Issue #1821, opened 2023-10-17
Tailwind text colors set via ".classes" selectively don't work for buttons


Rinzlov:
### Description

Trying to specify text color via ".classes" selector works well for labels and other elements but doesn't work for buttons when the color has a specified hue as can be seen on examples below:

Example code:
```py
from nicegui import ui

def main():
    ui.label('Label color works').classes("text-red-400")
    ui.button("Button color doesn't work").classes("text-red-400")
    ui.button("Unless it's pure color").classes("text-red")
    ui.run()

if __name__ in {"__main__", "__mp_main__"}:
    main()
```
Result:
![image](https://github.com/zauberzeug/nicegui/assets/1098586/37533911-dee2-4739-822c-6463cbb969ab)

### For a label the "text-red-400" tailwind class works:

![image](https://github.com/zauberzeug/nicegui/assets/1098586/b0f01da1-3f8d-43dc-9dd4-724af6d83c6c)

### For a button, the same class doesn't work:

![image](https://github.com/zauberzeug/nicegui/assets/1098586/d6602b35-efdd-43c9-86ae-b8f0898506f8)

### Although this way it works:

![image](https://github.com/zauberzeug/nicegui/assets/1098586/cb5d63b8-3aea-4d62-aed1-1a333a894d53)

Rinzlov:
Thank you for quick response.
This works for this use case but is there a general solution for other cases? I am guessing for other classes there won't be parameters that can be set to None.
My guess is that if it was easy to make all tailwind classes override the default styling of the component then it would already have been implemented? If not then maybe there could be a general UI setting "override_default_styles_with_the_ones_provided"?

I tried to use "replace" and "remove" options within the ".classes" and ".props" methods but had no luck so far. They are described in the documentation with "All three functions also provide remove and replace parameters in case the predefined look is not wanted in a particular styling." but I'm not sure if they can affect the default quasar styling?

natankeddem:
I am not sure what other cases you are referring to? We can try to handle them one at a time if you want to elaborate on them.

I am not a maintainer but, I would assume some of these elements have a `color` argument to allow a "Nice"™ default state to be used with `ui.colors`. As for why you can't remove those classes manually? NiceGUI can only remove classes that it inserted directly. In this case the `color` attribute actually propagates down to a quasar element `prop` for the button not a class. The prop then inserts those classes via quasar. Therefore instead of doing `color=None` you can also just remove the `color` prop.
```py
[...]

PR #2614, opened 2024-02-25
Introduce color module

falkoschindler:
This PR is an alternative, simplified implementation of PR #2537. It introduces a `color` module and allows to use color constants like this:

```py
from nicegui import color, ui

ui.icon('home', color=color.RED)
ui.icon('home', color=color.quasar.RED_5)
ui.icon('home', color=color.tailwind.RED_500)
```

Open tasks:
- [x] pytest
- [ ] documentation
- [ ] `color.RED` can result in Quasar color
- [ ] uppercase/lowercase?


rodja (approved):

falkoschindler:
I just noticed that
```py
ui.icon('home', color=color.RED)
```
results in a Quasar color, because Quasar's "red" has the same name. We could avoid that by translating to the hex value:
```py
RED = '#ff0000'
```

evnchn:
Shower (?, I'm in a bathtub rn) thoughs:

1. We should put this in context for a potential solution to #1821
2. Tailwind CSS V4 supports P3 colors. Let's not work on this PR before 3.X comes, or else we do twice the work for honestly little functionality improvement.

falkoschindler:
To be honest, like with #2613 I'm not certain if we really need this module. But at least there a fewer open questions. The module simply returns color strings.

Since this approach is so simple, I don't see much conflict with the discussion about button colors in #1821. If we'd come to the conclusion that you, e.g., should always make colors `!important` (hopefully we won't), we could easily do it by adjusting this module.

Regarding P3 colors, we could easily add them later like `color.tailwind_p3.RED_500` or `color.tailwind.p3.RED_500`.

evnchn:
All else not sure, but regarding P3 colors, I disagree with the approach `color.tailwind_p3.RED_500`. 

It is because, Tailwind CSS V4 promises that all existing code will go from sRGB to P3 colors without migration, by changing the definition of `text-red-500`, for example. 

In this case, `color.tailwind_p3.RED_500` would require manual migration. While changing `color.tailwind.RED_500` to refer to P3 colors could break existing code, namely Quasar if it isn't expecting it, or some color-parsing code. 

Therefore, it seems like if we want to introduce this module before 3.X, we are adding trouble. 

---

Perhaps this gets the same fate to be closed, as with #2613

falkoschindler:
Oh I see. Maybe it's indeed more similar to #2613 than I thought: It tries to hide some not-so-stable API behind a wrapper, which can add more harm than good.

falkoschindler:
Ok, let's close this PR for now. If we decide to give it another try, we can always come back an restore the branch.

PR #2620, opened 2024-02-26
Add a demo for custom icon set.

me21:
Add a demo for using [Themify](https://themify.me/themify-icons) icons.

---

See discussion #2617


falkoschindler (approved):
Thanks for the quick PR, @me21!
I even noticed that we can further simplify the Eva demo using `ui.icon` instead of `ui.element`. 🙂

PR #2623, opened 2024-02-26
Add webserial demo

BlankAdventure:
**WebSerial Demo**

This an example demonstrating the use of the JavaScript webserial interface to provide in-browser access to a local serial port. This demo interfaces with a PyBadge board and shows a live plotting of temperature (sending data from board to PC/app) and provides buttons to toggle the NeoPixel LEDs ( PC/app sending data to board).

There has already been some prior discussions on this project, here:

https://old.reddit.com/r/nicegui/comments/19e9r4n/triggering_nicegui_events_from_javascript/kjlodns/

https://github.com/zauberzeug/nicegui/discussions/334

https://github.com/zauberzeug/nicegui/wiki#community-projects

I have created a draft pull request to incorporate this webserial demo into the official collection of nicegui demos. There is probably some code cleanup to be done, and some discussion to generalize the embedded side. I'm happy to work through such suggestions.


rodja:
Thanks for the code @BlankAdventure. I'm looking forward to integrate this example as soon as it is ready. On a first glance I would like to get rid of the Arduino code and any dependencies to actual hardware. Maybe you could look into using serial simulators like socat (Linux,Mac, WSL) or com0com (Windows) to make it easier to run for a broad audience?

falkoschindler:
Thanks, @BlankAdventure!
I like the combination with Arduino, since it's a pretty popular framework. Sure, you need real hardware to try it yourself, but we also have other examples that are not easily accessible for everyone, like the ROS2 example.

But I would like to reduce the example to an LED and a button, just enough to demonstrate bidirectional communication. Currently you need an Adafruit Arcada board with a light sensor (I guess), which is not very common.

BlankAdventure:
@rodja Thinking more about it the serial simulation approach might not be ideal. The UI could feature buttons to toggle an LED (browser to board) and indicate a button press (board to browser) but these commands would just disappear into the virtual COM port. For more interactivity, one could envision adding a mock board UI featuring an LED element and button element, with the two UIs now linked through the virtual COM ports, but this seems to be going in the wrong direction of a simple demo.

I think @falkoschindler's suggestion may be cleaner. Keep the hardware, but remove the PyBadge specifics (NeoPixels and temperature sensor) and instead just stick with LED and push-button actions.

rodja:
>  but these commands would just disappear into the virtual COM port

Well, I would expect them to be printed in the console where socat is running. But let's move forward with a simple Arduino example. Maybe we'll add a variant for socat/com2com later...

BlankAdventure:
> > but these commands would just disappear into the virtual COM port
> 
> Well, I would expect them to be printed in the console where socat is running. But let's move forward with a simple Arduino example. Maybe we'll add a variant for socat/com2com later...

You are right that they would be printed to the console, and you could also send a command back to the UI - I just didn't think this would make for a very interesting demo!

BlankAdventure:
Alright, pushed changes per earlier discussion. In the GUI, buttons can be used to toggle and LED on/off (example of sending data to the board). Pushing a button on the board* triggers a message display in the GUI (example of receiving data)

*Arduino code was similarly updated to monitor the serial connection and pull an LED high/low as needed, or send the button count out. I suppose a schematic would be helpful as well.

falkoschindler (approved):
Thanks, @BlankAdventure!
I just reviewed your pull request, simplified the code a bit and tested the example on an ESP32 I had lying around. Looks good to me. 👍🏻

falkoschindler:
@rodja To get this pull request finally merged, I would leave it like this without the extra hassle of a simulation mode. What do you think?

rodja (approved):
Nice!

PR #2655, opened 2024-03-03
Add demo for altering the number of steps in a ui.stepper

rodja:
This pull request demonstrates the creation of a new `ui.step` depending on the value of a checkbox.


falkoschindler (commented):
Very nice. I just don't understand why the section is called "Storage".

falkoschindler (approved):

Issue #2556, opened 2024-02-15
Ag-Grid timeout error


ducnva:
### Description

Hi everyone,
I am encountering a timeout error when using the getDisplayedRowAtIndex API. Despite increasing the timeout to 600 seconds, I continue to encounter this error. What should I do? Can someone help me identify the cause of this error?"
`rowNode = await grid.run_grid_method('getDisplayedRowAtIndex', 0, timeout= 600)`
The message error:
```
JavaScript did not respond within 1.0 s
Traceback (most recent call last):
  File "/test/test/vendors/nicegui/nicegui/events.py", line 406, in wait_for_result
    await result
  File "/test/test/test.py", line 63, in detect_row_change
    rowNode = await grid.run_grid_method('getDisplayedRowAtIndex', rowIndex, timeout= 600)
  File "/test/test/vendors/nicegui/nicegui/client.py", line 211, in send_and_wait
    raise TimeoutError(f'JavaScript did not respond within {timeout:.1f} s')
TimeoutError: JavaScript did not respond within 600.0 s
```

dr-yd:
Same issue here, for pretty much all grid API methods. (Not ALL of them, though - `getDisplayedRowCount` works in the same place.) Example code:

```python
from nicegui import ui

grid = ui.aggrid({
    'columnDefs': [{'field': 'name', 'checkboxSelection': True}, { 'field': 'age'}, ],
    'rowData': [ {  'name': 'Alice',  'age': 18  },   {  'name': 'Bob', 'age': 21  }, ],
    ':getRowId':
    '(params) => params.data.name',
})
async def focus_handler(event):
    print(await grid.run_grid_method('getDisplayedRowAtIndex', event.args['rowIndex']))
grid.on('cellFocused', lambda e: focus_handler(e))
ui.run()
```

falkoschindler:
Thanks for reporting this issue, @ducnva and @dr-yd!
There is a JavaScript error "Maximum call stack size exceeded", which could be caused by cycles in the return data. But I'm not sure.

dr-yd:
For comparison, you can also try `getFocusedCell`, also breaks - maybe even just iterating through all the grid API methods and running them would emerge some kind of pattern?

falkoschindler:
I _assume_ every method returning cells or rows will have the same problem.

lapnd:
This is because aggrid return object that have recursive data which leads the function hasBinary raise that exception.
In the `templates/index.html`
```js
[...]

PR #2671, opened 2024-03-06
Add `return_exclusions` to `runJavascript()`

Pikaro:
Due to potential recursion in the objects returned by runJavascript, it is necessary to remove parts of the tree without modifying the objects in place. Added an optional argument that allows the exclusion of data from the return value in a dotted-path notation.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -173,7 +173,8 @@ class Client:
 
     def run_javascript(self, code: str, *,
                        respond: Optional[bool] = None,  # DEPRECATED
-                       timeout: float = 1.0, check_interval: float = 0.01) -> AwaitableResponse:
+                       timeout: float = 1.0, check_interval: float = 0.01,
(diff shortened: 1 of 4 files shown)


Pikaro:
Note: Might need to be formatted slightly because I couldn't get VSCode autoformatting to work and autopep8 produced different results from the CLI.

falkoschindler:
Thanks for the pull request, @Pikaro!

The formatting is no problem. We can fix that later - if necessary.

I'm a bit undecided whether the API change is worth it, or if there is another, more intuitive and/or more flexible solution to the problem. I left some more thoughts here: https://github.com/zauberzeug/nicegui/issues/2556#issuecomment-1981878154.

falkoschindler:
Thanks again, @Pikaro!
But I'll close this PR in favor of PR #2705 with a simpler yet more flexible API.

Issue #1869, opened 2023-10-24
Log viewer inside tabs issue


marcin-koziol:
### Description

I want to place the Log component in the tab, everything is OK, but to see the logs, you first need to switch to the Log viewer tab

```python
from datetime import datetime
from nicegui import ui


with ui.row():
    with ui.tabs().classes('w-full').props(add="dense align=left") as tabs:
        one = ui.tab('Something')
        logs = ui.tab('Logs')
    ui.separator()

    with ui.tab_panels(tabs, value=one).classes('w-full').style('min-height: 268px'):
        with ui.tab_panel(one):
            ui.image('https://picsum.photos/id/377/640/360').style('max-width: 420px')
        with ui.tab_panel(logs):
            log = ui.log(max_lines=10).classes('w-full h-20')


with ui.row():
    ui.button('Log time', on_click=lambda: log.push(datetime.now().strftime('%X.%f')[:-5]))

ui.run()
```

In this example:
 * run
 * click on "Log time" button
 * switch to the "Logs" tab - no entries here
 * click again
 * enties are visible

falkoschindler:
Wow, thanks for reporting this issue, @marcin-koziol!

It seems like the `log` element doesn't exist before opening the tab. When I print its ID `print(log.id)` (→"13") and try to get the element from the JavaScript console `getElement(13)`, it is `undefined.

Right now I'm not sure how to fix it. When calling `log.update()` right after pushing new content, it works. But the main idea of `ui.log` was to _not_ update the whole element for each new line. I guess there's a better way...

marcin-koziol:
maybe we could add the ability to bind a variable to the log (same as for inputs), so we'll be able to push logs to this var , without having rendered component? Is the truncating logic implemented only on the client?

falkoschindler:
I still have no idea how to fix this issue. Therefore I'll remove the milestone for now.

falkoschindler:
We have spotted a similar problem with AG Grids:
```py
with ui.tabs().classes('w-full') as tabs:
    one = ui.tab('One')
    two = ui.tab('Two')
with ui.tab_panels(tabs, value=one).classes('w-full'):
    with ui.tab_panel(one):
        async def get_data():
            ui.notify(await grid.get_client_data())
        ui.button('Get data', on_click=get_data)
    with ui.tab_panel(two):
        grid = ui.aggrid({
            'columnDefs': [{'headerName': 'Name', 'field': 'name'}],
[...]

PR #2686, opened 2024-03-11
New `ui.log` implementation

falkoschindler:
This PR attempts to solve #1869 by re-implementing the whole element. By simply adding new lines as `ui.label` elements, the log doesn't need keep track of its state and can simply use NiceGUI's core mechanics.

Open tasks:

- [x] Because the new log element derives from `ui.scroll_area`, it collapses to zero width by default and the user needs to specify some width. This breaks existing layouts. Can we find a solution to preserve backwards compatibility? Or do we need to target 2.0?
- [x] auto-scroll to bottom after initial push
- [x] stop auto-scrolling when scrolling up (#2690)


falkoschindler:
The new `ui.log` feels quite nice. My only concern is that adding a line "New log" now causes an update with the following payload:

```json
[
  "update",
  {
    "5": {
      "id": 5,
      "tag": "div",
      "class": [
        "nicegui-log"
      ],
      "style": {},
      "props": {
        "key": 5
      },
      "text": null,
      "slots": {
        "default": {
          "template": null,
          "ids": [
            6,
            7,
            8,
            9,
            10,
            11,
            12,
            13,
            14,
            15,
            16,
            17,
            18,
            19,
            20,
            39
          ]
        }
      },
      "events": [],
      "component": null,
      "libraries": []
    },
    "39": {
      "id": 39,
      "tag": "div",
      "class": [],
      "style": {},
      "props": {
        "key": 39
      },
      "text": "New log",
      "slots": {
        "default": {
          "template": null,
          "ids": []
        }
      },
      "events": [],
      "component": null,
      "libraries": []
    }
  }
]
```

In comparison, this was the update message for the old `ui.log`:

```json
[
  "run_javascript",
  {
    "code": "return runMethod(5, \"push\", [\"New%20log\",16])"
  }
]
```

Warchant:
How to get logs content?
I want to add a button which saves this logs as a file.

falkoschindler:
@Warchant You can get the content by iterating over its nested labels:
```py
ui.button('Print log', on_click=lambda: print('\n'.join(label.text for label in log)))
```

PR #2707, opened 2024-03-15
docs: enhance macos packaging instructions

gotev:


falkoschindler:
Hi @gotev, thanks for your pull request!

Just to clarify:

1. You removed "Mac M1" from the title. We thought these instructions solve issues with Apple Silicon hardware. Do you think it also applies to Intel Macs?
2. We added the statement about the `--windowed` parameter to address a specific issue some users where facing. Are you sure it really isn't needed anymore?

gotev:
@falkoschindler thanks to you and all the contributors for this amazing project! 

1. Yes, in the thread where this originated, the evidence I shared was from an Intel Mac, so the issue does not apply to only Apple Silicon Macs
2. Without the windowed flag, pyinstaller don't produce an .app bundle, but a mere executable you can launch only from cli. Not a viable option in real world scenarios. By ensuring the freeze support statements are on top of the main python script, you can safely pass --windowed option without process spawning

Let me know if the above is sufficient or you need further information or data.

falkoschindler (approved):
Thanks, @gotev! Sounds good to me. Let's merge! 🙂

PR #2708, opened 2024-03-15
Event registration, continued :)

petergaultney:
Here are what I believe to be the basic necessary code changes based on the list you made.

- [x] button
- [x] value elements
- [x] color_picker
- [x] echart
- [x] interactive_image
- [x] joystick
- [x] json_editor
- [x] keyboard
- [x] list-item (on_click)
- [x] menu_item
- [x] scene
- [x] scroll_area
- [x] table
- [x] tree
- [x] upload

Not included so far:
- changes to documentation
- new tests

~I had one error in `test_keyboard.py`~ EDIT: this turned out to be because I use a non-standard keyboard remapping tool, and all my `t` keys were getting remapped (i guess inside chromedriver?) into `f`s. :D

All the existing tests pass locally.

--- 

Also, I could use a sanity check on my autopep8/isort situation. I tried running it on the whole repo as per the readme and it reformatted way too many things. It also reformatted an import that was 121 characters long, which seems like a good thing, but unexpected since it was already part of the codebase.

### DIFF
--- a/nicegui/elements/json_editor.py
+++ b/nicegui/elements/json_editor.py
@@ -4,8 +4,7 @@ from typing_extensions import Self
 
 from ..awaitable_response import AwaitableResponse
 from ..element import Element
-from ..events import (GenericEventArguments, JsonEditorChangeEventArguments,
-                      JsonEditorSelectEventArguments, handle_event)
+from ..events import GenericEventArguments, JsonEditorChangeEventArguments, JsonEditorSelectEventArguments, handle_event
 
 
 class JsonEditor(Element, component='json_editor.js', exposed_libraries=['lib/vanilla-jsoneditor/index.js']):
@@ -22,8 +21,8 @@ class JsonEditor(Element, component='json_editor.js', exposed_libraries=['lib/va
         After data has changed, call the `update` method to refresh the editor.
 
         :param properties: dictionary of JSONEditor properties
-        :param on_select: callback function that is called when some of the content has been selected
-        :param on_change: callback function that is called when the content has changed
+        :param on_select: callback which is invoked when some of the content has been selected
+        :param on_change: callback which is invoked when the content has changed
         """
         super().__init__()
         self._props['properties'] = properties
--- a/nicegui/elements/json_editor.py
+++ b/nicegui/elements/json_editor.py
@@ -1,8 +1,11 @@
-from typing import Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional
+
+from typing_extensions import Self
 
 from ..awaitable_response import AwaitableResponse
 from ..element import Element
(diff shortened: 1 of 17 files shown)


nicegui/elements/json_editor.py
```diff
@@ -1,8 +1,11 @@
-from typing import Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional
+
+from typing_extensions import Self
 
 from ..awaitable_response import AwaitableResponse
 from ..element import Element
-from ..events import GenericEventArguments, JsonEditorChangeEventArguments, JsonEditorSelectEventArguments, handle_event
+from ..events import (GenericEventArguments, JsonEditorChangeEventArguments,
```
petergaultney, line 7: this formatting was not intentional, but it did happen when i ran autopep8, and it doesn't seem _wrong_, either - the line was previously 121 characters long, which is over the 120 character limit?

nicegui/elements/mixins/value_element.py
```diff
@@ -32,9 +32,10 @@ def handle_change(e: GenericEventArguments) -> None:
             self._send_update_on_value_change = True
         self.on(f'update:{self.VALUE_PROP}', handle_change, [None], throttle=throttle)
 
-    def on_value_change(self, callback: Callable[..., Any]) -> None:
-        """Set a callback to be invoked when the value changes."""
+    def on_value_change(self, callback: Callable[..., Any]) -> Self:
+        """Add a callback to be invoked when the value changes."""
```
petergaultney, line 36: I've changed your phrasing here to be "Add", since I believe that's a little more accurate than "Set".

petergaultney:
@falkoschindler no pressure, but i want to make sure you know that i'm not planning on working on tests or docs until you're happy with what you see here. 

I would of course love to have this as part of an official release, though, so I'm ready to move forward whenever you get some free time.

falkoschindler:
Thanks, @petergaultney! That totally makes sense. I'll have a look in the next few days. 👍🏻

nicegui/elements/json_editor.py
falkoschindler: Yeah, that's strange. See https://github.com/zauberzeug/nicegui/issues/2750.

falkoschindler (approved):
Very nice, @petergaultney!
I just review it and basically just unified the wording of the docstrings and variable names a little. There was only one tiny bug in `ui.scroll_area` where `callback` is passed as a parameter but `on_scroll` was used inside the method.
Let's merge it!

petergaultney:
> Very nice, @petergaultney! I just review it and basically just unified the wording of the docstrings and variable names a little. There was only one tiny bug in `ui.scroll_area` where `callback` is passed as a parameter but `on_scroll` was used inside the method. Let's merge it!

thanks for the catch!

Issue #2570, opened 2024-02-19
404 error page not rendered on static file routes


falkoschindler:
### Description

As noticed by @Anindya088 (#2568), the 404 error page isn't rendered correctly when accessed via a (sub-)path of a static route. Take this code for example:

```py
app.add_static_files('/static', '.')
ui.link('Static', '/static')
```

When visiting `/static` or `/static/subpage`, the HTML page is displayed, but without any CSS and JS assets. They are requested from a wrong path and, thus, not used for rendering the page.

rodja:
Thanks for the PR, @Lucas-lyh. It made me dig a bit deeper into the issue. I think the problem was introduced with #848 which sets the path prefix to `scope['root_path']` -- which is correct in my opinion:

https://github.com/zauberzeug/nicegui/blob/cc7770e210b415409131cc83128d649ed838a7a0/nicegui/client.py#L116

But, `app.add_static_files` uses `app.mount` internally:

https://github.com/zauberzeug/nicegui/blob/cc7770e210b415409131cc83128d649ed838a7a0/nicegui/app/app.py#L156

This is defined in the ASGI standard to alter the `scope['root_path']` -- which is not desired in this case. Because there is no  way to tell the FastAPI/Starlette (which implements ASGI) to not alter `scope['root_path']`, we may be forced to implement `app.add_static_files` in some other way :-(

PR #2722, opened 2024-03-19
fix: fix 404 page bug when accessing wrong static files.

Lucas-lyh:
This #2570 issue is because starlette modify `request.route['root_path']` when routing static files mount.  I just reset this field when handling 404 page request. 
I'm a beginner in nicegui and may miss something. On my machine, all test passed and the bug is fixed.
I also add a test for 404 page.

### DIFF
--- a/nicegui/app/app_config.py
+++ b/nicegui/app/app_config.py
@@ -36,6 +36,7 @@ class AppConfig:
     prod_js: bool = field(init=False)
     show_welcome_message: bool = field(init=False)
     _has_run_config: bool = False
+    root_path: str = field(init=False)
 
     def add_run_config(self,
                        *,
@@ -50,6 +51,7 @@ class AppConfig:
                        tailwind: bool,
                        prod_js: bool,
                        show_welcome_message: bool,
+                       root_path: str,
                        ) -> None:
         """Add the run config to the app config."""
         self.reload = reload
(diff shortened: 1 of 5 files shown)


rodja:
Interesting. Will it still work with subpath mounting? Like in our example https://github.com/zauberzeug/nicegui/tree/main/examples/nginx_subpath?

Lucas-lyh:
> Interesting. Will it still work with subpath mounting? Like in our example https://github.com/zauberzeug/nicegui/tree/main/examples/nginx_subpath?

Yes! 
I have to confess that I forgot to exam this situation. In my test, subpath mounting works well because prefix is assigned to `X-Forwarded-Prefix` in headers if exist, instead of `route['root_path']`. In fact because of that, 404 page bug is masked when working with subpath mounting.

Lucas-lyh:
Updated 404_test to prevent other test from being affected.

rodja:
Cool. And what about sub-mounting in another app as we do in the fastAPI example: https://github.com/zauberzeug/nicegui/tree/main/examples/fastapi

Lucas-lyh:
> Cool. And what about sub-mounting in another app as we do in the fastAPI example: https://github.com/zauberzeug/nicegui/tree/main/examples/fastapi

Thank you for you reminder, my fix gose wrong when working with app.run_with(). 
How about add a AppConfig field to save the correct root path, then use this path to generate 404page? I've tried to solve it but I'm not sure if it is reasonable.

falkoschindler:
Thank you so much for your contribution, @Lucas-lyh!
Your pull request sparked the idea of @rodja's PR #2746, which seems to have solved issue https://github.com/zauberzeug/nicegui/issues/2570. 🚀

PR #2733, opened 2024-03-20
run io_bound and cpu_bound maintain type signatures of their callbacks

petergaultney:
This is a pretty simple way for `nicegui` to not 'drop' the types of callbacks that need to get run in the background for whatever reason. I know `mypy` supports this, and I'm pretty sure other type checkers do too. It's an official part of the type system.

For this code:

Screenshot 2024-03-20 at 11 08 40 AM

we go from 

Screenshot 2024-03-20 at 11 09 13 AM

to 

Screenshot 2024-03-20 at 11 10 00 AM

And (although i have no accompanying screenshots) you would also get type checking on your inputs to the function.

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -2,17 +2,22 @@ import asyncio
 import sys
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from functools import partial
-from typing import Any, Callable
+from typing import Any, Callable, TypeVar, cast
+
+from typing_extensions import ParamSpec
 
 from . import core, helpers
 
 process_pool = ProcessPoolExecutor()
 thread_pool = ThreadPoolExecutor()
 
+P = ParamSpec('P')
+R = TypeVar('R')
 
-async def _run(executor: Any, callback: Callable, *args: Any, **kwargs: Any) -> Any:
+
+async def _run(executor: Any, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
     if core.app.is_stopping:
-        return
+        return  # type: ignore  # the assumption is that the user's code no longer cares about this value
     try:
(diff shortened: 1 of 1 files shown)


nicegui/run.py
```diff
@@ -21,9 +26,10 @@ async def _run(executor: Any, callback: Callable, *args: Any, **kwargs: Any) ->
             raise
     except asyncio.exceptions.CancelledError:
         pass
+    return cast(R, None)  # the assumption is that the user's code no longer cares that this is a lie
```
petergaultney, line 29: i don't actually know what the intent of just ignoring CancelledError was, but i'm assuming it's similar to "the app is stopping" case above. At least, it seems unlikely that the intent is that you might occasionally get `None` when awaiting `run.x_bound` ?

falkoschindler: > i don't actually know what the intent of just ignoring CancelledError was, but i'm assuming it's similar to "the app is stopping" case above.

I think so too.

falkoschindler (commented):
Thanks for this pull request, @petergaultney! I really like filling these gaps of missing type annotations.

Just one remark: I would prefer explicitly _ignoring_ the mypy warning (`# type: ignore`) instead of casting it to `R`, creating a "lie". What do you think?

petergaultney:
> Thanks for this pull request, @petergaultney! I really like filling these gaps of missing type annotations.
> 
> Just one remark: I would prefer explicitly _ignoring_ the mypy warning (`# type: ignore`) instead of casting it to `R`, creating a "lie". What do you think?

yeah, I can see how that might seem a little less confusing to people. I'll change that over.

falkoschindler (approved):
Thanks again, @petergaultney!
Let's merge! 🚀

PR #2746, opened 2024-03-22
implement static files serving without app.mount

### LINKED ISSUE #2570: 404 error page not rendered on static file routes

rodja:
This pull request fixes #2570 by not using `app.mount` to serve those files. See https://github.com/zauberzeug/nicegui/issues/2570#issuecomment-2014345947 for details of why I think this is the best solution.

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -153,7 +153,12 @@ class App(FastAPI):
         """
         if url_path == '/':
             raise ValueError('''Path cannot be "/", because it would hide NiceGUI's internal "/_nicegui" route.''')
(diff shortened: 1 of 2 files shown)


rodja:
@Lucas-lyh, it would be great if you could try this out for your scenarios.

falkoschindler (approved):

Issue #2611, opened 2024-02-24
ui.editor cursor jumps to start of text intermittently 


Gilser:
### Description

This is able to be replicated from the nicegui documentation pages in addition to in test applications.

To cause the issue, clear out any existing text in the ui.editor input control, then rapidly start typing characters into it. Especially when there are relatively few characters in the field, the cursor will jump to the beginning of the input which then causes text being typed to start going in before anything else inputted. 

Can happen a few times as text is being inputted but it seems to stop after some number of characters are already in the field. Fairly easy to trigger by clearing out all text before quickly inputting more.

Does seem to only be triggered when quickly entering information - by slowly inputting text I've been unable to replicate so far.

intrinsically:
i've been using nicegui to build some intranet apps. i don't get the cursor jumping problem when i'm deploying locally, but when i deploy on AWS behind a VPN then it's really really bad - cursor jumping every 5 characters or so even if i have existing text.

could it be latency related?

intrinsically:
as a further bit of info: i can reproduce the problem quite easily in the controls demo here: https://nicegui.io/documentation/editor

falkoschindler:
Thanks, @intrinsically!

I think I found a lead: When setting `LOOPBACK = False` and removing the following lines, the bug is gone.
https://github.com/zauberzeug/nicegui/blob/ee14497b1ee02c8f1dbff43964bc8d639704e116/nicegui/static/nicegui.js#L162-L164

Let me try to explain what I think is happening:

1. Usually value elements like `ui.switch` need to send an `update:modelValue` event to the server, which updates its model and sends the new value back to the front end. This is what we call "loopback".
2. For elements that can change rapidly like `ui.input` or `ui.number`, we disable this loopback procedure and set the new value directly on the client (see the code snippet above).
3. But there seems to be a third group of elements like `ui.editor` that shouldn't use loopback to prevent the cursor from jumping around, but also shouldn't set the model value programmatically on the client, because it also messes with the cursor position.

So I think we either need to change `LOOPBACK` from a boolean flag to something like a `Literal` string to distinguish these three types of value updates, or we add a second flag to define whether the new value should be set directly on the client or not.

intrinsically:
[...]

PR #2756, opened 2024-03-23
Fix jumping cursor for `ui.editor`

falkoschindler:
This PR tries to solve issue #2611 by disabling loopback _and_ disabling the "shortcut" which sets the new value directly on the client.


falkoschindler:
@rodja I improved the backward compatibility by switching back to a single `LOOPBACK` field which is now `Optional`. The new docstring describes its possible values. This way all value elements can remain unchanged, only the `ui.editor` gets `LOOPBACK = None` because it shouldn't receive updates from the server _nor_ set the value prop on the client.

rodja (commented):
Very cool. Great solution.

PR #2761, opened 2024-03-25
More elegant way to terminate On Air connection

rodja:
This pull-request get's rid of hacky signal handler registration to terminate the On Air connection. This also fixes issues with auto-reload when On Air is active.


falkoschindler (approved):
I can't really judge the difference, but the code looks plausible. 🤷🏻‍♂️
You can squash and merge, or we talk about it on Wednesday.

PR #2765, opened 2024-03-25
Adding upload_all argument for upload element

d-meehan:
Not sure what the most performant approach here is. I could remove the initial conditional check and have the lists populate regardless and only get used if the later conditional checks are met. Also had to adjust UploadEventArguments to accept lists, let me know if that's no good.

---

Feature request: #2698


falkoschindler (approved):
Thanks for the pull request, @d-meehan!
I just reviewed and improved the code a bit. Seems to work now. 👍🏻

rodja (approved):

PR #2054, opened 2023-11-21
FullCalendar Element

frankvp11:
Currently it's at the state where the element is working, however the documentation is not.
When I run the file websites/more_documentation/fullcalendar_documentation.py in it's current state, it works. However, when I comment out the last lines ( ui.run and main_demo()), and try to run the main.py file, it can't seem to find FullCalendar.
From the discord discussions, we believe this is because of a static file issue. 

Issues aside, in this current implementation, users can click on the events, add events "easily", and remove them aswell.
It also allows the user to choose their own options, a full list can be viewed here: https://fullcalendar.io/docs/view-specific-options
The basic ones though, are implemented in the documentation example.
Edit: One thing that I noticed is that I still need to remove the test_temp.py file (sorry about that, I will do that later today)

PR #2775, opened 2024-03-27
CodeMirror Element

frankvp11:
This is the PR that we discussed in #2760. Currently there are a few issues which I list below:
1. Sometimes it doesn't properly load all the resources, leading to the element not working properly - this can also be seen in the console, as it says CodeMirror is not defined
2. It currently does not yet support various things like 
   -Folding functions 
   -Autocomplete
and various other features that are supported.

Any help in making this work would be much appreciated.
One thing to note is that despite the fact that it logs that you are trying to fold it (when you do ctrl + ',' on the spot) it can't seem to figure out how to collapse it. The way that I am getting inspiration for this is from here https://codemirror.net/5/demo/folding.html. I realize that this is version 5, but I couldn't seem to figure anything else out that was semi-clear.

### DIFF
--- a/nicegui/elements/codemirror.js
+++ b/nicegui/elements/codemirror.js
@@ -1,0 +1,45 @@
+import { loadResource } from "../../static/utils/resources.js";
+
+export default {
+  template: `
+    <div style="width: 100%">
+      <textarea ref="myTextarea"></textarea>
+    </div>
+  `,
+  props: {
+    resource_path: String,
+  },
+  async mounted() {
+    // Load CodeMirror resources and initialize it
+    await this.$nextTick(); // NOTE: wait for window.path_prefix to be set
+    await Promise.all([
+      loadResource("https://cdnjs.cloudflare.com/ajax/libs/codemirror/6.65.7/codemirror.min.js"),
+      loadResource("https://cdnjs.cloudflare.com/ajax/libs/codemirror/6.65.7/codemirror.min.css"),
+      loadResource("https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.63.1/theme/dracula.min.css"),
+      loadResource("https://cdnjs.cloudflare.com/ajax/libs/codemirror/6.65.7/mode/python/python.min.js"),
+      loadResource("https://cdnjs.cloudflare.com/ajax/libs/codemirror/6.65.7/addon/edit/closebrackets.min.js"),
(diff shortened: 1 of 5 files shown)


falkoschindler:
Thanks, @frankvp11! I just fixed a few minor things, but it already looks quite nice. 🙂

frankvp11:
I agree. I have yet to figure out how to get the other features working, as unfortunately there aren't many good examples of how you could integrate them in version 6. 
I've also taken the liberty to start working on an Monaco integration aswell. While this one is more limited in terms of future features, it is ultimately easier to implement things.
Some of the differences in my the two elements (Monaco and CodeMirror) is as follows:
1. CodeMirror has the ability (potentially) to support collaborative editing.
2. Monaco has minimaps
3. While possible in CodeMirror, code folding comes out of the box with Monaco despite being difficult to implement (yet to do it) in CodeMirror.
4. Monaco has find and replace, and bracket/string matching "out of the box"

For the Monaco editor, please let me know if you want that in a seperate PR or if it can go in this PR aswell (as it pertains to code editors).

chrschorn:
I would be interesting in helping our here. Do we need/want all features supported by codemirror for a first version of this? I think even without e.g. folding this would be a really helpful component to have in NiceGUI as long as it is reliable (top comment says "Sometimes it doesn't properly load all the resources, leading to the element not working properly").

rodja:
I agree with @chrschorn. A minimal working version should be the goal of this PR. We can add code folding, auto-completion etc in later additions.

frankvp11:
I agree also. I think to solve the problem of the modules not loading properly, we should download the files similar to how we did in https://github.com/zauberzeug/nicegui/pull/2054 and then use the loadResource function. That should get rid of the issues that I was facing earlier where it would take a few refreshes for the issues to go away

chrschorn:
@frankvp11 I could spend some time looking into this later this week. Let me know if that makes sense or if you're working on this right now.

frankvp11:
That would be great thanks. I've got exams for the next 2 weeks so my development time has gone to zero :sweat_smile:

rodja:
I think we can close this in favour of https://github.com/zauberzeug/nicegui/pull/2913 which hopefully gets merged for the next release. What do you think @frankvp11?

chrschorn:
Thank you for getting this started btw @frankvp11!

PR #2784, opened 2024-03-29
Added Monaco Element and Example

frankvp11:
Currently you might be able to tell that there are basically duplicates of the files, one in example and one as an actual element.
For some reason, I can't figure out how to get the imports working, due to require not being allowed / use-able. 
So, currently running main.py won't work on the element that we are trying to implement. Note that this is purely because of the import issues I was facing, which prompted me to make the example under /examples.
If you run the example, (examples/monaco_editor/main.py, you can see what I'm after, and how it works. 
This example can be deleted once the issue with importing is fixed.
This PR is an extension of #2775 and discussion #2760, where in this PR I implemented the Monaco Code Editor. 
In the example, it has the following features:
1. Code Folding (ie, fold functions) 
2. Find and replace built in
3. Multiple-line selection built in
4. Minimap (toggle-able)
5. Themes
6. and much more

You might also notice that DEPENDENCIES.MD is updated - I didn't stage all the changes from running npm.py because it was ~200, none of which were relevant. However, DEPENDENCIES.MD auto-updates with these changes, and thus some other packages look like their dependency has been updated. This is worthwhile considering


frankvp11:
Might anyone have an idea of how I can get the import working? 
Also, I'm wondering on what we might do for features like the Code Diff. This would require a different setup in some sense, my personal opinion is that we make a seperate class for that, but I'm not so sure. 
Let me know. Thanks

rodja:
Thanks for the effort @frankvp11. We would love to see this integrated but do not find the time to delve into the PR at the moment. Is anyone else able to push this further with @frankvp11?

Issue #2549, opened 2024-02-13
PyInstaller not found


hrauch:
### Description

I installed nicegui, pyinstaller and pywebview in a virtual environment with Python 3.12 and Windows 11. 
Trying your example on **https://nicegui.io/documentation/section_configuration_deployment** I get this error message:
**C:\Program Files\Python312\python.exe: No module named PyInstaller**
If I change the last line of your build.py script in
**subprocess.call(cmd, shell=True)**
everything is ok. May be that this modification is only needed with Windows 11.

falkoschindler:
Thanks for reporting this issue @hrauch!

Can anyone with Windows 11 reproduce the problem? Should we simply add `shell=True`?

JS-Aibel:
Not able to reproduce the problem.
Example [Configuration & Deployment | NiceGUI](https://nicegui.io/documentation/section_configuration_deployment#package_for_installation) OK for Windows 11.
I have nicegui and pyinstaller installed, not pywebview.

alimate2023:
If I use my virtual environment, it says I can't find the package. Try installing pyinstall and packaging it without the virtual environment.

gotev:
Been able to reproduce that on Windows 10 running a Python 3.12.2 virtual envronment and to find a solution. My multiplatform build script looks like this (Linux can be added as well):
```python
import platform
import os
import shutil

APP_NAME = 'Your App Name'

def remove(folder):
    if os.path.exists(folder):
        shutil.rmtree(folder)

def build_windows():
    return [
        'pyinstaller',
        'main.py',  # your main file with ui.run()
        '--name', APP_NAME,  # name of your app
        '--onedir',
        '--windowed',
        '--icon', 'AppIcon.ico'
    ]

def build_macos():
    return [
        'python',
        '-m', 'PyInstaller',
        'main.py',  # your main file with ui.run()
        '--name', APP_NAME,  # name of your app
        '--windowed',
        '--icon', 'AppIcon.icns'
    ]

build_matrix = {
    'Windows': build_windows,
    'Darwin': build_macos
}

if __name__ == "__main__":
    platform_name = platform.system()
    
    if platform_name not in build_matrix:
        print(f'Unsupported platform {platform_name}')
        sys.exit(1)

    subprocess.call(build_matrix[platform_name]())
```

falkoschindler:
Thanks @gotev!
[...]

PR #2789, opened 2024-03-30
#2549 feat: added build package script

gotev:
@falkoschindler added the script as per your instructions in #2549. The script is meant to be copied and used as a base package build script for nicegui users, who will need to also modify it for their own specific usage (e.g. adding resources, icons and so on)


falkoschindler (commented):
Thank you for the pull request, @gotev!

I just reviewed the code and refactored it a bit to organize common and platform-specific arguments more intuitively. And I introduced argparse to allow configuring the parameters from outside.

What do you think so far?

I wonder if we really need to call `pyinstaller` without `python -m` on Windows... And what's the reason for the extra `--onedir`? There might already be an explanation in one of the issues and discussions, but I honestly lost track of every detail. Good to have this script to capture them all! 🙂

falkoschindler:
@gotev You're right, we should be careful not to suggest or promise to wrap PyInstaller in all its glory, but to only give a starting point from which our users can deviate. I just thought it would be nice to have an executable documentation for how to package a NiceGUI app that is actually shipped with the library and which you can simply call for simple use cases.

(In the future) we can use it to guide the user to the most important arguments to consider and include some [hints](https://nicegui.io/documentation/section_configuration_deployment#package_for_installation), either in the help text or in form of console output, like "1. Make sure to set native=True...", "2. Note about packaging on macOS: ..." and so on.

Regarding the `--add-data` argument: It should already work as expected, i.e. `nicegui` is always added while you can additional data:

```bash
$ ./pack.py --add-data /foo --add-data /bar --dry-run main.py
python -m PyInstaller --name Your App Name --add-data /Users/falko/Projects/nicegui/nicegui --add-data /foo --add-data /bar --windowed main.py
```

gotev:
Got it, so let's proceed then. If we have a way to validate what the user is passing in `ui.run` before starting the packaging, it will be great, so we could simply fail the build printing out messages like `native=True is required` and so on.

Another useful thing we could add is to print out the full PyInstaller command before executing it, so it's clear we're providing a helper rather than a wrapper, reducing frictions for simple use cases and already giving the starting command for more complex scenarios.

falkoschindler:
I like the idea of always printing the command!
Detecting whether `native=True` is set could be challenging though. Users could make it dependent on some other flag like `native=PRODUCTION`, so that scanning for the string "native=True" is not enough. But we can postpone this feature and just print some short info text for now.

gotev:
Agreed!

As an idea to achieve that, we could detect if the app is running inside a pyinstaller package at runtime:

```python
def is_pyinstaller():
    return getattr(sys, 'frozen', False) or hasattr(sys, '_MEIPASS')
```
and validate run params. But that's for another PR 😉

falkoschindler:
I just looked into this PR and noticed some remaining points we need to address:

- [ ] Regarding our idea about checking for `native=True`: The [documentation](https://nicegui.io/documentation/section_configuration_deployment#package_for_installation) recommends using `--windowed` only with `native=True`. Conversely, we don't need `native=True` if `--windowed` is not set. So I wonder if we really should set `--windowed` by default. Maybe we don't set it, but add a new argument `--windowed` for pack.py with a description taken from the NiceGUI documentation.

- [ ] Same for `--onefile`: We should add it as an argument with a description from the NiceGUI documentation.

- [ ] We should add the remaining packaging tips to the help text.

- [ ] We should warn about not packaging with `reload=True`.

falkoschindler:
Ok, I added some more documentation to the pack.py script. Let's merge and improve it later.

PR #2800, opened 2024-04-01
add efficient_chat_with_ai

merlinarer:
add an  efficient chat_with_ai example discussed in [here](https://github.com/zauberzeug/nicegui/discussions/2772)

### DIFF
--- a/examples/efficient_chat_with_ai/main.py
+++ b/examples/efficient_chat_with_ai/main.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+from typing import List, Tuple
+
+from langchain.chains import ConversationChain
+from langchain.chat_models import ChatOpenAI
+from nicegui import context, ui
+OPENAI_API_KEY = 'not-set'  # TODO: set your OpenAI API key here
+
+
+class CustomChatMessage:
+    def __init__(self, cls):
+        self.pointer = 0
+        self.last_message = None
+        self.spinner = None
+        self.cls = cls
+
+    def chat(self, name, text):
+        with self.container:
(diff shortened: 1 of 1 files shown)


falkoschindler:
Thanks for the pull request, @merlinarer!

In order to keep it simple and more close to the original example, I updated the latter with the idea of incrementally building the UI from yours. Unfortunately, I'm always getting a rate limit when testing the UI. Can you or @rodja try if it works with your key?

rodja:
I made some changes to display error messages and fixed getting the content from a chunk. But somehow the response is still not streamed.@merlinarer, could you look into this?

merlinarer:
@rodja Have you tried my commit? I had already made it into streaming style.

merlinarer:
@rodja Have you tried my commit? I had already made it into streaming style.

merlinarer:
@rodja Have you tried my commit? I had already made it into streaming style.

merlinarer:
@rodja Have you tried my commit? I had already made it into streaming style.

rodja:
@merlinarer Yes, in `examples/efficient_chat_with_ai` the streaming works. @falkoschindler tried to move the ideas to `examples/chat_with_ai` because we rather would like to have only one very well working example. Maybe you can have a look at the implementation?

merlinarer:
@rodja @falkoschindler I abandoned `ConversationChain`, which would have some conflict with stream mode. Moreover, using `config={"callbacks": [NiceGuiLogElementCallbackHandler(log)]` is an alternative to set callbacks.

falkoschindler (approved):
Great! It works really nicely now. 🙂
Thanks a lot for the contribution, @merlinarer!

Issue #278, opened 2023-01-21
Mermaid in markdown


falkoschindler:
### Discussed in https://github.com/zauberzeug/nicegui/discussions/277

<div type='discussions-op-text'>

<sup>Originally posted by **SebastienDorgan** January 21, 2023</sup>
Hi,
I am am really new to nicegui and I try to figure out how to make mermaid work with ui.markdown elements.
Here is what I tried:
```python
async def content() -> None:
    mermaid_header =  """\
    <style>
    .mermaid-pre {
        visibility: hidden;
    }
    </style>
    """
    ui.add_head_html(mermaid_header)
    mermaid_footer = """\
        <script type="module" defer>
        import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@9/dist/mermaid.esm.min.mjs';
        mermaid.initialize({
            securityLevel: 'loose',
            startOnLoad: true
        });
        let observer = new MutationObserver(mutations => {
            for(let mutation of mutations) {
            mutation.target.style.visibility = "visible";
            }
        });
        document.querySelectorAll("pre.mermaid-pre div.mermaid").forEach(item => {
            observer.observe(item, { 
            attributes: true, 
            attributeFilter: ['data-processed'] });
        });
        </script>
        """
    ui.add_body_html(mermaid_footer)
    ui.label("Streams")
    content = """\
    ```mermaid
    graph TD;
        A-->B;
        A-->C;
        B-->D;
        C-->D;
    ```
    """
    ui.markdown(content , extras=['fenced-code-blocks', 'mermaid'])
```
unfortunately the result is not what I expected to be:
![image](https://user-images.githubusercontent.com/12728415/213860828-768c6703-e436-49e2-b7ea-4da915dfb604.png)
Can you help me?
</div>

---

We identified a bug in markdown2, but found a workaround. So we could extend `ui.markdown` to support Mermaid.

- [ ] update to markdown2 version 2.4.7
- [ ] let `ui.markdown` insert additional head and body HTML blocks (only once per client) if "mermaid" extra is used
- [ ] additional `ui.mermaid` element
- [ ] example(s) for API reference
- [ ] pytest?

falkoschindler:
I noticed that the workaround of adding the "mermaid" class ourselves won't work in general since we can't know which pre tag is Mermaid and which is not.

But markdown2 already has a solution which only needs to get released to PyPI. So we'll wait until this fix is out.

https://github.com/trentm/python-markdown2/issues/495

[...]

PR #2803, opened 2024-04-02
Update Markdown elements even if invisible (e.g. inside tooltip or tab)

### LINKED ISSUE #2779: Tooltip contents are not preserved
### Description

I'd like to use a textarea to edit Markdown and to display Markdown content in a tooltip:

```python
        with ui.textarea('Инструкции').bind_value(form, 'instructions') as textarea:
            with ui.tooltip():
                markdown = ui.markdown()
                textarea.bind_value_to(markdown, 'content', lambda t: t or '')  # doesn't like None as content
```

It mostly works, but the tooltip content is lost if I move the mouse out of textarea and then back. The empty tooltip appears, but after editing the textarea, its contents are updated properly.

me21:
This is an attempt to fix #2779. It seems to make Markdown elements update correctly but breaks Mermaid diagrams in Markdown elements, not sure why.
Additional changes are welcome.

### DIFF
--- a/nicegui/elements/markdown.js
+++ b/nicegui/elements/markdown.js
@@ -4,7 +4,7 @@ export default {
     this.ensure_codehilite_css();
     if (this.use_mermaid) {
       this.mermaid = (await import("mermaid")).default;
-      this.update(this.$el.innerHTML);
+      this.renderMermaid();
     }
   },
   data() {
@@ -12,9 +12,11 @@ export default {
       mermaid: null,
     };
   },
+  updated() {
+    this.renderMermaid();
+  },
   methods: {
-    update(content) {
-      this.$el.innerHTML = content;
+    renderMermaid() {
       this.$el.querySelectorAll(".mermaid-pre").forEach(async (pre, i) => {
         await this.mermaid.run({ nodes: [pre.children[0]] });
       });
--- a/nicegui/elements/markdown.py
+++ b/nicegui/elements/markdown.py
@@ -35,7 +35,6 @@ class Markdown(ContentElement, component='markdown.js'):
         if self._props.get('innerHTML') != html:
             self._props['innerHTML'] = html
             self.update()
-            self.run_method('update', html)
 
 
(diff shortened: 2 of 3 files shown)


me21:
Mermaid diagrams work now.

me21:
There is an annoying blinking during the update of the content if Mermaid diagram is shown. However, this bug is not so major, I think, and Markdown elements now are correctly shown in tooltips :)

me21:
I'm sorry, I force-pushed new fix - it appears to be a one-liner 😄 
But it introduces blinking of visible content if Mermaid diagram is displayed inside Markdown element.

falkoschindler:
Wow, that's a nice solution!
I just can't reproduce the blinking - or I don't know where to look. Do you have some example code to reproduce it?

me21:
The following minimal example reproduces blinking for me:

```
textarea = ui.textarea()
markdown = ui.markdown(extras=['fenced-code-blocks', 'tables', 'mermaid'])
markdown.bind_content_from(textarea, 'value')
```

Now type the following into textarea:

````
```mermaid
graph LR;
    A-->B;
    A-->C;
```
````

You should see Mermaid diagram in Markdown. Now type something in textarea below that code block. You will see that markdown contents on each change will shortly blink with the code of the Mermaid diagram before being replaced with the diagram itself.

**Comment by falkoschindler** [comment]
I think I found a rather nice solution: Instead of calling the "update" method with an HTML payload, we can simply update the element and handle the "updated" lifecycle hook on the client to render Mermaid diagrams.

I tested with the following snippet
```py
with ui.textarea() as textarea:
    with ui.tooltip():
        ui.markdown(extras=['fenced-code-blocks', 'mermaid']).bind_content_from(textarea, 'value')

ui.markdown(extras=['fenced-code-blocks', 'mermaid']).bind_content_from(textarea, 'value')
```

as well as with the code from https://github.com/zauberzeug/nicegui/issues/278#issuecomment-1415607101. All seems to work well. @me21 Would you like to try it out?

me21:
@falkoschindler works nicely!

falkoschindler (approved):
Great! Thank you for your contribution, @me21! I think together we found a really nice solution to this rendering problem. 🙂

PR #2809, opened 2024-04-03
Make libsass optional

rodja:
Because libsass is hard to compile on some platforms this pull request makes it a optional dependency.

### DIFF
--- a/examples/chat_app/main.py
+++ b/examples/chat_app/main.py
@@ -26,7 +26,7 @@ async def main(client: Client):
     user_id = str(uuid4())
     avatar = f'https://robohash.org/{user_id}?bgset=bg2'
 
-    ui.add_style(r'a:link, a:visited {color: inherit !important; text-decoration: none; font-weight: 500}')
+    ui.add_css(r'a:link, a:visited {color: inherit !important; text-decoration: none; font-weight: 500}')
     with ui.footer().classes('bg-white'), ui.column().classes('w-full max-w-3xl mx-auto my-6'):
         with ui.row().classes('w-full no-wrap items-center'):
             with ui.avatar().on('click', lambda: ui.navigate.to(main)):
--- a/examples/chat_with_ai/main.py
+++ b/examples/chat_with_ai/main.py
@@ -39,7 +39,7 @@ def main():
         thinking = False
         chat_messages.refresh()
 
-    ui.add_style(r'a:link, a:visited {color: inherit !important; text-decoration: none; font-weight: 500}')
+    ui.add_css(r'a:link, a:visited {color: inherit !important; text-decoration: none; font-weight: 500}')
 
     # the queries below are used to expand the contend down to the footer (content can then use flex-grow to expand)
     ui.query('.q-page').classes('flex')
--- a/nicegui/functions/style.py
+++ b/nicegui/functions/style.py
@@ -14,22 +14,55 @@ from ..logging import log
 from .html import add_head_html
 
 
-def add_style(content: Union[str, Path], indented: bool = False) -> None:
-    """Add style definitions to the page.
+def add_style(content: Union[str, Path], indented: bool = False) -> None:  # DEPRECATED
+    """Add style definitions to the page. [DEPRECATED]
 
-    This function can be used to add CSS, SCSS, or SASS style definitions to the head of the HTML page.
-
-    :param content: style content (string or file path)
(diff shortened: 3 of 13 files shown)


falkoschindler (changes requested):
First of all, I'm not 100% sure "libsass is hard to compile" is actually true. Do we have any evidence? I couldn't find any _recent_ bug report about it.

But if we want to make libsass optional, I don't think we can reliably recognize SCSS. [Here](https://sass-lang.com/guide/) is an overview of its features:

- **Variables** could be recognized by the dollar sign `$`. But this could also be part of a URL or a content property in CSS.

- **Nesting** could be recognized by consecutive opening brackets `{`. But they also appear when defining media queries in plain CSS.

- **Modules, mixins and inheritence** could be recognized by `@use`, `@mixin`, `@include` and `@extend`. That's fine.

- But **operators** are really hard to recognize.

If we can't recognize SCSS, we can't warn users that are, e.g., developing an app locally and then deploy it to a server, where libsass is missing. The app will silently serve a potentially broken style or layout.

Therefore I propose the following:

1. Make `ui.add_style` obsolete and show a warning when using it.
2. Provide a new `ui.add_css` function for plain CSS only.
3. Provide a new `ui.add_scss` function for SCSS and SASS. Like the current `ui.add_style` the user can switch between the different syntaxes via a `indented` parameter. This function requires libsass and will raise an ImportError if the optional library is missing.
4. For convenience we can also add a `ui.add_sass` function which calls `ui.add_scss` with `indented=True` internally.

I can try implementing and squeezing it into the 1.4.20 release tomorrow.

rodja:
@falkoschindler I like your solution. But I checked again: as soon as we remove functionality it is considered a breaking change in semantic versioning. It does not matter that we mark it deprecated or show a warning. The only non-2.0 way out of this may be the use of another (pure-Python) sass compiler like [django-sass](https://github.com/coderedcorp/django-sass) or [pyscss](https://pyscss.readthedocs.io/en/latest/).

BTW, my statement "libsass is hard to compile" was not precise. I meant "libsass must be compiled" -- which is really inconvenient. All other packages we depend upon are either pure Python or have precompiled libs (as far as I know). But libsass does need gcc. Reproduction:

```dockerfile
# Dockerfile
FROM python:3.8-slim

RUN pip install libsass
```

Buidling with `docker build .` results in 

```
2.338       gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DPy_LIMITED_API -I./libsass/include -I/usr/local/include/python3.8 -c _sass.c -o build/temp.linux-aarch64-3.8/_sass.o -fPIC -std=gnu++0x -Wall -Wno-parentheses -Werror=switch -DLIBSASS_VERSION="3.6.6"
2.338       unable to execute 'gcc': No such file or directory
2.338       error: command 'gcc' failed with exit status 1
2.338       [end of output]
2.338
2.338   note: This error originates from a subprocess, and is likely not a problem with pip.
2.340 error: legacy-install-failure
2.340
2.340 × Encountered error while trying to install package.
2.340 ╰─> libsass
```

And installing `gcc` is not enough, you need the `build-essential` meta-package. And even then the compilation is quite demanding. On my M3 Apple Silicon it's more than 30 seconds. I guess it will be quite a pain on a Raspberry-Pi (as reported in #2796).

falkoschindler:
Ok, so libsass is indeed a problem. Yesterday I also looked for alternatives, but without great success. [sass-lang](https://sass-lang.com/libsass/) mentions the following:

> Three other Python projects, [python-scss](https://github.com/pistolero/python-scss), [pylibsass](https://github.com/rsenk330/pylibsass) and [SassPython](https://github.com/marianoguerra/SassPython), haven’t been updated in a while.

You mentioned pyScss, which is stale for 2 years with 111 open issues, and django-sass, which is stale for 3 years and has only 71 stars.

None of them seems like a reasonable alternative.

I see your point regarding semantic versioning and version 2.0. But:

- Technically we're not following semantic versioning _yet_.
- I feel like keeping libsass until 2.0 causes more harm then deprecating `add_style` just 11 days after it has been introduced - before it gets widely adapted.
- One could argue that this change is more like a bug fix for the 1.4.19 release. Adding `add_style` that magically recognizes the language and depends on a heavy library like libsass was a mistake that should be fixed, better sooner than later.

rodja:
Ok. Convinced!

PR #2818, opened 2024-04-04
More robust On Air reconnect

rodja:
This PR improves the reconnection behavior to the On Air server.


falkoschindler:
@rodja Just to clarify: This PR is tagged with milestone 2.0.0, but targets the main branch. Is that correct?

rodja:
No, should be 1.4.20

falkoschindler (approved):

PR #2820, opened 2024-04-04
Provide app.storage.client as a location to store volatile data which only matters for the current connection

Alyxion:
As discussed in [pull request #2811](https://github.com/zauberzeug/nicegui/pull/2811) and the discussion [Add app.storage.tab or similar #1308](https://github.com/zauberzeug/nicegui/discussions/1308) I added a feature which allows the storage of data per Client connection.

This - in practice - and when either implementing an own single page application approach, e.g. based upon the example/single_page_app or when [#2811]((https://github.com/zauberzeug/nicegui/pull/2811)) is merged enables developers to create **per-browser tab user data instances**, for example with different logins and/or settings with the same comfort of as currently when using **app.storage.user** - so without handing over another per-page instance from event handler to handler.


nicegui/client.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, page: page, *, shared: bool = False) -> None:
         self._body_html = ''
 
         self.page = page
+        self.state = {}
```
rodja, line 76: I think this should be an OberservalDict to allow users to add callbacks for when the data changes. And it would be consistent with the other storage options. Or is that not necessary because we only store in memory and people can use bindings?

nicegui/client.py
```diff
@@ -84,6 +85,12 @@ def is_auto_index_client(self) -> bool:
         """Return True if this client is the auto-index client."""
         return self is self.auto_index_client
 
+    @staticmethod
+    def current_client() -> Optional[Client]:
+        """Returns the current client if obtainable from the current context."""
+        from .context import get_client
```
rodja, line 91: Should the import not be made at the top of the file?

nicegui/client.py
```diff
@@ -84,6 +85,12 @@ def is_auto_index_client(self) -> bool:
         """Return True if this client is the auto-index client."""
         return self is self.auto_index_client
 
+    @staticmethod
+    def current_client() -> Optional[Client]:
```
rodja, line 89: Do we really need this? Why not use 

```py
from nicegui import context
context.get_client()
```

in the user code base?

nicegui/storage.py
```diff
@@ -149,10 +150,21 @@ def general(self) -> Dict:
         """General storage shared between all users that is persisted on the server (where NiceGUI is executed)."""
         return self._general
 
+    @property
+    def session(self) -> Dict:
+        """Volatile client storage that is persisted on the server (where NiceGUI is
+        executed) on a per client/per connection basis.
+        Note that this kind of storage can only be used in single page applications
```
rodja, line 157: I wonder if we could implement something where we combine your approach here with the browser Session Storage to keep the storage as long as the browser tab is not closed or domain is switched (see #2378).

rodja (changes requested):

nicegui/client.py
Alyxion: Updated. As the automatic content update of bindable properties worked perfectly I implicated NiceGUI had another, active approach here.

nicegui/client.py
Alyxion: Changed. To minimize circular import issues I tend to use such "one-shot" usages into methods.

nicegui/client.py
Alyxion: After I worked myself through all examples of NiceGUI and all (by the way really well done) document pages I never had knowingly contact with the the library **context**. De facto is the only place where it is even used in the examples the ones where you do **context.get_client()**.

On the other hand the client class is already "well known" by the users as it is used all over the place, even an optional parameter of all pages.

Or with other words: For you who is working on this project since two years things like the context lib, slots etc. might seems the obvious place to look at. For newbies they ... in my opinion ... would find something defined in the Client easier, thus the "forwarding" of this call from "high" to "low" level.

nicegui/storage.py
Alyxion: It is possible by using the browser sessionStorage as you already discussed in 2378:

* When NiceGUI loads in the browser it checks if NICE_GUY_SESSION_GUID is stored in the sessionStorage, if not in generates one
* Upon connecting (and before the page is allowed to build) the GUID is sent from the browser to the client and would be searched in some kind of auto-garbage collected global dictionary, lets name it SessionDataBackup. If it were found it will be attached to Client.state
* When the WebSocket connection completely breaks the Client.state object is added to the SessionDataBackup.
* An async running garbage collector regularly checks the timeouts of the elements in the SessionDataBackup and removes outdates ones, e.g. after x minutes.

Requires of course that trying to access to the client.state variable only to be accessible when this hand-shake was done. This way you could transfer living/non-serializable objects from one page to another, even between full-reloads.

I think though that this should be an optional feature and should be part of another PR in the future. In streamlit if I press the reload button I get a fresh page / session... as I would assume it.

So do I when I click a link in my favorites (more by random than coincident, not thinking about that I am still in the same tab actually) and instead of a "fresh" app instance I already get an active one when keeping the session alive between real page changes. I would have to check how GDPR applies here but I think having this as default behavior would also force one to add a cookie info as well.

The only reason to implement this were if one would by force try to cling to the actual, full page reload approach between page changes. And having recently attached the SPA to the existing examples, e.g. examples/modules its like day and night when switching pages (like quite visible 0.2 seconds thus visible JavaScript load times) to like zero.

nicegui/client.py
rodja: Hmm... ok. @falkoschindler do you agree?

nicegui/storage.py
rodja: > I think though that this should be an optional feature 

I agree.

> and should be part of another PR in the future. 

Maybe. But we need to thing about how we would do it here, because it may impact the API (see below).

> In streamlit if I press the reload button I get a fresh page / session... as I would assume it.

That heavily depends on the use case. For example if you press refresh on a GitHub page, you do not end up with a new session (even if you are not logged in). Maybe we need `app.storage.tab` for a data storage which uses a key from the browsers Session Storage as you suggested and `app.storage.page` or `app.storage.client` which is dropped as soon as the Client is disconnected (eg. page reload or similar).

Alyxion: Yes, really a question of the use case and I think we should let this control remain in the hands of the NiceGUI dev.

You are right, when I refresh here my page on GitHub even the text here in this text box I'm typing right now in is **preserved**. If I open another tab **it is not**. So very likely using the sessionStorage mechanism, a potential future **app.storage.tab**. And this is good of course in this case b/c imagine having typed a lot of text and by accident tapping back button the mouse, e.g. default side button on the Logitech ones, fat fingered onto reload etc., you leave the page and all text is gone.

On the other hand there are of course all the scenarios I already mentioned where this might not be desired, can't go in detail here on a public discussion.

My take away for today: I will rename **app.storage.session** to **app.storage.client** so its transparent to the user it is bound to the client (and thus *volatile*) and then its up to him/her where he/she makes use of it.

rodja (changes requested):
Could you replace the double quotes with single quotes? See https://github.com/zauberzeug/nicegui/blob/main/CONTRIBUTING.md#single-vs-double-quotes

nicegui/storage.py
```diff
@@ -149,10 +151,26 @@ def general(self) -> Dict:
...
+        """
+        client = context.get_client()
+        return client.state
+
     def clear(self) -> None:
         """Clears all storage."""
         self._general.clear()
         self._users.clear()
+        if get_slot_stack():
+            self.client.clear()
```
rodja, line 173: Is `client.clear()` correct? I can not find it.

Alyxion: Yeap, it's correct. Property self.client returns Client.state which is an ObservableDict which declares .clear in observables.py:73

rodja: Ah indeed!

rodja:
I just started a first proof-of-conecept for `app.storage.tab` in https://github.com/zauberzeug/nicegui/pull/2837.

nicegui/client.py
rodja: Maybe it is better to add documentation on how to use the `context` object? For example instead of injecting the Client into a request in 

https://github.com/zauberzeug/nicegui/blob/bcdae282783ecabfdad11487d13e78cae223e40c/website/documentation/content/page_documentation.py#L42-L55

It is more flexible to use `context.get_client()` because this can globally be called from sub-functions.

nicegui/storage.py
```diff
@@ -149,10 +151,26 @@ def general(self) -> Dict:
...
+        """Client storage that is persisted on the server (where NiceGUI is executed) on a per client
+        connection basis.
+
+        The data is lost when the client disconnects through reloading the page, closing the tab or
+        navigating away from the page. It can be used to store data that is only relevant for the current view such
+        as filter settings on a dashboard or in-page navigation. As the data is not persisted it also allows the
+        storage of data structures such as database connections, pandas tables, numpy arrays, user specific ML models
+        or other living objects that are not serializable to JSON.
+        """
+        client = context.get_client()
```
rodja, line 165: I think we should add a safe guard to ensure the storage.client is not used with the [auto-index page](https://nicegui.io/documentation/section_pages_routing#auto-index_page). Probably the same as 

https://github.com/zauberzeug/nicegui/blob/7a115b0f34df35703d030da55cb85c417ecc5d4b/nicegui/storage.py#L161-L163

nicegui/client.py
Alyxion: That client parameter injection part in the current source code definitely was one of those ***ugh*** moments when I read through the source base even though its of course just picking up the approach of passing **request** into a fastapi endpoint.

I am personally more a fan of the Flask alike, context based approach where you just do from flask import request and be done.

Removing it will though likely brick many NiceGUI apps I can imagine, would need a healthy deprecation period....

nicegui/storage.py
Alyxion: Done, actually the get_slot triggered an exception as well but this way it more obvious to the user why.

nicegui/client.py
Alyxion: To minimize the changes in this PR and as this is obviously still something to sleep about I removed this change for now.

Alyxion:
Added the missing checks + unit tests, removed Client.current_client and added the sections to the documentation page based upon your examples in the .tab variant.

tests/test_client_state.py
```diff
@@ -1,23 +1,51 @@
...
-    app.storage.client['counter'] = 123
-
     def increment():
         app.storage.client['counter'] = app.storage.client['counter'] + 1
 
-    ui.button('Increment').on_click(increment)
-    ui.label().bind_text(app.storage.client, 'counter')
+    @ui.page('/')
+    async def page():
+        await context.get_client().connected()
```
rodja, line 13: Why is waiting for client connection needed here?

nicegui/storage.py
```diff
@@ -149,10 +151,27 @@ def general(self) -> Dict:
...
+    def client(self) -> ObservableDict:
+        """A volatile storage that is only kept during the current connection to the client.
+        
+        Like `app.storage.tab` data is unique per browser tab but is even more volatile as it is already discarded
+        when the connection to the client is lost through a page reload or a navigation."""
+        if self._is_in_auto_index_context():
+            raise RuntimeError('app.storage.client can only be used with page builder functions '
+                               '(https://nicegui.io/documentation/page)')
+        client = context.get_client()
+        if not client.has_socket_connection:
```
rodja, line 164: I don't think a socket connection is required for app.storage.client...

website/documentation/content/storage_documentation.py
```diff
@@ -86,3 +92,47 @@ def ui_state():
...
+          if the user closes the browser, reloads the page or navigates to another page.
+          This is may be beneficial for resource hungry or intentionally very short lived data such as a database 
+          connection which should be closed as soon as the user leaves the page, sensitive data or if you on purpose
+          want to return a page with the default settings every time the user reloads the page while keeping the
+          data alive during in-page navigation or when updating elements on the site in intervals such as a live feed.
+          ''')
+def tab_storage():
+    from nicegui import app, context
+
+    class DbConnection:  # dummy class to simulate a database connection
```
rodja, line 111: The demo is to complicated in my opinion. A simple counter like for `app.storage.tab` would be better to demonstrate the mechanism.

tests/test_client_state.py
Alyxion: Fixed. You are right, in difference to .tab - even though the  data is of course bound to the Client it of course does not need to be established yet.

website/documentation/content/storage_documentation.py
Alyxion: I wanted to give the user a better impression about when to use what. Simplified the demo.

Alyxion:
Updated. I already referred [#2837](https://github.com/zauberzeug/nicegui/pull/2837) in the documentation to explain when to use **tab** and when to use **client**. The count in *It features four built-in storage types:* then needs to be adjusted accordingly when boths PRs should be united in the main branch.

Alyxion:
Resolved the conflicts with the new main branch and adjusted the doc accordingly.

rodja (approved):
I've just made minimal changes to the documentation. In markdown we like start each sentence in a new line.
@falkoschindler it would be great if you can have a look, too.

falkoschindler (commented):
I just reviewed the code and made a few minor changes. Most prominently I moved the tests into test_storage.py. Even though this file is already pretty long, we tend to move tests for file x.py into a file test_x.py. So all storage tests are in one place.

I've got three questions remaining:

1. Is `client.state` a good name for this kind of storage? What do you think about `client.storage`? I'm undecided though. Since it is very volatile, it's somehow less than real storage. But it is implemented and documented as "storage".
2. Why do we need to call `get_slot_stack()` in `clear()`? I guess it's for some important but slightly unintuitive reason.
3. Why are there two buttons and two labels in `test_client_storage`?

Alyxion:
Thanks @falkoschindler, regarding your questions:

> 1. Is `client.state` a good name for this kind of storage? What do you think about `client.storage`? I'm undecided though. Since it is very volatile, it's somehow less than real storage. But it is implemented and documented as "storage".

The reason why I chose **state** was actually 1:1 taken from Streamlit st.session_state which has exactly the same behaviour - "keep the state" as long as the user is connected. I think it's a bit philosophical what one interpretes into the words "state" and "storage", after all its also called Solid State Drive and not Solid Storage Drive, right? ;-)

> 2. Why do we need to call `get_slot_stack()` in `clear()`? I guess it's for some important but slightly unintuitive reason.

Otherwise app.reset crashes during the tests because it is also called w/o an available slot and then context.get_client would crash.

> 3. Why are there two buttons and two labels in `test_client_storage`?

Fixed, I assume through a git merge which duplicated those lines.

falkoschindler:
> > 2. Why do we need to call get_slot_stack() in clear()? I guess it's for some important but slightly unintuitive reason.
>
> Otherwise app.reset crashes during the tests because it is also called w/o an available slot and then context.get_client would crash.

Ok, I've rewritten this part a little:
- I explicitly get the `client` and catch any runtime errors.
- I don't check for auto-index clients, because we wan't to clear _all_ storage.
- We don't need getting the slot stack anymore, which was a bit confusing in my opinion.

rodja (approved):

falkoschindler:
Code for testing all our storages:

```py
@ui.page('/')
async def index():
    if 'time' not in app.storage.browser:
        app.storage.browser['time'] = time.time()
    ui.label(f'{app.storage.browser["time"]=}')

    if 'time' not in app.storage.client:
        app.storage.client['time'] = time.time()
    ui.label(f'{app.storage.client["time"]=}')

    if 'time' not in app.storage.user:
        app.storage.user['time'] = time.time()
    ui.label(f'{app.storage.user["time"]=}')

    if 'time' not in app.storage.general:
        app.storage.general['time'] = time.time()
    ui.label(f'{app.storage.general["time"]=}')

    await context.get_client().connected()

    if 'time' not in app.storage.tab:
        app.storage.tab['time'] = time.time()
    ui.label(f'{app.storage.tab["time"]=}')
```

Issue #2482, opened 2024-01-31
Wrapping the menu_and_tabs example into a @ui.page("/") causes continuous 100% CPU load


pplno:
### Description

When I take the [menu_and_tabs ](https://github.com/zauberzeug/nicegui/blob/main/examples/menu_and_tabs/main.py) example and wrap it into a `@ui.page("/")` like below, the process then runs with continous 100% CPU load when Tab A or Tab C are active. The CPU load drops to 0% when Tab B is active. The issue doesn't persist for the original menu_and_tabs example.

I reproduced the issue in a fresh .venv with nicegui==1.4.13 under Python 3.10 and 3.12

```py
from nicegui import ui
@ui.page("/")
def index_page() -> None:

    with ui.header().classes(replace='row items-center') as header:
        ui.button(on_click=lambda: left_drawer.toggle(), icon='menu').props('flat color=white')
        with ui.tabs() as tabs:
            ui.tab('A')
            ui.tab('B')
            ui.tab('C')

    with ui.footer(value=False) as footer:
        ui.label('Footer')

    with ui.left_drawer().classes('bg-blue-100') as left_drawer:
        ui.label('Side menu')

    with ui.page_sticky(position='bottom-right', x_offset=20, y_offset=20):
        ui.button(on_click=footer.toggle, icon='contact_support').props('fab')

    with ui.tab_panels(tabs, value='A').classes('w-full'):
        with ui.tab_panel('A'):
            ui.label('Content of A')
        with ui.tab_panel('B'):
            ui.label('Content of B')
        with ui.tab_panel('C'):
            ui.label('Content of C')

ui.run()
```

edit: in the meantime I have reproduced the issue on a second machine.
The smallest code to reproduce the problem appears to be the ui.tabs themselves 
```py
from nicegui import ui

@ui.page("/")
def index_page() -> None:
    
    with ui.tabs():
        ui.tab('A')
        ui.tab('B')
        ui.tab('C')
        ui.tab('D')

ui.run()
```

falkoschindler:
Wow @pplno, that's very strange.
I couldn't reproduce it immediately though.
What OS and browser are you using? Does the CPU remain at 100% when closing the tab, even after waiting?

pplno:
I'm on Windows 11 with Chrome 121.0.6167.140 and Edge 121.0.2277.83. Both browser show the same behaviour.
The CPU goes down to 0% when you close the tab, but not when unfocusing it.

[...]

PR #2827, opened 2024-04-05
Replace check_interval for JavaScript requests with asyncio event

falkoschindler:
This PR solves one aspect of issue #2482 by replacing the 0.01s loop with asyncio events.

To verify everything is still working, I experimented with the following demo code:
```py
def alert():
    ui.run_javascript('alert("Hello!")')

async def get_date():
    time = await ui.run_javascript('Date()')
    ui.notify(f'Browser time: {time}')

async def fail():
    await ui.run_javascript('fail()')

ui.button('fire and forget', on_click=alert)
ui.button('receive result', on_click=get_date)
ui.button('timeout', on_click=fail)
```

Note that I didn't mark every single method parameter as DEPRECATED. I only marked the root implementation, which is called by all derived methods. Similarly, I only added a warning once. I even removed a duplicate marker and a warning for the deprecated `respond` parameter.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -18,6 +18,7 @@ from .awaitable_response import AwaitableResponse
 from .dependencies import generate_resources
 from .element import Element
 from .favicon import get_favicon_url
+from .javascript_request import JavaScriptRequest
 from .logging import log
 from .outbox import Outbox
 from .version import __version__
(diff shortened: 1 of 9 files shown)


rodja (approved):

rodja:
I love it!

PR #2837, opened 2024-04-06
Provide app.storage.tab

rodja:
This pull request picks up on the idea from https://github.com/zauberzeug/nicegui/pull/2820#discussion_r1554523968 and other discussion around this topic to implement `app.storage.tab`: a volatile storage which keeps a state dictionary as long as the browser tab lives. That means the data survives page reloads and switching urls.

It is implemented by storing the dictionary in the memory of the server. To decide which dictionary needs to be accessed the pull request introduces a `client.tab_id` which is automatically set when doing the websocket handshake.

### ToDos

- [x] When to delete the storage dicts? A tab could revisit the page even after a few days...
- [x] add tests
- [x] add demo

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -58,6 +58,7 @@ class Client:
         self.shared = shared
         self.on_air = False
         self._disconnect_task: Optional[asyncio.Task] = None
+        self.tab_id: Optional[str] = None
 
         self.outbox = Outbox(self)
 
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -158,7 +158,7 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
 
 
 @sio.on('handshake')
-async def _on_handshake(sid: str, data: dict[str, str]) -> bool:
+async def _on_handshake(sid: str, data: Dict[str, str]) -> bool:
     client = Client.instances.get(data['client_id'])
     if not client:
         return False
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -129,6 +129,8 @@ async def _startup() -> None:
     background_tasks.create(binding.refresh_loop(), name='refresh bindings')
     background_tasks.create(Client.prune_instances(), name='prune clients')
     background_tasks.create(Slot.prune_stacks(), name='prune slot stacks')
+    prune_tab_storages = core.app.storage._prune_tab_storage  # pylint: disable=protected-access
+    background_tasks.create(prune_tab_storages(), name='prune tab storages')
     air.connect()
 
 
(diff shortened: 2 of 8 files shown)


Alyxion:
> * [ ]  When to delete the storage dicts? A tab could revisit the page even after a few days...

I am one who always closes all browser tabs asap but I know many people who have like 50+ tabs over weeks who even have to scroll horizontally and just still identify the tab by the fav icon and first letter.

I think we need to differentiate here between important, un-recoverable, compact and most importantly serializable data... like a long text someone typed and it would really hurt if it were lost... or the user's avatar we just cached in storage.tab so we don't have to pull it from the database every time.

If one could flag such data as "important" then we could transfer it in a configurable intervals from the server to the client and store it next to "__nicegui_tab_id" in the local storage.

And when the connection was completely broken (PC shutdown, browser closed etc.) and is then re-established we could just send all this data along with the handshake back to the server and then the server can decide if it can be recovered.
Could become problematic of course between version changes of the server and such situations but  thats the same for cookies already.

rodja:
> I think we need to differentiate here between important, un-recoverable, compact and most importantly serializable data

I would rather like to not do that. Maybe we can leave that it in the hands of the user? I suggest we do:

1. keep the data only in memory (eg. accept full loss if server restarts) -- that way we can store any Python objects just like `app.storage.client` from #2820.
2. remove data which is not accessed for over a month (this is also the default cookie lifetime for `app.storage.browser`)
3. make the data lifetime configurable via parameter in `ui.run`

Maybe later we can add a feature that saves all serializable fields to a file in regular intervals -- and load this on server startup.

rodja:
I created #2841 as a feature request for persistence across server restarts.

Alyxion:
Looks great. Two remarks though:

* If one would store his data like so app.storage.tab['User'] = {'user_id':..., 'last_message'.'''} and then change the values within subdict an interaction will never reduce the timeout. If someone is not aware of this then the session will timeout even though he interacts with the page. May be an "active tab" should update the timeout automatically instead.
* The time delta of 30 days is mentioned nowhere in the documentation. (also seems a bit much as default buts thats another story)

Other than that the Exceptions triggered by auto-index and missing client connection in the unit tests are not yet covered, don't know if you guys aim here for high coverage. See [here](https://github.com/zauberzeug/nicegui/commit/08d73bdabd59b260100fb378619204318ea9fa32#diff-caecfc28a18979b431b701c10be4f6307a7795c605c22b13cb83b9a85dd63601).

rodja:
> If one would store his data like so app.storage.tab['User'] = {'user_id':..., 'last_message'.'''} and then change the values within subdict an interaction will never reduce the timeout. 

That should work. `ObservableDict` supports hierarchical observation 😎

falkoschindler (approved):

falkoschindler:
@rodja There seems to be an issue with this PR: https://github.com/zauberzeug/nicegui/issues/2866

PR #2861, opened 2024-04-09
audio recording example

CVxTz:
Hello ! 

I added this example that allows to record audio from the browser then play it back or download it as a file.
I used a custom Vue component and pass the audio data to the backend once the recording is done.

Let me know what you think !

---

Related feature request: #1610


falkoschindler (approved):
Thanks a lot for this great pull request, @CVxTz! It's a really nice example.
I just refactored it a little to simplify the usage even more, so there's no need for the extra `AudioData` class anymore.

CVxTz:
Thanks !

PR #2867, opened 2024-04-10
Replace repeated polling in Outbox.loop() with an asyncio event (see #2482)

afullerx:
This change replaces the continuous polling implementation of the Outbox.loop() method with one based on asyncio events. It, along with #2827, resolves issue #2482, while lowering CPU usage at idle by over 80%. (To just 0.02% on my system). 

I was unable to use the most straightforward approach due to the fact that in versions of Python prior to 3.10, asyncio synchronization primitives attach themselves to the current event loop during initialization. In this case, the event loop that Outbox is created on is different from the one executing its loop method. Therefore, I had to go with a slightly more complex lazy initialization approach in order to maintain compatibility with Python 3.8. 

At first, I removed the asyncio.sleep(), reasoning that lower latency in the delivery of messages and updates could result in better performance. However, during automated testing, this led to an increase in the number of random timing related errors. As a result, I concluded that some batching was actually beneficial for system performance. When adding it back, 5ms was chosen as the batch duration because this is the average delay in processing with the current implementation and didn't result in any increase in sporadic errors. This value could likely be optimized further.

I was getting random timeouts from client.connected() during test teardowns with the default timeout, so chose to effectively disable it with timeout=2**30, as it shouldn't be needed here. I thought this was better than wrapping it in a while/try.

The automated tests were run using Python 3.8 in the Docker dev container. Ad hoc testing was done natively on Windows 10 using Python 3.12.

Note: This is my first pull request ever. While I've been coding for 20+ years in several different languages, it has always been as a solo hobbyist. As a result, I've had to spend the last few days speed-climbing the Git/Github/PR learning curves while putting this together, so it's quite possible I've botched some part of the process. Don't hesitate to give me any suggestions on how to do anything correctly or more efficiently in the future. It would be greatly appreciated.


nicegui/outbox.py
```diff
@@ -23,35 +23,46 @@ def __init__(self, client: Client) -> None:
...
+
         while not self._should_stop:
             try:
-                await asyncio.sleep(0.01)
-
-                if not self.updates and not self.messages:
-                    continue
-
+                await self._enqueue_event.wait()
+                await asyncio.sleep(0.005)
```
falkoschindler, line 60: Why do we need to sleep after the event has been awaited?

nicegui/outbox.py
```diff
@@ -23,35 +23,46 @@ def __init__(self, client: Client) -> None:
...
+                
                 if not self.client.has_socket_connection:
-                    continue
-
+                    await self.client.connected(timeout=2**30)
```
falkoschindler, line 63: I'm not sure if waiting basically indefinitely is a good idea. The client _might_ never connect, e.g. if the internet connection is interrupted before the websocket connection is created. I think we need to set the timeout to something like 60 seconds (like we do in `timer.py`) and `return` in case of a `TimeoutError`. This way we avoid a memory leak with broken client connections.

nicegui/outbox.py
```diff
@@ -23,35 +23,46 @@ def __init__(self, client: Client) -> None:
         self.updates: Dict[ElementId, Optional[Element]] = {}
         self.messages: Deque[Message] = deque()
         self._should_stop = False
+        self._enqueue_event = None
```
falkoschindler, line 26: Why did you chose lazy initialization? Can't we simply create the event here and simply set it later in the enqueue methods? This would make `_set_enqueue_event` kind of obsolete.

falkoschindler (commented):
Thanks a lot for the pull request, @afullerx!
I like this approach very much. There are just a few things I don't quite understand or would implement differently. Can you check my comments, please? 🙂

nicegui/outbox.py
afullerx: With the current implementation, there is some batching of queue processing due to the sleep interval. I wasn't sure if this was an intentional part of the design, so I first left it out in order to minimize latency in the system. During automated testing, this led to an increase in the number of random errors. I surmised that this increase in errors was due to the user interface being less responsive without the batching behavior, leading to more missed input events and therefore random errors. When adding it back, I chose 5ms as the batch duration to approximate the average delay in processing with the current implementation. This prevented any increase in errors. If the sleep is to be left out, some adjustment of screen.wait() calls in the tests may be necessary.

nicegui/outbox.py
afullerx: I reasoned that if there's never a connection, we never have any work to do and that there's nothing we can do about that locally. The memory leak aspect didn't occur to me. With that in mind, I completely agree that it should be changed to 60 seconds.

nicegui/outbox.py
afullerx: That was exactly what I did at first and it worked flawlessly with Python 3.12. Testing with 3.8, however, caused the following exception when the event was accessed: `RuntimeError: Task got Future <Future pending> attached to a different loop`. This is because prior to Python 3.10, events attach themselves to the current event loop during initialization. In this case, it appears that the event loop Outbox is created on is different from the one executing its loop method. I switched to the lazy initialization approach to maintain compatibility with Python 3.8.

falkoschindler: Oh sorry, I just noticed your explanation in your original post. Sorry! 🫣

nicegui/outbox.py
afullerx: Changing that now.

afullerx: Actually, this will take me a bit. I want to be sure to get it right.

afullerx:
No worries, I thought maybe my explanations weren't clear enough, so I didn't mind expounding.

nicegui/outbox.py
afullerx: I had some more thoughts on this. I chose to include the batching because of the issues I was seeing in testing and out of an abundance of caution. Despite this, I'm not completely convinced of its necessity. 

Firstly, my test setup is relatively low-performance. It involves multiple layers of virtualization on a six-year-old CPU. This may have made the tests more sensitive to small changes in timing. It's quite possible you'd see no issues whatsoever on your typical test systems. It also occurred to me that the increased test failures might not be an indication of overall lower performance but instead the result of the test timings being tuned to the current implementation.

Secondly, in my ad hoc testing of my own GUIs, I didn't notice any difference in performance with or without the batching. This leads me to believe that whatever small difference there is, it might be negligible from the user standpoint.

I'm on the fence about leaving it in since, now that you're aware of the potential issue, it can trivially be added later if problems arise in your testing. Let me know what you think.

nicegui/outbox.py
afullerx: I made the change to the connection timeout handling. I modeled it after the example you mentioned in timer.py.

afullerx: I'm having some second thoughts about this. Might it not be better to wait so we can recover if the connection is re-established. Otherwise, the UI would die and need to be reloaded because of an interruption lasting more than 60 seconds. I only ask because this is the current implementation, and we would be changing it.

afullerx: I was just looking through the code more and I realized a few things. To be consistent with the current behavior. We would need to occasionally time out `client.connected()` and make sure that `self._should_stop` has not been set. I also realized that the `self._enqueue_event.wait()` should time out periodically and do the same, so we don't get stuck waiting forever. It get set when the `client` calls `stop()`in its `delete()` method, so I think this would prevent any new memory leaks.

nicegui/outbox.py
me21: Is it safe to `asyncio.sleep(0.005)` here given that we're removing `asyncio.sleep(0.01)` or replacing it with `asyncio.sleep(0.016)` to avoid issues with sleep interval being less than Windows timer resolution?

falkoschindler (commented):
I just reviewed and improved the code a bit:

- With `await asyncio.wait_for(self._enqueue_event.wait(), timeout=1.0)` we can await an asyncio event with a timeout. If the timeout occurs, we can simply `continue`. The while loop will simply continue iterating if `_should_stop` is not set.
- If there is no client connection, we continue after a short delay. This way we check for a client connection in 0.1-second intervals.
- The 0.1 second delay is a bit more defensive. I guess this is why I had to adjust two pytests.

Regarding the batching of multiple messages: If multiple updates happened within 0.01 seconds, the old implementation waited and submitted multiple updates at once. The new implementation will submit a single update, but only once the currently running task yields and the outbox loop is called for another cycle. So if you synchronously update multiple UI elements, their updates will be batched.

Anyway, @rodja and I will thouroughly review this pull request once again to make sure we don't break anything.

nicegui/outbox.py
falkoschindler: I think it was safe to remove this `sleep` in https://github.com/zauberzeug/nicegui/pull/2867/commits/a9d0d52bc01723bee0ad7d53d00061ab3d2114ca. But let me know if it works on your machine.

nicegui/outbox.py
falkoschindler: I think with `await asyncio.wait_for(self._enqueue_event.wait(), timeout=1.0)` we basically reproduce the old behavior, just with a larger interval for checking `_should_stop`. Do you agree?

nicegui/outbox.py
afullerx: Are you seeing increased test failures? I see you changed some tests.

afullerx:
Nice, just one note about `asyncio.wait_for()`:
> Changed in version 3.11: Raises [TimeoutError](https://docs.python.org/3/library/exceptions.html#TimeoutError) instead of >[asyncio.TimeoutError](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.TimeoutError).

So you'd need to catch both for Python 3.8 compatibility.

nicegui/outbox.py
afullerx: @me21 
> Is it safe to `asyncio.sleep(0.005)` here given that we're removing `asyncio.sleep(0.01)` or replacing it with `asyncio.sleep(0.016)` to avoid issues with sleep interval being less than Windows timer resolution?

It would be safe here because the `asyncio.Event` is what controls execution of the loop, so there's no chance of it running out of control, even with 0 sleep. It was only included as a possible mitigation against an increase in test failures.

nicegui/outbox.py
afullerx: Absolutely, I was about to do that myself.

afullerx:
@falkoschindler I like the improvements. I saw just one issue: Prior to Python 3.11 `asyncio.wait_for()` raised `asyncio.TimeoutError`, so we also need to catch that. I verified that it's necessary with testing, so I went ahead and added it.

Good point about the natural batching that occurs with synchronous updates. I should have thought of that. It makes me even more curious why I saw an increase in test failures and why adding a sleep fixed it. Hopefully just an anomaly with my low-performance test setup and small changes in timing.

afullerx:
@falkoschindler Since the addition of `asyncio.wait_for()`, I'm getting lockups about every 3rd test run in the dev container. I'm well on my way to picking apart what's happening, but it's taking a while because every experiment can take over an hour to complete.

afullerx:
Fixed in the latest commit. The fix is pretty simple, but figuring out what was going on definitely wasn't. I thought it was going to be an event-related deadlock. It turns out it was some kind of obscure issue preventing task termination when `asyncio.wait_for()` is called in a loop with the event already set.

The hangs occurred when `self.client.has_socket_connection` is set to `false` during the test exit while there are still updates in the queue. Since `self._should_stop` often isn't set during disorderly test teardowns, this led to us calling `wait_for()` in an infinite loop. This normally wouldn't be a problem because the `Outbox.loop()` task is abruptly terminated, regardless of what it's doing. For some unknown reason, the repeated calls to `wait_for()` prevented this typical behavior, resulting in a hanging of the process. The solution is simply to check `self._enqueue_event.is_set()` before calling `wait_for()`.

This change also provides a useful optimization, since every call to `asyncio.wait_for()` results in a new task being created and multiple context switches. I validated this fix by running the tests on a continuous loop for several hours without issue.

falkoschindler (approved):
Thank you so much for your thorough investigation and beautiful solution, @afullerx! It's a nice improvement to avoid creating tasks just to exit immediately because the event is already set. And if it resolves the hanging tests - the better! 😀

PR #2879, opened 2024-04-12
Introduce ui.context as a simpler method to get Client, Slot-Stack etc

rodja:
As I wrote https://github.com/zauberzeug/nicegui/pull/2811#issuecomment-2051112734, it hit me that we could provide `ui.context` to get the current client object or Slot-Stack. I also think it would later also be a good place to cluster `page_title`, `favicon`, `viewport`, ... which currently can only be configured via page builder decorator.


router.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
```
falkoschindler, line 1: This file seems to have been added unintentionally.

falkoschindler (commented):
I wonder an API like `ui.context.client` was possible. This would require `context` to be a class instance and I'm not sure if this is possible and practical. It would also hide the fact that `client` isn't a constant field, but the result of a function. So we might just leave it as `ui.context.get_client()`.

router.py
rodja: Oh yes. I'm sorry.

Alyxion:
> I wonder an API like `ui.context.client` was possible. This would require `context` to be a class instance and I'm not sure if this is possible and practical. It would also hide the fact that `client` isn't a constant field, but the result of a function. So we might just leave it as `ui.context.get_client()`.

In the end always the balance between lazyness/simplicity and readability. In Flask world thats basically exactly what `request` does when you import it from Flask. It will always contain the current context's / thread's request properties even though it looks like a global variable.

```
_cv_request: ContextVar[RequestContext] = ContextVar("flask.request_ctx")
request_ctx: RequestContext = LocalProxy(  # type: ignore[assignment]
    _cv_request, unbound_message=_no_req_msg
)
request: Request = LocalProxy(  # type: ignore[assignment]
    _cv_request, "request", unbound_message=_no_req_msg
)
session: SessionMixin = LocalProxy(  # type: ignore[assignment]
    _cv_request, "session", unbound_message=_no_req_msg
)
``` see https://github.com/pallets/flask/blob/main/src/flask/globals.py

PR #2890, opened 2024-04-14
Introduce orthographic camera for `ui.scene`

falkoschindler:
As requested in #2872, this PR introduces a parameter `camera_type` for `ui.scene` to switch between perspective and orthographic cameras.

### DIFF
--- a/nicegui/elements/scene.js
+++ b/nicegui/elements/scene.js
@@ -66,7 +66,12 @@ export default {
     window["scene_" + this.$el.id] = this.scene; // NOTE: for selenium tests only
 
     this.look_at = new THREE.Vector3(0, 0, 0);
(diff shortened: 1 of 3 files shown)


rodja (approved):

falkoschindler:
I just noticed that there's no way to adjust the field of view programmatically. For perspective cameras you can simply move the camera relative to the object. But this doesn't work for orthographic cameras, which require adjusting the parameters `left`, `right`, `top`, and `bottom`.

I guess we should make these camera parameters configurable (also `fov` for perspective cameras). Now that this PR is merged, we can do that in another PR.

falkoschindler:
I remove this feature from 1.4.22 in order to get the API right before releasing anything. 1.4.22 will probably be a quick bugfix release.

PR #2905, opened 2024-04-16
UI context properties

falkoschindler:
As an extension to PR #2879, this PR allows to access client, slot, and slot stack via properties like this:

```py
ui.context.client
ui.context.slot
ui.context.slot_stack
```

The existing getter functions have been preserved but deprecated.


rodja (approved):
😎

PR #2922, opened 2024-04-20
Add early exit when setting the value of `ui.state` variables

falkoschindler:
This PR solves a problem described in #2921 where setting the value of a `ui.state` variable causes an infinite recursion.

Take this simple example code:
```py
@ui.refreshable
def counter():
    count, set_count = ui.state(0)
    color, set_color = ui.state('black')
    ui.number(value=count, on_change=lambda e: set_count(e.value))
    ui.select(['black', 'red', 'green', 'blue'], value=color, on_change=lambda e: set_color(e.value))
    if count == 0:
        set_color('red')

counter()
```

Whenever the `count` is zero, the `color` is set to "red", which causes the refreshable to re-run - even if the color hasn't changed. The proposed change causes an early exit to avoid the infinite recursion.

### DIFF
--- a/nicegui/functions/refreshable.py
+++ b/nicegui/functions/refreshable.py
@@ -154,6 +154,8 @@ def state(value: Any) -> Tuple[Any, Callable[[Any], None]]:
         value = target.locals[target.next_index]
 
     def set_value(new_value: Any, index=target.next_index) -> None:
+        if target.locals[index] == new_value:
+            return
         target.locals[index] = new_value
         target.refreshable.refresh()
 


vickorian:
@falkoschindler 

I pushed your PR to my version and it solves the recursion but that only brought out a set_value problem, it does not play well with arrays, making some more complex things not possible. I'm planning on tackling that when i finish this project along with the documentation for advanced aesthetics. More complex state based systems aren't possible without using custom binary functions to specify exact states via precalculated dictionaries for each state, which can get, to say the least, cumbersome.

rodja (approved):

PR #2941, opened 2024-04-23
Example: Add zeromq example

keenanjohnson:
This change adds a basic zeromq implementation with nicegui.

In order to do this, the zmq python library is used to create a publisher and a subscriber. The main interesting aspect is that the zmq.async library is used to create a subscriber that can be used in an asyncio loop, which is necessary to run the NiceGUI server.

I hope this will be useful to the community as it took me a bit of time to figure out. Let me know what you think.


rodja (approved):
Looks cool! Thanks for the contribution @keenanjohnson!

falkoschindler (approved):
Thanks for the pull request, @keenanjohnson!
I just simplified the code a bit here and there. Let's merge. 🙂

keenanjohnson:
Excellent and thanks for the review! Glad I could contribute back a bit.

PR #2942, opened 2024-04-23
Add chip element

chrschorn:
This adds the [Quasar Chip element](https://quasar.dev/vue-components/chip). Quick demo from the documentation page:

https://github.com/zauberzeug/nicegui/assets/6242043/defaa6ce-30f9-4031-a04b-213977e653f3

I wanted to implement all three main features (clickable, selectable, removable) and my goal was to enable a label add/remove interface as shown in the lower part of the video. 

~~As a part of that, I also added Enter key support for `ui.input` which can be quite handy. Doing this via `ui.keyboard` seemed not easily possible as the callback in there does not know which element is focussed at the moment.~~ I realized the way to do this is `.on("keydown.enter")`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,7 @@
     "--disable=C0301", // Line too long (exceeds character limit)
     "--disable=C0302", // Too many lines in module
     "--disable=R0801", // Similar lines in files
+    "--disable=R0901", // Too many ancestors
     "--disable=R0902", // Too many instance attributes
     "--disable=R0903", // Too few public methods
     "--disable=R0904", // Too many public methods
--- a/nicegui/elements/chip.py
+++ b/nicegui/elements/chip.py
@@ -17,7 +17,6 @@ class Chip(ValueElement, TextElement, BackgroundColorElement, TextColorElement, 
                  text: str = '',
                  *,
                  icon: Optional[str] = None,
-                 clickable: bool = False,
                  selectable: bool = False,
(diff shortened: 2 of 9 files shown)


falkoschindler (approved):
Thanks for the pull request, @chrschorn! This turned out to be a rather powerful UI element. 🙂

I just reviewed and slightly updated your code:

- Like for `ui.item`, I removed the slightly redundand `clickable` parameter. The chip is automatically clickable when a click handler is provided.
- I removed the awaitable `clicked` method. Even though we added it to `ui.button` to support something like "UI flows", we don't think every clickable UI element needs to support it.
- You wrote "The update:selected event has no value, so we treat it as a toggle event". But as far as I can tell we can use `e.args` as the new selection state.
- For the pytests I replaced some hard assertions with `screen.should_contain` statements with implicit waits. This can reduce flakiness on less powerful build systems.
- I improved the contrast of some demo chips and simplified the second demo a bit. Even though I sacrificed the notifications with the complete chip list, I hope the reduced complexity helps to focus on the chip API.
- Since most of the remaining code in chip.py was basically about managing the selection, I decided to introduce a new mixin. This helps to focus on this single feature and to use other mixins as blueprint for the state synchronization. And this way it was easy to add binding support.

What do you think? Do you agree with the changes I made?

chrschorn:
Thanks for the review/polish/cleanup! They all make sense, I have nothing to add 👍

PR #2608, opened 2024-02-24
Multiple cameras

fabian0702:
This Pull Request introduces the ability to use multiple cameras in one scene, as discussed in [this discord Thread](https://discord.com/channels/1089836369431498784/1208029236535758860). This would allow to have multiple ui.scene elements which are showing the same scene, but from multiple perspectives at the same time.

PR #2947, opened 2024-04-24
Scene view component

fabian0702:
This PR contains the newly created ui.scene_view element as discussed in PR #2608. This would allow to have multiple cameras for the same scene, but from multiple perspectives at the same time.


fabian0702:
@falkoschindler I fixed the problems you mentioned and added the necessary camera parameters. Unfortunately I had to remove the text renderer from the scene_view component due to the way text is currently rendered.

falkoschindler (approved):
Great! I think this feature is ready to merge. Thanks again, @fabian0702!

Regarding the text renders: It looks like they can't handle rendering a scene into another div. So I removed them altogether and added a note about this limitation to the documentation of `ui.scene_view`.

And I think I found a simpler way to wait for the parent scene. Awaiting `this.$nextTick()` seems to work well in all automated and manual tests. What do you think?

fabian0702:
@falkoschindler I'm all for simplifying the code. The rest of the changes look good to me, thank you for the improvements.

Issue #2917, opened 2024-04-19
Plotly chart changes position after updating the chart.


drummerboy2543:
### Description

I would like to report a unexpected behavior/bug about plotly dynamically changes position, after being updated. 

Working off of the demo code, that is given via the [documentation](https://nicegui.io/documentation/plotly#plot_updates).  The only differences I made was  adding some rows for some inputs and labels before button (Going to be used later, but made a dummy to show a example). I also specified a different height, for the plotly graph

https://github.com/zauberzeug/nicegui/assets/15644157/69a09889-8c4e-471b-90eb-1929d2aa5990

As you can see in the video attached the formatting gets a bit messed up and does not adhere to the column and row structure that was outlined. 

The code is shown below.

```python
import plotly.graph_objects as go
from nicegui import ui
from random import random

fig = go.Figure()
fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
plot = ui.plotly(fig).classes('w-full h-full')




with ui.row():
    ui.input(label="Cord 1 (x,y)", placeholder="Enter a value").props("dense")
    ui.label("(45.5, 35.23)")
with ui.row():
    ui.input(label="Cord 2 (x,y)", placeholder="Enter a value").props("dense")
    ui.label("(45.5, 35.23)")
with ui.row():
    ui.input(label="Cord 3 (x,y)", placeholder="Enter a value").props("dense")
    ui.label("(45.5, 35.23)")
with ui.row():
    ui.input(label="Cord 4 (x,y)", placeholder="Enter a value").props("dense")
    ui.label("(45.5, 35.23)")
def add_trace():
    fig.add_trace(go.Scatter(x=[1, 2, 3], y=[random(), random(), random()]))
    plot.update()

ui.button('Add trace', on_click=add_trace)

ui.run() 
```
Notice at the beginning the formation that was given in code failed. Only after the update it would go back to what is expected behavior.  However constantly updating it changes back and forth. I wonder if this is some sort of race condition.

After talking about the issue on discord we were able to narrow down the issue of the demo code exactly, if height is not specified. 

The context of the discord conversation can be seen  [here
](https://discord.com/channels/1089836369431498784/1230680629067386931)

```python

import plotly.graph_objects as go
from nicegui import ui
from random import random


fig = go.Figure()
fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
plot = ui.plotly(fig).classes('w-full')

def add_trace():
[...]

PR #2516, opened 2024-02-05
Use Plotly.react instead of Plotly.newPlot

platinops:
This is just a refactor, triggered by the discussion in #2505.

It's better to use Plotly.react than Plotly.newPlot as stated in the [Plotly docs](https://plotly.com/javascript/plotlyjs-function-reference/#plotlyreact):

> `Plotly.react` has the same signature as `Plotly.newPlot` above, and can be used in its place to create a plot, but when called again on the same `<div>` will update it far more efficiently than `Plotly.newPlot`), which would destroy and recreate the plot. `Plotly.react` is as fast as `Plotly.restyle`/`Plotly.relayout` documented below.

Test code to verify events are still OK, also after changing the plot config: 

```python
from random import random

from nicegui import ui

def get_trace_data() -> dict:
    return {
        "x":[1, 2, 3],
        "y": [random(), random(), random()],
    }

data = [get_trace_data() for _ in range(1)]
config = {"displayModeBar": True}
fig = {
    "data":data,
    "config":config,
}
plot = ui.plotly(fig)
plot = plot.classes('w-full h-96').on('plotly_click', ui.notify)

def add_trace():
    fig["data"] = [*fig["data"], get_trace_data()]
    plot.update()

def toggle_mode_bar():
    fig["config"]["displayModeBar"] = not fig["config"]["displayModeBar"]
    plot.update()

ui.button('Add trace', on_click=add_trace)
ui.button('Toggle mode bar', on_click=toggle_mode_bar)

ui.run(host="0.0.0.0", port=5000)
```

PR #2957, opened 2024-04-25
Try to fix Plotly layout issues

falkoschindler:
This PR tries to solve #2917 by reverting PR #2516 to avoid Plotly issue https://github.com/plotly/plotly.js/issues/4856 (see /test1 below).
But reverting this PR doesn't seem to be enough, since demo plots on our documentation page still have a wrong height, which I reproduced with /test2 below. It turned out that we need to partly revert https://github.com/zauberzeug/nicegui/commit/328fbab10c449f593a0ace69960470af2b158e39 as well to avoid this problem.

```py
from random import random
from nicegui import ui

@ui.page('/test1')
def test1():
    fig = go.Figure()
    plot = ui.plotly(fig)

    def add_trace():
        fig.add_trace(go.Scatter(x=[1, 2, 3], y=[random(), random(), random()]))
        plot.update()

    ui.button('Add trace', on_click=add_trace)

@ui.page('/test2')
def test2():
    ui.timer(0.5, lambda: ui.plotly({}).classes('h-60'), once=True)
```

I'm still not very happy with this solution because it re-introduces unnecessary payload. So consider it as work in progress.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -193,16 +193,16 @@ class Element(Visibility):
     def _to_dict(self) -> Dict[str, Any]:
         return {
             'tag': self.tag,
-            **({'text': self._text} if self._text is not None else {}),
+            'text': self._text,
+            'style': self._style,
+            'slots': self._collect_slot_dict(),
+            'events': [listener.to_dict() for listener in self._event_listeners.values()],
             **{
                 key: value
                 for key, value in {
                     'class': self._classes,
-                    'style': self._style,
                     'props': self._props,
-                    'slots': self._collect_slot_dict(),
                     'children': [child.id for child in self.default_slot.children],
-                    'events': [listener.to_dict() for listener in self._event_listeners.values()],
                     'component': {
                         'key': self.component.key,
                         'name': self.component.name,
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -197,12 +197,12 @@ class Element(Visibility):
             **{
                 key: value
                 for key, value in {
-                    'style': self._style,
-                    'slots': self._collect_slot_dict(),
-                    'events': [listener.to_dict() for listener in self._event_listeners.values()],
                     'class': self._classes,
+                    'style': self._style,
                     'props': self._props,
+                    'slots': self._collect_slot_dict(),
                     'children': [child.id for child in self.default_slot.children],
+                    'events': [listener.to_dict() for listener in self._event_listeners.values()],
                     'component': {
                         'key': self.component.key,
                         'name': self.component.name,
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -193,13 +193,13 @@ class Element(Visibility):
     def _to_dict(self) -> Dict[str, Any]:
         return {
             'tag': self.tag,
-            'text': self._text,
-            'style': self._style,
-            'slots': self._collect_slot_dict(),
-            'events': [listener.to_dict() for listener in self._event_listeners.values()],
+            **({'text': self._text} if self._text is not None else {}),
             **{
                 key: value
                 for key, value in {
+                    'style': self._style,
+                    'slots': self._collect_slot_dict(),
+                    'events': [listener.to_dict() for listener in self._event_listeners.values()],
                     'class': self._classes,
                     'props': self._props,
                     'children': [child.id for child in self.default_slot.children],
--- a/nicegui/elements/plotly.vue
+++ b/nicegui/elements/plotly.vue
@@ -7,7 +7,6 @@ export default {
   async mounted() {
     await this.$nextTick();
     this.update();
-    this.set_handlers();
   },
   updated() {
     this.update();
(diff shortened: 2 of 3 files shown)


bmaranville:
Even if the behavior is reverted to use `Plotly.newPlot` on first draw, it will presumably still use `Plotly.react` for subsequent draws, so won't the issue in #2917 still occur?

Also I'm having trouble reproducing the initial height issue on Firefox and Chromium using the current nicegui development branch (`main`) - can you provide a minimal reproduction?  If I run this code in python with my browser resized to have a height < 450px it opens a new browser tab and displays properly.  Does it only fail in an `iframe`?

```python
from nicegui import ui

fig = go.Figure(go.Scatter(x=[1, 2, 3, 4], y=[1, 2, 3, 2.5]))
fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
ui.plotly(fig).classes('w-full h-40')

ui.run()
```
Result:
image

falkoschindler:
> Even if the behavior is reverted to use Plotly.newPlot on first draw, it will presumably still use Plotly.react for subsequent draws, so won't the issue in https://github.com/zauberzeug/nicegui/issues/2917 still occur?

As far as I know the issue doesn't occur when using `Plotly.react` for updates, only when using it for the initial creation.

The height issue seems to occur only if the creation of the Plotly element is slightly delayed:
```py
@ui.page('/test2')
def test2():
    ui.timer(0.5, lambda: ui.plotly({}).classes('h-60'), once=True)
```
The plot should be squeezed like this
Screenshot 2024-04-25 at 15 38 52

but grows to 450px like this
Screenshot 2024-04-25 at 15 38 39

bmaranville:
One issue might be that the `js-plotly-plot` class is actually added during the initial render of the plot, and is not present on the `div` element before then.  This change fixes the problem, but it's worth testing to see if it causes any additional problems (the presence of the class is used internally by `Plotly` to determine if there's already a plot there, but it also checks if the `div` is empty, so I don't think it will break anything)

```diff
diff --git a/nicegui/elements/plotly.py b/nicegui/elements/plotly.py
index cc912581..5c4b7294 100644
--- a/nicegui/elements/plotly.py
+++ b/nicegui/elements/plotly.py
@@ -35,8 +35,8 @@ class Plotly(Element, component='plotly.vue', libraries=['lib/plotly/plotly.min.
         super().__init__()
 
         self.figure = figure
+        self.classes('js-plotly-plot')
         self.update()
-        self._classes.append('js-plotly-plot')
 
     def update_figure(self, figure: Union[Dict, go.Figure]):
         """Overrides figure instance of this Plotly chart and updates chart on client side."""
```

bmaranville:
Is there a race condition between the application of the classes (like `h-60`) and the rendering of the widget?  Seems like there could be.

bmaranville:
I think the only way to control the height of the plotly plot in a stable way is to either

- set the height of the `div` into which the plot will be rendered _before_ rendering is started (add optional height kwarg to the ui.plotly command?)
- explicitly tell plotly what height you want, by adding it to the `layout` object, e.g. `ui.plotly({"layout": {"height": 260}})`, but this is awkward because plotly seems to only accept numerical values for height (not `em` or `rem`).

bmaranville:
There is another race condition: after creating a div with class `h-60` in a nicegui component, I am timing that it takes > 150ms before the height (as measured by `style.getPropertyValue('height')` becomes non-zero, which I'm guessing comes from a delay in loading the Tailwind css rules.

bmaranville:
I see that tailwind is being used client-side as a script rather than a CSS static file - I think that explains the delay (the script is analyzing the page content to dynamically insert style rules?)

It might be nice performance-wise to compile all the styles that would result from executing every combination in `tailwind_types` into a static CSS artifact that could be used in the client.  It would certainly avoid delays like these.

falkoschindler:
That's a very good point, @bmaranville. Quasar classes are generated via JavaScript, probably _after_ the DOM has been created. But I still don't understand and would like to find out why the original plotly.vue (before PR #2917) now requires these [changes in element.py](https://github.com/zauberzeug/nicegui/pull/2957/files#diff-4e6f7a582c83937f4400ccf05b2d69990ec81343908aa9c86b08f6f064b3bac9). And why exactly `text`, `style`, `slots` and `events` - nothing more, nothing less?

We _might_ end up setting an initial height somewhere via CSS or Plotly option. But it worked before PR #2917 and this PR fixes it again, so I'm not convinced that inducing another height value is necessary. I just don't understand and like the current solution, because it doesn't make much sense to me and it increases the payload again.

bmaranville:
The Quasar classes are not what I was referring to - it's the use of Tailwind as a script at runtime rather than using the Tailwind API to generate static CSS at build time.  The use of the script version is not recommended for production in the Tailwind documentation: https://tailwindcss.com/docs/installation/play-cdn

When using the Tailwind script there is automatically a race as an inline CSS is dynamically generated by a script that is reading the DOM and updating the CSS in real time, so after you add a class like `h-60`, it will have no immediate effect.  At some later time (> 150ms) the inline stylesheet will be updated, then the style of the class will change.

Most of the time you don't care, because everything just updates pretty fast, but if you have a library that is trying to read the style of a container before rendering itself, it won't work.  The times that it appears to work are probably because an additional `update()` is called, which then triggers the auto-resize (if `{config: {responsive: true}}` is set)

I noticed that in the demo code for reproducing the problem, `update()` was called twice (once by the `mounted()` function, once by the `updated()` function), in quick succession.  There seem to be multiple potential race conditions here then.  

I still suggest that the only (stable) way to make sure the first render of the plot is sized to the container is to size the container before rendering.  Sometimes other approaches will work because of race conditions.

falkoschindler:
After an exciting debugging session with @codingpaula we finally seem to understand what is happening when we simply revert PR #2516:

1. Vue receives a Plotly element with undefined attributes like `text`, `style`, `slots` or `events`, which are automatically set to `null`, `[]` or `{}`.
2. The plot is initially rendered using `Plotly.newPlot`.
3. Tailwind notices a new class "h-60" and adds it to the CSS definition in `<head>`, causing the Plotly container to shrink.
4. The updated `text` attribute causes a second rendering using `Plotly.react`, which is buggy and doesn't resize correctly.

We fixed it by replacing undefined element attributes _before_ rendering anything. This not only avoids the buggy resizing of `Plotly.react` (see https://github.com/plotly/plotly.js/issues/4856), it also should reduce the computation for other elements as well. Because we don't change the `elements` dictionary anymore while rendering, Vue doesn't try to re-render everything.

rodja (approved):

PR #2963, opened 2024-04-27
Add 'clear_color' attribute to ui.scene

chrschorn:
Adds a way to configure the `ui.scene` background color

### DIFF
--- a/nicegui/elements/scene.js
+++ b/nicegui/elements/scene.js
@@ -107,7 +107,7 @@ export default {
       this.$el.style.border = "1px solid silver";
       return;
     }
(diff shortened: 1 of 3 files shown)


falkoschindler (commented):
Thanks for the pull request, @chrschorn! Making the background color configurable is a great idea!

Just one question: You named the parameter `clear_color`, but always refer to it as "background color". Why not using the name `background_color`? I'd find it more intuitive.

chrschorn:
I got stuck halfway between the technical term and a user-friendly parameter name 😄 I renamed the user-facing to `background_color` now

falkoschindler (approved):
Nice! 🙂

PR #2964, opened 2024-04-27
Improve devcontainer experience

chrschorn:
Starting from a fresh clone, I faced these issues inside a no-cache build of the devcontainer:

`isort` is missing when you try to run `python main.py` -> I added it to the poetry dev dependencies

`pre-commit run` doesn't work out of the box (see screenshot below). This is because the way the devcontainer is setup is a little odd:
   - At build time, the source folder ("nicegui") is copied to the current directory and then installed https://github.com/zauberzeug/nicegui/blob/3fcfac637273ebcd051cd09c1ee5d7745863d044/.devcontainer/Dockerfile#L30-L31 With that, you end up with a copy of the `nicegui/` folder contents in `/` which is registered as a Python module (due to the `poetry install`), leading to a circular import. The source files are not needed to setup the dependencies in the container, which can be achieved with `poetry install --all-extras --no-root` and just the `poetry.lock`. 
   - The `.devcontainer` has an extra `}`, causing this to never run: https://github.com/zauberzeug/nicegui/blob/3fcfac637273ebcd051cd09c1ee5d7745863d044/.devcontainer/devcontainer.json#L38-L39 I removed the `}` and now it's correctly executed. Technically the devcontainer dockerfile wouldn't need to run `poetry` at all now which is worth considering.

Something else I noticed but didn't want to touch because I don't know how it impacts CI/CD: `CONTRIBUTING.md` mentions `autopep8 --max-line-length=120 --in-place --recursive .` It would be nice if this was included in pre-commit to not have to run multiple commands before commit/push.

Screenshot


pyproject.toml
```diff
@@ -61,6 +61,7 @@ urllib3 = ">=1.26.18,^1.26 || >=2.0.7" # https://github.com/zauberzeug/nicegui/s
 pyecharts = "^2.0.4"
 ruff = ">=0.3.5"
 pre-commit = ">=3.5.0"
+isort = "^5.11"
```
rodja, line 64: @falkoschindler isort seems to be required to run our website documentation:

```
nicegui-app-1  |     from . import anti_scroll_hack, documentation, fly, main_page, svg
nicegui-app-1  |   File "/app/website/documentation/__init__.py", line 2, in <module>
nicegui-app-1  |     from .intro import create_intro
nicegui-app-1  |   File "/app/website/documentation/intro.py", line 6, in <module>
nicegui-app-1  |     from .demo import demo
nicegui-app-1  |   File "/app/website/documentation/demo.py", line 5, in <module>
nicegui-app-1  |     import isort
nicegui-app-1  | ModuleNotFoundError: No module named 'isort'
```

rodja (approved):
Thanks for this PR @chrschorn. I tested the setup and it works as expected.

rodja:
> Something else I noticed but didn't want to touch because I don't know how it impacts CI/CD: CONTRIBUTING.md mentions autopep8 --max-line-length=120 --in-place --recursive . It would be nice if this was included in pre-commit to not have to run multiple commands before commit/push.

You are totally right! Would you like to create a pull request for this?

Issue #2951, opened 2024-04-24
When executing the run_method , a KeyError error occurs.


CrystalWindSnake:
### Description

Below is the code to reproduce the issue:
```python
from nicegui import ui

async def get_ticked():
    selected = await tree.run_method("getTickedNodes")
    ui.notify(selected)

tree = ui.tree(
    [{"id": "m1", "children": [{"id": "n1"}, {"id": "n2"}]}],
    label_key="id",
    tick_strategy="leaf",
)

ui.button("query", on_click=get_ticked)

ui.run()
```

Clicking the button prints the data normally. However, an error occurs on the backend.
And the error only occurs once; clicking the button again does not produce any error.

Error message:
```
  File "D:\Programs\Python\Python311\Lib\site-packages\nicegui\javascript_request.py", line 20, in resolve
    request = cls._instances[request_id]
              ~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: '3d068f2a-375a-4136-8318-0db5978e1cb0'
```

If the page creation is moved inside `@ui.page`, everything works fine.
```python
@ui.page('/')
def index():
      ...
```

Environment information:
- python: 3.11
- nicegui: 1.4.22

falkoschindler:
Hi @CrystalWindSnake,
Strangely I can't reproduce the error.
Do you see any JavaScript error on the developer console?

CrystalWindSnake:
Strange, everything is working fine now when I rerun the code today.

CrystalWindSnake:
@falkoschindler I encountered the same situation again. I suspect it's caused by the run_javascript method.
Here is the code to reproduce the issue. 

```python
from nicegui import ui

input = ui.element("input").classes("outline")

async def on_input():
    result = await ui.run_javascript(f"""return getElement({input.id}).value;""")
    print(f"input changed to {result}")

input.on("input", on_input)

ui.run()
```

When rapidly entering content into the input box, a KeyError error occurs on the backend.

falkoschindler:
Oh, I see. Both problems only occur if you don't use page functions and multiple browser tabs are connected to `/`. Then there are JavaScript responses from multiple tabs, but after the first one is processed the second one can't find a corresponding request.

First of all, you should move your code out of the [auto-index client](https://nicegui.io/documentation/section_pages_routing#auto-index_page) into a page function, when you're sending JavaScript requests to the client.

[...]

PR #2983, opened 2024-04-29
Raise Exception when awaiting JavaScript on auto-index page

falkoschindler:
This PR tries to solve #2951 by preventing the user from awaiting any JavaScript call on the auto-index page.

As it turned out, many existing pytests had to be updated, since they didn't use page functions but awaited JavaScript.

There's a last problem with `ui.notification`, which currently can't be used on auto-index pages at all. It continuously checks if it is still contained in the DOM or if it should be deleted. With this PR in place, this check raises a RuntimeError.

Apart from `ui.notification`, the `clipboard.read()` function isn't usable on the auto-index page. Conceptually this is correct. But we will need to update corresponding demos which don't use page functions at the moment.

- [x] update clipboard demos
- [x] make `ui.notification` usable on auto-index page --> trigger event when notification disappears on _one_ client and delete it for all others


falkoschindler:
As a side effect of fixing the deletion of notifications in 5235446e68c7130918a15ab1e1f186f1a720fa13, we now can register "dismiss" events using the new `on_dismiss` parameter and pass JavaScript handlers for action buttons like this:

```py
ui.notification('Hi!', close_button=True, on_dismiss=lambda: ui.notify('Dismissed'))
ui.notification('Ho!', actions=[{'label': 'Alert', ':handler': '() => alert("Alert!")'}])
```

falkoschindler:
There seems to be a problem with codemirror, but I'm not sure if it's coming from this PR or the main branch. I'll look into it later.

falkoschindler:
I think I fixed the problem in 154cb735cba5307dab1c88b55965ae3a08b13e3f.

rodja (approved):

PR #3017, opened 2024-05-04
Potential improvement uplift from grouping and sending Outbox messages at once

falkoschindler:
This PR experiments with sending all non-update messages at once to reduce the time for creating complex 3D scenes like discussed in #3009.

A demo with 10,000 boxes still takes a while:
```py
with ui.scene() as scene:
    scene.move_camera(y=-6, z=7)
    for x in range(100):
        for y in range(100):
            scene.box().scale(0.09).move(x / 10 - 5, y / 10 - 5)
```

I'm not sure if sending tens of thousands of messages at once is a good idea or if we should create chunks. And it is unclear what the remaining bottleneck is. Should we optimize the Python code, the payload or the JavaScript code on the client? Maybe combining multiple steps for creating a 3D object (8 at the moment) into one function call would also be a reasonable improvement.

### DIFF
--- a/nicegui/outbox.py
+++ b/nicegui/outbox.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import itertools
 from collections import deque
 from typing import TYPE_CHECKING, Any, Deque, Dict, Optional, Tuple
 
@@ -76,8 +77,8 @@ class Outbox:
                 coros.append(self._emit('update', data, self.client.id))
                 self.updates.clear()
 
-                for target_id, message_type, data in self.messages:
-                    coros.append(self._emit(message_type, data, target_id))
+                for target_id, messages in itertools.groupby(self.messages, key=lambda x: x[0]):
+                    coros.append(self._emit('messages', [(type_, data) for _, type_, data in messages], target_id))
                 self.messages.clear()
 
                 for coro in coros:
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -325,6 +325,9 @@ function createApp(elements, options) {
         disconnect: () => {
           document.getElementById("popup").style.opacity = 1;
         },
+        messages: (msg) => {
+          msg.forEach(([type, data]) => messageHandlers[type](data));
+        },
         update: async (msg) => {
           for (const [id, element] of Object.entries(msg)) {
             if (element === null) {


kovalp:
Let's see if I manage to test this...

kovalp:
I tested this change. It is a good move for a few reasons. 

 - the development version produces no flickering in Firefox and Chrome. 
 - the version in the main branch when viewed from Firefox produces a not completely filled square in many cases, while the version in the development branch always produced the expected field of boxes. 
 
 On the downside, the development version on Firefox results in a message "This page slows down the browser" with the button "Close".

falkoschindler:
Yes, processing all these messages takes a while, which freezes the browser for a short while.
In PR #3022 I'm experimenting with another promising approach to speed up the initialization of large 3D scenes.

evnchn:
Update: #3022 has been merged already.

philipsgithub:
Hi there:) any updates here?

philipsgithub:
since I am interested in this feature as well

evnchn:
This likely won't go anywhere since #3022 is merged. 

Only chance: if this can fix https://github.com/zauberzeug/nicegui/pull/3022#issuecomment-2816718045, then we can take another look at it.

evnchn:
@falkoschindler After looking at the PR code, I came to the conclusion that this PR is _not only about performance of 3D scenes, but may also have other positive effects in any bit of code which uses the Outbox_. 

However, I am not sure how great the benefit, if any. Particularly since UI updates, IIRC, are also groupped together natively already, so idk what else can group together additional to UI updates. 

AFAIK v2.13.0 already squeezed quite a lot of performance and efficiency, with its: 

- #4463 
- #4466 

So I am not sure if we should squeeze more. 

Also, **a longer message may not be necessarily better per-se**. 

- If the string's long enough, it can fall out of the CPU's cache, and the process will then be slow. Rather, separate short messages may outperform one massive JSON string. 
  - 1000000 bytes is 1MB after all. With cheap Celerons having 2MB L2 cache and no L3 cache, I'm very sure the browser isn't having half of L2 for the Socket.IO. 
- Consider also the network as well. If the network is slow, it is not a good idea to wait for all the messages to be sent in one big chunk, and _then_ execute the code. 
  - We may have to go so far as to measure the user's network connection and dynamically adjust how many messages we send per group, and that can get real nasty real quick. 

Still, **what crushes this PR is these two points**: 

- This PR is non-feasible as of writing, since it does not proactively try and ensure beneath 1000000 bytes, which it could well exceed in this PR if enough messages are groupped together. 
- This PR has merge conflict. 

---

Suggested action items:

- [ ] Retitle the PR for future reference, in case whoever needs this can search it. 
  - Suggested title: "Potential improvement uplift from grouping and sending Outbox messages at once"
- [ ] Close the PR. Whoever needs it can copy the code over, re-open it, and say why they need it. 
  - They must also bear the responsibility of benchmarking the grouping of messages under differing CPU speed, network bandwidth, network latency, as well as the trio of Chromium+Firefox+Safari (up to 24 combinations!)

PR #3022, opened 2024-05-05
Improve performance of 3D scenes by initializing all objects at once

falkoschindler:
This PR experiments with sending a single initialization message with all 3D objects at once to reduce the time for creating complex 3D scenes like discussed in https://github.com/zauberzeug/nicegui/discussions/3009.

A demo with 10,000 boxes runs even faster than with PR #3017:
```py
with ui.scene() as scene:
    scene.move_camera(y=-6, z=7)
    for x in range(100):
        for y in range(100):
            scene.box().scale(0.09).move(x / 10 - 5, y / 10 - 5)
```

This approach not only bundles multiple commands into one message, it also cuts the payload significantly by focusing on the actual data rather the sending loads of overhead for every attribute. It could be combined with PR #3017 to improve the speed for updating existing objects.

### DIFF
--- a/nicegui/elements/scene_object3d.py
+++ b/nicegui/elements/scene_object3d.py
@@ -42,10 +42,10 @@ class Object3D:
         self._name()
         return self
 
-    def send(self) -> None:
-        """Send the object to the client."""
-        self.scene.run_method(
-            'init',
+    @property
+    def data(self) -> List[Any]:
+        """Data to be sent to the frontend."""
+        return [
             self.type, self.id, self.parent.id, self.args,
             self.name,
             self.color, self.opacity, self.side_,
@@ -54,7 +54,7 @@ class Object3D:
             self.sx, self.sy, self.sz,
             self.visible_,
             self.draggable_,
-        )
+        ]
 
     def __enter__(self) -> Self:
         self.scene.stack.append(self)
--- a/nicegui/elements/scene_object3d.py
+++ b/nicegui/elements/scene_object3d.py
@@ -44,14 +44,17 @@ class Object3D:
 
     def send(self) -> None:
         """Send the object to the client."""
-        self._create()
-        self._name()
-        self._material()
-        self._move()
-        self._rotate()
-        self._scale()
-        self._visible()
-        self._draggable()
+        self.scene.run_method(
+            'init',
+            self.type, self.id, self.parent.id, self.args,
+            self.name,
+            self.color, self.opacity, self.side_,
+            self.x, self.y, self.z,
+            self.R,
+            self.sx, self.sy, self.sz,
+            self.visible_,
+            self.draggable_,
+        )
 
     def __enter__(self) -> Self:
         self.scene.stack.append(self)
--- a/nicegui/elements/scene.js
+++ b/nicegui/elements/scene.js
@@ -435,15 +435,35 @@ export default {
       }
       this.camera.updateProjectionMatrix();
     },
-    init(type, id, parent_id, args, name, color, opacity, side, x, y, z, R, sx, sy, sz, visible, draggable) {
-      this.create(type, id, parent_id, ...args);
-      this.name(id, name);
-      this.material(id, color, opacity, side);
-      this.move(id, x, y, z);
-      this.rotate(id, R);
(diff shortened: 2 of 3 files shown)


nicegui/elements/scene_object3d.py
```diff
@@ -42,16 +42,19 @@ def with_name(self, name: str) -> Self:
...
-        self._create()
-        self._name()
-        self._material()
-        self._move()
-        self._rotate()
-        self._scale()
-        self._visible()
-        self._draggable()
+    @property
+    def data(self) -> List[Any]:
```
kovalp, line 46: I would use a simple method, not a property.

kovalp (commented):
Let me see how this behaves.

kovalp:
This performs great! There is almost no delay for the shape 50x50 and maybe half-a-second delay for the 100x100 field.

Does this PR improve the updates too?

nicegui/elements/scene_object3d.py
falkoschindler: Why do you think a method would be better?

kovalp:
A method is easier to mock to facilitate unit testing.

falkoschindler:
> Does this PR improve the updates too?

@kovalp No, this PR only improves the object initialization. Updates will be tackled by PR #3017 or other improvements.

Anyway, I think we should merge this PR because it greatly improves the `ui.scene` element.

kovalp:
Yes, this is mergeable already. Thanks!

kovalp:
Is this improvement effective for existing scenes or only for the newly created scenes?

falkoschindler:
@kovalp It should also improve the performance when creating new scenes dynamically, e.g. on button click. Every scene sends an "init" event to the server which responds with an "init_objects" message containing all 3D object information. Thus it benefits from the new condensed format.

kovalp:
Ok, but what about existing scene objects?

I guess whenever I have time I will try to figure out it myself.

falkoschindler:
As I said in https://github.com/zauberzeug/nicegui/pull/3022#issuecomment-2095590511, updates to existing objects will be addressed by PR #3017 and other improvements.

kovalp:
I am looking forward to the acceleration of updates.

I will try to formulate my question in form of a code snippet later.

Maybe I am using the objects not in an optimal way.

evnchn:
Am I missing something? Since this is merged into main a long time ago, but there are still a lot of WebSocket messages going through to the client. 

{18314D8C-B4D4-4C45-9575-A763020CA97F}

falkoschindler:
Are you referring to the code from https://github.com/zauberzeug/nicegui/pull/3022#issue-2279500602?

I'm seeing only one "init_objects" message as expected:

Screenshot 2025-04-18 at 23 11 47

evnchn:
Yes. 

@falkoschindler Mind if you include the entire code snippet that you used for the MRE? From the `from nicegui import ui` to `ui.run`? (Suspect something's up with auto-index page VS page builder)

Also, I'm on 2.15.0, which I suppose the PR was merged and included long time ago.

evnchn:
```py
from nicegui import ui


@ui.page('/')
async def main_page():
    await ui.context.client.connected() # so that even the scene is created via SocketIO
    with ui.scene() as scene:
        scene.move_camera(y=-6, z=7)
        for x in range(20):
            for y in range(20):
                scene.box().scale(0.09).move(x / 10 - 5, y / 10 - 5)


ui.run(port=9999, reload=False)
```

Had to reduce the number of boxes since it lags like crazy on my Celeron N100.

falkoschindler:
Ok! I didn't `await ui.context.client.connected()`.

This PR works if the scene is already created when the client connects. The scene emits an "init" event and the server responds with an "init_objects" message containing all objects that have been created to this point. But if the scene is created after the client connects, individual object updates are sent right when objects are created(*). Only then the client sends "init" and receives "init_objects" once again.

(*) --> We should probably avoid sending these updates before receiving "init". Otherwise we send them twice.

This is just a quick assessment and I might be completely wrong. But if this is correct and we're sending duplicate information, we should tackle this problem in a separate issue.

PR #2932, opened 2024-04-22
Initialization animation failure in ECharts

CrystalWindSnake:
This PR is aimed at solving the issue where ECharts charts' animations are interrupted by resize operations.

Below is a code snippet to reproduce the issue:

```python
option = {
    "xAxis": {
        "type": "category",
        "data": ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"],
    },
    "yAxis": {"type": "value"},
    "series": [{"data": [150, 230, 224, 218, 135, 147, 260], "type": "line"}],
    "animationDuration": 5000,
}

ui.echart(option)
```

The expected behavior is that when the chart is initially displayed, the line chart should have a fade-in animation from left to right, lasting for 5 seconds.

PR #3056, opened 2024-05-12
Initialization animation failure in ECharts

CrystalWindSnake:
This PR is based on another [PR](https://github.com/zauberzeug/nicegui/pull/2932) and tries to solve the echarts initialization animation problem and the tailwindcss width style problem.


falkoschindler (approved):
Thanks, @CrystalWindSnake! This PR seems to solve the problem we noticed with PR #2932. I just had to adjust the timing of `test_nested_expansion` a little for the test to pass. All other tests, demos and my experiments from https://github.com/zauberzeug/nicegui/pull/2932#issuecomment-2071616561 seem to work. 👍🏻

PR #3090, opened 2024-05-17
Add example to subscribe to audio events

rodja:
This pull request adds a simple demo to illustrate the use of audio event subscriptions.


falkoschindler (approved):
Ok. There's already an event subscription in the main demo. But I guess another demo can't hurt.

PR #3107, opened 2024-05-20
Make adding Javascript properties easier

jdoiro3:
## Background

https://github.com/zauberzeug/nicegui/pull/896 introduced dynamic props via adding `:` at the beginning of the property name, where the value (a Python `str` representing JavaScript) would be `eval`uated on the client side. This works great but is somewhat limiting. For example, it would be nice to keep JavaScript code in a `.js` file instead of embedding it in Python.  ~Another limitation of the current implementation is that only top level props are `eval`uated (might be wrong). `Elements` such as `aggrid` may have dynamic properties deeper in the `options` for example.~ this was incorrect.

> [!NOTE]  
> Please let me know if any points above are wrong.

## What This PR Adds/Changes?

- ~Removes the need for `:` before properties that are dynamic in favor of explicitly using `ui.js` as the value.~ This was reverted. Dynamic props via the colon (`:myProp`) method can now also be at any level of the props.
- Adds the ability for dynamic props to be nested (see `aggrid` example below).
- Adds the ability to define a dynamic prop with an object from an external `JavaScript` file.

## Example with `AgGrid`

```python
#!/usr/bin/env python3
import time
import typing as ty
import random 
from pathlib import Path

from nicegui import ui


def str_time_prop(start, end, time_format, prop):
    stime = time.mktime(time.strptime(start, time_format))
    etime = time.mktime(time.strptime(end, time_format))
    ptime = stime + prop * (etime - stime)
    return time.strftime(time_format, time.localtime(ptime))


def random_date(start, end, prop):
    return str_time_prop(start, end, '%m/%d/%Y %I:%M %p', prop)


rows = [
    {'id': 0, 'name': 'Alice', 'age': 18},
    {'id': 1, 'name': 'Bob', 'age': 21},
    {'id': 2, 'name': 'Lionel', 'age': 19},
    {'id': 3, 'name': 'Michael', 'age': 32},
    {'id': 4, 'name': 'Julie', 'age': 12},
    {'id': 5, 'name': 'Livia', 'age': 25},
    {'id': 6, 'name': 'Carol'},
]

def get_column_defs(
    columns: ty.Iterable[str], date_columns: ty.Iterable[str]
):
    column_defs = []
    for i, col in enumerate(columns):
        col_def = dict(
            headerName=col.upper().replace("_", " "),
            checkboxSelection=True if i == 0 else False,
            field=col,
            filter="agDateColumnFilter" if col in date_columns else "agTextColumnFilter",
        )
        if col in date_columns:
            col_def["valueFormatter"] = ui.js("(value) => { value ? new Date(Date.parse(value)) : null }")
            col_def["filterParams"] = {
                "comparator": ui.js(
                    obj="comparator", 
                    file=Path(__file__).parent / "test.js", 
                )
            }
        column_defs.append(col_def)
    return column_defs


grid = ui.aggrid(
    options={
        "columnDefs": get_column_defs(["name", "birthday"], ["birthday"]),
        "rowData": [
            dict(name=name, birthday=random_date("1/1/2008 1:30 PM", "1/1/2024 4:50 AM", random.random())) 
            for name in ["Joe", "Ashlynn", "Dave", "Bill", "Steve"]
        ],
        "defaultColDef": dict(sortable=True, resizable=True),
        "pagination": "true",
        "rowSelection": "single",
        "onGridReady": ui.js("(params) => params.columnApi.autoSizeAllColumns()"),
    },
)

ui.run()

```

*test.js*
```javascript
export function comparator(filterDate, cellValue) {
  if (cellValue == null) return -1;
  let cellDate = new Date(Date.parse(cellValue)).setHours(0, 0, 0, 0);
  if (cellDate === filterDate) {
    return 0;
  }
  if (cellDate < filterDate) {
    return -1;
  } else if (cellDate > filterDate) {
    return 1;
  }
  return 0;
}
```

*result*
Screenshot 2024-05-20 at 4 16 34 PM

## TODO

- [x] don't deprecate the colon method.
- [ ] update docs if @falkoschindler likes the idea.
- [ ] add testing.
- [x] figure out how to make file imports cleaner

### DIFF
--- a/nicegui/js.py
+++ b/nicegui/js.py
@@ -6,13 +6,13 @@ from .nicegui import app
 
 class js(str):
 
-    def __new__(cls, obj: str, file: Optional[Path] = None, prop: Optional[str] = None):
+    def __new__(cls, obj: str, file: Optional[Path] = None):
         val: str
         if file:
             file_resolved = file.resolve()
             assert file_resolved.is_file() and file_resolved.suffix == '.js', f'{file_resolved} is not a .js file.'
             app.add_static_file(local_file=file_resolved, url_path=f'/user-static-js/{file.stem}')
-            val = f'_nicegui_js_import:{{"url": "/user-static-js/{file.stem}", "obj": "{obj}", "prop": "{prop or obj}"}}'
+            val = f'_nicegui_js_import:{{"url": "/user-static-js/{file.stem}", "obj": "{obj}"}}'
         else:
             val = f'_nicegui_need_to_eval_:{obj}'
         return str.__new__(cls, val)
--- a/nicegui/js.py
+++ b/nicegui/js.py
@@ -3,6 +3,10 @@ from typing import Optional
 
 from .nicegui import app
 
+# values should match ones in static/nicegui.js
+NEED_TO_EVAL = '_nicegui_need_to_eval_:'
+NEED_TO_IMPORT = '_nicegui_need_to_import_:'
+
 
 class js(str):
 
@@ -12,7 +16,7 @@ class js(str):
             file_resolved = file.resolve()
             assert file_resolved.is_file() and file_resolved.suffix == '.js', f'{file_resolved} is not a .js file.'
             app.add_static_file(local_file=file_resolved, url_path=f'/user-static-js/{file.stem}')
-            val = f'_nicegui_js_import:{{"url": "/user-static-js/{file.stem}", "obj": "{obj}"}}'
+            val = f'{NEED_TO_IMPORT}{{"url": "/user-static-js/{file.stem}", "obj": "{obj}"}}'
         else:
-            val = f'_nicegui_need_to_eval_:{obj}'
+            val = f'{NEED_TO_EVAL}{obj}'
         return str.__new__(cls, val)
--- a/nicegui/js.py
+++ b/nicegui/js.py
@@ -1,6 +1,7 @@
-from .nicegui import app
-from typing import Optional
 from pathlib import Path
+from typing import Optional
+
+from .nicegui import app
 
 
 class js(str):
@@ -9,9 +10,9 @@ class js(str):
         val: str
         if file:
             file_resolved = file.resolve()
-            assert file_resolved.is_file() and file_resolved.suffix == ".js", f"{file_resolved} is not a .js file."
-            app.add_static_file(local_file=file_resolved, url_path=f"/user-static-js/{file.stem}")
+            assert file_resolved.is_file() and file_resolved.suffix == '.js', f'{file_resolved} is not a .js file.'
+            app.add_static_file(local_file=file_resolved, url_path=f'/user-static-js/{file.stem}')
             val = f'_nicegui_js_import:{{"url": "/user-static-js/{file.stem}", "obj": "{obj}", "prop": "{prop or obj}"}}'
         else:
-            val = f"_nicegui_need_to_eval_:{obj}"
+            val = f'_nicegui_need_to_eval_:{obj}'
         return str.__new__(cls, val)
--- a/nicegui/js.py
+++ b/nicegui/js.py
@@ -1,0 +1,17 @@
+from .nicegui import app
+from typing import Optional
+from pathlib import Path
+
+
+class js(str):
+
+    def __new__(cls, obj: str, file: Optional[Path] = None, prop: Optional[str] = None):
+        val: str
+        if file:
+            file_resolved = file.resolve()
+            assert file_resolved.is_file() and file_resolved.suffix == ".js", f"{file_resolved} is not a .js file."
+            app.add_static_file(local_file=file_resolved, url_path=f"/user-static-js/{file.stem}")
+            val = f'_nicegui_js_import:{{"url": "/user-static-js/{file.stem}", "obj": "{obj}", "prop": "{prop or obj}"}}'
+        else:
+            val = f"_nicegui_need_to_eval_:{obj}"
+        return str.__new__(cls, val)
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -32,13 +32,13 @@ async function parseElements(raw_elements) {
     }
   );
   // JSON.parse reviver cannot be async so we perform this hacky way to import the module
-  let ref = importRefs[0];
-  let url = ref.url;
-  let obj = ref.obj;
-  let prop = ref.prop;
-  Object.keys(ref).forEach((key) => delete ref[key]);
-  ref[prop] = (await import(url))[obj];
-
+  for (let ref of importRefs) {
+    let url = ref.url;
+    let obj = ref.obj;
+    let prop = ref.prop;
+    Object.keys(ref).forEach((key) => delete ref[key]);
+    ref[prop] = (await import(url))[obj];
+  }
   console.log(parsedElements);
   return parsedElements;
 }
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -10,37 +10,46 @@ const loaded_components = new Set();
 const eval_identifier = "_nicegui_need_to_eval_:";
 const import_identifier = "_nicegui_js_import:";
 
+async function transformObj(item, transformer) {
+  if (Array.isArray(item)) {
+    let newArr = [];
+    for (const val of item) {
+      let result = await transformer(val);
+      newArr.push(await transformObj(result, transformer));
+    }
+    return newArr;
+  } else if (typeof item === "object") {
+    let newObj = {};
+    for (const k in item) {
+      newObj[k] = await transformObj(await transformer(item[k]), transformer);
+    }
+    return newObj;
+  } else {
+    let v = await transformer(item);
+    return v;
+  }
+}
+
 async function parseElements(raw_elements) {
-  let importRefs = [];
   let parsedElements = JSON.parse(
     raw_elements
       .replace(/&#36;/g, "$")
       .replace(/&#96;/g, "`")
       .replace(/&gt;/g, ">")
       .replace(/&lt;/g, "<")
-      .replace(/&amp;/g, "&"),
-    (_, val) => {
-      if (typeof val === "string" && val.startsWith(eval_identifier)) {
-        return eval(val.slice(eval_identifier.length));
-      } else if (typeof val === "string" && val.startsWith(import_identifier)) {
-        const js_import = JSON.parse(val.slice(import_identifier.length));
-        importRefs.push(js_import);
-        return js_import;
-      } else {
-        return val;
-      }
-    }
+      .replace(/&amp;/g, "&")
   );
-  // JSON.parse reviver cannot be async so we perform this hacky way to import the module
-  for (let ref of importRefs) {
-    let url = ref.url;
-    let obj = ref.obj;
-    let prop = ref.prop;
-    Object.keys(ref).forEach((key) => delete ref[key]);
-    ref[prop] = (await import(url))[obj];
-  }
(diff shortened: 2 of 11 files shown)


nicegui/static/nicegui.js
```diff
@@ -7,16 +7,40 @@ let mounted_app = undefined;
...
   );
+  // JSON.parse reviver cannot be async so we perform this hacky way to import the module
+  for (let ref of importRefs) {
+    let url = ref.url;
+    let obj = ref.obj;
+    let prop = ref.prop;
+    Object.keys(ref).forEach((key) => delete ref[key]);
+    ref[prop] = (await import(url))[obj];
+  }
+  console.log(parsedElements);
```
jdoiro3, line 42: @jdoiro3 - delete this.

jdoiro3:
@falkoschindler, feel free to take a look but I'm realizing this PR should still be in draft since the Js from an external file doesn't work in all cases. Might remove it.

nicegui/js.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def __new__(cls, obj: str, file: Optional[Path] = None):
+        val: str
+        if file:
+            file_resolved = file.resolve()
+            assert file_resolved.is_file() and file_resolved.suffix == '.js', f'{file_resolved} is not a .js file.'
+            app.add_static_file(local_file=file_resolved, url_path=f'/user-static-js/{file.stem}')
+            val = f'_nicegui_js_import:{{"url": "/user-static-js/{file.stem}", "obj": "{obj}"}}'
+        else:
+            val = f'_nicegui_need_to_eval_:{obj}'
```
jdoiro3, line 17: @jdoiro3 - think about these special string 'keys' more. Should they be something else?

nicegui/static/nicegui.js
```diff
@@ -7,16 +7,49 @@ let mounted_app = undefined;
...
+async function parseElements(raw_elements) {
+  let parsedElements = JSON.parse(
     raw_elements
       .replace(/&#36;/g, "$")
       .replace(/&#96;/g, "`")
       .replace(/&gt;/g, ">")
       .replace(/&lt;/g, "<")
       .replace(/&amp;/g, "&")
   );
+  const parsedElementsWithJs = await transformObj(parsedElements, async (val) => {
```
jdoiro3, line 42: Maybe change `parsedElementsWithJs` to `parsedElementsEvaluated`?

nicegui/static/nicegui.js
```diff
@@ -142,16 +176,6 @@ function renderRecursively(elements, id) {
...
-  Object.entries(props).forEach(([key, value]) => {
-    if (key.startsWith(":")) {
-      try {
-        props[key.substring(1)] = eval(value);
-        delete props[key];
-      } catch (e) {
-        console.error(`Error while converting ${key} attribute to function:`, e);
-      }
-    }
-  });
```
jdoiro3, line 154: @jdoiro3 - Revert this change.

jdoiro3: Resolved in https://github.com/zauberzeug/nicegui/pull/3107/commits/30ec33c8f059651137a5cb8562f28a9c877014e4.

epmb1n:
How do I make an AGgrid column as a button? its like make the cellRender as the  jsclass

falkoschindler:
Hi @jdoiro3,

Thanks for the pull request! I finally found the time to look into it. Let me try to break down the different aspects:

- **Nested JavaScript attributes**
    This is already possible with the existing implementation and the colon notation:

	```py
	ui.aggrid({
	    'columnDefs': [
	        {'headerName': 'Make', 'field': 'make'},
	        {'headerName': 'Model', 'field': 'model'},
	        {'headerName': 'Price', 'field': 'price', ':valueFormatter': '(params) => `$${params.value.toLocaleString()}`'},
	    ],
	    'rowData': [
	        {'make': 'Toyota', 'model': 'Celica', 'price': 35000},
	        {'make': 'Ford', 'model': 'Mondeo', 'price': 32000},
	        {'make': 'Porsche', 'model': 'Boxster', 'price': 72000}
	    ],
	})
	```

- **Nicer API via `ui.js`?**
    One could argue that using `ui.js` is more intuitive than prepending the key with a colon. I'm just not sure about the `ui` namespace, because this is primarily for UI elements and functions. And especially when working with Vue and Quasar, the colon notation feels quite natural as well.

- **Embedding code from js files**
    The new `ui.js` allows to include JavaScript code from js files. This is a handy feature. But can't it simply by replaced with `ui.js(Path('some/file.js').read_text())` using pathlib? I see that you can't easily import individual objects this way, but you could store complex JavaScript functions in separate files to improve syntax highlighting and code completion.

As you can tell, I'm struggling a bit with the necessity of the added features. They wouldn't harm if I was convinced by the new API and implementation. But I have concernes with both of them:

- As mentioned above, `ui.js` doesn't feel quite right. Maybe `from nicegui import js` and `js.function(code)` and `js.import(path, object)` might be better?

- The implementation is quite complex. It kind of duplicates the traversal from [dynamic_properties.js](https://github.com/zauberzeug/nicegui/blob/959961932035bf928b0456139c250da667832cbe/nicegui/static/utils/dynamic_properties.js).

- The need for async imports further complicates the implementation. I'm not sure about potential timing issues now that `parseElements` is async.

- It feels pretty arbitrary and fragile to use two special string suffixes to indicate the purpose of a string. I'd love to avoid that if possible.

Maybe adding something like `ui.add_js()` (in analogy with `ui.add_css` etc.) which accepts `str | Path` would be an alternative to importing JavaScript code as part of element props. Then you can simply call the imported functions by name.

Long story short: Even though I love the initiative, I'm not fully convinced by the API and implementation. A new `ui.add_js()` and sticking with the existing colon notation might be equally useful with much less impact on the existing codebase.

falkoschindler (commented):
Hi @jdoiro3,

Thanks for the pull request! I finally found the time to look into it. Let me try to break down the different aspects:

- **Nested JavaScript attributes**
    This is already possible with the existing implementation and the colon notation:

	```py
	ui.aggrid({
	    'columnDefs': [
	        {'headerName': 'Make', 'field': 'make'},
	        {'headerName': 'Model', 'field': 'model'},
	        {'headerName': 'Price', 'field': 'price', ':valueFormatter': '(params) => `$${params.value.toLocaleString()}`'},
	    ],
	    'rowData': [
	        {'make': 'Toyota', 'model': 'Celica', 'price': 35000},
	        {'make': 'Ford', 'model': 'Mondeo', 'price': 32000},
	        {'make': 'Porsche', 'model': 'Boxster', 'price': 72000}
	    ],
	})
	```

- **Nicer API via `ui.js`?**
    One could argue that using `ui.js` is more intuitive than prepending the key with a colon. I'm just not sure about the `ui` namespace, because this is primarily for UI elements and functions. And especially when working with Vue and Quasar, the colon notation feels quite natural as well.

- **Embedding code from js files**
    The new `ui.js` allows to include JavaScript code from js files. This is a handy feature. But can't it simply by replaced with `ui.js(Path('some/file.js').read_text())` using pathlib? I see that you can't easily import individual objects this way, but you could store complex JavaScript functions in separate files to improve syntax highlighting and code completion.

As you can tell, I'm struggling a bit with the necessity of the added features. They wouldn't harm if I was convinced by the new API and implementation. But I have concernes with both of them:

- As mentioned above, `ui.js` doesn't feel quite right. Maybe `from nicegui import js` and `js.function(code)` and `js.import(path, object)` might be better?

- The implementation is quite complex. It kind of duplicates the traversal from [dynamic_properties.js](https://github.com/zauberzeug/nicegui/blob/959961932035bf928b0456139c250da667832cbe/nicegui/static/utils/dynamic_properties.js).

- The need for async imports further complicates the implementation. I'm not sure about potential timing issues now that `parseElements` is async.

- It feels pretty arbitrary and fragile to use two special string suffixes to indicate the purpose of a string. I'd love to avoid that if possible.

Maybe adding something like `ui.add_js()` (in analogy with `ui.add_css` etc.) which accepts `str | Path` would be an alternative to importing JavaScript code as part of element props. Then you can simply call the imported functions by name.

Long story short: Even though I love the initiative, I'm not fully convinced by the API and implementation. A new `ui.add_js()` and sticking with the existing colon notation might be equally useful with much less impact on the existing codebase.

jdoiro3:
Thanks for taking the time to review this. I know it takes a lot of time and effort.

- When reviewing the source code I missed the use of [convertDynamicProperties](https://github.com/zauberzeug/nicegui/blob/bbab081de130a538d9f437df961e079fb0cc842d/nicegui/static/utils/dynamic_properties.js#L1-L25) in `aggrid.js`. I think I missed this fact because I was defining my `comparator` with a multiline python `str`.

    ```python
    ":comparator": """
                (filterDate, cellValue) => {
                    if (cellValue == null) return -1;
                    let cellDate = (new Date(Date.parse(cellValue))).setHours(0, 0, 0, 0);
                    if (cellDate === filterDate) { return 0; }
                    if (cellDate < filterDate) { return -1; }
                    else if (cellDate > filterDate) { return 1; }
                    return 0;
                };
                """
    ````

    The function got evaluated to `undefined`. Not sure if this is happening because of an error on my part or a limitation of 
    the current implementation. In either case I appreciate you pointing this out and apologies if I confused anyone with my 
    initial comment.

- I agree that the `ui` namespace is wrong, but my opinion (not a strong one lol) that the use of a colon (`:`) for defining all props that need to be `eval`uated to `Javascript` isn't right since it seems to be specific to `Vue`. The colon is shorthand for `v-bind` (see docs [here](https://vuejs.org/guide/components/props.html#static-vs-dynamic-props)) and in the case of `aggrid` doesn't have meaning outside of `nicegui`. Seeing some python string wrapped in something like `js` clearly tells a reader of the code this isn't a normal string/prop and is in fact `Javascript`.

- I need to test your suggestion to use `ui.js(Path('some/file.js').read_text())` instead but wondering if it would cause the same issue that I encountered with my first bullet point with newlines. I'll need to test it.

I'm happy to continue working on this some more if you think that makes sense but understand you have other priorities. My co-coworkers and myself at my company really love this project and the appreciate all the hard work you guys put into it.

jdoiro3:
Another thing I encountered is the fact that we can't define props that are `Javascript` objects. For example, to add custom `cellRenderer` to an `aggrid` field you need to define a custom renderer `object` (see example [here](https://www.ag-grid.com/vue-data-grid/component-cell-renderer/?ref=blog.ag-grid.com)), which currently isn't possible b/c dynamic props are evaluated to `functions`.

```python





with this PR you'd be able to get this to work with the below example.

**Comment by epmb1n** [comment]
hello,bro
After clicking the button, how to execute the functions in the py file?Let's say there's a function
```
def mynotify():
      ui.notify('hello launch')
```
How to make
`this.eventListener = () => alert("Software Launched");`
Replace with mynotify()

falkoschindler:
@jdoiro3, I thought about your pull request again and again. I understand your concern about the colon `:` notation and declaring JavaScript code more explicitly would be nice. But my concerns from https://github.com/zauberzeug/nicegui/pull/3107#pullrequestreview-2090270819 are still standing (namespace `ui.js`, complexity of the implementation, potential timing issues, fragile string suffixes) and I don't see an easy solution. My fear is that we mess with a core part of the NiceGUI library and introduce very subtle bugs that will popup over the next months or so.

I know that this is your first contribution to this project and I hugely appreciate your initiative! But I come to the conclusion to close this pull request without merging it into main. I hope you understand my decision and I'm looking forward to your next contribution! 🚀

jdoiro3:
Definitely understand and think overall your comments and concerns make sense. Thanks for taking the time to review/consider these changes.

Issue #2977, opened 2024-04-29
Nginx subpath example doesn't redirect correctly when navigating to subpage with trailing slash


falkoschindler:
### Description

The Nginx subpath example should redirect from

- http://localhost/nicegui/subpage/

to

- http://localhost/nicegui/subpage

but leads to

- https://app:8080/subpage

Lucas-lyh:
Hey! It seems that rodja is working on other issue. I've tried to help solving this issue in PR. 
Adding proxy_redirect is helpful.

PR #3108, opened 2024-05-21
fix: fix nginx_subpath example by add proxy_redirect

Lucas-lyh:
#2977 issue is solved in this PR. 
This problem is because starlette operating path which has `/` in the end by redirect to a new url+path without `/`. 
But when proxy by Nginx, this url in docker is not the real url outside the proxy.
Nginx provide `proxy_redirect` to modify this url when proxy the redirect response.


rodja (approved):

rodja:
Wow, that's great @Lucas-lyh! I did not know about the `proxy_redirect` option. Thanks a lot.

PR #3118, opened 2024-05-23
Use local files for testing `ui.interactive_image`

falkoschindler:
This PR tries to fix our flaky pytests for `ui.interactive_image`. It seems like most failures came from fetching dummy images from a remote server, which could even block requests with a rate limit. Therefore I replaced those images with locally served copies.

I also overloaded the `screen.wait_for` method to allow waiting for a condition. This is handy when waiting for events in `test_loaded_event`.

Furthermore I noticed a minor bug in `screen.should_not_contain`:
```py
self.selenium.implicitly_wait(wait)
with pytest.raises(AssertionError):
    self.find(text)
self.selenium.implicitly_wait(self.IMPLICIT_WAIT)
```
In case `pytest.raises` itself raised, the implicit wait hasn't been set back to `IMPLICIT_WAIT`. I fixed it with a try-finally block.


rodja (approved):
Great!

PR #1931, opened 2023-10-31
Introduce ui.get to query the element tree

rodja:
Originated in https://github.com/zauberzeug/nicegui/discussions/1923#discussioncomment-7419532, this PR introduces `ui.get` to retrieve elements of the current page with a query language inspired by tortoise-orm's `filter` or BeautifulSoup's `find_all`. Therefore it also introduces a ~~`.tags`~~ `.keys` property on Element with wich the developer can set a list of queryable words.

When complete this PR should allow these kinds of operations:

```py
buttons = ui.get(type=ui.button)
headers = ui.get(key='header')  
action_buttons = ui.get(type=ui.button, key='action')
buttons_with_text = ui.get(type=ui.button, text='Click me')
buttons_witout_text = ui.get(type=ui.button).without(text='Click me')
buttons_within_header = ui.get(type=ui.button).within(key='header')
buttons_within_header_or_footer = ui.get(type=ui.button).within(key=['header', 'footer'])
```

The object which is returned by `ui.get` is called `Elements` (eg. group of elements) and allows setting props,classes,style on all containing elements at once:

```py
ui.get(key='important').classes('text-bold')
```

And it allows iterating over all elements:

```py
for element in ui.get(key='important'):
    print(element)
```

If `type=` is used the iterator automatically casts to this type:

```py
for stepper in ui.get(type=ui.stepper):
    stepper.next()
```

# ToDos

- [x] `type=`
- [x] iterating with typing
- [x] .classes
- [x] .style
- [x] .props
- [x] `text=`
- [x] `key=` and `.keys`
- [x] `.within(type=`
- [x] `.within(key=`
- [x] `.exclude(type=`
- [x] `.exclude(key=`
- [x] `.exclude(text=`
- [x] `.not_within(type=`
- [x] `.not_within(key=`
- [x] `.key` with string or List
- [ ] ~~`.type` with class or List~~ [simplify usage of Mixins](https://github.com/zauberzeug/nicegui/pull/1931#issuecomment-1788420084)
- [x] `.text` with string or List
- [ ] docstrings
- [x] documentation_get.py with demos
- [ ] looking through the examples and checking where `ui.get` would make sense
- [ ] discuss general impact on maintenance etc.

PR #3121, opened 2024-05-24
Integration test framework

rodja:
This pull request introduces a test framework which allows to write high-level integration tests which run as fast as unit tests.

---

Note: Partly based on PR #1931.

---

Open tasks:

- [x] `__str__`: We need to clarify and test the general layout of the generated string.
- [x] `user.should_not_see`: We need to handle `ui.notify`?
- [x] `user.gather_elements`: Do we really need nested `list` initializers? And should we work with sets to avoid duplicates?
- [x] Documentation for `ElementFilter`: Maybe "Styling & Appearance" is a better place for it?
- [x] `pycheck nicegui` isn't happy yet (mainly unused fixure arguments, `T` vs. `Element`, missing `_iterator`, missing docstrings)
- [x] Why do we need `response.encoding = 'utf-8'`? --> We don't.
- [x] What is the purpose of test_run.py?
- [x] Can we run (most of) test_element_filter.py without `Screen`?
- [x] test_multiple_pages: What do we test by calling `activate()`? It doesn't seem to make any difference.
- [x] test_navigation: The "back" and "forward" buttons are currently unused. But the following lines breaks the test.
	```py
    user.find('back').click()
    await user.should_see('Main page')
    user.find('forward').click()
    await user.should_see('Other page')
    ```
- [x] user interaction: We don't seem to support typing in `ui.number`, `ui.editor`, `ui.codemirror` etc.
- [x] user interaction: We seem to replace text rather appending it.
- [x] I haven't reviewed fixtures.py, since I could use some guidance with the overall idea.
- [x] I haven't reviewed element_filter.py in detail, since there are still so many type and linter errors.
- [x] I'm sceptical about the way we handle different `element_contents`. It allows a test like this to succeed:
    ```py
    async def test_xxx(user: User) -> None:
        ui.input(placeholder='x', value='y')
    
        await user.open('/')
        await user.should_see('x y')
    ```
- [x] `ElementFilter.within_kinds` has no test and does not seem to work
- [x] The first parameter of `ElementFilter.exclude` may be an `element` and not `kind`...
- [x] Add filter rules to docstring of `ElementFilter`.
- [x] Fix the rule for ancestor kinds.
- [x] Simplify ancestor marker rules?
- [x] Ensure pytest plugin is working.
- [x] Is it safe to provide "user" and "screen" as fixture names in a pytest plugin? -> Yes, without defining an entry point.
- [x] Remove pytest entry point from pyproject.toml.
- [x] Add conftest.py for backward compatibility.
- [x] Documentation (new section about `screen`, `user`, `create_user`, link to examples) -> PR #3413
- [x] Remove `user.activate` and `user.deactivate`?


rodja:
There are still some tests failing:

```
FAILED tests/test_favicon.py::test_emoji[\U0001f44b] - assert '\n        <s... </svg>\n    ' == '\n        <s... </svg>\n    '
FAILED tests/test_navigate.py::test_navigate_to[False] - AssertionError: Could not find "Test page"
FAILED tests/test_navigate.py::test_navigate_to[True] - AssertionError: Could not find "Test page"
FAILED tests/test_serving_files.py::test_get_from_static_files_dir[/static] - NameError: name 'Screen' is not defined
FAILED tests/test_serving_files.py::test_get_from_static_files_dir[/static/] - NameError: name 'Screen' is not defined
FAILED tests/test_storage.py::test_access_user_storage_from_fastapi - IndexError: pop from empty list
FAILED tests/test_storage.py::test_access_user_storage_from_background_task - IndexError: pop from empty list
ERROR tests/test_navigate.py::test_navigate_to[False] - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_navigate.py::test_navigate_to[True] - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_select.py::test_add_new_values[add-False-True] - AssertionError
ERROR tests/test_select.py::test_add_new_values[add-True-True] - AssertionError
ERROR tests/test_serving_files.py::test_media_files_can_be_streamed - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_adding_single_media_file - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_adding_single_static_file - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_auto_serving_file_from_video_source - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_storage.py::test_access_user_storage_from_fastapi - Failed: There were unexpected logs. See "Captured log call" below.
```

rodja:
```
FAILED tests/test_storage.py::test_access_user_storage_from_fastapi - IndexError: pop from empty list
FAILED tests/test_storage.py::test_access_user_storage_from_background_task - IndexError: pop from empty list
ERROR tests/test_select.py::test_add_new_values[add-False-True] - AssertionError
ERROR tests/test_select.py::test_add_new_values[add-True-True] - AssertionError
ERROR tests/test_serving_files.py::test_media_files_can_be_streamed - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_adding_single_media_file - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_get_from_static_files_dir[/static] - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_get_from_static_files_dir[/static/] - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_adding_single_static_file - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_serving_files.py::test_auto_serving_file_from_video_source - Failed: There were unexpected logs. See "Captured log call" below.
ERROR tests/test_storage.py::test_access_user_storage_from_fastapi - Failed: There were unexpected logs. See "Captured log call" below.
```

rodja:
@falkoschindler I think this monster is ready for review. We can write the documentation after you have worked through all the code.

falkoschindler (changes requested):
Ok, after two review sessions I end up with a list of open questions, tasks and issues. I'll add it to the original post.

falkoschindler:
@rodja I spotted one last rule that needs clarification:

> [An element is yielded if] The ancestors of the element include at least one of the specified kinds defined via `within`.

Is it really enough to match _one_ kind? Or should we change it to

> [An element is yielded if] The ancestors of the element include all of the specified kinds defined via `within`.

We would need to change the implementation as well and add a test.

nicegui/element_filter.py
```diff
@@ -0,0 +1,223 @@
...
+        - The element has none of the excluded markers.
+        - The element contains all of the specified content.
+        - The element contains none of the excluded content.
+
+        - The ancestors of the element include all of the specified instances defined via ``within``.
+        - The ancestors of the element do not include any of the specified instances defined via ``not_within``.
+        - The ancestors of the element include all of the specified kinds defined via ``within``.
+        - The ancestors of the element do not include any of the specified kinds defined via ``not_within``.
+        - The ancestors of the element include at least one element with all of the specified markers defined via ``within``.
+        - The ancestors of the element do not include any element with all of the specified markers defined via ``not_within``.
```
falkoschindler, line 63:
```suggestion
        - The ancestors of the element include all of the specified instances defined via ``within``.
        - The ancestors of the element do not include any of the specified instances defined via ``not_within``.
        - The ancestors of the element include all of the specified kinds defined via ``within``.
        - The ancestors of the element do not include any of the specified kinds defined via ``not_within``.
        - The ancestors of the element include all of the specified markers defined via ``within``.
        - The ancestors of the element do not include any of the specified markers defined via ``not_within``.
```
@rodja Now that I have fixed the rule for ancestor kinds, I wonder if we should also change the last two rules for symmetry. Currently they enforce single elements to include or exclude all markers at once. But I feel it is easier to understand if the required markers apply to the whole ancestor stack.

rodja: Yes, that is much more consistent!

falkoschindler: Ok, I'll change the implementation and pytests accordingly.

rodja:
>  Is it safe to provide "user" and "screen" as fixture names in a pytest plugin?

True. The magic fixture loading from plugins occasionally leads to unpredictable name clashes. See https://github.com/pytest-dev/pytest/issues/3834 and https://github.com/pytest-dev/pytest/issues/3966 for a lengthly discussions. The example of `test_client` provided by pytest-sanic and pytest-aiohttp is often provided. As https://github.com/yunstanford/pytest-sanic/issues/22, shows sanic changed their fixture (but still provide the old one as depreciated).

Still, I would really like to keep the simple fixture names `screen` and `user` because they are the used so much. For example:

```py
@pytest.mark.module_under_test(main)
async def test_basic_startup_appearance(user: User) -> None:
    """Test basic appearance of the chat app."""
    await user.open('/')
    await user.should_see('simple chat app')
    await user.should_see('message')
    ...
```

I have not found any other plugins which use that name. But they plugin space is vast. 
An alternative would be the explicit import of fixtures as we do it in 1.4.
But the pytest documentation explicitly mentions that this should not be done:

> Sometimes users will import fixtures from other projects for use, however this is not recommended: importing fixtures into a module will register them in pytest as defined in that module. This has minor consequences, such as appearing multiple times in pytest --help, but it is not recommended because this behavior might change/stop working in future versions.

https://docs.pytest.org/en/6.2.x/fixture.html#use-fixtures-in-classes-and-modules-with-usefixtures:~:text=Sometimes%20users%20will,in%20future%20versions.

I have experimented with explicit activation of our fixtures and have a working prototype where you need to pass `--with-nicegui-fixtures` to pytest to enable `user` and `screen`. But that also is not very nice because every developer needs to include the option every time the tests should be run.

@falkoschindler came up with an interesting pattern which could solve the dillema: We would provide the fixtures `nicegui_screen` and `nicegui_user` which can then be shorthanded  by developers in their own conftest.py like this:

```py
@pytest.mark.fixture
def screen(nicegui_screen):
    return nicegui_screen

@pytest.mark.fixture
def user(nicegui_user):
    return nicegui_user
```

But that would make using nicegui tests a bit more complicated. All in all, I'm in favour of just keeping at as it is right now in this pull request: 

1. provide a pytest plugin via entrypoint which directly offers the fixtures
2. use `screen` and `user` as fixture names
3. hope that virtual envs shield the users from the worst case
4. provide guidance on how to disable plugins which might have name clashes (`-p no:name` command line option and other techniques)
4. wait for pytest to come up with a better way to load/define fixtures in plugins

falkoschindler:
@rodja I have an alternative suggestion based on @codingpaula's experiments from yesterday: She struggled to get our pytest plugin (provided via poetry entrypoint) to work in a separate test project. Even though it works in the NiceGUI example, it doesn't seem to be correctly registered when used outside of the NiceGUI workspace.

But it does work when using the `pytest_plugins` variable in the user project:
```py
pytest_plugins = ['nicegui.testing.fixtures']
```
Strangely, this works even without defining the plugin via poetry in NiceGUI. We couldn't figure out why. But currently we have doubts that the magic registration of pytest plugins via poetry entrypoints actually works. Yes, RoSys also provides such a plugin. But do we have proof it actually works? We couldn't find any project that actually uses fixtures from the RoSys pytest plugin.

Maybe we are completely on the wrong track, and it would certainly help to chat about it together on Monday. Nevertheless, the `pytest_plugins` variable seems to allow for a rather nice solution:

1. Don't provide any pytest plugin via poetry entrypoints.
2. Remain with the short and handy fixture names `screen` and `user`.
3. Let the users put `pytest_plugins = ['nicegui.testing.fixtures']` in their conftest.py to explicitly use our fixtures.
4. Add a conftest.py next to fixtures.py with `from .fixtures import *` and a deprecation warning for backward compatibility.

The latter still allows writing `from nicegui.testing.conftest import *` in the user's conftest.py, so we don't break anything.

falkoschindler:
Regarding the filter rules: I noticed that "within(instance=)" is a bit nonsensical when used with multiple elements. Since it now only matches when the element's ancestors include _all_ given instances, you could as well specify the one closest to the element. All others don't contribute anything. Example:
```py
root > A > B > C > label

within(instance=[A, C])
within(instance=C)  # equivalent
```
But in favor of a very symmetric rule set, I think we can leave it as it is.

rodja:
**About the pytest plugin**

> But do we have proof it actually works? We couldn't find any project that actually uses fixtures from the RoSys pytest plugin

The Field Friend is using the `integration` fixture from RoSys to [setup it's own `system` fixture](https://github.com/zauberzeug/field_friend/blob/2757ebeb662528456dfa0aaf886c93f067183285/tests/conftest.py#L20-L35).

> Strangely, [using `pyplugin = ...`] works even without defining the plugin via poetry in NiceGUI.

Oh wow. I did not get that in their documentation. And also ChatGPT 4o and Claude Sonnet did not tell me about it when I asked how I could "not autoload a plugin, but do it explicitly". That is great!

A slightly modified strategy of what you are suggesting would be:

1. Provide a pytest plugin via poetry entrypoints which have the `nicegui_screen` and `nicegui_user` fixtures.
2. Provide another plugin which offers the the short and handy fixture names `screen` and `user` to let users put `pytest_plugins = ['nicegui.testing.shorthands']` in their conftest.py.
4. Add a conftest.py next to fixtures.py with `from .fixtures import *` and a deprecation warning for backward compatibility.

While it would make the documentation a bit more lengthly, small projects like our `chat_app` example would not need a `conftest.py` at all (when they are ok with the longer fixture names).

**About the `within` filter rule**

>  I noticed that "within(instance=)" is a bit nonsensical when used with multiple elements

I'm not so sure. Consider the scenario where there is another `C` below root:

```
root > A > B > C > label 1
        > C > label 2
```

`within(instance=[A, C])` must only yield `label 1` where `within(instance=C)` provides access to `label 1` and `label 2`. I just added a test to ensure this is possible.

falkoschindler:
**About the pytest plugin**

Ok, that makes three ways to use `nicegui.testing` (poetry entry point, `pytest_plugins`, old conftest.py). But with a good documentation this might be optimal. Let's discuss it tomorrow with @codingpaula. I think she already implemented it (partly) in PR #3411.

**About the `within` filter rule**

Sure, there can be multiple elements with marker or content "C". But one `C` can't be the same _instance_ like another `C`.

falkoschindler (approved):
I think this PR is finally ready to merge.
In 161c51f8b1840cd8f70360ca8ef4070ad388bc69 I removed the need for `activate()` and `deactivate()` by replacing `ui.navigate` whenever an attribute of a user is accessed. Since the slot stack manipulation was obsolete anyway, we could get rid of quite some code.
@rodja Feel free to merge if you agree with the latest changes. 🙂

Issue #3028, opened 2024-05-07
KeyError in `ui.log` after client disconnected


Thermaltech:
### Description

Hey! 
Not sure if this is a bug, or an issue with my code (see below), but it seems that when the ui.log element is used with the python logging module, the page / element throws an error under the following conditions:
1) max_lines is non-zero ( Not Equal to None )
2) page is changed to another page / closed
3) a time period elapses (I assume when nicegui is cleaning up the disconnected page / elements)

If I set max_lines to None (default), then no error occurs. as this seems to be related to the nicegui codebase in log.py, line 28/29 where the LogeElementhandler in my code is attempting to log an event / trim the list, but nicegui has already cleaned up and deleted the log element, hence the error.

```python
        while self.max_lines is not None and len(self.default_slot.children) > self.max_lines:
            self.remove(0)
```

The Error:

```python
--- Logging error ---
Traceback (most recent call last):
  File "/mnt/datastore/wm/ui_event_log.py", line 22, in emit
    self.element.push(msg)
  File "/usr/local/lib/python3.10/dist-packages/nicegui/elements/log.py", line 29, in push
    self.remove(0)
  File "/usr/local/lib/python3.10/dist-packages/nicegui/element.py", line 525, in remove
    self.client.remove_elements(elements)
  File "/usr/local/lib/python3.10/dist-packages/nicegui/client.py", line 304, in remove_elements
    del self.elements[element_id]
KeyError: 35
```

My Code:
```python
import logging

from nicegui import ui, Tailwind
import modules.config_loader as cfg

import ui_header
import ui_footer
from time import strftime, localtime

logger = logging.getLogger('wm.ui.event_log')

class LogElementHandler(logging.Handler):
    """A logging handler that emits messages to a log element."""

    def __init__(self, element: ui.log, level: int = logging.NOTSET) -> None:
        self.element = element
        super().__init__(level)

    def emit(self, record: logging.LogRecord) -> None:
        try:
            msg = f'{strftime("%A, %B %d, %Y, %H:%M:%S", localtime(record.created))} - {record.levelname} - {record.name} - {record.message}'
            self.element.push(msg)
        except Exception:
            self.handleError(record)


@ui.page('/event_log')
def content():

    wm_config = cfg.configs['wm'].config

    # Page Configuration
    ui.page_title(f"{wm_config['SETTINGS']['TITLE']} - Event Log")
    ui_header.content()
    ui_footer.content()

[...]

PR #3123, opened 2024-05-24
Warn about communication attempts with disconnected clients

falkoschindler:
This PR tries to solve issue #3028 where a memory leak in the application code only becomes apparent when trying to delete a UI element. To warn the user about a deleted client that is still in use, this PR checks the client whenever a message is enqueued in the outbox.

I tested with the following demo:
```py
import asyncio
import time
from nicegui import app, ui

handlers = []

async def loop():
    while True:
        for handler in handlers:
            handler()
        await asyncio.sleep(1)

app.on_startup(loop)

@ui.page('/')
def page():
    log = ui.log(max_lines=10)
    handlers.append(lambda: log.push(f'Hello at {time.time()}'))
```

When closing a client and waiting a few seconds, a warning appears that a deleted client is used. Without this PR, we need to wait until elements are deleted from the overflowing log element, which causes a KeyError.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -59,6 +59,7 @@ class Client:
         self.shared = shared
         self.on_air = False
         self._disconnect_task: Optional[asyncio.Task] = None
+        self._deleted = False
         self.tab_id: Optional[str] = None
 
         self.outbox = Outbox(self)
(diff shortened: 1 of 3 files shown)


nicegui/client.py
```diff
@@ -320,6 +321,12 @@ def delete(self) -> None:
         self.remove_all_elements()
         self.outbox.stop()
         del Client.instances[self.id]
+        self._deleted = True
+
+    def check_existence(self) -> None:
+        """Check if the client still exists and print a warning if it doesn't."""
+        if self._deleted:
+            log.warning('Client has been deleted but is still being used. This is a bug in the application code.')
```
rodja, line 329: Maybe we should write "This is a bug in your application code". Otherwise the distinction between library and application code is unclear for novice programmers.

rodja (approved):

Issue #3035, opened 2024-05-08
Leaflet: map flickers between centers after flyTo


backbord:
### Description

Hi all,

thanks for the great project! Unfortunately, I came across a possible regression with an undesired effect:

I've updated from NiceGUI 1.4.21 to 1.4.24 and now this small example flickers between the old and the new center after the button was clicked.

```python
from functools import partial
from nicegui import ui

with ui.row().classes('h-96 w-96'):
    m = ui.leaflet(center=(48.1, 11.6), zoom=10)  # Munich
    ui.button('Berlin', on_click=partial(m.run_map_method, "flyTo", [52.5, 13.4], 9, {"duration": 1.0}))

    m.on("map-moveend", lambda e: print(e.args))

ui.run()
```

There appears to be an endless loop issuing `moveend` events, alternating between the old and the new center, causing the map to flicker and not draw as desired. A pan or zoom with the mouse breaks the loop.

Sometimes, it doesn't go into this bad loops and just works as desired.

Output:
```
[...]
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
{'type': 'moveend', 'center': [52.5, 13.4], 'zoom': 9}
{'type': 'moveend', 'center': [48.1, 11.6], 'zoom': 9}
[...]
```

[...]

PR #3125, opened 2024-05-25
Prevent `ui.leaflet` from flickering after calling `flyTo()`

falkoschindler:
This PR attempts to solve #3035 by adding a short delay on client and server to allow both zoom and center to update without jumping back and forth. The delay of 0.02s is deliberately chosen above the lower bound of 0.016s for Windows. I hope it isn't too dependent on processor and network speed.

Can be tested with this  code:
```py
m = ui.leaflet(center=(48.1, 11.6), zoom=10)
m.on('map-moveend', lambda e: print(e.args['center'], e.args['zoom']))
ui.button('Berlin', on_click=lambda: m.run_map_method('flyTo', [52.5, 13.4], 9, {'duration': 1.0}))
```

There might still be glitches with multiple tabs connected to the auto-index page. But maybe it's good enough for now.

### DIFF
--- a/nicegui/elements/leaflet.js
+++ b/nicegui/elements/leaflet.js
@@ -62,7 +62,8 @@ export default {
       "preclick",
       "zoomanim",
     ]) {
-      this.map.on(type, (e) => {
+      this.map.on(type, async (e) => {
+        await this.$nextTick(); // NOTE: allow zoom and center to both be updated
         this.$emit(`map-${type}`, {
           ...e,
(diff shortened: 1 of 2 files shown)


afullerx:
This fix works well on Windows. I did my best to break it by opening the UI in multiple tabs, maxing out the CPU, and throttling the network to "slow 3G". In all cases it continued to function nicely. 👍

rodja (approved):

PR #3138, opened 2024-05-29
Add encoding of generated URLs

### LINKED ISSUE #3124: Audio with hashtags in filenames are not loaded
### Description

I have a folder with two audio files and I'm trying to play them:
- artist - #title.mp3
- artist - title.mp3

Minimal app:
```python
import urllib
from pathlib import Path

ui.audio(src="artist - #title.mp3", controls=True)  # this triggers an error
ui.audio(src=urllib.parse.quote_plus("artist - #title.mp3"),
         controls=True)  # this triggers an error
ui.audio(src=Path("artist - #title.mp3"),
         controls=True)  # this triggers an error
ui.audio(src="artist - title.mp3", controls=True)  # this works

ui.run(
    reload=False,
    native=True,
    dark=True,
    port=native.find_open_port(),
    title='Audio Demo'
)
```
When the app is run this is what I see on screen:
Screenshot 2024-05-25 alle 09 51 52

And this is what I see in the console:
```shell
$ python app.py
NiceGUI ready to go on http://localhost:8002
http://127.0.0.1:8002/artist+-+#title.mp3 not found
http://127.0.0.1:8002/_nicegui/auto/media/9ea021d92f620e392dc6848d57e79cca/artist -  not found
http://127.0.0.1:8002/_nicegui/auto/media/9ea021d92f620e392dc6848d57e79cca/artist -  not found
```
using Python 3.11.7 on macOS Sonoma 14.4

Tried URL escaping the file name and parsing file name with Path, but without success. Has this already happened to others before?

afullerx:
This PR attempts to resolve #3124 by encoding any special characters in URLs auto-generated from file names. This was done by adding `urllib.parse.quote()` to  `app.add_media_file()` and `app.add_static_file()`. 

The only difficult question was whether or not to also encode caller-provided URLs. Since `url_path` must be supplied in unencoded form to work with the FastAPI decorator, there should be no risk of double-encoding. The only downside of encoding it is that if a user passes in an `url_path` that includes special characters **and** relies on the return value being the same unencoded string, it could break their code. 

Since this is a pretty narrow edge case, I opted for this behavior to keep the operation of the methods more consistent. If you'd rather go the more conservative route of just returning the `url_path` provided value unencoded, that's obviously a simple change. However, then the return value description would have to be more complicated to fully describe the behavior.

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -1,4 +1,5 @@
 import inspect
+import urllib
 from enum import Enum
 from pathlib import Path
 from typing import Any, Awaitable, Callable, List, Optional, Union
@@ -176,12 +177,14 @@ class App(FastAPI):
         :param local_file: local file to serve as static content
         :param url_path: string that starts with a slash "/" and identifies the path at which the file should be served (default: None -> auto-generated URL path)
         :param single_use: whether to remove the route after the file has been downloaded once (default: False)
-        :return: URL path which can be used to access the file
+        :return: URL path in encoded form which can be used to access the file
         """
         file = Path(local_file).resolve()
         if not file.is_file():
             raise ValueError(f'File not found: {file}')
+
         path = f'/_nicegui/auto/static/{helpers.hash_file_path(file)}/{file.name}' if url_path is None else url_path
+        encoded_url = urllib.parse.quote(path)
 
         @self.get(path)
         def read_item() -> FileResponse:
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Thanks for the pull request, @afullerx!

I wondered if we could only URL-encode the `file.name`. But as you mentioned, `@self.get` needs the unencoded `path`. So I guess your solution is pretty optimal.

Regarding the breaking change: I would call it a bug that the returned URL couldn't be used to access the file if it contained special characters, which is fixed with this PR. If you really need the original `url_path`, it makes little sense to rely on the return value.

PR #3151, opened 2024-05-30
Add option to loosen security to allow click to trigger javascript

kyloe:
Added a second function to the mermaid.js interface to initialize the securityLevel to be 'loose'

This allows triggers of the form

click <nodeid> href "javascript:<function_name>()" to be triggered.

A proof of concept demo (work in progress) can be seen at 

[https://github.com/kyloe/nicegui_mermaid_demo/tree/main](https://github.com/kyloe/nicegui_mermaid_demo/tree/main)

### DIFF
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -30,6 +30,9 @@ export default {
       }
       is_running = false;
     },
+    initialize() {
+      mermaid.initialize({securityLevel: 'loose' });
+    }
   },
   props: {
     content: String,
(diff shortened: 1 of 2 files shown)


falkoschindler (changes requested):
Thanks for the pull request, @kyloe! I think we should add some changes though:

Can you, please, add a `initialize(config: dict)` method to `ui.mermaid` including a docstring? Otherwise the user would basically need to have read the source code to understand how to change the security level.

Apart from that, your test is invalid because it immediately creates the "Success" label.

kyloe:
Hi Falko

Looks like my first PR leaves something to be desired!

I will re-submit shortly

cheers :)

Ian B

kyloe:
Ive delved further and things are more complex - see

-- Pass in custom config for mermaid component to allow callback / click
support #1906

Would be grateful for any pointers or guidance as to where to head.
rgds
Ian B.

PR #3159, opened 2024-06-01
Introduce `ui.teleport` element

CrystalWindSnake:
This PR introduces a `ui.teleport` component to NiceGUI, mirroring the functionality of Vue's `Teleport`. It is the key for enabling the direct creation of any NiceGUI component within table cell slots(#3148).


falkoschindler (commented):
Thanks a lot for this pull request, @CrystalWindSnake! The teleport element feels pretty magical!

Here are some thoughts about it:

- The target container is specified using a CSS select. I guess this is helpful when addressing an HTML element like in the demo. But sometimes you might want to use a NiceGUI element like in the pytests. Could we support both, CSS strings _and_ references to a UI element?

- The `force_update` method feels a bit like the existing `update` method. Do you think we could simply re-use `update()`? Otherwise I fear users will call `update()` and wonder why the content doesn't appear in the target container.

- You added the `ui.teleport` element to the documentation section about control element. I'd suggest to move it to [the section about page layout](https://nicegui.io/documentation/section_page_layout#clear_containers), since it isn't limited to control elements and it is basically a very powerful layouting feature.

- I noticed that the teleported content in disappears when reloading the page. This is ok on individual pages, but confusing on the auto-index page. Do you think there's anything we can do about that?

CrystalWindSnake:
@falkoschindler Thank you for your help.

- we should be able to construct appropriate CSS selectors using `element.id`.
- `force_update()` is used in scenarios where the target container is destroyed and rebuilt, which might be different from the scenarios for `update()`.Additionally, `force_update` forces a re-render of `teleport` by updating the `key`, but I don't understand how this differs from props updates. I'm concerned it might cause other issues . However, I respect your decision and defer to your judgment .
- Agreed. it is more appropriate to classify the `ui.teleport` component as a layout element.
- I found a solution from the [Vue repository](https://github.com/vuejs/core/issues/5184#issuecomment-1065132253), but I don't understand the reason behind it. The issue is indeed resolved.

falkoschindler (approved):
Thanks again, @CrystalWindSnake!
Looks like the PR is ready to merge. 🚀

PR #3162, opened 2024-06-02
Inaccurate error message in `ui.image` when file does not exist

CrystalWindSnake:
When using ui.image, if the parameter is a non-existent file path, the following error message will appear in the backend. It is difficult to determine the exact source of the error. This PR fixes this .

```python
from nicegui import ui
from pathlib import Path

missing_path = Path('missing.png')
ui.image(missing_path)
```

error(not contain useful information to locate the source of the error):
```
TypeError: Type is not JSON serializable: WindowsPath
```


falkoschindler (approved):
Thanks for the pull request, @CrystalWindSnake! The previous error message was indeed not very helpful.
I moved the check from `ImageElement` into the `SourceElement` mixin to make it effective for many more UI elements like `ui.audio`, `ui.video` etc.:
```py
ui.audio(Path('foo.mp3'))
```

PR #364, opened 2023-02-11
Layout builder

rodja:
One idea from #117 was to create a subset of layout commands which are easier to understand and discover than the classic css/quasar/tailwind techniques. The "Layout Builder" is one such attempt.

### Features

This branch currently sketches out the following features (meaning it's not a full implementation but rather a proof of concept):

- auto completion + documentation for simple to understand layout and styling aspects
- using the builder pattern in combination with string literals for auto completion
- logical grouping of layout aspects (e.g. `align.center` and `align.children` are fundamentally different in css but presented together)
- the Layout of an existing element can be modify by accessing its layout property
- Layouts can also be created independently because they are self contained classes
- element.props and element.classes take precedence over element.layout definitions to allow customizations
- Layouts can be used as contexts to create generic groups, rows, columns: `with ui.layout('horizontal')`
- specialized Layouts provide additional styling (e.g. `ButtonLayout` to configure shape etc.)
- a Layout can be applied to (multiple) other Layouts via `add`
- a Layout can be duplicated by calling `.copy()`
- invalid layout aspects can raise error reports (e.g. `margin.top.auto()` is not possible with css)
- binding to Layout values like opacity
- 

A demo of this features is implemented in the [`layout_demo.py`](https://github.com/zauberzeug/nicegui/blob/layout_builder/layout_demo.py).

### ToDos / Ideas

- Layout.row() and Layout.column() should use a GroupLayout which provides arrangement of children (e.g. gap, align.children, ...)
- Some mechanism to allow Grid and Masonry Layouts

### Questions

- How does the name "Layout" affect header/footer/drawer and tabs? Should we rename the latter? Or incorporate it?
- How to provide smart colors which change automatically for dark mode?
- Can we return the specialized Layout type in base class methods? (e.g. `ui.button().layout.align.center()` should return `ButtonLayout`)
- Should we use tailwind classes for color to support `dark:` selector or use Quasar colors to support Theming with `ui.colors`?

PR #3175, opened 2024-06-06
Introduce `align_items` parameter for `ui.row`, `ui.column` and `ui.card`

falkoschindler:
Inspired by question #3174, this PR introduces a parameter to control how items are aligned within common layout elements. It is based on Tailwind's ["Align Items"](https://tailwindcss.com/docs/align-items) classes.

```py
with ui.row(align_items='center'):
    ui.button('Button')
    ui.label('Label')

with ui.column(align_items='center'):
    ui.button('Button')
    ui.label('Label')

with ui.card(align_items='center'):
    ui.button('Button')
    ui.label('Label')
```

### DIFF
--- a/nicegui/elements/card.py
+++ b/nicegui/elements/card.py
@@ -1,3 +1,5 @@
+from typing import Literal, Optional
+
 from typing_extensions import Self
 
 from ..element import Element
@@ -5,7 +7,9 @@ from ..element import Element
 
 class Card(Element):
 
-    def __init__(self) -> None:
+    def __init__(self, *,
+                 align_items: Optional[Literal['start', 'end', 'center', 'baseline', 'stretch']] = None,
+                 ) -> None:
         """Card
 
         This element is based on Quasar's `QCard <https://quasar.dev/vue-components/card>`_ component.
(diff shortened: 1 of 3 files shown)


rodja (approved):
Ok. It's really quite common. But so are [justify](https://tailwindcss.com/docs/justify-items) and [content alignment](https://tailwindcss.com/docs/align-content). And what about gap, self-alignment, item placement and so on? We either add them step-by-step as parameters or give something like the [LayoutBuilder](https://github.com/zauberzeug/nicegui/pull/364) another try.

rodja:
Other relevant discussions might be https://github.com/zauberzeug/nicegui/discussions/1806 and https://github.com/zauberzeug/nicegui/discussions/117. Also the recently announced https://github.com/rio-labs/rio Python-UI framework has build it's own layout-mechanism which could be an inspiration.

falkoschindler:
Yes, but because a complete layout system is still out of reach, I think we should improve NiceGUI's layouting options in smaller steps. I agree that `gap` would also be a useful parameter. But in contrast to the other alignments you mentioned, `align_items` is by far the one I use most frequently. Simply putting a label next to an input field or a button, you need to vertically center the row. But because we might add other parameters later, I chose the name `align_items` instead of just `align` or something.

PR #3186, opened 2024-06-08
Fixed npm import script.

natankeddem:
Hello,
The npm import script has some syntax errors in the typing that cause errors during runtime.


falkoschindler (approved):
Thanks for the pull request, @natankeddem!
Even though this file is not meant to be used by NiceGUI users, it's good to fix these type annotations for Python 3.8 support. 👍🏻

PR #3195, opened 2024-06-10
Added support for 3D graphing for echarts.

natankeddem:
I added the javascript library and updated the element to support a conditional import of echarts-gl. The library will automatically be imported if any root keys in the `options` contain the string "3D". In addition the library can be force imported with the `enable_3d` argument.

I did try to add a test but there is some issue with webgl not being recognized.


falkoschindler (approved):
Thanks a lot for this pull request, @natankeddem! I like it very much.
I've been thinking about the `enable_3d` parameter for quite a while. It seems a bit redundant. But otherwise it's nice that 3D charts work automatically and that, at the same time, there's an explicit way to enforce 3D in case the options don't include "3D" keys. 👍🏻
Let's merge!

natankeddem:
Thank you @falkoschindler, a pleasure as always.

PR #3202, opened 2024-06-11
Add skeleton element 

frankvp11:
The skeleton element is added to the UI module to provide a loading placeholder for content.
Inspired by #3082.
The documentation for the Skeleton Element can be found here: https://quasar.dev/vue-components/skeleton/#predefined-types
I wasn't sure of where I should place this element, as it didn't seem to really fit under styling but I don't think that it really fits in Audiovisual elements either..
Also, the API that Quasar uses (QBtn, QToolbar, etc) for defining the type might be unfamiliar to NiceGUI users, but I'm not sure if we can fix that nicely.

### DIFF
--- a/nicegui/elements/skeleton.py
+++ b/nicegui/elements/skeleton.py
@@ -1,0 +1,17 @@
+
+from ..element import Element
+
+
+class Skeleton(Element):
+    def __init__(self, type : str):
+        """Skeleton
+
+        This element is based on Quasar's `QSkeleton <https://quasar.dev/vue-components/skeleton>`_ component.
+
+        It serves as a placeholder for loading content in cards, menus and other component containers.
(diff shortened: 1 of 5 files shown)


falkoschindler (approved):

I think it's ok to stick with types like "QBtn", "QSlider" etc. We could replace them with "button", "slider" etc. But then we could also replace "QToggle" with "switch", because the corresponding NiceGUI element is `ui.switch`. This would be a rather confusing deviation from the QSkeleton API and brought me to the conclusion that we should simply keep Quasar's type names. In the end it's a skeleton element that mimics Quasar elements.

Once we're at it, I extended the skeleton API a little to include more props like animation, square, bordered and size.

Let's merge! 🚀

PR #3213, opened 2024-06-13
Add details and examples in docs for running apps with https

tmlmt:
As agreed, one of the points mentioned in https://github.com/zauberzeug/nicegui/discussions/3155.

**Question:** Shall I make a PR per doc item I'd like to modify? Or shall I make one PR covering all?

**Note:** there are several points in `CONTRIBUTING.md` that might be a bit confusing:
- I read: 
https://github.com/zauberzeug/nicegui/blob/be3803bd1289e9d9ec203ffd82cb042353430e89/CONTRIBUTING.md?plain=1#L48 but the pre-commit config file requires python3.8 and any overwrite would have to be committed into the dev branch. It seems python 3.8 is required for development?
- Is the [Additional Demos](https://github.com/zauberzeug/nicegui/blob/main/CONTRIBUTING.md#additional-demos) section still valid? It mentions a `more()` and some decorators that I think do not exist anymore?


falkoschindler:
Thanks for the pull request, @tmlmt!
Since HTTPS and NGINX is @rodja's expertise, I requested a review from him.

> Shall I make a PR per doc item I'd like to modify? Or shall I make one PR covering all?

I think small pull requests like this one are better. This way we can merge some PRs quickly while others need some discussion.

> It seems python 3.8 is required for development?

Yes. With "no special Python version" we mean that any officially supported Python version should work for development.

> Is the [Additional Demos](https://github.com/zauberzeug/nicegui/blob/main/CONTRIBUTING.md#additional-demos) section still valid?

No, this section seems to need an update. We refactored the documentation system a while ago.

rodja (changes requested):
Thanks for the config and docs update @tmlmt. Each example should be "executable". Therefor it would be great if you could add a simple nicegui `main.py` and a small `README.md` on how to start the example. Similar to `examples/nginx_subpath/README.md`.

tmlmt:
OK ! I understand the principle and have pushed a commit to do so.

I was thinking though that there is some extra complexity related with using the containerized nginx (e.g. the `resolver` directive, replacing the entire `nginx.conf` file) and the containerized version of NiceGUI (`http://app`) while it is more straight-forward when it comes to adding an extra nginx conf file to an existing nginx server. I propose therefore to keep a standalone `nginx_site.conf` file with a corresponding section in the README for self-hosting.

rodja (approved):
Great!

PR #3226, opened 2024-06-15
Improve date picker demo with input element and add similar demo for time picker

tmlmt:
As agreed, one of the points mentioned in https://github.com/zauberzeug/nicegui/discussions/3155.


falkoschindler (approved):
Thanks for the pull request, @tmlmt!

I noticed that the "auto-close" prop causes the menu to disappear even when trying to navigate between months or years. So I added a "Close" button instead.

tmlmt:
Good catch! Indeed. Will change on my own app :) thanks

Issue #3191, opened 2024-06-09
Changing min/max limits of `ui.range` breaks the element


falkoschindler:
### Discussed in https://github.com/zauberzeug/nicegui/discussions/3167

<div type='discussions-op-text'>

<sup>Originally posted by **bojh** June  3, 2024</sup>
### Question

I want to programmatically re-set the ui.range min/max boundaries and min/max values. 
```python
from nicegui import app, ui

my_range = ui.range(min=0, max=100, value={'min':5, 'max':95})
my_value = ui.label().bind_text_from(my_range, 'value', 
    backward=lambda v: f'v_min: {v["min"]}, v_max: {v["max"]}')

def set_bnd_val(b_min, b_max, v_min, v_max):
        my_range.props((f'min={b_min} max={b_max}'))
        my_range.value = {'min':v_min, 'max':v_max}

my_button_set1 = ui.button("value In: [40- 60]", on_click=lambda: set_bnd_val(40, 60, 45,55))
my_button_set2 = ui.button("value In: [ 0-100]", on_click=lambda: set_bnd_val( 0,100,  5,95))

ui.run()
```
... pressing buttons seems working well. The range slider does update and the new proportion is OK. 
The troubles come up touching the value handles: The reported values does create 'None' values, depending which handle is moved and the slider is no longer correct.

The question is what is wrong or missing to do it in the right way?
</div>

falkoschindler:
Here is a minimum example:
```py
r = ui.range(min=0, max=100, value={'min': 30, 'max': 70})
ui.button('Change limits', on_click=lambda: r.props('min=10 max=90'))
ui.label().bind_text_from(r, 'value', str)
```

After changing the limits, adjusting the range causes it to disappear and the value is `None`.

A plain Quasar example works fine though:
```html
<html>
  <head>
    <link href="https://cdn.jsdelivr.net/npm/quasar@2.16.4/dist/quasar.prod.css" rel="stylesheet" type="text/css" />
  </head>
  <body>
    <div id="q-app">
      <q-range v-model="range" :min="rangeLimits.min" :max="rangeLimits.max"></q-range>
      <q-btn label="Change limits" @click="changeLimits"></q-btn>
      <div>Current range: {{ range.min }} - {{ range.max }}</div>
    </div>
    <script src="https://cdn.jsdelivr.net/npm/vue@3/dist/vue.global.prod.js"></script>
[...]

PR #3242, opened 2024-06-18
Avoid changing min/max limits of `ui.range` breaking the element

Ezbaze:
Possible fix for #3191


falkoschindler (commented):
Thanks for your contribution, @Ezbaze!

However, as explained in https://github.com/zauberzeug/nicegui/issues/3191#issuecomment-2183155202, I don't think this PR is needed. Only when using the public `.props()` method, using colons `:` helps to preserve the float or integer data type of `min` and `max`. When writing to the private `_props` dictionary, this is not an issue. Therefore I'll close this PR.

PR #3244, opened 2024-06-19
Add Google one-tap authentication example

rodja:
This pull request is inspired by #3243 and shows how we can use Google one-tap authentication with NiceGUI.


falkoschindler (changes requested):
Somehow the demo doesn't work for me:
"Can't continue with google.com. Something went wrong."

Apart from that, I think we should add dotenv to a requirements file.

falkoschindler (changes requested):
Sorry, I'm still getting the same error.
Strangely, NiceGUI connects to "us-west.on-air.io"... But the authentication should work nonetheless I guess.
And the corresponding startup test is failing for some reason.

falkoschindler (commented):
Ok, I think I understand why the login didn't work for me: If you only add "localhost" as authorized JavaScript origin, the login will fail when accessing the app at 127.0.0.1. Maybe we should mention this pitfall.

Apart from that, I'd like to remove the dotenv dependency. Since we only use the `GOOGLE_CLIENT_ID` constant, we can as well add them to main.py (like [here](https://github.com/zauberzeug/nicegui/blob/8ae51404897f65bf5c2fbbfb321a38f9643f7317/examples/chat_with_ai/main.py#L7)).

falkoschindler (approved):

PR #3264, opened 2024-06-22
Introduce `update_from_pandas` method for `ui.table`

tmlmt:
Resolves: https://github.com/zauberzeug/nicegui/discussions/2731

Opening this PR as draft to request a check as to whether this is going in the right direction.

Principles followed:
- In order to reuse  the part of the code which checks the DataFrame and potentially casts into string the special types, took it out of the from_pandas method and even out of the class to avoid it being confused with a class method. Alternative is to keep it in the class with one or two leading underscores in the name.
- Only expect a DataFrame to be given as parameter, the rest can be updated  manually relatively easily (I though intend to update this PR if you accept the introduction of #3263 which enables a finer grain of definition of the columns upon calling the function). Alternative is to accept some of same other parameters as in from_pandas

### DIFF
--- a/nicegui/elements/table.py
+++ b/nicegui/elements/table.py
@@ -14,6 +14,24 @@ except ImportError:
     pass
 
 
+def _clean_dataframe(df: 'pd.DataFrame') -> 'pd.DataFrame':
+    def is_special_dtype(dtype):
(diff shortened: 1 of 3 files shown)


falkoschindler:
This feature has been developed further in PR #3525 and will be released shortly in version 2.0.0.
Therefore I'll close this pull request.

Issue #2535, opened 2024-02-10
@ui.refreshable refreshes 1 session from multiple diffrent sessions


Ezbaze:
### Description

1. Im trying to have a refreshable outside the @ui.page decorator so it can be reused on different pages.

```python
from nicegui import ui

class Test:
    @ui.refreshable
    def ui(self, name):
        ui.notify(f"Refreshing {name}")


@ui.page("/other_page")
def page_layout():
    _ui = Test().ui
    ui.input("INPUT", on_change=lambda event: _ui.refresh(event.value))
    _ui("START")


if __name__ in {"__main__", "__mp_main__"}:
    ui.run(
        host="0.0.0.0",
        port=4040,
)
```

2. I expect each session to have a seperate refreshable that will only refresh for the session it was asked to refresh from
3. The refreshable refreshes the elements from any session but only for 1 session (see video)

https://github.com/zauberzeug/nicegui/assets/68749104/38f8049c-b509-4195-9360-c2b4f4c2eae8

Additional notes:
1. Having the @ui.refreshable on it's own without a Class enclosing it makes every session refresh instead of just one.
2. Having the function that will be refreshed outside of the @ui.page and then wrapping it in a function inside the @ui.page + decorating it with @ui.refreshable does give the desired effect but you'd have to do this for every page / every refreshable you'd want to use ._.

Thank you

falkoschindler:
Hi @Ezbaze,

I'm not 100% sure why your code doesn't work as expected. I think it has to do with the way `@refreshable` keeps track of the current instance that needs to be refreshed when calling `refresh()`. By holding a reference `_ui` to `Test().ui`, this mechanism fails.

It's better to keep a reference to the `Test` instance:
```py
class Test:
    @ui.refreshable
    def ui(self, name):
        ui.notify(f"Refreshing {name}")

@ui.page("/")
def page_layout():
    test = Test()
    ui.input("INPUT", on_change=lambda event: test.ui.refresh(event.value))
    test.ui("START")
```

Ezbaze:
Hi @falkoschindler ,

I didnt think of that, but that does solve the issue I was facing :) 

Thank you

PR #3279, opened 2024-06-24
Add demos for using `ui.refreshable` with local scope

falkoschindler:
Inspired by #2535 and #3233, this PR adds 4 demos showing how to use `ui.refreshable` with different scope.


falkoschindler:
@rodja I simplified the demos by removing the `ui.input` element with access to the user storage and showing the current time instead.

rodja (changes requested):
The demos are great. But maybe we should use "Local scope variant A", "Local scope variant B" etc as headlines instead of "Local Scope (1/3)". The latter may change if we add another example which would break any hyperlinks.

rodja (approved):

Issue #3142, opened 2024-05-29
`ui.number`: change event argument differs from `value` property


falkoschindler:
### Discussed in https://github.com/zauberzeug/nicegui/discussions/3110

<div type='discussions-op-text'>

<sup>Originally posted by **CrystalWindSnake** May 21, 2024</sup>
### Question

```python
# notify 0 when deleting value
ui.number(value=66, on_change=lambda e: ui.notify(f"{e.value=} {e.sender.value=}"))
```
When the input is cleared, e.value is `0` , expect it to be `None` . this is bug?

</div>

Lucas-lyh:
It seems like this is designed to be 0.
In elements/number.py, overwrited function `_value_to_event_value` function convert value to zero when value is None.
Should it keep None?
```python
    def _value_to_event_value(self, value: Any) -> Any:
        return float(value) if value else 0
```

frankvp11:
Hmm, this is interesting.
I was playing around with this and noticed some interesting things.
Firstly, as is mentioned above, the fix for this issue would be what @Lucas-lyh  described. Switch the 0 to a None and it works. I'll submit a PR for this once I get home (unless either of you want to do it @falkoschindler )
However, the reason I said this was interesting, is because I tried putting in letters into the number element. I realize that the reason you can input 'e' is because of numbers like 1e6, as this is a common representation of them. However, the reason I am intrigued is because you can put the 'e' there without a prefix. Is this the expected behaviour? Are we supposed to be able to put e7, or should we require it to have a prefix?

PR #3281, opened 2024-06-24
ui.number value discrepancy

frankvp11:
This is the fix as described in #3142

### DIFF
--- a/nicegui/elements/number.py
+++ b/nicegui/elements/number.py
@@ -133,6 +133,3 @@ class Number(ValidationElement, DisableableElement):
         if value == '':
             return 0
         return self.format % float(value)
-
-    def _value_to_event_value(self, value: Any) -> Any:
-        return float(value) if value else None
--- a/nicegui/elements/number.py
+++ b/nicegui/elements/number.py
@@ -135,4 +135,4 @@ class Number(ValidationElement, DisableableElement):
         return self.format % float(value)
 
     def _value_to_event_value(self, value: Any) -> Any:
-        return float(value) if value else 0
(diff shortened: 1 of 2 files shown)


falkoschindler (changes requested):

However, this will result in an event value `None` when entering "0". 🤔

frankvp11:
Huh, I didn't see that coming... 
Because you've said that I've thought about the other possible cases where this can happen, and I believe that this would be the only one.
I think that doing something like this would fix it. Do you agree?
```python
def _value_to_event_value(value: Any) -> Any:
    if value == 0:
        return 0
    if value:
        return float(value)
    return None
```
This covers both cases where the value is 0 or 0.0. The other cases where the previous code could have failed is from some empty collection (list, dict, set, etc). I don't think we should worry about those though and can safely return None in those cases.

falkoschindler:
Yeah, the devil is in the details...

I think we can simplify your solution:
- if value is 0: return `0` (equal to `float(value)`)
- if value is truthy (i.e. not `None`): return `float(value)`
- else: return `None` (which should be `value`)

So basically we are converting to a float whenever `value` is not None:
```py
return float(value) if value is not None else value
```

But now that I think of it, what's the purpose of this conversion anyway? `value` should be of type `float | None` already. Should we simply remove this override of `_value_to_event_value` and use its base implementation, which simply returns `value`? It seems like its sole purpose was to convert `None` to 0.

frankvp11:
That's what it looks like. As for your implementation, it does indeed look simpler, I can change mine to that if you'd like?

PR #3309, opened 2024-07-03
Teleport Example

frankvp11:
As requested in the discord, this is an example of moving a chart into a table with ui.teleport.
I basically stole the chart and table from the documentation and wrote the CSS selector


falkoschindler (approved):
Together with ChatGPT I modified it so that the charts make more sense in the context of a table.

PR #3319, opened 2024-07-05
Support named slots in Element.move

NiklasNeugebauer:
This adds support for named slots in Element.move()

Most importantly `Element.move(index)` will now keep the current slot rather than moving into the default slot.

The new parameter `target_slot` can be used to move an element into a specific slot of the current parent element or the specified `target_element`.

Some special cases:
- Moving into a new `target_element` without specifying a `target_slot` will use the default slot.
- Moving into the current parent element without specifying a `target_slot` will keep the current slot.

I would like some opinions on the last point. If `target_element` is specified while `target_slot` is `None` should we always use the default slot to mimic the first case (i.e. even if the target is the current parent)?

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -498,12 +498,15 @@ class Element(Visibility):
             slot.children.clear()
         self.update()
 
-    def move(self, target_container: Optional[Element] = None, target_index: int = -1, target_slot: Optional[str] = None) -> None:
+    def move(self,
+             target_container: Optional[Element] = None,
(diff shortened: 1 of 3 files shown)


falkoschindler (approved):
Thanks for the pull request, @NiklasNeugebauer!
We decided to change the logic so that "default" is always the target slot if it isn't explicitly specified, independent of whether an element is moved to another container or not. This simplifies the implementation and is easier to grasp as a user.

PR #3320, opened 2024-07-05
shutdown with SIGINT  if autoreload is enabled

NiklasNeugebauer:
This adds the ability to fully shutdown the app even if `reload=True`.

It was inspired by encode/uvicorn#1103 (more specifically [this comment](https://github.com/encode/uvicorn/discussions/1103#discussioncomment-5359710))

This is currently missing Windows support.
According to ChatGPT, the following might work, though:
```py
parent_process = psutil.Process(parent_pid)
parent_process.send_signal(signal.CTRL_C_EVENT)
 ```
 
 Could someone test this?

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -127,17 +127,11 @@ class App(FastAPI):
         """Shut down NiceGUI.
 
         This will programmatically stop the server.
-        When auto-reload is enabled, this will only work on Linux and MacOS.
         """
         if self.native.main_window:
             self.native.main_window.destroy()
         if self.config.reload:
-            parent_pid = os.getppid()
-            if platform.system() == 'Darwin' or platform.system() == 'Linux':
-                os.kill(parent_pid, signal.SIGINT)
-            else:
-                raise NotImplementedError(
-                    'Shutting down the server with auto-reload enabled is only supported on Linux and MacOS.')
+            os.kill(os.getppid(), getattr(signal, 'CTRL_C_EVENT' if platform.system() == 'Windows' else 'SIGINT'))
         else:
             Server.instance.should_exit = True
 
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -1,4 +1,7 @@
 import inspect
+import os
+import platform
+import signal
 import urllib
 from enum import Enum
 from pathlib import Path
@@ -124,12 +127,17 @@ class App(FastAPI):
         """Shut down NiceGUI.
 
         This will programmatically stop the server.
-        Only possible when auto-reload is disabled.
+        When auto-reload is enabled, this will only work on Linux and MacOS.
         """
-        if self.config.reload:
-            raise RuntimeError('calling shutdown() is not supported when auto-reload is enabled')
         if self.native.main_window:
             self.native.main_window.destroy()
+        if self.config.reload:
+            parent_pid = os.getppid()
+            if platform.system() == 'Darwin' or platform.system() == 'Linux':
+                os.kill(parent_pid, signal.SIGINT)
+            else:
+                raise NotImplementedError(
+                    'Shutting down the server with auto-reload enabled is only supported on Linux and MacOS.')
         else:
             Server.instance.should_exit = True
 

### CODE REVIEW

**Comment by falkoschindler** [comment]
@NiklasNeugebauer ChatGPT's solution for Windows assumes to know `parent_pid`. But according to https://github.com/encode/uvicorn/discussions/1103#discussioncomment-5417741, `os.getppid` is not available on Windows, so we don't know the parent process ID. [This link](https://www.w3schools.com/python/ref_os_getppid.asp#:~:text=getppid()%20method%20returns%20the,on%20UNIX%20and%20Windows%20platforms.) on the other hand suggests that `os.getppid` _is_ available on Windows. 🤔

Can anyone confirm that this shutdown button works on Windows?
```py
import os
import signal
from nicegui import ui

ui.button('Shutdown', on_click=lambda: os.kill(os.getppid(), signal.SIGTERM))

ui.run()
```

python-and-fiction:
@falkoschindler  unfortunately, it doesn't work well on windows 10 ltsc, main thread is  still running.

python-and-fiction:
@falkoschindler 
I test following code in vscode on windows 10 ltsc after installing psutil, 
It wokrs well when I use `signal.CTRL_C_EVENT` and `signal.CTRL_BREAK_EVENT` , there is on python threads left.
I think `signal.CTRL_C_EVENT` and `signal.CTRL_BREAK_EVENT` are better than `signal.SIGTERM`
```python3
import os
import signal
import psutil
from nicegui import ui

def test():
    parent_process = psutil.Process(os.getppid())
    print(parent_process)
    parent_process.send_signal(signal.SIGTERM)
def test2():
    parent_process = psutil.Process(os.getppid())
    print(parent_process)
    parent_process.send_signal(signal.CTRL_C_EVENT)
def test3():
    parent_process = psutil.Process(os.getppid())
    print(parent_process)
    parent_process.send_signal(signal.CTRL_BREAK_EVENT)
# after SIGTERM,python thread is still running.
ui.button('kill by SIGTERM',on_click=test)
# after CTRL_C_EVENT, no python thread.
ui.button('kill by CTRL_C_EVENT',on_click=test2)
# after CTRL_BREAK_EVENT, no python thread.
ui.button('kill by CTRL_BREAK_EVENT',on_click=test3)

ui.run(native=True)
```

falkoschindler:
Ah ok, thanks a lot @python-and-fiction!
Since my Python installation doesn't know `signal.CTRL_C_EVENT`, I thought we should use `signal.SIGTERM`. But we can switch the signal depending on the platform.

python-and-fiction:
> Ah ok, thanks a lot @python-and-fiction! Since my Python installation doesn't know `signal.CTRL_C_EVENT`, I thought we should use `signal.SIGTERM`. But we can switch the signal depending on the platform.

@falkoschindler  For windows, the message from psutil may help you:
`only SIGTERM, CTRL_C_EVENT and CTRL_BREAK_EVENT signals are supported on Windows`

falkoschindler:
@python-and-fiction I just updated this branch to support `app.shutdown` with `ui.run(reload=True)` for Windows.
Would you mind testing it for us? Thanks! 😊

python-and-fiction:
> @python-and-fiction I just updated this branch to support `app.shutdown` with `ui.run(reload=True)` for Windows. Would you mind testing it for us? Thanks! 😊

@falkoschindler 
I replaced `app.py` from ` origin/shutdown-with-auto-reload`
The test code in vscode is:
```python3
from nicegui import ui,app

ui.button('Shutdown', on_click=app.shutdown)

ui.run(native=True)
```
The output is
```
Process Process-1:
  File "D:\Programs\Python\Python312\Lib\multiprocessing\process.py", line 314, in _bootstrap
    self.run()
  File "D:\Programs\Python\Python312\Lib\multiprocessing\process.py", line 108, in run
  File "E:\PSF\nicegui_app\.venv\Lib\site-packages\nicegui\native\native_mode.py", line 50, in _open_window
    webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
  File "E:\PSF\nicegui_app\.venv\Lib\site-packages\webview\__init__.py", line 209, in start
    guilib.create_window(windows[0])
  File "E:\PSF\nicegui_app\.venv\Lib\site-packages\webview\platforms\winforms.py", line 619, in create_window
    thread.Join()
KeyboardInterrupt
```
No python thread left.
It works perfectly. 

But for:
```python3
from nicegui import ui,app
def test_shutdown():
    print('\r',end='',sep='')
    app.shutdown()
ui.button('Shutdown', on_click=test_shutdown)
ui.button('Shutdown directly', on_click=app.shutdown)
ui.run(reload=True)
```
In vscode terminal,after clicking `Shutdown directly` ,The program will not respond .When I press enter in terminal, the program terminated.If I click `Shutdown` which run 'print('\r',end='',sep='')',It works well.
In normal terminal,It works well.
I think it is a bug about vscode.Do you want to add `print('\r',end='',sep='')` or `print('\n',end='',sep='')` in `app.shutdown`?
Or note it in document？

python-and-fiction:
ok,I find it.
I set `terminal.integrated.windowsEnableConpty` in vscode to disable, run `app.shutdown` directly in vscode terminal as same as in normal terminal.
It seems that it is a bug of node-pty, in `microsoft/node-pty` ,I find this:
https://github.com/microsoft/node-pty/issues/640
We could note it in document: 
1, disable `terminal.integrated.windowsEnableConpty` in vscode.
2, or add `print('\r',end='')` after running `app.shutdown` if  `terminal.integrated.windowsEnableConpty` is enabled.

NiklasNeugebauer:
I went ahead and checked myself (Windows 10, Python3.12).

This looks good. The current solution does indeed work when starting directly from Powershell.
And as described by @python-and-fiction, VSCode needs the extra newline.

When running `native=True` and `reload=True` we get a `KeyboardInterrupt ` printed, but I'd say that's better than not supporting a shutdown.

falkoschindler (approved):

falkoschindler:
Ok, thanks everyone! Let's merge!

PR #3369, opened 2024-07-18
Use ui.notification for ui.notify to simplify user simulation tests

rodja:


falkoschindler (approved):
Ok, it feels a bit sketchy to basically replace the core of `ui.notify` with `ui.notification`, but should work nonetheless. 👍🏻

PR #3378, opened 2024-07-21
Get drop information out from windowview

EmberLightVFX:
This is my WIP PR for issue https://github.com/zauberzeug/nicegui/discussions/3372
Right now all it does is allow you to drag and drop files into the webview and it will print out the path to the file from the main thread.

The issue I'm still having is how to access the running nicegui instance on the main thread as currently the check for droped files happens in a separate thread that's running from the main thread (not multiprocessing).

Are there any other way for me to check the drop_queue from within the main thread and from there create an event that could be picked up like `ui.on("drop", my_func)`?

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -21,14 +21,33 @@ try:
         # webview depends on bottle which uses the deprecated CGI function (https://github.com/bottlepy/bottle/issues/1403)
         warnings.filterwarnings('ignore', category=DeprecationWarning)
         import webview
+        from webview.dom import DOMEventHandler
     optional_features.register('webview')
 except ModuleNotFoundError:
     pass
 
+def on_drag(e: dict[str, Any]):
+    pass
+
+def on_drop(e: dict[str, Any], drop_event:Event, drop_queue: mp.Queue):
+    files = e['dataTransfer']['files']
+    if len(files) == 0:
+        return
+
+    for file in files:
+        drop_queue.put(file.get('pywebviewFullPath'))
+    drop_event.set()
+
+def bind_drop(window: webview.Window, drop_event:Event, drop_queue:mp.Queue) -> None:
+    window.dom.document.events.dragenter += DOMEventHandler(on_drag, True, True)
+    window.dom.document.events.dragstart += DOMEventHandler(on_drag, True, True)
+    window.dom.document.events.dragover += DOMEventHandler(on_drag, True, True)
+    window.dom.document.events.drop += DOMEventHandler(lambda e: on_drop(e, drop_event, drop_queue), True, True)
+
 
 def _open_window(
     host: str, port: int, title: str, width: int, height: int, fullscreen: bool, frameless: bool,
-    method_queue: mp.Queue, response_queue: mp.Queue,
+    method_queue: mp.Queue, response_queue: mp.Queue, drop_event: Event, drop_queue: mp.Queue,
 ) -> None:
     while not helpers.is_port_open(host, port):
         time.sleep(0.1)
@@ -47,7 +66,7 @@ def _open_window(
     closed = Event()
     window.events.closed += closed.set
     _start_window_method_executor(window, method_queue, response_queue, closed)
-    webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
+    webview.start(func=bind_drop, args=(window, drop_event, drop_queue), storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
 
 
 def _start_window_method_executor(window: webview.Window,
@@ -94,6 +113,9 @@ def _start_window_method_executor(window: webview.Window,
 
     Thread(target=window_method_executor).start()
 
+def got_drop(item: str):
+    print("got_drop externally")
+    print(item)
 
 def activate(host: str, port: int, title: str, width: int, height: int, fullscreen: bool, frameless: bool) -> None:
     """Activate native mode."""
@@ -105,17 +127,28 @@ def activate(host: str, port: int, title: str, width: int, height: int, fullscre
             time.sleep(0.1)
         _thread.interrupt_main()
 
+    def check_drop(drop_queue: mp.Queue, drop_event: Event) -> None:
+        while True:
+            while not drop_event.is_set():
+                time.sleep(0.1)
+            got_drop(drop_queue.get())
+            drop_event.clear()
+
     if not optional_features.has('webview'):
         log.error('Native mode is not supported in this configuration.\n'
                   'Please run "pip install pywebview" to use it.')
         sys.exit(1)
 
+    drop_event = mp.Event()
+    drop_queue = mp.Queue()
+
     mp.freeze_support()
-    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue
+    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue, drop_event, drop_queue
     process = mp.Process(target=_open_window, args=args, daemon=True)
     process.start()
 
     Thread(target=check_shutdown, daemon=True).start()
+    Thread(target=check_drop, args=(drop_queue, drop_event), daemon=True).start()
 
 
 def find_open_port(start_port: int = 8000, end_port: int = 8999) -> int:
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -29,25 +29,24 @@ except ModuleNotFoundError:
 def on_drag(e: dict[str, Any]):
     pass
 
-def on_drop(e: dict[str, Any], drop_event:Event, drop_queue: mp.Queue):
+def on_drop(e: dict[str, Any],drop_queue: mp.Queue):
     files = e['dataTransfer']['files']
     if len(files) == 0:
         return
 
     for file in files:
         drop_queue.put(file.get('pywebviewFullPath'))
-    drop_event.set()
 
-def bind_drop(window: webview.Window, drop_event:Event, drop_queue:mp.Queue) -> None:
+def bind_drop(window: webview.Window, drop_queue:mp.Queue) -> None:
     window.dom.document.events.dragenter += DOMEventHandler(on_drag, True, True)
     window.dom.document.events.dragstart += DOMEventHandler(on_drag, True, True)
     window.dom.document.events.dragover += DOMEventHandler(on_drag, True, True)
-    window.dom.document.events.drop += DOMEventHandler(lambda e: on_drop(e, drop_event, drop_queue), True, True)
+    window.dom.document.events.drop += DOMEventHandler(lambda e: on_drop(e, , drop_queue), True, True)
 
 
 def _open_window(
     host: str, port: int, title: str, width: int, height: int, fullscreen: bool, frameless: bool,
-    method_queue: mp.Queue, response_queue: mp.Queue, drop_event: Event, drop_queue: mp.Queue,
+    method_queue: mp.Queue, response_queue: mp.Queue, drop_queue: mp.Queue,
 ) -> None:
     while not helpers.is_port_open(host, port):
         time.sleep(0.1)
@@ -66,7 +65,7 @@ def _open_window(
     closed = Event()
     window.events.closed += closed.set
     _start_window_method_executor(window, method_queue, response_queue, closed)
-    webview.start(func=bind_drop, args=(window, drop_event, drop_queue), storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
+    webview.start(func=bind_drop, args=(window, drop_queue), storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
 
 
 def _start_window_method_executor(window: webview.Window,
@@ -127,28 +126,26 @@ def activate(host: str, port: int, title: str, width: int, height: int, fullscre
             time.sleep(0.1)
         _thread.interrupt_main()
 
-    def check_drop(drop_queue: mp.Queue, drop_event: Event) -> None:
+    def check_drop(drop_queue: mp.Queue) -> None:
         while True:
-            while not drop_event.is_set():
-                time.sleep(0.1)
             got_drop(drop_queue.get())
-            drop_event.clear()
 
     if not optional_features.has('webview'):
         log.error('Native mode is not supported in this configuration.\n'
                   'Please run "pip install pywebview" to use it.')
         sys.exit(1)
 
-    drop_event = mp.Event()
     drop_queue = mp.Queue()
 
+    drop_queue.get()
+
     mp.freeze_support()
-    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue, drop_event, drop_queue
+    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue, drop_queue
     process = mp.Process(target=_open_window, args=args, daemon=True)
     process.start()
 
     Thread(target=check_shutdown, daemon=True).start()
-    Thread(target=check_drop, args=(drop_queue, drop_event), daemon=True).start()
+    Thread(target=check_drop, args=(drop_queue,), daemon=True).start()
 
 
 def find_open_port(start_port: int = 8000, end_port: int = 8999) -> int:
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -135,17 +135,13 @@ def activate(host: str, port: int, title: str, width: int, height: int, fullscre
                   'Please run "pip install pywebview" to use it.')
         sys.exit(1)
 
-    drop_queue = mp.Queue()
-
-    drop_queue.get()
-
     mp.freeze_support()
-    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue, drop_queue
+    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue, native.drop_queue
     process = mp.Process(target=_open_window, args=args, daemon=True)
     process.start()
 
     Thread(target=check_shutdown, daemon=True).start()
-    Thread(target=check_drop, args=(drop_queue,), daemon=True).start()
+    Thread(target=check_drop, args=(native.drop_queue,), daemon=True).start()
 
 
 def find_open_port(start_port: int = 8000, end_port: int = 8999) -> int:
--- a/nicegui/native/native.py
+++ b/nicegui/native/native.py
@@ -8,6 +8,7 @@ from ..logging import log
 
 method_queue: Queue = Queue()
 response_queue: Queue = Queue()
+drop_queue: Queue = Queue()
 
 try:
     with warnings.catch_warnings():


EmberLightVFX:
Got it working!!!
I had to create a new widget to fetch the file drop. The positive with this is that you have have multiple drop-locations.

https://github.com/user-attachments/assets/0601a1e3-07fb-4105-8941-07e56ccdc873

There are two problems still tho that I would love to get some help with.
1. Closing the native window doesn't close the listening thread I'm running
2. When auto-reload is enabled this doesn't work. I think it's because the extra subprocess that gets created and information goes to the wrong process.

I'll post the data tomorrow as it's quite late here 😄

EmberLightVFX:
Added all files to make the drag and drop working and included a _example.py in the root dir for you to try it out.
I have added a new element called `drop_zone`. It's just a wrapper like row or column and the user needs to fill the content of the drop zone.
When a file or folder is being draged over the zone the class `dragover` will get added so the user can change the style of the zone.

Drop zone does still not work with `refresh=True` as `drop_queue.put(file.get('pywebviewFullPath'))` (in the native_mode.py file) gets applied on the refresh-thread and not the thread the user is interacting on. I don't know how to really solve this...
But I got the closing working!

EmberLightVFX:
No worries! I know you're busy with other stuff :) I haven't had much time myself to look into the problems but when I get some time I'll do my best. Any community help is highly appreciated tho!

EmberLightVFX:
I finally got some time to try and tackle this again.
I think I have found a way to not create a separate thread to fetch when the drop-information from webview but currently I don't know how to emit/broadcast the data to any object listening through `dropzone.on("drop", func)` like I did before when the emit came from within dropzone.js.

So quick question. Are there any way to emit/broadcast directly from the python code?

EmberLightVFX:
@falkoschindler
Forgot to tag you in case you wouldn't see my message

falkoschindler:
Hi @EmberLightVFX, thanks for the update!

> Are there any way to emit/broadcast directly from the python code?

I'm not quite sure if I understand the question.
Isn't `window.dom.document.events.drop` conceptually different to `__file-dropped`? One is emitted when dropping _anywhere_ on the document, the other is emitted when dropping onto the drop zone. This means it is impossible to know which `ui.drop_zone` drop handlers to call from within `on_drop` in native_mode.py.

EmberLightVFX:
@falkoschindler
The difference between `window.dom.document.events.drop` and `__file-dropped` is that file-dropped is a javascript event while window.dom... is an event from pywebview and it includes the full path to what ever you just dropped. That's the information I'm trying to catch.

I just pushed an update where I added a asyncio check when the drop zone gets a __file-dropped event and it checks for data in the drop_queue (mp.Queue()).
At the same time window.dom.document.events.drop executes the `on_drop` function in native_mode.py in webview's thread. In that function we set the data for drop_queue that the drop zone now can read and thus we have passed the file/folder information have bin passed.

A bit cumbersome but it's the only way in how NiceGUI and webview works.

So, how the drop zone have the path information but need to send the emit to everything that's listening.
The current way I'm doing it is to now run `self.run_method("drop_emitter", data)` to pass the data to a javascript method and that method simply sends out the emit.
There would just be less failure points if I could send the emit directly from the python code instead of having to go through javascript.

I have also got a new problem. I did the dev on my old environment that was running NiceGUI 1.4.30 but when I updated the codebase to 2.11.1-dev, when I now drop a folder it simply opens the folder in a new explorer window (I'm running Windows).
Also, the on_drop function never gets executed. It seems like something is hogging the window dom, making the drop not reaching webview or something. I tried to compare the codebases but I don't see any direct difference that could be the cause of this. I have no idea what might have changed but everything works on NiceGUI 1.4.30.
Any idea what might be going on?

EmberLightVFX:
Btw, I left some print functions around just to be able to see what executes and when. I'll remove these when everything works as expected 😄

EmberLightVFX:
I have went through the code and fixed everything so it's now working as expected!
The only problem I have left is that the drop zone doesn't work if reload is set to True. There is some thread-comunication problem that would need to be sorted out.
And if it's possible to emit the data directly from python instead of javascript to reduce failure points, but the drop zone do work in its current form.

EmberLightVFX:
So I have bin looking through the code and I don't see a way to comunicate to the correct thread as the reloadis done by uvicorn.
I'm quite stuck. Any idea on how to sort this out?

EmberLightVFX:
From my understanding there isn't really any good way to get drop zone and reload to work at the same time because of all the different threads going on.
So my suggestion would be to simply show a warning that drop zones won't work if reload is enabled when launching the gui.
Would that be an acceptable solution @falkoschindler ?

Except that I think this PR is ready for review. I have still left my example python file but when all is accepted I'll remove it.

EmberLightVFX:
Took some time to change the event usage from `.on("drop_zone", func)` to `ui.drop_zone(on_drop=func)` to match other NiceGUI controls, did some cleanup and added documentation :)

EmberLightVFX:
Yeah sorry for all updates I wrote here. Maybe a big too much.
To answer your notes:

2. Sorry about that. I'll try to fix it. This was caused by me using Ruff for the formatting instead of autopep8. I'm now using autopep8 so it should be ok for future PRs.

3. I'll try to explain what happens when a user drops a file into a drop zone as good as I can.
    1. The user drop a file into a zone.
    2. The zone's javascript-part gets a drop event in the DOM and sends out the event `__file_dropped` (This event only gets what modifier key might have bin hold down when the file gets dropped and the files filename and extension, not the path)
    3. The zone's python-part listens for the `__file_dropped` event and when it gets it it starts to check for when `drop_queue` gets data attached to it.
    All this happens induvidually per drop zone.
    4. While step 2 and 3 happens Webview also gets a drop event when a file gets dropped into its window. This event contains the full path to the file that got dropped. When this happens `drop_queue` will get set with the path to the files that got dropped.
    5. The zone's python-part that is listning for changes to `drop_queue` now gets the file path from webview and will send out the info to all listening event handlers.

4. That's weird.... What OS are you running the test in? I have to see if I can find a way to check what web engine webview is using. It might be something to do with that maybe?

nicegui/native/native_mode.py
```diff
@@ -21,14 +21,32 @@
...
     pass
 
 
+def on_drop(e: dict[str, Any], drop_queue: mp.Queue):
+    files = e['dataTransfer']['files']
+    if len(files) == 0:
+        return
+
+    for file in files:
+        drop_queue.put(file.get('pywebviewFullPath'))
```
falkoschindler, line 36:
```suggestion
    for file in e['dataTransfer']['files']:
        drop_queue.put(file.get('pywebviewFullPath'))
```

The early exit is not needed. If there are no files, the for-loop doesn't do anything.

nicegui/native/native_mode.py
```diff
@@ -47,7 +65,12 @@ def _open_window(
     closed = Event()
     window.events.closed += closed.set
     _start_window_method_executor(window, method_queue, response_queue, closed)
-    webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
+    webview.start(
+        func=bind,
+        args=(window, drop_queue),
+        storage_path=tempfile.mkdtemp(),
+        **core.app.native.start_args,
+    )
```
falkoschindler, line 73:
```suggestion
    webview.start(bind, (window, drop_queue), storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
```

Let's keep it short and concise.

nicegui/native/native_mode.py
```diff
@@ -21,14 +21,32 @@
...
+    for file in files:
+        drop_queue.put(file.get('pywebviewFullPath'))
+
+
+def bind(window: webview.Window, drop_queue: mp.Queue) -> None:
+    window.dom.document.events.drop += DOMEventHandler(
+        lambda e: on_drop(e, drop_queue),
+        True,
+        True,
+    )
```
falkoschindler, line 44:
```suggestion
    window.dom.document.events.drop += DOMEventHandler(lambda e: on_drop(e, drop_queue), True, True)
```

No need for 5 lines.

Oh and mypy is complaining:
`Argument 1 to "__iadd__" of "Event" has incompatible type "DOMEventHandler"; expected "Callable[..., Any]"Mypy[arg-type](https://mypy.readthedocs.io/en/latest/_refs.html#code-arg-type)`

nicegui/server.py
```diff
@@ -27,10 +28,15 @@ def create_singleton(cls, config: CustomServerConfig) -> None:
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
         self.instance = self
         assert isinstance(self.config, CustomServerConfig)
-        if self.config.method_queue is not None and self.config.response_queue is not None:
+        if (
+            self.config.method_queue is not None
+            and self.config.response_queue is not None
+            and self.config.drop_queue is not None
```
falkoschindler, line 34:
```suggestion
            self.config.method_queue is not None and
            self.config.response_queue is not None and
            self.config.drop_queue is not None
```

I prefer this for its symmetry.

nicegui/elements/drop_zone.py
```diff
@@ -0,0 +1,104 @@
...
+        This controller only works in NiceGUI native mode.
+
+        :param on_drop:	callback to execute for each dropped object
+        :param hover_style: change the style that gets applied to the drop zone when an object is hovered
+        :param cleared_hover_style: custom style that gets applied after the object have bin dropped
+        """
+        super().__init__()
+
+        self.default_classes('relative')
+        self.classes('relative')
```
falkoschindler, line 35: Why do we need the "relative" class?

falkoschindler (commented):
Thanks @EmberLightVFX! And sorry for the confusion. I just tested \_example.py once again and it seems to be working. Maybe I dropped onto the second drop zone which isn't connected to an event handler.

Now that I can see your changes more clearly, I started reviewing them in more detail. My two main concerns are the following:

1. The event mechanism sounds rather complicated. I'll look into it more closely and maybe we can find a simplification somewhere. But requiring `native=True` and `reload=False` is quite a limitation. It would be great to fix it for `reload=True`. And ultimately I would hope to support `ui.drop_zone` for `native=False` as well.

2. The amount of code for the "hover style" is quite significant, including additional events just to toggle the Tailwind classes. Would something like
    ```py
    ui.drop_zone().classes('outline-gray hover:outline-dashed')
   ```
    be too much of a simplification? It switches even when hovering without dragging a file, but it simplifies the code sooo much.
    If we were to keep the extra styling parameter(s), I'd try to handle the drag events right in JavaScript and switch classes and/or style there without involving the server. This reduces the need for these two events as well as needless latency.
    Another remark regarding Tailwind classes: There is an option to run NiceGUI without Tailwind. For this scenario built-in elements shouldn't rely on Tailwind but use plain CSS instead.

EmberLightVFX:
I'll go through all your points when I get some time over but let me answer your two notes first:

1. There might be some way to solve that reload can be true. It's a real wrangle to manage all the threads and processes that NiceGUI creates with native and reload 😅
Sadly it's imposible to get the file path in the webbrowser as it's against Javascript security to get direct file paths.

2. I totally agree. I'm currently re-writing that part of the code. Instead of having a `hover_style` and `cleared_hover_style` argument for drop_zone I'm gonna use .classes() and .hover_classes() for the user to set the styles.
So ether the user can write it out like you did with `hover:outline-dashed` in the classes() function or they can use `.hover_classes("outline-dashed")` for the class to only change when you hover a file over the drop zone.

3. I'll change the internal tailwind css with plain css :)

nicegui/elements/drop_zone.js
```diff
@@ -0,0 +1,40 @@
...
+      el.addEventListener(eventName, () => this.$emit("drag_leave", "drag_leave"));
+    });
+
+    function preventDefaults(e) {
+      e.preventDefault();
+      e.stopPropagation();
+    }
+
+    const handleDrop = (e) => {
+      this.$emit("__file-dropped", e);
```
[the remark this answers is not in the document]
EmberLightVFX: I have added the two __ in front just so users won't accidentally use that event emitter, so it's just there to be sure.
But I did see now that I accidentally wrote file-dropped instead of file_dropped. Fixed that!

nicegui/native/native_mode.py
EmberLightVFX: I don't really know so I asked Claude and got this answer:
````
The window.dom.document.events.drop is an event from pywebview that expects to have handlers added to it, but mypy sees it as a standard Python Event (like from the threading module) which has a different expectation for its __iadd__ method.
```

I added `# type: ignore[arg-type]` after the string but it becomes too long for autopep8 so it places the comment on a new line.
I saved the file without formatting but for it to save with auto formatting I think we'll have to split the line up into multiple lines to let `# type: ignore[arg-type]` be on the correct line.
Should I do that?

nicegui/native/native_mode.py
EmberLightVFX: I don't really know so I asked Claude and got this answer:
````
The window.dom.document.events.drop is an event from pywebview that expects to have handlers added to it, but mypy sees it as a standard Python Event (like from the threading module) which has a different expectation for its __iadd__ method.
```

I added `# type: ignore[arg-type]` after the string but it becomes too long for autopep8 so it places the comment on a new line.
I saved the file without formatting but for it to save with auto formatting I think we'll have to split the line up into multiple lines to let `# type: ignore[arg-type]` be on the correct line.
Should I do that?

EmberLightVFX:
I have now updated the code with all your comments.

I have now also added two extra classes to the drop zone element.
`hover_classes` and `hover_overlay_classes`.
With these the user can pick to ether change/update the base class using `hover_classes` for example to change the background color on hover.
With `hover_overlay_classes` the user can access a div that lives above the drop zones content, so with it you can create borders or add a background color with opacity above everything (see _example.py for better understanding).

I think this is the best way to let users get the style they want for the drop zones.
Currently you have to add `relative` to the base class and `absolute` to the overlay class manually for the overlay to cover the drop zone div. Maybe it's best to just always attach these for ease of use for the users?

I'll need to update the docs with all this information but I'll do that when we have come to a conclusion on how everything should be set up.

EmberLightVFX:
1. `Why do we need the new drag counter? It wasn't there before.` It had something to do with how the drop-event was bubbling when having overlays on the drop zone but it seems like it works as expected now any way :S So no need for them!
2. The catch isn't longer needed. It was part of the old code. So hard to fully know what all code is for after all changes 😅
3. I like giving a warning when a file gets dropped so you get some responsse rather than nothing happens. While you develop you also might not want a notification that drop_zones won't work each time a reload happens. Thoughts?
4. A notification like when reload=True can be added. Where can I check if native has bin set to true? It doesn't seem like the native bool is being saved anywhere.

nicegui/elements/drop_zone.py
```diff
@@ -0,0 +1,99 @@
...
+                    modifiers=KeyboardModifiers(
+                        alt=event.args['altKey'],
+                        ctrl=event.args['ctrlKey'],
+                        meta=event.args['metaKey'],
+                        shift=event.args['shiftKey'],
+                    ),
+                )
+                for drop_handler in self._drop_handlers:
+                    handle_event(drop_handler, args)
+                return
```
falkoschindler, line 80: @EmberLightVFX This `return` prevents the while loop from looping. Should we remove it, or replace `while not drop_queue.empty():` with `if not drop_queue.empty():`?

EmberLightVFX:
I had to re-add the drag counter. Without it the drag_enter will bubble over and emit drag_leave directly after making the overlay just flicker. I most have missed reload the ui or save the file when I tried remove it before but I saw the problem now. So I had to re-add it.

Btw, I'm not directly fluid in javascript so if there is a better way to do it, go for it! But I haven't found another way to solve this.

falkoschindler (changes requested):
@EmberLightVFX Sorry for keeping you waiting. Yesterday we discussed this PR once again and came to the following conclusion:

Because the `ui.drop_zone` element doesn't work for non-native apps, we're hesitant to add it to the set of officially supported UI elements. People might want to use it in the browser, which does not work.

Therefore we would like to add it as an example, like the [local file picker](https://github.com/zauberzeug/nicegui/tree/main/examples/local_file_picker). This allows us to merge the code more quickly and close this PR. If we decide to add (limited) support for non-native apps later, we can do it in a separate PR and eventually move it to the core library.

Of course, the example requires certain infrastructure in native.py. But instead of a special `drop_queue`, we would like to introduce a more generic `event_queue`. We can start with emitting drop events and maybe a few more (see https://pywebview.flowrl.com/examples/events.html). If the performance impact is ok, we could simply emit all events. And we need a way for user code to subscribe to such events. This might be the trickiest part.

Long story short: Adding support for pywebview events would be an awesome feature for the NiceGUI library. And a drop zone for native apps would make a great example.

What do you think, @EmberLightVFX?

falkoschindler:
> That sounds like a great idea! So the drop zone addon would simply be moved into the examples folder then?

Exactly.

> To subscribe to pywebviews events, maybe a specific event handler would be needed like I'm currently doing for the drop zone addon? As all events is getting passed through one channel (event_queue) some kind of filtering would be needed.

Yes, we need to find a way to register handlers for individual event types. Sending events as a tuple with `type: str` and `payload: dict` might be a starting point. But I haven't thought about the details yet.

> Do you want me to fix something or would you like to take that part to make sure everything gets setup the way you want it?

If you could start moving things around and see if you can work out the event registration, that would be great. But let me know if/when you need help. Then I can support or take over. 🙂

EmberLightVFX:
I don't know when I'll have some time over to work on it so if your schedule is a bit empty, take a stab!
I'll write here if/when I get some time for this!

EmberLightVFX:
I'm back at it!
While trying to figure out where to place the pywebview event handler I saw that we create
https://github.com/zauberzeug/nicegui/blob/882b314823a7849fef43c6b2a0e5695edad6bb9e/nicegui/server.py#L31 in server.py but when initializing but in native_mode.py we run
https://github.com/zauberzeug/nicegui/blob/882b314823a7849fef43c6b2a0e5695edad6bb9e/nicegui/native/native_mode.py#L46

@falkoschindler  Maybe I'm just blind but why do we create a new window using webview and not WindowProxy? What's the purpuse of WindowProxy if it doesn't get used by the webview window we create in native_mode.py?

falkoschindler:
If I recall correctly, `WindowProxy` serves as a proxy to the "real" window which runs in a separate process. This way the user can interact with `core.app.native.main_window` and the proxy communicates with the window via queues.

EmberLightVFX:
I have now updated everything so there now is an event_queue for the native (pywebview) part.
I also made a custom event handler highly inspired how elements does it.

One thing to note tho is for NiceGUI to fetch the emit that pywebview does I'll have to constantly monitor event_queue for changes. This is because pywebview runs in another process, this is the only way to communicate between them.

Checking task manager I can see that without the event manager the python process runs at 0-1% but with the event manager it runs at ~12% on my laptop!
I wouldn't say that is acceptable really...

The reason why I didn't get the 12% hit in my old code was because I was listening for the drop-zone elements emit and only then I monitored drop_zone_queue (now named event_queue). This is still possible for the drop events as I can make pywebview and nicegui listen for a drop at the same time but nicegui can't listen for when the window moves, minimizes, maximizes etc.

So to remove the 12% hit we'll have to skip the [pywebview events](https://pywebview.flowrl.com/examples/events.html) and only use the [drag and drop dom events](https://pywebview.flowrl.com/examples/drag_drop.html). If anything, maybe it's ok to finalize the drop dom events, merge and later maybe make a PR for the pywebview events.

The code isn't super clean but I wanted to get this out so we could discus the overhead and what to do from here.

(Maybe the master @evnchn can see a way to solve this😏)

evnchn:
Bro I don't even use PyWebView. 😅 It'd take a long while for me to reach any familiarity on that aspect to be able to help...

EmberLightVFX:
@falkoschindler would love to get some input on my last post so we can finish this PR 😄

falkoschindler:
@EmberLightVFX 12% CPU is definitely unacceptable. But before limiting ourselves to a tiny subset, let's see if we can improve the `_event_loop`. I quickly asked ChatGPT and it basically recommends using a blocking `.get()` call. Otherwise this threads is looping constantly which is burning CPU.

Here is the full discussion with some additional hints: https://chatgpt.com/share/6863bc87-4d24-800c-b1ee-1d720a0bdeee

EmberLightVFX:
Changing get_nowait() to get() did no difference for the CPU, it's still at ~12%
I started to read up on Pipes again. I got confused about them last time and didn't use them but for events like this they are perfect so I have no updated the code to use pipe instead of queue and the CPU idles at 0% !

I found a bug in my drop zone example. Each zone saves the data coming from pywebview's drop event and then processes it when it gets the file_drop event from its ui element.
Pywebview will always emit drop to all zones so it's imposible to know for what zone the data is intended, thus we're having two events.
The problem is that on the zones that doesn't get a file_drop event the old data still lingers and when a file gets dropped in its zone the old data will get read.
The best solution I have found is to have the drop data being stored in the event_manager. It's maybe not the best place to store it but I can't really find any better location that makes sense. Thoughts on it?

So in short all zones now check if event_manager.drop_data have any data assigned to it, then reads the data and sets drop_data to None so the old data gets erased.

falkoschindler:
> so it's imposible to know for what zone the data is intended

This reminds me of https://github.com/zauberzeug/nicegui/pull/3378#issuecomment-2663269872:
"This means it is impossible to know which `ui.drop_zone` drop handlers to call from within `on_drop` in native_mode.py."

I guess it's time for another review. I'm buried in PRs right now, but I'll see what I can do.

EmberLightVFX:
Exactly like that comment! Forgot we talked about it 😅

unigrated-solutions:
I've been following this pull request from the beginning. With how fast Nicegui has been progressing, if drag and drop was an easy thing to solve, I know it would have been implemented already. I have much appreciation for anyone spending their time trying to provide solutions like you have. At the end of the day, nicegui is a web based framework which running in native mode has provided me quite a crutch to going back pyqt or tkinter. I've been going through the requested changes for the last 2 weeks or so and the thought I keep going back to is, why is the original file location so important? The data in the file can be easily imported for processing, with the results returned to replace the original at the users consent, but to do anything with the original file on the server side is a serious security concern. I can probably guess why you would want direct access to the file locations, being that its in native mode, but I can't see building it into the application core the best solution. 

Here's a simple example without any changes to the application itself that just copies the dropped file to the server to do whatever with after that:
```python
from nicegui import ui, app
from fastapi import UploadFile, File
import os

UPLOAD_DIR = './uploads'
os.makedirs(UPLOAD_DIR, exist_ok=True)

STYLE = '''

        html, body, #app {
            height: 100%;
            margin: 0;
            padding: 0;
            overflow: hidden; /* Prevent window scroll */
        }

        .main-layout {
            display: flex;
            flex-direction: column;
            height: 100%;
        }

        .scroll-area-fill {
            flex: 1;
            overflow-y: auto;
        }
        
        #drop-zone {
            position: relative;
            border: 2px dotted #3b82f6; /* blue-500 */
            background-color: #f0f9ff;  /* blue-50 */
            transition: background-color 0.2s ease, border-color 0.2s ease;
        }

        #drop-zone.drag-over {
            background-color: #dbeafe;  /* blue-100 */
            border-color: #1e40af;      /* blue-900 */
        }
    '''
    
@app.get('/upload-status')
def status():
    return {"status": "ready"}

@app.post('/upload')
async def upload_file(file: UploadFile = File(...)):
    save_path = os.path.join(UPLOAD_DIR, file.filename)
    with open(save_path, 'wb') as f:
        content = await file.read()
        f.write(content)
    return {'status': 'success', 'filename': file.filename}

@ui.page('/')
def main():
    ui.add_head_html('''
    <script>

    window.addEventListener('dragover', e => e.preventDefault(), true);
    window.addEventListener('drop', e => e.preventDefault(), true);
    
    document.addEventListener('DOMContentLoaded', () => {
        const dropZone = document.getElementById('drop-zone');

        dropZone.addEventListener('dragover', (event) => {
            event.preventDefault();
            event.stopPropagation();
            dropZone.classList.add('drag-over');
        });

        dropZone.addEventListener('dragleave', (event) => {
            dropZone.classList.remove('drag-over');
        });

        dropZone.addEventListener('drop', (event) => {
            event.preventDefault();
            event.stopPropagation();
            dropZone.classList.remove('drag-over');

            const file = event.dataTransfer.files[0];
            if (file) {
                const formData = new FormData();
                formData.append('file', file);

                fetch('/upload', {
                    method: 'POST',
                    body: formData
                }).then(response => response.json())
                .then(data => {
                    if (dropZone && dropZone.__vue__) {
                        dropZone.$emit('drop', data);
                    }
                });
            }
        });
    });
    </script>
    ''')

    ui.add_css(STYLE)

    @ui.refreshable
    def file_list():
        def get_uploaded_files():
            return sorted(os.listdir(UPLOAD_DIR))

        with ui.column().classes('w-full p-4'):
            files = get_uploaded_files()
            if files:
                for file in files:
                    ui.label(file).classes('text-blue-800').style('font-weight: 500;')
            else:
                ui.label("No uploaded files yet.").classes('text-gray-500')

    with ui.column().classes('w-full h-screen gap-4'):
        with ui.card().props('id=drop-zone').classes('w-full h-[300px] relative'):
            ui.label("Drop a file on this box") \
                .classes('text-black text-center w-full') \
                .style('font-weight: bold; font-size: 1.2rem; position: absolute; top: 50%; left: 50%; transform: translate(-50%, -50%);')

        with ui.card().props('draggable').classes('bg-blue-200'):
            ui.label('Another card that should be drag and droppable')

        with ui.scroll_area().classes('w-full flex-grow overflow-y-auto border border-gray-300 rounded'):

            file_list()

        with ui.row().classes('w-full justify-end p-2'):
            ui.button(icon='refresh', on_click=file_list.refresh)

ui.run(
    native=True,
    reload=False,
)
```

EmberLightVFX:
To answer your question on why I need/made a feature like this. I made a software that scans folders and files for broken image sequences and video files. It also has a watch-folder function to keep on checking a folder for changes.

These files are BIG. Tens of gigabyes, so uploading the data, even if it's locally, is not ideal at all.

In the end I think this PR ended up in a good place where we just fetch PyWebView's emits and the drop-zone code is just an example what you can make with those emits.

unigrated-solutions:
So was the drag and drop functionality to provide a more user friendly way to add which directories/ files for your application to watch? If so, I full understand that. But by moving those events to the pywebview thread on startup, it removed the majority of the functionality of drag events listening from all other elements. Although not currently working in native mode, it doesn't seem intuitive for every element to have a listener for whether its in native mode or not. It seems the handling should be done within the page rather than a custom made element that knows how to handle it. I stripped down you're example to almost nothing and it still worked. You have a much better understanding than I do but maybe there's a simpler path to moving all drag events to another thread

EmberLightVFX:
I don't fully understand your comment but the way it works is:

Pywebview have it's own events like moving window, resizeing and the full path to a file when you drop one or more files into the window. It also runs in it's own thread.

In my dropzone example each zone have a drop-listener that listens for when something gets dropped into it.

When you drop files or folders that dropzone will fetch the file/folde path information from pywebviews thread. It's the only way to get it.

After that the dropzone will process the path and continue on.

unigrated-solutions:
Yes but moving the handling of drag events and stopping the propagation of these events prevents them from reach other elements that aren't just looking for external file drops. Like my other card with the draggable property. Theres other ways to capture the pywebview thread from the main thread
https://github.com/zauberzeug/nicegui/discussions/980#discussioncomment-13838081

EmberLightVFX:
Could you share a working example on drag-and-drop like you described? When I did all RnD I couldn't get any other way to work than what I'm currently doing.

unigrated-solutions:
Is there  a way to continue this discussion on your forked repository rather than on the man pull request? You've done alot of work and I don't want to drown it out on the PR discussion.

EmberLightVFX:
Absolutely! Just create an issue over on my fork and we'll continue there

EmberLightVFX:
I haven't heard anything from [unigrated-solutions](https://github.com/unigrated-solutions) and I don't see it myself on how to do it any different so I say go on with the review of this PR.

If there is a better way to do it we can always update it later. It won't change how the end user uses it.

EmberLightVFX:
@falkoschindler Updated the code to work with NiceGUI 3 and it's ready to be reviewed when ever you have time :)

falkoschindler:
Hi @EmberLightVFX, first of all — I'm sorry for the long silence. You put real effort into this and deserved a timely response. I appreciate your patience and the work you invested here.

Your PR was the starting point for thinking about how to get native events back to the main process. In the meantime it has accumulated quite a few conflicts with `main`, so rather than asking you to rebase after all this time, I've opened #5866 with a fresh implementation. It picks up your `Pipe`-over-`Queue` idea to avoid idle CPU usage, and extends the approach to window lifecycle events (`shown`, `minimized`, `maximized`, `resized`, `moved`, `closed`, etc.) in addition to file drops — with a simple `app.native.on(event, handler)` API.

There's also been a separate request in #5852 for handling the `closing` event (e.g. "you have unsaved changes" dialogs). That one needs a synchronous round-trip to veto the close, so it's out of scope for now but on the radar.

Per-element drop zones (knowing *which* UI element a file was dropped on) are not yet supported but something we'd like to add in a follow-up.

I'm going to close this PR in favor of #5866. Thank you again for getting the ball rolling on this!

evnchn:
Two things I want to add, @EmberLightVFX:

1. Even if you do nothing, we will credit you as a contributor in the release notes. 
2. Should you decide to further contribute, while you can't push to #5866, you can let us know and we should be able to pull them in manually.

PR #3383, opened 2024-07-22
Make combining multiple content filters more robust

rodja:


falkoschindler (approved):
I'm approving this PR mostly for the additional test and because it will work in 99.999% rather than only 99.99% of all cases. But maybe we'll find an even better way to find elements in the DOM before merging into main.

PR #3385, opened 2024-07-22
Improve iteration of elements

falkoschindler:
This PR tries to improve the iteration of elements, mainly to increase code readability. It makes the existing `_collect_descendants` method public, changes it to return an iterator `descendants` and adds a very similar `ancestors`.

While working on the implementation, some details turned out to be unclear. E.g. what is the expected result if one of multiple markers matches? What if one of multiple ancestors fulfils none, any or all criteria? I guess the individual conditions need thorough review and maybe more test cases. And we should add more information about the final rules to the docstring.

The new iterator might be less efficient than before, because the element iteration is agnostic of the filter conditions and, thus, yields every element even if a subbranch could be skipped. But I'd prefer a readable and correct implementation to finialize the filter rules before optimizing performance.

Open tasks:

- [x] review conditions
- [x] fix failing tests
- [x] find simpler expressions for some ancestor conditions based on sets and/or itertools
- [x] does `ElementFilter._contents` need to be a list?

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import inspect
-import itertools
 import time
 import uuid
 from contextlib import contextmanager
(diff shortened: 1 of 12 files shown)


falkoschindler:
I finally finished the implementation and fixed all tests. Let's review this alternative solution together, @rodja.

rodja (approved):
Very nice!

PR #3395, opened 2024-07-24
Adding methods to get computed props

srobertson86:
Adding methods to get computed props from Quasar elements, using the QTable as a proof of principle

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -484,6 +484,19 @@ class Element(Visibility):
         return self.client.run_javascript(f'return runMethod({self.id}, "{name}", {json.dumps(args)})',
                                           timeout=timeout, check_interval=check_interval)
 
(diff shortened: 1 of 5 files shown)


falkoschindler (approved):
Thanks for the pull request, @srobertson86!

I just reviewed your code and made some minor changes. The one thing I wasn't so sure about were these async properties. Even though they are technically possible in Python, they kind of contradict the convention of properties being used for quick, cheap attribute access, while methods are for more costly acquisitions. In this case we need to await a websocket request, which can take significant time. And we don't want the users to accidentally access the async property repeatedly in a for-loop, but do it once before iterating over the resulting list. Long story short, I think it is better to use async methods, like we do, e.g., for `ui.aggrid.get_selected_rows`.

PR #3411, opened 2024-07-28
pytest configuration & fixtures

codingpaula:
I tested using the NiceGUI pytest plugin in a separate project with NiceGUI as a dependency, basically the `pytests` example. It worked quite well and all the defined fixtures are available through the pytest plugin.
To not "block" the generic terms`user` and `screen` as fixture names globally with the plugin, we could use `nicegui_user` and `nicegui_screen`. When using these in our tests, a simple renaming fixture does the trick to keep the simpler `user` and `screen` names in our tests.

- [x] also add `nicegui_` prefix to `create_user`, `capabilites` and `prepare_simulated_auto_index_client` fixtures? -> no, we don't ship the pytest plugin with nicegui by default through an entry point

## Changes:
- fixtures:
  - ~~rename `screen` to `nicegui_screen` and `user` to `nicegui_user` to not block generic names for fixtures~~
  - use pytest plugin fixtures in nicegui tests -> `pytest_plugins` instead of `from nicegui.testing.fixtures import *`
  - ~~add renaming fixtures~~
- `pytests` example:
  - add `pytest.ini` with asyncio setting
  - add `nicegui` to requirements.txt
  - add `conftest.py` to show ~~renaming with custom fixture~~ how to use pytest plugin
- ~~other examples: fix unfound `screen` and `user` fixtures~~


falkoschindler (approved):

rodja:
@codingpaula I changed the structure of pytests example to match the modular project description of #3413. What do you think?

codingpaula:
> @codingpaula I changed the structure of pytests example to match the modular project description of #3413. What do you think?

Looks good to me!

rodja (approved):
Ok, @falkoschindler you can merge.

falkoschindler (approved):

PR #3413, opened 2024-07-29
Add documentation of pytest fixtures and pytest configuration

rodja:
This PR introduces a new testing section to the docs with three topics: pytest project setup, user fixture and screen fixture.

**ToDos**

- [x] add an example for the screen fixture
- [x] document required `pytest_plugin` loading
- [x] decide if we use `user` or `nicegui_user` in the docs -> we will use `user` in combination with explicit loading via `pytest_plugin = ...`
- [x] improve overall layout and appearance (code indentation, window sizes, ...)
- [x] is there a nicer way than introducing `@docs.pytest`? -> yes, the new `@docs.part`
- [x] proof-read
- [x] @rodja check simplification of `doc.part`
- [x] does the overwriting fixture need to `yield` or `yield from`? Or is `return` ok?
- [x] fix test_storage
- [x] add info about `addopts = "--driver Chrome"` in `pytest.ini` for screen tests


rodja:
@falkoschindler I wrote the docs as if the changes from #3411 where already applied. I think you can proof-read now.

falkoschindler (approved):
Ok, looks good to me.
I think I could simplify the `doc.part` decorator a bit.

rodja:
I like your simplifcation @falkoschindler. I also tested with `return` instead of `yield`: it works fine. So `yield` is only needed if there is something to tear-down in the fixture. The user fixtures tear down is executed regardless of using `yield` or `return`.

falkoschindler (approved):

PR #3439, opened 2024-08-05
Remove icecream dependency

falkoschindler:
We decided that a library like NiceGUI shouldn't enforce a dependency to a debugging tool like [icecream](https://github.com/gruns/icecream), even if it is very handy. We leave it up to the user code to add it.


codingpaula (approved):
I scream, you scream, we all scream for ice cream!
No more ice cream 😞 🍨

PR #3443, opened 2024-08-05
Support raising to override messages of ValidationElement

pierrecdn:
This change aims to delegate the control of the error message to the validation functions.
This change preserve backward compatibility, while allowing to share behavior between pydantic or FastAPI endpoints [1] and NiceGUI.

[1] https://docs.pydantic.dev/2.8/concepts/validators/

### DIFF
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -41,8 +41,12 @@ class ValidationElement(ValueElement):
             return self.error is None
 
         for message, check in self.validation.items():
-            if not check(self.value):
-                self.error = message
+            try:
+                if not check(self.value):
+                    self.error = message
+                    return False
+            except (ValueError, AssertionError) as error:
+                self.error = str(error)
                 return False
 
         self.error = None
--- a/website/documentation/content/input_documentation.py
+++ b/website/documentation/content/input_documentation.py
@@ -46,11 +46,21 @@ def styling():
     You can validate the input in two ways:
 
     - by passing a callable that returns an error message or `None`, or
-    - by passing a dictionary that maps error messages to callables that return `True` if the input is valid.
+    - by passing a dictionary that maps error messages to callables that return `True` if the input is valid, or raise
+    ValueError or AssertionError to take control on the message to display.
 ''')
 def validation():
     ui.input('Name', validation=lambda value: 'Too short' if len(value) < 5 else None)
-    ui.input('Name', validation={'Too short': lambda value: len(value) >= 5})
+
+    def runtime_message(s: str):
+        if len(s) > 10:
+            raise ValueError('Too long')
+        return True
+
+    ui.input('Name', validation={
+        'Too short': lambda value: len(value) >= 5,
+        'Error message delegated to function': runtime_message
+    })
 
 
 doc.reference(ui.input)


falkoschindler:
Thanks for the pull request, @pierrecdn!

I just wonder if it's really a good idea to propagate exceptions to the UI. Exceptions are usually meant for the developer, while the error message is shown to the user. The validation function could, of course, make sure to raise an exception with a message for the user. But then it could also simply return the message itself as a string.

So maybe you can elaborate on how this helps to "share behavior between pydantic or FastAPI endpoints and NiceGUI". Can you give a rough example? Thanks!

pierrecdn:
Hi,
Sure it maybe missed a bit of context (but now you'll have a lot :sweat_smile:).

Overview: I'm trying to unify API and UI endpoints (think CRUD form auto-generated from API descriptors). The issue is that validation/error management in pydantic validators is performed using exceptions, rather than returning an error string.

Also I cannot wrap this into a "try_except_to_string_return" function, because it doesn't match their spec:
From https://docs.pydantic.dev/latest/concepts/validators/#handling-errors-in-validators:
```
validators should either return the parsed value or raise a ValueError or AssertionError (assert statements may be used).
```

Trying to illustrate my point with the simplest possible example:
* I define two FastAPI pages, one for the API endpoint, one for the UI:

```
app = FastAPI()

@app.post("/api/my_request")
def create_myrequest_api(
    req: MyRequest
):
    return {"status": f"Got your {req.fqdn}{req.path} request, thanks!"}

@ui.page("/ui/my_request")
async def create_myrequest_ui() -> None:
    model_to_ui_form(MyRequest, read_only=False)

def main() -> None:
    ui.run_with(app)
    uvicorn.run(app)

if __name__ == "__main__":
    main()
```

* They both make use of the `MyRequest` pydantic basemodel, for which a field is using a custom validator:

```
URL_PATH_VALIDATOR: dict[str, Callable] = {
    "Start with slash": lambda s: s[0] == "/",
    "No dots": lambda s: "." not in s,
    "Max. length": lambda s: len(s) <= 2048,
    "Unauthorized characters": lambda s: re.match(r"\A[\/\.\-\w]+\Z", s) is not None,
}

def url_path_validator(path: str) -> str:
    for description, validator in URL_PATH_VALIDATOR.items():
        if not validator(path):
            raise ValueError(description)
    return path

UrlPath = Annotated[str, AfterValidator(url_path_validator)]

class MyRequest(BaseModel):
    fqdn: str = Field(title="The FQDN", examples=["foo-bar.tld"])
    path: UrlPath = Field(title="The path",  examples=["/foo/bar"])
```

I can now call the API endpoint this way:
```
# Nominal case
$ curl -XPOST http://localhost:8000/api/request -H'content-type: application/json' -d'{"fqdn": "test.tld", "path": "/valid"}'  # issuing a HTTP/200
{"status":"Got your test.tld/valid request, thanks!"}

# Error case
$ curl -XPOST http://localhost:8000/api/request -H'content-type: application/json' -d'{"fqdn": "test.tld", "path": "invalid"}' # issuing a HTTP/422
{"detail":[{"type":"value_error","loc":["body","path"],"msg":"Value error, Start with slash","input":"invalid","ctx":{"error":{}},"url":"https://errors.pydantic.dev/2.8/v/value_error"}]}
```

I made it simple for the example, but the goal is obviously to do that when the operation to perform is complex enough to not fit into a [classic type annotation/field](https://docs.pydantic.dev/latest/concepts/fields/) OR when you want to perform N checks, ordered in sequence, to better guide the user.

The interesting part not shown for now is the `model_to_ui_form` function. TL;DR that's a ton of generic stuff, and at some point it does this to render a `str` subclass:
```
(...)
    # str to ui.input (https://nicegui.io/documentation/input)
    elif issubclass(expected_type, str):
        validation = {}
        for meta in validation_metadata:
            match meta:
                case MaxLen():
                    max_length = getattr(meta, "max_length")
                    validation[f"Max. length is {max_length}"] = lambda s, max_len=max_length: len(s) <= max_len
                case MinLen():
                    min_length = getattr(meta, "min_length")
                    validation[f"Min. length is {min_length}"] = lambda s, min_len=min_length: len(s) >= min_len
                case AfterValidator():
                    predicate = getattr(meta, "func")
                    validation[f"Respect '{predicate.__name__}'"] = lambda s: predicate(s)
        elem = ui.input(
            value=value,
            on_change=lambda e: write_callback(e.value),
            validation=validation,
        )
```

In the end, when patched, it produces a nice input form, with custom validation logic:
image

Currently without such ability it's very vague for user, I typically display "Respect 'url_path_validator'", which isn't that informative :smile:.

>  I just wonder if it's really a good idea to propagate exceptions to the UI. Exceptions are usually meant for the developer, while the error message is shown to the user. 

Actually, the goal was not to propagate the exception in any way, just be able to catch it to delegate the message display to third party code using exceptions in a rather standard way. The exception itself is serialized using `str(e)` and propagated to the UI using the `self.error` property, as before.

pierrecdn:
> I'm still struggling to understand how AfterValidator is ending up as a validator in the UI. I guess the magic happens somewhere in the match statement. But this code is far from a usual use case, I guess.

The match statement essentially analyze the BaseModel  [FieldInfo](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.FieldInfo) object to extract its metadata properties, including validators.

TL;DR: my custom code deal with [BaseModel](https://docs.pydantic.dev/latest/api/base_model/) objects and does the magic with the UI to generate nice forms with as much validators as possible.

Think this code:
```
class TestObject(BaseModel):
    description_string: str = Field(title="The string", description="Free text", min_length=10)
    an_integer: int = Field(title="The integer", description="An int between 10 and 100", ge=10, le=100)
    boolean_val: bool = Field(title="A boolean", default=True, description="Whether foo or bar")


app = FastAPI()


@ui.page("/")
async def create_request_ui() -> None:
    model_to_ui_form(TestObject.model_construct(), read_only=False)
```

Generating this form:

image

Now the thing becomes more complex when my `TestObject` is using a user-specific validation logic, such as:

```
def url_path_validator(path: str) -> str:
    if path == "/not-this":
        raise ValueError("Please don't do that!")
    return path

UrlPath = Annotated[str, AfterValidator(url_path_validator)]

class TestObject(BaseModel):
    (...)
    path: UrlPath = Field(title="The path")
```

The FieldInfo metadata will reference the [AfterValidator](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.AfterValidator) `func` property. My issue here is to return something informative in the UI, knowing that it should also be shared with FastAPI.

> Could you, please, show a minimum reproducible example of some model deriving from BaseModel with a simple validator used to validate user input in a UI? As far as I understand, such an example doesn't need FastAPI endpoints and ui.run_with, but only a model and a UI element. Am I correct?

Indeed, you are.

Doing my best to provide you with something minimal, only handling non-nested BaseModel objects containing strings:

```
from annotated_types import MaxLen, MinLen
from functools import partial
from typing import Annotated

from nicegui import ui
from pydantic import AfterValidator, BaseModel, Field


def model_to_ui_form(obj: BaseModel) -> None:
    with ui.grid(columns=2).classes("w-full"):
        for key, model_field in obj.model_fields.items():
            ui.label(f"{model_field.title}: ")
            write_callback = partial(setattr, obj, key)
            value = getattr(obj, key)

            # str to ui.input (https://nicegui.io/documentation/input)
            validation = {}
            for meta in model_field.metadata:
                match meta:
                    case MaxLen():
                        max_length = getattr(meta, "max_length")
                        validation[f"Max. length is {max_length}"] = lambda s, max_len=max_length: len(s) <= max_len
                    case MinLen():
                        min_length = getattr(meta, "min_length")
                        validation[f"Min. length is {min_length}"] = lambda s, min_len=min_length: len(s) >= min_len
                    case AfterValidator():
                        predicate = getattr(meta, "func")
                        validation[f"Respect '{predicate.__name__}'"] = lambda s: predicate(s)
            elem = ui.input(
                value=value,
                on_change=lambda e: write_callback(e.value),
                validation=validation,
            ).props("dense filled").classes("w-full")


def url_path_validator(path: str) -> str:
    if path == "/not-this":
        raise ValueError("Please don't do that!")
    return path


UrlPath = Annotated[str, AfterValidator(url_path_validator)]


class TestObject(BaseModel):
    path: UrlPath = Field(title="The path", default="test", min_length=5)


model_to_ui_form(TestObject.model_construct())

ui.run()
```

With my patch, the `/not-this` input displays the "Please don't do that!" user validation failure message.

falkoschindler:
Ok, I start to understand. Since pydantic works with validator functions that raise exceptions, you would like to use them in NiceGUI's validation dictionary.

Your code basically boils down to this:
```py
from typing import Annotated, Any, Callable, Dict
import annotated_types
from pydantic import AfterValidator, BaseModel, Field
from nicegui import ui

def url_path_validator(path: str) -> str:
    assert path != "/not-this", "Please don't do that!"
    return path

UrlPath = Annotated[str, AfterValidator(url_path_validator)]

class TestObject(BaseModel):
    path: UrlPath = Field(title="The path", default="/test", min_length=5, max_length=10)

obj = TestObject.model_construct()

for key, model_field in obj.model_fields.items():
    validation: Dict[str, Callable[..., Any]] = {}
    for meta in model_field.metadata:
        if isinstance(meta, annotated_types.MaxLen):
            validation[f"Maximum length is {meta.max_length}"] = lambda s, meta=meta: len(s) <= meta.max_length
        elif isinstance(meta, annotated_types.MinLen):
            validation[f"Minimum length is {meta.min_length}"] = lambda s, meta=meta: len(s) >= meta.min_length
        elif isinstance(meta, AfterValidator):
            validation[f"Respect '{meta.func.__name__}'"] = meta.func
    ui.input(model_field.title, value=getattr(obj, key), on_change=lambda e, key=key: setattr(obj, key, e.value),
             validation=validation)

ui.run()
```

But without this PR you can also use `ui.input` with a validation function `check` like this:
```py
from typing import Annotated, Any, Callable, Dict, Optional
import annotated_types
from pydantic import AfterValidator, BaseModel, Field
from nicegui import ui

def url_path_validator(path: str) -> str:
    assert path != "/not-this", "Please don't do that!"
    return path

UrlPath = Annotated[str, AfterValidator(url_path_validator)]

class TestObject(BaseModel):
    path: UrlPath = Field(title="The path", default="/test", min_length=5, max_length=10)

obj = TestObject.model_construct()

for key, model_field in obj.model_fields.items():
    def check(value: str, metadata=model_field.metadata) -> Optional[str]:
        for meta in metadata:
            if isinstance(meta, annotated_types.MaxLen) and len(value) > meta.max_length:
                return f"Maximum length is {meta.max_length}"
            elif isinstance(meta, annotated_types.MinLen) and len(value) < meta.min_length:
                return f"Minimum length is {meta.min_length}"
            elif isinstance(meta, AfterValidator):
                try:
                    meta.func(value)
                except AssertionError as e:
                    return str(e)
        return None
    ui.input(model_field.title, value=getattr(obj, key), on_change=lambda e, key=key: setattr(obj, key, e.value),
             validation=check)

ui.run()
```

This code isn't much longer, yields the same result and doesn't require a breaking change in NiceGUI's validation elements. What do you think?

pierrecdn:
Sorry for the delay, vacationed.

It seems to do the trick indeed. This being said, I found it way easier to have a native way of doing this (and BTW integrate it with fastapi/pydantic).

> require a breaking change in NiceGUI's validation elements

By breaking, you meant changing the behavior in case the validation function currently raise ValueError or AssertException? Except from that, it seemed to be a noop to me.

> What do you think?

I'll try to perform such changes on my side and will tell you. In between let's probably close.

falkoschindler:
> By breaking, you meant changing the behavior in case the validation function currently raise ValueError or AssertException? Except from that, it seemed to be a noop to me.

Exactly.

> I'll try to perform such changes on my side and will tell you. In between let's probably close.

Sounds good. 👍

PR #2301, opened 2024-01-06
Avoid shadows and borders of nested cards being removed

falkoschindler:
This PR tries to avoid the confusion discussed in #726, #1295, #2265, where child elements of `ui.card` loose shadow and borders. This happens due to Quasar's CSS definitions, which are optimized for "tight" cards. NiceGUI adds some padding and gap by default, so shadow and borders seem to be missing for no obvious reason.

Unfortunately we can't undo Quasar's CSS setting. But I found a way to make them ineffective by inserting another div layer, catching Quasar's CSS setting, while the actual card content isn't affected.

I tested with the following scenarios:
```py
with ui.card().classes('bg-blue-100'):
    with ui.card().classes('bg-green-100'):
        ui.label('Hello world!')

with ui.card().classes('no-shadow border-[1px]'):
    ui.label('See, no shadow!')

with ui.card().tight():
    ui.image('https://picsum.photos/id/684/640/360')
    with ui.card_section():
        ui.label('Lorem ipsum dolor sit amet, consectetur adipiscing elit, ...')

with ui.card():
    ui.table(
        [{'name': 'age', 'label': 'Age', 'field': 'age'}],
        [{'age': '16'}, {'age': '18'}, {'age': '21'}],
    ).props('flat bordered')
```
Screenshot 2024-01-06 at 14 37 20

Only when working with "tight" cards, the result is suboptimal. But I guess it's more intuitive to remove unwanted borders in this case than to wonder why they magically disappear.

```py
with ui.card().tight():
    with ui.card().classes('border'):
        ui.label('Hello world!')
    with ui.card().classes('border'):
        ui.label('Hello world!')
```

Screenshot 2024-01-06 at 14 37 37

Issue #2265, opened 2023-12-30
Nesting cards merges the elements


ghbm-itk:
### Description

When placing cards inside other cards they merge insted of nesting, which seems very unintentional.

Somehow this also affects toggles. I guess they're based around cards? Or maybe it's the drop shadow that's weird.

If I use intermediate containers like a row the cards don't merge.

```python
from nicegui import ui

ui.toggle(["One", "Two"])

with ui.card():
    ui.toggle(["One", "Two"])

with ui.card():
    with ui.row():
        ui.toggle(["One", "Two"])

with ui.card():
    ui.label("Hello")

with ui.card():
    with ui.card():
        ui.label("Hello")

with ui.card():
    with ui.row():
        with ui.card():
            ui.label("Hello")

ui.run()
```

![2023-12-30 16_41_51-NiceGUI og 3 flere sider - Arbejde - Microsoft​ Edge](https://github.com/zauberzeug/nicegui/assets/123645708/55a302eb-10c6-40ae-ad30-e16d8315f4e1)

It doesn't seems to be a problem in Quasar, but the q-card-sections need to be there for the cards to nest properly:

https://codepen.io/Hermasetas/pen/oNVXKGB

falkoschindler:
Hi @ghbm-itk, this is an unfortunate consequence of how NiceGUI adds default padding to `ui.card`, which Quasar doesn't. It has been discussed in #726 and https://github.com/zauberzeug/nicegui/discussions/1295#discussioncomment-6588788 and is documented [here](https://nicegui.io/documentation/card#the_issue_with_nested_borders).

ghbm-itk:
Since the problem is so easily fixed by simply wrapping the element in another element, it seems like something should be possible to fix. I don't understand the backend of NiceGui well enough yet, but I will look into it at some point.

Could a simple hotfix be something like: "If parent is a card wrap this element." ?

falkoschindler:
@ghbm-itk I don't think it's a good idea to automatically change the element hierarchy, because it would break certain assumptions (see #2040 for a discussion about a similar case).

I tried once again to undo Quasar's CSS magic - without success. But I found a way to wrap the card content in a separate div, so that it isn't affected by the Quasar's CSS; see PR #2301. What do you think?

ghbm-itk:
[...]

Issue #726, opened 2023-04-05
Element borders are removed when placed inside a card


miqsoft:
### Description

Hey,

I just figured out that the bordered property does not work when table is inside a card.
I sadly could not figure out the reason for that behaviour.

Here is a small example showing the issue:

``` python

from nicegui import ui

columns = [
    {'name': 'name', 'label': 'Name', 'field': 'name', 'required': True, 'align': 'left'},
    {'name': 'age', 'label': 'Age', 'field': 'age', 'sortable': True},
]
rows = [
    {'name': 'Alice', 'age': 18},
    {'name': 'Bob', 'age': 21},
    {'name': 'Carol'},
]

# here the border is present
ui.table(columns,rows).props('flat bordered separator=cell')

# here the border is missing
with ui.card().classes('w-full'):
    ui.table(columns,rows).props('flat bordered separator=cell')

```

falkoschindler:
Thanks for reporting this bug, @miqsoft! It seems like `ui.card` messes with borders of child elements:

```py
with ui.card():
    ui.label('A')
with ui.card():
    with ui.card():  # <-- this card has no visible border
        ui.label('B')
```

This bug sounds familiar. A few months ago we had very similar problems with `ui.card`. Maybe I'll find the solution in the commit history.

falkoschindler:
Apparently this is "normal" behavior in Quasar. Using the UMD example from https://quasar.dev/start/umd and nesting two cards, the inner card has not shadow:

```py
<div id="q-app">
    <q-card style="padding: 1em; background-color: #ffdddd">
    <q-card style="padding: 1em; background-color: #ddddff">Hello World!</q-card>
    </q-card>
</div>
```

Here are the rules from quasar.css that disable shadow and border in some cases:

```css
.q-card>div:first-child,.q-card>img:first-child {
    border-top: 0;
    border-top-left-radius: inherit;
    border-top-right-radius: inherit
}
.q-card>div:last-child,.q-card>img:last-child {
    border-bottom: 0;
    border-bottom-left-radius: inherit;
    border-bottom-right-radius: inherit
}
.q-card>div:not(:first-child),.q-card>img:not(:first-child) {
    border-top-left-radius: 0;
    border-top-right-radius: 0
}
.q-card>div:not(:last-child),.q-card>img:not(:last-child) {
    border-bottom-left-radius: 0;
    border-bottom-right-radius: 0
}
.q-card>div {
    border-left: 0;
    border-right: 0;
[...]

PR #3444, opened 2024-08-06
Fix Quasar's layout rules for `ui.card` that remove children's borders and shadows

falkoschindler:
Quasar's QCard is the base for `ui.card`. But in the early days of NiceGUI we decided to set a default gap and padding, so that you don't need to wrap content in card sections. If you do need Quasar's original tight layout, you can call `.tight()`.

But this decision caused quite some trouble, because Quasar keeps removing borders and shadows of child elements, which makes sense for the tight layout without padding, but not with NiceGUI's default card style (#726, #1295, #2265, #2301). We already planned to introduce a new `ui.box` element to avoid confusion between both behaviors, which would have caused a _major_ breaking change.

Now I finally found a rather neat solution to change Quasar's CSS rules to apply to tight cards only. In my point of view, this is exactly what we needed. I just didn't think of the possibility to manipulate existing rules using JavaScript, which turned out to be pretty easy and powerful.

With this change, the following cards behave as expected:
```py
with ui.card():
    ui.label('Borders and shadows').classes('border shadow')
    ui.label('of child divs').classes('border shadow')
    ui.label('are preserved').classes('border shadow')

with ui.card().tight():
    ui.label('Borders and shadows').classes('border shadow')
    ui.label('of child divs').classes('border shadow')
    ui.label('are removed').classes('border shadow')
```

Screenshot 2024-08-06 at 10 16 03

Old behavior for reference:

Screenshot 2024-08-06 at 10 16 57

Open tasks:

- [x] add pytest to avoid regression when updating Quasar
- [x] update documentation

### DIFF
--- a/nicegui/elements/card.py
+++ b/nicegui/elements/card.py
@@ -16,10 +16,9 @@ class Card(Element):
         It provides a container with a dropped shadow.
 
         Note:
-        There are subtle differences between the Quasar component and this element.
-        In contrast to this element, the original QCard has no padding by default and hides outer borders of nested elements.
+        In contrast to this element,
+        the original QCard has no padding by default and hides outer borders and shadows of nested elements.
         If you want the original behavior, use the `tight` method.
-        If you want the padding and borders for nested children, move the children into another container.
 
         :param align_items: alignment of the items in the card (default: `None`)
         """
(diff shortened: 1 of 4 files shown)


rodja (approved):
That's a wonderful solution :-)

PR #3452, opened 2024-08-07
Move "is_initialized" flag for 3D scene from server to client

### LINKED ISSUE #3339: 3D objects get stuck in scene when reloading.
### Description

To reproduce, run this example (adapted from https://github.com/zauberzeug/nicegui/discussions/1269#discussioncomment-6573818).  Reload repeatedly and you'll observe red spheres are often stranded in the scene.
```python
import numpy as np
from nicegui import ui

scene = ui.scene()

def update_points(points: np.ndarray):
    for obj in list(scene.objects.values()):
        if obj.name == 'point':
            obj.delete()

    with scene:
        for point in points:
            scene.sphere(0.05).material('red').move(*point).with_name('point')

ui.timer(.1, lambda: update_points(np.random.randn(10, 3)))
ui.run()
```
I think this is likely the same issue as in #659. While it was solved for `ui.page`, it still remains an issue for the auto-index page. If I'm right about the cause, I think it could easily be fixed by moving the `is_initialized` flag from that fix to the JavaScript side. I'll submit a PR once I'm done with the current one I'm working on.

afullerx:
This PR resolves #3339 by moving the existing `is_initialized` flag from  `scene.py` into  `scene.js`. This way, each client maintains its own copy. 

I also removed `is_initialized` from `scene_view.py`. It looks like it was copied over from `scene.py`, but didn't serve any purpose since there are no methods for creating objects.


afullerx:
I considered this, but decided it was OK since most of the methods begin with `if (!this.objects.has(object_id)) return`. So, they are effectively prevented from running until the scene is initialized. The only two exceptions are `move_camera()` and `resize()`. I assumed this is because they can run at any time, since they only affect the view of the scene and aren't dependent on its contents.

falkoschindler (approved):
Yes, that makes sense. 👍🏻

Issue #3393, opened 2024-07-24
Aggrid upgrade. v30 to v32


ruidiasbraga:
### Description

Hi, 

after some unexpected behaviours from what I see on the Aggrid documentation and what I experienced in nicegui, I found out that the version difference (v30) could be the problem.

nicegui is using v30 and the latest is v32 of aggrid. are you considering and upgrade?

python-and-fiction:
nicegui 2.0 will upgrade dependency。

falkoschindler:
Yes, we're working on version 2.0 which will update Python dependencies like FastAPI as well as JavaScript libraries like AG Grid. To avoid breaking changes, we need to wait for such a major update. We hope to get it out by the end of August.

ruidiasbraga:
Great news! Thanks for the answer :)

gaby:
@falkoschindler Probably worth adding a dependabot.yml to track dependencies.

falkoschindler:
@gaby Sounds interesting. So far we don't have experience with dependabot.yml. Would you like to create a new feature request or pull request, also explaining benefits over using Dependabot without such a config file? Thanks!

gaby:
@falkoschindler Using the `dependabot.yml` will enable the feature on the repo, you can set a schedule and it will auto-create a Pull Request any time a dependency has an update. It supports python, gh actions, docker, javascript, etc. I will do a PR and we can discuss schedule there.

PR #3453, opened 2024-08-07
Add support for dependabot

gaby:
- Track the following using dependabot on a daily basis:
  - github-actions
  - python
  - docker
- Bump base image from 3.11.3 to 3.11.9 which is the latest patch release: https://endoflife.date/python 

Related to #3393


falkoschindler (commented):
Thanks for the pull request, @gaby!

Regarding the dockerfiles: I'm not sure why we specify patch versions here. In development.dockerfile we simply use `python:3.8-slim` which should use the latest patch automatically. @rodja Do you remember a reason for using `python:3.11.3-slim` instead of `python:3.11-slim`?

rodja:
I think there was no special reason for fixing on the patch version. We can change it to `3.11-slim`.

falkoschindler (approved):
Ok, I removed the patch version from the dockerfiles. Let's merge! 👍🏻

gaby:
@falkoschindler Thanks! It already created a few PR's :-)

PR #3470, opened 2024-08-09
Initial implementation of anchor positioning

mikaelho:
**Please note:** I appreciate that the project is busy with many many things, and I did not really intend to create a PR towards the project yet. At the same time the PR is the most convenient way to look at the code, if anyone is interested. Please feel free to close the PR at any time.

This PR is an Initial draft implementation of anchor positioning, and is related to this discussion: https://github.com/zauberzeug/nicegui/discussions/3398

Features:
- `.at` property as a gateway from nicegui elements to anchor positioning
- Basic edge and center connections like `.at.a(top=bottom(previous), center=center(previous))`
- Convenience "docking" features like `.at.top_left` or `.at.below(previous)`
- Dynamic sizing like `.at.fit_height`

See [this](https://github.com/zauberzeug/nicegui/discussions/3398#discussioncomment-10290661) discussion comment for a picture of the demo app and the snippet of code used to create it.

### DIFF
--- a/anchor_demo.py
+++ b/anchor_demo.py
@@ -1,0 +1,47 @@
+from nicegui import context
+from nicegui import ui
+from nicegui.anchor import bottom
+from nicegui.anchor import parent
+from nicegui.anchor import previous
+from nicegui.anchor import right
+from nicegui.anchor import width
+
+context.client.content.classes("h-[100vh]")
+
+ui.page_title("Anchor demo")
+
+def main():
+    with ui.element().classes("w-full h-full"):
+        with ui.element().classes("bg-primary").at.top.fit_height as top_bar:
+            ui.icon("home", size="lg", color="white").at.top_left
+            ui.label("SAMPLE APP").style("margin: 10px").at.center.tailwind.text_color("white")
+            ui.button(icon="menu").at.top_right
+        with ui.element().at.bottom_left.a(top=bottom(top_bar)).classes("bg-primary") as drawer:
+            add_left_drawer_items()
+        with ui.element().at.bottom_right.a(top=bottom(top_bar), left=right(drawer)) as content_area:
+            with ui.element().at.center.fit_height.a(width=max(width(parent)/3, 200)):
+                input_with_id(placeholder="Username").at.top
+                input_with_id(placeholder="Password").at.below(previous).a(width=width(previous))
+                ui.button("Login").at.below(previous)
+        ui.button(icon="add").at.bottom_right.props("fab")
+
+
+def add_left_drawer_items():
+    with ui.list().props("separator").classes("text-white").style("margin: 10px"):
+        ui.item("Sections").props("header").classes("text-bold")
+        for i in range(1, 5):
+            with ui.item():
+                with ui.item_section().props("avatar"):
+                    ui.icon("chevron_right")
+                with ui.item_section():
+                    ui.item_label(f"Section {i}")
+
+def input_with_id(*args, **kwargs):
+    with ui.element() as wrapper:
+        ui.input(*args, **kwargs).classes("w-full")
+    return wrapper
+
+
+main()
+
+ui.run()
--- a/nicegui/anchor.py
+++ b/nicegui/anchor.py
@@ -1,0 +1,279 @@
+__all__ = [
+    "bottom",
+    "center",
+    "left",
+    "middle",
+    "parent",
+    "previous",
+    "right",
+    "top",
+]
+
+import numbers
+from functools import partial
+from string import Template
+from typing import Self
+
+from nicegui import Tailwind
+from nicegui.element import Element
+
+
+PseudoElement = str
+Elementish = Element|PseudoElement
+
+
+CALC_LEFT_FOR_CENTER = "(var(--internal-width-$parent_id) / 2 - var(--external-width-$self_id) / 2)"
+CALC_TOP_FOR_MIDDLE = "(var(--internal-height-$parent_id) / 2 - var(--external-height-$self_id) / 2)"
+LEFT_FOR_CENTER = f"margin-left: 0; left: calc({CALC_LEFT_FOR_CENTER})"
+TOP_FOR_MIDDLE = f"margin-top: 0; top: calc({CALC_TOP_FOR_MIDDLE})"
+
+
+class At:
+    def __init__(self, element: Element):
+        self.element = element
+
+    @staticmethod
+    def prop(spec):
+        def getter(self) -> Self:
+            parent = self.element.parent_slot.parent
+            if style := spec.get("style"):
+                if isinstance(style, Template):
+                    style = style.substitute({"parent_id": f"c{parent.id}", "self_id": f"c{self.element.id}"})
+                self.element.style(style)
+            if classes := spec.get("classes"):
+                self.element.classes(classes)
+            return self
+
+        return property(getter)
+
+    top = prop({"style": "left: 0; right: 0; top: 0"})
+    bottom = prop({"style": "left: 0; right: 0; bottom: 0"})
+    left = prop({"style": "left: 0; bottom: 0; top: 0"})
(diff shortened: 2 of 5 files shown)


rodja:
Wow, you really threw yourself into it. Looks impressive. I only played with it a short time and came up with the following thoughts:

- I think the naming of `At.a(...)` is problematic; while its short the purpose is not clear. I guess you picked `a` as abbreviation for anchor. But it does neither return or create an anchor but adds css styles. Maybe something like `At.and(...)` would be better: `ui.element().at.bottom_left.and(top=bottom(top_bar))`.
- The `Element.at` entry is quite destructive because it makes the element (and probably also its parent `relative`. That means it can not be easily mixed with other means of layouting.
- There is currently no way to change/remove the anchoring. For example what would a developer have to do to make the login button right aligned after the password was typed?
- I wonder for what kinds of layout-jobs the anchoring would be really useful. The example you gave was very easy to reconstruct with existing NiceGUI capabilities:

```py
with ui.row().classes("bg-primary w-full text-white p-2 justify-between"):
    ui.icon("home", size="lg", color="white")
    ui.label("SAMPLE APP").style("margin: 10px")
    ui.button(icon="menu")
with ui.row().classes("h-full w-full"):
    with ui.column().classes("bg-primary h-full"):
        add_left_drawer_items()
    with ui.column().classes("m-auto w-1/3 items-center"):
        input_with_id(placeholder="Username").classes("w-full")
        input_with_id(placeholder="Password").classes("w-full")
        ui.button("Login")
with ui.page_sticky(x_offset=18, y_offset=18):
    ui.button(icon="add").props("fab")
```

mikaelho:
I also need to say wow, and thank you for taking the time for the very thoughtful feedback.

> I think the naming of At.a(...) is problematic; while its short the purpose is not clear. I
> guess you picked a as abbreviation for anchor.

True. I am not happy having that method at all, and there is an issue with properties as well -- my
IDE complains about `ui.element().at.top` "doing nothing". I have been considering a lot of
alternatives, all with their pros and cons. For example, just having an `at()` method that would
contain everything, e.g. like this:

`ui.element().at(bottom_left, top=bottom(top_bar))`

... is nice and tight, but we lose the discoverability via IDE completion.

I will continue playing with the alternatives, as there is no hurry.

> The Element.at entry is quite destructive because it makes the element (and probably also its
> parent relative. That means it can not be easily mixed with other means of layouting.

Hmm. For anchor positioning, the element **must be**
`position: absolute`. And for an HTML element to be absolutely positioned, its parent **must be**
positioned somehow. Thus, the code checks whether the parent is positioned, and only if not, applies
the `relative` positioning - which is effectively a no-op, if I am not mistaken.

If you see that there are scenarios where this approach is actually destructive, we could just fail
if the dev tries to user anchor positioning when the parent does not define `position`. Would like
to avoid that, though.

As for the element itself being `position: absolute`, while you would want to combine anchor 
positioning with some other CSS layout model, that could be a show-stopper, although there I did 
not yet think of an actual use case.

> There is currently no way to change/remove the anchoring. For example what would a developer
> have to do to make the login button right aligned after the password was typed?

Fair point, and this was on my to do list. Technically, releasing an anchor is the same as setting
e.g. `top: initial`.

Implementation could:

- Support setting the edge to `None`.
- Add logic to implicitly release constraints when there are too many. Here we would need to decide
  what the heuristic is; elsewhere, I have used the priority order which resulted in dropping `left`
  if `right` is set when `width` is also set, but if you want to release `width`
  you need to do it explicitly.

> I wonder for what kinds of layout-jobs the anchoring would be really useful. The example you
> gave was very easy to reconstruct with existing NiceGUI capabilities.

This is the big question, yes. My thesis is that yes, if you are already familiar and comfortable
with all the different CSS layout modes and Tailwind classes, as well as used to thinking in terms
of (often invisible) rows and columns, anchor positioning might be of little interest to you.

But if you come from a non-HTML, non-flow-oriented UI design background, or indeed have no UI
background at all, anchor positioning with the "put that there" mentality might be much more
approachable. I know it is much more approachable to me, even after decades of intermittent exposure
to HTML layout.

Thus, I see anchor positioning as potentially valuable if you want to make nicegui approachable to
the Python programmer that wants to create a web UI, without needing to learn a lot
HTML/CSS/Tailwind to get a simple layout in place.

There is also one edge feature of the anchor positioning, as already implemented in this PR, that I
am not sure would be easy to cover without writing custom CSS by hand, and that is "nudging"
the position of an element with arbitrary arithmetic, e.g.:

`center=center(parent) + width(previous)/4`

falkoschindler:
Thanks @mikaelho! I'm very impressed by the work you already put into this feature. I have some concerns though:

- Is anchor positioning really easier (for the non-expert) when dealing with more than 10 elements? What sounds simple when starting with a blank page, might turn into a headache when formulating anchor-based relations for a complex UI. I can't speak from experience, but it's a concern of mine. Maybe it's a matter of taste.

- If everything is anchored with `position:absolute`, how to resolve size restrictions? Will elements overflow and overlap by default? Isn't it extra hard when trying to design for small (mobile) screens? Flow-based layouting seems to be much more robust and reactive.

- I don't think adding `position:relative` is a no-op in CSS. At least it adds a new positioning context, which can have impact on the position of child elements. This could cause subtle bugs that are hard to find, especially for beginners.

- I'd like to see how anchor positioning takes of in web development and if there's interest in the NiceGUI community before diving too deep into the rabbit hole. A proof of concept like this PR is a great way to discuss the basic idea. But turning it into a complete layout system, that plays nicely with the existing library, could be _pretty_ challenging.

mikaelho:
For clarity, while I have long been a fan of anchor positioning, this is still just fun exploration for me, and I am perfectly ok if you think it is not a good fit with what you want to do with NiceGUI. Especially since anchor positioning does not even have full browser support, yet.

That said:

> Is anchor positioning really easier (for the non-expert) when dealing with more than 10 elements?

I might have given the impression with the demo that I think that anchor positioning is something you would "have to" build the whole UI with, but that is certainly not my intent. I specifically added the `ui.list` in the demo to showcase peaceful coexistence between the different models. I would probably use a `ui.row` for the top bar, as it is a natural fit. I would definitely use a `ui.grid` or `ui.aggrid` component in the main content area, if a grid was needed. Then I could again use some anchoring within the cards in the grid, if that would be most natural, etc.

> If everything is anchored with `position:absolute`, how to resolve size restrictions? Will elements overflow and overlap by default? Isn't it extra hard when trying to design for small (mobile) screens? Flow-based layouting seems to be much more robust and reactive.

It is not hard to use the anchors and min/max conditionals to keep things quite flexible, but as said above, there is no reason not to use the best available tool in each case. Maybe I would use reactive rows/columns to let the top-level layout change with screen size/orientation, then do the layout within those with anchors, as appropriate.

As an aside, I have also been thinking of a context manager that would let you define distinctly different child element trees based on the current dimensions of the parent element (which could be the whole viewport), for maximum control over the reactivity.

> I don't think adding position:relative is a no-op in CSS. At least it adds a new positioning context, which can have impact on the position of child elements. This could cause subtle bugs that are hard to find, especially for beginners.

Stacking context can certainly be an issue in some cases, and overall this is something I am trying to get some hands-on experience with. In my previous experience, a lot of the HTML/CSS layout concepts are hard for beginners, and I am not sure that relative positioning, applied only when needed, would be a net negative.

> I'd like to see how anchor positioning takes of in web development and if there's interest in the NiceGUI community before diving too deep into the rabbit hole.

I fully agree. Currently, anchor positioning in most materials is presented as just a way to add some callout bubble to an element, which is really underselling the concept. CSS continues to have its quirks, still requiring JS support to make anchor positioning fully functional. And there is the issue of incomplete browser support, as already mentioned.

Thanks to both you for all the input. I think it makes sense to close this PR now. I will keep refining the API, also possibly experimenting with things like size and animation contexts, and radial anchors.

PR #3476, opened 2024-08-11
ui.tree select/deselect, tick/untick programmatically

arunaseva:
Link to the description:

https://github.com/zauberzeug/nicegui/discussions/3436#discussioncomment-10297039

### DIFF
--- a/nicegui/elements/tree.py
+++ b/nicegui/elements/tree.py
@@ -78,21 +78,19 @@ class Tree(Element):
         self._select_handlers.append(callback)
         return self
 
-    def select(self, node_key: Optional[str] = None) -> Self:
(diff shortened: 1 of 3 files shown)


falkoschindler (approved):
Thanks for the pull request, @arunaseva!

I just reviewed your code and did some minor changes:

- I changed `select(None)` to deselect the nodes instead of doing nothing. This feels more intuitive.
- The code for `untick()` could be simplified a bit. We don't need to intersect before computing the difference.
- The tests could be simplified with a label showing the current selection and tick list, resulting in rather readable `should_contain` statements.
- Last but not least, I decided to split the select and tick demo. This reduces complexity and is more in line with the existing expand/collapse demo.

Let's merge into main! 🚀

PR #3482, opened 2024-08-12
Allow filtering nodes in `ui.tree`

falkoschindler:
Inspired by https://github.com/zauberzeug/nicegui/discussions/3436#discussioncomment-10297039, this PR adds the `FilterElement` mixin to `ui.tree`. This allows to filter nodes by name.


rodja (approved):
Cool!

Issue #3486, opened 2024-08-13
Test user fails when using ElementFilter to look for select elements


marcuslimdw:
### Description

**What I'm trying to do**

Write integration tests using `nicegui.testing.User` for a NiceGUI page that contains a `Select` element:

```py
async def test_this(nicegui_user: testing.User):  # nicegui_user is a fixture I defined that sets up my app
    await nicegui_user.open("/new/vendors")
    await nicegui_user.should_see("Vendor name")
```

**What I expected**

The test runs successfully.

**What happened**

The test fails on the `should_see` call with:

```
self = <nicegui.element_filter.ElementFilter object at 0x13338d7e0>

    def __iter__(self) -> Iterator[T]:
        # pylint: disable=protected-access
        for element in self._scope.descendants():
            if self._kind and not isinstance(element, self._kind):
                continue
            if self._exclude_kinds and isinstance(element, tuple(self._exclude_kinds)):
                continue
    
            if any(marker not in element._markers for marker in self._markers):
                continue
            if any(marker in element._markers for marker in self._exclude_markers):
                continue
    
            if self._contents or self._exclude_content:
                element_contents = [content for content in (
                    element._props.get('text'),
                    element._props.get('label'),
                    element._props.get('icon'),
                    element._props.get('placeholder'),
                    element._props.get('value'),
>                   element._props.get('options', {}).get('message'),
                    element.text if isinstance(element, TextElement) else None,
                    element.content if isinstance(element, ContentElement) else None,
                    element.source if isinstance(element, SourceElement) else None,
                ) if content]
E               AttributeError: 'list' object has no attribute 'get'

../../../../../.venv/lib/python3.10/site-packages/nicegui/element_filter.py:111: AttributeError
```

For the specific case of a `SelectElement`, `element._props['options']` is a `list`. If I comment out the `ui.select` calls, the test passes.

falkoschindler:
Thanks for reporting this bug, @marcuslimdw!

@rodja Do you think we still need the line

```py
element._props.get('options', {}).get('message')
```

after changing `ui.notify` in PR #3457? `get('message')` looks like it was meant to handle notifications. But it turns out to be rather fragile.

rodja:
@falkoschindler, no, the line is needed to find `ui.notification` elements (and for example dismiss them). I just created #3487 to provide better support for `ui.select`. This also fixes this issue.

rodja:
Should be fixed with #3487 and hence available in next release 1.4.35.

PR #3487, opened 2024-08-13
Support ui.select elements for user simulation tests

rodja:
As @marcuslimdw described in #3486, the user simulation for integration tests fails with `ui.select` elements. This PR does not only fix the exception but also improves support for `ui.select` to allow actual selection via `user.find(...).click()`.


falkoschindler (approved):

marcuslimdw:
thank you for the quick fix @falkoschindler 🙏

PR #3511, opened 2024-08-14
Provide two additional pytest plugins to make user and screen able to load independently

rodja:
This PR implements feature request #3510 by creating two additional plugins `nicegui.testing.user_plugin` and `nicegui.testing.screen_plugin`.


rodja:
I noticed that the example projects chat_app, authentication and todo_list are showing `PytestUnknownMarkWarning` despite the marker being registered correctly. This also happens on main. After some digging I suspect it's a known issue of pytest (see #9677) and should not hinder this PR from being able to get merged.

nicegui/testing/general_fixtures.py
```diff
@@ -0,0 +1,80 @@
...
+from nicegui import Client, app, binding, core, run, ui
+from nicegui.page import page
+
+# pylint: disable=redefined-outer-name
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    """Add the "module_under_test" marker to the pytest configuration."""
+    config.addinivalue_line('markers',
+                            'module_under_test(module): specify the module under test which then gets automatically reloaded.')
```
falkoschindler, line 17: Why did you add `(module)`? What does it mean?

nicegui/testing/plugin.py
```diff
@@ -1,202 +1,3 @@
...
-        binding_refresh_interval=0.1,
-        reconnect_timeout=3.0,
-        tailwind=True,
-        prod_js=True,
-        show_welcome_message=False,
-    )
-    nicegui.storage.set_storage_secret('simulated secret')
+from .general_fixtures import nicegui_reset_globals, pytest_configure  # noqa: F401
+from .screen_plugin import nicegui_chrome_options, nicegui_driver, nicegui_remove_all_screenshots, screen  # noqa: F401
+from .user_plugin import create_user, prepare_simulated_auto_index_client, user  # noqa: F401
```
falkoschindler, line 4: Can't we add something like `__all__ = [...]` to avoid the need for `noqa` markers?

falkoschindler (approved):
Ok, looks good to me. I only added two minor comments to the code.

nicegui/testing/general_fixtures.py
rodja: It indicates that the marker needs a parameter. Like here: https://docs.pytest.org/en/stable/how-to/writing_plugins.html#registering-markers

nicegui/testing/plugin.py
rodja: Ruff still reports the error if I add a `__all__` statement. Another way would be to ignore F401 for the whole file. Which is more similar to what Ruff is doing with `__init__.py` files.

falkoschindler:
@rodja This PR is still marked as draft. Is that correct, or can we merge?

PR #3544, opened 2024-08-20
Fix orbit controls of `ui.scene` after changing the camera's up vector

voigta:
This PR solves the problem that scenes need to have their Z-Axis "upwards" in order to have working orbit controls. If one tries to manually move the camera with `scene.move_camera(up_z=-1)`, the orbit controls are not updated and the mouse-interaction gets weird.

PS: Thanks for creating and sharing this awesome library! 💪

### DIFF
--- a/nicegui/elements/scene.js
+++ b/nicegui/elements/scene.js
@@ -85,7 +85,7 @@ export default {
     }
     this.look_at = new THREE.Vector3(0, 0, 0);
     this.camera.lookAt(this.look_at);
-    this.camera.up = new THREE.Vector3(this.camera_up[0], this.camera_up[1], this.camera_up[2]);
+    this.camera.up = new THREE.Vector3(0, 0, 1);
     this.camera.position.set(0, -3, 5);
 
     this.scene.add(new THREE.AmbientLight(0xffffff, 0.7 * Math.PI));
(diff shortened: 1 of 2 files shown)


falkoschindler (commented):
Thanks for the pull request, @voigta!

But if I'm not mistaken, you can already change the camera's "up" vector via the `move_camera()` method:
```py
with ui.scene() as scene:
    scene.move_camera(up_x=1, up_y=0, up_z=0)
    scene.cylinder(0, 1)
```
Therefore I don't think this PR is necessary.

voigta:
Yes, you can already change the "up" vector via move_camera but the OrbitControls are not updated accordingly. It looks like OrbitControls doesn't allow for a change in "up" because it only uses the object.up property once here: https://github.com/zauberzeug/nicegui/blob/252a111740fc7c87bb8069284f55551321fca1af/nicegui/elements/lib/three/modules/OrbitControls.js#L170 

An example to show the problem I want to solve:
```python
from nicegui import ui

ui.label("A regular scene with correct mouse interaction:")
with ui.scene() as scene_regular:
    scene_regular.cylinder(0, 1)

ui.label("An inverted scene with inverted mouse interaction:")
with ui.scene() as scene_inverted:
    scene_inverted.move_camera(up_z=-1)
    scene_inverted.cylinder(0, 1)

ui.run()
```

Another (maybe cleaner?) solution would be to create a new OrbitControls object in case camera.up is changed here:
https://github.com/zauberzeug/nicegui/blob/252a111740fc7c87bb8069284f55551321fca1af/nicegui/elements/scene.js#L425-L431

falkoschindler (approved):
Nice! This solution seems reasonable. Let's merge!

PR #3546, opened 2024-08-20
New style of joystick

python-and-fiction:
Inspired by https://github.com/zauberzeug/nicegui/discussions/3531
Make styling joystick more easier.
We can customize the size and the dark mode background color of joystick by this way:
```python3
ui.joystick(color="red", size=50, mode="static").classes(
    "w-96 h-96 bg-red-100 dark:bg-green-100"
)
```
light mode:

dark mode:

Note:
After this change, joystick will lose default background color. If you use it without any parameter or additional style, it is transparent.

### DIFF
--- a/nicegui/elements/joystick.vue
+++ b/nicegui/elements/joystick.vue
@@ -1,5 +1,5 @@
 <template>
-  <div><div></div></div>
+  <div class="nicegui-joystick"><div></div></div>
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


falkoschindler (approved):
Wow, thank you for this awesome pull request, @python-and-fiction!
This also solves feature request #1707, because we finally can change the size of `ui.joystick` in an intuitive way. 🚀

PR #3547, opened 2024-08-20
Add coordinate_system object

voigta:
This PR adds a coordinate system object to the 3d scene, which can visualize the origin or any other transformation / 6D-pose.


falkoschindler (changes requested):
Thanks for the pull request, @voigta!

But I think the implementation is broken:
Screenshot 2024-08-29 at 10 29 35

Apart from that, there is an [`AxesHelper`](https://threejs.org/docs/index.html?q=axes#api/en/helpers/AxesHelper) object in Three.js. For consistency with other elements, it might be better to simply provide an API for the `AxesHelper` instead of creating a custom object.

I could, however, imagine adding yours as an additional demo "Custom coordinate system indicator". It nicely demonstrates how to combine existing objects to a custom object.

voigta:
The following changes have been implemented:
1. Add `axes_helper` using the Three.js `AxesHelper` object
2. Add (and fix) the custom `CoordinateSystem` object as an exmple

nicegui/elements/scene.js
```diff
@@ -275,6 +275,9 @@ export default {
           undefined,
           (error) => console.error(error)
         );
+      } else if (type == "axes_helper") {
+        mesh = new THREE.AxesHelper(args[0]);
+        mesh.material.transparent = true;
```
falkoschindler, line 280: What is the purpose of setting `transparent = true`? My feeling is that without this setting the axes tend to flicker less while orbiting.

falkoschindler (approved):
Thanks for the update, @voigta! Looks good to me.
I just wonder whether we need to set `transparent = true` (see my comment in the code).

nicegui/elements/scene.js
voigta: To be honest, I tried both options and haven‘t noticed a difference. In the end, I tried to stick to the line and curve materials:

https://github.com/zauberzeug/nicegui/blob/1a47d90b16292375459d10a7438ab2958dc34bee/nicegui/elements/scene.js#L226

But we can also remove it if you perfer

falkoschindler: Oh I see! If we omit this line, we can't reduce the opacity with `material(..., 0.5)`.
It might be worth thinking about automatically switch this flag based on the current opacity. But that's out of scope of this pull request. Let's merge!

PR #3549, opened 2024-08-20
Allow draggable groups

voigta:
This PR allows groups to be draggable. Currently there is no effect when we call group.draggable().

As noted in the [documentation of DragControls](https://threejs.org/docs/#examples/en/controls/DragControls.transformGroup), this only works if the scene contains a single draggable group. This might lead to some confusion but ignoring the `.draggable()` call for groups is also confusing...😅 At least setting the `transformGroup` property to true only affects the behavior for groups and not for other objects.

### DIFF
--- a/nicegui/elements/lib/three/modules/DragControls.js
+++ b/nicegui/elements/lib/three/modules/DragControls.js
@@ -334,12 +334,7 @@ function onPointerDown( event ) {
 
 			// look for the outermost group in the object's upper hierarchy
 
-			_selected = findGroup( _intersections[ 0 ].object );
-
-            // Use the object itself if no group was found
-            if ( !_selected ) {
-                _selected = _intersections[ 0 ].object;
-            }
+			_selected = findGroup( _intersections[ 0 ].object ) || _intersections[ 0 ].object;
 
 		} else {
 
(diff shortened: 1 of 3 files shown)


falkoschindler:
We see the following options:

1. Raise a `NotImplement` exception when calling `draggable()` on a group as a temporary workaround to avoid confusion.
2. Raising a `RuntimeError` when calling `draggable()` if there are other draggables defined causing a conflict.
3. Somehow solving the problem, e.g. by dynamically updating the `transformGroup` on mouse down.

What do you think, @voigta?

voigta:
Thanks for the feedback @falkoschindler.

Solving the problem (i.e. option 3) would definitely be the cleanest solution. 

There have been updates of DragControls.js in [three.js r162](https://github.com/mrdoob/three.js/releases/tag/r162) that allow multiple draggable groups. However, the situation of a draggable group and a draggable object is still not nicely handled in the official repo. Therefore, I've added a small modification that should fix this problem and your example is now working as expected.

falkoschindler (approved):
Thanks for the update, @voigta!

The new solution looks really neat and it behaves just as expected! What a nice incidence that we just upgraded three.js.
I just moved the change into our npm.py script, so that it is automatically applied when updating third-party libraries.

Let's merge! 🚀

PR #3559, opened 2024-08-21
Lazy-load NumPy only if it is needed

falkoschindler:
Inspired by https://github.com/zauberzeug/nicegui/discussions/3533#discussioncomment-10397937 and similar to PR #3548, this PR reduces NiceGUI's startup time by deferring the import of NumPy to when it is actually needed.

### DIFF
--- a/nicegui/json/builtin_wrapper.py
+++ b/nicegui/json/builtin_wrapper.py
@@ -1,14 +1,11 @@
+import importlib.util
 import json
 from datetime import date, datetime
 from typing import Any, Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


rodja (approved):

ed2050:
Seems to be an error.  Line 8 of builtin_wrapper.py says pandas instead of numpy:
```py
HAS_NUMPY = importlib.util.find_spec('pandas') is not None
```

falkoschindler:
Thanks for spotting this oversight, @ed2050! I'll fix it on main.

PR #3561, opened 2024-08-21
Fix bug when opening/closing `ui.select` on air

falkoschindler:
This PR fixes a bug unveiled by PR #3487 which registered for the "popup-show" and "popup-hide" events. Apparently, a check in air.py assumed the first argument to be a string, which it isn't. This PR checks for the correct type and, while at it, removes the arguments for the popup events altogether, because they are not needed.

### DIFF
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -140,10 +140,11 @@ class Air:
             if client_id not in Client.instances:
                 return
             client = Client.instances[client_id]
-            if data['msg']['args'] and data['msg']['args'][0].startswith('{"socket_id":'):
-                args = json.loads(data['msg']['args'][0])
-                args['socket_id'] = client_id  # HACK: translate socket_id of ui.scene's init event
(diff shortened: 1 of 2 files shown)


rodja (approved):
Setting the args to `[]` was not really necessary. Otherwise this fix is very good. On Air could not cope with **any** event which had no arguments (eg. `None`). For example our [custom events demo](https://nicegui.io/documentation/generic_events#custom_events) is also breaking. The On Air relay server is simply passing the event msg to through as it is -- and that means `None` in the case of js events with no arguments.

falkoschindler:
@rodja Yes, setting args to `[]` isn't necessary, but potentially reduces the payload. I feel it's good practice for the library code to register events without arguments if they are not needed at all like, e.g., for a button click:
https://github.com/zauberzeug/nicegui/blob/0939d863fdc90c944236307a9db5f345be0cf858/nicegui/elements/button.py#L44

PR #3567, opened 2024-08-22
Use PUID/PGID for chown statements

zan73:
When the passed PGID already exists in the base image, the groupadd is not executed hence a reference to appgroup in the chown statements fails


rodja (approved):
Works for me. Build and tested with

```bash
docker build --build-arg VERSION=1.4.27  -t nicegui-3567 -f release.dockerfile .
docker run --env PUID=100 --env=PGID=100 --rm -it nicegui-3567 id
```

I get the warning 

```
useradd warning: appuser's uid 100 outside of the UID_MIN 1000 and UID_MAX 60000 range.
```

But the startup does not crash and the id is also correctly set to 100.

PR #3580, opened 2024-08-23
Delay events before handshake

falkoschindler:
In https://github.com/zauberzeug/nicegui/discussions/3530#discussioncomment-10369711 we noticed that Firefox auto-fills input fields before the websocket handshake happens, so that the information is lost and the server doesn't update its state. In this example the notification shows empty credentials although Firefox shows filled out input fields:
```py
username = ui.input('Username')
password = ui.input('Password', password=True, password_toggle_button=True)
ui.button('Login', on_click=lambda: ui.notify(f'{username.value}:{password.value}'))
```

This PR fixes this problem by waiting before emitting events until the handshake happens.

Technically there's a new possibility for messing up the state:
If the server sends an value update right after the handshake happens, the client might still be holding back the change event, causing server and client getting out of sync again. But by using a very short interval of 10ms this problem seems very unlikely.
```py
@ui.page('/')
async def page():
    username = ui.input('Username')
    password = ui.input('Password', password=True, password_toggle_button=True)
    ui.button('Login', on_click=lambda: ui.notify(f'{username.value}:{password.value}'))
    username.value = 'foo'
```

Open tasks:

- [x] fix table pytests
- [ ] review

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -181,7 +181,11 @@ function renderRecursively(elements, id) {
           args: stringifyEventArgs(args, event.args),
         };
         const emitter = () => window.socket?.emit("event", data);
-        throttle(emitter, event.throttle, event.leading_events, event.trailing_events, event.listener_id);
+        const delayed_emitter = () => {
+          if (window.did_handshake) emitter();
+          else setTimeout(emitter, 10);
+        };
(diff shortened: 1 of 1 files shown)


falkoschindler:
Analyzing failing pytests, I noticed that this PR also emits events that used to be ignored before, like an initial "update:pagination" for `ui.table`. I hope we don't break any user code by not dropping such events.

rodja (approved):
Works as expected (also with the original reproduction).

Issue #3581, opened 2024-08-23
`aggrid.get_client_data` returns unsorted data


JeppeKlitgaard:
### Description

I am finding that `AgGrid.get_client_data` returns the unsorted and unfiltered data. This makes sense as it uses the `api.forEachNode` call internally. It would if the forEach method could be defined in the `get_client_data` call as an optional argument.

See: https://www.ag-grid.com/javascript-data-grid/accessing-data/

I will work on a PR that adds this

PR #3582, opened 2024-08-23
feat: implement access method for get_client_data

JeppeKlitgaard:
Allows access to client data in a way that also takes care of sorting and filtering.

Fixes: #3581

I think I have followed the steps outlined in CONTRIBUTING, but perhaps I have managed to miss something


falkoschindler (approved):
Thanks for the pull request, @JeppeKlitgaard!
I just condensed the documentation for the parameter a bit (because we already refer to the AG Grid documentation), renamed the parameter to "method", and added a pytest for `get_client_data`. Let's merge!

PR #3585, opened 2024-08-24
Allow nicegui.test to be used without selenium dependencies

rodja:
As a follow up to #3511 this PR tries to fix #3510 by enclosing the `Screen` import in a try/cecept.

### DIFF
--- a/nicegui/testing/__init__.py
+++ b/nicegui/testing/__init__.py
@@ -1,4 +1,9 @@
-from .screen import Screen
+try:
+    from .screen import Screen
+except ImportError:
+    # we simply define Screen as None if selenium is not installed
+    # this allows simpler dependency management when only using the "User" fixture
+    Screen = None  # type: ignore
 from .user import User
 from .user_interaction import UserInteraction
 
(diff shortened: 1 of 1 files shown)


nicegui/testing/__init__.py
```diff
@@ -1,4 +1,15 @@
-from .screen import Screen
+try:
+    from .screen import Screen
+except ImportError:
+    # we simply define Screen as None if selenium is not installed
+    # this allows simpler dependency management when only using the "User" fixture
+    class Screen:  # type: ignore
+        def __init__(self, *args, **kwargs):
+            raise ImportError('Selenium dependencies are missing. Please install selenium to use the Screen class.')
```
falkoschindler, line 8: Are we sure Selenium is missing?
Maybe we can re-raise the original import error?

rodja: Good idea. I added it in [5d6a236](https://github.com/zauberzeug/nicegui/pull/3585/commits/5d6a2365676e049b079fa852cfd1f2c559895563).

falkoschindler (approved):

PR #3586, opened 2024-08-24
Improve support for drawing items in `ui.leaflet`

falkoschindler:
In #2422 we noticed that `ui.leaflet` didn't preserve drawn items, making it impossible to edit or delete them.

- This PR adds newly created layers to the `drawnItems` feature group, allowing to modify them using the draw controls. It also removes the need to add drawn items in user code. This breaking change has been deliberately scheduled for the release of version 2.0.0.
- To allow rendering drawn items with custom options like before, a new parameter `hide_drawn_items` allows to opt-out of the new behavior.
- The PR also fixes cyclic references in some event arguments, allowing to subscribe to events like "draw:created", "draw:edited" and "draw:deleted".
- **It does not** implement synchronization of drawn layers with the server. This means that drawn items are not synchronized on the auto-index page and cleared when reloading it. Extracting all information from the event arguments and using it to create drawn layers programmatically on the client turned out pretty complicated. And because Leaflet is on the verge of releasing version 2.0, I decided not to spend too much time trying to dig into the API and data structures of version Leaflet 1.9.

---

Open tasks:

- [x] demo showing how to edit and remove drawn items
- [x] demo how to draw items with custom options

### DIFF
--- a/nicegui/elements/leaflet.js
+++ b/nicegui/elements/leaflet.js
@@ -9,6 +9,7 @@ export default {
     options: Object,
     draw_control: Object,
     resource_path: String,
+    hide_drawn_items: Boolean,
   },
   async mounted() {
     await this.$nextTick(); // NOTE: wait for window.path_prefix to be set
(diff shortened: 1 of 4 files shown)


rodja:
I think we should add demos on how a drawn item can be modified and another showing how to draw items with custom options.

rodja (approved):

PR #3588, opened 2024-08-24
Provide public API for accessing `_props`, `_classes` and `_style`.

falkoschindler:
This PR introduces a `Props` class derived from `dict` to replace the `props()` method. By implementing the `__call__` method, the new `.props` property can be called like before, but also accessed like a dictionary. This allows reading and writing the props dictionary without accessing the private `_props` attribute. Furthermore it moves most of the rather complex parsing logic into a separate file, relieving the pretty long element.py module.

Currently this PR is only a draft. If we like the idea, we should

- [x] fix pytests
- [x] apply the same concept to `.classes` and `.style`
- [x] fix linting errors in `ui.query`
- [x] fix linting errors in `ui.tree` and `ui.menu`
- [x] use the new API in tests, documentation and examples

### DIFF
--- a/tests/test_element.py
+++ b/tests/test_element.py
@@ -204,105 +204,105 @@ def test_default_props(nicegui_reset_globals):
     ui.button.default_props('rounded outline')
     button_a = ui.button('Button A')
     button_b = ui.button('Button B')
-    assert button_a._props.get('rounded') is True, 'default props are set'
-    assert button_a._props.get('outline') is True
-    assert button_b._props.get('rounded') is True
-    assert button_b._props.get('outline') is True
+    assert button_a.props.get('rounded') is True, 'default props are set'
+    assert button_a.props.get('outline') is True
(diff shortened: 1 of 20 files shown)


tests/test_element.py
```diff
@@ -34,37 +36,37 @@ def assert_classes(classes: str) -> None:
 
 def test_style_parsing(nicegui_reset_globals):
     # pylint: disable=protected-access
```
marcuslimdw, line 38: should this be removed (same for `test_props_parsing`?

also, would it be better to use test parameterization here, instead of having multiple unrelated `assert` statements together?

falkoschindler: Thanks for the suggestions, @marcuslimdw!
Removing the pylint markers is a good idea. They are not needed anymore.
And even though I think test parametrization is a bit a matter of taste, you and claude.ai convinced me that there are more pros than cons for following this approach.

PR #3631, opened 2024-08-28
[WIP] Zeitline component

Aluriak:
Following #3620 , i tried something. Does not work (got an empty page), which is visible through the test.

I don't know anything about vue.js, so i guess my zeitline.js file has several problems.

Can someone help me ?


rodja:
We do not have the time, but maybe someone from the community can jump in?

nicegui/elements/zeitline.js
```diff
@@ -0,0 +1,29 @@
...
+    },
+  },
+  data() {
+    return {
+      linePromise: new Promise((resolve) => {
+        this.resolveLine = resolve;
+      }),
+  },
+  async mounted() {
+    this.line = new Zeitline.Timeline(this.conf);
```
NiklasNeugebauer, line 25: seems like you're not importing the zeitline.min.bundle.js anywhere so i'd assume this is undefined?

Aluriak: Isn't the resource already added with `self.add_resource(Path(__file__).parent / 'lib' / 'zeitline') ` ?

I tried anyway, but the problem remains the same ; i got a `Uncaught SyntaxError: unexpected token: identifier` for the line `async mounted() {` in the javascript console log.

I will look into that and get back to you

falkoschindler: The syntax error is cause by a missing `}` in the `data()` definition above.

Regarding the resource: You also have to load it on the client side. See the fullcalendar example for a very similar case:
https://github.com/zauberzeug/nicegui/blob/ab927959cd1250815a99943142137b61606d1565/examples/fullcalendar/fullcalendar.js#L11

Aluriak: Okay, it was the syntax problem and the missing imports. I copied from codemirror.js and now i've got a problem : the `zeitline.bundle.min.js` import is failing because of an undefined statement.

The source code of zeitline starts with :  (deminimization and debugging with firefox)

```
!function (t, n) {
  'object' == typeof exports &&
  'object' == typeof module ? module.exports = n() : 'function' == typeof define &&
  define.amd ? define([], n) : 'object' == typeof exports ? exports.Zeitline = n() : t.Zeitline = n()
}(
```

And the `TypeError: t is undefined` is raised at the 4th line, so i guess somehow `n() : t.Zeitline = n()` does not know about the function parameter `t`. Looks magic to me, but i guess either it's because my zeitline.bundle.min.js is malformed, either i'm using it wrong.

Aluriak: > The syntax error is caused by a missing `}` in the `data()` definition above.
> 
> Regarding the resource: You also have to load it on the client side. See the fullcalendar example for a very similar case:
> 
> https://github.com/zauberzeug/nicegui/blob/ab927959cd1250815a99943142137b61606d1565/examples/fullcalendar/fullcalendar.js#L11

Yeah, i just found that too :)

However, i copied the client-side import from codemirror.js. Can you verify it's valid too ?

xaptronic:
Zeitline needs an svg element to render into, and also has a missing stylesheet.

image

Would it make more sense to do this as a 3p extension?

falkoschindler:
@xaptronic You mean moving `ui.zeitline` in a separate repo/package? Sure, this might be reasonable.

xaptronic:
WIP here: https://github.com/xaptronic/nicegui-zeitline

I am going to try and get event handlers working before pushing to pypi.

Issue #3564, opened 2024-08-21
Copy button on ui.code does not work


JeppeKlitgaard:
### Description

On `nicegui==1.4.35` pressing the copy button on a `ui.code(...)` block does not work and throws the following error in the browser console (tested on Brave and Edge on Windows):

```
TypeError: Cannot read properties of undefined (reading 'writeText')
    at eval (eval at <anonymous> (nicegui.js:174:17), <anonymous>:1:27)
    at vue.global.prod.js:6:62901
    at Lt (vue.global.prod.js:1:14656)
    at jt (vue.global.prod.js:1:14734)
    at jt (vue.global.prod.js:1:14835)
    at cn (vue.global.prod.js:1:16637)
    at x (quasar.umd.prod.js:6:34300)
    at k (quasar.umd.prod.js:6:43565)
    at Lt (vue.global.prod.js:1:14656)
    at jt (vue.global.prod.js:1:14734)
```

falkoschindler:
Oh wow, thanks for reporting this issue, @JeppeKlitgaard!

It seems like the JavaScript API `navigator.clipboard` is not always available, only in secure contexts like HTTPS or localhost.

- [ ] Since there doesn't seem to be a workaround, we should hide the copy button in non-secure cases.
- [ ] Furthermore, we should fix `ui.clipboard` to warn the user and the developer about a non-existing clipboard API.

JeppeKlitgaard:
Thank you for replying so quickly! I wish this worked over HTTP as that's what I am running for internal use, but I understand that this is simply a browser limitation

PR #3636, opened 2024-08-29
Improve copy-to-clipboard button and `ui.clipboard` for non-secure environments

falkoschindler:
This PR solves #3564 by improving the clipboard behavior for non-secure environments where the clipboard API is not available.

- It hides the copy button if there is no `navigator.clipboard`. By setting `display: none` right on the client, we don't need to await any JavaScript call, which would not be allowed on the auto-index page.
- It adds documentation and shows warnings in Python and JavaScript when reading or writing the clipboard with the clipboard API being unavailable.
- It also improves the general event handling by preventing an exception if an awaited JavaScript command doesn't return anything (i.e. `undefined`).

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -282,7 +282,7 @@ class Client:
 
     def handle_javascript_response(self, msg: Dict) -> None:
         """Store the result of a JavaScript command."""
-        JavaScriptRequest.resolve(msg['request_id'], msg['result'])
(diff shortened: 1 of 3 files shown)


rodja (approved):

rodja:
Code looks fine. But we should figure out why the darkmode tests fails on main...

falkoschindler:
@rodja I'm working on it. See PR #3638.

PR #3660, opened 2024-08-31
Hide bottom space for error message when validation is `None`

falkoschindler:
PR #3446 added a bottom space for the error message of validation messages. This turned out to be unnecessary when no validation is used. This PR changes the behavior to only reserve the bottom space if the `validation` property is not `None`.

Code for experimenting with the new implementation:

```py
input_ = ui.input(value='Hello, world!').classes('border')

def add_validation():
    input_.validation = {
        'Must start with "Hello"': lambda x: x.startswith('Hello'),
        'Must end with "world!"': lambda x: x.endswith('world!')
    }

def remove_validation():
    input_.validation = None

ui.button('Add validation', on_click=add_validation)
ui.button('Remove validation', on_click=remove_validation)
```

### DIFF
--- a/nicegui/elements/input.py
+++ b/nicegui/elements/input.py
@@ -46,7 +46,7 @@ class Input(ValidationElement, DisableableElement, component='input.js'):
         :param password_toggle_button: whether to show a button to toggle the password visibility (default: False)
         :param on_change: callback to execute when the value changes
         :param autocomplete: optional list of strings for autocompletion
-        :param validation: dictionary of validation rules or a callable that returns an optional error message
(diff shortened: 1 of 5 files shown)


rodja (approved):

ghbm-itk:
@falkoschindler I know I'm late to the party, but may I ask why the validation setter was changed to automatically validate? It seems like an unnecessary side effect.

falkoschindler:
@ghbm-itk It's been a while and I don't remember the exact reasoning. But to me it feels right to re-validate if you assign a new validation function or dictionary. Like on value change the validation usually runs automatically.

PR #3679, opened 2024-09-05
add debug logs for on air reconnects

rodja:
We saw sporadic problems with reconnects to the On Air server. This PR adds some debug logging to better see what is going on. It also ensure we disconnect from all streams and sets a disconnect timeout so we do not get stuck there.

### DIFF
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -178,8 +178,8 @@ class Air:
             self.log.debug('Already connecting.')
             return
         if self.relay.connected:
-            self.log.debug('Already connected.')
             return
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
Looks good. I just wonder if logging "Already connected." every 5 seconds is a bit too much. Or do you think it's up to the user to ignore debug messages except for short intervals?

rodja:
You are right @falkoschindler. It might be a bit too much. I changed it so that we see a log message in every other case. That should reduce the overall messages and still give us informations about every state.

PR #3689, opened 2024-09-06
Introduce proper download simulation for the user fixture

rodja:
Improving upon #3687 to solve #3686 this PR adds a `user.download` to access the list of performed downloads.

### DIFF
--- a/nicegui/testing/user.py
+++ b/nicegui/testing/user.py
@@ -33,6 +33,7 @@ class User:
         self.forward_history: List[str] = []
         self.navigate = UserNavigate(self)
         self.notify = UserNotify()
+        self.http_responses: List[httpx.Response] = []
 
     def __getattribute__(self, name: str) -> Any:
         if name not in {'notify', 'navigate'}:  # NOTE: avoid infinite recursion
@@ -43,6 +44,7 @@ class User:
     async def open(self, path: str, *, clear_forward_history: bool = True) -> Client:
         """Open the given path."""
         response = await self.http_client.get(path, follow_redirects=True)
+        self.http_responses.append(response)
         assert response.status_code == 200, f'Expected status code 200, got {response.status_code}'
         if response.headers.get('X-Nicegui-Content') != 'page':
             raise ValueError(f'Expected a page response, got {response.text}')
@@ -51,6 +53,7 @@ class User:
         assert match is not None
         client_id = match.group(1)
         self.client = Client.instances[client_id]
+        self.client.outbox._emit = self._on_emit  # type: ignore # NOTE: monkey-patch the outbox
         self.sio.on('connect')
         await _on_handshake(f'test-{uuid4()}', {'client_id': self.client.id, 'tab_id': str(uuid4())})
         self.back_history.append(path)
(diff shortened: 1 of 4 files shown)


nicegui/testing/user.py
```diff
@@ -33,14 +34,16 @@ def __init__(self, client: httpx.AsyncClient) -> None:
...
         self.navigate = UserNavigate(self)
         self.notify = UserNotify()
+        self.download = UserDownload(self)
 
     def __getattribute__(self, name: str) -> Any:
-        if name not in {'notify', 'navigate'}:  # NOTE: avoid infinite recursion
+        if name not in {'notify', 'navigate', 'download'}:  # NOTE: avoid infinite recursion
             ui.navigate = self.navigate
             ui.notify = self.notify
+            ui.download = self.download
```
marcuslimdw, line 43: How does this work? Does it monkeypatch the `ui` module?

If so, I'm wondering - what happens if the user does e.g. `from ui import download`? Wouldn't that break testing with `User`?

website/documentation/content/user_documentation.py
```diff
@@ -156,6 +156,40 @@ def receive_file(e: events.UploadEventArguments):
             ''')
 
 
+doc.text('Test Downloads', '''
+    You can verify that a download was triggered by checking `user.downloads.http_responses`.
+    By awaiting `user.downloads.next()` you can get the next download response.
```
marcuslimdw, line 161: This says `next`, but the method's name is "new"...?

(personally, I like "next")

nicegui/testing/user_download.py
```diff
@@ -0,0 +1,44 @@
...
+
+    async def _get(self,  src: Union[str, Path, bytes]) -> None:
+        if isinstance(src, bytes):
+            await asyncio.sleep(0)
+            response = httpx.Response(200, content=src)
+        else:
+            response = await self.user.http_client.get(str(src))
+        self.http_responses.append(response)
+
+    async def new(self) -> httpx.Response:
```
marcuslimdw, line 32: I wonder - what's the reasoning behind this returning a `httpx.Response`, as opposed to some other object that just encapsulates e.g. the filename and contents?

nicegui/testing/user_download.py
```diff
@@ -0,0 +1,44 @@
...
+        self.http_responses: List[httpx.Response] = []
+        self.user = user
+
+    def __call__(self, src: Union[str, Path, bytes], filename: Optional[str] = None, media_type: str = '') -> Any:
+        background_tasks.create(self._get(src))
+
+    async def _get(self,  src: Union[str, Path, bytes]) -> None:
+        if isinstance(src, bytes):
+            await asyncio.sleep(0)
+            response = httpx.Response(200, content=src)
```
marcuslimdw, line 27: Do you want to consider using `HTTP_200_OK`?

marcuslimdw (commented):
Thank you for working on this!

website/documentation/content/user_documentation.py
falkoschindler: Yes, we agree. I was just about to commit a fix. 👍🏻

falkoschindler (approved):

nicegui/testing/user_download.py
rodja: A custom encapsulation would require documentation and maintenance. And because the `Response` object has a fairly nice API and allows to even check for response headers etc. it is good suited for tests.

nicegui/testing/user_download.py
rodja: Sure!

nicegui/testing/user.py
rodja: You are right. The productive code must use `ui.download` for this to work. This fails:

```py
from nicegui import ui
from nicegui.ui import download

async def test_download(user: User):
    ui.button('get', on_click=lambda: download(b'some data'))

    await user.open('/')
    user.find('get').click()
    response = await user.download.next()
    assert response.text == 'some data'
```

I think we should not solve this in this PR. Other features like `ui.notify` or `ui.navigate` have the same problem. Maybe you have an idea on how we could improve this? Could you create a feature request?

PR #3697, opened 2024-09-08
Add dark_page props to ui.colors

python-and-fiction:
Add dark_page props to ui.colors, so we can change the color of page in dark mode.


falkoschindler (approved):
Thanks for the pull request, @python-and-fiction!
I just simplified the `setProperty` call a bit, so that we don't set the non-existing "dark_page" property.
And while at it, I added missing type annotation and the parameter list in the docstring.
Let's merge!

PR #3698, opened 2024-09-08
Improve error message when client is not yet initialised

marcuslimdw:
`User.client` is only set after `open` is called. Since having a valid `client` instance is necessary to test anything, whether `client` is set is checked in most methods. However, this is done with a simple `assert self.client`, which gives an opaque `AssertionError` if the user has indeed forgotten to call `open`. This PR improves the error message to specifically call that out.

### DIFF
--- a/nicegui/testing/user.py
+++ b/nicegui/testing/user.py
@@ -95,7 +95,7 @@ class User:
         By default `should_see` makes three attempts to find the element before failing.
         This can be adjusted with the `retries` parameter.
         """
-        assert self.client
+        self._assert_client()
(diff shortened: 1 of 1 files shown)


rodja (changes requested):
That's great @marcuslimdw. I like showing a better error message. Unfortunately your modification breaks the mypy type checking. Maybe you can modify it to make `client` a property which can raise the error?

falkoschindler (approved):
I quickly replaced the function with a new `_client` property that returns a (non-optional) `Client`. Now mypy is happy again.

PR #3699, opened 2024-09-08
Handle type error when navigating

marcuslimdw:
In the event that `target` does not satisfy the type hint for `to`, the user will get an `UnboundLocalError` for `path`, which is confusing. This PR explicitly handles the case where `to` gets an input of an unexpected type.

### DIFF
--- a/nicegui/functions/navigate.py
+++ b/nicegui/functions/navigate.py
@@ -60,6 +60,8 @@ class Navigate:
             path = f'#c{target.id}'
         elif callable(target):
             path = Client.page_routes[target]
+        else:
+            raise TypeError(f"Expected a str path, an Element, or a page builder, but got {target}")
         context.client.open(path, new_tab)
 
 
--- a/nicegui/functions/navigate.py
+++ b/nicegui/functions/navigate.py
@@ -61,7 +61,7 @@ class Navigate:
         elif callable(target):
             path = Client.page_routes[target]
         else:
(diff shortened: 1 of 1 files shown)


rodja:
Thanks for the PR, @marcuslimdw. This is only one special case. If we implement this, we should do it for the entire public API. And that is quite a lot of code which needs to be written, tested and maintained. Maybe it's better to rely on static type checkers like mypy to catch such errors.

marcuslimdw:
@rodja I agree that, in theory, all projects should incorporate static type checkers, because what can be checked before runtime should be checked before runtime.

From a user experience perspective, IMO, it is likely that sooner or later, a user will run into this (which is basically a pattern matching failure) and be confused by the `UnboundLocalError`, so the question is - is the maintenance burden worth the easier debugging?

Lastly, it's apposite to note that even in the stdlib, there is a [function provided for this exact case](https://docs.python.org/3/library/typing.html#typing.assert_never): `assert_never`, which might be better than a custom error.

falkoschindler (approved):
Thanks for the pull request, @marcuslimdw!

We came to the conclusion that raising a TypeError in an else block is a good idea. While it improves the error message for the user, it also improves code readability, since it is a clear indication that `path` will be defined after the if-elif-else block. I just shortened the error message a little to avoid the risk of it getting outdated in case we change the set of supported types in the future.

There are lots of other places where wrong input types lead to weird exceptions. But we won't be able (and willing) to address all of them. This issue is a great example for the value of static type checkers and linters like mypy and pylint.

PR #3703, opened 2024-09-09
Correct doc of  `run_chart_method`.

python-and-fiction:
`Run a method of the JSONEditor instance.` in https://nicegui.io/documentation/echart#methods should be `Run a method of the EChart instance.`


falkoschindler (approved):
Thanks for the quick pull request, @python-and-fiction!

PR #3708, opened 2024-09-10
Custom colors prop support

Kolterdyx:
Implements a feature I just suggested [here](https://github.com/zauberzeug/nicegui/discussions/3707#discussion-7163211).

It allows to add custom colors definition for branding/theming using the `ui.colors` function. It can be used as follows:

```python
from nicegui import ui

ui.colors(my_custom_color='#ff0')
ui.button('Custom color', color='my-custom-color')

ui.run()
```

The javascript added to make this work probably isn't optimal, but it works as intended, so that's probably what needs to be reviewed the most

### DIFF
--- a/nicegui/elements/colors.js
+++ b/nicegui/elements/colors.js
@@ -1,6 +1,20 @@
 export default {
   mounted() {
     for (let name in this.$props) {
+      if (name === "customColors") {
+        let customCSS = "";
+        for (let customColor in this.customColors) {
+          const colorName = customColor.replaceAll("_", "-");
+          const colorVar = "--q-" + colorName;
+          document.body.style.setProperty(colorVar, this.customColors[customColor]);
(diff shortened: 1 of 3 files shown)


rodja:
Thanks @Kolterdyx. Great idea. @falkoschindler will look at it after coming back from his well deserved vacation.

rodja:
Thanks @Kolterdyx. Great idea. @falkoschindler will look at it after coming back from his well deserved vacation.

falkoschindler (approved):
Thanks for the pull request, @Kolterdyx! That's a neat addition.

I just
- refactored the JavaScript for-loop a little,
- made sure to _replace_ the custom style tag to avoid growing the DOM when calling `ui.colors` repeatedly,
- added `custom_colors` to the docstring, and
- simplified the demo a bit.

Let's merge!

PR #3736, opened 2024-09-16
Fix for ChoiceElement duplicate `self.update()` calls and basic ChoiceElement unit testing

### LINKED ISSUE #3733: Current implementation of ChoiceElement.set_options() calls self.update() twice if a value is set, causing bugs in code that relies on data coming from `on_changed()` to be accurate.
### Description

Just ran into this issue while working with the `ui.radio()` element. When I call set_options() on the radio element with a value set, the `on_changed()` callback is triggered twice due to the duplicated `self.update()` calls within the `set_options()` function. Not only is it strange to call update() twice for no reason, it causes bugs. If the current value of the radio is not within the new set of options, it will call `on_changed()` once with a `None`, and then a second time with the correct value. This causes seemingly impossible NoneType errors downstream that now need to be accounted for.

Fortunately, this is incredibly simple to fix.

Original code from `nicegui/elements/choice_element.py > ChoiceElement() > set_options()`:

```py
def set_options(self, options: Union[List, Dict], *, value: Any = ...) -> None:
    """Set the options of this choice element.

    :param options: The new options.
    :param value: The new value. If not given, the current value is kept.
    """
    self.options = options
    self.update()
    if value is not ...:
        self.value = value
        self.update()
```

Fixed code:

```py
def set_options(self, options: Union[List, Dict], *, value: Any = ...) -> None:
    """Set the options of this choice element.

    :param options: The new options.
    :param value: The new value. If not given, the current value is kept.
    """
    self.options = options
    if value is not ...:
        self.value = value
    self.update()
```

If someone specifically _wanted_ this functionality for whatever reason, they could call `set_options()` and `set_value()` separately, achieving the duplicated call regardless, so nothing is lost by making this change.

sSasha-uwu:
Resolves #3733

Added unit testing as requested. Wasn't sure how ham to go exactly, but I just went ahead and added tests for all the functionality that I've used with the `ui.radio()` element before.

### DIFF
--- a/tests/test_choice_element.py
+++ b/tests/test_choice_element.py
@@ -1,4 +1,4 @@
-from nicegui import ui
+from nicegui import events, ui
 from nicegui.testing import Screen
 
 
@@ -70,5 +70,18 @@ def test_radio_set_options_value(screen: Screen):
     screen.open('/')
     r.set_options(['D', 'E', 'F'], value='E')
     assert r.value == 'E'
+
+# - When using `set_options` with the value argument set and the `on_change` callback active on the element,
+# `on_change` should never pass `None` through, even if the old value is not within the new list of element options
+
+def test_radio_set_options_value_callback(screen: Screen):
+
+    def check_change_is_not_none(e: events.ValueChangeEventArguments):
+        assert e.value is not None
+
+    r = ui.radio(['A', 'B', 'C'], on_change=check_change_is_not_none)
+
+    screen.open('/')
+    r.set_options(['D', 'E', 'F'], value='F')
 
 ###
--- a/tests/test_choice_element.py
+++ b/tests/test_choice_element.py
@@ -1,0 +1,74 @@
+from nicegui import ui
+from nicegui.testing import Screen
+
+
+###
+
+# Tests Needed:
+
+# - Radio options should be clickable, and the value should be set to the clicked option
+
+def test_radio_click(screen: Screen):
+    r = ui.radio(['A', 'B', 'C'])
+
+    screen.open('/')
+    screen.click('A')
+    assert r.value == 'A'
+    screen.click('B')
+    assert r.value == 'B'
+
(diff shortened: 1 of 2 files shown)


tests/test_choice_element.py
```diff
@@ -0,0 +1,87 @@
+from nicegui import events, ui
+from nicegui.testing import Screen
+
+
+###
```
rodja, line 5: Why this ### comment?

tests/test_choice_element.py
```diff
@@ -0,0 +1,87 @@
...
+
+# Tests Needed:
+
+# - Radio options should be clickable, and the value should be set to the clicked option
```
rodja, line 9: I think you can remove this explanations. The test function name should describe what is being tested.

tests/test_choice_element.py
```diff
@@ -0,0 +1,87 @@
...
+# - When using `set_options` with the value argument set, the new value should be properly set
+
+def test_radio_set_options_value(screen: Screen):
+    r = ui.radio(['A', 'B', 'C'])
+
+    screen.open('/')
+    r.set_options(['D', 'E', 'F'], value='E')
+    assert r.value == 'E'
+
+# - When using `set_options` with the value argument set and the `on_change` callback active on the element,
```
rodja, line 74: This might actually be better a docstring of the test. And please include the link to the issue.

tests/test_choice_element.py
rodja, line 1: It's fine to use the radio element for testing the choice mixin, but then I suggest you add the testes to `test_radio_element.py`

rodja (changes requested):
Thanks for adding the tests @sSasha-uwu. I just made a few comments but over all this improves the coverage quite a lot.

sSasha-uwu:
Comments were my bad, I forgot to remove them.

rodja (approved):

PR #3750, opened 2024-09-19
Simplify clearing of values in pytests with simulated user interaction

rodja:
This PR implement the feature request from #3747 by introducing a `clear` method for the `UserInteraction` class. Usage:

```py
user.find(ui.input).clear() # empties the content of an input element
```

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -6,6 +6,7 @@ from typing_extensions import Self
 
 from nicegui import background_tasks, events, ui
 from nicegui.element import Element
+from nicegui.elements.mixins.value_element import ValueElement
 
 if TYPE_CHECKING:
     from .user import User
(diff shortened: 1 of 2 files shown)


MasterSergius (approved):
Looks good

nicegui/testing/user_interaction.py
```diff
@@ -75,3 +78,14 @@ def click(self) -> Self:
...
+
+    def clear(self) -> Self:
+        """Clear the selected elements.
+
+        Note: All elements must have a ``value`` attribute)."""
+        assert self.user.client
+        with self.user.client:
+            for element in self.elements:
+                assert isinstance(element, (ValueElement))
+                element.value = ''
```
falkoschindler, line 92:
```suggestion
                element.value = None
```
I wonder if `None` would be the more natural value for clearing a value element. That's the value assigned when using the "clearable" prop on Quasar inputs. And it's the initial default value for `ui.number` (where an empty string "" is not applicable). But the default for `ui.input` is "", so it's not 100% obvious what `clear()` is supposed to do. For the sake of user tests the difference between `None` and "" might be irrelevant though. For compatibility with `ui.number` I'd vote for `None`.

PR #3799, opened 2024-09-28
Add test to ensure we can teleport to a table cell

rodja:
Inspired by #3794 this PR adds a test which uses `ui.teleport` to make a cell value editable with a `ui.select` element.


falkoschindler:
As it turns out, this test doesn't test if the `ui.select` is actually teleported. And it's not so easy to fix. Therefore we'll close it without merging.

PR #3811, opened 2024-09-30
Add handler type alias and use it in event handlers

marcuslimdw:
Addresses https://github.com/zauberzeug/nicegui/discussions/3797.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -4,20 +4,7 @@ import inspect
 import re
 from copy import copy
 from pathlib import Path
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    ClassVar,
-    Dict,
-    Iterator,
-    List,
-    Optional,
-    Sequence,
-    Union,
-    cast,
-    overload,
-)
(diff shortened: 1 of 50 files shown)


marcuslimdw:
@falkoschindler, I'm wondering about absolute/relative imports - it seems that the NiceGUI convention is to always use relative imports (though there were some imports that were absolute, which was why I defaulted to absolute).

specifically, sibling relative imports (`from .sibling import obj`) are pretty understandable, but when you get to stuff like `from ... import thing`, it feels a little ridiculous IMO

marcuslimdw:
PEP 8 does, and IMO point 1 doesn't really apply when an IDE can do that for you

that said, I feel like consistency is more important - it might be good to put this in `CONTRIBUTING.md`

falkoschindler (approved):
Thanks for the pull request, @marcuslimdw!

Apart from quickly replacing absolute imports with relative ones, I
- fixed a confusion between select and change events for `ui.json_editor`,
- added casting to stop mypy from complaining, and
- added the new `Handler` to the generic event registrations `.on()`.

Let's merge! 🙂

Issue #3809, opened 2024-09-29
Validation Element Error


GMezWheel:
### Description

Hi,
Sometimes I have this error :
```
File "/usr/local/lib/python3.10/dist-packages/nicegui/elements/mixins/validation_element.py", line 43, in error
    if self._error == error and self._props['error'] == new_error_prop:
KeyError: 'error'
```

Perhaps check that the key 'error' exists in the 'self._props' dictionary:
```py
if self._error == error and 'error' in self._props and self._props['error'] == new_error_prop:
```

falkoschindler:
Hi @GMezWheel,

The "error" prop is set right in the initializer and never removed. So I don't immediately see a way this KeyError can occur.
Are you using the most recent version 2.2 of NiceGUI?
And can you, by any chance, provide a minimal reproducible example so we can investigate further? Thanks!

GMezWheel:
No, I'm using the previous version of nicegui : 2.1.0.dev0

I'll upgrade to version 2.2 and let you know if the problem recurs.

Thanks for your quick help.

kevinraymond:
I ran into this yesterday for the first time, on 2.2.0. Not sure what element OP is using, but mine is a select element. 

When it has `value=""` or `value=1`, for example, instead of `value=None` or a valid options label (e.g., 'Apple'), it's failing in the `ValidationElement.__init__` (probably in the super call to `ValueElement`. This causes the error prop to never be set.

Here's an example showing all three with a select element. I didn't try with any other input.

```
import traceback
from nicegui import ui

# fails
try:
    ui.select(
        label="Choose a fruit",
        options=["Apple", "Banana", "Cherry", "Date"],
        value=""
    )
except Exception as e:
    print("======================================")
    print("Empty string error:")
    traceback.print_exc()
    

# fails
try:
    ui.select(
        label="Choose a fruit",
        options=["Apple", "Banana", "Cherry", "Date"],
        value=1
    )
except Exception as e:
    print("======================================")
    print("Integer value error:")
    traceback.print_exc()

# works
ui.select(
    label="Choose a fruit",
    options=["Apple", "Banana", "Cherry", "Date"],
    value=None
)

ui.run()

# one potential option to fix is changing validation_element.py:
# if self._error == error and self._props['error'] == new_error_prop:
# to
# if self._error == error and self._props.get('error') == new_error_prop:
```

arshad-muhammad:
[...]

PR #3829, opened 2024-10-04
Check value of choice elements against options

falkoschindler:
This PR raises a more readable exception when initializing a choice element with an invalid value, as discussed in #3809:
```py
ui.select(['Apple', 'Banana'], value='Cherry')
```

- [x] fix tests
- [x] add test for new behavior


rodja:
We should also add a user test to verify the behaviour.

rodja (approved):

Issue #3826, opened 2024-10-04
ui.link.default_classes(replace) doesn't replace the nicegui-link class


ghbm-itk:
### Description

When setting default classes with replace on a ui.link the link will still have the nicegui-link class.

```python
from nicegui import ui

ui.link.default_classes(replace='text-lg border-r-2')
ui.link("Link text", "/nowhere")

ui.link("Link text2", "/nowhere2").classes(replace='text-lg border-r-2')

ui.run()
```

python-and-fiction:
Do you want that link's style looks like normal text?

python-and-fiction:
```python3
from nicegui import ui
from typing import Any, Callable, Union
from nicegui.element import Element

ui.link.default_classes(replace='text-lg border-r-2')


class link_ext(ui.link):
    def __init__(self,
                 text: str = '',
                 target: Union[Callable[..., Any], str, Element] = '#',
                 new_tab: bool = False,
                 ) -> None:
        super().__init__(text=text, target=target, new_tab=new_tab)
        self._classes.remove('nicegui-link')


ui.link = link_ext

ui.link("Link text", "/nowhere")

ui.link("Link text2", "/nowhere2").classes(replace='text-lg border-r-2')

ui.run()
```

ghbm-itk:
Thanks, I know how to remove the class.
I still think it's a bug

falkoschindler:
Hi @ghbm-itk, thanks for reporting this issue!
The inconsistency is indeed unexpected. It basically boils down to this:

```py
ui.link.default_classes(remove='nicegui-link')
ui.link("Link 1", "/one")  # is still underlined

ui.link("Link 2", "/two").classes(remove='nicegui-link')  # is not underlined
```

We are, however, unsure how to fix it properly. We'll think about it.

falkoschindler:
PR #3843 should solve this issue by adding "nicegui-*" classes already when defining an element class, not only when initializing it. This way these classes can be manipulated via `.default_classes()`.

PR #3843, opened 2024-10-08
Define default classes while subclassing

falkoschindler:
This PR introduces `default_classes`, `default_style` and `default_props` arguments for `Element.__init_subclass__` so that defaults like "nicegui-*" classes can be defined while subclassing. This way they can be removed by the user calling something like `ui.link.default_classes(replace='text-red-500')`, solving issue #3826.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -84,6 +84,9 @@ class Element(Visibility):
                           libraries: List[Union[str, Path]] = [],  # noqa: B006  # DEPRECATED
                           exposed_libraries: List[Union[str, Path]] = [],  # noqa: B006  # DEPRECATED
                           extra_libraries: List[Union[str, Path]] = [],  # noqa: B006  # DEPRECATED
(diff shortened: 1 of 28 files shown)


ghbm-itk:
Isn't this technically a breaking change or can it be defined as a bugfix? Some users might be reliant on the existing behavior

falkoschindler:
@ghbm-itk That's a good question. At the moment even the pytests are broken. I'll try to fix them first. Then we should evaluate how exactly the behavior changed and if we consider it a bug fix or a breaking change.

falkoschindler:
Ok, now that the tests are passing, we need to decide whether this is a bugfix or breaking change.

- It is probably a rare case that someone is already using the `replace=` argument for default classes, since it was basically ineffective.
- And if someone _does_ use it, removing _all_ classes is the expected result. Therefore I think this PR is a bugfix.

PR #3848, opened 2024-10-08
Allow customizing the crosshair in `ui.interactive_image`

parlance-zz:
These new props enable or disable drawing the corresponding line when cross is enabled.

The motivation in my case was the need to show only the vertical crosshair line in an audio editing context (in this instance the audio is represented as an interactive_image element of a spectrogram).

Simple demo of the functionality:

```py
from nicegui import ui

image = ui.interactive_image(source='https://picsum.photos/id/29/640/360', cross=True)

def toggle_prop(prop: str) -> None:
    image._props[prop] = not image._props[prop]
    image.update()

ui.button("toggle vertical", on_click=lambda: toggle_prop("cross_vertical"))
ui.button("toggle horizontal", on_click=lambda: toggle_prop("cross_horizontal"))

ui.run()
```

### DIFF
--- a/nicegui/elements/interactive_image.js
+++ b/nicegui/elements/interactive_image.js
@@ -11,9 +11,10 @@ export default {
         draggable="false"
       />
       <svg ref="svg" style="position:absolute;top:0;left:0;pointer-events:none" :viewBox="viewBox">
-        <g v-if="cross" :style="{ display: showCross ? 'block' : 'none' }">
-          <line :x1="x" y1="0" :x2="x" y2="100%" :stroke="cross === true ? 'black' : cross" />
-          <line x1="0" :y1="y" x2="100%" :y2="y" :stroke="cross === true ? 'black' : cross" />
+        <g :style="{ display: showCross ? 'block' : 'none' }">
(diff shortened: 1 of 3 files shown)


parlance-zz:
As this is a very small change I've skipped over some of the requirements in the contribution guide. If any changes are required please let me know and I will amend the request.

falkoschindler (commented):
Thanks for the pull request, @parlance-zz! We like the idea and the code looks perfectly fine.

We have, however, concerns about adding more and more parameters for configuring the crosshair. As an alternative I experimented with adding a custom "cross" slot in PR #3870. What do you think? If you agree with this solution, I'm going to merge my changes into your PR so that your contribution is still officially recognized.

parlance-zz:
> Thanks for the pull request, @parlance-zz! We like the idea and the code looks perfectly fine.
> 
> 
> 
> We have, however, concerns about adding more and more parameters for configuring the crosshair. As an alternative I experimented with adding a custom "cross" slot in PR #3870. What do you think? If you agree with this solution, I'm going to merge my changes into your PR so that your contribution is still officially recognized.

Looks great! Thank you for all your work on this library.

falkoschindler (approved):
Ok, I replaced your implementation with mine from PR #3870.
Let's merge!

PR #3853, opened 2024-10-09
Expose additional PyInstaller arguments

pikaro:
The optional added arguments are useful for building on OSX.

It would probably make sense to expose all arguments of PyInstaller as far as they don't conflict, although this would require manually respecifying them all.

(Alternatively, you could also provide documentation on how to package with PyInstaller if the user prefers not to use the wrapper.)


falkoschindler (approved):
Thanks for the pull request, @pikaro!
Let's merge!

PR #3866, opened 2024-10-11
Run dynamic argument conversion when using run_grid_method / run_row_method

xaptronic:
Following up with https://github.com/zauberzeug/nicegui/pull/3757

Do dynamic property conversion when using `run_grid_method` / `run_row_method`, so you can dynamically add / change callbacks and other things using properties with `:` prefix.

### DIFF
--- a/nicegui/elements/aggrid.js
+++ b/nicegui/elements/aggrid.js
@@ -46,9 +46,11 @@ export default {
       this.api.addGlobalListener(this.handle_event);
     },
     run_grid_method(name, ...args) {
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):

I successfully tested it with this snippet:
```py
grid = ui.aggrid({'columnDefs': [{'field': 'name'}], 'rowData': [{'name': 'Alice'}]})
ui.button('Update', on_click=lambda: grid.run_grid_method('setGridOption', 'rowData', [{':name': '"Bob"'}]))
```

Let's merge!

PR #3868, opened 2024-10-12
doc storage small update re: duplicated tabs

kfalcami:
added small extra text to indicate that app.storage.tab is only unique across non-duplicated tabs

---

Related discussion: #3867


falkoschindler (approved):
Thanks for the pull request, @kfalcami!
It's a helpful addition to the documentation, even though we're thinking about fixing this unexpected behavior in #3872.

Issue #3872, opened 2024-10-14
`app.storage.tab` is not unique for duplicate tabs


falkoschindler:
### Discussed in https://github.com/zauberzeug/nicegui/discussions/3867

<div type='discussions-op-text'>

<sup>Originally posted by **kfalcami** October 11, 2024</sup>
### Question

I did a quick experiment to test app.storage.tab storage (below), running on localhost.  Every new tab that connects prints "new!..." with a unique value, but any tab I create with "Duplicate" in Chrome (or "Duplicate Tab" in Edge) prints "already..." with the same value as the duplicated tab.  This isn't mentioned on the [storage page](https://nicegui.io/documentation/storage#storage), not sure if this is a bug or expected behavior?  Is there any way to get unique storage for Duplicate tabs?

```py
@ui.page('/')
    async def index(request: Request) -> None:
        await ui.context.client.connected()
        chat_exchanges_key = 'chat_exchanges'
        if chat_exchanges_key in app.storage.tab:
            print(f'~~~ already set! {app.storage.tab[chat_exchanges_key]}')
        else:
            x = randint(1, 99)
            app.storage.tab[chat_exchanges_key] = x
            print(f'~~~ new! setting to {x}')
```
</div>

---

It turns out NiceGUI's tab storage behaves differently than the JavaScript session storage:

> The contents of sessionStorage is "copied" not "shared". Any changes made to sessionStorage after the tab is duplicated are not reflected in the other tab's sessionStorage.
https://stackoverflow.com/questions/21463696/duplicate-tab-copying-sessionstorage-in-chrome

The tab storage is copied _and_ shared. We should try to detect the duplication, create a new tab ID, copy the storage and continue working on the new instance with the new tab ID.

PR #3876, opened 2024-10-15
Improve generation of unique tab IDs

falkoschindler:
This PR improves the way tab IDs are generated to avoid duplicated tabs causing shared tab storage (#3872).

The ides originates from this post on StackOverflow: https://stackoverflow.com/a/36807854/3419103

I successfully tested it with this code snippet:
```py
@ui.page('/')
async def main():
    if 'id' not in app.storage.tab:
        app.storage.tab['id'] = str(uuid.uuid4())
    ui.label(app.storage.tab['id'])
```

An automated pytest would be great. But I guess we can't properly simulate duplicating tabs.

### DIFF
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -130,6 +130,8 @@ class Air:
                 return False
             client = Client.instances[client_id]
             client.environ = data['environ']
+            if data.get('old_tab_id'):
+                core.app.storage.copy_tab(data['old_tab_id'], data['tab_id'])
             client.tab_id = data['tab_id']
             client.on_air = True
             client.handle_handshake()
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -167,6 +167,8 @@ async def _on_handshake(sid: str, data: Dict[str, str]) -> bool:
     client = Client.instances.get(data['client_id'])
     if not client:
         return False
+    if data['old_tab_id']:
+        app.storage.copy_tab(data['old_tab_id'], data['tab_id'])
     client.tab_id = data['tab_id']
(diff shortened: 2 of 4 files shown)


rodja:
Should we not copy the tab storage? For example if we have

```py
ui.select(['A', 'B', 'C']).bind_value(app.storage.tab, 'selection')
```

I would expect the duplicated tab to have the same selection. But then each should be able to edit independently. That isat least the way the browser session storage works...

falkoschindler:
@rodja You're absolutely right! I forgot this part of the equation...

falkoschindler:
Ok, now the tab storage is copied if a tab duplication is detected. Can be tested with this snippet:

```py
import random
from nicegui import app, ui

@ui.page('/')
async def main():
    ui.add_body_html('''
        <script>
        document.addEventListener('DOMContentLoaded', function () {
            const span = document.createElement('span');
            span.innerHTML = `Tab ID: <pre>${TAB_ID}</pre>`
            if (OLD_TAB_ID) span.innerHTML += `Cloned from: <pre>${OLD_TAB_ID}</pre>`
            span.classList.add('p-2', 'rounded', 'whitespace-pre-line', 'border', 'absolute-center');
            document.body.appendChild(span);
        });
        </script>
    ''')
    if 'number' not in app.storage.tab:
        app.storage.tab['number'] = random.randint(0, 1000000)
    ui.label(f'Number: {app.storage.tab["number"]}')

    ui.select(['A', 'B', 'C']).bind_value(app.storage.tab, 'selection')
```

rodja (approved):
Great!

Issue #2566, opened 2024-02-18
interactive image force_reload behavior


scottrcarlson:
### Description

This works perfectly fine:

```py
cam1 = ui.interactive_image(cam1_url)
ui.button('Refresh', on_click=cam1.force_reload).style('font-size: 150%; font-weight: 300')
```

However, if I were to execute `cam1.force_reload()` inside a function, or lambda, it does not seem do anything.  For example, i'd like to setup a `ui.timer()` to call a function that reloads the image. I've tried a few variations using `lambda`, `callback=` and even `ui.refreshable` to attempt to resolve. 

The only way `.force_reload` seems to function is when utilized in `on_click`.

Any advice or guidance would be greatly appreciated!

falkoschindler:
@scottrcarlson Refreshing an interactive image via `ui.timer` should work like this:

```py
img = ui.interactive_image('https://picsum.photos/640/360')
ui.timer(1.0, img.force_reload)
```

Maybe you can share your code to see what you're doing differently?

scottrcarlson:
That worked just fine.  I don't know how I didn't try that permutation!

Thank you!

scottrcarlson:
Resolved.

rp10007:
I have a slightly different version of this same question that may just be me misunderstanding what should happen: I have an interactive image that reloads fine when force_reload is called in its own mousehandler, but I would like it to also reload (with changes) when some buttons are pushed.
```
src = to_scan #image created by picam2.capture_image(), should be PIL form
ii = ui.interactive_image(src, on_mouse=mouse_handler, events=['mousedown', 'mouseup'], cross=True)

ui.button('Zoom', on_click=cropsrc)

def cropsrc():  
    global src
    print (int(ii.boxxpos), int(ii.boxypos),  int(ii.boxxpos+ii.boxwid), int(ii.boxypos + ii.boxht))
    src = to_scan.crop((int(ii.boxxpos), int(ii.boxypos),  int(ii.boxxpos+ii.boxwid), int(ii.boxypos + ii.boxht)))
    print("cropped")
    ii.force_reload()
```
(pls forgive the print statements; I'm coding this for a headless raspberry pi)
Anyway, when I click to do things inside the image, that works reasonably well if slowly. But if I click the Zoom button, I get the console printouts of my coordinates and a message that the cropping happened, but the image does not reload. 

Is the reload not happening because something about the cropped image is malformed, or for other reasons?

rodja:
[...]

PR #3895, opened 2024-10-18
Show warning when force_reload is used on data sources

rodja:
This PR attempts to help the user understand that `force_reload` only makes sense on network urls. See https://github.com/zauberzeug/nicegui/issues/2566#issuecomment-2421168006 more

### DIFF
--- a/nicegui/elements/image.py
+++ b/nicegui/elements/image.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Union
 
 from .. import optional_features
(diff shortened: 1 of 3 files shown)


falkoschindler (changes requested):
Thanks, @rodja. But there are two problems with the current solution:

- If `source` is a `Path` or `PIL_Image`, `startswith` is not defined.
- We should update `ui.image` accordingly.

falkoschindler (approved):

PR #3900, opened 2024-10-18
Introduce an `html` module with plain HTML elements

falkoschindler:
Following up on https://github.com/zauberzeug/nicegui/discussions/572#discussioncomment-10476200, this PR introduces an `html` module for creating plain HTML elements. They are basically shortcuts for `ui.element(<tag>)` and allow passing inner HTML as optional positional argument and arbitrary attributes as keyword arguments. Nesting, classes, styling and event subscriptions work like with regular NiceGUI elements. Here's a quick demo:

```py
from nicegui import html

with html.p():
    html.a('Link', href='https://google.com').classes('text-red-500')
    html.hr()
    html.button('Button').on('click', lambda: print('Hello World'))
    html.img(src='https://picsum.photos/400/300', width='200')
```

Open tasks:

- [x] Should we exclude "unusable" elements like `html`, `head`, and `body`? This might not be obvious in all cases.
- [x] Add it to the documentation.
- [x] Add pytests for some elements like `button`?


rodja (approved):
Great. I just added a small note and demo on how to use the props.

PR #3936, opened 2024-10-31
echart: enable svg rendering

Piguite:
add `enable_svg` option on echart to change chart renderer from `canvas `to `svg`

### DIFF
--- a/nicegui/elements/echart.js
+++ b/nicegui/elements/echart.js
@@ -9,7 +9,12 @@ export default {
       await import("echarts-gl");
     }
 
-    this.chart = echarts.init(this.$el);
+    if (this.enable_svg) {
+      this.chart = echarts.init(this.$el, null, {renderer: 'svg'});
+    } else {
+      this.chart = echarts.init(this.$el);
+    }
+    
     this.chart.on("click", (e) => this.$emit("pointClick", e));
     for (const event of [
       "click",
@@ -90,5 +95,6 @@ export default {
   props: {
     options: Object,
     enable_3d: Boolean,
+    enable_svg: Boolean
   },
 };
--- a/nicegui/elements/echart.py
+++ b/nicegui/elements/echart.py
@@ -22,7 +22,7 @@ class EChart(Element,
              dependencies=['lib/echarts/echarts.min.js', 'lib/echarts-gl/echarts-gl.min.js'],
              default_classes='nicegui-echart'):
 
-    def __init__(self, options: Dict, on_point_click: Optional[Handler[EChartPointClickEventArguments]] = None, *, enable_3d: bool = False) -> None:
+    def __init__(self, options: Dict, on_point_click: Optional[Handler[EChartPointClickEventArguments]] = None, *, enable_3d: bool = False, enable_svg: bool = False) -> None:
         """Apache EChart
 
         An element to create a chart using `ECharts <https://echarts.apache.org/>`_.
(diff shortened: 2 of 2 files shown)


nicegui/elements/echart.js
```diff
@@ -9,7 +9,12 @@ export default {
       await import("echarts-gl");
     }
 
-    this.chart = echarts.init(this.$el);
+    if (this.enable_svg) {
+      this.chart = echarts.init(this.$el, null, {renderer: 'svg'});
+    } else {
+      this.chart = echarts.init(this.$el);
+    }
+
```
falkoschindler, line 17:
```suggestion
    this.chart = echarts.init(this.$el, null, { renderer: this.renderer });
```

nicegui/elements/echart.js
```diff
@@ -90,5 +95,6 @@ export default {
   props: {
     options: Object,
     enable_3d: Boolean,
+    enable_svg: Boolean
```
falkoschindler, line 98:
```suggestion
    renderer: String,
```

nicegui/elements/echart.py
```diff
@@ -22,7 +22,7 @@ class EChart(Element,
              dependencies=['lib/echarts/echarts.min.js', 'lib/echarts-gl/echarts-gl.min.js'],
              default_classes='nicegui-echart'):
 
-    def __init__(self, options: Dict, on_point_click: Optional[Handler[EChartPointClickEventArguments]] = None, *, enable_3d: bool = False) -> None:
+    def __init__(self, options: Dict, on_point_click: Optional[Handler[EChartPointClickEventArguments]] = None, *, enable_3d: bool = False, enable_svg: bool = False) -> None:
```
falkoschindler, line 25:
```suggestion
    def __init__(self,
                 options: Dict,
                 on_point_click: Optional[Handler[EChartPointClickEventArguments]] = None, *,
                 enable_3d: bool = False,
                 renderer: Literal['canvas', 'svg'] = 'canvas',
                 ) -> None:
```

nicegui/elements/echart.py
```diff
@@ -32,10 +32,12 @@ def __init__(self, options: Dict, on_point_click: Optional[Handler[EChartPointCl
         :param options: dictionary of EChart options
         :param on_click_point: callback that is invoked when a point is clicked
         :param enable_3d: enforce importing the echarts-gl library
+        :param enable_svg: enable svg rendering for this chart
```
falkoschindler, line 35:
```suggestion
        :param renderer: renderer to use ("canvas" or "svg")
```

nicegui/elements/echart.py
```diff
@@ -32,10 +32,12 @@ def __init__(self, options: Dict, on_point_click: Optional[Handler[EChartPointCl
...
         """
         super().__init__()
         self._props['options'] = options
         self._props['enable_3d'] = enable_3d or any('3D' in key for key in options)
+        self._props['enable_svg'] = enable_svg
```
falkoschindler, line 40:
```suggestion
        self._props['renderer'] = renderer
```

falkoschindler (changes requested):
Thanks a lot for this pull request, @Piguite! Being able to change the renderer is a great addition to the `ui.echart` element.

The code looks good to me, but I think we should change the parameter from `enable_svg: bool` to `renderer: Literal['canvas', 'svg']`. This is more in line with the original ECharts library, indicates both options more clearly, and is extendable should there be more renderers in the future.

I would have made the required changes myself, but I don't have the necessary rights to push to your branch. ("Maintainers are allowed to edit this pull request" is not enabled.) Therefore I proposed the code changes in comments. You should be able to accept them directly from the web interface. Thanks!

falkoschindler (changes requested):
Thanks for the update, @Piguite!

Unfortunately, echart.py is missing an import. Can you, please, add "Literal" between "Dict" and "Optional" in the very first line? Thanks!

falkoschindler (approved):
Thanks again, @Piguite!
Let's merge 🚀

PR #3947, opened 2024-11-05
Add demo to show how to setup submenus

xaptronic:
I was trying to figure out if submenus would work and was almost about to open a discussion, but then took a closer look at the example on [Quasar](https://quasar.dev/vue-components/menu#submenus) and realized it was just a matter of adding the right props, and setting `auto_close=False` so that the submenus can be opened without dismissing the menu.

Adding a basic example of how to set this up.


falkoschindler (approved):
I just made some minor changes to follow our single-quote convention and added two more remarks to the documentation.
Let's merge!

PR #3968, opened 2024-11-09
Multicasting of page-scoped content

aranvir:
PR for https://github.com/zauberzeug/nicegui/discussions/3828

As discussed there added an iterator to `app` for yielding clients based on path. Also added documentation (at Page and ElementFilter since it kinda applies to both of them). Added a test.

I was also experimenting a bit with adding a wrapper or context manager as in my initial example, but I think @falkoschindler is right: an iterator is cleaner and then the developer can just do what they want with it.

E.g., in my original example
```python
    def update_clock():
        def _update():
            for element in ElementFilter(kind=ui.label, marker='clock'):
                element.text = time.strftime('%H:%M:%S')
        with_clients(OTHER_PAGE_ROUTE, _update)
```
versus now
```python
    def update_clock():
        for client in app.clients(OTHER_PAGE_ROUTE):
            with client.content:
                for element in ElementFilter(kind=ui.label, marker='clock'):
                    element.text = time.strftime('%H:%M:%S')
```

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -5,7 +5,7 @@ import signal
 import urllib
 from enum import Enum
 from pathlib import Path
-from typing import Any, Awaitable, Callable, List, Optional, Union
+from typing import Any, Awaitable, Callable, List, Optional, Union, Iterator
 
 from fastapi import FastAPI, HTTPException, Request, Response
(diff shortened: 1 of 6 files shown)


aranvir:
Thanks for the adjustments! Happy to contribute!

Issue #2744, opened 2024-03-21
Browser can't load page when table rows contain lists


falkoschindler:
### Description

As @rohitsathish noticed on #2697, the following minimal example causes the app to freeze:
```py
ui.table(columns=[{'name': 'A', 'label': 'A', 'field': 'A'}], rows=[{'A': [1, 2, 3]}])
```

It's even more apparent when creating the table dynamically:
```py
ui.button('Table', on_click=lambda: ui.table(columns=[{'name': 'A', 'label': 'A', 'field': 'A'}], rows=[{'A': [1, 2, 3]}]))
```

Somehow Vue and/or Quasar can't handle the row content. But a minimal Quasar app works with the same data:
```html
<html>
  <head>
    <link href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900|Material+Icons" rel="stylesheet" type="text/css" />
    <link href="https://cdn.jsdelivr.net/npm/quasar@2.15.1/dist/quasar.prod.css" rel="stylesheet" type="text/css" />
  </head>
  <body>
    <div id="q-app">
      <q-table :columns="[{'name': 'A','label': 'A','field': 'A'}]" :rows="[{'A': [1, 2, 3]}]" row-key="id" />
    </div>
    <script src="https://cdn.jsdelivr.net/npm/vue@3/dist/vue.global.prod.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/quasar@2.15.1/dist/quasar.umd.prod.js"></script>
    <script>
      const app = Vue.createApp({ setup() { return {}; } });
      app.use(Quasar);
      app.mount("#q-app");
    </script>
  </body>
</html>
```

I'm out of ideas. Can anyone help to find out what's going on? Thanks!

petergaultney:
I'm running into this today. Thank you for creating this - I was beginning to wonder if I was sane. 😄 

I don't have a great idea of how I might go about debugging this, but I'm interested in talking it through a bit. I can see that there's high CPU usage for a period of time after the Python code has apparently stopped doing anything (I assume it has at least sent the data to the `nicegui` JS side).

Is there a reasonably easy way to enable some kind of debugging logic on the JS side, so that I can try to see what messages are being received by the client? 

[...]

Issue #2706, opened 2024-03-15
`ui.table.from_pandas` fails for datetimes with timezones


falkoschindler:
### Description

As reported by @rohitsathish on #1983, `ui.table.from_pandas` fails for datetimes with timezones:

```py
df = pd.DataFrame({'Name': ['John'], 'DOB': ['2021-06-25 03:00:00+00:00']})
df['DOB'] = pd.to_datetime(df['DOB'])
ui.table.from_pandas(df)
```

PR #1983, opened 2023-11-09
UI.table from_pandas method

zilch42:
Adds `from_pandas` method to `ui.table` based on the similar method in `ui.aggrid`. 

Addresses #1972, including conversion of the non-serializable datatypes identified in #1698 

I think I have updated all of the documentation and tests but let me know if I have missed anything.

Issue #1698, opened 2023-09-27
ui.aggrid.from_pandas not working with DatetimeIndex


johancj:
### Description

I have a `pandas.DataFrame` which has a column with type `DatetimeIndex`. 

When used in `ui.aggrid.from_pandas(df)` it throws an error: `TypeError: Type is not JSON serializable: Timestamp`

Small example:
``` python
import pandas as pd
from nicegui import ui

df = pd.DataFrame({'numbers': [i for i in range(3)],
                   'dates': pd.date_range('2023-01-01', periods=3)})

ui.aggrid.from_pandas(df)

ui.run()
```

I have found a work around: casting `DatetimeIndex` to `str`.
```python
date_cols = df.columns[df.dtypes=='datetime64[ns]']
df[date_cols] = df[date_cols].astype(str) # Convert datetime to str
```

I think this should be handled by nicegui, either by handeling the casting or in some other way.

johancj:
It probably is the top of the iceberg. I tested some different stpes and got the same issue with the following:
```python
datetime64[ns]
timedelta64[ns]
complex128
```

I tested the following types:
```python
int64
float64
object
bool
datetime64[ns]
category
object
timedelta64[ns]
complex128
int64
uint8
int32
float32
float64
bool
uint16
int16
int8
```

falkoschindler:
Ok, here is an example with all problematic datatypes I could find and their conversion:

```py
from datetime import datetime, timedelta
import pandas as pd
from nicegui import ui

df = pd.DataFrame({
    'datetime_col': [datetime.now(), datetime(2020, 1, 1), datetime(2022, 12, 31)],
    'timedelta_col': [timedelta(days=5), timedelta(days=10), timedelta(days=15)],
    'complex_col': [1 + 2j, 2 + 3j, 3 + 4j],
    'period_col': pd.Series([pd.Period('2021-01'), pd.Period('2021-02'), pd.Period('2021-03')]),
})

date_cols = df.columns[df.dtypes == 'datetime64[ns]']
time_cols = df.columns[df.dtypes == 'timedelta64[ns]']
complex_cols = df.columns[df.dtypes == 'complex128']
period_cols = df.columns[df.dtypes == 'period[M]']
df[date_cols] = df[date_cols].astype(str)
df[time_cols] = df[time_cols].astype(str)
df[complex_cols] = df[complex_cols].astype(str)
df[period_cols] = df[period_cols].astype(str)

[...]

PR #3969, opened 2024-11-10
Integrating Polars into ui.table and ui.aggrid

trivedihoney:
This PR allows users to use from_polars and update_from_polars functions in ui.table.
Users will be able to use from_polars in ui.aggrid as well.

Changes were made in pyproject.toml Python version since Poetry was not allowing Polars installation with Python version ^3.8.
I have not worked extensively with various datatypes in polars. So, test cases can be improved in that regard. 
This is a WIP. I have integrated polars in ui.table so far.

I'm new to contributing to open-source projects. Any feedback is greatly appreciated.


trivedihoney:
from_polars method in ui.aggrid has been implemented

falkoschindler (commented):
Thanks a lot for this pull request, @trivedihoney! Support for Polars dataframes is a great addition to the library!

I just reviewed your code and improved it a little:
- I changed the minimum Python version back to 3.8. Even though its end-of-life date has passed, we might support it a little longer to avoid breaking changes in applications that haven't switched to newer Python versions yet. Depending on the Python version we can simply choose an older Polars version that still supports Python 3.8.
- To avoid too much duplicate code, I parameterized the pytests. This way we can re-use quite some code for both dataframe types.
- I re-ordered the new methods so that Pandas always comes before Polars, which feels more intuitive.

The one thing I'm still unsure about is handling the "problematic datatypes". See #1698, #1972, #1983, and #2706 for a little background on this subject. Some datatypes like `datetime`, `timedelta` and complex numbers caused exceptions when extracting rows and columns, so we convert them beforehand. But most of this doesn't seem to apply to Polars.
- `datetime` seems to work without conversion.
- `timedelta` causes an error: "polars.exceptions.InvalidOperationError: casting from Duration(Microseconds) to String not supported"
- complex numbers cause an error: "polars.exceptions.ComputeError: cannot cast 'Object' type"
- `pd.Period` doesn't exist in Polars, so I guess we can ignore it.

Maybe we find ways to avoid these errors. Otherwise we can still improve it later in another pull request.

Apart from that, I think casting every row to string and replacing null values is not necessary. Commenting it out doesn't seem to make a difference. Do you agree to remove these lines?

trivedihoney:
Hi @falkoschindler,
Thanks for the detailed reply!

> `timedelta` causes an error: "polars.exceptions.InvalidOperationError: casting from Duration(Microseconds) to String not supported"

`timedelta` is internally represented as `pl.Duration` in polars. In the current version of polars, `to_string` operation is not supported for dtype `duration[μs]`. There is a [pull request](https://github.com/pola-rs/polars/pull/19697) open that is supposed to handle this issue. 
We can use a workaround like below to get duration representation in seconds for current version, but I think it will be best to wait for a newer version of polars for this particular issue.
```python
df = df.with_columns(pl.col(pl.Duration).to_physical()) # timedelta 1 day --> 86400000000
```

> complex numbers cause an error: "polars.exceptions.ComputeError: cannot cast 'Object' type"

When we store Complex numbers in polars dataframe column, the column datatype becomes `Object` (Used for arbitrary python objects). It is possible to convert an object column to string using `map_elements` as shown below, but it is slow and [discouraged in documentation](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.map_elements.html). I am not sure if there are efficient and vectorized ways to achieve this.
```python
df = df.with_columns(pl.col(pl.Object).map_elements(lambda x: str(x), return_dtype=pl.String))
```

> Apart from that, I think casting every row to string and replacing null values is not necessary. Commenting it out doesn't seem to make a difference. Do you agree to remove these lines?

I added cast every row as string code because I thought it will handle complicated data types like `list[i64]`, `pl.Duration`, Nested Columns, etc. But now, upon testing I realized that it doesn't make a difference. So, we can remove those lines for now.

Just want to add one more thing that I noticed while testing, 
If I add list as an element in column as shown below, then niceGUI is not throwing any errors, and the page keeps on loading. This behavior is present in `from_pandas` and `from_polars` both.
Below is the minimal code to reproduce the issue.
```python
from nicegui import ui
import polars as pl

@ui.page("/")
def index_page():
    df = pl.DataFrame( 
        {"nested" : [[1, 2, 3], [4, 5, 6], [7, 8, 9]]}
    )
    ui.table.from_polars(df)

ui.run()
```

falkoschindler (approved):
Thanks for your feedback, @trivedihoney!

1. Let's leave the problems with `pl.Duration` and complex numbers unresolved. There is no obvious solution, so we better leave it to the user to decide how to preprocess the data.
2. I've removed the casting and null replacement.
3. The problem with table rows containing lists is a known issue (see #2744) and occurs independently of Polars:
    ```py
    ui.table(columns=[{"name": "x", "label": "x", "field": "x"}], rows=[{"x": [1, 2, 3]}])
    ```
    Unfortunately I still don't have a clue. But that's out of scope of this PR.

Let's finish this pull request by merging into main 🚀

alexander-beedie:
@falkoschindler, @trivedihoney:

FYI, the Polars "to_string()" enhancement was merged/released, and is available in the latest Polars version (`1.14.0`). A single `cs.temporal().to_string()` expression can now be used to convert all temporal types to ISO strings (see the linked PR and the "to_string" docstring for examples). Seems like this can be of use to you? Regards.

PR #3987, opened 2024-11-13
Introduce bindable dataclass fields

balex89:
This PR follows up an idea #3957 to introduce bindable `dataclasses.dataclass` fields.

The main challenge is to marry `bindableProperty` with `dataclasses.field`, and by that preserve native dataclass features.
The proposed idea is to use a wrapper around dataclasses.field that updates passed `metadata`, adding nicegui-specific options (for now just a "bindable" flag). Then use it to add `bindableProperties` retroactively on dataclass type postprocessing (much like it is done in [ dataclasses_json.config](https://github.com/lidatong/dataclasses-json?tab=readme-ov-file#encode-or-decode-from-camelcase-or-kebab-case))

# Usage example
```py
@bindable_dataclass
@dataclass
class MyClass:
    x: float = dataclass_bindable_field(default=1.0)
```

# Known tradeoffs
- Access to default field value through class attribute (e.g. `MyClass.x`) is lost.


falkoschindler (commented):
Thank you for the pull request, @balex89!

@rodja and I thought about the API quite a bit. Requiring the `@bindable_dataclass` decorator _and_ the field wrapper feels a bit verbose. In #3957 you wrote "One might want to be able to pick specific attributes only". But let's question this requirement. Do we really need to handle this use case? I know, the `bindable_field` function was kind of my idea. But if 99% of our users want to mark all fields as bindable anyway, your initial approach of converting _all_ fields would be much simpler and nicer to use. What do you think?

Apart from that we think we should merge `@dataclass` and `@bindable_dataclass` so that the user doesn't have to write both decorators and remember the right order. The new decorator would simply forward all keyword arguments to the `dataclass` decorator.

balex89:
> But if 99% of our users want to mark all fields as bindable anyway, your initial approach of converting all fields would be much simpler and nicer to use

@falkoschindler, I do agree on making that a default behavior. And yet, shouldn't we leave an option to specify fields explicitly for those 1%? For instance: 
```py
@bindable_dataclass
class A:
    x: int  # bindable
    y: int  # bindable
    z: int  # bindable

@bindable_dataclass(bindable_fields=['x', 'y'])
class B:
    x: int  # bindable
    y: int  # bindable
    z: int  # not bindable
```
That doesn't seem to complicate implementation or interface much while providing full control on attribute picking, if ever needed.

As was said above, we forward additional `kwargs` to `dataclass()`.

falkoschindler:
@balex89 You're right. Even though we could add such an additional attribute later on demand, we can as well add it right now. 👍

balex89:
@falkoschindler, I pushed some improvements. Futher down your list:

1. Type hints optimized.
2. Using `slots=True` now forbidden and raises a `ValueError`. Also i tested every other option and found `frozen=True` also complains about attribute setting. So now both this options are handled.
3. As far as i understand, your test does not perform as you might expect because of recursive value update propagation. When you, say, use a `slider`, this event calls `_propagate` on  `slider` value, which calls it on dataclass field value, witch eventually immediately updates `number` value. 
I added this code lines to your test to make difference observable:
    ```py
    ui.timer(0.2, lambda: list(map(
        lambda attr, value: setattr(a, attr, value),
        ['x', 'y'], [(a.x + 1) % 11] * 2
    )))
    
    ui.run(binding_refresh_interval=2.0)
    ```
    Here i automatically update both fields rapidly. Setting a huge `binding_refresh_interval` will show the difference.

balex89:
Hi @falkoschindler!
Issues with mypy seem to be fixed now. 

Turns out I initially misused `dataclass_transform` arbitrary param option. It has nothing to do with dataclass decorator params extension.

falkoschindler (approved):
Awesome! Thanks again, @balex89!
All linters and type checkers seem happy now. Let's merge!

1081:
I am very interested in this feature. Since I need a simply way to create models with may parameters I can bind to.

Not sure if my solution is robust. I changed the __init__ method of the binding.BindableProperty() descriptor to accept an optinal name. With this i can create a bindable model like this `model = create_model(["a", "b", "c"])`.

```python
from typing import Any, Callable, Optional

def __init__mod(self, on_change: Optional[Callable[..., Any]] = None, name="") -> None:
    self._change_handler = on_change
    self.name = name
    print("--> __init__", self.name)


binding.BindableProperty.__init__ = __init__mod

def create_model(attr_names):
    class Model:
        pass

    for name in attr_names:
        setattr(Model, name, binding.BindableProperty(name=name))

    model = Model()

    for name in attr_names:
        setattr(model, name, None)

    return model


model = create_model(["a", "b", "c"])

print(model.a, model.b, model.c)

ui.run(reload=False)
```

MateoSaez:
> 1. While version 1 preserves the correct type for `a` and `a.x`, with version 2 both type hints are `Any`. This is suboptimal.

I think the hints still show `Any`

falkoschindler:
@MateoSaez It works for me. Maybe your IDE or Pylance extension is not up to date?

Screenshot 2025-04-19 at 15 34 29

PR #3993, opened 2024-11-15
Add Python 3.13 to test workflow

falkoschindler:
This PR adds Python 3.13 to the test workflow on GitHub. Only the pandas dependency needed a special condition because only pandas 2.2.3 supports Python 3.13.


codingpaula (approved):
👍

PR #3997, opened 2024-11-15
Update the test screen implementation to allow customizable ports

engineerjames:
This merge request adds the capability to customize the ports that the test `Screen` implementation uses.  This allows for running multiple selenium-backed tests in parallel, and also allows users who aren't able to run in a container (and may be using this port for other things).

Admittedly I was having some issues running the unit tests that I am still working on, but I wanted to get the code change out there to see what the community thought.

### DIFF
--- a/nicegui/testing/screen.py
+++ b/nicegui/testing/screen.py
@@ -22,11 +22,11 @@ from nicegui.server import Server
 
 
 class Screen:
-    DEFAULT_PORT = 3392
+    PORT = 3392
     IMPLICIT_WAIT = 4
     SCREENSHOT_DIR = Path('screenshots')
 
-    def __init__(self, selenium: webdriver.Chrome, caplog: pytest.LogCaptureFixture, port: int = DEFAULT_PORT) -> None:
+    def __init__(self, selenium: webdriver.Chrome, caplog: pytest.LogCaptureFixture, port: int = PORT) -> None:
         self.selenium = selenium
         self.caplog = caplog
         self.port = port
@@ -36,9 +36,8 @@ class Screen:
         app.on_connect(self.connected.set)
         self.url = f'http://localhost:{self.port}'
 
-    def start_server(self, **kwargs) -> None:
+    def start_server(self) -> None:
         """Start the webserver in a separate thread. This is the equivalent of `ui.run()` in a normal script."""
-        self.ui_run_kwargs.update(kwargs)
         self.server_thread = threading.Thread(target=ui.run, kwargs=self.ui_run_kwargs)
         self.server_thread.start()
 
--- a/nicegui/testing/screen.py
+++ b/nicegui/testing/screen.py
@@ -22,21 +22,23 @@ from nicegui.server import Server
 
 
 class Screen:
-    PORT = 3392
+    DEFAULT_PORT = 3392
     IMPLICIT_WAIT = 4
     SCREENSHOT_DIR = Path('screenshots')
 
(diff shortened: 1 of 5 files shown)


rodja (approved):

falkoschindler (changes requested):
Thanks for the pull request, @engineerjames!

After having another look at your proposed changes, we have some concerns:

- Renaming `PORT` to `DEFAULT_PORT` can be considered a breaking change. Therefore we'd prefer to keep the name `PORT` unchanged.
- You added `self.ui_run_kwargs.update(kwargs)` to the `start_server` method. Why is this necessary?
- How to you actually use the new `port` parameter? The `screen` fixture doesn't allow to change it. And to change the port in Docker environments you could simply change the static `PORT` without using the new parameter.

engineerjames:
Of course! These are great questions.

> Renaming PORT to DEFAULT_PORT can be considered a breaking change. Therefore we'd prefer to keep the name PORT unchanged.

That is a fair point--happy to revert that.

> You added self.ui_run_kwargs.update(kwargs) to the start_server method. Why is this necessary?
It isn't necessary for allowing the customization of the port (which is my main concern), but it does allow for users to customize other arguments that end up being forwarded to `ui.run`.  

I thought it would be a useful addition, but I'm also happy to revert that part of the MR if you all feel it isn't a good idea.

> How to you actually use the new port parameter? The screen fixture doesn't allow to change it. And to change the port in Docker environments you could simply change the static PORT without using the new parameter.

Unfortunately our current setup doesn't allow us to use Docker--for the use of the port parameter I was envisioning something like:

```python

@pytest.fixture(scope='function')
def screen_fixture(caplog: pytest.LogCaptureFixture):
    yield Screen(selenium.webdriver.Chrome(...), caplog, port=your_port_here)

def test_loading_ui(screen_fixture: Screen) -> None:
   # Your test details here
```

I'm still fairly new to Python though, so I may have overlooked some easier options to accomplish this.

falkoschindler:
Another thought regarding your custom `screen_fixture`: This should be possible without the new `port` parameter.
```py
@pytest.fixture(scope='function')
def screen_fixture(caplog: pytest.LogCaptureFixture):
	Screen.PORT = 1234
    screen = Screen(selenium.webdriver.Chrome(...), caplog)
    yield screen
```

engineerjames:
> Another thought regarding your custom `screen_fixture`: This should be possible without the new `port` parameter.
> 
> ```python
> @pytest.fixture(scope='function')
> def screen_fixture(caplog: pytest.LogCaptureFixture):
> 	Screen.PORT = 1234
>     screen = Screen(selenium.webdriver.Chrome(...), caplog)
>     yield screen
> ```

Unfortunately this won't work because the default port is used in the constructor to set other member variables that are used in the implementation. :(

engineerjames:
> I don't understand. If `Screen.PORT` is set before calling `Screen()`, the effect on `self.ui_run_kwargs` and `self.url` equivalent to using a parameter `port`. The only difference would be if your code relies on the new member `self.port` or if some test needs to access `Screen.PORT`.
> 
> To be clear, I'm just trying to understand the benefit of the new `port` parameter. If you need a custom screen fixture anyway, we might just leave it as it is - or find a better way to make it configurable.

My apologies it took me so long to get back to you all; I was on a business trip.  Oh! I hadn't realized that setting `Screen.PORT` prior to construction would effectively do the same thing as I was trying to do.  Honestly in that case I don't think there's a need for the change I submitted.  Thank you all for the feedback and the consideration--I'm a huge fan of nicegui, and will keep an eye out for other ways to contribute in the future.

Issue #3886, opened 2024-10-16
Testing: Recognition of Autocomplete Input


GrazingScientist:
### Description

Hi!

I found another non-working feature in the testing framework, when I want to test that the autocomplete is working.

The test below is failing, although it should not.

```python
from nicegui import ui

def setup() -> None:
    @ui.page("/")
    def main() -> None:
        options = ['AutoComplete', 'NiceGUI', 'Awesome']
        ui.input(label='Text', placeholder='start typing', autocomplete=options)

async def test_validation(user):
    setup()

    await user.open("/")

    await user.should_not_see("AutoComplete")
    user.find("Text").type("Auto")
    await user.should_see("AutoComplete")
```

I hope, you can fix it. ❤️

rodja:
Thanks for bringing up this example @GrazingScientist. I think it is a little bit more complicated. If you start your page with NiceGUI and type "Auto" in the browser, you must also press the "TAB" key to confirm the auto completion. So we would need to use `user.find("Text").type("Auto").trigger("keydown.tab")`.

GrazingScientist:
Hi @rodja ,

thanks for coming back so quickly. I eagerly tried your tip, but the result is the same. The test fails with:

`AssertionError: expected to see at least one element with marker=AutoComplete or content=AutoComplete on the page`

This is the code I used:

```python
from nicegui import ui

def setup() -> None:
    @ui.page("/")
    def main() -> None:
        options = ["AutoComplete", "NiceGUI", "Awesome"]
        ui.input(label="Text", placeholder="start typing", autocomplete=options)

async def test_validation(user):
    setup()

    await user.open("/")

    await user.should_not_see("AutoComplete")
    user.find("Text").type("Auto").trigger("keydown.tab")
    await user.should_see("AutoComplete")
´´´

rodja:
I'm sorry. This is not implemented yet. I just wanted to clarify the API. We are pretty busy at the moment. Would you like to try creating a pull request.

GrazingScientist:
I will try to dig into the code, if I have the time. Can you provide me a link to the code section, where I need to focus on?

rodja:
I think you could try adding a special case for `ui.input` in the trigger loop over all elements:

https://github.com/zauberzeug/nicegui/blob/809290e3037a837fd231330c096da6409143f6ab/nicegui/testing/user_interaction.py#L38

Similar to what we do for `ui.link` and `ui.select` in the `click` method:

[...]

PR #4001, opened 2024-11-17
Add ui.input trigger autocomplete trigger capability

GrazingScientist:
This fixes the issue mentioned in #3886 .

The PR implements a workaround in the testing/user_interaction module by checking if a trigger event is a "tab" event and then replacing the current text value by the first element in the autocomplete list.

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Generic, Set, Type, TypeVar, Union
+from typing import TYPE_CHECKING, Generic, List, Set, Type, TypeVar, Union
 
 from typing_extensions import Self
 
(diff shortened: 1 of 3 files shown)


rodja:
Thanks for implementing this feature. Looks good so far. Maybe you could also add an example to the documentation?

nicegui/testing/user_interaction.py
```diff
@@ -36,6 +36,17 @@ def trigger(self, event: str) -> Self:
...
         with self.user.client:
             for element in self.elements:
+                if isinstance(element, ui.input) and event == 'keydown.tab':
+                    autocomplete_list = element.props.get('_autocomplete')
+                    typed_string = element.value
+
+                    for elem in autocomplete_list:
+                        if typed_string in elem:
+                            element.value = elem
+                            break
```
falkoschindler, line 46: Let's simplify this code a bit and use `startswith` instead of `in` to reproduce the real behavior correctly:

```suggestion
                    for option in element.props.get('_autocomplete'):
                        if option.startswith(element.value):
                            element.value = option
                            break
```

falkoschindler (approved):
Thanks, @GrazingScientist!

I just reviewed your pull request and made a few minor changes. Most importantly I removed the early exit `return self` which would skip other event handlers listening for "keydown.tab".

Ready to merge! 🚀

PR #4018, opened 2024-11-21
Allow ungrouping 3D objects

falkoschindler:
This PR implements feature request #4015 introducing an `ungroup` method for 3D objects:

```py
import math
import time
from nicegui import ui

with ui.scene() as scene:
    with scene.group() as group:
        a = scene.sphere().move(-2)
        b = scene.sphere().move(0)
        c = scene.sphere().move(2)

ui.timer(0.1, lambda: group.move(y=math.sin(time.time())))
ui.button('Ungroup', on_click=a.ungroup)

ui.run()
```

Open tasks:

- [x] better name: "detach"
- [x] introduce "attach" for symmetry
- [ ] resolve pose so that the object doesn't move visually


falkoschindler:
@rodja Resolving the object's world pose turns out rather tricky.

- We could compute it on the server. But currently we don't have NumPy as a dependency. And concatenating and inverting 3D motion matrices by hand is pretty challenging.
- We could compute it on the client using Three.js. This is relatively easy, but requires the server to await the result to update its internal state. This makes `attach` and `detach` async methods that can't be called on the auto-index page.

So I wonder if this requirement is worth the effort.

Update: I couldn't resist and implemented the computation for preserving position and rotation. But when extending the concept to preserve scale as well, I noticed that this isn't possible with our current object model. When scaling a box in x direction, attaching it to a rotating group and detaching it again, the resulting transformation contains shear, which can't be represented.

Either we preserve position, rotation but no scale, or we drop this requirement alrogether.

rodja (approved):
That's great @falkoschindler. Keeping position and orientation is a handy feature. I suggest we simply place note in the documentation that scaling is not preserved.

PR #4041, opened 2024-11-26
Documentation: fix inverted validation rule

hephaisto:
Minor documentation error. 
The original `{'Too small!': lambda value: value < 3}` would return
`True` if the number is small (error),
`False` if the number is big (okay).

But `ValidationElement` expects `False` to indicate an error:
```
            for message, check in self._validation.items():
                if not check(self.value):
                    self.error = message
                    return False
```

Alternate fix would be to replace `small` -> `big`.


falkoschindler (approved):
You're absolutely right, @hephaisto! Thanks for fixing this mistake!

Issue #2149, opened 2023-12-10
Re-Opening dialog resets inputs


adosikas:
### Description

I am trying to "hide" some settings in a dialog, but whenever I reopen the dialog, the initial value from page load is reset, seeming ignoring binds.

The label changes as expected when I modify the input, but the default value (either from the element or from the storage) is restored every time the dialog opens.

```py
from nicegui import ui, app

@ui.page("/")
def index():
    with ui.dialog() as edit_dialog, ui.card():
        text_input = ui.input("Text", value="Default").bind_value(app.storage.user, "text")
    ui.button("Edit", on_click=edit_dialog.open)
    ui.label().bind_text_from(text_input, "value")

ui.run(storage_secret="123")
```

I can work around it by manually updating the input when opening the dialog, but that hardly seems intentional.

falkoschindler:
Oh wow, thanks for reporting this issue, @adosikas!

Here is a reproduction without binding and without app storage:
```py
model = {'text': 'test'}
with ui.dialog() as dialog, ui.card():
    input_ = ui.input(value=model['text'], on_change=lambda e: model.update(text=e.value))
ui.button('Edit', on_click=dialog.open)
ui.label().bind_text_from(input_, 'value')
```

It looks like the problem is that the HTML input doesn't exist while the dialog is closed. Vue re-creates it when the dialog is opened, but somehow an outdated value is set. I wonder if we can reproduce the behavior with a plain Vue app...

falkoschindler:
A plain Vue app works as expected:
```html
<html>
  <head>
    <script src="https://cdn.jsdelivr.net/npm/vue@2.6.14/dist/vue.min.js"></script>
  </head>
  <body>
    <div id="app">
      <button @click="showDialog = true">Edit</button>
      <div v-if="showDialog" style="border: 1px solid black; padding: 10px">
        <input id="inputText" v-model="text" />
        <button @click="showDialog = false">Close</button>
      </div>
      <p>Entered Text: {{ text }}</p>
    </div>
    <script>
      new Vue({ el: "#app", data: { text: "test", showDialog: false } });
    </script>
  </body>
</html>
```

falkoschindler:
A Quasar app works as well:
```html
<html>
  <head>
    <link href="https://cdn.jsdelivr.net/npm/quasar@2.14.1/dist/quasar.css" rel="stylesheet" type="text/css" />
  </head>
  <body>
    <div id="q-app">
[...]

Issue #287, opened 2023-01-23
Text input example drops keys


joshalbrecht:
If you type quickly in the Text Input example on the API Reference page, you can see that it drops some characters. This makes it almost completely unusable in practice.

claytonaalves:
Yes, I can confirm that. I've got the same behavior here.

Works ok if you type slowly thou.

falkoschindler:
You are right, typing in a text input _can_ drop keys.

It actually happens because every value change is sent to the server where the event is processed and the new value is sent back to the client. I implemented this roundtrip because Quasar's QToggle simply would change its that when clicking on it even though the event was sent to the server. Somehow I had to send the updated value back to the client for it to move.

While this technique might be needed for some elements for whatever reason, it might not be necessary for other. Apparently we don't need to send the new value for a QInput. So I'll introduce an option in NiceGUI's `ValueElement` base class so that `ui.toggle` can behave differently than `ui.input` in this regard. This will hopefully avoid dropping keys while typing.

falkoschindler:
I checked the behavior for all elements deriving from `ValueElement` with the following code:

```py
with ui.row():
    ui.checkbox('Checkbox')
    ui.switch('Switch')
with ui.row():
    ui.radio(['A', 'B', 'C'])
    ui.select(['A', 'B', 'C'])
    ui.toggle(['A', 'B', 'C'])
with ui.row():
    ui.input('Input')
    ui.number('Number')
    ui.color_input('Color input')
with ui.row():
    with ui.menu() as menu:
        ui.menu_item('Ok')
    ui.button('Menu', on_click=menu.open)
    with ui.dialog() as dialog:
        ui.button('Ok', on_click=dialog.close)
    ui.button('Dialog', on_click=dialog.open)
with ui.row().classes('w-96'):
    lp = ui.linear_progress()
    cp = ui.circular_progress()
    ui.slider(min=0, max=1, step=0.1).bind_value_to(lp, 'value').bind_value_to(cp, 'value')
with ui.row():
    ui.date()
    ui.time()
```

I tried it once with sending changed values back to the client and once without.
To detect problems caused by the roundtrip, I simulated a slow connection using the Chrome developer tools.

- All input elements involving typing (`ui.input`, `ui.number`, and `ui.color_input`) show flickering and lost characters. When disabling the roundtrip, the flickering is gone but they get cleared when loosing focus. So we probably need to set their state sometime, maybe on change but with some throttling.
[...]

PR #4045, opened 2024-11-28
fix ui.input value loss

NiklasNeugebauer:
I noticed that the ui.input does not keep its displayed value when a dialog is closed and opened again.
Even more confusingly it keeps the value that was last set by the nicegui server (either by setting input.value or just connecting the client while it had a value).

As far as I managed to understand it, the issue is with the value prop is not being set properly when the user changes the input (though input.value does change correctly). Setting LOOPBACK back to `True` seems to solve this.

I do not understand this setting enough to answer the following questions:
What does it change?
Why was it set to False in the first place?

### DIFF
--- a/nicegui/elements/input.py
+++ b/nicegui/elements/input.py
@@ -8,7 +8,6 @@ from .mixins.validation_element import ValidationDict, ValidationElement, Valida
 
 class Input(ValidationElement, DisableableElement, component='input.js'):
     VALUE_PROP: str = 'value'
-    LOOPBACK = False
 
     def __init__(self,
                  label: Optional[str] = None, *,


NiklasNeugebauer:
related to https://github.com/zauberzeug/nicegui/issues/2149

falkoschindler:
Thanks, @NiklasNeugebauer!

The static `LOOPBACK` attribute was introduced in #287 to define the value update behavior for certain elements: Usually the user triggers a change event, which is sent to the server to update the data model in Python, and finally sent back to the client to update the frontend. This can cause issues when typing quickly, because the server update arrives after the next keys have been pressed. Setting `LOOPBACK = False` for input elements disables this additional roundtrip and sets the new model value directly in the frontend. In case of dialogs this shortcut for input elements doesn't seem to work properly.

Even though I don't think we can merge this PR because it would re-introduce issue #287, it might provide a workaround: Set `input.LOOPBACK = True` for input elements on dialogs.

NiklasNeugebauer:
Thanks for the explanation! Seems like it works fine in our app.

PR #4052, opened 2024-11-30
update pre-commit and ruff

codingpaula:
This PR updates ruff (0.3.5 to 0.8.1) and pre-commit. It also includes a few sorting and linting fixes that resulted from the ruff update.


nicegui/json/__init__.py
```diff
@@ -15,7 +15,7 @@
 
 
 __all__ = [
+    'NiceGUIJSONResponse',
     'dumps',
-    'loads',
-    'NiceGUIJSONResponse'
+    'loads'
```
falkoschindler, line 20:
```suggestion
    'loads',
```

falkoschindler (approved):
Nice! I just noticed a missing comma (which was already missing before). Let's fix it once we're at it.

PR #4057, opened 2024-12-02
fix _gather_elements in User class to only return visible elements

rlcmtzc:
When using `User` simulation methods such as `should_see`, `should_not_see`, or `find`, the `visible` state of the element is ignored. This can result in unexpected behavior. For example, if an element is invisible, but `should_see` or `find` is used, the visibility state is disregarded.

Consider a scenario where an element is created and set to be invisible, but becomes visible at a later point. In the current version, testing with `should_see` would always pass because the 'visible' state is ignored. Ideally, `should_see` should raise an error if the element is not visible.

This behavior is similar for `find` and `should_not_see`, where the visibility state is not taken into account during the test.

This is fixed by only returning visible elements in `_gather_elements`. Additionally for better debugging the `__str__` method of `Element` includes the `visible` state of the element.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -519,6 +519,7 @@ class Element(Visibility):
             for key, value in self._props.items()
             if not key.startswith('_') and key not in IGNORED_PROPS and value
         ]
+        additions.append(f'visible={self.visible}')
         if additions:
             result += f' [{", ".join(additions)}]'
 
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -519,7 +519,8 @@ class Element(Visibility):
             for key, value in self._props.items()
             if not key.startswith('_') and key not in IGNORED_PROPS and value
         ]
-        additions.append(f'visible={self.visible}')
+        if not self.visible:
+            additions.append(f'visible={self.visible}')
         if additions:
             result += f' [{", ".join(additions)}]'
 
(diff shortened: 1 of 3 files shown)


rodja:
Thanks for this crucial fix, @rlcmtzc. I'm not sure we should add the visibility to the `__str__` output. Due to the fix, it will always be `True` anyway. And it hinders readability in my opinion.

rlcmtzc:
@falkoschindler Good point. This makes the `__str__` less verbose, but still helps for debugging.

rodja (approved):
I love it!

falkoschindler:
I fixed some Ruff issues on main: 6efd2479621824b778aacc7776e157d6f19d7463.

marcuslimdw:
I agree that this is a good thing to have (had to work around it in a few ways in my existing tests); what I used to have:

```
assert not page_user.find(kind=Spinner).elements.pop().visible
```

what I use now:

```
await page_user.should_not_see(kind=Spinner)
```

*however*, this change breaks my old code, and is therefore, strictly speaking, backwards incompatible. this is a bit of a small thing to do a major version bump over, but IMO, there should be at least a warning notice added to the documentation for `should_see`/`should_not_see`/`find`.

rodja:
Thanks for bringing this up @marcuslimdw. I updated the release notes a bit: https://github.com/zauberzeug/nicegui/releases/tag/v2.8.0

PR #4091, opened 2024-12-11
Introduce `app.timer`

falkoschindler:
This PR implements one of our most-wanted feature requests #3225, an API for timing UI-independent function calls.

The current draft can be tested like this:
```py
import time
from nicegui import app, ui

timer = app.timer(0.5, lambda: print(time.time()))
ui.button('Start', on_click=timer.start)
ui.button('Stop', on_click=timer.stop)
ui.switch('Running').bind_value(timer, 'running')

ui.run()
```

Open tasks:

- [ ] pytest
- [ ] documentation
- [ ] Should we start with sleeping or with calling the handler? `ui.timer` calls immediately, but RoSys' repeater doesn't: https://github.com/zauberzeug/rosys/blob/96da831c272f56daa664382e9ecb6dac159a4b85/rosys/rosys.py#L152
- [ ] Do we need the complicated exception handling?
- [ ] Do we need to call `stop_all` during shutdown?

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -18,6 +18,7 @@ from ..observables import ObservableSet
 from ..server import Server
 from ..staticfiles import CacheControlledStaticFiles
 from ..storage import Storage
+from ..timer import Timer
 from .app_config import AppConfig
 from .range_response import get_range_response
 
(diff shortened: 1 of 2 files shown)


rodja (approved):
Looks good. I just modified the `app.timer` test a little bit. It was too similar to the `ui.timer` test on an auto-index page.

PR #4098, opened 2024-12-14
Disallow typing to disabled UI elements in `UserInteraction.type`

### LINKED ISSUE #4080: Able to type to disabled Input element when using User Fixture in testing
### Description

First up, thank you for creating and maintaining this great project! 

## Description
I am using the User Fixture to write UI tests. I am trying to simulate typing to a **disabled** `Input` element using the User Fixture. 

## Version Information
* Python 3.11.9
* nicegui==2.8.1
* pytest==8.3.2
* pytest-asyncio==0.24.0

## Minimum Reproducible Example

```python
import pytest
from nicegui import ui

MARKER_INPUT = "marker-input"


def build_ui() -> None:
    @ui.page("/")
    def page() -> None:
        with ui.input(label="My Input") as input_element:
            input_element.disable()
            input_element.mark(MARKER_INPUT)


@pytest.fixture
async def my_user(user: User) -> User:
    build_ui()
    await user.open("/")
    return user


@pytest.mark.asyncio
async def test_type_to_disabled_input(my_user: User) -> None:
    my_user.find(marker=MARKER_INPUT).type("Hello World!")
    await my_user.should_not_see("Hello World!")

```

## Expected Result
I expect the **disabled** `ui.input` element's value not to be changed when using `UserInteraction.type`. 

## Actual Result
But `UserInteraction.type` changes the `ui.input` element's value although it is disabled.

randy-seng:
The `UserInteraction` method `type` allows typing to a disabled element. This should be disallowed. This commit fixes that by skipping typing to elements that have the 'disable' property set to `True`.

Resolves #4080

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -6,6 +6,7 @@ from typing_extensions import Self
 
 from nicegui import background_tasks, events, ui
 from nicegui.element import Element
+from nicegui.elements.mixins.disableable_element import DisableableElement
 from nicegui.elements.mixins.value_element import ValueElement
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


rodja (approved):
Thank you so much!

PR #4114, opened 2024-12-17
Make BindableProperty add to bindable_properties

nathanjshaffer:
Make BindableProperty add to bindable_properties during binding instead of waiting till next value change.  This fixes a potential issue where a large quantity of BindablePropertys were bound at once would cause a large resource usage if the \_\_set\_\_() method was not called immediately after binding

based on discussion https://github.com/zauberzeug/nicegui/discussions/4101

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -7,8 +7,6 @@ from typing import Any, Callable, DefaultDict, Dict, Iterable, List, Optional, S
 from . import core
 from .logging import log
 
-import pdb
-
 MAX_PROPAGATION_TIME = 0.01
 
 bindings: DefaultDict[Tuple[int, str], List] = defaultdict(list)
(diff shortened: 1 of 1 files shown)


nathanjshaffer:
based on discussion https://github.com/zauberzeug/nicegui/discussions/4101

falkoschindler:
Thanks a lot for this pull request, @nathanjshaffer!

Can you, please, share a reproduction for the specific problem of "large resource usage" when binding many bindable properties at once? I'm still not sure how this might happen.

Your example in https://github.com/zauberzeug/nicegui/discussions/4101#discussion-7682545 was missing an initializer `def __init__(self)` that creates the attribute, e.g., `self.bar = ''`. This is why the binding couldn't make use of the bindable property.

Issue #4061, opened 2024-12-03
Leaflet does not initialise properly and flickers


hal2000gh:
### Description

I have an issue with leaflet maps. I have an app where I want to select and display an GPS track. When the track is selected I move the view to the track with "flyToBounds" (or "fitBounds", which shows the same effect) and quite often the map flickers after moving to the specified area and only the track is visible but not the map.
The same effect appears on the nicegui example for "Wait for Initialization"... see the screen capture. The first reload works properly but the second does not:

https://github.com/user-attachments/assets/4d97bfe8-694f-4df2-b378-0f1b043158b3

There was a bugfix in [release 1.4.26](https://github.com/zauberzeug/nicegui/releases/tag/v1.4.26) ("Prevent `ui.leaflet` from flickering...") which I suppose should have fixed the issue, but it is still occurring (particularly prevalent on Safari on iPad).

falkoschindler:
No, 50ms doesn't help reliably. Here's my test setup, by the way:
```py
@ui.page('/')
async def page():
    m = ui.leaflet(zoom=5)
    central_park = m.generic_layer(name='polygon', args=[[
        (40.767809, -73.981249),
        (40.800273, -73.958291),
        (40.797011, -73.949683),
        (40.764704, -73.973741),
    ]])
    await m.initialized()
    bounds = await central_park.run_method('getBounds')
    m.run_map_method('fitBounds', [[bounds['_southWest'], bounds['_northEast']]])
```

falkoschindler:
I might have found a solution in PR #4121.

hal2000gh:
Great thank!

PR #4121, opened 2024-12-18
Prevent `ui.leaflet` from oscillating between two locations

falkoschindler:
This PR tries to solve #4061 by disabling update messages when new center and zoom values have been received from the client. This technique is also used for our `ValueElement`s.

A minor downside of this approach might be that maps on the auto-index page might not stay in sync between different browsers. But I think this is a reasonable tradeoff.

### DIFF
--- a/nicegui/elements/leaflet.js
+++ b/nicegui/elements/leaflet.js
@@ -65,7 +65,6 @@ export default {
       "zoomanim",
     ]) {
       this.map.on(type, async (e) => {
-        await this.$nextTick(); // NOTE: allow zoom and center to both be updated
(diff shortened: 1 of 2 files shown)


nicegui/elements/leaflet.js
```diff
@@ -65,7 +65,6 @@ export default {
       "zoomanim",
     ]) {
       this.map.on(type, async (e) => {
```
falkoschindler, line 67:
```suggestion
      this.map.on(type, (e) => {
```

rodja (approved):

Issue #4109, opened 2024-12-16
Lifetime of objects owning a `BindableProperty`


andybayer:
When I stumbled across `BindableProperty`, I thought it would also be very useful for modeling data that can be modified through various "widgets" and keyboard shortcuts. Basically, a model-like object that encapsulates the single source of truth for some data. A small [example on the official website](https://nicegui.io/documentation/section_binding_properties#bindable_properties_for_maximum_performance) seems to support that this is an intended use case.

So I built a small application using this idea. It seems to work quite well. Since I also needed to clean up some resources for which a `with`-statement would not be very practical (and would also require some minor refactoring), I simply used `weakref.finalize` to add a "finalizer" to the model that does the cleanup. This is similar to RAII in C++, but without the deterministic behavior of when exactly the finalizer will run.

Then I noticed that the resources were never cleaned up, even when the garbage collection was forced to run. So I looked for references to my model. Since I used `BindableProperty` and bindings in general a lot, there were several references to my model in the global objects in the `bindings` module. I then looked up how these are all cleaned up. It seems that they are only cleaned up when the client is deleted (except for some special behavior for `ui.scene`)? The client by default only calls `binding.remove` for elements added to it. But at least it seems to remove related references to the model's `BindableProperty` as well. 

What is left is a reference to the model in `binding.bindable_properties`, since it is the owner of the property.  This will keep the model alive if it is not explicitly deleted, and will also _leak memory_. You could add a deleter manually in `app.on_disconnect` or as a client finalizer, but I am not sure if this is really a good solution. Maybe it would be better to remove the reference to the owner in `binding.bindable_properties` as I cannot see that it is currently needed anywhere, and add a finalizer using `weakref.finalize` to clean up the `(id, name)` tuples? But maybe the reference to the owner has some use I have not seen.

Below is a small example that shows the behavior described above. The naive model is not deleted when you disconnect from the client. The fixed model is deleted when you disconnect and start the next session.

```python
import gc
import weakref
from collections.abc import Callable
from datetime import datetime

from nicegui import app, ui
from nicegui.binding import bindable_properties, BindableProperty


class Model:
    name = BindableProperty()
    instances: list[weakref.ReferenceType] = list()

    def __init__(self, name: str) -> None:
        self.created = datetime.now()
        self.name = name

        self.instances.append(weakref.ref(self))

        weakref.finalize(self, lambda: log(f'Model "{name}" FINALIZED'))
        log(f'Model "{name}" initialized')

[...]

PR #4122, opened 2024-12-18
Automatic unregistering of `BindableProperty` objects

andybayer:
First draft to fix the issue reported in #4109.

Replaces the values in the `binding.bindable_properties` data structure, which acts as a "registry" for available bindable properties, with `weakref.finalize` objects. Previously, there was a permanent reference to the owner of the `BindableProperty`, which was never cleared unless explicitly removed with `binding.remove`.

I also added some very basic tests for this behavior. You may need to refactor these slightly. 

*What I did not test, and what in theory should still be a problem, is when 2 models have bindable properties and one model binds to a value of the other. Then permanent references to the models are kept in `binding.bindings`, which are never automatically cleaned up.*

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -150,18 +150,11 @@ class BindableProperty:
         if has_attr and not value_changed:
             return
         setattr(owner, '___' + self.name, value)
-        self._register(owner)
+        key = (id(owner), str(self.name))
+        bindable_properties.setdefault(key, weakref.finalize(owner, lambda: bindable_properties.pop(key, None)))
         _propagate(owner, self.name)
         if value_changed and self._change_handler is not None:
             self._change_handler(owner, value)
-
-    def _register(self, owner: Any) -> None:
-        registry_key = (id(owner), str(self.name))
-
-        def try_unregister() -> None:
-            bindable_properties.pop(registry_key, None)
-
-        bindable_properties.setdefault(registry_key, weakref.finalize(owner, try_unregister))
 
 
 def remove(objects: Iterable[Any]) -> None:
@@ -183,10 +176,9 @@ def remove(objects: Iterable[Any]) -> None:
         ]
         if not binding_list:
             del bindings[key]
-    for registry_key, finalizer in list(bindable_properties.items()):
-        obj_id, _ = registry_key
+    for (obj_id, name), finalizer in list(bindable_properties.items()):
         if obj_id in object_ids:
-            del bindable_properties[registry_key]
+            del bindable_properties[(obj_id, name)]
             finalizer.detach()
 
 
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -1,5 +1,6 @@
 import asyncio
 import time
+import weakref
 from collections import defaultdict
 from collections.abc import Mapping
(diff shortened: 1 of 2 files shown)


andybayer:
Yes, I tried replacing the `Dict` with a `Set` and everything worked as expected. I started to implement this fix, but suddenly I thought that the `Set` also needs to be cleaned up, or else deleted items will clutter it up. So I thought an automatic cleanup action was needed. 

I added the cleanup action as a new value to the `Dict`. Actually, I think it's not the best design because the `Dict` now has two jobs. It serves as a quick lookup of whether an object-attribute-pair is a `BindableProperty` and also as a storage for abortable cleanup actions. Perhaps we should not store a reference to the cleanup action at all, since it does no harm to run it after the object has already been manually removed with `binding.remove`?

andybayer:
So what would you suggest as the next step(s) @falkoschindler?

Shall I change something?

falkoschindler (approved):
Hi @andybayer,

Sorry for taking sooo long to get back to you. Let me try to recap where we were:

Assigning a value to a `BindableProperty` for the first time stores the `owner` object in a `bindable_properties` dictionary:
https://github.com/zauberzeug/nicegui/blob/786e24fb65ba62909cef7d90cec633671666d262/nicegui/binding.py#L176

For NiceGUI's UI elements this is probably ok, because bindings get cleaned up when removing the UI object. But for custom objects the user must need to call `bindings.remove()`.

This is demonstrated with the following demo:

```py
import gc
import weakref

clients: list[weakref.ReferenceType] = []
models: list[weakref.ReferenceType] = []

class Model:
    answer = binding.BindableProperty()

    def __init__(self) -> None:
        self.answer = 42

@ui.page('/')
def main() -> None:
    client = ui.context.client
    clients.append(weakref.ref(client))  # to count the number of clients
    weakref.finalize(client, lambda: print('Client finalized'))  # to print when the client is finalized
    print(f'Number of clients: {len([c for c in clients if c() is not None])}')

    model = Model()
    models.append(weakref.ref(model))  # to count the number of models
    weakref.finalize(model, lambda: print('Model finalized'))  # to print when the model is finalized
    print(f'Number of models: {len([m for m in models if m() is not None])}')

    ui.button('GC', on_click=gc.collect)
    ui.button('Remove bindings', on_click=lambda: binding.remove([m() for m in models if m() is not None]))
    print(f'Number of bindings: {len(binding.bindable_properties)}')

ui.run()
```

1. Reloading the page increases both, the number of clients as well as the number of models.
2. Running the garbage collector ("GC") reduces the number of clients (need a few seconds to disconnect), but does not affect the number of models.
3. Only removing the bindings reduces the number of models as well.

I suggested to resolve this problem by replacing the `bindable_properties` dictionary with a set, because we don't really need the dictionary values. For clarity, I sketched this approach in PR #4332.

After pushing this PR, I suddenly understood your problem with this approach: Even though the models are not held in the `bindable_properties` dictionary anymore, the set still grows if not pruned via `bindings.remove()`. This can be seen in the "Number of bindings" output in the demo above. This number grows rapidly with the number of clients, and even when removing clients via "GC" the number doesn't fall back to where it was.

So _finally_ I fully understand the need for your weakrefs like in your approach. I just refactored the code a little and shorten it here and there. Tests are still green. Let's merge!

PR #4128, opened 2024-12-20
Introduce a "Map of NiceGUI"

falkoschindler:
This PR adds an overview of the NiceGUI namespace to the documentation page.

- [x] It's still a draft, because I'm planning to add links to the individual documentation pages where possible.
- [x] And although I like the blue color to highlight the code snippets, it can be confused with links. So maybe we find a better way to highlight them.


rodja:
You modified the default css styling. Probably to enhance readability. But now the two kinds of links are intermixed:
Screenshot 2025-01-11 at 04 49 18

I think we should not alter the link styling:
Screenshot 2025-01-11 at 04 52 06

PR #4132, opened 2024-12-21
Fix leaked semaphore object warning

### LINKED ISSUE #4131: Leaked Semaphore Objects
### Discussed in https://github.com/zauberzeug/nicegui/discussions/4129

<sup>Originally posted by **ChipsAHoai** December 20, 2024</sup>
### Question

I am having leaked semaphore issues when running nicegui with fastapi.  Initially, I thought it was my code, however, I just tested this on Zauberzeug's repo running the start.sh bash script under examples/fastapi and immediately closed down the uvicorn server and it still shows leaked semaphores.  Is there a solution for this? </div>

Reproduction:

```
cd examples/fastapi
./start.sh dev
Starting Uvicorn server in development mode...
INFO:     Will watch for changes in these directories: ['nicegui/examples/fastapi']
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [54107] using WatchFiles
INFO:     Started server process [54109]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
^CINFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [54109]
INFO:     Stopping reloader process [54107]
/opt/homebrew/Cellar/python@3.12/3.12.6/Frameworks/Python.framework/Versions/3.12/lib/python3.12/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 6 leaked semaphore objects to clean up at shutdown
```

`^CINFO:     Shutting down` is the line where I press CTRL-C to abort the execution.

It also happens when starting native mode and reloading:

```py
from nicegui import ui

ui.label("hello")

ui.run(native=True)
```

Pressing CTRL-C yields also a leaked semaphore objects warning.

rodja:
This pull request fixes #4131: a "leaked semaphore object" warning appears when aborting with CTRL-C in certain configurations. Semaphores are used internally in `multiprocessing.Queue` which never get cleaned up properly.

### Scenario 1: pressing CTRL-C in an FastAPI `app` stared via uvicron from the command line with auto-reloading

Because the `multiprocessing.Queue` objects in `native.py` have been created globally they where also instantiated when not in native mode.  This PR fixes this by only creating the objects when native mode is activated.

### Scenario 2: pressing CTRL-C when using `ui.run(native=True, reload=True)`

This is still unsolved. I'm not sure where to properly call the shutdown/cleanup function.

### DIFF
--- a/nicegui/native/__init__.py
+++ b/nicegui/native/__init__.py
@@ -1,4 +1,4 @@
-from .native import WindowProxy, method_queue, response_queue
+from .native import WindowProxy, method_queue, on_shutdown, on_startup, response_queue
 from .native_config import NativeConfig
 from .native_mode import activate, find_open_port
 
@@ -8,5 +8,7 @@ __all__ = [
     'activate',
     'find_open_port',
     'method_queue',
+    'on_shutdown',
+    'on_startup',
     'response_queue',
 ]
--- a/nicegui/native/native.py
+++ b/nicegui/native/native.py
@@ -2,13 +2,33 @@
 import inspect
 import warnings
 from multiprocessing import Queue
-from typing import Any, Callable, Tuple
+from typing import Any, Callable, Optional, Tuple
 
 from .. import run
 from ..logging import log
 
-method_queue: Queue = Queue()
-response_queue: Queue = Queue()
+method_queue: Optional[Queue] = None
+response_queue: Optional[Queue] = None
+
+
+def on_startup():
+    """Initialize the message queues."""
+    global method_queue, response_queue  # pylint: disable=global-statement # noqa: PLW0603
+    method_queue = Queue()
+    response_queue = Queue()
+
+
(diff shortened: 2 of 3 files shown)


rodja:
Pressing CTRL-C when using `ui.run(native=True, reload=True)` still shows the leaked semaphore object warning. I verified that the shutdown/cleanup function for the Queue objects are correct. But it seems that the pywebview window produces also leaked semaphores when pressing CTRL-C. I could not figure out a way to avoid it. Is anyone able to jump in?

rodja:
It seems the leaked semaphores warning which appears with Native Mode is caused by something internal when pywebviews `webview.start` is being shut down. I suggest to ignore this for the current pull request.

falkoschindler (approved):

PR #4144, opened 2024-12-25
add `read_image` api to `ui.clipboard`

liunux4odoo:
make it easy to read copied image from client clipboard.


falkoschindler (changes requested):
Thanks for the pull request, @liunux4odoo!

- [x] To be honest, I don't understand how to use this feature. How can I copy an image to the clipboard? Whatever I try, your demo says I need to copy an image first.

Apart from that, there are some problems with this implementation:

1. [x] We can't simply break existing API by renaming `read` and `write`. I suggest to leave them unchanged while adding `read_image`.
2. [ ] The demo doesn't work for two reasons:
    1. The call to `ui.clipboard.read_image()` needs to be awaited.
    2. Awaiting a JavaScript command isn't allowed on the auto-index client, so the demo needs to be wrapped in a page function. But in order for the demo to keep working when embedded on the NiceGUI website, we need to prevent the decorator from adding a new route. [Here](https://github.com/zauberzeug/nicegui/blob/920b9e372c01c0f1369968b3c11039fc266c1401/website/documentation/content/run_javascript_documentation.py#L6-L25) is an example how to do that.

liunux4odoo:
I have reversed the `clipboard.read` & `clipboard.write` APIs.
User must copy an image to the clipboad before he can read it.

falkoschindler:
Ok, I managed to copy an image like this:

1. Open an image in Preview (macOS)
2. Use Edit > Copy

falkoschindler (approved):
Thanks again, @liunux4odoo!
I just improved the way PIL is imported because it is an optional dependency.
Let's merge!

PR #4146, opened 2024-12-25
example async drag-n-drop

drewfish:
This example builds on the existing Trello Cards example and adds 2 features:
* Uses `app.storage.tab` for the currently-dragged card. This is more "production ready" than a global.
* Introduces some "processing time" between when the card is dropped and when it's "committed". This is to demonstrate handling drag-n-drop which takes time to take effect.

The approaches in this PR might not be the best 😞 
* I didn't use the nice OOP approach found in `examples/trello_cards/draganddrop.py`. I think there might be a way to do that. It was getting tricky, though, in my mind when I started thinking about also being able to drag columns to re-order them.
* In order to managed the skeletons, I had to monkey around in the internals of the slots 😞. I _think_ this might be refactorable if there's a new `ui.column.replace(old_element, new_element)` or something like that. (I might attempt that in a separate PR.)

https://github.com/user-attachments/assets/05786b7b-12be-4074-bcd5-8a5ab6b2b168

### DIFF
--- a/examples/trello_cards/async-main.py
+++ b/examples/trello_cards/async-main.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+import asyncio
+from dataclasses import dataclass
+from functools import partial
+from typing import Optional, Tuple
+
+from nicegui import app, ui
+
+
+@dataclass
+class ToDo:
+    title: str
+
+
+@ui.page('/')
+def main():
+    def make_droppable_column(name: str) -> ui.column:
+        col = ui.column().classes('bg-blue-grey-2 w-60 p-4 rounded shadow-2')
+        with col:
+            ui.label(name).classes('text-bold ml-1')
+        col.on('dragover.prevent', partial(on_dragover, col))
+        col.on('dragleave', partial(on_dragleave, col))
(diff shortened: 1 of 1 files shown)


falkoschindler (changes requested):
Thanks a lot for this pull request, @drewfish! Adding support for async drop handlers sounds like a reasonable extension.

But I have some concerns about your approach of adding a second app to the example:

- I would prefer improving the existing app rather than adding another one.
- As far as I understand, you didn't stick with the OOP approach because you "started thinking about also being able to drag columns". But dragging columns isn't part of this example, so I don't see a reason to change the whole architecture (at least in this PR).
- The `on_drop` function mixes basic drag&drop functionality (including skeletons) with user code (sleeping for one second in this example). This makes it hard to see how to apply the idea to a real-world scenario.
- I'm not sure about the use of skeletons while dropping cards: Shouldn't the drop happen very quickly? Even if there is an async database call, it could run in the background after updating the UI.

I'd suggest to clarify the goal of this PR first before improving the implementation. Do we want to show how to support async drop handlers? Then we can modify the existing example pretty easily. Do we really want to show skeletons? Then we'd need to think about how to incorporate it into the existing example.

PR #4192, opened 2025-01-08
Fix potential matplotlib warning when using ui.line_plot.push

drkspace:
## Description

When trying to use `ui.line_plot.push` with x or y values that are all equal, you get the following matplotlib warning:
```
lib/python3.12/site-packages/nicegui/elements/line_plot.py:72: UserWarning: Attempting to set identical low and high ylims makes transformation singular; automatically expanding.
  self.fig.gca().set_ylim(min_y - pad_y, max_y + pad_y)
```

This PR changes 2 things:

1. It will not update the x and/or y limits if their respective data points are all the same value.
2. Add `update_y_lims` and `update_x_lims` parameters to `push` that will disable the limits updating for that push. This will also allow for custom limits to be set (and not reset every call to push).

## Example

This example will have a warning on the current version of nicegui.

```python
from nicegui import ui

lp = ui.line_plot(n=1)

i = 0
def update_lp():
    global i
    lp.push([i], [[0]])
    # lp.push([i], [[0]], update_y_lims=False)

    i += 1
ui.timer(1, update_lp)
ui.run()
```

### DIFF
--- a/nicegui/elements/line_plot.py
+++ b/nicegui/elements/line_plot.py
@@ -42,19 +42,18 @@ class LinePlot(Pyplot):
         self._convert_to_html()
         return self
 
-    def push(self, x: List[float], Y: List[List[float]], x_limits: Union[None, Literal['auto'], Tuple[float, float]] = 'auto', y_limits: Union[None, Literal['auto'], Tuple[float, float]] = 'auto') -> None:
+    def push(self,
+             x: List[float],
+             Y: List[List[float]],
+             x_limits: Union[None, Literal['auto'], Tuple[float, float]] = 'auto',
+             y_limits: Union[None, Literal['auto'], Tuple[float, float]] = 'auto',
+             ) -> None:
         """Push new data to the plot.
 
         :param x: list of x values
         :param Y: list of lists of y values (one list per line)
-        :param x_limits: Update the x limits.
-            If None, then the limits are not updated.
-            If 'auto' and the x data points are all not the same value, then the limits will be auto updated.
-            If a tuple of values, then those values will be used for the limit. (default: 'auto')
-        :param y_limits: Update the x limits.
-            If None, then the limits are not updated.
-            If 'auto' and the y data points are all not the same value, then the limits will be auto updated.
-            If a tuple of values, then those values will be used for the limit. (default: 'auto')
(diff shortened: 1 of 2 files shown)


falkoschindler:
Thanks a lot for this pull request, @drkspace!

Checking for `x_min != x_max` and `y_min != y_max` is definitely a good idea. These warnings have been there for ages and can be fixed rather easily.

I just wonder if the two additional parameters `update_x_lims` and `update_y_lims` are a good choice. It makes sense to provide an opt out for these updates. But what exactly is the use case? If the users want to define custom limits themselves, they can already do so:
```py
line_plot.push([x], [[y]])
with line_plot:
    plt.ylim([0, 1])
```
Sure, this is not very intuitive, so we might want to provide a parameter like `ui.push(..., y_lim=[0, 1])`. But in combination with `update_y_lims` the API gets confusing. Therefore I'm thinking about something like
```py
def push(self, ..., y_limits: Union[None, Literal['auto'], Tuple[float, float]] = 'auto') -> None:
```
(same for `x_limits`). Passing `None` would leave the limits unchanged. The string "auto" would automatically update the limits like before. And a tuple would specify the new limits explicitly. (Note that we tend to avoid abbreviations in NiceGUI, so I'd prefer `y_limits` over `y_lim`.)

What do you think? Does that make sense, or am I missing something?

drkspace:
> If the users want to define custom limits themselves, they can already do so:

```python
line_plot.push([x], [[y]])
with line_plot:
    plt.ylim([0, 1])
```

I don't think this is a good solution since it would require knowing the needed limits everywhere push is called. I can certainly imagine a use case where someone has many different places they are calling `push`. This way would require them to pass around 2/4 different variables to all the places that need them.

> What do you think? Does that make sense, or am I missing something?

That makes sense to me. The reason I added the ability to just turn of the auto update is because, in my code, I am setting the limits in 1 part of a class and updating the plot in another. Your suggestion should also allow that. 

I've updated the PR.

falkoschindler (approved):
Great! I just shortened the docstring a little and decided to remove the extra demo because it was too similar to the first one. We can simply use the `y_limits` parameter there to make the point. I think the docstring is clear enough about its usage.

PR #4208, opened 2025-01-13
Remove duplicate section that is already in html_documentation.py

norlatif:
Currently on the website, this section already exists: https://nicegui.io/documentation/section_text_elements#other_html_elements
It exists here: https://nicegui.io/documentation/html (html_documentation.py)


website/documentation/content/section_text_elements.py
```diff
@@ -20,24 +20,3 @@
 doc.intro(restructured_text_documentation)
 doc.intro(mermaid_documentation)
 doc.intro(html_documentation)
```
norlatif, line 22: The removed code is inside the **doc.intro(html_documentation)** page

falkoschindler:
Hi @norlatif,

In PR #4027 we decided to copy&paste this section because we want the users to find it both unter "text elements" as well as when looking into the `ui.html` element. Duplicating the section isn't a very nice solution though. Maybe you have a better idea for keeping the information in both places? Maybe we show different demos and add links to the other demo?

norlatif:
> Hi @norlatif,
> 
> Maybe you have a better idea for keeping the information in both places?

Yes, maybe. See my new commit. I changed it to use DRY principles to prsent the same section in both places.

Issue #4016, opened 2024-11-20
`ui.download` doesn't work for all browsers connected to the auto-index page


PeterQuinn396:
### Description

First off, NiceGUI is a great tool and I love working with it, thank you so much for making it and supporting it!

# Problem

I've come across a strange behavior that I believe is potentially problematic.

I am using a nicegui webpage to monitor and control the state of a piece of hardware, using the auto index page to ensure that anyone viewing the webpage sees the same state of the hardware.

When using the auto index page with multiple clients (either different web browers on the same machine, or browsers on different machines), when one user clicks a `ui.button` element which triggers a `ui.download` call, the download is initiated on all connected clients. 

This triggers a race condition between the browsers / clients to download the file. The first client to access it receives the file, all the others show a download failed error, as the file download url can only be used once:
https://github.com/zauberzeug/nicegui/blob/1f78c863e7e2f3baaed03148ddc4e021f83eef97/nicegui/functions/download.py#L19

The situation can arise where the user who clicked the download button does not receive the file, but it goes to a user on a different machine, who also happens to be viewing the auto index page. 

This seems like a potentially problematic situation, as arbitrary files can be downloaded onto another persons machine, and the desired files don't up on the correct target machine. 

## Minimal working example

Using ubuntu 22 and python 3.10, and latest version of nicegui==2.5.0

1. Create a small file to download later 

    ```
    echo "sample text" >> small_file.txt
    ```

1. Run the following code from the same folder 

    ```python
    from nicegui import ui
    
    def main():
        ui.button("Download", on_click=lambda: ui.download('small_file.txt'))
        ui.run(reload=False)
    
    if __name__ == "__main__":
        main()

    ```

1. Open the URL from the terminal in two different webbrowers on the same machine (ex chrome and firefox)

1. Press the download button, both will attempt to download the file, but one will fail. It may be the browser that requested the download, or the other one, its not very predictable

    Closing one of the browsers fixes allows the single one to consistently download the file as expected

# Attempted fixes / investigation so far

I couldn't see anyway to work with `ui.download` function arguments to change this behaviour, so I looked for alternatives to have only the user that initiated the download receive the file.  

My understanding with NiceGUI is that per user interactions are done with the `ui.page()` decorator. So I tried a small example below:

```python

from nicegui import ui
from nicegui.events import ClickEventArguments

def main():

    # create a private page to handle the download, so it only triggers for the current user?
    def download(click_event_args: ClickEventArguments):
[...]

PR #4219, opened 2025-01-14
Keep single-use routes for a few seconds before pruning them

falkoschindler:
As discussed in issue #4016, `ui.download` can fail on the auto-index page if multiple tabs try to download from a single-use route at the same time. The first download will remove the route and subsequent downloads will fail.

This PR addresses this problem by keeping used single-use routes in a dictionary with an expiration timestamp. A pruning loop will check this dictionary every 10 seconds and remove outdated routes.

Note that simply delaying the removal after creating a single-use route did not work. The path of a single-use route is basically a hash of the filepath and, thus, not unique to a single call to `ui.download`. Therefore a route could be used again and again, so we need to keep it as long as `ui.download` is still being called.

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -23,6 +23,8 @@ from ..storage import Storage
 from .app_config import AppConfig
 from .range_response import get_range_response
 
+SINGLE_USE_ROUTE_TIMEOUT = 10.0
+
 
 class State(Enum):
     STOPPED = 0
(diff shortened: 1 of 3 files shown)


falkoschindler:
We decided to close this PR because single-use routes should really be for "single use" only. And triggering downloads on multiple browsers isn't really a use case. Instead we should add support for triggering downloads on individual sockets.

PR #4229, opened 2025-01-15
A couple of small improvements to the custom vue component example

josephburkhart:
Hi there, this is my first PR for this project and I'm somewhat new to collaborative Open Source development, so please be patient.

### Description
This PR modifies the Custom Vue Component [example](https://github.com/zauberzeug/nicegui/tree/main/examples/custom_vue_component) to make it a bit more helpful for developers who need to create custom components but haven't used Vue before.

### Motivation
In the example linked above, the Vue component is defined in `counter.js`. This way of bundling a component in a vanilla JavaScript file is obviously supported by Vue, since every built-in NiceGUI component is defined this way, but the official Vue documentation (and basically every other resource on the web) assumes that developers are writing `.vue` files, which are supported by additional tooling. I think the example should structure its Vue code in a way that makes it easier for new developers to understand how it works in relation to existing Vue documentation.

### Changes
In the new modified example, the python class definition now uses a new file, `counter.vue`, which re-implements `counter.js` in the more customary Vue Single-File Component format.

The call to `ui.run()` has been modified to ensure browser reload on changes to the `.js` and `.vue` files, which is important for iterative development on the Vue side.

### Testing
These changes do not add any new functionality that needs to be covered by tests. I have run the modified example and ensured that it does not error.

### Additional Notes
I am curious if the developers would be interested in adding a second, more advanced custom vue component example. I am in the process of creating a pannable and zoomable image component, which I would be happy to wrap up in a nice example and provide in a separate PR.


falkoschindler (approved):
Thanks a lot for this pull request, @josephburkhart! Showing how to include Single-File Components is a great idea.

We just decided to better split the component into two instead of instructing the user how to switch between js and vue files. While at it, I varied one of the components a little bit to make it a bit more interesting. And I included scoped style definitions, which is one of the main selling points for SFCs.

Let's merge!

Issue #3995, opened 2024-11-15
bindableProperty of an object's copy is not bindable anymore


balex89:
### Description

If we make a [`copy.copy()`](https://docs.python.org/3/library/copy.html) (or a [`deepcopy()`](https://docs.python.org/3/library/copy.html#copy.deepcopy)) of an object with a `bindableProperty`, the resulting copy's bindable attribute is not really bindable anymore (until its value is reassigned manually).

# Example

```py
import copy

from nicegui import binding


class MyClass:
    x = binding.BindableProperty()

    def __init__(self):
        self.x = 1


original = MyClass()
duplicate = copy.copy(original)

assert (id(original), 'x') in binding.bindable_properties  # ok
assert (id(duplicate), 'x') in binding.bindable_properties  # AssertionError
```

# Why is that?

The `copy` module techniques do not invoke `__setattr__()` at any point, so the `BindableProperty.__set__()` method never gets to update `binding.bindable_properties` with the new object's `id()`. 

And therefore, when binding, unexpectedly an active link will be created.   

# A workaround

A possible obvious approach might involve defining `__copy__` and `__deepcopy__` methods for a target class and updating `bindable_properties` in the process. 
This works just fine for a trivial case. But I personally don't like it much, as it lacks universality and might get trickier if complicated by descriptors or unnecessary `__init__()` logic.

What `copy` uses under the hood is the `pickle` approach. From the [docs](https://docs.python.org/3/library/copy.html):
> In fact, the [copy](https://docs.python.org/3/library/copy.html#module-copy) module uses the registered pickle functions from the [copyreg](https://docs.python.org/3/library/copyreg.html#module-copyreg) module.

That leads us to the following solution:
```py
import copy
import copyreg
from typing import TypeVar, Type, Any

from nicegui import binding

T = TypeVar('T')


def register_bindables(original_obj: T, copy_obj: T) -> None:
    """ Here we replicate the "`bindable_properties` representation" of an object
    WITHOUT any unnecessary attribute accessing in process.
    """
    for attr_name in dir(original_obj):
        if (id(original_obj), attr_name) in binding.bindable_properties:
            binding.bindable_properties[(id(copy_obj), attr_name)] = copy_obj


def pickle_hooked(obj: Any):
    """ This is a `obj.__reduce__()` wrapper that invoke bindables registration """
    reduced = obj.__reduce__()
    creator = reduced[0]

    def creator_with_hook(*args, **kwargs):
[...]

PR #4251, opened 2025-01-22
Copyable classes with bindable properties

balex89:
This PR follows #3995 issue discussion. 

I went with [Option  3](https://github.com/zauberzeug/nicegui/issues/3995#issuecomment-2598113869) for a list of reasons:
1. It uncompromisingly covers all aspects of the issue, including tracking of bindable attributes;
2. It doesn't really affect performance much, if done right (see below);
3. It isn't in fact as ridiculously complicated in terms of implementation as it might seem, once you see that it only adds a wrapper for function that reproduces an object instance. 

## Impact on performance 

Since we modify `BindableProperty` setter method, i ran a simple test to measure impact on code performance when object is a) initialized and b) updated:
```py
class TestClass:

    x = BindableProperty()

    def __init__(self):
        self.x = 1

    def update(self):
        self.x = 2

init_time = timeit.timeit(lambda: TestClass(), number=1_000_000)
a = TestClass()
update_time = timeit.timeit(lambda: a.update(), number=1_000_000)
print(f'creation: {init_time:.6f} s, update: {update_time:.6f} s')
````
I ran this test several times for the current and updated `BindableProperty` implementations and got following average figures:
1. For initialization it adds up to 5% performance time on average;
2. For update it adds ~1% (or less) time on average.

Since a user, aiming to create millions of instances, each meant for binding, will probably face more sever issues for other reasons, the real-life overhead seems to be insignificant.


falkoschindler (approved):
Hi @balex89, thank you so much for your pull request and the patience it took until I finally reviewed your code!

I have to admit, I like this solution. It isn't as complicated as expected. And I even simplified it a little more:
- I merged some functions into one. In my point of view it is more comprehensible when all copy-related code is in one place.
- The docstrings seemed a bit redundant, so I reduced them to one line. To understand the details, you have to look into the code anyway.
- We can use the `copyreg.dispatch_table` instead of our own set of `copyable_classes`.

Let's merge this PR!

PR #4264, opened 2025-01-24
User testing, save response and print method

Mte90:
As per title now the response is saved so it is possible to get more information like the URL where the user is or check the response code.

Also the `print` function probably isn't the best but let's to print something in the pytest output as it isn't very easy to do that.

I was trying to get the actual URL after the various redirects but `response.history` with:
```py
async def test_navigate_to(user: User) -> None:
    @ui.page('/')
    def index() -> None:
        ui.navigate.to('/login')

    @ui.page('/login')
    def login() -> None:
        ui.input('Username')

    await user.open('/')
    user.print(user.response.history)
    # await user.should_see('Username')
    user.find('Username').type('admin')
```

Return an empty list, I think because the redirect happens next the open so at that time there wasn't the reidrect but I am not finding a way to get the actual url where it is the user.

### DIFF
--- a/nicegui/testing/user.py
+++ b/nicegui/testing/user.py
@@ -35,6 +35,7 @@ class User:
         self.navigate = UserNavigate(self)
         self.notify = UserNotify()
         self.download = UserDownload(self)
+        self.response = None
 
     @property
     def _client(self) -> Client:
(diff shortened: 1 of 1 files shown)


Mte90:
So my idea to use `page_routes` return the routes registered not the one the user visited including redirects requires something more advanced.

rodja:
I suggest you start three different feature requests, so we can discuss problem, implementation and benefits individually.

Issue #4253, opened 2025-01-22
White screen after socket reconnect


chriswi93:
### Description

Hello,

at first I would like to take this opportunity to praise this great project. I think NiceGUI is a really great package for quickly developing a user interface with Python. Many thanks to the developers of this project!

However, we have been observing an issue with NiceGUI for some time now and we could not find a way to fix it. We hope someone here can help to fix this bug as we don't know what is going on under the hood of NiceGUI. The issue is quite critical for us and we have already put a lot of effort into debugging the issue, but have had little success so far. According to our observations, the issue is not limited to the latest versions and also occurs in earlier versions < 2.0.

The issue is quite difficult to debug. Below I will try to describe my observations as best as I can, but unfortunately I can't provide a minimal reproducible code example. It seems like every NiceGUI application is affected by this issue, even very simple applications, and so far I have not found a way to actually reproduce it.

## Description
We have noticed for some time now that the socket connection of our NiceGUI application in the browser is regularly closed from time to time. Afterwards the entire content of the application in the browser is deleted and a white screen is shown to the user. Contrary to our expectations, a reload of the page is not triggered. Sometimes this happens after ten minutes, sometimes after three hours or later:

![Image](https://github.com/user-attachments/assets/c0afe798-a37b-4f43-b339-cb4427a45c37)

There is no clear pattern that could indicate what is causing the issue. It is unclear why the connection is closed, as there does not appear to be an actual network problem. As you can see in the image, a new socket connection is successfully established immediately after the initial socket connection is closed. However, after the second socket connection is established, the server sends a “clear screen” instruction to the browser:

```
[...]

PR #4271, opened 2025-01-25
Keep multiple disconnect tasks in a list

falkoschindler:
This PR might solve #4253 by keeping a list of disconnect tasks instead of just one.
I'm marking it as "draft" because ultimately I'd like to better understand what is happening. How can a disconnect task be overwritten before being canceled? This would require two disconnects without a handshake in between. 🤔


falkoschindler:
Let's close this PR in favor of PR #4285.

PR #4285, opened 2025-01-28
Only delete page content if the number of connections is zero

falkoschindler:
This PR tries to solve #4253 with a new `_num_connections` counter. On handshake the counter is increased, on disconnect it is decreased. On disconnect, after awaiting the reconnect timeout, the page content is only deleted if the number of connections is zero.

Possible scenarios:

- normal reconnect:
    ```
    handshake (num=1) ... disconnect (num=0) ... handshake (num=1) ... after delay: don't delete
    ```
- strange reconnect with handshake before disconnect:
    ```
    handshake (num=1) ... handshake (num=2) ... disconnect (num=1) ... after delay: don't delete
    ```
- normal disconnect:
    ```
    handshake (num=1) ... disconnect (num=0) ... after delay: delete
    ```

While working on the handshake and disconnect implementation, I noticed an inconsistency with `app.on_connect` and `app.on_disconnect`: While connect handlers are called on reconnect, disconnect handlers are not. In this PR I changed it so that both handlers are called during reconnection. We should clarify if this is the desired behavior.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -237,9 +237,7 @@ class Client:
 
     def handle_handshake(self, next_message_id: Optional[int]) -> None:
         """Cancel pending disconnect task and invoke connect handlers."""
-        if self._delete_task:
-            self._delete_task.cancel()
-            self._delete_task = None
+        self._cancel_delete_task()
         self._num_connections += 1
         if next_message_id is not None:
             self.outbox.try_rewind(next_message_id)
@@ -251,6 +249,7 @@ class Client:
 
     def handle_disconnect(self) -> None:
         """Wait for the browser to reconnect; invoke disconnect handlers if it doesn't."""
+        self._cancel_delete_task()
         self._num_connections -= 1
         for t in self.disconnect_handlers:
             self.safe_invoke(t)
@@ -262,6 +261,11 @@ class Client:
                 if self._num_connections == 0:
                     self.delete()
             self._delete_task = background_tasks.create(delete_content())
+
+    def _cancel_delete_task(self) -> None:
+        if self._delete_task:
+            self._delete_task.cancel()
+            self._delete_task = None
 
     def handle_event(self, msg: Dict) -> None:
         """Forward an event to the corresponding element."""
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -59,7 +59,7 @@ class Client:
         self.environ: Optional[Dict[str, Any]] = None
         self.shared = request is None
         self.on_air = False
-        self._disconnect_task: Optional[asyncio.Task] = None
+        self._num_connections = 0
         self._deleted = False
         self.tab_id: Optional[str] = None
 
@@ -236,28 +236,25 @@ class Client:
 
     def handle_handshake(self, next_message_id: Optional[int]) -> None:
         """Cancel pending disconnect task and invoke connect handlers."""
+        self._num_connections += 1
         if next_message_id is not None:
             self.outbox.try_rewind(next_message_id)
-        if self._disconnect_task:
-            self._disconnect_task.cancel()
-            self._disconnect_task = None
         storage.request_contextvar.set(self.request)
         for t in self.connect_handlers:
             self.safe_invoke(t)
         for t in core.app._connect_handlers:  # pylint: disable=protected-access
             self.safe_invoke(t)
 
-    def handle_disconnect(self) -> None:
+    async def handle_disconnect(self) -> None:
         """Wait for the browser to reconnect; invoke disconnect handlers if it doesn't."""
-        async def handle_disconnect() -> None:
-            await asyncio.sleep(self.page.resolve_reconnect_timeout())
-            for t in self.disconnect_handlers:
-                self.safe_invoke(t)
-            for t in core.app._disconnect_handlers:  # pylint: disable=protected-access
-                self.safe_invoke(t)
-            if not self.shared:
-                self.delete()
-        self._disconnect_task = background_tasks.create(handle_disconnect())
+        self._num_connections -= 1
+        for t in self.disconnect_handlers:
+            self.safe_invoke(t)
+        for t in core.app._disconnect_handlers:  # pylint: disable=protected-access
+            self.safe_invoke(t)
+        await asyncio.sleep(self.page.resolve_reconnect_timeout())
+        if self._num_connections == 0 and not self.shared:
+            self.delete()
 
     def handle_event(self, msg: Dict) -> None:
         """Forward an event to the corresponding element."""
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -60,6 +60,7 @@ class Client:
         self.shared = request is None
         self.on_air = False
         self._num_connections = 0
+        self._delete_task: Optional[asyncio.Task] = None
         self._deleted = False
         self.tab_id: Optional[str] = None
 
@@ -236,6 +237,9 @@ class Client:
 
     def handle_handshake(self, next_message_id: Optional[int]) -> None:
         """Cancel pending disconnect task and invoke connect handlers."""
+        if self._delete_task:
+            self._delete_task.cancel()
+            self._delete_task = None
         self._num_connections += 1
         if next_message_id is not None:
             self.outbox.try_rewind(next_message_id)
@@ -245,16 +249,19 @@ class Client:
         for t in core.app._connect_handlers:  # pylint: disable=protected-access
             self.safe_invoke(t)
 
-    async def handle_disconnect(self) -> None:
+    def handle_disconnect(self) -> None:
         """Wait for the browser to reconnect; invoke disconnect handlers if it doesn't."""
+        async def delete_content() -> None:
+            await asyncio.sleep(self.page.resolve_reconnect_timeout())
+            if self._num_connections == 0:
+                self.delete()
         self._num_connections -= 1
         for t in self.disconnect_handlers:
             self.safe_invoke(t)
         for t in core.app._disconnect_handlers:  # pylint: disable=protected-access
             self.safe_invoke(t)
-        await asyncio.sleep(self.page.resolve_reconnect_timeout())
-        if self._num_connections == 0 and not self.shared:
-            self.delete()
+        if not self.shared:
+            self._delete_task = background_tasks.create(delete_content())
 
     def handle_event(self, msg: Dict) -> None:
         """Forward an event to the corresponding element."""
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -4,6 +4,7 @@ import asyncio
 import inspect
 import time
 import uuid
+from collections import defaultdict
 from contextlib import contextmanager
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, ClassVar, Dict, Iterable, Iterator, List, Optional, Union
@@ -59,7 +60,7 @@ class Client:
         self.environ: Optional[Dict[str, Any]] = None
         self.shared = request is None
         self.on_air = False
-        self._num_connections = 0
+        self._num_connections = defaultdict(int)
         self._delete_task: Optional[asyncio.Task] = None
         self._deleted = False
         self.tab_id: Optional[str] = None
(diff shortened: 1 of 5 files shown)


falkoschindler:
Something in this PR seems to slow down our pytests. Usually the ~500 tests run in around 15 minutes per machine. Now they timeout after 30 minutes.

nicegui/client.py
```diff
@@ -236,28 +236,25 @@ def on_disconnect(self, handler: Union[Callable[..., Any], Awaitable]) -> None:
...
-                self.safe_invoke(t)
-            if not self.shared:
-                self.delete()
-        self._disconnect_task = background_tasks.create(handle_disconnect())
+        self._num_connections -= 1
+        for t in self.disconnect_handlers:
+            self.safe_invoke(t)
+        for t in core.app._disconnect_handlers:  # pylint: disable=protected-access
+            self.safe_invoke(t)
+        await asyncio.sleep(self.page.resolve_reconnect_timeout())
```
chriswi93, line 255: I noted just one minor difference here if ``_num_connections = 2`` and ``handle_disconnect()`` is called:
The event loop will be released at ``asyncio.sleep()`` and there is the possibility that in the meanwhile another disconnect happens which calls ``handle_disconnect()`` a second time and decreases ``_num_connections`` to 0. So ``delete`` will be called twice after sleep:
```
async task 1
    call handle_disconnect()
    _num_connections - 1 = 1
    asyncio.sleep(..)

Event loop context switch: async task 2
    call handle_disconnect()
    _num_connections - 1 = 0
   asyncio.sleep(..)

Context switch: async task 1
   delete()

Context switch: async task 2
   delete()
```

Instead of
```python
await asyncio.sleep(self.page.resolve_reconnect_timeout())
```
I would prefer to stop before sleep if we expect another disconnect in the future that finally cleans up resources:
```python
if self._num_connections > 0:
   return
await asyncio.sleep(self.page.resolve_reconnect_timeout())
```

chriswi93:
> Something in this PR seems to slow down our pytests. Usually the ~500 tests run in around 15 minutes per machine. Now they timeout after 30 minutes.

If I had to guess I would say that ``await client.handle_disconnect()`` blocks until ``asyncio.sleep()`` is finished. There was a background task before and ``client.handle_disconnect()`` immediately returned.

nicegui/client.py
falkoschindler: You're right, @chriswi93. There might be the possibility for two disconnect handlers to end up deleting the page content. As mentioned below, this doesn't seem to be the only flaw, so I slightly changed the implementation. It cancels any deletion if there is another connection event, so only the last disconnect causes the page content to be deleted.

chriswi93:
> That's a good explanation, @chriswi93! So it is better to sleep in a background task because they are gracefully canceled during teardown.
> 
> Apart from that we noticed a theoretical flaw of the connection counter. The delay could end right in between a "normal reconnect" when the disconnect happened and the counter `num` is 0. This would cause the disconnect to delete the page impermissibly (⚠️):
> 
> ```
> num: 1            0   1     0   1           2   1                 0
>      H .......... D ----------> delete because num=0 ⚠️
>                       H ... D ----------> no delete because num>0
>                                 H ............. D ----------> no delete because num>0
>                                             H ................... D ----------> delete because num=0
> ```
> 
> To overcome this problem, we can get back to using a background task for the content deletion, which is canceled (❌) by subsequent handshakes:
> 
> ```
> num: 1            0   1     0   1           2   1                 0
>      H .......... D --> ❌
>                       H ... D --> ❌
>                                 H ............. D ----------> no delete because num>0
>                                             H ................... D ----------> delete because num=0
> ```

Good point! I think we are quite close 🙂
I tested your code from today and it seems to resolve the white screen issue!

But it turns out that it introduces a breaking change I was not aware of and I think many applications built on top of NiceGUI would be affected in the same way:
Our service uses a disconnect handler to clean up resources after a client disconnects. Since disconnect handlers are not called once per client just before delete, but now for each disconnect and even for reconnect, you can not use disconnect handlers anymore as reliable source to decide if a client has finally gone or not. So the resources for our service are cleaned up too early.

I would suggest two options to fix this:

1. Call connect and disconnect handlers right after the if statement inside the background task:
```python
if self._num_connections == 0:
   ...
```
And document that there is an intended inconsistency: A connect handler might be called many times for a single client, but disconnect handler is only called once per client just before resources are cleaned up.

2. Add another ```on_delete()``` handler such that applications can reliable clean up their resources when a client is finally gone.

falkoschindler:
@chriswi93 You're right, calling disconnect handlers more than once can break user code. Even though you can argue that it was a bug and should be fixed, the impact is quite severe.

Besides your option 1 (keep inconsistent behavior like before) and option 2 (consistent behavior but breaking change), I see another option 3 that calls both handlers only once. It would also be a breaking change, but maybe less critical.

1. Multiple connects, single disconnect
2. Multiple connects, multiple disconnects, single delete
3. Single connect, single disconnect

But it's hard to tell if there are users relying on a connect handler being called on reconnect.

chriswi93:
@falkoschindler Yes, that would be another way to make it consistent. All three options are valid in my opinion.
Option 1 might be unexpected behavior and if you don't know about it your application recreates or overwrites existing resources for a client.
Option 2 is the most flexible option, but at the moment I can not see any reason why you would need to know about reconnects if the connection is not completly lost.
Option 3 makes it simple for most users without having to think about reconnects.

falkoschindler:
Ok, I'm about to keep the current behavior of multiple connects and a single disconnect (option 1). We can improve consistency with a breaking change in version 3.0, but should fix this bug first.

However, I noticed another aspect we didn't think about: On the shared auto-index client there are multiple open connections. Currently we call multiple connects and a single disconnect like on regular pages. But the `_num_connections` counter doesn't work for multiple concurrent connections.

So I thought we could use a `defaultdict[int]` instead with some connection ID:
- We could use the `tab_id`, but it isn't awailable in the `@sio.on('disconnect')` handler.
- Or we use the socket ID, but it isn't awailable via On Air in the `@self.relay.on('client_disconnect')` handler.

Because we need to add socket IDs to On Air messages anyway (see #4218), I suggest to do that first and use them for the connection dictionary.

chriswi93:
> Ok, I'm about to keep the current behavior of multiple connects and a single disconnect (option 1). We can improve consistency with a breaking change in version 3.0, but should fix this bug first.

Yes, I think it's a good idea to postpone the breaking change to a major release.

> However, I noticed another aspect we didn't think about: On the shared auto-index client there are multiple open connections. Currently we call multiple connects and a single disconnect like on regular pages. But the `_num_connections` counter doesn't work for multiple concurrent connections.
> 
> So I thought we could use a `defaultdict[int]` instead with some connection ID:
> 
> * We could use the `tab_id`, but it isn't awailable in the `@sio.on('disconnect')` handler.
> * Or we use the socket ID, but it isn't awailable via On Air in the `@self.relay.on('client_disconnect')` handler.
> 
> Because we need to add socket IDs to On Air messages anyway (see #4218), I suggest to do that first and use them for the connection dictionary.

As far as I can see you share a single Client instance and all the elements among many users and tabs. So all users share the same client id. If this is true, using the socket id could resolve it partially. However, I think that the current code won't resolve the issue due to the following reason:

I would assume that a new socket id is assigned to **every** new socket connection and I guess this also applies to a connection that is the result of a reconnect. If my assumption is true, the counter in ```_num_connections``` will never exceed 1 for any socket id. So for clients that are not shared, the client resources are deleted too early on the first socket disconnect even if the ```_num_connections``` dict still contains another socket id.

### For non shared clients
If I am right, it should be fine to use a list and keep track of ``_socket_ids`` instead of ```_num_connections```. All socket connections for a client are closed if ``len(self._socket_ids) == 0``.

### For shared auto-index client
If all users share the same client id, information is lost since we can no longer use the client id to identify an individual user tab. Therefore we don't know when all connections for an individual user tab are actually closed and as a result it is hard to find out when the disconnect handlers should be called.

I don't know if tab id is something that you could use instead for the shared auto-index page. Otherwise, it might be a good idea to add a unique id that makes it possible to identify a unique user tab across reconnects and regardless of whether it is a shared client or not.

If this would involve too many changes, it might be sufficient for now to call the disconnect handlers for each disconnect event if it is a shared client:

```python
if self.shared or len(self._socket_ids) == 0:
    ...
```
Instead of:
```python
if self._num_connections[socket_id] == 0:
    ...
```
However, it should be noted that an application built on top of the auto-index feature can not allocate/deallocate resources for a single user tab which could be an intended limitation. Therefore the behavior when the disconnect handlers are called is also different.

Apart from that, I have a minor remark:
In the background task ``_num_connections`` is cleaned up when a client is deleted
```python
self._num_connections.pop(socket_id)
```
but the ``_delete_tasks`` dict is not cleaned up properly.

chriswi93:
> Ok, I'm about to keep the current behavior of multiple connects and a single disconnect (option 1). We can improve consistency with a breaking change in version 3.0, but should fix this bug first.

Yes, I think it's a good idea to postpone the breaking change to a major release.

> However, I noticed another aspect we didn't think about: On the shared auto-index client there are multiple open connections. Currently we call multiple connects and a single disconnect like on regular pages. But the `_num_connections` counter doesn't work for multiple concurrent connections.
> 
> So I thought we could use a `defaultdict[int]` instead with some connection ID:
> 
> * We could use the `tab_id`, but it isn't awailable in the `@sio.on('disconnect')` handler.
> * Or we use the socket ID, but it isn't awailable via On Air in the `@self.relay.on('client_disconnect')` handler.
> 
> Because we need to add socket IDs to On Air messages anyway (see #4218), I suggest to do that first and use them for the connection dictionary.

As far as I can see you share a single Client instance and all the elements among many users and tabs. So all users share the same client id. However, I think the socket id won't resolve the issue due to the following reason:

I would assume that a new socket id is assigned to **every** new socket connection and I guess this also applies to a connection that is the result of a reconnect.

### For non shared clients
If my assumption is true, the counter in ```_num_connections``` will never exceed 1 for any socket id and the client resources are deleted too early on the first socket disconnect even if the ```_num_connections``` dict still contains another socket id. Therefore, the socket id has no additional benefit and it should be fine to keep ``_num_connections`` as a simple integer counter.

### For shared auto-index client
If all users share the same client id, information is lost since we can no longer use the client id to identify an individual user tab. Therefore we don't know when all connections for an individual user tab are actually closed and as a result it is hard to find out when the disconnect handlers should be called.

I don't know if tab id is something that you could use instead for the shared auto-index page. Otherwise, it might be a good idea to add a unique id that makes it possible to identify a unique user tab across reconnects and regardless of whether it is a shared client or not.

If this would involve too many changes, it might be sufficient for now to just ignore the ``_num_connections`` counter for a shared client inside the background task:
```python
# for self.shared = True: asyncio.sleep(...) is also not required
if self.shared or self._num_connections == 0:
    ...
```
Instead of:
```python
if self._num_connections[socket_id] == 0:
    ...
```
Then for a shared client the disconnect handlers would be called after each disconnect event. However, it should be noted that an application built on top of the auto-index feature can not allocate/deallocate resources for a single user tab which could be an intended limitation. Therefore, in my opinion it is consistent to say that the behavior of how disconnect handlers are called is different for a shared client.

falkoschindler:
> I would assume that a new socket id is assigned to **every** new socket connection and I guess this also applies to a connection that is the result of a reconnect.

Oh, you're correct!

> [...] it might be a good idea to add a unique id that makes it possible to identify a unique user tab across reconnects and regardless of whether it is a shared client or not.

In e65b1366a42da5d206d8292f97c2d2976653f142 I introduced a new `document_id` which is created randomly when JavaScripts mounts the frontend. It remains constant during reconnects, but changes when reloading the page - independent of whether we are on the auto-index page or not.

Here's an overview of all client-related IDs. It shows how `document_id` fills the gap between `tab_id` and `socket_id`, allowing to distinguish page reloads from reconnects:
```
                                      client_id  tab_id  document_id  socket_id  
different per tab on auto-index page        -       X            X          X
different after reload                      -       -            X          X
different after reconnect                   -       -            -          X
```

This should resolve the flaw of my previous attempt 🤞🏻

By the way, the name "document_id" derives from the JavaScript object `document`, which is usually the main anchor point for the HTML DOM. I also thought about `dom_id`, which might be a little more cryptic for non-web developers. I'm open for alternative suggestions though.

nicegui/client.py
```diff
@@ -234,30 +237,45 @@ def on_disconnect(self, handler: Union[Callable[..., Any], Awaitable]) -> None:
...
-                self.safe_invoke(t)
-            if not self.shared:
-                self.delete()
-        self._disconnect_task = background_tasks.create(handle_disconnect())
+            if self._num_connections[document_id] == 0:
+                for t in self.disconnect_handlers:
+                    self.safe_invoke(t)
+                for t in core.app._disconnect_handlers:  # pylint: disable=protected-access
+                    self.safe_invoke(t)
+                self._num_connections.pop(document_id)
```
chriswi93, line 271: I think we should also clear up ``self._delete_tasks`` after the final disconnect

chriswi93:
Looks all fine from my side! I had just one last remark.
Thanks a lot for your help @falkoschindler

nicegui/client.py
falkoschindler: Oh yes, of course!

falkoschindler:
Thanks again for all your valuable input and testing, @chriswi93!
Looks like this PR is working now and ready to merge 🚀

PR #4324, opened 2025-02-08
Add zh-CN & zh-TW support when server down.

Yuerchu:
When the client fails to connect to NiceGUI, a prompt will appear in the lower left corner of the client: `Connection Lost. Trying to reconnect...`. This is not difficult to understand for those who know English. However, for those who do not understand English, it undoubtedly reduces the user experience. 

This pull request adds support for Simplified Chinese and Traditional Chinese. After using `ui.run(language='zh-CN')` and the browser is fully loaded, if the server is closed, the correct display on the user's browser will be `连接丢失` instead of `Connection Lost`. If an undefined language is set， the default English prompt message will be displayed. I am still thinking of a better way for developers to customize this content.

> ui.run(language='zh-CN')
zh-CN

> ui.run(language='zh-TW')
zh-TW

> ui.run() # Specify another language or do not specify any language
en-US

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -23,6 +23,7 @@ from .javascript_request import JavaScriptRequest
 from .logging import log
 from .observables import ObservableDict
 from .outbox import Outbox
+from .translations import translations
 from .version import __version__
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


Yuerchu:
I have added languages from other countries/regions as well, such as Japanese `ja-JP`:
image

falkoschindler (approved):
Thanks a lot for this pull request, @Yuerchu!
Or should I say: 非常感谢您的贡献! 😉

I just moved the translation definition into a separate Python file. This way we can resolve the translation right on the server when rendering the HTML template and don't have to send the whole dictionary to the frontend.

Let's merge! 🚀

PR #4327, opened 2025-02-09
Removing list args when emitting events

valleedelisle:
Apparently, the args are not passed when we use ... in front of args. Removing the dots solves the issue.

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -80,7 +80,7 @@ function getComputedProp(target, prop_name) {
 }
 
 function emitEvent(event_name, ...args) {
-  getElement(0).$emit(event_name, ...args);
+  getElement(0).$emit(event_name, args);
 }
 
 function stringifyEventArgs(args, event_args) {


falkoschindler:
Sorry, @valleedelisle, I don't understand the problem with the current implementation.
Please explain in more detail. Can you give a minimum reproducible example?

valleedelisle:
Sorry about the lack of details.

I have this javascript code:

```javascript
emitEvent("delete_object", { id: my_object_id });
```

Which I listen to with this:
```python
ui.on("delete_schedule", handler=lambda e:print(f"Event received {e} for object id {e.args}"))
```

`e.args` is always empty unless I remove the `...` from the `$emit` argument like proposed here.

Unless the `emitEvent` syntax is wrong? I'm no javascript guru. I could find any doc or example with `emitEvent` arguments.

falkoschindler:
Strange, this works for me:
```py
ui.button('Send event', on_click=lambda: ui.run_javascript('emitEvent("test", {id: 42})'))
ui.on('test', lambda e: ui.notify(f'Received "test" event with args {e.args}'))
```

With the current implementation
```js
function emitEvent(event_name, ...args) {
  getElement(0).$emit(event_name, ...args);
}
```
all arguments of `emitEvent` should be directly forwarded to `$emit` which sends them to the server.

falkoschindler:
I'm closing this pull request because its purpose is still not clear. Feel free to re-open it if it is still relevant to you.

PR #4332, opened 2025-02-10
Replace `bindable_properties` dictionary with a set

falkoschindler:
This PR tries to fix issue #4109 by removing references to bound objects from the binding module to avoid memory leaks.


falkoschindler:
I'm closing this PR in favor of PR #4122, which also takes care of pruning the `bindable_property` collection.

Issue #367, opened 2023-02-12
Drawers behave weirdly since they are derived from ValueElement in #308


falkoschindler:
In #308 we derived `ui.left_drawer` and `ui.right_drawer` from `ValueElement` in order to allow setting their initial state. As it turns out, this causes weird mobile behavior:

- On small screens Quasar's drawers are initially hidden. But the default `value=True` overwrites this automation and always shows the drawer, even on small screens.
- A workaround might be to set the initial value to `None` and to re-activate the prop `show-if-above` to automatically show the drawer above a certain media query width. But this triggers a value change on clients with a large screen right after loading the HTML, when the socket is not yet connected. The JavaScript console shows an error and the `value` on the server is not updated. This causes the next `toggle()` call being uneffective.

Switching to a default value of `None` and setting `show-if-above` in this case only seems promising. But we have to find out how to inform the server about a possible automatic value change.

Or we ditch the `ValueElement` approach and send the `value` and other props only one-way from server to client. But this wouldn't allow providing a `toggle()` method, since the server does not know the current `value`.

falkoschindler:
> Or we ditch the ValueElement approach and send the value and other props only one-way from server to client. But this wouldn't allow providing a toggle() method, since the server does not know the current value.

I just noticed there is a [client-side `toggle()` method](https://quasar.dev/layout/drawer) we could use.

falkoschindler:
@rodja Removing the inheritance from `ValueElement` I found a solution I'm pretty pleased with. If you agree, we could release version 1.1.7 shortly before users start to rely on drawers being `ValueElements` with a bindable `value` property.

rodja:
Sure. Releasing 1.1.7 quickly would be good. Go ahead.

Issue #308, opened 2023-01-28
Provide example to use hamburger menu button for drawer expanding/collapsing


wwoods:
Quasar seems to allow a hamburger menu button that expands or collapses a drawer: https://github.com/quasarframework/quasar/blob/dev/docs/public/examples/QDrawer/Menu.vue

It would be nice to see how to do this within NiceGUI, since it's a standard layout choice for more real estate.

falkoschindler:
In [517ec0f](https://github.com/zauberzeug/nicegui/commit/517ec0fa8a1478533e36eea9096ff63fef764f0b) I improved the page layout elements:

- added docstrings,
- derived header, footer and drawers from `ValueElement` so that you can toggle their `value` property,
- added new initializer parameters, and
- improved the type annotation using literal strings here and there.

Here is an example for hamburger menu buttons toggling all four panels:

```py
ui.label('This is the page content')

with ui.header().classes('items-center') as header:
    ui.button(on_click=lambda: left_drawer.set_value(not left_drawer.value)).props('flat color=white icon=menu')
    ui.label('This is a header').classes('mx-auto')
    ui.button(on_click=lambda: right_drawer.set_value(not right_drawer.value)).props('flat color=white icon=menu')

with ui.left_drawer().classes('bg-blue-100') as left_drawer:
    ui.label('This is a left drawer')

with ui.right_drawer().classes('bg-blue-100') as right_drawer:
    ui.label('This is a right drawer')

with ui.footer() as footer:
    ui.label('This is a footer')

with ui.page_sticky(position='top-right', x_offset=20, y_offset=20):
    ui.button(on_click=lambda: header.set_value(not header.value)).props('fab icon=menu')

with ui.page_sticky(position='bottom-right', x_offset=20, y_offset=20):
    ui.button(on_click=lambda: footer.set_value(not footer.value)).props('fab icon=menu')
```

falkoschindler:
I also updated the example in the API reference.

@rodja I think we can close this issue. Do you agree?

rodja:
Yes! I just pushed a new example combining a foldable menu with tabs: https://github.com/zauberzeug/nicegui/commit/5e1610bad6216e20a148fd6f221e41921523b268.

falkoschindler:
Small caveat:
Setting the initial `value` for a drawer to `False` doesn't seem to work.
But it works for headers and footer.

rodja:
> Setting the initial value for a drawer to False doesn't seem to work.

[...]

PR #4334, opened 2025-02-10
Drawer inherits ValueElement

xaptronic:
https://github.com/zauberzeug/nicegui/discussions/4320

Drawer/LeftDrawer/RightDrawer does not currently inherit from `ValueElement` and so property binding here does not work.

This PR updates Drawer component to inherit from ValueElement and changes it's toggle/hide/show methods to use `value` property instead of calling the Q elements show/hide/toggle.

Example

```py
from nicegui import ui

with ui.header():
    ui.label("Fancy App")

with ui.left_drawer(value=True, elevated=True) as left_drawer:
    ui.label("Left Drawer Content")

with ui.right_drawer(value=True, elevated=True) as right_drawer:
    ui.label("Right Drawer Content")

with ui.row():
    with ui.card():
        ui.label("Left Drawer")
        ui.button("Show", on_click=lambda: left_drawer.show())
        ui.button("Hide", on_click=lambda: left_drawer.hide())
        ui.button("Toggle", on_click=lambda: left_drawer.toggle())

    with ui.card():
        ui.label("Right Drawer")
        ui.button("Show", on_click=lambda: right_drawer.show())
        ui.button("Hide", on_click=lambda: right_drawer.hide())
        ui.button("Toggle", on_click=lambda: right_drawer.toggle())

ui.checkbox("Left Drawer State").bind_value(left_drawer, "value")
ui.checkbox("Right Drawer State").bind_value(right_drawer, "value")

ui.run(port=9001)
```

### DIFF
--- a/nicegui/page_layout.py
+++ b/nicegui/page_layout.py
@@ -83,7 +83,7 @@ class Header(ValueElement, default_classes='nicegui-header'):
         self.value = False
 
 
-class Drawer(Element, default_classes='nicegui-drawer'):
+class Drawer(ValueElement, default_classes='nicegui-drawer'):
 
     def __init__(self,
                  side: DrawerSides, *,
@@ -112,11 +112,9 @@ class Drawer(Element, default_classes='nicegui-drawer'):
         """
         _check_current_slot(self)
         with context.client.layout:
-            super().__init__('q-drawer')
+            super().__init__(tag='q-drawer', value=value, on_value_change=None)
         if value is None:
             self._props['show-if-above'] = True
-        else:
-            self._props['model-value'] = value
         self._props['side'] = side
         self._props['bordered'] = bordered
         self._props['elevated'] = elevated
@@ -131,15 +129,15 @@ class Drawer(Element, default_classes='nicegui-drawer'):
 
     def toggle(self) -> None:
         """Toggle the drawer"""
-        self.run_method('toggle')
+        self.value = not self.value
 
     def show(self) -> None:
         """Show the drawer"""
-        self.run_method('show')
+        self.value = True
 
     def hide(self) -> None:
         """Hide the drawer"""
-        self.run_method('hide')
+        self.value = False
 
 
 class LeftDrawer(Drawer):
--- a/nicegui/page_layout.py
+++ b/nicegui/page_layout.py
@@ -113,8 +113,7 @@ class Drawer(ValueElement, default_classes='nicegui-drawer'):
         _check_current_slot(self)
         with context.client.layout:
             super().__init__(tag='q-drawer', value=value, on_value_change=None)
-        if value is None:
-            self._props['show-if-above'] = True
+        self._props['show-if-above'] = value is None
         self._props['side'] = side
         self._props['bordered'] = bordered
         self._props['elevated'] = elevated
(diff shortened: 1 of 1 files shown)


falkoschindler (changes requested):
Hi @xaptronic, thanks for another valuable pull request!

During review I stumbled over the line `self._props['show-if-above'] = True` which I traced back to #367 from two years ago. As it turns out - and I totally forgot when you asked me about it in #4320 - we already tried deriving layout elements from `ValueElement` in #308. But as described in #367, there is a problem which also occurs with your implementation: The model value isn't in sync with the server when opening the page on a large (non-mobile) screen. This can be reproduced with the following code snippet.

```py
with ui.left_drawer(bordered=True) as drawer:
    ui.label('Side menu')

ui.button('Toggle drawer', on_click=drawer.toggle)
ui.label().bind_text_from(drawer, 'value', lambda value: 'Open' if value else 'Closed')
```

I'm not sure if there's an easy fix for this problem. Do you have an idea?

codingpaula:
Hello everyone,

I guess that's a rather hacky solution with the drawback that it won't work on the auto-index page due to the use of  `run_javascript`. But for my use case it's currently enough:

```python
class Drawer(ValueElement, default_classes='nicegui-drawer'):

    def __init__(self,
                 side: DrawerSides, *,
                 value: Optional[bool] = None,
                 fixed: bool = True,
                 bordered: bool = False,
                 elevated: bool = False,
                 top_corner: bool = False,
                 bottom_corner: bool = False) -> None:
        """Drawer documentation ..."""
        _check_current_slot(self)
        with context.client.layout:
            super().__init__(tag='q-drawer', value=value, on_value_change=None)
        if value is None:
            self._props['show-if-above'] = True
            timer(0., self._correct_value, once=True)
        else:
            self._props['model-value'] = value
        # ... rest of __init__ as is

    async def _correct_value(self) -> None:
        class_list: Dict[str, str] = await context.client.run_javascript(
            f'return document.querySelector(".q-drawer--{self._props["side"]}").classList',
        )
        self.value = 'q-layout--prevent-focus' not in class_list.values()

    def toggle(self) -> None:
        """Toggle the drawer"""
        self.value = not self.value
```

What do you guys think?

Alternatively, I'd argue for making `value` a required argument and adding a `show_if_above` argument (default `False`). This way we could keep the need to use the `_correct_value` function to a minimum and warn if `show_if_above` is used on an auto-index page.

falkoschindler:
Thanks, @codingpaula! That's an interesting idea.

I think this unveils an inherent problem with making `ui.drawer` a `ValueElement`: It wants to hold the current state on the server. But - when initialized with `value=None` (the default) - the state will be determined in the frontend. Requesting the state with a JavaScript call or via "show" and "hide" events might be a solution for page functions, but not for the auto-index page which might show different states to different browsers.

Maybe we can still find a tradeoff without breaking changes and warnings:

1. Regular clients (page functions):
    - initial value `None`: browsers decide depending on screen width
    - one-shot timer (or on-connected hook) requests current state when page is loaded
    - setting the value or using show, hide and toggle methods work as expected
2. Auto-index client:
    - initial value `None`: browsers decide depending on screen width
    - the actual value won't be requested, it stays `None` on the server
    - setting the value or using show and hide methods work as expected
    - if the value is still `None` on the server, the toggle method will simply call the toggle method in frontend, toggling every drawer no matter its current state

I'm just not sure if the following example will behave correctly, i.e. if `d.show()` will correctly avoid the server state being overwritten by the client:
```py
@ui.page('/')
def page():
    d = ui.drawer()  # value=None, drawer state is undefined and will be requested when connected
    d.show()  # value=True, drawer should be visible
```
We might need to adjust the "show-if-above" prop depending on `value`: Whenever changing the value (e.g. when calling `show()`), the prop should be removed unless `value is None`.

What do you think? Might be worth a try.

falkoschindler:
Alternatively we could treat `value=None` the same in both cases:
- The client decides depending on the screen width.
- The toggle method will toggle the drawer without changing `value=None` on the server.
- Setting the value to `True` or `False` or using `show()` or `hide()` will remove the "show-if-above" prop.

falkoschindler (approved):
Since the `toggle()` method or simply resizing the browser window triggers a value-change event and the server value is updated, my plan to keep `value=None` didn't work as intended. But nevertheless we found a pretty good solution. To cite from the documentation:

> A value of `None` will automatically open or close the drawer depending on the current layout width (breakpoint: >=1024 px).
> On the auto-index page, the value will remain `None` until the drawer is opened, closed or toggled.
> On other pages, the value will be requested from the client when the websocket connection is established.

Thanks again for your contribution, @xaptronic! Let's merge into main!

PR #4347, opened 2025-02-13
Introduce QRating element

thetableman:
This PR introduces a Rating element based on Quasars QRating https://quasar.dev/vue-components/rating


falkoschindler (approved):
Thanks a lot for this pull request, @thetableman! Adding `ui.rating` is a great idea!

Among minor changes here and there, I added a `icon_half` parameter (because why not), and added support for `color` being a list of strings.

thetableman:
Thanks, @falkoschindler 

> I added a icon_half parameter (because why not),

Yeah, I was indecisive on this. It's a challenge to boil down a lot of Quasar prop options to the few that are likely needed the most, particularly when it's so easy to add those not included as parameters using `.props(...)`. It probably does make sense to include though.

PR #4358, opened 2025-02-16
 Introduce QSlideItem element

thetableman:
This PR introduces a Slide Item element based on Quasar's QSlideItem https://quasar.dev/vue-components/slide-item

@falkoschindler and @rodja I would be keen to hear your feedback on how I handled the `LeftSlide`, `RightSlide`, etc relationship with `SlideItem`, I feel there may be a better way to add the association of their slot to the parent element.

I'm also interested to hear your feedback on whether an `on_change` callback is really needed if each slot already has an `on_slide` callback.

I was undecided whether creating `LeftSlide`, `RightSlide`, etc classes were truly necessary, a user could just use `SlideSide` with the `side` parameter, but noted the style of `Drawer`, `LeftDrawer`, `RightDrawer` so aligned to this.

---

Feature request: #4282

### DIFF
--- a/nicegui/elements/slide_item.py
+++ b/nicegui/elements/slide_item.py
@@ -5,6 +5,7 @@ from typing import Literal, Optional
 from typing_extensions import Self
 
 from ..events import ClickEventArguments, GenericEventArguments, Handler, handle_event
+from ..slot import Slot
 from .mixins.disableable_element import DisableableElement
 
 SlideSides = Literal['left', 'right', 'top', 'bottom']
@@ -33,7 +34,7 @@ class SlideItem(DisableableElement):
               side: SlideSides, *,
               on_slide: Optional[Handler[GenericEventArguments]] = None,
               color: Optional[str] = 'primary',
-              ) -> Self:
+              ) -> Slot:
         """Slide
 
         This method adds a slide action to a specified side of the `SlideItem`
--- a/nicegui/elements/slide_item.py
+++ b/nicegui/elements/slide_item.py
@@ -20,18 +20,18 @@ class SlideItem(DisableableElement):
 
         This element is based on Quasar's `QSlideItem <https://quasar.dev/vue-components/slide-item/>`_ component.
 
-        If the text parameter is provided, an item will be created with the given text.
+        If the ``text`` parameter is provided, a nested ``ui.item`` element will be created with the given text.
         If you want to customize how the text is displayed, you can place custom elements inside the slide item.
 
-        To fill slots for individual slide actions, use the ``action`` method with a side argument
-        ("left", "right", "top", or "bottom") or the ``left``, ``right``, ``top``, or ``bottom`` methods.
+        To fill slots for individual slide actions, use the ``left``, ``right``, ``top``, or ``bottom`` methods or
+        the ``action`` method with a side argument ("left", "right", "top", or "bottom").
 
         Once a slide action has occurred, the slide item can be reset back to its initial state using the ``reset`` method.
 
         *Added in version 2.12.0*
 
         :param text: text to be displayed (default: "")
-        :param on_change: callback which is invoked when any slide action is activated
+        :param on_slide: callback which is invoked when any slide action is activated
         """
         super().__init__(tag='q-slide-item')
 
--- a/nicegui/elements/slide_item.py
+++ b/nicegui/elements/slide_item.py
@@ -1,46 +1,59 @@
 from __future__ import annotations
 
-from typing import List, Literal, Optional
+from typing import Literal, Optional
 
 from typing_extensions import Self
 
-from ..events import ClickEventArguments, GenericEventArguments, Handler, handle_event
+from ..events import Handler, SlideEventArguments, handle_event
 from ..slot import Slot
+from .item import Item
+from .label import Label
 from .mixins.disableable_element import DisableableElement
 
-SlideSides = Literal['left', 'right', 'top', 'bottom']
+SlideSide = Literal['left', 'right', 'top', 'bottom']
 
 
 class SlideItem(DisableableElement):
 
-    def __init__(self, on_change: Optional[Handler[ClickEventArguments]] = None) -> None:
+    def __init__(self, text: str = '', *, on_slide: Optional[Handler[SlideEventArguments]] = None) -> None:
         """Slide Item
 
         This element is based on Quasar's `QSlideItem <https://quasar.dev/vue-components/slide-item/>`_ component.
 
-        To add slide actions to a specific side (`left`, `right`, `top`, `bottom`) use the `slide` method.
+        If the text parameter is provided, an item will be created with the given text.
+        If you want to customize how the text is displayed, you can place custom elements inside the slide item.
 
+        To fill slots for individual slide actions, use the ``action`` method with a side argument
+        ("left", "right", "top", or "bottom") or the ``left``, ``right``, ``top``, or ``bottom`` methods.
+
+        Once a slide action has occurred, the slide item can be reset back to its initial state using the ``reset`` method.
+
+        *Added in version 2.12.0*
+
+        :param text: text to be displayed (default: "")
         :param on_change: callback which is invoked when any slide action is activated
         """
         super().__init__(tag='q-slide-item')
 
-        self._active_slides: List[str] = []
+        if text:
+            with self:
+                Item(text)
 
-        if on_change:
-            self.on_change(on_change)
+        if on_slide:
+            self.on_slide(None, on_slide)
 
-    def slide(self,
-              side: SlideSides, *,
-              on_slide: Optional[Handler[GenericEventArguments]] = None,
-              color: Optional[str] = 'primary',
-              ) -> Slot:
-        """Slide
+    def action(self,
+               side: SlideSide,
+               text: str = '', *,
+               on_slide: Optional[Handler[SlideEventArguments]] = None,
+               color: Optional[str] = 'primary',
+               ) -> Slot:
+        """Add a slide action to a specified side.
 
-        This method adds a slide action to a specified side of the `SlideItem`
-
-        :param side: side of the Slide Item where the slide should be added (`left`, `right`, `top`, `bottom`)
+        :param side: side of the slide item where the slide should be added ("left", "right", "top", "bottom")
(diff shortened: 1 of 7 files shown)


falkoschindler:
Hi @thetableman, thanks for yet another awesome pull request!

Just some quick feedback (haven't looked at all the code yet):

- Instead of nesting `ui.item_section` as a direct child of `ui.slide_item`, I think the usage should be
    ```py
    with ui.list():
        ui.item('Item A')
        ui.item('Item B')
        with ui.slide_item():
            ui.item('Slide me!')
            with ui.left_slide(color='green'):
                ui.item_label('Left')
            with ui.right_slide(color='red'):
                ui.item_label('Right')
    ```
    Using `ui.item` inside `ui.slide_item()` yields a consistent layout with regular items.
- Regarding the API for individual slots: What about creating slots like [`ui.splitter`](https://nicegui.io/documentation/splitter) does with `before` and `after`? This way we only need to define a single UI element and the usage is pretty straight forward. Ok, we can't easily define colors for slots. But maybe we should simply add `left_color`, `right_color`, ... parameters to `ui.slide_item` and we're good. This might be more accessible and more aligned with QSlideItem.

falkoschindler:
Maybe we can make the predefined slots callable to assign color and event handler...
```py
with slide_item.left(color='green', on_slide=lambda: ui.notify('Left')):
    ui.item_label('Left')
with slide_item.right(color='red', on_slide=lambda: ui.notify('Right')):
    ui.item_label('Right')
```
Or `left` and `right` are methods that add these slots when called. So many possibilities... 😄

thetableman:
Thanks for reviewing @falkoschindler 

> Instead of nesting `ui.item_section` as a direct child of `ui.slide_item`, I think the usage should be

Are you suggesting here that a `ui.item_section` becomes the default slot when a `ui.slide_item` is created? If so, this was my original intention however navigating the slot stack to ensure that the nested `ui.left_slide` was in the right place was messy.

> Regarding the API for individual slots: What about creating slots like [ui.splitter](https://nicegui.io/documentation/splitter) does with before and after? This way we only need to define a single UI element and the usage is pretty straight forward. Ok, we can't easily define colors for slots. But maybe we should simply add left_color, right_color, ... parameters to ui.slide_item and we're good. This might be more accessible and more aligned with QSlideItem.

I did play with this idea too but ultimately found it restrictive. 
* Slide sides are optional - you should be able to have a 'slide left' action only if desired, adding all slide slots on creation means there exist until you remove. 
* Further to point one, you mention 'left' and 'right' but there are also 'top' and 'bottom'. These are likely to be used less often but exist as options all the same.
* Further to point two, adding color options to `ui.slide_item` would then need four color parameters, one for each side. I'm not a fan of parameters being available for potentially redundant slots.
* To avoid four `SlideSide` inherited classes then I would opt for just using `SlideSide` and specifying side.

> Maybe we can make the predefined slots callable to assign color and event handler...

I think I like this alternative approach best but wouldn't this example create two `ui.slide_item` elements, one with a left and one with a right, rather than a single `ui.slide_item` with both? It would reduce the number of elements, dependent specifically on a single element, appearing in `ui`, which I support.

Surely usage would need to be as follows... but then this is almost as before:
```
with ui.slide_item as slide_item:
    ui.item('Slide me')
    with slide_item.left(color='green', on_slide=lambda: ui.notify('Left')):
        ui.item_label('Left')
    with slide_item.right(color='green', on_slide=lambda: ui.notify('Right')):
        ui.item_label('Right')
```

> So many possibilities... 😄

Indeed! I had a lot of fun approaching this one. For such a simple element there are a lot ways to approach.

thetableman:
I've had a play with two different approaches to the `ui.slide_item.left` suggestion based on the `SlideSide` classes defined in the PR. I'm not sure which I prefer at this stage.

```python
class SlideItem(DisableableElement):
    def __init__(self,
                 on_change: Optional[Handler[ClickEventArguments]] = None):

        super().__init__(tag='q-slide-item')

        self._active_slides: list[str] = []

        if on_change:
            self.on_change(on_change)

        self._left_slide: LeftSlide = None
        

    class right_slide(SlideSide):
        def __init__(self,
                 on_slide: Optional[Handler[GenericEventArguments]] = None,
                 color: Optional[str] = 'primary',
                 ) -> None:
            super().__init__('right', color=color, on_slide=on_slide)

    @property
    def left_slide(self) -> LeftSlide:

        if self._left_slide:
            return self._left_slide
        
        self._left_slide = LeftSlide
        
        return self._left_slide

    def reset(self) -> None:
        """Reset the Slide Item to initial state"""
        self.run_method('reset')
```

```python
from nicegui import ui

with ui.list().props('dense separator'):
    with ui.slide_item() as slide_item:
        with ui.item_section().classes('h-20'):
            ui.item_label('Slide me Up or Down')
        with slide_item.left_slide(on_slide=lambda: ui.notify('Left')):
            ui.item_label('Left')
        with slide_item.right_slide(on_slide=lambda: ui.notify('Right')):
            ui.item_label('Right')

ui.button('Reset Slide Item', on_click=lambda: slide_item.reset())

ui.run(reload=False)
```

falkoschindler:
> Are you suggesting here that a `ui.item_section` becomes the default slot when a `ui.slide_item` is created?

No, I was referring to your main demo. If `ui.item_section` is a direct child of `ui.slide_item`, the layout looks off. And according to Quasar's documentation, this is not the intended use of `QSlideItem`. Instead the `ui.item_section` should be wrapped inside another `ui.item`. Or we use NiceGUI's `text` argument for `ui.item` which does the rest for us:
```py
ui.item('Item A')
```

> Slide sides are optional - you should be able to have a 'slide left' action only if desired, adding all slide slots on creation means they exist until you remove.

Good point! We don't want them to be always created.

> Further to point one, you mention 'left' and 'right' but there are also 'top' and 'bottom'. These are likely to be used less often but exist as options all the same.

Yes, ignored "top" and "bottom". But they are a good example for optional slots that definitely need to stay optional.

> Further to point two, adding color options to `ui.slide_item` would then need four color parameters, one for each side. I'm not a fan of parameters being available for potentially redundant slots.

Agreed.

> but wouldn't this example create two `ui.slide_item` elements, one with a left and one with a right, rather than a single `ui.slide_item` with both?

Ah, I omitted the wrapping `ui.slide_item()`. Calling `slide_item.left()` and `slide_item.right()` would only add slots but no additional slide items.

So I think we can agree on a usage like this:
```py
with ui.slide_item as slide_item:
    ui.item('Slide me')
    with slide_item.left(color='green', on_slide=lambda: ui.notify('Left')):
        ui.item_label('Left')
    with slide_item.right(color='green', on_slide=lambda: ui.notify('Right')):
        ui.item_label('Right')
```

> I've had a play with two different approaches to the `ui.slide_item.left` suggestion based on the `SlideSide` classes defined in the PR. I'm not sure which I prefer at this stage.

I think we can simplify it by defining `left` and `right` as methods. They can directly refer to `self` and don't need to find their parent via `_get_slide_parent`. Actually, I think we can remove the `SlideSide` class as well as their subclasses `LeftSlide`, `RightSlide` etc.

> Indeed! I had a lot of fun approaching this one. For such a simple element there are a lot ways to approach.

Oh yes, I love tinkering with such API decisions as well. Often you end up with _nice_ solutions you haven't thought of in the beginning. And ultimately those solutions make a great library. 🙂

thetableman:
> No, I was referring to your main demo. If `ui.item_section` is a direct child of `ui.slide_item`, the layout looks off. And according to Quasar's documentation, this is not the intended use of `QSlideItem`. Instead the `ui.item_section` should be wrapped inside another `ui.item`

Ah, yes, you're right!

How about this approach with methods then? We can add `left(...)`, `right(...)`, etc methods if needed but as a working concept:

```python
class SlideItem(DisableableElement):
    def __init__(self,
                 on_change: Optional[Handler[ClickEventArguments]] = None):

        super().__init__(tag='q-slide-item')

        self._active_slides: list[str] = []

        if on_change:
            self.on_change(on_change)

    def slide(self,
              side: SlideSides, *,
              on_slide: Optional[Handler[GenericEventArguments]] = None,
              color: Optional[str] = 'primary',
              ) -> Self:
        
        if color:
            self._props[f'{side}-color'] = color

        if on_slide:
            self.on_slide(side, on_slide)

        if side not in self._active_slides:
            self._active_slides.append(side)

        return self.add_slot(side)

        
    def on_change(self, callback: Handler[ClickEventArguments]) -> Self:
        """Add a callback to be invoked when the Slide Item is changed."""
        self.on('action', lambda _: handle_event(callback, ClickEventArguments(sender=self, client=self.client)))
        return self
    
    def on_slide(self, side: SlideSides, callback: Handler[GenericEventArguments]) -> Self:
        """Add a callback to be invoked when a Slide Side is activated."""
        self.on(side, lambda _: handle_event(callback, GenericEventArguments(sender=self, client=self.client, args=side)))
        return self

    def reset(self) -> None:
        """Reset the Slide Item to initial state"""
        self.run_method('reset')

    @property
    def active_slides(self) -> list:
        """Returns a list of active Slide Sides"""
        return self._active_slides
```

```python
from nicegui import ui

with ui.list().props('dense separator'):
    with ui.slide_item() as slide_item:
        ui.item('Slide Me Left or Right').classes('h-20')

        with slide_item.slide(side='left', color='positive', on_slide=lambda: ui.notify('Left')):
            ui.label('Left')
        
        with slide_item.slide(side='right', color='negative', on_slide=lambda: ui.notify('Right')):
            ui.label('Right')

ui.button('Reset Slide Item', on_click=lambda: slide_item.reset())

ui.run()
```

falkoschindler:
@thetableman Looks great! 👍🏻

thetableman:
Done!

I'll leave it to you to decide whether creating 'left', 'right', etc methods are necessary.

falkoschindler:
Nice! Quick question: What is the use case for the `active_slides` property? If in doubt, I'd like to remove it. If you really need this information in rare cases, there's still the `slots` dictionary.

thetableman:
You can remove `active_slides` if you don't see value. I just prefer to include additional information where relevant and the ability to check what sides are active in the element easily without navigating the `slots` dict may be beneficial. It's not a deal-breaker though, it's just an extra tool for user's should they need it.

thetableman:
Looks good.

PR #4360, opened 2025-02-17
Add Leaflet WMS support

platinops:
Currently only WMTS (Web Map Tile Service) Leaflet functionality is supported in NiceGUI (through `L.tileLayer`).

Leaflet.js also supports WMS (Web Map Service) though: https://leafletjs.com/examples/wms/wms.html

This PR implements Leaflet.js `L.tileLayer.wms` and adds an example to the documentation.

### DIFF
--- a/nicegui/elements/leaflet.js
+++ b/nicegui/elements/leaflet.js
@@ -127,11 +127,11 @@ export default {
   },
   methods: {
     add_layer(layer, id) {
-      if (layer.type == "tileLayer.wms") {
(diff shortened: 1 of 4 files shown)


falkoschindler (approved):
Thanks a lot for this pull request, @platinops!
I just added a version info for the new feature, replaced the comments with labels (comments are not rendered correctly within interactive demos) and used a more generic way to resolve the layer type in `add_layer()`.
Let's merge! 🚀

PR #4368, opened 2025-02-18
Add ruff quote-style

EmberLightVFX:
A super small PR that adds simple quotes settings for ruff to make sure we stay in the correct coding style.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,4 @@
 {
-  "autopep8.args": ["--max-line-length=120"],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
   "prettier.printWidth": 120,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,12 +97,15 @@ module = [
 ]
 ignore_missing_imports = true
 
+[tool.autopep8]
+max-line-length = 120
+in-place = true
+recursive = true
+experimental = true
+
 [tool.ruff]
 indent-width = 4
 line-length = 120
-
-[tool.ruff.format]
-quote-style = "single"
 
 [tool.ruff.lint]
 # See complete list: https://docs.astral.sh/ruff/rules/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,9 @@ ignore_missing_imports = true
 indent-width = 4
 line-length = 120
 
+[tool.ruff.format]
+quote-style = "single"
+
 [tool.ruff.lint]
 # See complete list: https://docs.astral.sh/ruff/rules/
 select = [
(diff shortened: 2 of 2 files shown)


codingpaula:
Hello @EmberLightVFX,

thanks for bringing this up here! We currently only use the linting function of ruff since the ruff formatter in our opinion doesn't work well with the way NiceGUI elements are written.
We usually write elements like this button for example:
```python
ui.button('Click me!', icon='left_click', on_click=self.on_click) \
    .bind_visibility_from(self.property, 'active') \
    .props('flat rounded dense').classes('m-0 p-2') \
    .tooltip('This is just a test')
```
If we now format this with ruff instead of autopep8, it becomes this:
```python
ui.button("Click me!", icon="left_click", on_click=self.on_click).bind_visibility_from(
    self.property, "active"
).props("flat rounded dense").classes("m-0 p-2").tooltip("This is just a test")
```
which is a lot harder to read and identify in one glance. Because of this, for the time being we want to stick to autopep8 as our formatter.

As far as I could find out from the ruff documentation the setting you suggest doesn't influence the `ruff check` command. So, I'd argue against having a `[tool.ruff.format]` section in our `pyproject.tom` since it would wrongly indicate that we use ruff for formatting.
But while at it, I also noticed that we don't have an autopep8 config section in our `pyproject.toml`, just some configuration in the vs code settings. Would you be fine with "rerouting" this PR to moving the autopep8 config into the `pyproject.toml` instead of adding the ruff format config?

EmberLightVFX:
I see. Yeah Ruff doesn't seem to be able to format the code as you described.
I don't know exactly how the autopep8 settings would look like in pyproject.toml so I think it's better for you to add it correctly. If you do it in this PR or if you make a new one is ok with me.

codingpaula:
That should do the trick :)

codingpaula (approved):

falkoschindler:
@codingpaula Now this PR seems to disable auto-formatting on save in VSCode. Take this line in label.py for example:
```py
class Label(TextElement ):
```
It should be converted to
```py
class Label(TextElement):
```
But this doesn't work when checking out this branch.

However, when running pre-commit, the line is fixed with this PR (and not on the main branch). It would be great if both applies: pre-commit and format on save.

EmberLightVFX:
Weird. For me it does convert it to with the latest PR (and with Ruff disabled just to be sure)
```
class Label(TextElement):
```

codingpaula:
@falkoschindler, I checked again and for my vscode you're right. autopep8 says that `--in-place and --recursive cannot be used with standard input`. Which seems to be caused by the difference between using autopep8 as a command line tool or with the vscode plugin. So removing both resolves the problem with vscode for me. I will add a commit to remove them. Can you both then check again if formatting still works?

However, I can't reproduce pre-commit fixing that whitespace. What pre-commit hook did fix it for you?

falkoschindler:
@codingpaula Looks like I was wrong. I can't reproduce pre-commit fixing this whitespace either.

falkoschindler (approved):
Now format-on-save seems to work again in VSCode. Thanks, @codingpaula!

PR #4371, opened 2025-02-19
Add Google OAuth2 example 

frankvp11:
Adds the Google OAuth2 example that I created [here](https://github.com/zauberzeug/nicegui/discussions/4361)
Unfortunately, I'm not very skilled with readme's, so I used some images and text where I could. I ask that you guys fix that (if you feel it's necessary).
Let me know what you guys think.


rodja:
@frankvp11 I finally found the time to spend some time on this PR. While your code worked I took the liberty to refactor and simplify it a bit. Please have a look and tell me what you think. Also it's interesting to compare it with the rework I've done to #3244, where I applied a lot of learnings from this PR.

rodja (changes requested):
It would be great if we can get rid of the guide_images and simplify the README.md. @frankvp11 is there not a document at Google which describes the steps? I think it's much better if we do not need to maintain our own documentation. For one-tap-auth I found such a description: https://github.com/zauberzeug/nicegui/blob/a366de0c5317186a63f20440ef3fa97ffda2cab2/examples/google_one_tap_auth/main.py#L11

Also the example should be added to https://github.com/zauberzeug/nicegui/blob/main/website/examples.py

frankvp11:
By the looks of it, that covers the exact same steps [(this link, that you provided)](https://developers.google.com/identity/gsi/web/guides/get-google-api-clientid#get_your_google_api_client_id). I think it wouldn't be a bad idea to just remove the README and images all together and add the comment to the `main.py` itself. 
What do you think?

frankvp11:
Were there any more updates required from my end?

rodja (approved):

Issue #4373, opened 2025-02-20
ui.notification progress bar not synced with timeout


pedropg898:
### Description

I'm trying to provide a timeout to ui.notification and have a synced dismiss with the progress bar

With the following code snippet, I expected the progress bar to be synced with the timeout I provide:
```python
a = ui.notification(message='Line has been updated.', icon='check_circle_outline', 
                    classes='success-notification', timeout=5, progress=True)
await asyncio.sleep(5)
a.dismiss()
```

But when I run this code snippet, the progress bar is out of sync with the timeout and on two notifications, sometimes the second one resets the progress bar on the first notification. The dismiss itself seems to be working on time.
**Example 1:** timeout of 5 seconds - click one time to show notification - the notification closes in 5 seconds, but the progress bar doesn't get to the end;
**Example 2:** timeout of 5 seconds - click once to show notification1 - wait a second - click again to show notification2 - this sometimes resets the progress bar on notification1 - notification1 closes in 5 seconds, but the progress bar doesn't get to the end - notification2 closes in 5 seconds, , but the progress bar doesn't get to the end.

falkoschindler:
Hi @pedropg898,

It looks like Quasar's notification progress bars are always a bit behind:
https://quasar.dev/quasar-plugins/notify#example--timeout-progress
When clicking the button "Show timeout progress", the first notification should dismiss after 5000ms, but stays a bit longer. Maybe the animation duration is added.
Why do you need to call `a.dismiss()` anyway? The notification will dismiss automatically.

falkoschindler:
Ok, so the main problem is that a notification with `progress=True` behaves weirdly:

```py
ui.button('Notify', on_click=lambda: ui.notification(message='Hi!', progress=True))
```

<img width="310" alt="Image" src="https://github.com/user-attachments/assets/4a5bd2e9-09fb-45d3-8823-ffc970fcbfbd" />

- When showing another notification, the previous progress bar resets to 100%.
- Only when _all_ progress bars have reached 0%, the notifications disappear.

A plain Quasar example, on the other hand, behaves correctly:
```html
<html>
  <head>
    <link href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900|Material+Icons" rel="stylesheet" type="text/css" />
    <link href="https://cdn.jsdelivr.net/npm/quasar@2.17.3/dist/quasar.prod.css" rel="stylesheet" type="text/css" />
  </head>
  <body>
[...]

PR #4379, opened 2025-02-22
improve update of notifications

falkoschindler:
As noticed in #4373, notifications received an update when another notification was created. This update called `this.notification()`, causing progress bars to reset. This PR solves this problem by explicitly calling an `update_notification()` method on UI updates, but preventing the accidental updates caused by other notifications.

Can be tested with a simple `ui.notification`:
```py
ui.button('Notify', on_click=lambda: ui.notification('Hi!'))
```
Clicking the button again and again will prevent any notification from disappearing after the default timeout of 5 seconds.

(By the way: I removed an obsolete "data-id" attribute. This hasn't been in use since #2983.)

### DIFF
--- a/nicegui/elements/notification.js
+++ b/nicegui/elements/notification.js
@@ -4,10 +4,10 @@ export default {
   mounted() {
     this.notification = Quasar.Notify.create(this.convertedOptions);
   },
-  updated() {
-    this.notification(this.convertedOptions);
-  },
   methods: {
+    update_notification() {
+      this.notification(this.convertedOptions);
+    },
     dismiss() {
       this.notification();
     },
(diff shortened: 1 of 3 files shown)


rodja:
> Clicking the button again and again will prevent any notification from disappearing after the default timeout of 5 seconds.

Strange. I can not reproduce the described issue on the main branch:

Feb-23-2025 14-49-46

rodja:
Ah, seems that the issue does not happen on auto-index pages. I could reproduce it with a page builder. Strangely the test still fails on main -- even if it does not use a page builder.

falkoschindler:
> Ah, seems that the issue does not happen on auto-index pages.

Yes it does. Another example:

```py
ui.timer(1, lambda: ui.notification('Hi!', timeout=3))
```

There shouldn't be more than 3 or 4 notifications on the screen at the same time. Yet they don't disappear.

PR #4382, opened 2025-02-24
Add function to update browser bar URL

platinops:
Sometimes when you change the state of you application, you also want to have this reflected in the browser URL bar. That way, users can bookmark/share that URL and return to the state where they left off. I could not immediately find this functionality in NiceGUI out of the box.

Simple use case:

``` python
@ui.page('/select/{id}')
def page(id: int):
    select1 = ui.select([1, 2, 3], value=id).on_value_change(
        lambda e: ui.navigate.set_browser_url(f'/select/{e.value}')
    )
    select2 = ui.select({1: 'One', 2: 'Two', 3: 'Three'}).bind_value(select1, 'value')
    ui.link('Link to selected option', f'/select/{id}')
```


trivedihoney:
Hello @platinops,
If we are to integrate the `pushState` functionality, it would be great to integrate [`replaceState`](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) as well!

Usage can look like, 
`ui.navigate.set_browser_url(f'/select/{e.value}', replace=True)`

Or two separate functions can be also created like `ui.navigate.push` & `ui.navigate.replace` for more clarity.

Thoughts?

rodja:
Great idea @platinops. Maybe we can simplify the name to make it more consistent with `navigate.to`, `navigate.backward`, ...? What do you think about `ui.navigate.update('/path')`? And for @trivedihoney's suggestion we either introduce `ui.navigate.replace('/path')` or introduce a parameter `ui.navigate.update('/path', with_history=False)`.

Having this build in would also be beneficial for #2811 where it is currently done by building js commands:

https://github.com/Alyxion/nicegui/blob/1e5b3267ab938da9d802be1fe9be2e36fe75e5e3/nicegui/single_page_router.py#L226

platinops:
@trivedihoney: Fair point, sounds logical to support replaceState too.

@rodja: OK for me to use `ui.navigate.update`. It's a bit more cryptic, but looks more compact/clean.

Also slightly related, getting the browser's geolocation like in [this example](https://nicegui.io/documentation/run_javascript#run_async_javascript) could also be a builtin NiceGUI function? Would you also include it under ui.navigate?

Not sure if it makes sense, but would it reasonable to consider the browser url and geolocation as `ui.Element`, similar to [dcc.Location](https://dash.plotly.com/dash-core-components/location) in Plotly Dash? This would allow binding other elements' values to the URL path and/or query parameters.

rodja:
Geolocation should not be part of `ui.navigation`. And binding to the url feels complicated. Better to leave it as functions.

platinops:
Okey, thanks for the feedback. I updated the PR accordingly.

falkoschindler (commented):
Thanks for the pull request, @platinops!

I'm still not 100% happy with the naming though:

1. I feel like `update_url` or even `set_url` would better describe _what_ is updated. And because we're not really updating anything but plainly _setting_ the URL, "set" might be the better verb.
2. The parameter `with_history` seems a bit misleading, because the URL is added to the history in both cases. The question is, is it _added_ or _replaced_? Better names might be `add_to_history` or `push_to_history`. Inverting the logic it could also be called `replace_current`, `replace_in_history` or simply `replace`.

My favorite is probably
```py
ui.navigate.set_url(url: str, *, replace_in_history: bool = False)
```

What do you think, @platinops and @rodja?

platinops:
> I feel like `update_url` or even `set_url` would better describe _what_ is updated. And because we're not really updating anything but plainly _setting_ the URL, "set" might be the better verb.

Agreed; that's why I used `set` in the original PR :)

I'll await Rodja's reply before pushing any further commits though.

rodja:
I'm not really happy with `update` `update_url` and `set_url`. Maybe we could introduce `ui.history.push` or `ui.navigate.history.push` to better convey meaning and be more in line with browser API?

xaptronic:
If there is push/replace would it make sense to consider how one might reactively handle pop state? (e.g.; user using browser nav buttons?)

rodja:
@xaptronic sure. Please create a separate Feature Request for this, so we can discuss the details.

PR #4388, opened 2025-02-26
Pass referrer_path to /login as a query parameter

tonyallan:
Don't use app.storage.user['referrer_path'] because it can get confused if multiple tabs are used by the user. Instead pass the referrer_path as the URL query parameter "r" to /login.


falkoschindler (approved):
Thanks for this pull request, @tonyallan!

I just renamed the parameter to the more verbose `redirect_to` for this example. Of course the user can reduce it to something minimal like `r` in a productive environment.

Issue #4395, opened 2025-02-27
Timing issue when loading leaflet plugins via additional resources


anthony-arnold:
### Description

Attempting to load a leaflet plugin can result in failure. I'm loading a locally-written plugin like so:

```python
ui.leaflet(additional_resources=['/static/js/my-leaflet-plugin.js'])
```

where the `static` directory is served using `app.add_static_files`. 

I'm expecting the leaflet library to load in the browser, followed by the `my-leaflet-plugin.js` code. However due to the leaflet resource [loading code](https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/leaflet.js#L20), both leaflet and the plugins are loaded simultaneously. This can (and does consistently, in my case) result in the plugin being loaded by the browser before leaflet. Consequently, `window.L` is undefined at the point of loading my plugin.

I'm currently working around this by waiting for the leaflet map to load, then injecting my plugin script manually:

```python
m = ui.leaflet()
async def post_setup():
  await ui.run_javascript("""
const p = new Promise((resolve, reject) => {
  const js = document.createElement('script');
  js.src = '/static/js/my-leaflet-plugin.js';
  document.head.appendChild(js)
  js.onload = resolve
  js.onerror = reject
});
return p;""")

m.on('init', post_setup)
```

I would suggest modifying the leaflet element javascript by waiting for leaflet to be loaded, then loading the additional resources:

```javascript
await Promise.all([
      loadResource(window.path_prefix + `${this.resource_path}/leaflet/leaflet.css`),
      loadResource(window.path_prefix + `${this.resource_path}/leaflet/leaflet.js`)
    ]);
await Promise.all(this.additional_resources.map((resource) => loadResource(resource)));
```

falkoschindler:
Thanks for reporting this issue, @anthony-arnold!
Loading custom resources _after_ loading the Leaflet library itself (and maybe even leaflet-draw) makes sense.
Would you like to create a pull request? 🙂

PR #4401, opened 2025-02-28
Wait for leaflet to load before loading additional resources.

anthony-arnold:
In reference to [4395](https://github.com/zauberzeug/nicegui/issues/4395), this fix lets leaflet load before loading additional resources.


falkoschindler (approved):
Thanks, @anthony-arnold! 👍🏻

Issue #4406, opened 2025-03-01
add_static_file should not raise ValueError


ed2050:
### Description

app.add_static_file raised ValueError when file doesn't exist:

```py
>>> url = app.add_static_file (local_file = "foo.mp4")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Python/3.8/site-packages/nicegui/app.py", line 110, in add_static_file
    raise ValueError(f'File not found: {file}')
ValueError: File not found: /Users/foobar/media/foo.mp4
```

This is incorrect behavior.  Many times you want to create a url to a non-existent file, for instance:
- when another process is creating the file, nicegui may want to publish a link immediately.  by the time user clicks the link, other process will be finished and file will exist.
- nicegui may be processing things out-of-order for responsiveness, ie first update page with link to file, then create the file in a longer operation.
- nicegui may want to show links to future files.  eg your nicegui app publishes a page on Monday with links to status report for each day this week.  Status report Files will be uploaded at end of each day by another user or process.

That's just off the top of my head.  Just like python doesn't verify filesystem paths until the user tries to access it, nicegui shouldn't verify file urls until the url is requested.

# Expected Behavior

add_static_file should always return a url for the given path, whether or not path exists.

rodja:
Maybe we can introduce a `strict=True` parameter -- similar to the one in [pathlib](https://docs.python.org/3/library/pathlib.html#:~:text=If%20a%20path%20doesn%E2%80%99t%20exist%20or%20a%20symlink%20loop%20is%20encountered%2C%20and%20strict%20is%20True%2C%20OSError%20is%20raised.%20If%20strict%20is%20False%2C%20the%20path%20is%20resolved%20as%20far%20as%20possible%20and%20any%20remainder%20is%20appended%20without%20checking%20whether%20it%20exists.).

ed2050:
Seems like a good approach.

falkoschindler:
@ed2050 Would you like to create a pull request adding a parameter `strict: bool = True`?

ed2050:
See [PR 4411](https://github.com/zauberzeug/nicegui/pull/4411).

## Notes

- Added flag strict to add_static_file params that determines whether to check if file exists or not.
- default value should be FALSE because:
1) python stdlibs like [pathlib](https://docs.python.org/3/library/pathlib.html) are default false, because paths are not checked for existence until accessed
[...]

PR #4411, opened 2025-03-03
flag for file existence checks on add_static_file

ed2050:
- Added flag `strict` to `add_static_file` params that determines whether to check if file exists or not.
- default value should be FALSE because:
1) python stdlibs like [pathlib](https://docs.python.org/3/library/pathlib.html) are default false, because paths are not checked for existence until accessed
2) existing code is very unlikely to use `add_static_file` raising an error to determine whether a file exists or not
3) raising error on file existence is undocumented behavior, making it even less likely for existing code to rely on
- changed error raised from general ValueError to more appropriate FileNotFoundError as used by stdlib

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -233,7 +233,7 @@ class App(FastAPI):
                        local_file: Union[str, Path],
                        url_path: Optional[str] = None,
                        single_use: bool = False,
-                       ) -> str:
+                       strict: bool = False) -> str:
         """Add a single media file.
 
         Allows a local file to be streamed.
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Thanks for this pull request, @ed2050!

As discussed in #4406, we'll merge it with default `strict=True`, raising `ValueError`s and adding a deprecation warning to change it to `FileNotFoundError` in NiceGUI 3.0.

Regarding the default value, there are plenty of arguments for `True` as well as for `False`. I'd like to summarize a discussion with Claude.ai with one sentence:

> Starlette likely chose check_dir=True as default for a very compelling reason: it aligns with the principle of "fail fast" in web frameworks.

Of course there are scenarios where this isn't what you want, so it's good to have options now.

PR #4429, opened 2025-03-06
Profile imports

falkoschindler:
Inspired by discussion #4353, this draft pull request serves as a playground to profile the import of NiceGUI's `ui` module (and maybe others).

Two insights so far:

1. Often we can avoid importing FastAPI by importing things like `Request` or `Response` from Starlette instead. At the moment, however, this only shifts the required import time to app.py where we want to subclass the FastAPI app.
2. When importing `ui` on my machine, 70% of the time is spent importing Matplotlib. And we already have an environment variable to disable it. So it should be straight forward to save some time by setting this variable when creating a subprocess.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ venv
 .nicegui/
 *.sqlite*
 .DS_Store
+*.lprof
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -7,8 +7,9 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, Awaitable, Callable, Iterator, List, Optional, Union
 
-from fastapi import FastAPI, HTTPException, Request, Response
-from fastapi.responses import FileResponse
+from fastapi import FastAPI, HTTPException
+from starlette.requests import Request
+from starlette.responses import FileResponse, Response
 
 from .. import background_tasks, helpers
 from ..client import Client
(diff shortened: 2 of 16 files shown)


falkoschindler:
3. We can trick the import system into thinking that a specific module has already been imported:
    ```py
    sys.modules['nicegui'] = 0
    ```
    or
    ```py
    sys.modules['fastapi'] = 0
    ```
    Of course, accessing `nicegui.ui` or `fastapi.FastAPI` would fail.

falkoschindler:
Continuing on insight 3, we can inject a dummy class that loads quickly and doesn't complain about anything at runtime:
```py
class WhateverModule(ModuleType):

    def __getattr__(self, name):
        return Whatever()

class Whatever:

    def __init__(self, *args, **kwargs):
        pass

    def __getattr__(self, name):
        return Whatever()

    def __call__(self, *args, **kwargs):
        return Whatever()

    def __mro_entries__(self, *args, **kwargs):
        return (Whatever, )

    def __add__(self, other):
        return Whatever()

    def __enter__(self):
        return self

    def __exit__(self, *args, **kwargs):
        pass

sys.modules['fastapi'] = WhateverModule('fastapi')
```
This reduces the overall time for importing nicegui (without Matplotlib) from 0.77s down to 0.43s.

evnchn:
https://pypi.org/project/lazy-imports/?

falkoschindler:
> https://pypi.org/project/lazy-imports/?

Interesting. But looks like a nightmare for type checkers.

Anyway, a proof of concept to address the slow initialization of subprocesses (see #4353) would be great!

falkoschindler:

PR #4444, opened 2025-03-08
feat: add `timeout` method for `ui.notification`

weinibuliu:
This PR is based on #4437 , which explores a way to make notifications dismiss after a certain period of time.

This PR provides a method called `timeout` that allows people to directly override the value of timeout (instead of manually modifying the value of the props)

I did a simple test.It worked well.
```python
# notification.py (This PR did)
class Notification(Element, component='notification.js'):
    ...
    @property
    def timeout(self) -> Optional[float]:
        """The timout value of the notification."""
        return self._props['options']['timeout']

    @timeout.setter
    def timeout(self,value: Optional[float]) -> None:
        self._props["options"]["timeout"] = (value or 0) *1000
        self.update()

# test.py
import asyncio

from nicegui import ui

async def test_set_timeout():
    n = ui.notification("Test Notification",timeout=None)
    n.on_dismiss(lambda:print("Notification Dismiss."))
    await asyncio.sleep(3)
    n.message="Test Done"
    n.timeout = 3 # This is a new method

ui.button("Test").on_click(test_set_timeout)

ui.run()
```
That's all.

If you have any suggestions for changes to this PR, please let me know. Thanks a lot.

### DIFF
--- a/nicegui/elements/notification.py
+++ b/nicegui/elements/notification.py
@@ -179,7 +179,10 @@ class Notification(Element, component='notification.js'):
 
     @property
     def timeout(self) -> float:
-        """Timeout of the notification in seconds."""
+        """Timeout of the notification in seconds.
+
+        *Added in version 2.13.0*
+        """
         return self._props['options']['timeout'] / 1000
 
     @timeout.setter
--- a/nicegui/elements/notification.py
+++ b/nicegui/elements/notification.py
@@ -199,6 +199,11 @@ class Notification(Element, component='notification.js'):
     def set_visibility(self, visible: bool) -> None:
         raise NotImplementedError('Use `dismiss()` to remove the notification. See #3670 for more information.')
 
+    def set_timeout(self,timeout: Optional[float]) -> None:
+        """Set the notification's timeout."""
+        self.props["options"]["timeout"] = (timeout or 0) *1000
+        self.update()
+
     def update(self) -> None:
         super().update()
         self.run_method('update_notification')
--- a/nicegui/elements/notification.py
+++ b/nicegui/elements/notification.py
@@ -178,7 +178,7 @@ class Notification(Element, component='notification.js'):
         self.update()
 
     @property
-    def timeout(self) -> Optional[float]:
+    def timeout(self) -> float:
         """Timeout of the notification in seconds."""
         return self._props['options']['timeout'] / 1000
 
--- a/nicegui/elements/notification.py
+++ b/nicegui/elements/notification.py
@@ -1,5 +1,6 @@
 import asyncio
 from typing import Any, Dict, Literal, Optional, Union
+from enum import Enum, auto
 
 from typing_extensions import Self
 
@@ -27,6 +28,8 @@ NotificationType = Optional[Literal[
     'ongoing',
 ]]
 
+class ChangeType(Enum):
+        NoneChange = auto()
 
 class Notification(Element, component='notification.js'):
 
@@ -178,6 +181,16 @@ class Notification(Element, component='notification.js'):
         self.update()
 
     @property
+    def timeout(self) -> Optional[float]:
+        """Timout of the notification."""
+        return self._props['options']['timeout']
+
+    @timeout.setter
+    def timeout(self,value: Optional[float]) -> None:
+        self._props["options"]["timeout"] = (value or 0) *1000
+        self.update()
+
+    @property
     def close_button(self) -> Union[bool, str]:
         """Whether the notification has a close button."""
         return self._props['options']['closeBtn']
@@ -198,11 +211,6 @@ class Notification(Element, component='notification.js'):
 
     def set_visibility(self, visible: bool) -> None:
         raise NotImplementedError('Use `dismiss()` to remove the notification. See #3670 for more information.')
-
-    def set_timeout(self,timeout: Optional[float]) -> None:
-        """Set the notification's timeout."""
-        self.props["options"]["timeout"] = (timeout or 0) *1000
-        self.update()
 
     def update(self) -> None:
         super().update()
--- a/nicegui/elements/notification.py
+++ b/nicegui/elements/notification.py
@@ -1,6 +1,5 @@
 import asyncio
 from typing import Any, Dict, Literal, Optional, Union
-from enum import Enum, auto
 
 from typing_extensions import Self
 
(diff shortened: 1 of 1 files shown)


weinibuliu:
Hi, @falkoschindler 
I'm thinking of a method called `reset` that will allow developers to change almost any value of a notification by calling a single line of code. It will work like below, please note that this is just a demo and it is not included in this PR.

```python
# notification.py

class ChangeType(Enum):
        NoneChange = auto()
...
    def reset(self,
              message: Any = ChangeType.NoneChange,
              position: Union[NotificationPosition, ChangeType] = ChangeType.NoneChange,
              type: Union[NotificationType, ChangeType] = ChangeType.NoneChange,
              timeout: Union[float, int, ChangeType] = ChangeType.NoneChange,) -> None: # For a demo,I only use some params.
        
        self.props["options"]["message"] = str(message) if message != ChangeType.NoneChange  else self.props["options"]["message"]
        self.props["options"]["position"] = position if position != ChangeType.NoneChange else self.props["options"]["position"]
        self.props["options"]["type"] = type if type != ChangeType.NoneChange else self.props["options"]["type"]
        self.props["options"]["timeout"] = (timeout or 0)*1000 if timeout != ChangeType.NoneChange else self.props["options"]["timeout"]
        self.update()
```

Developers only need to call it like this.
```python
import asyncio

from nicegui import ui

async def test_reset():
    n = ui.notification("Test Notification",timeout=None,position="bottom-right",type="info")
    n.on_dismiss(lambda:print("Notification Dismiss."))
    await asyncio.sleep(3) # Do something
    n.reset("Test Done",timeout=5,type="positive") # This is a new method

ui.button("Test").on_click(test_reset)

ui.run()
```

```python
n.reset("Test Done",timeout=5,type="positive")
# is approximately equivalent to the following code.
n.message = "Test Done"
n.timeout = 5
n.type = "positie"
```
I think this will be more convenient, can we expand on this? If you have any suggestions for improvements to this idea, please let me know, thank you very much.

falkoschindler (approved):
Thanks for the pull request, @weinibuliu!
I cleaned it up a bit and now it's ready to merge.

Regarding a reset method: I'm a bit hesitant because we'll basically need to duplicate the signature and docstring of the initializer. But I guess it would be a useful feature nonetheless. Would you like to create a separate pull request?

weinibuliu:
I have created a new discussion on it. #4460

evnchn:
Wait I think I finally figured it out. 

By `reset` I think @weinibuliu you actually mean to "set it again", or "re-set" if you will 😅

When I think about "reset" I think about "factory reset", so I am so confused why I need pass arguments. 

I would propose that this new method be called `configure` instead. 

About the docstring and function signature duplication issue, I have thought about it for a bit. 

* Parameters are set to sensible defaults in __init__, but to `object()` in `configure`
* `__init__` invokes `configure`, where `configure` is the method that actually writes the value to the element
  * Notably, `configure` never gets passed `object()` if it is `__init__` calling it, so it is the same as old behaviour
*  If `configure` sees values, it writes. If it sees an `object()` instead (the default parmater of this function), it does not act and maintains the unchanged value. 
* `configure` can just point the library users to look at the docs of `__init__` and say "any values you do not pass in, they will be unchanged and inherit the value of the original element before `configure` call"

@falkoschindler what do you think in terms of the codebase simplicity and DRY principle?

weinibuliu:
> Wait I think I finally figured it out.
> 
> By `reset` I think @weinibuliu you actually mean to "set it again", or "re-set" if you will 😅
> 
> When I think about "reset" I think about "factory reset", so I am so confused why I need pass arguments.
> 
> I would propose that this new method be called `configure` instead.
> 
> About the docstring and function signature duplication issue, I have thought about it for a bit.
> 
> * Parameters are set to sensible defaults in **init**, but to `object()` in `configure`
> * `__init__` invokes `configure`, where `configure` is the method that actually writes the value to the element
>   
>   * Notably, `configure` never gets passed `object()` if it is `__init__` calling it, so it is the same as old behaviour
> * If `configure` sees values, it writes. If it sees an `object()` instead (the default parmater of this function), it does not act and maintains the unchanged value.
> * `configure` can just point the library users to look at the docs of `__init__` and say "any values you do not pass in, they will be unchanged and inherit the value of the original element before `configure` call"
> 
> @falkoschindler what do you think in terms of the codebase simplicity and DRY principle?
Thanks your reply.

I'm sorry this caused confusion. I didn't think much about method naming, but I'm glad you understood what I meant.

I'm still thinking whether it's necessary to add this method to reset some parameters, because users can actually implement it themselves (in a very simple way)

weinibuliu:
> Wait I think I finally figured it out.
> 
> By `reset` I think @weinibuliu you actually mean to "set it again", or "re-set" if you will 😅
> 
> When I think about "reset" I think about "factory reset", so I am so confused why I need pass arguments.
> 
> I would propose that this new method be called `configure` instead.
> 
> About the docstring and function signature duplication issue, I have thought about it for a bit.
> 
> * Parameters are set to sensible defaults in **init**, but to `object()` in `configure`
> * `__init__` invokes `configure`, where `configure` is the method that actually writes the value to the element
>   
>   * Notably, `configure` never gets passed `object()` if it is `__init__` calling it, so it is the same as old behaviour
> * If `configure` sees values, it writes. If it sees an `object()` instead (the default parmater of this function), it does not act and maintains the unchanged value.
> * `configure` can just point the library users to look at the docs of `__init__` and say "any values you do not pass in, they will be unchanged and inherit the value of the original element before `configure` call"
> 
> @falkoschindler what do you think in terms of the codebase simplicity and DRY principle?
Thanks your reply.

I'm sorry this caused confusion. I didn't think much about method naming, but I'm glad you understood what I meant.

I'm still thinking whether it's necessary to add this method to reset some parameters, because users can actually implement it themselves (in a very simple way)

evnchn:
I think it may be beneficial to re-title the PRs and discussions to avoid any confusion moving forward. 

I think the necessity of adding the method depends on personal preference. I'm neutral to it.

falkoschindler:
@evnchn Shouldn't we discuss the `reset` method on #4460 or #4464? This PR was about `timeout` and has been merged and closed.

evnchn:
> @evnchn Shouldn't we discuss the `reset` method on #4460 or #4464? This PR was about `timeout` and has been merged and closed.

For the context of understanding the `reset` naming and why I suggest `configure`, I persoanlly think that it makes more sense to read it from here. 

But we should go back to #4460 as @weinibuliu also confirmed that the new naming is good

PR #4449, opened 2025-03-09
Better support select element with multiple values and dictionaries in user simulation

Daemo00:
First things first: Thanks for your wonderful project!

I have implemented a small application and have just begun adding tests.
My very first test is failing with error:
```
[...]
.tox/py313/lib/python3.13/site-packages/nicegui/testing/user.py:110: in should_see
    if self.notify.contains(target) or self._gather_elements(target, kind, marker, content):
.tox/py313/lib/python3.13/site-packages/nicegui/testing/user.py:211: in _gather_elements
    elements = set(ElementFilter(marker=target)).union(ElementFilter(content=target))
_ _ _ _ _ _ _ _ _ _ _


self = <nicegui.element_filter.ElementFilter object at 0x7fb35113c2d0>

    def __iter__(self) -> Iterator[T]:
        for element in self._scope.descendants():
            if self._kind and not isinstance(element, self._kind):
                continue
            if self._exclude_kinds and isinstance(element, tuple(self._exclude_kinds)):
                continue
    
            if any(marker not in element._markers for marker in self._markers):
                continue
            if any(marker in element._markers for marker in self._exclude_markers):
                continue
    
            if self._contents or self._exclude_content:
                element_contents = [content for content in (
                    element.props.get('text'),
                    element.props.get('label'),
                    element.props.get('icon'),
                    element.props.get('placeholder'),
                    element.props.get('value'),
                    element.props.get('error-message'),
                    element.text if isinstance(element, TextElement) else None,
                    element.content if isinstance(element, ContentElement) else None,
                    element.source if isinstance(element, SourceElement) else None,
                ) if content]
                if isinstance(element, Notification):
                    element_contents.append(element.message)
                if isinstance(element, (Select, Radio, Toggle)):
                    options = {option['value']: option['label'] for option in element.props.get('options', [])}
>                   element_contents.append(options.get(element.value, ''))
E                   TypeError: unhashable type: 'list'

.tox/py313/lib/python3.13/site-packages/nicegui/element_filter.py:123: TypeError
```

I investigated the issue and managed to fix it and create a dedicated test case too, let me know what you think!

### DIFF
--- a/nicegui/element_filter.py
+++ b/nicegui/element_filter.py
@@ -120,8 +120,14 @@ class ElementFilter(Generic[T]):
                 if isinstance(element, ChoiceElement):
                     is_options_dict = isinstance(element.options, dict)
                     options = element.options.values() if is_options_dict else element.options
-                    element_contents.append(
-                        element.options.get(element.value, '') if is_options_dict else element.value)
+                    if isinstance(element, Select) and element.multiple:
+                        element_contents.extend([
+                            element.options.get(selected_value, '')
+                            for selected_value in element.value
+                        ] if is_options_dict else element.value)
+                    else:
+                        element_contents.append(
+                            element.options.get(element.value, '') if is_options_dict else element.value)
                     if not isinstance(element, Select) or element.is_showing_popup:
                         element_contents.extend(options)
                 if any(all(needle not in str(haystack) for haystack in element_contents) for needle in self._contents):
--- a/nicegui/element_filter.py
+++ b/nicegui/element_filter.py
@@ -120,7 +120,13 @@ class ElementFilter(Generic[T]):
                     element_contents.append(element.message)
                 if isinstance(element, (Select, Radio, Toggle)):
                     options = {option['value']: option['label'] for option in element.props.get('options', [])}
-                    element_contents.append(options.get(element.value, ''))
+                    selected_values = element.value
+                    if not isinstance(selected_values, list):
+                        selected_values = [selected_values]
+                    element_contents.extend([
+                        options.get(selected_value, '')
+                        for selected_value in selected_values
+                    ])
                     if not isinstance(element, Select) or element.is_showing_popup:
                         element_contents.extend(options.values())
                 if any(all(needle not in str(haystack) for haystack in element_contents) for needle in self._contents):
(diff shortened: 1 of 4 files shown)


rodja:
Wow. I finally found the time to look at your PR @Daemo00. It seems that the select element is very poorly supported at the moment. I extended your test and added two more to uncover a lot of problems with the current implementation. I do not have a solution yet...

rodja:
Ok. I think I found an implementation which passes the tests. The code is just a first shot; there might be better ways to implement the behaviour we are seeking. And it's worth noting that the props are not updated. `model-value` for example will always stay at index `0`. Is that a problem for the user simulation?

nicegui/element_filter.py
```diff
@@ -118,11 +117,13 @@ def __iter__(self) -> Iterator[T]:
...
                 if isinstance(element, Notification):
                     element_contents.append(element.message)
-                if isinstance(element, (Select, Radio, Toggle)):
-                    options = {option['value']: option['label'] for option in element.props.get('options', [])}
-                    element_contents.append(options.get(element.value, ''))
+                if isinstance(element, ChoiceElement):
+                    is_options_dict = isinstance(element.options, dict)
+                    options = element.options.values() if is_options_dict else element.options
+                    element_contents.append(
+                        element.options.get(element.value, '') if is_options_dict else element.value)
```
Daemo00: When the `ui.select` is `multiple`, `element.value` is a `list` so
```py
element.options.get(element.value, '')
```
raises
> \>                       element.options.get(element.value, '') if is_options_dict else element.value)
> E                   TypeError: unhashable type: 'list'

I added a the commit https://github.com/zauberzeug/nicegui/pull/4449/commits/996fa08b3176bcfe9913e71b2b7b7260fb0b1407 with test and fix for this use-case.

Daemo00 (commented):
Thanks for having a look!

> Ok. I think I found an implementation which passes the tests. The code is just a first shot; there might be better ways to implement the behaviour we are seeking. And it's worth noting that the props are not updated. `model-value` for example will always stay at index `0`. Is that a problem for the user simulation?

I'm not using props that much so it's not a problem for me at the moment.

I have found an issue with the current implementation though, see the comment below.

rodja (approved):

falkoschindler (approved):

falkoschindler (approved):
Thanks a lot for your pull request, @Daemo00!
I also reviewed the code and cleaned it up here and there. Let's merge! 🚀

PR #4450, opened 2025-03-09
support check_interval argument in @ui.page decorator

evnchn:
This PR serves to enhance the loading speed of tiny async pages. 

It's a small PR. 3 lines of additions only. 😅 and it's my first PR here. 

```py
@ui.page("/async")
async def async_page(): # 0.1 seconds at minimum
    pass

@ui.page("/sync")
def sync_page(): # near instant
    pass

@ui.page("/async_enhanced", check_interval=0.01)
async def async_page_enhanced(): # now also near instant
    pass
```

async

sync

async_enhanced

These 3 pages all do nothing, and it is my personal instinct that it will take about as long to complete. 

However, as it turns out, because the way to handle async page functions in the decorator is to poll for its completion at 0.1 second interval, the async page takes 0.1 seconds. 

**Note: This isn't really a bottleneck for any decently large page, or when operating over the internet, but could matter for localhost applications where instantaneous response is desired. There are performance implications with polling at a rate too high, and the user of this new feature is expected to time their own code to find what suits themselves best.** 

In the PR, `@ui.page` decorator now supports `check_interval`, similar to `Client.connected`. 

Attached please also find the speedtest script and the associated middleware to add, should you want to recreate the speedtest. It could be useful to add to the examples or even bake the timing functionality into NiceGUI itself, but I am not sure I can handle that just now...

[speed_investigation.zip](https://github.com/user-attachments/files/19151009/speed_investigation.zip)

EDIT: A space after the bolded section made it non-bold in the markdown 😒

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3512,17 +3512,35 @@ files = [
 
 [[package]]
 name = "starlette"
-version = "0.44.0"
+version = "0.41.3"
 description = "The little ASGI library that shines."
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "starlette-0.44.0-py3-none-any.whl", hash = "sha256:19edeb75844c16dcd4f9dd72f22f9108c1539f3fc9c4c88885654fef64f85aea"},
-    {file = "starlette-0.44.0.tar.gz", hash = "sha256:e35166950a3ccccc701962fe0711db0bc14f2ecd37c6f9fe5e3eae0cbaea8715"},
+    {file = "starlette-0.41.3-py3-none-any.whl", hash = "sha256:44cedb2b7c77a9de33a8b74b2b90e9f50d11fcf25d8270ea525ad71a25374ff7"},
+    {file = "starlette-0.41.3.tar.gz", hash = "sha256:0e4ab3d16522a255be6b28260b938eae2482f98ce5cc934cb08dce8dc3ba5835"},
 ]
 
 [package.dependencies]
 anyio = ">=3.4.0,<5"
+typing-extensions = {version = ">=3.10.0", markers = "python_version < \"3.10\""}
+
+[package.extras]
+full = ["httpx (>=0.22.0)", "itsdangerous", "jinja2", "python-multipart (>=0.0.7)", "pyyaml"]
+
+[[package]]
+name = "starlette"
+version = "0.46.0"
+description = "The little ASGI library that shines."
+optional = false
+python-versions = ">=3.9"
+files = [
+    {file = "starlette-0.46.0-py3-none-any.whl", hash = "sha256:913f0798bd90ba90a9156383bcf1350a17d6259451d0d8ee27fc0cf2db609038"},
+    {file = "starlette-0.46.0.tar.gz", hash = "sha256:b359e4567456b28d473d0193f34c0de0ed49710d75ef183a74a5ce0499324f50"},
+]
+
+[package.dependencies]
+anyio = ">=3.6.2,<5"
 typing-extensions = {version = ">=3.10.0", markers = "python_version < \"3.10\""}
 
 [package.extras]
@@ -4221,4 +4239,4 @@ sass = ["libsass"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8"
-content-hash = "65f3728b11c9b7993e370258746c8546308ab88be5239a138e5595429d2571e5"
+content-hash = "9d56e02408ddf68596416c26400f7b6c3e93d88701b0ad9637fd08dbbefa9bc8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,10 @@ python = "^3.8"
 typing-extensions = ">=4.0.0"
 markdown2 = ">=2.4.7,!=2.4.11"
 Pygments = ">=2.15.1,<3.0.0"
+starlette = [
+    {version = "<0.42.0", python = "~3.8"},  # https://github.com/zauberzeug/nicegui/issues/4255
+    {version = ">=0.45.3", python = ">=3.9"}
+]
 uvicorn = {extras = ["standard"], version = ">=0.22.0"}
 fastapi = ">=0.109.1"
 python-socketio = {extras = ["asyncio-client"], version = ">=5.10.0"} # version min: see https://github.com/zauberzeug/nicegui/issues/1809


evnchn:
Oh, here's what I did to add some timing functionality to `client.py` and `page.py` with regards to the time taken to execute the page function, and build the page. Achieved by directly patching the source code of NiceGUI 2.11.1 (the version I happpened to be using for my main project)

`client.py`

```py
    def build_response(self, request: Request, status_code: int = 200, before_call_func_time="", after_call_func_time="") -> Response:
        """Build a FastAPI response for the client."""
        just_before_build_response_time = f"{time.perf_counter():.10f}"
        self.outbox.updates.clear()
        prefix = request.headers.get('X-Forwarded-Prefix', request.scope.get('root_path', ''))
        elements = json.dumps({
            id: element._to_dict() for id, element in self.elements.items()  # pylint: disable=protected-access
        })
        socket_io_js_query_params = {
            **core.app.config.socket_io_js_query_params,
            'client_id': self.id,
            'next_message_id': self.outbox.next_message_id,
        }
        vue_html, vue_styles, vue_scripts, imports, js_imports = generate_resources(prefix, self.elements.values())
        just_after_build_response_time = f"{time.perf_counter():.10f}"
        return templates.TemplateResponse(
            request=request,
            name='index.html',
            context={
                'request': request,
                'version': __version__,
                'elements': elements.replace('&', '&amp;')
                                    .replace('<', '&lt;')
                                    .replace('>', '&gt;')
                                    .replace('`', '&#96;')
                                    .replace('$', '&#36;'),
                'head_html': self.head_html,
                'body_html': '<style>' + '\n'.join(vue_styles) + '</style>\n' + self.body_html + '\n' + '\n'.join(vue_html),
                'vue_scripts': '\n'.join(vue_scripts),
                'imports': json.dumps(imports),
                'js_imports': '\n'.join(js_imports),
                'quasar_config': json.dumps(core.app.config.quasar_config),
                'title': self.resolve_title(),
                'viewport': self.page.resolve_viewport(),
                'favicon_url': get_favicon_url(self.page, prefix),
                'dark': str(self.page.resolve_dark()),
                'language': self.page.resolve_language(),
                'prefix': prefix,
                'tailwind': core.app.config.tailwind,
                'prod_js': core.app.config.prod_js,
                'socket_io_js_query_params': socket_io_js_query_params,
                'socket_io_js_extra_headers': core.app.config.socket_io_js_extra_headers,
                'socket_io_js_transports': core.app.config.socket_io_js_transports,
            },
            status_code=status_code,
            headers={'Cache-Control': 'no-store', 'X-NiceGUI-Content': 'page', 
                     'X-Just-Before-Build-Response-Time': just_before_build_response_time, 
                     'X-Just-After-Build-Response-Time': just_after_build_response_time, 
                     'X-Before-Call-Func': before_call_func_time, 'X-After-Call-Func': after_call_func_time}
        )
```

`page.py`

```py
        @wraps(func)
        async def decorated(*dec_args, **dec_kwargs) -> Response:
            request = dec_kwargs['request']
            # NOTE cleaning up the keyword args so the signature is consistent with "func" again
            dec_kwargs = {k: v for k, v in dec_kwargs.items() if k in parameters_of_decorated_func}
            with Client(self, request=request) as client:
                if any(p.name == 'client' for p in inspect.signature(func).parameters.values()):
                    dec_kwargs['client'] = client
                before_call_func_time = f"{time.perf_counter():.10f}"
                result = func(*dec_args, **dec_kwargs)
            if helpers.is_coroutine_function(func):
                async def wait_for_result() -> None:
                    with client:
                        return await result
                task = background_tasks.create(wait_for_result())
                deadline = time.time() + self.response_timeout
                while task and not client.is_waiting_for_connection and not task.done():
                    if time.time() > deadline:
                        raise TimeoutError(f'Response not ready after {self.response_timeout} seconds')
                    await asyncio.sleep(self.check_interval)
                if task.done():
                    result = task.result()
                else:
                    result = None
            if isinstance(result, Response):  # NOTE if setup returns a response, we don't need to render the page
                return result
            binding._refresh_step()  # pylint: disable=protected-access
            after_call_func_time = f"{time.perf_counter():.10f}"
            return client.build_response(request, before_call_func_time=before_call_func_time, after_call_func_time=after_call_func_time)
```

It is a bit ugly but get's the job done...

evnchn:
Here's 2 shower thoughts related to `while task and not client.is_waiting_for_connection and not task.done():`

1. If the page builder function never calls `await ui.context.client.connected` and never hangs, we can just await for the task as normal. Potentially, this can be the behaviour if `check_interval` is set to `None`, where the user acknowledges that they must not await for the client, and they must write good code that does not hang indefinitely or bring down the server. 

2. Still, it is quite desirable to perform actions such as awaiting for the client to connect then do a notify message. Maybe the page can accept a callback which is executed after client connects? Or the page is actually a class with its `create_page` (can't be the __init__ because __init__ can't be async) and `on_client_connect` methods

falkoschindler:
Thanks for sharing this interesting investigation, @evnchn!

I would have expected that the while-loop
```py
while task and not client.is_waiting_for_connection and not task.done():
    if time.time() > deadline:
        raise TimeoutError(f'Response not ready after {self.response_timeout} seconds')
    await asyncio.sleep(0.1)
```
exits immediately without sleeping once. But apparently the task has no chance to start, let alone to complete, before reaching the first `await asyncio.sleep(0.1)` call.

What would happen if we add an `await asyncio.sleep(0)` call right before starting the loop? It should avoid sleeping for 100ms, at least in trivial cases like yours.

Of course, making the check interval configurable wouldn't do much harm. But I'm always a bit resistent to add yet another parameter which complicates the API a little more.

Regarding your suggestions "Maybe the page can accept a callback which is executed after client connects?": You can use the client's `on_connect` hook.
```py
@ui.page('/')
def async_page():  # 0.1 seconds at minimum
    ui.context.client.on_connect(lambda: ui.label('The client has connected'))
```

falkoschindler:
My point is: `await asyncio.sleep(0)` will put the current task back into the event loop, giving other tasks a chance to run. This can be our new `task` that runs `wait_for_result`. If like in your case, the async page function doesn't await anything, this task will finish before any other task from the event loop can continue. Therefore `await asyncio.sleep(0)` is enough to let any async page function _without_ async calls complete without sleeping 100ms inside the while loop.

But maybe we need a more reasonable example. Your `async_page` without any async call is kind of an edge case. We assumed that, if the page function is async, there's probably a reason for it and it won't terminate in a few milliseconds. Given the overall load time caused by network transmission and other parts of the pipeline, 100ms did seem like a good choice. It doesn't spam the event loop with lots and lots of interruptions, and it should be barely noticeable be the user. Again: We assume there is significant work that needs to be awaited anyway.

Of course, this is no veto against another parameter. I'm just exploring alternatives. Leaving it up to the user can be suboptimal, because the devil is in the detail: In #2482, for example, we fought a nasty bug that happened due to a 16ms timing limit on Windows. So we would probably need to impose a minimum 16ms threshold for the new `check_interval`.

evnchn:
Wow. I learnt something new today! Conclusion: `await asyncio.sleep(0)` is actually useful! 

I actually didn't know that `await asyncio.sleep(0)` has this kind of behaviour. And I have no idea about the 16ms issue on Windows (thinking faster = better 😅). I actually running the server with 10ms check interval for JS, and so I should go back and change that. 

I agree that an async page without async calls is itself a minority, but I disagree in a sense that it does **not** _represent_ a minority. Note my word use here. I am saying this because in the context of this discussion, an async page without async calls is the same as an async page with all async calls after `await ui.context.client.connected()`, which occurs quite frequently. 

Again, our check is `while task and not client.is_waiting_for_connection and not task.done():`, so either the task completing or `client.is_waiting_for_connection==True` (which is set by `await ui.context.client.connected()`) should get us out of the loop. 

For example, consider this psuedocode. It's taken from my project I am working on right now, but I don't quite have the time to produce a minimal example for this one...

```py
async def get_client_js_speed():
    # probably something to do with const data = [...Array(1000).keys()]; data.find(x => x == 800)
    return await ui.run_javascript("some speedtesting code") 

@ui.page("/what_is_my_speed")
async def what_is_my_speed():
    ui.label("What is my speed?")
    await ui.contect.client.connected()
    result = await get_client_js_speed()
    ui.label(result)
```

It is necessary to run javascript only after client is connected, otherwise, it hangs the page response and it always times out. 

But, in the above implementation, 100ms will be spent waiting for the page function, despite it is already finished and waiting for the client in 2ms. 

Meanwhile, without the `await asyncio.sleep(0)` patch, the following should be good for accomplishing 100ms-free page load, but I feel my hands are forced to use sync page even though it should be fine to use either one. So, the `await asyncio.sleep(0)` would still be helpful. 

```py
async def get_client_js_speed():
    # probably something to do with const data = [...Array(1000).keys()]; data.find(x => x == 800)
    return await ui.run_javascript("some speedtesting code") 

async def generate_speed_element():
    result = await get_client_js_speed()
    ui.label(result)

@ui.page("/what_is_my_speed")
def what_is_my_speed(): # sync page, though
    ui.label("What is my speed?")
    ui.context.client.on_connect(generate_speed_element)
```

Would you want to add that one line yourself for simplicity's sake?

evnchn:
moving my change to another branch. Should have done that since the beginning 🤦‍♂️

PR #4456, opened 2025-03-11
Introduce `ValueChangeEventArguments.previous_value`

falkoschindler:
Inspired by discussion #4410, this PR introduces a `previous_value` field for `ValueChangeEventArguments`. This change is pretty straightforward and only requires adjustments for two mixins and two elements.

Right now it is lacking backward-compatibility though. If user code derives custom `ValueChangeEventArguments`, they are now missing the `previous_value` field. Should we add a default value? But it should be clear that this default is _not_ the previous value, but "not set". We might need a sentinel...

And in 3.0 I'd like to enforce the `previous_value` field. A deprecation warning for the default value would be great, but probably hard to achieve.


evnchn:
Can this work?

```py
@dataclass(**KWONLY_SLOTS)
class ValueChangeEventArguments:
    value: Any
    previous_value: Any = None

    def __post_init__(self):
        if self.previous_value is None:
            warnings.warn(
                "The 'previous_value' is now part of the class. Please update your code before NiceGUI 3.0",
                DeprecationWarning
            )
```

falkoschindler:
@evnchn You're right, thanks! I expected it to be more difficult because it reminded me of a problem from another project with a more complex deprecation scenario. Here it is pretty straightforward.

evnchn:
Less is more

More or less 😉

evnchn (commented):
First time to learn about the `...` (Elipsis) object in Python 😅, but that should work great in this contex

falkoschindler:
We just decided to postpone this PR to 3.0 because it could break existing user code:

- If we keep the default value `...`, it would break user code that derives from `ValueChangeEventArguments` and adds additional arguments without default value.
- If we remove the default value `...`, it would break user code that derives from `ValueChangeEventArguments` and doesn't initialize the `previous_value` attribute.

falkoschindler:
I think we can introduce the `previous_value` argument in 3.0 but keep the deprecation warning until 4.0. This way downstream projects have some time to migrate.

rodja (approved):

Issue #4397, opened 2025-02-27
ui.json_editor uncaught exceptions


natankeddem:
### Description

Hello,
I found some unexpected behaviour using ui.json_editor. There are `on_change` and `on_select` events firing with `e.args` as an empty dict (ie `{}`). I found this by putting a print messages in `json_editor.py`. I am not exactly sure how `e.args` can be empty since the `json_editor.js` explicitly populates the object for the `emit`. When putting breakpoints in the `json_editor.js` I don't hit anything but I still seem to get the empty dict over in the python side, which is strange to me. The condition that causes the errors seem to be manipulating the text path feature. This code will show notifications for events and a traceback on error:
```py
import traceback
from nicegui import ui, app

c = ui.column()
with c:
    json = {
        "xxx": {
            "yyy": {
                "zzz": {
                    "a": "b",
                    "c": "d",
                }
            }
        }
    }
    ui.json_editor({"content": {"json": json}}, on_select=lambda e: ui.notification(f"Select: {e}"), on_change=lambda e: ui.notification(f"Change: {e}"))


def on_error():
    trace = traceback.format_exc()
    with c:
        ui.notification(trace, type="negative")


app.on_exception(on_error)
ui.run(reload=False, show=False)
```

https://github.com/user-attachments/assets/a77e952f-07c1-44e5-87d2-06159d7ebcaa

natankeddem:
It is almost as if `svelte-jsoneditor` is emitting on its own. The only way I could make the events stop firing was to alter the actual `on` method `type` argument.
https://github.com/zauberzeug/nicegui/blob/b71074ef73498f532234a837ea04a08ffbf4e804/nicegui/elements/json_editor.py#L47C1-L59C20

```py
    def on_change(self, callback: Handler[JsonEditorChangeEventArguments]) -> Self:
        """Add a callback to be invoked when the content changes."""
        def handle_on_change(e: GenericEventArguments) -> None:
            handle_event(callback, JsonEditorChangeEventArguments(sender=self, client=self.client, **e.args))
        self.on('change_i_changed_this', handle_on_change, ['content', 'errors'])
        return self


    def on_select(self, callback: Handler[JsonEditorSelectEventArguments]) -> Self:
        """Add a callback to be invoked when some of the content has been selected."""
        def handle_on_select(e: GenericEventArguments) -> None:
            handle_event(callback, JsonEditorSelectEventArguments(sender=self, client=self.client, **e.args))
[...]

PR #4458, opened 2025-03-11
Fix native select and change event capture for `json_editor` element.

natankeddem:
The `svelte-jsoneditor` library seems to have some native events which utilize the `select` and `change` naming. This causes exceptions once these reach the `NiceGUI` event handlers without expected arguments. This PR changes these names to remove the conflict. Per issue https://github.com/zauberzeug/nicegui/issues/4397.


falkoschindler (approved):
Thanks for the pull request, @natankeddem!
Works as expected. Let's merge! 🚀

PR #4463, opened 2025-03-12
replaceUndefinedAttributes 5x faster with caching 

evnchn:
(and probably comparatively even faster with more on-screen elements)

Test page: https://github.com/[evnchn/nicegui-speed-investigation](https://github.com/evnchn/nicegui-speed-investigation)

Tested on /native/after_conn/basic
Before change: 
image

After change: 
image

For /native/after_conn/js, speed increase roughly lines up with https://github.com/zauberzeug/nicegui/discussions/4457#discussion-8067069

I am lazy to take more screenshots 😅

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -19,8 +19,20 @@ function parseElements(raw_elements) {
   );
 }
 
-function replaceUndefinedAttributes(elements, id) {
-  const element = elements[id];
+const noUndefinedAttributes = new Set(); // Expects integers
+
+// Deprecates the old replaceUndefinedAttributes function, as it was slow against Vue mounted objects and not needed anymore.
+function replaceUndefinedAttributes_singleelement(element, element_id) {
+  // speedtest results yields that this is 5x faster than the old solution
+
+  // DO NOT UNCOMMENT THE LINE BELOW!!! Because updated elements are also sent with compact format (#2702), it would yield errors. 
+  // if (noUndefinedAttributes.has(element_id)) return; 
+
+  // check type of element_id, it should be integer, if not, convert and log warning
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):

I had to clean up a bit to see what actually changed.
Two thoughts:
- I don't think we need to convert string IDs to int. It is a pity that JavaScript doesn't support integer object keys. But we can simply work with string IDs on the client. (See my first commit.)
- You're never resetting the set of processed element IDs. This means that future updates will only be handled because elements that are directly mentioned in an update message will be processed without checking the set. Only recursive calls are filtered. This made we wonder if we can simply remove the recursion (and the set). Empirically it seems to work. (See my second commit.) What do you think?

evnchn:
{2FFFC616-9BD0-4BB7-B119-0FC6A483BA66}

Well that answers the first point. But idk JavaScript enough about the speed of integer comparisons to string comparisons.

evnchn:
For point 2, maybe we should run the test. 

Honestly speaking, this recursive call thing with the set for anti-call-twice, had me some flashbacks when I was in the algorithms course. My brain hurts to think about it, and my approach was always to comment stuff / add stuff until it works.

evnchn:
Can confirm that it seems to work empirically. 

Also the "Update branch" is not my intention! Clicked it wrongly! @falkoschindler I am not sure I can fix... GitHub should NOT show a button that throws the changes from you into the bin...

If you want to be extra sure, maybe we can ship this as an optional feature that can be turned on with a special flag, and when enough people feedback it is OK, then we make the new behaviour the default.

falkoschindler (approved):
I'm pretty sure the code is fine. All tests are green and the performance improvement should be significant, even without cache. Let's merge! 🚀

PR #4464, opened 2025-03-12
feat: add `reset()` method for components

weinibuliu:
This PR is based on #4460 that which discusses if add `reset()` method for components. 

**It is still a draft.** Its main purpose at the moment is to show how I designed the `reset` method and to make it easy for people to see the design.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -38,6 +38,7 @@ class Element(Visibility):
     _default_props: ClassVar[Dict[str, Any]] = {}
     _default_classes: ClassVar[List[str]] = []
     _default_style: ClassVar[Dict[str, str]] = {}
+    _UNCHANGED = object()
 
     def __init__(self, tag: Optional[str] = None, *, _client: Optional[Client] = None) -> None:
         """Generic Element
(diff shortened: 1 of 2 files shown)


evnchn:
Well, the reconfiguration paradigm is what people expect, sometimes. Check https://github.com/zauberzeug/nicegui/discussions/4484

Not sure if we should proceed, though.

evnchn:
It's been a month. Do we want this feature? 

Personally, with the use case I seen @weinibuliu demonstrate, it is mostly for totally updating the element with a new one. Then, we won't need to do it the new way, and it's probably better to delete the old element and make one anew. 

While, for updating, writing to individual property still seems wise. 

So, the feature only has benefits where we are updating just enough properties for the element to make it a chore to do it via setters, while not so much it becomes desired to create element anew. 

---

Another burning question: what happens to bind values? Suppose I bind the value of an input field to a variable. If I do the `.reconfigure`, will it then pump the value into the variable? 

We'll best test this via running the code, but seems like the current implementation may not account for this, and we could be tearing the reality apart if bindings are broken with the new `.reconfigure` method.

falkoschindler:
As far as I can tell, there are still several [open questions](https://github.com/zauberzeug/nicegui/discussions/4460#discussioncomment-12487239) regarding this feature, and the binding subject is another one. Demand for this feature is limited and I agree that "the feature only has benefits where we are updating just enough properties for the element to make it a chore to do it via setters". We can think about dedicated methods for such situations like for `ui.notification` in PR #4444. But I don't see the need for a generic solution.

PR #4462, opened 2025-03-12
await asyncio.sleep(0) gives a chance the task to start

evnchn:
https://github.com/zauberzeug/nicegui/pull/4450#issuecomment-2717173679

Otherwise, async pages with (1) no awaits in them at all or (2) all awaits after `await ui.context.client.connected()` incur 100ms latency. 

Test page: 

```py
@ui.page("/example2")
async def page_example2():
    pass
```

With the change: 

{A43F5BF4-B79B-4062-9D94-3F0D7623C225}

Without the change: 

{4F85CB08-29BC-4EF8-9203-AD0CD8851255}

PR #4461, opened 2025-03-12
support check_interval argument in @ui.page decorator

evnchn:
Changes originally for https://github.com/zauberzeug/nicegui/pull/4450

But I made it to my repo's main, which was not a smart move at all!

Now it's back

PR #4466, opened 2025-03-12
Introduce `_waiting_for_connection` event for more efficient page function

falkoschindler:
@evnchn After thinking about PRs #4461 and #4462, I came up with another idea how we can wait for async page functions more efficiently. By introducing a new "waiting for connection" event, we can simultaneously wait for both:
1. our task waiting for the page function to return
2. the new event which occurs when the page function calls `await client.connected()`.

Furthermore, we can specify a timeout very easily.

@evnchn What do you think? Does that solve our problem?

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -56,6 +56,7 @@ class Client:
 
         self.elements: Dict[int, Element] = {}
         self.next_element_id: int = 0
+        self._waiting_for_connection: asyncio.Event = asyncio.Event()
         self.is_waiting_for_connection: bool = False
         self.is_waiting_for_disconnect: bool = False
         self.environ: Optional[Dict[str, Any]] = None
@@ -177,6 +178,7 @@ class Client:
     async def connected(self, timeout: float = 3.0, check_interval: float = 0.1) -> None:
         """Block execution until the client is connected."""
         self.is_waiting_for_connection = True
+        self._waiting_for_connection.set()
         deadline = time.time() + timeout
         while not self.has_socket_connection:
             if time.time() > deadline:
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import inspect
-import time
 from functools import wraps
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Optional, Union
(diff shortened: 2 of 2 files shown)


evnchn:
Amazing. No-polling solution is always better than any polling solution. 

I don't know the asyncio library to know there is `asyncio.wait` let alone the fact that there is `timeout` and `return_when=asyncio.FIRST_COMPLETED` 😅

evnchn:
Note: is_waiting_for_connection and is_waiting_for_disconnect appears to be unused?

falkoschindler:
> Note: is_waiting_for_connection and is_waiting_for_disconnect appears to be unused?

Yes, but technically it is public API, so we could only deprecate it. But maybe it's still a useful attribute for some user code.

evnchn:
@falkoschindler Look at the below:

https://github.com/zauberzeug/nicegui/blob/8d48ec7236fcd8ed55a0c499caf0785581515a0c/nicegui/client.py#L105-L108

https://github.com/zauberzeug/nicegui/blob/8d48ec7236fcd8ed55a0c499caf0785581515a0c/nicegui/outbox.py#L79-L81

https://github.com/zauberzeug/nicegui/blob/8d48ec7236fcd8ed55a0c499caf0785581515a0c/nicegui/client.py#L183-L186

By the same logic, we _could_ (?) introduce `_wait_for_socket_connection` event, so that there would be no need to poll against a socket connection, with whoever that sets the tab_id for the client also update the event: 

https://github.com/zauberzeug/nicegui/blob/8d48ec7236fcd8ed55a0c499caf0785581515a0c/nicegui/nicegui.py#L168

---

But, due to the code placed very close to the edited area of the infamous "software rugpull" commit https://github.com/zauberzeug/nicegui/pull/4487 (which I am still waiting for the review), I am hesitant to edit it...

evnchn:
Sudden thought: We should modify `is_waiting_for_connection`, at the very least, returning `_waiting_for_connection.is_set()`, so that we don't have duplicate logic. As of writing, when a connection is made, one would need to do these 2 lines: 

https://github.com/zauberzeug/nicegui/blob/ecd02ae746bf9154e98f74e9c15294bf9b1ab244/nicegui/client.py#L180-L181

which seems a bit duplicate...

rodja:
`is_waiting_for_connection` is part of the public api and can only be removed in 3.0.

evnchn:
Yes, but we can probably still refactor the code to make the underlying logic of `is_waiting_for_connection` simpler, without breaking compatibility. 

Though: Not a very high urgency item right now. I'd do the icon selector instead.

Issue #4412, opened 2025-03-03
Multiprocessing made code reload


zak-45:
### Description

Working platform: Win 11, Python 3.12, NiceGUI 2.11.1

Each time a new process is created, NiceGUI script is reloaded.

Here is my test code :
```py
import multiprocessing
import sys
from nicegui import ui

def mp_setup():
    if sys.platform.lower() != 'linux':
        return multiprocessing.Process, multiprocessing.Queue  # Direct return
    ctx = multiprocessing.get_context('spawn')
    return ctx.Process, ctx.Queue

Process, Queue = mp_setup()

def mycmd():
    print('into my cmd')

def proc():
    new_process = Process(target=mycmd)
    # start the child process
    new_process.start()
    print(f'new process started {new_process}')

print('start')

ui.button('start new process', on_click=proc)

ui.run(reload=False,)

print('end')

```

When you click on button, this is the output : 
```
> new process started 'Process name='Process-1' pid=27204 parent=6508 started'
start
end
into my cmd
new process started 'Process name='Process-2' pid=3496 parent=6508 started'
start
end
into my cmd
new process started 'Process name='Process-3' pid=14800 parent=6508 started'
start
end
into my cmd
```

Maybe I misunderstand, but I expected something like that: 
```
> start
new process started Process name='Process-1' pid=27204 parent=6508 started
into my cmd
new process started 'Process name='Process-2' pid=3496 parent=6508 started'
into my cmd
new process started 'Process name='Process-3' pid=14800 parent=6508 started'
into my cmd
end
```

falkoschindler:
Thanks for reporting this issue, @zak-45!

This problem is currently being discussed in #4353. It it somewhat expected behavior in Python, but we're still looking for a way to avoid it when running NiceGUI apps.

zak-45:
Thanks for your quick response. I take a look to the discussion, put the 'main guard' in my test code, this provide what expected (from my side): 

```
if __name__ == '__main__':
    print('start')
    ui.button('start new process', on_click=proc)
    ui.run(reload=False,)
    print('end')
```
this give me this output: 
>new process started <Process name='Process-1' pid=13540 parent=35804 started>
into my cmd
new process started <Process name='Process-2' pid=27436 parent=35804 started>
into my cmd
new process started <Process name='Process-3' pid=6472 parent=35804 started>
into my cmd
new process started <Process name='Process-4' pid=5124 parent=35804 started>
into my cmd

[...]

PR #4471, opened 2025-03-13
Create a proof-of-concept for fast cpu_bound with lazy import

falkoschindler:
This PR is a proof-of-concept for faster multiprocessing by lazily filling `sys.modules` with dummy modules if NiceGUI is already running. It is related to and inspired by the following discussions:

- https://github.com/zauberzeug/nicegui/issues/3356
- https://github.com/zauberzeug/nicegui/pull/3365
- https://github.com/zauberzeug/nicegui/discussions/4353
- https://github.com/zauberzeug/nicegui/issues/4412
- https://github.com/zauberzeug/nicegui/pull/4429
- https://github.com/zauberzeug/nicegui/discussions/4465

Note: We don't need to move run.py in a separate module as originally planned because nicegui is now cheap to import in subprocesses.

There are still import problems when running with `reload=True` and exceptions when stopping the app with Ctrl-C.


evnchn:
Just want to mention: 

https://github.com/hxu296/tariff

They patched the behaviour of `import` via `builtins.__import__ = _tariffed_import`

Seems an interesting approach. Not sure if we can use a similar approach to enable lazy importing.

falkoschindler:

PR #4480, opened 2025-03-15
Better error screen supporting clean traceback for sync/async

evnchn:
So, when an error occurs before this PR: 

* User sends me a screenshot which is useless in debugging
* Lots gets printed in the console
* Console traceback is very wordy with many calls from NiceGUI
* Traceback ends at `result = task.result()` for async page functions which makes it useless

After this PR: 
* User sends me actionable information
* Console remains quiet
* Traceback shown in the browser is clean with all calls from NiceGUI neatly hidden
* Async page functions fully supported

Test code: 
```py
from nicegui import ui

def my_errorenous_function():
    raise ValueError("This is an error")

@ui.page("/sync_error")
def error_page():
    ui.label("This is a sync error page")
    my_errorenous_function()

@ui.page("/async_error")
async def async_error_page():
    ui.label("This is an async error page")
    my_errorenous_function()

ui.run()
```

{50AFCDCF-916D-46CE-99A7-2B36850F6504}

After clicking button: 

{C6A8F061-A7AF-4F19-A113-72E5758BE2E5}

**NOTE: Could be privacy concern as the path of the file in question is printed.** 

Things to decide on: 
* Whether to put this new behaviour behind a switch, and if so, is that switch default-True or default-False
* What is the better UI for displaying the error

Things left to fix:
* I don't know why, but when I scroll horizontally the error, the "Hide details" button disappears


rodja:
We think it's not a good idea to show backend stacktraces on the web site error pages. This can be really problematic in production or even in quick shared demos. Of course it could be usefulness in development. Maybe it can be implemented as a separate package or even a gist?

falkoschindler:
You can simply add a custom exception handler in user code:
```py
@app.exception_handler(Exception)
async def _exception_handler_500(request: Request, exception: Exception) -> Response:
    with Client(ui.page(''), request=request) as client:
        ui.label('This is my custom error page')
        ui.log().push(traceback.format_exc(chain=False).strip())
    return client.build_response(request, 500)
```

evnchn:
IIRC the call stack would end at `result = task.result()` if that is the case. 

We could put this functionality behind a normally-off switch

evnchn:
Doesn't work when the page is async...

{15A559D2-171C-4EBF-BB7C-5EEF2326DDAE}

This has something to do with how async functions work. But, I don't really understand it let alone explain it here 😅. Maybe you can take a shot!

Also, the log is lengthy in both cases. 

{86E6C566-6D56-4866-909D-30BB1BFA1F46}

My understanding is that, if you want to handle an exception and the traceback call stack remains short, you must handle the exception then and there. If you wait until the exception to be raised all the way to the top at `@app.exception_handler(Exception)`, everything in between gets printed in the traceback. 

For this reason, I am thinking `app.add_exception_handler`, which puts the user's exception handler somewhere, so that it can be immediately called if supplied, or fallback to old behaviour if not. Then, the new error page can be put somewhere in the library. 

So, the clause to add the exception handler would be something like 

```py
from nicegui.error import traceback_error_page
app.add_exception_handler(traceback_error_page)
```

evnchn:
Mark as draft, because https://github.com/zauberzeug/nicegui/pull/4538 seems more favourable. If you agree, just close this PR and focus on the other one.

evnchn:
Just so you know, won't delete branch from my fork, because code here can still be valuable.

PR #4481, opened 2025-03-15
Set explicitly font-display: block to exchange "flash of unstyled text" (FOUT) for “flash of invisible text” (FOIT)?

evnchn:
https://github.com/zauberzeug/nicegui/discussions/4477


evnchn:
TL-DR: Avoid flashing text like "notifications" when the icons are loading, but instead stuff will just be invisible while load. 

After a long thought, I'm considering whether to do this, since it _could_ be possible to use codepoints to sidestep this issue.

evnchn:
Works for NiceGUI buttons which you have control over what is inside only. Does not work for Quasar Components: 

```py
from nicegui import ui

@ui.page("/")
def main_page():
    make_disappear_when_loaded = []
    make_appear_when_loaded = []

    with ui.page_sticky():
        ui.icon("notifications").classes("w-[0px] h-[0px] opacity-0") # force the icon to load

    with ui.button().props("fab"):
        system_bell = ui.label("🔔").style('font-family: "Segoe UI Symbol"; font-size: 24px;')
        make_disappear_when_loaded.append(system_bell.id)


        icon_bell = ui.icon("notifications")
        icon_bell.set_visibility(False)
        make_appear_when_loaded.append(icon_bell.id)

    ui.add_body_html(f"""
    <script>
    document.fonts.ready.then(function() {{
        const make_disappear_when_loaded = {make_disappear_when_loaded};
        const make_appear_when_loaded = {make_appear_when_loaded};
        for (const id of make_disappear_when_loaded) {{
            getHtmlElement(id).classList.add('hidden');
        }}
        for (const id of make_appear_when_loaded) {{
            getHtmlElement(id).classList.remove('hidden');
        }}
    }});
    </script>
    """)

ui.run(port=8082)
```

---

The need for this PR diminishes more and more

falkoschindler (changes requested):
Hi @evnchn,

Changing FOUT (flash of unstyled text) to FOIT (flash of invisible text) seems to be a good idea for icons. Let me comment on some of your points from #4477 before looking into this PR:

> Issue 1: The block period is 3 seconds for most browsers, but Safari is infinite (up to 1 year).

If you see unstyled icons for more than a few seconds, your connection is probably broken anyway, so a button with such an icon is of little use. Therefore we better hide them.

> Issue 2: Does Google recommends against this? Or not?

I agree that icons are not written content. They are more like SVG graphics that cannot be rendered before being fetched.

> Therefore, I propose that we change to font-display: swap for the Material Icons and Material Symbols.

Are you sure you mean "swap"? In this PR you use "block".

---

Regarding this PR:

The file fonts.css is auto-generated using fetch_google_fonts.py. We should update this script accordingly so that the "font-display" setting doesn't get lost when updating the fonts for the next major release. A line like this should do the trick:
```py
css = css.replace('font-family: "Material', 'font-display: block;\n  font-family: "Material')
```

evnchn:
1. Agreed
2. Agreed. We should think beyond what is written as a suggestion. 
3. Yes 😅 made a typo. To clarify: **set `font-display: block`**
4. Not exactly. Since `font-display: block` is only supposed to be set for `@font-face`. I doubt it would break any modern browser that can tolerate it, but I will see how to work towards a better solution with regards to the fact that "fonts.css is auto-generated using `fetch_google_fonts.py`"

falkoschindler:
> Not exactly.

Oh, you're right. The line should only be added in `@font-face` blocks. This will make a patch a little more complex...

evnchn:
Aparently google fonts updated their format, so at https://github.com/zauberzeug/nicegui/pull/4481/commits/c45a9ccf47c7fe1afc11b4030b10ce2744f34ecf, change the handling behaviour of duplicate file, from throwing error straight away, to check if the file are mismatched then throw error. 

Then, the following regex does the job at https://github.com/zauberzeug/nicegui/pull/4481/commits/9f50a247a23edd02794bd4ea17592e6dfa965bea:

```py
css = re.sub(r'@font-face\s*{\s*font-family:\s*"Material',
             r'@font-face {\n  font-display: block;\n  font-family: "Material', css)
```

evnchn:
See, icons like the menu don't show up when it is loading: 

{A826F9D2-85D1-4940-AD74-BE4B07288517}

falkoschindler (changes requested):
@evnchn Uhm, actually we don't want to update Google fonts before 3.0, but only change the way they are displayed (FOUT vs. FOIT). We just want to update the fetch script so that we don't accidentally revert this PR when updating fonts for 3.0 later.

So this PR shouldn't change any font files and probably doesn't need to change the duplication detection either.

It might feel a bit weird to work on the script, to try it out and then needing to revert most of the changes it produced. But I guess this is the way it needs to be done.

evnchn:
Well, I'm on a trip, so I can't code. 

But what you want should not involve coding. Simply get the manual-edited CSS from previous commit and deploy it for 2.X, then merge this PR and use the auto-generated CSS and font files for 3.X. 

---

It probably would not be feasible to auto-generate CSS in same format of 2.X, due to upstream Google API changes. AFAIK there isn't a "way back machine" for the Google fonts, so we can't go back and fetch the old set. 

---

Also, I personally don't see why we should not update Google fonts for upcoming, say 2.15.0. 


The fonts should be mostly inter-compatible, and staying on an older version of Google fonts in fact adds more confusion as the Google official icon selector and NiceGUI's icons steadily falls more and more out of sync (learnt this while working on the NiceGUI VSCode extension)

evnchn:
Just to clarify: 

1. The commit with the manual-edited CSS is here: https://github.com/zauberzeug/nicegui/pull/4481/commits/75b1427164481d1a4e8eb8dd7eae5e531a867225
2. I see your :eyes:. I hope you are not waiting for me, because I am waiting for you, and that would be a deadlock.

falkoschindler (approved):
I decided to keep the fonts unchanged, because theoretically Google can remove, rename or change icons anytime and it would be very hard to spot a broken icon in your production code.

Let's merge this PR! Any necessary changes for updating fonts can be addressed later in preparation for 3.0.

PR #4487, opened 2025-03-17
"Software rugpull" for multi-worker support

evnchn:
> we create another client with the same id, and, we let it take over of the socket connection. 

---

https://github.com/zauberzeug/nicegui/discussions/1539#discussioncomment-12513394

Everything just works. 

I make a web request to port 8000 using my custom balancer, which remembers I made an HTTP request to which port (say: 8080), and actively refuses to let me talk to such port in the websocket (picking from 8081, 8082, 8083). 

_This is why it is beneficial to make your own balancer @rodja_ 😉

Then, on the page load for async pages, it loads up to the part before await ui.context.client.connected()

{C03E4283-95E3-405E-B62B-6FE200076FF1}

Afterwards, the "Software Rugpull" kicks in, and the page is swapped out with what the other process responds with. 

{A196E690-F044-4D46-A2C2-B6F2C5235E8F}

For sync pages, the old content also briefly flashes before the new content from the client in the rugpull operation is written to the screen. 

{E8DA6F65-5E4A-4B49-84A2-9F6AE1DC1B20}

Afterwards, the interactivity works properly following the new client. 

{46ED10AA-AB8D-4379-B03D-84ECE79C04F2}

---

But, let me also explain how it works. It's fun, really!

When, on handshake, it realizes it just got "cold-called" by a browser whose proclaimed `client_id` in the data doesn't exist, it: 
* Matches the request by immeidately spawning one such client with a matching `client_id`
* Digs through the page route to function dictionary in reverse and calling it
* Sets up the communication just like it would otherwise
* Return true so that the browser won't refresh

This is enabled by:
* making `await ui.context.client.connected()` do absolutely nothing if the client was spawned in response to a cold call
* Browser now also passes the `path` of the page, making it possible to do the reverse lookup in the page route to function dictionary

---

Honestly it is amazing that it works so well as it stands. I can already forsee some code cleanups. 

And also, the args and kwargs of the page are completely neglected as it stands.

### DIFF
--- a/examples/multiple_wokers/main.py
+++ b/examples/multiple_wokers/main.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+import logging
+import os
+from contextlib import asynccontextmanager
+
+from fastapi import FastAPI
+
+from nicegui import ui
+
+logging.basicConfig(level=logging.INFO)
+
+
+@asynccontextmanager
+async def lifespan(_):
+    worker_id = os.getpid()
+    logging.info(f'Worker {worker_id} started')
+    yield
+
+
+@ui.page('/')
+def show():
+    logging.info(f'Handling request in worker {os.getpid()}')
+    ui.label('Hello, FastAPI!')
+
+
+app = FastAPI(lifespan=lifespan)
+ui.run_with(app)
+
+if __name__ == '__main__':
+    print('Please start the app with a "uvicorn" command as shown in the start.sh script')
--- a/examples/multiple_wokers/start.sh
+++ b/examples/multiple_wokers/start.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+cd "$(dirname "$0")" # use path of this example as working directory; enables starting this script from anywhere
+uvicorn main:app --log-level warning --workers 4 --port 8000
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -48,9 +48,14 @@ class Client:
     shared_body_html = ''
     """HTML to be inserted in the <body> of every page template."""
 
-    def __init__(self, page: page, *, request: Optional[Request]) -> None:
+    def __init__(self, page: page, *, request: Optional[Request], force_id: str = None) -> None:
         self.request: Optional[Request] = request
-        self.id = str(uuid.uuid4())
+        if force_id is not None:
+            self.id = force_id
+            self.forced_id = True
+        else:
+            self.id = str(uuid.uuid4())
+            self.forced_id = False
         self.created = time.time()
         self.instances[self.id] = self
 
@@ -135,6 +140,7 @@ class Client:
             **core.app.config.socket_io_js_query_params,
             'client_id': self.id,
             'next_message_id': self.outbox.next_message_id,
+            'path': request.url.path,
         }
         vue_html, vue_styles, vue_scripts, imports, js_imports = generate_resources(prefix, self.elements.values())
         return templates.TemplateResponse(
@@ -177,6 +183,9 @@ class Client:
 
     async def connected(self, timeout: float = 3.0, check_interval: float = 0.1) -> None:
         """Block execution until the client is connected."""
+        if self.forced_id:
+            # By the ID being forced, it means we are doing the "software rugpull" and we don't want to wait for the client
+            return
         self.is_waiting_for_connection = True
         self._waiting_for_connection.set()
         deadline = time.time() + timeout
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -162,7 +162,25 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
 async def _on_handshake(sid: str, data: Dict[str, Any]) -> bool:
     client = Client.instances.get(data['client_id'])
     if not client:
-        return False
+        print("We just got cold-called by a client that doesn't exist")
+        for key, value in Client.page_routes.items():
+            print("Apparently, the requested path is: ", data['path'])
+            if value == data['path']:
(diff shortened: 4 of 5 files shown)


evnchn:
Testing is done via scripts here: 

https://github.com/evnchn/nicegui-multiworker-alpha/commit/75523b971f7596a77197146445afa7ae39c6f15b

evnchn:
Oh yeah, forgot to mention: 

Since the page actually updates with the new content, you could argue that the page we sent with the HTTP using worker A at port 8080 is indeed the "super tiny, generic html page which establishes a websocket connection" that @falkoschindler wanted in https://github.com/zauberzeug/nicegui/discussions/1539#discussioncomment-6890853

Moreover, since the page is essentially re-evaluated in full by worker B in port 8081, I don't really see any data inconsistency issues where worker B works with missing data which only worker A have access to, as it will re-execute and come to (hopefully) the same conclusion as worker A. 

So, this behaviour could be the default behaviour for worker>1, no need `allow_worker_switchover=True` in the @ui.page

evnchn:
Why it is called a "Software Rugpull": 

https://www.youtube.com/watch?v=q5M0TwnkWUM&t=3818s

I can phrase it in the same language as the Phoenix Hyperspace patent as well! 

> At this point the behaviour of `_on_handshake` **diverges greatly** from the traditional behaviour. For, upon the reception of a non-existent client_id in the handshake, the handler **proceeds instead** to create a matching client with the correct id and invokes the page function, establishing communication properly to a browser that is none-the-wiser.

evnchn:
Would want to record a video demo with OBS but (1) internet slow and (2) tomorrow 8am class, so yeah maybe next time. 

Do remind me to do so, though, or unless you've ran the code yourself. 

No checks are ran. Doesn't matter anyways since it still has worker=1 restriction in the ui.run.

rodja:
Thanks for the proof-of-concept @evnchn. I've added a `multiple_workers` example which does not rely on your custom load balancer. That way I could be more sure what exactly is needed for your "rugpull".
While it works like a charm in simple cases, the complexity of integrating it into NiceGUI is huge in my point of view. Just from the top of my head:

- How do we handle downloads (where unique paths might be created on individual workers)?
- Is it safe to access storage files from multiple processes?
- What to do about `app.storage.tab` when a reload or "copy tab" is performed?
- Do we really want to maintain the confusion which arises from not being able to use global state?

What is the overall benefit of this in regards of the costs? The discussion in #1539 already leads into the direction that it's kind of "last resort" to use multiple workers at all.

evnchn:
Thank you for reading my PR 😊

---

For the following language, "I/me" is a worker, "he/him" is another worker, "we" is us NiceGUI devs

* Downloads would not work with our current implementation, there is no way for "me" to inform "him" that I want to let a client to download xxx file and let "him" handle it. We may hack around it by placing the files "I" want to let the client download into a folder, and let "him" read said folder to serve the file 
* Not sure what you mean by "storage files", but everything besides app.storage.browser should be out-of-the-window, because the only thing we agree on between "he" and "I" is the storage_secret. Similar workarounds to the above would be needed for the app.storage.tab/user/general/client. 
* File access should be done via file locks such as `portalocker` for cross-platform file locking, to prevent read-write conflicts. 

---

As to the important question: should we want to embark on the feature...

After thinking about it for a few days, I would believe that it is better to mark this PR as draft, until we can gracefully handle edge cases of this new paradigm, either by disabling features that would not work, or think about other ways of handling it. 

* We may need to branch the development from this point forward, with this branch focused on porting features for multi-worker
* As we work more and more, we can make more and more features available under multi-worker mode
* And, we'll see when this moves out of beta when 90% (that's an objective measure) of features work on this multi-worker mode. 
* All remaining features that will break under this new mode should be documented clearly and throw warning/error (not sure, on the fence for this one)

**More importantly, immediately shipping this PR and disabling the worker>1 restriction in ui.run() is likely to cause mass confusion among those who try out worker>1 and their codebase breaking severely.**

rodja:
Are you sure it's worth the effort? In my point of view, uvicorn workers are just a "hack". Real load balancers line Traefik or Nginx are much better in my opinion.

evnchn:
I went ahead and asked ChatGPT to explain the docker config in https://github.com/zauberzeug/nicegui/tree/main/examples/redis_storage

It seems that it isn't as complicated as what the syntax would suggest, and the template should require no modification for 99% use cases. 

With this in mind and reflecting, we could be potentially compromising a lot of the simplicity in the codebase, just to support uvicorn multi-worker which is basically a load balancer without sticky session support. There has to be another set of test cases just for that, and it doesn't seem worth it. 

---

So, I have converted this into a draft PR, and I'm not sure whether to push this forward. We may end up closing this PR for the time being. 

However, I think we can keep this idea in the back of our heads. The content of this PR may perhaps go into another feature: _making the reconnection process more smooth by skipping a page reload._

evnchn:
Correct me if I am wrong, but

1. Client lose connection
2. Client attempt to reconnect to server
3. Client connect to server, and
  a. client_id still exist on the server, fine, no page reload
  b. client_id no longer exist on the server, enter a sort of panic state and trigger browser page reload

Under this PR, 3b situation would not trigger browser page reload, since - essentially all client_id supplied by the browser is valid after the "rugpull" maneuver. 

Please maybe don't close this PR just yet, I'll do some more testing on this 🤔

evnchn:
More use cases for the login in this PR: Could also refresh page contents without actually reloading the browser, if we simply drop the existing client and force the browser to handshake again. Think: `ui.navigate.soft_reload()`

rodja:
@evnchn can we close this PR? It might be good to create a new feature request for your soft-reload idea.

evnchn:
Sure no problem. I need a few long days to look at it more.

Issue #4182, opened 2025-01-07
ui.leaflet location is reset when ui.plotly is updated


qutang:
### Description

minimal reproducible code:

```python
import plotly.graph_objects as go
from nicegui import ui
from random import random


@ui.page("/")
async def page():
    m = ui.leaflet(zoom=5)
    central_park = m.generic_layer(
        name="polygon",
        args=[
            [
                (40.767809, -73.981249),
                (40.800273, -73.958291),
                (40.797011, -73.949683),
                (40.764704, -73.973741),
            ]
        ],
    )
    await m.initialized()
    bounds = await central_park.run_method("getBounds")
    m.run_map_method("fitBounds", [[bounds["_southWest"], bounds["_northEast"]]])

    fig = go.Figure()
    fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
    plot = ui.plotly(fig).classes("w-full h-40")

    def add_trace():
        fig.add_trace(go.Scatter(x=[1, 2, 3], y=[random(), random(), random()]))
        plot.update()

    ui.button("Add trace", on_click=add_trace)


ui.run()

```

Once I click the button to update the graph, the leaflet map view was reset to the original state before running the map method to fit bounds.

I tested the code on windows platform.

falkoschindler:
Thanks for reporting this issue, @qutang!

It seems like any update to the `ui.plotly` element moves the map back to its original location, even when resizing the window so that the plot is affected. I could boil it down to three lines only:
```py
ui.leaflet(zoom=5)
plot = ui.plotly({})
ui.button("Add trace", on_click=plot.update)
```

falkoschindler:
Similar problem:
```py
@ui.refreshable
def refreshable_ui():
    ...

refreshable_ui()
ui.leaflet(zoom=5)
ui.button('Refresh', on_click=refreshable_ui.refresh)
```
Refreshing the (empty) refreshable UI causes the Leaflet map to jump back to its original location.

nathanWagenbach:
I am having the same issue calling set_source on an interactive_img.
Resets the leaflet map back to the center as well as resetting zoom to inital

Happens on nicegui 2.9.1 and 2.10.2

falkoschindler:
@nathanWagenbach Indeed! Here's an example:
```py
image = ui.interactive_image('https://picsum.photos/id/1/300/200')
ui.button('Change source', on_click=lambda: image.set_source('https://picsum.photos/id/2/300/200'))
[...]

PR #4496, opened 2025-03-18
Add a new method Leaflet.set_view

### LINKED ISSUE #4491: Leaflet `set_center` doesn't work when `set_zoom` comes after it
### Description

Test code:

```python3
from nicegui import ui

@ui.page('/test')
def test_page():

    LONDON = (51.555981, -0.046957)
    ATHENS = (38.139204, 23.612089)

    def set_center():
        leaflet.set_center(ATHENS)

    def set_center_zoom_in():
        leaflet.set_center(ATHENS)
        leaflet.set_zoom(13)

    leaflet = ui.leaflet(LONDON, zoom=6).style('height: 400px;')
    ui.button('Only Set Center', on_click=set_center)
    ui.button('Set center and zoom in', on_click=set_center_zoom_in)


ui.run(reload=False, show=False)
```

**Only Set Center Button:** 
- Expected: Sets the center to Athens
- Actual: Sets the center to Athens ✅

**Set center and zoom in:**
- Expected: Sets the center to Athens and zooms in
- Actual: Doesn't set the center, only zooms into London ❌

https://github.com/user-attachments/assets/ada86619-69e9-41d1-a475-4d10b8388766

amks1:
Added the `set_view` method to set the center and zoom in one operation, because calling `set_zoom` and `set_center` one after the other was running only the latter command.

Resolves #4491

### DIFF
--- a/nicegui/elements/leaflet.py
+++ b/nicegui/elements/leaflet.py
@@ -124,7 +124,7 @@ class Leaflet(Element, component='leaflet.js', default_classes='nicegui-leaflet'
         if self._send_update_on_value_change:
             self.run_map_method('setView', self.center, zoom)
 
-    def set_view(self, center: Tuple[float, float]=None, zoom:int=None) -> None:
+    def set_view(self, center: Optional[Tuple[float, float]]=None, zoom:Optional[int]=None) -> None:
         """Set the map to a new center location, zoom or both in a single operation."""
         if center is None and zoom is None:
             return
--- a/nicegui/elements/leaflet.py
+++ b/nicegui/elements/leaflet.py
@@ -124,6 +124,21 @@ class Leaflet(Element, component='leaflet.js', default_classes='nicegui-leaflet'
         if self._send_update_on_value_change:
             self.run_map_method('setView', self.center, zoom)
 
+    def set_view(self, center: Tuple[float, float]=None, zoom:int=None) -> None:
+        """Set the map to a new center location, zoom or both in a single operation."""
+        if center is None and zoom is None:
+            return
+        if center is not None and self._props['center'] != center:
+            self._props['center'] = center
+        else:
+            center = self.center
+        if zoom is not None and self._props['zoom'] != zoom:
+            self._props['zoom'] = zoom
+        else:
+            zoom = self.zoom
+        if self._send_update_on_value_change:
+            self.run_map_method('setView', center, zoom)
+
     def remove_layer(self, layer: Layer) -> None:
         """Remove a layer from the map."""
         self.layers.remove(layer)
--- a/tests/test_leaflet.py
+++ b/tests/test_leaflet.py
@@ -15,6 +15,13 @@ def test_leaflet(screen: Screen):
     ui.button('Berlin', on_click=lambda: m.set_center((52.520, 13.405)))
     ui.button('London', on_click=lambda: m.set_center((51.505, -0.090)))
 
+    ui.button('Zoom in (set_view)', on_click=lambda: m.set_view(zoom=m.zoom + 1))
+    ui.button('Zoom out (set_view)', on_click=lambda: m.set_view(zoom=m.zoom - 1))
+    ui.button('Berlin (set_view)', on_click=lambda: m.set_view((52.520, 13.405)))
+    ui.button('London (set_view)', on_click=lambda: m.set_view((51.505, -0.090)))
+    ui.button('Berlin Zoom In', on_click=lambda: m.set_view((52.520, 13.405), m.zoom + 1))
+    ui.button('London Zoom Out', on_click=lambda: m.set_view((51.505, -0.090), m.zoom - 1))
+
     screen.open('/')
     assert screen.find_all_by_class('leaflet-pane')
     assert screen.find_all_by_class('leaflet-control-container')
@@ -36,3 +43,27 @@ def test_leaflet(screen: Screen):
 
     screen.click('London')
     screen.should_contain('Center: 51.505, -0.090')
+
+    screen.click('Zoom in (set_view)')
+    screen.should_contain('Zoom: 14')
+    screen.should_contain('Center: 51.505, -0.090')
+
+    screen.click('Zoom out (set_view)')
+    screen.should_contain('Zoom: 13')
+    screen.should_contain('Center: 51.505, -0.090')
+
+    screen.click('Berlin (set_view)')
+    screen.should_contain('Center: 52.520, 13.405')
+    screen.should_contain('Zoom: 13')
+
+    screen.click('London (set_view)')
+    screen.should_contain('Center: 51.505, -0.090')
+    screen.should_contain('Zoom: 13')
+
+    screen.click('Berlin Zoom In')
+    screen.should_contain('Center: 52.520, 13.405')
+    screen.should_contain('Zoom: 14')
+
+    screen.click('London Zoom Out')
+    screen.should_contain('Center: 51.505, -0.090')
+    screen.should_contain('Zoom: 13')


evnchn:
Wait. I think the issue is deeper than what you think. 

Checking both `set_zoom` and `set_center`, what happens is: 

* `set_zoom` lets the library user set zoom, and use `self.center` for center, but the zoom is not commited to `self.zoom` right away, but rather only at `_handle_zoomend`. 
* `set_center` lets the library user set center, and use `self.zoom` for zoom, but the center is not commited to `self.center` right away, but rather only at `_handle_moveend`

This means that `set_zoom` and `set_center` are fundamentally illsuited to be called together, which we may consider changing instead of adding another function which is basically set_zoom and set_center bundled together in a trenchcoat to adhere to DRY principle. Or, for that, we can implement `set_zoom` and `set_center` via calling `set_view`. 

Also, your `set_view` commits to self.zoom and self.center right away. There may be some special reasons `set_zoom` and `set_center` doesn't do that, and ther may be some consequences of breaking the pattern. Maybe ask around for that first.

falkoschindler:
Thank you for contributing this pull request, @amks1!

The center and zoom methods of `ui.leaflet` have indeed quite some history:
- #3035
- #3125
- #4061
- #4121
- #4182
- #4265

At least three times we tried to fix some weird behavior related to changing the map's center or zoom. So before simply adding another method `set_view`, I'd like to understand what is actually going on. Even if we would add a new method, the two existing ones would remain broken if used in combination.

I suggest to keep this PR here for reference and continue discussing the problem over at #4491.

amks1:
> Checking both `set_zoom` and `set_center`, what happens is:
> 
> * `set_zoom` lets the library user set zoom, and use `self.center` for center, but the zoom is not commited to `self.zoom` right away, but rather only at `_handle_zoomend`.
> * `set_center` lets the library user set center, and use `self.zoom` for zoom, but the center is not commited to `self.center` right away, but rather only at `_handle_moveend`
>
> This means that `set_zoom` and `set_center` are fundamentally illsuited to be called together, which we may consider changing instead of adding another function which is basically set_zoom and set_center bundled together in a trenchcoat to adhere to DRY principle. Or, for that, we can implement `set_zoom` and `set_center` via calling `set_view`.

I did spend some time trying to figure out whether I could achieve this without adding another function, but didn't arrive at anything. I suspect it's something "deeper" in the code in the way javascript code is sequenced to execute on the client. The second command always overrides the first. 
Personally I feel letting users choose to set one or both of these parameters using the same function is more intuitive, but of course I didn't want to break anybody's code by removing existing methods.

> Also, your `set_view` commits to self.zoom and self.center right away. There may be some special reasons `set_zoom` and `set_center` doesn't do that, and ther may be some consequences of breaking the pattern. Maybe ask around for that first.

My code doesn't commit `self.zoom` or `self.center`, only `self._props['zoom']` and `self._props['center']`. Like you described above, `self.zoom` and `self.center` still get updated in the event handlers `_handle_zoomend` and `_handle_moveend`. I've confirmed that the `self.zoom` and `self.center` get set correctly after `set_view` (but I admit I don't know why this is being done in this roundabout way).

image

Code:
```python3
@ui.page('/test')
def test_page():

    LONDON = (51.555981, -0.046957)
    ATHENS = (38.139204, 23.612089)

    def set_marker_zoom_in():
        marker = leaflet.marker(latlng=ATHENS)
        leaflet.set_center(ATHENS)
        leaflet.set_zoom(13)

    def set_marker():
        marker = leaflet.marker(latlng=ATHENS)
        leaflet.set_center(ATHENS)

    def set_center():
        leaflet.set_center(ATHENS)

    def set_center_zoom_in():
        # leaflet.set_center(ATHENS)
        # leaflet.set_zoom(13)
        leaflet.set_view(ATHENS, 13)

    leaflet = ui.leaflet(LONDON, zoom=6).style('height: 400px;')
    ui.button('Set Marker and Zoom In', on_click=set_marker_zoom_in)
    ui.button('Set Marker', on_click=set_marker)
    ui.button('Only Set Center', on_click=set_center)
    ui.button('Set center and zoom in', on_click=set_center_zoom_in)
    ui.button('Show center and zoom', on_click=lambda: ui.notify(f'Center: {leaflet.center}, Zoom: {leaflet.zoom}'))


ui.run(reload=False, show=False)
```

evnchn:
@amks1 Sorry to keep you waiting. It's quite mind-bending, but after reviewing the code some more, I think I got it. 

`self.center` and `self.zoom`: Reported from Leaflet, after the zoom and center operation is completed. Check the below: 

https://github.com/zauberzeug/nicegui/blob/847c4abcbe6b6822f142bc84de1a6e25d47b136b/nicegui/elements/leaflet.py#L96-L104

`self._props['center']` and `self._props['zoom']`: Updated immediately on `set_center` and `set_zoom` methods. Check the below: 

https://github.com/zauberzeug/nicegui/blob/847c4abcbe6b6822f142bc84de1a6e25d47b136b/nicegui/elements/leaflet.py#L111-L125

Now, I think you will agree, that I can change the code slightly for easier readability: 

```py
def set_center(self, center: Tuple[float, float]) -> None:
    """Set the center location of the map."""
    if self._props['center'] == center:
        return
    self._props['center'] = center
    if self._send_update_on_value_change:
        self.run_map_method('setView', self._props['center'], self.zoom) # center = self._props['center']

def set_zoom(self, zoom: int) -> None:
    """Set the zoom level of the map."""
    if self._props['zoom'] == zoom:
        return
    self._props['zoom'] = zoom
    if self._send_update_on_value_change:
        self.run_map_method('setView', self.center, self._props['zoom']) # zoom = self._props['zoom']
```

This makes it obvious: When you do two consecutive `set_zoom` and `set_center`, since the reported value from Leaflet has not returned, the second call won't actually use the first call's zoom/center, instead using whatever's in `self.center` / `self.zoom`, which is out-of-date. 

So, indeed, there is no way with existing functions that a zoom-and-move maneuver can be done in one fell swoop. 

Do @ me and correct me, if you think I am wrong, though. 

---

Given the above, and how I had to modify the code to make it readable, I take issue with how you do `center = self.center`. It makes it rather confusing, since `center` is both the input and the output. I need to think deeply about it, and see if I can make sense of it. Meanwhile, you can try and refactor your new function such that the code flow is more readable and explainable.

evnchn:
Let's see if this logic is more nice: 

```py
def set_view(self, *, center: Optional[Tuple[float, float]] = None, zoom: Optional[int] = None) -> None:
    """Set the view of the map.

    :param center: new center location (latitude/longitude)
    :param zoom: new zoom level
    """
    value_changed = False
    if center is not None and self._props['center'] != center:
        self._props['center'] = center
        value_changed = True
    if zoom is not None and self._props['zoom'] != zoom:
        self._props['zoom'] = zoom
        value_changed = True
    if value_changed and self._send_update_on_value_change:
        self.run_map_method('setView', center or self.center, zoom or self.zoom)
```

Key points: 

1. Your old implementation would still trigger `self.run_map_method` if I pass in `center = self._props['center'], zoom = self._props['zoom']`. Mine, however, `value_changed` would remain `False`, skipping `self.run_map_method`, inline with the spirit of the two previous functions. 
2. Your old implementation sets the function input variables, which isn't quite readable. Mine, however, does `self.run_map_method('setView', center or self.center, zoom or self.zoom)`, which captures the spirit of "use `center` if passed in, else `self.center` if it was not. 
3. Namely due to point 1, if my logic thinking is right, `set_zoom` and `set_center` can call my newly defined `set_view` method, and the behaviour would be idempotent to the old behaviour, keeping backwards compatibility. 

@falkoschindler code review time. See if this is the better solution.

evnchn:
I think I have a trick in mind which would allow calling set_zoom and set_center one after the other to simply work

This negates the need for this `set_view`, which also eliminates potential user error where they don't know that this new function exists and don't use them. 

Let me cook.

falkoschindler:
@evnchn That would be great! Apart from being busy with lots of other things, the suboptimal solution with potential user errors is one reason why I haven't reviewed and merged this PR yet. I thought there must be a better way, but couldn't find one.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5453

falkoschindler:
I'm closing this in favor of PR #5453, which solves the original problem without adding a new API.
Thank you anyway for your effort, @amks1!

PR #4507, opened 2025-03-20
modulepreload the scripts needed for faster page load

evnchn:
https://github.com/zauberzeug/nicegui/discussions/4505#discussioncomment-12562699

Before: 
image

After: 
image

It's slightly better. _It ain't much but it's honest work._ 

When does this matter? 
* Initial page load
* Loading new and unused elements like `ui.input()`

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -136,13 +136,12 @@ class Client:
             'client_id': self.id,
             'next_message_id': self.outbox.next_message_id,
         }
-        vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls = generate_resources(
-            prefix, self.elements.values())
+        vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls = \
+            generate_resources(prefix, self.elements.values())
         return templates.TemplateResponse(
             request=request,
             name='index.html',
             context={
-                'js_imports_urls': js_imports_urls,
                 'request': request,
                 'version': __version__,
                 'elements': elements.replace('&', '&amp;')
@@ -155,6 +154,7 @@ class Client:
                 'vue_scripts': '\n'.join(vue_scripts),
                 'imports': json.dumps(imports),
                 'js_imports': '\n'.join(js_imports),
+                'js_imports_urls': js_imports_urls,
                 'quasar_config': json.dumps(core.app.config.quasar_config),
                 'title': self.resolve_title(),
                 'viewport': self.page.resolve_viewport(),
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -136,11 +136,13 @@ class Client:
             'client_id': self.id,
             'next_message_id': self.outbox.next_message_id,
         }
-        vue_html, vue_styles, vue_scripts, imports, js_imports = generate_resources(prefix, self.elements.values())
+        vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls = generate_resources(
+            prefix, self.elements.values())
         return templates.TemplateResponse(
             request=request,
             name='index.html',
             context={
+                'js_imports_urls': js_imports_urls,
                 'request': request,
                 'version': __version__,
                 'elements': elements.replace('&', '&amp;')
--- a/nicegui/dependencies.py
+++ b/nicegui/dependencies.py
@@ -130,6 +130,7 @@ def generate_resources(prefix: str, elements: Iterable[Element]) -> Tuple[List[s
                                                                           List[str],
                                                                           List[str],
                                                                           Dict[str, str],
+                                                                          List[str],
                                                                           List[str]]:
     """Generate the resources required by the elements to be sent to the client."""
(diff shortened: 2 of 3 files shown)


falkoschindler:

I have to admit, your throughput is insane! I'm blinking and my inbox is full again. Unfortunately this means it's getting harder and harder to follow every discussion in detail. Therefore it would be very helpful if you could summarize PRs briefly so that they can be reviewed without reading the preceding discussion (again). Discussions can get long and messy. But a PR is a good opportunity to bring everyone up to speed.

Regarding this PR: You seem to be
1. moving the importmap to the HTML head, and
2. preloading all JS modules.

What is the impact of defining the importmap in the head vs. body? I'm a bit worried about side-effects.
What exactly is the benefit/downside of preloading _all_ modules. Doesn't it delay page rendering?
And what do you mean with "It's slightly better"? Does it fix the original problem? What is the problem you're trying to solve anyway?

evnchn:
Sorry for my oversight. This PR can do with a TL-DR since otherwise everyone have to read #4505 to know what's going on. 

**Brief description:** 
This PR is about moving the importmap to the HTML head and preloading all JS modules via "modulepreload"

**Benefits:** 
This enables the browser to [preemptively fetch a module script, parse and compile it, and store it in the document's module map for later execution](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/modulepreload)

More clearly, the browser essentially "begins work" despite it is only done handling the `<head>`, which makes the scripts gets fetched, parsed, and compiled earlier (measuring from the moment page load begins). 

**Scale of benefits**
This only benefits the loading of the page when the network is sufficiently terrible in terms of latency and download bandwidth. We are squeezing 0.1 seconds in a 2-second load under 3G, and it is a blink of an eye (maybe -1 frame) if you don't throttle. But, there _is_ a performance increase. 

---

About your concerns: 

**Moving the import map earlier may have side-effects?**
MDN claims only that an import map [must be declared and processed before any <script> elements that import modules using specifiers declared in the map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap). There are no restrictions as to how early we can put it. 

**Will it slow down page rendering?**
If I understand correctly, no. By us informing the browser earlier, we in fact let the browser do work earlier, speeding up page rendering. Either way, when we reach the body, when the browser haven't fetched, parsed, and compiled the scripts, it will say "hold up" and let those imports be processed before proceeding to the `createApp`. We are just reducing the chance of this occuring.

evnchn:
P.S.: I follow and comment only on 1 GitHub repository so far, and that's NiceGUI, and I am otherwise idle. I hope I don't burn you and Rodja out. 

I'll go on a weekend trip starting tomorrow and returning at the middle-of-the-week, so there wouldn't be (I hope) any new PRs, so you can catch up for a bit.

evnchn:
Anyone got a browser (or know how to get one) which doesn't support ES modules? Want to know how this PR will play out with the ES-module-shim.js

evnchn:
**TL-DR: No adverse side effects found despite on ancient browser. Should be able to merge safely.** 

@falkoschindler do you have other concerns on side-effects? I can't think of any so far...

Also, I want to clatify: It does not preload _all_ modules. Only those which are actually in use by the page, and will appear in the main `<script>` tag in the `<body>`. It would be absurd otherwise. 

---

Went ahead and tested the PR on Pale Moon, a browser based on Firefox. Forked pre-Quantum, it is still using the old Gecko engine. Apparently that means it still does not support ES modules. That's a good thing for me to test compatibility. 

```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="UTF-8">
    </head>
<body>

<!-- Comment out the line below and the react will not be console.log for Pale Moon -->
<script async src="https://ga.jspm.io/npm:es-module-shims@2.0.10/dist/es-module-shims.js"></script>

<!-- https://generator.jspm.io/#U2NhYGBkDM0rySzJSU1hKEpNTC5xMLTQM9Az0C1K1jMAAKFS5w0gAA -->
<script type="importmap">
{
  "imports": {
    "react": "https://ga.jspm.io/npm:react@18.0.0-rc.0/index.js"
  },
  "scopes": {
    "https://ga.jspm.io/npm:react@18.0.0-rc.0/": {
      "object-assign": "https://ga.jspm.io/npm:object-assign@4.1.1/index.js"
    }
  }
}
</script>

<script type="module">
  import react from 'react';
  console.log(react);
</script>
</body>
</html>
```

image

Then, I tried it with the PR: 

{4C128A62-B48C-41BF-86A3-852BB740074D}


Yep, it is polyfilled and the aggrid loads, despite modulepreload and shifting importmap to the front.

falkoschindler (approved):
Ok, I think we can safely merge this PR.
I tested with a simple `ui.aggrid({})` and wondered if we could also preload libraries like ag-grid-community.min.js, not only the component aggrid.js. But it doesn't seem to be trivial, so we defer it to another PR. 👍🏻

PR #4510, opened 2025-03-20
Fix typing for ui.refreshable/ui.refreshable_method

drkspace:
If you try to use an `async` refreshable method, the fact that it is `async` is lost by the type checker.

For example (A bigger, full example is below)
```python
from nicegui import ui
from typing_extensions import reveal_type

@ui.refreshable
def sync() -> None:
    return

@ui.refreshable
async def a_sync() -> None:
    return

reveal_type(sync) # refreshable[(), None]
reveal_type(a_sync) # refreshable[(), None]
```

The typing for the input function is `Callable[_P, Union[_T, Awaitable[_T]]]` and this PR removes the `Awaitable[_T]` from the union. The type checker will know the return type of the method is `Coroutine` and not lose that information.

---

Full Example:
```python
from nicegui import ui
from typing_extensions import reveal_type

@ui.refreshable
def sync() -> None:
    return

@ui.refreshable
async def a_sync() -> None:
    return

class Foo:

    @ui.refreshable_method
    def sync(self) -> None:
        return

    @ui.refreshable_method
    async def a_sync(self) -> None:
        return

reveal_type(sync)
reveal_type(a_sync)

async def func():

    foo = Foo()

    sync()
    sync.refresh()

    await a_sync()
    a_sync.refresh()

    foo.sync()
    foo.sync.refresh()

    await foo.a_sync()
    foo.a_sync.refresh()
```

In the current version of nicegui, you get type errors on the `await` lines and both the reveal types are `refreshable[(), None]`. In my version, you get no type errors and the second reveal type is `refreshable[(), Coroutine[Any, Any, None]]`

---

I had to add a `type: ignore` on `refreshable.py:45` . I tried converting `is_coroutine_function` to a `TypeGuard`, but some typing information (the actual return value) is lost, so you'll have to add `type: ignore` or `cast` elsewhere. _I think_ this can be fixed by narrowing the type of the input of `is_coroutine_function` from `Any` to `Callable` with some overloads, but that would cause type errors elsewhere. If someone does figure out a "better" typing of `is_coroutine_function`,  then the `assert isinstance(result, Awaitable)`s can go away (I assume they were there for the type checker).

### DIFF
--- a/nicegui/functions/refreshable.py
+++ b/nicegui/functions/refreshable.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Any, Awaitable, Callable, ClassVar, Dict, Generic, List, Optional, Tuple, TypeVar, Union, cast
+from typing import Any, Awaitable, Callable, ClassVar, Dict, Generic, List, Optional, Tuple, TypeVar, cast
 
 from typing_extensions import Concatenate, ParamSpec, Self
 
@@ -42,7 +42,7 @@ class RefreshableTarget:
                         result = func(self.instance, *self.args, **self.kwargs)
                     assert isinstance(result, Awaitable)
                     return await result
-            return wait_for_result() # type: ignore
+            return wait_for_result()  # type: ignore
         else:
             with self.container:
                 if self.instance is None:
--- a/nicegui/functions/refreshable.py
+++ b/nicegui/functions/refreshable.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Any, Awaitable, Callable, ClassVar, Dict, Generic, List, Optional, Tuple, TypeVar, Union, cast
+from typing import Any, Callable, ClassVar, Dict, Generic, List, Optional, Tuple, TypeVar, Union, cast
 
 from typing_extensions import Concatenate, ParamSpec, Self
 
@@ -28,7 +28,7 @@ class RefreshableTarget:
     locals: List[Any] = field(default_factory=list)
     next_index: int = 0
 
-    def run(self, func: Callable[..., Union[_T, Awaitable[_T]]]) -> Union[_T, Awaitable[_T]]:
+    def run(self, func: Callable[..., _T]]) -> _T:
         """Run the function and return the result."""
         RefreshableTarget.current_target = self
         self.next_index = 0
@@ -42,7 +42,7 @@ class RefreshableTarget:
                         result = func(self.instance, *self.args, **self.kwargs)
                     assert isinstance(result, Awaitable)
                     return await result
-            return wait_for_result()
+            return wait_for_result() # type: ignore
         else:
             with self.container:
                 if self.instance is None:
@@ -57,7 +57,7 @@ class RefreshableContainer(Element, component='refreshable.js'):
 
 class refreshable(Generic[_P, _T]):
 
-    def __init__(self, func: Callable[_P, Union[_T, Awaitable[_T]]]) -> None:
+    def __init__(self, func: Callable[_P, _T]]) -> None:
         """Refreshable UI functions
 
         The ``@ui.refreshable`` decorator allows you to create functions that have a ``refresh`` method.
@@ -83,7 +83,7 @@ class refreshable(Generic[_P, _T]):
             return refresh
         return attribute
 
-    def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> Union[_T, Awaitable[_T]]:
+    def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _T:
         self.prune()
         target = RefreshableTarget(container=RefreshableContainer(), refreshable=self, instance=self.instance,
                                    args=args, kwargs=kwargs)
@@ -133,7 +133,7 @@ class refreshable(Generic[_P, _T]):
 
 class refreshable_method(Generic[_S, _P, _T], refreshable[_P, _T]):
 
-    def __init__(self, func: Callable[Concatenate[_S, _P], Union[_T, Awaitable[_T]]]) -> None:
+    def __init__(self, func: Callable[Concatenate[_S, _P], _T]) -> None:
         """Refreshable UI methods
 
         The `@ui.refreshable_method` decorator allows you to create methods that have a `refresh` method.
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
Thanks for the pull request, @drkspace!

The diff looks nice (only removing code...), but I'm still struggling to understand what is going on.

1. When running your examples, `revealtype` only shows "Runtime type is 'refreshable'" _for both cases_ and on both branches, main and your PR. When running mypy, I'm seeing
    ```
    test.py:28: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], None]"
    test.py:29: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], typing.Coroutine[Any, Any, None]]"
    ```
    on this PR, and
    ```
    test.py:28: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], Any]"
    test.py:29: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], Never]"
    ```
    on main (after adding `-> None` to all functions and methods). So I guess yours is more correct?

2. On main mypy complains about decorating async functions and methods as refreshable. But I don't get "get type errors on the `await` lines". Maybe it has something to do with the Python and mypy versions we're using?

3. Can you explain _why_ removing `Awaitable` from the `Union` helps to fix the problem you're experiencing? Was it utterly wrong in the first place? Or are there side-effects to be expected when removing it?

drkspace:
I was running on 3.12 with pyright. 

After testing on different python versions with mypy and pyright (I'll put what exactly I did at the end), I always got that the main branch errored and my branch did not. For example, here's the output of my script for python 3.11

```
=========main branch===============

[notice] A new release of pip is available: 24.0 -> 25.0.1
[notice] To update, run: pip install --upgrade pip
mypy:
tmp/tmp.oFMGsNefS6/test.py:8: error: Argument 1 to "refreshable" has incompatible type "Callable[[], Coroutine[Any, Any, None]]"; expected "Callable[[], Awaitable[Never]]"  [arg-type]
tmp/tmp.oFMGsNefS6/test.py:18: error: Argument 1 to "refreshable_method" has incompatible type "Callable[[Foo], Coroutine[Any, Any, None]]"; expected "Callable[[Foo], Awaitable[Never]]"  [arg-type]
tmp/tmp.oFMGsNefS6/test.py:18: note: This is likely because "a_sync of Foo" has named arguments: "self". Consider marking them positional-only
tmp/tmp.oFMGsNefS6/test.py:22: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], None]"
tmp/tmp.oFMGsNefS6/test.py:23: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], Never]"
Found 2 errors in 1 file (checked 1 source file)
pyright:
 * Install prebuilt node (23.10.0) ..... done.
/tmp/tmp.oFMGsNefS6/test.py
  /tmp/tmp.oFMGsNefS6/test.py:22:13 - information: Type of "sync" is "refreshable[(), None]"
  /tmp/tmp.oFMGsNefS6/test.py:23:13 - information: Type of "a_sync" is "refreshable[(), None]"
  /tmp/tmp.oFMGsNefS6/test.py:32:11 - error: "None" is not awaitable
    "None" is incompatible with protocol "Awaitable[_T_co@Awaitable]"
      "__await__" is not present (reportGeneralTypeIssues)
  /tmp/tmp.oFMGsNefS6/test.py:38:11 - error: "None" is not awaitable
    "None" is incompatible with protocol "Awaitable[_T_co@Awaitable]"
      "__await__" is not present (reportGeneralTypeIssues)
2 errors, 0 warnings, 2 informations 
=========drkspace branch===============

[notice] A new release of pip is available: 24.0 -> 25.0.1
[notice] To update, run: pip install --upgrade pip
mypy:
tmp/tmp.oFMGsNefS6/test.py:22: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], None]"
tmp/tmp.oFMGsNefS6/test.py:23: note: Revealed type is "nicegui.functions.refreshable.refreshable[[], typing.Coroutine[Any, Any, None]]"
Success: no issues found in 1 source file
pyright:
/tmp/tmp.oFMGsNefS6/test.py
  /tmp/tmp.oFMGsNefS6/test.py:22:13 - information: Type of "sync" is "refreshable[(), None]"
  /tmp/tmp.oFMGsNefS6/test.py:23:13 - information: Type of "a_sync" is "refreshable[(), CoroutineType[Any, Any, None]]"
0 errors, 0 warnings, 2 informations 
```

I think the 2 different error locations are 2 sides of the same "error". Making the type just `_T` seems to have fixed it on both type checkers.

> Can you explain why removing Awaitable from the Union helps to fix the problem you're experiencing? Was it utterly wrong in the first place? Or are there side-effects to be expected when removing it?

_I think_ it fixes the problem because, somewhere along the line, it recognizes the type as `Awaitable[_T]` but then it drops the `Awaitable` for some reason. I do not know if this is a bug in the type checking systems or expected behavior (I'm having difficulty creating a small example of this to make an issue on mypy/pyright). 

I don't think this will cause any side effects because `_T` "contains" `Awaitable[_T]`, so there is no information that `Awaitable[_T]` has that `_T` cannot also have/rederive (with something like `is_coroutine_function`).

---
## Running the tests

1. I cloned the main branch and my branch in the same parent folder (I named my folder `ng_drk` so they wouldn't collide)
2. Created the following script

```bash
#!/usr/bin/env bash

DATA_DIR=/data

python -m venv venv_ng
python -m venv venv_ng_drk

tmpd=$(mktemp -d)
cp $DATA_DIR/test.py $tmpd/.

. venv_ng/bin/activate

echo "=========main branch==============="

pip install $DATA_DIR/nicegui mypy pyright >/dev/null

echo "mypy:"
mypy $tmpd/test.py
echo "pyright:"
pyright $tmpd/test.py

deactivate

echo "=========drkspace branch==============="

. venv_ng_drk/bin/activate
pip install $DATA_DIR/ng_drk mypy pyright >/dev/null
echo "mypy:"
mypy $tmpd/test.py
echo "pyright:"
pyright $tmpd/test.py

deactivate
```

3. Copied my full example above into that containing folder as `test.py`. Note that I added `-> None` to the functions since mypy wasn't able to deduce they returned `None`.

Here's the folder structure as I had it

```bash
$ tree -L 1
.
├── ng_drk
├── nicegui
├── run_test.sh
└── test.py
```

5. Ran it with docker like `docker run -t -v .:/data python:3.8 /data/run_test.sh` (changing the python version each run)

falkoschindler (approved):
Thanks for the clarification, @drkspace!
Using pyright vs. mypy explains the different error messages. Let's merge! 🚀

PR #4520, opened 2025-03-23
Add example for `Xterm.js` connected to Bash

paco-sevilla:
The discussion about integrating [Xterm.js](https://github.com/xtermjs/xterm.js) into `nicegui` (https://github.com/zauberzeug/nicegui/discussions/1846) seems to be stale...
However, I'd find it great to have `Xterm.js` in `nicegui`!

This PR just adds an example, but **I would be more than happy to close this PR and work on a proper integration** (including a "bell" event, a resizing event, addons like [@xterm/addon-web-links](https://github.com/xtermjs/xterm.js/tree/master/addons/addon-web-links), etc.).

## Use-cases for Xterm in nicegui:
* Provide a Terminal & Shell for native (electron-like) applications
* Provide a access to the Shell of a server through a web application, e.g. for a robot (this is the example in this PR, which has obvious security risks!)
* Display the output of a `subprocess` containing ANSI escape codes (this is what I'm mostly interested in)

## Example in this PR
The example integrates `Xterm.js` as a new element called `Terminal`, which provides a simple interface for reading and writing `bytes` from/to it.
This new element is then used in `main.py`, connecting in to `/bin/bash` running in a `pty.fork()` process.

The page looks as follows:
image

### DIFF
--- a/examples/xterm/.gitignore
+++ b/examples/xterm/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+package-lock.json
--- a/examples/xterm/README.md
+++ b/examples/xterm/README.md
@@ -1,0 +1,27 @@
+# Use Xterm.js as Third-party Dependency to run Bash
+
+This example demonstrates how to use the [Xterm.js](https://github.com/xtermjs/xterm.js) node module as dependency in a
+NiceGUI app. The app starts Bash in a pseudo-terminal (pty) and connects it to the Xterm.js element.
+
+In package.json, the `@xterm/xterm` module is listed as a dependency, while `terminal.js` and `terminal.py` define the
+new UI element which is then used in `main.py`.
+
+To run this example:
+
+1. First, install the third-party node modules (assuming you have NPM installed):
+
+   ```bash
+   npm install
+   ```
+
+   This will create a node_modules directory containing the `@xterm/xterm` module.
+   To install `npm`, you can use a node version manager like [nvm](https://github.com/nvm-sh/nvm).
+
+2. Now you can run the app as usual:
+
+   ```bash
+   python3 main.py
+   ```
+
+> [!WARNING] 
+> This example gives the clients full access to the server through Bash. Use with caution!
--- a/examples/xterm/main.py
+++ b/examples/xterm/main.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+"""Example for integrating the `Terminal` component with `Bash` running in a pty.
+
+WARNING: This example gives the clients full access to the server through Bash. Use with caution!
+"""
+
+import asyncio
+import logging
+import os
+import pty
+import signal
+
+from terminal import Terminal
+
+from nicegui import app, context, ui
+from nicegui.logging import log
+
+# Firstly, we need to make the `xterm.css` file available to the application
+XTERM_CSS = app.add_static_file(local_file='node_modules/@xterm/xterm/css/xterm.css')
+
+
+# We create the terminals in a `ui.page`, so that each client gets its own terminal
+@ui.page("/")
+def _page():
+    # We need to add the `xterm.css` file to the head of the page
+    ui.add_head_html(f'<link href="{XTERM_CSS}" rel="stylesheet" type="text/css"/>')
+
+    # We now create the terminal (front-end) element for the client
+    client = context.client
+    ui.label(f'Terminal ({client.id})').style('font-weight: bold;')
+    terminal = Terminal()
+
+    async def run_bash():
+        # At this point, we need to create a new pseudo-terminal (pty) fork of the process
+        pty_pid, pty_fd = pty.fork()
+        if pty_pid == pty.CHILD:
+            # The child process of the fork gets replaced with 'bash'
+            os.execv('/bin/bash', ('bash',))
+        log.info('Terminal opened: %s', client.id)
+
+        def pty_to_terminal():
+            try:
+                data = os.read(pty_fd, 1024)
+            except OSError:
+                # There was an error reading the pty; probably bash was exited. Let's stop reading from it.
+                log.info('Stopping reading from pty: %s', client.id)
+                loop.remove_reader(pty_fd)
+            else:
+                terminal.write(data)
+
+        def terminal_to_pty(data: bytes):
+            try:
+                os.write(pty_fd, data)
+            except OSError:
+                # There was an error writing to the pty; probably bash was exited. Nothing to do.
+                pass
+
+        # Now we can connect the pty to the terminal
+        loop = asyncio.get_event_loop()
+        loop.add_reader(pty_fd, pty_to_terminal)
+        terminal.on_data(terminal_to_pty)
+
+        def kill_bash():
+            os.close(pty_fd)
+            os.kill(pty_pid, signal.SIGKILL)
+            log.info('Terminal closed: %s', client.id)
+
+        # Make sure to kill the bash process when the client disconnects
+        client.on_disconnect(kill_bash)
+        await client.disconnected()
+
+    client.safe_invoke(run_bash())
+
+
+if __name__ in {"__main__", "__mp_main__"}:
+    logging.basicConfig(level=logging.INFO)
+    ui.run()
--- a/examples/xterm/package.json
+++ b/examples/xterm/package.json
@@ -1,0 +1,5 @@
+{
+  "dependencies": {
+    "@xterm/xterm": "^5.5.0"
+  }
+}
--- a/examples/xterm/terminal.js
+++ b/examples/xterm/terminal.js
@@ -1,0 +1,19 @@
+import "xterm";
+
+export default {
+  template: "<div></div>",
+  props: {
+    options: Array,
+  },
+  async mounted() {
+    this.terminal = new Terminal();
+    this.terminal.open(this.$el)
+    
+    this.terminal.onData((e) => this.$emit(`data`, e));
+  },
+  methods: {
+    write(data) {
+      return this.terminal.write(data);
+    },
+  },
+};
--- a/examples/xterm/terminal.py
+++ b/examples/xterm/terminal.py
@@ -1,0 +1,29 @@
+from typing import Callable, Dict, Optional, Union
+
+from nicegui import ui
+
+
+class Terminal(ui.element,
+               component='terminal.js',
+               dependencies=['node_modules/@xterm/xterm/lib/xterm.js']):
+
+    def __init__(self, options: Optional[Dict] = None) -> None:
+        """Terminal
+
+        An element that integrates `xterm` to emulate a terminal.
+        Note: This element provides only a front-end component without an underlying shell.
+        """
+        super().__init__()
+        self._props['options'] = options or {}
+
+    def on_data(self, callback: Callable[[bytes], None]) -> None:
+        """Add a callback to be invoked when data is received from the terminal."""
+        self.on('data', lambda event: callback(str(event.args).encode()))
+
+    def write(self, data: Union[bytes, str]) -> None:
+        """Write data to the terminal."""
+        if isinstance(data, bytes):
+            # Xterm.js accepts an `Uint8Array`, which we can get by converting `bytes` to a `list`
+            self.run_method('write', list(data))
+        else:
+            self.run_method('write', data)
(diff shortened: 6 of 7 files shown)


evnchn:
Xterm.js is quite a well-used project at this point. 

I have seen it used in Proxmox, as well as various third-party router plugins which provide SSH. 

So, I can stand behind including it for NiceGUI. 

---

If I understand correctly, now we need to do `npm install` to grab xterm right? 

Check out npm.json and npm.py to see how other elements officially in NiceGUI currently fetches the dependencies from NPM (though that's up for change later, but should stick until rest of 2.X)

paco-sevilla:
@evnchn Thanks for your comment.
I'm aware of those files (`npm.py` and `npm.json`). As I wrote above and in https://github.com/zauberzeug/nicegui/discussions/1846, if the authors of NiceGUI think that adding `Xterm.js` is valuable, I can work on a proper integration: add the module to the npm files, move `terminal.py` to `nicegui/elements`, add tests, ...

This PR just adds an example, very similar to the one here: https://github.com/zauberzeug/nicegui/tree/main/examples/signature_pad

rodja:
I think xterm.js would be a wonderful (and very powerful) NiceGUI element. But let me discuss it with @falkoschindler first. Might be at the end of the week.

falkoschindler:
Thanks for the pull request, @paco-sevilla!

@rodja and I agreed that it would be great to integrate Xterm.js in NiceGUI. For such a useful UI element, it only adds ~300KB of minified JS and a few lines of CSS. (I haven't checked the addons though.)

Instead of `ui.terminal`, we would prefer the name `ui.xterm`. For non-Quasar elements we tend to use the original brand name to avoid confusion with similar elements - like `ui.aggrid` was originally called `ui.table`, which is now a Quasar table...

So we'd appreciate it very much if you could convert the example into an integrated UI element, either on this PR or a new one. As @evnchn [pointed out](https://github.com/zauberzeug/nicegui/pull/4520#issuecomment-2745971673), we currently manage our dependencies with npm.json and npm.py, which might be a bit tricky to use. Let us know if you need help with anything!

KiddoV:
+1

evnchn:
Add this to `npm.json`

```json
  "xterm": {
    "package": "@xterm/xterm",
    "destination": "nicegui/elements/lib",
    "keep": [
      "package/lib/xterm.js",
      "package/css/xterm.css"
    ],
    "rename": {
      "package/lib/": "",
      "package/css/": ""
    },
    "version": "5.5.0"
  }
```

Shift things into `nicegui/elements/`, move to using `self.add_resource` for the CSS, and it should be much more better integrated.

paco-sevilla:
Sorry that I've been offline for so long... I finally found some time to work on this.
I've added the new element as requested, including tests, and updated the example.

I'll add some documentation/demos to the website in the next few days.
But I'm happy to receive a first pass of comments 🙂

.devcontainer/Dockerfile
```diff
@@ -9,7 +9,7 @@ ENV POETRY_VERSION=1.6.1 \
 
 # Install packages
 RUN apt-get update && apt-get install --no-install-recommends -y \
-    sudo git build-essential chromium chromium-driver \
+    sudo git build-essential chromium chromium-driver chromium-sandbox \
```
paco-sevilla: This is now required to run tests in the devcontainer.

poetry.lock
```diff
@@ -1,12 +1,11 @@
-# This file is automatically @generated by Poetry 2.1.2 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
```
paco-sevilla: Poetry only supports Python 3.8 up to version `1.6.1`.
A lock file generated with a newer version produces warnings (e.g. in the devcontainer).

However, I'm happy to revert this change, if preferred.

evnchn:
This PR is so old it witnessed 2 changes of our JS dependency management 😮

falkoschindler:
Oh I should finally start reviewing it! Dependency management should be even easier now.
I'll move it up in my list. 🤞🏻

falkoschindler:
Ok, I reviewed most of the code and migrated the dependency management to use a separate package.json and the new `esm` parameter. I just had to remove the example audio, because the URL wasn't reachable anymore. We either need a different source or host the audio file ourselves.

Now I'm pondering about the purpose of `ui.xterm`... The first demos are short and simple, but could seem like a poor copy of `ui.log` or `ui.codemirror`. The example, on the other hand, is pretty impressive, but also very complex. So I wonder if we could provide a simpler API for connecting `ui.xterm` to, e.g., a PTY process:

```py
process = await asyncio.create_subprocess_exec(...)
ui.xterm().connect(process)
```
Or:
```py
pty_pid, pty_fd = pty.fork()
ui.xterm().connect(pty_pid, pty_fd)
```
Or could `process` and `pty_pid, pty_fd` be initializer parameters?

Maybe that's something for a later PR if it doesn't require breaking the API again. But I'd like to hear your thoughts. 🙂

evnchn:
I like the idea of connecting `ui.xterm()` with something else, but I am expecting an explosion of things we connect to `ui.xterm()`, and I fear the element becomes super bulky. 

Would propose the creation of binding functions which take 2+ arguments, first being the `ui.xterm()` element, second being whatever you want to bind to.

paco-sevilla:
> I just had to remove the example audio, because the URL wasn't reachable anymore. We either need a different source or host the audio file ourselves.

I'm still able to download the audio (https://www.soundjay.com/buttons/beep-07a.mp3). It's just a short beeping sound (7kB).
If it's ok for you, we could include it (or something similar) in this repo and in NiceGUI. But I don't think the audio is that important.

> Now I'm pondering about the purpose of `ui.xterm`... The first demos are short and simple, but could seem like a poor copy of `ui.log` or `ui.codemirror`.

Yeah, I sort of see your point... What makes Xterm special is actually the parsing of ANSI escape codes (which is [not trivial](https://stackoverflow.com/a/59521115)).
So, I think most use cases of the Xterm element will involve a process/command running in the background (either a subprocess or a pty process). However, there are some caveats...

#### Caveats for subprocesses:
We could add a method like `run_subprocess` to the element, which takes the same args as `asyncio.create_subprocess_exec` but with `stdout` and `stderr` set to `subrpocess.PIPE` by default. The logic would be very similar to the last demo in this PR.

The problem is that most commands detect that `stdout`/`stderr` are not tty's and don't print outputs with ANSI codes, which makes having `Xterm` a bit useless (you might as well print the output to `ui.log`). This might surprise some users.
Many commands have a `--color` flag (like `ls --color=always`) to at least print the ANSI codes for text formatting, but they will still not print something like a progress bar.
 
Therefore, you actually want to use a PTY in most cases. But...

#### Caveats for pty:
The `pty` module in the standard library is _very_ basic (and has an awkward API). If you want to —for example— resize the terminal, you'll probably want to use a third-party package like [ptyprocess](https://github.com/pexpect/ptyprocess/) or [shellous](https://github.com/byllyfish/shellous).

Even worse, all of the above only work on Unix systems. If you're on Windows, you'll need to use a different package: [pywinpty](https://github.com/andfoy/pywinpty). To my knowledge, there is no cross-platform solution out there (see also https://github.com/python/cpython/issues/85829).

### Final thoughts
This comment is already quite long 🙈 My proposal is:

* **If it's acceptable to have different behavior on Windows vs. Unix:**
We can add a `run_subprocess` method that takes the same args as `asyncio.create_subprocess_exec` but by default sets `stdout` and `stderr` (and `stdin`?) to `subrpocess.PIPE` on Windows and to a PTY file descriptor that we get from [pty.openpty](https://docs.python.org/3/library/pty.html#pty.openpty) on Unix. Instead of a method, it could also be a separate `Process` class that `ui.xterm` can  `.bind()` to.

* **Otherwise**:
We keep the API minimalistic and let the user decide how to connect to the `Xterm` element.

In any case, I think the work should happen in a separate PR. The current API shouldn't be affected.

falkoschindler:
Ok, thanks a lot for the detailed explanation, @paco-sevilla!
You've convinced me that a subprocess or pty integration is non-trivial and that we shouldn't squeeze it into this PR. I'll finish my review and try to get it ready for version 3.1. 🙂

nicegui/elements/xterm/xterm.js
```diff
@@ -0,0 +1,65 @@
...
+      "LineFeed",
+      "Render",
+      "WriteParsed",
+      "Resize",
+      "Scroll",
+      "SelectionChange",
+      "TitleChange",
+    ]) {
+      this.terminal[`on${event}`]((e) => this.$emit(event.toLowerCase(), e));
+    }
```
evnchn: While works, may not be scalable approach due to Xterm.js may add new event handler. Proposing 2 solutions:

1. enumerate the keys of `this.terminal` and find those which begin with `on` then register them all (but I can't manage to enumerate the keys, maybe my JS skills are too weak 🤡 )
2. Do stuff to `self.on` so that it triggers JS code to then `if(`on${newEvent}` in myObj)` and register the event dynamically when see fit.

nicegui/elements/xterm/xterm.py
```diff
@@ -0,0 +1,118 @@
...
+        return await self.run_method('getColumns')
+
+    def fit(self) -> AwaitableResponse:
+        """Fit the terminal to its container.
+
+        This method only resizes the terminal, making the number of rows and columns match the size of the container.
+        Note that you might also need to resize the backing pty to match the new size of the terminal.
+
+        :return: AwaitableResponse that, if awaited, returns ``None`` once the method call is finished.
+                 Otherwise, the method is executed without waiting for it to finish.
```
evnchn: We do `:return: AwaitableResponse that can be awaited to get the result of the method call` for other elements, so I am thinking of `:return: AwaitableResponse that can be awaited to wait for the completion of the method call` to keep it one-line, concise, and consistent. It's a new example though.

nicegui/events.py
```diff
@@ -396,6 +396,16 @@ class JsonEditorChangeEventArguments(UiEventArguments):
...
 
 
+@dataclass(**KWONLY_SLOTS)
+class XtermBellEventArguments(UiEventArguments):
+    pass
+
+
+@dataclass(**KWONLY_SLOTS)
+class XtermDataEventArguments(UiEventArguments):
+    data: str
```
evnchn, line 406: @falkoschindler Necessary? I did confirm `onBell` is `IEvent<void>` and `onData` is `IEvent<str>` so this should be correct as of now. 

Just praying to Xterm.js that they won't have a special mode where `onData` returns a `dict` instead 🙏

nicegui/elements/xterm/xterm.js
falkoschindler: Good point. I replaced the code with
```js
Object.getOwnPropertyNames(Object.getPrototypeOf(this.terminal))
  .filter((key) => key.startsWith("on") && typeof this.terminal[key] === "function")
  .forEach((key) => {
    this.terminal[key]((e) => this.$emit(key.slice(2).toLowerCase(), e));
  });
```

nicegui/events.py
falkoschindler: Yes, the type should be correct. And any deviation would be considered a breaking change which hopefully only happens in major releases.

I noticed that we use the base class `UiEventArguments` for some events like "dismiss" of a notification or upload events, so we could drop `XtermBellEventArguments` and use plain `UiEventArguments` instead. On the other hand, there is `ClickEventArguments` as a counterexample. So I guess we can leave it as it is.

nicegui/elements/xterm/xterm.py
falkoschindler: I updated the four docstrings accordingly.

nicegui/events.py
evnchn: Kinda asymmetric to not have the Bell when you have the Data. Maybe we can leave it.

falkoschindler (changes requested):
There are still some things to do:

- [x] I think the documentation should move to "Controls" (like `ui.codemirror`).
- [x] We need another bell sound because the current one takes seconds to load (at least here in Germany). Or maybe its easier to trigger some visual indicator like a notification.
- [x] The `print` call in one of the demo is invisible when run online. A notification or maybe a `ui.log` is better.
- [x] 3 out of 5 tests are red because they are missing a page function. In 3.0 we lost the ability to write tests without an explicit page.
- [x] `test_write` will probably be flaky because it relies on nicegui.io. We should use a local `@page` instead.
- [ ] The fit addon doesn't seem to work (test fails).

Therefore I'll move this PR to milestone 3.2.

PR #4529, opened 2025-03-25
refactor: Introduces `LabelElement`

weinibuliu:
This PR introduces the `LabelElement `and applies it to the inheritance of the following components to provide a series of methods about label:

- ui.input
- ui.select
- ui.color_input
- ui.number
- ui.tab
- ui.text_area
- ui.upload


evnchn:
It's great, but I'm thinking one step ahead. 

We may want to allow post-creation editing of other things, such as placeholder. But then, this would require us to add 2 blocks of code per property. 

Then we might as well push this forward: https://github.com/zauberzeug/nicegui/pull/4464, unifying all such update maneuvers to the `.configure` method as discussed

weinibuliu:
Thanks @evnchn .

#4464 is progressing a bit slowly, and the community seems not very interested in it.

But even if #4464 is officially put into development, I don't think it will conflict with this PR. #4464 hopes to provide developers with a convenient way to set certain values ​​again, but it does not mean that it will be the only way.

falkoschindler (changes requested):
Thanks for the pull request, @weinibuliu! Adding label properties to these two elements could be handy in certain situations. And I also think it is a more intuitive and convenient way to get/set the label than a generic "configure" method.

But I'd like to extend it a little nonetheless: Besides `ui.input` and `ui.select` there is `ui.color_input`, `ui.knob`, `ui.number`, `ui.tab`, `ui.text_area`, and `ui.upload` that would also deserve a `label` property. To avoid immense code duplication, we like to implement mixins that can simply be added by elements with that feature.

For this PR, I suggest to reuse the name_element.py mixin and to create a label_element.py mixin. It's mostly copy&paste. Then said UI elements can inherit from this mixin. This way they not only get a `label` property, but a `set_label` method, a bindable property `label` and `bind_label`/`bind_label_from`/`bind_label_to` methods as well.

weinibuliu:
Thanks @falkoschindler 

I was originally planning to do this, but I wasn't sure which elements needs the `label()`.

I'll do my best to improve this PR.

weinibuliu:
Hi @falkoschindler  I have pushed a new commit.

The improvement of `ui.knob` is not yet completed. I not sure how to edit it.
I think we should not make any changes to `ui.knob`. The `label` property of the `ui.knob` is bound to its `value`, and I am worried that providing label-related methods will cause some bugs.

At the same time, I wrote a test.
```python
from nicegui import ui

# Test `set_label()`
with ui.card(), ui.row():
    ui.button("Set Label", on_click=lambda: set_label()).props("no-caps")
    a0 = ui.input(label="Test Label 3")
    b0 = ui.select(["A Test Option"], label="Test Label 2")
    c0 = ui.color_input("Test Label 3")
    d0 = ui.number("Test Label 4")
    e0 = ui.textarea("Test Label 5")
    f0 = ui.upload(label="Test Label 6")

    def set_label():
        i = 1
        for element in (a0, b0, c0, d0, e0, f0):
            element.set_label(f"New Label {i}")
            i += 1

# Test `bind_label()`
with ui.card(), ui.row():
    i1 = ui.input("Bind Test")
    test_bind = ui.input().bind_label(i1, "value")
    test_bind.set_visibility(False)

    a1 = ui.input(label="Test Label 3").bind_label(test_bind)
    b1 = ui.select(["A Test Option"], label="Test Label 2").bind_label(test_bind)
    c1 = ui.color_input("Test Label 3").bind_label_from(test_bind)
    d1 = ui.number("Test Label 4").bind_label(test_bind)
    e1 = ui.textarea("Test Label 5").bind_label(test_bind)
    f1 = ui.upload(label="Test Label 6").bind_label(test_bind)


ui.run()
```

falkoschindler (approved):
Thanks, @weinibuliu! The test script is really helpful. 👍🏻
The PR looks good to me. I just cleaned it up a little (e.g. we don't have to set `_props['label']` outside the mixin anymore) and added `ui.chat_message` (I forgot to mention this one last time).
Let's merge!

Issue #4535, opened 2025-03-27
apache echart always renders twice, why?


falkoschindler:
### Discussed in https://github.com/zauberzeug/nicegui/discussions/4501

<div type='discussions-op-text'>

<sup>Originally posted by **depley** March 19, 2025</sup>

The following code creates an EChart with 50,000 points, which seems to get rendered twice:
```py
ui.echart({
    'xAxis': {'type': 'value'},
    'yAxis': {'type': 'value'},
    'series': [{'type': 'scatter', 'data': [[x, x] for x in range(50_000)]}],
})
```

The problem can be reproduced without NiceGUI like this:
```html
<html>
  <head>
    <script src="https://cdn.jsdelivr.net/npm/echarts@5.5.1/dist/echarts.min.js"></script>
  </head>
  <body>
    <div id="chart" style="width: 600px; height: 400px"></div>
    <script>
      const el = document.getElementById("chart");
      const chart = echarts.init(el);

      const createResizeObserver = () => {
        new ResizeObserver(chart.resize).observe(el);
        chart.off("finished", createResizeObserver);
      };
      chart.on("finished", createResizeObserver);

      chart.setOption({
        xAxis: { type: "value" },
        yAxis: { type: "value" },
        series: [{ type: "scatter", data: Array.from({ length: 50000 }, (_, i) => [i, i]) }],
      });
    </script>
  </body>
</html>
```

Without the resize observer, the point cloud is only rendered once. It has been introduced in PR #3056 to solve a problem with animation durations.

If we can't find a way to always avoid the second rendering, we might at least activate or deactivate the resize observer depending on animation options.

depley:
Meanwhile I used echarts more and investigated in this topic.

Reacting on chart options dictionary at the python side will not work (like is there  some key starting with 'animation') .
The discussed examples uses series of the type 'line' (in animation example) or 'scatter' (in no animation many data example).
Both have per default `animation: true`, `animationDuration:1000`, .. so it will not help do distinguish.

depley:
A more general solution could be the approach used here [ecomfe/vue-echarts  .. autoresize.ts](https://github.com/ecomfe/vue-echarts/blob/b7852ab6430084c54a8e1c0a05648748da94bc64/src/composables/autoresize.ts#L7)
I tested the approach in the two following MREs.

**scatter series chart:** 
Does not render twice initially, renders again after resizing. [x]  

```html
<html>
  <head>
[...]

PR #4536, opened 2025-03-27
ui.echart switch resizeobserver attach mode automatically

evnchn:
Fixes issue https://github.com/zauberzeug/nicegui/issues/4535

Solution is discussed in: https://github.com/zauberzeug/nicegui/discussions/4501#discussioncomment-12564597 and onwards

---

TL-DR: 

Was not sure whether to make this PR because I fear large chart with animation will still render twice under this PR. Empirically, it doesn't, so this solution is perfect in all regards. 

---

Background context: 
* If we attach `ResizeObserver` straight away, animation is broken. 
* If we attach `ResizeObserver` on-finished, large charts gets rendered twice. 

Solution: 
* Attach `ResizeObserver` straight away if there is no animation
* Attach `ResizeObserver` on-finished if there is animation

Testing: 

|                | Animation           | No Animation         |
|----------------|---------------------|----------------------|
| **Large chart**      | **renders once, animation working*** | renders once |
| **Small chart**      | renders once, animation working | renders once, as usual |

*unexpected, but a good thing! May have to do with Apache EChart internals but I can't comprehend it 😅
My hypothesis: If there is animation, `this.chart` emits "finished" immediately when the animation starts, attaching the `ResizeObserver` right away (unlike when there is no animation where it emits "finished" only when it is finished drawing all points.

---

Testing script: 

```py
from nicegui import ui

ui.label("20000 points, no animation")
chart = ui.echart({
    'xAxis': {'type': 'value'},
    'yAxis': {'type': 'value'},
    'series': [{'type': 'scatter', 'data': [[x, x] for x in range(20_000)]}],
})

ui.label("5 points, with animation")
chart2 = ui.echart({
    'xAxis': {
        'type': "category",
        'data': ["Mon", "Tue", "Wed", "Thu", "Fri"],
    },
    'yAxis': {'type': "value"},
    'series': [{
        'type': "line",
        'data': [150, 230, 224, 147, 260]
    }],
    'animationDuration': 5000,
})

ui.label("20000 points, with animation")
chart3 = ui.echart({
    'xAxis': {'type': 'value'},
    'yAxis': {'type': 'value'},
    'series': [{'type': 'line', 'data': [[x, x] for x in range(20_000)]}],
    'animationDuration': 5000,
})

ui.label("5 points, no animation")
chart4 = ui.echart({
    'xAxis': {
        'type': "category",
        'data': ["Mon", "Tue", "Wed", "Thu", "Fri"],
    },
    'yAxis': {'type': "value"},
    'series': [{
        'type': "line",
        'data': [150, 230, 224, 147, 260]
    }],
})
ui.run()
```


falkoschindler (commented):

I think your test script is a bit inconsistent since it mixes different plot types. Here is a more generic example:
```py
for animation in [False, True]:
    for count in [20_000, 5]:
        ui.echart({
            'title': {'text': f'{count} points {"(animated)" if animation else ""}'},
            'xAxis': {'type': 'value'},
            'yAxis': {'type': 'value'},
            'series': [{'type': 'scatter', 'data': [[x, x] for x in range(count)]}],
            **({'animationDuration': 5000} if animation else {}),
        })
```
As far as I can tell, 20,000 points with animation still renders twice. Can you confirm that?

Apart from that, I think we should treat an animation duration of 0 or `None` like a missing key. This line should work:
```py
self._props['direct_attach_resizeobserver'] = bool(options.get('animationDuration'))
```

And why did you add this logic to the Python code? You could do it on the client, avoiding the need for an extra prop.

evnchn:
Point 1: [Inconsistent test script, 2000 points with animation still render twice] You're right. 

* The issue aparently has to do with scatter plots only. For line and bar, it's perfect even without this PR. 
* 20,000 points with animation still renders twice

Point 2 and 3: [Treat 0 or None like a missing key, why server-side]

My original intention was that, the user can still override the rendering mode via setting `_props['direct_attach_resizeobserver']` or by setting `'animationDuration': None` in options to rollback to the behaviour before the PR, if desired. 

Not sure if necessary, though.

evnchn:
Marking as draft, since #4584 seems more promising.

falkoschindler:
Yes, it looks like we found a better solution in PR #4584. Let's close this one.

PR #4537, opened 2025-03-28
let add_css, add_scss and add_sass set shared=True

evnchn:
TL-DR: As said in the title. 

---

I am not sure why it wasn't added in the first place. 

`shared=True` was added in this commit: https://github.com/zauberzeug/nicegui/commit/a09e713ece9099bc2d7164894d2d17d646b1d618, which was in [v1.4.6](https://github.com/zauberzeug/nicegui/releases/tag/v1.4.6)

`ui.add_css` was added in this commit: https://github.com/zauberzeug/nicegui/commit/3e5f92b23c5c7dfa9ab7306c52c84236e5fc0a34 ([v1.4.20](https://github.com/zauberzeug/nicegui/releases/tag/v1.4.20)), with `ui.add_style` (the precursor) in https://github.com/zauberzeug/nicegui/commit/a0c49a9112cfd04bc8d83d61ea5fc224ab52c95f ([v1.4.19](https://github.com/zauberzeug/nicegui/releases/tag/v1.4.19))

So the function which used `add_head_html` was added after the addition of `shared=True`? Why not expose all parameters of `add_head_html` since day1, if `ui.add_css` was just a wrapper around it which opens a `<style>` tag for you? 🤔 Or am I missing something here...

### DIFF
--- a/nicegui/functions/style.py
+++ b/nicegui/functions/style.py
@@ -13,7 +13,7 @@ from .. import helpers
 from .html import add_head_html
 
 
-def add_css(content: Union[str, Path]) -> None:
+def add_css(content: Union[str, Path], shared: bool = False) -> None:
     """Add CSS style definitions to the page.
 
     This function can be used to add CSS style definitions to the head of the HTML page.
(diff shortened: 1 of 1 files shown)


evnchn:
I don't personally use SCSS and SASS, but I can confirm CSS is working:

Can now apply the following new code to this discussion: https://github.com/zauberzeug/nicegui/discussions/4505#discussioncomment-12562061

```py
# No need to use this just for shared=True
'''ui.add_head_html("""
<style>
body { background-color: #121212; color: white; }
</style>
""", shared=True)'''

# We can now use this after the PR is merged
ui.add_css("body { background-color: #121212; color: white; }", shared=True)
```

falkoschindler (approved):
You're right, @evnchn. We should add the `shared` parameter to these functions as well. We must have forgotten to do so when introducing them in version 1.4.

I just added some documentation and enforced keyword arguments. Let's merge!

PR #4538, opened 2025-03-28
Customizable error screen w/ clean traceback (even async)

evnchn:
Refer to https://github.com/zauberzeug/nicegui/pull/4480

TL-DR: 

* `@app.exception_handler(Exception)` can return custom error page, but traceback is (1) messy and (2) useless for async functions
* #4480 propose baking in the traceback into the library, which was not positive because of potential security/privacy concerns
* Still, a clean and usable traceback is desired for debugging purposes, provided the user know about the risks. 
* Thus, this PR lets the user customize the error page. 

---

Some slight issues with this PR: 

* Existing exception handler defined via `@app.exception_handler(Exception)` will never be ran as a result. Is this acceptable? I think it is, but if not, maybe we have to invoke such handler manually (since simply raising an error = cannot return a result)

---

Sample code after the PR merged: 

```py
import traceback
from nicegui import ui, app

from nicegui.error import SAD_FACE_SVG

def error_content(exception) -> None:
    title = 'Server error (custom)'

    if isinstance(exception, str):
        message = exception
        message2 = ''
    else:
        message = exception.__class__.__name__
        message2 = traceback.format_exc(chain=False).strip()
        if str(exception):
            message += ': ' + str(exception)
            message2 = message2[:len(message2) - len(str(message))].strip()
        # drop lines in traceback if they belong to NiceGUI
        message2 = '\n'.join(message2.split('\n')[:1] + message2.split('\n')[4:])

    with ui.column().style('width: 100%; padding: 5rem 0; align-items: center; gap: 0'):
        ui.html(SAD_FACE_SVG).style('width: 8rem; padding: 1.25rem 0')
        ui.label(500).style('font-size: 3.75rem; line-height: 1; padding: 1.25rem 0')
        ui.label(title).style('font-size: 1.25rem; line-height: 1.75rem; padding: 1.25rem 0')
        ui.label(message).style('font-size: 1.125rem; line-height: 1.75rem; color: rgb(107 114 128)')
        if message2:
            ui.button('Show details').on('click', js_handler="() => {document.querySelector('.error-details').style.display = 'block'; document.querySelector('.show-details').style.display = 'none'; document.querySelector('.hide-details').style.display = 'inline-block'}").style(
                'margin-top: 1rem').classes('show-details')
            ui.button('Hide details').on('click', js_handler="() => {document.querySelector('.error-details').style.display = 'none'; document.querySelector('.show-details').style.display = 'inline-block'; document.querySelector('.hide-details').style.display = 'none'}").style(
                'margin-top: 1rem; display: none').classes('hide-details')
            ui.code(message2).style(
                'margin-top: 1rem; width: 100%; height: 20vh; overflow-y: auto; display: none').classes("error-details")
            
app.on_page_exception(error_content)

def my_errorenous_function():
    raise ValueError("This is an error")

@ui.page("/sync_error")
def error_page():
    ui.label("This is a sync error page")
    my_errorenous_function()

@ui.page("/async_error")
async def async_error_page():
    ui.label("This is an async error page")
    my_errorenous_function()

@ui.page("/")
def main_page():
    ui.label("This is the main page")
    ui.button("Go to sync error page", on_click=lambda: ui.navigate.to("/sync_error"))
    ui.button("Go to async error page", on_click=lambda: ui.navigate.to("/async_error"))

ui.run()
```

Screenshots are the same as the last PR, too lazy to take again 😅


evnchn:
https://github.com/zauberzeug/nicegui/pull/4538/commits/b0b546a6132773a86814f40d64a3359102bd852c re-raises the error if the user did not attach any custom page exception handler, afterwards the behaviour is identical to before this PR.

falkoschindler:
@evnchn I'm struggling to understand the "slight issues with this PR". What exactly doesn't work? I tested with
```py
@app.exception_handler(RuntimeError)
def unicorn_exception_handler(request: Request, exc: RuntimeError):
    return PlainTextResponse('Oops!')

@ui.page('/')
def index():
    raise RuntimeError()
```
which seems to keep working.

evnchn:
I _think_ if you do `app.on_page_exception(error_content)` (the new way of handling exception via returning custom page content) and also `@app.exception_handler(RuntimeError)`, the latter will not run. 

If that's not the case, ignore me 😅

My issue is: Imagine implementing some advanced logic to email yourself when there is an error, only to find it to be overridden when `app.on_page_exception(error_content)` is added. That would be quite an unexpected behavior. Normally would expect both custom page content AND trigger my own error-handling logic...

evnchn:
Let's see:

Test script: 

```py
from fastapi.responses import PlainTextResponse
from nicegui import ui, app

import os
os.system('cls' if os.name == 'nt' else 'clear')

@app.exception_handler(RuntimeError)
def unicorn_exception_handler(request, exc: RuntimeError):
    print('########Unicorn exception handler', type(exc))
    return PlainTextResponse('Oops!')

def native_exception_handler(exc):
    print('########Native exception handler:', type(exc))

app.on_exception(native_exception_handler)

def my_errorenous_function():
    raise ValueError("This is an error")

# PR 4583 only
def error_content(exception):
    import traceback
    ui.code(traceback.format_exc())

app.on_page_exception(error_content)

@ui.page('/sync_raise')
def sync_raise():
    raise RuntimeError()

@ui.page('/sync_error')
def sync_error():
    my_errorenous_function()

@ui.page('/async_raise')
async def async_raise():
    raise RuntimeError()

@ui.page('/async_error')
async def async_error():
    my_errorenous_function()

@ui.page('/')
def index():
    ui.label('Hello world!')
    ui.link('Sync raise', '/sync_raise')
    ui.link('Sync error', '/sync_error')
    ui.link('Async raise', '/async_raise')
    ui.link('Async error', '/async_error')

ui.run(port=4538)
```

Results summary: 

{8CE6D04C-853E-4C5C-95DD-57117E9D47A0}

Takeaways: 

1. Apparently it is different outcome if the page function `raise` an error, versus it runs code which "produces" an error. 🧠🤯
2. The original behaviour is super confusing, with it possible to run the unicorn exception handler, NiceGUI's exception handler, both, or neither 🤮
3. The new `app.on_page_exception` unifies the behaviour 😎 but it ignores both the exception handler attached via `@app.exception_handler(RuntimeError)` and `app.on_exception(native_exception_handler)` 😅

Things to think about: 

1. It seems super sketchy if the exception handlers can be ran / not ran, just because the page is sync / async and the error was raised by the page function / by some other function which the page function uses. 
2. It seems super sketchy if I attach a page exception, then all the 2 other exception handlers become useless paperweight. 

@falkoschindler can you double-check my findings on a Linux system, just in case? And tell me what you think about thinking points 1 and 2. Thanks!

evnchn:
The more I think about it, the more it feels weird that the native exception handler and unicorn exception handler can be just "skipped" in certain scenarios. 

I'll check that out and fix in another separate PR if I manage to figure out why. 

Meanwhile, I think the way forward for this PR is to simply invoke those two error handlers when the page exception handler is attached.

evnchn:
Hey. I just want to conclude what we have in this PR for a bit.

1. Bug: error handlers only sometimes invoked, depending on raise / execute code with error, async / sync
2. Feature: Custom error screen
3. Bug of the feature: error handlers not invoked when the above feature is used. 

**Point 1 deserves more attention,** since I'm considering logging all errors occuring in my application for easier debugging of my Dockerized NiceGUI apps, and I can also imagine some many other use cases. 

I also don't know if anyone is silently affected by the issue in the wild / not inspect their logs careful to find that it's missing. That wakes me up a night. 

Timeline wise, I aim for 2.15. 

Point 2 and Point 2 can be dealt with later™. 2.X is still preferred personally, since this isn't necessarily a breaking change.

falkoschindler (commented):
Hi @evnchn,

I just refactored your code a bit to reduce duplication. But there's a problem:

The function `wait_for_result` runs in a background task after `decorated` returns. So if there is an exception during `await result`, it is too late to return a `Response`. Actually the original code `return await response` is already a bit nonsensical because `wait_for_result` is supposed to return `None` anyway. - But I don't understand why the page exception handler seems to work nonetheless:
```py
app.on_page_exception(lambda ex: ui.label('Error'))

@ui.page('/')
async def index():
    ui.label('Hello World')
    await asyncio.sleep(0.1)
    raise Exception('test')  # causes the error page to be shown
```

Your points 1 and 2 indeed feel sketchy. I haven't recreated your experiments though. But I don't expect different results on macOS vs. Windows.

Apart from that, if we want to merge this PR eventually, it would be great to add pytests (might help during development as well) and an interactive demo (or example if a demo is too complicated to create).

nicegui/page.py
```diff
@@ -120,24 +128,14 @@ async def decorated(*dec_args, **dec_kwargs) -> Response:
...
                         try:
                             return await result
                         except Exception as e:
-                            if core.app._page_exception_handler is not None:
-                                with Client(page(''), request=request) as error_client:
-                                    core.app._page_exception_handler(e)
-                                    return error_client.build_response(request, 500)
-                            else:
-                                raise e
+                            return create_error_page(e, request)  # FIXME: at this point we can't return a response
```
evnchn, line 138: Yes we can. 

```py
if isinstance(result, Response):  # NOTE if setup returns a response, we don't need to render the page
    return result
```

And also, we don't encounter the following at all, since `task.done()` is true, as we did not `await ui.context.client.connected()`: 

```py
result = None
```

evnchn:
Oh actually, it may blow up, if we do have `await ui.context.client.connected()` and the error occurs AFTER said code. 

But I never thought of that to be an issue to be honest. And we can probably work around it. 

Anyways I'm way too tired for the day. We'll pick it up later.

falkoschindler:
Oh, I see! If `wait_for_result` returns something (agains its current type annotation), it will be extracted via `result = task.result()` and `isinstance(result, Response)` will lead to `return result` - so we are indeed returning the result of `create_error_page`. Let's quickly fix the type annotation: 0a44dddc2f937480191b7ee814d51daf105f8ae1.

If I'm correct, the following points are still open for discussion, right?

> 1. It seems super sketchy if the exception handlers can be ran / not ran, just because the page is sync / async and the error was raised by the page function / by some other function which the page function uses.
> 2. It seems super sketchy if I attach a page exception, then all the 2 other exception handlers become useless paperweight.

Can you give examples? It's hard to do it all in the head. 😰

evnchn:
Good catch on the type annotation. I'm not a typed guy so thanks for that. 

> Can you give examples? It's hard to do it all in the head. 😰

Check out the test script here https://github.com/zauberzeug/nicegui/pull/4538#issuecomment-2764780460

And inspect the console output very carefully to see if you get `########Unicorn exception handler` and/or `########Native exception handler:`

Note that the behaviour is somehow different, whether you `raise RuntimeError()`, or you run some function which `raise RuntimeError()` itself.

evnchn:
Hi. I just got to point out the obvious: 

**I raised `ValueError` in `my_errorenous_function`. No wonder the uvicorn exception handler is not ran! 🤡**

image

So, it simplifies to: 

> When it is a sync page, only the uvicorn exception handler is ran, not the NiceGUI one. 

Updated test code: 

```py
from fastapi.responses import PlainTextResponse
from nicegui import ui, app

import os
os.system('cls' if os.name == 'nt' else 'clear')


@app.exception_handler(RuntimeError)
def unicorn_exception_handler(request, exc: RuntimeError):
    print('########Unicorn exception handler', type(exc))
    return PlainTextResponse('Oops!')


def native_exception_handler(exc):
    print('########Native exception handler:', type(exc))


app.on_exception(native_exception_handler)


def my_errorenous_function():
    raise RuntimeError("This is an error")

# PR 4583 only


def error_content(exception):
    import traceback
    ui.code(traceback.format_exc())


# app.on_page_exception(error_content)


@ui.page('/sync_raise')
def sync_raise():
    raise RuntimeError()


@ui.page('/sync_error')
def sync_error():
    my_errorenous_function()


@ui.page('/async_raise')
async def async_raise():
    raise RuntimeError()


@ui.page('/async_error')
async def async_error():
    my_errorenous_function()


@ui.page('/')
def index():
    ui.label('Hello world!')
    ui.link('Sync raise', '/sync_raise')
    ui.link('Sync error', '/sync_error')
    ui.link('Async raise', '/async_raise')
    ui.link('Async error', '/async_error')


ui.run(port=4538)
```

evnchn:
With https://github.com/zauberzeug/nicegui/pull/4538/commits/e0b59febf3b8fb7afbeb2d99603ab9af08b1fd2d, the code gracefully attempts to access task.result().context['is_nicegui_error_page']. 

Should it be True, then we don't do `ignoring XXX: it was returned after the HTML had been delivered to the client`

evnchn:
With https://github.com/zauberzeug/nicegui/pull/4538/commits/beeca7332c56cc6fc78f87565d59ca67bc802eba, the error which occurs after client connected is sent to the client via Socket.IO. Extra metadata is included in an error response to facilitate reverse loopup of the original client which the error originated from, and send stuff in its outbox. 

Before this commit, the error is just printed in the console. User never gets to know. 

{1B1D257E-F625-4DC7-826D-9C26AE76D106}

evnchn:
Please use this test code: 

```py
from fastapi.responses import PlainTextResponse
from nicegui import ui, app

import os
os.system('cls' if os.name == 'nt' else 'clear')


@app.exception_handler(RuntimeError)
def unicorn_exception_handler(request, exc: RuntimeError):
    print('########Unicorn exception handler', type(exc))
    return PlainTextResponse('Oops!')


def native_exception_handler(exc):
    print('########Native exception handler:', type(exc))


app.on_exception(native_exception_handler)


def error_content(exception):
    import traceback
    ui.code(traceback.format_exc())


app.on_page_exception(error_content)  # PR 4583 only


@ui.page('/sync_raise')
def sync_raise():
    raise RuntimeError("This is an error in /sync_raise")


@ui.page('/async_raise')
async def async_raise():
    raise RuntimeError("This is an error in /async_raise")


@ui.page('/async_raise_after_connected')
async def async_raise_after_connected():
    raise RuntimeError("This is an error in /async_raise_after_connected")


@ui.page('/return_after_connected')
async def return_after_connected():
    return 'This is a return value in /return_after_connected'


@ui.page('/')
def index():
    ui.label('Hello world!')
    ui.link('Sync raise', '/sync_raise')
    ui.link('Async raise', '/async_raise')
    ui.link('Async raise after connected', '/async_raise_after_connected')
    ui.link('Return after connected', '/return_after_connected')


ui.run(port=4538)
```

rodja:
Besides a bit of documentation it might be good to create a small example to show users how it can be done.

evnchn:
@rodja See if this makes sense. 

{5C806F9C-FE61-4865-9B48-8C37F52553B4}

And the result when click into "Raise special exception":

{3B6E3ACA-EABF-4EEB-BE3B-6C468D43F3AE}

I have blocked off the traceback with a fake one for the website server, so that we don't leak any information.

falkoschindler:
@evnchn Quick question: Is it possible that `nicegui_error_metadata` isn't actually used on the client and can be removed? (No need to do it yourself, I already have some uncommitted changes...)

evnchn:
It’s bee a while since I worked on this, and my memory is a bit rusty at this point…

However, If I am not mistaken, `nicegui_error_metadata` was used to pass stuff around, so that it can show error on the browser, if there is late return value (or some other functionality). For my implementation, that variable itself is needed for that functionality to work, but yes it is not actively being listened to nor used in the client-side. 

Please commit your changes and we can see for ourselves. Thanks 😊

falkoschindler:
Oh, now I understand where `nicegui_error_metadata` is used: Even though it is added to the `TemplateResponse` which is sent to the client (where the meta data isn't used), the `check_for_late_return_value` function in page.py extracts the information to send it via websocket.
I'm still looking for a simpler solution...

falkoschindler:
I simplified the code quite a bit (I guess), but it seems like some tests are failing. I'll probably have to continue working on it tomorrow. Or maybe you @evnchn can review my changes and let me know what you think.

Two random thoughts:
1. The lines
    ```py
    if exception_handler is None:
        raise e
    ```
    in `create_error_page` might cause even deeper stacktraces than before since we catch an re-raise. But I haven't investigated it neither do I know a good solution. I just noticed pretty long stacktraces.
2. Should we match error codes from 500 to 599 instead of `if key == 500`?

evnchn:
So I guess you were trying to check if it is an exception before doing the `isinstance`

But resource came up dry, in terms of how to check if something is an exception, without mypy complaining. 

I also tried to check the type using `type()` but that’s futile. 

IMG_0187

I wonder: Since it was added via FastAPI, (1) do FastAPI do the check? (2) if they do, how they do it? 

Can we copy them?

_And yes, it’s an iPad 😅 You see I am not quite ready for programming._

evnchn:
And for your two questions: 

1. I don’t see a better way to do it, unless you “conditionally catch the error” in the beginning, which may be a good way out (or it may not, since I think `if exception_handler is None:` `raise e` is sort-of a good way to introduce this feature (if no exception handler then we spit out the error)
2. Right now the intention is for it to catch “internal server errors”, so yes, we should catch all 5xx codes, in retrospect. But, actually I am thinking if it should be somewhat configurable what error and error codes this new customizable error screen is catching. Perhaps we can have a custom error screen to catch 404s, or even 403s (if raised by an AuthMiddleware / manually returned). But I guess that can be in a separate PR and is arguably much less urgent that this one, since at the end of the day this PR has a huge benefit of showing traceback, but customizing the 404 is not that big of a deal IMO

falkoschindler:
> So I guess you were trying to check if it is an exception before doing the `isinstance`

Oh, I mixed up types and instances. `isinstance(key, Exception)` checks if `key` is an _instance_ of type Exception. But I wanted to check if `key` is a _type_, which can be done via `isinstance(key, type)`.

> but customizing the 404 is not that big of a deal

Well, if we want to enable the user to show different error pages per error code, we would need to somehow pass the error to the page exception handler. Currently it only expects an exception object which doesn't contain this information. We better think about it now than needing to break the API later.

evnchn:
I'd like to backpedal on the point of:

- This PR covering cases like 404 errors, since this should regard "exceptions" only, as evident by `app.on_page_exception`
- Multiple page exception handlers, where each handle a type of error, since we'll see below my point. 

But, your point is right: Now we expect first argument to take the exception. This is so that we can have different behaviour based on the error, (`if isinstance(e, ...):`), and so as an exception on that should we pass more information (like the HTTP error code)?

I am also thinking about an alternate API which is something like `@app.exception_handler(RuntimeError)` of FastAPI, which is `@ui.exception_page(RuntimeError)`. What do you think? 

I think, code functionality wise, all tests are passing, and so we're more or less done on that front. API design is what' left, I'd imagine.

falkoschindler:
After thinking about it: `app.on_page_exception` is for catching errors during page creation. Errors like 404 are pretty much different and can already be handled by the user with custom error pages:
```py
@app.exception_handler(404)
def my_exception_handler_404(request: Request, exception: Exception) -> Response:
    with Client(page.page(''), request=request) as client:
        ui.label('Not found 🙁').classes('text-2xl absolute-center')
    return client.build_response(request, 404)
```
So I don't think we need to to pass error codes to the page exception handlers because it will by definition always be an error 500.

evnchn:
Yep. 500 is the only error code returned by NiceGUI natively. 

I'm thinking, for ease of adding extra arguments when we need to do it, should we put everything inside a class (similar to XXXEventArguments for UI events)?

That way, assuming they don't make their own class on top of that, then we can add parameters without fear it's a breaking change.

falkoschindler:
> I'm thinking, for ease of adding extra arguments when we need to do it, should we put everything inside a class (similar to XXXEventArguments for UI events)?

In principle this is a good thought, especially for things like event handlers where the number of arguments are likely to change. But for an exception handler it feels like overcomplicating things. And we already have `app.on_exception` which expects a plain exception, nothing else. So I'd prefer sticking with this pattern.

By the way, `app.on_exception` expects an _optional_ exception parameter. This means the handler doesn't have to take it. Therefore I'd like to adjust `app.on_page_exception` accordingly.

evnchn:
Sure. 

I am also contemplating if we should let `page_exception_handler` be async. 

For doing that, we just need to make `create_error_page` into async

falkoschindler:
> I am also contemplating if we should let `page_exception_handler` be async.

Given that `app.on_exception` accepts sync and async handlers, one could expect the same from `app.on_page_exception`. I'm just a bit worried what happens if a page exception handler does crazy things like awaiting `client.connected` and dividing by zero afterwards. But it might just work. We should try it out.

evnchn:
In retrospect, I personally think that, given that the error screen is the last line of defense, and also that this PR has been sitting for a long time, that I would be more inclined to ship this without async support. 

I seem to think that sync functions, their behaviour are more "predictable" and is more "simple-minded". It's a stereotype of mine. 

Whoever argues for async handler can then feature request and then we can get coding again. 

Or if you want to get async support since day1 of the feature release, then we need some time. I am thinking about edge cases that may occur.

falkoschindler:
I agree. To support async error pages (that can await the handshake), we'd have to duplicate the code to wait for a client connection:
https://github.com/zauberzeug/nicegui/blob/ab9f7e0fd13f2c06daab0c4779c4f9b46575536f/nicegui/page.py#L126-L140
Let's stick with sync error pages for now.

falkoschindler:
Ok, I'm almost done reviewing. My last concern is about the following example which should result in an error popup displaying "ZeroDivisionError: division by zero":
```py
@ui.page('/')
async def test():
    ui.number(value=1 / 0)
```
- The popup is only shown if a page exception handler is registered. But both things shouldn't depend on each other. 🤔
- Apart from the popup not working reliably, I wonder if it is really a good idea to expose server error messages to the client. (I know, it is a bit late for such concerns. But better late than after releasing it to the public.) The exception could come from anywhere within the user codebase and is therefore hard to predict. It could be considered a breaking change. And the error popup seems to be pretty independent of the custom error page and isn't needed for the remaining PR.

Unfortunately I can't find any reasoning about how you came up with the error popup and its pros and cons. But given the above-mentioned concerns I lean towards removing it from this PR. What do you think, @evnchn and @rodja?

evnchn:
Interesting. I forgot why since it's a while ago, but I reviewed my past comments* and inferred from there.

So basically, my thinking was, if your page runs into an error post-HTTP-response: 

1. Originally, we do nothing, and the error spews out into the server console
2. Now, with the page exception handler, it then creates some UI elements and returns another HTTP response, which the code shortly after complains. 

In fixing the second point, I went with my preferred solution: Add post-HTTP-response error popup (at the end of the day I'd like debugging information, the more the better). 

But you're right in that: 

- It makes case 2 behaviour different from case 1
- It could be considered a privacy leak, which at the end of the day, we are trying to prevent leaking private information as a default with a customizable error screen 

So, my recommended course of action: 

- Rip out the post-HTTP-response error popup logic
- But do include the logic in https://github.com/zauberzeug/nicegui/pull/4538/commits/e0b59febf3b8fb7afbeb2d99603ab9af08b1fd2d so that we don't log the late return for page exception handler
- Jot down the commit(s) which ripped out the logic, so that we can based on that commit and PR another feature

What do you think?

*No accuracy guarantee. A sign that we should review PRs early?

rodja:
In https://github.com/zauberzeug/nicegui/pull/4480#issuecomment-2728848057 I wrote:

> We think it's not a good idea to show backend stacktraces on the web site error pages. This can be really problematic in production or even in quick shared demos. Of course it could be usefulness in development. Maybe it can be implemented as a separate package or even a gist?

And @evnchn wrote in response

> We could put this functionality behind a normally-off switch

falkoschindler:
@rodja Yet another switch... I would prefer encouraging NiceGUI developers to keep an eye on the Python (or even JavaScript) console instead of relying on debug information being sent to the UI.

@evnchn I agree with your recommended course of action. I'm not sure about e0b59febf3b8fb7afbeb2d99603ab9af08b1fd2d, but we'll see if we need it.

> A sign that we should review PRs early?

I'd love to. But with around 20 PRs waiting for review this will remain difficult to realize. Instead I'd vote for smaller PRs. Sometimes it's tempting to put two or more related features into a single PR (see PR #4810), but it adds complexity to the discussion, the diff and the review process. If we can divide (and conquer), we should do so.

evnchn:
Checked out the latest status. Used https://github.com/zauberzeug/nicegui/pull/4538#issuecomment-2967482671 as a test case. 

Basically, without the logic from https://github.com/zauberzeug/nicegui/commit/e0b59febf3b8fb7afbeb2d99603ab9af08b1fd2d (or similar counterparts), you would have a "ignoring <starlette.templating._TemplateResponse object at 0x000001DD50C58B90>; it was returned after the HTTP response had been delivered to the client" in the console, if your page errors out after `await ui.context.client.connected()`, as the page exception handler attempts to return another HTTP response. 

Which, I think could be mildly infuriating. But in retrospect, hiding that message may not be a good idea, because we'd not have any indication of an error then. Should we throw the error back into the console, as it always has been?

falkoschindler:
> Should we throw the error back into the console, as it always has been?

Yes, that's what I'm thinking as well. It might not be the most descriptive message, but it correctly states that some response couldn't be sent to the client. I would leave it this way for now. It should rarely occur anyway.

falkoschindler (approved):
Ok, I think the PR is finished. It turned out pretty nicely!
Let's merge! 🚀

PR #4539, opened 2025-03-28
Cachebusting via include timestamp into SHA256

evnchn:
Related discussion: https://github.com/zauberzeug/nicegui/discussions/4532#discussioncomment-12644326

TL-DR: 

* NiceGUI about to cache the files served more aggressively, with up to 1 year of immutable cache planned so far. 
* For developers of custom components, it means that manual cache-busting via renaming the JS file would have otherwise needed to be done. (Think: `class Counter(Element, component='counter-final-updated-fix2.js'):`)
* This could be troublesome, so cache busting is now handled by NiceGUI, for which the path the file is served over will be dependent on file content, so any change in the file content invalidates the browser cache useless and force re-fetch. 

Potential cons: 
* Slightly slower server startup. 
* If `compute_path` is ever invoked with a folder (which I don't see why you would do that, and in reality it never does), then cache busting would not work. 

Tested with: 
https://github.com/zauberzeug/nicegui/tree/main/examples/custom_vue_component
* Changes in the `.vue` files immediately propagate before and after the PR.
* Changes in the `.js` files immediately propagate **after the PR only.** 
* _Did not bother testing `resource`, since it would probably also work._


falkoschindler (commented):
Thanks, @evnchn!

I always thought about the development process, where you can simply deactivate the browser cache. But of course, if you develop your own component, it is served under "/_nicegui/{\_\_version\_\_}/...". And if you deploy an update, visitors will get a cached version. Thus, cache busting is probably a good idea!

What still bugs me is the case that `path` is a directory. This is impossible for libraries and Vue components because they expect to be called with Vue or JS files. But according to `Element.add_resource`, `path` can be a folder containing CSS and JS files (see codemirror.py, leaflet.py and fullcalendar.py).

While I don't think we should recursively read whole directories on startup just to create unique hashes, it might be worth a thought how we can do cache busting in this case. A few random ideas:
1. Use the modification timestamp of files and/or folders? (Could be unreliable depending on OS and filesystem.)
2. Use UUIDs instead of hashes? (Invalidates the cache on every restart.)
3. Serve user components and resources under a different route? (Allows to use different cache control strategy for builtin and user components, but doesn't really solve the problem for user components.)

evnchn:
@falkoschindler check this out. Hash also the folder. And it's 10% faster for large files with an implementation taken from `hashlib.file_digest` 

(reference: https://stackoverflow.com/a/44873382, https://github.com/python/cpython/blob/66aa78cbe604a7c5731f074b869f92174a8e3b64/Lib/hashlib.py#L228-L238, 10% time increase comes from my testing, where I put the 20MB Python installer exe into CodeMirror lib folder)

Also, fixed an issue whereby the hash will be re-computed on every page reload for `add_resource`, which is not apparent before this PR, but will incur +50ms load time for `ui.codemirror`, for example. 

---

Regarding the possibility of using modification timestamp, UUID, or even for some users who like to revert to old behaviour, I propose (somewhere) that we add the option `cachebusting_strategy`
* `'hash'`: consider the file hash
* `'timestamp'`: consider the file last modified time via `os.path.getmtime`
* `'uuid'`: use a random UUID which is different on every server reboot
* `None`: don't do any of that, revert to pre-PR behaviour and look at the hash of path only. 

Tell me if you think this is OK, and where should the option be added. Then, I can work on it.

evnchn:
helper functions can't access core.app.config?

{D9544266-EB27-4712-8EC8-433B0A11231C}

Cachebusting logic is as follows: 

```py
def hash_file_path_and_contents(path: Path) -> str:
    """Hash the given path and file content(s)."""

    try:
        cachebusting_strategy = core.app.config.cachebusting_strategy
    except AttributeError:
        print("AttributeError: core.app.config.cachebusting_strategy")
        cachebusting_strategy = None

    print(f'cachebusting_strategy: {cachebusting_strategy}')

    if cachebusting_strategy == 'uuid':
        return str(uuid.uuid4().replace('-', ''))

    hasher = hashlib.sha256()
    nicegui_base = Path(__file__).parent
    try:
        path_shortened = path.relative_to(nicegui_base)
    except ValueError:
        path_shortened = path

    hasher.update(path_shortened.parent.as_posix().encode())

    if cachebusting_strategy is None:
        return hasher.hexdigest()[:32]

    if path.is_file():
        hasher = hash_file(path, hasher)
    else:
        for p in path.rglob('*'):
            if p.is_file():
                print("*", end='', flush=True)
                if cachebusting_strategy == 'hash':
                    hasher = hash_file(p, hasher)
                else:
                    hasher.update(p.stat().st_mtime.to_bytes(8, 'big'))

    return hasher.hexdigest()[:32]
```

evnchn:
@falkoschindler I still have some difficulty reading the config defined at `core.app.config` for functions defined in `helper.py`, and I may need your help. 

With that, I can't proceed in adding configurability of `cachebusting_strategy` so that it is possible to switch betweein pre-PR behaviour, timestamp, SHA256, or random-on-boot.

falkoschindler:
> I don't quite get what you mean. I'd imagine since this cachebusting thing relates to how the server is operated, that it go into ui.run, just like, say, port number, and uvicorn options.

Yes, but we only make use of this parameter in dependencies.py, which is why I thought I could be a parameter of `Element.__init_subclass__` or something similar. A `ui.run` parameter could be misinterpreted as controlling every route or at least every static file, which it doesn't.

> Sorry but no strategy is cheap and perfect.

Sure. But if one of them was "good enough", we don't need "perfect".

> If the directory is 100MB large

Yes, but it isn't. I'm trying to estimate how long it actually takes for dependencies we're actually serving.

evnchn:
Hmm should we just always do timestamp-based cachebusting, since apparently it is always the case that when you modify a file the timestamp is modifed, and you must be intentionally shooting yourselves on the foot if you overwrite the timestamp after the fact. 

https://poe.com/s/t1E57ePCaz4ctsbWKX2d

For timestamp-based cachebusting, I was worrying whether there is an imaginary `write_without_modify_timestamp` OS-level function which applications writing to the file can call if they want. Apparently that doesn't exist (not unless you manually edit the bits beneath the filesystem), so we should be golden on the timestamp approach. 

@falkoschindler Since you appear to be doing the performance measurement, I won't interfere and I'll work on other things. I'm more inclined to dropping the approach of SHA256 hash every file in the folder due to the high amount of disk read involved, though.

evnchn:
https://poe.com/s/XIT5ulMk30T2V8NpZOyU

I'd recommend timestamp+size approach. 

Since, we no need to worry about the case that a hacker was sliding in changes to the files while intentionally not wanting the hash value to update (bro why, you'd want to infect more clients more quicker, right), so accuracy is less of a concern. 

Then we drop the option from `ui.run`. How nice. 

Still have work to do to un-do whatever changes I done to add the option to `ui.run`, though

evnchn:
Benchmarking: 

```
None approach     : 0.008700000180397183

Timestamp approach: 0.008663699962198734

Hash file approach: 0.027972699841484427

UUID approach     : 0.03197609994094819
```

Notably:

* Generating random UUID is even slower than hashing file
* Timestamp is as fast as None (run-to-run error at this point)

> Use Timestamp is the way to go!

falkoschindler:
@evnchn What exactly did you measure? I can't imagine that generating UUIDs takes more time than hashing files.
But I think using timestamps is a good (enough) solution. 👍🏻

(By the way, your conclusions are always formatted like quotes. Maybe use "->" instead of ">" to start such lines? 😉)

evnchn:
@falkoschindler It's a dumb benchmarking script. Get everything you need for running it at https://github.com/zauberzeug/nicegui/tree/main/examples/custom_vue_component

Not sure if this script is fair, but I can say is, accessing timestamp is almost as fast as accessing nothing, while random and hashing the entire file is equally as slow. 

```py
import time
from counter import Counter
from on_off import OnOff

import nicegui.helpers as helpers

import uuid

TIMECONSUMED = 0

def update_hash_given_file_mode_none(path, digestobj):
    print("Doing nothing! hash was", digestobj.hexdigest()[:32])
    return digestobj

def update_hash_given_file_mode_uuid(path, digestobj):
    # hash some random data (cache-busting on every server restart)
    digestobj.update(uuid.uuid4().bytes)
    print("Hashing random data! hash was", digestobj.hexdigest()[:32])
    return helpers.hash_file(path, digestobj)

def update_hash_given_file_mode_timestamp(path, digestobj):
    digestobj.update(int(path.stat().st_mtime).to_bytes(8, 'big'))
    print("Hashing file timestamp! hash was", digestobj.hexdigest()[:32])
    return digestobj

def update_hash_given_file_mode_contents(path, digestobj):
    """Hash the given file using same algorithm as cpython/Lib/hashlib.py"""
    buf = bytearray(8192)
    view = memoryview(buf)
    with path.open('rb') as file:
        while size := file.readinto(view):
            digestobj.update(view[:size])
    print("Hashing file contents! hash was", digestobj.hexdigest()[:32])
    return digestobj

def benchmarked_update_hash_given_file(path, digestobj):
    global TIMECONSUMED
    start = time.perf_counter()
    result = update_hash_given_file_mode_none(path, digestobj) 
    TIMECONSUMED += time.perf_counter() - start
    print("Time consumed so far:", TIMECONSUMED)
    return result

helpers.update_hash_given_file = benchmarked_update_hash_given_file 

from nicegui import ui

with ui.row(align_items='center'):
    counter = Counter('Count', on_change=lambda e: ui.notify(f'The value changed to {e.args}.'))
    ui.button('Reset', on_click=counter.reset).props('outline')

with ui.row(align_items='center'):
    on_off = OnOff('State', on_change=lambda e: ui.notify(f'The value changed to {e.args}.'))
    ui.button('Reset', on_click=on_off.reset).props('outline')

ui.codemirror()
 
print("Time consumed so far:", TIMECONSUMED) 

ui.run(uvicorn_reload_includes='*.py,*.js,*.vue')
```

falkoschindler:
Two problems:

1. You should remove print statements for performance analysis.
2. `update_hash_given_file_mode_uuid` is using `helpers.hash_file`. I think it should simply return the updated `digestobj`.

With these changes I get the following plausible results:
```
None:      ~0.01 ms
UUID:      ~0.25 ms
Timestamp: ~0.21 ms
Contents:  ~3.30 ms
```

evnchn:
Regardless, I think that for scanning changes in the file, hashing the timestamp is the best way forward. This is why I dropped the `cachebusting_strategy` outright. 

About UUID (aka random ID for each server restart): It feels a bit wasteful that we may be calling `uuid.uuid4().bytes` 100 times for 100 files in the directory, even when we need only call it once and apply it to the hash for it to be totally-random. 

So, I also let `update_hash_given_file_path` to be overridable. 

Also, I dropped the return, since the hash updates just fine regardless.

falkoschindler (commented):
Ok, I'm starting to get an overview of all the new hashing helpers. Some questions that came up:

- Do you agree with using `functools.lru_cache` instead of a custom memoization?
- We have `hash_file_path` and `hash_file_path_and_contents`. Both are used, but I wonder if we should use the newer `hash_file_path_and_contents` exclusively.
- Do we really need to account for overriding hashing functions? If not, all helpers would condense into a single functions:

```py
def hash_file_path_and_contents(path: Path) -> str:
    """Hash the given path and file content(s)."""
    hasher = hashlib.sha256()
    nicegui_base = Path(__file__).parent
    try:
        path_shortened = path.relative_to(nicegui_base)
    except ValueError:
        path_shortened = path

    hasher.update(path_shortened.parent.as_posix().encode())

    if path.is_file():
        hasher.update(int(path.stat().st_mtime).to_bytes(8, 'big'))
    else:
        for p in path.rglob('*'):
            if p.is_file():
                hasher.update(p.as_posix().encode())

    return hasher.hexdigest()[:32]
```

evnchn:
TL-DR: 
- Yes to `functools.lru_cache`
- Prefer keeping `hash_file_path` for compatibility with `app.add_static_file` and `app.add_media_file`
- Would like to absolutely backtrack on the idea of letting the user override the helper hash functions. 

---

> Do you agree with using `functools.lru_cache` instead of a custom memoization?

Agreed. Makes total sense. 

> On uisng `hash_file_path_and_contents` exclusively

The old function is used exactly twice for `app.add_static_file` and `app.add_media_file` respectively. 

```py
path = f'/_nicegui/auto/static/{helpers.hash_file_path(file)}/{file.name}' if url_path is None else url_path
...
path = f'/_nicegui/auto/media/{helpers.hash_file_path(file)}/{file.name}' if url_path is None else url_path
```

Notably, the old function is dependent on the file path only, but the new `hash_file_path_and_contents` depends on both the file path **and** the file contents (though indirectly, via the timestamp). 

This would result in the following knock-on effects:
1. When adding the file multiple times, if the file has changed, it would return a different path. 
2. If it is `single_use`, accessing one path does not invalidate the other. 

**Point 1 can break existing code. Point 2 is straight-up a security shortcoming.** 

There are also some minor effects:
3. Since the responsibility of trying to shorten the path to be relative to NiceGUI folder is shifted from `compute_key` to `hash_file_path_and_contents` for the latter to read the file, if we use `hash_file_path_and_contents`, it will always try (and fail) to shorten the path. 


So, either: 
- We keep the old `hash_file_path`
- We need to add extra logic to `app.add_static_file`, `app.add_media_file`, and `hash_file_path_and_contents` so that they play along nicely. 

I'd much prefer the former. 

> Do we really need to account for overriding hashing functions? 

After thinking about it for some time, **we should not endorse overriding NiceGUI internal functions for patching the behaviour.** 

If people were to ask for alternate forms of cache busting (99% won't, to be frank), they can raise another feature request, and we'll assess whether it's good, whether it's safe, whether it causes undefined behaviour, and then only add that. 

Imagine the following scenario:

```py
def really_poorly_written_hash_file_function(path: Path, digestobj: hashlib._Hash) -> None:
    # Whole range of crazy behaviour, such as:
    # Doing a bad job of hashing, with way too many hash collisions?
    # Accidentally modifies / deletes the file it is hashing?
    # UTF-8 intolerant in some manner?
    # Uses up excessive memory?
    # Tries to make a web request (in sync code, mind you) and grind the server to a halt?
    ....

helpers.update_hash_given_file = really_poorly_written_hash_file_function
```

We would be in a very bad position, with no control to "sanity check" the user's function before executing. 

Therefore, allowing for overrides, we are opening a big can of worms.

evnchn:
@falkoschindler https://github.com/zauberzeug/nicegui/pull/4539#pullrequestreview-2823255620

Sorry it's my fault but somewhere the logic went wrong. `if p.is_file():` then we should call `update_hash_given_file` aka `hash_file` and not `hasher.update(p.as_posix().encode())`

evnchn:
Decided to keep helper function `hash_file` since the logic is rather bulky and too-much to inline IMHO. 

Though that would make overriding the function is still possible, we don't endorse it so it should be fine. 

After all, we can't inline everything just because we don't want people to override it. They'll find a way regardless, and a worse way I may add.

evnchn (commented):
Otherwise, it _could_ be possible to brute-force when somebody installed NiceGUI, given the hash. A slight privacy leak. Also `struct.pack` is just plain better I think.

evnchn:
https://github.com/zauberzeug/nicegui/pull/4539/commits/fe667034ea2e95a3e7608e0974a38cd827414250

@falkoschindler sure about that? IIRC then the `hash_file_path_and_contents` can't read the file properly and error out.

falkoschindler:
Oh, I didn't think about that. Now I wonder _why_ it works. What if we are in a different working directory? How can `path.stat()` still read the file? Anyway, I'm still refactoring and the question might vanish in a moment.

Another two questions for you:
- Why do we need `path.parent` in our hash? Shouldn't `path` be better?
- Would it make sense to hash the most recent timestamp only? I could imagine `max()` is a tad cheaper than hashing them all.

evnchn:
> Why do we need path.parent in our hash? Shouldn't path be better?

No idea but I was trying to mock old behaviour. 

> Would it make sense to hash the most recent timestamp only? I could imagine max() is a tad cheaper than hashing them all.

Theoretically a good idea, but perhaps we should benchmark this. Since hashlib is no slow thing either.

falkoschindler (approved):
Ok, I'm pretty pleased with the current implementation. The diff is rather small again:

- We use `@functools.lru_cache` for `compute_key`.
- The `NICEGUI_BASE` constant is now uppercase.
- I inlined `path.is_file()`.
- The hash now includes the maximum mtime of the file itself (if `path` is a file) or all globed files (if `path` is a directory). Note that we use non-relative paths for mtime, just to be save.

Moving the hash function into dependencies.py helps tailoring it to the specific needs of `compute_key` without worrying too much about a possibly public API. Now the PR is doing exactly one thing: It includes timestamps in the hash for `compute_key` (and it adds an LRU cache...).

What do you think, @evnchn? Can you confirm it's working? (After class. No rush! 😉)
From my point of view it's ready to merge.

evnchn:
The new code is very elegant and almost ready to merge in all regards. 

My testing method: 
- Start a NiceGUI page, view source
- Modify the aggrid a bit, view source
- Verify that the script in the importmap section has changed its cachebusting value. 

{4EF4ADB3-8AF1-439A-BF2D-B711349A0EA3}

{2D90C000-86E7-49A2-BE86-D3503E88CF11}

But, I have a concern:
- What happens if `files` iterable is empty? 
  - Then `max` would error out with `ValueError: max() iterable argument is empty`
  - Knock-on effect being that adding an empty folder would error out, then. 

Also, should we really have the two `hash_file_path` be under the same name, when they behave entirely differently? 

And, I have an extra thought, that we can put the new `hash_file_path` in `dependencies.py` back into `helpers.py`, provided we: 
- Let the new `hash_file_path` take `None` for `files`
- Assuming that we let `files` iterable be empty and not error, we pass an empty iterable, like `path = f'/_nicegui/auto/static/{helpers.hash_file_path(file, [])}/{file.name}' if url_path is None else url_path`

Can this patch work?

{D5B47F0E-9056-4041-B6CC-A0BC8C432D46}

First patch: Make `hash_file_path(blahpath)` work, for letting it serve in helpers.py and the `app.add_XXXX_file`
Second patch, make `hash_file_path(blahpath, some_empty_iterator)` work, in case glob matches nothing.

evnchn:
Hashes for mermaid in the importmap is allovertheplace. Why?

<details>
<summary> Those hashes </summary>

```json
{
    "imports": {
        "mermaid": "/_nicegui/2.13.0.dev0/libraries/7280f6d538de10a06b6db5769384fb95/mermaid.esm.min.mjs",
        "blockDiagram-NDWNTGEE": "/_nicegui/2.13.0.dev0/libraries/4050c3db6b5319c26c7d2372db3db956/blockDiagram-NDWNTGEE.mjs",
        "c4Diagram-AUYESYAG": "/_nicegui/2.13.0.dev0/libraries/4050c3db6b5319c26c7d2372db3db956/c4Diagram-AUYESYAG.mjs",
        "chunk-2RSIMOBZ": "/_nicegui/2.13.0.dev0/libraries/c442925a03b13014a1e9363ffdb90b5f/chunk-2RSIMOBZ.mjs",
        "chunk-4BPNZXC3": "/_nicegui/2.13.0.dev0/libraries/c442925a03b13014a1e9363ffdb90b5f/chunk-4BPNZXC3.mjs",
        "chunk-4KE642ED": "/_nicegui/2.13.0.dev0/libraries/c442925a03b13014a1e9363ffdb90b5f/chunk-4KE642ED.mjs",
        "chunk-4SRTBRON": "/_nicegui/2.13.0.dev0/libraries/c442925a03b13014a1e9363ffdb90b5f/chunk-4SRTBRON.mjs",
        "chunk-6BY5RJGC": "/_nicegui/2.13.0.dev0/libraries/8ccec8d9f2ecbd3f5cf60baa1afccbd9/chunk-6BY5RJGC.mjs",
        "chunk-6IZS222M": "/_nicegui/2.13.0.dev0/libraries/8ccec8d9f2ecbd3f5cf60baa1afccbd9/chunk-6IZS222M.mjs",
        "chunk-6XGRHI2A": "/_nicegui/2.13.0.dev0/libraries/da45e3eecb8dae5a9521219815c02cfa/chunk-6XGRHI2A.mjs",
        "chunk-AC3VT7B7": "/_nicegui/2.13.0.dev0/libraries/da45e3eecb8dae5a9521219815c02cfa/chunk-AC3VT7B7.mjs",
        "chunk-BKDDFIKN": "/_nicegui/2.13.0.dev0/libraries/da45e3eecb8dae5a9521219815c02cfa/chunk-BKDDFIKN.mjs",
        "chunk-BOP2KBYH": "/_nicegui/2.13.0.dev0/libraries/da45e3eecb8dae5a9521219815c02cfa/chunk-BOP2KBYH.mjs",
        "chunk-C7NU23FD": "/_nicegui/2.13.0.dev0/libraries/6a48737a6597bd5bce04126995d64ee7/chunk-C7NU23FD.mjs",
        "chunk-CBSWTUHP": "/_nicegui/2.13.0.dev0/libraries/6a48737a6597bd5bce04126995d64ee7/chunk-CBSWTUHP.mjs",
        "chunk-D3PZO57J": "/_nicegui/2.13.0.dev0/libraries/b422f44f919bdcbeadefe88f3d83495d/chunk-D3PZO57J.mjs",
        "chunk-DZFIHE2J": "/_nicegui/2.13.0.dev0/libraries/8035ce42052e598bf8dce57218513b52/chunk-DZFIHE2J.mjs",
        "chunk-E4AWDUZE": "/_nicegui/2.13.0.dev0/libraries/8035ce42052e598bf8dce57218513b52/chunk-E4AWDUZE.mjs",
        "chunk-F4773GRL": "/_nicegui/2.13.0.dev0/libraries/53a3b77145465fb73dfa82ab7d7d3d95/chunk-F4773GRL.mjs",
        "chunk-GTKDMUJJ": "/_nicegui/2.13.0.dev0/libraries/53a3b77145465fb73dfa82ab7d7d3d95/chunk-GTKDMUJJ.mjs",
        "chunk-KMOJB3TB": "/_nicegui/2.13.0.dev0/libraries/ede453829368b6f81944726d09aa1b0e/chunk-KMOJB3TB.mjs",
        "chunk-NQURTBEV": "/_nicegui/2.13.0.dev0/libraries/0cfe8afe3ed041579feab373395cb9d6/chunk-NQURTBEV.mjs",
        "chunk-OUDNNCD4": "/_nicegui/2.13.0.dev0/libraries/0cfe8afe3ed041579feab373395cb9d6/chunk-OUDNNCD4.mjs",
        "chunk-RRFB4HDS": "/_nicegui/2.13.0.dev0/libraries/386e2ea5145e048385474c263fcd3513/chunk-RRFB4HDS.mjs",
        "chunk-TI4EEUUG": "/_nicegui/2.13.0.dev0/libraries/386e2ea5145e048385474c263fcd3513/chunk-TI4EEUUG.mjs",
        "chunk-U6LOUQAF": "/_nicegui/2.13.0.dev0/libraries/386e2ea5145e048385474c263fcd3513/chunk-U6LOUQAF.mjs",
        "chunk-UWHJNN4Q": "/_nicegui/2.13.0.dev0/libraries/1ba7cc199f464337b57bbf05718df8e8/chunk-UWHJNN4Q.mjs",
        "chunk-VRGDDFRA": "/_nicegui/2.13.0.dev0/libraries/1ba7cc199f464337b57bbf05718df8e8/chunk-VRGDDFRA.mjs",
        "chunk-VSLJSFIP": "/_nicegui/2.13.0.dev0/libraries/1ba7cc199f464337b57bbf05718df8e8/chunk-VSLJSFIP.mjs",
        "chunk-WVHPJQMP": "/_nicegui/2.13.0.dev0/libraries/750bb7add409f84d312e1e1ce2c2c732/chunk-WVHPJQMP.mjs",
        "chunk-YFFLADYN": "/_nicegui/2.13.0.dev0/libraries/750bb7add409f84d312e1e1ce2c2c732/chunk-YFFLADYN.mjs",
        "chunk-YJEQJWB7": "/_nicegui/2.13.0.dev0/libraries/aaca6a3344c8c4764ddc3f59d79d9a3e/chunk-YJEQJWB7.mjs",
        "chunk-YPUTD6PB": "/_nicegui/2.13.0.dev0/libraries/aaca6a3344c8c4764ddc3f59d79d9a3e/chunk-YPUTD6PB.mjs",
        "classDiagram-EPVPYR3L": "/_nicegui/2.13.0.dev0/libraries/aaca6a3344c8c4764ddc3f59d79d9a3e/classDiagram-EPVPYR3L.mjs",
        "classDiagram-v2-NO4EPWGV": "/_nicegui/2.13.0.dev0/libraries/fef9f67de83ed6ffe877f42f3150a86b/classDiagram-v2-NO4EPWGV.mjs",
        "dagre-EVPMPUST": "/_nicegui/2.13.0.dev0/libraries/fef9f67de83ed6ffe877f42f3150a86b/dagre-EVPMPUST.mjs",
        "diagram-V25JEYTC": "/_nicegui/2.13.0.dev0/libraries/fef9f67de83ed6ffe877f42f3150a86b/diagram-V25JEYTC.mjs",
        "erDiagram-Y4N7DENO": "/_nicegui/2.13.0.dev0/libraries/8939e48bf514de965bdfb251d13f1f0d/erDiagram-Y4N7DENO.mjs",
        "flowDiagram-JTTVBJUY": "/_nicegui/2.13.0.dev0/libraries/ea018fb7f80d0e240a35d4027a9e1f4c/flowDiagram-JTTVBJUY.mjs",
        "ganttDiagram-6SR64PWN": "/_nicegui/2.13.0.dev0/libraries/0afa164bd6dca63eb8bdc655185a615d/ganttDiagram-6SR64PWN.mjs",
        "gitGraph-F2EDSAW4-LKDW3VZQ": "/_nicegui/2.13.0.dev0/libraries/0afa164bd6dca63eb8bdc655185a615d/gitGraph-F2EDSAW4-LKDW3VZQ.mjs",
        "gitGraphDiagram-LRIBUTDQ": "/_nicegui/2.13.0.dev0/libraries/0afa164bd6dca63eb8bdc655185a615d/gitGraphDiagram-LRIBUTDQ.mjs",
        "info-PWGDJKR5-T7ZKA3RY": "/_nicegui/2.13.0.dev0/libraries/878df731c7bbdbc2e76f13d3371af54e/info-PWGDJKR5-T7ZKA3RY.mjs",
        "infoDiagram-YTPSHCSX": "/_nicegui/2.13.0.dev0/libraries/878df731c7bbdbc2e76f13d3371af54e/infoDiagram-YTPSHCSX.mjs",
        "journeyDiagram-VRXW2F6L": "/_nicegui/2.13.0.dev0/libraries/b3ec34d9c9e55dae8619c8b0715e12bf/journeyDiagram-VRXW2F6L.mjs",
        "katex-SWYD7GD6": "/_nicegui/2.13.0.dev0/libraries/23cfa6139350c5836e30ad0329a9e448/katex-SWYD7GD6.mjs",
        "mindmap-definition-KLCIANCF": "/_nicegui/2.13.0.dev0/libraries/e35e4d0efbcaf30bc16ac80705269998/mindmap-definition-KLCIANCF.mjs",
        "packet-7PPW3X5M-O6SZ3T2R": "/_nicegui/2.13.0.dev0/libraries/e35e4d0efbcaf30bc16ac80705269998/packet-7PPW3X5M-O6SZ3T2R.mjs",
        "pie-RF5LNP4B-V656W2AT": "/_nicegui/2.13.0.dev0/libraries/e35e4d0efbcaf30bc16ac80705269998/pie-RF5LNP4B-V656W2AT.mjs",
        "pieDiagram-WYSUK7CQ": "/_nicegui/2.13.0.dev0/libraries/801f9da99e221e163b98df44b993ed38/pieDiagram-WYSUK7CQ.mjs",
        "quadrantDiagram-K5BY4R5E": "/_nicegui/2.13.0.dev0/libraries/801f9da99e221e163b98df44b993ed38/quadrantDiagram-K5BY4R5E.mjs",
        "requirementDiagram-2DBX4ZW4": "/_nicegui/2.13.0.dev0/libraries/7bcfb7bedba3dbbcfafd1d10f40da928/requirementDiagram-2DBX4ZW4.mjs",
        "sankeyDiagram-F3GI3WFI": "/_nicegui/2.13.0.dev0/libraries/7bcfb7bedba3dbbcfafd1d10f40da928/sankeyDiagram-F3GI3WFI.mjs",
        "sequenceDiagram-QBGN4KDC": "/_nicegui/2.13.0.dev0/libraries/dc636272210b1d185ce0a06a788232a5/sequenceDiagram-QBGN4KDC.mjs",
        "stateDiagram-S3XGF733": "/_nicegui/2.13.0.dev0/libraries/d0138d196adc1f42948786297c3e2f49/stateDiagram-S3XGF733.mjs",
        "stateDiagram-v2-BSO5MUCZ": "/_nicegui/2.13.0.dev0/libraries/d0138d196adc1f42948786297c3e2f49/stateDiagram-v2-BSO5MUCZ.mjs",
        "timeline-definition-MHTE3MCH": "/_nicegui/2.13.0.dev0/libraries/d0138d196adc1f42948786297c3e2f49/timeline-definition-MHTE3MCH.mjs",
        "xychartDiagram-RLS75X5Z": "/_nicegui/2.13.0.dev0/libraries/d0138d196adc1f42948786297c3e2f49/xychartDiagram-RLS75X5Z.mjs"
    }
}
```

</details>

evnchn:
Ah, I seem to figured out why. 

The test has always been failing for Mermaid. 

But this commit https://github.com/evnchn/nicegui/commit/6c88d08a7fa53a785f6671d4b1f804591d0de764, I mixed the two functions, leading to an effectively "less effective cachebusting implementation which doesn't hash the file (so it's the same as pre-PR) and doesn't break Mermaid. 

{692B5CE8-CC6D-4647-8F93-A066BC2EC0AB}


This commit made things right and the logic fails Mermaid from there. 

https://github.com/evnchn/nicegui/commit/cb7f29f725a2fc81c13afe48f53966073e2f2b04

image

Back to the drawing board, then!

evnchn:
{C6CC3476-523B-46C8-A258-7B3766FC1030}

All imports made by chunks are failing. 

Because they imported it like this. 

{7693F497-1ACD-4D66-B0F9-286CE75D008D}

Note the "./chunk-NQURTBEV.mjs"

So maybe the entire folder has to share one cachebusting value.

evnchn:
@falkoschindler 

I beat https://github.com/zauberzeug/nicegui/pull/4539/commits/1d42ad8b70c68e850ed3a33c7c57ca760f2c26f0 into workingH, but the code is very ugly IMHO

Should we instead of `register_libraries` and `register_vue_components`, and maybe `register_resources` which does the globbing and the `max_time` calculation all-in-one?

evnchn:
See I was drafting documentation text while I was on the train

But I don't think we have to document this much? It's just underlying improvement

falkoschindler:
> I beat https://github.com/zauberzeug/nicegui/commit/1d42ad8b70c68e850ed3a33c7c57ca760f2c26f0 into working, but the code is very ugly IMHO

Yes, I wonder if we can improve it. Do I understand correctly that you basically use the same `max_time` (the timestamp of the last file change) of the whole component/library/dependency for hashing each file?

> Should we instead of register_libraries and register_vue_components, and maybe register_resources which does the globbing and the max_time calculation all-in-one?

I don't understand this sentence. Seems like the main verb is missing. 😵‍💫

> But I don't think we have to document this much? It's just underlying improvement

Agree. We won't need much documentation for this PR.

evnchn:
I meant: "Should we instead **do** register_libraries and register_vue_components, and maybe register_resources which does the globbing and the max_time calculation all-in-one?"

Sorry for the confusion!

But hey the code's currently working, so we're getting somewhere. 

And yes, the concept is to use the same `max_time` for the whole component/library/dependency

falkoschindler (approved):
Ok, I guess I understand what you mean. But I don't think we should spend much time refactoring the dependency registration because most of it is deprecated and will be gone in 3.0 anyway. Only `dependencies` will still be supported.

During a quick test I noticed a (probably unrelated) quirk with codehilite.css which isn't cached at all. And it is served at an unusual route. But we should investigate it separately.

Maybe you can have a brief look into my last two commits. If you agree, this PR is ready to merge.

evnchn:
I see you are changing the calling parameter of the three `register_xxx` functions. 

I am not sure if we can do that, but it _seems_ that we can, because these functions are meant to be used exclusively by element.py after this commit https://github.com/zauberzeug/nicegui/commit/4c5745b1f9c46ef8a40dfa98dd278e8fcc1185dc

What do you think? 

And, for 3.0, we should truly mark these as private by prefixing with `_`. We should have done it back then, but oh well...


---

Oh and for cohehilite.css: It is registered in a completely special way: 

https://github.com/zauberzeug/nicegui/blob/be0ae72aa94f17f8623873381f8fc1b6e2ebf2eb/nicegui/elements/markdown.py#L13C1-L40

I guess we can do something about that later, but IMO that doesn't mean we have to hold up this PR, since it's just 1 special file.

falkoschindler:
> And, for 3.0, we should truly mark these as private by prefixing with `_`. We should have done it back then, but oh well...

I don't think so. The whole module isn't public API (not part of any \_\_init\_\_.py), so "public" functions are only public to the NiceGUI library. If you would mark those functions a private, linters will complain when you use them within the library and you would have to `# ignore` these warnings.

> Oh and for cohehilite.css: It is registered in a completely special way:

Indeed. And there's even a related bug report from this morging: #4774.

PR #4540, opened 2025-03-28
Support ui.run(cache_control_directives=)

evnchn:
https://github.com/zauberzeug/nicegui/discussions/4532#discussioncomment-12643633

TL-DR: 

* Cache the static assets more aggressively, but we don't know how much is enough, and afraid overdo it. 
* Expose `cache_control_directives` argument in `ui.run` and `ui.run_with`, with default being long 1-year cache*
* Better than original solution because: 
  * Centralized management, as @falkoschindler pointed out
  * Ability to set other directives such as `immutable` (not available for Chrome), `stale-while-revalidate` (available for Chrome)

Concerns: 

* Is `ui.run` and `ui.run_with` the best place to add the parameter? 
* *Should we set the default to "public, max-age=3600" to match the default behaviour from before the PR (maximize compatibility), or let the better (?) 1-year-cache be the default (maximize rollout of performance increase)?

Testing: 

```py
ui.run()
```

{28E98406-CA70-42CE-94FA-8C894D1E6E5C}

```py
ui.run(cache_control_directives='no-store')
```

{40FAC44F-07B0-44B8-910D-F0A31875726A}


evnchn:
1. Sorry for that oversight. Should be "cache control directives for NiceGUI internal files"
2. According to the config, a request made to `http://nginx.server/nicegui/hello` would be relayed to `http://localhost:8080/hello`. 

```
        location ~ ^/nicegui/(.*)$ {
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection $connection_upgrade;
            proxy_set_header Authorization $http_authorization;
            proxy_pass_header Authorization;

            proxy_pass http://app:8080/$1?$args;
            proxy_set_header X-Forwarded-Prefix /nicegui;
            proxy_redirect http://app:8080/ /nicegui/;
```

- It is mentioned that `the entire app is available below "/nicegui" but the code does not need to know that.`, so the only way the line `if request.url.path.startswith(f'/_nicegui/{__version__}/'):` fails, is if it parses `X-Forwarded-Prefix` and mutates the path, which would be a security vulnerability (what's to stop me sending such header directly to a bare NiceGUI server, [cough cough Next.js](https://nextjs.org/blog/cve-2025-29927) 🤡), and just doesn't make any sense.

evnchn:
Final note: If you want to add this PR before cachebusting PR is merged, then set the default parameter of `cache_control_directives` to be exactly what would be served before this PR (`public, max-age=3600`)

But I persoanlly don't recommend rushing it. Seems a bad look if we add a parameter only that the default value changes from one version to another in a minor release.

evnchn:
There seems to be a circular import issue which I can't work around, making it impossible to access the app config: 

```
  File "C:\Users\User\Documents\cachebusting\.venv\Lib\site-packages\nicegui\app\__init__.py", line 1, in <module>
    from .app import App
  File "C:\Users\User\Documents\cachebusting\.venv\Lib\site-packages\nicegui\app\app.py", line 14, in <module>
    from ..client import Client
  File "C:\Users\User\Documents\cachebusting\.venv\Lib\site-packages\nicegui\client.py", line 19, in <module>
    from .dependencies import generate_resources
  File "C:\Users\User\Documents\cachebusting\.venv\Lib\site-packages\nicegui\dependencies.py", line 10, in <module>
    from .helpers import hash_file_path_and_contents
```

So any attempt at importing the app for the config from helpers.py, will lead to circular import.

evnchn:
Unblocked since #4539 merged

evnchn:
Since #4778 is merged, I need to modify this to make sure this PR doesn't cache the dynamic response endpoint.

evnchn:
There we go, https://github.com/zauberzeug/nicegui/pull/4540/commits/d00fb6b2b513ffa5b0e4d39c556953a10f5fef5f and https://github.com/zauberzeug/nicegui/pull/4540/commits/9bfa6ad69baed906a1e3e3463be82914bc496934

The test fails before the fix was implemented, and passes afterwards.

PR #4543, opened 2025-03-28
Support on_begin_upload for ui.upload

evnchn:
Fulfills https://github.com/zauberzeug/nicegui/discussions/4541

TL-DR: run `on_begin_upload` handler as soon as the server receive an upload, without waiting for the rest of the data to flow in. 

Ref https://github.com/zauberzeug/nicegui/discussions/4541#discussioncomment-12651783 on existing solution using `ui.upload().on('uploading', ...`: My solution should be slightly more accurate, in case network congestion means that WebSocket goes through but the upload requests gets stuck for some reason. 

---

Testing script: 

```py
from nicegui import ui

@ui.page("/")
def main_page():
    my_upload = ui.upload(on_begin_upload=lambda e: print(f"Upload started: {e}"), 
                          on_upload=lambda e: print(f"Upload: {e}"), 
                          on_multi_upload=lambda e: print(f"Multi upload: {e}"), 
                          on_rejected=lambda e: print(f"Rejected: {e}"))

    my_upload.on_begin_upload(lambda e: ui.notify(f"Upload started: {e}"))
    my_upload.on_upload(lambda e: ui.notify(f"Upload: {e}"))
    my_upload.on_multi_upload(lambda e: ui.notify(f"Multi upload: {e}"))
    my_upload.on_rejected(lambda e: ui.notify(f"Rejected: {e}"))

ui.run()
```

Throttle the speed in DevTools and you'll see it work. 

{93479A85-6820-4623-B27F-538CA66839B8}

### DIFF
--- a/nicegui/elements/upload.py
+++ b/nicegui/elements/upload.py
@@ -16,6 +16,7 @@ class Upload(DisableableElement, component='upload.js'):
                  max_file_size: Optional[int] = None,
                  max_total_size: Optional[int] = None,
                  max_files: Optional[int] = None,
+                 on_begin_upload: Optional[Handler[UiEventArguments]] = None,
                  on_upload: Optional[Handler[UploadEventArguments]] = None,
                  on_multi_upload: Optional[Handler[MultiUploadEventArguments]] = None,
                  on_rejected: Optional[Handler[UiEventArguments]] = None,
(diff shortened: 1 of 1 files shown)


falkoschindler:
Oh, you're right: There is only one POST request containing all files, so there is only one "begin" event. 👍🏻
Without looking closely into the code, I thought there would be a request per file.

evnchn:
Yes, I know the name is confusing... But if you refer to the core logic, it isn't.

Should we want to introduce this feature, we need to document the calling sequence **very clearly**!

evnchn:
Great! I didn't notice the documentation issue, either. You have sharp eyes!

Issue #4500, opened 2025-03-19
Task was destroyed but it is pending!


JudeMcbeath:
### Description

Im getting this error and have no idea where its come from. my app has had no issues for a while and i havnt changed anything:

```
2025-03-19 11:40:06.494 | task: <Task pending name='Task-21393' coro=<Event.wait() running at /usr/local/lib/python3.12/asyncio/locks.py:212> wait_for=<Future pending cb=[Task.task_wakeup()]>> |  
-- | -- | --
  |   | 2025-03-19 11:40:06.494 | 2025-03-19 11:40 - ERROR - Task was destroyed but it is pending! |  
  |   | 2025-03-19 11:40:06.494 | task: <Task pending name='Task-20721' coro=<Event.wait() running at /usr/local/lib/python3.12/asyncio/locks.py:212> wait_for=<Future pending cb=[Task.task_wakeup()]>> |  
  |   | 2025-03-19 11:40:06.494 | 2025-03-19 11:40 - ERROR - Task was destroyed but it is pending! |  
  |   | 2025-03-19 11:40:06.493 | task: <Task pending name='Task-21315' coro=<Event.wait() running at /usr/local/lib/python3.12/asyncio/locks.py:212> wait_for=<Future pending cb=[Task.task_wakeup()]>> |  
  |   | 2025-03-19 11:40:06.493 | 2025-03-19 11:40 - ERROR - Task was destroyed but it is pending!
```

evnchn:
### [Why is it important to provide a minimal reproducible example?](https://github.com/zauberzeug/nicegui/wiki/FAQs#why-is-it-important-to-provide-a-minimal-reproducible-example)

> When you're asking for help, it is important to provide a [minimal executable example (MRE)](https://en.wikipedia.org/wiki/Minimal_reproducible_example), if possible. Here is why:
> 
> By providing an MRE, you increase the chances of getting a quick and accurate response to your question. People are more likely to help when they can easily reproduce the problem and test their solutions.
> 
> It also helps you understand your problem better by isolating the issue and removing any unnecessary code. This can lead to a better understanding of the problem and potential solutions.
> 
> When creating an MRE, make sure to remove any code that is not directly related to the issue, while keeping the example complete and runnable. This includes imports, functions, and data. Keep the example as simple as possible while still demonstrating the problem.

---

So, need a code snippet, and also some console logs. Don't forget Python version and NiceGUI version

JudeMcbeath:
[...]

PR #4550, opened 2025-03-30
Avoid RuntimeError while iterating over bindable properties

### LINKED ISSUE #4419: Bindable Properties: RuntimeError: dictionary changed size during iteration
### Description

Hello,

currently we observe the following exception in different parts of our code:

```python
  [...]
  File "/opt/conda/envs/.../lib/python3.10/site-packages/nicegui/element.py", line 489, in delete
    self.parent_slot.parent.remove(self)
  File "/opt/conda/envs/.../lib/python3.10/site-packages/nicegui/element.py", line 481, in remove
    self.client.remove_elements(element.descendants(include_self=True))
  File "/opt/conda/envs/.../lib/python3.10/site-packages/nicegui/client.py", line 319, in remove_elements
    binding.remove(element_list)
  File "/opt/conda/envs/.../lib/python3.10/site-packages/nicegui/binding.py", line 203, in remove
    for (obj_id, name), finalizer in list(bindable_properties.items()):
```

It is triggered if ``.delete()`` for an UI element is called or ``.refresh()`` for a refreshable UI element, but it does not happen every time when the two methods are called. It is important to note that those elements do not have any bindable properties, but there are other elements in the application that use bindable properties.

The exception occurs from time to time if there are many users working with the same NiceGUI application. At the moment there is no way to reproduce the issue.

I noticed that there were some changes in 2.11.1 that are related to line 203: https://github.com/zauberzeug/nicegui/pull/4122
Thus this issue might be related to adding ``weakref``, but I am not sure. We have not observe the issue before updating our service from 2.9.1 to 2.11.1.

What I can not understand is why the code below from ``binding.py`` can actually throw ```RuntimeError: dictionary changed size during iteration```:

```python
for (obj_id, name), finalizer in list(bindable_properties.items()):
    if obj_id in object_ids:
        del bindable_properties[(obj_id, name)]
        finalizer.detach()
```
The error must happen while ``bindable_properties.items()`` is running, but this would also mean that there is another thread modifying ```bindable_properties``` while iterating over the dictionary. As far as I know NiceGUI is based on async code and I can not see how a context switch is possible during an ``.items()`` call.

I know that there is the possibility that ```bindable_properties``` is modified outside the for loop in line 178:
```python
bindable_properties.setdefault(key, weakref.finalize(owner, lambda: bindable_properties.pop(key, None)))
```

But still I can not understand why the lambda method might be called while iterating over ``bindable_properties.items()``

It would be great if someone could help to fix this issue. Many thanks in advance.

NiceGUI Version:
2.11.1

falkoschindler:
This PR tries to fix #4419 by disabling the GC while copying the dictionary of bindable properties. Calling `list()` iterates over the dictionary, which can case a "RuntimeError: dictionary changed size during iteration" if the garbage collector runs at the same time and causes the finalizers to pop items from the dictionary.

The PR is still a draft because there is another iteration over `bindable_properties` in the `_make_copyable` function. So technically there is still a chance for the RuntimeError.

1. So we might need to disable the GC once again in `_make_copyable`.
2. Or shouldn't the finalizers pop themselves, but add to a set of obsolete keys instead?
3. Alternatively we could try to use some kind of lock.

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import copyreg
 import dataclasses
-import multiprocessing
+import threading
 import time
 import weakref
 from collections import defaultdict
@@ -36,7 +36,7 @@ MAX_PROPAGATION_TIME = 0.01
 
 bindings: DefaultDict[Tuple[int, str], List] = defaultdict(list)
 bindable_properties: Dict[Tuple[int, str], weakref.finalize] = {}
-bindable_properties_lock = multiprocessing.Lock()
+bindable_properties_lock = threading.Lock()
 active_links: List[Tuple[Any, str, Any, str, Callable[[Any], Any]]] = []
 
 TC = TypeVar('TC', bound=type)
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import copyreg
 import dataclasses
-import gc
+import multiprocessing
 import time
 import weakref
 from collections import defaultdict
@@ -36,6 +36,7 @@ MAX_PROPAGATION_TIME = 0.01
 
 bindings: DefaultDict[Tuple[int, str], List] = defaultdict(list)
 bindable_properties: Dict[Tuple[int, str], weakref.finalize] = {}
+bindable_properties_lock = multiprocessing.Lock()
 active_links: List[Tuple[Any, str, Any, str, Callable[[Any], Any]]] = []
 
 TC = TypeVar('TC', bound=type)
@@ -180,7 +181,11 @@ class BindableProperty:
             return
         setattr(owner, '___' + self.name, value)
         key = (id(owner), str(self.name))
-        bindable_properties.setdefault(key, weakref.finalize(owner, lambda: bindable_properties.pop(key, None)))
+
+        def remove_bindable_property():
+            with bindable_properties_lock:
+                bindable_properties.pop(key, None)
+        bindable_properties.setdefault(key, weakref.finalize(owner, remove_bindable_property))
         _propagate(owner, self.name)
         if value_changed and self._change_handler is not None:
             self._change_handler(owner, value)
@@ -205,18 +210,11 @@ def remove(objects: Iterable[Any]) -> None:
         ]
         if not binding_list:
             del bindings[key]
-
-    gc_was_enabled = gc.isenabled()
-    if gc_was_enabled:
-        gc.disable()  # disable garbage collection to avoid finalizers to pop items while iterating
-    try:
+    with bindable_properties_lock:
         for (obj_id, name), finalizer in list(bindable_properties.items()):
             if obj_id in object_ids:
                 del bindable_properties[(obj_id, name)]
                 finalizer.detach()
-    finally:
-        if gc_was_enabled:
-            gc.enable()
 
 
 def reset() -> None:
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import copyreg
 import dataclasses
+import gc
 import time
 import weakref
 from collections import defaultdict
@@ -204,10 +205,18 @@ def remove(objects: Iterable[Any]) -> None:
         ]
         if not binding_list:
             del bindings[key]
-    for (obj_id, name), finalizer in list(bindable_properties.items()):
-        if obj_id in object_ids:
-            del bindable_properties[(obj_id, name)]
-            finalizer.detach()
+
+    gc_was_enabled = gc.isenabled()
+    if gc_was_enabled:
+        gc.disable()  # disable garbage collection to avoid finalizers to pop items while iterating
+    try:
+        for (obj_id, name), finalizer in list(bindable_properties.items()):
+            if obj_id in object_ids:
+                del bindable_properties[(obj_id, name)]
+                finalizer.detach()
+    finally:
+        if gc_was_enabled:
+            gc.enable()
 
 
 def reset() -> None:


falkoschindler:
No, the line in `_make_copyable`
```py
if (id(obj), attr_name) in bindable_properties:
```
doesn't iterate, so it can't cause such a RuntimeError.

Nevertheless, we want to try option 2 because it feels more robust and avoids different threads accessing the dictionary:

> Or shouldn't the finalizers pop themselves, but add to a set of obsolete keys instead?

chriswi93:
@falkoschindler 
The second option might introduce a race condition. At any place you have to clear the keys and there is the chance that the set of obsolete keys is modified at the same time from the finalizers thread.

An alternative way could be to create a child class for ``dict``:

```python
bindable_properties: Dict[Tuple[int, str], weakref.finalize] = LockedDict()
```

and override methods that should be thread safe:

```python
class LockedDict(dict):
    def items(..):
        ..
    def pop(..):
        ..
...
```
This way, you don't need to modify the code wherever ``bindable_properties`` is accessed.

falkoschindler:
Good point, @chriswi93! But implementing a `LockedDict` sounds like quite some overhead just for this purpose.
Maybe option 3 is more straightforward, simply using a lock when creating the list `list(bindable_properties.items())` and when popping elements `bindable_properties.pop(key, None)`.

falkoschindler:
I just pushed a new version with a lock instead of disabling the GC. The effect should be similar, but I find this code a bit more intuitive. Any counterarguments I've missed?

evnchn:
You might want to retitle the PR to be something more generic, if we are exploring more options besides disabling GC

chriswi93:
> I just pushed a new version with a lock instead of disabling the GC. The effect should be similar, but I find this code a bit more intuitive. Any counterarguments I've missed?

Looks good from my side 👍
Just one question: What is the reason to use lock from multiprocessing here? I would be quite suprised if the finalizer actually runs in another process. I guess this kind of lock has some unnecessary overhead due to interprocess communication.

falkoschindler:
@chriswi93 Oh, I didn't really think about it. I remembered that there is a `Lock` class in the multiprocessing package and simply used it. But you're right, it's probably better to use `Lock` from the threading package in this case.

rodja:
After releasing 2.14.0 and updating https://nicegui.io we noticed servers stopped responding after a while. After some git bisecting I come to the conclusion that this PR causes the error. I was able to reproduce it locally:

1. start documentation `./main.py`
2. open http://127.0.0.1:8080/documentation/section_binding_properties
3. reload multiple times in fast succession. After about 10-40 reloads the whole app becomes unresponsive.

We should quickly revert this PR and make a bugfix release. Or maybe even pull 2.14.0 from pip.

evnchn:
Also, just so we don't blow up prod twice: 

https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Lock

As it is a "a close analog of `threading.Lock`", it would also likely not work.

evnchn:
Iterate from the back? 

https://stackoverflow.com/a/65469649

Worst case we double-index, or the list got deleted so much in another thread that one of the thread runs out of list bounds, but at least it's more harmless?

polyipseity:
@evnchn summoned me privately here.

tl;dr There are two solutions. Choose one of them \(I have not tested them\):

1. Use `threading.RLock` instead of `threading.Lock`.
2. Lock-free approach: Don't use locks. The idea is you keep trying to convert `bindable_properties.items()` into a list until it succeeds. Change `list(bindable_properties.items())` to

```Python
bindable_properties_keys = None
while bindable_properties_keys is None:
  try:
    bindable_properties_keys = list(bindable_properties.items())
  except RuntimeError as e:
    if e.message != "dictionary changed size during iteration":
      raise
```

---

My reading of #4419 is that the GC may run while running `list(bindable_propeties.items())`. Specficially, the steps are:

1. `list(bindable_properties.items())` starts running.
2. The `list` constructor iterates through `bindable_properties.items()` one-by-one.
    - If no GC occurs while iterating through `bindable_properties.items()`, then everything works as expected.
    - A GC occurs while iterating through `bindable_properties.items()`. Finalizers may be invoked.
        1. In particular, our code registers a finalizer that is: `lambda: bindable_properties.pop(key, None))`. If this finalizer runs, it modifies the dictionary size.
        2. When the GC finishes, the `list` constructor continues to iterate, but notices the dictionary size changes, thus a `RuntimeError`.
3. `list(bindable_properties.items()` finshes.

It would be best to clarify what thread the above steps run in. This post <https://discuss.python.org/t/does-garbage-collection-run-in-a-separate-thread-in-free-threaded-python/74770/4> indicate that the GC can run in any thread, including the current thread iterating through `list(bindable_properties.items())`. \(It also seem to indicate GC may run in separate threads in the future, so let's also say we want to future-proof that.\)

The fix in this PR will fail if the GC runs on the same thread as `list(bindable_properties.items())`. Before you run `list(bindable_properties.items())`, you acquire the lock `bindable_properties_lock`. Then, if GC occurs during iteration on the same thread, and the above finalizer runs, then the same thread will try to acquire the lock `bindable_properties_lock` again in the finalizer. Since the lock is not reentrant, it deadlocks. 

This leads to the first solution: use an reentrant lock, which `threading.RLock` provides.

The second solution is noticing that if we get the `RuntimeError`, we can simply retry running `list(bindable_properties.items())` again. Ignoring the error does not affect the correctness of our program.

falkoschindler:
@polyipseity Thank you so much for two brilliant solutions and a a very insightful explanation of why this PR can cause deadlocks! Now we only have to choose between `RLock` and a while loop. The lock feels more robust, but I like the while loop for its locality. I think I lean towards the lock, because it sounds like the "right" solution and we can re-use this PR and only add a single character.

evnchn:
@falkoschindler see if the following solution is better: 

- Use `RLock`
- Inside the `RLock`, do the while-loop thing
- In debugging phases, warn if the while loop ran more than once. If so, the `RLock` must have not worked, and we resort to just a while-loop.

chriswi93:
Quite confusing that the gc sometimes runs in the same thread and sometimes not.
I would also vote for the ``threading.RLock`` solution.

falkoschindler:
I just created PR #4615 with an `RLock`. The problem as described in https://github.com/zauberzeug/nicegui/pull/4550#issuecomment-2788275663 doesn't seem to occur.

evnchn:
But my main concern is: 

According to https://superfastpython.com/asyncio-use-threading-lock/#Example_of_threading_Reentrant_Lock_in_Asyncio, since asyncio runs on a single thread, and a reentrant lock always allow the same thread to acquire it again, that basically it is equivalent to `if True`, and the `RLock` does no meaningful work. 

Thats why I suggest, besides testing for deadlocks, we should also test for if it still throws the error. And that's why I suggest combine it with a try-except while loop with logged warnings to evaluate the effectiveness of `RLock`.

falkoschindler:
@polyipseity's suggestion in https://github.com/zauberzeug/nicegui/pull/4550#issuecomment-2790006973 sounded so nice and simple...
> 1. Use `threading.RLock` instead of `threading.Lock`.

But the more I think about it, the more I believe you're right. Whatever is triggering the finalizer (GC or simple reference counting) should run in the same thread, so `RLock` shouldn't help.

I would love to have a reproduction so that we can simply test it...

falkoschindler:
Great idea! Unfortunately though, rapid reloading doesn't cause a RuntimeError on main. That's probably because we use the lock far more often (allowing for deadlocks) than we would actually need to avoid conflicts with the GC.

evnchn:
Perhaps mess around with the GC threshold via `gc.set_threshold`? Lowering the threshold would make the automatic (aka the ones which run while we're in the middle of code execution) GC occur more frequently, showing the defects in our code. 

Sidenote: running `gc.collect` probably won't work, since that would be blocking the code until GC's over.

polyipseity:
Ah, you are right. Using `RLock` only solves the new problem of deadlocking when using `Lock`, but does not resolve the original problem. \(Async has nothing to do with it.\) I think solution 2 is still valid though.

Alternatively \(solution 3\), the finalizer simply removes the key associated with `owner` from `bindable_properties` when `owner` dies:

```Python
        def remove_bindable_property():
            with bindable_properties_lock:
                bindable_properties.pop(key, None)
        bindable_properties.setdefault(key, weakref.finalize(owner, remove_bindable_property))
```

Wouldn't it be better to use a [`WeakValueDictionary`](https://docs.python.org/3/library/weakref.html#weakref.WeakValueDictionary) directly for `bindable_properties` instead, and store the `owner` as the value? Then you do not need to set up the finalizers yourself and manually detach the finalizers in `remove`. Iteration should be safe: <https://stackoverflow.com/a/52693673>.

evnchn:
For those who want to join the team and debug this:

From what I have learnt while working on https://github.com/zauberzeug/nicegui/issues/4500, GC issues are much harder to replicate on Windows than Linux.

That pretty much rules out me, since I use Windows exclusively for NiceGUI contribution. 🤡

PR #4557, opened 2025-04-01
Server-Side Rendering by invoking NodeJS, Vue and Quasar for the HTML

evnchn:
https://github.com/zauberzeug/nicegui/discussions/4554

TL-DR: Server-side rendering makes the NiceGUI pages include valid HTML elements (instead of a string in JavaScript), enabling 
1. Browsers to show the elements much earlier in the page parsing process without waiting for the JavaScript to execute, 
2. Better SEO (search engine optimization) so that pages don't show up as ["Connection lost. Trying to reconnect..."](https://github.com/zauberzeug/nicegui/discussions/4506) for Google, Bing, Yandex, the alikes. 

---

Code is rather obvious: 

* `ssr.py` contains the logic of getting the HTML elements
* The chain of serving HTTP response (`client.py`, `index.html`, `nicegui.js` is reworked to call `Vue.createSSRApp` if elements are successfully yielded from SSR)
* For async, no choice but to change `client.build_response` to async. 

---

Remaining concerns: 

* Should separate this to a separate package, think: `pip install nicegui[ssr]` since not everyone wants +20MB just for the NodeJS instance
* Behaviour may not be acceptable for all pages, since +200ms processing time. Opt-in via `@ui.page` seems wiser. 
  * (in another PR) implement caching of the HTTP response? but the page will have outdated content and updates must be served via updating calls after `await ui.context.client.connected()`
* For SEO purposes, perhaps SSR is only done when the agent is known to be a web scraper? 
* Tried few hours at resolving hydration mismatch, no success!

{D2226007-C4B3-4DA6-B298-85213AF5D1D0}

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -25,6 +25,8 @@ from .observables import ObservableDict
 from .outbox import Outbox
 from .translations import translations
 from .version import __version__
+from .ssr import get_ssr_output
+from .run import io_bound
 
 if TYPE_CHECKING:
     from .page import page
@@ -137,17 +139,22 @@ class Client:
             'next_message_id': self.outbox.next_message_id,
         }
         vue_html, vue_styles, vue_scripts, imports, js_imports = generate_resources(prefix, self.elements.values())
+        elements_replaced = elements.replace('&', '&amp;') \
+                                    .replace('<', '&lt;') \
+                                    .replace('>', '&gt;') \
+                                    .replace('`', '&#96;') \
+                                    .replace('$', '&#36;')
+
+        def get_my_ssr_output():
+            return get_ssr_output(elements_replaced)
+        results_ssr = await io_bound(get_my_ssr_output)
         return templates.TemplateResponse(
             request=request,
             name='index.html',
             context={
                 'request': request,
                 'version': __version__,
-                'elements': elements.replace('&', '&amp;')
-                                    .replace('<', '&lt;')
-                                    .replace('>', '&gt;')
-                                    .replace('`', '&#96;')
-                                    .replace('$', '&#36;'),
+                'elements': elements_replaced,
                 'head_html': self.head_html,
                 'body_html': '<style>' + '\n'.join(vue_styles) + '</style>\n' + self.body_html + '\n' + '\n'.join(vue_html),
                 'vue_scripts': '\n'.join(vue_scripts),
@@ -166,6 +173,7 @@ class Client:
                 'socket_io_js_query_params': socket_io_js_query_params,
                 'socket_io_js_extra_headers': core.app.config.socket_io_js_extra_headers,
                 'socket_io_js_transports': core.app.config.socket_io_js_transports,
+                'results_ssr': results_ssr,
             },
             status_code=status_code,
             headers={'Cache-Control': 'no-store', 'X-NiceGUI-Content': 'page'},
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -124,7 +124,7 @@ class Client:
     def __exit__(self, *_) -> None:
         self.content.__exit__()
 
-    def build_response(self, request: Request, status_code: int = 200) -> Response:
+    async def build_response(self, request: Request, status_code: int = 200) -> Response:
         """Build a FastAPI response for the client."""
         self.outbox.updates.clear()
         prefix = request.headers.get('X-Forwarded-Prefix', request.scope.get('root_path', ''))
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -62,8 +62,8 @@ Client.auto_index_client = Client(page('/'), request=None).__enter__()  # pylint
 
 
 @app.get('/')
-def _get_index(request: Request) -> Response:
-    return Client.auto_index_client.build_response(request)
+async def _get_index(request: Request) -> Response:
+    return await Client.auto_index_client.build_response(request)
 
 
 @app.get(f'/_nicegui/{__version__}' + '/libraries/{key:path}')
@@ -147,7 +147,7 @@ async def _exception_handler_404(request: Request, exception: Exception) -> Resp
     log.warning(f'{request.url} not found')
     with Client(page(''), request=request) as client:
         error_content(404, exception)
-    return client.build_response(request, 404)
+    return await client.build_response(request, 404)
 
 
 @app.exception_handler(Exception)
@@ -155,7 +155,7 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
     log.exception(exception)
     with Client(page(''), request=request) as client:
         error_content(500, exception)
-    return client.build_response(request, 500)
+    return await client.build_response(request, 500)
 
 
 @sio.on('handshake')
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -138,7 +138,7 @@ class page:
             if isinstance(result, Response):  # NOTE if setup returns a response, we don't need to render the page
                 return result
             binding._refresh_step()  # pylint: disable=protected-access
-            return client.build_response(request)
+            return await client.build_response(request)
 
         parameters = [p for p in inspect.signature(func).parameters.values() if p.name != 'client']
         # NOTE adding request as a parameter so we can pass it to the client in the decorated function
--- a/nicegui/ssr.py
+++ b/nicegui/ssr.py
@@ -1,0 +1,283 @@
+import json
+from pathlib import Path
+import subprocess
+from nodejs_wheel import node, npm
+
+import uuid
+
+cwd = Path(__file__).parent
+
+print("cwd", cwd)
+
+package_json = cwd / 'package.json'
+
+quasar_path = cwd / 'node_modules' / 'quasar' / 'dist' / 'quasar.server.prod.js'
+
+
+def write_js_script(uuid: str, elements_string: str) -> None:
+    """Write the JavaScript script to a file."""
+    script = """
+// this runs in Node.js on the server.
+import { createSSRApp, withModifiers, withKeys, resolveComponent, h } from 'vue'
+// Vue's server-rendering API is exposed under `vue/server-renderer`.
+import { renderToString } from 'vue/server-renderer'
+// Quasar is a Vue.js framework for building responsive web applications.
+// @ts-ignore
+import { Quasar } from 'quasar'
+
+const True = true;
+const False = false;
+const None = undefined;
+
+const loaded_libraries = new Set();
+const loaded_components = new Set();
+
+function parseElements(raw_elements) {
+  return JSON.parse(
+    raw_elements
+      .replace(/&#36;/g, "$")
+      .replace(/&#96;/g, "`")
+      .replace(/&gt;/g, ">")
+      .replace(/&lt;/g, "<")
+      .replace(/&amp;/g, "&")
+  );
+}
+
+function replaceUndefinedAttributes(element) {
+  element.class ??= [];
+  element.style ??= {};
+  element.props ??= {};
+  element.text ??= null;
+  element.events ??= [];
+  element.update_method ??= null;
+  element.component ??= null;
+  element.libraries ??= [];
+  element.slots = {
+    default: { ids: element.children || [] },
+    ...(element.slots ?? {}),
+  };
+}
+
+function renderRecursively(elements, id) {
+  const element = elements[id];
+  if (element === undefined) {
+    return;
+  }
+
+  // @todo: Try avoid this with better handling of initial page load.
+  if (element.component) loaded_components.add(element.component.name);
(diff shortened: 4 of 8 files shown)


evnchn:
Appear to have some issues with accidentally rolling back #4507 in https://github.com/zauberzeug/nicegui/pull/4557/commits/1f9d4dfe386353c505b9489969091742e6a34b70
Merged properly now, I think...

evnchn:
Note the following: 
1. Hydration mismatch is caused by having many `style=""` and `class=""` in the rendering. Generally minor, and I can already think of a fix. Too tired to do it now, though
2. This PR is blocked by caching which is blocked by #4552. There is nothing wrong with this PR.

evnchn:
Timeline wise, if we want to do this, this will be in 3.X

Similar reason to https://github.com/zauberzeug/nicegui/pull/4552#issuecomment-2798196210

rodja:
This is a complex feature and maybe we find someone who want's to maintain it? As a external package (`pip install nicegui-ssr`), we only need to allow to hook into the NiceGUI internals.

evnchn:
@rodja that's an excellent idea. 

- feasible since the server-side rendering is just a single function, so we can register it like `app.register_prerender`
- doesn't even need to modify the NiceGUI package much. No need `nicegui[ssr]` optional package
- enables third-party to develop their own renderer which does a much better job than me
- for simply fulfilling SEO, one can write a dummy prerenderer which just outputs text that the search engine like
- change is minor enough that it can reasonably be slipped into a 2.X release

What do you think? I'm getting excited 🤩

evnchn:
It might just get slightly ugly, but it may be possible to achieve this with nothing but a Middleware. 

Let's see when I get back home.

evnchn:
Unfortunately, the current SSR approach (with / without modification to NiceGUI) has **terrible** compatibility with additional NiceGUI components. 

This is because, such components are expected to be imported by the browser, where the path has been created via `compute_key`. It is not possible to easily retrieve such files for the easy import of the server-side Node.js instance. Moreover, stuff such as `import { loadResource } from "../../static/utils/resources.js";` straight-up breaks, since it's 1 parent directory not 2 parent directory in terms of local file. 

With that issue, many pages are awfully broken visually once the hydration is completed. 

Without extensive effort, I cannot forsee this being into NiceGUI any time soon. Therefore, I am closing this PR for now. Will keep the branch in case anyone wants to try their hands on it. 

---

`ssr.py`

```py
import json
from pathlib import Path
import subprocess
from nodejs_wheel import node, npm

import uuid

cwd = Path(__file__).parent

print("cwd", cwd)

package_json = cwd / 'package.json'

quasar_path = cwd / 'node_modules' / 'quasar' / 'dist' / 'quasar.server.prod.js'


def write_js_script(uuid: str, elements_string: str) -> None:
    """Write the JavaScript script to a file."""
    script = """
// this runs in Node.js on the server.
import { createSSRApp, withModifiers, withKeys, resolveComponent, h } from 'vue'
// Vue's server-rendering API is exposed under `vue/server-renderer`.
import { renderToString } from 'vue/server-renderer'
// Quasar is a Vue.js framework for building responsive web applications.
// @ts-ignore
import { Quasar } from 'quasar'

const True = true;
const False = false;
const None = undefined;

const loaded_libraries = new Set();
const loaded_components = new Set();

function parseElements(raw_elements) {
  return JSON.parse(
    raw_elements
      .replace(/&#36;/g, "$")
      .replace(/&#96;/g, "`")
      .replace(/&gt;/g, ">")
      .replace(/&lt;/g, "<")
      .replace(/&amp;/g, "&")
  );
}

function replaceUndefinedAttributes(element) {
  return; 
  element.class ??= [];
  element.style ??= {};
  element.props ??= {};
  element.text ??= null;
  element.events ??= [];
  element.update_method ??= null;
  element.component ??= null;
  element.libraries ??= [];
  element.slots = {
    default: { ids: element.children || [] },
    ...(element.slots ?? {}),
  };
}

function renderRecursively(elements, id) {
  const element = elements[id];
  if (element === undefined) {
    return;
  }

  // @todo: Try avoid this with better handling of initial page load.
  if (element.component) loaded_components.add(element.component.name);
  if (element.libraries) element.libraries.forEach((library) => loaded_libraries.add(library.name));

  const props = {
    id: "c" + id,
    ref: "r" + id,
    key: id, // HACK: workaround for #600 and #898
    class: (element.class || []).join(" ") || undefined,
    // style: element.style ? Object.entries(element.style).reduce((str, [p, val]) => `${str}${p}:${val};`, "") || undefined : undefined,
    ...element.props,
  };
  if (element.style) {
    props.style = Object.entries(element.style).reduce((str, [p, val]) => `${str}${p}:${val};`, "") || undefined;
  }
  Object.entries(props).forEach(([key, value]) => {
    if (key.startsWith(":")) {
      try {
        try {
          props[key.substring(1)] = new Function(`return (${value})`)();
        } catch (e) {
          props[key.substring(1)] = eval(value);
        }
        delete props[key];
      } catch (e) {
        console.error(`Error while converting ${key} attribute to function:`, e);
      }
    }
  });
  (element.events || []).forEach((event) => {
    let event_name = "on" + event.type[0].toLocaleUpperCase() + event.type.substring(1);
    event.specials.forEach((s) => (event_name += s[0].toLocaleUpperCase() + s.substring(1)));

    let handler;
    if (event.js_handler) {
      handler = eval(event.js_handler);
    } else {
      handler = (...args) => {
        const emitter = () =>
          window.socket?.emit("event", {
            id: id,
            client_id: window.clientId,
            listener_id: event.listener_id,
            args: stringifyEventArgs(args, event.args),
          });
        const delayed_emitter = () => {
          if (window.did_handshake) emitter();
          else setTimeout(emitter, 10);
        };
        throttle(delayed_emitter, event.throttle, event.leading_events, event.trailing_events, event.listener_id);
        if (element.props["loopback"] === False && event.type == "update:modelValue") {
          element.props["model-value"] = args;
        }
      };
    }

    handler = withModifiers(handler, event.modifiers);
    handler = event.keys.length ? withKeys(handler, event.keys) : handler;
    if (props[event_name]) {
      props[event_name].push(handler);
    } else {
      props[event_name] = [handler];
    }
  });
  const slots = {};
  const element_slots = {
    default: { ids: element.children || [] },
    ...element.slots,
  };
  Object.entries(element_slots).forEach(([name, data]) => {
    slots[name] = (props) => {
      const rendered = [];
      if (data.template) {
        rendered.push(
          h(
            {
              props: { props: { type: Object, default: {} } },
              template: data.template,
            },
            {
              props: props,
            }
          )
        );
      }
      const children = data.ids.map((id) => renderRecursively(elements, id));
      if (name === "default" && element.text !== null) {
        children.unshift(element.text);
      }
      return [...rendered, ...children];
    };
  });
  return h(resolveComponent(element.tag), props, slots);
}

const elements = ELEMENTS_GO_HERE;

Object.entries(elements).forEach(([_, element]) => replaceUndefinedAttributes(element));

const app = createSSRApp({
  data() {
    return {
      elements,
    };
  },
  render() {
    return renderRecursively(elements, 0);
  },
});

// package.json
//    "type": "module",
//    "imports": {
//        "#root/*.js": "./*.js"
//    }

// Hmm. even the import is failing
import { default as colors } from "#root/.venv/Lib/site-packages/nicegui/elements/colors.js";
app.component("nicegui-colors", colors);

app.use(Quasar, {});

renderToString(app).then((html) => {
  console.log(html)
})
"""
    script = script.replace('ELEMENTS_GO_HERE', elements_string, 1)
    with open(cwd / f'ssr_{uuid}.js', 'w', encoding='utf-8') as file:
        file.write(script)


def delete_js_script(uuid: str) -> None:
    """Delete the JavaScript script file."""
    script_path = cwd / f'ssr_{uuid}.js'
    if script_path.exists():
        script_path.unlink()
        print(f'Deleted {script_path}')
    else:
        print(f'{script_path} does not exist')


def ensure_npm_init() -> None:
    """Check if npm is initialized in the current directory."""
    if package_json.exists():
        print('npm already initialized')
    else:
        print('npm not initialized')
        npm(['init', '-y'], cwd=cwd)


def check_npm_install() -> None:
    """Check if npm packages are installed in the current directory."""
    ensure_npm_init()  # make a call to the lower-level check

    def npm_install() -> None:
        npm(['install', 'vue@3.4.38', 'quasar@2.16.9'], cwd=cwd)

    with open(package_json, 'r') as file:
        package_data = json.load(file)
    for package in ['vue', 'quasar']:
        if package in package_data.get('dependencies', {}):
            print(f'{package} already installed')
        else:
            print("Did not install the required packages")
            npm_install()
            break

    if not quasar_path.exists():
        print('quasar.server.prod.js does not exist. This is strange but we will try again.')
        npm_install()

        if not quasar_path.exists():
            print('quasar.server.prod.js still does not exist. This is a critical error.')
            raise FileNotFoundError(f'{quasar_path} does not exist. Please check your installation.')


def ensure_type_module() -> None:
    """Check if the type module is set in package.json."""
    check_npm_install()  # make a call to the lower-level check

    with open(package_json, 'r') as file:
        package_data = json.load(file)
    if 'type' in package_data:
        print('Type module already set')
    else:
        print('Setting type module')
        package_data['type'] = 'module'
        with open(package_json, 'w') as file:
            json.dump(package_data, file, indent=4)


def ensure_patch_quasar() -> None:
    """Check if quasar is patched."""
    ensure_type_module()  # make a call to the lower-level check

    with open(quasar_path, 'r') as file:
        quasar_data = file.read()
    if '// modified by ssr.py' in quasar_data:
        print('quasar.server.prod.js already patched')
    else:
        print('Patching quasar.server.prod.js')
        quasar_data = quasar_data.replace(
            'e.req.headers["user-agent"]||e.req.headers["User-Agent"]', 'e.req?.headers["user-agent"]||e.req?.headers["User-Agent"]')
        quasar_data = quasar_data.replace('e,t={},o', 'e,t={},o={}')
        quasar_data += '\n// modified by ssr.py'
        with open(quasar_path, 'w') as file:
            file.write(quasar_data)


def get_ssr_output(elements_string: str) -> str:
    """Get the SSR output."""
    unique_id = str(uuid.uuid4())  # Generate a random UUID
    write_js_script(unique_id, elements_string)
    result = node([f'ssr_{unique_id}.js'], cwd=cwd, return_completed_process=True,
                  stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    if result.stderr:
        print("Error:", result.stderr.decode('utf-8').strip())
    # print(result)
    output = result.stdout.decode('utf-8').strip()
    delete_js_script(unique_id)
    return output

ensure_patch_quasar()

if __name__ == '__main__':
    elements_input = input('Enter elements string: ')
    print(get_ssr_output(elements_input))

```

Middleware: 

```py
from ssr import get_ssr_output
from nicegui.run import io_bound

class PrintHTMLMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request, call_next):
        response = await call_next(request)
        chunks = []
        async for chunk in response.body_iterator:
            chunks.append(chunk)
        response_body = b''.join(chunks)

        # Patch the HTML response to include SSR output, if possible

        if response_body.startswith(b'<!DOCTYPE html>'):
            body_text = response_body.decode('utf-8')
            # find content after <script type="module">
            _, _, script_section = body_text.partition('<script type="module">')
            script_content, _, _ = script_section.rpartition('</script>')
            script_content = script_content.strip()
            first_line_of_script = script_content.split('\n')[0].strip()
            # print('First line of script:', first_line_of_script)
            # find the input inside createApp(), matching as much as possible
            _, _, create_app_section = first_line_of_script.partition('createApp(')
            create_app_section, _, _ = create_app_section.rpartition(')')
            create_app_section = create_app_section.strip()
            create_app_section += ')'
            # print('createApp section:', create_app_section)
            def get_my_ssr_output():
                return get_ssr_output(create_app_section)
            elements_output = await io_bound(get_my_ssr_output) # get_ssr_output(create_app_section)
            # print('elements_output:', elements_output)

            if elements_output:
                # find <div id="app"></div>, put the elements_output inside it, as well as a script which sets window.isSSR = true
                body_text = body_text.replace('<div id="app"></div>', f'<div id="app">{elements_output}</div><script>window.isSSR = true;console.log("SSR");</script>')
                response_body = body_text.encode('utf-8')

        # Patch nicegui.js to use createSSRApp instead of createApp if window.isSSR is true

        # or if we find `return (app = Vue.createApp({``
        # replace with `const createVueApp = window.isSSR ? Vue.createSSRApp : Vue.createApp; return (app = createVueApp({``
        elif b'return (app = Vue.createApp(' in response_body:
            response_body = response_body.replace(b'return (app = Vue.createApp(', b'const createVueApp = window.isSSR ? Vue.createSSRApp : Vue.createApp; return (app = createVueApp(')

        return Response(content=response_body, status_code=response.status_code, 
            headers=dict(response.headers), media_type=response.media_type)

app.add_middleware(PrintHTMLMiddleware)
```

PR #4571, opened 2025-04-05
popup if socket.io message expected to be too long

### LINKED ISSUE #3410: Unpredictable Behavior of `ui.textarea` with Over a Million Characters
#### Description
When `ui.textarea` contains more than a million characters, it starts to behave unpredictably without any visible warnings.

#### Steps to Reproduce
To demonstrate the issue from different perspectives, I wrote the following code:

```python
from nicegui import ui

@ui.page('/')
def main():
    text_area = ui.textarea('Type something').props('rounded outlined input-class=mx-3') \
                .classes('w-full self-center')

    text_area.value = '123' + 'A'*1_000_000

    len_label = ui.label('Length').classes('mt-2')
    
    def text_area_length():
        print('Length:', len(text_area.value))
        len_label.set_text(f'Length: {len(text_area.value)}')
        print('first_character:', text_area.value[0])

    def text_area_delete_first():
        text_area.value = text_area.value[1:]
        text_area_length()

    ui.button('Calculate', on_click=text_area_length)
    ui.button('Delete first character', on_click=text_area_delete_first)

    text_area_length()

ui.run(title='Text Input Max')
```

When you run this code, it creates a textarea filled with over a million characters, plus three more to make the behavior more evident in different scenarios.

1. If you click the `Calculate` button immediately after launching (without interacting with the `ui.textarea`), everything is calculated correctly, and the console prints the appropriate values.
2. If you add or delete a character using the `ui.textarea` (e.g., delete the digit `1` at the beginning or add a new character), the `Calculate` button will no longer calculate the new value of `text_area` correctly. The code executes without errors, and you see the print output, but it shows the wrong number of characters and the first character is also incorrect, as if the connection between the frontend and backend is lost.
3. To further investigate, I created a button to delete the first character without using the cursor to interact with the `ui.textarea`. In this case, everything works as expected, the counter decreases, and the first characters are read correctly.

#### Additional Investigation
To see if this is a Quasar issue, I added character counting to their textarea code:
https://codepen.io/Xtreemrus/pen/XWLNPyY?editors=101

```HTML
<!--
Forked from:
https://quasar.dev/vue-components/input#example--textarea
-->
<div id="q-app" style="min-height: 100vh;">
  <div class="q-pa-md" style="max-width: 300px">
    <q-input
      v-model="text"
      filled
      type="textarea"
    ></q-input>
    <div class="q-mt-md">
      <p>Calculate len: {{ characterCount }}</p>
      <p>first_character: {{ firstCharacter }}</p>
    </div>
  </div>
</div>
```

JS part:
```Javascript
const { ref, computed } = Vue

const app = Vue.createApp({
  setup () {
    const text = ref('')

    const characterCount = computed(() => text.value.length)
    
    const firstCharacter = computed(() => {
      if (text.value.length > 0) {
        return text.value[0]
      } else {
        return 'No symbols'
      }
    })

    return {
      text,
      characterCount,
      firstCharacter
    }
  }
})

app.use(Quasar, { config: {} })
app.mount('#q-app')
```

In this case, working with a million characters is handled correctly.

#### Impact
A couple of years ago, cases requiring a million characters in a textarea were rare. However, with the development of LLMs, users are inserting increasingly large chunks of information into input fields. One million characters are approximately 250,000 tokens, which is now a feasible amount.

#### Difficulty
The challenge with this bug is that no visible errors occur, neither on the user side nor the developer side. There are no exceptions thrown, making it hard to find a workaround.

#### Environment
- Browser: Chrome
- NiceGUI version: 1.4.30

evnchn:
fixes #3410 

TL-DR: Nobody in the chain of Python-EngineIO -> Python-SocketIO -> SocketIO.js is taking up the responsibility of limiting the message length in a verbose manner. NiceGUI is taking up the responsibility, then. 

Notable changes: 
* Refactor CSS to share styling between the existing disconnect popup and the new too-long-message popup
* Add popup into `index.html`
* JS estimate message length via trial-and-error, make popup show for 5 seconds after the last too-long-message

To-do and considerations: 
* Internationalization similar to #4324 
* Will there be performance penalty if we stringify once, and then Socket.IO swiftly stringifys again? 

Result: 
{12461EE6-429C-411B-BCFA-9C327F550267}

### DIFF
--- a/nicegui/static/nicegui.css
+++ b/nicegui/static/nicegui.css
@@ -265,8 +265,8 @@ h6.q-timeline__title {
   outline: none;
 }
 
-/* connection popup */
-#popup {
+/* connection popup & too-long-message popup */
+.nicegui-error-popup {
   position: fixed;
   bottom: 0;
   left: 0;
@@ -282,13 +282,13 @@ h6.q-timeline__title {
   pointer-events: none;
   z-index: 10000;
 }
-.body--dark #popup {
+.body--dark .nicegui-error-popup {
   background-color: black;
 }
-.body--light #popup {
+.body--light .nicegui-error-popup {
   background-color: white;
 }
-#popup[aria-hidden="true"] {
+.nicegui-error-popup[aria-hidden="true"] {
   opacity: 0;
   visibility: hidden;
 }
@@ -297,10 +297,14 @@ h6.q-timeline__title {
   visibility: visible;
   transition-delay: 2000ms;
 }
-#popup > span:first-child {
+#popup_toolongmessage[aria-hidden="false"] {
+  opacity: 1;
+  visibility: visible;
+}
+.nicegui-error-popup > span:first-child {
   font-weight: bold;
 }
-#popup > span:first-child::before {
+.nicegui-error-popup > span:first-child::before {
   content: "⚠️";
   position: absolute;
   left: 1.5em;
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -178,13 +178,36 @@ function renderRecursively(elements, id) {
       handler = eval(event.js_handler);
     } else {
       handler = (...args) => {
-        const emitter = () =>
-          window.socket?.emit("event", {
+        const emitter = () => {
+          if (!window.socket) {
+            return;
+          }
+          const payload = {
             id: id,
             client_id: window.clientId,
             listener_id: event.listener_id,
             args: stringifyEventArgs(args, event.args),
-          });
+          };
+          const expected_socketio_payload = [
+            "event",
+            payload,
+          ]
+          console.log("Length of expected_socketio_payload: ", JSON.stringify(expected_socketio_payload).length + 2);
+
+          if (JSON.stringify(expected_socketio_payload).length + 2 > window.socket.io.engine.maxPayload) {
+            console.log("Payload size exceeds the maximum allowed limit.");
+            if (window.tooLongMessageTimer) {
+              clearTimeout(window.tooLongMessageTimer);
+            }
+            const popup = document.getElementById("popup_toolongmessage");
+            popup.ariaHidden = false;
+            window.tooLongMessageTimer = setTimeout(() => {
+              popup.ariaHidden = true;
+            }, 5000);
+            return;
+          }
+          window.socket.emit("event", payload);
+        };
         const delayed_emitter = () => {
           if (window.did_handshake) emitter();
           else setTimeout(emitter, 10);
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -40,9 +40,13 @@
     {{ body_html | safe }}
 
     <div id="app"></div>
-    <div id="popup" aria-hidden="true">
+    <div id="popup" class="nicegui-error-popup" aria-hidden="true">
       <span>{{ translations.connection_lost }}</span>
       <span>{{ translations.trying_to_reconnect }}</span>
+    </div>
+    <div id="popup_toolongmessage" class="nicegui-error-popup" aria-hidden="true">
+      <span>Message too long</span>
+      <span>The message is longer than what the server can handle.</span>
     </div>
     <script type="module">
       const app = createApp(parseElements(String.raw`{{ elements | safe }}`), {


evnchn:
Preliminary translation support https://github.com/zauberzeug/nicegui/pull/4571/commits/f9fc3cdcdc68e25b5ac8b1ed1cf713f96058aefb

Not sure if the strings are correct, though!

{3F063AE9-CF00-4DDB-9E63-745B0B4A9DC8}

rodja:
Would it not be much better to implement chunking instead of showing an error?

evnchn:
Sending 1MB already makes the server markedly slow even on localhost connection, and sometimes the NiceGUI page reloads despite staying under the limit, just because multiple events are sent without a gap in them (I suspect a blocking call somewhere?). Showing an error seems like the correct way out. 

Also it's not my idea, but @falkoschindler's. Though I stand by his idea, I'll say. 

https://github.com/zauberzeug/nicegui/issues/3410#issuecomment-2779202130

> I'm not so sure we're there yet. Engineio might not be responsible, but maybe the Socket.IO client? It is responsible for serializing and submitting messages. Therefore it should be able to _judge the total message size and report an error if it exceeds a limit._
> 
> Anyway, _would we really want to implement segmenting? Do we really want to allow sending MBs of data back an forth?_ I still think the primary goal is to inform the user about the message not being transmitted. And it would be great to keep the connection alive.

evnchn:
Also, the goal of the limit is to avoid excessive memory use, right? 

If we chunk the sending, not only do we not save memory (since we do not stream the response from the Socket.IO nor can we (I think)), we incur additional overhead as the server has to put, somewhere, the segments and put it back together in-order (not sure if Socket.IO / WebSocket ensures order... I've seen Long-Polling go out of order I think?)

So that's kind of besides the point. 

If there are need to send large data, for example a novel component which throws tons of data across to the server, perhaps that component can handle the chunking itself?

rodja:
Ah sorry. Have not seen the updates from #3410 yet. Let's continue over there.

evnchn:
Wanted to start work on the segmented message transmission as outlined in https://github.com/zauberzeug/nicegui/issues/3410#issuecomment-2780203933

The code may have significant merge conflicts with this PR, so I think either (1) this PR does both, or (2) begin work after this PR goes through.

falkoschindler:
Ok, after talking to Rodja once again I'm now convinced that we need some kind of popup either way - whether we chunk messages or not.

But I'm sceptical about the additional serialization. Can't we find a way to either configure the Socket.IO client to adhere to the max message size, raise an error, call a callback or use a custom serialization? Maybe we find related issues/discussions over at https://github.com/socketio/socket.io/?

evnchn:
@falkoschindler I absolutely can't find any public methods for what you want to achieve. Maybe I am dumb, which I am, but Socket.IO isn't much intended for people extending its functionality I believe. 

How about this: 

```js
const originalWriteFunction = window.socket.io.engine.write;
window.socket.io.engine.write = function(msg, options, fn) {
  console.log('Message:', msg, msg.length); // log the message before executing the main code
  if (msg.length > 1000000) {console.log("Too long!");return}
  return originalWriteFunction.call(this, msg, options, fn);
};
```

Then it intercepts the call pretty low-level.

evnchn:
But then, if you do this approach, by the time you intercepted the call, it's several layers deep into the NiceGUI's logic, and it's difficult to then do any kind of error-handling, retry with smaller message, etc. 

Works for websocket only, not for polling: 

```
const originalWriteFunction = window.socket.io.engine.transport.ws.send;
window.socket.io.engine.transport.ws.send = function(msg) {
  console.log('Message:', msg, msg.length); // log the message before executing the main code
  if (msg.length > 1000000) {console.log("Too long!");return}
  return originalWriteFunction.call(this, msg);
};
```

More accurate than the above since it accounts for any binary attachment, namespace, as well as the 1-character message type.

falkoschindler (changes requested):
Finally getting back to this PR... Your proposal from https://github.com/zauberzeug/nicegui/pull/4571#issuecomment-2784435335 looks really good to me! It allows us to show a popup precisely when exceeding `max_http_buffer_size`, without the need to serialize _every_ message twice. We could also use `emitEvent` to let the server handle the problem. But a `.nicegui-error-popup` is probably good enough for now.

Do you agree, @evnchn? Can you update the PR accordingly? Thanks!

evnchn:
I think our tapping into socketio is a bit sketchy, but it is something we have no choice but to do. 

To settle the sketchiness, we add tests, and we make sure to triple-check if the tests pass when we bump socketio.

evnchn:
https://github.com/zauberzeug/nicegui/pull/4571/commits/7628e446db614c10891cfe64eb4d4ea890fa6b10: introduced since prerender have error: https://github.com/evnchn/nicegui/actions/runs/18908339894/job/53972071093

This also means that the popup won't work when it is initially loaded via pre-render though 🤔

falkoschindler:
@evnchn I'm pretty close to accepting this PR... Just one more question:

- [x] What exactly is the reasoning behind a limit of 1_000_000 - 100? Is it completely empirical, or is it derived from some websocket specification?

A change event message for an empty input value has length 140 (for a single-digit element ID):
```
42["event",{"id":5,"client_id":"d62704b4-1e0e-4a4f-8d57-4d8ab0834d76","listener_id":"de329627-7838-4dac-a02a-63affac18475","args":["\"\""]}]
```
I guess this explains the breaking point of 1_000_000 - 240 characters: This is 140 short of 999_900.

evnchn:
`1_000_000 - 100`

The `-100` was to leave some buffer room. It _should_ work without the `-100` (I think)

PR #4572, opened 2025-04-05
Add @ui.page('/') to fix output_selected_row

### LINKED ISSUE #4570: Your documentation has one wrong examples for ui.aggrid (I think, could be wrong), using await with auto-indexed pages
### First Check

### Example Code

```python
code from https://nicegui.io/documentation/aggrid#select_ag_grid_rows
```

### Description

This is the error when trying to run exact code from the above example url (https://nicegui.io/documentation/aggrid#select_ag_grid_rows):
RuntimeError: Cannot await JavaScript responses on the auto-index page. There could be multiple clients connected and it is not clear which one to wait for.
You just have to add pagination or whatever just like in a similar example:
https://nicegui.io/documentation/aggrid#filter_return_values

### NiceGUI Version

2.13.0

### Python Version

Python 3.13.2

### Browser

Other

### Operating System

Windows

### Additional Context

_No response_

evnchn:
fixes #4570 

TL-DR: Wrap everything in `@ui.page`, so that we do not run_javascript on auto-index page, leading to "Cannot await JavaScript responses on the auto-index page"

---

We need to think systematically about how we can make this occur less often, potentially with a screening mechanism to show a warning when the documentation attempts to await JavaScript response anywhere (even in a method which internally calls it) without using the `# @ui.page('/')` and `page() # HIDE`

Moreover, perhaps the functionalities of doc.demo can be better documented itself. This also took me quite some time to figure out the syntax, which parses the source code of the demo code 🤯

### DIFF
--- a/website/documentation/content/aggrid_documentation.py
+++ b/website/documentation/content/aggrid_documentation.py
@@ -42,36 +42,39 @@ def main_demo() -> None:
     See the [AG Grid documentation](https://www.ag-grid.com/javascript-data-grid/row-selection/#example-single-row-selection) for more information.
 ''')
 def aggrid_with_selectable_rows():
-    grid = ui.aggrid({
-        'columnDefs': [
-            {'headerName': 'Name', 'field': 'name', 'checkboxSelection': True},
-            {'headerName': 'Age', 'field': 'age'},
-        ],
-        'rowData': [
-            {'name': 'Alice', 'age': 18},
-            {'name': 'Bob', 'age': 21},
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Thanks for the quick fix, @evnchn!

Yes, the inner workings of our documentation framework are pretty involved. They could certainly be improved here and there, but the code is pretty stable and the ordinary contributor doesn't have to deal with it. So I'd rather focus on other parts of the project.

PR #4574, opened 2025-04-05
Improve chunking large packages for NiceGUI On Air

rodja:
This PR improves the chunking for large files. Specifically:

- better chunk sizes for bulk transfer and streaming
- upload `data['src']` in chunks (for `ui.download` with bytes)
- ...

### DIFF
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -78,10 +78,9 @@ class Air:
                 'content': compressed,
             }
 
-            # NOTE: chunk large responses to stay within the SocketIO limit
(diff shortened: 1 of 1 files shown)


falkoschindler:
We came to the conclusion that it is a bad idea to transmit large messages over Socket.IO. This PR only "solves" the problem for downloads, but it would be better to use URL or file downloads instead of passing large content.

evnchn:
The code could perhaps be useful for chunking in the client side for big messages, but yeah, if we want to send files across, perhaps another way is better.

PR #4576, opened 2025-04-05
Fix docs section_styling_appearance label bug

evnchn:
Originally reported to me by @hwtam privately. 

TL-DR: minor bugfix for https://nicegui.io/documentation/section_styling_appearance#try_styling_nicegui_elements_

Existing issue: 

1. Button string content becomes "Button" when it should be `ui.button('element')`, when anything is typed in the props input field. 

image

2. Mangled label for especially `ui.checkbox` and `ui.switch`

{8D8C0ED1-66DF-4849-96DB-82FA460D8C87}

Proposed fix: 

* Change `element.props['label'] = 'Button'` to `element.props['label'] = 'element'`
* Never do `element.props['label'] = 'element'` if selected `ui.checkbox` or `ui.switch`

### DIFF
--- a/website/documentation/content/section_styling_appearance.py
+++ b/website/documentation/content/section_styling_appearance.py
@@ -75,9 +75,10 @@ def styling_demo():
                     with ui.row().classes('items-center gap-0 w-full'):
                         def handle_props(e: events.ValueChangeEventArguments):
                             element.props.clear()
-                            if select_element.options[select_element.value] not in ["ui.switch", "ui.checkbox"]:
+                            if isinstance(element, (ui.button, ui.input, ui.textarea)):
                                 element.props['label'] = 'element'
-                            element.props['color'] = 'primary'
+                            if isinstance(element, ui.button):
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Thanks, @evnchn!

I changed the code to the following. This should define each prop reset more clearly:
```py
if isinstance(element, (ui.button, ui.input, ui.textarea)):
    element.props['label'] = 'element'
if isinstance(element, ui.button):
    element.props['color'] = 'primary'
```

1. The first problem was a clear bug: The button label needs to be reset to "element", not "button".
2. The other glitch is a bit complicated:
    The reset was wrong. Only `ui.button`, `ui.input` and `ui.textarea` set the "label" prop. Thus only these three elements need to reset it to "element".
    If you enter "label=foo" for, e.g., `ui.checkbox`, you will get "fooelement" (after my recent commit). But this is correct because `ui.checkbox` uses the default slot. If you use the "label" prop on top, you get two labels.

Issue #3337, opened 2024-07-10
`ui.codemirror` not updating when value changes programmatically


zkx06111:
### Description

Using `nicegui==1.4.28`

```python
from nicegui import ui

shared_memory = {
    'code': '// Write your code here',
    'design': '',
    'test': '',
}


@ui.page('/')
def main():
    editor = ui.codemirror(language='python').bind_value(shared_memory, 'code')
    ui.markdown(shared_memory["code"]).bind_content(shared_memory, 'code')
    ui.button('clear code').on('click', lambda: shared_memory.update(code=''))
    
ui.run()
```

When the button is clicked the first time, it works as expected. Both the markdown text and the codemirror editor are cleared.

However, when it's clicked the second time, the markdown gets cleared while the codemirror remains the same.

When I try to edit what's in codemirror, I get the following error.

```
Traceback (most recent call last):
  File "/Users/kexunz/miniconda3/envs/ad/lib/python3.11/site-packages/nicegui/events.py", line 413, in handle_event
    result = handler(arguments) if expects_arguments else handler()
             ^^^^^^^^^^^^^^^^^^
  File "/Users/kexunz/miniconda3/envs/ad/lib/python3.11/site-packages/nicegui/elements/mixins/value_element.py", line 40, in handle_change
    self.set_value(self._event_args_to_value(e))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kexunz/miniconda3/envs/ad/lib/python3.11/site-packages/nicegui/elements/codemirror.py", line 294, in _event_args_to_value
    new_value = changeset.apply(self.value)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kexunz/miniconda3/envs/ad/lib/python3.11/site-packages/nicegui/elements/codemirror.py", line 322, in apply
    raise ValueError('Cannot apply change set to a document with the wrong length')
ValueError: Cannot apply change set to a document with the wrong length
```

python-and-fiction:
It is a little similar to #3217 .
I change `LOOPBACK = None` in `.venv\Lib\site-packages\nicegui\elements\codemirror.py` to `LOOPBACK = True`.
It works as expected.

Following code shows that `_props['value']` has changed,
but the value displayed in `ui.codemirror` is the same as the value before.
```python3
from nicegui import ui

shared_memory = {
    'code': '',
    'design': '',
    'test': '',
}

@ui.page('/')
def main():
[...]

PR #4578, opened 2025-04-05
fix ui.codemirror incompatibility with JS UTF-16-based indexing

### LINKED ISSUE #4575: `ui.codemirror` does not apply changes with special characters like 🔄
### First Check

### Example Code

```python
go and see: https://github.com/gbrault/ngcmbug
```

### Description

- start the app in vs-code with justmycode=false
- open a browser http://127.0.0.1:8081
- add a char on the first line

triggers

```
Cannot apply change set to a document with the wrong length
  File "d:\SeaDev\ngcmbug\env\Lib\site-packages\nicegui\events.py", line 436, in handle_event
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "d:\SeaDev\ngcmbug\env\Lib\site-packages\nicegui\elements\mixins\value_element.py", line 40, in handle_change
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "d:\SeaDev\ngcmbug\env\Lib\site-packages\nicegui\elements\codemirror.py", line 335, in _event_args_to_value
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "d:\SeaDev\ngcmbug\env\Lib\site-packages\nicegui\elements\codemirror.py", line 363, in apply
```

### NiceGUI Version

2.13.0

### Python Version

3.12

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

evnchn:
fixes #4575 

TL-DR: Who uses UTF-16! Dear JS: Don't you notice the `<meta charset="UTF-8">` in the HTML head? Codemirror has no choice but to use UTF-16 as well! JavaScript is terrible! Everyone has to work around you!

Minimal test script: 

```py
from nicegui import ui
from pathlib import Path

my_label = ui.label()

editor = ui.codemirror(language='html').classes('w-full').style('height: 80vh;')

editor.bind_value_to(my_label, 'text')

ui.run(show=False, reload=False, port=8081, title='HTML Editor')
```

Result after PR: 

{C917CFEB-A086-456F-9B92-5831DA656C5B}

Result before PR: 

{97C54352-6647-4246-B154-AF4F203299D9}

---

Potential concerns: 
* Speed issue, if we repeatedly do `text.encode('utf-16le')` and then subsequently `.decode('utf-16le', errors='ignore')` for every slice operation? Is it better to store the doc as UTF-16 bytes internally instead?

### DIFF
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -245,6 +245,23 @@ SUPPORTED_THEMES = Literal[
 ]
 
 
+def get_code_units_length(text: str) -> int:
+    """Calculate the length of a string in code units (UTF-16)."""
+    return len(text.encode('utf-16le')) // 2
+
+
+def get_first_n_characters_utf16(text: str, n: int) -> str:  # basically doc[:n] for UTF-16
+    """Get the first n characters of a string, considering UTF-16 encoding."""
+    code_units = text.encode('utf-16le')
+    return code_units[:2*n].decode('utf-16le', errors='ignore')
+
+
+def rest_of_string_from_n_utf16(text: str, n: int) -> str:  # basically doc[n:] for UTF-16
+    """Get the rest of the string from the nth character, considering UTF-16 encoding."""
+    code_units = text.encode('utf-16le')
+    return code_units[2*n:].decode('utf-16le', errors='ignore')
+
+
 class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', default_classes='nicegui-codemirror'):
     VALUE_PROP = 'value'
     LOOPBACK = None
@@ -359,7 +376,7 @@ class _ChangeSet:
 
     def apply(self, doc: str) -> str:
         """Apply the changes to a document, returning the modified document."""
-        if self.length() != len(doc):
+        if self.length() != get_code_units_length(doc):
             raise ValueError('Cannot apply change set to a document with the wrong length')
         return _iter_changes(self, doc, _replacement_func, individual=False)
 
@@ -403,7 +420,7 @@ def _iter_changes(
 
 
 def _replace_range(doc: str, from_: int, to: int, new: str) -> str:
-    return doc[:from_] + new + doc[to:]
+    return get_first_n_characters_utf16(doc, from_) + new + rest_of_string_from_n_utf16(doc, to)
 
 
 def _replacement_func(doc: str, from_a: int, to_a: int, from_b: int, _to_b: int, text: str) -> str:


gbrault:
@evnchn  Thanks! have a look to https://github.com/gbrault/ngcmbug. I have upgraded to your bug fixing branch and it works.

Will this be integrated in the main stream of nicegui?

Thanks again for your time!

evnchn:
> Will this be integrated in the main stream of nicegui?

Depends on review process of this PR. All I can say is: Soon™

falkoschindler:
@evnchn Can you, please, describe what this PR is doing? How does it solve the bug?

evnchn:
@falkoschindler My apologies. **An entire section went missing in the PR description** 😅 I really should not PR in the middle of the night. Here it is: 

Detailed explanation: 

As we all know, JavaScript handles the strings internally as UTF-16 codepoints, and the .length method of a string returns the number of codepoints used to represent a character, NOT the number of characters. 

In this instance, the 🔄 character takes 2 codepoints. This causes a mismatch in the way JS and Python calculates the length, the `ChangeSet` was never applied, and client and server falls out of sync. 

In my PR, I have _re-implemented_ length calculation, getting the first N characters, and getting the characters starting from the N character _in a UTF-16 codepoint manner_, such that it is essentially a "JavaScript compatible" method. 

In such a manner, Python and JavaScript shall agree, ChangeSet shall be applied, and the client and server stays in sync. 

If in doubt, kindly run the test script to see the difference. 

Before the PR, should the 🔄 be in the codemirror textbox, further typing anything or deleting anything will cause the client and server to lose sync and ValueError raised in console. 

After the PR, everything seamlessly works.

nicegui/elements/codemirror.py
```diff
@@ -359,7 +376,7 @@ def length(self) -> int:
 
     def apply(self, doc: str) -> str:
         """Apply the changes to a document, returning the modified document."""
-        if self.length() != len(doc):
+        if self.length() != get_code_units_length(doc):
```
evnchn, line 379: `get_code_units_length` is `len`, but counting in UTF-16 manner

evnchn:
Should fix #3337

evnchn:
@falkoschindler I would recommend looking at putting this into 2.16 if thought fit, so that we don't shake up codemirror twice.

evnchn:
@falkoschindler off topic: I see you love to use Cursor. How is it, generally? I may get on board. 

First off: When I made this PR, I want to make it work. Elegance had low-priority then. But now we can work on it, given a working copy. 

**TL-DR: Recommend server-side solution with "maintain internal UTF-16 representation" over "convert again and again" (what we have now) for relative coding ease and client-side performance. Those who would like to sidestep server-side performance impact can query bytestring from `ui.codemirror` and not use its default `.value` and access `.raw_value` directly (getting a UTF-16 byte string)**

---

Screening through cursor's points:

1. Seems good. We can then probably override the getter with `@property` or something. 
2. Not recommend since:
  a. codemirror is a black box and I don't quite want to mess with it
  b. juggling UTF-8 vs UTF-16 in JS seems much harder than in Python
  c. forces a conversion, which some may actually not need (if my algorithm is fine reading bytestrings)
  d. actually, reducing client load is most important in ensuring snappy response, lower power usage for mobile clients, and compatibility with low-spec devices (think: my Atom tablet). 
3. Cursor's hallucinating. 
4. Same as point 1 mostly.

evnchn:
New Proposal: 

https://www.reddit.com/r/pythontips/comments/10ukodx/tip_use_python_builtin_binary_search_bisect_on/?rdt=47974

Maintain a list `cumulative_num_code_points`, same length as the Python string, which stores the number of UTF-16 code points per character in the Python string from start. 

Do a binary search to find the starting index to slice the Python string into. Do another binary search to find the ending index to slice the Python string into as well. 

Seems like it'd be pretty efficient, since `cumulative_num_code_points` is sorted, so we get to enjoy `O(log n)` time complexity.

falkoschindler:
> New Proposal: [...] Do a binary search

Yeah, that might work.

Before adding additional state to `ui.codemirror`, I'm wondering if we can drastically simplify the current implementation. The `_ChangeSet` has been translated from TypeScript and might have been part of a more complex framework. But in `_event_args_to_value` we only need to loop over `sections` and `inserted` - basically what `_iter_changes` is doing. And `_iter_changes` also looks a bit convoluted and non-pythonic. The parameter `individual` which is always `False` might just be the tip of the iceberg.

Anyway, before refactoring this code, I started to write some tests to better understand how these change sets work:
```py
def test_change_set():
    assert _ChangeSet(sections=[0, 1], inserted=[['A']]).apply('') == 'A'
    assert _ChangeSet(sections=[0, 2], inserted=[['AB']]).apply('') == 'AB'
    assert _ChangeSet(sections=[1, 2], inserted=[['AB']]).apply('X') == 'AB'
    assert _ChangeSet(sections=[1, -1], inserted=[]).apply('X') == 'X'
    assert _ChangeSet(sections=[1, -1, 0, 1], inserted=[[], ['Y']]).apply('X') == 'XY'
    assert _ChangeSet(sections=[5, -1, 0, 8], inserted=[[], [', world!']]).apply('Hello') == 'Hello, world!'
    assert _ChangeSet(sections=[1, -1, 0, 1], inserted=[[], ['!']]).apply('🔄') == '🔄!'
```

But why is the last assertion passing? Isn't it exactly the situation in #4575?

---

Off-topic: Yes, I switched from VSCode with Copilot to Cursor and I don't want to go back. The AI integration is on whole new level, allowing for much more complex code suggestions and more seamless collaboration. Like giving your colleague access to your IDE instead of chatting with him on the phone.

evnchn:
> I'm wondering if we can drastically simplify the current implementation. 

Sure, but I need you to understand clearly the problem at hand since you seem to have some misunderstanding towards the issue. 

> a bit convoluted and non-pythonic

It is downright the most non-pythonic code in the entire NiceGUI codebase. 🤡

> Why is the last assertion passing

Note these points: 
- There is no issue on the last line, since the `value` doesn't yet contain any UTF-16 multi-code-point characters, and Python and JS will agree on the indices. 
- Apply any other `_ChangeSet` though (with JS indices*), it will blow up, since JS thinks the string has a `.length` of, say 5, but Python thinks it `len()` to 4. 
- *I really hope the indices you use in the test case are intercepted from the browser (insert print statement / inspecting WS messages) for the faithful recreation of the issue. 
- *If you use Python's UTF-8 indices for `sections`, it'll be fine regardless of `value` having UTF-16 multi-code-point characters or not. 

---

Off-topic: I'll see if I want to make the switch to Cursor, then.

evnchn:
Ah. I think I got a way to explain this in a manner you got to understand personally.  

Suppose I'm German-intolerant. So I can't really wrap my head around "Zauberzeug" as a single word. 

Internally, I must jot it down as "Magic things", which is 2 words to me, 1 word to everybody else. 

If we need to keep a notepad synchronized over a phone, announcing the changes by the word index, then everything would be fine up and until we need to write "Zauberzeug" in the notepad. 

Then, I'm going to tell you "insert `GmbH` after the 2rd word" and: 
- you're going to say "but `Zauberzeug` has 1 word, where's the 2nd word?"
- while I'm thinking "isn't the 2nd word `things` if the notepad has `Magic things` written on it?"

Thus, we never get to have "Zauberzeug GmbH" written on the notepad. 

Replace: 
- German with UTF-8
- Me with JS client
- You with Python server
- notepad with `.value` of Codemirror
- write words on notepad with `_ChangeSet(...).apply(...)`

And it'll make total sense. 

---

On(?)-topic: Your language is truly in the spirit of UTF-8. I can't imagine `Kraftfahrzeug-Haftpflichtversicherung` being a **single** word for "motor vehicle liability insurance". It's 4 words over here in English.

falkoschindler:
Thanks for the explanation. I get it. A pytest which completely lives in Python with UTF-8 doesn't experience the problem like when communicating with JavaScript.

Anyway, I just created PR #4676 in preparation for the proposed bisection. Reducing the code to a few lines only should help to keep the complexity manageable.

Would you like to tackle the bisection, @evnchn?

evnchn:
If we merge #4676, we won't work on this PR anymore most likely, and we work with a fresh start from there. Otherwise there will be so many merge conflict. 

So I would wait for that PR to settle before selecting between bisection approach, or maintain internal UTF-16 representation approach (with possibility to turn off decoding for speed improvement) or a fusion of both approaches.

evnchn:
There's no way this can be merged. Look at the merge conflict. 

Also, we have #4678 now. 

I'm closing this PR. Not deleting the branch is my convenience for referencing the old code.

PR #4579, opened 2025-04-05
advanced_serialize for Touch which JSON.stringify fails

### LINKED ISSUE #4577: Can not get proper touch event data in backend
### First Check

### Example Code

```python
from nicegui import ui 

label = ui.label('touch data')

with ui.element('div').style('width:80vw;height:80vh; background-color:lightblue;') as toucharea:
    toucharea.on('touchmove', lambda e: print(e) and label.set_text(f"{e.args['touches']['0']}"))

ui.run(host='0.0.0.0', port=8000)
```

### Description

start the app. open the app in a browser of  a touch screen device that is connected to the same network. move your finger across the touch area, no touch position is shown.

### NiceGUI Version

2.13.0

### Python Version

3.12

### Browser

Chrome

### Operating System

Android

### Additional Context

_No response_

evnchn:
fixes #4577 

Because the `Touch` cannot be successfully JSON-serialized via `JSON.stringify`

ChatGPT's solution which DOESN'T WORK: https://poe.com/s/9qlncNJzjKOpKfILA2Ts

Notably, obj.hasOwnProperty(key) returns false, but it is indeed the value we want, I also don't need the recursive call either, so I straight-up ignored it and always do `result[key] = obj[key]`, and do the `advanced_serialize` only when we encounter a `Touch` event. 

Burning question: 
* Why `Touch` is so bad, cannot JSON-serialize
* Are there any other problematic objects? 
* If you do `emitEvent`, it does get sent, so why the Socket.IO can serialize but JSON-serialize can't?
  * ~~Ties in with #4571 since currently use JSON.stringify to estimate the length of the message before sending via Socket.IO!~~ Check my update below.


nicegui/static/nicegui.js
```diff
@@ -8,6 +8,15 @@ let mounted_app = undefined;
...
 
+function advanced_serialize(obj) {
+  // Works for object for which JSON.stringify fails
+  const result = {};
+  for (const key in obj) {
+    result[key] = obj[key];
+  }
+  return result;
+}
+
```
evnchn, line 19: `advanced_serialize` is specifically for `Touch` (and potentially other tricky objects for which JSON.stringify returns `'{}'`)

evnchn:
Appears that the serialization of properties for which `hasOwnProperty` returns false is important. 

{8DFBE8F8-14D9-4424-BFA3-B178EB8F5549}

So that's why the serialization totally misses property like `altKey`, and relevant to the discussion in #4589 `button`

evnchn:
Hey. 

https://stackoverflow.com/a/39926468

```js
function flatten(obj) {
    var ret = {};
    for (var i in obj) {
        ret[i] = obj[i];
    }
    return ret;
}
```

Turns out what I am doing for `advanced_serialize` isn't advanced at all. Rather, it is standard affair.

evnchn:
I am still struggling to make a serializer which serializes everything under the sun, without running into `TypeError: Illegal invocation` or `RangeError: Maximum call stack size exceeded`

I'm simply trying out everything in the StackOverflow.

evnchn:
Marking this as draft, because: 

1. Still no idea how to serialize every key, including inherited ones
2. Even if we succeed, every single message sent will become larger, affecting responsiveness of client and the server
3. #4618 much more compelling

evnchn:
I'l closing this in hopes that we can focus more attention to #4689, which is **sitting dry for a week despite being 1 of 2 PR marked as major, and can resolve 5 issues as mentioned https://github.com/zauberzeug/nicegui/pull/4618#issuecomment-2845918657**

PR #4584, opened 2025-04-06
EChart resizeObserver skipping initial rendering

### LINKED ISSUE #4535: apache echart always renders twice, why?

depley:
Should fix #4535 overcoming the before discussed resizing approaches causing issues, illustrated [here](https://[depley.github.io/nicegui-custom-echarts/](https://depley.github.io/nicegui-custom-echarts/)) 

Suggested solution is based on [ecomfe/vue-echarts .. autoresize.ts](https://github.com/ecomfe/vue-echarts/blob/b7852ab6430084c54a8e1c0a05648748da94bc64/src/composables/autoresize.ts#L7).
Basically a direct attached resizeObserver but skipping an initial resize if the initial size of the container element hasn't changed.

The in [#4535 comment](https://github.com/zauberzeug/nicegui/issues/4535#issuecomment-2781492261) discussed MREs can be reproduced in nicegui via

```py
from nicegui import ui

# scatter series chart
# expected behaviour: should not render twice initially, render once after resizing
echart_many_data = ui.echart({
    'xAxis': {'type': 'value'},
    'yAxis': {'type': 'value'},
    'series': [{
        'type': 'scatter',
        'data': [[x, x] for x in range(20_000)]
    }],
})

# line series chart
# expected behaviour: animation drawing a line and points on initial rendering 
# (broken animation behaviour was that the line is already there)  
echart_animation = ui.echart({
    'xAxis': {
        'type': 'category',
        'data': ["Mon", "Tue", "Wed", "Thu", "Fri"],
    },
    'yAxis': {'type': 'value'},
    'series': [{
        'type': 'line',
        'data': [150, 230, 224, 147, 260],
    }],
})

ui.run()
```

### DIFF
--- a/nicegui/elements/echart.js
+++ b/nicegui/elements/echart.js
@@ -54,13 +54,10 @@ export default {
       this.chart.on(event, (e) => this.$emit(`chart:${event}`, e));
     }
 
-    // Prevent interruption of chart animations due to resize operations.
-    // It is recommended to register the callbacks for such an event before setOption.
-    const createResizeObserver = () => {
-      new ResizeObserver(this.chart.resize).observe(this.$el);
-      this.chart.off("finished", createResizeObserver);
-    };
-    this.chart.on("finished", createResizeObserver);
+    this.initialResizeTriggered = false;
+    const initialHeight = this.$el.offsetHeight;
+    const initialWidth = this.$el.offsetWidth;
+    const resizeObserver = new ResizeObserver(this.resizeChart).observe(this.$el);
 
     this.update_chart();
   },
@@ -71,6 +68,16 @@ export default {
     this.chart.dispose();
   },
   methods: {
+    resizeChart() {
+      if (!this.initialResizeTriggered) {
+        this.initialResizeTriggered = true;
+        if (this.$el.offsetWidth === initialWidth && this.$el.offsetHeight === initialHeight) {
+          return;
+        }
+      } else {
+        this.chart.resize();
+      }
+    },
     update_chart() {
       if (!this.chart) {
         setTimeout(this.update_chart, 10);


evnchn:
Appears to work. Tried with the test script: 

```py
from nicegui import ui

def get_charts(chart_type):
    for animation in [False, True]:
        for count in [10000, 5]:
            mychart = ui.echart({
                'title': {'text': f'{chart_type}: {count} points {"(animated)" if animation else ""}'},
                'xAxis': {'type': 'value'},
                'yAxis': {'type': 'value'},
                'series': [{'type': chart_type, 'data': [[x, x] for x in range(count)]}],
                **({'animationDuration': 5000} if animation else {}),
            })
            print(mychart.props.get("direct_attach_resizeobserver", "NOTEXIST"))

@ui.page('/')
def main_page():
    ui.link("Scatter", '/scatter')
    ui.link("Line", '/line')
    ui.link("Bar", '/bar')

@ui.page('/scatter')
def scatter_page():
    get_charts('scatter')
    ui.link('Back', '/')

@ui.page('/line')
def line_page():
    get_charts('line')
    ui.link('Back', '/')

@ui.page('/bar')
def bar_page():
    get_charts('bar')
    ui.link('Back', '/')

ui.run(port=4949)
```

Which is basically https://github.com/zauberzeug/nicegui/pull/4536#pullrequestreview-2740631836 but tests 3 chart types just to be safe.

evnchn:
So, If I understand correctly: 

When `resizeChart` is called **for the very first time**, there is an opportunity for skipping the actual `this.chart.resize()`, should the size have not been changed since the chart is rendered. 

Thereafter, there is no such opportunity for bypass, and `this.chart.resize()` is **always** called. 

---

_The below content fits another PR, probably not this one:_ 

Also, further comparing the difference between implementation, I notice the use of `throttle` in the Vue implementation. 

Can we also implement or borrow a throttle function from somewhere, so that it can make resize less laggy? 

Note: `import { throttle } from "echarts/core";` is in their implementation. 

This seems to be for another purpose...

https://github.com/zauberzeug/nicegui/blob/b9c9e8492750153d4b519b24da57b02751ff225f/nicegui/static/nicegui.js#L110-L139

depley:
> So, If I understand correctly:
> 
> When `resizeChart` is called **for the very first time**, there is an opportunity for skipping the actual `this.chart.resize()`, should the size have not been changed since the chart is initially created.
> 
> Thereafter, there is no such opportunity for bypass, and `this.chart.resize()` is **always** called.

Exactly. 
Thought also about a property like vue-echart is using to enable/disable the autoresize mechanism.
Let's say 'disableAutoResize' defaulting to False, activatable via ui.element.default_props('disableAutoResize').
But than it's again the discussion introducing a new property.

---
> 
> _The below content fits another PR, probably not this one:_
> 
> Also, further comparing the difference between implementation, I notice the use of `throttle` in the Vue implementation.
> 
> Can we also implement or borrow a throttle function from somewhere, so that it can make resize less laggy?
> 
> Note: `import { throttle } from "echarts/core";` is in their implementation.
> 
> This seems to be for another purpose...
> 
> https://github.com/zauberzeug/nicegui/blob/b9c9e8492750153d4b519b24da57b02751ff225f/nicegui/static/nicegui.js#L110-L139

Also noticed the `throttle`. You mean laggy resize for ui.echart or something more general?
When for echart, one could see the difference with an MRE 
pure echarts.js via CDN (without throttle) vs   (with throttle) vue-echarts via [CDN](https://github.com/ecomfe/vue-echarts/tree/b7852ab6430084c54a8e1c0a05648748da94bc64?tab=readme-ov-file#cdn)

depley:
Ah, you're totally right, @falkoschindler.

The first callback sets `initialResizeTriggered = true;` and breaks with error.
This codeline never completed.
```js
if (this.$el.offsetWidth === initialWidth && this.$el.offsetHeight === initialHeight)
```
This way skipping the initial resize was not indended.
With your fix it works without errors!

Found also my fault during debugging. Cache not disabled in browser DevTools.
Why vscode not marked 'unused' properly is under investigation.

falkoschindler (approved):
Awesome! Let's merge this one! 🚀

PR #4596, opened 2025-04-09
Add theme support to ECharts

thetableman:
This PR introduces the ability to specify a custom theme with an ECharts element.

Users can create a JSON theme file using the ECharts Theme Builder here: https://echarts.apache.org/en/theme-builder.html
The JSON can then be added to a static files path and the URL passed to the element or a dictionary (matching the format of the JSON theme) can be passed directly.

Themes are applied to an instance so multiple charts can be displayed on the same page with different themes.

Example: 
The first chart has a theme passed as a dict, the second is a URL to a JSON in static files, the third has no theme specified.
image

### DIFF
--- a/nicegui/elements/echart.js
+++ b/nicegui/elements/echart.js
@@ -9,7 +9,7 @@ export default {
       await import("echarts-gl");
     }
 
-    const theme_name = this.theme ? crypto.randomUUID() : null;
+    const theme_name = this.theme ? createRandomUUID() : null;
     try {
       if (typeof this.theme == "string") {
         const response = await fetch(this.theme);
--- a/nicegui/elements/echart.js
+++ b/nicegui/elements/echart.js
@@ -1,6 +1,31 @@
 import "echarts";
 import { convertDynamicProperties } from "../../static/utils/dynamic_properties.js";
 
+async function fetchAndParseJson(url) {
+  try {
+    const response = await fetch(url);
+
+    if (!response.ok) {
+      console.error(`HTTP error! Status: ${response.status}`);
+      return;
+    }
+
+    const text = await response.text();
+
+    try {
+      const json = JSON.parse(text);
+      return json;
+    } catch (e) {
+      console.warn("Response is not valid JSON.");
+      return;
+    }
+
+  } catch (error) {
+    console.error("Fetch error:", error);
+    return;
+  }
+}
+
 export default {
   template: "<div></div>",
   async mounted() {
@@ -9,30 +34,31 @@ export default {
       await import("echarts-gl");
     }
 
+    let theme_name = null
+    let theme_json = null
+
     if (this.theme) {
-      let custom_theme_json
-     if (typeof this.theme == 'object') {
-      custom_theme_json = this.theme
-     } else {
-      var custom_theme = await fetch(this.theme)
-      custom_theme_json = await custom_theme.json()
+
+      if (typeof this.theme == 'object') {
+        theme_json = this.theme;
+        theme_name = self.crypto.randomUUID();
+
+      } else {
+        theme_json = await fetchAndParseJson(this.theme);
+
+        if (theme_json) {
+          theme_name = self.crypto.randomUUID();
+        } else {
+          console.log("No valid JSON theme returned.");
+        }
      }
-     var custom_theme_name = 'custom_theme'
-     echarts.registerTheme(custom_theme_name, custom_theme_json)
-    } else {
-      var custom_theme_name = null
     }
 
-    // if (this.theme) {
-    //   var custom_theme_name = 'custom_theme'
-    //   var custom_theme = await fetch(this.theme)
-    //   var custom_theme_json = await custom_theme.json()
-    //   echarts.registerTheme(custom_theme_name, custom_theme_json)
-    // } else {
-    //   var custom_theme_name = null
-    // }
+    if (theme_name && theme_json) {
+     echarts.registerTheme(theme_name, theme_json);
+    }
 
-    this.chart = echarts.init(this.$el, custom_theme_name, { renderer: this.renderer });
+    this.chart = echarts.init(this.$el, theme_name, { renderer: this.renderer });
     this.chart.on("click", (e) => this.$emit("pointClick", e));
     for (const event of [
       "click",
--- a/nicegui/elements/echart.js
+++ b/nicegui/elements/echart.js
@@ -1,31 +1,6 @@
 import "echarts";
 import { convertDynamicProperties } from "../../static/utils/dynamic_properties.js";
 
-async function fetchAndParseJson(url) {
-  try {
-    const response = await fetch(url);
-
-    if (!response.ok) {
-      console.error(`HTTP error! Status: ${response.status}`);
-      return;
-    }
-
-    const text = await response.text();
-
-    try {
-      const json = JSON.parse(text);
-      return json;
-    } catch (e) {
-      console.warn("Response is not valid JSON.");
-      return;
-    }
-
-  } catch (error) {
-    console.error("Fetch error:", error);
-    return;
-  }
-}
-
 export default {
   template: "<div></div>",
   async mounted() {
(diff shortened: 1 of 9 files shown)


evnchn:
Assuming it works. 

Why is `theme: Optional[Union[str, Dict]] = None`, when `options: Dict`

Seems a bit inconsistent to me, and complicates the client logic a bit.

thetableman:
> Why is `theme: Optional[Union[str, Dict]] = None`, when `options: Dict`

Because they're different things. `options` is a required parameter and must be a dictionary, the proposed `theme` is not required (hence `Optional`) and can accept a url (`str`) or a dictionary (hence `Dict`).

evnchn:
Particularly referring to how it can be a string. Of course it's `Optional` 

```
      var custom_theme = await fetch(this.theme) 
       custom_theme_json = await custom_theme.json()
```

Rare sight to be seen on the client side.

thetableman:
> Particularly referring to how it can be a string. Of course it's `Optional`

It can be a str if it's a url.

> ```
>       var custom_theme = await fetch(this.theme) 
>        custom_theme_json = await custom_theme.json()
> ```
> 
> Rare sight to be seen on the client side.

This is how the official documentation suggests doing it, if there's a better way then make suggestion. I'm not a JavaScript person so just followed the examples provided and found it worked as expected.

evnchn:
If I'm not mistaken, if you pass in a dict, then `custom_theme_json = this.theme` right away. If not, it's two `await` calls and one network request later. Depending on distance between client and server, and client CPU speed, could be slow. 

I'm also just afraid of some nasty layout issues if the client awaits for the JSON, such that the element is not rendered in the meanwhile. Again I'm on the phone so can't test. 

Also, this may arbitrary encourage arbitrary fetch to third-party servers. May cause some privacy concerns, and if the third-party server doesn't have CORS, the theme would not get loaded, and the code should error out since there isn't error handling in your implementation?

Finally, since we have the power of Python, why not open the file and then serve the dict to the client right away. 

---

Will say though, I'm supportive of the theme option. Just that, perhaps we can think about the way to pass it in more. 

Falko probably isn't too available since we got some issues which 2.14.1 is rushed to fix.

thetableman:
> If I'm not mistaken, if you pass in a dict, then custom_theme_json = this.theme right away.

You're right, that's why I check if it's a dict first and only proceed to `await fetch(...)` if it is not an object.

> I'm also just afraid of some nasty layout issues if the client awaits for the JSON, such that the element is not rendered in the meanwhile.

Potentially though haven't experienced any such issues in testing or in general use as of yet.

> Also, this may arbitrary encourage arbitrary fetch to third-party servers. May cause some privacy concerns, and if the third-party server doesn't have CORS, the theme would not get loaded...

Why would any developer deliberately reference a URL they knew wouldn't work? Surely there is always an expectation of due-diligence when publishing an app and there are plenty of other places you could exploit as a developer. Regardless, as mentioned, I'm not a JS person so please feel free to suggest alternatives here. Or perhaps there's a restriction to ensure the file is only served from a relative path?

Also worth noting that 'fetch(...)` is used in multiple places so clearly the function is an acceptable solution in client cases, though admittedly not from a parameter.

> ...and the code should error out since there isn't error handling in your implementation?

So a `try...catch` then?

> Finally, since we have the power of Python, why not open the file and then serve the dict to the client right away.

Yeah, valid suggestion. Feel free to provide an example of how your would implement this idea. I think it's preferred to open the JSON from a URL because then it's cached but otherwise it could be a path to a file.

There is also another JS method suggested on the ECharts website that uses a head script tag, but I couldn't get that method working as the theme needs to be loaded after the echarts library and before an echart instance is created. Also I liked the idea of the theme being in JSON rather than the JS because it can then be more easily used in other ways, say if the developer wanted to loaded it and pull some colors from an array to use elsewhere.

Regarding the `str` / `dict` question, it's worth noting that my documentation example shows a `dict` with only a few colors - this is for simplicity. An actual JSON theme from the ECharts Theme Builder is over 350 lines (mine is closer to 500), so the idea behind passing a URL to the file hosted on static files then becomes obvious, particularly if you reuse the theme in multiple charts as I do in my projects.

evnchn:
Seeing a config can be **as long as 350 lines** (am quite surprised actually to hear that), then I personally would prefer the client fetch the file. As you have said, this enables caching. 

Then, perhaps for uniformity, the client always fetches a file? Even if you pass a dict, the element implementation will put the dict into a file for the client to fetch? Seems like this can save us some headache down the line. Experience speaking it's always the JS side which is the most tricky to deal with. 

My concerns over client-side await can be non-concerns, if you have tested it out and it's fine. Also my concerns may not be super valid while I'm on a trip. I'm at Amsterdam, btw. 

---

But then, shifting gears, perhaps you can also handle serving the file to the client, if the path is relative, much like how `ui.download` does it, and may involve your modified implementation to run add_static_file automatically on behalf of the user. 

Also, for more advanced stuff now, can the library user change the theme programmatically? Can the library user bind theme to a variable? 

And the catching of error, refer to https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

---

Disclaimer: I'm also not a JS guy!

thetableman:
> Seeing a config can be as long as 350 lines (am quite surprised actually to hear that), then I personally would prefer the client fetch the file. As you have said, this enables caching.

It surprised me too. But you need to bear in mind that the theme file can essentially be used customize every single type of chart individually with colors, fonts, padding, text-alignment, etc. In fact the name 'Theme' seems to be a bit miss-used, it's more like a default config for all the chart settings. Also, it's 350+ lines 'human-readable', it would be considerably less if formatting/white-space was removed so it's less impressive to look at than it initially sounds.

> Then, perhaps for uniformity, the client always fetches a file? 

Yeah, this could be a solution though wouldn't necessarily help with caching. Part of the current design is that a different theme can be applied to different charts so, by virtue, each theme needs to be a unique file and theme name. If a theme was passed as a dictionary then first it would need to be converted to JSON, then written to a file, then added to static files, then fetched in the current process. This would mean determining a unique file name so as to avoid clashing with another JSON theme files, the file name would then be random (or inconsistent) and so wouldn't useful when cached as the name would be different next run. This would also add several steps that, in my suggested implementation, wouldn't occur for a dictionary. This is just my initial thought though and perhaps there's a better way to implement this suggestion. 

In fact, a unique name for each theme is something I need to implement in my own version as currently it's just called 'custom_theme' which may clash if multiple themes are loaded for different charts.

> Also, for more advanced stuff now, can the library user change the theme programmatically?

Maybe, but it's not something I have a need for and not something I come across in BI. Theme JSON files support all chart types so a theme wouldn't need to be changed if a chart was switched to another layout - such as the EChart toolbox which can switch a chart from bars to lines live. I've never been asked to switch the entire color scheme based on some external factor before. But yes, in theory, this may be possible.

> ...to run add_static_file automatically on behalf of the user

Are you suggesting the JSON file is loaded from disk each time? If so then the `Element` base class already has a `add_resource` method that can do this. However, this would be reading in and adding a static route every time a chart is loaded with that theme. Using a URL to a file already hosted on static files avoids this. Some of my pages are essentially BI dashboards with 10 or so visuals that all load - I don't think calling a file from disk and adding a static route 10 times makes sense. But, in principle, the ability to load a once-off from disk does make sense so could be added.

I actually wonder whether themes should be registered as a class method, a bit like `default_classes`. A dev could then run the class method at server startup to create the routes and check the validity of the file or URL and raise errors then. The `theme` param would still be `str`/`dict` but the `str` would be the name of the theme registered rather than a URL. This would circumvent the risks of just any URL being passed when the EChart element is declared though doesn't do away with `fetch` client-side.

falkoschindler (approved):
Thanks for the pull request, @thetableman!

I simplified the code here and there and added another test for the URL case. From my point of view it is ready to merge. Or is there anything from the discussion that should be accounted for before merging?

evnchn:
https://github.com/zauberzeug/nicegui/blob/main/nicegui%2Fstatic%2Fnicegui.js#L295

Use this for UUID generation instead of straight-up `const theme_name = this.theme ? crypto.randomUUID() : null;`?

thetableman:
Thanks for your contributions, recommendations, and improvements @evnchn and @falkoschindler, this one's been on my wish list for almost a year!

evnchn:
Congratulations! 🎉

Issue #4607, opened 2025-04-11
User fixture - Unexpected logs in pytest


pmeising:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui.testing import User

async def test_basic_functionality(user: User) -> None:
    await user.open("/")
```

### Problem: 
When using the User fixture after change 745167c81c219beda5ebaf850dfb7a3e772f96ac , any logging in the application during integration tests is causing the tests to fail. 

### Reproduce
The example test case is shown above using the User fixture as recommended in the documentation https://nicegui.io/documentation/user
It seems that any log (even just DEBUG, INFO or WARNING level logs) are causing the test teardown to trigger the pytest.fail call.

### TEST OUTPUT
`make tests TYPE=integration/test_test.py`
`"Running integration/test_test.py tests ..."`
`==================================== test session starts ============================`
`platform win32 -- Python 3.12.7, pytest-8.3.5, pluggy-1.5.0`
`rootdir: C:\Userdirectors`
`configfile: pyproject.toml`
`plugins: anyio-4.9.0, dash-2.18.2, asyncio-0.25.3, cov-5.0.0`
`asyncio: mode=Mode.AUTO, asyncio_default_fixture_loop_scope=function`
`collected 1 item`

`tests\integration\test_test.py .E`

`============================================================================================================= ERRORS ============================================================================================================== `
`__________________________________________________________________________________________ ERROR at teardown of test_basic_functionality __________________________________________________________________________________________ `
`There were unexpected logs.`
`------------------------------------------------------------------------------------------------------ Captured stderr setup ------------------------------------------------------------------------------------------------------ 
2025-04-11 08:26:14,338 - INFO - sensor_handler.SensorHandlerClass - Sensor handler initialized.`

[ . . . ]

`------------------------------------------------------------------------------------------------------- Captured log setup -------------------------------------------------------------------------------------------------------- 
INFO     sensor_handler.SensorHandlerClass:SensorHandlerClass.py:19 Sensor handler initialized.`

[ . . . ]

`===================================================================================================== short test summary info ===================================================================================================== `
`ERROR tests/integration/test_test.py::test_basic_functionality - Failed: There were unexpected logs.`

### NiceGUI Version

2.14.1

### Python Version

[...]

PR #4608, opened 2025-04-11
Fix unexpected logs in pytest when using the user fixture

pmeising:
Attempt to resolve the catch all logs error when using the user fixture.

Relates to #4607


rodja (approved):

falkoschindler:
Thanks a lot for this PR, @pmeising!
Filtering for error logs makes sense. 👍🏻

Issue #3644, opened 2024-08-29
Importing `nicegui` affects `mp.set_start_method`, throws `context has already been set`


EricCousineau-TRI:
### Description

Using `nicegui == 1.4.34`, CPython 3.10.12, on Ubuntu 22.04.

The following works fine:
```sh
$ python -c 'import multiprocessing as mp; mp.set_start_method("spawn")'
```

The following, however, fails:
```sh
$ python -c 'import nicegui; import multiprocessing as mp; mp.set_start_method("spawn")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.10/multiprocessing/context.py", line 247, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
```

rodja:
I currently have not the time to dig deeper but it seems to be related to the auto-reloading mechanism of uvicorn. I could reproduce the same behaviour without NiceGUI:

```py
import multiprocessing as mp
import uvicorn

mp.set_start_method("spawn")

if __name__ == "__main__":
    uvicorn.run("main:app", host="0.0.0.0", port=8000, reload=True)
```

If you set `ui.run(reload=False)` the start method can be set.

falkoschindler:
We don't think this is an issue with NiceGUI.
If you wish to be able to change the start method in a NiceGUI app (without `force=True`), feel free to create a feature request or even a pull request. Thanks!

EricCousineau-TRI:
Thanks for the feedback!

The underlying issue is *importing* either `nicegui` (or something it imports) creates a side effect where we can neither call `multiprocessing.set_start_method("spawn")`.

Note that the error above happens before any code is actually "run" -- I am not calling `app.run(...)` or `uvicorn.run(...)`.
Perhaps the issue is not with `nicegui`, but something in the `import` chain is causing a functional side effect at `import` time, which ideally should happen when code is explicitly called (e.g. via the `main()` function).

For us, we simply worked around it by deferring the import of any `nicegui` components to our `main()` function. This works, but is a little "dirty".

Does this make sense?

EricCousineau-TRI:
Also note that the above still fails for `nicegui == 2.0.1`.

It does not fail for `uvicorn == 0.30.6`:
```
$ python -c 'import multiprocessing as mp; import uvicorn; mp.set_start_method("spawn")'
# No error.
```

EricCousineau-TRI:
[...]

Issue #189, opened 2022-12-06
Why is spawn context not supported ?


tchaton:
Hey there,

I have seen ui.run doesn't support spawn.

```python
    if inspect.stack()[-2].filename.endswith('spawn.py'):
        return
```

However, after removing those lines, it seems to work just fine.

falkoschindler:
If NiceGUI runs with auto-reload enabled, the uvicorn server spawns a new process to serve the app. When an auto-reload is triggered, this process is re-started.

The if-statement in `ui.run` prevents the spawned process from starting another server - which would not work since the port is already in use. In order to identify the process we are currently in, we inspect the stack and look for "spawn.py". Maybe there is a better way to achieve the same...

But what are you actually trying to do? Can you give some context or a minimum example?

tchaton:
Hey @falkoschindler,

With the Lightning App Framework, we need to spawn the processes in order to move the model on GPU without CUDA Context Error.

Here is an example where I had to patch the frames in order to avoid this return checl: https://github.com/Lightning-AI/lightning-diffusion-component/blob/main/diffusion2/app_interactive.py#L61

```python
    def stack_patch():
        class FakeFrame:
            filename = "random"
        return [FakeFrame(), None]
    inspect.stack = stack_patch
    ui.run(host=host, port=port, reload=False)
```

As `reload=False`, it works just fine.

falkoschindler:
@tchaton Would it help to replace

```python
if inspect.stack()[-2].filename.endswith('spawn.py'):
    return
```

with the following lines?

```python
if current_process().name != 'MainProcess':
    return
```

This seems to be another way to tell [if a process is parent or not](https://stackoverflow.com/a/50435263/3419103).

falkoschindler:
@tchaton Have you had the chance to try out `if current_process().name != 'MainProcess':`? We could probably change this line pretty easily, but I'd like to know beforehand if it is any better than using `inspect.stack`.

falkoschindler:
We think checking `current_process().name` is cleaner than inspecting the caller's stack. Therefore we changed the implementation in [e8407c0](https://github.com/zauberzeug/nicegui/commit/e8407c08ce3e046aec95221a4b7c58b40f88e9f7).

PR #4613, opened 2025-04-12
Add require_main_process to run

davetapley:
This allowed NiceGUI to be started programmatically in a separate `SpawnContext.Process`, as part of a larger application.

See also:
- https://github.com/zauberzeug/nicegui/issues/189#issuecomment-1359559285
- https://github.com/zauberzeug/nicegui/issues/3644#issuecomment-2798998896


evnchn:
Hello @davetapley. Thanks for the PR. 

Although you most definitely are submitting it because it solved a problem of yours, I'm quite a bit hesitant towards this. 

Seems like we are making a check optional, and I am not sure of the side effects of what will happen from that. So, I'd like to talk with you for a bit first. 

1. https://github.com/zauberzeug/nicegui/issues/3644#issuecomment-2787226834

You have mentioned that it is no longer reproducible upstream using uvicorn 0.34.0, and we have seen above it work with 0.30.6. 

However, https://github.com/zauberzeug/nicegui/issues/3644#issuecomment-2319862379 @rodja commented that he can reproduce using a probably-older but unknown version of uvicorn. 

Would you mind going through older uvicorn releases to find which release (and better, which commit) that lead to the fix, so that we can better assess what was changed and whether we should follow suit? Hint: think about what's the latest uvicorn available at the time. 

2. https://github.com/zauberzeug/nicegui/issues/3644#issuecomment-2324168585

@falkoschindler reply implies that it would have worked with `force=True` anyways. It's concerning here that you also used `force=True` to illustrate the effectiveness of your PR.

Would you be so kind as to test, for completeness sake, with / without the patch, force=True present / absent (4 combinations), and put the results in a table?

davetapley:
@evnchn thanks for the feedback.

1. I can try and identify the uvicorn version, but I'm not not sure it matters with respect to this PR because: So long as someone is starting nicegui in a child thread then `multiprocessing.current_process().name ` is not going to be `'MainProcess'`, and so `run` will exit prematurely. I.e. the problem this PR solves would have existed even if the upstream uvicorn issue had never existed.

2. Yes `force=True` is required regardless. I'm not sure of the semantics, but I _assume_ the `RuntimeError: context has already been set` (which necessitates `force=True`) is thrown if `set_start_method` is *ever* called a second time. For what its worth in my application I've had `force=True` for several years of production use and it has been fine.

---

I'll let you know if I encounter any weirdness, but so far it seems to be behaving predictably including building the whole app with PyInstaller (for which I own [hooks-contrib](https://github.com/JEFuller/pyinstaller-hooks-contrib/tree/master) my hook file.

evnchn:
Interesting. Let me conclude for a bit. 

We have 2 separate issues, here: 

1. So long as someone is starting nicegui in a child thread, `ui.run` will exit prematurely. (addressed in this PR)
2. Require `force=True` for `mp.set_start_method`, else throws error. (addressed in #3644)

Your PR seem to address issue 1 and issue 1 only. It does not do anything with regards to issue 2. 

If that's the case, then finding which release that uvicorn throws error on `mp.set_start_method` without `force=True` is indeed not related, as we are using `force=True` anyways. 

@davetapley see if I got you right this time. If not, @ me right away with a reply. 

---

Then, some questions just to make sure: 

1. Can you test the following: Does reload=True/False still work? Does native=True/False still work? (4 combinations). These are in-my-opinion the most "impactful" parameters of `ui.run`, which shakes up dramatically how the NiceGUI app is launched. 
2. Can you, somehow, run the test cases in pytest, so as to see if anything breaks when we make this check optional? 
3. Does it now mean that, if `require_main_process=False`, and `ui.run` is not put inside a main guard, the code would blow up? Previously, `ui.run` can be inside a main guard or not (iirc) and it'd still be fine. 

Other than that, LGTM.

rodja:
Sorry, @davetapley but can you please sumarize your problem for me? Do you have example code for when you need to `require_main_process=False`? And how is that related to `spawn` vs `fork`?

davetapley:
@rodja sure! I'm programmatically start nicegui on a new process from inside my existing app (I'm using it to build an admin interface, the app is currently controlled just with config files and a CLI).

Basically like this:

```py
from multiprocessing.context import SpawnContext
from nicegui import ui

class GuiRunnerProcess(SpawnContext.Process):
    def run(self):
        ui.run(reload=False, require_main_process=False, show=False)


if __name__ == "__main__":
    set_start_method("spawn", force=True) # see below for why
    gui_process =  GuiRunnerProcess()
    gui_process.start()

    # rest of processes I start up in my application
```


I regret mentioning `force=true` / #3644 in the PR 😁 

It's only related insomuch as if you are doing what I'm doing you will additionally need `force=true` if you are using `set_start_method` (which I am because I want consistency across platforms, rather than [defaulting](https://docs.python.org/3.11/library/multiprocessing.html#contexts-and-start-methods) to `fork` on Linux).

evnchn:
Read #189 in more details. 

It seems like it is also a mix of 2 points: 
1. Whether a main process check was needed (conclusion: yes, for empowering `reload=True`)
2. How should such a main process check be done (conclusion: settle on `if multiprocessing.current_process().name != 'MainProcess':`)

Sorry but I focused on issue 2 when I first read #189. 

With point 1, I'd like to ask, would it make more sense and less hassle, if we bypass the main process check if and only if `reload=True`?

From https://github.com/zauberzeug/nicegui/issues/189#issuecomment-1340019477, it appears that "the if-statement in `ui.run` prevents the spawned process from starting another server". 

There won't be another process which is spawned in the first place, if `reload=False`. 

Then, essentially, we are baking in the "optimal defaults" for the `require_main_process` of `ui.run`, which is `True` when `reload=True`, `False` when `reload=False`

davetapley:
@evnchn yep, that makes total sense. Thanks for clarifying.

If good with @rodja I will update my PR to remove `require_main_process` and update the early return check to:
```py
if reload and multiprocessing.current_process().name != 'MainProcess':
```

falkoschindler:
@davetapley Unfortunately this brakes `run.cpu_bound` in combination with `reload=False`:
```py
from nicegui import ui, run
import time

def compute_sum(a: float, b: float) -> float:
    time.sleep(1)  # simulate a long-running computation
    return a + b

async def handle_click():
    result = await run.cpu_bound(compute_sum, 1, 2)
    ui.notify(f'Sum is {result}')

ui.button('Compute', on_click=handle_click)

ui.run(reload=False)
```
The child process is not "MainProcess", so we want to exit early from `ui.run`.

evnchn:
To clarify: 

Nothing to do with #3644 _"I regret mentioning force=true / #3644 in the PR 😁"_

Nothing to do with #4627 and the likes: We are not doing native mode, and there are no config to pass. 

Tangentially related to #189 only, as that regards _only one of the reason_ why the main process check was required in the first place: 

> If NiceGUI runs with auto-reload enabled, the uvicorn server spawns a new process to serve the app. When an auto-reload is triggered, this process is re-started. The if-statement in ui.run prevents the spawned process from starting another server. 

The other reason being: `run.cpu_bound` https://github.com/zauberzeug/nicegui/pull/4613#issuecomment-2808850110

@davetapley Kindly check if I am right. React 👍 if so.

rodja:
In #4706, @mrasore98 describes that uv is also struggling with `if multiprocessing.current_process().name != 'MainProcess':`.

mrasore98:
> In #4706, @mrasore98 describes that uv is also struggling with `if multiprocessing.current_process().name != 'MainProcess':`.

@rodja I tested the uv workflow using the source branch for this PR and was able to successfully launch the app, so this change solves my problem as well (using native=True, reload=False)!

After reading through some of the background for the MainProcess check, it seems like it is needed for the correct uvicorn behavior in the `reload=True` case. If my understanding is correct, should there be a check in this PR to ensure that when `require_main_process` is False `reload` is also False? Maybe set `reload = require_main_process and reload`?

rodja:
@davetapley by using `ui.run_with` you can launch NiceGUI in a subprocess without any modifications to NiceGUI:

```py
from multiprocessing import get_context
import time
import uvicorn
from fastapi import FastAPI
from nicegui import ui

def run_gui():
    app = FastAPI()
    ui.label('Hello')
    ui.run_with(app, mount_path='/')
    uvicorn.run(app, host='0.0.0.0', port=8000)

if __name__ == '__main__':
    ctx = get_context('spawn')
    gui_process = ctx.Process(target=run_gui)
    gui_process.start()

    time.sleep(2)
    user_input = input('Confirm to stop the GUI process')

    gui_process.terminate()
    gui_process.join()
```

rodja:
Because @mrasore98 also [does not need `require_main_process`](https://github.com/zauberzeug/nicegui/discussions/4706#discussioncomment-13136015) I think there is no reason to add this flag. What do you think @davetapley?

davetapley:
@rodja yep, perfect, thanks! 🙏🏻

PR #4616, opened 2025-04-14
Add new Leaflet elements

jeffmelville:
These changes add:
- Polyline, CircleMarker, and Circle elements to Leaflet
- Add the ability to override the map URL template at construction

### DIFF
--- a/nicegui/elements/leaflet.py
+++ b/nicegui/elements/leaflet.py
@@ -29,6 +29,7 @@ class Leaflet(Element, component='leaflet.js', default_classes='nicegui-leaflet'
                  zoom: int = 13,
                  *,
                  options: Dict = {},  # noqa: B006
+                 url_template = None,
                  draw_control: Union[bool, Dict] = False,
                  hide_drawn_items: bool = False,
                  additional_resources: Optional[List[str]] = None,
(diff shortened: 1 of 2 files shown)


jeffmelville:
Yup can do

evnchn:
Leaflet may soon release 2.0

https://github.com/zauberzeug/nicegui/discussions/4655#discussioncomment-13019330

Your commits may be impacted by upstream changes in some manner. I didn't look too deep into it (assuming you're using every bit of the source code in the integration, so all changes could affect yoy. 

Lets be cautious. 

---

Polyline

{6FFA7109-9198-4D08-A845-8F522AA668BA}

CircleMarker

{1BB2B1CE-9761-4139-A2A0-64101E127986}

Circle

{3A01BE51-3364-457B-9F8A-E34DB7CE4EA3}

falkoschindler:
I think these new features won't add any extra incompatibilities for Leaflet 2.0. We'll have to migrate `ui.leaflet` anyway.

@jeffmelville Are you still considering adding documentation? Pytests are probably difficult to write and we don't have any for other map elements.

evnchn:
I would recommend to close and try again since our 3.0 JS dependencies changed quite a lot.

falkoschindler:
@evnchn You're probably right. The merge conflict is pretty nasty compared to the relatively small diff of this PR. Let's close it for now.

PR #2536, opened 2024-02-11
feat: javascript event listener

WSH032:
Discuss: https://github.com/zauberzeug/nicegui/discussions/2383#discussioncomment-8184894

PR #4618, opened 2025-04-15
feat: support delegated event handling

yihuang:
1. Allow setting `handler` and `js_handler` at the same time, the `js_handler` can be used to override the default behavior at js side and has the ability to emit event to call the `handler` at server side, specifically, it can change how event args are constructed.
2. Provide utility function `emitEventToSelf` at js side, only available in `js_handler` context

Please check the unit test case for usage:

https://github.com/zauberzeug/nicegui/pull/4618/files#diff-17a9b99a7b95f217c0c8d0f8d0c9d852368eb8e2e14883d26b443483e1b7597fR196-R212

---

Feature request: #4611

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -83,6 +83,15 @@ function emitEvent(event_name, ...args) {
   getElement(0).$emit(event_name, ...args);
 }
 
+function emitEventTo(id, event, ...args) {
+  window.socket?.emit("event", {
+    id: id,
+    client_id: window.clientId,
+    listener_id: event.listener_id,
+    args: stringifyEventArgs(args, event.args),
+  });
+}
+
 function stringifyEventArgs(args, event_args) {
   const result = [];
   args.forEach((arg, i) => {
@@ -182,13 +191,7 @@ function renderRecursively(elements, id) {
       handler = eval(event.js_handler);
     } else {
       handler = (...args) => {
-        const emitter = () =>
-          window.socket?.emit("event", {
-            id: id,
-            client_id: window.clientId,
-            listener_id: event.listener_id,
-            args: stringifyEventArgs(args, event.args),
-          });
+        const emitter = () => emitEventTo(id, event, ...args);
         const delayed_emitter = () => {
           if (window.did_handshake) emitter();
           else setTimeout(emitter, 10);
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -83,15 +83,6 @@ function emitEvent(event_name, ...args) {
   getElement(0).$emit(event_name, ...args);
 }
 
-function emitEventTo(id, event, ...args) {
-  window.socket?.emit("event", {
-    id: id,
-    client_id: window.clientId,
-    listener_id: event.listener_id,
-    args: stringifyEventArgs(args, event.args),
-  });
-}
-
 function stringifyEventArgs(args, event_args) {
   const result = [];
   args.forEach((arg, i) => {
@@ -186,21 +177,28 @@ function renderRecursively(elements, id) {
     let event_name = "on" + event.type[0].toLocaleUpperCase() + event.type.substring(1);
     event.specials.forEach((s) => (event_name += s[0].toLocaleUpperCase() + s.substring(1)));
 
+    let emitEventToSelf = function(...args) {
+      const emitter = () => window.socket?.emit("event", {
+        id: id,
+        client_id: window.clientId,
+        listener_id: event.listener_id,
+        args: stringifyEventArgs(args, event.args),
+      });
+      const delayed_emitter = () => {
+        if (window.did_handshake) emitter();
+        else setTimeout(emitter, 10);
+      };
+      throttle(delayed_emitter, event.throttle, event.leading_events, event.trailing_events, event.listener_id);
+      if (element.props["loopback"] === False && event.type == "update:modelValue") {
+        element.props["model-value"] = args;
+      }
+    }
+
     let handler;
     if (event.js_handler) {
       handler = eval(event.js_handler);
     } else {
-      handler = (...args) => {
-        const emitter = () => emitEventTo(id, event, ...args);
-        const delayed_emitter = () => {
-          if (window.did_handshake) emitter();
-          else setTimeout(emitter, 10);
-        };
-        throttle(delayed_emitter, event.throttle, event.leading_events, event.trailing_events, event.listener_id);
-        if (element.props["loopback"] === False && event.type == "update:modelValue") {
-          element.props["model-value"] = args;
-        }
-      };
+      handler = emitEventToSelf;
     }
 
     handler = Vue.withModifiers(handler, event.modifiers);
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -177,13 +177,14 @@ function renderRecursively(elements, id) {
     let event_name = "on" + event.type[0].toLocaleUpperCase() + event.type.substring(1);
     event.specials.forEach((s) => (event_name += s[0].toLocaleUpperCase() + s.substring(1)));
 
-    let emitEventToSelf = function(...args) {
-      const emitter = () => window.socket?.emit("event", {
-        id: id,
-        client_id: window.clientId,
-        listener_id: event.listener_id,
-        args: stringifyEventArgs(args, event.args),
-      });
+    let emitEventToSelf = (...args) => {
+      const emitter = () =>
+        window.socket?.emit("event", {
+          id: id,
+          client_id: window.clientId,
+          listener_id: event.listener_id,
+          args: stringifyEventArgs(args, event.args),
+        });
       const delayed_emitter = () => {
         if (window.did_handshake) emitter();
         else setTimeout(emitter, 10);
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -320,26 +320,6 @@ class Element(Visibility):
             Tooltip(text)
         return self
 
-    @overload
-    def on(self,
-           type: str,  # pylint: disable=redefined-builtin
-           *,
-           js_handler: Optional[str] = None,
-           ) -> Self:
-        ...
-
-    @overload
-    def on(self,
-           type: str,  # pylint: disable=redefined-builtin
-           handler: Optional[events.Handler[events.GenericEventArguments]] = None,
-           args: Union[None, Sequence[str], Sequence[Optional[Sequence[str]]]] = None,
-           *,
-           throttle: float = 0.0,
-           leading_events: bool = True,
-           trailing_events: bool = True,
-           ) -> Self:
-        ...
-
     def on(self,
            type: str,  # pylint: disable=redefined-builtin
            handler: Optional[events.Handler[events.GenericEventArguments]] = None,
@@ -360,9 +340,6 @@ class Element(Visibility):
         :param trailing_events: whether to trigger the event handler after the last event occurrence (default: `True`)
         :param js_handler: JavaScript code that is executed upon occurrence of the event, e.g. `(evt) => alert(evt)` (default: `None`)
         """
-        if handler and js_handler:
-            raise ValueError('Either handler or js_handler can be specified, but not both')
-
         if handler or js_handler:
             listener = EventListener(
                 element_id=self.id,
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -191,3 +191,22 @@ def test_js_handler(screen: Screen) -> None:
     screen.open('/')
     screen.click('Button')
     screen.should_contain('Click!')
+
+
+def test_delegated_event(screen: Screen) -> None:
+    data = [f'Item {i}' for i in range(5)]
+    snippet = ''.join(f'<li data-index={i}>{item}</li>' for i, item in enumerate(data))
+    root = ui.html(snippet, tag='ul')
+
+    clicked = []
+    def on_click(e):
+        clicked.append(int(e.args['index']))
+
+    root.on('click', on_click, js_handler=f'(e) => emitEventTo({root.id}, event, e.target.dataset)')
+
+    # test
+    screen.open('/')
+    for i, _ in enumerate(data):
+        screen.selenium.find_element(By.CSS_SELECTOR, f'li:nth-child({i+1})').click()
+
+    assert clicked == list(range(len(data)))
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -202,7 +202,7 @@ def test_delegated_event(screen: Screen) -> None:
     def on_click(e):
         clicked.append(int(e.args['index']))
 
-    root.on('click', on_click, js_handler=f'(e) => emitEventTo({root.id}, event, e.target.dataset)')
+    root.on('click', on_click, js_handler="(e) => emitEventToSelf(e.target.dataset)")
 
     # test
     screen.open('/')


nicegui/static/nicegui.js
```diff
@@ -101,11 +101,35 @@ function stringifyEventArgs(args, event_args) {
...
+        return v;
+    }));
   });
   return result;
 }
 
+function nodeArgs(node) {
+    return {
+        cid: findComponentId(node),
+        tag: node.tag,
```
yihuang, line 117:
```suggestion
        tag: node.tagName,
```

evnchn:
Hello @yihuang. Thank you for the PR. It makes things much more clear what you are trying to do. 

Correct me if I am wrong by @ me, for these following points: 

Benefits
1. A bunch of NiceGUI elements inside a parent, for which they all share a common behaviour, and it would be a bit insane to attach individual click handlers for each of them. 
2. Raw HTML elements, for which it is currently impossible* to attach NiceGUI event handler to them without this PR. 

Details: 
1. For achieving benefit 1, the `nodeArgs` `findComponentId` finds the NiceGUI element ID right away, and we can do work based on such ID. 
2. For achieving benefit 2, the `nodeArgs` `findComponentId` finds the parent NiceGUI element only. Thus, the `node.tagName` and `node.dataset` make it possible to identify the clicked raw HTML element by the tag name and the `data-*` fields in the properties. 
3. `findComponentId` always returns an integer. The id, if found, or -1 if not found. 

Impact: 
1. The PR only adds data returned to NiceGUI server, but does not take away any existing data (`undefined -> nodeArgs(v)`), so backwards compatibility is a guarantee. 

Collisions with existing PRs:
1. Area of code edit collides with #4579, but should still be fine, because you are regarding the `Window` and `Node` type, while I regard the `Touch` type. 

*may be possible via browser-side `onclick` and `emitEvent`, but a rather nasty solution.

yihuang:
Yeah, you summarize it very well, I have nothing to add 👍

falkoschindler (commented):
Thanks for the pull request, @yihuang!

I was sceptical at first, because adding payload to (almost) every event could impact the performance for all user interactions. But an ID, a tag name and the dataset should be reasonably small.

When handling rapid events like "mousemove", we have to traverse the DOM each time to find the closest component. If there was a way to identify it without a regex, we could drop the while loop and use a CSS attribute selector:
```js
component = v.closest('[id^="c"]')
```
But relying on a leading "c" would probably be too sketchy as the only criterion.

@evnchn What do you think about the impact on performance and bandwidth?

Last but not least, if we agree to merge this feature, we should add a pytest (test_events.py) and a short demo (generic_events_documentation.py).

evnchn:
@falkoschindler Thank you for the new mindset on performance issue. It really turned on a part of my brain which was off until now. 🤯

Yes. Navigating the DOM to "hunt for the ID" is a terrible idea. The DOM is always slow, it's a fact. https://stackoverflow.com/questions/6817093/but-whys-the-browser-dom-still-so-slow-after-10-years-of-effort

My insight: **We need only the `node.dataset` to identify the element, if whoever creates the element takes care to populate `data-cid`**, which isn't hard. 

Check this out: 
```py
from nicegui import ui

data = [f'Item {i}' for i in range(5)]

def notify_clicked(event):
    cid = event.args['target']['cid']
    data_index = event.args['target']['data'].get('index')
    data_cid = event.args['target']['data'].get('cid')
    ui.notify(f'Clicked traceback: cid={cid}, data-cid={data_cid}, data-index={data_index}')

# Use component for each child, the component id map to the child
ui.label("Use component for each child, the component id map to the child").classes('text-2xl')
root = ui.element('ul').on('click', lambda e: notify_clicked(e))
with root:
    for i, item in enumerate(data):
        props_str = f'data-index={i} data-cid={ui.context.client.next_element_id}'
        ui.html(item, tag='li').props(props_str)

# Use a single component for all children, the component id map to the parent
ui.label("Use a single component for all children, the component id map to the parent").classes('text-2xl')
snippet = ''.join(f'<li data-index={i} data-cid={ui.context.client.next_element_id}>{item}</li>' for i, item in enumerate(data))
ui.html(snippet, tag='ul').on('click', lambda e: notify_clicked(e))


ui.run(reload=False, show=False)
```

> Always use `data-cid={ui.context.client.next_element_id}` to "predict the next ID" and then attach to the upcoming element right away. If you are not comfortable with that, you can always `my_html = ui.html(tag='ul')` then `my_html.set_content` later.

{9B57F840-1CD4-42F5-A41A-3C78D9F471D6}

With that big part of time-spender out of the way, the rest (accessing the `node.dataset`, and putting it into the JSON payload) should be minimal and non-impactful. 

Optionally, we can get rid of `node.tagName` as well, since a similar argument applies. Then, we can achieve **zero-side-effect addition** of this feature, if `nodeArgs` returns `undefined` if no dataset, or return just the `{data: node.dataset}` if there is.

evnchn:
> One way to mitigate this issue is allow user to mask out some events for this feature.

@falkoschindler's point is that, navigating the DOM 60 times per second (higher if your screen's refresh rate is high) is not a good idea. I don't think he minds `mousemove` if we don't touch the DOM, since the element is already returned to us in JS-land on the mousemove event, so it isn't slow. 

So, we don't need to be so conservative, if we remove the DOM component searching already. If the library user complains that the `mousemove ` is slow after the PR, they can just, y'know, _not include so many data-* attributes_ or better _write their own js_handler which is much more lean than serializing the entire `mousemove` event_ (Consider: https://github.com/zauberzeug/nicegui/issues/3762#issuecomment-2370196275 https://github.com/orgs/zauberzeug/projects/6/views/1?pane=issue&itemId=82940887)

falkoschindler:
@evnchn Even though I agree with you in principle, the StackOverflow question (from 13 years ago) might be a bit drastic. For our application, we're not modifying the DOM, so no visible changes have to be computed. We only access the parent node two or three times until the regex matches and we exit the loop. It's not nothing, but not "a terrible idea" either.

Anyway, I'm a fan of questioning API changes. Reducing the additional attributes to the dataset is an interesting idea. But it feels more like a workaround than a really nice API for "delegated event handling": Place a data attribute via `.props()` to identify your element and peel it out of `e.args` when handling an event. With similar effort you can use a `js_handler` to process the event on the client and emit what you really need to the server:

```py
with ui.card().on('click', js_handler='(e) => emitEvent("item-clicked", e.target.dataset.index)'):
    ui.on('item-clicked', lambda e: ui.notify(f'Clicked on {e.args}'))
    for i in range(10):
        ui.label(f'Item {i}').props(f'data-index={i}')
```

evnchn:
Hmm interesting. 

I didn't know that the `emitEvent` can take extra argument other than the obvious first one. 

Consider the code search on the website source code: https://github.com/search?q=repo%3Azauberzeug%2Fnicegui+emitEvent+path%3A%2F%5Ewebsite%5C%2Fdocumentation%5C%2Fcontent%5C%2F%2F&type=code

{60A15CE2-A320-4613-B49E-657EC63EE84D}

With that, I am now more on the side to believe that **the existing API is enough for what we need**, and this PR may not be necessary. 

Rather, we may want to put your code into the documentation / FAQ / Wiki, whichever you think see fit.

yihuang:
it'll be ideal if we can `emitEvent` to the current component.

evnchn:
Oh wow. I did NOT know `emitEvent` was as simple as `getElement(0).$emit(event_name, ...args);`. 

Seems like we have ought to document this better, and I like this solution much more than including the dataset on `nicegui.js`

yihuang:
> Oh wow. I did NOT know `emitEvent` was as simple as `getElement(0).$emit(event_name, ...args);`.
> 
> Seems like we have ought to document this better, and I like this solution much more than including the dataset on `nicegui.js`

`getElement(id).$emit()` need the integer component id, so we need to access the current integer component id, or use regex to extract from `this.id`, might need a little bit support from library to make it clean.

falkoschindler:
At the risk of ruining the fun, I have to ask: What is actually wrong with defining event handlers _per_ clickable element?
```py
with ui.card():
    for i in range(10):
        ui.label(f'Item {i}').on('click', lambda i=i: ui.notify(f'Clicked on {i}'))
```

yihuang:
```python
from nicegui import ui
with ui.card() as c:
    c.on('click', js_handler=f'(e) => getElement({c.id}).$emit("item-clicked", e.target.dataset.index)')
    c.on('item-clicked', lambda e: ui.notify(f'Clicked on {e.args}'))
    for i in range(10):
        ui.label(f'Item {i}').props(f'data-index={i}')
ui.run()
```

the example works now.

yihuang:
> At the risk of ruining the fun, I have to ask: What is actually wrong with defining event handlers _per_ clickable element?
> 
> ```python
> with ui.card():
>     for i in range(10):
>         ui.label(f'Item {i}').on('click', lambda i=i: ui.notify(f'Clicked on {i}'))
> ```

In my case, I was trying to hack markdown which is a single complex component, with this I can manage the event handlings on the component at server side, where it has access the original parsed document.

evnchn:
> At the risk of ruining the fun, I have to ask: What is actually wrong with defining event handlers per clickable element?

To elaborate on @yihuang's above comment, and to make it clearer...

Read carefully what @yihuang initially posted: 

```py
snippet = ''.join(f'<li data-index={i}>{item}</li>' for i, item in enumerate(data))
ui.html(snippet, tag='ul').on('click', lambda e: ui.notify(f'Clicked on {clicked(e)}'))
```

And my description of it: 

> 2. Raw HTML elements, for which it is currently impossible* to attach NiceGUI event handler to them without this PR.
> 
> *may be possible via browser-side `onclick` and `emitEvent`, but a rather nasty solution.

So it's not a "fun" thing, but a "necessity" thing. 

---

You may ask why is the above a "necessity" in the first place. 

Obviously maybe for some lower overhead because NiceGUI elements have quite some overhead, and if you have, say, 10000s of elements, stuff gets slow real fast, so a native HTML element is decent speedup (used it in my private project). 

But on a real feature-add, recall @yihuang is trying to "hack extra functionality into the markdown component" in https://github.com/zauberzeug/nicegui/discussions/4611#discussioncomment-12810305 -> #4606. Most certainly the elements defined inside the markdown are NOT NiceGUI elements and do NOT support `.on('click', ...)`, and he want click interactivity @yihuang correct me if I am wrong or react 👀 otherwise. 

And so the idea was, attach one big event handler to the `ui.markdown`, and then that handler must have some pretty awesome logic to "figure out what was clicked inside the `ui.markdown`" and respond appropriately.

evnchn:
@yihuang major correction

> `getElement(id).$emit()` need the integer component id

Check out #4675. Before then, `getElement` only works with a numeric ID. Afterwards, you can pass in the HTMLElement reference to the element, and it'll "figure out the ID and do the lookup for you".

evnchn:
With regards to `this`, I think we got a train wreck going on 🚨🚨🚨. @yihuang (starter) @falkoschindler (follower) 

{4503C028-41BE-4A0F-84EE-5713CF1A2885}

```py
with ui.card().on('click', js_handler='(e) => {console.log(this);console.log(self);}'):
    ui.label('Click me!')
```

Both `this` and `self` is `window` since the script is attached by `nicegui.js`, not like `<button onclick="myFunction()">Click me</button>` or something. 

So: 

https://github.com/zauberzeug/nicegui/pull/4618#issuecomment-2840291735 @yihuang 

See `getElement(getComponentId(this.id))`? `this.id` is `undefined`. Ain't gonna work. 

https://github.com/zauberzeug/nicegui/pull/4618#issuecomment-2840312814 @falkoschindler 

See `this.element`? Also `undefined`. Ain't gonna work. You also can't assign to `window.element` on a by-instance basis, since otherwise you set the variable `element`, which I doubt is your intention.

yihuang:
@evnchn [this](https://github.com/zauberzeug/nicegui/pull/4618#issuecomment-2840328320) is the working example

evnchn:
Hey. We have another train wreck here 🚨🚨🚨. It's with regards to `$emit` on non-Vue components, leading to the original `ui.html` use case being broken. 

TL-DR: 

@yihuang Just because `emitEvent` is `getElement(0).$emit(event_name, ...args)` does not mean you can `getElement(xxx).$emit`, because some elements don't support. So your snippet at https://github.com/zauberzeug/nicegui/pull/4618#issuecomment-2840291735 is wrong at 2 levels. Think and check carefully before you type. We would hope we can rely safely on your results without having to double-check everything. 

@falkoschindler Not sure if this is a bug or intended behaviour. Do you think this is worth an investigation? But `$emit` is a Vue thing (https://vuejs.org/guide/components/events#emitting-and-listening-to-events) and I am not sure on performance / compatibiity aspects, if we make everything, even every little `div`, to become a Vue component in some capacity. 

---

```py
from nicegui import ui

snippet = ''.join(f'<li data-index={i}>Item {i}</li>' for i in range(5)) # it's the same snippet always

ui.label("Emit event globally (working)").classes('text-2xl')

ui.on('global-item-clicked', lambda e: ui.notify(f"Received custom item-clicked event (globally), e.args={e.args}"))

my_click_reactive_html_global1 = ui.html(snippet, tag='ul')
js_handler_string1 = f'(e) => {{emitEvent("global-item-clicked", e.target.dataset.index)}}'
ui.label(f"js_handler_string1: {js_handler_string1}")
my_click_reactive_html_global1.on('click', js_handler=js_handler_string1)

ui.label("Emit event locally with ID and getElement (not working)").classes('text-2xl')

my_click_reactive_html_global2 = ui.html(snippet, tag='ul')
my_click_reactive_html_global2.on("local-item-clicked", lambda e: ui.notify(f"Received custom item-clicked event (locally), e.args={e.args}"))
js_handler_string2 = f'(e) => {{getElement({my_click_reactive_html_global2.id}).$emit("local-item-clicked", e.target.dataset.index)}}'
ui.label(f"js_handler_string2: {js_handler_string2}")
my_click_reactive_html_global2.on('click', js_handler=js_handler_string2)

ui.label("Emit event locally with ID and mounted_app.elements[id] (not working)").classes('text-2xl')

my_click_reactive_html_global3 = ui.html(snippet, tag='ul')
my_click_reactive_html_global3.on("local-item-clicked", lambda e: ui.notify(f"Received custom item-clicked event (locally), e.args={e.args}"))
js_handler_string3 = f'(e) => {{mounted_app.elements[{my_click_reactive_html_global3.id}].$emit("local-item-clicked", e.target.dataset.index)}}'
ui.label(f"js_handler_string3: {js_handler_string3}")
my_click_reactive_html_global3.on('click', js_handler=js_handler_string3)

my_button = ui.button("This button has $emit, by the way")
ui.label(f"That button had ID {my_button.id}")

ui.run(port=9999, show=False, reload=False)
```






It would appear that, there is no `$emit` for elements such as `ui.label`, `ui.html`. 

---

Expanding further, I dug this out, because @yihuang mentioned `getElement(0).$emit(event_name, ...args);` was how `emitEvent` worked, and `getElement` is indexing into `mounted_app.$refs["r" + _id]` under the hood. 

{E1A10565-AF3C-42A0-8378-4C62E8C820A5}

Check the source code, what is passed into `createApp`: 

```json
{
    "0": {"tag": "q-layout", ...},
    "1": {"tag": "q-page-container", ...},
    "2": {"tag": "q-page", ...},
    "3": {"tag": "div", ...},
    "4": {"tag": "div", ...},
    "5": {"tag": "ul", ...},
    "6": {"tag": "div", ...},
    "7": {"tag": "div", ...},
    "8": {"tag": "ul", ...},
    "9": {"tag": "div", ...},
    "10": {"tag": "div", ...},
    "11": {"tag": "ul", ...},
    "12": {"tag": "div", ...},
    "13": {"tag": "q-btn", ...},
    "14": {"tag": "div", ...}
}
```

Notice the trend? If the tag is a native browser tag, Vue does nothing, and mounted_app.$refs and mounted_app.elements gets to have the raw browser element, no `$emit` attached. Only if it is a Vue component tag (like `q-layout`) then it gets to have `$emit`

evnchn:
So, in the end, if we want to do point 2 at the start: "Raw HTML elements, for which it is currently impossible* to attach NiceGUI event handler to them without this PR." without any library modification, as of now, only possibility is to leverage `js_handler`, **global `emitEvent` and `ui.on`**. All that `getElement` and `this` and `getComponentId` / `findComponentId` was smokescreen.

yihuang:
nice dig, I thought nicegui component map 1:1 to vue component, apparently it's not. But maybe emititng events to component id may still work, just sending the event message and dispatch at server side. I have to dig later.

yihuang:
@evnchn here it is, the basic idea is to allow `handler` and `js_handler` exists at the same time, so user can use the `js_handler` to override the default event args construction:

```diff
diff --git a/nicegui/element.py b/nicegui/element.py
index 6962425f..7c3c7b53 100644
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -360,9 +360,6 @@ class Element(Visibility):
         :param trailing_events: whether to trigger the event handler after the last event occurrence (default: `True`)
         :param js_handler: JavaScript code that is executed upon occurrence of the event, e.g. `(evt) => alert(evt)` (default: `None`)
         """
-        if handler and js_handler:
-            raise ValueError('Either handler or js_handler can be specified, but not both')
-
         if handler or js_handler:
             listener = EventListener(
                 element_id=self.id,
```

```python
from nicegui import ui
data = [f'Item {i}' for i in range(10)]
snippet = ''.join(f'<li data-index={i}>{item}</li>' for i, item in enumerate(data))
with ui.html(snippet, tag='ul') as c:
    c.on('click', lambda e: ui.notify(f'Clicked on {e.args["index"]}'), js_handler=f'''(e) => {{
        window.socket?.emit("event", {{
            id: {c.id},
            client_id: window.clientId,
            listener_id: event.listener_id,
            args: stringifyEventArgs([e.target.dataset], null),
          }});
     }}''')
ui.run()
```

falkoschindler:
>  the basic idea is to allow `handler` and `js_handler` exists at the same time, so user can use the `js_handler` to override the default event args construction.

I had a very similar idea: What if we implement an `args_filter` parameter (see https://github.com/orgs/zauberzeug/projects/6?pane=issue&itemId=82940887). It would behave very similar to a `js_handler` using `emitEventToSelf`, but might feel more intentional.
```py
with ui.card().on('click', lambda e: ui.notify(f'Clicked on {e.args}'),
                  args_filter='(e) => e.target.dataset.index'):
    ...
```
We could also re-use the existing `args` parameter, which is currently of type `Union[None, Sequence[str], Sequence[Optional[Sequence[str]]]]`. But maybe this is confusing.

evnchn:
@yihuang your newest approach looks **really good** to me! 💪

Here are a couple of observations: 
* Behaviour of `nicegui.js` shall be **exactly the same** as old releases if the new feature is not active*
* Should it be used, `emitEventToSelf` is faithful to the original implementation. Such being, delay until `window.did_handshake` set, using `throttle` properly (and potentially `element.props["loopback"]` handling if `event.type == "update:modelValue"`, though not sure why that would be the case that you attach `js_handler` to such event type)**
* Essentially enables what we want for `args_filter` (see below)

@falkoschindler I feel like at this point `args_filter` is just a convenience method / syntactic sugar, where `args_filter='(e) => BLAHBLAHBLAH'` = `js_handler='(e) => emitEventToSelf(BLAHBLAHBLAH)'`. Since that would take away the possibility of, say, sliding in a `console.log` to debug stuff and/or extra JS functionality, I feel like we should not do it? 

---

*Am not sure about that but it looks to be that way. Minor concern on why are the two `@overload` gone? From https://github.com/zauberzeug/nicegui/pull/2536, appears to be for type hinting. @falkoschindler I think you are more proficient in this. Will it cause any troubles? 

**I was about to propose expanding the functionality of `emitEvent` to allow it to emit event on behalf of any element given ID, beyond ID=0 (the root element, aka `ui.on`). But that (1) lowers the barrier to security concern if people use this and emit a clicked event to a disabled element for example, (2) we have to re-interpret everything from `mounted_app.elements[8].event`, with the camel-case "type" being the biggest pain, since NiceGUI uses kebab-case internally. 

{D55EEED2-A1CE-4531-845B-71D64D4B1BC0}

P.S.: I recommend starting over in another PR next time, if we are changing directions.

falkoschindler:
~~I don't understand why so much code is gone. Shouldn't `element.py` remain unchanged?~~

Oh, I see! You're allowing to register a Python and a JavaScript handler at the same time!

```py
root.on('click', handle_click, js_handler='(e) => emitEventToSelf(e.target.dataset)')
```

In contrast, this is the same line with `args_filter`:
```py
root.on('click', handle_click, args_filter='(e) => e.target.dataset')
```

Actually, I lean towards calling it `args` like the already existing parameter:
```py
root.on('click', handle_click, args='(e) => e.target.dataset')
```

I'm struggling with the `emitEventToSelf` approach: The user has to understand that only `js_handler` is evaluated on the client, emits an event to the server, where `handle_click` is called with the emitted event arguments. In contrast, using `args_filter` or `args` reads more like "let's define which arguments to send", which is exactly what both solutions are doing.

With the `args` approach we wouldn't need an `emitEventToSelf` function because `args` is directly handled in `stringifyEventArgs`.

If you want to do more than filtering arguments, like logging to the console, you can use a separate event with `js_handler` or simply log within the Python handler.

Note that either way this PR could solve #4577 where we struggle with sending touch events. Simply defining `args='e => e.touches[0]'` would be all it takes.

evnchn:
> lean towards calling it `args` like the already existing parameter

Then args would be of type `Union[None, str, Sequence[str], Sequence[Optional[Sequence[str]]]]`, where the `str` possibility corresponds to attaching a JS (preferably arrow because it clean) function which is applied before `emitEventToSelf`, then. 

Is this too complicated? Since we are re-wiring the args to go to JS execution. 

Perhaps we can have `js_filter` which under-the-hood sets `js_handler` as it stands. Also in retrospect `args_filter` seems a bit unclear that we are doing filtering via JS, which enables much more operation that just, y'know, filtering. 

---

I also want to correct what I was saying: We can still do JS operations such as `console.log`, and computations. Consider the following:

{73CB5D34-0105-4C98-9B96-32B8BE5226F0}

evnchn:
Lets see. 

Fix #3762: Do `e.touches[0]` to avoid sending over the rest of the point data. 

Fix #4611 and fix #4606: Enables attaching event handler to non-NiceGUI elements much more easily, useful for hacking markdown component to do extra stuff. Do `e.target.dataset.index`, for example

Fix #4577: You can manually select stuff from `e.touches[0]` to send over

Fix #4589: You can manually select `e.event.button` to send over

---

Seems like this PR will be a big one. @falkoschindler Should we treat as `Priority = Major` on this PR?

evnchn:
I **strongly believe** there has to be JS somewhere in the name, so that it is super obvious the logic is executed via JS on the client, not on Python. 

Otherwise the nature of this new feature is not obvious, potentially leading to under-utilization, and **maybe lead to Arbitrary Code Execution** if the library user is not highly unaware of the fact that it's JS under the hood and pass user-controllable input (like a f-string). 

Of which, `js_filter` is the one I like the most. Since we came to filter stuff, so that should be made obvious.

yihuang:
What's interesting about current approach is, it don't add a new field, instead it actually supersedes several existing fields `args`, `throttle`, `leading_events`, `trailing_events`, they can all be passed through the `js_handler` instead.

yihuang:
Closing this one in favor of js_handler focused one #4689

Issue #4617, opened 2025-04-14
ui.timer causes memory increase overtime


x2a-packet:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui
import datetime

@ui.page('/')
def index_page() -> None:
    def update_time_display():
        current_time = datetime.datetime.now().strftime('%H:%M:%S')
        header_time_display.set_text(current_time)
        date_text = f"Date: {datetime.datetime.now().strftime('%d.%m.%Y')}<br> Servertime: {current_time}"
        timeout_timestamp = None
        if timeout_timestamp:
            timeout_text = "Session Timeout: XYZ"
            #timeout_text = f"Session Timeout: {time.strftime('%H:%M:%S', time.localtime(timeout_timestamp))}"
           
        else:
            timeout_text = "Session Timeout: N/A"

        tooltip_content.set_content(f"{date_text}<br>{timeout_text}")
        
    
    with ui.element('div').style('position: absolute; top: 14px; right:400px; z-index: 1;') as element:
        with ui.label(datetime.datetime.now().strftime('%H:%M:%S')).classes('font-bold') as header_time_display:
            with ui.tooltip(), ui.html("") as tooltip_content:
                    pass

    ui.timer(1, update_time_display)
    
    ui.label('Some content!')

ui.run(reload=False)
```

### Description

Start the app and access '/'.
Surveile the RSS usage of the app with external tools like ps on linux.
I observe a steady increase when the timer update_time_display is activated.
When the timer is not active, there is no grow observable.
The attached screenshots show the both scenarios.

Why is it behaving like that? 

### NiceGUI Version

2.12.1, 2.14.1

### Python Version

Python 3.10.12

### Browser

Chrome

### Operating System

Linux

### Additional Context

Good:

![Image](https://github.com/user-attachments/assets/8be4c571-00ab-4b8c-829b-098666c02b46)
Bad:

![Image](https://github.com/user-attachments/assets/300b7348-7a02-4b26-bf3b-c710fd367c7c)

evnchn:
@x2a-packet thanks for the issue report. 

[...]

PR #4619, opened 2025-04-15
Use a WeakValueDictionary instead of custom finalizers

falkoschindler:
This PR tries to solve the memory leak #4617 as well as the RuntimeError #4419 using a WeakValueDictionary as proposed by @polyipseity. With the code from https://github.com/zauberzeug/nicegui/issues/4617#issuecomment-2804047378 the memory doesn't grow anymore. And the RuntimeError shouldn't be possible without our custom finalizer.

I think there might still be leaks related to the `bindings` dictionary which holds strong references to source and target objects. But the problems described by the two issues mentioned above should be fixed.

@chriswi93 @evnchn @polyipseity @x2a-packet Would you like to have a look and/or try it out?

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -11,7 +11,6 @@ from typing import (
     Any,
     Callable,
     DefaultDict,
-    Dict,
     Iterable,
     List,
     Mapping,
@@ -34,7 +33,7 @@ if TYPE_CHECKING:
 MAX_PROPAGATION_TIME = 0.01
 
 bindings: DefaultDict[Tuple[int, str], List] = defaultdict(list)
-bindable_properties: Dict[Tuple[int, str], weakref.finalize] = {}
+bindable_properties: weakref.WeakValueDictionary[Tuple[int, str], Any] = weakref.WeakValueDictionary()
 active_links: List[Tuple[Any, str, Any, str, Callable[[Any], Any]]] = []
 
 TC = TypeVar('TC', bound=type)
@@ -179,7 +178,7 @@ class BindableProperty:
             return
         setattr(owner, '___' + self.name, value)
         key = (id(owner), str(self.name))
-        bindable_properties.setdefault(key, weakref.finalize(owner, lambda: bindable_properties.pop(key, None)))
+        bindable_properties[key] = owner
         _propagate(owner, self.name)
         if value_changed and self._change_handler is not None:
             self._change_handler(owner, value)
@@ -204,10 +203,9 @@ def remove(objects: Iterable[Any]) -> None:
         ]
         if not binding_list:
             del bindings[key]
-    for (obj_id, name), finalizer in list(bindable_properties.items()):
+    for obj_id, name in list(bindable_properties):
         if obj_id in object_ids:
             del bindable_properties[(obj_id, name)]
-            finalizer.detach()
 
 
 def reset() -> None:


evnchn:
Obligated to do so as the Windows guy. After my groceries.

polyipseity:
Seems good. There is a line of code at the very end modifying `bindable_properties`:

https://github.com/zauberzeug/nicegui/blob/293930dc30aeaa46ffacf37507e61af27179e8e8/nicegui/binding.py#L276

But it seems like the line of code is correct after applying this PR.

Before applying this PR, this should not be correct (`copy` is the owning object, not a finalizer), probably because #4122 forgot to change this line of code.

evnchn:
@polyipseity Hmm. 

But if you git blame the line in question, it is introduced in https://github.com/zauberzeug/nicegui/pull/4251, which is before https://github.com/zauberzeug/nicegui/pull/4122

So, you may want to check again the point you are making.

polyipseity:
> @polyipseity Hmm.
> 
> But if you git blame the line in question, it is introduced in #4251, which is before #4122
> 
> So, you may want to check again the point you are making.

Hm, then kind of surprise how #4251 has not caused problems... As far as I can see, `creator` is the owning class, which `creator(...)` definitely does not create a finalizer.

evnchn:
The code is still a bit difficult to read. I'm working empirically for now. 

Just to be sure against `copy` situation, here is a modified test script. 

Modifications: 

* Last 500 `Demo` are a deepcopy of the first 500
* Keep on copying the first `Demo` to be the last `Demo`

Result: 
* No memory leak after the PR fix. Leaks without. 
* `Demo` at 0 (being copied all the time), 400 (being copied once), 800 (copy once), -1 (copying all the time) all have `.number` (which it wouldn't if the copy fails) and value remains uniform. 

```py
from nicegui import ui, binding
import tracemalloc
import copy

# Start tracing memory allocations
tracemalloc.start()

class Demo:
    number = binding.BindableProperty()
    def __init__(self) -> None:
        self.number = 1
    def increment(self) -> None:
        self.number += 1

demos = [Demo() for _ in range(500)]
demos += [copy.deepcopy(demo) for demo in demos]

def increment_all_demos() -> None:
    # make the last Demo a deep-copy of the first one
    demos[-1] = copy.deepcopy(demos[0])
    for demo in demos:
        demo.increment()

ui.timer(0.1, increment_all_demos)

def snapshot_memory_usage() -> None:
    snapshot = tracemalloc.take_snapshot()
    top_stats = snapshot.statistics('lineno')
    print("[ Top 5 memory allocations ]")
    for stat in top_stats[:5]:
        print(stat)
    print(demos[0].number, demos[400].number, demos[800].number, demos[-1].number)

ui.timer(5, snapshot_memory_usage)

ui.run(show=False)
```

rodja:
For verification we have deployed this PR to https://nicegui.io. So far everything looks fine 👍

x2a-packet:
I tested https://github.com/zauberzeug/nicegui/issues/4617#issuecomment-2804047378 and the growth pattern did change radically but still I'm able to observe an unexplainable growth (~264KB) in some cases of the test. It seems to happen most the time after some minutes. As long as I observed the test no additonal changes to the memory appear. In this case the test was done with ui.
Observation: 
image
Besides that I would as well call the change an improvement for https://github.com/zauberzeug/nicegui/issues/4617#issuecomment-2804047378.

I tested for https://github.com/zauberzeug/nicegui/issues/4617#issuecomment-2803073543 and as well the pattern changed radically and improved. But still there is one raise happening I can't explain.
In this case the test was done with ui.
Observation:
image
Beside that I would as well call the change an improvement for this test.


I tested for https://github.com/zauberzeug/nicegui/issues/4617#issue-2993500738 and observed a little growth as well besides the pattern changed. After 3 or 4 increments there was till now no other increase observed. In this case the test was done with ui.
Observation: 
image

For more insight and because I think an answer to the question what is a correct memory behave and what a false is important for me, I updated the example @falkoschindler  provided to visualize a  bit better what changes when.
Surveilance Script Line:
old=; oldt=$(date +%s); while :; do cur="$(ps -p "$(pgrep -f './test.py')" -o rss= 2>/dev/null)"; [ -z "$cur" ] && sleep 1 && continue; newt=$(date +%s); if [ "$cur" != "$old" ]; then diff=0; tdiff=$((newt - oldt)); [ -n "$old" ] && diff=$((cur - old)); printf '%s: %s (%+d) delta=%ds\n' "$(date +'%F %T')" "$cur" "$diff" "$tdiff"; old="$cur"; oldt=$newt; fi; sleep 1; done
`Code:`
```
from nicegui import ui, app,
import gc

@ui.page('/')
def index_page() -> None:
    label = ui.label()
    ui.timer(0.1, lambda: label.set_text('foo' if label.text == 'bar' else 'bar'))

#app.timer(1, gc.collect)

ui.run( )
```
I came to the conclussion that the observed small increase which happens still does not happen if there is no connection to the UI.
I tried to understand the behave by adding app.timer(1, gc.collect) and the increase does not happen even when connected to the UI.

What should be the correct behave?

falkoschindler:
> I tried to understand the behave by adding app.timer(1, gc.collect) and the increase does not happen even when connected to the UI.

Yes, calling `gc.collect` makes the experiment more predictable. Otherwise Python might decide not to run the GC for a while because there is still plenty of free memory.

rodja (approved):

evnchn:
@x2a-packet 

I think we are at least in the right direction, with my test script which toggles bindableproperty rapidly not leaking memory. If you do agree on that. 

---

On the remaining memory growth: 

Consider doing the `tracemalloc` thing that I have done in my previous test scripts, to see who's allocating the 264KB. 

I would like to ask: did anyone visit your website during the testing period? It's important that no one does, since on every HTTP request, we load in a `Client` instance (which is what runs the Python in the background). That takes some memory, and could be the memory growth you have obtained. 

Also, see if you can exacerbate the memory growth to the point that it is obvious. Best if the program freezes or becomes slow. Again, hunting down 264KB amongst 100MB is hard, but hunting down 264MB amongst 100MB is easier.

falkoschindler:
> On the remaining memory growth

@evnchn I suggest to open a new discussion to investigate any remaining memory problems. Now that this PR is merged and one hole is plugged, we can start fresh with a new MRE etc.

Issue #4623, opened 2025-04-16
`ui.markdown` always includes mermaid library, after using mermaid once


evnchn:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

def debug_markdown():
    ui.button("Create plain markdown with fenced code blocks", on_click=lambda: ui.markdown("> Dynamically-created markdown with no extras"))
    ui.button("Create Mer-maid", on_click=lambda: ui.markdown("""
    ```mermaid
    graph LR;
        A --> B;
        A --> C;
    ```
    """, extras=["fenced-code-blocks", "mermaid"]))
    ui.label(str(ui.markdown.libraries))
    if ui.markdown.libraries:
        ui.label('Mer-maid library is in ui.markdown, despite the fact that it is not used!').classes('text-red')
        ui.label('View page source. There will be references to mer-maid in createApp, which there should be none!').classes('text-red')

@ui.page('/')
def no_mermaid_page():
    ui.markdown("> Markdown with no extras")
    debug_markdown()

ui.run(port=9970)
```

### Description

Follow these steps carefully

1. Start the server brand-new
2. Check the page source code. There should be exactly 2 references to mermaid, none in `createApp` section. 
3. Check the page, it should show `[]`, since `ui.markdown.libraries` is empty. 
4. Click "Create plain markdown with fenced code blocks". It works
5. Click "Create Mer-maid". It doesn't work, which it should, since #4620 is still being investigated
6. Reload the page
7. Check the page source code. There are now more references to mermaid, with `"libraries":[{"key":"a022d63bc2efb68f616955feada73028/mermaid.js","name":"mermaid"}]` somewhere in the `createApp` section, indicating `ui.markdown` is including mermaid library for no reason*
8. Check the page, it should show `[Library(key='a022d63bc2efb68f616955feada73028/mermaid.js', name='mermaid', path=WindowsPath('C:/Users/User/Documents/RequireMainProcess/.venv/Lib/site-packages/nicegui/elements/lib/mermaid/mermaid.js'), expose=True)]`, since `ui.markdown.libraries` is empty*
9. Click "Create plain markdown with fenced code blocks". It works
10. Click "Create Mer-maid". It works, which masks issue #4620. 
[...]

PR #2991, opened 2024-04-30
Unify declaration of third-party dependencies

falkoschindler:
Currently there are three ways to include JavaScript libraries when subclassing UI elements: `libraries`, `exposed_libraries` and `extra_libraries`. This PR experiments with using `exposed_libraries` only, which seems to work for all elements. A new "signature pad" example demonstrates how to use NPM to define and install JavaScript dependencies in a NiceGUI project.

Open tasks:

- [x] deprecate `libraries` and `extra_libraries` parameters
- [x] replace `exposed_libraries` with a more intuitive name, like `dependencies`
- [x] demonstrate a more complex NPM example, like vue-signature-pad, which has other dependencies? --> No, there's currently no point in trying to integrate third-party Vue components since we're not able to support Vue 3 components at the moment.
- [x] demonstrate a more complex NPM example, like is-odd, using a bundler (see the new "is even" example)
- [x] add some documentation about custom UI elements with third-party dependencies --> I updated the readme files in the "is_even" and "signature_pad" examples. But where would be a good place in the documentation to add a pointer to these examples?

PR #4625, opened 2025-04-16
Rely on ES importmap to load `mermaid.esm.min.mjs`, never load `mermaid.js`

### LINKED ISSUE #4620: fail to initialize mermaid in markdown component
### First Check

### Example Code

~~~python
from nicegui import ui

container = ui.row()


def update_content():
    with container:
        doc = ui.markdown(
            """
```mermaid
flowchart TD
    A[Start] --> B{Decision}
    B -->|Yes| C[Action 1]
    B -->|No| D[Action 2]
    C --> E[End]
    D --> E
```
""",
            extras=[
                "fenced-code-blocks",
                "mermaid",
            ],
        )

ui.button("Create Mermaid").on("click", update_content)
ui.run()
~~~

### Description

If the markdown component is constructed after button click, it seems some mermaid library initialization fails.

errors in console:
```
nicegui.js:425 TypeError: Cannot read properties of undefined (reading 'default')
    at mermaid.js:133465:55
```

Refresh the page will make it work.

### NiceGUI Version

2.14.1

### Python Version

3.12.8

### Browser

Chrome

### Operating System

macOS

### Additional Context

_No response_

evnchn:
Fixes #4620. Reference: https://github.com/zauberzeug/nicegui/issues/4620#issuecomment-2810142095

### Updated TL-DR: 

In the end, ripped out a bunch of old mermaid JS files, and no need to do any `self.libraries.append`, since we move the responsibility of loading mermaid from `nicegui.js` to `markdown.js`, expecting `this.mermaid = (await import("mermaid")).default;` in `markdown.js` and `import mermaid from "mermaid";` in `mermaid.js` to straight-up work unless you are using an ancient browser without ES module support. Added a test and updated how `npm.json` runs as well. 

---

### TL-DR of the initial commit: 

`Mermaid.exposed_libraries[0]` has the 1st element being `mermaid.js`, which [blows up on `import`](https://github.com/zauberzeug/nicegui/issues/4620#issuecomment-2810107663) since it is a browser bundle, not a ES Module. This PR fixes it by screening through `exposed_libraries` to find the one which is `mermaid.esm.min.js` and suitable for import. 

**NOTE 🚨: Be sure to restart the server between testing, since #4623 makes the markdown element reference the mermaid right in the beginning (so statically, not dynamically load-in), making it work without the PR.** 

EDIT: Test script is at: https://github.com/zauberzeug/nicegui/issues/4620#issuecomment-2810102118

### Before: 

{A672CCF4-9231-4046-BE67-830556DD7EC0}

{38938EB3-5481-4DDF-8E66-E4838E278E08}

Observations: Doesn't work. `nicegui.js` attempts to load in `mermaid.js` which it shouldn't

### After: 

{9698C11A-B6F4-4048-96DB-61A88B944BFF}

{0F545DDF-0927-4D13-B352-C84D650687F5}

Observations: Works like a charm. `nicegui.js` attempts to load in `mermaid.esm.min.mjs` which it should

---

### History of `ui.markdown` Mermaid support: 

`ui.markdown` use to always reference Mermaid, when Mermaid support was initially added: https://github.com/zauberzeug/nicegui/commit/ced4b0f451a05979dbe59df600563bc890c777ae
> `register_component('markdown', __file__, 'markdown.js', ['lib/mermaid.min.js'])`

Then, it was made dynamic, referencing it only when needed: https://github.com/zauberzeug/nicegui/commit/1420a51512587e06ac465a0e0eab9491ccb91128
> `if 'mermaid' in extras:` `self.use_library('mermaid')`

Afterwards, the library inclusion logic got a rework, and we got this for a moment: https://github.com/zauberzeug/nicegui/commit/2671059537538be668606aa4e214b2b86809d66d
> `self.use_library(mermaid_library)`, where `from .mermaid import library as mermaid_library`, and `library = register_library(Path('mermaid', 'mermaid.esm.min.mjs'), expose=True)` in `mermaid.py`

Finally, commit https://github.com/zauberzeug/nicegui/commit/4c5745b1f9c46ef8a40dfa98dd278e8fcc1185dc introduced the `Mermaid.exposed_libraries[0]`, which: 
1. Doesn't seem to work to reference `mermaid.esm.min.js`, instead referencing `mermaid.js` as the first choice, and
2. Is the only instance in the entire codebase where it indexes into a particular item in `exposed_libraries`, where otherwise the sequence does not matter. 

@falkoschindler as you authored the last PR, am I missing something as to why you are indexing right into the first element of `exposed_libraries`?


falkoschindler:
@evnchn Brilliant analysis!
Phew, a PR from two years ago... I guess it was fine back than because there was only one element in `exposed_libraries`. It changed in PR #2991 which probably caused this bug.

I'll think about this fix again tomorrow. Right now I have so many questions... 😄

- Why does `ui.mermaid` need multiple exposed libraries and `ui.markdown` only one of them?
- Could we simply `self.libraries.extend` _all_ of `ui.markdown`'s exposed libraries?
- Why do we add them to `self.libraries` and not `self.exposed_libraries`?

evnchn:
I'm more "make the dynamic import work like the static import" when I think to resolve the issue, so I call it done if I add `mermaid.esm.min.mjs` to library, not `mermaid.js`, not really thinking about why. 

Your questions are good ones, though. But, we'll think about it tomorrow. Plus, I am not familiar with Mermaid internals. 

---

> it was fine back than because there was only one element in `exposed_libraries`

Back then, you should apply an `assert len(Mermaid.exposed_libraries) == 1` in the code, so that the issue becomes obvious. Oh well...

falkoschindler:
I just added a new test case for dynamically adding `ui.markdown` with Mermaid diagrams.

And I cleaned up the list of Mermaid dependencies that are actually used. Apparently we can remove around 100 files and tests are still green. I manually tested static Markdown with Mermaid as well as dynamically creating it.

Regarding my questions:

> Why does ui.mermaid need multiple exposed libraries and ui.markdown only one of them?

This should be addressed by commit a47065dc0a04cc4c828f6ab2517e56d6575c9809.

> Could we simply self.libraries.extend all of ui.markdown's exposed libraries?

**I'm still not sure. Could there be a scenario where we need anything from the chunks/ folder?**

> Why do we add them to self.libraries and not self.exposed_libraries?

I updated it to use `exposed_libraries` for consistency. But it doesn't seem to matter.

@evnchn What do you think? Did I miss anything else?

evnchn:
> I'm still not sure. Could there be a scenario where we need anything from the chunks/ folder?

Doubt it. AFAIK The approach with webpack chunking is to break up code in to several files (so that if your code doesn't use everything in the monolithic file, it becomes faster as there's less download going on)

So if we loaded the monolithic file, it should be the case that we need nothing more. 

If you aren't sure, you can do a code search to see if code in the chunk is in the code of the monolithic somewhere.

evnchn:
Interesting. 

My understanding is: `"mermaid":"/_nicegui/2.15.0/libraries/a022d63bc2efb68f616955feada73028/mermaid.esm.min.mjs"` in importmap already gives the `ui.markdown` plenty enough information such that it is possible to simply `this.mermaid = (await import("mermaid")).default;` in `markdown.js`. Same goes for `import mermaid from "mermaid";` in `mermaid.js`

So all that `self.exposed_libraries.append(...)` was futile from the get-go 🤡

---

Let me quickly try with Pale Moon, the browser which doesn't support ES Module natively, and check if the PR works as well. But, I am quite sure it ought to work.

evnchn:
And you can see: Now markdown.js bears the responsibility of importing `mermaid.esm.min.js`. 

{AAB9AB92-3697-443E-9C57-B934DE59F807}

This would also fix #4623, since we are no longer doing any sort of `self.exposed_libraries.append(...)`.

evnchn:
@falkoschindler wait a moment. 

**🚨Not working on Pale Moon!!!**

{5942460A-BE20-41D6-971B-4E97F55932F9}

Mind if you install the browser to test yourself? 

Though: ES Module support is pretty widespread at this point...

falkoschindler:
> So all that `self.exposed_libraries.append(...)` was futile from the get-go 🤡

Yeah, it might have been necessary back when the whole dependency management was more immature and you had to explicitly define libraries to be imported.

> This would also fix https://github.com/zauberzeug/nicegui/issues/4623

This would be awesome!

> Not working on Pale Moon!!!

Does the main branch work? Or is ES Module support already mandatory for NiceGUI?

evnchn:
> Not working on Pale Moon!!!

The issue seems to be a limitation of the Polyfill: https://github.com/guybedford/es-module-shims?tab=readme-ov-file#polyfill-edge-case-dynamic-import

Main branch also doesn't work for Pale Moon, so we can consider this a separate issue, and work on it if we want (or not, since the fix would be to include invisible elements to load in mermaid)

**Therefore, my recommendation is: Let's merge!**

```py
ui.mermaid("""
graph LR;
    A --> B;
    A --> C;
""").set_visibility(False)  # force load mermaid component
```

evnchn:
@falkoschindler also you may consider re-title and modify the description of this PR to avoid confusion down the line. We are doing _way_ more than "always add mermaid.esm.min.mjs to library, not mermaid.js" at this point.

falkoschindler (approved):

evnchn:
Final note: We should probably not spend time into supporting old browsers which does not support ES modules natively, any more than including es-module-shim. 

We have many dynamic imports: https://github.com/search?q=repo%3Azauberzeug%2Fnicegui+import+language%3AJavaScript&type=code&l=JavaScript

So: If your browser doesn't support ES modules natively, it either works for you (congrats), or it doesn't (then you'll figure it out, whether as to put some placeholder element to force import statically, or use another element).

Issue #4319, opened 2025-02-07
Running native mode from a script.exe ignores settings


J3ronimo:
Hi niceguys,

my native mode window ignores its settings when spawning it through an executable script.

Say we take the example from the website and installed it in a venv:

```python
# main.py
from nicegui import app, ui

app.native.window_args['resizable'] = False
app.native.start_args['debug'] = True
app.native.settings['ALLOW_DOWNLOADS'] = True

ui.label('app running in native mode')
ui.button('enlarge', on_click=lambda: app.native.main_window.resize(1000, 700))

ui.run(native=True, window_size=(400, 300), fullscreen=False)
```

By "settings" I mean these values like `"resizable"`.

Now I made this a "script" where script means the pip/distlib .exe that is created in `.venv\Scripts` when you `pip install` a project that has a `[project.scripts]` section in its pyproject.toml. Such script endpoints specify a callable from a module, which should be imported and called without args on launch of the script. Here this can be done by adding an empty dummy function at the bottom of the file, as long as the rest of the code runs on import, this is sufficient to spawn the native window.

Now though, the `app.native` settings don't get applied, i.e. no debug window shows, and the window is resizable.

Turns out for these settings to get applied, the `multiprocessing.Process` that runs the pywebview window has to import and run these lines of code again. This is the case when running as `__main__`, where mp runs the module again as `__mp_main__` (some sort of special behavior there I guess), but not if it was imported like it is when running a script.exe. Here, only the parent process runs the module, but not the subprocess, i.e. no `__mp_main__` will appear if you print `__name__` at the start of the file. As a result, any settings done to `app.native` are basically irrelevant, since they're not seen by or transferred to the mp subprocess, which is the one that needs them as it spawns the window from its app context.

Another effect is that even if you run this as `__main__`, any modification to `app.native` that you do under a `__main__` guard, say for example taking the value for the debug setting from a `--debug` flag from the command line, will not do anything. You will have to do the exact same argparsing again in an `__mp_main__` guard of the subprocess to setup any window related flags in `app.native` there.

-------------

I finally found a workaround for this, which is

1. Launching through runpy:

```python
# __init__.py
def run_main_from_script():
    runpy.run_module(f"{__name__}.main", run_name="__main__", alter_sys=True)
```

2. Sending all settings for `app.native` through env vars down to the subprocess, i.e.
- run cli / argparse if `__main__`
- somehow encode all relevant data into `os.environ`
[...]

PR #4627, opened 2025-04-17
Pass the NativeConfig object from MainProcess to native subprocess

civilwargeeky:
In order to fix [Issue 4319](https://github.com/zauberzeug/nicegui/issues/4319), I have added a few lines of code to explicitly pass the NativeConfig from the MainProcess through to the native_mode._open_window function.

While the original poster of the issue was discussing it in terms of problems with freezing Nicegui in an executable, the real problem (I believe) is simply that: since the execution for native code is done within a multiprocessing Process, the script is re-run, all dependencies are re-imported, so core.app.native is reset to be empty in the subprocess, and any code under a main guard is not run.

Minimal example:
```python
# This works fine
from nicegui import app, ui

ui.label("Test")

app.native.window_args['text_select'] = True
app.native.window_args['maximized'] = True

ui.run(title='Test', native=True, reload=False)
```

```python
# This will not apply native settings
from nicegui import app, ui

if __name__ == "__main__":
    ui.label("Test")

    app.native.window_args['text_select'] = True
    app.native.window_args['maximized'] = True

    ui.run(title='Test', native=True, reload=False)
```

My proposed patch simply passes the NativeConfig that exists in the main process to the subprocess when it initializes. This should reduce unexpected behavior from setting native settings from within a main guard

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -29,7 +29,7 @@ except ModuleNotFoundError:
 
 def _open_window(
     host: str, port: int, title: str, width: int, height: int, fullscreen: bool, frameless: bool,
-    method_queue: mp.Queue, response_queue: mp.Queue, native_config: NativeConfig
+    method_queue: mp.Queue, response_queue: mp.Queue, native_config: NativeConfig,
 ) -> None:
     while not helpers.is_port_open(host, port):
         time.sleep(0.1)
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -15,6 +15,7 @@ from .. import core, helpers, optional_features
 from ..logging import log
 from ..server import Server
 from . import native
+from .native_config import NativeConfig
 
 try:
     with warnings.catch_warnings():
@@ -28,10 +29,13 @@ except ModuleNotFoundError:
 
 def _open_window(
     host: str, port: int, title: str, width: int, height: int, fullscreen: bool, frameless: bool,
-    method_queue: mp.Queue, response_queue: mp.Queue,
+    method_queue: mp.Queue, response_queue: mp.Queue, native_config: NativeConfig
 ) -> None:
     while not helpers.is_port_open(host, port):
         time.sleep(0.1)
+
+    # Replace fresh, likely empty config with a config supplied to us via MainProcess
+    core.app.native = native_config
 
     window_kwargs = {
         'url': f'http://{host}:{port}',
@@ -111,7 +115,7 @@ def activate(host: str, port: int, title: str, width: int, height: int, fullscre
         sys.exit(1)
 
     mp.freeze_support()
-    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue
+    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue, core.app.native
     process = mp.Process(target=_open_window, args=args, daemon=True)
     process.start()
 


evnchn:
With https://github.com/zauberzeug/nicegui/issues/4319#issuecomment-2815192087, I question the need.

evnchn:
Would vote for documentation solution, because: 

* I don't see why the benefit must be available in future release (think: 2.16.0) via pass `NativeConfig` to subprocess, when it is easy to avoid the issue by handling the main guard better and enjoying the benefit with older release. 
* Slight concerns over unpicklable variables. 

---

On native mode running in multiprocessing process, I think (personally) it is pretty obvious in the source code: 

https://github.com/zauberzeug/nicegui/blob/3df8b2b16eb7ac471cc36bf9498a3cd652e95bb9/nicegui/ui_run.py#L174-L180

https://github.com/zauberzeug/nicegui/blob/3df8b2b16eb7ac471cc36bf9498a3cd652e95bb9/nicegui/native/native_mode.py#L98-L118

Though, with the wrong main guard and using `reload=True`, we show an error. 

```
   if __name__ == "__main__":
to allow for multiprocessing.
```

But there's no way to detect the user's argument parsing code is in the right main guard or not, so yeah we can't catch that even if we wanted to. 

---

So, in conclusion: 

* Wrong main guard with `reload=True`: we can catch it and throw error
* Wrong main guard with `native=True`: we can't catch it and throw error, so short of passing `NativeConfig` which I am doubtful on, we should best educate the library user in the docs.

civilwargeeky:
Fair enough. I can see your concerns.

And I do agree that it is simple enough to see that `native=True` is run from a multiprocessing.Process when you examine the source code of NiceGUI. I just went down a long rabbit hole of thinking that the issue was with `pywebview` misinterpreting my arguments or that I wasn't passing the objects in the correct manner, since I was unfamiliar with `pywebview` and thought the issue must certainly lie there.

As I'm unfamiliar with how pull requests are usually handled on Github, for this pull request, would you rather I
A) Amend this request to just be documentation changes (with a revert commit and a separate documentation commit)
B) Close this request and open a separate one for the documentation change
C) Close this request and you can make the documentation change at your leisure

Thanks!

evnchn:
Conclusion: B > C > A

First of all, I can't edit the pull request. Even then, I would not recommend to revert what we are doing in this PR and then repurpose this PR for the documentation change. Best create another PR for the documentation (rule out choice A, leaving B or C only) 

I'm bad at communications, so I can't help with documentation, unless you want mass confusion. Also I cut my pinky finger so I can easily type on phone only, so can't do docs as that would involve quite a lot of typing on desktop. (choice B is preferred)

J3ronimo:
Hi guys.

As I wrote in OP:

> This is the case when running as `__main__`, where mp runs the module again as `__mp_main__` (some sort of special behavior there I guess), but not if it was imported like it is when running a script.exe. Here, only the parent process runs the module, but not the subprocess, i.e. no `__mp_main__` will appear if you print `__name__` at the start of the file.

Which means the native subprocess in my case does not re-run the module at all! Therefore, it can never see and apply any settings, no matter if under a main guard or not.

These pip launcher scripts (exes on Windows, no freezing here!) import and run a function as defined in pyproject.toml. That import means the root process is not run as `__main__` but as whatever its name is. And that's what causes above situation.

And even if you run as `__main__`, which will now for some reason cause multiprocessing to run the module again as `__mp_main__`, an argparser like in my case would run twice, one in each process, possibly doing different things... 

So i agree with the @civilwargeeky that
- keeping everything behind a `__main__` guard (as is good style anyways) and
- creating and then passing the NativeConfig down to the mp subprocess

is the only clean way to fix this.

evnchn:
@J3ronimo I would need the "pip launcher scripts" you are mentioning to see what's going on. For instance, since you mentioned "[project.scripts] section in its pyproject.toml", it would help if you can provide the `pyproject.toml`. 

Right now, I am just running it as a script like `python.exe test.py --debug` and come to the conclusion in https://github.com/zauberzeug/nicegui/issues/4319#issuecomment-2815192087

J3ronimo:
@evnchn It should suffice to run `python.exe -c "import test"` to achieve the same.

falkoschindler:
Regarding @evnchn's concerns:

> I don't see why the benefit must be available in future release (think: 2.16.0) via pass NativeConfig to subprocess, when it is easy to avoid the issue by handling the main guard better and enjoying the benefit with older release.

It's always good to improve the library so that things are easier, more intuitive and less confusing. Having to set precise main guards is a pain. If we can avoid it, why not?

> Slight concerns over unpicklable variables.

Ok, this could be problematic. Is there something like callable a user might want to add to the config? Can we detect unpicklable objects and show a warning?

evnchn:
> The pip / pyproject.toml stuff is standard PEP stuff which you'll find plenty examples for on the web.

Sorry @J3ronimo but I am an absolute beginner with regards to pip / pyproject.toml 🤡. I really need your pointer to make it work.

civilwargeeky:
@falkoschindler 
> Ok, this could be problematic. Is there something like callable a user might want to add to the config? 

Looking at the [pywebview docs](https://pywebview.flowrl.com/api/#webview-active-window) for create window, it seems most everything should be simple types, but there are some potentially-unpickleable things that could be added.

For `webview.create_window`:
- `server=http.BottleServer` is a class that may be unreachable or unpickleable
- `server_args={}` and `js_api - Expose a python object to the Javascript domain of the current pywebview window.` are arbitrary python objects

For `webview.start`:
- `func`, `args`, and `server_args` are arbitrary python objects
- `menu` is a list of [arbitrary python functions](https://pywebview.flowrl.com/examples/menu.html)

For `webview.settings`:
Everything is just strings and booleans 😄 

___

> Can we detect unpicklable objects and show a warning?

I mean, starting the process would fail with an error, and maybe you could restart with fresh (or original) NativeConfig and a warning?
From some limited testing, it looks like `process.start()` fails with something like `TypeError: cannot pickle '_io.TextIOWrapper' object` for unpickleable objects like files and like `AttributeError: Can't get attribute 'my_func' on <module '__mp_main__' from '...'>` for things which are dynamically defined.

So perhaps we could do something like
```
# Replace fresh, likely empty config with a config supplied to us via MainProcess
original_config = core.app.native
core.app.native = native_config

...

process = mp.Process(target=_open_window, args=args, daemon=True) 
try:
    process.start()
except (TypeError, AssertionError):
    warnings.warn("pywebview arguments were not pickleable and so will be ignored. launching with default settings")
    core.app.native = original_config
    args = host, port, title, width, height, fullscreen, frameless, native.method_queue, native.response_queue, core.app.native
    process = mp.Process(target=_open_window, args=args, daemon=True)
    process.start()
```

However, I do not know if this error handling code would only be triggered for unpickleable objects, nor do I know if this is comprehensive for the errors that could be encountered launching with unpickleable objects.

civilwargeeky:
Hey @falkoschindler, I went ahead and made the documentation changes you requested. I'm unfortunately unfamiliar with Github pull requests, so instead of reverting my edits and making a documentation commit, I thought it would be cleaner to close this PR and make a new one in a clean branch. [Here](https://github.com/zauberzeug/nicegui/pull/4658).

Let me know if there's anything else you would like me to do, and feel free to clean up the wording that I wrote or remove the extra examples.

Issue #4626, opened 2025-04-17
bindable_dataclass chained property bindings trigger forward and reverse directions


sfeltman:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import binding, ui

values = [
    "one",
    "two",
    "three",
]

@binding.bindable_dataclass
class State:
    chain1: str = ""
    chain2: str = ""
    chain3: str = ""
    chain4: str = ""


def chain(prefix, value):
    print(f"{prefix} changed: {value}")
    return value


@ui.page('/')
def index():
    state = State()
    selector = ui.select(values, value=values[0], label="Selector")
    selector.bind_value_to(state, "chain1", forward=lambda x: chain("select", x))
    binding.bind_to(state, "chain1", state, "chain2", forward=lambda x: chain("chain1", x))
    binding.bind_to(state, "chain2", state, "chain3", forward=lambda x: chain("chain2", x))
    binding.bind_to(state, "chain3", state, "chain4", forward=lambda x: chain("chain3", x))


ui.run(show=False)
```

### Description

Start the app, notice the print statements are as expected:

```
select changed: one
chain1 changed: one
chain2 changed: one
chain3 changed: one
```

Now select "two" from the drop-down, notice we have additional output after the expected output in reverse order:

```
select changed: two
chain1 changed: two
chain2 changed: two
chain3 changed: two
chain3 changed: two
chain2 changed: two
chain1 changed: two
```

I do not expect the last three lines of output.

### NiceGUI Version

2.15.0

### Python Version

3.10.15

### Browser

Firefox

### Operating System

macOS

evnchn:
This is too tough for me to find the issue, but I did help to narrow it down. 

@falkoschindler you can pick up, since you seem to know the binding module pretty well (better than me I guess 😅)

---

Takeaways from my test script and modification to binding.py: 

* Issue does not reside with `@binding.bindable_dataclass`, since it can be recreated with `binding.BindableProperty()`
* There are no duplicate bindings in `binding.bindings`. There are exactly 4 entries, as we would expect (a reason why I am using the auto-index page, aka no `@ui.page`)
* Issue seem to reside in the line `_propagate(owner, self.name)`, in a sense it seem to have some recursive behaviour. 

---

[...]

PR #4628, opened 2025-04-17
Introduce a `_propagation_context` to keep track of visited properties

falkoschindler:
This PR tries to solve #4626 with a new context manager `_propagation_context` that keeps track of visited properties using a ContextVar `propagation_visited`. The new implementation better encapsulates the logic around the `visited` set. This way it simplifies other parts of the code as they don't have to deal with this set anymore.

It also fixes the incorrect check
```py
if source_obj_id in visited:
    return
```

@evnchn, @sfeltman Would you like to have a look?

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -6,6 +6,8 @@ import dataclasses
 import time
 import weakref
 from collections import defaultdict
+from contextlib import contextmanager
+from contextvars import ContextVar
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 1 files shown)


evnchn:
Sorry but my computer algorithms knowledge has been a bit rusty, and I am not exactly strong in that regard either. 

Therefore, I can't really comment on the code. I guess if it works it works, which it does. 

Meanwhile, just to be sure, I would like to clarify: 

Basically the binding is a directed graph, and we want to do a search (idk if it is DFS or BFS, must be either one) through the graph starting from the changed value, and update the value of all nodes reached by the search, right? 

And, the issue is that, we forgot to mark the nodes as visited properly, leading to duplicate visits and thus the triggering of the `forward` lambda function, right? 

If you shall agree, then it LGTM.

falkoschindler:
@evnchn Yes, you're correct: It is a depth-first search through a directed graph. The problem was that the set of visited nodes was not correctly passed to recursive calls of the `_propagate` function. There was a recursive `_propagate(..., visited)` call, but the `_set_attribute()` call in the line above triggered a propagation that created a _new_ `visited` set. And if everyone uses its own instance of `visited` nodes, it misses the whole point.

> I guess if it works it works, which it does.

I think so too. At least it is (probably) less broken than before. 🙃

PR #4633, opened 2025-04-18
Introduce element.html_id property

evnchn:
Addresses "Introduce an `element.html_id` property which returns the `id` attribute of the DOM element" in NiceGUI Wishlist. 

https://github.com/orgs/zauberzeug/projects/6/views/1?pane=issue&itemId=82941211

Test script: 

```py
from nicegui import ui

my_label = ui.label('Hello world!')
ui.label(f"ID of my_label: {my_label.id}")
ui.label(f"HTML ID of my_label: {my_label.html_id}")
ui.label(f"HTML ID of my_label (inferred from ID): {'c'+str(my_label.id)}")

ui.run(show=False)
```

Test result: 

{5C9DECA0-8482-4610-B44C-611351EC4A30}

{D688002B-A8F0-443D-B04A-3287AA1D714E}

> The "Hello world!" indeed has id "c4"

---

I'll be honest, not sure what's the use of this wishlist entry. Why bother, when `f'c{element.id}'` is the underlying logic, and it's trivial enough to just do that in your code?

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -531,3 +531,8 @@ class Element(Visibility):
                 result += f'\n {line}'
 
         return result
+
+    @property
+    def html_id(self) -> str:
+        """The ID of the element in the HTML DOM."""
+        return f'c{self.id}'
--- a/nicegui/elements/link.py
+++ b/nicegui/elements/link.py
@@ -27,7 +27,7 @@ class Link(TextElement, component='link.js', default_classes='nicegui-link'):
         if isinstance(target, str):
             self._props['href'] = target
         elif isinstance(target, Element):
-            self._props['href'] = f'#c{target.id}'
+            self._props['href'] = f'#{target.html_id}'
         elif callable(target):
             self._props['href'] = Client.page_routes[target]
         self._props['target'] = '_blank' if new_tab else '_self'
--- a/nicegui/elements/teleport.py
+++ b/nicegui/elements/teleport.py
@@ -14,6 +14,6 @@ class Teleport(Element, component='teleport.js'):
         """
         super().__init__()
         if isinstance(to, Element):
-            to = f'#c{to.id}'
+            to = f'#{to.html_id}'
         self._props['to'] = to
         self._update_method = 'update'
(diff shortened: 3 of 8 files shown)


falkoschindler:
Thanks, @evnchn!

> I'll be honest, not sure what's the use of this wishlist entry. Why bother, when f'c{element.id}' is the underlying logic, and it's trivial enough to just do that in your code?

Good point. We should have documented our intention more clearly.
The problem(?) is that the ID is re-generated in many places throughout the code base. You can search globally for `c{` (match case, match whole word) and get around 12 results in Python, two more for `c${` and `"c"` in nicegui.js, and maybe some more places implicitly adding or removing the "c". And if anyone needs to work with HTML IDs in user code, there is only the fragile and undocumented way to add or remove "c" to the integer element ID. Thus, a public API like `element.html_id` would be nice to have.

But if we introduce such an API, we should use it where possible. One duplication in nicegui.js might be reasonable, but at least that would make two distinct places instead of a dozen.

To check if we spotted all relevant places, we could change `html_id` to something like `f'xyz{id}'` and see which tests fail.

evnchn:
I can agree it's a bit fragile and undocumented to append c to id and call it a day for html_id, but it's not like we will change the prefix to add trouble to ourselves (and break any other hacky scripts which rely on the c prefix, trust me there's many), so actually I'm more fine with the status quo. 

However, explicitly using the newly defined `element.html_id` would help increase clarity in the codebase. Basically just a refactor at this point.

evnchn:
@falkoschindler I genuinely hope I got them all, but to say I am unsure would be an understatement. I am just doing VSCode and searching for `"c"`, `'c'`, `c{`. I feel like you can do a much better job than me with your mypy or something.

falkoschindler (approved):
Looks good to me! 👍🏻

evnchn:
@falkoschindler Got some issue: 

Cannot use `element.html_id` with `getHtmlElement`, which instinctively feels like the right thing to do. 

(`getElement` alro breaks, but that's less expected from naming)

Should we modify the functions to make them accept `"c5"` in addition to `5`? I see no harm.

PR #4586, opened 2025-04-07
Notify CodeMirror gracefully to avoid out-of-sync between CodeMirror and NiceGUI internal states

evnchn:
fix #3337 

This is because, when you bind the value, eventually `set_value` in `ValueElement` is called, which does this:

https://github.com/zauberzeug/nicegui/blob/b9c9e8492750153d4b519b24da57b02751ff225f/nicegui/elements/mixins/value_element.py#L103-L108

That's quite a direct approach, and the CodeMirror can't accept that (or it needs to poll the value in the prop upon every update). 

Instead, CodeMirror offers another method for setting value, which is `setEditorValue`

{F70AA4F6-3E5E-4381-8E83-9D5767FCA077}

We just need to call it, and CodeMirror will be "made aware" of a change in editor value, so the following ChangeSet can be applied properly.

PR #4635, opened 2025-04-18
Use `element._update_method` for `ui.codemirror`

### LINKED ISSUE #3337: `ui.codemirror` not updating when value changes programmatically

falkoschindler:
This PR is an alternative to #4586 to fix #3337. Watching the `value` in Vue didn't work, which is why the frontend didn't update when changing the editor value programmatically. This PR leverages the new `_update_method` to tell the client to always call `setEditorValue()` after receiving an update.

### DIFF
--- a/nicegui/elements/codemirror.js
+++ b/nicegui/elements/codemirror.js
@@ -87,12 +87,15 @@ export default {
         effects: this.themeConfig.reconfigure([new_theme]),
       });
     },
-    setEditorValue() {
+    setEditorValueFromProps() {
+      this.setEditorValue(this.value);
+    },
+    setEditorValue(value) {
       if (!this.editor) return;
-      if (this.editor.state.doc.toString() === this.value) return;
+      if (this.editor.state.doc.toString() === value) return;
 
       this.emitting = false;
-      this.editor.dispatch({ changes: { from: 0, to: this.editor.state.doc.length, insert: this.value } });
+      this.editor.dispatch({ changes: { from: 0, to: this.editor.state.doc.length, insert: value } });
       this.emitting = true;
     },
     setDisabled(disabled) {
--- a/nicegui/elements/codemirror.js
+++ b/nicegui/elements/codemirror.js
@@ -13,9 +13,6 @@ export default {
     highlightWhitespace: Boolean,
   },
   watch: {
-    value(newValue) {
-      this.setEditorValue(newValue);
-    },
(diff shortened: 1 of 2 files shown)


evnchn:
@falkoschindler Agree with the approach but disagree with modifying `setEditorValue`, so please don't merge right away. 

Because this would **break existing code which does `getElement(...).setEditorValue(...)`** (should you consider `setEditorValue` as a public API, which I do)

Rather, perhaps implement `setEditorValueFromProps()`, so that we can do `self._update_method = 'setEditorValueFromProps'` while keeping existing compatibility.

evnchn:
If I am not mistaken: 

```js
...
setEditorValueFromProps() {
    this.setEditorValue(this.value);
}
...
```

evnchn:
Tests are good. LGTM.

PR #4640, opened 2025-04-20
docs(upload_documentation.py): update documentation on ui.upload to use new parameter name

yudelevi:
Following the method described in the docs [here](https://nicegui.io/documentation/upload#uploading_large_files) no longer works. 

I did some digging and it's because of changes to starlette:
* `MultiPartParser`: Rename `max_file_size` to `spool_max_size` [#2780](https://github.com/encode/starlette/pull/2780).

### DIFF
--- a/website/documentation/content/upload_documentation.py
+++ b/website/documentation/content/upload_documentation.py
@@ -38,7 +38,7 @@ def show_file_content() -> None:
 
 @doc.demo('Uploading large files', '''
     Large file uploads may encounter issues due to the default file size parameter set within the underlying Starlette library.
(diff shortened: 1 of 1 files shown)


evnchn:
LGTM

But I will have to say, breaking changes in their API from starlette on version 0.46.0 (what looks to be a minor release)? That's not very great...

In NiceGUI, the old parameter would stick to the next version, with `DeprecationWarning` before then. 

We'll have to be more careful with the upstream when we bump dependencies.
 https://github.com/encode/starlette/releases/tag/0.46.0

falkoschindler (approved):
Thanks for the pull request, @yudelevi!
What a nasty breaking change in Starlette! They must have been aware of it:

>  I tried to add a deprecation warning, but the effort didn't seem to pay off.

https://github.com/encode/starlette/pull/2780#issuecomment-2676178310

It's good to keep our documentation up to date though. 👍🏻

PR #4641, opened 2025-04-21
Properly await coroutines registered with app.on_shutdown

### LINKED ISSUE #4312: app reset fails due to improper I/O
### Description

My `nicegui` tests run with pytest leave `.nfs` files in the `.nicegui` folder and cause my tests to fail.

I run my nicegui tests on a system that produces [.nfs](https://serverfault.com/questions/201294/nfsxxxx-files-appearing-what-are-those) files when a file is deleted before I/O is properly terminated. Annoyingly, these files cannot be deleted by the user and only get cleaned up after the process finishes execution. The only file in the cache is `storage-user-*.json` so I assume that is the file handle which gets closed improperly.

My workaround is to simply move the `.nicegui` cache folder to trick the app into behaving correctly.

```python
@pytest.fixture
def user(user: User) -> User:
    yield user
    try:
        app.reset()
    except OSError:
        # This is necessary since once of the file handles opened
        # by nicegui isn't closed properly and leaves a .nfs file
        # We move the cache folder and then delete later.
        if list(app.storage.path.glob(".nfs*")):
            tmp_path = Path(".to_delete", '.nicegui')
            tmp_path.parent.mkdir(exist_ok=True)
            if tmp_path.exists():
                shutil.rmtree(tmp_path)

            shutil.move(app.storage.path, tmp_path)
        app.reset()
```

### LINKED ISSUE #4592: pytest always RuntimeWarning coroutine 'FilePersistentDict.backup.<locals>.backup' was never awaited
### First Check

### Example Code

```python
Appears in https://github.com/zauberzeug/nicegui/actions/workflows/test.yml often

Passing tests: 
https://github.com/zauberzeug/nicegui/actions/runs/14312243561/job/40109877142
https://github.com/zauberzeug/nicegui/actions/runs/12870993631/job/35883341208 (Jan 20)

Failing tests: 
https://github.com/zauberzeug/nicegui/actions/runs/13815787631/job/38666492630
```

### Description

I always see `RuntimeWarning: coroutine 'FilePersistentDict.backup.<locals>.backup' was never awaited` in the pytest logs in the GitHub actions. 

Personally, I have some experiences where the storage defined via `FilePersistentDict` (`app.storage.general` and `app.storage.user` most affected) would be straight-up corrupted (being empty on next read) when I shutdown the server while lots of clients are accessing the website. This may be related, but recreating this issue is hard...

### NiceGUI Version

2.13.0, affects main branch on GitHub dating back at least Jan20

### Python Version

3.8-3.13 all affected

### Browser

Other

### Operating System

Linux

#### Preliminary cause analysis: 

https://github.com/zauberzeug/nicegui/blob/4e538bb9ba213054ea0991caacfa5df9a28c98bf/nicegui/persistence/file_persistent_dict.py#L14-L54

on_change, it runs `backup` method, which does the actual backing up via `background_tasks.create_lazy(backup(), name=self.filepath.stem)`. 

When the test is shut down, there is no cleanup mechanism to ensure all tasks created by `background_tasks.create_lazy` are completed, and so `coroutine 'FilePersistentDict.backup.<locals>.backup' was never awaited`

---

#### Preliminary proposed fix: 

* Provide some mechanism to find the task and await the tasks for their completion in `background_tasks` 
* `app.on_shutdown` await for all backup task for `FilePersistentDict` to finish.

rodja:
This PR started with the intent to fix #4592 which reported unawaited coroutines when running pytests. It turns out, the warning lead down a rabbit hole of problems. Here is a list of the major findings and fixes:

- the coroutines registered with `app.on_shutdown` have not been awaited properly
- `background_tasks` where not canceled on shutdown
- `outbox.loop` did not exit when receiving a cancel command
- handle cancellation in internal house-keeping loops (pruning, binding, ...)
- before Python 3.12 `asyncio.wait_for` misses cancel commands in certain conditions and must be fixed with [`wait-for2`](https://pypi.org/project/wait-for2/)
- ensure that all storage backups are written before exiting
- introduce `@background_tasks.await_on_shutdown` annotation to mark coroutines as "not to cancel on shutdown" (and thereby also fixes #4312)
- ~~clean up multiprocessing to not get "leaked semaphore" warnings (fix was described in https://github.com/zauberzeug/nicegui/issues/4131#issuecomment-2705100273)~~

Also this PR makes some minor improvements:

- disable pytest warning from upstream packages which we can do nothing about
- fix UTC warnings
- add names to background_tasks where ever possible
- only stop screen tests if console output contains "ERROR" (same as #4608 did for the user tests)
- first steps to add some `background_tasks` documentation

### DIFF
--- a/nicegui/background_tasks.py
+++ b/nicegui/background_tasks.py
@@ -84,8 +84,9 @@ async def _cancel_all(tasks: set[asyncio.Task]) -> None:
         try:
             await asyncio.wait(tasks, timeout=2.0)
         except asyncio.TimeoutError:
-            for task in [t for t in tasks if not t.done()]:
-                log.error(f'Task {task.get_name()} could not be aborted within timeout')
+            log.error('Could not cancel %s tasks within timeout: %s',
+                      len(tasks),
+                      ', '.join([t.get_name() for t in tasks if not t.done()]))
 
     running_tasks.clear()
     lazy_tasks_running.clear()
--- a/nicegui/background_tasks.py
+++ b/nicegui/background_tasks.py
@@ -89,5 +89,7 @@ async def _cancel_all(tasks: set[asyncio.Task]) -> None:
                 if not task.done():
                     task_name = task.get_name() if hasattr(task, 'get_name') else 'unknown'
                     print(f'Task {task_name} could not be aborted within timeout')
+
+    await asyncio.sleep(0)  # ensure the loop can cancel the tasks before it stops
     running_tasks.clear()
     lazy_tasks_running.clear()
--- a/nicegui/background_tasks.py
+++ b/nicegui/background_tasks.py
@@ -56,10 +56,19 @@ def _handle_task_result(task: asyncio.Task) -> None:
         core.app.handle_exception(e)
 
 
-def on_shutdown() -> None:
+async def on_shutdown() -> None:
+    """Cancel all running tasks and coroutines on shutdown."""
     for task in running_tasks:
-        task.cancel()
+        await _cancel_task(task)
     for task in lazy_tasks_running.values():
-        task.cancel()
+        await _cancel_task(task)
     for coroutine in lazy_coroutines_waiting.values():
         coroutine.close()
+
+
+async def _cancel_task(task: asyncio.Task) -> None:
+    task.cancel()
+    try:
+        await task  # NOTE we must run the task to completion
+    except asyncio.CancelledError:
+        pass
--- a/nicegui/background_tasks.py
+++ b/nicegui/background_tasks.py
@@ -68,9 +68,9 @@ def _handle_task_result(task: asyncio.Task) -> None:
 
 async def on_shutdown() -> None:
     """Cancel all running tasks and coroutines on shutdown."""
-    for task in running_tasks:
+    for task in list(running_tasks):
         await _cancel_task(task)
-    for task in lazy_tasks_running.values():
+    for task in list(lazy_tasks_running.values()):
         await _cancel_task(task)
     for coroutine in lazy_coroutines_waiting.values():
         coroutine.close()
--- a/nicegui/background_tasks.py
+++ b/nicegui/background_tasks.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+import functools
 from typing import Any, Awaitable, Callable, Coroutine, Dict, Set
 
 from . import core, helpers
@@ -51,8 +52,13 @@ def create_lazy(coroutine: Awaitable, *, name: str) -> None:
 
 def await_on_shutdown(fn: Callable) -> Callable:
     """Tag a coroutine function so tasks created from it won't be cancelled during shutdown."""
-    functions_awaited_on_shutdown.add(fn)
-    return fn
+    @functools.wraps(fn)
+    async def wrapper(*args, **kwargs):
+        coro = fn(*args, **kwargs)
+        task = create(coro)
+        task._await_on_shutdown = True  # type: ignore[attr-defined]  # pylint: disable=protected-access
+        return task
+    return wrapper
 
 
 def _ensure_coroutine(awaitable: Awaitable[Any]) -> Coroutine[Any, Any, Any]:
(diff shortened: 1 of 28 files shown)


rodja:
Ahh so close but still so far away... now I get "leaked semaphore" warnings (see #4131) whenever I close an app. Is anyone able to help me with this?

rodja:
Thanks to @TomFreudenberg I could fix the leaked semaphore warnings in c516d03. Now, this PR also fixes #4131.

falkoschindler (changes requested):
Ok, we're almost there. But right now we think collecting callables in `functions_awaited_on_shutdown` causes a potential memory leak, e.g. if the user decorates a function which is defined within a page function. So we have to either prune efficiently, work with weak references or find a different way to "mark" functions as "await on shutdown".

nicegui/background_tasks.py
```diff
@@ -33,24 +36,69 @@ def create_lazy(coroutine: Awaitable, *, name: str) -> None:
...
     except Exception as e:
         core.app.handle_exception(e)
+
+
+async def teardown() -> None:
+    """Cancel all running tasks and coroutines on shutdown. (For internal use only.)"""
+    while running_tasks or lazy_tasks_running:
+        tasks = set(running_tasks) | set(lazy_tasks_running.values())
+        for task in tasks:
+            if not task.done() and not task.cancelled() and not _should_await_on_shutdown(task):
```
falkoschindler, line 83:
```suggestion
            if not task.done() and not task.cancelled() and task.get_coro() not in functions_awaited_on_shutdown:
```
Is this possible?

nicegui/background_tasks.py
```diff
@@ -33,24 +36,69 @@ def create_lazy(coroutine: Awaitable, *, name: str) -> None:
...
     except asyncio.CancelledError:
         pass
     except Exception as e:
         core.app.handle_exception(e)
+
+
+async def teardown() -> None:
+    """Cancel all running tasks and coroutines on shutdown. (For internal use only.)"""
+    while running_tasks or lazy_tasks_running:
+        tasks = set(running_tasks) | set(lazy_tasks_running.values())
```
falkoschindler, line 81:
```suggestion
        tasks = running_tasks | set(lazy_tasks_running.values())
```
`running_tasks` is already a set.

falkoschindler (commented):
I wonder if we really need `_should_await_on_shutdown`. See my corresponding comment below.

Apart from that I tried to use the new decorator with the following snippet:
```py
async def test1():
    print('test1 start')
    await asyncio.sleep(3)
    print('test1 end')


@background_tasks.await_on_shutdown
async def test2():
    print('test2 start')
    await asyncio.sleep(3)
    print('test2 end')

ui.button('Test1', on_click=lambda: background_tasks.create(test1()))
ui.button('Test2', on_click=lambda: background_tasks.create(test2()))
ui.button('Shutdown', on_click=app.shutdown)
```

But when quickly clicking "Test1"-"Test2"-"Shutdown", both tasks seem to be cancelled before the second print statement is executed. Am I doing anything wrong?

rodja:
Oh wow. The click_handler creates another background task which does not have the annotation, on shutdown the annotated one gets preserved but because the outer background task from the click handler is canceled, the inner one still receives a cancellation. I added two user tests to demonstrate the problem.

rodja:
I could fix the nested-background-tasks problem by making sure "tasks" are not awaited in the click handler:

https://github.com/zauberzeug/nicegui/blob/c5d989428067bb5f8804abaa3cd5bf505dc6af35/nicegui/events.py#L440-L441

`asyncio.Tasks` are already running in the background so it does not make sense to run them in another background task.

nicegui/background_tasks.py
rodja: No, that does not work because `functions_awaited_on_shutdown` contains function objects (Callables), not coroutine objects.

rodja:
`client.safe_invoke` used by `app.on_startup`, `app.on_shutdown`, ... also awaited tasks and hence started background tasks in background tasks. I think now all is covered. @falkoschindler can you have a look at it again?

falkoschindler (approved):

PR #4651, opened 2025-04-23
Add Star History to README.md

evnchn:
It just looks beautiful so I want to add it 😊 If not we can decide on another provider or drop it.


rodja:
Thanks for the PR. We are happy for every star on GitHub. But stars are not a good indicator of quality; they shouldn’t be overrated.

A history chart adds another external call, can stall on firewalls, and risks highlighting plateaus more than progress. A simple shields-badge or a link to a metrics page keeps the README lean while still satisfying curiosity.
Let’s skip the chart here and focus the front page on code, docs, and real-world usage.

PR #4656, opened 2025-04-24
Add Sortablejs as element (dragable objects!)

EmberLightVFX:
This PR adds Sortable to NiceGUI! ([](https://sortablejs.github.io/Sortable/))
I used the original sortable.js and not any vue adaptations as those were a bit old and were made out of multiple files so I have no idea how to implement them.

Any way, I think I have gotten everything to work correctly. I copied Sortable's examples into the NiceGUI sortable example so you can try out all the different options you can use.

Currently the example contains a bunch of custom css. It would be best to add them into the sortable.py but I would like to know what the best way would be to implement it?

### DIFF
--- a/examples/sortable/main.py
+++ b/examples/sortable/main.py
@@ -1,0 +1,351 @@
+from random import randint
+
+from nicegui import ui
+
+
+def on_end(e):
+    ui.notify(f'Item moved from {e.args["oldIndex"]} to {e.args["newIndex"]}')
+
+
+def on_add(e):
+    ui.notify('Item added from another list')
+
+
+def on_sort(e):
+    ui.notify('List order changed')
+
+
+def on_move(e):
+    ui.notify('Item being moved')
+
+
+def on_clone_add(e):
+    ui.notify(f'Cloned item added to the target list at index {e.args["newIndex"]}')
+
+
+def on_filter(e):
+    ui.notify(f'Filtered item was clicked: {e.args["item"]}')
+
+
+def on_select(e):
+    ui.notify(f'Item selected: {e.args["item"]}')
+
+
+def on_deselect(e):
+    ui.notify(f'Item deselected: {e.args["item"]}')
+
+
+def on_spill(e):
+    ui.notify(f'Item spilled: {e.args["item"]}')
+
+
+# Example 1: Simple List
+ui.label("")
+with ui.card():
+    ui.label('Example 1: Simple List').classes('text-h5')
+    ui.label('Basic sortable list with animation')
+
+    with ui.sortable(on_end=on_end, animation=150, ghost_class='blue-background-class') as simple_sortable:
+        for i in range(1, 7):
+            with ui.card().classes('p-4 mb-2 list-group-item'):
+                ui.label(f'Item {i}')
+
+# Control Panel
+    ui.label('Control Panel').classes('text-h5')
+
+    with ui.row():
+        ui.button('Enable Simple List', on_click=lambda: simple_sortable.enable())
+        ui.button('Disable Simple List', on_click=lambda: simple_sortable.disable())
+        ui.button('Reverse Order', on_click=lambda: simple_sortable.sort(
+            list(reversed(simple_sortable.default_slot.children)), True))
+
+
+# Example 2: Shared Lists
+with ui.card():
+    ui.label('Example 2: Shared Lists').classes('text-h5')
+    ui.label('You can drag between these two lists:')
+
+    with ui.row():
+        with ui.card().classes('w-64'):
+            ui.label('List 1').classes('text-h6')
+            with ui.sortable(group='shared', animation=150, on_add=on_add, on_sort=on_sort) as list1:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item'):
+                        ui.label(f'Item {i}')
+
+        with ui.card().classes('w-64 ml-4'):
+            ui.label('List 2').classes('text-h6')
+            with ui.sortable(group='shared', animation=150, on_add=on_add, on_sort=on_sort) as list2:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item tinted'):
+                        ui.label(f'Item {i}')
+
+# Example 3: Cloning
+with ui.card():
+    ui.label('Example 3: Cloning').classes('text-h5')
+    ui.label('Try dragging from one list to another. The item you drag will be cloned and the clone will stay in the original list.')
+
+    with ui.row():
+        with ui.card().classes('w-64'):
+            ui.label('List 1 (Clone Source)').classes('text-h6')
+            with ui.sortable(group={'name': 'clone-example', 'pull': 'clone'},
+                             animation=150) as clone_list1:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item'):
+                        ui.label(f'Item {i}')
+
+        with ui.card().classes('w-64 ml-4'):
+            ui.label('List 2 (Clone Source)').classes('text-h6')
+            with ui.sortable(group={'name': 'clone-example', 'pull': 'clone'},
+                             animation=150) as clone_list2:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item tinted'):
+                        ui.label(f'Item {i}')
+
+# Example 4: Disabling Sorting
+with ui.card():
+    ui.label('Example 4: Disabling Sorting').classes('text-h5')
+    ui.label('Try sorting the list on the left. It is not possible because it has its "sort" option set to false. However, you can still drag from the list on the left to the list on the right.')
+
+    with ui.row():
+        with ui.card().classes('w-64'):
+            ui.label('Non-Sortable List (Pull Only)').classes('text-h6')
+            with ui.sortable(group={'name': 'shared-disabled', 'pull': 'clone', 'put': False},
+                             animation=150, sort=False) as disabled_sort_list1:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item'):
+                        ui.label(f'Item {i}')
+
+        with ui.card().classes('w-64 ml-4'):
+            ui.label('Sortable List').classes('text-h6')
+            with ui.sortable(group='shared-disabled', animation=150) as disabled_sort_list2:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item tinted'):
+                        ui.label(f'Item {i}')
+
+# Example 5: Handle Example
+with ui.card():
+    ui.label('Example 5: Handle').classes('text-h5')
+    ui.label('Dragging is only possible using the handle')
+
+    with ui.sortable(handle='.handle', animation=150, on_move=on_move) as handle_sortable:
+        for i in range(1, 7):
+            with ui.card().classes('p-2 mb-1 list-group-item'):
+                with ui.row().classes('items-center w-full'):
+                    ui.icon('drag_handle').classes("handle").style("cursor: move;")
+                    ui.label(f'Item {i}')
+
+# Example 6: Filter Example
+with ui.card():
+    ui.label('Example 6: Filter').classes('text-h5')
+    ui.label('Try dragging the item with a red background. It cannot be done, because that item is filtered out using the "filter" option.')
+
+    with ui.sortable(filter='.filtered', animation=150, on_filter=on_filter) as filter_sortable:
+        for i in range(1, 7):
+            if i == 4:
+                with ui.card().classes('p-2 mb-1 list-group-item filtered bg-red-200'):
+                    ui.label('Filtered').classes('text-red-800')
+            else:
+                with ui.card().classes('p-2 mb-1 list-group-item'):
+                    ui.label(f'Item {i}')
+
+# Example 7: MultiDrag Plugin
+with ui.card():
+    ui.label('Example 7: MultiDrag Plugin').classes('text-h5')
+    ui.label('Hold down Ctrl/Cmd key while selecting items to select multiple. Then drag them as a group.')
+
+    with ui.sortable(
+        animation=150,
+        multi_drag=True,
+        multi_drag_key='ctrl',
+        selected_class='multi-selected',
+        on_select=on_select,
+        on_deselect=on_deselect
+    ) as multi_drag_sortable:
+        for i in range(1, 7):
+            with ui.card().classes('p-2 mb-1 list-group-item'):
+                ui.label(f'Item {i} (hold Ctrl/Cmd to select multiple)')
+
+# Example 8: Swap Plugin
+with ui.card():
+    ui.label('Example 8: Swap Plugin').classes('text-h5')
+    ui.label('Items will swap places instead of sorting when dragged over each other.')
+
+    with ui.sortable(
+        animation=150,
+        swap=True,
+        swap_class='swap-highlight'
+    ) as swap_sortable:
+        for i in range(1, 7):
+            with ui.card().classes('p-2 mb-1 list-group-item'):
+                ui.label(f'Item {i} (will swap instead of sort)')
+
+# Example 9: OnSpill Plugins - Remove and Revert
+with ui.card():
+    ui.label('Example 9: OnSpill Plugins').classes('text-h5')
+    with ui.row():
+        with ui.card().classes('w-64 mr-4'):
+            ui.label('RemoveOnSpill Plugin').classes('text-h5')
+            ui.label('Drag an item outside the list to remove it.')
+
+            with ui.sortable(
+                animation=150,
+                remove_on_spill=True,
+                on_spill=on_spill
+            ) as remove_spill_sortable:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item bg-red-100'):
+                        ui.label(f'Drag outside to remove {i}')
+
+        with ui.card().classes('w-64'):
+            ui.label('RevertOnSpill Plugin').classes('text-h5')
+            ui.label('Drag an item outside the list and it will revert to original position.')
+
+            with ui.sortable(
+                animation=150,
+                revert_on_spill=True,
+                on_spill=on_spill
+            ) as revert_spill_sortable:
+                for i in range(1, 7):
+                    with ui.card().classes('p-2 mb-1 list-group-item bg-green-100'):
+                        ui.label(f'Drag outside to revert {i}')
+
+# Example 10: AutoScroll Plugin
+with ui.card():
+    ui.label('Example 10: AutoScroll Plugin').classes('text-h5')
+    ui.label('This example demonstrates auto-scrolling when dragging near the edges of the container.')
+
+    with ui.element('div').style('max-height: 300px; overflow-y: auto; padding: 10px;'):
+        with ui.sortable(
+            animation=150,
+            auto_scroll=True,
+            scroll_sensitivity=20,
+            scroll_speed=15
+        ) as auto_scroll_sortable:
+            for i in range(1, 15):
+                with ui.card().classes('p-2 mb-1 list-group-item'):
+                    ui.label(f'Item {i} (drag up/down to auto-scroll)')
+
+
+# Example 11: Add/Remove Items in Sortable List
+def add_deletable_object(label: str):
+    with ui.card().classes('p-2 mb-1 list-group-item'):
+        with ui.row().classes('items-center w-full'):
+            ui.label(label).classes('flex-grow')
+            ui.button(icon='delete', color='red',
+                      on_click=lambda e: e.sender.parent_slot.parent.parent_slot.parent.delete()).classes('min-w-0 w-8 h-8')
+
+
+def add_new_item(string: str):
+    with dynamic_sortable:
+        add_deletable_object(string)
+    ui.notify('Item added to the list')
+
+
+with ui.card():
+    ui.label('Example 11: Add/Remove Items').classes('text-h5')
+    ui.label('This example shows how to dynamically add and remove items from a sortable list.')
+
+    # Form for adding new items
+    with ui.row().classes('w-full items-center mb-4'):
+        new_item_input = ui.input('Enter new item text').classes('flex-grow mr-2')
+        ui.button('Add Item', on_click=lambda: add_new_item(new_item_input.value)).classes('bg-green-500')
+        new_item_input.value = ''  # Clear the input
+
+    with ui.sortable(animation=150, on_end=on_end) as dynamic_sortable:
+        # Pre-populate with some items
+        for i in range(1, 5):
+            add_deletable_object(f'Sortable Item {i} - Drag to reorder')
+
+    # Action buttons
+    with ui.row().classes('mt-4'):
+        ui.button('Delete All', color='red', on_click=lambda: dynamic_sortable.clear()).classes('mr-2')
+        ui.button('Add 3 Items', color='green', on_click=lambda: [
+                  add_new_item(str(randint(1, 100))) for _ in range(3)]).classes('mr-2')
+
+        async def show_current_order():
+            items_text = [item.default_slot.children[0].default_slot.children[0].text
+                          for item in dynamic_sortable.default_slot.children]
+            ui.notify(f'Current order: {items_text}')
+
+        ui.button('Show Current Order', on_click=show_current_order)
+
+
+# Add CSS for styling
+ui.add_head_html("""
+<style>
+.nicegui-sortable {
+    padding: 8px;
+    border-radius: 4px;
+    background-color: #f5f5f5;
+    min-height: 40px;
+}
+
+.sortable-ghost {
+    opacity: 0.4;
+    background-color: #c8ebfb;
+}
+
+.sortable-chosen {
+    background-color: #e8f4fd;
+}
+
+.sortable-drag {
+    opacity: 0.8;
+    box-shadow: 0 5px 10px rgba(0, 0, 0, 0.15);
+}
+
+.blue-background-class {
+    background-color: #c8ebfb;
+}
+
+.tinted {
+    background-color: #f8d7da;
+}
+
+.list-group-item {
+    position: relative;
+    display: block;
+    margin-bottom: -1px;
+    background-color: #fff;
+    border: 1px solid rgba(0,0,0,.125);
+}
+
+.handle {
+    cursor: move;
+}
+
+.multi-selected {
+    background-color: #c8e6fa !important;
+    border: 1px solid #0d6efd !important;
+}
+
+.swap-highlight {
+    background-color: #ffeeba !important;
+    border: 1px solid #ffc107 !important;
+}
+
+.grid-container {
+    width: 100%;
+    padding: 10px;
+}
+
+.grid-square {
+    display: inline-block;
+    width: 80px;
+    height: 80px;
+    margin: 5px;
+    text-align: center;
+    line-height: 80px;
+    background-color: #fff;
+    border: 1px solid #ddd;
+}
+
+.filtered {
+    background-color: #f8d7da !important;
+    cursor: not-allowed;
+}
+</style>
+""")
+
+ui.run(title="SortableJS Complete Plugin Examples", native=True)
--- a/nicegui/elements/lib/sortable/sortable.complete.esm.js
+++ b/nicegui/elements/lib/sortable/sortable.complete.esm.js
@@ -1,0 +1,3380 @@
+/**!
+ * Sortable 1.15.6
+ * @author	RubaXa   <trash@rubaxa.org>
+ * @author	owenm    <owen23355@gmail.com>
+ * @license MIT
+ */
+function ownKeys(object, enumerableOnly) {
+  var keys = Object.keys(object);
+  if (Object.getOwnPropertySymbols) {
+    var symbols = Object.getOwnPropertySymbols(object);
+    if (enumerableOnly) {
+      symbols = symbols.filter(function (sym) {
+        return Object.getOwnPropertyDescriptor(object, sym).enumerable;
+      });
+    }
+    keys.push.apply(keys, symbols);
+  }
+  return keys;
+}
+function _objectSpread2(target) {
+  for (var i = 1; i < arguments.length; i++) {
+    var source = arguments[i] != null ? arguments[i] : {};
+    if (i % 2) {
+      ownKeys(Object(source), true).forEach(function (key) {
+        _defineProperty(target, key, source[key]);
+      });
+    } else if (Object.getOwnPropertyDescriptors) {
+      Object.defineProperties(target, Object.getOwnPropertyDescriptors(source));
+    } else {
+      ownKeys(Object(source)).forEach(function (key) {
+        Object.defineProperty(target, key, Object.getOwnPropertyDescriptor(source, key));
+      });
+    }
+  }
+  return target;
+}
+function _typeof(obj) {
+  "@babel/helpers - typeof";
+
+  if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") {
+    _typeof = function (obj) {
+      return typeof obj;
+    };
+  } else {
+    _typeof = function (obj) {
+      return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj;
+    };
+  }
+  return _typeof(obj);
+}
+function _defineProperty(obj, key, value) {
+  if (key in obj) {
+    Object.defineProperty(obj, key, {
+      value: value,
+      enumerable: true,
+      configurable: true,
+      writable: true
+    });
+  } else {
+    obj[key] = value;
+  }
+  return obj;
+}
+function _extends() {
+  _extends = Object.assign || function (target) {
+    for (var i = 1; i < arguments.length; i++) {
+      var source = arguments[i];
+      for (var key in source) {
+        if (Object.prototype.hasOwnProperty.call(source, key)) {
+          target[key] = source[key];
+        }
+      }
+    }
+    return target;
+  };
+  return _extends.apply(this, arguments);
+}
+function _objectWithoutPropertiesLoose(source, excluded) {
+  if (source == null) return {};
+  var target = {};
+  var sourceKeys = Object.keys(source);
+  var key, i;
+  for (i = 0; i < sourceKeys.length; i++) {
+    key = sourceKeys[i];
+    if (excluded.indexOf(key) >= 0) continue;
+    target[key] = source[key];
+  }
+  return target;
+}
+function _objectWithoutProperties(source, excluded) {
+  if (source == null) return {};
+  var target = _objectWithoutPropertiesLoose(source, excluded);
+  var key, i;
+  if (Object.getOwnPropertySymbols) {
+    var sourceSymbolKeys = Object.getOwnPropertySymbols(source);
+    for (i = 0; i < sourceSymbolKeys.length; i++) {
+      key = sourceSymbolKeys[i];
+      if (excluded.indexOf(key) >= 0) continue;
+      if (!Object.prototype.propertyIsEnumerable.call(source, key)) continue;
+      target[key] = source[key];
+    }
+  }
+  return target;
+}
+function _toConsumableArray(arr) {
+  return _arrayWithoutHoles(arr) || _iterableToArray(arr) || _unsupportedIterableToArray(arr) || _nonIterableSpread();
+}
+function _arrayWithoutHoles(arr) {
+  if (Array.isArray(arr)) return _arrayLikeToArray(arr);
+}
+function _iterableToArray(iter) {
+  if (typeof Symbol !== "undefined" && iter[Symbol.iterator] != null || iter["@@iterator"] != null) return Array.from(iter);
+}
+function _unsupportedIterableToArray(o, minLen) {
+  if (!o) return;
+  if (typeof o === "string") return _arrayLikeToArray(o, minLen);
+  var n = Object.prototype.toString.call(o).slice(8, -1);
+  if (n === "Object" && o.constructor) n = o.constructor.name;
+  if (n === "Map" || n === "Set") return Array.from(o);
+  if (n === "Arguments" || /^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(n)) return _arrayLikeToArray(o, minLen);
+}
+function _arrayLikeToArray(arr, len) {
+  if (len == null || len > arr.length) len = arr.length;
+  for (var i = 0, arr2 = new Array(len); i < len; i++) arr2[i] = arr[i];
+  return arr2;
+}
+function _nonIterableSpread() {
+  throw new TypeError("Invalid attempt to spread non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method.");
+}
+
+var version = "1.15.6";
+
+function userAgent(pattern) {
+  if (typeof window !== 'undefined' && window.navigator) {
+    return !! /*@__PURE__*/navigator.userAgent.match(pattern);
+  }
+}
+var IE11OrLess = userAgent(/(?:Trident.*rv[ :]?11\.|msie|iemobile|Windows Phone)/i);
+var Edge = userAgent(/Edge/i);
+var FireFox = userAgent(/firefox/i);
+var Safari = userAgent(/safari/i) && !userAgent(/chrome/i) && !userAgent(/android/i);
+var IOS = userAgent(/iP(ad|od|hone)/i);
+var ChromeForAndroid = userAgent(/chrome/i) && userAgent(/android/i);
+
+var captureMode = {
+  capture: false,
+  passive: false
+};
+function on(el, event, fn) {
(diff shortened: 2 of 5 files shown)


evnchn:
> not any vue adaptations as those were a bit old

Wise move. https://github.com/SortableJS/vue.draggable.next is 4 years old. 

To clarify, I think your file `sortable.complete.esm.js` is the same one as in https://www.npmjs.com/package/sortablejs?activeTab=code, right? 

This is because, eventually, we would need a script to fetch it automatically (check out `npm.py` and `npm.json`)

> Currently the example contains a bunch of custom css. It would be best to add them into the sortable.py

Almost. If you have a good bit of reasonable default for styles, they would commonly go to https://github.com/zauberzeug/nicegui/blob/main/nicegui/static/nicegui.css

Typically, styles are applied to elements with class `nicegui-*`. A bunch of other elements got the treatment there. 

It would be great to include just the mininal style so that the sortable doesnt look like hot garbage, but not anything else on top. After all, the user may also want to apply their own styles as well. 

> Any way, I think I have gotten everything to work correctly. 

I'm fine to take your word for it now, but eventually there should be tests. Not sure how we may "drag" stuff with ChromeDriver, though. 

---

@falkoschindler correct me if I am wrong.

EmberLightVFX:
Updated with some missing examples and options for sortablejs.
I also added some default css in the css file :)
The only example not currenty working is the Thresholds one. Need to take a better look at it...

About the npm,py/json I would gladly take some help with that, or some instructions on how that part of NiceGUI works.
[https://www.npmjs.com/package/sortablejs?activeTab=code](https://www.npmjs.com/package/sortablejs?activeTab=code) is the correct one. I'm only using the file **sortable.complete.esm.js** that's under /sortable/modular/sortable.complete.esm.js as it contains all official plugins for sortable.

EmberLightVFX:
I have now fixed the threshold example to fully work and also made it possible to drag cards into other cards in the nested example.

The only thing left is the npm part and I would love some help with that as I have no idea how you would set it up correctly.

falkoschindler:
Thanks for the pull request, @EmberLightVFX! Bringing SortableJS to NiceGUI would be an awesome feature!
I'll review your code as soon as possible. The update of npm.json shouldn't be a problem. I can do that. 👍🏻

evnchn:
@EmberLightVFX Include the following in your `npm.json`:

```json
  "sortable": {
    "package": "sortablejs",
    "destination": "nicegui/elements/lib",
    "keep": [
      "package/modular/sortable.complete.esm.js"
    ],
    "rename": {
      "package/modular/": ""
    },
    "version": "1.15.6"
  }
```

Hopefully, this should download `sortable.complete.esm.js` and place the file at `nicegui/elements/lib/sortable/sortable.complete.esm.js`

falkoschindler (changes requested):
Hi @EmberLightVFX, thanks for this massive pull request!

I roughly reviewed your code and played around with `ui.sortable` for a bit. Here are some of my findings:

- With 1600 lines of NiceGUI code (not including 3300 lines of SortableJS) this PR is huge and hard to review in its current form.
- In contrast to other elements like `ui.aggrid`, `ui.echart`, or `ui.leaflet`, the new `ui.sortable` doesn't simply expect an `options` dictionary, but provides explicit parameters for every single option. Even though this might feel nice from the user perspective, it makes the element very hard to maintain. Every upstream change can break our API, requiring extensive regression tests and potentially slowing down major releases like 3.0. Generic options, on the other hand, reduce complexity on our side while providing full flexibility for the user.
- Alternatively, if you'd be willing to invest that much time and devotion into this element, you could split it into a separate repository and PyPI package. This would clearly communicate that the NiceGUI core development team isn't responsible for maintaining `ui.sortable`. (It's not my prefered option though.)
- We should move the example main.py into the documentation. By splitting it into small, independent, interactive demos they become more accessible and reproducible.
- At the moment there are no pytests at all. We should at least test basic functionality to ensure the library is loaded and sortable content is shown on the screen. Testing drag and drop is challenging, but we should cover it in some basic test.
- While I'm arguing for a more generic implementation, it would be great to have special event arguments including the most relevant information.
- I'm wondering if we could render `ui.sortable` with flex layout by default, like most of NiceGUI's layout elements (`ui.row`, `ui.column`, `ui.card`, `ui.scroll_area`, ...). This would allow you to nest some cards inside `ui.sortable` and it just looks nice.
- On the other hand, when trying to reproduce lists with list items, you might want to nest `ui.item` inside `ui.list` and somehow combine it with `ui.sortable`. This brings me to maybe the most challenging idea: Is "sortable" really a UI element, or a property of _any_ existing container element? I mean, maybe we should reconsider the API and check if we could create something like `with ui.row().sortable(): ui.label(A); ...`.
- I noticed some confusing animation in certain situations, which also happens with the original online examples. So I hope SortableJS is stable enough and won't cause tons of bug reports on our end. (At the time of writing, they have 468 unresolved issues...)

Please don't get me wrong: I love what you did so far and the sheer amount of code is impressive. But before merging anything on main, we have to challenge our decisions. Once a feature is out in the wild, it's very hard to change or take back anything. Allowing to make containers sortable is a very high priority feature and we definitely need something like this in NiceGUI. We just have to make sure to do it right.

Let's start with the most fundamental questions:

- Is SortableJS stable enough to be included in NiceGUI? (I think so.)
- Should this PR and its code remain in this repo? (I do hope so.)
- Could we turn it into a feature of any `ui.element` like `column.sortable()`?

When these questions are answered, other tasks like moving examples, writing tests, simplifying the parameter list, adding special event arguments etc. will be relatively straight forward.

evnchn:
First off, if not for @falkoschindler mentioning, I would not have noticed the sheer scale of this PR. @EmberLightVFX it is impressive work. Be proud of yourselves and have some rest if need be. 

---

But, @falkoschindler I need to address **one important point:** 

> Could we turn it into a feature of any `ui.element` like `column.sortable()`

As you have mentioned, SortableJS may not be the most stable, and we may or may not have to limit SortableJS children elements to an allowlist (can't imagine what sorting Leaflet maps or Plotly diagrams would be like 🤯). 

Turning it into a feature of any `ui.element`, with your wishful programming example of `with ui.row().sortable():`, means we are getting our feets **deep into the water**, which may not be a good idea. 

At least certainly not a good idea now. Maybe later, when a good chunk of people have used SortableJS with NiceGUI and find it stable and generic enough to work with a bunch of elements. But, we'll have to re-review by then. 

---

With that said, I think the following action items in the following order makes the most sense: 

- [ ] Let `ui.sortable` simply expect an `options` dictionary to decouple itself from upstream parameter definition changes, and simplify the NiceGUI implementation by a whole lot. 
  - Remeber NiceGUI is all for [a shallow layer around existing libraries](https://github.com/zauberzeug/nicegui/discussions/4734#discussioncomment-13126503). 
  - This can make the PR more reviewable by reducing line count
- [ ] Move the examples in `examples/sortable/main.py` into separate ones in documentation. 
  - This can make the PR more reviewable by visual confirmation
  - We can also report more explicitly, say, demo 6 is broken, and we only need to look at demo 6 which is isolated. 
  - Also helps with getting new users to use this element. 
- [ ] Adopt `flex` layout if possible
  - Helps make it easy to assess whether it is possible to leverage the entire / most of NiceGUI elements with `ui.sortable`
  - ["flex layout by default" is in our website front page](https://nicegui.io/#:~:text=flex%20layout%20by%20default)
- [ ] Pytest test cases
  - It's super useful since sometimes we break JS libraries when we mess with the underlying infrastructure. Saved us in #4539 or else we would have shipped broken Mermaid. 

Not saying the PR is necessarily "done" whe all these action items are done, but it'll be closer than ever, and we can spot the remaining issues much more easily by then.

EmberLightVFX:
Finally I got some time over to work on this PR again.

Some updates from the feedback!
* Use options dict like aggrid, leaflet etc to match all other elements
* Split up the example into documentation
* Write some pytests
* Use flex layout by default

I added a drag-and-drop function to NiceGUIs screen testing as selenium have that functionality internally. Maybe it should be its own PR but it's only 6 lines of code and (currently) only sortable that uses it, so I added it to this PR.


* `While I'm arguing for a more generic implementation, it would be great to have special event arguments including the most relevant information.`
I did not fully understand what you're after here. Could you explain in another way for me?

-----

If there is anything more that needs to be fixed with this PR, just say and I'll do my best to do it as fast as possible :)

evnchn:
Wow. You sure are efficient! Now I see some potential improvements to the codebase. I am not sure if I can collaborate but I will put them out here first. 

- [ ] Can you do `dependencies=['lib/sortable/sortable.complete.esm.js']` instead of `self.add_resource(Path(__file__).parent / 'lib' / 'sortable')`, then you can `import { default as Sortable } from 'sortable';` in `sortable.js` and change dynamic import to static import, and avoid defining `this.Sortable`. 

{0C4A73BC-9FFA-486E-859D-43B956510D20}

{D1191EB8-65EF-4E98-A45B-95FB0ADB4329}

- [ ] If you already did `default_classes='nicegui-sortable'`, I don't think we need `self.classes('nicegui-sortable')`? 

- [ ] `on_end` to `on_deselect` callbacks, should be `Handler` of `EventArguments` or maybe even `ValueChangeEventArguments`. Do a code search on `Handler[ValueChangeEventArguments]` and you'll instantly get what I mean. 
  - [ ] Then `e` will not be `Any` anymore <br> {7FFE60E3-8ADB-4B29-8A59-5433E5B6F4FA}

---

That's what I can spot so far. There may be more, but it this _should_ be enough to keep you busy for a short while.

evnchn:
idk if it works but I put down all my changes so far in https://github.com/EmberLightVFX/nicegui/pull/1

evnchn:
- [ ] Enable retroactively add handler & mutiple handler, by means of following:
  - Code search `tree.py` and it will give you some inspirations. 
  - ```py
    self._select_handlers = [on_select] if on_select else []
    ```
  - ```py
    def on_select(self, callback: Handler[ValueChangeEventArguments]) -> Self:
        """Add a callback to be invoked when the selection changes."""
        self._props.setdefault('selected', None)
        self._select_handlers.append(callback)
        return self
    ```
  - ```py
        def handle_selected(e: GenericEventArguments) -> None:
            update_prop('selected', e.args)
            for handler in self._select_handlers:
                handle_event(handler, ValueChangeEventArguments(sender=self, client=self.client, value=e.args))
        self.on('update:selected', handle_selected)
    ```

EmberLightVFX:
Thank you @evnchn for all the notes and the PR! I merged it and fixed some small stuff around the code.

I didn't really understand your comment https://github.com/zauberzeug/nicegui/pull/4656#issuecomment-2903289729 about retroactively add handler & mutiple handler so I would love to get that cleared a bit.

Thanks!

evnchn:
@EmberLightVFX _Generally speaking_ a NiceGUI element accepts both passing in the change handler on construction, and retroactively. 

Consider `ui.tree`

{75FAAF3C-DB97-49EA-9693-F7D688DAF63C}

So I wonder if you can make it the case too?

evnchn:
https://github.com/EmberLightVFX/nicegui/pull/2

@EmberLightVFX don't merge right away since idk if this is right or not

@falkoschindler can you see if I am messing things up or making things better? Thanks.

EmberLightVFX:
@evnchn I'm actually working on that right now. I have done exactly like you except I made a generic handle function and passed the correct handler list to it to reduce the amount of code.

I'm currently fighting with pytest after updating this as I'm getting `Message: timeout: Timed out receiving message from renderer: 3.734` for the last tests if I run all the tests. Induvidual runs passes... Does pytest pass fully for you?

evnchn:
Sorry. As I have said I'm developing on Windows, and per test case I only have 50% success rate (WSL does NOT help), making it infeasible to run all tests locally. 

I'm waiting for the GitHub actions result to see how much, if any, did I mess up the tests.

EmberLightVFX:
Oh really? I'm also running on Windows but I have 100% successfully run all the tests. I'm running locally and not any docker over WSL.

I'm using UV so I init my NiceGUI environment like this:
```bash
uv venv -p 3.8

.venv/Scrupts/activate

uv pip install Poetry

Poetry install
```

evnchn:
Tests are green in my attempt: https://github.com/evnchn/nicegui/actions/runs/15212866217

So idk what problem you are facing...

falkoschindler:
> can you see if I am messing things up or making things better? Thanks.

@evnchn Looks good. But it's really a lot of events. Do we really need all of them? Maybe we can reduce them to the most common ones and keep the rest of them available via the generic `.on()`?

EmberLightVFX:
The current state of this code is a bit of a mess.
I'm currently fighting the true clone example. It works mostly but some times the cloning breaks and what ever I do I'm only getting `DOMException: Node.insertBefore: Child to insert before is not a child of this node` errors.

Checking the content of the python object and the DOM object I can see they aren't matching but I don't understand why...

I have also added all missing events like onUnchoose etc. When I have gotten the clone problem fixed we could scale down the amount of events to maybe 1-2 or something.

I added a debug example to the docs where you can see what event gets triggered when. 

So in short: The code is now a bit of a mess while trying to figure out what's going on. After that the code will get cleaned up.

evnchn:
Add oil! Like idk why this is marked with a major. I hope this doesn't stress you out... Love to see the final result 💪

falkoschindler:
Oh, the "major" tag says: "Important, but not urgent". In other words: It would be a really useful feature we'd love to add to NiceGUI, but it's not a critical bug that needs fixing as soon as possible.

Regarding the list of events we need to find a sweet spot. Take `ui.input` for example: It would be a pain to use the generic `.on('update:model-value', ...)` and extract the `value` from GenericEventArguments every time. On the other hand, the user rarely needs the events "blur", "focus" or "clear", so the generic API is a good trade-off.

And maybe we don't need cloning in the very first version of `ui.sortable`. As long as we don't make bad API choices that might prevent us from adding it later, we _can_ decide to focus on the core functionality for now.

phifuh:
Will this work when there are 2 sortables in one page @EmberLightVFX? I have a case where I would like to implement two sortable trees: one for content sorting and another for sorting a list of objects

evnchn:
In principle it should, since there's multiple sortable in the demo page I'd assume. 

Also if not, I'd consider it a bug, and we need to fix it. Basically all UI elements (short of header, footer, left right drawer) can have more than one of each kind, after all.

EmberLightVFX:
It works with any amount of sortable lists!

TM0088:
> It works with any amount of sortable lists!

The example is amazing, most of the features are working great.

I have found that 'show_current_order' cannot get the correct result right after adding item. Properly because item is added as the 1st item by using nicegui 'with' function while item is added to the last in sortablejs logic.

Besides, i can not get 'getSelected', 'select' and 'deselect' function working properly in multi-drag sortable object @EmberLightVFX , the only response from the 3 functions is deselecting all selection, it still in process or may you provide some example?

jochso:
Great feature! Would it be possible to add expandable/collapsible lists as well? I think that is a feature NiceGui would really benefit from 👍

EmberLightVFX:
Update time!
1. Changed the init kwargs to the ones I wrote in my eariler post: https://github.com/zauberzeug/nicegui/pull/4656#issuecomment-2907518159
2. I removed the getSelected function as the selection would get de-selected if you pressed a button or something like that. It's also not part of native SortableJS.
3. Fixed the code for select/deselect so it now works as expected.

I tried to find the bug @TM0088 talked about with show_current_order but I couldn't replicate it.
Could you record a video showing how you get the bug? And if you're using custom code, please share it also.

If there isn't any show_current_order bug then I would say this PR is ready for its last review and merge!

TM0088:
> Update time!
> 1. Changed the init kwargs to the ones I wrote in my eariler post: https://github.com/zauberzeug/nicegui/pull/4656#issuecomment-2907518159
> 2. I removed the getSelected function as the selection would get de-selected if you pressed a button or something like that. It's also not part of native SortableJS.
> 3. Fixed the code for select/deselect so it now works as expected.
> 
> I tried to find the bug @TM0088 talked about with show_current_order but I couldn't replicate it.
> Could you record a video showing how you get the bug? And if you're using custom code, please share it also.
> 
> If there isn't any show_current_order bug then I would say this PR is ready for its last review and merge!

Is there any workaround for getSelected function for multi-drag objects?

EmberLightVFX:
> Is there any workaround for getSelected function for multi-drag objects?

Not currently. As it's a feature outside of SortableJS it's better to add it in a separate PR.

phifuh:
Any status on when this will be merged?

@EmberLightVFX @evnchn have you considered to add expanding/collapseable parent items? I have a usecase where there will be 100+ items so its in dire need 🩸

EmberLightVFX:
I'm waiting for @falkoschindler to review the code but he's swamped with other PRs so no eta yet.

For collapsable lists, just populate the sortable element with expansions: https://nicegui.io/documentation/expansion

phifuh:
@EmberLightVFX I think you missed to implement the on_move argument?

```py
    def __init__(
        self,
        options: Optional[Dict] = None, *,
        on_change: Optional[Handler[GenericEventArguments]] = None,
        on_end: Optional[Handler[GenericEventArguments]] = None,
        on_add: Optional[Handler[GenericEventArguments]] = None,
        on_filter: Optional[Handler[GenericEventArguments]] = None,
        on_spill: Optional[Handler[GenericEventArguments]] = None,
        on_select: Optional[Handler[GenericEventArguments]] = None,
        on_deselect: Optional[Handler[GenericEventArguments]] = None,
        on_cancel_clone: Optional[Handler[GenericEventArguments]] = None,
    ) -> None:
        """Initialize the sortable element.
        Args:
            options: Dictionary of options to pass to SortableJS. See https://github.com/SortableJS/Sortable#options for available options.
            on_change: Callback when the list order changes
            on_end: Callback when element dragging ends
            on_add: Callback when element is dropped into the list from another list
            on_move: Callback when you move an item
            on_filter: Callback when filtered item is clicked
            on_spill: Callback when an item is spilled outside a list
            on_select: Callback when an item is selected (MultiDrag)
            on_deselect: Callback when an item is deselected (MultiDrag)
            on_cancel_clone: Callback when an item is removed from the list into another list
        """
        super().__init__()
     ````

**Comment by EmberLightVFX** [comment]
You get the same respons as if you would use onChange

**Comment by phifuh** [comment]
I might be wrong about this, but my understanding is that they shouldnt be the same. 

In SortableJS, onChange and onMove behave differently. onChange is a passive event—it fires continuously while you drag and you can’t cancel it. onMove, on the other hand, is cancellable: if it returns false, the move is blocked.

Example usecase: I want to use onMove to enforce a maximum nesting depth. When an item is dragged over another item that already has children, I calculate the resulting nesting level (including the dragged item’s own children). If that total would exceed my "maxNestedLevel", I return false from onMove to prevent the drop. Otherwise, the move proceeds and onChange will continue to fire as normal.

**Comment by phifuh** [comment]
@EmberLightVFX What are your thoughts on this?

> I might be wrong about this, but my understanding is that they shouldnt be the same.
> 
> In SortableJS, onChange and onMove behave differently. onChange is a passive event—it fires continuously while you drag and you can’t cancel it. onMove, on the other hand, is cancellable: if it returns false, the move is blocked.
> 
> Example usecase: I want to use onMove to enforce a maximum nesting depth. When an item is dragged over another item that already has children, I calculate the resulting nesting level (including the dragged item’s own children). If that total would exceed my "maxNestedLevel", I return false from onMove to prevent the drop. Otherwise, the move proceeds and onChange will continue to fire as normal.

If this wont be implemented I kinda need to do snapshots of multiple sortable trees to be able to do a rollback to the previous tree state, which I am not a huge fan of 💀

**Comment by phifuh** [comment]
To be honest onMove is still a crucial missing feature. 
Another thing I found is that after a user is sorting some items, the UI is updating but the backend isnt updating the childrens order. 

I implemented another list which i syncronise and where i sort the children, but this shouldnt be the default solution.

Example 
```py
from nicegui import ui, ElementFilter

def dynamic_list_management() -> None:
    def add_deletable_object(label: str):
        with ui.card():
            with ui.row().classes('items-center w-full'):
                ui.label(label).classes('flex-grow')

    with ui.sortable({'fallbackOnBody': True}) as dynamic_sortable:
        for i in range(1, 4):
            add_deletable_object(f'Sortable Item {i} - Drag to reorder')

    with ui.row():

        async def show_current_order():
            items_text = [item.default_slot.children[0].default_slot.children[0].text
                          for item in dynamic_sortable.default_slot.children]
            print(f'Current order: {items_text}')
            print(", ".join(b.text for b in ElementFilter(kind=ui.label)))

        ui.button('Show Current Order', on_click=show_current_order)

dynamic_list_management()

ui.run(title='Sortable', dark=True)
```
result = Sortable Item 1 - Drag to reorder, Sortable Item 2 - Drag to reorder, Sortable Item 3 - Drag to reorder
Even tho items was dragged. 

Further more how do you test it with pytest? Is there implemented an action event like .click() ? onDrag() or something?
And is there a label_1 = user.should_see("text_label", position = 0)
label_1.id == id2
So that I can validate positions of specific elements which got re-ordered?

EmberLightVFX:
Currently I don't have super much time to work on this but if you can you can make a PR on my fork that implements what you feel is missing

falkoschindler:
But please note that - apart from us being busy with the 3.0 release at the moment - the main challenge of this PR is its sheer size. So we should avoid adding too many additional features before merging a first version.

Most of my remarks from https://github.com/zauberzeug/nicegui/pull/4656#pullrequestreview-3087348029 are probably still standing. Addressing them, especially by reducing complexity of the documentation, would be very helpful.

evnchn:
https://github.com/zauberzeug/nicegui/pull/4656/commits/a2aea470f97b23b9b952ccc0c4800e35211348a2: Yes but then our JS is then not updated by any script anymore. Moving forward full migration to #5021's approach is the way to go eventually.

EmberLightVFX:
Thanks @evnchn !
I was actually about to ask about this.

Why removing the npn.json became it's own commit was because something happened in the merge and I ended up with an empty npm file but got said no files has bin changed, so I had to manually remove it. No idea why that happened 🤷

evnchn:
Not urgent, by the way, since nicegui-highcharts still rocking the old approach, and so it has got to work until at least 4.0 where we can make a breaking change and take out the old JS dependency code.

EmberLightVFX:
but npm.json is gone from the root of the project? Or has it moved to somewhere else?

evnchn:
You can check #5021 and what happened there. But in short t's gone and move into each element.

EmberLightVFX:
I'm soon done fixing all items in the review https://github.com/zauberzeug/nicegui/pull/4656#pullrequestreview-3087348029 but I would love some help on how I should add the `esm={}` information as mentioned in https://github.com/zauberzeug/nicegui/pull/5021

I tried to understand by looking on the other elements but I'm quite lost...
This is how the sortable information looked in its npm.json file before:
```json
  "sortable": {
    "package": "sortablejs",
    "destination": "nicegui/elements/lib",
    "keep": [
      "package/modular/sortable.complete.esm.js"
    ],
    "rename": {
      "package/modular/": ""
    },
    "version": "1.15.6"
  }
```

How should I structure the information in `esm={}` with that data?

EmberLightVFX:
I would say that all notes have now bin taken care of.
I have removed 10(!) examples so it's now only those who have edge-cases or interesting cases.
Some of the demos I also merged in the Event Debugging example, like handle and filter so users can look at that code to see how those features work. I hope the amount of examples is acceptable :)

@phifuh You have access to all events that SortableJS has but you might need to add them yourself like I'm doing in the `event debugging` demo. We're only showing the most used and needed ones in the initialization of the element to have it as clean as possible.

EmberLightVFX:
I also just saw, 100 commits in this pr 🥳 😅 and almost 50 comments (52 with this one)

phifuh:
@EmberLightVFX I think there is an error in the  `dynamic_list_management` function.

I am running the example code local and even tho its deleting the element its throwing an error in my console
```
    parent_slot.children.remove(element)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
ValueError: list.remove(x): x not in list
```

phifuh:
Another thing, is it even possible to intersect the serverside sortable.js? 
on_move would be called on the client right now? and wouldnt that be to late to prevent falsy movement?

If you cant execute the inbuild on_move somehow, taking snapshot on_start and restoring it on_end would be the only feasable option?

phifuh:
So I guess something like this 
```python
def validation_with_revert() -> None:
    original_order = []

    def on_start(e: GenericEventArguments):
        """Store the original order of items."""
        nonlocal original_order
        # Store element IDs in order
        original_order = [child for child in sortable1.default_slot.children]
        ui.notify(f"Saved order: {len(original_order)} items")

    def on_end(e: GenericEventArguments):
        """Handle the end of a sorting event."""
        old_index = e.args["oldIndex"]
        new_index = e.args["newIndex"]

        # Validation: Don't allow moving first item to last position
        if old_index == 0 and new_index >= 2:
            ui.notify("Invalid move! Reverting...", color="negative")
            sortable1.sort(list(original_order))

        else:
            ui.notify(f"Valid move: {old_index} → {new_index}", color="positive")

    with ui.card():
        ui.label("Try moving Item 1 to the bottom (will be reverted)").classes("text-h6")
        with ui.sortable() as sortable1:
            sortable1.on("sort_start", on_start)
            sortable1.on("sort_end", on_end)
            for i in range(1, 5):
                with ui.card():
                    ui.label(f"Item {i}")
```
Not sure if thats a usefull example to add

EmberLightVFX:
I'll take a look at the error on Monday.
About the feature you mentioned. I don't know it's currently possible to communicate with the client side js SJS. The solution you came up with looks good tho but to keep the code to the minimal I don't think it's something that will be implemented, at least not now.
After this PR has bin merged it's more than welcome to add features like that if it's of the interest of NiceGUI.

EmberLightVFX:
Fixed the bug and added a test when an item gets removed.
Ready for review if it feels ok @falkoschindler !

falkoschindler:
@EmberLightVFX I was just about to finally review this PR, when I saw your merge commit. Just to be sure: Are you currently working on it? We probably need to update the ESM dependency management to see if everything works with NiceGUI 3.

EmberLightVFX:
I were suppose to update my drop_zone branch but accidentally updated this one 😅 I'm not working on this one.
I would love some help on how to correctly set up the esm dependencie for NiceGUI 3!

falkoschindler:
After thinking about this PR once again, I came to the following conclusion:

1. I'm still not sure about the desired API. Currently we're mimicking SortableJS, and yes, we can drag and drop elements. But to actually _use_ `ui.sortable`, e.g., to organize items in a database, the user has to convert IDs from generic events into useful information, resulting in rather complex user code.
2. I'm still not convinced that SortableJS is the right library for us. It's debatable if the repository is "stale" or just "stable". But there are other strong contenders that need to be considered before regretting a poor decision it later.

In order to keep this discussion focussed on this PR and to reach a broader audience, I posted these two open questions on the corresponding feature request:
- https://github.com/zauberzeug/nicegui/discussions/932#discussioncomment-14935058
- https://github.com/zauberzeug/nicegui/discussions/932#discussioncomment-14935062

Let's pause development on this PR and continue the discussion over there until we've agreed on the desired API and library.

EmberLightVFX:
Sounds all good for me! I made this element for a project of mine but if it should be part of NiceGUI it should follow NiceGUIs needs! If anything it could end up being an external NiceGUI addon that people could pip install. I would just need to know how to make that work with NiceGUI 😄

phifuh:
I can definitely see both the pros and cons of exploring other frameworks. In one of our projects, we rely heavily on both flat and nested sortables, and honestly, we couldn’t live without them. So, in the worst case, I really hope this could at least live on as a plugin or addon.

From my perspective, even if SortableJS might seem a bit “stale,” it’s still extremely capable and stable. It already offers a lot of functionality out of the box and works reliably. So I’m wondering, what concrete new needs would NiceGUI have that SortableJS doesn’t already cover?

I’m even tempted to make a short video showcase, because I genuinely think ui.sortable is awesome and @EmberLightVFX has put a lot of great work into it and It adds a ton of value right now.

Even _**if**_ a different framework were to be chosen in the future, sortable.js could be marked as end-of-life with a legacy period before being moved into a standalone plugin or addon without official support. Realistically, adopting a new drag-and-drop library, designing a new API, reviewing, testing, and polishing it would probably take at least 6–12 months. So having a solid working solution now, even if, at some point in 2026, there’s a decision to adopt a different framework that might be implemented in 2027 seems completely reasonable and pragmatic.

In short, I think it’s worth keeping and using this solution for the time being. It’s functional, stable, and valuable today, while still leaving the door open for future improvements.

evnchn:
Maybe we can get a feel of both our API design and whether SortableJS is good enough, if we have more people using it. 

I propose `ui.beta` API will change in minor version, such that `ui.beta.sortablejs` (sortablejs is intentional, to indicate that we are using SortableJS) can let us experiment. 

Inspired by ChatGPT Python library.

falkoschindler:
Hi everyone,

Let me give a short update on what I've been thinking about the last few weeks.

### Which library

SortableJS is probably good enough for now and we don't need to change it. If we find a good abstraction and the user doesn't have to deal with implementation details, we can always replace the underlying JavaScript library with something else in the future.

### API design

I still don't like the fact that the current implementation requires the user to deal with generic events and map IDs to actual data items. PR #5464 tried to simplify things and gathered valuable [feedback](https://github.com/zauberzeug/nicegui/pull/5464#issuecomment-3556601262).

Right now I'm leaning towards a `.sortable()` method that can be called from container elements like `ui.column`, `ui.row`, `ui.card`, `ui.list`, and a `.sortable_item()` method to mark children as draggable and to attach data:
```py
with ui.column().sortable() as column:
    ui.label('A').sortable_item(data='A')
    ui.label('B').sortable_item(data='B')
    with ui.label('C').sortable_item(data='C', handle='.handle'):
        ui.icon('drag_handle').classes('handle')
```

The `.sortable` namespace can also hold additional properties and methods, e.g. to activate and deactivate the sortable feature:
```py
ui.button('Activate', on_click=column.sortable.activate)
ui.button('Deactivate', on_click=column.sortable.deactivate)
```

The `.sortable()` method can accept an event handler which can access the current data list of the child elements.

### Beta namespace

@evnchn proposed a `ui.beta` namespace to hold experimental features. I'm not so sure about it because NiceGUI could more and more become a collection of unfinished features. Instead I think it would be better to release elements like `ui.sortable` as standalone packages so that users can try them out and provide feedback.

I'm planning to continue working on this feature in January. But of course, I can't promise any release date.

phifuh:
> ### API design
> I still don't like the fact that the current implementation requires the user to deal with generic events and map IDs to actual data items. PR #5464 tried to simplify things and gathered valuable [feedback](https://github.com/zauberzeug/nicegui/pull/5464#issuecomment-3556601262).
> 
> Right now I'm leaning towards a `.sortable()` method that can be called from container elements like `ui.column`, `ui.row`, `ui.card`, `ui.list`, and a `.sortable_item()` method to mark children as draggable and to attach data:
> 
> ```python
> with ui.column().sortable() as column:
>     ui.label('A').sortable_item(data='A')
>     ui.label('B').sortable_item(data='B')
>     with ui.label('C').sortable_item(data='C', handle='.handle'):
>         ui.icon('drag_handle').classes('handle')
> ```
> 
> The `.sortable` namespace can also hold additional properties and methods, e.g. to activate and deactivate the sortable feature:
> 
> ```python
> ui.button('Activate', on_click=column.sortable.activate)
> ui.button('Deactivate', on_click=column.sortable.deactivate)
> ```
> 
> The `.sortable()` method can accept an event handler which can access the current data list of the child elements.


Happy to hear that SortableJS is back in the mix! 🎉

I fully agree that NiceGUI should do the heavy lifting and define a clean, stable API for users. To be honest, I’m not satisfied with the current secondary data.dict() approach in either the nested or flat sortable implementations in our project.

Since we’re close to Christmas, allow me to dream big and rephrase what I already outlined here:
https://github.com/zauberzeug/nicegui/pull/5464#issuecomment-3556601262

### Suggested Core APIs (MVP):

+ get_all_data()
+ insert_at_index(position, data)
+ get_data(by: element_name | id | position)
+ delete(by: element_name | id | position)
+ update(by: element_name | id | position)

These operations allow users to maintain a stable 1:1 relationship between UI elements and application data.
Common use cases, including:

- Simple lists
- Todo apps
- Settings pages
- Dynamic forms

### More Advanced APIs

+ duplicate(by: element_name | id | position)
+ allow_nested_sortable()
+ set_dropzone(size, range, etc.)
+ undo_last_move() -> None
+ set_max_nested_depth(number)
+ get_max_depth()
+ get_path(by: element_name | id | position)
+ sort_by(xxx)
+ delete_all()

I believe these advanced APIs would provide tremendous value, especially since implementing nested structures is never fun.
Creating mechanisms such as blocking moves when max_depth is exceeded is exactly the kind of problem the framework is perfectly positioned to solve for users. Dropzones is another canidate for nested sortables.

I’m also considering a new project involving another sortable use case, where nested sortables visually represent dependencies.
Imagine a questionnaire: if a user answers A, a follow-up question is shown. This relationship could be naturally expressed using nested sortable functionality.

To give more inspiration, I’d like to share the current state of the project.
In the video, you can see:

- Fast loading from JSON
- A nested sortable with two different item types:
  - ui.expansions, which act as nested containers to group items (rename, add, delete)
  - ui.rows, which are normal items with fewer options in the vertical menu (rename, delete)
- Dropzones for items inside ui.expansions
- A maximum depth constraint preventing more than four nested layers (Shown in the bottom of the video as ui.notfication, a little bit hard to see if the video isnt fullscreen)
- A flat sortable where nesting is explicitly prevented

I think the APIs outlined above provide significant value and will make it much easier for users to build powerful, flexible sortable lists in the future 🛩️

https://github.com/user-attachments/assets/37968837-b2f6-4854-a1a3-63a125cbcd0f

EmberLightVFX:
I'm super glad to see conversations about this! I feel I'm a bit low on energy on this PR and I don't have much time to put into it for at least a couple of months so I'm more than happy to give this PR over to @phifuh if you're up for it?

phifuh:
@evnchn @falkoschindler I found 2 issues, but I am not sure if they are related to the sortable branch or have a deeper link. So before opning new issues I thought I ping u guys here.

1) We are still working on the sortable branch and when adding about 100items into the sortable tree the app gets a little bit laggy. (This can most likely be fixed) but when this happens we get a `reloading because handshake failed for clientID xxxx`. Nicegui isnt picking up on the disconnect/handshake errors, because I asume its happnings after a client disconnection inside js? I think it would be very valueable to have that error bubble up or have some sort of heartbeat, even the official website has somethings the "disconnect" toast and I am assuming you guys neither have logs on that? 

Currently we are running a instance of the sortable treecontroller, so its pretty noticable when the whole data vanish :D This could again be fixed by using storage but I am still curios about the `handshake` issue and if we can help tracking it down.

2) When running sortable with nicegui 3.2 we can update a badge which is hooked up to singleclass (Validation storage, which has delta errors/warnings). When running exactly the same code with nicegui 3.6 + sortable branch this ui element wont update. So my question is, was there some "breaking" change from v3.2 -> 3.6 which could have an influence on this? 

```py
from collections.abc import Callable
from typing import ClassVar


class ReportValidationStorage:
    """Tracks total errors/warnings via delta updates using classmethods.

    Templates call update() with delta values when their validation state changes.
    Workspace subscribes via add_listener() to react to changes.

    Example:
        ReportValidationStorage.update(error_delta=-1, warning_delta=0)
        total = ReportValidationStorage.total_errors()

    """

    _total_errors: ClassVar[int] = 0
    _total_warnings: ClassVar[int] = 0
    _listeners: ClassVar[list[Callable[[], None]]] = []

    @classmethod
    def reset_issues(cls) -> None:
        """Reset the amount of warnings and errors to 0, called by 'load json data'."""
        cls._total_warnings = 0
        cls._total_errors = 0

    @classmethod
    def add_listener(cls, callback: Callable[[], None]) -> None:
        """Subscribe to validation state changes.

        Args:
            callback: Function to call when totals change (no arguments).

        """
        if callback not in cls._listeners:
            cls._listeners.append(callback)

    @classmethod
    def remove_listener(cls, callback: Callable[[], None]) -> None:
        """Unsubscribe from validation state changes.

        Args:
            callback: The callback to remove.

        """
        if callback in cls._listeners:
            cls._listeners.remove(callback)

    @classmethod
    def update(cls, error_delta: int, warning_delta: int) -> None:
        """Update totals with delta values and notify listeners.

        Args:
            error_delta: Change in error count (+1, -1, 0, etc.)
            warning_delta: Change in warning count (+1, -1, 0, etc.)

        """
        if error_delta == 0 and warning_delta == 0:
            return  # No change, skip notification

        cls._total_errors += error_delta
        cls._total_warnings += warning_delta

        # Ensure we don't go negative (safety check)
        cls._total_errors = max(0, cls._total_errors)
        cls._total_warnings = max(0, cls._total_warnings)

        cls._notify()

    @classmethod
    def _notify(cls) -> None:
        """Notify all listeners of state change."""
        for listener in cls._listeners:
            listener()

    @classmethod
    def total_errors(cls) -> int:
        """Get the total error count across all templates."""
        return cls._total_errors

    @classmethod
    def total_warnings(cls) -> int:
        """Get the total warning count across all templates."""
        return cls._total_warnings

    @classmethod
    def has_errors(cls) -> bool:
        """Check if there are any errors."""
        return cls._total_errors > 0

    @classmethod
    def has_warnings(cls) -> bool:
        """Check if there are any warnings."""
        return cls._total_warnings > 0

```

```py
    def update_validation_badges(self) -> None:
        """Update the error/warning chips from ReportValidationStorage. O(1) operation."""
        errors = ReportValidationStorage.total_errors()
        warnings = ReportValidationStorage.total_warnings()

        if self._error_chip is not None:
            error_text = f"{errors} Kritiske fejl"
            self._error_chip.set_text(error_text)
            self._error_chip.set_visibility(False)
            if errors > 0:
                self._error_chip.set_visibility(True)

        if self._warning_chip is not None:
            warning_text = f"{warnings} Advarsler"
            self._warning_chip.set_text(warning_text)
            self._warning_chip.set_visibility(False)
            if warnings > 0:
                self._warning_chip.set_visibility(True)
```

& we simply do `ReportValidationStorage.update(error_delta, warning_delta)`

evnchn:
I still haven't got myself into this PR, but: 

- If your JS lags hard enough that you see `reloading because handshake failed for clientID xxxx` then you may have blocked the JS main thread long enough that Socket.IO connection cannot be sustained. While #5784 will salvage the situation, it is not ideal. 
- Nothing comes to mind for big changes in 3.2 through 3.6 which could break your `ReportValidationStorage`. I would recommend that you can bisect by yourself to land on a specific PR which broke things.

phifuh:
> I still haven't got myself into this PR, but:
> 
> * If your JS lags hard enough that you see `reloading because handshake failed for clientID xxxx` then you may have blocked the JS main thread long enough that Socket.IO connection cannot be sustained. While [Use Web Worker heartbeat to keep connection alive during blocking dialogs #5784](https://github.com/zauberzeug/nicegui/pull/5784) will salvage the situation, it is not ideal.
> * Nothing comes to mind for big changes in 3.2 through 3.6 which could break your `ReportValidationStorage`. I would recommend that you can bisect by yourself to land on a specific PR which broke things.

@evnchn Thanks I will give it a try to identify the change. 

Regarding the WebSocket overload, we currently validate each template sequentially. I wondered whether running the validations asynchronously and in parallel could help, even though we have only one FastAPI worker. In theory, some threads would remain idle. However, I’m concerned that this approach might flood the WebSocket even more.

So the question is: how can we reduce the strain on the WebSocket while still delivering the necessary UI updates?
As mentioned in: https://github.com/zauberzeug/nicegui/discussions/4758#discussioncomment-15823092 we have the communication broker -> validate / update ui flow.

falkoschindler:
Thanks for reporting these, @phifuh. As @evnchn noted, issue 1 is likely a general Socket.IO/WebSocket resilience problem (see #5784), and issue 2 sounds like a NiceGUI version regression unrelated to sortable -- `git bisect` should help narrow it down. Either way, both deserve their own issues or discussion threads rather than being discussed here.

Which brings me to a broader point about this PR.

With 100+ commits, 2500+ lines, and 70+ comments covering many side topics, this PR has become very hard to track and maintain. I'm still working on a concept for a proper sortable/draggable API built into NiceGUI, and this PR in its current form probably won't be merged as-is. The API surface isn't where it needs to be for long-term stability, and since I only find time for this every now and then, I can't give a timeline for when it'll be ready.

That said, I know people are already using and relying on this code, and I don't want to leave you stranded on an unreleased branch. Therefore I'd propose to move the current sortable implementation to a separate repository (e.g. `nicegui-sortable` or similar). This would:

- Provide official releases so users don't have to depend on a feature branch
- Give room for focused discussions about sortable-specific issues (like the two above)
- Allow the community to iterate and improve independently

Meanwhile, I'll continue working on the built-in NiceGUI API at my own pace, drawing on the lessons learned here. When that's ready, we can look at how to provide a smooth migration path.

Denubis:
I'm also using a versioned instance of this in my code, so having this be a proper extension would be very nice. Currently struggling to get the e2e draggables testing properly.

falkoschindler:
Thanks once again for this PR, @EmberLightVFX — it kicked off a great discussion and helped shape our thinking on how to best integrate SortableJS into NiceGUI.

After extensive discussion in #932, we concluded that "sortable" is better modeled as a _feature on existing containers_ rather than a standalone element. A `ui.column` that can be reordered is still a column. PR #5855 implements this approach via a `.make_sortable()` method on container elements like `ui.column`, `ui.row`, `ui.card`, etc.

Closing this in favor of #5855. Your contribution and the ideas from this PR were instrumental in getting us there — thank you!

PR #4658, opened 2025-04-25
Added documentation on how native app settings are ignored when using a main guard

civilwargeeky:
Following discussion in [Pull Request 4627](https://github.com/zauberzeug/nicegui/pull/4627), it was concluded that we should  change documentation instead of code to deal with the fact that native mode settings are not applied if they come from behind a main guard.

This PR adds documentation to the [Configuration & Deployment](https://nicegui.io/documentation/section_configuration_deployment#native_mode) page, along with an example of bad code and good code.

Feel free to remove the examples if they take up too much space or are unclear.

### DIFF
--- a/website/documentation/content/section_configuration_deployment.py
+++ b/website/documentation/content/section_configuration_deployment.py
@@ -52,6 +52,44 @@ def native_mode_demo():
     # END OF DEMO
     ui.button('enlarge', on_click=lambda: ui.notify('window will be set to 1000x700 in native mode'))
 
+# Currently, options passed via app.native are not used if they are set behind a main guard
+# See discussion at: https://github.com/zauberzeug/nicegui/pull/4627
+doc.text('', '''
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Very nice, @civilwargeeky! This demonstrates the point very clearly. 👍🏻

I just wrapped the text on multiple lines and shortened the demo codes to reduce duplication. Ready to merge!

J3ronimo:
>Therefore any configuration changes from code run under a main guard is ignored by the native app.

The main pitfall I guess is that any configuration changes that are not run at load-time of the module are ignored,
even if those lines of code were executed initially by the main process. 

I know this may be a bit outside of nicegui's scope, but it's currently what prevents launching nicegui native apps from [script entrypoints](https://packaging.python.org/en/latest/specifications/pyproject-toml/#entry-points), since those always have to link to functions, which - under what was said above - are never run at load time. 🫤 Or in other words: whenever your way of launching a native app is more complex than just running a module as `__main__`, those configs may break.

evnchn:
We can potentially have the functionality in #4627 behind a Boolean on-off, so as to satiate your needs (assuming code refactoring is not a way out, while user has to acknowledge the limitation on unpickleable values)

Maybe you can PR as such?

PR #4665, opened 2025-04-25
fix: do not use global objects to pass information around between processes

chaosct:
I am being affected by a very similar instance of https://github.com/zauberzeug/nicegui/discussions/1714

In a big project, somehow the global objects are not copied in the new process after multiprocessing.Process is started.

It is happening to me when running it without compilation through pyinstaller. I am, however, unable to reproduce this with a small example.

Anyway, this is a small change that solves this issue for me, and I guess for other people that may encounter the same issue in other situations, as it happens in the linked issue.

Otherwise, the change is innocuous.

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -14,7 +14,7 @@ from typing import Any, Callable, Dict, List, Tuple
 from .. import core, helpers, optional_features
 from ..logging import log
 from ..server import Server
-from . import native
+from . import NativeConfig, native
 
 try:
     with warnings.catch_warnings():
@@ -28,7 +28,7 @@ except ModuleNotFoundError:
 
 def _open_window(
     host: str, port: int, title: str, width: int, height: int, fullscreen: bool, frameless: bool,
-    method_queue: mp.Queue, response_queue: mp.Queue,
+    native_config: NativeConfig, method_queue: mp.Queue, response_queue: mp.Queue,
 ) -> None:
     while not helpers.is_port_open(host, port):
         time.sleep(0.1)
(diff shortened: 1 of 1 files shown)


falkoschindler:
Hi @chaosct, thanks for the pull request!

It looks very similar to #4627, which has been closed due to concerns about the native config not being picklable in all cases. But maybe this isn't a problem to fix your problem.

chaosct:
Thanks for the quick response. If this is a concern, I could craft a PR
that uses only native types. I am very interested to solve this upstream.

El dv., 25 d’abr. 2025, 18:06, Falko Schindler ***@***.***>
va escriure:

FabianGoessling:
Hi @falkoschindler, 

Do you see this connected to #1714 ? I will take a look if PR solves #1714, this would finally allow me to update my nicegui version and get rid of my "in-prod-monkeypatch"..

chaosct:
> Hi @chaosct, thanks for the pull request!
> 
> It looks very similar to #4627, which has been closed due to concerns about the native config not being picklable in all cases. But maybe this isn't a problem to fix your problem.

I just looked into #4627 and this is EXACTLY what is happening to me. I realised that this happens when I run the program as a pip script.

falkoschindler:
Brief overview over the two very similar PRs:
- This PR #4665 attempts to solve #1714.
- PR #4627 wanted to solve issue #4319, but was replaced by PR #4658 to update the documentation.

@chaosct Do I understand correctly that the improved documentation from PR #4658 doesn't help in your case?

> I am, however, unable to reproduce this with a small example.

This is very strange. I'd love to better understand what it takes to reproduce the problem before merging a solution.

> I could craft a PR that uses only native types.

How would that be possible? The config object is controlled by the user, not by the NiceGUI library.

chaosct:
Sorry for the confusion, @falkoschindler I assumed and responded before looking into the links with sufficient detail.

What was in fact impacting me was setting app.native stuff inside a function, instead of in the global scope, so it can be executed again when run through multiprocessing.

So, yes, PR https://github.com/zauberzeug/nicegui/pull/4658 in fact solves my problem.

After this change I got it working.

I'll delete this PR as it is no longer relevant. Thanks!

PR #4671, opened 2025-04-28
Add `reset`, `trigger`, and change handler for `ui.timer`

evnchn:
This PR fix #4670 by adding `reset()` and `trigger()` methods to `ui.timer`, enabling the user to skip the existing sleep when changing the interval, while choosing between not running (for `reset()`) or running (for `trigger()`) the callback while doing so. 

Notably, `_run_once` is removed, since it is possible to `reset()` a timer which is supposed to run once again and again, and the code needs a while loop logic, which I leverage the existing `_run_in_loop` to achieve. Rather, now we raise `AlreadyRanOnceError` if `once` is set, which also enables setting `once` dyanmically while the `ui.timer` is running. 

Some remaining issues: 
* Cannot do `timer.on_interval_changed(timer.trigger)` since then `timer.trigger` would receive too many arguments
* Let alone the one-line convenience method `timer = ui.timer(10, last_update.refresh, immediate=False, on_interval_changed=TRIGGER)`, since otherwise `timer = ui.timer(10, last_update.refresh, immediate=False, on_interval_changed=timer.trigger)` would not work since `timer` is not defined.

### DIFF
--- a/nicegui/timer.py
+++ b/nicegui/timer.py
@@ -1,16 +1,23 @@
 import asyncio
 import time
 from contextlib import nullcontext
-from typing import Any, Awaitable, Callable, ContextManager, Optional
+from typing import Any, Awaitable, Callable, ContextManager, Optional, cast, Self
 
 from . import background_tasks, core
 from .awaitable_response import AwaitableResponse
 from .binding import BindableProperty
 
 
+class AlreadyRanOnceError(Exception):
+    """Exception raised when the timer has already run once, so as to stop the outer while loop in _run_in_loop."""
+    pass
+
+
 class Timer:
-    active = BindableProperty()
-    interval = BindableProperty()
+    active = BindableProperty(
+        on_change=lambda sender, active: cast(Self, sender)._handle_active_change(active))  # pylint: disable=protected-access
+    interval = BindableProperty(
+        on_change=lambda sender, interval: cast(Self, sender)._handle_interval_change(interval))  # pylint: disable=protected-access
 
     def __init__(self,
                  interval: float,
@@ -18,6 +25,8 @@ class Timer:
                  active: bool = True,
                  once: bool = False,
                  immediate: bool = True,
+                 on_active_changed: Optional[Callable[[bool], None]] = None,
+                 on_interval_changed: Optional[Callable[[float], None]] = None,
                  ) -> None:
         """Timer
 
@@ -28,21 +37,41 @@ class Timer:
         :param interval: the interval in which the timer is called (can be changed during runtime)
         :param callback: function or coroutine to execute when interval elapses
         :param active: whether the callback should be executed or not (can be changed during runtime)
-        :param once: whether the callback is only executed once after a delay specified by `interval` (default: `False`)
+        :param once: whether the callback is only executed once after a delay specified by `interval` (default: `False`, can be changed during runtime)
         :param immediate: whether the callback should be executed immediately (default: `True`, ignored if `once` is `True`, *added in version 2.9.0*)
+        :param on_active_changed: callback which is invoked when the active state is changed (default: `None`)
+        :param on_interval_changed: callback which is invoked when the interval is changed (default: `None`)
         """
         super().__init__()
         self.interval = interval
         self.callback: Optional[Callable[..., Any]] = callback
         self.active = active
         self._is_canceled = False
-        self._immediate = immediate
+        self.once = once
+        self._immediate = immediate if not once else False
+        self._should_abort_sleep = asyncio.Event()
+        self._skip_callback_once_for_reset = False
+        self._on_active_changed: Optional[Callable[[bool], None]] = None
+        self._on_interval_changed: Optional[Callable[[float], None]] = None
 
-        coroutine = self._run_once if once else self._run_in_loop
+        if on_active_changed:
+            self.on_active_changed(on_active_changed)
+        if on_interval_changed:
+            self.on_interval_changed(on_interval_changed)
+
+        coroutine = self._run_in_loop
         if core.app.is_started:
             background_tasks.create(coroutine(), name=str(callback))
         else:
             core.app.on_startup(coroutine)
+
+    def _handle_active_change(self, active: bool) -> None:
+        if self._on_active_changed:
+            self._on_active_changed(active)
+
+    def _handle_interval_change(self, interval: float) -> None:
+        if self._on_interval_changed:
+            self._on_interval_changed(interval)
 
     def _get_context(self) -> ContextManager:
         return nullcontext()
@@ -60,21 +89,41 @@ class Timer:
         """Cancel the timer."""
         self._is_canceled = True
 
-    async def _run_once(self) -> None:
+    def trigger(self) -> None:
+        """Abort sleep to reset the remaining sleep time to the interval, triggering the callback immediately in doing so."""
+        assert not self._is_canceled, 'Cannot trigger a canceled timer'
+        self._should_abort_sleep.set()
+
+    def reset(self) -> None:
+        """Abort sleep to reset the remaining sleep time to the interval, but without triggering the callback."""
+        assert not self._is_canceled, 'Cannot reset a canceled timer'
+        self._skip_callback_once_for_reset = True
+        self.trigger()
+
+    def on_active_changed(self, callback: Callable[[bool], None]) -> Self:
+        """Set a callback which is invoked when the active state is changed."""
+        self._on_active_changed = callback
+        return self
+
+    def on_interval_changed(self, callback: Callable[[float], None]) -> Self:
+        """Set a callback which is invoked when the interval is changed."""
+        self._on_interval_changed = callback
+        return self
+
+    async def _sleep_with_abort(self, seconds: float) -> None:
+        """Sleep for a given number of seconds, but allow to abort the sleep."""
+        if seconds <= 0:
+            return
         try:
-            if not await self._can_start():
-                return
-            with self._get_context():
-                await asyncio.sleep(self.interval)
-                if self.active and not self._should_stop():
-                    await self._invoke_callback()
-        finally:
-            self._cleanup()
+            await asyncio.wait_for(self._should_abort_sleep.wait(), timeout=seconds)
+            self._should_abort_sleep.clear()
+        except asyncio.TimeoutError:
+            pass
 
     async def _run_in_loop(self) -> None:
         try:
             if not self._immediate:
-                await asyncio.sleep(self.interval)
+                await self._sleep_with_abort(self.interval)
             if not await self._can_start():
                 return
             with self._get_context():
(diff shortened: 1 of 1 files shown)


evnchn:
Sorry forgot attach test code: 

```py
from datetime import datetime

from nicegui import ui

def set_last_change_value_time_label(label: ui.label) -> None:
    last_change_value_time = datetime.now().strftime("%H:%M:%S")
    label.set_text(f'Last Change Value Time: {last_change_value_time}')

@ui.page("/on_interval_changed_trigger")
def on_interval_changed_trigger_page() -> None:
    ui.label("This label triggers on interval change").classes('text-2xl')

    @ui.refreshable
    def last_update():
        last_update_time = datetime.now().strftime("%H:%M:%S")
        ui.label(f'Last Update: {last_update_time}')
        ui.notify(f'Last Update: {last_update_time}')

    last_update()

    last_change_value_time_label = ui.label()
    set_last_change_value_time_label(last_change_value_time_label)

    timer = ui.timer(10, last_update.refresh, immediate=False)

    def my_on_interval_changed(interval: float) -> None:
        set_last_change_value_time_label(last_change_value_time_label)
        timer.trigger()

    timer.on_interval_changed(my_on_interval_changed)

    ui.select(options=[10, 20, 30, 600], value=10, label='Refresh Interval (s)').props('stack-label'
            ).classes('w-40'
            ).bind_value(timer, 'interval' ) # changes the timer interval for next run
    
    ui.link("Back to main page", "/")

@ui.page("/on_interval_changed_reset")
def on_interval_changed_reset_page() -> None:
    ui.label("This label resets on interval change").classes('text-2xl')
    @ui.refreshable
    def last_update2():
        last_update_time = datetime.now().strftime("%H:%M:%S")
        ui.label(f'Last Update: {last_update_time}')
        ui.notify(f'Last Update: {last_update_time}')

    last_update2()

    last_change_value_time_label2 = ui.label()
    set_last_change_value_time_label(last_change_value_time_label2)

    timer2 = ui.timer(10, last_update2.refresh, immediate=False)

    def my_on_interval_changed2(interval: float) -> None:
        set_last_change_value_time_label(last_change_value_time_label2)
        timer2.reset()

    timer2.on_interval_changed(my_on_interval_changed2)

    ui.select(options=[10, 20, 30, 600], value=10, label='Refresh Interval (s)').props('stack-label'
            ).classes('w-40'
            ).bind_value(timer2, 'interval' ) # changes the timer interval for next run\
    
    ui.link("Back to main page", "/")

@ui.page("/")
def main_page() -> None:
    ui.link("See on_interval_changed_trigger", "/on_interval_changed_trigger")
    ui.link("See on_interval_changed_reset", "/on_interval_changed_reset")

    def run_me_once():
        print('run me once')

    ui.button('Run me once, after 3 seconds', on_click=lambda: ui.timer(3, run_me_once, once=True))

ui.run(show=False,reload=False)
```

falkoschindler:
Hi @evnchn,
Thanks for another promissing pull request! I just briefly looked into it (haven't read the complete diff though).
To clarify, despite "some remaining issues", the following code works:
```py
t = ui.timer(1.0, lambda: ui.notify('Tick'))
t.on_interval_changed(lambda _: t.trigger())
ui.slider(min=0.1, max=3.0, step=0.1).bind_value_to(t, 'interval')
```

I even thought you could write
```py
t = ui.timer(1.0, lambda: ui.notify('Tick')).on_interval_changed(lambda e: e.sender.reset())
ui.number(step=0.1).bind_value(t, 'interval')
```
But since the change handler is called with the new interval, there is no `sender`. This is a bit unexpected, because usually event handlers are called with `EventArguments`, or `ValueChangeEventArguments` in this case. Do you think it's possible to change `on_active_changed` and `on_interval_changed` in this regard? This would solve both issues you mentioned, because using `handle_event` from events.py will automatically make the argument optional, allowing to write
```py
t.on_interval_changed(t.reset)
```
and it will allow one-liners like
```py
ui.timer(1.0, lambda: ui.notify('Tick'), on_interval_changed=lambda e: e.sender.reset())
```
Searching for `handle_event` in the code base should give you a good reference for how to use it to call event handlers with event arguments.

evnchn:
Aha! Sure thing, will do. Sort of forgot about those `EventArgument` things while developing this PR.

evnchn:
@falkoschindler It appears that the situation is more complex than you may think. Please double-check my work before merging, since I am really not confident with this one. 

This is because, there is 2 variant of timer, one in `timer.py` (for `app.timer`), one in `elements/timer.py` (for `ui.timer`)

The former, as it seems, have no access to `client` whatsoever. But it's still useful to know which timer was the sender, so the solution for the former was: 

```py
@dataclass(**KWONLY_SLOTS)
class BaseTimerIntervalChangeEventArguments(EventArguments):
    sender: 'Timer'
    interval: float


@dataclass(**KWONLY_SLOTS)
class BaseTimerActiveChangeEventArguments(EventArguments):
    sender: 'Timer'
    active: bool
```

While the latter gets: 

```py
@dataclass(**KWONLY_SLOTS)
class TimerIntervalChangeEventArguments(BaseTimerIntervalChangeEventArguments):
    client: Client


@dataclass(**KWONLY_SLOTS)
class TimerActiveChangeEventArguments(BaseTimerActiveChangeEventArguments):
    client: Client
```

Then, we override the `__init__`, `_handle_XXX_change` and `on_XXX_changed` and it appears to work. 

Updated test script (yours isn't working, since (1) it tries to assign None to interval, breaking the code, (2) it is difficult to change the value exactly once)

```py
timer_ui = ui.timer(1.0, lambda: ui.notify('UI Timer Tick'))
timer_ui.on_interval_changed(timer_ui.trigger)
timer_ui.on_interval_changed(print_event)
ui.select(options=[10, 20, 30, 600], value=10, label="UI Timer Interval").bind_value_to(
    timer_ui, 'interval').classes('w-full')

timer_app = app.timer(1.0, lambda: print('App Timer tick'))
timer_app.on_interval_changed(timer_app.trigger)
timer_app.on_interval_changed(print_event)
ui.select(options=[10, 20, 30, 600], value=10, label="App Timer Interval").bind_value_to(
    timer_app, 'interval').classes('w-full')
```

Checking, you indeed see `TimerIntervalChangeEventArguments` if it came from `ui.timer`, `BaseTimerIntervalChangeEventArguments` if it came from `app.timer`.

evnchn:
@falkoschindler Last thing to fix: get pylint to stay silent. I can't quite do that. VSCode is fine with the type hint and the assignment, showing TimerIntervalChangeEventArguments appropriately on mouse hover. 

```py
nicegui/elements/timer.py:53: error: Incompatible types in assignment (expression has type "List[Union[Callable[[TimerActiveChangeEventArguments], Any], Callable[[], Any]]]", variable has type "List[Union[Callable[[BaseTimerActiveChangeEventArguments], Any], Callable[[], Any]]]")  [assignment]
nicegui/elements/timer.py:54: error: Incompatible types in assignment (expression has type "List[Union[Callable[[TimerIntervalChangeEventArguments], Any], Callable[[], Any]]]", variable has type "List[Union[Callable[[BaseTimerIntervalChangeEventArguments], Any], Callable[[], Any]]]")  [assignment]
nicegui/elements/timer.py:56: error: List item 0 has incompatible type "Union[Callable[[TimerActiveChangeEventArguments], Any], Callable[[], Any]]"; expected "Union[Callable[[BaseTimerActiveChangeEventArguments], Any], Callable[[], Any]]"  [list-item]
nicegui/elements/timer.py:57: error: List item 0 has incompatible type "Union[Callable[[TimerIntervalChangeEventArguments], Any], Callable[[], Any]]"; expected "Union[Callable[[BaseTimerIntervalChangeEventArguments], Any], Callable[[], Any]]"  [list-item]
nicegui/elements/timer.py:69: error: Argument 1 to "append" of "list" has incompatible type "Union[Callable[[TimerActiveChangeEventArguments], Any], Callable[[], Any]]"; expected "Union[Callable[[BaseTimerActiveChangeEventArguments], Any], Callable[[], Any]]"  [arg-type]
nicegui/elements/timer.py:74: error: Argument 1 to "append" of "list" has incompatible type "Union[Callable[[TimerIntervalChangeEventArguments], Any], Callable[[], Any]]"; expected "Union[Callable[[BaseTimerIntervalChangeEventArguments], Any], Callable[[], Any]]"  [arg-type]
```

evnchn:
Done. Code check passed after I refactor to let [`ui.timer` maintain a separate list of handlers, avoid conflict](https://github.com/zauberzeug/nicegui/pull/4671/commits/93b34bcc9be1dcd21fb922f7640d6769df25f928)

```
Success: no issues found in 353 source files
------------------------------------
Your code has been rated at 10.00/10
```

evnchn:
@falkoschindler Here is a problem which I think you can potentially solve much better than I can. 



See, on Python 3.8-3.9, `self._should_abort_sleep = asyncio.Event()` sets the Event up with `self._should_abort_sleep._loop = <ProactorEventLoop running=False closed=False debug=False>`, whereas future versions sets it up as `self._should_abort_sleep._loop = None`. 

As such, on `await asyncio.wait_for(self._should_abort_sleep.wait(), timeout=seconds)`, Python 3.8-3.9 complaints that it is "attached to a different loop" and refuse to work. 

My approach: 

1. try-except, reattach a proper Event if it fails (currently using)

```py
    async def _sleep_with_abort(self, seconds: float) -> None:
        """Sleep for a given number of seconds, but allow to abort the sleep."""
        if seconds <= 0:
            return
        while True:
            try:
                await asyncio.wait_for(self._should_abort_sleep.wait(), timeout=seconds)
                self._should_abort_sleep.clear()
                break
            except asyncio.TimeoutError:
                break
            except RuntimeError:
                # This is a workaround for Python 3.8-3.9 where the `._loop` is explicitly set (future versions have None as default),
                # causing "attached to a different loop" error upon awaiting the event.
                self._should_abort_sleep = asyncio.Event()
```
> Seems sketchy that we are reattaching the event...

2. Check explicitly the `self._should_abort_sleep._loop`. Something like: 

```py
    def _initialize_should_abort_sleep(self) -> None:
        """Initialize the event used to abort sleep."""
        if self._should_abort_sleep_internal is None or self._should_abort_sleep_internal._loop != asyncio.get_event_loop():
            print(
                f"!!!!! Timer._initialize_should_abort_sleep() - creating new event because {self._should_abort_sleep_internal._loop} != {asyncio.get_event_loop()}")
            self._should_abort_sleep_internal = asyncio.Event()
```

> Accessing `._loop` comes with no stability guarantees...

Do you have any better solution? Thanks!

rodja:
In 83dfe7b2923 we replaced all `asyncio.wait_for` calls with `wait_for2.wait_for` because the Python version is buggy when trying to cancel. This should also solve your problem with Python 3.8-3.9.

evnchn:
Thanks for the heads up @rodja but I doubt it will fix things. 

It is because, `wait_for2` mostly handles what happens when you cancel and when it times out. 

But with Python 3.8-3.9, you can't even begin to await in the first place, since the event loop is considered mismatched. 

You can consider Python 3.10+ having a "tolerant approach" by simply having `._loop=None` meaning using whatever event loop, and that approach simply doesn't exist for Python 3.8-3.9 AFAIK. 

Regardless, I'll switch to `wait_for2` for consistency. I'll also do a code search for the old `wait_for` and change out any instances of it if so.

evnchn:
Proposed documentation text

#### Reset or Trigger timer on interval change

It may be desirable to immediately honor the new interval when it is set. 

For example, if a timer was set with long interval (600 seconds), and it was updated to shorter interval (10 seconds), the intention may not be to wait for the remaining time, which may be well in excess of the new interval. 

`ui.timer` supports `on_interval_changed` to run a callback of your desire, which can be: 

- `.reset`: Simply reset the remaining time to the updated interval, but will not immediately run the timer callback. 
- `.trigger`: The above, but also runs the timer callback (imagine turning a clock forward until the alarm rings)
- A custom callback which you can run the above methods based on whether the interval was increased or decreased.

We believe that the default behavior, reset-on-change, trigger-on-change all have their own use cases, and we offer the maximum flexibility to the developer.

evnchn:
Documentation done

evnchn:
Slightly a bit over-engineered, possibly.

Maybe not need the change event handlers since, y-know, the person triggering and resetting the timer in Python-land knows that they are doing that, so no need to emit an event for something they are doing. 

And the merge conflicts.

falkoschindler:
> Slightly a bit over-engineered, possibly.
>
> Maybe not need the change event handlers [...]
>
> And the merge conflicts.

@evnchn Given your concerns, should we simply close this PR? Maybe it would be more efficient to restart from scratch - if there is demand. Otherwise I'd need to ask you to update the PR and resolve any conflicts. 😕

evnchn:
Let's close it for now. It'll come back in another way later™

PR #4675, opened 2025-04-28
Make passing `Element` to `getElement` work

### LINKED ISSUE #4674: Passing `HTMLElement` to `getElement` never works
### First Check

### Example Code

```js
c30 instanceof HTMLElement // true

getElement(c30) // null
```

### Description

Go to NiceGUI website

Go to DevTools and run `getElement(c30)`, returns `null`, despite `c30` is well-defined as a `HTMLElement` (do `c30 instanceof HTMLElement` to verify)

Image

### NiceGUI Version

2.15.0

### Python Version

3.12.9

### Browser

Samsung Internet, Firefox, Chrome, Edge

### Operating System

Windows

### Additional Context

https://github.com/zauberzeug/nicegui/blob/d9bf81f35ac5d89958765dd1664d48672ee9e9b7/nicegui/static/nicegui.js#L37-L40

This infers it trying to do `mounted_app.$refs["rc30"]` upon `getElement(c30)`, since `_id` is set to "c30" by `id.id`. 

Shouldn't we drop the "c" to get the numeric ID, then do `mounted_app.$refs["r30"]`?

Seems like an obvious logic bug...

evnchn:
Fix #4674 and other issues. 

Most obviously: Passing HTMLElement now works (consider `getElement(c30)`)

More implicitly: Checking `id instanceof HTMLElement` is wrong logic, since the `id` property is introduced by `Element` (the base class which `HTMLElement` inherits from). 

Otherwise, code would not work, if we pass in, say, an `SVGElement`. 

Check here for more details about `Element.id`: https://developer.mozilla.org/en-US/docs/Web/API/Element/id

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -35,7 +35,7 @@ function replaceUndefinedAttributes(element) {
 }
 
 function getElement(id) {
-  const _id = id instanceof Element ? id.id : id;
+  const _id = id instanceof Element ? id.id.slice(1) : id;
   return mounted_app.$refs["r" + _id];
 }
 
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -35,7 +35,7 @@ function replaceUndefinedAttributes(element) {
 }
 
 function getElement(id) {
-  const _id = id instanceof HTMLElement ? id.id : id;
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Great catch! Thanks, @evnchn! 👍🏻

PR #4676, opened 2025-04-28
Simplify change set processing for `ui.codemirror`

falkoschindler:
In preparation for PR #4578, this PR significantly simplifies the processing of change sets in `ui.codemirror`.

An optimization, the inner while loop, has been removed. I think user input, which is processed in `_event_args_to_value`, usually consists of one section only and we don't need the more complex algorithm to process consecutive changes at once.

### DIFF
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -1,6 +1,6 @@
 from itertools import zip_longest
 from pathlib import Path
-from typing import List, Literal, Optional, get_args
+from typing import List, Literal, Optional, Tuple, cast, get_args
 
 from nicegui.elements.mixins.disableable_element import DisableableElement
 from nicegui.elements.mixins.value_element import ValueElement
@@ -341,7 +341,8 @@ def _apply_change_set(doc, sections: List[int], inserted: List[List[str]]) -> st
     assert sum(sections[::2]) == len(doc), 'Cannot apply change set to document due to length mismatch'
     pos = 0
     joined_inserts = ('\n'.join(ins) for ins in inserted)
-    for old_len, new_len, ins in zip_longest(sections[::2], sections[1::2], joined_inserts, fillvalue=''):
+    for section in zip_longest(sections[::2], sections[1::2], joined_inserts, fillvalue=''):
+        old_len, new_len, ins = cast(Tuple[int, int, str], section)
         if new_len >= 0:
             doc = doc[:pos] + ins + doc[pos + old_len:]
         pos += old_len
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -1,5 +1,6 @@
+from itertools import zip_longest
 from pathlib import Path
-from typing import Callable, List, Literal, Optional, get_args
+from typing import List, Literal, Optional, get_args
 
 from nicegui.elements.mixins.disableable_element import DisableableElement
 from nicegui.elements.mixins.value_element import ValueElement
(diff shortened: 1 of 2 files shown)


evnchn:
> I think user input, which is processed in _event_args_to_value, usually consists of one section only and we don't need the more complex algorithm to process consecutive changes at once.

CodeMirror is a black-box which I don't comprehend, but I think it may still be beneficial if we process multiple changes, _in the rare case that if we have multiple changes_ we can process them all. 

Still, I understand your desire for clean code. Perhaps we can get rid of the nested while loop, by defining the inner while loop's functionality into a function (which we will shortly modify for the UTF-16 compatibility)

Personally don't want to "take away" functionality in a PR unless we are absolutely sure, which we aren't, I'm afraid.

falkoschindler:
@evnchn I don't think this PR "takes away" functionality. It just removes a potential performance improvement, which - as far as I can tell - is irrelevant for our application.

Here is an explanation from Cursor (referring to the old implementation), which shows how simply these change sets actually are:

> The `_iter_changes` function processes a `ChangeSet` which describes modifications to a document. The `changeset.sections` list encodes these modifications in pairs of integers: `(length_in_old_doc, length_in_new_doc_or_-1)`.
> 
> - A pair like `(length, -1)` means a section of `length` characters remains unchanged.
> - A pair like `(old_length, new_length)` (where `new_length >= 0`) signifies a change (insertion, deletion, or replacement).
> 
> The outer `while` loop iterates through these section pairs.
> 
> The inner `while True` loop specifically handles sequences of *consecutive* changes. It accumulates the changes from adjacent modified sections until it encounters:
> 1.  An unchanged section (`changeset.sections[i + 1] < 0`).
> 2.  The end of the `changeset.sections` list.
> 3.  The `individual` flag is set (though in the `apply` method call, `individual` is `False`).
> 
> By grouping consecutive changes, the inner loop calculates the total span of the modification in the original document (`posA` to `endA`) and the resulting span in the new document (`posB` to `endB`), along with concatenating all the inserted text (`text`) for that entire block.
> 
> After the inner loop breaks, the collected change (represented by `posA`, `endA`, `posB`, `endB`, and `text`) is applied to the document in a single operation using the `func`. This approach avoids applying many small, adjacent changes individually, potentially improving efficiency.

My implementation captures the essence, without the need of an extra class, an iterator, function arguments, and multiple smaller functions calling each other. Hopefully it makes the algorithm more comprehensible. And the additional pytest should make sure that we don't introduce regressions in the future.

evnchn:
It hurts to look at the old code, but I have convinced myself that your code always does the same thing as the original code, minus: 

* `individual = True` which was never used anyways (I think that was used internally by CodeMirror to step through changes or something, since the internal ChangeSet of CodeMirror can be very long)
* Slight speed optimization when applying several consecutive changes, as Cursor mentioned (no more accumulation of `text`), which empirically we have observed CodeMirror does not send stuff back that way.

Should we want to pursue point 2, I think (1) leave it later, since consecutive changes don't occur and (2) we can potentially pre-process the pair `sections: List[int], inserted: List[List[str]]` before entering the for-loop logic, keeping the for-loop concise. 

Advice: Merge if tests pass. 

---

Notably: 

Pair-wise step through of `sections`: 

Old: 
```py
                len_ = changeset.sections[i]
                i += 1
                ins = changeset.sections[i]
                i += 1
```
New: 
```py
for section in zip_longest(sections[::2], sections[1::2], joined_inserts, fillvalue=''):
```

Establish the `inserted`:
Old: `self.inserted: List[str] = ['\n'.join(ins) for ins in inserted]`
New: `joined_inserts = ('\n'.join(ins) for ins in inserted)`

**`posA` and `posB` change in sync, if you make the `while True` logic never occur (don't join changes into one)**

evnchn:
PS: I figure out some additional details, on why it is so brain-twisting to read the code: 

Old code has `self.inserted` but your new code have `joined_inserts = ('\n'.join(ins) for ins in inserted)`. Though the look different, `joined_inserts` == `self.inserted` == `inserted` (in `_iter_changes`) in all regards. 

"second number" `ins` (old code) and `new_len` is just flat-out weird. It is used more like a boolean value, where if it's -1 we keep that segment of original, and otherwise we always grab from `joined_inserts` without verifying whether it is really "the length of the replacement content otherwise".

falkoschindler:
Yes, I used the term "joined inserts" because I'm pre-processing the `inserted` list and didn't want to re-use the same variable which would cause a type conflict. The old code used a member `self.inserted` to hold the pre-processed ("joined") lists.

Interesting observation that we only use the sign of `new_len`. The whole datastructure of a change set seems pretty suboptimal. But I'm sure there are reasons.

PR #4678, opened 2025-04-28
fix `ui.codemirror` on JS UTF-16-based indexing (memoize JS length per char + bisect) 

### LINKED ISSUE #4575: `ui.codemirror` does not apply changes with special characters like 🔄

evnchn:
This PR is in response to #4676 being merged, implementing the approach at https://github.com/zauberzeug/nicegui/pull/4578#issuecomment-2816803235 to fix #4575 

Recap of the technique: 

> Maintain a list `self._cumulative_js_length`, same length as the Python string, which stores the number of UTF-16 code points per character in the Python string. 
> 
> Do a binary search to find the starting index to slice the Python string into. Do another binary search to find the ending index to slice the Python string into as well.
> 
> Don't forget to recompute `self._cumulative_js_length` on change of `self.value` if it has been mutated by others (obvious by the fact that the new value is different to `self._cumulative_corresponds_to_string`

Changes in the code:
* `import bisect` (duh)
* Do not attach the `on_change` via `super().__init__(value=value, on_value_change=on_change)`. We want our internal handler `self._update_cumulative` to always go first. Attach the `on_change` via `super().on_value_change(on_change)`
* Move `_apply_change_set` inside the class (since it is no longer in a vacuum but relies on several instance variables)
* Helper function `get_cumulative_js_length` is introduced, returning `[1, 2, 4, 5, 6]` if passed "aa🔄bb"
* Inner function `find_python_index_given_js_index`: consider "aa🔄bb", say we want to find where to slice, if JS-side calls for `doc[:4]` in its minds. `bisect.bisect_right([1,2,4,5,6],4)` yields 3, meaning we should in-return do `doc[:3]` to mirror JS-side intention. 
* Helper function `get_total_length_from_cumulative` is introduced. Would return `0` if input `[]`, `5` if input `[1, 2, 3, 4, 5]`

---

Remaining tasks:

- [x] avoid counting 1s twice?
- [x] update docstrings
- [x] ~~use separate event for emitting change sets, avoiding the need for selectively blocking `_update_codepoints`~~ it was easier to stick with the "update:model-value" event but to use `_send_update_on_value_change` for an early exit
- [x] check variable naming

### DIFF
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -284,7 +284,7 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
         """
         super().__init__(value=value, on_value_change=self._update_cumulative)
         self._cumulative_corresponds_to_string = value
-        self._cumulative_js_length = []
+        self._cumulative_js_length: List[int] = []
         self._update_cumulative(forced=True)
         if on_change is not None:
             super().on_value_change(on_change)
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -283,7 +283,6 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
         """
         super().__init__(value=value, on_value_change=self._update_codepoints)
         self._codepoints: bytes = b''
-        self._codepoints_represents: str = ''
         self._update_codepoints()
         if on_change is not None:
             super().on_value_change(on_change)
@@ -345,18 +344,13 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
         return b''.join(b'\0\1' if ord(c) > 0xFFFF else b'\1' for c in doc)
 
     def _update_codepoints(self) -> None:
-        """Update `self._codepoints` and `self._codepoints_represents`.
+        """Update `self._codepoints` as a concatenation of "1" for code points <=0xFFFF and "01" for code points >0xFFFF.
 
-        `self._codepoints` are a concatenation of '1' for code points <=0xFFFF and '01' for code points >0xFFFF.
         This captures how many Unicode code points are encoded by each UTF-16 code unit.
         This is used to convert JavaScript string indices to Python by summing `self._codepoints` up to the JavaScript index.
-
-        `self._codepoints_represents` is the string which matches `self._codepoints`.
-        This is used to check if the value has changed and to avoid unnecessary updates.
         """
-        if self.value == self._codepoints_represents:
-            return
-        self._codepoints_represents = self.value
+        if not self._send_update_on_value_change:
+            return  # the update is triggered by the user and codepoints are updated incrementally
         self._codepoints = self._encode_codepoints(self.value or '')
 
     def _apply_change_set(self, sections: List[int], inserted: List[List[str]]) -> str:
@@ -377,5 +371,4 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
                 parts.append(ins)
                 parts_codepoint.append(self._encode_codepoints(ins))
         self._codepoints = b''.join(parts_codepoint)
-        self._codepoints_represents = ''.join(parts)
-        return self._codepoints_represents
+        return ''.join(parts)
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -351,7 +351,11 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
         doc = self.value
 
         def find_python_index(js_index: int) -> int:
-            return bisect.bisect_right(self._cumulative_js_length, js_index)
+            if js_index == 0:
+                return 0
+            lo = max(0, len(self._cumulative_js_length) - (get_total_js_length(self._cumulative_js_length) - js_index))
+            hi = min(js_index, len(self._cumulative_js_length))
+            return bisect.bisect_right(self._cumulative_js_length, js_index, lo, hi)
         assert sum(sections[::2]) == get_total_js_length(
             self._cumulative_js_length), 'Cannot apply change set to document due to length mismatch'
         pos = 0
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -389,10 +389,10 @@ def get_total_js_length(cumulative_js_length: List[int]) -> int:
 def find_python_index(js_index: int, cumulative_js_length: List[int]) -> int:
     if js_index == 0:
         return 0
-    lo = max(0, len(cumulative_js_length) - (get_total_js_length(cumulative_js_length) - js_index))
-    hi = min(js_index, len(cumulative_js_length))
+    lo1 = len(cumulative_js_length) - (get_total_js_length(cumulative_js_length) - js_index)
+    hi1 = js_index
     lo2 = (js_index + 1) // 2
-    hi2 = min(js_index, len(cumulative_js_length) - (get_total_js_length(cumulative_js_length) - js_index + 1) // 2)
-    lo = max(lo, lo2)
-    hi = min(hi, hi2)
+    hi2 = len(cumulative_js_length) - (get_total_js_length(cumulative_js_length) - js_index + 1) // 2
+    lo = max(lo1, lo2, 0)
+    hi = min(hi1, hi2, len(cumulative_js_length))
     return bisect.bisect_right(cumulative_js_length, js_index, lo, hi)
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -1,5 +1,6 @@
+from itertools import zip_longest
 from pathlib import Path
-from typing import Callable, List, Literal, Optional, get_args
+from typing import List, Literal, Optional, Tuple, cast, get_args
 
 from nicegui.elements.mixins.disableable_element import DisableableElement
 from nicegui.elements.mixins.value_element import ValueElement
@@ -332,80 +333,17 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
 
     def _event_args_to_value(self, e: GenericEventArguments) -> str:
         """The event contains a change set which is applied to the current value."""
-        changeset = _ChangeSet(sections=e.args['sections'], inserted=e.args['inserted'])
-        new_value = changeset.apply(self.value)
-        return new_value
+        return _apply_change_set(self.value, e.args['sections'], e.args['inserted'])
 
 
-# Below is a Python implementation of relevant parts of https://github.com/codemirror/state/blob/main/src/change.ts
-# to apply a ChangeSet to a text document.
-
-
-class _ChangeSet:
-    """A change set represents a group of modifications to a document."""
-
-    def __init__(self, sections: List[int], inserted: List[List[str]]) -> None:
-        # From https://github.com/codemirror/state/blob/main/src/change.ts#L21:
-        # Sections are encoded as pairs of integers. The first is the
-        # length in the current document, and the second is -1 for
-        # unaffected sections, and the length of the replacement content
-        # otherwise. So an insertion would be (0, n>0), a deletion (n>0,
-        # 0), and a replacement two positive numbers.
-        self.sections: List[int] = sections
-        self.inserted: List[str] = ['\n'.join(ins) for ins in inserted]
-
-    def length(self) -> int:
-        """Calculate the length of the document before the change."""
-        return sum(self.sections[::2])
-
-    def apply(self, doc: str) -> str:
-        """Apply the changes to a document, returning the modified document."""
-        if self.length() != len(doc):
-            raise ValueError('Cannot apply change set to a document with the wrong length')
-        return _iter_changes(self, doc, _replacement_func, individual=False)
-
-    def __str__(self) -> str:
-        return f'ChangeSet(sections={self.sections}, inserted={self.inserted})'
-
-
-def _iter_changes(
-    changeset: _ChangeSet, doc: str, func: Callable[[str, int, int, int, int, str], str], individual: bool
-) -> str:
-    inserted = changeset.inserted
-    posA, posB, i = 0, 0, 0
-
-    while i < len(changeset.sections):
-        len_ = changeset.sections[i]
-        i += 1
-        ins = changeset.sections[i]
-        i += 1
-
-        if ins < 0:
-            posA += len_
-            posB += len_
-        else:
-            endA, endB = posA, posB
-            text = ''
-            while True:
-                endA += len_
-                endB += ins
-                if ins and inserted:
-                    text = text + inserted[(i - 2) // 2]
-                if individual or i == len(changeset.sections) or changeset.sections[i + 1] < 0:
-                    break
-                len_ = changeset.sections[i]
-                i += 1
-                ins = changeset.sections[i]
-                i += 1
-            doc = func(doc, posA, endA, posB, endB, text)
-            posA, posB = endA, endB
-
+def _apply_change_set(doc, sections: List[int], inserted: List[List[str]]) -> str:
+    # based on https://github.com/codemirror/state/blob/main/src/change.ts
+    assert sum(sections[::2]) == len(doc), 'Cannot apply change set to document due to length mismatch'
+    pos = 0
+    joined_inserts = ('\n'.join(ins) for ins in inserted)
+    for section in zip_longest(sections[::2], sections[1::2], joined_inserts, fillvalue=''):
+        old_len, new_len, ins = cast(Tuple[int, int, str], section)
+        if new_len >= 0:
+            doc = doc[:pos] + ins + doc[pos + old_len:]
+        pos += old_len
     return doc
-
-
-def _replace_range(doc: str, from_: int, to: int, new: str) -> str:
-    return doc[:from_] + new + doc[to:]
-
-
-def _replacement_func(doc: str, from_a: int, to_a: int, from_b: int, _to_b: int, text: str) -> str:
-    return _replace_range(doc, from_b, from_b + (to_a - from_a), text)
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -350,26 +350,26 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
         # based on https://github.com/codemirror/state/blob/main/src/change.ts
         doc = self.value
 
-        def find_python_index_given_js_index(js_index: int) -> int:
+        def find_python_index(js_index: int) -> int:
             return bisect.bisect_right(self._cumulative_js_length, js_index)
-        assert sum(sections[::2]) == get_total_length_from_cumulative(
+        assert sum(sections[::2]) == get_total_js_length(
             self._cumulative_js_length), 'Cannot apply change set to document due to length mismatch'
         pos = 0
         joined_inserts = ('\n'.join(ins) for ins in inserted)
         for section in zip_longest(sections[::2], sections[1::2], joined_inserts, fillvalue=''):
             old_len, new_len, ins = cast(Tuple[int, int, str], section)
             if new_len >= 0:
-                first_part_index = find_python_index_given_js_index(pos)
-                second_part_index = find_python_index_given_js_index(pos + old_len)
-                doc = doc[:first_part_index] + ins + doc[second_part_index:]
+                first_index = find_python_index(pos)
+                second_index = find_python_index(pos + old_len)
+                doc = doc[:first_index] + ins + doc[second_index:]
 
                 ins_cumulative_js_length = get_cumulative_js_length(ins)
-                first_part_cumulative_js_length = self._cumulative_js_length[:first_part_index]
+                first_part_cumulative_js_length = self._cumulative_js_length[:first_index]
 
                 self._cumulative_js_length = first_part_cumulative_js_length + \
-                    [x + get_total_length_from_cumulative(first_part_cumulative_js_length) for x in ins_cumulative_js_length] + \
-                    [x + get_total_length_from_cumulative(ins_cumulative_js_length)
-                     for x in self._cumulative_js_length[second_part_index:]]
+                    [x + get_total_js_length(first_part_cumulative_js_length) for x in ins_cumulative_js_length] + \
+                    [x + get_total_js_length(ins_cumulative_js_length)
+                     for x in self._cumulative_js_length[second_index:]]
             pos += old_len
             self._cumulative_corresponds_to_string = doc
         return doc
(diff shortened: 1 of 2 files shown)


evnchn:
@falkoschindler sorry but I accidentally pressed request review from Copilot. 

It's not counting towards quota before May 8, so it'd be fine? https://docs.github.com/en/copilot/using-github-copilot/code-review/using-copilot-code-review#code-review-monthly-quota

How much quota do I / we (idk which entity the quota counts towards) have?

evnchn:
Sorry but as it turns out the old insertion logic is one buggy nightmare. 

It was revealed when I want to use Windows Emoji keyboard to insert the emoji 🔄️

Actually, it doesn't just insert 1 character, but 2 (don't ask me why)

{C3F3F7D0-8A8D-483B-BEA3-EA059E7FC3B3}

Which reveals that my old logic of preserving the cumulative have some issues. 

---

Now, the logic goes more like:

* First chunk, write straight to self._cumulative_js_length
* Any chunk onwards follows the logic of `self._cumulative_js_length += [x + get_total_js_length(self._cumulative_js_length) for x in NEW_CUMULATIVE]` where `NEW_CUMULATIVE` starts counting all on its own. 
* Particularly for tacking on the old previous chunk, we got to be careful! It is NOT safe to assume the first character has JS-length of 1 and work from there. Instead, go backwards 1 character and get the cumulative up till then (or 0 is can't go backward) and subtract from it. 

Example: 

`[1,2,3,5,6,7,8]`, I want to tack-on the last 4 characters. 

My target segment would be `[5,6,7,8]`, and I subtract `3` from each, getting `[2,3,4,5]`, which I can safely tack on via the `self._cumulative_js_length += [x + get_total_js_length(self._cumulative_js_length) for x in NEW_CUMULATIVE]` approach.

evnchn:
@falkoschindler Is something not quite right with the test framework: 

`TypeError: 'NoneType' object is not iterable`

https://github.com/evnchn/nicegui/actions/runs/14766138486/job/41457866769

But it never should be `None`. Value is supposed to be a `str`

Here I patch around it by making it work in case someone did something like `mycodemirror.set_value(None)`, but maybe we can fix the test some way too?

evnchn:
Hey @falkoschindler. I just want to clarify that this PR is officially working. 

There are test cases, and the tests are passing. 

https://github.com/evnchn/nicegui/actions/runs/14774823858

evnchn:
If you want to get your head into this, you must read, **very carefully,** https://en.wikipedia.org/wiki/UTF-16 and https://en.wikipedia.org/wiki/Unicode#UTF

Be careful about the difference in "code point" (the single number represent the character) and "code units" (a single unit of representation of 2 bytes in UTF-16 which either can mean a code point, or a pair stuck together to represent code point beyond U+010000)

Note that JS `.length` counts UTF-16 code units, while Python `len()` counts how many characters / how many UTF-8 code point (it's 1:1 for UTF-8)

Note also the strategy to quickly know what to slice into in Python-land to match the JS-land index is by means of a Prefix Sum Array https://en.wikipedia.org/wiki/Prefix_sum and binary search https://en.wikipedia.org/wiki/Binary_search

The binary search is sped up by two sets of upper bound and lower bound, which eliminates some numbers from the search by leveraging the fact each character can take 1 or 2 code units, no third possibility.

falkoschindler:
@evnchn Before looking into your code, I still have a conceptual question:

I read about the two encodings. UTF-8 uses 1 to 4 bytes depending on the character (or code point), and UTF-16 uses 2 or 4 bytes. This made me think: Shouldn't we have more compatibility problems when interpreting change sets from JavaScript in Python?

Here's a quick test:
```py
for encoding in ['utf-8', 'utf-16']:
    for c in 'AÄ🙂':
        print(f'"{c}" in {encoding}: {[int(b) for b in c.encode(encoding)]}')
```
Output:
```
"A" in utf-8: [65]
"Ä" in utf-8: [195, 132]
"🙂" in utf-8: [240, 159, 153, 130]
"A" in utf-16: [255, 254, 65, 0]
"Ä" in utf-16: [255, 254, 196, 0]
"🙂" in utf-16: [255, 254, 61, 216, 66, 222]
```

It seems like UTF-16 always prepends the BOM (Byte Order Mark) [255, 254] at the beginning of a string, so we shouldn't count it as part of the character encoding. But even then we have
- 1 vs. 2 bytes for "A",
- 2 vs. 2 bytes for "Ä", and
- 4 vs. 4  bytes for "🙂".

Why does issue #4575 only occur when using emojis? Shouldn't it also break for umlauts like "Ä"? Actually "Ä" and "🙂" are the characters with matching byte counts, so the issue should only occur for ASCII characters like "A". I'm confused. It looks like there is already some compensation going on. Or I'm missing something obvious.

evnchn:
@falkoschindler I'm glad you are still into the issue. 

TL-DR: Python and JS only disagree for U+10000 and above (`ord(x) >= 65536`), since it's always 1 Unicode Code Point for Python regardless of how many UTF-8 bytes underneath, but it is 2 UTF-16 Code Units for JS regardless of it's representing 1 Unicode Code Point with 2 UTF-16 Code Units. 

---

First off, the obvious: Use 'utf-16le' or 'utf-16be' so that the BOM (`255, 254`) doesn't show up. Will do so for the rest of the discussion below. 

---


As a recap to how Python and JS counts length: 

- Python `len()`: Counts the number of Unicode **Code Points**, where each can be represented by UTF-8 **single-byte characters** / **multi-byte sequences**, which can take 1 / 2-4 bytes. 
  - A higher-level "Unicode Code Point First" approach, which ensures the number of Unicode code points, but the number of bytes underneath is varying
- JS `.length`: Counts the number of UTF-16 **Code Units**, which is always 2 bytes. It requires 1 **Code Unit** for any Unicode **Code Points** in the Basic Multilingual Plane (U+0000 - U+FFFF), but may require 2 **Code Units** for anything Unicode **Code Points** beyond U+10000. 
  - A more bare-metal "bytes first" approach, which ensures number of bytes underneath is varying, but the number of Unicode code points which is represented is varying. 

By you counting bytes in Python-land, I have every reason to suspect that you have misunderstood the Python `len()` calculation. **Check your understanding very carefully.** 

---

**Below, what is bracketed in {} has a length of 1 in their respective languages.** 

### Ä

Python: `{195, 132}` => Length = 1
JavaScript: `{196, 0}` => Length = 1

### 🙂

Python: `{240, 159, 153, 130}` => Length = 1
JavaScript: `{61, 216}`, `{66, 222}` => Length = 2

---

We can also work backwards, and figure out a discriminator between the cases, which divides the working and the non-working: 

```
"A" in utf-8: [65]  # working
"Ä" in utf-8: [195, 132] # working
"🙂" in utf-8: [240, 159, 153, 130] # NON-WORKING
"A" in utf-16: [65, 0] # working
"Ä" in utf-16: [196, 0] # working
"🙂" in utf-16: [61, 216, 66, 222] # NON-WORKING
```

Since Ä works equally as well as A, the number of bytes in UTF-8 can't possible be the cause of the issue. 

We come to the conclusion that `number of bytes of a character in UTF-16 != 2` => NON-WORKING. 

---

Extra reading materials: 

1. The bombshell quote from MDN:

> The length data property of a [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) value contains the length of the string in **UTF-16 code units**.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length

2. The bombshell quote from Python:

> Strings are immutable [sequences](https://docs.python.org/3/library/stdtypes.html#typesseq) of **Unicode code points**.

https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str

3. Looks almost exactly like my initial solution:

{3C09CEB9-11C6-4451-90E3-1B379B67B97B}


https://youtu.be/R6HN92RCTJs?si=JMMUsEJ0HCSNne7O

4. This examples shows the 2 code units per emoji situation very well


{FA74507A-93DD-4E85-A118-664F4C01FFF2}

https://youtu.be/LuFSIEGKmBg?si=wDn7mAcCpur6tvk5

5. Best StackOverflow question on the issue so far

https://stackoverflow.com/q/38345372

6. Best article on the issue so far (if not being slightly out of scope)

https://cestoliv.com/blog/how-to-count-emojis-with-javascript/en (Please ignore everything about the emojis being joined with Zero Width Joiner. That's out of scope for this discussion because it has to do with character limits (think: tweets) 

7. Decent article on the issue, but at least it's not out of scope 

https://blog.jonnew.com/posts/poo-dot-length-equals-two

falkoschindler:
Oh wow, reading your response after waking up this morning was like 🤦🏻‍♂️. I think I finally got it! It's not only the number of bytes that matter, but that Python is counting code points and JavaScript is counting code units, which only differs for U+10000 and above.

falkoschindler:
Ok, I've started to review the code, but I'm still analyzing and learning what is going on.

A first result is that the synchronization still can be broken. Consider the following setup:
```py
ui.codemirror('Hello world 🙂\nHello world 🙂')
```
Now select one smiley, press Cmd-D (or the corresponding Windows equivalent) to get a second cursor, and finally delete both smileys at once. This causes the good old "AssertionError: Cannot apply change set to document due to length mismatch".

@evnchn What to you think, should I continue reviewing the code so that I can hopefully help fixing the algorithm? Or do you want to take over and let me know when you're done (or need help)?

evnchn:
@falkoschindler Sorry. As you may know, I need to prepare for exam, which is on May 19-28. 

Therefore, let's divide the responsibilities: 

- Issues pertaining to #4701: Can you fix it yourself? 
  - It is because #4676 really hurts my brain to think about. 
- Issue pertaining to single-edit scenario in this PR: Tell me if you find them, and I'll fix it. Though I'm quite confident that won't happen easily since I tested a lot. 
- Issue pertaining to code quality: Can you fix it yourself? There's no use in finding me at all

evnchn:
Given how #4702 was implemented, means this PR has to go back to square-one (or square-two, maybe. Still a lot has to be re-done.)

falkoschindler:
Sorry for the trouble. But let me take over for a bit. I don't think there's a dramatic difference and - to be honest - after PR #4702 the code feels even more natural. I'll study your implementation anyway, so I can fix it while being at it.

falkoschindler:
Oh dear, this was challenging! I'll spare you the details, but it was quite a ride trying to get this working either by fixing the server or doing the conversion on the client. Now I finally have something which I'm pretty pleased with.

My main insight: It looks like bisecting doesn't help much because we have to update each individual cumulative sum every time - or at least half of them on average. And while not being very effective, keeping these cumulative sums updated complicates the code so much. There's a similar remark on [`bisect.insort`](https://docs.python.org/3/library/bisect.html#performance-notes):
> The insort() functions are O(n) because the logarithmic search step is dominated by the linear time insertion step.

Therefore I thought about a simpler lookup table (LUT) which is easier to update and easier to use. The result is `self._emojies` (I like to call characters >=0x10000 "emojies" for simplicity) which is of type `bytes` and contains a single byte 0 for "normal" characters and two bytes (0, 1) for emojies. This means you can take a JavaScript string index `pos` and get the _offset_ to the corresponding Python string index via `sum(self._emojies[:pos])`.

Now the update algorithm is as follows. For each section, reuse a part from the old `doc` if the section didn't change (count emojies to convert JavaScript positions to Python string indices), or use `ins` sent from the client. While at it, build a new version of the `emojies` LUT. _After_ iterating over all sections, we can replace `self._emojies` and return the joined document `parts`.
Note that the `inserted` list is sometimes shorter than the number of sections, which is why we extend them with an infinite `repeat('')` iterator. This allows us to use `zip` instead if `zip_longest`, which avoids typing issues.

I tried to keep performance in mind. I use iterators where possible to avoid instantiating many lists. And if you look closely, we only sum over `self._emojies` once (in parts). I think this is optimal because even bisecting would require to keep a cumulative LUT up to date, causing the complexity.

I also added a few more test cases which are passing. Testing manually also feels nice.

What do you think, @evnchn?

evnchn:
@falkoschindler 

**First off: I'm assuming you are not working from #4702, because you "capture the spirit" of that PR**

Even though we can absolutely use something like `blist` to achieve O(log n) insert, I think we are bottlenecked by O(n) insertion time in the _string itself_. Oh well... 

Then, it took me quite a while to figure out the `0` and `01` thing in `self._emojis`. I think you calling it a LUT is underselling it. You are, in fact, creating a **faithful representation of how many Unicode characters beyond U+10000 are encoded by each code Unit**. 

See, you can consider each byte in the byte string as a one-to-one corrspondance to the underlying code units that JavaScript would have used. 

- `0` captures that those are up to U+FFFF
- `01` captures that (1) it is U+10000 and beyond, and (2) the high surrogate does nothing on itself, and only when the low surrogate is also there, it represents a Unicode code point together. 

With that, I am as always impressed by your ability to find simple solutions 🤯

---

But I think we can do better with an even simplier solution: 

Check out my commit, which creates a **faithful representation of how many Unicode characters (in general) are encoded by each code Unit**. 

You need **one** change only

- **`1` captures that up to U+FFFF, a single code unit represents a Unicode code point.** 
- `01` captures that U+10000 and beyond, the high surrogate does nothing on itself, and only when the low surrogate is also there, it represents a Unicode code point together. 

Then, we can get rid of `pos - old_len - emojies_before_start: pos - emojies_before_end`. 

---

Also: 

- I did a speedtest. Apparently it is faster to count explicitly for 1s instead of summing the byte string. 
- Remember to block off calls to `_update_codepoints` if the underlying string matches. Otherwise we are running that function again and again, and your effort of maintaining codepoints via `self._codepoints = b''.join(parts_codepoint)` is in vain. 

Open questions:

- Variable naming? (I am not good at that)
- Does it actually work (Should, but pytest on Windows is really flakey)

falkoschindler:
> With that, I am as always impressed by your ability to find simple solutions

I was sooo close to failing on this one 🥵

> But I think we can do better with an even simplier solution

Nice!! The code is getting really pretty!

> I did a speedtest. Apparently it is faster to count explicitly for 1s instead of summing the byte string.

Interesting! I chose bytes especially because we can `join` and `sum` them easily. But it's still a good choice I guess.
Anyway, it took me a while to reuse the `emojies_before_start` to compute `emojies_before_end` without re-counting most emojis. But you're indeed re-counting. Sure that this is faster? I'd expect it to take approximately twice the time.

> Remember to block off calls to _update_codepoints if the underlying string matches.

I intentionally did not block, because the change handler is only called if the string changes, right? But we better check this assumption.

> Variable naming?

Yes, let's review variable names once the implementation is settled.

> Does it actually work

Tests are still green. 👍🏻

One more thing: We should update the docstrings to match the new naming and LUT content.

evnchn:
> I intentionally did not block, because the change handler is only called if the string changes, right? But we better check this assumption.

No. You can try not blocking it off, and add a print statement. Every single time you type anything, you recompute `_update_codepoints`, which loses the point of maintaining it via `self._codepoints = b''.join(parts_codepoint)` entirely. 

I always resort to print debugging for this kind of stuff, because computer brain >>> my brain. 

> Sure that this is faster? I'd expect it to take approximately twice the time.

There are 2 separate issues: 

- Use `.count(1)` in place of `sum()`: indeed faster. I'll attach the test script. 
- Counting the part `self._codepoints[:pos-old_len]` twice: fair point. Will see how to fix. 


Test script from chatgpt: 

```py
import timeit
import random

# Create a byte string of length 100,000 with 50/50 distribution of 0s and 1s
byte_string = b'011110001010100010' * 6250  # Repeat to reach the desired length

# Function to sum bits
def sum_bits(s):
    return sum(s)

# Function to count '1's
def count_ones(s):
    return s.count(1)

# Randomly slice the byte string
def random_slice_and_sum():
    index = random.randint(0, len(byte_string) - 1)
    sliced = byte_string[:index]
    return sum_bits(sliced)

def random_slice_and_count():
    index = random.randint(0, len(byte_string) - 1)
    sliced = byte_string[:index]
    return count_ones(sliced)

# Timing the functions
sum_time = timeit.timeit(random_slice_and_sum, number=1000)
count_time = timeit.timeit(random_slice_and_count, number=1000)

print(f"Time taken to sum bits: {sum_time:.5f} seconds")
print(f"Time taken to count '1's: {count_time:.5f} seconds")
```

```
===== RESTART: C:/Users/User/AppData/Local/Programs/Python/Python312/vs.py =====
Time taken to sum bits: 0.21321 seconds
Time taken to count '1's: 0.03040 seconds

===== RESTART: C:/Users/User/AppData/Local/Programs/Python/Python312/vs.py =====
Time taken to sum bits: 0.19715 seconds
Time taken to count '1's: 0.03097 seconds

===== RESTART: C:/Users/User/AppData/Local/Programs/Python/Python312/vs.py =====
Time taken to sum bits: 0.18572 seconds
Time taken to count '1's: 0.03005 seconds

===== RESTART: C:/Users/User/AppData/Local/Programs/Python/Python312/vs.py =====
Time taken to sum bits: 0.18661 seconds
Time taken to count '1's: 0.02956 seconds
```

evnchn:
About the check in `_update_codepoints`:

{8F423050-B9FC-4714-8A80-229D50D71124}

image

Pressed my key 7 times in the codemirror.

falkoschindler:
> Every single time you type anything, you recompute `_update_codepoints`

Oh, I didn't consider that user interaction also triggers a change event.
But keeping basically another copy of `value` is a pity. Yet another state item of size O(n).

This reminds me of another idea I had today: Maybe we don't use the "update:value" event to emit changes to the client, but a custom "change_set" or whatever. Then it doesn't trigger the change handler, but can be handled separately before setting the new `value`.

> There are 2 separate issues:

Exactly!

evnchn:
> But keeping basically another copy of `value` is a pity. Yet another state item of size O(n).

You could `hash()` the string, and check the `hash()` instead? But not sure about hash collisions and stuff.

falkoschindler:
> You could hash() the string, and check the hash() instead?

Another O(n) operation looking at every character... I'll see if we can separate the events. Then we don't need this check at all.

I just added a task list to the description of this PR to keep an overview of the remaining tasks.

evnchn:
> Maybe we don't use the "update:value" event to emit changes to the client, but a custom "change_set" or whatever.

Sorry but I am really weak in that aspect and the code of ValueElement always pains me when I read it. Can you please handle it?

I think:

- Spend reasonable time to come up with a more elegant solution to avoid duplicated `_update_codepoints`. 
- If you can't figure it out before 2.17.0 rolls around, ship whatever we got now, since although it's 2x the memory use, at least it doesn't break on UTF-16. 

---

Also: I don't get why `_send_update_on_value_change` can be True, False, or None. 

- `if self._send_update_on_value_change:`: True only if it is `True`
- `self._send_update_on_value_change = self.LOOPBACK is True`: True only if it is `True`

So `False` and `None` totally feels identical to me, not sure the distinction.

evnchn:
https://github.com/zauberzeug/nicegui/pull/4678/commits/4dfc1156d26812a0c56b964892538fa3e8340b07: Docstring
https://github.com/zauberzeug/nicegui/pull/4678/commits/af4f56245dfabc12fd79056c2afd606d3d3e2643:  avoid counting 1s twice

falkoschindler:
> Also: I don't get why _send_update_on_value_change can be True, False, or None.

The attribute `_send_update_on_value_change` controls whether a value change should be sent as an update to the client. It is mainly controlled by the static `LOOPBACK` attribute which is defined per value element class. Referring to the docstring, there are three cases:

- ``True``: The value is updated by sending a change event to the server which responds with an update.
    This is the default. An example is `ui.checkbox`. The advantage of a loopback is that if there is a connection problem, the box won't change its appearance because a message needs to the server and back to actually trigger the state change.

- ``False``: The value is updated by setting the VALUE_PROP directly on the client.
    This is the second most common case. An example is `ui.input`. Typing in an input field would feel laggy and can cause glitches if we would send the whole text back and forth after each key press. Instead we update the "model-value" prop on the client without waiting for an update from the server. This happens generically in the event handling code in nicegui.js.

- ``None``: The value is updated automatically by the Vue element.
    This is for elements that don't expect a response from the server (similar to `ui.input`) but also don't need nicegui.js to update the "model-value" prop. An example is `ui.codemirror`.

But `_send_update_on_value_change` is not always equal to `LOOPBACK`, but changes its value in certain situations. Initially it is `True` so that changing the `value` programmatically triggers and update for the client. But if a change handler is triggered by a _client event_, we set `_send_update_on_value_change = LOOPBACK`, call `self.set_value(...)`, and finally set `_send_update_on_value_change = True` again.

---

In my latest commit I'm using this very mechanism to selectively run `_update_codepoints` for server-induced changes only. When the user is typing, `_send_update_on_value_change` will be set to `LOOPBACK` (which is `None`) and we exit early from `_update_codepoints`.

falkoschindler (approved):
Ok, from my point of view this PR is ready to merge! 🙂

Note about naming: I changed `parts` and `parts_codepoint` to `document_parts` and `codepoint_parts`. I understand the urge to have them beginning with the same word "parts" to increase visual symmetry of both lines. But in the long term I think it's better to use names that work well as part of natural language. You wouldn't say "We need to join the parts codepoint", but "We need to join the codepoint parts". Similarly, we have "old lengths" and "new lengths", "end positions" etc.

@evnchn Let me know if you're happy with the result. Then I would squash and merge this PR. 🚀

evnchn:
Awesome @falkoschindler 

Here is my test, by the way: 

```py
@ui.page('/codemirror')
def codemirror_page():
    mycodemirror = ui.codemirror('Hello world 🙂\nHello world 🙂')
    ui.code(language="").bind_content_from(mycodemirror, 'value')
    ui.button("Set to sad message", on_click=lambda: mycodemirror.set_value('So sad emoji 😢'))
    ui.button("Set to happy message", on_click=lambda: mycodemirror.set_value('So happy emoji 🙂'))
```

It does indeed skip the `_update_codepoints` successfully. The check `if not self._send_update_on_value_change:` is concise but works amazingly. Love it.

PR #4679, opened 2025-04-29
Add HTTPS protocol support to `ui.run`

thetableman:
This PR introduces a `protocol` parameter to `ui.run(...)` to support launching the app on a `https` protocol if required.

Uvicorn automatically launches using `https` when the `ssl_certfile` and `ssl_keyfile` parameters are specified. However, this isn't the case with NiceGUI's `ui.run` when passed as kwargs (as per https://nicegui.io/documentation/section_configuration_deployment#server_hosting).

This is confirmed by these two minimal examples.

Uvicorn recognizes the specified certificate and key files and launches on https, you don't need to specify the port as 443.
```python
import uvicorn

async def app(scope, receive, send):
    assert scope['type'] == 'http'

    await send({
        'type': 'http.response.start',
        'status': 200,
        'headers': [
            (b'content-type', b'text/plain'),
            (b'content-length', b'13'),
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })

uvicorn.run(app, port=8000, ssl_certfile='<certfile_path>', ssl_keyfile='<keyfile_path>')
```

NiceGUI passes the certificate and key files to Uvicorn but still launches on a http protocol.
```python
from nicegui import ui, app

ui.label('Hello, world!')

ui.run(ssl_certfile='<certfile_path>', ssl_keyfile='<keyfile_path>')
```

This is because, as the maintainers are aware, NiceGUI handles the app launch rather than using Uvicorn directly.

~I opted for a `protocol` parameter over auto-detection, despite `ui.run` already having a considerable number, to allow for greater control and allow for the fringe cases where certificates are provided but a standard http protocol is required.~

Edit:
I'm aware of #387 and the workaround to launch a NiceGUI app using `uvicorn.run`.


thetableman:
To generate a certificate and key for `localhost`, the following generated by ChatGPT works perfectly (Python 3.10).

```python

from cryptography import x509
from cryptography.x509.oid import NameOID, ExtendedKeyUsageOID
from cryptography.hazmat.primitives import serialization, hashes
from cryptography.hazmat.primitives.asymmetric import rsa
from cryptography.hazmat.primitives.serialization import pkcs12, Encoding, PrivateFormat, NoEncryption
from datetime import datetime, timedelta
import ipaddress

SSL_PATH = '<your file path>'
PFX_PASSWORD = b'P@ssw0rd'
HOST_NAME = 'localhost'
HOST_IP = '127.0.0.1'

# Generate private key
key = rsa.generate_private_key(public_exponent=65537, key_size=2048)

# Certificate subject/issuer
subject = issuer = x509.Name([
    x509.NameAttribute(NameOID.COMMON_NAME, HOST_NAME),
])

# Subject Alternative Names (SAN)
san = x509.SubjectAlternativeName([
    x509.DNSName(HOST_NAME),
    x509.IPAddress(ipaddress.IPv4Address(HOST_IP))
])

# Create the certificate
cert = (
    x509.CertificateBuilder()
    .subject_name(subject)
    .issuer_name(issuer)
    .public_key(key.public_key())
    .serial_number(x509.random_serial_number())
    .not_valid_before(datetime.utcnow())
    .not_valid_after(datetime.utcnow() + timedelta(days=365 * 5))  # 5 years
    .add_extension(san, critical=False)
    .add_extension(
        x509.ExtendedKeyUsage([ExtendedKeyUsageOID.SERVER_AUTH]),
        critical=False
    )
    .sign(private_key=key, algorithm=hashes.SHA256())
)

# Serialize to PFX
pfx_data = pkcs12.serialize_key_and_certificates(
    name=HOST_NAME.encode(),
    key=key,
    cert=cert,
    cas=None,
    encryption_algorithm=serialization.BestAvailableEncryption(PFX_PASSWORD)
)

# Define a base name for the pfx, crt, and key files
base_name = f"{HOST_NAME}-{HOST_IP.split('.')[0]}"

# Save to file
with open(f'{SSL_PATH}{base_name}.pfx', 'wb') as f:
    f.write(pfx_data)

print(f'PFX file created: {base_name}.pfx')

# Load the PFX contents
private_key, cert, additional_certs = pkcs12.load_key_and_certificates(pfx_data, PFX_PASSWORD)

# Write the private key to a .key file
with open(f'{SSL_PATH}{base_name}.key', 'wb') as f:
    f.write(private_key.private_bytes(
        encoding=Encoding.PEM,
        format=PrivateFormat.TraditionalOpenSSL,
        encryption_algorithm=NoEncryption()
    ))

# Write the certificate to a .crt file
with open(f'{SSL_PATH}{base_name}.crt', 'wb') as f:
    f.write(cert.public_bytes(encoding=Encoding.PEM))

print('Extracted .key and .crt files.')
```

thetableman:
@evnchn Yes... but like I said:
> I opted for a protocol parameter over auto-detection, despite ui.run already having a considerable number, to allow for greater control and allow for the fringe cases where certificates are provided but a standard http protocol is required.

Personally I think it does make sense to replicate uvicorn (the ssl parameters determine the protocol) but on this I caved to ChatGPT which advised explicit vs implicit. Also note that the default `None` of the new parameter does result in following uvicorn's logic.

When I enquired about this implicit behavior ChatGPT had the following to say

> Should protocol be determined solely by cert/key presence?
> 
> In general, yes — but with important nuances:
> 
> - If both ssl_certfile and ssl_keyfile are provided and valid, it’s reasonable to assume the user intends to run HTTPS.
> Uvicorn itself will bind to an SSL context if those are provided.
> 
> - However, some users might pass cert/key for other tooling purposes but still intend to handle requests over HTTP (uncommon, but possible) — or might have conditional launch code where ssl_certfile is set but not intended to actually enable HTTPS on a given run.
> 
> So while automatic protocol detection is safe in 95%+ cases, it’s technically an implicit behavior change for anyone relying on the current hardcoded http:// in open_browser() and print_welcome().
> Potential breaking cases:
> 
> - Scripts or tooling that parse/expect the exact protocol string in the console output.
> 
> - Automated tests that mock or capture the console output or the webbrowser.open_new() URL.
> 
> - Environments where cert/key files are supplied as config defaults but HTTPS isn’t actually intended (unlikely, but possible).
> 
> Best practice solution:
> 
> Explicit is better than implicit.

So, food for thought, is uvicorn's approach actually the best way!?

> So it feels like letting people shooting ourselves in the foot, if we let them set manually the protocol.

Totally understand and agree... but, if you don't know what you're doing and the impact then why purposely change the setting from the default which does follow uvicorn's approach.

Honestly I'm happier either way, for my projects the implicit assumption of https when a certificate is provided works for me, but NiceGUI isn't designed for just me and the specifics of my projects. Sometimes its 'nicer' being able to customize a bit more and I always find it irritating when an api limits some of these things (#4217)

Regardless, if the maintainers prefer a purely auto method I'll happily adjust the PR.

thetableman:
> PS: @falkoschindler is https://github.com/zauberzeug/nicegui/pull/4539#pullrequestreview-2799677551

On this, which I completely agree with because it is a bit messy, I had wondered whether `ui.run` could support a config file, something like a toml or yaml? Perhaps add a `ui.run_from_config(...)` to compliment `ui.run_with(...)` to avoid the irony of adding a another parameter to `ui.run(...)` to accept a config.

This would allow for circumstances like my own where I run the app in development, staging, and production environments with varying parameters - you just load up the config file relevant to the environment.

evnchn:
TL-DR: I don't see why you would want to set protocol to HTTP when SSL cert is passed in and recreate the old behaviour, when the old behaviour does nothing good at all. This is why I believe we should not look back and march forward with only automatic HTTPS protocol detection, no adding protocol input to `ui.run`. 

---

Let me respond directly to a couple of points which I believe are factually wrong. 

> fringe cases where certificates are provided but a standard http protocol is required
> pass cert/key for other tooling purposes but still intend to handle requests over HTTP (uncommon, but possible)
> Environments where cert/key files are supplied as config defaults but HTTPS isn’t actually intended (unlikely, but possible)

**Cannot possibly happen.** Nobody would pass in certificates and willingly have their browser go to port 8080 over HTTP because that is for-certain not working. They _could_ have some HTTP service, but that can't possibly bind to the same 8080 port and must be served over say port 8090. At this point our web browser launch logic totally doesn't work for their purpose. So in fact there CANNOT be any existing & working code which relies on the old behaviour of opening the HTTP at 8080, since that would just point the browser to a dead page ALWAYS. 

_P.S.: You may want to switch to DeepSeek R1 or other reasoning models..._

> Scripts or tooling that parse/expect the exact protocol string in the console output.

Not supported use case. We can't be forced to keep our console output consistent just because of that. Look at `apt` on Linux: `WARNING : apt does not have a stable CLI interface. Use with caution in scripts.`. I guess we are just not warning about it but it's obviously true for NiceGUI and any other software which doesn't explicitly let you parse the output. You can't blame Google for changing the HTML on their webpage and breaking your web scraper, can you? 

> Automated tests that mock or capture the console output or the webbrowser.open_new() URL

Same argument as above. We can't be forced to keep how we call `webbrowser.open_new()` consistent just because of that.

> #4217 

Note that in your case, you are arguing that the old behaviour have some value, and the new behaviour changed that. Then it's more nuanced. But, opening the browser to go to HTTP port 8080 while the server serves HTTPS and HTTPS only on port 8080 have **absolutely no use** and thus I don't feel any hesitation on simply changing the behaviour to the new one, since we are strictly adding something and taking away nothing. 

---

Meanwhile: 

> So, food for thought, is uvicorn's approach actually the best way!?

Maybe you can try convincing uvicorn devs.

thetableman:
@evnchn I'm really confused by your response, the reasoning of some of the ideas in this PR aren't intended as a personal attack on you but your response reads, almost aggressively, as though this is the case.

This is simply a PR presenting an idea for consideration to include in the project and I noted in my last comment that I was completely open to an auto https/http option so I don't understand the need for your most recent response.

evnchn:
My apologies as I am a bit unstable right now, but I'll clarify that it is nothing personal, and I can take it both ways.

Sorry for being a bit harsh towards ChatGPT. I begin to not like it but Llama or DeepSeek since it sometimes hallucinates (as you see) in my previous response. 

At the end of the day, we'll have to check out what @falkoschindler have to say. 

To be clear, we are friends, @thetableman

thetableman:
It's all good, we all have our days and ultimately there's no hard feelings. The end goal for both of us is clearly the further development of NiceGUI; ensuring key app settings work correctly is important and worthy of detailed discussion.

evnchn:
@thetableman glad you forgived me. 

Two things: 
1. Check the NiceGUI Discord #off-topic if you want to chitchat. I'm looking to vent. 
2. It seems like you got hit by a breaking change back in #4217 (https://github.com/zauberzeug/nicegui/commit/cd9e5472ade27197765dc949b2a81af9dfc12134). That's bad but we should be able to promise that never happens again, since we [implement semantic version from 2.0.0 onwards](https://github.com/zauberzeug/nicegui/releases/tag/v2.0.0), and don't have any breaking changes until the next major version, 3.x, is released.

falkoschindler:
Thanks, @thetableman for this pull request and the fruitful discussion!
We would prefer to stick to uvicorn's convention. If we come across a use case where an explicit `protocol` parameter is needed in the future, we can easily add one.

evnchn:
Code LGTM

@thetableman may want to edit the first comment so that the merge message reflects accurately what you have done. 

@falkoschindler Check it out. No more param anymore.

thetableman:
I've removed the `protocol` parameter and updated the logic to check the presence of the key and cert files in determining whether to run over https.

I also challenged ChatGPT over it's suggestion that the protocol be a parameter and asked it to either provide real-word examples of the arguments it had made for the case or acknowledge it had potentially miss-informed me.

It's response was long - too long to cleanly share here though happy to if requested out of interest - but the TLDR was essentially: Yes, there are real-word examples but that some were over-represented in importance and other in likelihood.

evnchn:
Sure no problem. 

---

Sidenote on ChatGPT and LLM: It'd be fun, if you can share with me the prompt and output of ChatGPT. Am quite interested in LLM use. Then maybe test on some other models than ChatGPT. 

We've been combating hallucinations with AI coding assistants since it started. Most extreme case: half of Cursor's points are bogus: https://github.com/zauberzeug/nicegui/pull/4578#issuecomment-2816725655

So anything prompt engineering that can be done to make it better, can potentially drastically improve development efficiency.

falkoschindler (approved):
Great! I reviewed and tested the automatic protocol selection successfully.
Let's merge! 🎉

PR #4682, opened 2025-04-30
Add context manager to the user fixture

barrywhart:
This PR fix https://github.com/zauberzeug/nicegui/discussions/4667 by making user a context manager. By implementing `__enter__` and `__exit__` for user, we can with user: and then use the `ElementFilter` directly, overcoming how `user.find` returns a set, and enabling checking the order of filtered elements in testing.

### DIFF
--- a/nicegui/testing/user.py
+++ b/nicegui/testing/user.py
@@ -54,7 +54,7 @@ class User:
 
         Example:
             with user:
-                table = one(ElementFilter(kind=ui.table))
+                tables = list(ElementFilter(kind=ui.table))
         """
         return self._client.__enter__()
 
--- a/nicegui/testing/user.py
+++ b/nicegui/testing/user.py
@@ -46,6 +46,21 @@ class User:
             raise ValueError('This user has not opened a page yet. Did you forgot to call .open()?')
         return self.client
 
+    def __enter__(self):
+        """Enter context manager for the underlying page.
+
+        Within a context manager block, create an ElementFilter to find elements
+        on the page.
+
+        Example:
+            with user:
+                table = one(ElementFilter(kind=ui.table))
+        """
+        return self._client.__enter__()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return self._client.__exit__(exc_type, exc_val, exc_tb)
+
     def __getattribute__(self, name: str) -> Any:
         if name not in {'notify', 'navigate', 'download'}:  # NOTE: avoid infinite recursion
             ui.navigate = self.navigate
--- a/nicegui/testing/user.py
+++ b/nicegui/testing/user.py
@@ -46,19 +46,10 @@ class User:
             raise ValueError('This user has not opened a page yet. Did you forgot to call .open()?')
         return self.client
 
-    def __enter__(self):
-        """Enter context manager for the underlying page.
-
-        Within a context manager block, create an ElementFilter to find elements
-        on the page.
-
-        Example:
-            with user:
-                tables = list(ElementFilter(kind=ui.table))
-        """
+    def __enter__(self) -> Client:
         return self._client.__enter__()
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
         return self._client.__exit__(exc_type, exc_val, exc_tb)
 
     def __getattribute__(self, name: str) -> Any:
--- a/website/documentation/content/user_documentation.py
+++ b/website/documentation/content/user_documentation.py
@@ -120,8 +120,7 @@ doc.text('Using an ElementFilter', '''
 def using_an_elementfilter():
     with ui.row().classes('gap-4 items-stretch'):
         with python_window(classes='w-[400px]', title='UI code'):
-            ui.markdown(
-                '''
+            ui.markdown('''
                 ```python
                 ui.label('1').mark('number')
                 ui.label('2').mark('number')
--- a/website/documentation/content/user_documentation.py
+++ b/website/documentation/content/user_documentation.py
@@ -78,15 +78,6 @@ doc.text('Querying', '''
 
 @doc.ui
 def querying():
-    ui.markdown('''
-        You can also use `ElementFilter` directly with `user`'s context manager
-        syntax. This can be useful in the following cases:
-
-        - Retrieving an element and interacting with it directly
-        - Retrieving elements in the order they appear on the page (for example,
-          if a test needs to verify the sequence of elements)
-    ''')
-
     with ui.row().classes('gap-4 items-stretch'):
         with python_window(classes='w-[400px]', title='some UI code'):
             ui.markdown('''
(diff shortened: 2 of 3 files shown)


evnchn:
@barrywhart Mind if you make the PR self-descriptive by looking at only the first comment? Since it will be the merge message when we eventually merge the PR*

Something like: 

This PR fix #4667 by making user a context manager. By implementing `__enter__` and `__exit__` for `user`, we can `with user:` and then use the ElementFilter directly, overcoming how `user.find` returns a set, and enabling checking the order of filtered elements in testing. 

---

Meanwhile, some issues, from important to less important: 

* Did we not agree to hold back adding helper function `one`? Why is it in the docstring? 
* Perhaps we need to add documentation, as Rodja suggested? 

*the same mistake I make when I contribute the first time 😅

nicegui/testing/user.py
```diff
@@ -46,6 +46,21 @@ def _client(self) -> Client:
...
 
+    def __enter__(self):
+        """Enter context manager for the underlying page.
+
+        Within a context manager block, create an ElementFilter to find elements
+        on the page.
+
+        Example:
+            with user:
+                table = one(ElementFilter(kind=ui.table))
```
evnchn, line 57: Did we not agree to hold back adding helper function `one`? Why is it in the docstring?

evnchn (changes requested):
Did we not agree to hold back adding helper function `one`? Why is it in the docstring?

nicegui/testing/user.py
barrywhart: Oops, mindless copy paste from internal code. Will modify.

barrywhart:
I changed the PR description as you suggested. Is that what you had in mind?

barrywhart:
I updated the docstring to remove any mention of `one()` -- good catch!

barrywhart:
What do you have in mind for documentation? Perhaps updating this file`website/documentation/content/user_documentation.py`? How can I build and view the documentation after changing it?

website/documentation/content/user_documentation.py
```diff
@@ -101,6 +110,9 @@ def querying():
                 await user.should_see('Hello')
                 await user.should_see(marker='greeting')
                 await user.should_see(kind=ui.icon)
+                
+                with user:
+                    icons = list(ElementFilter(kind=ui.icon))
```
evnchn, line 115: Sorry but I don't think it is a good idea to (1) blend the new feature into the existing documentation part, and (2) not actually checking the sequence after getting the list.

evnchn (changes requested):

evnchn:
Would prefer having another `doc.text` and `@doc.ui` section. 

Drafted some text and code for the documentation. See if you think it's fine. 

Text: 
### Using ElementFilter directly
> Since `user.find` uses `set` to avoid duplicates, it does not preserve the order. 
> Consider how the order is not preserved after converting to a set and back in `list(set(["1", "2", "3"])) = ["3", "1", "2"]`. 
> Instead, it is possible to use ElementFilter directly by entering the user context. 

UI code would be: 
```py
ui.label("1").mark("text-123")
ui.label("2").mark("text-123")
ui.label("3").mark("text-123")
```

Testing code would be:
```py
with user:
    expected_text = ["1", "2", "3"]
    assert [label.text == expected for label, expected in zip(ElementFilter(marker="text-123"), expected_text)]
```

Other than that, LGTM.

barrywhart:
@evnchn: I updated the documentation as you suggested.

evnchn (approved):
LGTM as established previously

website/documentation/content/user_documentation.py
```diff
@@ -105,6 +105,53 @@ def querying():
             ''')
 
 
+doc.text(
+    'Using ElementFilter directly',
+    '''
+    Since `user.find` uses a set to avoid duplicates, it does not preserve the order.
```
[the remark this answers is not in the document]
evnchn: Interesting viewpoint on how `UserInteraction` may change. 

Well, maybe we can talk directly about the advantages of `ElementFilter`, instead of beating on existing disadvantages of `UserInteraction`. 

---

### Using `ElementFilter` directly

It may be desirable to use `ElementFilter` for the following purposes: 
- Preserve order of elements to check their order on the page
- More granular filtering options _(link to ElementFilter doc page)_, such as `ElementFilter(...).within(...)`

The `user` fixture supports the use of `ElementFilter` directly, after you enter its context (`with user:`)

barrywhart:
@falkoschindler, @rodja: This is ready for review.

rodja (approved):

falkoschindler (commented):
Thanks for the pull request, @barrywhart!

I just reviewed it and made some changes here and there:

- The docstring of `__enter__` is hard to find and doesn't show up when typing `with user` in the IDE. So I decided to remove it and leave it up to the documentation to explain things. The code itself is pretty transparent.
- The new methods `__enter__` and `__exit__` were missing type annotations.
- The test could be simplified a little bit. We don't need a page decorator or a click handler.
- Apart from some cleanup in the documentation, I think the "user assertions" didn't work. If you `assert [...]` a list, the test will always pass as long as the list isn't empty. It wouldn't matter if all its items were `False`.

The PR is basically ready to merge. But I have to ask one last question: Couldn't we simply write `with user.client:`? Do we really need to wrap the client context manager inside a user context manager and basically hide the fact that `with user:` opens a client context? I mean, we can leave pytest and documentation almost unchanged. But in my point of view we don't need the change in user.py.
@barrywhart, @rodja, what do you think?

barrywhart:
@falkoschindler: From my perspective, `user` represents an imaginary "human" user, and `client` is an _implementation detail_ of `user`.

By enabling `user` to expose the context manager behavior of the underlying user, we make it easier to handle two testing use cases that I believe will be pretty common:
- Building and using more complex filters, for example finding elements `within()` another one (not the entire page)
- Getting elements in the order they appear on the page. (We're using this to verify that dynamically built menus appear in the expected order.)

Are there any other situations where a test writer needs to directly use or be aware of the `client`? If the answer is "no" or "rarely", that seems like a good reason to make this change.

evnchn:
@falkoschindler I just noticed a strange thing:

- `self.client: Optional[Client] = None` is the **internal** one which is not supposed to be accessed because it can be None, and you would run into a cryptic error `TypeError: 'NoneType' object does not support the context manager protocol` if you try `with` into it, when it is `None`
- `def _client(self) -> Client:` is the **public-facing getter** which raises `ValueError('This user has not opened a page yet. Did you forgot to call .open()?')`, making it clear what's going on. 

Why is the internal one not getting the `_` prefix, but the public-facing one getting the `_` prefix? 

I feel like we did it wrongly the first time around...

Can we afford to swap the two, without breaking existing code? 

---

Otherwise, we should not endorse `with user.client:`, at least not without user education that they should `assert user.client` beforehand.

evnchn:
Perhaps both `.client` and `._client` will raise `ValueError('This user has not opened a page yet. Did you forgot to call .open()?')`, where we introduce a super-secret `.__client` which is for-real internal-use-only this time around.

evnchn:
@barrywhart From my perspective, whenever a user opens a NiceGUI page, we spawn in a `Client` in the Python side, which follows them throughout the page, whatever they do is handled by said `Client` as well. 

https://github.com/zauberzeug/nicegui/blob/096adcc54aa54dbe9130ebe5d380d1eae4452004/nicegui/page.py#L112-L120

Therefore, each user is associated with a `client`, and `ElementFilter` is a functionality offered by the server-side. 

So I see no issue with `with user.client:`. 

---

`with` into the `client` can do some more powerful things than just leveraging the `ElementFilter`. Consider:

https://nicegui.io/documentation/page#multicasting

```py
with client:
    ui.notify(message)
```

Heck, you can even the entire suite of UI functions, such as `ui.label`

https://github.com/zauberzeug/nicegui/blob/096adcc54aa54dbe9130ebe5d380d1eae4452004/tests/test_page.py#L330-L331

---

Though I'll grant that `with user:` is an awesome shorthand.

rodja:
I also like the shorthand abstraction of `with user` and think we should make the optional `user.client` private and the `user_client` property public in 3.0. I can not remember why we have done it this way...

falkoschindler (approved):
Ok, convinced. Let's keep the shorthand `with user:`.

Seeing `client` as implementation detail of `User` also explains the `client`/`_client` confusion:
Originally `User` had only a `client`. Even though it isn't marked as private, it was basically meant to be private (as far as I can remember). But because it can be `None`, we had to add checks throughout the `User` class. For convenience, we added a private `_client` property which wraps this check and always returns a `Client`. So where ever we expect a `Client`, we can use `_client`. And even though the getter is private, the exception can occur when the user calls other public methods like `should_see` that use `_client` internally. So if you call `user.should_see(...)` without opening a page first, you will get "This user has not opened a page yet."

Maybe we want to keep `client` public, but rename `_client` into a public property with a better name. This wouldn't require a breaking change. But this is where we struggled last time and simply fell back to `_client` instead.

PR #4683, opened 2025-05-01
run.py: little refactoring to gain basic android compatibility

jeff-dh:
This PR refactors the initialization of the ThreadPoolExecutor and the ProcessPoolExecutor in `run.py` a little. As a result they get initialized on the first call to `io_bound` respectively `cpu_bound` on demand.

This is necessary to achieve basic android compatibility (#578) since named semaphores are not available on android which the named Executors use and therefor throw an exception on initialization.

With this change nicegui runs on android under the following conditions:
- calling `run.io_bound` and `run.cpu_bound` will throw exceptions (named semaphores not available)
- host must not equal '0.0.0.0' for now (because otherwise `welcome.py` makes an `io_bound` call)

`run.io_bound` and `run.cpu_bound` support might follow in a separate PR.

 With this PR the basic example app at https://github.com/jeff-dh/niceApp works on android (at least in my environment)

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -27,6 +27,7 @@ def setup() -> None:
     except NotImplementedError:
         logging.warning('Failed to initialize ProcessPoolExecutor')
 
+
 class SubprocessException(Exception):
     """A picklable exception to represent exceptions raised in subprocesses."""
 
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -102,11 +102,9 @@ def tear_down() -> None:
         return
 
     kwargs = {'cancel_futures': True} if sys.version_info >= (3, 9) else {}
-
     if process_pool is not None:
         for p in process_pool._processes.values():  # pylint: disable=protected-access
             p.kill()
         process_pool.shutdown(wait=True, **kwargs)
-
     if thread_pool is not None:
         thread_pool.shutdown(wait=False, **kwargs)
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -21,8 +21,7 @@ def setup() -> None:
     """Setup the process pool. (For internal use only.)"""
     global process_pool  # pylint: disable=global-statement # noqa: PLW0603
     try:
-        if process_pool is None:
-            process_pool = ProcessPoolExecutor()
+        process_pool = ProcessPoolExecutor()
     except NotImplementedError:
         logging.warning('Failed to initialize ProcessPoolExecutor')
 
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -16,6 +16,21 @@ P = ParamSpec('P')
 R = TypeVar('R')
 
 
+def setup() -> None:
+    """Setup the process pool and the thread pool. (For internal use only.)"""
+    global process_pool  # pylint: disable=global-statement # noqa: PLW0603
+    global thread_pool  # pylint: disable=global-statement # noqa: PLW0603
+    try:
+        if process_pool is None:
+            process_pool = ProcessPoolExecutor()
+    except NotImplementedError:
+        pass
+    try:
+        if thread_pool is None:
+            thread_pool = ThreadPoolExecutor()
+    except NotImplementedError:
+        pass
+
 class SubprocessException(Exception):
     """A picklable exception to represent exceptions raised in subprocesses."""
 
@@ -66,17 +81,11 @@ async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs)
     It is encouraged to create static methods (or free functions) which get all the data as simple parameters (eg. no class/ui logic)
     and return the result (instead of writing it in class properties or global variables).
     """
-    global process_pool # pylint: disable=global-statement # noqa: PLW0603
-    if process_pool is None:
-        process_pool = ProcessPoolExecutor()
     return await _run(process_pool, safe_callback, callback, *args, **kwargs)
 
 
 async def io_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
     """Run an I/O-bound function in a separate thread."""
-    global thread_pool # pylint: disable=global-statement # noqa: PLW0603
-    if thread_pool is None:
-        thread_pool = ThreadPoolExecutor()
     return await _run(thread_pool, callback, *args, **kwargs)
 
 
@@ -87,10 +96,10 @@ def tear_down() -> None:
 
     kwargs = {'cancel_futures': True} if sys.version_info >= (3, 9) else {}
 
-    if not process_pool is None:
+    if process_pool is not None:
         for p in process_pool._processes.values():  # pylint: disable=protected-access
             p.kill()
         process_pool.shutdown(wait=True, **kwargs)
 
-    if not thread_pool is None:
+    if thread_pool is not None:
         thread_pool.shutdown(wait=False, **kwargs)
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 import sys
 import traceback
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
@@ -24,12 +25,12 @@ def setup() -> None:
         if process_pool is None:
             process_pool = ProcessPoolExecutor()
     except NotImplementedError:
-        pass
+        logging.warning('Failed to initialize ProcessPoolExecutor')
     try:
         if thread_pool is None:
             thread_pool = ThreadPoolExecutor()
     except NotImplementedError:
-        pass
+        logging.warning('Failed to initialize ThreadPoolExecutor')
 
 class SubprocessException(Exception):
     """A picklable exception to represent exceptions raised in subprocesses."""
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -66,16 +66,16 @@ async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs)
     It is encouraged to create static methods (or free functions) which get all the data as simple parameters (eg. no class/ui logic)
     and return the result (instead of writing it in class properties or global variables).
     """
-    global process_pool
-    if not process_pool:
+    global process_pool # pylint: disable=global-statement # noqa: PLW0603
+    if process_pool is None:
         process_pool = ProcessPoolExecutor()
     return await _run(process_pool, safe_callback, callback, *args, **kwargs)
 
 
 async def io_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
     """Run an I/O-bound function in a separate thread."""
-    global thread_pool
-    if not thread_pool:
+    global thread_pool # pylint: disable=global-statement # noqa: PLW0603
+    if thread_pool is None:
         thread_pool = ThreadPoolExecutor()
     return await _run(thread_pool, callback, *args, **kwargs)
 
@@ -84,12 +84,13 @@ def tear_down() -> None:
     """Kill all processes and threads."""
     if helpers.is_pytest():
         return
-    assert process_pool is not None
-    for p in process_pool._processes.values():  # pylint: disable=protected-access
-        p.kill()
+
     kwargs = {'cancel_futures': True} if sys.version_info >= (3, 9) else {}
 
-    if process_pool:
+    if not process_pool is None:
+        for p in process_pool._processes.values():  # pylint: disable=protected-access
+            p.kill()
         process_pool.shutdown(wait=True, **kwargs)
-    if thread_pool:
+
+    if not thread_pool is None:
         thread_pool.shutdown(wait=False, **kwargs)
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -10,16 +10,10 @@ from typing_extensions import ParamSpec
 from . import core, helpers
 
 process_pool: Optional[ProcessPoolExecutor] = None
-thread_pool = ThreadPoolExecutor()
+thread_pool: Optional[ThreadPoolExecutor] = None
 
 P = ParamSpec('P')
 R = TypeVar('R')
-
-
-def setup() -> None:
-    """Setup the process pool. (For internal use only.)"""
-    global process_pool  # pylint: disable=global-statement # noqa: PLW0603
-    process_pool = ProcessPoolExecutor()
 
 
 class SubprocessException(Exception):
@@ -72,11 +66,17 @@ async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs)
     It is encouraged to create static methods (or free functions) which get all the data as simple parameters (eg. no class/ui logic)
     and return the result (instead of writing it in class properties or global variables).
     """
+    global process_pool
+    if not process_pool:
+        process_pool = ProcessPoolExecutor()
     return await _run(process_pool, safe_callback, callback, *args, **kwargs)
 
 
 async def io_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
     """Run an I/O-bound function in a separate thread."""
+    global thread_pool
+    if not thread_pool:
+        thread_pool = ThreadPoolExecutor()
     return await _run(thread_pool, callback, *args, **kwargs)
 
 
@@ -88,5 +88,8 @@ def tear_down() -> None:
     for p in process_pool._processes.values():  # pylint: disable=protected-access
         p.kill()
     kwargs = {'cancel_futures': True} if sys.version_info >= (3, 9) else {}
-    process_pool.shutdown(wait=True, **kwargs)
-    thread_pool.shutdown(wait=False, **kwargs)
+
+    if process_pool:
+        process_pool.shutdown(wait=True, **kwargs)
+    if thread_pool:
+        thread_pool.shutdown(wait=False, **kwargs)
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -81,11 +81,17 @@ async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs)
     It is encouraged to create static methods (or free functions) which get all the data as simple parameters (eg. no class/ui logic)
     and return the result (instead of writing it in class properties or global variables).
     """
+    if process_pool is None:
+        raise RuntimeError('Process pool not set up. Call run.setup() first.')
+
     return await _run(process_pool, safe_callback, callback, *args, **kwargs)
 
 
 async def io_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
     """Run an I/O-bound function in a separate thread."""
+    if thread_pool is None:
+        raise RuntimeError('Thread pool not set up. Call run.setup() first.')
+
     return await _run(thread_pool, callback, *args, **kwargs)
 
 
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -18,9 +18,8 @@ R = TypeVar('R')
 
 
 def setup() -> None:
-    """Setup the process pool and the thread pool. (For internal use only.)"""
+    """Setup the process pool. (For internal use only.)"""
     global process_pool  # pylint: disable=global-statement # noqa: PLW0603
-    global thread_pool  # pylint: disable=global-statement # noqa: PLW0603
     try:
         if process_pool is None:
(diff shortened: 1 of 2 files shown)


evnchn:
The PR should have close to zero impact on existing code. 

Non-impacts: 
* `run.setup()` was never used outside of our code, and we said it was internal use only, so nobody can complain we dropped a function. 
{2588BBEF-1C48-4F68-A6B7-D6BB27A4DEA8}

* `process_pool` and `thread_pool` are also both internal to `run.py`. As it makes no sense to interact to these two variables without using `run.cpu_bound` or `run.io_bound`, nobody can complain we change the type and make it lazy-intitialized. 

{97EFB7D8-C9CF-45BF-95A9-728B680798E4}

{97EFB7D8-C9CF-45BF-95A9-728B680798E4}


* New `tear_down()` behaves just like the old one when `process_pool` and `thread_pool` are both initialized, while the new functionality of selectively skipping the shutdown of either pools if they are `None` being a backwards-compatible superset functionality. 

Minor impacts:
* First call to `run.cpu_bound` and `run.io_bound` may be ever so slightly slower. Perhaps we can restore the `setup()` function, which initializes the pools around a try-except. Then, we can avoid dropping a function, and bring the performance back to pre-PR levels. 

Perhaps:
```py
def setup() -> None:
    """Setup the process pool and the thread pool. (For internal use only.)"""
    global process_pool  # pylint: disable=global-statement # noqa: PLW0603
    global thread_pool  # pylint: disable=global-statement # noqa: PLW0603
    try:
        if process_pool is not None:
            process_pool = ProcessPoolExecutor()
    except Exception:
        pass
    try:
        if thread_pool is not None:
            thread_pool = ThreadPoolExecutor()
    except Exception:
        pass
```

@jeff-dh see what you think? I have even beefed up the logic to allow `run.setup()` to be ran multiple times, just in case for Android, the `ProcessPoolExecutor` is not available on boot but somehow it is after we do some fixing.

jeff-dh:
@evnchn I completely agree with you. I would use a single `try except` because if one fails I assume the second will also fail. you could also wrap it in `if 'android' in platform.uname().release`.

To add `io_bound` and `cpu_bound` support, I assume we might end up with some sort of platform independent replacement or wrapper for the executor pools or even the whole `run` module (which only consists of `io_bound` and `cpu_bound` + deps, right?). For me it's hard to predict how that's gonna turn out in detail and what would be the best preparation for it now.

evnchn:
> I would use a single `try except` because if one fails I assume the second will also fail.

It is an assumption I am not willing to make, and I am cranking the code robustness up to 11 here. 

I also don't want to do `if 'android' in platform.uname().release`, since I can imagine that the benefit of this PR extending beyond Android which doesn't have `ProcessPoolExecutor` and/or `ThreadPoolExecutor` (think: MicroPython? Non-CPython Python implementations? Python on the browser? etc...)

---

We can think about how to re-add support for `io_bound` and `cpu_bound`. The purpose of this PR, I think, is explicitly that on Android, NiceGUI doesn't blow up, just like FastAPI.

evnchn:
Use the "Hide whitespaces" view to see how cleanly this PR is at doing its job. 

{9E6658FB-3D2E-4307-859B-EABEB7F9B028}

jeff-dh:
By accident this turns out to have a side effect:

Since the setup function will cause the pool variables to be `None` this actually results in a call to `loop.run_in_executor(None,....)` and this DOES NOT crash, because:

>The executor argument should be an [concurrent.futures.Executor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) instance. The default executor is used if executor is None. The default executor can be set by [loop.set_default_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor), otherwise, a [concurrent.futures.ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) will be lazy-initialized and used by [run_in_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) if needed.

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor

This means we get the platform independent `io_bound` for free and the `cpu_bound` does "work" but is buggy (it's a thread that will block the GIL).

Btw: Please double check the setup function. I think the snippet proposed above has a bug, I assume it's supposed to be `if process_pool is None: process_pool = ....`

jeff-dh:
> The default executor can be set by [loop.set_default_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor), otherwise, a [concurrent.futures.ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) will be lazy-initialized and used by [run_in_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) if needed.

Even though I don't understand why "they" somehow manage to initialize a `ThreadPoolExecutor`?!? (assuming the docs are correct.... or is buildozer maybe applying a patch?)

evnchn:
> it's supposed to be `if process_pool is None: process_pool = ...`

It sure is! 🤦‍♂️Guess I'm too tired!

> `loop.run_in_executor(None,....)` and this DOES NOT crash

Hmm?! Thought it would crash with `NoneType object has no attribute XXX`. Sorry for that. But check the below as **we must fix this.** 

> we get the platform independent io_bound for free and the cpu_bound does "work" but is buggy

We can't afford to earn this free platform-independent `io_bound` by changing how the existing `io_bound` works for who-knows-how-many existing users. Also a buggy cpu_bound is not acceptable either. **We must raise an exception when the corresponding pool is None for those two functions.** 

> I don't understand why "they" somehow manage to initialize a `ThreadPoolExecutor`?!?

So do I! Why does it work?

https://github.com/python/cpython/blob/b8633f9aca9b198e5592106b649389d638cbc620/Lib/asyncio/base_events.py#L884-L898

{0F252D95-600A-4E3F-9C09-99DE9308F49A}

I think the possibility of a patch from buildozer / python for android is likely. They may be using a custom event loop which handles run_in_executor in a special manner. 

---

I think: 

`run`: The good-ol
`run2`: Using the default executor for `io_bound`, and launch-our-own-process technique for `cpu_bound`
`run_asyncio`: Use `asyncio.to_thread` for `io_bound`, and launch-our-own-process technique for `cpu_bound`

Reference: https://github.com/zauberzeug/nicegui/discussions/578#discussioncomment-13007174

jeff-dh:
> changing how the existing `io_bound` works for who-knows-how-many existing users. Also a buggy cpu_bound is not acceptable either. **We must raise an exception when the corresponding pool is None for those two functions.**

we only change the behavior if the initialization of the corresponding pool raises an `NotImplementedError`

>I think:
>
>run: The good-ol
>run2: Using the default executor for io_bound, and launch-our-own-process technique for cpu_bound
>run_asyncio: Use asyncio.to_thread for io_bound, and launch-our-own-process technique for cpu_bound

from my perspective this has one major drawback, all "legacy" code (e.g. `welcome.py`) needs to potentially be changed to be platform independent.

I know you don't like the android specific platform check, but for me it's common to have platform specific code / modules in some kind of a `#ifdef platform_is_blablub` guard. Maybe there's a better way to check which platform(s) we're running on / which modules we want to use.

```python
try:
    import run #good old unmodified run.py
except NotImplementedError:
    import run_asyncio as run
```

would probably work.

evnchn:
First of all: This is **off-topic** since this PR regards to gain basic android compat only. Using whatever method, block `io_bound` and `cpu_bound` from running in this PR when the corresponding pools are None, and that wraps it up. 

@jeff-dh I highly recommend to open a separate PR regarding implementing `run.io_bound` and `run.cpu_bound`. It's getting messy in this PR. 

---

> all "legacy" code (e.g. welcome.py) needs to potentially be changed to be platform independent

But that is explicitly the point. I do actively want them to switchover to the platform independent version, because we can't guarantee whether the platform independent version will behave exactly the same as existing one. 

Consider: 
- `special_function()` works if you run it in a separate ThreadPoolExecutor, but grinds to a halt otherwise. 
- If they do the switchover, then all is good, same code on all platforms, so if it fails in Dev it fails in Prod. 
- If they don't do the switchover, perhaps the code works on Dev, but fails in Prod. 

---

However, I think the switchover can be eased, if we have `run.set_run_method(method)`, where `method` can be `created_executor`, `default_executor`, or `asyncio_to_thread`. Then, we redirect `run.io_bound` to `run.created_executor.io_bound`, `run.default_executor.io_bound`, or `run.asyncio_to_thread.io_bound`. 

(Inspired heavily by `loop.set_default_executor`, `loop.set_event_loop` and the likes. Note how `asyncio` remains cross platform compatible, despite the event loops being all different on Windows and Linux, with some event loops (notable `uvloop`) not running on Windows at all.)

So, for Android `run.io_bound` and `run.io_bound`: We may just need `run.set_run_method(asyncio_to_thread)`, which then applies to the code everywhere. 

We will make the `set_run_method` error out if it doesn't work. Since you can argue having default run method being `created_executor` is also a set operation, on incompatible platforms such as Android, we warn and promote the switch to platform independent versions (we can't error out, since that would break functionality altogether)

evnchn:
@jeff-dh Can this work? I am way too tired to try this on Windows, let alone complete the rest and fiddle with Buildozer to test if this can work for Android. 

```py
import asyncio
import logging
import sys
import traceback
from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
from functools import partial
from typing import Any, Callable, Optional, TypeVar

from typing_extensions import ParamSpec

from . import core, helpers

P = ParamSpec('P')
R = TypeVar('R')


class SubprocessException(Exception):
    """A picklable exception to represent exceptions raised in subprocesses."""

    def __init__(self, original_type, original_message, original_traceback) -> None:
        self.original_type = original_type
        self.original_message = original_message
        self.original_traceback = original_traceback
        super().__init__(f'{original_type}: {original_message}')

    def __reduce__(self):
        return (SubprocessException, (self.original_type, self.original_message, self.original_traceback))

    def __str__(self):
        return (f'Exception in subprocess:\n'
                f'  Type: {self.original_type}\n'
                f'  Message: {self.original_message}\n'
                f'  {self.original_traceback}')


def safe_callback(callback: Callable, *args, **kwargs) -> Any:
    """Run a callback; catch and wrap any exceptions that might occur."""
    try:
        return callback(*args, **kwargs)
    except Exception as e:
        # NOTE: we do not want to pass the original exception because it might be unpicklable
        raise SubprocessException(type(e).__name__, str(e), traceback.format_exc()) from None


async def _run(executor: Any, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
    if core.app.is_stopping:
        return  # type: ignore  # the assumption is that the user's code no longer cares about this value
    try:
        loop = asyncio.get_running_loop()
        return await loop.run_in_executor(executor, partial(callback, *args, **kwargs))
    except RuntimeError as e:
        if 'cannot schedule new futures after shutdown' not in str(e):
            raise
        pass
    return  # type: ignore  # the assumption is that the user's code no longer cares about this value


class BaseIOBoundRunner:
    """A base class to run I/O-bound functions in a separate thread."""

    async def io_bound(self, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
        raise NotImplementedError('This method should be overridden by subclasses')

    def _cleanup(self) -> None:
        """Cleanup resources."""
        pass


class DefaultIOBoundRunner:
    """A class to run I/O-bound functions in a separate thread."""

    def __init__(self) -> None:
        self._thread_pool = ThreadPoolExecutor()

    async def io_bound(self, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
        return await _run(self._thread_pool, callback, *args, **kwargs)

    def _cleanup(self) -> None:
        kwargs = {'cancel_futures': True} if sys.version_info >= (3, 9) else {}
        self._thread_pool.shutdown(wait=True, **kwargs)


class BaseCPUBoundRunner:
    """A base class to run CPU-bound functions in a separate process."""

    async def cpu_bound(self, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
        raise NotImplementedError('This method should be overridden by subclasses')

    def _cleanup(self) -> None:
        """Cleanup resources."""
        pass


class DefaultCPUBoundRunner:
    """A class to run CPU-bound functions in a separate process."""

    def __init__(self) -> None:
        self._process_pool = ProcessPoolExecutor()

    async def cpu_bound(self, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
        return await _run(self._process_pool, callback, *args, **kwargs)

    def _cleanup(self) -> None:
        for p in self._process_pool._processes.values():  # pylint: disable=protected-access
            p.kill()
        kwargs = {'cancel_futures': True} if sys.version_info >= (3, 9) else {}
        self._process_pool.shutdown(wait=True, **kwargs)


cpu_bound_runner: Optional[BaseCPUBoundRunner] = None
io_bound_runner: Optional[BaseIOBoundRunner] = None


def setup(cpu_bound_runner_cls: Optional[type[BaseCPUBoundRunner]] = None,
          io_bound_runner_cls: Optional[type[BaseIOBoundRunner]] = None) -> None:
    """Setup the process pool and thread pool."""
    global cpu_bound_runner, io_bound_runner  # pylint: disable=global-statement # noqa: PLW0603

    try:
        cpu_bound_runner = (cpu_bound_runner_cls or DefaultCPUBoundRunner)()
    except Exception as e:
        logging.warning(f'Failed to create CPU-bound runner: {e}')

    try:
        io_bound_runner = (io_bound_runner_cls or DefaultIOBoundRunner)()
    except Exception as e:
        logging.warning(f'Failed to create I/O-bound runner: {e}')


async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
    """Run a CPU-bound function in a separate process.

    `run.cpu_bound` needs to execute the function in a separate process.
    For this it needs to transfer the whole state of the passed function to the process (which is done with pickle).
    It is encouraged to create static methods (or free functions) which get all the data as simple parameters (eg. no class/ui logic)
    and return the result (instead of writing it in class properties or global variables).
    """
    if cpu_bound_runner is None:
        raise RuntimeError('CPU-bound runner not set up. Please call setup() first.')
    return await cpu_bound_runner.cpu_bound(callback, *args, **kwargs)


async def io_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
    """Run an I/O-bound function in a separate thread."""
    if io_bound_runner is None:
        raise RuntimeError('I/O-bound runner not set up. Please call setup() first.')
    return await io_bound_runner.io_bound(callback, *args, **kwargs)


def tear_down() -> None:
    """Kill all processes and threads."""
    if helpers.is_pytest():
        return
    if cpu_bound_runner:
        cpu_bound_runner._cleanup()
    if io_bound_runner:
        io_bound_runner._cleanup()
```

evnchn:
@jeff-dh With regards to my mess at https://github.com/zauberzeug/nicegui/pull/4683#issuecomment-2846085636, I think it can potentially work, if we take some time to polish it, and include 3 sets of `XXXCPUBoundRunner` for the user to choose from. Do you want to try do it?

jeff-dh:
@evnchn I agree (and assume the `XXXBoundRunner` approach works). But I think it might end up in more than "little refactoring". As a result I don't think it aligns as well with this PR as the other solution does (little refactoring -> basic android support) unless we're talking about the bigger picture, which has some open questions for me, e.g.:
- how to get the new parameters of `run.setup` into it (it get's called from the lifetime hook in `nicegui.py` which gets initialized on import.... I think it might be tricky to pass a flag from the user into it)
- what's the design for platform independent `io/cpu_bound` in the big picture? Add a new `ui.run` parameter which somehow gets passed to `run.setup`? Or maybe something like the `run_async` approach or something completely different?
- ....

I think we probably have all the parts on the table by now and it's a design decision how to continue. Since I understand that this PR is supposed to make NiceGUI compile and start on android with as little impact as possible, I would prefer the non `XXXBoundRunner` approach and whether to use `XXXBoundRunner` or `run_async` or whatever is to be discussed in a further (`io_bound` & `cpu_bound`) discussion.

PS: Please have a look at the `nicegui.py` change. I think it should not have side effects..... The alternative would be to move `run.setup` to the top.

Btw: It turns out the `ThreadPoolExecutor` is initializable and works on android (which aligns with "why the default_executor works"). Don't know why I got this wrong before, sorry!

nicegui/run.py
```diff
@@ -10,17 +11,26 @@
 from . import core, helpers
 
 process_pool: Optional[ProcessPoolExecutor] = None
-thread_pool = ThreadPoolExecutor()
+thread_pool: Optional[ThreadPoolExecutor] = None
```
rodja, line 14: @jeff-dh, in https://github.com/zauberzeug/nicegui/pull/4683#issuecomment-2847680047 you wrote

> It turns out the ThreadPoolExecutor is initializable and works on android (which aligns with "why the default_executor works").

Why does this PR still needs it to be initialized in `run.setup()`?

nicegui/run.py
```diff
@@ -72,21 +82,31 @@ async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs)
...
-    process_pool.shutdown(wait=True, **kwargs)
-    thread_pool.shutdown(wait=False, **kwargs)
+
+    if process_pool is not None:
+        for p in process_pool._processes.values():  # pylint: disable=protected-access
+            p.kill()
+        process_pool.shutdown(wait=True, **kwargs)
+
+    if thread_pool is not None:
+        thread_pool.shutdown(wait=False, **kwargs)
```
rodja, line 112:
```suggestion
    if process_pool is not None:
        for p in process_pool._processes.values():  # pylint: disable=protected-access
            p.kill()
        process_pool.shutdown(wait=True, **kwargs)
    if thread_pool is not None:
        thread_pool.shutdown(wait=False, **kwargs)
```

Blank lines should [only be used sparingly](https://legacy.python.org/dev/peps/pep-0008/#blank-lines:~:text=Use%20blank%20lines%20in%20functions%2C%20sparingly%2C%20to%20indicate%20logical%20sections.).

rodja (changes requested):

nicegui/run.py
evnchn: Extending beyond Android compatibility, this can potentially offer better resilience against even more feature-restrictive Python environments. 

Consider iOS, which Buildozer suppport via Kivy iOS. 

Given how locked-down iOS is, I am not sure if `ThreadPoolExecutor` is available on there. I can't test, either.

jeff-dh: When I started working on this, the `ThreadPoolExecutor` caused issues or it at least heavily seemed to cause issues. Now niceGUI runs without the `ThreadPoolExecutor` changes in my environment.

I can't reproduce the former behvaviour and have no explanation why this could have changed (I'm not aware of local changes or other changes in my environment).

It might (not) be necessary........ ;) It's probably whether to be precaution or wait until this might cause trouble (again).

Can anyone else check this in another environment?

rodja: I would prefer to not make the thread pool optional. Let's keep this PR as small as possible. It's better to know and understand the exact requirements before implementing a mechanism like this.

nicegui/run.py
```diff
@@ -19,7 +20,11 @@
 def setup() -> None:
     """Setup the process pool. (For internal use only.)"""
     global process_pool  # pylint: disable=global-statement # noqa: PLW0603
-    process_pool = ProcessPoolExecutor()
+    try:
+        if process_pool is None:
```
jeff-dh, line 24: what about this line? I think is unnecessary too....

falkoschindler: Yes, I think we don't need the if statement.

rodja (approved):

evnchn:
I think I get what @rodja was getting at. 

Since Android actually supports `ThreadPoolExecutor`, then we can afford to make _as minimal as changes can be in this PR_, deliverying the **bare minimum** to get Android working. 

That means the only thing we are doing is turning asserts into if-elses and adding try-except around `ProcessPoolExecutor` initialization. 

I still have concerns over telling the user to "Call run.setup() first", as `run.setup()` remains an internal function ("For internal use only."). Thus, we should not endorse calling internal functions. Do you want to change that directly?

falkoschindler (approved):
Thanks everyone for the fruitful discussion, and especially @jeff-dh for proposing the pull request! Moving towards Android support with only a few minor code changes is a great result. Let's merge! 🚀

PR #4688, opened 2025-05-02
Sitewide index for AI

evnchn:
This PR fix #3412 by serving another 2 JSON files in addition to the existing `search_index.json`. 

- /static/sitewide_index.json: Compared to `search_index.json`, it has the code of the documentation pages, but it does not have the examples

- /static/examples_index.json: Just the examples as we would otherwise see in `search_index.json`

This greatly benefits AI as they can learn about the syntax of NiceGUI, and no need to worry about 2 mixed items in the `search_index.json` muddying the waters and making AI difficult to learn from the pattern. 

---

There should not be any impact on the existing NiceGUI website.

### DIFF
--- a/website/documentation/code_extraction.py
+++ b/website/documentation/code_extraction.py
@@ -1,11 +1,10 @@
 
 
-import re
-import isort
-
-
 import inspect
+import re
 from typing import Callable
+
+import isort
 
 UNCOMMENT_PATTERN = re.compile(r'^(\s*)# ?')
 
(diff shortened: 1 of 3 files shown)


evnchn:
Next move would be, hopefully, to implement some sort of LLM agent. This PR prepares for me and others to do that.

evnchn:
The following is the upper bound of what the AI agent can achieve. Using DeepSeek R1 with putting the entire JSON in the context window via attachment. 

https://poe.com/s/vqlCa2Se7VTQTGDW6yrm

```py
from nicegui import ui

def caesar_cipher(text: str, shift: int = 3) -> str:
    result = []
    for char in text:
        if char.isalpha():
            start = ord('A') if char.isupper() else ord('a')
            new_char = chr((ord(char) - start + shift) % 26 + start)
            result.append(new_char)
        else:
            result.append(char)
    return ''.join(result)

@ui.page('/text_transformer')
def text_transformer():
    with ui.column().classes('w-full items-stretch gap-4 p-4'):
        input_text = ui.textarea(label='Input Text').classes('w-full')
        output = ui.label().classes('text-lg font-mono')
        
        with ui.row().classes('gap-2'):
            ui.button('UPPERCASE', on_click=lambda: output.set_text(input_text.value.upper()))
            ui.button('lowercase', on_click=lambda: output.set_text(input_text.value.lower()))
            ui.button('Caesar Cipher', on_click=lambda: output.set_text(caesar_cipher(input_text.value)))
            ui.button('Reverse', on_click=lambda: output.set_text(input_text.value[::-1]))
            ui.button('Clear', on_click=lambda: (input_text.set_value(''), output.set_text('')))
        
        ui.separator()
        with ui.expansion('Advanced Options', icon='settings').classes('w-full'):
            with ui.row():
                shift = ui.number('Caesar Shift', value=3, min=1, max=25)
                ui.button('Apply Shift', on_click=lambda: output.set_text(
                    caesar_cipher(input_text.value, int(shift.value))
                ))

ui.run()
```

{079F2F60-3FD8-4C8E-831A-496C8EEC9765}

falkoschindler (approved):
Awesome, @evnchn!
I just refactored the code a bit to reduce duplication and to make it more obvious what each index is supposed to contain:
```py
def build_search_index() -> None:
    """Build search index."""
    search_index[:] = _collect_documentation_parts(include_code=False) + _collect_examples()
    sitewide_index[:] = _collect_documentation_parts(include_code=True)
    examples_index[:] = _collect_examples()
```

PR #1134, opened 2023-07-06
using hierarchical path names to identify JS dependencies

rodja:
ToDos

- [x] fix importmap to remove mjs warnings
- [x] adapt examples (custom vue component etc)
- [x] make tests pass again
- [x] define dependencies while subclassing?
- [x] timing issue when asynchronously loading dependencies
- [x] Check `Path('foo/bar')` on Windows
- [x] convert `element.components` into `element.component`
- [x] use ID instead of relative path for identifying dependency
- [x] new pytests for adding elements with dependencies on button click
- [x] new pytest for adding AG Grid on button click and selecting rows
- [x] globbing?

PR #4692, opened 2025-05-03
Use the Mermaid API directly to fix nested diagram inside `ui.dialog`

### LINKED ISSUE #4518: `ui.mermaid` renders incorrectly within `ui.dialog`
### First Check

### Example Code

```python
with ui.dialog(value=True), ui.card():
    ui.mermaid('graph TD; A --> B')
```

### Description

I'm expecting to see this:
Image

But I'm getting this:
Image

### NiceGUI Version

2.13.0

### Python Version

3.8

### Browser

Chrome

### Operating System

macOS

### Additional Context

This problem was originally reported by @ntmarkus as part of issue #4502. After quite some discussion I isolated this aspect as a separate issue.

A very suboptimal workaround is to add a delay:
```py
with ui.dialog(value=True), ui.card():
    ui.timer(2.0, lambda: ui.mermaid('graph TD; A --> B'), once=True)
```

evnchn:
This PR fix #4518 magically by invoking the Mermaid API exactly as prescribed here: 

https://mermaid.js.org/config/usage.html#api-usage

I think the commit wraps it up very well: 

> Dear mermaid: we'll call API ourselves thankyouverymuch

Their code: 
```js
  element = document.querySelector('#graphDiv');
  const graphDefinition = 'graph TB\na-->b';
  const { svg, bindFunctions } = await mermaid.render('graphDiv', graphDefinition);
  element.innerHTML = svg;
  // This can also be written as `bindFunctions?.(element);` using the `?` shorthand.
  if (bindFunctions) {
    bindFunctions(element);
  }
```

My code:
```js
let element = queue.shift(); // same mindset
let innertext = this.content; // mine's from Vue
const { svg, bindFunctions } = await mermaid.render('graphDiv', innertext); // literally the same except variable names
element.innerHTML = svg; // literally the same
if (bindFunctions) {
   bindFunctions(element); // literally the same
}
```

If I had to guess, the reason it was fixed, is because '#graphDiv' does not exist, and so Mermaid just "assumes full size" and renders the proper diagram without being pinched to absolutely tiny. 

Note that there are a couple of shortcomings which I haven't tested / haven't fixed, so don't merge just yet. 

- [x] what happens when we change the content? Does the mermaid diagram update appropriately? 
- [x] As of how, if there is syntax error, we lose the good-ol "Mermaid Bomb" error screen. Should we have better means? 
- [x] Are the Mermaid config respected (`{'securityLevel': 'loose', ...}` - allow running JavaScript when a node is clicked; `{'logLevel': 'info', ...}` - log debug info to the console) respected under this approach? 
- [x] Does this PR impact existing code? If so, how much? 
- [x] Passing 'graphDiv' when '#graphDiv' doesn't exist seems kinda sketchy. Are we absolutely sure that's the way to go?

---

image

Perfect diagram, despite disable cache is false / disable cache is on and no network throttle.

### DIFF
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -24,21 +24,12 @@ export default {
       try {
         const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", content);
         this.$el.innerHTML = svg;
-        if (bindFunctions) {
-          bindFunctions(this.$el);
-        }
+        bindFunctions?.(this.$el);
       } catch (error) {
         const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", "error");
         this.$el.innerHTML = svg;
-        if (bindFunctions) {
-          bindFunctions(this.$el);
-        }
-        const mermaidErrorFormat = {
-          str: error.message,
-          message: error.message,
-          hash: error.name,
-          error
-        }
+        bindFunctions?.(this.$el);
+        const mermaidErrorFormat = { str: error.message, message: error.message, hash: error.name, error };
         console.error(mermaidErrorFormat);
         this.$emit("error", mermaidErrorFormat);
       }
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -31,7 +31,17 @@ export default {
       is_running = true;
       while (queue.length) {
         try {
-          await mermaid.run({ nodes: [queue.shift()] });
+          let element = queue.shift();
+
+          let innertext = this.content;
+
+          const { svg, bindFunctions } = await mermaid.render('graphDiv', innertext);
+
+          element.innerHTML = svg;
+
+          if (bindFunctions) {
+            bindFunctions(element);
+          }
         } catch (error) {
           console.error(error);
           this.$emit("error", error);
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -1,10 +1,5 @@
 import mermaid from "mermaid";
 
-window.mermaid = mermaid;
-
-let is_running = false;
-const queue = [];
-
 export default {
   template: `<div></div>`,
   data: () => ({
@@ -27,9 +22,8 @@ export default {
       if (this.last_content === content) return;
       this.last_content = content;
       try {
-        const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", content)
+        const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", content);
         this.$el.innerHTML = svg;
-
         if (bindFunctions) {
           bindFunctions(this.$el);
         }
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -24,8 +24,7 @@ export default {
     async update(content) {
       if (this.last_content === content) return;
       this.last_content = content;
-      this.$el.innerHTML = content;
-      this.$el.removeAttribute("data-processed");
+      this.$el.setAttribute("mermaid_content", content);
       queue.push(this.$el);
       if (is_running) return;
       is_running = true;
@@ -33,9 +32,7 @@ export default {
         try {
           let element = queue.shift();
 
-          let innertext = this.content;
-
-          const { svg, bindFunctions } = await mermaid.render('graphDiv', innertext);
+          const { svg, bindFunctions } = await mermaid.render(element.id+"_mermaid", element.getAttribute("mermaid_content"))
 
           element.innerHTML = svg;
 
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -24,27 +24,17 @@ export default {
     async update(content) {
       if (this.last_content === content) return;
       this.last_content = content;
-      this.$el.setAttribute("mermaid_content", content);
-      queue.push(this.$el);
-      if (is_running) return;
-      is_running = true;
-      while (queue.length) {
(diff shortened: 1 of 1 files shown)


evnchn:
Hi @falkoschindler 

Since you are awake, mind if you tell me why https://github.com/zauberzeug/nicegui/commit/d86cc1d9309a32212a7e220df9d19fd02c8131ac was implemented? 

You wrote at https://github.com/zauberzeug/nicegui/pull/1134#pullrequestreview-1524617793:

> Ok, I think we're finally ready to merge. Today I struggled a bit with another Mermaid bug (not caused by this PR but already present in the 1.3 branch), but I could fix it with a queue in mermaid.js handling async processes one at a time.

Could you recall what bug is it? Since we are, apparently, computing the SVG from an non-existent `graphDiv` element (with Mermaid renderer having no idea that we are going to apply the SVG to `element`, really), perhaps we can get rid of the `queue` entirely. 

This can be a performance uplift as we enable simultaneous rendering of multiple Mermaid diagram at the same time, and potentially allow parallel rendering which you mention not working well in https://github.com/zauberzeug/nicegui/pull/4170#discussion_r2046341683

evnchn:
Notable changes: 

- Drop the queue, since we are doing the rendering ourselves, we no need to worry about conflicting `mermaid.run` calls. 
- Use ID "c6_mermaid" given element ID "c6" for mermaid, and deleting the svg content before render, to avoid mermaid not updating on update. 
- Draw also the Mermaid error screen if error
- Give error in Mermaid format if error to maintain compatibility with existing code. 

@falkoschindler All 5 elements in the initial checklist are completed. Feel like this is safe for merge and ready for your review. 


Test page: 

```py
from nicegui import ui
import random


@ui.page('/mermaid')
def mermaid_page():
    with ui.dialog(value=True).props("transition-duration=1500"), ui.card():
        ui.mermaid('graph TD; A --> B')


@ui.page('/many_mermaid')
def mermaid_page():
    with ui.dialog(value=True).props("transition-duration=1500"), ui.card().classes("w-full"), ui.grid(columns=3):
        def generate_random_graph():
            nodes = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J']
            edges = [f"{random.choice(nodes)} --> {random.choice(nodes)}" for _ in range(1)]
            return f"graph TD; {' '.join(edges)}"

        mermaids = [ui.mermaid(generate_random_graph()) for _ in range(9)]

        def set_random_graphs():
            for mermaid in mermaids:
                mermaid.set_content(generate_random_graph())

        ui.button('Set all to random', on_click=set_random_graphs)


@ui.page('/click_mermaid')
def click_mermaid_page():
    with ui.dialog(value=True).props("transition-duration=1500"), ui.card():
        ui.mermaid('''
        graph LR;
            A((Click me!));
            click A call emitEvent("mermaid_click", "You clicked me!")
        ui.on('mermaid_click', lambda e: ui.notify(e.args))


@ui.page('/click_mermaid2')
def click_mermaid_page2():
    with ui.dialog(value=True).props("transition-duration=1500"), ui.card():
        ui.mermaid('''
        graph LR;
            A((Click me!));
            click A call emitEvent("mermaid_click", "You clicked me!")
        ''')
        ui.on('mermaid_click', lambda e: ui.notify(e.args))
        ui.label('Not supposed to work due to missing config {"securityLevel": "loose"}')


def print_mermaid_event(event):
    try:
        print(event.args['message'])
    except KeyError:
        print("WARNING: No message in event args")
        print(event.args)


@ui.page('/broken_mermaid')
def broken_mermaid_page():
    with ui.dialog(value=True).props("transition-duration=1500"), ui.card():
        ui.mermaid('''
        graph LR;
            A --> B;
            A -> C;
        ''').on('error', lambda e: print(e.args['message']))


ui.run(port=9191, show=False)
```

falkoschindler:
Thanks, @evnchn! This PR looks amazing!

> Since you are awake, mind if you tell me why d86cc1d9309a32212a7e220df9d19fd02c8131ac was implemented?

Unfortunately I don't remember. Mermaid has always been one of the more tricky elements. But maybe we just didn't handle it correctly. Your implementation seems much cleaner.

Just a few questions:

1. Why are you creating `mermaidErrorFormat`? Why not simply logging and emitting `error`?
2. Why are you passing "error" to `mermaid.render`? Does it have a special meaning to Mermaid?

evnchn:
> 1. Why are you creating `mermaidErrorFormat`? Why not simply logging and emitting `error`?

Otherwise, you'd break existing code, such as: https://nicegui.io/documentation/mermaid#handle_errors

Of which, `e.args['message']` would fail. 

with `const mermaidErrorFormat = { str: error.message, message: error.message, hash: error.name, error };`, I make the error return format exactly the same as Mermaid, ensuring copatibility. 

Look for this in `mermaid.esm.js`

{33A37042-D433-4847-AE25-ABA94683579A}


> 2. Why are you passing "error" to `mermaid.render`? Does it have a special meaning to Mermaid?

For it to show it's bomb thing. Passing "error" does have special meaning to Mermaid. It is what is used to draw the bomb, which is itself a type of Mermaid diagram. 

{6D5B3193-BA28-4AAC-B727-E6E4EE2F54EA}

{1D68596B-6104-4543-9B9F-013243381A05}

evnchn:
https://www.npmjs.com/package/mermaid?activeTab=code

Source: `dist/mermaid.esm.mjs`

falkoschindler (approved):
Ah I see! It's a weird format with redundant payload, but to prevent breaking changes we need to stick with it for now.

All questions answered. Code looks clean. Ready to merge! 😎

PR #4696, opened 2025-05-04
Document how to update context menu items

evnchn:
This PR fix #4695 by mentioning how to update context menu items. 

As it turns out, it's decently tricky. Took me a while to figure out as well, and it's not mentioned in the docs AFAIK. 

{CE1DB11B-3F19-4AE2-BD30-B8B03A364E88}

### DIFF
--- a/website/documentation/content/context_menu_documentation.py
+++ b/website/documentation/content/context_menu_documentation.py
@@ -13,29 +13,22 @@ def main_demo() -> None:
             ui.menu_item('Reset', auto_close=False)
 
 
-@doc.demo('Updating context menu items', '''
-    Only the first `ui.context_menu` instance is respected by Quasar. 
-          
-    Therefore, the following paradigm is used to update the context menu items:
-          
-    - Create a new context menu while keeping a reference (`with ui.context_menu() as cmenu:`).
-    - Clear any existing items in the context menu (`cmenu.clear()`).
-    - Add new items to the existing context menu (`with cmenu: ...`).
+@doc.demo('Context menus with dynamic content', '''
+    To show a context menu with content that changes dynamically, e.g. based on the position of the mouse,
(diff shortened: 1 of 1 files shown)


evnchn:
@Anindya088 check this out

falkoschindler:
Hi @evnchn,
I'm not sure this new demo is 100% spot on.

> Only the first `ui.context_menu` instance is respected by Quasar.

Well, only the first instance _within one UI element_ is respected by Quasar.

> Therefore, the following paradigm is used to update the context menu items

Even though updating an existing context menu is better than having more than one context menu, you could also remove the old one and create a new context menu. I mean, the demo is still useful, but in my opinion it's not the logical consequence.

The problem in #4695 was that the "cellContextMenu" event is raised without any existing context menu, which is why the system's context menu appears. Then a NiceGUI context menu is created with is opened from then on, ignoring all future context menus. The essential trick is to have a possibly empty context menu which is then populated with new items.

What do you think about this alternative?
```py
@doc.demo('Context menus with dynamic content', '''
    To show a context menu with content that changes dynamically, e.g. based on the position of the mouse,
    it is recommended to re-use the same context menu instance.
    This demo shows how to clear the context menu and add new items to it.
''')
def update_context_menu() -> None:
    from nicegui import events

    def update_menu(e: events.MouseEventArguments) -> None:
        context_menu.clear()
        with context_menu:
            ui.menu_item(f'Add circle at ({e.image_x:.0f}, {e.image_y:.0f})')

    source = 'https://picsum.photos/id/377/640/360'
    with ui.interactive_image(source, on_mouse=update_menu, events=['contextmenu']):
        context_menu = ui.context_menu()
```
It is closer to the problem in #4695 because the menu items depend on information from the event itself.

falkoschindler (approved):
Alright, I quickly did it myself, so we co-author this PR. 😊

PR #4708, opened 2025-05-06
`run.set_cpu_bound_executor_type`: enable custom (android compatible) `cpu_bound`-exeuctors

jeff-dh:
This PR proposes a new function in the run module -- called `set_cpu_bound_executor_type` -- which allows to exchange the type of the `(ProcessPool)Executor` used in `cpu_bound`.

As discussed in #4683 and #578 `run.cpu_bound` is not android compatible. The reason is that `concurrent.futures.ProcessPoolExecutor` is not android compatible (uses `multiprocessing.Queue` which uses named semaphores which are not available on android).
We gain (`cpu_bound`-) android compatibility as soon as we use an android compatible executor.

The proposed `set_cpu_bound_executor_type` function would outsource the mentioned issue to a custom executor which could be choosen and installed by the user.

With this PR, this works (on Linux, Windows (@evnchn right?) and Android):

```python
from nicegui import run
from aio_process_pool import Executor as AsyncExecutor

run.set_cpu_bound_executor_type(AsyncExecutor)
```

Note: `aio_process_pool` is an -- at least proof of concept -- implementation for an android compatible executor and is [available](https://github.com/jeff-dh/aio_process_pool).

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -19,17 +19,8 @@ R = TypeVar('R')
 
 
 def set_cpu_bound_executor_type(executor_type: type):
-    """Set concurrent.futures.Executor compliant type for cpu_bound execution
-
-        Note: unfortunately this function will (very likely) be called before
-              all functions that are supposed to be called through the executor
-              are defined. As a result we can not instantiate the executor in
-              this function. If we would, the forked worker processes would not
-              be able to call these functions, because they are not defined yet
-              and thus don't exist in processes that are forked NOW....
-              Therefor this uncommon approach.
-    """
-    # makes no sense to call this function after ui.run which calls setup, right???
+    """Set concurrent.futures.Executor compliant type for cpu_bound execution"""
+    # makes no sense to call this function after ui.run which calls setup
     assert process_pool is None
     assert type(executor_type) is type
 


jeff-dh:
Idea for a minimal impact integration:

_TODO I_
instead of adding a custom executor interface we could adapt to `concurrent.futures.Executor`. As a result `set_process_pool_executor` should be able to simply apply a custom executor to the regular `run.process_pool` and hopefully no further changes are necessary.

_TODO II_
extract the custom process pool implementation (which is now supposed to be a `concurrent.futures.Executor` compliant executor) into a separate pypi package.

The result should be:

```python
from nicegui import run
from alternativeProcessPoolPyPiPackage import AlternativeProcessPoolExecutor

run.set_cpu_bound_executor(AlternativeProcessPoolExecutor())
```

 Necessary changes in `run.py` hopefully reduce to:
```python
def set_cpu_bound_executor(executor: Executor):
    global process_pool
    if process_pool is not None:
        shutdown_process_pool()

    process_pool = executor

# extracted a couple of lines from tear_town
def shutdown_process_pool():
     ...
```

jeff-dh:
```python
from nicegui import ui, run
from aio_process_pool import AsyncProcessPool

run.set_cpu_bound_executor_type(AsyncProcessPool)
```

semms to work. `aio_process_pool` is [available](https://pypi.org/project/aio_process_pool/).

I had to adjust the planed interface slightly (+ `_type`), see comments in code for details why.

nicegui/run.py
```diff
@@ -2,26 +2,46 @@
...
+              be able to call these functions, because they are not defined yet
+              and thus don't exist in processes that are forked NOW....
+              Therefor this uncommon approach.
+    """
+    # makes no sense to call this function after ui.run which calls setup, right???
+    assert process_pool is None
+    assert type(executor_type) is type
+
+    global ProcessPoolExecutorType # pylint: disable=global-statement # noqa: PLW0603
+    ProcessPoolExecutorType = executor_type
```
jeff-dh, line 37: we could rename `executor_type` to `executor_factory`.... ;)

evnchn:
Sorry @jeff-dh I am not sure if your `AsyncProcessPool` is fully "concurrent.futures.Executor compliant"

These two should be minor since NiceGUI doesn't use either one. 
1. No `.map`
2. Does not support the context manager protocol

This one I am not sure:

3. For `pool.submit`:
  - your `AsyncProcessPool` implementation returns a `<class '_asyncio.Task'>` which should be `await`ed. 
  - Meanwhile `ProcessPoolExecutor` returns a `<class 'concurrent.futures._base.Future'>` which you definitely can't `await` it (or `TypeError: object Future can't be used in 'await' expression`), rather you apparently should `loop.run_in_executor` that thing (according to https://stackoverflow.com/a/34376938)

Do you want to discuss it here or move to the issues of the aio_process_pool repo?

jeff-dh:
> Sorry @jeff-dh I am not sure if your `AsyncProcessPool` is fully "concurrent.futures.Executor compliant"

yeah, your right, there is at least one more (`wait=False` in `shutdown`). Atm it's compliant enough to be used as an executor in `asyncio.run_in_executor` -- at least in the nicegui context -- to be precise, right?

I'll think about it, probably rephrase it and / or add the missing features (as necessary).

>     * your `AsyncProcessPool` implementation returns a `<class '_asyncio.Task'>` which should be `await`ed.
>
>     * Meanwhile `ProcessPoolExecutor` returns a `<class 'concurrent.futures._base.Future'>` which you definitely can't `await` it (or `TypeError: object Future can't be used in 'await' expression`), rather you apparently should `loop.run_in_executor` that thing (according to https://stackoverflow.com/a/34376938)

oopos, didn't notice that... gonna have a look at it.

> Do you want to discuss it here or move to the issues of the aio_process_pool repo?

I guess belongs into `aio_process_pool` now and this will turn into a "added `run.set_cpu_bound_executor_type`" PR.

jeff-dh:
I guess we're pretty much ready to go, right?

@evnchn would you mind hiding the `aio_process_pool` related posts to clean this PR up? (feel free to create `aio_process_pool` issues)

jeff-dh:
>     * Is `assert isinstance(executor_type, Executor)` more robust?
it's not supposed to be an Executor instance, it's supposed to be an Executor `type`. As mentioned we could replace it with a factory which could probably be specified in more detail (return type). `assert issubclass(executor_type, Executor)` might also be a solution for this.

>     * Since `assert process_pool is None`, I can't use a custom executor if the default executor successfully loaded?

The default executor is loaded in `run.setup` (which gets called by the `lifespan` hook from `nicegui.App`). Therefor this function will and must be called before `setup` is executed (resp. `ui.run` which I assume triggers the hook), right? We simply exchange the type which gets instantiated (later on) in `setup`.

evnchn:
Interesting. I think my Python type / class knowledge is too shallow for this. But this should be open for review.

jeff-dh:
Think of it as a factory. A `type` -- which is (a kind of) a reference to it's constructor, right? -- is a factory that returns an instance of that type.

```python
class A:
    ...

class B(A):
    ...

factory = B
assert issubclass(factory, A)

instance = factory()
assert isinstance(instance, A)
```

jeff-dh:
(obviously) there's a monkey patch alternative

```python
def monkey_patch_nicegui_before_import():
    import nicegui.run, aio_process_pool
    nicegui.run.ProcessPoolExecutor = aio_process_pool.Executor

    def patched_shutdown():
        assert nicegui.run.process_pool
        nicegui.run.process_pool.shutdown(wait=True, cancel_futures=True)
        nicegui.run.thread_pool.shutdown(wait=False, cancel_futures=True)

    nicegui.run.tear_down = patched_shutdown

monkey_patch_nicegui_before_import()
```

rodja (changes requested):
Thank you @jeff-dh. The current implementation does not allow to pass arguments for creating the `process_pool`. It would be great to allow users to set their own executor, not just the type. For example to change the number of workers or replacements provided by [MPIRE](https://github.com/sybrenjansen/mpire), [Pebble](https://pebble.readthedocs.io/en/latest/), [Loky](https://github.com/joblib/loky) or [Deadpool](https://github.com/cjrh/deadpool).
As you showed in the monkey-patch version this could easily be archived by calling `nicegui.run.ProcessPoolExecutor = ...`. Could we use that instead of making the type configurable?

jeff-dh:
> The current implementation does not allow to pass arguments for creating the `process_pool`. It would be great to allow users to set their own executor, not just the type.

yep, I agree.

Would it be save with regards to 1b87221fc76cd62ae203133a83e20dc622028c95?

> As you showed in the monkey-patch version this could easily be archived by calling `nicegui.run.ProcessPoolExecutor = ...`. Could we use that instead of making the type configurable?

I not sure whether I understand this correctly: Use the monkey patch instead of the `run.py` change? Sure, I'm fine with it.

I'm wondering whether this monkey patch actually makes most sense:

```python

import nicegui.run, aio_process_pool, atexit

pool = aio_process_pool.ProcessPool()
atexit.register(pool.shutdown)

nicegui.run.cpu_bound = pool.run
# or use pool.run instead of run.cpu_bound
```

evnchn:
Hello @jeff-dh, still there? https://github.com/zauberzeug/nicegui/discussions/5574 shows the advantages of Loky.

evnchn:
@jeff-dh seems like you have deleted the nicegui fork so the PR auto-closes. We'll still credit you if we used any lines of code found here.

PR #4712, opened 2025-05-07
Interactive image layer

falkoschindler:
This PR allows adding layers to `ui.interactive_image`.

In some cases you might want to add different groups of SVG elements to an image. Maybe there is one element that needs frequent updates, while the other elements are rarely changed. Putting all elements in the same SVG can lead to performance issues, because the whole SVG needs to be sent to the client whenever one of the elements changes.

The solution is to add multiple layers to the image. Each layer is a separate SVG element, which means that each layer can be updated independently.

```py
def mouse_handler(e: events.MouseEventArguments):
    image.content += f'<circle cx="{e.image_x}" cy="{e.image_y}" r="30" fill="none" stroke="red" stroke-width="4" />'
    highlight.content = f'<circle cx="{e.image_x}" cy="{e.image_y}" r="28" fill="yellow" opacity="0.5" />'

image = ui.interactive_image('https://picsum.photos/id/674/640/360', on_mouse=mouse_handler, cross=True)
highlight = image.add_layer()
```

Screenshot 2025-05-07 at 15 56 36

### DIFF
--- a/nicegui/elements/interactive_image.py
+++ b/nicegui/elements/interactive_image.py
@@ -101,3 +101,12 @@ class InteractiveImage(SourceElement, ContentElement, component='interactive_ima
             return
         self._props['t'] = time.time()
         self.update()
+
+    def add_layer(self) -> InteractiveImage:
+        """Add a new layer with its own content.
+
+        *Added in version 2.17.0*
+        """
+        with self:
+            return InteractiveImage(source=self._props['src'], size=self._props['size']) \
+                .classes('nicegui-interactive-image-layer')
--- a/nicegui/elements/interactive_image.py
+++ b/nicegui/elements/interactive_image.py
@@ -102,11 +102,11 @@ class InteractiveImage(SourceElement, ContentElement, component='interactive_ima
         self._props['t'] = time.time()
         self.update()
 
-    def add_layer(self) -> InteractiveImage:
+    def add_layer(self, *, content: str = '') -> InteractiveImage:
         """Add a new layer with its own content.
 
         *Added in version 2.17.0*
         """
         with self:
-            return InteractiveImage(source=self._props['src'], size=self._props['size']) \
+            return InteractiveImage(self._props['src'], content=content, size=self._props['size']) \
                 .classes('nicegui-interactive-image-layer')
--- a/nicegui/static/nicegui.css
+++ b/nicegui/static/nicegui.css
@@ -263,8 +263,14 @@ h6.q-timeline__title {
 .nicegui-scene-view[data-initializing] {
   visibility: hidden;
 }
-
-/* CodeMirror */
+.nicegui-interactive-image-layer {
+  position: absolute !important;
+  inset: 0;
+  pointer-events: none;
+}
+.nicegui-interactive-image-layer img {
+  display: none;
+}
 .nicegui-codemirror .cm-editor.cm-focused {
   outline: none;
 }
--- a/tests/test_interactive_image.py
+++ b/tests/test_interactive_image.py
@@ -95,3 +95,14 @@ def test_loaded_event(screen: Screen):
     screen.click('Change Source')
     screen.should_contain('loaded')
     assert (screen.find_by_tag('img').get_attribute('src') or '').endswith(URL_PATH2)
+
+
+def test_add_layer(screen: Screen):
+    ii = ui.interactive_image(URL_PATH1, content='<rect x="0" y="0" width="100" height="100" fill="red" />')
+    ii.add_layer(content='<circle cx="100" cy="100" r="15" />')
+
+    screen.open('/')
+    screen.find_by_tag('svg')
+    with screen.implicitly_wait(0.5):
+        assert len(screen.find_all_by_tag('rect')) == 1
+        assert len(screen.find_all_by_tag('circle')) == 1
--- a/website/documentation/content/interactive_image_documentation.py
+++ b/website/documentation/content/interactive_image_documentation.py
@@ -16,6 +16,31 @@ def main_demo() -> None:
     ii = ui.interactive_image(src, on_mouse=mouse_handler, events=['mousedown', 'mouseup'], cross=True)
 
 
+@doc.demo('Adding layers', '''
+    In some cases you might want to add different groups of SVG elements to an image.
+    Maybe there is one element that needs frequent updates, while the other elements are rarely changed.
+    Putting all elements in the same SVG can lead to performance issues,
+    because the whole SVG needs to be sent to the client whenever one of the elements changes.
+
+    The solution is to add multiple layers to the image.
+    Each layer is a separate SVG element, which means that each layer can be updated independently.
+
+    The following demo shows this concept in action, even though both layers are changed at the same time.
+
+    *Added in version 2.17.0*
+''')
+def adding_layers():
+    from nicegui import events
+
+    def mouse_handler(e: events.MouseEventArguments):
+        image.content += f'<circle cx="{e.image_x}" cy="{e.image_y}" r="30" fill="none" stroke="red" stroke-width="4" />'
+        highlight.content = f'<circle cx="{e.image_x}" cy="{e.image_y}" r="28" fill="yellow" opacity="0.5" />'
+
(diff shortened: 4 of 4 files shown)


evnchn:
@falkoschindler interesting. I think I have finally figured out what you are trying to do:

- `nicegui-interactive-image-layer` is a special class which has the following behaviour:
  - "stack" on top of existing elements
  - block the image from displaying

Then, by `with self` and instantiate another `InteractiveImage`, then we basically "keep the interactive but drop the image" of the new `InteractiveImage`, allowing for the SVG layer you were asking for. 

But then, since the new layer is also `InteractiveImage`, which still inherits from `SourceElement`, so `highlight.set_source` still works. It just does nothing. 

```py


def mouse_handler(e: events.MouseEventArguments):
    image.content += f'<circle cx="{e.image_x}" cy="{e.image_y}" r="30" fill="none" stroke="red" stroke-width="4" />'
    highlight.content = f'<circle cx="{e.image_x}" cy="{e.image_y}" r="28" fill="yellow" opacity="0.5" />'


image = ui.interactive_image('https://picsum.photos/id/674/640/360', on_mouse=mouse_handler, cross=True)
ui.button('Set alternative image for "image"', on_click=lambda: image.set_source('https://picsum.photos/id/237/640/360'))
ui.button('Revert image for "image"', on_click=lambda: image.set_source('https://picsum.photos/id/674/640/360'))

highlight = image.add_layer()
ui.button('Set alternative image for "highlight"',
          on_click=lambda: highlight.set_source('https://picsum.photos/id/237/640/360'))
ui.button('Revert image for "highlight"', on_click=lambda: highlight.set_source('https://picsum.photos/id/674/640/360'))

ui.run()
```

{94831415-A860-4A29-A442-975F68261AD3}

This could lead to quite a bit of confusion. 

---

I am thinking whether we should somehow:

- Block the setting of source on an additional layer, making `highlight.set_source` error out
- Warn on attempt of setting of source on an additional layer. 
- `image.add_layer()` return something which does not support `.set_source` at all.

falkoschindler:
> I think I have finally figured out what you are trying to do

Yes, it's a common use case for our internal projects where something like a robot pose is updating constantly and other SVG content like an obstacle map almost never changes. From such projects I'm re-cycling the idea of stacking interactive images to get independent contents but with a precisely overlapping geometry. And in order to automatically set the correct `size`, I copy the parent source, which will be cached by the browser anyway, hide it via CSS, and let the child image derive its size from the "downloaded" image.

But your concern about layers being interactive images with "broken" methods is very valid. Basically all methods are pointless (at the moment): `set_source`, `on_mouse`, `force_reload`, `add_layer`. I tend to implement an internal `InteractiveImageLayer` class without any methods.

evnchn:
@falkoschindler maybe not work on that first, since I got another idea. 

Can it be possible, that the newly-added layers also support setting images? 

This would enable use cases, where we want to overlay a PNG image which has alignment markers, onto the underlying image. 

So, the newly added layers would not only just be for SVG, but fully-functional `InteractiveImage` instances?

evnchn:
But then if that's the case, then what is the difference between: 

```py
image = ui.interactive_image(...)
highlight1 = image.add_layer(...)
highlight2 = image.add_layer(...)
```

And:

```py
image = ui.interactive_image(...)
highlight1 = image.add_layer(...)
highlight2 = highlight1.add_layer(...)
```

Seems like no difference in behaviour, but then there's a completely dfferent hierachy inside going on? 

---

Hypothetical API design. Can this work? 

```py
with ui.interactive_image_stack():
    image = ui.interactive_image(...)
    highlight1 = ui.interactive_image(...)
    highlight2 = ui.interactive_image(...)
```

falkoschindler:
> what is the difference

Basically none.

> Hypothetical API design. Can this work?

Maybe. But I feel like this is less discoverable than `add_layer`. And the different images don't know that they should overlap exactly.

evnchn:
> But I feel like this is less discoverable than add_layer

But I think `add_layer` is abstracting away the fact that you are basically: 

```py
with ui.interactive_image(...) as image:
    highlight1 = ui.interactive_image(...)
    highlight2 = ui.interactive_image(...)
```

In fact, the above looks like pretty good API design. Maybe we don't need the "nicegui-interactive-image-layer" class, but the inner interactive image instances can "figure out they are inside another interactive image" using another way, and act correspondingly. 

About image with mismatched size if image support is enabled: I think it should fit in "contain" manner inside the first and outer most `interactive_image`.

falkoschindler:
> So, the newly added layers would not only just be for SVG, but fully-functional `InteractiveImage` instances?

This would be great. But I see two conceptional problems:

- What if layers have sources with different sizes? Should nested images stretch or overflow? Both is weird.
- What happens if you register event handlers via `on_mouse` on a nested image? If this would work, they would block events on the parent image.

To keep things reasonably simple for this PR, I still lean towards a dedicated `InteractiveImageLayer` with limited functionality (which can be extended in the future).

Anyway, I'm leaving this open for discussion for a while before finalizing the implementation.

evnchn:
> What happens if you register event handlers via on_mouse on a nested image? If this would work, they would block events on the parent image.

I personally don't want to endorse attaching event handler on the nested image, even if we can passthrough and trigger both the parent and the child. 

Thus, seems like `InteractiveImageLayer` is the way to go, since we fundamentally need different treatment between the parent `InteractiveImage` and the child `InteractiveImageLayer`. I don't think it is appropriate to return `InteractiveImage`, and have them figure out whether they are parent or they are child. 

Then, for this PR we can first enable just the SVG functionality first. We can figure out the image part, should we want to pursue it (put it in NiceGUI wishlist, then)

---

Still thinking whether we should do `.add_layer(...)` or use potentially:

```py
with ui.interactive_image(...) as image:
    highlight1 = ui.interactive_image_layer(...)
    highlight2 = ui.interactive_image_layer(...)
```

Latter seems more pythonic. After all, NiceGUI is all for making use of the context manager to add stuff inside some parent container.

falkoschindler:
> After all, NiceGUI is all for making use of the context manager to add stuff inside some parent container.

There are, however, counter-examples like Leaflet layers or 3D scene objects that can be added via "methods" on the parent element. Adding such children of a certain sub-type can be seen as a function of the parent. Or to put it that way: The children can't exist without a parent. And it keeps the `ui` namespaces tidy. Otherwise Leaflet maps, 3D scenes and SVG images might all want to claim `ui.circle`, or you need prefixes like `ui.leaflet_circle`, `ui.scene_circle` etc. Not nice.

falkoschindler:
Because the parent image could change size when loading a new source, layers should react accordingly. Luckily there is a "loaded" event which we can use together with a new "updateViewbox" method.

This is my test code:
```py
def mouse_handler(e: events.MouseEventArguments):
    circles.content += f'<circle cx="{e.image_x}" cy="{e.image_y}" r="30" fill="none" stroke="red" stroke-width="4" />'
    highlight.content = f'<circle cx="{e.image_x}" cy="{e.image_y}" r="28" fill="yellow" opacity="0.5" />'

def update_outlines(e: events.GenericEventArguments):
    w = e.args['width']
    h = e.args['height']
    image.set_content(f'<rect x="0" y="0" width="{w}" height="{h}" fill="none" stroke="blue" stroke-width="18" />')
    circles.set_content(f'<rect x="0" y="0" width="{w}" height="{h}" fill="none" stroke="red" stroke-width="12" />')
    highlight.set_content(f'<rect x="0" y="0" width="{w}" height="{h}" fill="none" stroke="yellow" stroke-width="6" />')

SRC_16_9 = 'https://picsum.photos/id/674/640/360'
SRC_3_4 = 'https://picsum.photos/id/674/480/640'
image = ui.interactive_image(SRC_16_9, on_mouse=mouse_handler, cross=True)
image.on('loaded', update_outlines)
circles = image.add_layer()
highlight = image.add_layer()

ui.button('16:9', on_click=lambda: image.set_source(SRC_16_9))
ui.button('3:4', on_click=lambda: image.set_source(SRC_3_4))
```

codingpaula (approved):
Nice! I tested this with our use case, works well. You can even change the order of the layers by adding them in order ;)

Issue #4468, opened 2025-03-12
When I set a class "grow-14" in the element classes, the tailwind css class is not generated


spacether:
### Description

When I set a class "grow-14" in the element classes, the tailwind css class is not generated

evnchn:
https://github.com/zauberzeug/nicegui/blob/main/DEPENDENCIES.md

NiceGUI so far uses v3 of Tailwind CSS.

![Image](https://github.com/user-attachments/assets/d252cd7e-280a-4dee-9ad2-2aea1f5ee96a)

First result if you search "tailwind flex grow" yields https://tailwindcss.com/docs/flex-grow, which, as of writing, links to the v4 documentation. 

_The website is a bit counter-intuitive_ in a sense that when you select v3.4.17 in the top-left, it redirects you to the _homepage_ of Tailwind CSS v3... 😡

But, you can manually go to https://v3.tailwindcss.com/docs/flex-grow

![Image](https://github.com/user-attachments/assets/38ead2ce-d3fc-4bef-987e-e8513ca871ba)

Getting back to how to do the equivalent of `grow-14`, at the end of the day, Tailwind CSS classes are just CSS styles under a trenchcoat. You can freely apply `.style()` to the elements. 

grow-<number> would be flex-grow: <number>; 

So, grow-14 would be flex-grow: 14;

@falkoschindler any plans to go to Tailwind CSS v4? Perhaps milestone 3.0?

falkoschindler:
@evnchn Yes, we're usually upgrading third-party dependencies during major releases. So Tailwind version 4 will probably come with NiceGUI 3.0. We don't have a release schedule yet, but given 2.0 came in August 2024, I guess some time this summer would make sense.

evnchn:
Should we link to the C3 Tailwind CSS page and also put some description at https://nicegui.io/documentation/section_styling_appearance ?

Since Tailwind CSS changed the `https://tailwindcss.com/` link to serve V4 content instead

falkoschindler:
Good idea. I've updated the links in d87ee220d018ece775fa02ba40933a47a3cdc711.

falkoschindler:
I'm closing this issue because this particular problem can be solved using "grow-[14]" instead of "grow-14".

Support for classes like "grow-14" depends on an upgrade to Tailwind 4 which is in discussion. We are, however, thinking about UnoCSS with a [Wind4 preset](https://unocss.dev/presets/wind4) as an alternative (PR #4832).

Issue #3854, opened 2024-10-10
AG-Grid's `'rowSelection': {'mode':"multiRow",'headerCheckbox':False}` won't work


python-and-fiction:
### Description

It is same as https://github.com/ag-grid/ag-grid/issues/8855
Following code won't work as expected.
```python3
from nicegui import ui

grid = ui.aggrid({
    'defaultColDef': {'flex': 1},
    'columnDefs': [
        {'headerName': 'Name', 'field': 'name'},
        {'headerName': 'Age', 'field': 'age'},
        {'headerName': 'Parent', 'field': 'parent', 'hide': True},
    ],
    'rowData': [
        {'name': 'Alice', 'age': 18, 'parent': 'David'},
        {'name': 'Bob', 'age': 21, 'parent': 'Eve'},
        {'name': 'Carol', 'age': 42, 'parent': 'Frank'},
    ],
    'rowSelection': {'mode':"multiRow",'headerCheckbox':False},
    
}).classes('max-h-40')

ui.run(native=True)
```
![E5NCGJC5{9{3(KOODE P52M](https://github.com/user-attachments/assets/a84000b0-d1b5-44c9-9871-2aeaa9d0c19f)

When  i upgrade ag-grid to 32.2.1, it works well.
![6 D5EYT52YGFI {~1VJG@4P](https://github.com/user-attachments/assets/dac37007-83bf-4a8d-b77d-33eed4d303bf)

falkoschindler:
Thanks for reporting this issue, @python-and-fiction!
It looks like we would need to upgrade from our current AG Grid version 32.1.0 to at least 32.2.1 or the most recent 32.2.2. Usually we only upgrade dependencies for major releases. But I guess going from 32.1 to 32.2 shouldn't introduce breaking changes.

falkoschindler:
Unfortunately, upgrading from 32.1 to 32.2 breaks existing behavior, e.g. calling the "selectAll" method doesn't work without specifying the "rowSelection" option.

@python-and-fiction Is it possible that the feature you're describing didn't exist in AG Grid <=32.2? I just tried your code with NiceGUI 1.4.37 and it doesn't work either. In this case, this issue isn't a bug but a feature request. And since it depends on upgrading third-party libraries, it would need to wait for NiceGUI 3.0.

mockey:
I was playing around with aggrid and found this issue.
@falkoschindler Did you try with a current version?
There have been changes in 33.0.0:
[...]

PR #4718, opened 2025-05-08
Feat: Override dependencies at lib/

evnchn:
This PR allows the user to set `NICEGUI_LIB_OVERRIDE_DIRECTORY` in env to a directory, for which it contains what _would have been_ in NiceGUI's `lib/` folder. 

This way, the user can override NiceGUI's dependencies with newer ones, skipping forward in the release cycle. 

Combined with a beta feedback system, we can collect information on whether a new version of JS dependency will work with NiceGUI or not, potentially easin upcoming releases, and empowering us to move faster 🚀

Demo code:

```py
from nicegui import ui
import os
from pathlib import Path
os.environ['NICEGUI_LIB_OVERRIDE_DIRECTORY'] = str(Path(__file__).parent / 'lib_override')

grid = ui.aggrid({
    'defaultColDef': {'flex': 1},
    'columnDefs': [
    ],
    'rowData': [
        {'name': 'Alice', 'age': 18},
        {'name': 'Bob', 'age': 21},
        {'name': 'Carol', 'age': 42},
    ],
    'rowSelection': 'multiple',
}).classes('max-h-40')

ui.run(reload=True)
```

Ensure in same folder as your test script, `lib_override/lib/aggrid/ag-grid-community.min.js` is accessible. 

Without override:

{A4504147-9A07-487E-8D34-3A899EDBEBE7}

With override: 

{BD7CE54B-AA7E-4C59-8181-D49892436680}

Notice the new styling. 

Remaining tasks: 

- [ ] Should we add this?
- [ ] Where should we place the option? Is environment variables a good idea? 
- [ ] Why it doesn't work when I try and shove the option into core.app.config? 
- [ ] Why the environment variable is ignored when I do `reload=False`?
- [ ] Documentation?
- [ ] Underlying framework to receive beta feedback?

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import inspect
+import os
 import re
 from copy import copy
 from pathlib import Path
@@ -93,8 +94,16 @@ class Element(Visibility):
         base = Path(inspect.getfile(cls)).parent
 
         def glob_absolute_paths(file: Union[str, Path]) -> List[Path]:
+            LIB_OVERRIDE_DIRECTORY = Path(os.getenv('NICEGUI_LIB_OVERRIDE_DIRECTORY', ''))
             path = Path(file)
             if not path.is_absolute():
+                if LIB_OVERRIDE_DIRECTORY and str(path).startswith('lib'):
+                    path_override = LIB_OVERRIDE_DIRECTORY / path
+                    if path_override.exists():
+                        path = path_override
+                        print(f'Using override path: {path}')
+                    else:
+                        print(f'Override path does not exist: {path_override}')
                 path = base / path
             return sorted(path.parent.glob(path.name), key=lambda p: p.stem)
 
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -94,16 +94,13 @@ class Element(Visibility):
         base = Path(inspect.getfile(cls)).parent
 
         def glob_absolute_paths(file: Union[str, Path]) -> List[Path]:
-            LIB_OVERRIDE_DIRECTORY = Path(os.getenv('NICEGUI_LIB_OVERRIDE_DIRECTORY', ''))
             path = Path(file)
             if not path.is_absolute():
-                if LIB_OVERRIDE_DIRECTORY and str(path).startswith('lib'):
-                    path_override = LIB_OVERRIDE_DIRECTORY / path
-                    if path_override.exists():
-                        path = path_override
-                        print(f'Using override path: {path}')
-                    else:
-                        print(f'Override path does not exist: {path_override}')
+                lib_path = os.getenv('NICEGUI_LIB_OVERRIDE_DIRECTORY')
+                if lib_path and path.parts[0] == 'lib':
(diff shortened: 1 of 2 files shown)


evnchn:
If implemented, would help with #3854, and possibly #4468

falkoschindler (commented):
I quickly reviewed this pull request, which could be useful in certain situations. But I'm undecided about the details yet.

evnchn:
Me too. 

I think setting via environment variable is a bit jank and unreliable, especially since it doesn't work if you:

```py
from dotenv import load_dotenv
load_dotenv()
```

Or

```py
import os
os.environ['NICEGUI_LIB_OVERRIDE_DIRECTORY'] = 'blahblahblah'
```

You seem to MUST set it via command line before even launching the Python script for it to work. 

---

However, overriding lib is still desirable. While we're at it, maybe even allow overriding the Tailwind to V4. Those beta testers (i.e. me and a few brave people) can speed things up when we roll around to implementing Tailwind V4 in NiceGUI for real.

rodja:
I'm not sure an env variable is very handy. One would need to have all the NiceGUI code in that dir. Why not simply copy the lib you want to test/update in the pip installed location. With venv, Docker etc this is well encapsulated.

evnchn:
Fair point. 

I think for alpha-level testing, overriding the JS files directly at `Lib/site-packages/nicegui/elements/lib` provided the use of an isolated environment like venv or Docker makes sense. 

Besides a slightly better way to keep track of the old copy, and being able to do this in non-venv situations, this PR doesn't seem to add much to the table. Alpha-testers should be able to back their files up and handle a venv anyways. 

So, I'm more inclined to closing this PR in favor of a documentation update somewhere in `CONTRIBUTING.md` or somewhere in the NiceGUI website (though former seems more plausible since this is alpha-level thing, and I don't feel comfortable endorsing writing to `Lib/site-packages/nicegui/elements/lib` out loud)

---

However, for beta-level testing, perhaps a more streamlined experience is still desirable, which this PR does not provide so far. 

Wishful programming:

```py
from nicegui.beta import dependency_override 
# put in nicegui.beta, and we declare we may have breaking changes in minor versions, so that we move quickly
# could be an optional package, or external package

dependency_override('aggrid': '33.3.0')
dependency_override('leaflet': 'latest')
dependency_override('plotly': '/path/to/plotly/js')
```

It is best, if the function can be called after `ui.run` is called, such that we can live-swap dependencies in-and-out. 

```py
ui.input('Enter AG Grid version or path to test', lambda e: dependency_override('aggrid', e.args['value'])
```

**To avoid shifting the topic of the conversation here, whoever thinks this is useful, if there isn't a thread already, please open a thread in https://github.com/zauberzeug/nicegui/discussions/categories/ideas-feature-requests and comment the link to the new thread below to restart the conversation.**

evnchn:
Apparently swapping dependencies is as easy as https://github.com/zauberzeug/nicegui/discussions/4739#discussioncomment-13141055

So there's no need for this override lib functionality for elements. 

Only though remaining is if we can let user swap out Vue, Quasar and Tailwind (<- This one most tricky since v4 has breaking changes)

Issue #1430, opened 2023-08-16
Getting a "connection lost" error when trying to use the file picker in native mode


varchasgopalaswamy:
### Description

I'm trying to use the example case provided in https://github.com/zauberzeug/nicegui/discussions/283

```python
from nicegui import app, ui

async def choose_file():
    files = await app.native.main_window.create_file_dialog(allow_multiple=True)
    for file in files:
        ui.notify(file)

ui.button('choose file', on_click=choose_file)

ui.run(native=True)

```

but when I try to run this, I get a message "Connection lost", and the file browser doesn't show up.

rodja:
The posted code is working for me. Maybe its an operating system issue? I'm on Mac OS with Chrome.

varchasgopalaswamy:
Interesting. I'm on RHEL 9 and using GTK for Webview. I found QT did not work (I was getting a white blank screen). I am able to launch the file browser using Webview, e.g. 

```python 

import webview


def open_file_dialog(window):
    file_types = ('Image Files (*.bmp;*.jpg;*.gif)', 'All files (*.*)')

    result = window.create_file_dialog(webview.OPEN_DIALOG, allow_multiple=True, file_types=file_types)
    print(result)


if __name__ == '__main__':
    window = webview.create_window('Open file dialog example', 'https://pywebview.flowrl.com/hello')
    webview.start(open_file_dialog, window)


```

so I guess it's not an issue with Webview?

varchasgopalaswamy:
I turned on uvicorn_debug and got 

`DEBUG:    < TEXT '42["event",{"id":12,"listener_id":"a96b632f-5e0...3cfa3ff0eb","args":[]}]' [84 bytes]`

when I clicked on the "open file" dialog. I guess it's not able to get a connection to something since there's no reply?

rodja:
It may be some code [is blocking the main thread](https://github.com/zauberzeug/nicegui/wiki/FAQs#how-to-not-get-the-reconnect-popup-or-avoid-laggy-interaction). Can anyone on Linux with GTK confirm this issue?

falkoschindler:
@NiklasNeugebauer, @JensOgorek, I already tried to reproduce this issue with @denniswittich, but we couldn't get GTK running. Could one of you, please, try to run the code from the [original post](https://github.com/zauberzeug/nicegui/issues/1430#issue-1853341094) and see if the file dialog works? Thanks!

JensOgorek:
[...]

PR #4719, opened 2025-05-08
Project BASE - Faster HTTP Reconnect

evnchn:
This PR introduces a HTTP-based special mechanism for loading in the page which is must faster than the browser. 

#### Brief explainer

The server, upon the setting of a special private header `'X-NiceGUI-Client-ID'`, instead of serving HTML, _proceeds instead_ to serve it in JSON machine-readable format. 

Which, the special JavaScript function `softReload` will:

- Disconnect the old Socket.IO
- Fetch with special header and parse JSON
- Re-run `createApp(...)`
- Dynamically import any JS modules and add them to add component
- Re-execute `vue_script` for the Plotly and similar elements which rely on it
- Mount the app to `#app`

Essentially, it does what a page load will do, but without a page load. 

---

#### Name meaning of BASE

BASE stands for "Because Asked Server Explicitly", which is an (admittedly not very good) acryonym reflecting on the nature that the client asks the server explicitly for reconnection advice via HTTP, including the crucial Client ID. 

This contrasts with project ACID #4552 before it, which the client can say it had any arbitrary Client ID via Socket.IO, and the server must bend around it by then suddenly spawning in the Client, bypassing HTTP layer entirely. 

---

#### Advantages

Advantages shared common between ACID and BASE: 

- Fast page switching (implemented via softReload) 
- Zero Flash Of Unstyled Content as a baseline guarantee. 
- Fast reconnection in case client destroyed, because disconnected over the timeout duration / server reboot (to be implemented)

Advantages exclusive to BASE: 

- Maintains the one-to-one relationship between HTTP request and NiceGUI Client instances
- Ensures Middlewares stay effective
- Avoids fallacies such as HTTP GET parameters being lost
- Practically zero-impact on userspace code, since server-side all new logic is basically not involved if `'X-NiceGUI-Client-ID'` is not set in header
- May make creating custom NiceGUI frontends (like an ESP32 touchscreen UI) slightly easier, though not by much. 

---

#### Remaining tasks

- [ ] Assess benefit VS impact
- [ ] Finish off the fast reconnection in client destroyed / server reboot situation
- [ ] `softReload` doesn't set browser URL bar
- [ ] Extensive testing (non-negotiable)
- [ ] Extensive documentation 
- [ ] Drop unused fields from the JSONResponse if they are not needed by the reload process
- [ ] Can we make the typical page load in code and the `softReload` code to be unified? This can further ensure exact same behaviour.  

---

#### Testing code

<details>
<summary> My testing code </summary>

```py
from nicegui import ui
from random import random


def menu():
    with ui.row():
        ui.label('Menu').classes('text-2xl text-gray-700')
        ui.button('Page 1', on_click=lambda: ui.navigate.to('/page1'))
        ui.button('Page 2', on_click=lambda: ui.navigate.to('/page2'))
        ui.button('Page 3', on_click=lambda: ui.navigate.to('/page3'))
        ui.button('Page 4', on_click=lambda: ui.navigate.to('/page4'))
        ui.button('Page 5', on_click=lambda: ui.navigate.to('/page5'))
    with ui.row():
        ui.button('Soft Page 1', on_click=lambda: ui.run_javascript('softReload("/page1")'))
        ui.button('Soft Page 2', on_click=lambda: ui.run_javascript('softReload("/page2")'))
        ui.button('Soft Page 3', on_click=lambda: ui.run_javascript('softReload("/page3")'))
        ui.button('Soft Page 4', on_click=lambda: ui.run_javascript('softReload("/page4")'))
        ui.button('Soft Page 5', on_click=lambda: ui.run_javascript('softReload("/page5")'))


@ui.page('/page1')
def page1():
    menu()
    ui.label('Page 1').classes('text-2xl text-red-500')


@ui.page('/page2')
def page2():
    menu()
    ui.label('Page 2').classes('text-2xl text-blue-500')


@ui.page('/page3')
def page3():
    menu()
    with ui.card():
        ui.mermaid('''
            graph TD
            A --> B
            A --> C
            A --> D
        ''')

# page4 with plotly


@ui.page('/page4')
def page4():
    menu()
    fig = go.Figure()
    fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
    plot = ui.plotly(fig).classes('w-full h-40')

    def add_trace():
        fig.add_trace(go.Scatter(x=[1, 2, 3], y=[random(), random(), random()]))
        plot.update()

    ui.button('Add trace', on_click=add_trace)

# page5 with dark mode


@ui.page('/page5')
def page5():
    menu()
    dark = ui.dark_mode()
    ui.switch('Dark mode').bind_value(dark)


menu()

ui.run()
```

</details>

Results: 

- Switching through all 5 pages work seamlessly, despite starting on whichever page. 
- Since Mermaid loaded in while starting from page 1, loading ES modules works. 
- Since Plotly loaded in while starting from page 1, loading `vue_scripts` works.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -10,7 +10,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, ClassVar, Dict, Iterable, Iterator, List, Optional, Union
 
 from fastapi import Request
-from fastapi.responses import Response
+from fastapi.responses import JSONResponse, Response
 from fastapi.templating import Jinja2Templates
 from typing_extensions import Self
 
@@ -124,7 +124,7 @@ class Client:
     def __exit__(self, *_) -> None:
         self.content.__exit__()
 
-    def build_response(self, request: Request, status_code: int = 200) -> Response:
+    def build_response(self, request: Request, status_code: int = 200, *, send_json: bool = False) -> Response:
         """Build a FastAPI response for the client."""
         self.outbox.updates.clear()
         prefix = request.headers.get('X-Forwarded-Prefix', request.scope.get('root_path', ''))
@@ -138,37 +138,44 @@ class Client:
         }
         vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls = \
             generate_resources(prefix, self.elements.values())
+        context = {
+            'version': __version__,
+            'elements': elements.replace('&', '&amp;')
+            .replace('<', '&lt;')
+            .replace('>', '&gt;')
+            .replace('`', '&#96;')
+            .replace('$', '&#36;'),
+            'head_html': self.head_html,
+            'body_html': '<style>' + '\n'.join(vue_styles) + '</style>\n' + self.body_html + '\n' + '\n'.join(vue_html),
+            'vue_scripts': '\n'.join(vue_scripts),
+            'imports': json.dumps(imports),
+            'js_imports': '\n'.join(js_imports),
+            'js_imports_urls': js_imports_urls,
+            'quasar_config': json.dumps(core.app.config.quasar_config),
+            'title': self.resolve_title(),
+            'viewport': self.page.resolve_viewport(),
+            'favicon_url': get_favicon_url(self.page, prefix),
+            'dark': str(self.page.resolve_dark()),
+            'language': self.page.resolve_language(),
+            'translations': translations.get(self.page.resolve_language(), translations['en-US']),
+            'prefix': prefix,
+            'tailwind': core.app.config.tailwind,
+            'prod_js': core.app.config.prod_js,
+            'socket_io_js_query_params': socket_io_js_query_params,
+            'socket_io_js_extra_headers': core.app.config.socket_io_js_extra_headers,
+            'socket_io_js_transports': core.app.config.socket_io_js_transports,
+        }
+        if send_json:
+            return JSONResponse(
+                content=context,
+                status_code=status_code,
+                headers={'Cache-Control': 'no-store', 'X-NiceGUI-Content': 'jsonpage'},
+            )
+        context['request'] = request
         return templates.TemplateResponse(
             request=request,
             name='index.html',
-            context={
-                'request': request,
-                'version': __version__,
-                'elements': elements.replace('&', '&amp;')
-                                    .replace('<', '&lt;')
-                                    .replace('>', '&gt;')
-                                    .replace('`', '&#96;')
-                                    .replace('$', '&#36;'),
-                'head_html': self.head_html,
-                'body_html': '<style>' + '\n'.join(vue_styles) + '</style>\n' + self.body_html + '\n' + '\n'.join(vue_html),
-                'vue_scripts': '\n'.join(vue_scripts),
-                'imports': json.dumps(imports),
-                'js_imports': '\n'.join(js_imports),
-                'js_imports_urls': js_imports_urls,
-                'quasar_config': json.dumps(core.app.config.quasar_config),
-                'title': self.resolve_title(),
-                'viewport': self.page.resolve_viewport(),
-                'favicon_url': get_favicon_url(self.page, prefix),
-                'dark': str(self.page.resolve_dark()),
-                'language': self.page.resolve_language(),
-                'translations': translations.get(self.page.resolve_language(), translations['en-US']),
-                'prefix': prefix,
-                'tailwind': core.app.config.tailwind,
-                'prod_js': core.app.config.prod_js,
-                'socket_io_js_query_params': socket_io_js_query_params,
-                'socket_io_js_extra_headers': core.app.config.socket_io_js_extra_headers,
-                'socket_io_js_transports': core.app.config.socket_io_js_transports,
-            },
+            context=context,
             status_code=status_code,
             headers={'Cache-Control': 'no-store', 'X-NiceGUI-Content': 'page'},
         )
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -136,7 +136,7 @@ class Client:
             'client_id': self.id,
             'next_message_id': self.outbox.next_message_id,
         }
-        vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls = \
+        vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls, js_imports_raw = \
             generate_resources(prefix, self.elements.values())
         context = {
             'version': __version__,
@@ -166,6 +166,7 @@ class Client:
             'socket_io_js_transports': core.app.config.socket_io_js_transports,
         }
         if send_json:
+            context['js_import_raw'] = js_imports_raw
             return JSONResponse(
                 content=context,
                 status_code=status_code,
--- a/nicegui/dependencies.py
+++ b/nicegui/dependencies.py
@@ -141,6 +141,7 @@ def generate_resources(prefix: str, elements: Iterable[Element]) -> Tuple[List[s
     imports: Dict[str, str] = {}
     js_imports: List[str] = []
     js_imports_urls: List[str] = []
+    js_imports_raw: Dict[str, Dict[str, str]] = {}
 
     # build the importmap structure for exposed libraries
     for key, library in libraries.items():
@@ -172,6 +173,10 @@ def generate_resources(prefix: str, elements: Iterable[Element]) -> Tuple[List[s
                 url = f'{prefix}/_nicegui/{__version__}/components/{js_component.key}'
                 js_imports.append(f'import {{ default as {js_component.name} }} from "{url}";')
                 js_imports.append(f'app.component("{js_component.tag}", {js_component.name});')
+                js_imports_raw[js_component.name] = {
+                    'tag': js_component.tag,
+                    'url': url,
+                }
                 js_imports_urls.append(url)
                 done_components.add(js_component.key)
-    return vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls
+    return vue_html, vue_styles, vue_scripts, imports, js_imports, js_imports_urls, js_imports_raw
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -112,6 +112,7 @@ class page:
         @wraps(func)
         async def decorated(*dec_args, **dec_kwargs) -> Response:
             request = dec_kwargs['request']
+            special_header_set_in_request = 'X-NiceGUI-Client-ID' in request.headers
             # NOTE cleaning up the keyword args so the signature is consistent with "func" again
             dec_kwargs = {k: v for k, v in dec_kwargs.items() if k in parameters_of_decorated_func}
             with Client(self, request=request) as client:
@@ -141,6 +142,8 @@ class page:
             if isinstance(result, Response):  # NOTE if setup returns a response, we don't need to render the page
                 return result
             binding._refresh_step()  # pylint: disable=protected-access
+            if special_header_set_in_request:
+                return client.build_response(request, send_json=True)
             return client.build_response(request)
 
         parameters = [p for p in inspect.signature(func).parameters.values() if p.name != 'client']
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -447,3 +447,36 @@ for (let sheet of document.styleSheets) {
     }
   }
 }
+
+function softReload(url) {
+  // Make the GET request
+  window.socket.disconnect();
+  fetch(url, {
+    method: 'GET',
+    headers: new Headers({
+      'X-NiceGUI-Client-ID': 'your-client-id' // Replace with your actual client ID
+    })
+  })
+    .then(response => {
+      if (!response.ok) {
+        throw new Error('Network response was not ok')
+      }
+      return response.json()
+    })
+    .then(data => {
+      console.log(data)
+      // Handle the response data
+      createApp(parseElements(data.elements), {
+        version: data.version,
+        prefix: data.prefix,
+        query: data.socket_io_js_query_params,
+        extraHeaders: data.socket_io_js_extra_headers,
+        transports: data.socket_io_js_transports,
+        quasarConfig: JSON.parse(data.quasar_config),
+      })
+      app.mount("#app")
+    })
+    .catch(error => {
+      console.error('There was a problem with the fetch operation:', error)
+    })
+}
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -473,8 +473,19 @@ function softReload(url) {
         extraHeaders: data.socket_io_js_extra_headers,
         transports: data.socket_io_js_transports,
         quasarConfig: JSON.parse(data.quasar_config),
-      })
-      app.mount("#app")
+      });
+      const importPromises = Object.entries(data.js_import_raw).map(([name, value]) => {
+        console.log(`Importing element ${name}`);
+        return import(value.url).then((module) => {
+          console.log(`Imported element ${name} from ${value.url}`, module);
+          app.component(value.tag, module.default);
+        });
+      });
+
+      Promise.all(importPromises).then(() => {
+        eval(data.vue_scripts); // required for plotly and some libraries
+        app.mount("#app");
+      });
     })
     .catch(error => {
       console.error('There was a problem with the fetch operation:', error)


falkoschindler:
Very interesting idea, @evnchn! Seems to be a comparatively small change with a big impact. 👌🏻

evnchn:
Appears to work fine for the most part. 

However, QTable not showing up, if you visit, in this sequence: 

http://127.0.0.1:8080/documentation
http://127.0.0.1:8080/documentation/section_data_elements
http://127.0.0.1:8080/documentation/table

evnchn:
https://github.com/zauberzeug/nicegui/pull/4719#issuecomment-2869129979 addressed with https://github.com/zauberzeug/nicegui/pull/4719/commits/f955e429d9504fcb2084e3de284d26b1802e0856

evnchn:
So @rodja and I was discussing how this PR relates to Single-Page Application in another thread. 

As a recap, the definition of a Single-Page Application is: 

1. Only one browser load
2. Capture browser's pushstate/popstate
3. Load in only the updated content

For this PR it is itself mostly oriented towards solving point 1, since as you know the single-megabit uplink **hate** page loads. However, as we have seen in `main.py` (the NiceGUI documentation), Project BASE is handling point 2 with just a bit of extra JavaScript. Check below: 

https://github.com/zauberzeug/nicegui/blob/fa34e466257e5a6bd4b9f1c7d26c010e8423b7c8/main.py#L64-L91

Point 3 is real, though, and I should have thought of that as the "performance-focused guy" on the team 😅. Anyways, I think it is doable, if we: 
1. Have the user cache all elements repeatedly sent via #4796, or 
2. Similar to the cookie approach in #4796, we set the cookie indicating what `Client` is associated with ourselves right now. Such that, when we spawn the new `Client`, we can compare to the old `Client` and deduplicate the elements if it's clearly the same between the pages. 

Point 2 elaboration: Proposed elements JSON string: 

`{"0":"SAME", "1":"SAME", ... "39": {"tag":"nicegui-link","text":"New Link...",... }`

rodja:
I'm not sure if a combination of this PR with a UI element cache can provide a sufficient SPA implementation. From my point of view, a NiceGUI SPA is not only defined by "Load in only the updated content" but also "Not rerun backend code". And to generate a JSON string like `{"0":"SAME", "1":"SAME", ... "39": {"tag":"nicegui-link","text":"New Link...",... }` you would need to re-execute the whole page function including headers etc for every sub-page-change.

evnchn:
> From my point of view, a NiceGUI SPA is not only defined by "Load in only the updated content" but also "Not rerun backend code"

I'd like to disagree by pointing out a half-baked (IMO) SPA, namely GitHub itself. 

Ever noticed, when the number of PRs has gone up, that some old tabs stays on the old PR count, until you reload or go inside and outside of NiceGUI repository? I sure did, as the source of those PRs 😅

This is because, GitHub does not re-execute the backend code responsible to calculating how many PRs are there and update the header, while just pulling in the body. 

But this could work for this PR: 

```py
def get_number_of_PRs():
    # some function

def common_header():
    ui.label(f'Number of PRs: {get_number_of_PRs()}')

@ui.page('/page1')
def page1():
    common_header()
    # content

@ui.page('/page2')
def page2():
    common_header()
    # content
```

Every single time you re-navigate using soft reload, it re-executes the `common_header`, and instructing the client to update with the new header if changed, or use the old one if not changed. 

---

In contrast, assuming #4821 syntax, this would keep the stale PR count on navigation: 

```py
def get_number_of_PRs():
    # some function

def common_header():
    ui.label(f'Number of PRs: {get_number_of_PRs()}')

@ui.page('/')
@ui.page('/{_:path}') # tells FastAPI/Starlette that all path's should land on our SPA
def index():
    common_header()
    ui.sub_pages({'/': child, '/other': child2})

def child():
    ui.link('goto other', '/other')

def child2():
    ui.link('goto main', '/')
```

Rather, to change the common header, we need to actively pass the elements which may change into the subpage, as prescribed in https://github.com/zauberzeug/nicegui/pull/4821#issue-3106729591

```py
@ui.page('/')
@ui.page('/{_:path}')
def index():
    title = ui.label()
    ui.sub_pages({
        '/': lambda: child(title),
        '/other': lambda: child2(title)
    })

def child(title: ui.label):
    title.set_text('child title')
    ui.link('goto other', '/other')

def child2(title: ui.label):
    title.set_text('other title')
    ui.link('goto main', '/')
```

---

Overall, I am inclined to believe, if implemented well, this PR can be a better implementation to SPA than #4821.

evnchn:
@rodja check this out https://github.com/zauberzeug/nicegui/pull/4719/commits/18f83e743ad381a0ae78cdee250c2fcff75d1a7a

```py
from nicegui import ui


def reused_ui_elements():
    ui.label('This is a label that can be reused.')
    ui.table(rows=[{'id': 1, 'name': 'Alice'}, {'id': 2, 'name': 'Bob'}])
    ui.markdown('#### This is a markdown element that can be reused.')
    ui.link('Go to another page', '/another')


@ui.page('/')
def main_page():
    ui.label('This is the main page.')
    reused_ui_elements()


@ui.page('/another')
def another_page():
    ui.label('This is another page.')
    reused_ui_elements()

# Copy from main.py
ui.add_body_html('''<script>
    // Your custom function
    function customFunction(event) {
        let target = event.target;
        // Traverse up the DOM tree to find the nearest anchor tag
        while (target && target.tagName !== 'A') {
            target = target.parentElement;
        }
        if (target && target.tagName === 'A') {
            console.log('Link clicked!'); // Replace with your custom logic
            console.log(target.href); // Log the href of the clicked link
            try {
                window.softReload(target.href);
                event.preventDefault();
            } catch (e) {
                console.error('Error in softReload:', e);
            }
        }
    }
    // Attach the event listener to the document
    document.addEventListener('click', customFunction);

    window.addEventListener("popstate", (event) => {
        console.log("popstate", event.state);
        softReload(window.location.href, event.state?.x, event.state?.y);
    });
    </script>
''', shared=True)


ui.run()
```

Betwen page soft reloads, it detects that the elements in `reused_ui_elements()` did not change. 

As such, Project REBASE (name loosely inspired from `git rebase`) short-formed it to `{"0":{"=":true}, ...}`, reducing data transmission. If there were any changes, it would send them over. 

{751F8ADE-FCE0-4BCD-8C79-F4C00538CFA8}

Note that the table is re-transmitted, because I did not fully steal the logic from #4796. If I did, then it would detect that the table, only the event handlers are changed, and pass something like the below (transmitting only the new event handlers):

```'{"6":{"=": true, "events":[{"listener_id":"d8134f52-fa89-4232-a9be-5f34f5636162","type":"selection","specials":[],"modifiers":[],"keys":[],"args":[["added","rows","keys"]],"throttle":0,"leading_events":true,"trailing_events":true,"js_handler":null},{"listener_id":"89168e4d-90d2-4d29-842c-af6ae29ffa9d","type":"update:pagination","specials":[],"modifiers":[],"keys":[],"args":null,"throttle":0,"leading_events":true,"trailing_events":true,"js_handler":null}]}}'```

Seems like this is a much easier approach over SPA, with NiceGUI taking up the heavy lifting of finding identical elements between pages, instead of user having to divide their content between the main page and `ui.sub_page`.

evnchn:
The logic in https://github.com/zauberzeug/nicegui/pull/4719/commits/21c2fad85183731b9c8e7c50c1dc4766070267e7 is: 

- Server-side, mark via `'=': True`, such that `replaceUndefinedAttributes` base off its response from `mounted_app.elements[id]`
- Server-side, for keys in the current element, we add on top, if it is different to the past element
- Client-side, `replaceUndefinedAttributes` will take the past element, add on top the changed keys from the previous step
- Client-side, we update the page, while sending minimal data across.

rodja:
I'm really fond of the friendly competition here, @evnchn. It is awesome how you transformed this "Faster HTTP Reconnect" into a SPA concept 😃

For me it is not an either #4821 or this PR. They server both quite different purposes. Even if this PR might be a the groundwork for an "SPA behind-the-scenes", at the beginning it was so neat and tight. Pushing the concept towards SPA might be better off in a follow up pull request. I just tested the implementation and directly ran into errors where the page switch was broken:

Jun-04-2025 17-34-09

rodja:
About your comment about [full-page reevaluation vs partial execution](https://github.com/zauberzeug/nicegui/pull/4719#issuecomment-2929948735):

> Every single time you re-navigate using soft reload, it re-executes [everything]

While this can be nice in some circumstances, it can also be very problematic in others. For example if you have a costly database request in the outer layout. Or a search filter which should not be reset. In short: re-execution of the whole page might be costly. It depends on the scenario. I suggest to 

1. get "faster HTTP Reconnect" out of the door (eg. revert this PR to 2549b1c, before adding push/popstate)
2. hone and polish `ui.sub_pages`
3. (maybe) optimize page loads by building upon "faster HTTP Reconnect" and adding push/pop state + sending only the changed parts

evnchn:
First of all, this PR is evolving so quickly! I was cleaning my room, and it took me a solid 15 minutes to re-familarize myself with this PR. Now, we need a summary. 

- Originally, Project BASE (Because Asked Server Explicitly) was designed to target #4552 (Project ACID) which came before it, with the selling points of:
  - Fast page switching (implemented via softReload)
  - Zero Flash Of Unstyled Content as a baseline guarantee.
  - Fast reconnection in case client destroyed, because disconnected over the timeout duration / server reboot (STILL to be implemented 😅)
- But then, the "Fast page switching" part happens to coincide with the Single-Page Application thing we are trying to make
  - Since I tapped into pushstate and popstate in https://github.com/zauberzeug/nicegui/pull/4719/commits/783cc34736900ceead3f4f26e6f84c28d9b99bb5
- The concern of repeatedly sending constant elements between pages came, and Project REBASE was born, to deduplicate elements sent, so as to leverage the elements already in memory
  - First generation, only elements which are an exact match would be deduplicated. https://github.com/zauberzeug/nicegui/pull/4719/commits/18f83e743ad381a0ae78cdee250c2fcff75d1a7a
  - Second generation, it was done on a key by key basis. https://github.com/zauberzeug/nicegui/commit/21c2fad85183731b9c8e7c50c1dc4766070267e7

---

@rodja, I am so sorry for shipping broken commit in https://github.com/zauberzeug/nicegui/commit/21c2fad85183731b9c8e7c50c1dc4766070267e7, leading to a general bad impression of Project BASE. 

I was too happy at looking at my key-diffing code serving small payloads, that the page was broken and I did not realize. 

Apparently, (1) there must have been some problem with the logic which diffs the elements on a key-by-key basis, in a sense that old keys are not dropped, and (2) even if I fixed it, some elements are still messed up, since Vue doesn't force a re-render, and so the old element sticks around. 

Check the below, where I made sure to manually compare the elements: 

{0885BB78-71F9-4C16-B7FC-78A9C66610A1}

JSON-wise, key 97 is fine. But, c97 is still messed up 

{16EF4BCB-717D-47BF-AC57-DAE4F1D65508}

So, we really can't proceed with key-level diffing, unless we dig deeper into Vue, which we need to get #4828 done first, in order to be in a better headspace to solve these kinds of issue. 

But, regardless, assuming we do the commit before that, we still get project REBASE with SPA-like functionality.  

---

For your suggestions: 

1. Get "faster HTTP Reconnect" out of the door (eg. revert this PR to https://github.com/zauberzeug/nicegui/commit/2549b1cb47482ee7f425ff0558159e6f7d38c8ff, before adding push/popstate)

I would then need to achieve the "Fast reconnection in case client destroyed, because disconnected over the timeout duration / server reboot (STILL to be implemented 😅)" which I have dragged on. 

2. hone and polish `ui.sub_pages`

Sure, but I think you can handle it, right? 

3. (maybe) optimize page loads by building upon "faster HTTP Reconnect" and adding push/pop state + sending only the changed parts

But I think this part is what the PR truly shines, and I am not sure if it would be better to finalize this before we ship the feature. Otherwise, I feel like `softReload` is just a function sitting in memory that is only used when the client is disconnected for a long time, making the utilization very low...

evnchn:
3. (maybe) optimize page loads by building upon "faster HTTP Reconnect" and adding push/pop state + sending only the changed parts

Moreover, as we can see after reverting https://github.com/zauberzeug/nicegui/commit/21c2fad85183731b9c8e7c50c1dc4766070267e7, push/pop state + sending only the changed parts already works on documentation. 

2025-06-0420-56-21-ezgif com-video-to-gif-converter

So on that front, I'm not sure what work we have to do, other than pytest and documentation.

rodja:
Ah ok. Only reverting the key-diffing but keeping pushState/popState might also be a good cut. But still I can break the documentation page quite easily. For example if you open `/` and then click on "Features" or "Demos" the page does not scroll. And after that when clicking on "Documentation" the page is white.

evnchn:
Can you send me a reproduction?

Though, I'm thinking if we can't get it to work after enough attempts, that we remove the duplication removal, and just send the entire unmodified element dictionary over...

This is because, we have #4796 on our side, if we want to do deduplication as well. Moreover, the bugs we are seeing looks very similar to #4828, so it looks like it'd not be an easy problem to solve.

rodja:
It seems that auto-scrolling works sometimes. But I found another, simpler reproduction:
Jun-05-2025 07-16-57
Just repeatedly clicking on the logo shows a blank page.

> I'm thinking if we can't get it to work after enough attempts, that we remove the duplication removal, and just send the entire unmodified element dictionary over...

Yes, the simpler this PR gets the better.

evnchn:
Although I fixed @rodja's issue on repeated clicks, **there is a big problem with this PR: memory leak in the browser-side**

{F3A626C4-BBB3-43A0-B700-4AEE1E9E272B}

I have banged my head against the wall for an hour to no avail. If we do not address this, this PR cannot proceed. 

I have tried: 

- Clearing out the elements with `mounted_app.elements = []`. 
  - Clearing out the events of all elements before doing that. 
- Try ensure complete deletion of the old Vue app besides unmounting it. 
- Take another approach by updating the elements of the existing Vue app. 

All attempts came up dry and I am lost as to how I can fix it, especially since the JavaScript-side is so complex...

---

~~At least I am pretty sure that #4821 won't suffer from this issue, since it doing things in the NiceGUI-layer.~~ #4821 would also be affected, since we're also unmounting and loading in new elements in that PR.

evnchn:
The plot thickens, as we see that there is no memory leak with a simple page (as seen by how the DOM nodes can come down naturally)

image

evnchn:
Hey @rodja and @falkoschindler 

**I would like to raise the importance of the issue of memory leak when our NiceGUI elements are unmounted.** 

This is because, in retrospect, #4821, as well as any other PR which attempts to achieve Single-Page Applications, will suffer from the issue. 

Fundamentally, the concept of SPA is, we're dragging one client along the lifespan of multiple pages, with us removing the elements of the old page and adding in the elements of the new page for a page transition. 

Thus, if this PR have memory leak, so will #4821. 

Now, I have already banged my wall at fixing the entire NiceGUI documentation website at once. 

Remaining approaches: 

- Go through each element one by one to see which one leak memory. 
- Fix the elements which leak memory, or we'd have to block them from being used in SPA.

rodja:
> Thus, if this PR have memory leak, so will https://github.com/zauberzeug/nicegui/pull/4821.

@evnchn can you create a separate issue for the memory leak? We are not sure how to reproduce, analyze or even fix it right now. As I understand it seems to not be related to this PR or [`ui.sub_pages`](https://github.com/zauberzeug/nicegui/pull/4821) but appears whenever we use `.clear` (which is also used to implement `ui.refreshable` for example).

evnchn:
Interesting. Let me try it. 

Originally I thought that the memory leak was only because I was getting NiceGUI to do something which it was never intended to do, and so it's not a bug. 

But with your refreshable argument, I believe that NiceGUI may be silently affected by it, but went unnoticed until now since not a lot of memory is leaked, if you only change not a lot of elements. 

Moving forward, I'll put the entire NiceGUI documentation page in a refreshable, and refresh it in rapid succession, to see if it also memory leaks. If it does, it's a bug, and the issue report will follow shortly. 

Thanks for the pointer!

evnchn:
Speaking of. The memory leak is fixed in 3.0 according to https://github.com/zauberzeug/nicegui/issues/4892#issuecomment-3210558779 by bumping Quasar. This code also has a lot of merge conflicts at this point. 

I don't see how my code is leaking memory any other way. I hope I am right but I may be wrong. 

Let's focus on what's urgent for 3.0 (AG Grid is one), and then introduce this as a non-breaking new feature in 3.x.

evnchn:
Was looking at old PRs. I just want to mention with how many performance-related PRs I have shoved into NiceGUI, the initial connect is already very fast (unless you have awful network, then trust me more than initial-connect will break). 

I begin to doubt whether this is necessary. 

At least, if this is necessary (maybe, considering JS-dead situations like #1430 or you are in an elevator or you are trying to make PWA (#2410, #3810, #3844)), then I prefer to work with a fresh mind. 

Let's close this PR. The spirit of the change is still contained within this thread so it's fine.

PR #4722, opened 2025-05-09
Use Mermaid API directly to fix nested diagram inside `ui.dialog` inside `ui.markdown`

evnchn:
This PR is extremely similar in nature to #4692, in that it fixes: 
- broken text as in https://github.com/zauberzeug/nicegui/pull/4170#issuecomment-2865321544
- showing the mermaid diagrams source code for about half a second whenever a page is refreshed containing the markdown component using mermaid, as in https://github.com/zauberzeug/nicegui/pull/4170#issue-2767685163

by, again invoking the Mermaid API exactly as prescribed at https://mermaid.js.org/config/usage.html#api-usage

Notable differences compared to past attempt at #4170:

- Uses `mermaid.render` not `mermaid.run`, more easy API integration
- `this.mermaid.initialize({ startOnLoad: false });` should not be awaited, according to the Mermaid docs.
- Since the SVG is added by me because I control the API usage, I always make the text hidden no-matter-what in the CSS, and add the SVG in another class so that it shows up. No need stuff like `.mermaid[data-processed="true"]``

Notable differences compared to #4692:

- Stores the last rendered Mermaid diagram, since the markdown is somehow updated when the `ui.dialog` finishes the animation, and otherwise we'd render twice and there'd be Layout Shift. 

There are still tasks to do:

- [x] Address multiple-mermaid situation (I'd imagine this implementation breaking, especially with the caching of rendered diagrams).

But the mermaid integration with markdown isn't perfect. There are still some tasks on my wishlist. I hope to address some of them some day soon:

- Cannot influence the mermaid initialize parameters so as to enable click events like https://nicegui.io/documentation/mermaid#handle_click_events
- Cannot listen for mermaid error messages like https://nicegui.io/documentation/mermaid#handle_errors

---

Test Script (updated to include multi-mermaid scenario):

<details>
<summary> It's a bit long but it works </summary>

```py
from nicegui import ui
import random


def generate_random_graph():
    nodes = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J']
    two_random_nodes = random.sample(nodes, 2)
    # Generate a random graph with 1 edge
    edges = [f"{two_random_nodes[0]} --> {two_random_nodes[1]}"]
    return f"graph TD; {' '.join(edges)}"


@ui.page('/mermaid_inside_markdown')
def mermaid_inside_markdown():
    md = ui.markdown('''
    - Mermaid inside markdown

    ```mermaid
    graph TD;
        A-->B;
        A-->C;
        B-->D;
        C-->D;
    ```
''', extras=['mermaid'])

# mermaid inside markdown inside ui.dialog

@ui.page('/mermaid_inside_markdown_dialog')
def mermaid_inside_markdown_dialog():
    with ui.dialog(value=True).props('transition-duration=1500'), ui.card():
        my_markdown = ui.markdown('''
        - Mermaid inside markdown inside dialog        

        ```mermaid
        graph TD;
            A-->B;
            A-->C;
            B-->D;
            C-->D;
        ```
        ''', extras=['mermaid'])

        def change_to_random_markdown_with_mermaid():
            my_markdown.set_content(f'''
            - Mermaid inside markdown inside dialog        

            ```mermaid
            {generate_random_graph()}
            ```
            ''')

        def change_to_markdown_with_errorneous_mermaid():
            my_markdown.set_content('''
            - Mermaid inside markdown inside dialog        

            ```mermaid
            graph TD;
                A-->B;
                A->C;
            ```
            ''')

        def change_to_markdown_with_many_mermaid():
            my_markdown.set_content(f'''
            - Mermaid inside markdown inside dialog        

            ```mermaid
            {generate_random_graph()}
            ```
            ```mermaid
            {generate_random_graph()}
            ```
            ```mermaid
            {generate_random_graph()}
            ```
            ''')
        ui.button('Change to random graph', on_click=change_to_random_markdown_with_mermaid)
        ui.button('Change to erroneous graph', on_click=change_to_markdown_with_errorneous_mermaid)
        ui.button('Change to many graphs', on_click=change_to_markdown_with_many_mermaid)

# many meriaid inside markdown inside dialog

@ui.page('/mermaid_inside_markdown_dialog_many')
def mermaid_inside_markdown_dialog_many():
    with ui.dialog(value=True).props('transition-duration=1500'), ui.card():
        my_markdown = ui.markdown(f'''
        - Mermaid inside markdown inside dialog        

        ```mermaid
        {generate_random_graph()}
        ```
        ```mermaid
        {generate_random_graph()}
        ```
        ```mermaid
        {generate_random_graph()}
        ```
        ''', extras=['mermaid'])

ui.link('mermaid inside markdown', '/mermaid_inside_markdown')
ui.link('mermaid inside markdown dialog', '/mermaid_inside_markdown_dialog')
ui.link('mermaid inside markdown dialog many', '/mermaid_inside_markdown_dialog_many')

ui.run(show=False, port=9191)

```

</details>

<img width="253" alt="{7C71FC62-1331-43CC-900D-5682586111C3}" src="https://github.com/user-attachments/assets/3a2191f9-fe0c-4439-a279-12e6a2562792" />

<img width="271" alt="{28A96891-BF42-43B3-8027-D0C7E6CDD3C2}" src="https://github.com/user-attachments/assets/34c3afd5-528b-4177-bd1b-30a54c612736" />

### DIFF
--- a/nicegui/elements/markdown.js
+++ b/nicegui/elements/markdown.js
@@ -14,6 +14,7 @@ export default {
   data() {
     return {
       mermaid: null,
+      stored: {}, // Initialize stored to avoid undefined errors
     };
   },
   updated() {
@@ -21,15 +22,26 @@ export default {
   },
   methods: {
     renderMermaid() {
+      // drop all keys in this.stored that is not used in this rendering
+      const diagramsToBeRendered = new Set();
+      this.$el.querySelectorAll(".mermaid-pre").forEach((pre, i) => {
+        diagramsToBeRendered.add(pre.children[0].innerText + i);
+      });
+      for (const key in this.stored) {
+        if (!diagramsToBeRendered.has(key)) {
+          delete this.stored[key];
+        }
+      }
       this.$el.querySelectorAll(".mermaid-pre").forEach(async (pre, i) => {
         try {
-          if (pre.children[0].innerText != this.lastContent) {
+          if (!this.stored[pre.children[0].innerText + i]) {
             const { svg, bindFunctions } = await this.mermaid.render(this.$el.id + "_mermaid_" + i, pre.children[0].innerText);
-            this.lastSvg = svg;
-            this.lastBindFunctions = bindFunctions;
-            this.lastContent = pre.children[0].innerText;
+            this.stored[pre.children[0].innerText + i] = {
+              svg: svg,
+              bindFunctions: bindFunctions,
+            }
           }
-          this.addMermaidToElement(pre, this.lastSvg, this.lastBindFunctions);
+          this.addMermaidToElement(pre, this.stored[pre.children[0].innerText + i].svg, this.stored[pre.children[0].innerText + i].bindFunctions);
         } catch (error) {
           const { svg, bindFunctions } = await this.mermaid.render(this.$el.id + "_mermaid_" + i, "error");
           this.addMermaidToElement(pre, svg, bindFunctions);
--- a/nicegui/elements/markdown.js
+++ b/nicegui/elements/markdown.js
@@ -14,7 +14,7 @@ export default {
   data() {
     return {
       mermaid: null,
-      stored: {}, // Initialize stored to avoid undefined errors
+      diagrams: {},
     };
   },
   updated() {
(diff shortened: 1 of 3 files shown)


evnchn:
Multiple-mermaid is done with https://github.com/zauberzeug/nicegui/pull/4722/commits/0661f6ef864ecb72879b11c5081e180297f9a9a5

Test script is updated in the first comment

evnchn:
Hi @falkoschindler I need some help with pytest. Not sure if what I done is right, but the tests are passing. 

I don't want to make the test too "dumb" and always pass. That just takes away from the purpose. So please assess.

falkoschindler (approved):
Very nice, @evnchn! All demo pages feel really great now!

I reviewed and refactored the code a little. At first I wasn't sure what `stored` is supposed to be. Now it is called `diagrams` which at least gives an idea. By introducing variables like `key` and re-arranging the code a little, I managed to simplify the loop a bit and inlined the extra method. And I noticed that we can collect `usedKeys` while creating new diagrams and prune unused keys afterwards.

The tests are indeed interesting. Using `shouldSee` matches the Mermaid source code which is invisible, so the tests fail. But as far as I can tell, using an XPATH doesn't guarantee that the resulting element is visible (or, e.g., stuck hidden in an animation). Therefore I replaced the rather useless "nodeLabel" check with an `is_displayed`. This should precisely check if the node is visible to the user.

Do you agree with the resulting implementation? Should we merge?

evnchn:
**Code in `mermaid.js` is very clean, and that part should be ready to merge.** 

If I traced the code correct, it should be identical in behaviour to the old one.

Particularly, `const key = pre.children[0].innerText + i;` captures concisely that we reuse the diagram, only if the innerText AND the order in the markdown match. Otherwise, the ID wouls not be the same (`c6_mermaid_0` -> `c6_mermaid_1`) and that can mess things up real bad. 

The using of a Set of avoid iterating over `this.$el.querySelectorAll(".mermaid-pre").forEach` is also clever. I don't know why I got stuck on the mindset purging the dictionary before the render in the past. Oh well...

---

But, sorry but I don't think I can comment on the pytest with my limited understanding and ability on that aspect. 

Overall, I find working with pytest a massive pain, because I can't interactively see what I am about to change. I wish I can make pytest alongside a Chromium browser where I can see the browser's logic...

If you are unsure, you need to get someone to double-check your work, or try and intentionally break `mermaid.js` and see if the tests flag you red.

falkoschindler:
Yeah, pytesting JavaScript libraries like Mermaid isn't fun. But I'm rather confident that the tests make sense.

One last thing: I added a newline separator to the diagram key in order to prevent the rare edge case that `innerText` and `i` mingle, e.g. `"A --> Node12" + 3` would equal `"A --> Node 1" + 23`.

falkoschindler (approved):

PR #4731, opened 2025-05-10
Website: Faster search page

evnchn:
This PR speeds up the documentation search page in the following ways: 

- Client, who is responsible* for running Fuse.js fuzzy search, needs to send back the indices only, not the entire content. 
  - This saves unnecessary uplink bandwidth from client to server. Messages went from 37KB to 0.5KB. 
- When the page is narrow, an indicator which means the page is on mobile clients, return 25 (a quarter of the original) results
  - According to Lighthouse, the benchmark for page speed, mobile browsers are a quarter as fast as desktop browsers, so I chose the value from there. 
- Let Fuse.js do the limiting of results, instead of us `.slice`
  - Theoreticlly slightly faster, since Fuse.js doesn't have to sort the entire list and provide a big return object. 

As far as what I can tell, the search results before and after this PR are exactly the same. 

*tried to find a Python substitute but came up dry...

### DIFF
--- a/website/search.py
+++ b/website/search.py
@@ -1,6 +1,7 @@
 from nicegui import __version__, background_tasks, events, ui
 
 from .documentation import CustomRestructuredText as custom_restructured_text
+from .documentation.search import search_index
 
 
 class Search:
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Amazing! Search is now so much faster! 🚀

I just improved the JavaScript code a little.

(Sorry for the back and forth in search.py. First I thought I could reduce the diff and improve the import, but the latter actually introduced a circular import.)

PR #4632, opened 2025-04-18
Pin version for JS dependencies

evnchn:
This PR supersedes #4482 and addresses #4478, while extending beyond a stopgap solution to satiate GitHub Dependabot and SBOM generation requirements. 

Notably (summarized from https://github.com/zauberzeug/nicegui/discussions/4478#discussioncomment-12872618):
* Since we are using version information from `package.json`, which is generated from `npm`, we can be sure that there won't be conflicts, which `npm` would have checked for us
* Makes `npm.py` repeatable (whereas currently it would generate a different output since it always gets the latest version)
* Makes it easy to bump JS dependencies minor versions, should we want to do so in 2.X (bumping major should be done with caution as always)

Steps to usage after this PR: 
* Init `package.json` (for me, I just do `{}`)
* Run `npm install` commands to install all JS dependencies NiceGUI needs (or do removal or upgrade)
* Run `pin_version.py` to copy those to `npm.json`. It will warn you if `package.json` has any extra entries or missing entries compared to `npm.json` (perhaps you add/remove a package via `npm`, and did not update `npm.json` accordingly)
* Run `npm.py` as usual. 

As of now, if I am not mistaken, the `npm.py` should be able to generate an exact copy of the dependencies we are using right now in NiceGUI 2.15.0. However, since I am on Windows and the script expects a Linux environment, I can't really test it out, but my mental tracethrough suggests that it should be fine. 

Normal console output when given an unpinned `npm.json`: 

```
  Changing: @tweenjs/tween.js from N/A to 25.0.0
  Changing: ag-grid-community from N/A to 32.1.0
  Changing: echarts from N/A to 5.5.1
  Changing: echarts-gl from N/A to 2.0.9
  Changing: es-module-shims from N/A to 1.10.0
  Changing: leaflet from N/A to 1.9.4
  Changing: leaflet-draw from N/A to 1.0.4
  Changing: mermaid from N/A to 11.0.2
  Changing: nipplejs from N/A to 0.10.2
  Changing: plotly.js from N/A to 2.35.0
  Changing: quasar from N/A to 2.16.9
  Changing: socket.io from N/A to 4.7.5
  Changing: tailwindcss from N/A to 3.4.10
  Changing: three from N/A to 0.168.0
  Changing: vanilla-jsoneditor from N/A to 0.23.8
  Changing: vue from N/A to 3.4.38
```

Normal console output after pinning: 

```
  Keeping:  @tweenjs/tween.js pinned to 25.0.0
[...]

PR #4733, opened 2025-05-10
Let npm.py run on Windows as well (hopefully same function as old)

evnchn:
This PR lets npm.py run on Windows as well. It also allows it to be ran repeatedly without throwing errors

Core changes: 
- Use `tempfile.TemporaryDirectory()`
- `unlink` before `rename` so that the old file is overwritten
- Try-except the failing line `if not any(destination.iterdir()):` `destination.rmdir()`

---

Admittedly, this is a minor concern, because by all means you should be on Linux or using a Dev Container, according to the [contributing guide](https://github.com/zauberzeug/nicegui/blob/42d24d40129785038d6fce9c2aafe24c62482478/CONTRIBUTING.md). 

But, both has failed me so far, and so I'm still on Windows. And, I need `npm.py` to work, so as to help out in #4656 and #4520. 

I personally believe that making `npm.py` runnable on cross-platform may speed up development of new elements, as we lower the barrier to custom element development. 

---

It is best to answer the following before we proceed: 

- [ ] What's the purpose of the `if not any(destination.iterdir()):` `destination.rmdir()`? 
  - It seems to always fail on Windows, so I add a try-except
  - But I don't want to just carelessly do that and call it a day
- [x] Is the output on Linux still working? 
  - You may want to grab the `npm.json` with pinned version in #4632, and test. 
  - It's the same on my side, minus Tailwind, which they changed stuff upstream and so we'll never get the exact same copy.

### DIFF
--- a/npm.py
+++ b/npm.py
@@ -20,6 +20,9 @@ from pathlib import Path
 from typing import Dict, List
 
 import requests
+import tempfile
+
+temp_dir = tempfile.TemporaryDirectory()
 
 parser = ArgumentParser()
 parser.add_argument('path', default='.', help='path to the root of the repository')
@@ -44,9 +47,7 @@ def url_to_filename(url: str) -> str:
 
 
 def download_buffered(url: str) -> Path:
-    path = Path('/tmp/nicegui_dependencies')
-    path.mkdir(exist_ok=True)
-    filepath = path / url_to_filename(url)
+    filepath = Path(temp_dir.name) / url_to_filename(url)
     if not filepath.exists():
         response = requests.get(url, headers={'User-Agent': 'Mozilla/5.0'}, timeout=3)
         filepath.write_bytes(response.content)
@@ -121,6 +122,8 @@ for key, dependency in dependencies.items():
                 filename = filename.replace(rename, dependency['rename'][rename])
 
             newfile = prepare(Path(destination, filename))
+            if newfile.exists():
+                newfile.unlink()
             Path(tmp, key, extracted.name).rename(newfile)
 
             if 'GLTFLoader' in filename:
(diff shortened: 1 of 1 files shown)


falkoschindler:
Using `TemporaryDirectory` and `unlink` make a lot of sense.

> What's the purpose of the `if not any(destination.iterdir()): destination.rmdir()`?

I'm not sure. The line is originally from @dclause. I guess it was meant to remove empty directories in case you remove a dependency from npm.json or change its target directory. What is the problem? Why does it fail on Windows?

> Is the output on Linux still working?

Seems to be working nicely.

evnchn:
> Why does it fail on Windows?

Errors generated are: 

```
[WinError 3] The system cannot find the path specified: 'nicegui\\static\\vue'
[WinError 3] The system cannot find the path specified: 'nicegui\\static\\quasar'
[WinError 3] The system cannot find the path specified: 'nicegui\\static\\tailwindcss'
[WinError 3] The system cannot find the path specified: 'nicegui\\static\\socket.io'
[WinError 3] The system cannot find the path specified: 'nicegui\\static\\es-module-shims'
```

No clue. 

---

But I don't think we should care that much, because:

- It works on Windows and Linux right now. 
- All that's going to happen is, we run the script, and commit only the files we need to Git by manual inspection. 
  - Like if we are adding xterm, we would commit only the xterm files after running the script. 
- `npm.py` constantly evolves with the changing landscape of the upstream. We can fix it later if it doesn't work for some reason.

falkoschindler (approved):

dclause:
From memory, I don't remember why the code was so "unix specifix".
@falkoschindler the line you mentionned was intended to manually clean the temporary folder so you could restart the process. This was because how specific is the file path, it was always the same folder reused, hence needed cleaning (you can't rely on autoclean delay to restart the script).
On windows, the files are cleaned right away when unused and flag "in-used" before, so it might be the issue faced before: the files where actually in used and not deletable. 
On linux, the files are cleaned at service restart or based on a temporal value. 

The new refactoring looks good at first sight (although I did not tried it) because the folder is dynamically created by the system itself so it is not reused and not system specific, and the folder deletion is left at the standard sytem behavior appreciation. 
Looks good indeed.

PR #4754, opened 2025-05-16
Delay emitter until NiceGUI-level handshake done

### LINKED ISSUE #4753: `delayed_emitter` and Leaflet `$emit` runs too early, emitting events before Client initialized
### First Check

### Example Code

```python
from nicegui import ui


@ui.page('/')
def leaflet_page():
    m = ui.leaflet(center=(40.73997376331186, -74.22363281250001), zoom=11).classes('w-96 h-96')
    rawimg = ui.element('img').props('src=https://broken.com')
    rawimg.on('error', lambda: print('Image error triggered'))


ui.run()
```

### Description

Under any normal circumstance, the code works: 

- Leaflet map shows up
- "Image error triggered" shows up in the console. 

But, if you artificially make NiceGUI respond to Socket.IO "handshake" event slower (check additional context), then the code breaks: 

- Leaflet map does not show up
- "Image error triggered" does not show up in the console. 

So, this bug is really hidden in any decently-performant system, but will break if: 

- NiceGUI server is slow enough
- Some proxy/firewall delays the packets sent via WebSocket to inspect them. 
- When I eventually shift the Socket.IO initialization earlier, as in my branch https://github.com/evnchn/nicegui/tree/earlier-socketio

Also affects any other elements which emits "init" using a similar checking mechanism, and any early events emitted. 

### NiceGUI Version

2.17.0

### Python Version

3.12.9

### Browser

Chrome

### Operating System

Windows

### Additional Context

Image

Suspected reason: 

`nicegui.js` considers the handshake to the server "done", immediately after sending the Socket.IO message to the server. 

https://github.com/zauberzeug/nicegui/blob/3172fe8cb19dd7fba137fb1e141f41073cc46543/nicegui/static/nicegui.js#L347-L364

But this is wrong. It should do it on the callback, and only when received `ok = true`. 

`leaflet.js` considers the handshake to the server "done", if `window.socket.id` no longer is `undefined`

https://github.com/zauberzeug/nicegui/blob/3172fe8cb19dd7fba137fb1e141f41073cc46543/nicegui/elements/leaflet.js#L122-L126

But this is wrong. It should check for `did_handshake` (provided the above is fixed).

evnchn:
This PR fix #4753, possibly paves the way to shifting the Socket.IO intialization earlier. 

Notable changes: 

- [Handshake is done only when receiver server ok](https://github.com/zauberzeug/nicegui/commit/793adb4814b382c755749289f9583fc90e0b24dd)
- [Retry emitting message repeatedly until handshake is done](https://github.com/zauberzeug/nicegui/commit/a107315d1d117bd5ab5ce130a10e968c33532157)

Old logic: 
- Handshake is done when we sent the message. Who cares how long that message may take to reach the server
- If the handshake is not done, we wait 10ms and emit regardless. 
  - Meaning if the handshake is not done after 10ms, the message gets lost. 

New logic: 
- We wait for the NiceGUI side to return True before considering the handshake is done
- If the handshake is not done, we keep on retrying at 10ms interval
  - Meaning if the handshake is not done after 10ms, the message does not get lost. 

About the handshakes: 

- Engine.IO-level handshake, receiving `0{sid: "m-Lwo1fStO8e-QAbAAAU", upgrades: [] ...`
- Socket.IO-level handshake, receiving `40{"sid":"oxVbOYM32Hieu1anAAAV"}`
  - Some components wait for this handshake to emit, because they need `window.socket.id`, like Leaflet
- NiceGUI-level handshake, sending `420["handshake", {client_id: "7b6b9df6-e24a-489c-9dd2-012a08b34a10"...` and receiving `430[true]`
  - This PR refers to waiting for this handshake, as it is the most important one. 
  - Before this handshake, any event emitted will not be registered, since the Socket.IO ID is not associated with the `client_id` yet.

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -358,9 +358,9 @@ function createApp(elements, options) {
               console.log("reloading because handshake failed for clientId " + window.clientId);
               window.location.reload();
             }
+            window.did_handshake = true;
             document.getElementById("popup").ariaHidden = true;
           });
-          window.did_handshake = true;
         },
         connect_error: (err) => {
           if (err.message == "timeout") {
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -191,7 +191,7 @@ function renderRecursively(elements, id) {
           });
         const delayed_emitter = () => {
           if (window.did_handshake) emitter();
-          else setTimeout(emitter, 10);
+          else setTimeout(delayed_emitter, 10);
         };
         throttle(delayed_emitter, event.throttle, event.leading_events, event.trailing_events, event.listener_id);
         if (element.props["loopback"] === False && event.type == "update:modelValue") {


falkoschindler (approved):
Brilliant!

1. The `delayed_emitter` is an obvious mistake (once you see it).
2. Moving `did_handshake = true` into the handler makes totally sense. Even though you could argue that the socket connection is _connected_ which is enough for emitting events. But then the name "did_handshake" would be wrong. Anyway, waiting for the handshake to complete seems reasonable.

evnchn:
> socket connection is connected which is enough for emitting events

Let me be clear (for my future reference as well), if just the socket connection is connected, though you can emit events, those are events which NiceGUI could not make sense of natively. 

Because the Socket.IO socket is not associated with the NiceGUI client. 

So, unless you are doing `window.socket.emit`, `@sio.on` and `window.did_handshake` (which is a rare combo IMO, why would you do that) there would be no harm done by this PR. 

Even if it did, (1) it's a few milliseconds slower only, (2) you can always move to doing `window.socket.id !== undefined` for the check as you should, this way you are precisely checking against the handshake layer you need. 

Or use the NiceGUI-native native `$emit` or `emitEvent` directly. 🙄

Also `window.did_handshake` I think its not a public API, so we by-all-means can change it.

PR #4755, opened 2025-05-16
move websocket initialization code to create hook

xaptronic:
Part of an effort to have custom vue component emit an `init` type event led to investigating how NiceGUI app component uses `mounted` hook to initialize websocket, which runs after subcomponent hooks. 

In order for the component to emit and have server receive, sub components can now use window.socket with a `connect` handler (or potentially other types of events?) 

This moves the socket initialization code to `beforeCreate()`, but leaves `mounted_app = this` in the mounted hook.

Early vibe testing is that the functionality I have that is invoked by custom component init calls is now much faster than waiting for everything to mount and then being able to either invoke the function via a `run_method` server call, or polling for socket availability.

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -331,7 +331,8 @@ function createApp(elements, options) {
     render() {
       return renderRecursively(this.elements, 0);
     },
-    beforeCreate() {
+    created() {
+      mounted_app = this;
       window.documentId = createRandomUUID();
       window.clientId = options.query.client_id;
       const url = window.location.protocol === "https:" ? "wss://" : "ws://" + window.location.host;
@@ -434,9 +435,6 @@ function createApp(elements, options) {
           }
         });
       }
-    },
-    mounted() {
-      mounted_app = this;
     },
   }).use(Quasar, {
     config: options.quasarConfig,


evnchn:
I also think the code would break, if we receive a Socket.IO message in the gap time that is after `beforeCreate` but after `mounted`

Because `this` would be `undefined` otherwise. 

This is because I recall in `beforeCreate`, I once tried to inject a `console.log(mounted_app)` and it prints `undefined`

evnchn:
{61105B37-EC14-47E7-93C8-0689146D8DE5}


{A4462985-6CB3-44FB-9C19-60390DCE3588}

It seems like we are gaining not even a single frame of speed increase. Still if that 4ms is the difference between being able to use `connect` handler of `window.socket`, it'd make all the difference.

falkoschindler (changes requested):
Ok, the PR looks good so far.
But what do you guys think, should we replace the polling
```js
const connectInterval = setInterval(() => {
  if (window.socket.id === undefined) return;
  this.$emit(...);
  clearInterval(connectInterval);
}, 100);
```
with
```js
window.socket.on('connect', () => this.$emit(...));
```
in router_frame.js, leaflet.js, scene_view.js and scene.js? Seems like a good opportunity to get rid of these bulky workarounds.

And I think we can replace
```js
window.socket?.emit("event", {...})
```
with
```js
window.socket.emit("event", {...})
```
in nicegui.js because `created()` runs before the first `render()`.

evnchn:
Spotted a potential breaking change while staring at the diagram. 

image

Do you agree that, if we immediately receive a message from Socket.IO which is to `run_javascript` (say: `document.getElementFromId('c4').innerHTML = 'blah'`), and that is before we mount everything to `#app` (assuming slow Vue, fast localhost network), that we run into a race-condition and the JavaScript request would fail? 

It seems like it is best not to change when the functions are defined, but rather keep it in `mounted` while shifting websocket init earlier as in #4756 is a more safe change in that regard (though that PR isn't without its issues)

falkoschindler:
Yes, that could be a problem.
So you're saying PR #4756 would be better anyway and could replace this PR #4755?

> though that PR isn't without its issues

I guess you're referring to "we are re-defining what `await ui.context.client.connected(timeout=30)` entails".

Given all these nitty gritty details that could break user code in a hard-to-foresee manner: Is it worth it?

I just noticed that we can replace the interval with this one-liner without this PR:
```js
this.$nextTick(() => window.socket.on("connect", () => this.$emit("init", { socket_id: window.socket.id })));
```
`$nextTick` waits for `window.socket` to be defined in `app.mounted`, and `on("connect", ...)` waits for the socket connection. At least this should be a viable alternative to the original motivation of this PR.

evnchn:
> So you're saying PR https://github.com/zauberzeug/nicegui/pull/4756 would be better anyway and could replace this PR https://github.com/zauberzeug/nicegui/pull/4755?

Yes. We would then be respecting the old behaviour of processing the messages in `mounted`, ensuring the most compliant and backwards-compatible behaviour on that front. 

> I guess you're referring to "we are re-defining what await ui.context.client.connected(timeout=30) entails".

Yes, but my counterargument was "it's indistinguishable from a client which just had an awful network spike right after connection"

So, I'm thinking that it'd be fine and the user's code is not broken any more than a bad network connection would. Users may have to increase timeouts to be more generous, but I don't see any big changes for user code. 

Let's discuss in #4756, if you want to dig deeper on that aspect. 

> Given all these nitty gritty details that could break user code in a hard-to-foresee manner: Is it worth it?

Do you think that applies to #4755, #4756, or both? I'd argue both PRs require changes to the user code in _some_ way, but comparing, I believe #4756 is a slightly less bitter pill to swallow, if we can get away with increasing timeouts. Otherwise, someone (user / NiceGUI) need to try-except their JavaScript calls (or determine if it's ready to run otherwise), which may be quite a pain. 

> `this.$nextTick` solution

That's awesome! I think it is more elegant, since we already use `await this.$nextTick();` for waiting `window.path_prefix` to be set for some elements, so that just brings the codebase in line with itself. Do you want to start a PR on that? 

---

Overall, I believe that, since @falkoschindler just raised the `this.$nextTick` solution which can enable the use of `window.socket.on("connect", ...)` without moving where we handle the events, it seems that this PR doesn't bring much to the table, especially when the gains is only 4ms. 

Note, however, that `window.socket.on("connect", ...)` would NOT work with #4756 most likely, since the Socket.IO is connected _well before_ the Vue app is even loaded. We may need to set a flag somewhere to skip the `window.socket.on("connect", ...)` and run `this.$emit("init", { socket_id: window.socket.id })` right away.

falkoschindler:
A quick update: PR #4756 has been closed, so this PR is technically unblocked. However, after reviewing the current state, there are a few concerns:

**Branch is significantly outdated.** `main` has diverged substantially since this was opened — the `mounted()` hook now includes `finishHandshake`, `load_js_components`, message size wrapping, and other changes. A rebase would likely produce non-trivial conflicts.

**`mounted_app` timing risk.** Moving the assignment `mounted_app = this` from `mounted()` to `created()` means it's set before `$refs` are populated. `getElement()` relies on `mounted_app.$refs[...]`, so any WebSocket message arriving between `created()` and `mounted()` could hit undefined refs.

**DOM access in `created()`.** Socket handlers reference `document.getElementById("popup")` etc. While these likely exist in the HTML before Vue mounts, the `update` handler uses `this.$nextTick()` and reactive element access, which assume the component is fully mounted.

**Unrelated change.** The `.use(Quasar, { config: options.quasarConfig })` addition isn't mentioned in the PR description — it should be explained or handled separately.

**No test coverage.** Given that this touches the core initialization path for every NiceGUI app, we'd need evidence that existing functionality isn't broken and that the performance improvement is measurable.

**Hard to follow.** Multiple force-pushes have rewritten the branch history, and the existing discussion references code that is no longer part of the current diff. This makes it difficult to review the PR in its current form.

Given the amount of drift from `main` and the issues above, I'd suggest closing this PR. If making `window.socket` available earlier for custom component hooks is still something you'd like to pursue, a fresh PR against current `main` with a narrower scope (e.g., initializing only the socket in `created()` while keeping `mounted_app = this` and handler registration in `mounted()`) would be a better starting point. We're happy to discuss the approach in an issue first.

PR #4756, opened 2025-05-16
Move Socket.IO initialization all the way to head + Pre-Mount Message Caching

evnchn:
This PR is a more heavy-handed approach in making sure we establish Socket.IO initialization as soon as possible. It is benchmarked to be at least 50% faster than the current implementation, with higher performance gains under more adverse network connection. 

Before PR: 

image

> "Connected in" time > DOMContentLoaded, since we only establish Socket.IO connection after mounted

After PR: 

{FA188AE4-E9B2-4588-8118-3FE6CBA9DE94}

> "Connected in" time < DOMContentLoaded, since we establish Socket.IO connection as early as possible, and that occurs before Vue was loaded in at all. 

Explainer: 

Currently, NiceGUI loads like this: 

- Load in the CSS files in head
- Load in the JS files in body
- `createApp`
- When the app is mounted, setup Socket.IO
- The triple handshake (Engine.IO, Socket.IO, then NiceGUI)
- `await ui.context.client.connected()` finishes

My PR change it to be like this

- Load in the CSS files in head
- **Load in just enough JS to setup Socket.IO with a pre-mount message queue**
- The triple handshake (Engine.IO, Socket.IO, then NiceGUI)
- `await ui.context.client.connected()` finishes
- Load in the JS files in body
- `createApp`
- When the app is mounted, **execute the messages stored in the message queue and re-setup Socket.IO**

Since we begin talking to Socket.IO earlier, we:

- Complete the triple-handshake earlier
- `await ui.context.client.connected()` finishes earlier
- Obtain the messages to be ran after mounted earlier

To-do items: 

- [ ] Heard pytest wasn't too happy about this. Let's fix either pytest or my code! (likely former)
- [ ] Remove debugging `console.log` prints. 

---

PS: #4755 change it to be like this

- Load in the CSS files in head
- Load in the JS files in body
- `createApp`
- When the app is **`beforeCreate`**, setup Socket.IO
- The triple handshake occurs from here (Engine.IO, Socket.IO, then NiceGUI)
- Assign `mounted_app = this;` when app is `mounted` only
- `await ui.context.client.connected()` finishes

### DIFF
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -4,6 +4,26 @@
     <title>{{ title }}</title>
     <meta name="viewport" content="{{ viewport }}" />
     <link href="{{ favicon_url }}" rel="shortcut icon" />
+    <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/socket.io.min.js" fetchpriority="high"></script>
+    <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/nicegui_head.js" fetchpriority="high"></script>
+    <script>
+      console.log('before initSocketIO');
+      console.time('initSocketIO');
+      window.nicegui_options = {
+        version: "{{ version }}",
+        prefix: "{{ prefix | safe }}",
+        query: {{ socket_io_js_query_params | safe }},
+        extraHeaders: {{ socket_io_js_extra_headers | safe }},
+        transports: {{ socket_io_js_transports | safe }},
+        quasarConfig: {{ quasar_config | safe }},
+      }
+      initSocketIO(window.nicegui_options);
+      console.timeEnd('initSocketIO');
+      console.log('initSocketIO');
+      window.socket.on('connect', () => {
+        console.log('connected_debug');
+      });
+    </script>
     <link href="{{ prefix | safe }}/_nicegui/{{version}}/static/nicegui.css" rel="stylesheet" type="text/css" />
     <link href="{{ prefix | safe }}/_nicegui/{{version}}/static/fonts.css" rel="stylesheet" type="text/css" />
     {% if prod_js %}
@@ -19,20 +39,6 @@
     {% for url in js_imports_urls %}
     <link rel="modulepreload" href="{{ url }}" />
     {% endfor %}
-    <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/socket.io.min.js"></script>
-    <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/nicegui_head.js"></script>
-    <script>
-      window.nicegui_options = {
-        version: "{{ version }}",
-        prefix: "{{ prefix | safe }}",
-        query: {{ socket_io_js_query_params | safe }},
-        extraHeaders: {{ socket_io_js_extra_headers | safe }},
-        transports: {{ socket_io_js_transports | safe }},
-        quasarConfig: {{ quasar_config | safe }},
-      }
-      initSocketIO(window.nicegui_options);
-      console.log('initSocketIO');
-    </script>
   </head>
   <body>
     <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/es-module-shims.js"></script>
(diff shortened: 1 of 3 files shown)


evnchn:
Tests also completed slightly faster. 

Before

{44D2F0E3-132D-4219-8ABD-A5A5E2FFAE6F}

After

{B47F85B7-D58B-41F4-A646-4190F0101222}

But not by much, since everything should be cached.

evnchn:
Got some Git issues going on but I believe up to commit https://github.com/zauberzeug/nicegui/pull/4756/commits/c5160b491c3a2ee06b6c8d22b0c306646453c9d7 this PR contains #4754 

If not I'd be sad...

evnchn:
> Perhaps we need to introduce a flag in ui.context.client.connected to wait for the app to be mounted, which is default-True in 2.x for backwards-compatible behaviour, but potentially default-False in 3.x to encourage users sending message to the Outbox ASAP.

No. This is because our new behavior is in-principle "impossible to distinguish from a browser which _just so happens_ to suffer a network lag shortly after load". The transport layer is in NiceGUI's hands. We can do whatever we want. 

If they need to time exactly from when the browser is fully loaded, they need `await ui.context.client.connected()` then `await ui.run_javascript('blah')` to guage when did the JavaScript actually get ran (since receiving a result is an indication that the message is not in `premountMessageQueue` but actually ran)

falkoschindler:
Ok, you say this PR is 50% faster and this is basically the time it takes to `await ui.context.client.connected()`. But at the same time you're changing what `connected()` means. So I'm not yet convinced that we're getting any noticeable improvement. The client might be connected, but still can't process messages because the client is still busy loading dependencies and mounting the app. The timelines in your screenshots look pretty similar, indicating both pages being ready almost at the same time. So what is the actual benefit for the user?

evnchn:
The page does indeed get load in faster, since you are in essence parallelizing your network requests instead of one-by-one. 

image

Since Socket.IO handshake is actually 3 handshakes (Engine-IO, Socket-IO, and NiceGUI, check #4754 if you'd like a refresher), we incur time penalty of **3x the round-trip network latency**, so it can be quite a lot, if your network is bad enough. 

So, sorry but I can't agree that "timelines in your screenshots look pretty similar", and there is definitely noticeable improvement. 

But my simplified timeline also shows the change: `await ui.context.client.connected()` will return earlier, and earlier than the browser Vue actually being ready, so messages sent after the orange block but before the yellow block will be held in a queue.

evnchn:
I have an analogy. 

Imagine Socket.IO is the engine in a gasoline vehicle. 

#4755 is tuning the engine start sequence so that when you turn the key, the engine spins up a bit faster. 

#4756 is getting someone to go downstairs and start the car for you ahead-of-time. 

If that doesn't make things clear, I don't know what will...

evnchn:
Since the meaning of `await ui.context.client.connected()` will change slightly after this PR (client connected doesn't necessarily mean Vue is ready, though the messages will still be processed only when Vue is ready though), not sure if this is best to be done in 3.0, or maybe the next opportunity is 4.0 then.

nicegui/templates/index.html
```diff
@@ -4,6 +4,18 @@
     <title>{{ title }}</title>
     <meta name="viewport" content="{{ viewport }}" />
     <link href="{{ favicon_url }}" rel="shortcut icon" />
+    <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/socket.io.min.js" fetchpriority="high"></script>
+    <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/nicegui_head.js" fetchpriority="high"></script>
```
evnchn, line 8: Should be `async` so that we do not block the HTML parser while the download takes place. 

We will set `window.nicegui_options` early and `initSocketIO(window.nicegui_options);` would be part of `nicegui_head.js`

This way this PR becomes even faster.

evnchn:
As the time gained is not a whole lot, and the application logic becomes significantly more complex, I would close this PR. (considering also the review load)

PR #4757, opened 2025-05-16
Document use of user fixture with ui.select

barrywhart:
This PR addresses #4737 by documenting how the `user` fixture supports the `ui.select` control.


evnchn:
seems to be good from first sight

just want to ask if there are any uses of the user fixture still left on the table undocumented...

barrywhart:
> just want to ask if there are any uses of the user fixture still left on the table undocumented...

I'm still learning NiceGUI and the user fixture, so I can't say for sure. I briefly reviewed user.py and userinteraction.py. I see that UserInteraction has a `clear()` method that is _documented_, but there's no example of _using_ it. On the other hand, the method sounds straightforward, so perhaps no example is required.

rodja (approved):
Thanks for the PR @barrywhart! I generalized the text from the autocomplete demo to provide a high level introduction to the concept of the `UserInteraction` element. Also I boiled down your new demo to the multi-select version because it's the superset and no so different from the special case.

What do you think?

website/documentation/content/user_documentation.py
```diff
@@ -105,6 +105,75 @@ def querying():
             ''')
 
 
+doc.text('User Interaction', '''
+    `user.find(...)` returns an `UserInteraction` object which provides methods to type text,
```
barrywhart, line 109:
```suggestion
    `user.find(...)` returns a `UserInteraction` object which provides methods to type text,
```

Weird quirk of English: the letter "u" at the beginning of a word _sounds like_ a consonant.

website/documentation/content/user_documentation.py
```diff
@@ -105,6 +105,75 @@ def querying():
             ''')
 
 
+doc.text('User Interaction', '''
```
barrywhart, line 108: Would it be possible to make "User Interaction" a higher-level heading, with two sub-headings:
* Autocomplete
* Selecting options

barrywhart:
I commented on the changes. Looks good, with a couple of suggestions.

I have another testing-related question (we can move it to a discussion if you like): Does `UserInteraction` support interacting with a `ui.table`, for example, a table with a clickable "View details" button for each row?

```
            <!-- View details Button Column -->
            <q-td auto-width>
                <q-btn
                    size="sm"
                    color="positive"
                    round
                    dense
                    icon="zoom_in"
                    @click="() => $parent.$emit('view_logs', props.row)"
                >
                    <q-tooltip>View details</q-tooltip>
                </q-btn>
            </q-td>
```

website/documentation/content/user_documentation.py
rodja: Unfortunately not. The documentation system has no sub-sections.

rodja:
> Does UserInteraction support interacting with a ui.table

Not jet. It would be great to open a Feature Request for this so we can discuss different approaches.

barrywhart:
> > Does UserInteraction support interacting with a ui.table
> 
> Not jet. It would be great to open a Feature Request for this so we can discuss different approaches.

Will do.

falkoschindler (commented):
There seems to be an inconsistency between the User fixture and a regular UI:
To select multiple fruits from a select element like this
```py
ui.select(
    label='Fruits',
    options=['Apple', 'Banana', 'Cherry'],
    multiple=True,
    on_change=lambda e: ui.notify(', '.join(e.value)),
)
```
you need to click [ui.select], "Apple", "Banana", ... But the User fixture seems to close the select element, so that you need to simulate clicks to [ui.select], "Apple", [ui.select], "Banana".

Looks like a bug that needs to get fixed.

barrywhart:
@falkoschindler: Do you need any information or changes from me?

evnchn:
I think it is likely that we will hold off this PR until we finish investigating the bug that he mentioned. 

Since otherwise, user fixture ≠ real UI, and the testing becomes meaningless. 

If you want, you can join in on the investigation to move things along.

falkoschindler:
I just wrote two tests to compare both test fixtures. But there seems to be even another bug: After selecting Apple and Banana with `multiple=True`, the `value` seems to contain "select". 🤦🏻‍♂️

barrywhart:
I don't have time this week to help with the investigation. Good luck! I'll keep an eye on this PR. Definitely not urgent, but I hope we can get this merged in the next release or two. 🙌🏽

falkoschindler:
I think I fixed both problems: Not only do we need to open and close the simulated popup more carefully; we also need to check if the target is actually a possible value. Otherwise we need to close the popup.

Looks good to me now. If there are no objections, I'll merge it tomorrow.

falkoschindler (approved):

evnchn (commented):
There may be slight difference between the user test and the screen test?

tests/test_select.py
```diff
@@ -205,3 +205,58 @@ def test_invalid_value(screen: Screen):
...
+
+
+@pytest.mark.parametrize('multiple', [False, True])
+def test_opening_and_closing_popup_with_screen(multiple: bool, screen: Screen):
+    select = ui.select(options=['Apple', 'Banana', 'Cherry'], label='Fruits', multiple=multiple).classes('w-24')
+    ui.label().bind_text_from(select, 'is_showing_popup', lambda v: 'open' if v else 'closed')
+    ui.label().bind_text_from(select, 'value', lambda v: f'value = {v}')
+
+    screen.open('/')
+    fruits = screen.find_element(select)
```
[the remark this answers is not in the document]
rodja: @falkoschindler mainly wrote the test to show the discrepancy between user test and screen test. But the increased symmetry does not hurt in this case.

rodja (approved):

tests/test_select.py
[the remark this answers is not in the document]
falkoschindler: Oh yes, I missed this line.

PR #4759, opened 2025-05-16
Generalize `js_handler` with Auto-Return-Value-Intercept (ARVI)

evnchn:
This PR should end the saga of PR #4689, PR #4618, while addressing 5 issue / discussions / feature request (#3762, #4611, #4606, #4577, and #4589) as stated in https://github.com/zauberzeug/nicegui/pull/4618#issuecomment-2845918657

#### Background

The crux of the issue is, we want to be able to return custom values which are only possible to be extracted via JavaScript and/or too deep in nested property which `args` does not handle. Stuff gets pretty complex and case-by-case, so read up those 7 referenced items above!

#### Established Requirements

We debated a long time over the API design, and we needed the following requirements: 

- Avoid adding extra parameters whenever possible
  - Option `js_transform = "(e) => e.offsetY"` was eliminated 
- JavaScript transformations should be as easy to add as `"(e) => e.offsetY"`
  - Option `js_handler = "(e) => {defaultHandler(e.offsetY)}"` was eliminated 
- Must be as clear and minimize confusion
  - Option "shoehorning `args`", as well as all other above options, were eliminated

This PR illustrates the idea at https://github.com/zauberzeug/nicegui/pull/4689#issuecomment-2887683005 to comply with all of the aforementioned requirements. 

- Avoid adding extra parameters whenever possible
  - Yes
- JavaScript transformations should be as easy to add as `"(e) => e.offsetY"`
  - `js_handler = "(e) => e.offsetY"`
- Must be as clear and minimize confusion
  - I think so???

#### Logic behind this PR

The keyword is **ARVI**, which stands for Auto-Return-Value-Intercept. It means we intercept the `js_handler`'s return value, if and only if Python-side `handler` is attached. 

As a recap from https://github.com/zauberzeug/nicegui/pull/4689#issuecomment-2887683005: 

- `.on('blah', handler=python_handler, js_handler='(e) => e.event.button')`
  - Since we pass both `handler` and `js_handler`, we now **do** care about the return value, and if there is a return value, we pass to `defaultHandler` right away. 
- `.on('blah', js_handler='(e) => e.event.button')`
  - Old behaviour. `js_handler` doesn't care about return value. 
- `.on('blah', handler=python_handler)`
  - Old behaviour 🙄

#### Notable implementation details

- **`args` and `js_handler` remain mutally exclusive**. 
  - Makes no sense otherwise. You wanna filter twice?
- Logic check `if (event.js_handler) {` changed safely to `if (!event.handler) {`
  - Because no handler + code is even executed => js_handler must be present
  - Since we have a `if handler or js_handler:` on Python-side

#### Remaining to-do 

- [ ] Persuading @falkoschindler
- [ ] Documentation
- [ ] Pytest

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -4,7 +4,7 @@ import inspect
 import re
 from copy import copy
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, Iterator, List, Optional, Sequence, Union, cast, overload
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, Iterator, List, Optional, Sequence, Union, cast
 
 from typing_extensions import Self
 
@@ -320,26 +320,6 @@ class Element(Visibility):
             Tooltip(text)
         return self
 
-    @overload
-    def on(self,
-           type: str,  # pylint: disable=redefined-builtin
-           *,
-           js_handler: Optional[str] = None,
-           ) -> Self:
-        ...
-
-    @overload
-    def on(self,
-           type: str,  # pylint: disable=redefined-builtin
-           handler: Optional[events.Handler[events.GenericEventArguments]] = None,
-           args: Union[None, Sequence[str], Sequence[Optional[Sequence[str]]]] = None,
-           *,
-           throttle: float = 0.0,
-           leading_events: bool = True,
-           trailing_events: bool = True,
-           ) -> Self:
-        ...
-
     def on(self,
            type: str,  # pylint: disable=redefined-builtin
            handler: Optional[events.Handler[events.GenericEventArguments]] = None,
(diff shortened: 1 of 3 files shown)


evnchn:
So we sort of said "imagine setting a flag JS_TRANSFORMATION = True somewhere, marking that we now suddenly do care about the return value"

Actually that flag is implemented here. 

https://github.com/zauberzeug/nicegui/blob/9198a3716cb9303d03665cce9f46ba8140692af7/nicegui/event_listener.py#L44

With `event.handler == true`, the JS side, instead of `handler = eval(event.js_handler);`, will `processed_args = [JSON.stringify(eval(event.js_handler)(...args))];` instead. 

Note that `processed_args` is a list of JSON-strings, so we match the original format.

evnchn:
Closing this to further encourage converging on https://github.com/zauberzeug/nicegui/pull/4689#issuecomment-2888516247

Branch not immediately deleted just in case.

PR #4769, opened 2025-05-20
Use Plausible Analytics on nicegui.io

rodja:
Spiked by discussion #4746, this PR adds to our https://nicegui.io page:

- Plausible Analytics (wich is also used by FastAPI for their project website)
- imprint & privacy page


evnchn:
If I understand correctly, the previous Prometheus puts `prometheus.py` in the Docker container, but this time around, it should stay dormant without the script tag in `index.html`

Because I hear that swapping out the main.py in the Docker container is a legit way of running your own NiceGUI apps,and I was thinking whether it is possible that people who pull the Docker container may be getting analytics even though they don't want any.

rodja:
Thanks @evnchn. Thats a very valuable insight. As soon as `main.py` is replaced, no tracking will take place. But the container is also used to view the documentation locally. To avoid analytics in such cases, I just changed the code to only include the Plausible JS when docas are deployed via fly.io.

falkoschindler (approved):

PR #4778, opened 2025-05-21
Introduce "Dynamic Resources" for robust inclusion of `codehilite.css` against match-all `@ui.page("/{_:path}")`

### LINKED ISSUE #4774: codehilite.css not loaded in single_page_app example
### First Check

### Example Code

```python
#!/usr/bin/env python3
from router import Router

from nicegui import ui


@ui.page("/")  # normal index page (e.g. the entry point of the app)
@ui.page(
    "/{_:path}"
)  # all other pages will be handled by the router but must be registered to also show the SPA index page
def main():
    router = Router()

    @router.add("/")
    def show_one():
        ui.label("Content One").classes("text-2xl")

    @router.add("/two")
    def show_two():
        ui.label("Content Two").classes("text-2xl")

    @router.add("/three")
    def show_three():
        ui.label("Content Three").classes("text-2xl")
        ui.markdown("""
                    print("hello world")
                    """)

    # adding some navigation buttons to switch between the different pages
    with ui.row():
        ui.button("One", on_click=lambda: router.open(show_one)).classes("w-32")
        ui.button("Two", on_click=lambda: router.open(show_two)).classes("w-32")
        ui.button("Three", on_click=lambda: router.open(show_three)).classes("w-32")

    # this places the content which should be displayed
    router.frame().classes("w-full p-4 bg-gray-100")


ui.run()
```

### Description

Above is a small variation on the [single_page_app example](https://github.com/zauberzeug/nicegui/tree/main/examples/single_page_app) that uses the `ui.markdown` element.

When you load the page "three", the code is not properly highlighted. The JS console shows an error: 
`The stylesheet http://localhost:8080/_nicegui/2.17.0/codehilite.css was not loaded because its MIME type, "text/html", is not "text/css".`

At first sight, this only occurs for `/codehilite.css`, not for `/static/...` or `/libraries/...`.

Is this due to the way how/when `codehilite.css` is mounted? 

### NiceGUI Version

2.17.0

### Python Version

3.11.9

### Browser

Firefox

### Operating System

Windows

### Additional Context

If I add `ui.markdown()` just after `from nicegui import ui`, the issue is resolved. I guess that forces the codehilite.css route to be added before `{_:path}`, thereby getting higher priority.

Can this be fixed in the source, so this workaround is not necessary? E.g. add `markdown: bool` to `ui.run()` which is default `True`.

evnchn:
This PR fix #4774 by introducing the concept of "Dynamic Resources", resources whose `Response` (and thus content) are provided as return value by a callable. 

This is a new and unique mechanism compared to libraries/components/resources, which are all static and served-from-the-file. 

#### Reason for dynamic generation

This is because we have established that we'd like to keep dynamically generating the Codehilite CSS such that pygments styling work appropriately in https://github.com/zauberzeug/nicegui/issues/4774#issuecomment-2897824724

#### Code implementation highlights

This implementation is much better than the `core.app.get(CODEHILITE_CSS_URL)(lambda: PlainTextResponse(...)`, in a sense that: 

- Properly uses `app.get` as a decorator, instead of a sketchy way. 
- Definition and serving of dynamic resources is centralized. 
- Works against match-all `@ui.page("/{_:path}")`, as much as the rest of the resources work. 

The dynamic resources are available at `f'/_nicegui/{__version__}/dynamic_resources/{{filename}}'`. Note that there is no `key`, since we don't need to do cachebusting since we should always serve these resources with caching turned off. 

#### Notable point

In #4540, a middleware will be introduced. We need to modify it so that `dynamic_resources` does not get the long `cache_control` header.


evnchn:
As referenced by https://github.com/zauberzeug/nicegui/issues/4774#issuecomment-2898906052, the test script boils down to: 

```py
@ui.page("/")
@ui.page("/{_:path}")
def main():
    ui.markdown("""```python
print("hello world")
```""")
```

This PR: 

{C0EE9D9A-F751-4CD7-9E95-EADD0C67B69E}

Main: 

{DA72FE82-FCEB-40F3-8DDB-352DD4872FF5}

falkoschindler (approved):
Thanks @evnchn for another very useful pull request!

I just renamed
- `filename` to `name` because the resource isn't a file, and
- `result_callable` to `function` because this is simpler and I can't find a really good name.

And I fixed the user test without the need for a regex.

Ready to merge! 🙃

evnchn:
Good. 

I originally had `callable` as the parameter, but we had `<built-in function callable>` in the way. There is no `function`, so we can safely use that.

PR #4779, opened 2025-05-22
Warn if non-relative, since cross-origin download has many problems

### LINKED ISSUE #4723: ui.download.from_url() opens url for download and ignores filename
### First Check

### Example Code

```python
from nicegui import ui

ui.button(icon='download', on_click=lambda: ui.download.from_url('https://nicegui.io/logo.png', filename='my_custom_name.png'))

ui.run()
```

### Description

Using `ui.download.from_url()` with the `filename` parameter specified, I expect that the downloaded is triggered directly (and not opened in a new tab first) and that the file will be served with whatever filename was passed. However, the file is opened in a new tab instead, and if this triggers a download, it retains the original name of the server. 

In the example found in the [documentation](https://nicegui.io/documentation/download#download_from_a_url), the same file is directly downloaded and not opened in a new tab. The example does not show the usage of the `filename`-parameter.

### NiceGUI Version

2.16.1.dev0

### Python Version

Python 3.13.0

### Browser

Chrome

### Operating System

Windows

### Additional Context

Tested on Chrome and Edge

evnchn:
This PR fix #4723 to avoid a common pitfall that is cross-origin download on the browser-side. 

Due to browser security reasons outlined https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#download, if it is cross-origin, then the browser ignores the `download` and thus `filename` and `media_type` alike. 

The conclusion from #4723 appears to be, there is no legitimate use for running `ui.download.from_url` from a cross-origin URL. Since it'll only work as the user expected only if: 
- Server responds a file type the browser can't otherwise open a new tab,
- User did not pass filename (which would otherwise be ignored). 

Thus, it is recommended that either: 
- User change their code to relative URL
- Or if they insist, simply use `ui.navigate.to` with `new_tab = True` (https://nicegui.io/documentation/navigate#ui_navigate_to_(formerly_ui_open))

### DIFF
--- a/nicegui/functions/download.py
+++ b/nicegui/functions/download.py
@@ -3,6 +3,7 @@ from typing import Optional, Union
 
 from .. import core, helpers
 from ..context import context
+from ..logging import log
 
 
 class Download:
@@ -55,6 +56,14 @@ class Download:
         :param filename: name of the file to download (default: name of the file on the server)
         :param media_type: media type of the file to download (default: "")
         """
+        is_relative = url.startswith('/') or url.startswith('./') or url.startswith('../')
+        if not is_relative:
+            log.warning(f'ui.download.from_url: You have initiated a cross-origin download from {url}.')
+            log.warning('If cross-origin, browser ignores the instruction to start a download, and tries to view the file in a new tab if possible (such as images, PDFs, etc.). \n'
+                        'Incidentally, download may work for some file types (such as .zip, .db, etc.), but it is not guaranteed to work for all file types.\n')
+            if filename is not None or media_type != '':
+                log.warning(
+                    'Moreover, browser ignores filename and media_type parameters, respecting the origin server\'s headers instead.')
         context.client.download(url, filename, media_type)
 
     def content(self, content: Union[bytes, str], filename: Optional[str] = None, media_type: str = '') -> None:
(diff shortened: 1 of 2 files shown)


evnchn:
Test script, btw:

```py
@app.get('/relative_url')
def relative_url():
    return 'This is a relative URL response.'


@ui.page('/')
def main():
    ui.button('Download relative URL with filename',
              on_click=lambda: ui.download.from_url('/relative_url', 'myresponse.txt', 'text/plain'))
    ui.button('Download absolute URL (JSON, viewable by browser)',
              on_click=lambda: ui.download.from_url('https://nicegui.io/static/sitewide_index.json'))
    ui.button('Download absolute URL (ZIP, not viewable by browser) with filename',
              on_click=lambda: ui.download.from_url('https://github.com/zauberzeug/nicegui/archive/refs/heads/main.zip', 'mysourcecode.zip'))
```

1. OK, no problem
2. Warning, without the "moreover" phrase since no filename is defined
3. Warning and "moreover" phrase since filename is defined

falkoschindler (approved):
Thanks, @evnchn!

I decided to move the lengthy explanation to the documentation and shorten the warning instead. I also updated the demo and other documentation to mention "relative URLs".

And I thought about deprecating absolute URLs and warning about them raising an error in future versions. But actually we don't achieve anything by _enforcing_ relative URLs. If a user keeps using absolute URLs and ignores (or doesn't see) our warning, opening files in a new tab is still better than crashing the server. So we better stick with a warning.

Ready to merge if you ask me. 🙂

PR #4783, opened 2025-05-23
Add flake8-quotes information to pyproject.toml for ruff fixable

EmberLightVFX:
### Motivation

I kept having "problems" with using " instead of ' for strings when commiting changes. As we're currently using ruff for some fixing

```toml
fixable = [
    "I",  # isort
    "RUF022", # `__all__` is not sorted
]
```
and
```json
  "[python]": {
    "editor.defaultFormatter": "ms-python.autopep8",
    "editor.codeActionsOnSave": {
      "source.fixAll.ruff": "always"
    }
  }
```

Adding quote fixing to the fixable list would automatically fix this and we'll have less build-errors when commiting.

In an earlier PR I made (https://github.com/zauberzeug/nicegui/pull/4368) we come to the conclusion to not use ruff for the formating but some time after fixable got added so I don't see any problems adding ruff's quote fixing there as it only will fix the quotes.

### Progress

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,6 +137,12 @@ target-version = "py38"
 indent-width = 4
 line-length = 120
 
+[tool.ruff.lint.flake8-quotes]
+inline-quotes = "single"
+multiline-quotes = "single"
+docstring-quotes = "double"
+avoid-escape = true
+
 [tool.ruff.lint]
 # See complete list: https://docs.astral.sh/ruff/rules/
 select = [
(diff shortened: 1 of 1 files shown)


EmberLightVFX:
Yes, that's true. The pre-commit do fix it but that means that you'll first get an error that the commit didn't pass, re-stage the file/files and commit again. This is such a small fix that would help people not fixing/possible fixing the pre-commit for their PRs.

I don't see adding this small change would make super big difference for the over all project but it would help formatting the code while working on it directly. Like we're already fixing imports etc when saving the file so why not also fix quotes?

falkoschindler (approved):
Thanks, @evnchn! It's a great idea to let Ruff fix quotes on the fly. And having a second tool taking care of it is good because not everyone is using pre-commit.
I even decided to let Ruff fix the other quotation rules as well.
The VSCode setting for fixing the end-of-file is also nice.
Ready to merge!

PR #4790, opened 2025-05-24
Preload documentation pages

### LINKED ISSUE #4728: Subpages in documentation inaccessible until corresponding documentation page is visited
### First Check

### Example Code

Kindly search "@ui.page" in website, but EXCLUDE commented results which begins with "# "

Or. use the following regex: 

```regex
^(?!\s*#)\s*.*?@ui\.page
```

Or, check my search result below

### Description

One day, I was trying to improve documentation, but then I saw a sudden 404 at https://nicegui.io/local_refreshable_c, which I have failed to recreate. 

I looked into it by starting my own server, and realized, that _on a fresh server start_, all of the subpages defined in `@ui.page` in the documentation's code content are not actually created because they are nested in a function. 

Only when the corresponding documentation page is opened, those subpages are then created and the page works. 

Consider: 
- `/local_refreshable_c` is unavailable, until you visit `/documentation/refreshable`
- `/page_layout` is unavailable, until you visit `/documentation/section_page_layout`

Though I do not know about the details since I'm not in charge of the website hosting, I highly suspect that, since we have multiple instances, that the server which served me https://nicegui.io/local_refreshable_c on that day, did not have any visitors attempt to open https://nicegui.io/documentation/refreshable from its spinup till that moment. 

Perhaps currently the page scales down to one server, and so the issue does not show. 

Other than that, the impact is pretty minor, since one is always expected to visit the documentation page before clicking into one of the demos. 

### NiceGUI Version

2.16.1, current main

### Python Version

3.12.9

### Browser

Chrome

### Operating System

Windows

### Additional Context

Affected pages:

Image

I'm not sure whether there's existing issue, because searching for "docs" or "documentation" in the issues page leads to so many issues being filtered out. However, seeing the behaviour still exists, I doubt it has be raised before.

rodja:
### Motivation

In #4728, @evnchn noticed that documentation demo pages are only available after the corresponding sections have been visited once.

### Implementation

This PR fixes #4728 by executing each demo function once and thereby registering all defined pages.

### Progress

- [x] Pytests are not necessary


evnchn:
Actually I have tried my hands on preloading documentation pages, but I was tired and the code was awful so I didn't save my progress. 

But I can see what's going on here after the exam on 26 and 28. Perhaps other can join in as well. 

Still, my concerns, last attempt and this attempt, is the side-effects of this change, especially to other pages who don't need preloading. 

I'm thinking possibly a more simplistic manner: 

- Copy-paste the page definition into `main.py` or some other file. 
- Preferably with ability to "copy source code from there" in the documentation so that no need repeat. 

But I haven't got around to trying that yet.

evnchn:
I think #4799 supersedes this PR. 

Preloading _all_ documentation pages may or may not be a good idea, given the wide variety in nature of content among the parts across those pages, and #4799 offers basically an on-off switch, in a sense that you can switch on the new behaviour (and deal with the bugs that it may cause if you use fancy awaits) or revert to the old behaviour. 

Seems to capture the essense of "eliminating side-effects of this change, especially to other pages who don't need preloading"

PR #4732, opened 2025-05-10
Show the documentation hierarchy in the sidebar

evnchn:
This PR shows the documentation hierarchy in the sidebar. 

- Completes https://github.com/orgs/zauberzeug/projects/6/views/1?pane=issue&itemId=82941340
- Mentioned in https://github.com/zauberzeug/nicegui/discussions/2976

{A8984628-61CD-40C3-BEEF-7B7834085616}

{40313142-2528-40CD-92E5-75E6773CBDE0}

---

_If, in the unlucky case that you have found some deficiencies in the tree generation code, here is an old copy of the code with a lot of debugging prints:_ 

<details>
<summary> Ugly code ahead! </summary>

```py
@ui.page('/debug_documentation')
def _debug_documentation_page() -> None:
    def display_dict_pretty(d: dict) -> None:
        for k, v in d.items():
            ui.label(f'{k}: {v}').classes('text-xs font-mono')
    ui.label('Debug documentation')

    all_registry_keys = [k for k in documentation.registry.keys()]

    single_key_empty = {k: v for k, v in documentation.registry.items() if k == ''}
    no_back_link = {k: v for k, v in documentation.registry.items() if k != '' and v.back_link is None}
    rest = {k: v for k, v in documentation.registry.items() if k != '' and v.back_link is not None}

    """ui.label('Single key empty:').classes('font-bold')
    display_dict_pretty(single_key_empty)

    for k, v in documentation.content.overview.tiles:
        ui.label(k.__name__.rpartition(".")[2]).classes('text-xs font-mono')

    ui.label('No back link:').classes('font-bold')
    display_dict_pretty(no_back_link)

    ui.label('Rest:').classes('font-bold')
    display_dict_pretty(rest)"""

    all_registry_keys.remove('')

    # First build the adjacency list
    adjacency_list: List[tuple[str, str, str]] = []
    for k, v in documentation.content.overview.tiles:
        adjacency_list.append(('', k.__name__.rpartition(
            ".")[2], documentation.registry[k.__name__.rpartition(".")[2]].title))
        all_registry_keys.remove(k.__name__.rpartition(".")[2])

    """for k, v in rest.items():
        adjacency_list.append((v.back_link, k))"""

    i = 0
    while i < len(adjacency_list):
        if i > 1000000:
            break  # no way
        _, v, _ = adjacency_list[i]
        if '#' in v:
            i += 1
            continue
        registry_entry = documentation.registry.get(v)
        if registry_entry and registry_entry.parts:
[...]

PR #4796, opened 2025-05-26
Browser Data Store: Retransmission Avoidance (-40% for docs page alone)

evnchn:
### Motivation

As outlined in #4794, I felt that the NiceGUI site became more sluggish after #4732. Be it true or not, I found that over 30% of the page's content data is the sidebar hierarchy, which is sent again and again on every page load. 

This PR introduces a two new mechanisms: 

- Browser Data Store (at `app.browser_data_store`): 
  - **[User usage]** which the user can write strings (strings only, for ease of hashing) to, 
  - **[End result]** that they can **rest assured NiceGUI will help them keep synchronized** with what is in the browser's `localStorage`
- Fetch from Browser Data Store: (at `client.fetch_xxx_from_browser_data_store`, where `xxx` can be `string`, `list`, or `dict`), 
  - **[User usage]** which the user can use as a placeholder of the large amount of data, 
  - **[End result]** and upon `nicegui.js` `parseElements`, it will replace such placeholders with _whatever's in the corresponding Browser Data Store key_, achieving the outcome of **retransmission avoidance**. 

Real motivation: @falkoschindler ["I'm not sure how this can work in general."](https://github.com/zauberzeug/nicegui/discussions/4794#discussioncomment-13270220) You know I like to challenge the impossible. 

#### Browser Data Store synchronization routine: 

- If populated, browser sends cookie `nicegui_data_store` with the keys it has and the hash of those values to the server on every request. 
- Server inspects such cookie if present, checks the hash, and sends only the keys it needs to send to update the browser with the latest data from the Browser Data Store. 
- Browser receives the data from `updateBrowserDataStore` and updates `nicegui_data_store` such that next request (if nothing has changed) it gets to save a re-transmission. 

Notably: 
- Server also sends value `None` (which is `null` in JSON-form) so as to delete old keys. 

#### Fetch from Browser Data Store routine:

- To avoid plain-text injection, the fetching is triggered with a random token in `client.browser_data_store_token` which is also transmitted to the browser. In such a manner, people who control the arbitrary input to elements (e.g. `ui.label(user_provided_input)` can NOT trigger the fetch, since they don't know the random token. 
  - For injection to occur, the `user_provided_input` must read `ui.context.client.browser_data_store_token` dynamically on client creation, which is impossible for a string / dict / list. 
- `nicegui.js` `parseElements` `JSON.parse` is passed with a custom `reviver`, which detects for these values and replace them (using Python notation for clarity):
  - If it finds **string** `"BDS-TOKEN-sometoken:long_string"`, it replaces the string with the value of key `long_string` in localStorage
  - If it finds **list** `["BDS-TOKEN-sometoken", "long_string"]`, it replaces the list with the JSON-decoded value of key `long_string` in localStorage
  - If it finds **dict** `{"BDS-TOKEN-sometoken": "long_string"}`, it replaces the dict with the JSON-decoded value of key `long_string` in localStorage

### Progress

-> We need to think about whether we need pytest. Documentation seems necessary but I'd like to get someone else to look at the idea before doing documentation, since this idea is quite big!

### Results showcase

Before: 

image

-> 30.3 KB

After: 

image

-> 18.3 KB

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,7 +7,7 @@ from fastapi import HTTPException, Request
 from fastapi.responses import RedirectResponse
 from starlette.middleware.sessions import SessionMiddleware
 
-from nicegui import app, ui
+from nicegui import app, json, ui
 from website import anti_scroll_hack, documentation, fly, imprint_privacy, main_page, svg
 
 # session middleware is required for demo in documentation
@@ -24,6 +24,7 @@ app.add_static_file(local_file=svg.PATH / 'logo_square.png', url_path='/logo_squ
 
 documentation.build_search_index()
 documentation.build_tree()
+app.browser_data_store['side_tree_hierarchy'] = json.dumps(documentation.tree.nodes)
 
 
 @app.post('/dark_mode')
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -5,7 +5,7 @@ import signal
 import urllib
 from enum import Enum
 from pathlib import Path
-from typing import Any, Awaitable, Callable, Iterator, List, Optional, Union
+from typing import Any, Awaitable, Callable, Dict, Iterator, List, Optional, Union
 
 from fastapi import FastAPI, HTTPException, Request, Response
 from fastapi.responses import FileResponse
@@ -45,6 +45,7 @@ class App(FastAPI):
         self._connect_handlers: List[Union[Callable[..., Any], Awaitable]] = []
         self._disconnect_handlers: List[Union[Callable[..., Any], Awaitable]] = []
         self._exception_handlers: List[Callable[..., Any]] = [log.exception]
+        self.browser_data_store: Dict[str, str] = {}
 
     @property
     def is_starting(self) -> bool:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -19,6 +19,7 @@ from .awaitable_response import AwaitableResponse
 from .dependencies import generate_resources
 from .element import Element
 from .favicon import get_favicon_url
+from .helpers import hash_data_store_entry
 from .javascript_request import JavaScriptRequest
 from .logging import log
 from .observables import ObservableDict
@@ -67,6 +68,7 @@ class Client:
         self._deleted = False
         self._socket_to_document_id: Dict[str, str] = {}
         self.tab_id: Optional[str] = None
+        self.browser_data_store_token = f'BDS-TOKEN-{uuid.uuid4().hex}'
 
         self.page = page
         self.outbox = Outbox(self)
@@ -87,6 +89,15 @@ class Client:
         self.disconnect_handlers: List[Union[Callable[..., Any], Awaitable]] = []
 
         self._temporary_socket_id: Optional[str] = None
+
+    def fetch_string_from_browser_data_store(self, key: str) -> str:
+        return f'{self.browser_data_store_token}:{key}'
+
+    def fetch_list_from_browser_data_store(self, key: str) -> List[str]:
+        return [self.browser_data_store_token, key]
+
+    def fetch_dict_from_browser_data_store(self, key: str) -> Dict[str, str]:
+        return {self.browser_data_store_token: key}
 
     @property
     def is_auto_index_client(self) -> bool:
@@ -128,6 +139,18 @@ class Client:
         """Build a FastAPI response for the client."""
         self.outbox.updates.clear()
         prefix = request.headers.get('X-Forwarded-Prefix', request.scope.get('root_path', ''))
+        client_declared_data_store_entries_string: str = request.cookies.get('nicegui_data_store', '{}')
+        client_declared_data_store_entries: Dict[str, str] = json.loads(client_declared_data_store_entries_string)
+        filtered_browser_data_store = {
+            key: value
+            for key, value in core.app.browser_data_store.items()
+            if hash_data_store_entry(value) != client_declared_data_store_entries.get(key, '')
+        }
+        # value = None, for keys which the client declare contain, but not exist in the server's browser data store anymore
+        for key in client_declared_data_store_entries:
+            if key not in core.app.browser_data_store:
+                filtered_browser_data_store[key] = None
+        filtered_browser_data_store = json.dumps(filtered_browser_data_store)
         elements = json.dumps({
             id: element._to_dict() for id, element in self.elements.items()  # pylint: disable=protected-access
         })
@@ -165,6 +188,12 @@ class Client:
                 'prefix': prefix,
                 'tailwind': core.app.config.tailwind,
                 'prod_js': core.app.config.prod_js,
+                'filtered_browser_data_store': filtered_browser_data_store.replace('&', '&amp;')
+                .replace('<', '&lt;')
+                .replace('>', '&gt;')
+                .replace('`', '&#96;')
+                .replace('$', '&#36;'),
+                'browser_data_store_token': self.browser_data_store_token,
                 'socket_io_js_query_params': socket_io_js_query_params,
                 'socket_io_js_extra_headers': core.app.config.socket_io_js_extra_headers,
                 'socket_io_js_transports': core.app.config.socket_io_js_transports,
--- a/nicegui/helpers.py
+++ b/nicegui/helpers.py
@@ -53,6 +53,11 @@ def is_file(path: Optional[Union[str, Path]]) -> bool:
         return False
 
 
+def hash_data_store_entry(entry: str) -> str:
+    """Hashes the data store entry, which must be a string."""
+    return hashlib.sha256(entry.encode('utf-8')).hexdigest()
+
+
 def hash_file_path(path: Path, *, max_time: Optional[float] = None) -> str:
     """Hash the given path based on its string representation and optionally the last modification time of given files."""
     hasher = hashlib.sha256(path.as_posix().encode())
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -8,14 +8,122 @@ let mounted_app = undefined;
 const loaded_libraries = new Set();
 const loaded_components = new Set();
 
-function parseElements(raw_elements) {
+function setCookie(cname, cvalue, exdays) {
+  const d = new Date();
+  d.setTime(d.getTime() + (exdays * 24 * 60 * 60 * 1000));
+  let expires = "expires=" + d.toUTCString();
+  document.cookie = cname + "=" + cvalue + ";" + expires + ";path=/";
+}
+
+function updateBrowserDataStore(str_in) {
+  dict_in = JSON.parse(str_in
+    .replace(/&#36;/g, "$")
+    .replace(/&#96;/g, "`")
+    .replace(/&gt;/g, ">")
+    .replace(/&lt;/g, "<")
+    .replace(/&amp;/g, "&")
+  );
+  // Maintain a list of all keys in __nicegui_data_store_keys__ in localStorage
+  let allKeys = [];
+  try {
+    const storedKeys = localStorage.getItem("__nicegui_data_store_keys__");
+    if (storedKeys) {
+      allKeys = JSON.parse(storedKeys);
+      if (!Array.isArray(allKeys)) allKeys = [];
+    }
+  } catch (e) {
+    allKeys = [];
+  }
+
+  // Prepare for hashing
+  const encoder = new TextEncoder();
+  const hashPromises = [];
+  const keysToAdd = [];
+
+  // Iterate dict_in once for all operations
+  for (const [key, value] of Object.entries(dict_in)) {
+    if (value === null || value === undefined) {
+      localStorage.removeItem(key);
+      // Remove key from allKeys if present
+      const idx = allKeys.indexOf(key);
+      if (idx !== -1) allKeys.splice(idx, 1);
+      console.log(`Removed localStorage key "${key}"`);
+    } else {
+      try {
+        console.log(`Setting localStorage key "${key}" to value:`, value);
+        localStorage.setItem(key, value); // value is always string
+        if (!allKeys.includes(key)) allKeys.push(key);
+      } catch (e) {
+        console.error(`Error setting localStorage key "${key}":`, e);
+      }
+    }
+  }
+
+  localStorage.setItem("__nicegui_data_store_keys__", JSON.stringify(allKeys));
+
+  // Hash values for all keys in __nicegui_data_store_keys__ for nicegui_data_store
+  // Also hash any keys that were already present but not updated in this call
+  for (const key of allKeys) {
+    const value = localStorage.getItem(key);
+    if (value !== null) {
+      hashPromises.push(
+        (async () => {
+          try {
+            const data = encoder.encode(value);
+            const hashBuffer = await crypto.subtle.digest('SHA-256', data);
+            const hashArray = Array.from(new Uint8Array(hashBuffer));
+            const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+            return { key, hashHex };
+          } catch (e) {
+            console.error(`Error hashing value for key "${key}":`, e);
+            return null;
+          }
+        })()
+      );
+    }
+  }
+  console.log("Hash promises:", hashPromises);
+
+  Promise.all(hashPromises).then(results => {
+    const data_store = {};
+    results.forEach(item => {
+      if (item) {
+        data_store[item.key] = item.hashHex;
+      }
+    });
+    console.log("Updated data_store:", data_store);
+    setCookie("nicegui_data_store", JSON.stringify(data_store), 365);
+  });
+}
+
+function parseElements(raw_elements, token) {
+  console.log("Token is", token);
+  function reviver(_, value) {
+    string_detection_prefix = token + ":";
+    if (typeof value === "string" && value.startsWith(string_detection_prefix)) {
+      const localKey = value.slice(string_detection_prefix.length);
+      const storedValue = localStorage.getItem(localKey);
+      console.log(`String: Retrieving "${localKey}" from localStorage, since found ${value} in the elements as a string.`);
+      return storedValue;
+    } else if (Array.isArray(value) && value[0] === token) {
+      const storedValue = JSON.parse(localStorage.getItem(value[1]));
+      console.log(`Array: Retrieving "${value[1]}" from localStorage, since found ${value} in the elements as an array.`);
+      return storedValue;
+    } else if (value !== null && typeof value === "object" && token in value) {
+      const storedValue = JSON.parse(localStorage.getItem(value[token]));
+      console.log(`Dictionary: Retrieving "${value[token]}" from localStorage, since found ${JSON.stringify(value)} in the elements as an object.`);
+      return storedValue;
+    }
+    return value;
+  }
   return JSON.parse(
     raw_elements
       .replace(/&#36;/g, "$")
       .replace(/&#96;/g, "`")
       .replace(/&gt;/g, ">")
       .replace(/&lt;/g, "<")
-      .replace(/&amp;/g, "&")
+      .replace(/&amp;/g, "&"),
+    reviver
   );
 }
 
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -45,7 +45,9 @@
       <span>{{ translations.trying_to_reconnect }}</span>
     </div>
     <script type="module">
-      const app = createApp(parseElements(String.raw`{{ elements | safe }}`), {
+      updateBrowserDataStore(String.raw`{{ filtered_browser_data_store | safe }}`);
+
+      const app = createApp(parseElements(String.raw`{{ elements | safe }}`, "{{browser_data_store_token}}"), {
         version: "{{ version }}",
         prefix: "{{ prefix | safe }}",
         query: {{ socket_io_js_query_params | safe }},
--- a/website/documentation/rendering.py
+++ b/website/documentation/rendering.py
@@ -19,7 +19,8 @@ def render_page(documentation: DocumentationPage, *, with_menu: bool = True) -> 
         with ui.left_drawer() \
                 .classes('column no-wrap gap-1 bg-[#eee] dark:bg-[#1b1b1b] mt-[-20px] px-8 py-20') \
                 .style('height: calc(100% + 20px) !important') as menu:
-            tree = ui.tree(nodes, label_key='title').classes('w-full').props('accordion no-connectors')
+            tree = ui.tree(ui.context.client.fetch_dict_from_browser_data_store('side_tree_hierarchy'),
+                           label_key='title').classes('w-full').props('accordion no-connectors')
             tree.add_slot('default-header', '''
                 <a :href="'/documentation/' + props.node.id" onclick="event.stopPropagation()">{{ props.node.title }}</a>
             ''')
--- a/website/svg.py
+++ b/website/svg.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 
-from nicegui import ui
+from nicegui import app, ui
 
 PATH = Path(__file__).parent / 'static'
 HAPPY_FACE_SVG = (PATH / 'happy_face.svg').read_text(encoding='utf-8')
@@ -9,25 +9,33 @@ GITHUB_SVG = (PATH / 'github.svg').read_text(encoding='utf-8')
 DISCORD_SVG = (PATH / 'discord.svg').read_text(encoding='utf-8')
 REDDIT_SVG = (PATH / 'reddit.svg').read_text(encoding='utf-8')
 
+app.browser_data_store['happy_face_svg'] = HAPPY_FACE_SVG
+app.browser_data_store['nicegui_word_svg'] = NICEGUI_WORD_SVG
+app.browser_data_store['github_svg'] = GITHUB_SVG
+app.browser_data_store['discord_svg'] = DISCORD_SVG
+app.browser_data_store['reddit_svg'] = REDDIT_SVG
+
+app.browser_data_store['half_happy_face_svg'] = HAPPY_FACE_SVG.replace(
+    'viewBox="0 0 62.44 71.74"', 'viewBox="31.22 0 31.22 71.74"')
+
 
 def face(half: bool = False) -> ui.html:
-    code = HAPPY_FACE_SVG
     if half:
-        code = code.replace('viewBox="0 0 62.44 71.74"', 'viewBox="31.22 0 31.22 71.74"')
-    return ui.html(code)
+        return ui.html(ui.context.client.fetch_string_from_browser_data_store('half_happy_face_svg'))
+    return ui.html(ui.context.client.fetch_string_from_browser_data_store('happy_face_svg'))
 
 
 def word() -> ui.html:
-    return ui.html(NICEGUI_WORD_SVG)
+    return ui.html(ui.context.client.fetch_string_from_browser_data_store('nicegui_word_svg'))
 
 
 def github() -> ui.html:
-    return ui.html(GITHUB_SVG)
+    return ui.html(ui.context.client.fetch_string_from_browser_data_store('github_svg'))
 
 
 def discord() -> ui.html:
-    return ui.html(DISCORD_SVG)
+    return ui.html(ui.context.client.fetch_string_from_browser_data_store('discord_svg'))
 
 
 def reddit() -> ui.html:
-    return ui.html(REDDIT_SVG)
+    return ui.html(ui.context.client.fetch_string_from_browser_data_store('reddit_svg'))


evnchn:
Test script: 

```py
from nicegui import json, ui, app
import uuid


def generate_important_data() -> str:
    """Generates a unique identifier for important data."""
    # This function generates a UUID to represent important data.
    # It can be used to track or reference specific data points in the application.
    return f"IMPORTANT_DATA_{uuid.uuid4()}"


def generate_long_string() -> str:
    """Generates a long string for testing purposes."""
    return generate_important_data() + 'This is a very long string that you don\'t want to transmit again and again. ' * 100


def put_long_string_in_browser_data_store() -> None:
    """Button handler to put a long string in the browser data store."""
    app.browser_data_store['long_string'] = generate_long_string()


def generate_large_tree() -> list:
    """Generates a large tree structure for testing purposes."""
    return [
        {'id': f'node_{i} {generate_important_data()}', 'children': [{'id': f'child_{i}_{j}'} for j in range(20)]}
        for i in range(5)
    ]


def put_tree_in_browser_data_store() -> None:
    """Button handler to put a large tree in the browser data store."""
    app.browser_data_store['large_tree'] = json.dumps(generate_large_tree())


def generate_large_table() -> dict:
    """Generates a large table structure for testing purposes."""
    return {
        'defaultColDef': {'flex': 1},
        'columnDefs': [
        ],
        'rowData': [
            {
                'name': f'Name {i} {generate_important_data()}',
                'age': 20 + i,
                'parent': f'Parent {i}'
            }
            for i in range(50)
        ],
        'rowSelection': 'multiple',
    }


def put_large_table_in_browser_data_store() -> None:
    """Button handler to put a large table in the browser data store."""
    app.browser_data_store['large_table'] = json.dumps(generate_large_table())


@ui.page('/')
def main_page() -> None:
    """Main page of the application."""

    if not app.browser_data_store.get('long_string'):
        put_long_string_in_browser_data_store()
    if not app.browser_data_store.get('large_tree'):
        put_tree_in_browser_data_store()
    if not app.browser_data_store.get('large_table'):
        put_large_table_in_browser_data_store()

    ui.label('Welcome to the Browser Data Store Test Page!')

    ui.label("Label with Browser Data Store")
    ui.button('Put New Long String in Browser Data Store', on_click=lambda: (
        put_long_string_in_browser_data_store(), ui.navigate.reload()))
    ui.label(ui.context.client.fetch_string_from_browser_data_store(
        'long_string')).classes('max-h-40 overflow-y-scroll')

    ui.label("Tree with browser data store")
    ui.button('Put New Tree in Browser Data Store', on_click=lambda: (
        put_tree_in_browser_data_store(), ui.navigate.reload()))
    ui.tree(ui.context.client.fetch_list_from_browser_data_store('large_tree'), label_key='id')

    ui.label("AG Grid with browser data store")
    ui.button('Put New Large Table in Browser Data Store', on_click=lambda: (
        put_large_table_in_browser_data_store(), ui.navigate.reload()))
    ui.aggrid(ui.context.client.fetch_dict_from_browser_data_store('large_table')).classes('max-h-40')

    if app.browser_data_store.get('key_sometimes_missing'):
        ui.button(
            'Remove Key from Browser Data Store',
            on_click=lambda: (
                app.browser_data_store.pop('key_sometimes_missing', None),
                ui.navigate.reload()
            )
        )
    else:
        ui.button(
            'Put Key in Browser Data Store',
            on_click=lambda: (
                app.browser_data_store.update({'key_sometimes_missing': 'value'}),
                ui.navigate.reload()
            )
        )


ui.run()
```
Notice the size of the response and console logs for the following: 

- Fresh browser, no cookies
- Just eload
- Press any of the button and the reload afterwards

-> You may need to Disable Cahce!

falkoschindler:
Caching individual page content is an interesting idea, @evnchn! We'll certainly need some time to look into it and give feedback. But can you show a brief "hello world" demo on how to use the new feature? This way we can start with a high-level overview before getting into implementation details. Thanks!

evnchn:
TODO: 

- [ ] If the page does indeed serve over 5MB of content to be shoved into browser data store, the data which has exceeded capacity can't fit in localStorage, and thus on the replace, it fails. We should still server best-effort by putting the data that can't fit in a dictionary, such that although yes we aren't caching the content, it means that we are not dropping content.

rodja:
Your idea has a lot of potential @evnchn. But I think the API with `app.browser_data_store['...'] = json.dumps(...)` and `ui.context.client.fetch_string_from_browser_data_store` a bit complicated. When doing wishful programming, I could imagine to do:

```py
def some_static_ui():
    ui.label('Hello NiceGUI')

@ui.page('/')
def index():
    ui.label('Index Page')
    ui.cached_content(some_static_ui)

@ui.page('/other')
def other():
    ui.label('Other Page')
    ui.cached_content(some_static_ui)
```

What do you think? Is that possible?

evnchn:
Managed to got it working with 2 core changes: 

1. Separate the element' keys into static and dynamic, cache only the static ones, and serving only the dynamic ones on cache hit. 
2. On browser side, if it's a dict, apply the rest of the original keys (aka the dynamic keys of an element) if we got a hit. 

Notably: 

- `element._populate_browser_data_store_if_needed()` is needed otherwise the cache will be empty on first page load, instructing the client to drop all keys, while also serving a request which uses said keys (aka it doesn't load)

_Apparently calling it an "uphill battle" is enough to drive up my spirits to get things done!_

Test script:

```py
from nicegui import ui, app
import uuid

if not app.storage.general.get('special_uuid'):
    app.storage.general['special_uuid'] = str(uuid.uuid4())


@ui.page('/')
def main_page():
    my_cached_label = ui.label('This should be cached'+app.storage.general['special_uuid'])
    my_cached_label.cache_name = 'cached_label'

    columns = [
        {'name': 'name', 'label': 'Name', 'field': 'name', 'required': True, 'align': 'left'},
        {'name': 'age', 'label': 'Age', 'field': 'age', 'sortable': True},
    ]
    rows = [
        {'name': app.storage.general['special_uuid'], 'age': 18},
        {'name': 'Bob', 'age': 21},
        {'name': 'Carol', 'age': 25},
    ]
    my_cached_table = ui.table(columns=columns, rows=rows, row_key='name')
    my_cached_table.cache_name = 'cached_table'

    my_cached_button = ui.button('Click me (here is some long text to test caching)'+app.storage.general['special_uuid'],
                                 on_click=lambda: ui.notify('Button clicked!'))
    my_cached_button.cache_name = 'cached_button'

    with ui.card() as my_cached_card:
        my_cached_card.cache_name = 'cached_card'
        ui.label('This is a cached card'+app.storage.general['special_uuid'])
        my_cached_button_inside_card = ui.button('Click me', on_click=lambda: ui.notify('Card button clicked!'))
        my_cached_button_inside_card.cache_name = 'cached_button_inside_card'

    # another cahced label
    my_cached_label2 = ui.label('This is another cached label'+app.storage.general['special_uuid'])
    my_cached_label2.cache_name = 'cached_label2'

    # button to change the special UUID and reload the page
    ui.button('Change UUID and reload', on_click=lambda: (app.storage.general.update({'special_uuid': str(uuid.uuid4())}),
                                                          ui.navigate.reload()))


ui.run()
```

evnchn:
> Assuming I have a page with a big `ui.markdown`. How can I cache it?

As of https://github.com/zauberzeug/nicegui/pull/4796/commits/2ad9e3c1237134f5b524c6506648ec6424921cc0:

```py
long_markdown_element = ui.markdown("""# This is a long markdown text""")
long_markdown_element.cache_name = 'long_markdown_cache'  # makes it cached
```

Originally I propose:

```py
long_markdown_text = """# This is a long markdown text"""
app.browser_data_store['long_markdown_text'] = long_markdown_text
ui.markdown(ui.context.client.fetch_string_from_browser_data_store('long_markdown_text'))
```

But that doesn't work, since markdown mangles the innerHTML into `<p>BDS-TOKEN-sometoken:long<em>markdown</em>text</p>`, making it impossible to recognize by `nicegui.js` to do the substitution...

rodja:
Why do you need a cache name, @evnchn? Would not an internal uuid be sufficient? Maybe something like

```py
# on the hidden auto-index page, or anywhere else
with ui.column() as my_content:
   ui.label('A')
   ui.label('B')

@ui.page('/')
def index():
    ui.label('main')
    my_content.from_cache()

@ui.page('/other')
def other():
    ui.label('other')
    my_content.from_cache()
```

falkoschindler:
@evnchn I think @rodja doesn't want to _copy_ elements. Instead `from_cache()` would insert some kind of marker with enough information for the client to insert the element from memory.

evnchn:
Having SHA256 has the automatically populated cache name doesn't work. 

Assume I have this in `elements.py`

```py
def cached(self) -> Self:
    """Mark the element as cached.

    This will store the element's data in the browser data store.
    The element can then be referenced by its cache name.
    """
    self.cache_name = f'auto_{hash_data_store_entry(json.dumps(self._to_dict_internal()[0]))}'
    self._populate_browser_data_store_if_needed()
    return self
```

Then, note what happens in the browser console. 

{D89902C6-A8EF-4B4F-B096-68226F0CB389}

After a couple updates to the content, we see that we have more and more old unused cache entries. This is because, when the element is created again with the new content, it has a new cache_name, and the old entries are thus not updated, and left in memory. 

Test script: 
```py
from nicegui import ui, app
import uuid

if not app.storage.general.get('special_uuid'):
    app.storage.general['special_uuid'] = str(uuid.uuid4())


@ui.page('/')
def main_page():
    markdown1 = ui.markdown("# This is markdown element 1 " + app.storage.general['special_uuid'])
    markdown1.cache_name = 'long_markdown_cache'  # makes it cached

    markdown2 = ui.markdown("# This is markdown element 2 " + app.storage.general['special_uuid']).cached()  # auto-cached
    print(markdown2.cache_name)

    # Button to change the special UUID and reload the page
    ui.button('Change UUID and reload', on_click=lambda: (app.storage.general.update({'special_uuid': str(uuid.uuid4())}),
                                                          ui.navigate.reload()))


ui.run()
```

---

This goes to show the importance of an unchanging cache_name for element for caching to properly work, otherwise cache invalidation becomes an issue, which is "one of two hardest things in computer science" 😅

evnchn:
Can we first agree that we need some invariant cache_name / ID, that is NOT dynamically derived from randomness / content being cached? Therefore `.cached()` and `.from_cache()` wont suffice. 

If you're OK with this agreement, 👍 and we can move on, or else 👎 and I will continue my half-written analogy to (hopefully) get you to understand)

rodja:
> Can we first agree that we need some invariant cache_name / ID, that is NOT dynamically derived from randomness / content being cached? Therefore .cached() and .from_cache() wont suffice.

Hmmm.... I thought about using the Python id of the element (`self.cache_name = id(self)') or a uuid generated on first use (`self.cache_name = uuid4() if self.cache_name == None else self.cache_name`).

evnchn:
> (1) create a route serving the content dictionary and (2) to inform the client to fetch the content from this route

This was exactly what I was thinking of in https://github.com/zauberzeug/nicegui/discussions/4794#discussioncomment-13266025, but it's _not the best design_, since you need to make another access to the endpoint to see if you can use such data, slowing page load down (and potentially causing TOCTOU - Time Of Check to Time Of Use errors, since you did the check 100ms later than the initial page load)

This is why in my PR, I shove things into the cookies, in a sense that when you access any NiceGUI page, it basically **bundles the response of the responses that the API would otherwise have to make**

But I will ponder the API design of `with ui.cached_element([id: str]):`. Seems _nice_, but the issue is, if you notice with this PR, that if you currently set `.cache_name` for the parent element, it does nothing to its childrens (since we also can't easily replace the children in `nicegui.js`, since the JSON.parse reviver works on individual value, and the parent&child would otherwise span several values. 

Perhaps `with ui.cached_element([id: str]):` can auto-set the `cache_name` of the childrens. I'll look into it, but IMO the case of repeated childrens, you can just always manually assign the children some `cache_name`...

evnchn:
@rodja Using those dynamically generated IDs, which are ephemeral in nature, **as it stands in this PR**, will not work. Let me offer you 2 counter-examples: 

#### Using `id(self)`

Since the element was initialized many times, in Python it has different IDs. It means we would create a new key in `app.browser_data_store` every time, but we don't clean up the old one. 

#### Using `str(uuid4()))`

Since the element was initialized many times, it has different UUID. It means we would create a new key in `app.browser_data_store` every time, but we don't clean up the old one. 

---

**HOWEVER** I just noticed a possibility: What if when the element is deallocated, we automatically delete the entry in `app.browser_data_storage`? That way, we can use ephemeral IDs, and make the API simplier. However, I am not sure if we will face the issue of, we're trying to delete the key while writing to the same key in another thread? (I don't know asyncio enough to answer this question...)

EmberLightVFX:
Wouldn't it be possible to use inspect to get the function name and use that as the cache-id?
```
import inspect

def foo():
   print(inspect.stack()[0][3])
   print(inspect.stack()[1][3])  # will give the caller of foos name, if something called foo

foo()
```

```
output:

    foo
    <module_caller_of_foo>
```

So combine that with

```
@ui.page('/')
def page():
    ui.cached(foo)
```

(I'm writing on my phone so can't give super clear examples. Hope it's understandable any way!)

evnchn:
`inspect.stack()` would be even more confusing IMO. 

I am thinking, even though we have so-far mentioned that IDs which are random with respect to object initialization's time (Python's `id()`, `str(uuid())`) will not work, since the key flys around. How about if we have the key to be the hash, and then we also delete the key in the object's destruction? 

I'll try it, but it seems like it could work...

evnchn:
With https://github.com/zauberzeug/nicegui/pull/4796/commits/461241cca25a518bf48b1f798981cb6ddb8f4152, I do `.cache(cache_name)` to make it more in line with the rest of the NiceGUI's methods. (Commit name is wrong, it's not `id` since Ruff would be mad otherwise)

It is possible to not give any ID, but then _should you change the cached element's content_, it takes 30 seconds (or whatever your client disconnect timeout is) for the old element to be deleted and its cache invalidated. 

Means that it's fine if the element is known will never change, but could lead to unexpected localStorage overflows if the user refresh rapidly, such that old elements are not invalidated in time, and we have too many cache items. 

---

I am still biased to not allow caching without `cache_name` though.

evnchn:
https://github.com/zauberzeug/nicegui/pull/4796/commits/6fb39574d08d1aa1dfda67fc22d02f33f9b08e6f

```py
from nicegui import ui, app
import uuid

if not app.storage.general.get('special_uuid'):
    app.storage.general['special_uuid'] = str(uuid.uuid4())


@ui.page('/')
def main_page():
    my_cached_label = ui.label('This should be cached'+app.storage.general['special_uuid'])
    my_cached_label.cache('cached_label')

    columns = [
        {'name': 'name', 'label': 'Name', 'field': 'name', 'required': True, 'align': 'left'},
        {'name': 'age', 'label': 'Age', 'field': 'age', 'sortable': True},
    ]
    rows = [
        {'name': app.storage.general['special_uuid'], 'age': 18},
        {'name': 'Bob', 'age': 21},
        {'name': 'Carol', 'age': 25},
    ]
    my_cached_table = ui.table(columns=columns, rows=rows, row_key='name')
    my_cached_table.cache('cached_table')

    ui.label('Elements from cache:').classes('text-lg font-bold')

    ui.label().from_cache('cached_label')  # this will load the cached label
    ui.table(rows=[]).from_cache('cached_table')  # this breaks for no reason

    # button to change the special UUID and reload the page
    ui.button('Change UUID and reload', on_click=lambda: (app.storage.general.update({'special_uuid': str(uuid.uuid4())}),
                                                          ui.navigate.reload()))


ui.run()
```

For some reason, the replacement for the second `ui.table` breaks:

{D20CEAC0-3854-40C4-980B-A526BD5C445B}

All in all, I don't think `from_cache` will really work out 😥

evnchn:
OK so I have gotten to the bottom of why it doesn't work. 

{4D91F1F9-72B4-4334-A195-12FC0A4305B4}

In the Socket.IO level, it is found that we were compressing the data transmission with Browser Data Store. 

{0395E014-1127-4297-8503-7A1571D29446}

This brings an interesting conundrum: 

- The premise of `my_element.from_cache()` is so that we never initialize the element twice, and we use one element `my_element` in one spot only, which is easy to keep consistent. 
- Then, there will always be a "master copy" element, and another "cloned copy" element. In this case, "master copy" is `my_cached_table.cache('cached_table')`, "cloned copy" is `ui.table(rows=[]).from_cache('cached_table')`
- But, these elements will soon fall apart in synchronization, if in the case of `ui.table()`, that we want to mutate the element in any way. Such as `ui.table` immediately sending `update`. 
  - We can keep on refering to the "cloned copy" by the short-form, using the BDS notation, but we'd need to extend the sychronization of the Browser Data Store into Socket.IO, which deviates from the original goal of this PR to reduce the HTTP traffic. 
  - We can detach the "cloned copy", but then:
    - If we did `ui.table(rows=[]).from_cache('cached_table')`, then the underlying element does not have the original rows and columns, and the tables fall out of sync
    - If we did `ui.table(columns=[...], rows=[...]).from_cache('cached_table')`, then we'd be initializing the element twice, and defeat the original point. 

Overall, I believe the API `my_content.from_cache()` can hardly lead anywhere. Moreover, if we want to share an element, we can already do it using Global Refreshable (I think), so that seems a much more elegant solution. I'll continue exploring, but I am more inclined to reverting https://github.com/zauberzeug/nicegui/pull/4796/commits/6fb39574d08d1aa1dfda67fc22d02f33f9b08e6f and call it a day.

rodja:
Interesting observation @evnchn. From my understanding, cached ui elements always have the problem of being decoupled from the Python object. Basically they should be regarded "backend-less". Or not?

evnchn:
Basically, if you really want to make a UI element from cache, **give up trying to update it in any sort of manner**. This means making a `ui.table` from cache, among many other elements where any sort of interactivity, is not possible. 

And my argument was, if the function is so limited, then my earlier APIs `ui.context.client.fetch_XXX_from_browser_data_store` may suffice, where we recycle the data used to make the UI element, not the element itself. 

I also pitched the concept of caching a global refreshable, if you want to derive element from cache, while actually we keep them in sync with each other because global refreshables are global. 

So this is why I am more inclined to drop `element.from_cache()` entirely, and if you do agree, we'll revert https://github.com/zauberzeug/nicegui/commit/6fb39574d08d1aa1dfda67fc22d02f33f9b08e6f to allow this PR to move on.

rodja:
I'm not sure reverting will help to overcome the core issues with the concept. But of course you can try and show me the way 🙃

evnchn:
Using this, we can achieve the goal of "re-use once-defined element objects", without the extremely sketchy `.from_cache(id)` API, which "creates a soul-less doppelganger of an Element". Rather, the concept of define-once-use-many is already well-established from `ui.refreshable`, and we need not reinvent the wheel. 

```py
from nicegui import ui, app
import uuid


def menu():
    ui.link('Main Page', '/')
    ui.link('Page 1', '/page1')
    ui.button('Change UUID and reload',
              on_click=lambda: (app.storage.general.update({'uuid': str(uuid.uuid4())}), ui.navigate.reload()))


if not app.storage.general.get('uuid'):
    app.storage.general['uuid'] = str(uuid.uuid4())


@ui.refreshable
def repeated_elements():
    ui.label(f'This is a long label that should be cached: {app.storage.general["uuid"]}').cache('cached_label')
    ui.table(rows=[{'id': i, 'uuid': app.storage.general['uuid']} for i in range(10)]).cache('cached_table')


@ui.page('/')
def main_page():
    ui.label('This is the main page')
    repeated_elements()
    menu()


@ui.page('/page1')
def page1():
    ui.label('This is page 1')
    repeated_elements()
    menu()


ui.run()
```

Result: 

{B4C251DF-C4E8-4DB2-B9B2-EF2841356BF2}

{F678D775-D397-410D-A7AA-1EF0E809BFF4}

In both pages, the same label and the same table is shown, and as we can see from `Updating browser data store with: {}` at the beginning of the console log, we see that our cached definitions of the element remains valid, and we stick to using that definition on page load. 

With that, IMO https://github.com/zauberzeug/nicegui/commit/6fb39574d08d1aa1dfda67fc22d02f33f9b08e6f serves no value, and I'll revert it. If you want to explore `.from_cache(id)` further, just let me know. Thanks!

rodja:
I found some time to dig deeper into your current implementation. Here my summary within the framework [suggested by Falko](https://github.com/zauberzeug/nicegui/pull/4796#issuecomment-2916544719):

### Use cases

In the current state, the PR supports caching unchanged (static) content when reloading a page and caching same static content on multiple pages. But only for single elements, not (automatically) handling element hierarchies.

### API

The PR uses the `.cache(id=...)` API because `.from_cache()` is breaking the backend/frontend relation. 

### Storage

The currently used server-side data store minimizes http requests. The first access to a page would be longer if HTTP-cachable endpoints would be used (because additional requests must be performed). Same is true for every cache invalidation.

rodja:
And here some comments / thoughts from my side:

- it would be cool if we could allow hierarchies of elements to be cached automatically. For example `with ui.row().cache('navigation'):` may open a context which holds all child elements in cache.
- I'm not sure if `.cache(id=...)` or `with ui.cached_element(id=...)` would be the better API
- while the implementation with browser storage cache is more complicated, I think it's worth it for maximal performance
- is the `Client. browser_data_store_token` really necessary?
- I think it's time to write a bunch of tests to verify the inner workings; it would also help me (and others) to better understand cache loading, invalidating, ...

evnchn:
I'm still thinking about the, `.cache` (individual element), `.cache` (affecting all childrens), or `with ui.cached_element`

---

But I can say, a random-per-client `Client.browser_data_store_token` would be necessary. 

### Core mechanism of the Browser Data Store - Ctrl-F principle

At the end of the day, it is like we are doing Ctrl-F on the JSON data

- See string "TOKEN:hello"? Replace with the string keyed "hello" in localStorage
- See list [TOKEN, "hello"]? Replace with the JSON-decoded list keyed "hello" in localStorage
- See dict {TOKEN: "hello"}? Replace with the JSON-decoded dict keyed "hello" in localStorage

So, immediately, we need _some_ token, or else we'd be replacing plain strings, lists, and dicts, even when no replacement is desired. Question is, does it need to be random per client, or can be a global constant? 

### Discerning Python-land instruction from user's injection attempts

Issue is, we can't really discern whether it is the Python server's intention, or some dude entering arbitrary strings, without a randomized token which is randomized per request and used before any user input is respected. 

Consider the following: 

```py
my_input = ui.input('Enter some string to show')
ui.button('Show the string', on_click=lambda: ui.label(my_input.value))
```

Assume token is `123` and is constant. What's to stop the user from entering "123:hello" into the input? 

Assume the token is `some_random_token` set randomly once on server startup. What's to stop the user from inspecting element and check the token is `some_random_token` as seen in the call to `parseElements`, and use that token to enter "some_random_token:hello" into the input? 

Similar argument applies, when you have code which parses arbitrary JSON from the user, and shortly after pass into a NiceGUI element. 

### Similar example: Cross-Site Request Forgery (CSRF) tokens

Cross-Site Request Forgery (CSRF) tokens are security measures used to protect web applications from unauthorized actions performed by malicious websites. A CSRF attack tricks a user into executing unwanted actions on a different site where they are authenticated.

- **Purpose**: CSRF tokens help ensure that requests sent to a web application are legitimate and intentional.
- **Implementation**: When a user accesses a form, the server generates a unique token and includes it in the form. This token is validated when the form is submitted.
- **Validation**: If the submitted token matches the server's stored token, the request is processed; otherwise, it is rejected.
- **Protection**: CSRF tokens mitigate the risk of attackers performing actions on behalf of authenticated users without their consent.

Consider `Client.browser_data_store_token` a token against **forged placeholders** for the Browser Data Store find-and-replace mechanism. 

### TL-DR

By using a random token per Client instantiation, we ensure that, short of Arbitrary Code Execution in the page function (which if there is, the website developer got bigger problems to worry about), the user CANNOT enter anything in the inputted string, that would inadvertantly trigger the Browser Data Store find-and-replace mechanism.

evnchn:
https://github.com/zauberzeug/nicegui/pull/4796/commits/087b2d9346e4dbaca823aeb78712e3d380c2fbb2

Spotlights: 

- Can enable caching for all child elements, if pass `.cache(..., apply_to_child=True)`
- Child will look at the parent. If parent indicate apply to child, then:
  - Assign its cache name by considering the slot name and order of itself among the elements in that slot (to ensure unique cache name per element across slots)
  - Inherit `auto_id` and `cache_apply_to_child` from parent
- At any time, it is possible to deviate from parent by manually calling `cache()` or `disable_cache()`

Test script: 

```py
from nicegui import ui, app
import uuid

if not app.storage.general.get('uuid'):
    app.storage.general['uuid'] = str(uuid.uuid4())


@ui.page('/')
def main_page():
    def render_section(with_cache_name: bool):
        ui.label('With setting cache name to cached_column' if with_cache_name else 'With automatic cache name')
        column = ui.column().classes('bg-gray-100').cache('cached_column' if with_cache_name else None, apply_to_child=True)
        with column:
            with ui.card():
                ui.label(f'This is a long label that should be cached: {app.storage.general["uuid"]}')
                ui.label(
                    f'This label is inside a card and that should be cached as well: {app.storage.general["uuid"]}')
            ui.table(rows=[{'id': i, 'uuid': app.storage.general['uuid']} for i in range(2)])
            with ui.input(f'Cached input {app.storage.general["uuid"]}').add_slot('append'):
                ui.icon('face')
            ui.label('This label has caching explicitly disabled, despite being inside a cached element').disable_cache()
            with ui.card().cache('cached_card' if with_cache_name else None):
                ui.label(
                    f'Although card is cached, this label is not, since we did not set apply_to_child = True: {app.storage.general["uuid"]}')

    render_section(with_cache_name=True)
    render_section(with_cache_name=False)

    ui.button('Change UUID and reload',
              on_click=lambda: (app.storage.general.update({'uuid': str(uuid.uuid4())}), ui.navigate.reload()))


ui.run()
```

Results: 

{4C653514-1341-465D-96BC-67E10FBF1E3E}

Where: 

- Gray column: `cached_column`
  - First card in gray column: `cached_column.default:0`
    - First and second label in first card in gray column: `cached_column.default:0.default:0` and `cached_column.default:0.default:1`
  - Table in gray column: `cached_column.default:1`
  - Input in gray column: `cached_column.default:2`
    - Icon in input in gray column: `cached_column.default:2.append:0`
  - "caching explicitly disabled" label: nowhere to be found
  - "Although the card is cached" card: `cached_card`
    - Label in "Although the card is cached" card: nowhere to be found

evnchn:
Contemplating whether we should keep the `client.fetch_xxx_from_browser_data_store` methods, since really quite a lot of elements expect to be able to read the str/list/dict input to it, and possibly mutate it before sending it out. Both of which is incompatible if we pass a placeholder as in `client.fetch_xxx_from_browser_data_store`

string case: `ui.markdown` mangles into `<p>BDS-TOKEN-sometoken:long<em>markdown</em>text</p>`

list case: `ui.chat_message` iterates through the list and spawns several individual chat messages, breaking up the list

dict case: haven't found, but I am sure there is. 

@rodja I want to hear what do you think. Are those methods, which doesn't even work half of the time, safe to be deleted, now that we have `element.cache`?

rodja:
> So, immediately, we need some token, or else we'd be replacing plain strings, lists, and dicts, even when no replacement is desired.

I had a hard time to understand your statement and only after contemplating "user's injection attempts" and the analogy to CSRF I now might have an idea of what you mean. Let me rephrase in my own words:

The server sends a json object with elements to the browser and needs a way to mark parts which should be loaded from browser storage. A malicious user might sneak some string into the persistence, which then gets loaded on a new page and tricks the javascript in thinking that it should load some browser storage data. Right?

But why not use a well known key like "CACHE" to mark cached elements. They appear in the json structure as placeholders for entire objects/arrays or root-level values. So even if user enters "CACHE":evil_key" I don't see how the js could confuse this with a cache loading instruction because the entered text will only appear as string value in a property.

> Can enable caching for all child elements, if pass .cache(..., apply_to_child=True)
> it is possible to deviate from parent by manually calling cache() or disable_cache()

What would be scenarios where you want to cache the parent but not the children? It would simplify the API if we always cache children (and hence get rid of `apply_to_child` and `disable_cache`).

> Are [the `client.fetch_*`] methods, which doesn't even work half of the time, safe to be deleted, now that we have element.cache?

We are always looking for minimal, straight forward APIs. So, yes I think its good to remove these methods.

evnchn:
Let's gather for a bit, since dropping `client.fetch_*` methods shifts the landscape. 

Originally: 

- Need to detect placeholders in _any level_ of the JSON hierarchy, of which it is user-controlled
- Need to have a random token to discern
- Cached elements leverages the above mechanism

For example:

`ui.json_editor(USER_CONTROLLED_DICT)` can be mistaken as `ui.json_editor(ui.context.client.fetch_dict_from_browser_data_store('EVIL_KEY'))`, if `USER_CONTROLLED_DICT` turns out to be `{'CACHE': 'EVIL_KEY'}`

---

Now: 

- We drop the `client.fetch_*` methods, and now we need a dedicated mechanism for replacing placeholders for elements (and that only)
- So, placeholders occur in _root-level_ of the JSON hierarchy only, of which it is non-user-controllable (minus some cursed elements which override `_to_dict()` to spit out user-controlled content
- We can discern by a simple `CACHE` key in the element's dictionary

**I think this is getting somewhere soon. Let me work on it later.**

evnchn:
> What would be scenarios where you want to cache the parent but not the children? It would simplify the API if we always cache children (and hence get rid of `apply_to_child` and `disable_cache`).

I was thinking about this 

```py
for _ in range(10): # point 2
    with ui.card().cache('repetitive-card', apply_to_child=True)
        ui.label('This is some boilerplate label that barely changes') # and some other elements which don't really change a lot
        ui.label(dynamic_data()).disable_cache() # point 1
```

1. If `dynamic_data()` will be too large to fit in localStorage, `.disable_cache()` marks it out of the cache, and it'd work
2. If we disable the warning in https://github.com/zauberzeug/nicegui/pull/4796/commits/b3416e6c61da6592f507948ec0438b514a6ac985 to allow the same cache to be used for multiple elements, then we can have several of these cards, and we share the rest of the boilerplate in the ui.card() definition, while the `dynamic_data()` remains different across cards.

evnchn:
Proposed new signature: 

1. `apply_to_child=True` as a default for `.cache()`. Adopt manual opt-out for by disabling cache as below. 
2. For `.cache()`, accept three-way: 
  a. `.cache(name)`: Explicit name
  b. `.cache()`: Automatic name derived from hash
  c. `.cache(False)`: **New**, for disabling cache (does the old functionality of `.disable_cache()`)
3. It is possible to opt-out of caching children by putting those elements inside a placeholder div with cache disabled

```py
with ui.card().cache('cache-just-the-card'), ui.column().cache(None):
    ui.label('This label is not cached, since the `ui.column().cache(None)` stops the propagation of applying the cache to the child')
```

rodja:
Ok. The main reason for not wanting to cache something is the 5 mb limit in the browser. The documentation should educate the users about that limitation. And I'm curious what @falkoschindler thinks about this API. Maybe we should use the plural `apply_to_children`.

evnchn:
Very interesting development so far. 

Status: 

- SVGs are mostly cached, except the variants of the happy face SVG with differing classes, which since I used automatic cache names, pops in and out of the cache as the elements flip-flip from being present and absent across pages. 
- Sidebar hierarchy caching is not working, since the nodes are in key `nodes` of `props` dictionary, and on an element level, we can't cache the element. <- Point of discussion

1. Apparently, under the old code, if we update the element in any way via the Outbox, since we call `element._to_dict()` with nothing else passed, it means that we are not caching the element definition, and we'd send the entire element, making the caching useless if we want to use most of all functionalities which requires writing to the `props` of the element, including `tree.expand` of the sidebar hierarchy. 
  a. In https://github.com/zauberzeug/nicegui/pull/4796/commits/9c3fcfa637b650b738c5b3477c581cc92ed41d1a, you see that we serve the `CACHE` short-form no matter what, and it is now the responsibility of the user to register dynamic keys, in such a manner that they opt-out that part from the caching and into the transmission (still thinking about this since this is quite suboptimal... I am thinking if we do the check to see if the cache content is old, and send the full definition)
  b. Note that the nodes are in key `nodes` of `props` dictionary. We would otherwise need to deep-diff the two dictionaries (even more complex than Project BASE's once-had diffing in https://github.com/zauberzeug/nicegui/commit/21c2fad85183731b9c8e7c50c1dc4766070267e7, which checks the root keys only.  
2. Apparently, there are 4 variations of the NiceGUI happy face SVG as used in the NiceGUI documentation website, as we applied `.classes()` shortly afterwards. 
  a. Right now you see me use the automatic `.cache()`, since it would blow up if you do `.cache('name')`, or the elements would be different but share the same cache name. Issue is that, since not all pages have all 4 variations, the variations which do not occur in every page, they come in and out of the cache (since right now we drop the element from the cache immediately when it is not used)
    - I am thinking if we should not drop automatic cache name elements, and instead we clear the oldest unused elements when they are soon over 5MB total
    - We could manually assign cache names on each element like `svg.face().classes(...).cache(...)`, but that is a tad bulky

@rodja Do you think we should re-think whether we do the `fetch_*` methods, otherwise we can't cache the sidebar hierarchy without much headache? Since we repeat content on a data-level, and element-level two `ui.tree`s expanded differently are technically two different elements. Thanks.

rodja:
I would rather not like to go back to tokens and the `fetch_*` methods. Let's explore other possibilities. What about allowing each element to overwrite `.cache` to define a custom cache strategy? `ui.tree` would produce just `nodes`,  the SVG underlying `ui.html` would use `innerHTML` and so forth.

evnchn:
A custom cache strategy per element is a very interesting idea indeed, especially when the old `fetch_*` methods suffer when the cached data is mutated before making its way to the JSON definition (e.g. `ui.markdown` as we have seen)

I'll take a look at that idea

Meanwhile, I have these thoughts:

- We should not address the pain point of transmitting the entire element definition in the Socket.IO communication (aka in the Outbox stage) in this PR, since I think if we should enhance that aspect, then it may apply to non-cached elements as well.

evnchn:
Initially, I'm quite perplexed by the idea of a "cache strategy" since that would imply executable code on the browser, so each element need to have its own JavaScript snippet, which hardly seems ideal. 

Now, I'm thinking perhaps the "strategy" is not executable code, but a list of props root keys for which (1) value composes of most of the element's size, (2) unlikely to be mutated, (3) when mutated, it could be argued that it's an entirely different element. In this case, for `ui.tree`, it'd be "nodes" in props. 

I think that this strategy makes more sense, but we need to make this for **every** element, which could be quite a burden and could take a while (or I can ask some LLMs to help once I've laid out the groundwork). Still it's better than the only-working-sometimes `fetch_*` I guess

evnchn:
@rodja Thank you for your pointer in the right direction! Now we are back to the 30KB -> 18KB for documentation (specifically, testing on /documentation/section_text_elements), while: 

- No token nonsense, since we act on root keys only (I tapped onto `replaceUndefinedAttributes`, after all)
- Able to overcome the different classes of the 4 variations of the NiceGUI happy face SVG, by excluding it from the cached keys with `happy_face_svg_element.dynamic_keys.add('class')`, meaning the classes will be dynamic across cache instances, and so we share 1 cache. 
- Able to cache `ui.tree` and `ui.html` post-change, with `tree.static_prop_keys.add('nodes')` and `html_element.static_prop_keys.add('innerHTML')`

It went much smoother than what I expected. Shall we review the API? I think manually calling `.add` on the sets breaks the builder pattern, and we should also review the unified `.cache()` API at https://github.com/zauberzeug/nicegui/pull/4796#issuecomment-2951928262

rodja:
Why are you using `static_prop_keys.add(..)` in the user code (namely the documentation website). Would it not be much better to do this per Element in their respective constructors? I think it's more of an implementation detail and should not be cared about by normal developers.

evnchn:
Just to be clear, it is not expected for userspace code to write to `static_prop_key` and `dynamic_keys`

Hopefully, each element should have its sensible default, and manual configuration should be rare and far between. 

However, the need to configure still exists (one shared element with shared styles, or the element may have differing styles, as we have seen with the happy face SVG), so there still needs to be some configurability, but not often. 

For that, I think we'd need to builder pattern to continue, so we may need some helper methods in the same calling manner as classes and props. 

So I'm waiting for confirmation before adding caching awareness to all elements, since the workload would be huge.

evnchn:
https://github.com/zauberzeug/nicegui/commit/6ecb1c5410746f869559b3b80730ffb24a1ffdbc shows caching applied to `ui.tree`, and `ui.html` (and other elements from `ContentElement`, though their docstring is not updated)

For caching of all elements to work*, I have to set `static_props_keys` for **all** of the elements.

So that's why I am hesitant on continuing, and I'm not sure if this PR is in progress if that's the case @falkoschindler 

*by work, I mean actually doing caching. An element which haven't set the `static_props_keys`, when cached, will cache not a lot of stuff. It won't grind the page to a halt, but it'd be ineffective.

evnchn:
All good suggestions above. I generally have no disagreements with it. 

For the implementation wise: 

- I'll see if I have time left on the table for doing it in the weekends? Weekday's busy...
- You can definitely edit my branch as maintainers. 
- To anyone reading, you can actually PR into my branch at https://github.com/evnchn/nicegui/tree/browser-data-store, and I'll review it ASAP (or just let it pass). This enables us to contribute to this PR. 

As we have seen, this simple concept has grown to something huge (has been a tendency so far but it's good I think). 

Let's let this sit in the oven for a bit longer 💪

evnchn:
Between internship, university, this, and #4900, I need quite a long period of time to sort this out. 

Also, given SPA in 2.22.1, NiceGUI is fast enough in most use cases already. 

Let's focus on other PRs in preparation for 3.0 release.

evnchn:
Too far gone! I am in https://github.com/zauberzeug/nicegui/pull/5493

PR #4799, opened 2025-05-27
Preload only documentation pages marked with `@doc.auto_execute`

evnchn:
### Motivation

First, in #4728, I noticed that documentation demo pages are only available after the corresponding sections have been visited once.

Then, in #4790, @rodja tried his hands on preloading the documentation pages. But he got bugs and tests only pass on 3.12+, which is suspicious. 

Past experience in https://github.com/zauberzeug/nicegui/pull/4671#issuecomment-2848552821 indicate that, if tests fail on old versions only, they may have something to do with asyncio, which there shouldn't if we are simply defining a page. 

I then watched a YouTube video about how a CVE boils down to "Python function decorators execute code as well" https://youtu.be/T2nBvNBzrP8

Then I asked LLM for "python function decorator for executing the function immediately after definition": https://poe.com/s/YoZKe2s8uRpY4uWZEpVo

Seeing light at the end of the tunnel, I proceed with the PR. 

### Implementation

The code boils down to this: 

```py
def auto_execute(func):
    """Decorator to automatically execute the function when the module is imported."""
    func()
    return func
```

This means, if we decorate the function with `@doc.auto_execute`, that function gets executed immediately. 

The idea is simple: Apply this decorator to **only** the demo functions which are simple `@ui.page` wrappers, and we'll go from there. 

### Progress

Not sure how to Pytest the documentation site, but I have tested every single page and it works. Additionally, the one demo about `@app.get` FastAPI endpoint is also covered.

### DIFF
--- a/website/documentation/content/doc/__init__.py
+++ b/website/documentation/content/doc/__init__.py
@@ -1,6 +1,7 @@
-from .api import demo, extra_column, get_page, intro, part, redirects, reference, registry, text, title, ui
+from .api import demo, extra_column, get_page, intro, part, redirects, reference, registry, text, title, ui, auto_execute
 
 __all__ = [
+    'auto_execute',
     'demo',
     'extra_column',
     'get_page',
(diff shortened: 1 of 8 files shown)


falkoschindler (approved):
Very nice! Such a clean solution... 🌟

I thought about decorating the page functions. But decorating the demo functions is the way to go.

PR #4801, opened 2025-05-27
Skip ES module shim for modern browsers entirely, ~10% faster page load (CPU-bound)

evnchn:
### Motivation

First, I discovered the `nomodule` attribute of `script` completely by accident in #4761. I pitched that it can speed up page load, but never got around to testing it. I was going on a trip back then. 

I promptly forgot about it. 😅

Until #4798 where we discovered the importance of applying a CPU throttle to make speed issues show themselves, and I was reminded of the topic, and proceeded to implement it and benchmark it. I found a significant speed increase. 

### Implementation

Simply set `nomodule` to the ES module shim. 

### Progress

### Results showcase: 

Low-tier mobile speed throttling (8.5x on my machine), no need disable cache and network throttling (we want to show **only** JS execution time)

Before: 

image

After: 

image

Finish: 7.71s -> 6.97s (Difference: -0.74s, Percentage Difference: -9.60%)
DOMContentLoaded: 6.48s -> 5.85s (Difference: -0.63s, Percentage Difference: -9.72%)
Load: 7.57s -> 6.82s (Difference: -0.75s, Percentage Difference: -9.91%)

### Potential impacts

We can no longer use the functionalities that ES Module Shims offer to us

- `importShim.getImportMap()`
- `importShim.addImportMap(importMap);`
- Shim import `importShim('/path/to/module.js').then(x => console.log(x));`
- Dynamically injecting import map like
```js
document.head.appendChild(Object.assign(document.createElement('script'), {
  type: 'importmap',
  innerHTML: JSON.stringify({ imports: { x: './y.js' } }),
}));
```

I doubt they are useful, since if your browser doesn't support ES Module, ES Module Shims does **not** magically make NiceGUI run with better compatibility. Notably, `ui.markdown` with Mermaid diagram always breaks.

### DIFF
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -21,7 +21,7 @@
     {% endfor %}
   </head>
   <body>
-    <script src="{{ prefix | safe }}/_nicegui/{{version}}/static/es-module-shims.js"></script>
+    <script nomodule src="{{ prefix | safe }}/_nicegui/{{version}}/static/es-module-shims.js"></script>
     <script defer src="{{ prefix | safe }}/_nicegui/{{version}}/static/socket.io.min.js"></script>
     {% if tailwind %}
     <script defer src="{{ prefix | safe }}/_nicegui/{{version}}/static/tailwindcss.min.js"></script>


falkoschindler (approved):
Looks like a good solution to reduce costs for ESM-capable browsers while keeping functionality for all others.
Maybe we can consider dropping support for non-ESM browsers in 3.0?

evnchn:
But, what do you mean by "support non-ESM browsers"? 

IMO: 

- We never did support non-ESM browsers, since as I have mentioned, part of NiceGUI already falls apart long before I began contributing. Check https://github.com/zauberzeug/nicegui/pull/4625#issuecomment-2812752236
- We included the ES module shim merely as a "best effort attempt" at making things somewhat compatible with non-ESM browsers, in a sense that the page won't be blank, but rather only some elements break. 

And so, since with this PR, ES module shim is loaded in only for non-ESM browsers and skipped entirely for ESM browsers, we are already at the best possible state for NiceGUI's ESM support: **Mandatory, and as best-effort we put in the ES module shim if you don't have ESM**

Please define clearly what we are doing to "support non-ESM browsers" and what we'd like to do to "drop support non-ESM browsers" if you would like to proceed.

evnchn:
Like basically for no reason whatsoever `import("mermaid")` and `import {default as mermaid} from "mermaid"` doesn't work on Pale Moon and Basilisk (two browsers without ES module support), despite we having an import map. 

I did explore the possibility of using "shim mode" for ES module shim such that `importShim("mermaid")` works, but we'd slow down all ESM browsers / add an extra try-except to all functions, which is a pain, and goes against the spirit of "not focusing on non-ESM browsers"

evnchn:
@falkoschindler 

> drop the shim only when core functionality began to break, even with ES module shim on non-ESM browser

You know what? I was investigating whether I can minify ES module shims to save 5KB, so I installed Basilisk today on my Mac, and I found that #5351 totally broke NiceGUI (nothing works) for browsers which need the shim to "use" ES modules. 

Even if we go before #5351, it looks like this, because CSS layer support is likely not there, if your browser is old enough to not support ES modules. 

image

#5351 shipped 3.3.0, and now is 3.3.1, with nobody calling in to complain. 

Maybe we should just drop the whole thing altogether.

falkoschindler:
Very interesting, @evnchn! Maybe 3.3 hasn't been out there long enough. But if there are no complains until - say - next year, we might as well remove it.

falkoschindler:
Ok, given that 3.0 already breaks styling for such browsers, we might already have reached the point of no return. I'll look into PR #5514. 👍🏻

PR #4807, opened 2025-05-28
Correction of the link to JavaScript's History API in the documentation

Igor-san:
### Progress


evnchn:
You should work off a branch instead of off your own fork's main, and keep the main up to date, to avoid last time's commit being in this PR. 

Other than that, LGTM

evnchn:
Wait. This is from Copilot:

> markdown link should not have < and > right? 

You are correct. In standard Markdown, the syntax for a link is:

```markdown
[link text](URL)
```

There should not be angle brackets `<` and `>` around the URL inside the parentheses.

So this is correct:

```markdown
[JavaScript's History API](https://developer.mozilla.org/en-US/docs/Web/API/History)
```

But this (with `<` and `>`) is not standard Markdown and may not render properly in all Markdown parsers:

```markdown
[JavaScript's History API](<https://developer.mozilla.org/en-US/docs/Web/API/History>)
```

**In the selection you highlighted:**
```diff
-    See `JavaScript's History API <https://developer.mozilla.org/en-US/docs/Web/API/History>`_ for more information.
+    See [JavaScript's History API](<https://developer.mozilla.org/en-US/docs/Web/API/History>) for more information.
```
The new version uses `[JavaScript's History API](<https://developer.mozilla.org/en-US/docs/Web/API/History>)`, which includes the angle brackets. This is not standard Markdown and should be changed to:

```markdown
[JavaScript's History API](https://developer.mozilla.org/en-US/docs/Web/API/History)
```

**Summary:**  
You are right—remove the `<` and `>` for best Markdown compatibility.

falkoschindler (changes requested):
I agree with @evnchn:

1. The change in element.py is confusing, but probably doesn't hurt in this case.
2. The syntax `[...](<...>)` is incorrect. Please remove `<` and `>`.

Igor-san:
I apologize, I don’t know my way around the git (and Englsih too :( ) well, I’ll improve.

PR #4811, opened 2025-05-30
Fixes #4805 Allows PyWebView's `storage_path` and `private_mode=False` to work aka private mode off

### LINKED ISSUE #4805: Can't enable persistant objects on native mode due to repeated argument error
### First Check

### Example Code

```python
from nicegui import ui, app

# arguments for pywebview outside of main guard as its a different process
app.native.start_args['private_mode'] = False
app.native.start_args['storage_path'] = r'%APPDATA%\myapp-cookies'
 
if __name__ == '__main__':
    ui.run(native=True)
```

### Description

Trying to enable object persistence, aka cookies, on the native PyWebview does not work with `private_mode` off and `storage_path` set. 

NiceGUI code kills any attempts at persistence by overwriting the storage path with a random temporary file every time; therefore, `storage_path` is overdefined when passed in, and `private_mode` becomes useless because of the randomization. 

```
  File "C:\Users\x\AppData\Local\Programs\Python\Python312\Lib\multiprocessing\process.py", line 314, in _bootstrap
    self.run()
  File "C:\Users\x\AppData\Local\Programs\Python\Python312\Lib\multiprocessing\process.py", line 108, in run
  File "C:\Users\x\Documents\Projects\touch-screen\env\Lib\site-packages\nicegui\native\native_mode.py", line 50, in _open_window
    webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
TypeError: webview.start() got multiple values for keyword argument 'storage_path'
```

I have referenced the NiceGUI documentation, code, and PyWebView documentation 
https://pywebview.flowrl.com/api/#webview-start and the problem comes from here:

https://github.com/zauberzeug/nicegui/blob/134c3308bcd0c8dd7d1d5e932b637f32f3dcbc12/nicegui/native/native_mode.py#L50

Looking at the PyWebView  documentation, we can be sure that two things are provided for us:
1. When `storage_path` is set to `None` it defaults a OS dependant default value for example Windows its `%APPDATA%\pywebview`
2. When `private_mode` is set to True (default True), then there is no persistence. 
   - therefore it's safe to leave `storage_path` empty

The solution is to change:

```python
    webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
```
to:

```python
    webview.start(**core.app.native.start_args)
```

we can then use the following to specify what we want without any conflicts:
```python
# main
app.native.start_args['private_mode'] = False
app.native.start_args['storage_path'] = r'%APPDATA%\myapp-cookies'
```

### NiceGUI Version

2.11.1

### Python Version

3.12.3

### Browser

Other

### Operating System

Windows

### Additional Context

_No response_

nachobacanful:
Allows `storage_path` to be overwritten without incurring in TypeError repeated arguments. It also allows `private_mode=False` to work as expected (only if `storage_path` is passed in). 

### Motivation
Fixes issues detailed in #4805. 
Resolves issues regarding multiple values if the user passed in:
```python
app.native.start_args['storage_path'] = r'%APPDATA%\myapp-cookies'
```
which resulted in:
```
 File "C:\Users\x\Documents\Projects\touch-screen\env\Lib\site-packages\nicegui\native\native_mode.py", line 50, in _open_window
    webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
TypeError: webview.start() got multiple values for keyword argument 'storage_path'
```
Additionally, the original code:
```python
webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
```
would prevent `private_mode=False`  in PyWebView from working as intended. This patch addresses that. 

### Implementation
As per issue #4805 discussion, the fix is done by using dictionary unpacking to allow NiceGui defaults `storage_path=tempfile.mkdtemp()` to be overwritten to the user's liking without incurring multiple `storage_path` 
 arguments. 

By fixing/allowing `storage_path` to be user-settable, it restores the ability to use PyWebView's `private_mode=False` functionality. 

### Progress

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -47,7 +47,9 @@ def _open_window(
     closed = Event()
     window.events.closed += closed.set
     _start_window_method_executor(window, method_queue, response_queue, closed)
-    webview.start(storage_path=tempfile.mkdtemp(), **core.app.native.start_args)
+    if core.app.native.start_args.get("private_mode") == False and "storage_path" not in core.app.native.start_args:
+        log.warning('Pass in your own `storage_path` to enable proper `private_mode=False` behaviour')
+    webview.start(**{'storage_path': tempfile.mkdtemp(), **core.app.native.start_args})
 
 
 def _start_window_method_executor(window: webview.Window,


evnchn:
First off, we use single-quotes. **It would fail the code quality check as it stands.** 

Second, I was thinking about `core.app.native.start_args.get("private_mode") == False`: 

I was thinking about the user passing _falsy_ values (`core.app.native.start_args['private_mode'] = 0` or some weird thing), and considering something like `not core.app.native.start_args.get("private_mode", True)` so as to work around that case. 

But I m not sure I am overly worryful...

falkoschindler (approved):
Thanks for this pull request, @nachobacanful!

Ruff doesn't like comparisons `== False` and recommends `is False` instead. But I took the opportunity and followed @evnchn's suggestion using `if not ... get('private_mode', True)` to cover _falsy_ values as well.

Ready to merge! 🚀

evnchn:
@falkoschindler Check if you pushed your latest `if not ... get('private_mode', True)` change. Can't see it here...

falkoschindler:
@evnchn Ooops, there it is.

PR #4329, opened 2025-02-10
Switched to python 3.9

Solipsistmonkey:
I changed some instances where it had python 3.8 as Poetry 2.0.0 dropped support for 3.8. I noticed this when trying to run "docker.sh buildup app"

PR #4812, opened 2025-05-30
Fix the development container to keep working with Python 3.8

ProfessorTom:
### Motivation

This PR solves the problem of the development container not building. Specifically, `poetry` was failing during the install.

I posted about it [here](https://discord.com/channels/1089836369431498784/1377358120279609405) in the NiceGUI Discord. (If necessary, I can edit this PR and/or add screenshots of the old behavior.)

### Implementation

It's literally a two character change `3.8` to `3.11`

### Progress

### DIFF
--- a/development.dockerfile
+++ b/development.dockerfile
@@ -1,12 +1,21 @@
-FROM python:3.11-slim
+FROM python:3.8-slim
 
 RUN apt update && apt install curl build-essential -y
 
+# The next few lines are the original command commented out
+# but kept in the file to quickly swap back once development.dockerfile
+# is upgraded to use python 3.11.
+
+
 # We use Poetry for dependency management
-RUN curl -sSL https://install.python-poetry.org | python3 - && \
-    cd /usr/local/bin && \
-    ln -s ~/.local/bin/poetry && \
-    poetry config virtualenvs.create false
+#RUN curl -sSL https://install.python-poetry.org | python3 - && \
+#    cd /usr/local/bin && \
+#    ln -s ~/.local/bin/poetry && \
+#    poetry config virtualenvs.create fals
+
+# pin the last seed packages that still support 3.8
+RUN python -m pip install --upgrade "pip<25" "setuptools<80" wheel
+RUN python -m pip install "virtualenv<20.31" "poetry==1.8.5" && poetry config virtualenvs.create false
 
 WORKDIR /app
 
--- a/development.dockerfile
+++ b/development.dockerfile
@@ -2,20 +2,10 @@ FROM python:3.8-slim
 
 RUN apt update && apt install curl build-essential -y
 
-# The next few lines are the original command commented out
-# but kept in the file to quickly swap back once development.dockerfile
-# is upgraded to use python 3.11.
-
-
-# We use Poetry for dependency management
-#RUN curl -sSL https://install.python-poetry.org | python3 - && \
-#    cd /usr/local/bin && \
-#    ln -s ~/.local/bin/poetry && \
-#    poetry config virtualenvs.create fals
-
-# pin the last seed packages that still support 3.8
-RUN python -m pip install --upgrade "pip<25" "setuptools<80" wheel
-RUN python -m pip install "virtualenv<20.31" "poetry~=1.8" && poetry config virtualenvs.create false
+# keep toolchain compatible with Python 3.8
+RUN python -m pip install --no-cache-dir --upgrade "virtualenv<20.31" "pip<25" "setuptools<80" wheel \
+    && python -m pip install --no-cache-dir "poetry~=1.8" \
+    && poetry config virtualenvs.create false
 
 WORKDIR /app
 
--- a/development.dockerfile
+++ b/development.dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8-slim
+FROM python:3.11-slim
 
 RUN apt update && apt install curl build-essential -y
 
--- a/development.dockerfile
+++ b/development.dockerfile
@@ -15,7 +15,7 @@ RUN apt update && apt install curl build-essential -y
 
 # pin the last seed packages that still support 3.8
 RUN python -m pip install --upgrade "pip<25" "setuptools<80" wheel
-RUN python -m pip install "virtualenv<20.31" "poetry==1.8.5" && poetry config virtualenvs.create false
+RUN python -m pip install "virtualenv<20.31" "poetry~=1.8" && poetry config virtualenvs.create false
 
 WORKDIR /app
 


evnchn:
Hello @ProfessorTom. It's good to see you here on GitHub. 

Regarding the PR, I understand that you are coming from the perspective of getting the development container working, but I think that we need to approach this very carefully, since Python 3.8 is still actively used in a couple of places. 

Code search: https://github.com/search?q=repo%3Azauberzeug%2Fnicegui%203.8&type=code

1. We still using Python 3.8 for the Devcontainer

https://github.com/zauberzeug/nicegui/blob/26765aaa7ec49c6e92b3e628fbd3c5a191fa1462/.devcontainer/Dockerfile#L1

(Also mentioned in https://github.com/zauberzeug/nicegui/blob/26765aaa7ec49c6e92b3e628fbd3c5a191fa1462/CONTRIBUTING.md)

2. We actively use Python 3.8 for pre-commit, which from my understanding, **would break after your change**, since there is no Python 3.8 to be found. 

https://github.com/zauberzeug/nicegui/blob/26765aaa7ec49c6e92b3e628fbd3c5a191fa1462/.pre-commit-config.yaml#L2

How NiceGUI depends on Python 3.8 for pre-commit is apparent by the fact that, when I develop locally without Docker on Windows, I have to manually install Python 3.8 on my machine, to get things to work. 

3. We actively use Python 3.8 for GitHub actions for testing. From my understanding, it'd still work after your change, but we will no longer check against Python 3.8 compatibility when we develop locally, straining the test servers. 

https://github.com/zauberzeug/nicegui/blob/26765aaa7ec49c6e92b3e628fbd3c5a191fa1462/.github/workflows/test.yml#L19

@falkoschindler I am not very proficient in NiceGUI's containers, testing systems, and pre-commit. I am not sure what I have said is 100% accurate, but it still goes to show that we do **not** take lightly a two-character change, or any change for that matter. 

Regardless, I still hope that the container builds successfully. I'm sure we can figure the details out with time. Cheers!

falkoschindler:
Hi @ProfessorTom,

Thanks a lot for the PR.

Why we keep the dev-container on Python 3.8:
- 3.8 is still the lowest version NiceGUI officially supports.
- Keeping the development image on the oldest supported runtime lets us catch compatibility problems early, long before they reach users running legacy systems.
- We'll only raise this baseline when we formally drop 3.8 support (see discussion #4135). For now, 3.8 remains the reference point. 

Because of that policy we can't merge a switch to 3.11, even though the production image happens to use it.

I asked ChatGPT about the original problem and this is what it came up with:

> The build failure is caused upstream releases, not by the container itself:
> 
> * virtualenv 20.31 ➜ now bundles pip 25 / setuptools 80
> * pip 25 + setuptools 80 ➜ officially dropped Python 3.8 support  
>   (see https://github.com/pypa/pip/issues/12989, https://github.com/pypa/setuptools/pull/4718)
> 
> Because the dev container intentionally sticks to **Python 3.8** to expose compat-bugs early, the Poetry installer aborts as soon as those wheels are seeded.
> 
> **Fix (without raising the baseline):**
> 
> ```Dockerfile
> # pin the last seed packages that still support 3.8
> RUN python -m pip install --upgrade "pip<25" "setuptools<80" wheel
> RUN python -m pip install "poetry==2.1.3" && poetry config virtualenvs.create false
> ```
> 
> (or pin virtualenv<20.31, whichever you prefer).
> 
> Once we drop 3.8 we can delete the pin again.

Maybe you can try this fix and adjust the PR accordingly? Thanks!

ProfessorTom:
Attempting to solve this problem, I tried the two lines you mentioned from ChatGPT. But poetry 2.1.3 isn't a version that is compatible with python 3.8:

```[4/9] RUN python -m pip install "poetry==2.1.3" && poetry config virtualenvs.create false:
0.459 ERROR: Ignored the following versions that require a different python version: 2.0.0 Requires-Python <4.0,>=3.9; 2.0.1 Requires-Python <4.0,>=3.9; 2.1.0 Requires-Python <4.0,>=3.9; 2.1.1 Requires-Python <4.0,>=3.9; 2.1.2 Requires-Python <4.0,>=3.9; 2.1.3 Requires-Python <4.0,>=3.9
0.459 ERROR: Could not find a version that satisfies the requirement poetry==2.1.3 (from versions: 0.1.0, 0.2.0, 0.3.0, 0.4.0, 0.4.0.post1, 0.4.1, 0.4.2, 0.5.0b1, 0.5.0b2, 0.5.0, 0.6.0, 0.6.1, 0.6.2, 0.6.3b1, 0.6.3b2, 0.6.3b3, 0.6.3b4, 0.6.3b5, 0.6.3b6, 0.6.3b7, 0.6.3, 0.6.4b1, 0.6.4, 0.6.5, 0.7.0b1, 0.7.0b2, 0.7.0b3, 0.7.0b4, 0.7.0, 0.7.1, 0.8.0a0, 0.8.0a1, 0.8.0a2, 0.8.0a3, 0.8.0a4, 0.8.0, 0.8.1a0, 0.8.1, 0.8.2, 0.8.3, 0.8.4, 0.8.5a0, 0.8.5, 0.8.6, 0.9.0a0, 0.9.0a1, 0.9.0a2, 0.9.0a3, 0.9.0, 0.9.1, 0.10.0a0, 0.10.0a1, 0.10.0a2, 0.10.0a3, 0.10.0, 0.10.1, 0.10.2, 0.10.3, 0.11.0a0, 0.11.0a1, 0.11.0a2, 0.11.0a3, 0.11.0a4, 0.11.0, 0.11.1, 0.11.2, 0.11.3, 0.11.4, 0.11.5, 0.12.0a0, 0.12.0a1, 0.12.0a2, 0.12.0a3, 0.12.0a4, 0.12.0a5, 0.12.0, 0.12.1, 0.12.2, 0.12.3, 0.12.4, 0.12.5, 0.12.6, 0.12.7, 0.12.8, 0.12.9, 0.12.10, 0.12.11, 0.12.12, 0.12.13, 0.12.14, 0.12.15, 0.12.16, 0.12.17, 1.0.0a0, 1.0.0a1, 1.0.0a2, 1.0.0a3, 1.0.0a4, 1.0.0a5, 1.0.0b1, 1.0.0b2, 1.0.0b3, 1.0.0b4, 1.0.0b5, 1.0.0b6, 1.0.0b7, 1.0.0b8, 1.0.0b9, 1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.4, 1.0.5, 1.0.6, 1.0.7, 1.0.8, 1.0.9, 1.0.10, 1.1.0a1, 1.1.0a2, 1.1.0a3, 1.1.0b1, 1.1.0b2, 1.1.0b3, 1.1.0b4, 1.1.0rc1, 1.1.0, 1.1.1, 1.1.2, 1.1.3, 1.1.4, 1.1.5, 1.1.6, 1.1.7, 1.1.8, 1.1.9, 1.1.10, 1.1.11, 1.1.12, 1.1.13, 1.1.14, 1.1.15, 1.2.0a1, 1.2.0a2, 1.2.0b1, 1.2.0b2, 1.2.0b3, 1.2.0rc1, 1.2.0rc2, 1.2.0, 1.2.1, 1.2.2, 1.3.0, 1.3.1, 1.3.2, 1.4.0, 1.4.1, 1.4.2, 1.5.0, 1.5.1, 1.6.0, 1.6.1, 1.7.0, 1.7.1, 1.8.0, 1.8.1, 1.8.2, 1.8.3, 1.8.4, 1.8.5)
0.461
0.461 [notice] A new release of pip is available: 24.3.1 -> 25.0.1
0.461 [notice] To update, run: pip install --upgrade pip
0.461 ERROR: No matching distribution found for poetry==2.1.3
```

I tried asking grok how to pin `virtualenv<20.31` and was told I could add that package with version number in double quotes on the `poetry` line but in front of `poetry`. (I assume that `virtualenv` news to be installed before `poetry`.

Here is my final diff

image

I made a comment about version pinning and commented the original code out since I assume we'll want to go back to that when we upgrade to 3.11 (or at least  3.9). Yes, I know, it's usually not a best practice to leave commented code in; my thinking here was we could do a quick revert without having to look at previous commits.

I'm holding off on committing and pushing in case you think we should do something different. Just because the container built, that doesn't mean that everything is going to work correctly. And I lack confidence because I'm changing the version of two dependencies: `poetry` and `virtualenv`. Perhaps neither of these changes are that major: one is a package manager and the other is a virtual environment.

Anyway, if you think this is a good fix, I'll push my changes and ask for PR approval.

Here's the `docker build` output:

image

falkoschindler:
@ProfessorTom Yes, please go ahead and update the PR.

Minor remark about the poetry version: We should use `poetry~=1.8` (equivalent to `>=1.8,<1.9`). This way we’ll continue to receive patch releases while staying on the 1.x line.

For easier reference, here is the command to reproduce the problem:
```bash
docker build -f development.dockerfile .
```

ProfessorTom:
Rebased my branch and pushed the changes that were in the screenshot of the diff.

Based on our conversation, I think this code should be ready to merge.

ProfessorTom:
Yeah, I can go either way with leaving the comment code in or out.

My reason for leaving it in was that it would reduce mental capacity needed to revert it when the time came. And, you wouldn't have to dig through revision history.AND… if you didn't know the history of that file, you could still do the work. Otherwise, it may never get changed. (although, pinning versions of dependencies might help surface that a change needs to be made.)

falkoschindler:
@ProfessorTom I just started working on dropping Python 3.8 support in PR #4329. Do I remember correctly that we can revert development.dockerfile back to
```dockerfile
RUN curl -sSL https://install.python-poetry.org | python3 - && \
    cd /usr/local/bin && \
    ln -s ~/.local/bin/poetry && \
    poetry config virtualenvs.create false
```
Or should we stick with pip like this:
```dockerfile
RUN python -m pip install --no-cache-dir "poetry~=1.8" \
    && poetry config virtualenvs.create false
```

**Update:** I think I'll stick with pip. It seems much simpler and cleaner.

PR #4813, opened 2025-05-31
Feat: `ui.fab` and `ui.fab_action` element

evnchn:
### Motivation

Addresses https://github.com/orgs/zauberzeug/projects/6?pane=issue&itemId=82939936, how NiceGUI does not have native `ui.fab` and `ui.fab_action` element, having to use raw `ui.element`

### Implementation

- Blend code between `ui.menu`, `ui.button`, since:
  - Much like `ui.menu` nested with `ui.menu_item`, the Quasar FAB is `ui.fab` nested with `ui.fab_action` 
  - Much like `ui.button` (it is a button after all 😅)
- Steal the documentation implementation from `ui.menu`
- Move the old example to the FAB documentation, so as to get people to switch. 

-> If there were Turnitin for Git, this would score like a 95% similarity score. It's a good thing though, IMO. 

### Progress

**My first NiceGUI element. Review with care!**


evnchn:
🚀

PR #4816, opened 2025-05-31
Feat: `ui.page_scroller`

evnchn:
### Motivation

Addresses https://github.com/orgs/zauberzeug/projects/6/views/1?pane=issue&itemId=82939968

### Implementation

- As mentioned, [QPageScroller is very similar to QPageSticky](https://quasar.dev/layout/page-scroller/#:~:text=QPageScroller%20is%20very%20similar%20to%20QPageSticky), so we copy over the code and the function signature from `ui.page_sticky`
- Then, we add on the 3 new properties, `scroll-offset`, `reverse` and `duration`, inheriting the default values of 300ms duration and 1000ms scroll-offset from Quasar. 
- Reworked the documentation, because the old one was so messy and hard to read. 

### Progress

-> No pytest for `ui.page_sticky` so no pytest for `ui.page_scroller` too.

### DIFF
--- a/nicegui/elements/page_scroller.py
+++ b/nicegui/elements/page_scroller.py
@@ -10,8 +10,8 @@ class PageScroller(Element):
                  y_offset: float = 0,
                  *,
                  expand: bool = False,
-                 scroll_offset: int = 1000,
-                 duration: int = 0.3,
+                 scroll_offset: float = 1000,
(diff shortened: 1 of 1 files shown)


evnchn:
Did you ask Copilot to help you resolve merge conflict? Do try it. It's not 100% but it's decent.

PR #1970, opened 2023-11-06
Raise RuntimeError if a page layout element is nested

falkoschindler:
This PR tries to avoid the confusion from #1953 where a `ui.header` was placed inside a refreshable UI container.

**But** I'm not 100% sure if it is too harsh to exit with an exception just because the element is created from the wrong spot. This might break existing code where it works perfectly fine to call `ui.header` from anywhere. Should we print a warning instead? Or is it a reasonable requirement to create such elements from the top-most slot?

PR #4821, opened 2025-06-01
Introduce ui.sub_pages element to allow implementing single page applications (SPAs)

rodja:
### Motivation

We already have [an example on how to hand-build a Singe Page Application (SPA)](https://github.com/zauberzeug/nicegui/tree/main/examples/single_page_app). Then, @Alyxion worked out the feature-rich but super large pull request #2811 to integrate SPAs directly into NiceGUI. While working/reviewing the code I came up with this alternative pull request which allows the following SPA syntax:

```py
@ui.page('/')
@ui.page('/{_:path}') # tells FastAPI/Starlette that all path's should land on our SPA
def index():
    ui.label('some text before main content')
    ui.sub_pages({'/': child, '/other': child2})

def child():
    ui.link('goto other', '/other')

def child2():
    ui.link('goto main', '/')
```

Sub pages can also be nested. But only one `ui.sub_pages` element should be used on each hierarchy level.
If a sub page needs to modify content from it's parent, the object is simply passed via lambda:

```py
@ui.page('/')
@ui.page('/{_:path}')
def index():
    title = ui.label()
    ui.sub_pages({
        '/': lambda: child(title),
        '/other': lambda: child2(title)
    })

def child(title: ui.label):
    title.set_text('child title')
    ui.link('goto other', '/other')

def child2(title: ui.label):
    title.set_text('other title')
    ui.link('goto main', '/')
```

Paths to sub pages can contain parameters (not yet fully implemented):

```py
@ui.page('/')
@ui.page('/{_:path}')
def index(_):
    ui.select(label='item', options=['a', 'b', 'c'], on_change=lambda e: ui.navigate.to(f'/{e.value}'))
    ui.sub_pages({'/{item}': main_content})

def main_content(item: str = 'nothing'):
    ui.label(f'item: {item}')
```

Key differences compared to #2811 (@Alyxion's SPA PR)

 - much less code (but obviously also only a fraction of the features)
 - self-explanatory shared state between pages via parameter passing (not `yield` and object storage)
 - just a new UI element `ui.sub_pages` instead of decorators, routers etc. (for custom routing with sub pages, users should derive from `ui.sub_pages` and override 'show()')

### Implementation

I tried to create a PR as small as possible. But adding nested sub pages, async support and path parameters made it a lot more complicated than I originally thought. It is also still far from done.

If you want to dive into this, I recommend looking at `test_sub_pages.py` for examples on how the `ui.sub_pages` api works.

### Progress

  - [ ] make `ui.navigate.to` work with sub pages
  - [ ] allow default parameters for sub page functions
  - [ ] ensure only one `ui.sub_pages` is used on each hierarchy level
- [ ] Documentation has been added.

### DIFF
--- a/nicegui/elements/sub_pages.js
+++ b/nicegui/elements/sub_pages.js
@@ -2,11 +2,13 @@ export default {
   template: "<slot></slot>",
   mounted() {
     const initial_path = window.location.pathname;
-    //this.$emit("open", initial_path);
-    // Listen for browser back/forward navigation
     window.addEventListener("popstate", (event) => {
       console.log("popstate", event);
-      this.$emit("route_change", event.state?.page || window.location.pathname);
+      this.$emit("open", event.state?.page || window.location.pathname);
+    });
+    window.addEventListener("pushstate", (event) => {
+      console.log("pushstate", event);
+      this.$emit("open", event.state?.page || window.location.pathname);
     });
 
     document.addEventListener("click", (e) => {
--- a/nicegui/elements/sub_pages.js
+++ b/nicegui/elements/sub_pages.js
@@ -1,5 +1,9 @@
 export default {
-  template: "<slot></slot>",
+  template: `
+    <div class="nicegui-column">
+      <slot></slot>
+    </div>
+  `,
   mounted() {
     const initial_path = window.location.pathname;
     window.addEventListener("popstate", (event) => {
--- a/nicegui/elements/sub_pages.js
+++ b/nicegui/elements/sub_pages.js
@@ -1,0 +1,22 @@
+export default {
+  template: "<slot></slot>",
+  mounted() {
+    const initial_path = window.location.pathname;
+    //this.$emit("open", initial_path);
+    // Listen for browser back/forward navigation
+    window.addEventListener("popstate", (event) => {
+      console.log("popstate", event);
+      this.$emit("route_change", event.state?.page || window.location.pathname);
+    });
+
+    document.addEventListener("click", (e) => {
+      const a = e.target.closest("a[href]");
+      e.preventDefault();
+      if (!a) return;
+      const path = a.getAttribute("href");
+      history.pushState({}, "", path);
+      this.$emit("open", path);
+      console.log("click", path);
+    });
+  },
+};
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -1,0 +1,31 @@
+from asyncio import iscoroutine
+from typing import Callable, Dict, Self
+
+from .. import background_tasks, ui
+from ..element import Element
+
+
+class SubPages(Element, component='sub_pages.js'):
+
+    def __init__(self, routes: Dict[str, Callable]):
+        super().__init__()
+        self._routes = routes
+        self.show(ui.context.client.request.url.path if ui.context.client.request else '/')
+        self.on('open', lambda e: self.show(e.args))
+
+    def add(self, path: str, page: Callable) -> Self:
+        """Add a new route to the sub pages."""
+        self._routes[path] = page
+        return self
+
+    def show(self, path: str):
+        """Show the page for the given path."""
+        if path in self._routes:
+            self.clear()
+            with self:
+                result = self._routes[path]()
+            if iscoroutine(result):
+                async def background_task():
+                    with self:
+                        await result
+                background_tasks.create(background_task(), name=f'sub_pages {path}')
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -1,5 +1,7 @@
+import inspect
+import re
 from asyncio import iscoroutine
-from typing import Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional
 
 from typing_extensions import Self
 
@@ -36,8 +38,9 @@ class SubPages(Element, component='sub_pages.js'):
             if not sub_path.startswith('/'):
                 sub_path = '/' + sub_path
             for path, builder in self._routes.items():
-                if sub_path == path:
-                    self._replace_content(path, builder)
+                params = self._match_path(path, sub_path)
+                if params is not None:
+                    self._replace_content(path, builder, params)
                     child_sub_pages = find_child_sub_pages(self)
                     if child_sub_pages:
                         child_sub_pages.show(full_path.removeprefix(path))
@@ -47,14 +50,36 @@ class SubPages(Element, component='sub_pages.js'):
             ui.label(f'404: sub page "{full_path}" not found')
         return False
 
-    def _find_base_path(self):
-        parent = self
-        while True:
-            if parent.parent_slot is None:
-                return '/'
-            parent = parent.parent_slot.parent
-            if isinstance(parent, SubPages):
-                return parent._current_path
+    @staticmethod
+    def _match_path(pattern: str, path: str) -> Optional[Dict[str, str]]:
+        """Match a path pattern against an actual path and extract parameters.
+
+        :return: empty dict for exact matches, parameter dict for parameterized matches, None for no match.
+        """
+        if '{' not in pattern:
+            return {} if pattern == path else None
+        regex_pattern = pattern
+        for match in re.finditer(r'\{(\w+)\}', pattern):
+            param_name = match.group(1)
+            regex_pattern = regex_pattern.replace(f'{{{param_name}}}', f'(?P<{param_name}>[^/]+)')
+        regex_pattern = f'^{regex_pattern}$'
+        regex_match = re.match(regex_pattern, path)
+        if regex_match:
+            return regex_match.groupdict()
+        return None
+
+    def _convert_parameter(self, value: str, param_type: type) -> Any:
+        """Convert a string parameter to the specified type."""
+        if param_type is str:
+            return value
+        elif param_type is int:
+            return int(value)
+        elif param_type is float:
+            return float(value)
+        elif param_type is bool:
+            return value.lower() in ('true', '1', 'yes', 'on')
+        else:
+            return value
 
     def _is_root(self) -> bool:
         parent: ui.element = self
@@ -64,10 +89,23 @@ class SubPages(Element, component='sub_pages.js'):
                 return False
         return True
 
-    def _replace_content(self, path: str, builder: Callable):
+    def _replace_content(self, path: str, builder: Callable, params: Optional[Dict[str, str]] = None):
         self.clear()
         with self:
-            result = builder()
+            if params:
+                sig = inspect.signature(builder)
+                converted_params = {}
+                for param_name, param_value in params.items():
+                    if param_name in sig.parameters:
+                        param_info = sig.parameters[param_name]
+                        param_type = param_info.annotation
+                        if param_type != inspect.Parameter.empty:
+                            converted_params[param_name] = self._convert_parameter(param_value, param_type)
+                        else:
+                            converted_params[param_name] = param_value
+                result = builder(**converted_params)
+            else:
+                result = builder()
         if iscoroutine(result):
             async def background_task():
                 with self:
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -1,5 +1,7 @@
 from asyncio import iscoroutine
-from typing import Callable, Dict, Self
+from typing import Callable, Dict
+
+from typing_extensions import Self
 
 from .. import background_tasks, ui
 from ..element import Element
@@ -10,6 +12,8 @@ class SubPages(Element, component='sub_pages.js'):
     def __init__(self, routes: Dict[str, Callable]):
         super().__init__()
         self._routes = routes
+        self._base_path = self._find_base_path()
+        self._current_path = '/'
         self.show(ui.context.client.request.url.path if ui.context.client.request else '/')
         self.on('open', lambda e: self.show(e.args))
 
@@ -18,13 +22,20 @@ class SubPages(Element, component='sub_pages.js'):
         self._routes[path] = page
         return self
 
-    def show(self, path: str):
+    def show(self, full_path: str):
         """Show the page for the given path."""
+        if full_path.startswith(self._base_path):
+            path = full_path[len(self._base_path):]
+            if not path.startswith('/'):
+                path = '/' + path
+        else:
+            path = full_path
         if path in self._routes:
+            self._current_path = path
             self.clear()
             ui.run_javascript(f'''
-                if (window.location.pathname !== "{path}") {{
-                    history.pushState({{page: "{path}"}}, "", "{path}");
+                if (window.location.pathname !== "{full_path}") {{
+                    history.pushState({{page: "{full_path}"}}, "", "{full_path}");
                 }}
             ''')
             with self:
@@ -33,4 +44,18 @@ class SubPages(Element, component='sub_pages.js'):
                 async def background_task():
                     with self:
                         await result
-                background_tasks.create(background_task(), name=f'sub_pages {path}')
+                background_tasks.create(background_task(), name=f'building sub_page {full_path}')
+        else:
+            if any(full_path.startswith(route) for route in self._routes):
+                return
+            with self:
+                ui.label(f'404: sub page "{path}" not found on {self._base_path}')
+
+    def _find_base_path(self):
+        parent = self
+        while True:
+            if parent.parent_slot is None:
+                return '/'
+            parent = parent.parent_slot.parent
+            if isinstance(parent, SubPages):
+                return parent._current_path
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -40,14 +40,26 @@ class SubPages(Element, component='sub_pages.js'):
             for path, builder in self._routes.items():
                 params = self._match_path(path, sub_path)
                 if params is not None:
+                    if sub_path == '/':
+                        remaining_path = full_path
+                    else:
+                        remaining_path = full_path[len(sub_path):]
+                    if not remaining_path:
+                        remaining_path = '/'
+                    if sub_path == '/' and remaining_path == full_path and sub_path != full_path:
+                        continue
                     self._replace_content(path, builder, params)
                     child_sub_pages = find_child_sub_pages(self)
-                    if child_sub_pages:
-                        child_sub_pages.show(full_path.removeprefix(path))
-                    return True
+
+                    if child_sub_pages and remaining_path != sub_path:
+                        if child_sub_pages.show(remaining_path):
+                            return True
+                    elif sub_path == full_path:
+                        return True
             segments.pop()
+        self.clear()
         with self:
-            ui.label(f'404: sub page "{full_path}" not found')
+            ui.label(f'404: sub page {full_path} not found')
         return False
 
     @staticmethod
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -1,5 +1,5 @@
 from asyncio import iscoroutine
-from typing import Callable, Dict
+from typing import Callable, Dict, Optional
 
 from typing_extensions import Self
 
@@ -12,44 +12,40 @@ class SubPages(Element, component='sub_pages.js'):
     def __init__(self, routes: Dict[str, Callable]):
         super().__init__()
         self._routes = routes
-        self._base_path = self._find_base_path()
-        self._current_path = '/'
-        self.show(ui.context.client.request.url.path if ui.context.client.request else '/')
-        self.on('open', lambda e: self.show(e.args))
+        if self._is_root():
+            path = ui.context.client.request.url.path if ui.context.client.request else '/'
+            if self.show(path):
+                ui.run_javascript(f'''
+                    if (window.location.pathname !== "{path}") {{
+                        history.pushState({{page: "{path}"}}, "", "{path}");
+                    }}
+                ''')
+
+            self.on('open', lambda e: self.show(e.args))
 
     def add(self, path: str, page: Callable) -> Self:
         """Add a new route to the sub pages."""
         self._routes[path] = page
         return self
 
-    def show(self, full_path: str):
+    def show(self, full_path: str) -> bool:
         """Show the page for the given path."""
-        if full_path.startswith(self._base_path):
-            path = full_path[len(self._base_path):]
-            if not path.startswith('/'):
-                path = '/' + path
-        else:
-            path = full_path
-        if path in self._routes:
-            self._current_path = path
-            self.clear()
-            ui.run_javascript(f'''
-                if (window.location.pathname !== "{full_path}") {{
-                    history.pushState({{page: "{full_path}"}}, "", "{full_path}");
-                }}
-            ''')
-            with self:
-                result = self._routes[path]()
-            if iscoroutine(result):
-                async def background_task():
-                    with self:
-                        await result
-                background_tasks.create(background_task(), name=f'building sub_page {full_path}')
-        else:
-            if any(full_path.startswith(route) for route in self._routes):
-                return
-            with self:
-                ui.label(f'404: sub page "{path}" not found on {self._base_path}')
+        segments = full_path.split('/')
+        while segments:
+            sub_path = '/'.join(segments)
+            if not sub_path.startswith('/'):
+                sub_path = '/' + sub_path
+            for path, builder in self._routes.items():
+                if sub_path == path:
+                    self._replace_content(path, builder)
+                    child_sub_pages = find_child_sub_pages(self)
+                    if child_sub_pages:
+                        child_sub_pages.show(full_path.removeprefix(path))
+                    return True
+            segments.pop()
+        with self:
+            ui.label(f'404: sub page "{full_path}" not found')
+        return False
 
     def _find_base_path(self):
         parent = self
@@ -59,3 +55,31 @@ class SubPages(Element, component='sub_pages.js'):
             parent = parent.parent_slot.parent
             if isinstance(parent, SubPages):
                 return parent._current_path
+
+    def _is_root(self) -> bool:
+        parent: ui.element = self
+        while parent.parent_slot is not None:
+            parent = parent.parent_slot.parent
+            if isinstance(parent, SubPages):
+                return False
+        return True
+
+    def _replace_content(self, path: str, builder: Callable):
+        self.clear()
+        with self:
+            result = builder()
+        if iscoroutine(result):
+            async def background_task():
+                with self:
+                    await result
+            background_tasks.create(background_task(), name=f'building sub_page {path}')
+
+
+def find_child_sub_pages(element: ui.element) -> Optional[SubPages]:
+    for child in element.default_slot.children:
+        if isinstance(child, SubPages):
+            return child
+        result = find_child_sub_pages(child)
+        if result is not None:
+            return result
+    return None
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -22,6 +22,11 @@ class SubPages(Element, component='sub_pages.js'):
         """Show the page for the given path."""
         if path in self._routes:
             self.clear()
+            ui.run_javascript(f'''
+                if (window.location.pathname !== "{path}") {{
+                    history.pushState({{page: "{path}"}}, "", "{path}");
+                }}
+            ''')
             with self:
                 result = self._routes[path]()
             if iscoroutine(result):
--- a/tests/test_sub_pages.py
+++ b/tests/test_sub_pages.py
@@ -1,8 +1,10 @@
 from nicegui import ui
 from nicegui.testing import Screen
 
+# pylint: disable=missing-function-docstring
 
-def test_switching_between_sub_pages_keeps_other_content(screen: Screen):
+
+def test_switching_between_sub_pages(screen: Screen):
     index_calls = 0
 
     @ui.page('/')
@@ -29,9 +31,19 @@ def test_switching_between_sub_pages_keeps_other_content(screen: Screen):
     screen.should_contain('some text before main content')
     screen.should_contain('goto main')
     screen.should_not_contain('goto other')
+    screen.selenium.back()
+    assert index_calls == 1
+    screen.should_contain('some text before main content')
+    screen.should_contain('goto other')
+    screen.should_not_contain('goto main')
+    screen.selenium.forward()
+    assert index_calls == 1
+    screen.should_contain('some text before main content')
+    screen.should_contain('goto main')
+    screen.should_not_contain('goto other')
 
 
-def test_opening_sub_page(screen: Screen):
+def test_accessing_sub_page_directly(screen: Screen):
     @ui.page('/two')
     @ui.page('/one')
     def index():
--- a/tests/test_sub_pages.py
+++ b/tests/test_sub_pages.py
@@ -112,12 +112,15 @@ def test_sub_page_in_sub_page(screen: Screen):
 
 
 def test_parameterized_sub_pages(screen: Screen):
+    main_content_calls = 0
+
     @ui.page('/')
     @ui.page('/{_:path}')
     def index(_):
         ui.link('goto main', '/')
         ui.link('goto one', '/one-1')
         ui.link('goto two', '/two/two')
+        ui.link('goto wrong match', '/two-3')
         ui.sub_pages({
             '/': main,
             '/one-{idx}': one,
@@ -125,7 +128,9 @@ def test_parameterized_sub_pages(screen: Screen):
         })
 
     def main():
-        ui.label('main')
+        nonlocal main_content_calls
+        main_content_calls += 1
+        ui.label('main_content')
 
     def one(idx: int):
         ui.label(f'one-{idx}-{isinstance(idx, int)}')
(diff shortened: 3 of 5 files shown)


rodja:
@Alyxion and @evnchn what do you think about this approach? I already talked with @falkoschindler last week, but I would much appreciate your feedback. Where do you see problems or inherent limitations? Especially compared to #2811.

evnchn:
@rodja Thanks for inviting me on to this journey. 

Overall, this PR, #2811 (the original SPA), #4552 (Project ACID) and #4719 (Project BASE) should be viewed together, as at the end of the day, they achieve building a single-page application, hereby defined as a page which does NOT reload but loads-in new content nonetheless. `pushstate` and `popstate`, I consider it a cherry-on-top. 

### Parallels between Project ACID and this PR - re-implmementing the HTTP layer

Why I am saying this, is because I think @rodja you mentioned that you have to implement "async support and path parameters", which sounds very similar to back in #4552 (project ACID), where I re-implemented client initialization with async support (but left out path parameters since I didn't got around to do it) when the ACID maneuver occurs, and we'd like to spin up a client to match the browser. 

In the end, we came to the conclusion that we'd be re-inventing the HTTP layer and all the middleware / path parameters / pydantic classes that FastAPI offers, and we did not pursue. 

**So, high-level speaking, although this PR is obviously way less cursed than #4552 (project ACID), I fear that there will equally be a high maintenance burden in "async support and path parameters"** 

### Leveraging the existing HTTP layer with Project BASE

Then, if I am not mistaken, that in #4719, since we are making a HTTP request to the server for the latest content, that we do not need to re-implement "async support and path parameters", since we are leveraging the original functionality in `@ui.page`. 

**That PR has its own problems, but high-level speaking, the maintenance burden may be lower.** 

### Can we move forward with the Original SPA PR, with this new mindset in mind?

Seems like this PR was created to supersede the original SPA PR, but given the above concerns over re-implementing the HTTP layer and maintenance burden, it may or may not be a good idea. 

Though the original SPA PR is also very long and bulky, but at least we see, `nicegui/content.py` using `@ui.page` under the hood. This means that it could be using the existing page definition infrastructure, and **if we rework that PR to be simplier, there may be a way out as well.** 

@rodja I'd recommend to check out #4719 for a fresh perspective on how this SPA problem may be solved, now that a bug has been resolved in that PR, which made it work in the entire NiceGUI documentation (though tests are failing for some version for some reason)

nicegui/elements/sub_pages.js
```diff
@@ -0,0 +1,28 @@
+export default {
+  template: `
+    <div class="nicegui-column">
+      <slot></slot>
+    </div>
+  `,
+  mounted() {
+    const initial_path = window.location.pathname;
```
evnchn: I want to take this sub_pages element and fit it for #4719. I see that `initial_path` appears to be unused. Can you confirm? Thanks

evnchn (commented):
I want to take this sub_pages element and fit it for https://github.com/zauberzeug/nicegui/pull/4719. I see that `initial_path` appears to be unused. Can you confirm? Thanks

nicegui/elements/sub_pages.js
rodja: Thanks @evnchn. This was indeed a leftover 👍

Alyxion:
First of all thank you very much helping to drive the implementation forward @rodja.

As already discussed in the video call I think **yielding** data can easily be replaced by storing in within the client storage for the most part - if calls are deeply nested - or by passing it directly to the sub pages as in your example. As the client storage is not context aware this requires manual cleanup in some scenarios though where as the yield apporach did this automatically when a sub page was left which yielded objects.

**Async** is cruicial in my opinion, as NiceGUI is basically a single-threaded app no serious apps with mulitple users could be realized without page and the sub page being declarable as **async**.

The samples above didn't contain code behind the sub page declaration, but I assume thats as well possible as code before?

Regarding the general concept at least for us the actual approach with outlet and outlet.view allowed far cleaner structuring of larger apps in my opinion. It behaved like an ApiRouter in FastAPI or a Blueprint in Flask where you could define the outlet in a shared, central module file and then all "plugins" could import the outlet and register themselves to it... like in an APIRouter.

With this approach now thats not possibl anymore. At the point of implementation the main page would already need to know all children and/or do some workaround everytime, e.g. like a registry where all sub pages had to enlist themselves so that the main page could then register them. Don't like it to be honest. Preferred the "router style".

What I like though is that you used the same pattern for the declaration of sub pages as FastAPI, e.g. `{item}`. Using wildcards here to catch everything wasn't the right approach as it would not allow the declaration of endpoints starting with the same URL dynamically - such as when using an `ui.upload`.

rodja:
Response to @evnchn:

> Overall, this PR, https://github.com/zauberzeug/nicegui/pull/2811 (the original SPA), https://github.com/zauberzeug/nicegui/pull/4552 (Project ACID) and https://github.com/zauberzeug/nicegui/pull/4719 (Project BASE) should be viewed together

Exactly.

>  pushstate and popstate, I consider it a cherry-on-top.

No. In my opinion, the url-path-handling is crucial to SPAs. Especially in NiceGUI where we offer a "backend-first" approach and handle the web-development details internally.

> [in project [ACID](https://github.com/zauberzeug/nicegui/pull/4552)] we came to the conclusion that we'd be re-inventing the HTTP layer and all the middleware / path parameters / pydantic classes that FastAPI offers,

This PR's approach utilizes `ui.page` for http routing, middlewares, pydantic, etc. when accessing a page with sub pages through their full url. `ui.sub_pages` also does not reimplement routing as it is done in the original SPA PR #2811. Instead this PR just implements a minimalistic path-matching which is executed **after** FastAPI/Starlette does all it's magic. I'll update the description above.

> [in project [BASE](https://github.com/zauberzeug/nicegui/pull/4719)], the maintenance burden may be lower.

True. I really like the concept and want to see project BASE (#4719) merged. Its remarkable elegant, small and brings a great user improvement. But it is not solution for SPAs in my opinion. In an SPA you want to only load once (in this its similar to BASE), provide browser navigation (eg. pushstate/popstate) and most importantly only update the changed content via efficient backend communication.

> Though the original SPA PR #2811 is also very long and bulky, but at least we see, nicegui/content.py using @ui.page under the hood. This means that it could be using the existing page definition infrastructure, and if we rework that PR to be simplier, there may be a way out as well.

I tried. Which led me to this PR which avoids the reimplementation of full-fledged routing.

---

Response to @Alyxion:

> Async is cruicial in my opinion, as NiceGUI is basically a single-threaded app no serious apps with mulitple users could be realized without page and the sub page being declarable as async.

Yes, `ui.sub_pages` allows the builder function to be async. I'll update the description of this PR.

> Regarding the general concept at least for us the actual approach with outlet and outlet.view allowed far cleaner structuring of larger apps in my opinion. It behaved like an ApiRouter in FastAPI or a Blueprint in Flask where you could define the outlet in a shared, central module file and then all "plugins" could import the outlet and register themselves to it... like in an APIRouter.

True. Outlet, ApiRouter, Blueprint etc. are a lot more decentral by allowing self-registration. This "Inversion of Control" IoC is very modular and provides a loose coupling between main file and drop-in modules. But this startup-magic comes at the cost of hidden side-effects where an simple import mutates the global app state and it can be quite hard to understand the routing (because it is scattered through-out the project). `ui.sub_pages` on the other hand is implemented as a composition root where all routing is defined in one place (which also simplifies control over the registration order).

I'm leaning towards the central composition root architecture because it's simpler to use in small code bases. And -- as you said -- it can be extended to allow IoC by adding a registration mechanism for applications which require a more modular approach.

evnchn:
Interesting. I think we can define a Single-Page Appication as follows: 

1. Only one browser load
2. Capture browser's pushstate/popstate
3. Load in only the updated content

This definition shall be useful across all PRs attempting to take a shot at the problem. Let us bear this in mind. 

I have stuff to talk about how project BASE can achieve SPA, but I will put it at that PR's comment so avoid muddying waters. 

So the way I understand it, in this PR we're dragging on one `Client` across SPA navigations, and with it state held in memory. 

```py
@ui.page('/')
@ui.page('/{_:path}') # tells FastAPI/Starlette that all path's should land on our SPA
def index():
    ui.label('some text before main content')
    # some state-dependent element
    ui.sub_pages({'/': child, '/other': child2, '/mutate': child3})

def child():
    ui.link('goto other', '/other')
    ui.link('goto mutate', '/mutate')

def child2():
    ui.link('goto main', '/')

def child3():
    # some stuff which modifies the state...
    ui.link('goto main', '/')
```

If I go from `/` to `/mutate` and back to `/`, I can get a different page result, despite being on `/`. I imagine this can get messy real quick. Do you agree? Or am I overthinking?

rodja:
I do not quite understand your concern @evnchn. Can you provide an example with problematic code? Above you only placed comments which I do not know how to fill.

`ui.sub_pages` simply calls `.clear()` on the container before adding new content. In this regard it is very similar to `ui.refreshable`.

evnchn:
Depending on what you decide to do, the content in the main page can become wildly path-dependent. This could lead to confusion. 

```py
from nicegui import ui, app


@ui.page('/')
@ui.page('/{_:path}')
def index():
    app.storage.client['some_value'] = 0
    ui.label('some text before main content')
    ui.label().bind_text_from(app.storage.client, 'some_value')
    ui.sub_pages({'/': child, '/other': child2, '/mutate': child3})


def child():
    ui.link('goto other', '/other')
    ui.link('goto mutate', '/mutate')


def child2():
    ui.link('goto main', '/')


def child3():
    app.storage.client['some_value'] += 1
    ui.link('goto main', '/')


ui.run()
```

Initial load: 

{8B260310-A423-4180-A597-06A38C5D51B6}

Clicking around in the page: 

{AA37F2AD-1BAE-4412-9025-D6815FB08567}

rodja:
@falkoschindler you can start with a first review if you find the time. I'll look into creating the documentation.

rodja:
Oh, sorry @evnchn. I totally missed your comment

> Depending on what you decide to do, the content in the main page can become wildly path-dependent. This could lead to confusion.

I do not see the problem. If you have pages which change a storage value -- of course they will alter the internal state. That behaviour is not specific to sub pages:

```py
app.storage.general['some_value'] = 0

@ui.page('/')
def index():
    ui.label().bind_text_from(app.storage.general, 'some_value')
    ui.link('goto mutate', '/mutate')

@ui.page('/mutate')
def mutate():
    app.storage.general['some_value'] += 1
    ui.link('goto main', '/')
```

The main difference is that you can do such things with `app.storage.client` because it is a single page app (eg. single client app).

Alyxion:
Wow, a lot of progress over Pfingsten, awesome.

One remark regarding the new SPA feature b/c either I did it wrong or may be it's a real limitation of Quasar.

Elements like header, footer, drawer but also add_head_html are at the moment required to be in the "layout" page.

If you have a quite large application though then the chance is very high that not "one size fits all" - resulting in at the moment that we were forced to set up three root outlets.

To come to the point: Will it be possible to have headers / drawer / footer, potentially "head html" configured on "per sub-page" level or if not: how high is the hurdle?

evnchn:
Important message: As discovered in comments leading up to https://github.com/zauberzeug/nicegui/pull/4719#issuecomment-2957297018, NiceGUI currently leaks memory in the browser if we unmount _some_ elements (not all, since a simple page doesn't leak memory)

Without finding the root cause, this PR, among others which attempts SPA, will probably become slower and slower with more page navigations, with 20 navigations before considerable slowdown for Desktop, and I think at most 5 if we're lucky for mobile.

rodja:
> Will it be possible to have headers / drawer / footer, potentially "head html" configured on "per sub-page" level or if not: how high is the hurdle?

Thanks for this remark @Alyxion. I agree that it would simplify modularization if a sub page could also inject head-html or layout stuff. I'll check in the next few days.

rodja:
Somehow tests are now flaky. Also adding documentation uncovered multiple scenarios like navigating to pages which are not part of the sub pages. The implementation progresses but will still take some time.

rodja:
@Alyxion what do you think about this minimal example of a IoC `Router` which uses `ui.sub_pages` under the hood?

```py
from nicegui import ui

class Router:
    def __init__(self):
        self.routes = {}

    def add(self, path: str):
        def decorator(func):
            self.routes[path] = func
            return func
        return decorator

    def create_frame(self):
        ui.sub_pages(self.routes)

router = Router()

@ui.page('/')
@ui.page('/{_:path}')
def index():
    router.create_frame()

@router.add('/')
def main_content():
    ui.link('goto sub', '/sub')

@router.add('/sub')
def sub_content():
    ui.link('goto main', '/')

ui.run()
```

rodja:
> Elements like header, footer, drawer but also add_head_html are at the moment required to be in the "layout" page.

@Alyxion Yes. We show an error if page layout elements are used outside top-level (#1970). But if you want different drawers on each of your sub pages for example, you can remove the existing one with an `ElementFilter`:

```py
from nicegui import ElementFilter, ui

@ui.page('/')
@ui.page('/{_:path}')
def index():
    with ui.header().classes('text-white'):
        ui.link('goto main', '/').classes('text-white')
        ui.link('goto other', '/other').classes('text-white')
    ui.sub_pages({
        '/': main_content,
        '/other': other_content
    })

def main_content():
    ui.label('main content')
    clear_drawer()
    with ui.context.client:
        with ui.left_drawer().classes('w-20'):
            ui.label('main content drawer')

def other_content():
    ui.label('other content')
    clear_drawer()
    with ui.context.client:
        with ui.left_drawer().classes('w-20'):
            ui.label('other content drawer')

def clear_drawer():
    try:
        next(iter(ElementFilter(kind=ui.left_drawer))).delete()
    except StopIteration:
        pass

ui.run()
```

A similar approach would work for `ui.add_head_html`:

```py
...

def main_content():
    ui.label('main content')
    change_style('body { background-color: red; }')

def other_content():
    ui.label('other content')
    change_style('body { background-color: blue; }')

def change_style(style: str):
    ui.run_javascript('document.getElementById("my-style").remove()')
    ui.add_head_html(f'<style id="my-style">{style}</style>')
```

I would rather not generalize this for `ui.sub_pages` at the moment. The PR is complex enough as it is.

examples/single_page_app/custom_sub_pages.py
```diff
@@ -0,0 +1,70 @@
...
+        self.clear()
+        with self:
+            with ui.column().classes('items-center text-center p-8'):
+                ui.icon('error_outline', size='4rem').classes('text-red-500 mb-4')
+                ui.label('404 - Page Not Found').classes('text-3xl text-red-500 mb-2')
+                ui.label(f'The page "{path}" does not exist.').classes('text-gray-600 mb-4')
+                ui.separator().classes('w-32 mb-4')
+                with ui.row().classes('gap-4'):
+                    ui.button('Go Home', on_click=lambda: ui.navigate.to('/')).props('outline')
+                    ui.button('Go Back', on_click='history.back()').props('outline')
```
falkoschindler: TypeError: 'history.back()' is not a callable object

main.py
```diff
@@ -41,7 +41,7 @@ def _documentation_page() -> None:
     documentation.render_page(documentation.registry[''], with_menu=False)
 
 
-@ui.page('/documentation/{name}')
+@ui.page('/documentation/{name}/{_:path}')
```
falkoschindler, line 44: Why is this change necessary?

nicegui/functions/navigate.py
```diff
@@ -66,6 +67,11 @@ def to(self, target: Union[Callable[..., Any], str, Element], new_tab: bool = Fa
             path = Client.page_routes[target]
         else:
             raise TypeError(f'Invalid target type: {type(target)}')
+
+        if not new_tab and isinstance(target, str):
```
falkoschindler, line 70: What about callable targets? Would it possible/reasonable to use `navigate_to_sub_page` for them as well?

main.py
falkoschindler: It looks like this causes fonts not being found, e.g.:
> http://127.0.0.1:8080/documentation/fonts/fira-sans-v16-latin-700.woff2 not found

website/documentation/rendering.py
```diff
@@ -46,7 +46,7 @@ def render_content():
             if part.title:
                 if part.link_target:
                     ui.link_target(part.link_target)
-                subheading(part.title, link=part.link, major=part.reference is not None)
+                subheading(part.title, link=f'/documentation/{part.link}', major=part.reference is not None)
```
falkoschindler: Why is this change needed?

website/documentation/rendering.py
```diff
@@ -62,7 +62,7 @@ def render_content():
             if part.reference:
                 generate_class_doc(part.reference, part.title)
             if part.link:
-                ui.markdown(f'See [more...]({part.link})').classes('bold-links arrow-links')
+                ui.markdown(f'See [more...](/documentation/{part.link})').classes('bold-links arrow-links')
```
falkoschindler, line 67: Same here: Why is this change needed?

nicegui/elements/sub_pages.py
```diff
@@ -0,0 +1,245 @@
...
+
+def navigate_to_sub_page(path: str) -> bool:
+    """Handle navigation through sub_pages system.
+
+    :param path: the path to navigate to
+    :return: True if handled by sub_pages, False otherwise
+    """
+    try:
+        client = context.client
+        if client and client.content:
```
falkoschindler: How can `client` or `client.content` be falsy? I think we don't need this condition.

nicegui/elements/sub_pages.js
```diff
@@ -0,0 +1,25 @@
+export default {
+  template: `
+    <div class="nicegui-column">
+      <slot></slot>
+    </div>
+  `,
```
falkoschindler: What do you think about removing the container div?
```suggestion
  template: `<slot></slot>`,
```
This would place subpage content directly in the current context without nesting it in another (column-like) div, like a refreshable.
On the other hand, it might feel more intuitive for `ui.sub_pages` to behave like a real element, accepting style, classes etc. 🤔

nicegui/elements/sub_pages.py
```diff
@@ -0,0 +1,243 @@
...
+            match_result = self.find_route_match(full_path)
+            if match_result is None:
+                Label(f'404: sub page {full_path} not found')
+                return
+            self._place_content(match_result)
+            child_sub_pages = find_child_sub_pages_element(self)
+            if child_sub_pages:
+                child_sub_pages.show(full_path[len(match_result.path):])
+
+    def show_and_update_history(self, path: str) -> None:
```
falkoschindler: Two thoughts about this method:

- It is public. Is it meant to be called from outside?
- `_is_root` is checked many times:
    - before registering the `show_and_update_history` event handler in `__init__`
    - before calling `show_and_update_history` in `_on_routes_changed`
    - within `show_and_update_history`

  So it doesn't seem to be necessary to check it inside `show_and_update_history`.

nicegui/elements/sub_pages.py
```diff
@@ -0,0 +1,243 @@
...
+        self.show(path)
+        if self._is_root():
+            run_javascript(f'''
+                const fullPath = (window.path_prefix || '') + "{path}";
+                if (window.location.pathname !== fullPath) {{
+                    history.pushState({{page: "{path}"}}, "", fullPath);
+                }}
+            ''')
+
+    def find_route_match(self, path: str) -> Optional[RouteMatch]:
```
falkoschindler: Does this method need to be public?

nicegui/elements/sub_pages.py
```diff
@@ -0,0 +1,243 @@
...
+        """
+        super().__init__()
+        self._routes = routes or {}
+        self._root_path = root_path
+        self._on_routes_changed()
+        if self._is_root():
+            self.on('open', lambda e: self.show_and_update_history(e.args))
+            self.on('navigate', lambda e: self._handle_navigate(e.args))
+
+    def add(self, path: str, page: Callable) -> Self:
```
falkoschindler, line 62: Should we make `_routes` a public observable dictionary calling its change handler automatically? Then we could remove `.add()` and the user can simply call `sub_pages.routes['/new_route'] = show_new_route`.

nicegui/elements/sub_pages.py
```diff
@@ -0,0 +1,243 @@
...
+    def _convert_parameter(value: str, param_type: type) -> Any:
+        """Convert a string parameter to the specified type."""
+        if param_type is str or param_type is inspect.Parameter.empty:
+            return value
+        elif param_type is int:
+            return int(value)
+        elif param_type is float:
+            return float(value)
+        elif param_type is bool:
+            return value.lower() in ('true', '1', 'yes', 'on')
```
falkoschindler: I'm not sure about this list. Does it follow any convention? If "on" is `True`, why not "enabled", "active", "ok", or "positive"? And what about "1.0" or "42"? Or "Y"? I mean, it becomes a bit unpredictable.

nicegui/testing/general_fixtures.py
```diff
@@ -21,18 +21,32 @@ def pytest_configure(config: pytest.Config) -> None:
...
+
+    routes_to_remove = []
+    for route in list(app.routes):
+        if isinstance(route, Route) and '{' in route.path and '}' in route.path:
+            if (not route.path.startswith('/_nicegui/') and
+                    not ('{path:path}' in route.path or '{filename:path}' in route.path)):
+                routes_to_remove.append(route.path)
+
+    for path in routes_to_remove:
+        app.remove_route(path)
```
falkoschindler: Why do we need to collect `routes_to_remove` first before calling `remove_route`?
And the two if-conditions and the combination of `not`, `and` and `or` is pretty cryptic. Can we do better? What is the purpose of this block?

tests/test_sub_pages.py
```diff
@@ -0,0 +1,453 @@
...
+
+    screen.open('/')
+    screen.should_contain('child title')
+    screen.click('goto other')
+    screen.should_contain('other title')
+    screen.click('goto main')
+    screen.should_contain('child title')
+
+
+def test_accessing_sub_page_directly(screen: Screen):
```
falkoschindler: I don't understand the name of this test. Where are we "accessing [a] sub page directly"?

website/documentation/rendering.py
falkoschindler: This change breaks the subheading if `part.link` is `None`.

website/documentation/rendering.py
falkoschindler: This change breaks the "more" link if `part.link` is `None`.

website/documentation/content/sub_pages_documentation.py
```diff
@@ -0,0 +1,137 @@
...
+from . import doc
+
+
+@doc.demo('Sub Pages', '''
+    Sub pages provide url based navigation between different views.
+    This allows you to easily build a single page application (SPA).
+    The `ui.sub_pages` element itself functions as the container for the currently active sub page.
+    You only need to provide the routes for each view builder function.
+    NiceGUI takes care of replacing the content without triggering a full page reload when the url changes.
+    The `ui.sub_pages` element can be used as a context manager to automatically navigate to the correct sub page when the context is entered or exited.
```
falkoschindler: I don't understand this sentence. How would I use `ui.sub_pages` as a context manager? Should we add a demo?

website/documentation/content/sub_pages_documentation.py
```diff
@@ -0,0 +1,137 @@
...
+''')
+def main_demo() -> None:
```
falkoschindler, line 15: I'm not sure if we should mock these demos using a real `ui.sub_pages` element. By updating the browser URL, all demos interact and cause weird behavior. Can we simply update a label text to fake the navigation?

website/documentation/content/sub_pages_documentation.py
```diff
@@ -0,0 +1,137 @@
...
+        # ui.link('Go to main page', '/')
+        ui.link('Go to main page', f'/documentation/{location}')  # HIDE
+
+    # END OF DEMO
+    ui.label(f'This id {str(uuid4())[:6]} changes only on reload')
+    ui.separator()
+    ui.sub_pages({'/': main, '/other': other}, root_path=f'/documentation/{location}')
+
+
+@doc.demo('Passing Parameters to Sub Page', '''
```
falkoschindler, line 42: This demo seems to be broken. Maybe it's just because of the async main function, which also causes problems in the other async demo. But even when removing `async` it doesn't seem to work as expected.

website/documentation/content/sub_pages_documentation.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # END OF DEMO
+    with ui.row():
+        ui.label('Title:')
+        title = ui.label()
+    ui.separator()
+    ui.sub_pages({'/': lambda: main(title), '/other': lambda: other(title)}, root_path='/documentation/sub_pages')
+
+
+@doc.demo('Async Sub Pages', '''
```
falkoschindler, line 77: This demo seems to be broken (on the website as well as in isolation). I don't see any "delayed..." label.

nicegui/elements/sub_pages.js
falkoschindler: As discussed in person, we find it more intuitive to keep the div. 👍🏻

nicegui/elements/sub_pages.py
falkoschindler: Note that I converted the private method `_is_root()` into a public property `is_root`.

main.py
rodja: But why is the url to the font suddenly containing `/documentation`? The fonts are located in `http://127.0.0.1:8080/fonts/fira-sans-v16-latin-700.woff2`. I'll need to dig deeper here.

nicegui/functions/navigate.py
falkoschindler: No, I confused sub-page functions with page functions. Only page functions support `navigate.to` with a callable parameter.

nicegui/elements/sub_pages.py
rodja: No.

nicegui/elements/sub_pages.py
rodja: I would like to keep `_routes` as an implementation detail. That way we could later change it if needed.

nicegui/elements/sub_pages.py
rodja: We can drop `bool` altogether. It does not make much sense in paths.

website/documentation/content/sub_pages_documentation.py
rodja: Oh yes. That is totally wrong.

nicegui/testing/general_fixtures.py
falkoschindler: I think we broke test\_serving\_files.py.

rodja: No. We uncovered an existing issue: trailing slashes in URL paths (like /static/) created invalid double-slash route patterns (/static//{path:path}) that couldn't match normal requests. I fixed it by normalizing paths with rstrip('/') before route creation.

main.py
rodja: Ah, the problem was the relative links in `style.css`. I changed it to the more correct absolute links.

website/documentation/content/sub_pages_documentation.py
falkoschindler: We agreed to keep the interactive documentation as it is. It shows the power of `ui.sub_pages` more spectacularly.

website/documentation/content/sub_pages_documentation.py
falkoschindler: The async demo is working now.

website/documentation/content/sub_pages_documentation.py
falkoschindler: The async demo is working now.

website/documentation/rendering.py
falkoschindler: No, there is a `if part.link:` check above.

rodja:
After spending quite some time to implement #4905 and #4906, I came back to #4902. There I discovered an issue with the current core architecture: the root sub pages are doing all the rendering (eg. calling `show`) for itself and all its children. I designed it that way because `ui.sub_pages` inside `ui.sub_pages` should not care about the full url path but only the part which is not handled by it's parent already. That works quite well, until the child `ui.sub_pages` is added async or via background task to the element tree.

I'm now trying to change the architecture so that every `ui.sub_pages` element takes care of it's own rendering as soon as it appears. The most tricky part is the path stripping so that you can still write:

```py
    @ui.page('/')
    @ui.page('/{_:path}')
    def index():
        ui.sub_pages({'/demo': main})

    def main():
        ui.link('Go to A', '/sub/a')
        ui.link('Go to B', '/sub/b')
        ui.sub_pages({
            '/': sub_main,
            '/a': sub_page_a,
            '/b': sub_page_b
        })

    def sub_main():
        ui.label('sub main page')

    def sub_page_a():
        ui.label('sub A page')

    def sub_page_b():
        ui.label('sub B page')
```

Instead of

```py
    def main():
        ui.link('Go to A', '/sub/a')
        ui.link('Go to B', '/sub/b')
        ui.sub_pages({
            '/demo': sub_main,
            '/demo/a': sub_page_a,
            '/demo/b': sub_page_b
        })
```

rodja:
> I'm now trying to change the architecture so that every `ui.sub_pages` element takes care of it's own rendering as soon as it appears.

My first approach was to let every sub pages element be independent from all others. I tried to determine if a navigation event should be handled by a specific sub pages element. But I could not get it to work. Depending on the action, the path must match either exactly one of the routes or just a prefix of the path. After quite some forth and back (see 329aa0f), I'll now try to implement a more centralized routing which can keep track of everything.

rodja:
When working on #4902 I discovered that there were unintended full-page reloads when query parameters changed. But I could work that out in 3159c87.
Now I'm struggling with two rather tricky high-level requirements:

1. show a 404 error if a sub pages element can not completely resolve an url
2. allow normal FastAPI routes beside the single page routing (eg. for RESTful endpoints or other NiceGUI pages)

rodja:
**About the 404 error displaying**

`ui.sub_pages` can be nested. Therefore, when a sub pages element is looking for a matching route and only finds a partial path the remaining string is expected to be handled by nested sub pages elements. But whether there are any elements which can handle this can only be determined after building the page of the partial matching route. Only then it's clear if a 404 page should be shown or not. This changes the current assumptions about how often a page builder is called, and makes the routing over all a bit more complicated.

**About allowing normal FastAPI routes**

While working on the above I discovered that it sometimes is not correct to show a 404 page at all but rather use the normal FastAPI routing. I think I found a solution to this by checking if there are FastAPI routes to handle the navigation. But it does intermesh with my WIP of the 404 error displaying which is not yet ready to push.

main.py
falkoschindler: I'm still wondering why this change is needed (outside of PR #4902). If this PR requires any change to the "infrastructural" files inside the website directory, wouldn't that be a breaking change?

examples/single_page_app/custom_sub_pages.py
```diff
@@ -0,0 +1,70 @@
...
+
+class CustomSubPages(ui.sub_pages):
+    """Custom ui.sub_pages with built-in authentication and custom 404 handling."""
+
+    def _show_page(self, match: RouteMatch) -> None:
+        if self._is_route_protected(match.builder):
+            if not self._is_authenticated():
+                self._show_login_form(match.full_url)
+                return
+        super()._show_page(match)
```
falkoschindler: I think this method is called `_render_page` now.

examples/single_page_app/main.py
```diff
@@ -1,34 +1,50 @@
...
 
-    # this places the content which should be displayed
-    router.frame().classes('w-full p-4 bg-gray-100')
 
+def error():
+    raise ValueError('some error message')
 
-ui.run()
+
+@protected
```
falkoschindler, line 40: This decorator doesn't seem to work. My browser isn't asking for any credentials.

nicegui/elements/sub_pages.js
```diff
@@ -0,0 +1,67 @@
...
+  const cleanPath = event.state?.page || getCleanCurrentPath();
+  emitEvent("sub_pages_open", cleanPath);
+}
+
+function handleFragmentNavigation(href, targetUrl) {
+  const fragmentName = targetUrl.hash.substring(1);
+  const target = document.getElementById(fragmentName) || document.querySelector(`a[name="${fragmentName}"]`);
+  if (!target) return false;
+
+  target.scrollIntoView({ behavior: "smooth" });
```
falkoschindler, line 32: While our NiceGUI website uses smooth scrolling, other apps might prefer a different behavior. Maybe we should make it configurable.

falkoschindler (commented):
So far I reviewed a part of this PR, but the large files are still missing. I might not have time to continue before the weekend, therefore I'm submitting my first comments now.

examples/single_page_app/custom_sub_pages.py
rodja: I'm sorry. It's fixed now in 39986e7.

examples/single_page_app/main.py
rodja: That was caused by the wrong method overrides. Fixed in 39986e7.

nicegui/elements/sub_pages.js
rodja: Sure. But we could do this in later updates. Or do you think it will likely break the existing API?

main.py
rodja: > wouldn't that be a breaking change?

I don't think so. The change is necessary because the documentation now contains `ui.sub_pages` which has stricter requirements. As long as an existing code base does not use the new feature, there are no changes required.

rodja:
There had to be made [quite some changes](https://github.com/zauberzeug/nicegui/pull/4821/commits/3f278f7e8a3de370f2f6023fd4c424c558a8f35d) to satisfy the requirements "return 404 http status code when initial sub pages routing fails". It would have been bad to always return status code 200 if you create single page application without defining explicit routes in the page handler.

lanbaoshen:
How do I access static files when using `@ui.page('/{_:path}') `?

rodja:
@lanbaoshen currently you must define any other routes before the main page with the catch-all route. For example:

```py
app.add_static_files('/static', '/assets')

@ui.page('/')
@ui.page('/{_:path}')
def index():
    ui.label('some text before main content')
    ui.sub_pages({'/': child, '/other': child2})

...
```

This is how FastAPI determines the routing. Unfortunately this also means that auto-route creation like `ui.image('assets/logo.png')` does not work with the catch all. The route is created after the catch all is defined so FastAPI will match the catch-all first. We will improve/fix this in 3.0, see #5005.

PR #4822, opened 2025-06-01
Perf: Crackdown on style recals

evnchn:
### Motivation

Discord, someone was complaining that NiceGUI website slow and takes up over 20% of the CPU: https://discord.com/channels/1089836369431498784/1357872936694059078

### Implementation

- Drop the following which constantly causes style recals:
  - Drop the `star-pulse` animation (no replacement so far)
  - Drop the use of `ui.spinner` in favor of custom-made loading GIF

### Progress

### Results

Considering `/documentation/section_text_elements`

#### Before

{BE3FE9C2-A4DC-4CEA-8D8A-316B9ED30C4A}

#### After

{8312AF0F-BED1-4868-86A3-9939AA05F627}

### Script for generating the loading GIF

Surprised to find no royalty-free loading GIF. Proceed to made myself one:

```py
from PIL import Image, ImageDraw

import numpy as np

def create_loading_gif(size=128):
    frames = []
    num_dots = 8
    radius = size / 2.5
    center = (size // 2, size // 2)
    color = (88, 152, 212)

    for i in range(num_dots):
        frame = Image.new('RGBA', (size, size), (255, 255, 255, 0))
        draw = ImageDraw.Draw(frame)

        for j in range(num_dots):
            angle = 2 * np.pi * j / num_dots
            x = center[0] + radius * np.cos(angle)
            y = center[1] + radius * np.sin(angle)
            opacity = 255 if (0 < (j - i)%num_dots <= 3) else 0
            print(f"Drawing dot {j} at angle {angle:.2f} radians with opacity {color + (opacity,)}")
            draw.ellipse([(x - 10, y - 10), (x + 10, y + 10)], fill=color + (opacity,))

        frames.append(frame)

    # Save as GIF
    frames[0].save('loading.gif', save_all=True, append_images=frames[1:], optimize=False, duration=100, loop=0, disposal=2)

create_loading_gif()
```


evnchn:
# 100% CPU usage in Quasar QSpinner demo 🤡

{F184FAD8-BCB7-466F-83D7-16ED8A314484}

EmberLightVFX:
Feels like this should be reported to Quasar too 😄

evnchn:
IMO No, since style recals are unavoidable if you use CSS animations

falkoschindler (approved):
I just stumbled across this PR. Looks ready to merge! Thanks, @evnchn!

evnchn:
Note that the star no longer pulses, but I think it's a minor thing to compromise for massively better performance (we can make a GIF later)

PR #4824, opened 2025-06-01
Fix hierarchy scroll-to for mobile situation

evnchn:
### Motivation

While working on another PR, I noticed that the NiceGUI documentation kept throwing errors in the console, and I realized that the scroll-into-view functionality never worked for mobile users, since the DOM is not populated until the tree is expanded. 

### Implementation

- Use `?.` such that it doesn't throw an error, instead fails silently. 
- Run the scroll-into-view on toggling the left drawer. 

### Progress

### Results

Before: 

{5497DE04-201A-448D-9516-65AC9775660C}

-> Wrong scroll position, errors thrown

After: 

{6B9B70DB-8718-4852-A5E7-3BBED9A3232F}

-> Correct scroll position, no errors


evnchn:
This PR is un-mergable after #4902. The entire logic is different and the scroll-to appears to be gone.

PR #4820, opened 2025-05-31
Feat: More controls for `ui.scene`

evnchn:
### Motivation

Addresses https://github.com/orgs/zauberzeug/projects/6/views/1?pane=issue&itemId=82940012, how we are stuck with OrbitControls

### Implementation

- Grab all of the controls related to moving the camera
- Add `setup_controls` and `fetch_controls` to `scene.js`
- Add `THREE.Clock();` and feed the delta to get FlyControls and FirstPersonControls happy. 
- Some controls don't have target. Use `this.controls.target?.` to get those happy. 

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests have been added (or are not necessary).
- [x] Documentation has been added (or is not necessary).

Issue: Some controls don't work properly. Do we want to drop them and release the working ones first, or spend more time get them working? 

Test script: 

```py
control_types = [
    ('Orbit controls (default)', None),  # Before the PR
    ('Arcball controls', 'arcball'),
    ('Trackball controls', 'trackball'),
    ('Map controls', 'map'),
    ('Fly controls', 'fly'),
    ('First-person controls', 'first-person'),
    ('Pointer lock controls', 'pointer-lock'),
]

for label, control_type in control_types:
    ui.label(label)
    with ui.scene(width=285, height=220, **({'control_type': control_type} if control_type else {})) as scene:
        scene.sphere()
        scene.move_camera(x=1, y=-1, z=1.5)
```

We have decided to ship the working ones. 

- WORKING: Orbit controls: Working
- Arcball controls: Has a tendency to jump around sproadically
- WORKING: Trackball controls: Working
- WORKING: Map controls: Working
- Fly controls: Cannot control direction
- First-person controls: Cannot control direction
- Pointer lock controls: Cannot move camera

---

Feature request: #3710

PR #4825, opened 2025-06-01
Feat: `fps` and `show_stats` for `ui.scene`

evnchn:
### Motivation

While working on #4820, I noticed that the FPS of NiceGUI's `ui.scene` is low. I originally though it was Three.js performance issue, but their demo's running at a smooth 120FPS, so I thought what gives?

As it turns out, it's hard-coded to 20FPS (`requestAnimationFrame(() => setTimeout(() => render(), 1000 / 20));`)

While we're at it, added on the `Stats` module from `Three.js` to visualize difference. 

### Implementation

- Don't hardcode it, accept `fps` prop
- Add on the `Stats` module from `Three.js` to visualize difference. 
  - Update `npm.json`

### Progress

-> Pytest should be unnecessary, otherwise the testing load may be huge.

### DIFF
--- a/nicegui/elements/lib/three/modules/stats.module.js
+++ b/nicegui/elements/lib/three/modules/stats.module.js
@@ -1,0 +1,167 @@
+var Stats = function () {
+
+	var mode = 0;
+
+	var container = document.createElement( 'div' );
+	container.style.cssText = 'position:fixed;top:0;left:0;cursor:pointer;opacity:0.9;z-index:10000';
+	container.addEventListener( 'click', function ( event ) {
+
+		event.preventDefault();
+		showPanel( ++ mode % container.children.length );
+
+	}, false );
+
+	//
+
+	function addPanel( panel ) {
+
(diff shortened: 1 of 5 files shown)


falkoschindler:
Thanks, @evnchn!

Quick question: Am I right to assume that we need to add Stats.js to npm.json?

evnchn:
Yes. You'd need that to download the file. 

See image

evnchn:
@falkoschindler According to my instinct, you may be wanting this PR for the customizable FPS to facilitate some robots powered by RoSys. 

If that's the case we should deliver the FPS customization change first, then we do the stats later (TM).

evnchn:
In favor of #5344 and evnchn#52

falkoschindler:
> According to my instinct, you may be wanting this PR for the customizable FPS to facilitate some robots powered by RoSys.

Not really. Three.js runs in the browser, so the frame rate doesn't affect our robots.

PR #4826, opened 2025-06-01
Eliminates B006 warnings from root-cause

evnchn:
### Motivation

Addresses https://github.com/orgs/zauberzeug/projects/6?pane=issue&itemId=82940985

### Implementation

- I tried `click_events: List[str] = ('click', 'dblclick')`, but that would cause mypy to complain. 
- `click_events: Sequence[str] = ('click', 'dblclick')` could be doable, but we are changing `List` to `Sequence`, not sure if that'd be a breaking change
- Wound up asking Copilot to "no mutable default" and its solution was `click_events: Optional[List[str]] = None` with `if click_events is None:` `click_events = ['click', 'dblclick']`

### Progress

Tests are green at https://github.com/evnchn/nicegui/actions/runs/15380276062, but the point about whether this is a breaking change is still worthy.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -88,14 +88,18 @@ class Element(Visibility):
 
     def __init_subclass__(cls, *,
                           component: Union[str, Path, None] = None,
-                          dependencies: List[Union[str, Path]] = [],  # noqa: B006
-                          libraries: List[Union[str, Path]] = [],  # noqa: B006  # DEPRECATED
-                          exposed_libraries: List[Union[str, Path]] = [],  # noqa: B006  # DEPRECATED
-                          extra_libraries: List[Union[str, Path]] = [],  # noqa: B006  # DEPRECATED
+                          dependencies: Optional[List[Union[str, Path]]] = None,
+                          libraries: Optional[List[Union[str, Path]]] = None,  # DEPRECATED
+                          exposed_libraries: Optional[List[Union[str, Path]]] = None,  # DEPRECATED
+                          extra_libraries: Optional[List[Union[str, Path]]] = None,  # DEPRECATED
                           default_classes: Optional[str] = None,
                           default_style: Optional[str] = None,
                           default_props: Optional[str] = None,
                           ) -> None:
+        dependencies = dependencies or []
+        libraries = libraries or []
+        exposed_libraries = exposed_libraries or []
+        extra_libraries = extra_libraries or []
         super().__init_subclass__()
         base = Path(inspect.getfile(cls)).parent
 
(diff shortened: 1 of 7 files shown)


falkoschindler:
> `click_events: Sequence[str] = ('click', 'dblclick')` could be doable, but we are changing List to Sequence, not sure if that'd be a breaking change

In my point of view this is not a breaking change because we're loosening the type constraint. Since `List` is a `Sequence`, existing user code is still valid. Note that return types would work the other way around: Changing `Sequence` to the more specific `List` would be ok, but going from `List` to `Sequence` would be a breaking change because the user might rely on the return value being mutable which `Sequence` isn't guaranteed to be.

In my opinion, using `Optional` is a tiny bit confusing because the user might think: "Ok, it can be a `List` or `None`. I don't want any click events. So lets pass `None`." But `None` leads to two default click arguments, not an empty list.

And `Sequence` has the advantage that the user can pass tuples. This is considered good practice for type annotation: accept broad, return narrow.

evnchn:
Interesting. 

How about `Dict`? No `frozendict`. `MappingProxyType`? https://adamj.eu/tech/2022/01/05/how-to-make-immutable-dict-in-python/

evnchn:
Also, if I understand correctly, since tuples can alo be `Sequence`, then it means that we no longer can leverage `append` or mutate the input in any way, Thus, we'd need to copy it to a new list if we need to mutate it, or we need to change our code to never mutate, right?

falkoschindler:
> How about `Dict`?

Good question. `options: Mapping: MappingProxyType({})` seems to be technically correct. I just hope that it doesn't confuse anyone too much.

Actually, one reason why we kept using `List` and `Dict` instead of `Sequence` and `Mapping` was that it feels more readable for beginners who might not know these advanced datatypes.

Another aspect I haven't though of: If someone derived, e.g., `ui.interactive_image` and assumes the `events` parameter is of type `List`, changing it to `Sequence` would indeed break the derived class. Therefore we need to wait for 3.0.

> Thus, we'd need to copy it to a new list if we need to mutate it, or we need to change our code to never mutate, right?

That's the point of the whole exercise. Our code should never mutate such input parameters because their default value is "static". If one `ui.interactive_image` would append something to the default `events` list, it would affect the default value of all other `ui.interactive_image`s. (See https://docs.astral.sh/ruff/rules/mutable-argument-default/.)

As far as I can tell, within the NiceGUI library we already never mutate default arguments. But user code could - accidentally or intentionally - break this convention, allowing for nasty bugs down the line. By changing the signatures to immutable types, we enforce derived classes to follow this rule as well.

evnchn:
Well https://github.com/zauberzeug/nicegui/pull/5505 fools the type checker to not have B006 errors. Accidental win I guess.

PR #4828, opened 2025-06-02
Force Vue re-render on late event registration

### LINKED ISSUE #4154: Events such as keydown and keyup are not triggered by an ui.input when not assigned directly upon creation
### Description

During the development of a Typeahead component which attaches itself visually to the bottom of an ui.input field to (later) show search result data from a DB, i stumbled over a potential bug when modifying certain event_listeners such as `keydown` and `keyup` via the `element.on` method "post creation".

image

Situation:
* An ui.input field is created
* A focus handler is set up via `.on("focus", handle_focus)` and a blur_handler is setup via `on("blur", handle_blur)`
* In `handle_focus` the`.on("keydown", handle_keydown)` is assigned to track cursor key presses
* In `handle_blur` the event listener is removed again.
   * ***As NiceGUI doesn't officially support removal of event listeners added via `.on` yet there is a helper class involved which "cleans up again" / can unassign event listeners, but that just as side note and has no impact on this report / test as you can see in the example code attached below.***
* Expected behavior: handle_keydown will be called when a key is pressed.
* Actual behavior: It is not

If a "keypress" or "keyup" listener is assigned *later* and not directly upon creation of the input field, so e.g. delayed by a timer or an event, then the handlers will never be called. If they are assigned directly upon creation, so before everything goes through the websocket from Python to JS, then they work though.

If I assign the same events directly via JavaScript to the DOM element w/ help of the new `getHtmlElement` function though then everything works perfectly fine. Thus there seems to be some synchronization kind of issue for certain events. A delayed assigned `blur` event just for example works like a charm.

```python
from nicegui import ui

@ui.page('/')
async def index():
    log_area = ui.scroll_area().style('height: 200px; width: 400px')

    def log_key(e):
        with log_area:
            ui.label(f"Key pressed: {e.args['key']}").classes('font-mono text-xs').style('line-height: 1; margin: 0; padding: 0;')
        log_area.scroll_to(percent=1.0)

    with ui.row():
        input1 = ui.input()
        ui.run_javascript(f'''
            getHtmlElement({input1.id}).addEventListener("keydown", () => console.log("keydown on input1 via js"));
            ''')
        input1.on('keydown', log_key)   # will be triggered
        input1.on('keydown', js_handler='() => console.log("keydown on input1")')  # will be triggered
        input2 = ui.input()

    def attach_event():
        input2.on('keydown', log_key)  # will NOT be triggered
        input2.on('keydown', js_handler='() => console.log("keydown on input2")')  # will NOT be triggered
        input2.on('blur', lambda e: print('blur'))  # will be triggered
        ui.run_javascript(f'''
            getHtmlElement({input2.id}).addEventListener("keydown", () => console.log("keydown on input2 via js"));
            ''')  # will be triggered    

    # setup timer which attaches the event
    ui.timer(1.0, attach_event, once=True, immediate=False)
    
ui.run()
```

evnchn:
### Motivation

This PR fix #4154, enabling late event registration like:

```py
@ui.page('/')
async def page():
    ui.label('This is a test for late event handling')
    i = ui.input()
    i.on('keydown', ui.notify)
```

{A7397251-83CE-4739-93CB-0EB4690EA8D5}

As you can see, `i.on('keydown', ui.notify)` was ran after the client was connected, so it must be sent via Socket.IO and client-side shall register the event after initial render. Before, it doesn't work, as mentioned in #4154 somewhere. Now, with this PR, it works. 

### Implementation

Before we process the list of elements, we do a check to see if there are any new events sent in by the server. 
- If there is, then:
  - We remove that element from Vue's `this.elements`
  - We mark that we need to `await this.$nextTick();`
  - We proceed to `await this.$nextTick();` to force the Vue to render without that element
    - Do it ONCE only, for performance
  - In such a way, when we process the list of elements as usual, Vue will be like 
    - "hey I haven't seen this element before, let me render it"
- If there is none, then fine, do nothing. 

### Progress

I don't think this needs documentation. ~~Thinking about how to Pytest this.~~ Pytest done

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -392,7 +392,6 @@ function createApp(elements, options) {
 
             const oldTypes = new Set((this.elements[id]?.events || []).map(ev => ev.type));
             if (element.events?.some(e => !oldTypes.has(e.type))) {
-              console.log("Re-rendering element " + id + " because of new events");
               delete this.elements[id];
               needAwaitNextTick = true;
             }
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -385,6 +385,21 @@ function createApp(elements, options) {
             .map(([_, element]) => loadDependencies(element, options.prefix, options.version));
           await Promise.all(loadPromises);
 
+          let needAwaitNextTick = false;
+
+          for (const [id, element] of Object.entries(msg)) {
+            if (element === null) continue;
+
+            const oldTypes = new Set((this.elements[id]?.events || []).map(ev => ev.type));
+            if (element.events?.some(e => !oldTypes.has(e.type))) {
+              console.log("Re-rendering element " + id + " because of new events");
+              delete this.elements[id];
+              needAwaitNextTick = true;
+            }
+          }
+
+          if (needAwaitNextTick) await this.$nextTick();
+
           for (const [id, element] of Object.entries(msg)) {
             if (element === null) {
               delete this.elements[id];
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -206,3 +206,18 @@ def test_delegated_event_with_argument_filtering(screen: Screen) -> None:
     screen.click('Item B')
     screen.click('Item C')
     assert ids == ['A', 'B', 'C']
+
+
+async def test_late_event_registration(screen: Screen):
+    events = []
+
+    @ui.page('/')
+    async def page():
+        a_input = ui.input('A')
+        await ui.context.client.connected()
+        a_input.on('keydown', lambda _: events.append('A'), [])
+
+    screen.open('/')
+    await asyncio.sleep(1)  # wait for the page to load and the event to be attached
+    screen.selenium.find_element(By.XPATH, '//*[@aria-label="A"]').send_keys('x')
+    assert events == ['A']


evnchn:
Additional notes: 

Force Vue re-render: https://michaelnthiessen.com/force-re-render/

-> Did not figure out where to call `$forceUpdate()`, did not have a key to change, but deleting the element and shortly populating it back, is in-spirit to the **v-if hack** it mentioned. 

@xaptronic on Vue's issue on late-attaching event handler on https://github.com/zauberzeug/nicegui/issues/4154#issuecomment-2885368501

-> It was super insightful and was the inspiration to this solution.

evnchn:
Pytest done in https://github.com/zauberzeug/nicegui/pull/4828/commits/27b3b1a08949629392f1fdd233c6c37a7b26ff48

The test fails on main, but with this PR it passes. Commenting out `delete this.elements[id];` or `if (needAwaitNextTick) await this.$nextTick();` also breaks the test, showing the validity of this PR.

evnchn:
Reminder to self: Since we're quickly deleting and then restoring the element: 

- If I place my cursor in the input, will it move around? Will it drop focus?
- If I am on mobile, does my keyboard quickly withdraw (and does it come back)?
- Will the text I type still be present afterwards?

Hmm 🤔

evnchn:
So, 

- Focus is dropped (desktop, Chrome 126)
- As a result, mobile keyboard is likely withdrawn and never coming back automatically unless user click on it
- Text typed still present, though, so that's one good thing I guess

-> Perhaps the `$forceUpdate()` or key-changing method is better, but I can't figure it out how to do it

-> Since we always endorse early-attachment of events, and then acting on the events selectively, can we accept the defect of dropping focus as a known-but-wont-fix or known-but-fix-later? I'd assume it's better to dop focus than to miss events outright.

falkoschindler:
@evnchn Are you still working on this one, or should I start reviewing?

evnchn:
Nope, I'm not working. 

You can review this and the rest of the PRs. Thanks!

falkoschindler:
After noticing that #4154 seems to be fixed after updating Vue and Quasar, I don't think we need this PR anymore. No need to add extra complexity. 😌

evnchn:
@falkoschindler how about the 3 questions in https://github.com/zauberzeug/nicegui/pull/4828#issuecomment-2940042694 ?

I'd assume they all lead to positive answers since the upstream definitely do a better job than downstream hotfixes...

falkoschindler:
@evnchn I don't understand. If the original bug #4154 doesn't occur anymore, your questions about "quickly deleting and then restoring the element" seem obsolete. 🤔

evnchn:
Oh I just mean if focus is dropped when we dynamically add event listener after the bump in dependency. 

You're right in that we are no longer deleting an element, so I'm guessing focus is not dropped.

evnchn:
https://github.com/zauberzeug/nicegui/issues/4154#issuecomment-3199294757

Should we re-consider this?

falkoschindler:
@evnchn Well, as it turned out, I closed it by accident. We can re-open and assign it to 3.x. But you'd need to restore the branch first.

PR #4836, opened 2025-06-04
Provide autocompletion hints for ui.codemirror()

Yuerchu:
This PR optimizes the syntax hints for the `set_language()` and `set_theme()` methods in `ui.codemirror()`.  

I noticed there are two Literal types in [codemirror.py](https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/codemirror.py), which define the supported languages and themes for codemirror.  

However, when calling `set_language()`, I found `language: Optional[str]`. I believe changing it to `language: Optional[SUPPORTED_LANGUAGES] = None` would provide a better IDE autocompletion experience.  

This was the motivation behind my improvements to this method.  

I also took the opportunity to modify the `set_theme()` method so that it can benefit from similar syntax autocompletion.

## Before
image
image

## After
image
image

### DIFF
--- a/nicegui/elements/codemirror.py
+++ b/nicegui/elements/codemirror.py
@@ -301,11 +301,11 @@ class CodeMirror(ValueElement, DisableableElement, component='codemirror.js', de
         return self._props['theme']
 
     @theme.setter
-    def theme(self, theme: str) -> None:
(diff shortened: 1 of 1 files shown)


evnchn:
Makes total sense. Should have been this way since day1, since the initializer takes those Literal types already.

PR #4299, opened 2025-02-03
Add to `redis.from_url` connection parameters

alydersen:
This pull request includes an enhancement to the `RedisPersistentDict` class in the `nicegui/persistence/redis_persistent_dict.py` file to improve the reliability and performance of the Redis client connection.

Connection improvements to `RedisPersistentDict`:

* [`nicegui/persistence/redis_persistent_dict.py`](diffhunk://#diff-d52efa243ec1e0f3d58104376c85f0e294b7f2371679b32296c7fdd426195978L17-R23): Updated the initialization of the `redis_client` to include additional parameters such as `health_check_interval`, `socket_connect_timeout`, `retry_on_timeout`, and `socket_keepalive` to enhance connection stability and performance.

PR #4840, opened 2025-06-05
Better Redis control for users

alydersen:
### Motivation

This PR adds some more docs to using Redis, as well as allows for the following environment variables to be set to control Redis connection properties:
- `NICEGUI_REDIS_HEALTH_CHECK_INTERVAL` (default: 10) as connection argument `health_check_interval`
- `NICEGUI_REDIS_SOCKET_CONNECT_TIMEOUT` (default: 5) as connection argument `socket_connect_timeout`
- `NICEGUI_REDIS_RETRY_ON_TIMEOUT` (default: True) as connection argument `retry_on_timeout`
- `NICEGUI_REDIS_SOCKET_KEEPALIVE` (default: True) as connection argument `socket_keepalive`

### Implementation

`redis_persistant_dict.py` will check for the variables above and convert strings to booleans

### Progress

If applied, this PR will improve docs on using Redis and allow for more user control in the Redis connection.

### DIFF
--- a/examples/redis_storage/README.md
+++ b/examples/redis_storage/README.md
@@ -1,0 +1,22 @@
+# Using Redis for session storage
+
+NiceGUI can use Redis as the session storage, which is particularly interesting in environments where you want session persistance or there are multiple workers. To make this work, you set environment variables:
+
+## `NICEGUI_REDIS_URL`
+
+This is the only required variable. Setting it to a full Redis server connection string will make NiceGUI switch to Redis for storage
+
+## `NICEGUI_REDIS_PREFIX`
+
+The prefix that will be used on the Redis server for the values set from the NiceGUI server. Default is `nicegui:`. If you have two environments using the same Redis server/cluster, use this prefix to separate them.
+
+## Control how NiceGUI establishes the Redis connection
+
+When NiceGUI starts, and the `NICEGUI_REDIS_URL` is set, it will establish the connection. Sometimes you want to control the arguments for this connection (you can read more on the Python Redis package documentation). The following are defined as environment variables:
+
+- `NICEGUI_REDIS_HEALTH_CHECK_INTERVAL` (default: 10) as connection argument `health_check_interval`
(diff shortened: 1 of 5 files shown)


evnchn:
OK so last time we took a shot at making it support multiple workers, the issue was https://github.com/zauberzeug/nicegui/discussions/1539#discussioncomment-12496301

Note how in NiceGUI, there’s the HTTP response (initial), and the Socket.IO connection (follow-up)

The issue was, that for Uvicorn workers, we can’t guarantee that the worker who served you the HTTP response is the same worker for the Socket.IO connection (aka no session affinity), and so the Socket.IO workers are like “bro who are you, I don’t even know you, why are you suddenly talking to me” and the connection fails. 

Preliminarily, I don’t see how this PR addresses this fundamental limitation, and so I would tend to believe that we will more likely close this. 

---

Not all hope is lost, since we can run multiple instances as stated in https://github.com/zauberzeug/nicegui/discussions/1539#discussioncomment-12517031

evnchn:
So the PR simplifies to "Allow control Redis connection properties via environment variables", then?

alydersen:
Sorry for not reading the existing Github issues closely on this - don't know how that happened. I've changed the purpose of this PR to just allow for more user control @evnchn

rodja:
I'm not sure about having these many new env variables. But a single `REDIS_CONFIG` which contains json is also not perfect. What about reading a `.redis_config.yaml`?

evnchn:
Sure. Both solutions makes more sense than having one environment variable per connection parameter. 

Imagine if Redis (or some Redis-compatible alternative) suddenly took one more parameter. Or 100 of them. Then it seems that one centralized file for Redis-related connection configuration helps with organization. 

In retrospect, I'm not sure why some parameters are configurable. I get why some timeout values need to be tweaked: maybe you have a really slow server? But, to ensure that the Redis is always available, shouldn't we always turn on retry on timeout and socket keepalive?

alydersen:
I'm totally OK with this not being implemented. The reason for adding them was a follow-up of the comments made in #4299

PR #4841, opened 2025-06-05
Remove wait_for2 dependency

rodja:
### Motivation

I introduced the `wait_for2` package in #4641 because it fixed cancellation in test tear downs. But further modifications to the code seems to have made the package obsolete again.

### Implementation

I removed the dependency and the helper. Now ~~`asyncio.wait_for2`~~  `asyncio.wait_for` can be used directly again.

### Progress

- [x] New Pytests are not necessary

### DIFF
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import gzip
 import json
 import logging
@@ -225,7 +226,7 @@ class Air:
         self.connecting = True
         try:
             if self.relay.connected:
-                await helpers.wait_for(self.disconnect(), timeout=5)
+                await asyncio.wait_for(self.disconnect(), timeout=5)
             self.log.debug('Connecting...')
             await self.relay.connect(
                 f'{RELAY_HOST}?device_token={self.token}',
--- a/nicegui/background_tasks.py
+++ b/nicegui/background_tasks.py
@@ -5,7 +5,7 @@ import asyncio
 import weakref
 from typing import Any, Awaitable, Callable, Coroutine, Dict, Set, TypeVar
 
-from . import core, helpers
+from . import core
 from .logging import log
 
 running_tasks: Set[asyncio.Task] = set()
@@ -22,7 +22,7 @@ def create(coroutine: Awaitable, *, name: str = 'unnamed task') -> asyncio.Task:
     See https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task.
     """
     assert core.loop is not None
-    coroutine = coroutine if asyncio.iscoroutine(coroutine) else helpers.wait_for(coroutine, None)
+    coroutine = coroutine if asyncio.iscoroutine(coroutine) else asyncio.wait_for(coroutine, None)
     task: asyncio.Task = core.loop.create_task(coroutine, name=name)
     task.add_done_callback(_handle_task_result)
     running_tasks.add(task)
@@ -91,7 +91,7 @@ async def teardown() -> None:
         if tasks:
             await asyncio.sleep(0)  # NOTE: ensure the loop can cancel the tasks before it shuts down
             try:
-                await helpers.wait_for(asyncio.gather(*tasks, return_exceptions=True), timeout=2.0)
+                await asyncio.wait_for(asyncio.gather(*tasks, return_exceptions=True), timeout=2.0)
             except asyncio.TimeoutError:
                 log.error('Could not cancel %s tasks within timeout: %s',
                           len(tasks),
--- a/nicegui/helpers.py
+++ b/nicegui/helpers.py
@@ -7,11 +7,8 @@ import struct
 import threading
 import time
 import webbrowser
-from collections.abc import Awaitable
 from pathlib import Path
 from typing import Any, Optional, Set, Tuple, Union
-
-import wait_for2
 
 from .logging import log
 
@@ -113,12 +110,3 @@ def kebab_to_camel_case(string: str) -> str:
 def event_type_to_camel_case(string: str) -> str:
     """Convert an event type string to camelCase."""
     return '.'.join(kebab_to_camel_case(part) if part != '-' else part for part in string.split('.'))
-
-
-async def wait_for(fut: Awaitable, timeout: Optional[float] = None) -> None:
-    """Wait for a future to complete.
-
-    This function is a wrapper around ``wait_for2.wait_for`` which is a drop-in replacement for ``asyncio.wait_for``.
-    It can be removed once we drop support for older versions than Python 3.13 which fixes ``asyncio.wait_for``.
-    """
-    return await wait_for2.wait_for(fut, timeout)
--- a/nicegui/javascript_request.py
+++ b/nicegui/javascript_request.py
@@ -3,8 +3,6 @@ from __future__ import annotations
 import asyncio
 from typing import Any, ClassVar, Dict
 
-from . import helpers
-
 
 class JavaScriptRequest:
     _instances: ClassVar[Dict[str, JavaScriptRequest]] = {}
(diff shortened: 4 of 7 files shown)


falkoschindler:
> Now `asyncio.wait_for2` can be used directly again.

I assume you mean `asyncio.wait_for`.

> But further modifications to the code seems to have made the package obsolete again.

Can you tell _which_ modifications made it obsolete?
How can we be sure that the problem with `asyncio.wait_for` is gone?

rodja:
> I assume you mean asyncio.wait_for.

Yes, of course. Sorry.

> Can you tell which modifications made it obsolete?

I think it was just a misconception on my side. When running 

```sh
docker compose run --remove-orphans app ptw . --now -k "user_simulation"
```

on commit [c8419bb](https://github.com/zauberzeug/nicegui/pull/4641/commits/c8419bb56bf701ea1fed2e9c8059c6e5e1459555), I get 

```
tests/test_user_simulation.py .........E...............................                                                                                                    [100%]

===================================================================================== ERRORS =====================================================================================
______________________________________________________________________ ERROR at teardown of test_navigation ______________________________________________________________________
There were unexpected ERROR logs.
------------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------------
ERROR    asyncio:base_events.py:1865 Task was destroyed but it is pending!
task: <Task pending name='storage-user-58cc531f-0567-4305-858b-91e3156b1e3c' coro=<FilePersistentDict.backup.<locals>.async_backup() done, defined at /app/nicegui/persistence/file_persistent_dict.py:48> wait_for=<Future pending cb=[_chain_future.<locals>._call_check_cancel() at /usr/local/lib/python3.13/asyncio/futures.py:384, Task.task_wakeup()]> cb=[_handle_task_result() at /app/nicegui/background_tasks.py:67, <1 more>, create_lazy.<locals>.<lambda>() at /app/nicegui/background_tasks.py:48]>
============================================================================ short test summary info =============================================================================
ERROR tests/test_user_simulation.py::test_navigation - Failed: There were unexpected ERROR logs.
================================================================== 40 passed, 474 deselected, 1 error in 4.80s ===================================================================
```

But after introducing `wait_for2` and fixing some other issues in [50782f9](https://github.com/zauberzeug/nicegui/pull/4641/commits/50782f9d491f9077803655fddf9f05cd59a7fca1) tests pass. Even if I replace `wait_for2` with `asyncio`.

--- 

Here is Cursor (with Sonnet-4) assessing the risk of standard ~~`asyncio.wait_for2`~~ `asyncio.wait_for`:

Based on the `wait_for2` documentation from the GitHub repository, it addresses several specific race conditions that can occur with `asyncio.wait_for()`. Let me explain the key race conditions it guards against:

## 1. **Simultaneous Cancellation and Completion Race**

**The Problem:**
```python
# Task A
task = asyncio.create_task(some_operation())

# Task B 
wait_task = asyncio.create_task(asyncio.wait_for(task, timeout=5.0))

# If Task A completes AND Task B gets cancelled at the exact same moment:
# - Python 3.7: You lose the completed result 
# - Python 3.8+: You lose the cancellation request
```

**Real-world scenario in NiceGUI context:**
```python
# JavaScript request waiting for response
request_task = asyncio.create_task(event.wait())
wait_task = asyncio.create_task(asyncio.wait_for(request_task, timeout=5.0))

# User closes browser tab (triggers cancellation) 
# EXACTLY when JavaScript response arrives
# Result: Either lose the response data OR ignore the cancellation
```

## 2. **Timeout + Completion Race**

**The Problem:**
```python
# Timeout expires at 5.000000 seconds
# Future completes at 5.000001 seconds  
result = await asyncio.wait_for(future, timeout=5.0)

# Different Python versions handle this differently:
# - Some prioritize TimeoutError (lose the result)
# - Others prioritize the actual result
```

**Real-world scenario:**
```python
# Background task cleanup during shutdown
await asyncio.wait_for(
    asyncio.gather(*tasks, return_exceptions=True), 
    timeout=2.0
)
# If tasks complete just after 2.0 seconds timeout,
# you might miss cleanup completion status
```

## 3. **Cancellation During Timeout Handling**

**The Problem:**
```python
# Complex scenario: timeout starts, then external cancellation occurs
# Some Python versions don't properly wait for inner future termination
# Result: "zombie" tasks that keep running after wait_for returns
```

## **Analysis for NiceGUI**

Looking at NiceGUI's usage patterns:

### **Low Risk Areas:**
- **JavaScript requests**: Usually short-lived, clear timeout bounds
- **Event waiting**: Simple event.wait() patterns
- **Background task teardown**: Happens during controlled shutdown

### **Potential Risk Areas:**
1. **Client disconnection during active requests**:
   ```python
   # nicegui/javascript_request.py
   yield from asyncio.wait_for(self._event.wait(), self.timeout).__await__()
   # If client disconnects exactly when JS responds, could lose response
   ```

2. **Shutdown race conditions**:
   ```python
   # nicegui/background_tasks.py  
   await asyncio.wait_for(asyncio.gather(*tasks, return_exceptions=True), timeout=2.0)
   # If tasks complete just after timeout during shutdown, cleanup status unclear
   ```

3. **Air relay disconnection timing**:
   ```python
   # nicegui/air.py
   await asyncio.wait_for(self.disconnect(), timeout=5)
   # Cancellation during connection cleanup could leave connections hanging
   ```

## **Verdict**

For NiceGUI, these race conditions are **theoretical edge cases** that would:
- Occur in microsecond timing windows
- Require very specific concurrent conditions  
- Generally have minimal user-visible impact
- Be extremely difficult to reproduce or test

The practical benefits of `wait_for2` don't justify the dependency for NiceGUI's use cases. The standard `asyncio.wait_for()` is sufficient, and any edge case issues would be masked by the web application's natural resilience patterns (reconnection, retry logic, etc.).

**Good decision to remove it!** The complexity and maintenance burden outweigh the protection against these rare edge cases.

falkoschindler:
> Here is Cursor (with Sonnet-4) assessing the risk of standard `asyncio.wait_for2`:

I assume you mean `asyncio.wait_for`. 😜

falkoschindler (approved):

evnchn:
Theoretically though, `wait_for2` does make the normal wait_for more robust, but if those edge cases are not hit by NiceGUI (which is apparently a yes), we can remove `wait_for2`.

@rodja have you tried to put this in production and serve many clients? I have concerns over that, since it's a part of why in NiceGUI 2.14.0 we didn't catch the issue...

Getting back, but what's the maintenance cost of including and using `wait_for2`? I felt that it's like an umbrella for a potential rain. Better than nothing.

falkoschindler:
@evnchn Yes, I had a similar thought. That's why I hesitated for a moment before merging the PR. Once the dependency had been added (and the corresponding Conda package has been released), we could simply keep it for extra safety.

But as mentioned, to release [NiceGUI on Anaconda](https://anaconda.org/conda-forge/nicegui), we had to release [wait_for2](https://anaconda.org/conda-forge/wait_for2) first. Other distributions have to do similar work (like https://www.freshports.org/devel/py-wait-for2/), so it helps making NiceGUI accessible for other distributions to come by removing such questionable dependencies.

evnchn:
Interesting. I think thats a strong point for dropping `wait_for2`

However, I do insist that we need to real-world test this by serving a large website, so as to avoid any troubles. Can you try it? (Or you already tried)

evnchn:
To anyone that's reading, it doesn't mean NiceGUI will be less stable or reliable after the change. Rather, it will be as stable and reliable as the Python version you are using, as well as any other asynchronous library which uses the native `wait_for`

As we can see in the `wait_for2` documentation, generally speaking the behavior of the native `wait_for` is more predictable and stable, the more recent the Python version is. 

So, when in doubt, and also as a best practice, alwasy upgrade your Python version to the newest possible, for a peace of mind that NiceGUI, among other libraries, will work as smoothly as possible.

rodja:
We have just deployed a update which contains this PR on https://nicegui.io. Please let us know about any possibly related issues to the use of standard `asyncio.wait_for`.

evnchn:
Tested. `run.cpu_bound`, `run.io_bound`, `ui.run_javascript` (both fire-and-forget and await version), as well as `background_tasks.create` all work. Also the Outbox works, if not the page would fall apart. 

Safe to merge IMO.

PR #4845, opened 2025-06-07
Add `on_node_click` functionality to `ui.mermaid` element

thetableman:
### Motivation

The current `click` event handler for `ui.mermaid` relies on the global js `emitEvent` function and registering the event with `ui.on(...)`, see https://nicegui.io/documentation/mermaid#handle_click_events, but this differs to most other elements which allow specifying a callback as a parameter when the element is created. 

If applied, this PR will introduce a dedicated `on_node_click` parameter to the `ui.mermaid` element.

### Implementation

A new `on_node_click` parameter has been added to `ui.mermaid`. The user is still required to add a `click {nodeName} call nodeClick()` declaration to their markdown to signal which nodes are to trigger the callback. 

Ensuring that multiple Mermaid diagram instances on the same page all triggered the callback proved difficult. During testing it was found that:
- If multiple diagrams on the same page had overlapping node names this caused a double-trigger on the last diagram and none on the prior. 
- If the `nodeClick()` js function isn't ready when the chart is rendered it can break the diagram resulting in a blank element, refreshing the page resolves this.
- The last Mermaid diagram rendered to a page has `click` js but all prior diagrams were missing this.
- Unless passed as an argument to the js function in the markdown, Mermaid can't return the node id.
- These issues appear to all be at the Mermaid api level. 

To overcome these issues: 
- The markdown `content` is inspected and altered prior to sending to the javascript component to ensure that the `nodeClick()` function name is unique to each diagram, a timestamp is used to do this. Each diagram then has it's own caller attached to the instance without the user needing to manage this.
- The `content` needs to be updated with the unique function name prior to the first initial js load. If done afterwards it triggers a `content` refresh which breaks the diagram.
- Mermaid's own `click` js events, generated when a click reference is added to the markdown, are removed (if they exist) and replaced with a new `click` js event.
- The node id can be extracted from the nodes DOM element.

This is a fully working implementation but it is by no means perfect and there are definitely improvements & efficiencies to be found. 

### Progress

### Key Discussion Points

- Is there a better way to generate unique function names? Although this implementation works it doesn't feel right to alter the markdown secretly.
- Originally I used the elements 'id' to create the unique function names but this isn't available before `super()`. If the content is updated after `super()` it triggers an element refresh and this resulted in a broken diagram. Is there a better way to handle this situation?

### Test Example

```python
from nicegui import ui

ui.mermaid('''
    graph LR;
        A((Click Me));
        B((Or Click Me));
        A --> B;
        click A call nodeClick()
        click B call nodeClick()
    ''', on_node_click=lambda e: ui.notify(e.args))


ui.mermaid('''
    graph LR;
        A((Click Me 2));
        B((Or Click Me 2));
        A --> B;
        click A call nodeClick()
        click B call nodeClick()
    ''', on_node_click=lambda e: ui.notify(e.args))

ui.run()
```

### DIFF
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -9,7 +9,17 @@ export default {
     this.initialize();
     this.update(this.content);
   },
+  beforeUnmount() {
+    this.clearHandler();
+  },
   methods: {
+    getHandlerName() {
+      return `nodeClick_${this.clickInstance}`;
+    },
+    clearHandler() {
+      const handlerName = this.getHandlerName();
+      if (window[handlerName]) delete window[handlerName];
+    },
     initialize() {
       try {
         mermaid.initialize(this.config || {});
@@ -25,6 +35,22 @@ export default {
         const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", content);
         this.$el.innerHTML = svg;
         bindFunctions?.(this.$el);
+
+        if (this.clickInstance) {
+          const handlerName = this.getHandlerName();
+          window[handlerName] = (nodeId, nodeText) => {
+            this.$emit("nodeClick", {
+              node: this.getNodeName(nodeId),
+              nodeId,
+              nodeText,
+            });
+          };
+
+          this.$nextTick(() => {
+            this.attachClickHandlers(this.getHandlerName());
+          });
+        };
+
       } catch (error) {
         const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", "error");
         this.$el.innerHTML = svg;
@@ -34,9 +60,32 @@ export default {
         this.$emit("error", mermaidErrorFormat);
       }
     },
+    attachClickHandlers(handlerName) {
+      const clickables = this.$el.querySelectorAll("g.node.clickable");
+      clickables.forEach(node => {
+        node.removeAttribute("onclick");
+        node.onclick = null;
+
+        const newNode = node.cloneNode(true);
+        node.parentNode.replaceChild(newNode, node);
+
+        const nodeText = newNode.textContent.trim();
+        newNode.addEventListener("click", () => {
+          window[handlerName](newNode.id, nodeText);
+        });
+      });
+    },
+    getNodeName(domId) {
+      if (!domId) return undefined;
+      const parts = domId.split("-");
+      if (parts.length >= 3) return parts.slice(1, -1).join("-");
+      if (parts.length === 2) return parts[1];
+      return domId;
+    },
   },
   props: {
     config: Object,
     content: String,
+    clickInstance: Number,
   },
 };
--- a/nicegui/elements/mermaid.py
+++ b/nicegui/elements/mermaid.py
@@ -1,5 +1,8 @@
-from typing import Dict, Optional
+import re
+from datetime import datetime
+from typing import Dict, Match, Optional
 
+from ..events import GenericEventArguments, Handler
 from .mixins.content_element import ContentElement
 
 
@@ -11,26 +14,47 @@ class Mermaid(ContentElement,
               ]):
     CONTENT_PROP = 'content'
 
-    def __init__(self, content: str, config: Optional[Dict] = None) -> None:
+    def __init__(self, content: str, config: Optional[Dict] = None, on_node_click: Optional[Handler[GenericEventArguments]] = None) -> None:
         """Mermaid Diagrams
 
-        Renders diagrams and charts written in the Markdown-inspired `Mermaid <https://mermaid.js.org/>`_ language.
-        The mermaid syntax can also be used inside Markdown elements by providing the extension string 'mermaid' to the ``ui.markdown`` element.
+            Renders diagrams and charts written in the Markdown-inspired `Mermaid <https://mermaid.js.org/>`_ language.
+            The mermaid syntax can also be used inside Markdown elements by providing the extension string 'mermaid' to the ``ui.markdown`` element.
 
-        The optional configuration dictionary is passed directly to mermaid before the first diagram is rendered.
-        This can be used to set such options as
+            The optional configuration dictionary is passed directly to mermaid before the first diagram is rendered.
+            This can be used to set such options as
 
-            ``{'securityLevel': 'loose', ...}`` - allow running JavaScript when a node is clicked
-            ``{'logLevel': 'info', ...}`` - log debug info to the console
+                ``{'securityLevel': 'loose', ...}`` - allow running JavaScript when a node is clicked, applied automatically when `on_node_clicked` is specified
+                ``{'logLevel': 'info', ...}`` - log debug info to the console
 
-        Refer to the Mermaid documentation for the ``mermaid.initialize()`` method for a full list of options.
+            Refer to the Mermaid documentation for the ``mermaid.initialize()`` method for a full list of options.
 
-        :param content: the Mermaid content to be displayed
-        :param config: configuration dictionary to be passed to ``mermaid.initialize()``
-        """
+            :param content: the Mermaid content to be displayed
+            :param config: configuration dictionary to be passed to ``mermaid.initialize()``
+            :param on_node_click: callback that is invoked when a node is clicked, applies ``{'securityLevel': 'loose'}`` automatically
+            """
+        content, timestamp = self._update_click_handler(content)
+
         super().__init__(content=content)
-        self._props['config'] = config
+
+        self._props['config'] = config or {}
+
+        if on_node_click:
+            self.on('nodeClick', on_node_click)
+            self._props['config']['securityLevel'] = 'loose'
+            self._props['clickInstance'] = timestamp
 
     def _handle_content_change(self, content: str) -> None:
         self._props[self.CONTENT_PROP] = content.strip()
         self.run_method('update', content.strip())
+
+    def _update_click_handler(self, content: str) -> str:
+        pattern = r'click\s+(\w+)\s+(?:call\s+)?nodeClick(\([^)]*\))?'
+
+        dt = datetime.now()
+        timestamp = int(dt.timestamp())
+
+        def repl(m: Match[str]) -> str:
+            node = m.group(1)
+            return f'click {node} call nodeClick_c{timestamp}()'
+
+        return re.sub(pattern, repl, content), timestamp
--- a/nicegui/elements/mermaid.py
+++ b/nicegui/elements/mermaid.py
@@ -1,6 +1,6 @@
 import re
 from datetime import datetime
-from typing import Dict, Match, Optional
+from typing import Dict, Match, Optional, Tuple
 
 from ..events import GenericEventArguments, Handler
 from .mixins.content_element import ContentElement
@@ -47,7 +47,7 @@ class Mermaid(ContentElement,
         self._props[self.CONTENT_PROP] = content.strip()
         self.run_method('update', content.strip())
 
-    def _update_click_handler(self, content: str) -> str:
+    def _update_click_handler(self, content: str) -> Tuple[str, int]:
         pattern = r'click\s+(\w+)\s+(?:call\s+)?nodeClick(\([^)]*\))?'
 
         dt = datetime.now()
--- a/tests/test_mermaid.py
+++ b/tests/test_mermaid.py
@@ -1,4 +1,3 @@
-import pytest
 from selenium.webdriver.common.by import By
 
 from nicegui import ui
@@ -82,18 +81,19 @@ def test_error(screen: Screen):
     screen.should_contain('Parse error on line 3')
 
 
-@pytest.mark.parametrize('security_level', ['loose', 'strict'])
-def test_click_mermaid_node(security_level: str, screen: Screen):
+def test_click_mermaid_node(screen: Screen):
     ui.mermaid('''
         flowchart TD;
-            A;
-            click A call document.write("Success")
-    ''', config={'securityLevel': security_level})
+            A[Node A];
+            B[Node B];
+            click A call nodeClick()
+            click B call nodeClick()
+    ''', on_node_click=lambda e: label.set_text(str(e.args)))
+    label = ui.label('')
 
     screen.open('/')
-    screen.click('A')
-    screen.wait(0.5)
-    if security_level == 'loose':
-        screen.should_contain('Success')
-    else:
-        screen.should_not_contain('Success')
+    screen.click('Node A')
+    assert 'A' in label.text
+
+    screen.click('Node B')
+    assert 'Node B' in label.text
--- a/website/documentation/content/mermaid_documentation.py
+++ b/website/documentation/content/mermaid_documentation.py
@@ -14,17 +14,19 @@ def main_demo() -> None:
     list(ui.context.client.elements.values())[-1].props['config'] = {'securityLevel': 'loose'}  # HACK: for click_demo
 
 
-@doc.demo('Handle click events', '''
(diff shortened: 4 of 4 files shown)


evnchn:
I think that `click SOMENODE call nodeClick()` is possibly a bad API design because:

- User suffers from low flexibility, as they cannot have 1 event handler attached to 2 mermaid node, at least not easily
- Library maintainers suffers from high maintenance burden, with us having to inspect the markdown content, which may break if passed input which slips between our understanding of Mermaid syntax intricacies and the library's code interpretation
- NiceGUI philosophy of a lightweight and thin layer around existing libraries falls apart, as we're doing some magic in `ui.mermaid`

I would counter-propose an alias function `emit(...)` which corresponds to `this.$emit(...)`, as a convenience method of emitting events exclusive to the Mermaid element only. It's similar to what we have done in #4689, so it's somewhat consistent as well. 


What do you think?

thetableman:
Thanks for the feedback @evnchn.

> I think that click SOMENODE call nodeClick() is possibly a bad API design

Ha ha... I totally agree... but it's not my idea, this is the official way you register a click event in Mermaid. They aren't even that consistent with it either, in `graph` diagrams you can just do `click SOMENODE func` with no `call` or `()` but other diagrams like `sequence` require the full reference. Some diagrams don't officially support click events at all.

> ...as they cannot have 1 event handler attached to 2 mermaid node...

Not sure I understand what you mean here. If using the official Mermaid click events then every node can have it's own event handler or you can add the same event handler to every node.

> Library maintainers suffers from high maintenance burden, with us having to inspect the markdown content, which may break if passed input which slips between our understanding of Mermaid syntax intricacies and the library's code interpretation.

Completely agree! I don't like it either but after days of going around in circles it always came back to that requirement. Also because, as noted above, there's no consistency in how Mermaid wants those commands.

> ...as we're doing some magic in ui.mermaid

Agree again.

> I would counter-propose an alias function emit(...) which corresponds to this.$emit(...)

I'll take a look. 

The whole custom function name and inspect & alter the markdown is all just because I (initially) wanted the ability for the Mermaid markdown to be used as intended when a click is needed - that you use the `click SOMENODE call FUNC()` pattern as per their documentation. I (initially) wanted this because it meant the user could control which nodes had a click event and which didn't.

I keep saying "initially" because perhaps that is also unnecessary - you don' get that with many of the other elements in NiceGUI. You can't easily say which chart points/bars in `ui.echarts` are or are not clickable. 

So, based on your feedback, I'm now thinking:
- Dump the requirement for Mermaid `click SOMENODE call FUNC()` in the markdown. Just add the callback function to the `on_node_click` parameter.
- Drop the inspect & alter markdown to make the functions unique.
- Send a `clickable` prop over to js to inform it that `on_node_click` is specified, this will add the correct cursor styling to the nodes
- Set up the emit for all the nodes, users can filter the responses for the ones they won't just like a clickable bar/point in `ui.echarts`.

Thoughts?

evnchn:
@thetableman Some discovery

> nodeId is the id of the node
> callback is the name of a javascript function defined on the page displaying the graph, the function will be called with the nodeId as parameter.

Reference: https://mermaid.js.org/syntax/flowchart.html#:~:text=nodeId%20is%20the,nodeId%20as%20parameter.

So, possibly we just need to expose `emit = (e) => this.$emit('nodeClick', e)` and it'd work, no need for the inspection of Mermaid diagram content, and the hacks on top. 

Example here: 

```
flowchart LR
    A-->B
    B-->C
    C-->D
    D-->E
    click A alert
```

image

thetableman:
@evnchn, yes, you can declare the function like that as I said above, and yes, it does return the node id. However, as I mentioned in the PR intro, if two diagrams on the same page have the same node id, then they both trigger.

thetableman:
I'm not near a computer to run a test, but wouldn't this cause a double trigger too. Once from Mermaids own process and once from this one?

> possibly we just need to expose emit = (e) => this.$emit('nodeClick', e) and it'd work,

thetableman:
So I've had a play with the `js_handler` of a `.on(...)` as discussed in #4689 and did get some success. Yes the emit fires and yes there are `args` returned, however, the node id is not one of the `args` but I fully appreciate this is probably somehting I doing wrong from lack of full understanding.

One issue, as noted previously, persists in this soltuion though - when two diagrams are on the same page the second diagram has `click` events and the first doesn't. This evident from the DOM where the `click` event is clearly missing from the first diagram nodes, and from the fact that the `alert` doesn't trigger on the first diagram but does on the second.

```py
from nicegui import ui

ui.mermaid('''
    graph LR;
        A((Click Me));
        B((Or Click Me));
        A --> B;
        click A alert
        click B alert
    ''', config={'securityLevel': 'loose'}).on('click', lambda e: ui.notify(f'click: {e.args}'), js_handler="(e) => emit('nodeClick', e)")


ui.mermaid('''
    graph LR;
        C((Click Me 2));
        D((Or Click Me 2));
        C --> D;
        click C alert
        click D alert
    ''', config={'securityLevel': 'loose'}).on('click', lambda e: ui.notify(f'click: {e.args}'), js_handler="(e) => emit('nodeClick', e)")

ui.run()
```

image

I'm not entirely sure why this is but throughout my testing this has been a consistent issue and this is exactly why I looped the nodes and added my own `click` event, because it was the only guaranteed way to ensure all diagrams on a page had them.

evnchn:
> One issue, as noted previously, persists in this soltuion though - when two diagrams are on the same page the second diagram has click events and the first doesn't. This evident from the DOM where the click event is clearly missing from the first diagram nodes, and from the fact that the alert doesn't trigger on the first diagram but does on the second.

That could classify as a bug outside of this PR. Can you quickly check if #4692 could have introduced this regression? Thanks (I don't want to have messed up the Mermaid in a PR which I made...)

thetableman:
> Can you quickly check if https://github.com/zauberzeug/nicegui/pull/4692 could have introduced this regression?

When I roll back the js to the version prior to that introduced in #4692 then both diagrams have the `click` event on the nodes, when I re-instate the changes made in the PR, the first diagram loses it's `click` events. 

This strongly suggests the PR introduced this issue but I'm not sure why or how. The biggest difference between the two versions seems to be using `render` with `bindFunctions` rather than `run`.

NiceGUI dependencies says we're using Mermaid 11+ (https://github.com/zauberzeug/nicegui/blob/main/DEPENDENCIES.md) and the Mermaid documentation suggests that, from 10+, using `run` "...is the preferred way of handling more complex integration" (https://mermaid.js.org/config/usage.html#using-mermaid-run). Equally, though, there are clear instructions (without deprecation warnings) that the `render` process is also okay as was noted in the links provided in PR #4692

I've queried the changes with ChatGPT but it strugged to explain and identify what the cause is, or, more importantly, find a solution. But, what was ascertained, after going around in circles for a bit, is that moving away from using Mermaid's `click` events to defining our own may be the best way to ensure all diagrams trigger correctly.

evnchn:
I think you should file an issue reporting the bug, since it's a real bug and not related to this PR strictly. 

However I may be a bit slow at resolving it, partly since I'm busy, and also since emitEvent solution works.

thetableman:
@evnchn see #4852 

> and also since emitEvent solution works.

The `emitEvent` solution doesn't work for the first diagram, it can't work because the `click` event is missing.

evnchn:
I see. I marked #4852 and #4853 as a medium in terms of importance as a result. Let's wait for that PR to be merged before working on this one?

evnchn:
#4853 merged. 

Resolve merge conflicts, and you should be good to go. 

I'm still thinking about the approach in this PR, though...

thetableman:
I think we need to clearly define and agree on how the user designates that `click` events are needed in the markdown first, they all have their pros and cons.

#### Using `click <nodeName> <eventHandler>` syntax
- Pro: This always returns the nodes name in the response
- Pro: The results are consistent, the same arguments are returned each time
- Con: No other parameters can be set
- Con: Only works with some diagrams, causing a breaking error in others (by design of the api)
- Con: Further testing required to ensure that this doesn't cause double-triggers if multiple diagrams share similar node names
- Con: Inconsistent definition in markdown may cause confusion when using - eg should the user use the below syntax or this syntax for a diagram

#### Using `click <nodeName> call <eventHandler>(<params>)` syntax
- Pro: Works with all diagram types that permit `click` events
- Pro: Allows user to pass any parameters to the handler for return
- Con: Relies on user to define parameters to be returned as arguments, if none specified then none are returned
- Con: Creates inconsistent arguments as the handler parameters don't necessarily include node name
- Con: Further testing required to ensure that this doesn't cause double-triggers if multiple diagrams share similar node names
- Con: Inconsistent definition in markdown may cause confusion when using - eg should the user use the above syntax or this syntax for a diagram

#### No `click` statement included in the markdown
- Pro: No additional markdown required to enable `click` events, they are added when a callback is added to the `ui.mermaid` element.
- Pro: Consistent arguments returned, an `event` is added to each node by the NiceGUI api
- Pro: No confusion over syntax in the markdown
- Pro: Potential to add `click` events to diagram types that don't support them natively
- Con: No other parameters can be set
- Con: Moves away from Mermaid api definitions
- Con: Subject to potential breaking by documented & un-document changes to the Mermaid api

@evnchn Feel free to add pros/cons where you see fit.

Personally, despite the cons, I prefer the last option now - yes, moving away from my initial proposal. The lack of consistency in the mermaid api `click` syntax, coupled with the lack of consistency in the arguments from the event (because they need to be user defined), irritate me.

evnchn:
I think I prefer that we define `emit` in the Mermaid component, so that the user can do choice 1 and choice 2. 

- For "Con: No other parameters can be set" and "Con: Relies on user to define parameters to be returned as arguments, if none specified then none are returned", I can only say, you did it to yourself (Mermaid held partly responsible due to its API design)
- For "Con: Only works with some diagrams, causing a breaking error in others (by design of the api)" and "Con: Inconsistent definition in markdown may cause confusion when using - eg should the user use the below syntax or this syntax for a diagram", it is Mermaid fault. We can't be responsible to cover-up Mermaid API's deficiencies, right? 
- We need to work hard against "Con: Further testing required to ensure that this doesn't cause double-triggers if multiple diagrams share similar node names". The third choice won't help you escape this workload.

thetableman:
> I think I prefer that we define emit in the Mermaid component, so that the user can do choice 1 and choice 2.

We might need to defer to maintainers for guidance here then?

> I can only say, you did it to yourself

Presumably you mean the user writing the `click` events in the markdown? The reason why I think this is a con is that in most  other NiceGUI elements the arguments returned from an event are consistent (not with other elements, but within themselves) and don't require a user to define exactly what they want returned.

> ...it is Mermaid fault. We can't be responsible to cover-up Mermaid API's deficiencies, right?

Responsible or Cover-up? No, not the suggestion at all. Finding solutions to overcome difficulties in underlying apis is literally why projects like NiceGUI exist and are popular - never forget who the target market is here! If we didn't want to simplify and smooth over the complexities of Mermaid then we wouldn't have a convenient `ui.mermaid` (or many other elements) to do a lot of the hard work.

> ... ensure that this doesn't cause double-triggers... The third choice won't help you escape this workload.

In fact, not that much of a workload at all! After dropping the idea that I needed to rely on the Mermaid markdown to initiate the `click` events the whole project became a lot simpler (and smaller). Whilst duplicate node names can exist in different charts, when they are populated to the DOM they do have individual ids, thus making them unique. It's therefore quite straight forward to attach `click` events to each node knowing that it's unlikely to clash. 

I don't offer these responses as further weight to my own preference - I've already made my opinions clear - but rather to ensure we all understand the specifics of the differences in approach. 

I'm also not against the suggestion that we define `emit` as you have noted, that's a nice clean approach. It's the way we decide the nodes are clickable that I think needs to be simpler. Let's not forget the comment from earlier...

> I think that click SOMENODE call nodeClick() is possibly a bad API design

... it is, but here's the cool part, we don't have to use it ;)

evnchn:
Well say, the situation seems more nuanced than what I thought to be (I thought it was cut-and-dry). I think opinion from the maintainers is helpful here.

evnchn:
@thetableman I see that you have mentioned that only some diagrams support `click <nodeName> <eventHandler>`, while only some other diagrams support `click <nodeName> call <eventHandler>(<params>)`

Can you tell me:

- Which notable diagrams support which?
- Generally, do the diagrams support one or the other? 
  - Or do some diagrams support both? 
  - Or do some diagrams support none? 

It'll help in my investigation towards the API design as well as maintainers reviewing this PR against a variety of diagram types.

falkoschindler:
> I think opinion from the maintainers is helpful here.

I'm struggling to imagine the final usage of each option (click, click-call, no click). And I'm probably missing some of the plenty pros and cons you collected. But intuitively it feels sketchy to programmatically add click events (you mean editing the Mermaid code?), trying to iron out Mermaids inconsistent event API. @evnchn You often mention NiceGUI providing a shallow layer around existing libraries. By taking over event registrations it feels like we're doing a little much, which could backfire in the long run. And if Mermaid's event arguments are inconsistent, we can forward them as a generic dictionary, like for plain HTML events.

But maybe one of you can compare the three options from the user's perspective - how does a minimal example look like, when would an event be fired, what would be contained in its arguments.

evnchn:
> you mean editing the Mermaid code?

Yes this PR is doing that right now. And it's very sketchy IMO. More sketchy than what I could tolerant, even though I do write some sketchy code from time to time. Check around `content, timestamp = self._update_click_handler(content)`

> if Mermaid's event arguments are inconsistent, we can forward them as a generic dictionary

Agreed. We can probably get away with `ValueChangeEventArguments` without explaining, or do the following for Mermaid (though I think it's a bit too much)

```py
@dataclass(**KWONLY_SLOTS)
class PossiblyANodeNameOrItMayBeYourOwnInputChangeEventArguments(UiEventArguments):
    value: Any
```

But you get the point 😅

I want to take a shot at this problem using the light-weight `emit` approach, so that we can compare that approach to this PR's approach, and see for ourselves which is better. I can agree that "struggling to imagine the final usage" is the major pain point in understanding the 3 proposed directions, and we need some code to get our minds spinning properly

thetableman:
Lot's to unpack here!

> Yes this PR is doing that right now. And it's very sketchy IMO

Yes, the initial PR did suggest "sketchy" methods but if you refer to pretty much all subsequent comments, I made it clear that alternatives were available and that I was in favour of moving towards them.

> But intuitively it feels sketchy to programmatically add click events (you mean editing the Mermaid code?)

It is "sketchy", I noted this early on and suggested alternatives.

> I'm struggling to imagine the final usage of each option (click, click-call, no click). 

It's simple, either the user adds `click <nodeName> <function>` or `click <nodeName> call <function>()` to the markdown for items 1 and 2 respectively, or they don't in number 3. All 3 suggestions are non-"sketchy" proposals, none of them require altering the markdown. In fact this whole comment was about the pros/cons of whether the user needed to declare a`click` event in the markdown at all - there was no suggestion of editing the markdown in any of those discussion points.

I think it's best we close this PR opening up the opportunity for others to find a fresh solution.

falkoschindler:
Sorry, @thetableman, I realize I joined this thread a bit late, and with the discussion already quite long, it's been hard to fully catch up on all the context. I've read "despite the cons, I prefer the last option now", but apparently couldn't visualize how it would be done. Maybe the implementation of this approach is easier than thought.
Anyway, a fresh start might help to focus on the preferred approach and avoid confusion with other alternatives.

thetableman:
@falkoschindler It is a long thread to follow for sure!

I think it's just reached the point where I need to evaluate the time spent testing different approaches and finding alternative methods - which has all been at the detriment of my actual work and family time.

evnchn:
Got #4861 now. See if that works better. 

Please don't stress yourself out 💪🏻

PR #4849, opened 2025-06-09
UserInteraction.click(): Handle radio elements

barrywhart:
### Motivation

Currently, `UserInteraction.click()` does not support `ui.radio` elements. This PR updates it so radio elements work similarly to `ui.select` elements.

### Implementation

It adds a new section of logic to `UserInteraction.click()` for radio elements.

I have not added documentation for this case, but it's so similar to `ui.select` that perhaps no additional documentation is needed. (But I'm happy to add documentation if that would be useful.)

### Progress


barrywhart:
Thanks!

PR #4853, opened 2025-06-09
Reintroduce queue for Mermaid, since parallel rendering = no onclick

### LINKED ISSUE #4852: Mermaid `click` events missing if multiple diagrams are rendered
### First Check

### Example Code

```python
from nicegui import ui

ui.mermaid('''
    graph LR;
        A --> B;
        click A alert
        click B alert

ui.mermaid('''
    graph LR;
        C --> D;
        click C alert
        click D alert

ui.run()
```

### Description

When two Mermaid diagrams with `click` events defined in the markdown are rendered on the same page, the first diagram is missing the `click` event in the DOM, the second is created correctly.

This results in the first diagram having no click action while the second works as expected, triggering the `alert`.

The expected outcome is that both diagrams render with `click` events and in the example provided all nodes trigger an `alert`. It is expected that each diagram be a separate instance and as such not impact the settings of the other.

### NiceGUI Version

2.19.0.dev0

### Python Version

3.12.5

### Browser

Firefox, Chrome

### Operating System

Windows

### Additional Context

This issue was identified during the drafting of #4845 

I've tested with the implementation of `mermaid.js` prior to PR #4692 and the issue was not present - two diagrams with `click` events rendered correctly.

Example of diagram 1, node A missing the `click` event:
Image

Example of diagram 2, node D with `click` event:
Image

evnchn:
### Motivation

This PR fix #4852, by reintroducing the rendering queue that we dropped in #4692 because of "unknown reasons". 

### Implementation

- Copy over the `let is_running = false;` and `const queue = [];` from prior to #4692 
- Let GitHub copilot rewrite the rest of the rendering code to respect the queue

### Progress

I think we should add Pytest to avoid accidentally regressing this in the future, but Mermaid pytest is _really_ not my strong suit, and I'd like to do what I am best at - rapid fixes. 

Also I am not sure if we should hold back the fix, just because of missing pytest. 

### Test code

Stolen from #4852

```py
from nicegui import ui

ui.mermaid('''
    graph LR;
        A --> B;
        click A alert
        click B alert

ui.mermaid('''
    graph LR;
        C --> D;
        click C alert
        click D alert

ui.run()
```

Before: 

{2D867831-AA7B-4CD3-8FF7-B0D681EE0F5C}

-> No `click`

After: 

{E35456E5-6B65-4506-8DC9-16FBE7B78FBE}

-> Yes `click`

### DIFF
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -1,5 +1,8 @@
 import mermaid from "mermaid";
 
+let is_running = false;
+const queue = [];
+
 export default {
   template: `<div></div>`,
   data: () => ({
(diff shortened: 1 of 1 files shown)


evnchn:
https://github.com/zauberzeug/nicegui/pull/4853/commits/44f3a0a1664f3b2ea82363a4eff525dc3c4a91c4 was simple: Add on some Mermaid diagram before and after the one to be tested, such that it will break if our `ui.mermaid` respects only the last Mermaid's click events.

PR #1287, opened 2023-07-28
Move Quasar config to Python

rodja:
By making the Quasar config available through `globals.quasar_config` it can be configured/changed by the app. This can for example be used to disable the loading bar:

```py
from nicegui import globals as nicegui_globals
...

nicegui_globals.quasar_config.["loadingBar"]["skipHijack"] = False

...
```

PR #4859, opened 2025-06-11
Let configure UI framework other than Quasar

evnchn:
### Motivation

In direct response to #4858 and #3818, this PR attempts at decoupling NiceGUI from Quasar at the browser-side Vue app configuration level, such that we can then assess which UI frameworks can work with NiceGUI, and then take a shot at creating native elements in Python. 

We are beginning a new project, preliminarily named NFM: NiceGUI Framework Multiverse

### Implementation

- Rip out Quasar in NiceGUI's `createApp`
- Introduce the concept of `ui_framework_config_script`, which defaults to `DEFAULT_APP_CONFIGURATOR`, that we can override to not interact with `Quasar` but another UI framework in `app.use` and other configuration and bootstrapping
- Set the above by `app.configure_vue_ui_framework`, which could be userspace code (if they urgently want another UI framework), another library's code, or even in an optional package of NiceGUI library

### Progress

### Results

I don't have a lot of time (it's 4am around where I am right now!) but I did get Element Plus working: 

```py
from nicegui import ui, app

ui.add_body_html('''
<link rel="stylesheet" href="//unpkg.com/element-plus/dist/index.css" />
<script defer src="//unpkg.com/element-plus"></script>
''', shared=True)

app.configure_vue_ui_framework('''
app.use(ElementPlus);
''')

with ui.element('el-button'):
    ui.html('I am a button from Element Plus')

ui.run()
```

{93B8D525-9EB4-4F81-8CBE-0EC54C3E5A32}

-> No `el-button`, instead we have `button`, since Element Plus loaded in properly. 

**Your mileage may vary with other UI frameworks!!!**

My intention: Merge this, since this doesn't break the existing code, and then we can work from here on out.

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -18,7 +18,7 @@ from ..observables import ObservableSet
 from ..server import Server
 from ..staticfiles import CacheControlledStaticFiles
 from ..storage import Storage
-from .app_config import AppConfig
+from .app_config import DEFAULT_APP_CONFIGURATOR, AppConfig
 from .range_response import get_range_response
 
 
@@ -300,3 +300,14 @@ class App(FastAPI):
         for client in Client.instances.values():
             if client.page.path == path:
                 yield client
+
+    def configure_vue_ui_framework(self, config_script: str = DEFAULT_APP_CONFIGURATOR):
+        """Configure the UI framework used with the Vue app.
+        This is used to set up the UI framework (e.g. Quasar) with the provided configuration.
+
+        :param config_script: JavaScript code to configure the UI framework (default: DEFAULT_APP_CONFIGURATOR)
+
+        Note: the config_script should be a format string of JavaScript code that will be executed in the browser.
+        It will be called with `.format(config=json.dumps({...})` to populate the configuration.
+        """
+        self.config.ui_framework_config_script = config_script
--- a/nicegui/app/app_config.py
+++ b/nicegui/app/app_config.py
@@ -5,6 +5,12 @@ from typing import Dict, List, Literal, Optional, Union
 from ..dataclasses import KWONLY_SLOTS
 from ..language import Language
 
+DEFAULT_APP_CONFIGURATOR = '''
+app.use(Quasar, {{config: {config}}});
+Quasar.lang.set(Quasar.lang[language.replace('-', '')]);
+Quasar.Dark.set(dark === null ? "auto" : dark);
+'''
+
 
 @dataclass(**KWONLY_SLOTS)
 class AppConfig:
@@ -23,6 +29,7 @@ class AppConfig:
                 'skipHijack': False,
             },
         })
+    ui_framework_config_script: str = field(default_factory=lambda: DEFAULT_APP_CONFIGURATOR)
 
     reload: bool = field(init=False)
     title: str = field(init=False)
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -155,7 +155,7 @@ class Client:
                 'imports': json.dumps(imports),
                 'js_imports': '\n'.join(js_imports),
                 'js_imports_urls': js_imports_urls,
-                'quasar_config': json.dumps(core.app.config.quasar_config),
+                'ui_framework_config_script': core.app.config.ui_framework_config_script.format(config=json.dumps(core.app.config.quasar_config)),
                 'title': self.resolve_title(),
                 'viewport': self.page.resolve_viewport(),
                 'favicon_url': get_favicon_url(self.page, prefix),
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -436,8 +436,6 @@ function createApp(elements, options) {
         });
       }
     },
-  }).use(Quasar, {
-    config: options.quasarConfig,
   }));
 }
 
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -51,15 +51,14 @@
         query: {{ socket_io_js_query_params | safe }},
         extraHeaders: {{ socket_io_js_extra_headers | safe }},
         transports: {{ socket_io_js_transports | safe }},
-        quasarConfig: {{ quasar_config | safe }},
       });
+      const dark = {{ dark }};
+      const language = "{{ language }}";
+      {{ ui_framework_config_script | safe }}
 
       {{ js_imports | safe }}
       {{ vue_scripts | safe }}
 
-      const dark = {{ dark }};
-      Quasar.lang.set(Quasar.lang["{{ language }}".replace('-', '')]);
-      Quasar.Dark.set(dark === None ? "auto" : dark);
       {% if tailwind %}
       if (dark !== None) tailwind.config.darkMode = "class";
       if (dark === True) document.body.classList.add("dark");


rodja:
Thanks @evnchn. Your PR rate is really astonishing. I like the idea of making the UI framework configurable and have not imagined it to be to such a small diff.

rodja:
In 317748b I made some naming suggestions. It was easier to do in the code directly instead of listing it in a comment here. But of corse it is open for discussion.

Also I think that we can remove the `AppConfig.quasar_config` because that can now be put directly in `AppConfig.DEFAULT_VUE_CONFIG`.

evnchn:
> we can remove the `AppConfig.quasar_config` because that can now be put directly in `AppConfig.DEFAULT_VUE_CONFIG`

Actually I was contemplating that, and my solution beats around changing that. 

`quasar_config` is outdated naming, and best to be absorbed into the string, but I wonder if some existing code writes to `AppConfig.quasar_config` directly, and could be considered a breaking change. 

At the end of the day, it depends on whether you classify `quasar_config` as public API.

evnchn:
Technically not public API, since no examples use it, and it's undocumented. 

But still, I hesitated on removing it. I'm willing to do it with your confirmation, though.

rodja:
I really would like to drop it here and now. But on the caution side, let's mark it as depreciated and remove it in 3.0.

falkoschindler:
The `quasar_config` has been moved to Python in PR #1287 by @rodja and has been mentioned in #1333. Unfortunately it is undocumented, but I would consider it public API.

nicegui/app/app_config.py
```diff
@@ -23,6 +29,7 @@ class AppConfig:
                 'skipHijack': False,
             },
         })
+    vue_config: str = field(default_factory=lambda: DEFAULT_VUE_CONFIG)
```
falkoschindler: @evnchn Do you think we need the `default_factory`? Or can we simply assign the default value?
```suggestion
    vue_config: str = DEFAULT_VUE_CONFIG
```

falkoschindler: About deprecating the `quasar_config`:

1. I think it's difficult to print an warning whenever the field is used. We might just add a `# DEPRECATED` marker and inform about the breaking change in 3.0.
2. But do we really want to drop it? Assuming the majority of NiceGUI will remain based on Quasar (at least for 3.x and maybe 4.x), it is a very handy field. Instead of writing JS code, you can simply add or modify dictionary values. Therefore I'd prefer keeping it for now (without deprecation).

nicegui/app/app.py
```diff
@@ -300,3 +301,13 @@ def clients(path: str) -> Iterator[Client]:
         for client in Client.instances.values():
             if client.page.path == path:
                 yield client
+
+    def set_vue_config(self, script: str = DEFAULT_VUE_CONFIG) -> None:
```
falkoschindler: Do we really need this method? Depending on how official the new config option is meant to be, we could simply access `config.vue_config` directly, reducing public API which is hard to change in the future. And since it's just a one-liner, the method doesn't add much value.

nicegui/app/app_config.py
evnchn: Asked GitHub Copilot. So long as `DEFAULT_VUE_CONFIG` is immutable, we no need to use the `default_factory`, whose role is to handle mutable defaults so that we don't share one mutable instance (or something)

The more I learn every day!

evnchn: TL-DR: Do rename since the name will be inaccurate, but don't drop it since Python-land configurability is still useful. 

1. I agree to renaming `quasar_config` to something more generic (`config`? `vue_app_config`? `ui_framework_config`?). So, deprecating `quasar_config` and moving people to the new naming. 
  a. `# DEPRECATED` marker could do, and we'd just make the switch in 3.0 to the new name. 
  b. But it could possible to enable the use of the new name in 2.X so as to ease transition? Although it could be a pain to keep 2 members synchronized? 
2. Thinking about it for longer, dropping `quasar_config` may not be a good idea. Many UI frameworks (including Quasar) have some options you can pass for initializing it. Those options are useful to be able to be modified by Python-land, without mutating the vue config script text. Therefore, I would tend to disagree with the approach outlined in https://github.com/zauberzeug/nicegui/pull/4859#issuecomment-2965165262. @rodja What do you think?

nicegui/app/app.py
evnchn: Good idea. Had more in mind for `set_vue_config` but it became a string-setter in the end...

Done in https://github.com/zauberzeug/nicegui/pull/4859/commits/0d480de36db44c9ee42b58799f8761100d49183f

evnchn:
> This PR seems to break `ui.run(dark=None)`

Fixed in https://github.com/zauberzeug/nicegui/pull/4859/commits/a0a163e2cdb2d41c9cdfa92f925b3b9880a9d1c5

I mean it's a trivial change: We're just shifting code around, and after this PR, the code should behave exactly the same, if you don't set `app.config.vue_config`

If I am not mistaken, this is not tested in our test suite, since we don't set `prefers-color-scheme` in our testing at all. It may be outside of the scope of this PR, but do you think we can cover that with a test?

evnchn:
It says tests are failing on my notifications. 

Wait for a bit.

nicegui/app/app_config.py
rodja: I still think that `quasar_config` can be dropped in 3.0. If someone wants to modify the config of Quasar, they can easily replace the entire `app_config.vue_config`.

evnchn: If we endorse everyone to write to `app.config.vue_config`, it would be difficult to ship changes to said config, since everyone would be overriding it. 

Think about what would happen, if we need `Quasar.init();` from Quasar from now on. We can't patch it if the user sets `app.config.vue_config` manually, and they'll have to update their code manually.

rodja: Hmmm... good argument. I'm ok with keeping it for now -- even though it feels a bit to complicated.

evnchn:
So (I think) we have agreed to, at least for now, not do anything about `quasar_config`

Renaming that to something more generic, `vue_ui_framework_config`, figure out some other mechanism for both customizable Vue Script and Python-land parameter passing, or dropping it altogether, we'll think about it at a later stage (a subsequent PR following this one, most probably)

Though, I did have a thought: can we drop the `.format`, if we `const vue_app_config = {{quasar_config}};` in the source code? Then our Vue Script can just use that variable in JS! @rodja what do you think?

@falkoschindler regarding mocking the ElementPlus button, 3 choice I have thought of:

1. In the Vue Script, read a boolean variable, which can be set using `ui.add_head_html` script tag, and we'll divert to load ElementPlus only on the special test page
2. We can just lie serve a plain Vue+ElementPlus app, can't we?
3. We can do with a screenshot?

evnchn:
https://github.com/zauberzeug/nicegui/pull/4859#issuecomment-2988357903

- Achieved config dict drop-in without `.format`, by setting `const vue_config` in JS-land
- Demo at `/element_plus` by a Vue config script with conditionally loads Element Plus or Quasar, based on a JS-land flag.

evnchn:
Congratulations on cloning the ElementPlus button style. 

However I just like to mention that this is super alpha-level, since if we drop Quasar, then the q-app container (iirc) all becomes invalid tags, and so we need to do quite a bit of rework in NiceGUI library to get basic stuff working, in addition to adapting the existing elements / creating a new set of elements.

rodja:
> However I just like to mention that this is super alpha-level, since if we drop Quasar, then the q-app container (iirc) all becomes invalid tags,

You are right, @evnchn. This PR only enables very brave and skilled devs to change some deep internals of NiceGUI. Good thing that it is marked as experimental in the docs. Although, as Falko just showed me, we could use `app.config.vue_config_script +=` in the demo which loads stuff in addition to Quasar and hence does not produce invalid tags. What do you think?

evnchn:
Interesting. All is good with me. 

1. `app.config.vue_config_script` is in fact the default. If we need the constant, we can `DEFAULT_VUE_CONFIG_SCRIPT = app.config.vue_config_script` right after importing NiceGUI. Inlining the constant is a good idea. 
2. Somehow I _thought_ that UI frameworks would be mutually exclusive to each other. I'm sure you must have done it already, but be sure to manually test the code in the demo, since we're not importing ElementPlus but mocking it in the documentation page. 
3. And if we can use both Quasar and some other UI frameworks, short of the obvious concern of bulkiness, that really makes the devwork of adapting the other UI framework easier. It's a great idea in general, and vibes with the general theme of this PR of "making NiceGUI work with another UI framework experimentally, while we figure out the details later"

dzjwwm:
There is a BUG in the use of vue_config_script here. I only added it once and opened the browser to look at the references in js twice
image

evnchn:
Oh. That's a bug. Are you using hot-reload? Try reload=False in `ui.run()` and see if it fixes things.

falkoschindler:
@dzjwwm Ah yes, that's probably because you're running in script mode (no `@page` or `root` function). In this case the main file is evaluated twice. But is it really a bug? I mean, does it hurt to call `app.use(...)` twice?

Anyway, I you think it's a bug, I recommend opening a new issue. Thanks!

PR #4861, opened 2025-06-12
Mermaid `on_node_click`, with `{handler}` drop-in approach

evnchn:
### Motivation

@thetableman, I feel sorry for pushing back against #4845, so I spent about 2 hours to code what I think could work to let Mermaid have node click functionality, with a minimally-sketchy code change. Notably, I used string format instead of regex, so the change NiceGUI is making should be obvious and understandable to everyone. 

### Implementation

If we desire to add `on_node_click` for Mermaid: 
- Setup function `mermaid_click_handler_SOMEUUID` (since `this` is not available), . 
- Instead of using regex, which are less predictable and a bit of a black box, I use string format (formatting against `{handler}`) and replace that with the actual function name

Shows up clearly in the IDE as well: 

{562E07C9-AB7D-46E6-8BD8-D01AB95B5D1D}

### Progress

_Pytest pending. Documentation pending. I'm tired TBH._ 

### Test script

```py
from nicegui import ui


@ui.page('/')
def main_page():
    ui.mermaid('''
        graph LR;
            A --> B;
            B --> C;
            click A {handler}
            click B call {handler}()
            click C call {handler}("C", "C some args")


ui.run()
```

### Result

{BA0717C3-3735-48B0-9FC0-2ACCE953FEA3}

### DIFF
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -20,6 +20,11 @@ export default {
         console.error(error);
         this.$emit("error", error);
       }
+      if (this.function_name) {
+        window[this.function_name] = (node, param) => {
+          this.$emit("nodeClicked", {node: node, param: param});
+        }
+      }
     },
     async update(content) {
       if (this.last_content === content) return;
@@ -48,5 +53,6 @@ export default {
   props: {
     config: Object,
     content: String,
+    function_name: String,
   },
 };
--- a/nicegui/elements/mermaid.py
+++ b/nicegui/elements/mermaid.py
@@ -1,5 +1,8 @@
+import uuid
 from typing import Dict, Optional
 
+from ..events import GenericEventArguments, Handler
+from ..logging import log
 from .mixins.content_element import ContentElement
 
 
@@ -11,7 +14,7 @@ class Mermaid(ContentElement,
               ]):
     CONTENT_PROP = 'content'
 
-    def __init__(self, content: str, config: Optional[Dict] = None) -> None:
+    def __init__(self, content: str, config: Optional[Dict] = None, *, on_node_click: Optional[Handler[GenericEventArguments]] = None) -> None:
         """Mermaid Diagrams
 
         Renders diagrams and charts written in the Markdown-inspired `Mermaid <https://mermaid.js.org/>`_ language.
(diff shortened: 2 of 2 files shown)


evnchn:
@falkoschindler I forgot how to do custom event arguments 😅, so it is `GenericEventArguments` as it is. I expect in the future that it's something like: 

```py
@dataclass(**KWONLY_SLOTS)
class JsonEditorChangeEventArguments(UiEventArguments):
    node: str
    param: Any
```

Of course, the user has to do due diligence and check the node whether it is valid node, since if they do `click C call {handler}("blah", "C some args")`, then they record to the backend that node `blah` was clicked, which doesn't exist in the first place, and is likely going to cause some errors down the line.

thetableman:
@evnchn, thanks for taking this on. 

What's the outcome if that mermaid markdown is created using an f-string, won't that `{handler}` placeholder become an issue?

evnchn:
Supposedly, they need to ensure that their f-string, after substitution, gets to have a `{handler}` in them

Could be some very nasty syntax to get that working. Let me assess.

evnchn:
Ok so apparently for f-string, you need `{{handler}}` so that after f-string does its substitution, it becomes `{handler}` and the format in our `ui.mermaid` works. 

Syntax is a bit nasty, but i think it's better than a direct replace (cannot escape the keyword if keyword turns out to be inside Mermaid diagram) / regex (hard to grasp)

falkoschindler:
Thanks for the proposal, @evnchn!

I'm still sceptical. Consider the following diagram:
```py
ui.mermaid('''
    graph LR
        A[trigger] -->|event| B{handler}
''')
```
Screenshot 2025-06-12 at 17 11 42

The diamond shape looks like our new event handler, but is normal Mermaid syntax. If we want to extend the Mermaid language with a custom event handler notation, we need to find something that isn't already part of the language. This could, however,  get pretty cryptic.

evnchn:
@falkoschindler Can agree that the Python f-string is getting cryptic. I also pondered the possibility that someone would invoke the browser-side Mermaid content update function by means of `getElement(5).update`, which this PR would be incompatible with, since the find-and-replace logic is at Python-side. 

With that, I made #4862, the diff is smaller, and I don't think it would be possible to run into collision issues unlike this PR. 

Marking this as draft, unless you want to switch gears back to this approach, which I think is rather unlikely.

evnchn:
It's either #4862 or #4871, not this.

PR #4871, opened 2025-06-16
Introduce Mermaid `on_node_click` using a no-markdown solution

thetableman:
### Motivation

This PR introduces another alternative way of adding node click events to Mermaid diagrams without use of the global `emitEvent`.

The method was first raised in #4845 and demonstrated in #4862. This implementation differs to the previous working demonstration by ensuring the rendering queue is included as per #4852 and #4853.

### Implementation

This approach bypasses Mermaid's own `click` event methods - the adding of `click <nodeName> call <func(params)>` to the markdown - by attaching a handler directing to the node DOM elements. The method utilizes `querySelectorAll` to identify the nodes within each diagram based on html tag and class names.

### Progress

454577347-5bc288bb-d98f-49be-ac4c-84e57ff24152

#### Pros
- No requirement for declaring click events in Mermaid
- No confusion about Mermaid's click syntax
- No placeholder handlers in markdown
- No unique event handlers
- Returns consistent data, both within itself (each trigger has the same arguments) and as expected from a NiceGUI element
- Returns node name, node html id, and node text
- Adds protection against double-triggers by preventing multiple event handlers being attached
- Works with multiple diagrams on the same page
- Works where multiple diagrams share similar node names
- Doesn't stop or prevent traditional Mermaid click events if defined in the markdown
- Adds the ability for further functionality through node html id

#### Cons
- Relies on finding g.node with a query selector, leading to potential breaking changes from Mermaid (unlikely)
- Relies on extracting node name from its html id, leading to potential breaking changes from Mermaid (unlikely)
- Currently only works with diagrams with g.node like 'graph', 'state', and 'class' (though these are also the most popular diagram types)

#### Inbetweens
- Ability to add additional query selectors to support other diagram types that even Mermaid doesn't support click events for
- Adds click events in a manner that doesn't match Mermaid documentation

#### Mermaid click events still work

Because no changes are made to the markdown or the Mermaid click definitions, they still work with existing methods and as per the Mermaid documentation.

```python

from nicegui import ui

ui.mermaid('''
    graph LR;
        A --> B;
        click A alert

ui.run()
```
454578200-a7540950-1652-41e9-b633-47cde690464a

#### You can reference the nodes in the DOM directly

Because we have the node html id (and not just the node name) we can reference it directly and make changes.

```python

from nicegui import ui

node_styles = {}

def style_node(e):
    node_id = e.args['nodeId']
    node_el = ui.query(f'#{node_id} rect')
    
    if node_id not in node_styles:
        node_styles[node_id] = True
    else:
        node_styles[node_id] = not node_styles[node_id]
    
    if node_styles[node_id]:
        node_el.style(add='fill: yellow;')
    else:
        node_el.style(remove='fill: yellow;')

    e.args['state'] = node_styles[node_id]
    ui.notify(e)

ui.mermaid('''
    graph LR;
        A --> B;
    ''', on_node_click=lambda e: style_node(e))
    
ui.run()
```
454584526-1f342412-761a-4419-a8a0-e694d429c33c

### DIFF
--- a/website/documentation/content/mermaid_documentation.py
+++ b/website/documentation/content/mermaid_documentation.py
@@ -14,17 +14,19 @@ def main_demo() -> None:
     list(ui.context.client.elements.values())[-1].props['config'] = {'securityLevel': 'loose'}  # HACK: for click_demo
 
 
-@doc.demo('Handle click events', '''
-    You can register to click events by adding a `click` directive to a node and emitting a custom event.
-    Make sure to set the `securityLevel` to `loose` in the `config` parameter to allow JavaScript execution.
+@doc.demo('Handle node events', '''
+    You can register to click events by adding a `click` directive to a node to call `nodeClick()`, this will trigger the callback defined at `on_node_click`.
+    When a callback is specified the `config` is updated to include ``{"securityLevel": "loose"}`` to allow JavaScript execution.
 ''')
 def click_demo() -> None:
     ui.mermaid('''
     graph LR;
-        A((Click me!));
-        click A call emitEvent("mermaid_click", "You clicked me!")
-    ''', config={'securityLevel': 'loose'})
-    ui.on('mermaid_click', lambda e: ui.notify(e.args))
+        A((Click Me));
+        B((Or Click Me));
+        A --> B;
+        click A call nodeClick()
+        click B call nodeClick()
+    ''', on_node_click=lambda e: ui.notify(e.args))
 
 
 @doc.demo('Handle errors', '''
--- a/website/documentation/content/mermaid_documentation.py
+++ b/website/documentation/content/mermaid_documentation.py
@@ -15,7 +15,7 @@ def main_demo() -> None:
 
 
 @doc.demo('Handle node events', '''
-    You can register to click events by adding a `click` directive to a node to call `nodeClick()`, this will trigger the callback defined at `on_node_click`.
+    You can register to click events by adding a callback to the `on_node_click` parameter.
     When a callback is specified the `config` is updated to include ``{"securityLevel": "loose"}`` to allow JavaScript execution.
 ''')
 def click_demo() -> None:
@@ -24,8 +24,6 @@ def click_demo() -> None:
         A((Click Me));
         B((Or Click Me));
         A --> B;
-        click A call nodeClick()
-        click B call nodeClick()
     ''', on_node_click=lambda e: ui.notify(e.args))
 
 
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -9,7 +9,17 @@ export default {
     this.initialize();
     this.update(this.content);
   },
+  beforeUnmount() {
+    this.clearHandler();
+  },
   methods: {
+    getHandlerName() {
+      return `nodeClick_${this.clickInstance}`;
+    },
+    clearHandler() {
+      const handlerName = this.getHandlerName();
+      if (window[handlerName]) delete window[handlerName];
+    },
     initialize() {
       try {
         mermaid.initialize(this.config || {});
@@ -25,6 +35,22 @@ export default {
         const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", content);
         this.$el.innerHTML = svg;
         bindFunctions?.(this.$el);
+
+        if (this.clickInstance) {
+          const handlerName = this.getHandlerName();
+          window[handlerName] = (nodeId, nodeText) => {
+            this.$emit("nodeClick", {
+              node: this.getNodeName(nodeId),
+              nodeId,
+              nodeText,
+            });
+          };
+
+          this.$nextTick(() => {
+            this.attachClickHandlers(this.getHandlerName());
+          });
+        };
+
       } catch (error) {
         const { svg, bindFunctions } = await mermaid.render(this.$el.id + "_mermaid", "error");
         this.$el.innerHTML = svg;
@@ -34,9 +60,32 @@ export default {
         this.$emit("error", mermaidErrorFormat);
       }
     },
+    attachClickHandlers(handlerName) {
+      const clickables = this.$el.querySelectorAll("g.node.clickable");
+      clickables.forEach(node => {
+        node.removeAttribute("onclick");
+        node.onclick = null;
+
+        const newNode = node.cloneNode(true);
+        node.parentNode.replaceChild(newNode, node);
+
+        const nodeText = newNode.textContent.trim();
+        newNode.addEventListener("click", () => {
+          window[handlerName](newNode.id, nodeText);
+        });
+      });
+    },
+    getNodeName(domId) {
+      if (!domId) return undefined;
+      const parts = domId.split("-");
+      if (parts.length >= 3) return parts.slice(1, -1).join("-");
+      if (parts.length === 2) return parts[1];
+      return domId;
+    },
   },
   props: {
     config: Object,
     content: String,
+    clickInstance: Number,
   },
 };
--- a/nicegui/elements/mermaid.js
+++ b/nicegui/elements/mermaid.js
@@ -12,17 +12,7 @@ export default {
     this.initialize();
     this.update(this.content);
   },
-  beforeUnmount() {
-    this.clearHandler();
-  },
   methods: {
-    getHandlerName() {
-      return `nodeClick_${this.clickInstance}`;
-    },
-    clearHandler() {
-      const handlerName = this.getHandlerName();
-      if (window[handlerName]) delete window[handlerName];
-    },
     initialize() {
       try {
         mermaid.initialize(this.config || {});
@@ -34,7 +24,6 @@ export default {
     async update(content) {
       if (this.last_content === content) return;
       this.last_content = content;
-
       queue.push({ element: this.$el, content: content });
       if (is_running) return;
       is_running = true;
@@ -44,6 +33,10 @@ export default {
           const { svg, bindFunctions } = await mermaid.render(element.id + "_mermaid", content);
           element.innerHTML = svg;
           bindFunctions?.(element);
+          if (this.clickInstance) {
+            await this.$nextTick();
+            this.attachClickHandlers(element);
+          };
         } catch (error) {
           const { svg, bindFunctions } = await mermaid.render(element.id + "_mermaid", "error");
           element.innerHTML = svg;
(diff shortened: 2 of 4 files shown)


falkoschindler:
Thanks for the pull request, @thetableman! To be honest, it got a bit lost in the long list of open PRs.
@evnchn Since you worked on a similar PR, would you mind having a look into this one?

What I noticed at a first glimpse:
- `ui.mermaid`'s docstring has been partly indented, causing unnecessary diff.
- `test_click_mermaid_node` and `click_demo` has been changed which looks like an API change. This would require us to wait for NiceGUI 4.0 to avoid breaking user code. Or does the solution using `emitEvent` still work? Then we should restore the old test as an alternative way to handle clicks.
- We should merge from main to see if everything works with NiceGUI 3.0.

evnchn:
Note that https://github.com/zauberzeug/nicegui/pull/4871/commits/79bb1cec673a25ca5e03f490a0258ba9d990c0d6 may not be a very good job! In progress...

nicegui/elements/mermaid/mermaid.js
```diff
@@ -44,9 +48,37 @@ export default {
       }
       is_running = false;
     },
+    attachClickHandlers(element) {
```
evnchn: According to my instinct Falko will ask to inline this function since it used once only.

nicegui/elements/mermaid/mermaid.py
```diff
@@ -1,12 +1,13 @@
...
 
+from ...events import GenericEventArguments, Handler
 from ..mixins.content_element import ContentElement
 
 
 class Mermaid(ContentElement, component='mermaid.js', esm={'nicegui-mermaid': 'dist'}):
     CONTENT_PROP = 'content'
 
-    def __init__(self, content: str, config: Optional[dict] = None) -> None:
+    def __init__(self, content: str, config: Optional[dict] = None, on_node_click: Optional[Handler[GenericEventArguments]] = None) -> None:
```
evnchn: Custom event arguments type?

nicegui/elements/mermaid/mermaid.py
```diff
@@ -25,9 +26,15 @@ def __init__(self, content: str, config: Optional[dict] = None) -> None:
...
         :param config: configuration dictionary to be passed to ``mermaid.initialize()``
+        :param on_node_click: callback that is invoked when a node is clicked
         """
         super().__init__(content=content)
-        self._props['config'] = config
+
+        self._props['config'] = config or {}
+
+        if on_node_click:
+            self.on('nodeClick', on_node_click)
```
evnchn: Python-land events are `snake_case`

website/documentation/content/mermaid_documentation.py
```diff
@@ -14,17 +14,23 @@ def main_demo() -> None:
...
+        A((Click me!));
+        click A call emitEvent("mermaid_click", "You clicked me!")
     ''', config={'securityLevel': 'loose'})
     ui.on('mermaid_click', lambda e: ui.notify(e.args))
+    ui.mermaid('''
+    graph LR;
+        A((Click Me));
+        B((Or Click Me));
+        A --> B;
+    ''', on_node_click=lambda e: ui.notify(e.args))
```
evnchn: Messy leftover from the merge 🤡

nicegui/elements/mermaid/mermaid.js
```diff
@@ -44,9 +48,37 @@ export default {
...
+      const clickables = element.querySelectorAll("g.node");
+      console.log(clickables)
+      clickables.forEach(node => {
```
evnchn: debugging print will be gone shortly and we can inline `clickables`

evnchn (commented):
It's a to-do-list.

nicegui/elements/mermaid/mermaid.py
evnchn: https://github.com/zauberzeug/nicegui/pull/4871/commits/b4352f8b72219565b0fa115e4227180f22af9dc3

evnchn:
https://github.com/zauberzeug/nicegui/pull/4871/commits/65f577bb85d1a0dfd1e7b30a44288baf629209b2 is a non-simple-win change: 

I tried extensively and it works. I think since you `element.innerHTML = svg;` that (1) all event handlers are gone, and (2) even if thery are not gone, the data attribute is also gone.

website/documentation/content/mermaid_documentation.py
evnchn: https://github.com/zauberzeug/nicegui/pull/4871/commits/48bbe19f8d759bf869b21a0b141a55c6fbf8096e

nicegui/elements/mermaid/mermaid.js
evnchn: https://github.com/zauberzeug/nicegui/pull/4871/commits/7fb729ab595ac2665b460d0286df5d3e7170a664

nicegui/elements/mermaid/mermaid.js
evnchn: https://github.com/zauberzeug/nicegui/pull/4871/commits/999ebf5c297cf962ab0a38d2467e82329a0cd61f

nicegui/elements/mermaid/mermaid.js
```diff
@@ -44,9 +56,17 @@ export default {
       }
       is_running = false;
     },
+    getNodeName(domId) {
+      if (!domId) return undefined;
+      const parts = domId.split("-");
+      if (parts.length >= 3) return parts.slice(1, -1).join("-");
+      if (parts.length === 2) return parts[1];
+      return domId;
+    },
```
evnchn: Is this accurate? I never seen 2 parts, never seen 4 parts

nicegui/elements/mermaid/mermaid.js
```diff
@@ -24,15 +24,27 @@ export default {
...
           element.innerHTML = svg;
           bindFunctions?.(element);
+          if (clickInstance) {
+            element.querySelectorAll("g.node").forEach((node) => {
+              node.style.cursor = "pointer";
+              node.addEventListener("click", () => {
+                getElement(element).$emit("node_click", {
+                  node: this.getNodeName(node.id),
+                  nodeId: node.id,
+                  nodeText: node.textContent.trim(),
```
evnchn: Potentially costly to send the entire text content. Also encourages malpractice (lookup by value). So I don't think this is a good idea.

nicegui/elements/mermaid/mermaid.py
```diff
@@ -25,10 +26,15 @@ def __init__(self, content: str, config: Optional[dict] = None) -> None:
 
         :param content: the Mermaid content to be displayed
         :param config: configuration dictionary to be passed to ``mermaid.initialize()``
+        :param on_node_click: callback that is invoked when a node is clicked
         """
         super().__init__(content=content)
         self._props['config'] = config
 
+        if on_node_click:
+            self.on('node_click', on_node_click)
```
evnchn: Under this code, how do I use the wonderful `js_handler` (https://github.com/zauberzeug/nicegui/pull/4689)?

evnchn: I think they will call `.on` themselves, at their own risk.

nicegui/elements/mermaid/mermaid.py
evnchn: At https://github.com/evnchn/nicegui/tree/mncea, if see fit.

nicegui/elements/mermaid/mermaid.js
falkoschindler: Although it's handy for testing and demos, I removed it. It's easier to add it later if needed than to remove it and break things.

nicegui/elements/mermaid/mermaid.js
falkoschindler: With Sonnet 4.5 I came to the conclusion that there are always 3+ parts: `{diagram-type}-{node-name}-{index}`. The node name can have hyphens.
Anyway, the `>=3` case should be sufficient.

falkoschindler (approved):
Thanks everyone! This PR turned out pretty nicely. 🙂

While reviewing it, I added `MermaidNodeClickEventArguments` with node ID and name, inlined a simplified `getNodeName`, added an `on_node_click` method for late event registrations and simplified the test a little.

Should be ready to merge.
@evnchn Do you agree?

evnchn:
@falkoschindler this PR is good in its own right but the only thing which may be troublesome down the line is the fact that **we are actively inferring the node name from the DOM ID for the user.**

All signs point to it _not_ happening, but if Mermaid decides to be a 🤡 in arbitrarily changing how the DOM ID works, we'll be in trouble as the man in the middle of user <-> Mermaid. 

I did consider briefly the following API design:

- we do not do any DOM ID to name conversion
- user will pass DOM-ID-specific handlers (side benefit: can make only some nodes clickable, which affects pointer hover style)
- if the DOM ID is non-existent, it will raise error in the JS console (and possibly the Python console) and fail any tests, at which point (1) we'll announce breaking change, (2) breaks the tests of any user who uses pytest on their project


But Id also like to know what do you think on this before proceeding with the change. Thanks!

falkoschindler:
@evnchn Yes, the current implementation is not 100% future-proof. But I can't find a good way around it.

How would your approach look like? How would the user register event handlers?

- [ ] While thinking about node IDs and HTML element IDs I noticed that we currently store the Mermaid ID under the attribute `name`. This might confuse Mermaid users. Maybe we don't even need the HTML `id` (it is internal and undocumented anyway) and can simply put the Mermaid ID into the only event argument `id`.

evnchn:
> But I can't find a good way around it.

It could be great if you let me cook for a while, because if we release this into public API then it's there until another breaking...

I already have something in mind.

evnchn:
> Interplay between `name`, and `id`

@falkoschindler IMO there shall be neither. 

- `id` is easy to confuse with our numeric `id` for each NiceGUI element. 
- `name` is inferred from HTML ID and is inaccurate in the long run. 

As shown in https://github.com/zauberzeug/nicegui/pull/4871/commits/7dd64e4f8baabdf025abe11c383fee0ffc7e41ef we should only have `html_id` as the sole attribute that is the only unchanging parameter and is the "single-source-of-truth". 

Feel free to remove my last 2 commits if they don't look good. Too confusing to branch off and PR again.

evnchn:
Further comments:

> With Sonnet 4.5 I came to the conclusion 

LLMs always agree with you... Risky move...

> HTML ID is internal

It is to Mermaid, but to NiceGUI it is the only point of attachment. These 2 changes attempt to formalize and respect this fact by having it front-and-center while failing loud in case of mismatch.

falkoschindler:
@evnchn Although communicating HTML IDs is more future-proof, it feels not very user-friendly to me. When defining a node as `A[Node A]`, I'd expect a click event to yield either "A" or maybe "Node A", but not something apparently unrelated like "flowchart" or something fragile like the index which will change as soon as you re-organize your nodes. Sure, the user can split the HTML ID and extract the node ID. But doing it right is nontrivial (as we noticed) and adds complexity to the user code. It would be _nice_ if NiceGUI would communicate Node IDs which the users explicitly defined themselves.

Actually the danger of a future incompatible HTML ID pattern is low, because we're bundling Mermaid.js with NiceGUI and can write regression tests to make sure a dependency update won't break anything.

Your proposed `clickable_nodes` parameter seams like an additional feature which I'd rather not add to this PR at this moment. I'm not sure if there is any demand and it doesn't really solve the ID problem if most users will omit this parameter.

Long story short: I'd prefer to revert your two last commits. 🫤

evnchn:
Actually we already have the test, and it can detect regressions, and so we should be able to just revert these 2 commits and get going. 

Note that this means we're bearing the responsibility for HTML ID to node translation, but I read that you're ready for the commitment in exchange for a more nice user experience. On that I don't have any concerns on my side. 

We'll drop a note to self somewhere that the tests are for anti-regression in HTML ID parsing logic because it may be a real headscratcher a year later if the test breaks then. 

@falkoschindler can you do the reverting? Thanks.

Issue #4878, opened 2025-06-21
KeyEventArguments.key.space is incorrect


eulerleibniz:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui
from nicegui.events import KeyEventArguments


async def keyboard_event(e: KeyEventArguments) -> None:
    if e.action.keydown:
        if not e.action.repeat:
            print("e.action:        ", e.action)
            print("e.modifiers:     ", e.modifiers)

            # For Space key only
            print("e.key.space:     ", e.key.space)  # False
            print("e.key.name:      ", e.key.name)  # "" -> Empty String
            print("e.key.code:      ", e.key.code)  # "Space"


ui.label("Hi. Press Space And Check Terminal Please")
ui.keyboard(on_key=keyboard_event)
ui.run()
```

### Description

Hi all the Nice people.

There is a simple issue with how the `KeyEventArguments.key.space` property is calculated.
In this file https://github.com/zauberzeug/nicegui/blob/main/nicegui/events.py, the property is checked as follow:

```    @property
    def space(self) -> bool:
        """Whether the key is the space key."""
        return self.name == 'Space'
```

But when pressing Space key, it sends an event that its `KeyEventArguments.key.name` value is empty string. So the above property wont work. When pressing space, `KeyEventArguments.key.code` value is Space which currently i use that in my code.

That's All. Thank You.

### NiceGUI Version

2.16.1

### Python Version

Python 3.12.9

### Browser

Firefox

### Operating System

Windows

falkoschindler:
Oh wow, thanks for reporting this issue, @eulerleibniz! I wonder why this went unnoticed for so long!

By the way, the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values) say it used to be "Spacebar", but never "Space":
> Older browsers may return "Spacebar" instead of " " for the Space Bar key. Firefox did so until version 37.

Would you like to create a quick pull request, making you an official contributor of NiceGUI? 🙂

eulerleibniz:
[...]

PR #4879, opened 2025-06-22
Fix: Corrected how space key press is checked in the class KeyboardKey of events.py file

eulerleibniz:
### Motivation
Fix for issue [KeyEventArguments.key.space is incorrect #4878](https://github.com/zauberzeug/nicegui/issues/4878)

### Implementation
in events.py, class KeyboardKey replaced the following:
    @property
    def space(self) -> bool:
        """Whether the key is the space key."""
        return self.name == 'Space'

with this:
    @property
    def space(self) -> bool:
        """Whether the key is the space key."""
        return self.name == ' '

***Note***: We could use `return self.code == 'Space'` as well, but it would be inconsistent with the rest of the class which always uses `return self.name == value`

### Progress

### DIFF
--- a/nicegui/events.py
+++ b/nicegui/events.py
@@ -249,7 +249,7 @@ class KeyboardKey:
     @property
     def space(self) -> bool:
         """Whether the key is the space key."""
-        return self.name == 'Space'
+        return self.name == ' '
 
     @property
     def page_up(self) -> bool:


falkoschindler (approved):
Thanks a lot for the pull request, @eulerleibniz!

Looks good and works perfectly fine. 👍🏻

Issue #4875, opened 2025-06-19
deselection of rows in table does not work when the row_key is set after the table is initialised


thugentobler:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
import pandas as pd
from nicegui import ui


class GuiResults:
    def __init__(self):
        with ui.dialog() as self.dialog, ui.card():
            ui.markdown("#### Title")
            self.table = ui.table(columns=[], rows=[], selection="multiple")

    async def show(self, check_name: str, data: pd.DataFrame):
        # Update the table with the new data
        self.table.columns = [
            {"name": col, "label": col, "field": col} for col in data.columns
        ]

        # add data to table
        self.table.rows = data.to_dict(orient="records")

        # set the row key
        if "Number" in data.columns:
            self.table.row_key = "Number"

        self.dialog.open()


if __name__ in {"__main__", "__mp_main__"}:

    gui_results = GuiResults()

    # Example data
    example_data = pd.DataFrame(
        {
            "Number": [1, 2, 3],
            "Name": ["Alice", "Bob", "Charlie"],
            "Score": [90, 85, 88],
        }
    )

    ui.button(
        "Show Example Data",
        on_click=lambda: gui_results.show("Example Check", example_data),
    )

    # Show the dialog with example data
    ui.run()
```

### Description

When the row_key is set after the table is initialised, and the table is set to selection="multiple", the rows cannot be deselected, there is a keyerror raised.

```
Traceback (most recent call last):
    File "[...]\.venv\Lib\site-packages\nicegui\events.py", line 436, in handle_event
        result = cast(Callable[[EventT], Any], handler)(arguments)
    File "[...]\.venv\Lib\site-packages\nicegui\elements\table.py", line 85, in handle_selection
        self.selected = [row for row in self.selected if row[row_key] not in e.args['keys']]
                                                                                ~~~^^^^^^
KeyError: 'id'
```

In my humble opinion, `row_key `should be replaced by `self.row_key`, else it will always default to the argument of row_key given in the initialisation of the table.

### NiceGUI Version

2.12.1

### Python Version

3.13

### Browser

Chrome

### Operating System

Windows

falkoschindler:
Thanks for reporting this issue, @thugentobler!

[...]

PR #4893, opened 2025-06-26
Fix deselection of table rows by using the correct row key

thugentobler:
### Motivation

Discovered the bug in Issue #4875 where rows in a table can be selected, but not deselected. This is due to the function `handle selection` which refers to the row_key which is given in the initialisation of the table, but if it is assigned afterwards, it does not update in the function. Therefore replacing the row_key with self.row_key, to ensure that the class variable is used.


falkoschindler (approved):
Very nice, @thugentobler! Thanks for the pull request!
Let's merge 🚀

thugentobler:
Thanks for your help @falkoschindler  :)

PR #4902, opened 2025-06-28
Use ui.sub_pages to greatly speed up the documentation

rodja:
### Motivation

The documentation currently use @ui.page declarations. Due to http and socketio handshakes, content transfer and Quasar app loading, these have a noticeable delay when switching from one page to the next.

### Implementation

This PR uses the new `ui.sub_page` element from #4821 to implement the whole documentation as a single page application. A major difficulty was the left-drawer implementation for the element tree. To decide wether it should be opened or closed is tricky because non-desktop sized screens should never open it automatically. To make it work, I had to send the media breakpoint information to the backend via a custom javascript event.

### Progress

  - [ ] fix styling of index page for ultra-wide screens
  - [ ] add url-fragment capabilities to `ui.sub_pages` (in a separate PR)

### DIFF
--- a/main.py
+++ b/main.py
@@ -8,7 +8,7 @@ from fastapi.responses import RedirectResponse
 from starlette.middleware.sessions import SessionMiddleware
 
 from nicegui import app, ui
-from website import anti_scroll_hack, documentation, fly, imprint_privacy, main_page, svg
+from website import anti_scroll_hack, documentation, fly, header, imprint_privacy, main_page, svg
 
 # session middleware is required for demo in documentation
 app.add_middleware(SessionMiddleware, secret_key=os.environ.get('NICEGUI_SECRET_KEY', ''))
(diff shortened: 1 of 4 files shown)


falkoschindler (commented):
I'm not sure if this PR is meant to be "ready", but I noticed at least three things:

1. The NiceGUI face on the index page is not placed at the edge of the screen on wide displays.
2. Clicking on the imprint link yields error 500 "add_header() missing 1 required positional argument: 'menu'".
3. Clicking on items in the tree menu or the search dialog causes the page to reload. Don't they support the SPA mechanism yet?

Update: I just noticed that some of my points are mentioned in the "Progress" task list. 🤦🏻‍♂️

rodja:
@falkoschindler I fixed the imprint and tree navigation. I could not reproduce page reloads when using the search dialog. And I hope that you can fix the styling for wide screens -- I'm quite lost here.

main.py
```diff
@@ -32,28 +30,55 @@ async def _post_dark_mode(request: Request) -> None:
 
 
 @ui.page('/')
+@ui.page('{path:path}')
```
falkoschindler:
```suggestion
@ui.page('/{path:path}')
```
Might not make a difference, but we usually register "/" and "/{path:path}".

PR #4905, opened 2025-06-29
Introduce PageArgs for ui.sub_pages for query parameters and data access

rodja:
### Motivation

This PR extends the `ui.sub_pages` functionality from #4821 to support `PageArgs`, enabling sub pages to access query parameters and shared data in a unified style. This is also in preparation for handling url fragments.

### Implementation

- **PageArgs Integration**: Sub page functions can accept a `PageArgs` parameter for unified access to path parameters, query parameters, and shared data
- **Data Sharing**: The `data` parameter in `ui.sub_pages()` allows passing shared objects to sub pages via `PageArgs`
- **Automatic Injection**: As before, builder functions can use type annotated parameter names to automatically receive individual parameters from `PageArgs`.
- **Optional Parameters**: builder functions now also support optional parameters

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests have been added (or are not necessary).
- [x] Documentation has been added (or is not necessary).

PR #4906, opened 2025-06-29
Make ui.sub_pages navigation aware of url fragments

rodja:
### Motivation

Sub-pages in NiceGUI didn't support URL fragments (e.g., `/page#section`), preventing users from:
- Linking directly to specific sections within sub-pages
- Using `ui.link_target()` elements for in-page navigation
- Proper scroll positioning when navigating with fragments

### Implementation

Building on top of #4905 the implementation was straight forward. Only remark might be the rather large addition to `sub_pages.js` which handles the "same page scrolling" which can and should be fully on the browser side.

### Progress

### DIFF
--- a/examples/single_page_app/custom_sub_pages.py
+++ b/examples/single_page_app/custom_sub_pages.py
@@ -1,7 +1,7 @@
 from typing import Callable
 
 from nicegui import app, ui
-from nicegui.elements.sub_pages import SubPages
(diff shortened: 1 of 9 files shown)


rodja:
Because the architecture changed greatly, we decided the review will be done after merging this and the other stacked branches into #4821.

PR #4913, opened 2025-06-30
`ui.code` drop `on_connect` by defining class using JS dynamically

### LINKED ISSUE #4912: Server side memory leak when clearing ui.code()
### First Check

### Example Code

```python
from nicegui import ui

row = ui.row()
i = 0

def more_code():
    global i
    row.clear()
    with row:
        for k in range(50):
            ui.code(language='xml', content=f'<data>{i}</data>')
            i += 1

ui.timer(callback=more_code, interval=1)
ui.run()
```

### Description

I noticed that when I create, clear and re-create a bunch of ui.code() elements, memory usage on the server creeps up. The given example creeps by 1 MB every 1-3 seconds. A real life example reached 4 gigs before becoming so slow that it had to be restarted. Tested in real life on Linux and the minimal example tested on Windows.

### NiceGUI Version

2.20.0

### Python Version

Python 3.13.3 (tags/v3.13.3:6280bb5, Apr  8 2025, 15:28:43) [MSC v.1943 64 bit (ARM64)] on win32

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

evnchn:
### Motivation

This PR fix #4912 by not using `self.client.on_connect`, which is know to cause memory leak as it adds a lambda to `Client.connect_handlers` list, but that lambda is never deleted at all. 

### Implementation

To achieve the original functionality of "hide copy button if no `navigator.clipboard`, this PR defines `nicegui-invisible-if-no-clipboard` dynamically via JS in `nicegui.js`

### Progress

- [x] Pytests are impossible? (how can we test for memory leak?)

### DIFF
--- a/nicegui/elements/code.py
+++ b/nicegui/elements/code.py
@@ -25,7 +25,7 @@ class Code(ContentElement, default_classes='nicegui-code'):
         super().__init__(content=remove_indentation(content))
 
         with self:
-            self.markdown = markdown().classes('overflow-auto') \
+            self.markdown = markdown().classes('overflow-auto nicegui-invisible-if-no-clipboard') \
                 .bind_content_from(self, 'content', lambda content: f'```{language}\n{content}\n```')
             self.copy_button = button(icon='content_copy', on_click=self.show_checkmark) \
                 .props('round flat size=sm').classes('absolute right-2 top-2 opacity-20 hover:opacity-80') \
@@ -34,10 +34,6 @@ class Code(ContentElement, default_classes='nicegui-code'):
         self._last_scroll: float = 0.0
         self.markdown.on('scroll', self._handle_scroll)
         timer(0.1, self._update_copy_button)
-
-        self.client.on_connect(lambda: self.client.run_javascript(f'''
-            if (!navigator.clipboard) getHtmlElement({self.copy_button.id}).style.display = 'none';
-        '''))
 
     async def show_checkmark(self) -> None:
         """Show a checkmark icon for 3 seconds."""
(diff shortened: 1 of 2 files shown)


evnchn:
Notably the clipboard button is hidden for-real: 

{374823FF-7AA3-4916-B69C-DE0C5C99426E}

evnchn:
Had a thought. 

Though this PR is obvious more efficient than checking the clipboard every time the element is created, there could be an edge case where the browser's clipboard went from unavailable to available in the page's lifecycle (could occur for some enterprise locked-down machines). 

Then this would be technically a different behavior and break stuff, and #4919 seems like a better solution. I just have to test whether that leaks memory.

evnchn:
https://github.com/zauberzeug/nicegui/pull/4913#issuecomment-3019572847 

Well actually this behaviour is wrong. 127.0.0.1 should show the copy button. `navigator.clipboard` is available, when (1) secure context or (2) localhost.

evnchn:
After https://github.com/zauberzeug/nicegui/pull/4913/commits/5bd0b149d688b04d0abdef79127bfb647cde0cbf, this PR should be functionally correct. 

Here is what I am looking for: 

1. Do not leak memory (duh): Resolve #4912 
2. Copy button in 127.0.0.1, no copy button in 192.168.xxx.xxx, despite element transmitted post-connect: Resolve #4921

falkoschindler:
Thanks for the update and your thorough analysis, @evnchn!

I slightly prefer PR #4919 (based on the other idea of yours). It solves the problem within code.\* and doesn't bleed into global library code.

Let's close this one.

PR #4914, opened 2025-06-30
Unregister connect handler when deleting a `ui.code` element

falkoschindler:
### Motivation

The `ui.on_connect` registration of a lambda function in the initializer of `ui.code` causes a memory leak when removing and re-creating many `ui.code` elements.

### Implementation

By using a method `_hide_copy_button_if_no_clipboard` that can be removed from `client.connect_handlers` when the element is deleted, the leak is fixed.

### Progress

### DIFF
--- a/nicegui/elements/code.py
+++ b/nicegui/elements/code.py
@@ -35,9 +35,16 @@ class Code(ContentElement, default_classes='nicegui-code'):
         self.markdown.on('scroll', self._handle_scroll)
         timer(0.1, self._update_copy_button)
 
-        self.client.on_connect(lambda: self.client.run_javascript(f'''
+        self.client.on_connect(self._hide_copy_button_if_no_clipboard)
+
+    def _hide_copy_button_if_no_clipboard(self) -> None:
+        self.client.run_javascript(f'''
             if (!navigator.clipboard) getHtmlElement({self.copy_button.id}).style.display = 'none';
-        '''))
+        ''')
+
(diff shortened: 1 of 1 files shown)


evnchn:
For some reason the JavaScript buttons are still showing. 

image

I use 0.1s as an interval, may be a tad extreme, but anyways...

evnchn:
Using the test script in https://github.com/zauberzeug/nicegui/issues/4912#issuecomment-3019375041, this PR reloads the page, with console message reason (visible if you do preserve log): 

`reloading because handshake failed for clientId e22fd8e9-5f22-4c30-b32c-e02319a5263d`

evnchn:
```py
import asyncio
from nicegui import ui

@ui.page('/')
async def main_page():
    ui.code('# ui.code added before connect')
    await asyncio.sleep(1)
    ui.code('# ui.code added after connect')

ui.run()
```


main: 

{539A0628-DDA0-4DC2-BB68-BDBEF32B9171}

#4913: 

{E83261F0-5EB5-46E5-867D-13DE196123E3}

#4914 (this):

{B47A662A-2C59-4366-A821-D7F4331D18EE}

Luckily both PRs doesn't leak memory. But this PR uses more memory, though (to be expected)

evnchn:
@falkoschindler Let's close this since it inherits the bug #4921 from main. 

{AA88BFEC-E498-4C0E-A2B7-F628659D32B3}

Issue #4896, opened 2025-06-26
`ui.select` add unique does not work fully as intented


BaccanoMob:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
x = ui.select(
    [],
    new_value_mode="add-unique",
    multiple=True,
).on_value_change(lambda e: print(e.value))
ui.label().bind_text_from(x, "value")
```

### Description

https://github.com/user-attachments/assets/d0bd62bd-1e8c-488e-a51a-abad89b6a6cb

Note that I press enter when the dropdown is not highlighted (by pressing up arrow).

Ideally it should act as a toogle. Only difference between `add` in this case, is that the options do not populate. And sometimes I run into list out of index errors (with props("hide-dropdown-icon popup-content-class='hidden' use-chips")).

```
IndexError: list index out of range
list index out of range
Traceback (most recent call last):
  File "/project/.venv/lib/python3.12/site-packages/nicegui/events.py", line 431, in handle_event
    result = cast(Callable[[EventT], Any], handler)(arguments)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/project/.venv/lib/python3.12/site-packages/nicegui/elements/mixins/value_element.py", line 40, in handle_change
    self.set_value(self._event_args_to_value(e))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/project/.venv/lib/python3.12/site-packages/nicegui/elements/select.py", line 96, in _event_args_to_value
    args = [self._values[arg['value']] if isinstance(arg, dict) else arg for arg in e.args]
```

`toggle` and `add` mode works as intended AFAIK.

Also this does not happen in [quasar docs](https://quasar.dev/vue-components/select#using-menu-and-filtering) so it might have been possibly fixed in a later release.

### NiceGUI Version

2.20.0

### Python Version

3.12

### Browser

Chrome

### Operating System

Linux

falkoschindler:
Hi @BaccanoMob,

I don't think I understand the problem yet. Let's improve the example to make clearer what is going on:
```py
label = ui.label()
ui.select([], new_value_mode='add-unique', multiple=True).props('use-chips') \
[...]

PR #4916, opened 2025-07-01
Avoid duplicate values in `ui.select` in "add-unique" mode

falkoschindler:
### Motivation

In #4896 we noticed that there is a way to get duplicate values in `ui.select` even when using the "add-unique" mode:

```py
label = ui.label()
ui.select([], new_value_mode='add-unique', multiple=True).props('use-chips') \
    .bind_value_to(label, 'text', lambda x: ', '.join(x))
```

1. Type "abc"
2. Press Enter
3. Type "abc"
4. Close the popup by clicking on the arrow
5. Press Enter

### Implementation

By analyzing the received event arguments containing the existing selection `{"value": 0, "label": "abc"}` as well as the typed input `"abc"`, we can find and eliminate such duplicates before treating them as new values.

### Progress

### DIFF
--- a/nicegui/elements/select.py
+++ b/nicegui/elements/select.py
@@ -93,6 +93,14 @@ class Select(LabelElement, ValidationElement, ChoiceElement, DisableableElement,
             if e.args is None:
                 return []
             else:
+                if self._props.get('new-value-mode') == 'add-unique':
+                    # handle issue #4896: eliminate duplicate arguments
+                    for arg1 in [a for a in e.args if isinstance(a, str)]:
+                        for arg2 in [a for a in e.args if isinstance(a, dict)]:
+                            if arg1 == arg2['label']:
+                                e.args.remove(arg1)
+                                e.args.remove(arg2)
+                                break
                 args = [self._values[arg['value']] if isinstance(arg, dict) else arg for arg in e.args]
                 for arg in e.args:
                     if isinstance(arg, str):
--- a/nicegui/elements/select.py
+++ b/nicegui/elements/select.py
@@ -89,6 +89,7 @@ class Select(LabelElement, ValidationElement, ChoiceElement, DisableableElement,
         return self._is_showing_popup
 
     def _event_args_to_value(self, e: GenericEventArguments) -> Any:
+        # pylint: disable=too-many-nested-blocks
         if self.multiple:
             if e.args is None:
                 return []
--- a/nicegui/elements/select.py
+++ b/nicegui/elements/select.py
@@ -100,7 +100,6 @@ class Select(LabelElement, ValidationElement, ChoiceElement, DisableableElement,
                         for arg2 in [a for a in e.args if isinstance(a, dict)]:
                             if arg1 == arg2['label']:
                                 e.args.remove(arg1)
-                                e.args.remove(arg2)
                                 break
                 args = [self._values[arg['value']] if isinstance(arg, dict) else arg for arg in e.args]
                 for arg in e.args:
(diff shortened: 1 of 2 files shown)


BaccanoMob:
@falkoschindler I made a mistake by saying `Ideally it should act as a toggle` but `add-unique` should ignore duplicate values. `toggle` was for toggling if there was duplicate. So I below code changes would work as intended (that is, remove `e.args.remove(arg2)` line).

```python
                    for arg1 in [a for a in e.args if isinstance(a, str)]:
                        for arg2 in [a for a in e.args if isinstance(a, dict)]:
                            if arg1 == arg2['label']:
                                e.args.remove(arg1)
                                break
```

and for test it should be something like below (probably)

```python
screen.should_contain("value = ['a', 'd']" if multiple else 'value = d')
```

Also `Close the popup by clicking on the arrow` is also a bug. I just found out that this only happens for the first element when I was testing the above code snippet.

https://github.com/user-attachments/assets/94a096d5-bb8f-428f-8e7b-c73ea5267bcf

Notice the difference in dropdown highlight for `abc` and `def` where `abc` needs the up arrow only. (I patched the ui.select with above code to remove duplicates values).

Even with prop `popup-content-class='hidden'` the problem presists.

https://github.com/user-attachments/assets/9aa40249-41de-4018-9e4f-72a351c9fdc6

I have to press up arrow even if there isnt a dropdown to fix it. 

Should I make a separate issue for this?

falkoschindler:
### `e.args.remove(arg2)` or not

Are you sure the second "abc" shouldn't remove "abc" from the value? Shouldn't typing "abc" behave identically to clicking on "abc" in the dropdown?

### Close the popup by clicking on the arrow

I still don't understand the problem. But yes, if it is unrelated to the duplicate value, please create a separate issue with a more detailed description. What _exactly_ are you doing, what do you expect to happen, what happens instead? The videos don't show the arrow, so I don't see how they help to illustrate the problem.

BaccanoMob:
## `e.args.remove(arg2)` or not

Check [Quasar Select docs](https://quasar.dev/vue-components/select#the-new-value-mode-prop). Just try adding `abc` three times in the fields for add, add-unique, toggle features. You will notice that,
    1. `add` just adds value to the list allowing duplicate entries
    2. `add-unique` only allows adding new values to the list which makes the list behave like a set  (but ordered I guess)
    3. `toggle` adds value if its not on list and removes if its in the list.

> Are you sure the second "abc" shouldn't remove "abc" from the value

This is toggling which `toggle` mode handles. AFAIK, `toggle` mode works as intended 👍 

> Shouldn't typing "abc" behave identically to clicking on "abc" in the dropdown?

No, sending via input and clicking in the dropdown are two separate things. 

Clicking from dropdown fulling involves with toggling values alone. You can filter the dropdown with input. When `new-value-mode` is set then it should not highlight the first value of dropdown since the value from input must be retreated as a new value to make `add` or `add-unique` work as intended. 

Sending via input fully deals with how you want to use the value and its particularly different when you set `new-value-mode` since you have to `process the value` to define the updated list. It also involves toggling if `new-value-mode` is `toggle` which will be the only case identical to clicking the dropdown.

I hope I made some sense trying to explain it.

## Close the popup by clicking on the arrow
I'll open a new issue regarding it then!

falkoschindler:
I think the example https://quasar.dev/vue-components/select#the-new-value-mode-prop unveils a conceptional difference between QSelect and `ui.select`:
- When using the input field with `new_value_mode`, QSelect adds the user-defined string to the `value` - but _not_ to the `options`.
- Our `ui.select`, however, modifies the list of `options`.

This is probably why I'm keeping confusing values and options. Quasar's new-value-mode prop only defines what happens to the `value`. The `options` shouldn't change, and we could debate whether NiceGUI should change its current behavior to be more in line with QSelect.

Anyway, I removed the line `e.args.remove(arg2)` in f902127f4d2d1f939995f966ab9e06c94d5a21e7. This should conclude this PR, right?

BaccanoMob:
I guess Quasar view on options is as preset of default/expected values rather than a cumulative of values entered. 

Current behavior is not a big problem, though keeping options fixed might improve performance on larger lists. 

> This should conclude this PR, right?

Yes 👍 , Ill work on #4931 once this is merged!

PR #4919, opened 2025-07-01
Use code.js component to avoid memory-leaky on_connect registration

falkoschindler:
### Motivation

The `ui.on_connect` registration of a lambda function in the initializer of `ui.code` causes a memory leak when removing and re-creating many `ui.code` elements.

### Implementation

After PR #4914 didn't seem to work correctly, I ended up implementing @evnchn's approach 2 from https://github.com/zauberzeug/nicegui/issues/4912#issuecomment-3019408080. It introduces a code.js component setting the display style for the copy button on `mounted`.

### Progress

### DIFF
--- a/nicegui/elements/code.js
+++ b/nicegui/elements/code.js
@@ -1,0 +1,6 @@
+export default {
+  template: `<div><slot></slot></div>`,
+  mounted() {
+    if (!navigator.clipboard) this.$el.querySelector(".q-btn").style.display = "none";
+  },
+};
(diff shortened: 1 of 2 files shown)


evnchn (approved):
To test this comprehensively: 

1. Doesn't leak memory: Basically same memory usage as #4913. 

{7317E700-1AA7-4A81-9698-F451EE23CF56}

2. `localhost` context, button shows

{8B80632B-40C3-4060-8401-3D105E1F5340}


3. Insecure context, button does not show. 

{92B29581-81D3-4716-AF5A-76CDAA26EB53}

LGTM

PR #4933, opened 2025-07-03
Fix nested propagation of bindable properties

falkoschindler:
### Motivation

We noticed that changing the value of a bindable property multiple times within the change handler of another bindable property doesn't work correctly:

```py
class Demo:
    a = binding.BindableProperty()
    b = binding.BindableProperty(on_change=lambda obj, _: obj.change_a())

    def __init__(self) -> None:
        self.a = 0
        self.b = 0

    def change_a(self) -> None:
        self.a = 1
        self.a = 2

demo = Demo()
ui.label().bind_text_from(demo, 'a', lambda a: f'a = {a}')  # stuck at 1
ui.number().bind_value_to(demo, 'b')  # should set a to 1 and then 2
```

### Implementation

We correctly assumed that the context variable `propagation_visited` wasn't used correctly. At the first assignment to `self.a` the property is added to the set of visited properties. Then the propagation of the second assignment falsely assumes that `self.a` has already been propagated.

Now we make sure to always open a new propagation context whenever a value assignment or a refresh step of the binding loop triggers a propagation. This eliminates the need for distinguishing between root and non-root calls.

This PR also adds a pytest to cover this case.

### Progress

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -75,53 +75,50 @@ async def refresh_loop() -> None:
 
 @contextmanager
 def _propagation_context():
-    visited = propagation_visited.get()
-    is_root_call = visited is None
-    if is_root_call:
-        visited = set()
-        token = propagation_visited.set(visited)
+    token = propagation_visited.set(set())
     try:
-        yield visited
+        yield
     finally:
-        if is_root_call:
-            propagation_visited.reset(token)
+        propagation_visited.reset(token)
 
 
 def _refresh_step() -> None:
     t = time.time()
-    with _propagation_context():
-        for link in active_links:
-            (source_obj, source_name, target_obj, target_name, transform) = link
-            if _has_attribute(source_obj, source_name):
-                source_value = _get_attribute(source_obj, source_name)
-                value = transform(source_value) if transform else source_value
-                if not _has_attribute(target_obj, target_name) or _get_attribute(target_obj, target_name) != value:
-                    _set_attribute(target_obj, target_name, value)
+    for link in active_links:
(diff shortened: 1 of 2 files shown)


evnchn:
What happens if the user have something insane like:

```py
    a = binding.BindableProperty(on_change=lambda obj, _: obj.change_b())
    b = binding.BindableProperty(on_change=lambda obj, _: obj.change_a())
```

falkoschindler:
@evnchn This is what I asked myself as well. After this PR it will probably trigger an endless loop. Before it would have stopped due to the shared propagation context. Even though an endless loop is bad, it could be considered more correct than simply stopping the propagation.

NiklasNeugebauer:
I agree that an endless loop is the expected behavior.
I see this as equivalent to something like the following, which also results in an endless recursion.

```py
class MyClass:
    def __init__(self):
        self.a = 0
        self.b = 0

    def set_a(self, value):
        self.a = value
        self.set_b(value)

    def set_b(self, value):
        self.b = value
        self.set_a(value)

x = MyClass()

x.set_a(1)
```

evnchn:
I think it will go into an endless loop (and probably Python will trigger max recursion limit)

My concern is, since we are creating context variables in the process, will we consume a lot of memory?

In guaranteeing that the server will not crash, I feel like we should gracefully handle and detect endless loops.

NiklasNeugebauer:
In my opinion this is rather easy to detect by the user and not in Nicegui's responsibility to handle.
We catch recursions from direct bindings but should not need to handle these cases.

NiklasNeugebauer (approved):

Issue #4925, opened 2025-07-02
`ui.log` shouldn't grow when pushing lines


e-nesse:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

# Imagine this column is inside something that has height less than the browser viewport:
ui.query(".nicegui-content").classes("absolute-full p-0 m-0")
with ui.column().classes("w-full h-full p-0 gap-0"):
    # nav_area is light blue in the picture
    nav_area = ui.row(align_items="center", wrap=False).classes(
        "w-full gap-0 bg-gray-100 h-24"
    )
    content_area = ui.column().classes("w-full h-full p-0 gap-0 shadow-3")

with nav_area:
    ui.label("Nav area")

with content_area:
    with ui.splitter(value=33, limits=[25, 50]).classes("w-full h-full") as splitter:
        with splitter.before:
            left_col = ui.column().classes("w-full h-full p-2 gap-1")

        with splitter.after:
            right_col = ui.column().classes("w-full h-full p-2 gap-1")


with right_col:
    with ui.row().classes(
        "w-full gap-2 bg-red"
    ) as title_row:  # title_row is pink in the picture
        ui.label("title row")
    # same phenomenon happens with or without log_container:
    with ui.column().classes("w-full h-full bg-green") as log_container:
        # my_log is blacked in the picture and its scroll bars are visible
        # because it had many log events pushed to it when screenshotted:
        my_log = ui.log(max_lines=400).classes("w-full h-2/3")


with left_col:
    ui.button("Log", on_click=lambda: my_log.push("*" * 150, classes="text-xs"))

ui.run()
```

### Description

I am unable to produce an *exact* minimal reproducible example in my environment. However, ui.log() elements do not behave like other elements when it comes to respecting their own height settings (e.g., via .classes('h-full') or .classes('h-1/2') etc.) or staying within the bounds of a parent container such as a ui.column().

Initially, the empty ui.log() element will appear to render correctly-sized (its visual boundaries fall where you'd expect).

[...]

PR #4935, opened 2025-07-04
Use q-scroll_area for `ui.log` to preserve a fixed height

falkoschindler:
### Motivation

In #4925 we noticed that the height of a `ui.log` element can be affected by its content, which is unexpected and different than, e.g., for `ui.scroll_area`:
```py
with ui.column().classes('w-60 items-stretch h-72'):
    ui.label('h-24').classes('border h-24')
    with ui.column().classes('h-full'):
        log = ui.log().classes('h-full')

with ui.column().classes('w-60 items-stretch h-72'):
    ui.label('h-24').classes('border h-24')
    with ui.column().classes('h-full'):
        scroll = ui.scroll_area().classes('h-full border')

def add_content():
    log.push(time.time())
    with scroll:
        ui.label(time.time())

ui.button('Log', on_click=add_content)
```

### Implementation

The problem seems to be that `ui.log` doesn't wrap its content in multiple layers like `ui.scroll_area` does. To keep things simple, this PR implements `ui.log` based on `<q-scroll_area>` instead of `<div>`. This way we only need to adjust the CSS rules a little bit and now the layout behavior of `ui.log` is identical to `ui.scroll_area`.

### Caveat

I don't think that the different underlying Vue component is a breaking change.
**But** now `ui.log` collapses if placed inside a container with an unspecified width:
```py
ui.log()  # ok

with ui.column():
    ui.log()  # collapses

with ui.column().classes('w-60'):
    ui.log()  # ok
```

Therefore I'm not sure if it is safe to release this PR before 3.0.

### Progress

### DIFF
--- a/tests/test_log.py
+++ b/tests/test_log.py
@@ -28,7 +28,7 @@ def test_log_with_newlines(screen: Screen):
 
 
 def test_replace_log(screen: Screen):
-    with ui.row() as container:
+    with ui.row().classes('w-full') as container:
         ui.log().push('A')
 
     def replace():


falkoschindler:
We agreed to wait for NiceGUI 3.0 because this PR can break existing layouts.

davetapley:
I would love to use this in my 2.x app. Before I fork and cherry-pick: Is there a better way?

I know of `ui.css`, but the `super()` call looks like a dead end?

falkoschindler:
Hi @davetapley,

You can extract your own log element like this:
```py
ui.add_css('''
    .my-nicegui-log {
        outline: 1px solid rgba(127, 159, 191, 0.15);
    }
    .my-nicegui-log .q-scrollarea__container {
        scroll-padding-bottom: 0.5rem;
        scroll-snap-type: y proximity;
        background-color: rgba(127, 159, 191, 0.05);
    }
    .my-nicegui-log .q-scrollarea__content {
        white-space: pre;
        font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
        padding: 0.5rem;
    }
    .my-nicegui-log .q-scrollarea__content > :last-child {
        scroll-snap-align: end;
    }
''', shared=True)

class MyLog(ui.element, default_classes='my-nicegui-log'):

    def __init__(self, max_lines: Optional[int] = None) -> None:
        super().__init__('q-scroll-area')
        self.max_lines = max_lines

    def push(self, line: Any, *,
             classes: Optional[str] = None,
             style: Optional[str] = None,
             props: Optional[str] = None) -> None:
        for text in str(line).splitlines():
            with self:
                label = ui.label(text)
                if classes is not None:
                    label.classes(replace=classes)
                if style is not None:
                    label.style(replace=style)
                if props is not None:
                    label.props.clear()
                    label.props(props)
        while self.max_lines is not None and len(self.default_slot.children) > self.max_lines:
            self.remove(0)

log = MyLog(max_lines=10).classes('w-full h-20')
ui.button('Log time', on_click=lambda: log.push(datetime.now().strftime('%X.%f')[:-5]))
```

Issue #4927, opened 2025-07-02
`ui.select` dropdown auto selects first element in list when typed in input


BaccanoMob:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
def handle_key(e: KeyEventArguments):
    if not e.action.repeat and e.action.keydown:
        keylabel.text = " ".join([keylabel.text, str(e.key)])

ui.keyboard(on_key=handle_key, ignore=[])

keylabel = ui.label()
label = ui.label()
ui.select([], new_value_mode="add-unique", multiple=True).props(
    "use-chips"
).bind_value_to(label, "text", lambda x: ", ".join(x))
```

### Description

https://github.com/user-attachments/assets/caf50046-28c7-4995-8c3e-962dc5899036

Note that after every three letters I pressed `enter` and only the first element of the list autoselects from the dropdown on enter. Expected behavior is how `def` and `ghj` values behave which is not seen for `abc` aka the first element of the list.

https://github.com/user-attachments/assets/1ad37677-cc27-49cc-99af-a1bc1799c1fd

I pressed up arrow only for abc to remove the highlight on the dropdown.

https://github.com/user-attachments/assets/ff95f151-90e5-4844-8898-24c2bfe0682c

If replace props as `"hide-dropdown-icon popup-content-class='hidden' use-chips"` to hide the dropdown. The problem still presists. First half is without pressing up arrow and the latter pressing up arrow to avoid toggling.

This kind of degrades the purpose of `add` and `add-unique` props for the input fields where the value from input must be treated as new rather than just toggle.

Hope I sufficiently explained the issue!

### NiceGUI Version

2.20.0

### Python Version

3.12

### Browser

Chrome

### Operating System

Linux

falkoschindler:
Thanks, @BaccanoMob.

It took me quite a while, but I think I understand what you mean:

```py
ui.select(['x', 'y', 'z'], value=['x', 'y', 'z'], new_value_mode='add-unique', multiple=True).props('use-chips')
```
- When focussing the `ui.select` and typing "x", the dropdown shows a _highlighted_ "x" item.
- When focussing the `ui.select` and typing "y" or "z", the dropdown shows an _unhighlighted_ "y" or "z" item.

[...]

PR #4936, opened 2025-07-04
Generating multiple values for `ui.select`

BaccanoMob:
### Motivation

This PR allows users to a value delimited by comma [or one of these `,;|、،`] to split the values. I used the delimiting chars `,;|` from Quasar docs and further included `、` (Chinese and Japanese comma) and `،` (Arabic comma) to the list incase the GUI is used in other language. 

This feature is only available for new_value_mode `add` and `add-unique`.

It also uses #4916 to resolve duplicate values with `add-unique`. Partly improved upon to handle when the value is delimited.

Closes #4931 #4896

### Implementation

- Pretty much splits the value into list of strings before the usual process. And only works if the value is not None (when removing choice condition).

Limitations:

- I excluded `new_value_mode` as None, as input did not allow me to choose, other than using the dropdown as filter.
- I excluded `new_value_mode`'s `toggle` because it will include more for loops to remove the value in pairs which could lead to severe performance degradation on long lists. I found it difficult to toggle (removing in pairs) if duplicate values are involved when split.

### Progress

### Problems I am facing

I could not get the pytest `test_add_new_values_delimited` working for `add-False` condition for some reason while `add-unique` works completely fine. Locally when I test in my browser it shows the expected output (also tried copy pasting to the test as well) but not in pytest.

List of things I tried to make pytest work so far:

- Checked if value is not splitting - It does split
- Checked if `value` or `options` can be seen (excluding list) - It is seen (The only time pytest passed)
- Checked if value/options are behaving like `add-unique` - No it does not give the same result

#4927 has a working video captured from my browser.

I have trouble setting up pytest as well so I had to rely on github actions for the tests. Locally I test by editing `select.py` file. Also, I do not have experience with pytests with nicegui and barely tried my best by mimicking based on #4916.

@falkoschindler Could I please get some assistance with this? I really could not find a reason why this does work.

I didn't know about adding documentation till this PR, will update if needed after fixing up pytests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,8 +50,8 @@ jobs:
           poetry run pip install -r tests/requirements.txt
           # try fix issue with importlib_resources
           poetry run pip install importlib-resources
-      - name: test startup
-        run: poetry run ./test_startup.sh
+      # - name: test startup
+      #   run: poetry run ./test_startup.sh
       - name: setup chromedriver
         uses: nanasess/setup-chromedriver@v2.3.0
       - name: pytest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
       - name: setup chromedriver
         uses: nanasess/setup-chromedriver@v2.3.0
       - name: pytest
-        run: poetry run pytest
+        run: poetry run pytest tests/test_select.py
 
   slack:
     needs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,12 +50,12 @@ jobs:
           poetry run pip install -r tests/requirements.txt
           # try fix issue with importlib_resources
           poetry run pip install importlib-resources
-      # - name: test startup
-      #   run: poetry run ./test_startup.sh
+      - name: test startup
+        run: poetry run ./test_startup.sh
       - name: setup chromedriver
         uses: nanasess/setup-chromedriver@v2.3.0
       - name: pytest
-        run: poetry run pytest tests/test_select.py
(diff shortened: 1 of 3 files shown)


BaccanoMob:
Ok I found the cause of pytest fail for `add`... I forgot the for loop and the final value should have had 6 elements instead of 3 so it ended up failing. `add-unique` passed because it just removed the duplicates leaving 2 elements with/without for loop. Ill update the code in a bit

BaccanoMob:
Everything should be working fine now.

falkoschindler:
Thanks for this pull request, @BaccanoMob! It looks like a handy feature.

I'll try to merge #4916 first before reviewing this one though.

falkoschindler:
Hi @BaccanoMob,

We are sceptical about adding additional complexity to `ui.select` without a clear benefit for the user. Our concerns are:
- The user does not expect typing "abc,def"+Enter to work because this functionality is not common.
- The user might want to include "," or another separator within an item.
- The user could simply type "abc"+Enter+"def"+Enter, resulting in the same number of keystrokes.

Therefore I'm closing this PR for now. Feel free to create a feature request to discuss pros and cons with the community and collect upvotes to re-address this PR.

BaccanoMob:
Ill just respond for the concerns from my POV (prolly useful for anyone looking through this PR). 

- My use case was copy pasting a list of values and having this feature it would reduce time to copy paste one by one.
- I did think have use_delimiter as a string rather than bool. But ultimately, went with bool since I thought only commas and similar symbols should do.

But then again I most likely wont use this feature anyway because of the bug with the dropdown and `new_value_mode`. So I'm cool with closing this PR. Thanks for reviewing @falkoschindler !

PR #4938, opened 2025-07-04
Remove cyclic references within clients and elements

falkoschindler:
### Motivation

We noticed that cyclic references between `Client`, `Element`, `Slot`, `Props`, `Classes`, `Style`, and `Tailwind` classes prevent simple reference counting and require the garbage collector to find deleted instances with a more costly search. This PR investigates if we can break all cycles with weak references.

Demonstration:
```py
import weakref
from nicegui import ui

elements: weakref.WeakSet = weakref.WeakSet()

with ui.card() as card:
    for _ in range(10):
        elements.add(ui.element())

card.clear()
ui.label(f'{len(elements)} elements in memory')  # should be 0

ui.run()
```

### Implementation

Instead of storing hard references like `self.client = client` in element.py, this PR only stores a weak reference `self._client = weakref.ref(client)` and provides a property `self.client` for read access.

A new pytest counts references before the garbage collector has a chance to run, proving that the weak references are effective.

### Pros

- The garbage collector has much less to do.
- The read-only properties come with the additional benefit of preventing write access.

### Con

- Calling the property getter and resolving the reference could be more costly than the direct attribute access before.

### Progress

- [x] Fix similar cycles for `ui.scene` and its 3D objects. --> Not needed apparently. See new pytest.
- [x] Fix similar cycles for `Client` and its `Outbox`.
- [x] Fix ElementFilter tests.
- [x] Are we sure the pros outweigh the cons? --> Yes, according to [ChatGPT](https://chatgpt.com/share/6867d2e8-4e98-800c-82af-36ea97b03ce1).


nicegui/outbox.py
```diff
@@ -76,7 +92,8 @@ async def loop(self) -> None:
                     except (TimeoutError, asyncio.TimeoutError):
                         continue
 
-                if not self.client.has_socket_connection:
+                client = self.client
+                if not client or not self.client.has_socket_connection:
```
falkoschindler:
```suggestion
                if not client or not client.has_socket_connection:
```

burnpanck:
Question: I assume the `Client` still is being kept alive by the framework until the actual client disconnects? I think this PR broke my application, and I'm trying to figure out what exactly did. My application is an interface to some local hardware, so there are per-client (user-interface) resources, and global resources. My `@ui.page("/")` builds a number of elements that reflect global resource state. A callback is attached to the global resources (not using nicegui-binding but a custom reactive library) keyed on a weak reference to the element, so that these callbacks can update UI state of live clients, but get dropped when clients disconnect. With this change, somehow, these callbacks start to throw `RuntimeError: The client this element belongs to has been deleted.` when they try to update their target element, which still is live - otherwise they would not be called. (I tried to make sure that the callback itself doesn't retain a hard reference to the element in its closure; it receives a hard reference to the "owner" as an argument, materialised by the reactive library). The application does not use multi-threading, and there should not be any suspend-point between materialising the element and the callback trying to update the element. Do you see any other reason why that the client could be dead while the elements remain alive for some time? I was hoping that I don't have to explicitly manage client lifetimes in my code...

falkoschindler:
@burnpanck It is very hard to tell the reason for this RuntimeError without seeing any code. Like with issues #4948 and #5014 the only way forward is to extract a minimum reproducible code example (preferably in a new thread) and see if it is a bug in NiceGUI or a problem with the application code.

burnpanck:
Ignore my previous (deleted) message. There may be a bug or unintended behaviour in nicegui, and I may have reproducer of that: #5052

PR #4940, opened 2025-07-04
New Element - InputChips

BaccanoMob:
### Motivation

Note that #4927 can not be resolved since Quasar framework has same behavior as well. The root cause lies in the dropdown itself which is directly coupled with the behavior of `options`. Setting props `hide-dropdown-icon popup-content-class='hidden' use-chips` does not help either. You may accidentally toggle the first element if a substring is typed in (#4927 shows some clips on this behavior). This will be troublesome in long lists.

`ui.select` is at the moment can be an alias for `Dropdown Selection`. Since options is more or less the backbone of ui.selecct,  making the dropdown optional may break ui.select.

Therefore, I created a new element `InputChips` to overcome the issue for users who want `new_value_mode` and safely use it. Using this element will add values to list with input and removing can be done with chip's clear button rather than dropdown. Having an element like this ready to use would be helpful for myself (because this was exactly what I wanted to create a form) or users like #1649.

### Implementation

- It replaces `ChoiceElement` with `ValueElement` to remove the dependency on options/dropdown and the JS file contains barebones of Qselect.
- It converts the value label dict from Qselect to list of values and compares with the new values. So this simplifies backend work done to keep the values validated according to `new_value_mode` set for the element.
- Unlike my previous PR, `use_delimiter` works for all cases since this only deals with list.

Partly I feel there will be better performance for larger list since this element quickly deals with the list of values in transit while ui.select have to run `_hande_new_value` after. Remember that `_hande_new_value` runs multiple times if `use_delimiter` is True, for cases like 2+ values are added in one go. (Though this will be the case if #4936 or this PR is merged at some point to use `use_delimiter`)

### Progress

I created this draft PR before starting pytest/docs, to get some opinions/thoughts of this feature whether it will be good addition or not.

### DIFF
--- a/nicegui/elements/input_chips.js
+++ b/nicegui/elements/input_chips.js
@@ -1,0 +1,9 @@
+export default {
+  template: `
+    <q-select
+      ref="qRef"
+      v-bind="$attrs"
+    >
+    </q-select>
+  `,
+};
--- a/nicegui/elements/input_chips.py
+++ b/nicegui/elements/input_chips.py
@@ -20,12 +20,11 @@ class InputChips(LabelElement, ValidationElement, ValueElement, DisableableEleme
                  use_delimiter: bool = False,
                  validation: Optional[Union[ValidationFunction, ValidationDict]] = None,
                  ) -> None:
-        """Dropdown Selection
+        """Input Chips
 
         This element is based on Quasar's `QSelect <https://quasar.dev/vue-components/select>`_ component.
-
-        The options can be specified as a list of values, or as a dictionary mapping values to labels.
-        After manipulating the options, call `update()` to update the options in the UI.
+        This variant focuses solely on using chips without dropdown menu which is suitable for longer lists.
+        Hence, it acts as an input with list of chips rather than a dropdown of choices.
 
         You can use the `validation` parameter to define a dictionary of validation rules,
         e.g. ``{'Too long!': lambda value: len(value) < 3}``.
@@ -68,7 +67,11 @@ class InputChips(LabelElement, ValidationElement, ValueElement, DisableableEleme
             new_value = next((arg for arg in e.args if isinstance(arg, str)), None)
             split_args = []
             if new_value is not None:
-                split_args = [value.strip() for value in re.split(r'[,;|]+', new_value) if len(value)> 0] if self.use_delimiter else [new_value]
+                split_args = (
+                    [value.strip() for value in re.split(r'[,;|]+', new_value)if len(value)> 0]
+                    if self.use_delimiter
+                    else [new_value]
+                )
 
             args = [arg['label'] for arg in e.args if isinstance(arg, dict)]
             mode = self._props['new-value-mode']
--- a/nicegui/elements/input_chips.py
+++ b/nicegui/elements/input_chips.py
@@ -1,0 +1,92 @@
+import re
+from typing import Any, Literal, Optional, Union
+
+from ..events import GenericEventArguments, Handler, ValueChangeEventArguments
+from .mixins.disableable_element import DisableableElement
+from .mixins.label_element import LabelElement
+from .mixins.validation_element import ValidationDict, ValidationElement, ValidationFunction
+from .mixins.value_element import ValueElement
+
+
+class InputChips(LabelElement, ValidationElement, ValueElement, DisableableElement, component='input_chips.js'):
+
+    def __init__(self,
+                 value: Any = None,
+                 *,
+                 label: Optional[str] = None,
+                 on_change: Optional[Handler[ValueChangeEventArguments]] = None,
+                 new_value_mode: Literal['add', 'add-unique', 'toggle'] = 'toggle',
+                 clearable: bool = False,
+                 use_delimiter: bool = False,
+                 validation: Optional[Union[ValidationFunction, ValidationDict]] = None,
+                 ) -> None:
+        """Dropdown Selection
+
+        This element is based on Quasar's `QSelect <https://quasar.dev/vue-components/select>`_ component.
+
+        The options can be specified as a list of values, or as a dictionary mapping values to labels.
+        After manipulating the options, call `update()` to update the options in the UI.
+
+        You can use the `validation` parameter to define a dictionary of validation rules,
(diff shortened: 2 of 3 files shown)


falkoschindler (commented):
Thanks a lot for this pull request, @BaccanoMob!

I like the idea of a `ui.input_chips` element. Also it seems like a good choice to start with a separate element instead of deriving from `ui.select`.

I just reviewed the code and found the following simplifications:

- We don't need the "empty" input_chips.js. We can directly use "q-select" in input_chips.py.
- Quasar is already handling the new-value-mode for us. So we don't need to update the value ourselves. The only difference is that we want to default to `[]` instead of `None` after clearing the element. At least this is how `ui.select` behaves, so it is a good idea to adopt the behavior.
- I decided to remove the delimiter argument. This controversial feature complicates the implementation quite a bit and still doesn't capture all cases correctly (e.g. `if len(value)> 0` didn't consider the _stripped_ `value`). As it turns out, the users can delimit values themselves:
    ```py
    def split_values(e: events.ValueChangeEventArguments):
        for value in e.value[:]:
            if ',' in value:
                e.value.remove(value)
                e.sender.value = e.value + value.split(',')
    
    c = ui.input_chips(on_change=split_values)
    ui.label().bind_text_from(c, 'value', backward=', '.join)
    ```
    Therefore I suggest to add this demo to the documentation and leave it up to the user to add any details on demand.

What do you think? If you agree with my suggestion, you can go ahead and add a short pytest and a documentation page. 🙂

BaccanoMob:
@falkoschindler I looked at your commit, and I agree with the changes and it actually makes more sense than have a built in delimiter. Since I also use flutter, I got confused with `onChanged` from flutter and `on_change` from nicegui.  `onChanged` keeps track of input itself so I assumed `on_change` would do something similar.  Maybe `on_submit` would be more close to its purpose than `on_change` (This would be a breaking change i guess, so you can have thoughts on this later).

I looked at the example and gave it try. I found couple of issues and solving it could improve user experience with `ui.input_chips`.

## Points to discuss

### Looping

```python
for value in e.value[:]:
```
I usually tend to avoid looping if possible. While delimiting is a simple case, what if the value is manipulated or validated or compares thorough database, etc. Looping every single element in the list is a waste of resource since you only have to process the only the new value. I thought of 2 solutions.

#### Last element

When you add a new value, the last element is the new value. Though it is not always the case, like in removing an element.

```python
new_value = None if len(e.value) == 0 else e.value[-1] 
```

This way you can save the iteration to a singluar value.

#### New value from ui.input_chips

Personally, I want to use the below approach where ui.input_chips keeps track of new_value that is entered. You can use `get_new_value` to get new value if something was entered or None if something was deleted. This will be more verbose and user friendly. Alternatively, you can also get the `new_value` directly as a list (in case Quasar framework makes some update to have multiple values) and `get_new_value` is named so to be singular since at the moment it is only one value. Therefore, nothing should break ideally even if Quasar is updated.

```python
    def _event_args_to_value(self, e: GenericEventArguments) -> Any:
        self.new_value =  [val for val in e.args if val not in self._old_values]
        self._old_values = e.args
        return e.args or []

    def get_new_value(self):
        return None if len(self.new_value) == 0 else self.new_value[0]
```

### Appending values

```python
e.sender.value = e.value + value.split(',') #  e.value = e.value + value.split(',')  # this also works btw!
```

Above appends the value regardless of `new_value_mode` (aka treats like `add`). This will make users to write code similar to below if they want to check against `new_value_mode`.

```python
new_value = None if len(e.value) == 0 else e.value[-1]

if new_value is not None and "," in new_value:
    new_value_mode = e.sender.props["new-value-mode"]
    e.value.remove(new_value)
    for split_val in new_value.split(","):
        split_val = split_val.strip()
        if split_val != "":
            # Below is the logic to append according to new_value_mode
            if new_value_mode == "add":
                e.value.append(split_val)
            else:
                if split_val not in e.value:
                    e.value.append(split_val)
                elif new_value_mode == "toggle":
                    e.value.remove(split_val)
```

Well you could technically argue that the user should write code to the `new_value_mode` of their choosing but the reusability of this function would be less. So the solution I came up was to have another helper function similar to `get_new_value` which uses the above append value logic internally.

```python
    def append_value(self, value: str):
        new_value_mode = self._props['new-value-mode']
        if new_value_mode == "add":
            self.value.append(value)
        else:
            if value not in self.value:
                self.value.append(value)
            elif new_value_mode == "toggle":
                self.value.remove(value)
```

`append_value` would append the value according to the `new_value_mode` that is currently set.

This way `ui.input_chips` would be fully utilize and exhibit `new_value_mode`. 

### Example if above solutions implemented

```python
new_value = e.sender.get_new_value()
if new_value is not None and "," in new_value:
    e.value.remove(new_value)
    for split_val in new_value.split(","):
        split_val = split_val.strip()
        if split_val != "":
            e.sender.append_value(split_val)
```

You can see that the example, is more verbose and would be user friendly. Below would more of a generic template where user can just change the `doing_some_magic_on` function to their needs and if `updated_value` becomes a list, then they need a for loop to append the values.

```python
new_value = e.sender.get_new_value()
if new_value is not None:
    e.value.remove(new_value)
    updated_value = doing_some_magic_on(new_value)
    e.sender.append_value(updated_value)
```

## TLDR (?)

@falkoschindler Let me know the thoughts on `get_new_value` and `append_value` functions being built into `ui.input_chips`. The function names are self explanatory which is also a plus point. The last code snippet should give a rough idea on what it does if implemented. 

Also I am thinking of making `append_value` accept `list[str]` or `str` (so I guess it becomes `append_values`) and include the logic of how the value is handled in init function that way the user does have to worry about looping the new values.

If its good to go, Ill add the above functions and start writing pytests/docs for `ui.input_chips`.

BaccanoMob:
Inside init

```python
        if value is None:
            value = []
        elif not isinstance(value, list):
            value = [value]
        else:
            value = value[:]  # NOTE: avoid modifying the original list which could be the list of options (#3014)
```
I dont know if this was overlooked or necessary. Since options is no more in this version of select, I can remove the `else` statement right? or is there a reason I should keep this as is?

BaccanoMob:
@falkoschindler I understand your points, so I reverted the changes. I think its good to go now.

PR #4943, opened 2025-07-07
Support async nested sub pages

rodja:
### Motivation

As described in https://github.com/zauberzeug/nicegui/pull/4821#issuecomment-3017729418, when child `ui.sub_page` elements are added async or via background task to the element tree, they get not rendered.

### Implementation

This PR changes the core architecture: instead of root sub pages elements rendering all their children, now each sub pages element takes care of it's own rendering. That way, even when added via background task it just creates it's content. For this to work the element determines its own contribution of the complete path (with prefix, root, parent sub pages elements, ...) to decide whether re-rendering is necessary or not.

### Progress

  - [x] try to use `_root_path` instead of complicated anchestor walking in `_find_matching_path`
  - [x] make js code global in `sub_pages.js` (to get rid of `add_head_html`)
  - [x] check if example still works and is up to date
- [x] Pytests have been added/updated/extended.


rodja:
I think it's best to drop the feature of binding the (relative) url on a sub pages element for now. It leads to quite some complications with the chain of updates and general URL modification responsibility. Especially if you consider that there might be multiple sub pages elements side by side.

rodja:
Because the architecture changed greatly, we decided the review will be done after merging this and the other stacked branches into #4821.

Issue #4837, opened 2025-06-04
ui.table row will not accept list - memory build up and crash.


WilliamAughenbaugh:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

columns = [
{'name':'col1', 'label': 'Column 1', 'field': 'col1', 'required': True},
{'name':'col2', 'label': 'Column 2', 'field': 'col2',},
{'name':'col3', 'label': 'Column 3', 'field': 'col3', 'required': True},
]

rows = [
{'col1': '1 Data point', 'col2' : '1 Data point', 'col3': '1 Data point'}
{'col1': '1 Data point', 'col2' : '1 Data point', 'col3': '1 Data point'}
{'col1': '1 Data point', 'col2' : '1 Data point', 'col3': '1 Data point'}
{'col1': '1 Data point', 'col2' : '1 Data point', 'col3': ['Point 1', 'Point 2', 'Point 3']}
]

ui.table(columns=columns, rows=rows, row_key='name')

ui.run()

# will work if list is cast to string using .__str__()
#{'col1': '1 Data point', 'col2' : '1 Data point', 'col3': ['Point 1', 'Point 2', 'Point 3'].__str__()}
```

### Description

When attempting to insert a list of items within a single field  of a row, and loading the ui.table, the browser ends up eating up a large amount of memory before ultimately crashing.

When removing the list, and inserting just a single item, loads as expected, with the expected speeds dependent upon the data amount.

Loading the 'ui.table' is the crucial step in repeating this issue.

Casting the list to string does resolve the issue - unsure if this is a bug or a feature request but considering it leads to a crashed browser I presumed bug.

### NiceGUI Version

2.19.0

### Python Version

3.13.3

### Browser

Chrome

### Operating System

Windows

falkoschindler:
Hi @WilliamAughenbaugh,
This bug has been discussed and partly solved in #2744. Here is a quick summary generated with ChatGPT:

---

| Aspect                            | Details                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
[...]

PR #4949, opened 2025-07-10
Fix: Handle lists in ui.table cells

pratikscfr:
### Motivation

This PR solves the problem where ui.table would experience significant memory buildup and eventually crash the browser if a Python list was provided as a value for a cell in a manually defined rows prop. The underlying Quasar table component does not natively render Python list objects directly within a standard table cell, leading to this instability.

This issue is documented in [Github Issue #4837](https://github.com/zauberzeug/nicegui/issues/4837).

This PR ensures that such lists are gracefully handled by default, preventing the crash and providing a readable string representation of the list content in the table cell.

### Implementation

The implementation modifies the _normalize_columns method within the Table class in nicegui/elements/table.py.

The core idea is to automatically wrap the field accessor for each column with a new lambda function if the original field is a simple string key (i.e., not an existing callable/formatter provided by you).

This wrapper function checks the type of the data retrieved for a cell using the field key:

If the data is an instance of list, it converts the list into a comma-separated string (e.g., ['Apple', 'Banana'] becomes "Apple, Banana").
If the data is not a list, it's returned as-is.
This conversion happens before the data is passed to the Vue/Quasar frontend, ensuring that the QTable component only receives data types it can readily display (primarily strings, numbers, booleans for basic cells).

This approach was chosen because:

- It directly addresses the root cause – the incompatibility of raw list objects with Quasar's default cell rendering.
- It provides a sensible default behavior without requiring you to manually convert lists to strings for basic display.
- It doesn't interfere with more advanced user customizations where a field might already be a callable or when users employ slots for custom cell rendering (in which case they have full control over how list data is presented).
- The trade-off is a slight overhead for type checking and string conversion for list data, but this is minimal and significantly outweighs the previous crashing behaviour.

This fix complements PR #4775, which handled list-like data within Pandas DataFrames for ui.table.from_pandas(); this PR addresses lists in manually defined rows for ui.table().

### Progress

   - A new test test_table_with_list_in_cell was added to tests/test_table.py to verify this specific scenario

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -365,3 +365,21 @@ def test_columns_from_df(screen: Screen, df_type: str):
     screen.click('Update cars with columns')  # updated columns via parameter
     screen.should_contain('Hyundai')
     screen.should_contain('i30')
+
+
+def test_table_with_list_in_cell(screen: Screen):
+    columns_with_list = [
+        {'name': 'name', 'label': 'Name', 'field': 'name', 'required': True},
+        {'name': 'items', 'label': 'Items', 'field': 'items'},
+    ]
+    rows_with_list = [
+        {'id': 0, 'name': 'Alice', 'items': ['Apple', 'Banana']},
+        {'id': 1, 'name': 'Bob', 'items': ['Cherry', 'Date', 'Elderberry']},
+    ]
+    ui.table(columns=columns_with_list, rows=rows_with_list)
+
+    screen.open('/')
+    screen.should_contain('Alice')
+    screen.should_contain('Apple, Banana')
+    screen.should_contain('Bob')
+    screen.should_contain('Cherry, Date, Elderberry')


falkoschindler:
Hi @pratikscfr,

To be honest, I don't understand this pull request. Do be more specific:

- I don't understand how it is _supposed_ to work. The `columns` list is supposed to contain dictionaries, not callables. Where and how should callables be evaluated?
- Consequently, the test fails with an assertion error.
- The test code also doesn't work when copied to a plain NiceGUI test app.

Can you, please, explain your intention with this PR? Or is it AI-generated?

pratikscfr:
Hi @falkoschindler, yes, I was trying to test the AI generated code. Sorry, this didn't work.  I am also working on this by myself. Will raise the changes separately since this is not working. Once again sorry for the AI code .

PR #4966, opened 2025-07-17
Add dynamic icon color to ColorInput in preview mode

gabriblas:
### Motivation

When passing `preview=True` to `ColorInput`, the button icon and background are separated by adding a shadow. If applied, this PR modifies this behavior to changing the icon color dynamically based on the button background color.

|Before|After|
|--|--|
|before_dark before_light | after_light after_dark|

### Implementation

The implementation uses the `rgb_to_yiq` function from the `colorsys` module from Python's standard library to calculate the luminance of the chosen color. The YIQ color space slightly biases the choice towards a white icon for vivid colors.

### Progress

---

Feature request: #4965

### DIFF
--- a/nicegui/elements/color_input.py
+++ b/nicegui/elements/color_input.py
@@ -1,3 +1,4 @@
+import re
 from colorsys import rgb_to_yiq
 from typing import Any, Optional
 
@@ -7,6 +8,9 @@ from .color_picker import ColorPicker as color_picker
 from .mixins.disableable_element import DisableableElement
 from .mixins.label_element import LabelElement
 from .mixins.value_element import ValueElement
+
+HEX_COLOR_PATTERN_6 = re.compile(r'^#([0-9a-fA-F]{6})$')
+HEX_COLOR_PATTERN_3 = re.compile(r'^#([0-9a-fA-F]{3})$')
 
 
 class ColorInput(LabelElement, ValueElement, DisableableElement):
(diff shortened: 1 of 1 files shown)


falkoschindler:
Thanks a lot for this pull request, @gabriblas!

Quick question: What is the purpose of the "inline" prop? It doesn't seem to be related with the new color mechanism.

gabriblas:
Hi @falkoschindler! My bad, I'm quite new to NiceGUI and Quasar and must've misinterpreted some other example somewhere. Indeed, the inline prop is not necessary!

gabriblas:
Yeah, I see. Removing padding, detecting and catching most color formats should be fairly straightforward, but handling named colors is trickier.

A possible solution could be using `ui.run_javascript` to call Quasar's [`getPaletteColor`](https://quasar.dev/quasar-utils/color-utils#helper-getpalettecolor), which converts all kinds of named colors (brand, palette, custom) into hexadecimal colors. However, it requires importing the Quasar Javascript library (which could be undesired for some users?) and sanitize the arbitrary string `event.value` before passing it to `ui.run_javascript`. Also, the documentation states "Note that the method below is not cheap to run, so use it with care". The basic demo I pasted below runs very smoothly, but what about larger projects?

All in all, I realize this is a lot for such a small detail, so I would understand if this gets closed as not planned.

Quick and dirty demo for `getPaletteColor`:

```python
from nicegui import ui
from collections import namedtuple
from itertools import cycle

ui.add_head_html(
    '<script src=" https://cdn.jsdelivr.net/npm/quasar@2.18.2/dist/quasar.umd.prod.min.js "></script>'
)


@ui.page("/")
def page():
    async def on_pick(event):
        # ! passing arbitrary strings to javascript is very unsafe
        res = await ui.run_javascript(
            f'window.Quasar.colors.getPaletteColor("{event.value}")'
        )
        ui.notify(res)

    colors = cycle(["red", "grey-10", "primary"])

    async def on_click(event):
        e = namedtuple("Event", ["value"])(next(colors))
        await on_pick(e)

    ui.color_input("Color", on_change=on_pick, preview=True)
    ui.button("Test", on_click=on_click)


ui.run()
```

falkoschindler:
Hm, full support for color names doesn't seem feasible. Waiting for a client-server roundtrip just to get the color value is a bit much. And it only works for Quasar colors ("red-4"), not CSS ("DarkRed") or Tailwind ("red-400").

But such color names don't really "work" on main either. Even though CSS colors are correctly applied as a background color (because the CSS property accepts both, color names and hex values), Tailwind and Quasar colors are completely ignored.

My main concern is that this PR raises exceptions for values that have been ignored previously. If we would sanity-check the value before converting color spaces we should be fine. Or we simply catch exceptions if they occur.

Kind of a second-order problem is that, even if this PR doesn't understand the value "white", it wouldn't know which text color to use in combination. a) If "white" is ignored as an invalid value, we don't have a contrast problem, but a breaking change wrt. to the main branch. b) If "white" is applied as background color like before, the icon might be invisible due to missing contrast.

Thinking about the "breaking change": The fact that entering a CSS color like "DarkRed" causes the button to turn dark red - is this really a feature? We could argue that this is undefined behavior for invalid color input. There is no value for the user in interpreting color names either way. The point of this element is to _pick_ a color and display its hex code.

Therefore I suggest to sanitize the `value` before determining its brightness and to simply ignore invalid values. (I was thinking about using the old implementation as a fallback, but it's probably not worth it.)

Issue #457, opened 2023-03-06
Unable to limit options in Date Input


kazaarun:
I am trying to limit date options in the Date Input element by passing a list of acceptable dates, as required by the _options_ prop in [QDate](https://quasar.dev/vue-components/date#limiting-options). But this has no effect on the component. If I understand correctly from the source code, the _props_ function can only parse the string input to a `dict[str, str]`.
Would it be possible to allow _props_ to parse lists or are there other ways to limit date options?

Example code:
```py
from nicegui import ui

date_options = ['2023-01-01', '2023-01-12', '2023-01-17']

ui.date(value='2023-01-01').props(f"options={date_options}")

ui.run()
```

falkoschindler:
Good question! As far as I can tell, there are multiple layers to the problem:

1. The argument `f"options={date_options}"` results in the string "options=['2023-01-01', '2023-01-12', '2023-01-17']" which is incorrectly parsed into the props dictionary. Values containing whitespace need to be wrapped in quotes: `f'options="{date_options}"'`.

2. This still converts the array into a string, but Quasar expects an array. We can workaround this limitation by setting the dictionary value directly:

    ```py
    d = ui.date(value='2023-01-01')
    d._props['options'] = ['2023-01-01', '2023-01-12', '2023-01-17']
    ```

3. But this still doesn't work and I haven't found out why. 🫤

falkoschindler:
Oh hey, I found a solution! It turns out the dates have to be in YYYY/MM/DD format:

```py
d = ui.date(value='2023-01-01')
d._props['options'] = ['2023/01/01', '2023/01/12', '2023/01/17']
```

You can even use the `.prop()` method. Here the options are a list of strings wrapped in double quotes to guide the props parser:

```py
ui.date(value='2023-01-01').props('''options="['2023/01/01', '2023/01/12', '2023/01/17']"''')
```

Or using the `date_options` list from the original post:

```py
date_options = ['2023/01/01', '2023/01/12', '2023/01/17']
ui.date(value='2023-01-01').props(f'options="{date_options}"')
```

PR #4969, opened 2025-07-20
allow props() to take a dict or list without quotes

lawrenceakka:
Occasionally, props can be an javascript array or object, eg the style props for [QScrollArea](https://quasar.dev/vue-components/scroll-area/).  It is not obvious how to set these props with NiceGui. Neither the docs nor the examples help very much, but there is some information in #457 and #706.  There are quite a lot of quotes involved!

Whilst it is possible to add to the `_props` list directly, this is not obvious, and feels like a bit of a hack.

This PR allows a list or dict to be passed to `Element.props()` in what feels like an intuitive fashion. It also updates the `PROPS_PATTERN` regex in `props.py` to use named capture groups, so the logic is clearer 

Includes additional pytests.  All existing tests pass.

Now its easy to make beautiful scrollbars like this 😉 :

scroll

```python
from nicegui import ui

with ui.scroll_area() as scroll:
    for i in range(10):
        ui.label("Commodo nostrud aliquip ad ad consequat aliquip reprehenderit ut. Ea in id est nostrud anim est" 
                    "amet. Minim commodo fugiat ullamco tempor nisi. Nulla consectetur eiusmod tempor mollit.Enim "
                    "exercitation ea culpa aliquip et elit nulla cupidatat id reprehenderit voluptate. Incididunt "
                    "ullamco do ipsum amet qui tempor fugiat ut. Officia esse ea anim magna minim id quis consequat "
                    "labore ad pariatur culpa. Sit ullamco minim velit aliquip ut pariatur irure. Laborum cillum est "
                    "tempor mollit incididunt id eu mollit elit incididunt ad minim. Aute exercitation ad non "
                    "adipisicing. Exercitation commodo et Lorem eiusmod proident elit sit minim laborum enim anim"
                    "ullamco non nostrud. \n")
    
scroll.props('thumb-style={"background": "linear-gradient(0deg, blue 0%, green 100%)", "borderRadius": "5px", "opacity": 1}')
scroll.props('visible bar-style={"background": "linear-gradient(0deg, red 0%, yellow 100%)", "borderRadius": "5px", "opacity": 1}')

ui.run()
```

### DIFF
--- a/nicegui/props.py
+++ b/nicegui/props.py
@@ -9,30 +9,45 @@ if TYPE_CHECKING:
     from .element import Element
 
 PROPS_PATTERN = re.compile(r'''
-# Match a key-value pair optionally followed by whitespace or end of string
-([:\w\-]+)          # Capture group 1: Key
-(?:                 # Optional non-capturing group for value
-    =               # Match the equal sign
-    (?:             # Non-capturing group for value options
-        (           # Capture group 2: Value enclosed in double quotes
-            "       # Match  double quote
-            [^"\\]* # Match any character except quotes or backslashes zero or more times
-            (?:\\.[^"\\]*)*  # Match any escaped character followed by any character except quotes or backslashes zero or more times
-            "       # Match the closing quote
+# Match a key or key-value pair optionally followed by whitespace or end of string
+# The value can be unquoted, or enclosed in quotes, or square or curly brackets
+(?P<key>[:\w\-]+)           # Capture group 1: Key
+(                           # Optional non-capturing group for value
+    =                       # An equals sign
+    (                       # followed by one of ...
+        (?P<double>         # a value enclosed in double quotes
+            "                   # Match double quote
+            [^"\\]*             # Match any character except quotes or backslashes zero or more times
+            (\\.[^"\\]*)*       # Match any escaped character followed by any character except quotes or backslashes zero or more times
+            "                   # Match the closing quote
         )
-        |
-        (           # Capture group 3: Value enclosed in single quotes
-            '       # Match a single quote
-            [^'\\]* # Match any character except quotes or backslashes zero or more times
-            (?:\\.[^'\\]*)*  # Match any escaped character followed by any character except quotes or backslashes zero or more times
-            '       # Match the closing quote
+        |                   # or
+        (?P<single>         # a value enclosed in single quotes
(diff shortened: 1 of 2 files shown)


falkoschindler:
Hi, @lawrenceakka! Thank you so much for this PR!

The ability to write
```py
scroll.props('thumb-style={"background": "red"}')
```
instead of
```py
scroll.props('''thumb-style='{"background": "red"}' ''')
```
is pretty nice. It would improve the readability quite a bit in some cases.

It is, however, a breaking change. Consider this hypothetical example which won't work after this PR:
```py
x.props('left-symbol={ right-symbol=}')
```

Therefore this PR needs to wait for the upcoming release of version 3.0.

While thinking about your PR, I came up with an alternative simplification demonstrated in PR #4978. The core idea is to update the element as soon as the props dictionary changes, e.g. by calling `props.update()`:
```py
scroll.props.update(color='blue')
scroll.props.update({'thumb-style': '{"background": "red"}'})
```

What do you think? It requires less complex strings and no breaking change. But somehow tests are still failing and I haven't figured out why.

lawrenceakka:
Hi @falkoschindler ,

Of course, this can wait until v3!

But ```python x.props('left-symbol={ right-symbol=}')```

does not work even under the existing code: 

```python
>>> from nicegui.props import Props
>>> print(Props.parse('left-symbol={ right-symbol=}'))
{}
```

So I don't think the PR would break existing code - it just fails in a different way.

However, with the PR applied, the following does work (but again fails under existing code):

```python
>>> from nicegui.props import Props
>>> print(Props.parse('left-symbol="{" right-symbol="}"'))
{'left-symbol': '{', 'right-symbol': '}'}
```

I could take a look at #4978 next week (I think) and see why it is failing.  Are you suggesting it as an alternative to `x.props()`, or as something else?

falkoschindler:
@lawrenceakka Oh, you're right! I forgot that parentheses, brackets and braces are not treated as part of a prop value without enclosing quotes. So it indeed wouldn't be a breaking change. 🙂

> Are you suggesting it as an alternative to `x.props()`

Yes, I thought of it as an alternative. But given the fact that `Props.parse('left-symbol={ right-symbol=}')` never worked, your PR #4969 is useful independent of PR #4978.

PR #4978, opened 2025-07-22
Make props, classes and style observable to send updates automatically

falkoschindler:
### Motivation

When reviewing PR #4969, we noticed that it would be handy if the props dictionary would automatically update the frontend when changed:
```py
b = ui.button('test', on_click=lambda: b.props.update(icon='face'))
```

### Implementation

This PR replaces the `dict` base class with an `ObservableDict` and calls a new `_update` method. In order to avoid keeping strong reference to `element`, this method needs to use the weak reference `self._element`.

I'm marking the PR as draft because we should carefully evaluate memory and performance, add pytests and documentation. Although the change shouldn't break anything, it might be a good idea to release this feature in 3.0.

### Progress

- [x] Introduce `@helpers.prevent_recursion`
- [x] Existing pytests are passing.
- [x] This PR breaks https://nicegui.io/documentation/table#show_and_hide_columns [BREAKING CHANGE].
- [x] The condition `if self._send_update_on_value_change:` might be ineffective now.
- [x] `test_apply_format_on_blur` is red


falkoschindler:
Now I understand why many tests are running into infinite recursions:

Take `ui.plotly` for example. It defines a custom `update` method like this:
```py
def update(self) -> None:
    self._props['options'] = self._get_figure_json()
    super().update()
```

But modifying `self._props` calls `self.update()`, causing an infinite recursion.

Triggering a recursion is already unfortunate. But why is it infinite? It turns out, our observable collections call change handlers even if the value actually doesn't change. So even if `self._props['options']` is assigned the same dictionary, the `update` method is called again and again. We could change the behavior of observables to compare old and new values before calling change handlers. But I'm worried about the performance implication.

NiceGUI itself could cleverly avoid recursions within its `update` methods. But this PR would still break user code in a pretty nasty way.

falkoschindler:
I just tested a new `_no_updates` context manager:
```py
def update(self) -> None:
    with self._no_update():
        self._props['options'] = self._get_figure_json()
    super().update()
```

But it doesn't resolve the infinite recursion because the update to `self._props` calls `update()` which again updates `self._props`. The fact that `super().update()` is ineffective within the `_no_update` context doesn't help because we don't even reach `super().update()`. So we also need to add early exits to such derived `update` methods, which isn't very nice...

```py
if not self._updating:
    return
with self._no_update():
    self._props['options'] = self._get_figure_json()
super().update()
```

lawrenceakka:
@falkoschindler Just a thought - and I haven't tested this yet because I am on the road - but would it be better/more efficient just to change [ObservableCollection._handle_change()](https://github.com/zauberzeug/nicegui/blob/41543750be4b3d705eb2ef346b2fcc91e501c27c/nicegui/observables.py#L34C5-L37C93) to make sure that the change_handlers are not called if an update is already in progress? Something like:

```python
def _handle_change(self) -> None:
    if self._updating:
        return
    self._updating = True
    self.last_modified = time.time()
        for handler in self.change_handlers:
            events.handle_event(handler, events.ObservableChangeEventArguments(sender=self))
    self._updating = False
```

This might be complete nonsense - I didn't get much sleep yesterday!

falkoschindler:
@lawrenceakka Interesting idea. But I'm not sure if this is what we want. Consider two change handlers:

1. The first handler simply logs the current list.
2. The second handler makes sure no item exceeds a certain maximum. Whenever the list changes, it traverses it and updates individual items if necessary.

If we conditionally early-exit `_handle_change`, the second handler might update too large items, but `_handle_change` isn't executed again to give the first handler a chance to log the updated value. This causes unintuitive inconsistencies.

Therefore I think we need to handle recursions outside of the ObservableCollection where we have a better understanding of the use-case.

falkoschindler:
I noticed that this PR breaks the following code that modifies the original list of rows which has been passed to a table:
```py
rows = [{'name': 'Alice', 'age': 25}, {'name': 'Bob', 'age': 30}]
table = ui.table(rows=rows)
ui.button('Add row', on_click=lambda: (
    rows.append({'name': 'Charlie', 'age': 35}),
    table.update(),
))
```
Pressing the button updates the table on main, but leaves it unchanged with this PR.

This is probably because `rows` is written to `self._props['rows']` where it is converted into an observable list. So if we still want to get this PR merged, we will finally have to stop supporting this way of manipulating table rows and columns.

rodja:
> So if we still want to get this PR merged, we will finally have to stop supporting this way of manipulating table rows and columns.

Yes, we should do that. Devs can easily use `table.rows` instead of the original. And then they do not need to call `table.update()` anymore.

rodja:
And it still works when passing an `ObservableList`:

```py
rows = ObservableList([{'name': 'Alice', 'age': 25}, {'name': 'Bob', 'age': 30}])
ui.table(rows=rows)
ui.button('Add row', on_click=lambda: rows.append({'name': 'Charlie', 'age': 35}))
```

falkoschindler:
This PR was about to bypass the `_send_update_on_value_change` flag which should avoid loopback messages. Take a simple `ui.input` for example: Entering characters shouldn't cause update messages being sent back to the client. But if the value is assigned to the props dict on the server, an update is triggered.

I fixed this problem by introducing a `suspend_updates()` context manager (consistently for props, classes, and style) to selectively suspend updates on value change. Furthermore it can be used when overwriting the `update` method, making the slightly sketchier `prevent_recursion` helper decorator obsolete.

From my point of view this PR is ready for review.

rodja (approved):
This is awesome! So nice to see all these update calls gone from user code.

Issue #4979, opened 2025-07-22
NiceGUI creates app.storage.user = {} and new pubsub connection on every single API Call


lweighall:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import APIRouter
# from fastapi import APIRouter
from nicegui import app, ui, run
import asyncio
import json
from typing import Dict, Any
from dotenv import load_dotenv

load_dotenv('bug.env')
api_router = APIRouter(prefix='/api/v1', tags=['v1'])

import redis


def get_redis_client_info(host='localhost', port=6379, password=None, db=0):
    try:
        # Create Redis connection
        r = redis.Redis(host=host, port=port, password=password, db=db, decode_responses=True)

        # Get client info
        info = r.info('clients')

        return {
            'connected_clients': info.get('connected_clients', 0),
            'pubsub_clients': info.get('pubsub_clients', 0)
        }

    except redis.RedisError as e:
        print(f"Redis error: {e}")
        return None
    except Exception as e:
        print(f"Error: {e}")
        return None

@api_router.get("/redis-stats")
async def redis_stats():
    client_info = get_redis_client_info()
    return client_info


@api_router.get("/clear-app-storage")
async def clear_app_storage():
    app.storage.clear()
    return {"message": "app storage cleared -- but you'll see the user keys in redis still exist"}


@api_router.get("/iterate-user-storage")
async def iterate_user_storage():
    ret = {}
    for user, user_data in app.storage._users.items():
        ret[user] = user_data
    return ret


@api_router.get("/clear-user-storage")
async def clear_user_storage():
    ret = []
    for user, user_data in app.storage._users.items():
        user_data.clear()
        await user_data.close()
        ret.append(user)
    return ret


@api_router.get("/test-endpoint")
async def test_endpoint():
    """
    Simple API endpoint that should NOT create user storage.
    However, due to the bug, this creates app.storage.user = {} in Redis
    and opens a pub/sub connection to monitor it.
    """
    return {
        "message": "This is a simple API call",
        "status": "success",
        "bug_note": "This call should NOT create Redis user storage keys"
    }


app.include_router(api_router)
ui.run(storage_secret="test12345681239129312")
```

START TEST SCRIPT
```bash
#!/bin/bash

[...]

PR #4981, opened 2025-07-24
Management of Redis client connections in app.storage.user

lweighall:
Related to Issue #4979  -- This pull request contains extra code that needs to be removed if it is to be accepted. It is left this way to help explain and demonstrate the problems and solution. 

@rodja requested specifically, that I put this in, rather than him cloning my fork, or viewing my commits here:

- https://github.com/zauberzeug/nicegui/compare/main...lweighall:nicegui:lw_issue4979_api_path_exclude  (this is my current implementation)
- https://github.com/zauberzeug/nicegui/compare/main...lweighall:nicegui:lw_issue4979_original_implementation_notes (these are my notes on the original code)

As a result it is not yet complete and I am submitting as a draft PR. 

I am used to submitting PR's as complete, but in this case it was specifically requested in Issue #4979. 

### Motivation

NiceGUI creates `app.storage.user = {}` with a new session UUID on every API call via `RequestTrackingMiddleware`, regardless of whether storage is needed. This opens a new Redis pub/sub connection per call, eventually hitting Redis's 10K connection limit and making the entire app unavailable.

**Need:** Better control over Redis client connection creation. `RedisPersistentDict` should use context managers and only create persistent connections when actually needed (existing data or new writes), not on every API call that doesn't use `app.storage.user`.

### Implementation

**Goal:** Improve Redis connection management and allow excluding specific URI prefixes (e.g., `/api`) from storage creation.

**Changes:**
- **storage.py:** Added `NICEGUI_REDIS_USER_STORAGE_IGNORE_URI_PREFIXES` env var (CSV of URI prefixes). `RequestTrackingMiddleware` skips creating `app.storage.user` for matching prefixes and passes `skip_redis_publish=True` to prevent unnecessary connections.

- **redis_persistent_dict.py:** Moved Redis connection from `__init__` to lazy `_init_redis_client()`. Non-pubsub operations now use context managers. Persistent connections only created when `_start_listening()` is called. `close()`/`clear()` skip Redis operations when `skip_redis_publish=True`.

**Documentation:** Need to document `NICEGUI_REDIS_USER_STORAGE_IGNORE_URI_PREFIXES` behavior.

### Progress

### DIFF
--- a/main.py
+++ b/main.py
@@ -65,7 +65,7 @@ def some_json():
 
 
 @ui.page('/')
-async def root_path():
+async def root_page():
     # use this page to check that app.storage.user behaves as expected '
     ui.label('root page')
     if not app.storage.user.get('root_page_counter'):
@@ -76,7 +76,7 @@ async def root_path():
 
 
 @ui.page('/api/hello')
-async def root_path():
+async def api_hello_page():
     # use this page to that app.storage.user changes are not published to redis
     # check redis before going back to /
     ui.label('api hello page')
--- a/main.py
+++ b/main.py
@@ -1,65 +1,90 @@
-#!/usr/bin/env python3
-import os
-from pathlib import Path
-from typing import Optional
-
-from fastapi import HTTPException, Request
-from fastapi.responses import RedirectResponse
-from starlette.middleware.sessions import SessionMiddleware
-
+from copy import deepcopy
+import redis
+import time
 from nicegui import app, ui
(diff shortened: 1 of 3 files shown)


rodja:
Thanks @lweighall, but the code is quite complicated and has a lot of changes. Could you have a look at my alternative implementation #4982? How does it compare?

lweighall:
Closing my PR in favor of @rodja 's implementation on PR #4982 -- my notes are on that PR.

PR #4982, opened 2025-07-25
Reduce redis data and connections for API requests

rodja:
### Motivation

As described by @lweighall in #4979, a basic API request creates an empty entry in the redis storage and keeps a subscription until app shutdown. Reproduction:

Start redis server with monitor in one terminal:

```bash
docker run --rm -it -p 6379:6379 redis:7-alpine sh -c 'redis-server --protected-mode no & sleep 0.2; exec redis-cli MONITOR'
```

Then run `main.py` in another terminal with `NICEGUI_REDIS_URL=redis://localhost python main.py`:

```py
from nicegui import app, ui

@app.get('/api')
def some_json():
    return {'message': 'Hello, World!'}

ui.run(storage_secret='my_secret')
```

Each execution of `curl localhost:8080/api` in a third terminal will create a new empty redis entry, as can be seen in the output of the running redis container. Also a subscription is registered which counts towards the connection limit in redis store.

### Implementation

This pull request adds an early exit in the `backup` function similar to the one in `FilePersistentDict` which already skips creating empty entries. Additionally quite a bit of code had to be added to directly unsubscribe from redis as soon as the request is complete. That way, even a longer request (with await's) can react to updates to the redis entry. I admit that this scenario is quite rare but the storage implementation should keep it's promise to
1. make `user.storage` available for API requests
2. allow sync between multiple app instances via redis store

### Progress

### DIFF
--- a/.cursor/rules/coding.mdc
+++ b/.cursor/rules/coding.mdc
@@ -3,6 +3,7 @@ description: writing code
 globs: *.py
 alwaysApply: false
 ---
+
 # Coding Guidelines for AI Assistants
 
 **Primary Reference**: @CONTRIBUTING.md
--- a/.cursor/rules/coding.mdc
+++ b/.cursor/rules/coding.mdc
@@ -3,7 +3,6 @@ description: writing code
 globs: *.py
 alwaysApply: false
 ---
-
 # Coding Guidelines for AI Assistants
 
 **Primary Reference**: @CONTRIBUTING.md
--- a/examples/redis_storage/docker-compose.yml
+++ b/examples/redis_storage/docker-compose.yml
@@ -20,6 +20,8 @@ services:
 
   redis:
     image: redis:alpine
+    environment:
+      - REDIS_TIMEOUT=10
     ports:
       - "6379:6379"
 
--- a/fly.toml
+++ b/fly.toml
@@ -37,7 +37,7 @@ swap_size_mb = 2048
     handlers = ["tls", "http"]
   [services.concurrency]
     type = "requests"
-    hard_limit = 2000
+    hard_limit = 200
     soft_limit = 100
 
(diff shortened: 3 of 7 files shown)


rodja:
@lweighall, could you try this pull request and tell me if it fixes your problem? What would be possible disadvantages compared to your pull request #4981?

lweighall:
@rodja Thank you for this. I knew you folks would have a better idea of how to implement this. The response header check is a great way to deal with the pubsub connections. It makes sense because in the storage documentation it's stated that app.storage.user requires a page builder function.

I've cloned this PR and given it some testing, and it's handling the unnecessary pubsub connections and empty keys the way I would expect it to. 

The overall `connected_clients` from `redis-cli info clients` still goes up even though the pubsub connections do not, even though you're clearly closing the redis connection in your implementation. We can deal with this using a redis client timeout setting, so it should work. That being said we should modify the redis storage documentation to note this redis configuration setting being recommended?

The CLI setting is: `--timeout <SECONDS>`
The ENV var is: `REDIS_TIMEOUT=<SECONDS>`

The important thing was dealing with the pubsub connections we *can't* expire automatically and I think that's been accomplished here. 

Thank you for your work on this. We appreciate you. Feel free to move forward with this PR instead of mine -- it's not important to me that my name be on it, and you clearly know this code base better than I do. I'll close mine out.

rodja:
Thanks @lweighall, I added the hint to the documentation. I guess this is ready for review!

rodja:
> The response header check is a great way to deal with the pubsub connections.

As we found out in #4998 it is not so great after all. The dictionaries are not removed, only the connections are closed. Hence, I removed the header check in #5074 and replaced it with regular pruning of all unused user storage objects.

PR #4994, opened 2025-07-29
Simplify fastapi example

rodja:
### Motivation

The FastAPI example is overly complicated and the concepts in play are hard to grasp.

### Implementation

In each example we try to focus on one concept only and therefore now use a single file which only demonstrates `ui.run_with`.

### Progress


falkoschindler (approved):

addaguob:
> ### Motivation
> The FastAPI example is overly complicated and the concepts in play are hard to grasp. See discussion #4974.
> 
> ### Implementation
> In each example we try to focus on one concept only and therefore now use a single file which only demonstrates `ui.run_with`.

Great work! Thanks! As of this moment, I'm glad to learn about `app` as a FastAPI object and that I wasn't really paying attention. It feels native with `app` like Node's `app.get` thingy. My first app with NiceGUI was based on the old example of this and it looked something like:
```python
#!/usr/bin/env python3

import backend
import frontend

# Initialize NiceGUI frontend with FastAPI integration
frontend.init(backend.fastapi_app)

# Expose the app for Uvicorn
app = backend.fastapi_app

if __name__ == "__main__":
    print(
        'Please start the app with the "uvicorn" command as shown in the start.sh script'
    )
```

PR #4999, opened 2025-08-01
Fix: only use SPA router for relative URLs in ui.navigate.to

mohankumarelec:
### Motivation

Fix SPA navigation so that `ui.navigate.to()` only uses the SPA router for relative URLs. Absolute URLs now trigger standard navigation as expected.

### Implementation

- Fixed `ui.navigate.to()` to distinguish between relative and absolute URLs.
- Added tests for both SPA and external navigation.

### Progress

- [x] Meaningful title
- [x] Complete implementation
- [x] Tests added
- [x] Documentation update not needed

#4821

### DIFF
--- a/nicegui/functions/navigate.py
+++ b/nicegui/functions/navigate.py
@@ -1,9 +1,14 @@
 from typing import Any, Callable, Union
+from urllib.parse import urlparse
 
 from ..client import Client
 from ..context import context
 from ..element import Element
 from .javascript import run_javascript
+
+
+def is_absolute_url(url: str) -> bool:
+    return bool(urlparse(url).netloc)
 
 
 class Navigate:
@@ -67,7 +72,7 @@ class Navigate:
         else:
             raise TypeError(f'Invalid target type: {type(target)}')
 
-        if not new_tab and isinstance(target, str):
+        if not new_tab and isinstance(target, str) and not is_absolute_url(path):
             context.client.sub_pages_router._handle_navigate(path)  # pylint: disable=protected-access
             return
 
--- a/nicegui/functions/navigate.py
+++ b/nicegui/functions/navigate.py
@@ -7,10 +7,6 @@ from ..element import Element
 from .javascript import run_javascript
 
 
-def is_absolute_url(url: str) -> bool:
-    return bool(urlparse(url).netloc)
-
-
 class Navigate:
     """Navigation functions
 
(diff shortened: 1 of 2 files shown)


falkoschindler:
Hi @mohankumarelec, thanks for this pull request!

You seem to be fixing a bug. But could you, please, give a minimum reproduction of the problem? I'm struggling to understand how the current implementation doesn't correctly navigate to absolute URLs.

rodja:
@nightwatcher74 showed in #5002 how to reproduce it. Here is my even more basic reproduction:

```py
from nicegui import ui

@ui.page('/')
def index():
    ui.label('Hello World')
    ui.navigate.to('https://www.google.com/')

ui.run()
```

The problem is that `navigate.to` sends all strings to `client.sub_pages_router`, even if there are no sub pages elements involved and the path is actually a real network location.

rodja:
Or, even shorter:

```py
from nicegui import ui

ui.button('Go external', on_click=lambda: ui.navigate.to('https://nicegui.io'))

ui.run()
```

A variant of what this PR uses to implement `test_navigate_to_absolute_url`.

**NOTE:** in https://github.com/zauberzeug/nicegui/pull/4999#issuecomment-3146991853 I discovered that this reproduction does not work. Sorry.

rodja (changes requested):
Thanks for the PR @mohankumarelec! I think you can remove `test_navigate_to_relative_url_uses_spa` because there are several tests in `test_sub_pages.py` which ensure that relative urls are working with `ui.sub_pages`.

falkoschindler:
Sorry, but I don't see the point of `test_navigate_to_absolute_url`. It is already passing in 2.22.0, which is why I couldn't reproduce the problem based on the code in this PR.

And @rodja's example
```py
ui.button('Go external', on_click=lambda: ui.navigate.to('https://nicegui.io'))
```
is also working for me in 2.22.0. What is going on?

rodja:
You are right @falkoschindler. The original test did not fail on 2.22.0. I'm sorry to not have checked this before. After tracing the different execution paths, I have an explanation for the different behaviours: 
`SubPagesRouter._other_page_builder_matches_path` we have an early exit that returns `True` if you are on an auto-index page. That results in `SubPagesRouter._handle_navigate` to call `client.open` -- and all works fine (in test and presumably also your reproduction). But if you have a page builder, there is no other route that matches an absolute url and hence `SubPagesRouter._handle_navigate` just tries to push the history state.

In retrospect, my comment https://github.com/zauberzeug/nicegui/pull/4999#issuecomment-3145378395 is wrong and you need the page builder to reproduce the issue. I fixed the test on `main` with commit bb0da98.

MicaelJarniac:
I believe this might have caused #5023, or at least be related to it.

Issue #5001, opened 2025-08-01
Documentation subpages broken after SPA update


evnchn:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
https://nicegui.io/documentation/section_action_events#custom_error_page
```

### Description

When go to the URL above, you may observe that https://nicegui.io/raise_timeout_error and https://nicegui.io/raise_timeout_error no longer lead to error pages

### NiceGUI Version

2.22.0

### Python Version

3.12.9

### Browser

Chrome

### Operating System

Windows

### Additional Context

I think the match for the SPA is "too greedy"?

evnchn:
<img width="465" height="145" alt="Image" src="https://github.com/user-attachments/assets/4692cfdb-39e1-4990-add7-161e3f92cbda" />

evnchn:
Persists on 2.22.1

rodja:
I created a fix in #5004 and have just deployed it.

Issue #5002, opened 2025-08-01
ui.navigate.to not working with verison 2.22.0


nightwatcher74:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
import uuid
from fastapi import FastAPI
from nicegui import APIRouter, app, ui
import uvicorn

router = APIRouter(prefix="")

@router.page("/")
def startpage():
    ui.label('Hello World')
    ui.navigate.to('https://www.google.com/')

fastapi_app = FastAPI()

app.include_router(router)  # Startseite (/)

ui.run_with(
    fastapi_app,
    storage_secret=str(uuid.uuid4()),  # für app.storage
    reconnect_timeout=10,  # 10 Sekunden
)

if __name__ == "__main__":
    uvicorn.run("test:fastapi_app", host="127.0.0.1", port=8000, reload=True)
```

### Description

In the nicegui app above, the redirect is not working since nicegui version 2.22.0
With version 2.21.1 I will be redirected to google.com

I tested it with the latest Versions of Firefox and Google Chrome.

with the smaller example 
`ui.label('Hello World')`
`ui.navigate.to('https://www.google.com/')`
`ui.run()`

the browser will be opend and google.com will be shown but this will work exactly one time. When i call it again... only the "Hello World" will be shown.

This are tho small examples but required, because we made redirects for an EntraID oAuth logon with ui.navigate.to()

Thanks 

Frank

### NiceGUI Version

2.22.0

### Python Version

3.12

### Browser

Chrome

### Operating System

Windows

rodja:
Thanks for reporting. Seems like @mohankumarelec found a fix in #4999.

PR #5004, opened 2025-08-02
fix routing to other FastAPI pages in documentation

rodja:
### Motivation

As @evnchn noted in #5001, migrating the documentation to a single-page application (#4902) introduced a bug that makes the FastAPI pages unreachable. The issue was caused by the FastAPI catch-all route for the SPA entry point.

### Implementation

As an temporary fix, this PR uses explicit routes for the SPA plus a custom `ui.sub_pages` implementation which short-stops rendering the main page from rendering before the SPA can decide to route to another FastAPI route.
A better solution might be possible if we implement #4964 in 3.0.

### Progress

### DIFF
--- a/main.py
+++ b/main.py
@@ -6,6 +6,7 @@ from fastapi import Request
 from starlette.middleware.sessions import SessionMiddleware
 
 from nicegui import app, ui
+from nicegui.page_arguments import RouteMatch
 from website import anti_scroll_hack, documentation, fly, header, imprint_privacy, main_page, rate_limits, svg
 
 # session middleware is required for demo in documentation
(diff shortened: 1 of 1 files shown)


evnchn:
Thanks @rodja for the rapid fix!

I think that given the current architecture where things are built on FastAPI, then we should have this as a default usage and default behavior for `ui.sub_pages`:

- User should be explicit in listing the path that is served using `@ui.page`, and don't do `@ui.page('/{path:path}')`, else they "disable" the FastAPI routing entirely. 
- With regards to `ui.sub_page` matching `/`, maybe `ui.sub_page` should give up rendering if the path isn't a complete match, just like `custom_sub_page`. 
  - Perhaps even the case for other entries as well. 

Luckily, from what I gather, as long as you don't use `ui.sub_page`, you won't be affected by this.

evnchn:
But it seems like if it's FastAPI then actually you can manipulate the sequence of defining the routes so that the catch-all is the last, and so more specific routes gets the priority: https://stackoverflow.com/a/76528465

And why the documentation subpages not served is possibly because they are defined after the `@ui.page('/{path:path}')`

Either we can educate the user on that, or maybe NiceGUI can massage it so that, even if you did `@ui.page('/{path:path}')`, the catch-all route is always the last in the priority.

rodja:
Yes, we considered manipulating the order of FastAPI routes. But that is not very intuitive. Have a look at https://github.com/zauberzeug/nicegui/discussions/4964#discussioncomment-13967645 and the draft PR #5002 for an alternative idea which we might implement for NiceGUI 3.0.

PR #5005, opened 2025-08-02
Drop auto-index pages in favour of streamlined usage of SPAs via positional root parameter in ui.run

rodja:
### Motivation

As discussed in #4964, we could streamline the the setup of NiceGUI as an SPA by providing a positional `root` parameter in `ui.run`. The basic pattern becomes so simple that we do not need auto-index pages anymore 🎉

### Implementation

This PR implements an idea form @falkoschindler and me while discussing #4964: by evaluating the `root` builder function in the FastAPI default 404 handler we do not need to register catch-all FastAPI routes at all. Thereby any other defined FastAPI routes take precedence. Only if the FastAPI routing fails we route the path to the `root` builder function passed via `ui.run`.  A basic (already working) example looks like this:

```py
from nicegui import ui

def root():
    ui.link('Go to other page', '/other')
    ui.sub_pages({
        '/': main,
        '/other': other,
    })

def main():
    ui.label('Main page content')

def other():
    ui.label('Another page content')

@ui.page('/test')
def test():
    ui.label('Test page content')

ui.run(root)
```

Of course that only works when auto-index page does not register a `/` page on it's own.

### Progress

  - [ ] remove all auto-index client related code
  - [ ] provide support for async root pages
  - [ ] inject required parameters into root page similar to what ui.sub_pages does

### DIFF
--- a/nicegui/core.py
+++ b/nicegui/core.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import asyncio
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Callable, Optional
 
 from socketio import AsyncServer
 
@@ -13,3 +13,4 @@ app: App
 sio: AsyncServer
 loop: Optional[asyncio.AbstractEventLoop] = None
 air: Optional[Air] = None
+spa: Optional[Callable] = None
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -13,6 +13,7 @@ from . import air, background_tasks, binding, core, favicon, helpers, json, run,
 from .app import App
 from .client import Client
 from .dependencies import dynamic_resources, js_components, libraries, resources
+from .elements.sub_pages import SubPages
 from .error import error_content
 from .json import NiceGUIJSONResponse
 from .logging import log
@@ -57,13 +58,6 @@ static_files = CacheControlledStaticFiles(
     follow_symlink=True,
 )
 app.mount(f'/_nicegui/{__version__}/static', static_files, name='static')
-
-Client.auto_index_client = Client(page('/'), request=None).__enter__()  # pylint: disable=unnecessary-dunder-call
-
-
-@app.get('/')
-def _get_index(request: Request) -> Response:
-    return Client.auto_index_client.build_response(request)
 
 
 @app.get(f'/_nicegui/{__version__}' + '/libraries/{key:path}')
@@ -149,6 +143,13 @@ async def _shutdown() -> None:
 
 @app.exception_handler(404)
 async def _exception_handler_404(request: Request, exception: Exception) -> Response:
+    if core.spa is not None:
+        with Client(page(''), request=request) as client:
+            core.spa()
+            # NOTE: after building the page, there might be sub pages that have 404 -- and initial requests should send 404 status request in such cases
+            sub_pages_elements = [e for e in client.elements.values() if isinstance(e, SubPages)]
+            if not any(sub_pages.has_404 for sub_pages in sub_pages_elements):
+                return client.build_response(request)
     log.warning(f'{request.url} not found')
     with Client(page(''), request=request) as client:
         error_content(404, exception)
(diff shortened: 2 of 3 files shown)


evnchn:
This doesn't have to be mutually exclusive with the auto-index page (I think), if we have the default root to serve the auto-index page. Then, when you pass the root that auto-index page will be ignored (add a warning).

Also, it may be a bit of a pity to lose the shared client of the auto-index page. We may want to look at https://github.com/zauberzeug/nicegui/pull/4729 with a fresh pairs of eyes and keep NiceGUI competitive in the shared-client web development framework aspect as well.

rodja:
> it may be a bit of a pity to lose the shared client of the auto-index page.

I disagree. It would be a relief for the code base. Let's move the discussion to https://github.com/zauberzeug/nicegui/discussions/4472#discussioncomment-13969621 and focus on the implementation of this PR here.

falkoschindler:
I don't think `ui.run(root)` requires dropping the auto-index client. We could limit this PR to extending `ui.run` and discuss implications for the auto-index client afterwards. This PR might further decrease the necessity for the auto-index page, but it shouldn't be held back by the ongoing discussion.

rodja:
> I don't think ui.run(root) requires dropping the auto-index client.

@falkoschindler I do not see the point in keeping the auto-index client. It will make the code -- and behavior more complicated. As @evnchn said,

> Then, when you pass the root that auto-index page will be ignored (add a warning).

This warning will be gone or change significantly when dropping auto-index client in 3.0.

rodja:
@falkoschindler I experimented with the test infrastructure to support SPAs in d6532f8. But wonder if there is a better way... lets discuss this in person later.

rodja (commented):
Looks good so far!

rodja:
Should we not use the new root page mechanism for the ROS2 example?

falkoschindler:
@rodja I'm not sure. `NiceGuiNode` is instantiated in `ros_main`, which runs in a separate thread. At least it is not trivial to make a root function accessible from the outside. It's not a priority right now.
Apart from that we could certainly update many other examples. Maybe we can do that after this PR is merged and when we re-visit every example anyway for a final test?

rodja:
Looks good. I just would like to see more root page use instead of page decorators in the examples. But we can also do this in a separate PR...

defkev:
This discussion should unironically be framed and taught in STEM classes on how NOT to add new features to mature software:

Screenshot 2025-11-07 at 06-47-58 Drop auto-index pages in favour of streamlined usage of SPAs via positional root parameter in ui run by rodja · Pull Request #5005 · zauberzeug_nicegui


Not only is this train of thought extremely destructive, as already pointed out by both @evnchn and @falkoschindler, it's also the definition of anti-user.

Imagine Microsoft releasing an update tomorrow which opens every *.exe using Notepad instead of the kernel by default.
Definitely hilarious but also really, really stupid. Just like this PR!

evnchn:
To clarify, there is thinking and voting process in #4472. Do give it a read if not already. 

We are and should be open to feedback but we would appreciate focusing on what can be improved. Let's take it chill and calm. 

Do you have any concrete suggestions? I would love to hear. 

---

Finally about Microsoft. It brings a good point:

Microsoft is really good at maintaining backwards compatibility, but their products have technical debt as a direct result. 

As a (potentially inappropriate) analogy, I would say that if they did semantic versioning for Windows then if Windows 2000 is 1.0 then possibly Windows 11 is still on 1.x. It's the same NT kernel all the way. 

NiceGUI is stepping from 2.x to 3.0 and so we are delivering breaking changes. Dropping the auto-index client got us a lot less issues. Think of it like MacOS to keep the analogy.

defkev:
Microsoft is indeed great when it comes to backwards-compatibility. I just recently saw 25-year-old legacy 16-bit code run on a 32-bit NTVDM Windows 10. But that wasn't my point.

Edit: After glancing over the linked discussion, it might be worth noting we use nicegui exclusively for rapid UI prototyping, which probably doesn't correlate with the idea of the author what it should be used for. It's probably really for the best to just move to something else.

PR #4749, opened 2025-05-15
Support for ajv-formats in `ui.json_editor`

evnchn:
This PR fix #4748 by introducing ajv-formats such that `'format': 'uuid'` and the likes can be supported in `ui.json_editor(..., schema=schema)` 

@phnmn kindly check if this fixes what you are trying to do, which I think it does. You can use this new code in a pinch before we decide if this is a worthy addition and finalize this PR. 

Test script: 

```py
from nicegui import ui

schema = {
    'type': 'object',
    'properties': {
        'id': {
            'type': 'integer',
        },
        'uuid': {
            'type': 'string',
            'format': 'uuid',  # troublesome line
        },
        'name': {
            'type': 'string',
        },
        'price': {
            'type': 'number',
            'exclusiveMinimum': 0,
        },
    },
    'required': ['id', 'name', 'price', 'uuid'],
}
data = {
    'id': 42,
    'uuid': '123e4567-e89b-12d3-a456-426614174000',
    'name': 'Banana',
    'price': 15.0,
}
ui.json_editor({'content': {'json': data}}, schema=schema)

ui.run()
```

Before the PR: 

{41AD2F12-2231-45B0-B396-0BC568C642DA}

After the PR: 

{F0085D14-B88C-47C9-9DE5-1716A117CE8F}

{2C28517E-FF6A-41B4-B2C2-B92CA479CAE3}

To-do items: 

- [ ] I can't figure out how best to integrate this with `npm.py` and `npm.json`
  - I get import errors. 
  - Also it will occupy "index" in importmap, which hardly sounds like a good idea
  - Then, we can get rid of the access to jsdelivr.net. 
  - @falkoschindler this I think you are more proficient in `npm.py` and `npm.json`...

Issue #4163, opened 2025-01-01
No control over the key used for the generated importmap


simontaurus:
### Description

Import map key generation seems to be based on the dependencies filename:
https://github.com/zauberzeug/nicegui/blob/09f4c34049b37907174bfe3109aae4b834f80863/nicegui/dependencies.py#L126

e.g. 
```py
dependencies=['node_modules/@json-editor/json-editor/dist/jsoneditor.js']
```

leads to the key `jsoneditor` which is also required in dynamic imports within custom vue components:
```js
import("jsoneditor")
```
(see also #4162)

However, this may lead to conflicts if two libs use the same filename. Also, it would be preferable to use the correct package name to stay interoperable with other environments (e.g. vite):
```js
import("@json-editor/json-editor")
```

falkoschindler:
Thanks for reporting this issue, @simontaurus!
And sorry for keeping you waiting. Now I finally found the time and focus to look into it.

You're absolutely right: Basically using the file _name_ for the importmap key is very fragile and can lead to conflicts rather quickly. I'm just not sure how to improve it.

This is what `_get_name` is called with when running the NiceGUI website:
```
/Users/falko/Projects/nicegui/nicegui/elements/aggrid.js
/Users/falko/Projects/nicegui/nicegui/elements/lib/aggrid/ag-grid-community.min.js
/Users/falko/Projects/nicegui/nicegui/elements/audio.js
/Users/falko/Projects/nicegui/nicegui/elements/mermaid.js
/Users/falko/Projects/nicegui/nicegui/elements/lib/mermaid/mermaid.js
/Users/falko/Projects/nicegui/nicegui/elements/lib/mermaid/mermaid.min.js
/Users/falko/Projects/nicegui/nicegui/elements/lib/mermaid/mermaid.esm.min.mjs
/Users/falko/Projects/nicegui/nicegui/elements/lib/mermaid/chunks/mermaid.esm.min/blockDiagram-NDWNTGEE.mjs
/Users/falko/Projects/nicegui/nicegui/elements/lib/mermaid/chunks/mermaid.esm.min/c4Diagram-AUYESYAG.mjs
/Users/falko/Projects/nicegui/nicegui/elements/lib/mermaid/chunks/mermaid.esm.min/chunk-2RSIMOBZ.mjs
...
/Users/falko/Projects/nicegui/nicegui/elements/lib/mermaid/chunks/mermaid.esm.min/xychartDiagram-RLS75X5Z.mjs
[...]

PR #5021, opened 2025-08-08
Re-organize and upgrade NPM dependencies

falkoschindler:
### Motivation

There are several problems with the existing dependency management via npm.json and npm.py.

**No control over importmap key**

Example: Newer Three.js versions come with three.module.js which points to three.core.js. We can declare both files as a dependency, but they use the same key "three" in the importmap. A similar problem came up in #4163.

**"Manual" code modifications needed**

If a file like three.module.js references `./three.core.js`, we have to replace the relative link because the file three.core.js isn't found at this relative URL. Similarly there is an import of `../utils/BufferGeometryUtils.js` in Three.js' GLTFLoader which we need to replace with `BufferGeometryUtils`. These corrections happen in npm.py.

**Complex dependencies hard or impossible to integrate**

There is an extra script at scripts/codemirror/bundle.bash to integrate CodeMirror. And we wanted to integrate ajv-formats for `ui.json_editor` (https://cdn.jsdelivr.net/npm/ajv-formats@3.0.1/+esm), but it includes references to e.g. "/npm/ajv@8.12.0/+esm" which don't work in a local NiceGUI app (see #4749).

### Implementation

For this PR I decided to move modules with dependencies into separate directories (e.g. nicegui/elements/scene/) instead of holding all dependencies in one giant lib/ directory. This includes codemirror, scene, aggrid, joystick, json_editor, plotly, echarts, leaflet and mermaid.

Each of these subdirectories contains an independent package.json, bundler configuration (rollup or vite), src/ directory with an index.js, a Git-ignored node-modules/ directory, and a generated dist/ directory with the resulting (sometimes chunked) bundle. This way we can configure the build individually for each UI element and don't have to find a one-fits-all solution. Furthermore, we can rely on standard NPM tools without the need for our own custom JSON format and Python scripts for downloading and extracting packages. Now the build process is 1. `npm install` and 2. `npm run build`.

For registering the module in NiceGUI, I introduced a new `esm` parameter when subclassing an `Element`, e.g.:
```py
class JsonEditor(Element, component='json_editor.js', esm={'nicegui-json-editor': 'dist'}):
    ...
```
This defines an entry to the importmap named "nicegui-json-editor" pointing to the relative directory dist/. This replaces the definition of individual `dependencies=[...]`, which we will probably deprecate.

### Progress

The PR is still work in progress. The change has quite some implications that still need to be implemented.

    - [x] remove deprecated parameters from `Element.__init_subclass__`
    - [x] create custom package.jsons per UI element
    - [x] create package.json for NiceGUI's core dependencies (Vue, Quasar, Tailwind)
    - [x] remove npm.json/npm.py
    - [x] update examples for custom components (number_checker, signature_pad)
    - [x] update gitattributes
    - [x] generate DEPENDENCIES.md?
    - [x] what about minification and maps?

### DIFF
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -146,8 +146,6 @@ export default {
     },
   },
   async mounted() {
-    await this.$nextTick(); // NOTE: wait for window.path_prefix to be set
-
     // This is used to prevent emitting the value we just received from the server.
     this.emitting = true;
 


falkoschindler:
To be clear: The change _feels great_! By removing the indirection of npm.json and npm.py we enable to work with NPM more directly, which reduces mental complexity (for humans and AI alike) and brings the whole NPM infrastructure with its huge feature set closer to our fingertips.
And while the checklist looks like we didn't make much progress yet, moving and transforming the dependency management for all UI elements was probably the hardest and most time-consuming step.

PR #5011, opened 2025-08-04
Allow canceling the currently executed invoctation of ui.timer

rodja:
### Motivation

Originally, this PR was motivated by #5010, where the async demo in `ui.sub_pages` documentation does not work properly when rapidly clicking the links. When trying to fix the bug, I discovered that there is currently not an option in `ui.timer` to cancel the ongoing task.

### Implementation

This PR introduces `timer.cancel(with_current_invocation=True)` to allow canceling the internal awaited coroutine.

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests have been added.
- [x] Documentation has been added.

PR #5025, opened 2025-08-09
Fix bad navigation of relative urls

rodja:
### Motivation

In #5023, @MicaelJarniac reported that the google oauth example is broken. But the problem is much larger. All relative navigations do not work if you use page builder but not the new `ui.sub_pages`. We fixed absolute urls in #4999. But relative urls are also problematic.

### Implementation

This PR fixes #5023 by ensuring sub pages router only handles navigation if there are ui.sub_pages elements.

### Progress

### DIFF
--- a/nicegui/functions/navigate.py
+++ b/nicegui/functions/navigate.py
@@ -4,6 +4,7 @@ from urllib.parse import urlparse
 from ..client import Client
 from ..context import context
 from ..element import Element
+from ..elements.sub_pages import SubPages
 from .javascript import run_javascript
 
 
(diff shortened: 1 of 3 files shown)


rodja:
@falkoschindler maybe it would be good to release a 2.22.2 with just this fix and #5011?

falkoschindler (approved):
I decided to move the extra condition to navigate.py. This way it isn't evaluated on "sub_pages_navigate", but only when the router is called from outside. Basically you shouldn't ask the sub-pages router to `_handle_navigate` if there is no `sub_pages` element.

Apart from that I refined the assertions in `test_navigate_to_relative_url` because `'\' in screen.selenium.current_url` is always true.

Let's see if all tests are green and merge afterwards.

PR #5034, opened 2025-08-12
Upgrade third-party dependencies

falkoschindler:
### Motivation

For the upcoming release of version 3.0 we want to upgrade all our third-party dependencies.

### Implementation

Based on the new package management (PR #5021) this PR basically ran
```bash
ncu -u
npm install
npm clean
npm build
```
in each directory with a package.json and
```bash
./summarize_dependencies.py
```
in the root directory.

### Progress

  - [x] echarts-gl is not compatible with echart 6.0 (https://github.com/ecomfe/echarts-gl/issues/556, but seems pretty stale)
  - [x] The "Expand" button at https://nicegui.io/documentation/json_editor#run_methods is broken


falkoschindler:
The dependency https://github.com/ecomfe/echarts-gl for 3D graphs in `ui.echart` requires ECharts 5 and isn't officially compatible with the latest version of ECharts 6. But because echarts-gl seems pretty stale and I don't want us to remain on EChart 5 forever, I decided to simply ignore the incompatibility. Basic functionality seems to work nonetheless. And if we should discover nuanced problems in the future, we can hope for an upstream fix or drop support for 3D plots. That's better than limiting 2D ECharts in NiceGUI.

PR #5038, opened 2025-08-12
Starlette/fastAPI Routing Fix

git2212:
### Motivation

fastAPI/starlette does not set up proper routes for NiceGUI endpoints (html/wss) when registering the NiceGUI as a sub-app with fastAPI.

for example:

```py
main_app = FastAPI(root_path=/my_nice_ui)
ui.run_with(
    main_app,
    mount_path='/niceui', 
    storage_secret='pick your private secret here',  # NOTE setting a secret is optional but allows for persistent storage per user
)
```

you'd expect your NiceGUI sub app to be available at '/my_nice_ui/niceui' and while it is registered as such by fastAPI, there is no routing setup for it by fastAPI/starlette. Developers who are fastAPI/starlette experts won't loose a beat and quickly register their own router... the rest of the world will spend a week following one silly recommendation after another from various online fori.

While this is not a NiceGUI deficiency per-se, the fix for this is rather simple

### Implementation

For proper routing of html/wss requests to the sub-app a custom starlette middleware is registered when NiceGUI is running as a sub-app. The middleware is specific to NiceGUI and ensures that both http and wss requests are routed to the correct endpoints.

this class (RedirectWithPrefixMiddlewareSubApp) is added to middlewares.py with a two liner modification to ui_run_with.py

### Progress

- [ ] I chose a meaningful title that completes the sentence: "If applied, this PR will ensure that when a NiceGUI app is registered as a fastAPI sub-app, the routing is configured without any guesswork"
- [ ] There is not specific testing required for this fix. other than the example code above
- [ ] Current documentation does not require updates, in fact this PR would ensure that the expectation of proper routing in a variety of use-cases is as per the current docs.

### DIFF
--- a/nicegui/middlewares.py
+++ b/nicegui/middlewares.py
@@ -5,6 +5,9 @@ from starlette.responses import Response
 from . import core
 from .version import __version__
 
+import re
+from starlette.middleware import Middleware
+from starlette.types import ASGIApp, Receive, Scope, Send
 
 class RedirectWithPrefixMiddleware(BaseHTTPMiddleware):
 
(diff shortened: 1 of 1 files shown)


git2212:
> Thanks for the PR @git2212. Why have you added a new middleware? It should be enough to concatenate the values in
> 
> https://github.com/zauberzeug/nicegui/blob/c798539870750bf6949004327f36633d76185020/nicegui/client.py#L134

I do agree that there are many other ways to address what is a well know issue with fastAPI... namely, routing for sub-apps is left to the sub-app... Adding a new middleware in my view is the cleanest way to provide the proper routing... I am using this in my project, I figured I'd share it with the community... it's on offer, not a demand. If you don't want it you can reject it.

rodja:
Thanks @git2212. I think the middleware approach might have too many side-effects which are hard to anticipate. Hence I created #5063 as an alternative.

PR #5040, opened 2025-08-13
add an option to bindings to check if target exists

codingpaula:
### Motivation

When refactoring or renaming variables their string names in bindings are often missed since the rename IDE action doesn't pick it up. We discussed creating a method that gets the variable name as a string from a reference which proved possible but costly. As an alternative to raise awareness to missed variable names after renaming we propose to add an option that checks if the binding target exists and warns if it's not.

### Implementation

The implementation is rather straight forward and involves a lot of trivial parameter adding. The check takes place in the functions `bind`, `bind_to` and `bind_from` in `bindings.py` which are used in all the other occurrences.

### Progress

  - [x] update all docstrings
  - [x] `direction` could be a `Literal`
  - [x] check both, source and target
  - [x] change it to `check: bool | None = None` and check non-dicts by default

### DIFF
--- a/tests/test_binding.py
+++ b/tests/test_binding.py
@@ -2,6 +2,7 @@ import copy
 import weakref
 from typing import Optional
 
+import pytest
 from selenium.webdriver.common.keys import Keys
 
 from nicegui import binding, ui
@@ -210,12 +211,10 @@ async def test_nested_propagation(user: User):
 def test_binding_other_dict_is_strict(screen: Screen):
     data: dict[str, str] = {}
     label = ui.label()
-    binding.bind(label, 'text', data, 'non_existent_key', other_strict=True)
+    with pytest.raises(KeyError):
+        binding.bind(label, 'text', data, 'non_existent_key', other_strict=True)
 
     screen.open('/')
-    screen.assert_py_logger('WARNING',
-                            'Binding a non-existing attribute "non_existent_key" of target object of type dict. '
-                            'Proceeding with binding, keeping the value unset.')
 
 
 def test_binding_object_is_strict(screen: Screen):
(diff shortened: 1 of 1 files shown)


falkoschindler:
Some thoughts:

- Not all docstrings are up-to-date yet.
- The `direction` could be a literal. (It doesn't really matter, but `str` feels a bit arbitrary.)
- Should we check both, source _and_ target?
- Should we change it to `check: bool | None = None` and check non-dicts by default? (Would need to wait for 3.0.)
- Should we add a way to configure the check globally?

falkoschindler:
@codingpaula I updated the checklist in the original post with the points we discussed yesterday. A global configuration option isn't needed at the moment. It would complicate things unnecessarily.

falkoschindler:
@codingpaula @rodja The current API is like
```py
label.bind_text_from(model, 'text', check_exists=True)
binding.bind(model, 'text', label, 'text', check_self=True, check_other=True)
```
What do you think about changing it to
```py
label.bind_text_from(model, 'text', strict=True)
binding.bind(model, 'text', label, 'text', self_strict=True, other_strict=True)
```
It's shorter, less sketchy English ("check exists" instead of "check if exists"), and a common term for some additional check or verification that can happen behind the curtain (like `zip(..., ..., strict=True)`).

codingpaula:
That sounds nice! Also tools like mypy have the strict option, so it's something people automatically have a feeling for what it does. I'll change it to what you suggested.

falkoschindler:
@codingpaula @rodja Now that we named the new parameter `strict` (which I like very much), I do wonder if we should raise instead of showing a warning - which doesn't feel very "strict". I mean, you get a warning if you're either a) binding to a missing dictionary entry with `strict=True` or b) binding to a missing object attribute without `strict=False`. In both cases an exception might be adequate.

codingpaula:
@falkoschindler I‘d argue against raising an exception. It seems like a really big step to change the default behavior from completely ignoring if an attribute exists (<3.0.0) to raising on all non-dictionary bindings (>=3.0.0). Also, I don’t like the intensity of raising for a missed binding. If I miss changing a binding, the whole page could fail to show and I‘d have no other choice than to immediately fix it. That seems quite drastic for something that NiceGUI ignored up until now and that usually is a missing text, value or false visibility status.
Maybe the middle ground would be to log an error or even an exception that's caught by the binding itself to make it clearer to developers that there's something not working?

Issue #4886, opened 2025-06-25
UI collapses when dark_mode = None and page contains grid with ag grid


BaccanoMob:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

@ui.page(path="/")
async def error_causer():
    ui.dark_mode(None)

    with ui.grid().classes("items-start place-items-stretch w-full grid-cols-2"):
        ui.aggrid({})
        ui.button()

ui.run()
```

### Description

With commenting ui.dark_mode,

![Image](https://github.com/user-attachments/assets/bbd8eb9f-087c-4b0c-b06c-a52e12381fe9)

Without commenting  ui.dark_mode,

![Image](https://github.com/user-attachments/assets/d5bfb25b-2e70-476b-9b6e-e89533aacb6b)

If I set dark=None in run,

![Image](https://github.com/user-attachments/assets/b7eca0e8-1bdf-41a2-ae69-519d7a0537a9)

The bug does not happen if there are buttons only (possibly if its only quasar components I believe)

I wanted to have an ag grid on left some content on the right. And I have managed to get a minimal reproducible code. I do not know what causes it so please change the title if the cause is something else. 

For now I will stick with True/False for dark mode.

### NiceGUI Version

2.20.0

### Python Version

3.12

### Browser

Chrome

### Operating System

Linux

falkoschindler:
Oh wow, thanks for reporting this issue, @BaccanoMob!

First I thought this is a duplicate of #3753, but the behavior is quite different. Here is an even more compact reproduction:

```py
ui.dark_mode(None)

with ui.card().classes('w-full'):
    ui.aggrid({})
```

Currently I have no idea why `ui.dark_mode(None)` has such an effect on the layout.

The auto-resizing in `ui.aggrid` doesn't seem to be the reason. The problem persists after removing this block.
https://github.com/zauberzeug/nicegui/blob/0695e11e2959eee43bdf451025be5c14a20469a8/nicegui/elements/aggrid.js#L58-L60

evnchn:
I'm a tad busy but can anyone can check if the UnoCSS implementation at #4832 is also affected?

At this point, I consider #4832's UnoCSS as the "reference implementation" of Tailwind CSS on the browser. 

[...]

Issue #3753, opened 2024-09-19
nicegui 2.0 seems to change/break some tailwind behavior?


petergaultney:
### Description

> I'm not (currently) able to give a minimal reproducible example. In lieu of that, I'm going to write down the basic things I'm observing here, and then go spend some time debugging the NiceGUI code to see if I can figure out what's going on. I'll report back with any findings.

We upgraded from NiceGUI 1.4.30 to 2.1.0, and a lot of our CSS styling broke. We think it mostly has to do with tailwind classes somehow no longer being present/applied. I have narrowed it down to the 1.4.37 -> 2.0.0 upgrade by installing those versions specifically.

Screenshots, before and after the upgrade:

(expected layout, which we get in 1.4.37 and earlier)
<img width="1715" alt="Screenshot 2024-09-19 at 12 42 57" src="https://github.com/user-attachments/assets/5827ee2c-c5b1-454e-b088-5e540e91b7d8">

(what we see in 2.0)
<img width="1715" alt="Screenshot 2024-09-19 at 12 42 04" src="https://github.com/user-attachments/assets/3926fdf2-b4bb-48ef-9c14-4d81ec8dba1c">

The relevant code, which was unchanged:

```python
def _render_task_counts_display(task_counts: ty.Collection[QueueCount]) -> None:
    with ui.row(wrap=False).classes("w-full"):
        with ui.column().classes("w-1/2"):
            viz.render_task_counts_by_intent(task_counts)
            viz.render_task_counts_by_typename(task_counts)
        with ui.column().classes("w-1/2"):
            viz.render_task_counts_by_typename_and_intent(task_counts)


async def queue_stats_area():
    task_counts = await list_queue_counts("task")
    policy_counts = await list_queue_counts("policy")

    with ui.column().classes("w-full"):
        viz.render_total_task_count(task_counts)
        with ui.row(wrap=False).classes("w-full"):
            with ui.column().classes("w-1/2"):
                _render_task_counts_display(task_counts)
            with ui.column().classes("w-1/2"):
                viz.render_policy_counts(policy_counts)
```

petergaultney:
aha - so, interestingly, the `<head><style>` tag for tailwind on 1.4.37 contains the following clear reference to the assigned background color:

<img width="1363" alt="Screenshot 2024-09-19 at 13 27 47" src="https://github.com/user-attachments/assets/8ba4203a-4069-41fb-aa6f-ab86caf2a700">
[...]

PR #5042, opened 2025-08-14
Fix Tailwind for `ui.dark_mode` in combination with `ui.run(dark=None)`

falkoschindler:
### Motivation

While working on migrating to Tailwind 4, I learned how to properly configure dark mode to use Quasar's `body.body--dark` class and noticed that we can use it for Tailwind 3 as well. As it turns out, this solves both #3753 as well as #4886. 🎉

### Implementation

This PR configures Tailwind to watch Quasar's `body.body--dark` class:
```js
tailwind.config.darkMode = ["class", "body.body--dark"];
```
This way Tailwind follows Quasar's dark mode precisely and we can remove all other code related to Tailwind's dark mode.

### Progress

- [x] Pytests are not necessary (or at least very difficult if we want to simulate the system preference).

### DIFF
--- a/nicegui/elements/dark_mode.js
+++ b/nicegui/elements/dark_mode.js
@@ -8,12 +8,6 @@ export default {
   methods: {
     update() {
       Quasar.Dark.set(this.value === null ? "auto" : this.value);
-      if (window.tailwind) {
-        const mode = this.value === null ? "media" : "class";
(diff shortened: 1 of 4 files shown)


rodja (approved):

petergaultney:
y'all are amazing. thanks for coming back to this.

PR #5044, opened 2025-08-15
Show a screenshot per example on the website

leocjj:
### Motivation

There is an open Enhancement in NiceGUI Wishlist project [No. 10](https://github.com/orgs/zauberzeug/projects/6?pane=issue&itemId=103129975): Show a screenshot per example on the website

PR #4391 was created but has requested changes, so I'm creating this new PR to follow the proposed plan by @falkoschindler:

### Implementation

This PR has a sample of 5 examples with screenshots and README files to get some feedback before doing the others 45.

### Progress

- [5/50] Regenerate screenshots manually to better fit the UI content.
- [5/50] Add a README.md to every example folder, including at least: a title, 
a short description, an image pointing to the corresponding screenshot.png [examples.py](https://github.com/zauberzeug/nicegui/blob/main/website/examples.py)
- [0/1] Update [examples.py](https://github.com/zauberzeug/nicegui/blob/main/website/examples.py) to automatically load data from the examples folder, fetching title and description from the readme files.
- [0/50] Improve the example tiles on the main page to somehow include screenshots. Popups might be a good start.

### DIFF
--- a/website/main_page.py
+++ b/website/main_page.py
@@ -157,7 +157,7 @@ def create() -> None:
         section_heading('In-depth examples', 'Pick your *solution*')
         with ui.row().classes('w-full text-lg leading-tight grid grid-cols-1 sm:grid-cols-2 xl:grid-cols-3 gap-4'):
             for example in examples:
-                if example.title in ['Authentication', 'Chat App', 'Todo List']:
+                if example.title in {'Authentication', 'Chat App', 'Todo List'}:
                     example_link(example)
             example_link().classes('min-h-40 text-center items-center justify-center xl:col-span-3')
 


falkoschindler:
Thank you for the initiative to improve the example tiles, @leocjj! Starting with a few examples only before generating all 50 screenshots and markdown files is definitely a great idea.

Let's continue updating examples.py and the example tiles to see how the final result looks like. 🙂

falkoschindler:
@leocjj I appreciate your work. But let's focus on how to integrate screenshots into the [grid tiles](https://nicegui.io/#examples) on the website before spending more time and effort on creating screenshots. E.g. the orange title bars might look weird within the blueish theme, so we probably need to adjust them anyway.

leocjj:
Hi @falkoschindler, I'm working on the [grid tiles](https://nicegui.io/#examples), this is how it looks like with the images.

image

You're right about the orange title bars, what do you think about this two styles? Second (AI interface) and third (Audio recorder) columns?

---
Here the same two styles for another example

## Style 1: 
screenshot_style_1

## Style 2
screenshot_style_2

I prefer style 1 because for some examples the url is important

rodja:
Maybe we just show the webpage without any chrome?

falkoschindler:
I also think we can remove the frame and only show the page content. It helps to recognize an example and to get a brief impression of what to expect. But we don't need to show every detail.

Oh and right now the screenshots are 1920x1080 (but zoomed in?). Maybe a smaller size would be better? I would suggest something like 960x540 or something in this ballpark which nicely fits the content.

Apart from that we might want to use WEBP instead of PNG. It usually reduces the file size significantly without any quality loss. We just need to check if it works as part of a README.md on GitHub.

leocjj:
Thank you for your feedback @rodja and @falkoschindler, what do you think about this? Images are now webp at 960x540 (average size of just 10 kb)

image

falkoschindler:
I like it very much! 👍🏻
@rodja do you agree?

rodja:
Yes. That could work. We can still ask our designer to make some fine-tuning to the appearance. But the cropped screenshots are the right way.

leocjj:
Hi @falkoschindler and @rodja , it's almost done, do you want to take a look?
Please let me know in case of some feedback!

website/examples.py
```diff
@@ -10,68 +10,47 @@ class Example:
     title: str
     description: str
     url: str = field(init=False)
+    screenshot: str = field(init=False, default='screenshot.webp')
```
evnchn: I think there's no need for a default, if you're going to set it in the post-init anyways.

website/examples.py
```diff
@@ -10,68 +10,47 @@ class Example:
...
+            nonempty = [line.strip() for line in text.splitlines() if line.strip()]
+            if nonempty:
+                # first non-empty line is the title (strip any leading markdown header markers)
+                title_candidate = nonempty[0].lstrip('#').strip()
+                title = title_candidate or default_title
+                # second non-empty line (if present) is used as description
+                description = nonempty[1] if len(nonempty) > 1 else default_description
+        except Exception:
+            # keep defaults on read/encoding errors
+            pass
```
evnchn: Can we log the errors in the console so that people can notice and fix their stuff when they are creating new examples?

website/examples.py
```diff
@@ -10,68 +10,47 @@ class Example:
...
+    if readme.exists():
+        try:
+            text = readme.read_text(encoding='utf-8')
+            nonempty = [line.strip() for line in text.splitlines() if line.strip()]
+            if nonempty:
+                # first non-empty line is the title (strip any leading markdown header markers)
+                title_candidate = nonempty[0].lstrip('#').strip()
+                title = title_candidate or default_title
+                # second non-empty line (if present) is used as description
+                description = nonempty[1] if len(nonempty) > 1 else default_description
```
evnchn: BONUS: This reads all the text into memory, but you're looking at the first 2 lines only. 

Try and open the file and `readline()` / `next()` it twice. 

This will be useful as sometimes we have additional stuff in the readme which is not in the website.

website/examples.py
```diff
@@ -10,68 +10,47 @@ class Example:
...
-    Example('OpenAI Assistant', "Using OpenAI's Assistant API with async/await"),
-    Example('Redis Storage', 'Use Redis storage to share data across multiple instances behind a reverse proxy or load balancer'),
-    Example('Google One-Tap Auth', 'Authenticate users via Google One-Tap'),
-    Example('Google OAuth2', 'Authenticate with Google OAuth2'),
-    Example('Stripe', 'Use Stripe to process payments'),
-]
+# discover README files for each example directory (no symlink recursion)
+readme_files_paths: List[Path] = sorted(
+    (p / 'README.md' for p in PATH.iterdir() if p.is_dir(follow_symlinks=False)),
+    key=lambda p: p.parent.name.lower(),
```
evnchn: Hmm isn't parent of p is always PATH though... Not sure I'm a bit rusty in Pathlib stuff...

Do you intend for `p.name.lower()`?

website/examples.py
```diff
@@ -10,68 +10,47 @@ class Example:
...
+)
+
+examples: List[Example] = []
+for readme in readme_files_paths:
+    # sensible defaults derived from the folder name
+    default_title = readme.parent.name.replace('_', ' ').title()
+    default_description = default_title
+
+    title = default_title
+    description = default_description
```
evnchn: You can write the defaults directly into title and description and they will be overwritten later when the code below runs successfully.

website/examples.py
```diff
@@ -10,68 +10,47 @@ class Example:
...
-    Example('Signature Pad', 'A custom element based on [signature_pad](https://www.npmjs.com/package/signature_pad'),
-    Example('OpenAI Assistant', "Using OpenAI's Assistant API with async/await"),
-    Example('Redis Storage', 'Use Redis storage to share data across multiple instances behind a reverse proxy or load balancer'),
-    Example('Google One-Tap Auth', 'Authenticate users via Google One-Tap'),
-    Example('Google OAuth2', 'Authenticate with Google OAuth2'),
-    Example('Stripe', 'Use Stripe to process payments'),
-]
+# discover README files for each example directory (no symlink recursion)
+readme_files_paths: List[Path] = sorted(
+    (p / 'README.md' for p in PATH.iterdir() if p.is_dir(follow_symlinks=False)),
```
evnchn: I think the code logic will be better, if you first list all example folder paths, then below you check if the readme exists.

Now it's like, we'll go over all the possible readme file paths (may or may not exist) while we rely on `.parent` for the original example folder.

evnchn (changes requested):
Thank you for the PR. This may be a bit extra but see I'm cranking up the detection sensitivity all the way to 200% in hopes that when @falkoschindler returns from the holiday he can skip one "code quality" / "code refactoring" commit on this PR. 

Please feel free to speak up if you think some of the comments don't apply. Thanks!

website/examples.py
leocjj: You're right, no needed

website/examples.py
leocjj: It is p.parent.name.lower() to order examples based on the example folder name. I changed the lambda parameter by r instead of p to improve readability, which is the path for each readme file. Sor r.parent.name.lower() will be the folder of that readme file.

website/examples.py
leocjj: logic changed to use example folders paths instead of the readme files paths

website/examples.py
leocjj: Left this way for simplicity, to avoid a manual parser that would increase complexity, for example, we don't know how many lines to read because we are not sure how many empty lines are before and after title line and description line.

leocjj:
Thanks @evnchn for your comments, commit sent!

leocjj:
Hi @falkoschindler and @rodja, busy days I think, same for me!
Just let me know if I can help with something else in this PR, I really want to see this deployed!
Have a nice day!

falkoschindler (commented):
Thanks again for your contribution, @leocjj! We're almost there!

I just reviewed the PR and cleaned it up here and there:
- Added a blank line after the title (as my auto-formatter would do). Now examples.py can assume to always find the title in line 0 and the description in line 1.
- Removed several checks and try-excepts. If an example can't be parsed, we'll notice and fix it.
- Added missing readmes and screenshots for new examples.
- Removed unused screenshots for Editable AG Grid and Editable Table examples. One screenshot per example is enough I guess.
- Removed the placeholder image. We should add missing screenshots rather than providing a fallback.
- Switched back to opening in the same tab (if the user doesn't actively chooses to use a new tab). There is debate about the optimal default for external links. But I chose to remove `new_tab=True` for consistency with other external links.

The automatic example detection is a great improvement on its own! I already spotted examples that were missing in the list. And it simplifies the code quite a bit because we don't need to "guess" the file path anymore.

There are just a few open questions I'd like to discuss with @rodja:

- [x] Do we want to leave the layout as it is? Or do we prefer popups? Note that the screenshots are rather bright in dark mode and don't really blend in with the remaining page.
- [x] Do we still need the "Single Page Application" example now that we have `ui.sub_pages`?
- [x] Do we still need the "Download Text as File" example? `ui.download.content` seems to be doing the same. Or it the `StreamingResponse` important? Then we might want to extend the readme.

leocjj:
Thank you @falkoschindler for your feedback. Please let me know if I can help with something else (this PR or another).

falkoschindler:
> Do we want to leave the layout as it is? Or do we prefer popups? Note that the screenshots are rather bright in dark mode and don't really blend in with the remaining page.

Popups are not very accessible and mobile-friendly, but dozens of permanently visible screenshots clutter the main page quite a lot.

As a proposal I limited the overview to three examples only and added a link to a dedicated examples page. Maybe this is good enough for now and can be improved later. What do you guys think?

evnchn:
3 examples in homepage is enough. But can we randomize which 3 to show on each page visit? This will make things interesting and ensure even exposure. 

```py
for example in random.sample(examples, k=3):
    example_link(example)
```

Image in dark mode: Can we do some CSS trickery? Possibly `filter: invert(100%);`? Maybe even toss `hue-rotate(180deg)` on top. 

I, uh, may want to not commit to this PR to avoid cluttering things up 🙇

evnchn:
```css
.body--dark .invert-on-dark {
  filter: invert(1) hue-rotate(180deg);
}
```

image

Passable?

falkoschindler:
**Random examples**

I'm not sure. Actually I picked specific examples that go nicely with the overall theme to optimize the visual appearance of the main page. At the same time I wanted to show relevant examples that motivate to visit the extra page.

**CSS trickery**

I don't think faking dark mode will look good.

Note that we're working on a new design for the main page. At least a touch-up, maybe even a larger redesign. Anyway, we don't have to aim for perfection now, since we'll revisit this section in near future anyway.

evnchn:
Sure. How about let's keep this PR strictly focused on the functional aspects for now, and leave the styling later (TM) as it will involve extra discussion, design, and coding. 

If that then I have no concerns, including deleting examples as mentioned in https://github.com/zauberzeug/nicegui/pull/5044#pullrequestreview-3458689420

falkoschindler (approved):

website/main_page.py
```diff
@@ -157,7 +157,9 @@ def create() -> None:
         section_heading('In-depth examples', 'Pick your *solution*')
         with ui.row().classes('w-full text-lg leading-tight grid grid-cols-1 sm:grid-cols-2 xl:grid-cols-3 gap-4'):
             for example in examples:
-                example_link(example)
+                if example.title in ['Authentication', 'Chat App', 'Todo List']:
```
evnchn: nit: use a set?

falkoschindler: Good catch! Let's use a set. 👍🏻

evnchn:
Wait @leocjj I just noticed: You are working off your branch's main

Let's not do that next time: read #5375 for more details.

Issue #5052, opened 2025-08-18
`Element` may remain alive substantially longer than their `Client`


burnpanck:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
import contextlib
import weakref

import nicegui
from nicegui import ui
import anyio

element_refs = weakref.WeakSet()

@ui.page("/")
def main():
    label = ui.label("Fresh!")
    element_refs.add(label)

async def monitor():
    """ Busy-loop monitoring elements. """
    n_prev = len(element_refs)
    loop_count = 0
    while True:
        # give other tasks a chance to run
        await anyio.sleep(0.01)
        # 
        loop_count += 1
        n_cur = len(element_refs)
        new_text = f"Iteration {loop_count}: Have {n_cur} clients" if n_cur != n_prev or not loop_count % 20 else None
        n_prev = n_cur
        for lbl in element_refs:
            c = lbl._client()
            if not c:
                print(f"In loop {loop_count}, had live element with dead parent ({n_prev=} {n_cur=})")
                continue
            if new_text is not None:
                lbl.set_text(new_text)
            del lbl

if __name__ in {"__main__", "__mp_main__"}:
    _orig_lifespan_context = nicegui.app.router.lifespan_context

    @contextlib.asynccontextmanager
    async def lifespan(nicegui_app):
        async with contextlib.AsyncExitStack() as stack:
            await stack.enter_async_context(_orig_lifespan_context(nicegui_app))
            tg = await stack.enter_async_context(anyio.create_task_group())
            stack.callback(lambda: tg.cancel_scope.cancel())
            tg.start_soon(monitor)
            yield

    nicegui.app.router.lifespan_context = lifespan

    ui.run()
```

### Description

[...]

PR #5053, opened 2025-08-18
Use weak references for parent slot to avoid cyclic references

falkoschindler:
### Motivation

In #5052 we noticed that there are still cyclic references between elements and their parent slots. To reduce work for the garbage collector and use much cheaper reference counting, we should avoid such cycles if possible.

### Implementation

Similar to other weak references, this PR introduces a private attribute `self._parent_slot` and a public property `self.parent_slot`. Internal access to the parent slot is optimized to reduce the number of dereferentiations.

### Progress

- [x] Pytests have been added (or are not necessary): I'm not sure how because we'd need to disconnected the client and wait.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -76,17 +76,18 @@ class Element(Visibility):
         self._deleted: bool = False
 
         client.elements[self.id] = self
-        self.parent_slot: Optional[Slot] = None
+        self._parent_slot: Optional[weakref.ref[Slot]] = None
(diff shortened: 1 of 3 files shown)


rodja:
> Pytests have been added (or are not necessary): I'm not sure how because we'd need to disconnected the client and wait.

If you open another page the old client disconnects...

rodja (approved):

PR #5057, opened 2025-08-19
Upgrade to Tailwind 4

falkoschindler:
### Motivation

NiceGUI 3.0 should upgrade its Tailwind dependency to version 4 which has been released quite a while ago.

### Implementation

This PR had to address several challenges:

- update package.json and extract_core_libraries to use Tailwind from node_modules/ instead of http://cdn.tailwindcss.com/
- remove warning about not using this version in production
- introduce CSS layers to properly integrate Tailwind with Quasar and NiceGUI's CSS rules
- update our hack to preserve borders of card children
- fix a conflict between Tailwind's "inline" class a Quasar components like QSelect, QCheckbox and QRadio
- update documentation and tests to use new border classes
- update links in documentation
- ~~update~~ remove the `.tailwind` API

An open question is what to do with the `.tailwind` API: It is getting more and more useless after Tailwind removed classes like "p-0" and mainly kept generic classes like "p-<number>", "p-(<custom-property>)", and "p-[<value>]". We could try to improve the API to support numbers beside literal strings, but it would require significant work. Apart from that the current PR causes several breaking changes because, e.g., [borders](https://tailwindcss.com/docs/border-style) now can start with "border-" _or_ "divide-". Therefore the literal string options change from "solid" etc. to "border-solid", "divide-solid" etc. It feels like Tailwind is getting to complex and flexible to keep trying to mirror its API. A potential way forward could be to release the new API as deprecated and discourage using it until we shut it down in NiceGUI 4.0.

### Progress

- [x] Remove the `.tailwind` API.


rodja:
I think we should `.tailwind`. Maybe we can deprecate it in 2.23.0 and drop it already in 3.0?

evnchn:
Just realized `.tailwind` was never used in any of the examples (quite surprising find). I think deprecation in 2.23.0 and removal in 3.0 is good. 

It can also help un-tie NiceGUI with a particular Tailwind version, which is useful when you consider UnoCSS allows you to use subset of V3, V3, V4, and/or neither.

evnchn:
Huh, the warning is gone entirely

https://github.com/tailwindlabs/tailwindcss/pull/18732

falkoschindler:
@evnchn Interesting! I just integrated Tailwind 4.1.13 in PR #5168.

Issue #5020, opened 2025-08-08
ui.echart() shows blank page using native=True in NiceGUI v2.19+


coolynx:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

@ui.page('/')
def page() -> None:
    ui.echart({
        'xAxis': {'type': 'category', 'data': ['Mon', 'Tue', 'Wed', 'Thu', 'Fri']},
        'yAxis': {'type': 'value'},
        'series': [{'type': 'line', 'data': [150, 230, 224, 218, 135]}],
    })

ui.run(native=True)
```

### Description

I took example code from https://nicegui.io/documentation/echart#echart_with_clickable_points

This code works on:
```sh
pip install --upgrade nicegui==2.18.0
```
But produces white page in future versions.
```sh
pip install --upgrade nicegui==2.19.0
```

using
```py
ui.run(native=True)
```

However, the code works as expected in browser.

So, the issue is **only** in **native app** - compiled or not.

### NiceGUI Version

2.19.0

### Python Version

Python 3.13.2

### Browser

Firefox

### Operating System

macOS

### Additional Context

<details> <summary><strong>Working 2.18.0 pip list output:</strong></summary>
<code>
Package                   Version
------------------------- -----------
aiofiles                  24.1.0
aiohappyeyeballs          2.4.4
aiohttp                   3.11.10
aiosignal                 1.3.2
altgraph                  0.17.4
annotated-types           0.7.0
anyio                     4.7.0
astroid                   3.3.8
attrs                     24.3.0
bidict                    0.23.1
bottle                    0.13.2
certifi                   2024.12.14
charset-normalizer        3.4.0
click                     8.1.7
dill                      0.3.9
docutils                  0.21.2
fastapi                   0.116.1
frozenlist                1.5.0
h11                       0.16.0
httpcore                  1.0.9
httptools                 0.6.4
httpx                     0.28.1
idna                      3.10
ifaddr                    0.2.0
isort                     5.13.2
itsdangerous              2.2.0
Jinja2                    3.1.6
macholib                  1.16.3
markdown2                 2.5.2
MarkupSafe                3.0.2
mccabe                    0.7.0
multidict                 6.1.0
nicegui                   2.18.0
[...]

Issue #3935, opened 2024-10-31
AG Grid JS error when running in native window


juledwar:
### Description

This only happens when running in a native window. I have an ag-grid on a page, and when loading the page I get this error in the console output:
```
js: TypeError: Failed to resolve module specifier "ag-grid-community". Relative references must start with either "/", "./", or "../".
```

This halts further set up of the page (anything constructed after the error does not appear).

I'm also seeing
```
Registered new object after initialization, existing clients won't be notified!
```
which seems to be a Qt WebEngine error, not sure if related to this.

Can be reproduced with this code (I just copy pasted what I am doing, you can probably repro without the extra classes and settings):
```py
from nicegui import ui

ui.aggrid(
    {
        'columnDefs': [
            {
                'field': 'value',
                'type': 'numericColumn',
                'width': 100,
            },
            {'field': 'stat'},
        ],
        'defaultColDef': {
            'flex': 1,
            'enableCellChangeFlash': True,
        },
        'rowData': [],
        'headerHeight': 0,
    }
).classes('min-w-fit')

ui.run(native=True, show=False)
```

juledwar:
nicegui==2.5.0
PyQt5==5.15.11
PyQt5-Qt5==5.15.15
PyQt5_sip==12.15.0
PyQtWebEngine==5.15.7
PyQtWebEngine-Qt5==5.15.15
QtPy==2.4.1

falkoschindler:
Strange. I can't reproduce the problem on macOS. What OS are you using? Can anyone else reproduce the error?

juledwar:
Ubuntu 24.04, under KDE using the QT view (not Gnome).

mbclarkengineering:
I can reproduce on rocky linux 8.6, on firefox browser. Same issue

evnchn:
Hey. 

Your browser may be old, or otherwise it doesn't support the ES module importmap. 

Check https://github.com/zauberzeug/nicegui/pull/4625#issuecomment-2812752236

Remind me to install Pale Moon on my laptop I brought to travel and do some testing. You can do that too. Basilisk is also another Firefox fork like Pale Moon that doesn't support ES modules (at least not completely)

evnchn:
Reminder to self: IIRC "shim mode" of es-module-shims seem to be more compatible against dynamic imports. It would require adding `-shim` suffix to link `modulepreload`, script type `importmap` and script type `module`. 

[...]

PR #5067, opened 2025-08-21
Always load ESM module shims, but async

falkoschindler:
### Motivation

Introducing the "nomodule" attribute in PR #4801 causes problems like #3935 or #5020 in some old browsers or webkits which don't understand "nomodule".

### Implementation

This PR removes the "nomodule" attribute, but adds "async" as recommended by the shim. This way the shims are loaded in the background and don't slow down the rendering.

### Progress

### DIFF
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -21,7 +21,7 @@
     {% endfor %}
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


rodja (approved):
Great solution!

PR #5069, opened 2025-08-21
Implement "user.should_see" and "click" for ui.tree

rodja:
### Motivation

As @davetapley asked for in #5068, user simulation currently misses the feature to work with `ui.tree`.

### Implementation

I implemented `UserInteraction.click` and `User.find` including a pytest to verify it's working (including node ids).

### Progress


falkoschindler (approved):
@rodja I reverted the breaking change we discussed and refactored the remaining code. The result looks pretty good to me. Please have a look if you agree.

Note that I made sure that the click simulation for `ui.tree` behaves like for `ui.select`: The `target` needs to be a label text, not a key. If we want to change it in the future, we should preserve consistency across all elements.

rodja:
Yes. Very nice. Thanks!

Issue #5085, opened 2025-08-28
Nested sub page not working if using `async` function


alexzaech:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

@ui.page('/')
@ui.page('/{_:path}')  # NOTE: our page should catch all paths
def index():
    ui.link('Go to main', '/')
    ui.link('Go to other', '/other')
    ui.sub_pages({
        '/': main,
        '/other': other,
    }).classes('border-2 p-2')

def main():
    ui.label('main page')

async def other():
    ui.label('sub page')
    ui.link('Go to A', '/other/a')
    ui.link('Go to B', '/other/b')
    ui.sub_pages({
        '/': sub_main,
        '/a': sub_page_a,
        '/b': sub_page_b
    }).classes('border-2 p-2')

def sub_main():
    ui.label('sub main page')

def sub_page_a():
    ui.label('sub A page')

def sub_page_b():
    ui.label('sub B page')

ui.run()
```

### Description

This is the original example. The only modification is using `async def other()`. As long as you use the page links, it seems to work as expected, but performing a browser refresh on `/other/a` or `/other/b` will result in a 404 error. `/other` refreshes without problems. 

Using `async` for any of the `sub_*` functions does not change the behavior. 

This seems to be related to https://github.com/zauberzeug/nicegui/issues/5071#issuecomment-3216187816

### NiceGUI Version

2.23.3

### Python Version

3.13.7

### Browser

Chrome

### Operating System

macOS

rodja:
Thanks for reporting @alexzaech. This is a tricky one. Extending our tests to cover this scenario was easy (see #5089). But I have not come up with a fix yet.

PR #5089, opened 2025-08-29
Fix direct opening of async nested sub pages

rodja:
### Motivation

As @alexzaech reported in #5085, opening a nested sub page which is inside a async sub page shows the 404 error page.

### Implementation

Currently this PR only reproduces the problem in a screen test. 
A minimal reproduction:

```py
from nicegui import ui

@ui.page('/')
@ui.page('/{_:path}')
def index():
    ui.sub_pages({'/': main})

async def main():
    ui.sub_pages({
        '/': sub_main,
        '/sub': sub_sub_page,
    })

def sub_main():
    ui.link('Go to sub page', '/sub')

def sub_sub_page():
    ui.label('sub page')

ui.run()
```

If you access `/` everything works fine. And clicking on "Go to sub page" also works. But if you directly open `/sub` you see a 404 page.

The problem is this: due to the `async` keyword the execution of `main` is a coroutine which is not executed until the event loop ticks it forward. Hence the sub pages element in main is not available and 

https://github.com/zauberzeug/nicegui/blob/78bf3151b645573834e46aecd8e2c50977d4ac26/nicegui/elements/sub_pages.py#L111-L115

determines that the 404 should be shown. It works for the `/` because `match.remaining_path` is empty.

### Progress

  - [x] try to find a way to allow 404 http errors **and** evaluate async builder functions properly
  - [ ] verify that the approach will still work with 3.0 (especially #5005, where SPA root page is processed in 404 handler)

### DIFF
--- a/nicegui/functions/navigate.py
+++ b/nicegui/functions/navigate.py
@@ -74,12 +74,10 @@ class Navigate:
             parsed = urlparse(path)
             if not parsed.scheme and not parsed.netloc and \
                     any(isinstance(el, SubPages) for el in context.client.elements.values()):
-                client = context.client
-
-                async def navigate_sub_pages():
+                async def navigate_sub_pages(client: Client) -> None:
                     with client:
                         await client.sub_pages_router._handle_navigate(path)  # pylint: disable=protected-access
-                background_tasks.create(navigate_sub_pages(), name='navigate_sub_pages')
+                background_tasks.create(navigate_sub_pages(context.client), name='navigate_sub_pages')
                 return
 
         context.client.open(path, new_tab)
--- a/nicegui/sub_pages_router.py
+++ b/nicegui/sub_pages_router.py
@@ -49,10 +49,7 @@ class SubPagesRouter:
             callback(path)
         for child in context.client.layout.descendants():
             if isinstance(child, SubPages):
-                try:
-                    child._show()  # pylint: disable=protected-access
-                except ValueError:
-                    pass
+                child._show()  # pylint: disable=protected-access
         return await self._can_resolve_full_path(context.client)
 
     async def _handle_navigate(self, path: str) -> None:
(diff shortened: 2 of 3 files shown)


rodja:
Ok. I found an easy solution to the issue: simply skipping the early exit if the sub page builder is a coroutine:

https://github.com/zauberzeug/nicegui/blob/b8e452b6bd3dc52bd898a77925d6fc1ddde50feb/nicegui/elements/sub_pages.py#L111-L114

But that means we can not return 404 http status anymore (if any sub pages builder in path is async).

alexzaech:
Hi @rodja. This resolves the issue reported in #5085. 
Not showing the 404 error page for subpages using async is fine for me since the `404: sub page ... not found` message is still displayed. Even if `show_404=False` is used and `/other/c` is visited, the content of `/other` is shown without an error message. For my use case, this is a nice "feature"

rodja:
@alexzaech True. But consider the case where you have static assets or other routes defined via FastAPI: as soon as you use an async page builder, every response will have http status code `200`. Even if the url is not resolving to one of the defined routes.

alexzaech:
You're right, @rodja . This isn't right, and unfortunately, I didn't consider this circumstance. Static assets and FastAPI routes must continue to work.

rodja:
I think I found a way: when creating the FastAPI page, we should to give sub pages the opportunity to process the first part of an async builder by yielding to the event loop (via `await asyncio.sleep(0)`). After that, we can  reliably check if a sub page produces 404. This approach not only works for my newly introduced tests, but nicely simplifies the overall code in `sub_pages.py`.

As an addition, I discovered some routes bleeding from one test into the other. By fixing that the pytest fixtures are now much simpler. And it uncovered some misconceptions in a few test cases -- which had to be fixed.

rodja:
Unfortunately that is not enough: async sub pages currently do not set `match(None)` (and hence `has_404` stays `False`) when they have a remaining path which can not be resolved. This means that any nested sub pages inside an async sub pages will always report 200. But I might have an idea...

rodja:
Yes. I managed to put all the logic into one method (`sub_pages_router._can_resolve_full_path`). This must be called after page is build and after a sub pages navigation took place. Because it is async, there was some trickery involved to keep the nice API without awaiting `ui.sub_pages`, `navigate.to` etc.

falkoschindler (approved):

PR #5093, opened 2025-08-30
Update checkbox value type to support indeterminate state

daya0576:
### Motivation
The checkbox component already supports `null` value for indeterminate state. This change simplely update type annotation to reflect this behavior.

A simpe example:
```python
from nicegui import ui

checkbox = ui.checkbox("checkbox")

with ui.row():
    ui.button("True", on_click=lambda: setattr(checkbox, "value", True))
    ui.button("False", on_click=lambda: setattr(checkbox, "value", False))
    ui.button("Indet", on_click=lambda: setattr(checkbox, "value", None))

ui.run()
```

### Progress

### DIFF
--- a/nicegui/elements/checkbox.py
+++ b/nicegui/elements/checkbox.py
@@ -8,7 +8,10 @@ from .mixins.value_element import ValueElement
 
 class Checkbox(TextElement, ValueElement, DisableableElement):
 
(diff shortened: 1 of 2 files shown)


falkoschindler (approved):
Thank for the pull request, @daya0576!

I guess it makes sense to officially support `None` values here.
For consistency I updated `ui.switch` as well.

PR #5097, opened 2025-08-31
implement an example chat interface using NiceGUI and Google's Generative AI model, enabling users to interact with the Gemini model through a web-based chat application that streams the LLM's reponse.

CVxTz:
### Motivation

This PR focuses on enabling real-time streaming of AI responses within the chat interface. Instead of waiting for complete responses, users will see text appear incrementally as it's generated by the Gemini model. This provides a more natural conversation experience and better user feedback during processing.
This is in response to a need expressed by a reddit user that could not get the streaming+binding feature to work: https://www.reddit.com/r/nicegui/comments/1n4ajzm/comment/nblnu2g/?context=1 

### Implementation

The key lies in binding the markdown component to the dictionary. As each chunk is received, it's appended to the shared data structure. When cleared and re-rendered, this binding ensures the UI updates immediately to reflect new content. `data['response']`

### Result

Peek 2025-08-31 16-05

### Progress

### DIFF
--- a/examples/streaming_chat/main.py
+++ b/examples/streaming_chat/main.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+from langchain_google_genai import ChatGoogleGenerativeAI
+
+from nicegui import ui
+
+GOOGLE_API_KEY = 'not-set'  # TODO: set your GEMINI API key here
+
+
+@ui.page('/')
+def main():
+    llm = ChatGoogleGenerativeAI(
+        api_key=GOOGLE_API_KEY, model="gemini-2.5-flash", temperature=0, max_retries=10
+    )
+
+
+    async def send() -> None:
+        question = text.value
+        text.value = ''
+
+        with message_container:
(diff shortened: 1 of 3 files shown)


rodja:
@CVxTz Thanks for the PR, but do we really need it? We already have a langchain example: https://github.com/zauberzeug/nicegui/blob/main/examples/chat_with_ai/main.py

CVxTz:
This one streams the output, can be useful for long answers to avoid making the user wait too long and is the common behavior in most chat UIs. But if you think its not necessary I can close the PR.

rodja:
But our chat example also streams the output. Or not?

CVxTz:
You are right ! I don't know how I missed it. Sorry :) 
The code is out of date with the langchain requirements so I could not run it as is to check.

rodja:
> The code is out of date with the langchain requirements so I could not run it as is to check.

Oh. would you like to create a PR to bring it up to date?

PR #5102, opened 2025-09-02
Pytests can load a main file with runpy

rodja:
### Motivation

With the introduction of the root page concept, we need to rethink our pytest infrastructure (he `module_under_test` marker and the modular setup with a startup function)

### Implementation

`main.py` is loaded via `runpy.run_path` in the fixture setup.

### Progress


website/documentation/content/section_testing.py
```diff
@@ -1,12 +1,92 @@
...
+def project_code():
+    with python_window(classes='max-w-[820px] w-full h-48'):
+        ui.markdown('''
+            ```python
+            from nicegui import ui
+
+            def root() -> None:
+                ui.button('Click me', lambda: ui.notify('Hello World!'))
+
+            if __name__ in {'__main__', '__mp_main__'}:
```
falkoschindler: I don't think we need the main guard anymore - right?

PR #5104, opened 2025-09-03
Use `str.translate` instead of repeated `str.replace` where possible

falkoschindler:
### Motivation

I stumbled upon the built-in function `str.translate` which can be used to replace multiple characters more efficiently than calling `str.replace` repeatedly. This reminded me of our `client.build_response` method which calls `str.replace` five times to escape special characters for HTML.

### Implementation

`client.build_response` is further optimized by globally defining a translation dictionary which is re-used by every client.
~~While being at it, two other occurrences of repeated `str.replace` calls have been replaced.~~

### Progress

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -33,6 +33,14 @@ if TYPE_CHECKING:
 
 templates = Jinja2Templates(Path(__file__).parent / 'templates')
 
+HTML_ESCAPE_TABLE = str.maketrans({
+    '&': '&amp;',
+    '<': '&lt;',
+    '>': '&gt;',
+    '`': '&#96;',
+    '$': '&#36;',
+})
+
 
 class Client:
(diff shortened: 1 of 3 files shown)


evnchn:
I think inline `str.translate` is where the readability fell off, not that `str.translate` itself is less readable. 

Consider:
- `str.translate(NEWLINES_TO_SPACES)`
- `str.translate(ENSURE_VALID_NAME)`

It's likely just personal preference at this point, performance gains are extremely minimal to none.

rodja (approved):

rodja:
Could this already be be done in 2.24?

falkoschindler:
@rodja Yes, that should be possible. First I thought there were more places that would benefit from `str.translate` so I branched off of 3.0 to avoid massive merge conflicts. But in the end the diff is pretty small. Let's merge into 3.0 and cherry pick on main. (I don't see a better way without re-creating the whole PR.)

PR #5107, opened 2025-09-03
Introduce Event class for sending events between different parts of the code

falkoschindler:
### Motivation

While working on PR #5005, we noticed that some projects like our ROS2 example might depend on the shared nature of the long-living auto-index client. If there is some kind of "event" in the data model (or hardware controller), they can simply update the UI if there is only one and exactly one client. When switching to page functions, they updating UI becomes more complicated in this scenario.

Therefore we decided to move our event system from [RoSys](https://github.com/zauberzeug/rosys/) to NiceGUI. It provides a simple API to define events, subscribe arbitrary callbacks to them, and to emit/call them if the event happens.

### Implementation

The core implementation is taken from RoSys. But on a second look some details have changed, partly by relaxing some assumptions or because we can make better use of NiceGUI's infrastructure.

- I chose the method names "subscribe" and "subscribe_ui" instead of "register" and "register_ui".
- Callbacks can still be sync or async.
- Callbacks can - but don't have to - receive arguments.
- Callbacks are automatically called within the same slot where they subscribed.

### Progress

  - [x] Merge `subscribe` and `subscribe_ui` into a single `subscribe(..., unsubscribe_on_disconnect: bool | None = None)`.

### DIFF
--- a/nicegui/__init__.py
+++ b/nicegui/__init__.py
@@ -4,6 +4,7 @@ from .app.app import App
 from .client import Client
 from .context import context
 from .element_filter import ElementFilter
+from .event import Event
 from .nicegui import app
 from .page_arguments import PageArguments
 from .version import __version__
@@ -13,6 +14,7 @@ __all__ = [
     'App',
     'Client',
     'ElementFilter',
+    'Event',
     'PageArguments',
     '__version__',
     'app',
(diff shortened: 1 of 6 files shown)


rodja (approved):
I like it very much. Especially the automatic unsubscription 😎

PR #5058, opened 2025-08-19
Improve support for languages with right-to-left text direction

falkoschindler:
### Motivation

As noticed by @M6stafa in #5045, Quasar elements and the connection popup haven't been rendered correctly for languages with right-to-left text direction.

### Implementation

This PR loads the quasar.rtl*.css files and adjusts the CSS rules for the connection popup for `:dir(ltr)` and `:dir(rtl)` respectively.

It can be tested with some input like
```py
ui.input(label='متن ورودی', placeholder='متن خود را وارد کنیم')
```

and choosing a language like
```py
ui.run(language='fa-IR')
```

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests are not necessary.
- [x] Documentation is not necessary.

PR #5037, opened 2025-08-12
Remove deprecated code and APIs

falkoschindler:
This PR removes code and APIs which have been marked as deprecated. Exceptions:

- Since we probably won't update to Tailwind 4, the documentation links to version 3 remain unchanged.
- Parameters for subclassing `ui.element` are already updated in PR #5021.
- `previous_value` will only be enforced in NiceGUI 4

PR #5109, opened 2025-09-04
NiceGUI 3.0

### LINKED ISSUE #3854: AG-Grid's `'rowSelection': {'mode':"multiRow",'headerCheckbox':False}` won't work

### LINKED ISSUE #4892: Client-Side Memory Leak
### First Check

### Example Code

```python
from nicegui import ui
from datetime import datetime

class TimeManager:
    def __init__(self):
        self.setup_ui()

    def setup_ui(self):

        with ui.row().classes('flex w-full justify-between'):
            ui.label('Current Time : ').classes('flex-start')
            self.clock_label = ui.label('...').classes('flex-end')

        ui.timer(0.1, self.update_clock)

    def update_clock(self):
        """Update the clock label with the current time."""
        current_time = datetime.now()
        self.clock_label.set_text(f'{current_time:%Y-%m-%d %H:%M:%S}')

def build_gui():
    ui.separator()
    with ui.row().classes('w-full no-wrap'):
        with ui.column().classes('md:w-4/12 w-full'):
            TimeManager()

    # ui.image('https://picsum.photos/id/377/640/360')

    build_gui()

    ui.run()
```

### Description

Hopefully this is just something I am doing wrong and is a quick fix! I've been encountering a slow memory leak in the client-side (browser tab) that seems to be tied to a simple static ui.image(). I have tried to provide a single file MWE for testing. The layout is similar to the larger application I am developing, where there are several separate class files that are instantiated by the main file.  It can likely be made even simpler, but maybe the error is in the order. 

In any case, if the image is uncommented, I get a slow leak. If I omit the image, there is no leak. The timer refresh is fast only as an example, and doesn't seem to leak on its own. I test for leaks by right clicking in Google Chrome and using the Inspect, then recording a couple hundred seconds of the performance with the memory button checked. 

Here's no leak with just the timer
Image

And here's the leak with the ui.image uncommented.

Image

### NiceGUI Version

2.19.0

### Python Version

Python 3.10.12

### Browser

Chrome

### Operating System

macOS

### Additional Context

_No response_

falkoschindler:
This PR collects all changes and release notes for NiceGUI 3.0.
It shouldn't get squashed.

---

### New features and enhancements, breaking changes and migration guide

This major release introduces several new features and enhancements, as well as breaking changes.
We always try to keep breaking changes to a minimum, guide you through the migration process using deprecation warnings, and provide migration instructions.
Please read the following release notes carefully to understand the changes and adapt your code accordingly before upgrading.

- **Drop shared auto-index pages** (#4964, #5005 by @rodja, @falkoschindler)

	⚠️ BREAKING: UI elements defined in global scope have been added to a static shared "auto-index client" served at "/".
	This caused a multitude of problems throughout the code base, so we decided to remove this shared client.

	In version 3.0 you have the following options:

	- Keep putting (all) UI elements in global scope. We call such apps without page functions **"NiceGUI scripts"**. They will automatically re-evaluated inside an implicit page function when visiting "/". This is almost a drop-in replacement for the auto-index client, but:
	    - NiceGUI scripts can't contain page functions.
	    - The UI isn't shared but re-created on every page visit.

	- Wrap all UI in a singe function and pass it to the new positional **`root` parameter in `ui.run`**. This is especially handy in combination with `ui.sub_pages`. This way you can create rich single-page applications without worrying about defining the correct routes with `@page` decorators.

	- Use page functions for all your pages, including the index page at "/".

	Note that we are introducing a new `Event` class (see below). Together with the binding module this helps to synchronize long-living objects with short-living UI without relying on a long-living shared client.

- **Introduce a new `Event` class** (#5107 by @falkoschindler, @denniswittich)

	In some cases it can be tricky to send updates to the UI, especially after removing the shared auto-index client. Therefore we are introducing a new event system which allows you to define custom events, subscribe handlers, and emit or call them.

- **Introduce a `strict` option for binding methods** (#5040 by @denniswittich, @codingpaula, @falkoschindler)

	When binding from or to object properties or dictionary items, the binding failed silently if one of the attributes doesn't exist. In case of dictionaries this was intended, because they could be are often used to bind to persistent storage which is empty by default. But for object properties this can lead to very subtle bugs, e.g. after renaming properties and not updating the attribue names in binding functions.

	⚠️ BREAKING: In 3.0 object properties will be checked for existence by default. Missing dictionaries will continue to be ignored.

	You can fine-tune this behavior using the `strict: bool | None = None` parameter (`None`: check object properties and ignore dictionaries).

- **Improve support for languages with right-to-left text direction** (#5045, #5058 by @M6stafa, @falkoschindler)

	NiceGUI now includes the required Quasar library to correctly display UI in languages with a different text direction. The language can be chosen via the `language` parameter in `ui.run`.

- **Make props, classes and style observable to send updates automatically** (#4978 by @lawrenceakka, @falkoschindler, @rodja)

	Calling `.props()`, `.classes()` or `.style()` will automatically update the UI element. A subsequent `.update()` is not necessary anymore.

	⚠️ BREAKING: When overwriting the `update` method in custom elements, infinite recursions can occur. If, e.g., the `update` methods uses `.prop()` before calling `super().update()`, the `.prop()` call will cause an infinite cycle. Wrap it with `with self._props.suspend_updates():` (and similar for classes and style) to pause automatic updates in such cases.

- **Use `q-scroll_area` for `ui.log` to preserve a fixed height** (#4925, #4935 by @e-nesse, @falkoschindler)

	We noticed that the height of `ui.log` could be affected by its content, which is unexpected. Therefore we decided to use a scroll area for a more robust layout.

	⚠️ BREAKING: Now the width of `ui.log` can collapse when placed inside a container with an unspecified width. Either give the container some width or specify the width of `ui.log`.

- **Introduce `ValueChangeEventArguments.previous_value`** (#4410, #4456 by @me21, @falkoschindler, @evnchn)

	In some situations it might be helpful to have access to both, the current and the previous value. Therefore we added `previous_value` to the `ValueChangeEventArguments`.

	⚠️ BREAKING: Custom elements that emit `ValueChangeEventArguments` need to provide the previous value.

- **Let `ui.clipboard.read()` return `None` if the clipboard API isn't available** (#4236, #4237 by @weinibuliu, @falkoschindler, @evnchn)

	The `read()` function used to return an empty string if the clipboard API is not available. This was indistinguishable from an empty clipboard.

	⚠️ BREAKING: Now the `read()` function returns `None` if the clipboard API is not available.

- **Remove deprecated code and APIs** (#5037 by @falkoschindler)

	Several deprecated code has been removed. This shouldn't cause any breaks.

- **Drop support for Python 3.8** (#4329 by @Solipsistmonkey, @falkoschindler)

	Almost one year after Python 3.8 reached its end-of-life, it was time to drop support. This allowed us to update the code to a newer standard and resolve some issues with Python dependencies that already dropped 3.8 a while ago.

### Testing

- Simplify pytests by loading a main file with runpy (#5102 by @rodja, @falkoschindler)

### JavaScript Dependencies

The infrastructure for managing node packages has been improved significantly (#4163, #5021 by @simontaurus, @evnchn, @falkoschindler).
The following JavaScript dependencies have been updated to the latest versions (#5034, #5057 by @falkoschindler):

- @tailwindcss/browser: 3.4.10 → 4.1.11
- es-module-shims: 1.10.0 → 2.6.2
- quasar: 2.16.9 → 2.18.2 (resolves #4892 by @eddie3ruff, @evnchn, @xaptronic, @falkoschindler)
- socket.io: 4.7.5 → 4.8.1
- vue: 3.4.28 → 3.5.18
- ag-grid-community: 32.1.0 → 34.1.1 (resolves #3854 by @python-and-novella, @mockey, @falkoschindler)
- codemirror: 6.0.1 → 6.0.2
- @codemirror/language-data: 6.5.1 (new)
- @codemirror/theme-one-dark: 6.1.3 (new)
- @uiw/codemirror-themes-all: 4.24.2 (new)
- @babel/runtime: 7.28.2 (new)
- echarts: 5.5.1 → 6.0.0
- echarts-gl: 2.0.9 (unchanged)
- nipplejs: 0.10.2 (unchanged)
- vanilla-jsoneditor: 0.23.8 → 3.8.0
- leaflet: 1.9.4 (unchanged)
- leaflet-draw: 1.0.4 (unchanged)
- mermaid: 11.0.2 → 11.9.0
- plotly.js: 2.35.0 → 3.1.0
- three: 0.168.0 → 0.179.1
- @tweenjs/tween.js: 25.0.0 (unchanged)


evnchn:
Would be better if we can document also the breaking changes which we deprecated for a while and is breaking now with #5037, so that just looking at the 3.0 release notes is enough to bring you from as old as 2.0.0 to 3.0.0?

Currently one would need to read through from 2.0.0 to 3.0.0 release notes to hunt for any deprecated code which they may have used.

evnchn:
Also can we potentially make the searching a bit easier by re-phrasing "upstream breaking change" to "upstream changes which _broke_ our code" in old release notes, so that searching for "break" or "breaking" yields **only** the breaking changes? 

Or the other way round with "our breaking changes" so that we can search by "our break" / "our breaking"

evnchn:
About #5037 

High severity (can error out the whole server, or popular elements):

- `app.add_static_file` and `app.add_media_file` raises `FileNotFoundError` instead of `ValueError` in case of non-existing files. Since `FileNotFoundError` is NOT a subclass of `ValueError`, existing code which `except ValueError` will fail to catch the new `FileNotFoundError`. 
- `ui.aggrid` `run_column_method` is gone. Use `run_grid_method` instead. 
- `ui.table` calling `add_rows()`/`remove_rows()` with variable-length arguments no longer works. Pass a list instead or use `add_row()`/`remove_row()` for a single row.

Medium severity (less popular elements and functions)

- `ui.open` is gone. Use `ui.navigate.to`
- (Where is this? Native?) `load_html` parameter `content` renamed to `html` to match the parameter name in pywebview>=5.4
- `nicegui.testing.conftest` is gone and you can no longer import it. Use pytest_plugins = ["nicegui.testing.plugin"] instead.

Low severity (you can get away with it in certain scenarios):

- `element.on` parameter `js_handler` now has type `str` instead of `str | None`. You can pass a `None` and it still works, but your type checker won't be happy, and we can't promise it will work for much longer afterwards.

falkoschindler:
@evnchn We updated the release notes to include deprecations.

Updating old release notes is an interesting idea. But it will probably break (pun not intended) in the future as soon as someone forgets that "break" is a forbidden word for non-breaking changes.
Alternatively you can search for "⚠️". I just don't understand why v2.7.0 is popping up. Maybe because of my falcepalm 🤦🏻‍♂️.

evnchn:
I read through the release notes. 

Although may not affect everyone, aren't the JavaScript version bumps also potentially breaking? 

I think I'll do a search.

evnchn:
THIS MAY NOT BE ACCURATE!!!

JavaScript dependencies with major version bump:

- @tailwindcss/browser: 3.4.10 → 4.1.11 (https://github.com/zauberzeug/nicegui/pull/5057 by @falkoschindler) 
  - Upgrade guide at https://tailwindcss.com/docs/upgrade-guide
- es-module-shims: 1.10.0 → 2.6.2 
  - No breaking. Some deprecation: https://github.com/guybedford/es-module-shims/releases?q=deprecate&expanded=true

- ag-grid-community: 32.1.0 → 34.1.1 (resolves https://github.com/zauberzeug/nicegui/issues/3854 by @python-and-novella, @mockey, @falkoschindler)
  - Does NOT semantic version! Check https://www.ag-grid.com/changelog/ carefully
- echarts: 5.5.1 → 6.0.0
  - Breaking changes documented at https://github.com/apache/echarts/releases/tag/6.0.0

- vanilla-jsoneditor: 0.23.8 → 3.8.0

  - No breaking changes nor deprecations. Still, check https://github.com/josdejong/svelte-jsoneditor/releases
- plotly.js: 2.35.0 → 3.1.0
  - Removals documented at https://github.com/plotly/plotly.js/releases/tag/v3.0.0
- three: 0.168.0 → 0.179.1
  - Look at the migration guide https://github.com/mrdoob/three.js/wiki/Migration-Guide



JavaScript dependencies with minor version bump:
- quasar: 2.16.9 → 2.18.2 (resolves https://github.com/zauberzeug/nicegui/issues/4892 by @eddie3ruff, @evnchn, @xaptronic, @falkoschindler)
- socket.io: 4.7.5 → 4.8.1
- vue: 3.4.28 → 3.5.18
- codemirror: 6.0.1 → 6.0.2
- mermaid: 11.0.2 → 11.9.0



New JavaScript dependencies:
- @codemirror/language-data: 6.5.1
- @codemirror/theme-one-dark: 6.1.3
- @uiw/codemirror-themes-all: 4.24.2
- @babel/runtime: 7.28.2

Unchanged JavaScript dependencies:
- echarts-gl: 2.0.9 (unchanged)
- nipplejs: 0.10.2 (unchanged)
- leaflet: 1.9.4 (unchanged)
- leaflet-draw: 1.0.4 (unchanged)
- @tweenjs/tween.js: 25.0.0 (unchanged)

ustas4:
Paragraph 12. Uploade files.
Unfortunately, I tested the read and text functions to load a small txt file. Both functions e.file.read() and e.file.text(encoding='utf-8') returns the error: nicegui: 'coroutine' object is not iterable

Linux, chrome,  nice gui 3.2.0

evnchn:
Hello @ustas4 

> 'coroutine' object is not iterable

coroutine, in human words, means "something you should `await`". So yeah, add that and it should work. 

If you are still confused feel free to open a new discussion at https://github.com/zauberzeug/nicegui/discussions. 

Let's not discuss here since it's a closed PR and we don't really read the comments down here often.

PR #5112, opened 2025-09-05
`ui.log` multiline-push scroll-to-bottom via component approach

### LINKED ISSUE #4830: `ui.log` doesn't scroll to bottom when pushing multiple lines at once
As shown in the following example, the log view stops scrolling to the bottom when multiple lines are pushed at once.

```py
from nicegui import ui
import time

log = ui.log().classes('w-full h-30vh')

def write_log():
    log.push(f'Log entry at {time.strftime("%H:%M:%S")}')
    log.push(f'Log entry at {time.strftime("%H:%M:%S")}')
    log.push(f'Log entry at {time.strftime("%H:%M:%S")}')
    log.push(f'Log entry at {time.strftime("%H:%M:%S")}')

ui.timer(0.5, write_log, active=True)

ui.run()
```

_Originally posted by @pandabearcodes in https://github.com/zauberzeug/nicegui/discussions/3909#discussioncomment-13342547_

evnchn:
### Motivation

This PR fixes #4830 and #3909 via introducing a Vue component for `ui.log`, which includes logic to scroll-to-bottom in a smart manner. 

### Implementation

- Use `log.js` component in `log.py` element
- `ui.log`:
  - `beforeUpdate()` checks if the user at the bottom
  - `updated()` scrolls to the bottom if the user were at the bottom

### Progress

- [x] Did NOT do pre-commit since I'm using a new Windows computer and is significantly underpowered. 
  - Let's look at the pipeline then check this task. 

### Additional explanation from GitHub Copilot which wrote the `log.js`

In Vue.js, the `beforeUpdate` and `updated` lifecycle hooks are part of the component's lifecycle and are triggered during the update phase. Here's what they do:

#### `beforeUpdate`
- **When it runs**: This hook is called **before the DOM is updated** with the new data or state.
- **Purpose**: It allows you to perform actions or calculations based on the current state of the DOM before Vue applies changes.
- **In your code**:
  ```javascript
  beforeUpdate() {
      const container = this.$refs.logContainer;
      if (container) {
          this.shouldScroll = container.scrollTop + container.clientHeight >= container.scrollHeight;
      }
  }
  ```
  - This checks if the user has scrolled to the bottom of the log container.
  - `this.shouldScroll` is set to `true` if the user is at the bottom, or `false` if they are not.
  - This ensures that the component knows whether it should auto-scroll after the DOM updates.

---

#### `updated`
- **When it runs**: This hook is called **after the DOM has been updated** with the new data or state.
- **Purpose**: It allows you to perform actions that depend on the updated DOM.
- **In your code**:
  ```javascript
  updated() {
      const container = this.$refs.logContainer;
      if (container && this.shouldScroll) {
          container.scrollTop = container.scrollHeight;
      }
  }
  ```
  - If `this.shouldScroll` is `true` (i.e., the user was at the bottom before the update), the log container is scrolled to the bottom.
  - This ensures that new content added to the log remains visible to the user.

---

#### How They Work Together
1. **`beforeUpdate`**: Checks the current scroll position before the DOM changes.
2. **`updated`**: Scrolls the container to the bottom if the user was already at the bottom before the update.

This combination ensures a smooth auto-scrolling behavior for the log container while respecting the user's manual scrolling. If the user scrolls up, the auto-scroll stops until they scroll back to the bottom.

### DIFF
--- a/nicegui/elements/log.js
+++ b/nicegui/elements/log.js
@@ -1,20 +1,18 @@
 export default {
-    template: `<div class="nicegui-log" ref="logContainer"><slot></slot></div>`,
+    template: `<div><slot></slot></div>`,
     data() {
         return {
             shouldScroll: true,
         };
     },
     beforeUpdate() {
-        const container = this.$refs.logContainer;
-        if (container) {
-            this.shouldScroll = container.scrollTop + container.clientHeight >= container.scrollHeight;
+        if (this.$el) {
+            this.shouldScroll = this.$el.scrollTop + this.$el.clientHeight >= this.$el.scrollHeight;
         }
     },
     updated() {
-        const container = this.$refs.logContainer;
-        if (container && this.shouldScroll) {
-            container.scrollTop = container.scrollHeight;
+        if (this.$el && this.shouldScroll) {
+            this.$el.scrollTop = this.$el.scrollHeight;
         }
     }
 };
--- a/nicegui/elements/log.js
+++ b/nicegui/elements/log.js
@@ -1,0 +1,20 @@
+export default {
+    template: `<div class="nicegui-log" ref="logContainer"><slot></slot></div>`,
+    data() {
+        return {
+            shouldScroll: true,
+        };
+    },
+    beforeUpdate() {
+        const container = this.$refs.logContainer;
+        if (container) {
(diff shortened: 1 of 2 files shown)


evnchn:
Good point. Just another reminder that Copilot isn't perfect. Will get to fixing when at home.

PR #5113, opened 2025-09-06
Make NiceGUI slash-agnostic by default

rodja:
### Motivation

A long time ago, @xinetzone created the feature request #2792 to add `APIRouter` support to NiceGUI On Air. A few days ago I stumbled over the issue again when working on #5083. The issue seems to be related to https://github.com/Kludex/starlette/issues/869 and https://github.com/Kludex/starlette/discussions/2514. When accessing `/router/` it works, but `/router` results in a 307 redirect to the **absolute** url http://localhost/router/ which does not include the `root_path` or path-prefix.

### Implementation

We introduce `SlashAgnosticMiddleware` which rewrites the path of **incoming requests** if the current path does not have a match but the alternative (with or without a slash) does. This avoids the (potentially wrong) redirect **response** which would otherwise be send from Starlette. From NiceGUI's point of view I think it is ok to be slash-agnostic by default (which is a breaking change and hence should be released with 3.0). But of curse it can be turned of by the developer with a parameter.

### Progress

- [ ] Pytests have been added.


falkoschindler:
I'm a bit worried about cases like this:
```py
@ui.page('/page/')
def page():
    ui.link('Go to sub page', 'sub')

@ui.page('/page/sub')
def sub():
    ui.label('sub page')
```

When visiting /page, the middleware won't redirect to /page/. This causes the link with target "sub" to point to /sub instead of /page/sub. This way existing websites might suddenly have broken links.

evnchn:
The edge case pointed out is a valid one. Note that despite the name `/sub` it has nothing to do with the `ui.sub_pages`, just to be clear. 

> This avoids the (potentially wrong) redirect response which would otherwise be send from Starlette

If slash agnostic via rewrite is problematic, can we fix the redirect response instead?

evnchn:
Alternatively, consider `history.replaceState` to modify the client into thinking it is at `/page/` despite it originally fetched the HTML of `/page`. Check https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState

rodja:
Yes @evnchn. In my opinion that would be the best option. But also that would more complexity. An alternative strategy could be to replace Starletts middleware with a custom one which performs redirects with the correct path. Not my favourite, but will be non-breaking and could be postponded to 3.x.

evnchn:
https://github.com/Kludex/starlette/issues/869#issuecomment-1686790185

> undocumented option in Starlette, to disable its (potentially flawed) slash 307 redirects

🤡, but maybe we can give that a shot. 

---

Also I think we are close. Can we return a **proper** redirect response after we computed in the `SlashAgnosticMiddleware` that we have a match?

rodja:
I could not reproduce the issue without NiceGUI On Air. Consider a modified `main.py` in [examples/nginx_subpath](https://github.com/zauberzeug/nicegui/tree/main/examples/nginx_subpath) with this code:

```py
import uvicorn
from nicegui import app, ui

fastapi_app = FastAPI()

@ui.page('/')
def index(request: Request):
    ui.label(f'URL path: {request.url.path}')
    ui.label(f'X-Forwarded-Prefix: {request.headers.get("X-Forwarded-Prefix", "")}')
    ui.label(f'Root path: {request.scope.get("root_path", "")}')

    ui.link('Navigate to a subpage.', 'subpage')


@ui.page('/subpage')
def subpage(request: Request):
    ui.label(f'URL path: {request.url.path}')
    ui.label(f'X-Forwarded-Prefix: {request.headers.get("X-Forwarded-Prefix", "")}')
    ui.label(f'Root path: {request.scope.get("root_path", "")}')

    ui.link('Navigate to root.', '/')

ui.run_with(fastapi_app, mount_path='/mount')

@fastapi_app.get('/')
def get_root():
    return {'message': 'Hello, FastAPI! Browse to /mount to see the NiceGUI app.'}

if __name__ == '__main__':
    uvicorn.run('main:fastapi_app', log_level='info', reload=True, port=8080, host='0.0.0.0')
```

When running `docker compose pull && docker compose up` to start the setup access and redirects work fine. It seems that #1501 fixed the redirect by re-injecting the X-Forwarded-Prefix if it is missing.

But with On Air, this reproduction fails:

```py
import uvicorn
from fastapi import FastAPI
from nicegui import ui

fastapi_app = FastAPI()

@ui.page('/')
def index():
    ui.link('go to subpage', '/subpage/')

@ui.page('/subpage')
def subpage():
    ui.label('Hello from subpage!')

ui.run_with(fastapi_app, mount_path='/mount', on_air=True)

@fastapi_app.get('/')
def get_root():
    return {'message': 'Hello, FastAPI! Browse to /mount to see the NiceGUI app.'}

if __name__ == '__main__':
    uvicorn.run('main:fastapi_app', log_level='info', reload=True, port=8080, host='0.0.0.0')
```

The problem here is that Starlette already prepends the proxy server to the redirect url and hence `if 'Location' in response.headers and response.headers['Location'].startswith('/'):` in `RedirectWithPrefixMiddleware` fails to inject the prefix.

rodja:
I'll close this PR because the rewrite strategy leads to errors (as @falkoschindler showed) and can only be fixed by adding javascript code (not desired to add more complexity). Starlette redirect logic is already enhanced with a two year old middleware in NiceGUI -- but does not jet work with On Air. We should look into that in a separate PR.

Issue #5115, opened 2025-09-07
In NiceGUI 3.0.0rc1, open the same url twice, the terminal reports `Could not register a disconnect handler for callback`


python-and-novella:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui,Event

event = Event()

@ui.page('/')
def page():
    ui.button('Click me', on_click=lambda: event.emit(42))
    event.subscribe(lambda: ui.notify('clicked'))
    event.subscribe(lambda x: ui.notify(f'{x = }'))

ui.run(port=80,reload=False,show=False)
```

### Description

In NiceGUI 3.0.0rc1, open the same url(`http://127.0.0.1/`) twice, the terminal reports `Could not register a disconnect handler for callback`

<img width="1201" height="128" alt="Image" src="https://github.com/user-attachments/assets/b97c3f97-74c5-4d6c-b94e-9d3d8e7af82b" />

### NiceGUI Version

3.0.0rc1

### Python Version

Python 3.12.10

### Browser

Other

### Operating System

Windows

python-and-novella:
It is a brower problem, paste 127.0.0.1  in url bar, the browser will visit it once quietly(I don't know reason, it's not a normal connection).
The problem dispeard whem I visit 127.0.0.1 in chrome.
All right, my browser is weird.

python-and-novella:
I insert `print(client.id)` in `event.py` line 95, and I tested the MRE:

![Image](https://github.com/user-attachments/assets/108d7580-ceed-4618-9cee-4ef58497be3f)

rodja:
Yes, I can reproduce the issue and created a draft PR at #5116.

PR #2272, opened 2024-01-01
Keep updates and messages in outbox until client is connected

falkoschindler:
This PR is trying to delay updates and messages in case the corresponding client is not yet connected. This will solve problems like this:

```py
@ui.page('/')
def page():
    label = ui.label('Hello world!')

    async def update():
        label.text = 'Hello again!'  # problem 1
        with label:
            ui.run_javascript('alert("Hello from JavaScript!")')  # problem 2
    background_tasks.create(update())
```

Because element updates and other messages are kept in separate queues, there are basically two distinct problems. Messages are already solved, updates are still work in progress.

PR #5116, opened 2025-09-08
Support client connection when Chrome performs speculative loading

rodja:
### Motivation

As @python-and-novella discovered in #5115, Chrome pre-loads content when typing an url. This raises `TimeoutError` messages in different kind of scenarios. See this reproduction:

```py
import asyncio

from nicegui import Event, ui

async def root():
    ui.link('event', '/event')
    ui.link('js', '/js')
    ui.link('connected', '/connected')
    ui.link('longbuild', '/longbuild')
    ui.link('timer', '/timer')

@ui.page('/event')
def event():
    Event().subscribe(print)

@ui.page('/js')
async def js():
    await ui.run_javascript('console.log("hello")')

@ui.page('/connected')
async def connected():

@ui.page('/longbuild')
async def longbuild():
    await asyncio.sleep(4)

@ui.page('/timer')
def timer():
    ui.timer(1, lambda: print('timer'), once=True)

ui.run(root, port=8990)  # NOTE: this syntax is new in 3.0
```

If you simply type in the URL of one of the paths without hitting enter, you'll see a raised error in the console. You may have to switch the port because Chrome has internal heuristics when to do the prefetch. You can also provoke these errors by configuring speculation rules (like in https://github.com/zauberzeug/nicegui/pull/5116#issuecomment-3277456754) or by accessing the urls with curl (which does not establish a websocket connection, obviously).

### Implementation

In `page.py` I discovered that the try/except around awaiting async page to finish never worked before (because `asyncio.wait` does not raise). This is now fixed by the the new `.cancel()` calls on the tasks.

The core contribution of this PR is the discovery that Chrome holds back websocket connections in speculation requests (prerender, prefetch). For prefetch, the browser does not even execute JavaScript. In that case, we can simply wait for navigation to happen until client get's purged after 60 seconds. For prerender we can use socketio polling instead of websocket to establish a connection in the background.
Also this PR makes `run_javascript` awaiting `client.connected` internally to support prefetch (where we can not rely on outbox delivery in time).

Because missing websocket connection can be seen as "normal behavior" (curl, crawlers, ...), this PR also reduces the noise by converting warnings in the console to debug messages.

### Progress

    - [x] investigate and decide about `ui.run_javascript` shows error if client is not connected after a while (as it is [stated in the docstring](https://github.com/zauberzeug/nicegui/blob/d7e21d6ae7aec022851b99551b20a1fe9240d0eb/nicegui/client.py#L210) but worked around in #2272)
    - [x] find solution for preload (which does not execute javascript)
    - [x] is the new `ClientTimeoutError` raised and handled in all scenarios?
    - [x] keep 1 second timeouts for JavaScript calls (use 3.0 timeout for implicitly awaiting connection)
    - [x] use same 3.0 connection timeout for event registration

### 3.0 migration guide

- `run_javascript` now internally awaits `client.connected`. This possible delay is not covered by the `timeout` parameter anymore
- the timeout of `client.connected` might be extended if browser performs a speculative prefetch

### DIFF
--- a/nicegui/api_router.py
+++ b/nicegui/api_router.py
@@ -14,7 +14,7 @@ class APIRouter(fastapi.APIRouter):
              viewport: Optional[str] = None,
              favicon: Optional[Union[str, Path]] = None,
              dark: Optional[bool] = ...,  # type: ignore
-             response_timeout: float = 20.0,
+             response_timeout: float = 3.0,
              **kwargs,
              ) -> Callable:
         """Page
@@ -28,7 +28,7 @@ class APIRouter(fastapi.APIRouter):
         :param viewport: optional viewport meta tag content
         :param favicon: optional relative filepath or absolute URL to a favicon (default: `None`, NiceGUI icon will be used)
         :param dark: whether to use Quasar's dark mode (defaults to `dark` argument of `run` command)
-        :param response_timeout: maximum time for the decorated function to build the page (default: 20.0)
+        :param response_timeout: maximum time for the decorated function to build the page (default: 3.0)
         :param kwargs: additional keyword arguments passed to FastAPI's @app.get method
         """
         return ui_page(
--- a/nicegui/api_router.py
+++ b/nicegui/api_router.py
@@ -14,7 +14,7 @@ class APIRouter(fastapi.APIRouter):
              viewport: Optional[str] = None,
              favicon: Optional[Union[str, Path]] = None,
              dark: Optional[bool] = ...,  # type: ignore
-             response_timeout: float = 3.0,
+             response_timeout: float = 20.0,
              **kwargs,
              ) -> Callable:
         """Page
@@ -28,7 +28,7 @@ class APIRouter(fastapi.APIRouter):
         :param viewport: optional viewport meta tag content
         :param favicon: optional relative filepath or absolute URL to a favicon (default: `None`, NiceGUI icon will be used)
         :param dark: whether to use Quasar's dark mode (defaults to `dark` argument of `run` command)
-        :param response_timeout: maximum time for the decorated function to build the page (default: 3.0)
+        :param response_timeout: maximum time for the decorated function to build the page (default: 20.0)
         :param kwargs: additional keyword arguments passed to FastAPI's @app.get method
         """
         return ui_page(
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -251,7 +251,7 @@ class Client:
         Otherwise, the JavaScript code is executed without waiting for a response.
 
         Obviously the javascript code is only executed after the client is connected.
-        Internally, `await ui.context.client.connected()` is called before the JavaScript code is executed.
+        Internally, `await ui.context.client.connected(timeout=timeout)` is called before the JavaScript code is executed.
 
         :param code: JavaScript code to run
         :param timeout: timeout in seconds (default: `3.0`)
@@ -413,6 +413,7 @@ class Client:
                 if not client.has_socket_connection and client.created <= time.time() - client_age_threshold
             ]
             for client in stale_clients:
+                log.debug(f'Pruning stale client {client.id}')
                 client.delete()
 
         except Exception:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -252,6 +252,7 @@ class Client:
 
         Obviously the javascript code is only executed after the client is connected.
         Internally, `await ui.context.client.connected(timeout=timeout)` is called before the JavaScript code is executed.
+        This might delay the execution of the JavaScript code and is not covered by the `timeout` parameter.
 
         :param code: JavaScript code to run
         :param timeout: timeout in seconds (default: `3.0`)
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -229,11 +229,13 @@ class Client:
         if not self.has_socket_connection:
             self._waiting_for_connection.set()
             self._is_connected.clear()
-            timeout = max(timeout, self.page.response_timeout) if self.is_speculation_prefetch else timeout
-            try:
-                await asyncio.wait_for(self._is_connected.wait(), timeout=timeout)
-            except asyncio.TimeoutError as e:
-                raise ClientConnectionTimeout(self) from e
+            if self.is_speculation_prefetch:
+                await self._is_connected.wait()  # NOTE: wait for connection without timeout until client pruning takes care of cancellation
+            else:
+                try:
+                    await asyncio.wait_for(self._is_connected.wait(), timeout=timeout)
+                except asyncio.TimeoutError as e:
+                    raise ClientConnectionTimeout(self) from e
 
     async def disconnected(self) -> None:
         """Block execution until the client disconnects."""
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -63,8 +63,9 @@ class Client:
         self.elements: dict[int, Element] = {}
         self.next_element_id: int = 0
         self._waiting_for_connection: asyncio.Event = asyncio.Event()
-        self.is_waiting_for_connection: bool = False
-        self.is_waiting_for_disconnect: bool = False
+        self._waiting_for_disconnect: asyncio.Event = asyncio.Event()
+        self._is_connected: asyncio.Event = asyncio.Event()
+        self._deleted_event: asyncio.Event = asyncio.Event()
         self.environ: dict[str, Any] | None = None
         self.on_air = False
         self._num_connections: defaultdict[str, int] = defaultdict(int)
@@ -116,6 +117,16 @@ class Client:
     def has_socket_connection(self) -> bool:
         """Whether the client is connected."""
         return self.tab_id is not None
+
+    @property
+    def is_waiting_for_connection(self) -> bool:
+        """Whether the client is currently waiting for a connection."""
+        return self._waiting_for_connection.is_set()
+
+    @property
+    def is_waiting_for_disconnect(self) -> bool:
+        """Whether the client is currently waiting for a disconnect."""
+        return self._waiting_for_disconnect.is_set()
 
     @property
     def head_html(self) -> str:
@@ -184,25 +195,21 @@ class Client:
         """Return the title of the page."""
         return self.page.resolve_title() if self.title is None else self.title
 
-    async def connected(self, timeout: float = 3.0, check_interval: float = 0.1) -> None:
+    async def connected(self, timeout: float = 3.0) -> None:
         """Block execution until the client is connected."""
-        self.is_waiting_for_connection = True
-        self._waiting_for_connection.set()
-        deadline = time.time() + timeout
-        while not self.has_socket_connection:
-            if time.time() > deadline:
-                raise TimeoutError(f'No connection after {timeout} seconds')
-            await asyncio.sleep(check_interval)
-        self.is_waiting_for_connection = False
+        if not self.has_socket_connection:
+            self._waiting_for_connection.set()
+            self._is_connected.clear()
+            await asyncio.wait_for(self._is_connected.wait(), timeout=timeout)
 
-    async def disconnected(self, check_interval: float = 0.1) -> None:
+    async def disconnected(self) -> None:
         """Block execution until the client disconnects."""
         if not self.has_socket_connection:
             await self.connected()
-        self.is_waiting_for_disconnect = True
-        while self.id in self.instances:
-            await asyncio.sleep(check_interval)
-        self.is_waiting_for_disconnect = False
+        if self.id in self.instances:
+            self._waiting_for_disconnect.set()
+            self._deleted_event.clear()
+            await self._deleted_event.wait()
 
     def run_javascript(self, code: str, *, timeout: float = 3.0) -> AwaitableResponse:
         """Execute JavaScript on the client.
@@ -249,6 +256,8 @@ class Client:
 
     def handle_handshake(self, socket_id: str, document_id: str, next_message_id: int | None) -> None:
         """Cancel pending disconnect task and invoke connect handlers."""
+        self._waiting_for_connection.clear()
+        self._is_connected.set()
         self._socket_to_document_id[socket_id] = document_id
         self._cancel_delete_task(document_id)
         self._num_connections[document_id] += 1
@@ -345,6 +354,8 @@ class Client:
         If the global clients dictionary does not contain the client, its elements are still removed and a KeyError is raised.
         Normally this should never happen, but has been observed (see #1826).
         """
+        self._waiting_for_disconnect.clear()
+        self._deleted_event.set()
         self.remove_all_elements()
         self.outbox.stop()
         del Client.instances[self.id]
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -68,10 +68,10 @@ class Client:
 
         self.elements: dict[int, Element] = {}
         self.next_element_id: int = 0
-        self._waiting_for_connection: asyncio.Event = asyncio.Event()
-        self._waiting_for_disconnect: asyncio.Event = asyncio.Event()
-        self._is_connected: asyncio.Event = asyncio.Event()
-        self._deleted_event: asyncio.Event = asyncio.Event()
+        self._waiting_for_connection = asyncio.Event()
+        self._waiting_for_disconnect = asyncio.Event()
+        self._connected = asyncio.Event()
+        self._deleted_event = asyncio.Event()
         self.environ: dict[str, Any] | None = None
         self.on_air = False
         self._num_connections: defaultdict[str, int] = defaultdict(int)
@@ -120,42 +120,9 @@ class Client:
         return self.request.client.host if self.request.client is not None else ''
 
     @property
-    def is_speculation_prerender(self) -> bool:
-        """Whether this request was initiated by a prerender (Speculation Rules)."""
-        return 'prerender' in self._sec_purpose
-
-    @property
-    def is_speculation_prefetch(self) -> bool:
-        """Whether this request was initiated by a prefetch (Speculation Rules).
-
-        We consider it prefetch when the header contains 'prefetch' and does not contain 'prerender'.
-        """
-        sp = self._sec_purpose
-        return 'prefetch' in sp and 'prerender' not in sp
-
-    @property
-    def _sec_purpose(self) -> str:
-        """Return the Sec-Purpose or Purpose header value in lower case."""
-        try:
-            header = self.request.headers.get('Sec-Purpose') or self.request.headers.get('Purpose') or ''
-            return header.lower()
-        except Exception:
-            return ''
-
-    @property
     def has_socket_connection(self) -> bool:
         """Whether the client is connected."""
         return self.tab_id is not None
-
-    @property
-    def is_waiting_for_connection(self) -> bool:
-        """Whether the client is currently waiting for a connection."""
-        return self._waiting_for_connection.is_set()
-
-    @property
-    def is_waiting_for_disconnect(self) -> bool:
-        """Whether the client is currently waiting for a disconnect."""
-        return self._waiting_for_disconnect.is_set()
 
     @property
     def head_html(self) -> str:
@@ -226,16 +193,16 @@ class Client:
 
     async def connected(self, timeout: float = 3.0) -> None:
         """Block execution until the client is connected."""
-        if not self.has_socket_connection:
-            self._waiting_for_connection.set()
-            self._is_connected.clear()
-            if self.is_speculation_prefetch:
-                await self._is_connected.wait()  # NOTE: wait for connection without timeout until client pruning takes care of cancellation
-            else:
-                try:
-                    await asyncio.wait_for(self._is_connected.wait(), timeout=timeout)
-                except asyncio.TimeoutError as e:
-                    raise ClientConnectionTimeout(self) from e
+        if self.has_socket_connection:
+            return
+        self._waiting_for_connection.set()
+        self._connected.clear()
+        purpose = (self.request.headers.get('Sec-Purpose') or self.request.headers.get('Purpose') or '').lower()
+        is_prefetch = 'prefetch' in purpose and 'prerender' not in purpose
+        try:
+            await asyncio.wait_for(self._connected.wait(), timeout=None if is_prefetch else timeout)
+        except asyncio.TimeoutError as e:
+            raise ClientConnectionTimeout(self) from e
 
     async def disconnected(self) -> None:
         """Block execution until the client disconnects."""
@@ -246,18 +213,18 @@ class Client:
             self._deleted_event.clear()
             await self._deleted_event.wait()
 
-    def run_javascript(self, code: str, *, timeout: float = 3.0) -> AwaitableResponse:
+    def run_javascript(self, code: str, *, timeout: float = 1.0) -> AwaitableResponse:
         """Execute JavaScript on the client.
 
         If the function is awaited, the result of the JavaScript code is returned.
         Otherwise, the JavaScript code is executed without waiting for a response.
 
-        Obviously the javascript code is only executed after the client is connected.
-        Internally, `await ui.context.client.connected(timeout=timeout)` is called before the JavaScript code is executed.
-        This might delay the execution of the JavaScript code and is not covered by the `timeout` parameter.
+        Obviously the JavaScript code is only executed after the client is connected.
+        Internally, ``await client.connected()`` is called before the JavaScript code is executed (*since version 3.0.0*).
+        This might delay the execution of the JavaScript code and is not covered by the ``timeout`` parameter.
 
         :param code: JavaScript code to run
-        :param timeout: timeout in seconds (default: `3.0`)
+        :param timeout: timeout in seconds (default: 1.0)
 
         :return: AwaitableResponse that can be awaited to get the result of the JavaScript code
         """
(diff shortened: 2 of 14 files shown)


rodja:
I'm still undecided what to do about the js-error. On the one hand it is annoying to have this error, on the other hand directly awaiting javascript without awaiting a connection before is not good practice in the first place.

rodja:
In #2272 we used the `Outbox` to collect js commands, which made `ui.run_javascript` possible even before a client is connected. But the default 1 second timeout is way to short if a prefetch is involved.
To reproduce more reliable and not depend on the speculation engine of Chrome, we can force prerender with some `add_head_html`.

```py
from nicegui import ui

def root():
    ui.add_head_html('''<script type="speculationrules">
        {
            "prerender": [
                {"source":"list","urls":["/test"],"eagerness":"eager","tag":"t1"}
            ]
        }
        </script>''')
    ui.link('test', '/test')

@ui.page('/test')
async def test():
    await ui.run_javascript('console.log("hello")')

ui.run(root) # NOTE: this syntax is new in 3.0
```

Simply open the `/` page, wait a sec and the JavaScript Timeout error will appear. If you then click the link which brings you to the `/test` page, a 500 error will be rendered.

rodja:
Oh, by the way: the javascript issue above is also happening in 2.x.

rodja:
Before thinking about how to work around the issue, let's take a step back: why is the client not connected when Chrome tries to prerender in the first place? It seems that the websocket connection is held back. ChatGPT says

> Chrome won’t let you “whitelist” WebSockets during prerender. In a prerendered doc Chrome treats the page as hidden and defers/blocks various side-effectful APIs until activation; you can only detect prerender and adapt. There’s no header, flag, or CSP you can send to enable WS in prerender.

It seems to work fine when the protocol uses polling instead of websocket. I just had to increase the js timeout because polling is slower.

rodja:
I really struggled to find a proper implementation to support preload. Here are some observations about the general behaviour (without this PR):

- no js is executed, just the html
- when page builder awaits javasscript or connection, we return the http response which is then held by chrome in its specualtion result list
- if the navigation happens before page timeout or javascript timeout, the handshake works and everything is fine (prefeched builded page can be used)
- if the navigation happens after page timeout or javascript timeout, the handshake **does not fail** (because the timeout does not remove the client) but does not render the proper content (because of the raised timeout)

For a reproduction use

```py
@ui.page('/')
def root():
    ui.add_head_html('''<script type="speculationrules">
        {
            "prefetch": [
                {"source":"list","urls":["/test"],"eagerness":"eager","tag":"t1"}
            ]
        }
        </script>''')
    ui.link('test', '/test')

@ui.page('/test', response_timeout=4)
async def test():
    await ui.run_javascript('console.log("hello")', timeout=2)
    ui.label('should be shown on page')
``` 

you'll see the label only if you navigate to `/test` before the two second timeout.

The solution needs multiple changes:

- awaiting the connection must use the page response timeout when in prefetch (because default timeout is too short)
- `run_javascript` must await the connection to benefit from the longer timeout
- a new `ClientTimeoutError` must be raised to signal the elapsed connection attempt
- page logic needs to catch this error and delete the client, so a connection attempt produces a failed handshake which in turn triggers a reload in the browser

rodja:
@evnchn the PR is not yet completely done, but because @falkoschindler is on vacation for another week I would love you to do the review later on.

evnchn:
Interesting PR. I read https://github.com/zauberzeug/nicegui/pull/5116#issuecomment-3289216797 and got hooked. 

**Major edit: Handshake occurs for prerendering since we're adding polling to the transports if we detect prerendering request. No handshake for prefetch though...** 

@rodja Something clicked within my brain related to past PRs and I feel like I must say this. Otherwise this PR is headed to do a change that doesn't improve _prefetch_ loading performance due to **the inherent bottleneck**. 

**Please use mindset of #4552 and #4719 to read this. Good-ol times.** 

---

Consider this series of event, ordered chronologically, which will happen without any `await ui.run_javascript(...)` and `await ui.context.client.connected()` calls, really (so it's the case before or after this PR):

1. Chrome fetches the HTML, but no client connection (no JS execution in prefetch)
2. Client spawned by the HTML fetch is "destined to die lonely" as the `reconnect_timeout` lapses
3. When Chrome uses the prefetched HTML, its `client_id` will be associated with a dead client
4. Handshake fails, page forcibly reloads, and our pre-fetched HTML is useless... **(the inherent bottleneck)**

---

**See, we are at the age-old question of "browser's Client UUID is a non-existent one". If this isn't ACID and BASE nothing is.** 

For full effect of spectulative loading able to speed up the page connection, we need the following:

- Mismatched `client_id` between browser and client should be fine
  - Browser client should be OK to use a stale `client_id` to connect (ACID approach, not recommended)
  - Or, browser client should be "open to the possibility" that the `client_id` is going to be different (BASE approach, recommended)
- Between the old and the new client, what's valuable is the **JSON-list of elements** which we should avoid re-send
  - _(If you recall, both in ACID and BASE we re-send the entire list over, so here is where things diverges!)_
  - Hash the JSON-list of elements
  - If on handshake, browser's hash same as new client's, let the browser connect one way or another
  - If on handshake, browser's hash mismatch:
    - Reload the browser
    - Communicate the differences (hard, not recommended)
- User education shall be done to try your very best to serve identical content (at least for the pre-connection part)
  - Documentation
  - Log warning in the page, if the hash mismatches frequently for a page

As of tradition, we need an awful acronym for such a project regarding browser-server client UUID trickery. I'm thinking Project SALT for Speculate A Load Technique 🤣 

---

**This does not say that this PR is bad or wrong, though (I'll see if I have time for a read but I feel like if Falko's in a vacation we should also take it easy as well).** Just be prepared that the prefetch load is still awful speed even after this PR if we do not do Project SALT....

evnchn:
Right. This PR is a pain to do since the internal speculation details of modern browsers are such a black box. To save your sanity I offer you this: 

```py
@app.middleware('http')
async def log_requests(request: Request, call_next):
    print(f'Request: {request.method} {request.url}')
    response = await call_next(request)
    print(f'Response Code: {response.status_code}')
    return response
```

Self-explanatory. Logs the requests immediately when they come in, then the status code when the response is ready.

nicegui/page.py
```diff
@@ -161,17 +166,19 @@ async def wait_for_result() -> Response | None:
                             return await result
                         except Exception as e:
                             return create_error_page(e, request)
-                task = background_tasks.create(wait_for_result())
+                task = asyncio.create_task(wait_for_result())
```
evnchn: We have agreed to never use bare `asyncio.create_task` in #4551, no? 

To be honest I would even promote adding check against `asyncio.create_task` into pre-commit so that it never appears in our codebase, since it causes `Task was destroyed but it is pending!` if the task is GC-ed while it is pending,

nicegui/client.py
```diff
@@ -184,37 +224,37 @@ def resolve_title(self) -> str:
...
         """Execute JavaScript on the client.
 
-        The client connection must be established before this method is called.
-        You can do this by `await client.connected()` or register a callback with `client.on_connect(...)`.
-
         If the function is awaited, the result of the JavaScript code is returned.
         Otherwise, the JavaScript code is executed without waiting for a response.
 
+        Obviously the javascript code is only executed after the client is connected.
+        Internally, `await ui.context.client.connected(timeout=timeout)` is called before the JavaScript code is executed.
```
evnchn: Not sure on linking the Socket.IO connection timeout with the JS execution timeout. I don't like it for 2 reasons: 

- Just because I need to run a 20-second long JS task doesn't mean I would tolerate a client which takes 20 second just to Socket.IO connect. 
- Setting the timeout, I also have expectations on when the function has to finish (either return value or raise error). If I do `await ui.run_javascript(..., timeout = 10)` I totally expect the function to finish in 10 seconds. Now it may take up to 20 seconds. 

---

Would advocate `def run_javascript(self, code: str, *, timeout: float = 3.0, connect_timeout: float = 3.0)`. Better control and clarity on how long the function will take. 

Side benefit I can also pass `connect_timeout = 0` to revert to old behaviour which may be useful when you know, for sure, that the client has connected and time limit only the JS execution

nicegui/functions/javascript.py
```diff
@@ -2,19 +2,21 @@
...
     This function runs arbitrary JavaScript code on a page that is executed in the browser.
-    The client must be connected before this function is called.
     To access a client-side Vue component or HTML element by ID,
     use the JavaScript functions `getElement()` or `getHtmlElement()` (*added in version 2.9.0*).
 
     If the function is awaited, the result of the JavaScript code is returned.
     Otherwise, the JavaScript code is executed without waiting for a response.
 
+    Obviously the javascript code is only executed after the client is connected.
+    Use `await ui.context.client.connected()` to wait for the client to connect.
```
evnchn: In `client.py` we already by-default `await ui.context.client.connected()`. Let's match the docstring.

evnchn:
Yes. It is indeed handshake failed. Check this out. 

image

evnchn:
Also here's an important section in Devtools: speculation section inside Application tab. Surprised this didn't come up in this discussion. 

image

evnchn (changes requested):
I have some feedback over the timeout configuration of `run_javascript`, use of bare `asyncio.create_task` and docstring alignment between `run_javascript` of `nicegui/functions/javascript.py` and `nicegui/client.py`. Kindly check. Thanks!

nicegui/client.py
evnchn: Also the docstring: **If the result is awaited, then** internally, `await ui.context.client.connected(timeout=timeout)` is called ...

Consider that if we don't await the result then we don't await client connection.

nicegui/client.py
```diff
@@ -41,6 +41,12 @@
 })
 
 
+class ClientConnectionTimeout(Exception):
```
evnchn: For backwards compatibility, `ClientConnectionTimeout` should also be a `TimeoutError`, so that we don't break this possible user code:

```py
try:
except TimeoutError:
    print('It is fine')
```

nicegui/page.py
rodja: Oh wow. Good catch. Indeed, the task may be removed by GC, because we `add_done_callback` only holds a reference inside the task. The problem is this: `background_tasks.create` catches all exceptions -- which is problematic because `check_for_late_return_value` needs to check for `ClientConnectionTimeout`. Maybe we should add an optional parameter  `background_tasks.create(..., catch_exceptions=True)`?

nicegui/api_router.py
```diff
@@ -14,7 +14,7 @@ def page(self,
              viewport: Optional[str] = None,
              favicon: Optional[Union[str, Path]] = None,
              dark: Optional[bool] = ...,  # type: ignore
-             response_timeout: float = 3.0,
+             response_timeout: float = 20.0,
```
evnchn: Wanting to wait 20 seconds in hopes of the user clicking the speculated link should not also mean we do not timeout if the page logic takes like 20 seconds.

nicegui/client.py
rodja: >  If I do await ui.run_javascript(..., timeout = 10) I totally expect the function to finish in 10 seconds. Now it may take up to 20 seconds.

Yes. Your are right.

> Would advocate def run_javascript(self, code: str, *, timeout: float = 3.0, connect_timeout: float = 3.0).

I wanted to avoid jet another parameter which we need to maintain. But I get your point and maybe it's the best solution.

nicegui/client.py
```diff
@@ -184,37 +224,37 @@ def resolve_title(self) -> str:
...
-            if time.time() > deadline:
-                raise TimeoutError(f'No connection after {timeout} seconds')
-            await asyncio.sleep(check_interval)
-        self.is_waiting_for_connection = False
-
-    async def disconnected(self, check_interval: float = 0.1) -> None:
+        if not self.has_socket_connection:
+            self._waiting_for_connection.set()
+            self._is_connected.clear()
+            timeout = max(timeout, self.page.response_timeout) if self.is_speculation_prefetch else timeout
```
evnchn: Currently we wait 20 seconds in hopes of the user clicking the speculated link by bending the implementation of `await ui.context.client.connected()`. Similar to `await ui.run_javascript(...)` it also breaks user expectations:

```py
@ui.page('/')
async def use_a_token_which_valid_for_5_seconds():
    token = get_token()
    await ui.context.client.connected(timeout=5)
    begin_transaction(token)
```

After this PR, `begin_transaction` can easily take a token which is stale for more than 5 seconds, which is not obvious in the code. I don't think this is a good idea.

evnchn (commented):
I have looked at this long and hard. Basically currently the prefetch handling goes as follows: 

- Stretch `response_timeout` to 20 seconds
- Bend `await ui.context.client.connected()` to await longer than passed-in timeout in case of prefetch. 

_As a reminder, `response_timeout` is the time from async page function execution to `await ui.context.client.connected()` or full-completion of the async page function, not applicable to sync page function._ 

Besides the possible issue stated in the code review, namely (1) letting long-running page functions slip by, (2) breaking of user expectation of the `timeout` parameter of `await ui.context.client.connected()`, there's another issue that **this trick doesn't work if the page function is sync, or there's neither `await ui.run_javascript` nor `await ui.context.client.connected()`**, where both cases `response_timeout` will finish counting and `reconnect_timeout` takes over. 

---

I think that Project SALT is the best way forward, as it can enable the browser to connect literally any time (it can be many minutes and even hours from the prefetch to the clicking of the link and it'd still work, assuming the browser didn't already purge the prefetch due to the long delay). 

Still, if that's not attainable within a short time, I would recommend: 

- Normally after counting the `response_timeout` (3->20?) we count the `reconnect_timeout` (10) directly. 
- Suppose we do not change `response_timeout`, keep it at 3. 
- Then, can we, for the first time waiting for connection, have a longer `reconnect_timeout`? How about `first_connect_timeout`?

(Merge the changes of including `await ui.context.client.connected()` in `await ui.run_javascript` and the `asyncio.Event`-based client connection-disconnection logic first, if the above steps take long time)

nicegui/page.py
evnchn: Extra parameter sounds good. After all, we don't need to catch exceptions all the time, but still need the anti-task-GC.

nicegui/client.py
evnchn: Well if we call `connected` inside `run_javascript`, and we don't have sensible defaults for the parameter for `connected`, I don't see how we can get away with not including that parameter in the parameters of `run_javascript`

I did think about `def run_javascript(self, code: str, *, timeout: float = 3.0, await_connection: bool = True)`. Although it enables the configuration if someone did `await_connection=False` then call `await ui.context.client.connected(timeout=7.7)` themselves, it still wouldn't be obvious that the default call can take up to +3 seconds on top of the `timeout` though, which is what we came for.

rodja: I tried introducing the `connect_timeout` but we would need to introduce it all over the place in `Element` we have `run_method` and `get_computed_prop`, `aggrid` has it's own `get_client_data`, ...
Sometimes it's [slightly better to be simple than correct](https://en.wikipedia.org/wiki/Worse_is_better#:~:text=It%20is%20slightly%20better%20to%20be%20simple%20than%20correct.). Hence I advocate of keeping it as it is and accept that the timeout might be longer than expected.

rodja:
Thanks for all the feedback @evnchn.

> For full effect of spectulative loading able to speed up the page connection, we need [...] Project SALT [based on the ideas of #4552 and #4719]

That might be true. But this PR mainly focus on not breaking pages when Chrome does speculative loading. Everything else can come after that. For example in https://github.com/zauberzeug/nicegui/pull/5116#issuecomment-3289216797 I described a scenario which shows a partially rendered page without this PR. Thats why I still think this is a first good (and mandatory) step: 1. make prerender connect via polling and 2. make prefetch a bit more efficient by allowing a connection in 20.0 seconds. Even after that time prefetch might have gained something by populating the cache with assets etc. Good enough in my opinion. For better performance we provide `ui.sub_pages`.

The three issues you identified with this PR are indeed worth considering:

> (1) letting long-running page functions slip by,

As written in the description

> In page.py I discovered that the try/except around awaiting async page to finish never worked before (because asyncio.wait does not raise).

So, at least with this PR it does not slip indefinitely but raises after 20 seconds. I consider this a huge improvement without breaking existing user code too much.

> (2) breaking of user expectation of the timeout parameter of await ui.context.client.connected(),

As written in https://github.com/zauberzeug/nicegui/pull/5116#discussion_r2347491974 I now think that it's ok to not strictly follow the timeout parameter. Maybe we can simply write "at least x seconds of timeout" in the docstring...

> there's another issue that this trick doesn't work if the page function is sync, or there's neither await ui.run_javascript nor await ui.context.client.connected()

Why would prefetch fail? In a sync function the client stays connectable until it is purged after 60.0 seconds.

nicegui/client.py
evnchn: Fair point. Didn't realize how much `run_javascript` is used internally in the codebase. We can document this change from 2.x to 3.0 then.

nicegui/api_router.py
rodja: But in 2.x there is no timeout at all for this. There is a try/except block 

https://github.com/zauberzeug/nicegui/blob/1ed7e72001079380f123a45cd84a2e8344a8c244/nicegui/page.py#L160C2-L166C107

but `asyncio.wait` does not throw an exception.

nicegui/client.py
rodja: Yes it can. But as above with the hidden delay in `run_javascript`, [this simplification is slightly better than being correct](https://en.wikipedia.org/wiki/Worse_is_better#:~:text=It%20is%20slightly%20better%20to%20be%20simple%20than%20correct.).

nicegui/page.py
rodja: Ok, I just pushed 9e15343ae33139d27604b7706c8c29c945d4f2bc

evnchn:
To rethink a bit: what even is the purpose of having `await ui.context.client.connected()` of 3 seconds, when the `response_timeout` defaults to 20 seconds? 

Your code, which needs the client connection, I presume, either errors out due to the timeout, or sits lonely for 17 seconds without one. 

🤔

As such, I would propose in 3.0 as a breaking change, to default the timeout of `await ui.context.client.connected()` to be that page's `response_timeout`. 

~~If we are more brave we can drop the parameter directly, even.~~ No we need control of the timeout since `ui.run_javascript` uses it and we're not down for waiting 20 seconds for a JS call at all. 

This would drop the duality of function of `await ui.context.client.connected()` where it would, in the prefetch case, wait longer than the timeout.

nicegui/client.py
evnchn: Since I am proposing that `await ui.context.client.connected()` will default to page response timeout, then we should always do `await ui.context.client.connected(timeout=timeout)` in `await ui.run_javascript`. 

We will then document that users are expected to allocate **twice** the time stated in `timeout`, counting once for waiting the client connection, another time for the JS execution.

rodja:
@evnchn The response timeout was meant to ensure that a page builder produces a response in a certain time. But as soon as the builder sees `await client.connection()`, it produces a response (and hence the guarantee is fulfilled). Waiting 3 seconds for socketio handshake is more than enough in normal circumstances. If it takes longer network is so bad that , NiceGUI won't work at all. Except Chromes prefetch mechanism breaks this assumption.

Maybe we should introduce a `Client` class variable `PREFETCH_TIMEOUT=60.0` or similar which can be used in `client.connected` to extend the wait for the prefetch use case. Should be much better that misusing the `response_timeout` from page.

evnchn:
> But as soon as the builder sees `await client.connection()`, it produces a response (and hence the guarantee is fulfilled).

Oh yes `response_timeout` is from initial execution to the BEGINNING of `await client.connection()` if there is one. 

_If that's the case the analysis above of mine may be slightly inaccurate. I'll look at it again._

---

But if that's the case then I should be able to keep the old `response_timeout` of 3 and use `await client.connection()` with a large timeout when detected prefetch mode (this PR's behavior). This would still allow late Socket.IO handshake. 

Then we don't need to change the default value of `response_timeout`, then. 

If that's the case, would agree that setting the timeout of `await client.connection()` to `response_timeout` is an abuse under current understanding. 

Maybe we can default the timeout to 60 seconds if detected prefetch? (The time before which the client is pruned). If we let that 60 be configurable may be a bonus. 

---

Y'know I originally thought `response_timeout` were to be from initial execution to the END of `await client.connection()` if there is one. This is why I suggested what I want to suggest. Sorry for that. 

But adjusting the definition of `response_timeout` can make sense in a nice way: 
- Page is at a stage where "it requires the client to be fullt-connedted to proceed" when call `await client.connection()` 
- If the `response_timeout` is 5 second, the page has 5 seconds all-inclusive of computing the page AND network latency for the client to connect. 
- Even if we send out a HTML response at the 1-second mark we don't call it done for the counting of `response_timeout`

rodja:
Thanks @evnchn. I had assumed that the `response_timeout` is not working on main and 3.0 because there is no exception raised. But the timeout works -- you just don't get any info. So it's better to keep it at 3 seconds.

Also your idea of binding `await client.connected` to the pruning of the client is great. Instead of simply setting 60 seconds, in 2947fab I modified `connected` to simply have no timeout at all. When client gets pruned the task is canceled anyway. What do you think?

evnchn:
I hope Falko will be happy when he returns from vacation and sees this ❤️

nicegui/client.py
rodja: Resolved with https://github.com/zauberzeug/nicegui/pull/5116#issuecomment-3301268948

nicegui/api_router.py
rodja: Resolved with https://github.com/zauberzeug/nicegui/pull/5116#issuecomment-3301268948

nicegui/static/nicegui.js
```diff
@@ -269,6 +269,7 @@ function download(src, filename, mediaType, prefix) {
 }
 
 function ack() {
+  if (!window.socket || !window.did_handshake) return;
```
falkoschindler, line 272: I don't see how this change has anything to do with the transport change in 799f7159c8c3cb1e55bbfda41fa5ab6148881468. Do we really need the extra condition?

nicegui/elements/timer.py
```diff
@@ -17,13 +17,12 @@ async def _can_start(self) -> bool:
         See https://github.com/zauberzeug/nicegui/issues/206 for details.
         Returns True if the client is connected, False if the client is not connected and the timer should be cancelled.
         """
-        # ignore served pages which do not reconnect to backend (e.g. monitoring requests, scrapers etc.)
-        TIMEOUT = 60.0
         try:
-            await self.client.connected(timeout=TIMEOUT)
+            await self.client.connected()
```
falkoschindler, line 21: I can't remember why we chose a connection timeout of 60 seconds before stopping timers. But is it really safe to reduce it down to 3 seconds? Are there any situations where we might need more time? And is it essential for this PR or mainly a matter of consistency?

evnchn:
Hmm. @falkoschindler JS timeout was increased from 1 to 3 in https://github.com/zauberzeug/nicegui/pull/5116/commits/c37384ec5a591fd243a4bf83cc412f10cfd0a6d7 "to support polling websocket, which is default when Chrome prerenders", so I am not sure whether to revert that. 

For the 60 second timeout it was introduced in https://github.com/zauberzeug/nicegui/commit/e4d50659747a881ab824efe725b6195b500dd030 directly on main but absolutely can't find the reason why. Not even in https://github.com/zauberzeug/nicegui/releases/tag/v1.1.9, but I think maybe we should unify the timeout since it's kinda awkward if `await ui.context.client.connected()` fails but `ui.timer` runs, don't you think? 

The rest you will need to check with @rodja on the details...

nicegui/static/nicegui.js
falkoschindler: Ok, technically mounting the app doesn't have to finish before `ack()` is called for the first time (after 3 seconds), so this extra condition is kind of reasonable. Just unrelated to the PR.

nicegui/elements/timer.py
falkoschindler: Oh, this change breaks the following example when simulating a "Slow 4G" connection:
```py
@ui.page('/')
def page():
    ui.label('Page is ready.')
    ui.timer(1.0, lambda: ui.notify('Tic toc'))
```
On main the timer patiently waits for the client to connect (independent of the default timeout of 3 seconds for `client.connected`). But with this change the timer is canceled long before the page is ready.

evnchn: Hmm. Since we're "extending the timeout of `client.connected` to 60 seconds in a detected prefetch" anyways, what do you think if we **globally** made the timeout to be 60 seconds to unify everything? 

Then there would be no ambiguity. 60 seconds is what you get no matter where the request took place.

falkoschindler:
@evnchn Good point. The question is what we would expect to happen on a slow connection.

Currently the following page fails to load on "Slow 4G" ("JavaScript did not respond within 1.0 s"):
```py
@ui.page('/')
async def page():
    date = await ui.run_javascript('Date()')
    ui.label(f'Ready. {date}')
```

But even worse: This PR causes an endless reload-loop for this example on "Slow 4G" without cache. 😕

evnchn:
Works with https://github.com/zauberzeug/nicegui/pull/5116/commits/adff733902a49ec303a58b8559d29fd180c6ad7e

Yeah let's just await the client indefinitely since we always have the 60-seconds prune client to catch our back.

evnchn:
Oh I wish we just made, permanently, without any parameter, that `client.connected()` have no timeout. This way we get rid of `ClientConnectionTimeout` exception entirely. 

Might as well go for it. since 3.0 is for breaking changes, right?

falkoschindler:
Sorry, @evnchn, I've noticed your message just after merging. But there might be use cases for a connection timeout parameter, so we decided to keep it. We can think about removing it in 4.0, but somehow we need to feature-freeze 3.0.

evnchn:
See you next year for https://github.com/zauberzeug/nicegui/pull/5172

PR #5129, opened 2025-09-11
Re-initialize `process_pool` if it was broken by a bad `run.cpu_bound`

### LINKED ISSUE #4980: Broken process pool on run.cpu_bound if process terminates unexpectedly
### First Check

### Example Code

```python
from nicegui import ui, run, app
import os

def crash_process():
    # Cause the process to exit with a non-zero exit status
    os._exit(1)

def fine_process():
    print('Hey this should be printed')

async def on_long_operations():
    try:
        print('First task')
        await run.cpu_bound(crash_process)
    except Exception as e:
        print(f'First task error: {e}')

    try:
        print('Second task')
        await run.cpu_bound(fine_process)
    except Exception as e:
        print(f'Second task error: {e}')

    try:
        print('Third task')
        await run.cpu_bound(crash_process)
    except Exception as e:
        print(f'Third task error: {e}')

ui.button('Do the Job', on_click=on_long_operations)


ui.run()
```

### Description

Just like in issue #2774 , the same problem arises when using the run.cpu_bound function. If the called function terminates unexpectedly, the process pool can no longer be used. If you run the code you get the following result:

```
First task
First task error: A process in the process pool was terminated abruptly while the future was running or pending.
Second task
Second task error: A child process terminated abruptly, the process pool is not usable anymore
Third task
Third task error: A child process terminated abruptly, the process pool is not usable anymore

```
The only solution is to restart the whole application, which is bad for other users of that application. 

### NiceGUI Version

2.21.1

### Python Version

Python 3.10.12

### Browser

Chrome

### Operating System

Linux

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #4980, where we see if the process pool is broken in a manner of process termination, there is no way for the user to recover it without resorting to directly writing to `run.process_pool = ProcessPoolExecutor()`

### Implementation

This implementation checks if the process pool is broken after an error has been raise by asking it to run an empty lambda. If that fails then the process pool is broken and we re-initialize it. Nevertheless the error will still be raise. 

Effect is, no matter how mangled your calls to `run.cpu_bound`, you can never wreck it and affect the next normal call. 

### Progress

 - Not sure if I should add some. 

### DIFF
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -74,12 +74,20 @@ async def test_run_cpu_bound_function_which_raises_problematic_exception(user: U
     await user.open('/')
 
 
+def bad_function() -> None:
+    os._exit(1)  # pylint: disable=protected-access
+
+
+def good_function() -> bool:
+    return True
+
+
 async def test_run_cpu_bound_survive_bad_function(user: User):
     @ui.page('/')
     async def index():
         with pytest.raises(BrokenProcessPool):
-            await run.cpu_bound(lambda: os._exit(1))  # bad function kills the process pool
-        assert isinstance(await run.cpu_bound(time.time), float)  # good function returns without error
+            await run.cpu_bound(bad_function)
+        assert await run.cpu_bound(good_function)
         ui.label('excellent')
 
     await user.open('/')


nicegui/run.py
```diff
@@ -76,10 +77,20 @@ async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs)
...
-    return await _run(process_pool, safe_callback, callback, *args, **kwargs)
+    try:
+        return await _run(process_pool, safe_callback, callback, *args, **kwargs)
+    except BrokenProcessPool as e:
+        try:
+            await _run(process_pool, safe_callback, lambda: None)
+        except BrokenProcessPool:
+            process_pool = ProcessPoolExecutor()
+        finally:
+            raise e from None
```
rodja: Is that correct? Should it not be something like `raise RuntimeError from e` or a bare raise?

evnchn: Did `raise e from None` in hopes that the new traceback is same as old one, keeping things simple and avoid confusion (e.g. "Why did NiceGUI re-raise the error when it already try-excepted it?")

Retrospect, a bare raise would help make things more obvious, at cost of longer traceback. 

I'm fine either way. What do you think? I'm not at home for the weekend so you can push to it directly.

evnchn: Still `During handling of the above exception, another exception occurred` _sounds_ like NiceGUI messed up, which is not the intent. 

If only we can change that message...

evnchn:
Original traceback:

```
  File "/workspaces/nicegui/dist/test_bpe.py", line 18, in on_long_operations
    await run.cpu_bound(crash_process)
  File "/workspaces/nicegui/nicegui/run.py", line 86, in cpu_bound
    return await _run(process_pool, safe_callback, callback, *args, **kwargs)
  File "/workspaces/nicegui/nicegui/run.py", line 63, in _run
    return await loop.run_in_executor(executor, partial(callback, *args, **kwargs))
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

Traceback with repair, `raise e from None`

```
  File "/workspaces/nicegui/dist/test_bpe.py", line 18, in on_long_operations
    await run.cpu_bound(crash_process)
  File "/workspaces/nicegui/nicegui/run.py", line 93, in cpu_bound
    raise e from None
  File "/workspaces/nicegui/nicegui/run.py", line 86, in cpu_bound
    return await _run(process_pool, safe_callback, callback, *args, **kwargs)
  File "/workspaces/nicegui/nicegui/run.py", line 63, in _run
    return await loop.run_in_executor(executor, partial(callback, *args, **kwargs))
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

Traceback with repair, `raise e`

```
  File "/workspaces/nicegui/dist/test_bpe.py", line 18, in on_long_operations
    await run.cpu_bound(crash_process)
  File "/workspaces/nicegui/nicegui/run.py", line 93, in cpu_bound
    raise e
  File "/workspaces/nicegui/nicegui/run.py", line 86, in cpu_bound
    return await _run(process_pool, safe_callback, callback, *args, **kwargs)
  File "/workspaces/nicegui/nicegui/run.py", line 63, in _run
    return await loop.run_in_executor(executor, partial(callback, *args, **kwargs))
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

Huh, I may have some misunderstanding towards the syntax. Why does `During handling of the above exception, another exception occurred` not show up? 

Well then, let's do `raise e`.

rodja:
I think we should put this in 3.0... we plan no new feature release 2.25.

evnchn:
Actually true. Imagine 3.0 is releasing, and (touchwood) something went wrong. The more features we have in 3.0 the tougher it will be for debug.

falkoschindler:
We just thought about merging this PR into 3.0 nonetheless to get it out of our feet. But tests are failing for Python 3.12 and 3.13. Unfortunately I can't reproduce it locally.

evnchn:
# 🤡 

image

falkoschindler:
Interesting! But a change of `cpu_count` in Python 3.13 wouldn't explain 3.12 _and_ 3.13 to fail...

Now only 3.13 fails. Maybe they both (or all) are flaky? 🤔

evnchn:
I'm thinking more along the lines of "since GitHub Actions is cloud-based, that they may be using a more cut-down environment such that there is no concept of CPU cores"

evnchn:
Continuing on with the theme of "potentially we're working in a Cloud environment" I decided to go back to using `os._exit(1)` since that calls the C `_exit()` directly. See if it works!

falkoschindler:
Ok, with the original `bad_function` and `good_function` tests are green again. Let's leave it this way and merge into 3.0.

PR #5135, opened 2025-09-13
Drop `check_interval` for `Client.disconnected()`

evnchn:
### Motivation

We basically had "war on polling based solution" since #4466 (the good-ol days). This brings it to `Client.disconnected()`. Now it doesn't poll. 

### Implementation

In this implementation, I use `on_disconnect`. Not sure if it is necessarily the same as `while self.id in self.instances:`, though. Tests pass, so that's one thing. 

Will it cause trouble, if many calls to `Client.disconnected()`, and we add a lot of lambdas to `Client.disconnect_handlers`?

### Progress

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -195,13 +195,14 @@ class Client:
             await asyncio.sleep(check_interval)
         self.is_waiting_for_connection = False
 
-    async def disconnected(self, check_interval: float = 0.1) -> None:
+    async def disconnected(self) -> None:
         """Block execution until the client disconnects."""
         if not self.has_socket_connection:
             await self.connected()
         self.is_waiting_for_disconnect = True
-        while self.id in self.instances:
-            await asyncio.sleep(check_interval)
+        disconnect_event = asyncio.Event()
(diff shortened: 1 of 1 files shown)


evnchn:
How about this for `Client.connected()`?

```py
    async def connected(self, timeout: float = 3.0) -> None:
        """Block execution until the client is connected."""
        self.is_waiting_for_connection = True
        self._waiting_for_connection.set()
        connect_event = asyncio.Event()
        self.on_connect(lambda _: connect_event.set())
        try:
            await asyncio.wait_for(connect_event.wait(), timeout=timeout)
        except asyncio.TimeoutError:
            raise TimeoutError(f'No connection after {timeout} seconds')
        self.is_waiting_for_connection = False
```

evnchn:
Also in progress in https://github.com/zauberzeug/nicegui/pull/5116

Maybe we can focus on that instead.

rodja:
@evnchn Yes. In #5116 I had to rework the `connected` and `disconnected`. The last few days I really struggled to get a grip on the Chrome speculation requests but did not push my changes 😞
Your PR here reminded me about that.

PR #5137, opened 2025-09-13
Add Anywidget Front-End Module spec support with `ui.anywidget()`

azjps:
### Motivation

Refer to #5096 for details: add support for anywidgets to be rendered in NiceGUI, such as the 70+ examples from the anywidgets gallery: https://try.anywidget.dev/. For example, this solves #657 (adding `altair` support); for convenience, this PR exposes a `ui.altair(chart: altair.Chart)` although its a one-line wrapper around `ui.anywidget()`.

Roughly speaking the flow is:
- python->JS: Use `traitlets.HasTraits.observe()` to listen to changes to anywidget's traitlets; on observing changes, call `nicegui.Element.run_method("update_trait", <trait>)` to invoke any callbacks on JS side registered by `model.on("change:<trait>")` in the anywidget's`esm`
- JS->python: `model.save_changes()` `$emit`s the state of `traits` to `ValueElement._handle_value_change`, where any changed traits will be set on the anywidget 

It also might simplify feature requests for other pure-JS libraries, for example it took me a couple of minutes to vibe code a TradingView LightWeight chart widget (#813) that I iterated on in Jupyter and then copy-pasted into NiceGUI, without needing to add any code to NiceGUI.

### Implementation

I borrowed most of the structure behind the [AFM implementation](https://anywidget.dev/en/afm/) from the few examples out there such as [marimo](https://github.com/marimo-team/marimo/blob/7f3023ff0caef22b2bf4c1b5a18ad1899bd40fa3/frontend/src/plugins/impl/anywidget/AnyWidgetPlugin.tsx#L161-L267).

Known issues:
- The `model.send()` method in AFM is left unimplemented, haven't checked how its used
- The cleanup methods returned by `initialize() / render()` are not currently used
- `model.save_changes()` currently sends all traits back to python including unchanged traits, which is unnecessarily expensive
- `altair`: JS errors when `altair.params` are updated but this is an issue with `altair`: https://github.com/vega/altair/issues/3868
- Right now linking nicegui --> anywidget via python-side `on_change` callbacks is very jittery -- see the recording below, I'm not sure what throttling / debouncing needs to applied. Is there an easy way to link the two NiceGUI elements on the JS side?

https://github.com/user-attachments/assets/8214591a-caef-4526-8a42-854bc2d2b081

### Progress

### DIFF
--- a/examples/altair/main.py
+++ b/examples/altair/main.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+"""Demonstrates embedding altair charts in NiceGUI.
+
+This shows how to embed altair charts in NiceGUI, and how to synchronize
+NiceGUI elements with charts using ``.bind_*`` (chart -> NiceGUI) and ``on_*``
+callbacks (NiceGUI -> chart).
+"""
+
+import altair as alt
+import numpy as np
+import pandas as pd
+
+from nicegui import ui
+
+
+def create_altair_chart() -> alt.Chart:
+    """Create simple altair chart with slider-based coloring
+
+    Refer to:
+    https://altair-viz.github.io/user_guide/interactions/jupyter_chart.html#accessing-variable-params
+    """
+    rand = np.random.RandomState(42)
+
+    df = pd.DataFrame({
+        'xval': range(100),
+        'yval': rand.randn(100).cumsum()
+    })
+
+    slider = alt.binding_range(min=0, max=100, step=1)
+    cutoff = alt.param(name='cutoff', bind=slider, value=50)
+
+    chart = alt.Chart(df).mark_point().encode(
+        x='xval',
+        y='yval',
+        color=alt.condition(
+            alt.datum.xval < cutoff,
+            alt.value('red'), alt.value('blue')
+        )
+    ).add_params(
+        cutoff
+    )
+    return chart
+
+
+@ui.page('/')
+def page():
+    # Example 1: getting started example from altair documentation
+    # https://altair-viz.github.io/getting_started/starting.html#customizing-your-visualization
+    with ui.card().classes('min-w-200px max-w-300px'):
+        ui.label('Static altair chart').classes('text-xl font-bold')
+
+        data = pd.DataFrame({'a': list('CCCDDDEEE'), 'b': [2, 7, 4, 1, 2, 6, 8, 4, 7]})
+        chart = alt.Chart(data).mark_bar(color='firebrick').encode(
+            alt.Y('a').title('category'),
+            alt.X('average(b)').title('avg(b) by category')
+        )
+        ui.altair(chart)
+
+    # Example 2: altair chart widget with slider synchronized between nicegui & altair
+    # This is the JupyterChart variable params example from the altair documentation:
+    # https://altair-viz.github.io/user_guide/interactions/jupyter_chart.html#accessing-variable-params
+    with ui.card().classes('min-w-200px max-w-300px'):
+        ui.label('Synchronized altair and nicegui sliders').classes('text-xl font-bold')
+        # jchart = AltairChart(create_altair_chart())
+        chart = create_altair_chart()
+        widget = ui.altair(chart, throttle=0.5)
+
+        def update_cutoff(change):
+            # NOTE: as of altair v5.5, there is a JavaScript side bug in altair.JupyterChart
+            # that causes this callback to fail when accessing .changed (only supported in Jupyter):
+            # https://github.com/vega/altair/issues/3868
+            widget._widget.params.cutoff = change.value
+        slider = ui.slider(
+            min=0, max=100, value=widget._widget.params.cutoff, on_change=update_cutoff, throttle=0.2
+        ).bind_value_from(widget._widget, '_params', backward=lambda p: p['cutoff'])
+        ui.label().bind_text_from(slider, 'value', backward=lambda c: f'Cutoff is {c}')
+
+
+ui.run(show=False)
--- a/examples/anywidget/main.py
+++ b/examples/anywidget/main.py
@@ -1,0 +1,337 @@
+#!/usr/bin/env python3
+"""Demonstrates embedding anywidget widgets in NiceGUI.
+
+This shows how to embed anywidget widgets in NiceGUI, and how to synchronize
+NiceGUI elements with widgets using ``.bind_*`` (widget -> NiceGUI) and ``on_*``
+callbacks (NiceGUI -> widget).
+
+The example shows how to embed a counter widget and an altair chart widget
+from the getting started examples in anywidget and altair.
+"""
+
+import altair
+import anywidget
+import traitlets
+
+from nicegui import ui
+
+
+class CounterWidget(anywidget.AnyWidget):
+    """Baseline anywidget example"""
+    _esm = '''
+    function render({ model, el }) {
+      let button = document.createElement("button");
+      button.innerHTML = `anywidget count is ${model.get("value")}`;
+      button.addEventListener("click", () => {
+        model.set("value", model.get("value") + 1);
+        model.save_changes();
+      });
+      model.on("change:value", () => {
+        button.innerHTML = `anywidget count is ${model.get("value")}`;
+      });
+      el.classList.add("counter-widget");
+      el.appendChild(button);
+    }
+    export default { render };
+    '''
+    _css = '''
+    .counter-widget button { color: white; font-size: 1.75rem; background-color: #ea580c; padding: 0.5rem 1rem; border: none; border-radius: 0.25rem; }
+    .counter-widget button:hover { background-color: #9a3412; }
+    '''
+    value = traitlets.Int(0).tag(sync=True)
+
+
+class AltairChart(altair.JupyterChart):
+    """
+    This is a wrapper around `altair.JupyterChart` that just adds some console
+    debugging output, and fixes a minor bug with accessing `.changed` (only
+    usable in Jupyter-widget-based environments).
+    """
+    _esm = '''
+import vegaEmbed from "https://esm.sh/vega-embed@v7?deps=vega@6&deps=vega-lite@5.33.0";
+import lodashDebounce from "https://esm.sh/lodash-es@4.17.21/debounce";
+
+// Note: For offline support, the import lines above are removed and the remaining script
+// is bundled using vl-convert's javascript_bundle function. See the documentation of
+// the javascript_bundle function for details on the available imports and their names.
+// If an additional import is required in the future, it will need to be added to vl-convert
+// in order to preserve offline support.
+async function render({ model, el }) {
+    let finalize;
+
+    function showError(error){
+        el.innerHTML = (
+            '<div style="color:red;">'
+            + '<p>JavaScript Error: ' + error.message + '</p>'
+            + "<p>This usually means there's a typo in your chart specification. "
+            + "See the javascript console for the full traceback.</p>"
+            + '</div>'
+        );
+    }
+
+    const reembed = async () => {
+        if (finalize != null) {
+          finalize();
+        }
+
+        model.set("local_tz", Intl.DateTimeFormat().resolvedOptions().timeZone);
+
+        let spec = structuredClone(model.get("spec"));
+        if (spec == null) {
+            // Remove any existing chart and return
+            while (el.firstChild) {
+                el.removeChild(el.lastChild);
+            }
+            model.save_changes();
+            return;
+        }
+        let embedOptions = structuredClone(model.get("embed_options")) ?? undefined;
+
+        let api;
+        try {
+            api = await vegaEmbed(el, spec, embedOptions);
+        } catch (error) {
+            showError(error)
+            return;
+        }
+
+        finalize = api.finalize;
+
+        // Debounce config
+        const wait = model.get("debounce_wait") ?? 10;
+        const debounceOpts = {leading: false, trailing: true};
+        if (model.get("max_wait") ?? true) {
+            debounceOpts["maxWait"] = wait;
+        }
+
+        const initialSelections = {};
+        for (const selectionName of Object.keys(model.get("_vl_selections"))) {
+            const storeName = `${selectionName}_store`;
+            const selectionHandler = (_, value) => {
+                const newSelections = cleanJson(model.get("_vl_selections") ?? {});
+                const store = cleanJson(api.view.data(storeName) ?? []);
+
+                newSelections[selectionName] = {value, store};
+                model.set("_vl_selections", newSelections);
+                model.save_changes();
+            };
+            api.view.addSignalListener(selectionName, lodashDebounce(selectionHandler, wait, debounceOpts));
+
+            initialSelections[selectionName] = {
+                value: cleanJson(api.view.signal(selectionName) ?? {}),
+                store: cleanJson(api.view.data(storeName) ?? [])
+            }
+        }
+        model.set("_vl_selections", initialSelections);
+
+        const initialParams = {};
+        for (const paramName of Object.keys(model.get("_params"))) {
+            const paramHandler = (_, value) => {
+                const newParams = JSON.parse(JSON.stringify(model.get("_params"))) || {};
+                newParams[paramName] = value;
+                model.set("_params", newParams);
+                model.save_changes();
+            };
+            api.view.addSignalListener(paramName, lodashDebounce(paramHandler, wait, debounceOpts));
+
+            initialParams[paramName] = api.view.signal(paramName) ?? null
+        }
+        model.set("_params", initialParams);
+        model.save_changes();
+
+        // Param change callback
+        model.on('change:_params', async (new_params) => {
+            console.log('change:_params', new_params);
+            for (const [param, value] of Object.entries(new_params)) {
+                console.log('change:_params', new_params, param, value);
+                api.view.signal(param, value);
+            }
+            await api.view.runAsync();
+            console.log('change:_params done');
+        });
+
+        // Add signal/data listeners
+        for (const watch of model.get("_js_watch_plan") ?? []) {
+            if (watch.namespace === "data") {
+                const dataHandler = (_, value) => {
+                    model.set("_js_to_py_updates", [{
+                        namespace: "data",
+                        name: watch.name,
+                        scope: watch.scope,
+                        value: cleanJson(value)
+                    }]);
+                    model.save_changes();
+                };
+                addDataListener(api.view, watch.name, watch.scope, lodashDebounce(dataHandler, wait, debounceOpts))
+
+            } else if (watch.namespace === "signal") {
+                const signalHandler = (_, value) => {
+                    model.set("_js_to_py_updates", [{
+                        namespace: "signal",
+                        name: watch.name,
+                        scope: watch.scope,
+                        value: cleanJson(value)
+                    }]);
+                    model.save_changes();
+                };
+
+                addSignalListener(api.view, watch.name, watch.scope, lodashDebounce(signalHandler, wait, debounceOpts))
+            }
+        }
+
+        // Add signal/data updaters
+        model.on('change:_py_to_js_updates', async (updates) => {
+            for (const update of updates.changed._py_to_js_updates ?? []) {
+                if (update.namespace === "signal") {
+                    console.log('change:_py_to_js_updates signal', update.name, update.scope, update.value);
+                    setSignalValue(api.view, update.name, update.scope, update.value);
+                } else if (update.namespace === "data") {
+                    console.log('change:_py_to_js_updates data', update.name, update.scope, update.value);
+                    setDataValue(api.view, update.name, update.scope, update.value);
+                }
+            }
+            await api.view.runAsync();
+        });
+    }
+
+    model.on('change:spec', reembed);
+    model.on('change:embed_options', reembed);
+    model.on('change:debounce_wait', reembed);
+    model.on('change:max_wait', reembed);
+    await reembed();
+}
+
+function cleanJson(data) {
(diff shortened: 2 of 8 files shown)


nicegui/elements/anywidget/anywidget.js
```diff
@@ -0,0 +1,129 @@
...
+        // Implement AFM: https://anywidget.dev/en/afm/
+        // References:
+        // * Marimo AFM impl:
+        // https://github.com/marimo-team/marimo/blob/7f3023ff0caef22b2bf4c1b5a18ad1899bd40fa3/frontend/src/plugins/impl/anywidget/AnyWidgetPlugin.tsx#L161-L267
+        const model = {
+          attributes: { ...this.traits },
+          callbacks: {},
+          get: function (key) {
+            log('Getting value for', key, ':', this.attributes[key]);
+            const value = this.attributes[key];
```
azjps: If this just `return value`, this raises an error:
```javascript
Uncaught (in promise) DataCloneError: Failed to execute 'structuredClone' on 'Window': #<Object> could not be cloned.
    at reembed (1b7ef5d7-1cf7-4947-8f06-0f33e5db9f95:29:20)
    at Object.render (1b7ef5d7-1cf7-4947-8f06-0f33e5db9f95:146:11)
    at anywidget.js:96:47
```

evnchn:
Thank you @azjps!

A very interesting PR which opens up a lot of possibilities. 

May be a better idea to target this for 3.0 branch since I'm not sure how many releases will be left for 2.x, and 3.0 has different JS dependency management and other changes. Also considering release schedule and workload. 


But please don't make that change until either @falkoschindler or @rodja agree - I'm just suggesting.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5129#issuecomment-3298776923

> no new feature release ... (2.4.25? Assume is typo)

@rodja if I understand correctly there will not be a 2.25, if so @azjps unfortunately you may have to adapt the code for 3.0's JS dependency management...

azjps:
@evnchn I suppose for https://github.com/zauberzeug/nicegui/pull/5021? Currently there's no 3rd party npm dependencies so if I rebase on top of the 3.0 branch, the examples still run fine. But I'm not too sure where the JS files should go now (in particular this one: `nicegui/elements/lib/anywidget/widget.js`) or if there's some better way to organize.

evnchn:
Oh I thought `nicegui/elements/lib/anywidget/widget.js` was from NPM... I also recalled that there were merge conflicts, but now I don't see any (maybe I misremembered)? Sorry for the confusion. 

Yes there is some changes in how things are organized in 3.0, but no I'm not familiar with it since I resumed contribution recently. 

Regardless, it looks like this will be in 3.0 so whether we merge this into main then the 3.0 branch, or merge this into 3.0 then that goes into main, just Git logistics at this point. It'll be fine either way.

falkoschindler:
Very very interesting, @azjps! Unfortunately it'll have to wait until after 3.0. I'm looking forward to reviewing it!

azjps:
@falkoschindler no problem, maybe it would be nice to include it as a selling feature with the 3.0 release? The current PR still works with 3.0.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5129#pullrequestreview-3253707600

Trying to feature-freeze for 3.0, but you can argue this is an independent new element so we can slip it in though. 

Main limiting factor is manpower to review, I think. 

To be honest, feature-freezing is mutually in opposition to a banger 3.0 release, but I guess we have to strike a balance somewhere.

azjps:
@evnchn Got it, yeah IMHO this is a new feature that is disjoint from any existing code, but up to you guys when you have capacity to roll out. I've added a trivial unit test and some barebones website documentation now to hopefully make it easier to review, can maybe flesh out a fuller example if I have some time but it would effectively just be an addendum to the `anywidget` documentation (https://anywidget.dev/en/getting-started/).

falkoschindler:
I finally looked into this PR to move forward with this very interesting feature. Here are some high-level thoughts:

1. I'm wondering if anywidget should be a strict dependency of NiceGUI. With around 10MB the zip isn't particularly small though (most of it is documentation 🙄).
2. altair.py should probably move out of the anywidget module because it is its own UI element.
3. Do we really need examples for `ui.anywidget` and `ui.altair`? It looks like we could easily show them as demos.
4. The file widget.js is copied from the anywidget repository, but the anywidget package is installed as a dependency. How can we make sure they won't diverge eventually? Do we need to pin the anywidget version? Or can we extract widget.js from the installed anywidget package?
5. We should merge from main (or rebase?), fix the conflicting pyproject.toml, and see if everything still works with the most recent library code.

evnchn:
> 10MB dependency, most docs

But I imagine if it is a strict dependency the friction is lower and the percentage of utilitzation is higher. Though I imagine if this is really a banger then the utilization will be high no matter the friction.

Still, some ideas: 

- Can we possibly have a mirror using Git submodule but strip the docs? 
- Or maintain a "lite" version which syncs with the parent? 
- Will either of this get us into license trouble?

Noghpu:
There's been some movement on that front from the anywidget team.

https://github.com/manzt/anywidget/discussions/901
https://github.com/manzt/anywidget/pull/902

Fingers crossed this helps.

falkoschindler:
Nice! 🤩

- anywidget-0.9.18.tar.gz (9.8 MB)
- anywidget-0.9.19.tar.gz (358.7 kB)

Thanks for the initiative, @Noghpu!

azjps:
> * I'm wondering if anywidget should be a strict dependency of NiceGUI. With around 10MB the zip isn't particularly small though (most of it is documentation 🙄).

Currently (as mentioned by `@manzt`) `anywidget` depends on `ipywidgets` which pulls in various [jupyter dependencies](https://github.com/jupyter-widgets/ipywidgets/blob/main/python/ipywidgets/setup.cfg#L36), so even though the `anywidget` package is not that heavy by itself anymore (thanks @Noghpu), I'd personally advise against it to be conservative. 

> * altair.py should probably move out of the anywidget module because it is its own UI element.

Sure, moving `altair.py` out seems sensible; into its own `altair/` directory?

> * Do we really need examples for `ui.anywidget` and `ui.altair`? It looks like we could easily show them as demos.

Re: examples, probably not needed, although I felt more comfortable adding verbose notes in the `examples/` as opposed to in the `documentation/` where I tried to be more concise

> * The file widget.js is copied from the anywidget repository, but the anywidget package is installed as a dependency. How can we make sure they won't diverge eventually? Do we need to pin the anywidget version? Or can we extract widget.js from the installed anywidget package?

Yeah I had the same question. The `anywidget` JS package itself depends on `solid-js`, `jupyter-widgets`, and a few other dependencies. My knowledge about JS vendoring is extremely limited so I figured I would just copy the few utility methods I needed, which have zero dependencies. But it would be nice if those utility methods could be vendored without any dependencies, as I imagine any UI framework that wants to support AFM would want to call into these methods or direct equivalents. I'll ask the maintainer for thoughts.

> * We should merge from main (or rebase?), fix the conflicting pyproject.toml, and see if everything still works with the most recent library code.

Sure, can do

falkoschindler:
> Currently (as mentioned by `@manzt`) `anywidget` depends on `ipywidgets` which pulls in various [jupyter dependencies](https://github.com/jupyter-widgets/ipywidgets/blob/main/python/ipywidgets/setup.cfg#L36), so even though the `anywidget` package is not that heavy by itself anymore (thanks @Noghpu), I'd personally advise against it to be conservative.

Oh wow. I just checked by pip-installing into a clean environment and the overall size went up from 149MB for NiceGUI to 190MB including anywidget. That's a lot...

> Sure, moving `altair.py` out seems sensible; into its own `altair/` directory?

I think, since it's a single file, it can live next to the other element files like audio.py and avatar.py.

> Re: examples, probably not needed, although I felt more comfortable adding verbose notes in the `examples/` as opposed to in the `documentation/` where I tried to be more concise

Ok. It's a bit confusion to have two small examples per main.py. It feels like we could massage them into multiple demos on the documentation page.
Bet let's leave them for now. Right now I'm getting a JavaScript error "Cannot read properties of undefined (reading '_params')" when using the slider for the second Altair example. We should fix that first.

> Yeah I had the same question. The `anywidget` JS package itself depends on `solid-js`, `jupyter-widgets`, and a few other dependencies. My knowledge about JS vendoring is extremely limited so I figured I would just copy the few utility methods I needed, which have zero dependencies. But it would be nice if those utility methods could be vendored without any dependencies, as I imagine any UI framework that wants to support AFM would want to call into these methods or direct equivalents. I'll ask the maintainer for thoughts.

Ok, maybe it's best to leave it like this. But let's update .gitattributes to mark this file as vendored.

> Sure, can do

👍🏻

azjps:
I've made the suggested changes, seems like everything is still working as expected.

> Bet let's leave them for now. Right now I'm getting a JavaScript error "Cannot read properties of undefined (reading '_params')" when using the slider for the second Altair example. We should fix that first.

This was an issue with `altair`, it should be fixed after https://github.com/vega/altair/pull/3871 (which was released 2 days ago in `altair==6.0`).

nicegui/elements/anywidget/anywidget.js
```diff
@@ -0,0 +1,129 @@
...
+      // If you have an API to add listeners, do so here (placeholder)
+      // this.api.addGlobalListener(this.handle_event);
+    },
+    update_trait(change) {
+      // Callback from Python traitlet backend change event
+      // change is a dictionary with 'trait', 'new', and 'old' keys
+      convertDynamicProperties(change, true);
+      this._log('Updating trait:', change);
+      if (change) {
+        this.model.attributes[change['trait']] = change['new'];
```
azjps: Should replace this with `this.model.set(change['trait'], change['new'])`

nicegui/elements/anywidget/anywidget.js
```diff
@@ -0,0 +1,129 @@
...
+            }
+            this.callbacks[event]?.delete(callback);
+          },
+          emit: function (event, value) {
+            if (this.callbacks[event]) {
+              this.callbacks[event].forEach(cb => cb(value));
+            }
+          },
+          send: function (content, callbacks, buffers) {
+            if (buffers) {
```
azjps: After looking at the `ipywidgets` docs, I found out that the python widget receives this in a callback registered by `on_msg()` so it seems like it should not be too hard to handle JS -> Python arbitrary communication.  https://ipywidgets.readthedocs.io/en/latest/_modules/ipywidgets/widgets/widget.html#Widget.on_msg 

Conversely for Python -> JS though I am not sure what one would do on an arbitrary `AnyWidget.send()`: https://ipywidgets.readthedocs.io/en/latest/_modules/ipywidgets/widgets/widget.html#Widget.send I don't see an obvious way to intercept the `send()` callback and redirect it through NiceGUI comms instead.

nicegui/elements/anywidget/anywidget.js
falkoschindler: We can use `cleanObject` from ../../static/utils/json.js for this purpose.

nicegui/elements/anywidget/anywidget.js
falkoschindler: Done. ✔️

falkoschindler:
Ok, I finally reviewed the whole PR. Looks really nice!

@azjps Is there anything left to do from your side? There's an [open conversation](https://github.com/zauberzeug/nicegui/pull/5137#discussion_r2566568134) about the JavaScript-Python communication. And there are "known issues" mentioned in the description. Maybe we can release it as it is and improve on it later?

azjps:
Thanks @falkoschindler for the cleanup. Yes actually I was about to ask about this; we did end up doing further work to avoid callback bouncing (e.g. JS -> Python -> JS; if there are multiple quick updates from JS then currently it can get stuck in a loop) by implementing the https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Low%20Level.html#comms `comms` API. We hadn't had a chance to consolidate the changes yet, possibly we can open that as another PR on top of this one?

evnchn:
@azjps Can't push to your branch since you worked on main. Only @falkoschindler can. 

I implemented JS->Py in https://github.com/evnchn/nicegui/tree/azjps/main (particularly https://github.com/evnchn/nicegui/commit/3a1a78359b1c10330212b259ff25b17e3c094f95). Both of you, please kindly check and adopt if see fit.

s-meza:
Hi, I've been working on implementing Python<-->JS communications for `ui.anywidget` (plus other compatibility things). 

Changes are here: https://github.com/s-meza/nicegui/tree/main (squashed, actual commits in the [`raw-commits`](https://github.com/s-meza/nicegui/tree/raw-commits) branch)

It's sizable amount of changes, so was wondering if you would like me to make a PR into @azjps's fork or a PR into NiceGUI itself (as a draft until this gets merged).

I explain a little more on the description of d98469e, but the gist is that I added shim `Comm` object which replaces `widget.comm` so that all communications are done through NiceGUI. I also implemented what was missing from the AFM and added some logic from Jupyter in order to handle recursive `model.set`s.

evnchn:
Hello @s-meza. Welcome to NiceGUI GitHub. Let me on-board with 2 things which I must mention:

#### Please do NOT put commits in main

Stated in https://github.com/zauberzeug/nicegui/blob/main/CONTRIBUTING.md#pull-requests. Let me recap here:

> [!IMPORTANT]
> Always work on a feature branch, never on `main`:
>
> 1. Working on `main` blocks you from contributing to multiple things simultaneously,
> 2. It highly increases the chance of an unclean commit history,
> 3. Your `main` branch should stay in sync with the upstream repository.

#### Good idea to open another Pull Request if the changes are drastic enough. 

I see https://github.com/s-meza/nicegui/tree/main diverges quite a bit from the implementation of https://github.com/azjps/nicegui/tree/main. I would recommend opening a new PR so that we are comparing your work directly to main, not your work <-> other's work <-> main. 

---

And now something to everyone here:

**I think we need to be careful and not over-engineer our implementation.** 

I read the following in the discussion which is concerning:

> Many links to https://ipywidgets.readthedocs.io/
> Python-side API definitions borrowed from ipywidgets such as `on_msg()`, `AnyWidget.send()`
> shim `Comm` object which exists in Python-side as well

However, it is of my personal opinion that we are implementing the Anywidget Front-End Module spec, which is very limited in scope: whatever is in https://anywidget.dev/en/afm/#model-interface

Implications include:

1. We may choose not refer to stuff in https://ipywidgets.readthedocs.io/
2. We may choose to use NiceGUI patterns for Python-side APIs, where we prefer `on_message()`
3. We may choose to not implement shim objects, like `Comm`, exactly as prescribed as Jupyter 

These divergances:

- [pros] Reduces work we have to do on our side
- [cons] Some frontend modules which are not fully in AFM spec (need extra functionality) may break

It is a decision that we best make up-front.

falkoschindler:
Yes, I think we should limit ourselves to the AFM spec for now. We can always extend it later.

Let me know if you guys think this PR is good enough for now and can be merged as a first version of anywidget support.

azjps:
Sure @evnchn sorry about that and thanks for pointing out, we'll keep to feature branches in the future. @falkoschindler yes, I reckon the current PR is fine for a first-cut, since for example static `altair` plots would be pretty useful on its own, we can iterate on improvements afterwards (e.g. those proposed by @s-meza, @evnchn). Thanks for reviewing :)

There is one issue that we for sure want to address (can be separate PR), which is that currently multiple fast updates from JS can bounce indefinitely. For example a state change followed by another change in quick succession while the first is in-flight: both changes will be received by Python, which will send both changes to JS, and so forth indefinitely. @s-meza and I are looking at a couple of options to address this.  

----

Regarding `comm.Comm`, it is an attempt to fulfill the AFM spec -- the rationale is that:
- `widget.observe()` lets us listen to changes to changes to `trait`s, but there's no corresponding public API that listens to [`widget.send()`](https://github.com/jupyter-widgets/ipywidgets/blob/fbc0d6e9cd63727b5951f2bf559f1a5769e4455d/python/ipywidgets/ipywidgets/widgets/widget.py#L657). So currently, to my knowledge, the only way to implement this corner of the AFM (Python `widget.send()` -> JS `model.on("custom", ..)`) is to implement the `comm` API. Alternatively there's probably some way to patch the widget object but that'd probably be considered more intrusive.
- We hadn't noticed this originally, but [`marimo` also implements `anywidget` by implementing the `Comm` API](https://github.com/marimo-team/marimo/blob/ad9e11c4f7ce607e64e5d62c7d4d47a73273a58d/marimo/_plugins/ui/_impl/anywidget/init.py#L36). Notably [`marimo` does not support `ipywidgets`](https://github.com/marimo-team/marimo/issues/5099).
- Functionally, `comm` just prescribes how to relay messages, so the [implementation of `Comm` here is just to call nicegui's `run_method()`](https://github.com/azjps/nicegui/blob/d98469eb804240c7b0b86ffb7d69fd90e1f9cb6b/nicegui/elements/anywidget/comm.py#L44 ). There's a bit more boilerplate but its not too much imho 
- The nice part is that we no longer need to `observe()` the traits ourselves, instead everything automatically comes out of the internal `ipywidgets` machinery. It also handles the bouncing issue mentioned above directly, since `widget.set_state()` does not call `comm.publish()` and hence breaks the Python->JS bouncing.
- The unfortunate part is that this does mutate the `anywidget` object to set the `comm` object on the `Anywidget` instance. But this is also exactly what `jupyter` & `marimo` do.  

All that being said, I wished the `anywidget` docs explicitly mentioned whether one should implement the `Comm` API, I believe their goal is that it will not be necessary in the future, but it seems to be the only practical way currently.

s-meza:
Just as a follow up, following @evnchn's advice I've extracted some small-but-important fixes from my original branch onto [`anywidget-fixes`](https://github.com/s-meza/nicegui/tree/anywidget-fixes). I believe these should ideally be included with this first version of anywidget support (see https://github.com/azjps/nicegui/pull/2):
1. Stops the Python <--> JS update loop
2. Makes the `model.on("change:...")` callback arguments match Jupyter's behaviour

evnchn:
@s-meza https://github.com/azjps/nicegui/pull/2 looks good. Diff small and should function. It also goes along with https://github.com/azjps/nicegui/pull/1 as well (no conflicts I think)

---

@azjps personally I would recommend YOLO-merge the two PRs on your branch then @falkoschindler can still modify / revert commits if for some reason it isn't liked. Having the discussion at 3 positions is not ideal and I kinda regret starting the trend, though I have no choice since you are merging from your fork's main and I cannot commit directly to your branch. **You have the final call though, and you may want to wait for @falkoschindler's opinion.** 

---

On the `comm` API, my intepretation of https://github.com/zauberzeug/nicegui/pull/5137#issuecomment-3708756221 is that **while it is not de-jure, it is de-facto**. Considering that part of the `anywidget` ecosystem may already be entrenched into the `comm` API, I see the best move is for us to follow suit, even though it would be doing extra on top of strictly the AFM spec. 

In other words, they need a "Anywidget Backend Specification", which while is not a thing yet, we can infer from other implementers. 

@s-meza If that is the case then you may need to dig up some commits in https://github.com/s-meza/nicegui/tree/raw-commits since I see your `main` is cleared. I would still recommend either a cleaner diff based on this PR's code, or open another PR if the implementation is sufficiently different and the diff to this PR would be big.

evnchn:
@falkoschindler is https://github.com/azjps/nicegui/pull/1 needed as well? Implements what was log-warned in `console.warn("anywidget.send() is not yet implemented in NiceGUI", content);`

falkoschindler:
@evnchn As far as I understand, it is not needed for the demos in this PR. At least I don't get any "not yet implemented" warnings when running the demos. So we can implement `anywidget.send()` later and improve `ui.anywidget` incrementally.

But I just noticed that the "Interactive charts" demo logs errors when moving the slider:
```
Uncaught (in promise) Error: Unrecognized signal name: "attributes"
    at j (vega-util.mjs:2:321)
    at m (vega-view.mjs:3:1647)
    at _t.signal (vega-view.mjs:3:2569)
    at f6713122-2ebf-488e-9208-f486dba43080:95:26
    at anywidget.js:39:48
    at Array.forEach (<anonymous>)
    at Object.emit (anywidget.js:39:32)
    at Object.set (anywidget.js:21:14)
    at paramHandler (f6713122-2ebf-488e-9208-f486dba43080:82:23)
    at S (debounce.js:95:19)
```
It doesn't happen when reverting https://github.com/azjps/nicegui/pull/2/changes/2c1d5145b05dfd55a80e6c474d8effcb4df8cf34.
@azjps Can you, please, have a look what's wrong? Thanks!

azjps:
@falkoschindler yeah, we noticed too -- as it turns out `Jupyter` and `Marimo` have completely different signatures of the JS callbacks triggered on trait change, and the AFM currently doesn't specify what the signature should be. [Although presumably it was intended to be that used in Jupyter, since it came first.] I had originally followed the signature in Marimo, then we changed it to the one used in Jupyter, which then broke the `altair` handling (which uses a Jupyter-internal attribute of one of the arguments passed in the Jupyter signature).

We'll revert that part of the change for now so that this PR can be merged as-is, and start up discussions in anywidget and marimo to get the signatures to line up. Likely small changes will be needed in the anywidgets docs, marimo, altair, and here; hopefully can get the anywidget maintainers to weigh in.

azjps:
Wonderful! 

> I still find the whole Jupyter/Marimo/AFM universe a bit confusing. Therefore I'm glad that you're contributing the knowledge for this powerful feature. Discussing the open questions with the anywidget and/or Marimo teams would certainly be very helpful.

Yes its been pretty confusing for us too, I've opened https://github.com/manzt/anywidget/issues/931 & https://github.com/manzt/anywidget/issues/932 to hopefully get more clarity. After that it should inform us about what to do with regards to https://github.com/azjps/nicegui/pull/1 (@evnchn).

Issue #5142, opened 2025-09-15
ui.sub_pages.add with async function results in `RuntimeWarning: coroutine 'main' was never awaited`


platinops:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui


@ui.page("/")
@ui.page("/{_:path}")
def index():
    # ui.sub_pages({"/": main, "/other": other}) # -> this works fine
    pages = ui.sub_pages()
    pages.add("/", main)
    pages.add("/other", other) # -> this results in warning


async def main():
    ui.label("main page")


async def other():
    ui.label("other page")


ui.run()
```

### Description

Run the example.

Browse to http://127.0.0.1:8080/other

In terminal: `RuntimeWarning: coroutine 'main' was never awaited`

### NiceGUI Version

2.24.1 and 3.0.0rc1

### Python Version

3.11.9

### Browser

Firefox

### Operating System

Windows

evnchn:
Hello @platinops 

I managed to find where the problem is, but don't yet have a fix. 

1. If you pass a dict into `ui.sub_pages` it writes to `self._routes` since `self._routes = routes or {}`. That works fine so surely no problem with `ui.sub_pages` in handling the routes in `self._routes`
2. Interestingly in `pages.add` besides `self._routes[path] = page` it also does `self._show()`. 

So yeah, likely `self._show()` is the culprit. But I can't just comment it out or else I get `HTTPException: 404: /other not found`

evnchn:
The `ui.sub_pages` codebase is too foreign to me. I can't understand the logic of it, and didn't come up with a fix.

python-and-novella:
try this:
```python3
from nicegui import ui

@ui.page("/")
@ui.page("/{_:path}")
def index():
    # ui.sub_pages({"/": main, "/other": other}) # -> this works fine
    pages = ui.sub_pages()
    pages.add("/other", other) # -> this results in warning
    pages.add("/", main)

async def main():
    ui.label("main page")

async def other():
    ui.label("other page")

ui.run()
```

python-and-novella:
or visit other page by click link:
```python3
from nicegui import ui

@ui.page("/")
def index():
[...]

PR #5145, opened 2025-09-16
Close sub_pages render coroutine on cancellation

rodja:
### Motivation

As @platinops described in #5142, using an async function as `/` and then calling `sub_pages.add` results in a "coroutine was never awaited" warning. Reproduction:

```py
@ui.page('/')
@ui.page('/{_:path}')
def index():
    pages = ui.sub_pages({'/': main})
    pages.add('/other', lambda: ui.label('other page'))

async def main():
    ui.label('main page')
```

### Implementation

If the renderer task was cancelled before it could start, we must close it. I also added a test, but discovered that it's just a warning. After some experiments I decidet to not added "warnings as errors" to the `pyproject.toml`. So the test passes but prints a warning. We can turn on "warning as errors" in a different PR (targeting 3.0).

### Progress

### DIFF
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -113,9 +113,16 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
             async def background_task():
                 with self:
                     await result
+
             task = background_tasks.create(background_task(), name=f'building sub_page {match.pattern}')
             self._active_tasks.add(task)
-            task.add_done_callback(self._active_tasks.discard)
+
+            def _close_if_canceled(t: asyncio.Task) -> None:
+                if t.cancelled():
+                    result.close()
+                self._active_tasks.discard(t)
+
+            task.add_done_callback(_close_if_canceled)
         return True
 
     def _render_404(self) -> None:
--- a/nicegui/testing/general_fixtures.py
+++ b/nicegui/testing/general_fixtures.py
@@ -1,4 +1,3 @@
-import asyncio
 import importlib
 from copy import copy
 from typing import Generator, List, Type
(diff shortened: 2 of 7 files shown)


evnchn:
Consider `error::RuntimeWarning`,  `error` turning all warnings into errors may be too strict (Selenium has some warnings from time to time, iirc)

evnchn:
May be useful if you are doing asyncio and it breaks on py3.8-3.9

https://github.com/zauberzeug/nicegui/pull/4671#issuecomment-2847601782

It's 5 months old but I WROTE IT so I remember it very well.

rodja:
@evnchn you are right. "warnings as errors" does not fit well with this small bugfix. I reverted everything -- mainly because in 3.0 the issues with 3.8/3.9 are gone (because we removed the auto-index client)

Would you like to create a PR with "warnings as errors". From this PR, just the commit "fix warning for test_user_storage_is_pruned" (0f7494a) might be reusable. The other attempts are obsolete with 3.0.

evnchn:
Hello @rodja

The PR looks good. Manually checked it does indeed resolve #5142. Solution is simple too. Great observation that 3.0 the issues with 3.8/3.9 are gone, it simplified things. **Good to merge if all other tests pass.** 

But, try as I might (tried `@pytest.mark.filterwarnings('error::RuntimeWarning')`) I can't make `test_adding_page_after_async_sub_page` error when I revert to the old code. The test passes with warnings. Maybe some pytest expert can help us so that we can avoid accidentally regressing this thing. 

About warnings as errors so as to get rid of all warnings in the test suite, it will be a long-term goal. Let's not let that block the bugfix. Maybe we can put in NiceGUI Wishlist and create a draft branch to work for that progress.

evnchn (approved):
It's good

PR #5150, opened 2025-09-18
Add table demo: Delete button in Actions column using Quasar template

Mogit21:
### Motivation

Add an example to the NiceGUI table documentation showing how to include a Delete button in a custom “Actions” column using Quasar template slots. This addresses a common question in the discussions about rendering per-row buttons in tables, which wasn’t clearly covered in the existing docs.

### Implementation

Added a new demo in website/documentation/content/table_documentation.py using table.add_slot() with a Quasar template.

The demo shows a red Delete button in each row’s Actions column that triggers a notification when clicked.

Could you ensure the button displays correctly and is aligned with the row content?

### Progress

- [x] Pytests have been added (or are not necessary). [x] (not required for UI demo)
- [x] Documentation has been added (or is not necessary). [x] (demo added directly to table documentation)

Table_delete_button

### DIFF
--- a/website/documentation/content/table_documentation.py
+++ b/website/documentation/content/table_documentation.py
@@ -111,6 +111,40 @@ def table_with_expandable_rows():
         </q-tr>
     ''')
 
+@doc.demo('Table with Delete button', '''
+    Each row has a Delete button in the Actions column.
+    Clicking the button shows a notification with the row name.
+''')
+def table_with_delete_button():
+    columns = [
+        {'name': 'id', 'label': 'ID', 'field': 'id'},
+        {'name': 'name', 'label': 'Name', 'field': 'name'},
+        {'name': 'age', 'label': 'Age', 'field': 'age'},
+        {'name': 'actions', 'label': 'Actions'},
+    ]
+
+    rows = [{'id': 0, 'name': 'Paul', 'age': 38}]
+
+    table = ui.table(columns=columns, rows=rows, row_key='id').classes('w-full')
+
+    # Quasar template for Actions column
+    table.add_slot(
+        'body-cell-actions',
+        '''
+        <q-td auto-width :props="props">
+            <q-btn 
+                size="sm" 
+                color="red" 
+                label="Delete" 
+                flat 
+                dense 
+                round
+                @click="() => $q.notify({message: 'Deleted ' + props.row.name + '!'})"
+            />
+        </q-td>
+        '''
+    )
+
 
 @doc.demo('Show and hide columns', '''
     Here is an example of how to show and hide columns in a table.


website/documentation/content/table_documentation.py
```diff
@@ -111,6 +111,40 @@ def table_with_expandable_rows():
...
+@doc.demo('Table with Delete button', '''
+    Each row has a Delete button in the Actions column.
+    Clicking the button shows a notification with the row name.
+''')
+def table_with_delete_button():
+    columns = [
+        {'name': 'id', 'label': 'ID', 'field': 'id'},
+        {'name': 'name', 'label': 'Name', 'field': 'name'},
+        {'name': 'age', 'label': 'Age', 'field': 'age'},
+        {'name': 'actions', 'label': 'Actions'},
```
evnchn: If you want to align this column center, check https://quasar.dev/vue-components/table/#defining-the-columns

website/documentation/content/table_documentation.py
```diff
@@ -111,6 +111,40 @@ def table_with_expandable_rows():
...
+    ]
+
+    rows = [{'id': 0, 'name': 'Paul', 'age': 38}]
+
+    table = ui.table(columns=columns, rows=rows, row_key='id').classes('w-full')
+
+    # Quasar template for Actions column
```
evnchn: Is this right? IIRC our documentation strips comments and so we would literally have `Quasar template for Actions column` in the source code displayed. 

https://github.com/zauberzeug/nicegui/blob/main/website/documentation/code_extraction.py#L7-L11

website/documentation/content/table_documentation.py
```diff
@@ -111,6 +111,40 @@ def table_with_expandable_rows():
...
+        '''
+        <q-td auto-width :props="props">
+            <q-btn 
+                size="sm" 
+                color="red" 
+                label="Delete" 
+                flat 
+                dense 
+                round
+                @click="() => $q.notify({message: 'Deleted ' + props.row.name + '!'})"
```
evnchn: Hmm... This doesn't delete anything though...

Now our original goal is to show how to add buttons. 

- Should we also demo how to actually delete the row? 
- If not how about simpler actions such as simply showing the age of the person in a notify?

evnchn (changes requested):
Provided suggestions on how to align center, point out potential issue with including comments in the code (verify later) and suggested against adding a delete button which doesn't actually delete things (2 alternatives suggested)

website/documentation/content/table_documentation.py
evnchn: How about "Say hello" which does "Hello Alice!"?

Mogit21: @evnchn Thank you for your suggestions! I will go through the links that you have provided this evening (ADT) and will make the changes.
I also agree regarding the action that would greet with a "Hello name" instead of "Delete" and keep the example simple enough, the entry is in fact not deleted and it feels that the action is incomplete.

Mogit21: Thanks again for the suggestions! I’ve updated the table example:

- The button now says “Say hello” instead of “Delete”.

- The Actions column is centered.

- Added a second row to demonstrate multiple rows.

- Changed the button color for better visibility.

- Removed the outdated comment about “Quasar template for Actions column”.

- Updated the @doc.demo decorator: the example now clearly highlights the concept of adding interactive buttons to table cells, with each button working independently for its row.

Screenshot attached for referenc
Table_greet_button
e.

website/documentation/content/table_documentation.py
```diff
@@ -112,6 +112,39 @@ def table_with_expandable_rows():
     ''')
 
 
+@doc.demo('Table with Say Hello button', '''
+    Each row has a "Say hello" button added to the Actions column.
+    Clicking the button shows a notification greeting the row name.
+    This demonstrates how to add interactive buttons to table cells 
+    in a specific column, with each button working independently for its row.
+''')
```
evnchn: You're almost there, but I don't think that's how the title and the content is intended:

- the title needs to be generic and concise (e.g. "Table with Button in table cells")
  - this is because the title is weighted quite heavily in the search, additional terms may lower the match confidence
- the gist of the content, having a "Say hello" button is an _instance_ to show the above title's functionality
  - so the 3rd and 4th line goes before the 1st and 2nd line, essentially, with rephrasing afterwards

website/documentation/content/table_documentation.py
```diff
@@ -112,6 +112,39 @@ def table_with_expandable_rows():
...
+    table.add_slot('body-cell-actions', r'''
+        <q-td auto-width :props="props">
+            <q-btn 
+                size="sm" 
+                color="blue" 
+                label="Say hello" 
+                flat 
+                dense 
+                round
+                @click="() => $q.notify({message: 'Hello ' + props.row.name + '!'})"
```
evnchn: No need to force yourself, but I'm thinking about whether we can emit events using JavaScript so that we trigger Python code. 

This will be so much more powerful than what we can do in JS-land. 

Check https://nicegui.io/documentation/generic_events#custom_events, and also try something like `this.$emit` (this is everywhere in the codebase, do search for it). Latter solution preferred if possible since that event will be specific to the table element. 

If not possible it can be done as an enhancement to this documentation after this has been merged. Thanks!

evnchn (changes requested):
We're almost there. Just suggested changes to the title and content to facilitate searching and match the other documentation's style. 

Emitting event to the server to trigger Python action is a bonus item and so please don't get stuck at trying to do it.

website/documentation/content/table_documentation.py
```diff
@@ -112,6 +112,39 @@ def table_with_expandable_rows():
...
+def table_with_say_hello_button():
+    columns = [
+        {'name': 'id', 'label': 'ID', 'field': 'id'},
+        {'name': 'name', 'label': 'Name', 'field': 'name'},
+        {'name': 'age', 'label': 'Age', 'field': 'age'},
+        {'name': 'actions', 'label': 'Actions', 'align': 'center'},
+    ]
+    rows = [
+    {'id': 0, 'name': 'Alice', 'age': 25},
+    {'id': 1, 'name': 'Bob', 'age': 30},
```
evnchn: Indentation issue?

Mogit21:
Hi @evnchn , I explored your suggestion to emit events from JavaScript so that Python code is triggered. I tried implementing it using $emit from table slots to Python, but in the current NiceGUI version, $emit inside table slots does not propagate to Python.
I propose we keep the current example for now, which shows interactive buttons triggering a JS notification. I can revisit adding Python event handling as an enhancement in the future when I'll know how the NiceGUI table component fully supports it.
Please let me know your thoughts.
Thanks!

evnchn:
I did see `$parent.$emit` in https://nicegui.io/documentation/table#table_with_drop_down_selection so maybe we can give it one more shot, but yeah I agree that's not the focus and we can return to it at a later stage.

Mogit21:
> I did see `$parent.$emit` in https://nicegui.io/documentation/table#table_with_drop_down_selection so maybe we can give it one more shot, but yeah I agree that's not the focus and we can return to it at a later stage.

No worries, I should be able to have a look at it starting from tomorrow, and git it one more try. I'll keep you updated. Thank you,

evnchn:
That's great. Note also that the milestone changed to 3.x, keeping 2.24.2 laser-focused as a bugfix. So, be in no rush.

falkoschindler:
Hi @Mogit21, thanks for contributing to the documentation of NiceGUI!

I just had a brief look into your PR. At the moment the button is a bit of a dead end because it doesn't send any information back to the server. I'd suggest to use the "drop down selection" demo as an inspiration and implement the button in a similar way. You can even move your demo above the "drop down selection" since it kind of serves as a preparation.

Feel free to let us know if you're not working on it anymore and if we should take over to finish the PR. Thanks again!

Mogit21:
Thanks for reviewing my PR! I’m not able to continue with the changes at the moment, so please feel free to take over.
Appreciate the opportunity to contribute!

PR #5161, opened 2025-09-22
No orjson for PyPy

evnchn:
### Motivation

As mentioned in https://github.com/zauberzeug/rosys/pull/301#issuecomment-3310849125, NiceGUI does not work under PyPy normally. I discovered the reason (at least one of it) being orjson, as "orjson does not and will not support PyPy" (https://github.com/ijl/orjson/blob/aa8c946cfbe2aba29016bc9c9ddc8713cb00173b/README.md?plain=1#L35)

It would be a pity if we can't use NiceGUI under PyPy since its garbage collector does not "stop the world" unlike CPython's, an awesome use case for the robots which RoSys may empower

### Implementation

Add `platform_python_implementation != 'PyPy'` to the markers of orjson marking it out of 32-bit systems, such that if it is PyPy orjson will not be installed. 

This is inspired from https://github.com/pyca/cryptography/blob/cac6426aafd989e4d8d9c4f6b920a332a5280bba/pyproject.toml#L8

### Progress

- [x] Pytests have been are not necessary.
  - Original Pytests already handle missing orjson. Check https://github.com/zauberzeug/nicegui/blob/a6354dc56d2837ede75b8498766ce85a5fc97d8e/tests/test_json.py#L13-L20
  - But do make it very clear that this is BETA or even ALPHA level support...


evnchn:
It works. I tried it in the devcontainer. 

- `sudo apt install pypy3 pypy3-dev pypy3-venv`
- `pypy3 -m venv venv_pypy` then `source venv_pypy/bin/activate`
- `pip install git+https://github.com/MagicStack/httptools@e265ae0` (Why: Check https://github.com/MagicStack/httptools/issues/122#issue-2892480985)
- `pip install -e .`
- `pip install isort` to run the documentation website, by the way.

evnchn:
Feel like we should YOLO and merge this. 

Originally: 

- Works for CPython. 
- Does NOT work for PyPy whatsover. 

Now:

- Still works for CPython, `platform_python_implementation != 'PyPy'` is true after all. 
- If it doesn't work for PyPy, we're not regressing; But if it does then great!

Tests pass, by the way (https://github.com/evnchn/nicegui/actions/runs/17919770924). Please ignore 3.9-3.12 results on my branch since their cache's messed up.

evnchn:
#5163 is better.

evnchn:
I'm not up to resolving merge conflict for a partial PR...

evnchn:
This PR will likely have unsalvageable merge conflicts after #5311 merged...

falkoschindler:
@evnchn I just merged main and resolved (rewrote) the conflicting line in pyproject.toml. Can you check if it works?

evnchn:
@falkoschindler I am not quite familiar with uv and so I am very confused when I get:

```py
vscode@b1ae8fe6a0ad:/workspaces/nicegui$ uv run pip install pydantic-core 
  × Failed to build `pydantic-core==2.41.4`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `maturin.build_wheel` failed (exit status: 1)

      [stdout]
      Running `maturin pep517 build-wheel -i /home/vscode/.cache/uv/builds-v0/.tmpfet4m7/bin/python --compatibility off`
      Rust not found, installing into a temporary directory

      [stderr]
      Warning: cannot find your CPU L2 & L3 cache size in /sys/devices/system/cpu/cpuX/cache
      Python reports SOABI: pypy310-pp73
      Unsupported platform: pp73

      hint: This usually indicates a problem with the package or the build environment.
  help: `pydantic-core` (v2.41.4) was included because `nicegui` depends on `pydantic-core`
```

Despite there is prebuilt pypy wheels for pypy310-pp73 at https://pypi.org/project/pydantic_core/#files

Anyways I think we just need the package to be installable, so let's just roll with it?

PR #532, opened 2023-03-16
Fallback to Python's built-in json module if orjson unavailable

rbeeli:
`orjson` does not support all architectures. For those cases, if the `orjson` module import fails, the json serializer abstraction falls back to Python's built-in `json` module.

The two implementations have been adjusted/extended in order to output equal JSON strings, e.g. for date/datetime/numpy arrays etc. To ensure this, a new unit test `test_json.py` has been added.

PR #5163, opened 2025-09-22
PyPy support with test pipeline

evnchn:
### Motivation

Further develop on #5161. For PyPy to be first-class (or second-class) citizen of NiceGUI it has to, at the very least, run our test suite

### Implementation

- Add PyPy into the pipeline. 
- Expand the timeout to 120 minutes for now
- Drop LangChain, OpenCV, numpy (and other things which use numpy) from test
- Drop reference counter related test
- Loosen timing for event tests

### Progress

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5758,4 +5758,4 @@ sass = ["libsass"]
 [metadata]
 lock-version = "2.1"
 python-versions = "^3.9"
-content-hash = "28faa44617fe650e73bf260e08d68ffe595f3928347c86b1a4f3d9b84d561f12"
+content-hash = "5f870593ce31ddeb7941a344666c99997368cdc79d2add53f6648f4a0f95f859"


evnchn:
Missing pydantic core. Huh. But pydantic was installed, no?

https://github.com/evnchn/nicegui/actions/runs/17921019299

evnchn:
pydantic-core is a rust component. Until we figure out why it fails in the pipeline (it doesn't fail installing on my machine) we can use Pydantic V1 for now: https://github.com/fastapi/fastapi/blob/master/pyproject.toml#L31

evnchn:
@falkoschindler Should we reconsider simply letting NiceGUI install with PyPy in #5161 first? Because as it turns out https://github.com/zauberzeug/nicegui/pull/532#issuecomment-3478416665 there's desire for it. 

As discussed somewhere else, this PR may take a long time / never make it, due to the many test exceptions and large change size...

falkoschindler:
@evnchn Yes, the small change in PR #5161 seems very appealing compared to this heavy PR. I guess we can agree to provide second-class support for PyPy: We'll see that it works, fix bugs if necessary, but don't add tons of workarounds to get all 600+ tests green for PyPy.

evnchn:
Give me ~~5~~ 3 minutes

evnchn:
This PR may still be useful if first-class support is needed. After all, we would not want to run PyPy on the field robots without a tested NiceGUI as a base.

PR #5166, opened 2025-09-23
Move Tailwind layer less important than Quasar so that Tailwind wins when with !important

### LINKED ISSUE #5156: In NiceGUI 3.0.0rc1,I can't apply TailWindCSS color forcely
### First Check

### Example Code

```python
from nicegui import ui

def index():

index()

ui.run(native=True)
```

### Description

Does it works as expected?

### NiceGUI Version

3.0.0rc1

### Python Version

Python 3.12.10

### Browser

Other

### Operating System

Windows

### Additional Context

It shows this:

Image

Image

evnchn:
### Motivation

### Implementation

Now we move Tailwind layer to be lower than Quasar. 

- Tailwind still loses to Quasar's `!important` if Tailwind isn't begin with `!`. It didn't have a chance without `!` anyways
- Tailwind now wins Quasar precisely because layer order is reversed if both rules are `!important`

### Progress

  - I'm contemplating whether to Pytest this. 
  - We **need** to document this.

### DIFF
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -5,7 +5,7 @@
     <meta name="viewport" content="{{ viewport }}" />
     <link href="{{ favicon_url }}" rel="shortcut icon" />
     <style>
-      @layer theme, base, quasar, nicegui, components, utilities, overrides;
+      @layer theme, base, utilities, quasar, nicegui, components, overrides;
       @import url("{{ prefix | safe }}/_nicegui/{{version}}/static/fonts.css") layer(base);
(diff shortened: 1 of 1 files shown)


falkoschindler:
Unfortunately this breaks the layout in various places (see main.py).

evnchn:
image

😶‍🌫️

evnchn:
I have to give up on this since it's way too hard for me...

evnchn:
Not even the tests pass. 

🤡

Issue #4415, opened 2025-03-03
Quasar color palette colors generate classes that are all !important, pollutes override colors with !important


spacether:
### Description

Generation of css with specified colors results in css like (quasar.prod.css)
```css
.text-dark {
    color: var(--q-dark)!important
}

.bg-dark {
    background: var(--q-dark)!important
}

.text-primary {
    color: var(--q-primary)!important
}

.bg-primary {
    background: var(--q-primary)!important;
}

.text-secondary {
    color: var(--q-secondary)!important
}

.bg-secondary {
    background: var(--q-secondary)!important
}

.text-accent {
    color: var(--q-accent)!important
}

.bg-accent {
    background: var(--q-accent)!important
}

.text-positive {
    color: var(--q-positive)!important
}

.bg-positive {
    background: var(--q-positive)!important
}

.text-negative {
    color: var(--q-negative)!important
}
```
All of these colors are marked as !important.
So if any colors need to override these they also need to use !important.
This results in important pollution when one should use more specific css classes instead.
How about removing the !importants from the css class definitions?

falkoschindler:
Hi @spacether,

I agree: Using "!important" everywhere is indeed annoying.

But modifying quasar.prod.css doesn't sound great either. The impact is hard to foresee and we would need to check _why_ Quasar does things this way. By sticking to the original source, users can trust the Quasar documentation without exceptions introduced by NiceGUI. Therefore we're not planning to do that.

Issue #371, opened 2023-02-12
The .bg-primary class is messing with Tailwind's Background Color styling system on Buttons


Diegiwg:
Referenced class:

```css
.bg-primary {
    background: var(--q-primary) !important;
}
```

When we remove the `!important`, all the buttons turn white.

I tried
```python
styles.Backgrounds(ui.button("Test")).background_color("bg-red-800")
```
and it didn't work. It continued with the normal NiceGUI color.

Then, when I disabled the style of the class in the developer tool, it worked.

---

Task list:

- [x] add a parameter `ui.button(..., color: Optional[str] = 'primary')`
    - [x] handle `None`
    - [x] handle Quasar color palette
    - [x] handle CSS color
    - [x] handle Tailwind colors?
- [x] same for `ui.avatar` and other elements
- [x] tests
- [x] documentation

falkoschindler:
Thanks for reaching out, @Diegiwg!
Your description is a bit vague, but I guess I understand what you're saying:
You're referring to Quasar's CSS definition of `.bg-primary` which is used by `ui.button`. And `background_color` basically sets the Tailwind class "bg-red-800". Since Quasar defines its background color as `important`, it's not easy to overwrite it with a Tailwind color. That's the downside of combining two UI frameworks with overlapping style definitions.

But I'm not sure what to do about it. `ui.button`, which is based on [QBtn](https://quasar.dev/vue-components/button), creates a rather complex HTML element with various styling classes. I doubt we can avoid clashes with Tailwind in such cases.

Diegiwg:
I tried using classes(remove="bg-primary") on the button, but it didn't work.

The class continues to exist in the button. I also don't know how to solve this problem, other than applying the style manually with !important.

falkoschindler:
I'll close this issue. We don't see any (easy) way to avoid these kind of conflicts between Quasar and TailwindCSS.

miqsoft:
Hey,

I used the following solution for the problem and it solved it.

``` python      
btn = ui.button('CustomButton')
del btn._props['color']
btn.classes('bg-slate-300')
```

It's still not optimal since we need to modify the internal _props dictionary.
An option would be to implement a method to the button class that removes color from props.

rodja:
@Diegiwg we want track the idea @falkoschindler described https://github.com/zauberzeug/nicegui/issues/371#issuecomment-1494242035 on this card to make it more simple for users to actually change the background color of buttons (and other elements).

[...]

PR #5167, opened 2025-09-23
Remove important flags from Quasar's CSS definitions

falkoschindler:
### Motivation

Quasar's "!important" CSS definitions are very hard to override, especially after switching to Tailwind 4 which uses CSS layers.

### Examples

```py
ui.button('button').classes('bg-red-700')  # doesn't work in 2.x and 3.0.0rc1, but with this PR
ui.button('button').classes('!bg-red-700')  # doesn't work in 3.0.0rc1, but used to work in 2.x
ui.button('button', color=None).classes('!bg-red-700')  # workaround for 3.0.0rc1
```

```py
dark = ui.dark_mode()
ui.switch().bind_value(dark, 'value')
ui.label('Hi!').classes('invisible dark:visible')  # doesn't work in 2.x and 3.0.0rc1, but with this PR
ui.label('Hi!').classes('invisible dark:!visible')  # doesn't work in 3.0.0rc1, but used to work in 2.x
ui.label('Hi!').classes('!invisible dark:!visible')  # workaround for 3.0.0rc1
```

### Implementation

This PR strips all "!important" flags from quasar.css and quasar.prod.css.
While at it, it also move the QCard hack to extract_core_libraries.py so that all hacks are done in one place.

### Background

See https://github.com/zauberzeug/nicegui/issues/5156#issuecomment-3324426595 for ChatGPT's explanation why Quasar is using "important" everywhere for stability. With this context in mind, it seems reasonable to remove it for the use with NiceGUI and Tailwind as additional UI frameworks.

### Progress

  - [x] remove "!" from demos and examples where possible?
  - [ ] fix `ui.color_input`/`ui.color_picker` which cause the tests to fail


evnchn:
Did some history digging. Seems like we discussed whether to remove `!important` in #371 #1821 #4415. Back then, we hesistated before but now with 3.0 and the `@layer` stuff, our hands are rather forced and we _have_ to remove it otherwise I can't see how it works. 

I'm fine with it since we already checked for the documentation site at least that it doesn't break things (at least not too hard), but not sure about other users though. Feels more of a "heavy decision" after realize this has been discussed 3 times in the past, y'know? 

Maybe 3.0.0rc2 is needed after all.

falkoschindler:
@evnchn Thanks for the research. A 3.0.0rc2 sounds reasonable before rushing it.

One test is failing by the way: The color picker in `ui.color_input` has some margin now which - by sheer luck - causes the test to fail. I suspect there are more of these subtle differences. 😕

evnchn:
Hmm @falkoschindler. I know that it doesn't break much of NiceGUI documentation, but I checked and that is because those mostly use Tailwind CSS for styling. 

Check this out: I am removing `!important` from Quasar's documentation. 

{4623DCD8-1182-4819-86C0-640C5A2D2738}

We see that the text doesn't show up since Quasar's `.q-dark` is broken. 

I don't think it will be a stable approach if we made Quasar less-important. Rather, we can only make Tailwind more important. What do you think?

evnchn:
Original is like this, by the way:

{073FBD57-EF5B-4A94-AD08-21CEF191C160}

falkoschindler:
Yes, I agree that this is probably a dead end. Even if the documentation turns out identically, we won't be able to guarantee that every Quasar feature works like before, and we'd risk major regressions in the future whenever we update Quasar.

PR #5170, opened 2025-09-24
Move Tailwind's important rules to a layer before Quasar to make sure they win

### LINKED ISSUE #5156: In NiceGUI 3.0.0rc1,I can't apply TailWindCSS color forcely

evnchn:
### Motivation

We did try #5166, that didn't work (NiceGUI-native documentation site is broken), while #5167 also didn't work (Quasar-native documentation site shall be broken)

**Real motivation 3: I look like a 🤡 in #5166**

### Implementation

- `tailwind_importants` layer before all layers, normally empty and does not affect page operation
- `moveRules` to be ran, if we detect any changes in `<head>`
- `moveRules` to move all `!important` rules from Tailwind away from `utilities` into `tailwind_importants` so that accounting for the inverse of layer priority, Tailwind wins. 

### Progress

  - I'm considering it. 
  - May need document this to keep users in-the-loop of what is happening

### Comprehensive test script

Combines https://github.com/zauberzeug/nicegui/issues/5156#issuecomment-3324250564 and https://github.com/zauberzeug/nicegui/pull/5167#issue-3445876247

```py
@ui.page('/')
def index():
    ui.label('Tailwind & Quasar CSS Comprehensive Test Script').classes('text-2xl')

    ui.label('I want a red button').classes('text-lg')

    ui.button('button').classes('bg-red-700')  # doesn't work in 2.x and 3.0.0rc1 and this PR
    ui.button('button').classes('!bg-red-700')  # doesn't work in 3.0.0rc1, but work in 2.x and this PR

    ui.label('I want "Hi!" to be visible only in dark mode').classes('text-lg')

    dark = ui.dark_mode()
    ui.switch().bind_value(dark, 'value')
    ui.label('Using invisible dark:visible')
    ui.label('Hi!').classes('invisible dark:visible')  # doesn't work in 2.x and 3.0.0rc1 and this PR
    ui.label('Using invisible dark:!visible')
    ui.label('Hi!').classes('invisible dark:!visible')  # doesn't work in 3.0.0rc1, but work in 2.x and this PR

    ui.label('If it is green, Tailwind is winning over Quasar').classes('text-lg')

    ui.label('Quasar Red, Tailwind (no important) Green')
    ui.button('button').props('color="red-10"').classes('bg-green-700')  # never wins

    ui.label('Quasar Red, Tailwind (with important) Green')
    ui.button('button').props('color="red-10"').classes('!bg-green-700')  # wins in 2.x and this PR, but not in 3.0.0rc1
```

### Results

image

### Documentation page screenshot

image

### Additional notes

The value of UnoCSS (#4832) is that we can potentially do the same without the MutationObserver.

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -429,60 +429,48 @@ function createApp(elements, options) {
   }));
 }
 
-// HACK: remove Quasar's rules for divs in QCard (#2265, #2301)
-for (const importRule of document.styleSheets[0].cssRules) {
+const t0 = performance.now();
+const NEUTRALIZE_CLASSES = [
+  "cursor-not-allowed",
+  "cursor-pointer",
+  "block",
+  "inline-block",
+  "hidden",
+  "invisible",
+  "overflow-auto",
+  "overflow-hidden",
+];
+const CLS_ALT = NEUTRALIZE_CLASSES.map((s) => s.replace(/[-/\\^$*+?.()|[\]{}]/g, "\\$&")).join("|");
+const NEUTRALIZE_RE = new RegExp(`(^|[^\\w-])\\.(${CLS_ALT})(?![\\w-])`, "g");
+
+const rootSheet = document.styleSheets[0];
+const rules = Array.from(rootSheet.cssRules);
+const source = rules.find((r) => r.layerName === "quasar").styleSheet;
+const target = rules.find((r) => r.name === "quasar_importants");
+rootSheet.ownerNode.disabled = true;
+
+for (const importRule of rootSheet.cssRules) {
   if (importRule instanceof CSSImportRule && /quasar/.test(importRule.styleSheet?.href)) {
     for (const rule of Array.from(importRule.styleSheet.cssRules)) {
-      if (rule instanceof CSSStyleRule && /\.q-card > div/.test(rule.selectorText)) {
+      if (rule instanceof CSSStyleRule) {
+        // HACK: remove Quasar's rules for divs in QCard (#2265, #2301)
         if (/\.q-card > div/.test(rule.selectorText)) rule.selectorText = ".nicegui-card-tight" + rule.selectorText;
+
+        // HACK: neutralize Quasar's utility classes without affecting Tailwind
+        rule.selectorText = rule.selectorText.replace(NEUTRALIZE_RE, "$1._quasar_$2_");
       }
     }
   }
 }
 
-// HACK: Move Tailwind's important rules to a layer before Quasar to make sure they win
-function moveRules() {
-  const rulesInLayer = [];
-  for (const stylesheet of document.styleSheets) {
-    try {
-      for (const rule of stylesheet.cssRules) {
-        if (rule instanceof CSSLayerBlockRule && rule.name === "utilities") {
-          for (const rule2 of rule.cssRules) {
-            if (typeof rule2.selectorText === "string") {
-              const idx = rule2.cssText.indexOf("!important");
-              if (idx !== -1 && rule2.cssText.charAt(idx - 1) !== "\\") {
-                rulesInLayer.push(rule2);
-              }
-            }
-          }
-        }
-      }
-    } catch (e) {
-      console.warn(`Could not access stylesheet: ${stylesheet.href}`, e);
-    }
-  }
-  console.log(`CSS rules in layer "utilities":`, rulesInLayer);
-  for (const stylesheet of document.styleSheets) {
-    try {
-      for (const rule of stylesheet.cssRules) {
-        if (rule instanceof CSSLayerBlockRule && rule.name === "tailwind_importants") {
-          while (rule.cssRules.length > 0) {
-            rule.deleteRule(0);
-          }
-          for (const applyRules of rulesInLayer) {
-            console.log(`Re-inserting rule in layer "tailwind_importants":`, applyRules.cssText);
-            rule.insertRule(applyRules.cssText);
-          }
-          console.log(rule);
-          console.log(rule.cssText);
-        }
-      }
-    } catch (e) {
-      console.warn(`Could not access stylesheet: ${stylesheet.href}`, e);
-    }
+// HACK: Move Quasar's important rules to a layer to control precedence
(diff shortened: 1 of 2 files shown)


falkoschindler:
Interesting approach, @evnchn!

It feels a bit sketchy to tear apart Tailwind CSS rules, but it's certainly better then what we tried before.

I'm a bit concerned about the performance implications of observing and re-processing the document head. That's why I tried moving Quasar rules instead (they don't change). It took me quite a while, but this seems to do the job:
```css
@layer theme, base, quasar, nicegui, components, utilities, overrides, quasar_importants;
...
@layer quasar_importants {
  /* Do not remove this layer, it is used in the JS to move !important rules from the "quasar" layer here */
}
```
```js
const rules = Array.from(document.styleSheets[0].cssRules);
const source = rules.find((r) => r.layerName === "quasar").styleSheet;
const target = rules.find((r) => r.name === "quasar_importants");
for (let i = source.cssRules.length - 1; i >= 0; i--) {
  const rule = source.cssRules[i];
  if (rule instanceof CSSStyleRule && /!important/.test(rule.cssText)) {
    source.deleteRule(i);
    target.insertRule(rule.cssText);
  }
}
```
It just bugs me that this snippet randomly jumps between taking 4ms and 800ms. Still investigating...

falkoschindler:
Here we go:
```js
const rootSheet = document.styleSheets[0];
const rules = Array.from(rootSheet.cssRules);
const source = rules.find((r) => r.layerName === "quasar").styleSheet;
const target = rules.find((r) => r.name === "quasar_importants");
rootSheet.ownerNode.disabled = true;
for (let i = source.cssRules.length - 1; i >= 0; i--) {
  const rule = source.cssRules[i];
  if (rule instanceof CSSStyleRule && /!important/.test(rule.cssText)) {
    source.deleteRule(i);
    target.insertRule(rule.cssText);
  }
}
rootSheet.ownerNode.disabled = false;
```
Now it always takes around 5ms. The trick is to avoid re-rendering while working on the CSS sheet.

falkoschindler:
Here is a performance-optimized version for removing all conflicting Quasar classes (we should do this only if `ui.run(tailwind=True)` (the default):
```js
const NEUTRALIZE_CLASSES = [
  "cursor-not-allowed",
  "cursor-pointer",
  "block",
  "inline-block",
  "hidden",
  "invisible",
  "overflow-auto",
  "overflow-hidden",
];
const CLS_ALT = NEUTRALIZE_CLASSES.map((s) => s.replace(/[-/\\^$*+?.()|[\]{}]/g, "\\$&")).join("|");
const NEUTRALIZE_RE = new RegExp(`(^|[^\\w-])\\.(${CLS_ALT})(?![\\w-])`, "g");
const sheet = document.styleSheets[0];
sheet.ownerNode.disabled = true;
for (const importRule of sheet.cssRules) {
  if (importRule instanceof CSSImportRule && /quasar/.test(importRule.styleSheet?.href)) {
    for (const rule of Array.from(importRule.styleSheet.cssRules)) {
      if (rule instanceof CSSStyleRule) {
        rule.selectorText = rule.selectorText.replace(NEUTRALIZE_RE, "$1._quasar_$2_");
      }
    }
  }
}
sheet.ownerNode.disabled = false;
```

evnchn:
@falkoschindler Awesome idea to move Quasar's importants to a less-important layer than to move Tailwind's importants to a more-important layer (considering importance after `!important` order reversal), since Quasar styles do not change (in general, I think? I can't exclude possibility of Quasar setting style via JS but if they do I will be **mad**)

But why do it in the browser? Can't we have:

```
      {% if prod_js %}
        @import url("{{ prefix | safe }}/_nicegui/{{version}}/static/quasar.prod.css") layer(quasar);
        @import url("{{ prefix | safe }}/_nicegui/{{version}}/static/quasar.importants.prod.css") layer(quasar_importants);
      {% else %}
        @import url("{{ prefix | safe }}/_nicegui/{{version}}/static/quasar.css") layer(quasar);
        @import url("{{ prefix | safe }}/_nicegui/{{version}}/static/quasar.importants.css") layer(quasar_importants);
      {% endif %}
```
 
Assuming you have created the `quasar.importants*.css` files somewhere else.

falkoschindler:
@evnchn Good idea! It was easier to do it in JS in the browser, but it should be possible as a preprocessing step, ideally in extract_core_libraries.py.

evnchn:
Heck, I even think that the segregation of important and no-important styles can be officially in Quasar, offering advanced technical people (like us, integrating Quasar into NiceGUI) the flexibility, while everybody else can just import 2 stylesheets / 1 combined stylesheet and get along with their normal usage. 

Can propose later if it works here, I guess

falkoschindler:
Not sure if they like it, but worth a try.

Two caveats with the current implementation:
- The script moves whole rules like `.q-icon::after, .q-icon::before { width: 100%; height: 100%; align-items: center; justify-content: center; display: flex !important; }`, including non-important declarations. We might want to extract the important ones and leave the non-important ones in place.
- The script only handles instances of `CSSStyleRule` and skips rules like
    ```css
    @media (orientation: landscape) {
      .orientation-portrait { display: none !important; }
    }
    ```

evnchn:
image

Works

evnchn:
Wait but does this capture and move this?  

image

evnchn:
~~This fails both of our implementations~~ I fixed it on mine 🤪

```py
from nicegui import ui


@ui.page('/')
def page():
    ui.label('This text is always visible')
    # Quasar: print-only is hidden on screen, visible on print
    ui.label('print-only').classes('print-only')

    # Test if user's custom important CSS can override Quasar's print-only class
    ui.add_css('''
    @layer utilities {
        @media screen {
            .print-only {
                display: block !important;
            }
        }
    }''')


ui.run(show=False, port=9999)
```

falkoschindler:
Let's close this PR in favor of PR #5171.

PR #5171, opened 2025-09-24
Segregate Quasar's important rules in Python-side

### LINKED ISSUE #5156: In NiceGUI 3.0.0rc1,I can't apply TailWindCSS color forcely

evnchn:
### Motivation

We did try #5166, that didn't work (NiceGUI-native documentation site is broken), while #5167 also didn't work (Quasar-native documentation site shall be broken). #5170 we're almost there but as said in https://github.com/zauberzeug/nicegui/pull/5170#issuecomment-3328192605 we should instead segregate the Quasar styles in the server side Python pre-comute. 

**Real motivation 3: I look like a 🤡 in #5166**

**Real motivation 4: #5170 is not adopted due to frequent updates on head**

### Implementation

`quasar_segregator.py` creates the 3 CSS files. 

### Progress

  - [x] Separate the mixed ones. I've been stuck for a while
  - [x] Modify the `index.html` to serve these segregated styles
  - [x] Minify with `rcssmin`
  - [x] Beautify with `cssbeautifier`
  - [x] Document pip install commands
  - Again I'm considering it. 
  - We should probably also document this somewhere just for avoid confusion down the line.
- [x] fix comments being moved to wrong code block
- [x] integrate with extract_core_libraries.py
- [x] add new dev dependencies to pyproject.toml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,3 @@ venv
 .DS_Store
 .env
 node_modules/
-quasar.reference.css
-quasar.reference.prod.css
--- a/extract_core_libraries.py
+++ b/extract_core_libraries.py
@@ -11,7 +11,9 @@ import rcssmin
 import tinycss2
 from tinycss2 import ast
 
-from library_path_constants import NODE_MODULES, STATIC
+ROOT = Path(__file__).parent
+STATIC = ROOT / 'nicegui' / 'static'
+NODE_MODULES = ROOT / 'node_modules'
 
 
 def _extract_quasar_css(css_path: Path) -> None:
--- a/library_path_constants.py
+++ b/library_path_constants.py
@@ -1,5 +1,0 @@
-from pathlib import Path
-
-ROOT = Path(__file__).parent
-STATIC = ROOT / 'nicegui' / 'static'
-NODE_MODULES = ROOT / 'node_modules'


evnchn:
Obligatory screenshot (test script at #5170)

image

Before you ask, the documentation screenshot:

image

quasar_segregator.py
```diff
@@ -0,0 +1,55 @@
...
+            has_important = True
+        if isinstance(declaration, ast.Declaration) and not declaration.important:
+            has_unimportant = True
+    if has_important and has_unimportant:
+        rule_copy1 = copy.deepcopy(rule)
+        rule_copy2 = copy.deepcopy(rule)
+        rule_copy1.content = [d for d in declarations if isinstance(
+            d, ast.Declaration) and d.important]  # keep only important
+        rule_copy2.content = [d for d in declarations if isinstance(
+            d, ast.Declaration) and not d.important]  # keep only unimportant
```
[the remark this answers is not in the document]
evnchn: `d if not isinstance(d, ast.Declaration) else d` 🤡 

Done in https://github.com/zauberzeug/nicegui/pull/5171/commits/665ec52b050ff1678a65cb1d3aaaedbada94bf17

falkoschindler:
Thanks, @evnchn! While we were debating whether it's worth trying, you already did. Great work!

Our main concern, however, is the complexity and number of dependencies we're adding. Parsing and re-arranging CSS isn't trivial and could lead to nasty bugs. cssbeautify, rcssmin and tinycss2 are not too popular, might get inactive in the future (looking at you, vbuild) and could introduce bugs in edge cases.

Regarding one detail which caught my eye, ChatGPT explained:
> `.q-icon:before` and `.q-icon:after` are valid selectors. They target the `::before` and `::after` pseudo-elements of elements with the class `.q-icon`. Using a single colon (`:`) instead of double (`::`) is allowed for compatibility, though the modern standard is `::before` and `::after`.

Anyway, we would prefer leaving this PR maybe as an inspiration for 4.0 and follow option 1.2 (see https://github.com/zauberzeug/nicegui/issues/5156#issuecomment-3327983473).

evnchn:
@falkoschindler It has always been `.q-icon:before` in `quasar.css` even before this PR. I guess I will prefer follow them. 

For the dependencies, 2 of them are good news: 

- `cssbeautify` empowers https://beautifier.io/ and I use that quite often from time to time. Quite popular JS beautifier so should be fine? 
- `tinycss2` is "based on the [CSS Syntax Level 3](https://www.w3.org/TR/css-syntax-3/) specification, knows the grammar of CSS but doesn’t know specific rules, properties or values supported in various CSS modules" so I don't see how it can break.

The one that I have concern is `rcssmin ` since it is a "[re-implementation](https://pypi.org/project/rcssmin/#:~:text=module%20is%20a-,re%2Dimplementation,-aiming%20for%20speed)"

We could do https://pypi.org/project/lightningcss/ which provides bindings to the lightningcss library written in rust, but as you know I'm on PyPy right now and I can't install to evaluate at the moment ...

---

I'm fine with both solutions but I feel, genuinely, that option 1.2 (represented by #5170 now, I think) is more "sketchy" and more "hacky". This options feels more "we're making our own CSS post-processor" and more "legit" somehow.

evnchn:
This PR got some issues: Doesn't catch all `!important` and some remain in the unimportant.css

image

evnchn:
Example code which shows that this PR fails: 

https://github.com/zauberzeug/nicegui/pull/5170#issuecomment-3329276410

falkoschindler:
Oh, with option 1.2 I meant none of our existing PRs, but:

- Add a demo about how to use custom text and background colors for buttons.
- String-replace conflicting "important" Quasar classes, e.g. from ".invisible {...}" to ".q-invisible {...}" (or remove them, but string-replacing their class names is easier).
- While we're at it, string-replace the card classes to move the corresponding hack from JS to Python.

That's it, no sketchy CSS stunts, but still solving 99% of our problems. (🤞🏻)

evnchn:
> That's it, no sketchy CSS stunts, but still solving 99% of our problems. (🤞🏻)

Bold claim. Do show me the code. We'll see which one is less sketchy and/or solves more problems (not sure the judgement criteria to be honest but we'll see)

---

Anyways if you recall this PR may not just be for NiceGUI but also for Quasar as well so I still want to push for completion, so check out https://github.com/zauberzeug/nicegui/pull/5171/commits/31d3f9078d51858e1d08b08138a9d871982f10e2

Using the code above, we see that user can override Quasar's `print-only`

image

evnchn:
> String-replace conflicting "important" Quasar classes, e.g. from ".invisible {...}" to ".q-invisible {...}" (or remove them, but string-replacing their class names is easier).

So you're saying, after the maneuver, Quasar's `.invisible` rules will cease to exist, and then any Quasar elements which uses `.invisible` will use Tailwind's rules? I think this is pretty "sketchy", considering the case where Tailwind and Quasar do not agree, which may happen in long-term. 

---

Suppose that Tailwind got hacked and they changed every class to simply apply `color: red; background-color: red`. Wouldn't this element show up red under your proposed implementation, but not this one? 

`ui.button('Invisible button').classes('invisible')`

---

I'm exaggerating since in reality Tailwind will not get hacked. What's likely the case is: 

- Either there are already slight differences between Tailwind's definition and Quasar's definition for the common classes and we didn't catch them
- Or what's likely the case, Tailwind can change their implementation if they want. 

If that's the case, we must: 

- Keep track of the classes which are common between Tailwind and Quasar
- Check to see if their implementation is the same
  - If same, fine. 
  - If not we will keep one:
    - Keep Tailwind's definition, by delete Quasar's
    - Keep Quasar's definition which is default more important than Tailwind's

falkoschindler:
Ok, you're starting to convince me...

I still think that conflicting classes like "invisible" have a trivial CSS (like `display: invisible`) which should be equal among frameworks. But technically there _could_ be a mismatch. And yes, Quasar elements could depend on the "invisible" class being treated as important.

So let's continue the path of this PR. I just checked the generated CSS and added some beautifier options to reduce the diff. What I then noticed is that comments strangely moved down one block. It becomes visible when viewing the diff between quasar.css and quasar.unimportant.css:
Screenshot 2025-09-25 at 14 19 57
This could get confusing when inspecting CSS in debug mode.

Reducing the diff is one way to verify that the script is doing what we expect. There are some other diffs I don't know how to get rid of though. But that's ok.

I also moved `quasar_<...>.css` to `quasar.<...>.css` so that .gitattributes and .pre-commit-config.yaml still match.
And I added some more tasks to your checklist (see above).

quasar_segregator.py
```diff
@@ -0,0 +1,110 @@
...
+        important_split_rules.append(important_rule_copy)
+        unimportant_split_rules.append(unimportant_rule_copy)
+        important_split_rules.extend(pending_comments)
+        unimportant_split_rules.extend(pending_comments)
+    elif contains_important_declarations:
+        important_split_rules.append(qualified_rule)
+        important_split_rules.extend(pending_comments)
+    else:
+        unimportant_split_rules.append(qualified_rule)
+        unimportant_split_rules.extend(pending_comments)
```
evnchn: NEEDFIX: comments first, then the rules. Then we won't have the comments being 1 block lower

evnchn (commented):
🤡

evnchn:
> added some beautifier options to reduce the diff

It is unlikely that there exists a config such that cssbeautifier works **identically** to whatever beautifier that Quasar uses. And maybe they change their formatter one day...

If you want to reduce the diff, consider formatting `quasar.css` into `quasar.reformat.css` and have the segregator base off the latter.

quasar_segregator.py
evnchn: https://github.com/zauberzeug/nicegui/pull/5171/commits/9cf608080a17079a4a129bc87de4b62c24a41f71

evnchn:
@falkoschindler Literally only line-removals are found from `quasar.reference.css` to `quasar.unimportant.css`

Looks pretty good so far. May want to add a check. 

`quasar.reference.css` will never look the same as `quasar.css` since we `skip_whitespace=True` but the CSS should still be same-functioned, otherwise `tinycss2` issue...

evnchn:
image

Docs shoved here. Intentional not to create a new subsection in case in future NiceGUI versions we backtrack on this.

falkoschindler:
I just reviewed and refactored the code quite a bit:

- Most importantly I simply dropped the comment nodes. They don't seem very informative to me and just complicate the segregation script.
- The second most complicated part (in my point of view) was the verification at the end, especially steps 4 and 5 to check if important and unimportant rules interleave. I figured we can achieve _almost_ the same result by 1) using sequence matchers to verify that the parts don't contain any new lines (only delete and equal operations, no inserts or replacements) and 2) comparing the number of declarations (sufficiently approximated by the number of semicolons). This simplifies the code dramatically and confirms the correctness of the implementation sufficiently.

With these changes I hope to improve maintainability, especially if we re-visit the code in one year.

falkoschindler:
And now that the segregation code is so short, I moved it into extract_core_libraries.py. This isn't perfect for the readability of this file, it doesn't really make sense as a standalone script. It should immediately generate two parts and skip the original quasar.css file. And it avoids cluttering the root directory with too many scripts.

evnchn:
I am impressed by the refactoring! This looks more ready than ever. 

Couple of notes useful for one year later: 

> sufficiently approximated by the number of semicolons

1. If I recall correctly there can be `\;` to target, for example, class names with semicolons (?) So the check may break one day. 

2. You're using `difflib`. I don't know how it works but there may be a chance where it mis-detects removals as removal+addition, similar to Git diff.

3. As such, **no amount of checks will be a substitute for manual verification**. 

4. Working backwards, we can also diff the CSS of Quasar from now to the release one year afterwards, note the changes, and ensure applied in our segregated CSS as well.

falkoschindler (approved):
Ok, I think this PR is ready to merge.

@evnchn I agree that nothing beats manual verification. With the current safety net we at least should notice if something is wrong, even if there could potentially be occur some false positives.

PR #5172, opened 2025-09-24
Deprecate `ClientConnectionTimeout` and `timeout` for `await client.connected()`

evnchn:
### Motivation

In #5116 `ClientConnectionTimeout` was an exception raised (and NiceGUI internally handled) that would occur when `await client.connected()` timed out. 

But as #5116 progressed we made `await client.connected()` never timeout unless you pass in a timeout parameter, and so if we just got rid of that parameter we can get rid of `ClientConnectionTimeout` entirely. 

I was trying to slip this into #5116 but (1) we need to feature-freeze 3.0 and (2) indeed there may be uses for `ClientConnectionTimeout` and so we can deprecate the timeout parameter in 3.x and remove it in 4.0? 

### Implementation

- See `ClientConnectionTimeout`? Remove it
- Timer can always start when the client is eventually connected; Assuming entire client will be destroyed alongside its timer. 
- Do not unsubscribe the callback on client disconnect; Assuming the entire client will be destroyed alongside its callback. 
- Page late return value will not delete client; Client is destroyed by the purge anyways. 

I made 2 assumptions there, so this code may or may not work!

### Progress

@falkoschindler Milestone 4.0?

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -41,12 +41,6 @@ HTML_ESCAPE_TABLE = str.maketrans({
 })
 
 
-class ClientConnectionTimeout(TimeoutError):
-    def __init__(self, client: Client) -> None:
-        super().__init__(f'ClientConnectionTimeout: {client.id}')
-        self.client = client
-
-
 class Client:
     page_routes: ClassVar[dict[Callable[..., Any], str]] = {}
     '''Maps page builders to their routes.'''
@@ -191,21 +185,13 @@ class Client:
         """Return the title of the page."""
         return self.page.resolve_title() if self.title is None else self.title
 
-    async def connected(self, timeout: float | None = None) -> None:
-        """Block execution until the client is connected.
-
-        :param timeout: timeout in seconds (default: ``None``)
-        """
+    async def connected(self) -> None:
+        """Block execution until the client is connected."""
         if self.has_socket_connection:
             return
         self._waiting_for_connection.set()
         self._connected.clear()
-        purpose = (self.request.headers.get('Sec-Purpose') or self.request.headers.get('Purpose') or '').lower()
-        is_prefetch = 'prefetch' in purpose and 'prerender' not in purpose
-        try:
-            await asyncio.wait_for(self._connected.wait(), timeout=None if is_prefetch else timeout)
-        except asyncio.TimeoutError as e:
-            raise ClientConnectionTimeout(self) from e
+        await asyncio.wait_for(self._connected.wait(), None)
 
     async def disconnected(self) -> None:
         """Block execution until the client disconnects."""
--- a/nicegui/elements/timer.py
+++ b/nicegui/elements/timer.py
@@ -1,8 +1,7 @@
 from contextlib import AbstractContextManager, nullcontext
 
-from ..client import Client, ClientConnectionTimeout
+from ..client import Client
 from ..element import Element
-from ..logging import log
 from ..timer import Timer as BaseTimer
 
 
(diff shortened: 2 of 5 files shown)


falkoschindler:
Nice to see how much code would be gone.
But my main concern is that some user code relies on a configurable connection timeout and would be stranded without it.

evnchn:
Users will never be fully stranded, I think?

Consider this:

```py
try:
    await asyncio.wait_for(ui.context.client.connected(), timeout=5)
except asyncio.TimeoutError:
    pass # give up waiting after 5 seconds
```

I guess the question is, do we want 31 extra lines in the codebase to cut user's 3 lines of boilerplate?

falkoschindler:
Then my question would be: Can _we_ replace our 31 lines with the 3 lines using `asyncio.wait_for`?

evnchn:
Retrospect, because that 3 lines of code is more like "OK it timed out, deal with it" but our current code logic is more like "OK it timed out, I'm going to delete the client for you and the timers and the event handlers and stuff"

So actually maybe I have to backtrack by saying if we remove this code there's no real ways around it on the user side. 🧐

rodja:
I like the idea of deprecating the timeout parameter a lot. Currently we assume that there is user code relying on it. But I'm i think it is unlikely. If we deprecate now and let it run until 4.0 these potential users can reach out to us and explain their scenario. Worst case: we remove the deprecation in 3.x.

falkoschindler:
From my point of view there are three options:

| Option | Description                        | Benefits                                                               |
| ------ | ---------------------------------- | ---------------------------------------------------------------------- |
| 1      | Leave everything as is.            | User can control the timeout. Client pruned immediately after timeout. |
| 2      | Use `asyncio.wait_for` internally. | User can control the timeout. Less complex code.                       |
| 3      | Remove the timeout altogether.     | Less complex code.                                                     |

This PR implements option 3 and would need to wait for 4.0. We'd need another PR to deprecate the parameter in 3.x.

But given my humble table this option "only" reduces code complexity, while removing a feature without improving performance or memory usage.

falkoschindler:
Oh sorry, @evnchn! This PR got closed by removing the target 3.0 branch and now I can't reopen it.
But maybe we should
1. first come to a conclusion if we really want to remove this timeout parameter (maybe with feedback from the community),
2. then deprecate the parameter, and
3. finally create a PR to remove it in 4.0.

Since it's probably at least a year until 4.0, doing step 3 first might be impractical since the code could change quite a lot until then.

evnchn:

**4.0 is rolling around — can we get this one to a decision?** @falkoschindler @rodja

The ordering here was conclude → deprecate → remove, and doing anything was blocked on *"probably at least a year until 4.0"*. It's milestoned for November now, so that objection has mostly expired — but step 1 was never actually concluded, so there's still nothing to deprecate.

Flagging against my own PR while we're here: I backtracked in-thread and think it's load-bearing — `asyncio.wait_for` abandons the wait but skips the client cleanup, so it isn't the clean user-side substitute I originally pitched. Option 3 is weaker than I made it sound.

Option 1, 2, or 3?

<details><summary><b>The quotes, for context</b></summary>

@falkoschindler, 2025-10-03 — the ordering and the blocker:

> maybe we should 1. first come to a conclusion if we really want to remove this timeout parameter (maybe with feedback from the community), 2. then deprecate the parameter, and 3. finally create a PR to remove it in 4.0. Since it's probably at least a year until 4.0, doing step 3 first might be impractical since the code could change quite a lot until then.

@falkoschindler, 2025-09-29 — the case against option 3:

> given my humble table this option "only" reduces code complexity, while removing a feature without improving performance or memory usage.

Me, 2025-09-24 — the backtrack:

> Retrospect, because that 3 lines of code is more like "OK it timed out, deal with it" but our current code logic is more like "OK it timed out, I'm going to delete the client for you and the timers and the event handlers and stuff" […] So actually maybe I have to backtrack by saying if we remove this code there's no real ways around it on the user side 🧐

@rodja, 2025-09-25 — why deprecating is cheap even if we keep it:

> If we deprecate now and let it run until 4.0 these potential users can reach out to us and explain their scenario. Worst case: we remove the deprecation in 3.x.

</details>

falkoschindler:
Thanks for pushing for a conclusion, @evnchn! Let's conclude: **Option 1 — keep the `timeout` parameter.**

Looking at it once more, the only benefit of removal is ~30 fewer lines of code. There's no performance or memory gain, and no correctness gain either: clients that never connect are pruned after 60 seconds anyway, so the exception handlers merely speed up the cleanup. On the other side of the ledger we'd have a deprecation warning nagging 3.x users, a breaking change in 4.0, and — as you noticed yourself — no clean user-side replacement, since a plain `asyncio.wait_for` skips the immediate cleanup. Removing a small but real user-facing feature to save 30 internal lines just isn't a good trade.

So my decision: keep everything as is, no deprecation for 4.0. Sorry it took a poke to get there. 🙂

PR #5173, opened 2025-09-25
Call lifecycle event handlers more consistently

falkoschindler:
### Motivation

In #4285 we noticed that connect and disconnect handlers are not both called when reconnecting. To avoid breaking changes, we left it unchanged until now. With version 3.0 we now have the chance to unify how lifecycle event handlers are called.

### Implementation

This PR introduces new event handlers to complete the following list:

- `on_handshake`: when the first socket connection is established
- `on_connect`: when the socket connection is established (even during reconnect)
- `on_disconnect`: when the socket connection is interrupted (even during reconnect)
- `on_deleted`: when a client is deleted after it didn't reconnect for a while

Note that the (internal) `client.handle_handshake` method is called on every reconnect, but the handshake even handler is only called at the first connection.

I also added a new pytest, but didn't manage to simulate connection losses.

### Progress

- [x] Pytests have been added (although without maximum test coverage).
- [ ] decision about event names, especially "on_handshake"

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -44,8 +44,10 @@ class App(FastAPI):
 
         self._startup_handlers: list[Union[Callable[..., Any], Awaitable]] = []
         self._shutdown_handlers: list[Union[Callable[..., Any], Awaitable]] = []
+        self._handshake_handlers: list[Union[Callable[..., Any], Awaitable]] = []
         self._connect_handlers: list[Union[Callable[..., Any], Awaitable]] = []
         self._disconnect_handlers: list[Union[Callable[..., Any], Awaitable]] = []
+        self._deletion_handlers: list[Union[Callable[..., Any], Awaitable]] = []
         self._exception_handlers: list[Callable[..., Any]] = [log.exception]
         self._page_exception_handler: Optional[Callable[..., Any]] = None
 
@@ -107,6 +109,13 @@ class App(FastAPI):
         except Exception as e:
             self.handle_exception(e)
 
+    def on_handshake(self, handler: Union[Callable, Awaitable]) -> None:
+        """Called when a client completes the handshake with NiceGUI.
+
+        The callback has an optional parameter of `nicegui.Client`.
+        """
+        self._handshake_handlers.append(handler)
+
     def on_connect(self, handler: Union[Callable, Awaitable]) -> None:
         """Called every time a new client connects to NiceGUI.
 
@@ -120,6 +129,13 @@ class App(FastAPI):
         The callback has an optional parameter of `nicegui.Client`.
         """
         self._disconnect_handlers.append(handler)
+
+    def on_deletion(self, handler: Union[Callable, Awaitable]) -> None:
+        """Called when a client is deleted.
+
+        The callback has an optional parameter of `nicegui.Client`.
+        """
+        self._deletion_handlers.append(handler)
 
     def on_startup(self, handler: Union[Callable, Awaitable]) -> None:
         """Called when NiceGUI is started or restarted.
@@ -306,8 +322,10 @@ class App(FastAPI):
         self.storage.clear()
         self._startup_handlers.clear()
         self._shutdown_handlers.clear()
+        self._handshake_handlers.clear()
         self._connect_handlers.clear()
         self._disconnect_handlers.clear()
+        self._deletion_handlers.clear()
         self._exception_handlers[:] = [log.exception]
         self.config = AppConfig()
 
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -44,10 +44,9 @@ class App(FastAPI):
 
         self._startup_handlers: list[Union[Callable[..., Any], Awaitable]] = []
         self._shutdown_handlers: list[Union[Callable[..., Any], Awaitable]] = []
-        self._handshake_handlers: list[Union[Callable[..., Any], Awaitable]] = []
         self._connect_handlers: list[Union[Callable[..., Any], Awaitable]] = []
         self._disconnect_handlers: list[Union[Callable[..., Any], Awaitable]] = []
-        self._deletion_handlers: list[Union[Callable[..., Any], Awaitable]] = []
+        self._delete_handlers: list[Union[Callable[..., Any], Awaitable]] = []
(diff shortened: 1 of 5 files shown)


evnchn:
### Lifetime diagram

```
->CLIENT| (on HTTP)
        ->SOCKET|
        |SOCKET<-
           ...  (as many times within timeout)
        ->SOCKET|
        |SOCKET<-
|CLIENT<- (timed out)

Y-axis: time
X-axis legends:

| Python client missing, let alone socket link
        | Python client ready, missing socket link
                | Python client and socket link ready
````

### Event comparison table:

| Event                               | Old Handler               | New Handler                           |
|-------------------------------------|---------------------------|---------------------------------------|
| Python: `Client` created            | `@ui.page`                | `@ui.page`                            |
| Network: Socket.IO connected         | `on_connect`              | `on_connect`, _`on_handshake` if first time (I don't like this)_ |
| Network: Socket.IO disconnected      | None                       | `on_disconnect`                       |
| Python: `Client` deleted             | `on_disconnect`           | `on_delete`                           |

@falkoschindler Kindly check if this table is accurate. Will be in the docs / migration notes eventually. 

---

Before this PR:
- it is completely impossible to detect Socket.IO disconnected. 
- `connect` and `disconnect` handlers are assymetrical, so to speak. 

About this PR: 
- Good to fix both of the above issues
- Not too sure about `on_handshake` though, it is an odds-one-out in these pairs of functions. 
  - Especially since a handshake (Socket.IO one) occurs during re-connect. 
  - I don't want `on_first_handshake` nor `on_first_handshake`

Suggestions: 

- Can we just let the `on_connect` (and maybe `on_disconnect`) read that it is the `n`-th connection/disconnections?
  - No more confusing `on_handshake`
  - `n`, stored somewhere, increases with connect but does not decrease with disconnect
    - 1st connect; 1st disconnect; 2nd connect; 2nd disconnect...
  - Then we maybe pass that as a parameter into `on_connect` and `on_connect`? 
    - This also has an inadvertant advantage of making `on_disconnect` non-backwards-compatible, which in this case can be a good thing since the function's changed. 
  - Or we let users read it themselves?
    - They must read it immediately or else there are race-condition since the handlers can be ran in async. 
- Can we rework or deprecate `_num_connections`? 
  - It's 3.0, no more shared client, why more than 1 connect?
  - Become `_is_connected` with boolean values, or clamp `_num_connections` between 0 and 1?
  - Block new connection attempts if the old one did not terminate / terminate the old connection attempt upon the new one arrival?

evnchn:
Not that easy to deprecate `_num_connections`, as seen in #5174

falkoschindler:
@evnchn Yes, the table looks correct.

**`_num_connections`**: I don't think we can remove it easily. It has not only been relevant for the old auto-index client, but for "normal" clients as well. We noticed that sometimes a reconnect happens in reverse order, so we need to count a second "connect" before receiving a "disconnect". Therefore I don't think we can remove the counter (without too much effort and risk). See #4285 for more details.

**`on_handshake` and `on_connect`**: I agree that the terminology isn't optimal, especially given the fact that "handshakes" also take place during reconnection. What about the following event names:

- connect
- interrupt
- reconnect
- disconnect ("delete" would be more correct, but then "connect" had no obvious counterpart)

or:

- establish/created/started (I can't find a really good name...)
- connect
- disconnect
- deleted

falkoschindler:
More naming ideas (order by personal preference):

1. on_established, on_connect, on_disconnect, on_expired
2. on_established, on_connect, on_disconnect, on_delete
3. on_open, on_connect, on_disconnect, on_close
4. on_activate, on_connect, on_disconnect, on_deactivate

`app.on_expired` sounds slightly more connection-related than `app.on_delete`.

evnchn:
@falkoschindler TL-DR: **Consider dropping the first proposed function** and simply do `on_connect` `on_disconnect` and `on_deleted`, and make `on_deleted` run on explicit `Client.delete`. Naming rationale: `on_connect` `on_disconnect` are the unbeatable duo, while `on_deleted` roughly mirrors `__del__` expected as a Python object. 

---

1st function: **Hesitant on the first function no matter the name**. 
  - Can't we just:
    - ```py
       # your first connect code here?
    - ```py
      def my_on_connect(client):
          if client.is_first_connect:
              ... # your first connect code here?
  - I think the purpose of an event handler is to capture events that is _outside_ of the `@ui.page` lifetime...
  - Nobody asked for it yet? So maybe we can implement one when people feature request?
  - If you insist, then I think `on_first_connect` would be the most truthful (but a mouthful), while `on_ready` is a decent short one. 
  - My naming criteria: 
    - An action which can be done once only (fails `on_open` since you can `close` it)
    - An action that doesn't sound like we are allocating memory in Python, but awaiting network (fails `on_established`)
  
2nd function `on_connect`: **It's golden**. It was there, it is here now, we're not moving it, and it matches the reality

3rd function `on_disconnect`: **It's basically golden**. Now we're moving it to the direct-opposite of `connect`, making a perfect duo. Sure, we can have 2 back-to-back `connect` since #5174 is not fully implemented, but you can consider 2 separate connections

4rd function `on_deleted`: **Slight issues.** I just realized that it is not ran, if I manually invoke `Client.delete()` (assume I am actively kicking people out of my server)
  - Maybe make `deleted` to also be ran under the above case?
  - If not we have to resort to `on_expired` but what's the use of such a handler?

falkoschindler:
You're right: We had no "first connect" event, nobody asked for it, and it is of little use since you can `await client.connected()` instead.

Regarding the final event, I'm undecided between `on_delete` which should always be triggered when calling `client.delete()`, or `on_expired` for clients that fail to reconnect. That latter might be useful for freeing up resources in user code. But when thinking about it, `on_delete` would serve the same purpose with the benefit of _always_ working even when called manually.

Rodja and I also liked the open/close combination. But it could be confused with opening/closing the _connection_. And if we don't need the first event, we can as well choose a better name for the last one.

This brings us to the combination **connect, disconnect, delete**. I'll adjust the implementation accordingly.

falkoschindler:
Yes, I really like how it turned out! Triggering the delete event when calling `delete()` finally allows to pytest all events rather efficiently (without waiting for expiration).
Thanks again for your valuable feedback, @evnchn!

@rodja I'll leave it up to you to approve and merge this PR.

rodja (approved):

PR #5174, opened 2025-09-25
Fully deprecate multi-connection logic and  `_num_connections`

evnchn:
### Motivation

In 3.0 we dropped the shared client, and so now each `Client`expected to talk with 1 browser-side client only. 

So then, why are we keeping account of `_num_connections`? Rather pointless as it ideally be 0 or 1 only (still not the case for 3.0, I'll get into this)

### Implementation

See `_num_connections`? Remove it. 

### Progress

  - [ ] Fully ensure 0 or 1 socket connection

### Further considerations

This is hinging on the assumption that connect and disconnect are called interweavingly (so no two connects in a row, no two disconnects in a row) so that essentially `_num_connections` would really be oscilate between 0 and 1

But actually I immediately see how this is not the case. When you switch from WiFi to mobile data, then before the old Socket.IO connection disconnects (since it takes some time from network down to Socket.IO giving uo), the new Socket.IO connection comes. 

There are 2 solutions: 

- Block the new connection until the old one disconnected 
  - Poor UX
- Terminate the old connection when the new one comes
  - No idea how to do this right now... Still thinking

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -74,7 +74,6 @@ class Client:
         self._deleted_event = asyncio.Event()
         self.environ: dict[str, Any] | None = None
         self.on_air = False
-        self._num_connections: defaultdict[str, int] = defaultdict(int)
         self._delete_tasks: dict[str, asyncio.Task] = {}
         self._deleted = False
         self._socket_to_document_id: dict[str, str] = {}
@@ -267,7 +266,6 @@ class Client:
         self._connected.set()
         self._socket_to_document_id[socket_id] = document_id
         self._cancel_delete_task(document_id)
-        self._num_connections[document_id] += 1
         if next_message_id is not None:
             self.outbox.try_rewind(next_message_id)
         storage.request_contextvar.set(self.request)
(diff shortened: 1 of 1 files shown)


falkoschindler:
> This is hinging on the assumption that connect and disconnect are called interweavingly [...] But actually I immediately see how this is not the case.

Exactly. This is why we ended up with a counter in #4285. It's a simple solution that does the job well.

evnchn:
> > This is hinging on the assumption that connect and disconnect are called interweavingly [...] But actually I immediately see how this is not the case.

> Exactly. This is why we ended up with a counter in https://github.com/zauberzeug/nicegui/pull/4285. It's a simple solution that does the job well.

I agree with the "exactly" and the "simple solution that does the job" but not really on the "well". 

I think if we kick the old connection offline, we can:

- simply the code and get rid of `_num_connections`
- further enforce one-to-one relationship between `Client` and browser-client, aligning with 3.0's design
- potentially save resources as we stop tring the old socket entirely
- `connect` and `disconnect` run alternatively, never two-in-a-row, which makes much more sense

---

I still want to pursue this PR, but if I can't make it into 3.0 then it has to go into 4.0, and I don't think it's practical nor safe to rush it. You can make the decision.

falkoschindler:
While removing the `_num_connections` is straight-forward, the crux is how to handle an early connect event (when connect and disconnect are triggered in reverse order during a reconnect). As long as we have no concept of how to do that, we need the connection counter.

I marked the PR as draft. We can assign a milestone later or simply close it.

falkoschindler:
Oh sorry, @evnchn! This PR got closed by removing the target 3.0 branch and now I can't reopen it.
But since we're still missing an idea for the core problem mentioned above, we might as well leave it closed and create a new PR once we know how to avoid multi-connections.

PR #5153, opened 2025-09-19
pytest: All warnings are to be treated as errors

evnchn:
### Motivation

We don't want to block #5145 from resolving #5142 but long-term speaking we still want warnings gone from our tests. So, we're beginning the journey of the fixes with this PR. 

I have some difficulty running tests locally and have the warnings show up, so GitHub Actions will be taking the load, I guess. 

### Implementation

Note that this is based off 3.0 sinc fixing all warnings will be impossible in 2.x due to python 3.8/3.9 and the existence of the shared auto-index client. 

NOTE: If you want to suggest changes, maintainers go ahead. anyone else you can PR into my branch in the fork, and I will review ASAP (most likely let you merge anyways)

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [ ] The implementation is complete.
- [ ] Pytests have been added (or are not necessary).
- [ ] Documentation has been added (or is not necessary).

PR #5182, opened 2025-09-28
Persisting on teardown fails on 3.9

rodja:
### Motivation

Working on #5178 uncovered the brittle implementation of the `background_tasks.await_on_shutdown` annotation done in #4641. Especially on Python 3.9.

### Implementation

This PR suggests to use a simple list of "task to be awaited" instead of the weak ref set. In addition the detection of the decorator is now implemented in a more robust and standard way via class wrapping.

### Progress


evnchn:
The value of this PR will shine in https://github.com/zauberzeug/nicegui/pull/5153 (warnings are not shown by default) so maybe we can base this off that branch and let the pipeline results speak for this PR's change?

rodja:
@evnchn it seems that I can only base the PR on branches of this repository and not your to one of the forks. Therefore I'll mirror your branch on this repo.

evnchn:
`persisting-on-teardown-fails-on-3.9` is still 3 commits ahead of `main` only, as such pipeline will likely still show green, but I expect it to show red with **only** `test_uploading_from_two_tabs` failing. 

Maybe we can branch off this branch then `git merge ci/warnings-as-errors` then check the pipeline result of _that_. 

But yeah for complex issues like "pytest warnings as errors" we're having way too many branches right now...

evnchn:
I can also do it on my side, just realized. We now have https://github.com/evnchn/nicegui/tree/evaluate-5182 for that. 

Let's wait and observe https://github.com/evnchn/nicegui/actions/runs/18071271779 then.

evnchn:
Something must have went horribly wrong while merging this PR into the experimental branch https://github.com/evnchn/nicegui/tree/evaluate-5182. Tests are failing. 

But after https://github.com/zauberzeug/nicegui/pull/5182/commits/45c2f24d753352b8ea255607a1127e969a715351 we see here, that the pipeline fails with `test_uploading_from_two_tabs` only. It's the expected result. 

So yeah, all is good, let's not touch stuff anymore and wait for @falkoschindler then.

rodja:
I agree. The failing tests are caused by the wrong implementation of `ui.upload` which will be fixed with #5178.

PR #5184, opened 2025-09-29
Make user.should_see work with ui.icon

### LINKED ISSUE #5183: user.should_see cant see icon names as shown on the official website
### First Check

### Example Code

```python
from nicegui import ui


with ui.row():
    ui.label('Hello World!').mark('greeting')
    ui.icon('star')
with ui.row():
    ui.label('Hello Universe!')
    ui.input(placeholder='Type here')

    ui.run()


----

test_file.py


import pytest
from nicegui import ui

import main_report_builder


@pytest.mark.module_under_test(main_report_builder)
async def test_complete_tree_workflow(user: User) -> None:
    """Integration test covering the complete tree workflow."""
    await user.open('/')
    await user.should_see('greeting')
    await user.should_see('star')  # <<<<<<< this is breaking
    await user.should_see('Hello Universe!')
    await user.should_see('Type here')
    await user.should_see('Hello')
    await user.should_see(marker='greeting')
    await user.should_see(kind=ui.icon)


Conftest file

"""This file is used by pytest to discover plugins and fixtures."""

pytest_plugins = ["nicegui.testing.user_plugin"]

import warnings
import tracemalloc

# Enable tracemalloc for better debugging
tracemalloc.start()
warnings.filterwarnings("ignore", category=ResourceWarning)
warnings.filterwarnings("ignore", message=".*pkgutil.find_loader.*", category=DeprecationWarning)
```

### Description

I was trying to find some icons inside a custom expansion header and at first I thought that an custom expansion header slot, was the issue but digging deeper I just noticed that the `user.should()` with icon name is broken. 

So i copy pasted the example from the official website (https://nicegui.io/documentation/user#querying) and that code broke too.  The `await user.should_see(kind=ui.icon)` is working tho.

Might be a good idea to add this as new pytest :) 

### NiceGUI Version

2.24.2

### Python Version

3.13.1

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

rodja:
### Motivation

This PR fixes #5183 where @phifuh noticed that the [documentation](https://nicegui.io/documentation/user#querying:~:text=await%20user.should_see(%27star%27)) demonstrates `user.should_see(...)` with an `ui.icon`. But that does not work.

### Implementation

Simply added a test and added "name" to the properties that are searched by the `ElementFilter`.

### Progress


evnchn:
Is there risk of over-capturing?

image

Tab and Step

rodja:
Ah, good point @evnchn. `should_see` and hence `ElementFilter` content should match what the user "sees" on the screen. We should check if name is generally rendered or just used internally.

phifuh:
I tested this
```python
from nicegui import app, ui

with ui.row().style("align-items: center").classes("mb-4 bg-black w-full").props("gap='md'"):
    ui.icon(name="bar_chart")
    ui.icon(name=None).props("name='bar_chart'")

ui.run(reload=True, title="Test", port=8080)
```
In both cases
```html
<i class="q-icon notranslate material-icons" aria-hidden="true" role="presentation" id="c5">bar_chart</i>
```

I would argue that this is "visible" for the user

but @rodja are there other pytest´s testing the other functionality? 
Else I would highly recommend adding the other tests too

```python
with ui.row():
    ui.label('Hello World!').mark('greeting')
    ui.icon('star')
with ui.row():
    ui.label('Hello Universe!')
    ui.input(placeholder='Type here')
```
    
```python
await user.open('/')
await user.should_see('greeting')
await user.should_see('star')
await user.should_see('Hello Universe!')
await user.should_see('Type here')
await user.should_see('Hello')
await user.should_see(marker='greeting')
await user.should_see(kind=ui.icon)
```

rodja:
Yea. In carousel slides `name` is just used to identify slides, bind with v-model, and control navigation. It is not rendered into the DOM, so it does not show up in accessibility tree, nor is it “visible” to screen readers. Same goes for tabs and stepper. But for icon and chat_message it is shown. I updated the code and test accordingly.

evnchn:
That works, but I am also thinking, in the future each element can have a `.visible_props` which is a list of props that the `ElementFilter` should consider for matching. This way, we can avoid `ElementFilter` having to import everybody and calling `isinstance` often.

rodja:
@evnchn yes that could work. Or we introduce a "can_see" method in element which is free to be implemented as complicated as the element needs it to be. But let's leave that to other PRs.

PR #5187, opened 2025-09-30
Introduce new `sanitize` argument for `ui.html`

falkoschindler:
### Motivation

It has been pointed out to us that `ui.html` could be considered unsafe because it doesn't sanitize its content. It could be filled with user input, opening the door for nasty things.

### Implementation

Therefore we decided to introduce a new required `sanatize` parameter which either accepts a callable sanitization function like `Sanitizer().sanitize` from the `html-sanitizer` package, or `False` to keep the content unchanged. This way we force the developers to think about the intention of `ui.html` elements.

### Progress

    - [x] What to do with `ui.chat_message`?

### DIFF
--- a/examples/chat_with_ai/requirements.txt
+++ b/examples/chat_with_ai/requirements.txt
@@ -1,4 +1,4 @@
-html-sanitizer
+html-sanitizer>=2.6.0
 langchain>=0.2
 langchain-community
 langchain_openai
--- a/examples/generate_pdf/requirements.txt
+++ b/examples/generate_pdf/requirements.txt
@@ -1,3 +1,3 @@
-html-sanitizer
+html-sanitizer>=2.6.0
 nicegui>=3.0
 pycairo


evnchn:
Possibly, maybe sanitize can be `True` and then it will invoke a HTML sanitizer bundled with NiceGUI?

This way, we can keep NiceGUI "nice" and easy-to-use. 

Maybe we have to make it optional dependency and passing `True` will raise error / not render the element without it, otherwise the standalone library becomes larger.

falkoschindler:
@evnchn Because html-sanitizer depends on beautifulsoup4, which is a pretty large dependency for such a rare use-case, we wanted to avoid packing it with NiceGUI. But we haven't thought about `ui.chat_message` which should be sanitized when running with `text_html=True` (not the default).

Maybe we should simply add an optional `sanitize: Callable | Literal[False] = False` parameter to `ui.chat_message` which causes an exception if `text_html=True` and `sanitize=False`. Then the developer experience is: You enable HTML support by `text_html=True` and NiceGUI informs you that you should use a `sanitize` function or opt-out explicitly. This doesn't feel too bad.

rodja (approved):

PR #5190, opened 2025-10-01
Update third-party node libraries

falkoschindler:
### Motivation

This PR updates third-party node libraries a last time before releasing 3.0.

### Implementation

- Called `ncu -u && npm install && npm run build` in nicegui/elements/*.
- Exception: `npm install --force` in nicegui/elements/echart due to outdated echarts-gl.
- Called `ncu -u && npm install && npm run build && ./summarize_dependencies.py` in the root path.

### Progress

- [x] Fix failing Mermaid test

### DIFF
--- a/nicegui/elements/mermaid/mermaid.py
+++ b/nicegui/elements/mermaid/mermaid.py
@@ -20,6 +20,9 @@ class Mermaid(ContentElement, component='mermaid.js', esm={'nicegui-mermaid': 'd
 
         Refer to the Mermaid documentation for the ``mermaid.initialize()`` method for a full list of options.
 
+        **Note:** When using click handlers with ``securityLevel: 'loose'``,
+        make sure to use unique node IDs across diagrams to avoid click events being bound to the wrong elements.
+
         :param content: the Mermaid content to be displayed
         :param config: configuration dictionary to be passed to ``mermaid.initialize()``
         """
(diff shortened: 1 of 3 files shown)


evnchn:
Maybe we can consider bump minor/patch versions for all future NiceGUI 3.x releases, seeing 2.x suffered from a client side memory leak due to old Quasar version, now that the dependency management is more streamlined.

falkoschindler:
@evnchn Yes, we can certainly bump patch versions now and then before 4.0. But some libraries break behavior in minor versions. Right now I'm fighting with Mermaid, which started confusing click handlers between nodes in separate diagrams with same node IDs. This must have been introduced between 11.0.2 and 11.9.0.

evnchn:
> which started confusing click handlers between nodes in separate diagrams with same node IDs

Retrospect, I can't eliminate that it is potentially something to do with how we implemented things (plus both the mermaid and markdown's mermaid are done by me, so 🧐)

Glad that the tests works now so that makes things _a bit_ more firm. We'll see in the RC if things are truly right I guess.

falkoschindler:
@evnchn I traced the Mermaid problem down to a plain HTML example which breaks at version 11.5.0 when they changed how nodes are identified. So it's not our fault and we can't protect our users against all upstream shenanigans.

PR #5193, opened 2025-10-02
Implement ui.sub_pages navigation for user simulated tests

rodja:
### Motivation

Currently the user fixture always triggers a full page reload — even if the navigation should be done with `ui.sub_pages`. The core problem: the user fixture injects its own `ui.navigate.to` implementation to skip over any browser and JavaScript behaviour, and this custom implementation was not aware of sub pages navigation.

### Implementation

`UserNavigate.to/back/forward` and `ui.link` clicks in the user simulation now funnel through a single sub-pages-aware helper `UserNavigate._open()`. The helper mirrors the production logic in `nicegui/functions/navigate.py`: when the path is relative and the current layout contains a `ui.sub_pages`, it routes through `sub_pages_router._handle_open()`. If the resulting state is unresolved and another `@ui.page` builder claims the path, it falls back to a full `user.open()`, matching the `_handle_navigate` fallback used in production.

`UserNavigate.reload()` deliberately bypasses `_open()` and always calls `user.open()` directly, mirroring production's `history.go(0)` semantics (full page reload, not partial sub_pages routing).

Drive-by: `Navigate.to` now detects `SubPages` via `client.layout.descendants()` instead of the flat `client.elements` registry, aligning with `_handle_open`, `has_any_unresolved_path`, and the new user-simulation helper.

### Progress

### DIFF
--- a/nicegui/functions/navigate.py
+++ b/nicegui/functions/navigate.py
@@ -69,7 +69,7 @@ class Navigate:
         if not new_tab and isinstance(target, str):
             parsed = urlparse(path)
             if not parsed.scheme and not parsed.netloc and \
-                    any(isinstance(el, SubPages) for el in context.client.elements.values()):
+                    any(isinstance(el, SubPages) for el in context.client.layout.descendants()):
                 client = context.client
                 navigate_coro = client.sub_pages_router._handle_navigate(path)  # pylint: disable=protected-access
                 background_tasks.create(helpers.await_with_context(navigate_coro, client), name='navigate_sub_pages')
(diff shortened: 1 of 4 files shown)


falkoschindler:
Rebased onto main and refactored to unify the navigation paths: `UserNavigate.to/back/forward/reload` and `ui.link` clicks now share a single `_open()` helper that mirrors `Navigate.to()` in `nicegui/functions/navigate.py` (urlparse + SubPages-in-layout check, fallback to full `user.open`). All 70 user-simulation tests and 68 sub-pages tests still pass, linters green. Marking ready for review.

falkoschindler (approved):
After resurrecting and updating this old PR with massive support by Claude, I think it's ready to merge. The description is updated to cover all additional changes. Now the simulation should better mirror the real behavior and a pytest checks the various navigation paths.

PR #5199, opened 2025-10-03
Respect show_404=False in sub_pages navigation

rodja:
### Motivation

The `show_404` parameter in `ui.sub_pages` was not properly respected during navigation. When set to `False`, the router would still block navigation to other page builders even though no 404 message was shown.

It was a problem in the docs. To reproduce go to `/documentation/page_layout/` and click on "show page with fancy layout". The expected page is not shown until you refresh the browser. NOTE: on https://nicegui.io the bug is already fixed by applying the idea of this PR.

### Implementation

Separated path resolution state (`has_404`) from 404 display behavior (`_404_enabled`):
- `has_404` is now set unconditionally when a path cannot be resolved
- Navigation decisions only block when `has_404=True` AND `_404_enabled=True`
- Added `_has_any_unresolved_path()` helper method for cleaner logic
- Extended test coverage to validate both `show_404=True` and `False` scenarios

### Progress

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Bugfixes
 
+- Respect `show_404=False` in `ui.sub_pages` navigation (#5199 by @rodja, @falkoschindler)
 - Fix docker builds by not excluding dist/ directories from the build context (#5200, #5202 by @Yuerchu, @rodja, @falkoschindler)
 
 ### Testing


evnchn:
Hmm we may have entirely forgot about `nicegui-highcharts`. It's still using the 2.x syntax for JS dependency management, and so it's not the case of just relaxing the `pyproject.toml` for a fix...

falkoschindler (approved):

PR #5202, opened 2025-10-04
Fix docker builds where element dist/ dirs had been excluded from the build context

### LINKED ISSUE #5200: Some NiceGUI documentation can't be opened
### First Check

### Example Code

```python

```

### Description

Today when I tried to open the NiceGUI documentation, some pages were blank, just like in this video:

https://github.com/user-attachments/assets/32cf0bf0-c630-4270-9ecd-478ce0d74078

In the document content, I can't open the details from this `See more` no matter what, unless I can successfully enter from the left sidebar.
And after this blank page appeared, I tried using the browser's forward and back functions, and they were completely unresponsive as well.

### NiceGUI Version

3.0

### Python Version

None

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

rodja:
### Motivation

This PR fixes #5200 where @Yuerchu reported that the public documentation has issues displaying certain content. After some investigation me and claude-4.5-sonnet found the core issue: our `.dockerignore` file excluded all `dist/` dirs from the docker build. But in #5021 we changed the dependency management so an element can provide their js code in `dist/` folders.

### Implementation

This PR just ensures elements `dist/` directories are not excluded from docker context.

NOTE: the fix is already deployed on https://nicegui.io

### Progress

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@
 **/*.egg-info
 **/.*.swp
 **/dist
+!nicegui/elements/*/dist
 test.py
 demo.py
 **/*.pickle


evnchn:
`dist` was gone from `.gitignore` in https://github.com/zauberzeug/nicegui/commit/7762d62b362958a8f997bf07296a261e2fba93b8#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947L4 but we may have forgot to update the `.dockerignore` which _should_ typically sync with `.gitignore`

On the remaining sync between `.gitignore` and `.dockerignore`:

- We may run the risk of packing `node_modules/` into the docker image, if a person packed their local directory which they are doing NiceGUI devwork on. It makes the image bulky but is not harmful I guess.

PR #5204, opened 2025-10-04
Re-run `poetry lock` to fix `nicegui-highcharts` incompatibility

evnchn:
### Motivation

Pipeline's red. Dev container cannot be rebuilt. Because poetry still uses `nicegui-highcharts` 2.1.0 as in the lockfile

### Implementation

Re-run `poetry lock`, the commit the lockfile

### Progress


evnchn:
@rodja You'll want this for the pipeline gives better sense of security when delivering other changes.

PR #5206, opened 2025-10-04
Allow non-AMD64 devcontainer, such as ARM (reverts #3328)

evnchn:
### Motivation

I just checked and we can install `chromium` and `chromium-driver` on ARM64, covering a major use case of developing on Apple Silicon Macs

image

I can't trace back which release it was supported, though. 

So we can revert #3328 and improve devcontainer performance. 

For those who need AMD64 they can specify it themselves? 

### Implementation

Revert #3328, drop `--platform=linux/amd64`

### Progress

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 python:3.9
+FROM python:3.9
 
 ENV POETRY_VERSION=1.6.1 \
     POETRY_NO_INTERACTION=1 \


evnchn:
I'm still on the ARM64 devcontainer, by the way.

evnchn:
Tests cannot run??

# 🤡

image

I don't have that much experience. Let's make this draft until someone can look at this and make the tests pass, so that we don't accidentally merge this and mess up the devcontainer for everybody.

evnchn:
Don't ask me why but with https://github.com/zauberzeug/nicegui/pull/5206/commits/e2c6ba94e7439a8cf00e0eba05e0e96f08aa98b9 it works

Not all tests pass though, some fail, but I don't know why. The tests which do pass though, they are very fast.

evnchn:
I just realized that actually for Polars it will complain if you use AMD64 devcontainer on Apple Silicon

https://github.com/pola-rs/polars/blob/afdbf3056d1228cf493901e45f536b0905cec8ea/py-polars/src/polars/_cpu_check.py#L250-L262

So for Apple Silicon Macs, this PR is more important than what it seems.

falkoschindler (approved):

PR #5209, opened 2025-10-04
Check disabled state before clicking or clearing via user simulated interaction in pytests

### LINKED ISSUE #5191: Able to click to disabled Button element when using User Fixture in testing
### First Check

### Example Code

```python
import pytest
from nicegui import ui

BUTTON_NAME = "My Button"
BUTTON_TEXT = "Button clicked"


def build_ui() -> None:
    @ui.page("/")
    def page() -> None:
        with ui.button(BUTTON_NAME) as button_element:
            button_element.disable()
            button_element.on_click(lambda: ui.notify(BUTTON_TEXT))


@pytest.fixture
async def my_user(user: User) -> User:
    build_ui()
    await user.open("/")
    return user


@pytest.mark.asyncio
async def test_type_to_disabled_input(my_user: User) -> None:
    my_user.find(BUTTON_NAME).click()
    await my_user.should_not_see(BUTTON_TEXT)
```

### Description

This is similar to https://github.com/zauberzeug/nicegui/issues/4080

# Expected Result
User fixture can click disabled button

# Actual Result
User fixture cannot click disabled button

# pytest versions:
pytest            8.4.2
pytest-asyncio    1.2.0

### NiceGUI Version

2.24.2

### Python Version

3.11.10

### Browser

Other

### Operating System

macOS

### Additional Context

_No response_

rodja:
### Motivation

This PR fixes #5191, where @igro-marczak reported that the disabled state is currently ignored when clicking with the user fixture.

### Implementation

The fix ensures that disabled elements (buttons, checkboxes, switches, etc.) behave correctly when clicked or cleared, matching real browser behavior -- same way as it was already implemented for `type(...)`.

### Progress

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## [Unreleased]
 
+### Bugfixes
+
+- Check disabled state before clicking or clearing via user simulated interaction in pytests (#5209 by @igro-marczak, @rodja)
+
 ### Infrastructure
 
 - Re-run poetry lock to fix nicegui-highcharts incompatibility (#5204 by @evnchn)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## [Unreleased]
 
-### Bugfixes
+### Testing
 
-- Check disabled state before clicking or clearing via user simulated interaction in pytests (#5209 by @igro-marczak, @rodja)
+- Check disabled state before clicking or clearing via user simulated interaction (#5191, #5209 by @igro-marczak, @rodja)
 
 ### Infrastructure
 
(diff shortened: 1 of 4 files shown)


evnchn (approved):
Code works (tested by disabling the patch and we see the tests fail, with the patch the tests pass). 

---

I am just thinking if we should cover the case where I have an element but I did `set_visibility(False)`

- On one hand, user can typically NOT click or clear the element
- But, the element is not disabled, and our server-side validation is not enforced anyways. 

🤔 I am definitely overthing though, this PR is fine.

rodja:
@evnchn you would not find the element if it is invisible. Of course you could store a `UserInteraction` and click/clear it later. But that is not the intended usage. Definitely needs a discussion first (not on this PR).

PR #5213, opened 2025-10-05
Enable setting session cookie attributes from ui.run()

samuller:
### Motivation

When using `app.storage.user` it creates a session cookie with permissions that disallow cross-origin usage. While this is a good secure default, there's currently no way to override it when needed, e.g. if you want your NiceGUI site to be embeddable as an iframe from another domain. See discussion https://github.com/zauberzeug/nicegui/discussions/4252.

### Implementation

A new argument is added to `ui.run()` which allows passing keyword arguments that will be forwarded to the [SessionMiddleware](https://www.starlette.dev/middleware/#sessionmiddleware) that is used to create the session cookie.

### Progress


falkoschindler (approved):
Thanks for the pull request, @samuller!

I just simplified the pytests a bit. Let's merge!

PR #5214, opened 2025-10-05
Fix support for Redis Unix-Domainsocket-connections

senfomat:
### Motivation

I usually use Redis with UnixDomainsocket-connections, using connection-URLs beginning with `unix://`. When trying to use this for persistence in NiceGUI I discovered, that this throws the error 'Could not load data from Redis with key ...'. The reason is, that the option `socket_keepalive` is not present when `redis.from_url()` leads to the use of Redis' class `UnixDomainSocketConnection`.

### Implementation

The fix builds a connection-parameter-dict in the class `RedisPersistentDict` and adds the option 'socket_keepalive' ony when the URL is not beginning with `unix://`. Furthermore it reuses this dict for the Redis-connection in the same Class on `initialize_sync()`.

### Progress

- [x] Pytests are are not necessary.


evnchn:
Kinda awkward to mix args and kwargs into `_redis_client_params` and the `@property` the `self.url`. Maybe let's not over-engineer things.

falkoschindler (approved):
Ok, I agree. We don't have to over-engineer it.

- I just marked the parameter dictionary as private.
- And I refactored the dictionary definition a little bit to have all parameters in one block right when creating the dictionary. The nested dictionary spreading might be a bit peculiar, but we already use this pattern in other places. I like the fact that you can clearly see the four parameters without skimming another if-condition.

Let's merge! 🚀

PR #5220, opened 2025-10-06
`ui.log` scroll-to-bottom working in `ui.tab_panel`

evnchn:
### Motivation

Fix #5118, where we find #5112 (or, actually, the reimplementation of it in 3.0, since we do `q-scroll-area` for `ui.log`) does not work inside `ui.tab`. 

### Implementation

I did 2 things: 

- Act, only when `document.body.contains(this.$refs.qRef.$el)` (aka only when body contains the element, which is false when the `ui.log` hidden due to inactive in `ui.tab_panel`
- Go back to `scrollTop + clientHeight >= scrollHeight` approach for checking current scroll but we do it on `this.$refs.qRef.$el.childNodes[0]` this time around. 
  - This I not sure but I think `getScroll().verticalPercentage` isn't accurate while `ui.tab_panel` under animation effect. 

### Progress

  - @falkoschindler Please do look at this, I am unsure. 


evnchn:
In my testing I found `scrollTop + clientHeight >= scrollHeight` more reliable. I can't explain it but it is real though. 

Still, if the problem persists, then we need to try again.

PR #5222, opened 2025-10-06
Allow preventing default handlers for key events

falkoschindler:
### Motivation

In #4924 @samuller mentioned that the "/" key on the NiceGUI website doesn't work flawlessly to open the search dialog on Firefox because the default handler opens Firefox' built-in search field. We also discussed how to bring the `preventDefault()` and `stopPropagation()` functions to `ui.keyboard`.

### Implementation

- This PR adds the original JavaScript event object to the argument dictionary which is passed to `js_handler`.
- A new demo shows how to use `.on('key', ..., js_handler=...)` to handle key events on the client and conditionally call `preventDefault()` or `stopPropagation()`.
- A similar `js_handler` function is used to improve the "/" key behavior on the NiceGUI website.
- The formatting of some version specifiers is fixed.

### Progress

### DIFF
--- a/nicegui/elements/button_dropdown.py
+++ b/nicegui/elements/button_dropdown.py
@@ -55,7 +55,7 @@ class DropdownButton(IconElement, TextElement, DisableableElement, BackgroundCol
     def on_click(self, callback: Handler[ClickEventArguments]) -> Self:
         """Add a callback to be invoked when the dropdown button is clicked.
 
-        **Added in version 2.22.0**
+        *Added in version 2.22.0*
         """
         self.on('click', lambda _: handle_event(callback, ClickEventArguments(sender=self, client=self.client)), [])
         return self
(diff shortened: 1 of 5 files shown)


samuller:
Does the `ui.keyboard`'s `ignore=` argument continue working, or will the `js_handler` side-step it?

If so, then `preventDefault()` would also prevent the characters from being typed in input boxes (same issue that happened to me in https://github.com/zauberzeug/nicegui/discussions/4924).

falkoschindler:
@samuller Good question! I haven't thought about that.
But the `ignore` list is handled before the `js_handler` is called. So typing in an input field should still work.

rodja (approved):
Elegant!

evnchn:
Briefly looked where else we are emitting events and did not expose the browser's event. I don't see any. No problems with this PR. 

Just got extra stuff to address later when see fit. 

On the formatting of version specifiers and other things in the docstring, it's such an easy thing to miss and I'm thinking one day we may need to put it in the CONTRIBUTING.md so that AI agents can help us check things, or maybe even put this in the pipeline code-check if we have a reliable detection method.

PR #5231, opened 2025-10-08
Make Event sync between instances in distributed setups

rodja:
### Motivation

Currently it's tricky to appreciate the [`Event` demo in the documentation](https://nicegui.io/documentation/event#event): when sending a "tweet" it only appears on visitors of the same instance. Due to the distributed deployment via fly.io, if a user opens two tabs, they have a high chance of landing on different instances.

### Implementation

This PR is currently just an (AI-aided) sketch of an idea I had after some research. By using [Zenoh](https://zenoh.io/)  to transparently sync events across instances. Zenoh is great for this because it does not need a broker or seperate server (like MQTT or Redis). Instances simply auto-discorver. To test it, I created a new example called `distributed`: A docker swarm starts three NiceGUI instances and a Traefik proxy in front:

Oct-08-2025 06-36-34

### Progress

  - [ ] error handling, edge cases, ... there is still much much to do before it might be ready
- [ ] Documentation has been added.
   - [x] example
   - [ ] link to example
   - [ ] info in documentaiton


rodja:
It might even be possible to create a `ZenohPersistentDict` and hence allow a distributed storage without a seperate Redis server. But we should leave that for an other PR.

evnchn:
Could also make https://nicegui.io/documentation/refreshable#global_scope work, where we faking it with the use of `Event`

But in another PR.

NiklasNeugebauer:
https://chatgpt.com/share/68f0d6ca-6f5c-800c-8d4c-2f7abbda6632

I wonder if the topic should be generated differently. In particular, I think events that are generated from a library like rosys would always result in the same topic right now.

I am not sure if it is worth considering this in the current state, but I wanted to leave a note here.
See the small chatgpt conversation about it above.

falkoschindler:
Good point, @NiklasNeugebauer. Definitely worth checking.

---

I'm wondering if we really should add the code for distributed events into the `Event` class, or if something like a `DistributedEvent` inheriting from `Event` would be better. This would allow the user to explicitly create `Event`s or `DistributedEvent`s.

@rodja, you probably want a global switch `ui.run(distributed=True)` so that _all_ events are distributed at once. But this feels a bit unintuitive and implicit to me. Is it clear to the user that this parameter affects all events (and nothing else)? And do we only have two use cases, either _no_ event distribution or _complete_ event distribution?

Maybe we need to think about specific real-world applications to answer such questions about the desired API. Since this PR is only a draft, there's no urge in answering them now. I just wanted to leave my thoughts here for reference.

rodja:
> [I wonder] if something like a DistributedEvent inheriting from Event would be better

Yes @falkoschindler. Feels great and keeps the responsibilities apart.

> I wonder if the topic should be generated differently.

You are right @NiklasNeugebauer. It would be better to incorporate the app which is running into the auto-generated topic. Maybe the app title is unique enough in combination with the existing logic? Of course it should also be possible to optionally define your own topic string.

NiklasNeugebauer:
I think some manually set instance id would be useful so that a user does not have to define every event topic but is still able to differentiate the instances (by version, specific application, whatever).

evnchn:
@rodja Now I am an owner of a homelab with at least 4 nodes able to participate in this PR, what do you think if I battle-test on their with me and my agent? If OK, I can take this PR and pushforward, because the immersiveness of the demos in https://nicegui.io and the "able to deploy NiceGUI to many instances" argument can be strengthened by this PR if we push for it.

rodja:
@evnchn it would be super cool if you could take over. I think the event sync is a very valuable feature. I just do not have the time to work on it.

evnchn:
*Posted on Evan's behalf by Claude Code — see [CLAUDE.md](https://github.com/evnchn/nicegui/blob/distributed-cleaned/CLAUDE.md).*

Heads-up @rodja: I took the handoff seriously and opened #6088 with a re-implementation against current `main`.

Quick status:
- Battle-tested on multi-node homelab (real LAN, not just same-daemon replicas) - cross-instance emoji float-up confirmed end-to-end via two-tab Playwright.
- Ran the surface through a multi-reviewer pass (a Claude-family code reviewer + GitHub Copilot) before opening upstream; the Copilot pass caught a real Zenoh-thread → asyncio thread-safety bug that source-read alone missed.
- 12/12 pytest, mypy clean, pylint 10.00/10, pre-commit clean.

The implementation diverged from this PR enough that range-diff wouldn't help (`storage_secret` required for namespace, peers-list shortcut, no absolute filepath in topic), so it's a fresh PR rather than a rebase.

**No rush at all** - explicitly not for the upcoming release. Wanted to let you know the homelab handoff worked, and to ask: are you OK closing this in favour of #6088 so you and @falkoschindler can review whenever you feel like it?

Full evidence trail (multi-reviewer logs + empirical-verification probe + the followup commits that fixed each reviewer finding) lives on the fork PR https://github.com/evnchn/nicegui/pull/164 for anyone curious.

falkoschindler:
Thanks for picking this up and battle-testing it on real hardware, @evnchn! 🙏

Yes, let's close this in favour of #6088 (answering on @rodja's behalf). It's the right call:

- This PR was always an early draft — incomplete and by now conflicting with `main`.
- #6088 is a complete, fresh re-implementation against current `main` (not a rebase, since the API diverged too much for a range-diff to help), and it's tested cross-host.
- Nothing is lost: the design discussion here stays preserved and is linked from #6088, and a closed PR can always be reopened if needed.

PR #5232, opened 2025-10-08
Pytest: wait for Mermaid to render

evnchn:
### Motivation

Infuriating to wake up to a red pipeline in my fork. Consider https://github.com/evnchn/nicegui/actions/runs/18337127046

### Implementation

This pull request enhances the test coverage and reliability for Mermaid diagram rendering in both `test_markdown.py` and `test_mermaid.py` by adding explicit assertions for node content and display. The changes ensure that key nodes are present or absent as expected after user interactions.

**Improvements to test assertions for Mermaid diagram rendering:**

* Added checks in `tests/test_markdown.py` to assert that 'Node_A' and 'Node_C' are present or absent as appropriate after setting new content, and that these nodes are displayed in the rendered output.
* Included assertions after clicking 'Create Mermaid' in `tests/test_markdown.py` to verify that both 'Node_A' and 'Node_B' are present and displayed.
* Enhanced `tests/test_mermaid.py` by adding assertions to confirm that 'Node_A' and 'Node_C' are present and have the correct class, and that 'Node_A' is absent after content changes.

HUMAN NOTE: By the "checks" I mean `screen.should_contain` which has a built-in retry mechanism. Check https://github.com/zauberzeug/nicegui/pull/5232#issuecomment-3380852682 and https://github.com/zauberzeug/nicegui/blob/c5c9dfe9952baad414acf0c4f5ceef3ebee3912b/nicegui/testing/screen.py#L195-L210

### Progress

### DIFF
--- a/tests/test_markdown.py
+++ b/tests/test_markdown.py
@@ -46,6 +46,7 @@ def test_markdown_with_mermaid(screen: Screen):
 
     screen.open('/')
     screen.should_contain('Mermaid')
+    screen.wait(1.0)  # wait for Mermaid to render
     assert screen.find_by_tag('svg').get_attribute('id') == f'{m.html_id}_mermaid_0'
     _wait_for_xpath_displayed(screen, '//span[p[contains(text(), "Node_A")]]')
 


falkoschindler:
Thanks for looking into this flaky test, @evnchn! It also caught my attention a few days ago.

But I think your new delays are irrelevant to the specific problem. If the test fails, it also happens in test_markdown.py in `test_markdown_with_mermaid()` at the line
```py
assert screen.selenium.find_element(By.XPATH, '//span[p[contains(text(), "Node_A")]]').is_displayed()
```
That's above the new delay this PR is adding. Before this line there is already a delay which I recently increased to 1 second in https://github.com/zauberzeug/nicegui/commit/3b2d6a070234db78a33c770a47736d74fdfa2940. Unfortunately this didn't help.

So it remains unclear how to fix this problem. We could further increase the duration, but this adds to the overall time it takes to run all tests. Maybe calling `screen.wait(1.0)` _after_ `screen.should_contain('Mermaid')` could help, so we don't spend our precious waiting time while the page is still loading and rendering hasn't started yet?

evnchn:
```
        screen.open('/')
        screen.wait(1.0)  # wait for Mermaid to render
        screen.should_contain('Mermaid')
        assert screen.find_by_tag('svg').get_attribute('id') == f'{m.html_id}_mermaid_0'
>       assert screen.selenium.find_element(By.XPATH, '//span[p[contains(text(), "Node_A")]]').is_displayed()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Indeed the waits, as inserted, do not fix the issue. Because the issue surfaces even with the waits just like how I inserted it. My apologies for the 🤡. 

~~Consider my latest approach at https://github.com/zauberzeug/nicegui/pull/5232/commits/8a651431f605ca24732a6696f1a1146a7ee800ff, which uses `screen.should_contain` (has a built-in retry mechanism) to probe for the Node (not the text "Mermaid", since that shows up always instantly as part of the Markdown) before trying to do `screen.selenium.find_element` on it.~~ Not updated...

Note that with #4722 and #4692 we no longer flash-render the Mermaid raw syntax text since we use the Mermaid API directly. I wrote it so I know it.

---

Well, as with pipeline stuff, there's no better way but to try out different and basically fire shots into the dark (tests always pass locally...). It's more "merge it if it doesn't break existing tests, then pray that it isn't flaky". 

@falkoschindler Two things though, for another PR: 

- Maybe we can submit the artifact of the screenshot, if any, so that we see them visually? I don't think theres a way with GitHub Actions right now, and it will be an important tool.
- How about using https://pypi.org/project/pytest-retry/ everywhere? Doesn't it sound more convincing that a test is failing, when it has failed 10 times in a row?

evnchn:
https://github.com/zauberzeug/nicegui/actions/runs/18405814901/job/52445352379

image

Huh... So Mermaid _is_ broken when we have a failing test...

falkoschindler:
Very interesting! This confirms my suspicion that we shouldn't simply repeat tests that don't pass on first try.

evnchn:
We will have to re-think, if that's the case. 

- This PR: Still safe to merge. Maybe with my more tolerant logic Mermaid loads. 
- #5236: If the purpose of adding reruns is merely to shove problems under a rug, we will say no to it. 
- #5249: If the purpose of adding reruns is to make parallelized tests work, then maybe we can say yes, but we may need to run the the test suite in a fully non-parallelized manner before release / at midnight, where we are not actively waiting for the pipeline to inform our coding.

falkoschindler:
@evnchn Do we really need `_wait_for_xpath_displayed`, or is it sufficient to swap the first `wait` and the first `should_see`?

evnchn:
@falkoschindler AFAIK `screen.should_contain`, although internally tries a couple of times, errors out when the element is available but hidden, and so my `_wait_for_xpath_displayed` is the only way to try a couple of times on the "Node_A" which could be hidden for a while until it's displayed.

evnchn:
Closing this in favor of https://github.com/zauberzeug/nicegui/pull/5366 since the JS error logs are "irrefutable evidence" that more tolerant pytest checking will not fix the issue, i.e. you can wait 1000 seconds and the plot still won't be there.

PR #5234, opened 2025-10-08
Upload screenshot on fail

evnchn:
### Motivation

When the tests fail in GitHub Actions, cannot see visually what's wrong. 

### Implementation

This pull request enhances the CI workflow to improve the handling of failed test screenshots and error reporting. The main changes include adding a script to collect screenshots for failed tests, updating the test workflow to generate and upload these screenshots, and introducing a deliberate test failure to verify the process.

**CI workflow improvements:**

* Added a new script `collect_failed_screenshots.py` that parses `reports/junit.xml` for failed tests and copies matching screenshots from `screenshots/` to `failed-screenshots/` for easier artifact collection.
* Updated `.github/workflows/test.yml` to:
  - Run pytest with `--junitxml=reports/junit.xml` to generate a JUnit XML report.
  - Always run the screenshot collection script after tests.
  - Upload failed screenshots as an artifact if tests fail.
  - Always upload the JUnit XML report as an artifact.

**Test reliability and debugging (TEMPORARY, REVERT BEFORE MERGE):**

* Added a deliberate error (`raise ValueError(...)`) at the end of the `page()` test in `tests/test_mermaid.py` to ensure screenshot capturing and error reporting are functioning as expected.

### Progress

- [ ] Remoce the deliberate error after everything is finalized, just before merging this PR.

### DIFF
--- a/.github/scripts/collect_failed_screenshots.py
+++ b/.github/scripts/collect_failed_screenshots.py
@@ -17,7 +17,8 @@ def main() -> int:
     junit = Path('reports/junit.xml')
     screenshots = Path('screenshots')
     out = Path('failed-screenshots')
-    shutil.rmtree(out)
+    if out.exists():
+        shutil.rmtree(out)
     out.mkdir(parents=True, exist_ok=True)
 
     if not junit.exists():
--- a/.github/scripts/collect_failed_screenshots.py
+++ b/.github/scripts/collect_failed_screenshots.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+"""Collect screenshots for failed pytest tests.
+
+Reads `reports/junit.xml` for `<failure>` or `<error>` entries and copies
+matching files from `screenshots/` to `failed-screenshots/`.
+
+It supports matching multiple files per test name using glob patterns.
+"""
+from __future__ import annotations
+
+import shutil
+import xml.etree.ElementTree as ET
+from pathlib import Path
+
+
+def main() -> int:
+    junit = Path('reports/junit.xml')
+    screenshots = Path('screenshots')
+    out = Path('failed-screenshots')
+    shutil.rmtree(out)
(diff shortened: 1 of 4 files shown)


evnchn:
https://github.com/evnchn/nicegui/actions/runs/18354718667/job/52283629698

Works when there's error. Will wait for @falkoschindler to review what we have now, before reverting https://github.com/zauberzeug/nicegui/pull/5234/commits/98184f8bc4ee6cd0f54131de2fce4411bc9c35a3 to see if things work when the tests do not error out.

falkoschindler:
Thanks for the PR, @evnchn! The screenshots not being available when run by the CI pipeline is bothering me for quite a while now. Great that it's finally working!

I just wonder if we can simplify things a little.

- The script smells a bit like AI-generated.
    - `raise SystemExit(main())` is a pattern I haven't seen before.
    - `from __future__ import annotations` seems useless since the code isn't basically not typed.
- Do we really need to add the two new folders to .gitignore? They are only created within the pipeline anyway.
- If we would only store screenshots for failed tests in the first place (or in separate directories depending on the test outcome), we wouldn't need to split them later. I'm experimenting with such an approach on a new branch: https://github.com/zauberzeug/nicegui/tree/upload-failed-screenshots.

evnchn:
Oh yes the script is AI generated and I left it that way for now. I figured until we agree with the approach we don't touch it, and we'll have to decide between the approach of (1) don't mess with Pytest, but save the JUnit XML and parse it, or (2) get Pytest to segregate the screenshots. 

The adding of `.gitignore` is for facilitate running the pytest locally without causing uncommited file.

falkoschindler:
Ok, I created PR #5251 as a comparison. Let's hear what @rodja thinks about both approaches.

evnchn:
There's no competition. 

The XML parser script is 🤡 courtesy of GitHub Copilot after all...

If you really want to do this you can re-open the PR.

PR #5237, opened 2025-10-08
Also test CPython 3.14

evnchn:
### Motivation

Python 3.14 is released! https://www.python.org/downloads/release/python-3140/

### Implementation

This pull request makes a small update to the test workflow by adding Python 3.14 to the matrix of Python versions tested in `.github/workflows/test.yml`. This ensures that the project is tested against the upcoming Python release.

NOTE: I am not sure why but in PyPI we are listed for 3.14 already. Check https://pypi.org/project/nicegui/

### Progress

### DIFF
--- a/test_startup.sh
+++ b/test_startup.sh
@@ -49,6 +49,11 @@ do
         continue # until https://github.com/omnilib/aiosqlite/issues/241 is fixed
     fi
 
+    # Skip examples/ai_interface for Python 3.14
+    if [[ $(python3 --version | cut -d' ' -f2 | cut -d'.' -f1,2) =~ ^3.14$ ]] && [[ $path == "examples/ai_interface" ]]; then
+        continue # It still uses Pydantic V1, which breaks horribly with Python 3.14
+    fi
+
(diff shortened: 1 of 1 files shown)


evnchn:
# 🤡 

image

falkoschindler:
> I am not sure why but in PyPI we are listed for 3.14 already.

Our python dependency is defined as `python = "^3.9"` in pyproject.toml, so everything above 3.8 and below 4.0 is valid.

evnchn:
Blocked by release of https://github.com/manatlan/vbuild/pull/16

evnchn:
Due to vbuild not released their Python 3.14 compatibility, if you do Python 3.14 for NiceGUI now, it errors out. the hack to use vbuild off git for Python 3.14 isn't as bad as it seems. I take 0->0.5 as a win.

evnchn:
As we can see in #5242 things are a big mess when we update `poetry.lock` with a shotgun approach. I'm going to try again instead of force-push to preserve this branch's history.

PR #5239, opened 2025-10-08
Only unsubscribe events when deleting a client

falkoschindler:
### Motivation

Within a UI context event handlers are automatically unsubscribed when the client disconnects. But this is slightly incorrect because the client can reconnect.

### Implementation

This PR uses `on_delete` instead of `on_disconnect` to register the unsubscription.

### Progress

- [x] When uncommenting the whole auto-unsubscribe block, I still can't cause any memory leak. So do we even need this feature? --> Ah ok, clients are still pruned, but the list of `callbacks` would grow indefinitely. So it is good to unsubscribe.
- [x] The name of the optional parameter is still `unsubscribe_on_disconnect` to avoid breaking the API and the documentation still talks about client disconnection. Is this ok?

### DIFF
--- a/nicegui/event.py
+++ b/nicegui/event.py
@@ -63,17 +63,16 @@ class Event(Generic[P]):
         self.instances.add(self)
 
     def subscribe(self, callback: Callable[P, Any] | Callable[[], Any], *,
-                  unsubscribe_on_disconnect: bool | None = None) -> None:
+                  unsubscribe_on_delete: bool | None = None) -> None:
         """Subscribe to the event.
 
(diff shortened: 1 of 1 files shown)


evnchn:
In theory, `unsubscribe_on_disconnect` and `unsubscribe_on_client_deletion` can exist in parallel, so that we don't make breaking changes. 

The first one is good for some events which you don't want to react to when the client is temporarily disconnected (you need to re-subscribe to these events)

The second one is good for events which you don't mind reacting to when the client is temporarily disconnected. 

But obvious the second one is so much more useful than the first one.

PR #5242, opened 2025-10-08
Just poetry relock (as well as fix for yanked propcache 0.4.0)

evnchn:
### Motivation

The dependencies in `poetry.lock` are OLD.

### Implementation

After realizing that Poetry sort-of "recomputes everything" when you delete `poetry.lock` and run `poetry lock` again, I promptly did that, only to find my pipeline's red. I also did that for #5237, which I'm going to undo that and force-push without this / open another PR

Additionally, this PR makes a minor update to the `pyproject.toml` dependencies. The change excludes the `0.4.0` version of the `propcache` package, which is used transitively by `aiohttp`, because that version has been yanked and there is currently no replacement.

- Dependency management:
  * Excluded `propcache` version `0.4.0` in `pyproject.toml` since it has been yanked and is used transitively by `aiohttp`.

### Progress

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -26,98 +26,132 @@ files = [
 
 [[package]]
 name = "aiohttp"
-version = "3.12.15"
+version = "3.13.0"
 description = "Async http client/server framework (asyncio)"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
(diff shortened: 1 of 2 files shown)


evnchn:
Pipeline results: https://github.com/evnchn/nicegui/actions/runs/18351721947

Too deep to understand 🤡

evnchn:
VSCode hangs when I try and resolve merge conflict. Also resolving merge conflict for these computer-generated files isn't the best idea either.

PR #5243, opened 2025-10-08
Also test CPython 3.14 (V2)

evnchn:
### Motivation

Python 3.14 is released! https://www.python.org/downloads/release/python-3140/

**Real motivation: #5237 did not work out...**

### Implementation

This pull request adds support for Python 3.14 to the project and addresses compatibility issues with certain dependencies. The most important changes are grouped below:

**Python 3.14 support and dependency compatibility:**

* Added Python 3.14 to the test matrix in the GitHub Actions workflow to ensure CI coverage for the new version.
* Updated the `vbuild` dependency in `pyproject.toml` to use a specific Git commit for Python 3.14, as the released version is not yet compatible.
* Added a conditional requirement for `pydantic` in `pyproject.toml` to exclude versions 2.0–2.11 for Python 3.14 and above, since these versions are incompatible.

**Test script adjustments:**

* Modified `test_startup.sh` to skip the `examples/ai_interface` test when running under Python 3.14, due to its reliance on Pydantic V1 which is not compatible with Python 3.14.

### Progress

- [x] The implementation is complete (We'll stare at the pipeline to make that conclusion).

### DIFF
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -3,6 +3,7 @@ import os
 import time
 from collections.abc import Awaitable, Generator
 from concurrent.futures.process import BrokenProcessPool
+from pickle import PicklingError
 
 import pytest
 
@@ -50,7 +51,7 @@ async def test_run_unpickable_exception_in_cpu_bound_callback(user: User):
 
     @ui.page('/')
     async def index():
-        with pytest.raises(AttributeError, match="Can't pickle local object|Can't get local object"):
+        with pytest.raises((AttributeError, PicklingError), match="Can't pickle local object|Can't get local object"):
             ui.label(await run.cpu_bound(raise_unpicklable_exception))
 
     await user.open('/')


evnchn:
https://github.com/evnchn/nicegui/actions/runs/18365797268

There we go.

falkoschindler:
Nice! 🙂
Quick question: Could we _always_ use `inspect.iscoroutinefunction`? It should be defined [since Python 3.5](https://docs.python.org/3.14/library/inspect.html#inspect.iscoroutinefunction).

evnchn:
https://github.com/zauberzeug/nicegui/pull/5243/commits/e903ec1eb023ced253300e20c4bf211084b31c32. 

Related to bug in CPython (https://github.com/python/cpython/issues/94924), and it is a hotfix for `unittest.mock.patch` to work. 

Note: 
1. I am quoting from https://github.com/fastapi/fastapi/pull/14022 where I copy-pasted their code. 
2. We used `unittest.mock.patch` once in our code: At https://github.com/zauberzeug/nicegui/blob/1bb276c3a2cf9b25f569319a07dbd2946a2225ad/tests/test_fullscreen.py#L31

burnpanck:
FYI: The `git` override for `vbuild` may not work for downstream libraries. Specifically, it works for my applications which I manage using `pdm`, where I can apply an "override" during package resolution to select the git version. But for `pip`, which also has a  `-c override.txt` option where one can specify a `git+https://` package spec, it fails. The problem is that the `vbuild` wheel that gets automatically built by `pip` when installing from a git url will pretend to be version `0.0.0`, which is incompatible with the version requirement `vbuild>=0.8.2` given in the `nicegui` package metadata.

evnchn:
@burnpanck We learnt the hard way and proceeded to in-house vbuild.

PR #5236, opened 2025-10-08
Use `pytest-rerunfailures` to re-run flaky Mermaid tests

evnchn:
### Motivation

#5232 (make Mermaid tests not flaky) harder than it looks. We may have to live with flakiness. This PR does exactly that. 

### Implementation

This pull request introduces improvements to the test suite's reliability by addressing flaky tests related to Mermaid rendering. The main changes involve adding the `pytest-rerunfailures` plugin and marking relevant tests to automatically rerun on failure, reducing false negatives in CI.

**Test reliability improvements:**

* Added `pytest-rerunfailures` as a dependency in both `pyproject.toml` and `tests/requirements.txt` to enable automatic rerunning of flaky tests. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R78) [[2]](diffhunk://#diff-27c5febcc9ce93e8e5341f263b7d68962733b8a2a86c583723d025a1db39d7a3R3)
* Marked Mermaid-related tests in `tests/test_markdown.py` with `@pytest.mark.flaky(reruns=2)` to allow up to two reruns on failure. [[1]](diffhunk://#diff-f0f45280def5f73305be099b06f97da1b2a3da1f964471c94dd1c58cac6fe799R28) [[2]](diffhunk://#diff-f0f45280def5f73305be099b06f97da1b2a3da1f964471c94dd1c58cac6fe799R62)
* Applied a module-level `pytest.mark.flaky(reruns=2)` marker to all tests in `tests/test_mermaid.py` for consistent rerun behavior.
* Imported `pytest` in relevant test files to support the new flaky test markers.

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests have been added (or are not necessary).
- [x] Documentation has been added (or is not necessary).

PR #5249, opened 2025-10-09
`pytest-xdist` support (parallelized tests!)

evnchn:
### Motivation

I want a fast dev experience. 20 minutes to wait for the pipeline in GitHub Action isn't what I want. 

**Real motivation: Using 1 core out of 10 in my laptop is unacceptable...**

### Implementation

This pull request adds support for running parallelized tests using `pytest-xdist`, allowing tests to be distributed across multiple workers for faster execution. The main changes involve updating test infrastructure to handle parallelism safely, including dynamic port assignment and ensuring that shared resources are not accessed concurrently by multiple workers.

Test parallelization and infrastructure updates:

* Added `pytest-xdist` to the dependencies in both `pyproject.toml` and `tests/requirements.txt` to enable parallel test execution. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R83) [[2]](diffhunk://#diff-27c5febcc9ce93e8e5341f263b7d68962733b8a2a86c583723d025a1db39d7a3R3)
* Modified the GitHub Actions workflow (`.github/workflows/test.yml`) to run `pytest` with the `-n auto` flag, enabling automatic parallelization based on available CPU cores.

Parallel test safety and resource management:

* Updated the `Screen` class in `nicegui/testing/screen.py` to respect the `NICEGUI_SCREEN_TEST_PORT` environment variable, allowing each test worker to use a unique port and avoid conflicts.
* Introduced a `_get_worker_id` function in `nicegui/testing/screen_plugin.py` to determine the current pytest-xdist worker ID, which is used to offset ports and manage resources per worker.
* Adjusted the `nicegui_remove_all_screenshots` fixture to only remove screenshots in the main worker (worker 0), preventing race conditions when cleaning up shared directories.
* Modified the `screen` fixture to assign a unique port to each worker by incrementing the base port with the worker ID, ensuring parallel test servers do not conflict.

### End results

`pytest` stage went from 16 minutes to 5 minutes. 

### Progress

  - Tests are no longer flaky provided use of retryfailures. 

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,11 +49,11 @@ jobs:
         uses: nanasess/setup-chromedriver@v2.3.0
       - name: pytest (parallelized)
         run: |
-          poetry run pytest -n auto --ignore=tests/test_storage.py --ignore=tests/test_download.py --ignore=tests/test_run.py --ignore=tests/test_binding.py --ignore=tests/test_upload.py --ignore=tests/test_user_simulation.py --ignore=tests/test_event.py
+          poetry run pytest -n 2 --ignore=tests/test_download.py --ignore=tests/test_storage.py
       - name: pytest (non-parallelized)
         if: always()
         run: |
-          poetry run pytest tests/test_storage.py tests/test_download.py tests/test_run.py tests/test_binding.py tests/test_upload.py tests/test_user_simulation.py tests/test_event.py
+          poetry run pytest tests/test_download.py tests/test_storage.py
       - name: Upload failed screenshots
         if: failure()
         uses: actions/upload-artifact@v4


evnchn:
Pretty sure this was closed by the GitHub outage...

evnchn:
So there's also another branch in my fork which adds #5236 globally onto all tests, and I have got one 100% green run at https://github.com/evnchn/nicegui/actions/runs/18383137104 based off https://github.com/zauberzeug/nicegui/pull/5249/commits/6b0f15a9e28efa1853256d7bff12cecb12bcb411

But I don't want to deprecate `screen_plugin.DOWNLOAD_DIR` since I want to keep the deprecation to `Screen.PORT` which is global to all pytest.

rodja:
We would really like to avoid reruns. We must find the issues or live with the flakyness until we find them. Why is the flakyness higher than when running without parallel tests?

evnchn:
OK then. We will face flakiness directly. 

Short-term action will explore why in another branch. 

When ready, to revert https://github.com/evnchn/nicegui/pull/44 / make another PR.

evnchn:
Gor you an answer already @rodja 

image

Looks familiar? 

You recall I said something privately about running `gc.collect()` before and after the `yield` in the fixture to catch ResourceErrors early, as well as an unclosed Event Loop? 

That's exactly what [GC before and after test (hope to catch ResourceWarning earlier)](https://github.com/zauberzeug/nicegui/commit/53dd53c4784eb3824b55f3ac51c95a743b5c326a), [Drop using asyncio.run in pytest](https://github.com/zauberzeug/nicegui/commit/2afac8b62b6761ff47345b4ce912a36f922baffd) and [Stop using asyncio.run in test_upload.py](https://github.com/zauberzeug/nicegui/commit/d4b6c4c829068fc4219ae94b73118c6ab9ac5e71)
was about in https://github.com/zauberzeug/nicegui/compare/main...evnchn:nicegui:clean-test-dependencies-relock

So basically I can apply those patches over at this branch and we should be able to achieve retry-free and flaky-free parallized tests.

evnchn:
That can't be right (de-parallelizing way more tests than I expected) and totally unlike https://github.com/zauberzeug/nicegui/compare/main...evnchn:nicegui:clean-test-dependencies-relock where things just "works". 

I may not be at the core issue but I still want to try things and see how it goes. Worst case we revert back to https://github.com/zauberzeug/nicegui/pull/5249/commits/ae1a35f96f9f655a9e055947821959d05a0521a6

evnchn:
https://github.com/evnchn/nicegui/actions/runs/18463882807

Timeout from 4 to 15 seconds, and selenium is still timing out. At least that only 2/6 Python versions suffer from an error, now that I have removed most of the most flaky tests under parallelization. 

I don't forsee full completion under parallelization with no retries when using all 4 cores in the GitHub pipelin. It's just way too stretched. Maybe it will work with 3, better when 2. 

@rodja nevertheless rerunning individual failed tests in hope of getting every test to at least pass once is not good and I got rid of that logic. But how about a mixed approach using a mix of the below criteria?

- at least 1 Python version need to fully complete all tests under parallelization?
- for Python versions which failed the parallelized tests, can we accept it if we re-ran the suite, parallelized, and it passed?
- for Python versions which failed the parallelized tests, can we accept it if we took just the failing tests, ran it in non-parallized manner successfully 5 times in a row?

evnchn:
https://github.com/evnchn/nicegui/actions/runs/18463882807

All-pass

evnchn:
I can't really come to a point, at least in my current understanding and the current state of the repo, where:

1. Tests are reliable, AND
2. I do not ignore quite a lot of tests in the parallelized part. 

I think #5258 would help with the situation, and so maybe let's wait until that is merged before we grind on this.

evnchn:
Well https://github.com/zauberzeug/nicegui/pull/5249/commits/45cf058340b23bc8e3036e85ec492528b95f8091 works with 2 workers, skipping only the needed tests, so you _can_ take that as a win. 

But still I have 4 cores on GitHub Actions and want to use all though.

evnchn:
No. We will never reach the stability of the tests using 4 workers, since there's not only the 4 Python process, but 4 headless Chromium instances, and that's just too much for the 4 vCPUs. 

If we can only achieve a marginal speed increase going from 1 to 2 workers then I don't think this is worth it. 

If you do agree we can close this PR as it is. 

---

A fast pipeline is still desirable, so maybe we can pivot by having another GitHub workflow which runs parallelized tests with retries on just the fastest Python version available (that would be Python 3.13 as of writing, considering the faster dependencies installation)

That also have the benefit that the tests results arrives sooner, since we're not having the startup tests in that extra-fast pipeline. 

And, it can help manage the hardware installation when I eventually consider a self-hosted worker for the tests at my branch.

rodja:
Yes @evnchn, executing tests in parallel is super tricky. There are all kind of side effects (like storage dir, cpu strain, ...). It is very unfortunate to have such a long running test pipeline -- and we should further look into optimizations.

PR #5251, opened 2025-10-09
Split screenshots and upload failed ones in CI pipeline

falkoschindler:
### Motivation

In PR #5234 @evnchn extended the test workflow to automatically collect screenshots of failed tests and to upload them as artefacts. I wondered if it would be easier to directly store them in two separate dictionaries depending on the individual test result, so the test pipeline only needs an additional upload step.

**We need to decide which approach we prefer, PR #5234 or this PR.**

### Implementation

Apart from some trivial changes to handle two screenshot directories instead of one, we need an extra `pytest_runtest_makereport` fixture to store the test outcome. Later we can access this information via
```py
test_failed = hasattr(request.node, 'rep_call') and request.node.rep_call.failed
```
and pass the information to the `screen_.shot` method.

### Progress

- [x] An extra pytest has been added for testing.
- [x] cleanup
  - [x] The extra pytest has been removed before merging.
  - [x] All tests and all Python versions are executed again (revert 886e03fb964fc2fc253f4eb1d74f6801284ef9b8 and 526e67ae9fe2b626a591ba096ae017bab0cd2cad)

### DIFF
--- a/tests/test_label.py
+++ b/tests/test_label.py
@@ -18,13 +18,3 @@ def test_text_0(screen: Screen):
 
     screen.open('/')
     screen.should_contain('0')
-
-
-def test_failure(screen: Screen):
-    """This test should fail. We will remove it later."""
-    @ui.page('/')
-    def page():
-        ui.label('Hi!')
-
-    screen.open('/')
-    screen.should_contain('This text does not exist')


evnchn:
I like yours better. Yours is basically what I want to do in the very very beginning, but I & Copilot both cannot figure out the `@pytest.hookimpl`

nicegui/testing/screen.py
```diff
@@ -29,7 +28,8 @@
 class Screen:
     PORT = 3392
     IMPLICIT_WAIT = 4
-    SCREENSHOT_DIR = Path('screenshots')
```
evnchn, line 32: Is this considered public, though... Maybe we can still save a combined screenshots at the old path, and have 2 new paths (green and red)

falkoschindler: Good catch! What about keeping a single directory and add a suffix if a test failed:

test_hello_world.png
vs.
test_hello_world.fail.png

Then we can specify the upload path like this:
```
path: 'screenshots/*.fail.png'
```

evnchn: For this I think is fine in the 99.999% use cases

Technically considerable as a "breaking change" but accounting for the most popular usage (`fname.split('.')[0]`) we can still find the screenshot by the name. 

The code which is broken by this are those who (1) care about performance and do not use `split` and (2) used `fname.rpartition('.')[0]` instead of `fname.partition('.')[0]`, which since pytest names cannot have `.` in them I call it user issue.

evnchn: Retrospect, we never condone the scraping and computer-processing of the `screenshots/` output, so we can do whatever.

PR #5252, opened 2025-10-10
Document CSS layers feature in NiceGUI v3.0

amks1:
Added documentation for CSS layers feature in NiceGUI v3.0, explaining how to override Quasar's rules and providing an example of defining custom styles.

### Motivation

The documentation was not clear about how to override default styles using custom css. Regular classes and even using `!important` do not work after v3. Discussed in https://github.com/zauberzeug/nicegui/discussions/5240.

### Implementation

Added a description and demo to the doc.

### Progress

### DIFF
--- a/website/documentation/content/section_styling_appearance.py
+++ b/website/documentation/content/section_styling_appearance.py
@@ -118,24 +118,30 @@ doc.text('', '''
 
 
 @doc.demo('CSS Layers', '''
-    Since NiceGUI 3.0 you can override Quasar's rules by defining your CSS rules in a layer
-    that takes precedence over Quasar's.
-    The layers in order of priority from high to low are:
-    `theme`, `base`, `quasar`, `nicegui`, `components`, `utilities`, `overrides`, `quasar_importants`.
-    Layers allow you to control the priority of whole groups of styles,
-    independent of selector specificity or `!important`.
+    NiceGUI defines the following CSS layers (in order of increasing priority):
+    "theme", "base", "quasar", "nicegui", "components", "utilities", "overrides", and "quasar_importants".
 
-    So to override a button's color, for example, you could simply do the following:
+    You don't need to put your custom CSS into layers for basic styling.
+    However, to override Quasar's `!important` rules, you should define your CSS in an appropriate layer:
+    use "components" for component-specific styles or "utilities" for utility classes,
+    depending on the purpose of your custom styles.
+    Note that you need to use `!important` in your custom styles
+    because Quasar defines most of its CSS with `!important`,
+    which would otherwise take precedence.
+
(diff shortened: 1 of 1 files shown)


website/documentation/content/section_styling_appearance.py
```diff
@@ -117,6 +117,27 @@ def handle_style(e: events.ValueChangeEventArguments):
 ''')
 
 
+@doc.demo('CSS Layers', '''
+    Since NiceGUI 3.0 you can override Quasar's rules by defining your CSS rules in a layer
+    that takes precedence over Quasar's.
+    The layers in order of priority from high to low are:
+    `theme`, `base`, `quasar`, `nicegui`, `components`, `utilities`, `overrides`, `quasar_importants`.
+    Layers allow you to control the priority of whole groups of styles,
```
evnchn: Content isn't that accurate. Check #5171 and https://developer.mozilla.org/en-US/docs/Web/CSS/@layer again. 

> The declaration order matters. The first declared layer gets the lowest priority and the last declared layer gets the highest priority. However, the priority is reversed when the [!important](https://developer.mozilla.org/en-US/docs/Web/CSS/important) flag is used.

And so you need `!important` for the reversed ordering which let's you win Quasar (I added a comment below)

website/documentation/content/section_styling_appearance.py
```diff
@@ -117,6 +117,27 @@ def handle_style(e: events.ValueChangeEventArguments):
...
+    independent of selector specificity or `!important`.
+
+    So to override a button's color, for example, you could simply do the following:
+''')
+def css_layers():
+    ui.add_css('''
+        @layer base {
```
evnchn: Tailwind CSS uses `utilities` so I recommend everyone to use that layer if they want same functionality as 2.X. 

(Note: You can feel free to do whatever if you don't have a strong desire for backwards compatibility, but the docs we should suggest the general approach)

website/documentation/content/section_styling_appearance.py
```diff
@@ -117,6 +117,27 @@ def handle_style(e: events.ValueChangeEventArguments):
...
+           .my-custom-button {
+               background-color: red;
```
evnchn: Need `!important` here.

website/documentation/content/section_styling_appearance.py
amks1: Sorry my bad for misunderstanding this, I got the wrong info from Chatgpt. So the reason why my example worked in https://github.com/zauberzeug/nicegui/discussions/5240 is because of the !important flag in the rules defined in Quasar? If !important is present for a rule in any layer, the order for that rule is reversed? 🤔

PR #5254, opened 2025-10-10
Let pytest run first in the pipeline, startups later.

evnchn:
### Motivation

There are various issues with running the `test_startup.sh` before the `pytest`, in ascending order of seriousness:

1. Delayed to see the results of the pytest
2. Installation of the example's requirements.txt, **defeats the purpose of having pinned dependencies in `poetry.lock`**, as we can see in https://github.com/zauberzeug/nicegui/actions/runs/18400412564/job/52428146319 that packages indicated in `poetry.lock` are then uninstalled in place of other ones. 

We can tolerate with point 1 but point 2 definitely needs improvement. 

This could be the cause of the many errors in #5242. Hence this PR. 

**Real motivation: I squeezed pytest into 5 minutes in #5249, but I need to wait 10 minutes for `test_startup.sh`?**

### Implementation

Move stage `test startup` to the back. 

### Progress

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,8 +50,6 @@ jobs:
           poetry run pip install -r tests/requirements.txt
           # try fix issue with importlib_resources
           poetry run pip install importlib-resources
-      - name: test startup
-        run: poetry run ./test_startup.sh
       - name: setup chromedriver
         uses: nanasess/setup-chromedriver@v2.3.0
       - name: pytest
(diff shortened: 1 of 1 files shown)


evnchn:
Startups still messes with the cache... We see Py3.14 always need an extra 5 minutes to install dependencies. 

https://github.com/zauberzeug/nicegui/actions/runs/18411862410/job/52466073580

2 things: 

1. Can the test_startups be done in another pipeline, or another stage in the test pipeline? (We dont need all 6 Python versions, no?)
2. Can we skip `test_startups` if we did not change the website, `main.py` nor anything in `examples`?

falkoschindler:
> Can the test_startups be done in another pipeline

I think to, yes. Theoretically there could be a problem with one Python version only, but I can't remember that this ever happened for the startup tests.

> Can we skip `test_startups` if...

No, I don't think so. Startup tests sometimes break after changing UI elements or some core parts of the library.

PR #5255, opened 2025-10-10
Move vbuild into our repository

falkoschindler:
### Motivation

The latest release of vbuild doesn't support Python 3.14 (yet) even though the problem is fixed on main.
The release of NiceGUI 3.0.4 is currently blocked because PyPI refuses to accept a direct dependency to a Git repository.

### Implementation

Since vbuild is only a single file of Python code, this PR simply moves it into our repository. To silence linters and type checkers, I drastically cleaned up and refactored the code (545 lines down to 120). I hope everything is still working, but tests are green and Single-File Components like `ui.plotly` and `ui.joystick` are working.

### Progress

### DIFF
--- a/nicegui/vbuild.py
+++ b/nicegui/vbuild.py
@@ -22,7 +22,7 @@ class VBuild:
         self.script = create_vue_component(filepath.stem, f'#tpl-{name}', parser.script)
 
 
-class VueParser(HTMLParser):
+class VueParser(HTMLParser):  # pylint: disable=abstract-method  # pylint assumes there is an abstract ``error`` method
 
     def __init__(self, filepath: Path):
         HTMLParser.__init__(self)


evnchn:
Thanks @falkoschindler for doing what I want to do, but lack the time. Quite busy so high-level comments. 

**Definitely non-backwards-compatible** since the `pscript` thing, as well some some other switches and knobs, are gone. 

But, I think it is fair since:

1. Our subset targets 99.999% of a "normal user"'s use case. After all, users are loading `.js` and `.vue` files via NiceGUI, and IDEs will be **very angry** if you do any `pscript` or anything like that. 
2. `pscript` has come caveats to begin with (https://pscript.readthedocs.io/en/latest/intro.html)
3. Users are not left in the dark. They are able to use `vbuild` standalone to build existing stuff which requires it into workable Vue component, then include that component into existing NiceGUI code. 

Generally speaking I'm fine with this PR since we have NiceGUI tests to back things up. And we are delivering non-backwards-compatible code in #5239 due to "not too many users have used this" anyways. 

Suggestion: 

1. For a better user experience, maybe we detect the attempt to use the public `vbuild` specific features and log warning appropriately?
2. If you want more confidence, consider to copy applicable tests from https://github.com/manatlan/vbuild/tree/master/tests (warning: not all of their tests pass, so take it with a big grain of salt)

We can do these in another PR though, to not block 3.0.4

falkoschindler:
Thanks, @evnchn! I agree.

- Technically we might break things, but non of those features like pscript have been documented anywhere. vbuild is an internal tool and we can replace it if needed.
- I thought about copying the original tests. But if we do that, we need to filter, clean and update the code as well. And if we don't _understand_ what the tests are actually testing, there is no other safety net. So we might be better off by creating tests from scratch.

evnchn:
OK so here's where the clarifications on the refactoring will happen. 

I have every reason to believe that `mkClassicVueComponent` became `create_vue_component`, @falkoschindler Can you confirm? 

I need this info because the gist in https://github.com/zauberzeug/nicegui/pull/4673#issuecomment-2852867163 was to cut that function out of the equation.

falkoschindler:
@evnchn Yes, `mkClassicVueComponent` became `create_vue_component`. Since we only need a single function for creating Vue components, the "classic" could go.

PR #5257, opened 2025-10-10
Modernize our `vbuild` to use ESM import rather than inlining the JS

evnchn:
### Motivation

The mastermind is and always has been @xaptronic (see #4673), but things went stale and that PR becomes harder and harder to merge as time marches onwards. 

The potential advantages are at https://github.com/zauberzeug/nicegui/pull/4673#issue-3025392294 but you should read above the first `---` because the rest are about the refactoring of `vbuild` and no AI can beat @falkoschindler (look at #5255)

### Implementation

The logic is 3:

1. Cut the middleman `create_vue_component` between `self.script` and `parser.script` (check https://github.com/zauberzeug/nicegui/pull/4673#issuecomment-2852867163)
2. Use `js_imports` instead of `vue_scripts`
3. Serve `vue_components` from `_get_component`

### Progress

  - [x] The pytest should pass, obviously
  - [x] We _really_ need to have our own set of pytest to test whether our Vue integration works. We already winged it in #5255, and we don't want to double-wing it. 

### DIFF
--- a/tests/test_vbuild.py
+++ b/tests/test_vbuild.py
@@ -27,9 +27,7 @@ def test_template_only():
             <h1 data-TEST>Hello, World!</h1>
         </script>
     ''', css='''
-    ''', js='''
-        var TEST = Vue.component(\'TEST\', {template:"#tpl-TEST",});
-    ''')
+    ''', js='')
 
 
 def test_template_with_style():
@@ -48,9 +46,7 @@ def test_template_with_style():
         </script>
     ''', css='''
         h1 {color: red; }
-    ''', js='''
-        var TEST = Vue.component(\'TEST\', {template:"#tpl-TEST",});
-    ''')
+    ''', js='')
 
 
 def test_template_with_scoped_style():
@@ -69,9 +65,7 @@ def test_template_with_scoped_style():
         </script>
     ''', css='''
         *[data-TEST] h1 {color: red; }
-    ''', js='''
-        var TEST = Vue.component(\'TEST\', {template:"#tpl-TEST",});
-    ''')
+    ''', js='')
 
 
 def test_template_with_script():
@@ -94,13 +88,13 @@ def test_template_with_script():
         </script>
     ''', css='''
     ''', js='''
-        var TEST = Vue.component(\'TEST\', {template:"#tpl-TEST",
-                methods: {
-                    hello() {
-                        alert('Hello, World!');
-                    }
+        export default {
+            methods: {
+                hello() {
+                    alert('Hello, World!');
                 }
-            });
+            }
+        }
     ''')
 
 


evnchn:
@falkoschindler Strong argument for this PR: Everything outside of `{}` is ignored before this PR, so cannot import vue stuff. `import { ref } from 'vue'` won't work since it's not ESM but even `const { ref } = Vue` won't work.  

image

falkoschindler:
@evnchn Ok, but what's the status? I thought you were still working on pytesting our vbuild.

evnchn:
I didn't really have the time for that for now though...

evnchn:
GitHub Copilot splat out like 10 tests, then I boiled it down in https://github.com/zauberzeug/nicegui/pull/5257/commits/a7de3db0da04389470f0ebbca44e2d68567a239b 

(Oh I wish the output is as condensed straight off GitHub Copilot but no...)

Should be pretty good to avoid accidental regression. No idea the code coverage though.

falkoschindler:
The test looks pretty cryptic. I'd rather create tests like this:
```py
import tempfile
import textwrap
from pathlib import Path
from nicegui.vbuild import VBuild

TMP_FILE = Path(tempfile.mktemp())

def test_template_only():
    TMP_FILE.write_text(textwrap.dedent('''\
        <template>
            <div>
                <h1>Hello, World!</h1>
            </div>
        </template>
    '''), encoding='utf-8')
    build = VBuild(TMP_FILE)
    assert build.html == textwrap.dedent(f'''\
        <script type="text/x-template" id="tpl-{TMP_FILE.stem}"><div data-{TMP_FILE.stem}>
                <h1>Hello, World!</h1>
            </div></script>''')
    assert build.style == '\n'
    assert build.script == f'var {TMP_FILE.stem} = Vue.component(\'{TMP_FILE.stem}\', {{template:"#tpl-{TMP_FILE.stem}",}});'
```

Similarly we can test scenarios like
- component with JavaScript
- component with CSS
- component with scoped CSS
- raise if more than one template,
- ...

Possible improvements to the test setup:
- Add newlines in `VBuild` such that `build.html` results in a cleaner layout.
- Extract a helper function to take care of writing the temp file and dedenting multi-line strings.
- Use a temp file with a fixed stem like "TEST" to avoid the need for f-strings (`id="tpl-TEST"` instead of `id="tpl-{TMP_FILE.stem}"`).

Maybe this is something for a separate PR _before_ continuing with this PR?

evnchn:
My concern with explicit `==` is that the tests will tie down the exact output of our VBuild and when we want to change _anything_ the tests have to be changed. 

Especially when in the given context: There is no one correct output. Case in point: 

> Add newlines in VBuild such that build.html results in a cleaner layout. 

Nevertheless I agree with you still: The tests right now are cryptic, and we could do with more commenting in retrospect, but using `in` it captures the _behaviour_ of the VBuild which, hopefully, we don't have to change the tests often to keep in sync. 

Feel free to takeover by the way. I'm not really expert in pytest.

falkoschindler:
@evnchn See PR #5393. I chose to stick with strict `==` comparisons for simplicity. When we notice regressions for whatever reason in the future, we can decide whether to loosen or adjust the tests.

evnchn:
In light of #5393 I propose to get rid of https://github.com/zauberzeug/nicegui/commit/a7de3db0da04389470f0ebbca44e2d68567a239b via force-push, then when that is merged we merge main and see how the tests fair. 

Retrospect kinda not the best idea to add tests in the same PR which you are _changing_ something which was untested beforehand.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5257/commits/c7cf14209aea05cac3641f8821440538914b6e24 demonstrates pretty clearly what the PR is about: the `js` becomes a ES module which can be imported.

PR #5258, opened 2025-10-10
Recreate `poetry.lock` after deletion

evnchn:
### Motivation

The dependencies in `poetry.lock` are OLD.

**Real motivation: #5242 pipeline's not happy, but this might just work, as we have #5254**

### Implementation

- Delete `poetry.lock`
- Run `poetry lock`

NOTE: Unlike #5242 I don't have to explicitly excluded `propcache` version 0.4.0 since we have 0.4.1 already. 

### Progress

- [x] The implementation is complete. (We'll look at the pipeline)

### DIFF
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -342,7 +342,7 @@ def test_tab_storage_holds_non_serializable_objects(screen: Screen):
     screen.wait(0.5)
 
 
-def test_user_storage_is_pruned(screen: Screen):
+async def test_user_storage_is_pruned(screen: Screen):
     @ui.page('/')
     def page():
         ui.label(f'clients: {len(Client.instances)}')
@@ -367,7 +367,7 @@ def test_user_storage_is_pruned(screen: Screen):
 
     screen.close()
     Client.prune_instances(client_age_threshold=0)
-    asyncio.run(nicegui.prune_user_storage(force=True))
+    await nicegui.prune_user_storage(force=True)
     assert len(Client.instances) == 0
     assert len(app.storage._users) == 0
 
--- a/tests/test_upload.py
+++ b/tests/test_upload.py
@@ -1,4 +1,3 @@
-import asyncio
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


evnchn:
https://github.com/evnchn/nicegui/actions/runs/18412996019

Pipeline's still red. 

This can be adopted together with #5249 but we're not solving the actual problem.

evnchn:
image

Can't be right. I think our pipeline dependencies are still kinda messed up. 

https://github.com/evnchn/nicegui/tree/refs/heads/clean-test-dependencies

evnchn:
No actually more clearly since we haven't arranged the dependency in #5261 it will always install asyncio-timeout 4.0.3 forcefully into a Poetry environment which expects 5.0.1 and everything's off the rails from there. 

So this PR is actually blocked by #5261

evnchn:
This PR is un-mergable without "gc.collect() before and after yield" and "No more asyncio.run", so we have no choice but to expand the scope of this PR to also include those 2 sets of changes.

evnchn:
Pipeline's green 👍 

https://github.com/evnchn/nicegui/actions/runs/18471917995

Recommended action: Review and merge which reduce difference between test environment and user's typical environment.

evnchn:
I ran that particular test 100 times in my local machine and can't reproduce it. (Using `pytest-repeat`)

falkoschindler:
Oh, I see! On my machine `test_storage_access_in_binding_function` is flaky on main, especially when running _all_ storage tests. So it doesn't have anything to do with this PR.

Regarding the garbage collection: Why exactly do we need it? I don't find any explanation in this PR.

evnchn:
> Regarding the garbage collection: Why exactly do we need it?

We need it because the nasty `ResourceWarning` is caused when the event loop becomes GC-ed, which occurs in the tests which uses `asyncio.run`, but then the error only shows when the test progresses into another test's element init, leading to ResourceWarning on a new object's `__init__`. 

In short: Avoid the leak of event loop to shift blame into another irrelevant test.

rodja (approved):

PR #5266, opened 2025-10-13
Apply margin to all `pre` in `.nicegui-markdown`

### LINKED ISSUE #5264: Missing spacing in ui.code without language specification
### First Check

### Example Code

```python
#!/usr/bin/env python3
from nicegui import ui

ui.code('''
    from nicegui import ui

    ui.label('Code inception!')

    ui.run()
''').classes('w-full')

ui.code('''
    from nicegui import ui

    ui.label('Code inception!')

    ui.run()
''', language='').classes('w-full')

ui.run()
```

### Description

Occasionally, language highlighting isn't desired, so I specify `language=''`. This works, but then the margins are missing.

### NiceGUI Version

3.0.4

### Python Version

3.11.13

### Browser

Firefox

### Operating System

Linux

### Additional Context

Occasionally, language highlighting isn't desired, so I specify `language=''`. This works, but then the margins are missing.

evnchn:
### Motivation

Fix #5264, where the margin is inconsistent when `language=''`

### Implementation

* Changed the selector from `.nicegui-markdown .codehilite pre` to `.nicegui-markdown pre` and updated the margin to `0.5rem 0.5rem` to provide both horizontal and vertical margin to both codehilite and non-codehilite code blocks. 
* Removed unnecessary padding from `.nicegui-code .codehilite`. 

### Visual difference

Before:

image

After: 

image

### Progress

### DIFF
--- a/nicegui/static/nicegui.css
+++ b/nicegui/static/nicegui.css
@@ -196,7 +196,7 @@
   border: 1px solid #8884;
 }
 .nicegui-markdown pre {
-  margin: 0.5rem 0.5rem;
+  margin: 0.5rem 0;
 }
 .nicegui-markdown .mermaid-pre > .mermaid {
   display: none;
@@ -256,6 +256,9 @@ h6.q-timeline__title {
   border: 1pt solid rgba(127, 159, 191, 0.15);
   box-shadow: 0 0 0.5em rgba(127, 159, 191, 0.05);
   border-radius: 0.25rem;
+}
+.nicegui-code pre {
+  padding: 0 0.5rem;
 }
 .nicegui-matplotlib > svg,
 .nicegui-pyplot > svg {


falkoschindler (changes requested):
Thanks, @evnchn!
Unfortunately the result is not 100% correct. I'm testing with this code:
````py
ui.code('print("Code with language")')
ui.code("print('Code without language')", language='')
ui.markdown('''
    ```python
    print("Code with language")
    ```
''').classes('border')
ui.markdown('''
    ```
    print("Code without language")
    ```
''').classes('border')
````

Image

The border around `ui.markdown` is just for visualizing the layout. Usually `ui.markdown` is _not_ meant to be bordered.

What we want:
- `ui.code` needs space around its text.
- Code in `ui.markdown` should have vertical space (like plain text in `ui.markdown`), but no horizontal space (again, like plain text in `ui.markdown`).

The problem with the current implementation can be seen on the documentation page: All demo code snippets have a bit more spacing now.

I'll have a look and try to find a solution that works. Maybe we only have to adjust the `.nicegui-markdown .codehilite pre` selector because code without a language isn't wrapped in a codehilite div - as I just learned.

evnchn:
Glad you got it. Should have looked more carefully at other parts of the code which uses the style.

PR #5271, opened 2025-10-13
Make the style of `ui.separator` be different in different orientation

### LINKED ISSUE #5270: `ui.separator().props("vertical")` takes `width: 100%` instead of `1px` with new `nicegui.css`
### First Check

### Example Code

```python
from nicegui import ui

ui.separator().props("vertical")

ui.run()
```

### Description

Hi,

I noticed that my `ui.separator.props("vertical")` elements stretched to occupy all the horizontal space instead of just 1px after updating to NiceGUI 3.

It seems to stem from https://github.com/zauberzeug/nicegui/commit/f571ff197b98014bd8f69be9d56ae582fcc79f31#diff-5d9345fb48521f49b1b64d5d6e3cec49ab5a59b10621f21532e1d87613e3e69bR43

Can you advise on how to re-enable the "vertical" feature without having to add a style like `ui.separator().props("vertical").style("width: 1px")`?

Thanks!
Tim

### NiceGUI Version

3.0.4

### Python Version

3.12.11

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

python-and-novella:
### Motivation
Fix #5270, where the style of `ui.separator().props('vertical')` doesn't work as expected in NiceGUI 3.0.0. above.

### Implementation
change
```css
.nicegui-separator {
  width: 100%;
}
```
into
```css
.q-separator--vertical.nicegui-separator {
  width: 1px;
}
.q-separator--horizontal.nicegui-separator {
  width: 100%;
}
```

### Progress

### Test

Testing it by:
```python3
from nicegui import ui

def index():
    with ui.column():
        ui.button('1')
        ui.separator()
        ui.button('2')
    with ui.row():
        ui.button('1')
        ui.separator().props('vertical')
        ui.button('2')

ui.run(
    root=index
)
```
It should work as following:
image

### DIFF
--- a/nicegui/static/nicegui.css
+++ b/nicegui/static/nicegui.css
@@ -215,10 +215,7 @@
 .nicegui-link:active {
   opacity: 0.8;
 }
-.q-separator--vertical.nicegui-separator {
-  width: 1px;
-}
-.q-separator--horizontal.nicegui-separator {
+.nicegui-separator.q-separator--horizontal {
   width: 100%;
 }
 .nicegui-aggrid,


python-and-novella:
I'm sorry for that I have no time to write unit test. And I have no idea of writing unit test about this.
There isn't a unit test about `ui.separator`, I think the unit test is necessary.

evnchn:
I don't think that's how the tests are supposed to work. (You're more like checking whether the styles which we applied are applied, which they sure are, from the tests which tests that part)

Maybe you can test the aspect ratio of the rendered divider, instead.

python-and-novella:
It is difficult for me without learning carefully. I will try it when I am free.
I am glad for that  if someone can make unit test complete.

evnchn:
Honestly I am not that good at pytest either, but looking back I think we can get away with it. 

I didn't have any tests in #5266 either.

falkoschindler:
Thanks for the pull request, @python-and-novella!
I haven't had the chance to think about it, yet. But either way, I don't think pytests are necessary. We don't check the visual appearance for most elements, so we don't need to start here. If we would start checking aspect ratios, I would also need to check color, shadow, exact dimensions etc. to prevent regressions. This isn't feasible.

falkoschindler (approved):
Thanks for the pull request, @python-and-novella!

I simplified it a bit: We don't have to set the width of 1px for vertical separators. This is already done by Quasar. But limiting our rule to horizontal separators is a great fix. 👍🏻

PR #5274, opened 2025-10-14
introduce code transformation to simplify writing ui.sub_pages demos

rodja:
### Motivation

Writing demos for Single Page Applications (SPAs) with `ui.sub_pages` has been cumbersome and unnatural. Demo code doesn't match what users should actually write in their own applications:

1. Unnatural code patterns: Demo code used `FakeSubPages` and manual workarounds (sub_pages.link(), FakeArguments, etc.) instead of the clean, straightforward API (ui.link(), ui.sub_pages())
2. Quite some code duplications which makes it hard to find areas where the rendered demo didn't match what users should write

### Implementation

This PR introduces an automatic code transformation system that allows demo authors to write natural, idiomatic NiceGUI code while still supporting browser-based demo execution:

1. New transformation infrastructure (`website/documentation/code_extraction.py`):  
   * Transform `ui.sub_pages()` → `FakeSubPages.init()`, `ui.link()` → `pages.link()`
   * Handles nested sub-pages, URL parameters, query strings, and chained methods  
   * Uses special `# TRANSFORM` comments for complex cases (e.g., parameterized routes)  
   * test cases covering transformation of sub-pages, nested routes, URL parameters, TRANSFORM hints  
1. Demo execution engine (`website/documentation/demo.py`):  
   * Displays non-transformed code block
   * Executes transformed code with automatic root() function detection  
1. Updated sub-pages documentation (`website/documentation/content/sub_pages_documentation.py`):  
   * All demos now use natural ui.link() and ui.sub\_pages() API  
   * Removed \~140 lines of workaround code  

### Progress

  - [ ] The transformation itself is AI generated hand I have not yet found the time to clean it up

### DIFF
--- a/tests/test_demo_transformation.py
+++ b/tests/test_demo_transformation.py
@@ -1,0 +1,320 @@
+from textwrap import dedent
+
+from website.documentation.code_extraction import get_full_code, transform_for_demo_execution
+
+
+def prepare(text: str) -> str:
+    return dedent(text).strip() + '\n'
+
+
+def test_simple_sub_pages():
+    assert transform_for_demo_execution(prepare('''
+        from nicegui import ui
+
+        def root():
+            ui.sub_pages({'/': main, '/other': other})
+
+        ui.run(root)
+    ''')) == prepare('''
+        from nicegui import ui
+        from website.documentation.content.sub_pages_documentation import FakeSubPages
+
+        pages = FakeSubPages()
+
+        def root():
+            pages.routes = {'/': main, '/other': other}
+            pages.init()
+
+        ui.run(root)
+    ''')
+
+
+def test_sub_pages_with_data():
+    assert transform_for_demo_execution(prepare('''
+        from nicegui import ui
+
+        def root():
+            ui.sub_pages({
+                '/': main,
+                '/other': other,
+            }, data={'title': title})
+    ''')) == prepare('''
+        from nicegui import ui
+        from website.documentation.content.sub_pages_documentation import FakeSubPages
+
+        pages = FakeSubPages()
+
+        def root():
+            pages.routes = {
+                '/': main,
+                '/other': other,
+            }
+            pages.data = {'title': title}
+            pages.init()
+    ''')
+
+
+def test_empty_sub_pages():
+    assert transform_for_demo_execution(prepare('''
+        from nicegui import ui
+
+        def root():
+            pages = ui.sub_pages()
+            pages.add('/', main)
+    ''')) == prepare('''
+        from nicegui import ui
+        from website.documentation.content.sub_pages_documentation import FakeSubPages
+
+        pages = FakeSubPages()
+
+        def root():
+            pages.init()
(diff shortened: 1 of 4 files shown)


rodja:
While I generally like the idea I'm not sure the actual implementation is worth it. I guess it depends on how many `ui.sub_pages` tests we want to write/maintain.

evnchn:
Prefer not to add code transformation to _all_ of the demos just for `ui.sub_page` demos because we can't really be certain our new code transformer is the same as the old one if it is not a `ui.sub_page` demo. 

How about we can pass in a `transformer: Callable[[str], str]` somewhere?

evnchn:
> pass in a transformer

That is a good idea _in general_ because we can't have comments in demo code, and I can't say it (or any other constraints) definitely did not hold us back at one point. 

Also the code extraction is quite complex. Describing the logic as a series of transformers can also help with clarity.

falkoschindler:
Closing this one — the motivation is right (single-source demo code, no duplicated `FakeSubPages` workaround), but the approach has drifted too far from where `main` is today to be worth rebasing.

A few concrete reasons:

1. **`main` already converged on a `def root(): ...; return root` convention.** `get_full_code` strips the `return root` for display and `browser_window` calls the returned callable. That's the same goal this PR was reaching for, minus the `return root`. Keeping the PR's "no `return root`" variant forces a separate exec-into-namespace step that wouldn't be needed if the demos followed the same convention as e.g. `awaitable_refresh`.

2. **The string-based transformer is structurally fragile** — regex + manual paren counting + substring checks for `, data=` / `ui.link(` / `# HIDE` all have known holes (strings containing those tokens break the parser). The `# TRANSFORM:` mini-DSL is also invented vocabulary that future contributors would have to learn from tests. If we rebuild this on an AST (`ast` + `ast.unparse`, or `libcst` if we want to preserve comments), the transformation logic gets meaningfully more robust by construction — those edge cases stop being possible.

3. **Barely anyone missed this since October.** The branch has been sitting on milestone "Someday" for ~6 months while `main` moved 400+ commits ahead, including three new sub-pages demos (`path_changed_event_demo`, `wildcard_demo`, `error_handling_demo`) that all use the old `FakeSubPages` workaround pattern.

4. **The merge into current `main` is non-trivial** — 3-way conflicts in `code_extraction.py`, `demo.py`, and `sub_pages_documentation.py`, plus the new demos would need to be ported to the new style for consistency. That's a rewrite either way.

If we pick this up again, the shape I'd suggest:

- libcst-based rewriter (no regex, no `# TRANSFORM:` comment DSL — infer route literals and nesting structurally).
- `def root(): ...; return root` to match the rest of the docs.
- Migrate all sub-pages demos in the same PR so the file ends up with one style.

Closing in favor of a fresh attempt when someone has time.

evnchn:
@falkoschindler since https://github.com/evnchn/nicegui/pull/102 share mostly the same issues, I think can close it also then?

PR #5275, opened 2025-10-14
Replace intro demos with more educational code/text

rodja:
### Motivation

The main page demos are often the first code examples new users see. The previous demos (Styling, Common UI Elements, and Value Binding) showed what was possible but didn't effectively communicate NiceGUI's new core paradigm: building SPAs with a root function as the single entry point. This is a fundamental pattern that users should grasp early, especially those transitioning from frameworks like Streamlit.
The old demos also introduced too many concepts at once without clear educational flow, making it harder for newcomers to form the right mental model.

### Implementation

- replace old intro-demos with two new demos (single page app & reactive transformations). I would have liked a third one which demonstrates global state but had no good idea yet. So I'll leave it for another PR.
- changed `@ui.page` decorator in feature list with `ui.sub_pages` to establish SPA as the primary pattern.
- some improved infrastructure
  - by returning `root` the `@demo` helper automatically executes the root function and replaces the code to display with `ui.run(root)` (it is not used everywhere yet to minimize diff of this PR and reduce collisions with ideas from #5274)
  - made `FakeSubPages.link()` return `ui.label` to enable method chaining
 
### Progress

### DIFF
--- a/website/documentation/intro.py
+++ b/website/documentation/intro.py
@@ -68,7 +68,7 @@ def create_intro() -> None:
         When a new value arrives, it emits an event for the chart to be updated.
     ''')
     def event_system_demo():
-        from datetime import datetime
+        import time
 
         from nicegui import Event, app
 
@@ -87,7 +87,7 @@ def create_intro() -> None:
 
             def update_chart(temperature: float):
                 data = chart.options['series'][0]['data']
-                data.append([datetime.now().timestamp() * 1000, temperature])
+                data.append([time.time(), temperature])
                 if len(data) > 10:
                     data.pop(0)
 
@@ -95,7 +95,7 @@ def create_intro() -> None:
             # END OF DEMO
 
             data = chart.options['series'][0]['data']
-            data.append([datetime.now().timestamp() * 1000, 24.0])
+            data.append([time.time(), 24.0])
             ui.timer(1.0, lambda: update_chart(round(data[-1][1] + (random.random() - 0.5), 1)), immediate=False)
         app  # noqa: B018 to avoid unused import warning
 


website/documentation/intro.py
```diff
@@ -3,57 +3,58 @@
...
-        ui.icon('thumb_up')
-        ui.markdown('This is **Markdown**.')
-        ui.html('This is <strong>HTML</strong>.', sanitize=False)
-        with ui.row():
-            ui.label('CSS').style('color: #888; font-weight: bold')
-            ui.label('Tailwind').classes('font-serif')
-            ui.label('Quasar').classes('q-ml-xl')
-        ui.link('NiceGUI on GitHub', 'https://github.com/zauberzeug/nicegui')
+    @_main_page_demo('Single Page App', '''
+        Build applications with fast client-side routing using [`ui.sub_pages`](/documentation/sub_pages) and a `root` function as single entry point.
```
evnchn: But we still have this. 

image

> NOTE: This is an experimental feature, and the API is subject to change.

rodja:
> Importantly, I don't think we can put sub-pages at the spotlight while the API is still warned to be subject-to-change in the documentations.

You are right. I just talked to @falkoschindler and we agreed to remove the warning. It's about time!

rodja:
> While we're at it can we also spotlight the new Event system? 

Yes, I would really like that. But such a demo is blocked by #5231. Otherwise opening two tabs will have a high chance of not seeing the triggered events (because you land on another instance). Let's leave it for another PR.

evnchn:
> ...  spotlight the new Event system ... blocked by #5231

If the source of the events are not triggerable by user but come from hypothetical sources, we can easily get away with it. 

Imagine a demo:

```py
from nicegui import ui, app, Event

temperature_event = Event[int]()

@app.get('/temperature_sensor')
def temperature_sensor_webhook(temperature: float):
    temperature_event.emit(temperature)

@ui.page('/')
def temperature_dashboard():
    # some logic to subscribe to the temperature_event
```

We won't really expose `/temperature_sensor` in reality, though. 

And then we mock the demo with fluctuating temperatures from 24 to 26 degree celsius.

rodja:
> If the source of the events are not triggerable by user but come from hypothetical sources, we can easily get away with it.

Great idea. I also like that it would demonstrate the integration with FastAPI RESTful APIs!

rodja:
@evnchn would you like to take a shot at the third demo and test your new write permissions?

evnchn:
https://github.com/zauberzeug/nicegui/pull/5275/commits/8fcfbadd6ceb3b17f363273df1c3840c4f558a64

Did not wound up adding a link to RoSys since the rest of the demo text is so short.

rodja:
Thanks @evnchn. I replaced the log with an echart to squeeze in some data visualization. Even if the demo is a bit longer that way, I think its worth it.

evnchn:
> more "wow"

A good idea in general because some of the examples are too simple, really. 

Going forward I envision a "level select" for any demo's technical depth, possibly from "basic", "advanced" to "expert". 

See if that's considerable. If so can go to https://github.com/orgs/zauberzeug/projects/6 (may need designer's help to elegantly put a selector in the existing demo)

rodja:
> Going forward I envision a "level select" for any demo's technical depth, possibly from "basic", "advanced" to "expert".

I'm not sure if that is a good idea. It would add quite a bit of complexity. I think the intro-demos should be a a mix of 33% "nice", 33% "wow" and 33% "aha". We have the in-depth examples for advanced and expert users.

falkoschindler (approved):
As discussed with @rodja, I filled the SPA with a `ui.table` and a `ui.leaflet` map to make it visually more interesting. At the same time I tried to keep the code short and simple. The generic "row-click" event is a bit tricky, but I think it's good enough for now.

Let's merge this and further improve on a separate PR.

Issue #5244, opened 2025-10-08
ag-grid "jerking" on updates after v3 (significant decline in user experience)


amks1:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
@ui.page('/datatable_update')
async def datatable_update_test():
    columns = [
        {
            'field': 'product',
            'headerName': 'Product Name',
            'editable': False,
        },
        {
            'field': 'quantity',
            'headerName': 'Quantity',
            'editable': True
        },
        {
            'field': 'price',
            'headerName': 'Unit Price',
            'editable': True
        },
        {
            'field': 'category',
            'headerName': 'Category',
            'editable': False
        },
        {
            'field': 'in_stock',
            'headerName': 'In Stock',
            'editable': True
        },
        {
            'field': 'supplier',
            'headerName': 'Supplier',
            'editable': False
        },
        {
            'field': 'rating',
            'headerName': 'Rating',
            'editable': True
        },
    ]

    sample_data = [
        {
            'product': 'Laptop',
            'quantity': 10,
            'price': 1200.00,
            'category': 'Electronics',
            'in_stock': True,
            'supplier': 'TechCorp',
            'rating': 4.5
        },
        {
            'product': 'Mouse',
            'quantity': 25,
            'price': 25.50,
            'category': 'Accessories',
            'in_stock': True,
            'supplier': 'GadgetPro',
            'rating': 4.2
        },
        {
            'product': 'Keyboard',
            'quantity': 15,
            'price': 45.00,
            'category': 'Accessories',
            'in_stock': False,
            'supplier': 'GadgetPro',
            'rating': 4.0
        },
        {
            'product': 'Monitor',
            'quantity': 8,
            'price': 300.00,
            'category': 'Electronics',
            'in_stock': True,
            'supplier': 'DisplayWorld',
            'rating': 4.7
        },
        {
            'product': 'Desk Lamp',
            'quantity': 20,
            'price': 18.99,
            'category': 'Furniture',
            'in_stock': True,
            'supplier': 'HomeEssentials',
            'rating': 4.1
        },
        {
            'product': 'Chair',
            'quantity': 5,
            'price': 85.00,
            'category': 'Furniture',
            'in_stock': False,
            'supplier': 'HomeEssentials',
[...]

Issue #5276, opened 2025-10-14
ui.aggrid render failed when `options["columnDefs"]["flex"] = 1` after upgrade to version 3.0


guyingbo:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
ui.aggrid({
    "columnDefs": [
        {"headerName": "Name", "field": "name", "width": 200},
        {"headerName": "Age", "field": "age", "minWidth": 200, "flex": 1},
        {"headerName": "Parent", "field": "parent", "hide": True},
    ],
    "rowData": [
        {"name": "Alice", "age": 18, "parent": "David"},
        {"name": "Bob", "age": 21, "parent": "Eve"},
        {"name": "Carol", "age": 42, "parent": "Frank"},
    ],
    "rowSelection": {"mode": "multiRow"},
})
```

### Description

When `"flex": 1` is added, grid failed to render on version 3.x

### NiceGUI Version

3.0.4

### Python Version

3.12

### Browser

Chrome

### Operating System

macOS

falkoschindler:
Thanks for reporting this issue, @guyingbo!
Let me share the following analysis from Claude 4.5 Sonnet, which is pretty spot on:

> **The Issue**: The grid fails to render when `"flex": 1` is added to column definitions in NiceGUI 3.0.4.
> 
> **Root Cause Hypothesis**: I suspect there's a **conflict between the `auto_size_columns` feature and AG Grid's `flex` property**. 
> 
> Looking at the code in `aggrid.js` (lines 77-79):
> 
> ```javascript
> if (type === "gridSizeChanged" && this.auto_size_columns) {
>   this.api.sizeColumnsToFit();
> }
> ```
> 
> The `auto_size_columns` parameter defaults to `True` in NiceGUI's `AgGrid` class (line 27 in `aggrid.py`). This means every time the grid size changes, NiceGUI calls `sizeColumnsToFit()`, which overrides AG Grid's built-in flex sizing behavior.
> 
> ### Why This is a Problem
> 
> 1. **`flex` is a native AG Grid property** for responsive column sizing that should automatically distribute available space
> 2. **`sizeColumnsToFit()`** is a method that proportionally sizes columns to fill the grid width, but it conflicts with `flex`-based sizing
[...]

PR #5277, opened 2025-10-14
Fix auto-sizing of `ui.aggrid`

falkoschindler:
### Motivation

In #5276 we noticed that `ui.aggrid`'s column auto-sizing break in the presence of flex columns.

### Implementation

~~This PR conditionally ignores the `auto_size_columns` parameter if any column has a `flex` attribute.~~
This PR uses the "autoSizeStrategy" to implement the `auto_size_columns` parameter instead of calling a JavaScript function on size change.

### Progress

### DIFF
--- a/nicegui/elements/aggrid/aggrid.js
+++ b/nicegui/elements/aggrid/aggrid.js
@@ -74,13 +74,6 @@ export default {
       return runMethod(this.api.getRowNode(row_id), name, args);
     },
     handle_event(type, args) {
-      if (
-        type === "gridSizeChanged" &&
-        this.auto_size_columns &&
-        !this.gridOptions.columnDefs?.some((col) => col.flex !== undefined)
-      ) {
-        this.api.sizeColumnsToFit();
-      }
       this.$emit(type, {
         value: args.value,
         oldValue: args.oldValue,
@@ -118,6 +111,5 @@ export default {
   props: {
     options: Object,
     html_columns: Array,
-    auto_size_columns: Boolean,
   },
 };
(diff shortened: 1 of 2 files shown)


falkoschindler:
@evnchn Maybe calling `this.api.sizeColumnsToFit()` isn't such a great idea. I just experimented with adding
```
'autoSizeStrategy': {'type': 'fitGridWidth'},
```
to the `options` dictionary and the result is flawless. It also solves the flickering from #5244.

evnchn:
Sure, maybe we can explore that. 

But still, if the library's sane, no amount of reasonable tinkering with the options should result in a _blank table_, so I am calling it 🤡 in AG Grid.

falkoschindler:
I changed this PR to use the "autoSizeStrategy" - just to notice that this breaks when adding `'flex': 1` as in #5276. Not sure what I've done differently in https://github.com/zauberzeug/nicegui/pull/5277#issuecomment-3402085512.

falkoschindler:
My current conclusion would be:

- Using "autoSizeStrategy" instead of calling `sizeColumnsToFit()` is still better to reduce flickering when updating the grid (fixes #5244).
- Combining "flex" with either `sizeColumnsToFit()` _or_ "autoSizeStrategy" isn't supported by AG Grid. E.g. it warns that "colDef.flex is not supported with gridOptions.autoSizeStrategy". So we can add a note to the documentation (to address #5276), but shouldn't try to solve the impossible.

evnchn:
I'm fine with this PR as it is, because we are at a state where if the config is invalid, console errors are produced. 

Still though, no warnings, when doing `sizeColumnsToFit` on tables `flex` columns? 🤡 

---

On a separate note, I'm considering to create a mechanism where browser warnings and errors are logged to the Python console. AG Grid will be a **top user** of this functionality.

PR #5278, opened 2025-10-14
Add documentation for upload event arguments

falkoschindler:
### Motivation

The Discord user MasterOfNone asked about documentation for the event arguments of `ui.upload`. So far we only described them in the [release notes](https://github.com/zauberzeug/nicegui/releases/tag/v3.0.0) for version 3.0.

### Implementation

This PR doesn't try to demonstrate all properties and methods, but simply lists them with a short description. In combination with other demos this should be sufficient.

### Progress

### DIFF
--- a/website/documentation/content/upload_documentation.py
+++ b/website/documentation/content/upload_documentation.py
@@ -73,4 +73,5 @@ def uploading_large_files() -> None:
     ui.upload(on_upload=lambda e: ui.notify(f'Uploaded {e.file.name}')).classes('max-w-full')
 
 
-doc.reference(ui.upload)
+doc.reference(ui.upload, title='Reference for ui.upload')
+doc.reference(ui.upload.FileUpload, title='Reference for ui.upload.FileUpload')


evnchn:
I am also considering a simplier solution where we are less likely to run into stale documentation, if we did:

```py
doc.reference(ui.upload.FileUpload, title="Reference for FileUpload")
```

falkoschindler:
> I am thinking:
> 
> ```python
> ui.upload(on_upload=lambda e: ui.notify(f'Uploaded {e.file.name} of size {e.file.size()}')).classes('max-w-full')
> ```

There's another demo using `e.file.text()`. So we basically cover calling methods on `e.file`. And I'd rather not make the lambda expression any longer (way beyond the normal window size).

> I am also considering a simplier solution where we are less likely to run into stale documentation, if we did:
> 
> ```python
> doc.reference(ui.upload.FileUpload, title="Reference for FileUpload")
> ```

Good idea. I added this reference section in 67b2e54e3aca986232128ae603a5c49a2d1892de.
I'd still keep the section about "Upload event arguments" because most properties and methods are not covered by any demos and the reference section is a bit hard to find.

PR #5281, opened 2025-10-14
expose leaflet by default (unless no_conflict mode)

### LINKED ISSUE #5279: Leaflet not available when draw_control is False
### First Check

### Example Code

```python
from nicegui import ui

m = ui.leaflet(center=(51.505, -0.09)).classes('h-32')
marker = m.marker(latlng=m.center)
ui.button('Hide', on_click=lambda: marker.run_method('setOpacity', 0.3))
ui.button('Show', on_click=lambda: marker.run_method('setOpacity', 1.0))

icon = 'L.icon({iconUrl: "https://leafletjs.com/examples/custom-icons/leaf-green.png"})'
ui.button('Change icon', on_click=lambda: marker.run_method(':setIcon', icon))

ui.run()
```

### Description

We use the run methods of the Leaflet Layer objects. Since Nicegui 3 our implementation does not work anymore. The [Run Layer Methods example](https://nicegui.io/documentation/leaflet#run_layer_methods) in the documentation does not work anymore as well.
The icon will not change when clicking "Change Icon" and the Browser console shows the following error message:
`Uncaught (in promise) ReferenceError: L is not defined`
When passing an empty dictionary to ui.leaflet as draw_control, L is accessible. It seems like loadLeafletDraw in elements/leaflet/leaflet.js imports leaflet, but is only called when draw_control is not false.

### NiceGUI Version

3.0.4

### Python Version

3.12.10

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

evnchn:
### Motivation

We have 2 issues with regards to the current Leaflet implementation in NiceGUI. 

1. `window.L` unavailable unless `draw_control` is set, which breaks user code which this PR fix #5279
2. Unlike old Leaflet which you can call `L.noConflict` to not pollute the window namespace, there is no such function in NiceGUI 3.0

### Implementation

- `exposeLeaflet` exposes `L` into `window.L`, runs by default
- ~~Unless you did `no_conflict=True`~~
- ~~Unless you are specifying also `draw_control`,  in which case the latter takes precedence.~~

### Progress

- [x] Documentation has been added (I change the docstring).


evnchn:
Can we save adding `no_conflict` by defaulting `draw_control` to `{}`, and the proposed no-conflict mode can be entered with an explicit `draw_control=False`?

falkoschindler:
@evnchn Two thoughts:

1. Why did you add `exposeLeaflet` to index.mjs? Can't you simply inline the assignment `window.L = L` where it is used? This would avoid the need to re-bundle, keeps all logic in leaflet.js and reduces the diff of this PR.
2. Do you really think we need the `no_conflict` option?
    - Using this parameter sounds like a good idea, but it can break things like #5279. This can certainly be surprising and confusing for our users.
    - The name and default value is confusing: `no_conflict: bool = False` means conflict=true by default. Why should anyone want that?
    - Leaflet used to pollute `window.L` before 3.0 (I guess), so we could look away and allow it to proceed.
    - The chance of an accidental conflict is pretty low.
    - Leaflet 2.0 is [scheduled for November 2025](https://github.com/Leaflet/Leaflet/issues/9869) and won't come with a global `L`. We will need to wait for NiceGUI 4.0 though before migrating, but at least there is improvement on the horizon.

evnchn:
> Can't you simply inline the assignment window.L = L where it is used?

True. Should have thought of that!

> Do you really think we need the no_conflict option?

I agree to your arguments to anything which I had, which I also forgot as well. 

Anyways, whoever wants a `no_conflict` can pull up https://github.com/zauberzeug/nicegui/pull/5281/commits/fd0b63d9f1f9aeb1bf8dc65dc1a61a6dce2a3aae and PR from there, who shall be responsible for elucidating the reasons in due course. 

I am busy, maybe you can remove it. @falkoschindler

PR #5283, opened 2025-10-14
Run publisher in background and use pkill to end example

OpenSourceIronman:
### Motivation

Make running the ZeroMQ example easier for beginners that don't understand they will need two terminals open to run this example. For experts this  speeds up example boot-up with a friendly reminder to use "&" at the end of command to run code in the background and give terminal control back to the user. 

https://www.linux.com/training-tutorials/and-ampersand-and-linux/

### Implementation

Uses "&" and "pkill" to simplify running example 

I used "pkill -9" instead on "kill -9"  to send SIGKILL to all processes whose name matches “python”.

✅ Pros:
	• Convenient: No need to find PIDs manually.
	• Powerful for bulk termination (e.g., stop all instances of Python).

⚠️ Cons:
	• Riskier — might kill more than you intend.
        • No confirmation prompt by default.

### Progress

- [X] Pytests are not necessary

### DIFF
--- a/examples/zeromq/README.md
+++ b/examples/zeromq/README.md
@@ -32,12 +32,12 @@ python main.py
 
 ### Results
 
-Once both the publisher and GUI server are running, you will see an updating plot on the UI. 
+Once both the publisher and GUI server are running, you will see an updating plot on the UI.
 
 ![plot](images/plot.png)
 
 To kill zmq-server.py running in the background, execute the following command after main.py ends.
 
 ```bash
-kill -9 %% 
+kill -9 %%
 ```


examples/zeromq/README.md
```diff
@@ -32,6 +32,12 @@ python main.py
 
 ### Results
 
-Once both the publisher and GUI server are running, you will see an updating plot on the UI.
+Once both the publisher and GUI server are running, you will see an updating plot on the UI. 
+
+To kill both zmq-server.py and main.py run the following command:
+
+```bash
+pkill -9 Python
```
rodja: I personally like `kill -9 %%` which kills the last process send to background. Should be fine here and ensures you only kill the intended process.

PR #5290, opened 2025-10-15
Add the anti-hang slot template for the user if they have lists in the cells

### LINKED ISSUE #2744: Browser can't load page when table rows contain lists

evnchn:
### Motivation

Fixes #2744 and #4837 for good, in which row values being lists cause Vue to grind to a halt. 

### Implementation

Contrast to #4694 this PR's approach is to:

1. Apply the slot template which ceases the hanging for the user (https://github.com/zauberzeug/nicegui/issues/2744#issuecomment-2881607557)
2. Informs the user they better change their code or they must live with console warnings. 

Not discovering this solution earlier was caused by how the discussion was spread in multiple places and that the comment is hidden by "show more items"

### Progress

### DIFF
--- a/nicegui/elements/table.py
+++ b/nicegui/elements/table.py
@@ -80,8 +80,6 @@ class Table(FilterElement, component='table.js'):
         self._selection_handlers = [on_select] if on_select else []
         self._pagination_change_handlers = [on_pagination_change] if on_pagination_change else []
 
-        self._scan_rows_for_lists()
-
         def handle_selection(e: GenericEventArguments) -> None:
             if e.args['added']:
                 if self.selection == 'single':
@@ -103,27 +101,27 @@ class Table(FilterElement, component='table.js'):
                 handle_event(handler, arguments)
         self.on('update:pagination', handle_pagination_change)
 
-    def _scan_rows_for_lists(self) -> None:
-        """Check if any cell in the rows contains a list.
-
-        Also catches anything that is eventually converted to a list in the conversion to JSON."""
-        for row in self._props['rows']:
-            for key, value in row.items():
-                if isinstance(value, (list, set, tuple)) and f'body-cell-{key}' not in self.slots:
-                    log.warning(f'Found list in row in column "{key}": {value}.\n'
-                                'Unless there is slot template, '
-                                'table rows must not contain lists. '
-                                'or the browser will crash.\n'
-                                'NiceGUI is intervening by adding a slot template to display the list as comma-separated values.\n'
-                                )
+    def _to_dict(self) -> dict[str, Any]:
+        # scan rows for lists and add slot templates if needed
+        for column in self._props['columns']:
+            key = column.get('field')
+            if not key or f'body-cell-{key}' in self.slots:
+                continue
+            for row in self._props['rows']:
+                value = row.get(key)
+                if isinstance(value, (list, set, tuple)):
+                    log.warning(
+                        f'Found list in column "{key}": {value}.\n'
+                        'Unless there is slot template, table rows must not contain lists or the browser will crash.\n'
+                        'NiceGUI is intervening by adding a slot template to display the list as comma-separated values.'
+                    )
                     self.add_slot(f'body-cell-{key}', '''
                         <td class="text-right" :props="props">
                             {{ Array.isArray(props.value) ? props.value.join(', ') : props.value }}
                         </td>
                     ''')
+                    break
 
-    def _to_dict(self) -> dict[str, Any]:
-        self._scan_rows_for_lists()
         return super()._to_dict()
 
     def on_select(self, callback: Handler[TableSelectionEventArguments]) -> Self:


falkoschindler:
This PR implements what we considered in #2744, but hesitated because it could be costly - without adding value for 99.9% of our users: https://github.com/zauberzeug/nicegui/issues/2744#issuecomment-2887757633
I'm still worried that such a check is more harm than good.

evnchn:
> ... costly ...

#4694 is costly but I don't think this is. Consider you need to go through all the data once to serialize it to JSON, and so everything is still O(N) given N cells, as always. 

I take a freezing browser as unacceptable and is willing to add 10% performance penalty in order NOT have that experience.

falkoschindler:
Ah, I overlooked the fact that this PR isn't correcting all cells, but only adds a named slot. I'm starting to like the idea... 😎

We could iterate over columns first and
- `continue` if the slot already exists and
- `break` from the inner row loop if a list has been found.

But note that we're not catching all cases, yet. Modifying the observable `rows` can still cause a freeze:
```py
table = ui.table(rows=[{'name': 'John', 'age': 20}, {'name': 'Jane', 'age': 21}])
table.rows.append({'name': 'Jim', 'age': [22, 23]})
```
Same problem:
```py
table.add_row({'name': 'Jim', 'age': [22, 23]})
```

evnchn:
> ... not catching all cases ...

Caught them all with https://github.com/zauberzeug/nicegui/pull/5290/commits/7674d865619c03b4900a4c242d1e0dac0c14e85f, I think.

evnchn:
> break from the inner row loop if a list has been found

I don't think so... since the slot is specific to the column, and so if there are 2 columns in the same row which have the list then the `break` would cause the code not catch the second column

falkoschindler:
I thought more like
```py
for column in columns:
    if slot_exists:
        continue
    for row in rows:
        if bad_type:
            add_slot()
            break
```

---

Apart from that, consider an application like
```py
@ui.button('Increment all').on_click
def increment_all():
    for row in table.rows:
        row['age'] += 1
```
Every `+= 1` will check again all cells. This is O(N²), which you could call "costly"...

evnchn:
> Every `+= 1` will check again all cells

Fair. Can you tell me where is the insertion point, if I want to add a function which is ran _just before_ an element is serialized to the outbox? 

`self._props.on_change(self._scan_rows_for_lists)` is close, but that triggers every time even in intermediate operations lilke `increment_all`, which we don't need. Conceptually we just need to scan the rows "one last time before we ship the rows to the browser".

falkoschindler:
Maybe we can override `_to_dict`? This is where data is collected before sending it to the client.

evnchn:
That works. 

I also tried to tap into this:

```py
    def update(self) -> None:
        """Update the element on the client side."""
        if self.is_deleted:
            return
        self.client.outbox.enqueue_update(self)
```

But for some reason it `KeyError` with "rows"...

falkoschindler (approved):
@evnchn I refactored the nested for-loops as outlined in https://github.com/zauberzeug/nicegui/pull/5290#issuecomment-3446857509. What do you think? Ready to merge from my side.

evnchn:
Finally got what your for loop was about. It was good.

nicegui/elements/table.py
```diff
@@ -100,6 +101,29 @@ def handle_pagination_change(e: GenericEventArguments) -> None:
...
+            for row in self._props['rows']:
+                value = row.get(key)
+                if isinstance(value, (list, set, tuple)):
+                    log.warning(
+                        f'Found list in column "{key}": {value}.\n'
+                        'Unless there is slot template, table rows must not contain lists or the browser will crash.\n'
+                        'NiceGUI is intervening by adding a slot template to display the list as comma-separated values.'
+                    )
+                    self.add_slot(f'body-cell-{key}', '''
+                        <td class="text-right" :props="props">
```
evnchn, line 119: Is `class="text-right"` needed though? I stole this from https://github.com/zauberzeug/nicegui/issues/2744#issuecomment-2881607557 without thinking too much...

evnchn (commented):
I mean we're applying the slot for the user so we can do _whatever_ but still just to question everything we wrote...

nicegui/elements/table.py
falkoschindler: This seems to be the default style. When switching from
```py
rows=[{'A': [1, 2, 3]}]
```
to
```py
rows=[{'A': '[1, 2, 3]'}]
```
the HTML looks the same with `class="text-right"`.

PR #5292, opened 2025-10-17
Default sanitize value in html element

TM0088:
ui.html(sanitize = False)

### Motivation
Avoid breaking existing html elements 

### Implementation
Set default value

### DIFF
--- a/nicegui/elements/html.py
+++ b/nicegui/elements/html.py
@@ -7,7 +7,7 @@ from .mixins.content_element import ContentElement
 
 class Html(ContentElement):
 
-    def __init__(self, content: str = '', *, sanitize: Callable[[str], str] | Literal[False], tag: str = 'div') -> None:
+    def __init__(self, content: str = '', *, sanitize: Callable[[str], str] | Literal[False] = False, tag: str = 'div') -> None:
         """HTML Element
 
         Renders arbitrary HTML onto the page, wrapped in the specified tag.


evnchn:
Hello @TM0088 

You should read https://github.com/zauberzeug/nicegui/pull/5187, where the gist is, we enforce that the developer must explicitly opt-out of sanitization. 

Otherwise, there is no way to easily tell that the input to `ui.html` isn't a f-string which depends on user input, and it would be a vulnerability. 

It requires load on the developer but we figured that it is better than the developer resting easy at home but the security team waking up to a XSS attack 2am in the midnight. 

I'm inclined to close this.

falkoschindler:
Exactly. The breaking change is intentional to address https://github.com/zauberzeug/nicegui/security/advisories/GHSA-8c95-hpq2-w46f.

> Framework authors, however, should make the _safe_ path the _easy_ path.

If `sanitize` was `False` by default, unexperienced users probably wouldn't change it (easy) and potentially open a security vulnerability (not safe).

We're not perfectly happy with this extra parameter, but there was no better way to combine security with ease of use.

WolfgangFahl:
really? _there was no better way to combine security with ease of use._

Just create a new ui.html_secure instead of breaking the unsecure old one.

evnchn:
Yes there wasn't. At least we didn't think of any. 

On proposal of `ui.html_secure`:

- It's no tougher to change from `ui.html(...)` to `ui.html(..., sanitize=False)` than to `ui.html_secure`
- Should `ui.html_secure` have sanitization, library install size larger
- And `ui.html_secure` can't possibly resolve to `ui.html(..., sanitize=False)`
- We prefer not to keep 2 copies of functions, near identical in function, in the codebase. 
  - Can you tell me, in 5 seconds, what is the difference between PHP's `mysql_escape_string` and `mysql_real_escape_string`? 

Rebuttals and alternative proposals welcome 😄

WolfgangFahl:
**It's no tougher to change from ui.html(...) to ui.html(..., sanitize=False) than to ui.html_secure**

True but in one case it is a deliberate decision and in the other cases it is forced and the system breaks if you do not. I would rather have my system be insecure under circumstances i know e.g. in development and testing and secure environments than having them break with no further notice. Dependency hell is all around already it must not get worse on items that have options to behave differently.

rodja:
> in one case it is a deliberate decision and in the other cases it is forced and the system breaks if you do not

That is exactly the purpose. The breaking change was intentional to inform **every developer** that there is a potential security hole in the code base. You **must** think about it and then decided for your self if you want to sanitize or not. Leaving the unsecure version as default was not an option in our opinion.

> having them break with no further notice

Come on. That is hardly true. We made a major version jump and [wrote a migration guide](https://github.com/zauberzeug/nicegui/releases/tag/v3.0.0). Have you read it?

Issue #5247, opened 2025-10-09
`SyntaxError: source code string cannot contain null bytes` after packaging with `nicegui-pack`


romaintailhurat:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Description

Hi!

I wrote a very simple app in order to test the packaging of NiceGUI:

```python
from multiprocessing import freeze_support
freeze_support()
from nicegui import ui, native  # noqa

ui.label("TEST")

ui.run(reload=False, port=native.find_open_port())
```

My project is managed using `uv`, which I used to install NiceGUI (`3.0.3`) and the latest version of PyInstaller (`6.16.0`).

My system is macOS Tahoe 26.0.1.

I packaged the app with: `uv run nicegui-pack --onefile --name "psg" main.py`.

I'm trying to run the app with: `./dist/psg`, the app launches but it display a 500 error page with the message `SyntaxError: source code string cannot contain null bytes`.

I checked: my source file doesn't contains any hidden null character.

I tested the app with both Safari and Firefox, same issue.

Thanks for considering my troubles.

### NiceGUI Version

3.0.3

### Python Version

3.13

### Browser

Safari

### Operating System

macOS

Fleckso:
Hey, i have the same problem and couldn't solve it.
Im using:
- Windows
- Python==3.10.11
- NiceGUI==3.0.2
- Pyinstaller==6.16.0

I tried different build commands with any combination of onefile, windowed.
Will let you know if I find smth.

Fleckso:
I solved my problem.... you have to add a page which you want to show... like this:

```py
@ui.page('/')
def page():
    ui.link('Visit other page', other_page)
    ui.link('Visit dark page', dark_page)
```

https://nicegui.io/documentation/section_pages_routing

In the error message you can read this: 
`return await page('')`
thats why i checked the NiceGui Page thing.

romaintailhurat:
Wow indeed that also solved my problem! Thanks @Fleckso .

I'm still a bit confused, [the deployment page](https://nicegui.io/documentation/section_configuration_deployment#package_for_installation) of the website gives an example that specifically doesn't use the page construct. Should It be updated?

evnchn:
> gives an example that specifically doesn't use the page construct

[...]

PR #5297, opened 2025-10-17
Add a note about packaging NiceGUI scripts

falkoschindler:
### Motivation

In #5247 we noticed that packaging NiceGUI scripts (apps without `@page` or `root` functions) does not work - especially using the `--onefile` argument.

### Implementation

This PR adds a note about this limitation and adjusts the demo accordingly.

### Progress

### DIFF
--- a/website/documentation/content/section_configuration_deployment.py
+++ b/website/documentation/content/section_configuration_deployment.py
@@ -257,7 +257,7 @@ doc.text('Package for Installation', '''
     Just make sure
 
     - to call `ui.run` with `reload=False` in your main script to disable the auto-reload feature, and
-    - to use either a `root` page function in `ui.run` or decorated `@page` functions. The script mode is not supported.
+    - to pass a `root` page function to `ui.run` or define at least one decorated `@page` function.
 
     Running the `nicegui-pack` command below will create an executable `myapp` in the `dist` folder:
 ''')


website/documentation/content/section_configuration_deployment.py
```diff
@@ -254,7 +254,11 @@ def uvicorn_ssl():
     NiceGUI apps can also be bundled into an executable with `nicegui-pack` which is based on [PyInstaller](https://www.pyinstaller.org/).
     This allows you to distribute your app as a single file that can be executed on any computer.
 
-    Just make sure to call `ui.run` with `reload=False` in your main script to disable the auto-reload feature.
+    Just make sure
+
+    - to call `ui.run` with `reload=False` in your main script to disable the auto-reload feature, and
+    - to use either a `root` page function in `ui.run` or decorated `@page` functions. The script mode is not supported.
```
evnchn: I think we are "dereferencing a null pointer" on script mode, because curiously it is never documented in NiceGUI documentation (at least I can't find it). Release notes for 3.0.0 links to #5005 only. 

People who aren't terminally online on our GitHub repo may not know much about it...

Other than that I see no issues.

evnchn (changes requested):
TL-DR: We can't talk about script mode in the docs if we didn't document script mode.

falkoschindler:
@evnchn You're right. It's probably not necessary to mention the "script mode" here. I've updated the text.

Issue #5177, opened 2025-09-26
copying the content of ui.code after set_content yields to the initial content


jlmanaloto:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

c = ui.code('''
    from nicegui import ui

    ui.label('Code inception!')

    ui.run()
''').classes('w-full')
c.set_content('new content')
c.update()

ui.run()
```

### Description

Copying the content of ui.code always gives the initial content even after using `set_content()` and `update()`. Given the example code, I would expect to get "new content" when pasting to a text editor but instead,  I still get the code block from the initialized value of ui.code content.

### NiceGUI Version

2.24.2

### Python Version

Python 3.11.9

### Browser

Chrome

### Operating System

Windows

python-and-novella:
The truth is, f-string in js_handler is only pure string.
So, clicking copy button just copies the current content:
```python3
from nicegui import ui
import json
c = ui.code('''
    from nicegui import ui

    ui.label('Code inception!')

    ui.run()
''').classes('w-full')

ui.button('copy before update').on('click', js_handler=f'() => navigator.clipboard.writeText({json.dumps(obj=c.content)})')

c.set_content('new content')
c.update()

ui.button('copy after update').on('click', js_handler=f'() => navigator.clipboard.writeText({json.dumps(obj=c.content)})')
ui.run()
```

python-and-novella:
I make a little patch:
```python3
from nicegui import ui
import json
c = ui.code('''
    from nicegui import ui

    ui.label('Code inception!')

    ui.run()
''').classes('w-full')

c.set_content('new content')
c.update()

with c:
    c.copy_button.delete()
    c.copy_button = ui.button(icon='content_copy', on_click=c.show_checkmark) \
                .props('round flat size=sm').classes('absolute right-2 top-2 opacity-20 hover:opacity-80') \
                .on('click', js_handler=f'() => navigator.clipboard.writeText({json.dumps(c.content)})')

ui.run()
```

python-and-novella:
Or try this:
```python3
from nicegui import ui

import json
class Code(ui.code):
    def _handle_content_change(self, content):
        if not hasattr(self,'copy_button'): return
        with self:
            self.copy_button.delete()
[...]

PR #5298, opened 2025-10-17
Fix clipboard button in `ui.code` after content has been changed

falkoschindler:
### Motivation

In #5177 we noticed that the copy button writes old content to the clipboard when the content of `ui.code` has been updated:
```py
c = ui.code('old content')
c.content = 'new content'
```
Clicking the button will write "old content" to the clipboard.

### Implementation

This PR introduces a new "content" prop to hold the raw (non-HTML) version of the content. When clicking the copy button, a JavaScript call will write the value of this prop to the clipboard.

### Progress

- [x] Pytests are probably not necessary (and probably pretty hard to write).

### DIFF
--- a/nicegui/elements/code.js
+++ b/nicegui/elements/code.js
@@ -3,4 +3,7 @@ export default {
   mounted() {
     if (!navigator.clipboard) this.$el.querySelector(".q-btn").style.display = "none";
   },
+  props: {
+    content: String,
+  },
 };
--- a/nicegui/elements/code.py
+++ b/nicegui/elements/code.py
@@ -2,7 +2,6 @@ import asyncio
 import time
 from typing import Optional
 
-from .. import json
 from .button import Button as button
 from .markdown import Markdown as markdown
 from .markdown import remove_indentation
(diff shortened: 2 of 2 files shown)


evnchn:
On https://github.com/zauberzeug/nicegui/issues/5177#issuecomment-3341897096: I had the wrong concept since I often forget that `ui.code` is nesting `ui.markdown` inside itself. It's tricky...

falkoschindler:
> Identical to [#5177 (comment)](https://github.com/zauberzeug/nicegui/issues/5177#issuecomment-3339015643), which I have tested to be working before.

Almost. I noticed that we need to put `self.content` into the prop, not `content` (which still has the full indentation).

> On [#5177 (comment)](https://github.com/zauberzeug/nicegui/issues/5177#issuecomment-3341897096): I had the wrong concept since I often forget that `ui.code` is nesting `ui.markdown` inside itself. It's tricky...

Exactly. The prop name "content" might be a bit confusing. On the other hand that's exactly what it contains. Anyway, it's an implementation detail.

PR #5299, opened 2025-10-17
Document the script mode

evnchn:
### Motivation

We can't talk about not using the script mode in https://github.com/zauberzeug/nicegui/pull/5297, if we did not document what that even is. 

### Implementation

Add a small section on what is script mode in the documentation. 

### Progress

### DIFF
--- a/website/documentation/content/section_pages_routing.py
+++ b/website/documentation/content/section_pages_routing.py
@@ -22,17 +22,13 @@ doc.intro(sub_pages_documentation)
 
 
 @doc.demo('Script Mode', '''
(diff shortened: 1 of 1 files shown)


falkoschindler:
Thanks, @evnchn!
But you're referring to the root function, which is also not explicitly documented. This is an open task on #5109.
@rodja Are you still planning to write a section about the `root` parameter? If so, we should coordinate our efforts.

rodja:
Yes, root pages must also get better documentation. I took a first step with #5275. If that is through I want to update examples and also some of the demos where @ui.page is currently used. I think this can be an independent, and useful PR.

PR #5300, opened 2025-10-19
Clear module cache between pytests for proper isolation

rodja:
### Motivation

NiceGUI's testing infrastructure has a problematic test isolation issue with module imports:

In user fixture and screen fixture setup `runpy` is used to execute the main file, application modules with `@ui.page()` decorators get imported and register routes. However, Python caches these modules in `sys.modules` across test runs. On subsequent tests, the cached modules aren't re-executed, so their `@ui.page()` decorators don't run again, and routes aren't registered in the fresh NiceGUI app instance.

### Implementation

**Core Fix:**
Modified the `nicegui_reset_globals()` fixture to track and clean up dynamically imported modules between tests. The fixture now:

1. Captures `sys.modules` state before test execution
2. Allows the test to run normally
3. Removes any modules added during test execution which registered page routes

This ensures all application modules are re-imported and their decorators re-executed for each test.

**Trade-offs:**

- Removing only modules with page routes avoids problems with re-imporing libs like numpy and others which have hidden state (for example in c code)
- Added `pytest-order` dependency to make test execution order explicit for the isolation verification tests
- Slight overhead from re-importing modules is acceptable given the guarantee of true test isolation

**Test Coverage:**
Added two sequential tests that verify module isolation works correctly across test boundaries.

### Progress


evnchn:
@rodja:
1. Got mypy to be silent. 
2. According to https://github.com/evnchn/nicegui/actions/runs/18658724781 the pipeline shall be green after the merge from main.

PR #5303, opened 2025-10-19
Lazy imports for `ui` module using PEP 562 `__getattr__`

evnchn:
### Motivation

`from nicegui import ui` eagerly imports every UI element, pulling in heavy transitive dependencies (pyecharts, plotly, etc.) even when only a few elements are used. This adds significant startup time. Lazy imports defer these loads until the element is actually accessed.

See also: #5684

### Implementation

Uses Python's PEP 562 module-level `__getattr__` (stdlib, no dependencies) to lazily import UI elements on first access:

- **`_LAZY_IMPORTS` dict** is the single source of truth mapping attribute names to `(module_path, attr_name)` tuples
- **`__all__`** is derived from `_LAZY_IMPORTS` keys
- **`TYPE_CHECKING` block** provides real imports for mypy/pyright static analysis
- **`__getattr__`** resolves imports on first access and caches via `setattr(sys.modules[__name__], ...)`
- **ESM element packages** with `dist/` directories are eagerly imported so `__init_subclass__` registers their modules in the importmap. Uses `importlib.util.find_spec()` for robust path resolution.
- **`clipboard` stays a module** — `None` as `attr_name` means "return the module itself", no breaking change
- **ESM `__init__.py` files** use `TYPE_CHECKING` imports to satisfy pylint (no more `# pylint: disable=undefined-all-variable`)

No new dependencies. No changes to `clipboard.py`. No breaking API changes.

### Benchmark

Scenario 2 from #5684 shows overhead dropping from 0.46s to 0.27s (~41% reduction) on startup.
Scenario 3 from #5684 shows overhead dropping from 0.53s to 0.32s (~40% reduction) on each call.
Scenario 4 from #5684 shows overhead dropping from 0.63s to 0.33s (~48% reduction) on first call.

### Progress

- [x] Pytests have been added (`tests/test_lazy_imports.py`: all names resolve, `dir(ui)` correct, `AttributeError` on bad attr, `_LAZY_IMPORTS`/`__all__` sync).

### DIFF
--- a/nicegui/dependencies.py
+++ b/nicegui/dependencies.py
@@ -157,16 +157,15 @@ def register_esm(name: str, path: Path, *, max_time: float | None) -> None:
     if key in esm_modules:
         if esm_modules[key].name == name:
             return
-        raise ValueError(f'Conflicting ESM registration for key "{key}": '
-                         f'"{esm_modules[key].name}" vs "{name}"')
+        raise ValueError(f'Conflicting ESM registration for key "{key}": "{esm_modules[key].name}" vs "{name}"')
     esm_modules[key] = EsmModule(name=name, path=path)
 
 
-def setup_esm_package(package_file: str, package_name: str, esm_name: str,
-                      exports: dict[str, str]) -> tuple[Callable[[str], object], Callable[[], list[str]]]:
+def setup_esm_package(package_file: str, package_name: str, esm_name: str, exports: dict[str, str]) \
+        -> tuple[Callable[[str], object], Callable[[], list[str]]]:
     """Register an ESM module and return ``__getattr__`` and ``__dir__`` for lazy class loading.
 
-    Each ESM element package can use this in its ``__init__.py``::
+    Each ESM element package can use this in its ``__init__.py``:
 
         from ...dependencies import setup_esm_package
         __getattr__, __dir__ = setup_esm_package(__file__, __name__, 'nicegui-aggrid', {'AgGrid': '.aggrid'})
@@ -176,16 +175,16 @@ def setup_esm_package(package_file: str, package_name: str, esm_name: str,
     register_esm(esm_name, dist, max_time=dist.stat().st_mtime)
 
     by_submodule: dict[str, list[str]] = {}
-    for attr, submod in exports.items():
-        by_submodule.setdefault(submod, []).append(attr)
+    for attr, submodule in exports.items():
+        by_submodule.setdefault(submodule, []).append(attr)
 
     def __getattr__(name: str) -> object:
         if name not in exports:
-            raise AttributeError(name)
-        submod = exports[name]
-        module = importlib.import_module(submod, package=package_name)
+            raise AttributeError(f'module {package_name!r} has no attribute {name!r}')
+        submodule = exports[name]
+        module = importlib.import_module(submodule, package=package_name)
         pkg = sys.modules[package_name]
-        for attr in by_submodule[submod]:
+        for attr in by_submodule[submodule]:
             setattr(pkg, attr, getattr(module, attr))
         return getattr(module, name)
 
--- a/nicegui/ui.py
+++ b/nicegui/ui.py
@@ -289,8 +289,6 @@ __all__ = [
     'xterm',
 ]
 
-assert set(__all__) == set(_LAZY_IMPORTS), 'All public names must be in _LAZY_IMPORTS'
-
 if TYPE_CHECKING:
     from .context import context
     from .element import Element as element
@@ -433,20 +431,18 @@ def __dir__() -> builtins.list[str]:
 
 
 def __getattr__(name: str) -> object:
-    if name in _LAZY_IMPORTS:
-        module_path, attr_name = _LAZY_IMPORTS[name]
-        module = importlib.import_module(module_path, package='nicegui')
-        value = getattr(module, attr_name) if attr_name else module
-        setattr(sys.modules[__name__], name, value)
-        return value
-    raise AttributeError(f"module 'nicegui.ui' has no attribute {name!r}")
+    if name not in _LAZY_IMPORTS:
+        raise AttributeError(f"module 'nicegui.ui' has no attribute {name!r}")
+    module_path, attr_name = _LAZY_IMPORTS[name]
+    module = importlib.import_module(module_path, package='nicegui')
+    value = getattr(module, attr_name) if attr_name else module
+    setattr(sys.modules[__name__], name, value)
+    return value
 
 
 # Eagerly import element packages with 'dist' dirs so their __init__.py registers ESM modules in the importmap.
 for _module_path in {mp for mp, _ in _LAZY_IMPORTS.values()}:
     _spec = importlib.util.find_spec(_module_path, package='nicegui')
-    if _spec and _spec.submodule_search_locations:
-        _pkg_dir = Path(_spec.submodule_search_locations[0])
-        if (_pkg_dir / 'dist').is_dir():
-            importlib.import_module(_module_path, package='nicegui')
-del _module_path, _spec, _pkg_dir  # pylint: disable=undefined-loop-variable
+    if _spec and _spec.submodule_search_locations and (Path(_spec.submodule_search_locations[0]) / 'dist').is_dir():
+        importlib.import_module(_module_path, package='nicegui')
+del _module_path, _spec  # pylint: disable=undefined-loop-variable
--- a/tests/test_lazy_imports.py
+++ b/tests/test_lazy_imports.py
@@ -1,4 +1,3 @@
-import importlib
 import subprocess
 from textwrap import dedent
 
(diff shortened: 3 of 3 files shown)


evnchn:
Currently, anything originally imported using `from nicegui import ui`, such as `ui.label`, errors out twice when you access it:

```py
from nicegui import ui

ui.label # surprisingly this line is fine
ui.label # this line errors out
```

```py
from nicegui import ui

@ui.page('/')
def main_page():
    ui.label('Hello, World!')

ui.run()
```

But anything imported like `from nicegui.ui import label` is fine. 

```py
from nicegui.ui import label

label
label 
```

```py
from nicegui.ui import label, page, run

@page('/')
def main_page():
    label('Hello, World!')

run()
```

I believe I am on to something in this PR, but the code is broken somewhere.

evnchn:
Reading the docs again:

> [!IMPORTANT]
> Lazy attributes have to be proper attributes (defined in `__init__.py`) of the module they are imported from and cannot be submodules.
> In this example, if `function` is a submodule of `absolute.package`, the lazy module won't (try to) load `absolute.package.function` as a module by itself; in this case, an access of the lazy module's attribute `function` might fail.

`label` is a sub-module of `nicegui.ui`, and so the note applies exactly to us ...

evnchn:
https://github.com/zauberzeug/nicegui/pull/5303/commits/f41557588dfcbe0bf3735211dc05007c0ef04901 fixed the double-access on `ui.label` breaking. 

Now the basic code works:

```py
from nicegui import ui

ui.label('First label')
ui.label('Second label. Did not break!')

ui.run()
```

As always pipeline will be our guiding star. Let's see tomorrow.

evnchn:
Pipeline's less dramatic than I think. 2 things are broken, mainly:

1. `ui.clipboard` totally unavailable (I think this is _truly_ what the docs warned us)
2. `importmap` is empty from the start, and so `ui.mermaid` and `ui.echarts` does not work. 

Actually point 2 can theoretically affect someone who's

- creating custom elements, 
- expect to use importmap for importing your stuff, 
- and 🤡 enough to _dynamically_ define the element's class. 

Gist is: with `lazy-imports` what we call 🤡 before is actually the 🧠 move empowering lazy import's efficiency.

evnchn:
Well, as of https://github.com/zauberzeug/nicegui/pull/5303/commits/425cb4fa46a3512dcd8931299bf761c51155e900 the only failing tests is the `ui.clipboard` one. 

@falkoschindler This is getting somewhere, somewhere farther than anywhere we had been. Consider closing:
- #3365 
- #4429 
- #4471

evnchn:
If in https://github.com/zauberzeug/nicegui/pull/4564 we swapped a function for a Class, may we in this case swap a module for a Class and then we can solve issue 1.

falkoschindler:
@evnchn I added the other PRs to the description of this PR. If this PR converges and is merged eventually, we can close them.

evnchn:
Actually this PR is more mergable if we incorporate https://github.com/evnchn/nicegui/tree/feature/lazy-imports-clipboard-class @falkoschindler can you review the changes there and then if it looks good we can merge that into this PR

evnchn:
Crazily efficient 💯 

image

Running:

```py
from nicegui import ui


@ui.page('/')
def page():
    ui.highchart({'series': {'data': [1.0]}}, type='stockChart', extras=['stock'])


ui.run(show=False, reload=False)
```

falkoschindler:
@evnchn What's the status of this PR?
As I just described in #5684, it's currently our best bet to improve the initialization of multiprocesses.

evnchn:
@falkoschindler I chose to rebase because the merge conflict is unreal. Also nobody remember what happened those 98 days ago, so rebase over merge is more similar to how our memory works 🧠 

Nevertheless, I want to report: Despite what the original description says, this PR is **working**!

`test.py` (frankly you can put whatever)

```py
from nicegui import ui

@ui.page('/')
def page():
    ui.json_editor({})

ui.run(show=False, reload=False)
```

`time_script.sh`

```bash
#!/bin/bash

source /workspaces/nicegui/.venv/bin/activate

start=$(date +%s.%N)

coproc python test.py

while read -r line <&${COPROC[0]}; do
    echo "$line"
    if [[ "$line" == *"NiceGUI ready to go on"* ]]; then
        end=$(date +%s.%N)
        elapsed=$(awk "BEGIN {print $end - $start}")
        echo "Time taken: $elapsed seconds"
        kill $COPROC_PID
        break
    fi
done
```

- Before the PR: 0.6 seconds
- After this PR: 0.35 seconds

evnchn:
Would consider that this is not draft. If you agree, you can mark it ready for review and review it.

falkoschindler:
As @evnchn wrote (privately): "maybe early 3.8 is better than late 3.7" 👍🏻

evnchn:
After all, the PR require changes in `nicegui/ui.py`. One wrong move and it's a critical bug.

evnchn:
Need static `__all__` for ruff/IDE compatibility, but https://github.com/evnchn/nicegui/pull/80 should be good.

falkoschindler:
We're getting there... But I'm not very happy about the 10 exceptions in client.py. Here's an idea:

---

### Proposal: Lightweight ESM registration via element `__init__.py`

#### Problem

With lazy imports in `ui.py`, the 10 importmap-based elements (AgGrid, CodeMirror, EChart, etc.) are no longer imported at startup. But their ESM modules must be in the global `esm_modules` registry before any page is served — because the `<script type="importmap">` is baked into the initial HTML and can't be amended later (elements can be created dynamically after page load).

The current solution eagerly imports all 10 element **classes** in `client.py`, which triggers `__init_subclass__` → `register_esm()`. This works but defeats lazy loading for these elements and is labeled as a "HACK."

#### Idea

Each ESM element is already a **package** (directory with `__init__.py`). Currently those `__init__.py` files just re-export the class:

```python
# nicegui/elements/aggrid/__init__.py (before)
from .aggrid import AgGrid
__all__ = ['AgGrid']
```

We can make `__init__.py` the lightweight registration point — it calls `register_esm()` directly using local path info, then lazy-loads the heavy class only on demand:

```python
# nicegui/elements/aggrid/__init__.py (after)
from pathlib import Path
from ...dependencies import register_esm

# Register ESM module — cheap, no heavy imports needed
register_esm('nicegui-aggrid', Path(__file__).parent / 'dist')

def __getattr__(name: str):
    if name == 'AgGrid':
        from .aggrid import AgGrid
        return AgGrid
    raise AttributeError(name)

__all__ = ['AgGrid']
```

#### Auto-discovery from `ui.py`

Rather than maintaining a hardcoded list of ESM elements in `client.py`, `ui.py` already knows all module paths via `_LAZY_IMPORTS`. It can detect which ones are packages and eagerly import just their (lightweight) `__init__.py`:

```python
# At the end of ui.py — auto-import element packages for ESM registration
for _module_path, _ in _LAZY_IMPORTS.values():
    if (Path(__file__).parent / _module_path.lstrip('.').replace('.', '/')).is_dir():
        importlib.import_module(_module_path, package='nicegui')
```

- `.elements.button` → resolves to `nicegui/elements/button.py` → not a directory → skipped
- `.elements.aggrid` → resolves to `nicegui/elements/aggrid/` → is a directory → import its `__init__.py`

This means `client.py` drops all 10 imports entirely — **zero knowledge** of ESM elements. New ESM element packages are auto-discovered with no list to maintain. The `is_dir()` calls are just stat syscalls — negligible cost for ~137 entries.

#### Required supporting change

`register_esm` creates an `EsmModule` dataclass whose `__post_init__` asserts name uniqueness. When the class is later lazily imported, `__init_subclass__` calls `register_esm()` again with the same name → assertion failure. Fix: make registration idempotent if `name` and `path` match.

#### Pros

- **ESM metadata stays with the element** — declared in the element's own package, not in `client.py`
- **Auto-discovery** — `ui.py` detects packages generically; no hardcoded list to maintain
- **No heavy imports at startup** — `__init__.py` only imports `pathlib.Path` and `dependencies.register_esm`; the class (and its transitive imports like pandas detection, pyecharts, asyncio, etc.) is deferred
- **User elements are unaffected** — `__init_subclass__` still works for custom elements
- **Removes the "HACK" from `client.py`** — `client.py` has zero knowledge of ESM elements

#### Cons

- **10 `__init__.py` files gain complexity** — they go from 2-line re-exports to ~15 lines with `__getattr__` + registration
- **Dual registration path** — built-in elements register in `__init__.py`, user elements register via `__init_subclass__`; `register_esm` must be made idempotent to handle both
- **Marginal real-world gain** — startup saves ~10 class definitions, but the saving is only meaningful if the user never actually uses these elements in a given session

falkoschindler:
Let's postpone to 3.9. There are still some things to do, decisions to be made and tests to be made.

evnchn:
Haha, the "Lightweight ESM registration" is exactly what I have in mind but I left it because back then I didn't use so much AI.

evnchn:
Since this has been in milestone 3.7, 3.8, 3.9, 3.10, I am thinking more like "hmm this might just fit in 4.0 better". Not sure. 

Again, for this PR it has to be in an early milestone to catch issues, and we can make it if we start **now**, but any later then it's not "early 3.10" anymore.

evnchn:
Addressed the remaining review items:

- **Pylint E0603**: Added `TYPE_CHECKING` import blocks to all 10 ESM `__init__.py` files so pylint sees the exported names statically. Pylint now scores 10/10.
- **Tests**: Added `tests/test_lazy_imports.py` — verifies all `__all__` names resolve, `dir(ui)` is correct, bad attributes raise `AttributeError`, and `_LAZY_IMPORTS`/`__all__` stay in sync.
- **ESM path fragility**: Replaced `lstrip('.').replace('.', '/')` with `importlib.util.find_spec()` for correct relative import resolution.
- **Removed `# pylint: disable=undefined-all-variable`** from all 10 ESM `__init__.py` files (no longer needed with the `TYPE_CHECKING` imports).
- **PR description**: Still needs updating via the GitHub UI — will do that separately.

evnchn:
@falkoschindler Tests passes, and the review items are done to my eyes. See you in early 3.11!

falkoschindler:
Had a small chat with Claude about the overall idea. After arguing against it at first, we came to the agreement that the reduced startup times justify the added complexity. 👍🏻

---

## Is it worth going for lazy imports?

**Yes, for two key use cases:**

- **Hot reload during development.** The import cost isn't one-time -- it's paid on every code change. Saving ~0.3s per reload cycle adds up significantly over a development session.
- **Packaged apps** (PyInstaller etc.) import everything twice due to multiprocessing spawn, and cold-import cost is higher from a frozen archive. The ~0.3s becomes ~0.6s+ of visible delay before the app window appears -- that's user-facing latency (see #5944).

**Concerns to watch:**

- **Maintenance cost.** Every new element needs to be added in three places (`_LAZY_IMPORTS`, `__all__`, `TYPE_CHECKING` block) instead of one import line. The sync test catches drift, but it's still more friction for contributors.
- **Complexity.** The lazy version adds `_LAZY_IMPORTS` dict, `__getattr__`, `__dir__`, `TYPE_CHECKING` block, eager ESM loop, and `setup_esm_package` helper across multiple files -- a real step up from the current simple imports.

**Verdict:** The maintenance cost is the main thing to watch, but the hot-reload and packaged-app use cases justify the trade-off.

PR #5307, opened 2025-10-20
Faster pipeline via tasks parallelization

evnchn:
### Motivation

We know that `pytest-xdist` unreliable from pain. Still, want faster pipeline. 

Question being: why must startup tests be ran after the pytest? 

This PR challenges that assumption, and the pipeline speed went from 33mins to 20mins. Great!

### Implementation

* Added a new `scope` dimension to the matrix strategy, allowing jobs to run either `pytest` or `startup` tests independently.

* Updated step conditions to run only when appropriate for the current `scope`. 

### Progress

> Pipeline results: https://github.com/evnchn/nicegui/actions/runs/18659245087


evnchn:
And, if we chop up the pytest somewhat cleanly into half (by alphabet?) then the pipeline should be done in 10 seconds. 

Edit: Yes. https://github.com/evnchn/nicegui/tree/refs/heads/even-faster-pipeline-tasks-parallelization.

falkoschindler:
I understand the motivation to shorten the overall time it takes to get test results. But I'm a bit sceptical if we should simply throw more compute on it. By parallizing pytests and startup tests, we're doubling the required number workers. So what feels like a reduction in computation, still takes the same amount of CPU time, energy, heat. Therefore I'm thinking about alternatives.

So why not going one step further and questioning the need for startup tests: We do want to run them from time to time, but we might not need them on every push. What if we only run them as part of the release workflow? Sure, it's too late to fix a bad PR. But - in the rare case of a failing startup test - the release would be cancelled and we can create a bugfix PR. And maybe there is a way to manually trigger startup tests on GitHub in case we want to check a critical PR during development.

evnchn:
Yes indeed this spreads out the compute across more workers for the apparent speedup. 

I still find the startup tests not very useful though. Since you just start the server then shut it down, even if the page function contains erroneous code it is not caught because the page isn't visited. 

Startup tests are still useful in ensuring that the examples at least work to _some degree_ and help us change the examples if we ship a breaking change to the API though.

falkoschindler:
> Since you just start the server then shut it down, even if the page function contains erroneous code it is not caught because the page isn't visited.

That's a good point! After removing the auto-index client and migrating most examples to using page functions, starting them doesn't even generate UI. It basically only checks imports. Should we ditch them altogether?

evnchn:
Some examples do have tests. We can run those. All others we can basically skip them.

evnchn:
But then spreading the tests across more workers are still compelling if you want to know the results sooner. 

And that's taken to the next level in https://github.com/zauberzeug/nicegui/pull/5307#issuecomment-3423062546

falkoschindler:
@evnchn We thought about the following plan to make startup tests useful again:

1. We should try running Selenium as part of test\_startup.sh to actually request the index page and create a websocket connection. If anything fails while creating the UI, we'll get unexpected log messages like before.
2. "On push" we should run startup tests for the _oldest_ Python version only (in parallel to pytests). This should be enough to spot mistakes while developing on a feature branch.
3. As part of the release pipeline we should run startup tests for _all_ Python versions as a last line of defence against extra crazy bugs (in parallel to pytests).

What do you think? Do you think it's worth looking into Selenium before merging this PR and continuing on another one?

evnchn:
1 and 3 looks good (when combined: 2.x level of guarantee before release)

Personally I don't see a particular strong point for point 2, but I think you two have talked over it and maybe you're thinking of something that I missed. 

So yeah we can proceed with this direction. 

A bit occupied these days so maintainers feel free to takeover and anyone who wants to work on this feel free to PR into my fork.

evnchn:
Retrospect: https://github.com/zauberzeug/nicegui/pull/5307#issuecomment-3443022728 is valid but we may consider addressing this in another PR?

evnchn:
@falkoschindler Sudden revelation time: Do you know what we _really_ miss? A merge queue! 

We had that back when I used GitLab, but it was called a merge train, and I didn't put the two together. 

I was brought aware when I checked out Mermaid on mobile GitHub which shows "merge queue". 

Check https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

So, the plan is: 

- Run test only on Py3.9 on push. Possibly also Py3.14. 
- Merge queue will run the full test, pytest and startup tests, so that our main is always green. 
- Then, we don't need the pre-release full test (which may cause technical debt as we find many things broken).

falkoschindler:
@evnchn That's an awesome idea! It would greatly improve our test strategy, but also simplify the overall development process. Right now I'm waiting for the tests of PR #5422 before I can merge. With a merge queue - as far as I understand - I could simply queue the merge and the CI pipeline will take care of merging if and only if all tests are green.

What would be the next steps? I guess you could update the workflows accordingly and we would need to setup the merge queue in GitHub?

evnchn:
Yes. Gone are the days of staring at the pipeline, wait for it to be green, and then merge. 

But if your tests are flaky then I (think) it would forbid merge unless you bypass the merge queue. 

Anyways we'll see.

evnchn:
Nah I think want to test Py3.9 only, since Py3.14 should be _mostly_ backwards compatible. There are exceptions (`PicklingError`, if you recall) but those are rare and caught by merge queue. 

But then, we are having 6 workers before, so I am thinking that we keep the old number of workers, but then (1) pytest-xdist with retry is coming back (3x+ speed improvement) and (2) we're spreading all the way to 6 workers per python version (6x speed improvement). 

Napkin math would suggest the new per-push pipeline completing in 2 minute, at which point the environment setup is the bottleneck 🚀 

Sounds exciting 🪨

falkoschindler (approved):
This turned into a huge refactoring of our CI configuration files. Let me quote a summary by Sonnet 4.5:

> ### 1. **Perfect Modularity**
> ```
> Reusable Components (_*.yml):        Orchestrators:
> ├── _check.yml (3 parallel checks)   ├── test-on-push.yml (fast feedback)
> ├── _test.yml (matrix testing)       ├── test-on-merge.yml (full validation)
> ├── _slack.yml (notifications)       └── publish-on-tag.yml (release)
> ├── _pypi.yml (PyPI + GH release)
> ├── _docker.yml (multi-arch builds)
> ├── _citation.yml (CITATION.cff)
> └── _verify.yml (container check)
> ```
> 
> Each component has **one clear responsibility** and can be tested/modified independently.
> 
> ### 2. **Excellent Performance Strategy**
> - **On push**: 4 test workers (Py3.9, 3.14 × pytest only) → ~8-10 min
> - **On merge**: 12 test workers (all versions × pytest+startup) → ~20 min
> - **Code checks**: 3 parallel checks instead of sequential → ~2-3x faster
> 
> ### 3. **Clean Naming Convention**
> - `_*.yml` = Reusable components (clear visual distinction)
> - `*-on-{event}.yml` = Event-triggered orchestrators
> - Instantly recognizable structure
> 
> ### 4. **Smart Merge Queue Strategy**
> - Fast feedback during development
> - Comprehensive validation before merge
> - Main branch always green
> - No more manual waiting for CI
> 
> ### 5. **Excellent Code Quality**
> - No redundant names (using good defaults)
> - Minimal syntax (`on: push` not `on: [push]`)
> - Parameterized matrix for checks (easy to extend)
> - Simplified secret checking (1 line vs 10)

Personally I like how the three "orchestrators" define what jobs are triggered when (on push, merge or tag), while reusable components define what exact steps are executed. This makes the code readable and manageable again.

After merging this PR on main, I'll setup the merge queue so that upcoming PRs will be able to use it.

evnchn:
Please be extra careful with Claude. It ranks high in the 🤡 score. 

I trust that it works, but I disagree with `_*.yml = Reusable components (clear visual distinction)` since the titles don't distinct themselves). 



And for components used only once, I don't know if we should break it down so much. Especially since this will cause spawning many many workers? 

But the reuse of test component for merge test and push test, that's just so good though.

falkoschindler:
> clear visual distinction

I was more focused on the file list. But you're right: The displayed title could be more distinct. Actually I didn't expect the `_*.yml` files to appear as workflows. Anyway, I pinned the three entry points:

Screenshot 2025-11-13 at 16 33 20

> many many workers

I don't think so. It's just breaking up large workflow files into smaller chunks. Even though some of them are only used once, I like that you can, e.g., focus on \_docker.yml without worrying about when and how it's triggered and what else might happen in the larger workflow.

evnchn:
image

beyond wonderful

PR #5311, opened 2025-10-21
Migrate to uv

falkoschindler:
### Motivation

I'd like to see how our project setup would look like based on uv.

### Implementation

This PR is almost 100% AI-generated:

1. **`pyproject.toml`** - Converted from Poetry format to PEP 621 standard format:
   - Changed `[tool.poetry]` to `[project]`
   - Converted dependencies to standard format
   - Moved optional dependencies to `[project.optional-dependencies]`
   - Created `[dependency-groups]` section for dev dependencies (uv standard)
   - Updated build system to use `hatchling` instead of Poetry
   - Preserved all important security and version comments

2. **`.github/workflows/test.yml`** - Updated to use uv:
   - Replaced `abatilo/actions-poetry@v4` with `astral-sh/setup-uv@v7`
   - Removed `cache: "poetry"` (uv handles caching automatically)
   - Changed `poetry install` to `uv sync --all-extras --group dev`
   - Changed `poetry run` to `uv run`

3. **`.github/workflows/publish.yml`** - Updated to use uv:
   - Replaced Poetry setup with uv
   - Changed version setting to use `sed` instead of `poetry version`
   - Updated build command to `uv build`
   - Updated publish command to `uv publish` with `UV_PUBLISH_TOKEN`

4. **`fly.dockerfile`** - Updated to use uv:
   - Removed Poetry installation
   - Added uv installation from official Docker image
   - Changed `poetry install` to `uv sync` with appropriate flags
   - Updated to use `uv pip install --system`

5. **`development.dockerfile`** - Updated to use uv:
   - Removed Poetry installation
   - Added uv installation
   - Changed to `uv sync --all-extras`

6. **`CONTRIBUTING.md`** - Updated documentation:
   - Added uv installation instructions
   - Changed setup command from `pip install -e .` to `uv sync --all-extras --group dev`

7. **`poetry.lock`** - Deleted (replaced with `uv.lock`)

8. **`uv.lock`** - Generated with all dependencies resolved (164 packages)

### Progress

    - [x] Let's wait for the build pipeline.

### DIFF
--- a/test_startup.sh
+++ b/test_startup.sh
@@ -45,12 +45,12 @@ do
     fi
 
     # Skip examples/sqlite_database for Python 3.11 and 3.12
-    if [[ $(python3 --version | cut -d' ' -f2 | cut -d'.' -f1,2) =~ ^3.1[12]$ ]] && [[ $path == "examples/sqlite_database" ]]; then
+    if [[ $(uv run python3 --version | cut -d' ' -f2 | cut -d'.' -f1,2) =~ ^3.1[12]$ ]] && [[ $path == "examples/sqlite_database" ]]; then
         continue # until https://github.com/omnilib/aiosqlite/issues/241 is fixed
     fi
 
     # Skip examples/ai_interface for Python 3.14
-    if [[ $(python3 --version | cut -d' ' -f2 | cut -d'.' -f1,2) =~ ^3.14$ ]] && [[ $path == "examples/ai_interface" ]]; then
+    if [[ $(uv run python3 --version | cut -d' ' -f2 | cut -d'.' -f1,2) =~ ^3.14$ ]] && [[ $path == "examples/ai_interface" ]]; then
         continue # It still uses Pydantic V1, which breaks horribly with Python 3.14
     fi
 


evnchn:
@falkoschindler good thing to give this a shot, and I appreciate your bravery and efforts in kick-starting this big change, but we may want to do this at the later end of 3.x?

One thing immediately that I need to point out because I did look into this:

> pyproject.toml - Converted from Poetry format to PEP 621 standard format

Consider `uvx migrate-to-uv` (https://github.com/mkniewallner/migrate-to-uv) because AI may not be accurate. AI have been observed to miss lines while copying stuff verbatim, even.

EmberLightVFX:
I tried migrate-to-uv before, it works great!
For NiceGUI I'm currently using UV to create a venv, `uv pip install poetry` to install poetry and then run like normal.
Personally I do like UV more than Poetry, mostly as you install UV outside a python instance so I don't need to install python on a new system and it's quite faster, but that's just me 😄

falkoschindler:
@codingpaula @pascalzauberzeug I wonder if our dynamic versioning will still work. Claude recommended to keep the "poetry-dynamic-versioning" build system. What do you think? Maybe we simply have to try it out.

pascalzauberzeug:
@falkoschindler I think you just have to install it locally to test it. Since version 3.0.0 is hardcoded, it should be easy to see

falkoschindler:
@pascalzauberzeug
```bash
uv pip install -e .
uv run python -c 'import nicegui; print(nicegui.__version__)'
```
Output:
```
3.0.4.dev0
```
This doesn't look right...

pascalzauberzeug:
@falkoschindler 😞 I will look into it

pascalzauberzeug:
There seems to be no dynamic versioning in uv

himbeles:
There is, if you use [poetry-dynamic-versioning](https://pypi.org/project/poetry-dynamic-versioning/) and the poetry build backend, as you already do. 
It is just not properly configured. As pointed out in the review above.

Alternatively, you can use `hatchling` + `uv-dynamic-versioning`.

.github/workflows/_pypi.yml
```diff
@@ -13,27 +13,32 @@ jobs:
...
-      - run: poetry publish --build
-        env:
-          POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_API_TOKEN }}
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
+      - name: Install Python
+        run: uv python install
+      - name: Update version
+        run: |
+          sed -i "/^\[project\]/,/^version = /s/version = .*/version = \"${{ inputs.version }}\"/" pyproject.toml
```
codingpaula: If we now use `dynamic = ["version"]`, this is obsolete. Or do we want to put the correct version into the `tool.poetry` section? The poetry dynamic versioning project says this in its README: "Poetry still requires the tool.poetry.version field to be present in pyproject.toml, but you are encouraged to use version = "0.0.0" as a standard placeholder."

rodja (approved):

codingpaula (approved):

evnchn:
Startup tests are very dead. 

https://github.com/zauberzeug/nicegui/actions/runs/19854770670/job/56889849424

evnchn:
Maybe it is simply:

image

codingpaula:
Just `uv pip` works for me locally, let's see if the test runs through with it. I also added `uv run` to the pytest command.

evnchn:
`python3 -m pip` runs fine locally so I can't test, hence the "maybe". 

We have no other thing to do other than to pray at the pipeline gods.

codingpaula (dismissed):

evnchn:
Python 3.14 pydantic v1 issues. 

https://github.com/zauberzeug/nicegui/actions/runs/19857070086/job/56897511745

evnchn:
Python version checking for `test_startup.sh` appears broken. 

Just that https://github.com/omnilib/aiosqlite/issues/241 is fixed so Py3.11 Py3.12 don't have extra errors.

evnchn:
We'll toss the Py3.11 and Py3.12 exclusion later (want to keep the diff minimal to avoid ALL issues), playing very cautiously hgere.

falkoschindler:
Thanks, @evnchn!
I guess the sqlite_database example still doesn't work because https://github.com/omnilib/aiosqlite/issues/241 hasn't been released. Let's keep it excluded for now.

PR #5312, opened 2025-10-21
Add dynamic versioning for Poetry

pascalzauberzeug:
### Motivation

I noticed that NiceGUI's version in the pyproject.toml always lags behind one release, because the version update is triggered by the release itself.
For example NiceGUI 3.0.4 has 3.0.3.dev0 in its pyproject.toml
https://github.com/zauberzeug/nicegui/blob/v3.0.4/pyproject.toml

In RoSys we already added in [dynamic versioning](https://github.com/mtkennerly/poetry-dynamic-versioning) with https://github.com/zauberzeug/rosys/pull/346.

### Implementation

- Add [poetry-dynamic-versioning](https://github.com/mtkennerly/poetry-dynamic-versioning) with the same configuration as we use for RoSys
- Remove the version update from the release workflow. With the current implementation CITATION.cff will still lag behind.

### Progress

    - [x] Fix Highcharts error -> We will set the version to the latest major version manually
    - [x] Fix/discuss citation.cff -> We will keep it like this for now

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -904,7 +904,7 @@ version = "1.25.0"
 description = "Dynamic version generation"
 optional = false
 python-versions = ">=3.5"
-groups = ["main"]
+groups = ["dev"]
 files = [
     {file = "dunamai-1.25.0-py3-none-any.whl", hash = "sha256:7f9dc687dd3256e613b6cc978d9daabfd2bb5deb8adc541fc135ee423ffa98ab"},
     {file = "dunamai-1.25.0.tar.gz", hash = "sha256:a7f8360ea286d3dbaf0b6a1473f9253280ac93d619836ad4514facb70c0719d1"},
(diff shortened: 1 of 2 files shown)


pascalzauberzeug:
poetry-dynamic-versioning expects you to set the version to 0.0.0, but that causes an issue because nicegui-highcharts is not compatible with 0.0.0.
Locally, I solved this problem by installing NiceGUI first when updating poetry.lock, because then the version can be evaluated correctly. We have to find a solution before this can be merged. 
```
Because no versions of nicegui-highcharts match >2.0.2,<2.1.0 || >2.1.0,<2.2.0 || >2.2.0,<3.0.0
 and nicegui-highcharts (2.0.2) depends on nicegui (==2.*), nicegui-highcharts (>=2.0.2,<2.1.0 || >2.1.0,<2.2.0 || >2.2.0,<3.0.0) requires nicegui (==2.*).
And because nicegui-highcharts (2.1.0) depends on nicegui (==2.*)
 and nicegui-highcharts (2.2.0) depends on nicegui (>=2.0.0,<4.0.0), nicegui-highcharts (>=2.0.2,<3.0.0) requires nicegui (>=2.dev0,<4.0.0).
So, because nicegui depends on nicegui-highcharts (^2.0.2)
 and nicegui is 0.0.0, version solving failed.
```

falkoschindler (approved):

PR #5335, opened 2025-10-23
Add Quasar QResponsive Component

AlePiccin:
### Motivation

Expose the Quasar QResponsive component through NiceGUI so developers can easily preserve aspect ratios while following Quasar’s single-child and ratio-string guidelines directly within the Python API and documentation.

### Implementation

Introduced a Responsive element that forwards the optional ratio prop, integrated it into the public ui namespace, and documented its constraints and usage patterns (including wrapping content in ui.card) in the layout section of the docs.

---

Feature request: #5330

### DIFF
--- a/nicegui/elements/aspect_ratio.py
+++ b/nicegui/elements/aspect_ratio.py
@@ -3,7 +3,7 @@ from ..element import Element
 
 class AspectRatio(Element):
 
-    def __init__(self, *, ratio: float) -> None:
+    def __init__(self, ratio: float) -> None:
         """Aspect Ratio
 
         This element is based on Quasar's `QResponsive <https://quasar.dev/vue-components/responsive>`_ component.
--- a/website/documentation/content/aspect_ratio_documentation.py
+++ b/website/documentation/content/aspect_ratio_documentation.py
@@ -5,7 +5,7 @@ from . import doc
 
 @doc.demo(ui.aspect_ratio)
 def main_demo() -> None:
-    with ui.aspect_ratio(ratio=16 / 9):
+    with ui.aspect_ratio(16 / 9):
         with ui.card():
             ui.icon('live_tv')
             ui.label('16:9 content')


nicegui/elements/responsive.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+
+from ..element import Element
+
+
+class Responsive(Element):
+
+    def __init__(
+        self,
+        *,
```
evnchn: 1 parameter then we can let it be positional, no need `*`, facilitate `with ui.aspect_ratio(16/9):` as mentioned in https://github.com/zauberzeug/nicegui/discussions/5330#discussioncomment-14779457

nicegui/elements/responsive.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+
+from ..element import Element
+
+
+class Responsive(Element):
```
evnchn: We are thinking of doing `ui.aspect_ratio` as in https://github.com/zauberzeug/nicegui/discussions/5330#discussioncomment-14779457. Do you think so as well? If so we can make the change.

alexandre-piccin: Sure, ui.aspect_ratio sounds good!

falkoschindler (changes requested):
Thanks a lot for this pull request, @AlePiccin!

Just a few points that need clarification:
1. What would be the use for `ui.aspect_ratio` without a `ratio`? I assume the argument is required, therefore I removed its default value.
2. The demo output looks a bit weird:
    Image
    Is there a problem in `ui.aspect_ratio`, or are we using it wrong?
3. I'm not sure if we need the warning about a single child only, or if we can wrap the content into a container automatically. But first we need to address point 2, then we can think about further improvements.

alexandre-piccin:
Given better thought, I think we should use css native aspect-ratio instead of quasar QResponsive. IA helped me explain it:

The current pull request attempted to introduce Quasar's `QResponsive` into NiceGUI. After testing it in real layouts, we found that the component does not behave the way we initially expected, and that its core design has side effects that make it awkward to integrate as a first-class NiceGUI layout primitive.

**Why `QResponsive` didn't behave as expected**

`QResponsive` is implemented using the classic "intrinsic ratio box" trick: it creates a wrapper whose dimensions are determined by a target aspect ratio, and then absolutely positions the inner content to fill that box.

Concretely:

* The outer `<q-responsive>` sets up a box with a given `ratio` (e.g. 16/9).
* The actual slot content (what we render inside it) is placed in a child element with `position: absolute; top: 0; left: 0; right: 0; bottom: 0;`.

This is important, because absolutely positioned children do not contribute to their parent's intrinsic size. In other words, from the browser's layout engine point of view, `<q-responsive>` looks "empty": its size no longer comes from its children, its size is supposed to come from the ratio math.

That leads directly to the problem we observed: when `<q-responsive>` is used inside a flex container that doesn't force a width (for example, a column with `align-items: center` instead of `stretch`), the responsive box can end up with no intrinsic width. With no width, its computed height becomes 0 as well (height is derived from width * ratio). The visual effect is that the content "disappears". It didn't actually fail to render; it just collapsed to 0×0.

The only way to make it visible in that situation is to manually force width on the component (e.g. by adding `class="full-width"` or giving it an explicit width style, or wrapping it in another full-width element). This is what we saw in practice: removing `full-width` caused the card to vanish.

**Why this is undesirable in NiceGUI**

NiceGUI encourages declarative nesting like:

```python

@doc.demo(ui.aspect_ratio)
def main_demo() -> None:
    with ui.aspect_ratio(16 / 9):
        with ui.card():
             ...
```

From the user's point of view, this looks like a normal layout block that should size itself based on its content. But `QResponsive` is not "normal": it breaks that mental model because it suppresses the child from the normal flow. The result is surprising: components randomly collapse unless you also know about Quasar-specific layout classes and flexbox edge cases.

This makes `QResponsive` fragile as a general-purpose building block in NiceGUI. It works, but only if you understand and remember to always give it an explicit width context. That’s not ergonomic.

**Why native CSS `aspect-ratio` is a better fit**

Modern CSS has first-class support for aspect ratio via the `aspect-ratio` property. We can express the same intent without relying on absolute positioning tricks:

```python
@doc.demo(ui.aspect_ratio)
def main_demo() -> None:
    with ui.card().style("aspect-ratio: 16/9;"):
        ui.icon('live_tv')
        ui.label('16:9 content')
```

Key advantages of using a native `aspect-ratio` container:

1. The element participates in normal layout flow.
   The container still has a real box model, so flexbox/grid can size and align it naturally. The child content does not need to be absolutely positioned to fill the area unless we explicitly want that.

2. No accidental collapse to 0×0.
   Because the container itself owns an intrinsic aspect ratio, browsers can derive both width and height in a predictable way. We don't rely on the child to "exist" and then immediately remove it from flow.

3. Cleaner mental model.
   Users think "this is a box that keeps a 16:9 ratio", and that’s exactly what `ui.aspect_ratio` does. There is no extra requirement like "also remember to add `full-width` or it will disappear in some flex alignments."

4. Simpler API surface for NiceGUI.
   We don't have to leak Quasar's implementation details (absolute positioning of the slot, full-width classes, etc.). We can expose a Pythonic `ui.aspect_ratio(ratio)` component that behaves intuitively across layouts.

**Conclusion**

`QResponsive` solves an older cross-browser problem by faking aspect ratios with absolute positioning. That approach has two drawbacks in NiceGUI:

* The child content is taken out of normal layout flow, so the parent can collapse unless you force width manually.
* The resulting behavior is surprising for users who expect "put card inside box" to just work.

Given that modern browsers natively support `aspect-ratio`, we get the same visual result (fixed ratio containers like 16:9) with fewer surprises, better integration with flex/grid, and a simpler mental model. For this reason, instead of integrating `QResponsive` directly, **we should prefer to use the CSS `aspect-ratio` property**.

falkoschindler:
Oh wow, thanks for the clarification, @AlePiccin!

So we can simply use plain CSS or even a Tailwind class. That's nice!
```py
with ui.card().classes('aspect-16/9'):
    ui.icon('live_tv')
    ui.label('16:9 content')
```
Screenshot 2025-10-28 at 15 21 05

I guess there's no need for this PR anymore. Maybe we just create a stand-alone demo in the ["Page Layout"](https://nicegui.io/documentation/section_page_layout) section?

alexandre-piccin:
Sure! A Demo would be helpful

PR #5338, opened 2025-10-24
Fix code extraction for documentation demos

rodja:
### Motivation

The documentation code extraction had two issues:
1. Commented `def root()` functions were incorrectly detected as actual root functions
2. Multiple blank lines appeared before `ui.run()` calls after isort processing

### Changes

- Check for root functions before uncommenting lines to avoid false positives
- Normalize whitespace to ensure exactly one blank line before `ui.run()` calls

### Progress

### DIFF
--- a/website/documentation/code_extraction.py
+++ b/website/documentation/code_extraction.py
@@ -31,16 +31,12 @@ def get_full_code(f: Callable) -> str:
 
     if has_root_function:
         code = [line for line in code if line.strip() != 'return root']
-        code.append('ui.run(root)')
-    elif not code[-1].startswith('ui.run('):
-        code.append('ui.run()')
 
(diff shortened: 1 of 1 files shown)


falkoschindler:
@rodja Can you name examples where the two issues caused problems? In other words: Where should I look to see the effect of this PR?

falkoschindler (approved):
I guess this PR is only relevant for intro demos at the moment since they are the only ones using root functions.

Anyway, I simplified the code a little. `code` is already a list of lines, so we don't need to join and split it again. I further uncoupled the individual processing steps (removing `return root`, adding `ui.run`, fixing number of blank lines) to make the code more readable.

We should think about unit-testing this function. But I'll leave this for another PR.

PR #5339, opened 2025-10-24
Allow passing event arguments to User.trigger()

rodja:
### Motivation

Testing event handlers that rely on event arguments (like click coordinates, key codes, or custom data) was not possible with the `User.find().trigger()` method. The method always passed an empty dictionary as event arguments, making it impossible to test handlers that expect specific argument values.

### Implementation

Added an optional `args` parameter to `UserInteraction.trigger()` that allows passing custom event arguments to event handlers during test simulation:

- Parameter defaults to `None` to maintain full backward compatibility
- When provided, the dictionary is passed to `GenericEventArguments` instead of an empty dict
- Type hint uses `Any` for flexibility (event args can be dict or other types depending on context)
- Documentation follows NiceGUI's standard `:param` style

**Public API Impact:** Backward compatible enhancement – existing code continues to work unchanged.

**Testing:** Added `test_trigger_with_event_arguments()` demonstrating usage with click event coordinates.

### Progress

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -50,11 +50,7 @@ class UserInteraction(Generic[T]):
                 for listener in element._event_listeners.values():  # pylint: disable=protected-access
                     if listener.type != event:
                         continue
-                    event_arguments = events.GenericEventArguments(
-                        sender=element,
-                        client=self.user.client,
-                        args=args,
-                    )
+                    event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args=args)
                     events.handle_event(listener.handler, event_arguments)
         return self
 


nicegui/testing/user_interaction.py
```diff
@@ -29,10 +29,11 @@ def __init__(self, user: User, elements: set[T], target: str | type[T] | None) -
         self.elements = elements
         self.target = target
 
-    def trigger(self, event: str) -> Self:
+    def trigger(self, event: str, args: Any = None) -> Self:
         """Trigger the given event on the elements selected by the simulated user.
 
-        Examples: "keydown.enter", "click", ...
+        :param event: the event type to trigger (e.g. "keydown.enter", "click", ...)
+        :param args: optional event arguments to pass to the handler (default: None)
```
falkoschindler:
```suggestion
        :param args: optional event arguments to pass to the handler (default: ``None``)
```

nicegui/testing/user_interaction.py
```diff
@@ -47,7 +48,11 @@ def trigger(self, event: str) -> Self:
                 for listener in element._event_listeners.values():  # pylint: disable=protected-access
                     if listener.type != event:
                         continue
-                    event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args={})
+                    event_arguments = events.GenericEventArguments(
+                        sender=element,
+                        client=self.user.client,
+                        args=args if args is not None else {}
```
falkoschindler:
```suggestion
                        args=args or {},
```

falkoschindler: If we replace the ternary expression with `args or {}`, we can probably bring it back into one line...

rodja: Unfortunately its still to long...

nicegui/testing/user_interaction.py
```diff
@@ -47,7 +48,11 @@ def trigger(self, event: str) -> Self:
                 for listener in element._event_listeners.values():  # pylint: disable=protected-access
                     if listener.type != event:
                         continue
-                    event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args={})
+                    event_arguments = events.GenericEventArguments(
+                        sender=element,
+                        client=self.user.client,
+                        args=args or {},
```
evnchn: I cannot pass `False` or 0 or any other falsy args, then.

Maybe `args if args is not None else {}` just to be on the safe side?

CONTRIBUTING.md
```diff
@@ -241,6 +241,7 @@ If you plan to implement a new element you can follow these suggestions:
 8. Add a documentation file in `website/documentation/content`.
    By calling the `@doc.demo(...)` function with an element as a parameter the docstring is used as a description.
    The docstrings are written in restructured-text.
+   Make sure to use the correct syntax (like double backticks for code).
```
evnchn, line 244: OFF TOPIC: I think for minor and simple-win changes into docs, (additions only, can't do any harm), since you are maintainer, you can just commit directly to main. 

We can see the commits anyways.

evnchn:
Wait but my suggestion is opposite of https://github.com/zauberzeug/nicegui/commit/e3bb26868597ffae3322bc485fa55453203d7a8a

@falkoschindler is passing falsy args a possible thing in general? Off my head I feel like it is a possibility with the new and improved `js_handler` nowadays.

evnchn:
This breaks:

```py
async def test_trigger_with_event_arguments(user: User) -> None:
    @ui.page('/')
    def page():
        ui.button('Click').on('click', lambda e: ui.notify(
            f'{e.args} with type={type(e.args)}'), js_handler='() => emit(false)')

    await user.open('/')
    user.find('Click').trigger('click', args=False)
    await user.should_see("False with type=<class 'bool'>")
```

Before you ask:

image

nicegui/testing/user_interaction.py
rodja: You are right, @evnchn. I made sure all typical values are handled correctly in e6215cc408de4b35f82bbf5841c087aa1fbcfd3e

CONTRIBUTING.md
rodja: Sure. Just in this case the info came up in the review so I thought it would be appropriate to put it into the PR.

nicegui/testing/user_interaction.py
```diff
@@ -47,7 +50,11 @@ def trigger(self, event: str) -> Self:
                 for listener in element._event_listeners.values():  # pylint: disable=protected-access
                     if listener.type != event:
                         continue
-                    event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args={})
+                    event_arguments = events.GenericEventArguments(
+                        sender=element,
+                        client=self.user.client,
+                        args=args,
+                    )
```
evnchn: Totally a nitpick and just personal preference at this point: 

I counted `                    event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args=args)` to be **119** characters, should just be enough to fit in 1 line as @falkoschindler originally desired. 

But then the code may lose the readability at that point, so actually I am sitting on the fence on whether to apply this (after all I could have commited to this branch but now I am writing this comment)

nicegui/testing/user_interaction.py
falkoschindler: Ooops, my bad.

falkoschindler (approved):

PR #5344, opened 2025-10-24
Add `fps` to `ui.scene` (3.0)

evnchn:
### Motivation

Who doesn't want more FPS than 20, but #4825 is too far gone as it is based off 2.x branch code...

### Implementation

- Replace `20` with `this.fps`
- Add `fps` option
- Add documentation

### Progress

-> Established in #4825 pytest is not necessary.

### DIFF
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -552,6 +552,6 @@ export default {
     click_events: Array,
     drag_constraints: String,
     background_color: String,
-    fps: { type: Number, default: 20 },
+    fps: Number,
   },
 };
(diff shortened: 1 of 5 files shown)


evnchn:
Adding `show_stats` is easier than expected. Check https://github.com/evnchn/nicegui/pull/52

evnchn:
> parameters are positional

Yes I find that weird as well, because I wanted to put fps after the width and height, but I can't do that right now. 

Anyways here's the real question: Where you put the deprecation marker, leads me to believe that width and height will be positional in `ui.scene` but not `ui.scene_view`, which is a bit strange and may be confusing?

falkoschindler:
> Where you put the deprecation marker, leads me to believe that width and height will be positional in `ui.scene` but not `ui.scene_view`, which is a bit strange and may be confusing?

You're right. I misplaced one of the markers. I'll certainly think about it once again when doing the change next year... 🤞🏻

PR #5345, opened 2025-10-25
Basic CSP support

evnchn:
### Motivation

CSP is good for web security as it reduces the attack vector. 

There are attempts at #4864 but without library modifications into NiceGUI you will get a DeadGUI instead 💀  

### Implementation

- Use SebWeb (https://pypi.org/project/Secweb/)
- No HSTS because (1) we are http, (2) breaks websocket for some reason (room for need issue report)
- Add nonce to script

### Progress

  - [x] Inline scripts (not styles, though)
  - [ ] Vue.js template compiler
- [ ] **Put the functionality behind a switch**
  - [ ] Improve the pytest so that it tests under CSP environment.

-> Goal as of writing: Centralize the effort at adding CSP. Check the pipeline results.


evnchn:
Stuck at "Vue.js template compiler" problem.

evnchn:
Consider soft-blocked by #5351 due to whatever template compiled will eventually have `import { BLAH } from "vue";` which can't work. 

Hacks can be done to change the compiled output into `const { BLAH } = Vue;` but why hack when you can do it the proper way.

falkoschindler:
Since PR #5351 is merged and release in version 3.3, I think we can unblock this PR.

evnchn:
@falkoschindler Please look at https://github.com/zauberzeug/nicegui/pull/5345/commits/e37b41ea31600222fbb3444cf5ed1d82ade07372. It works. Tests are passing too. 

Might want to stop here and review before I go any further on the other elements, I figured.

evnchn:
I tried taking https://github.com/zauberzeug/nicegui/pull/5345/commits/e37b41ea31600222fbb3444cf5ed1d82ade07372 beyond `ui.audio` and took `ui.xterm` for my first shot, then got stuck. 

@frankier We may need your help to migrate elements to the syntax enabled by https://github.com/zauberzeug/nicegui/pull/5351, namely `examples/vue_vite`. Though I am not sure since for some elements we're using Rollup already. 

Anyways I think this PR will be a massive one and definitely not simple-win by any imaginations. Let's let this cook for a long time.

falkoschindler:
> Let's let this cook for a long time.

@evnchn So what's that status - who is cooking? I fear that the PR gets stale and outdated pretty quickly and the chance of merging it declines. At the same time, I'm _so_ unfamiliar with the whole CSP topic that I won't be of much help finishing it. I can only promise a thorough review once the implementation is complete.

evnchn:
> who is cooking

Sorry I tried but there's like 5 failed attempts with me and GitHub copilot 😭. And so nothing was pushed...

I am familiar with CSP now but the issue is with Vue and the Vite/Rollup, which former I am unfamiliar and latter I know nothing frankly. 

We can probably leave this PR open since I don't se it diverging into unmergable territory unless 4.0 comes around. Or we can close it. 

But please, whoever wants CSP for NiceGUI probably needs to fight for it themselves seeing that they would know their stuff when they asked about it. 

CC: @benvc

evnchn:
Considering limited bandwidth and expertise, here is a proposal:

- CSP support released as alpha-stage, behind-a-switch, untested, if it works it works for each element, and may be removed entirely in a minor version. 
- We can maintain a list of elements which work and which doesn't work from community feedback. 
- For elements which doesn't work, PRs are welcome, but we will not actively make them work unless we have much time on our hands, which we don't.

evnchn:
Note that there is no point in shipping something half-baked in main as prescribed in https://github.com/zauberzeug/nicegui/pull/5345#issuecomment-3580469903, because a trick I keep forgetting is that you can install NiceGUI to a branch. To get up-to-date CSP shenanigans, simply `pip install git+https://github.com/evnchn/nicegui.git@CSP`.

And unfortunately, I think that, without CSP and Vue experts on board, this PR will not move very far. As I am unable to carry-forward the devwork, I am closing this PR.

PR #5349, opened 2025-10-25
Add `show_stats` to `ui.scene` (3.0)

evnchn:
### Motivation

Originally I guessed that Falko would like high-FPS `ui.scene` for the RoSys robots for their projects, this is why he has looked specifically at #4825 out of the many "feat" PRs that I made was the time. 

So then I did #5344 which standalone the FPS parameter adjustment. 

But then, in #4825 we got away with a more concise documentation by documenting the stat and the FPS together. Therefore I thought we should do this in time for 3.2

### Implementation

Take the logic from #4825 and apply to here. 

- `this.show_stats` props`
- Spawn and update the element appropriately
- Change the packed JS

### Progress

### DIFF
--- a/nicegui/elements/scene/scene.py
+++ b/nicegui/elements/scene/scene.py
@@ -71,7 +71,7 @@ class Scene(Element, component='scene.js', esm={'nicegui-scene': 'dist'}, defaul
                  drag_constraints: str = '',
                  background_color: str = '#eee',
                  fps: int = 20,
-                 show_stats: bool = False
+                 show_stats: bool = False,
                  ) -> None:
         """3D Scene
 
@@ -91,7 +91,7 @@ class Scene(Element, component='scene.js', esm={'nicegui-scene': 'dist'}, defaul
         :param drag_constraints: comma-separated JavaScript expression for constraining positions of dragged objects (e.g. ``'x = 0, z = y / 2'``)
         :param background_color: background color of the scene (default: "#eee")
         :param fps: target frame rate for the scene in frames per second (default: 20, *added in version 3.2.0*)
-        :param show_stats: whether to show performance stats (default: False)
+        :param show_stats: whether to show performance stats (default: ``False``, *added in version 3.2.0*)
         """
         super().__init__()
         self._props['width'] = width
--- a/website/documentation/content/scene_documentation.py
+++ b/website/documentation/content/scene_documentation.py
@@ -193,7 +193,8 @@ def scene_views():
 
 @doc.demo('Frame Rate and Statistics', '''
     You can configure the target frames per second (FPS) of the scene using the `fps` argument.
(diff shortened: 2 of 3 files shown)


evnchn:
@falkoschindler preference time: `show_stats` or `stats`?

nicegui/elements/scene/scene.py
```diff
@@ -71,6 +71,7 @@ def __init__(self,
                  drag_constraints: str = '',
                  background_color: str = '#eee',
                  fps: int = 20,
+                 show_stats: bool = False
```
falkoschindler:
```suggestion
                 show_stats: bool = False,
```

nicegui/elements/scene/scene.py
```diff
@@ -90,11 +91,13 @@ def __init__(self,
         :param drag_constraints: comma-separated JavaScript expression for constraining positions of dragged objects (e.g. ``'x = 0, z = y / 2'``)
         :param background_color: background color of the scene (default: "#eee")
         :param fps: target frame rate for the scene in frames per second (default: 20, *added in version 3.2.0*)
+        :param show_stats: whether to show performance stats (default: False)
```
falkoschindler:
```suggestion
        :param show_stats: whether to show performance stats (default: ``False``, *added in version 3.2.0*)
```

website/documentation/content/scene_documentation.py
```diff
@@ -191,22 +191,22 @@ def scene_views():
             scene_view2.move_camera(x=0, y=4, z=3)
 
 
-@doc.demo('Frame Rate', '''
+@doc.demo('Frame Rate and Statistics', '''
     You can configure the target frames per second (FPS) of the scene using the `fps` argument.
-    The default value is 20.
+    The default value is 20. To see the changes for yourself, enable the statistics display using the `show_stats` argument.
```
falkoschindler:
```suggestion
    The default value is 20.
    To see the changes for yourself, enable the statistics display using the `show_stats` argument.
```

falkoschindler (commented):
Thanks, @evnchn!

There's just a tiny issue: Now there are too many scene demos on one page. When you scroll all the way to the bottom of the page, JavaScript says "WARNING: Too many active WebGL contexts. Oldest context will be lost." And indeed, the first demo doesn't render anymore.

This is probably not caused by this PR, but by PR #5344. But we should look into it. Maybe we can increase the limit?

falkoschindler:
> @falkoschindler preference time: `show_stats` or `stats`?

Oh I like `show_stats` 👍🏻

evnchn:
> WARNING: Too many active WebGL contexts. Oldest context will be lost.

IntersectionObserver when too far away should disable the rendering. A bit lost as to how your IntersectionObserver is done though (did use IntersectionObserver but mine was not in Python-land, it was in JS-land)

PR #5350, opened 2025-10-25
Mount `ui.xterm` in sync to avoid `this.terminal` undefined error

### LINKED ISSUE #5343: First `ui.xterm` demo is broken on the website
### First Check

### Example Code

```python
terminal = ui.xterm({'cols': 30, 'rows': 9})
ui.timer(0.1, lambda: terminal.write('Hello NiceGUI!'), once=True)
```

### Description

Visit https://nicegui.io/documentation/xterm and reload the page without cache.
The terminal window is blank, "Hello NiceGUI!" is not visible and there is a JavaScript error "Cannot read properties of undefined (reading 'write')".

### NiceGUI Version

3.1.0

### Python Version

3.11

### Browser

Chrome

### Operating System

macOS

### Additional Context

The problem was reported by u/jugglist on Reddit: https://www.reddit.com/r/nicegui/comments/1odaihz/comment/nkzzzp2/

evnchn:
### Motivation

Fix #5343, where since the methods are sync but mount is async, we "mount too late" and methods cannot write to the terminal. 

### Implementation

Move from async mounted, into mounted. Since we're loading CSS we can load it whenever(TM), no need to block the rest of the mount logic. 

### Progress

@falkoschindler The problem is indeed solved, but is going from async component back to sync component a breaking change?

### DIFF
--- a/nicegui/elements/xterm/xterm.js
+++ b/nicegui/elements/xterm/xterm.js
@@ -17,10 +17,8 @@ export default {
         this.terminal[key]((e) => this.$emit(key.slice(2).toLowerCase(), e));
       });
 
-    this.$nextTick().then(() => {
-      // NOTE: wait for window.path_prefix to be set
-      loadResource(window.path_prefix + `${this.resource_path}/xterm.css`);
-    });
+    // NOTE: wait for window.path_prefix to be set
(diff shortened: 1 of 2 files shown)


evnchn:
> But I'm not sure. It might cause FOUC (flash of unstyled content).

There would be a White Flash anyways, so I call it fine.

PR #5351, opened 2025-10-25
Load Vue as ES module and example of building Vue SFC with Vite

frankier:
### Motivation

 * This PR allows syntax like `import { ref } from 'vue'` *as well as* the currently supported `const { ref } = Vue` in Vue components
   * Notably the newly supported syntax is the one used in the official Vue documentation
 * The newly supported syntax matches that which is output when Vue SFCs are built with Vite
   * An example of how to build SFCs with Vite is provided in this PR
 * This PR additionally complements https://github.com/zauberzeug/nicegui/pull/5257
   * https://github.com/zauberzeug/nicegui/pull/4673#issue-3025392294 notes that "to access various vue3 functions / utilities which will get a reference to the global Vue object - one could also choose to use [vue via ESM](https://vuejs.org/guide/quick-start#using-the-es-module-build), but this would I think duplicate code for not really any reason" --- this PR does not duplicate Vue, but does allow it to be used as an ESM

The changes were previously discussed at https://github.com/zauberzeug/nicegui/discussions/5328

### Implementation

 * The Vue build is switched from `vue.global.js` to `vue.esm-browser.js`.
 * A new dependency type `BuiltInLibrary` has been added with a production/non-production path
   * This dependency type is only used for Vue currently, and is unconditionally added in `client.py` (is this the best place to put this?)
   * It doesn't use timestamping or anything, but I think it's not necessary here.
 * It's then added to the import map under the name `'vue'` (matching the Vue documentation) inside `generate_resources`
 * To support the old-style usage with the global `Vue` object, it's the module is imported in a script tag and set as a global.

### Progress

- [x] An example has been added.

### DIFF
--- a/examples/vue_vite/README.md
+++ b/examples/vue_vite/README.md
@@ -11,25 +11,38 @@ There are two implementations of a counter component.
 `CounterOptions` is implemented using the options API, and `CounterComposition` uses the composition API.
 Both use TypeScript and Vue's Single-File Component (SFC) syntax.
 
-## Usage
+## Quick Start
 
-Run the example with `python main.py`.
+The example includes pre-built components, so you can run it immediately:
 
-To enable building/rebuilding the components, install the npm dependencies with:
+```bash
+python main.py
+```
+
+## Development Setup
+
+To modify and rebuild the Vue components, install the NPM dependencies:
 
 ```bash
 npm install
 ```
 
-In order to have modifications to the Vue components be reflected live, you can run Vite's build-watch process in parallel with the NiceGUI server:
+Then run Vite's watch process in one terminal:
 
 ```bash
-$ npm run watch &
-$ python main.py
+npm run watch
 ```
 
-Alternatively you can rebuild components manually with:
+And the NiceGUI server in another terminal:
 
 ```bash
-$ npm run build
+python main.py
+```
+
+This will automatically rebuild your Vue components whenever you save changes.
+
+Alternatively, you can rebuild components manually after making changes:
+
+```bash
+npm run build
 ```


evnchn:
I don't think we need the whole `register_built_in_library` infrastructure to register the Vue. Using the old code, we can just slip 2 entries into the importmap, and Vue can be imported already. 

And, we can avoid having to register the Vue repeatedly as well.

evnchn:
> I don't think we need ...

Yeah, tried it out. Look at the diff at https://github.com/zauberzeug/nicegui/compare/main...evnchn:nicegui:refs/heads/simple-vue-esm it's much cleaner. Add on your example and we should be good to go!

evnchn:
@falkoschindler Hear me out... If we have this PR (vue available in ES module level), and then we build all of our components (so no more template, only render functions, just like the components which @frankier built in `examples/vue_vite/vue_vite/components/CounterComposition.js`), then we can straight-up get CSP working then!

We should pull together for an epic. 💪

frankier:
> > I don't think we need ...
> 
> Yeah, tried it out. Look at the diff at [main...evnchn:nicegui:refs/heads/simple-vue-esm](https://github.com/zauberzeug/nicegui/compare/main...evnchn:nicegui:refs/heads/simple-vue-esm) it's much cleaner. Add on your example and we should be good to go!

Yeah as I was writing it I was wondering whether the `BuiltInDependendency` was too much -- I just wasn't sure whether hard coding stuff into generate_resources was the correct "layer". If it seem okay to you then yes having less code looks good.

What's the next step? Are you making a PR with this or should I rebase this one?

> @falkoschindler Hear me out... If we have this PR (vue available in ES module level), and then we build all of our components (so no more template, only render functions, just like the components which @frankier built in `examples/vue_vite/vue_vite/components/CounterComposition.js`), then we can straight-up get CSP working then!
> 
> We should pull together for an epic. 💪

Interesting -- so my understanding from this is that VBuild currently produces code which uses eval? If so, this is indeed another advantage of using Vite. In this PR I have deliberately been a bit diplomatic in simply trying to offer Vite as an alternative -- especially in light of the work done to improve VBuild. Also it's not like adding a build step comes with no drawbacks. It either has to be done manually or else NiceGUI will have to run/manage the Vite process -- and this adds a runtime dependency on node.

These are not necessarily straightforward decisions. If possible, I would like to first get this PR through on its own merits, but I'd be interested in discussing next steps after.

evnchn:
> What's the next step? Are you making a PR with this or should I rebase this one?

Go ahead and copy my code. I don't mind. 

> VBuild currently produces code which uses eval?

Not just. When Vue parses a `template` it uses eval. So whether `.vue` or `.js`, both affected. 

> adds a runtime dependency on node

Hmm I don't think so. The future looks bright. Consider: We can **pre-build** (keyword here) all Vue components, serve the runtime-only Vue, and make NiceGUI CSP-friendly, while the user can flick a switch, turn off CSP and switch to the old full Vue. There is no runtime dependency, and there is no regression.

frankier:
As you wish -- I've cherry-picked your changes together with my example.

Yes -- combining prebuilding and in-browser builds is another option to consider in future PRs.

Is this ready to merge yet?

falkoschindler:
Indeed! The changes look surprisingly minimal. 😊
Let's schedule this for 3.3. But before merging, I'd like to give it a thorough review (trying to understand the example and all implications myself), which probably won't happen before next week.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5351/commits/4706cb3fd32d97a9fdffa432daeedc7478f3590f was intentional: 

With #5257 merged, I want to see if the ESM Vue would work with `.vue` components (namely, `ui.plotly` and `ui.joystick`)

evnchn:
@frankier Can you enable workflow runs at https://github.com/frankier/nicegui/actions? Thanks

falkoschindler (changes requested):
Thanks for the pull request, @frankier!

After reviewing the code I just have a few remarks:

- The readme mentioned the command `npm watch` twice. I changed it to `npm run watch` because there is no `npm watch`. But the second instance is probably still wrong. Can you check?
- Can you add a section on how to build the components? I guess `npm install` and `npm build` is sufficient.
- What do you think about the following AI-generated assessment about potential breaking changes to user code? Is there a potential race condition? And should we set `window.Vue` or something for backward compatibility?

---

### Potential Issues:

**Race Condition Risk** ⚠️

Looking at the HTML template:

```54:63:nicegui/templates/index.html
<script type="module">
  import * as Vue from "vue";
  globalThis.Vue = Vue;
</script>
{% if prod_js %}
<script defer src="{{ prefix | safe }}/_nicegui/{{version}}/static/quasar.umd.prod.js"></script>
{% else %}
<script defer src="{{ prefix | safe }}/_nicegui/{{version}}/static/quasar.umd.js"></script>
{% endif %}
<script defer src="{{ prefix | safe }}/_nicegui/{{version}}/static/nicegui.js"></script>
```

- The `<script type="module">` without `defer` starts loading/executing immediately but asynchronously
- The `defer` scripts execute in order after HTML parsing, but there's no guarantee `globalThis.Vue` is set yet
- Module scripts execute after `DOMContentLoaded` by default

### User Code Impact (Probably Low):

1. **Custom Vue components** - ✅ Safe - They use ES modules with `export default`, loaded via dynamic `import()`
2. **User JavaScript via `ui.add_head_html()` or inline scripts** - ⚠️ Potentially affected if they access `window.Vue` or `globalThis.Vue`
3. **No examples in the repo use global Vue** - Good sign for users

### Recommendation:

Consider adding `defer` to the Vue import script OR using `async` with proper dependency management to ensure `globalThis.Vue` is available before Quasar and nicegui.js try to access it. The safest approach would be to load Vue as a regular script tag that blocks, or ensure the module completes before the deferred scripts run.

evnchn:
@falkoschindler To refresh your memory: 

We gave up on supporting people who need `Vue` in non-defer JS long time ago. Specifically in https://github.com/zauberzeug/nicegui/pull/4798#issuecomment-2911832833. 

We agreed that "we should be held back by that very niche use case"

This was so far back, I was still in Denmark doing exchange / holiday. 

So yeah I think this is a non-issue.

evnchn:
> And should we set `window.Vue` or something for backward compatibility?

Experimentally `globalThis.Vue = Vue` sets also `window.Vue` and even `Vue` so it should be fine?

falkoschindler:
@evnchn Ok, I just wanted to be sure we don't overlook anything important. And I'm not very proficient with advanced JS, so I better ask someone more familiar with it. But you're right, I vaguely remember a discussion about this subject... Good old times...
Then we only need to fix/extend the readme. 💪🏻

frankier:
Okay great that we're close now. By the way, that part of the LLM review about defer is a bit muddled. From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#other_differences_between_modules_and_classic_scripts `type="module"` is *always* defer. (It does sounds convincing though.)

The two issues with the README were related. I meant to copy-paste/modify watch to build, but fumbled it. I've now expanded the instructions a bit, so hopefully they're completely clear and self-contained now.

evnchn:
> LLM review

I'm guessing Claude, and Claude hallucinates _bad_. @falkoschindler you recall which model and the initial prompt? Good opportunity for LLM cross-comparison.

PR #5355, opened 2025-10-26
Fix of recursion in case of table rows modification

steweg:
### Motivation

This patch fixes invalid recursion, which is being called, when original table rows are being used within update_rows call. That results in infinite recursion as table rows are Observable, which triggers change event to itself

### Implementation

Simple check to avoid any observable to push itself in list of change handlers. Also specific test is added to showcase the issues

### Progress

### DIFF
--- a/tests/test_observables.py
+++ b/tests/test_observables.py
@@ -173,8 +173,8 @@ async def test_no_infinite_recursion(user: User):
     @ui.page('/')
     def page():
         list_ = ObservableList([1, 2, 3])
-        list_[:] = list_
+        list_ += list_
         ui.label(str(list_))
 
     await user.open('/')
-    await user.should_see('1, 2, 3, 1, 2, 3')
+    await user.should_see('[1, 2, 3, 1, 2, 3]')


tests/test_table.py
```diff
@@ -189,8 +189,14 @@ def replace_rows_with_carol():
         def replace_rows_with_daniel():
             t.update_rows([{'id': 4, 'name': 'Daniel', 'age': 33}])
 
+        def update_rows():
+            rows = t.rows
+            rows[0]['age'] = 35
+            t.update_rows(rows)
```
evnchn: VSCode complains: Redefines global variable `rows` due to `def rows():` above

evnchn (changes requested):
A bit of code issue, but yes, right now the code grinds to a halt with `RecursionError`. @falkoschindler I think this is major or above.

tests/test_table.py
steweg: Fixed

steweg:
After latest re-push, some unit tests failed. Apparently this is not related to this patch, but rather timeout on `screen.open()` is not big enough to cover the testing scenario. So do I have to do something from my end for this specific PR to get it through? Thanks

evnchn:
> some unit tests failed

Just rerun. It's been flaky for a while. We can address in separate PR.

steweg:
Ok, all the jobs were successful after re-run

falkoschindler:
Minimum reproduction:
```py
table = ui.table(rows=[])
ui.button('Update', on_click=lambda: table.update_rows(table.rows))
```
```
RecursionError: maximum recursion depth exceeded while calling a Python object
```

falkoschindler:
Ah, the bug isn't exclusive to `ui.table`, but can occur with any observable collection:
```py
list_ = observables.ObservableList([])
list_[:] = list_  # RecursionError
```
This PR already fixed the problem within `ObservableCollection`. 👍🏻

evnchn:
> due to unexpected logs containing a RecursionError

No, didn't see any `RecursionError`... 

But why `user.should_see('1, 2, 3, 1, 2, 3')` though. Shouldn't it be `[1, 2, 3]`?

Edit: `await user.should_see('1, 2, 3')`: Tests pass

image

falkoschindler:
@evnchn Sorry, I tried so many things back and forth and finally ended up with an inconsistent assertion. What I wanted to test was
```py
list_ += list_
```
which should match
```py
await user.should_see('[1, 2, 3, 1, 2, 3]')
```
(with or without brackets doesn't matter because substrings will also match).
But this test is still failing (at least locally).

evnchn:
> test is still failing (at least locally)

But the test passes in the pipeline and on my machine as well...

image

https://github.com/steweg/nicegui/actions/runs/18842134547/job/53756778246

Clear your pycache, maybe?

falkoschindler:
Thanks, @evnchn!
Somehow, my local installation was broken. After `pip install -e .` tests are green again. I should definitely move towards uv...

PR #5357, opened 2025-10-27
Document NiceGUI-specific coding patterns and architectural conventions

rodja:
### Motivation

This PR documents NiceGUI-specific coding patterns and architectural conventions that were previously undocumented, making it easier for contributors to understand and follow the codebase's unique design decisions.

NiceGUI has developed specific patterns to handle real-time UI synchronization, memory management, and fluent API design. These patterns (weakref usage, triple underscore storage, `KWONLY_SLOTS`, etc.) may seem unusual to new contributors but serve important architectural purposes. Documenting them reduces friction and helps contributors write code that fits naturally into the existing codebase.

### Implementation

No code changes; documentation only.

Added comprehensive "NiceGUI-Specific Patterns" section to `CONTRIBUTING.md` documenting:
- Element architecture (mixins, props vs attributes, context managers)
- Memory management patterns (weakref, WeakValueDictionary)
- Binding system internals (BindableProperty, triple underscore storage)
- Async patterns and background task handling
- Fluent interface conventions (Self return types, method chaining)
- Error handling approaches and dataclass patterns

Also consolidated coding standards from `AGENTS.md` into `CONTRIBUTING.md` as single source of truth, and restructured text for better clarity.

### Progress

  - [x] add things @evnchn identifies in his review
  - [x] add things @falkoschindler identifies in his review

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -246,8 +246,9 @@ Understanding these patterns will help you write code that fits naturally into t
 ### Memory Management
 
 - **Weakref Pattern**: Use `self._element = weakref.ref(element)` to avoid circular references
-  (which require more costly garbage collection); always check if `element()` returns `None` before using
-- **WeakValueDictionary**: Use for caches that shouldn't prevent garbage collection
+  (which require more costly garbage collection).
+  When dereferencing, always check for `None`: `element = self._element(); if element is not None: element.update()`
+- **WeakValueDictionary**: Use for caches that shouldn't prevent garbage collection by means of the reference counting mechanism of CPython
 
 ### Binding System
 


rodja:
@falkoschindler and @evnchn: I tried to add all the implicit conventions to the `CONTRIBUTING.md` I could find. There are possibly more. It would be great if you review the changes and add (or clarify) other things worth mentioning.

evnchn:
@rodja Appreciate the initiative. It's quite complete. 

I won't have anything to add, since I think in a sproadic way. (i.e.:`list(memory)` is not implemented and never will 🤡)

I'm more thinking of the flow and feeling of contribution in general. 

Note the following observations:

- Currently, the reviewing load of PRs comes from the **non-awareness of the coding patterns** especially for new contributors, since those patterns **exist in word-of-mouth** and transmitted from Falko & Rodja downwards.  
- Still, we should welcome some rapid PRs to break new ground, provided it is not too extreme (for reference: I'm at or exceeding the upper bound at times), and I fear the current modifications would **add to developmental load** and **put off new contributors.**

Central point to keep ourselves on the same page: "At NiceGUI, coding is not just for its function, but sort of an art."

- I don't think we can balance the both sides, if in the PR template, we have a single task "The implementation is complete" 
- Perhaps we can break it down into "The implementation is factualy correct" and "The implementation complies with coding patterns and conventions (insert hyperlink here)". 
- We would, if the PR is compelling enough (think: major, or sometimes medium), to take a PR even if the latter is missing, but we still _gently_ nudge things forwards by _prioritizing_ PRs which already has the latter. 
- The task will be self-proclaimed for now, but manual reviews will take place. 
- Honestly speaking, if you frequently tick yes but it's actually a no, we'll get to remember your username (I have a list in my mind but again I can't list it)

evnchn:
Second point: I would do

WeakValueDictionary: Use for caches that shouldn't prevent garbage collection **by means of the reference counting mechanism of CPython**

Reference counting is a garbage collector but one which is fast but doesn't work all the time. 

And reference counting doesn't exist for PyPy so yeah.

PR #5362, opened 2025-10-27
Add "convertEol" parameter to `ui.xterm` subprocess demo

falkoschindler:
### Motivation

In #5353 we noticed that the demo about `ui.xterm` with subprocesses should enable `convertEol` to avoid missing carriage returns.

### Implementation

This PR adds the parameter (even though it doesn't make a difference for the single-line output in this demo) and describes it briefly.

### Progress

### DIFF
--- a/website/documentation/content/xterm_documentation.py
+++ b/website/documentation/content/xterm_documentation.py
@@ -66,10 +66,13 @@ def connecting_to_subprocess():
     async def run_subprocess():
         button.disable()
         process = await asyncio.create_subprocess_exec(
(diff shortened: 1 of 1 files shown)


falkoschindler:
> Do we have more useful options for xterm worth mentioning?

I'm pretty new to xterm so I don't know its options. Therefore I'd add new insights step by step.

> Should we make our demo output more than 1 line

Oh yes. The code is a bit cryptic anyway. Do you have an idea for a similarly short Python script with multi-line output?

PR #5365, opened 2025-10-28
Introduce `Screen.CATCH_JS_ERRORS`

falkoschindler:
### Motivation

With PR #5225 being merged, users don't have an easy way to ignore JavaScript errors like before.

### Implementation

This PR adds a static `Screen.CATCH_JS_ERRORS` flag which can be set to `False` if needed.
Apart from that a new "Configuration" section also describes the three other static attributes.

### Progress

### DIFF
--- a/nicegui/testing/screen.py
+++ b/nicegui/testing/screen.py
@@ -29,6 +29,7 @@ class Screen:
     PORT = 3392
     IMPLICIT_WAIT = 4
     SCREENSHOT_DIR = Path('screenshots')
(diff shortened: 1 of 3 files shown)


evnchn:
That works. 

So for the default value being true:

- Original behaviour considered as a bug, since Pytest should capture all errors
- Even if you consider it a breaking change, I think SemVer should come after "writing good code", and code which errors out in the JS side most definitely isn't.

PR #5369, opened 2025-10-28
Implement responsive behavior for Plotly charts

AlePiccin:
### Motivation

Plotly charts rendered by `ui.plotly` did not automatically adapt to container size changes.

Typical cases where this caused problems:
- Resizing the browser window
- Showing/hiding side panels or changing flex/grid layout
- Rendering the same chart in containers of different sizes (dashboards, split views)

Without additional work, the Plotly figure would keep the initial dimensions and could overflow or get cropped instead of resizing.

The goal of this PR is to make Plotly charts responsive out of the box, without users having to register resize hooks themselves.

### Implementation

This PR adds automatic responsiveness to the built-in `Plotly` element by observing size changes of the chart’s DOM node and calling Plotly’s own resize routine.

1. A new private method `_enable_responsiveness` was added to the `Plotly` class:

```python
async def _enable_responsiveness(self):
    await ui.run_javascript(
        f"""
            (function () {{
                const plotRef = getElement({self.id});
                if (!plotRef) {{
                    return;  // DOM not ready yet
                }}

                const plotEl = plotRef.$el ?? plotRef;

                const ro = new ResizeObserver(() => {{
                    try {{
                        Plotly.Plots.resize(plotEl);
                    }} catch (err) {{
                        // Plotly may still be initializing; safe to ignore
                    }}
                }});

                // observe size changes of the Plotly root element itself
                ro.observe(plotEl);
            }})();
        """
    )
 ```
    
How it works:

We grab the DOM element for this Plotly instance using getElement(self.id).

We create a native ResizeObserver in the browser.

On every size change, we call Plotly.Plots.resize(plotEl), which is the officially supported way to have Plotly recompute its layout for the new size.

This all happens client-side; nothing is sent back to Python on resize.

The Plotly constructor now schedules _enable_responsiveness once after mount:

```python

class Plotly(Element, component='plotly.vue', dependencies=['lib/plotly/plotly.min.js']):

    def __init__(self, figure: Union[Dict, go.Figure]) -> None:
        """Plotly Element

        Renders a Plotly chart.
        There are two ways to pass a Plotly figure for rendering, see parameter `figure`:

        * Pass a `go.Figure` object, see https://plotly.com/python/
        * Pass a Python `dict` with keys `data`, `layout`, `config` (optional),
          see https://plotly.com/javascript/

        For best performance, prefer the declarative `dict` approach.
        """
        super().__init__()

        self.figure = figure
        self.update()
        self._classes.append('js-plotly-plot')
        self._update_method = 'update'

        # set up responsive behavior once the element is in the DOM
        ui.timer(0, self._enable_responsiveness, once=True)
 ```
Key design decisions:

We use ui.timer(0, ..., once=True) to run _enable_responsiveness right after the element is added to the DOM. This guarantees getElement(self.id) is available.

We rely on the browser’s native ResizeObserver that is broadly supported.

The resize logic stays entirely in the browser and calls Plotly.Plots.resize(...), which is what Plotly expects for responsive redraws.

From the user’s perspective, usage does not change:

```python

from nicegui import ui

fig = go.Figure(data=[go.Bar(x=['A', 'B', 'C'], y=[3, 1, 2])])
ui.plotly(fig).classes('full-width full-height')
```
Now the chart automatically adapts whenever its container grows or shrinks.

Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will add responsiveness to Plotly charts using ResizeObserver."
(This change only injects client-side resize handling; no new Python API surface or return values.)
(Behavior is automatic and backward compatible; no new user-facing arguments are required.)

### DIFF
--- a/nicegui/elements/plotly/plotly.vue
+++ b/nicegui/elements/plotly/plotly.vue
@@ -13,9 +13,8 @@ export default {
       this.resizeObserver.observe(this.$el);
     });
   },
-  beforeUnmount() {
+  unmounted() {
     this.resizeObserver?.disconnect();
-    this.resizeObserver = null;
   },
   methods: {
     update() {


evnchn:
@AlePiccin thanks for the PR

Hmm executing JavaScript immediately is a bit weird, can the logic be in https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/plotly/plotly.vue instead?

alexandre-piccin:
@evnchn Sure! I have done the modifications. Can you review it please?

nicegui/elements/plotly/plotly.vue
```diff
@@ -74,6 +100,7 @@ export default {
   data() {
     return {
       last_options: {},
+       __ro: null,
```
evnchn: We don't really do `__` in JS for marking private. https://github.com/zauberzeug/nicegui/blob/0162874380748736b9f6b51bf2b9771f3e93fb39/nicegui/elements/aggrid/aggrid.js#L11

nicegui/elements/plotly/plotly.vue
```diff
@@ -33,6 +45,20 @@ export default {
       // store last options
       this.last_options = options;
     },
+    initResizeObserver() {
+      if (this.__ro) {
+        return;
+      }
+
+      const el = this.$el;
```
evnchn: I'm guessing @falkoschindler prefers it inline, based on my past experience working with him.

nicegui/elements/plotly/plotly.vue
```diff
@@ -8,7 +8,19 @@ export default {
...
+    });
+  },
+
+  beforeUnmount() {
+    if (this.__ro) {
+      this.__ro.disconnect();
+      this.__ro = null;
+    }
   },
+
```
evnchn: accidental extra line?

evnchn:
@falkoschindler Important question coming up from this PR, but not related to this PR necessarily: 

We use MutationObserver and we did disconnect it: https://github.com/search?q=repo%3Azauberzeug%2Fnicegui+MutationObserver&type=code

But when we use ResizeObserver we don't disconnect it? https://github.com/search?q=repo%3Azauberzeug%2Fnicegui+ResizeObserver&type=code

I suspect this may constitute a memory leak. Should we work to fix the two instances?

evnchn:
Speaking of: @AlePiccin I also spotted another difference: for disconnect the observer do you think we should stick with `beforeUnmount()` (your impl) or `unmounted()` (`ui.aggrid`'s impl)? I am not sure which one is more correct...

alexandre-piccin:
@evnchn I'm not an expert in javascript or any frontend language, so I will follow any recommendations you guys give. Actually, if you could help me edit and review this code, I would be very grateful. 

Between `beforeUnmount()` and `unmounted()` I couldn't find any reason why one would be better than other. We could stick to `unmounted()` to keep the patterns.

nicegui/elements/plotly/plotly.vue
falkoschindler: Yes, and I'd rather call it `resizeObserver` instead of `ro`.

By the way, I wonder if we really need to store it in `data`. It should be possible to simply assign it to `this.resizeObserver` and use it in `beforeUnmount`.

falkoschindler:
> But when we use ResizeObserver we don't disconnect it?

@evnchn That's what I also noticed when reading this PR. 😀 Yes, we need to fix echart.js and header.js - either in this PR or a separate one.

> Between `beforeUnmount()` and `unmounted()`

GPT says:
> The key difference:
> - `beforeUnmount`: "I'm about to be destroyed, let me clean up while I still fully exist"
> - `unmounted`: "I'm gone, now clean up my global side effects"

In this case I guess either is fine. But since we're doing something with the observer which belongs to the component, `beforeUnmount` might be semantically better.

By the way, `beforeDetroy` is used at some places, which doesn't exist in Vue 3 anymore (if I'm correct). Maybe this another argument for a separate PR cleaning up the lifecycle hooks and making sure observers are disconnected.

evnchn:
Vue suggest use unmounted for clean up DOM event listeners. https://vuejs.org/api/options-lifecycle#unmounted so @AlePiccin let's make the switch. 

And indeed beforeDetroy is gone 🤡, we should get rid of it.

nicegui/elements/plotly/plotly.vue
evnchn: You're right we don't need that. See `aggrid.js`.

nicegui/elements/plotly/plotly.vue
```diff
@@ -8,6 +8,17 @@ export default {
     const { Plotly } = await import("nicegui-plotly");
     this.Plotly = Plotly;
     this.update();
+
+    this.$nextTick(() => {
+      this.initResizeObserver();
+    });
```
falkoschindler, line 14:
```suggestion
    this.$nextTick(() => this.initResizeObserver());
```

nicegui/elements/plotly/plotly.vue
```diff
@@ -8,6 +8,17 @@ export default {
...
+  },
+
+  beforeUnmount() {
+    if (this.resizeObserver) {
+      this.resizeObserver.disconnect();
+      this.resizeObserver = null;
+    }
```
falkoschindler:
```suggestion
    this.resizeObserver?.disconnect();
    this.resizeObserver = null;
```

nicegui/elements/plotly/plotly.vue
```diff
@@ -33,6 +44,16 @@ export default {
       // store last options
       this.last_options = options;
     },
+    initResizeObserver() {
+      if (this.resizeObserver) {
+        return;
+      }
+      this.resizeObserver = new ResizeObserver(() => {
+        this.Plotly.Plots.resize(this.$el);
+      });
```
falkoschindler:
```suggestion
      this.resizeObserver = new ResizeObserver(() => this.Plotly.Plots.resize(this.$el));
```

nicegui/elements/plotly/plotly.vue
evnchn: Ykw I was drafting this, did not commit, and then went for dinner. 

Sure let's do this.

nicegui/elements/plotly/plotly.vue
evnchn: Optional chaining is good. Explicit-set to null is a good practice which we may bring to other elements (according to https://poe.com/s/hIwBfSDySQ08aoXYsPx7)

evnchn:
How did you do the "apply suggestion" thing?

falkoschindler:
@evnchn I guess you mean step 7 in https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request

Screenshot 2025-11-03 at 14 16 50

falkoschindler:
Ok, I further simplified the diff by inlining the registration and removing the unnecessary check (`mounted` is never called more than once).

But I'm struggling to reproduce the original problem. The above-mentioned code already resizes with the browser window without this PR (tested with Chrome and Firefox):
```py
fig = go.Figure(data=[go.Bar(x=['A', 'B', 'C'], y=[3, 1, 2])])
ui.plotly(fig).classes('full-width full-height')
```

@AlePiccin Can you, please, check this code snippet once again? Do we really need the resize observer?

evnchn:
> struggling to reproduce the original problem

Me neither...

In retrospect we have responsive as true unless user explicitly pass false...

https://github.com/zauberzeug/nicegui/blob/1668bd3c7633c7e9f1f6df172a0b1a5f6a67f370/nicegui/elements/plotly/plotly.vue#L20-L23

I apologize for not spotting this earlier 🤡 though in my defense my focus went to the implementation (https://github.com/zauberzeug/nicegui/pull/5369/commits/1c2cf8aa9cf72fb63deef290da6ee8eb09f976f4) so I forgot to question the need. 

@AlePiccin While attempting to recreate the issue, could you also tell me are you starting with NiceGUI 3.0.0+ right away or upgrading from a old (possibly very old) release of NiceGUI? Since we're running the same code, I am suspecting niche issues like caching.

alexandre-piccin:
It listens only to window resizing. If it is inside a container that changes its size based on other reason than window resizing, the plotly chart won't update. For exemple, one of my apps implement a drawer that expands when mouse enters its area.

```python
with (
        ui.left_drawer(bordered=True)
        .props("mini")
        .props("mini-width=72")
        .classes("bg-dark")
        .on("mouseenter", lambda: left_drawer.props("mini=False"))
        .on("mouseleave", lambda: left_drawer.props("mini")) as left_drawer
    ):
    ...
```

That's one exemple. But there are other reasons to implement that. Another is when you have a plotly chart and another element in a container that have stretch as a configuration. If this other element changes its side (because of late mounting or other reasons), plotly wont stretch as is expected.

falkoschindler:
Ah, I see! So we have a new MRE 🎉
```py
ui.switch().bind_value(ui.left_drawer(), 'value')
ui.plotly({}).classes('w-full')
```

evnchn:
Two questions: 

- Is this a Plotly bug / deficiency? 
- Also with this PR I basically cannot turn off responsiveness anymore(?). Is that a bug on our side?

falkoschindler:
> Is this a Plotly bug / deficiency?

I'm not sure. The [documentation](https://plotly.com/python/configuration-options/#turning-off-responsiveness) says:
"Responsive figures automatically change their height and width when the size of the window they are displayed in changes." It doesn't say it reacts to its parent _container_.

> Also with this PR I basically cannot turn off responsiveness anymore(?). Is that a bug on our side?

Good question. Should we respect the `responsive` option before resizing?

alexandre-piccin:
Sadly, this is a plotly deficiency. 

> Should we respect the `responsive` option before resizing?

As you let the user choose the resposive configuration, I would recommend to only add this listener if responsive is True. But to me honest, I can't find a use case to responsive=False.

evnchn:
> Sadly, this is a plotly deficiency.

Anyone who got time and can work on a plain (so no NiceGUI) reproduction, consider filing a report at https://github.com/plotly/plotly.js? Anyways I take this PR as a win as we're doing more than upstream. 

Nevertheless I think we should respect responsive=False (idk why some user would do that, but we should let them make their own (bad) decisions, if we once did in the past). That can be in another PR.

evnchn:
I can't manage to figure out how to set responsive to False in userspace code. Giving up for now...

falkoschindler:
@evnchn This shouldn't resize with the browser window:
```py
ui.plotly({'config': {'responsive': False}}).classes('w-full')
```
I created PR #5397 to fix it.

alexandre-piccin:
> idk why some user would do that, but we should let them make their own (bad) decisions, if we once did in the past. 

haha that's true. 

Thanks for the help and the amazing job with the library!

evnchn:
Hey @AlePiccin I also noticed: You are working off your branch's main

Let's not do that next time: read #5375 for more details.

PR #5374, opened 2025-10-29
Store the `ResizeObserver` and cleanup in `unmounted`

evnchn:
### Motivation

Also acknowledged in https://github.com/zauberzeug/nicegui/pull/5369#issuecomment-3460665023, we `new ResizeObserver` without storing a reference of that anywhere, let alone disconnect it when unmounted. 

In some linters this is even flagged as a violation: https://eslint-react.xyz/docs/rules/web-api-no-leaked-resize-observer, so we better fix it. 

### Implementation

Store into `this.resizeObserver`, and disconnect when `unmounted()` as recommend by Vue docs https://vuejs.org/api/options-lifecycle#unmounted 

### Progress

### Additional notes

Should not merge conflict with #5373...

### DIFF
--- a/nicegui/elements/echart/echart.js
+++ b/nicegui/elements/echart/echart.js
@@ -88,7 +88,7 @@ export default {
     this.chart.dispose();
   },
   unmounted() {
-    this.resizeObserver.disconnect();
+    this.resizeObserver?.disconnect();
   },
   methods: {
     update_chart() {


nicegui/elements/header.js
```diff
@@ -2,11 +2,16 @@ export default {
...
-      new ResizeObserver(() => {
+      this.resizeObserver = new ResizeObserver(() => {
         document.documentElement.style.scrollPaddingTop = `${this.$el.offsetHeight}px`;
       }).observe(this.$el);
     }
   },
+  unmounted() {
+    if (this.resizeObserver) {
+      this.resizeObserver.disconnect();
+    }
```
falkoschindler:
```suggestion
    this.resizeObserver?.disconnect();
```
(untested)

evnchn: Tested. Tests pass on local. We'll see the pipeline.

PR #5380, opened 2025-10-30
Add possibility to perform tests without NiceGUI pytest plugins

himbeles:
### Motivation

As discussed and proposed in https://github.com/zauberzeug/nicegui/discussions/5377,
this adds a context manager `async run(ui_code) as user: ...`  within the NiceGUI testing framework.
It enables testing of UI code without configuration of the NiceGUI plugin setup proposed in the documentation. 

This can be necessary if NiceGUI is used in a setup where the pytest configuration can't be adapted.
As discussed in https://github.com/zauberzeug/nicegui/discussions/5377, this was necessary for UI tests required within the pyinstaller plugin project. 

It can also be helpful as another way to get started with testing in a new user's project -- without the need to configure pytest at all for NiceGUI, minimal API overhead in the call sites, and without hidden plugin "magic". 

An example of a test function would be: 
```python
async def test_button_click():
    def ui_code() -> None:
        ui.button('Click me', on_click=lambda: ui.notify('Hello World!'))

    async with run(ui_code) as user:
        await user.open('/')
        await user.should_see('Click me')
        user.find(ui.button).click()
        await user.should_see('Hello World!')
```

### Implementation

The implementation relies on an async context manager which executes `ui.run` and points an httpx.AsyncClient as a `User` to it, in line with the setup and de-setup performed in `nicegui.testing.user_plugin.user`. The context manager exposes the `User` for UI tests.

```python
@asynccontextmanager
async def run(ui_code):
    try:
        # simulate user and keep NiceGUI fully headless for tests
        os.environ['NICEGUI_USER_SIMULATION'] = 'true'

        # don't spawn reloader/native window; don't open browser
        ui.run(ui_code, reload=False, native=False, show=False)

        async with core.app.router.lifespan_context(core.app):
            async with httpx.AsyncClient(
                transport=httpx.ASGITransport(core.app),
                base_url='http://test'
            ) as client:
                yield User(client)
    finally:
        os.environ.pop('NICEGUI_USER_SIMULATION', None)
        ui.navigate = Navigate()
        ui.notify = notify
        ui.download = download
```

Open points in the architecture would be:
- Where could this implementation live in the code base?
- Suitable name: Candidates would be `run`, `user_run` to not risk overloading the very simple `run` later,  `user_of`, ...
- Option to create multiple users for the same UI code. Could yield multiple users in context manager if optional argument `n_users>1`, for example.

I'm open for discussion and refine the implementation if you consider this useful. 

### Progress

### DIFF
--- a/nicegui/testing/user_simulation.py
+++ b/nicegui/testing/user_simulation.py
@@ -1,9 +1,11 @@
+from __future__ import annotations
+
 import os
 import runpy
 from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
 from pathlib import Path
-from typing import Callable, Optional, Union
+from typing import Callable
 
 import httpx
 
@@ -17,47 +19,32 @@ from .user import User
 
 @asynccontextmanager
 async def user_simulation(
-    root: Optional[Callable] = None, *, main_file: Optional[Union[str, os.PathLike]] = None
+    root: Callable | None = None, *, main_file: str | os.PathLike | None = None,
 ) -> AsyncGenerator[User, None]:
     """Context manager for test user simulation.
 
-    Context manager that yields a ``User`` connected to a NiceGUI app within an isolated test context.
+    This context manager yields a ``User`` connected to a NiceGUI app within an isolated test context.
 
     :param root: root function which is passed directly to ``ui.run``; mutually exclusive with ``main_file`` argument.
     :param main_file: path to a NiceGUI main file executed via ``runpy.run_path``; mutually exclusive with ``root`` argument.
     """
-
     if main_file is not None and root is not None:
         raise ValueError('Cannot specify both `main_file` and `root` function simultaneously.')
-
-    if root is not None and not callable(root):
-        raise ValueError('`root` must be a callable or None')
-
-    main_file_path = None
-    if main_file is not None:
-        try:
-            main_file_path = Path(main_file)
-        except TypeError as e:
-            raise TypeError('main_file must be convertible to Path') from e
-        if not main_file_path.exists():
-            raise FileNotFoundError(f'Main file not found at {main_file_path}')
 
     with nicegui_reset_globals():
         os.environ['NICEGUI_USER_SIMULATION'] = 'true'
         try:
-            if main_file_path is not None:
-                runpy.run_path(str(main_file_path), run_name='__main__')
+            if main_file is not None:
+                if not Path(main_file).exists():
+                    raise FileNotFoundError(f'Main file not found at {main_file}')
+                runpy.run_path(str(main_file), run_name='__main__')
             else:
                 prepare_simulation()
-                if root is None:
-                    ui.run(storage_secret='simulated secret')
-                else:
-                    ui.run(root, storage_secret='simulated secret')
+                ui.run(root, storage_secret='simulated secret')
 
             async with core.app.router.lifespan_context(core.app):
                 async with httpx.AsyncClient(transport=httpx.ASGITransport(core.app), base_url='http://test') as client:
-                    user = User(client)
-                    yield user
+                    yield User(client)
         finally:
             os.environ.pop('NICEGUI_USER_SIMULATION', None)
             ui.navigate = Navigate()
--- a/website/documentation/content/user_documentation.py
+++ b/website/documentation/content/user_documentation.py
@@ -366,9 +366,11 @@ doc.text('Comparison with the screen fixture', '''
 
 
 doc.text('User Simulation Context', '''
-    The [`user_simulation`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/testing/user_simulation.py) context is the low-level building block behind the `user` fixture.
+    The [`user_simulation`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/testing/user_simulation.py) context
+    is the low-level building block behind the `user` fixture.
     It spins up a NiceGUI app inside the same event loop, providing deterministic test control without Selenium.
-    Unlike the `user` fixture, it does not rely on pytest-specific infrastructure and can be used with `unittest` or within plain async code.
+    Unlike the `user` fixture, it does not rely on pytest-specific infrastructure
+    and can be used with `unittest` or within plain async code.
 
     The context supports three testing approaches:
 
@@ -376,7 +378,8 @@ doc.text('User Simulation Context', '''
     - Test a specific main file by passing its path.
     - Define `ui.page` definitions inline within the context.
 
-    More usage examples can be found in [`tests/test_user_simulation_context.py`](https://github.com/zauberzeug/nicegui/blob/main/tests/test_user_simulation_context.py).
+    More usage examples can be found in
+    [`tests/test_user_simulation_context.py`](https://github.com/zauberzeug/nicegui/blob/main/tests/test_user_simulation_context.py).
 ''')
 
 
--- a/nicegui/testing/user_plugin.py
+++ b/nicegui/testing/user_plugin.py
@@ -21,9 +21,7 @@ from .user_simulation import prepare_simulation, user_simulation
 
 
 @pytest.fixture
-async def user(caplog: pytest.LogCaptureFixture,  # pylint: disable=unused-argument
-               request: pytest.FixtureRequest,
-               ) -> AsyncGenerator[User, None]:
+async def user(caplog: pytest.LogCaptureFixture, request: pytest.FixtureRequest) -> AsyncGenerator[User, None]:
     """Create a new user fixture."""
     async with user_simulation(main_file=get_path_to_main_file(request)) as user:
         yield user
--- a/tests/test_user_simulation_context.py
+++ b/tests/test_user_simulation_context.py
@@ -1,4 +1,5 @@
 import textwrap
+from pathlib import Path
 
 import pytest
 
@@ -13,6 +14,7 @@ async def test_basic_root_function():
     async with user_simulation(root) as user:
         await user.open('/')
         await user.should_see('Click me')
+
         user.find(ui.button).click()
         await user.should_see('Hello World!')
 
@@ -30,6 +32,7 @@ async def test_root_function_with_subpages():
     async with user_simulation(root) as user:
         await user.open('/')
         await user.should_see('Main')
+
         await user.open('/other')
         await user.should_see('Another')
 
@@ -47,6 +50,7 @@ async def test_with_page_definitions():
 
         await user.open('/')
         await user.should_see('Main page')
+
         await user.open('/a')
         await user.should_see('Page A')
 
@@ -70,23 +74,13 @@ async def test_with_page_definitions():
         ui.run(root)
     '''),
 ])
-async def test_main_file(tmp_path, file_content):
+async def test_main_file(tmp_path: Path, file_content: str):
     main_file = tmp_path / 'main.py'
     main_file.write_text(file_content, encoding='utf-8')
 
     async with user_simulation(main_file=main_file) as user:
         await user.open('/')
         await user.should_see('Main file content')
-
-
-async def test_providing_page_inside_context():
-    async with user_simulation() as user:
-        @ui.page('/')
-        def page():
-            ui.label('Some page content')
-
-        await user.open('/')
-        await user.should_see('Some page content')
 
 
 async def test_adding_page_to_root_function():
@@ -100,7 +94,9 @@ async def test_adding_page_to_root_function():
 
         await user.open('/')
         await user.should_see('Click me')
+
         user.find(ui.button).click()
         await user.should_see('Hello World!')
+
         await user.open('/a')
         await user.should_see('Page A')


nicegui/testing/user_run.py
```diff
@@ -0,0 +1,51 @@
+import os
+from contextlib import asynccontextmanager
+
+import httpx
+
+from nicegui import core, ui
+from nicegui.functions.download import download
+from nicegui.functions.navigate import Navigate
+from nicegui.functions.notify import notify
+from nicegui.testing.user_plugin import User
```
evnchn: We typically do `from .. import core` for NiceGUI internal code.

nicegui/testing/user_run.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@asynccontextmanager
+async def run(ui_code):
+    """A user context manager for the given `nicegui.ui` code.
+    
+    Example use in a plain pytest function without user plugin use:
```
evnchn: I'm rethinking: The function is just a plain test function now, have very little to do with pytest.

evnchn (commented):
2 things:

1. How the import is written, it doesn't match the codebase
2. I am initially lost over the purpose of this PR. Now, I think this PR is doing, really, "testing without pytest", do you agree? 

thanks

nicegui/testing/user_run.py
himbeles: Indeed. You can use run(ui_code) in a pytest test function or with any other test library, or just in regular functions. 
But it in the end, it enables you to run pytest tests without using the nicegui pytest plugins.

himbeles:
> 2 things:
> 
>     1. How the import is written, it doesn't match the codebase
> 


Fully agreed, of course. I have copied the code from my implementation for the pyinstaller hook, which was not within the nicegui codebase. 

Many other things would need to be adapted, also. Like type hints, doc string, etc.

himbeles:
>     2. I am initially lost over the purpose of this PR. Now, I think this PR is doing, really, "testing without pytest", do you agree?

The main purpose is not to "test without pytest" but to remove the need for the nicegui pytest plugins.
This could make it easier to set up your test environment initially for new users. (Users can use pytest but no special pytest config is required)
Also, in other scenarios, like mine for pyinstaller, you might not have the freedom to configure pytest.

nicegui/testing/user_run.py
himbeles: Changed to relative imports now, instead of absolute.

nicegui/testing/user_run.py
evnchn: Minor issue in retrospect. 

But still I think the main selling point for this PR is "test without pytest" because that is a superset of "pytest without plugins" and if we describe this PR by "pytest without plugins" we're underselling it.

nicegui/testing/user_run.py
```diff
@@ -0,0 +1,49 @@
...
+            async with httpx.AsyncClient(
+                transport=httpx.ASGITransport(core.app),
+                base_url='http://test'
+            ) as client:
+                yield User(client)
+    finally:
+        os.environ.pop('NICEGUI_USER_SIMULATION', None)
+        ui.navigate = Navigate()
+        ui.notify = notify
+        ui.download = download
```
evnchn: Some logic is shared with `nicegui/testing/user_plugin.py`. 

If that is the case, could we adhere to the DRY principle and better highlight the (I think quite high) similarity between the user fixture and this new function, either by:

1. User fixture can simply call this new function? 
2. User fixture and this new function call some common function?

nicegui/testing/user_run.py
```diff
@@ -0,0 +1,49 @@
...
+
+from .. import core, ui
+from ..functions.download import download
+from ..functions.navigate import Navigate
+from ..functions.notify import notify
+from .user import User
+
+
+@asynccontextmanager
+async def run(ui_code):
```
evnchn: `run` may cause confusion with `ui.run`. And also doesn't show the fact that this outputs a `User` which then unlocks all of https://nicegui.io/documentation/user

I am proposing `get_user` or `make_user` to hammer-down the fact that we're getting a `User`. Not sure if this name really works out though.

evnchn (commented):
- Function naming suggestions: `run_user` or `make_user` can help hammer down we're getting a `User`
- PR positioning: Though "pytest without plugins" was the intention, suggest keep "test without pytest" as how we communicate because that is the happy-accident and also a superset of the former. 
- Code duplication: Suggest to refactor instead of copy the code twice.

rodja:
> Function naming suggestions: `run_user` or `make_user` can help hammer down we're getting a `User`

I like `run` better because it reads so well in `async with run(ui_code) as user` and the analogy to `ui.run`. The few times it might collide with `run.io_bound`/`cpu_bound`, people provide another name via `as` when importing.

> PR positioning: Though "pytest without plugins" was the intention, suggest keep "test without pytest" as how we communicate because that is the happy-accident and also a superset of the former.

I find "test without pytest" quite irritating because you still need to run your tests with `pytest`. It's a more meaningful description is "user testing without using the fixture (which is made available via pytest plugin)". While I like this new feature and shortly thought if this approach could be the new go-to default, I now think that most users are better of using the plugin. By defining the `main_file` setup is super easy. Modifying the code so that all UI can be build from a single function feels like more hassel.

> How the import is written, it doesn't match the codebase

It might be good to do the import via `from nicegui.testing.user import run`, because a similar feature could be implemented for screen tests.

himbeles:
Thank you a lot for your renewed feedback.
I will have a more in-depth look at all your replies, the API and implementation over the weekend.

In the meantime, I have come up with a slightly modified approach with even simpler user interface. You can now 

test code snippets without defining a ui_code callback function

```python
async def test_get_user_context():
    async with get_user() as user:
        ui.label('as')
        await user.should_see('as')
```

or access multiple pages directly in the context manager

```python
async def test_get_user_context_with_page_definitios():
    async with get_user() as user:
        @ui.page('/')
        def page():
            ui.label('Main page')

        @ui.page('/a')
        def pageA():
            ui.label('Page A')

        await user.open('/')
        await user.should_see('Main page')
        await user.open('/a')
        await user.should_see('Page A')
```

or access a main file by specifying its path `path_to_mainfile`

```python
async def test_get_user_context():
    async with get_user(path_to_mainfile) as user:
        await user.should_see('as')
```

(Implementation is not yet committed to the PR branch.)

What do you think?

evnchn:
When necessary, ignore some of my previous comments to respect the majority vote. 

---

For `get_user` note the following code is basically equivalent. 

```py
async with run(some_code) as user:
    ...
```

```py
async with get_user() as user:
    some_code()
    ...
```

So I quite like `get_user`. But if I'm not mistaken:

- With the old `run` if we want to test a `@ui.page` function we need to pass an empty function because cannot have page functions inside root function (I think?)
- With the new `get_user` cannot easily test SPAs? (recall root functions are introduced specifically to ease SPA implementation)

So possibly `get_user(ui_code=None)` can facilitate both use cases?

rodja:
Yes, I like your "slightly modified approach" @himbeles where the context can also provides an enclosure for ui code. As @evnchn suggested the `get_user` could have an optional root-function parameter, and as you suggested an optional path to main file. That covers most of the use-cases and hence might even become the new default way.

I just struggle with the naming. `async with get_user() as user` does not read so fluent. What about 

- `async with run_user() as user`
- `async with user_simulation() as user`
- `async with start_simulation() as user`
- `async with run_simulation() as user`

himbeles:
I have made quite some progress on the implementation.

Picked `user_simulation` for the async context manager name, as suggested by @rodja. 
Although it is not the shortest, it has multiple advantages:
- Conveys the actual functionality (simulates a user connected to and NiceGUI code/app)
- Explicit enough such that it doesn't overload simpler function names (like run, simulate, open, ...)
- Could be extended with more arguments (e.g., an optional argument n_users, which defines the number of users to return). 
- Being a noun, it hints at it being a context manager. 

`user_simulation` now accepts either a 
- `root` parameter for script mode and component level tests, 
- a `main_path` parameter to point to a main file
- or no parameters, to have in-line UI code in the context

Parameters are documented in the doc string and new code has type annotations.
 
See the test cases for the different usage scenarios, implemented here: 
https://github.com/zauberzeug/nicegui/pull/5380/files#diff-969568fb0740c47dc251166615e00889c2416cf6c2d14804b72d795f51240737

For example 
```python
# Test script mode, root function input 
async def test_user_simulation_script_mode():
    def root():
        ui.button('Click me', on_click=lambda: ui.notify('Hello World!'))

    async with user_simulation(root) as user:
        await user.open('/')
        await user.should_see('Click me')
        user.find(ui.button).click()
        await user.should_see('Hello World!')

# Main file input, with script mode and normal mode
@pytest.mark.parametrize('file_content', [
    textwrap.dedent('''
        from nicegui import ui
        @ui.page("/")
        def page():
            ui.label("Main file content")
        ui.run()
    '''),
    textwrap.dedent('''
        from nicegui import ui
        def root():
            ui.label("Main file content")
        ui.run(root, storage_secret="simulated secret")
    '''),
])
async def test_user_simulation_script_main_file(tmp_path, file_content):
    main_path = tmp_path / 'main.py'
    main_path.write_text(file_content, encoding='utf-8')

    async with user_simulation(main_path=main_path) as user:
        await user.open('/')
        await user.should_see('Main file content')


# Inline UI
async def test_user_simulation_with_page_definitios():
     async with user_simulation() as user:
        @ui.page('/')
        def page():
            ui.label('Main page')

        @ui.page('/a')
        def pageA():
            ui.label('Page A')

        await user.open('/')
        await user.should_see('Main page')
        await user.open('/a')
        await user.should_see('Page A')
```

The source code for `user_simulation` lives in `nicegui/testing/user_simulation.py` and does not use pytest functionality and imports.
Existing pytest-specific code in `nicegui/testing/user_plugin.py` has been refactored to make use of `user_simulation`, to avoid code duplication. 

Key selling points of the new API are:
- Simulate users without dependency on pytest. Could use standard library unittest for example, or pytest without plugins.
- Easier to switch between testing root functions in script mode and main files. 
- Easy to test at component level (using script mode).
- Explicit API without pytest plugin fixture "magic". Might be easier to understand for some users.

himbeles:
One more idea would be to have only a single input to user_simulation, and a runtime type check would decide if it is a callable (root function) or path-like (main file). Might make the API easier/shorter, but perhaps more difficult to maintain.

rodja (changes requested):
> One more idea would be to have only a single input to user_simulation

For documentation clarity, it is better to have multiple different parameters.

Here my review commends (aided by claude-4.5):

- **Inverted dependency architecture**: `nicegui_reset_globals()` is general testing infrastructure used by both `user` AND `screen` fixtures, but now lives in `user_simulation.py`. This creates backwards dependency where `general_fixtures.py` depends on `user_simulation.py`. Was the change intentional? Similarly, `find_all_subclasses()` helper should also move back to `general_fixtures.py` since it's general test infrastructure, not user-simulation-specific.

- **Inconsistent error handling**: In `user_plugin.py`, the ERROR log check happens after the `user_simulation` context exits (lines 32-34). This means cleanup happens before checking for errors, which could mask issues if cleanup itself produces errors. 

- **`create_user` fixture inconsistency**: The `create_user` fixture (lines 37-48 in `user_plugin.py`) still manually manages the lifespan context and cleanup, not using `user_simulation()`. This creates duplicate logic and inconsistency with the new pattern. 

- **Missing documentation**: The new `user_simulation()` context manager should be documented in the website. It's a quite powerful feature but needs a few more words of explanation. I think we should stick to the user plugin as default way of writing tests because it is so easy and works for >90% of all projects. `user_simulation` should be part further down in `user_documentation.py`.

himbeles:
> > One more idea would be to have only a single input to user_simulation
> 
> For documentation clarity, it is better to have multiple different parameters.
> 

Agreed. 

> Here my review commends (aided by claude-4.5):
> 
> * **Inverted dependency architecture**: `nicegui_reset_globals()` is general testing infrastructure used by both `user` AND `screen` fixtures, but now lives in `user_simulation.py`. This creates backwards dependency where `general_fixtures.py` depends on `user_simulation.py`. Was the change intentional? Similarly, `find_all_subclasses()` helper should also move back to `general_fixtures.py` since it's general test infrastructure, not user-simulation-specific.

Indeed, the changes were intentional. 
I think that within the testing framework, we should further enhance the separation between two "layers":
- general logic on how to set up user simulation (user.py, screen.py, user_simulation.py), **independent on pytest**
- pytest plugin logic (user_plugin.py, screen_plugin.py, general_**fixtures**.py): these files depend (=import) on pytest. 

This separation was already there, and it was further carried through here. One main motivation of the PR was to be able to test independent of pytest (and its plugins). 

Note that `nicegui_reset_globals` was not moved to `user_simulation.py`. Instead, there is a plugin-version of it still in `general_fixtures`, based on a context-version in `user_simulation.py`, to avoid code duplication. 
I'd propose to move the context version of `nicegui_reset_globals` from user_simulation.py into a **new file** `general.py`. 
This would implement the same naming logic and code separation as for screen.py versus screen_plugin.py, user.py versus user_plugin.py. 

`find_all_subclasses` is only a local helper function used in  `nicegui_reset_globals`, and I think, is not documented. So I have moved it together with `nicegui_reset_globals`. I would propose to put it into `general.py`, too. 

> * **Inconsistent error handling**: In `user_plugin.py`, the ERROR log check happens after the `user_simulation` context exits (lines 32-34). This means cleanup happens before checking for errors, which could mask issues if cleanup itself produces errors.
> * **`create_user` fixture inconsistency**: The `create_user` fixture (lines 37-48 in `user_plugin.py`) still manually manages the lifespan context and cleanup, not using `user_simulation()`. This creates duplicate logic and inconsistency with the new pattern.

This duplication of logic was already present before the PR. So, I think we should treat it as a separate, independent issue.
(Also, I don't see a clever way of getting rid of the (2 lines of) redundancy, since `create_user` yields a different type.)
Indeed, for the new `user_simulation ` context manager, I have not implemented a variant that can yield a user generator, similar to create_user. This could be a future feature. 

> * **Missing documentation**: The new `user_simulation()` context manager should be documented in the website. It's a quite powerful feature but needs a few more words of explanation. I think we should stick to the user plugin as default way of writing tests because it is so easy and works for >90% of all projects. `user_simulation` should be part further down in `user_documentation.py`.

It is of course your choice on how to position the feature, and how prominently to document it at all. I can propose some minimal documentation in `user_documentation.py`, as you suggested.

rodja:
> [Layer 1:] general logic on how to set up user simulation

Ok. Sounds good. I just struggle with the naming. In the past, "user simulation" always means the user being simulated (eg. no browser) while screen tests is "automated browser" (eg. selenium). How could we keep that nomenclature? Introducing `general.py` sounds like a good way...

> This duplication of logic was already present before the PR. So, I think we should treat it as a separate, independent issue.

Ok.

> I can propose some minimal documentation in user_documentation.py, as you suggested.

That would be great! Thanks.

himbeles:
> * **Inconsistent error handling**: In `user_plugin.py`, the ERROR log check happens after the `user_simulation` context exits (lines 32-34). This means cleanup happens before checking for errors, which could mask issues if cleanup itself produces errors.

This was addressed by latest commit:
Moved the ERROR log inspection inside the user_simulation context so it runs before cleanup, ensuring unexpected cleanup errors can’t hide test failures.

himbeles:
> > [Layer 1:] general logic on how to set up user simulation
> 
> Ok. Sounds good. I just struggle with the naming. In the past, "user simulation" always means the user being simulated (eg. no browser) while screen tests is "automated browser" (eg. selenium). How could we keep that nomenclature? Introducing `general.py` sounds like a good way...
> 
> > This duplication of logic was already present before the PR. So, I think we should treat it as a separate, independent issue.
> 
> Ok.
> 
> > I can propose some minimal documentation in user_documentation.py, as you suggested.
> 
> That would be great! Thanks.

Will do so. General test logic was moved into `general.py` in latest commit.

himbeles:
Added documentation. @rodja, if you could have a look.

nicegui/testing/user_simulation.py
```diff
@@ -0,0 +1,64 @@
...
+from ..functions.download import download
+from ..functions.navigate import Navigate
+from ..functions.notify import notify
+from .general import nicegui_reset_globals, prepare_simulation
+from .user import User
+
+
+@asynccontextmanager
+async def user_simulation(
+    root: Optional[Callable] = None, *, main_file: Optional[Union[str, bytes, os.PathLike]] = None
```
rodja: Do we need `bytes`? I get typing warnings below ...

website/documentation/content/user_documentation.py
```diff
@@ -364,5 +364,71 @@ async def page():
...
+doc.text('User Simulation Context', '''
+    The [`user_simulation`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/testing/user_simulation.py) context is the low-level building block behind the `user` fixture.
+    It spins up a NiceGUI app inside the same event loop, giving tests deterministic control without Selenium.
+    More usage examples can be found in [`tests/test_user_simulation_context.py`](https://github.com/zauberzeug/nicegui/blob/main/tests/test_user_simulation_context.py).
+
+    It can be used directly to
+
+    - simulate a `User` client without relying on pytest-specific fixtures.
+    That means, it can be used with `unittest` or within plain async code.
+    - test script-mode `root` functions passed to context as callables.
```
rodja: not script mode

rodja:
I made these modifications which I hope are ok with you:

- Fixed import statements in `general_fixtures.py` (corrected import paths)
- Renamed parameter from `script_path` to `main_file` across user simulation components for better clarity
- removed "script mode" from documentation because the "root function" in user simulation has nothing to do with script mode
- Improved test names and structure for better readability (and not mentioning "script mode" anymore)
- Made user simulation available in `nicegui.testing` module

Now I just wonder if we really need the `bytes` typing for `main_file` in `user_simulation`. Currently I get a typing error because `Path` does not expect `bytes` as parameter...

himbeles:
Thank you for this early-morning review :) 
All your changes are much appreciated. 

I had added bytes in accordance with the suggestion in PEP 519. 
https://peps.python.org/pep-0519/#provide-specific-type-hinting-support
But, indeed, `Path` typing rules only accepts `str` or `Path` as input. 
I think it is fine to remove `bytes`, and I just did so. 
Passing `bytes` would be a rare scenario, and only affects the type checks anyway. 

Concerning the renaming from `main_path` to `main_file`, this is fine with me. I was preserving the existing naming (`main_path`) in the `user` pytest fixture. Perhaps you want to double check this point.

himbeles:
I would be happy with the PR as is and you could merge it at any point.

rodja:
I'm also happy with the PR. @falkoschindler please have a quick look at the code. I think it's fine but you always find another thing ;-)

himbeles:
@falkoschindler Thanks for reviewing and merging. One minor remark: in your last commit, you have introduced union typing with "|". I had explicitly avoided that because NiceGUI specifies a minimum Python version of 3.9, and only 3.10 has introduced union typing with "|".

evnchn:
@himbeles No harm done when union typing is already _everywhere in the codebase_? 

image

I am not sure why also. My venv is Py3.9 and things works flawlessly. Does it work for you? 

https://stackoverflow.com/q/76579085 may or may not be a good explanation...

falkoschindler:
@himbeles Good catch! But I used `|` intentionally. Even though it isn't supported by Python 3.9, you can opt-in via `from __future__ import annotations`. Then Python 3.9 - magically - understands these advanced type annotations. Many parts of the codebase already use this new syntax, but as far we haven't replaced it everywhere to avoid a huge diff. But in new files I tend to use the import so that the code is already "3.10-ready" and we'll only need to remove the import when dropping support for 3.9.

himbeles:
Perfect. Just wanted to be sure that this is not missed.

Issue #4882, opened 2025-06-24
WebView2 init issue


phifuh:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import app, ui
import src.ui.window

app.native.window_args["resizable"] = False
app.native.start_args["debug"] = True
app.native.settings["ALLOW_DOWNLOADS"] = True


# Run the NiceGUI application
ui.run(
    native=False,
    title="Test Data Generator",
    window_size=(1280, 820),
    language="da",
    fullscreen=False,
    port=8080,
)
```

This code generates this error:

```log
NiceGUI ready to go on http://127.0.0.1:8080
[pywebview] Using WinForms / Chromium
[pywebview] WebView2 initialization failed with exception:
  (0x80080005): Server-udførelse mislykkedes (Undtagelse fra HRESULT: 0x80080005 (CO_E_SERVER_EXEC_FAILURE))
   ved System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   ved Microsoft.Web.WebView2.Core.CoreWebView2Environment.<CreateCoreWebView2ControllerAsync>d__10.MoveNext()
--- Afslutningen på staksporingen fra den tidligere placering, hvor undtagelsen blev udløst ---
   ved System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   ved System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   ved Microsoft.Web.WebView2.WinForms.WebView2.<InitCoreWebView2Async>d__25.MoveNext()
```

The intresting part is tho, I wouldnt expect this behaviour because native is set to False, 
when removing window_size and fullscreen argument, the issue is gone. So I kinda that the arguments kinda overwrite some of the naitive arument? The native flag should have the highest prio in my opinion, so that you just can swap between each mode. 

(Any help on the webview2 issue is tho also appreciated)

### Description

Not sure how to replicate it, because you kinda need to have the webview error to start with to showcase the naitive loading issue. But it should be traceable in the code

### NiceGUI Version

2.20

### Python Version

3.13

### Browser

Other, Edge

### Operating System

Windows

[...]

PR #5391, opened 2025-11-03
Try-except `tempfile.mkdtemp()` and show error if it fails

falkoschindler:
### Motivation

In #4882 we noticed that the native mode can fail because the creation of a temp directory fails.

### Implementation

This PR shows an explicit error and a suggestion how to fix the problem if `mkdtemp()` raises.

While working on this PR I noticed that the documentation says:
> Caller is responsible for deleting the directory when done with it.

So it looks like we should remove the temp dir after use.

### Progress

  - [x] Remove temp dir after use?
  - [x] Avoid error if "storage_path" is already in `start_args`.

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import _thread
+import contextlib
 import multiprocessing as mp
 import queue
 import socket
@@ -8,6 +9,7 @@ import sys
 import tempfile
 import time
 import warnings
+from collections.abc import Generator
 from threading import Event, Thread
 from typing import Any, Callable
 
@@ -50,13 +52,26 @@ def _open_window(
     _start_window_method_executor(window, method_queue, response_queue, closed)
     if not core.app.native.start_args.get('private_mode', True) and 'storage_path' not in core.app.native.start_args:
         log.warning('Pass in a `storage_path` to properly disable `private_mode` for the native app.')
+    with _provide_storage():
+        webview.start(**core.app.native.start_args)
+
+
+@contextlib.contextmanager
+def _provide_storage() -> Generator[None, None, None]:
+    if 'storage_path' in core.app.native.start_args:
+        yield
+        return
+
     try:
-        storage_path = tempfile.mkdtemp()
-    except Exception:
-        log.error('Error creating temporary storage path. '
+        temp_dir = tempfile.TemporaryDirectory()
+    except OSError as e:
+        log.error(f'Error creating temporary storage path: {e}. '
                   'Please set `app.native.start_args["storage_path"]` to a valid path.')
         sys.exit(1)
-    webview.start(**{'storage_path': storage_path, **core.app.native.start_args})
+
+    with temp_dir:
+        core.app.native.start_args['storage_path'] = temp_dir.name
+        yield
 
 
 def _start_window_method_executor(window: webview.Window,


evnchn:
Would `tempfile.TemporaryDirectory` be better since it tosses the directory for you when you're done with the context?

phifuh:
@falkoschindler  Using 3.2.0.post5.dev0+4b738c30 I still get the same Error as before. App isnt opning but I get a windows error.

Are you sure that the temp folders are needed? AI prompted us for a solution, which skips creating the tmp folders

webview_fix.py
```py
"""WebView2 Compatibility Fix.

This module patches pywebview to avoid WebView2 initialization issues
on environments where WebView2 Runtime may be restricted or corrupted.

By removing the storage_path parameter, pywebview falls back to WinForms/Chromium
which is more reliable on managed Windows environments.

Usage:
    Import this module BEFORE importing NiceGUI:

    import webview_fix  # Must be first!
    from nicegui import ui

    ui.run(native=True)
"""

import logging

import webview

logger = logging.getLogger(__name__)

# Store the original start function
_original_start = webview.start


def patched_start(*args, storage_path: str | None = None, **kwargs):
    """Patched webview.start that ignores storage_path to avoid WebView2.

    This forces pywebview to use WinForms/EdgeHTML/Chromium backend instead
    of WebView2, which may fail on restricted Citrix environments.

    Returns:
        The result of the original webview.start function without storage_path.

    """
    if storage_path is not None:
        logger.info("Citrix fix: Ignoring storage_path={storage_path} to avoid WebView2")

    return _original_start(*args, **kwargs)


# Apply the patch
webview.start = patched_start
logger.info("Applied Citrix WebView2 compatibility patch")
```

main.py
```python
import webview_fix

from nicegui import app, native, ui

ui.run(
	native=True,
	title="Webview temp fix",
	reload=False,
)
```

evnchn:
> Are you sure that the temp folders are needed?

While I'm unsure: 

1. This PR doesn't change the general way of doing things, while handling errors and deleting the tempdir courtesy of context manager. 
2. Because we `yield` on `if 'storage_path' in core.app.native.start_args:` you can disable NiceGUI's storage path provision if it's not necessary to you. 

So I would still take this PR as a win.

falkoschindler:
@phifuh As far as I remember, the temp folder is needed for `app.storage`. Because the native app isn't supposed to leave persistent data behind, a temporary file is used.

@evnchn Yes, but adding +22-1 lines of code without actually solving the original problem is a bit much. Therefore I talked to Sonnet 4.5 about it:

> You're absolutely right - you're catching the exception at the wrong place! The `OSError` catch around `tempfile.TemporaryDirectory()` will only catch failures when **creating** the temp directory itself, not when **WebView2 tries to use it**.
> 
> The actual WebView2 initialization failure (the `0x8000FFFF E_UNEXPECTED` error) happens later when `webview.start()` is called on line 53. At that point, WebView2 tries to initialize with the storage path, and that's where it can fail due to:
> 1. Path being too long (Windows MAX_PATH limitations)
> 2. Special characters or non-ASCII characters in the path
> 3. Path in a restricted location
> 4. Permissions issues
> 5. Corrupted WebView2 user data folder

So I'll try improving the PR to hopefully catch this exception.

falkoschindler:
I came to the conclusion that this PR is overcomplicating things and, meanwhile, not solving the actual problem. Therefore I'll close it. See https://github.com/zauberzeug/nicegui/issues/4882#issuecomment-3503024779 for my most recent insights.

PR #5393, opened 2025-11-03
Introduce pytests for `VBuild`

falkoschindler:
### Motivation

Improvements to `VBuild` like PR #5257 are dangerous because the implementation is currently untested.

### Implementation

This PR adds six simple test cases to verify the core functionality of the Vue parser. A helper function `check` allows to write tests with almost no overhead, allowing to focus on the SFC input and the HTML/CSS/JS output.

A few minor changes in vbuild.py (whitespace only) ensures a clean test result.

### Progress

### DIFF
--- a/tests/test_vbuild.py
+++ b/tests/test_vbuild.py
@@ -7,14 +7,14 @@ from nicegui.elements.markdown import remove_indentation
 from nicegui.vbuild import VBuild
 
 
-def check(text: str, html: str, style: str, script: str) -> None:
+def check(text: str, *, html: str, css: str, js: str) -> None:
     with tempfile.TemporaryDirectory() as tmp_dir:
         temp_file = Path(tmp_dir) / 'TEST.vue'
         temp_file.write_text(remove_indentation(text), encoding='utf-8')
         vbuild = VBuild(temp_file)
         assert remove_indentation(vbuild.html).strip() == remove_indentation(html).strip()
-        assert remove_indentation(vbuild.style).strip() == remove_indentation(style).strip()
-        assert remove_indentation(vbuild.script).strip() == remove_indentation(script).strip()
+        assert remove_indentation(vbuild.style).strip() == remove_indentation(css).strip()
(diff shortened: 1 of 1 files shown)


nicegui/vbuild.py
```diff
@@ -15,8 +15,8 @@ def __init__(self, filepath: Path) -> None:
             raise ValueError(f'{filepath} has no template')
 
         name = filepath.stem.replace('/', '-').replace('\\', '-').replace(':', '-').replace('.', '-')
-        html = re.sub(r'^<([\w-]+)', rf'<\1 data-{name}', parser.html)
-        self.html = f'<script type="text/x-template" id="tpl-{name}">{html}</script>'
+        html = re.sub(r'^<([\w-]+)', rf'<\1 data-"{name}"', parser.html)
```
evnchn: Not too sure why this change. I don't think it helps with testing (we are using `==` anyways)

Though to be honest, I didn't fully understand the line before nor after the change. I'm more like if it ain't broke don't fix it.

tests/test_vbuild.py
```diff
@@ -0,0 +1,126 @@
+import tempfile
+from pathlib import Path
+
+import pytest
+
+from nicegui.elements.markdown import remove_indentation
+from nicegui.vbuild import VBuild
+
+
+def check(text: str, html: str, style: str, script: str) -> None:
```
evnchn: I had to look at this line quite often to remind myself of the calling order. Should we make these keyword-only instead of allow positional?

nicegui/vbuild.py
falkoschindler: That was a mistake. Tests failed. I misread it as `<... data={name}>` and wanted to turn it into `<... data="{name}">`. But it doesn't make sense for `<... data-{name}>`.

tests/test_vbuild.py
falkoschindler: Yes, maybe this is too much minimalism.
Actually, we could call them `html`, `css` and `js`. This is shorter and a bit more precise.

PR #5397, opened 2025-11-04
Don't auto-resize `ui.plotly` if `responsive === false`

falkoschindler:
### Motivation

With PR #5369 `ui.plotly will auto-resize even if the "responsive" option is set to `False`:
```py
ui.plotly({'config': {'responsive': False}}).classes('w-full')
```

### Implementation

This PR conditionally early-exits the auto-resize callback if the user opted-out the "responsive" configuration.

### Progress

### DIFF
--- a/nicegui/elements/plotly/plotly.vue
+++ b/nicegui/elements/plotly/plotly.vue
@@ -9,7 +9,10 @@ export default {
     this.Plotly = Plotly;
     this.update();
     this.$nextTick(() => {
-      this.resizeObserver = new ResizeObserver(() => this.Plotly.Plots.resize(this.$el));
+      this.resizeObserver = new ResizeObserver(() => {
+        if (this.options.config?.responsive === false) return;
+        this.Plotly.Plots.resize(this.$el);
+      });
       this.resizeObserver.observe(this.$el);
     });
   },


nicegui/elements/plotly/plotly.vue
```diff
@@ -9,7 +9,10 @@ export default {
     this.Plotly = Plotly;
     this.update();
     this.$nextTick(() => {
-      this.resizeObserver = new ResizeObserver(() => this.Plotly.Plots.resize(this.$el));
+      this.resizeObserver = new ResizeObserver(() => {
```
evnchn, line 12: We're creating an observer which doesn't nothing in this implementation, but this can possibly facilitate switch on or off the responsive.

evnchn (commented):
It's fine. Extremely similar to what I have wrote but I lacked the MRE to test it. 

Minor point: I'm contemplating whether we should create the resizeObserver in the first place.

nicegui/elements/plotly/plotly.vue
falkoschindler: Yes, options can change. So we need to check the current value of `options.config.responsive` whenever we observe a resize.

evnchn: Fair. Let's have that :rocket:

Issue #4870, opened 2025-06-16
3D Plotly figure jumps


lucasimi:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
import plotly.graph_objects as go
from nicegui import ui

fig = go.Figure(data=[go.Surface(z=[[1, 2], [3, 4]], colorscale="Viridis")])
fig.update_layout(uirevision="constant")
plot = ui.plotly(fig)


def change_colormap():
    fig.update_traces(colorscale="Cividis")
    plot.update_figure(fig)


ui.button("Change colormap", on_click=change_colormap)
ui.run()
```

### Description

When changing the colormap of a 3D Plotly figure rendered inside a NiceGUI app, the figure visibly "jumps" or the camera view resets briefly, which is jarring to the user.

After the page loads, and the user rotates and zooms the image, if the button is clicked the camera makes a little jump.

### NiceGUI Version

2.18.0

### Python Version

3.11.12

### Browser

Firefox

### Operating System

Linux

falkoschindler:
Thanks for reporting this issue @lucasimi. Unfortunately, I can't reproduce any "jump" with Chrome, Firefox or Safari on macOS.
Can anyone from the community reproduce the problem?

evnchn:
I also can't recreate on Chrome and Firefox. For Chrome, I have tried to apply a 6x CPU slowdown and Slow 4G profile in DevTools to see if I can expose the issue, but nope. 

@lucasimi Can you record a video and attach here? You may need to convert to GIF, by the way.

lucasimi:
Hi, thank you for taking a look at this. To be fair, it's not clear to me if this is an issue with plotly or with nicegui. However, in this recording I try to pan and zoom. When I click the button, the zoom level is inconsistent.

https://github.com/user-attachments/assets/e99f61c0-703e-461f-97b3-f6ff1a2080b7

falkoschindler:
Strange. Even if I try to closely replicate your pan and zoom movements, I don't see any jump. 😕

lucasimi:
To add more context, I'm using `plotly` version `6.1.2`.

falkoschindler:
Even after upgrading from 6.0.0 to 6.1.2, I still can't reproduce the problem.

[...]

PR #5404, opened 2025-11-05
Avoid conflict between `responsive` and `resizeObserver`

falkoschindler:
### Motivation

In #4870 we observed a jumping button in combination with a Plotly element:
```py
ui.plotly(go.Figure())
ui.button('Change colormap', on_click=plot.update)
```
As it turns out, it is caused by Plotly's "responsive" setting which causes the plot to shrink to size 0x0.

### Implementation

This PR removes the "responsive" setting if it is `true`. In this case our `resizeObserver` from PR #5369 handles the change of the container size anyway and would conflict with this setting.

### Progress

### DIFF
--- a/nicegui/elements/plotly/plotly.vue
+++ b/nicegui/elements/plotly/plotly.vue
@@ -29,8 +29,7 @@ export default {
 
       // default responsive to true
       const options = this.options;
-      if (options.config === undefined) options.config = { responsive: true };
-      if (options.config.responsive === undefined) options.config.responsive = true;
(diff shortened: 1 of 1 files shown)


evnchn (approved):
Go ahead. Simple-win. 

Though my immediate thought was whether this would lead to subtle differences in the responsive behaviour for existing users. 

Perhaps someone was looking forward to one specific thing from Plotly's responsive that calling `resize` on `ResizeObserver` cannot deliver. 

But then we can't have that hypothetical need block us from fixing a clear and reproducible button-jumping issue.

PR #5419, opened 2025-11-07
Let pywebview create a storage path if needed

falkoschindler:
### Motivation

In #4882 we noticed problems with NiceGUI's random storage path for native apps based on `tempfile.mkdtemp()`. A bad path can cause `0x8000FFFF` WebView2 errors when calling `webview.start()`.

### Implementation

This PR removes our custom storage path and relies on webview's default. This kind of contradicts our discussion in #4805 where we hesitated to leave it up to pywebview. But experiments show that storage works as expected with privacy mode (the default) and without, even with multiple app instances.

### Progress

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -5,7 +5,6 @@ import multiprocessing as mp
 import queue
 import socket
 import sys
-import tempfile
 import time
 import warnings
 from threading import Event, Thread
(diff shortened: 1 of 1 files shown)


evnchn (approved):
@falkoschindler we have discussed elsewhere that this code works. 

However, I can't believe that you were coding up some unnecessary stuff back in the day. That just isn't happening in my envision. So I would believe _the old code is once useful_. 

With this assumption, I hypothesize that the _old code was once necessary but now no-longer due to some changes in pywebview_, which is quite reasonable. 

Under this assumption, we can proceed with this PR, but to go the extra mile we may consider **put this at the top of the release notes** and/or **hyperlink to file a new issue** if this broke even after bumping pywebview / fresh Python environment (latter is unlikely since works on your machine, but former is still possible)

PR #5421, opened 2025-11-08
Experiment with shared pages

falkoschindler:
### Motivation

Inspired by discussion #5201, I asked myself (and Sonnet 4.5) what it takes to _share_ NiceGUI pages.

### Implementation

A core idea is that we don't need to immediately evaluate the page function, but we can wait until it is visited for the first time. Instead of re-creating a new `client` on every visit, we store it and re-use it for subsequent visits. To avoid cyclic references (bad for reference counting), we use a weak reference. The resulting implementation is rather simple.

At the moment this is only an experiment.

Example usage:
```py
@ui.page('/', shared=True)
def page():
    ui.label(f'The time is {time.time()}')
    ui.input('Enter your name').on('keydown.enter', lambda e: ui.label(f'Hello, {e.sender.value}!'))
    ui.button('Notify', on_click=lambda: ui.notify('Hello, world!'))
```

### Progress


rodja:
Interesting! But it will re-introduce some of the [strangeness which we encountered with auto-index client](https://github.com/zauberzeug/nicegui/discussions/4472#discussioncomment-13969621):

1. all visitors have the `request` object of the first visitor (and therefore also the same `client.ip`)
2. client cleanup (eg. pruning) requires special attention
3. what about `ui.download`?
4. will `app.storage.tab` work?
5. how to handle returns of `await ui.run_javascript`? Will there be multiple return values?

evnchn:
By principle, one-to-one actions will not work when the Python<->Browser relationship is one-to-many. 

I foresee we need to state very clearly that when `shared=True` stuff may break and that is just an inherent limitation, and possibly for the user's sake we may need to add back the checks we dropped in #5005. 

Nevertheless I think this PR is good. Second-class support is better than no support, I guess. I'll never look back for my code, though.

weinibuliu:
Hi! After a long busy time, I finally have empty time to follow this PR.

Well, I noticed it is still a work in progress but the last change committed 2 months ago. There is any difficulty in dev? And what can I do for this?

evnchn:
Feel like it's a whole lot of testing work to find edge cases of shared pages and then either fix them or document them.

weinibuliu:
> Feel like it's a whole lot of testing work to find edge cases of shared pages and then either fix them or document them.感觉要进行大量的测试工作，才能找到共享页面的极端情况，然后修复它们或记录它们。

Hi @evnchn ! So, is there a basic, usable experimental API available now? I'd be happy to assist with testing, for example, trying it out in my project.

evnchn:
@weinibuliu The API should be functionable right now, but you miss all the guardrails for the fallacies of shared page, some of which are in https://github.com/zauberzeug/nicegui/pull/5421#issuecomment-3507528799

Feel free to checkout this PR locally and run it!

falkoschindler:
I'm closing this experiment. Working through #6279 gave us measurements that answer the question this draft was asking, and the answer is that the approach — reusing one `Client` for all visitors — doesn't survive contact with the post-3.0 architecture, where `Client` is per-visitor state through and through.

@evnchn measured the sharp edges on top of this diff (details in #6279): the first visitor to leave reaches `delete_content` and deletes the shared client for everyone (1 disconnect → 0 clients, 0 tabs, with 3 viewers still connected) — with the weakref design, the page state this feature exists to share doesn't survive the first departure. `client.tab_id` is single-valued, so the last handshake wins and the first disconnect nulls it, dropping the remaining viewers' events; teardown then raises a `KeyError` per extra viewer. Add @rodja's earlier list (shared `request`/IP, `ui.download`, `await ui.run_javascript` return values, `app.storage.tab`, pruning) and it's clear these aren't edge cases of the flag but one systematic mismatch: per-client state that would have to become per-socket.

So the honest path to shared pages is a per-socket/per-viewer layer inside `Client` plus answers to what `app.storage.tab` and the delete lifecycle mean with N viewers — a refactoring project, not a 15-line flag. Given the modest demand (#5201) and that "everyone sees the same data" is achievable today with `app.storage.general` plus bindings or `ui.refreshable`, that project doesn't earn its place on the roadmap right now, and security hardening like #6279 shouldn't be constrained by a draft that would need the big rework anyway.

If we ever pick this up again, this PR and the measurements in #6279 are the map: `tab_id`, `delete_content`, and the handshake identity are the three per-client rows that need to become per-socket first. Until then, #5201 remains the place to register interest. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Decision%20by%20the%20author%3B%20arguments%20assembled%20and%20worded%20by%20the%20model%20from%20the%20review%20discussion%20in%20PR%206279.&a=claude-code "claude-fable-5: Decision by the author; arguments assembled and worded by the model from the review discussion in PR 6279.")

Issue #5119, opened 2025-09-08
Still seeing `KeyError: 'XXX.....'` after 2.24.0 bug fix


alydersen:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python

```

### Description

In #5090 the issue about `KeyError` was reported. After applying the latest version 2.24.0 I'm still seeing these issues in the log:

```
'd7cbb5c1-e428-4050-a20f-94c4a0f8b49b'
Traceback (most recent call last):
  File ".../.venv/lib/python3.13/site-packages/nicegui/background_tasks.py", line 77, in _handle_task_result
    task.result()
    ~~~~~~~~~~~^^
  File ".../.venv/lib/python3.13/site-packages/nicegui/client.py", line 292, in delete_content
    self.delete()
    ~~~~~~~~~~~^^
  File ".../.venv/lib/python3.13/site-packages/nicegui/client.py", line 356, in delete
    del Client.instances[self.id]
        ~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'd7cbb5c1-e428-4050-a20f-94c4a0f8b49b'
```

### NiceGUI Version

2.24.0

### Python Version

3.13.3

### Browser

Other

### Operating System

macOS

evnchn:
Related to #5090

rodja:
Oh, while investigating #5115 and prerender behaviour of Chrome, I discovered that I can reproduce the `KeyError` with this code:

```py
from nicegui import ui


def root():
    ui.add_head_html('''<script type="speculationrules">
        {
            "prerender": [
                {"source":"list","urls":["/test"],"eagerness":"eager","tag":"t1"}
            ]
        }
        </script>''')
    ui.link('test', '/test')


@ui.page('/test')
async def test():
    await ui.context.client.connected()
    print('connected')


ui.run(root)
```

@alydersen are you using `await ui.context.client.connected()`?

alydersen:
> `await ui.context.client.connected()`

@rodja I'm not specifically using that, but I do have a lot of async functions and ui.page()

evnchn:
Trying to backport the MRE to 2.x since the issue is with regards to 2.24.0: 

```py
from nicegui import ui


@ui.page('/')
def main_page():
[...]

Issue #5090, opened 2025-08-29
Lots of `KeyError` after Redis connection close fix


alydersen:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python

```

### Description

After the storage cleanup was implemented in 2.23.1 ref. issue #4998, I'm seeing a lot of these errors in our logs:

`KeyError: 'c529aa89-d2f0-468e-bd86-213403ba9968'` (the key will be different every time).

The stacktrace is like this:

```
Traceback (most recent call last):
  File ".../.venv/lib/python3.13/site-packages/nicegui/background_tasks.py", line 77, in _handle_task_result
    task.result()
    ~~~~~~~~~~~^^
  File ".../.venv/lib/python3.13/site-packages/nicegui/client.py", line 288, in delete_content
    self.delete()
    ~~~~~~~~~~~^^
  File ".../.venv/lib/python3.13/site-packages/nicegui/client.py", line 352, in delete
    del Client.instances[self.id]
```

I've tried this in two different environments and with two different caching db's (Redis and Valkey).

This did not appear before upgrading

### NiceGUI Version

2.23.3

### Python Version

3.13.3

### Browser

Chrome

### Operating System

macOS

sonicno1:
```
ERROR:nicegui:'0f736f1b-aaa5-3ed4-5eb9-d62c04097b2c'
Traceback (most recent call last):
  File "nicegui/background_tasks.py", line 77, in _handle_task_result
    task.result()
  File "nicegui/client.py", line 288, in delete_content
    self.delete()
  File "nicegui/client.py", line 352, in delete
    del Client.instances[self.id]
KeyError: '0f736f1b-aaa5-3ed4-5eb9-d62c04097b2c'
```

zbxiany:
I also have this problem with the debian12 system, python3.11.9

evnchn:
Everyone, thanks for the report. All of it helps. With 3 cases, I'm tending to believe it's a bug. (exactly why all report helps)

`Client.delete()` called in 2 instances:

- `delete_content`: no changes since 7 months ago
- `prune_instances`: changed in 2.23.1 (<- **sus**)

[...]

PR #5422, opened 2025-11-08
Avoid clients being deleted too early

falkoschindler:
### Motivation

This PR finally solves the KeyError described in #5119, which has been silenced in #5090 without understanding the root cause. As it turned out, `Client.prune_instances` falsely treated reconnecting clients as stale and deleted them. After the client gave up waiting for a reconnect, it got deleted as second time.

### Implementation

This PR
1. checks if there are `client._delete_tasks`, and
2. removes the condition `if self.id in Client.instances` from #5090.

This should fix the root cause and stops hiding KeyErrors in case clients are still deleted multiple times.

We could also make `delete()` idempotent so that it can be called repeatedly. But as this bug shows, it's better to crash if an assumption doesn't hold than to hide it with too many precautions.

### Progress

- [x] `test_user_storage_is_pruned` is failing. Why?


evnchn:
I'm thinking of a test for this. 

Meanwhile I'm thinking also, how can we speed up the debugging should it come next time with improvements in the codebase and tooling? 

- `Client.prune_instances` is "meant to find clients that only requested the HTTP response and never created a websocket connection" is non-obvious from first sight and the docstring "Prune stale clients" would imply that it is responsible for pruning _all_ stale clients. 
  - We should update that. I'm proposing _"Prune stale clients which never connected via websocket. Those which did, the `_delete_tasks` ensures that they self-destruct"_
  - Possibly we can refactor the if-condition at line 405-408 to `if never_socket(client) and is_stale(client):` by defining two helper functions?
- For VSCode, is there any way to mark `@property` such as `has_socket_connection` in a different color?

evnchn:
https://github.com/zauberzeug/nicegui/pull/5422/commits/14a6954df53ff88050adfa756a2a6569f6c21114 mirrors case 1 in https://github.com/zauberzeug/nicegui/issues/5119#issuecomment-3507024397. 

Can't manage to do case 2 though, but case 1 is more representative of the original issue (`KeyError` in the logs) anyways.

evnchn:
Question: Why the dual-mode? 

- Client to be destroyed by `Client.prune_instances` before socket connection. 
- Client self-destroy after socket connection. 

Can we populate a delete task in `_delete_task` on time-0 so that client always self-destroy? Then we can get rid of `Client.prune_instances` entirely.

falkoschindler:
@evnchn Interesting idea to get rid of the pruning loop! I'm just a bit concerned about the amount of code change to get it working. Before the first handshake we have no `document_id`, so we don't have an obvious key for populating the dictionaries. But in principle it should be possible.

By the way, I wonder if we actually need document IDs now that we don't have shared clients anymore (ignoring PR #5421). So maybe there is even more room for simplification.

Anyway, let's leave this PR as it is for now, allowing to merge this critical bugfix. We can try refactoring the client lifecycle in a separate PR.

evnchn:
`document_id` is public API so axing it will come at 4.0 aka later (TM), or never, if #5421 catches on. 

Other than that, code simplifications are always welcome. 

In fact I wonder whether getting rid of the 60-second client scanning can get rid of some periodic lag spikes, which I have sometimes noticed with NiceGUI apps 🤔

falkoschindler:
> `document_id` is public API

As far as I can tell it's only exchanged between client and server, but not stored as a public member in `client`. And all dictionaries, that use it as a key, are private.

evnchn:
But we have https://github.com/zauberzeug/nicegui/blob/main/nicegui%2Fstatic%2Fnicegui.js#L327

Not underscore or anything, straight up in `window`

falkoschindler:
Ah ok, I didn't consider the JavaScript world as public. At least NiceGUI is trying to hide it from the user as much as possible. But at the same time we encourage to `run_javascript` etc., so it is debatable for sure.

evnchn:
Hmm why is `_cancel_delete_task` associated with each `document_id`? 

#5421 is going to be very unreliable as a result, since: 

1. document A connect
2. document A disconnects
3. document B connect, but does not cancel the self-destruct task stored in `self._delete_tasks` caused by (2)
4. document B client deleted in the middle of active client usage

What gives? Possibly due to auto-index client (which never dies) we have flawed-logic from the getgo? 🤔

Issue #5423, opened 2025-11-08
default_props for tab_panels are ignored


rolfn:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

ui.tab_panels.default_props('animated=false')

with ui.tabs() as tabs:
    ui.tab('Tab 1.1')
    ui.tab('Tab 1.2')
    ui.tab('Tab 1.3')    
with ui.tab_panels(tabs, value='h').classes('w-full'):
    with ui.tab_panel('Tab 1.1'):
        ui.label('Content 1.1')
    with ui.tab_panel('Tab 1.2'):
        ui.label('Content 1.2')
    with ui.tab_panel('Tab 1.3'):
        ui.label('Content 1.3')

with ui.tabs() as tabs:
    ui.tab('Tab 2.1')
    ui.tab('Tab 2.2')
    ui.tab('Tab 2.3')    
with ui.tab_panels(tabs, value='h').classes('w-full').props('animated=false'):
    with ui.tab_panel('Tab 2.1'):
        ui.label('Content 2.1')
    with ui.tab_panel('Tab 2.2'):
        ui.label('Content 2.2')
    with ui.tab_panel('Tab 2.3'):
        ui.label('Content 2.3')        
            
ui.run()
```

### Description

The first tab example shows animation when switching between content, even though `animated=false` is specified as `default_props`. As the second tab example shows, `animated=false` must be specified directly to achieve the desired effect. Why doesn't `default_props` work here?

### NiceGUI Version

3.2.0

### Python Version

3.11

### Browser

Firefox

### Operating System

Linux

python-and-novella:
In `tabs.py`, `__init__` covers the `animated` props with `True` which is from para `animated`.
```python3
self._props['animated'] = animated
```

rolfn:
This answer doesn't solve the problem. From a user's perspective, like mine, it should be possible to set properties for all elements using `default_props`. What other purpose would `default_props` serve? What other way is there to disable animation for all instances of `tab_panels`?

python-and-novella:
It is not a resolution, it is just a reason.
I don't know the orignal purpose, so let's wait maintainer.

python-and-novella:
[...]

Issue #4857, opened 2025-06-11
Default parameter values overwrite default props


falkoschindler:
### Discussed in https://github.com/zauberzeug/nicegui/discussions/4856

<div type='discussions-op-text'>

<sup>Originally posted by **SHDocter** June 10, 2025</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I have checked the [FAQs](/zauberzeug/nicegui/wiki/FAQs) for a similar question and didn't find it.
- [X] I used the GitHub search to find a similar [question](https://github.com/zauberzeug/nicegui/discussions/categories/q-a) and didn't find it.
- [X] I searched the NiceGUI documentation, with the integrated search.
- [X] I already searched online "How to X in NiceGUI" and didn't find any information.

### Example Code

```python
from nicegui import ui

ui.button.default_props('color="secondary"')

ui.button()

ui.run()
```

### Description

When I run it, the button color is still 'primary'.
How can I modify the color parameters of many buttons?

### NiceGUI Version

2.15.0

### Python Version

3.11.5

### Browser

Firefox

### Operating System

Windows

### Additional Context

_No response_</div>

falkoschindler:
Thanks for bringing this up, @SHDocter!

That's an interesting problem caused by the default value of the `color` parameter:
```py
ui.button(..., color: Optional[str] = 'primary', ...)
```
When the initializer is called without a `color`, it sees the value "primary" and uses ignoring any default props. A similar problem has been discussed in #4330 without finding a good solution.

As a hacky workaround you can manipulate the default initializer value like this:
```py
ui.button.__init__.__kwdefaults__['color'] = 'secondary'
```

Apart from this we might need to once again think about how we define defaults:

**💡 Idea 1** `color: Optional[str] = 'primary'`
This is very readable and you don't have to check the documentation to find the default "primary". But default props will be ignored.

**💡 Idea 2** `color: Optional[str] = ...`
This is more cryptic and you have to check the documentation. The initializer clearly sees if the `color` is still `...` (use default prop) or if it has been set to a different value (`None` or a color string). But type checkers will complain about `...` not being an `Optional[str]` (requiring a `# type: ignore` marker).

**💡 Idea 3** `color: Union[None, str, EllipsisType] = ...`
This is technically correct, but - again - needs a look into the documentation to find the actual default color. And it explicitly accepts `...`, which doesn't make much sense from the user's perspective, but doesn't hurt either.

[...]

PR #5424, opened 2025-11-09
`props._set_and_warn`: NiceGUI take precedence, nicely

evnchn:
### Motivation

**TL-DR: Stand weak and we're no longer nice. Stand strong but warn.** 

After reviewing #5423 and #4857 with a new perspective, I have come to realize:

1. 👍 What makes NiceGUI nice is its Python API, that's _beautiful_, _pythonic_ and _well-documented_. 
2. 👎 What makes NiceGUI nice is NOT `.props` and `.default_props`, that's _ugly_ (to be clear: not Quasar's fault). 
3. 👍 What makes NiceGUI un-nice is NOT that we override user's props. We should do it otherwise we endorse (2).
4. 👎 What makes NiceGUI un-nice is that we override user's props _without telling them_, leading to head-scratchers like #5423.  

(read the above **carefully**, it's a brain-teaser)

So, instead of thinking hard in #4857's mentality to honor user's props (let's be honest: `PRIMARY_UNLESS_OVERRIDDEN` is horrendous, so is `color: Union[None, str, EllipsisType] = ...` 🤢) **and** break existing code by changing our behaviour (so that's at least 4.0 if not later since this is one big change and need time to deprecate), we should double-down on our (arguably correct) existing behaviour and inform the user as to what is going on. Hence this PR. 

### Implementation

This PR introduces `_set_and_warn`. As seen from the private method, it is intended for library use only. 

When the library use it, should there be user-set prop (not even a conflict), messages will show, informing the user to use the Python API. 

Reason why we show despite no conflict is, since NiceGUI will override user, we want to inform user early, instead of user wonder "why primary is OK but everything else not". 

### Results

Obviously I have not applied it to every element, but it works for `ui.button` (#4857) and `ui.tab_panels` (#5203)

```py
from nicegui import ui

ui.button.default_props('color="secondary"')
ui.tab_panels.default_props('animated=false')

ui.button()
ui.tab_panels()

ui.run()
```

Console will have: 

```
Conflict in Button.props['color']: User set 'secondary', but NiceGUI set 'primary' instead. Please use the Python API.
Conflict in TabPanels.props['animated']: User set 'false', but NiceGUI set 'True' instead. Please use the Python API.
```

Which _should_ prompt the user to do the following instead:

```py
ui.button(color='secondary')
ui.tab_panels(animated=False)
```

### Progress

  - [ ] Have we covered ALL NiceGUI library code? Require @falkoschindler or @rodja's review
- [x] Documentation is not necessary (error message self-explanatory...)

### PS

We do lack a Python API for setting defaults which persist across Python-land, but that is a non-blocking item and should go to the NiceGUI wishlist. 

And you can always do: 

```py
def my_secondary_button():
    return ui.button(color='secondary')
```

### DIFF
--- a/nicegui/elements/mixins/color_elements.py
+++ b/nicegui/elements/mixins/color_elements.py
@@ -23,7 +23,7 @@ class BackgroundColorElement(Element):
     def __init__(self, *, background_color: Optional[str], **kwargs: Any) -> None:
         super().__init__(**kwargs)
         if background_color in QUASAR_COLORS:
-            self._props[self.BACKGROUND_COLOR_PROP] = background_color
+            self._props._set_and_warn(self.BACKGROUND_COLOR_PROP, background_color)
         elif background_color in TAILWIND_COLORS:
             self._classes.append(f'bg-{background_color}')
         elif background_color is not None:
--- a/nicegui/elements/tabs.py
+++ b/nicegui/elements/tabs.py
@@ -76,8 +76,8 @@ class TabPanels(ValueElement):
         super().__init__(tag='q-tab-panels', value=value, on_value_change=on_change)
         if tabs is not None:
             tabs.bind_value(self, 'value')
-        self._props['animated'] = animated
-        self._props['keep-alive'] = keep_alive
+        self._props._set_and_warn('animated', animated)
+        self._props._set_and_warn('keep-alive', keep_alive)
 
     def _value_to_model_value(self, value: Any) -> Any:
         return value.props['name'] if isinstance(value, (Tab, TabPanel)) else value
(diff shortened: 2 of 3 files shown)


evnchn:
There may be cases where NiceGUI needs to give way, but I don't think that should be the default. Instead, passing `None` should be actively done by the user if they like their `props`.

falkoschindler:
Thanks for the PR, @evnchn! This might be a reasonable way to improve the status quo.

But if possible I would prefer a solution as sketched in PR #5476. If we need to touch all prop initializations anyway, we could use a technique like the `DEFAULT_PROP` sentinel to improve the API, documentation, and behavior.

evnchn:
Enough to inspire the setter approach in #5476. That PR is better in every regard. And stance is shifting in a sense that we actually want default props to take precedence. So yeah this will be closed.

PR #5425, opened 2025-11-09
One singular `self._delete_task`: remove `_num_connections` and deprecate `prune_instances`

evnchn:
### Motivation

**TL-DR: If you want something to always get cleaned up, clean it up in one place, never switch PICs. Keep logic clean and firm by always cleaning up when the function elapses - only way to stop it is to cancel the function.** 

So a client should, ideally, self-destruct. 

But now a client is destroyed in a variety of places, and only _sometimes_ it actually destroys. 

- Before socket connect: by `Client.prune_instances`
- After socket connect: by the `_delete_tasks`, which can be multiple, but we act on the last one because we only actually delete when `self._num_connections[document_id] == 0`

This is quite some technical debt (no we are not 🤡 enough to write this on day one). Some left from 2.x when we had auto-index client. This PR seeks to eliminate it. 

#### Removals and deprecations: 

- No more `_num_connections`. 
- No more `_delete_tasks` (plural). 
- Deprecate `Client.prune_instances` (does nothing now)

#### Additions

- We do a singular `_delete_task`. Considering the above, only the last one matter anyways, so why bother. 
- `self._reset_self_delete`: For share the logic between the existing socket-disconnect self-delete, and the before-connect 60-second self-delete

### Progress

  - [x] Pray to pytests
  - [x] Possibly to warn the user if they manually call `Client.prune_instances` though.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -58,7 +58,7 @@ class Client:
     shared_body_html = ''
     '''HTML to be inserted in the <body> of every page template.'''
 
-    def __init__(self, page: page, *, request: Request | None = None, _init_self_delete: bool = True) -> None:
+    def __init__(self, page: page, *, request: Request | None = None) -> None:
         self._request = request
         self.id = str(uuid.uuid4())
         self.created = time.time()
@@ -101,8 +101,7 @@ class Client:
         with self:
             self.sub_pages_router = SubPagesRouter(request)
 
-        if _init_self_delete:  # don't pass this unless you will call client.delete() yourself
-            self._reset_self_delete(timeout=60.0)
+        self._reset_self_delete(timeout=60.0)
 
     @property
     def request(self) -> Request:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -393,3 +393,4 @@ class Client:
     @classmethod
     def prune_instances(cls, *, client_age_threshold: float = 60.0) -> None:
         """Deprecated since client should self-delete"""
+        helpers.warn_once('Client.prune_instances is deprecated and will be removed in a future version.')
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -312,8 +312,7 @@ class Client:
 
         async def delete_content() -> None:
             await asyncio.sleep(timeout or self.page.resolve_reconnect_timeout())
-            if not self._deleted:
-                self.delete()
+            self.delete()
         self._cancel_delete_task()
         self._delete_task = background_tasks.create(delete_content(), name=f'delete content {self.id}')
 
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -306,6 +306,10 @@ class Client:
         self._reset_self_delete()
 
     def _reset_self_delete(self, *, timeout: float | None = None) -> None:
+        if not (core.loop and core.loop.is_running()):
+            core.app.on_startup(lambda: self._reset_self_delete(timeout=timeout))
+            return
+
         async def delete_content() -> None:
             await asyncio.sleep(timeout or self.page.resolve_reconnect_timeout())
             if not self._deleted:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -58,7 +58,7 @@ class Client:
     shared_body_html = ''
     '''HTML to be inserted in the <body> of every page template.'''
 
-    def __init__(self, page: page, *, request: Request | None = None) -> None:
+    def __init__(self, page: page, *, request: Request | None = None, _init_self_delete: bool = True) -> None:
         self._request = request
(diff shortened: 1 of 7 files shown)


evnchn:
Wait but then would https://github.com/zauberzeug/nicegui/pull/5425/commits/8414e1d9546c07bad75e80f5a55fc539fc0687af indicate that our code can possibly break existing user code which spawns `Client` before `core.loop` is ready, causing error like:

```
...
  File "/home/runner/work/nicegui/nicegui/nicegui/client.py", line 104, in __init__
    self._reset_self_delete(timeout=60.0)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/runner/work/nicegui/nicegui/nicegui/client.py", line 313, in _reset_self_delete
    self._delete_task = background_tasks.create(delete_content(), name=f'delete content {self.id}')
                        ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/nicegui/nicegui/nicegui/background_tasks.py", line 27, in create
    assert core.loop is not None
           ^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

🤔

evnchn:
The force-pushes are due to incomplete pylint fixes.

We can agree that force-push are for "commits you don't want people to see"

evnchn:
Recap on the two reverts: 

https://github.com/zauberzeug/nicegui/pull/5425/commits/8414e1d9546c07bad75e80f5a55fc539fc0687af is a bad idea, since it causes a breaking change. 

Rather, https://github.com/zauberzeug/nicegui/pull/5425/commits/4ada9e4aaf89afe11ad0302e0998467740ebffc9 makes `_reset_self_delete` tolerant in a manner that we don't need extra config for `Client` created before the core loop is ready. Added benefit is, if such client persists to a stage where the core loop _is_ ready, then we can take care of deleting the client. 

https://github.com/zauberzeug/nicegui/pull/5425/commits/1548190c0b1d083c7a8144ac585bece89377a910 is a bad idea, since when a client is deleted by something other than the self-destruct mechanism, it will leave behind a delete task which does nothing. Ideally, a deleted client should be as dead as a doorknob. 

In https://github.com/zauberzeug/nicegui/pull/5425/commits/f212c5c59295bf139deea388ffc16518edb181ab I cancel the delete task if the client is deleted, so it really _is_ as dead as a doorknob

falkoschindler:
@evnchn Today I talked to @rodja about PR #5421 (introducing shared pages). Even though we're definitely going to mark them as experimental with a list of known limitations, they might benefit from the current `_num_connections` dictionary to handle multiple browser tabs with poor connection visiting a single shared page. Therefore I think we should at least put this PR on hold and see if PR #5421 gets released.

evnchn:
We can still simplify the logic while keeping `_num_connections` (I think). 

But still I don't think it's worth the effort to do that when #5421 is a "maybe". Let's leave it later.

evnchn:
https://github.com/evnchn/nicegui/pull/59 would also necessitate giving up shared pages.

falkoschindler:
@evnchn This PR came up as one of the few non-draft PRs that isn't on our maintainer board, which prompted me to revisit its status. After looking it over, I'd like to propose closing it. My reasoning:

1. **The motivation has weakened.** The original pitch was consolidating two teardown paths, which is legitimate tech debt. But `_num_connections` is not just cleanup overhead — it's the mechanism that lets multiple sockets share a client per `document_id`, which is exactly what shared pages (#5421) would rely on. Removing it now and potentially reintroducing equivalent logic later is churn we can avoid.

2. **No external demand.** Nobody has reported a bug, leak, or confusion around `prune_instances` or the current teardown. The gain is purely internal aesthetics.

3. **The new approach adds its own wrinkle.** Moving the self-delete into `Client.__init__` means every constructor has to reason about whether `core.loop` is running yet (hence the `on_startup` fallback in `_reset_self_delete`). The old `prune_instances` timer didn't have that concern because it was registered inside `_startup()`. So the refactor trades one kind of complexity for another rather than net-simplifying.

4. **State.** The branch has merge conflicts and no new tests; keeping it alive means ongoing rebase cost for a change we're unsure we want.

My suggestion: close this, and if shared pages (#5421) is ultimately dropped, we can revisit with a narrower scope (e.g., just deprecating `prune_instances` without touching `_num_connections`). If shared pages lands, the current structure is closer to what's needed anyway.

evnchn:
Makes sense. We can track the branch in https://github.com/evnchn/nicegui (in fact I thought I did already, maybe I missed something)

PR #5432, opened 2025-11-10
Stop using `__version__` for URLs: Use secure `version_signature` instead

evnchn:
### Motivation

The security-focused would suggest that version information be kept private. 

For example, browsers are freezing the version show in User-Agent-string. 

So, being able to list the exact NiceGUI versions used for the websites in https://github.com/zauberzeug/nicegui/discussions/4506, may not be a good idea after all. 

Impact-wise it's not that big though. It doesn't even qualify for a security advisory. 

### Implementation

NiceGUI versions don't have a lot, so hashing that is a no-no (can rainbow table / enumerate)

So, the new `version_signature` is based off the storage secret under HMAC-SHA256 scheme, which is (iirc) exactly how our client-side JWT cookies are signed. 

So long as the storage secret is private, there is no good way to brute force through all NiceGUI versions, without first brute-forcing the storage secret. 

### Progress

- [x] The implementation is complete (pray to pytest)
  - ~~Fix cyclic import. Need help!~~
  - Possibly a breaking change if users are registering their own resources without going through NiceGUI?
  - But I don't feel like we should cover that case?

### Final notes

Script kiddies will spam _all_ exploits, so if you're vulnerable you're doomed anyways. 

Security through obscurity is no good. It isn't even security.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -25,7 +25,6 @@ from .observables import ObservableDict
 from .outbox import Outbox
 from .sub_pages_router import SubPagesRouter
 from .translations import translations
-from .version import __version__
 
 if TYPE_CHECKING:
     from .page import page
@@ -161,7 +160,7 @@ class Client:
             name='index.html',
             context={
                 'request': request,
-                'version': __version__,
+                'version': helpers.version_signature(),
                 'elements': elements.translate(HTML_ESCAPE_TABLE),
                 'head_html': self.head_html,
                 'body_html': '<style>' + '\n'.join(vue_styles) + '</style>\n' + self.body_html + '\n' + '\n'.join(vue_html),
--- a/nicegui/dependencies.py
+++ b/nicegui/dependencies.py
@@ -7,9 +7,8 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Callable
 
 from .dataclasses import KWONLY_SLOTS
-from .helpers import hash_file_path
+from .helpers import hash_file_path, version_signature
 from .vbuild import VBuild
-from .version import __version__
 
 if TYPE_CHECKING:
     from .element import Element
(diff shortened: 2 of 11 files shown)


falkoschindler:
Quick feedback after talking to Rodja:

- Recomputing the version signature again and again seems wasteful. Since it doesn't change at runtime, we could probably store it as a constant in `core.py`.
- More importantly: Before merging this PR, we need to update our On Air server. It currently caches NiceGUI resources to avoid requesting them from the NiceGUI app. But if every app has its own signature, we need to somehow communicate the version during handshake.

evnchn:
Version signature function is cached, so I think it is not wasteful? Still if we want to be more explicit in the constant nature we can move away from a cached function. 

The part about On Air is fair, but I don't know how to work on it since I don't know about its inner workings...

evnchn:
Well for On Air I don't think a secretive version is possible considering CDN use. 

Definitely we won't use a separated secret: that way the CDN is incredibly ineffective. 

Using a shared secret is as good as no secret, because the hash of a particular version can simply be stored in a lookup dictionary somewhere. 

So for On Air we should rather stick with the original plaintext version.

rodja:
> So for On Air we should rather stick with the original plaintext version.

On Air is a smart CDN. If the on air connection provides the information which NiceGUI version the app is running plus it's version_signature, the On Air server can use that to serve the right version from CDN if requests are made to that signature.

evnchn:
> the On Air server can use that to serve the right version from CDN if requests are made to that signature

2 concerns:

- Hash collision
- How do you "serve the right version" that way? If you redirect then it's revealed, but caching can't be efficient if we do direct serving (cache from one app does not benefit another app?)

Anyways I think you will need to take charge on anything that has to do with On Air, including security and client<->server interplay.

evnchn:
Marked as "In progress" because somebody other than me must do the On Air part.

falkoschindler:
Thanks @evnchn, and sorry to leave this hanging for so long. After thinking it through with the On Air angle from earlier in the thread, I'm going to close this rather than keep it open waiting on the relay-side work.

The honest reason is cost vs. benefit. The On Air change Rodja sketched (announce version + signature on the connection, route in the relay) is real protocol work in a separate repo, and the security gain on this side is small enough that you already framed it that way in the PR description. The version also leaks through static-asset content and feature detection regardless, so hiding the URL segment doesn't really close the door. I'd rather not carry the lifecycle complexity (storage-secret now affects URL routing, init-order constraints) for that.

One small thing worth flagging in case you reuse the helper elsewhere: `@functools.cache` on `version_signature()` combined with the import-time `app.mount(...)` / `@app.get(...)` calls in `nicegui/nicegui.py` means the first call happens before `ui.run(storage_secret=...)`, so `Storage.secret` is `None`, the function returns plaintext `__version__`, and the cache locks that in. The HMAC branch never actually fires — the tests pass only because they compare the cached value against itself. So in addition to the On Air work, the implementation would also need either lazy route registration or a cache invalidation on `set_storage_secret`.

Genuinely appreciate the security-minded eye, and I'm happy to revisit if On Air ever grows the handshake-time version channel for other reasons. Closing for now.

Issue #4237, opened 2025-01-18
How to determine whether the current connection is secure


falkoschindler:
As discussed in #4236, `ui.clipboard` warns and returns an empty string if the clipboard API isn't available. I think it should raise instead so that the user code can decide how to handle the error case. As this is technically a breaking change, this issue is scheduled for 3.0.

evnchn:
I have thought about it for a bit. 

I don't think it should raise an exception, but it should return `None`. 

None and an empty string is distinguishable by `is None`, and I can pretty much say for 99% use cases, since people (1) don't type check and (2) this is not TypeScript, that it will be fine and work with situations such as `if clipboard_result:` statements. 

Also, there is no concern that the clipboard API, when working, returns something like `undefined` or `null`, because according to MDN Web Docs, it [returns an empty string if the clipboard is empty, does not contain text, or does not include a textual representation among the objects representing the clipboard's contents](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#:~:text=Returns%20an%20empty%20string%20if%20the%20clipboard%20is%20empty%2C%20does%20not%20contain%20text%2C%20or%20does%20not%20include%20a%20textual%20representation%20among%20the%20objects%20representing%20the%20clipboard%27s%20contents)

Compared to raising an exception, which breaks all code if no action is taken when upgrading, and empty string, which breaks only if you do something like `assert isinstance(a, str)` 

We can also push out the feature earlier. 

Also, exceptions are just generally hard to deal with in my opinion. There is no way to deal with an exception without indenting in, but you can absolutely use the ternary operator in python for basically one-line if-else.

evnchn:
But then, in the Zen of Python by Tim Peters: 

> Although practicality beats purity.
> Errors should never pass silently.
> Unless explicitly silenced.

Question: Is not having access to the Clipboard API under HTTP (non-localhost) an "error"? Is raising a warning via `log.warning` considered "silent" or not? 

And the most important question of all which we wont get the answer to (not easily): Can Tim Peters please come explain what he means?

weinibuliu:
Well, the main reason I raised this question at that time was `ui.clipboard.read()` always returns a string type. This leads to a problem that I can't confirm under what circumstances an empty string is returned.

At the time, my solution was to adjust the return type of `ui.clipboard.read()` to Optional[str], which allows me to implement the judgment by judging whether the return value is None.

[...]

Issue #2335, opened 2024-01-10
Performance issues in `binding.py`'s `remove` function causing high CPU usage


ruiheng:
### Description

I have a web app that occasionally shows 100% CPU usage and becomes unresponsive. Looking at the call stacks from `faulthandler` during these incidents, there is always a thread executing the `remove` function in `binding.py`. Although I have not been able to reliably reproduce this in testing, observing the `remove` code shows obvious performance problems.

The `remove` function mainly modifies three global variables to delete some instances from them. However, there is unnecessary copying of objects even though it is shallow copies. With large numbers of objects, this still has a performance impact. 

I wrote a simple test program to create 1000 page element objects and delete them one-by-one. The test code is:

```python
from nicegui import ui, binding
import os
import time


def mk_element():
    dat = { 'text': 'hello' }
    return ui.label().bind_text_from(dat, 'text')

@ui.page('/')
async def home_page():
    num = 10000
    ele_list = [ mk_element() for _ in range(num) ]

    print(f'remove function implementation: {binding.remove.__name__}')
    print(f'num of active_links: {len(binding.active_links)}')
    print(f'num of bindings: {len(binding.bindings)}')
    print(f'num of bindable_properties: {len(binding.bindable_properties)}')

    t0 = time.time()

    for i in range(num):
        ele_list[i].delete()
 
    t1 = time.time()
    print(f'''{num} elements deleted in {t1-t0:.3f} seconds''')


if __name__ == '__mp_main__' or __name__ == '__main__':
    ui.run(
        binding_refresh_interval=1,
        port=4000,
        show=False,
        reload=True,
        )

```

Modifying the `remove` implementation shows large differences in execution times:

- remove_original: ~ 151 seconds  
- remove_0: ~ 68 seconds
- remove_1: ~ 3.x seconds  
- remove_2: ~ 0.07 seconds

So the modifications provide large performance improvements.

Here is the code of different implementations for the test. To support `remove_2`, I also add a global variable, which now show here. See below.

```python
def remove_original(objects: Iterable[Any], type_: Type) -> None:
    """Remove all bindings that involve the given objects.

    The ``type_`` argument is as a quick pre-filter.

    :param objects: The objects to remove.
    :param type_: The type of the objects to remove.
    """
    active_links[:] = [
        (source_obj, source_name, target_obj, target_name, transform)
        for source_obj, source_name, target_obj, target_name, transform in active_links
[...]

PR #5438, opened 2025-11-12
Lazily start active link refresher task

bulletmark:
This is a simple change that merely starts the active-links refresher background periodic task when needed, i.e. when there are one or more links to be checked. When there are no active links, the task is stopped. For an application that never uses active-links, no background refresher task is ever started.

This means a well crafted nicegui application (which only uses bindable properties/dataclass etc) will not wear the overhead of an unnecessary task running every 0.1 secs.

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -52,6 +52,8 @@ def _set_attribute(obj: object | Mapping, name: str, value: Any) -> None:
 
 async def refresh_loop() -> None:
     """Refresh all bindings in an endless loop."""
+    global _active_links_added  # pylint: disable=global-statement # noqa: PLW0603
+    _active_links_added = asyncio.Event()
     await _active_links_added.wait()
     if core.app.config.binding_refresh_interval is None:
         core.app.config.binding_refresh_interval = 0.1
@@ -265,11 +267,9 @@ def reset() -> None:
 
     This function is intended for testing purposes only.
     """
-    global _active_links_added  # pylint: disable=global-statement # noqa: PLW0603
     bindings.clear()
     bindable_properties.clear()
     active_links.clear()
-    _active_links_added = asyncio.Event()
 
 
 @dataclass_transform()


evnchn:
@bulletmark Thanks for the PR. 

https://github.com/zauberzeug/nicegui/blob/9555de9d6ae28aa89c86091b6b68da3daabb7e32/nicegui/binding.py#L62-L74

If `active_links` is `[]`, which it is in a well-written application, then `_refresh_step()` boils down to:

```py
def _refresh_step() -> None:
    t = time.time()
    if time.time() - t > MAX_PROPAGATION_TIME:
        log.warning(f'binding propagation for {len(active_links)} active links took {time.time() - t:.3f} s')
```

Which is not a lot of overhead. So I doubt whether this PR is needed. 

Do you have any performance metrics to back up the need for this PR? 

Finally, I note that the original PR template is gone in this PR. Try and restore it. 

Thanks again!

nicegui/binding.py
```diff
@@ -49,7 +49,7 @@ def _set_attribute(obj: object | Mapping, name: str, value: Any) -> None:
         setattr(obj, name, value)
 
 
-async def refresh_loop() -> None:
+async def _refresh_loop() -> None:
```
evnchn: Originally public API. People may intentionally call `await binding.refresh_loop()` to expedite their bindings being refreshed. Better leave it that way.

nicegui/binding.py
```diff
@@ -74,6 +74,21 @@ def _refresh_step() -> None:
...
+    @classmethod
+    def start(cls) -> None:
+        if not cls.task:
+            cls.task = background_tasks.create(_refresh_loop(), name='refresh bindings')
+
+    @classmethod
+    def stop(cls) -> None:
+        if cls.task:
+            cls.task.cancel()
+            cls.task = None
```
evnchn: Note: these two functions: 

- First one starts the task if not started already
- Second one stops the task if it has started

Relevant later

nicegui/binding.py
```diff
@@ -172,6 +189,8 @@ def bind_from(self_obj: Any, self_name: str, other_obj: Any, other_name: str,
     _check_self_and_other_attribute(self_obj, self_name, other_obj, other_name, self_strict, other_strict)
     bindings[(id(other_obj), other_name)].append((other_obj, self_obj, self_name, backward))
     if (id(other_obj), other_name) not in bindable_properties:
+        if not active_links:
+            _ActiveLinkRefresher.start()
         active_links.append((other_obj, other_name, self_obj, self_name, backward))
```
evnchn: Since you're adding to `active_links` so you know you need to start the active link refresher. 

Combined with how `_ActiveLinkRefresher.start()` starts the task if not started already, you should be able to straight-up call `_ActiveLinkRefresher.start()` with no if-condition. Same goes for above.

nicegui/binding.py
```diff
@@ -240,6 +259,11 @@ def remove(objects: Iterable[Any]) -> None:
         for source_obj, source_name, target_obj, target_name, transform in active_links
         if id(source_obj) not in object_ids and id(target_obj) not in object_ids
     ]
+    if active_links:
+        _ActiveLinkRefresher.start()
+    else:
+        _ActiveLinkRefresher.stop()
+
```
evnchn: This is a remove operation, so there is no need to consider calling `_ActiveLinkRefresher.start()` because it would have been started already in the addition process. 

Instead, you only need to consider calling `_ActiveLinkRefresher.stop()` if `active_links` is empty.

evnchn (changes requested):
Some code logic and API compatibility issues. 

But no need to rush and fix them. Let's affirm the need for this PR first.

bulletmark:
I read the template and considered that my commit words already addressed everything asked so deleted the headers etc sorry.

### Motivation

See 2nd paragraph of opening comment.

### Implementation

See 1st paragraph of opening comment.

### Progress

- [x] Documentation change is not necessary as no user facing effect.

Regarding your question about whether this PR is needed. I feel using a 0.1 sec async task to check variables frequently has a rather strong "code smell" about it, particularly when I learned that a user can easily write an the application so that task is not even needed. I originally considered to just set my "binding_refresh_interval" to something like 99999 but, even apart from the nicegui code you quote, the Python `asyncio` scheduler is still baring the cost of managing a task + timer overhead. So I looked at the nicegui code and found it is very trivial to change/improve as done here. The overhead of creating/deleting the task only occurs when the count goes to/from 0 and 1 which is infrequent and I think is far better than wearing the cost of a permanent redundant 0.1 sec periodic task.

I don't have any metrics although I did run the original main and my branch against [a test I found](https://github.com/zauberzeug/nicegui/issues/2335) where somebody was exercising thousands of bind additions and deletions just to confirm I didn't break anything and that performance was at least no worse - which it wasn't.

I know I could call `_ActiveLinkRefresher.start()` every time but just thought the single `if` test removed the redundant python function call overhead. I agree with your last comment that I should only call `_ActiveLinkRefresher.stop()` in `remove()`.

bulletmark:
I force pushed to change `binding._refresh_loop()` to `binding.refresh_loop()` and remove the test + potential call to `_ActiveLinkRefresher.start()` in `remove()`.

I left the tests of `active_links` in the code because that simple test removes the overhead of the function call and it is more clear to a reader of the surrounding code that the function is only required when the list of links becomes/goes non-zero.

nicegui/binding.py
evnchn: @bulletmark I think that, since you call the `start` and `stop` accurately, then you can get rid of the checks `if not cls.task:` and `if cls.task` entirely.

bulletmark:
I know I could change `_ActiveLinkRefresher.start()` to remove the `if` test but I can't remove the complementary `if` test in the companion `_ActiveLinkRefresher.stop()` because it is essential there (and the almighty type checker demands it!) and thus I considered the code would look more consistent and symmetrical if left both tests there as a pair. Can you see that point, or do you  want me to remove that first `if`?

evnchn:
The point about type checker is correct. Anyways I think we can wait for more reviewers, especially for whether the performance gain is worth it. Thanks for your work!

bulletmark:
Just force pushed a minor change. After the single line I removed in `nicegui.py` there were 2 imports no longer used so I removed those also.

evnchn:
Please use merge instead of force-push to stay in sync with main, and no need to stay in sync with main unless something you need from main is absent. Thanks!

evnchn:
> raise on using active link when binding interval set as None

I don't think we should raise. That could be quite annoying to debug. I think we should warn instead. 

And we'll activate the refresh loop lazily in a one-way manner (no means of deactivating) if an active link is used. So the refreshable case would not lead to rapid start-stop. 

How does that sound?

bulletmark:
@falkoschindler my original plan for a PR was to add an option to set `binding_refresh_interval=-1` so the user could disable the refresh, much as you are proposing but then thinking about it I thought it better to just enable/disable the loop as required. However, I agree with the last proposal here by @evnchn. By default the loop is disabled but enabled and kept running forever once any active link binding is created. That makes fastidious users like myself happy, as we won't have the refresh loop running every 0.1 sec, and you happy because it avoids the possible thrashing problem you raise. I will redo this PR if you agree.

evnchn:
@bulletmark I think you can get started right away if you have time. I "won the argument" before on not raising errors unless absolutely necessary, so I think I'll win this time as well? (Check https://github.com/zauberzeug/nicegui/issues/4237#issuecomment-2816084272)

bulletmark:
@evnchn the change is so straight forward I have already done it :) Just waiting for the ok to push.

falkoschindler:
@evnchn @bulletmark You're right. Raising would be a bit harsh. 👍🏻

evnchn:
Saw the force-push back to 1 commit. Now the code is really simple and clean, I don't see any problems on my side. Let's go.

falkoschindler:
Ok, so you decided against an optional `binding_refresh_interval=None` and a warning for the user?
I think I'm fine with the current PR. It improves the performance for applications without active links. I'm just a bit worried that there is no feedback for the developer if they choose to work without active links but accidentally add one.

bulletmark:
I'm setting my `binding_refresh_interval=99999` so it won't be refreshing and should be obvious.

I am quite happy to add a check if `binding_refresh_interval == None` and then warn or raise if the user sets a binding as I think this is a good idea also (of course it requires me to add documentation change here). Please let me know if you want me to add this.

bulletmark:
Can one of you guys please look at why pytest is failing? Just looking at the first failing test for example (`test_binding_to_input`) , it fails also if  I run all the tests on my machine, but it passes if I run that single test alone. I can't see how my simple change could affect that test unless it is a test environment issue?

falkoschindler:
> I am quite happy to add a check if `binding_refresh_interval == None` and then warn or raise if the user sets a binding as I think this is a good idea also (of course it requires me to add documentation change here). Please let me know if you want me to add this.

Yes, let's do this. 👍🏻

> Can one of you guys please look at why pytest is failing?

We need to reset the state of `_ActiveLinkRefresher` in `reset()`. Otherwise the next test won't start the loop because it is assumed to be still running. ✔️

evnchn:
> (manually) reset the state of `_ActiveLinkRefresher`

While it works, this may not be the best idea, and may cause subtle bugs in the future. I think it is best if the class can always be accurate as to the current state without external intervention. 

Some commits ago we had `cls.task = background_tasks.create`. If we do that, can we then check `cls.task` to see if the task is still _actually_ running?

bulletmark:
I implemented as last discussed here then rebased off latest main and cleaned up to one commit.

If/after this change is committed to main then I will also submit a separate PR to add a test for this change. Didn't want to include it here because that makes it awkward over the transition.

falkoschindler:
> I implemented as last discussed here

Nice! 👍🏻

> then rebased off latest main and cleaned up to one commit.

Please try not to force-push to a pull request. This makes it harder to keep track of already reviewed files and can cause conflicts with local checkouts of this branch. When merging we will squash all changes into a single commit anyway, so the history looks clean again.

> If/after this change is committed to main then I will also submit a separate PR to add a test for this change. Didn't want to include it here because that makes it awkward over the transition.

What transition do you mean? What exactly do you find awkward? Usually tests are an essential part of the PR.

evnchn:
I thought we wanted to warn instead of raise? https://github.com/zauberzeug/nicegui/pull/5438#issuecomment-3544720245

nicegui/binding.py
```diff
@@ -51,10 +51,12 @@ def _set_attribute(obj: object | Mapping, name: str, value: Any) -> None:
 
 async def refresh_loop() -> None:
     """Refresh all bindings in an endless loop."""
+    binding_refresh_interval = core.app.config.binding_refresh_interval
+    assert binding_refresh_interval is not None, 'binding_refresh_interval is None'
```
evnchn:
```suggestion
    binding_refresh_interval = core.app.config.binding_refresh_interval or 0.1
```

When user do `binding_refresh_interval=None` and accidentally use active links, then `assert binding_refresh_interval is not None` fails out. 

I think we can safely fallback to 0.1 second polling if the user makes such mistake? This goes with how we want a warning but functional code.

bulletmark: @falkoschindler wanted `binding_refresh_interval` to be able to be set to `None`, not `0` to disable the refresh. So presumably he wants to allow the user to be able to set it to `0`? Also, I don't really like hard coding that default in two places?

evnchn: But as agreed it should not error out unless necessary though.

And setting it to 0 would basically kill the server, so we shouldn't empower the user to do that. Otherwise maybe `binding_refresh_interval = core.app.config.binding_refresh_interval if core.app.config.binding_refresh_interval is not None else 0.1` but that's getting unwieldy...

Hardcoding the default in 2 places is generally not good but we did it with `ui.run` and `ui.run_with` already so I guess it's fine? https://github.com/zauberzeug/nicegui/blob/main/nicegui/ui_run.py and https://github.com/zauberzeug/nicegui/blob/main/nicegui/ui_run_with.py

Anyways let's wait for him and see what he thinks.

falkoschindler: I think `None` is a more intuitive value for disabling the loop. An interval of 0 suggests a loop running at maximum speed with no time in between cycles.

The line
```py
binding_refresh_interval = core.app.config.binding_refresh_interval if core.app.config.binding_refresh_interval is not None else 0.1
```
is indeed not very handy. But I guess it is technically more correct, because at the moment nothing is stopping you from setting the interval to 0, even though it might be costly.

Slightly shorter:
```py
binding_refresh_interval = 0.1 if core.app.config.binding_refresh_interval is None else core.app.config.binding_refresh_interval
```

bulletmark: I don't like dereferrencing `core.app..` twice. If you don't really want the user to set the value to 0 or negative then this is a good place to limit check that so perhaps:
```python
    if (interval := core.app.config.binding_refresh_interval) is None or interval <= 0:
        interval = 0.1
````

If you don't want to check for `<= 0` then same test but just remove the 2nd `or` part?

**Comment by evnchn** [comment]
Diffs clean. Let's go.

**Comment by bulletmark** [comment]
It's your code but I don't like double deferencing (the triple deep dict) in the following code in the tight 0.1 sec loop:

```python
assert core.app.config.binding_refresh_interval is not None
await asyncio.sleep(core.app.config.binding_refresh_interval)
```

Best to do the dererence only once:

```python
interval = core.app.config.binding_refresh_interval 
assert interval is not None
await asyncio.sleep(interval)
```

I originally also used a simple function + global and only changed it to a class to pacify the the linter because, as a non-maintainer, I did not feel I had the privilege to add a suppression comment.

evnchn:
Oh that's right! I think I'll let Falko decide himself on the final call. I just want to ensure the tests actually pass last midnight 🙇🏻‍♂️

evnchn:
Oh that's right! I think I'll let Falko decide himself on the final call. I just want to ensure the tests actually pass last midnight 🙇🏻‍♂️

evnchn:
But then note that "do the dererence only once" will cause it no longer possible to change the frequency midway during runtime, which may be occuring in user code (we can't say for sure)

Anyways it's the final stretch for this PR, just some minor decisions but overall the logic is established. As always, thanks for the work 😊

bulletmark:
> But then note that "do the dererence only once" will cause it no longer possible to change the frequency midway during runtime, which may be occuring in user code (we can't say for sure)

The two lines I quote above are adjacent and within the 0.1 sec loop. There can be no user code running between them.

evnchn:
Because in https://github.com/zauberzeug/nicegui/commit/33033529555cabb6885afd6a1c3afb5a70997172, before the change the `interval` was set once outside the loop. By moving it inside the loop we should be good to go?

evnchn:
> It shows how ... could cause a major bug

I feel "sniped" by the contextvar 😭 ... 

It makes the function, essentially, "dependent on where it was called". 

Anyways https://github.com/zauberzeug/nicegui/pull/5438/commits/8de03fa0a95a2ecf055cf0e5ce0a089212fe1a0d is my take at solving the problem without messing with the contextvar by keeping where it was called. 

I hope you like this solution. If not, either of you can force-push it away `git reset --hard HEAD~1` `git push --force`

evnchn:
Wishlist entry: document our use of contextvar and ask agents to be weary of it when a function's call position is moved, despite otherwise may yield same results. 

Feel like this could cause some nasty bugs if words are not spread about it.

evnchn:
`assert interval is not None` fails on Py3.9 but not other versions. No idea why. 

https://github.com/zauberzeug/nicegui/pull/5438/commits/c655c62a80fa5493aabdf61c171c0725cd285116 fixes things symtomatically. 

And `_ensure_refresh_loop_is_running` is no more (name is no longer accurate as well). 

The commit has an advantage: The binding loop will not crash despite repeated attempts to set `core.app.config.binding_refresh_interval = None`. Just a bit more robust.

falkoschindler:
I really like the approach, but somehow the test is still failing.
And I wonder why we need need to re-assign `_refresh_loop_active` in an except block.

evnchn:
> why we need need to re-assign _refresh_loop_active in an except block.

Due to asyncio event loop changed, which happens in pytest. 

Failing test is mysterious though....

evnchn:
Tests are passing @falkoschindler

falkoschindler:
@evnchn Tests were failing for some Python versions because the exception message slightly changed from "...attached to a different loop" to "bound to a different loop". But I think I found an even better, more robust solution. Let's see if the CI pipeline is happy.

evnchn:
> Let's see if the CI pipeline is happy.

Quite unhappy indeed. The moment the `await _active_links_added.wait()` spans another event loop, which it does with a new pytest due to the pytest event loop scoping, it errors out. Without a try-except it always blows stuff up. 

I don't see how looking for `attached to a different loop` / `bound to a different loop` can be optional...

falkoschindler:
Before trying to find a new (or old) fix, I'd like to reproduce the problem locally. But tests are green on my machine. 🫤

evnchn:
> But tests are green on my machine.

Are you on Py3.9? later versions just kinda got over the differing event loop issue.

falkoschindler:
I don't get it... At least I can reproduce the problem with my new approach when running pytest on Python 3.9.
But `pytest -k binding` even fails on e43466d1162bdb18c8ca99845fb08d30e6561f04 (your last commit). I tried with pyenv and Python 3.9.18 as well as the corresponding devcontainer.

evnchn:
According to GitHub, https://github.com/zauberzeug/nicegui/commit/e43466d1162bdb18c8ca99845fb08d30e6561f04 passes all tests though.

falkoschindler:
5ea4559b8ac785b393bebcdbd66ce4c5fb121607 seems to work.
Ready to merge if you ask me. 👍🏻

evnchn:
I get your logic. Basically by assigning the new event, we ensure it's always at the right event loop. 

Let's go 🚀

PR #5439, opened 2025-11-12
Patch and warn late event registrations

evnchn:
### Motivation

Reading #4154 and #4828 again, I come to the following conclusion:

1. Late event registrations are fundamentally misaligned with Vue's design. (reference: https://github.com/zauberzeug/nicegui/issues/4154#issuecomment-2885368501)
2. Therefore, we do not need to aim for perfect behaviour if late event registrations do occur (so https://github.com/zauberzeug/nicegui/pull/4828#issuecomment-2940042694 should not matter)
3. Rather, "best effort patch" and "warning to the user" is the way to go. 

Hence this PR. 

### Implementation

Code is extremely similar to #4828 

> Before we process the list of elements, we do a check to see if there are any new events sent in by the server. 
> - If there is, then:
>   - We remove that element from Vue's `this.elements`
>   - We mark that we need to `await this.$nextTick();`
>   - We proceed to `await this.$nextTick();` to force the Vue to render without that element
>     - Do it ONCE only, for performance
>   - In such a way, when we process the list of elements as usual, Vue will be like 
>     - "hey I haven't seen this element before, let me render it"
> - If there is none, then fine, do nothing. 

But on top, I added warnings both in Python-land and JS-land. 

### Progress

- [x] Documentation is not necessary (warning message self-explanatory). 

### Final notes

How I read it, #4828 was closed because: 

- Accidental closure
- Significant merge conflict due to NiceGUI 3.0
- Focus on perfect behaviour (no focus drops)

Both of which should no longer be blockers, given the new code and new positioning on the issue. 

So hopefully this gets a pass. Thanks!

### DIFF
--- a/tests/test_socketio_too_long.py
+++ b/tests/test_socketio_too_long.py
@@ -22,3 +22,6 @@ def test_socketio_too_long(screen: Screen, transport: Literal['websocket', 'poll
 
     assert events == ['changed'] * 2, 'two more characters are ok'
     assert len([log for log in screen.selenium.get_log('browser') if 'Payload size' in log['message']]) == 3
+    screen.assert_py_logger('ERROR', 'Payload size 999901 exceeds the maximum allowed limit.')
+    screen.assert_py_logger('ERROR', 'Payload size 999902 exceeds the maximum allowed limit.')
+    screen.assert_py_logger('ERROR', 'Payload size 999903 exceeds the maximum allowed limit.')


evnchn:
1. yes
2. https://github.com/zauberzeug/nicegui/pull/5439/commits/6a1abb684b1e46454ee177cb92d190a41225a52b
3. https://github.com/zauberzeug/nicegui/pull/5439/commits/5d65b759ebb2af6cadbdca93be4b37bc3a60749d
4. https://github.com/zauberzeug/nicegui/pull/5439/commits/61c7cfa4c39a658a4df0d6fb04295e0c3925269d

evnchn:
Actually do we really need such a big popup for re-rendering the element. Maybe just invoking the Quasar loading bar at the top of the page for a split second is enough to indicate some loading is happening. 

Regardless I think both is fine because either way late-registrations are not recommended.

falkoschindler:
I didn't expect you to change any popups. I was just thinking about merging multiple errors/warnings into a single message topic. We probably shouldn't implement so many popups, especially for technical details like an event registration. Logging the warning should suffice for the developer to notice.

Oh and tests are failing right now.

evnchn:
Should be good:

- No more popup. It just `console.warn`s. Should be enough to get an attentive person's attention. 
- Fix the tests and make it stricter in the meantime.

falkoschindler:
I just introduced a "log" message as an easy way to log warnings and errors both in JavaScript and Python. This replaces the hacky "too_long_message" message. The test is green. Here is my code for experimentation:
```py
# experiment with large payloads
ui.textarea(value='x' * (1_000_000 - 242), on_change=lambda e: ui.notify(f'{len(e.value)} characters'))

# experiment with late event registration
name = ui.input()
name.on('keydown.a', ui.notify)
ui.timer(1.0, lambda: name.on('keydown.b', ui.notify), once=True)
```
Do you agree, @evnchn?

nicegui/static/nicegui.js
```diff
@@ -348,12 +360,12 @@ function createApp(elements, options) {
             return function (...args) {
               const msg = args[0];
               if (typeof msg === "string" && msg.length > MAX_WEBSOCKET_MESSAGE_SIZE) {
-                console.error(`Payload size ${msg.length} exceeds the maximum allowed limit.`);
-                args[0] = '42["too_long_message"]';
+                logAndEmit("error", `Payload size ${msg.length} exceeds the maximum allowed limit.`);
```
evnchn: This is not a valid move if you use polling transport. Each polling transport message must be delivered to the server or there will be HTTP response code 400 and Socket.IO emergency resets.

nicegui/static/nicegui.js
```diff
@@ -139,6 +150,7 @@ function throttle(callback, time, leading, trailing, id) {
     }
   }
 }
+
```
evnchn: extra blank line?

falkoschindler: Yes, unrelated to this PR, but I spotted a missing blank line between top-level functions. Never mind, we can leave it unchanged.

nicegui/static/nicegui.js
falkoschindler: Good catch!

PR #5441, opened 2025-11-12
Use local audio to avoid flaky test

evnchn:
### Motivation

Flaky test in https://github.com/evnchn/nicegui/actions/runs/19286240615/job/55147628997 due to upstream HTTP error. 

### Implementation

- Generate the audio files ourselves. 
- Use those audio files for testing using custom fixture. 

### Progress

- [x] Documentation is not necessary (it's a pytest).

### Final notes

If there are any external HTTP requests in pytest, do let me know, I'll do my best to get rid of them.


evnchn:
Yes we can leave them until they cause a test to fail, at which point we can fix in due course.

Issue #5444, opened 2025-11-12
q-select__dropdown-icon does not animate after the introduction of PR  5405


Moldeador:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

select1 = ui.select([1, 2, 3], value=1)

ui.run()
```

### Description

The rotation of the icon in the `ui.select` element and in the `ui.dropdown_button` is not animated.

There was recently a fix in PR #5405 that fixed the rotation angle for any `ui.icon`, an issue that was raised in #5399 .
Now the angle is fixed, but the fix has stopped the Quasar CSS transition in bigger components that use the icon, like `ui.select` and `ui.dropdown_button`, from working, so the rotation is not animated.

I think the cause is that by fixing it, in #5405 the Quasar CSS rule for rotation was changed from `transform` to `rotate`. However, the Quasar CSS rule in the selec component that animated the rotation still refers to the `transform` transition. Therefore the Quasar select component does not animate anymore. The CSS rules for this icon inside the component, after the regular expression, are now:

```
@layer quasar {
  .rotate-180 {
    rotate: 180deg;
  }
}
@layer quasar {
  .q-select__dropdown-icon {
    cursor: pointer;
    transition: transform 0.28s;
  }
}
```

### NiceGUI Version

3.2.0.post2.dev0+b72e7bbe

### Python Version

Python 3.9.25

### Browser

Edge, Firefox

### Operating System

Windows

### Additional Context

In this screenshot the green arrow shows that Quasar and Tailwind do not clash on the angle of rotation after the #5405 fix. The red arrow shows that now the Quasar transition still targets a `transform` rule that no longer exists.

<img width="1488" height="909" alt="Image" src="https://github.com/user-attachments/assets/5980183f-9444-4323-a799-61062923d9ad" />

evnchn:
Hello @Moldeador 

You are totally right. 

I'm just stuck as to how I can fix this, because I can't effectively grep the CSS when the conflict comes from apply/unapply `rotate-*` class. That logic should rather be in Quasar JS. 

Compared to #5399 (doesn't work) this is a smaller issue (works without animations) so I don't think we would revert #5405. 

For the time being animations may be unavailable. Sorry for that 😐 

[...]

PR #5405, opened 2025-11-05
Introduce `_patch_rotate` on Quasar CSS

evnchn:
### Motivation

> People who mix Quasar with Tailwind or other utility frameworks are a minority (https://github.com/zauberzeug/nicegui/issues/5156#issuecomment-3324426595)

We knew we were fighting an uphill battle alone with using both Quasar and Tailwind from day 1, but we really didn't envision a situation like #5399 where:

- Past impl: both uses `transform` so we get once the effect
- New impl: one uses `transform` one uses `rotate` so we get twice the effect. 

Well, welcome to web dev I guess 🤡 

### Implementation

Regex will look for this:

```css
.rotate-BLAH {
  transform: rotate(BLAHdeg)
    /* rtl:ignore */
  ;
}
```

and make it into this:

```css
.rotate-BLAH {
  rotate: BLAHdeg
    /* rtl:ignore */
  ;
}
```

Then the issue is fixed. 

### MRE

```py
ui.icon('arrow_forward').classes('rotate-90')
```
Should point down, not left. 

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests have been added (slightly necessary but don't know how to read computed rotation style...).
- [x] Documentation has been added (or is not necessary).

Issue #5399, opened 2025-11-04
Tailwind rotate-... classes applied twice in NiceGUI 3.x


henriknoren:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

ui.icon('arrow_forward').classes('rotate-90')

ui.run()
```

### Description

Adding rotate-90 results in a 180 degree rotation since the class is applied twice. Looking at chrome dev tools you can see the difference between NiceGUI 2 and 3.

### NiceGUI Version

3.2.0

### Python Version

3.13

### Browser

Chrome

### Operating System

Windows

### Additional Context

In Chrome DevTools you can see that the class is applied twice. Disabling one of them gives the right behavior. 
In NiceGUI 2.24.0 one of the styles is crossed over, not sure exactly what it means, but it works as intended.

<img width="405" height="807" alt="Image" src="https://github.com/user-attachments/assets/5b0ba4f9-532b-4e5e-95c3-6098234a9705" />

<img width="406" height="846" alt="Image" src="https://github.com/user-attachments/assets/9caea669-8161-45e6-9654-ec9d5eccc826" />

falkoschindler:
Thanks for reporting this bug, @henriknoren!

It looks like the Tailwind version in NiceGUI 2.x implemented the class "rotate-90" like Quasar via `transform:`, so one rule replaced the other one.
Now Tailwind implements rotations via `rotate:`, so both rules complement each other.

At the moment I'm not sure how to fix that.

falkoschindler:
As a workaround you can mark the class as important:
```py
ui.icon('arrow_forward').classes('!rotate-90')
```
The prefix `!` only works for Tailwind classes, so Quasar won't add its 90 degrees.

henriknoren:
Thanks for that!

I was using the workaround .style('rotate: 90deg') instead, but I will change to yours instead.
This is a low priority issue, just thought you should know.

Thanks for your great work!

evnchn:
Proposed solution (see if this checks out):

1. Grep the quasar CSS to look for classes which correspond to Tailwind CSS classes
2. Manually screen for shenanigans
3. Change Quasar around Tailwind. (because Tailwind CDN JS is a black box)

[...]

PR #5445, opened 2025-11-12
`headwind.css`: Quasar-Tailwind compatibitlity by disabling Tailwind

evnchn:
### Motivation

Props to @Moldeador for #5444, which highlights how #5405 caused Quasar to lose animations in fixing #5399. 

Moral of the story: Do not touch Quasar internals. 

### Implementation

So, I went ahead and disabled the conflicting classes of Tailwind, calling it `headwind.css` (headwind is opposite to tailwind, I thought). 

And indeed as mentioned in https://github.com/zauberzeug/nicegui/issues/5444#issuecomment-3522024598 I act only on classes which Quasar has an equivalent. 

### Progress

### Final note

**Will this break Tailwind animations which animate from `rotate-45` (Quasar's class) to `rotate-100` (not Quasar's class)???**

### DIFF
--- a/nicegui/static/headwind.css
+++ b/nicegui/static/headwind.css
@@ -1,21 +1,21 @@
 .rotate-45 {
-  rotate: 0deg !important;
+  rotate: 0deg;
 }
 .rotate-90 {
-  rotate: 0deg !important;
+  rotate: 0deg;
 }
 .rotate-135 {
-  rotate: 0deg !important;
+  rotate: 0deg;
 }
 .rotate-180 {
-  rotate: 0deg !important;
+  rotate: 0deg;
 }
 .rotate-225 {
-  rotate: 0deg !important;
+  rotate: 0deg;
 }
 .rotate-270 {
-  rotate: 0deg !important;
+  rotate: 0deg;
 }
 .rotate-315 {
-  rotate: 0deg !important;
+  rotate: 0deg;
 }


evnchn:
> Will this break Tailwind animations

Worst case if it does, we just have to educate user to do `rotate-[45deg]` to dodge Quasar. I don't see any other ways around it.

falkoschindler:
> Will this break Tailwind animations which animate from `rotate-45` (Quasar's class) to `rotate-100` (not Quasar's class)???

Good question! I tested with the following code:
```py
def test(from_angle: int, to_angle: int):
    icon = ui.icon('arrow_forward').classes(f'rotate-{from_angle} transition-transform')
    ui.button('Rotate', on_click=lambda: icon.classes(remove=f'rotate-{from_angle}', add=f'rotate-{to_angle}'))

test(45, 90)
test(45, 91)
test(46, 90)
test(46, 91)
```

Everything works fine. And I can't find another way to implement animations which would work in 3.2 and break with this PR.

falkoschindler:
@Moldeador Yes, sorry for the confusion. You're totally right!

falkoschindler:
@evnchn Thanks for the quick solution! Seems to work pretty nicely. Just one question before merging:
Why do you think we need `!important` in headwind.css? Everything seems to work without it.

evnchn:
> `!important` in headwind.css

We don't need it. I thought I dropped it before I commit but it was gone for some reason 🤡. Anyways https://github.com/zauberzeug/nicegui/pull/5445/commits/0ecbb1c45b0490af8f75a3f3f34f66eea6c71aad

evnchn:
While it doesn't break things right now, best practice we should disable this when `tailwind=False`

evnchn:
https://github.com/zauberzeug/nicegui/pull/5448

Issue #5318, opened 2025-10-22
ui.tree.update() not working


roddy-zhang:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

data = [{"id":"1", "label":"apple"}{"id":"2", "label":"banana"}]
tree = ui.tree(data)

# after do something
# or
# in some method
data.append({"id":"3", "label":"peach"})
tree.update()  # it doesn't work!
```

### Description

When I update the data and call the ​update()​​ method, the ​ui.tree​ object does not refresh its display. However, this method worked correctly in version ​2.24.2.

### NiceGUI Version

3.0.4

### Python Version

Python 3.13.2

### Browser

Firefox

### Operating System

macOS

evnchn:
Hello @roddy-zhang thanks for the report. 

Most certainly it is the breaking change mentioned in the release notes of 3.0.0, but turns out there is one more element affected by it. 

> 3. Modifying the content of `ui.table`, `ui.aggrid`, `ui.echart`, `ui.highchart`, and `ui.json_editor`
> 
>    Before NiceGUI 3.0 it was possible create tables or AG Grids like `ui.table(columns=columns, rows=rows, ...)` or `ui.aggrid({'columnDefs': columns, 'rowData': rows, ...)` and update it later by modifying the _original_ `columns` or `rows`. This is not possible anymore.
> 
>    In NiceGUI 3.0 you need to modify `table.rows`, `table.columns` or `aggrid.options` instead. A subsequent `table.update()` or `aggrid.update` is not necessary anymore.
> 
>    Same holds for `ui.echart`, `ui.highchart` and `ui.json_editor`.

Consider this fix:

```py
from nicegui import ui

data = [{'id': '1', 'label': 'apple'}, {'id': '2', 'label': 'banana'}]
tree = ui.tree(data)

tree.props['nodes'].append({'id': '3', 'label': 'peach'})
tree.update()

ui.run()
```

@falkoschindler It seems like we should document this better...

roddy-zhang:
Thank you for your reply. I just started trying to migrate a project from version 2.24.2 to version 3.0 yesterday, but I couldn't find any information on what to pay attention to during the version transition. Could you please provide me with a link?

[...]

PR #5446, opened 2025-11-12
Improve docstrings and demos about updating content elements

falkoschindler:
### Motivation

In #5318 we noticed that the documentation is not very clear about how to update tables, AG Grids, and other content elements in NiceGUI 3.

### Implementation

Improved docstrings:
- `ui.table`: "Updates can be pushed to the table by updating the `rows` or `columns` properties."
- `ui.aggrid`: "Updates can be pushed to the grid by updating the `options` property."
- `ui.highchart`: remove sentence about calling the `update` method

Updated demos:
- `ui.table` demo "Adding rows" to use `table.rows.append`
- add "Adding rows" demo for `ui.aggrid`
- add "Update content" demo for `ui.json_editor`

### Progress

### DIFF
--- a/nicegui/elements/tree.py
+++ b/nicegui/elements/tree.py
@@ -22,6 +22,7 @@ class Tree(FilterElement):
         """Tree
 
         Display hierarchical data using Quasar's `QTree <https://quasar.dev/vue-components/tree>`_ component.
+        Updates can be pushed to the tree by updating ``.props['nodes']``.
 
         If using IDs, make sure they are unique within the whole tree.
 
--- a/website/documentation/content/tree_documentation.py
+++ b/website/documentation/content/tree_documentation.py
@@ -11,6 +11,19 @@ def main_demo() -> None:
     ], label_key='id', on_select=lambda e: ui.notify(e.value))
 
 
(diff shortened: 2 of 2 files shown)


falkoschindler:
@evnchn I just noticed that I completely overlooked the actual critical element `ui.tree`, which was the root cause for #5318. I'll see what I can do about `ui.tree`, which - unfortunately - doesn't have a `nodes` property for updates.

falkoschindler:
Unfortunately the name `nodes` is already taken and it would be a hassle with breaking changes to arrive at a `nodes` property containing the original node hierarchy. I chose to simply document the status quo.

evnchn (approved):
All documentation. Simple-win. Port 8080 is occupied as of writing so I trust that you have tried everything out. If so let's go 🚀 

---

I think we should prioritize documenting the things how it is, and then figure out if we need any API changes later (TM).

This helps to divide things into manageable chunks into separate PRs and helps keep track of what we have changed by looking how the documentation changed. 

I would prefer everything a user needs to do is done in Python API (no writing to `.props`) because it is easier to discover (think VSCode autocomplete), but you just said it was not possible? We can investigate later.

PR #5449, opened 2025-11-13
Put SASS and SCSS compiling in the browser

evnchn:
### Motivation

- Annoying to have to install a library to use SASS or SCSS
- LibSass is deprecated, and that comes with it https://github.com/sass/libsass-python
- No plans for a Dart Sass Python binding replacement: https://github.com/sass/dart-sass/issues/2166
- **We will die on November 30, 2025 if nothing gets done**: https://github.com/zauberzeug/nicegui/discussions/5321#discussioncomment-14960698

### Implementation

- Let the importmap be before the head HTML for it to import stuff (should have been the case since day1 but anyways...)
- Follow https://github.com/sass/dart-sass/blob/main/README.md#dart-sass-in-the-browser
- Drop check for installed Python library
- Deprecate `indented` since it doesn't matter any more (cannot configure)
- `immutable.es.js` had trailing whitespaces. Patch that. 

### Progress

- [x] Pytests already have.
- [x] Documentation is not necessary (self-explanatory).

### Final notes

https://github.com/zauberzeug/nicegui/discussions/5321#discussioncomment-14961222 lasted 33 minutes before it was outdate 🚀

### DIFF
--- a/nicegui/functions/style.py
+++ b/nicegui/functions/style.py
@@ -20,12 +20,14 @@ def add_css(content: Union[str, Path], *, shared: bool = False) -> None:
     add_head_html(f'<style>{content}</style>', shared=shared)
 
 
-def add_scss(content: Union[str, Path], *, indented: bool = False, shared: bool = False) -> None:
-    """Add SCSS style definitions to the page.
+def add_scss(content: Union[str, Path], *, indented: bool = False, shared: bool = False) -> None:  # DEPRECATED
+    """Add SCSS style definitions to the page (deprecated).
 
     This function can be used to add SCSS style definitions to the head of the HTML page.
 
     *Added in version 2.0.0*
+
+    **Note: This function is deprecated and will be removed in NiceGUI 4.0. Use add_css instead.**
 
     :param content: SCSS content (string or file path)
     :param indented: whether the content is indented (SASS) or not (SCSS) (default: `False`)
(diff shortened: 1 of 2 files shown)


evnchn:
I don't know enough about SASS and Dart Sass to debug https://github.com/evnchn/nicegui/actions/runs/19340914024/job/55328764256

falkoschindler:
This was quick, @evnchn! 🚀

I haven't tried it out, but the PR looks rather straightforward. A few remarks though:

- `add_head_html` is called for every `add_scss` call. This might accumulate lots of \<script\> tags on the client. But maybe there's no better solution without overcomplicating things. `add_scss` is probably not called many times per page.
- I'm worried about 5.2MB of JavaScript added to the library. Strangely there doesn't seem to be a minimized JS file in node_modules/sass/, even though the example on https://www.jsdelivr.com/package/npm/sass shows it.
- See https://github.com/zauberzeug/nicegui/discussions/5321#discussioncomment-14967468: I'm thinking about simply dropping the whole SASS/SCSS feature.

evnchn:
- We would have many <style> tags anyways. Let's leave it. 
- 3MB min, 0.6MB min+gzipped. https://manus.im/share/U7m7TBcTZrcrNJv4MEXcmW?replay=1
  - Still large but not terrible. Nevertheless the performance-aware should not use this. 
- I'm fine dropping it, but we should have deprecation.

evnchn:
For the size concern, can we put this new functionality behind a package (think `nicegui-sass`) so that not everyone downloads 3MB extra on disk?

And we can log warn when install that package saying it may be gone any time?

For proper deprecation we need a stopgap and notice time, but for this functionality I think forgoing it is a possibility

falkoschindler:
Somehow rjsmin only achieved 3.9 MB.
So switched to using `npx terser`. This way we don't need an additional dev dependency (we need npx though) and the file size goes down to 3.0 MB.

evnchn:
Good. Let's merge queue.

Issue #5395, opened 2025-11-03
Intermittent 404 error on files in /_nicegui/3.2.0/components/* using Azure Web App


hub-il:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

@ui.page('/')
def main_page():
    ui.label('Hello from NiceGUI on Azure!')

    columns = [
    {'name': 'name', 'label': 'Name', 'field': 'name', 'required': True, 'align': 'left'},
    {'name': 'age', 'label': 'Age', 'field': 'age', 'sortable': True},
    ]
    rows = [
        {'name': 'Alice', 'age': 18},
        {'name': 'Bob', 'age': 21},
        {'name': 'Carol'},
    ]
    ui.table(columns=columns, rows=rows, row_key='name')

    ui.button('Click me', on_click=lambda: ui.notify('Button clicked!'))

ui.run(port=8000)
```

### Description

Trying to run the application in Azure Linux WebApp (Linux, Python 3.13).  FastAPI applications appear to work correctly without issue.

When running the code above **WITH** the table I will get a 404 file not found error for the file INFO:      `/_nicegui/3.2.0/components/36bfe96fc17f456afc8755ab027b54ee/table.js` but if I keep refreshing that file specifically, the file it will eventually get served to me.  If I try to pull the file using `curl` it will never be served, it always comes up a 404.

This is also true for all other files in the `/_nicegui/3.2.0/components/` folder that my larger application uses including `header.js`, `query.js`, `link.js`, `image.js`, `table.js`, `timer.js`.

If I remove the table, the site is served as expected but will refresh between 1 and 5 times before it is able to be interacted with.  No errors noted in Chrome Dev tools.

I do not get this issue when running locally.  As of now, the site is non-functional until I can resolve this issue.  

### NiceGUI Version

3.2

### Python Version

3.13

### Browser

Chrome

### Operating System

Linux

evnchn:
Hello @hub-il 

I think I know where this is pointing from the clue below: 

> remove the table, will refresh between 1 and 5 times before it is able to be interacted with

[...]

PR #5450, opened 2025-11-13
Check `core.app._uuid` and warn if repeated mismatch (no session affinity?)

evnchn:
### Motivation

In light of NiceGUI used in cloud with auto worker scaling, we found that the fact that NiceGUI needs session affinity isn't really well known and leads to quite some headscratchers. 

https://github.com/zauberzeug/nicegui/issues/5395#issuecomment-3483971400

https://github.com/orgs/zauberzeug/projects/6/views/1?pane=issue&itemId=137123267

### Implementation

- Add `core.app._uuid` which is served HTML-land and checked SocketIO-land. 
- If mismatch 3+ times, warn and ask for further retries. 
- Should catch more than 1 worker: the more workers there are, there easier this gets triggered. 

### Progress

### Final notes

Locally, to simulate the mismatch, do:

```py
if server_id != core.app._uuid or True: # pylint: disable=protected-access
```


falkoschindler:
@evnchn Thanks for the pull request, that's a great idea!

But we think the error should be shown on the server, not on the client. This would simplify the JavaScript code and provide opportunity to link to educational documentation for the developers.

evnchn:
#### Concern 1: Not sure the detection accuracy on the server side. 

Obviously wrong `server_id` from an incoming connection can be due to worker mismatch. 

But it can also be due to a client which while it was disconnected from internet, the server has rebooted. And there can be infinitely many `server_id`s caused by this, as clients may be from an arbitrary number of server reboots ago. 

I am not quite sure of a reliable heuristic which I can use to distinguish the former from the latter, and I chose browser-based detection because it's easier. If you have any ideas on detection on server side, please tell me. 

#### Concern 2: Stopping browser infinite-reload is still necessary

The infinite reload may put great stress at the server hosting setup, and may manifest into large cloud bills if cloud was used. 

So even if the warning is shown on the console, I think the client should still stop trying after 10 attempts or so.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5450/commits/0c2db43fc3bb5b4261f9a11f4f52d9bdc4a926b0: 

- Detection accuracy is no worse than before, since on 3 - 20 failed connection attempts, client is just retrying to "spread the word to workers"
- Browser still stops at 20 reloads

I consider it a pretty good solution. I am not sure about the error messages though.

falkoschindler:
Some tests seem to be failing. Let's see if a merge from main helps.
Anyway, I'll move it to version 3.5.

falkoschindler:
Since some pytests are still red and I'm busy with other PRs, I'll reschedule this to 3.x.

evnchn:
_(posted by Claude Code on Evan's behalf)_

Rebased onto current `main` and integrated with the `implicit_handshake` path that landed upstream after this PR was written. TL;DR: `_on_handshake` returning the truthy string `'WRONG_SERVER_ID'` would have silently slipped through the new `if not await _on_handshake(...)` check in `_on_connect`, so a multi-worker-no-affinity client whose connection landed on the wrong worker would proceed past the connect handler and break later — not what either of us wanted.

CI: mypy clean, pylint 10.00/10 on the touched files, 89 local pytests green (`test_user_simulation`, `test_socketio_too_long`, `test_reconnect`, `test_run`, `test_user_simulation_context`). Full diff vs `main` is +42/-5 across 5 files.

<details>
<summary>What's new since the last review</summary>

**Server side**

- `client.py` — `server_id` is now lifted into `socket_io_js_query_params` (instead of just the template context). It rides along with the connection query string so the **implicit_handshake** path in `_on_connect` can verify it, not just the explicit handshake event.
- `nicegui.py` — `_on_connect` now distinguishes the `'WRONG_SERVER_ID'` return from a generic `False` and raises a separate `ConnectionRefusedError('Wrong server ID')`. Without this, the `not result` check let mismatched clients slip through.
- `nicegui.py` — `_on_handshake` int-casts `data.get('mismatches', 0)` because in the implicit path it arrives as a URL-encoded string from `parse_qs`, not as a JSON int from the explicit handshake payload. Without the cast, `'5' > 3` raises `TypeError` in Python 3.
- `templates/index.html` — dropped the now-redundant `serverId: "{{ server_id }}"` since the value comes through `options.query` instead.

**Client side (nicegui.js)**

- `handleWrongServerId()` hoisted to `mounted()` scope so the same logic is reachable from `finishHandshake` (explicit) AND `connect_error` (implicit, via the new `"Wrong server ID"` error message).
- `options.query.mismatches` is injected from `sessionStorage` before `io()` init, so the server-side warn-threshold works the same in both modes.
- Explicit-handshake JS now uses `options.query` directly instead of building a separate `args` dict — matches upstream's pattern after #5xxx.

**Testing**

- `testing/user.py` — fixture handshake now passes `server_id: core.app._uuid` so user-mode tests don't trip the new check.

</details>

Threshold-tightening discussion from #5802 (whether `mismatches > 3` is too sensitive against pathological clients like the iOS 26 / WebKit bug-302823 reload-looper) is intentionally **out of scope here** — this PR is about plumbing the detection through; the threshold tuning can be a follow-up if the warning surfaces false positives in the wild.

evnchn:
@falkoschindler Note: Somehow https://github.com/zauberzeug/nicegui/pull/5450/commits/a10c78e5e93c283817e224abb9efcf6c9ba65e20 missed Claude co-authorship again. I think maybe Claude Code model behaviour change or something messed up with my install's memory. Will investigate. 

I went through https://github.com/zauberzeug/nicegui/milestone/140 for triage PRs by me, and this turns out is the easiest.

evnchn:
_(posted by Claude Code on Evan's behalf)_

Heads-up: force-pushed `a10c78e5 → f558f303` to fold a ruff-check fix and the Claude Code co-authorship trailer into the merge commit, since the prior push left `pre-commit` red. Content delta is only the type-annotation modernization:

```diff
-from typing import Any, Literal, Union
+from typing import Any, Literal
 ...
-async def _on_handshake(sid: str, data: dict[str, Any]) -> Union[bool, Literal['WRONG_SERVER_ID']]:
+async def _on_handshake(sid: str, data: dict[str, Any]) -> bool | Literal['WRONG_SERVER_ID']:
```

`git range-diff 308f04c2..a10c78e5 308f04c2..f558f303` shows every other commit unchanged.

falkoschindler:
This has been postponed many times, I haven't read the code in detail for a long time, and nobody seems to be missing it. Therefore I'm downgrading it to "Next".

evnchn:
You know what, I am closing it directly. 

I don't think NiceGUI must obliged to add warnings for "N ways you can mess up NiceGUI", especially when the detection isn't low cost, and I can't verify if it works with iOS 26 devices with sketchy websocket.

PR #5454, opened 2025-11-15
Fix spurious ui.table "found list in column" warnings

Mick235711:
### Motivation

Follow-up to #5290. Unfortunately, QTable's slots `body-cell-[name]` actually refers to column names, not column fields. This discrepancy caused a spurious "found list in column" warning (and unnecessary slot addition) for columns that have custom slots defined already.

### Implementation

Just a simple one-line fix.

### Progress


Mick235711:
Ah, my bad for not noticing the usage of field below. The example should work properly now.

Issue #5459, opened 2025-11-17
Event on ui.plotly is not working (even on the official webpage)


mathisschmidt:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
import plotly.graph_objects as go
from nicegui import ui

fig = go.Figure(go.Scatter(x=[1, 2, 3, 4], y=[1, 2, 3, 2.5]))
fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
plot = ui.plotly(fig).classes('w-full h-40')
plot.on('plotly_click', ui.notify)

ui.run()
```

### Description

Hello, 

I tried to use the event 'plotly_click' but it is not working. I tried the example given in the documentation on Firefox & edge, same result nothing works.

### NiceGUI Version

3.3.0

### Python Version

Python 3.13.2

### Browser

Edge, Firefox

### Operating System

Linux

falkoschindler:
Oh wow, thank you so much for reporting this bug, @mathisschmidt!

MRE:
```py
ui.plotly({'data': [{'x': [1], 'y': [1]}]}).on('plotly_click', ui.notify)
```

It looks like PR #5404 broke it.

mathisschmidt:
So, do you know when the fix will be published?

falkoschindler:
@mathisschmidt I'm just waiting for the CI pipeline to finish testing before triggering the release. 🚀

mathisschmidt:
Perfect thank you, very fast team! And nice tool.

PR #5460, opened 2025-11-17
Plotly `this.last_options_set`: fix wrongful re-use plotly instance on first-spawn

evnchn:
### Motivation

A logic error in value change check in `ui.plotly` revealed itself with #5404, which instead of force-set `options.config = { responsive: true }` on behalf of user, force-unsets it instead. 

### Implementation

Since `this.last_options.config` is undefined at first, `if (JSON.stringify(options.config) == JSON.stringify(this.last_options.config))` is not a sufficient check in the case that `this.options.config` is also undefined, which with #5404 is the likely possibility. 

I added `this.last_options_set` to force the first time render to always false the if-condition and so `this.set_handlers();` is ran. 

### Progress

- [x] Pytests are not necessary
  - I am considering to add regression tests, but #5459 is a critical. 

### DIFF
--- a/nicegui/elements/plotly/plotly.vue
+++ b/nicegui/elements/plotly/plotly.vue
@@ -32,7 +32,7 @@ export default {
       if (options.config?.responsive === true) options.config.responsive = undefined;
 
       // re-use plotly instance if config is the same
-      if (this.last_options_set && JSON.stringify(options.config) == JSON.stringify(this.last_options.config)) {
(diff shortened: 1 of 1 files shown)


evnchn:
Actually https://github.com/zauberzeug/nicegui/pull/5460/commits/4b098b420696f018b3e32d11e15e080d596c5ce1 is better since `{}` is `true` in JS, so we can basically use that for the `this.last_options_set`

PR #5461, opened 2025-11-17
Let `element.clear()` return the element (builder pattern)

falkoschindler:
### Motivation

It would be nice if we could replace
```py
container.clear()
with container:
    ...
```
with
```py
with container.clear():
    ...
```

### Implementation

This PR lets `element.clear()` return the `element` so that the method can be chained in a builder pattern like `classes()`, `props()`, `style()` etc.

### Progress

### DIFF
--- a/tests/test_navigate.py
+++ b/tests/test_navigate.py
@@ -31,7 +31,7 @@ def test_navigate_to(screen: Screen, new_tab: bool):
 
 
 def test_navigate_to_absolute_url(screen: Screen):
-    external_url = 'https://nicegui.io'
+    external_url = 'https://www.google.com/'
 
     @ui.page('/')
     def page():


falkoschindler:
I fear that this is technically a breaking change and needs to wait for 4.0. A user could have overwritten `clear()` in a subclass and would need to return `self` now.

evnchn:
@falkoschindler It is not a breaking change in my eyes. 

Not returning `self` for a custom element defining `clear()` doesn't break code right away since users of such element would likely still in the old syntax which works. 

At worst their type checker will suddenly get unhappy but I think it's expected collateral. 

This is why I propose a more granular change management system than breaking/non-breaking. We should not let small technicalities get in the way of the Nice nature of the library.

evnchn:
Besides the point on whether this is to be delayed into 4.0, there is nothing to write home about this PR: quite simple-win in nature.

evnchn:
@falkoschindler I have thought long and hard as to whether this can be a breaking change. 

Short of "soft" breaks like type checker complaining, I did conjure up one scenario, but I think we actually should break that user's code because it is malpractice in the first place. See if you agree. 

Imagine a custom element `NeedsPrepBeforeEnter`

```py
npbe = NeedsPrepBeforeEnter()

# with nbpe: # raises RuntimeError('please prep_enter first')
#     ...

npbe.prep_enter() # possibly costly initialization

with nbpe: # now this will work
    ...
```

The new builder pattern essentially mandates that we return the element in a state which it can be entered in context. 

The maintainer of `NeedsPrepBeforeEnter` may see themselves facing a conundrum:

- break the mandate and return `self` as is, but `with npbe.clear():` will error out
- follow the mandate call `npbe.prep_enter()` inside `npbe.clear()` which makes `clear()` costly as well. 

But I'll counter with the following:

- none of NiceGUI's element methods should have been made costly in the first place, especially since it is sync (blocks main thread!)
- we never condone having an element which is only sometimes can be entered in context

---

And I did not think of any more scenario. So yeah I think we're safe to go, even consider advanced downstream use such as RoSys?

falkoschindler:
Right, Python doesn't complain about old code still overwriting `clear` without returning `self`. It's only mypy which isn't happy about it. But I think this is fine (and easy to fix).
I also checked RoSys, but there is no overwrite of `clear`.

pascalzauberzeug (approved):
Cool feature 🚀

PR #5463, opened 2025-11-18
Fix padding of horizontal `ui.stepper`

falkoschindler:
### Motivation

In #5434 we noticed that the padding of horizontal steppers is wrong since NiceGUI 3.0. More specifically, it is broken since f18f902780d850d503b6acfc3f847f87687941f8 where I fixed it for vertical steppers, but forgot to check horizontal ones.

### Implementation

Instead of patching the problem with a more precise CSS selector, I chose to improve the overall strategy. The critical CSS rules were described as "original padding for some Quasar elements". They were only necessary because the padding had been set to `var(--nicegui-default-padding)`.

This PR distinguishes between two groups of flex containers: For some we apply the NiceGUI default padding (`.nicegui-content`, `.nicegui-header`, ...) and for some we don't (`.nicegui-splitter .q-splitter__panel`, `.nicegui-step .q-stepper__nav`, ...). This way we don't need to fix the padding afterwards.

The only patch that remains is the 8px top padding for the stepper navigation. It compensates the gap which NiceGUI adds to children of `ui.step`. Otherwise the gap above the navigation would be too large.

Because this PR affects multiple UI elements, I tested with the following comprehensive code:
```py
ui.step.default_classes('border border-blue-500')
ui.label.default_classes('border border-red-500')
ui.timeline_entry.default_classes('border border-green-500')

with ui.stepper().props('vertical').classes('w-full') as stepper:
    with ui.step('Step 1'):
        ui.label('Label 1')
        ui.label('Label 2')
        with ui.stepper_navigation():
            ui.button('Next', on_click=stepper.next)
    ui.step('Step 2')

with ui.stepper().classes('w-full') as stepper:
    with ui.step('Step 1'):
        ui.label('Label 1')
        ui.label('Label 2')
        with ui.stepper_navigation():
            ui.button('Next', on_click=stepper.next)
    ui.step('Step 2')

with ui.timeline(side='right').classes('border'):
    ui.timeline_entry('Body 1', title='Title 1', subtitle='Subtitle 1')
    ui.timeline_entry('Body 2', title='Title 2', subtitle='Subtitle 2', icon='rocket')

with ui.splitter().classes('border') as splitter:
    with splitter.before:
        ui.label('Before content')
    with splitter.after:
        ui.label('After content')

with ui.row().classes('border'):
    ui.label('Label 1')
    ui.label('Label 2')

with ui.column().classes('border'):
    ui.label('Label 1')
    ui.label('Label 2')
```

This is the result (before and after this PR):

### Progress


evnchn:
Code works from the provided comprehensive code, which as far as I see, pixel-perfectly the same with exception to the horizontal `ui.stepper` which is fixed. 

The diff is a bit tough to read though. I tried https://semanticdiff.com/online-diff/css/ and to summarize it told me there's basically only 2 changes:

For "for some we apply the NiceGUI default padding, for some we don't" (`padding: var(--nicegui-default-padding);`), I got the exact list of changed elements: 
```.nicegui-splitter .q-splitter__panel,
.nicegui-step .q-stepper__nav,
.nicegui-step .q-stepper__step-inner,
.nicegui-timeline-entry .q-timeline__content,
.nicegui-sub-pages,
.nicegui-row,
.nicegui-column
```

And "patch that remains is the 8px top padding for the stepper navigation" refers to
```
.nicegui-step .q-stepper__nav {
  padding-top: 8px;
}
```

Which I think is reasonable. Full diff below: 

image


Finally a bonus item for peace of mind, can we quicky check if https://nicegui.io still works as intended? 

With that this PR is good to go. Thanks!

nicegui/static/nicegui.css
```diff
@@ -18,52 +18,37 @@
...
-}
 .nicegui-card-tight {
   padding: 0;
   gap: 0;
 }
 
-/* original padding for some Quasar elements */
-.nicegui-step .q-stepper__nav {
-  padding: 8px 0 0 0;
-  gap: 8px;
```
evnchn, line 55: Is this a change in behavior? 

Gap for `.nicegui-step .q-stepper__nav` used to be `8px` but now is `var(--nicegui-default-gap)`?

nicegui/static/nicegui.css
```diff
@@ -76,6 +61,9 @@
 .nicegui-stepper .q-stepper__step-inner {
   height: 100%;
 }
+.nicegui-step .q-stepper__nav {
+  padding-top: 8px;
+}
```
evnchn, line 67: And so do we shove `gap: 8px` here?

evnchn (changes requested):
Might have spotted an unnecessary change in behavior? (we're fixing padding, why change the gap?)

This, and possibly https://nicegui.io live testing (@rodja responsible for deployment?), should be what we need for this PR.

evnchn:
Before:

image

Plain #5463

image

<details>
<summary>Testing code</summary>

```py
with ui.stepper().props('vertical').classes('w-full') as stepper:
    with ui.step('step'):
        ui.label('step label')
        with ui.stepper_navigation():
            ui.button('btn1', on_click=stepper.next)
            ui.button('btn2', on_click=stepper.next) # problem lies here
```
</details>

Note the distance between btn1 and btn2 is significantly wider than before.

nicegui/static/nicegui.css
falkoschindler: Good catch! We increase the gap between nav buttons from 8px to 1rem (the default gap).

If I remember correctly, I thought that this more what you'd expect from a NiceGUI element. But since we're integrating Quasar's QStepper we should stick to the original look, especially if we'd break existing layout otherwise.

I'll update the PR to revert it back to 8px.

nicegui/static/nicegui.css
falkoschindler: Usually there is a 24px gap around `.q-stepper__step-inner`. But the way we put nav buttons inside `.q-stepper__step-inner`, the space between step content and nav buttons is defined by NiceGUI's default gap of 1rem=16px. Therefore we're adding 8px top padding to the nav buttons to end up with 24px of free space.

falkoschindler:
I don't think we need to test on nicegui.io. Testing with main.py should be sufficient: As far as I can tell, everything looks correct.

nicegui/static/nicegui.css
evnchn: I had no issues with `padding-top: 8px;` seeing that it is identical in the end, but this was just to ask if `gap: 8px` should go here, which it did with https://github.com/zauberzeug/nicegui/pull/5463/commits/50894c722f81c981c0af6aac34d3e51007c6d1d1

evnchn:
On wanting to "increase the gap between nav buttons from 8px to 1rem (the default gap)", retrospect I am in agreement to it, but perhaps leave it to another PR since it can be considered breaking. Though may not have to be stashed all the way to 4.0 since actually it is good for visual coherence.

PR #5464, opened 2025-11-18
Experiment with a minimum implementation for `ui.sortable`

falkoschindler:
### Motivation

In PR #4656 we're still discussing the ideal API for a sortable UI element. Because this PR is already in a later stage, it becomes tricky to play around with alternatives.

### Implementation

This PR only introduces the SortableJS dependency and sets up a minimal `ui.sortable`.

Some code for experimentation:
```py
ui.label.default_classes('border border-gray-200 p-2 cursor-grab')

def handle_change():
    print('New order: ', [child.text for child in container])

with ui.sortable(on_change=handle_change) as container:
    ui.label('Label 1')
    ui.label('Label 2')
    ui.label('Label 3')
```

### Progress


phifuh:
> Thanks, @phifuh! So you're taking care of the mapping between HTML IDs and your data entries yourself.
>
> I just created PR https://github.com/zauberzeug/nicegui/pull/5464 to go back a few steps and think about what we really need. > My ui.sortable calls a change handler without any arguments. This seems not very useful at first. But since you need to lookup  > corresponding data anyway, you can simply inspect the child elements of the sortable container and use whatever attribute serves your needs. It can be a label text, the element ID or even a marker set via .mark().
>
> I'd love to hear your thoughts over there. Maybe we can use this as a starting point to start with a minimal feature set and see what we need later.

I am not sure if it would help to look at our implementation to maybe get a little bit inspired? (it´s messy tho)

### Proposal: Define Core APIs and Improve NiceGUI’s Sortable Integration

We have been experimenting with different approaches for managing data within sortable components for our nested-sortable system (currently around 900 lines). We totally agree that creating api´s for the most common usecases is the way to go. Ideally, sortable should handle most of the heavy lifting so that NiceGUI users don’t have to implement their own state-management layer.

#### Motivation

In practice, almost every sortable element corresponds to some form of underlying data. A sortable without associated data is technically possible, but uncommon. I would assume that most users typically need to:

- Retrieve data for a given sortable element
- Update data
- Insert new elements
- Remove elements
- Reflect drag-and-drop actions directly in their data structures

Earlier in our project, we tried attaching data directly to elements and retrieving it via element.filter (kinda like u suposed here - >

> Alternatively the user could attach value information to the child elements. If the event handlers would receive elements instead of element IDs, the user could access the sorted values more easily. But linters/type checkers will complain about adding attributes like 
>
> ```py
> with ui.sortable({...}, on_change=lambda e: ...):
>     item1 = ui.label('Item 1')
>     item1.value = 1
>     item2 = ui.label('Item 2')
>     item2.value = 2
>     item3 = ui.label('Item 3')
>     item3.value = 3
> ```

However, this approach broke because reordering didn’t properly synchronize with the underlying widget tree (But I am pretty sure EmberLightVFXThis fixed it in a later commit). 
We later implemented a “secondary list” strategy, which works but introduces a lot of complexity—especially with nested sortables.

To prevent every user from reinventing this logic, we propose a minimal but solid set of sortable APIs.

### Suggested Core APIs (MVP)

A first version of sortable should provide:

- get_all_data()   # Returns the complete list of data items in their current sorted order.
- get_data(element_or_id)  # Returns the data associated with a specific sortable element.
- insert_at_index(position, data)  # Inserts a new data item at the given position.
- delete(element_or_id)  # Removes a sortable element and its associated data.
- update(element_or_id, data)  # Updates the data associated with an element.

Why these are enough for MVP

These operations allow users to maintain a stable 1:1 relationship between UI elements and application data. This alone will solve 80–90% of use cases, including simple lists, todo apps, settings pages, and dynamic forms.

Once the MVP is solid, more advanced features could be added:

- Nested sortable support  #All core APIs should work recursively with nested structures.
- Official dropzones for nested items (tree-like structures).
- undo_last_move()
- get_max_depth()
- get_path(element_or_id)

These would greatly simplify complex use cases such as editors, page builders, menus, and nested data structures.

I am not sure if this makes sense? I am looking forward for your response :)

falkoschindler:
I'm closing this PR in favor of PR #5855.

PR #5465, opened 2025-11-18
Write all union types as X | Y (PEP 604)

evnchn:
### Motivation

Ref: https://github.com/zauberzeug/nicegui/pull/5380#issuecomment-3548069208

- I don't think we should mix new and old code. 
- And I wanted to see how big the diff is

### Implementation

Add `from __future__ import annotations` such that Py3.9 pylint does not throw up. 

Then we can use the new syntax. 

We end up with a huge diff, really. 

### Progress

### Final notes

Not sure whether this is for 3.x or 4.0.

### DIFF
--- a/examples/audio_recorder/audio_recorder.py
+++ b/examples/audio_recorder/audio_recorder.py
@@ -1,12 +1,14 @@
+from __future__ import annotations
+
 import base64
-from typing import Callable, Optional
+from typing import Callable
 
 from nicegui import events, ui
 
 
 class AudioRecorder(ui.element, component='audio_recorder.vue'):
 
-    def __init__(self, *, on_audio_ready: Optional[Callable] = None) -> None:
+    def __init__(self, *, on_audio_ready: Callable | None = None) -> None:
         super().__init__()
         self.recording = b''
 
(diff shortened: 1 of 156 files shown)


falkoschindler:
I'm planning to do that when we drop Python 3.9 (not sure when exactly) using [pyupgrade](https://github.com/asottile/pyupgrade).

evnchn:
Hmm but this (hopefully) still works with Py3.9, but what you imply means we're dropping Py3.9

Can we possibly run pyupgrade in-place to Py3.9 (or some other command line parameter) to do what I am doing right now?

falkoschindler:
Maybe. But what's the point? We'll update oldschool annotations in a few months anyway.

Regarding your motiviation:

- "I don't think we should mix new and old code." - I agree. But we don't need to rush it.
- "And I wanted to see how big the diff is." - You did. It's huge.

evnchn:
While reviewing the Ruff notes, the following would indicate that such changes are more than surface-level and may have actual impacts: 

> Fix safety
> 
> This rule's fix is marked as unsafe, as it may lead to runtime errors when alongside libraries that rely on runtime type annotations, like Pydantic, on Python versions prior to Python 3.10. 

Two questions arise: 

- Will this code break things? 
- Did any existing code inside the codebase already broke things, but we perhaps didn't test it enough?

And this PR is draft in nature because I think it's still an open question. Let's discuss, but I don't think priority for this is particularly high.

evnchn:
I think let's keep the existing solution of "if it ain't broke don't fix it" and do not try and move to new union syntax in Py3.9. 

The PR's intention was to "save time when going to NiceGUI 4.0 which supports new union syntax natively" but I don't think we are saving time with this one.

PR #5467, opened 2025-11-18
Provide proper default value to ui.range

Mick235711:
### Motivation

Currently, `ui.range` is completely unusable if you don't provide a `value`, despite its calling signature suggests that the `value` parameter is optional. For example, given the following code:
```python3
from nicegui import ui

ui.range(min=0, max=100)
ui.range(min=0, max=100, value={"min": 0, "max": 100})
ui.run()
```
Only one range slider will be shown, which seems to be confusing for beginners. Worse, there are no warnings/errors printed to the console, and the HTML is just blank for the first element.

### Implementation

Provide a proper default value to the `value` parameter, initializing to the full range. Alternatively, we can also move `value` to be a required parameter (remove `= None`) to accurately reflect its nature.

### Progress

- [x] Pytests are added. 

### DIFF
--- a/nicegui/elements/range.py
+++ b/nicegui/elements/range.py
@@ -21,7 +21,7 @@ class Range(ValueElement, DisableableElement):
         :param min: lower bound of the range
         :param max: upper bound of the range
         :param step: step size
-        :param value: initial value to set min and max position of the range (default: `min` to `max`)
+        :param value: initial value to set min and max position of the range (default: ``min`` to ``max``)
         :param on_change: callback which is invoked when the user releases the range
         """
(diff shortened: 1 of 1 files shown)


evnchn:
Hello @Mick235711 thanks for the PR. 

It's a fair fix. And retrospect we're actually kinda sloppy in terms of the types passed into Quasar across the library, passing `None`s despite the intended type is `str`, for example. (NiceGUI wishlist, maybe?)

image

Generally fine for `str` but in the case of `ui.range` it broke things because it expects a `dict`. 

@codingpaula As the initial author for `ui.range`, do read this in case I missed anything. Thanks!

PR #5469, opened 2025-11-19
Forward `--clean` and `--noconfirm` to pyinstaller in nicegui-pack

himbeles:
### Motivation

Currently, `nicegui-pack` attempts to delete the `build` and `dist` folders before a new build with PyInstaller. 
However, the implementation 
```python
for directory in ['build', 'dist']:
    if Path(directory).exists():
        Path(directory).rmdir()
```
fails to delete these folders with an error, and cancels the build, if there is any content of a previous build. This is because `Path.rmdir()` only works if directory is not empty. 

I think this is not the intended behaviour. The user must manually delete these folders after every build. And deleting these folders if they are already empty has no effect on the build.

This PR instead introduces the optional arguments `--clean` and `--noconfirm` for `nicegui-pack`, which are directly forwarded to PyInstaller. 
- `--clean`: Clean PyInstaller cache (build folder) and remove temporary files before building.
- `--noconfirm`:  Replace output directory (default: SPECPATH/dist/SPECNAME) without asking for confirmation

If arguments are **not** given, no cleanup is performed, which allows the user to rely on pyinstaller caching. 
This is currently prevented because one would run into the error raised from `rmdir()` before reaching PyInstaller.

### Implementation

Two more args for nicegui-pack have been introduced, directly forwarded to pyinstaller.
CLI documentation and logic are borrowed directly from PyInstaller, similar to most other arguments. 

### Progress

- [x] Pytests have been added (or are not necessary) -> currently there are no tests at all for `nicegui-pack`.

### DIFF
--- a/nicegui/scripts/pack.py
+++ b/nicegui/scripts/pack.py
@@ -53,7 +53,7 @@ def main() -> None:
         'Clean PyInstaller cache (in ./build folder) and remove temporary files before building.'
     ))
     parser.add_argument('--noconfirm', action='store_true', default=False, help=(
-        'Replace output directory (./dist/SPECNAME) without asking for confirmation'
+        'Replace output directory (./dist/SPECNAME) without asking for confirmation.'
     ))
     parser.add_argument('main', default='main.py', help='Main file which calls `ui.run()`.')
     args = parser.parse_args()
--- a/website/documentation/content/section_configuration_deployment.py
+++ b/website/documentation/content/section_configuration_deployment.py
@@ -319,7 +319,6 @@ doc.text('', '''
 
     - Specifying `--noconfirm` to `nicegui-pack` will replace the output directory (`./dist/SPECNAME`) without asking for confirmation.
 
-
     - Summary of user experience for different options:
 
         | `nicegui-pack`           | `ui.run(...)`  | Explanation |


evnchn:
These changes makes sense. 

---

Progressively, I have 3 bonus items, not blocking this PR. 

1. Are there any arguments which `nicegui-pack` should forward?

2. If `nicegui-pack`'s responsibilty becomes simply to forward as much arguments to pyinstaller, then I wonder how long should we maintain `nicegui-pack`? (ref: https://github.com/zauberzeug/nicegui/discussions/5377)

3. Is it fine if it will be gone in, say, NiceGUI 4.0?

himbeles:
> 1. Are there any arguments which `nicegui-pack` should forward?

This is an architectural choice for the maintainers.
- The more arguments, the more complicated of an API you need to maintain (document/test/etc.), but more flexibility for the user. 
- If the arguments are too tightly coupled to PyInstaller, it becomes harder to migrate to a different packaging backend without completely swapping out the API. 

> 2. If `nicegui-pack`'s responsibilty becomes simply to forward as much arguments to pyinstaller, then I wonder how long should we maintain `nicegui-pack`? (ref: [PyInstaller Hooks #5377](https://github.com/zauberzeug/nicegui/discussions/5377))

Indeed, I had initiated this discussion based on the hook I have contributed to PyInstaller. 
In the future (~next month), you will be able to just use PyInstaller directly with NiceGUI. 
Then it becomes a question of backwards compatibility for older versions of PyInstaller.
Also, a wrapper might still be useful for future evolution of the packaging backend.

falkoschindler:
To be honest: I'd love to drop nicegui-pack in the future. Its main purpose was to turn all packaging information and best practices, which we collected over the years on GitHub and in our documentation, into an executable tool. But maintaining it is additional work. And we at Zauberzeug rarely use it, so we don't have much experiences with all its nuances.

If packaging with a new version of PyInstaller "just works", removing nicegui-pack in 4.0 sounds like a reasonable plan. We could start the transition by deprecating it and see if our users have trouble migrating or not.

Anyway, let's merge this PR. 🚀

PR #5476, opened 2025-11-20
Introduce sentinels for default parameter values

falkoschindler:
### Motivation

Inspired by #4857 and #5424, this PR attempts to implement idea 5: a very readable sentinel for marking default parameter values that might be overwritten by default props/classes/style:
```py
color: Optional[str] = DEFAULT_PROP or 'primary',  # type: ignore[assignment]
```

### Implementation

**Utilities**

helpers.py
- A new `_DefaultSentinel` class overwrites the `__or__` method so that an instance can be or-ed with a value and still return the sentinel.
- Three instances `DEFAULT_PROP`, `DEFAULT_CLASS`, and `DEFAULT_STYLE` can be used to mark the origin on a default parameter value.

props.py
- A new `set` method takes care of assigning a new prop value, but automatically ignoring the sentinel and handling `None`.

**Examples**

tabs.py
- As a simple example, `TabPanels` now uses the `DEFAULT_PROP` sentinel and the `set` method for `animated` and `keep_alive`.

row.py and column.py
- A bit more complicated is the case of `wrap` and `align_items` for rows and columns. But the sentinel allows a very clear distinction between and explicit value passed to the initializer or the default which should be ignored.

button.py, icon\_element.py, and color\_element.py
- The button is an example for arguments being passed to mixins.
- The implementation for the icon element is a bit simpler now, because the new `set` method takes care of `None` values. The large diff is mainly caused by using PEP 604 type annotations to keep the `icon` signature tidy.
- The color element also uses PEP 604 type annotations and the `set` method. It now also handles default values more correctly and pops the default prop if a class or style is given.

### Progress

### DIFF
--- a/nicegui/elements/mixins/color_elements.py
+++ b/nicegui/elements/mixins/color_elements.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from typing import Any
 
 from ...element import Element
-from ...helpers import DEFAULT_PROP
+from ...helpers import DEFAULT_PROP, _DefaultSentinel
 
 QUASAR_COLORS = {'primary', 'secondary', 'accent', 'dark', 'positive', 'negative', 'info', 'warning'}
 for color in ['red', 'pink', 'purple', 'deep-purple', 'indigo', 'blue', 'light-blue', 'cyan', 'teal', 'green',
@@ -23,7 +23,7 @@ TAILWIND_COLORS = {
 class BackgroundColorElement(Element):
     BACKGROUND_COLOR_PROP = 'color'
 
-    def __init__(self, *, background_color: str | None | DEFAULT_PROP, **kwargs: Any) -> None:
+    def __init__(self, *, background_color: str | None | _DefaultSentinel, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         if background_color in QUASAR_COLORS or background_color is DEFAULT_PROP or background_color is None:
             self._props.set(self.BACKGROUND_COLOR_PROP, background_color)
@@ -38,7 +38,7 @@ class BackgroundColorElement(Element):
 class TextColorElement(Element):
     TEXT_COLOR_PROP = 'color'
 
-    def __init__(self, *, text_color: str | None | DEFAULT_PROP, **kwargs: Any) -> None:
+    def __init__(self, *, text_color: str | None | _DefaultSentinel, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         if text_color in QUASAR_COLORS or text_color is DEFAULT_PROP or text_color is None:
             self._props[self.TEXT_COLOR_PROP] = text_color
--- a/nicegui/elements/mixins/icon_element.py
+++ b/nicegui/elements/mixins/icon_element.py
@@ -6,14 +6,14 @@ from typing_extensions import Self
 
 from ...binding import BindableProperty, bind, bind_from, bind_to
 from ...element import Element
-from ...helpers import DEFAULT_PROP
+from ...helpers import _DefaultSentinel
 
 
 class IconElement(Element):
     icon = BindableProperty(
         on_change=lambda sender, icon: cast(Self, sender)._handle_icon_change(icon))  # pylint: disable=protected-access
 
-    def __init__(self, *, icon: str | None | DEFAULT_PROP, **kwargs: Any) -> None:
+    def __init__(self, *, icon: str | None | _DefaultSentinel, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self.icon = icon
         self._props.set('icon', icon)
--- a/nicegui/elements/row.py
+++ b/nicegui/elements/row.py
@@ -9,7 +9,7 @@ from ..helpers import DEFAULT_CLASS, DEFAULT_STYLE
 class Row(Element, default_classes='nicegui-row'):
 
     def __init__(self, *,
-                 wrap: bool = DEFAULT_STYLE or True,
+                 wrap: bool = DEFAULT_STYLE or True,  # type: ignore[assignment]
                  align_items: Literal[
                      'start',
                      'end',


evnchn:
_Sorry for the misclick. My iPhone 12's touchscreen may be a bit wonky. Time for replacement?_ 

#### Sentinels are good and https://github.com/zauberzeug/nicegui/issues/4857#issuecomment-3552692421 IMO undersold it. 

Seeing in NiceGUI library code context, the sentinel is a good idea: In status quo or #5424, it is impossible to know whether a parameter has origins from props/class/style unless you know the inner workings, while this PR shows it out. 

#### Naming enhancements? Non-default props also take precedence under this approach?

I think `DEFAULT_PROP or 'primary'` was coined because, in human words, "**follow** the default props, if not then primary". However, I also think that non-default (ad-hoc applied using `.props`) would also take precedence, and **follow** is the keyword of the sentence, so maybe `FOLLOW_PROPS or 'primary'` is better?

#### The real decision point: Behaviour change?

Note that: 

- In #5424 and status quo, NiceGUI wins, just that the PR shows a warning to let you know. 
- In this PR, user's props wins. 

This would be classifiable as a breaking change in behaviour and may have to be delayed to NiceGUI 4.0. Combined with my argument that we should actually let NiceGUI win, I'm not too sure on this. 

But this is just library design. Code-wise to change this PR to let NiceGUI win, it would just be a one-line change in `props.set`. The sentinel's still excellent.

falkoschindler:
### Naming

I don't think the sentinel applies to "normal" props. How could you set such props _before_ calling the initializer? I think there can only be default props at the time of initialization.

Oh, you mean overwriting a parameter later? `ui.button(color='primary').props(color='secondary')`? I would argue that we don't need to communicate that a prop might change the look and feel of an existing element. Instead the sentinel is meant to resolve the ambiguity of a color parameter being "primary" and NiceGUI doesn't know if the user is trying to overrule a default prop with a parameter or if the user didn't pass any color parameter and this is just the default value.

Anyway, I still like `DEFAULT_PROP` more. Because at the time of calling the initializer, the default value of this parameter is _the value of a corresponding default prop or "primary"_.

### Breaking change?

Technically this can be considered a breaking change. But I'd argue this fixes bugs, because after setting a default prop `color=red-5` and then calling `ui.button` without any color, I would expect the button to be red. The fact that it ignored the default prop can be considered a bug. (This is why #4857 is labeled as bug.)

evnchn:
**Agreed let's do `DEFAULT_PROP`**. While `FOLLOW_PROPS` better highlight the relationship between the `color` in `ui.button(color='primary').props(color='secondary')` and was why I proposed it, `DEFAULT_PROP` might just be good enough to imply it, and has a stronger presence in highlighting that its introduction was to resolve the default props overwriting problem. 

**Merging this will likely necessitate a warning emoji** (not typed here to not mess up the GitHub searches) since the behaviour is fundamentally different. After all, people may accidentally set props, which while ineffective in past versions, suddenly become effective. 

Citing how "[this subject keeps popping up](https://github.com/zauberzeug/nicegui/discussions/4856)", we see a strong user demand for default_props to take precedence. At the end of the day the library's here to empower users, not to fulfill SemVer requirements, so maybe we can have this in 3.4 after all? 

We can also argue that #4857 was positioned as a bug since Jun 11 and 6 months is _more than enough time_ to voice-out it was not a bug if you do think so that way.

nicegui/helpers.py
```diff
@@ -28,6 +30,16 @@
     from inspect import iscoroutinefunction
 
 
+class _DefaultSentinel:
+    def __or__(self, other: int) -> Self:
+        return self
```
evnchn, line 35: The value of `other` is totally lost this way? `DEFAULT_PROP or 'foo'` and `DEFAULT_PROP or 'bar'` is indistinguishable...

nicegui/props.py
```diff
@@ -138,3 +138,12 @@ def parse(text: Optional[str]) -> dict[str, Any]:
                     props[key] = True
 
         return props
+
+    def set(self, key: str, value: Any) -> None:
+        """Set a prop value unless it is ``DEFAULT_PROPS``; remove it if the value is ``None``."""
+        if value is helpers.DEFAULT_PROP:
+            return
```
evnchn, line 145: Knock-on effect of "value of other is totally lost in `__or__`": When the `set()` sees a `DEFAULT_PROP` it just gives up and does nothing. Would be a regression.

nicegui/props.py
```diff
@@ -138,3 +138,12 @@ def parse(text: Optional[str]) -> dict[str, Any]:
                     props[key] = True
 
         return props
+
+    def set(self, key: str, value: Any) -> None:
+        """Set a prop value unless it is ``DEFAULT_PROPS``; remove it if the value is ``None``."""
```
evnchn, line 143: The rest of the code had `DEFAULT_PROP`? I kinda think `DEFAULT_PROPS` is more correct since we have `.props`

nicegui/elements/button.py
```diff
@@ -4,19 +4,20 @@
...
 from ..events import ClickEventArguments, Handler, handle_event
+from ..helpers import DEFAULT_PROP
 from .mixins.color_elements import BackgroundColorElement
 from .mixins.disableable_element import DisableableElement
 from .mixins.icon_element import IconElement
 from .mixins.text_element import TextElement
 
 
-class Button(IconElement, TextElement, DisableableElement, BackgroundColorElement):
+class Button(IconElement, TextElement, DisableableElement, BackgroundColorElement, default_props='color=primary'):
```
evnchn, line 14: Ah... So the real `'primary'` was _here_, and the `DEFAULT_PROP or 'primary'` was just "for decoration" and actually idempotent to `DEFAULT_PROP`.

Kinda cursed not gonna lie. 

And the concern for passing `DEFAULT_PROP` around a codebase which originally expects strs and ints still hold.

falkoschindler: > Ah... So the real `'primary'` was _here_, and the `DEFAULT_PROP or 'primary'` was just "for decoration" and actually idempotent to `DEFAULT_PROP`.
>
> Kinda cursed not gonna lie.

Yes, that's the extra mile NiceGUI is trying to go to achieve a nice API for the user. But one could argue that it's not worse than keeping a docstring up-to-date. If you change the default prop, you need to update the documentation including the signature. (This basically never happens as it would be a breaking change.)

> And the concern for passing `DEFAULT_PROP` around a codebase which originally expects strs and ints still hold.

I think we can't have both: We either want to have the information if the value is supposed the default, or not. But I don't think it is being passed around very often. The color mixins are a rare example, I assume. But we'll only know for sure once we applied the new concept to the remaining code base.

nicegui/props.py
falkoschindler: Ah yes, I changed it from `DEFAULT_PROPS` to `DEFAULT_PROP` and forgot to update the docstring. I thought `DEFAULT_PROP` is more correct because the value is either defined as a "default prop" or the mentioned value.

nicegui/elements/button.py
evnchn: `DEFAULT_PROP` being memory-free means we always have 2 code paths, one which user passes in stuff or else it is `DEFAULT_PROP`, another being the library default value. This may or may not scale well. 

I did think about sentinel with memory which `unwrap(x)` returns `x.value` if it is the sentinel, if not return `x` straight away. But did not go far with it because I thought #5492 was the endgame (until it isn't)

nicegui/helpers.py
```diff
@@ -28,6 +30,16 @@
     from inspect import iscoroutinefunction
 
 
+class _DefaultSentinel:
+    def __or__(self, other: int) -> Self:
```
evnchn, line 34: Just realized: this is for `DEFAULT | 'blah'` and `DEFAULT or 'blah'` doesn't even trigger this (`DEFAULT` is truthy)

falkoschindler: Oh, I wasn't aware of that! So it still works for "memory-free" default annotation, but not if you'd want to store the value somewhere.

falkoschindler:
I'm closing this PR in favor of PR #5505.

Issue #5472, opened 2025-11-20
add_head_html causes RuntimeError without shared=True


davetapley:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

ui.add_head_html('''
    <style>
        .my-red-label {
            color: Crimson;
            font-weight: bold;
        }
    </style>
''')


@ui.page('/')
async def root():
    ui.label('RED').classes('my-red-label')

ui.run()
```

### Description

Per [the docs](https://nicegui.io/documentation/section_pages_routing#add_html_to_the_page), while implementing a workaround for https://github.com/zauberzeug/nicegui/discussions/5471

I ran into:
```
RuntimeError: ui.page cannot be used in NiceGUI scripts where you define UI in the global scope. To use multiple pages, either move all UI into page functions or use ui.sub_pages.
```

The solution is to pass `shared=True` to `add_head_html`, but this isn't immediately obvious.

Can with improve the error message, or since it's a `RuntimeError` just assume `shared=True` in this case?

### NiceGUI Version

3.2.0

### Python Version

Python 3.12.2

### Browser

Other

### Operating System

Linux

evnchn:
Hi @davetapley thanks for the report. 

Kinda intended behaviour? But I get why it is confusing. 

https://github.com/zauberzeug/nicegui/blob/54c00a60fc25efd4b59627409567dad7d12a8e45/nicegui/functions/html.py#L5-L17

See, `ui.add_head_html` has a dual-mode behaviour:

- When you do `shared=False` since it accesses the `client` it will invoke script mode for you (intended behaviour for any UI element), but which you don't want in this instance. 
- When you do `shared=True`, it "betrays" how it is belong to `ui` and actually just adds the head html straight away for you, and acts more like like `app.add_head_html` (where we put non-UI elements)

I have sketched out a PR to make `ui.add_head_html` do NOT raise error by having it auto-select which mode to operate in, but I think it just muddies water further. 

I don't think anyone will like it but I can show you in a draft PR (gimme 10 mins). Maybe it's salvageable by some wiser person. 

Finally: @falkoschindler What do you think if we had `app.add_head_html` in NiceGUI 4.0 to cut the dual-mode?

evnchn:
[...]

PR #5478, opened 2025-11-20
Warn accidental script mode due only to `ui.add_*` without `shared=True`

evnchn:
### Motivation

As see in https://github.com/zauberzeug/nicegui/issues/5472#issue-3645073473, it is easy to accidentally invoke NiceGUI script mode due to use of `ui.add_head_html` or other `ui.add_*` functions without `shared=True`

```py
from nicegui import ui

ui.add_head_html('''
    <style>
        .my-red-label {
            color: Crimson;
            font-weight: bold;
        }
    </style>
''')

@ui.page('/')
async def root():
    ui.label('RED').classes('my-red-label')

ui.run()
```

Without knowledge in NiceGUI inner-workings, the following error message is quite a head-scratcher. 

```
RuntimeError: ui.page cannot be used in NiceGUI scripts where you define UI in the global scope. To use multiple pages, either move all UI into page functions or use ui.sub_pages.
```

### Implementation

This PR detects such accidental invocations of NiceGUI script mode by checking for both:

- No other elements in the script mode client's content
- Either `_body_html` or `_head_html` was set

And provides a message with action to resolve this issue. 

```
Note: You had no other UI elements in the global scope than ui.add_head_html / ui.add_body_html / ui.add_css. 
To resolve the RuntimeError below, consider using shared=True, as they do not constitute UI elements. 
```

### Progress

- [x] Pytests are not necessary (we don't test script mode).
- [x] Documentation is not necessary (self-explanatory).


falkoschindler (approved):
Great idea to use `core.script_client._head_html` and `core.script_client._body_html` in this way, @evnchn!

I just combined the warning and exception to a more targeted error message.

Ready to merge if you ask me. 🚀

evnchn:
Let's go! 🚀

PR #5481, opened 2025-11-21
Add default slot to the table component

### LINKED ISSUE #5447: ui.table tooltip method does not seem to do anything
### First Check

### Example Code

```python
from nicegui import ui

table = ui.table(rows=[
    {'make': 'Toyota', 'model': 'Celica', 'price': 35000},
    {'make': 'Ford', 'model': 'Mondeo', 'price': 32000},
    {'make': 'Porsche', 'model': 'Boxster', 'price': 72000},
])

table.tooltip("It's a Ferrari.")

ui.run()
```

### Description

Hi!

When trying different ways of adding a disclaimer to a table, i happily stumbled upon the tooltip method that is described by the documentation and also implemented for ui.table. However this does not seem to trigger anything, no matter where the cursor is placed i am not shown the tooltip.

I'm already sorry in case i misunderstood the functionality, but the reproduction of the issue is simple enough.

Thank you very much for all your work!

love

### NiceGUI Version

3.2.0 / 3.0.3 / 2.24.0

### Python Version

3.13.0 / 3.13.0 / 3.10.11

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

himbeles:
### Motivation

Fixes #5447: enables missing tooltip for the `ui.table` component. 

### Implementation

Wrapped the table component template in a neutral container (div) with a default slot, forwarding all `attrs` to the inner q-table and keeping named slots intact, while exposing a default slot so `table.tooltip(...)` can mount.
Filtered slot forwarding to skip the default slot when passing through to `q-table`, preventing tooltips from being swallowed.

### Progress

- [x] Pytests have been added (or are not necessary). -> can't test mouse hover

### DIFF
--- a/nicegui/elements/table.js
+++ b/nicegui/elements/table.js
@@ -20,9 +20,5 @@ export default {
       this.columns.forEach((column) => convertDynamicProperties(column, false));
       return this.columns;
     },
-    forwardedSlots() {
-      const { default: _default, ...nonDefaultSlots } = this.$slots;
-      return nonDefaultSlots;
-    },
   },
 };


himbeles:
Same approach could be taken for `ui.uploader` ...

nicegui/elements/table.js
```diff
@@ -1,12 +1,16 @@
...
+  inheritAttrs: false,
   template: `
-    <q-table ref="qRef" :columns="convertedColumns">
-      <template v-for="(_, slot) in $slots" v-slot:[slot]="slotProps">
-        <slot :name="slot" v-bind="slotProps || {}" />
-      </template>
-    </q-table>
+    <div>
+      <q-table ref="qRef" v-bind="$attrs" :columns="convertedColumns">
+        <template v-for="(_, slot) in forwardedSlots" v-slot:[slot]="slotProps">
```
falkoschindler: What to you think about this alternative without an extra computed variable?
```suggestion
        <template v-for="(_, slot) in $slots" v-if="slot !== 'default'" v-slot:[slot]="slotProps">
```

himbeles: sounds good!

himbeles:
There seems to be an issue with passing properties to the table.

falkoschindler:
Very nice, @himbeles! This looks like a great solution to fix tooltips for `ui.table` (and possible other elements like `ui.upload`).

Let's check if there are other elements that can be fixed the same way. As far as I can tell, there are `ui.table`, `ui.upload`, `ui.editor`, and `ui.dialog`.

Regarding the failing test: I guess the test needs an update. Passing props to the table seems to work though.

nicegui/elements/table.js
```diff
@@ -16,5 +20,9 @@ export default {
       this.columns.forEach((column) => convertDynamicProperties(column, false));
       return this.columns;
     },
+    forwardedSlots() {
+      const { default: _default, ...nonDefaultSlots } = this.$slots;
+      return nonDefaultSlots;
+    },
```
evnchn:
```suggestion
```

It has to go because unused?

himbeles: Sure. Sorry...

evnchn:
Sorry, I may be a bit vested-interest here, but I wonder will this PR affect user's CSS / Tailwind selectors, as the DOM structure becomes slightly different (though no visual changes)

falkoschindler:
@evnchn This is very well possible, even though rather unlikely. I thought fixing tooltips for `ui.table` would justify the DOM change. But it's not perfect.

himbeles:
You are right @evnchn. This might not be the optimal solution. Also, I am introducing a default slot for an element which perhaps should not have one. For example, one could now open the table as a context and UI in this context would be rendered as a second element of the new div. This is not necessarily bad, but definitely a choice...

evnchn:
"open the table as a context" is a unintended behaviour of this PR, and we should not condone. 

Should we go with this route we will need active measures to ensure **only** tooltip gets through for `ui.table`, nothing else (raise an error?)

With this, #5482 looks even more appealing though...

himbeles:
I agree. Feel free to close this one in favor of yours.

evnchn:
Sure no problem. But perhaps please don't delete the branch so quick just in case #5482 doesn't work out.

PR #5420, opened 2025-11-07
Add HTML IDs to complex Quasar elements

falkoschindler:
### Motivation

In #5417 we noticed that some UI elements still don't get an ID in the HTML DOM.

### Implementation

This PR uses a new technique by setting the "for" prop in Python. This saves us dedicated JS components for all these elements. The `ui.input` element, which already fixed the missing ID with an ":for" attribute in JS, has been updated for consistency.

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests have been added.
- [x] Documentation is not necessary.

PR #5482, opened 2025-11-21
Manually attach target for `tooltip` method (works in more cases)

evnchn:
### Motivation

Fixes https://github.com/zauberzeug/nicegui/issues/5447 and so much more cases where the mechanism to place child elements is not enabled by the element. 

**Real motivation: I thought the approach in https://github.com/zauberzeug/nicegui/pull/5481 isn't scalable...**

### Results

Even `ui.markdown`, which populates `innerHTML` directly (so all child are long gone) works. 

```py
from nicegui import ui

ui.markdown('''
Try _tooltip_!
''').tooltip('Works!')

ui.run(show=False)
```

image

### Implementation

Attach the targets ourselves using HTML ID instead of defaulting to using parent: https://quasar.dev/vue-components/tooltip/#qtooltip-api

### Progress

- [x] Pytests are added: https://github.com/zauberzeug/nicegui/pull/5482/commits/1dea497f8397b27f51b00ca8f8349d429651862f
- [x] Documentation is not necessary (enhances existing functionality).

### DIFF
--- a/website/documentation/content/tooltip_documentation.py
+++ b/website/documentation/content/tooltip_documentation.py
@@ -37,23 +37,4 @@ def tooltip_with_other_content():
             ui.image('https://picsum.photos/id/377/640/360').classes('w-64')
 
 
-@doc.demo('Tooltip on HTML and Markdown', '''
-    Some elements like `ui.html` and `ui.markdown` do not support nested elements.
-    In this case, you can nest such elements inside a container element with a tooltip.
-''')
-def tooltip_on_html_and_markdown():
-    with ui.element().tooltip('...with a tooltip!'):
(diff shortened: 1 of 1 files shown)


falkoschindler:
🤯

evnchn:
Question for @falkoschindler: Did you get everything for https://github.com/zauberzeug/nicegui/pull/5420 (adding HTML ID)? 

Because if not then this may regress and break some elements which it used to work. Thanks!

falkoschindler:
> Did you get everything for #5420 (adding HTML ID)?

I do think so, yes. As far as I remember I checked all Quasar elements and collected the tricky ones for a pytest. All others should be fine be design (nice rhyme...).

PR #5483, opened 2025-11-21
Unify dark mode API and add color-scheme meta tag for browser/extension compatibility

evnchn:
### Motivation

There are 2 problems this PR want to address:

1. #5471, where select browsers and Dark Reader extension do not cope well with NiceGUI's dark mode because we did not use `<meta>` tags to inform them
2. `ui.dark_mode` is hard-coded to use `Quasar.Dark.set` which gets in the way of [using other UI frameworks than Quasar](https://nicegui.io/documentation/section_styling_appearance#using_other_vue_ui_frameworks) (#4859)

**Real motivation: to experiment, since I have doubts over "demo is good enough instead of API addition" https://github.com/zauberzeug/nicegui/discussions/5471#discussioncomment-15036953**

### Implementation

- `Quasar.Dark.set` will never be called directly: It will be wrapped to take true/false/undefined, then set as `darkSetter`
- `darkSetter` can be another UI library's dark mode function, if you need that. 
- `setDark` calls the `darkSetter` while also handles the `<meta>` tags for you. 
- To prevent FOUC, the `<meta>` tags are also populated in HTML.  

### Progress

- [x] Pytests are not necessary? (not sure...)
- [x] Documentation is not necessary? (not sure...)
- [x] Maybe let's use more unique id names to avoid collision with user's code?

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -80,7 +80,6 @@ function setDark(dark) {
   document
     .getElementById("nicegui-color-scheme")
     .setAttribute("content", dark === None ? "normal" : dark ? "dark" : "light");
-  document.getElementById("nicegui-darkreader-lock").setAttribute("name", dark ? "darkreader-lock" : "");
 }
 
 function parseElements(raw_elements) {
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -8,7 +8,6 @@
       name="color-scheme"
       content="{{ ({'True': 'dark', 'False': 'light'}).get(dark, 'normal') }}"
     />
-    <meta id="nicegui-darkreader-lock" name="{{'darkreader-lock' if dark=='True' else ''}}" />
     <link href="{{ favicon_url }}" rel="shortcut icon" />
     <style>
       @layer theme, base, quasar, nicegui, components, utilities, overrides, quasar_importants;
--- a/website/documentation/content/dark_mode_documentation.py
+++ b/website/documentation/content/dark_mode_documentation.py
@@ -36,4 +36,19 @@ def bind_to_switch() -> None:
     ))
 
 
+@doc.demo('Disable Dark Reader extension', '''
+    NiceGUI sets a `<meta name="color-scheme">` tag to inform browsers and extensions about your app's color scheme.
+    However, the [Dark Reader](https://darkreader.org/) browser extension may still transform your page.
+    To prevent this, you can add a `<meta name="darkreader-lock">` tag using `ui.add_head_html`.
+    Use `shared=True` to apply it to all pages, not just the current one.
+    This is an [officially supported mechanism](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#disabling-dark-reader-on-your-site).
+
+    Note: Users can still enable Dark Reader manually in their extension settings if they prefer it over your dark theme.
+''')
+def darkreader_lock() -> None:
+    # ui.add_head_html('<meta name="darkreader-lock">')
+
+    ui.label('Dark Reader is disabled on this page.')
+
+
 doc.reference(ui.dark_mode)


davetapley:
@evnchn well this was fun to test...

TLDR a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8 subtly changed the behavior to make my original repro from #5471 not work, so this PR is still _probably correct_, but in a way I can't verify (without cherry-picking fix back).

---

I _presume_ the Dark Reader extension has some built in semantics to guess if the page is 'dark', and the changes in a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8 make the page settle fast enough for it to automatically detect when `dark = True`, and so the extension doesn't try to inject its own CSS.

To further complicate things only a hard browser reload would causes the issue (regardless of a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8).

Here's hard / soft refresh before / after a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8:

https://github.com/user-attachments/assets/7ccf7cc9-084f-45cd-a61b-b073f294955b


So, with that said, LGTM 😁

evnchn:
> https://github.com/zauberzeug/nicegui/commit/a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8

Consider to change your calls to `add_css` to `add_head_html` and include the style tag by yourself. This way the CSS is in the head right away and is "slightly earlier"

davetapley:
@evnchn to be sure: I'm not doing any custom styling. But that is a good tip.

evnchn:
> not doing any custom styling
> https://github.com/zauberzeug/nicegui/commit/a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8 broke the repro

@davetapley Sorry but I can't see how the repro can be broken while you are not using custom styling (so no `add_css` let alone `add_scss`/`add_sass`)

If you really have me guess, then I have to guess that Dark Reader is reacting to the following body snippet:

```html
    <script>
      addStyle = (c) => document.head.append(Object.assign(document.createElement("style"), { textContent: c }));
    </script>
```

Which I am so shocked if that is the case, think should not be the case, and if that really is the case then bug report to Dark Reader (https://github.com/darkreader/darkreader) is highly recommended.

davetapley:
@evnchn hmm, I was using ⬇️  to repro (obvs with `dark: True`):
https://github.com/zauberzeug/nicegui/blob/b11972f063b976174f0cd4a4af4a9ead18ac2a10/examples/todo_list/main.py#L61

Not sure how to proceed.

falkoschindler:
@evnchn Since it's your PR: How should we proceed? Are you and @davetapley still testing, or should I review? Personally I'd love to hear that it actually serves a purpose in solving #5471 before spending time reviewing and possibly discussing about API details.

evnchn:
Let me clean up the conversation history. @davetapley kindly check 🙏  

Context: **Test page is `dark=True`, Dark Reader shall NOT invoke.**

1. How https://github.com/zauberzeug/nicegui/discussions/5471 is phrased, it would appear that before https://github.com/zauberzeug/nicegui/commit/a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8, Dark Reader is always invoked. If not I am sure he would note the inconsistency. 
2. https://github.com/zauberzeug/nicegui/pull/5483#issuecomment-3666960845 1/2: After https://github.com/zauberzeug/nicegui/commit/a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8, for some bespoke reason, actually it brought a positive impact, as Dark Reader is invoked only on first page visit, and subsequent visits do not invoke. NOTE: While "broke the _repro_" was stated, it did not break but rather mended _functionality_.
3. https://github.com/zauberzeug/nicegui/pull/5483#issuecomment-3666960845 2/2: After this very PR, Dark Reader is invoked never, which is the desired outcome, hence why "with that said, LGTM 😁" was replied. 

@falkoschindler TL-DR: https://github.com/zauberzeug/nicegui/commit/a8fd25b7d5e23afb1952d0f60a1940e18b5f1ca8 muddied waters by mysterious fixing the Dark Reader invocation except the first one, but the purpose of this PR still stands as it fixes it 100%. 

I think we can move straight to review and API design. Though you may want to wait for @davetapley and check out the other PRs (many) in the meantime.

falkoschindler:
Let's re-schedule this PR to 3.7. At the moment I see the following open questions:

- Should we add add prefix "nicegui-" to the IDs "color-scheme" and "darkreader-lock"? I do think so.
- Should we really add a "darkreader-lock" tag by default? As far as I understand, the user should be able to decide whether to use darkreader or not. It is generally not recommended to _force_ the user to disable darkreader. While a NiceGUI page in dark mode already looks pretty dark, it might not be the darkreader result the user is hoping for.

evnchn:
> Do we "darkreader-lock"?

I think you can still force-enable Dark Reader even with it. Considering that, I prefer that it is there, to preserve the "Nice" and works-out-of-the-box nature of the library.

falkoschindler:
I'm still unsure about the darkreader-lock:
Currently it is only set for `dark=True`. For `dark=None` it isn't set, so Dark Reader might be active depending on the system setting. Is this intentional? I wonder if we need to watch the system preference to toggle it accordingly.
On the other hand, why don't we _always_ use the lock? It won't hurt in light mode, right?

But it is still unclear to me whether we should use the darkreader-lock at all - or what the "nice" and prefered result is. There are arguments for both scenarios:

**Scenario 1 (prefer Dark Reader):**
- Dark Reader users are often power users who've customized their settings (contrast, brightness, sepia, etc.) and want consistency across all sites
- They installed an extension specifically to control dark mode - taking that control away could feel intrusive
- "I chose Dark Reader, not your dark theme"

**Scenario 2 (prefer native dark mode):**
- Dark Reader is often used as a fallback for sites *without* dark mode
- Native dark themes are usually higher quality (designed for the specific UI)
- Dark Reader can produce visual artifacts with complex UIs
- "Dark Reader is for sites that don't try - this site tries"

I think, if we introduce the darkreader-lock, we should allow the developer to opt in or out. We can provide it in form of a demo, a `ui.run` parameter, a `ui.dark_reader_lock` element, a `ui.dark_mode.dark_reader_lock` method or whatever. In the end it's just static head HTML.

Personally I slightly lean towards omitting the lock by default and show how to add it in form of a demo at https://nicegui.io/documentation/dark_mode: `ui.add_head_html('<meta name="darkreader-lock">')`. This is easier and potentially less confusing than adding the lock by default and trying to remove it on demand. In the future we can think about a more handy API and whether to enable the lock by default (maybe in 4.0?).

evnchn:
Note how there are equally valid arguments against both including `darkreader-lock` and never, and so `ui.dark_reader_lock` configurability sounds reasonable. 

But how about `<meta name="color-scheme">`? I guess we are keeping that?

falkoschindler:
Yes, I think we should keep `<meta name="color-scheme">`.

Meanwhile I noticed another problem with the darkreader-lock: If dark mode is `False`, we (as the NiceGUI maintainers) don't know
- if the site supports dark mode but it is currently deactivated, i.e. the visitor wants to see the light version, (→ lock!),
- if the site implements a single mode only with a specific look that shouldn't be messed around with (→ maybe lock?),
- or if the developer never thought about light and dark, and it would be good to have Dark Reader do its thing (→ no lock!).

Anyway, I'll remove the lock and add a demo explaining how to add it.

PR #5485, opened 2025-11-21
UnoCSS support (3.0)

evnchn:
### Motivation

Due to various reasons it may not be good to have ourselves tied down to using Tailwind CDN JS

- Large size
- Performance reasons
- Lack of configurability

So this PR want to enable UnoCSS support, an atomic and configurable CSS engine. 

### Implementation

Compared to predecessor #4832, this is more clean because:

- Dark mode works without sketchy hacks
- Code is cleaned up in some places (shorter name `allClasses`, which `generateStylesFromClasses` takes no argument, cleaner for loop, etc)
- No need preset-wind <-> preset-wind3 hack, because of https://github.com/unocss/unocss/releases/tag/v66.5.3 by yours truly
- Now that we have CSS layering, stuff becomes more organized and simple. 

Going over the changes briefly:

- Extract core libraries
- Expose option in `ui.run` `ui.run_with` and `AppConfig`
- Use UnoCSS in HTML and JS
- Ignore for precommit codespell

### Progress

- [x] Pytests have been added (https://github.com/zauberzeug/nicegui/pull/5485/commits/b0fb62a995ab27d5fe069bfeb144e74850454189).
- [x] Documentation has been added (https://github.com/zauberzeug/nicegui/pull/5485/commits/20e39169adb4dc27fe85e2a7c7fa9e0ed44e5526).

### DIFF
--- a/website/documentation/content/section_styling_appearance.py
+++ b/website/documentation/content/section_styling_appearance.py
@@ -183,7 +183,7 @@ def tailwind_layers():
     - "wind4": [UnoCSS Wind4 preset](https://unocss.dev/presets/wind4)
 
     UnoCSS is a smaller library and more performant, especially on small pages.
-    On "low-tier mobile" CPU throttling profile, load time for this page went from 1.1s down to 0.7s.
+    In Lighthouse Desktop, load time for this page went from 1.1s down to 0.7s.
 
     However, full compatibility with Tailwind CSS is not guaranteed.
     For example, Tailwind CSS Layers (see above) do not work with UnoCSS.


evnchn:
**https://github.com/zauberzeug/nicegui/pull/5485/commits/57a3b3370ee82f55be332ead07ca1ae2c23c4bc1 untested**

evnchn:
https://github.com/zauberzeug/nicegui/commit/57a3b3370ee82f55be332ead07ca1ae2c23c4bc1 tested and works. 

```js
// Modified from examples/custom_vue_element/counter.js to use Tailwind classes
export default {
  template: `
    <button
      @click="handle_click"
      :class="[
        'px-4 py-2 rounded-md transition-colors',
        value > 0 ? 'bg-green-200 hover:bg-green-300' : 'bg-gray-200 hover:bg-gray-300'
      ]"
    >
      <strong>{{ title }}: {{ value }}</strong>
    </button>
  `,
  props: {
    title: String,
  },
  data() {
    return {
      value: 0,
    };
  },
  methods: {
    handle_click() {
      this.value += 1;
      this.$emit("change", this.value);
    },
    reset() {
      this.value = 0;
    },
  },
};
```

nicegui/static/nicegui.js
```diff
@@ -413,6 +498,7 @@ function createApp(elements, options) {
               getElement(id)[element.update_method]();
             }
           }
+          addObservers(msg);
```
evnchn: This should be moved right after `await this.$nextTick();` to avoid edge case where the element added new class due to update method executing.

nicegui/static/nicegui.js
```diff
@@ -398,9 +478,14 @@ function createApp(elements, options) {
             .map(([_, element]) => loadDependencies(element, options.prefix, options.version));
           await Promise.all(loadPromises);
 
+          const originalClassesCount = allClasses.size;
+          for (const el of Object.values(msg)) for (const c of el?.class || []) allClasses.add(c);
+          if (allClasses.size > originalClassesCount) await generateStylesFromClasses();
```
evnchn: Do not run this if UnoCSS not enabled

nicegui/static/nicegui.js
```diff
@@ -308,11 +381,14 @@ const TAB_ID =
...
 window.onbeforeunload = function () {
   sessionStorage.__nicegui_tab_closed = "true";
+  dropAllObservers();
 };
 
 function createApp(elements, options) {
   Object.entries(elements).forEach(([_, element]) => replaceUndefinedAttributes(element));
   setInterval(() => ack(), 3000);
+  for (const el of Object.values(elements)) for (const c of el?.class || []) allClasses.add(c);
+  generateStylesFromClasses().then(() => document.getElementById("app").style.removeProperty("display"));
```
evnchn: Do not run this if UnoCSS not enabled

And possibly to use some other way of hiding the body (our own class?) other than straight-up manipulating the style, in case the user is hiding the body for some longer duration that what we are doing for some other reason.

evnchn (commented):
Spotted some mistakes while on the train

nicegui/static/nicegui.js
evnchn: https://github.com/zauberzeug/nicegui/pull/5485/commits/4a29052f7610eb28043737f6cff2966ab2a83663

nicegui/static/nicegui.js
evnchn: https://github.com/zauberzeug/nicegui/pull/5485/commits/4a29052f7610eb28043737f6cff2966ab2a83663

nicegui/static/nicegui.js
evnchn: https://github.com/zauberzeug/nicegui/pull/5485/commits/4a29052f7610eb28043737f6cff2966ab2a83663 and https://github.com/zauberzeug/nicegui/pull/5485/commits/9263cb0e6bf0177e9b8b2b88029cd88643bbf610

main.py
```diff
@@ -99,4 +99,5 @@ def _status():
 
 
 # NOTE: do not reload on fly.io (see https://github.com/zauberzeug/nicegui/discussions/1720#discussioncomment-7288741)
-ui.run(uvicorn_reload_includes='*.py, *.css, *.html', reload=not on_fly, reconnect_timeout=10.0)
+ui.run(uvicorn_reload_includes='*.py, *.css, *.html', reload=not on_fly,
+       reconnect_timeout=10.0, tailwind=False, unocss_preset='wind4')
```
evnchn: Accidentally left this in there. Oh well. Let's remove this before we merge (or not, since it works)

evnchn:
Actually is the observer needed? All of NiceGUI components do not add Tailwind classes programmatically... 

Perhaps configurable? `observer` can be one of 3 values:

- `None`: No observer, most performant
- `'components'`: Catches components which add Tailwind classes programmatically
- `'full'`: Observes the entire body and misses nothing

evnchn:
I really want to drop the observer because I feel that UnoCSS is for "advanced users" who know what they are doing...

falkoschindler:
Actually @rodja and I came to the conclusion that we'd like to keep the "full" observer for now so that UnoCSS becomes an optional drop-in replacement for Tailwind. Then we can start trying it out in production and see if there are any unforeseen problems. In 4.0 we could then think about making UnoCSS the default.

Independently we could make the observer customizable in a separate PR. But to keep this PR reasonably "small", I would choose full compatibility for now.

falkoschindler:
This is basically the last PR in 3.6. But it's rather big and we shouldn't rush it. Let's release 3.6 without it and finish UnoCSS as one of the first things we do for 3.7.

falkoschindler:
I finally started working on this PR again. Although I already proposed to

> keep the "full" observer for now

here is a more detailed analysis of our options. Now I'm pretty sure that option C, a single document observer, is indeed a great solution that reduces code complexity and eliminates unsupported edge cases, without losing the performance benefit of UnoCSS over TailwindCSS. I'll go on an try changing the implementation accordingly.

### Option A: No Observer

**Approach**: Only process classes from NiceGUI's WebSocket `update` messages.

**Pros**:
- Simplest implementation
- Zero runtime overhead
- Works for 99% of use cases (Python-driven class changes)

**Cons**:
- Classes added via `ui.run_javascript()` won't work
- Classes in scoped slot templates (e.g., `<q-badge :class="condition ? 'bg-red' : 'bg-green'">`) won't work
- Custom Vue components with dynamic classes won't work
- Requires users to "hoist" classes they plan to use dynamically

### Option B: Conditional Observers (Current PR)

**Approach**: Only observe elements with `component` field (custom Vue components).

**Pros**:
- Catches dynamic classes in custom components
- Lower overhead than full observer

**Cons**:
- Added complexity (tracking observers per element, lifecycle management)
- Inconsistent: `ui.run_javascript` on regular elements still won't work
- Scoped slot templates still won't work
- The `component` distinction feels arbitrary to users

### Option C: Single Document Observer ⭐️

**Approach**: One MutationObserver on `document.body` with `subtree: true`, watching only `class` attribute changes.

**Pros**:
- Catches **all** class changes regardless of source
- Simpler code (no per-element tracking, no lifecycle management)
- No surprising limitations for users
- Full compatibility as drop-in Tailwind replacement

**Cons**:
- Slightly more mutations to process

**Why the overhead is negligible**:
- MutationObservers are highly optimized in modern browsers
- We filter to `attributeFilter: ['class']` only
- We only collect class names (O(1) Set lookup per class)
- The expensive part (UnoCSS `extract()`) is already batched and only processes *new* classes
- The 3s→1.5s speedup came from bypassing UnoCSS's full-document regex parsing, not from avoiding observers

evnchn:
@falkoschindler Sounds reasonable. You can implement it, and I will benchmark it with some slow machines I have lying around to back up the numbers.

evnchn:
I have a correction on the AI summary:

> The 3s→1.5s speedup came from bypassing UnoCSS's full-document regex parsing, not from avoiding observers

- 3s is Tailwind CDN JS
- 1.5s is bare UnoCSS with no observer

So provided your observer implementation fits within 1.5s of runtime under "low-tier mobile" CPU throttling profile then we have a win, which I am confident in.

falkoschindler:
I noticed that template slots only work if we observe "childList" mutations. The new document observer seems to work correctly in various scenarios:
```py
ui.table(rows=[]).add_slot('no-data', '<div class="text-red-500">This is red.</div>')

label = ui.label('Label')
ui.button('Make green', on_click=lambda: ui.run_javascript(f'{label.html_id}.classList.add("text-green-500")'))
ui.button('Make yellow', on_click=lambda: label.classes('text-yellow-500'))

ui.button('Create a blue label', on_click=lambda: ui.label('Blue label').classes('text-blue-500'))
```

evnchn:
~~3.66s vs 2.86s~~ Wrong but shows improvement. See below.

Note that my testing methodology may not be the same, and I am using a MacBook with a much larger slowdown factor, so the numbers may not be comparable. Nevertheless I can say for sure UnoCSS is more performant, and I am just on `wind4` preset. Surely further performance / network transmission gains can be had, if we use `mini` preset. 

So yeah, let's take this. I think that global observer is the most "nice" and works-out-of-the-box approach out of everything.

evnchn:
Sorry I did not use incognito mode and it's timing my chrome extensions. Results above is void, but still shows speed improvement 🤡

evnchn:
OK. Lighthouse Desktop in incognito, LCP 1.1s (main) vs 0.7s (this PR)

I am getting rusty with performance testing but I still know my way.

falkoschindler:
Wow, 36% down - that's something! 👍🏻

evnchn:
Two and a half questions:

- Do we need to minify `tailwind-v4.css` and `tailwind.css`?
- `preset-wind4` can inject preflights for us. Do we even _need_ `tailwind-v4.css` in the first place?
  - Consider that `tailwind-v4.css` is basically in the source code of `preset-wind4.global.js`

falkoschindler:
Opus 4.5:

**Current setup:**
```html
@import url(".../unocss/tailwind-v4.css") layer(base);
```
```javascript
presets: [
  () => window.__unocss_runtime.presets.presetWind4({
    preflight: false,  // disabled!
  }),
],
```

**The redundancy:** You're loading the preflight CSS as a separate file, *and* preset-wind4.global.js contains the same preflight code (just disabled).

**The fix:** Enable preflights and map them to the correct layer:

```javascript
window.__unocss = {
  presets: [
    () => window.__unocss_runtime.presets.presetWind4({
      dark: { ... },
      preflight: true,  // enable it
    }),
  ],
  outputToCssLayers: {
    cssLayerName: (layer) => {
      if (layer === 'preflights') return 'base';
      if (layer === 'default') return 'utilities';
      return undefined;
    },
  },
};
```

Then you can **remove** both `tailwind.css` and `tailwind-v4.css`.

**Benefits:**
- One fewer HTTP request
- ~10KB less to download (the CSS is already in the JS bundle)
- Simpler code

**Regarding minification:** If you do keep the files, yes minifying helps, but removing them entirely is even better.

evnchn:
The commit comment says it all: "preflight: false never worked in the first place"

For V3 we need it, and let's go with minification.

evnchn:
Wait there's also preflights for mini and wind3? That doesn't fully line up with the docs. 

mini
image

wind3
image

But the layer name is `preflights` so we missed it. Let me wire it over.

evnchn:
Ref: http://jsben.ch/#/OJ3vo (from https://stackoverflow.com/a/7296616)

falkoschindler:
> Ref: http://jsben.ch/#/OJ3vo (from https://stackoverflow.com/a/7296616)

@evnchn I'm not 100% sure if our use case is equivalent to simply concatenating an array of strings since we're filtering by uniqueness using a set. And the performance gain should be marginal while the code grows.

Here's Opus' analysis which unveils a potential race condition with your solution:

---

The `appendClass` approach is **reasonable but the performance claim is dubious**. Here's my analysis:

**What's actually faster:**
- String concatenation (`+=`) is fast in modern JS engines
- BUT `renderedClasses.has()` and `renderedClasses.add()` are still O(1) Set operations (unchanged)

**What was removed:**
- `new Set()` per mutation batch (was: `const classes = new Set()`)
- `[...classes].filter()` spread + filter

**What was added:**
- String concatenation with `+=`

**My take:**

The performance difference is negligible for this use case. We're talking about maybe a few microseconds saved while:
- `querySelectorAll("*")` takes milliseconds on large DOMs
- `extract()` is async and does CSS parsing

**However**, there's a subtle bug risk: `newClassesString` is module-scoped (declared with `let` outside the observer). If the observer fires twice rapidly before the first `queue.then()` runs, the second mutation will overwrite `newClassesString = ""` while the first is still queued.

The previous Set-per-callback approach didn't have this race condition because `newClasses` was captured in the closure passed to `.then()`.

**Verdict:** The optimization is premature, and the code is slightly less safe. I'd prefer the original Set approach for clarity and correctness, unless there's actual profiling showing this is a bottleneck (which it isn't).

evnchn:
Oh yes `newClassesString is module-scoped` is a breaker for https://github.com/zauberzeug/nicegui/pull/5485/commits/737b479c6195485ace321f94e571bb9c106a4446. @falkoschindler May I ask how does Opus think about https://github.com/zauberzeug/nicegui/pull/5485/commits/62d224bf2681673fea1fb66e337ec492f6d827c2? Thanks.

falkoschindler:
**The code looks good now.** A few minor observations:

1. **`appendClass` inside callback** - recreated every mutation. Tiny overhead, but acceptable for clarity. Alternative would be passing `newClassesString` by reference, which is messier.

2. **String vs Set tradeoff** - Still using string concatenation. It works, but I personally find the Set approach more readable:
   ```javascript
   const newClasses = [...classes].filter(c => !renderedClasses.has(c));
   ```
   vs
   ```javascript
   let str = ""; function appendClass(c) { if (!set.has(c)) { set.add(c); str += c + " "; } }
   ```

3. **Trailing space** - `newClassesString` ends with a trailing space (`"class-a class-b "`). Harmless for `extract()`, but slightly untidy.

**Verdict:** The code is correct and the race condition is fixed. The string approach is fine - it's a stylistic preference. Ship it! 🚀

evnchn:
@falkoschindler sidenote: If you used LLM and _continued the conversation_ from https://github.com/zauberzeug/nicegui/pull/5485#issuecomment-3812237498 to https://github.com/zauberzeug/nicegui/pull/5485#issuecomment-3812283783, while contextual continuity is useful, please do bear in mind that LLMs are statistically more likely in changing their stance in the second shot, so you may want double-sanity-check with a fresh start. Nevertheless Opus 4.5 looks to be less biased at times so maybe it's fine.

falkoschindler:
Out of curiosity I created a benchmark to actually _measure_ the performance of various approaches by copy&pasting the following code into the dev console:
```js
// Collect real class names from the page
const allClasses = [];
document.querySelectorAll("*").forEach((el) => el.classList.forEach((c) => allClasses.push(c)));
console.log(`Found ${allClasses.length} class occurrences`);

const iterations = 10_000;
const seen = new Set(); // Simulate renderedClasses check

// Approach 1: Set + join
const t1Start = performance.now();
for (let i = 0; i < iterations; i++) {
  seen.clear();
  const newClasses = [];
  for (const c of allClasses) {
    if (!seen.has(c)) {
      seen.add(c);
      newClasses.push(c);
    }
  }
  const result1 = newClasses.join(" ");
}
const t1End = performance.now();
console.log(`Set + join:    ${(t1End - t1Start).toFixed(2)}ms for ${iterations} iterations`);

// Approach 2: Set + filter + join (no check when adding, filter at end)
const t2Start = performance.now();
for (let i = 0; i < iterations; i++) {
  seen.clear();
  const classes = new Set();
  for (const c of allClasses) classes.add(c); // no check, Set dedupes automatically
  const newClasses = [...classes].filter((c) => !seen.has(c));
  for (const c of newClasses) seen.add(c);
  const result2 = newClasses.join(" ");
}
const t2End = performance.now();
console.log(`Set + filter:  ${(t2End - t2Start).toFixed(2)}ms for ${iterations} iterations`);

// Approach 3: String concatenation
const t3Start = performance.now();
for (let i = 0; i < iterations; i++) {
  seen.clear();
  let newClassesString = "";
  for (const className of allClasses) {
    if (!seen.has(className)) {
      seen.add(className);
      newClassesString += className + " ";
    }
  }
  const result3 = newClassesString;
}
const t3End = performance.now();
console.log(`String concat: ${(t3End - t3Start).toFixed(2)}ms for ${iterations} iterations`);

// Approach 4: String concatenation with addClass function
const t4Start = performance.now();
for (let i = 0; i < iterations; i++) {
  seen.clear();
  let newClassesString = "";
  function addClass(className) {
    if (seen.has(className)) return;
    seen.add(className);
    newClassesString += className + " ";
  }
  for (const className of allClasses) addClass(className);
  const result4 = newClassesString;
}
const t4End = performance.now();
console.log(`String concat: ${(t4End - t4Start).toFixed(2)}ms for ${iterations} iterations`);
```

Results on [nicegui.io](https://nicegui.io/):
```
Found 14906 class occurrences
Set + join:    667.80ms for 10000 iterations
Set + filter:  826.00ms for 10000 iterations
String concat: 622.70ms for 10000 iterations
String concat: 615.80ms for 10000 iterations
```

Results on [zauberzeug.com](https://zauberzeug.com/):
```
Found 895 class occurrences
Set + join:    96.00ms for 10000 iterations
Set + filter:  116.50ms for 10000 iterations
String concat: 65.00ms for 10000 iterations
String concat: 64.60ms for 10000 iterations
```

Given that these times are for 10,000 iterations, we're optimizing in the order of microseconds per mutation. Anyway, I'm fine with keeping the string approach. It can be significantly faster and the code isn't too complex. 👍🏻

website/documentation/content/section_styling_appearance.py
```diff
@@ -163,8 +165,39 @@ def tailwind_layers():
...
+    you can also use the [UnoCSS engine](https://unocss.dev/) to let Tailwind CSS classes take effect.
+
+    Pass one of the following presets to `ui.run(unocss=...)`:
+
+    - "mini": [UnoCSS Mini preset](https://unocss.dev/presets/mini)
+    - "wind3": [UnoCSS Wind3 preset](https://unocss.dev/presets/wind3)
+    - "wind4": [UnoCSS Wind4 preset](https://unocss.dev/presets/wind4)
+
+    UnoCSS is a smaller library and more performant, especially on small pages.
+    On "low-tier mobile" CPU throttling profile, load time for this page went from 1.1s down to 0.7s.
```
evnchn:
```suggestion
    In Lighthouse Desktop, load time for this page went from 1.1s down to 0.7s.
```

evnchn (commented):
The only change left, and let's merge afterwards.

Issue #5487, opened 2025-11-22
ui.navigate.to not triggering sub_pages_navigate event


GinjiJizai:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui


def root():
    ui.on('sub_pages_navigate', lambda event: ui.notify(f'Navigating to 2 {event.args}'))

    ui.label('main page')
    ui.separator()
    ui.sub_pages({'/': main, '/other': other})

def main():
    ui.label('Main page content')
    ui.link('Go to other page', '/other')
    ui.button('Click me', on_click=lambda: ui.navigate.to('/other'))  

def other():
    ui.label('Another page content')
    ui.link('Go to main page', '/')

ui.run(root, dark=True)
```

### Description

I'm using sub-pages to create a single-page application and I want to execute some code when the URL changes.

After researching the source code of the sub-pages module, I found that the `sub_pages_navigate` event is fired on sub-page navigation.

When I click a `ui.link`, the event is fired correctly and my notification is shown.
However, when I use `ui.navigate.to`, the event is not fired.
The event is also not fired on going back and forward in the browser.

Is there a reason why `ui.navigate.to` behaves differently from the `ui.link` implementation?

### NiceGUI Version

3.2.0

### Python Version

3.13.5

### Browser

Firefox

### Operating System

Windows

### Additional Context

My final goal is to check if my user is still authenticated if a new sub page is accessed. Instead of checking it on every sub page i try to find a central solution like the `starlette.middleware.base.BaseHTTPMiddleware`, which does not work in that case because sub pages use url based navigation.

evnchn:
Hello @GinjiJizai thanks for the report. 

I got you an explanation but it is in the form of a PR. Please wait ~~15~~ 7 minutes.

GinjiJizai:
@evnchn awesome, thank you for the fast response!

falkoschindler:
I don't think this is a bug because the "sub_pages_navigate" event is for internal use only, especially for the client to inform the server about a path change. It is not intended or documented as a user-facing API.

Instead, you can use the sub pages router:
```py
ui.context.client.sub_pages_router.on_path_changed(lambda path: ui.notify(f'Navigating to {path}'))
```

[...]

PR #5490, opened 2025-11-22
`navigate_sub_pages` mock `_handle_event` to let userspace handlers work

evnchn:
### Motivation

So in `ui.navigate.to` we kinda "cheat" and call `await client.sub_pages_router._handle_navigate(path)` directly. 

While it works to navigate the page, userspace handlers (`ui.on('sub_pages_navigate', ...)`) do not work, as seen in https://github.com/zauberzeug/nicegui/issues/5487

### Implementation

Don't cheat I guess? Go through the event listener system. 

- We go through the event listener(s) and invoke them with mocked dictionaries. 
- Userspace handlers work, of course, as of this change. 
- While, `client.sub_pages_router._handle_navigate` is still being ran because of the following code.

https://github.com/zauberzeug/nicegui/blob/7ff456b15397bde1ee6550fbd351b613962227ba/nicegui/sub_pages_router.py#L18-L22

### Progress

- [x] Pytests have been added (https://github.com/zauberzeug/nicegui/pull/5490/commits/c0547efe3a8378338ef73864a8c67d73a4b5850c).


falkoschindler:
@evnchn I'm curious what @rodja thinks about this PR.
But either way, we should probably add a pytest that fails on main and passes with this PR.

rodja (dismissed):
I like it!

rodja:
```run_javascript(f'emitEvent("sub_pages_navigate", "{path}");')``` might be simpler, but then we will need a roundtrip to the client.

**Comment by evnchn** [comment]
I did think of that (my initial instinct), but did not go with it.

**Review by falkoschindler — COMMENTED** [review]
Sorry, but I disagree with this PR. It is fixing something which isn't broken.

The "sub_pages_navigate" event is for internal use when the client needs to tell the server to update the UI because the user decided to navigate. But it is not intended or documented as a feature to detect page navigations. Therefore I don't think we should "mock" it.

Actually, the sub pages router has a method for registering navigation handlers:
```py
ui.context.client.sub_pages_router.on_path_changed(lambda path: ui.notify(f'Navigating to {path}'))
```

As far as I know, this isn't documented yet. But we can do that to resolve #5487.

PR #5492, opened 2025-11-22
Identify NiceGUI defaults using `new_class` type marker and let `default_props` win

evnchn:
### Motivation

As stated in https://github.com/zauberzeug/nicegui/issues/4857, we want user's default props to win NiceGUI's default values. 

### Implementation

We mark NiceGUI values with `default_props_or`, which, say passed `str`, returns a special type with name `nicegui!default_str` inheriting from `str`. 

Exclamation mark was used because we want to avoid user creating their own class which conflict with ours. And so this would never happen because it is invalid Python syntax: 

```py
class nicegui!default_str(str): # Causes SyntaxError: invalid syntax
    pass # imagine it was something else other than a pass
```

The cleverness comes from the inheritance: comparisons should simply work, because of Python's duck typing. 

Only at `__setitem__` of `Props` do we break away from the duck typing and check explicitly the type name, revealing that it is a default value, and skipping when `__key in self`

### Progress

### Inspiration

This PR is inspired by https://github.com/zauberzeug/nicegui/pull/4900/files, but more generic because `default_props_or` can wrap value of _any_ type. 

### Discussion

Seems like a brighter future than https://github.com/zauberzeug/nicegui/pull/5476 because for that, we are passing stuff with type `_DefaultSentinel` around code which expects `str`s and `int`s. 

But then actually I don't think we should let user's styles and classes win NiceGUI's, because:

- For classes, there is no key-value relationship, and so if user did `bg-red-500` and NiceGUI did `bg-gray-500`, it is difficult to then recognize user's `bg-red-500` as being able to conflict with our `bg-gray-500` and then give up the setting. 
- For styles, are you really sure you want to endorse, say, `ui.row.default_style('flex-wrap: wrap')`?
  - One thing to follow Quasar docs and then complain to us that Quasar props are not working. 
  - Another thing to pull `.default_style('flex-wrap: wrap')` out of thin air and expect it to work...

### DIFF
--- a/nicegui/elements/button.py
+++ b/nicegui/elements/button.py
@@ -3,6 +3,8 @@ from typing import Optional
 
 from typing_extensions import Self
 
+from nicegui.helpers import default_props_or
+
 from ..events import ClickEventArguments, Handler, handle_event
 from .mixins.color_elements import BackgroundColorElement
 from .mixins.disableable_element import DisableableElement
@@ -15,7 +17,7 @@ class Button(IconElement, TextElement, DisableableElement, BackgroundColorElemen
     def __init__(self,
                  text: str = '', *,
                  on_click: Optional[Handler[ClickEventArguments]] = None,
-                 color: Optional[str] = 'primary',
+                 color: Optional[str] = default_props_or('primary'),
                  icon: Optional[str] = None,
                  ) -> None:
         """Button
--- a/nicegui/helpers.py
+++ b/nicegui/helpers.py
@@ -8,11 +8,12 @@ import struct
 import sys
 import threading
 import time
+import types
 import webbrowser
 from collections.abc import Callable
 from inspect import Parameter, signature
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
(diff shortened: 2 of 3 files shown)


falkoschindler:
@evnchn The problem with this approach is that some types can't be used as base types:
```py
default_props_or(None)
```
```
TypeError: type 'NoneType' is not an acceptable base type
```
Same for `True` and `False`.

I did similar experiments in March this year and simply couldn't find a way to mark defaults without changing their type.

evnchn:
I think the answer is somewhere between these two PRs. What about:

- Marked type solution for classes which can be base type, seeing the ease when we want to do mutations. 
- Use `DEFAULT_PROP or None` for classes which cannot be base type, seeing we have no other choice.

evnchn:
The entire thing have a risk of becoming very cursed. Consider `x = default_props_or(None)`. 

If that succeeds then we would have something which is not `None` but is true for `x is None`. 

This PR's logic, while may stand in other languages, is fundamentally at odds with Python's types. Therefore I'm inclined to closing this. Close if you agree. Thanks!

PR #5497, opened 2025-11-23
Load codehilite.css faster (cache headers, file name hash cache-busting)

evnchn:
### Motivation

While working on #5493 I also noticed that, full cached, NiceGUI documentation reaches out to:

- `index.html` (obviously)
- `codehilite.css`

The latter we should be able to get rid of? 

### Implementation

- Filename may change so we do `${this.filename}` with the accompanying props. 
- Compute the file name and return the response using `_generate_codehilite_css()`, which is shared. 
- [ ] (Questionable) Apply `@lru_cache(maxsize=1)` to `_generate_codehilite_css()` because it doesn't ever change in runtime (an asusmption, is this right?)

### Progress

### Results

Slow 4G no CPU throttling (because network is the bottleneck)

Before: 1.46s load
After: 1.00s load

### DIFF
--- a/nicegui/elements/markdown.js
+++ b/nicegui/elements/markdown.js
@@ -4,7 +4,7 @@ export default {
   template: `<div></div>`,
   async mounted() {
     await this.$nextTick(); // NOTE: wait for window.path_prefix to be set
-    await loadResource(window.path_prefix + `${this.dynamic_resource_path}/${this.filename}`);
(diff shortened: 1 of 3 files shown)


evnchn:
> Would it be possible/reasonable to move it into /_nicegui/<version>/

Probably not, since we don't know what combo of library the user has installed (could be super out of date or extremely recent)

Yeah let's leave this PR as it is for now...

PR #5499, opened 2025-11-24
Implicit handshake: 33% faster time-to-first-interactivity

evnchn:
### Motivation

The "speed-focused" evnchn is back with another performance-improving PR, introducing the concept of "implicit handshake". 

This cuts the handshaking from 3 round-trips to 2 round-trips, hence the 33% faster time-to-first-interactivity

#### Recap before this PR

The handshake consists of 3 sequential steps, each requiring a round-trip:

1. EngineIO handshake
2. SocketIO handshake
3. NiceGUI's `handshake` message

#### The relevation

If SocketIO handshake takes query, why not: 

- Shove whatever we want to communicate in the 3rd step into the 2nd step?
- Instead of returning true or false explicitly, keep the SocketIO connection open if success, or close it down if not. 
- Simply execute the code for the 3rd step's callback `window.socket.emit("handshake", args, (ok) => {CODE}` in 2nd step's callback

This PR does exactly that. 

#### Cherry-on-tops

- Functionality signature of `_on_handshake` kept: For backwards-compatibility with NiceGUI testing code. The code is more robust because it calls `int` explicitly. 
- Less code in client side
- Cope with Python-SocketIO passing the 3rd parameter which we don't care with `_=None`

### Progress

  - [x] What is up with `test_prefetch_connects_after_navigation`? Has to do with #5116
- [x] Documentation is not necessary (no functional differences, just faster).

### DIFF
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -188,10 +188,11 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
 
 
 @sio.on('connect')
-async def _on_connect(sid: str, data: dict[str, Any], _=None) -> None:
+async def _on_connect(sid: str, data: dict[str, Any], _=None) -> bool:
     query = {k: v[0] for k, v in urllib.parse.parse_qs(data.get('QUERY_STRING', '')).items()}
     if query.get('implicit_handshake', '') == 'true' and not await _on_handshake(sid, query):
-        raise ConnectionRefusedError('handshake failed')
+        return False
+    return True
 
 
 @sio.on('handshake')


evnchn:
```py
import time

from nicegui import app, ui

app.config.socket_io_js_transports = ['polling'] # Chrome does not simulate latency in WebSocket


@ui.page('/')
async def page():
    time_begin = time.time()
    time_end = time.time()
    ui.label(f'Connection established in {time_end - time_begin:.2f} seconds.')


ui.run(show=False, reload=False)
```

Slow 4G, no disable cache, no CPU bottleneck (focus on network latency here)

Before PR: 1.93s
After PR: 1.31s

evnchn:
~~Did we purge the client in test_prefetch_connects_after_navigation?~~

~~I have a suspicion that because the handshake is so efficient, that the client is actually connected in prefetch.~~

I printed `len(Client.instances)` and it does go down.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5499/commits/bd986b6fe374f4b41fb1ff3ac160f0f196a87781 gets `test_prefetch_connects_after_navigation` to pass by disabling this PR's new logic under prefetch mode. 

While I would love to make the tests pass without the dual-mode in behaviour, I can't manage to do it... @rodja If you can, take it away, or maybe we can leave this for another PR (prefetch and prerender is so hard to debug 😭)

nicegui/nicegui.py
```diff
@@ -187,6 +187,13 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
     return client.build_response(request, 500)
 
 
+@sio.on('connect')
+async def _on_connect(sid: str, data: dict[str, Any], _=None) -> None:
+    query = {k: v[0] for k, v in urllib.parse.parse_qs(data.get('QUERY_STRING', '')).items()}
+    if query.get('implicit_handshake', '') == 'true' and not await _on_handshake(sid, query):
+        await sio.disconnect(sid)
```
falkoschindler: @evnchn Wouldn't it be better to `raise ConnectionRefusedError("handshake failed")` or `return False` instead of connect and immediately disconnect?

evnchn: The point of the implicit handshake is to disconnect ASAP if the provided data isn't right, else it is considered as a successful connection attempt by the client. 

I think that `await sio.disconnect` will be necessary. However I think you can play around with it so long as the tests pass and the thing works.

falkoschindler: I'm not so sure if we have tests for edge cases like a failing `_on_handshake`...

According to the [docs](https://python-socketio.readthedocs.io/en/v4/server.html#defining-event-handlers), the connect handler can return `False` or raise. I thought this would be even faster and more idiomatic than calling `disconnect`.

evnchn: Since client does not read the reason of disconnect, I think `return False` is good enough.

falkoschindler: I just pushed `raise ConnectionRefusedError('handshake failed')`. Should be fine. It's just a few bytes more payload in a rare edge case.

falkoschindler (dismissed):

evnchn:
Due to `raise` causing error logs, maybe `return False` is better...

falkoschindler:
For reference, here is a visual summary of this PR:

### Previous: 3 Round-Trips

```
 Browser                                    Server
    │                                          │
    │─────── 1. EngineIO Handshake ───────────>│
    │<────────── (transport setup) ────────────│
    │                                          │
    │─────── 2. SocketIO Connect ─────────────>│
    │<────────── (namespace join) ─────────────│
    │                                          │
    │─────── 3. emit("handshake", { ──────────>│
    │            client_id,                    │
    │            document_id,                  │  Server calls
    │            tab_id,                       │  _on_handshake()
    │            old_tab_id,                   │
    │            next_message_id               │
    │        })                                │
    │<────────── callback(ok) ─────────────────│
    │                                          │
    │           ✓ Ready to interact            │
```

### New: 2 Round-Trips (Implicit Handshake)

```
 Browser                                    Server
    │                                          │
    │─────── 1. EngineIO Handshake ───────────>│
    │<────────── (transport setup) ────────────│
    │                                          │
    │─────── 2. SocketIO Connect ─────────────>│
    │            ?client_id=...                │
    │            &document_id=...              │  Server's _on_connect()
    │            &tab_id=...                   │  sees implicit_handshake=true,
    │            &old_tab_id=...               │  calls _on_handshake()
    │            &next_message_id=...          │  immediately
    │            &implicit_handshake=true      │
    │<────────── (namespace join) ────────────-│
    │                                          │
    │           ✓ Ready to interact            │
```

### Key Insight

The NiceGUI handshake data (`document_id`, `tab_id`, etc.) was being sent as a separate message *after* SocketIO connected. But SocketIO's connect already supports query parameters - so the PR piggybacks the handshake data onto the connect request itself.

The server's `_on_connect` handler checks for `implicit_handshake=true` in the query, and if present, calls `_on_handshake()` right there instead of waiting for a separate `"handshake"` message.

**Result**: One fewer round-trip = 33% faster time-to-first-interactivity (3→2 round-trips).

falkoschindler:
Unfortunately this PR breaks hot reload (see #5671) and connections via NiceGUI On Air. To act quickly, I'm reverting this PR. Later we can think about a new attempt that works in all cases - if it's worth the effort.

PR #5500, opened 2025-11-24
Initial-load element definition never needed component key nor tag so we drop it

evnchn:
### Motivation

While working on https://github.com/zauberzeug/nicegui/pull/5495#issuecomment-3568186974, I realize there's literally space savings on-the-table, by inspecting code behaviour. 

#### Observation: Only 2 places where JS dependencies are loaded

1. In the HTML, we add JS imports statically

https://github.com/zauberzeug/nicegui/blob/eea9a48071f274657e65409571c255b9b35830e8/nicegui/dependencies.py#L193-L206

2. In JS, we load them

https://github.com/zauberzeug/nicegui/blob/eea9a48071f274657e65409571c255b9b35830e8/nicegui/static/nicegui.js#L281-L290

There is no other place. I looked. 

#### Fulfill the bare-minimum for the code logic

For 1 to not trigger 2, we do the following:

https://github.com/zauberzeug/nicegui/blob/eea9a48071f274657e65409571c255b9b35830e8/nicegui/static/nicegui.js#L148-L149

But note we only care about the `.name`, nothing else. So why ship the bulk? 

### Progress

  - [x] Does existing tests pass?
- [x] Documentation is not necessary (identical behaviour but faster).

### Final note

This re-affirms the belief in #5495 that we key by name, not by hash-key. 

**Is this a simple-win? I think it is if you memorize the codebase.**

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -226,7 +226,6 @@ class Element(Visibility):
                     'children': [child.id for child in self.default_slot.children],
                     'events': [listener.to_dict() for listener in self._event_listeners.values()],
                     'update_method': self._update_method,
-                    'component': self.component.name if self.component else None,
                 }.items()
                 if value
             },
--- a/nicegui/outbox.py
+++ b/nicegui/outbox.py
@@ -108,15 +108,17 @@ class Outbox:
                         for element_id, element in self.updates.items()
                     }
                     js_components = [
-                        {'key': component.key, 'name': component.name, 'tag': component.tag}
+                        component
                         for element in self.updates.values()
                         if not isinstance(element, Deleted)
                         and isinstance((component := element.component), JsComponent)
                         and component.name not in self._loaded_components
                     ]
                     if js_components:
-                        coros.append(self._emit((client.id, 'load_js_components', {'components': js_components})))
-                        self._loaded_components.update(c['name'] for c in js_components)
+                        coros.append(self._emit((client.id, 'load_js_components', {
+                            'components': [{'key': c.key, 'tag': c.tag} for c in js_components],
+                        })))
+                        self._loaded_components.update(c.name for c in js_components)
                     coros.append(self._emit((client.id, 'update', data)))
                     self.updates.clear()
 
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -43,7 +43,6 @@ function replaceUndefinedAttributes(element) {
   element.text ??= null;
   element.events ??= [];
   element.update_method ??= null;
-  element.component ??= null;
   element.slots = {
     default: { ids: element.children || [] },
     ...(element.slots ?? {}),


falkoschindler (commented):

However, I'm wondering if the benefit is worth the complexity and additional computation. Let's hear what Opus 4.5 thinks about it (and I tend to agree):

---

**The benefit:**
- Saves ~55-60 bytes per element *with a custom component* (`key` is a ~40 char hash, `tag` is ~15 chars like `"nicegui-plotly"`)
- But most elements are native Quasar/HTML elements (buttons, labels, cards, etc.) which don't have `component` at all

**The cost:**
- Calls `_strip_extra_component_info()` for **every** element on every page load
- Adds a dict lookup (`if 'component' in element_dict`) for each element
- Adds code complexity and cognitive overhead for maintainers

**The math doesn't favor it:**
- A typical page might have 100+ elements but only 0-5 custom component elements (plotly, joystick, echart, etc.)
- So you're paying the overhead on 100 elements to save ~300 bytes total
- The page likely loads megabytes of CSS/JS libraries anyway

**My take:** I agree with your skepticism. The complexity-to-benefit ratio is poor. The optimization feels premature - it solves a problem that probably doesn't exist in practice.

If payload size were truly a concern, a cleaner approach would be:

1. Modify `_to_dict()` to accept an optional parameter, or
2. Have a separate lightweight dict builder for initial load, or  
3. Just leave it as-is - the extra bytes are negligible

I'd lean toward **closing the PR** unless there's a demonstrated performance issue with large pages full of custom components. The current code is simpler and the "waste" is trivial.

evnchn:
The thing I am tring to strip, is repeated **per use** of elements which need a component. Please use a Ctrl-F search. I definitely need to get that straight*

image

Indeed it could be an exception of our documentation website that we use many `ui.markdown`, and indeed a function call + 2 pops is hardly efficient in server-side. 

However, the grand spirit of this PR is actually rather to rework, with the enlightenment in https://github.com/zauberzeug/nicegui/pull/5495, the way we notate the inclusion of a component. 

Moving forward, we will index by name only, no key, no path. This way, we are doing dict lookup only on the elements with components, and reducing repetition from root source.

I would argue an environment for which data redundancy is minimized is the lowest cognitive overhead. 

And with this grand goal in mind, I do not agree with the partial-solutions 1 2 and 3 proposed by Opus. 

---

With the following phrase:

> A typical page might have 100+ elements but only 0-5 custom component elements (plotly, joystick, echart, etc.)

I quite doubt LLM's analytical ability, as it directly conflicts with point (*) stated above, which is super easy to debunk. 

@falkoschindler noting the response had "I agree with your skepticism", I would prefer if it is possible to prompt the LLM in a way that does not let it feel the user is skeptic in one way or another. If you still had the prompt, we can engineer it together.

evnchn:
The fact that this PR currently moves the loaded-component set from client to server, it is blocked by https://github.com/zauberzeug/nicegui/pull/5421

falkoschindler:
> The thing I am trying to strip, is repeated per use of elements which need a component.

Yes, I know. But I (and Opus) didn't think of `ui.markdown` which indeed can occur very often on a page.

> I would prefer if it is possible to prompt the LLM in a way that does not let it feel the user is skeptic in one way or another

You're right, and I do try to prompt LLMs such that I get an objective unbiased answer. But I'll try to be even more careful in the future.

The new "simplification" looks really interesting. Now we're removing complexity - at the cost of the server needing to keep track of loaded components. I think this might work.

Regarding the shared client (PR #5421): You think it would falsely assume that components had already been loaded even if a new browser tab connects. But is this really the case? Wouldn't a new tab get all `js_imports` up to this point via HTTP? Apart from that, at the moment PR #5421 is just an experiment. While it is good to think about possible impacts of this PR in the future, we might not necessarily need to block it.

evnchn:
> Wouldn't a new tab get all `js_imports` up to this point via HTTP?

Oh yes. So basically all further connected clients are "up to speed" and the server just needs to broadcast for the elements which did not make it. 

Assuming fully ensured message delivery from Socket.IO, then this should be fine. If the message is lost, then the client will not work (is this an extra failure point that this PR is introducing?)

Issue #2181, opened 2023-12-15
Can NOT set ui.input's prop autocomplete


rjzhou:
### Description

```py
ui.input().props('autocomplete="off"')
```
doesn't work

In the code input.py & input.js, I think this._props['autocomplete'] should be replaced with this._props['_autocomplete'] in input.py, and change input.js accordingly.

falkoschindler:
Interesting, @rjzhou! So you're trying to control the [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute of HTML's input element. I haven't thought of this conflict. Luckily your solution works perfectly.

I tested with the following code snippet:
```py
ui.input('First name').props('autocomplete="first-name"')
ui.input('Family name').props('autocomplete="family-name"')
ui.input('Something else').props('autocomplete="off"')
```

PR #5505, opened 2025-11-24
Resolve default parameter values with a decorator

evnchn:
### Motivation

With #5476 and #5492 it is obvious that there is so much we can do to, say, `button.py`, to:

- make default props win
- keep readable syntax
- keep simple downstream code
- have the solution work for many types including `None` and `True` and `False`

But why keep yourselves into just changing `button.py`? What about some external patching? 

This PR explores just that. 

### Implementation

It makes this simply work. 

```py
from nicegui import ui

ui.button.default_props('color="secondary"')


@ui.page('/')
def test_page():
    ui.button('Test Button 1')  # Should use color='secondary'
    ui.button('Test Button 2', color='red')  # Should use color='red'


ui.run(reload=False)

```

And it works like this:

image

image

Where `DEFAULT_PROPS` is a sign that, the `'primary'` over there? You will win over it even if you did not pass the argument, if you have default props set. 

And if you click inside the definition then you will see `DEFAULT_PROPS['color']` which tells you that, to win, kindly set `default_props('color="BLAH"')`

### Progress

### Discussions

- Where in the library to do the wrapping? (GitHub Copilot suggested https://github.com/zauberzeug/nicegui/pull/5505/commits/29bd97f58ccc541c9b4e5f694ed8bcb8fb41d209)
- Performance degredations? (We don't care?)

### DIFF
--- a/nicegui/elements/select.py
+++ b/nicegui/elements/select.py
@@ -1,8 +1,6 @@
-from __future__ import annotations
-
 from collections.abc import Generator, Iterable, Iterator
 from copy import deepcopy
-from typing import Any, Callable, Literal
+from typing import Any, Callable, Literal, Optional, Union
 
 from ..defaults import DEFAULT_PROP, DEFAULT_PROPS, resolve_defaults
 from ..events import GenericEventArguments, Handler, ValueChangeEventArguments
@@ -16,16 +14,16 @@ class Select(LabelElement, ValidationElement, ChoiceElement, DisableableElement,
 
     @resolve_defaults
     def __init__(self,
-                 options: list | dict, *,
-                 label: str | None = DEFAULT_PROP | None,
+                 options: Union[list, dict], *,
+                 label: Optional[str] = DEFAULT_PROP | None,
                  value: Any = DEFAULT_PROPS['model-value'] | None,
-                 on_change: Handler[ValueChangeEventArguments] | None = None,
+                 on_change: Optional[Handler[ValueChangeEventArguments]] = None,
                  with_input: bool = False,
-                 new_value_mode: Literal['add', 'add-unique', 'toggle'] | None = DEFAULT_PROP | None,
+                 new_value_mode: Optional[Literal['add', 'add-unique', 'toggle']] = DEFAULT_PROP | None,
                  multiple: bool = DEFAULT_PROP | False,
                  clearable: bool = DEFAULT_PROP | False,
-                 validation: ValidationFunction | ValidationDict | None = None,
-                 key_generator: Callable[[Any], Any] | Iterator[Any] | None = None,
+                 validation: Optional[Union[ValidationFunction, ValidationDict]] = None,
+                 key_generator: Optional[Union[Callable[[Any], Any], Iterator[Any]]] = None,
                  ) -> None:
         """Dropdown Selection
 


evnchn:
Why patch when you can decorate? 

https://poe.com/s/2SXdlMHOwwDl6t1BZjBT?utm_source=link

falkoschindler:
Very interesting approach, @evnchn!

I like that the value is resolved right in `__init__` and the sentinel doesn't have to be passed to mixins.

But the decorator is hard to get a head around, therefore hard to prove right in every case, and based on assumptions that might not always hold: Think about `ui.column` with a parameter `align_items` mapping to classes "items-start", "items-end", ... I guess this approach doesn't work here as well as PR #5476.

And it is debatable whether
```py
color: Optional[str] = DEFAULT_PROP or 'primary'
```
vs.
```py
color: Union[Optional[str], DEFAULT_PROPS['color']] = 'primary'
```
is nicer to read. Or in the more modern version:
```py
color: str | None = DEFAULT_PROP or 'primary'
```
vs.
```py
color: str | None | DEFAULT_PROPS['color'] = 'primary'
```
Your approach suggests `color` could be a default prop, but is "primary".
Mine reads more like `color` can be `str` or `None` and is whatever you put in `default_props` or (if nothing) "primary".

Anyway, the decorator is an interesting technique, but we should start with defining the signature first and find an implementation second. Although the library needs to be tidy and maintainable, most users only get to see the facade and don't care about the inner workings.

evnchn:
Man https://github.com/zauberzeug/nicegui/pull/5505/commits/8b4940e2bc8b8b48b1d6fa3f60c2920c9017c7b1 is so good:

- Nice signature: `color: Optional[str] = DEFAULT_PROPS('color') | 'primary',`
  - Retains the original type!
  - Expanding the type to Union an extra type `DEFAULT_PROPS` may throw some analyzers off...
- **Fully automatic application** of default props lookup: No more `handle_default_props`
- Not a single `#type: ignore` or other ignore statements in sight for UI element code. 

To the point I chose to direct-commit in this branch because the old one didn't stand a chance.

evnchn:
> decorator is hard to get a head around

The new one is workable? Please tell me if anything seems unclear. 

```py
@functools.wraps(original_func)
def decorated(*args, **kwargs):
    bound = inspect.signature(original_func).bind(*args, **kwargs)
    bound.apply_defaults()

    element: Element = bound.arguments['self']

    return original_func(**{
        param_name: (element._default_props.get(value.prop_key, value.default_value)  # pylint: disable=protected-access
                        if isinstance(value, DEFAULT_PROPS) else value)
        for param_name, value in bound.arguments.items()
    })

return decorated
```

evnchn:
> Although the library needs to be tidy and maintainable, most users only get to see the facade and don't care about the inner workings.

Sorry, but tricky internals would impede NiceGUI having more third-party components and elements. Let's avoid that whenever possible.

falkoschindler:
Oh WOW! 8b4940e2bc8b8b48b1d6fa3f60c2920c9017c7b1 is insanely good! 😍
It checks so many boxes:

- human-readable signature
- typing
- nice library code
- nice user code
- efficient
    (I added a cache for `signature(original_func)` so that we don't have to recompute it for every instance.)

- [x] A remaining issue: Currently the function is untyped. When adding types like
    ```py
    from typing import Callable, TypeVar, ParamSpec
    
    _P = ParamSpec('_P')
    _T = TypeVar('_T')
    
    def honor_default_props(original_func: Callable[_P, _T]) -> Callable[_P, _T]:
        # ... rest of implementation
    ```
    we notice that mypy isn't happy about plugging `element` as a keyword argument because it is positional only. So far I haven't found a handy solution, but it should be possible for sure.

falkoschindler:
I wonder how we would write the signature for `ui.row` with `align_items: Optional[Literal['start', 'end', 'center', 'baseline', 'stretch']] = None`. After setting default classes to "start", creating a `ui.row(align_items='end')` would cause both classes to be present. But maybe we keep the signature unchanged and the initializer simply removes existing "items-" classes in this case (see PR #5476) before adding a new one.

evnchn:
> Calling `ui.row.default_classes('items-center')` vis-a-vis `ui.row(align_items='end')`

Here is my brainstorming

#### Default classes take priority

`align_items: Optional[Literal['start', 'end', 'center', 'baseline', 'stretch']] = DEFAULT_CLASSES(prefix='items-') | None,`

which indicates NiceGUI will give up setting `align_items` if any class in `default_classes` matches prefix `items-`

#### NiceGUI take priority

`align_items: Optional[Literal['start', 'end', 'center', 'baseline', 'stretch']] = OVERRIDE_CLASSES(prefix='items-') & None,`

which indicates NiceGUI will override and remove existing classes if they match prefix `items-`

---

Right now both have a downside: What if I had a custom class `items-blah`? If default classes take priority, we are over-sensitive. If NiceGUI take priority, we induce collateral damage. 

2 solutions: 

1. We may shove information into the smart class placeholder like followings:

`align_items: Optional[Literal['start', 'end', 'center', 'baseline', 'stretch']] = DEFAULT_CLASSES('items-start', 'items-end', 'items-center', 'items-baseline', 'items-stretch') | None,`

`align_items: Optional[Literal['start', 'end', 'center', 'baseline', 'stretch']] = OVERRIDE_CLASSES('items-start', 'items-end', 'items-center', 'items-baseline', 'items-stretch') & None,`

But this is getting unwieldy...

2. Or we may have the `DEFAULT_CLASSES`/`OVERRIDE_CLASSES` inspect the function argument typing for inference, but then the decorator gets unwieldy instead...

# 😵‍💫

Perhaps on that, #5476's approach is good enough...

falkoschindler:
Regarding `ui.row`, I would leave the signature for now. We can use parts of PR #5476's approach to correctly overwrite default classes. But this doesn't interfere with this PR.

I've got some more thoughts though:

- [x] Currently mypy doesn't spot mistakes like this:
    ```py
    color: Optional[str] = DEFAULT_PROPS('color') | 42
    ```
    A solution might be to replace `Any` with a type var `T`:
    ```py
    def __or__(self, other: T) -> T:
    ```
    and ignore type warnings in `__or__`.
- [x] As an alternative to
    ```py
    DEFAULT_PROPS('color')
    ```
    we could certainly achieve a syntax with square brackets
    ```py
    DEFAULT_PROPS['color']
    ```
    Do we want that?

evnchn:
> we could certainly achieve a syntax with square brackets

I don't see how, somehow. Maybe you can try it out.

evnchn:
TypeVar should be working: 

image

To elaborate: I don't see how we can achieve a syntax with square brackets, because we need to store the value of `'color'` per-instance into `DEFAULT_PROPS`, and `DEFAULT_PROPS['color']` is not creating a new instance AFAIK.

evnchn:
> mypy isn't happy after we typed the decorator

But according to https://poe.com/s/2W7avPZhuA0x47DPrhno, ParamSpec is Py3.10+?

evnchn:
I think: 

- The overhead involved for getting the square bracket syntax may not be worth it over the syntactic sugar. 
- I don't think we can leverage args and kwargs for more customization in the instantiation of the sentinel if we go with square bracket. Consider round-bracket we could have had `DEFAULT_CLASSES(prefix='items-')` if the situation called for it. 

Maybe we should roll it back. Then again I may be biased.

falkoschindler:
I refactored the code once again, tried square brackets, added a pytest, and the result looks pretty good if you ask me.

Yes, `ParamSpec` is Py3.10+, but you can already import it `from typing_extensions`.

> The overhead involved for getting the square bracket syntax may not be worth it over the syntactic sugar.

I don't think it's too much overhead. It's basically a three-line factory class. And we won't need to touch it ever again.

> I don't think we can leverage args and kwargs for more customization in the instantiation of the sentinel if we go with square bracket. Consider round-bracket we could have had `DEFAULT_CLASSES(prefix='items-')` if the situation called for it.

We can simply choose `DEFAULT_CLASSES = Sentinel()` to manage default classes. For default props `DEFAULT_PROPS[...]` feels more intuitive to me.

Regarding the decorator: What do you think about a new name?
- `apply_default_props`
- `resolve_default_props`
- `use_default_props`
- `with_default_props`

Apart from that, we might be ready to convert the element signatures (supported by some AI agent I guess). This will unveil edge cases, maybe default classes or default style we need to handle.

evnchn:
In other words, my main issue with `DEFAULT_PROPS[...]` is that it does not allow me to do `DEFAULT_PROPS[..., option=...]` to add more config into the sentinel. If you are comfortable in forfeiting that flexibility (which may be useful down the line) for the better syntax, go for it. 

I think "honor" is the only word I can think of that highlights the fact the default props are indeed set in userland, but just sitting there collecting dust as it is overshadowed. 

If you don't think this emphasis is important, then I would prefer either `apply_default_props` first and `resolve_default_props` secondary. 

We are already "using" default props, and "with" is a "reserved keyword in programmer's mind" and should not appear in variable names _ever_. Though there is some resolution done in the decorator, the operation is so minimal that it does not deserve the word "resolve".

falkoschindler:
> In other words, my main issue with `DEFAULT_PROPS[...]` is that it does not allow me to do `DEFAULT_PROPS[..., option=...]` to add more config into the sentinel. If you are comfortable in forfeiting that flexibility (which may be useful down the line) for the better syntax, go for it.

Yes, I'm comfortable with taking this route because we still can add support for `DEFAULT_PROPS('foo', strict=True)` by implementing `__call__`. This can live side by side, like the `.props` method supporting indexing. And for other applications like `DEFAULT_CLASSES` we'll first have to define requirements before deciding about the API.

> I think "honor" is the only word I can think of that highlights the fact the default props are indeed set in userland, but just sitting there collecting dust as it is overshadowed.

"honor" sounds a bit vague for me as a developer. The verb is less common in programming.

> If you don't think this emphasis is important, then I would prefer either `apply_default_props` first and `resolve_default_props` secondary.

I lean towards "resolve" because the decorator doesn't really _apply_ anything, but it compares input arguments with default props and resolves the value which is fed into the initializer.

Thinking of other possible sentinels like `DEFAULT_CLASSES`, we might want to use a more generic name like `resolve_defaults`.

> We are already "using" default props, and "with" is a "reserved keyword in programmer's mind" and should not appear in variable names _ever_.

Agreed.

> Though there is some resolution done in the decorator, the operation is so minimal that it does not deserve the word "resolve".

Sure, the resolution is not too complicated. But it deals with default props, the default argument value and the actual user value. I'd call it resolution.

Last thought: nouns are more common for decorators than verbs (`@property`, `@dataclass`, ...). But I can't find a good one. `@wraps` is an exception. Maybe `resolves_defaults` is a good choice?

My top picks so far: `resolve_defaults` or `resolves_defaults`.

evnchn:
Let's do `resolve_defaults` because the two s in `resolves_defaults` is a bit of a tongue twister

falkoschindler:
I think this is ready for review. Some remarks:

- In `slide_item.py`, the `action()` method has a `color` parameter without `DEFAULT_PROPS`. The wrapper methods (`left()`, `right()`, `top()`, `bottom()`) use `DEFAULT_PROPS['left-color']`, etc., and call `action()`. This works because the wrappers resolve the defaults first, but if someone calls `action()` directly, it won't respect `default_props()`. At the moment `DEFAULT_PROPS` doesn't support matching a key based on another `side` argument. But I think this is an edge case and can be ignored for now.

- We use `DEFAULT_PROPS` everywhere where a parameter could conflict with default props. As it turns out, this is the case in many places and many signatures now contain a `DEFAULT_PROPS` expression. But it probably doesn't make sense to limit its usage to fewer cases because only _with_ our `DEFAULT_PROPS` sentinel we correctly resolve default values.

- I noticed inconsistent naming schemes for props: Most props use kebab-case, but some snake_case and some camelCase. We could attempt to unify it, but this might surprise users that use these props unofficially in their code. On the other hand: Once we publish our new signatures using `DEFAULT_PROPS['xy']`, we kind of add prop names to the public API. So maybe we better break the undocumented API now before documenting it.

- This PR doesn't address default props for `ui.notification` (see #4330). This can be done in a separate PR by splitting `options` into individual props.

evnchn (commented):
ColorElement used to do nothing when passed None. Now it pops the key. Not sure. 

- considerable as breaking?
- but old code not popping the key is more of a bug (failure to unset)

I'm sided on the latter. 

---

On the inconsistent naming, I would not recommend unifying them in this PR so as to not land people with a surprise in a minor release. While the props are undocumented, we condone its use in the discussions forums. 

When the naming is sufficiently unified (or maybe now), can we go for a short form? Where `DEFAULT_PROPS` standalone without the brackets stand for "fetch the value from the key of the same name as the parameter name".

In other words: 

```py
                 placeholder: Optional[str] = DEFAULT_PROPS['placeholder'] | None,
```

becomes simply:


```py
                 placeholder: Optional[str] = DEFAULT_PROPS | None,
```

falkoschindler:
> ColorElement used to do nothing when passed None. Now it pops the key. Not sure.
> 
> * considerable as breaking?
> * but old code not popping the key is more of a bug (failure to unset)
> 
> I'm sided on the latter.

Yes, I think this was a bug.

> On the inconsistent naming, I would not recommend unifying them in this PR so as to not land people with a surprise in a minor release. While the props are undocumented, we condone its use in the discussions forums.

I do think we should use a consistent naming _now_ because

- The change will be even "more breaking" after exposing the internal prop names now and changing them later in 4.0. (Of course, we could also postpone this PR to 4.0. But since it fixes so many bugs, I'd rather not wait with releasing it.)
- It allows for automatic inference of prop names from parameter names (see below).
- Props mentioned in discussion are mostly well-documented props of Quasar elements which already follow a consistent kebab scheme. The ones we might need to change are often a bit more internal. Here's an AI-generated list:

    | File | Prop |
    |------|------|
    | `input.py` | `_autocomplete` |
    | `header.py` | `add_scroll_padding` |
    | `leaflet/leaflet.py` | `additional_resources` |
    | `scene/scene.py` | `background_color` |
    | `scene/scene.py` | `click_events` |
    | `colors.py` | `dark_page` |
    | `scene/scene.py` | `drag_constraints` |
    | `leaflet/leaflet.py` | `draw_control` |
    | `echart/echart.py` | `enable_3d` |
    | `leaflet/leaflet.py` | `hide_drawn_items` |
    | `codemirror/codemirror.py` | `highlightWhitespace` |
    | `aggrid/aggrid.py` | `html_columns` |
    | `codemirror/codemirror.py` | `lineWrapping` |
    | `fullscreen.py` | `requireEscapeHold` |
    | `scene/scene.py` | `show_stats` |
    | `scene/scene_view.py` | `show_stats` |

For props with wrong casing: Maybe we can detect them and warn the user?

> When the naming is sufficiently unified (or maybe now), can we go for a short form? Where `DEFAULT_PROPS` standalone without the brackets stand for "fetch the value from the key of the same name as the parameter name".

Great idea! This would remove much duplication. Unfortunately, there are cases where prop and parameter won't match, like `value` and "model-value". Here's an AI-generated list (some of them could probably be renamed):

| File | Parameter | Prop |
|------|-----------|------|
| `input.py` | `autocomplete` | `_autocomplete` |
| `expansion.py` | `text` | `label` |
| `editor.py` | `value` | `model` |
| `slide_item.py` | `color` | `left-color` |
| `slide_item.py` | `color` | `right-color` |
| `slide_item.py` | `color` | `top-color` |
| `slide_item.py` | `color` | `bottom-color` |
| (many files) | `value` | `model-value` |

But maybe a combination works well:
```py
value: Optional[str] = DEFAULT_PROPS['model-value'] | None,
placeholder: Optional[str] = DEFAULT_PROP | None,
```
This way the singular `DEFAULT_PROP` acts more like a plain sentinel and `DEFAULT_PROPS['model-value']` more like a dictionary access.

evnchn:
> use a consistent naming now

Fair argument. If we want to pursue that then I want to at-least ask, on the behalf of the end-user, to have a **confirmed** list of what props are changed (old name -> new name). 

Such that the messaging behind the change is:

- Not "sucks if your code doesn't work anymore"
- But "here is the exact list, grep for them and you should be good to go". 

The "AI-generated list" is a good start but we should **read and double-check.** 

> detect them and warn the user

I think we can go beyond detecting wrong casing but also the old prop name. 

And specifically for old prop name, why stop at detection? We may rectify the name _for_ the user while warning them to switch by 4.0 to have a seamless switchover

Though interventions are not recommend do this for wrong casing due to possible edge cases and considerable as introducing new behaviour. 

> combination of sentinel / dictionary access for `DEFAULT_PROPS`

Good idea. But we should document `DEFAULT_PROPS` then, because having 2 behaviour for 1 thing is possibly stepping into confusing territory.

falkoschindler:
Sure, the AI-generated list was just a starting point. Here is a unified, confirmed list of all name inconsistencies:

| File                     | Parameter              | Prop                   | Type          |
|--------------------------|------------------------|------------------------|---------------|
| colors.py                | `dark_page`            | `dark_page`            | ⚪️ snake case |
| fullscreen.py            | `require_escape_hold`  | `requireEscapeHold`    | 🟡 kebab case |
| header.py                | `add_scroll_padding`   | `add_scroll_padding`   | ⚪️ snake case |
| aggrid/aggrid.py         | `html_columns`         | `html_columns`         | ⚪️ snake case |
| codemirror/codemirror.py | `line_wrapping`        | `lineWrapping`         | 🟡 kebab case |
| codemirror/codemirror.py | `highlight_whitespace` | `highlightWhitespace`  | 🟡 kebab case |
| echart/echart.py         | `enable_3d`            | `enable_3d`            | ⚪️ snake case |
| leaflet/leaflet.py       | `additional_resources` | `additional_resources` | ⚪️ snake case |
| leaflet/leaflet.py       | `draw_control`         | `draw_control`         | ⚪️ snake case |
| leaflet/leaflet.py       | `hide_drawn_items`     | `hide_drawn_items`     | ⚪️ snake case |
| scene/scene.py           | `background_color`     | `background_color`     | ⚪️ snake case |
| scene/scene.py           | `click_events`         | `click_events`         | ⚪️ snake case |
| scene/scene.py           | `drag_constraints`     | `drag_constraints`     | ⚪️ snake case |
| scene/scene.py           | `show_stats`           | `show_stats`           | ⚪️ snake case |
| scene/scene\_view.py     | `show_stats`           | `show_stats`           | ⚪️ snake case |
| (many files)             | `value`                | `model-value`          | 🟠 other name |
| expansion.py             | `text`                 | `label`                | 🟠 other name |
| input.py                 | `autocomplete`         | `_autocomplete`        | 🟠 other name |
| slide\_item.py           | `color`                | `left-color`           | 🟠 other name |
| slide\_item.py           | `color`                | `right-color`          | 🟠 other name |
| slide\_item.py           | `color`                | `top-color`            | 🟠 other name |
| slide\_item.py           | `color`                | `bottom-color`         | 🟠 other name |

The case problems should be rather straight-forward.
The name inconsistencies probably have to remain unchanged:
- "model-value" and "...-color" are Quasar props which we don't want to change.
- `text` vs. "label" is an unfortunate mismatch, but I'd rather rename the positional parameter `text` in 4.0 instead of changing the Quasar prop "label".
- "_autocomplete" is a workaround to avoid a conflict with the HTML attribute `autocomplete` (see #2181).

By the way: This rigorous check of prop names unveiled a mistake which I swiftly fixed in 35f4e61c0e9cf250e16be3ceb4244828afd41009.

> We may rectify the name _for_ the user

This might be tricky. But I'll look into it.

> But we should document `DEFAULT_PROPS` then, because having 2 behaviour for 1 thing is possibly stepping into confusing territory.

Note that I proposed to use `DEFAULT_PROP` _and_ `DEFAULT_PROPS`. It reads rather nicely and this way they only have 1 behavior each.
```py
value: Optional[str] = DEFAULT_PROPS['model-value'] | None,
placeholder: Optional[str] = DEFAULT_PROP | None,
```

falkoschindler:
The updated list of all renamings is as follows:

| File                     | Parameter              | Old prop                | New prop              |
|--------------------------|------------------------|-------------------------|-----------------------|
| element.py               | -                      | resource\_path          | resource-path         |
| element.py               | -                      | dynamic\_resource\_path | dynamic-resource-path |
| colors.py                | `dark_page`            | dark\_page              | dark-page             |
| colors.py                | `custom_colors`        | customColors            | custom-colors         |
| fullscreen.py            | `require_escape_hold`  | requireEscapeHold       | require-escape-hold   |
| header.py                | `add_scroll_padding`   | add\_scroll\_padding    | add-scroll-padding    |
| markdown.py              | -                      | resource\_name          | resource-name         |
| markdown.py              | -                      | use\_mermaid            | use-mermaid           |
| aggrid/aggrid.py         | `html_columns`         | html\_columns           | html-columns          |
| codemirror/codemirror.py | `highlight_whitespace` | highlightWhitespace     | highlight-whitespace  |
| codemirror/codemirror.py | `line_wrapping`        | lineWrapping            | line-wrapping         |
| echart/echart.py         | `enable_3d`            | enable\_3d              | enable-3d             |
| leaflet/leaflet.py       | `additional_resources` | additional\_resources   | additional-resources  |
| leaflet/leaflet.py       | `draw_control`         | draw\_control           | draw-control          |
| leaflet/leaflet.py       | `hide_drawn_items`     | hide\_drawn\_items      | hide-drawn-items      |
| scene/scene\_view.py     | -                      | camera\_params          | camera-params         |
| scene/scene\_view.py     | -                      | camera\_type            | camera-type           |
| scene/scene\_view.py     | -                      | scene\_id               | scene-id              |
| scene/scene\_view.py     | `show_stats`           | show\_stats             | show-stats            |
| scene/scene.py           | `background_color`     | background\_color       | background-color      |
| scene/scene.py           | -                      | camera\_params          | camera-params         |
| scene/scene.py           | -                      | camera\_type            | camera-type           |
| scene/scene.py           | `click_events`         | click\_events           | click-events          |
| scene/scene.py           | `drag_constraints`     | drag\_constraints       | drag-constraints      |
| scene/scene.py           | `show_stats`           | show\_stats             | show-stats            |

The following props remain inconsistent with their parameter name:

| File           | Parameter      | Prop           |
|----------------|----------------|----------------|
| (many files)   | `value`        | model-value    |
| expansion.py   | `text`         | label          |
| input.py       | `autocomplete` | \_autocomplete |
| slide\_item.py | `color`        | left-color     |
| slide\_item.py | `color`        | right-color    |
| slide\_item.py | `color`        | top-color      |
| slide\_item.py | `color`        | bottom-color   |

falkoschindler:
The `Props` class already had a method `add_warning` to register unsupported props. I extended it with a method `add_rename` and have both called whenever the observable dictionary updates. This way we should be able to notice every case where an old prop is written. The value is automatically moved to the new prop so that existing apps should remain fully functional. I tested with this example:
```py
options = {
    'rowData': [{'name': 'M&uuml;nster'}],
}
ui.aggrid(options, html_columns=[0]).classes('h-24')
ui.aggrid(options).props('html_columns=[0]').classes('h-24')
ui.aggrid(options).props('html-columns=[0]').classes('h-24')
```
The HTML cell value is correctly interpreted in all three cases. A warning is shown for the deprecated "html_columns".

falkoschindler (approved):
Ok, it has been quite a ride. But I think this PR is ready to merge! 🎉

nicegui/props.py
```diff
@@ -105,6 +108,25 @@ def add_warning(self, prop: str, message: str) -> None:
...
+                with self.suspend_updates():
+                    del self[name]
+                helpers.warn_once(message)
+
+    def _check_renames(self) -> None:
+        for name, rename in self._renames.items():
+            if name in self:
+                with self.suspend_updates():
+                    self[rename] = self[name]
+                    del self[name]
```
evnchn, line 127: I don't think we should delete the old name just in case some code reads back from the old name

evnchn (commented):
Small issue left, and also does `_check_renames` handle props set by `default_props` (just want to be sure)

falkoschindler:
Oh dear, I though default props would show warnings as well, but I forgot to check.
As it turns out, we get _no_ warning:
```py
ui.aggrid.default_props('html_columns=[0]')
ui.aggrid({'columnDefs': [{'headerName': 'Name', 'field': 'name'}]})
```

The problem is: When `props` are created, warnings haven't been defined. But we can't define warnings via `props.add_warnings` before creating `props`.

Maybe can check warnings right when they are added. I'll create a new PR.

evnchn:
Oops I was on it
https://github.com/zauberzeug/nicegui/pull/5621

nicegui/props.py
falkoschindler: Resolved in PR #5622.

PR #5510, opened 2025-11-26
Ensure raw error for API exceptions by marking NiceGUI page requests in request scope

### LINKED ISSUE #5508: Error page should only apply to ui.page endpoints
### First Check

### Example Code

```python
# File1.py
from nicegui import app, ui

@ui.page("/")
def gui():
    ui.label("Hello!")

@app.get("/api")
def api():
    return {"response": "hello from api"}

@app.get("/api_error")
def api():
    return {"response": 1/0}

ui.run()

# File2.py
import requests

result = requests.get("http://127.0.0.1:8080/api_error")
print(result)
print(result.text)
```

### Description

I think the default error page should only apply to ui.page endpoints and not endpoints defined by app.get etc.

It creates a very ugly error response for api users, when they receive the error page as html when not viewed in a browser.
Instead api endpoints should be handled by fastapi's normal error handling.

### NiceGUI Version

3.11

### Python Version

3.11

### Browser

Edge

### Operating System

Windows

### Additional Context

_No response_

evnchn:
### Motivation

Fix #5508, where NiceGUI returns the HTML for the "sad face error page" under this FastAPI GET request:

```py
@app.get("/api_error")
def api():
    return {"response": 1/0}
```

Problems:

- Wasteful in bandwidth
- Difficult to parse and detect (remember: `client_id` is dynamic)
- Non API-like behaviour

Optimally for `@app`, it should act as much like FastAPI as possible. This PR does exactly that.

### Implementation

- Mark NiceGUI page requests in request scope
- Only `return client.build_response(request, 500)` if found
- If not we `raise exception` and FastAPI proceeds to "Internal Server Error" exactly like if a plain FastAPI app would. 

### Results

image

### Progress

- [x] Pytests have been added (https://github.com/zauberzeug/nicegui/pull/5510/commits/058d488f69f333436f117a447ccf2d8bde1a069b).
- [x] Documentation is not necessary (I consider this a bug, frankly).

### DIFF
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -164,7 +164,6 @@ def test_api_exception(screen: Screen):
     screen.allowed_js_errors.append('/ - Failed to load resource')
     screen.open('/')
     screen.should_contain('Internal Server Error')
-    screen.assert_py_logger('ERROR', 'some exception in a GET endpoint')
 
 
 def test_page_with_args(screen: Screen):


nicegui/nicegui.py
```diff
@@ -180,6 +180,8 @@ async def _exception_handler_404(request: Request, exception: Exception) -> Resp
 @app.exception_handler(Exception)
 async def _exception_handler_500(request: Request, exception: Exception) -> Response:
     log.exception(exception)
+    if not request.scope.get('nicegui_page_path'):
+        raise exception  # Simply return "Internal Server Error", just like FastAPI would do
```
evnchn:
```suggestion
    if not request.scope.get('nicegui_page_path'):
        raise exception  # Simply return "Internal Server Error", just like FastAPI would do
    log.exception(exception)
```

falkoschindler (approved):

evnchn:
And if I am not mistaken, the native FastAPI behavior is to not log. Not nice to the developer, but again, it's not called NiceAPI.

PR #5512, opened 2025-11-26
Color setters and bindings

evnchn:
### Motivation

In #5511, according to the rest of the library's code logic, this should work, but there is no such logic right now.

```py
from nicegui import ui

myinput = ui.input(label='Color?')

ui.button('Color-changing?').bind_background_color(myinput, 'value')

ui.run(reload=False, port=9091)
``` 

While there are ways around it (https://github.com/zauberzeug/nicegui/discussions/5511#discussioncomment-15083749), it doesn't work with all 3 of Quasar/Tailwind/CSS color, which users expect should work. 

This PR makes it work. 

### Implementation

- `set_*_color` to set color, defaulting to `clear=True` because need to clear previous color. 
- Refactor the color setting logic called on init to `set_*_color` with `clear=False` to ensure identical init behaviour. 
- Copy binding code from the rest of the library. 

### Progress

- [x] Pytests are complete.
  - [x] https://github.com/zauberzeug/nicegui/pull/5512/commits/5543d4a6eb3b32ebf2aeea18f39c587f7463b02d tests the setter and bind_to
  - [x] How about bind_from / ~~normal bind~~ (not doing it since it should work)? https://github.com/zauberzeug/nicegui/pull/5512/commits/fa4c9e475dc90bdfbe44d04a2a9a8e6c13e09954
- [x] No need docs? Clear enough? https://github.com/zauberzeug/nicegui/pull/5512#issuecomment-3580984225
- [x] Do we need all 3 binding functions? (uncheck and rectify if disagree)
  - IMO yes, since you can still write to a button's color via `mybutton.color = ...` 

### Final notes

While this isn't simple win, I would classify as "high similarity score" (imagine a Turnitin for GitHub), which rather indicates that the code is easier to review despite the mass.

### DIFF
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -1,62 +1,68 @@
+from collections import namedtuple
+
 from nicegui import ui
 from nicegui.testing import Screen
 
+ColorCase = namedtuple('ColorCase', ['label', 'color', 'result'])
+
 COLOR_CASES = [
-    ('1', 'rgba(239, 83, 80, 1)', 'red-5'),
-    ('2', 'oklch(0.637 0.237 25.331)', 'red-500'),
-    ('3', 'rgba(255, 0, 0, 1)', '#ff0000'),
-    ('4', 'rgba(0, 255, 255, 1)', '#00ffff'),
+    ColorCase('Quasar Red-5',     'red-5',   'rgba(239, 83, 80, 1)'),
+    ColorCase('Tailwind Red-500', 'red-500', 'oklch(0.637 0.237 25.331)'),
+    ColorCase('CSS Red',          '#ff0000', 'rgba(255, 0, 0, 1)'),
+    ColorCase('CSS Cyan',         '#00ffff', 'rgba(0, 255, 255, 1)'),
 ]
 
 
-def test_quasar_colors(screen: Screen):
+def test_colors_via_color_parameter(screen: Screen):
     @ui.page('/')
     def page():
         ui.button()
         ui.button(color=None)
-        for _, _, color in COLOR_CASES:
-            ui.button(color=color)
+        for case in COLOR_CASES:
+            ui.button(color=case.color)
 
     screen.open('/')
     assert screen.find_all_by_tag('button')[0].value_of_css_property('background-color') == 'rgba(88, 152, 212, 1)'
     assert screen.find_all_by_tag('button')[1].value_of_css_property('background-color') == 'rgba(0, 0, 0, 0)'
-    for i, (_, expected_color, _) in enumerate(COLOR_CASES, start=2):
-        assert screen.find_all_by_tag('button')[i].value_of_css_property('background-color') == expected_color
+    for i, case in enumerate(COLOR_CASES, start=2):
+        assert screen.find_all_by_tag('button')[i].value_of_css_property('background-color') == case.result
 
 
-def test_quasar_colors_via_setter(screen: Screen):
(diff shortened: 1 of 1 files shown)


evnchn:
We need all 3 bindings:

```py
mybutton = ui.button('Color-changing?')

def to_color(color):
    mybutton.background_color = color

ui.button('Red', on_click=lambda: to_color('red'))
ui.button('Green', on_click=lambda: to_color('green'))

current_color = ui.label()
mybutton.bind_background_color_to(current_color, 'text')
```

evnchn:
I think no need docs. 

image

evnchn:
On your 4 points: 

1. Yes
2. They did it to themselves
3. That's a dealbreaker. Didn't thought of that
4. Required to backwards-compatibility (no clearing of the color unless use the binding)

3 is a dealbreaker: I agree that without a reliable way of removing existing colors, the PR is not workable. This leaves us with:

- Write a regex for tailwind color classes
- Give up

I'm not sure. Should we do the regex?

evnchn:
https://github.com/zauberzeug/nicegui/pull/5512/commits/203ec9c827db6f8404edcc1b524df74243014764 should address your concerns.

I won't make color setter work for all Tailwind classes - that is just not time-feasible. Rather, we will document the bounds of ability for the color setters and bindings.

falkoschindler:
203ec9c827db6f8404edcc1b524df74243014764 looks pretty good to me.

> won't make color setter work for all Tailwind classes

You mean colors like "!bg-red-400" or "bg-[#123456]"? Sure, we can simply document this limitation.

Another idea though: Can't we keep track of the last color that has been set via `set_*_color` or binding? Then we only need to remove a single prop, class or style. This would even work for custom Tailwind colors like `set_text_color('[#123456]')`.

Regarding `_skip_clear` for backwards compatibility: Can you give an example where an existing color shouldn't be removed? I feel like preserving a color while assigning a new one can be considered a bug and should be fixed.

evnchn:
> `set_text_color('[#123456]')`

That would never work unfortunately, since square brackets are not recognized, and it falls through to setting style (which doesn't work)

> `_skip_clear` for backwards compatibility

No I do not have a case calling for the backwards compatibility. If you're comfortable making the change I'm comfortable too. I just prefer to keep changes initially backwards-compatible then possibly not afterwards.

falkoschindler:
> > `set_text_color('[#123456]')`
> 
> That would never work unfortunately, since square brackets are not recognized, and it falls through to setting style (which doesn't work)

Oh, you're right!

> > `_skip_clear` for backwards compatibility
> 
> No I do not have a case calling for the backwards compatibility. If you're comfortable making the change I'm comfortable too. I just prefer to keep changes initially backwards-compatible then possibly not afterwards.

As far as I can see, `_skip_clear` is only `True` during initialization. So the user has no chance to manually set a color before the element is created. The only chance I see would be a subclass that adds some Tailwind class before calling `super().__init__()`. But this is also basically impossible, because only in `Element.__init__` the `props` and `classes` collections are created.
So yes, I guess we don't need `_skip_clear`.

evnchn:
> keep track of the last color that has been set via `set_*_color` or binding?

While that would work, any user error which sets color using non-official means would totally stop this new color setter and binding from working. Is that a good idea?

evnchn:
> wouldn't expect binding/setter cleanup to handle their manual additions

I am ready to embark on making the cleanup work, but only if users raise the need. Let's merge, release, and see.

PR #5516, opened 2025-11-27
Streamable pages with `yield`

evnchn:
### Motivation

The gist of pipeline parallelization is that, should a task involve several steps which _could_ have been in parallel but must be completed serially for each task, we want to break them down so that the parallelization takes place. 

- There are multiple steps for rendering a NiceGUI page: Page function execution, network transmission
- But the page function is monolithic and the respinse is also monolithic 😞 

This PR challenges the status quo. 

### Implementation

Even if nothing is done, there is already speedup, because the template is broken into 2, and now we have 3 parts:

- Before elements string
- elements string
- After elements string

On top, this code is now possible:

```py
@ui.page('/')
async def main():
    for _1 in range(5):
        for _2 in range(10000):
            ui.label('1')
        await asyncio.sleep(0.1)
        yield
```

Here, we have 5 chunks of 10000 elements, each generated 0.1 second apart. 

**Fast 4G**

Before this PR: 2.38s 
After this PR: 1.73s

### Progress

- [ ] The implementation is complete. (I'm not sure)
- [ ] Pytests are very necessary and pending. 
- [ ] Documentation is necessary and pending. 
- [x] While the math checks out, we should do more testing to see if the time savings are worth it in real life. (2.38 vs 1.73)
- [x] Do we need explicit `stream=True` or can we infer?
- [x] How does this compare to just `await client.connected()` then push content changes through the outbox?
  - Push via outbox will lead to rendering and is slow that way.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -196,7 +196,6 @@ class Client:
                     id: _strip_children(element._to_dict()) for id, element in self.elements.items() if id not in serialized_elements  # pylint: disable=protected-access
                 }).translate(HTML_ESCAPE_TABLE)[1:-1]  # strip {}
                 serialized_elements.update(self.elements.keys())
-                print(serialized_elements)
                 return retval
 
             yield templates.get_template('index1.html').render(**context)


evnchn:
Erm, with https://github.com/zauberzeug/nicegui/pull/5516/commits/2f3e1230e90f1a1e965cf0ca41f2c03ce5bf3302, the code doesn't work: because of the children...

evnchn:
@falkoschindler Hmm I'm considering the user side of things: 

- Will this PR encourage further creating bulkier and bulkier pages instead of doing what you should do: pagination and lazy-loading?
- While our architecture works with a typical page size, we can't say for sure if the page is many times bigger than what it is now...

Considering the risks to both good user habits and computational pressure on our library code, let's not pursue this PR as it is. This means no more `yield` in userspace function. 

But if that is gone, then breaking the page into 3 chunks we aren't saving _that_ much time, so maybe we should scrap this altogether?

falkoschindler:
Yes, I don't think the additional complexity is worth it. In a situation like your example, you can already stream page content in chunks after awaiting `client.connected()`:
```py
@ui.page('/')
async def main():
    for _1 in range(5):
        for _2 in range(10000):
            ui.label('1')
        await asyncio.sleep(0.1)
```
In less extrem scenarios the benefit of `yield` or 3 chunks is probably marginal. So I don't think we should confuse the users with yet another way to write page functions (at the moment) and spare us more business logic in page.py and client.py.

Issue #5479, opened 2025-11-21
Interactive_image svg element sizing problems when in grid


daniel-anderberg:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import events, ui

def mouse_handler(e: events.MouseEventArguments):
    color = 'SkyBlue' if e.type == 'mousedown' else 'SteelBlue'
    ii.content += f'<circle cx="{e.image_x}" cy="{e.image_y}" r="15" fill="none" stroke="{color}" stroke-width="4" />'
    ui.notify(f'{e.type} at ({e.image_x:.1f}, {e.image_y:.1f})')

src = 'https://picsum.photos/id/565/640/360'
with ui.grid(columns=2,rows=2).classes('size-100'):
    ii = ui.interactive_image(src, on_mouse=mouse_handler, events=['mousedown', 'mouseup'], cross=True)

ui.run()
```

### Description

When I place an interactive_image in a sized grid, the image and the svg layers are not resized eqaually. The example code is a very lightly modified example from the interactive_image documentation. In this case the result is that the svg drawing does not follow the mouse pointer.

### NiceGUI Version

3.3.1

### Python Version

3.12.3

### Browser

Firefox

### Operating System

Linux

### Additional Context

Have also tested with chromium. Same result.

evnchn:
Hello @daniel-anderberg 

Sorry I don't know enough about browser CSS to get to why it currently breaks, but I do have a hotfix for you. Hope it helps!

```py
from nicegui import events, ui


def mouse_handler(e: events.MouseEventArguments):
    color = 'SkyBlue' if e.type == 'mousedown' else 'SteelBlue'
    ii.content += f'<circle cx="{e.image_x}" cy="{e.image_y}" r="15" fill="none" stroke="{color}" stroke-width="4" />'
    ui.notify(f'{e.type} at ({e.image_x:.1f}, {e.image_y:.1f})')


src = 'https://picsum.photos/id/565/640/360'
with ui.grid(columns=2, rows=2).classes('size-100'):
    with ui.element('div'):  # HOTFIX HERE
        ii = ui.interactive_image(src, on_mouse=mouse_handler, events=['mousedown', 'mouseup'], cross=True)

ui.run(show=False)
```

[...]

PR #5517, opened 2025-11-27
Make sure the crosshair overlay matches the image

falkoschindler:
### Motivation

In #5479 we noticed that the crosshair of a `ui.interactive_image` with non-default aspect ratio is displayed incorrectly:
```py
ui.interactive_image('https://picsum.photos/id/565/640/360', on_mouse=ui.notify, cross=True).classes('size-100')
```

### Implementation

This PR simply adds `width:100%;height:100%` and `preserveAspectRatio="none"`.

### Progress

### DIFF
--- a/nicegui/elements/interactive_image.js
+++ b/nicegui/elements/interactive_image.js
@@ -10,7 +10,7 @@ export default {
         v-on="onUserEvents"
         draggable="false"
       />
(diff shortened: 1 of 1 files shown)


evnchn:
btw I didn't just test the display: Checked the coordinates, they are reasonable as well.

Issue #5529, opened 2025-12-01
query_parameters not returned with wildcard routing


davetapley:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui
from nicegui.page_arguments import PageArguments


def main():
    ui.link('Go to item with query', '/item?color=red')
    ui.label('Main page')


def item_handler(args: PageArguments):
    ui.label(str(args))
    ui.link('Back', '/')


ui.sub_pages({
    '/': main,
    '/item': item_handler
}, show_404=False)

ui.run()
```

### Description

The wildcard routing introduced on #5440 doesn't populate `PageArguments` `query_parameters`

### NiceGUI Version

3.2.0 with #5440 cherry-picked on top while I wait for 3.4

### Python Version

Python 3.12.2

### Browser

Other

### Operating System

Linux

### Additional Context

Already fixed on my branch, PR incoming...

evnchn:
#5541 fixed it

PR #5530, opened 2025-12-01
Capture query_parameters with wildcard routing

davetapley:
### Motivation

- https://github.com/zauberzeug/nicegui/issues/5529

### Implementation

Since a URL can only have query params once, we collect it as we go and apply to `match` when found.

I also added coverage for `show_404` behavior to `test_sub_page_with_query_parameters` while testing, just to be sure it's specifically wildcard routing causing the problem (it is).

### Progress

### DIFF
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -4,7 +4,7 @@ import asyncio
 import inspect
 import re
 from typing import Any, Callable
-from urllib.parse import urlparse
+from urllib.parse import urlparse, ParseResult
 
 from starlette.datastructures import QueryParams
 from typing_extensions import Self
@@ -147,6 +147,7 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
 
     def _find_matching_path(self) -> RouteMatch | None:
         match: RouteMatch | None = None
+        query_params: QueryParams | None = None
         relative_path = self._router.current_path[len(self._root_path or ''):]
         if not relative_path.startswith('/'):
             relative_path = '/' + relative_path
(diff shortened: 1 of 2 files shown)


evnchn:
While it works, can the PR be achieved without changing the function signature of `def _match_route(self, path: str) -> RouteMatch | None:`?

evnchn:
How about this? 

```py
    def _find_matching_path(self) -> RouteMatch | None:
        match: RouteMatch | None = None
        relative_path = self._router.current_path[len(self._root_path or ''):]
        if not relative_path.startswith('/'):
            relative_path = '/' + relative_path
        segments = relative_path.split('/')
        query_params: QueryParams | None = None
        while segments:
            path = '/'.join(segments)
            if not path:
                path = '/'
            match, query_params = self._match_route(path, query_params)
            if match is not None:
                match.remaining_path = urlparse(relative_path).path.rstrip('/')[len(match.path):]
                break
            segments.pop()
        return match

    def _match_route(self, path: str, past_query_params: QueryParams | None) -> tuple[RouteMatch | None, QueryParams | None]:
        parsed_url = urlparse(path)
        path_only = parsed_url.path.rstrip('/')
        query_params = past_query_params or (QueryParams(parsed_url.query) if parsed_url.query else QueryParams())
        fragment = parsed_url.fragment
        if not path_only.startswith('/'):
            path_only = '/' + path_only

        for route, builder in self._routes.items():
            parameters = self._match_path(route, path_only)
            if parameters is not None:
                return RouteMatch(
                    path=path_only,
                    pattern=route,
                    builder=builder,
                    parameters=parameters,
                    query_params=query_params,
                    fragment=fragment,
                ), query_params
        return None, query_params
```

falkoschindler:
@evnchn @davetapley Are you still pondering about this PR? I'm thinking about merging it as one of the last PRs for 3.4 (since it fixes something which will be introduced in 3.4). But it would certainly make sense for @rodja to have a quick review. Or we postpone it to 3.5 so that 3.4 with its critical bugfix isn't held back.

evnchn:
Functionality wise there's no question (current implementation tosses the query string when trying to do the "one step back" checking of parent)

Question is, if we want to, and if so how might we, keep the diff smaller. 

I just didn't commit https://github.com/zauberzeug/nicegui/pull/5530#issuecomment-3600294963 because I'm not sure if everyone's on board. But you can look at it now 😉 

According to my memory, @davetapley is sporadic in terms of availability. I think we can proceed directly.

falkoschindler:
@evnchn Your code seems to work and indeed reduces the diff. Unfortunately I'm not allowed to push to this PR. 🫤

evnchn:
> not allowed to push to this PR

@davetapley sorry this PR un-mergable at this moment due to failing tests. Your attention is much appreciated!

falkoschindler:
In order to move forward and fix this bug before it gets released with #5440, I created PR #5541 as an alternative to this PR.

evnchn:
We went for #5541 in the end in the interest of time. @falkoschindler please be sure to credit @davetapley for the code. 🙏🏻

PR #5541, opened 2025-12-04
Capture query_parameters with wildcard routing

falkoschindler:
### Motivation

As noticed in #5529, the new wildcard routing drops query parameters:

```py
def item_page(args: PageArguments):
    ui.label(f'Query parameters: {args.query_parameters}')

ui.sub_pages({
    '/': lambda: ui.link('Go to item 123 with color red', '/item/123?color=red'),
    '/item': item_page,
}, show_404=False)
```

After clicking the link, we should see "color=red", but we don't.

### Implementation

As an alternative to PR #5530, this PR follows @evnchn's suggestion from https://github.com/zauberzeug/nicegui/pull/5530#issuecomment-3600294963.

### Progress

### DIFF
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -151,21 +151,22 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
         if not relative_path.startswith('/'):
             relative_path = '/' + relative_path
         segments = relative_path.split('/')
+        query_params: QueryParams | None = None
         while segments:
             path = '/'.join(segments)
             if not path:
                 path = '/'
-            match = self._match_route(path)
+            match, query_params = self._match_route(path, query_params)
             if match is not None:
                 match.remaining_path = urlparse(relative_path).path.rstrip('/')[len(match.path):]
                 break
             segments.pop()
         return match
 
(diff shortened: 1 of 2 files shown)


nicegui/elements/sub_pages.py
```diff
@@ -151,21 +151,22 @@ def _find_matching_path(self) -> RouteMatch | None:
...
                 break
             segments.pop()
         return match
 
-    def _match_route(self, path: str) -> RouteMatch | None:
+    def _match_route(self, path: str, query_params: QueryParams | None) -> tuple[RouteMatch | None, QueryParams | None]:
         parsed_url = urlparse(path)
         path_only = parsed_url.path.rstrip('/')
-        query_params = QueryParams(parsed_url.query) if parsed_url.query else QueryParams()
+        query_params = query_params or QueryParams(parsed_url.query)
```
evnchn, line 169: Went the extra mile to verify QueryParams is same if passed nothing VS passed an empty string. 

```py
from starlette.datastructures import QueryParams
print(QueryParams() == QueryParams(''))
```

Prints `True`, so the original code could have been simply `query_params = QueryParams(parsed_url.query)` and the new code is valid.

Issue #4985, opened 2025-07-26
"pop from empty list" exception when using long-running contexts in io_bound


MaidScientistIzutsumiMarin:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from functools import partial
from time import sleep

from nicegui import ui
from nicegui.run import io_bound


def run() -> None:
    with row:
        sleep(10)


row = ui.row()
ui.button(on_click=partial(io_bound, run))
ui.run(reload=False)
```

### Description

Start the app.
The browser starts with a button.
Pressing the button will cause the following error: `IndexError: pop from empty list`

### NiceGUI Version

2.21.1

### Python Version

Python 3.13.5

### Browser

Other, Firefox

### Operating System

Windows

### Additional Context

Functionality seems to be unaffected.

This error happens on native and the Firefox browser. We have not tested other browsers. The length of time spent in the context manager in `run()` seems to determine if the error will happen. Testing with 5 seconds only sometimes has an error, while testing with 10 seconds seems to always cause it on our end.

This error also happens when performing other long-running processes in this fashion, even without using `time.sleep()`.

rodja:
Thanks for reporting @MaidScientistIzutsumiMarin! The problem seems to be caused by `Slot.prune_stacks` which was introduced in [NiceGUI 1.0 to avoid a memory leak](https://github.com/zauberzeug/nicegui/issues/182). If the pruning is monkey patched the error is gone:

```py
import time
from nicegui import app, run, slot, ui

async def dummy() -> None:
    pass

# without this line, we get "IndexError: pop from empty list"
slot.Slot.prune_stacks = dummy

def task() -> None:
    with row:
        time.sleep(10)
    print('done')

row = ui.row()
app.on_startup(run.io_bound(task))

ui.run()
```

@falkoschindler do you have an idea why the stack is pruned?

falkoschindler:
@rodja The stack is pruned to avoid the dictionary of slot stacks to grow infinitely. If a task is finished or cancelled, the (possibly empty) stack isn't needed anymore, so it can be removed from the dictionary.

[...]

PR #5546, opened 2025-12-05
Treat task id 0 as always running for Slot

ftilde:
Previously, a long running (or unlucky) callback running with a Slot context outside of an async task could have its stack pruned between `__enter__` (pushing on the stack) and `__exit__` (poping from the stack). The latter would find the stack missing or empty and would crash.

### Motivation

Consider the following MRE (courtesy of @falkoschindler):
```python
PING = Event[[]]()

app.timer(1, lambda: run.io_bound(PING.emit), once=True)

@ui.page('/')
def page():
    PING.subscribe(lambda: time.sleep(11))
```
Running this will always crash with the following stacktrace:
```
  File "/home/dominik/git/nicegui-test/.venv/lib/python3.12/site-packages/nicegui/event.py", line 145, in _invoke_and_forget
    result = callback.run(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dominik/git/nicegui-test/.venv/lib/python3.12/site-packages/nicegui/event.py", line 34, in run
    with (self.slot and self.slot()) or nullcontext():
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dominik/git/nicegui-test/.venv/lib/python3.12/site-packages/nicegui/slot.py", line 39, in __exit__
    self.get_stack().pop()
IndexError: pop from empty list
```

The issue is that the task is run without an active task, resulting a the task id of 0 within the Slot context. Entering the context will push an entry to the stack associated with id 0, but exiting will find the stack missing (or recreated and empty) since the periodic pruning task (running every 10 seconds) has not treated id 0 as associated with a running task and pruned the corresponding stack.

Notably, this is the same underlying issue as in #4985, but while there it was decided to be unsupported to create ui elements outside of the corresponding context, here it surfaced by simply emitting an event.

### Implementation

We simply add task id 0 to the list of running tasks as a special case. This will mean that the associated stack is never pruned, but the memory cost of this is miniscule.

### Progress

### DIFF
--- a/nicegui/slot.py
+++ b/nicegui/slot.py
@@ -61,13 +61,12 @@ class Slot:
     async def prune_stacks(cls) -> None:
         """Remove stale slot stacks."""
         try:
-            running = [id(task) for task in asyncio.tasks.all_tasks() if not task.done() and not task.cancelled()]
+            running = {id(task) for task in asyncio.tasks.all_tasks() if not task.done() and not task.cancelled()}
 
-            # Id 0 is the special id used in cases where no task is active (see
-            # get_task_id). Since it has no associated task, we always assume
-            # it to be active in order to avoid pruning the associated stack
-            # while the Slot is active.
-            running.append(0)
+            # ID 0 is the special ID used in cases where no task is active (see get_task_id).
+            # Since it has no associated task, we always assume it to be active
+            # in order to avoid pruning the associated stack while the slot is active.
+            running.add(0)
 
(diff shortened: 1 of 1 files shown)


evnchn:
Just in case, I checked and `id()` function will not return 0 for CPython and PyPy, so there's no risk of false positive. 

Though this is an implementation detail, I take it that this is an established norm.

PR #5547, opened 2025-12-05
Add extras to dev dependencies

falkoschindler:
### Motivation

After checking out the repository, a contributor had to
```bash
uv sync --all-extras
```
before being able to run things like
```bash
uv run pytest
```

It would be more intuitive if all extras also were dev dependencies so that `uv run ...` just works.

### Implementation

Due to lack of a better solution, this PR simply copies all extra dependencies into the "dev" dependency group. Additional comments should help to keep both blocks in sync.

Now we can directly run
```bash
uv run pytest
```

### Progress

- [x] CONTRIBUTING.md has been updated.

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -30,6 +30,8 @@ jobs:
           cache-dependency-glob: "uv.lock"
       - name: Install Python
         run: uv python install ${{ matrix.python }}
+      - name: Install dependencies
+        run: uv sync
       - uses: nanasess/setup-chromedriver@v2.3.0
       - name: ${{ matrix.python }} / ${{ matrix.scope }}
         if: matrix.scope == 'pytest'


codingpaula (dismissed):

evnchn:
Makes sense. 

Bonus item for later: add some checks somewhere (pre-commit? pipeline?) to check if all extras are in dev, just in case forgot.

evnchn:
All startup tests are dead?

https://github.com/zauberzeug/nicegui/actions/runs/19987211219

falkoschindler:
Ooops, I removed the "Install dependencies" step because I thought `uv run pytest` is enough. But this is only for `matrix.scope == 'pytest'`. The other scope needs those dependencies. I'll fix it.

evnchn:
The test failed in a non-conventional way which smells like #5119 / #5090 all over again. 

While we can get away with merging this PR via re-running the tests, I think we should look into it further. 

https://github.com/zauberzeug/nicegui/actions/runs/19988169807/job/57325066289

Issue #4748, opened 2025-05-15
support in JSONEditor schema validations 'format' properties


phnmn:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from typing import Union
from uuid import UUID

from nicegui import ui
from pydantic import BaseModel, Field


class Schema(BaseModel):
    client_session_id: Union[UUID, None] = Field(default=None)


ui.json_editor(
    {
        "content": {"json": ""},
        "readOnly": True,
        "mode": "text",
    },
    schema=Schema.model_json_schema(),
)

ui.run()
```

### Description

When the JSON schema contains format: uuid, the JSON editor fails to work and displays an error in the browser console
`vue.global.prod.js:5 Error: unknown format "uuid" ignored in schema at path "#/properties/client_session_id/anyOf/0"
    at format.js:63:27
    at format.js:50:17
    at Object.code (format.js:22:13)
    at $ (index.js:464:13)
    at index.js:222:17
    at e.value (index.js:439:13)
    at e.value (index.js:568:18)
    at x (index.js:219:9)
    at p (index.js:200:13)
    at index.js:192:13
(anonymous) @ vue.global.prod.js:5
content.bundle.js:55109 will try to open idb now: content
`
I think this can be fixed by using [ajv-formats](https://github.com/ajv-validator/ajv-formats) for the AJV validator.

### NiceGUI Version

2.17.0

### Python Version

Python 3.11.12

### Browser

Chrome

### Operating System

macOS

evnchn:
@phnmn Kindly check out #4749. You can early-enjoy the benefits of that PR if you edit the source code of your local NiceGUI install (best if you do this in a virtual environment like venv or Docker so that you don't mess up your global NiceGUI installation)

evnchn:
Sorry but the approach of #5556 will not help since it is not something that we can solve by simply manipulating the `await import("nicegui-json-editor")` instance (I think)

@phnmn the reason this is stuck is because `ajv-formats` tough to pack into ESM format, for some reason. While jsdelivr manages to do it, we are a bit hesitant on pulling files from there instead of doing the normal procedure of rollup.

falkoschindler:
[...]

PR #5559, opened 2025-12-08
expandable ajv by ajv_addins parameter

evnchn:
### Motivation

Owing to [just so many addins and extensions for AJV](https://www.perplexity.ai/search/now-i-know-that-for-ajv-i-can-oK0LuUMJQgec3N93IG5BJQ) I believe it is not feasible to carry-through the "see something? Add it to NiceGUI" approach in #4749, especially with shenanigans such as [tough to pack `ajv-formats` into our own ESM module](https://github.com/zauberzeug/nicegui/issues/4748#issuecomment-3628309903)

But we still want those extensions...

### Implementation

This is a "batteries not included" PR but it enables the use of AJV addins. 

Users are supposed to "bring their own addins" then register using `ajv_addins` parameter. 

Example code:

```py
from nicegui import ui

ui.add_head_html('''
<script type="module">
import ajvFormats from 'https://cdn.jsdelivr.net/npm/ajv-formats@3.0.1/+esm';
window.ajvFormats = ajvFormats;
</script>''')

schema = {
    'type': 'object',
    'properties': {
        'uuid': {
            'type': 'string',
            'format': 'uuid',  # breaks on main, not on this PR
        },
    },
    'required': ['uuid'],
}
data = {
    'uuid': '123e4567-e89b-12d3-a456-426614174000',
}
ui.json_editor({'content': {'json': data}}, schema=schema, ajv_addins=['ajvFormats'])

ui.run()
```

### Progress

- [ ] Pytests are pending. 
- [ ] Documentation is needed (several showcases, I'd imagine).

### DIFF
--- a/nicegui/elements/json_editor/json_editor.js
+++ b/nicegui/elements/json_editor/json_editor.js
@@ -22,7 +22,14 @@ export default {
   methods: {
     checkValidation() {
       if (this.schema !== undefined) {
-        this.properties.validator = createAjvValidator({ schema: this.schema, schemaDefinitions: {}, ajvOptions: {} });
+        this.properties.validator = createAjvValidator({
+          schema: this.schema,
+          schemaDefinitions: {},
+          ajvOptions: {},
+          onCreateAjv: (ajv) => {
+            for (const ajvAddin of this.ajvAddins) eval(ajvAddin)(ajv);
+          },
+        });
       }
     },
     update_editor() {
@@ -49,5 +56,6 @@ export default {
   props: {
     properties: Object,
     schema: Object,
+    ajvAddins: Array,
   },
 };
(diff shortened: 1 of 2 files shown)


falkoschindler:
@evnchn Just to be clear: You think this approach is better than including the batteries? Or is it just too hard (or impossible) to bundle ajv-formats?

evnchn:
1. IMO this approach is better than the batteries because there are so many batteries. 
2. I can't do the other approach. As said I have tried many times to pack `ajv-formats` but I just can't do it for some reason.

evnchn:
Sorry this is bad idea. Because the batteries are bad. 

For the 5th time let's inspect https://cdn.jsdelivr.net/npm/ajv-formats@3.0.1/+esm

image

It tries to load in another copy of Ajv, which is not good because that is not the same as the one internally we use (we have 8.17.1), and this fact alone may cause compatibility issues. 

Moreover, the syntax is kinda sketchy. 

On top of that, we need an external web request. 

I think @falkoschindler pressed 👍 on the basis of point 2: I can't do the other approach. Well I have done it (PR coming up). Please close this PR if the other one is better (I think it will be).

evnchn:
The more I look the more I don't like this...

PR #5562, opened 2025-12-08
Element bundling

evnchn:
### Motivation

While checking the networks tab under the 1000ms latency infinite bandwidth profile, one fact is clear: **there is max 6 outgoing connections**

So, it would seem to me that it is a bit wasteful to spread the many NiceGUI elements into separate internet fetches. 

### Implementation

The code, in the end, facilitates: 

```js
import { dark_mode, dialog, header, image, input, interactive_image, intersection_observer, keyboard, link, markdown, select, sub_pages } from "/_nicegui/3.3.1.post34.dev0+126d78ec/component_pack/_/0d1c4bb62d23bfa495950af74dac3f84:interactive_image.js,311b3a916364748cbafd1ec125726dc6:link.js,38e19a26114c7edb90c14be093c9f2a6:image.js,51ab70dc0ec662d81efaab63af1fad5b:input.js,7cdbee5c631d8f004a9e106a2ae159cf:intersection_observer.js,8dfd5dd265260e79ad82c16e12c1b00a:dialog.js,8f15f776281fe21ebff32ce6748dc6ec:dark_mode.js,97777276820998629fdce5a3f2dd9aa7:select.js,c9d39871fffac82064d23f93001b6f90:header.js,d3d4f0d4a388c8e3671bde2b9049c742:markdown.js,f21f17e318a37dbed449582b131ee2ea:sub_pages.js,f36d55691403482aed2b65ee6a1c6820:keyboard.js";
```

Rather than doing `import { default as BLAH } from "URL"` N times for N elements. 

### Progress

  - Not a good idea to just roll out this PR since the element bundle will change even if we include one more or one less element
  - [x] The element bundle is created by considering the elements used in the `'/'` page, in an increase-only manner to combat against flip-flop in case the homepage "sometimes show an element"
  - [x] We write to `loaded_components` JS-land variable by Python-land directly. 
- [x] Need to add pytest for, at least, mermaid and markdown with mermaid, knowing from prior experience that those two are the "first to fail" if our JS component implementation isn't right, to ensure that those two elements work perfectly whether they are in the bundle or not in the bundle. https://github.com/zauberzeug/nicegui/pull/5562/commits/262faed76c25b8f35de0d1634dce11d1682cbd12
- [x] Documentation turns not necessary: just an implicit non-configurable performance improvement under-the-hood

### Results

Before:

image

After:

image

Which is **expected**, since we shoved 12 requests into 1, and so we should be 2 seconds faster because we save ourselves 2 rounds of having to wait 1 second for the last 6 requests to finish.

### DIFF
--- a/tests/test_speculative_loading.py
+++ b/tests/test_speculative_loading.py
@@ -123,7 +123,7 @@ def test_prefetch_connects_after_navigation(screen: Screen, event_log: EventLog)
 
 
 def add_speculation_rule(url: str, *, kind: str = 'prerender') -> None:
-    rules = {kind: [{'source': 'list', 'urls': [url], 'eagerness': 'eager'}]}
+    rules = {kind: [{'source': 'list', 'urls': [url], 'eagerness': 'immediate'}]}
     script = '<script type="speculationrules">' + json.dumps(rules) + '</script>'
     ui.add_head_html(script)
 


evnchn:
Thoughts:

1. `ui.run` can add option: `homepage='/'` because (1) we will need this for SEO in the future, (2) when we serve the homepage we will pack the element bundle and then all the other pages will get the bundle from homepage. 
2. On the bundling set selection, prevent the edge case that homepage have dynamically-added elements, the bundle set can only grow bigger, not smaller, to avoid wasteful caching in back-and-forth situation.

evnchn:
Slow 4G: 5.06s -> 4.05s

Reasonable since the network latency of Slow 4G is 562.5ms https://stackoverflow.com/a/77173765 so we are saving 2 roundtrips essentially.

evnchn:
Further got Slow 4G down to 3.63s. 

Graphical before-and-after:

Before: 

image

After:

image

**Note the differing x-axis** and recall that the "lighter-yellow" part is due to overhead of network latency which we **don't** want because that is just waiting around without transferring data.

evnchn:
I can pack the CSS as well, will become something like:

```css
@layer base {
  {{ base_css | safe }}
}
@layer quasar {
  {{ quasar_css | safe }}
}
@layer quasar_importants {
  {{ quasar_importants_css | safe }}
}
@layer nicegui {
  {{ nicegui_css | safe }}
}
@layer quasar {
  blockquote, dd, dl, figure, h1, h2, h3, h4, h5, h6, hr, p, pre {
    margin: revert-layer; /* HACK: revert Quasar's margin styles */
  }
}
@layer overrides {
  {{ headwind_css | safe }}
  .column.inline, .flex.inline, .row.inline {
    display: inline-flex;  /* HACK: re-apply Quasar's inline styles to override Tailwind */
  }
  .nicegui-dialog-open,
  .nicegui-select-popup-open {
    scroll-behavior: auto !important; /* HACK: avoid smooth-scrolling when dialogs or select popups close */
  }
}
```

Shaves 300ms, and at this point the longest request is now the Quasar Socket NiceGUI Vue 4-in-1, which we may consider breaking up again to parallelize the download 🤯

falkoschindler:
I just chatted a bit with GPT 5.2 to make myself familiar with this PR. Make some of the mentioned concerns are worth investigating:

> If I put my “maintainer hat” on: the core idea is valid, but I would be cautious about merging this PR *as-is* (especially as a silent, always-on change), because it mixes a solid optimization with a few brittle/mechanically-risky choices.
> 
> ### What the PR really buys you (even with caching and HTTP/2)
> 
> This isn’t “just transport”: it changes the module graph from “N module fetches” to “1 fetch that re-exports N things”. The PR’s motivation is specifically the browser’s visible “6 outgoing connections” behavior under a high-latency profile, and the improvement claim is essentially “we saved rounds”. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562)][1])
> 
> Even on HTTP/2, there can still be a win because:
> 
> * fewer request/response headers and scheduling overhead
> * fewer module parse/instantiate steps (module bookkeeping isn’t free)
> * simpler critical path (one dependency instead of many)
> 
> But the huge wins shown in the PR description are mostly a *cold cache + big RTT* story. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562)][1])
> 
> ### The parts I like
> 
> 1. **A single endpoint that can return a pack module**
>    The patch adds a `/_nicegui/<version>/component_pack/_/<keys>` endpoint that concatenates component sources into one module by rewriting `export default` into named exports (`pack_<name>`). ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562.patch)][2])
>    That’s a straightforward mechanism and it’s easy to reason about from a networking perspective.
> 
> 2. **Bundling “always-loaded” built-in Vue components**
>    Later in the patch, it moves toward always packing the Vue components into the packset. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562.patch)][2])
>    This is the highest-confidence subset: it’s stable across pages and (likely) across most apps.
> 
> ### The parts that make me nervous as a maintainer
> 
> 1. **Very long URLs as the “pack manifest”**
>    The pack URL literally contains a comma-separated list of component keys (with `/` replaced by `:`). ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562)][1])
>    That can run into practical limits (reverse proxies, CDNs, WAFs, even some servers) and it makes caching behavior more chaotic than it needs to be.
> 
> If you want this approach, I’d strongly prefer a *short* cache key:
> 
> * `/component_pack/<hash>.js` where `<hash>` is computed from the sorted key list, with the server keeping a map `hash -> list-of-keys`
> * optionally plus an LRU / size cap and a max “keys per pack” guardrail
> 
> 2. **The “latching” behavior tied to homepage traffic**
>    The PR’s current strategy is “pack components based on what’s used on `/`, and only grow the set to avoid flip-flop”. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562)][1])
>    That makes performance dependent on which page users hit first, and it introduces global-ish state semantics that can surprise people (and complicate multi-worker setups).
> 
> If you keep bundling, I’d rather see one of these:
> 
> * a fixed “core pack” (built-ins) + per-page incremental loads
> * or a deterministic “pack policy” (e.g., “pack built-ins + top K most common components”) decided at startup/config time, not by runtime traffic
> 
> 3. **Rewriting `export default` by string replacement**
>    The endpoint uses a simple `replace('export default', ...)` approach. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562.patch)][2])
>    That’s the kind of thing that works for 95% of your own components… until you hit the 5% (comments/strings, formatting, different export styles, generated code, etc.). You already see the PR author calling out “mermaid and markdown … first to fail” as a test focus. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562)][1])
> 
> If you merge bundling, I’d want:
> 
> * a stricter rewrite (at least regex anchored to actual JS syntax line starts)
> * or (better) build-time generation of packable wrappers for built-in components so runtime bundling doesn’t need to parse/transform JS at all
> 
> 4. **Caching headers and compute cost**
>    The pack endpoint returns dynamically generated JS. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562.patch)][2])
>    To make this worthwhile broadly, you’ll want:
> 
> * strong cache headers (because the URL is versioned anyway)
> * server-side memoization of generated pack content (don’t re-read/concat files every time)
> 
> ### My practical judgement
> 
> * **As a concept:** yes, this can be worth it for “NiceGUI out-of-the-box” deployments where people run behind HTTP/1.1, on mobile networks, behind VPNs, etc. The user benefit is real in those worst-case conditions. ([[GitHub](https://github.com/zauberzeug/nicegui/pull/5562)][1])
> * **As currently designed:** I wouldn’t merge it as a silent always-on behavior yet. The URL-length/caching-key issue and the runtime rewrite brittleness are the two big “support burden” multipliers.
> 
> ### A merge-friendly direction (if you want the win without the risk)
> 
> If you want a version I’d be comfortable shipping broadly:
> 
> 1. Ship a **stable core pack** (built-in always-needed components) with a **short URL**.
> 2. Keep per-element components as separate modules (or add optional per-page packs behind a flag).
> 3. Add **strong caching + memoization** for the pack response.
> 4. Harden the export rewrite (or eliminate it for built-ins via pre-generated wrappers).
> 
> [1]: https://github.com/zauberzeug/nicegui/pull/5562 "Element bundling by evnchn · Pull Request #5562 · zauberzeug/nicegui · GitHub"
> [2]: https://github.com/zauberzeug/nicegui/pull/5562.patch "patch-diff.githubusercontent.com"

falkoschindler:
Given the additional complexity and downsides mentioned above, we're not sure if this is the right path. Usually proxies would achieve simultaneous downloads via HTTP/2, so this PR mainly fixes cases where HTTP/2 is not in use, e.g. a local test setup (with artificially high latency).

Anyway, it would be interesting to see a comparison with more realistic latencies like 100ms. Is the benefit still in the order of seconds?

evnchn:
> more realistic latencies like 100ms

We can estimate right away. We shaved 1.43s with HTTP-level network latency assumed 562.5ms (so 2.5 round-trips)

If HTTP latency is 100ms then we would save circa 250ms. 

But then note that if the network speed up 5.625 times I expect the downlink to also be 5.625 times, so the complete story would more be 900ms to 650ms. 

---

But 2 critical observations in retrospect: 

- What I am doing in this PR only matters in initial-load without cache...
- BUT in future-load, our packed one-file code may actually hinder browser-side loading parallelizations, if any...

So perhaps lighthouse is too tunnel-visioned of a metric and pursuing it at the cost of library maintenance and future-load speed is not a good idea? 

Especially given risk of subtle incompatibilities in the components due to JS-rewrite and other decisions which come around it (like which components are packed). 

Kinda sad to say, but maybe we should scrap this.

falkoschindler:
> So perhaps lighthouse is too tunnel-visioned of a metric and pursuing it at the cost of library maintenance and future-load speed is not a good idea?

Yes, this resembles my feeling pretty well. Performance is great and I'd love to improve it where possible. But additional complexity is a significant burden in the long run. That's why I'm questioning such metrics to focus on PRs that are _really_ beneficial with reasonably complex implementations.

evnchn:
If someone needs this for HTTP1.1 needs, tell me and reopen I guess.

PR #5567, opened 2025-12-10
Have `ui.joystick` use `.js` component instead

evnchn:
### Motivation

Vue components have the problem in that they are always loaded into the HTML regardless of whether the element is on-sceeen because of the need of HTML and CSS. 

I don't quite like that, so this PR seeks to have `ui.joystick` use `.js` component instead

### Implementation

- Template is in the JS now. 
- Import nipplejs statically is possible (perf).
- Scoped style is achieved via targeting `.nicegui-joystick`, similar to the other elements. 
  - We already had `.nicegui-joystick` selector in `nicegui.css` so I figured it is no harm to add another one. 

### Progress

- [x] Existing pytest do us good. 
- [x] Documentation no need for a background improvement.

### DIFF
--- a/nicegui/elements/joystick/joystick.js
+++ b/nicegui/elements/joystick/joystick.js
@@ -1,11 +1,8 @@
-<template>
-  <div><div></div></div>
-</template>
+import { nipplejs } from "nicegui-joystick";
 
-<script>
 export default {
-  async mounted() {
-    const { nipplejs } = await import("nicegui-joystick");
+  template: "<div><div></div></div>",
+  mounted() {
     const joystick = nipplejs.create({
       zone: this.$el.children[0],
       position: { left: "50%", top: "50%" },
@@ -20,12 +17,3 @@ export default {
     options: Object,
   },
 };
-</script>
-
-<style scoped>
-:scope > div {
-  width: 100%;
-  height: 100%;
-  position: relative;
-}
-</style>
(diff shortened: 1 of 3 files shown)


evnchn:
Let's keep the discussion here:

I can't find it but @falkoschindler I think you mentioned that these elements "have not been converted yet for some reason". Now I am doing the converting, and it appears to work, but it may be better if we can get down to _why_ it has not be converted. 

Especially since `joystick.vue` was actually changed in 3.0 in https://github.com/zauberzeug/nicegui/commit/7762d62b362958a8f997bf07296a261e2fba93b8, and I figured that you are the kind that would have migrated joystick to `.js` as soon when you came around to the file.

falkoschindler:
@evnchn As far as I remember, these `.vue` components kind of serve as examples and help to validate that all component types still work. But this is a weak argument, as there should be independent tests to verify our dependency mechanism. So I like the idea of converting them to `.js` files.

We just have to check if there are any regression tests for `.vue` components. Our startup tests don't really check the page content, so our "Custom Vue Component" could be broken after changing something in dependencies.py and we won't notice.

evnchn:
> should be independent tests to verify our dependency mechanism; regression tests 

I think we should enhance https://github.com/zauberzeug/nicegui/blob/7320746795b7ec5fa609ab64e57e8abbf6238808/tests/test_vbuild.py to involve the screen fixture on that end.

PR #5571, opened 2025-12-12
Bake `ajv-formats` into `ui.json_editor`

evnchn:
### Motivation

This PR is in response to #4748 in which UUID schema is not supported. This requires `ajv-formats`, which this PR _nicely_ packs into NiceGUI as a permanent inclusion

### Implementation

Contrast to #5559 this is a very "batteries-included" PR, to the point it is not possible to opt-out of `ajv-formats`. 

It is done this way because the size increase isn't that large (+10% only) and we can do so much more with `ajv-formats` installed. 

I can tell you, the trick to install `@rollup/plugin-commonjs` and `@rollup/plugin-json` foils both ChatGPT, Grok, and Claude. 

By manually putting in the work, I prove that [I do not have skill issue](https://github.com/zauberzeug/nicegui/issues/4748#issuecomment-3628309903). 

### Demo code

```py
schema = {
    'type': 'object',
    'properties': {
        'uuid': {
            'type': 'string',
            'format': 'uuid',  # breaks on main, not on this PR
        },
    },
    'required': ['uuid'],
}
data = {
    'uuid': '123e4567-e89b-12d3-a456-426614174000',
}
ui.json_editor({'content': {'json': data}}, schema=schema)
```

image

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5571/commits/e6fd20bfd8f363a9ccf39b7884dfe8a38b674b3f
- [x] Documentation https://github.com/zauberzeug/nicegui/pull/5571/commits/0b00564eca4b67a983104d2f09108646e4f729ad

### DIFF
--- a/website/documentation/content/json_editor_documentation.py
+++ b/website/documentation/content/json_editor_documentation.py
@@ -45,7 +45,9 @@ def update_content():
     - `price` must be a number greater than 0
     - `uuid` must be a valid UUID (requires NiceGUI version 3.5.0 or higher)
 
-    *Added in version 2.8.0*, with formats from [ajv-formats](https://ajv.js.org/packages/ajv-formats.html) now supported *since version 3.5.0*
+    *Added in version 2.8.0*
+
+    *Updated in version 3.5.0: Added support for [ajv-formats](https://ajv.js.org/packages/ajv-formats.html)*
 ''')
 def schema_demo() -> None:
     schema = {


evnchn:
But we may still need some provision to load only some plugins which NiceGUI has packed into the library, as the number of plugins which we pack increases with user demand. 

Perhaps that can be a task for later, because we can have `ajv_plugins=['ajvFormats']` as the default to push that out non-breakingly.

nicegui/elements/json_editor/json_editor.py
```diff
@@ -13,7 +13,7 @@
 )
 
 
-class JsonEditor(Element, component='json_editor.js', esm={'nicegui-json-editor': 'dist'}):
+class JsonEditor(Element, component='json_editor.js', esm={'nicegui-json-editor': 'dist', 'nicegui-ajv-formats': 'dist2'}):
```
evnchn:
```suggestion
class JsonEditor(Element, component='json_editor.js', esm={'nicegui-json-editor': 'dist'}):
```

We don't need dist2

PR #5578, opened 2025-12-15
Introduce on_click for ui.echart

### LINKED ISSUE #5576: KeyError: 'seriesType' when clicking on a indicator item when using the "radar" plot in an "ui.echart"
### First Check

### Example Code

```python
def makeIndicator(name, max_value):
    return {"name": name, "max": max_value}
def makeLegendData(data):
    return {
        "name":data,
    }
def makeRadar(data,legend):
    dataObj={
                "name": str(legend),
                "value": data.tolist()
            } 
    return dataObj
def makeSeries(datas,legend):
    dataObj=list(range(len(datas)))
    for d in range(len(datas)):
        data=datas[d]
        dataObj[d]={
                "type": "radar",
                "radarIndex": d,
                "data":[makeRadar(data[i],legend[i]) for i,l in enumerate(legend)]
        }
    return dataObj
def point_clicked(e):
    print(e)
    ui.label(str(e))
def radial_chart(data,indicators,legend,title):
    """This function shows data as a radar plot
    Args:
        data (list of lists ([[],[]])): the data to be plotted, each number in a sublist corresponds to an indicator
        indicators (_type_): The labels that goes with each number in the sub-list of data
        legend (_type_): The instance of the data that the indicators are referring to
        title (_type_): The title of the plot
    Returns:
        _type_: A NiceGUI radar plot ui element 
    """
    lineStyle = {"width": 1,"opacity": 0.9}
    option = {
        "tooltip": {
            "triggerEvent":True,
            "confine":True,
            "enterable":True,
            "textstyle":{"overflow":"break"}
        },
        "backgroundColor": "#FFF",
        "title": {
            "text": title,
            "left": "center",
            "textStyle": {
                "color": "#333"
            }
        },
        "legend": {
            "triggerEvent":True,
            "tooltip":{},
            "bottom":5,
            "data": legend,
            "itemGap": 20,
            "textStyle": {
                "color": "red",
                "fontSize": 14
            },
            "selectedMode": "multiple",
            "coordinateSystem":"radar"
        },
        "radar": {
            "triggerEvent":True,
            "indicator": [makeIndicator(name, 100) for name in indicators[0]],
            "shape": "circle",
            "splitNumber": 5,
            "axisName": {
                "color": "rgb(0, 0, 0)"
            },
            "splitArea": {
                "show": True
            },
            "axisLine": {
                "lineStyle": {
                    "color": "rgba(238, 197, 102, 0.5)"
                }
            }
        },
        "series": makeSeries(data,legend)
    }
    # ui.label(str(option))
    return ui.echart(option,on_point_click=point_clicked,renderer='svg').classes('w-full h-128')

skills_have0=np.array([[77.0,50.0,50.0,74.0,50.0,74.,50,71,50,50,76,50,50],[67,73,50,50,50,50,89,50,50,50,50,50,50],[74,76,78,79,48,79,50,74,50,50,50,50,50]])
skills_label0=[['AI and Machine Learning ','Bioengineering', 'Business',
'Community Building', 'Electronics', 'Medicine', 'Robotics', 'Microbiology',
'Neuroscience', 'Teaching/Learning', 'Government', 'Quantum physics',
'Sustainable construction'],['AI and Machine Learning ','Bioengineering', 'Business',
'Community Building', 'Electronics', 'Medicine', 'Robotics', 'Microbiology',
'Neuroscience', 'Teaching/Learning', 'Government', 'Quantum physics',
'Sustainable construction'],['AI and Machine Learning ','Bioengineering', 'Business',
'Community Building', 'Electronics', 'Medicine', 'Robotics', 'Microbiology',
'Neuroscience', 'Teaching/Learning', 'Government', 'Quantum physics',
'Sustainable construction']]
names0=['Randomitity', 'Someone', 'Someone Else']
radial_chart([skills_have0],skills_label0,names0,"What skills do you currently have?")
```

### Description

I am, trying to implement a radar plot using the Apache ECharts ui function.  It largely is going well, but I would to make it so that you can click on an indicator (the text identifier on a radar plot, see example screenshot) and it triggers an event.  If I click on something that is actually part of the data, it successfully fires the event (again see screenshot).  I am not sure if this is a problem with the 

Here is what happens when I click on an indicator:
```
'seriesType'
  File "C:\Users\Someone\micromamba\envs\otherus\Lib\site-packages\nicegui\events.py", line 440, in handle_event
  File "C:\Users\Someone\micromamba\envs\otherus\Lib\site-packages\nicegui\elements\echart\echart.py", line 57, in handle_point_click
    series_type=e.args['seriesType'],
                ~~~~~~^^^^^^^^^^^^^^
KeyError: 'seriesType'
```
And here is what happens if I click on a some of the actual data points:
```
EChartPointClickEventArguments(sender=<nicegui.elements.echart.echart.EChart object at 0x000001C38D631480>, client=<nicegui.client.Client object at 0x000001C38D5756A0>, component_type='series', series_type='radar', series_index=0, series_name='series\x000', name='Randomitity', data_index=0, data={'name': 'Randomitity', 'value': [61, 50, 64, 50, 50, 68, 50, 63, 50, 50, 71, 50, 50]}, data_type=None, value=[61, 50, 64, 50, 50, 68, 50, 63, 50, 50, 71, 50, 50])
```
Radar Echarts Plot

### NiceGUI Version

3.4.0

### Python Version

3.14.2

### Browser

Firefox

### Operating System

Windows

### Additional Context

I tried reproducing the error using just javascript (pure Echarts) and the issue did not present itself:

```
    ...
var myChart = echarts.init(dom, null, {
  renderer: 'svg',
  useDirtyRect: true
}).on('click',function(e){console.log(e)});
   ...
```

evnchn:
### Motivation

Fix #5576, where in the case of clicking indicator label, all we have in `e.args` is key `componentType` and `name`, causing library code to throw an error. 

But #5577 no good because it shoehorns the component click into point click handler, which has a risk of breaking downstream user code expecteding to take a point from the handler args. 

### Implementation

We have to segregate: 

- All mandatory keys available for point click? Go to `on_point_click` handler with `EChartPointClickEventArguments`
- Any else case, go to `on_component_click` handler with `EChartComponentClickEventArguments`

### Progress

### DIFF
--- a/nicegui/events.py
+++ b/nicegui/events.py
@@ -65,7 +65,7 @@ class EChartPointClickEventArguments(UiEventArguments):
     series_name: str
     data_index: int
     data: float | int | str
-    data_type: str
+    data_type: str | None
     value: float | int | list
 
 


falkoschindler:
I refactored the code a bit to keep both events separate and the diff smaller.

A design decision is whether the new "click" or "component click" event should include point clicks. Conceptually I don't see a reason to exclude them because they _are_ component clicks after all. Therefore I removed an if condition.

And I'd rename "component click" to "click" - like the original JS event.

~~The client still emits a "pointClick" event. This is mostly for backward compatibility in case some user code relies on this internal event name. But I'm leaning towards renaming it to "click".~~ I chose the internal event name "componentClick".

Anyway, we should add tests and a demo.

falkoschindler:
I solved the bug directly on main, so this PR can focus on the new "click" event.

evnchn:
> Shoving None into non-None field

We are actually doing that right now:

```
EChartPointClickEventArguments(sender=<nicegui.elements.echart.echart.EChart object at 0xffff81e196d0>, client=<nicegui.client.Client object at 0xffff81d37a10>, component_type='series', name='Test', series_type='radar', series_index=0, series_name='series\x000', data_index=0, data={'name': 'Test', 'value': [77, 50, 90]}, data_type=None, value=[77, 50, 90])
```

data_type cannot be None but it is. So hence why I left it as is.

We should rework this in 4.0 ~~to increase the shared fields of inheritance, as well as~~ to make it more accurate typing-wise.

evnchn:
> inheritance

Hmm but is inheritance really necessary? 

- How often is the user passing a handler intended for `on_click` into `on_point_click`? 
- Isn't that user error? 
- Maybe we should _not_ enable that misuse? 

I am afraid that the inheritance will land us into tricky spot should EChart change something upstream. It is best to not tie two things together via inheritance if there is no solid benefit.

falkoschindler:
You're right! Let's remove the extra layer of inheritance!

I also changed the type of `data_type` to `str | None`. Although it is kind of a breaking change, I consider it a bugfix. If some user code would really assume `data_type: str`, it would break at runtime anyway. So it's better to communicate the possibility of `None` up front.

falkoschindler (approved):

Issue #5579, opened 2025-12-15
HTTPException for `root` function with `ui.sub_pages` when opening dev tools


falkoschindler:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

def root():
    ui.sub_pages({'/': lambda: ui.label('Home')})

ui.run(root)
```

### Description

When opening the Chrome dev tools after visiting the index page /, an exception with a long stack traces is logged:
<details>
<summary>
fastapi.exceptions.HTTPException: 404: /.well-known/appspecific/com.chrome.devtools.json not found
</summary>

```
404: /.well-known/appspecific/com.chrome.devtools.json not found
Traceback (most recent call last):
  File "/Users/falko/.pyenv/versions/3.11.11/lib/python3.11/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
    await app(scope, receive, sender)
  File "/Users/falko/.pyenv/versions/3.11.11/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/Users/falko/.pyenv/versions/3.11.11/lib/python3.11/site-packages/starlette/routing.py", line 716, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/falko/.pyenv/versions/3.11.11/lib/python3.11/site-packages/starlette/routing.py", line 766, in app
    await self.default(scope, receive, send)
  File "/Users/falko/.pyenv/versions/3.11.11/lib/python3.11/site-packages/starlette/routing.py", line 652, in not_found
    raise HTTPException(status_code=404)
starlette.exceptions.HTTPException: 404: Not Found

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/falko/.pyenv/versions/3.11.11/lib/python3.11/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/Users/falko/.pyenv/versions/3.11.11/lib/python3.11/site-packages/starlette/middleware/sessions.py", line 85, in __call__
    await self.app(scope, receive, send_wrapper)
[...]

PR #5580, opened 2025-12-15
Handle 404 gracefully without re-raising exception

falkoschindler:
### Motivation

In #5579 we noticed that a 404 (e.g. when opening the Chrome dev tools which requests com.chrome.devtools.json) causes an exception with a long stacktrace being printed to the console:

```py
from nicegui import ui

def root():

ui.run(root)
```

### Implementation

This PR avoids calling `create_error_page` which would re-raise the HTTP exception.
Instead it logs a warning and responds with a default error page.

### Progress

### DIFF
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -123,7 +123,7 @@ class page:
             except Exception as e:
                 core.app.handle_exception(e)
 
-        def create_error_page(e: Exception, request: Request) -> Response:
+        def create_500_error_page(e: Exception, request: Request) -> Response:
             page_exception_handler = core.app._page_exception_handler  # pylint: disable=protected-access
             if page_exception_handler is None:
                 raise e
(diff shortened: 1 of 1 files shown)


evnchn (dismissed):
@falkoschindler Excellent observation and fix. Can merge right away if urgent. 

---

Suspected root cause of the bug: Nature of `create_error_page` that it should not be called as a shorthand to `with client:` `error_content(...)` is not obvious especially given the name. Considered oversight of #4538 🙇 

**Proposed bonus fix (takes ~5mins, considerable)**: Rename the function to `_create_error_page` and add a docstring to avoid this from happening again. Especially in light of coding agents which may fall easily for this "trap" if they lack the domain knowledge.

rodja (approved):

falkoschindler:
I tried to find a good docstring, but came up with an even better result (I think):
The most important fact about `create_error_page` is that it deals with server errors, HTTP 500. Changing the name to `create_500_error_page` makes it obvious that the previous usage was wrong:
```py
create_500_error_page(HTTPException(404, ...))
```

evnchn:
Nice. With the intended use case self-explanatory, we don't need to prefix with `_` to force coding agents to not use it - they can use it at the right moment if it is called for.

PR #5581, opened 2025-12-15
Fix automatic event unsubscription

falkoschindler:
### Motivation

@ftilde noticed a memory leak in our `Event` class:
```py
EVENT = Event[[]]()

@ui.page('/')
def root():
    ui.label('Hi!')
    EVENT.subscribe(ui.notify)

app.timer(1, lambda: print(f'There are {len(EVENT.callbacks)} subscribers.'))
```

After visiting this page with a browser, the event is automatically unsubscribed after a few seconds.
But when requesting the page with curl (no JavaScript, no handshake, no socket connection), the unsubscription fails.

### Implementation

Here is the old `register_unsubscribe` code:
```py
async def register_unsubscribe() -> None:
    try:
        await client.connected()
        client.on_delete(lambda: self.unsubscribe(callback))
    except ClientConnectionTimeout:
        log.debug('Could not register a disconnect handler for callback %s', callback)
        self.unsubscribe(callback)
if core.loop and core.loop.is_running():
    background_tasks.create(register_unsubscribe())
else:
    core.app.on_startup(register_unsubscribe())
```
The default `timeout` for `client.connected` is `None`, so it waits indefinitely and the `on_delete` handler is never registered.

The code dates back to a [commit](https://github.com/zauberzeug/rosys/commit/1e36df2506f9fa47b1a0e82939ab90150e35c6dd) in RoSys 0.6.7 on January 17, 2023 (with NiceGUI at version 1.1.3). It replaced
```py
client.on_disconnect(lambda: self.unregister(callback))
```
with
```py
async def register_disconnect():
    try:
        await client.connected()
        client.on_disconnect(lambda: self.unregister(callback))
    except TimeoutError:
        self.unregister(callback)
task_logger.create_task(register_disconnect())
```
This was necessary because there was no "delete" event, but only "disconnect". But if a client doesn't connect (e.g. with curl), there would be no disconnect and the event unsubscription wouldn't happen.

Thanks to the new "delete" event, we can replace the whole background task for `register_ubsubscribe` with a single line:
```py
client.on_delete(lambda: self.unsubscribe(callback))
```
The "delete" event will always occur, even without a socket connection. Just note that the minimum reproduction above will take around a minute or two for the callback to be removed because stale clients are only pruned once a minute.

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5581/commits/59fbf7d547b5d24a9051eeb03936df71b5c32b60

### DIFF
--- a/tests/test_event.py
+++ b/tests/test_event.py
@@ -144,16 +144,12 @@ def test_reconnect(screen: Screen):
 async def test_event_memory_leak(screen: Screen):
     event = Event()
 
-    @ui.page('/')
-    def page():
-        ui.label('Keep screen occupied')
-
     @ui.page('/memory_leak')
     def memory_leak():
         event.subscribe(ui.notify)
 
-    screen.open('/')
+    screen.start_server()
     httpx.get(f'http://localhost:{Screen.PORT}/memory_leak', timeout=5)
     await asyncio.sleep(1)
     Client.prune_instances(client_age_threshold=0)
-    assert len(event.callbacks) == 0
+    assert not event.callbacks, 'event callbacks should be cleared after pruning clients'


evnchn (dismissed):
@falkoschindler Yikes! Good catch! 

2 observations:

1. **We should skip the creation of pytest if we want to ship 3.4.1 early. Hence the early approval.** 
2. I have identified exactly why the bug occured. Check below. 

In #5116, the last 2 commits:

1. [use default timeout of 3 seconds for registering event handlers](https://github.com/zauberzeug/nicegui/pull/5116/commits/e1acfc6e5466f1ddbe39fe111f86410800797664)
2. [remove default connection timeout](https://github.com/zauberzeug/nicegui/pull/5116/commits/adff733902a49ec303a58b8559d29fd180c6ad7e)

The last commit is conflict with the _intention_ of the second-last commit. And it flew right under the radar because we are very tired and exhausted after battling with Chrome's Speculative Loading...

tests/test_event.py
```diff
@@ -26,7 +27,7 @@ async def test_event_with_args(user: User):
     def page():
         ui.button('Click me', on_click=lambda: event.emit(42))
         event.subscribe(lambda: ui.notify('clicked'))
-        event.subscribe(lambda x: ui.notify(f'{x = }'))
+        event.subscribe(lambda x: ui.notify(f'{x=}'))
```
evnchn:
```suggestion
        event.subscribe(lambda x: ui.notify(f'{x = }'))
```

Keep the diff minimal.

evnchn (approved):
Merge whenever ready

rodja (dismissed):

PR #5582, opened 2025-12-15
Allow disabling or setting a custom GZipMiddleware

falkoschindler:
### Motivation

At the moment there is no way to disable or customize the GZipMiddleware. In some situations, like for our [Learning Loop](https://zauberzeug.com/products/learning-loop), this can be disadvantageous.

### Implementation

This PR adds a parameter to `ui.run`, defined as `gzip_middleware: type[GZipMiddleware] | None = GZipMiddleware`. This allows the user to disable the middleware by setting `gzip_middleware` to `None`, or to configure it via a custom subclass.

### Progress

- [ ] Documentation has been added (necessary?).

### DIFF
--- a/nicegui/ui_run.py
+++ b/nicegui/ui_run.py
@@ -60,8 +60,7 @@ def run(root: Optional[Callable] = None, *,
         reconnect_timeout: float = 3.0,
         message_history_length: int = 1000,
         cache_control_directives: str = 'public, max-age=31536000, immutable, stale-while-revalidate=31536000',
-        gzip_middleware_factory: Optional[
-            Callable[[ASGIApp], GZipMiddleware]] = lambda app: GZipMiddleware(app, minimum_size=500, compresslevel=9),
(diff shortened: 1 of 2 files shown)


evnchn:
@falkoschindler While this definitely works, and I have no issues merging it right away if you want to (we don't even need documentation since the new parameter automatically shows up):

**I was taught to avoid adding new parameters to `ui.run`.** 

And so I have a thought: Owing to how it is obvious to only GZip once, can NiceGUI "detect" if GZip middlewares are used in userland, and accomodate automatically by not GZipping twice if detected?

denniswittich:
@falkoschindler the last commit implements the alternative API approach we discussed

denniswittich (approved):

evnchn:
I'll respect the majority vote and have this over #5585  

🚀

sklibanov:
Thanks for this, I needed to turn off gzip compression for large file responses. Very nice.

PR #5585, opened 2025-12-15
Add GZip if not used by userspace only

evnchn:
### Motivation

Same as #5582:

> At the moment there is no way to disable or customize the GZipMiddleware. In some situations, like for our [Learning Loop](https://zauberzeug.com/products/learning-loop), this can be disadvantageous.

### Implementation

When the user attempts to add GZipMiddleware, NiceGUI automatically gives way by not adding it twice. 

This allows the userspace config of GZipMiddleware without adding extra param. 

You may in-fact disable GZip this way: `app.add_middleware(GZipMiddleware, minimum_size=9999999999999999)`

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5585/commits/4dd1933a0834451260f6137102b932f40a70ca88
- [x] Documentation is not necessary (considered a bug).

https://github.com/zauberzeug/nicegui/blob/dcfda542b3a6e57bebfc2ab5d019e60c3d2c1a51/nicegui/ui_run.py#L171-L172

### DIFF
--- a/nicegui/helpers.py
+++ b/nicegui/helpers.py
@@ -35,8 +35,10 @@ def warn_once(message: str, *, stack_info: bool = False) -> None:
         _shown_warnings.add(message)
 
 
-def is_pytest() -> bool:
-    """Check if the code is running in pytest."""
+def is_pytest(*, is_mockable: bool = False) -> bool:
+    """Check if the code is running in pytest, unless mockable and overridden"""
+    if is_mockable and 'HIDE_IS_PYTEST' in os.environ:
+        return False
     return 'PYTEST_CURRENT_TEST' in os.environ
 
 
--- a/nicegui/ui_run.py
+++ b/nicegui/ui_run.py
@@ -168,7 +168,7 @@ def run(root: Optional[Callable] = None, *,
     )
     core.root = root
     core.app.config.endpoint_documentation = endpoint_documentation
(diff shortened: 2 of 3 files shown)


falkoschindler:
Hi @evnchn,

I totally agree that we shouldn't solve every configuration problem with new parameters to `ui.run`. But it might be justified in this case, as I don't see a better alternative.

My concerns with your approach:

- Disabling the middleware via
    ```py
    app.add_middleware(GZipMiddleware, minimum_size=9999999999999999)
    ```
    is rather unintuitive.
- To replace the middleware, the user is forced to call `app.add_middleware` in user code. This changes the order of middlewares, which could change the behavior.

We have some ideas for improving PR #5582. Let's compare afterwards which we like more.

evnchn:
I think this PR is gone if "changes the order of middlewares" affects how GZip works. But from my understanding I can apply GZipMiddleware at any sequence and it's the same. (?)

I am indifferent to `app.add_middleware(GZipMiddleware, minimum_size=9999999999999999)` but maybe I am biased as the PR creator...

evnchn:
Middleware sequence does matter. https://share.google/aimode/mKHcUDYOt05ikAMur

2 fixes to your aforementioned concerns:

- Can this PR grab the user space GZipMiddleware and re-order it to the right place?
- For easy disabling of GZip, can we have `NoGZipMiddleware` somewhere in our library code?

But the ultimate question: **where is the line between "too much magic" and "Nice"?**

evnchn:
The majority vote sided with #5582 rather than PR. It has been a ride.

Issue #3939, opened 2024-11-01
Scoped slots are not passed to python described elements


zceemja:
### Description

This is on the same topic as #167 and #501. Here is the code to expand the idea (tested on nicegui 2.5.0):
```python
from nicegui import ui

columns = [
    {'name': 'id', 'label': 'ID', 'field': 'id'},
    {'name': 'name', 'label': 'Name', 'field': 'name'},
]
rows = [{'id': i, 'name': f'row-{i}'} for i in range(5)]

table = ui.table(columns=columns, rows=rows, row_key='id')
table.add_slot('body', r'''
<q-tr :props="props">
    <q-td key="id" :props="props">{{props.row.id}}</q-td>
</q-tr>''')

ui.run(prod_js=False)
```
And this works as expected. Props are passed down to `q-td` and is visible from vue debugging on browser:

![image](https://github.com/user-attachments/assets/d6c74b0b-6ed0-4bda-a179-2c62161d8788)

But this does not work when describing with body slot with python:
```python
...
table = ui.table(columns=columns, rows=rows, row_key='id')
with table.add_slot('body'):
    with table.row().props(':props=props'):
        with table.cell().props('key=id :props=props'):
            ...
...
```
![image](https://github.com/user-attachments/assets/ebad2af3-7b44-4bf1-bec7-cfe24446a083)
However there seems to be some disconnect (missing anonymous component?) between body slot and `QTr` and `props.row` are not accessible. 

Reason why I find this useful is that body description in python enables inserting custom vue components, and the only way I see to do it now is rewriting whole table with custom component as new combined component, which is a lot of duplicated boilerplate code.

falkoschindler:
Hi @zceemja,

Currently there is - as far as I know - no way to use scoped slots without writing template strings. This problem is discussed in #3148 and addressed in PR #3261. Hopefully there will be a solution in the near future.

Since passing props like `.props(':props=props')` isn't expected to work, I'll close this issue.

PR #3261, opened 2024-06-21
Table cell slot without string template

CrystalWindSnake:
This PR is based on discussion #3148. The goal is to table cell slots without writing any Vue string templates.

PR #5592, opened 2025-12-18
Allow defining table cell slots without string templates

falkoschindler:
### Motivation

Currently, customizing table cells requires Vue template strings:

```python
table.add_slot('body-cell-age', '''
    <q-td key="age" :props="props">
        <q-badge :color="props.value < 21 ? 'red' : 'green'">
            {{ props.value }}
        </q-badge>
    </q-td>
''')
```

This has several drawbacks:
- Not Pythonic
- No IDE support (no auto-completion, no type hints)
- Easy to make syntax errors
- Requires knowledge of Vue template syntax

This PR enables using NiceGUI elements directly:

```python
with table.add_slot('body-cell-age'):
    with table.cell():
        ui.badge().props(':label="props.value"').props(':color="props.value < 21 ? \'red\' : \'green\'"')
```

### Implementation

Modified `nicegui/static/nicegui.js`:

1. **Added slot props context parameter** to `renderRecursively(elements, id, slotPropsContext)`
2. **Enhanced dynamic property evaluation**: When evaluating `:prop="expression"`, slot props are passed as the `props` variable
3. **Implemented slot props inheritance**: Child slots inherit parent context using `slotProps || slotPropsContext`

### Full example

```py
table = ui.table(rows=[{'name': 'Alice', 'age': 18}, {'name': 'Bob', 'age': 21}, {'name': 'Carol', 'age': 42}])
with table.add_slot('body-cell-name'):
    with table.cell():
        ui.button().props(''' flat :label="props.value" ''') \
            .on('click', js_handler='() => emit(props.label)', handler=lambda e: ui.notify(f'Clicked {e.args}'))
with table.add_slot('body-cell-age'):
    with table.cell():
        ui.badge().props(''' :label="props.value" :color="props.value < 21 ? 'red' : 'green'" ''')
```

Screenshot 2025-12-18 at 09 02 27

### Progress

---

Related discussions:

- https://github.com/zauberzeug/nicegui/discussions/846
- https://github.com/zauberzeug/nicegui/discussions/1138
- https://github.com/zauberzeug/nicegui/discussions/2080
- https://github.com/zauberzeug/nicegui/discussions/2625
- https://github.com/zauberzeug/nicegui/discussions/3148
- https://github.com/zauberzeug/nicegui/pull/3261 (← PR)
- https://github.com/zauberzeug/nicegui/discussions/3317
- https://github.com/zauberzeug/nicegui/discussions/3540
- https://github.com/zauberzeug/nicegui/discussions/3662
- https://github.com/zauberzeug/nicegui/issues/3939
- https://github.com/zauberzeug/nicegui/discussions/5092
- https://github.com/zauberzeug/nicegui/discussions/5591

### DIFF
--- a/website/documentation/content/table_documentation.py
+++ b/website/documentation/content/table_documentation.py
@@ -169,7 +169,12 @@ def table_with_buttons():
 
 @doc.demo('Table with drop down selection', '''
     Here is an example of how to use a drop down selection in a table.
-    After emitting a "rename" event from the scoped slot, the `rename` function updates the table rows.
+    After emitting a "update:model-value" event from the `ui.select` element,
+    the `rename` function updates the table rows.
+
+    *Since version 3.5.0:*
+    UI elements in scoped slots can access the `props` object to get the current row and column.
+    Previously, the `props` object was only available in Vue templates.
 ''')
 def table_with_drop_down_selection():
     from nicegui import events
@@ -386,6 +391,10 @@ def computed_fields():
     We use the `body-cell-age` slot to insert the `q-badge` into the `age` column.
     The ":color" attribute of the `q-badge` is set to "red" if the age is under 21, otherwise it is set to "green".
     The colon in front of the "color" attribute indicates that the value is a JavaScript expression.
+
+    *Since version 3.5.0:*
+    UI elements in scoped slots can access the `props` object to get the current cell value.
+    Previously, the `props` object was only available in Vue templates.
 ''')
 def conditional_formatting():
     columns = [
@@ -409,6 +418,10 @@ def conditional_formatting():
 @doc.demo('Table cells with links', '''
     Here is a demo of how to insert links into table cells.
     We use the `body-cell-link` slot to insert an `<a>` tag into the `link` column.
+
+    *Since version 3.5.0:*
+    UI elements in scoped slots can access the `props` object to get the current cell value.
+    Previously, the `props` object was only available in Vue templates.
 ''')
 def table_cells_with_links():
     columns = [


evnchn:
I know this is in progress, but possibly is this missing a commit? This just changes the nicegui.js

falkoschindler:
@evnchn Yes, that's all it takes. The example I've given should work with these changes in nicegui.js.

falkoschindler:
What I'm still unsure about:

- Is there a better way for event handling than via `js_handler`?
    ```py
    .on('click', js_handler='() => emit(props.label)', handler=lambda e: ui.notify(f'Clicked {e.args}'))
    ```
    What if the button label is something different? How to access the `row` object? Is there a better solution than
    ```py
    ui.button('Notify').props(':props=props') \
        .on('click', js_handler='() => emit(props.props.row.name)', handler=lambda e: ui.notify(f'Clicked {e.args}'))
    ```

- How to document the new usage? Do we want to rewrite all slot demos? Or do we want to keep the old documentation (for now) and add an additional demo? Maybe it's better to do it step by step.

evnchn:
I'm still wrapping my head around the PR so not sure about the js_handler thing (I'm considering to use data-* props, just like the js_handler PR originally enables)

But for the documentation, I think this is a **strong** case for having both the old and the new code. I think it will ease understanding and migration.

nicegui/static/nicegui.js
```diff
@@ -160,9 +160,11 @@ function renderRecursively(elements, id) {
     if (key.startsWith(":")) {
       try {
         try {
-          props[key.substring(1)] = new Function(`return (${value})`)();
+          props[key.substring(1)] = slotPropsContext
+            ? new Function("props", `return (${value})`)(slotPropsContext)
+            : new Function(`return (${value})`)();
         } catch (e) {
-          props[key.substring(1)] = eval(value);
+          props[key.substring(1)] = slotPropsContext ? undefined : eval(value);
```
evnchn:
```suggestion
          props[key.substring(1)] = eval(value);
```

In the new code logic, if:

- somehow we land in this code (`Function` failed),
- `value` is a JS expression that is critically dependent on `props`

I would prefer we still try to do `eval(value)` so that the error is propagated to ```console.error(`Error while converting ${key} attribute to function:`, e);``` instead of silently doing nothing via the `undefined`. 

This facilitates easier debugging and most importantly fails pytests.

nicegui/static/nicegui.js
```diff
@@ -160,9 +160,11 @@ function renderRecursively(elements, id) {
     if (key.startsWith(":")) {
       try {
         try {
-          props[key.substring(1)] = new Function(`return (${value})`)();
+          props[key.substring(1)] = slotPropsContext
+            ? new Function("props", `return (${value})`)(slotPropsContext)
+            : new Function(`return (${value})`)();
```
evnchn:
```suggestion
          props[key.substring(1)] = new Function("props", `return (${value})`)(slotPropsContext);
```

Under the old code logic, the following is satisfied:

- `slotPropsContext` is undefined
- `value` is a JS-expression which does not depend on `props` because it is not set anywhere

Therefore, it is safe to directly migrate to ```Function("props", `return (${value})`)(slotPropsContext)``` without a ternary operator check

nicegui/static/nicegui.js
```diff
@@ -223,12 +225,12 @@ function renderRecursively(elements, id) {
...
             },
             {
-              props: props,
+              props: slotProps,
             }
           )
         );
       }
-      const children = data.ids.map((id) => renderRecursively(elements, id));
+      const children = data.ids.map((id) => renderRecursively(elements, id, slotProps || slotPropsContext));
```
evnchn: I think it may be better to revert name to `props` and use `propsContext` as the props aren't inherently related to slot rendering. 

More that we are inheriting the context of the parent props when rendering the child for further processing. 

It will keep the diff minimal and the change more understandable.

evnchn (commented):
I finally understood the changes now. 

IMO my understanding of this PR was held back by the variable renaming and the ternary operators. I have review them and personally found both to be not strictly necessary. 

If you do agree, please adopt my changes and the PR should be good to go **short of documentation.** Thanks!

falkoschindler:
With a60775f1160fc747c4c2236c244fd97715b29206 event handlers automatically have access to a `props` object and we can simplify the event registration from
```py
ui.button('Notify').props(':props=props') \
    .on('click', js_handler='() => emit(props.props.row.name)', handler=lambda e: ui.notify(f'Clicked {e.args}'))
```
to
```py
ui.button('Notify') \
    .on('click', js_handler='() => emit(props.row.name)', handler=lambda e: ui.notify(f'Clicked {e.args}'))
```

evnchn (commented):
The rest of the PR is a banger. 4-line change in core-logic, a self-descriptive test. 

But, being the first of 3 features mentioned in the [NiceGUI 2025 Recap](https://www.reddit.com/r/nicegui/comments/1q07kau/nicegui_2025_recap_and_whats_coming_in_2026/), do you _really_ think that we should simply add a new entry to the **bottom** of the existing docs?

- Put at the top? 
- Existing demos should be modernized with old code hidden in `ui.expansion` until NiceGUI 4.0 rolls around?

evnchn:
@falkoschindler Bad news

I try and do https://nicegui.io/documentation/table#table_with_expandable_rows like this

```py
    with table.add_slot('header'):
        with table.row():
            table.header().props('auto-width')
            for col in table.columns:
                with table.header():
                    ui.label(col['label'])
```

but then `text-right` is not longer applied for us from Quasar and the names are not right-aligned anymore

image

No idea how to fix for now.

falkoschindler:
@evnchn Good point. Maybe we should go the extra mile and convert all demos now we're at it. I'm just not sure how to present old and new code without adding a whole new feature to the documentation system. Maybe we don't need the old versions at all? There are still plenty of examples online (GitHub discussions etc.).

And yes, the horizontal alignment is annoying. Maybe we should try to find and fix the root cause. Right now it will annoy everyone using it and fixing it later could be considered a breaking change.

evnchn:
I would actually promote keeping the old code until forever, because not many people realize NiceGUI is vue underneath, and the table documentation is like a proof. A weak but readily-accessible one. 

For the horizontal alignment, I agree that we better get it straight on day1, but seeing that the missing class is `text-right` which came from the hands of Quasar, patch may be tough, and we may have to let this PR cook for longer. 

https://github.com/zauberzeug/nicegui/blob/b6554e1e6993e8959243d1972bb072adc290aa1f/nicegui/static/quasar.unimportant.css#L12333

falkoschindler:
> I would actually promote keeping the old code until forever, because not many people realize NiceGUI is vue underneath, and the table documentation is like a proof. A weak but readily-accessible one.

Ok, but then we'd need a way to optionally hide it. Otherwise the table documentation page get's too convoluted.

Alternatively we could convert all demos to the new API and add a single dedicated demo for custom Vue templates (like the demo I added, but using the old API). This should be sufficient to show off the Vue layer, but keeps the focus on the new preferred way of doing things.

> For the horizontal alignment, I agree that we better get it straight on day1, but seeing that the missing class is `text-right` which came from the hands of Quasar, patch may be tough, and we may have to let this PR cook for longer.

I still have hope for getting it ready for 3.5. But we'll see.
Maybe we simply need to add `.text-right` to `table.cell`'s default classes?

evnchn:
> Simply add `.text-right`

Oh I wish it was so simple. It actually depends on the columns data. But it does rely on it in a simple manner though. Check this out:

```py
columns = [
    {'name': 'name', 'label': 'Name', 'field': 'name', 'align': 'evnchn'},  # hmm
    {'name': 'age', 'label': 'Age', 'field': 'age'},
]
```

Then the name column has class `text-evnchn` even though it is non-existent. 

Still, we need ability to read the column, which may be tough.

evnchn:
Suppose need `:props=props` but why when add it, it degenerates and does not show?

```py
        with table.add_slot('header'):
            with table.row().props(':props=props'):
                table.header().props('auto-width :props=props')
                for col in table.columns:
                    with table.header().props(':props=props'):
                        ui.label(col['label'])
```

evnchn:
This works:

```py
        with table.add_slot('header'):
            with table.row():
                table.header()
                for col in table.columns:
                    with table.header().props(f':props=props key={col["name"]}'):
                        ui.label(col['label'])
```

`table.header()`'s underlying `q-td` need both the props and the key to find itself and figure out how it should text-align. 

- Having `props` but failing to pass `key` causes it to degenerate. (explains https://github.com/zauberzeug/nicegui/pull/5592#issuecomment-3715722253)
- Having `key` but failing to pass `props` causes it to degenerate. (means that we cannot just pass `props` to all elements and expect no casualties)
  - Or we instruct the user to do `ui.element('th')` instead of `table.header()` in due course?

falkoschindler:
Interesting. But the vertical alignment is off:

Screenshot 2026-01-06 at 19 25 42

Maybe this error message helps understanding the problem:

> Error while converting :key attribute to function: ReferenceError: col is not defined

falkoschindler:
This works:
```py
table = ui.table(rows=rows, row_key='name').classes('w-72')
with table.add_slot('header'):
    with table.row():
        for col in table.columns:
            with table.header().props(f':props=props key={col["name"]}'):
                ui.label(col['label']).classes('inline')
```

falkoschindler:
Ok, the alignment problem is solved. Now I can try converting the demos.

The table with expandable rows seems to be particularly difficult:
```py
columns = [
    {'name': 'name', 'label': 'Name', 'field': 'name'},
    {'name': 'age', 'label': 'Age', 'field': 'age'},
]
rows = [
    {'name': 'Alice', 'age': 18},
    {'name': 'Bob', 'age': 21},
    {'name': 'Carol'},
]

table = ui.table(columns=columns, rows=rows, row_key='name').classes('w-72')
with table.add_slot('header'):
    with table.row():
        table.header().props('auto-width')
        for col in table.columns:
            with table.header(col['name']):
                ui.label(col['label'])
with table.add_slot('body'):
    with table.row():
        with table.cell().props('auto-width'):
            ui.button(color='accent').props('''
                round dense size=sm
                :icon="props.expand ? 'remove' : 'add'"
            ''').on('click', js_handler='() => (props.expand = !props.expand)')
        for col in table.columns:
            with table.cell(col['name']):
                ui.label().props(f':innerHTML=props.row.{col["name"]}')
    with table.row().props(':v-show="props.expand"'):
        with table.cell().props('colspan=100%'):
            ui.label().props(':innerHTML="`This is ${props.row.name}.`"')
```

I'm struggling with `:v-show="props.expand"`.

falkoschindler:
@evnchn I rewrote all examples and demos as much as possible. As far as I can tell, they are all working correctly and visually identical (or equivalent) to the original version. For the remaining instances I couldn't find a good solution. Let's leave them like this. The table documentation page nicely represents the two possible ways to define slots and the user can choose what suits best.

May I ask for a final review to please the CI pipeline? Thanks!

PR #5594, opened 2025-12-19
Set Redis TTL on tab storage keys

quantumdark:
### Motivation

When using Redis as a storage backend (NICEGUI_REDIS_URL), tab storage keys accumulate forever in Redis. The existing prune_tab_storage() function only clears keys from the in-memory _tabs dictionary. If the application restarts, _tabs is empty and the app has no knowledge of orphaned Redis keys, causing them to persist until Redis runs out of memory (OOM).

Related to the max_tab_storage_age configuration which currently only works for in-memory cleanup but has no effect on Redis key expiration.

### Implementation

Added TTL (Time To Live) support to RedisPersistentDict for tab storage:

  - Added optional ttl: float | None parameter to RedisPersistentDict.__init__ (follows type of max_tab_storage_age)
  - Modified publish() to use Redis SETEX command (atomic set-with-expiry) when TTL is configured, otherwise falls back to regular SET
  - Tab storage now passes ttl=Storage.max_tab_storage_age when creating RedisPersistentDict instances                                                                                                                                                                                                                 
  - TTL is refreshed on every write (every data change), so active tabs never expire while inactive/orphaned tabs are automatically cleaned up by Redis
 
  Key behavior: 
  - Only tab storage uses TTL (user and general storage remain permanent)
  - TTL defaults to max_tab_storage_age (30 days) for consistency with existing pruning behavior
  - Redis handles cleanup automatically, even after app restarts

### Progress

### DIFF
--- a/nicegui/storage.py
+++ b/nicegui/storage.py
@@ -22,6 +22,7 @@ request_contextvar: contextvars.ContextVar[Request | None] = contextvars.Context
 GENERAL_ID = 'general'
 USER_PREFIX = 'user-'
 TAB_PREFIX = 'tab-'
+TTL_BUFFER_SECONDS = 20  # Buffer to avoid race with prune_tab_storage polling
 
 
 class RequestTrackingMiddleware(BaseHTTPMiddleware):
@@ -75,8 +76,7 @@ class Storage:
     @staticmethod
     def _create_persistent_dict(id: str) -> PersistentDict:  # pylint: disable=redefined-builtin
         if Storage.redis_url:
-            BUFFER_SECONDS = 20  # Buffer to avoid race with prune_tab_storage polling
-            ttl = int(core.app.storage.max_tab_storage_age + BUFFER_SECONDS) if id.startswith(TAB_PREFIX) else None
+            ttl = int(core.app.storage.max_tab_storage_age + TTL_BUFFER_SECONDS) if id.startswith(TAB_PREFIX) else None
             return RedisPersistentDict(url=Storage.redis_url, id=id, key_prefix=Storage.redis_key_prefix, ttl=ttl)
         else:
             return FilePersistentDict(Storage.path / f'storage-{id}.json', encoding='utf-8')


nicegui/persistence/redis_persistent_dict.py
```diff
@@ -81,8 +82,12 @@ async def backup() -> None:
             if not await self.redis_client.exists(self.key) and not self:
                 return
             pipeline = self.redis_client.pipeline()
-            pipeline.set(self.key, json.dumps(self))
-            pipeline.publish(self.key + 'changes', json.dumps(self))
+            data = json.dumps(self)
+            if self.ttl is not None:
+                pipeline.setex(self.key, int(self.ttl), data)
+            else:
+                pipeline.set(self.key, data)
```
evnchn: Maybe we can just directly use the `ex` parameter of `pipeline.set`. 

`setex` is deprecated anyways: https://share.google/aimode/YpTRXfFGQScTxYe7x

nicegui/persistence/redis_persistent_dict.py
```diff
@@ -81,8 +82,12 @@ async def backup() -> None:
             if not await self.redis_client.exists(self.key) and not self:
                 return
             pipeline = self.redis_client.pipeline()
-            pipeline.set(self.key, json.dumps(self))
-            pipeline.publish(self.key + 'changes', json.dumps(self))
+            data = json.dumps(self)
```
evnchn, line 93: Actually a good thing: Guards against `json.dumps` mutating between set and publish. Please keep this!

evnchn (changes requested):
Surface-level-speaking, let's use `set` for expiry if we really want to handle expiry in Redis. 

And, generally we do not prefer to pass a parameter around for single-purpose. `RedisPersistentDict` can read `core.app.storage.max_tab_storage_age` (not `Storage.max_tab_storage_age`, let's align with the codebase) directly. 

---

But I want to ask: Is that really a good idea? 

Tracing back `max_tab_storage_age`, it is originally exclusively used in:

https://github.com/zauberzeug/nicegui/blob/dc49827d812850649935142c7356ba7f7aa8ee2d/nicegui/nicegui.py#L246-L254

and notably: it is invoked **every 10 seconds** as seen in:
 https://github.com/zauberzeug/nicegui/blob/dc49827d812850649935142c7356ba7f7aa8ee2d/nicegui/nicegui.py#L146

Then, I have the following questions:

- What happens to the tab storage, if Redis tossed the key because it is sharp-pass the deadline, but NiceGUI did not prune it due to the polling interval?
  - Should, in the next polling interval, the tab storage is expired, does the pruning still occur normally?
  - Should the tab storage is further mutated before the next NiceGUI poll occurs, then isn't the state torn between Redis and NiceGUI? Are there side-effects?
- Owing to the above concerns, is it better to handle the purging of Redis tab storage at `prune_tab_storage` instead?

Thank you for reading my longwinded comment!

evnchn:
Sorry, did I scare you away? 🥺

quantumdark:
Apologies, I was off for the holidays.

I've chosen the parameter approach because to me the alternative was too implicit. To address your feedback I added a flag based on the id prefix to avoid an explicit parameter, but I'm open for other ideas.

To answer your question is this a good idea, to me this approach seems very reasonable. We have a Redis server that we have to reboot from time to time due to is running out of memory because of the orphaned tab storage keys. When using Redis storage, if a NiceGUI server crashes or restarts:
  - The in-memory _tabs dict is lost
  - Redis keys persist forever
  - On restart, prune_tab_storage() only knows about tabs in the current _tabs dict
  - Old Redis keys from previous server instances become orphaned forever

The drift between the expiration time counted in Python vs in Redis should be pretty close because they both update on writes. If you think this is still a concern we could add a small delta (max_tab_storage_age + delta) that is greater than polling interval when saving to Redis, this way if it's not pruned by that time, it is safe to assume it is orphaned.

evnchn:
Thanks. It seems that using `ex` is non-negotiable because the server does not know about tab IDs in the past-life. 

I simplified the code a bit, and added a delta. Rationale for the delta value chosen is in the commit message of https://github.com/zauberzeug/nicegui/pull/5594/commits/f64322651c6f013b040572eac21ef5faf6a381c9

> Rationale: 1 cycle is definitely not enough due to potentially last-modified being 70 seconds:
> 
> - NiceGUI's not pruning it
> - Redis, due to network latency, gets the request at 70.1 seconds, expired
> 
> 20 seconds is a really conservative setting considering that attempts to connect to a Redis server with 10 seconds of network latency will likely fail horribly, but better safe than sorry!

@falkoschindler Should be ready for review!

PR #5598, opened 2025-12-20
Terminate `.connected()` waits on client deletion to ensure bounded waiting

### LINKED ISSUE #5595: `ui.timer` leaks when client disconnects immediately
### First Check

### Example Code

```python
import objgraph

from nicegui import ui, background_tasks


def print_something():
    num_timers = len(objgraph.by_type('Timer'))
    print(f'Number of Timer instances: {num_timers}')


@ui.page('/')
def page():
    ui.label('# running tasks')
    ui.label().bind_text_from(background_tasks, 'running_tasks', backward=len)
    ui.timer(1, print_something)


ui.run(port=8082, reload=False)
```

### Description

When running this code and opening the website with a short lived client (curl or hammering refresh in the browser), the number of running timers (as printed in the console) and the number of background tasks (shown in UI) will steadily increase. This issue has the same underlying problem as #5581:

In [ui.timer](https://github.com/zauberzeug/nicegui/blob/dc49827d812850649935142c7356ba7f7aa8ee2d/nicegui/elements/timer.py#L21) we wait for a client to connect without a timeout. Since `_can_start` is ultimately waited for in a background task created when calling `ui.timer` the task never completes and referenced objects are never cleaned up.

One fix would be to just add some timeout value for the `.connected` call in ui.timer, but maybe we want to consider defaulting to a timeout again (i.e. reversing part of the change in b9c8a192e8d53f4c15ac2bab121c1a50c1f0fbb6).

### NiceGUI Version

3.4.1

### Python Version

3.12.11

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

evnchn:
### Motivation

In #5116 the decision was made to have `client.connected()` assume an infinite timeout unless set. While this works for user code fine, **it breaks library code.** 

While we have fixed it in #5581 by not using `client.connected()` (using `client.on_delete` instead), to fix #5595, I see that [the code structure of `ui.timer`](https://github.com/zauberzeug/nicegui/blob/dc49827d812850649935142c7356ba7f7aa8ee2d/nicegui/elements/timer.py#L21), calls for a `client.connected()` with bounded waiting. 

### Implementation

This PR momentarily sets `self._connected` in the `client.delete()`. This stops the await of `client.connected()`. And with the assertation we raise `ClientConnectionTimeout` as intended (it is timed out because the client is deleted and will never connect)

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5598/commits/3f1e67465a86f54d6ac2d9b0077e8450db61ef32
- [x] Documentation is not necessary (userspace code will not execute after `client.connected()` since page function is outright cancelled + error is raised).

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -208,9 +208,8 @@ class Client:
         is_prefetch = 'prefetch' in purpose and 'prerender' not in purpose
         try:
             await asyncio.wait_for(self._connected.wait(), timeout=None if is_prefetch else timeout)
-            assert self._deleted is not True, 'Client was deleted while waiting for connection'
-        except AssertionError:
-            pass
+            if self._deleted:
+                return  # Client was deleted while waiting for connection
         except asyncio.TimeoutError as e:
             raise ClientConnectionTimeout(self) from e
 


evnchn:
https://github.com/zauberzeug/nicegui/pull/5598/commits/3f1e67465a86f54d6ac2d9b0077e8450db61ef32 fails on main and passes on this PR.

falkoschindler (approved):
Thanks, @evnchn! Looks like a good solution. 👍🏻

Commit 902613b6a4a6ad024ecd549a1f5fad4bcd696f74 simplifies it a bit with a straightforward if-condition rather than catching an assertion.

Let's merge! 🚀

nicegui/client.py
```diff
@@ -208,6 +208,8 @@ async def connected(self, timeout: float | None = None) -> None:
         is_prefetch = 'prefetch' in purpose and 'prerender' not in purpose
         try:
             await asyncio.wait_for(self._connected.wait(), timeout=None if is_prefetch else timeout)
+            if self._deleted:
+                return  # Client was deleted while waiting for connection
```
evnchn, line 212: What's to stop us from removing these two lines, as we are returning anyways?

evnchn (commented):
May have some simplifications on the table?

nicegui/client.py
falkoschindler: Oh! You're right!
I'll test and push it directly on main (since this PR is already merged).

Issue #5597, opened 2025-12-19
ui.mermaid(...).on(error, ...) - attached only once, the first


AwMalka:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
def render_mermaid_safe(code: str, title: str = '', container=None):
    """
    Safely render a mermaid diagram. If syntax is invalid, skip rendering.
    Note: This renders directly - caller should already be in the right container context.
    """
    if not is_valid_mermaid_syntax(code):
        print(f'[visualizations] Skipping invalid mermaid diagram: {code[:100]}...')
        return False
    
    def hide_on_error(e, mermaid_wrapper, id):
        print (f'!!!!!!!!!!!!!!! - hiding mermaid due to error - {id} {mermaid_wrapper.html_id}')
        mermaid_wrapper.set_visibility(False)

    try:
        if container:
            with container:
                # Wrap in a column so we can hide just this diagram on error
                with ui.column().classes('w-full') as mermaid_wrapper:
                    if title:
                        ui.label(title).classes('text-sm font-medium text-gray-700 mt-2')
                    print (f'================ - setting mermaid - {mermaid_wrapper.html_id}')
                    ui.mermaid(code).classes('w-full').on('error', lambda e, w=mermaid_wrapper, id=mermaid_wrapper.html_id: hide_on_error(e, w, id))
                return True
        else:
            # Wrap in a column so we can hide just this diagram on error
            with ui.column().classes('w-full') as mermaid_wrapper:
                if title:
                    ui.label(title).classes('text-sm font-medium text-gray-700 mt-2')
                print (f'================ - setting mermaid - {mermaid_wrapper.html_id}')
                ui.mermaid(code).classes('w-full').on('error', lambda e, w=mermaid_wrapper, id=mermaid_wrapper.html_id: hide_on_error(e, w, id))
            return True
    except Exception as e:
        print(f'[visualizations] Mermaid rendering error: {e}')
        return False
```

### Description

i have ai gerenerating charts
ive saved the state where:
1. valid
2. invalid sytax
3. invalid sytax

for each i call render_mermaid_safe(...)
on creation:
================ - setting mermaid - c2091
================ - setting mermaid - c2094
================ - setting mermaid - c2097
[...]

PR #5599, opened 2025-12-20
Fix `ui.mermaid` error ID confusion

evnchn:
### Motivation

While trying to investigate #5597 I found a bug. It is the root cause of #5597. 

```py
def mermaid_with_error_notify(code: str):
    el = ui.mermaid(code)
    return el.on('error', lambda el=el: ui.notify(f'Mermaid error for element {el.html_id}'))

ui.label(mermaid_with_error_notify('graph TD\nA --> B').html_id)

ui.label(mermaid_with_error_notify('BAD SYNTAX').html_id)
```

Obviously, c4 mermaid is fine, c6 mermaid has error. 

But in the notify, it shows "Mermaid error for element c4" which is wrong. 

### Implementation

Since `ui.mermaid` uses a single element as the renderer, calling `this.$emit` is wrong as it raises the error on the renderer despite rendering some element's Mermaid content. 

As such, use `getElement` to find the corresponding element to emit the error. 

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5599/commits/abbacbff122dfb4768e3e59d7c49c7df36a4801e

### DIFF
--- a/tests/test_mermaid.py
+++ b/tests/test_mermaid.py
@@ -1,7 +1,7 @@
 import pytest
 from selenium.webdriver.common.by import By
 
-from nicegui import ui
+from nicegui import events, ui
 from nicegui.testing import Screen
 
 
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Nice! I just replaced `getElement(element.id.substring(1))` with `getElement(element)` to remove dependency of knowing the HTML ID format.
Ready to merge! 🚀

PR #5600, opened 2025-12-20
Fix Docker bind-mount regression after uv migration

### LINKED ISSUE #5593: Docker image ≥3.4.0 breaks when bind-mounting /app (regression from 3.3.1)
### First Check

### Example Code

```python

```

### Description

You are going to get a beautiful: ```ModuleNotFoundError: No module named 'nicegui'```

if you use `FROM zauberzeug/nicegui:latest`

### NiceGUI Version

Latest docker image

### Python Version

not relevant

### Browser

Other

### Operating System

Linux

### Additional Context

_No response_

lpellicer:
This restores compatibility with the documented `-v $(pwd):/app` workflow.

Since 3.4.0 (uv migration), the Docker image installs its runtime environment under `/app` and uses `uv run` at runtime. When users bind-mount `/app` (as documented), this hides the image’s `pyproject.toml` and venv, leading to `ModuleNotFoundError: nicegui` (and `uv` warns that `--no-sync` has no effect outside a project).

This PR:
- moves the uv project environment outside `/app` (UV_PROJECT_ENVIRONMENT=/opt/venv)
- runs the venv Python directly at runtime (CMD ["/opt/venv/bin/python", "main.py"])

Build-time dependency resolution and installation still use `uv`; the change is purely to make `/app` safe to bind-mount again.

Verified locally with 3.4.1 using a host directory containing only `main.py`:
- `zauberzeug/nicegui:3.4.1` fails with `ModuleNotFoundError`
- the patched image runs successfully with the same bind-mount

Fixes #5593

### DIFF
--- a/release.dockerfile
+++ b/release.dockerfile
@@ -5,6 +5,9 @@ LABEL maintainer="Zauberzeug GmbH <info@zauberzeug.com>"
 # Install uv
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
+ENV UV_PROJECT_ENVIRONMENT=/opt/venv
+ENV PATH="/opt/venv/bin:${PATH}"
+
 WORKDIR /app
 
 # Copy dependency files first for better layer caching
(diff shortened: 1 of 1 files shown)


evnchn:
Thinking about the /app/.venv: 

1. can we have backwards compatibility with 3.4.0 by means of symlinking (or some other kind of linking, whatever kind suits best) which, we should remove on next major version.
2. Is it worth it?

evnchn:
> Consider what was in 3.4.0 as a mistake

Then I would recommend releasing 3.4.2, so that _across the last patch versions of each minor version_, the behaviour is consistent. **Though it may be tough since some non-bugfixes are already merged to main...**

Issue #5110, opened 2025-09-04
Strong reference cycles in slot children preventing cleanup by reference counting


thevinchi:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui


class MyLabel(ui.label):
    def __del__(self) -> None:
        print(f'Garbage collected element {self}')

    def _handle_delete(self) -> None:
        print(f'Deleting element {self}')


@ui.page('/')
async def _index() -> None:
    with ui.column():
        MyLabel('test1')
        ui.link('other', '/other')

    with ui.splitter() as splitter:
        with splitter.before:
            MyLabel('test2')
        with splitter.after:
            MyLabel('test3')


@ui.page('/other')
async def _other() -> None:
    with ui.column():
        MyLabel('test4')
        ui.link('home','/')


ui.run()
```

### Description

Child elements of non-default slots do not appear to be garbage collected. Upon cursory inspection of the `Element` class, it appears `self.slots` is never referenced during `remove` or `delete` or otherwise. This appears to be causing strong references to the descendants of these slots to hang around and prevent garbage collection. When a client disconnects, the strong references to `Elements` are popped off, but this does not remove the remaining strong reference within sibling `Slot.children` lists. 

Due to the recursive nature of `Element` and `Slot` I could surely be missing something here. However, the simple example helps illustrate what I'm seeing.

### NiceGUI Version

2.23.3

### Python Version

3.11.13

### Browser

Chrome

### Operating System

Linux

### Additional Context

When running the sample code, once you click on the link, the console output will show:

```bash
Deleting element MyLabel [text=test1]
Deleting element MyLabel [text=test2]
Deleting element MyLabel [text=test3]
Garbage collected element MyLabel [text=test1]
```

However, I would expect to see:
```bash
Deleting element MyLabel [text=test1]
Deleting element MyLabel [text=test2]
Deleting element MyLabel [text=test3]
Garbage collected element MyLabel [text=test1]
Garbage collected element MyLabel [text=test2]
Garbage collected element MyLabel [text=test3]
```

thevinchi:
[...]

Issue #4116, opened 2024-12-17
`ui.dialog` hidden when created inside a `ui.menu`


tgbl-mk:
### Description

I am trying to use ui.menu and ui.menu_item to create a dropdown menu (think file menu in most applications). If i try and create a dialog window from the function called by on_click,  then I get an unexpected response when clicking the menu item.

This enough code to reproduce the issue:
```python
from nicegui import ui


class MenuItemBug:

    def compose(self):
        with ui.row():
            with ui.button("Menu 1"):
                with ui.menu():
                    ui.menu_item("Item 1", on_click=self.on_item_click)
                    ui.menu_item("Item 2", on_click=self.on_item_click, auto_close=False)
            with ui.button("Menu 1"):
                with ui.menu():
                    ui.menu_item("Item 1", on_click=self.on_item_click, auto_close=False)
                    ui.menu_item("Item 2", on_click=self.on_item_click)
        return self

    def run(self):
        ui.run(port=1004)

    def dialog(self):
        with ui.dialog(value=True) as _, ui.card():
            with ui.column():
                ui.label("Bit of a journey to get here!")

    def on_item_click(self):
        print("click")
        self.dialog()


MenuItemBug().compose().run()
```
If you click a menu item that has auto_close set to False then the dialog opens as expected. However if you click a menu item with auto_close set to True(the default value), "click" is printed to the terminal and the menu closes without opening the dialog. If you reopen the menu then the dialog will then open. I have attached a video showing the behaviour below.

https://github.com/user-attachments/assets/53da9af8-4d1c-4431-b03b-1474d91eed65

It seems that the dialog will only be displayed whilst the menu is open.

Is this expected behaviour? An acceptable solution for me would be to leave the menu open while the dialog is displayed, and to close the menu when the dialog is closed. I suspect this is already possible with the existing methods.

I am using python 3.11.0 and nicegui 2.8.1 have attached the requirements.txt from my .venv below 

[requirements.txt](https://github.com/user-attachments/files/18165514/requirements.txt)

Please let me know if any additional information is required! Thanks

falkoschindler:
Thanks for reporting this issue, @tgbl-mk!

See https://github.com/zauberzeug/nicegui/discussions/4092#discussioncomment-11553083 for an explanation what is going on.

[...]

PR #5602, opened 2025-12-23
Create non-local dialogs

evnchn:
### Motivation

#4116 highlights how all `ui.dialog`s are local in nature. This means they are bound by the element they are in. 

In the case of `ui.menu`, this causes the dialog to not show up. 

However, the attempt to make `ui.dialog`'s non-local has stalled with https://github.com/zauberzeug/nicegui/issues/4116#issuecomment-2551829739, where the discussion thereafter revolves around warnings and workarounds. 

This PR addresses the issue directly. 

### Implementation

At the end of the day, the concept of a container clear (pivotal to `ui.refreshable`) is that, all of its child elements should be deleted when the container is cleared. 

Now, we are placing the dialog in `self.client.content` and the element is not cleared. 

Ask ourselves: Clear that element, and we are good to go!

In this PR, I introduce the concept of `DeletePropagationElement`, whose sole role is to be placed in the container, to be deleted normally, and when it is deleted it also deletes whatever element was pointed to it. 

As such `DeletePropagationElement(element_to_delete=self)` reads in English "even though I am not in the container, please delete me when my corresponding `DeletePropagationElement`, which is in the container, is deleted"

_And indeed local will be False by default in 4.0. I don't see why not._

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5602/commits/295bc857c59249b82dbad3c16425486ec7cedb4b
- [x] Documentation is not necessary as past behaviour considered a bug.

### Testing code

```py
from nicegui import ui


@ui.refreshable
def my_refreshable():
    ui.label('Hello, NiceGUI!')
    with ui.menu():
        ui.dialog(value=True, local=False)


@ui.page('/')
def page():
    my_refreshable()
    ui.timer(2.0, my_refreshable.refresh)


ui.run()
```

If the code works, the page should be dimmed (`ui.dialog` working), but the page should not be dimmed more than once (past `ui.dialog` cleaned up)

### DIFF
--- a/nicegui/elements/dialog.py
+++ b/nicegui/elements/dialog.py
@@ -30,7 +30,9 @@ class Dialog(ValueElement, component='dialog.js'):
         # create a canary element in the current context to trigger the deletion of the dialog when its parent is deleted
         canary = Element()
         canary.visible = False
-        weakref.finalize(canary, lambda: self.delete() if self._parent_slot and self._parent_slot() else None)
+        weakref.finalize(
+            canary, lambda: self.delete() if not self.is_deleted and self._parent_slot and self._parent_slot() else None
+        )
 
         self._result: Any = None
         self._submitted: Optional[asyncio.Event] = None


evnchn:
Should we move the class definition to elsewhere?

I shoved it in the file because I can't find a good place...

falkoschindler:
Interesting approach, @evnchn!

I was a bit confused by your testing code. So here's mine:
```py
def create_dialog():
    with ui.dialog(value=True, local=False), ui.card():
        ui.label('Hi!')

with ui.button('Open menu'):
    with ui.menu() as menu:
        ui.menu_item('Create dialog', on_click=create_dialog)

ui.button('Delete menu (and dialog) in 2 seconds', on_click=lambda: ui.timer(2, menu.delete, once=True))
```

1. Clicking "Open menu" and "Create dialog" should open a visible dialog saying "Hi!".
2. Clicking "Delete menu..." and then opening the dialog should let it disappear after 2 seconds because its parent container is deleted.

I have to think about the concept of local vs. global dialogs. Can't dialogs be always global from now on?

An alternative idea I've been carrying around for a while now (see https://github.com/orgs/zauberzeug/projects/8/views/2?pane=issue&itemId=124462692) is to create a new `ui.overlay` element that is always added to `client.layout` and is auto-deleted on close. Though I don't remember what I thought about the use case where you want to re-use a dialog. Maybe that's a bad idea because even having the sheer possibility of invisible dialogs laying around in memory allows for nasty memory leaks.

evnchn:
> Can't dialogs be always global from now on?

As mentioned somewhere, a code may leverage the existing behavior by pre-creating a dialog in a hidden container, then unhide that container for the dialog to show up. 

This PR would break their code because the dialog would show up immediately. Although, there is a strong argument to be made that this is unintended behavior and shouldn't be relied on in the first place. Not when dialog offers dedicated APIs to show/hide.

evnchn:
> `ui.overlay`

https://github.com/zauberzeug/nicegui/discussions/2040 mentioned it **once** which is not enough to get a solid idea of what it really entails. Could you kindly elaborate? 

But, if the intention is to simply place the dialog in an element created somewhere in `client.layout` and _do nothing else about it_ then this would break rule that container-clearing clears all child elements, and we still need either something similar to this PR, or the fact that it auto-deletes on close, both of which ensure a bounded lifetime of the element.

falkoschindler:
> Could you kindly elaborate [on `ui.overlay`]?

Sure. A dialog used to be an _element_ that is created in its current container, disappears if this container is deleted or temporarily removed from the DOM, and can be shown and hidden multiple times. This caused confusion like #4116. But changing the fundamental behavior of `ui.dialog` could lead to more confusion.

This is why I thought about introducing a whole new element with a different name to underline the difference to Quasar's QDialog. It would behave more like a single-use dialog that is auto-deleted when closed (like `ui.notify`) and is created globally (like `ui.header`, `ui.drawer`, ...) to avoid surprises when its parent disappears.

But introducing a new element with rather non-obvious differences also creates confusion. So maybe this PR is the better approach.

To move forward with this PR, let's

- add pytests for the new parameter,
- add a parameter description for the new parameter,
- add a deprecation marker for inverting the default behavior in 4.0, and
- move the `DeletePropagationElement` definition to the global scope to avoid redefining the class for every dialog instance.
- And I wonder if we should rename (and invert) the parameter to `top_level`. This way it would be a feature you can opt-in - which feels more intuitive to me.

evnchn:
> element which is single-use

Sorry, I would disagree: 

- Elements should be many-use as a default. 
- `ui.notify` is single-use because it isn't an element: It is in `nicegui/functions`, even. 
- If user demand single-use elements they should clear the elements themselves. 
- I don't think we should make the exception. 
- If we make an exception we may risk making many exceptions, for single-use cards, single-use XXX, etc...

For your 5 tasks, totally yes, got one query though:

- Pytests on the table
- Parameter description part of documentation
- Deprecation marker is a good idea. 
- move the `DeletePropagationElement`: **where should we move it?**
- Rename to `top_level`: While I do not see how it is having to do with opt-in/opt-out (we can just manipulate the default value), I do agree and think that the new parameter being `top_level` makes more sense than `local` because elements are implicitly `local=True` by default, if you really think about it.

falkoschindler:
I agree that my single-use `ui.overlay` has some flaws. So let's discard this idea.

I would move `DeletePropagationElement` out of the `Dialog` class, maybe at the end of the file. And we can mark it private with an underscore. We can move it somewhere else once we need it for other elements. But now it's just about not defining one class per dialog instance.

Regarding my opt-in/opt-out argument: I guess my explanation was a bit confusing, but you got the idea.

evnchn:
For the 5 tasks in https://github.com/zauberzeug/nicegui/pull/5602#issuecomment-3738670209

https://github.com/zauberzeug/nicegui/pull/5602/commits/295bc857c59249b82dbad3c16425486ec7cedb4b

Does task 1

https://github.com/zauberzeug/nicegui/pull/5602/commits/7220949d0f2ff02e2ad568c3227b5c47effbb164

Does task 2 3 4 5

falkoschindler:
@evnchn Two more thoughts:

- Layout elements like `ui.header` and `ui.drawer` are placed in `ui.context.client.layout`. Should we use this container for `ui.dialog` as well instead of `ui.context.client.content`?
- Do we really need to wait for 4.0? And do we really need the `top_level` parameter at all? The "breaking" change is that a nested dialog won't disappear when closing a menu. But this is classified as a bug. And I can't think of a reasonable scenario where this (undocumented) behavior is used productively.

evnchn:
Oh wait yes. I originally intended `ui.context.client.content` because I haven't thought of the `_DeletePropagationElement` yet and thought you can only clear the dialog with `ui.context.client.content.clear()`. Now that we have auto-element-deletion, `ui.context.client.layout` makes more sense. 

We should double-check and see if the dialog shows up fine being inside `ui.context.client.layout` though (I recall some strangeness in another PR but forgot which, not sure if this would apply)

I'm totally fine with top-level dialogs being the non-configurable default since day1 after this PR. You can make the final decision on that.

nicegui/elements/dialog.py
```diff
@@ -23,23 +21,15 @@ def __init__(self, *,
...
-        You can now create a dialog as a top level element by setting ``top_level=True``.
-        This will ensure that the dialog is always visible and not hidden when nested inside a hidden container.
-        The dialog will still be automatically deleted when the original context is deleted.
-        This behavior will be the default in NiceGUI 4.0.
-
         :param value: whether the dialog should be opened on creation (default: `False`)
-        :param top_level: whether the dialog is created as a top level element (default: `False`, *added in version 3.6.0*, will default to `True` in NiceGUI 4.0)
         """
-        super().__init__(value=value, on_value_change=None)
+        with context.client.layout:
```
evnchn, line 27: Hmm why not `self.client` as in before?

falkoschindler: `super().__init__` hasn't been called, so there is no `self.client` yet.

evnchn: Oh 😮 

Then it should be fine, since `Element` always follows `context.client` unless specified so we're doing the same thing.

falkoschindler:
I'm thinking about using a `weakref.finalizer` that calls `self.delete` when a canary element is deleted. This could replace the `_DeletePropagationElement`.
But for this to work without waiting for the garbage collector, we need to fix #5110. Which is why I finally looked into this issue again and created PR #5644. 😎

falkoschindler:
The new pytest is red until we merge PR #5644. 😬

falkoschindler (approved):
AI found another case we need to address: If the dialog is deleted programmatically, the canary is left behind. Once the menu (the dialog's parent container) is deleted, the canary dies and wants to delete the already deleted dialog. With another `if not self.is_deleted` we should be safe.

Ready to merge.

Issue #5604, opened 2025-12-29
NiceGUI documentation example images _sometimes_ breaks


evnchn:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
N/A
```

### Description

Happens every 1 in 20 loads (circa)

<img width="2940" height="1756" alt="Image" src="https://github.com/user-attachments/assets/649f75ab-85f7-4e01-90bc-b840f9d435a5" />

### NiceGUI Version

3.4.1

### Python Version

3.13.0

### Browser

Edge

### Operating System

macOS

### Additional Context

Likely due to fly.io multi-instance

PR #5605, opened 2025-12-29
Use deterministic URL generation for example image

evnchn:
### Motivation

As revealed in #5604, using auto URL for NiceGUI documentation website example images leads to issues when hosting across multiple instances. 

### Implementation

We call `app.add_media_file` directly and store the URL in the `screenshot` instead of a `Path`. 

### Progress

- [x] Pytests are not necessary for documentation website generally.
- [x] Documentation is not necessary for under-the-hood bugfix.

### DIFF
--- a/website/examples.py
+++ b/website/examples.py
@@ -25,7 +25,8 @@ class Example:
             url=f'https://github.com/zauberzeug/nicegui/tree/main/examples/{path.name}/main.py',
             screenshot=app.add_media_file(
                 local_file=path / 'screenshot.webp',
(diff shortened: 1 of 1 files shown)


evnchn:
Without https://github.com/zauberzeug/nicegui/pull/5605/commits/179b3ff1f1181bcf482ac0e658fa78f68ffc0632, it would have been no-cache. 1 day is a decent compromise. 30 days to pass Lighthouse but I don't think we can afford to do that.

PR #5610, opened 2026-01-01
Introduce `Markdown.default_extras`

evnchn:
### Motivation

In #5548, it was requested to set by-default the markdown extras, citing the influx of LLMs which forgets to add newline for a markdown list as the need driver. 

### Implementation

It is not possible to offer the `ui.markdown.extras` API since `extras` already exists per-instance. I did the next best thing: `ui.markdown.default_extras`. 

### Progress

- [x] Pytests in `tests/test_markdown.py`
- [x] Documentation is not necessary: the docstring shows up automatically.


evnchn:
Well but to what end? Do all elements will have many `default_*` `ClassVar`s lying around?

We may need a more systematic solution, otherwise we are pushing out a non-scalable API out to a single element which we have hard time withdrawing (not until 4.0)

evnchn:
Oh I have an idea now

All elements will expose `.defaults(...)` which user can call with kwargs. 

The `@resolve_defaults` decorator will be responsible to resolve the defaults both for those which are affected by `DEFAULT_PROP` but also for those which are not affected by it, overwrites the default values to follow the one set by the above `.defaults(...)` mean

This way we do not add N classvars for an element which takes N arguments.

falkoschindler:
Hm, there's another pattern we could use:

| Variable | Files |
|----------|-------|
| `LOOPBACK` | `color_input.py`, `editor.py`, `time_input.py`, `input.py`, `number.py`, `date_input.py`, `codemirror.py`, `fullscreen.py` |
| `CONTENT_PROP` | `mermaid.py`, `interactive_image.py`, `content_element.py` |
| `TEXT_COLOR_PROP` | `color_elements.py`, `badge.py`, `avatar.py`, `chip.py` |
| `BACKGROUND_COLOR_PROP` | `color_elements.py` |
| `VALUE_PROP` | `progress.py`, `codemirror.py`, `dark_mode.py` |
| `FILTER_PROP` | `filter_element.py` |
| `SOURCE_IS_MEDIA_FILE` | `video.py`, `audio.py` |
| `PIL_CONVERT_FORMAT` | `image.py`, `interactive_image.py` |

Most of these variables are mostly for internal use (`LOOPBACK` and `PROP`s), but `PIL_CONVERT_FORMAT` is - like Markdown's `DEFAULT_EXTRAS` - something a user might want to change.

Should we keep uppercase class vars for internals and deprecate `PIL_CONVERT_FORMAT` in favor of a new `.defaults(...)` API? I just don't like kwargs too much because it won't support proper linting and typing.

Or maybe the assumption that "all elements will have many `default_*` ClassVars lying around" isn't right. Maybe it's way simpler to just add static class vars like `ui.markdown.default_extras`. They are a bit hard to distinguish from methods, props and instance vars, which is probably why I chose uppercase vars for above-mentioned internals.

Let's summarize our options so far:

1. `ui.markdown.defaults(extras=...)` (and `ui.image.defaults(pil_convert_format=...)`?)
2. `ui.markdown.EXTRAS` or `ui.markdown.DEFAULT_EXTRAS`
3. `ui.markdown.default_extras` (and move towards `ui.image.pil_convert_format`?)

evnchn:
> don't like kwargs too much

Just to be clear, me neither. But I am not sure if it is a legal move mypy-speaking but can elements define their own `.defaults()` with per-element-specific arguments, and then `Element` has a `.defaults()` (or possibly ._defaults()`) which takes kwargs and is basically a fancy writer to a classvar dict? 

Maybe a good idea to try something out and see how mypy reacts because I am not too familiar with classes to know if this is a legal move or not.

evnchn:
Be sure to check out https://github.com/zauberzeug/nicegui/pull/5643 as well.

evnchn:
> legal move

No it's not. VSCode complaints, so mypy must be very upset.

evnchn:

Revived this after five months: merged current `main` in (resolving the conflict with the new `sanitize` parameter from #6082-era changes — both features now coexist in `__init__`) and updated to the `list[str] | None` union style `main` migrated to. Locally green on the full golden set: `pre-commit run`, `mypy ./nicegui` (241 files), `pylint` 10.00/10, `pytest tests/test_markdown.py` 7/7.

On the stalled design question, summarizing where the thread actually landed so we can move:

- The `.defaults(**kwargs)` alternative I floated is **not type-legal** (confirmed back in January — per-element kwargs overloads over a generic base `.defaults()` upset mypy).
- The `default_extras` ClassVar this PR ships matches the existing house pattern you tabulated (`LOOPBACK`, `CONTENT_PROP`, `TEXT_COLOR_PROP`, `VALUE_PROP`, …) — one well-named class variable, no new API machinery.

So unless you see a third shape, I'd say the PR as-is *is* the systematic solution we were looking for, and the scalability worry resolves itself by precedent. Ready for your review.

falkoschindler (approved):
Thanks for reviving this, @evnchn! I agree with your summary: with `.defaults(**kwargs)` ruled out by typing and the class-variable pattern already established in the codebase, the simple `default_extras` ClassVar is the systematic solution — no new API machinery needed. Restoring it in `nicegui_reset_globals` was a good catch, too.

I pushed a few follow-up commits:

- **Decoupled internal Markdown consumers from `default_extras`:** `ui.code` now requests `fenced-code-blocks` explicitly (otherwise e.g. `ui.markdown.default_extras = ['breaks-on-newline']` — the motivating use case from #5548 — would have broken code highlighting), and `ReStructuredText` passes `extras=[]` since it doesn't use them (a `"mermaid"` entry would have needlessly loaded Mermaid for RST elements).
- **Replaced the mermaid test parametrization with a dedicated `test_default_extras`:** it verifies that the global default actually affects rendering and that per-instance `extras` still override it, while the mermaid test goes back to testing one thing.
- **Simplified `__init__`** to resolve and copy `extras` in a single expression, with a defensive copy on both branches so neither the caller's list nor the class default can be mutated through an instance.
- Some cosmetics (docstring formatting, a clearer variable name in the test fixture).

PR #5611, opened 2026-01-01
Introduce `app.colors`

evnchn:
### Motivation

A global `app.colors()` was initially pitched back in #3787 but subsequently stalled. 

With attention towards styling NiceGUI and the uprise of coding LLMs it was mentioned a second time in #5470, which leads me to believe this is worth it of a feature to pursue. 

### Implementation

`app.colors()` directly supersedes doing `app.config.quasar_config['brand']['primary'] = 'red'`, which is rather uncontrolled and undocumented. 

- API totally mirrors `ui.colors()`. 
- Documentation added (same page as `ui.colors`)
- `app.config.quasar_config['brand']` will cease to work (**considerable breaking?**) as it inherits from `app.colors()`

### Progress

- [x] Existing pytest `test_quasar_colors` covers us for regression in setting the primary color
- [x] Pytest https://github.com/zauberzeug/nicegui/pull/5611/commits/5584022030a242ca313a698c34898dabf336e642
- [x] Documentation in `website/documentation/content/colors_documentation.py`

### DIFF
--- a/tests/test_colors.py
+++ b/tests/test_colors.py
@@ -3,14 +3,16 @@ from nicegui.testing import Screen
 
 
 def test_appwide_colors(screen: Screen):
-    app.colors(primary='#00ffff')
+    app.colors(primary='#ff0000', brand='#00ff00')
 
     @ui.page('/')
     def page():
         ui.button('Test Button')
+        ui.button('Brand Button', color='brand')
 
     screen.open('/')
-    assert screen.find_by_tag('button').value_of_css_property('background-color') == 'rgba(0, 255, 255, 1)'
+    assert screen.find_all_by_tag('button')[0].value_of_css_property('background-color') == 'rgba(255, 0, 0, 1)'
+    assert screen.find_all_by_tag('button')[1].value_of_css_property('background-color') == 'rgba(0, 255, 0, 1)'
 
 
 def test_replace_colors(screen: Screen):


falkoschindler:
Ah! `ui.colors` dynamically creates extra CSS for custom colors to define "text-..." and "bg-..." classes. We either need to implement something similar for `app.colors` or limit this PR to built-in colors only (remove `custom_colors` parameter).

evnchn:
Check if https://github.com/zauberzeug/nicegui/pull/5611/commits/6d9025c1a5645483419810afe559256dba9ff348 makes sense. By reusing existing code, we are confident (up to a certain degree) that the behaviour is the same, while expanding `app.colors` to work with arbitrary color names.

phifuh:
@falkoschindler It´s not related to this PR but is it a good idea to use so much javascript? We encountered some disconnects related to custom javascript code which the program couldnt recover after. (thats the reason i am asking this)

evnchn:
> We encountered some disconnects related to custom javascript code which the program couldnt recover after

On that I think you mean `ui.run_javascript`? 

I don't think _traditional_ JavaScript inclusion should cause much issues...

falkoschindler:
Hm, what do you mean with "so much javascript"? You can't do much in the browser without JavaScript.

phifuh:
> > We encountered some disconnects related to custom javascript code which the program couldnt recover after
> 
> On that I think you mean `ui.run_javascript`?
> 
> I don't think _traditional_ JavaScript inclusion should cause much issues...

Yes thats what I meant. I understand that js is still the goto in web but there are some projects like pyscript. I just believe its harder to track js related connection issues so my point was more about saying the more the backend can handle the better i think? 
(In our case e.x using a tkinter to get the filepath because we save that information in a data file, but if the user cancel the window we have a timeout. And we had another timeout related issue when renaming some label in our frontend)

falkoschindler:
Thanks for the suggestion. Just to make sure it lands in the right place: if you’re proposing PyScript (or similar) as a new/alternative mechanism for client-side execution, could you open a [feature request](https://github.com/zauberzeug/nicegui/discussions/categories/ideas-feature-requests) with a bit more detail on the motivation and the concrete API/UX you're envisioning?

If instead the concern is that `ui.run_javascript` (or related helpers) behaves incorrectly or unreliably in specific scenarios, please open a [bug report](https://github.com/zauberzeug/nicegui/issues) with a minimal reproducible example and relevant details (browser, NiceGUI version, etc.). That will make it much easier to investigate and fix.

PR #5612, opened 2026-01-01
Client-side input value persistence

evnchn:
### Motivation

Fixes https://github.com/zauberzeug/nicegui/issues/2149

Let us recap the problem space:

https://github.com/zauberzeug/nicegui/blob/b11972f063b976174f0cd4a4af4a9ead18ac2a10/nicegui/elements/input.js#L19-L28

- When an `ui.input` is unmounted then remounted client-side (from being inside `ui.dialog`, for example), the JS component is supposed to mount itself **without involving the server**.
- `this.value` does not change with user input (you can't even set it), only `this.inputValue` does. 
- But the value in the latter is tossed on unmount, and re-initialized to `this.value` which is stale. 

This prompts for the titled solution: **Client-side input value persistence**

### Implementation

- Identify each `ui.input` with a `internal_id`
- When the element is unmounted, keep the value with `window.__nicegui_input_values__[this.internal_id] = this.inputValue;`
- When the element is remounted, prioritize the value in `window.__nicegui_input_values__[this.internal_id]` over `this.value`

### Progress

- [ ] Oh this is going to be tough to pytest. I will try it

### Testing code

For https://github.com/zauberzeug/nicegui/discussions/1125

```py
from nicegui import app, ui

if not app.storage.general.get('name'):
    app.storage.general['name'] = 'NiceGUI User'

with ui.label() as label:
    label.bind_text_from(app.storage.general, 'name')
    with ui.element('q-popup-edit'):
        ui.input().bind_value(app.storage.general, 'name')

with ui.dialog() as mydialog, ui.card():
    ui.input().bind_value(app.storage.general, 'name')
ui.button('Edit name in dialog', on_click=mydialog.open)

ui.run()
```

### DIFF
--- a/nicegui/elements/input.js
+++ b/nicegui/elements/input.js
@@ -19,15 +19,14 @@ export default {
   props: {
     _autocomplete: Array,
     value: String,
-    internal_id: String,
+    clientStorageId: String,
   },
   unmounted() {
-    window.__nicegui_input_values__[this.internal_id] = this.inputValue;
+    setClientStorageValue(this.clientStorageId, "oldInputValue", this.inputValue);
   },
   data() {
-    if (!window.__nicegui_input_values__) window.__nicegui_input_values__ = {};
     return {
-      inputValue: window.__nicegui_input_values__[this.internal_id] || this.value,
+      inputValue: getClientStorageValue(this.clientStorageId, "oldInputValue") || this.value,
       emitting: true,
     };
   },
--- a/nicegui/elements/input.py
+++ b/nicegui/elements/input.py
@@ -1,14 +1,14 @@
-import uuid
 from typing import Any, Optional, Union
 
 from ..events import Handler, ValueChangeEventArguments
 from .icon import Icon
+from .mixins.client_storage_element import ClientStorageElement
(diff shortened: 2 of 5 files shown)


evnchn:
`window.__nicegui_input_values__` looks ugly. Can `nicegui.js` offer some basic key-value store?

evnchn:
Oh actually yes!

I think the lifetime of the persistent data on the browser should match the element in the Python-side. 

If an input element with a particular internal ID is destroyed on Python-side, then we need to inform the browser in some manner. 

Regardless I think _having_ persistent data is the method to solve the bug; As to how, we need to fine-tune.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5612/commits/16b88cdcf3aaa49d569bc903567803ce47bb5f07 should not leak, but then there is an issue: In the testing code of https://github.com/zauberzeug/nicegui/pull/5612#issue-3774629738, the input in the popup-edit and the dialog does **not** sync.

evnchn:
Since this breaks bindings, I am closing this in favour of https://github.com/zauberzeug/nicegui/pull/5652

PR #5614, opened 2026-01-01
`ui.leaflet`: invalidateSize on unhide

### LINKED ISSUE #2338: ui.leaflet and bind_visibility: tiles not loading
### Description

1. Using ui.leaflet, create a map that is initially hidden and later (when location etc are known) made visible through bind_visibility.
2. Expected behavior: when the bound value changes to True, the map is shown with the given center & zoom.
3. Instead, the map tiles are not shown -- they appear only if/when the window is resized. 

There seems to be a known issue with Leaflet when the container size is not known when initializing, but I don't know how to further isolate it for Nicegui. Likely a CSS issue?

See example below. Both maps are initially hidden. The left map is then made visible using *tailwind classes*, that works. Visibility for the map on the right hand side is enabled with *bind_visibility*; the map tiles are not correctly shown. 

Screenshot 2024-01-10 at 16 32 39

```python
from nicegui import ui


def toggle_visibility():
    a, r = ("visible", "invisible") if onoff1.value else ("invisible", "visible")
    mapcard.classes(add=a, remove=r)


onoff1 = ui.switch("via tailwind classes", value=False, on_change=toggle_visibility)

onoff2 = ui.switch("via bind_visibility", value=False)

with ui.card().classes("w-full flex flex-row"):
    with ui.card().classes("w-1/3 flex invisible") as mapcard:
        m = ui.leaflet(center=(51.505, -0.09)).classes("h-screen visible")

    with ui.card().classes("w-1/3 flex").bind_visibility_from(onoff2, "value"):
        m = ui.leaflet(center=(51.505, -0.09)).classes("h-screen")

ui.run()
```
The flex classes are only for display here, AFAICT the difference between the two methods remains with all styling I've tried. As a workaround I can work with the Tailwind classes, but that is of course a bit more clumsy.

evnchn:
### Motivation

Fixes #2338, in which we see that `ui.leaflet` broken if it was hidden on creation, then subsequently unhidden. 

While Leaflet 2.0 fixes it, the [alpha was in May last year but now we are in January the next year](https://leafletjs.com/2025/05/18/leaflet-2.0.0-alpha.html)...

### Implementation

- [Uses an IntersectionObserver to detect unhide](https://stackoverflow.com/a/74967237)
- Run `this.run_map_method("invalidateSize")` on unhide
- Be sure to disconnect the observer on unmount

### Progress

- [x] Pytest https://github.com/zauberzeug/nicegui/pull/5614/commits/d12b4108a54be17250e5831cc59e9a6180ba345e

### DIFF
--- a/tests/test_leaflet.py
+++ b/tests/test_leaflet.py
@@ -57,7 +57,7 @@ def test_leaflet_unhide(screen: Screen):
     @ui.page('/')
     def page():
         with ui.card().classes('w-full h-64') as card:
-            ui.leaflet().wms_layer('/mock_tile/{{z}}/{{x}}/{{y}}')
+            ui.leaflet().wms_layer(url_template='/mock_tile/{{z}}/{{x}}/{{y}}')
             card.visible = False
         ui.button('Show map card', on_click=lambda: card.set_visibility(True))
 


falkoschindler (commented):
Thanks for the pull request, @evnchn! Finally fixing this issue would be great!

Here is a quick reproduction for reference (map tiles missing when clicking "Show map card"):
```py
ui.button('Show map card', on_click=lambda: card.set_visibility(True))
with ui.card() as card:
    ui.leaflet(center=(51.505, -0.09)).classes('w-96 h-64')
    card.set_visibility(False)
```

I refactored the observer and the pytest a little. I don't think it change any behavior. But then I noticed that the test is failing. Maybe it is just flaky and depends on external server availability and/or rate limits. This is why I'm worried that this test will haunt us with flakiness in our CI pipeline. But I'm not sure if there's an "offline" way to test it. 🤔

evnchn:
> Tests failing



... 

Offline-testing sure sounds like a good idea. Let's see how.

falkoschindler (approved):

Issue #5617, opened 2026-01-02
No error page handling with async sub-pages


CatamountJack:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python

```

### Description

When using sub-pages (single page apps) with async functions, errors do not trigger handling with an error page.

Async functions seem to work just fine otherwise, but I just ran into this in my app where I'm raising authentication errors when a user's session is expired, but I can't figure out how to catch it in a way I can redirect them to login again.  

Using the [single page app example](https://github.com/zauberzeug/nicegui/tree/main/examples/single_page_app), if you simply change the `def error():` function in `main.py` to `async def error():`, no error page will be displayed and the following is logged to the console:

```py
some error message
Traceback (most recent call last):
  File "C:\dev\.venv\Lib\site-packages\nicegui\background_tasks.py", line 93, in _handle_exceptions
    task.result()
    ~~~~~~~~~~~^^
  File "C:\dev\.venv\Lib\site-packages\nicegui\elements\sub_pages.py", line 116, in background_task
    await result
  File "C:\dev\test\main.py", line 38, in error
    raise ValueError('some error message')
ValueError: some error message
```

#### Example:

*main.py* **(change made on line 38)**
```py
#!/usr/bin/env python3
from custom_sub_pages import custom_sub_pages, protected

from nicegui import app, ui


@ui.page('/')
@ui.page('/{_:path}')
def main_page():
    with ui.header().classes('items-center bg-blue-100'):
        ui.button('Home', on_click=lambda: ui.navigate.to('/')).props('flat')
        ui.button('Secret', on_click=lambda: ui.navigate.to('/secret')).props('flat')
        ui.button('Invalid', on_click=lambda: ui.navigate.to('/invalid')).props('flat')
        ui.button('Error', on_click=lambda: ui.navigate.to('/error')).props('flat')
        ui.space()
        ui.button('Logout', icon='logout').props('flat') \
            .bind_visibility_from(app.storage.user, 'authenticated') \
            .on_click(lambda: app.storage.user.update(authenticated=False)) \
            .on_click(lambda: ui.navigate.to('/'))

    custom_sub_pages({
        '/': home,
        '/secret': secret,
        '/error': error,
    }).classes('flex-grow p-4')


def home():
[...]

PR #5618, opened 2026-01-03
A backwards-compatible and customizable API to handle async page errors

evnchn:
### Motivation

Back in https://github.com/zauberzeug/nicegui/pull/4538#issuecomment-2967482671 we had functionality to capture the error of a page, and what we did with it was to toss it to the frontend using a _highly-dedicated delivery channel_ (with a custom `server_error` Socket.IO message handled in the frontend, even)

It was indeed quite overengineered and rigid, has a big diff, and was shortly dropped. Partly why my original stance in https://github.com/zauberzeug/nicegui/issues/5617 was to simply let the fact that async page errors (specifically the post-connect ones) be unhandled. However, the more I look at the issue, the more my stance changed, especially in light of `ui.sub_pages` which have almost all UI logic to be ran post-connect in daily use. 

### Implementation

With an eureka moment, I believe I have provided a sufficiently **backwards-compatible**, **customizable**, **opt-in** and **small-diff** API to handle async page errors for NiceGUI to fill up that functionality gap. 

The core lies in `client._emit_error`, which is a private method which is similar in spirit to the trick in https://github.com/zauberzeug/nicegui/pull/5490, in that it emits an event as if it came from the client despite it actually not. 

This is used to create a phantom event `__error__` whose args is the error itself. 

Handler of the event may choose to:

- Do nothing (default): the error still gets logged anyways
- Popup an error dialog in a fashion similar to other full-stack UI frameworks
- Clear the page of content to forbid the user from progressing further
- Log the error somewhere else **though this is not recommended in my personaly opinion, use `@app.exception`**
- A mix of the above, depending on a criteria of the user's liking, such as what kind of error it is

It is highly-flexible. That's the point. Partly with that to thank, the diff is small.  

### Notes

`__error__` was chosen because:

1. `ui.on` targets the JS counterpart `emitEvent` which user may emit anything. A sufficiently unique name is demanded else our event collides with existing userland evemts. 
2. the `__` is highly related to Python

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5618/commits/2fcf40eeabb264835936685a2caf6355c86a5ac2
- [x] Documentation https://github.com/zauberzeug/nicegui/pull/5618/commits/a94b5157574ad3bfd53c5247e26ed07c69ecbb37

~~Should we consume #5490 and this PR together, then refactor the mocking into `_mock_event` for DRY?~~ Not relevant anymore

### Demo code

```py
import asyncio
import traceback

from nicegui import ui

def render_error_details(error, code_classes=''):
    ui.label('Page error').classes('text-lg font-bold')
    ui.label(f'{error} ({type(error).__name__})').classes('text-red-600')
    ui.code(traceback.format_exc(chain=False)).classes(code_classes)

async def error_out():
    ui.label('This @ui.page errors out post-HTML-response in 3 seconds')
    await asyncio.sleep(3)
    raise ValueError('Test exception handling')

@ui.page('/')
def menu():
    ui.link('@ui.page raises error, shows error dialog', '/1')
    ui.link('@ui.page raises error, clears the body and shows the error dialog', '/2')
    ui.link('ui.sub_pages: sub page without error', '/3')
    ui.link('ui.sub_pages: sub page with error', '/4')

@ui.page('/1')
async def main_page1():
    menu()
    with ui.page_sticky(x_offset=16, y_offset=16):
        fab_error = ui.button(icon='error', color='negative').props('fab')
        fab_error.set_visibility(False)

    def show_error_dialog(error):
        fab_error.set_visibility(True)
        with ui.dialog() as error_dialog, ui.card():
            render_error_details(error, 'max-w-[calc(560px-2rem)]')
            ui.button('Close', on_click=error_dialog.close)
        fab_error.on('click', error_dialog.open)

    ui.on('__error__', lambda e: show_error_dialog(e.args))
    await error_out()


@ui.page('/2')
async def main_page2():
    menu()

    def clear_content_and_show_error(error):
        with ui.context.client.content.clear():
            render_error_details(error, 'w-full')
            ui.link('Back to menu', '/')

    ui.on('__error__', lambda e: clear_content_and_show_error(e.args))
    await error_out()

@ui.page('/3')
@ui.page('/4')
async def main_page3_4():
    menu()
    ui.on('__error__', ui.notify)  # Lazy treatment

    async def error():
        ui.label('This page will raise an error.')
        raise ValueError('This is a test error on the /error page')

    ui.sub_pages({'/3': lambda: ui.label('Sub page without error.'), '/4': error})

ui.run()
```

### Final notes

I'm quite proud of this PR honestly. Thank you @CatamountJack for sufficiently challenging the status quo. You must be responsible for the eureka moment I had.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -284,7 +284,7 @@ class Client:
         self.delete_handlers.append(handler)
 
     def on_exception(self, handler: Callable[[Exception], Any] | Callable[[], Any]) -> None:
-        """Add a callback to be invoked when non-critical exceptions occur in the client's UI context.
+        """Add a callback to be invoked for in-page exceptions (after the page has been sent to the browser).
 
         The callback has an optional parameter of `Exception`.
         """
@@ -381,7 +381,7 @@ class Client:
         self.remove_elements(self.elements.values())
 
     def handle_exception(self, exception: Exception) -> None:
-        """Handle a non-critical exception to be handled by callers of `ui.on_exception(...)`."""
+        """Handle an in-page exception by invoking handlers registered via `ui.on_exception(...)`."""
         for handler in self._exception_handlers:
             with self.content:
                 if helpers.expects_arguments(handler):
--- a/nicegui/functions/on_exception.py
+++ b/nicegui/functions/on_exception.py
@@ -4,7 +4,7 @@ from ..context import context
 
 
 def on_exception(handler: Union[Callable[[Exception], Any], Callable[[], Any]]) -> None:
-    """Register a handler for non-critical exceptions occurring in the current page's UI context.
+    """Register a handler for in-page exceptions (after the page has been sent to the browser).
 
     The callback has an optional parameter of ``Exception``.
 
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -145,7 +145,7 @@ def test_exception(screen: Screen):
     screen.should_contain('500')
     screen.should_contain('Server error')
     screen.assert_py_logger('ERROR', 'some exception')
-    assert not exceptions, 'ui.on_exception is used for non-critical exceptions'
+    assert not exceptions, 'ui.on_exception is for in-page exceptions (after page sent to browser)'
 
 
 def test_exception_after_connected(screen: Screen):
@@ -161,7 +161,7 @@ def test_exception_after_connected(screen: Screen):
     screen.open('/')
     screen.should_contain('this is shown')
     screen.assert_py_logger('ERROR', 'some exception')
-    assert exceptions, 'non-critical exception should be caught by ui.on_exception'
+    assert exceptions, 'in-page exception should be caught by ui.on_exception'
 
 
 def test_page_with_args(screen: Screen):
--- a/website/documentation/content/overview.py
+++ b/website/documentation/content/overview.py
@@ -303,7 +303,7 @@ def map_of_nicegui():
             - [`ui.navigate`](/documentation/navigate): let the browser navigate to another location
             - [`ui.notify`](/documentation/notify): show a notification
             - [`ui.on`](/documentation/generic_events#custom_events): register an event handler
-            - [`ui.on_exception`](/documentation/on_exception): register a handler for non-critical exceptions occurring in the current page's UI context
+            - [`ui.on_exception`](/documentation/on_exception): register a handler for in-page exceptions (after the page has been sent to the browser)
             - [`ui.page_title`](/documentation/page_title): change the current page title
             - [`ui.query`](/documentation/query): query HTML elements on the client side to modify props, classes and style definitions
             - [`ui.run`](/documentation/run) and `ui.run_with`: run the app (standalone or attached to a FastAPI app)
--- a/website/documentation/content/section_action_events.py
+++ b/website/documentation/content/section_action_events.py
@@ -110,18 +110,18 @@ doc.intro(event_documentation)
 doc.text('Error handling', '''
     There are 3 error handling means in NiceGUI:
 
-    1. [`app.on_exception`](#lifecycle_events): Global exception handler
-        - Works for all exceptions in the NiceGUI app.
+    1. [`app.on_exception`](#lifecycle_events):
+        Global exception handler for **all** exceptions in the NiceGUI app.
         - Applied app-wide.
         - Handler has no UI context (cannot use `ui.*`).
         - Common sources: `app.timer`, `background_tasks.create`, `run.io_bound`, `run.cpu_bound`.
-    2. [`@app.on_page_exception`](#custom_error_page): Custom error page for critical exceptions
-        - Works for UI-context exceptions raised **before** the Python-side client is ready.
+    2. [`@app.on_page_exception`](#custom_error_page):
+        Custom error page for page-blocking exceptions (**before** page sent to browser)
         - Applied app-wide.
         - Handler may use UI elements but in a new client.
         - Common sources: sync `@ui.page` functions, exceptions in async `@ui.page` functions before `await ui.context.client.connected()`.
-    3. [`ui.on_exception`](#ui_on_exception): Page exception handler for non-critical exceptions
-        - Works for UI-context exceptions raised **after** the Python-side client is ready.
+    3. [`ui.on_exception`](#ui_on_exception):
+        Handler for in-page exceptions (**after** page sent to browser)
         - Applied per-page.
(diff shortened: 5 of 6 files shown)


CatamountJack:
With a *very* quick test, this seems to work nicely.  My app is sub-classing `ui.sub_pages` (based directly on the [Single Page App example](https://github.com/zauberzeug/nicegui/tree/main/examples/single_page_app)) and it appears I can add `ui.on('__error__', lambda e: self.clear_content_and_show_error(e.args))` in the `def _render_page()` method to catch and handle the error from any sub-page.  Per-previous discussion, my use-case was to check for user authentication within the async sub-pages (raise exception when unauthorized) and catch it a higher level so could redirect to a login page - and it seems to work well for that.

I'll do more real testing in the next day or so.  Thanks!!

evnchn:
@CatamountJack I changed the API away from `ui.on('__error__', ...)` to `ui.on_exception(...)`. 

If your code works on this PR's branch you will need to update your code.

CatamountJack:
Working well here so far.  Curious though, is it possible to suppress logging the exception when it's handled by `ui.on_exception()`?  I'm handling them, but I'm still getting tracebacks to the console.

evnchn:
@CatamountJack There is no such provision for now, and is not planned in this PR. 

You can use this for now if you wish:

```py
from nicegui import app, context, logging, ui

app._exception_handlers.clear()

@app.on_exception
def my_log_exception(e):
        return  # processed by ui.on_exception
    logging.log.exception(e)
```

But this does not work for errors from `ui.sub_pages` since it directly invoked `log.error` 

https://github.com/zauberzeug/nicegui/blob/049bae14ab4005a47766a1fd0c02bdb244e31b7d/nicegui/elements/sub_pages.py#L136

For having the console be silent for handled errors, we can embark on a separate PR. 

I envision:

- either the UI-level handlers (`ui.on_exception`, `@app.on_page_exception`) may `return True` if they deem that the error has been sufficiently handled. 
  - However for `ui.on_exception` it currently supports async handlers, and for those it is "too late to return True" unfortunately. We may either reserve the ability to block-off the `app.on_exception` handler 
- Or we just have a global switch
  - but the switch name is hard to name

evnchn:
While trying to investigate "is it possible to suppress logging the exception", I found an interesting phenomenon. 

1. On error call `client._emit_error(e)`
2. `_emit_error` calls `handle_event(listener.handler, event_args)`
3. `handle_event` on exception calls `core.app.handle_exception(e)`
4. `handle_exception` on exception calls `context.client._emit_error(exception)`

So, **if the handler attached via `ui.on_exception` errors out, we risk an infinite loop.**

We need to test if it is the case, and if it is we need to fix it until it isn't. 

---

And, to supress logging the exception, `handle_event` shallr return the value of the handler if it is sync handler. This was we capture `return True` as a sign to supress logging the exception.

falkoschindler:
Thanks @evnchn, here are some first review findings:

1. Infinite loop vulnerability (as you noted in the PR comments):

    The flow `_emit_error` → `handle_event` → `app.handle_exception` → `_emit_error` creates an infinite loop if the error handler itself throws.
    The fix might be simple: I _think_ `handle_exception` shouldn't emit the error event, because it is already an exception handler.

2. `raise e` should be `raise` in sub\_pages.py:

    Using `raise e` resets the traceback, losing context.

3. Documentation clarity: The mutual exclusivity explanation could be slightly clearer.

    Consider:

    - All exceptions are logged via `app.on_exception` (1)
    - UI-context exceptions go to *either*:
      - `@app.on_page_exception` (2) if raised before client connection
      - `ui.on_exception` (3) if raised after client connection

4. Error handlers instead of "\_\_error\_\_" event?

    Instead of using a pseudo "\_\_error\_\_" event, wouldn't it be cleaner to implement independent error handlers? I mean `on_exception` would add a handler to a list and instead of `_emit_error` we would call all handlers.

5. Naming:

    I'd suggest to talk about exceptions, not errors. So it would be an "\_\_exception\_\_" event and we would call `_emit_exception`. But see point 4 above.

evnchn:
> Error handlers instead of "__error__" event?

This is good since currently the error is in `e.args` but you would expect it to be in `e`. It's been an itch but I just tolerated it for the time being 😂 

I think this could also help us rid the infinite loop as well. However `handle_exception` would still need to invoke the handler since it is responsible for capturing errors in `ui.button` callbacks (iirc).

evnchn:
No infinite loop issue now. `During handling of the above exception, another exception occurred:` saved us.

nicegui/client.py
```diff
@@ -383,9 +383,15 @@ def remove_all_elements(self) -> None:
...
         for handler in self._exception_handlers:
-            result = handler() if not inspect.signature(handler).parameters else handler(exception)
+            if helpers.expects_arguments(handler):
+                result = cast(Callable[[Exception], Any], handler)(exception)
+            else:
+                result = cast(Callable[[], Any], handler)()
             if helpers.is_coroutine_function(handler):
-                background_tasks.create(result, name=f'UI exception {handler.__name__}')
+                async def wait_for_result(result: Any = result) -> None:
+                    with self.layout:
```
evnchn: Oh this one I think `self.content` is more fair, in case one would prefer a `document.write` style error log which appends the error beneath the page. 

Throughout the codebase, entering the layout element is done by code which needs it. 

image

evnchn: And we may have to `with` before invoking even a sync `handler` to ensure uniform context among sync and async handling.

falkoschindler:
Thanks, @evnchn, I'll use the context `content` for both cases.

One final thought: The term "non-critical exception" feels a bit unintuitive. Who is to judge which exception is critical and which not. I think we should use a term independent of the severity.
What do you think about "in-page exceptions"?

evnchn:
"in-page exceptions" is very good, much better than "critical", but in the context of `@ui.page` can get confusing. 

Why is it that:

```py
@ui.page('/') # this is a page, isn't it?
async def main_page():
    # if I error here, this is "out of page"?
    # suddenly this is "in-page"?
```

We need to be extra clear in the phrasing that we refer to the browser-page, not the Python-page.

evnchn:
The combo of "page-blocking & in-page" cleared everything up 🤩

nicegui/client.py
evnchn: Done in https://github.com/zauberzeug/nicegui/pull/5618/commits/df532f02144ee1a5e328b9b2f1236effbefcfea9

PR #5629, opened 2026-01-09
Allow bring-your-own-aggrid

evnchn:
### Motivation

As mentioned by @falkoschindler at https://github.com/zauberzeug/nicegui/discussions/2431#discussioncomment-15455977, the [AG Grid Enterprise EULA](https://www.ag-grid.com/eula/AG-Grid-Enterprise-License-Latest.html), and clause 3.7(b) explicitly prohibits redistribution as part of:

> "any application that can be described as a development toolkit or library, an application builder, a website builder, a user interface designer, or any application intended for use by software, application, or website developers or designers"

For which, NiceGUI and any extensions fall squarely into. 

While this may seem like a dead end for NiceGUI implementing AG Grid Enterprise, I have thought of a way: what if we never redistributed it, but just linked to a CDN? Hence this PR. 

**Real motivation:** I got a reply ["Not quite sure what you mean..."](https://github.com/zauberzeug/nicegui/discussions/2431#discussioncomment-15456797) and a PR is the best way to clarify. 

### Implementation

- https://github.com/zauberzeug/nicegui/commit/7faef362ffa1c619b64e071e77ba917e1df7bd4d removes the AG Grid module registration from the bundled AG Grid community, moves away from default exports, and got rid of the themes dictionary. 
  - This brings our `nicegui-aggrid` much more in line with a plain AG Grid ESM module, to the point it is compatible with JSDelivr bundles.
- New parameters `modules` used to control what modules to load
- New function `register_importmap_override` to override the `nicegui-aggrid`

### Demo code

```py

from nicegui import ui

ui.aggrid.set_esm_module('https://cdn.jsdelivr.net/npm/ag-grid-enterprise@35.0.0/+esm')

def create_aggrid_page(modules):
    grid = ui.aggrid({
        'columnDefs': [
        ],
        'rowData': [
            {'name': 'Alice', 'age': 18, 'parent': 'David'},
            {'name': 'Bob', 'age': 21, 'parent': 'Eve'},
            {'name': 'Carol', 'age': 42, 'parent': 'Frank'},
        ],
        'rowSelection': {'mode': 'multiRow'},
    }, modules=modules)

    def update():
        grid.options['rowData'][0]['age'] += 1

    ui.button('Update', on_click=update)
    ui.button('Select all', on_click=lambda: grid.run_grid_method('selectAll'))
    ui.button('Show parent', on_click=lambda: grid.run_grid_method('setColumnsVisible', ['parent'], True))

@ui.page('/')
def aggrid_page():
    create_aggrid_page(modules=['AllCommunityModule'])

@ui.page('/enterprise')
def aggrid_enterprise_page():
    create_aggrid_page(modules=['AllEnterpriseModule'])

ui.run()
```

### Progress

- [x] Pytest https://github.com/zauberzeug/nicegui/pull/5629/commits/db137739718081aef2f8e27eb3b5edf86b4b7579
- [x] Documentation via docstring: https://github.com/zauberzeug/nicegui/pull/5629/commits/4cae858184efa36e0dcf552aaebf3afeef0c7d1d

### DIFF
--- a/nicegui/elements/aggrid/aggrid.py
+++ b/nicegui/elements/aggrid/aggrid.py
@@ -105,8 +105,7 @@ class AgGrid(Element, component='aggrid.js', esm={'nicegui-aggrid': 'dist'}, def
             'suppressFieldDotNotation': True,
             **options,
             'theme': theme or options.get('theme', 'quartz'),
-            'modules': modules,
-        }, html_columns=html_columns, theme=theme, auto_size_columns=auto_size_columns)
+        }, html_columns=html_columns, theme=theme, auto_size_columns=auto_size_columns, modules=modules)
 
     @classmethod
     def from_polars(cls,
@@ -138,8 +137,7 @@ class AgGrid(Element, component='aggrid.js', esm={'nicegui-aggrid': 'dist'}, def
             'suppressFieldDotNotation': True,
             **options,
             'theme': theme or options.get('theme', 'quartz'),
-            'modules': modules,
-        }, html_columns=html_columns, theme=theme, auto_size_columns=auto_size_columns)
+        }, html_columns=html_columns, theme=theme, auto_size_columns=auto_size_columns, modules=modules)
 
     @property
     def options(self) -> dict:


evnchn:
Need https://github.com/zauberzeug/nicegui/pull/5629/commits/37941c4162e7195a88f904ae9c45496262f46dcd because of failing test

evnchn:
I believe that, generally speaking, moving to a plain `export * from "BLAH";` is a good approach in ensuring the libraries that NiceGUI depends upon can be easily swapped for a newer version. Because that is the most direct and transparent way of exporting the entire library un-touched and un-remapped. 

Any mapping that we do (e.g. `export { default as mermaid } from "mermaid/dist/mermaid.esm.min.mjs";` to facilitate `import { mermaid } from "nicegui-mermaid";`) makes it impossible to use the mermaid from JSDelivr.

evnchn:
`modules` is a list because you can load only some of the modules for a lighter experience:


https://www.ag-grid.com/javascript-data-grid/modules/

The `All*Module` is a shorthand for loading all of the modules. Enterprise loads also the Community.

falkoschindler (changes requested):
1. Ok, `Literal['AllCommunityModule', 'AllEnterpriseModule']` could be a bit too restrictive if we want to allow customizing the exact set of loaded modules. But do we really need that? Don't we actually want to distinguish two modes only? Maybe an `enterprise: bool = False` would suffice?
2. See 1.
3. Let's rename the `import_name` parameter and extend the docstring, maybe
    ```py
    """Override the ESM module URL for all AG Grid elements.
    
    This sets a global import map override, affecting all pages and clients.
    Use this to switch to AG Grid Enterprise or a self-hosted bundle.
    
    :param import_name: the ESM module URL (e.g., "https://cdn.jsdelivr.net/npm/ag-grid-enterprise@35.0.0/+esm")
    """
    ```
4. How could we test this without depending on quick requests to jsdelivr and without bundling the enterprise version? Maybe just another mock endpoint to check if the request happens?
5. `from_pandas` and `from_polars` are still missing the `modules` parameter (or `enterprise`).

evnchn:
> `enterprise: bool = False`

Not a good idea in case AG Grid comes up with more tiers of offering (think "AG Grid Premium"). Still simplicity is key and `AllCommunityModule` is one heck of a name. I came up with a compromise. See what you think. 

Your new docstring is adopted, but it may be a bad idea to condone using AG Grid 35.0.0 owing to us being in 34.2.0 and (1) AG Grid isn't known for backwards-compatibility in a minor release, (2) 35.0.0 is a major bump.

PR #5631, opened 2026-01-11
Avoid duplication in methods documentation and use the best docstring available

### LINKED ISSUE #5630: Duplicated overridden methods in documentation
### First Check

### Example Link

https://nicegui.io/documentation/matplotlib

### Description

The `update()` method shows up twice. 

### NiceGUI Version

3.5.0

### Python Version

3.14.0

### Browser

Safari, Edge

### Operating System

Windows

### Additional Context

This regression is traced to #5484

However, before the regression, it only shows the `update()` with no docstring from `ui.matplotlib` not the inherited `ui.element` one with the docstring, so either way the implementation isn't perfect. 

CC @himbeles

evnchn:
### Motivation

Fixes #5630, which boils down to:

- We would not want duplicated methods in documentation. 
- When a method is overridden without docstring, the docstring of the original method should not be masked. 

### Implementation

- Use the alternative docstring from the inherited method if the native method does not have docstring
- Stop showing the inherited method if the native method is already shown

### Limitation

This code would not well-handle the case where multiple levels of method override, each level having its own docstring. In this implementation, only one docstring would show from the many levels, and there isn't concatenation logic. 

But I decided to let it slide because there isn't such thing in the docs right now. 

### Progress

### DIFF
--- a/website/documentation/reference.py
+++ b/website/documentation/reference.py
@@ -50,21 +50,15 @@ def generate_class_doc(class_obj: type, part_title: str) -> None:
         ui.markdown('\n'.join(f'- `{ancestor.__name__}`' for ancestor in ancestors))
 
 
-def _get_doc(obj: object) -> str:
-    return getattr(obj, '__doc__', '') or ''
-
-
 def _render_section(class_obj: type, attributes: list[Attribute], *, method_section: bool) -> None:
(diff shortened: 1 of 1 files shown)


evnchn:
I am not sure if there are edge cases to this PR's code.

PR #5632, opened 2026-01-11
ui.log scroll-to-bottom working in ui.tab_panel using onScroll

evnchn:
### Motivation

Fix #5118, where we find #5112 (or, actually, the reimplementation of it in 3.0, since we do `q-scroll-area` for `ui.log`) does not work inside `ui.tab`. 

### Implementation

Compared to the predecessor, #5220, the following are the realizations:

- The core reason behind flaky `verticalPercentage` measurements is that, during the animation, `clientHeight` degenerates to 0px tall. We can ignore it with `if (!this.$refs.qRef.$el.childNodes[0].clientHeight) return;`
- During tab animate-in, the DOM during `beforeUpdate` is really messy. I just cannot find a reliable measure. 
- Use `onScroll`, because the user's intention on whether to scroll-to-bottom changes on scroll. 

### Progress

- [x] We probably can't easily pytest this...

### DIFF
--- a/nicegui/elements/log.js
+++ b/nicegui/elements/log.js
@@ -7,13 +7,12 @@ export default {
     };
   },
   updated() {
-    if (this.$refs.qRef && this.shouldScroll) {
+    if (this.shouldScroll) {
       this.$nextTick(() => this.$refs.qRef.setScrollPosition("vertical", Number.MAX_SAFE_INTEGER));
     }
   },
   methods: {
     onScroll(info) {
-      if (!this.$refs.qRef) return;
       if (info.verticalContainerSize === 0 || info.horizontalContainerSize === 0) return;
       if (info.verticalContainerSize !== this.lastHeight) {
         this.lastHeight = info.verticalContainerSize;


falkoschindler:
@evnchn I'm a bit confused what exactly this PR is trying to achieve.

I tried to
1. run the code from #5118
    ```py
    with ui.left_drawer().props('width=200').classes('bg-secondary'):
        with ui.tabs().props('vertical').classes('w-full') as tabs:
            one = ui.tab('One')
            two = ui.tab('Two')
    
        with ui.tab_panel(one):
            log = ui.log().classes('w-full h-30vh')
        with ui.tab_panel(two):
            ui.label('nicegui is my favorite.').classes('text-2xl')
    
    def write_log():
        log.push(f'Log entry at {time.strftime("%H:%M:%S")}')
        log.push(f'Log entry at {time.strftime("%H:%M:%S")}')
        log.push(f'Log entry at {time.strftime("%H:%M:%S")}')
        log.push(f'Log entry at {time.strftime("%H:%M:%S")}')
    
    ui.timer(0.5, write_log, active=True)
    ```
2. wait for the log to scroll (sticking to the bottom),
3. switch to tab 2 and wait a moment,
4. switch back to tab 1.

No the view doesn't stick to the bottom anymore. But it should, right?

evnchn:
My code works on the demo in https://github.com/zauberzeug/nicegui/pull/5220#pullrequestreview-3349832332, Microsoft Edge

If it doesn't work we'll have to find another heuristic. It's much trial and error but the goal is, should scroll to bottom if it was scroll to bottom before.

evnchn:
https://github.com/zauberzeug/nicegui/pull/5632/commits/c3a6545b09a49ac1976929967cb0a38738b1f517 tested against https://github.com/zauberzeug/nicegui/pull/5632#issuecomment-3745001316, and https://github.com/zauberzeug/nicegui/pull/5220#pullrequestreview-3349832332

Both works on my machine. 

The heuristics are: 

- If container height or width is 0, give up
- If the container height changed (tab change), then skip setting `this.shouldScroll` once.

falkoschindler (commented):
Great, @evnchn! Now it's working nicely!

Just one question: Is `if (!this.$refs.qRef) return;` really needed? Shouldn't `this.$refs.qRef` always be defined inside an event handler?

evnchn:
`if (!this.$refs.qRef) return;`

But it was there for updated and the old beforeUpdate so I kept it. Not sure. You may want to ask Claude or ChatGPT on that.

evnchn:
https://vuejs.org/guide/essentials/template-refs#accessing-the-refs

Well then, I think we are safe to not only drop that from `onScroll` but also `mounted` since `updated` run after mount as in https://vuejs.org/guide/essentials/lifecycle#lifecycle-diagram

I will commit first. Drop it if it breaks. Thanks!

Issue #5033, opened 2025-08-12
Plotly animation play button doesn't start the animation


fdion:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui
import plotly.express as px

df = px.data.gapminder()

fig = px.scatter(df, x="gdpPercap", y="lifeExp", animation_frame="year", animation_group="country",
           size="pop", color="continent", hover_name="country", title="Life expectancy vs gdp per cap",
           log_x=True, size_max=55, range_x=[100,100000], range_y=[25,90])

ui.plotly(fig)

ui.run()
```

### Description

Run the code. Click on the "play" button, no animation. This is the case in native or with various web browsers.

### NiceGUI Version

2.22.2

### Python Version

3.11.13

### Browser

Safari

### Operating System

macOS

### Additional Context

Also ran in chrome, and under windows, same problem.

No messages in the javascript console.

falkoschindler:
Thanks a lot for reporting this issue, @fdion!

It turned out the figure hasn't correctly been passed to Plotly.js. Instead of
```js
Plotly.react(this.$el.id, this.options.data, this.options.layout);
```
we need to call
```js
Plotly.react(this.$el.id, this.options, options.config);
```
which includes `this.options.frames`.

It should be fixed on main and will be released in version 2.23.0.

PR #5633, opened 2026-01-11
Do not run `update_method` for not-yet-mounted elements

evnchn:
### Motivation

In #5033 we see that elements with `update_method` throwing error when `update` is eventually invoked and the element is not yet mounted (maybe inside `tab_panel`)

### Implementation

Well, just don't invoke it. 

- There's no point since the element is not yet mounted, no work to do. 
- By the time the element gets mounted, it can render itself fine without extra invocations to `update_method`

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5633/commits/128cee273d3732cf4033eda5432144199c53a72e

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -410,8 +410,8 @@ function createApp(elements, options) {
 
           await this.$nextTick();
           for (const [id, element] of Object.entries(msg)) {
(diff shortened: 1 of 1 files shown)


evnchn:
I didn't know you can `?.[` 😅

PR #5636, opened 2026-01-12
Allow propagation of nested updates back to frontend in AnyWidget

### LINKED ISSUE #5626: AnyWidget nested updates don't propagate back to frontend
### First Check

### Example Code

```python
import anywidget as aw
from traitlets import Int
from nicegui import ui

class UpdateWidget(aw.AnyWidget):
    _esm = """
    function render({model, el}) {
        const div = document.createElement("div");
        div.innerText = "nothing";
        el.appendChild(div);
        model.on("change:b", () => {
            const b = model.get("b");
            div.innerText = `b=${b}`;
        });

        model.set("a", 1);
        model.save_changes();
    }

    export default { render };
    """
    a = Int(0).tag(sync=True)
    b = Int(0).tag(sync=True)

uw = UpdateWidget()
def change_b(change):
    uw.b = 1
uw.observe(change_b, names=["a"])
ui.anywidget(uw)
ui.run()
```

### Description

The browser should show `b=1`, not `nothing`. It remains as `nothing` because the update is never sent.

### NiceGUI Version

3.5.0

### Python Version

Python 3.11.14

### Browser

Chrome

### Operating System

macOS

### Additional Context

This breaks altair when using the vegafusion data_transformer.

Introduced in 5512bff0af6e98cee221c9c1d951f77227d37188 :'), I have a fix ready: #5627.

s-meza:
### Motivation

Closes #5626

As mentioned there, altair charts don't render when using the vegafusion data_transformer. I think it uses traitlets to communicate with the frontend.

Altair Example:
```python
from nicegui import ui
import altair as alt
import pandas as pd

alt.data_transformers.enable("vegafusion")
c = alt.Chart(pd.DataFrame({"data": [0,1,2,3]})).mark_point().encode(x="data")

ui.altair(c)
ui.run()
```

### Implementation

It keeps track of the current dictionary of changes made as a result of the front-end code, and avoids calling `update_traits` if there's been no new changes.

The one concern I have with NaNs, since NaNs are always not-equal to each other, the check will return True, perhaps causing more updates than intended.

### Progress

- [X] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.

### DIFF
--- a/nicegui/elements/anywidget/anywidget.py
+++ b/nicegui/elements/anywidget/anywidget.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import importlib.util
 import inspect
+import math
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
@@ -90,5 +91,7 @@ def _equal(a: Any, b: Any) -> bool:
     """Check if two values are equal, considering NaN as equal."""
     if a == b:
         return True
-    # could still be equal if both are NaN
-    return a != a and b != b  # pylint: disable=comparison-with-itself # noqa: PLR0124
+    try:
+        return math.isnan(a) and math.isnan(b)
+    except TypeError:
+        return False


falkoschindler (dismissed):
Thanks, @s-meza!

I simply added a custom `_equal` function that handles NaNs. Since they are the only objects unequal to themselves, I'm using `a != a` and `b != b` to check for NaN without importing math and try-catching `math.isnan`.

Let's merge! 👍🏻

evnchn:
Is `math.isnan` possibly better? Because two distinct objects both with overridden `__eq__` to be always false may slip through the cracks?

falkoschindler:
@evnchn You're probably right. It is a very odd edge case and I wanted to keep the `_equal` method short and concise. But using `math.isnan` might be more readable, more correct and equally cheap to evaluate.

s-meza:
Thank you !

PR #5642, opened 2026-01-14
Allow passing a path to `ui.run(show=...)`

falkoschindler:
### Motivation

For https://github.com/zauberzeug/rosys/issues/380 we would like to automatically open the browser at a specific path.

### Implementation

This PR relaxes the type for `show` in `ui.run` to accept a path.

### Progress

- [x] Pytests have been adjusted.
- [x] Documentation has been adjusted.

### DIFF
--- a/nicegui/ui_run.py
+++ b/nicegui/ui_run.py
@@ -100,7 +100,7 @@ def run(root: Optional[Callable] = None, *,
     :param cache_control_directives: cache control directives for internal static files (default: `'public, max-age=31536000, immutable, stale-while-revalidate=31536000'`)
     :param gzip_middleware_factory: GZipMiddleware factory function (e.g. ``lambda app: GZipMiddleware(app, minimum_size=500, compresslevel=9)``, defaults to Starlette's ``GZipMiddleware``, can be ``None`` to disable, *added in version 3.5.0*)
     :param fastapi_docs: enable FastAPI's automatic documentation with Swagger UI, ReDoc, and OpenAPI JSON (bool or dictionary as described `here <https://fastapi.tiangolo.com/tutorial/metadata/>`_, default: `False`, *updated in version 2.9.0*)
-    :param show: automatically open the UI in a browser tab (default: `True`) or a specific path (e.g. ``'/path/to/page'``, *since version 3.6.0*)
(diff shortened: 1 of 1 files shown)


nicegui/ui_run.py
```diff
@@ -100,7 +100,7 @@ def run(root: Optional[Callable] = None, *,
     :param cache_control_directives: cache control directives for internal static files (default: `'public, max-age=31536000, immutable, stale-while-revalidate=31536000'`)
     :param gzip_middleware_factory: GZipMiddleware factory function (e.g. ``lambda app: GZipMiddleware(app, minimum_size=500, compresslevel=9)``, defaults to Starlette's ``GZipMiddleware``, can be ``None`` to disable, *added in version 3.5.0*)
     :param fastapi_docs: enable FastAPI's automatic documentation with Swagger UI, ReDoc, and OpenAPI JSON (bool or dictionary as described `here <https://fastapi.tiangolo.com/tutorial/metadata/>`_, default: `False`, *updated in version 2.9.0*)
-    :param show: automatically open the UI in a browser tab (default: `True`)
+    :param show: automatically open the UI in a browser tab (default: `True`) or a specific path (e.g. ``'/path/to/page'``, *since version 3.6.0*)
```
evnchn:
```suggestion
    :param show: automatically open the UI in a browser tab (default: `True` opens `'/'`, pass a specific path (e.g. ``'/path/to/page'``) *since version 3.6.0*)
```

The original docstring forms X **or** Y outside of the brackets, as such the docstring kinda reads like "when a specific path is passed then 'automatically open the UI in a browser tab' does not take place anymore", until one realizes that can't make sense and the "English parser backtracks" 😵

evnchn (approved):
Just an optional docstring phrasing enhancement. The code is fine.

evnchn:
Also how about changing the type to just `str` or `str | Literal[False]` in the future? I don't expect readily usage of `show=True`. 

`str` is easy to implement but `show=''` is ugly...

nicegui/ui_run.py
falkoschindler: Ok, I improved the docstring. 👍🏻

falkoschindler:
> Also how about changing the type to just `str` or `str | Literal[False]` in the future?

I think we should keep the `True` option, especially for cases like script mode where you work with a single page and don't care about routes and paths.

PR #5643, opened 2026-01-14
Configurable defaults using `resolve_defaults`

evnchn:
### Motivation

The discussion in https://github.com/zauberzeug/nicegui/pull/5610 explored:

1. `ui.markdown.defaults(extras=...)` (and `ui.image.defaults(pil_convert_format=...)`?)
2. `ui.markdown.EXTRAS` or `ui.markdown.DEFAULT_EXTRAS`
3. `ui.markdown.default_extras` (and move towards `ui.image.pil_convert_format`?)

But these all come with API surface expansion / code duplication / use of kwargs (or a mix of the three)

1. `.defaults` if implemented at `Element` level is kwargs; if per individual element then the function signature of `.defaults` will duplicate `__init__` leading to maintenance burden
2. So many ClassVars
3. Also so many ClassVars, and they don't look like ClassVars

Can we not? I have an interesting idea. 

### Implementation

For a parameter responsive to this new functionality (marked by `DEFAULTS | ...`), you may mark a parameter which you would like to be a default for subsequent calls with `ui.default(...)`

Once passed into the `__init__` decorated by `@resolve_defaults`, it will unravel the inner value and pass on to the function as usual, but store the value into `_defaults` ClassVar dict. 

Next time when that parameter is unpopulated, it checks `_defaults` and grabs the value from there. 

### Demo code

```py
m = ui.markdown('''
    ```mermaid
    graph TD;
        Node_A --> Node_B;
    ```
''', extras=ui.default(['mermaid', 'fenced-code-blocks']))
ui.code(str(m.extras))
m2 = ui.markdown('''
    ```mermaid
    graph TD;
        Node_A --> Node_B;
    ```
''')  # applied automatically from defaults
ui.code(str(m2.extras))
```

### Progress

- [ ] Apply to all elements if see fit.
- [ ] Code currently simply does not work if try to use `ui.default` on non-marked / `DEFAULT_PROP`-marked / `DEFAULT_PROPS[...]`-marked elements. 
  - Either always warn, or better for the props one, we can write directly to the default props on behalf of the user's intention.

### DIFF
--- a/nicegui/defaults.py
+++ b/nicegui/defaults.py
@@ -20,15 +20,17 @@ R = TypeVar('R')
 @dataclass(**KWONLY_SLOTS)
 class Sentinel:
     key: str | None
+    source: str | None = None
 
     def __or__(self, other: T) -> T:
-        return SentinelValue(key=self.key, default=other)  # type: ignore[return-value]
+        return SentinelValue(key=self.key, default=other, source=self.source)  # type: ignore[return-value]
 
 
 @dataclass(**KWONLY_SLOTS)
 class SentinelValue:
     key: str | None
     default: Any
+    source: str | None = None
 
 
 class SentinelFactory:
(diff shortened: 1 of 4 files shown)


evnchn:
Wait but looking at the list in https://github.com/zauberzeug/nicegui/pull/5610#issuecomment-3743568187, do we only have 2 class-wide parameters where one would like to adjust? Namely only `PIL_CONVERT_FORMAT` and the new `DEFAULT_EXTRAS`? 

If that is the case then I think we should not over-engineer and go through all this decorator thing.

falkoschindler:
Oh, this looks fragile. 😕
Imagine you move around markdown elements, comment out on of them, or conditionally render some of them. If this is the recommended way to set defaults, it would break every time you change what is the first element.

Yes exactly, at the moment there are only two use cases, which is why I'd prefer simple class vars.

evnchn:
Wish I knew that there were only 2 class wide defaults earlier 🤡

PR #5128, opened 2025-09-11
Let `Client.delete()` do not cause KeyError if called multiple times

evnchn:
### Motivation

Fixes #5119. As discussed in https://github.com/zauberzeug/nicegui/issues/5119#issuecomment-3277041615, `Client.delete()` may get called twice, if within a prune, the client disconnects. 

### Implementation

This code follows https://github.com/zauberzeug/nicegui/issues/5090#issuecomment-3240147134 and https://github.com/zauberzeug/nicegui/issues/1826#issuecomment-3275126149 and modifies it so that should `Client.delete()` can be called as many times on the same client, resolving the problem. 

Could be considered a breaking change, but I don't see how a `Client.delete()` which _sometimes errors out_ is helpful, really. 

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] Pytests are not necessary.
- [x] Documentation is not necessary.

PR #5644, opened 2026-01-14
Delete slot children and event listeners to break up reference cycles

falkoschindler:
### Motivation

In #5110 we noticed that reference counting doesn't work in all cases and the garbage collector needs to cleanup leftovers.

After looking into this problem once again, I identified two scenarios:
1. Elements hold strong references to their slots (via `self.slots` dict and sometimes direct attributes like `splitter.before`). Each slot has a `children` list with strong references to child elements. When a parent is deleted, these references remain, preventing children from being garbage collected.
2. Event listeners can reference their element causing a reference cycle (e.g., `button.on('click', lambda: button.set_text('clicked'))`).

### Implementation

This PR implements `_handle_delete` to
- clear all slot children and
- clear all event listeners.

It also adds two new pytests covering these two cases.
The original example from #5110 also behaves as expected now.

### Progress

### DIFF
--- a/tests/test_element_delete.py
+++ b/tests/test_element_delete.py
@@ -1,3 +1,5 @@
+import weakref
+
 from nicegui import binding, ui
 from nicegui.testing import Screen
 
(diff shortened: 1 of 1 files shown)


falkoschindler:
@evnchn I was skeptical at first because I didn't think about it thoroughly. I feared that we might sweep something under the rug (remember PR #5128?) that bites us later. But after carefully evaluating the implementation with time and focus (and some AI), I came to the conclusion that the three new lines in `_handle_delete` fill a gap that was missed before. Without clearing the `children` and `event_listeners`, reference counting fails and the GC is needed. And as far as I can tell, this hasn't been done anywhere else, so there's no danger of doing things recursively.

PR #5645, opened 2026-01-15
Use a secondary AG Grid for un-mounted `get_client_data`

### LINKED ISSUE #3033: AG Grids inside hidden tabs don't get mounted

evnchn:
### Motivation

Reading un-mounted AG Grid's data is not necessarily an uncommon use case: https://github.com/zauberzeug/nicegui/issues/3033#issuecomment-3748668328

### Implementation

Spawn another AG Grid whose sole role is to host the `.api` object before the main AG Grid is mounted. 

Once the main AG Grid mounts itself, **on next tick** (very important, in the case that the main AG Grid is mounted as soon as it is there (i.e. normal situation)) destroy the secondary AG Grid's API.

If it isn't obvious, at any point in time, there is one and only one AG Grid API, even when the main AG Grid is inside `ui.tab` of non-active-selection. 

### Progress

- [x] Pytests https://github.com/zauberzeug/nicegui/pull/5645/commits/7b36bb16a0afe61977653fc772800b4aabfe2b3c
- [x] Documentation not necessary.


evnchn:
I don't like https://github.com/zauberzeug/nicegui/commit/65389ad6a605d22cd53959d6e311a62bf63dfdcb because it uses `$emit` which involves the server and is one round-trip slower...

falkoschindler:
Hm, ok...

- Why do you think the extra roundtrip is a problem? We're not sending much data and I don't see much room for race conditions. The user would have to switch tabs, interact with the grid and request user data before the server notices that the phantom is gone.
- I find your approach a bit fragile with the "secondary-id" prop being passed around. This is why I introduced the member `self._phantom_grid`.
- And your secondary UI element remains in memory until the primary `ui.aggrid` is deleted. To clean up immediately, the client _has_ to send the information to the server that the primary grid is mounted.

Maybe we should postpone the decision to 3.8?

evnchn:
Agreed on postponement. Consider also the whole thing's still kinda sketchy and so early 3.8 treatment is deserved. 

I actually have nothing against `$emit` specifically, and you're right in that it clears up the server-side. 

But, I would prefer that we keep the original logic of calling `.destroy` on the client-side phantom AG Grid ASAP without roundtrip server involvement. The storing of the phantom's ID is necessary as such. 

I think the solution lies somewhere between our approached.

evnchn:
Another "issue": the AG Grid data is transmitted twice under the current approach. 

Can the phantom "steal" the data from the main grid on the JS side to adhere to the DRY principle?

evnchn:
I don't think I can address "AG Grid data is transmitted twice" in a reasonable amount of time or effort. Let's leave it.

evnchn:
We should look at https://github.com/zauberzeug/nicegui/issues/5839#issuecomment-4018816101 for a systematic solution before looking at this AG Grid specific solution.

evnchn:
Well I am closing this because `ui.keep_alive`, once well done, is so much more elegant.

PR #5652, opened 2026-01-16
Write to props beforeUnmount for `LOOPBACK = False` elements

evnchn:
### Motivation

Magically fixes https://github.com/zauberzeug/nicegui/issues/2149. 

Stolen from predecessor #5612

> Let us recap the problem space:
> 
> https://github.com/zauberzeug/nicegui/blob/b11972f063b976174f0cd4a4af4a9ead18ac2a10/nicegui/elements/input.js#L19-L28
> 
> - When an `ui.input` is unmounted then remounted client-side (from being inside `ui.dialog`, for example), the JS component is supposed to mount itself **without involving the server**.
> - `this.value` does not change with user input (you can't even set it), only `this.inputValue` does. 
> - But the value in the latter is tossed on unmount, and re-initialized to `this.value` which is stale. 

**Question:** Is it really that we cannot set to `this.value`? I challenge the status quo!

### Implementation

I never did manage to write to `this.value` because Vue complains. 

But, I did find success writing to `mounted_app.elements[this.$props.id.slice(1)].props.value`

_And then everything magically worked and I am not quite sure why. @falkoschindler AI agents explanation time..._

Implementation details:

- For the input elements other than `input.py` which they uses `tag='q-input'` before, I wired them to use `editor.js` by expanding it to render both `q-editor` and `q-input`.
- It is legal to add `id: String` to props, as it is always set. 

### Progress

### DIFF
--- a/tests/test_dialog.py
+++ b/tests/test_dialog.py
@@ -129,6 +129,7 @@ def test_reopening_dialog_with_various_inputs(element_factory: Callable, selecto
     screen.click('Edit')
     screen.should_contain('Dialog open')
 
-    element = screen.selenium.find_element(By.XPATH, selector)
-    actual = element.get_attribute('value') or element.text
-    assert actual == text, f'Value of {selector} did not persist: expected {text!r}, got {actual!r}'
+    @screen.wait_for
+    def element_has_expected_content():
+        element = screen.selenium.find_element(By.XPATH, selector)
+        return element.get_attribute('value') or element.text == text


falkoschindler:
@evnchn Great idea to tackle the props dictionary! 🚀
And I get the intention of
```py
mounted_app.elements[this.$props.id.slice(1)].props.value = this.inputValue;
```
But I don't understand why you're wiring input elements to editor.js. Can you elaborate?

Apart from that, a test is red. 😕

evnchn:
> wiring input elements to editor.js 

We need it because otherwise those elements, which simply do `super().__init__(tag='q-input', label=label, value=value, on_value_change=on_change)`, we have no place to put in this new `beforeUnmount` logic. 

I will look at the test soon.

evnchn:
https://github.com/zauberzeug/nicegui/compare/eb5d8ab1f237080415762e17cd4e9e217a8ac7eb..9466c31954a720be376244b6fb9c0869b0a2c2e9: I got a more elegant fix without changing the tests (a sign of a breaking API change if left to be)

nicegui/elements/basic_input.js
evnchn, line 1: This file is just `editor.js` without the `template`. 

Do we rework `editor.py` to use this?

evnchn: It doesn't work. Text input becomes `undefined` for some reason

falkoschindler: Interesting... A common core-implementation for the input value handling would certainly be nice, but certainly more work than the single line of magic so far. 🪄

evnchn: I think we should leave it for another PR 😪

evnchn:
Not sure if https://github.com/zauberzeug/nicegui/pull/5652/commits/2074bec6bf4451f6e830b4f4f59ddd339ce12fe1 is the right thing to do: I do not see thi pattern elsewhere. 

If you have a more elegant fix, go ahead!

evnchn:
> :id="id" is probably not needed in editor.js

Without it, for some reason the element is mounted without HTML ID. I can't figure out how to fix it without this rather hacky hack...

The rest I agree with. I am quite shocked the 4 already work (maybe my testing is messed up 🙇). Anyways thanks!

falkoschindler:
> Without it, for some reason the element is mounted without HTML ID. I can't figure out how to fix it without this rather hacky hack...

Maybe you need the line
```py
self._props['for'] = self.html_id
```
like in input.py?

Oh and I just noticed that we also should add `ui.codemirror` to the test to address #5661.

evnchn:
`basic_input.js` gone. It really works on main already 🤡 

`self._props['for'] = self.html_id`

Then it becomes `<div class="q-editor q-editor--default nicegui-editor" loopback="false" for="c4">` 🤔 

If you have a fix then go ahead.

evnchn:
@falkoschindler Oh no! Something is wrong...

Let's read Quasar docs again:

> `for`: String
> 
> Used to specify the 'id' of the control and also the 'for' attribute of the label that wraps it; If no 'name' prop is specified, then it is used for this attribute as well

It means the following: 

- For all elements which was "fixed" by #5420, it is slightly inaccurate as the element with the ID is the control, not the entire Quasar element.
- `for` prop will not work for `ui.editor` because there is no such prop in Quasar.
- For elements with `id: String` in `props`, using `:id="id"` is the only correct way (that I know) for adding an ID, because `id` is not in `this.$attrs` which is "undeclared attributes" 
- The elements in #5420 need a fixing-up, but may subtly break existing user code which expects the DOM a certain way. 

What do you think about this situation? Thanks.

falkoschindler:
@evnchn Yes, PR #5420 intentionally used the "for" prop to add an ID to some nested element. The main goal was that every NiceGUI element has a corresponding HTML element with a matching `html_id`. Since QInput already added an ID to the nested \<input\> element, I followed this pattern adding IDs to nested elements.

For `ui.editor` we can use `:id="id"` if the "for" prop doesn't work.

If we want to ensure that the outermost element get's the ID, we'd need to add js files for various elements, find a solution for `ui.input` which already had a nested ID, and wait until 4.0 with such a breaking change.

evnchn:
I don't think this is in-progress anymore. Am I missing something?

Issue #5583, opened 2025-12-15
Slow SVG updates for interactive images with PIL source


denniswittich:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
import time
from io import BytesIO

import numpy as np
import PIL.Image
from fastapi import Response
from nicegui import app, ui


def startup():
    @ui.page('/')
    def index_page():
        ui.label('Choose a test page:')
        ui.link('Fast Images', '/fast')
        ui.link('Slow Images', '/slow')

    @ui.page('/fast')
    def fast_images_page():
        ui.label('Showing image from resource endpoint')
        timestamp = time.time()
        ii = ui.interactive_image(source=f'/image/{timestamp}').classes('w-1/2')
        ui.button('Click me!', on_click=lambda: ii.set_content(_random_svg()))

    @ui.page('/slow')
    def slow_images_page():
        ui.label('Showing image from generated PIL image')
        pil_image = _gen_pil_image()
        ii = ui.interactive_image(source=pil_image).classes('w-1/2')
        ui.button('Click me!', on_click=lambda: ii.set_content(_random_svg()))

    app.add_api_route('/image/{timestamp}', _get_image)


def _gen_pil_image() -> PIL.Image.Image:
    random_pixels = np.random.randint(0, 255, (2000, 2000, 3), dtype=np.uint8)
    pil_image = PIL.Image.fromarray(random_pixels)
    return pil_image


def _get_image(timestamp: str) -> Response:  # pylint: disable=unused-argument
    pil_image = _gen_pil_image()
    buffer = BytesIO()
    pil_image.save(buffer, format='JPEG', quality=85)
    jpeg_bytes = buffer.getvalue()
    return Response(content=jpeg_bytes, media_type='image/jpeg')


def _random_svg() -> str:
    colors = ['red', 'green', 'blue', 'yellow', 'purple', 'orange']
    svg_parts = ['<svg xmlns="http://www.w3.org/2000/svg" width="2000" height="2000">']
    for _ in range(10):
        x = np.random.randint(0, 2000)
        y = np.random.randint(0, 2000)
        radius = np.random.randint(10, 50)
        color = np.random.choice(colors)
[...]

PR #5653, opened 2026-01-16
Use path instead of base64 for PIL handling

evnchn:
### Motivation

#5583 reveals that using base64 encoding for PIL images is a bad idea when the PIL image is large. 

This is due especially to how we send the entire props dict over

### Implementation

- Use `pil_to_tempfile` instead.
- SourceElement expose a new private property `self._source_for_cleanup`
  - Responsible for cleaning up the temporary file, when:
    - Either a new source arrives
    - Or the element is deleted

### Progress

- [x] Documentation is not necessary for an underlying performance uplift.

### DIFF
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -130,5 +130,4 @@ def test_pil_image_cleanup(screen: Screen):
     screen.should_load_image(image)
 
     screen.click('Delete')
-    screen.wait(0.2)
-    assert not Path(temp_path_str).exists()
+    screen.wait_for(lambda: not Path(temp_path_str).exists())


falkoschindler:
@evnchn I just updated this branch and started with a quick AI review. There seem to be some problems with the current implementation:

#### 1. **Critical: `weakref.finalize` won't work correctly**

In `source_element.py:26`:
```python
weakref.finalize(self, self._cleanup_source)
```

This is problematic because `self._cleanup_source` is a **bound method** that holds a reference to `self`. This creates a reference cycle that prevents `self` from being garbage collected, meaning the finalizer will never trigger.

The correct pattern is to pass the data needing cleanup as arguments to a static/module-level function:

```python
def _cleanup_temp_file(path: Path | None) -> None:
    if path is not None:
        path.unlink(missing_ok=True)

# In __init__:
# Need to use a mutable container since _source_for_cleanup changes
self._cleanup_holder: list[Path | None] = [None]
weakref.finalize(self, lambda holder=self._cleanup_holder: _cleanup_temp_file(holder[0]))
```

Or use `atexit` with a different cleanup strategy.

#### 2. **Temp file leak when source is set via property binding**

When the source is changed via `element.source = new_value` or through bindings, `_handle_source_change()` is called but `_cleanup_source()` is not. The cleanup only happens through `set_source()`.

In `source_element.py:106-111`:
```python
def _handle_source_change(self, source: Any) -> None:
    self._set_props(source)  # No cleanup of old temp file!
```

Should be:
```python
def _handle_source_change(self, source: Any) -> None:
    self._cleanup_source()  # Clean up old temp file first
    self._set_props(source)
```

And then `set_source()` may not need its own cleanup call since setting `self.source` will trigger `_handle_source_change`.

#### 3. **Missing `_handle_delete` override in Image/InteractiveImage**

When an element is explicitly deleted (not garbage collected), `_handle_delete()` is called. The base `SourceElement._handle_delete()` doesn't clean up `_source_for_cleanup`. While the finalizer is supposed to handle this, given the finalizer issue above, temp files may leak when elements are deleted.

Consider adding cleanup in `_handle_delete`:
```python
def _handle_delete(self) -> None:
    self._cleanup_source()
    return super()._handle_delete()
```

#### 4. **Minor: `force_reload()` warning message**

In `image.py:43-44`, the check for `data:` prefix is still present. While this won't cause issues (PIL images won't have this prefix anymore), the warning message mentioning "base64" might be slightly confusing now.

### Test Coverage

The checklist indicates tests haven't been added. Given the cleanup complexity, tests would be valuable:
- Test that temp files are created when using PIL images
- Test that temp files are cleaned up when source changes
- Test that temp files are cleaned up when element is deleted

### Recommendation

The core idea is sound and addresses a real performance problem, but the cleanup mechanism has bugs that could cause temp file accumulation. I'd suggest:

1. Fix the `weakref.finalize` usage to properly clean up on GC
2. Add cleanup to `_handle_source_change` to prevent leaks via property/binding changes  
3. Consider adding cleanup to `_handle_delete` as a belt-and-suspenders approach
4. Add tests for the cleanup behavior

evnchn:
Got it. Should be good now.

evnchn:
Great. I was a bit concerned at the `class _TempPath(type(Path())):`, but actually:

- Path() returns a platform-specific instance - WindowsPath on Windows, PosixPath on Unix
- type(Path()) gets the concrete class

Chaotic but works. Tests pass. What's not to love!

.gitignore
```diff
@@ -1,4 +1,7 @@
 __pycache__/
+.mypy_cache/
+.pytest_cache/
+.ruff_cache/
```
evnchn, line 4: These folders are not scanned by Git even in main though. Do we need it?

nicegui/elements/image.py
```diff
@@ -34,26 +33,32 @@ def set_source(self, source: str | Path | PIL_Image) -> None:
...
         if self._props['src'].startswith('data:'):
-            log.warning('ui.image: force_reload() only works with network sources (not base64)')
+            log.warning('ui.image: force_reload() only works with network sources (not data URIs)')
             return
         self._props['t'] = time.time()
 
 
-def pil_to_base64(pil_image: PIL_Image, image_format: str) -> str:
-    """Convert a PIL image to a base64 string which can be used as image source.
+def pil_to_tempfile(pil_image: PIL_Image, image_format: str) -> Path:
```
evnchn:
```suggestion
def pil_to_tempfile(pil_image: PIL_Image, image_format: str) -> _TempPath:
```

.gitignore
falkoschindler: I guess .mypy\_cache contained a .gitignore to ignore its content. But after removing it locally and running mypy again, Git wanted to add 4000 files to the repo. So I thought it won't hurt to exclude them explicitly.

PR #5654, opened 2026-01-17
Do not `Vue.resolveComponent` if `app.config.isNativeTag`

### LINKED ISSUE #3179: Component registration warning always emitted
### Description

Whenever I load or reload the "app's" browser page, Vue produces the warning: 
```
vue.global.js:1503 [Vue warn]: Failed to resolve component: div
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement. 
  at <QPage id="c2" ref="r2" key=2  ... > 
  at <QPageContainer id="c1" ref="r1" key=1  ... > 
  at <QLayout id="c0" ref="r0" key=0  ... > 
  at <App>
```

### Reproduce
1. put the following code in a file (`test.py`)
   ```py
   from nicegui import ui
   ui.label("Test")
   ui.run(prod_js=False)
   ```
2. Run the script: `python test.py`
3. Open the browser console and observe the warning.

With the minified / "production" JS assets (`prod_js=True`) the warning isn't being produced. I assume that's simply because the logging error level is set higher than warn for "production". This issue makes it annoying to develop using non-minified JS-sources. It also indicates that there is a real issue with how nicegui is leveraging Vue for custom HTML injection, so worth investigating if you have the bandwidth.

### Version information
```
Python 3.10.13
```

Package versions (clean virtual environment with only nicegui installed)
```
nicegui==1.4.26

aiofiles==23.2.1
aiohttp==3.9.5
aiosignal==1.3.1
annotated-types==0.7.0
anyio==4.4.0
async-timeout==4.0.3
attrs==23.2.0
bidict==0.23.1
certifi==2024.6.2
charset-normalizer==3.3.2
click==8.1.7
colorama==0.4.6
docutils==0.19
exceptiongroup==1.2.1
fastapi==0.109.2
frozenlist==1.4.1
h11==0.14.0
httpcore==1.0.5
httptools==0.6.1
httpx==0.27.0
idna==3.7
ifaddr==0.2.0
itsdangerous==2.2.0
Jinja2==3.1.4
markdown2==2.4.13
MarkupSafe==2.1.5
multidict==6.0.5
orjson==3.10.3
pscript==0.7.7
pydantic==2.7.3
pydantic_core==2.18.4
Pygments==2.18.0
python-dotenv==1.0.1
python-engineio==4.9.1
python-multipart==0.0.9
python-socketio==5.11.2
PyYAML==6.0.1
requests==2.32.3
simple-websocket==1.0.0
sniffio==1.3.1
starlette==0.36.3
typing_extensions==4.12.1
urllib3==2.2.1
uvicorn==0.30.1
vbuild==0.8.2
watchfiles==0.22.0
websockets==12.0
wsproto==1.2.0
yarl==1.9.4
```

evnchn:
### Motivation

Finally fixes #3179, where `[Vue warn]: Failed to resolve component: div` always shows up when `prod_js=False`

### Implementation

Why this does not happen with a plain Vue recreation is that we call `Vue.resolveComponent` directly. 

Consider `app.config.compilerOptions.isCustomElement = (tag) => {console.log(tag); return ['div'].includes(tag)}` doesn't even log the tag. 

Rather, just [do not `Vue.resolveComponent` if `app.config.isNativeTag`](https://github.com/zauberzeug/nicegui/commit/0ad94e5bf4822f9c3a17550d89a2610c4b358aac) and you should be good to go!

[Tighten `test_dev_mode`](https://github.com/zauberzeug/nicegui/commit/4c15e58cbc1520f9474f2b91d943adeb509fa3e5) to look for Vue warn and fail the test. 

### Progress

- [x] Pytests 4c15e58cbc1520f9474f2b91d943adeb509fa3e5

### Final notes

**A simple-win**

### DIFF
--- a/tests/test_prod_js.py
+++ b/tests/test_prod_js.py
@@ -29,3 +29,4 @@ def test_prod_mode(screen: Screen):
     importmap = screen.selenium.find_element(By.XPATH, '//script[@type="importmap"]')
     assert f'/_nicegui/{__version__}/static/vue.esm-browser.prod.js' in (importmap.get_attribute('innerHTML') or '')
     screen.selenium.find_element(By.XPATH, f'//script[@src="/_nicegui/{__version__}/static/quasar.umd.prod.js"]')
+    assert 'Vue warn' not in screen.render_js_logs()


falkoschindler (commented):
Brilliant! 🤩

Just one possible edge case: What if someone defined a Vue component with name "input" (which exists as HTML element). I could imagine that this PR changes behavior by rendering the HTML element, not the Vue component.

evnchn:
> defined a Vue component with name "input"

I don't think we should let those people get away with it. That's quite a weird use case...

falkoschindler:
Bad example. But think of any lesser known tag (see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements). A NiceGUI developer that didn't know all available HTML tag names might have chosen one of them by accident without negative consequences - until now.

Maybe I should try it out to see if this PR _really_ breaks behavior in this case. If it does, we might need to send a list of registered components to the client...

evnchn:
But NiceGUI tags are always prefixed. 

https://github.com/zauberzeug/nicegui/blob/e81503c36414cdbe4de58f8cbb8bfa511ad04dc8/nicegui/dependencies.py#L33-L36

So this would only break custom JS code.

evnchn:
Test script:

```py
TRY_NAME = 'address'

app.config.vue_config_script += f'''
app.component('{TRY_NAME}', {{
    template: `<div>Hello, World!</div>`
}});
'''

@ui.page('/')
def main_page():
    ui.element(TRY_NAME)

ui.run(prod_js=False)
```

Before:

image


After: 

image

So yes, this PR does break existing behaviour, but a behaviour which is not condone by NiceGUI nor Vue in the first place. I believe we should be safe to proceed.

evnchn:
NOTE: In https://github.com/zauberzeug/nicegui/pull/5654/commits/c91bb04ee4ebc8ac6acbdbd6a0d8b699d7d40cbf I used `app._context.components` which is private API. Be sure to check if it still works when bumping Vue version. 

Also, it only stops working in `prod_js=False`. In `pros_js=True` tag-named components still work since `app.config.isNativeTag` becomes always-false in that case, and so we always `Vue.resolveComponent`.

evnchn:
But I kinda feel like we should just not have this extra compatibility. The impact and thus the cost is low - just `prod_js=False` if your website breaks, fix your code, and turn it back on.

falkoschindler:
I was just about to write that I agree with `app.config.isNativeTag(element.tag) ? element.tag : Vue.resolveComponent(element.tag)` because not only builtin components use a "nicegui-" prefix, but also the ones defined in user code. So it is very unlikely that someone managed to use a custom "article" component or something like this.

> But I kinda feel like we should just not have this extra compatibility.

I think so too.

> In `pros_js=True` tag-named components still work since `app.config.isNativeTag` becomes always-false in that case, and so we always `Vue.resolveComponent`.

That's interesting. So `pros_js` changes how components are resolved. I wonder why. But it's probably fine.

evnchn:
Then let's drop it 🗑️

evnchn:
> probably fine

It is fine, because if the component isn't found then the raw tag name lands in the DOM. 

Having a functional `app.config.isNativeTag` is just for non-production warnings only. 

On this I totally agree with Vue's design to short-circuit it in production.

falkoschindler (approved):

PR #5656, opened 2026-01-17
Click-to-reload when `ui.scene` WebGL context lost

### LINKED ISSUE #5360: Some `ui.scene` demos are killed when scrolling to the bottom of the page
### First Check

### Description

1. Navigate to https://nicegui.io/documentation/scene.
2. Scroll to the bottom.
3. Scroll back to the top.

The first demo stopped rendering, displays a sad face instead and JavaScript says "WARNING: Too many active WebGL contexts. Oldest context will be lost."

Image

Note: The reproduction might depend on PR #5344 and/or PR #5349 being deployed because they add additional demos which exceed the limit.

### NiceGUI Version

3.1.0

### Python Version

3.11

### Browser

Chrome

### Operating System

macOS

evnchn:
### Motivation

Fixes #5360, where some `ui.scene` suffers irrevocably from lost WebGL context and there is no way to get it work again without a page reload. 

### Implementation

When detect the context is gone (`this.renderer.getContext().drawingBufferHeight === 0 && this.renderer.getContext().drawingBufferWidth === 0`), force Vue to re-render by temporarily setting `tag` to be `''` then restoring it on the next tick

### Test script

```py
from nicegui import ui

with ui.grid(columns=5):
    for _ in range(25):
        ui.scene(width=100, height=100)

ui.run(prod_js=True, dark=True) # to ensure the overlay shows up in dark mode
```

image

### Progress

- [x] I am not sure if it is worthwhile to pytest this?
- [x] Documentation is not necessary for a quality-of-life improvement.

### DIFF
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -71,10 +71,7 @@ export default {
       <canvas style="position:relative"></canvas>
       <div style="position:absolute;pointer-events:none;top:0"></div>
       <div style="position:absolute;pointer-events:none;top:0"></div>
-      <div style="position:absolute;display:none;inset:0;padding:8px;color:black;background-color:#ffffffcc;cursor:pointer">
-        WebGL context lost.
-        Click to re-initialize.
-      </div>
+      <div style="position:absolute;display:none;inset:0;cursor:pointer">WebGL context lost. Click to re-initialize.</div>
     </div>`,
 
   mounted() {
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Nice! I just made a few minor changes:

- use `inset:0` instead of specifying left, top, width and height,
- toggle visibility of children 0 to 2 so that we don't need to cover them with the right color (better appearance in dark mode),
- use `once: true` to indicate single-use event handler,
- use the "webglcontextlost" event instead of measuring the drawing buffer.

Let's merge! 🏁

falkoschindler:
> I am not sure if it is worthwhile to pytest this?

We could test if the info message appears. But clicking it and checking if this individual scene is rendered again sounds complicated. Let's leave it untested as a poorly supported edge case.

Issue #5663, opened 2026-01-19
Misleading warning when navigating on `ui.sub_pages`


falkoschindler:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

def root():
    ui.sub_pages({'/': main})

def main():
    ui.link('Go to other page (and come back via back button)', '/other')
    ui.button('Button', on_click=lambda: None)

ui.run(root)
```

### Description

I noticed a problem with PR #5439 on main:

1. Click on the link (the target doesn't exist, but it doesn't matter),
2. use the browser's "Back" button to navigate back to the index page,
3. a warning appears: "Event listeners changed after initial definition. Re-rendering affected elements."

So our event-listener-change detection seems to be broken.

### NiceGUI Version

main branch

### Python Version

3.9.25

### Browser

Chrome

### Operating System

macOS

evnchn:
Ah yes

And I don't think you need to use `ui.sub_pages` even. A refreshable with a button with an on click may have false warning on refresh as well (you can try it out, I'm doing errands.

falkoschindler:
Correct!

```py
@ui.refreshable
def my_ui():
    ui.button('Button', on_click=lambda: None)

my_ui()
ui.button('Refresh', on_click=my_ui.refresh)
```

PR #5664, opened 2026-01-19
Fix errorneous late event registration warning

evnchn:
### Motivation

In #5663, newly added elements also trigger late event registration warning

### Implementation

Skip the check if the new element's ID does not exist in `this.elements`, or else `this.elements[id]?.events || []` becomes `[]` and **all new elements with an event gets trigger the warning.** 

### Progress

- [x] Do we need pytest for anti-regression?

### Final notes

**A simple-win**


falkoschindler:
> Do we need pytest for anti-regression?

Hm, I don't think so. Our tests would get pretty crazy if we'd add such "second order" regression tests.
In this particular case you'd notice the bug as soon as you open a documentation page.

PR #5665, opened 2026-01-19
Add `on_path_changed` demo and reference for `SubPagesRouter`

### LINKED ISSUE #5487: ui.navigate.to not triggering sub_pages_navigate event

falkoschindler:
### Motivation & Implementation

Fixes #5487 by adding documentation for how to register a path-changed event handler with `ui.sub_pages`.

### Progress

### DIFF
--- a/website/documentation/content/sub_pages_documentation.py
+++ b/website/documentation/content/sub_pages_documentation.py
@@ -2,6 +2,7 @@ import asyncio
 from typing import Any, Awaitable, Callable, Optional
 
 from nicegui import PageArguments, background_tasks, ui
+from nicegui.sub_pages_router import SubPagesRouter
 
 from . import doc
 
@@ -13,6 +14,11 @@ class FakeSubPages(ui.column):
         self.routes = routes
         self.data = data
         self.task: Optional[asyncio.Task] = None
+        self.path_changed_handlers: list[Callable[[str], None]] = []
+        self.route: Optional[str] = None
+
+    def on_path_changed(self, handler: Callable[[str], None]) -> None:
+        self.path_changed_handlers.append(handler)
 
     def init(self) -> None:
         self._render('/')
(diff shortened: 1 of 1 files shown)


evnchn:
Fundamnetally there is nothing wrong with this PR:

- Demo code works
- Demo in website works
- Text is concise
- Shows the full reference via `doc.reference`

But: Why are async handlers not supported? Seems kinda weird when it is everywhere else...

falkoschindler:
> Why are async handlers not supported?

Good question. But let's leave it for another PR if there is demand.

PR #5673, opened 2026-01-21
Implicit handshake + Bugfix

evnchn:
### Motivation

#5499 was good for performance but when the `Client` is gone (common case when server reloads) then the browser does not reload. 

### Implementation

Check https://github.com/zauberzeug/nicegui/commit/785650ecb150c68fa8a179fece559b947b5ae735 for the fix code

Notably:

- It is not a good idea to infinitely-retry on failed socketio connect, since that could lead to resource exhaustion. 
- Sending a custom error message to the browser is a must, as a result. 
- [Python-socketio's docs demo code](https://python-socketio.readthedocs.io/en/v4/server.html#defining-event-handlers) is not very accurate. Though the code mentions `raise ConnectionRefusedError('authentication failed')`, the error that works to send a custom message is `socketio.exceptions.ConnectionRefusedError`
- Native `ConnectionRefusedError` does not work! (https://github.com/zauberzeug/nicegui/pull/5499/commits/305f7e5c0a985dce8c6d9e74561b667feb8910ba fails test: https://github.com/zauberzeug/nicegui/actions/runs/21179651294/job/60918132159)

### Progress

- [x] Should we pytest?

### DIFF
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -188,11 +188,10 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
 
 
 @sio.on('connect')
-async def _on_connect(sid: str, data: dict[str, Any], _=None) -> bool:
+async def _on_connect(sid: str, data: dict[str, Any], _=None) -> None:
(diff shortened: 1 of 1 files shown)


evnchn:
I don't think pytest is possible within our current tooling.

PR #5685, opened 2026-01-26
Deprecate custom AG Grid checkboxRenderer in favor of built-in agCheckboxCellRenderer

falkoschindler:
### Motivation

Inspired by https://github.com/zauberzeug/nicegui/discussions/5681#discussioncomment-15607246.

The custom `checkboxRenderer` cell renderer for AG Grid was added in 2020 based on [this blog post](https://blog.ag-grid.com/binding-boolean-values-to-checkboxes-in-ag-grid/) when AG Grid didn't have built-in support for boolean checkboxes.

AG Grid now provides the following options.

1. Separate renderer and editor
    ```python
    {
        'field': 'is_active',
        'cellRenderer': 'agCheckboxCellRenderer',
        'cellEditor': 'agCheckboxCellEditor',
        'editable': True,
    }
    ```

2. Always-interactive renderer (no editing flow)
    ```python
    {
        'field': 'is_active',
        'cellRenderer': 'agCheckboxCellRenderer',
        'cellRendererParams': {'disabled': False},
    }
    ```

3. Use cellDataType (recommended)
    ```python
    {
        'field': 'is_active',
        'cellDataType': 'boolean',
        'editable': True,
    }
    ```
    This automatically sets up `agCheckboxCellRenderer` + `agCheckboxCellEditor` + keyboard support.

The built-in solutions are better integrated with AG Grid's theming, have proper keyboard support (spacebar to toggle), and are maintained by the AG Grid team.

This PR deprecates the custom `checkboxRenderer` in preparation for removal in NiceGUI 4.0.

### Implementation

Added a static method `_check_deprecated_checkbox_renderer()` that scans column definitions for usage of `'cellRenderer': 'checkboxRenderer'` and emits a `warn_once` message with migration guidance.

The warning provides two migration paths:
1. **Recommended**: Use `"cellDataType": "boolean"` with `"editable": True` for interactive checkboxes (uses AG Grid's editing flow)
2. **Legacy behavior**: Use `"cellRenderer": "agCheckboxCellRenderer"` with `"cellRendererParams": {"disabled": False}` to keep the always-interactive behavior

Note: This is a breaking change because `cellDataType: 'boolean'` makes checkboxes read-only by default (requires `editable: True`), whereas the current `checkboxRenderer` is always interactive. Users need to consciously migrate.

### Update

Changed from warning-only to **auto-migration**: The custom `checkboxRenderer` is now automatically replaced with `agCheckboxCellRenderer` + `cellRendererParams: {'disabled': False}` at runtime, preserving identical behavior. This allows removing the custom JS code immediately rather than keeping it until 4.0.

Users still see a deprecation warning suggesting they update their code to use `cellDataType: 'boolean'` with `editable: True` for the modern approach with keyboard support and proper editing flow.

### Progress

### DIFF
--- a/nicegui/elements/aggrid/aggrid.py
+++ b/nicegui/elements/aggrid/aggrid.py
@@ -73,12 +73,13 @@ class AgGrid(Element, component='aggrid.js', esm={'nicegui-aggrid': 'dist'}, def
         if migrated:
             helpers.warn_once(
                 "AG Grid: 'checkboxRenderer' is deprecated.\n"
-                'Your code currently contain:\n'
+                'Your code currently contains:\n'
                 "    'cellRenderer': 'checkboxRenderer',\n"
                 'But the native renderer is preferred for accessibility and styling:\n'
                 "    'cellDataType': 'boolean',\n"
                 "    'editable': True,\n"
-                'Please migrate ASAP as the backwards-compatibility will be removed in NiceGUI 4.0.')
+                'Please migrate ASAP as the backwards-compatibility will be removed in NiceGUI 4.0.'
+            )
 
     @classmethod
     def from_pandas(cls,


evnchn:
Some questions:

- `'cellRenderer': 'agCheckboxCellEditor'` errors out in the console with `Uncaught TypeError: e.validate is not a function`. Could you look into that before recommending it?
- Why is this PR "technically a breaking change candidate" when it **IS** a breaking change by-definition? We are taking away `checkboxRenderer`. If that isn't a breaking change I don't know what is. 
- While users must migrate by 4.0, what is needed for the "conscious" migration? It seems to me that since `checkboxRenderer` is always editable (no provision in code for making it disabled, even), that a direct search-and-replace of `'cellRenderer': 'checkboxRenderer'` with `'cellRenderer': 'agCheckboxCellRenderer', 'editable': False` would suffice
- Add-on to the previous question, could we drop `checkboxRenderer` right now and perform the option adjustment for the user until NiceGUI 4.0 to net us less technical debt right now?

falkoschindler:
> * `'cellRenderer': 'agCheckboxCellEditor'` errors out in the console with `Uncaught TypeError: e.validate is not a function`. Could you look into that before recommending it?

I updated the description to clarify the usage of "agCheckboxCellRenderer" and "agCheckboxCellEditor".

> * Why is this PR "technically a breaking change candidate" when it **IS** a breaking change by-definition? We are taking away `checkboxRenderer`. If that isn't a breaking change I don't know what is.

Ok, I tightened the formulation.

> * While users must migrate by 4.0, what is needed for the "conscious" migration? It seems to me that since `checkboxRenderer` is always editable (no provision in code for making it disabled, even), that a direct search-and-replace of `'cellRenderer': 'checkboxRenderer'` with `'cellRenderer': 'agCheckboxCellRenderer', 'editable': False` would suffice
>
> * Add-on to the previous question, could we drop `checkboxRenderer` right now and perform the option adjustment for the user until NiceGUI 4.0 to net us less technical debt right now?

I think the replacement would be
```py
'cellRenderer': 'agCheckboxCellRenderer',
'cellRendererParams': {'disabled': False},
```
I'll try it out. 👍🏻

evnchn:
Some observations: 

**Can we deduplicate option 1 and 2 above?** 

Considering:

- `'cellEditor': 'agCheckboxCellEditor',` is the default when you use `'cellRenderer': 'agCheckboxCellRenderer',`
- No functional difference between setting `'editable': True,` over setting `'cellRendererParams': {'disabled': False},`

I may be missing something, though, so please fact-check!

**Why can't we replace `'cellRenderer': 'checkboxRenderer'` with `'cellDataType': 'boolean', 'editable': True`** on day 1?

I finally get the keyboard support thing. Option 1 and 2 seems like half-solutions to me. 

**The message is long and not quite accurate**

- Do we really need `stack_info=True`? Used sparingly and I don't think we need this. 
- The phrasing sounds like the user can `Consider` to do nothing but really they got to do the migration and they can `Consider` to migrate to something _better_. It is a bit confusing. Let's make it more strong to ensure nobody misses the memo. Add-on to point 2, I would promote to migrate right-away to the best solution.

falkoschindler:
Yes, I like the two improvements! Auto-upgrading the users to the new, improved UX should be fine.
Ready to merge if you approve, @evnchn 🙂

evnchn (approved):
I like this. 

Bonus item for later: add a demo?

PR #5686, opened 2026-01-26
Kill Uvicorn server via `gc.get_objects()` in `ui.run_with`

### LINKED ISSUE #3253: `app.shutdown` does not work with `ui.run_with`
### Description

```python
import uvicorn
from fastapi import FastAPI
from nicegui import ui, app

server = FastAPI()


def init() -> None:
    @ui.page("/")
    def main():
        ui.button('shutdown', on_click=app.shutdown)

    ui.run_with(server)


init()

if __name__ == "__main__":
    uvicorn.run(server, host="127.0.0.1", port=7778)
```

My app is started using fastapi. I want to stop the service through a button. How to do it?
app.shutdown will report an exception.

a

evnchn:
### Motivation

This PR fixes #3253, how `app.shutdown` does not work when we do `ui.run_with` as `Server.instance` was never set. 

### Analysis

Note that upstream is extremely unlikely to be able to supply it to us:

- FastAPI CLI simply calls `uvicorn.run` https://github.com/fastapi/fastapi-cli/blob/e62b61d6c56b77d39310a1c49e8eb7abdad2785a/src/fastapi_cli/cli.py#L218-L233
- `uvicorn.run` doesn't even return `server` https://github.com/Kludex/uvicorn/blob/918dae6ef91917a5824ab48c46026e07b7c80beb/uvicorn/main.py#L579-L602

### Implementation (Copilot)

**Improvements to shutdown handling:**

* Updated the `shutdown` method in `nicegui/app/app.py` to use garbage collection to find and set the `should_exit` flag on the correct `uvicorn.Server` instance if the standard reference is unavailable, increasing reliability of server shutdown.
* Added the `gc` and `uvicorn` imports to support the improved shutdown logic in `nicegui/app/app.py`. [[1]](diffhunk://#diff-adae3337464d5c5adf411c1b2e4b52034747030db2de3f87cfc83fea39763f96R2) [[2]](diffhunk://#diff-adae3337464d5c5adf411c1b2e4b52034747030db2de3f87cfc83fea39763f96R13)
* **Human note: Server instance is always unavailable when using `ui.run_with`, and always available when using `ui.run`**

**Application instance management:**

* Introduced a new global variable `fastapi_app` in `nicegui/core.py` to hold a reference to the FastAPI application, and ensured it is set during application startup in `nicegui/ui_run_with.py`. [[1]](diffhunk://#diff-5ed898650c8cdc47c39575a2e752cfe07edba4c755c1c2390460373640983081R6) [[2]](diffhunk://#diff-5ed898650c8cdc47c39575a2e752cfe07edba4c755c1c2390460373640983081R15) [[3]](diffhunk://#diff-8af9e19f881b73ad9d426d4e341acfbc54b5f68e57e68d8eb2e7866fdde76f24R95-R96)
* **Human note: Prevents killing the wrong server, in case there are multiple!**

### Progress

- [x] Pytests is out-of-scope.
- [x] Documentation is not necessary for a capability expansion of an already-there function.

### DIFF
--- a/nicegui/ui_run_with.py
+++ b/nicegui/ui_run_with.py
@@ -90,11 +90,11 @@ def run_with(
     @asynccontextmanager
     async def lifespan_wrapper(app):
         def _get_server_instance() -> Optional[uvicorn.Server]:
-            for srv in (o for o in gc.get_objects() if isinstance(o, uvicorn.Server)):
-                wrapped = srv.config.loaded_app
+            for server in (obj for obj in gc.get_objects() if isinstance(obj, uvicorn.Server)):
+                wrapped = server.config.loaded_app
                 while wrapped is not None:
                     if wrapped is app:
-                        return srv
+                        return server
                     wrapped = getattr(wrapped, 'app', None)
             return None
         if (instance := _get_server_instance()) is not None:


falkoschindler:

I just pushed two major changes. Feel free to revert if you disagree.

1. Instead of searching for the server on shutdown, we can do it on startup. This way we can use the existing `Server.instance` field to store it and avoid the need for another (partly unused) field in core.py. We only need to relax the type for the server instance.
2. I tested with nicegui/examples/fastapi/main.py, adding a shutdown button and removing `reload=True`. The original PR didn't work and the app kept running. I had to traverse the `obj.config.loaded_app` and look for nested `app` objects.

The search takes about 3-4ms on my machine, which should be ok. But of course, moving a costly search to the startup is a downside of this approach.

Apart from that, I wonder if we can improve the `if self.config.reload:` condition in `shutdown()`. Currently the shutdown button fails if we run the uvicorn server with `reload=True`. But forcing an `os.kill` doesn't work either.

evnchn:
- Whether done on startup or shutdown, can we traverse `gc.get_objects()` only when the original `Server.instance` is not there? I prefer to keep the impact of this PR _only_ to those who use `ui.run_with`
- If `reload=True` fails for this PR after some effort, perhaps let's leave it. Something is better than nothing, after all.

falkoschindler:
> * Whether done on startup or shutdown, can we traverse `gc.get_objects()` only when the original `Server.instance` is not there? I prefer to keep the impact of this PR _only_ to those who use `ui.run_with`

The traversal is implemented within `ui.run_with`. So yes, the impact is limited to `ui.run_with`.

> * If `reload=True` fails for this PR after some effort, perhaps let's leave it. Something is better than nothing, after all.

Sure, I was just wondering if there's a solution insight. But fixing `reload=False` is good enough for now. 👍🏻

falkoschindler:
I think I found a simple solution: https://github.com/zauberzeug/nicegui/pull/5686/changes/3eecaf16400f1f7eeb7eba5edf97cd07dd22ce9b

All four cases (reload True/False, ui.run/ui.run_width) seem to work correctly.

evnchn:
Interesting 

When does `self.config.reload` diverge from `Server.instance.config.should_reload`

If it is never, could we check the latter only?

falkoschindler:
For `ui.run(reload=True)` the server instance is undefined in the child process.

PR #5697, opened 2026-01-29
Fix test isolation for pages defined in submodules

### LINKED ISSUE #5692: Test isolation failure in user fixture, resulting in 404s
### First Check

### Example Code

```python

```

## Summary 

My app uses `@ui.pages` defined in modules. When running multiple UI tests using the `user` fixture with a `main_file` configuration, the second (and subsequent) tests gets 404 errors because routes are not re-registered. The root cause seems to be that `nicegui_reset_globals()` only removes direct page modules from `sys.modules` but leaves parent packages cached.

### 1. Project structure

```
my_project/
├── main.py
├── pages/
│   ├── __init__.py
│   └── dashboard.py
├── tests/
│   └── test_ui.py
└── pytest.ini
```

### 2. main.py

```python
from nicegui import ui
import pages  # This imports and registers the @ui.page routes

    ui.run()
```

### 3. pages/__init__.py

```python
from .dashboard import dashboard_page
```

### 4. pages/dashboard.py

```python
from nicegui import ui

@ui.page('/dashboard')
def dashboard_page():
    ui.label('Dashboard')
```

### 5. pytest.ini

```ini
[pytest]
asyncio_mode = auto
main_file = main.py
addopts = -p nicegui.testing.user_plugin
```

### 6. tests/test_ui.py

```python
import pytest

@pytest.mark.asyncio
async def test_first(user: User):
    await user.open('/dashboard')
    await user.should_see('Dashboard')

@pytest.mark.asyncio
async def test_second(user: User):
    await user.open('/dashboard')  # This fails with 404!
    await user.should_see('Dashboard')
```

### 7. Run tests

```bash
python -m pytest tests/test_ui.py -v
```

## Expected Behavior

Both tests should pass, with routes being properly re-registered between tests.

## Actual Behavior

- First test passes
- Second test fails with: `AssertionError: Expected status code 200, got 404`

## Root Cause Analysis

The issue seems to be in [`nicegui/testing/general.py`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/testing/general.py) in the `nicegui_reset_globals()` function.

In the `finally` block:

```python
for func in Client.page_routes:
    if not func.__module__.startswith('tests.'):
        sys.modules.pop(func.__module__, None)
```

This only removes the direct module containing the page function (e.g., `pages.dashboard`) but leaves parent packages (`pages`) in `sys.modules`.

When the second test starts:
1. `nicegui_reset_globals()` removes all routes at the start
2. `runpy.run_path(main_file)` executes `main.py`
3. `main.py` does `import pages`
4. Since `pages` is still in `sys.modules`, Python returns the cached module
5. The cached module doesn't re-execute its imports
6. `@ui.page` decorators in `pages.dashboard` never run
7. Routes are not registered → 404 error

## Proposed Fix

In the `finally` block of `nicegui_reset_globals()`, also remove parent packages of the page modules:

```python
# Collect all modules to remove, including parent packages
modules_to_remove = set()
for func in Client.page_routes:
    if not func.__module__.startswith('tests.'):
        module_name = func.__module__
        modules_to_remove.add(module_name)
        # Also add parent packages
        parts = module_name.split('.')
        for i in range(1, len(parts)):
            modules_to_remove.add('.'.join(parts[:i]))

for mod_name in modules_to_remove:
    sys.modules.pop(mod_name, None)
```

Alternatively, a more targeted approach could remove all modules that are submodules of modules that were imported by the main_file.

## Temporary fix / workaround

In `conftest.py`, add an autouse fixture that clears page modules before each test:

```python
import sys
import pytest

@pytest.fixture(autouse=True)
def clear_page_modules():
    """Clear page modules to force re-registration."""
    modules_to_remove = [
        mod_name for mod_name in sys.modules.keys()
        if mod_name == 'pages' or mod_name.startswith('pages.')
    ]
    for mod_name in modules_to_remove:
        sys.modules.pop(mod_name, None)
    yield
```

This workaround forces a clean re-import of all page modules for each test.

### NiceGUI Version

3.5.0

### Python Version

3.14.2

### Browser

Chrome

### Operating System

macOS

### Additional Context

This issue affects any project that organizes routes in submodules/packages. It does not affect projects where all `@ui.page` decorators are in `main.py` directly.

[Personal note: this issue report mostly drawn up by Cline/claude-opus-4.5, after jointly wrestling with the issue for a few hours. -PK]

falkoschindler:
### Motivation

Fixes #5692

When using the `user` fixture with a `main_file` that imports `@ui.page` routes from submodules (e.g., `pages/dashboard.py`), the second and subsequent tests fail with 404 errors because routes are not re-registered.

The root cause is that `nicegui_reset_globals()` only removes the direct module containing the page function (e.g., `pages.dashboard`) from `sys.modules`, but leaves parent packages (e.g., `pages`) cached. When the next test runs `import pages`, Python returns the cached module without re-executing its imports, so `@ui.page` decorators never run again.

### Implementation

Modified the module cleanup logic in `nicegui_reset_globals()` to also remove parent packages of page modules. For example, if a page function is in `pages.dashboard`, we now remove both `pages.dashboard` and `pages` from `sys.modules`, ensuring a clean re-import on the next test.

### Progress

- [x] Pytest has been added.

### DIFF
--- a/tests/test_user_simulation_context.py
+++ b/tests/test_user_simulation_context.py
@@ -100,3 +100,28 @@ async def test_adding_page_to_root_function():
 
         await user.open('/a')
         await user.should_see('Page A')
+
+
+@pytest.mark.parametrize('_first_run', [True, False])
+async def test_main_file_with_submodule_pages(tmp_path: Path, _first_run: bool):
+    (pages_dir := tmp_path / 'pages').mkdir()
+    (pages_dir / '__init__.py').write_text('from .dashboard import dashboard_page')
+    (pages_dir / 'dashboard.py').write_text(textwrap.dedent('''
+        from nicegui import ui
+        @ui.page('/dashboard')
+        def dashboard_page():
+            ui.label('Dashboard Content')
+    '''))
+
+    main_file = tmp_path / 'main.py'
+    main_file.write_text(textwrap.dedent('''
+        import sys
+        sys.path.insert(0, str(__file__).rsplit('/', 1)[0])
+        import pages
+        from nicegui import ui
+        ui.run()
+    '''))
+
+    async with user_simulation(main_file=main_file) as user:
+        await user.open('/dashboard')
+        await user.should_see('Dashboard Content')


nicegui/testing/general.py
```diff
@@ -76,7 +76,9 @@ def nicegui_reset_globals():
 
         for func in Client.page_routes:
             if not func.__module__.startswith('tests.'):
-                sys.modules.pop(func.__module__, None)
+                parts = func.__module__.split('.')
+                for i in range(len(parts)):
+                    sys.modules.pop('.'.join(parts[:i+1]), None)  # remove the module and all its parents
```
evnchn, line 81:
```suggestion
                module = func.__module__
                while module:  # remove the module and all its parents
                    sys.modules.pop(module, None)
                    module = module.rpartition('.')[0]
```

Considering:

- More performant since less joins. 
- First removes `aaa.bbb.ccc` then remove `aaa.bbb`, could avoid interesting edge cases where first deleting the parent throws off the child. 

Note:

```
>>> "aaabbb".rpartition(".")
('', '', 'aaabbb')
>>>
```

So when no dot inside `modules`, next call to `module = module.rpartition('.')[0]` empties `modules`

evnchn (dismissed):
Should work. Two optional questions:

- Do we add tests? We could but it may be hacky though...
- How is my suggested code?

If you think is adoptable, proceed. If not, then feel free to merge.

nicegui/testing/general.py
falkoschindler: The suggestion is reasonable but the claimed benefits are marginal:

1. **Performance**: The claim about "less joins" is technically true, but the difference is negligible. Module names typically have 1-3 dots, so we're talking about a handful of string operations during test teardown. Not a meaningful optimization.

2. **Order of removal (child-first)**: You suggest removing `aaa.bbb.ccc` before `aaa.bbb` "could avoid interesting edge cases where first deleting the parent throws off the child." However, `sys.modules` is just a dict - removing one key doesn't affect other keys. There's no parent-child relationship at the dict level. Both orders work correctly.

3. **Readability**: The `while` + `rpartition` approach is slightly more elegant conceptually ("keep chopping off the last segment"), but the `for` + `split`/`join` approach is more explicit about what's happening. It's a wash.

Let's keep the code unchanged.

falkoschindler:
Yes, a pytest won't hurt. It's a bit hacky and regressions in general.py are unlikely, but user simulation tests are cheap, so why not.

PR #5702, opened 2026-02-02
Fix shutdown handler not called with native mode and reload enabled

### LINKED ISSUE #2107: When native is True and reload is True, port is ignored and app_shutdown never fires
### Description

Platform:
Linux Mint 20.03 (ubuntu)
NiceGUI 1.4.3

Steps to replicate:
Save the code below as main.py and run `python main.py`

```py
from nicegui import app, ui

def app_startup():
    print('app startup')

def app_shutdown():
    print('app shutdown')

app.on_startup(app_startup)
app.on_shutdown(app_shutdown)

ui.run(
    title="Some app",
    port=8502,
    native=True,
    reload=True,
    window_size=(1920, 1200),
    storage_secret='my secret'
)
```

Expected: Window app opens up, server starts on port 8502, app startup printed out.
Actual:  Window app opens up, app startup printed out, but server starts on port 8000 (note: I double checked that 8502 is NOT taken by some other process)

Then close the window by clicking "X"

Expected: app shutdown printed out
Actual: nothing is printed out, the app_shutdown function is never called

evnchn:
### Motivation

Fixes #2107, where when using native=True and reload=True together, closing the native window would not trigger the app.on_shutdown() handler. 

### Implementation

**Claude is used: First get it to have _a_ working solution, then ask it to "condense" the diff for 4 rounds until we arrive here.**

It was due to the way uvicorn manages processes in reload mode, preventing the shutdown signal from propagating correctly.

The fix introduces a multiprocessing.Event that is shared between the native mode process and the server. When the native window is closed, it sets this event, and a monitoring thread in the server detects this and triggers a graceful shutdown.

### Test script

```py
from nicegui import app, ui

def app_startup():
    print('app startup')

def app_shutdown():
    print('app shutdown')

app.on_startup(app_startup)
app.on_shutdown(app_shutdown)

@ui.page('/')
def main():
    return # remove this if under Windows, and you want to auto-close the window. Useful for agentic coding. 
    import subprocess

    def close_native_window():
        subprocess.run([
            'powershell', '-Command',
            '''
            Add-Type @"
                using System;
                using System.Runtime.InteropServices;
                public class Win32 {
                    [DllImport("user32.dll")]
                    public static extern IntPtr FindWindow(string lpClassName, string lpWindowName);
                    [DllImport("user32.dll")]
                    public static extern bool PostMessage(IntPtr hWnd, uint Msg, IntPtr wParam, IntPtr lParam);
                    public const uint WM_CLOSE = 0x0010;
                }
"@
            $hwnd = [Win32]::FindWindow([NullString]::Value, "NiceGUI")
            if ($hwnd -ne [IntPtr]::Zero) {
                $result = [Win32]::PostMessage($hwnd, [Win32]::WM_CLOSE, [IntPtr]::Zero, [IntPtr]::Zero)
            }
            '''
        ])
        print('Native window closed', flush=True)

    ui.timer(0.5, close_native_window, once=True)

ui.run(
    native=True,
    reload=True, # reload=False works always fine before and after this PR
)
```

<details>
<summary> Development tooling (the script that Claude made, and subsequently grinded against </summary>

```py
"""Test runner with timestamped output and 25-second timeout."""
import subprocess
import sys
import time
import threading

TIMEOUT = 25
EXPECTED_LINES = ['app startup', 'Native window closed', 'app shutdown']

def run_test():
    start_time = time.time()
    found_lines = []

    proc = subprocess.Popen(
        [sys.executable, 'test.py'],
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        text=True,
        bufsize=1,
        cwd=r'c:\Users\User\nicegui'
    )

    def read_output():
        for line in proc.stdout:
            elapsed = time.time() - start_time
            line_stripped = line.rstrip()
            print(f'[{elapsed:6.2f}s] {line_stripped}', flush=True)
            for expected in EXPECTED_LINES:
                if expected in line_stripped and expected not in found_lines:
                    found_lines.append(expected)

    reader_thread = threading.Thread(target=read_output, daemon=True)
    reader_thread.start()

    try:
        proc.wait(timeout=TIMEOUT)
    except subprocess.TimeoutExpired:
        print(f'\n[TIMEOUT] Process exceeded {TIMEOUT}s limit, killing...', flush=True)
        proc.kill()
        proc.wait()

    reader_thread.join(timeout=1)
    elapsed_total = time.time() - start_time

    print(f'\n{"="*50}')
    print(f'Total time: {elapsed_total:.2f}s')
    print(f'Expected lines found: {len(found_lines)}/{len(EXPECTED_LINES)}')
    for expected in EXPECTED_LINES:
        status = '[Y]' if expected in found_lines else '[N]'
        print(f'  {status} "{expected}"')

    if len(found_lines) == len(EXPECTED_LINES) and elapsed_total < TIMEOUT:
        print('\n[PASS] All lines found within TIMEOUTs')
        return 0
    else:
        print('\n[FAIL] Missing lines or exceeded TIMEOUTs')
        return 1

if __name__ == '__main__':
    sys.exit(run_test())
```

</details>

### Progress

- [x] Native mode out-of-scope for Pytest coverage
- [x] Documentation is not necessary for an underlying bugfix

### DIFF
--- a/nicegui/server.py
+++ b/nicegui/server.py
@@ -39,7 +39,7 @@ class Server(uvicorn.Server):
                 def monitor_shutdown_event() -> None:
                     event.wait()
                     self.should_exit = True
-                threading.Thread(target=monitor_shutdown_event).start()
+                threading.Thread(target=monitor_shutdown_event, daemon=True).start()
 
         storage.set_storage_secret(self.config.storage_secret, self.config.session_middleware_kwargs)
         super().run(sockets=sockets)


nicegui/server.py
```diff
@@ -32,6 +35,11 @@ def run(self, sockets: list[socket.socket] | None = None) -> None:
             core.app.native.main_window = native.WindowProxy()
             native.method_queue = self.config.method_queue
             native.response_queue = self.config.response_queue
+            if (event := self.config.shutdown_event) is not None:
+                def monitor_shutdown_event() -> None:
+                    event.wait()
+                    self.should_exit = True
+                threading.Thread(target=monitor_shutdown_event).start()
```
falkoschindler: Opus 4.5 suggests setting `daemon=True` to avoid blocking process exit in edge cases. What do you think?

evnchn: It had `daemon=True` in the beginning but then I asked it to "remove all code that is not strictly necessary for the Test script to pass", and it was gone. 

A question we have to ask ourselves is: How likely is the server to attempt shutdown without calling `shutdown_event.set()`? As that is the only case where `daemon=True` can matter. 

I would say that it would never happen in sunny day, but I am not sure if there are rainy day cases.

evnchn: One rainy day possibility is "Unhandled exception in main thread", where the server shut itself down. 

In spirit of https://github.com/zauberzeug/nicegui/pull/5706 I think `daemon=True` is better to ensure bounded waiting. Drop if disagree.

falkoschindler: Yes, I think `daemon=True` is better: _Maybe_ it handles weird edge cases, but it also documents that this process isn't meant to keep running after the parent process ends.

PR #5703, opened 2026-02-02
inherit SessionMiddleware from parent app in ui.run_with

### LINKED ISSUE #2578: Attaching to a FastAPI instance with another storage_secret breaks session storage
### Description

I noticed that when attaching to a FastAPI instance using `ui.run_with` one has to use the same secret like the FastAPI SessionMiddleware - otherwise the session data is not persistence across page reloads. This seems to be problematic in two ways: 
 1. Setting a storage_secret seems to be redundant as it is already set in the middleware.  
 2. Setting another secret does not result in an error but instead results in non-persistent sessions. 
 
Debugging, took me a while, so I would recommend to improve the feedback if the user tries to set another password and/or to improve the web examples for attaching to an existing FastAPI instance (e.g. show the example how to attach to an app with a installed middleware).
 
When writing the MRCE I also noticed, that initializing NiceGUI will break all endpoints defined later in the code via FastAPI. I don't know if this is problematic, but it seems weird. You may have a look at this.
 
MRCE:

File 1: main.py
```py
import uvicorn
from ng_page import storage
from nicegui import ui

# /test1 : http://localhost:8080/test1
# /test2 : http://localhost:8080/test2
# /testR : http://localhost:8080/testR
# /storage : http://localhost:8080/storage


app = FastAPI()
refresh_count = 0


@app.get("/testR")
def tR(request: Request):
    global refresh_count
    request.session['refresh_count'] = refresh_count
    refresh_count += 1
    return f"refresh to increase counter: {refresh_count}"


@app.get("/test1")
def t1():
    return "test1"


case = 3

if case == 0:  # ----- No key set -----
    pass
    # /test1 is accessible
    # /test2 is accessible
    # /testR gives an error (because the SessionMiddleware is not installed)
    # /storage is not accessible (because nicegui is not running)

if case == 1:  # ----- Setting a key only for the fastapi app -----
    app.add_middleware(SessionMiddleware, secret_key='themagickey')

    # /test1 is accessible
    # /test2 is accessible
    # /testR is accessible and the session is working
    # /storage is not accessible (because nicegui is not running)

if case == 2:  # ----- Setting a key only for the storage -----
    ui.run_with(app, storage_secret='aregularkey')

    # /test1 is accessible
    # /test2 is not accessible (initializing nicegui breraks any later defined access points in the fastapi app)
    # /testR gives an error (because the SessionMiddleware is not installed)
    # /storage is accessible and the storage is working

if case == 3:  # ----- Setting key for the fastapi app but not for NiceGUI
    app.add_middleware(SessionMiddleware, secret_key='themagickey')
    ui.run_with(app)

    # /test1 is accessible
    # /test2 is accessible
    # /testR is accessible and the session is working
    # /storage gives hint to set the storage_secret

if case == 4:  # ----- Setting different keys for the app and the storage -----
    app.add_middleware(SessionMiddleware, secret_key='themagickey')
    ui.run_with(app, storage_secret='aregularkey')

    # /test1 is accessible
    # /test2 is not accessible (see case 2)
    # /testR is accessible and the session is working
    # /storage is accessible but the storage is not working

if case == 5:  # ----- Setting the same key for both the app and the storage -----
    app.add_middleware(SessionMiddleware, secret_key='themagickey')
    ui.run_with(app, storage_secret='themagickey')

    # /test1 is accessible
    # /test2 is not accessible (see case 2)
    # /testR is accessible and the session is working
    # /storage is accessible and the storage is working


@app.get("/test2")
def t2():
    return "test2"


if __name__ == "__main__":
    uvicorn.run("main:app", host="0.0.0.0", port=8080, lifespan='on', use_colors=True, reload=True,)

```

File 2: ng_page.py

```py
from nicegui import app, ui


@ui.page("/storage")
def storage():

    status_label = ui.label('')
    detail_label = ui.label('')

    def update_labels():
        status_label.text = f'User is authenticated: {app.storage.user.get("authenticated", False)}'
        detail_label.text = f'"authenticated" in storage: {"authenticated" in app.storage.user.keys()}'

    def on_authenticate():
        app.storage.user['authenticated'] = True
        update_labels()

    def on_deauthenticate():
        app.storage.user['authenticated'] = False
        update_labels()

    def clear_storage():
        app.storage.clear()
        update_labels()

    ui.button('Authenticate', on_click=on_authenticate)
    ui.button('Deauthenticate', on_click=on_deauthenticate)
    ui.button('Clear storage', on_click=clear_storage)
    ui.button('Reload', on_click=lambda: ui.run_javascript('window.location.reload()'))

    update_labels()
```

Expected/ideal behavior:
- case 2: I would expect '/test2' to work.
- case 3: I would expect '/storage' to work, because I set a secret_key in the middleware
- case 4: I would expect an error because I used two different keys

evnchn:
### Motivation

Fixes #2578, which boils down to how `ui.run_with()` absolutely does not play well with parent FastAPI app's `SessionMiddleware`

### Implementation

When using ui.run_with() with a parent FastAPI app that already has SessionMiddleware installed, NiceGUI now:

- Inherits the secret if no storage_secret is provided or if the same secret is given (fixes case where storage silently failed)
- Detects conflicts when different secrets are used with the same cookie name, and automatically renames NiceGUI's cookie to avoid silent failures, with a warning guiding the user

### Progress

- [x] Running with a FastAPI instance is not-in-scope for the Pytest
- [x] Docstring will be self-explanatory and serve as documentation. But it does not show in https://nicegui.io though.

### DIFF
--- a/nicegui/storage.py
+++ b/nicegui/storage.py
@@ -37,7 +37,7 @@ class RequestTrackingMiddleware(BaseHTTPMiddleware):
 
 
 def set_storage_secret(storage_secret: str | None = None,
-                       session_middleware_kwargs: dict[str, Any] | None = None,
+                       session_middleware_kwargs: dict[str, Any] | None = None, *,
                        parent_app: FastAPI | None = None) -> None:
     """Set storage_secret and add request tracking middleware.
 


falkoschindler:
@evnchn I just wanted to finally finish reviewing this PR, but once again struggled to fully understand it. Therefore I started from first principles and created PR #5857. Maybe you can have a look and tell me what is missing. Thanks!

evnchn:
Ref https://github.com/zauberzeug/nicegui/pull/5857: Feature-complete isn't always the NiceGUI answer, but I am glad this PR did exist. 

It's easy (may need AI but still) to cut away features, but hard (not even Opus can do it) to think about what missing features are there. It's always "You're absolutely right!"

Had `session_cookie` be a mandatory feature and we started with #5857 we would never discover this PR's code and we'd be caught in a bad surprise and have to ship a patch release.

PR #5704, opened 2026-02-02
Add dynamic resampling for large Plotly datasets

### LINKED ISSUE #3340: Updating Ploty plot with large dataset causes the UI to hang
## Description

I'm trying to plot some audio data with a sample rate of 48000 Hz. Updating the Plotly data with a large dataset causes the whole NiceGUI to hang. I'm not quite sure how to reproduce the issue because same code sometimes works but sometimes does not.

## Minimal code

Here is my minimal code, hope someone knows what's going on.

```python
import random
import asyncio


fig = go.Figure(
    data={"type": 'scatter'},
    layout={"margin": dict(l=20, r=20, t=20, b=20)}
)
plot = ui.plotly(fig)

async def update_data():
    update_bt.props('loading')  # update button
    try:
        # Clear data
        fig.update_traces(x=[], y=[])
        plot.update()
        
        await asyncio.sleep(1)  # refresh UI
        
        # Update plot data
        fig.update_traces(x=x, y=y)
        plot.update()
    finally:
        update_bt.props(remove='loading')  # update button
        print('The Plotly plot should be refreshed')  # the ploty plot does not be refreshed after this line

update_bt = ui.button('Update data', on_click=update_data)
ui.button('Force reload', on_click=ui.navigate.reload)  # force reload always works

# 50-sec audio data with 48000 Hz sample rate
x = [ float(n) for n in range(48000*50) ]  # using numpy arrays might avoid ui hanging
y = [ random.uniform(-i, i) for i in x ]  # using numpy arrays might avoid ui hanging

ui.run(port=native.find_open_port(), native=True)
```

I've added a force reload button at the bottom. If the plotly plot does not be refreshed after the `update_data` returns, clicking the force reload button always works.

## My hardware

M2 Pro Mac Mini
macOS 14.5

## Python packages

python 3.10
nicegui==1.4.29
pywebview==5.1
plotly==5.22.0

Thank you for any help you can provide.

evnchn:
### Motivation

Part of #4714 (built-in support for plotting large datasets). Even with Plotly's binary `bdata` encoding making the *transfer* of many points cheap, plotly.js still has to *render* every point it receives, so it chokes on very large traces. This PR raises that rendering ceiling by downsampling on the server.

(Originally opened against #3340, which turned out to be a Plotly list-vs-NumPy encoding issue and was closed with a docs note. This PR solves the complementary rendering-ceiling problem, so it is re-anchored to the broader #4714 effort.)

### Implementation

Implement LTTB (Largest Triangle Three Buckets) downsampling for large scatter data in Plotly charts.

- Add opt-in `n_samples` parameter to `ui.plotly()` (default `None` → no change in behavior)
- Retain the full dataset on the server; send at most `n_samples` points per trace to the client
- Resample the visible range on zoom/pan via `plotly_relayout`; resample the full range on axis reset
- Include a pure-NumPy LTTB implementation as a dependency-free fallback
- Support the optional `tsdownsample` package for a faster Rust-based implementation (`pip install nicegui[tsdownsample]`)

### Progress

- [x] If this PR addresses a security issue, it has been coordinated via the security advisory process.
- [x] Pytests validate that resampling took place: assert only `n_samples` points reach the client (handling both plain-list and `bdata` encodings), the full dataset is retained server-side, and zoom / reset resample correctly.
- [x] The feature is showcased in the `ui.plotly` documentation ("Dynamic resampling" demo).


evnchn:
Welcome to add documentation and pytest, as well as bug-spotting.

NvNielen:
Hello @evnchn, I had a look at your implementation and it looks great! 

I am also not quite sure how to test the downsampling. All I could think of is checking for each trace, that when x and y exceed n_samples, that they should have reduced down to length n_samples after downsampling.

I have a question regarding plot types. You check when extracting the data:
```py
if trace_dict.get('type', 'scatter') not in ('scatter', 'scattergl'):
    continue
```
Wouldn't `line` also be a valid time series type?

Lastly, maybe it would also be nice to add an option to use [datashader](https://datashader.org/). Both for some [applications where tsdownsampler doesn't quite work as well](https://predict-idlab.github.io/plotly-resampler/v0.9.0/FAQ/) (please see section _What is the difference in approach between plotly-resampler and datashader?_) and for rasterization of large/dense scatterplots or other types such as heatmaps. Though I am uncertain whether that would overcomplicate things and should be part of this PR.

evnchn:
> Wouldn't line also be a valid time series type?

Not sure. Maybe you can override locally, test out the code with a line time series type, and let me know if it works properly. If it does, let's include it. 

I think we should focus on plotly-resampler, not datashader, for now. Let's keep the PR size manageable. You are more than welcome to make another PR though!

falkoschindler:
Thanks for putting this together, @evnchn — the LTTB implementation is nice work.

I dug into the root cause of #3340 (see my comment over there) and I think it changes how we should frame this PR, though not whether it's valuable.

It turns out the specific freeze in #3340 is a Plotly encoding issue: a Python list is sent as a plain JSON array of numbers, while a NumPy array is encoded as Plotly's compact binary `bdata` typed array. So passing NumPy arrays already resolves that particular case, and I'd like to close #3340 with a docs note rather than a code change.

That said, I don't think this PR becomes unnecessary — it solves a different bottleneck. The `bdata` encoding makes _transferring_ N points cheaper, but it doesn't raise plotly.js's _rendering_ ceiling: even plain Plotly (no NiceGUI) chokes on millions of points. Downsampling is what keeps rendering smooth regardless of source size. So this is complementary to the encoding fix, not redundant with it.

Given that, I'd suggest re-anchoring this PR from "Fixes #3340" to the broader large-dataset effort in #4714, which is exactly the problem LTTB resampling addresses.

A few things I'd still like to work through before merging (happy to discuss):

- The open question in your checklist about how to test that resampling actually happened.
- Whether automatic resampling should be opt-in via `n_samples` (as you have it) and how it interacts with user-driven zoom/relayout in edge cases.
- A small showcase/demo so the feature is discoverable.

What do you think about pointing this at #4714?

evnchn:
Thanks @falkoschindler — that root-cause on #3340 (list vs NumPy `bdata`) is a great catch, and I agree with closing it via the docs note. Happy to re-anchor: I've dropped `Fixes #3340` and pointed this at #4714, framed as the rendering-ceiling complement to the transport fix rather than a replacement.

I've also pushed an update that rebases onto `main` (the PR had gone conflicting) and works through the three points you raised. Since it was a rebase I force-pushed with `--force-with-lease` and squashed to a single clean commit — net feature diff is +480 across 5 files; the rest of the range-diff is just the `main` catch-up.

On your three points:

- **Testing that resampling happened** — it's server-side, so no browser needed. Added `pytest` cases asserting that only `n_samples` points reach the client (decoding both the plain-list and binary `bdata` forms), that the full dataset is retained server-side, and that zoom / combined-range-key / autorange-reset resample correctly.
- **Opt-in via `n_samples` + edge cases** — kept it opt-in (`None` default is fully non-breaking). I ran the diff through a review pass and it surfaced some real sharp edges around the edge cases you flagged; I fixed the correctness ones and documented the rest (details folded below).
- **Showcase** — added a "Dynamic resampling" demo to the `ui.plotly` docs (100k-point random walk).

<details>
<summary><b>Edge cases found &amp; how they're handled</b> (review pass — fixed vs documented)</summary>

| Edge case | Before | Now |
|---|---|---|
| **datetime / string x** (the "time series" case) | `np.asarray(x, dtype=float64)` → crashes on `datetime`/date-string x, or renders `datetime64` as bare epoch floats | Original x kept and indexed with the LTTB result, so datetime/int dtype survives the round-trip; numeric strings and `None` still behave as numbers; genuinely non-numeric traces are skipped, not crashed |
| **Caller's figure mutated** | resampling wrote the downsampled arrays back into the user's own `dict`/`go.Figure` | figure is deep-copied when `n_samples` is set — the caller's object is never touched (also decodes Plotly's binary `bdata` typed-array form on the way in) |
| **`n_samples < 3`** | `tsdownsample` (Rust) panics | guarded — keeps the endpoints |
| **Combined `xaxis.range` relayout key** | only the split `range[0]`/`range[1]` keys were parsed | both forms parsed; datetime range strings converted |
| **Subplots / multiple x-axes** | a zoom on `xaxis2` was applied to every trace | now tracks only the **primary** x-axis; subplots are left un-resampled (documented). Open to doing per-axis mapping if you'd like subplots in scope — felt like a design call worth your input rather than something to silently commit |
| **Unsorted x (line traces)** | silently sorted by x, reordering the drawn path | unchanged (LTTB needs sorted x) but now documented in the docstring |

</details>

<details>
<summary><b>Local CI</b></summary>

```
pre-commit  … all passed
mypy ./nicegui … Success: no issues found in 244 source files
pylint ./nicegui … 10.00/10
pytest tests/test_plotly.py … 12 passed
```

Both downsampler paths (Rust `tsdownsample` and the pure-NumPy fallback) verified end-to-end, including datetime x and zoom/reset.

</details>

What do you think about the subplot question — leave it primary-axis-only for now, or is multi-axis support something you'd want before this lands?

evnchn:
IMO: Support one plot for now. Why subplots when can compose plots using NiceGUI and achieve 90% function?

evnchn:
@NvNielen — you kindly looked at this back in March, and since you authored the uPlot element you've got the sharpest plotting eyes here. Up for a quick vibe-check on the current diff?

Since your last look it's had real work: re-anchored from #3340 to the large-dataset effort in #4714 (Falko's steer), added server-side pytest coverage that resampling actually happens, kept it opt-in via `n_samples` (fully non-breaking), and added a docs demo. CI's green.

On your March question — whether `line` traces should count as a time-series type — I kept it scatter/scattergl for now to hold scope, but you're the right person to say if line should be in from the start. Also kept it single-plot deliberately (no subplot resampling).

No rush — just value your read on whether the shape vibes before it goes back to Falko for merge review. 🙏

falkoschindler:
We didn't manage to get this into 3.15 — moving it to the 3.16 milestone. Thanks for your patience, it's still on our radar.

evnchn:

**Tests moved onto the public `props` accessor** — the seven `protected-access` disables and the private `_hf_data` assertion are gone, per the testing policy in `#6219`.

<details>
<summary><b>What changed, and why the payload assertion stays</b></summary>

The tests previously read `plot._props['options']['data']` behind seven `# pylint: disable=protected-access` comments and asserted on the private `plot._hf_data`. They now go through `Element.props` — the public accessor, documented as "The props of the element", with existing precedent in `tests/test_defaults.py` and `tests/test_element_filter.py`. The `_hf_data` assertion is deleted: the claim it made (the full dataset is retained server-side) is instead proved observably, by zooming into a 51-point sub-range and getting all 51 points back at full resolution, for both the `go.Figure` and `dict` input forms.

The remaining coupling to `props['options']` is deliberate — the feature's contract is *"only `n_samples` points are transmitted to the client"*, so the payload **is** the behavior under test. Both alternatives were tried and rejected:

- **Asserting on the public `plot.figure` attribute.** With `update()` removed from the relayout handler (resample server-side, never push), `figure` shows the correct 51 zoomed points while the client still holds a stale 100-point payload — so a `figure`-based test would pass that bug.
- **Recording outgoing messages.** There is no public testing API for it; `User` patches `outbox._emit` only to answer `run_javascript`, and re-doing that patch in a test is the "patch to observe an internal call" the policy names. A `Screen` test would have to read plotly.js internal state to see the same thing, at browser cost.

Revert-and-fail, four mutations, each restored afterwards:

| Mutation | Result |
|---|---|
| `if n_samples:` → `if False:` (feature reverted) | 6 failed |
| drop `self.update()` in `_handle_relayout` | 3 failed |
| server keeps only `n_samples` points | 6 failed — the dropped `_hf_data` claim, covered observably |
| retention bug on `go.Figure` only | 1 failed |

That last row is a gap found in review, not by the rewrite: the deleted `_hf_data` assertion had run over *both* figure forms, while the zoom test replacing it covered `dict` only, so `go.Figure` server-side retention briefly went unguarded — a `go.Figure`-only truncation passed 12/12. Closed by extending the test that already builds both forms, and re-verified failing.

12 passed, 0 skipped; pre-commit, `mypy ./nicegui` (246 files), `pylint ./nicegui` 10.00/10 all green.
</details>

PR #5700, opened 2026-02-01
Fix video loading issues in Windows in native mode

evnchn:
### Motivation

The following fails:

```py
from nicegui import ui

N = 2 #4970

for _ in range(N):
    ui.video("test.mp4")

ui.run(native=True, reload=False)
```

with `RuntimeError: No response returned.`

### Implementation

Fix 1: async generator fixes #4970

I just let Claude to hammer on it for half an hour. Apparently switching to an async generator fixed it. 

I **do not** speak for or against its reasoning attached below, but I **can** confirm that this consistently fixes it on my machine out of 10 trials without the fix, 10 trials with. 

I triple-checked that the error-masking [`ApparentlyUselessMiddleware`](https://github.com/zauberzeug/nicegui/issues/4970#issuecomment-3830672666) was not used the whole time. 

<details>
<summary> Its reasoning </summary>

When you have 2+ video elements requesting the same file:

1. Multiple concurrent range requests hit the same streaming endpoint
2. The synchronous generator blocks the event loop during file I/O
3. Starlette's middleware coordination gets disrupted when generators don't yield control
4. This causes the "No response returned" error when clients disconnect or make concurrent requests

</details>

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.
- [x] Pytests is non-feasible as we do not have Windows worker configured right now. 
- [x] Documentation is not necessary for a bugfix. 

### Final notes

CC @MaidScientistIzutsumiMarin @python-and-novella, kindly test and see if this does the trick. Thanks!

`pip install git+https://github.com/evnchn/nicegui.git@mysterious-fix-4970`

PR #5706, opened 2026-02-02
set timeout_graceful_shutdown to prevent hanging on ConnectionResetError

### LINKED ISSUE #5443: ConnectionResetError from loading too many video elements in native browser
### First Check

### Example Code

```python

import anyio
from nicegui import app, ui


        try:
        except RuntimeError as exc:
            raise
        except anyio.EndOfStream:


def main() -> None:
    for _ in range(20):
        ui.video("test.mp4")


```

### Description

Start the app.
```python
```

### NiceGUI Version

3.2.0

### Python Version

Python 3.14.0

### Browser

Other

### Operating System

Windows

### Additional Context

evnchn:
### Motivation

Fixes #5443, where when ConnectionResetError occurs (particularly with multiple videos in native mode on Windows), the server may hang indefinitely, unresponsive to Ctrl+C termination. 

Additional debugging context in #5700's comments. 

### Implementation

By setting timeout_graceful_shutdown (default 10 seconds, which is plenty long, considering Redis also times out in 10 seconds), the server will forcefully terminate requests after this timeout elapses, ensuring the application exits rather than hanging.

### Progress

### Final notes

It appears that `ConnectionResetError` is inevitable on Windows. Nevertheless I think this PR is also good to ensure a bounded-wait termination (surprised it isn't originally the case tbh)

CC @python-and-novella 

At this point we can consider to cherry-pick:

- https://github.com/zauberzeug/nicegui/pull/5700/commits/1ce1a72b46d75389197e47f5facb560c4d531a02
- https://github.com/zauberzeug/nicegui/pull/5700/commits/1c656164e9e8d37a4b0ff747c52b88a05600f43d
- https://github.com/zauberzeug/nicegui/pull/5700/commits/6c245ceaee3ea840b205e27807dfc988beeb2600

### DIFF
--- a/nicegui/ui_run.py
+++ b/nicegui/ui_run.py
@@ -278,6 +278,7 @@ def run(root: Callable | None = None, *,
         reload_dirs=split_args(uvicorn_reload_dirs) if reload else None,
         log_level=uvicorn_logging_level,
         ws='wsproto',
+        timeout_graceful_shutdown=kwargs.pop('timeout_graceful_shutdown', 10),
         **kwargs,
     )
     config.storage_secret = storage_secret


evnchn:
I think that having it at 10 seconds may snipe some people who have long background downloads. 

What about warning, when on Windows, if `timeout_graceful_shutdown` is not set? This way the user is kept in-the-loop.

falkoschindler:
After discussing the matter with Opus 4.5, I'd suggest:

1. Update the PR description or issue to clarify this is a **partial mitigation**, not a full fix
2. Keep #5443 open (or create a follow-up) for the lag/performance aspect
3. Consider whether the cherry-picks add value beyond cosmetic error suppression - if they just hide the errors without fixing anything, that might mask future debugging

The real fix for #5443 would likely require investigating *why* Windows connections reset when loading multiple videos - that's a different, harder problem.

evnchn:
@falkoschindler I want to re-position this as the canonical fix. Let's consider cross-platform. 

On Linux and macOS (so UNIX systems), Ctrl+C terminates all in-flight HTTP requests quickly and forcefully (milisecond-level). It is only Windows who hang indefinitely. 

A bounded and forceful termination is the way to go to align NiceGUI's Windows behaviour to UNIX's, which is arguably the correct one per the "bounded waiting" principle of OS. 

Similar to how Windows waits, wall-clock, 3 seconds before force-closing a C++ program which leaked memory, we should populate `timeout_graceful_shutdown` to `3` for Windows users. 

`10` is already too lose when the canonical value to ideally aim for is `0` (but we can't put that because legitimate file writes don't get flushed that way). 

Sure, there may be legitimate uses for the current unbounded termination, but that is likely a hack-job with a `taskkill` armed somewhere because a `taskkill` is, AFAIK, the only way to close the process which hanged indefinitely in this bug. 

Hence, the ultimate recommendation is to **auto-populate `timeout_graceful_shutdown` to `3`, no need to log this, but in the release notes we note this new behaviour which is considered as a bugfix (so can ship in minor).**

evnchn:
*Posted by Claude Code on @evnchn's behalf — Evan asked me to summarize where we landed after probing his empirical Windows testing.*

**TL;DR:** Agree with @evnchn's auto-populate proposal and "bugfix in minor" framing. One disagreement: I'd ship the value at **10s, not 3s**, and keep it cross-platform rather than Windows-only.

<details>
<summary>Rationale</summary>

Three things sharpen the picture after probing the `ui.video` native-mode repro:

1. **Mechanism is exactly right.** With this PR (`timeout_graceful_shutdown=10`), the process cleanly terminates ~10s after Ctrl+C on Windows. That confirms uvicorn's `asyncio.wait_for` timeout is the only reliable escape — Ctrl+C spam doesn't help because Python's SIGINT handler on Windows only runs when the main thread re-enters the interpreter (`PyErr_CheckSignals`), and the Proactor event loop's IOCP polling starves signal delivery. The asyncio scheduler itself stays alive, so the `wait_for` timer fires even when signal-driven shutdown is wedged. `taskkill` afterwards immediately releases port 8080, ruling out a Winsock leak — the hang is purely Python-side, waiting on a `ConnectionResetError`-orphaned entry in `server_state.connections` that never completes.

2. **3s is over-tuned without evidence.** 10s already works for the repro with zero indication it's too long. Going to 3s risks truncating legitimate ~3–5s flushes (DB commits, file syncs in shutdown handlers) for no demonstrated gain. The "Windows force-closes leaked C++ programs in 3s" analogy doesn't quite hold either — Windows' own `WaitToKillServiceTimeout` default is 5000ms.

3. **Cross-platform costs nothing.** UNIX doesn't hit the orphan-connection path, so a 10s bound is a no-op there. Branching for Windows-only adds complexity without benefit and avoids platform-dependent surprises.

Re: logging — uvicorn already emits `"Cancel N running task(s), timeout graceful shutdown exceeded"` at `server.py:293` when the timeout fires, so debuggability is preserved without adding anything.

</details>

@falkoschindler — over to you:

- If you're happy with **10s**, this PR can merge as-is.
- If you'd rather ship **3s**, edit the literal in `ui_run.py` and merge.

Either way, please add a line to the release notes flagging the new bounded-shutdown behavior so anyone relying on unbounded graceful shutdown isn't surprised.

falkoschindler:
See https://github.com/zauberzeug/nicegui/issues/5443#issuecomment-4509935595 for why I'm still hesitating to accept this PR.

falkoschindler:
I guess this can be closed as superseded by PR #6105.

evnchn:
We might fallback to this if #6105 doesn't go our way, but no need to keep an open PR when the diff is literally:

```
timeout_graceful_shutdown=kwargs.pop('timeout_graceful_shutdown', 10),
```

PR #5707, opened 2026-02-02
`uv.lock` revision = 3

evnchn:
### Motivation

This keeps coming up when I try to add dependencies, but I just "revert block" to dodge it for the time being.

### Implementation

By adding a fake dependency then subsequently deleting it, it tricked `uv` into bumping up to `revision = 3`. 

### Progress

### Final notes

**A very very quick win. Please merge ASAP to get this friction point away.**

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.10, <4"
 resolution-markers = [
     "python_full_version >= '3.14'",


falkoschindler:
@evnchn Please update the description and tell what you actually did. The diff is hidden by default and the idea of the PR should be clear without reading the code. Thanks!

evnchn:
Can do, but note that the diff is literally:

image

falkoschindler:
Ok, now I get it:

When contributors use newer versions of `uv` to add or modify dependencies, they encounter a prompt indicating that the lock file revision is outdated and needs to be updated. This creates unnecessary friction during development, as the contributor must either:

1. Accept the revision bump as part of their unrelated PR, or
2. Manually revert the lock file changes to keep their PR focused

By proactively updating the `uv.lock` revision to 3, we eliminate this recurring prompt and allow contributors to work with dependencies without unexpected lock file changes.

PR #5708, opened 2026-02-02
Fix material settings not applying to GLTF models

### LINKED ISSUE #3515: Double-sided material property not applied to objects loaded from gltf
### Description

Hi all
I'm trying to display a garment 3D object, loaded as gltf. The material in my glb file is one-sided, so I tried to use the material settings from niceGUI to convert it to a double-sided material as follows:
```
# Camera setup
camera_location = [0, -200., 1.25]  
camera = ui.scene.perspective_camera(fov=np.pi / 6.)
camera.x = cam_location[0]
camera.y = cam_location[1]
camera.z = cam_location[2]
camera.look_at_x = 0
camera.look_at_y = 0
camera.look_at_z = cam_location[2] * 2/3

# Create scene
with ui.scene(
	width=1024,
	height=800, 
	camera=camera, 
	grid=False, 
	) as ui_3d_scene:
	ui_3d_scene.gltf(
		'geo/Configured_design_3D.glb', 
	).scale(0.01).rotate(np.pi / 2, 0., 0.).material(side='double')
```

However, the displayed object is still one-sided (the back of the garment visible through the neck and sleeve holes is not rendered): 
image 

Here is how I expect it to look like: 
image

Seems like currently, material setup in gltf object takes precedence over the additional material setup, but I would expect it to be the other way around, for the material properties that are explicitly set

Here is the glb file: [Uploading Configured_design_3D.zip…]()

evnchn:
### Motivation

Fix #3515, where GLTF models load asynchronously as Groups, so material settings were being ignored.

Particularly, there are 3 cases to resolve (from https://github.com/zauberzeug/nicegui/issues/3515#issuecomment-2325885092):

> 1. Calling `material()` on a group (or GLTF object) does not work, because the JavaScript implementation of the `material()` method doesn't traverse children of groups.
> 2. Calling `material()` on a group (or GLTF object) does not work, because the group is added first and their children are added later. So traversing the group doesn't help, because there are no children yet.
> 3. Calling `material()` immediately after creating the GLTF object does not work, because the GLTF loader loads the object(s) asynchronously. Only when this is finished, `material()` can actually find the corresponding objects.

**NOTE: right now the test sample I can only test the GLTF case, not the `group` case though.**

### Implementation

TL-DR: Now stores pending material info on Group objects and applies settings to all mesh children after the model loads.

**Material handling improvements:**

- Refactored the `material` method: Now, if the target object is a GLTF model that has not finished loading, material settings are stashed in `userData.pendingMaterialInfo` and applied to all child meshes (including `Material[]` arrays) once the model is loaded. For non-GLTF objects, material settings are applied directly as before.
- Consolidated the per-material update logic into a local `apply` closure inside `material()`, used for both direct meshes and the GLTF traversal path.

**GLTF loading enhancements:**

- Modified the GLTF loader callback to set a `userData.loaded` flag and re-enter `material()` with any stashed `pendingMaterialInfo`, so material updates requested before model load are not lost.

### Progress

- [x] Pytest has been extended.

### DIFF
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -316,7 +316,16 @@ export default {
         mesh = new THREE.Group();
         this.gltf_loader.load(
           url,
-          (gltf) => mesh.add(gltf.scene),
+          (gltf) => {
+            mesh.add(gltf.scene);
+            if (mesh.pendingMaterialInfo) {
+              const { color, opacity, side } = mesh.pendingMaterialInfo;
+              mesh.traverse((child) => {
+                if (child.isMesh && child.material) this.applyMaterialSettings(child.material, color, opacity, side);
+              });
+              delete mesh.pendingMaterialInfo;
+            }
+          },
           undefined,
           (error) => console.error(error),
         );
@@ -374,9 +383,7 @@ export default {
       if (!this.objects.has(object_id)) return;
       this.objects.get(object_id).name = name;
     },
-    material(object_id, color, opacity, side) {
-      if (!this.objects.has(object_id)) return;
-      const material = this.objects.get(object_id).material;
+    applyMaterialSettings(material, color, opacity, side) {
       if (!material) return;
       const vertexColors = color === null;
       material.color.set(vertexColors ? "#ffffff" : color);
@@ -386,6 +393,15 @@ export default {
       if (side == "front") material.side = THREE.FrontSide;
       else if (side == "back") material.side = THREE.BackSide;
       else material.side = THREE.DoubleSide;
+    },
+    material(object_id, color, opacity, side) {
+      const object = this.objects.get(object_id);
+      if (!object) return;
+      if (object.isGroup) {
+        object.pendingMaterialInfo = { color, opacity, side };
+      } else if (object.material) {
+        this.applyMaterialSettings(object.material, color, opacity, side);
+      }
     },
     move(object_id, x, y, z) {
       if (!this.objects.has(object_id)) return;
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -324,7 +324,7 @@ export default {
             if (mesh.userData.pendingMaterialInfo) {
               const { color, opacity, side } = mesh.userData.pendingMaterialInfo;
               delete mesh.userData.pendingMaterialInfo;
-              this.material(mesh.object_id, color, opacity, side);
+              this.material(id, color, opacity, side);
             }
           },
           undefined,
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -321,10 +321,8 @@ export default {
             mesh.add(gltf.scene);
             if (mesh.pendingMaterialInfo) {
               const { color, opacity, side } = mesh.pendingMaterialInfo;
-              mesh.traverse((child) => {
-                if (child.isMesh && child.material) this.applyMaterialSettings(child.material, color, opacity, side);
-              });
               delete mesh.pendingMaterialInfo;
+              this.material(mesh.object_id, color, opacity, side);
             }
           },
           undefined,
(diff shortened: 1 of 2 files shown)


evnchn:
This indeed does not fix the group case:

```py
    with ui_3d_scene.group().material('#00ffff', side='both'):
        ui_3d_scene.box(1, 1, 1)
```

Box still white...

evnchn:
Not sure if we really should fix the group case. I got a fix in local which does it, but then I don't see a way to override the material once an object is in a group, which sounds hardly like a good idea.

falkoschindler:
I just added MREs for each of the three problems: https://github.com/zauberzeug/nicegui/issues/3515#issuecomment-3841729822.

```py
# Problem 1: Group material is not applied to child objects
with ui.scene() as scene:
    with scene.group() as group:
        scene.sphere()
    group.material('red')

# Problem 2: Group is still empty when created with material
with ui.scene() as scene:
    with scene.group().material('red'):
        scene.sphere()

# Problem 3: GLTF loader is not ready when applying material
avocado = 'https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Assets/main/Models/Avocado/glTF-Binary/Avocado.glb'
with ui.scene() as scene:
    scene.gltf(avocado).material('red').scale(100)
```

At the moment, this PR only solves **problem 3**.

Solving **problem 2** could be considered a breaking change, because it would turn this blue sphere red:
```py
with scene.group().material('red'):
    scene.sphere().material('blue')
```
Instead we could agree that material is only applied to the currently existing children and you need to apply it later (like in the first example) if you want to color all children.

**Problem 1** could probably be easily solved by traversing the group when applying material.

Overall I'm not 100% sure about the desired behavior. If Three.js doesn't automatically apply group material to children, maybe we shouldn't try doing it ourselves. Maybe there's a way to initialize objects with _undefined_ material so that Three.js automatically applies their group's material (if any)?

falkoschindler:
@evnchn This PR came up as one of the few non-draft PRs that isn't on our maintainer board, so I revisited it today. After thinking through the three cases from #3515, here's where I landed and what I pushed to the branch.

## Position

I'd like to merge this as a narrow fix for **case 3 only** (the async GLTF case) and leave cases 1 and 2 unchanged — by design, not as a deferred TODO.

Reason: **Three.js itself doesn't propagate material from a parent group to its children.** Transforms propagate via the scene graph; material does not. Groups don't even have a `material` property. Any group-inheritance model would be something we invent and maintain ourselves, with no upstream help — and I don't think we want that commitment without a concrete driver (and no one's asked for it). `ui.scene.group().material()` should stay a no-op on children; users apply material to mesh objects directly, matching Three.js semantics.

## What I pushed to the branch

Two commits on top of yours:

1. **`cf388729` — Also apply GLTF material on post-load calls.** Your PR fixes the "call `.material()` immediately after `.gltf(url)`" half of case 3, but the "call `.material()` after the load has already finished" half was still broken (group, no `.material` property → silent no-op). I replaced the `isGroup` branch with a permanent `isGltf` marker and added post-load traversal, so both halves of case 3 work. This also avoids stashing dead `pendingMaterialInfo` on unrelated groups (`ui.scene.group()`, `spot_light`).

2. **`39051748` — Consolidate material application in `material()`.** Inlined `applyMaterialSettings` back into `material()` as a local closure, and routed the load callback through `this.material(...)` with the stashed args. Result: one place handles all material logic, GLTF-specific branching (stash while loading vs. traverse once loaded) is visible in a single function.

A thorough review is still due. But before polishing, does the reasoning above — case 3 only, cases 1 & 2 left as-is by design — match how you see it?

falkoschindler (approved):
Approving — went through this end-to-end and pushed four small follow-ups on top of your branch:

- **`8916d4f7` Use `userData` for GLTF flags and handle multi-material meshes** — moves `isGltf`, `loaded`, and `pendingMaterialInfo` under `mesh.userData` so we don't squat on Three.js-reserved property names, replaces the `children.length === 0` "still loading" heuristic with an explicit `userData.loaded` flag set inside the loader callback, and guards the traversal against `Material[]` (multi-material meshes are common in GLTF assets and would have crashed `material.color.set`).
- **`194aa29a` Fold array unwrap into `apply` and shorten traverse callback** — merges the temporary `applyTo` wrapper into `apply` itself so one closure handles both single materials and arrays, and collapses the traverse callback to an expression-arrow with short-circuit guards.
- **`e9652788` Use captured `id` in GLTF loader callback** — uses the closure-captured `id` instead of round-tripping through `mesh.object_id` in the recursive `this.material(...)` call.
- **`40ac7096` Verify GLTF material assignment in `test_gltf`** — extends the existing `test_gltf` with `.material('#ff0000')` right after creation and asserts the loaded mesh ends up red, locking in the async case from #3515 that this PR fixes.

Scope stays as described in https://github.com/zauberzeug/nicegui/pull/5708#issuecomment-4313191130: case 3 from #3515 only. Cases 1 & 2 (`ui.scene.group().material()` propagating to children) are intentionally left alone — Three.js itself doesn't propagate material through groups, and inventing that semantic would be a NiceGUI-only commitment with no concrete user driver.

If anything in those follow-ups looks off, happy to drop or rework. Otherwise I'll merge once CI's green.

PR #5710, opened 2026-02-03
Fix REPL detection for Python 3.13's new pyrepl

### LINKED ISSUE #5675: Hint "NiceGUI doesn't work in REPL" doesn't show up anymore
### First Check

### Example Code

```python
from nicegui import ui

ui.label('Lorem ipsum')

ui.run()
```

### Description

I was busted by NiceGUI script mode not being supported in REPL.

To notice that more quickly https://github.com/zauberzeug/nicegui/pull/5456 added a message making this obvious. However it seems it doesn't show up anymore.

To reproduce:
-  Open a Python REPL
- Type the provided example code

The result is this:
```
ERROR:    Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/starlette/routing.py", line 694, in lifespan
    async with self.lifespan_context(app) as maybe_state:
               ~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/contextlib.py", line 214, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/nicegui/nicegui.py", line 31, in _lifespan
    await _startup()
  File "/usr/local/lib/python3.13/site-packages/nicegui/nicegui.py", line 122, in _startup
    raise RuntimeError('\n\n'
    ...<5 lines>...
                       'to allow for multiprocessing.')
RuntimeError: 

   if __name__ == "__main__":
to allow for multiprocessing.

ERROR:    Application startup failed. Exiting.
```

I would expect that the message added in that PR appears.

Furthermore, the second suggestion from that message, using a root function, doesn't seem to help either and leads to the same result. To reproduce type this code into the REPL:
```python
from nicegui import ui

def root():
     ui.label('Lorem ipsum')

ui.run(root=root)
```

I would expect that NiceGUI starts up normally.

### NiceGUI Version

3.6.1

### Python Version

3.13.11

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

falkoschindler:
### Motivation

Fixes #5675

In Python 3.13, the new `_pyrepl` REPL sets `__main__.__file__` to an actual file path (`/path/to/python3.13/_pyrepl/__main__.py`), causing the REPL detection to fail. This results in uvicorn spawning a subprocess where `ui.run()` was never called, producing a confusing error instead of the helpful "NiceGUI doesn't work in REPL" hint.

### Implementation

Changed the REPL detection from checking `__main__.__file__` to using `sys.ps1` and `sys.flags.interactive`:

```python
is_repl = bool(getattr(sys, 'ps1', sys.flags.interactive))
```

- `sys.ps1` exists when Python is running interactively (REPL mode)
- `sys.flags.interactive` is set when Python is started with the `-i` flag

This approach works reliably across all Python versions, including 3.13.

Also fixed a typo in the warning message: "is is only" → "it is only".

### Progress

### DIFF
--- a/nicegui/ui_run.py
+++ b/nicegui/ui_run.py
@@ -12,8 +12,6 @@ from starlette.types import ASGIApp
 from uvicorn.main import STARTUP_FAILURE
 from uvicorn.supervisors import ChangeReload, Multiprocess
 
-import __main__
-
 from . import core, helpers
 from . import native as native_module
 from .air import Air
@@ -217,8 +215,9 @@ def run(root: Callable | None = None, *,
     if multiprocessing.current_process().name != 'MainProcess':
         return
 
-    if reload and not hasattr(__main__, '__file__'):
-        log.warning('disabling auto-reloading because is is only supported when running from a file')
+    is_repl = bool(getattr(sys, 'ps1', sys.flags.interactive))
+    if reload and is_repl:
+        log.warning('disabling auto-reloading because it is only supported when running from a file')
         core.app.config.reload = reload = False
 
     if kwargs.get('ssl_certfile') and kwargs.get('ssl_keyfile'):


KrilleGH:
> `python -c "from nicegui import ui; ui.label('hey'); ui.run()"`

Good catch, I may not have noticed it.

Just for curiousity, is it ever possible for `sys.argv` to be empty?

nicegui/ui_run.py
```diff
@@ -217,8 +215,9 @@ def run_script() -> None:
     if multiprocessing.current_process().name != 'MainProcess':
         return
 
-    if reload and not hasattr(__main__, '__file__'):
-        log.warning('disabling auto-reloading because is is only supported when running from a file')
+    is_repl = bool(getattr(sys, 'ps1', sys.flags.interactive))
```
KrilleGH, line 218: It might be even better to use `hasattr()`, for the absurd case where the interpreter is interactive and `sys.ps1` is defined but empty for some reason.

evnchn: Maybe we will switch when we see such a crazy interpreter

PR #5719, opened 2026-02-04
improve storage documentation clarity and fix inconsistencies

evnchn:
### Motivation

Addresses feedback from discussion #5717 where users reported significant confusion with storage documentation, including:
- Critical table inconsistency about tab storage persistence across restarts
- Unclear terminology ("needs client connection")
- Missing guidance for multi-instance deployments
- Undocumented storage behavior in native mode
- Confusion about when/how storage initialization occurs

### Implementation

**Fixed Critical Inconsistencies:**
- Corrected storage comparison table: tab storage "across server restarts" now shows "No*" with footnote explaining it's only "Yes" with Redis
- Added footnote explaining "needs client connection" means awaiting client.connected() before accessing storage

**Added Storage Initialization Documentation:**
- Documented when and how each storage type is initialized
- Explained multi-tab behavior and data sharing between tabs
- Documented tab duplication behavior (copies initially, then independent)

**Enhanced Native Mode Documentation:**
- Documented that automatic port scanning via find_open_port() is already implemented and used by default (scans ports 8000-8999)
- **Critical addition:** Documented storage path conflicts when running multiple instances of the same file from the same location
- Provided solutions: separate NICEGUI_STORAGE_PATH per instance or Redis

**Key Distinction Clarified:**
Port conflicts are automatically handled in native mode, but storage path conflicts require manual configuration.

### Progress

- [x] If this PR addresses a security issue, it has been coordinated via the security advisory process. (N/A)
- [x] Pytests have been added (or are not necessary). (Documentation only)
- [x] Documentation has been added (or is not necessary). (This IS the documentation)

### Final notes

This is also written by Claude. Rough readthrough by me.


evnchn:
Though the PR description may be a bit slack, I screened the content of this PR quite carefully. Should not have technical inaccuracies.

phifuh:
@evnchn 

A) Where is the storage located?
If I remember correctly, it’s a dictionary saved somewhere, but I’m wondering why the “tab” storage type can be persisted with Redis while the browser/user storage cannot.

B) In discussion thread #5719 I asked whether port scanning runs when native=False. Is it active in that configuration?

C) Even when we run an app in native mode, the app remains accessible in the browser.
Does this mean the browser storage for the user (the “browser” storage) and the technical “open‑tab” storage can retain persistent data across native‑window restarts? I’m not sure whether this behavior is intentional or caused by having the reload argument set to true.

D) It would be helpful to link directly to the "section_configuration_deployment.py" documentation whenever native/Redis is mentioned.
Perhaps a brief standalone subsection could be added there.

E) What does “Write only before response” mean exactly, and why does it apply only to the browser storage and not to the tab storage?

F) Are the tab storage and browser storage the same kind of dictionary?
If so, why doesn’t tab storage require serializable data?

G) A colleague pointed out that “Needs serializable data” and “Needs storage_secret” are implementation concerns, whereas the other options are more about choosing the appropriate storage type.
It feels like mixing concerns, but it might still make sense to keep them together in the same table.

falkoschindler:
@evnchn What's that status of this PR? Looks like there are many open questions. Should we postpone?

evnchn:

Revived this: merged current `main` in (conflict was with the new Windows-favicon note in the native-mode docs; both texts now coexist) and pre-commit is green locally, so the PR is reviewable again.

@phifuh — your three questions (storage location/Redis asymmetry, port scanning outside native mode, persistence across native-window restarts) are good ones, but they go beyond what this docs-clarity PR should grow to answer. Rather than expanding scope per specific issue faced, they seem better suited to a [discussion](https://github.com/zauberzeug/nicegui/discussions) where the answers can be worked out properly — and whatever's worth keeping can then flow into the docs in a follow-up.

@falkoschindler what do you think? If you agree, this PR can be reviewed on its current scope.

falkoschindler (approved):
Thanks @evnchn, this addresses real confusion from #5717 — especially the storage table, which wrongly claimed tab storage survives server restarts. I verified that against the code: tab storage has been an in-memory dictionary since its introduction in #2837 and only persists with Redis (since #4074), so the correction is spot on. The tab duplication behavior, the port scanning range and the storage path default all check out as well.

I pushed a few follow-up commits to polish the result:

- **Footnote markers:** Replaced the escaped asterisks with `<sup>1)</sup>`/`<sup>2)</sup>`. The `\*` sequences were invalid Python escapes (a `SyntaxWarning` on 3.12+, hidden by bytecode caching and our ruff exclusion of the docs folder), and numbered footnotes read nicer anyway.
- **Folded the new sections into the existing bullets:** The "How Storage Initialization Works" and "Multi-Tab Behavior" sections largely restated the storage type descriptions above them. The genuinely new facts now live in the `app.storage.tab` bullet: retention for up to `max_tab_storage_age` (30 days by default), persistence across restarts with Redis, and the copy-on-duplicate behavior. This also fixed the claim that tab storage only persists "as long as the tab remains open" — it is actually kept on the server up to the maximum age, which is why it survives page reloads.
- **Client connection footnote:** Clarified that awaiting `client.connected()` is only needed while building the page; event handlers already have an established connection.
- **Native mode:** Made the multi-instance warning more precise — the shared storage path is not specific to native mode (any instances started from the same working directory share it), and the actual failure mode is that instances don't see each other's data and silently overwrite each other's writes.
- **Port selection:** Added that browser mode defaults to port 8080 without scanning, and that `port=native.find_open_port()` can be passed explicitly to run multiple instances side by side — answering @phifuh's follow-up question in #5717.

Merging once CI is green. Thanks again!

PR #5720, opened 2026-02-04
Symlink CLAUDE.md to AGENTS.md

evnchn:
### Motivation

I find it surprising that the code Claude Opus 4.5 produce using Claude Code have so many comments. Isn't it state-of-the-art coding model, and which @falkoschindler uses as well? 

After auditing its thinking process, aparently it never read CONTRIBUTING.md, and not even AGENTS.md for that matter. No wonder the poor output. 

### Implementation

I symlinked it: `ln -s AGENTS.md CLAUDE.md`

### Progress

- [x] If this PR addresses a security issue, it has been coordinated via the [security advisory](hxttps://github.com/zauberzeug/nicegui/security/advisories/new) process.

### Final note

May waste some context window, but it's for the better.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


evnchn:
image

falkoschindler:
To be honest, I've been aware of the missing CLAUDE.md. When I changed from Cursor to VS Code + Claude Code on Monday, I was wondering about my low token usage. Current working theory: AGENTS.md is eating quite a lot of tokens while I didn't notice much difference without it. But I haven't done large refactorings this week, so maybe my impression would be different with different tasks.

Anyway, it's a shame that Claude Code doesn't read the AGENTS.md which is supposed to be an "open format for guiding coding agents, used by over 60k open-source projects." Instead of adding dot files for every tool developers might use, maybe we can configure it in our local Claude settings?

In the end it might be simpler to just merge this PR though. 🤷🏻‍♂️

However, we should check if CONTRIBUTING.ms (linked from AGENTS.md) can be condensed. E.g. we don't need to mention every formatting rule if Ruff handles it automatically on save.

rodja:
Yes we definitely should merge this. In my experience Claude Code (as any LLM) greatly improves with proper context. The AGENTS.md might benefit from a good sweep of token reduction with the `claude-md-management` plugin. But that can be done in a separate PR (which also condenses README.md and CONTRIBUTING.md).

evnchn:
> maybe we can configure it in our local Claude settings

There is no option that I know of, to have Claude Code natively read anything other than CLAUDE.md on conversation init

Maybe we can use some hooks, but it may not be the same functionality as the native one. 

It's unfortunate that there isn't such an option, but for now I think symlinking is what we can do.

falkoschindler:
> Yes we definitely should merge this.

Sorry, but I'm not so sure. I mean, you can outvote me and I can live with it. But here are my thoughts:

- Symlinks can be problematic on Windows. Maybe @evnchn can check if it works nonetheless, but they might only work on macOS and Linux.
- For maximum compatibility it might be better to use a single-line instruction like "See AGENTS.md for guidelines".
- Even though I marked this PR for 3.7, it is not critical and can be merged at any time. No need to rush.
- To get a working CLAUDE.md right now, everyone of us can add a global CLAUDE.md with above-mentioned instruction to read the AGENTS.md.
- @rodja Are you sure claude-md-management is for shortening md files? I heard its primarily for adding information.
- Personally I didn't miss AGENTS.md the last few days, so I'd rather revisit its content first before simply symlinking the old one.

rodja:
> For maximum compatibility it might be better to use a single-line instruction like "See AGENTS.md for guidelines".

I do not think that would work. `CLAUDE.md` is preloaded and then you can start your prompting. The context window will not be filled with instructions from something referenced there.

> Symlinks can be problematic on Windows.

Ok, that is true. Maybe a git hook that just runs `cp AGENTS.md CLAUDE.md` with `CLAUDE.md` to `.gitignore`, and keep `AGENTS.md` be the committed source of truth.

> Personally I didn't miss AGENTS.md the last few days

That is fascinating and suggests we must improve `AGENTS.md`. Quite a lot of Claude Code documentation and usage reports, my own experience, and Evans motivation for this PR suggests that it has tremendous impact.

evnchn:
Git symlinks breaks hard on Windows :pensive::

- Doesn't work out of the box
- Tried `git config core.symlinks true`
- Then `error: unable to create symlink CLAUDE.md: Permission denied`

evnchn:
Windows require Admin priviledges for symlinks 🤡 

I think @rodja's "git hook that just runs cp AGENTS.md CLAUDE.md with CLAUDE.md to .gitignore, and keep AGENTS.md be the committed source of truth" is good.

evnchn:
Hmm, but 2 issues:

1. Beginners in NiceGUI development miss installing `pre-commit` and may miss out on `CLAUDE.md` entirely
2. If the source-of-truth is `AGENTS.md`, then it becomes cumbersome to commit once edit the memory using Claude Code UI

Is it good enough if we add check `CLAUDE.md == AGENTS.md` in precommit and CI? This way the onboarding friction is the lowest.

PR #5722, opened 2026-02-05
Add CLAUDE.md with pre-commit sync enforcement

evnchn:
### Motivation

Claude Code (Anthropic's official CLI tool) uses `CLAUDE.md` for project-specific instructions, similar to how other AI tools use `AGENTS.md`. This PR introduces `CLAUDE.md` as a copy of `AGENTS.md` and adds a pre-commit hook to ensure they stay in sync.

### Implementation

- Added `CLAUDE.md` with identical content to `AGENTS.md`
- Added a local pre-commit hook (`sync-agent-files`) that uses Python to compare the files
- The hook only triggers when either file is modified and fails if they differ
- CI already enforces pre-commit via `_check.yml`, so this is automatically enforced
- Uses Python for cross-platform compatibility (works on both Unix and Windows)

### Progress

### Human notes

Above was written by Claude. Kinda obvious for how non-negative the motivation statement is. 

How is "Claude Code use `CLAUDE.md` similar to how other AI tools use `AGENTS.md`"? The fact that the file name is not the same mean it is NOT similar. 

Anyways this works on Windows, over the predecessor #5720

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,3 +36,11 @@ repos:
           )$
         additional_dependencies:
           - tomli; python_version<'3.11'
+  - repo: local
+    hooks:
+      - id: sync-agent-files
+        name: Ensure AGENTS.md and CLAUDE.md are identical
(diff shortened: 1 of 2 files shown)


evnchn:
I like #5730

PR #5727, opened 2026-02-06
Add ui.aggrid.VERSION constant for AG Grid version reference

evnchn:
### Motivation

In #5726, it was pointed out that it is best to read off a constant instead of hardcode the version. 

### Implementation

Asked Copilot for the pull request, but [its commit history is poor](https://github.com/evnchn/nicegui/tree/copilot/add-ag-grid-version-property) and does not create good PR according to template, so ask Claude to clean things up. 

As in commit message: 

> Add a VERSION class attribute to AgGrid that exposes the AG Grid version (34.2.0) used by NiceGUI. This allows users to programmatically reference the version when building custom CDN URLs for enterprise features, ensuring compatibility between the core library and enterprise modules.

This enables:

```py
from nicegui import ui

bundle_url = f'https://cdn.jsdelivr.net/npm/ag-grid-enterprise@{ui.aggrid.VERSION}/+esm'
ui.aggrid.set_module_source(bundle_url)
```

### Progress

### DIFF
--- a/website/documentation/content/aggrid_documentation.py
+++ b/website/documentation/content/aggrid_documentation.py
@@ -292,7 +292,7 @@ def aggrid_handle_theme_change():
     (either from a CDN or from a self-hosted bundle)
     and passing `modules='enterprise'` to the `ui.aggrid` constructor.
 
-    Use `ui.aggrid.VERSION` to programmatically reference the AG Grid version used by NiceGUI.
+    Use `ui.aggrid.VERSION` (*since version 3.8.0*) to programmatically reference the AG Grid version used by NiceGUI.
     This ensures compatibility when building custom CDN URLs.
 
     *Added in version 3.6.0*


falkoschindler:
Thanks, @evnchn!
I wonder
- if we should introduce a version attribute for _all_ UI elements with third-party dependencies, and
- if we could automatically extract them to reduce the risk for regressions.

Maybe we can at least add a pytest that compares `ui.aggrid.VERSION` with a version obtained via JavaScript?

evnchn:
Adding version attribute for all UI elements is a good-to-have but I don't firmly see the need because _only_ `ui.aggrid` comes with `.set_module_source` method which runs `register_importmap_override`. 

As `register_importmap_override` is private API (not in `app` not `ui`) we could argue it is a "your mileage may vary". 

I mostly disagree with automatic extraction because I feel it can get nasty real quick. Either:

- `VERSION = _some_complex_function_which_extracts_it_from_JS_in_runtime()`, compromise IDE value visibilty. 
- A script which greps for the `VERSION` line and modifys Python source code, which is easy to get wrong. 

Anti-regression test is good though.

Issue #5185, opened 2025-09-29
Character loss in ui.input validation with high network latency


Xitod:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

def simple_validation(value: str) -> str | None:
    """Simple validation that requires minimum length and specific characters"""
    if len(value) < 3:
        return "Must be at least 3 characters long"
    if not value.isalnum():
        return "Must contain only letters and numbers"
    return None

# Minimal reproduction case
test_input = ui.input(
    label="Test Input",
    validation=simple_validation,
)

...
```

### Description

When typing quickly in `ui.input` fields with validation functions, characters are being deleted/lost during the validation process. This only occurs with network latency - the same application works perfectly when running locally. I live in southern hemisphere so I can say that I have an estimated latency of ~200 to ~400ms. 

### Observation

- Works fine locally: No character loss when running the same code locally
- Geographic dependency: Colleagues near the AZ don’t experience this issue
- Reproducible: Happens consistently when typing at normal speed with validation enabled

This appears to be the same underlying problem described in #[287](https://github.com/zauberzeug/nicegui/issues/287), specifically related to this comment:
		“All input elements involving typing (ui.input, ui.number, and ui.color_input) show flickering and lost characters. … So we probably need to set their state sometime, maybe on change but with some throttling.”
The issue seems to be the roundtrip validation process where each keystroke triggers a client→server→client cycle for validation, and with high latency, this causes desynchronization between user input and the field state.

### Expected Behavior
Input validation should work smoothly regardless of network latency, possibly by implementing automatic throttling for validation functions or debouncing validation calls

### NiceGUI Version

2.24.1

### Python Version

3.12.11

### Browser

Chrome

### Operating System

macOS

falkoschindler:
Thanks for reporting this issue, @Xitod!

In #287 the problem occurred even without sanitization, which has been fixed in version 1.1.5.

But now the problem is that an updated error message is sent to the client together with the whole state dictionary of the input element, including an outdated `value`. Here is a minimum reproduction:

```py
ui.input(validation=lambda x: f'Still {1000 - len(x)} characters missing')
```

[...]

PR #5728, opened 2026-02-06
Implement differential updates in outbox to reduce bandwidth

evnchn:
### Motivation

The issue in #5185 is that the `value` must be retransmitted from the Outbox since we transmit the entire element definition when something changed. 

As such, from the moment the server receives the character leading to a change in error message, to the moment the client received the outbox message, all character typed within that window is gone. 

- Client: aaa
- Client: aaaa
- Server: OK, hide the error dialog, value is aaa
- Client: aaa (note how aaaa is lost)

### Implementation

- Track last-sent state per element to compute diffs
- Send only changed properties instead of full element state
- Add client-side deep merge for incoming diffs
- Optimize LOOPBACK=False elements to avoid redundant updates

### Progress

### Final notes

It _could_ be possible to minimize the PR to just target the `value` key, but I thought (1) why not also push for a performance increase, (2) we can slim things down later.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -446,7 +446,7 @@ class Element(Visibility):
         """Return diff from last sent state and update _last_sent.
 
         Calls _to_dict() exactly once. Use for differential updates.
-        Returns None if nothing changed.
+        Returns None if nothing changed (unless element has update_method).
         """
         current = self._to_dict()
         previous = self._last_sent
@@ -454,7 +454,10 @@ class Element(Visibility):
 
         if previous is None:
             return current
-        return _compute_diff(current, previous)
+        diff = _compute_diff(current, previous)
+        if diff is None and self._update_method:
+            return {}  # ensure element is included so Vue update method gets called
+        return diff
 
     def run_method(self, name: str, *args: Any, timeout: float = 1) -> AwaitableResponse:
         """Run a method on the client side.


falkoschindler (changes requested):

This is a pretty complicated undertaking. Here's what Claude has to say:

<details>
<summary>
Full review
</summary>

**BLOCKER**

1. **Performance regression on hot path (deepcopy):** `_to_dict_diff` (element.py:453) calls `deepcopy(current)` on every element on every outbox tick. For elements with large prop dicts (e.g., `ag_grid` with thousands of rows, `table`, `plotly`), this could be significantly more expensive than the bandwidth it saves. Has this been benchmarked on heavy elements?

2. **Missing tests:** No test files are changed. `_compute_diff` and `_to_dict_diff` are new core logic governing what gets sent to every client — at minimum they need unit tests covering: identical dicts → `None`, nested changes, key removal, and the `_update_method` empty-dict path. The `_set_client_prop` LOOPBACK optimization also needs coverage.

3. **Reconnection correctness:** `try_rewind` replays messages from `message_history`, which now contain diffs instead of full state. If a client does a socket-only reconnect (no page reload), replaying partial diffs on top of potentially stale client state could produce inconsistencies. Is this scenario handled, or does reconnect always go through `build_response` → `_to_dict_full`?

**MAJOR**

1. **Falsy-key disappearance:** `_to_dict` (line 253) omits keys with falsy values (`if value`). If `classes` goes from `"foo"` to `""`, the key disappears from the dict. `_compute_diff` then produces `classes: None`, and the JS `deepMergeElement` _deletes_ the key rather than setting it to empty. This could cause stale classes/styles/props to persist on the client instead of being cleared.

2. **Loose equality in LOOPBACK check** (value_element.py:45): `str(model_value) == str(e.args)` will match values that aren't semantically equal (e.g., `"1"` vs `1`, `"None"` vs `None`). Could this cause `_set_client_prop` to incorrectly suppress a needed update?

**CLEANUP**

1. **`replaceUndefinedAttributes` semantic change** (nicegui.js:103-105): Changed from `element.children || []` to `element.children ??= []` and reordered slot construction so `default` always wins. This is more correct but is a subtle behavior change bundled into the diff PR — worth a comment or separate commit.

2. **Consider lighter copy:** For JSON-serializable dicts, `json.loads(json.dumps(x))` is often faster than `deepcopy`. Or copy only when the diff is non-None (i.e., skip the copy when nothing changed and we're about to return `None` anyway).
</details>

Given these complications, could we scope this down to just the value problem? The full diff machinery (`deepcopy` per element, `_compute_diff`, `deepMergeElement`, null-means-delete) adds a lot of complexity and performance risk for what's essentially a value loopback issue.

What about a simpler approach: Can we use a **boolean flag** on `ValueElement` to track whether the current value came from the client or the server? When building the outbox update, omit `VALUE_PROP` from props if it originated from the client. The client keeps its local value because the key simply isn't in the update. No copies, no diffs, no new merge semantics — just one bool per input element.

evnchn:
> what's essentially a value loopback issue

Maybe I'm always stuck with a performance frame-of-mind. I thought it was the inefficient outbox holding back the value element, not the value element using the outbox wrong. 

Honestly we could pursue the latter, and the PR would be much more simple. Though we're absorbing the performance gains, which, depending on the device and the website size, could be OK since the browser/OS is more overhead. 

@falkoschindler Do you want to take a shot before my usage limit comes back tomorrow?

falkoschindler:
@evnchn If we, as I proposed, don't resend values that came from the client, we do improve performance. For many elements this is the main part of the payload. I'm not ruling out another performance improvement by dropping other unchanged attributes. But at the moment I'm worried about the cost of storing, comparing and merging _everything_.

I probably won't have time to implement it myself. Since we have still plenty of open tickets for 3.9 and already a full milestone 3.10, I postponed it to 3.x. We can always squeeze it in earlier, but I have to triage a bit.

evnchn:
Superseded by #5933, which implements the simpler bool-flag approach proposed by @falkoschindler in his review.

The full differential outbox optimization (deepcopy + diff + merge) from this PR is preserved as a future option — if we want to go beyond just VALUE_PROP and drop other unchanged attributes from outbox updates, the diff machinery concept from this PR could be revisited. Tracking that idea separately.

PR #5729, opened 2026-02-06
Reuse Chrome driver across screen tests for faster execution

evnchn:
### Motivation

Starting a new Chrome browser for each test using the `Screen` fixture is slow. By reusing the same browser session across all tests, we can significantly reduce test execution time.

### Implementation

- Made `nicegui_driver` and `nicegui_chrome_options` fixtures session-scoped so the browser is started once and reused
- Added `_reset_browser_state()` helper to clean up browser state between tests (closes extra tabs, navigates to about:blank, clears browser logs)
- Updated `Screen.close()` to navigate to about:blank instead of closing the last tab, which would invalidate the session-scoped driver
- Consolidated `capabilities` fixture into `nicegui_chrome_options` (moved logging prefs to `set_capability()`)

### Progress

### DIFF
--- a/nicegui/testing/screen.py
+++ b/nicegui/testing/screen.py
@@ -97,6 +97,7 @@ class Screen:
                 assert self.server_thread is not None
                 if not self.server_thread.is_alive():
                     raise RuntimeError('The NiceGUI server has stopped running') from e
+        self._wait_for_socket_idle()
 
     def close(self) -> None:
         """Close the browser tab.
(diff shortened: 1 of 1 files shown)


evnchn:
I can't tell if the tests are flaky already on main or only with this PR. Also MacBook Air the CPU performance fluctuates. Let's pray for the pipeline 🙏

nicegui/testing/screen_plugin.py
```diff
@@ -45,28 +46,22 @@ def nicegui_chrome_options(chrome_options: webdriver.ChromeOptions) -> webdriver
...
     return chrome_options
 
 
-@pytest.fixture
-def capabilities(capabilities: dict) -> dict:
-    """Configure the Chrome driver capabilities."""
-    capabilities['goog:loggingPrefs'] = {'browser': 'ALL'}
-    return capabilities
-
-
```
evnchn, line 59: Appears never used in the first place?

evnchn:
17m 48s only, normally is 22m 35s. 

So we shabe 5 minutes circa. 

1 failing test: will see what to do with it.

nicegui/testing/screen_plugin.py
rodja: I think it was being used via pytest-selenium's selenium fixture until 64990112. After the switch to a custom driver, it became a leftover.

evnchn:
I think LLM really hates the idea of moving helper functions below their usage. Inspecting the thinking process, it's like "it can't possibly work"

Not sure but maybe let's reverse it one day.

PR #5732, opened 2026-02-07
Fix race condition: use static DOMPurify import to avoid mid-module yield

evnchn:
### Motivation

Fixes race condition error introduced by https://github.com/zauberzeug/nicegui/commit/f1f7533577875af7d23f161ed3627f73584cb561 on https://nicegui.io where `emitEvent` is called before app initialization completes:

```
Uncaught TypeError: Cannot read properties of undefined (reading '$refs')
    at getElement (nicegui.js:111:22)
    at emitEvent (nicegui.js:155:3)
    at (index):308:51
```

https://github.com/zauberzeug/nicegui/blob/2941a963ed4600690d45664d6eb19009a1e1324f/main.py#L74-L80

This occurs when `window.addEventListener('load', ...)` fires before `createApp` finishes, caused by async DOMPurify import delaying app initialization.

### Implementation

- Store browser feature detection result in cookie on first load
- Use cookie to conditionally render DOMPurify import server-side
- Browsers with native `setHTML` skip import entirely → synchronous, immediate app initialization
- Self-healing: if cookie assumption is wrong, clear and reload once

### Progress

- [x] If this PR addresses a security issue, it has been coordinated via the security advisory process.

### DIFF
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -31,8 +31,7 @@
           display: inline-flex;  /* HACK: re-apply Quasar's inline styles to override Tailwind */
         }
         .nicegui-dialog-open,
-        .nicegui-select-popup-open,
-        .nicegui-table-fullscreen {
+        .nicegui-select-popup-open {
           scroll-behavior: auto !important; /* HACK: avoid smooth-scrolling when dialogs or select popups close */
         }
       }
@@ -114,6 +113,7 @@
       <span>{{ translations.message_too_long_body }}</span>
     </div>
     <script type="module">
+      // Load DOMPurify for HTML sanitization, polyfill for browsers without native setHTML
       import DOMPurify from "dompurify";
       if (typeof Element.prototype.setHTML !== "function") {
         Element.prototype.setHTML = function (html) {


evnchn:
@falkoschindler Since this messes up the public API, may I have a medium or a major and review ASAP? Thanks.

evnchn:
A more hotfix-style approach is in https://github.com/evnchn/nicegui/tree/early-emitevent-fix, but not sure how useful it is, since besides `emitEvent` there's also other variables accessible in JS which broke.

rodja:
I'll check it soon

rodja:
Thanks for the thorough analysis of the race condition — the diagnosis is spot on. The `await import("dompurify")` creates a yield point that lets the `load` event fire before `createApp` completes.

I think the same fix can be achieved with a much simpler change (2 lines, no cookies):

**1. `nicegui/templates/index.html`** — static import instead of dynamic await:
```diff
-      if (typeof Element.prototype.setHTML !== "function") {
-        const { default: DOMPurify } = await import("dompurify");
-        Element.prototype.setHTML = function (html) {
-          this.innerHTML = DOMPurify.sanitize(html);
-        };
-      }
+      import DOMPurify from "dompurify";
+      if (typeof Element.prototype.setHTML !== "function") {
+        Element.prototype.setHTML = function (html) {
+          this.innerHTML = DOMPurify.sanitize(html);
+        };
+      }
```

**2. `nicegui/static/nicegui.js`** — guard against pre-mount calls:
```diff
 function emitEvent(event_name, ...args) {
+  if (!mounted_app) return;
   getElement(0).$emit(event_name, ...args);
 }
```

The static `import` is the actual fix: it eliminates the mid-execution yield point so `createApp` → `app.mount()` completes before `load` fires. The `emitEvent` guard is defense-in-depth for any future race conditions.

I verified both changes locally — with the static import, `mounted_app` is already defined when `load` fires, and the `is_desktop` event is delivered successfully (no dropped events).

The cookie-based DOMPurify skip could be a nice follow-up optimization once `setHTML` has broader adoption (currently ~0.02% on caniuse).

evnchn:
@rodja I think if you statically load DOMPurify (`import DOMPurify from "dompurify";`) then you don't even need the guard `if (!mounted_app) return;`.

Moreover, the public JS API surface of NiceGUI is quite wide. While certainly we can guard the `emitEvent` there's nothing to stop people from expecting `mounted_app` and `app` to be available on initial JS load. 

image

Hence if you would like to proceed to cut things down, you only need item (1) in your suggestion.

evnchn:
Sidenote: 

If we aren't doing feature detection, then it means we either keep it or remove it. 

If that's the case, since DOMPurify is used for SVG sanitization, I guess DOMPurify will be kept essentially forever (or until browsers come up with a setSVG / we find a lighter-weight SVG-sanitization-only library)

PR #5738, opened 2026-02-08
Add nested dictionary binding support

evnchn:
### Motivation

Accomplishes https://github.com/orgs/zauberzeug/projects/6?pane=issue&itemId=82941257 (https://github.com/zauberzeug/nicegui/discussions/2978)

Binding UI elements to nested dictionary keys currently requires manually managing intermediate
structures. This PR adds first-class support for nested paths using tuple syntax, e.g.
`ui.input().bind_value(data, ('profile', 'username'))`, making it easy to bind directly into
nested dicts/objects without boilerplate.

### Implementation

- New `PropertyName = str | tuple[str, ...]` type alias allows property names to be either
  a single string (backward compatible) or a tuple of strings for nested access
- Helper module `binding_nested.py` provides `_normalize_name`, `_path_contains_dict` to
  keep the core module clean
- `_get_attribute` / `_set_attribute` / `_has_attribute` now walk nested paths, with
  `_set_attribute` auto-creating intermediate dicts as needed
- All mixin `bind_*` methods (`bind_value`, `bind_text`, `bind_visibility`, `bind_enabled`,
  `bind_icon`, `bind_content`) accept the new tuple syntax
- Strict mode error messages distinguish "key" from "nested key" for clarity
- Internal binding keys normalized to `tuple[str, ...]` for consistent lookups

### Progress

- [x] Documentation has been added (docstring update sufficient).
- [x] But do we need a demo, still?

### DIFF
--- a/website/documentation/content/section_binding_properties.py
+++ b/website/documentation/content/section_binding_properties.py
@@ -64,6 +64,21 @@ def bind_dictionary():
     ui.button('Turn 18', on_click=lambda: data.update(age=18))
 
 
+@doc.demo('Bind to nested properties', '''
+    The binding methods also accept a tuple of strings for nested keys.
+    This allows you to bind to nested properties of complex data structures, like dictionaries of dictionaries or nested dataclasses.
+
+    *Added in version 3.10.0*
+''')
+def bind_nested():
+    data = {'user': {'name': 'Bob', 'age': 17}}
+
+    ui.label().bind_text_from(data, ('user', 'name'), backward=lambda n: f'Name: {n}')
+    ui.label().bind_text_from(data, ('user', 'age'), backward=lambda a: f'Age: {a}')
+
+    ui.button('Turn 18', on_click=lambda: data['user'].update(age=18))
+
+
 @doc.demo('Bind to variable', '''
     Here we are binding the value from the datepicker to a bare variable.
     Therefore we use the dictionary `globals()` which contains all global variables.


evnchn:
IMO we are one demo short of greatness on this one

evnchn:
The bulk comes from the tests, which while is complete (covers all the code paths), is a bit excessive. Though I don't see a simplier codebase though if we want to make the API `bind_value(data, ('profile', 'username'))` work. 

We could `init_binding(data, ('profile', 'username'))` and use the existing `bind_value(data.profile, 'username')`, though that would lead to the need of type ignore likely, since `data.profile` does not exist. The nested binding machinary must be baked inside, and I don't see a way around it.

evnchn:
@falkoschindler I have done a bit. Not 100% effort because tired and also you mentioned "Given these complications I'm not certain if we're heading in the right direction", so maybe this will be a chance for us to answer that question before we move on. My Claude Code will soon comment soon, because it did not agree with all your Claude Code's points.

evnchn:
## Independent review (from @evnchn's Claude Code)

I ran an independent review of this PR to get a second opinion on the points raised. Here's where I agree, disagree, or think the severity is off:

### On the blockers

**BLOCKER 1 (auto-creating intermediate attrs via `setattr` on non-dict objects)** — **Agree, this is the biggest design issue.** Silently mutating user objects by injecting dict attributes they never defined is surprising for a binding system. Fix applied: `_set_attribute` now raises `AttributeError` for missing intermediates on non-dict objects; only dict intermediates are auto-created.

**BLOCKER 2 (`type(current)()` fallback)** — **Agree on the fix, but severity is overstated.** In practice the common case is plain `dict` where `type(current)()` == `dict()` == `{}`. The risk only manifests with custom dict subclasses taking required constructor args — a narrow edge case. That said, the fix is trivial (just use `{}`), so it's been applied.

**BLOCKER 3 (`_has_attribute` swallows `TypeError`)** — **Agree.** `None['key']` raises `TypeError` which was being silently swallowed, making debugging very hard. Fixed: only `KeyError` and `AttributeError` are caught now. Same fix applied to `_path_contains_dict`.

### On the majors

**MAJOR 1 (redundant `_normalize_name` in hot path)** — **Agree, this is real and measurable.** `active_links` already stores normalized tuples, yet every `_get_attribute` / `_set_attribute` / `_has_attribute` call was re-normalizing. Fixed: internal functions now accept `tuple[str, ...]` directly, normalization happens once at the public API boundary (`bind_to`, `bind_from`, `bind`).

**MAJOR 2 (`_path_contains_dict` changes strict-mode defaults)** — **Overstated.** For single-key paths the behavior is identical to before (`isinstance(obj, Mapping)` root check). The semantics only change for the *new* tuple paths, which have no pre-existing behavior to break. This is correct design — you want lazy binding when traversing through dicts.

**MAJOR 3 (`PropertyName` not exported)** — **Agree.** Fixed: now importable from `nicegui`.

### Additional issues found (not in the original review)

1. **Double traversal pattern** — `_has_attribute` + `_get_attribute` was doing the full path traversal twice per check. Introduced `_try_get_attribute` with a `_MISSING` sentinel to combine both operations into one traversal.

2. **Tests use `Screen` fixture when `User` would suffice** — Per CONTRIBUTING.md, `User` should be preferred when no JavaScript/DOM interaction is needed. Converted 14 of 16 new tests to `User` fixture (only visibility and enabled tests remain on `Screen` since they check CSS classes/DOM attributes). This makes the test suite faster and less flaky.

### On the overall direction

The core idea is sound. Tuple syntax is the right API choice — it's explicit, type-safe, and avoids ambiguity with dot-containing keys. The +457/-88 stat is misleading: mixin changes are mechanical signature updates, the real logic is ~90 lines, and tests are proportional. Nested dict binding is a genuine pain point (discussion #2978).

With the fixes applied (5 commits on the branch), the hot-path overhead for existing single-key bindings is essentially zero, and the safety concerns have been addressed.

evnchn:
Hi — this is evnchn's Claude Code (Opus 4.6). Evan asked me to do an independent review of this PR before addressing the feedback, specifically to hear from both sides and avoid bias.

**TL;DR:** All blockers, majors, and cleanups are addressed across 5 commits. The net diff is +162/−162 (zero line count change). We disagreed on two points: BLOCKER 2 was overstated, and MAJOR 2 is not a real issue for existing bindings. We also found two issues the original review missed (double traversal on hot path, tests using `Screen` instead of `User`).

<details>
<summary>Full independent assessment and what changed</summary>

## Independent assessment of the original review

I agreed with most of the review, but had some differing takes:

**BLOCKER 2 (`type(current)()` fallback) — overstated.** In practice the most common case is plain `dict`, where `type(current)()` is identical to `{}`. The risk is real but narrow (custom dict subclasses with required constructor args). I'd call this a strong MAJOR rather than a BLOCKER. That said, the fix is trivial (just use `{}` always), so we did it anyway.

**MAJOR 2 (`_path_contains_dict` changes strict-mode semantics) — not a real issue.** The concern was that this changes behavior for existing single-key bindings. I traced through the logic: for a single-key name, `_path_contains_dict` with `len(name) == 1` returns `False` unless the root object is a `Mapping` — which is exactly what the old `isinstance(other_obj, dict)` check did. The semantics only differ for the *new* tuple paths, which have no pre-existing behavior to break.

**Things the original review missed:**
- `_has_attribute` + `_get_attribute` were called back-to-back in `_refresh_step` and `_propagate_recursively`, doing a full path traversal *twice*. With nested paths this doubles the cost.
- All new tests used the `Screen` fixture (browser-based) when `User` fixture would suffice per CONTRIBUTING.md guidelines — slower and more flaky.

## What's been addressed (5 commits)

| Commit | What it fixes |
|--------|--------------|
| `25f6960` Fix safety issues in nested binding attribute access | BLOCKER 1 (raises `AttributeError` instead of silent `setattr`), BLOCKER 2 (always `{}`), BLOCKER 3 (`_try_get_attribute` catches only `KeyError`/`AttributeError`) |
| `4a4adbf` Optimize binding hot path: normalize once, traverse once | MAJOR 1 (normalize at entry points only; internal functions take `tuple[str, ...]`), plus the missed double-traversal issue (new `_try_get_attribute` with `_MISSING` sentinel) |
| `e433305` Use dot notation in display names and standardize docstrings | CLEANUP 3 (`->` → `.`), CLEANUP 4 (consistent docstring wording), CLEANUP 1 (PEP 8 blank lines) |
| `eb6e252` Export PropertyName type from nicegui package | MAJOR 3 |
| `3bbd55b` Convert nested binding tests from Screen to User fixture | CLEANUP 2 (no more `screen.wait()`), plus the missed Screen→User issue |

</details>

## What's still open

- **Demo** — acknowledged in the PR checklist, not yet done.
- **"Is this the right direction?"** — this is the big question. The tuple API is explicit and type-safe (no ambiguity with dots in key names). The hot path overhead for non-nested bindings is now essentially zero: internal functions work exclusively with `tuple[str, ...]`, no normalization in the loop. Happy to discuss further.

evnchn:
Added one more commit (`5549ade`) — uses `MutableMapping` instead of `dict` in `_set_attribute`, so nested binding writes work with `OrderedDict`, `defaultdict`, and any custom `MutableMapping` subclass.

<details>
<summary>Details</summary>

`_get_attribute` already used `Mapping` (read-only ABC) for traversal, but `_set_attribute` was checking `isinstance(obj, dict)` — meaning non-`dict` mappings would fall through to the `setattr`/`getattr` branch and fail. Now `_set_attribute` uses `MutableMapping` (the writable counterpart), giving a clean `Mapping`-for-reads / `MutableMapping`-for-writes symmetry.

The change is 3 lines: one import, two `isinstance` checks.

</details>

evnchn:
@falkoschindler I think this is good to go, then. The last item is whether we need a live demo in the NiceGUI docs site, but other than that it's rock solid.

PR #5741, opened 2026-02-08
Add GitHub CLI and pre-commit hooks to devcontainer

evnchn:
### Motivation

The devcontainer currently lacks GitHub CLI (`gh`) integration and pre-commit hook setup, making it harder for contributors to interact with GitHub (e.g., creating PRs, reviewing issues) and to catch linting/formatting issues before pushing.

### Implementation

- **Dockerfile**: Added `gh` (GitHub CLI) to the installed packages.
- **devcontainer.json**:
  - Added a persistent named volume (`gh-config`) mounted at `~/.config/gh` so that `gh auth` tokens survive container rebuilds.
  - Extended `postCreateCommand` to fix volume ownership, run `pre-commit install`, and prompt for `gh auth login` if not already authenticated.
- **CONTRIBUTING.md**: Updated the devcontainer setup instructions to note the GitHub authentication prompt on first launch.

### Progress

### DIFF
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -9,4 +9,4 @@ uv sync
 uv run pre-commit install
 
 # Authenticate with GitHub CLI if not already logged in
-gh auth status >/dev/null 2>&1 || gh auth login -h github.com -p https -w
+gh auth status >/dev/null 2>&1 || gh auth login -h github.com -p https -w || true


evnchn:
contributors = Claude instances, if it isn't obvious

I still haven't wrapped my head around the Git subtree thing, and my Macbook Air has 32GB RAM 🤷

evnchn:
Hello @falkoschindler 

1. I tried `gh auth login -h github.com -p https -w` and **many other combo myself** and I never got it to NOT show `? Authenticate Git with your GitHub credentials? (Y/n)`

image

If you have one, please point out, but do test it thx

2. Intentional. When user messes with the `pyproject.toml`, `uv.lock`, or `.pre-commit-config.yaml`, the instinctive (though brutal) solution is to rebuild the Devcontainer. I have done it several times. As the commands precedding it may fail, I thought it's kinda rude to leave the user / AI stranded without a working `gh` to revert things. 

3. Seems like it tried to poke holes on the markdown change, but failed midway. 

4. `nicegui-gh-config` makes sense. I do not see why not to share the `gh` auth across all NiceGUI Devcontainers, and were we not to share it, then I don't think we can't persist it at all, since the devcontainer are random-named.

evnchn:
`sudo chown -R vscode:vscode /home/vscode/.config/gh` isn't complete because ChromeDriver breaks and `matplotlib` always complains. Seems like we may need more careful permission management.

evnchn:
image

.devcontainer/post-create.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -e
+
+# Fix ownership of .config/gh (Docker volume mounts it as root)
+sudo chown vscode:vscode /home/vscode/.config
+sudo chown -R vscode:vscode /home/vscode/.config/gh
```
evnchn, line 6: What's wrong with the old `sudo chown -R vscode:vscode /home/vscode/.config`?

Right now, this would leave non-`gh` directories inside `.config` to be owned by root, if I am not mistaken, though I am not sure.

evnchn: https://poe.com/s/oWto0JpoETTWhvH4KUdC btw

evnchn: I mean the two thing which broke:

- A nag message of `matplotlib` complaining it can't write to it
- ChromeDriver not working as it can't write to it

are both gone, so I guess it's a preference thing at this point. Feel free to close this comment thread and merge if you think otherwise.

falkoschindler: Yes, we discussed this intentional change. Here's Opus' explanation:

> Good question. The targeted approach is intentional: the only thing Docker mounts as root is the `gh` volume at `.config/gh`, which can also cause the parent `.config` directory itself to become root-owned. The two `chown` lines fix exactly that:
> 
> 1. `chown vscode:vscode .config` (non-recursive) — fixes the parent dir so vscode can create subdirectories in it
> 2. `chown -R vscode:vscode .config/gh` (recursive) — fixes the volume mount contents
> 
> Any other subdirectories (matplotlib, ChromeDriver, etc.) don't exist yet at this point — they get created later by processes running as vscode, so they'll naturally be vscode-owned. A blanket `chown -R` on all of `.config` would work too, but it's broader than necessary and could mask future ownership issues from other sources.

evnchn: It works now, so maybe we can make it broader when it doesn't later. No need to over-do it.

PR #5752, opened 2026-02-10
fix: declare missing 'radius' variable in Leaflet.draw circle resize

MicaelJarniac:
## Problem

The upstream `leaflet-draw` library's `L.Edit.Circle._resize` function uses a `radius` variable without declaring it with `var`. This was harmless in sloppy mode but causes a `ReferenceError` when loaded as an ES module (strict mode) via the rollup bundle.

## Fix

- Added a `patch-package` patch to insert `var radius;` in the `_resize` function of `leaflet.draw-src.js`
- Switched the dynamic import to use the unminified source (`leaflet.draw-src.js`) so the patch applies cleanly
- Added `patch-package` as a dev dependency with a `postinstall` script
- Rebuilt the bundle

## Upstream issue

This is a known bug in `leaflet-draw` (Leaflet/Leaflet.draw#927) that has never been fixed upstream since the library is unmaintained.

Fixes #5751

PR #5756, opened 2026-02-10
Load leaflet-draw via script tag to avoid ES module strict mode

### LINKED ISSUE #5751: Leaflet Draw circle resize broken: minifier drops var declaration (strict mode ReferenceError)
## Description

The minified Leaflet Draw bundle (`leaflet.draw-CFfIVNdp.js`) shipped with NiceGUI 3.x has a bug that completely breaks circle resize in edit mode.

## Root Cause

In the original Leaflet Draw source, `L.Edit.Circle.prototype._resize` declares two variables in a single `var` statement:

```javascript
var moveLatLng = this._moveMarker.getLatLng(),
    radius;
```

The minifier converted this to:

```javascript
var e=this._moveMarker.getLatLng();
```

The `radius` declaration was dropped. Later in the function, `radius = e.distanceTo(t)` assigns to the now-undeclared variable.

Since NiceGUI 3.x loads the Leaflet Draw bundle via dynamic `import()` (`loadLeafletDraw = () => import("./leaflet.draw-CFfIVNdp.js")`), it runs as an ES module in **strict mode**. In strict mode, assigning to an undeclared variable throws a `ReferenceError`, silently killing the resize handler.

This was likely harmless in older NiceGUI versions where Leaflet Draw may have been loaded as a classic script (sloppy mode allows implicit globals).

## Impact

- Drawing circles: **works** (doesn't invoke `_resize`)
- Moving circles in edit mode: **works** (`_move` has no undeclared variables)
- Resizing circles in edit mode: **broken** (`_resize` throws `ReferenceError: radius is not defined`)

## Steps to Reproduce

1. Create a Leaflet map with `draw_control` enabling circles and editing
2. Draw a circle on the map
3. Click the edit (pencil) toolbar button
4. Try to drag the resize handle on the circle's edge
5. Nothing happens (check browser console for `ReferenceError: radius is not defined`)

```python
from nicegui import ui

m = ui.leaflet(
    center=(0.0, 0.0),
    zoom=2,
    draw_control={
        "draw": {"circle": True},
        "edit": {"edit": True, "remove": True},
    },
)

ui.run()
```

## Workaround

Monkey-patch `L.Edit.Circle.prototype._resize` after Leaflet Draw loads:

```python
await ui.run_javascript("""
    if (L.Edit && L.Edit.Circle) {
        L.Edit.Circle.prototype._resize = function (latlng) {
            var moveLatLng = this._moveMarker.getLatLng(),
                radius;
            if (L.GeometryUtil.isVersion07x()) {
                radius = moveLatLng.distanceTo(latlng);
            } else {
                radius = this._map.distance(moveLatLng, latlng);
            }
            this._shape.setRadius(radius);
            if (this._map._editTooltip) {
                this._map._editTooltip.updateContent({
                    text: L.drawLocal.edit.handlers.edit.tooltip.subtext
                        + '<br />'
                        + L.drawLocal.edit.handlers.edit.tooltip.text,
                    subtext: L.drawLocal.draw.handlers.circle.radius
                        + ': '
                        + L.GeometryUtil.readableDistance(
                            radius, true,
                            this.options.feet,
                            this.options.nautic
                        ),
                });
            }
            this._shape.setRadius(radius);
            this._map.fire(L.Draw.Event.EDITRESIZE, {
                layer: this._shape,
            });
        };
    }
""")
```

## Environment

- NiceGUI 3.7.1
- Bundle file: `nicegui/elements/leaflet/dist/leaflet.draw-CFfIVNdp.js`
- Tested on macOS with Chrome

evnchn:
### Motivation

Fixes #5751.
Alternative approach to #5752.

The upstream `leaflet-draw` library predates ES modules and expects to run in sloppy mode. When Rollup bundles it as an ES module chunk, it is forced into strict mode with no way to opt out, causing `ReferenceError` for undeclared variables (e.g. `radius` in `L.Edit.Circle._resize` — see Leaflet/Leaflet.draw#927).

Rather than patching the library source, this PR sidesteps the issue entirely by loading `leaflet.draw.js` via a plain `<script>` tag, which runs in sloppy mode by default.

### Implementation

1. **Removed the dynamic `import()` of leaflet-draw** from `src/index.mjs` — it is no longer bundled by Rollup as an ES module chunk.
2. **Copy `leaflet.draw.js` byte-for-byte** into `dist/leaflet-draw/` via the existing `rollup-plugin-copy` (no transform, preserving the original file exactly).
3. **Load via `loadResource()`** in `leaflet.js` — this helper already creates a plain `<script>` tag (without `type="module"`), so the script executes in sloppy mode where undeclared variable assignments are permitted.

This avoids the need for `patch-package` or any source modifications to `leaflet-draw`, and is resilient to any other sloppy-mode assumptions the library may have beyond the known `radius` bug.

### Progress

- [x] Existing Pytests cover us well for regression, and I think this is out-of-scope for our testing.
- [x] Documentation is not necessary for underlying bugfix.

### Note

This PR was authored with the assistance of Claude Code on behalf of @evnchn. Local Selenium tests could not be run (Chromium crashes in the ARM devcontainer), so CI results should be checked for confirmation.


evnchn:
Test case stolen from #5751 works:

```py
from nicegui import ui

m = ui.leaflet(
    center=(0.0, 0.0),
    zoom=2,
    draw_control={
        "draw": {"circle": True},
        "edit": {"edit": True, "remove": True},
    },
)

ui.run()
```

So yeah, who wouldn't love a root-cause fix 😄 

@MicaelJarniac I should credit you for discovering the bug and pinning it down to Sloppy Mode. Insightful analysis!

MicaelJarniac:
I like your solution more than my (vibe-coded) one. I agree it's best to "fix" the root cause (strict mode), than to patch all side-effects. Nice work!

evnchn:
I just have to say that AI agents are super-good at patching side-effects, to the point it's better than us humans. 

While useful of a skill, one-liners and simple fixes are not always the best solution.

PR #5757, opened 2026-02-10
Skip redundant innerHTML updates in ui.html to preserve client-side DOM modifications

### LINKED ISSUE #5749: 3.7.1: server-side updates destroy client-side DOM modifications in ui.html elements
### First Check

### Example Code

```python
"""Minimal repro: NiceGUI 3.7.x destroys client-side DOM mods.

Text should have alternating pink/blue character backgrounds.
Click the button to trigger a server-side element update.
In 3.7.x the coloring disappears. In 3.6.1 it stays.

Usage (no install needed, uv creates a temp env):
    uv run --with nicegui==3.6.1 python repro.py  # works
    uv run --with nicegui==3.7.1 python repro.py  # broken
"""

import nicegui
from nicegui import ui


@ui.page("/")
def index():
    ui.label(f"NiceGUI {nicegui.__version__}").classes("text-h5")
    ui.label("Text below has alternating pink/blue chars.")
    ui.label("Click the button. In 3.7.x the color vanishes.")

    ui.html(
        "<p>Hello world sample text for char spans</p>",
        sanitize=False,
    ).props('id="target"')

    # Inject styled spans wrapping each character
    ui.run_javascript("""
        const el = document.getElementById('target');
        if (!el) return;
        let idx = 0;
        function wrap(node) {
            if (node.nodeType === Node.TEXT_NODE) {
                const frag = document.createDocumentFragment();
                for (const ch of node.textContent) {
                    const s = document.createElement('span');
                    s.dataset.charIndex = idx;
                    s.style.backgroundColor =
                        idx % 2 === 0 ? '#ffcccc' : '#ccccff';
                    s.textContent = ch;
                    idx++;
                    frag.appendChild(s);
                }
                node.parentNode.replaceChild(frag, node);
            } else if (node.nodeType === Node.ELEMENT_NODE) {
                Array.from(node.childNodes).forEach(wrap);
            }
        }
        Array.from(el.childNodes).forEach(wrap);
        console.log('[REPRO] Injected ' + idx + ' spans');
    """)

    # Button triggers a server-side element update
    target = ui.label("I am visible").classes("text-lg")

    def on_click() -> None:
        target.set_visibility(not target.visible)

    ui.button(
        "Toggle (triggers server update)",
        on_click=on_click,
    )


ui.run(port=8090)
```

## Steps to reproduce

  1. Run `uv run --with nicegui==3.7.1 python repro.py`
  2. Open http://localhost:8090
  3. Observe alternating pink/blue character backgrounds
  4. Click "Toggle (triggers server update)"
  5. Character backgrounds disappear — DOM modifications wiped

  ## Expected behavior

  Client-side DOM modifications to a `ui.html()` element survive
  server-side updates (as they do in 3.6.1).

  ## Root cause

  The `update` handler in `nicegui.js` (~line 481) deletes elements
  from Vue reactive state when it detects changed event listener IDs,
  then re-adds them after `$nextTick`. Vue unmounts the old DOM node
  (destroying JS-injected content) and creates a fresh one from the
  server's element tree.

  This cycle appears to trigger more aggressively in 3.7.x than 3.6.1.

  ## Secondary issue

  The `beforeUnmount` hook added in #5652 crashes during this cycle:

      TypeError: Cannot read properties of undefined (reading 'props')
          at Proxy.beforeUnmount (input.js:31:51)

  `mounted_app.elements[id]` is already deleted (line 486) before
  Vue's unmount hook fires. Needs a null guard.

  ## Workaround

  Pin `nicegui==3.6.1`.

### NiceGUI Version

3.7.1

### Python Version

Python 3.14.2

### Browser

Chrome

### Operating System

Linux

### 3.7.1

before button click, 3.7.1 
Image

### 3.6.1

Image
Image

evnchn:
### Motivation

Fixes #5749

In NiceGUI 3.7.x, server-side updates (e.g. toggling visibility of an unrelated element) cause `ui.html` to re-assign `innerHTML` even when the content hasn't changed. This destroys any client-side DOM modifications (such as JavaScript-injected styled spans) that were applied after the initial render. This worked correctly in 3.6.1.

### Implementation

Added a `previousInnerHTML` data property to the html component. The `renderContent()` method now compares the current `innerHTML` prop against the cached value and returns early if unchanged, avoiding unnecessary DOM rewrites that would wipe client-side modifications.

### Progress

### DIFF
--- a/nicegui/elements/html.js
+++ b/nicegui/elements/html.js
@@ -1,5 +1,8 @@
 export default {
   template: `<component :is="tag"></component>`,
+  data() {
+    return { previousInnerHTML: null };
+  },
   mounted() {
     this.renderContent();
   },
@@ -8,11 +11,13 @@ export default {
   },
   methods: {
     renderContent() {
+      if (this.innerHTML === this.previousInnerHTML) return;
       if (this.sanitize) {
         this.$el.setHTML(this.innerHTML);
       } else {
         this.$el.innerHTML = this.innerHTML;
       }
+      this.previousInnerHTML = this.innerHTML;
     },
   },
   props: {


evnchn:
As #5761 fixed things from the root-cause and is intended for merge for 3.8, the most recent future release, the duty of this PR as a stopgap fix has vanished. Closing as such.

PR #5759, opened 2026-02-10
Use msgpack for Socket.IO serialization

evnchn:
### Motivation

Socket.IO communication between the Python server and JavaScript client currently uses JSON (via orjson) for packet serialization. MessagePack (msgpack) is a more efficient binary serialization format that reduces payload sizes over the wire.

On evaluation with the https://nicegui.io documentation website, the _transition_ payload from `/documentation` to `/` dropped from **7.3 KB to 5.2 KB** (~29% reduction).

### Implementation

Both python-socketio (server) and socket.io (client) have built-in msgpack support, making this a straightforward switch:

- **Server**: Pass a custom `MsgPackPacket` subclass (via `MsgPackPacket.configure()`) as the `serializer` to `socketio.AsyncServer`. The subclass includes a `dumps_default` handler for types msgpack can't natively serialize (NumPy arrays, NumPy scalars, `Decimal`), mirroring what the orjson wrapper handled previously.
- **Client**: Socket.IO 4.8.1 ships with `socket.io.msgpack.min.js` alongside the standard JSON build. We now copy this variant as the static `socket.io.min.js`.
- **Dependency**: Added `msgpack>=1.0.0` to `pyproject.toml`.

**What does NOT change:**
- `nicegui/json/` module — still used for FastAPI responses, JavaScript generation, persistence
- `nicegui/templates/index.html` — still references `socket.io.min.js` (same filename, different build)
- Air relay — uses its own separate `socketio.AsyncClient()` connecting to the on-air server
- Test `User` fixture — uses a dummy socketio client that doesn't establish real connections

### Progress

- [ ] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.

### DIFF
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -49,6 +49,7 @@ class SocketIoApp(socketio.ASGIApp):
 
 def _make_msgpack_packet_class():
     """Create a MsgPackPacket subclass that handles custom types like NumPy arrays and Decimals."""
+    import datetime  # pylint: disable=import-outside-toplevel
     from decimal import Decimal  # pylint: disable=import-outside-toplevel
 
     from socketio.msgpack_packet import MsgPackPacket  # pylint: disable=import-outside-toplevel
@@ -60,6 +61,10 @@ def _make_msgpack_packet_class():
                 return obj.tolist()
         if isinstance(obj, Decimal):
             return float(obj)
+        if isinstance(obj, (datetime.datetime, datetime.date, datetime.time)):
+            return obj.isoformat()
+        if isinstance(obj, datetime.timedelta):
+            return str(obj)
         raise TypeError(f'Object of type {obj.__class__.__name__} is not msgpack serializable')
 
     return MsgPackPacket.configure(dumps_default=_msgpack_default)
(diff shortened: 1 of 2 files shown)


evnchn:
Too many tests have failed though 🤦

evnchn:
I wish I chose to cook this in evnchn/nicegui instead. Anyways I think we are just 1 test off (`tests/test_socketio_too_long.py::test_socketio_too_long[polling-250240]`) so it means the rest of the codebase works with msgpack.

evnchn:
Two observations:

1. (not related to this PR but) Implicit handshake #5673 actually shoved more params per long-polling call. Not sure if we should turn it off if polling is one of the transport options
2. Long-polling transmits the messagepacked binary message in base64, significantly lowering the capacity, complicating calculations, and worsens performance. Essentially non-workable.

It would be quite a hassle to simultaneously support MessagePack and non-MessagePack clients. As such, this PR can only be merged when NiceGUI is commited to fully dropping long-polling support. 

Considering several users have already had mysterious failures under weird network environments, which were ultimately not resolved without switching to long-polling, I think that day wll never come. As such, I should close this for now.

PR #5761, opened 2026-02-11
Cache vnodes to avoid unnecessary re-renders

falkoschindler:
### Motivation

Relates to #5749. Supersedes #5757.

`renderRecursively` creates new vnode objects and slot function closures on every render pass, which makes Vue treat every component as dirty — even when its data hasn't changed. This has always been the case, but became visible in 3.7 when `ui.html` gained an `updated()` hook that re-assigns `innerHTML`: any unrelated server-side update now triggers that hook and destroys client-side DOM modifications.

PR #5757 addresses the symptom at the component level by skipping redundant `innerHTML` writes in `html.js`. This PR fixes the underlying inefficiency so that unchanged components are never re-rendered in the first place.

### Implementation

Add a vnode cache (`Map<elementId, {vnode, propsContext}>`) to `renderRecursively`. When a server update arrives, the cache is invalidated for the changed elements **and their ancestors** (since ancestor vnodes embed child vnodes via slot closures). Unchanged elements return the same vnode reference, which Vue recognizes as identical and skips entirely — no prop diffing, no slot evaluation, no lifecycle hooks.

This benefits all components, not just `ui.html`: any component with side effects in lifecycle hooks is protected from spurious re-renders, and the overall rendering workload is reduced for every update.

### Progress

- [x] New pytests are not necessary.
- [x] Existing Pytests are passing.
- [x] Add a pytest ensuring `$stable` is still there and `ui.html` isn't re-rendered unexpectedly.

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -47,3 +47,21 @@ def test_xss_sanitization(screen: Screen):
     screen.allowed_js_errors.append('/x - Failed to load resource')
     screen.open('/')
     assert screen.find_by_tag('img').get_attribute('onerror') is None
+
+
+def test_html_manipulation(screen: Screen):
+    @ui.page('/')
+    def page():
+        html_element = ui.html('Old HTML')
+        ui.button('1. Manipulate HTML', on_click=lambda: ui.run_javascript(f'''
+            getHtmlElement({html_element.id}).innerHTML = "New HTML";
+        '''))
+        label = ui.label('Hi')
+        ui.button('2. Manipulate Label', on_click=lambda: label.set_text('Ho'))
+
+    screen.open('/')
+    screen.click('1. Manipulate HTML')
+    screen.should_contain('New HTML')
+    screen.click('2. Manipulate Label')
+    screen.should_contain('Ho')
+    screen.should_not_contain('Old HTML')


falkoschindler:
At the moment this PR doesn't seem to fix the original problem:
```py
ui.html('HTML').props('id=target')
ui.button('1. Manipulate HTML', on_click=lambda: ui.run_javascript('''
    document.getElementById("target").innerHTML += " (changed by JavaScript)";
'''))
label = ui.label('Hi')
ui.button('2. Manipulate Label', on_click=lambda: label.set_text('Ho'))
```
Clicking buttons 1 and 2, the HTML change is reverted by updating the label. 😕

evnchn:
### Analysis: Why vnode-cache doesn't fix #5749

The page hierarchy is: `q-layout → q-page-container → q-page → div.nicegui-content → [ui.html, ui.label, ...]`

When `label.set_text('Ho')` triggers a server update, `invalidateVnodeCache` invalidates the label **and all ancestors** (up to root). During re-render, Quasar components (`q-page`, etc.) provide **scoped slot props** — new object references each time. The cache check uses reference equality (`cached.propsContext === propsContext`), so all children of invalidated ancestors get cache misses, including the unchanged `ui.html`. A new vnode is created, Vue patches it, `updated()` fires, and `innerHTML` is overwritten.

**TL;DR**: ancestor invalidation + Quasar scoped slot props = cache miss for all siblings, making the cache ineffective for the common case.

### Why #5757 works

Instead of preventing the `updated()` hook from firing, #5757 makes `renderContent()` in `html.js` idempotent by tracking the previous prop value. If the server hasn't changed the `innerHTML` prop, the DOM write is skipped.

### Are the two PRs complementary?

The vnode cache is still a valid **performance optimization** (reduces rendering work for subtrees with different ancestors than the changed element), but it doesn't address #5749. The two PRs are complementary: #5757 as the targeted fix, #5761 as a broader optimization on its own merits.

evnchn:
Is there even a `$stable` API?

Can't find it anywhere...

Edit: got it in https://github.com/vuejs/rfcs/blob/master/active-rfcs/0006-slots-unification.md#further-manual-optimization

evnchn:
Wait, all ancestors? Doesn't that mean the root is always re-rendered, and we don't have any performance uplift at all?

falkoschindler:
Opus, the stage is yours:

---

**On `$stable`**: Yes, it's real — it's the mechanism Vue's own template compiler uses to tell the runtime "these slots haven't changed, skip the update check." The RFC they found confirms it. It's not a *documented public API*, but it's a stable internal contract that Vue's compiler has relied on since Vue 3.0. NiceGUI already bypasses the compiler by using `Vue.h()` directly in its render function, so using `$stable` is consistent with that approach. We're essentially doing what the compiler would do if it could prove the slots are unchanged — which our cache invalidation guarantees.

**On "all ancestors always re-rendered"**: Yes, ancestors up to the root are always invalidated. But the performance gain comes from **siblings and their subtrees**, not ancestors. Consider:

```
root → q-layout → q-page-container → q-page → div
  → [ui.html, ui.button, ui.table(1000 rows), ui.chart, ...]
```

When the label changes, the 5 ancestor vnodes are recreated (cheap — just `Vue.h` calls). But the `ui.html`, `ui.table`, `ui.chart`, etc. all return cached vnodes. Vue clones them but (with `$stable`) skips their component updates entirely — no prop diffing, no re-render, no `updated()` hooks. Without the cache, all siblings would get fresh vnodes, full prop comparison, and potentially unnecessary lifecycle hooks.

The ancestor chain is typically ~5 layout elements. The siblings can be hundreds. That's where the uplift is.

I'd also note that evnchn's earlier analysis about Quasar scoped slot props was incorrect — I checked the actual source: `hSlot(slots.default)` calls `slot()` with no arguments. `propsContext` stays `undefined` throughout. The real culprit was `cloneIfMounted` + missing `$stable`, not scoped slot props.

evnchn:
Well, conclusion-speaking, after https://github.com/zauberzeug/nicegui/pull/5761/commits/4b8472ceae1dd0f41cc3a533cc4484a6538c2782 the test code at https://github.com/zauberzeug/nicegui/issues/5749 works

For AI agents ease, a fully-auto test snippet is as follows:

```py
@ui.page('/')
async def page():
    ui.html('HTML').props('id=target')
    label = ui.label('Hi')
    await ui.run_javascript('''
        document.getElementById("target").innerHTML += " (changed by JavaScript)";
    ''')
    label.set_text('Ho')
```

It can only mean the vnode for `ui.html` was not remounted. Great news! With this I think this PR now finally supersedes #5757. 

@falkoschindler I think we should examine the current pytests to see if they can double-duty to check that the `$stable` API is still there. I would desire that at least 1 test is red if the API is suddenly gone from Vue.

PR #5762, opened 2026-02-11
Update bundled Google Fonts (Material Symbols + Roboto v50)

evnchn:
### Motivation

Users report that newer Material Symbols icons like `database_upload`, `cloud_alert`, and `desktop_cloud` render incorrectly as two separate icons instead of one. This is because the bundled Material Symbols fonts were last updated in January 2024 and are missing icons added since then.

See discussion: https://github.com/zauberzeug/nicegui/discussions/5743

An audit script confirmed the update is purely additive -- 519 new codepoints added, 0 removed across all font variants. Full audit results posted here: https://github.com/zauberzeug/nicegui/discussions/5743#discussioncomment-15769778

### Implementation

- Re-ran `fetch_google_fonts.py` to download latest fonts from Google Fonts
- Fixed `fetch_google_fonts.py` to handle duplicate font URLs: Google now serves Roboto v50 as a variable font where the same file is shared across multiple weights, which caused the script to crash on duplicate filenames
- Removed stale version-pinned URL fallback replacements (`v140`, `v30`) that were dead code after the download loop

**Font changes:**
- Material Icons (all 4 variants): byte-for-byte identical, no change
- Material Symbols Outlined: 254KB -> 313KB (+519 codepoints)
- Material Symbols Rounded: 294KB -> 361KB (+519 codepoints)
- Material Symbols Sharp: 241KB -> 298KB (+519 codepoints)
- Roboto: v30 -> v50 (variable font), 42 files -> 9 files
- Total font files: 49 -> 16

### Progress

### DIFF
--- a/fetch_google_fonts.py
+++ b/fetch_google_fonts.py
@@ -28,17 +28,14 @@ css += '\n' + httpx.get(f'https://fonts.googleapis.com/css?family={"|".join(css_
                         headers={'User-Agent': AGENT}, timeout=5).content.decode()
 css += '\n' + httpx.get(f'https://fonts.googleapis.com/css?family={"|".join(css2_families)}',
                         headers={'User-Agent': AGENT}, timeout=5).content.decode()
-downloaded: dict[str, Path] = {}
-for font_url in re.findall(r'url\((.*?)\)', css):
-    if font_url not in downloaded:
-        font = httpx.get(font_url, timeout=5).content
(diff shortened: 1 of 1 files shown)


evnchn:
I like this for the Lighthouse improvement as well: variable fonts are the future!

falkoschindler (approved):
Nice! 👌🏻

PR #5766, opened 2026-02-11
Guard against missing element in beforeUnmount hooks

### LINKED ISSUE #5765: TypeError in input.js beforeUnmount when element is already removed
### First Check

### Example Code

```python
@ui.refreshable
def show_input():
    ui.input()

show_input()
ui.button('Refresh', on_click=show_input.refresh)
```

### Description

When using `ui.input()` inside a `@ui.refreshable`, clicking the Refresh button causes a `TypeError` in the browser console.

**Steps to reproduce:**
1. Run the example code above.
2. Type something into the input field.
3. Click the "Refresh" button.

**Expected:** The input refreshes without errors.

**Actual:** A JavaScript error is thrown:
```
TypeError: Cannot read properties of undefined (reading 'props')
```

This happens in `input.js` at the `beforeUnmount` hook (line 31). When a refreshable rebuilds its content, the element may already be removed from `mounted_app.elements` by the time Vue's `beforeUnmount` fires. The code assumes `mounted_app.elements[this.$props.id.slice(1)]` always exists, but it can be `undefined`.

### NiceGUI Version

3.7.1

### Python Version

3.10.19

### Browser

Chrome

### Operating System

Linux

evnchn:
### Motivation

Fixes #5765

When a `ui.input()` is inside a `@ui.refreshable`, clicking the refresh button causes a `TypeError` in the browser console:

```
TypeError: Cannot read properties of undefined (reading 'props')
```

This happens because `mounted_app.elements[id]` may already be removed by the time Vue's `beforeUnmount` hook fires during a refreshable rebuild.

### Implementation

Add a simple guard check in the `beforeUnmount` hook in `input.js` to verify the element still exists in `mounted_app.elements` before accessing its `props.value`.

```diff
- mounted_app.elements[this.$props.id.slice(1)].props.value = this.inputValue;
+ const element = mounted_app.elements[this.$props.id.slice(1)];
+ if (element) element.props.value = this.inputValue;
```

### Progress

- [x] Existing Pytests cover the regression.

🤖 Generated with [Claude Code](https://claude.com/claude-code), then modified a bit.

### DIFF
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -39,7 +39,8 @@ export default {
   },
   beforeUnmount() {
     if (this.editor) {
-      mounted_app.elements[this.$props.id.slice(1)].props.value = this.editor.state.doc.toString();
+      const element = mounted_app.elements[this.$props.id.slice(1)];
+      if (element) element.props.value = this.editor.state.doc.toString();
     }
   },
   methods: {
--- a/nicegui/elements/editor.js
+++ b/nicegui/elements/editor.js
@@ -17,7 +17,8 @@ export default {
     };
   },
   beforeUnmount() {
-    mounted_app.elements[this.$props.id.slice(1)].props.value = this.inputValue;
+    const element = mounted_app.elements[this.$props.id.slice(1)];
+    if (element) element.props.value = this.inputValue;
   },
   watch: {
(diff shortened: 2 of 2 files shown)


evnchn:
https://github.com/evnchn/nicegui/pull/81

Let's wait if it does a good job then.

PR #5767, opened 2026-02-11
Add comprehensive SEO support for documentation website

evnchn:
### Motivation

Improve the discoverability and search engine ranking of the NiceGUI documentation website. Currently the site lacks per-page meta descriptions, Open Graph/Twitter Card tags, structured data, a sitemap, and other SEO fundamentals, which limits how well pages appear in search results and social media shares.

### Implementation

- **`website/seo.py`** (new): Central module for SEO helpers — `extract_description()` pulls page descriptions from docstrings, `build_meta_tags()` generates Open Graph/Twitter Card/canonical meta tags, `build_structured_data()` produces JSON-LD (Organization, SoftwareApplication, BreadcrumbList)
- **`main.py`**: Adds `/sitemap.xml` and `/robots.txt` endpoints dynamically generated from the documentation registry; adds `<noscript>` fallback and JSON-LD structured data to all pages via `app.on_page_ready`
- **`website/documentation/rendering.py`**: Injects per-page meta descriptions, OG tags, and canonical URLs into documentation pages
- **`website/header.py`**: Adds breadcrumb navigation and `<noscript>` content
- **`nicegui/templates/index.html`**: Adds `lang="en"` attribute to `<html>` tag
- **`nicegui/app/app.py`**: Resets `_page_exception_handler` in `App.reset()` for test isolation (needed because website imports trigger demo code that registers a custom exception handler)
- **`tests/test_seo.py`** (new): Tests for description extraction, meta tag generation, and structured data output

Page titles are kept under 60 characters per Google's recommendation. The static meta description in `header.html` is removed in favor of dynamic per-page descriptions.

### Progress

### DIFF
--- a/main.py
+++ b/main.py
@@ -148,6 +148,7 @@ def _documentation_detail_page(name: str, tree: ui.tree) -> None:
     elif name in documentation.redirects:
         ui.navigate.to('/documentation/' + documentation.redirects[name])
     else:
+        ui.status_code(404)
         ui.label(f'Documentation for "{name}" could not be found.').classes('absolute-center')
 
 
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -95,6 +95,7 @@ class Client:
                     self.content = Element('div').classes('nicegui-content')
 
         self.title: str | None = None
+        self.status_code: int | None = None
 
         self._head_html = ''
         self._body_html = ''
@@ -193,7 +194,7 @@ class Client:
                 'socket_io_js_extra_headers': core.app.config.socket_io_js_extra_headers,
                 'socket_io_js_transports': core.app.config.socket_io_js_transports,
             },
-            status_code=status_code,
+            status_code=self.status_code if self.status_code is not None else status_code,
             headers={'Cache-Control': 'no-store', 'X-NiceGUI-Content': 'page'},
         )
 
--- a/nicegui/functions/status_code.py
+++ b/nicegui/functions/status_code.py
@@ -1,0 +1,11 @@
+from ..context import context
+
+
+def status_code(code: int) -> None:
+    """Status code
+
+    Set the HTTP status code for the current page response.
+
+    :param code: HTTP status code (e.g. 200, 404, 503)
+    """
+    context.client.status_code = code
--- a/nicegui/ui.py
+++ b/nicegui/ui.py
@@ -112,6 +112,7 @@ __all__ = [
     'spinner',
     'splitter',
     'state',
+    'status_code',
     'step',
     'stepper',
     'stepper_navigation',
@@ -265,6 +266,7 @@ from .functions.on import on
 from .functions.on_exception import on_exception
 from .functions.page_title import page_title
 from .functions.refreshable import refreshable, refreshable_method, state
+from .functions.status_code import status_code
 from .functions.style import add_css, add_sass, add_scss
 from .functions.update import update
 from .page import page
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -2,6 +2,7 @@ import asyncio
 import re
 from typing import Literal
 
+import httpx
 import pytest
 from fastapi.responses import PlainTextResponse
 from selenium.webdriver.common.by import By
@@ -336,3 +337,22 @@ def test_warning_if_response_takes_too_long(screen: Screen):
     screen.allowed_js_errors.append('/ - Failed to load resource')
     screen.open('/')
     screen.assert_py_logger('WARNING', re.compile('Response for / not ready after 0.5 seconds'))
+
+
+def test_status_code(screen: Screen):
+    @ui.page('/')
+    def page():
+        ui.status_code(404)
+        ui.label('not found')
+
+    screen.start_server()
+    assert httpx.get(f'http://localhost:{Screen.PORT}/').status_code == 404
+
+
+def test_status_code_default(screen: Screen):
+    @ui.page('/')
+    def page():
+        ui.label('hello')
+
+    screen.start_server()
+    assert httpx.get(f'http://localhost:{Screen.PORT}/').status_code == 200


evnchn:
We have the most advanced auto-generating Python-based documentation using our own library but we don't have a sitemap?

image


I am surprised that NiceGUI _shows up_ on Google _at all_.

evnchn:
Hmm, should NiceGUI auto-populate `sitemap.xml` based on `@ui.page` and `ui.sub_page` calling attempts?

evnchn:
@falkoschindler Since:

- Not having a sitemap really messes up the Google Search Console (a bunch of non-indexed URLs from a prior mishap + non-indexable stuff like `components` and `esm` showing up)
- SEO effectiveness takes a while to propagate

I would recommend to merge this in 3.8 so that we can iterate another round circa end of 3.9

evnchn:
@falkoschindler to have all SEO-facing changes in the pipeline in one SEO re-index run, can we slip #5792 and #5793 into 3.8?

Trust me it's very small changes and quick-win in nature. 

If it's too tough then perhaps all 3 can go in 3.9

falkoschindler:
@evnchn This one is pretty tough to review: +502 lines and lots of concerns in one PR.

By the way, have you asked an LLM for review before posting the PR?

Here are Opus' findings. Some are debatable, but especially point 1 resonates with me: The contribution would benefit from splitting different concerns into separate PRs.

---

### BLOCKER

**1. Too many unrelated concerns bundled together**

This PR mixes at least 4 distinct changes with different risk profiles:

| Change | Scope | Risk |
|--------|-------|------|
| SEO meta/sitemap/robots.txt | `website/` only | Low |
| `ui.status_code()` | **Public API** (`nicegui.*`) | High — needs deprecation-aware review |
| ESM empty-path 404 | Core library | Medium |
| `<html lang>` template change | **All NiceGUI users** | Medium |
| `_page_exception_handler` reset | Core library | Low-Medium |

Per AGENTS.md: *"`nicegui.*` is public API – breaking changes need deprecation paths"*. `ui.status_code()` deserves its own PR with dedicated discussion about API design (e.g., should it validate the code? should it return the previous code? what happens if called outside a page context?).

Recommendation: Split into at least 2-3 PRs: (a) website SEO, (b) `ui.status_code()`, (c) ESM path fix.

**2. Sitemap `lastmod` is always `date.today()`**

```python
today = date.today().isoformat()
# ...every URL gets today's date as lastmod
```

Google [explicitly warns](https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap) against inaccurate `lastmod` values. Setting every page to "today" on every request means crawlers will never trust the value. This can **hurt** crawl efficiency rather than help it. Either track actual modification dates or omit `lastmod` entirely.

**3. `<html lang>` template change affects all NiceGUI users**

```html
<html lang="{{ language.split('-')[0] }}">
```

This is applied to the core template in `nicegui/templates/index.html`, meaning it affects **every** NiceGUI application, not just the website. While `lang` is a good addition, this:
- Should be its own change with its own tests
- Needs verification that `language` is never `None` at template render time (if `Language` type allows `None`, `None.split('-')` will crash the template)
- Should arguably use the full BCP-47 tag (`en-US`) not just the primary subtag (`en`), or make this configurable

---

### MAJOR

**4. `ui.status_code()` has no input validation**

```python
def status_code(code: int) -> None:
    context.client.status_code = code
```

No validation that `code` is a valid HTTP status code (100-599). No guard against being called outside a client context. Per AGENTS.md: *"Validate user inputs at API boundaries for critical operations."*

**5. `_page_exception_handler = None` reset is a workaround, not a fix**

The commit message says: *"needed because website imports trigger demo code that registers a custom exception handler"*. This sounds like a test isolation problem that should be fixed at its source rather than adding a reset. If demo code has side effects at import time, that's the real bug.

**6. Missing breadcrumb JSON-LD on main page**

`main_page.py` adds `page_seo_html` and `noscript_fallback` but omits `breadcrumb_jsonld`, unlike `examples_page.py`, `imprint_privacy.py`, and all documentation pages. This is inconsistent.

**7. `test_seo.py` tests website code, not library code**

`tests/test_seo.py` (170 lines) tests `website.seo` helper functions (`extract_description`, `page_seo_html`, `breadcrumb_jsonld`), which are website-specific code, not library code. NiceGUI's `tests/` directory is for testing the `nicegui` package. Mixing website tests into the library test suite:
- Creates a dependency from tests → `website/` code, which is unusual
- Makes CI run website-specific tests for every library PR
- Blurs the boundary between "library" and "website that happens to live in the same repo"

If these tests should exist, they'd fit better as `website/tests/test_seo.py` or similar, separate from the library test suite.

---

### CLEANUP

**8. `noscript_fallback` has limited SEO value**

Googlebot executes JavaScript. The `<noscript>` blocks are mostly useful for users with JS disabled, not search engine crawlers. This adds HTML weight to every page for marginal benefit.

**9. No integration tests for rendered pages**

Tests only cover pure helpers in `seo.py`. No tests verify that documentation pages actually render with correct meta tags, or that `/sitemap.xml` and `/robots.txt` return valid content.

**10. Page titles aren't actually capped at 60 chars**

The PR description claims *"Page titles are kept under 60 characters per Google's recommendation"* but there's no enforcement. `seo_title = f'{title} - NiceGUI Documentation'` can exceed 60 chars for longer element titles.

evnchn:
> have you asked an LLM for review

Most definitely. Several rounds of it in fact. Each round I run /review and also have another agent pretend to be SEO expert. 

I'll go through the rest of the comments after my family dinner.

falkoschindler:
No rush!

evnchn:
Hmm. Why is your Opus so much better than my Opus at SEO? How did you prompt it? 

I asked mine to "web-search SEO knowledge", did someone poison the internet databank? 

For the concerns:

1. [many concern] Disagree. Given our repo's situation, I think you would appreciate not to have 5 PRs instead of 1. I always welcome to cut things down when we agree to it. I just want to show you completely what's on the table. 

2. [lastmod always today] Agree. Genuinely do not know about this one OMG! I think we _can_ get away with a valid lastmod if we update it per version bump. However there will still exist cases where the Googlebot scrapes our docs and finds nothing new and becomes disappointed. To counter that, I recommend to include a new link "Release notes" which has path `/release-notes/3.8.0` (version is intentional) such that when we bump version, _at least one entry_ will be changed. 

3. (1) Agree. (2) Disagree. Read `nicegui/language.py` it's never `None` if the user follow our instruction. Still may be good to do some defensive programming, (3) Not sure, let's web-search

4. Disagree. Out-of-normal-scope HTTP error codes are commonly used for internal operations. I remember using `666` in my personal project. Cloudflare even tosses you `10xx` HTTP error codes, so yeah. 

5. Disagree. It would require substantial effort to refactor to limit the import side effect, and I mean we should honestly test `app.on_page_exception` to be honest. 

6. Not sure. My Opus promised me everything is there and I saw it run quite many curl commands against `http://localhost:8080`. Imma check

7. Agree. Though testing is intentional, we should probably move it somewhere else (I have plans to selective opt-out of running stuff in `tests/` if no library code is changed, and so we better get the separation right now)

8. Disagree. "Googlebot executes JavaScript" is not quite true. It is only for large known websites. Otherwise we would not have so many websites indexed as "Connection lost. Trying to reconnect..." as in https://github.com/zauberzeug/nicegui/discussions/4506

9. Agree, but not sure if it's worth it - Google Search Console is the ultimate integration test. 

10. `seo_title = f'{title} - NiceGUI Documentation'` can, but I already shortened the rest. We may cut it down to `seo_title = f'{title} - NiceGUI Docs'` or `seo_title = f'{title} - NiceGUI'` for exceptionally long titles, but there is only so much the code can do. IMHO SEO awareness needs to permeate the team for effective SEO.

falkoschindler:
> Why is your Opus so much better than my Opus at SEO?

Not sure. Usually I simply prompt "/review 5767". This command should be universal. But it seems like separate chats have some kind of shared memory, so maybe yours is a bit biased because it wrote the PR itself?

> Given our repo's situation, I think you would appreciate not to have 5 PRs instead of 1.

Well, the sum of 5 small PRs isn't easier to review then each of them separately. The larger complexity makes the review of the mono-PR harder. So, in the end, more code is in limbo much longer.

Therefore I'd like to ask you to split this into separate PRs. I know that's extra work and I'm sorry about that, but it would really simplify the review process - each piece has different review criteria and risk profiles. I see the following three separate areas. If they depend on each other, we can review and merge them sequentially. And if not everything makes it into 3.8, that's perfectly fine - better to land things right than to rush them.

### 1. `ui.status_code()` — New public API
- `nicegui/functions/status_code.py` (new)
- `nicegui/client.py` (adds `status_code` field)
- `nicegui/ui.py` (exports it)
- `tests/test_page.py` (tests)

This is a **public API addition** to `nicegui.*` — the highest-stability part of the codebase per AGENTS.md. It deserves its own focused review. Questions like: should it validate the code range? What happens if called multiple times? Does it need documentation on the website? None of that is addressed here because it's buried in an SEO PR.

### 2. Unrelated bug fixes
- `nicegui/nicegui.py` — ESM empty-path guard (completely unrelated)
- `nicegui/app/app.py` — `_page_exception_handler` reset (test isolation fix)
- `nicegui/templates/index.html` — `lang` attribute on `<html>` (library-wide template change affecting **all** NiceGUI users, not just the website)

These are small, easy to review independently, and shouldn't be held up by or entangled with the SEO discussion.

### 3. Website SEO (the actual feature)
- `website/seo.py` (new, ~118 lines)
- `website/header.py`, `website/main_page.py`, `website/examples_page.py`, `website/imprint_privacy.py`, `website/documentation/rendering.py`
- `main.py` (sitemap.xml, robots.txt endpoints)
- `website/static/header.html`
- `tests/test_seo.py`

evnchn:
Sure. Will do. 

Retrospect, it is fine if #5792 and #5793 in 3.8 and this is released whenever. 

The only problematic case is when this is released _earlier_ that the two fixes. 

So yeah, we can take our time. 

Regarding the shared memory: I have seen `MEMORY.md` explicitly edited **exactly once** when I tried to ask it to remember something. I'm not sure if the agent silently edits the file when performing tasks. 

I'll be sure to check it from now on, but I also believe (1) it's product design failure on Claude Code if this is not apparent to me as an end user, and (2) perhaps we should document this quirk as more and more people contribute to the repo using Claude Code.

evnchn:
I don't see any MEMORY.md files. Could be because I rebuilt the container recently. 

On the PR focusing:

- `_page_exception_handler` reset will be in its own PR and **merge with high priority** because it currently blocks all documentation features from being pytested, secretly blocking https://github.com/evnchn/nicegui/tree/docs-code-transformation-callables (supersedes #5274) from being a PR. 
  - "website imports trigger demo code that registers a custom exception handler" doesn't sound ideal, so if we can reduce the import side effects such that we do not need the `_page_exception_handler` reset, it is welcome and even better of a solution. 
- `ui.status_code()` will be in its own PR. In addition to the additional tasks needed to bring it up to speed, that PR it will be responsible to add `ui.status_code()` to the documentation page's 404. 
- `lang` attribute on `<html>` will be in its own PR as well. 
- I prefer `ESM empty-path guard` still should belong to this PR because it serves the explicit purpose to satisfy the Googlebot (non-user-facing whatsoever as that path never triggers in real life) but I can make it into another one-liner PR if you prefer. 
- The rest stays in this PR.

evnchn:
Superseded by focused PRs split per review feedback:

- #5809 — `_page_exception_handler` reset
- #5810 — `ui.status_code()`
- #5811 — `<html lang>` attribute
- #5813 — SEO support (this PR's core scope, cleaned up)

🤖 Closed by Claude Code on behalf of @evnchn

PR #5772, opened 2026-02-13
Add support for typing numbers in UserInteraction

falkoschindler:
### Motivation

See discussion #5771: Calling `user.find(ui.number).type('42')` in user simulation tests raises an `AssertionError` because `type()` only supports `ui.input`, `ui.editor`, and `ui.codemirror`. There's no reason `ui.number` shouldn't work too.

### Implementation

- Added `ui.number` as a supported element in `UserInteraction.type()`
- For `ui.number`, the typed text is converted to `float` and set directly (no string concatenation, since that doesn't apply to numeric values)
- Replaced the bare `assert` with an explicit `TypeError` for unsupported element types, giving a clear error message
- Added `test_type_number` to verify typing and clearing on `ui.number` elements

### Limitation

~~Unlike text inputs, `type()` on `ui.number` does not simulate character-by-character typing. It converts the full text to a float in one step. Simulating intermediate keystroke states (e.g., `type('7.')` → NaN → `type('0')` → 7.0) would require modeling browser-level input behavior, which is out of scope for this change.~~

Each `type()` call must produce a valid float. Intermediate states that aren't valid floats (like `"."` or `"1e"`) can't be represented, and trailing dots get lost (e.g., `type("7.")` + `type("5")` yields `75.0` instead of `7.5` because `"7."` is parsed as `7`).

### Progress

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -65,7 +65,8 @@ class UserInteraction(Generic[T]):
                 if isinstance(element, DisableableElement) and not element.enabled:
                     continue
                 if isinstance(element, ui.number):
-                    element.value = float(text)
+                    current = element._value_to_model_value(element.value) or ''  # pylint: disable=protected-access
+                    element.value = float(current + text)
                 elif isinstance(element, (ui.input, ui.editor, ui.codemirror)):
                     element.value = (element.value or '') + text
                 else:
--- a/tests/test_user_simulation.py
+++ b/tests/test_user_simulation.py
@@ -234,19 +234,23 @@ async def test_input(user: User, kind: type) -> None:
 async def test_type_number(user: User) -> None:
     @ui.page('/')
     def page():
-        number = ui.number(on_change=lambda e: ui.notify(f'Changed: {e.value}'))
+        number = ui.number()
         ui.label().bind_text_from(number, 'value', lambda v: f'Value: {v}')
 
     await user.open('/')
 
-    user.find(ui.number).type('42')
+    user.find(ui.number).type('4')
+    await user.should_see('Value: 4.0')
+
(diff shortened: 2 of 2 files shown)


nicegui/testing/user_interaction.py
```diff
@@ -64,8 +64,12 @@ def type(self, text: str) -> Self:
             for element in self.elements:
                 if isinstance(element, DisableableElement) and not element.enabled:
                     continue
-                assert isinstance(element, (ui.input, ui.editor, ui.codemirror))
-                element.value = (element.value or '') + text
+                if isinstance(element, ui.number):
+                    element.value = float(text)
```
evnchn: Unlike the 3 other elements, `type` on `ui.number` does not take the prior value into account. Is this intentional?

evnchn (commented):
Not sure about the behavioural discrepancy. 

But if we take the old value into consideration, then things may not work out always. 

What if the value was 1.2, and I thought it was 1 so then I typed ".2"?

Then we would have 1.2.2?

evnchn:
I think we should clearly document somewhere this discrepancy. Otherwise it's easy to confuse people.

falkoschindler:
@evnchn You're right, this is confusing. In eff92c19f32830d834f2db290ecdbde31f680ede I changed it so that incremental typing is possible. Now the only limitation is that you have to end in a valid float value:
- Typing ".5" is ok and results in 0.5, but typing "." + "5" doesn't work because "." can't be parsed.
- Typing "7" + ".5" is ok and results in 7.5, but typing "7." + "5" results in 75.0 because the intermediate "7." is parsed as 7.

I think this is a reasonable tradeoff.

evnchn:
What happens when the user does land in an invalid situation? Error out? (Good thing, considering pytest scenario)

PR #5775, opened 2026-02-13
Preserve cursor position when calling `ui.codemirror.set_value`

falkoschindler:
### Motivation

When calling `set_value` on `ui.codemirror`, the entire document was replaced, which reset the cursor position and any active selection. This made it impossible to programmatically update part of the editor content (e.g. via a timer or external event) without disrupting the user who is actively typing.

> **Note:** This does not enable true multi-user collaborative editing. Simultaneous edits from multiple clients still follow last-writer-wins semantics, which can cause lost keystrokes. Real-time collaboration would require CRDT or OT, which is out of scope here.

### Implementation

**Minimal diff in `setEditorValue`** ([codemirror.js](nicegui/elements/codemirror/codemirror.js)):
- Instead of replacing the full document, `setEditorValue` now finds the common prefix and suffix between the old and new text, then dispatches only the changed region to CodeMirror.
- Cursors and selections outside the edited region are preserved automatically by CodeMirror's transaction system.
- O(n) character scan, no new dependencies. Behavior is unchanged for equal documents (early return) and full replacements.

**New "Preserving Cursor Position" documentation demo** ([codemirror_documentation.py](website/documentation/content/codemirror_documentation.py)):
- A timer updates the first line every second while the user edits freely below, demonstrating that the cursor stays in place.

**Pytest** ([test_codemirror.py](tests/test_codemirror.py)):
- `test_set_value_preserves_cursor`: places the cursor after "Hello" in "Hello World", calls `set_value("Hello Earth")`, then types a comma — asserts the result is "Hello, Earth".

### Progress

### DIFF
--- a/website/documentation/content/codemirror_documentation.py
+++ b/website/documentation/content/codemirror_documentation.py
@@ -1,5 +1,3 @@
-from collections.abc import Callable
-
 from nicegui import ui
 
 from . import doc
@@ -16,28 +14,17 @@ def main_demo() -> None:
                 on_change=lambda e: editor.set_line_wrapping(e.value))
 
 
-@doc.demo('Collaborative Editor', '''
-    Multiple users can edit the same document simultaneously.
-    A shared ``Event`` notifies all editors when the content changes.
-
-    *Since version 3.8.0:*
-    ``set_value`` applies only the modified region to preserve cursor positions.
+@doc.demo('Preserving Cursor Position', '''
+    ``set_value`` applies only the modified region, so cursor positions and selections outside the change are preserved.
+    Try editing the code below while the first line updates automatically.
 ''')
-def collaborative_editor_demo() -> Callable:
-    from nicegui import Event, events
+def preserve_cursor_demo() -> None:
+    from datetime import datetime
 
-    document = {'value': 'print("Hello, world!")'}
-    change = Event[str]()
-
-    def root():
-        def on_change(e: events.ValueChangeEventArguments) -> None:
-            document['value'] = e.value
-            change.emit(e.value)
-
-        editor = ui.codemirror(value=document['value'], on_change=on_change)
-        change.subscribe(editor.set_value)
-
-    return root
+    editor = ui.codemirror(f'# {datetime.now():%H:%M:%S}\n', language='Python')
+    ui.timer(1, lambda: editor.set_value(
+        f'# {datetime.now():%H:%M:%S}\n' + editor.value.split('\n', 1)[-1]
+    ))
 
 
 doc.reference(ui.codemirror)


evnchn:
Occupied, but I literally have flashbacks when looking at `ui.codemirror` diff processing. Can you test and comment whether this would work considering Emojis? Ref: https://github.com/zauberzeug/nicegui/pull/4678

falkoschindler:
## Analysis: Does the new `setEditorValue` work with emojis?

**Yes, it works correctly.** Here's why:

### The new diff logic operates entirely in JavaScript's UTF-16 space

The key code in codemirror.js:104-122:

```javascript
const old = this.editor.state.doc.toString();  // JS string (UTF-16)
// ...
while (start < old.length && start < value.length && old[start] === value[start]) start++;
// ...
this.editor.dispatch({ changes: { from: start, to: oldEnd, insert: value.slice(start, newEnd) } });
```

Both `old` and `value` are JavaScript strings. The comparison `old[start] === value[start]` iterates through UTF-16 code units, and `start`/`oldEnd` are UTF-16 code unit offsets. CodeMirror 6 positions are also UTF-16 code unit offsets. So the `from`, `to`, and `insert` values are all in the same coordinate system — no cross-language conversion happens.

### Surrogate pair splitting is harmless here

Consider replacing `"a🔄b"` with `"a🔥b"` (where 🔄 = `\uD83D\uDD04` and 🔥 = `\uD83D\uDD25`):

- Prefix scan: `'a'` matches, `\uD83D` matches (shared high surrogate), `\uDD04 ≠ \uDD25` → `start = 2`
- Suffix scan: `'b'` matches → `oldEnd = 3, newEnd = 3`
- Dispatch: `{ from: 2, to: 3, insert: "\uDD25" }`

The diff boundary falls *inside* the surrogate pair, replacing only the low surrogate. This is fine because CodeMirror's position model is UTF-16-based — position 2 is a valid position. The resulting document `"\uD83D\uDD25"` correctly forms 🔥.

### No interaction with the Python-side codepoints mechanism

This is the crucial distinction from PR #4678. The problematic path in #4678 was **JS → Python** (change sets with JS indices needing conversion to Python code points via `_codepoints`). The `setEditorValue` path is **Python → JS**:

1. Python calls `set_value()` → updates `_codepoints` for the new value
2. JS receives the new value string and diffs it against the current document
3. `emitting = false` prevents `changeSender` from firing, so **no change set goes back to Python**
4. No cross-language index conversion occurs at all

The entire operation stays within JavaScript's UTF-16 world, so the UTF-16 vs Unicode code point mismatch that #4678 addressed is simply not in play here.

### Summary

The new `setEditorValue` diff is emoji-safe because it never crosses the JS↔Python boundary with positions — the diff is computed and applied entirely in JavaScript using UTF-16 indices, which matches CodeMirror's native position model. The `_codepoints` / `_apply_change_set` machinery from #4678 is only used in the opposite direction (when the user types in the editor), and this PR doesn't change that path.

evnchn:
1. Can we pytest this?
2. `Event` doesn't work well on https://nicegui.io until #5231 merged. Can you document?

evnchn:
Another two questions:

- What would happen, if two users type on almost the same time (delta < network propagation latency)? Does there exist a scenario where since the changesets are applied in different order, that the sync falls apart?
- Follow-up the above, consider how before this PR, since the full value is always sent, the sync is always self-healing (there is no room for falling out of sync). Now, it doesn't look like it, so what happens when some sync doesn't line up?

I think if we want to get serious with collaborative editing, we should propagate the content hash of the JS codemirror to the server. When the hashes do not align, there has to be a mechanism to snapback (preferably with a decision-window so that the user doesn't lose their local changes?)

And as such, I don't think this is prime for 3.8

falkoschindler:
I'll let Opus speak:

---

**Comment 1 (testing + Event on nicegui.io):**

Both points are reasonable. A browser test verifying cursor preservation after `set_value` would be valuable. And noting that the collaborative demo requires single-instance deployment (until #5231 lands) is fair — though that's an `Event` limitation, not specific to this PR.

**Comment 2 (race conditions / sync divergence):**

I think this misreads the architecture. This PR does **not** introduce distributed changesets or OT. The server still sends the **full target value** to each client. The diff is computed **locally inside `setEditorValue`** by comparing the client's current document against the incoming complete string. So:

- Before this PR: client receives `"Hello X"` → replaces entire document → cursors reset
- After this PR: client receives `"Hello X"` → computes minimal edit from current doc to `"Hello X"` → applies only that region → cursors preserved

The end-state of the document is identical in both cases. The diff is purely a **presentation optimization** — it determines the smallest CodeMirror transaction to reach the target state, not a semantic changeset that could be misordered. There's no scenario where applying these local diffs causes divergence that wouldn't also happen with full replacement, because the target is always the authoritative full value from the server.

The "last writer wins" race condition exists equally before and after this PR. If two users type simultaneously, whichever change the server processes last will overwrite the other. That's inherent to the `Event`-based broadcast pattern, not to the diff logic.

The suggestion about content hashing and snapback mechanisms would be relevant if this were building toward a CRDT/OT system, but that's a much larger scope than what this PR aims for.

---

So I'll add a test and a note for the demo. But I don't think we need to postpone to 3.9.

evnchn:
Bitten by Microsoft PowerPoint online which asks me to keep either my slides or my teammate's slides back when I was in high school, **I believe we should actively refrain from saying something is "collaborative editing" if it isn't working over CRDT for at least 99.999% of the time.** 

Since this PR does not address "last writer wins" but still brings value by enabling "preserve cursor positions", I would recommend to add a mutex lock so that only one `ui.codemirror` instance is editable at any point in time, so as to highlight the value of this PR without stretching NiceGUI beyond its current capabilities.

falkoschindler:
Test added.

Regarding the note, I came to the conclusion that it isn't necessary here. The demo is rendered per-visitor on the documentation page. There's no second tab to sync with anyway. So I think we don't need a disclaimer.

evnchn:
> The demo is rendered per-visitor on the documentation page. There's no second tab to sync with anyway. So I think we don't need a disclaimer.

But the moment someone tries to spin up this demo themselves and realize "hey it seems like I can edit this textbox from my phone and my PC" with "and" really being "or" since he only has 2 hands, I believe there is a chance of miscommunication and disappointment. Really reconsider 🥺

falkoschindler:
@evnchn I don't really get your point. If users run the demo locally, they _can_ edit from multiple devices and the content will be in sync. Only a distributed setup with a loadbalancer will fail. But that's a special case that doesn't have much to do with `ui.codemirror`.

evnchn:
I just tried it on my phone and laptop. **Simultaneously**, my laptop I type "p" and my phone I type "q"

I pressed equal times:

image

Each time, 1 character is kept only. 

The promised "collaborative editing" is not happening, if I try and type "banana" on my phone and "apple" on my laptop, neither parties would make a complete word. 

And, what's worse, if both parties are editing on the same line, then the cursor position of one party will jump around as the "find the changed region" becomes super confused. 

@falkoschindler Have you tried to edit, _simultaneously_, on multiple devices?

evnchn:
Hmm. I did try to implement single-device edit, but I lost the cursor position. 

```py
from nicegui import Event, binding, events, ui

change = Event[str]()


class State:
    editing_client_id: binding.BindableProperty

    def __init__(self):
        self.editing_client_id = None
        self.document: dict[str, str] = {'value': 'print("Hello, world!")'}


state = State()


def root():
    my_client_id = ui.context.client.id

    def on_change(e: events.ValueChangeEventArguments) -> None:
        state.document['value'] = e.value
        change.emit(e.value)

    editor = ui.codemirror(value=state.document['value'], on_change=on_change)
    change.subscribe(editor.set_value)

    editor.on_value_change(lambda: setattr(state, 'editing_client_id', my_client_id))
    editor.bind_enabled_from(state, 'editing_client_id', lambda client_id: client_id in (None, my_client_id))

    ui.label().bind_text_from(state, 'editing_client_id',
                              lambda client_id: f'Editing client: {client_id}' if client_id else 'No one is editing right now')

    def relinquish():
        if state.editing_client_id == my_client_id:
            state.editing_client_id = None
    ui.context.client.on_disconnect(relinquish)

    ui.button('Relinquish editing', on_click=relinquish)


ui.run(root)
```

Is there even a solid benefit of this PR relaibly demonstratable, then?

falkoschindler:
@evnchn Ah ok, I understand. _Simultaneous_ edits don't always work.

So you think we should simply drop the example? Would be fine for me.

evnchn:
Please definitely drop this example. It will lead to pitchforks. 

I think you don't need collaborative editing to demonstrate the value. Simply implementing a keyboard shortcut to programmatically append/prepend the codemirror text should already demonstrate the value of the PR. 

*has to be keyboard-shortcut since pressing a button = lose the cursor.

evnchn:
`FAILED tests/test_page.py::test_warning_if_response_takes_too_long - selenium.common.exceptions.TimeoutException: Message: timeout`

Flaky test but not related to this PR I suppose.

PR #5780, opened 2026-02-14
Exclude Python prefix directory from reload file watcher

### LINKED ISSUE #5750: Nicegui fork local installation
### First Check

### Example Code

```python

```

### Description

I got a wierd 'first time only error' trying to fork/run the main file, but I thought it might still be helpfull to post it

1) Fork from nicegui
2) Change to sortable branch
3) Run `uv sync`
4) Run `uv run main
5) Error but after that normal bootup of nicegui

[nicegui_startup_issues.txt](https://github.com/user-attachments/files/25204415/nicegui_startup_issues.txt)

This error was only happning the very first time running the main file. Deleting the .venv and trying to reproduce it didnt provoke the error

### NiceGUI Version

3.7.1

### Python Version

3.13.9

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #5750. When running NiceGUI from a local fork/clone with `reload=True` (the default), the first startup can trigger spurious reloads because uvicorn's watchfiles detects changes inside `.venv/Lib/site-packages/` (e.g. from lazy bytecode compilation or package extraction).

The root cause is that uvicorn's `FileFilter` applies exclude patterns via `pathlib.Path.match()`, which only matches the **filename** (last path component). So the default `.*` exclude pattern matches files *named* with a leading dot (e.g. `.hidden.py`), but does **not** exclude files *inside* dot-prefixed directories like `.venv/`. Files like `.venv/Lib/site-packages/IPython/core/display.py` pass right through because `display.py` doesn't match `.*`.

### Implementation

Appends `sys.prefix` to uvicorn's `reload_excludes` list. Uvicorn's `FileFilter` checks each exclude entry with `Path(e).is_dir()` — when it resolves to an actual directory, it adds it to an `exclude_dirs` list and skips all files within that directory tree.

Using `sys.prefix` (rather than the `VIRTUAL_ENV` env var) works regardless of how the virtualenv was entered — activated, `uv run`, direct interpreter path, poetry, etc. It also covers the edge case of someone running `main.py` from within a bare-metal Python installation directory.

### Progress


phifuh:
@evnchn I cant reproduce the error, so it seems to work

PR #5783, opened 2026-02-14
Add reaktiv order calculator example

evnchn:
### Motivation

Addresses https://github.com/zauberzeug/nicegui/discussions/4758#discussioncomment-13175516

Adds a new example demonstrating integration with the [reaktiv](https://github.com/buiapp/reaktiv) reactive state management library. This showcases an alternative approach to state management in NiceGUI using fine-grained reactive signals.

### Implementation

The example is an order calculator with:
- **Signals** for user inputs (price, quantity, tax rate)
- **Computed** values for derived state (subtotal, tax, total) forming a diamond dependency graph
- **Effects** to automatically update NiceGUI labels when any dependency changes

Requires reaktiv >= 0.21.1 which fixes a diamond dependency graph bug ([buiapp/reaktiv#27](https://github.com/buiapp/reaktiv/issues/27)).

### Progress

### DIFF
--- a/examples/reaktiv/main.py
+++ b/examples/reaktiv/main.py
@@ -40,10 +40,15 @@ with ui.row():
             total: float = 0.0
 
             def __post_init__(self) -> None:
+                # multi-input binding for subtotal = price * quantity
                 binding.bind_to(self, 'price', self, 'subtotal', lambda price: price * self.quantity)
                 binding.bind_to(self, 'quantity', self, 'subtotal', lambda quantity: self.price * quantity)
+
+                # multi-input binding for tax = tax_rate * subtotal / 100
                 binding.bind_to(self, 'tax_rate', self, 'tax', lambda tax_rate: tax_rate * self.subtotal / 100)
                 binding.bind_to(self, 'subtotal', self, 'tax', lambda subtotal: self.tax_rate * subtotal / 100)
+
+                # multi-input binding for total = subtotal + tax
                 binding.bind_to(self, 'tax', self, 'total', lambda tax: self.subtotal + tax)
                 binding.bind_to(self, 'subtotal', self, 'total', lambda subtotal: self.tax + subtotal)
 


examples/reaktiv/main.py
```diff
@@ -3,72 +3,56 @@
...
+            tax_rate = 10.0
+            subtotal = 0.0
+            tax = 0.0
+            total = 0.0
+
+            def __init__(self):
+                binding.bind_to(self, 'price', self, 'subtotal', lambda price: price * self.quantity)
+                binding.bind_to(self, 'quantity', self, 'subtotal', lambda quantity: quantity * self.price)
+                binding.bind_to(self, 'tax_rate', self, 'tax', lambda tax_rate: tax_rate * self.subtotal / 100)
+                binding.bind_to(self, 'subtotal', self, 'total', lambda subtotal: subtotal + self.tax)
```
evnchn: Hmm. I don't think this example is complete. 

Owing to how there is no multiple bind-target support for binding, one must bind manually for multi-bind. 

In `subtotal = price * quantity` we bound twice, but we did not bind twice for `total = subtotal + tax`

We would need an extra line here.

evnchn: Ref: https://github.com/zauberzeug/nicegui/discussions/5315#discussioncomment-14748335

falkoschindler: Hm, but why does it work nonetheless?

falkoschindler: I think it is because there was no type annotation for dataclass attributes, which is why they were simple static fields, which is why NiceGUI considered them active links instead of bindable properties. The active polling caused the result to be in sync (slightly delayed).

evnchn: Indeed. It's also the conclusion I came to after artificially inflating `binding_refresh_interval`

evnchn:
Hmm. Should add some comments to group the related bindings. 

And we should give some more thought to `binding.combine` in another PR. 

Perhaps a `strict=True` to disable the polling fallback for non-type-annotation dataclass attributes.

falkoschindler (approved):
Ok, I think this is good enough for now.
Let's merge 👍🏻

PR #5784, opened 2026-02-14
Use Web Worker heartbeat to keep connection alive during blocking dialogs

evnchn:
### Motivation

When blocking browser dialogs (`alert()`, `confirm()`, `print()`) freeze the main JavaScript thread, Socket.IO cannot respond to server pings. This causes the server to consider the client disconnected, eventually deleting the client and forcing a full page reload — losing all user state.

Addresses https://github.com/zauberzeug/nicegui/discussions/2410

### Implementation

- **Web Worker heartbeat** (`nicegui-heartbeat.js`): A lightweight Web Worker runs on a separate thread, sending periodic HTTP POST requests to a new `/_nicegui/heartbeat` endpoint. Since Web Workers are unaffected by main thread blocking, heartbeats continue even during blocking dialogs.
- **Server-side heartbeat handler** (`nicegui.py`): Receives heartbeat POSTs and calls `client._handle_heartbeat()`, which reschedules (cancels and recreates) any pending delete task so the client stays alive while the worker keeps reporting in. If heartbeats stop, the rescheduled delete task still fires after `reconnect_timeout` and cleanup happens normally.
- **Stale `next_message_id` fix** (was in `nicegui.js`): superseded by main's `reconnect_attempt` handler from #6019 — only the heartbeat-related JS remains in this PR after rebase.
- **`reconnect_timeout` passed to client JS** (`client.py`, `index.html`): The heartbeat interval is derived from `reconnect_timeout * 0.5` with a 2 s floor so the worker adapts to the configured timeout without being too aggressive.

### How to test manually

1. Run any NiceGUI app with a default `reconnect_timeout`.
2. Open the page, open DevTools → Console, enable **Preserve log**, and type something into a focused input.
3. Run `alert("blocking")` in the console.
4. Walk away for at least `reconnect_timeout` seconds (default 3 s — try 30 s for confidence).
5. Return and click **OK** to dismiss the alert.
6. Expected: the page does **not** reload; the input value is preserved; reconnect happens silently.
7. Without this PR (i.e. on `main`), the same recipe forces a full page reload after the timeout.

### Progress

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -210,6 +210,7 @@ class Client:
                 'socket_io_js_query_params': socket_io_js_query_params,
                 'socket_io_js_extra_headers': core.app.config.socket_io_js_extra_headers,
                 'socket_io_js_transports': core.app.config.socket_io_js_transports,
+                'reconnect_timeout': self.page.resolve_reconnect_timeout(),
             },
             status_code=status_code,
             headers={'Cache-Control': 'no-store', 'X-NiceGUI-Content': 'page'},
@@ -351,6 +352,11 @@ class Client:
                 self.delete()
         self._delete_tasks[document_id] = \
             background_tasks.create(delete_content(), name=f'delete content {document_id}')
+
+    def _handle_heartbeat(self) -> None:
+        """Cancel pending delete tasks to keep the client alive during blocking browser dialogs."""
+        for document_id in list(self._delete_tasks):
+            self._cancel_delete_task(document_id)
 
     def _cancel_delete_task(self, document_id: str) -> None:
         if document_id in self._delete_tasks:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -79,6 +79,7 @@ class Client:
         self.on_air = False
         self._num_connections: defaultdict[str, int] = defaultdict(int)
         self._delete_tasks: dict[str, asyncio.Task] = {}
+        self._pending_tab_ids: dict[str, str | None] = {}
         self._deleted = False
         self._socket_to_document_id: dict[str, str] = {}
         self.tab_id: str | None = None
@@ -335,7 +336,7 @@ class Client:
         document_id = self._socket_to_document_id.pop(socket_id)
         self._cancel_delete_task(document_id)
         self._num_connections[document_id] -= 1
-        tab_id_to_close = self.tab_id
+        self._pending_tab_ids[document_id] = self.tab_id
         self.tab_id = None
 
         for t in self.disconnect_handlers:
@@ -343,20 +344,33 @@ class Client:
         for t in core.app._disconnect_handlers:  # pylint: disable=protected-access
             self.safe_invoke(t)
 
+        self._create_delete_task(document_id)
+
+    def _create_delete_task(self, document_id: str) -> None:
+        """Schedule deletion of client content after reconnect timeout."""
         async def delete_content() -> None:
             await asyncio.sleep(self.page.resolve_reconnect_timeout())
             if self._num_connections[document_id] == 0:
-                self._num_connections.pop(document_id)
-                self._delete_tasks.pop(document_id)
-                await core.app.storage.close_tab(tab_id_to_close)
+                self._num_connections.pop(document_id, None)
+                self._delete_tasks.pop(document_id, None)
+                tab_id = self._pending_tab_ids.pop(document_id, None)
+                if tab_id is not None:
+                    await core.app.storage.close_tab(tab_id)
                 self.delete()
         self._delete_tasks[document_id] = \
             background_tasks.create(delete_content(), name=f'delete content {document_id}')
 
     def _handle_heartbeat(self) -> None:
-        """Cancel pending delete tasks to keep the client alive during blocking browser dialogs."""
+        """Reset pending delete tasks to keep the client alive during blocking browser dialogs."""
         for document_id in list(self._delete_tasks):
-            self._cancel_delete_task(document_id)
+            self._reschedule_delete_task(document_id)
+
+    def _reschedule_delete_task(self, document_id: str) -> None:
+        """Cancel the current delete task and create a new one with a fresh timeout."""
+        if document_id not in self._delete_tasks:
+            return
+        self._delete_tasks.pop(document_id).cancel()
+        self._create_delete_task(document_id)
 
     def _cancel_delete_task(self, document_id: str) -> None:
         if document_id in self._delete_tasks:
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -195,6 +195,14 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
     return client.build_response(request, 500)
 
 
+@app.post('/_nicegui/heartbeat')
+async def _heartbeat(request: Request) -> Response:
+    data = await request.json()
+    if client := Client.instances.get(data['client_id']):
+        client._handle_heartbeat()  # pylint: disable=protected-access
+    return Response(status_code=200)
+
+
 @sio.on('connect')
 async def _on_connect(sid: str, data: dict[str, Any], _=None) -> None:
     query = {k: v[0] for k, v in urllib.parse.parse_qs(data.get('QUERY_STRING', '')).items()}
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -197,8 +197,11 @@ async def _exception_handler_500(request: Request, exception: Exception) -> Resp
 
 @app.post('/_nicegui/heartbeat')
 async def _heartbeat(request: Request) -> Response:
-    data = await request.json()
-    if client := Client.instances.get(data['client_id']):
+    try:
+        client_id = (await request.json()).get('client_id')
+    except ValueError:
+        return Response(status_code=400)
+    if client_id and (client := Client.instances.get(client_id)):
         client._handle_heartbeat()  # pylint: disable=protected-access
     return Response(status_code=200)
 
--- a/nicegui/static/nicegui-heartbeat.js
+++ b/nicegui/static/nicegui-heartbeat.js
@@ -1,0 +1,18 @@
+let intervalId = null;
+
+self.onmessage = function (e) {
+  if (e.data.type === "start") {
+    const { url, clientId, interval } = e.data;
+    if (intervalId) clearInterval(intervalId);
+    intervalId = setInterval(() => {
+      fetch(url, {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ client_id: clientId }),
+      }).catch(() => {});
+    }, interval);
+  } else if (e.data.type === "stop") {
+    if (intervalId) clearInterval(intervalId);
+    intervalId = null;
+  }
+};
--- a/nicegui/static/nicegui-heartbeat.js
+++ b/nicegui/static/nicegui-heartbeat.js
@@ -9,7 +9,7 @@ self.onmessage = function (e) {
         method: "POST",
         headers: { "Content-Type": "application/json" },
         body: JSON.stringify({ client_id: clientId }),
-      }).catch(() => {});
+      }).catch((err) => console.debug("heartbeat failed:", err));
     }, interval);
   } else if (e.data.type === "stop") {
     if (intervalId) clearInterval(intervalId);
--- a/nicegui/static/nicegui-heartbeat.js
+++ b/nicegui/static/nicegui-heartbeat.js
@@ -1,26 +1,28 @@
-let intervalId = null;
-let inFlight = false;
+let timeoutId = null;
+
+function scheduleNext(url, clientId, interval) {
+  // jitter ±25% so synchronously-loaded clients don't stampede the server
+  const jittered = interval * (0.75 + Math.random() * 0.5);
+  timeoutId = setTimeout(() => {
+    fetch(url, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ client_id: clientId }),
+    })
+      .catch((err) => console.debug("heartbeat failed:", err))
+      .finally(() => {
+        if (timeoutId !== null) scheduleNext(url, clientId, interval);
+      });
+  }, jittered);
+}
 
 self.onmessage = function (e) {
   if (e.data.type === "start") {
     const { url, clientId, interval } = e.data;
-    if (intervalId) clearInterval(intervalId);
-    intervalId = setInterval(() => {
-      if (inFlight) return; // skip while a previous heartbeat is still pending
-      inFlight = true;
-      fetch(url, {
-        method: "POST",
-        headers: { "Content-Type": "application/json" },
-        body: JSON.stringify({ client_id: clientId }),
-      })
-        .catch((err) => console.debug("heartbeat failed:", err))
-        .finally(() => {
-          inFlight = false;
-        });
-    }, interval);
+    if (timeoutId !== null) clearTimeout(timeoutId);
+    scheduleNext(url, clientId, interval);
   } else if (e.data.type === "stop") {
-    if (intervalId) clearInterval(intervalId);
-    intervalId = null;
-    inFlight = false;
+    if (timeoutId !== null) clearTimeout(timeoutId);
+    timeoutId = null;
   }
 };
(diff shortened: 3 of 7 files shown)


evnchn:
Testing method:

- Open console
- Turn on Preserve log
- Run `alert()`
- Have lunch
- Return to press OK. 
- The page should NOT reload.

evnchn:
**What would happen** if we moved the entire Socket.IO management into a Web Worker? 

When the main thread is blocked, what happens to the messages though...

phifuh:
I can confirm this behaviour :d We had a simple rename javascript dialog and just waiting a minute or so disconnected the client. Our solution was simple, just do an async real ui.dialog() and await it. 

Is it possible to add an async `add_js_async` element? Currently we are just injecting js code if i am not mistaken? Maybe this could be some sort of bridge like your web-worker idea

evnchn:
@phifuh On your `add_js` idea:

- Async JS code will not salvage total-JS-blocking stuff like `dialog()`. If it did, we would not need a Web Worker.
- As of writing NiceGUI offers `ui.run_javascript` which does not support getting the value of async JS*. Nevertheless, well-written userland JS can process user input asynchronously and submit the result to the server via `emitEvent` once available. 

*: `await ui.run_javascript('await new Promise(resolve => setTimeout(() => resolve("Hello World"), 100));')` returns `None`...

Overall, I am not inclined to add another JS execution means beside `ui.run_javascript`. Expanding its functionality to better deal with async JS may be a good idea though, but that's for another PR.

evnchn:
@falkoschindler If memory serves, none of the files touched here is actually changed in what's merged in 3.10, so this is rather good candidate for 3.10 inclusion. 

I'd imagine it's good to "change all files, but each just a little" over "10 PRs stabbed at a particular file" in one release because it gives us easier bisection for free.

falkoschindler (changes requested):
Thanks for this PR — the Web Worker approach is a clever way to work around the main thread freeze.

I haven't done a detailed manual review yet because I had Claude look at it first, and it flagged a number of issues that seem worth addressing before I dig deeper. Posting them below so you can take a pass at them.

It would also be great if the PR description included a short "how to test manually" section (e.g., open a page, trigger `alert()`, wait past the reconnect timeout, dismiss, and verify state is preserved).

1. **Heartbeat cancels delete tasks permanently — client may never be cleaned up**

   `_handle_heartbeat()` cancels all pending delete tasks but does **not** reschedule them. If the browser is truly gone (e.g., crashed with the worker still running briefly, or a misbehaving client), the delete task is cancelled and never recreated. The client instance will leak in `Client.instances` indefinitely — no mechanism re-creates the delete task after a heartbeat.

   The heartbeat should **reset** the timer rather than cancel it outright. Consider replacing the cancel-only approach with a "reschedule" pattern: cancel the current delete task and create a new one that sleeps for another `reconnect_timeout` period. This way, if heartbeats stop, cleanup still happens.

   ```python
   def _handle_heartbeat(self) -> None:
       """Reset pending delete tasks to keep the client alive during blocking browser dialogs."""
       for document_id in list(self._delete_tasks):
           self._reschedule_delete_task(document_id)
   ```

2. **Heartbeat worker never stops on normal page unload**

   The worker is started on page load but there is no `beforeunload` or `pagehide` listener to send a `stop` message to the worker. While the worker is garbage-collected when the page is destroyed in most browsers, this is not guaranteed (especially with bfcache). Adding a cleanup listener would be defensive and cheap:

   ```js
   window.addEventListener("pagehide", () => {
     if (window.heartbeatWorker) {
       window.heartbeatWorker.postMessage({ type: "stop" });
       window.heartbeatWorker.terminate();
     }
   });
   ```

3. **On Air compatibility not verified**

   With On Air, the browser talks to a relay server. The heartbeat POST goes to `window.location.origin` (the relay), which needs to proxy it back to the NiceGUI server via the `_handle_http` handler in `air.py`. This _should_ work since the relay proxies arbitrary HTTP requests, but it hasn't been tested. Please manually verify heartbeat behavior with `on_air=True` and confirm it works.

4. **Heartbeat URL doesn't account for reverse proxy / non-origin deployments**

   The heartbeat URL is built as `window.location.origin + options.prefix + ...`, but `window.location.origin` may not be correct behind a reverse proxy with path rewriting. Other NiceGUI network calls go through Socket.IO (which uses the configured path), so they don't have this issue. Consider using a relative URL instead:

   ```js
   url: `${options.prefix}/_nicegui/heartbeat`,
   ```

   `fetch()` with a relative URL will use the page's origin automatically and works correctly behind proxies.

5. **Heartbeat interval may be too aggressive for large deployments**

   With `reconnect_timeout=3.0`, the heartbeat fires every 1.5 seconds. For apps with many concurrent clients, this adds significant HTTP overhead. The minimum of 0.5s seems very low. Consider a higher floor (e.g., 2-3s minimum) and document the tradeoff.

6. **Tests use long `time.sleep()` calls (15s)**

   `test_connection_survives_alert_with_high_reconnect_timeout` sleeps for 15 seconds in a single test. This will significantly slow down the test suite. Consider reducing `reconnect_timeout` and `ping_interval`/`ping_timeout` for this test to keep the sleep under 5s, or mark the test as `@pytest.mark.slow` if such a marker exists.

7. **`window.heartbeatWorker` pollutes the global namespace**

   Consider using a local variable inside `createApp` instead of attaching to `window`, unless it needs to be accessed externally (e.g., for debugging). If it does, document why.

8. **`.catch(() => {})` silently swallows all fetch errors in the worker**

   This is understandable (heartbeat failures are expected during shutdown), but a `console.debug` would help with debugging connectivity issues during development.

9. **Test uses `counter.__setitem__` lambda hack**

   In `test_connection_survives_alert_with_high_reconnect_timeout`, this pattern is unnecessarily obscure:

   ```python
   lambda: label.set_text(str(counter.__setitem__('value', counter['value'] + 1) or counter['value']))
   ```

   A simple helper function or `nonlocal` variable would be clearer.

evnchn:
Moving to 3.12 — coordinated with Falko, this needs more work and shouldn't block the 3.11 release.

evnchn:
**TL;DR:** Went through all 9 of @falkoschindler's points — 6 were already done in `4baede0e`, 2 were done in two new commits (`08775d33`, `fd26d4db`), and 1 (On Air) is code-path-verified-only since the relay is Zauberzeug-internal. PR description updated with a "How to test manually" section.

<details><summary>Per-point status table</summary>

| # | Status | Notes |
|---|---|---|
| 1 — heartbeat resets, doesn't permanently cancel | ✅ Done in `4baede0e` | `_handle_heartbeat()` now calls `_reschedule_delete_task()`, which cancels the current delete task and creates a new one with a fresh `reconnect_timeout`. If heartbeats stop, cleanup still fires. |
| 2 — `pagehide` cleanup | ✅ Done in `4baede0e` (and tightened in `08775d33`) | Listener registered on `pagehide`. |
| 3 — On Air verification | ⚠️ Not tested live (no public On Air relay) — code-path looks fine | `air.py:_handle_http` is a generic ASGI proxy with no path filter, so a POST to `/_nicegui/heartbeat` is forwarded transparently to the device. Happy to spin up a verification on the relay if Zauberzeug can grant access. |
| 4 — relative URL behind reverse proxy | ✅ Done in `4baede0e` | `${options.prefix}/_nicegui/heartbeat`. |
| 5 — interval floor too low | ✅ Done in `4baede0e` | Floor raised from 0.5 s to 2 s. With the default `reconnect_timeout=3.0`, heartbeat interval = max(1.5 s, 2 s) = 2 s. |
| 6 — long `time.sleep(15)` in tests | ✅ Done in `fd26d4db` | Two test sleeps shrunk: `0.5 + 5.0 → 3.5` s each, plus `screen.wait(3.0) → 1.5` s in the message-flood test (timer 50 ms → 20 ms keeps the prune). Net ~10 s saved across the file. The 3.5 s figure is "past `reconnect_timeout` but past at least one heartbeat refresh too". |
| 7 — `window.heartbeatWorker` pollutes globals | ✅ Done in `08775d33` | Now a local `const heartbeatWorker` inside `createApp`. Dropped the `test_heartbeat_worker_is_started` smoke test that probed `window.heartbeatWorker` — the survival tests already prove the worker is running (they would fail without it). |
| 8 — `.catch(() => {})` swallows errors | ✅ Done in `4baede0e` | Now `.catch((err) => console.debug("heartbeat failed:", err))`. |
| 9 — `counter.__setitem__` lambda | ✅ Done in `4baede0e` | Refactored to `counter = [0]` + `increment()` helper. |

PR description updated with a "How to test manually" section, and the conflict against `main` resolved (the disconnect-handler change was now redundant given #6019's `reconnect_attempt` handler — dropped that hunk during rebase, so only the heartbeat-specific JS remains).

Pushed as two new commits on top of `4baede0e` (no force-push). Happy to squash before merge.

</details>

evnchn:
**TL;DR:** Force-pushed a clean rebase onto current `main` (`fb8edc21`) to clear the `CONFLICTING` status. SHAs `c0eb1e92..fd26d4db` → `07e4e96b..733b08ba`. Three of five commits are byte-identical; the other two only have shifted unified-diff context. No commits squashed, dropped, or reordered. Used `--force-with-lease`. Happy to drop the rebase if you'd rather resolve at merge time.

<details><summary>What changed and proof</summary>

- The conflict in `nicegui/static/nicegui.js` was the disconnect-handler `options.query.next_message_id = window.nextMessageId;` line from the original PR. Main's new `reconnect_attempt` handler (#6019) already does the same job, so I dropped that hunk during the rebase. Net behavior unchanged — `next_message_id` still gets refreshed before reconnect, just via the new upstream handler.
- Everything else replayed cleanly. No commits were squashed, dropped, or reordered.

Verifiable via `git range-diff`:

```
$ git range-diff 877b2b1c..fd26d4db fb8edc21..733b08ba
1:  c0eb1e92 ! 1:  07e4e96b Use Web Worker heartbeat to keep connection alive during blocking dialogs
2:  3ab501c4 = 2:  3ba1ebf9 Add heartbeat endpoint to endpoint documentation tests
3:  4baede0e = 3:  440d190b Address review feedback: reschedule delete tasks, pagehide cleanup, relative URL, raise interval floor
4:  08775d33 ! 4:  198d0bd0 Scope heartbeat worker to a local const (review #7)
5:  fd26d4db = 5:  733b08ba Trim heartbeat test sleep durations (review #6)
```

Three commits are byte-identical (`=`); the two flagged with `!` differ only in surrounding context — the heartbeat block now sits below the new `reconnect_attempt` handler instead of below the `transports:` line. Same lines of code added; only the unified-diff context shifted.

Apologies for the force-push; the conflict was blocking review and the rebase is purely mechanical.

</details>

evnchn:
**TL;DR:** Five new commits (`a3a1bb41`..`3c79dfac`), one per Copilot finding. Took #1, #2, #4, #5 mostly verbatim. For #3 I took the jitter idea but pushed back on the opt-in/configurable suggestion. The big one is **#5** — Copilot caught that the previous survival tests were passing trivially because `ping_interval+ping_timeout` (≥6s) means `_delete_tasks` doesn't even populate during a 3.5s alert, so the heartbeat had nothing to refresh. Fixed; tests are slower but now actually exercise the heartbeat.

<details><summary>Per-finding response</summary>

| # | Copilot finding | My take | Commit |
|---|---|---|---|
| 1 — `nicegui.py:201` JSON / missing `client_id` raises 500 | Valid, minor. Heartbeat is fire-and-forget; a 500 with stack trace is just log noise on a malformed POST. Wrapped `request.json()` in `try/except ValueError` (returns 400) and used `.get('client_id')` so a missing key is a no-op rather than a `KeyError`. | `a3a1bb41` |
| 2 — `nicegui-heartbeat.js:16` overlapping fetches | Valid. Originally added an explicit `inFlight` flag, then realized commit 3's switch to recursive `setTimeout` (next tick scheduled from `.finally`) makes overlap impossible by construction. The flag would have been dead code, so it was dropped in `7ed881b4`. Net guarantee is the same. | `63b29b1e` (superseded by `7ed881b4`) |
| 3 — `nicegui.js:438` steady-state load: reduce frequency, add jitter, or make opt-in | Took the **jitter** part (±25%) — cheap insurance against synchronized stampedes after a deploy or LB restart. **Declined "make opt-in"**: the heartbeat is a no-op server-side when `_delete_tasks` is empty (which is the steady state on a healthy connection — `_handle_heartbeat` loops over an empty dict), so cost is one HTTP round-trip and a method call. Putting it behind a flag adds API surface for negligible gain. **Declined "lower frequency"**: the floor is already at 2s per Falko's earlier review and must remain shorter than `reconnect_timeout` for the worker to refresh the delete task before it fires. | `7ed881b4` |
| 4 — test name vs body inconsistent | Yes. Renamed to `test_connection_survives_alert_after_many_messages` and dropped the misleading docstring. Also stripped docstrings + most comments from this test file to match the rest of `tests/` (looks like the codebase convention is "function name is the documentation"). | `11b1cc1a` |
| 5 — fixed-sleep test passes without heartbeat | **Sharp catch and the most important one.** With the default `reconnect_timeout=3.0`, `ping_interval = max(2.4, 4) = 4s` and `ping_timeout = max(1.2, 2) = 2s` (set in `nicegui.py:134-135`). So socket.io needs ~6s to register the disconnect, and the previous 3.5s sleep meant `_delete_tasks` was empty the entire time. Tests would have passed without any heartbeat code. Fixed: poll `Client.instances[client_id]._delete_tasks` becoming non-empty as a precondition (proves the disconnect is real), then sleep past `reconnect_timeout` and assert the client is still alive. The third test gets a longer post-navigate sleep for the symmetric reason. Per-test runtime grows to ~10s; the only honest alternative was to make the test wrong. | `3c79dfac` |

</details>

Pushed as a fast-forward on top of `733b08ba` — no rebase, no force. Open to dropping any of these if the trade-offs read differently to you.

phifuh:
@evnchn Quick question; is it possible to add a fallback logging in case of connection timeout with can add a helpfull diagnostic where the blocking event happend or what else caused the disconnect?

falkoschindler:
@evnchn — thanks for working through the review rounds on this. Before we go further on the merge path I'd like us to weigh the heartbeat against an alternative I sketched in [#2410](https://github.com/zauberzeug/nicegui/discussions/2410) (beacon-on-pagehide instead of a heartbeat worker). Smaller permanent surface, no steady-state HTTP traffic, slower zombie-cleanup as the trade-off. The design question feels load-bearing enough to settle first — if we pivot to the beacon, most of the implementation review becomes moot.

evnchn:
We still haven't got time to decide which, so let's handle this later. 

Also, a retrospect: blocking dialogs would block since our first NiceGUI release (to be frank, it messes up ALL web frameworks), and the reports are all in #2xxx era (we are now into the 6000s), so maybe we can tuck `alert()` usage into "skill issue" which is a non-issue as even the most basic LLM can tell you it is not the right thing to do.

evnchn:
*Closing note drafted by Claude Code, working with @evnchn.*

**TL;DR:**

**Closing** — weighed again, the always-on heartbeat isn't worth carrying against current demand.

The one genuinely non-skill-issue case is **`window.print()`** — a legitimate, unavoidable native blocking dialog (unlike `alert()`/`confirm()`, which are better replaced with `ui.dialog`). But demand is stale, half of it is already superseded, and the steady-state heartbeat is a permanent surface a lighter approach avoids.

If print-during-session state-loss resurfaces as a fresh report, the cleaner path is Falko's **beacon-on-pagehide** sketch (discussion #2410), not this always-on worker. Happy to revive from there.

<details><summary><b>Why close, in full</b></summary>

- **Scope of the value:** the PR keeps the socket alive when a blocking browser dialog freezes the main JS thread. `alert()`/`confirm()` are avoidable (use `ui.dialog`) → user-side skill issue. **`window.print()` is not avoidable** — printing is a legitimate feature that unavoidably uses a native blocking dialog. That is the only non-skill-issue kernel here.
- **Demand:** the motivating reports are all from the #2xxx-era; nothing recent in the current #6xxx-era.
- **Already partly superseded:** the stale `next_message_id` fix folded into main's `reconnect_attempt` handler (#6019); only the heartbeat JS remained.
- **Cost:** an always-on Web Worker POSTing to `/_nicegui/heartbeat` is permanent steady-state traffic + a new endpoint surface.
- **Lighter alternative:** beacon-on-pagehide (#2410) — smaller permanent surface, no steady-state traffic (trade-off: slower zombie cleanup).
- **Review state:** all of the Apr-1 change-requests and Copilot's 5 inline comments were addressed and CI is green on the head — this is a scope/architecture close, not an unresolved-work close.

**Revival trigger:** a fresh, current-era report of `print()`-during-session dropping the connection → revisit via the beacon approach.
</details>

PR #5785, opened 2026-02-14
Make ValueElement and ValueChangeEventArguments generic

evnchn:
### Motivation

Closes #2677.

Type checkers currently infer `Any` for `.value` on every `ValueElement` subclass and for `e.value` / `e.previous_value` in change handler callbacks. This makes static analysis ineffective for the most commonly accessed property in NiceGUI.

This PR makes `ValueElement` generic (`ValueElement[V]`) and `ValueChangeEventArguments` generic (`ValueChangeEventArguments[ValueT]`) so that type checkers can properly infer value types — e.g. `ui.input().value` → `str`, `ui.slider().value` → `float | None`, `ui.switch().value` → `bool | None`.

### Implementation

**Commit 1 — Make `ValueElement` generic with TypeVar `V`:**

- Add `V = TypeVar('V')` and `Generic[V]` base class to `ValueElement`
- Use a `TYPE_CHECKING`-only `value: V` annotation to override the `BindableProperty` descriptor for type checkers (the descriptor continues to work at runtime)
- Update all method signatures (`set_value`, `_handle_value_change`, `_event_args_to_value`, `bind_value_to/from`, etc.) to use `V`
- Make `ValidationElement` propagate `V` (`ValidationElement(ValueElement[V])`)
- Pin `ChoiceElement` to `Any` since choice values are runtime-dependent
- Parameterize all ~35 leaf element subclasses with their concrete value types

**Commit 2 — Make `ValueChangeEventArguments` generic with TypeVar `ValueT`:**

- Add `ValueT = TypeVar('ValueT')` and `Generic[ValueT]` to the dataclass
- Type `value` and `previous_value` fields as `ValueT`
- Update all `Handler[ValueChangeEventArguments]` annotations across all elements to carry the correct type parameter

**Key design decisions:**

- `BindableProperty` is **not** made generic — Python's descriptor protocol cannot infer the return type from the owner's generic parameter. The `TYPE_CHECKING` annotation override is the standard workaround.
- Elements with complex/heterogeneous value types (`Date`, `Tabs`, `Carousel`, `Stepper`, `Select`, etc.) use `Any` as their type parameter, which preserves the current behavior without making incorrect type claims.
- This is a **typing-only change** — `Generic[V]` has no runtime effect, and the `TYPE_CHECKING` block is never executed.

### Progress

### DIFF
--- a/nicegui/elements/tabs.py
+++ b/nicegui/elements/tabs.py
@@ -64,10 +64,10 @@ class Tabs(ValueElement[str | Tab | TabPanel | None]):
         """
         super().__init__(tag='q-tabs', value=value, on_value_change=on_change)
 
-    def _value_to_model_value(self, value: Any) -> str | None:
+    def _value_to_model_value(self, value: Any) -> Any:
         return value.props['name'] if isinstance(value, (Tab, TabPanel)) else value
 
-    def _value_to_event_value(self, value: Any) -> str | None:
+    def _value_to_event_value(self, value: Any) -> Any:
         return self._value_to_model_value(value)
 
 
@@ -102,8 +102,8 @@ class TabPanels(ValueElement[str | Tab | TabPanel | None]):
         self._props.set_bool('animated', animated)
         self._props.set_bool('keep-alive', keep_alive)
 
-    def _value_to_model_value(self, value: Any) -> str | None:
+    def _value_to_model_value(self, value: Any) -> Any:
         return value.props['name'] if isinstance(value, (Tab, TabPanel)) else value
 
-    def _value_to_event_value(self, value: Any) -> str | None:
+    def _value_to_event_value(self, value: Any) -> Any:
         return self._value_to_model_value(value)


evnchn:
### Why `BindableProperty` isn't made generic (and why that's fine)

Ref: https://github.com/zauberzeug/nicegui/discussions/3988

**TL;DR:** Python's descriptor protocol can't infer the return type from the owner class's generic parameter. The `TYPE_CHECKING` annotation override used in this PR is the standard workaround (same pattern as SQLAlchemy, Django, Pydantic). No Python PEP currently addresses this gap.

<details>
<summary>Full analysis</summary>

### The problem

`BindableProperty` is a descriptor. Its `__get__` signature is:

```python
def __get__(self, owner: Any, owner_cls=None) -> Any
```

For it to return `V` (the value type), it would need to know that it lives on a `ValueElement[V]` and that `V` is bound to e.g. `float | None` for `Slider`. Python's type system has no way to express "return the generic parameter of my owner class."

### Approaches considered

**Approach A — `BindableProperty(Generic[T])` with `__get__` overloads:**

```python
class BindableProperty(Generic[T]):
    @overload
    def __get__(self, owner: None, owner_cls: type) -> Self: ...
    @overload
    def __get__(self, owner: Any, owner_cls: type) -> T: ...
```

Problem: `T` is fixed at descriptor *definition* time, not at owner *class* time. You'd need `value = BindableProperty[float | None](...)` in every element class, defeating the purpose of inheriting the type from the class generic.

**Approach B — Per-class `__get__` return type via `__class_getitem__`:**

Not supported. Descriptors can't conditionally return different types based on the class they're accessed from.

**Approach C — Protocol-based descriptor typing (PEP 688 / future):**

Some type checkers (pyright) have experimental heuristics, but nothing is standardized and it doesn't work with class-level generic parameters.

### What we did instead (Approach D)

```python
value = BindableProperty(on_change=...)
if TYPE_CHECKING:
    value: V  # type: ignore[assignment]
```

This gives correct type inference in all type checkers, has zero runtime cost, and is one line. The `# type: ignore` is the only cosmetic downside.

### Difficulty assessment

| Factor | Assessment |
|--------|------------|
| Conceptual complexity | Very high — "descriptor + owner generic" inference is unsolved |
| Python type system support | Insufficient — no PEP covers this |
| Type checker support | Partial (pyright has heuristics, mypy doesn't) |
| Scope if attempted | 15+ mixin classes with their own `BindableProperty` |
| Benefit over current workaround | Marginal — `TYPE_CHECKING` override already works |

### Recommendation

Don't pursue. A proper solution requires a future PEP for descriptor-generic inference. The same `TYPE_CHECKING` pattern could optionally be applied to other `BindableProperty` usages (`text: str`, `label: str`, `visible: bool`, `enabled: bool`, etc.) in a follow-up PR for completeness.

</details>

evnchn:
I'll sit until 3.9 until getting around to fix the mypy. 

If anyone want to take a shot, feel free to do so (open a PR against my branch, I suppose)

evnchn:
Yeah indeed this is OK for review. I forgot that I asked copilot to fix the mypy errors.

evnchn:
In 4.0, could we rid the unions by narrowing the API?

PR #5791, opened 2026-02-16
ui.tiptap concurent multi editing with WYSIWYG editor

phifuh:
### Motivation

With inspiration of #5775 and as discussed in #5774 there is a usecase for having a WYSIWYG editor with realtime collab. @falkoschindler gave the hint to maybe look at tiptap with py-y which this PR introduces as new element. 

### Implementation

y-py is optional and only needed for the state management, maybe it should be added as core tho?
A few extra extensions got added to extend the tiptap js starterkit most notable `@tiptap/extension-underline`
This can be extended in the future

### Progress

--- 

## Comments: 

There is at least one minor outstanding issue which I am stuck on and thats the handling of threads in nicegui and garbage collection. 
E.x when running `uv run pytest test/test_tiptap.py` i get a warning `RuntimeError: y_py::y_doc::YDoc is unsendbale, but is dropped on another thread!`
I even asked my popular friend `claude` with no luck, so I am hoping that @evnchn  @falkoschindler  that you can have a look at it , since you guys no a lot more about the internal wiring. 

Another issue was trying to run it on a different machine, which resultet in some y-py dependency issue which I havent looked into further but that might just be a pyproject.toml setup issue

I am not sure if there are any other short comming, I tested it on a render.com server to be sure with a little demo and that worked execelent.

---

### AI summary:

This pull request introduces a collaborative rich-text editor element (`Tiptap`) to NiceGUI, powered by Tiptap and Yjs for real-time editing and CRDT-based document merging. It adds both frontend and backend support for collaborative editing, state persistence, and awareness/cursor sharing. The changes also integrate the new element into NiceGUI's import system and documentation.

**Collaborative editor integration:**

* Added the `Tiptap` element, a collaborative rich-text editor using Tiptap and Yjs, including state persistence and awareness features. 
* Implemented a server-side Yjs room manager (`tiptap_room.py`) to handle document state, client synchronization, and awareness messages.
* Registered and cleaned up Yjs event handlers in the main NiceGUI server lifecycle, ensuring proper room management and client disconnection handling.

**Frontend bundling and exports:**

* Added a Node.js package (`package.json`, `rollup.config.mjs`, `src/index.mjs`) to bundle Tiptap core, extensions, and Yjs-related modules for frontend usage. 

**NiceGUI integration and documentation:**

* Registered the `Tiptap` element in NiceGUI's UI import system and documentation sections.


evnchn:
For a first-time PR, this is very impressive! You don't have to aim for perfection right away (this is likely scheduled for 3.9 if not 3.10) and **please do not take me as inspiration (it's 4AM where I live right now)**

I'd assume AI helped you in the process. It's too late to add AI co-authorship though (prefer not to force-push when PR is open). If you have any friction points in the process, please be sure to voice out, as we are about to revamp the `AGENTS.md`.

phifuh:
Thank you, my friend Claude did a lot of the heavy lifting. My first little “baby” script was written back in Python  2.5, and I’ve just been promoted to senior developer at work. It’s crazy to see how far agent‑style coding has come tho in the past year.

I mention this because I am very interested in hybrid intelligence. At my current job I see junior developers struggling to handle the direction of AI, mainly because they lack the necessary knowledge, company insights and experience and thats an issue because AI is in the driving seat without the needed supervision and direction in the steeringwheel. 

I feel like this PR/draft is already far along—not perfect, but good enough as a draft. I wouldn’t have tackled this PR if it weren’t for Claude; it took me the whole weekend, and the amount of time I would have needed to do everything manually would have been insane.

evnchn:
Just to be clear and set the record straight for other people and possibly AI Agents reading: **Please try to ensure you are familiar with NiceGUI before opening a PR.** IMO not that you _must_, but there will be a lot of back-and-forth otherwise (not sure for other maintainers, they reserve the right to close your PR)

@phifuh, with 11 issues and 6 discussions, you are well-past-the-mark on that one 💪

evnchn:
Hello @phifuh. Happy Lunar New Year!

It's been a busy few days. I don't think I have the time to look at this in detail just yet, but I think I should point out 2 obvious things right away: 

#### Mysterious error `RuntimeError: y_py::y_doc::YDoc is unsendbale, but is dropped on another thread!` at the end of passing pytest

**I am positively sure** that comes from `y-py` or its dependencies because NiceGUI uses `asyncio` which is single-threaded concurrency, while `y-py` calls Rust and it definitely has to be on another thread. 

We run `gc.collect()` in the test fixture, and previously we have leaked stuff which complains when being GCed and the fixture's GC catches that. The fact that tests passes means that message has nothing to do with garbage collection. Ref: https://github.com/zauberzeug/nicegui/blob/ab47f68040f87a436f90fe9a720a62228d114295/nicegui/testing/general.py#L61-L66

#### Library selection and usage

Speaking of `y-py`, [it is abandoned](https://github.com/y-crdt/ypy). Can we switch to maintained alternative such as https://github.com/y-crdt/pycrdt which it linked? 

Nevertheless, that also invokes Rust, so if there's no [prebuilt wheel](https://pypi.org/project/pycrdt/#files) then stuff gets complicated real quick. We should strive to not have `y-crdt` as a hard-requirement of NiceGUI as such. Ref: https://github.com/zauberzeug/nicegui/blob/ab47f68040f87a436f90fe9a720a62228d114295/nicegui/optional_features.py#L1-L27

phifuh:
I always thought you were from Norway or Sweden, but the other day when you mentioned its 4AM it was clear you are currently far in the east :D  
happy New Year tho! May the Year of the Fire Horse be a lucky charm for the coming year.

I tracked the pytest that raises the warning  and it seems to be related to `screen.open('/')` which makes sense because I would assume it opens a new thread inside pytest. I’ll try the garbage‑collector controller and see if that fixes it.

I don’t see any reason not to switch to y‑crdt. 
From a user perspective, the most important thing is to make it clear that this adds an extra dependency and to document that accordingly.

phifuh:
@evnchn 

Moving from y-py to py_crdt fixed the `y_py::y_doc::YDoc is unsendbale, but is dropped on another thread!` but it could be that the other libary just doesnt have this specific warning, so I am unsure how to debug/profile if the gc is doing its job.

---

Changlog:

  1. pyproject.toml:
    - Updateded dependencies
    
  2. nicegui/tiptap_room.py:
    - Renamed HAS_Y_PY → PYCRDT_AVAILABLE    
    - Updated imports
    - Updated API calls:
    - Updated docstrings and comments
    
  3. tests/test_tiptap.py:
    - Updated imports to use from pycrdt import Doc, Map
    - Updated API calls in tests for transactions and map access
    - Renamed test functions to reflect pycrdt naming
    
  4. website/documentation/content/tiptap_documentation.py:
    - Updated doc about optional pycrdt installation and Toolbar()
  
  5. Added a new Toolbar class to be more explicit about creation of the toolbar and for future extion like `preset`, `sticky`, `position`, `style` or what ever. So calling ui.tiptap() will return a default toolbar, while toolbar=None is hiding it
  
  ---
  
  I guess I could add another test about the toolbar but I am done for today

evnchn:
*Drafted by Claude Code working with @evnchn — checking in on the PR after the Feb pause and asking @falkoschindler for a scope read before more engineering goes in.*

Hey @phifuh — sorry for going dark on this. The PR is in good shape for what it is, but I think before either of us spends more cycles on conflicts/CI/wiring, it's worth a scope check from @falkoschindler, because the answer changes which fixes are even worth doing.

@falkoschindler — quick direction question: does a collab-rich-text editor (Tiptap + Yjs/pycrdt, server-side room manager in `nicegui/tiptap_room.py`, ws lifecycle hooks in `nicegui.py`) belong in core, or is this better as a separate extension package? My read of the cost:

<details><summary>Cost / risk breakdown</summary>

- **New runtime dep on `pycrdt`** (Rust-built CRDT lib). It has prebuilt wheels, but adds platform-matrix surface. Has to go through `optional_features.py` either way.
- **Server-side state** — `tiptap_room.py` is 176 LOC of room/awareness management plus ws lifecycle hooks in `nicegui.py`. Unlike most `ui.*` elements this is *not* stateless on the server.
- **Bundled JS** — `nicegui/elements/tiptap/` ships a `package.json` + rollup build for Tiptap+Yjs (~1900 LOC of `package-lock.json`). First element in tree with its own JS build step that I can think of (codemirror ships prebuilt).
- **Scope precedent** — once collab is in core, expectation for collab on `ui.codemirror`, `ui.editor`, etc. follows.

None of these are dealbreakers — phifuh's implementation is clean — but they're the kind of decisions that are cheaper to make now than after merge.

</details>

If you greenlight core inclusion, I'm happy to pitch in on the small, well-scoped pieces: `optional_features.py` wiring for `pycrdt`, the merge conflict against `main`, and the three red CI gates (pre-commit / pylint / quick-test-gate). If you'd rather see it as an extension package, that's a different conversation phifuh and I can take elsewhere.

Either way, @phifuh — the work itself is good and the pycrdt swap was the right call.

evnchn:
I think the previous response is more towards where to put the completed code, in NiceGUI or in an external package. However, assuming a target audience exists for the full-form `ui.tiptap` with WISYWYG, the course of action is to get this PR's code to be polished and done. 

Let me take a shot at task division instead.

evnchn:

Took a fresh review pass through the latest commits as the "shot at task division" promised earlier today. Here's what came back, organized by category. The collaborative-editing scaffolding is solid: the pycrdt migration is clean, the `_initialSyncComplete` + `_hasReceivedServerState` initial-sync race handling deliberately avoids the classic Yjs duplicate-state-on-reconnect trap, `LOOPBACK = None` is the right opt-out because Yjs owns canonical state, and `set_state` recreating the Doc (rather than CRDT-merging) is the only correct way to restore deletions. The scope you've delivered for a draft PR is genuinely impressive — toolbar + collab cursors + persistence demo with debounce, all working.

A handful of items remain. Most are NiceGUI-integration conventions, a few are multi-user safety items, and two are open design questions worth surfacing to @falkoschindler before any code lands.

---

**Items I'm happy to handle** if useful — these are the repetitive NiceGUI-integration plumbing I've already wired into other elements (echart, codemirror), so they're faster for me than for someone new to the codebase, and offloading them frees you for the substantive parts:

- **C2** Defer `tiptap_room.setup()` until first `Tiptap.__init__` (lifecycle pattern; touches `nicegui.py`)
- **C5** Walk-through value-pipeline end-to-end + remove `_event_args_to_value` if dead
- **I1** Awareness-removal broadcast on client disconnect (wire into Socket.IO disconnect hook)
- **I3** `asyncio.to_thread` for large pycrdt operations + Socket.IO binary frames
- **I7** Real multi-user E2E test (two browser sessions, shared `doc_id`, edit-and-assert-propagation)
- **I10** Single shared `yjs_*` dispatcher per page (replace per-instance `window.socket.on`)
- **M4** Replace `setTimeout(100)` with explicit `yjs_init` ack

You stay driver. I'll send these as commits against your branch if you'd like, or drop diff sketches in comments for you to apply — whichever feels less imposing. Just let me know.

---

<details><summary>Convention compliance (NiceGUI patterns)</summary>

**C1 — `optional_features` registration missing**
`pyproject.toml` correctly adds `pycrdt` as an optional extra, but `nicegui/elements/tiptap/tiptap.py` doesn't call `optional_features.register('pycrdt')`. Two steps: (a) extend `optional_features.py`'s `FEATURE` Literal (currently lists 11 names — altair, anywidget, highcharts, matplotlib, pandas, pillow, plotly, polars, pyecharts, redis, webview — `pycrdt` would need to be added before any `register('pycrdt')` call would type-check), (b) call `optional_features.register('pycrdt')` at the top of `tiptap.py`. Compare `altair.py:10`, `highchart.py:7`, `image.py:14`, `pyplot.py:18`, `echart/echart.py:24`. This is the headline convention compliance ask from the earlier round.

**C2 — `tiptap_room.setup()` runs at module import for all NiceGUI users**
`nicegui/nicegui.py:55-56` imports and calls `setup()` unconditionally, so every NiceGUI process registers four Socket.IO handlers (`yjs_join`, `yjs_leave`, `yjs_update`, `yjs_awareness`) even if no `ui.tiptap` is ever instantiated. Other elements (codemirror, echart, plotly) carry no module-level hook injection into `nicegui.py`. Should defer until first `Tiptap.__init__`.

**I6 — `Tiptap.Toolbar = Toolbar  # type: ignore`**
Works at runtime but IDE autocomplete won't see it. Standard NiceGUI pattern is nested classes inside the class body (e.g. `Timeline.TimelineEntry`).

**I8 — `mounted_app.elements[this.$props.id.slice(1)]` (tiptap.js:429)**
Pokes a NiceGUI runtime internal. Only other element using this pattern is codemirror. Worth @falkoschindler's call on the sanctioned API.

**I10 — `window.socket.on('yjs_update', ...)` per Tiptap instance**
N editors on a page = N handlers per `yjs_update`. A single shared dispatcher would scale better.

</details>

<details><summary>Multi-user safety / correctness</summary>

**C3 — `doc_id` rooms are unauthenticated** ⚠️ design call
Any browser can emit `socket.emit('yjs_join', {doc_id: 'persist-demo'})` and read/write any room. For demo `doc_id='shared-room'` that's intended; for app `doc_id`s used per-user-per-document, it means any other user can guess and join. Recommend at minimum a docs warning + a `doc_id_validator: Callable[[str, Client], bool]` hook for future-proofing. **This needs @falkoschindler's call before code lands.**

**C4 — pycrdt GC verification still open from Feb**
The migration from y-py to pycrdt was the right call, but the gc-verification concern you flagged isn't resolved. Lowest-cost confirmation: run `pytest tests/test_tiptap.py --count=50` (with `pytest-repeat`) and grep stderr for any pycrdt/Rust panic. If clean, the `del doc` / `del ref_doc` rituals at `test_tiptap.py:211-214, 261-264` can be dropped.

**I1 — Awareness never removed on disconnect**
`_on_yjs_leave` discards the sid from `_rooms`, but no awareness-removal is broadcast. Closed tabs leave ghost cursors visible to other collaborators until y-protocols' 30s timeout fires.

**I2 — `_docs` and `_rooms` grow unbounded**
`_get_or_create_doc` creates a `Doc` and never evicts. Long-running apps with dynamic doc_ids leak. Recommend eviction when a room becomes empty.

**I4 — `'server'` origin string used across 7 call sites**
`applyUpdate(ydoc, updateBytes, 'server')` + `applyAwarenessUpdate(..., 'server')` at 5 write sites, paired with `if (origin === 'server') return;` guards at 2 read sites — all in tiptap.js. The infinite-update-storm guard depends on one bare string literal matching across all of them. Promote to a named constant.

</details>

<details><summary>Async / performance</summary>

**I3 — Blocking pycrdt calls on the asyncio event loop**
`doc.apply_update(bytes(raw))` and `doc.get_update()` are blocking Rust calls. Small typing events are fine; a 5 MB pasted image blocks the entire NiceGUI server. Wire format also round-trips bytes as `list[int]` JSON — ~5× size inflation vs. binary frames. Switch to `asyncio.to_thread` for large payloads + Socket.IO binary frames.

**I9 — Bundle artifacts**
444 KB minified JS + 2.2 MB `.map`. codemirror commits `.map` files; worth confirming the project convention. terser's `compress: true` would shave 10-15% (currently only `mangle: true`).

**M4 — `setTimeout(100)` for "wait for yjs_init"** (tiptap.js:368)
A 100ms guess on a slow link or large state will race. Replace with an explicit `yjs_init`-received check, with the timeout as fallback.

</details>

<details><summary>Value pipeline architecture</summary>

**I5 — `value` prop vs `Y.Doc` dual source of truth** ⚠️ design call
If a user sets `editor.value = '<p>Foo</p>'` after construction, `setContentFromProps` calls Tiptap's `setContent`, which fires a Yjs update, which gets relayed to all collaborators — appending "Foo" on top of whatever they had. The first-mount initial-value path special-cases this, but runtime changes don't coordinate with Yjs. Likely answer: deprecate the runtime `value=` setter, expose `set_state` only. **Worth @falkoschindler's call.**

**C5 — `_event_args_to_value` may be dead code**
The JS emits `update:value` via Vue v-model, which goes through prop-update — not through `e.args`. The converter is registered but may never sit on the actual value-change path. The existing test `test_event_args_to_value_non_string` only exercises it via direct call. Worth tracing end-to-end and either removing or rewiring.

</details>

<details><summary>Tests</summary>

**I7 — No actual multi-user collaboration tests**
Despite ~427 lines of tests, no test opens two browser sessions and verifies that edits in one propagate to the other. `test_shared_doc_id_returns_identical_state` only checks two instances on one page. All multi-user-flavored tests reach in via `tiptap_room._get_or_create_doc` + `with doc.transaction():` — the Socket.IO relay handlers (`_on_yjs_join`, `_on_yjs_update`, `_on_yjs_awareness`) are never invoked under test. For a feature whose entire reason for existence is multi-user collaboration, this is the biggest coverage gap.

</details>

<details><summary>Cleanup / polish (minor)</summary>

- **M1** Redundant `TYPE_CHECKING` block for `pycrdt.Doc` — `from __future__ import annotations` makes the runtime `try:` import sufficient.
- **M2** `except Exception:` in `_on_yjs_update` / `_on_yjs_awareness` — narrow to the specific pycrdt exception to avoid hiding real errors.
- **M3** `Math.random()` for user color — non-trivial collision chance. Stable palette indexed by a user-id hash.
- **M5** `DEFAULT_TOOLBAR_BUTTONS` (Python) and `DEFAULT_TOOLBAR` (JS) — two sources of truth for the same defaults. Drift risk.
- **M6** Module-level `pylint: disable=protected-access` in tests — fine; worth a one-line comment marking the test-exposed items.
- **M7** "Yjs encodes an empty document as exactly 2 bytes" — implementation detail, not Yjs spec. Compare against a fresh `Doc().get_update()` instead.
- **M8** CHANGELOG/CONTRIBUTING line for the new `pycrdt` extra.

</details>

---

Let me know which items you want to keep, and which (if any) of the items in the top list you'd like me to send commits for. No rush on timeline.

evnchn:
@Denubis Just realized CRDT is involved in your project. Do you think CRDT related code from https://github.com/MQFacultyOfArts/PromptGrimoireTool can be mainlined, and do you think we should do it via this PR?

Denubis:
@evnchn The lag is *real* on my side of things. The main issue here is that I'm also using milkdown rather than tiptap, but beyond that the lag and scale issues are deeply deeply real. On the other hand, I've deployed to 2500 students? Very happy to have my claude contribute to this if you have a sense of what it should grep through?

falkoschindler:
@phifuh @evnchn — thanks both for keeping this alive, and apologies for the long silence on my end. After reading through the recent rounds, here's where I land.

## Direction: core, with reservations

I'm willing to take this into core, but I want to be honest up front: review is going to span multiple rounds, and there's no guarantee we'll converge. The architectural surface is meaningfully larger than a typical element (per-element cross-client state, awareness, room lifecycle), and I'd rather we work through that carefully than rush a merge and regret it. So please treat the green light as "let's keep going" rather than "this will land."

## The reframe I'd like to see

What feels right to me is to **stop treating this PR as "adding a Tiptap element"** and start treating it as **"adding a Yjs collaboration primitive, with Tiptap as its first consumer"**. Concretely:

- Split `tiptap_room.py` into a general `yjs_room.py` (rooms, awareness, Socket.IO handlers, doc registry) and the Tiptap-specific glue.
- That primitive then becomes available for `ui.codemirror` later via `y-codemirror`, which is the natural second consumer.
- `ui.editor` (Quasar's q-editor) stays non-collab — it has no operation-based model, so there's no clean y-binding for it. Better to be honest about that than ship a lossy HTML-diff layer.

This reframing changes how I read the rest of the PR: the Tiptap-specific code can be small and focused, and the heavier infrastructure earns its keep by serving multiple elements over time.

## The three items @evnchn tagged me on

**C3 — auth.** Yes, this needs a hook before merge. Per-instance callable, called on join, sync, default `None` = open (preserves the demo-style usage with `doc_id='shared-room'`). I'd call it `access_check` rather than `doc_id_validator` — "validator" implies format checking, and `access_check` is the right semantic level (and scope-flexible if we later gate updates too). Open question: where does it live? Per-instance is most flexible, but a per-page or per-app default would cover the "I always want this" case. I lean per-instance with the option to layer a global default later.

**I5 — `value` vs `Y.Doc`.** Route the runtime setter through a server-side Y.Doc rebuild rather than client-side `setContent`. So `editor.value = '<p>foo</p>'` means "set the canonical document content to this", server tears down and rebuilds the Y.Doc, broadcasts the new state to all connected clients. This keeps the standard `ValueElement` idiom working, has well-defined semantics, eliminates the need for a separate `set_state` method, and handles multi-tab / offline-originator cases honestly. Surgical edits from server code (e.g., "replace just this paragraph") would reach for the Y.Doc API directly — which is correct, since CRDT-aware mutations need to be ops, not HTML.

For the _read_ side: clients already emit HTML via Tiptap's `onUpdate` hook (captured by `_event_args_to_value` and stored via `ValueElement`), so a cached HTML snapshot exists today — it's just per-element-instance. Small refactor: move the cache next to the Y.Doc in `yjs_room`, keyed by `doc_id`, updated from any connected client's emission. The fully-correct version (server walks the Y.Doc XmlFragment and serializes HTML without trusting clients) needs a Python port of `y-prosemirror` and is out of scope here; the cache approach is honestly stale-prone but cheap.

**I8 — `mounted_app.elements[...]`.** Minor point: codemirror isn't the only precedent — `ui.editor`, `ui.input`, and `ui.scene` reach into the same registry. I'm fine with the pattern as-is.

## Other items

@evnchn's offer to drive the integration plumbing (C2/C5/I1/I3/I7/I10/M4) is much appreciated — please go ahead on those whenever it fits, ideally as separate small commits against @phifuh's branch so review stays digestible.

## Timeline

This is a "Next" milestone item and I'll iterate as bandwidth allows. Don't read silence as rejection; if a round goes quiet for a couple of weeks, just ping. And @phifuh — even if this takes a while, the work is genuinely good and the architectural questions it surfaces are useful regardless of the outcome.

evnchn:
Suddenly compelled to say this, typing my thoughts out loud:

@Denubis subject to clarification from @falkoschindler but I think the super-TLDR is...

> With continued efforts, NiceGUI shall become **the** framework for **custom** collaborative platforms.

Because I genuinely think that NiceGUI is well-posed, as a server-logic-dominant framework, to enable this.

In any other framework, frontend logic would drive you positively nuts if you want to get CRDT working, I'd reckon. 

I'd also say that CRDT is a genuinely hard task, even Microsoft PowerPoint doesn't get it right. If we get it right then we are on to a banger. 

The future is bright. It's just far away. Though we'd get there in no time with coding agents 💪 

---

> Let's make CRDT Nice!

Denubis:
@evnchn I am happy to contribute my insights into the problems of running CRDT at scale. Not sure this is the right thread, but... 400-500 consecutive users has a way of concentrating the mind.

phifuh:
> Suddenly compelled to say this, typing my thoughts out loud:
> 
> @Denubis subject to clarification from @falkoschindler but I think the super-TLDR is...
> 
> > With continued efforts, NiceGUI shall become **the** framework for **custom** collaborative platforms.
> 
> Because I genuinely think that NiceGUI is well-posed, as a server-logic-dominant framework, to enable this.
> 
> In any other framework, frontend logic would drive you positively nuts if you want to get CRDT working, I'd reckon.
> 
> I'd also say that CRDT is a genuinely hard task, even Microsoft PowerPoint doesn't get it right. If we get it right then we are on to a banger.
> 
> The future is bright. It's just far away. Though we'd get there in no time with coding agents 💪
> 
> > Let's make CRDT Nice!

That sounds great & honestly I would hope that more people would choose NiceGUI, my personal nightmare is tho the "connection-timeout" & no proper way tracking it

PR #5793, opened 2026-02-16
Enhance Googlebot homepage screenshot: show first demo

evnchn:
### Motivation

image

While may not relevant before the AI boom, I highly suspect Google uses VLM on the screenshot for SEO purposes. Otherwise the screenshot preview wouldn't be necessary in the first place. 

As such, having the demo a spinner when the screenshot happens is a dealbreaker. 

### Implementation

Introduce `has_demo` to `DocumentationPage`, such that each `@doc.demo` call can check in O(1) time whether there is already a demo, and fallback to `False` instead of `True` if not passed. 

Introduction of constant `TRUE_EXCEPT_FIRST_DEMO` is intentional to get the messaging across:

image

### Progress

### DIFF
--- a/website/documentation/content/doc/api.py
+++ b/website/documentation/content/doc/api.py
@@ -21,6 +21,8 @@ from .part import Demo, DocumentationPart
 registry: dict[str, DocumentationPage] = {}
 redirects: dict[str, str] = {}
 
+TRUE_EXCEPT_FIRST_DEMO = True
+
 
 @contextmanager
 def dummy_client() -> Generator[Client]:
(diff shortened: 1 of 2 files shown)


evnchn:
TL-DR: You implementation is less flexible, but is the flexibility needed? If not, simplify away 💪 

Line of interest: `kwargs.get('lazy', page.has_demo)`

Assumption when coding: the first demo _may be_ bulky enough that we deem we want lazy-loading despite the SEO downsides. 

- My implementation retains the ability to pass explicit `lazy=True`
- I don't see this being possible in your implementation? 
  - First demo must be `lazy=False` since the logic is `blah and False` which is `False` always. 


The assumption is worth questioning though: if we really have a big demo then we should not make it the first demo in the first place. 

Off the top of my head, I don't think any demos are bulky enough for now (near-offender may be `ui.scene` or some of the charts)

evnchn:
If we really have a bulky demo and we don't want to move it away from being the first one, we may investigate `lazy` to take an image.

But it'd be a task for later. Anyways I don't think we really need the flexibility.

evnchn:
I'm cooking another one based off https://github.com/zauberzeug/nicegui/pull/5793#pullrequestreview-3812353324

PR #5801, opened 2026-02-17
Radical INP improvements

evnchn:
### Motivation

As discussed privately with @rodja, driving down INP ([Interaction to Next Paint](https://web.dev/articles/inp)) can increase SEO ranking. 

However, spinning up more servers come with money concern 💸 

### Implementation

As titled, this PR is quite radical. Let's order by radicality still. 

#### LoadingBar INP trick

- Implement a Quasar LoadingBar which will be invoked when server response is needed but not yet come. As we are rendering _something_ this constitutes "next rendered frame" and INP should go down. 
- Tap into supposed private API `sub_pages_open` and `sub_pages_navigate` to trigger the LoadingBar without server involvement. 
- Hide the LoadingBar on public API `on_path_changed` which is involed with server involvement

#### Menu INP trick

- When trying `LOOPBACK=False` for `ui.menu` it doesn't work, but `args` was `[true]`
- If I did `element.props["model-value"] = args.length === 1 ? args[0] : args;` then it worked, so I kept it for this PR. 

Both of the above are still reasonable and could hit the core library one day if desired.

#### Dialog and Left-drawer INP trick

- For dialog, need to expose the `qRef` for messing with it in JS-land
- Callers to `self.open_dialog` to invoke `getElement({self.dialog.id}).$refs.qRef.show();` in JS-land directly. 
- Original responsibility of `self.input.run_method('select')` moved to `self.dialog.on('show', js_handler=f'() => {{{self.input.html_id}.select();}}')`
- When user clicks somewhere other than the dialog, directly close the dialog (use `window.searchDialogManipulating` to avoid triggering the JS handler in the JS handler's code. 
- Similar for the left drawer, but use `menuManipulating` to prevent collsion. 

Now this one is not very "Nice" of NiceGUI to be honest. I have thought of an API to make this less of a hack, but let me cook thanks. 

### Progress

### Final notes

The fact that I can come up with this PR proves that human ingenuity beats Claude Code. 

While this is SEO-facing, this is radical enough that it should be scheduled for 3.10 if not later. I guess either of those versions will be SEO-focused, which is fair since 3.8 likely will be, so we have 2 versions to collect data and see how we did. 

Note that this **does not mean** that all pages will have lower INP. TL-DR: Some will be non-optimizable. But it could be good enough. 

- Consider INP is measured by taking the 98th percentile. For our purposes (our pages have less than 50 interactions), it is the most slowest interaction. 
- Now who is the slowest interaction? If besides SPA navigation, Dialog&Drawer open/close, it must be the demo then. 

Consider the grouping, provided enough pages do not have an interactive demo to drive up the INP for that group, we should be good to go. 

- 11 pages in the `sections` group? Maybe...

image

- 111 pages in the documentation details pages group? Likely yes (some demos are quite boring after all)

image

### DIFF
--- a/main.py
+++ b/main.py
@@ -83,6 +83,12 @@ def _main_page() -> None:
             .classes('w-full').props('accordion no-connectors no-selection-unset')
     menu_button = header.add_header(menu)
 
+    menu.on('update:model-value',
+            js_handler=f'(v) => {{if (!v && !window.menuManipulating)'
+            f'{{window.menuManipulating=true;'
+            f'getElement({menu.id}).hide();'
+            f'window.menuManipulating=false;}}}}')
+
     window_state = {'is_desktop': None}
     ui.on('is_desktop', lambda v: window_state.update(is_desktop=v.args))
     ui.add_head_html('''
--- a/website/header.py
+++ b/website/header.py
@@ -41,7 +41,8 @@ def add_header(menu: ui.left_drawer) -> ui.button:
     with ui.header() \
             .classes('items-center duration-200 p-0 px-4 no-wrap') \
             .style('box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1)'):
-        menu_button = ui.button(on_click=menu.toggle, icon='menu').props('flat color=white round').classes('lg:hidden')
+        menu_button = ui.button(icon='menu').props('flat color=white round').classes('lg:hidden')\
+            .on('click', js_handler=f'() => {{getElement({menu.id}).toggle();}}')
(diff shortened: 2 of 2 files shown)


evnchn:
Should consider https://github.com/zauberzeug/nicegui/pull/5824 before considering this beast.

evnchn:
I think this is too cursed to make it into mainline NiceGUI, considering it will be LLM feeding material. This can stay where it is, while we work on #5824 to deliver INP improvements across the board while keeping a Nice and clean API. 

Note: technical speaking nothing wrong. This would land us with ~100 INP though.

PR #5806, opened 2026-02-18
Cancel connection-wait task when page coroutine completes first

### LINKED ISSUE #5803: page.py leaks Event.wait tasks on every async page request
### First Check

### Example Code

```python
# /// script
# dependencies = [
#     "nicegui @ git+https://github.com/zauberzeug/nicegui.git",
#     "httpx",
#     "pytest",
#     "pytest-asyncio",
# ]
# requires-python = ">=3.11"
# ///

from __future__ import annotations

import asyncio
import sys
from pathlib import Path

import pytest

# NiceGUI app source -- page decorators + ui.run() as required
_APP_SOURCE = """\
from nicegui import ui

@ui.page("/")
async def index():
    ui.label("Hello")

@ui.page("/sync")
def sync_page():
    ui.label("Hello sync")

ui.run()
"""


def _count_event_wait_tasks() -> int:
    return sum(
        1
        for t in asyncio.all_tasks()
        if not t.done() and getattr(t.get_coro(), "__qualname__", "") == "Event.wait"
    )


@pytest.fixture(scope="module")
def app_file(tmp_path_factory: pytest.TempPathFactory) -> Path:
    p = tmp_path_factory.mktemp("app") / "app.py"
    p.write_text(_APP_SOURCE)
    return p


@pytest.mark.asyncio
async def test_async_page_leaks_event_wait_tasks(
    app_file: Path,
) -> None:
    """Each async page open leaks one Event.wait task."""
    from nicegui.testing.user_simulation import user_simulation

    async with user_simulation(main_file=str(app_file)) as user:
        before = _count_event_wait_tasks()
        for _ in range(10):
            await user.open("/")
        after = _count_event_wait_tasks()
        leaked = after - before

        print(f"\n  Event.wait tasks before: {before}")
        print(f"  After 10 async page loads: {after}")
        print(f"  Leaked: {leaked}")

        assert leaked > 0, (
            f"Expected Event.wait task leak, but count didn't grow "
            f"(before={before}, after={after})"
        )


@pytest.mark.asyncio
async def test_sync_page_does_not_leak(
    app_file: Path,
) -> None:
    """Sync pages don't go through asyncio.wait -- no leak."""
    from nicegui.testing.user_simulation import user_simulation

    async with user_simulation(main_file=str(app_file)) as user:
        before = _count_event_wait_tasks()
        for _ in range(10):
            await user.open("/sync")
        after = _count_event_wait_tasks()
        leaked = after - before

        print(f"\n  Event.wait tasks before: {before}")
        print(f"  After 10 sync page loads: {after}")
        print(f"  Leaked: {leaked}")

        assert leaked == 0, (
            f"Sync pages should not leak Event.wait tasks "
            f"(before={before}, after={after})"
        )


if __name__ == "__main__":
    sys.exit(pytest.main([__file__, "-v", "-s"]))
```

### Description

This is a bug I found while spending most of a day getting Claude to debug 40 randomly failing end to end tests. 

Claude reports (and has demonstrated that there are different numbers being printed):

In ``nicegui/page.py``, the ``decorated()`` function creates two competing
tasks for every async page handler:

```python

    task = background_tasks.create(wait_for_result(), ...)
    task_wait_for_connection = background_tasks.create(
        client._waiting_for_connection.wait(),
    )
    await asyncio.wait(
        [task, task_wait_for_connection],
        timeout=self.response_timeout,
        return_when=asyncio.FIRST_COMPLETED,
    )
    if not task_wait_for_connection.done() and not task.done():
        task_wait_for_connection.cancel()
        task.cancel()
```

The cancellation guard (``if not ... and not ...``) only fires when
**both** tasks are still running (the timeout path). When ``task``
completes first -- the normal case for pages that don't call
``await client.connected()`` -- ``task_wait_for_connection`` is **never
cancelled**. It persists in ``background_tasks.running_tasks`` and
``asyncio.all_tasks()`` forever.

Impact
------
Each async page request leaks one ``asyncio.Event.wait()`` task. In
long-running servers this is a slow memory/task leak. In E2E test suites
with rapid page navigations it accumulates to 40-100+ tasks and degrades
server performance.

Suggested fix
-------------
After the ``asyncio.wait`` block, unconditionally cancel the connection
wait task::

    if not task_wait_for_connection.done():
        task_wait_for_connection.cancel()

(Human again)

Running the MVP scripts produce:
```bash
❮ uv run minimal/bug1_task_leak.py
============================================================================ test session starts =============================================================================
platform linux -- Python 3.14.2, pytest-9.0.2, pluggy-1.6.0 -- /home/brian/.cache/uv/environments-v2/bug1-task-leak-4b38229f5e3d5c21/bin/python3
cachedir: .pytest_cache
rootdir: /home/brian/people/Brian/nicegui-bug-repro
configfile: pyproject.toml
plugins: anyio-4.12.1, asyncio-1.3.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 2 items

minimal/bug1_task_leak.py::test_async_page_leaks_event_wait_tasks
  Event.wait tasks before: 0
  After 10 async page loads: 10
  Leaked: 10
PASSED
minimal/bug1_task_leak.py::test_sync_page_does_not_leak
  Event.wait tasks before: 0
  After 10 sync page loads: 0
  Leaked: 0
PASSED
```

This issue was making running e2e tests quite problematic. 

### NiceGUI Version

3.7.1.post30.dev0+ab47f680 (cloned from main)

### Python Version

3.14.2

### Browser

Other

### Operating System

Linux

### Additional Context

[bug1_task_leak.py](https://github.com/user-attachments/files/25382268/bug1_task_leak.py)

Denubis:
(Ala Claude, the investigation logs are in the issue. This investigation was spurred when I kicked off a bunch of playwright e2e tests and they started failing randomly, inconsistently.)

### Motivation

`page.py`'s `decorated()` creates two competing tasks via `asyncio.wait(return_when=FIRST_COMPLETED)`: the page coroutine (`task`) and a connection wait (`task_wait_for_connection`). The cancellation guard only fires on timeout — when the page coroutine completes first (the normal case for pages that don't call `await client.connected()`), `task_wait_for_connection` is never cancelled.

The leaked task wraps `client._waiting_for_connection.wait()`. Since `_waiting_for_connection` is only `.set()` inside `connected()` (client.py:211), and `handle_handshake()` calls `.clear()` not `.set()` (client.py:298), the event is never set for pages that don't call `connected()`. The task persists in `background_tasks.running_tasks` until server shutdown. Each async page load that doesn't call `connected()` adds one such task.

This is a task count leak, not a memory leak — each `Event.wait` coroutine frame is small and doesn't pin the `Client` (the Event doesn't back-reference its owner). But leaked tasks accumulate in `background_tasks.running_tasks` and `asyncio.all_tasks()` without bound over the lifetime of the server.

Closes #5803

### Implementation

`asyncio.wait(return_when=FIRST_COMPLETED)` has four possible outcomes after returning:

| Outcome | task.done() | task_wait.done() | Action |
|---------|-------------|------------------|--------|
| Page completes first | True | False | Cancel task_wait **(the fix)** |
| Client connects first | False | True | Let task finish via callback |
| Timeout | False | False | Cancel both, warn, delete |
| Both complete | True | True | No cleanup needed |

The existing `if not task_wait.done() and not task.done()` correctly handles timeout (outcome 3). That `and` is load-bearing — it distinguishes "client connected, page still loading" (outcome 2) from "timeout" (outcome 3).

The fix adds `elif not task_wait_for_connection.done(): task_wait_for_connection.cancel()` to handle outcome 1. The `elif` makes the branches mutually exclusive: timeout block handles outcome 3, the elif handles outcome 1, outcomes 2 and 4 need no action.

Cancelling `task_wait_for_connection` is safe: it wraps `asyncio.Event.wait()`, `CancelledError` is caught by `background_tasks._handle_exceptions`, and the done callback removes it from `running_tasks`. The underlying `_waiting_for_connection` event is unaffected — `handle_handshake()` can still clear/set it independently.

### Progress

- [X] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.

### DIFF
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -174,6 +174,7 @@ class page:
                                                handle_exceptions=False)
                 task_wait_for_connection = background_tasks.create(
                     client._waiting_for_connection.wait(),  # pylint: disable=protected-access
+                    name=f'wait for connection {client.page.path}',
                 )
                 done, _ = await asyncio.wait([
                     task,
--- a/nicegui/testing/user.py
+++ b/nicegui/testing/user.py
@@ -63,8 +63,6 @@ class User:
 
     async def open(self, path: str, *, clear_forward_history: bool = True) -> Client:
         """Open the given path."""
-        if self.client is not None:
-            self.client.delete()
         response = await self.http_client.get(path, follow_redirects=True)
         assert response.status_code == 200, f'Expected status code 200, got {response.status_code}'
         if response.headers.get('X-Nicegui-Content') != 'page':
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -349,8 +349,7 @@ async def test_async_page_does_not_leak_event_wait_tasks(user: User):
     for _ in range(5):
         await user.open('/')
     await asyncio.sleep(0.1)
-    assert sum(
-        1
+    assert not any(
+        t.get_name().startswith('wait for connection') and not t.done()
         for t in asyncio.all_tasks()
-        if not t.done() and getattr(t.get_coro(), '__qualname__', '') == 'Event.wait'
-    ) == 1, 'only the Event.wait task for the last opened page should remain, all previous ones should be cleaned up'
+    ), 'connection-wait tasks should be cancelled after page coroutine completes'


evnchn:
First time I see Copilot have no comments. Neither do I. @falkoschindler it's your turn now.

Denubis:
Also dug into this one. Here are the revising down of the claims.

Claude says:

Following up with what we've confirmed since the original report.

The E2E failures had three primary causes on our end: (1) `asyncio.set_debug(True)` blocking the event loop via `linecache.checkcache()` — this was the 5-7s block; (2) cleanup in `on_disconnect` instead of `on_delete`, causing state accumulation on reconnects; (3) missing inter-test cleanup boundaries.

The page.py task leak (#5806) and outbox linger (#5805) were discovered during the investigation. They're real bugs — the page.py leak accumulates one undone `Event.wait` task per async page load, and `Outbox.stop()` doesn't wake the loop — but they weren't the direct cause of our test failures. They amplified the debug mode problem (more tasks = more `checkcache()` overhead per `create_task()`) and forced workaround sleeps in our test harness.

We initially overclaimed the performance impact of #5805. As Falko identified, the outbox linger is concurrent and doesn't block the caller. We've retracted that claim on the PR.

The `cr_frame.f_locals` technique for tracing leaked tasks back to their origin might be useful for the diagnostics tooling discussion (#5660) — it lets you attribute orphan `Event.wait` tasks to specific Client instances without guessing.

There are exactly four occurrences of `_waiting_for_connection` in the codebase:

```
client.py:73   self._waiting_for_connection = asyncio.Event()   # constructor, unset
client.py:211  self._waiting_for_connection.set()                # connected(), the only .set()
client.py:298  self._waiting_for_connection.clear()              # handle_handshake(), clears not sets
page.py:176    client._waiting_for_connection.wait()             # the leaked task
```

No `.set()` outside `connected()`. No cancellation outside the timeout branch. `delete()` doesn't touch it.

evnchn:
@Denubis I have bad news:

- The test you added passes both with and without your fix, so it's not testing anything meaningful. \
- Adding a delay in hopes of getting the Outbox to stop in time (ref #5805) only made the test fail both with and without your fix

The good news is:

- Now with the tightened test, without your fix we leak 10 tasks, with it we leak 5 tasks. So the value of the PR is still there, but just not quite complete yet. 

Maybe we need to do more work...

Denubis:
Yep. Absolutely need to dig more. I wonder why my setup is so different. Back to the drawing board.

evnchn:
@Denubis since 3.8 is round the corner I did it in https://github.com/zauberzeug/nicegui/pull/5806/commits/d14f1a45316a8565442bfa69f7edc0dde60439e0

I think the moral of the story is: Ensure the test fails on main and passes on your PR to solidify the value of your PR. Even if the test is not pytest.

Denubis:
Thank you. I appreciate it.

Denubis:
(It's friday night, and I've had 2 gins, so -- I can attest that I, as human, have hit f5 a bunch on both main and my branch, and have observed the difference, but I don't have the foggiest clue *why*)

Documenting my verification 1) because I'm somewhat drinky, and 2) because human in the loop

first, on main
> uv run ../mre1_event_wait_leak.py

After 4 f5's
> Event.wait tasks: 4

Then on my branch, robust on 10 f5's
> Event.wait tasks: 1

```python
"""MRE 1: Event.wait task leak. Reload (F5) and watch count grow."""
import asyncio
from nicegui import ui

@ui.page('/')
async def index():
    n = sum(1 for t in asyncio.all_tasks()
            if not t.done() and getattr(t.get_coro(), '__qualname__', '') == 'Event.wait')
    ui.label(f'Event.wait tasks: {n}').classes('text-h4')

ui.run(port=8081)
```
[mre1_event_wait_leak.py](https://github.com/user-attachments/files/25439566/mre1_event_wait_leak.py)


Claude asserts:

```
  Question 1: Does the leak exist?

  ┌───────────────────────────────┬──────────────────────┬──────────────────────┐
  │           Evidence            │         main         │      PR branch       │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ MRE via curl (10 requests)    │ 38 → 47, +1/request  │ 1 → 1, flat          │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ MRE via browser F5            │ You verified: grows  │ You verified: stable │
  ├───────────────────────────────┼──────────────────────┼──────────────────────┤
  │ pytest (same test both sides) │ FAILS: leaked 5 of 5 │ PASSES: leaked 0     │
  └───────────────────────────────┴──────────────────────┴──────────────────────┘

  Question 2: Why Falko's MRE doesn't show it

  His MRE counts len(asyncio.all_tasks()) — all tasks. The +1 leaked Event.wait task per reload is masked by ~17 other framework tasks churning (timers, outbox, cleanup). Filtering
  to Event.wait by __qualname__ isolates the signal.
```

Then the other ask you made.

ah. After poking it, it's a bug in the test fixture itself. (WTF -- maybe worth another PR, I leave that in your capable hands and will not object either way)

> The client.delete() change is in User.open() — the test fixture, not the server. A browser/curl MRE can't demonstrate it because:
>   - Browser: WebSocket disconnect → reconnect timeout → client.delete() happens automatically
>  - curl: HTTP-only, no WebSocket, no cleanup — same as User.open without the fix
>  - The fix: only fires inside User.open(), which only runs in pytest


I, as human, attest that I've seen these two differentials. I also cannot for the life of me explain them. 

And, for the delete (a few rounds of back and forth)

```
git reset --hard HEAD
    cp ../mre2_user_open_cleanup.py tests/
    uv run pytest tests/mre2_user_open_cleanup.py -x -v -s
```
on main produces:
> E       AssertionError: User.open accumulated 5 stale clients (expected 1)

Claude asserts:
> MRE 2 isolates the User.open change only. It counts Client.instances, which is unaffected by the page.py fix. So:

then on worktree
```fish
  git reset --hard HEAD
  cp ../../mre2_user_open_cleanup.py tests/
  .venv/bin/pytest tests/mre2_user_open_cleanup.py -x -v -s
```

> PASSED

Then to really prove those two lines

```
  git reset --hard HEAD
  sed -i '66,67d' nicegui/testing/user.py
  cp ../../mre2_user_open_cleanup.py tests/
  .venv/bin/pytest tests/mre2_user_open_cleanup.py -x -v -s
```
Fails with the error
[mre2_user_open_cleanup.py](https://github.com/user-attachments/files/25440038/mre2_user_open_cleanup.py)

```py
"""MRE 3: User.open stale client accumulation.

  cp to tests/ and run from inside a NiceGUI checkout:

  FAIL (main):  Client.instances grows by 5
  PASS (fix):   Client.instances grows by 1 (only the current page)
"""
import asyncio

from nicegui import ui


async def test_user_open_cleans_up_old_clients(user: User):
    @ui.page('/')
    async def page():
        ui.label('Hello')

    before = len(Client.instances)
    for _ in range(5):
        await user.open('/')
    after = len(Client.instances)
    new = after - before

    print(f'\nClient.instances: before={before}, after={after}, new={new}')
    assert new <= 1, f'User.open accumulated {new} stale clients (expected 1)'
```

I cannot speak to the *impact* of this, but I have observed the state change.

evnchn:
Issue 1:

The code _does_ do something because injecting a print shows it being done. 

image

As to whether it is useful? Not sure. 

To my dry-tracing: `task_wait_for_connection` waits for `client._waiting_for_connection` which is only `set()` by `Client.connected()`. If not for this new statement, the task was never `set()` and `task_wait_for_connection` could indeed wait forever. 

However, I have failed to isolate just those tasks and get a growing number figure. 

---

Issue 2:

It is very obvious that we should call `client.delete()` before getting rid of the `client` instance. I think it's a simple change that should be adopted.

falkoschindler:
Ok, strange. Now I _do_ get a growing number of tasks with my MRE from https://github.com/zauberzeug/nicegui/pull/5806#pullrequestreview-3827280737. 👍🏻

evnchn:
Are we expected to re-implement the `User.open change` in another PR in another day?

falkoschindler:
@evnchn I don't think so. Here's why:

The `client.delete()` added to `User.open()` doesn't match production behavior — in production, stale clients are pruned by `Client.prune_instances()` every 60s or after a reconnect interval, not eagerly deleted on navigation. The eager delete was also masking a problem in the test: the assertion filtered by `Event.wait` qualname, which is shared by *all* `asyncio.Event.wait()` calls. Without the delete, the test caught unrelated `Outbox._enqueue_event.wait()` tasks and failed. So the test only passed because `client.delete()` stopped each old client's outbox loop as a side effect.

Three adjustments in this commit:

1. **Reverted `client.delete()` in `User.open()`** to match real-world behavior.
2. **Added a `name` to the `task_wait_for_connection` task** (`wait for connection /path`) so it's identifiable for debugging and testing.
3. **Rewrote the test assertion** to filter by task name instead of generic qualname, asserting exactly 0 connection-wait tasks remain.

evnchn:
That makes sense. We shouldn't purely chase for lower amount of tasks. Let's merge!

PR #5810, opened 2026-02-19
Add `ui.status_code()` to set HTTP status codes from page builders

evnchn:
### Motivation

Page builder functions currently have no way to control the HTTP status code of the response. This is needed for proper SEO (returning 404 for not-found pages) and for any scenario where a page needs to signal a non-200 status.

Split out from #5767 per review feedback — this is a **public API addition** that deserves focused review.

### Implementation

- **`nicegui/functions/status_code.py`** (new): `ui.status_code(code)` sets `context.client.status_code`
- **`nicegui/client.py`**: Adds `status_code` field to `Client`, uses it in `build_response()` when set
- **`nicegui/ui.py`**: Exports `status_code` in `__all__` and imports
- **`main.py`**: Uses `ui.status_code(404)` for unknown documentation pages
- **`tests/test_page.py`**: Tests for custom and default status codes

### Open questions for review

- Should `code` be validated (e.g. must be in 100-599 range)?
- Should it return the previous code?
- Does it need documentation on the website?

### Progress

### DIFF
--- a/website/documentation/content/status_code_documentation.py
+++ b/website/documentation/content/status_code_documentation.py
@@ -14,3 +14,24 @@ def main_demo() -> None:
     def page():
         ui.link('Visit the teapot page', '/teapot')
     page()  # HIDE
+
+
+@doc.demo('Conditional 404', '''
+    You can use `ui.status_code` to return a 404 status code when an item is not found.
+    This is useful for SEO and for signaling to clients that a page does not exist.
+''')
+def item_demo() -> None:
+    @ui.page('/item/{item_id}')
+    def item_page(item_id: str):
+        if 'a' <= item_id <= 'c':
+            ui.label(f'This is item {item_id}')
+        else:
+            ui.status_code(404)
+            ui.label('Page not found')
+
+    # @ui.page('/')
+    def page():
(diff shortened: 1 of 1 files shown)


falkoschindler:
Quick clarification:

> This is needed for proper SEO (returning 404 for not-found pages)

Undefined routes already return 404. Or am I missing something?

```py
@ui.page('/')
def page():
    ui.label('Hello, world!')
```
```
$ curl -I http://127.0.0.1:1234/foo
HTTP/1.1 404 Not Found
...
```

falkoschindler:
Ah! You're focusing on scenario 2:

---

The distinction is between two different scenarios:

1. **No matching route** (e.g., visiting `/totally-random-url`): NiceGUI/FastAPI already returns HTTP 404 — you're right about this.

2. **Route matches but content doesn't exist** (e.g., `/documentation/{name}` where `name` is nonsense): The `{name}` wildcard catches the request, the page builder runs, renders the "not found" label, and returns **HTTP 200** — because from the framework's perspective, the page builder succeeded.

evnchn:
If we want proper SEO, then we'd **have** to merge this early, since currently the status code is **hard-coded** to 200 with no way to change it.

PR #5812, opened 2026-02-19
fix(reaktiv example): assign Effects to variables to prevent garbage collection

buiapp:
### Motivation

Effects in reaktiv must be assigned to variables, otherwise they get garbage collected and stop working. The previous example only worked due to a bug in reaktiv that has since been fixed. Updated the example to properly store Effect instances.

This API behavior is documented in the reaktiv docs.

References: #5783, #4758

### Implementation

We need just to assign Effects to variables.

### Progress

### DIFF
--- a/examples/reaktiv/main.py
+++ b/examples/reaktiv/main.py
@@ -23,9 +23,10 @@ with ui.row():
         tax_label = ui.label()
         total_label = ui.label().classes('text-bold')
 
-        Effect(lambda: subtotal_label.set_text(f'Subtotal: ${subtotal():.2f}'))
-        Effect(lambda: tax_label.set_text(f'Tax: ${tax():.2f}'))
-        Effect(lambda: total_label.set_text(f'Total: ${total():.2f}'))
+        # NOTE: Effects must be assigned to variables to prevent garbage collection
+        subtotal_effect = Effect(lambda: subtotal_label.set_text(f'Subtotal: ${subtotal():.2f}'))
+        tax_effect = Effect(lambda: tax_label.set_text(f'Tax: ${tax():.2f}'))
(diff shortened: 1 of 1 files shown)


evnchn:
Ref: https://github.com/buiapp/reaktiv/commit/7f7a383bf0c8ca5337c251fc18afd85acad7b8c3

Would prefer to wait for a reaktiv release with the fix, and update `requirements.txt`, before merging.

falkoschindler:
@evnchn I don't understand. The fix doesn't change the fact that we should keep references. The example works before and after.
I'd rather release it with 3.8 to avoid publishing a possibly broken example.

evnchn:
@falkoschindler Thats fair. Perhaps the update of `requirements.txt` should be a non-blocker to this PR instead :bow:

evnchn (approved):
As mentioned, the code is correct and should be merged ASAP to not ship a bad example.

buiapp:
Thank you for the approved pull request! I am also very sorry for this inconvenience. 🙇‍♂️ I just pushed the new reaktiv version to PyPI.

In our company we are using reaktiv in combination with NiceGUI very extensively and always keep a reference to Effects to prevent it from being GC'd. But it seems that it didn't worked as intended.. The weakref behavior is to prevent memory leaks (https://github.com/buiapp/reaktiv/issues/28). I thought that my unit tests already catched the garbage collection case... but it seems I was wrong about it.

Thank you for everything! You guys rock!

Greetings from Hamburg!

evnchn:
Oops, since you PR against your main, I can't easily slip in bumping `requirements.txt` into this PR on GitHub mobile without doing some nasty tricks on desktop.

PR #5820, opened 2026-02-19
Update reaktiv example to use non-leaking version

evnchn:
### Motivation

The example code pattern that previously leaked subscriptions was already fixed in the example. This PR makes the complementary change: it bumps the reaktiv dependency used by the example to the release that contains the non-leaking fix so users get the corrected behavior without further code changes. No related issue exists.

### Implementation

- Bump the reaktiv dependency/version used by the example to the release that includes the non-leaking fix.
- The example code itself is unchanged; this is a minimal dependency update (one-line change) that ensures the example does not increase memory/resource usage over time.
- Low-risk change limited to the example dependency specification.

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will bump the reaktiv dependency used by the example to a version that fixes subscription/resource leaks"
- [x] If this PR addresses a security issue, it has been coordinated via the [security advisory] (not applicable).

### Final notes

I'm on the bus, this is written by GitHub Copilot. Please cut me some slack 🙇

### DIFF
--- a/examples/reaktiv/requirements.txt
+++ b/examples/reaktiv/requirements.txt
@@ -1,2 +1,2 @@
 nicegui>=3.0
-reaktiv>=0.21.1
+reaktiv>=0.21.2


falkoschindler:
Thanks, @evnchn! However, I don't think this change is necessary.

PR #5812 fixed the example code so that Effects are properly held in variables instead of being garbage collected while still in use. With that fix in place, the example works correctly on `reaktiv>=0.21.1`.

The leak fix in reaktiv 0.21.2 addresses a bug in the reaktiv library itself - but it's not the example's job to pin dependencies to work around bugs in third-party libraries. Users will naturally get the latest version via `pip install` anyway since we use `>=`.

Therefore I'm closing this PR.

PR #5821, opened 2026-02-19
Fix blank page after client-side sub_pages navigation

### LINKED ISSUE #5816: Client-side sub_pages navigation renders blank page when clicking links
### First Check

### Example Code

```python
from nicegui import ui

@ui.page('/')
@ui.page('/page_a')
@ui.page('/page_b')
def main():
    with ui.header():
        ui.link('Home', '/')
        ui.link('Page A', '/page_a')
        ui.link('Page B', '/page_b')

    ui.sub_pages({
        '/': lambda: ui.label('Welcome to the homepage'),
        '/page_a': lambda: ui.label('This is page A'),
        '/page_b': lambda: ui.label('This is page B'),
    })

ui.run()
```

### Description

When navigating between pages using client-side routing via `ui.sub_pages`, clicking a `ui.link` in the header causes the URL to update correctly but the page content is blank — only the header remains visible. A manual browser refresh (F5) then loads the page correctly.

**Steps to reproduce:**

1. Run the example code above (or `python main.py` for the NiceGUI website)
2. Open `http://localhost:8080/` — the homepage loads correctly
3. Click "Page A" (or "Documentation" on the website) in the header
4. URL changes to `/page_a` (or `/documentation`) but the content area is empty

**Expected behavior:** The target page content should render after clicking the link.

**Actual behavior:** The URL updates but the content area below the header is blank. Only the header navigation items are visible.

This also affects the NiceGUI website itself (`main.py`): clicking any header link (Documentation, Examples, etc.) from any page results in a blank content area. Originally reported in https://github.com/zauberzeug/nicegui/discussions/5794#discussioncomment-12623277.

### NiceGUI Version

3.7.1.post28.dev0+5685aec8 (main branch, commit 06162052)

### Python Version

3.10.19

### Browser

Chrome

### Operating System

Linux

### Additional Context

The issue is 100% reproducible, not intermittent. Direct navigation (typing the URL or refreshing) works — only client-side SPA navigation via clicking links is broken.

evnchn:
### Motivation

Fixes #5816.

The vnode cache introduced in #5761 prevents Vue 3.5's dependency tracking from capturing all elements during partial re-renders. After a small update triggers a cached re-render, the render effect's dep list shrinks from hundreds of elements to ~40 (only those whose vnodes were invalidated). Subsequent changes to elements outside that set — such as `ui.sub_pages` content during client-side navigation — go undetected by Vue, leaving the page blank.

Originally reported in https://github.com/zauberzeug/nicegui/discussions/5794#discussioncomment-12623277.

### Implementation

Add a reactive `renderToggle` boolean to the Vue app's data. The render function subscribes to it (`void this.renderToggle`), and the update handler flips it after every update. This guarantees Vue always re-enters the render function when elements change, while the vnode cache still prevents unnecessary DOM work for unchanged elements.

### Progress

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -420,11 +420,9 @@ function createApp(elements, options) {
     data() {
       return {
         elements,
-        renderToggle: false,
       };
     },
     render() {
-      void this.renderToggle; // NOTE: subscribe to renderToggle so the update handler can trigger re-renders
       return renderRecursively(this.elements, 0);
     },
     mounted() {
@@ -536,7 +534,7 @@ function createApp(elements, options) {
           }
 
           invalidateVnodeCache(Object.keys(msg).map(Number));
-          this.renderToggle = !this.renderToggle; // NOTE: trigger re-render since vnode cache prevents Vue's dependency tracking from capturing all elements
+          this.$forceUpdate(); // NOTE: vnode cache prevents Vue's dependency tracking from capturing all elements
 
           await this.$nextTick();
           for (const [id, element] of Object.entries(msg)) {


falkoschindler:
MRE:

```py
from nicegui import ui

def root():
    with ui.header():
        ui.link('Index', '/')
        ui.link('Other', '/other')

    ui.sub_pages({
        '/': lambda: ui.label('Index page'),
        '/other': lambda: ui.label('Other page'),
    })

ui.run(root)
```

falkoschindler:
I "simplified" the fix and added a pytest. But now Opus seems to doubt the whole vNodeCache being effective at all. If we force re-rendering, we might as well drop the whole cache. This is getting way over my Vue knowledge... 🫤

evnchn:
**TL;DR:** No, `$forceUpdate()` does not invalidate the vnode cache's performance gains. They operate at different levels — `$forceUpdate()` ensures Vue calls `render()`, while the vnode cache makes each `render()` call cheap by skipping unchanged subtrees.

<details>
<summary>Full analysis</summary>

### What `$forceUpdate()` does
It tells Vue to call the component's `render()` function again. That's all — it schedules a re-render pass.

### What the vnode cache does
Inside `renderRecursively()`, it short-circuits vnode creation for unchanged elements. When the cache hits, no props processing, no event binding, no `Vue.h()` call, no recursion into children — just return the cached vnode immediately.

### The update flow with `$forceUpdate()`

1. Update handler receives changed elements, writes them into `this.elements`
2. `invalidateVnodeCache()` removes cache entries for **only** the changed IDs + their ancestors up the tree
3. `$forceUpdate()` ensures Vue calls `render()`
4. `render()` → `renderRecursively(elements, 0)` traverses the tree:
   - Changed elements + ancestors: **cache miss** → rebuild vnode
   - All other elements: **cache hit** → return instantly, skip entire subtree

So if you have 500 elements and only 3 change, the render function still runs, but it only does real work for those 3 elements plus their ancestor chain. The other ~490+ elements return cached vnodes in O(1).

### Key insight

The render function was *already supposed to run* on every update — that's how Vue works. The bug (#5816) was that the vnode cache broke Vue 3.5's dependency tracking: since cached elements' properties were never accessed during render, Vue stopped tracking them as deps and would sometimes skip the render entirely. `$forceUpdate()` is a correctness fix that restores the pre-cache behavior (render always runs), while the cache still makes each render *cheaper*.

`$forceUpdate()` adds essentially zero overhead beyond what a normal reactive update would trigger. It's functionally equivalent to the earlier `renderToggle` approach but cleaner.

</details>

---
*This comment was written by Claude Code on behalf of @evnchn. Initial prompt: "Does the forcedUpdate invalidate all performance gains of the VNode cache?"*

falkoschindler:
Ok, my AI agrees. But it found something else:

---

The `$forceUpdate()` fix in the `update` handler is correct and the vnode cache remains valuable (it avoids vnode construction and lets Vue skip patching same-reference subtrees).

However, there is a gap in the loopback handler at line 295–298:

```js
if (element.props["loopback"] === False && event.type == "update:modelValue") {
    element.props["model-value"] = args;
    invalidateVnodeCache([id]);
}
```

This invalidates the cache but never triggers a re-render. Before the vnode cache existed, the reactive mutation to `element.props["model-value"]` always triggered Vue's reactivity because every element was read during render. With the cache, Vue's dependency set narrows after partial renders — elements served from cache aren't tracked as dependencies. So the mutation can go undetected.

Crucially, `LOOPBACK = False` elements (`ui.input`, `ui.number`, `ui.editor`, `ui.color_input`, `ui.date_input`, `ui.time_input`) deliberately skip the server round-trip — `_send_update_on_value_change` is set to `False`, so `self.update()` is never called and no update message is sent back. There is no `$forceUpdate()` safety net for these elements.

In practice the user still sees their typed text because Quasar components manage their own display state internally. But the mechanism is architecturally incomplete: it mutates data and invalidates the cache, then hopes something else will trigger a render.

The fix would be to also call `mounted_app.$forceUpdate()` (which is available in the closure scope) after invalidating the cache:

```js
if (element.props["loopback"] === False && event.type == "update:modelValue") {
    element.props["model-value"] = args;
    invalidateVnodeCache([id]);
    mounted_app.$forceUpdate();
}
```

This matches the pattern used in the `update` handler and restores the same render frequency these elements had before the vnode cache was introduced.

evnchn:
Hmm. Is there an MRE for that? 

I asked "Do we need to force a re-render for all call sites of `invalidateVnodeCache`" before and the answer _was_ no.

falkoschindler:
**TL;DR** The loopback handler (line 297) has a theoretical dep-tracking gap but no observable bug today because Quasar compensates. Not worth blocking the PR over.

<details>
<summary>
Full analysis
</summary>
Good question. Let me think about whether the dep-narrowing in the loopback path actually produces a visible bug.

The scenario requires:
1. A `LOOPBACK=False` element (e.g., `ui.input`)
2. A prior partial update that narrows Vue's dep set (so the input drops out of tracking)
3. Typing in the input — the local mutation goes undetected
4. Something that makes the stale Vue state **observable**

The problem: step 4 is hard to achieve because of compensating mechanisms:

- **Quasar manages its own display** — the user sees their typed text regardless of whether Vue re-renders
- **The NiceGUI data model is correct** — `element.props["model-value"]` was set locally
- **The cache is invalidated** — so any future render picks up the right value
- **The event still reaches the server** — Python-side `handle_change` runs, `self.value` is correct
- **Any server update to any element** calls `$forceUpdate()`, which catches everything up

For a visible glitch you'd need: dep-narrowing has occurred, user types in the input, no server update follows, AND something that reads the Vue-rendered state rather than Quasar's internal state or NiceGUI's data model.

I don't think there's a practical MRE. The loopback issue is a correctness gap in the reactivity wiring — Vue's dep tracking is wrong for these elements — but the visible behavior is correct because Quasar's internal state and the NiceGUI data model both compensate. It's the kind of thing that might surface under a future Vue or Quasar change but isn't observable today.

So the `mounted_app.$forceUpdate()` addition in the loopback handler would be a **defensive** fix, not a fix for a user-visible bug. Up to you whether it's worth flagging on the PR.
</details>

evnchn:
Let's leave it for now. I think this is mergeable.

PR #5824, opened 2026-02-19
Cut header and tree visual effects for mobile

evnchn:
### Motivation

While good INP is strongly desired for SEO purposes, the approach at #5801 is hacky, to say the least. 

Are there any lower-hanging fruits before we go the nuclear option?

### Implementation

I spotted 4 places for optimization:

- Mobile devices lack room. They would benefit from the shrunk header almost always. Removing dynamic header grow-shrink logic cuts layout recomputations for frames in a 200ms window. 
- CSS blur effects have a high computational cost, and it honestly doesn't matter on mobile, so I substituted the colors with what they would be assuming empty background to minimize felt change. 
- While likely broken just because https://github.com/zauberzeug/nicegui/pull/5732 is not merged yet, the docs tree is supposed to collapse-on-click for mobile. As such, remove the animation which is not readily seen by the user but would have been rendered anyways. 
- The search dialog now uses a conditional `fade` transition on mobile (via `Quasar.Screen.lt.md`) instead of the default scale animation, reducing compositing work.

While the network bottleneck is still there, we do everything in our power to ensure the CPU load is relieved as much as possible. 

As we need only to shave 11ms for `https://nicegui.io/documentation/section_*` we might just have a win for that. For the others 39ms is needed, and I am not super hopeful on that. 

image

### Progress

### DIFF
--- a/main.py
+++ b/main.py
@@ -66,7 +66,7 @@ def _main_page() -> None:
             .style('height: calc(100% + 20px) !important') as menu:
         tree = ui.tree(documentation.tree.nodes, label_key='title',
                        on_select=lambda e: ui.navigate.to(f'/documentation/{e.value}')) \
-            .classes('w-full').props('accordion no-connectors no-selection-unset')
+            .classes('w-full nav-tree').props('accordion no-connectors no-selection-unset')
     menu_button = header.add_header(menu)
 
     window_state = {'is_desktop': None}
--- a/main.py
+++ b/main.py
@@ -66,7 +66,7 @@ def _main_page() -> None:
             .style('height: calc(100% + 20px) !important') as menu:
         tree = ui.tree(documentation.tree.nodes, label_key='title',
                        on_select=lambda e: ui.navigate.to(f'/documentation/{e.value}')) \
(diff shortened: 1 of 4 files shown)


evnchn:
About 50% faster, if we consider the Performance Monitor tab

evnchn:
Also according to https://pagespeed.web.dev/analysis/https-nicegui-io/li5rniz6jo?hl=en_GB&form_factor=mobile, the header padding is a ["non-composited animation" and we best avoid it](https://developer.chrome.com/docs/lighthouse/performance/non-composited-animations/).

evnchn:
image

I think the visual balance for the "iPad form-factor" is wrong, and we should limit the non-moving header to mobile form-factor only. After all, the Googlebot is either a smartphone or a desktop, no in-between.

falkoschindler:
@evnchn This PR looks pretty outdated. Do you think it's worth rebasing or should we just close it?

evnchn:
@falkoschindler I appear to have missed this PR, while closing PRs noting how the redesign already fixed INP issues (double checked just now, we are green since 25 Mar, so we're stable for a month)

I'm closing this.

Issue #5823, opened 2026-02-19
`ui.log` auto-scroll broken by vnode caching


falkoschindler:
### Description

PR #5761 ("Cache vnodes to avoid unnecessary re-renders") introduced a vnode cache in `nicegui.js` that skips re-rendering unchanged components. This broke the `updated()` lifecycle hook in `log.js`, which relies on being called whenever new child labels are added via `log.push()` in order to auto-scroll to the bottom.

Two scenarios are affected:

1. **`ui.log` in a `ui.tab_panel`**: After switching away from the tab containing the log and switching back, newly pushed lines are no longer auto-scrolled into view. This was already reported in the review of #5220 and was previously working on `be4d520` but broke on `6ba80ed` (the merge commit of #5761).

2. **`ui.log` scroll-to-bottom in general**: The `updated()` hook in `log.js` may not fire when new children are added if the vnode cache considers the log element unchanged, preventing the `setScrollPosition` call.

### Root Cause

`log.push()` creates new child `Label` elements inside the log's default slot. The vnode cache in `renderRecursively` may return a cached vnode for the log element if it doesn't detect the slot content change, so Vue skips the component update entirely — meaning `updated()` never fires and the scroll position is never adjusted.

### Steps to Reproduce

```python
import time
from nicegui import ui

with ui.tabs() as tabs:
    one = ui.tab('One')
    two = ui.tab('Two')

with ui.tab_panels(tabs, value=one).classes('w-full'):
    with ui.tab_panel(one):
        log = ui.log().classes('h-20')
    with ui.tab_panel(two):
        ...

ui.timer(0.5, lambda: log.push(time.strftime('%H:%M:%S')))

ui.run()
```

1. Run the app
2. Switch to tab "Two"
3. Switch back to tab "One"
4. Observe that the log is no longer scrolling to the bottom

### Expected Behavior

`ui.log` should auto-scroll to the bottom whenever new lines are pushed, regardless of vnode caching.

### Context

- Reported in #5788 ([comment](https://github.com/zauberzeug/nicegui/issues/5788#issuecomment-3916203728), [comment](https://github.com/zauberzeug/nicegui/issues/5788#issuecomment-3916225205))
- Broken by #5761 (merge commit `6ba80ed`)
- Last known working commit: `be4d520`
- Blocks 3.8 release

falkoschindler:
[...]

PR #5826, opened 2026-02-20
Revert vnode cache and guard innerHTML writes

falkoschindler:
### Motivation

Reverts the vnode cache from #5761, which broke `ui.log` auto-scroll (#5823) and potentially other elements relying on Vue's `updated()` lifecycle hook.

Instead applies the component-level approach from #5757: guards `renderContent()` in `html.js`, `markdown.js`, and `interactive_image.js` to skip redundant `innerHTML` writes, preserving client-side DOM modifications (#5749).

The vnode cache is the right long-term fix, but needs more work to handle elements that legitimately depend on `updated()` firing when slot content changes (like `ui.log`). For 3.8, the component-level guards are the safer path.

### Implementation

**`nicegui.js`**: Reverted `vNodeCache`, `parentOf`, `invalidateVnodeCache()`, cache lookup/storage in `renderRecursively`, and the `$forceUpdate()` from #5821.

**`html.js`, `markdown.js`, `interactive_image.js`**: Added a `previous*` guard in `renderContent()` that compares the current content against the previously rendered value and returns early if unchanged. These are the only three elements that gained destructive `innerHTML` side effects in `updated()` via f1f7533 (the sanitize feature). All other elements with `updated()` hooks are either idempotent or already have their own guards.

### Progress

- [x] New pytests are not necessary.

### DIFF
--- a/nicegui/elements/html.js
+++ b/nicegui/elements/html.js
@@ -12,12 +12,12 @@ export default {
   methods: {
     renderContent() {
       if (this.innerHTML === this.previousInnerHTML) return;
-      this.previousInnerHTML = this.innerHTML;
       if (this.sanitize) {
         this.$el.setHTML(this.innerHTML);
       } else {
         this.$el.innerHTML = this.innerHTML;
       }
+      this.previousInnerHTML = this.innerHTML;
     },
   },
   props: {
--- a/nicegui/elements/interactive_image.js
+++ b/nicegui/elements/interactive_image.js
@@ -77,7 +77,6 @@ export default {
     renderContent() {
       const content = this.content || "";
       if (content === this.previousContent) return;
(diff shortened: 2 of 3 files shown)


falkoschindler:
Great: nicegui.js is reverted to its pre-vnode-cache state:
https://github.com/zauberzeug/nicegui/compare/be4d5201d87f0ec63b8e941583486dc2213d107d..aeef08c6070855beddcbaa067bb88cf7a817bd1e#diff-2e7a902681849d4a8afbb7f3bb101fb145be4f0c2e07a62c15ff9a928b8bc4a7

falkoschindler:
@evnchn I did notice the new line. But it makes sense to separate top-level functions this way. So it's a nice leftover.

PR #5830, opened 2026-02-22
Plausible SPA compatibility

evnchn:
### Motivation

image

NiceGUI's SPA is totally not known to Plausible, as such the analytics is quite messed up. 

### Implementation

This pull request updates the Plausible analytics script to a newer version that automatically handles Single-Page Application (SPA) tracking. 

### Progress

### DIFF
--- a/website/header.py
+++ b/website/header.py
@@ -15,11 +15,14 @@ def add_head_html() -> None:
     """Add the code from header.html and reference style.css."""
     ui.add_head_html(HEADER_HTML + f'<style>{STYLE_CSS}</style>')
     if os.environ.get('ENABLE_ANALYTICS', 'false').lower() == 'true':
-        ui.add_head_html(
-            '<script defer data-domain="nicegui.io" src="https://plausible.io/js/script.outbound-links.js">'
-            '</script>'
-        )
-        ui.context.client.sub_pages_router.on_path_changed(lambda _: ui.run_javascript('plausible("pageview")'))
+        ui.add_head_html('''
+            <!-- Privacy-friendly analytics by Plausible -->
+            <script async src="https://plausible.io/js/pa-d9tl9pdueh-ArCJfoLv3H.js"></script>
+            <script>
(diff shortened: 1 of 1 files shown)


evnchn:
https://github.com/zauberzeug/nicegui/pull/5830/commits/9c596dc36e3ce497b1ee7e46457c371730932b66 is tested by using `{captureOnLocalhost: true}` in `plausible.init`

I can verify that the `pageview` event is emitted without the SPA `on_path_changed` change on localhost. 

`async`, the HTML comment, and the missing semicolon in `plausible.init()` is all came from the below:

image

rodja (changes requested):
Thanks for digging into this @evnchn. Just two things I noticed:
- non-sub-page hashes are not tracked anymore (for example there are no events fired to plausible if you go to documentation/markdown and then use the side bar to navigate to different demos)
- description of the PR still says "Added a handler... " which is not in the code

evnchn:
@rodja as mentioned privately, I do not think we should track the hash when our page isn't doing hash based routing.

As a reminder, hash-based routing is intended for websites which use `/#pages/mypage` to serve distinct and click-navigated pages. NiceGUI is not that. Sections appear on the same page and is use scroll to navigate among them. 

Since we use the hash for sections inside a page, insist on tracking the hash with Plausible assuming hash-based routing has several disadvantages:

- The visits for the different hashes are separated with no way to aggregate them on the dashboard.
- It is a flawed metric anyways, since it only captures sidebar clicks, not when the users scrolls into it. 

A real solution would be to use custom events to track if the particular hash's content appeared on screen.

Nevertheless, if you prefer following old behaviour, you may change to 

```JS
plausible.init({
  hashBasedRouting: true
})
```

---

~~I intend to change the PR description when I am back home.~~ done with AI while not at home

evnchn:
When this is merged, I intend to export all data off Plausible, then reset site data, modify it to unify all hash based paths, and re-load the modified data. We are rewriting the history as if hash-based navigation logging was never enabled on day 1.

PR #1411, opened 2023-08-14
Add Async to ObservableCollections

thetableman:
This PR adds the ability for an `ObservableCollection` object to call an awaitable as per discussion #1384.

@falkoschindler I wasn't able to find a way for a `ui` element to be run within the called awaitable using a parent slot, I also wasn't sure if you wanted the `arguments` parameter of `handle_event` to be optional hence why I passed a `NoneType`.

PR #5841, opened 2026-02-26
Add favicon support for Windows native

DearRude:
### Motivation

pywebview's `icon` parameter works only on Linux (GTK/Qt). On Windows and macOS, the window icon must typically be set at **packaging time** (via PyInstaller, Nuitka, etc.), so native NiceGUI windows on Windows show the default Python/executable icon instead of the app icon.

**Related issues and discussions:**

- **[NiceGUI Discussion #1745](https://github.com/zauberzeug/nicegui/discussions/1745)** – Main discussion requesting custom window icon support for native mode.
- **[NiceGUI Discussion #1005](https://github.com/zauberzeug/nicegui/discussions/1005)**  Feature request for changing the default app icon; packaging (PyInstaller/Nuitka) discussed as the usual approach.
- **[pywebview #1411](https://github.com/r0x0r/pywebview/issues/1411)** pywebview states icon must be specified when freezing on Windows/macOS; community workaround used `LoadImageW` + `WM_SETICON` to set the icon from a file path.

This PR adds support for using the existing `favicon` parameter from `ui.run()` as the native window icon on Windows when running in native mode. Users can pass a file path (e.g. `favicon='icon.ico'`) and it will be applied to the taskbar and title bar **at runtime**, without requiring packaging changes.

---

### Implementation

- **Reuse of `favicon`**: No new API. When `native=True` and `favicon` is a file path, it is passed to the native process.
- **`ui_run.py`**: Resolves `favicon` to an absolute path when it's a file and passes it to `native_module.activate()`.
- **`native_mode.py`**: `activate()` and `_open_window()` accept a `favicon` argument. On Windows, a `window.events.shown` callback finds the window by title and sets the icon.
- **`window_icon.py`** (new): Windows-only helpers using ctypes:
  - `find_window_by_title()` – uses `EnumWindows` to get the HWND
  - `set_window_icon_windows()` – uses `LoadImageW` and `WM_SETICON` to set the icon
- **Scope**: Windows only. Linux behavior unchanged (pywebview already supports `icon` there).
- **Trade-off**: Icon is set after the window is shown, so there may be a brief moment where the default icon appears before the custom one.

---

### Progress


evnchn:
@DearRude Wow. I nominate this for "top 3 first-time PR". I'm grabbing my Windows laptop for this one.

evnchn:
@DearRude Works for titlebar, not for taskbar though. 

{F3131DEA-6344-4B11-B326-822D335190BA}

Still, pretty good. Is taskbar functionality intended?

evnchn:
{BBABF106-2FA4-4DB1-A48A-FA7792F030CC}

evnchn:
@DearRude Can you fix the pipeline? mypy and pylint isn't very happy...

DearRude:
@evnchn Done. Let me know if I should squash the commits.

falkoschindler:
Thanks, @DearRude!
No need to squash now. We'll do that when we merge on main.

evnchn:
@DearRude I am guessing you came from GitLab which encourages force pushes? Because we don't really do force pushes for GitHub because the GitHub UI naturally encourages against it. Please don't do it unless it's necessary. Thanks. 

Example: GitLab you rebase a branch for latest main; GitHub you merge.

DearRude:
Nice. Thank you for specifying that.

DearRude:
@falkoschindler @evnchn Anything I have to do for now or do you think it is ready to be merged?

evnchn:
@DearRude since this is schedule for 3.10, it is likely that we will look at this after at least 2 weeks from now. 

I'd consider this PR done for now. Please be on the watch out when we come around to review it. Meanwhile you can consider to work on items on the NiceGUI wishlist if you feel like it. MCP integration is a big one but not on the list just now.

evnchn:
@falkoschindler I think it's your turn. Just find me if you want to test on Windows hardware.

falkoschindler:
Thanks for the contribution, @DearRude! This is a nice addition for Windows users.

I'm not very familiar with the Windows side of things, so I used Claude to help review the code. Here are some points that came up — it would be great if you could have a look:

1. **Silent error swallowing in `set_window_property_store`** (`window_icon.py:128`) — The bare `except Exception: return False` around the entire function body hides any failure with no logging. Catching arbitrary exceptions is generally bad practice — tighter try-blocks with more specific exception types would be better, if possible. At minimum, could you log the exception? Otherwise debugging taskbar icon issues will be very hard.

2. **Redundant `sys.platform == 'win32'` checks** (`window_icon.py:34,54,67`) — The caller in `_open_window` (line 57) already guards on `win32`. The three helper functions each re-check internally. Could we check once at the entry point and let the helpers assume Windows?

3. **Duplicate `Path(icon_path).resolve()`** — The path gets resolved in `ui_run.py:243`, then again in `_open_window` at line 65, and a third time in `set_window_icon_windows` at line 56. Once should be enough.

4. **Non-file favicon passed through unnecessarily** (`ui_run.py:243`) — When `favicon` is a URL or data URI, the raw value gets passed all the way to `_open_window` where `helpers.is_file()` catches it. Passing `None` for the non-file case would be cleaner.

5. **Docs could clarify supported formats** — The doc note recommends `.ico`, but it might help to mention that `LoadImageW` with `IMAGE_ICON` only supports `.ico` (and `.cur`/`.ani`). A user passing `.png` would get a silent failure.

6. **Tests** — The tests check function signatures and platform guard early returns, but none of the actual Win32 logic. Since the real behavior can only be verified with an actual native window on Windows, I'd suggest removing them rather than keeping tests that give a false sense of coverage.

7. **Icon handle not freed** (`window_icon.py:57-62`) — `LoadImageW` returns an `HICON` that is never destroyed with `DestroyIcon`. Probably negligible for a single-window app, but wanted to flag it.

8. **`app_id` not sanitized** (`native_mode.py:70`) — `f'nicegui.{title.replace(" ", "_")}'` could exceed the Windows AppUserModelID limit of 128 characters with a very long title. Edge case, but worth considering.

evnchn:
@DearRude @falkoschindler I got some fixes at https://github.com/evnchn/nicegui/tree/evnchn/windows-native-icon but I don't have the time to review them yet.

DearRude:
> it would be great if you could have a look

@falkoschindler Sorry for the late reply and fix. I've updated based on the reviews you provided.

falkoschindler (commented):
Thanks again for the PR, @DearRude!

I pushed 11 commits on top with the following cleanups:

- **Moved all Windows icon logic into `window_icon.py`** -- `native_mode.py` now just calls `window_icon.apply_icon()`, keeping the module focused on window lifecycle
- **Replaced `EnumWindows` title lookup with `window.native.Handle.ToInt32()`** -- pywebview exposes the native handle directly, which is more reliable than matching by title
- **Removed `_find_window_by_title`** entirely (no longer needed)
- **Moved `PROPVARIANT` and `VT_LPWSTR` to module level** for cleaner structure
- **Fixed `PROPVARIANT` struct layout for 32-bit Windows** -- changed `padding` from `c_ulonglong` to `c_void_p` so the struct is 16 bytes on 32-bit and 24 bytes on 64-bit, matching the real Win32 layout
- **Unsubscribed the `on_shown` handler** after first invocation so it only runs once
- **Logged HRESULT values in hex** (`0x80070005`) instead of decimal for easier Windows error lookup
- **Added `# type: ignore[attr-defined]`** for Windows-only ctypes attributes (`HRESULT`, `WINFUNCTYPE`)
- **Simplified and tightened the favicon documentation**
- **Added tests for `_create_windows_app_id`**
- **Consolidated logging** -- consistent caller-level warnings for all property store failures

The code looks good to me now. One caveat: I haven't tested on Windows, so @DearRude please verify that the `window.native.Handle.ToInt32()` approach works correctly for both the title bar icon and the taskbar icon before we merge.

DearRude:
Thanks for the changes @falkoschindler.
I can confirm that it works on Windows.

evnchn (commented):
Overall no problems after 2 rounds of improvements:

<details>
<summary>
Two minor suggestions, non-blocking:
</summary>


**1. VTable index comments**

The magic indices `vtable[2]`, `vtable[6]`, `vtable[7]` are correct per COM ABI but opaque to future readers. Consider inline comments:

```python
set_value = ... (vtable[6])   # IPropertyStore::SetValue
commit = ... (vtable[7])      # IPropertyStore::Commit
release_fn = ... (vtable[2])  # IUnknown::Release
```

**2. Icon correctness is caller responsibility**

The current `LoadImageW` call with `LR_LOADFROMFILE` and size `0,0` is a straightforward pass-through — NiceGUI shouldn't be in the business of validating `.ico` files. Ensuring the icon loads correctly across Windows versions and DPI configurations (e.g. multi-size `.ico` with 16/32/48/256px frames) is the caller's responsibility. The existing warning on `LoadImageW` failure is sufficient.

</details>

evnchn (approved):
Approved (sorry the previous comment should be approve)

PR #5843, opened 2026-02-28
Add @js_action decorator for client-side open/close/toggle optimization

evnchn:
### Motivation

When using `on_click=dialog.open` or similar patterns, the action currently requires a full server round-trip before the UI updates. This introduces noticeable latency, especially on slower connections. By running a JavaScript equivalent on the client side *in addition to* the server-side handler, the UI responds instantly while server state stays in sync.

### Implementation

Introduces a `@js_action` decorator (`nicegui/js_action.py`) that marks element methods (like `open`, `close`, `toggle`) as having a client-side JavaScript equivalent:

- **`JsAction` wrapper**: When a decorated method is accessed on an element instance, it returns a `JsAction` object that carries both the Python callable and a JS handler string.
- **`Element.on()` integration**: When a `JsAction` is passed as a handler, `Element.on()` automatically installs the JS handler for instant client-side feedback and suppresses the redundant server→client update (loopback) since the client already applied the change.
- **`js_action.value(v)` / `js_action.toggle()`**: Convenience constructors for the common patterns of setting or toggling a `model-value` prop.
- **`has_js_action()`**: Utility to detect `JsAction` handlers, used by elements with custom `on_click` wrappers (`Button`, `Chip`, `Item`, `FabAction`, `DropdownButton`) to pass the handler directly to `on()` instead of wrapping it in a lambda.
- **Backward compatible**: Direct calls like `dialog.open()` still work identically. Subclasses that override decorated methods lose the JS action (by design), so custom logic always runs server-side.

Applied `@js_action` to: `Dialog.open/close`, `Menu.open/close/toggle`, `DropdownButton.open/close/toggle`, `Fab.open/close/toggle`.

### Progress

- [ ] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.
- [x] Fix mypy/pylint
- [x] Clicking elsewhere in the `ui.dialog` to close it still causes server involvement
  - [x] Afterwards, the js_action ones are broken.

### DIFF
--- a/nicegui/__init__.py
+++ b/nicegui/__init__.py
@@ -5,6 +5,7 @@ from .client import Client
 from .context import context
 from .element_filter import ElementFilter
 from .event import Event
+from .js_action import JsAction, js_action
 from .nicegui import app
 from .page_arguments import PageArguments
 from .version import __version__
@@ -15,6 +16,7 @@ __all__ = [
     'Client',
     'ElementFilter',
     'Event',
+    'JsAction',
     'PageArguments',
     '__version__',
     'app',
@@ -22,6 +24,7 @@ __all__ = [
     'context',
     'elements',
     'html',
+    'js_action',
     'run',
     'storage',
     'ui',
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -338,7 +338,7 @@ class Element(Visibility):
 
     def on(self,
            type: str,  # pylint: disable=redefined-builtin
-           handler: events.Handler[events.GenericEventArguments] | None = None,
+           handler: events.Handler[events.GenericEventArguments] | JsAction | None = None,
            args: None | Sequence[str] | Sequence[Sequence[str] | None] = None,
            *,
            throttle: float = 0.0,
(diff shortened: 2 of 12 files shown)


evnchn:
Human note: This is either 3.10 or 3.11, so this will take a while. Let's slow-cook this. Meanwhile, it is a sneak peak as to what could be done. 

I would have went for evnchn/nicegui treatment, but I need to show this for the SEO context: This PR would supersede the hack in #5801

evnchn:
image

Server does not need to intervene when opening-closing dialogs 🎉

evnchn:

Closing this one and moving it to the fork (evnchn/nicegui#166) rather than keeping it on the upstream queue:

1. **Audience:** the performance-aware crowd this optimizes for isn't the majority of NiceGUI users, so it's hard to justify the extra API surface (`@js_action`) upstream.
2. **The motivating signal resolved itself:** nicegui.io's INP scores recovered in Google Search Console after the docs revamp. We don't fully understand why, but the empirical result stands — and without the INP pain, the optimization loses its urgency.

If client-side action optimization becomes relevant again, the work continues on the fork and can come back upstream from there.

PR #5825, opened 2026-02-20
Add i18n support for the NiceGUI documentation website

evnchn:
### Motivation

Adds internationalization (i18n) infrastructure to the NiceGUI documentation website, enabling translation of the UI into other languages (starting with zh-CN).

Driven by community interest: https://github.com/zauberzeug/nicegui/discussions/4025#discussioncomment-15863674

### Implementation

**Core i18n module** (`website/i18n.py`):
- `t(english)` function that returns the translated string for the current language, falling back to English
- `language(lang)` context manager using `contextvars.ContextVar` for async-safe language overrides (e.g. for building search indexes per language)
- Translations loaded from `website/translate.csv` (CSV with `en` + language columns)

**Enrolled pages**: main page, intro demos, header, search, examples page, imprint/privacy, style components, documentation overview tiles

**Documentation tree** (`website/documentation/tree.py`):
- `translated_nodes()` returns the sidebar tree with titles translated at render time

**Search indexes** (`website/documentation/search.py`):
- All three endpoints (`search_index.json`, `sitewide_index.json`, `examples_index.json`) accept `?lang=XX`
- Collectors are i18n-aware: `t()` wraps individual fields (heading, title, description) *before* composition, so translations match CSV entries
- Translated indexes are lazily built and cached per language

**Bootstrap tooling** (`i18n_bootstrap.py`):
- Extracts translatable strings from three sources: `t()` calls (AST), documentation registry (runtime), and `examples/*/README.md` files
- Syncs to `website/translate.csv`

**Helper scripts** (pre-existing, included as-is):
- `i18n_translate.py`, `i18n_salvage.py`, `i18n_batch_extract.py`, `i18n_check.py`, `i18n_merge.py`

> **Note:** Code cleanup is still needed. This draft PR is created so that others can see it working and provide early feedback on the approach.
> Not all pages are enrolled yet (e.g. `doc.text()` prose blocks, Map of NiceGUI). Helper scripts need cleanup.

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [ ] The implementation is complete.
- [x] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.
- [ ] Pytests have been added (or are not necessary).
- [ ] Documentation has been added (or is not necessary).

PR #5848, opened 2026-03-04
Add i18n support for the NiceGUI website (focused)

evnchn:
### Motivation

Adds internationalization (i18n) infrastructure to the NiceGUI website, starting with the main landing page and core UI — the highest-value surface area for non-English speakers.

Driven by community interest: https://github.com/zauberzeug/nicegui/discussions/4025#discussioncomment-15863674

This is the focused follow-up to #5825, which served as a proof-of-concept and drew valuable early feedback. This PR is deliberately scoped down to address the key concern raised by @falkoschindler: the original was too large (+8 200 lines) to review in good conscience. This version is 15 files, ~670 insertions.

### Implementation

**Core i18n module** (`website/i18n.py`):
- `t(english)` function that returns the translated string for the current language, falling back to English
- `set_language(lang)` / `get_language()` using `contextvars.ContextVar` for async-safe, per-request language state
- Internal documentation links are rewritten to include the language prefix (e.g. `/documentation/…` → `/zh/documentation/…`)
- Translations loaded from per-language JSON files in `website/translations/`

**Scope**: `main_page.py`, `header.py`, `search.py`, `star.py`, `style.py`, `examples_page.py`, and the documentation intro — the most user-visible surface area, deliberately excluding the full API reference docs

**Language-prefixed routes** (`main.py`):
- Routes registered for each language (e.g. `/de/`, `/zh/documentation/…`)
- Language selector dropdown added to the header

**Translation files** (`website/translations/*.json`): de, ja, ko, zh — 70 strings each, covering the complete translated surface

**Validation tooling** (`website/check_translations.py` + pre-commit hook):
- Extracts all `t()` call keys via AST
- Validates that every key has an entry in every translation file
- Exits non-zero if any keys are missing

**What is intentionally excluded** (vs #5825):
- Full documentation API reference translation (docstrings, `doc.text()` prose)
- Helper scripts (`i18n_bootstrap.py`, `i18n_salvage.py`, etc.)
- The `website/translate.csv` single-file approach (replaced by per-language JSON)
- `<html lang>` / `<link rel="alternate" hreflang>` SEO tags (can follow separately, see TODO in commit)

### Progress


falkoschindler:
Thanks, @evnchn!
Two things before I'll look into it in more detail:

- There's a conflict with main.py. Can you, please, look into it?
- Most multi-line strings have been converted into multiple strings and line breaks have been added. I don't see why. The only diff should be one `t(` at the beginning and one `)` at the end, don't you think?

    Screenshot 2026-03-05 at 09 56 11

falkoschindler:
Oh, if splitting the strings is to handle indentation, we should improve the `t()` function to auto-dedent the input string.

website/i18n.py
```diff
@@ -0,0 +1,70 @@
...
+    English string when no translation is available.  Internal documentation
+    links are rewritten to include the current language prefix.
+    """
+    lang = _current_language.get()
+    if lang == 'en':
+        return english
+    text = _load_translations(lang).get(english, english)
+    prefix = _url_prefix.get()
+    if prefix and '](' in text:
+        text = text.replace('](/', f']({prefix}/')
```
MicaelJarniac, line 69: This feels like it could go wrong easily.

I believe it's to handle Markdown links (`[Example](example)`), but I can see it going wrong unpredictably.

At the same time, I don't know of a better way to implement it. Maybe some RegEx?

website/header.py
```diff
@@ -74,10 +73,39 @@ def add_header(menu: ui.left_drawer) -> ui.button:
...
+                    break;
+                }}
+            }}
+            window.location.href = newPrefix + basePath;
+        ''')
+
+    with ui.button(icon='language').props('flat color=white round').classes('max-[470px]:hidden'):
+        with ui.menu().classes('bg-primary text-white'):
+            for lang, name in SUPPORTED_LANGUAGES.items():
+                ui.menu_item(name, on_click=lambda lang=lang: switch_language(lang))
```
MicaelJarniac, line 111: Lambda inside a loop gives me flashbacks to bugs I introduced, where the value of the variable is always the last iteration of the loop, or something like that.

Not sure if this is the case here, but I'd suggest testing it thoroughly.

MicaelJarniac: I think the bug I coded went something like this:

```py
my_list = ["foo", "bar"]
my_dict = {}

for my_item in my_list:
    def print_item() -> None:
        print(my_item)
    my_dict[my_item] = print_item

my_dict["foo"]()  # bar
my_dict["bar"]()  # bar
```

So again, double-check and be careful.

website/i18n.py
```diff
@@ -0,0 +1,70 @@
...
+from pathlib import Path
+
+# URL slug -> display name
+SUPPORTED_LANGUAGES: dict[str, str] = {
+    'en': 'English',
+    'de': 'Deutsch',
+    'ja': '日本語',
+    'ko': '한국어',
+    'zh': '中文',
+}
```
MicaelJarniac, line 19: This _could_ be automated by scanning the list of translations inside the `website/translations` directory.

That way, to add a new language, one just has to add a `<lang>.json` to that folder, without needing to touch this `i18n.py` file at all, making diffs simpler and avoiding conflicts.

That's more or less what I'm doing in my own web app.

MicaelJarniac (changes requested):

MicaelJarniac:
> Oh, if splitting the strings is to handle indentation, we should improve the `t()` function to auto-dedent the input string.

You're both probably already aware, but Python comes with a [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

website/i18n.py
MicaelJarniac: For this, you could either use a lib (like Babel) to go from lang abbreviation (`de`) to lang name (Deutsch), or you could add a field to the JSON files themselves for the language name, something like `"__LANG_NAME": "Deutsch"`.

MicaelJarniac: The only language that would need to be "hard-coded" to the list would then be English itself, as it doesn't have a JSON file.

falkoschindler:
> You're both probably already aware, but Python comes with a [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

Yes, but the behavior is slightly different. For `textwrap.dedent` to work you need to add `\` behind the opening `'''`. NiceGUI elements like `ui.markdown` or `ui.html`, on the other hand, automatically ignore this empty line when determining the common indentation. This is the behavior I'd prefer for `t()`.

website/header.py
evnchn: Hence the `lang=lang`. Without it, the last `lang` is used.

website/i18n.py
evnchn: `[Example](example)` seems like a good idea actually. What's wrong with it? Let's rewrite all links to not use relative-to-/ and we should be good to go.

MicaelJarniac: A few scenarios where this `str.replace("](/")` approach could break:

**1. Markdown images → broken assets**
If a translated string contains an image reference:
```
![Screenshot](/screenshots/demo.png)
```
It becomes `Screenshot` → 404, since static assets don't have language-prefixed routes.

**2. Links to non-translated paths (static, API, etc.)**
```
[Download the PDF](/static/nicegui-cheatsheet.pdf)
```
→ `](/de/static/nicegui-cheatsheet.pdf)` → 404.

**3. Double-prefixing on repeated calls**
If `t()` is ever called on already-translated text, links get double-prefixed: `](/de/docs)` → `](/de/de/docs)`.

A regex that excludes known static prefixes (e.g. `/static`, `/assets`, `/_`) would be safer, or handling link rewriting at the route/template level instead of string-munging the translations.

MicaelJarniac: This is what my AI agent had to say about this.
I won't say this is certainly going to break, but it feels fragile in my opinion, and would need special care to work correctly.

MicaelJarniac:
> > You're both probably already aware, but Python comes with a [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).
> 
> Yes, but the behavior is slightly different. For `textwrap.dedent` to work you need to add `\` behind the opening `'''`. NiceGUI elements like `ui.markdown` or `ui.html`, on the other hand, automatically ignore this empty line when determining the common indentation. This is the behavior I'd prefer for `t()`.

Now that you mentioned it, I finally noticed there's even a comment on the link I sent about this behavior.

I suppose it's fair to want to avoid the `\` trick.

MicaelJarniac:
Does Google (and other search engines) not use the language header to probe pages for multi language content?

As in, if I had a website with `/foo`, but depending on the user's accepted language header, it displayed translated content, will Google never pick it up?

Because, as far as I can tell, the main reason you're going with `/foo` and `/de/foo` is for SEO.

I personally find it nicer for a website to have the same route (always `/foo`), and provide internationalized content using other mechanisms (cookies, accepted language headers, etc), but I'm super new to this stuff and don't really know how SEO works.

MicaelJarniac:
To answer my own question — **no, Google does not use the `Accept-Language` header to discover translated content.** Googlebot crawls with fixed headers and won't vary them to probe for alternate languages.

Google's [official guidance](https://developers.google.com/search/docs/specialty/international/managing-multi-regional-sites) is explicit: **use different URLs for different languages** (subdirectories, subdomains, or separate domains), and link them together with `<link rel="alternate" hreflang="...">` tags.

Without distinct URLs:
- Google only indexes whichever language it gets on the single URL — the rest are invisible to search.
- You can't use `hreflang` (it requires distinct URLs per language).
- Shared links give unpredictable results depending on the recipient's browser.
- CDN caching becomes fragmented (`Vary: Accept-Language`).

So the `/de/foo` approach in this PR is the right call for SEO. **However — this PR is currently missing the `<link rel="alternate" hreflang="...">` meta tags**, which are essential for search engines to actually understand the relationship between the language variants. Without them, Google may not associate `/foo` and `/de/foo` as the same page in different languages. The PR's TODO mentions this, but it's worth flagging as a priority before merge.

evnchn:
If rewriting the structure is so difficult, can we do /XXXXXX?lang=zh instead?

MicaelJarniac:
Query params (`?lang=zh`) are technically easier to implement, but they largely defeat the SEO purpose that motivated separate URLs in the first place:

- Google treats query params inconsistently — they can be ignored, treated as duplicates, or not indexed reliably
- `hreflang` tags work best with clean, distinct URL paths
- Google's [own docs](https://developers.google.com/search/docs/specialty/international/managing-multi-regional-sites) explicitly recommend subdirectories, subdomains, or separate domains — not query params
- Users can't easily tell the language from the URL, and sharing links becomes less intuitive

If SEO is a goal here (and for the NiceGUI website it probably should be), `/zh/documentation` is the right approach.

MicaelJarniac:
By the way, I'm personally not opposed to `/<lang>/<path>`, it's just that I wish we didn't need to change the URL for SEO to work.

Also, I know it's likely obvious, but I'm alternating between me and my AI agent answering. If it gets annoying, just let me know.

falkoschindler:
Thanks for the work on this, @evnchn — the architecture here (ContextVar-based `t()`, URL prefix routing, JSON translation files, AST validation) is solid and well thought out.

Since the website has been completely redesigned in the meantime (the monolithic `main_page.py` was split into 9 component files, header/footer/design system all changed), I think rebasing this would be harder than starting fresh. What do you think — would it make sense to close this and start a new PR that builds on the learnings from this discussion?

Here's roughly what I have in mind:

**Scope**: Static strings on landing page, header, footer, imprint, examples page, and documentation overview. Documentation content and code examples stay English for now.

**Decisions**:

- URL prefix (`/zh/...`), English at `/` — as discussed, this is the right call for SEO
- English fallback when no translation exists
- JSON files, one per language (`website/translations/zh.json`)
- Starting with Chinese (`zh`)

**Incorporating feedback from this PR**:

1. **Auto-dedent in `t()`** — wrapping triple-quoted strings should just work without splitting them up; `t()` will strip the leading blank line and dedent like `ui.markdown` does
2. **No naive link rewriting** — the `text.replace('](/', ...)` approach breaks images and static assets; we'll handle URL prefixing more carefully
3. **Auto-discover languages** — scan `translations/` directory instead of hardcoding `SUPPORTED_LANGUAGES`; display name can live inside the JSON file itself

Thank you to @MicaelJarniac for the thorough review comments that helped shape this direction. Of course, @evnchn, if you'd prefer to rework this PR yourself against the new website structure, that's also welcome — happy to discuss either way!

evnchn:
Whatever it is, it's not in 3.10. I have said:

> Right, with docs getting a huge update, I think for 3.10 we can shift all the SEO/Perf/Accessibility PRs to 3.11. Easier to bisect if something goes wrong, and lightens the review workload for 3.10 as well. 

I'm kinda open to perf / accessibility (we already did the keyboard nav), but i18n and SEO is definitely not it. Also the GoogleBot would need its chill time to acclimatize to the redesigned NiceGUI homepage anyways.

evnchn:
Closing to rebase onto main — the website redesign in #5910 restructured `website/` (moved sections to `website/components/`, added `design.py`, rewrote `header.py`/`main_page.py`/`style.py`), which conflicts heavily with this branch. I'll rework against the new structure in my fork first, then reopen a fresh PR once it's clean. Thanks for your patience.

PR #5853, opened 2026-03-06
allow `app.clients()` to return all clients when path is `None`

AlePiccin:
### Motivation

Currently, the `app.clients()` method requires a matching `path` string. However, a common use case is needing to get all connected clients across the entire application (e.g., for global server broadcasts, maintenance messages, or gathering metrics) without having to access the underlying `Client.instances` dictionary directly.

This PR solves this by allowing `path` to be `None`, returning all connected clients.

### Implementation

- Changed the type hint of `path` from `str` to `str | None = None`.
- Added an early return block: `if path is None:`, which uses `yield from Client.instances.values()` to efficiently yield all clients.
- The original logic for filtering by a specific path remains completely untouched, ensuring total backward compatibility.
- Updated the docstring to reflect the new default behavior.

### Progress

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -380,15 +380,15 @@ class App(FastAPI):
 
     @staticmethod
     def clients(path: str | None = None) -> Iterator[Client]:
-        """Iterate over connected clients, with a matching path if provided.
+        """Iterate over clients, with a matching path if provided.
 
-        When using `@ui.page("/path")` each client gets a private view of this page.
+        When using ``@ui.page("/path")``, each client gets a private view of this page.
         Updates must be sent to each client individually, which this iterator simplifies.
-        If `path` is `None` (default), it yields all connected clients regardless of their path.
+        If ``path`` is ``None`` (default), it yields all clients regardless of their path.
 
         *Added in version 2.7.0*
 
-        :param path: string to filter clients by (or None to get all clients)
+        :param path: string to filter clients by (or ``None`` to get all clients, *added in version 3.9.0*)
         """
         for client in Client.instances.values():
             if path is None or client.page.path == path:


nicegui/app/app.py
```diff
@@ -379,16 +379,21 @@ def reset(self) -> None:
...
 
-        :param path: string to filter clients by
+        :param path: string to filter clients by (or None to get all clients)
         """
+        if path is None:
+            yield from Client.instances.values()
+            return
+
         for client in Client.instances.values():
             if client.page.path == path:
```
evnchn: Can we use `if path is None or ...` so as to use one for loop for both functionalities?

nicegui/app/app.py
```diff
@@ -379,16 +379,21 @@ def reset(self) -> None:
         self.colors()  # reset colors to default
 
     @staticmethod
-    def clients(path: str) -> Iterator[Client]:
-        """Iterate over all connected clients with a matching path.
+    def clients(path: str | None = None) -> Iterator[Client]:
+        """Iterate over connected clients.
```
evnchn: Iterate over connected clients, with a matching path if provided.

This seems to be more self-descriptive

nicegui/app/app.py
AlePiccin: We could, but the check `path is None` check would be executed n times

tests/test_page.py
```diff
@@ -328,6 +328,25 @@ def page():
...
+    def page():
+        ui.button('add label', on_click=update)
+
+    screen.open('/')
+    screen.switch_to(1)
+    screen.open('/')
+    screen.click('add label')
+    screen.should_contain('added')
+    screen.switch_to(0)
+    screen.should_contain('added')
```
falkoschindler: Looks like this duplication can be avoided using `pytest.parameterize`:
```py
@pytest.mark.parametrize('path', ['/', None])
def test_multicast(screen: Screen, path: str | None):
    ...
```

falkoschindler:
@AlePiccin Thanks for the pull request!

I'd just love to tweak the `clients` method a little bit to
```py
    @staticmethod
    def clients(path: str | None = None) -> Iterator[Client]:
        """Iterate over clients, with a matching path if provided.

        When using ``@ui.page("/path")``, each client gets a private view of this page.
        Updates must be sent to each client individually, which this iterator simplifies.
        If ``path`` is ``None`` (default), it yields all clients regardless of their path.

        *Added in version 2.7.0*

        :param path: string to filter clients by (or ``None`` to get all clients, *added in version 3.9.0*)
        """
        for client in Client.instances.values():
            if path is None or client.page.path == path:
                yield client
```
and add a missing blank line above
```py
@pytest.mark.parametrize('path', ['/', None])
```
(should be done by the auto-formatter).

But I'm not allowed to push to this PR, probably because you're working on main instead of a feature branch. See https://github.com/zauberzeug/nicegui/blob/main/CONTRIBUTING.md#pull-requests for details.

So maybe you can apply above-mentioned changes so that we can merge this PR for the 3.9 release. Thanks!

alexandre-piccin:
> @AlePiccin Thanks for the pull request!
> 
> I'd just love to tweak the `clients` method a little bit to
> 
> ```python
>     @staticmethod
>     def clients(path: str | None = None) -> Iterator[Client]:
>         """Iterate over clients, with a matching path if provided.
> 
>         When using ``@ui.page("/path")``, each client gets a private view of this page.
>         Updates must be sent to each client individually, which this iterator simplifies.
>         If ``path`` is ``None`` (default), it yields all clients regardless of their path.
> 
>         *Added in version 2.7.0*
> 
>         :param path: string to filter clients by (or ``None`` to get all clients, *added in version 3.9.0*)
>         """
>         for client in Client.instances.values():
>             if path is None or client.page.path == path:
>                 yield client
> ```
> 
> and add a missing blank line above
> 
> ```python
> @pytest.mark.parametrize('path', ['/', None])
> ```
> 
> (should be done by the auto-formatter).
> 
> But I'm not allowed to push to this PR, probably because you're working on main instead of a feature branch. See https://github.com/zauberzeug/nicegui/blob/main/CONTRIBUTING.md#pull-requests for details.
> 
> So maybe you can apply above-mentioned changes so that we can merge this PR for the 3.9 release. Thanks!

Are you sure you want to make this change? The `path is None` check would be done multiple times...

falkoschindler:
Yes, the `path is None` check runs on every iteration, but it's a single boolean comparison — virtually free compared to the dictionary iteration and attribute access already happening in the loop. The number of connected clients is small in practice, so there's no measurable performance difference. The benefit is simpler, more readable code with a single code path instead of two.

evnchn:
Good question! I asked Claude Code to benchmark this -- the `path is None` check is a single `IS_OP` bytecode instruction (pointer comparison), and it short-circuits so `client.page.path == path` is never evaluated when `path` is `None`.

Actual overhead of the single-loop version vs the early-return:

| Clients | Overhead per call |
|---------|-------------------|
| 10      | +0.04 us          |
| 100     | +0.47 us          |
| 1,000   | +5.1 us           |
| 10,000  | +47 us            |

Even at 10K clients, the extra cost is 47 microseconds -- roughly 0.001% of the downstream WebSocket I/O work. The single-loop version is cleaner and the performance difference is well below anything measurable in practice.

<details>
<summary>Full benchmark report</summary>

## Context

The PR currently has an early-return with `yield from` for `path=None`, and a separate loop for path filtering:

```python
def clients(path: str | None = None) -> Iterator[Client]:
    if path is None:
        yield from Client.instances.values()
        return
    for client in Client.instances.values():
        if client.page.path == path:
            yield client
```

Falko suggested simplifying to a single unified loop:

```python
def clients(path: str | None = None) -> Iterator[Client]:
    for client in Client.instances.values():
        if path is None or client.page.path == path:
            yield client
```

AlePiccin raised a fair question about whether the repeated `path is None` check per iteration is a concern. Let's find out.

## Short Answer: No, it doesn't matter.

### Absolute numbers at every scale

| Clients | `yield from` (early return) | Single loop (`is None or`) | Overhead per call |
|---------|----------------------------|---------------------------|-------------------|
| 10      | 0.35 us                    | 0.39 us                   | **+0.04 us**      |
| 100     | 1.68 us                    | 2.15 us                   | **+0.47 us**      |
| 1,000   | 12.9 us                    | 18.0 us                   | **+5.1 us**       |
| 10,000  | 133 us                     | 179 us                    | **+47 us**        |

The worst case (10,000 simultaneously connected clients) costs an extra **47 microseconds** -- 0.047 milliseconds. For context, the work this function enables (sending WebSocket messages to each client, re-rendering UI) is on the order of 50-500 us *per client*. At 10,000 clients, you're looking at 0.5-5 *seconds* of actual I/O. The 47 us is noise -- roughly 0.001% of total work.

### Why there's any difference at all

The bytecode tells the story:

- `yield from` delegates to C-level `tp_iternext` on the `dict_values` object -- pure C, no Python bytecode per element.
- The single loop adds 3 bytecode ops per iteration (`LOAD_FAST`, `LOAD_CONST None`, `IS_OP`), but `is None` is one of the cheapest possible Python operations -- it's a pointer comparison, no method call, no `__eq__`.

When `path is not None`, both versions are nearly identical because `is None` short-circuits to false and both then do the same `client.page.path == path` attribute lookup + comparison. The overhead there is ~8% relative, ~20 us at 10K clients.

### Realistic perspective

A NiceGUI app with 10,000 simultaneous WebSocket clients is already a heavy-duty deployment. At that scale, your bottleneck is network I/O, not a pointer comparison in a Python loop. At the typical 10-100 client range, we're talking about 0.04-0.47 microseconds of overhead -- less time than a single `print()` call.

## Three Versions Compared

### Version A: Early-return (current PR)

```python
def clients(path: str | None = None) -> Iterator[Client]:
    if path is None:
        yield from Client.instances.values()
        return
    for client in Client.instances.values():
        if client.page.path == path:
            yield client
```

- Fastest for `path=None` (C-level iteration via `yield from`)
- Two separate code paths; the `yield from` + bare `return` in a generator is a non-obvious pattern
- Duplicates the `for client in Client.instances.values()` iteration setup

### Version B: Single loop (falkoschindler's suggestion)

```python
def clients(path: str | None = None) -> Iterator[Client]:
    for client in Client.instances.values():
        if path is None or client.page.path == path:
            yield client
```

- Cleanest and most readable -- one loop, one condition, done
- Pays a trivial `is None` pointer check per iteration
- Easiest to maintain and least likely to diverge if logic changes

### Version C: Hoisted dispatch (no per-iteration check, no generator trick)

```python
def clients(path: str | None = None) -> Iterator[Client]:
    if path is None:
        return iter(Client.instances.values())
    return (c for c in Client.instances.values() if c.page.path == path)
```

- Zero per-iteration overhead for `path=None` (returns the dict_values iterator directly)
- No duplicated loop body
- **Subtle difference**: this is a regular function returning an iterator, not a generator function. Callers won't notice (both are `Iterator[Client]`), but exceptions inside the genexpr are raised at consumption time rather than at call time -- which is the same behavior as a generator anyway, so no practical difference.
- Arguably *less* clear than Version B: two `return` statements returning different iterator types, and the genexpr syntax is less scannable than an explicit `for`/`if`/`yield`.

## Recommendation

**Go with Version B** (falkoschindler's single loop). The readability gain is real; the performance cost is fictional at any practical scale. The concern is technically correct -- the check *is* repeated -- but `is None` is a single `IS_OP` bytecode instruction comparing two pointers. At 100 clients it costs 0.47 microseconds. There is no scenario where this is the bottleneck.

Version C exists as a "have your cake and eat it too" option, but it sacrifices readability for an optimization that saves microseconds in a function whose downstream work takes milliseconds to seconds.

</details>

evnchn:
@AlePiccin pylint and pre-commit is sad again. Can you fix it, and try to install pre-commit (and run mypy and pylint) yourselves if possible? Though using CI to gauge is fine if local dev environment would mean it's tough to do, since public repo run minute is free.

alexandre-piccin:
I’m currently editing directly on GitHub rather than from a local clone, so I can’t run pre-commit locally at the moment. I’ll use the CI output to fix the remaining issues.

nicegui/app/app.py
```diff
@@ -379,16 +379,18 @@ def reset(self) -> None:
...
         When using `@ui.page("/path")` each client gets a private view of this page.
         Updates must be sent to each client individually, which this iterator simplifies.
+        If `path` is `None` (default), it yields all connected clients regardless of their path.
 
         *Added in version 2.7.0*
 
-        :param path: string to filter clients by
+        :param path: string to filter clients by (or None to get all clients)
         """
+
```
evnchn: nit: extra new line?

falkoschindler: Oh yes!
```suggestion
```

PR #5854, opened 2026-03-07
Minor changes to the SVG clock

rolfn:
### Motivation

Two minor changes:

* `continuous_minute = False` makes the minute jump instead of continuously scrolling. This reflects the behavior of such a clock in the real world ("Bahnhofsuhr").

* `viewBox` allows correct scaling of the ui.html element using classes or style attributes.

### DIFF
--- a/examples/svg_clock/main.py
+++ b/examples/svg_clock/main.py
@@ -9,12 +9,16 @@ def build_svg() -> str:
 
         Original was borrowed from https://de.m.wikipedia.org/wiki/Datei:Station_Clock.svg.
     """
+    continuous_minute = True
+
     now = datetime.now()
     seconds_angle = now.second / 60 * 360
-    minutes_angle = now.minute / 60 * 360 + seconds_angle / 60
+    minutes_angle = now.minute / 60 * 360
+    if continuous_minute:
+        minutes_angle += seconds_angle / 60
     hours_angle = now.hour / 12 * 360 + minutes_angle / 12
     return f'''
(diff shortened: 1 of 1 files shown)


falkoschindler:
Ooops, somehow I mixed up your main with our main branch and merged accidentally. Anyway, thanks for the pull request, @rolfn! I decided to remove the always-true parameter `continuous_minute` and use the new non-continuous behavior by default. 👍🏻

evnchn:
@falkoschindler @rolfn I thought the consensus was all 3 hands be smooth-moving in https://github.com/zauberzeug/nicegui/pull/5288? IMO having smooth-moving hands is the "cheapest and most standard" way to make a clock and we should do that. The Bahnhofsuhr is just an exception. 

If we diff https://github.com/zauberzeug/nicegui/commit/d2474291ea6bae2d98b3f9d9d829dc5afe3da521 and https://github.com/zauberzeug/nicegui/commit/41181a41707ad60b80a9005cd5b12966ee9ee0e7 then the changes are three:

1. Minutes hand smoothness removed (line 14)
2. SVG made responsive (line 16)
3. Clock element gets a width class (line 48)

which I think we should adopt just 2 and 3, not 1 IMO.

falkoschindler:
@evnchn Yes, I see your point regarding change 1. But since the design is very closely inspired by the iconic German station clocks (see http://www.digital-clock.org/db/), I get that users don't expect continuous minutes and agreed to merge the change. In the end it's just an example and should be more about integrating SVG with NiceGUI than the exact outcome.

falkoschindler:
By the way, it looks like @rolfn is our 200th contributor! 🥳

PR #5855, opened 2026-03-09
Add `make_sortable()` for drag-and-drop sorting on container elements

falkoschindler:
### Motivation

Feature request: #932
Supersedes PRs #4656 and #5464

Users have long requested built-in drag-and-drop sorting support. PR #4656 introduced SortableJS as a dedicated `ui.sortable` element, and #5464 was another experiment along the same lines. After extensive discussion in #932, we concluded that "sortable" is a feature, not an element — a `ui.column` that can be reordered is still a column.

This PR implements that conclusion: a `.make_sortable()` method on container elements (`ui.column`, `ui.row`, `ui.card`, `ui.grid`, `ui.list`, `ui.expansion`, `ui.scroll_area`). It wraps SortableJS and keeps the server-side element tree in sync automatically. The container keeps its original type — a `ui.column` stays a `ui.column`.

### Implementation

- **`SortableElement` mixin** adds `make_sortable()` to container elements, returning a `Sortable` controller with `enable()`, `disable()`, ~~`destroy()`~~, and read/write properties (`animation`, `handle`, `group`, `ghost_class`). ~~An `options` property (backed by `ObservableDict`) auto-pushes changes to the client, following the same pattern as AG Grid.~~
- **`Sortable` controller** ~~initializes SortableJS on the client, syncs the client-side element tree on drop (preventing snap-back), and fires a `sortend` event with element references (`item`, `source`, `target`) and indices.~~ is a hidden Vue component (`Element` subclass) that manages SortableJS via a `.js` component file. Options are synced reactively through props. The JS-side `onEnd` callback syncs the client-side element tree (preventing snap-back) and dispatches a `sortend` CustomEvent on the container element with element references (`item`, `source`, `target`) and indices.
- **`SortableEventArguments`** dataclass provides typed event args.
- **SortableJS is bundled** as an ESM module (same rollup pattern as ag-grid).
- **Auto-cleanup**: ~~destroying the container element automatically destroys the sortable.~~ Vue's `unmounted()` lifecycle hook automatically destroys the SortableJS instance — no manual cleanup needed.
- **Duplicate protection**: calling `make_sortable()` twice raises `RuntimeError`.

### Progress

### DIFF
--- a/website/documentation/content/sortable_documentation.py
+++ b/website/documentation/content/sortable_documentation.py
@@ -25,16 +25,16 @@ def cross_container() -> None:
     with ui.row():
         with ui.card():
             ui.label('Card 1').classes('font-bold')
-            with ui.column() as col1:
+            with ui.column() as column1:
                 for name in ['Alice', 'Bob', 'Carol']:
                     ui.label(name).classes('cursor-grab active:cursor-grabbing')
         with ui.card():
             ui.label('Card 2').classes('font-bold')
-            with ui.column() as col2:
+            with ui.column() as column2:
                 for name in ['Dave', 'Eve', 'Frank']:
                     ui.label(name).classes('cursor-grab active:cursor-grabbing')
-    col1.make_sortable(group='shared')
-    col2.make_sortable(group='shared')
+    column1.make_sortable(group='shared')
+    column2.make_sortable(group='shared')
 
 
 @doc.demo('Drag Handle', '''


falkoschindler:
There are still two review comments from Claude I need to look into. Apart from that the PR is complete. Should be good for early 3.11.

---

1. **Sortable element placement and lifecycle**

   `Sortable(self, ...)` in `make_sortable()` creates the element in whatever build context is current at call time — typically the page root, **not** as a child of the container being made sortable.

   If the container is programmatically deleted (e.g. `container.delete()`), the `Sortable` element is **not** automatically cleaned up because it isn't a descendant. The SortableJS instance would linger with a stale DOM reference until the page is closed.

   **Suggestion:** Either create the Sortable as a child of the container (using `with self:`) or register an `on_delete` callback on the container to destroy the Sortable.

   Note: Option A would make the Sortable a visible child in the slot, which would be affected by sorting. So Option B (on_delete callback) or using a non-default slot might be better. Worth thinking through.

2. **Missing test coverage**
   - **`enable()` / `disable()`**: No test verifies these methods work
   - **Runtime property changes**: No test for changing `animation`, `handle`, `group`, or `ghost_class` after initialization
   - **Duplicate `make_sortable()` call**: No test verifying the `RuntimeError` is raised

evnchn:
https://github.com/zauberzeug/nicegui/pull/5855/commits/c4b88e8e001bd902f8252c078f9c060e3669d50d: bonus item for later: modernize the example. Though this PR is complex enough already...

falkoschindler:
Good idea! We actually discussed this and intentionally kept the example as-is for now. It's currently the only example showing a DIY drag-and-drop solution using HTML5 events, which is still useful for cases where you need more control (custom acceptance logic, visual feedback, etc.). The `make_sortable()` docs already cover the common use cases with inline demos. But agreed it could be worth revisiting — maybe as a separate follow-up after this PR lands.

evnchn (approved):
Reviewed by @evnchn with Claude Code. evnchn directed the review strategy, identified the cross-container demo UX issue (card titles being draggable), chose the `ui.column` fix over `handle`, and validated all changes before push. LGTM — three small commits pushed to the branch.

<details>
<summary>Review journey</summary>

### What was reviewed
- All new/modified source: `SortableElement` mixin, `Sortable` controller, `sortable.js`, bundling setup, tests, docs, container element diffs

### Investigation results
- **`self.update()` not needed** on property setters — NiceGUI props auto-sync. Confirmed by writing `test_disable_and_enable` (disable → drag → no effect; enable → drag → works)
- **`pop()` pattern wrong for setters** — SortableJS needs explicit `null` via `.option()` to clear a value; removing the key means the watcher never fires. Confirmed by `test_handle_property` (set handle → clear to None → drag works only with direct assignment, not pop)
- **Ablation-tested both new tests**: gutting `disable()` or `handle` setter → test fails as expected

### Commits pushed
1. **NaN guard in `sortable.js` `onEnd`** — `parseInt` on non-NiceGUI child IDs returns `NaN`, would silently corrupt slot arrays
2. **Two new tests** — `test_disable_and_enable` and `test_handle_property` cover runtime property mutation (not covered by existing tests)
3. **Cross-container demo** — wrapped draggable items in `ui.column` so card titles stay fixed, avoiding `handle` before that concept is introduced in the next demo

### Not flagged (confirmed fine)
- Test `card2` lambda with 0 args — `handle_event` introspects signature, works by design
- No `self.update()` anywhere — verified unnecessary via tests
- Existing setter style (simple assignment) — verified correct via tests

</details>

falkoschindler:
Thanks, @evnchn! Great feedback!
I just felt the urge to simplify the pytests a little - now that we have yet another test checking the order of sortable elements.

PR #5856, opened 2026-03-09
Remove `KWONLY_SLOTS` constant and inline `@dataclass` arguments

falkoschindler:
### Motivation

Now that Python 3.9 support has been dropped, the version guard in `KWONLY_SLOTS` is always true.
The indirection makes the decorator less readable without providing any benefit.

### Implementation

- Replaced all `@dataclass(**KWONLY_SLOTS)` usages with `@dataclass(kw_only=True, slots=True)`,
- removed the corresponding imports,
- deleted `nicegui/dataclasses.py`, and
- updated CONTRIBUTING.md to reflect the new convention.

### Progress

### DIFF
--- a/nicegui/dataclasses.py
+++ b/nicegui/dataclasses.py
@@ -1,0 +1,10 @@
+from .helpers import warn_once
+
+# DEPRECATED: This module will be removed in NiceGUI 4.0
+
+
+def __getattr__(name: str) -> dict:
+    if name == 'KWONLY_SLOTS':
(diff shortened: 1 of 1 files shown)


evnchn:
Do we even need the deprecation? Because I don't think it's part of the public API. 

Other than that, nothing remarkable.

falkoschindler:
@evnchn I thought so too. But nicegui-highcharts is using it for its dataclasses, causing failing tests. So we at least need to keep it for a short migration period. So we can as well wait for 4.0 before removing it.

PR #5857, opened 2026-03-09
reuse parent app's SessionMiddleware in `ui.run_with()`

### LINKED ISSUE #2578: Attaching to a FastAPI instance with another storage_secret breaks session storage

falkoschindler:
### Motivation

Fixes #2578. When using `ui.run_with()` with a parent FastAPI app that already has `SessionMiddleware`, NiceGUI's session storage silently fails because both apps fight over the same session cookie with different secrets.

### Implementation

Instead of adding a second `SessionMiddleware`, `set_storage_secret` now checks whether the parent app already has one. If so, NiceGUI reuses the parent's secret and only adds its own `RequestTrackingMiddleware`. If the user explicitly passed a different `storage_secret`, a warning is logged and the parent's secret takes precedence (first comes, first serves).

This is an alternative to #5703 with a simpler approach: no conflict detection, no cookie renaming, no flags — just reuse what's already there.

### Progress

### DIFF
--- a/nicegui/storage.py
+++ b/nicegui/storage.py
@@ -51,7 +51,7 @@ def set_storage_secret(storage_secret: str | None = None,
     A warning is issued if a different ``storage_secret`` was provided explicitly.
     """
     if parent_app and (parent_sm := next((m for m in parent_app.user_middleware if m.cls == SessionMiddleware), None)):
-        parent_secret = str(parent_sm.kwargs.get('secret_key') or parent_sm.args[0])
+        parent_secret = str(parent_sm.kwargs['secret_key'] if 'secret_key' in parent_sm.kwargs else parent_sm.args[0])
         if storage_secret is not None and storage_secret != parent_secret:
             helpers.warn_once('Ignoring storage_secret because the parent app already has SessionMiddleware')
         if session_middleware_kwargs:


evnchn:
@falkoschindler This is about what #5703 used to look like (at least line-count speaking), before I consider the case where NiceGUI and FastAPI have differing secret keys, but they would have still worked together considering that they have different `session_cookie` cookie. 

And when there's 4 cases of [session_cookie match / unmatch] * [storage_secret match / unmatch] I don't see a way to boil #5703 to become simple. Maybe we can, but should we? 

The question becomes, are we comfortable with a slightly higher false positive rate by neglecting check session_cookie? If that means higher library maintainability while the code still works for 99.99% of the users, maybe it's worth it?

falkoschindler:
@evnchn Thanks for thinking through the edge cases. This PR sidesteps the 4-case matrix from #5703 entirely: if the parent app has `SessionMiddleware`, we simply reuse it and only add `RequestTrackingMiddleware`. No conflict detection, no cookie renaming, no flags.

The trade-off is that we don't support the "different secrets + different cookies" case where both could technically coexist. But I think that's the right call:

- It covers 99.99% of real usage (the original issue #2578 and all its cases)
- If someone truly needs independent sessions, they can add `SessionMiddleware` directly to NiceGUI's app before calling `ui.run_with()`
- The resulting code is ~15 lines instead of ~60, and easy to understand at a glance

@denniswittich Since you opened #2578 originally -- would this simpler approach cover your use cases? The idea is: when the parent FastAPI app already has `SessionMiddleware`, NiceGUI reuses its secret and session cookie rather than adding its own. If you pass a different `storage_secret`, we warn and ignore it (parent's middleware wins).

evnchn:
Hot take: like the `KWONLY_SLOTS`, we should consider out-of-library use cases. 

IMHO we should totally not condone setting NiceGUI-land `session_cookie` because if our JS cookie jumps around, it may cause headache for NiceGUI custom element developers, who expect the `session` cookie to always be there with NiceGUI cookie contents. 

Though we do compromise 0.01% of the edgest of edge case that even #2578 did not cover, we make the codebase Nice, and the custom element development Nice.

falkoschindler:
Ok, @denniswittich also agrees with the 99.99% solution. We can always add complexity later if there is demand.

PR #5866, opened 2026-03-10
Add native window event IPC bridge (shown, resized, drop, etc.)

falkoschindler:
### Motivation

When running NiceGUI with `native=True`, the app window is powered by pywebview in a separate process. Currently, communication only goes one way: the main process can tell the window what to do, but the window can't report back what's happening to it.

This means there's no way to know when the user minimizes, maximizes, or resizes the window — or when they drag and drop files onto it. This PR adds that missing event channel.

Relates to #5852 and supersedes #3378 with a fresh implementation. Uses `Pipe` instead of `Queue` for the IPC transport, which avoids idle CPU overhead from polling.

### Usage

```python
from nicegui import app, ui

app.native.on('minimized', lambda: print('Window minimized'))
app.native.on('maximized', lambda: print('Window maximized'))
app.native.on('resized', lambda e: print(f'{e.args["width"]}x{e.args["height"]}'))
app.native.on('drop', lambda e: print(f'Dropped files: {e.args["files"]}'))

ui.run(native=True)
```

Handlers can be sync or async and optionally accept a `NativeEventArguments` parameter.

Supported events: `shown`, `loaded`, `minimized`, `maximized`, `restored`, `resized`, `moved`, `closed`, `drop`.

### Implementation

The pywebview subprocess sends events through a `multiprocessing.Pipe` back to the main process, where a background thread receives them and dispatches to registered handlers.

- **Pipe transport** (`native.py`): A one-directional `Pipe(duplex=False)` is created alongside the existing method/response queues. The receiver stays in the main process; the sender is passed to the subprocess.
- **Event binding** (`native_mode.py`): `_bind_pywebview_events()` hooks into pywebview's window lifecycle callbacks and sends each event as a `{'type': name, 'args': kwargs}` dict through the pipe. For `drop`, it injects a small JavaScript snippet after the page loads that captures `dragover`/`drop` and extracts file paths via pywebview's `pywebviewFullPath`.
- **Event manager** (`event_manager.py`, new file): A daemon thread blocks on `pipe.recv()` and dispatches incoming events to registered handlers. A sentinel `None` value is used for clean shutdown.
- **Public API** (`native_config.py`): `app.native.on(event_type, handler)` delegates to the event manager.

### Known limitation: per-element drop zones

The `drop` event fires globally for the entire window. Targeting a specific UI element (e.g. "only trigger when files are dropped on this upload area") is not yet supported. The difficulty is that browser-side drag events (which know the target element) travel via WebSocket, while the native `drop` event (which carries the file paths) travels via the subprocess pipe — and these two channels race against each other. Possible follow-up approaches include sending drop coordinates through the pipe for server-side hit testing, or resolving the target element entirely in JavaScript before sending through the pipe.

### Progress

- [x] Pytests are out of scope (native mode is currently untested).

### DIFF
--- a/nicegui/native/event_manager.py
+++ b/nicegui/native/event_manager.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections import defaultdict
 from threading import Thread
 
+from .. import core
 from ..events import Handler, NativeEventArguments, handle_event
 from . import native
 
@@ -31,6 +32,11 @@ class EventManager:
             except OSError:
                 pass  # the pipe might already be closed
 
+    def _dispatch(self, data: dict) -> None:
+        args = NativeEventArguments(type=data['type'], args=data.get('args', {}))
+        for handler in self._handlers.get(args.type, []):
+            handle_event(handler, args)
+
     def _event_loop(self) -> None:
         assert native.event_receiver is not None
         while True:
@@ -40,9 +46,10 @@ class EventManager:
                 break  # the pipe was closed
             if data is None:
                 break
-            args = NativeEventArguments(type=data['type'], args=data.get('args', {}))
-            for handler in self._handlers.get(args.type, []):
-                handle_event(handler, args)
+            if core.loop and core.loop.is_running():
+                core.loop.call_soon_threadsafe(self._dispatch, data)
+            else:
+                self._dispatch(data)
 
 
 event_manager = EventManager()


falkoschindler:
For the record, this is what I tried to create a local drop zone like in #3378 (without success):

```py
class DropZone(ui.card):

    def __init__(self, title: str, *, on_drop: Callable[[], list[Path]]) -> None:
        super().__init__()

        self.on('dragenter', lambda: self.classes('bg-red-100'))
        self.on('dragleave', lambda: self.classes(remove='bg-red-100'))

        with self:
            ui.label(title).classes('m-auto')

        def handle_drop(files: events.NativeEventArguments) -> None:
            if 'bg-red-100' in self.classes:
                on_drop([Path(file_) for file_ in files.args['files']])
                self.classes(remove='bg-red-100')

        app.native.on('drop', handle_drop)

DropZone('Zone A', on_drop=lambda files: print('Dropped in A:', files)).classes('w-64 h-64')
DropZone('Zone B', on_drop=lambda files: print('Dropped in B:', files)).classes('w-64 h-64')
```

EmberLightVFX:
Glad to see this getting some development! Seems like a leaner way that in my old PR but getting a per-element drop seems like the final piece to try and solve.

falkoschindler:
Hi @EmberLightVFX, would you mind giving this PR a quick test on Windows? We've only tested on macOS so far, and since pywebview uses a different backend on Windows (EdgeChromium), it'd be great to verify things work there too.

A minimal smoke test would be something like:

```python
from nicegui import app, ui

app.native.on('shown', lambda: print('shown'))
app.native.on('resized', lambda e: print(f'resized: {e.args["width"]}x{e.args["height"]}'))
app.native.on('minimized', lambda: print('minimized'))
app.native.on('drop', lambda e: print(f'drop: {e.args["files"]}'))

ui.label('Check the terminal for native window events.')

ui.run(native=True)
```

Mainly curious about:
1. Do the lifecycle events (`shown`, `resized`, `minimized`, etc.) fire as expected?
2. Does dragging a file onto the window print the file paths?

Thanks!

evnchn:
Two small fixes pushed to this branch:

1. **`f151f86`** — Added missing "drop" to the `on()` docstring (was listed in website docs but not in the code docstring).
2. **`9ed4dea`** — Dispatched native events onto the event loop thread via `call_soon_threadsafe`, fixing a thread-safety issue with async handlers.

<details>
<summary>Analysis and research details</summary>

### Testing

Tested on Windows 11 with Python 3.12 and pywebview 6.1. Wrote a NiceGUI native app registering handlers for all event types, then automated window interactions (minimize, maximize, restore, resize, move, close) via Win32 API calls from PowerShell. All events fire correctly with both sync and async handlers.

### Thread-safety issue (commit 2)

`EventManager._event_loop` runs in a daemon thread. When an async handler is registered, the call chain is:

```
_event_loop (daemon thread)
  → handle_event(handler, args)
    → handler() returns a coroutine
    → background_tasks.create(...)
      → core.loop.create_task(...)   ← not thread-safe
      → running_tasks.add(task)      ← set mutation from wrong thread
```

`loop.create_task()` internally calls `loop.call_soon()`, which the [Python docs explicitly state is not thread-safe](https://docs.python.org/3/library/asyncio-eventloop.html). The docs also state: ["Almost all asyncio objects are not thread safe"](https://docs.python.org/3/library/asyncio-dev.html), and that [`call_soon_threadsafe()` must be used when scheduling from another thread](https://docs.python.org/3/library/asyncio-eventloop.html).

Cross-thread `create_task` has caused real crashes in CPython itself due to races on the shared `_all_tasks` WeakSet — see [CPython #89125](https://github.com/python/cpython/issues/89125) (`IndexError: pop from empty list`) and [CPython #36607](https://bugs.python.org/issue36607) (`RuntimeError: Set changed size during iteration`).

The fix uses `core.loop.call_soon_threadsafe(self._dispatch, data)` so that `handle_event` (and any subsequent `create_task`) runs on the event loop thread.

### Duplicate events observation

During testing, every event appeared to fire twice. Investigation traced this to NiceGUI's script mode re-executing the user's file via `runpy.run_path` in `ui_run.py:153`, which re-runs top-level `app.native.on()` calls and double-registers handlers. This is expected behavior — users who need single-execution should use `root=` or `@ui.page`.

</details>

---

*I am Claude Code acting on behalf of @evnchn.*

Issue #5804, opened 2026-02-18
Outbox.stop() doesn't wake the sleeping loop


Denubis:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
# /// script
# dependencies = [
#     "nicegui @ git+https://github.com/zauberzeug/nicegui.git",
#     "httpx",
#     "pytest",
#     "pytest-asyncio",
# ]
# requires-python = ">=3.11"
# ///

from __future__ import annotations

import asyncio
import sys
import time
from pathlib import Path

import pytest

_APP_SOURCE = """\
from nicegui import ui

@ui.page("/")
async def index():
    ui.label("Hello")

ui.run()
"""


def _find_outbox_task(client_id: str) -> asyncio.Task | None:  # type: ignore[type-arg]
    for t in asyncio.all_tasks():
        if not t.done():
            name = t.get_name() or ""
            if "outbox loop" in name and client_id in name:
                return t
    return None


@pytest.fixture(scope="module")
def app_file(tmp_path_factory: pytest.TempPathFactory) -> Path:
    p = tmp_path_factory.mktemp("app") / "app.py"
    p.write_text(_APP_SOURCE)
    return p


@pytest.mark.asyncio
async def test_outbox_stop_takes_one_second(
    app_file: Path,
) -> None:
    """stop() should return promptly but the loop lingers ~1s."""
    from nicegui.testing.user_simulation import user_simulation

    async with user_simulation(main_file=str(app_file)) as user:
        await user.open("/")
        assert user.client is not None
        client_id = user.client.id
        await asyncio.sleep(0.3)

        outbox_task = _find_outbox_task(client_id)
        assert outbox_task is not None, (
            f"No outbox loop task found for client {client_id}"
        )

        start = time.monotonic()
        user.client.outbox.stop()

        while not outbox_task.done() and time.monotonic() - start < 2.0:
            await asyncio.sleep(0.01)
        elapsed = time.monotonic() - start

        print(f"\n  Outbox stop elapsed: {elapsed:.3f}s")
        print("  Expected with fix: <0.1s")
[...]

PR #5867, opened 2026-03-11
Add opt-in diagnostics endpoint and ui.diagnostics_view element

Denubis:
### Motivation

When debugging NiceGUI applications, runtime state (asyncio tasks, memory usage, client connections, server configuration) requires ad-hoc inspection code.

This adds an opt-in `diagnostics=True` parameter to `ui.run()` that registers a JSON endpoint at `/_nicegui/diagnostics`, and a `ui.diagnostics_view()` element for in-page display.

We have not pre-discussed whether `ui.diagnostics_view` belongs in core (per CONTRIBUTING.md step 1 for new elements). Happy to move it to a separate package if you'd prefer. (Noting that this is stemming from the discussion in #5660.) I renamed it to diagnostics from "health" due to a bit of concept mismatch.

### Implementation

- `diagnostics` field on `AppConfig`, threaded through `ui.run()` → `add_run_config()`
- `nicegui/diagnostics.py`: `collect_snapshot()` returns task counts grouped by coroutine, memory metrics (RSS from `/proc/self/status` and `resource.getrusage`), per-client detail (element count, outbox queue
  lengths, socket status), and server config (`async_handlers`, transports, `reconnect_timeout`, `binding_refresh_interval`)
- `/_nicegui/diagnostics` endpoint registered only when `diagnostics=True`; module not imported otherwise
- `ui.diagnostics_view()` composes Log + Button + optional Timer, calls `collect_snapshot()` directly (no HTTP request)
- `user_simulation` gains `**run_kwargs` forwarding to `ui.run()` to enable testing with `diagnostics=True`
- Memory metrics degrade gracefully on non-Linux/non-POSIX platforms via `ImportError` and `contextlib.suppress(OSError)` guards
- 15 diagnostics tests + 2 `user_simulation` validation tests, all using `User` fixture

### Progress

- [ ] The implementation is complete. (Nope, this is phase 1 draft, just so that @evnchn can comment)

### DIFF
--- a/nicegui/diagnostics.py
+++ b/nicegui/diagnostics.py
@@ -1,0 +1,141 @@
+"""Runtime diagnostics for NiceGUI applications."""
+import asyncio
+import collections
+import contextlib
+import datetime
+import sys
+from typing import Any
+
+from starlette.requests import Request
+from starlette.responses import JSONResponse
+
+from . import background_tasks, core
+from .client import Client
+
+# NOTE: resource module is POSIX-only; gracefully degrade on other platforms
+try:
+    import resource
+except ImportError:
+    resource = None  # type: ignore[assignment]
+
+
+def _collect_task_summary() -> dict[str, Any]:
+    """Collect asyncio task summary grouped by coroutine name."""
+    tasks = asyncio.all_tasks()
+    counter: collections.Counter[str] = collections.Counter()
+    names_by_group: dict[str, list[str]] = {}
+    for task in tasks:
+        qualname = getattr(task.get_coro(), '__qualname__', '')
+        # NOTE: extract last two dotted segments for readable grouping
+        key = '.'.join(qualname.rsplit('.', 2)[-2:])
+        counter[key] += 1
+        names_by_group.setdefault(key, []).append(task.get_name())
+    by_coroutine = {
+        key: {'count': count, 'names': names_by_group[key]}
+        for key, count in counter.most_common()
+    }
+    return {
+        'total': len(tasks),
+        'by_coroutine': by_coroutine,
+    }
+
+
+def _collect_memory() -> dict[str, Any]:
+    """Collect memory usage metrics with source labels."""
+    if resource is not None:
+        peak_rss_raw = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+        # NOTE: on Linux, ru_maxrss is in KB; on macOS it is in bytes
+        peak_rss_bytes = peak_rss_raw * 1024 if sys.platform == 'linux' else peak_rss_raw
+        peak_rss_source = 'resource.getrusage(RUSAGE_SELF).ru_maxrss'
+    else:
+        peak_rss_bytes = None
+        peak_rss_source = 'resource module not available (non-POSIX platform)'
+
+    current_rss_bytes = None
+    current_rss_source = '/proc/self/status not available (non-Linux platform)'
+    with contextlib.suppress(OSError):
+        with open('/proc/self/status', encoding='utf-8') as f:
+            for line in f:
+                if line.startswith('VmRSS:'):
+                    # NOTE: VmRSS line format is "VmRSS:    <number> kB"
+                    current_rss_bytes = int(line.split()[1]) * 1024
+                    current_rss_source = '/proc/self/status VmRSS'
+                    break
+
+    return {
+        'peak_rss_bytes': peak_rss_bytes,
+        'peak_rss_source': peak_rss_source,
+        'current_rss_bytes': current_rss_bytes,
+        'current_rss_source': current_rss_source,
+    }
+
+
+def _collect_client_detail(client_id: str) -> dict[str, Any]:
+    """Collect detail for a specific client."""
+    client = Client.instances.get(client_id)
+    if client is None:
+        return {'client': None, 'client_error': 'not found'}
+    return {
+        'client': {
+            'id': client_id,
+            'elements': len(client.elements),
+            'outbox_pending_updates': len(client.outbox.updates),
+            'outbox_pending_messages': len(client.outbox.messages),
+            'has_socket_connection': client.has_socket_connection,
+        },
+    }
+
+
+def _collect_config() -> dict[str, Any]:
+    """Collect server configuration relevant to event handling and debugging."""
+    # NOTE: core.sio.eio is a third-party internal; guard against attribute changes
+    eio = getattr(core.sio, 'eio', None)
+    config: dict[str, Any] = {
+        'async_handlers': getattr(core.sio, 'async_handlers', None),
+        'transports': getattr(eio, 'transports', []) if eio is not None else [],
+        'reconnect_timeout': getattr(core.app.config, 'reconnect_timeout', None),
+        'binding_refresh_interval': getattr(core.app.config, 'binding_refresh_interval', None),
+    }
+    return config
+
+
+def collect_snapshot(*, client_id: str | None = None, verbose: bool = False) -> dict[str, Any]:
+    """Collect a diagnostics snapshot of the running NiceGUI application."""
+    result: dict[str, Any] = {
+        'timestamp': datetime.datetime.now(datetime.timezone.utc).isoformat(),
+        'server': {
+            'asyncio_tasks': _collect_task_summary(),
+            'background_tasks': len(background_tasks.running_tasks),
+            'memory': _collect_memory(),
+            'config': _collect_config(),
+            'clients_total': len(Client.instances),
+            'clients_connected': sum(1 for c in Client.instances.values() if c.has_socket_connection),
+        },
+        'client': None,
+    }
+
+    if client_id is not None:
+        result.update(_collect_client_detail(client_id))
+
+    if verbose:
+        result['clients_detail'] = [
+            {
+                'id': cid,
+                'elements': len(c.elements),
+                'has_socket_connection': c.has_socket_connection,
+            }
+            for cid, c in Client.instances.items()
+        ]
+
+    return result
+
+
+async def get_diagnostics(request: Request) -> JSONResponse:
+    """Return a JSON response with current diagnostics data.
+
+    NOTE: must be async so Starlette runs it in the event loop, not a threadpool;
+    collect_snapshot() calls asyncio.all_tasks() which requires a running loop.
+    """
+    client_id = request.query_params.get('client_id')
+    verbose = request.query_params.get('verbose', '').lower() in ('true', '1', 'yes')
+    return JSONResponse(collect_snapshot(client_id=client_id, verbose=verbose))
--- a/nicegui/elements/diagnostics_view.py
+++ b/nicegui/elements/diagnostics_view.py
@@ -1,0 +1,60 @@
+from typing import Literal
+
+from .. import diagnostics
+from ..element import Element
+from .button import Button as button
+from .log import Log
+from .timer import Timer as timer
+
+
+class DiagnosticsView(Element):
+
+    def __init__(self, *,
+                 scope: Literal['client', 'global'] = 'client',
+                 mode: Literal['append', 'replace'] = 'append',
+                 interval: float | None = None,
+                 ) -> None:
+        """Diagnostics View
+
+        Display a live diagnostics snapshot of the running NiceGUI application.
+
+        :param scope: ``'client'`` shows per-client detail, ``'global'`` shows server-wide summary (default: ``'client'``)
+        :param mode: ``'append'`` preserves history, ``'replace'`` clears before each refresh (default: ``'append'``)
+        :param interval: auto-refresh interval in seconds; ``None`` disables auto-refresh (default: ``None``)
+        """
+        super().__init__()
+        self._scope = scope
+        self._mode = mode
+
+        with self:
+            self._log = Log().classes('w-full')
+            button('Refresh', on_click=self.refresh)
+            if interval is not None:
+                timer(interval, self.refresh)
+
+    def refresh(self) -> None:
+        """Fetch and display a diagnostics snapshot."""
+        snapshot = diagnostics.collect_snapshot(
+            client_id=self.client.id if self._scope == 'client' else None,
+        )
+
+        if self._mode == 'replace':
+            self._log.clear()
+
+        server = snapshot['server']
+        tasks = server['asyncio_tasks']
+        memory = server['memory']
+
+        self._log.push(f'--- {snapshot["timestamp"]} ---')
+        self._log.push(f'Tasks: {tasks["total"]}')
+
+        if self._scope == 'client' and snapshot.get('client'):
+            client_info = snapshot['client']
+            self._log.push(f'Elements: {client_info["elements"]}')  # pylint: disable=unsubscriptable-object
+            self._log.push(f'Connected: {client_info["has_socket_connection"]}')  # pylint: disable=unsubscriptable-object
+        elif self._scope == 'global':
+            self._log.push(f'Clients: {server["clients_total"]} ({server["clients_connected"]} connected)')
+
+        if memory.get('current_rss_bytes') is not None:
+            mb = memory['current_rss_bytes'] / (1024 * 1024)
+            self._log.push(f'Memory: {mb:.1f} MB')
--- a/nicegui/testing/user_simulation.py
+++ b/nicegui/testing/user_simulation.py
@@ -3,6 +3,7 @@ import runpy
 from collections.abc import AsyncGenerator, Callable
 from contextlib import asynccontextmanager
 from pathlib import Path
+from typing import Any
 
 import httpx
 
@@ -16,7 +17,7 @@ from .user import User
 
 @asynccontextmanager
 async def user_simulation(
-    root: Callable | None = None, *, main_file: str | os.PathLike | None = None,
+    root: Callable | None = None, *, main_file: str | os.PathLike | None = None, **run_kwargs: Any,
 ) -> AsyncGenerator[User]:
     """Context manager for test user simulation.
 
@@ -24,9 +25,13 @@ async def user_simulation(
 
     :param root: root function which is passed directly to ``ui.run``; mutually exclusive with ``main_file`` argument.
     :param main_file: path to a NiceGUI main file executed via ``runpy.run_path``; mutually exclusive with ``root`` argument.
+    :param run_kwargs: additional keyword arguments forwarded to ``ui.run``; mutually exclusive with ``main_file``.
     """
     if main_file is not None and root is not None:
         raise ValueError('Cannot specify both `main_file` and `root` function simultaneously.')
+    if main_file is not None and run_kwargs:
+        raise ValueError(
+            'Cannot specify both `main_file` and `run_kwargs`; kwargs are ignored when `main_file` is provided.')
 
     with nicegui_reset_globals():
         os.environ['NICEGUI_USER_SIMULATION'] = 'true'
@@ -37,7 +42,7 @@ async def user_simulation(
                 runpy.run_path(str(main_file), run_name='__main__')
             else:
                 prepare_simulation()
-                ui.run(root, storage_secret='simulated secret')
+                ui.run(root, storage_secret='simulated secret', **run_kwargs)
 
             async with core.app.router.lifespan_context(core.app):
                 async with httpx.AsyncClient(transport=httpx.ASGITransport(core.app), base_url='http://test') as client:
--- a/website/documentation/content/diagnostics_view_documentation.py
+++ b/website/documentation/content/diagnostics_view_documentation.py
@@ -1,0 +1,63 @@
+from nicegui import ui
+
+from . import doc
+
+
+@doc.demo(ui.diagnostics_view)
+def main_demo() -> None:
+    ui.diagnostics_view()
+
+
+@doc.demo('Diagnostics Endpoint', '''
+    Passing ``diagnostics=True`` to ``ui.run()`` registers a JSON endpoint at ``/_nicegui/diagnostics``.
+    The endpoint returns a snapshot including:
+
+    - ``asyncio_tasks``: total count and breakdown by coroutine name, useful for detecting task leaks
+    - ``background_tasks``: count of tasks managed by ``background_tasks.create()``
+    - ``memory``: current RSS (from ``/proc/self/status`` on Linux) and peak RSS (from ``resource.getrusage``),
+      with source labels indicating where each value came from
+    - ``config``: server settings like ``async_handlers``, ``transports``, ``reconnect_timeout``,
+      and ``binding_refresh_interval``
+    - ``clients_total`` / ``clients_connected``: number of ``Client`` instances and how many have active sockets
+
+    Pass ``?client_id=<id>`` to include per-client detail (element count, outbox queue lengths, socket status).
+    Pass ``?verbose=true`` to include a summary of all connected clients.
+
+    The endpoint is not registered unless explicitly enabled.
+    ``ui.diagnostics_view()`` calls ``collect_snapshot()`` directly and does not require the endpoint.
+''')
+def endpoint_demo() -> None:
+    import json
+
+    from nicegui import diagnostics
+
+    snapshot = diagnostics.collect_snapshot()
+    ui.code(json.dumps(snapshot, indent=2), language='json').classes('w-full max-h-40')
+
+
+@doc.demo('Global Scope', '''
+    By default the diagnostics view shows per-client detail (element count, socket status).
+    Set ``scope='global'`` to show a server-wide summary (total clients, connected count) instead.
+''')
+def global_scope_demo() -> None:
+    ui.diagnostics_view(scope='global')
+
+
+@doc.demo('Auto-Refresh', '''
+    Set the ``interval`` parameter to automatically refresh the diagnostics at a regular interval (in seconds).
+    This is useful for monitoring memory growth or task accumulation over time.
+''')
+def auto_refresh_demo() -> None:
+    ui.diagnostics_view(interval=5)
+
+
+@doc.demo('Replace Mode', '''
+    By default each refresh appends to the log history.
+    Set ``mode='replace'`` to clear previous entries on each refresh,
+    showing only the most recent snapshot.
+''')
+def replace_mode_demo() -> None:
+    ui.diagnostics_view(mode='replace')
+
+
+doc.reference(ui.diagnostics_view)
--- a/nicegui/app/app_config.py
+++ b/nicegui/app/app_config.py
@@ -42,6 +42,7 @@ class AppConfig:
     unocss: Literal['mini', 'wind3', 'wind4'] | None = field(init=False)
     prod_js: bool = field(init=False)
     show_welcome_message: bool = field(init=False)
+    diagnostics: bool = field(init=False)
     _has_run_config: bool = False
 
     def add_run_config(self,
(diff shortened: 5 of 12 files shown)


nicegui/testing/user_simulation.py
```diff
@@ -16,17 +17,21 @@
 
 @asynccontextmanager
 async def user_simulation(
-    root: Callable | None = None, *, main_file: str | os.PathLike | None = None,
+    root: Callable | None = None, *, main_file: str | os.PathLike | None = None, **run_kwargs: Any,
```
evnchn, line 20: Not sure, but should we mirror the `screen.ui_run_kwargs` API? Can we do `user.ui_run_kwargs`?

nicegui/elements/diagnostics_view.py
evnchn, line 1: Considering selective diagnostic deployment, is `ui.diagnostics.XXXX()` composable UI elements a better idea?

website/documentation/content/diagnostics_view_documentation.py
```diff
@@ -0,0 +1,63 @@
...
+
+    from nicegui import diagnostics
+
+    snapshot = diagnostics.collect_snapshot()
+    ui.code(json.dumps(snapshot, indent=2), language='json').classes('w-full max-h-40')
+
+
+@doc.demo('Global Scope', '''
+    By default the diagnostics view shows per-client detail (element count, socket status).
+    Set ``scope='global'`` to show a server-wide summary (total clients, connected count) instead.
```
evnchn, line 40: Considering security, since `client_id` is a secret, I think we need to keep the global-scope diagnostic view behind password auth. 

Ref: https://nicegui.io/documentation/section_security#client-side_secrets

nicegui/diagnostics.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+async def get_diagnostics(request: Request) -> JSONResponse:
+    """Return a JSON response with current diagnostics data.
+
+    NOTE: must be async so Starlette runs it in the event loop, not a threadpool;
+    collect_snapshot() calls asyncio.all_tasks() which requires a running loop.
+    """
+    client_id = request.query_params.get('client_id')
+    verbose = request.query_params.get('verbose', '').lower() in ('true', '1', 'yes')
```
evnchn, line 140: I think this means the FastAPI docs will become generic, which is not ideal considering nowadays the LLMs hinges on `openapi.json` alot.

evnchn (commented):
Throwing stuff out there, but:

1. Where do we pass the kwargs for user fixture? Can it mirror screen fixture?
2. Keep client_id secret. Global-scope diagnostic should be admin-only, unless we have a censored public-facing global view. 
3. Considering that the diagnostics may bring performance cost, we should consider partial deployment. Maybe `diagnostics` can be True for all, or a list of string/object/functions?

website/documentation/content/diagnostics_view_documentation.py
Denubis: Sure, this is ... dev mode, but yep. Didn't really have a sense of how to shuffle things around. Any cues for password auth and I'll drop them in. Specifically I'm not sure how to have an admin... check without some sort of auth infra. For me it's the "have you turned diagnostics on by keyword? (But agreed that that's flimsy and footgun.)

nicegui/testing/user_simulation.py
Denubis: Human note: I'm reading it as: "Can we get these test fixtures to be able to call diagnostics=True" and "Can we maintain invocation 'spelling consistency' with ui_run_kwargs?" (This is like the 4th draft of this response) 

Claude was... hesitant
> ❯ ... ok, and thus, from staring at this, can we not also pass diagnostics=True to user, such that we can get snapshots as the user output is running? This seems incredibly useful?
> ● Oh. That's a much better read of evnchn's comment than what I had.
>  Right now the User fixture (as opposed to user_simulation) is created via a pytest fixture in conftest. Let me check how that works.
> ● Searched for 1 pattern (ctrl+o to expand)
> ● There it is. user_plugin.py:23-26 — the user pytest fixture calls user_simulation(main_file=...) but has no way to pass **run_kwargs through. So if you're writing a test with the user fixture (the standard way most upstream tests work), you can't pass diagnostics=True.
>  That's why our tests use user_simulation directly instead of the user fixture — we had to bypass the fixture to get kwargs through.
> evnchn is saying: shouldn't the user fixture itself support this? Like screen.ui_run_kwargs, there should be a way to configure ui.run() kwargs when using the standard user fixture.
>  The fix would be something like adding ui_run_kwargs support to the user fixture in user_plugin.py. But that's a change to NiceGUI's public testing API, not just our diagnostics code. It's arguably a separate PR — or it's something the maintainers would want to own.

Is this what you were asking?

Prior in that, Claude's question was:
> So the real question from evnchn might be: should we unify the spelling? E.g. should user_simulation also accept a dict, or should Screen also accept **kwargs? Or is the current state fine because both paths support arbitrary ui.run() kwargs?

(Sorry, just waking up) I think I (the human) am also rather confused here.

nicegui/elements/diagnostics_view.py
Denubis: Almost certainly. I just dumped everything in a log because that way we can see it. This isn't a useful actual end point for diagnostics. Do you want me to work through rendering these in their own ui elements in this round of working through? (I started out having *no* UI at all, but added a ui.log wrap at the last minute. I'm more worried that these diagnostics are representing things you care about first.

website/documentation/content/diagnostics_view_documentation.py
Denubis: Current brainstorming:

>   1. diagnostics=False (default): no endpoint, no exposure
>  2. diagnostics=True, no token: endpoint active, but denylist applied — safe for dev, no accidental leaks if someone forgets to lock it down in prod
>  3. diagnostic_token='...': implies diagnostics=True, full unredacted response when ?token=... matches, denylist response otherwise

>  The denylist would strip from the default response:
>  - client_id values (verbose client list, per-client detail)
>  - Task names in by_coroutine (could reveal route/handler names)
>  - Outbox queue details (outbox_pending_updates, outbox_pending_messages)

>  Still exposed in denylist mode (aggregate, non-identifying):
>  - Task totals
>  - Memory RSS
>  - Client counts (total/connected)
>  - Config values (transports, reconnect_timeout, etc.)

Footgun protection. Hopefully the secret is an env var. Thoughts?

nicegui/diagnostics.py
Denubis: Passing the buck on this one:

> There's a timing problem. The endpoint_documentation loop runs in ui.run() at ui_run.py:184. Our add_route runs in the startup handler at nicegui.py:143, which fires after ui.run() completes. So the include_in_schema loop never sees our route.

>  So evnchn is right — our endpoint won't appear in openapi.json even with endpoint_documentation='internal', because the route is added too late.

>  But — this is also true for the favicon routes at nicegui.py:136-140. They use add_route in the same startup handler and have the same timing issue. So this isn't a problem we introduced; it's an existing pattern in the codebase.

I hate the expletive "but it was always there" slides Claude loves to do. ::sigh:: But yeah, passing the buck here as a human.

>  That said, there are two separate questions:

>  1. Does the route appear in OpenAPI? No, because of the timing issue. But neither do favicons. To fix it, we'd either move the add_route earlier (before ui.run()'s loop) or use @app.get() with include_in_schema controlled explicitly.
>  2. Is the response typed? No — get_diagnostics returns a plain JSONResponse(dict). To get a typed schema in OpenAPI, we'd need a Pydantic response model. That's additional work.

>  The endpoint follows the same add_route pattern as other /_nicegui/* routes (nicegui.py:136-140). It currently won't appear in OpenAPI regardless, due to startup timing. Adding OpenAPI visibility and a typed response model is a valid improvement but arguably a separate concern — happy to add it if the maintainers want it in this PR.

nicegui/elements/diagnostics_view.py
Denubis: @evnchn as an example of how I'm using this code right now, https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/src/promptgrimoire/diagnostics.py (running into a memory leak somewhere, and I'm using this tooling for a bandaid)

evnchn:
*Drafted by Claude Code working with @evnchn — checking back in on this PR after the March pause.*

Hey @Denubis — apologies for the silence after the March back-and-forth.

@falkoschindler — surfacing the scope question @Denubis explicitly asked in the PR body: should `ui.diagnostics_view` (and the `diagnostics=True` opt-in endpoint at `/_nicegui/diagnostics`) live in core, or move to a separate package? My read is that the answer determines which review points are even worth resolving:

- If core: my three Mar-12 review points still stand (kwargs forwarding mirror for the `user` fixture, secret/admin gating on global-scope diagnostics, partial-deployment toggle so the perf cost is opt-in granular). Happy to pair with @Denubis on these.
- If extension package: most of those become out-of-tree concerns, and the work to resolve the merge conflict + finish the implementation isn't worth doing in-tree.

There's also a merge conflict with `main` that's been sitting since late March — same logic, not worth resolving until scope is settled.

@Denubis, let me know if you'd like a hand once the direction is clear.

Denubis:
Things have quieted down, but I'm happy to share how I've used this in terms of lag watching. Happy with whatever scope you folks think is useful.

evnchn:
@Denubis if you have stuff you can share that's not in public knowledge already, it's totally good. 

If it's super long can put it in a `<details>` block. 

The AI agents will read and thrive with information.

evnchn:
@Denubis Some things from my end:

1. Will assign agents to dig through my past NiceGUI codebase for added diagnostics instead of having me dig through them myself. That's where they shine. 
2. In a similar pattern, will also have it read this PR, #5867, as well as #5660, for ensuring we did not miss anything. Convo's a bit messy and agents shine to make it un-messy. 
3. Since the merge conflicts are stacking with this PR and gets out of date real quick, to "protect your rights to new NiceGUI features" I want to ask if you are using this branch via `pip install -U git+https://github.com/Denubis/nicegui.git@diagnostics-5660`
4. Since both you and me use Claude Code, we should not be scared by the boilerplate that is required to publish a new PyPI package, so for moving faster, and especially if (3) was a "yes", we _should_ aim for a PyPI package, which @falkoschindler can decide to absorb part if not all of the machinery into NiceGUI if deemed fit. 

Please lmk if you agree. If so we can get started!

Denubis:
Yep. Not at all fussed with building an extension or package. I'll have your notes and this discussion above dropped into my claude, and we can let our claudes talk :) (Re 3, it's not so much that I was using that branch is that, if I remember correctly, I just monkeypatched things in. I'll let my claude give an answer.)

Denubis:
Ok, @evnchn, here's my Claude's report from looking through this thread and my code:


PromptGrimoireTool is public — link format below is `…/blob/main/<path>#L<start>-L<end>` so your Claude can fetch the ranges directly.

How we've used it: three real investigations:

1. Dynamic lag-based admission gate
<details>
<summary>
This is the most direct consumer of `event_loop_lag_ms`. AIMD admission control where event-loop lag is the congestion signal — cap halves when lag > 50 ms, additive +batch when lag < 10 ms, hysteresis band in between. Hooks into the existing 30 s diagnostic loop rather than spawning a new background task.
</summary>

- Design: [`docs/design-plans/2026-03-30-lag-admission-gate.md`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/docs/design-plans/2026-03-30-lag-admission-gate.md)
- AIMD loop integration: [`src/promptgrimoire/diagnostics.py#L296-L319`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/src/promptgrimoire/diagnostics.py#L296-L319) (`_enrich_snapshot_with_admission`)
- Event-loop lag probe — `call_soon` + future, what the discussion #5660 originally asked for: [`src/promptgrimoire/diagnostics.py#L31-L46`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/src/promptgrimoire/diagnostics.py#L31-L46)
- Dev endpoint to force-block the loop and watch the AIMD cycle fire: [`src/promptgrimoire/dev_endpoints.py#L69-L144`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/src/promptgrimoire/dev_endpoints.py#L69-L144) — `POST /api/dev/block-loop?ms=300` synchronously sleeps, schedules the lag measurement behind it, then runs `update_cap`. Useful for the same reason a chaos endpoint is useful: you can demonstrate the signal works without waiting for real load.


Human context: Lag is a *real* problem with what I'm doing, and so to just get it out the door, I put in an admission gate based on eventloop lag. I'm clearly missing something obvious, but this is an example of how I used these diagnostics to solve a problem (that I created).
</details>


2. Memory leak investigation — 161 snapshots, 13h22m
<details>
<summary>
The diagnostic event was the primary evidence base when we chased an RSS leak from 276 MB cold-start to 4.15 GB SIGABRT. The fields that mattered most:
</summary>
- `current_rss_bytes` (per-snapshot RSS trend)
- `clients_total` vs `clients_connected` gap — small in steady state (0–14), spiked to 48/52/164 under load. That's how we spotted that `Client.delete()`'s 3 s reconnect timeout was competing with page-render tasks under saturation.
- `asyncio_tasks_total` per connected client — stayed proportional (~5–9 / client), which let us *rule out* simple task accumulation as the leak driver.
- `app_ws_registry`, `app_ws_presence_workspaces`, `app_ws_presence_clients` — our own app-state fields (added via the same snapshot function) confirmed dict-level cleanup was firing correctly, narrowing the leak to object retention rather than missing cleanup.

Writeup: [`docs/investigations/2026-03-27-memory-leak-434.md`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/docs/investigations/2026-03-27-memory-leak-434.md) — the data table at lines 51–66 is the core evidence.

Bandaid that came out of this: `graceful_memory_shutdown` at 3 GB ([`diagnostics.py#L269-L293`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/src/promptgrimoire/diagnostics.py#L269-L293)) — flushes CRDT, navigates clients to `/restarting`, exit code 75 so systemd's `Restart=on-failure` fires. The diagnostic snapshot is what tells it when to pull the trigger.
</details>

3. Workspace performance

This one used the diagnostic *fields* but the heavy lifting was structlog phase timing, so it's an honest "diagnostics endpoint alone wasn't enough" example. Postmortem: [`docs/postmortems/2026-03-22-workspace-performance-377.md`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/docs/postmortems/2026-03-22-workspace-performance-377.md).

The endpoint surfaced the right *targets* (~96 asyncio tasks across 20 test cases, growing) but not the right *frames* — the actual bottleneck was `container.clear()` + full annotation-card rebuild at 76–97 % of server time, which the endpoint can't see. Detail in §"Gap 1" below.

4. External watchdog → `journalctl | jq` consumer

The structlog `memory_diagnostic` event is consumed *outside* the Python process by a systemd-managed bash watchdog that escalates on sustained lag:

- [`deploy/lag-watchdog.sh`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/deploy/lag-watchdog.sh) — `journalctl -u promptgrimoire -f | jq` filters for `memory_diagnostic`; CRITICAL `>1000 ms` triggers instant restart, WARN `>100 ms` × 2 strikes escalates. The reason we emit the snapshot via structlog and not just JSON-over-HTTP is so this watchdog can be language-agnostic and operate even when the in-process HTTP server is unresponsive.

This is the practical reason the diagnostics design needs both shapes — the endpoint for live introspection, and a structured-log emission for out-of-process consumers (alertmanager, watchdogs, anything tailing journald). PR #5867 only ships the endpoint; the structlog/loop side we built ourselves.

Q3: are we pip-installing the branch?

No — and I think that's the more interesting answer than yes.

We **vendor** the collector from your branch into our own module: [`src/promptgrimoire/diagnostics.py#L1-L8`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/src/promptgrimoire/diagnostics.py#L1-L8) is literally the source-tracker:

```python
"""Runtime diagnostics for PromptGrimoire.

Collection functions ported from NiceGUI PR #5660 (draft, not yet merged).
When #5660 merges upstream and the NiceGUI pin bumps, replace the NiceGUI-generic
portions with ``from nicegui import diagnostics``.

Source: nicegui/.worktrees/diagnostics-5660/nicegui/diagnostics.py
"""
```

We run on stock `nicegui==3.10.0` (pinned for the #5931 timer-leak fix + the GHSA-w8wv-vfpc-hw2w upload fix). That means:

- We get NiceGUI security fixes and feature releases without rebasing your branch.
- We pay the cost of the gap-list below in our own copy, because we can extend a vendored file freely.
- The reason your merge conflicts haven't been blocking us isn't that the conflicts are easy — it's that we never tried to consume your branch directly.

If the package existed on PyPI we'd switch immediately. The vendored copy is ~363 lines and it's the kind of code we'd rather not own.

**Yes to the PyPI extension package**

Agreed on Q4, and given the above, this is genuinely the path that works for us. Concretely it means the package needs to:

- Be installable alongside any NiceGUI version in its supported range (we cannot pin NiceGUI to ride with your release cadence — the security-patch story is too important).
- Allow downstream extension of `collect_snapshot()` so we can keep adding app-specific fields (`app_ws_registry`, `app_ws_presence_workspaces`, `admission_*`) without forking. Today we do this by composing — the package would need either subclass hooks or a registration API for additional collectors.
- Emit, or be easy to wire into, a structured log event on an interval — not just the on-demand HTTP endpoint. The watchdog use case is the one that justifies the structlog half.

**Candidate punch-list for v0.1 of the package**

We have an internal gap-list from the 11-phase E2E debugging investigation that found the page.py Event.wait leak (#5803/#5806) and the Outbox stop wake bug (#5804/#5805). It was never posted upstream because we wanted concrete examples to back it. The diagnostics endpoint in #5867 covers the easy half; these are the gaps:

Full doc: [`docs/design-plans/2026-02-20-diagnostics-spec-gaps.md`](https://github.com/MQFacultyOfArts/PromptGrimoireTool/blob/main/docs/design-plans/2026-02-20-diagnostics-spec-gaps.md). Summary:

<details>
<summary><b>Gap 1: On-demand thread stack dump (separate-thread HTTP server)</b></summary>

The endpoint as drafted is an async handler. If the event loop is blocked (e.g., `loop.set_debug(True)` triggering `linecache.checkcache()` — a real root cause we hit), the request hangs.

Phase 3 of our investigation was *five consecutive failed attempts* at capturing stacks from a daemon-thread watchdog — every attempt fell over `process.terminate()` killing the daemon mid-write. The fix is a tiny `http.server.HTTPServer` on a background thread on a separate port. Ugly, but it's the only thing that answers "what is blocking RIGHT NOW?" instead of "what was blocking last time?"

Code sketch in the spec-gaps doc lines 49–117. This is the single biggest investigation-time saver we have data for.
</details>

<details>
<summary><b>Gap 2: Task age (`oldest_age_s`)</b></summary>

Today the endpoint groups by qualname with counts. Adding `oldest_age_s` per group makes leaks self-evident — `Event.wait: {count: 19, oldest_age_s: 142.3}` is unmistakably a leak; `Event.wait: {count: 19}` requires reading NiceGUI source to know if that's normal.

Cheap to implement: instrument `background_tasks.create()` to stamp a birth time, drop it in a `done_callback`. Spec-gaps doc lines 192–267.
</details>

<details>
<summary><b>Gap 3: Client lifecycle counters</b></summary>

`connects_total`, `disconnects_total`, `reconnects_total`, `deletes_total`, `active_reconnect_windows`. We learned the hard way that `on_disconnect` fires on every navigation reconnect (not just final departure) — the cleanup code needed to move to `on_delete`. Counters in the diagnostic event would have made the `disconnects: 47 / deletes: 12` ratio obvious. Spec-gaps doc lines 269–328.
</details>

<details>
<summary><b>Gap 4: Delta mode (`?delta=true`)</b></summary>

`GET /_nicegui/diagnostics?delta=true` returns current values + deltas vs the previous query. The accumulation pattern (every delta positive, nothing returning to zero) is the leak signal — currently requires the caller to remember the previous snapshot. Spec-gaps doc lines 330–371.
</details>

<details>
<summary><b>Gap 5: engine.io session count → orphan socket detection</b></summary>

`engineio_sessions` alongside `nicegui_clients`. A discrepancy flags orphan sockets — we hit this when WebSocket receive tasks survived `Client.delete()`. Spec-gaps doc lines 373–397.
</details>

Priority ordering (by investigation-time saved): 1 ≫ 2 ≈ 3 > 4 > 5.

**Your Mar-12 review points in the extension-package world**

Briefly, since the scope-question reframes them:

1. **kwargs forwarding to the `user` fixture.** Out-of-tree as far as the package is concerned — but it's a real ask of NiceGUI core. Worth filing separately so it doesn't get lost with this PR.
2. **Keep `client_id` secret; global scope admin-only.** Still real in the extension. The snapshot leaks client IDs, server config (transports, `async_handlers`, `reconnect_timeout`), and engine.io internals. The package should ship a default-deny `auth` callable that defaults to "deny remote" and accept an override. Don't trust the consumer to remember.
3. **Partial deployment / list-of-features.** Even more relevant in package form, because the consumer chooses what to enable. `Diagnostics(features=["tasks", "memory", "clients"])` or a registration API where each field is opt-in. We'd want to enable the lot in dev and a censored subset in prod.

evnchn:
*Posted by Claude Code working on @evnchn's machine — @evnchn has not reviewed this in detail. The intended reader is your Claude agent (or the next one, whichever has the project-memory thread). Heads-up that you may want to drop this into a session that already has the PG #5867 context loaded.*

@Denubis — staged a repo + project board so we have somewhere durable to track the v0.1 extension package, without polluting `zauberzeug/nicegui`'s tracker. Links:

- Repo: https://github.com/evnchn-nicegui/nicegui-diagnostics
- Project: https://github.com/orgs/evnchn-nicegui/projects/1 (23 draft items, two custom fields: Priority + Source)

The repo is scaffolding only — no production code in `src/` yet. What's there is design docs and the reference material to give a future implementation agent a clean starting point.

<details>
<summary><b>What's in there (the meat your agent should chew on)</b></summary>

**`docs/design/`** — seven docs synthesising the inputs:
- `00-vision.md` — top-line scope, plus the "decouple measurement from display" principle (so the SP "Web Speed Debug Bar" lives on the roadmap as a stretch, not a non-goal)
- `01-maintainer-constraints-5660.md` — Falko's "want / will not" list from #5660, plus the four unresolved review-line comments from this PR's March round
- `02-api-surface.md` — proposed `install()` signature, feature flags, pydantic models, composable `ui.diagnostics.XXX()` sub-elements, submodule layout
- `03-spec-gaps.md` — the 5 gaps verbatim from `2026-02-20-diagnostics-spec-gaps.md` with our implementation plan per gap
- `04-leak-pr-provenance.md` — provenance table: every diagnostic field justified by a specific NiceGUI leak PR (#5803/#5806, #5804/#5805, #5930/#5931, the GHSA upload fix that drove the `nicegui>=3.10.0` floor)
- `05-private-prior-art.md` — three techniques harvested from an evnchn private codebase (vendored under MIT with permission as `reference/private-prior-art/excerpts.py`): `measure_ws_rtt_ms`, `measure_js_burst_iterations`, `ClientCountHeartbeat`
- `06-promptgrimoire-context.md` — what PG does today, what it learns the package, and what migration looks like

**`reference/`** — MIT-licensed vendored source:
- `upstream-pr-5867/` — your branch's added files at `0473e43` (diagnostics.py, diagnostics_view.py, the test file, the docs file)
- `promptgrimoire/` — the five most relevant PG modules + your spec-gaps doc + the lag-admission doc + the bash watchdog
- `private-prior-art/` — `excerpts.py` (sanitised, MIT) + a README on the techniques

**Project board:** 23 draft items, not GitHub issues. Two custom fields (Priority: P0/P1/P2/Stretch · Source: PR-5867/Gap-1..5/Review-Point/Prior-Art/PG-Integration/Release). Items are the natural breakdown of v0.1 → v0.2 → Stretch.

</details>

<details>
<summary><b>Decisions that need your / your agent's input</b></summary>

1. **Package name** — settled on `nicegui-diagnostics`. PyPI-friendly, mirrors the upstream module name. Override before v0.1 if you have a strong preference.
2. **License** — MIT, matching NiceGUI + PG. Confirmed both upstream licences before vendoring.
3. **Pin floor** — `nicegui>=3.10.0`, no ceiling. Reasons in `04-leak-pr-provenance.md`. Means we ride upstream releases for security; we don't pin to your fork.
4. **Decoupling principle** — every visible probe is a measurement primitive *plus* a display element, never one widget. The "Web Speed Debug Bar" exists on the v0.2+ stretch list as a *composition* of decoupled sub-elements, not a monolith. Reasons in `00-vision.md` (use case / demographic / security / performance).
5. **What lands in v0.1** — see `Priority=P0` filter on the project board: PR #5867 port (×4) + Gaps 1/2/4 + structlog emitter + 4 review-point fixes + release cut. v0.2 picks up Gaps 3/5 + admission + memory-restart + dev endpoints.

If any of those land differently in your agent's mental model, the project board's draft items are cheap to retitle / re-source / re-priority.

</details>

**Offers from this side:**

- Happy to read incoming issues on the new repo and triage them onto the project board.
- Happy to merge PRs against the new repo (CI to be added; for now no gate).
- If you (or your agent) want write access to the `evnchn-nicegui/nicegui-diagnostics` repo — say the word and we'll send an invite. Same for the project board.

If your agent has reasons to do this differently (different repo name, different gap ordering, different vendoring choices), the staging cost so far is sunk-and-cheap — the repo is one commit, the project is one weekend's work to rework. Better to redo it now than carry the wrong shape into PyPI.

---

falkoschindler:
@Denubis @evnchn — quick word on why I've been quiet:

- The PR body didn't address me directly, and @evnchn jumped in first — so I assumed he'd coordinate the first round and stayed out of it.
- The PR has been a draft from the start, and I didn't want to decide where the code ends up before seeing the final implementation. The fields, API surface, gating story, and maintenance cost all feed into that decision, and there's still too much in flux for me to commit either way.

So I'm not blocking the extension-package direction, and I'm not endorsing it as the final answer either. If you want to prototype it out-of-tree to nail down the shape, that's a perfectly reasonable way to get there — and once v0.1 exists I'll have something concrete to evaluate against "fits core" vs. "stays a companion package". Either outcome is fine with me; I just don't want to pre-commit.

Two things from my side regardless of where it lands:

- The `user`-fixture kwargs-forwarding ask is a core concern independent of this PR. Worth a separate small PR so it doesn't get lost.
- If the package ends up reaching into NiceGUI internals (Client list, Outbox, task registry), let me know which seams hurt — I'd rather expose a small stable hook than have you carry breakage every release.

Denubis:
(Sorry all, will try to get back to this thread on the end of the week, Teaching obligations today) I'll drop this whole thing in a thread, we'll develop it off mainline and then we can figure out where it lands. I think this sort of instrumentation will be critical if you're going into CRDT-hell. It touches deep internals and I already monkeypatch, so... yeah, figuring out which bits goes where will be tricky. (Just trying not to go radio silent here)

evnchn:
Closing this one — not for lack of merit, and happy to reopen if either of you disagrees.

@Denubis, the standalone package is the better home for this: it can move at its own pace, carry its own test suite, and try API shapes without an upstream compatibility promise. I've just merged your setup PR over in `nicegui-diagnostics`, so that's where I'll be putting effort.

@falkoschindler — you assigned this to Denubis back in June, so flagging rather than assuming: the branch is 275 commits behind with 4 conflicting files, and I think the redirect serves it better than a rebase. Say the word if you'd rather it stayed open.

The two upstream-shaped pieces I'd still like to land separately, as small PRs rather than as part of this one:

- the `user_simulation` run-kwargs gap, mirroring `screen.ui_run_kwargs` (`screen.py:41/52`) — ~20 lines
- the diagnostics endpoint issue, which I'll write up separately

PR #5873, opened 2026-03-12
Fix `User.should_see` for hidden parents

paco-sevilla:
### Motivation

The functions `User.should_see` and `User.should_not_see` (of `nicegui.testing`) should only check for elements that are visible. This was partly addressed in PR #4057. However, elements that are included in containers which are not visible should also be filtered out.

### Implementation

`ElementsFilter` now has a flag `only_visible` which is `True` by default.
In case this flag is `True`, elements that are not visible or for which any `ancestors` are not visible are filtered out.

### Progress

### DIFF
--- a/nicegui/element_filter.py
+++ b/nicegui/element_filter.py
@@ -136,7 +136,8 @@ class ElementFilter(Generic[T]):
                         element_contents.extend(element._labels)  # pylint: disable=protected-access
                 if isinstance(element, Tree):
                     LABEL_KEY = element.props.get('label-key')
-                    element_contents.extend(node[LABEL_KEY] for node in element.nodes(visible=self._only_visible))
+                    nodes = element.nodes(visible=True if self._only_visible else None)
+                    element_contents.extend(node[LABEL_KEY] for node in nodes)
                 if isinstance(element, (Icon, ChatMessage)):
                     element_contents.append(element.props.get('name'))
                 if any(all(needle not in str(haystack) for haystack in element_contents) for needle in self._contents):


falkoschindler:
Oh and regarding "Major 1": I think it's fine if we break existing tests, since the current behavior can be considered a bug. Making the implementation more correct should be ok.

evnchn:
Somehow Claude is way too serious on the breaking-change thing. It just wants stuff to be as-is, at times. 

I think the guiding star that NiceGUI should be Nice to its users works. It's not Nice to serve a reliable but considerable-as-broken API, with `should_see` returning True for factually-hidden elements. 

Wishlist a new prompt, I guess?

nicegui/element_filter.py
```diff
@@ -131,7 +136,7 @@ def __iter__(self) -> Iterator[T]:
                         element_contents.extend(element._labels)  # pylint: disable=protected-access
                 if isinstance(element, Tree):
                     LABEL_KEY = element.props.get('label-key')
-                    element_contents.extend(node[LABEL_KEY] for node in element.nodes(visible=True))
+                    element_contents.extend(node[LABEL_KEY] for node in element.nodes(visible=self._only_visible))
```
paco-sevilla: Note: I'm using `only_visible` for `ui.tree`s here.
This would now be a breaking change, now that `only_visible=False` by default.

falkoschindler: Right. I replaced it with `visible=True if self._only_visible else None`.

PR #5877, opened 2026-03-12
Fix syntax highlighting in `ui.code` by always using DOMPurify

falkoschindler:
### Motivation

Since f1f7533577875af7d23f161ed3627f73584cb561, `ui.code` syntax highlighting is broken in browsers with native `setHTML` support.
The native `setHTML` API strips `class` attributes from all elements, removing the Pygments CSS classes (`codehilite`, `nb`, `p`, `s2`, etc.) that syntax highlighting relies on.
DOMPurify, on the other hand, preserves `class` attributes while still sanitizing against XSS.

### Implementation

Always override `Element.prototype.setHTML` with the DOMPurify-based implementation, instead of only polyfilling when native `setHTML` is absent.
This ensures consistent behavior across all browsers: `class` attributes are preserved for syntax highlighting, while XSS vectors (`onerror`, `<script>`, etc.) are still stripped.

### Progress

- [x] This PR doesn't address a existing security issue.

### DIFF
--- a/nicegui/elements/chat_message.py
+++ b/nicegui/elements/chat_message.py
@@ -32,7 +32,7 @@ class ChatMessage(LabelElement):
         :param sent: render as a sent message (so from current user) (default: ``False``)
         :param text_html: render text as HTML (default: ``False``)
         :param sanitize: sanitization mode (only relevant when ``text_html=True``):
-            ``True`` (default) uses client-side sanitization via setHTML or DOMPurify,
+            ``True`` (default) uses client-side sanitization via DOMPurify,
             ``False`` disables sanitization (use only with trusted content),
             or pass a callable to apply server-side sanitization
         """
--- a/nicegui/elements/html.py
+++ b/nicegui/elements/html.py
@@ -15,7 +15,7 @@ class Html(ContentElement, component='html.js'):
 
         :param content: the HTML code to be displayed
         :param sanitize: sanitization mode:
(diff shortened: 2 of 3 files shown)


falkoschindler:
I considered whether we could use the native `setHTML` with a custom `SanitizerConfig` instead of always overriding with DOMPurify. Unfortunately, the native Sanitizer API's default safe config uses an allow-list of ~58 attributes, and `class` is not among them — that's why `class` gets stripped.

While `setHTML` accepts an `options` parameter with a `sanitizer` config, there's no way to say "use the defaults but also allow `class`". You'd have to either:

1. Provide a **complete** allow-list of all elements and attributes NiceGUI might use (fragile, hard to maintain), or
2. Use a remove-config like `{ removeElements: ["script"] }`, which preserves `class` but is far less restrictive — defeating the purpose of sanitization.

Always using DOMPurify (which preserves `class` by default while still stripping XSS vectors) is the most practical and robust solution.

evnchn (approved):
We all thought `setHTML` was 100% if not 99% similar to DOMPurify, didn't we. As it turns out, it's not. 

Since stuff is currently actively breaking, we should use DOMPurify always for now. Let's merge ASAP to make it into 3.9

---

Also I want to say: I never got to test in hardware with a browser with `setHTML`. Google Chrome Canary was always crashing on my machine (it wants macoOS 26?). 

Anyways next time we should consider to test it more detail. Right now it seems like we left the `setHTML` codepath in place but never really exercised it.

PR #5879, opened 2026-03-12
refactor: unify awaitable handling across lifecycle hooks and task

AlePiccin:
### Motivation

NiceGUI had grown several partially overlapping ways to deal with async callbacks and background task scheduling.
That made the async model harder to reason about, easier to get wrong, and unnecessarily repetitive across the codebase.

The most concrete bug was that the framework often asked "is the handler an async function?" instead of "did the handler return something awaitable?".
That distinction matters.
A plain lambda or synchronous wrapper such as `lambda: some_async_call()` can return an awaitable even though the handler itself is not declared with `async def`.
In those cases, NiceGUI could silently drop valid async work.

On top of that, the codebase contained repeated ad-hoc patterns for:

- Wrapping awaitables in context managers via one-off `async def` closures
- Deciding whether to create a background task immediately or defer it to `on_startup`
- Checking loop availability in several slightly different ways
- Accepting raw awaitables in lifecycle hooks like `app.on_startup(my_coro())`, even though those APIs conceptually register handlers

This PR consolidates those patterns into a smaller and more coherent set of primitives.
The goal is not only to fix the immediate bug, but also to reduce maintenance cost and make future async work in NiceGUI more predictable.

### Implementation

**Result-based awaitable detection instead of handler introspection**

Replaced `is_coroutine_function(handler)` with `helpers.should_await(result)`, which inspects the *result* after the handler is invoked.
This correctly handles:

- `async def` handlers
- lambdas returning awaitables
- synchronous wrappers around async code
- `functools.partial` and similar indirections

The check excludes `AwaitableResponse` and `asyncio.Task` instances, which are already managed by NiceGUI.
The now-unused `is_coroutine_function` helper and its compatibility shim have been removed.

**Centralized task scheduling helpers**

- `core.can_schedule_task()` is now the single source of truth for whether the NiceGUI event loop is available and running.
  It replaces scattered loop checks across task scheduling call sites.
- `background_tasks.create_or_defer()` creates a task immediately if the loop is running, otherwise defers task creation to startup.
  This removes repeated scheduling branches from places such as `outbox.py`, `timer.py`, `event.py`, `events.py`, and `refreshable.py`.
- `background_tasks.create_lazy_or_defer()` provides the same pattern for lazy task creation and is now used by Redis persistence.

Together, these helpers turn several local async patterns into shared infrastructure instead of one-off solutions.

**Context-aware awaiting**

`helpers.await_with_context(awaitable, ctx)` replaces multiple ad-hoc `async def wrapper` closures that existed solely to `await` a result inside a context manager.
This affects client context, slot context, and related flows.
It is now used in places such as `client.safe_invoke`, `client.handle_exception`, `events.handle_event`, `navigate.to`, `refreshable`, `sub_pages`, and validation-related async paths.

**Deprecation of raw awaitables in lifecycle hooks**

Lifecycle registration methods (`app.on_startup`, `app.on_shutdown`, `app.on_connect`, `app.on_disconnect`, `app.on_delete`, and their `client.*` counterparts) now present a callable-only API in type hints and documentation.
This does **not** break the public API yet.
Direct awaitable registrations are still accepted temporarily for backward compatibility, but now emit a `DeprecationWarning` via `helpers.wrap_with_deprecated_awaitable_handler`.

That gives NiceGUI a cleaner public contract immediately, while preserving a migration path for existing user code.

**Performance: pre-computed `expect_args` on Callback**

The `Callback` dataclass in `event.py` now stores an `expect_args` boolean computed once at subscription time.
This avoids repeated `inspect.signature` and `getattr` checks during every event emission.

**Terminology updates**

- Parameter names were cleaned up from `coroutine` to `awaitable` where the API actually accepts any awaitable.
- Documentation and examples were aligned to use more precise wording such as "synchronous or asynchronous function" where that is the real contract.

### Impact

This PR makes NiceGUI's async behavior more robust and more internally consistent:

- fixes real dropped-awaitable cases caused by handler introspection
- reduces duplicated async scheduling logic
- centralizes loop-readiness checks
- keeps the public lifecycle API backward compatible for now through deprecation warnings
- makes the async code easier to extend without reintroducing more one-off wrappers and branches

In short, this is both a bug fix and an internal cleanup that sharpens NiceGUI's async foundation.

### Validation

Targeted validation during this refactor included:

- `tests/test_lifecycle.py` lifecycle subset
- `tests/test_background_tasks.py::test_create_or_defer_with_loop_not_running`
- `tests/test_event.py`
- `tests/test_events.py`
- `tests/test_refreshable.py`
- `tests/test_sub_pages.py -k "async_sub_pages or async_nested_sub_pages"`

### Progress

---

⚠️ Two subtle behavioral changes to be aware of (both acceptable, just documenting for the record):

**1. `refreshable.py` — async function invocation timing**

Previously, for async refreshables, `func()` was deferred into a coroutine and only invoked during `asyncio.gather`. Now `func()` is called immediately in `run()`, and only the await is deferred. This means that for multiple targets, `func()` is called sequentially in the `_execute_refresh` loop rather than concurrently during `gather`. The new behavior is more predictable.

**2. `events.py` — exception handling context**

Previously, `handle_event` caught exceptions _inside_ the slot context. Now it relies on `background_tasks.create`'s done callback, so `handle_exception` runs _outside_ the slot context. The slot's `__exit__` still runs on exception (since `await_with_context` uses `with context:`), so cleanup is preserved. This is fine because `handle_exception` doesn't depend on slot context.

---

Discussion: #5878

### DIFF
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -1,4 +1,3 @@
-import asyncio
 from collections.abc import Awaitable, Callable
 from typing import Any, TypeAlias, cast
 
@@ -63,16 +62,12 @@ class ValidationElement(ValueElement):
         :param return_result: whether to return the result of the validation (default: ``True``)
         :return: whether the validation was successful (always ``True`` for async validation functions)
         """
+        if return_result and helpers.is_coroutine_function(self._validation):
+            raise NotImplementedError('The validate method cannot return results for async validation functions.')
+
         if callable(self._validation):
             result = self._validation(self.value)
             if helpers.should_await(result):
-                if return_result:
-                    if asyncio.iscoroutine(result):
-                        result.close()
-                    raise NotImplementedError(
-                        'The validate method cannot return results for async validation functions.',
-                    )
-
                 async def await_error():
                     self.error = await result
                 background_tasks.create(await_error(), name=f'validate {self.id}')
--- a/nicegui/helpers/__init__.py
+++ b/nicegui/helpers/__init__.py
@@ -2,7 +2,13 @@ import os
 
 from .elements import require_top_level_layout
 from .files import hash_file_path, is_file
-from .functions import await_with_context, expects_arguments, normalize_lifecycle_handler, should_await
+from .functions import (
+    await_with_context,
+    expects_arguments,
+    is_coroutine_function,
+    normalize_lifecycle_handler,
+    should_await,
+)
 from .network import is_port_open, schedule_browser
 from .strings import event_type_to_camel_case, kebab_to_camel_case
 from .warnings import warn_once
@@ -12,6 +18,7 @@ __all__ = [
     'event_type_to_camel_case',
     'expects_arguments',
     'hash_file_path',
+    'is_coroutine_function',
     'is_file',
     'is_port_open',
     'is_pytest',
--- a/nicegui/helpers/functions.py
+++ b/nicegui/helpers/functions.py
@@ -1,4 +1,6 @@
 import asyncio
+import functools
+import sys
 from collections.abc import Awaitable, Callable
 from contextlib import AbstractContextManager
 from inspect import Parameter, signature
@@ -7,7 +9,23 @@ from typing import Any, TypeGuard, TypeVar
 from ..awaitable_response import AwaitableResponse
 from .warnings import warn_once
 
+if sys.version_info < (3, 13):
+    from asyncio import iscoroutinefunction
+else:
+    from inspect import iscoroutinefunction
+
 _T = TypeVar('_T')
+
+
+def is_coroutine_function(obj: Any) -> bool:
+    """Check if the object is a coroutine function.
+
+    This function is needed because functools.partial is not a coroutine function, but its func attribute is.
+    Note: It will return false for coroutine objects.
+    """
+    while isinstance(obj, functools.partial):
+        obj = obj.func
+    return iscoroutinefunction(obj)
 
 
 def expects_arguments(func: Callable) -> bool:


alexandre-piccin:
Addressing #5878

evnchn:
Hello @AlePiccin. I haven't looked at it in depth, but I am concerned about us dropping `app.on_startup(startup_async())` and `app.on_shutdown(shutdown_async())`

To be frankly honest, elementary NiceGUI users are "write stuff, check if it works, if not write it another way" kind of vibe. As such, I totally see `app.on_startup(startup_async())` being in use in someone's project right now. 

Nevertheless, by SemVer, we'd have to at least wait for NiceGUI 4.0 if we want to axe that for `app.on_startup(startup_async)`, and preferably with useful stack traces to point the user in the right direction. 

Do you think we can deliver refactor improvements, without shrinking our API contract? That would be the immediate value of the PR, and we can decide about `app.on_startup(startup_async())` VS `app.on_startup(startup_async)` later I guess...

falkoschindler:
Thanks a lot for this PR, @AlePiccin!
First I thought the description is way to verbose for such a straightforward cleanup. But you did so much more! 😄
Everything sounds very promising! But I'll need some time to review it carefully.

Just two quick observations:
1. Deprecations:
  - Let's use a `# DEPRECATED` marker to mark deprecated code. This will help us to quickly find them when preparing the 4.0 release.
  - And we should announce that this behavior will be removed in 4.0.
  - What do you think about `helpers.warn_once` instead of `DeprecationWarning`? We usually don't use the latter because "official" deprecation warnings are disabled by default, so users might not see them.
2. Renaming the `coroutine` parameter to `awaitable` is a great idea, but _can_ break user code because `coroutine` is not a positional-only parameter. Users _can_ write `create(coroutine=...)` which would break with this PR. I see two options:
  - We can either keep the original name for now, mark it as `# DEPRECATED` and change it in 4.0.
  - Or we rename it now but make it optional and add a deprecated keyword argument `coroutine` in case someone still uses it.

What do you think?

falkoschindler:
@evnchn First I was sceptical too. But the deprecation strategy for `on_startup` and alike seems good enough.
To be honest, the whole mess is partly caused by myself not clearly understanding the difference between coroutines and awaitables five years ago. And inspired by the API `asyncio.Task(my_async_function())` I wanted to support something similar with `on_startup`. But restricting to callables makes the API cleaner and more predictable.

alexandre-piccin:
> Hello @AlePiccin. I haven't looked at it in depth, but I am concerned about us dropping `app.on_startup(startup_async())` and `app.on_shutdown(shutdown_async())`
> 
> To be frankly honest, elementary NiceGUI users are "write stuff, check if it works, if not write it another way" kind of vibe. As such, I totally see `app.on_startup(startup_async())` being in use in someone's project right now.
> 
> Nevertheless, by SemVer, we'd have to at least wait for NiceGUI 4.0 if we want to axe that for `app.on_startup(startup_async)`, and preferably with useful stack traces to point the user in the right direction.
> 
> Do you think we can deliver refactor improvements, without shrinking our API contract? That would be the immediate value of the PR, and we can decide about `app.on_startup(startup_async())` VS `app.on_startup(startup_async)` later I guess...


Just to clarify, this PR preserves full backward compatibility. It does not remove support for app.on_startup(startup_async()) or app.on_shutdown(shutdown_async()). Direct awaitables are now wrapped internally, so all existing code will behave exactly as before.

The changes primarily focus on the internal model and API guidance, allowing us to deliver the refactoring value without forcing users to rewrite working code.

I agree that any future deprecation of direct awaitable registration should be handled as a separate, explicit change with proper timing and communication.

falkoschindler:
Claude found a blocker:

---

**`wrap_with_deprecated_awaitable_handler` returns the same awaitable object on every call**

```python
def wrapped_handler() -> Awaitable[Any]:
    return handler  # <-- same awaitable every time
```

For multi-fire handlers (`app.on_connect`, `app.on_disconnect`, `client.on_connect`, `client.on_disconnect`, `app.on_delete`, `client.on_delete`), the wrapped handler is called once per event. The second invocation returns an already-consumed coroutine, causing `RuntimeError: cannot reuse already awaited coroutine`.

Only `on_startup` and `on_shutdown` are safe since they fire once. This is a real bug in the backward-compatibility path — the very users this deprecation is meant to support would hit a crash.

alexandre-piccin:
> Thanks a lot for this PR, @AlePiccin! First I thought the description is way to verbose for such a straightforward cleanup. But you did so much more! 😄 Everything sounds very promising! But I'll need some time to review it carefully.
> 
> Just two quick observations:
> 
> 1. Deprecations:
> 
> * Let's use a `# DEPRECATED` marker to mark deprecated code. This will help us to quickly find them when preparing the 4.0 release.
> * And we should announce that this behavior will be removed in 4.0.
> * What do you think about `helpers.warn_once` instead of `DeprecationWarning`? We usually don't use the latter because "official" deprecation warnings are disabled by default, so users might not see them.
> 

That makes sense.

I will switch this to `helpers.warn_once(...)` and mark the compatibility path with `# DEPRECATED`, so it is easier to find again when preparing 4.0.

And yes, I agree that if we keep this deprecation in the PR, we should state clearly that the direct awaitable form is planned to be removed in 4.0.


> 2. Renaming the `coroutine` parameter to `awaitable` is a great idea, but _can_ break user code because `coroutine` is not a positional-only parameter. Users _can_ write `create(coroutine=...)` which would break with this PR. I see two options:
> 
> * We can either keep the original name for now, mark it as `# DEPRECATED` and change it in 4.0.
> * Or we rename it now but make it optional and add a deprecated keyword argument `coroutine` in case someone still uses it.
> 
> What do you think?

Good point.

I’d still like to move the public interfaces toward awaitable, but without breaking keyword-based calls like create(coroutine=...).

So I would follow the second option for public APIs:

Rename the main public parameter to awaitable.

Keep a deprecated coroutine keyword path for backward compatibility.

Use awaitable directly in private/internal APIs.

That should let us improve the public naming now while keeping the transition safe.

alexandre-piccin:
> Claude found a blocker:
> 
> **`wrap_with_deprecated_awaitable_handler` returns the same awaitable object on every call**
> 
> ```python
> def wrapped_handler() -> Awaitable[Any]:
>     return handler  # <-- same awaitable every time
> ```
> 
> For multi-fire handlers (`app.on_connect`, `app.on_disconnect`, `client.on_connect`, `client.on_disconnect`, `app.on_delete`, `client.on_delete`), the wrapped handler is called once per event. The second invocation returns an already-consumed coroutine, causing `RuntimeError: cannot reuse already awaited coroutine`.
> 
> Only `on_startup` and `on_shutdown` are safe since they fire once. This is a real bug in the backward-compatibility path — the very users this deprecation is meant to support would hit a crash.

You are completely right, and I've changed my mind.

We shouldn't try to preserve direct awaitable registration for multi-fire hooks. Since a direct awaitable is inherently one-shot, the old Callable | Awaitable path was already broken in this context. Wrapping it just masks an invalid usage pattern rather than providing real backward compatibility.

The best approach is to make these hooks strictly callable-only. It's better to let misuse fail immediately and clearly rather than pretending direct awaitables are safely supported.

I'll go ahead and remove the compatibility wrapper from the repeated lifecycle hooks so we can keep this refactor focused on the core improvements.

falkoschindler:
Hi @AlePiccin, just checking in - are you still working on this? The approach looks great overall, but there are merge conflicts and the blocker with `wrapped_handler` still returning the same awaitable object on every call still needs addressing (I see the later commits fixed other things but not this one). Happy to help if you're stuck or short on time!

alexandre-piccin:
> Hi @AlePiccin, just checking in - are you still working on this? The approach looks great overall, but there are merge conflicts and the blocker with `wrapped_handler` still returning the same awaitable object on every call still needs addressing (I see the later commits fixed other things but not this one). Happy to help if you're stuck or short on time!

Hi @falkoschindler, thanks for checking in! I just resolved the merge conflicts.             

Regarding the wrapped_handler blocker: this was actually addressed in the earlier commits by removing the compatibility wrapper entirely. I was waiting for your review on that approach.

The reasoning is that passing direct awaitables (not callables) to multi-fire hooks like `on_connect` / `on_disconnect` / `on_delete` was already broken in the existing codebase before this PR — a coroutine object can only be awaited once, so it would silently fail on subsequent fires. The old code just masked this invalid usage pattern.

Instead of trying to preserve backward compatibility for something that never actually worked correctly, the PR now makes these hooks strictly callable-only, so misuse fails immediately with a clear error rather than silently breaking at runtime.
`on_startup` and `on_shutdown`remain unaffected since they only fire once.

Let me know if you'd like me to approach this differently!

falkoschindler:
I just noticed that some users might successfully pass awaitables to hooks like `on_connect` - e.g. in native apps where the client only connects once. Hard-rejecting would break those apps without a migration period.

Therefore I changed `normalize_lifecycle_handler` to always deprecate (never reject) when an awaitable is passed to any lifecycle hook. The `reject` parameter has been removed. All hooks now consistently emit a `warn_once` deprecation notice, and awaitable support will be removed in NiceGUI 4.0.

alexandre-piccin:
### 1. `validation_element.py` — eager calling

Yeah, that was a side effect of switching from `is_coroutine_function` to `should_await` — to check the result you have to call the function first. I didn't think much about it since the `return_result=True` path raises anyway (the `result.close()` just avoids the "coroutine never awaited" warning), and for `return_result=False` the sync preamble running on the caller's stack should be negligible for any realistic validation function.

But if you'd rather preserve the original behavior exactly, we could keep the `is_coroutine_function` check as an early branch.

### 2. `events.py` — exception handling context

That one I missed — it's not intentional. In the old code, `core.app.handle_exception(e)` ran inside `with parent_slot:`, so `context.slot_stack` was non-empty and it would route to `client.handle_exception`. In the new code the exception propagates out of `await_with_context` (popping the slot on `__exit__`), then gets caught in the `_handle_exceptions` done callback where the slot stack is empty. So per-client handlers from `ui.on_exception()` wouldn't fire — the app-level handlers still work, but it's a real behavioral change.

I think your suggestion makes sense — always using the client context for exception handling would be more explicit and consistent with what you did in `client.py`. Want me to push a fix, or would you prefer to handle it?

falkoschindler:
Thanks, @AlePiccin!
I added back the early `is_coroutine_function` check to validation\_element.py.
And I noticed that the current implementation is already consistent in that it handles all exceptions in a global context. So we should be fine.

I'll check the code one last time and than it should be ready to merge 🎉

PR #5881, opened 2026-03-15
Fix invalid Python code shown in SPA demo

jmerle:
### Motivation

The demo code at [https://nicegui.io/#single-page_applications](https://nicegui.io/#single-page_applications) currently shows invalid Python code (`Table.on(...)` is sort-of called twice):

image

After this change:

image

### Implementation

Moved a misplaced `# HIDE` comment to the correct line.

### Progress

### DIFF
--- a/website/documentation/intro.py
+++ b/website/documentation/intro.py
@@ -35,7 +35,8 @@ def create_intro() -> None:
                 {'name': 'New York', 'lat': 40.7119, 'lon': -74.0027},
                 {'name': 'London', 'lat': 51.5074, 'lon': -0.1278},
                 {'name': 'Tokyo', 'lat': 35.6863, 'lon': 139.7722},
-            ]).on('row-click', lambda e: sub_pages._render('/map/{lat}/{lon}', lat=e.args[1]['lat'], lon=e.args[1]['lon']))  # nopep8 # HIDE
+            ]).on('row-click',  # HIDE
+                  lambda e: sub_pages._render('/map/{lat}/{lon}', lat=e.args[1]['lat'], lon=e.args[1]['lon']))  # HIDE
             # ]).on('row-click', lambda e: ui.navigate.to(f'/map/{e.args[1]["lat"]}/{e.args[1]["lon"]}'))
 
         def map_page(lat: float, lon: float):


evnchn:
Hello @jmerle 

Yep it's a bug. But I doubt it was written wrong by demo author, and I suspect it's ruff linting doing its own thing and wrecking the demo. 

I'm running the CI. Let's see if it's happy. If it's not, please be prepared to add ignore rules or otherwise for a CI-compliant fix. Thanks!

evnchn:
Hmm ruff is happy. Maybe it is a typo, then. 

Perhaps run format on a VSCode instance configured to CONTRIBUTING.md as a final check, but I think it's a "we can merge right away". Cc @falkoschindler

jmerle:
Oops, my bad, I should've ran autopep8 as per the CONTRIBUTING.md. That seems to be the cause of the comment getting moved away, I've made it ignore the relevant line in the latest commit.

website/documentation/intro.py
```diff
@@ -35,8 +35,7 @@ def table_page():
                 {'name': 'New York', 'lat': 40.7119, 'lon': -74.0027},
                 {'name': 'London', 'lat': 51.5074, 'lon': -0.1278},
                 {'name': 'Tokyo', 'lat': 35.6863, 'lon': 139.7722},
-                # HIDE
-            ]).on('row-click', lambda e: sub_pages._render('/map/{lat}/{lon}', lat=e.args[1]['lat'], lon=e.args[1]['lon']))
+            ]).on('row-click', lambda e: sub_pages._render('/map/{lat}/{lon}', lat=e.args[1]['lat'], lon=e.args[1]['lon']))  # nopep8 # HIDE
```
falkoschindler: Thanks for the pull request, @jmerle!
I assume some kind of auto-formatter moved the `# HIDE` comment in PR #5696. But I can't reproduce it with Ruff or autopep8. The line seems to be too complex to be split automatically by autopep8. Therefore I'd like to remove the nopep8 marker (which we don't use anywhere else). Or is it needed from your point of view?

```suggestion
            ]).on('row-click', lambda e: sub_pages._render('/map/{lat}/{lon}', lat=e.args[1]['lat'], lon=e.args[1]['lon']))  # HIDE
```

evnchn: Since I asked "Perhaps run format on a VSCode instance configured to CONTRIBUTING.md as a final check" and then he had to do this, I suspect it's VSCode doing the formatting mangle, not autopep8 nor ruff.

falkoschindler: Yes, but my VSCode doesn't change this line although ms-python.autopep8 is installed, up-to-date and running. Therefore I'm _suggesting_ to remove the `# nopep8` marker.

jmerle: Hmm strange, Autopep8 was doing the mangling for me, not VS Code. I ran this command from the CONTRIBUTING
md:
```bash
uv run autopep8 --max-line-length=120 --in-place --recursive .
```

jmerle: I don't have access to my personal laptop for the next 12 hours by the way. so I can't give more information than that for now.

falkoschindler: Ok, then there's some difference between our `autopep8` tools which I don't understand...
Anyway, I decided to simply break the long line to avoid the need for any "ignore this line" markers.
Should be good to merge. 👍🏻

PR #5883, opened 2026-03-16
Fix `User` test simulation for select options with `None` as value

### LINKED ISSUE #5880: Allow "None" as value for ui.select
### First Check

### Example Code

```python
import pytest
from nicegui import ui
from typing import Literal


@pytest.mark.nicegui_main_file(__file__)
async def test_asdf(user: User) -> None:
    await user.open("/")

    # Find the select dropdown
    select = user.find(ui.select)
    select_element = next(iter(select.elements))
    assert isinstance(select_element, ui.select)
    assert select is not None

    # Verify the select dropdown shows "-" initially
    await user.should_see("-")

    # Verify both options are available in the dropdown
    select.click()
    await user.should_see("--mode-fast")

    # Select the first option (--mode-fast)
    user.find("--mode-fast").click()

    # Wait a moment for the UI to update
    await user.should_see("Fast mode")

    # Verify that the correct element is selected
    assert select_element.value == "fast"

    # Switch to empty/none option to test deselection
    select.click()
    await user.should_see("-")
    user.find("-").click()

    # After deselecting, the attribute should be reset to default (None)
    assert select_element.value is None


def main():
    def _render_lit(name: None | Literal["fast"]) -> ui.element:
        match name:
            case None:
                return ui.element()
            case "fast":
                with ui.row() as row:
                    ui.label("Fast mode")
                    ui.checkbox()
                    return row

    render_lit = ui.refreshable(_render_lit)
    choices = {None: "-", "fast": "--mode-fast"}
    default_choice = None

    with ui.row(align_items="center") as root:
        selector = ui.select(
            choices,
            value=default_choice,
            on_change=lambda ev: render_lit.refresh(ev.value),
        )
        render_lit(selector.value)
    return root

if __name__ == "__main__":
    ui.run(main)
```

### Description

The test code will fail on the last assertion. That is because `ui.select` will update it's visuals but will not actually change it's value if it is `None`. 
If I replace the value of the choice with anything else, e.g. the string `"None`", it works as expected.

``` 
'fast' != None
Expected :None
Actual   :'fast'
```

As a side note, I found ValueElements to struggle with `None` in general. `ui.input`, for example, will show the string `'None'` when setting the value to `None`.

### NiceGUI Version

3.8.0

### Python Version

3.13.11

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

falkoschindler:
### Motivation

Closes #5880.

When using `None` as a dictionary key in `ui.select` options (e.g. `{None: '-', 'fast': '--mode-fast'}`), the `User` test fixture fails to set the value back to `None` when clicking that option. The real browser behavior is correct — this is purely a bug in the simulated click logic.

### Implementation

The `UserInteraction.click()` method used `None` as the default sentinel for `next()` when looking up a select option key by its label. When `None` is a legitimate option key, the subsequent `if target_value is not None` check incorrectly treats it as "not found" and skips the value update.

Replaced `None` with a local `_NOT_FOUND` sentinel object and switched to identity comparison (`is not _NOT_FOUND`).

### Progress

- [x] Does `ui.radio` simulation need a similar fix? → No.

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -87,11 +87,12 @@ class UserInteraction(Generic[T]):
 
                 if isinstance(element, ui.select):
                     if element.is_showing_popup:
+                        _NOT_FOUND = object()
                         if isinstance(element.options, dict):
-                            target_value = next((k for k, v in element.options.items() if v == self.target), None)
+                            target_value = next((k for k, v in element.options.items() if v == self.target), _NOT_FOUND)
                         else:
-                            target_value = self.target if self.target in element._values else None  # pylint: disable=protected-access
-                        if target_value is not None:
+                            target_value = self.target if self.target in element._values else _NOT_FOUND  # pylint: disable=protected-access
+                        if target_value is not _NOT_FOUND:
                             # User clicked on a valid option: update the value and close the popup
                             if element.multiple:
                                 if target_value in element.value:
--- a/tests/test_user_simulation.py
+++ b/tests/test_user_simulation.py
@@ -555,6 +555,22 @@ async def test_select_keeps_value_when_toggling_popup(user: User):
     await user.should_see('value = Apple')
 
 
+async def test_select_none_value(user: User) -> None:
+    @ui.page('/')
+    def _():
+        select = ui.select({'a': 'A', None: 'B'}, value=None)
+        ui.label().bind_text_from(select, 'value', lambda v: f'Value: {v}')
+
+    await user.open('/')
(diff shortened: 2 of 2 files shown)


falkoschindler:
Regarding the checklist item about `ui.radio`: **no, it doesn't need the same fix.** The `ui.select` bug was caused by the `if target_value is not None` guard, which treated a legitimate `None` key as "not found." The `ui.radio` code path has no such guard — it always assigns `element.value = target_value`, so `None` keys work correctly as-is.

PR #5887, opened 2026-03-17
Support clicking `ui.tab` in user simulation

### LINKED ISSUE #5885: UserInteraction.click() does not work properly for ui.tabs
### First Check

### Example Code

```python
from nicegui import ui

async def test_select_none_value(user: User) -> None:
    event_history = []

    def on_event(v: str):
        event_history.append(v)

    @ui.page('/')
    def _():
        # Tabs and a Button; each click should log an event
        with ui.tabs() as tabs:
            ui.tab("A")
            ui.tab("B")
        tabs.on_value_change(lambda ev: on_event(ev.value))
        ui.button("X").on_click(lambda ev: on_event("X"))

    ui.run()

    await user.open("/")

    assert event_history == []

    # Clicking the button to demonstrate that the list can be modified by callbacks.
    user.find("X").click()
    assert event_history == ["X"]

    # Clicking the tab doesn't work though.
    user.find("A").click()
    assert event_history == ["X", "A"]  # fail
```

### Description

Clicking a `ui.tab` via `UserInteraction` in tests does not actually change the value of the tab. The `value` does not change and no change event is emitted.

### NiceGUI Version

3.8.0

### Python Version

3.13.11

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

falkoschindler:
### Motivation

Clicking a `ui.tab` element in the user simulation testing framework had no effect — the parent `ui.tabs` value was never updated, so `on_change` callbacks were not triggered. This made it impossible to test tab switching with the `User` fixture.

Fixes #5885.

### Implementation

Added a `ui.tab` case to `UserInteraction.click()`, following the same pattern used for `ui.radio` and `ui.select`: when a tab is clicked, its parent `ui.tabs` element's value is set to the tab's name, which triggers the `on_change` callback.

A corresponding test (`test_switching_tabs`) verifies the fix.

### Progress

- [x] Pytest has been added.

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -114,6 +114,11 @@ class UserInteraction(Generic[T]):
                     element.value = target_value
                     return self
 
+                elif isinstance(element, ui.tab):
+                    if isinstance(element.tabs, ui.tabs):
+                        element.tabs.value = element.props['name']
+                    return self
+
                 elif isinstance(element, ui.tree) and isinstance(self.target, str):
                     NODE_KEY = element.props.get('node-key')
                     LABEL_KEY = element.props.get('label-key')
--- a/tests/test_user_simulation.py
+++ b/tests/test_user_simulation.py
@@ -843,3 +843,15 @@ async def test_storage_tab_persists_across_navigation(user: User) -> None:
     await user.open('/other')
     user.find('Read value').click()
     await user.should_see('ABC')
+
+
+async def test_switching_tabs(user: User) -> None:
+    @ui.page('/')
+    def _():
+        with ui.tabs(on_change=lambda e: ui.notify(f'Switching to {e.value}')):
+            ui.tab('A')
+            ui.tab('B')
+
+    await user.open('/')
+    user.find('A').click()
+    await user.should_see('Switching to A')


evnchn:
CC: The test was clicking tab A, which may already be the active tab by default. Pushed a small fix: click B first, then A, so the test actually proves tab switching works in both directions.

falkoschindler:
@evnchn Ok. The active tab is `None` by default (no tab panel visible). But your commit makes it a bit more explicit what we're doing.

falkoschindler:
On second thought, the original test was stricter and secured us against a possible regression of "A" becoming the default. So I lean towards reverting 5d50e1852bca3eaeeab0d7b38b9bfa55bdcd9d7f.

evnchn:
Not sure. I don't see your point yet. Maybe we can add both, since user fixture test so cheap?

falkoschindler:
The original test (commit 1) only clicks tab "A" — the first tab. The default `ui.tabs` value is `None` (no tab selected, no panel visible). This is actually a stricter test because:

- With the correct default of `None`, clicking "A" changes the value from `None` to `"A"` and fires `on_change` → test passes.
- If a regression causes "A" to become the default value on page load, clicking "A" would **not** change the value, so `on_change` would **not** fire → test fails, catching the regression.

The second commit (clicking B then A) weakens this: even if "A" silently becomes the default, clicking B still changes the value (from "A" to "B"), and clicking A after B changes it again (from "B" to "A"). Both `on_change` events fire regardless of what the initial value was — so the test passes either way, masking the regression.

That's why I lean towards reverting the second commit: the simpler, original test is more sensitive to unexpected changes in default/initial value behavior.

evnchn:
Sure. I agree with you now.

PR #5834, opened 2026-02-22
Surface documentation index for AI tooling and discoverability

evnchn:
### Motivation

When developers evaluate UI libraries, AI compatibility is increasingly a deciding factor — both for direct LLM-assisted development and because AI-powered search summaries (Google AI Overview, Perplexity, ChatGPT search) now heavily influence which libraries get recommended. NiceGUI already serves its full documentation as machine-readable JSON indices, but this capability was completely undiscoverable — buried as an implementation detail with no public-facing documentation.

This should have been surfaced from day one. The indices power our own site search and are ideal for RAG pipelines, MCP servers, and AI agent tooling, yet no one could find them without reading the source code.

This PR makes the documentation index a first-class, documented feature.

### Implementation

**New documentation section** in Configuration & Deployment:
- Schema table describing the JSON structure (`title`, `content`, `format`, `url`, `demo` fields)
- Comparison table of the three available endpoints (`sitewide_index.json` ~140k tokens, `search_index.json` ~100k tokens, `examples_index.json` ~3k tokens) with entry counts, token estimates, and use cases
- Call for community contributions (MCP servers, AI agent skills, enhanced RAG implementations)

**Homepage AI CTA** (dark-box section between Installation and Features):
- Headline: "Or, let your AI do it"
- Copy highlighting that most LLMs already know NiceGUI, with a link to the documentation index for when that isn't enough
- "Learn more" button linking to the new documentation section

**Documentation overview** update:
- Added an "If you are an AI" note with a direct link to the machine-readable JSON index, so AI consumers landing on the documentation entry point can immediately find the structured data

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.
- [x] Pytests have been added (or are not necessary).
- [x] Documentation has been added (or is not necessary).

PR #5889, opened 2026-03-18
Add `Accept: text/markdown` content negotiation for NiceGUI pages

evnchn:
### Motivation

As mentioned in https://github.com/zauberzeug/nicegui/discussions/5833#discussioncomment-15895255, there is now the `Accept: text/markdown` header, which I believe Claude Code actively uses (since when it visits https://code.claude.com/docs/en/setup#alpine-linux-and-musl-based-distributions it correctly gets the Markdown version of the site).

However, currently NiceGUI websites shows up as a monolithic non-understandable JSON blob to AI agents as evidenced by https://github.com/zauberzeug/nicegui/discussions/5833#discussioncomment-16102263

I would like to implement a **NiceGUI-level API** such that ~~all~~ opted-in NiceGUI websites are **automatically agent-friendly** (short of interactivity, but that's expected).

### Implementation

Allow LLMs, CLI tools, and agents to request a markdown representation of any NiceGUI page by sending `Accept: text/markdown` in the HTTP request. Browsers continue to receive HTML as usual.

**Opt-in (experimental):** existing apps keep their current behaviour. A page only serves Markdown if it has been opted in:

- Global default on `ui.run(markdown=True)` / `ui.run_with(markdown=True)`.
- Per-page override on `@ui.page('/', markdown=True|False)`; `None` (default) falls back to the global flag.
- Marked experimental; the default may flip to `True` in NiceGUI 4.0.

**Rendering:**

~~The implementation adds a `_to_markdown()` method to the base `Element` class with smart duck-typing dispatch (content → text → label+value → recurse children), minimizing per-element overrides. Non-visual elements opt out via a `MARKDOWN_SKIP = True` class attribute.~~

Every element defines its own `_render_markdown()` returning a string ~~(or `None` to be skipped)~~. Built-in visible elements (`ui.label`, `ui.markdown`, `ui.link`, `ui.image`, `ui.table`, `ui.checkbox`, …) carry explicit one-liner overrides; layout containers (`ui.card`, `ui.row`, `ui.column`, `ui.expansion`, …) recurse into children; non-visual elements (`ui.timer`, `ui.dark_mode`, …) inherit ~~`None`~~ empty strings from `Element` ~~and are skipped~~. This removes the MRO traps of dispatch-by-mixin and makes behaviour self-documenting per element. User-defined subclasses of `ui.label`, `ui.card`, etc. inherit the correct markdown behavior automatically; subclasses of `ui.element` that render via their own Vue component are skipped unless they provide their own `_render_markdown`.

**Lifecycle:** Clients created for markdown responses are cleaned up via a Starlette `BackgroundTask` since they will never receive a WebSocket connection.

### End-to-end results

image

(note: cannot use the typical Fetch tool, since that only works for publically-accessible URLs)

### Progress

- [x] Documentation is not necessary for an underlying AI-facing change?
  - [ ] Or do we want to make some docs regardless?


rodja (dismissed):
I really like the concept. Might also be noteworthy in https://github.com/zauberzeug/nicegui/pull/5834. I have not tested/reviewd the PR in detail and hope that @falkoschindler can do that part.

evnchn:
@rodja If we have `Accept: text/markdown` then we likely would not highly advertise the sitewide index, but rather the markdown compatibility then. Because with the big JSON you need the agent to learn to read it, but agent already know how to read Markdown files.

rodja:
We should also track markdown deliveries in plausible so we can see how much llms looking things up.

evnchn:
Perhaps, but it will consume custom events, and we need to do Plausible on server-side which could be a hassle (at least we aren't doing it right now, so we need bootstrapping)

rodja:
Yes, its only nice to have and should not block the merge of this feature.

evnchn:
I'd say 3.11?

falkoschindler (changes requested):
Yes, 3.11 sounds good. I just did a first review and there are still some things to do and/or to think about:

1. **`ChatMessage._markdown_text` stores HTML-processed text, making the roundtrip lossy**

   In `chat_message.py:58`, `self._markdown_text = text` is assigned _after_ the text has been HTML-escaped and `\n` has been replaced with `<br />` (lines 49-51). The `_to_markdown` method then tries to reverse this with `html.unescape()` and `.replace('<br />', '\n')`, but this roundtrip is lossy:
   - If the original text _literally_ contained `<br />` or HTML entities like `&amp;`, these would be incorrectly transformed.
   - When `text_html=True`, `_markdown_text` stores raw HTML which `_to_markdown` doesn't properly handle (it would leave HTML tags in the markdown output).

   **Fix**: Store the original, unprocessed text before escaping. Add a `self._original_text = text` (or rename) _before_ line 49, and use that in `_to_markdown`.

   ```diff
   +       self._original_text = list(text)  # store before HTML processing
           if not text_html:
               text = [html.escape(part) for part in text]
               text = [part.replace('\n', '<br />') for part in text]
               sanitize = False
   -       self._markdown_text = text
   ```

   And adjust `_to_markdown` to use `self._original_text` directly instead of reversing the escaping.

2. **Duck-typing dispatch can produce misleading markdown for layout/value elements**

   The base `_to_markdown()` in `element.py:229` checks `hasattr(self, 'label') and hasattr(self, 'value')` as a fallback. This matches any `ValueElement` with a label property, including elements where `value` has a non-textual meaning:
   - `Splitter` (value = split percentage, e.g. `50`) → would produce `": 50"` or just `"50"`
   - `Drawer` (value = open/closed boolean) → `": True"`
   - `Carousel` (value = current slide name) → could be confusing
   - `LinearProgress` / `CircularProgress` (value = progress float) → `": 0.7"`
   - `Rating` (value = star count) → `": 3"`
   - `Knob` (value = angle/number) → `": 42"`
   - `Pagination` (value = current page) → `": 1"`

   These elements should either override `_to_markdown()` to skip their `value` and recurse into children (for containers like `Drawer`, `Footer`, `ScrollArea` whose children carry the real content), or produce meaningful output (for `Rating`, `Progress`), or be skipped entirely (for `Splitter` where the value is a split percentage).

   **Suggestion**: Containers (`Drawer`, `Footer`, `ScrollArea`, `Carousel`) should override to recurse into children only. `Splitter` should skip its value and recurse. `Fullscreen` should use `MARKDOWN_SKIP`. For `Slider`, `Knob`, `Rating`, `Pagination`, and `Progress`, consider custom overrides or skipping.

3. **`MARKDOWN_SKIP` class attribute placement in `Element`**

   The `MARKDOWN_SKIP` class attribute at `element.py:215` is placed in the middle of the class body between `_collect_slot_dict()` and `_to_dict()`. Class-level attributes should typically be grouped at the top of the class body, near other class-level declarations. This improves discoverability and follows the existing pattern in NiceGUI's element subclasses.

4. **Repeated visibility check pattern in every override**

   Every `_to_markdown` override begins with:

   ```python
   if not self.visible:
       return ''
   ```

   This is needed because overrides don't call `super()._to_markdown()` (which already has this check). Consider a template method pattern:

   ```python
   def _to_markdown(self) -> str:
       if not self.visible or self.MARKDOWN_SKIP:
           return ''
       return self._render_markdown()

   def _render_markdown(self) -> str:
       """Override this in subclasses."""
       # ... current duck-typing dispatch ...
   ```

   This eliminates the duplicated guard in every override and makes it impossible to forget the visibility check.

5. **`Dialog` and `Menu` should not unconditionally skip markdown**

   `Dialog`, `Menu`, and `Notification` all have `MARKDOWN_SKIP = True`, but all can be visible to the user:
   - `Dialog` and `Menu` can be open by default (`value=True`), so they should render children only when open.
   - `Notification` is always visible when created and carries a user-facing `message` — it should render that message instead of being skipped.

   The test `test_dialog_skipped` would need updating to cover both open/closed cases. Similar tests should be added for `Menu` and `Notification`.

6. **`Fullscreen` missing `MARKDOWN_SKIP`**

   `Fullscreen` is a non-visual control element (`ValueElement` with a boolean value) that would hit the label+value fallback and produce misleading output. It should have `MARKDOWN_SKIP = True`. (Other childless non-visual elements like `PageScroller`, `Skeleton`, and `Joystick` would produce empty strings via the base class recurse-into-children path, so they're harmless — but `Fullscreen` has a `value` that triggers the wrong dispatch.)

7. **Accept header parsing is simplistic**

   `'text/markdown' in accept` (client.py:155-156) doesn't handle quality values (`text/markdown;q=0.9`) or wildcard patterns (`text/*`). Starlette doesn't provide a built-in Accept parser, so a proper implementation would require a small custom parser. The current approach works correctly for the real-world use case (agents sending exactly `Accept: text/markdown`), but a code comment noting this simplification would be helpful.

8. **Button markdown: ambiguous syntax and icon-only buttons silently dropped**

   `[Click me]` (button.py:48) looks like an incomplete markdown link. Worse, icon-only buttons (no label) return an empty string and are silently omitted from the output. An icon button is still a meaningful interactive element. Consider using the icon name as fallback (e.g. `[icon:thumb_up]`) or always emitting `[Button]` when the label is empty. The `[...]` syntax itself could also be reconsidered — `**Click me**` or `[Button: Click me]` would be less ambiguous.

9. **Expansion hardcodes `###` (h3)**

   In `expansion.py:49`, the label is always rendered as `### label`. Nested expansions would all render at the same heading level. Consider making this relative to nesting depth, or use bold (`**Details**`) to avoid heading hierarchy issues.

10. **Simple `_to_markdown` overrides should be one-liners**

    Most overrides (`Checkbox`, `Switch`, `Button`, `Link`, `Image`, `Separator`, `Code`) are simple enough to condense into a single expression, e.g.:

    ```python
    def _to_markdown(self) -> str:
        return f'- [{"x" if self.value else " "}] {self._text or ""}' if self.visible else ''
    ```

    This reduces visual noise across 15+ overrides and makes the pattern scannable. Multi-step methods like `Table._to_markdown` and `Expansion._to_markdown` should stay as-is.

11. **Test coverage gaps**

    No tests for:
    - `ui.select` / `ui.radio` / `ui.slider` / `ui.number` without label
    - `ui.badge` / `ui.chip` (text elements)
    - `ui.html` element (content passthrough)
    - `ui.mermaid` (content passthrough)
    - `ui.date` / `ui.time` (value elements)
    - Nested elements that trigger the duck-typing fallback in unexpected ways
    - `text_html=True` on `ChatMessage`
    - Pages with `@ui.page` async handlers
    - Error scenarios (e.g., what happens when `_to_markdown` raises)

12. **`X-NiceGUI-Content: page` header**

    The custom header in `markdown_response.py:25` is undocumented. If it's intended for programmatic detection, a brief comment explaining its purpose would help.

evnchn:
@falkoschindler Your 12 feedbacks was in https://github.com/evnchn/nicegui/pull/130 and I looked over them before merging into this PR's branch

falkoschindler:
Apart from two minor commits, I have one proposal left to discuss:

## Proposal: Replace duck-typing dispatch with explicit per-class rendering

### Problem

The base `Element._render_markdown` uses a duck-typing chain to guess how to render an element:

```python
if hasattr(self, 'content') and isinstance(self.content, str):
    return self.content
if self._text is not None:
    return self._text
if hasattr(self, 'label') and hasattr(self, 'value'):
    # check if value is str/int/float...
    return f'{label}: {value}'
return self._children_to_markdown()
```

This probes for attributes at runtime without knowing what the element actually _is_. It works by accident for some elements and produces wrong output for others:

| Element                        | Attribute matched            | Duck-typing produces                       | Correct output                     |
| ------------------------------ | ---------------------------- | ------------------------------------------ | ---------------------------------- |
| **InteractiveImage**           | `content: str` (SVG overlay) | Raw SVG markup                             | `` like Image              |
| **Stepper**                    | `value: str` (step name)     | `": step_name"` — looks like form data     | Recurse into step content          |
| **Tabs**                       | `value: str` (tab name)      | `": tab_name"` — looks like form data      | Recurse into tab panels            |
| **Fab**                        | `label` + `value: bool`      | Skipped only because bool fails type check | Should show children (FAB actions) |
| **Drawer**, **Carousel**, etc. | `value` (structural)         | Need explicit override to bypass dispatch  | Just recurse into children         |

The last row is why `Drawer`, `Carousel`, `Footer`, `ScrollArea`, and `Splitter` all have identical `_render_markdown` overrides that just call `self._children_to_markdown()` — they exist solely to prevent the duck-typing from kicking in.

It also means every new element must be checked against the dispatch chain to see if it accidentally matches — a maintenance trap.

### Solution

Use `None` as the default return value to mean "skip this element in markdown", and require explicit `_render_markdown` overrides on each leaf element.

- **`Element._render_markdown`**: returns `None` — skip by default. No `MARKDOWN_SKIP` class variable needed.
- **No mixin-level overrides** on `ContentElement`, `TextElement`, or `ValueElement`. NiceGUI's elements mix multiple base classes freely (e.g., `Chip` inherits from both `TextElement` and `ValueElement`), so adding `_render_markdown` to mixins creates MRO conflicts and unpredictable behavior. Each element knows best how to represent itself.
- **Each element with visible content** defines its own `_render_markdown`. Most are trivial one-liners (e.g., `return self._text or ''` for Label, `return self.content or ''` for Markdown). Elements that are already overridden in the PR stay as-is.
- **Layout containers** (`ui.card`, `ui.row`, `ui.column`, etc.) return `self._children_to_markdown()` to recurse into children.
- **Non-visual elements** (Timer, Keyboard, DarkMode, etc.) need no annotation — they inherit `None` from `Element` and are skipped automatically.

`_children_to_markdown` and `_to_markdown` treat `None` returns from children as "nothing to contribute" and skip them.

**A CI test** verifies that every built-in element explicitly defines `_render_markdown` (returning either a string or `None`). This catches new elements that forget to handle markdown — the test fails, the author adds a one-liner, done.

**What this gives us:**

- No duck-typing dispatch, no `MARKDOWN_SKIP`, no MRO surprises.
- New elements default to invisible in markdown. Authors opt in by defining `_render_markdown`.
- User-defined elements without `_render_markdown` are silently skipped — no crash.
- ~30 trivial overrides across built-in elements, each self-documenting and easy to review.

evnchn:
**TL;DR**: Full agreement with the proposal. The duck-typing dispatch can go — idiomatic user-defined elements inherit correct markdown behavior from the official NiceGUI element they extend.

<details>
<summary>How we got there (with Claude Code's help)</summary>

I initially wanted to push back — the duck-typing dispatch gave user-defined elements "free" markdown support, and I thought that was the only way to make it work without users knowing about `_render_markdown`.

Claude Code (Opus, acting on my behalf) walked me through why that concern doesn't hold up:

- **Leaf elements**: Users should compose with `ui.label`, `ui.markdown`, etc., which already have explicit `_render_markdown`. The `_text`/`content` duck-typing branches never needed to fire for user elements.
- **Containers**: Users should extend `ui.column`, `ui.card`, `ui.row`, etc., which return `self._children_to_markdown()`. Children render correctly via inheritance.

My critical misunderstanding was not realizing that `class MyWidget(ui.column)` inherits the recursion behavior automatically. I was stuck thinking about `class MyWidget(ui.element)` — which is like grabbing the uncooked batter from the back kitchen instead of ordering the pancake from the menu. Non-idiomatic usage shouldn't drive the design.

Once that clicked, the duck-typing's downsides (your InteractiveImage/Stepper/Tabs table) clearly outweigh its benefits. The proposal correctly pushes rendering responsibility to the official NiceGUI elements — the "menu items" users should actually reach for — with no surprises.

</details>

falkoschindler:
I finished reviewing and refactoring the code. Just one question remains:

Subclasses of existing elements like `ui.card` or `ui.label` inherit a working `_render_markdown` from their parent and show up correctly in Markdown responses without any extra effort.

My concern is custom elements that subclass `ui.element` directly and register their own Vue component, with the visible content living in props rather than in children or `_text`. A user-defined `Rating` element that draws stars from an integer prop would be a typical example. Neither `ui.element` nor the Vue component declaration contributes a `_render_markdown`, so the element silently disappears from the Markdown response. The HTML response keeps rendering fine, so the author has little reason to notice.

evnchn:
Short of a generic Vue-to-Markdown bridge, which surely isn't happening, we cannot improve the situation directly...

On "The HTML response keeps rendering fine, so the author has little reason to notice.", indeed the _current situation_ is that, since the Markdown transport is only used by agents, that it basically is invisible to humans, and **active effort** must be spent to ensure content readability. 

To fix the effort situation, we can consider to add `(browser-only content hidden)` to the markdown for "foreign" Vue templates (so those done by the user, not NiceGUI) to be surfaced to the user **via** the agent?

falkoschindler:
I just talked to Rodja and we came to the following conclusion:

- A placeholder like `(browser-only content hidden)` is an interesting idea. But I guess we would need to implement overrides for every individual built-in element to distinguish our empty-by-default case from a user-missed-an-override case. Sounds like a lot of overhead for a rare edge case.
- We're going to add an opt-in so that existing NiceGUI apps don't change their LLM representation automatically. In NiceGUI 4.0 the Markdown rendering might be the new default. We can mark it as experimental to give us a bit more wiggle room for the final API.

falkoschindler (approved):
Ok, after adding an opt-in `markdown` parameter I think this PR is ready to merge 🚀

evnchn:
Moving to 3.12 — coordinated with Falko, this needs further discussion and shouldn't block the 3.11 release.

evnchn:
Wait @falkoschindler you got it and we can 3.11?

falkoschindler:
@evnchn Yes, we can merge and release in 3.11 if you agree.

evnchn:
@falkoschindler Consider setting `markdown=True` for https://nicegui.io for an instant boost in AI understandability. Then let's merge. 

To clarify though: The implementation is likely not perfect, and we will be polishing it up into 4.0 until it's good.

falkoschindler:
@evnchn Yes, after merging and releasing this PR, nicegui.io will support Markdown responses: https://github.com/zauberzeug/nicegui/pull/5889/changes#diff-b10564ab7d2c520cdd0243874879fb0a782862c3c902ab535faabe57d5a505e1

PR #5917, opened 2026-03-26
Add uPlot Element

NvNielen:
### Motivation

This PR introduces support for uPlot, a high-performance, memory-efficient 2D plotting library. This was originally requested in #2669, but also motivated by issues related to large dataset plotting such as #3340. Compared to other plotting libraries, it has [a smaller memory footprint and less overhead on rendering](https://github.com/leeoniya/uPlot/blob/master/perf.png), which makes it ideal for processing large datasets at high refresh rates. 

### Implementation

The core implementation wraps uPlot as a NiceGUI element, based on [uplot-vue](https://github.com/skalinichev/uplot-wrappers), which exposes uPlot's config and data update mechanisms.

The update logic should be structured to allow zoom/drag state to persist across data updates, as demonstrated in #2669.
Some important technical considerations include keeping the API simple and extensible and implementing tests.

### Progress

- [X] uPlot base element and functionality
- [x] Persist zoom/drag across updates
- [x] Pytests, should include rendering and updates.
- [x] Add documentation and provide clear, runnable demos, since uPlot's documentation is limited.


NvNielen:
Any input and/or feedback very much appreciated!

falkoschindler:
Thanks for this PR! uPlot is a great choice for high-performance charting, and this is a solid start.

We tested the element successfully with the following code:

```python
from nicegui import ui
from nicegui.elements.uplot.uplot import UPlot
import math

x_data = list(range(100))
y_data = [math.sin(i * 0.1) for i in range(100)]

chart = UPlot({
    'width': 800,
    'height': 400,
    'title': 'Sine Wave',
    'series': [{}, {'stroke': 'red', 'label': 'sin(x)'}],
}, data=[x_data, y_data])

@ui.button('Add Point').on_click
def add_point():
    x_data.append(len(x_data))
    y_data.append(math.sin(len(y_data) * 0.1))
    chart.update_data([x_data, y_data])

ui.run()
```

A few things we noticed while looking at the code:

- The other ESM-based elements (plotly, echart, codemirror) have an `__init__.py` that re-exports the class, plus a `ui.plotly(...)` / `ui.echart(...)` registration. Something to keep in mind for when this gets wired up to the `ui` namespace.

- `uplot.py` imports NumPy unconditionally, but it's only used in type hints. A `TYPE_CHECKING` guard would avoid making NumPy a hard dependency:

  ```python
  from typing import TYPE_CHECKING
  if TYPE_CHECKING:
      import numpy as np
  ```

Looking forward to seeing this evolve!

NvNielen:
Hello @falkoschindler, thank you for the feedback! 

I have incorporated your suggestions and added a few more things, including zoom persistence and a plugin system. Need to play around with it a little more to see if there are any bugs. Here's an example with some integrated plugins:

```py
import math

from nicegui import ui
from nicegui.elements.uplot.uplot import UPlot

x_data = list(range(100))
y_data = [math.sin(i * 0.1) for i in range(100)]

chart = UPlot({
    'width': 800,
    'height': 400,
    'title': 'Sine Wave',
    'series': [{}, {'auto': False, 'stroke': 'red', 'label': 'sin(x)'}],
    'scales': {
        'x': {'time': False},
        'y': {'range': [-2.5, 2.5]},
    },
    'plugins': [{'name': 'wheelZoom', 'options': {'factor': 0.8}}, {'name': 'legendAsTooltip'}]
}, data=[x_data, y_data])

def add_point():
    x_data.append(x_data[-1] + 1 if x_data else 0)
    y_data.append(math.sin(x_data[-1] * 0.1))
    if len(x_data) > 100:
        x_data.pop(0)
        y_data.pop(0)
    chart.update_data([x_data, y_data])

ui.timer(0.05, add_point)

ui.run()
```

One can also import custom plugins using NiceGUI and reference them.

That's it for now, once again any feedback is welcome!

falkoschindler:
Thanks for pushing this forward, @NvNielen!

Before we dive into a full review, a few quick nudges so the remaining checklist work happens in the right direction:

### Housekeeping

- The branch is **56 commits behind main**. We don't usually ask contributors to rebase, but at this distance it's probably worth a merge/rebase from main so CI runs against a current baseline.
- `check / pylint` fails: move `from ...element import Element` in `nicegui/elements/uplot/uplot.py` above the `if TYPE_CHECKING:` block (C0413 wrong-import-position).
- The element isn't exported yet — please wire it into `nicegui/ui.py` and `nicegui/elements/__init__.py` (grep for `echart` to see all the spots).

### Suggestion: shrink the scope of this first PR

Right now the PR carries quite a lot: the base element, zoom/drag persistence, three bundled plugins, plus downsampling and docs still to come. For a first contribution I'd strongly suggest trimming it down to just the **base uPlot element** — a minimal, well-tested wrapper with clean data/options flow. Everything else (plugins, zoom persistence, downsampling) can land as follow-up PRs once the core is solid.

This makes the review tractable, lets us merge something useful sooner, and gives you a cleaner foundation to build on. The current scope is large enough that it's slowing the PR down rather than speeding it up.

### High-level direction for the remaining scope

A few things worth aligning on now, regardless of which items stay in this PR:

1. **Bundled plugins are manually copied from upstream demos.** `plugins/wheelzoom.js`, `plugins/legendtooltip.js` and `plugins/touchzoom.js` are lifted from https://leeoniya.github.io/uPlot/demos/ (the URLs are still in the comments), and `utils/uplot_helpers.js` looks adapted from [uplot-vue](https://github.com/skalinichev/uplot-wrappers). We'd like to avoid the maintenance burden of manually kept forks — upstream fixes won't reach us automatically, and it's not obvious to a future maintainer what's "ours" vs. "theirs". Unfortunately uPlot's npm package only ships `dist/`; the demo plugins aren't published anywhere, so `package.json` alone can't fetch them. Preferred approach for us is an **automated sync at build time**: extend `rollup.config.mjs` to download the plugin files from a pinned uPlot tag (`https://raw.githubusercontent.com/leeoniya/uPlot/v${version}/demos/...`, with `${version}` read from the `uplot` dep in `package.json`) and bundle them into `dist/`. That way the plugins never live in our source tree, there's one version to bump, and they can't silently drift. If that turns out to be impractical, the fallback is to drop the plugins from this PR and expose a plugin hook so users (or follow-up PRs) can provide their own.

2. **Plugin API surface.** Plugins are currently selected by string names (`wheelZoom`, `touchZoom`, `legendAsTooltip`) resolved via a JS registry, with a `window[name]` fallback. That ties users to the three bundled plugins unless they do custom JS. Before committing to it in tests/docs, could you sketch the Python-side plugin API? (e.g. a string in `options['plugins']`, or a dedicated `ui.uplot.plugin(...)` concept?) Easier to get right now than to revise later.

3. **Data updates and `reset_scales`.** The behavior is subtle: `reset_scales=True` currently _does_ call `setData(data)` (which resets), but then also tries to restore zoom if the x-range matches the previous data. That mixed behavior is hard to reason about.
   - Should `reset_scales` be a constructor default that `update_data` can override per-call? Right now `update_data` has no way to opt out.
   - Would two explicit modes (e.g. `append` vs. `replace`) be cleaner than a boolean whose meaning shifts?

4. **`update_data` / `update_options` vs. direct prop assignment.** NiceGUI elements typically expose data as bindable props (see `ui.echart` — users just mutate `chart.options`). Is there a reason `update_*` methods are necessary here, or can we drop them and let the existing `_props` flow handle it? Fewer methods = less surface to document and test.

5. **Downsampling.** Before implementing: where do you plan to do it — Python-side, JS-side, or configurable? uPlot itself doesn't downsample, but on large datasets you usually want to avoid shipping the full array over the websocket. A brief design note in the PR description would help before coding. (My vote: defer this to a follow-up PR.)

### Next steps

I'd suggest: rebase + fix pylint, then decide on scope (my recommendation: trim to base element only), then align on the open API questions before writing tests and docs. Happy to do a thorough review once that's in.

NvNielen:
Hello @falkoschindler, thank you very much for your already in-depth review. There are indeed some good points you brought up worth aligning on, so let's dive into those:

### Points regarding scope

1. For the plugins, while they are derived from the demos (hence the references), both `plugins/wheelzoom` and `plugins/legendtooltip` have been adjusted/improved. This also applies to the helper functions, which are indeed derived from [uplot-vue](https://github.com/skalinichev/uplot-wrappers), but adjusted to work for the plugins and data persistence. Unfortunately this means they cannot be simply be copied over. Therefore I think that we should keep the plugin support, but certainly drop the plugins themselves for now and expose a plugin hook. How do you envision this could be done?
2. Plugins are defined at the python-end in `options['plugins']` as one or more of:
    - a string with the plugin name, or
    - an object containing the plugin name and options that will be passed to the plugin initializer

    Then, depending on whether a plugin exists in the registry or attached to the window, they can be resolved. Now this leaves a bit of responsibility to the user for the correctness of the input, as this is done at the browser-end. Therefore this will be mentioned in the documentation.

3. You are right about the ambiguity here as one is a subset of the other. I will replace this with a scale_mode parameter, containing three states:
    - "preserve", restore scales + zoom
    - "reset_with_zoom", reset scales but restore zoom
    - "reset". reset scales + zoom

4. The `update_*` methods are not necessary here. Let's stick to direct prop assignment.

5. Downsampling should definitely be done at the python-end. It will be very similar, if not identical to how it is handled in #5704 - [tsdownsample](https://github.com/predict-idlab/tsdownsample) is hard to beat in terms of performance.  Let's create a follow-up PR once that is up and running.

### How I plan to proceed

I will indeed rebase + fix pylint, trim to base element + plugin support + zoom persistence, then once we agree on the above, I will proceed with the tests and docs.

falkoschindler:
Thanks for the thorough response, @NvNielen — your plan sounds good to me. Replies in your numbering:

### 1. Plugins → postpone

You're right that we agreed to drop the bundled plugins, so let's not over-design now. I'd suggest going further and removing the plugin registry / resolver / `plugin_loader.js` plumbing too for this PR — no plugin code at all. Once there's a concrete plugin to ship (yours, ours, or a user request), we'll have real requirements to design against. Right now anything we sketch is speculative.

### 2. Plugin API form

Moot if we drop plugin support entirely for this PR — let's revisit when (1) comes back.

### 3. Scale mode

Before we settle on names, can you spell out what each mode does in concrete uPlot terms (which `setData` flag, which scale gets reset, etc.)? I want to make sure all three are genuinely distinct user-visible behaviors and not just two variations of "reset". `reset_with_zoom` in particular reads ambiguously to me.

If three modes really are needed, something like `'preserve'` / `'rescale_keep_zoom'` / `'reset'` (or `'preserve'` / `'reset_data'` / `'reset_all'`) would be more parallel. Also: which one should be the default?

### 4. `update_*` methods

Great — direct prop assignment is the right call.

### 5. Downsampling

Agreed, follow-up PR using `tsdownsample`. Alignment with #5704 is a nice plus.

### Scope

With (1) and (2) postponed, the scope becomes **base element + zoom persistence**, which is a much cleaner first PR. Zoom persistence stays because it's intertwined with the data update flow — pulling it out and re-adding later would mean redoing the same redesign twice.

Looking forward to the next push.

NvNielen:
Hello @falkoschindler! Thank you once again for your replies. I have fixed pylint and the element export, rebased and trimmed down to the base element + zoom persistence. Some notes:

### Plugins
I agree that plugin support introduces unnecessary complexity to this PR, so it has been dropped for now. I do hope we can follow up on this at some point, since it does add some nice customizability :)

### Scale mode
There indeed still was some ambiguity. The three modes are now named: `preserve_all`, `preserve_zoom`, and `reset`. Their effects in uPlot terms are as follows:

| Mode | setData `_resetScales` flag | When are x/y scales reset? |
| ------ | -------------- | ---------------------------------------- |
| preserve_all | false        | Never (scales/zoom perserved)          |
| preserve_zoom | true         | Only when not zoomed |
| reset | true         | Always (scales/zoom reset) |

Therefore, zoom preservation is an extension of `setData`'s `_resetScales `option, which allows for scales to be preserved specifically on zoom. To still support the old behavior, having three modes instead of two makes more sense in my opinion.

I have set `reset` as the default mode, since this is `uPlot-vue`'s default behavior. 

Hope this clears things up! Let me know whether the naming now makes sense within the context of their behavior.

falkoschindler:
Thanks @NvNielen — the scale mode design is convincing, happy with `preserve_all` / `preserve_zoom` / `reset` and `reset` as the default.

I pushed a few small follow-ups directly: an `options` setter for symmetry with `data`, a `Literal` annotation on `scale_mode`, a `default: 'reset'` on the JS prop, and some docstring polish. Nothing structural — feel free to push back if anything looks off.

Please go ahead with tests and docs. I'll do a thorough end-to-end review after.

NvNielen:
Thank you for the follow-ups @falkoschindler! I have adjusted the data typing slightly and added some tests. 

While making the tests I ran into a problem; the uPlot element does not auto scale to the parent container/window if no height/width is defined (see `test_in_card`). Unfortunately, uPlot does not have built-in functionality for that. Fortunately, [there is a way to rescale to the window size](https://leeoniya.github.io/uPlot/demos/resize.html), but that does not fix the problem for parent NiceGUI elements.

I assume this functionality is required. Therefore, my question to you: Is there a way to fix this scaling issue for NiceGUI elements?

falkoschindler:
Thanks for the update, @NvNielen!

If I understand the question correctly, you can find a solution in the implementation of `ui.echart`. It uses a `ResizeObserver` on the host element and calls the chart's resize method whenever the parent size changes — see [`echart.js#L72-L80`](https://github.com/zauberzeug/nicegui/blob/44cc19e5b9fc86d398a92060eaad1e8b21f9444e/nicegui/elements/echart/echart.js#L72-L80). For uPlot, the equivalent is:

```js
this.resizeObserver = new ResizeObserver(() => {
  if (this._chart) {
    this._chart.setSize({ width: this.$el.offsetWidth, height: this.$el.offsetHeight });
  }
}).observe(this.$el);
```

…and disconnect it in `unmounted()`. The echart implementation also has a small guard against the initial spurious resize callback that you may want to copy.

One caveat: uPlot requires `width` and `height` in its initial options. So the user still passes them — they just act as the initial dimensions, and the `ResizeObserver` keeps the chart in sync afterwards. (`ui.echart` works the same way.)

Once that's in and the docs demo is ready, ping me and I'll do a thorough end-to-end review.

Noghpu:
Hi, I'd very much be excited to see this finished.

To hopefully help it along, I vibe-coded the auto-resize suggested above, a `ResizeObserver` on the host, modeled on `ui.echart`. Width worked right away.

While demoing I found the problem with the **title and legend**: uPlot renders them as DOM siblings of the plot, and `setSize()` sizes only the plotting area, so filling the host with the canvas pushes the title/legend out of the box.

The solution I went with: reserve room for the title/legend and shrink the plot so the **whole chart** fits its container, the same contract as `ui.echart` (with a `.nicegui-uplot` default, you size it the usual way, `w-full h-full`). Worth flagging that this **isn't idiomatic to uPlot's own sizing**: per uPlot's docs, `setSize`/`width`/`height` cover only the plotting area and treat the title/legend as variable extra, i.e. uPlot expects the element to grow by them, rather than the plot to shrink. I went with the echart-style "whole chart fits the box" because it matches the rest of NiceGUI's layout model.

evnchn:
*Posted by Claude Code on @evnchn's behalf — every empirical below was run locally against the actual branches; commands included for replay.*

**TL;DR:**

@Noghpu (Phuong Dao) actually solved the two things still blocking this PR — container auto-resize and documentation — on a fork branch that was never linked here, so the work was invisible. It's real and it's good.

I folded those 5 commits onto the PR head (authorship preserved) plus one small commit de-flaking two timing-fragile tests. Result: **CI goes from red → green** — 17/17 uPlot tests, stable across 3 runs, with `mypy`/`pylint`/`pre-commit` all clean.

One honest carry-over: the `preserve_zoom` path still has a large-data crash (`Math.min(...)` on the full x-series) that predates this work — flagged below, not fixed here.

This is a second-pair-of-eyes assist, not a master-order — @falkoschindler, the end-to-end review you reserved is still yours to make.

<details>
<summary><b>Before → after (ran both branches)</b></summary>

| | Before — PR head `aa72725d` | After — folded `d63378ac` |
|---|---|---|
| `quick-test` pytest (CI gate) | 🔴 **red** (confirmed on the PR's own checks + locally 3/8 fail) | 🟢 **17/17**, 3 consecutive runs, 0 flakes |
| Resize to container | ❌ none — `test_in_card` returns canvas height `'0'` | ✅ fills card **568×273**, title + legend reserved inside the box |
| Documentation (last unchecked box) | ❌ none | ✅ 4 `@doc.demo` blocks + `doc.reference` |
| `mypy ./nicegui` | ✅ | ✅ 240 files |
| `pylint ./nicegui` | ✅ | ✅ 10.00/10 |

The 3 "before" failures: `test_in_card` (no resize implemented), plus `test_update_*` which raced the async round-trip. The resize one is a genuine feature gap Phuong Dao closed; the update ones just needed a `screen.wait` after the click.
</details>

<details>
<summary><b>What landed (6 commits, authorship intact)</b></summary>

```
7fe62d08  Phuong Dao  Auto-resize ui.uplot to its container and add documentation
f3241ba2  Phuong Dao  Reserve room for uPlot title and legend when resizing
3ba0fb8a  Phuong Dao  Simplify ui.uplot scale_mode to use uPlot's setData(resetScales)
51b7c12e  Phuong Dao  Add resize, zoom, lifecycle and legend tests for ui.uplot
a6a44b04  Phuong Dao  Test that a content-sized ui.uplot host does not grow unbounded
d63378ac  evnchn      Stabilize flaky uPlot resize/update tests with explicit waits
```

Highlights of Phuong Dao's work — idiomatic to the rest of NiceGUI:
- `ResizeObserver` on the host modeled on `ui.echart`, **plus** the subtle bit: uPlot renders title/legend as DOM siblings, so a naive `setSize` overflows the box. He reserves their measured height (re-measured on the next `rAF`) so the *whole* chart fits — same contract as other NiceGUI chart elements.
- `default_classes='nicegui-uplot'` + adding `.nicegui-uplot` to the shared `width:100%; height:16rem` block in `nicegui.css` — exactly how `ui.echart`/`ui.leaflet` default-size.
- Replaces the hand-rolled `_restoreZoomState` with uPlot-native `setData(data, !zoomedIn)`, dropping dead code.

6-file footprint, nothing unrelated rode along. Cherry-picks onto the PR head with **zero conflicts**.
</details>

<details>
<summary><b>Carry-over: large-data crash in `preserve_zoom` (pre-existing, not fixed here)</b></summary>

`uplot.js` (the `data` watcher) computes the zoom check with a spread:

```js
const zoomedIn = x.min > Math.min(...prevData[0]) || x.max < Math.max(...prevData[0]);
```

`Math.min(...arr)` passes every x-value as a function argument, which overflows the call stack on large series — the dataset size uPlot is chosen for (cf. #3340).

Confirmed (Node, V8):
```
n=100000 -> Math.min OK
n=500000 -> THROWS: Maximum call stack size exceeded
```
Safari's argument limit is far lower (~65k). Empty `prevData` also makes `prevData[0]` `undefined` → throws. One-line fix (a loop instead of spread) if wanted; I left it out so this push stays a faithful fold-in of the resize/docs work rather than mixing a third concern.
</details>

<details>
<summary><b>Replay</b></summary>

```bash
# fold-in branch
git fetch origin pull/5917/head:pr-5917
git fetch https://github.com/Noghpu/nicegui.git uplot-element-work
git checkout -B uplot-combined pr-5917
git cherry-pick 785ea010^..34c3c7ba         # Phuong Dao's 5 commits

# gates (all green)
.venv/bin/python -m pytest tests/test_uplot.py    # 17 passed
.venv/bin/mypy ./nicegui                            # Success, 240 files
.venv/bin/pylint ./nicegui                          # 10.00/10
```
</details>

Open question for the thread: fold this in as above (resize + docs + green CI now), and track the `Math.min` large-data fix as a quick follow-up — or would you rather keep zoom-on-large-data in scope here?

NvNielen:
Hello @Noghpu, thanks a lot for helping out! Apologies for the delay. 

The additions look good to me. There were two small things that I found during testing:
1. The `width` and `height` in the options didn't really do anything - the chart always scaled to its parent regardless of what you set. I've made them act as the chart's default size (still responsive and overridable via Tailwind classes or `.style()`), and changed the docs/docstring which incorrectly stated they were required as initial dimensions.
2. On the dark documentation site the charts were hard to see, since uPlot renders a transparent background by default. I have added `.classes('bg-white text-black')` to the doc examples so they stay legible in both light and dark mode.

I also addressed the call stack bug found by @evnchn. Since uPlot requires the x-values to be sorted ascending, it now just reads the first and last entries instead.

Ready for your review now @falkoschindler. Feel free to revert any of these changes if you disagree with them!

falkoschindler:
We didn't manage to review this in time for 3.15, so I'm moving it to the 3.16 milestone. It's a sizeable addition and we want to give it a proper review. Thanks for your patience!

falkoschindler:
Thanks again for this, @NvNielen – and @evnchn for the resize/test work. We did a full AI-assisted review pass on the branch and fixed what was clear-cut directly (see the last commits): `setData(data, false)` never redraws in uPlot, so `preserve_all` and a zoomed-in `preserve_zoom` showed stale data; any `:`-prefixed option made every data update recreate the chart (functions compared by identity); plus some simplification and packaging cleanup (CSS now shipped like xterm/leaflet, unused dev dependencies removed, docs links added).

Two design points are left where we'd rather hear from you than decide on our own. **Neither blocks the merge** – if we don't hear back within about a week we'll merge as is and turn them into follow-up issues.

**1. `preserve_zoom` zoom detection (@NvNielen)** – the "zoomed in" check compares `scales.x.min/max` against the raw first/last x value. With `scales.x.distr: 2` uPlot keeps the x scale in index space, so the check is always "zoomed" and the chart never refits; a fixed `scales.x.range` inside the data extent has the same effect; and a y-only zoom isn't detected. Our suggestion: compare against uPlot's own auto range instead – `let [min, max] = x.distr === 2 ? [0, xs.length - 1] : [xs[0], xs[xs.length - 1]]; [min, max] = x.range(u, min, max, "x");` and treat `x.min !== min || x.max !== max` as zoomed – and document that `preserve_zoom` refers to the x axis (uPlot re-autoscales y on any x change anyway). Does that match how you intended the mode? A test with `distr: 2` would be good either way.

**2. Title/legend reservation on content-sized hosts (@evnchn)** – with `.style('height: auto')` the plot ends up at 219 px instead of the configured 200 (the first `_resize` runs before uPlot's commit with `chrome` still 0), and `test_content_sized_host_does_not_grow` only asserts that the host stops growing. More generally the rAF re-measure loop only converges by interleaving and doesn't notice a legend that re-wraps without a host resize. Our suggestion: measure `.u-title`/`.u-legend` heights directly (they don't depend on the plot height, so no feedback loop), call `setSize` only when the size actually changed, skip the immediate `_resize` when the host had no height before creation, and let the `ResizeObserver` observe the legend/title too. Happy to take this one ourselves if you'd rather not – just say so.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20and%20fix%20proposals%20from%20the%20model%27s%20review%20of%20the%20branch%2C%20framing%20and%20decisions%20from%20the%20maintainer.&a=claude-code "claude-fable-5: Findings and fix proposals from the model's review of the branch, framing and decisions from the maintainer.")

evnchn:
*Drafted by Claude Code working with @evnchn — answering the ownership question on point 2.*

I'll take point 2 — the root cause is exactly the `rAF` you spotted (the first `setSize` runs with `chrome` still 0), fix and measurements to follow shortly.

evnchn:
Taking point 2 — @Noghpu (Phuong Dao) wrote that resize code and I folded it onto the branch, so it is mine to fix; patch and writeup to follow shortly.

evnchn:
Pushed 89bad7b6.

**The root cause was only the `rAF`.** `chrome` was applied a frame late, so the very first `setSize` ran with `chrome = 0` — and on a content-sized host that wrong first value feeds the next measurement, so the loop settled on a fixed point contaminated by its own mistake. Measuring it synchronously takes the plot from 219 back to the configured 200.

**All four of your suggestions ended up in — but I nearly shipped without three of them.** I bisected them by "do the tests still pass", they did, so I called them not load-bearing. That oracle was too weak: the tests did not cover the cases they protect. Two were real.

- **`setSize` only when the size changed** — uPlot's own guard covers the bbox recompute but not the redraw, and `ResizeObserver` fires on sub-pixel changes that leave the integer `offsetWidth`/`offsetHeight` untouched.
- **Observe the legend and title** — reachable by moving the mouse: the live legend shows wider values, grows 62 → 92 px with no host resize, and the chart overflows its box at 330 against a 300 px host. Now covered by a test.

Only "skip the immediate `_resize`" stayed out — on its own it gives 258, worse than the 219 baseline, and it has no job left once chrome is measured synchronously.

<details>
<summary><b>Why <code>root.offsetHeight - chart.height</code> instead of summing the two elements</b></summary>

It is the same quantity — `root` is title + wrap + legend, and `wrap` is exactly `chart.height`, so the plot height cancels and there is no feedback loop either way. Two reasons to prefer the subtraction:

- `offsetHeight` excludes margins. Give the legend `margin: 20px 0` and the direct sum under-reserves, overflowing the host by 20 px.
- It absorbs any future chrome sibling automatically, where a `.u-title`/`.u-legend` sum silently under-reserves if uPlot adds one or renames a class. Its only assumption is that `chart.height` matches the wrap's laid-out height, which is uPlot's own sizing invariant.

</details>

<details>
<summary><b>Evidence for the two I got wrong</b></summary>

**The redraw guard.** From the vendored bundle — the `&&` gates only `kt` (the bbox recompute); `qn()`, the commit and redraw, runs unconditionally:

```js
function vt(e,l,t){(t||e!=c.width||l!=c.height)&&kt(e,l),Wn(!1),gt=!0,pt=!0,qn()}
```

**The legend growth.** Fixed 220x300 host, three labelled series, wide values, mouse moved over the plot:

| | host | chart root | legend |
|---|---|---|---|
| before hover | 300 | 300 | 62 |
| after hover, without observing the legend | 300 | **330** | 92 |
| after hover, with it | 300 | 300 | 92 |

</details>

<details>
<summary><b>Tests</b></summary>

23 passed, 0 skipped, `pre-commit` clean. Both new guards were verified to fail against the pre-fix file, not merely to pass against the new one:

- `test_content_sized_host_does_not_grow` — strengthened as you asked, now asserts the canvas keeps the configured 200 rather than only that the host stops growing. Fails `assert '219' == '200'`.
- `test_growing_legend_fits_in_box` — new, covers the hover case above. Fails `assert 330 <= 300`.

Local only so far: the CI Gate run for this push is sitting at `action_required` and needs a maintainer to approve it, so treat the above as local evidence rather than a green tick.

</details>

<details>
<summary><b>One pre-existing issue I did not fix</b></summary>

`_resize` measures the host with `offsetHeight`, the border box. If the host element itself carries a border or padding, the chart is sized against the full box instead of the content box and overflows. On `.classes('w-full h-full').style('border: 10px solid red; padding: 8px')` in a 400x300 parent: host `offsetHeight` 300 against `clientHeight` 280, canvas 242, chart root bottom 333.8 against a host bottom of 316.

Identical numbers before and after this patch — that line is untouched — so it is out of scope here. The fix is to subtract the computed padding from `clientHeight` on both axes. Happy to send it separately.

</details>

<details>
<summary><b>Point 1</b></summary>

Not mine to answer — @NvNielen knows what he meant by the mode better than I do. Your `x.range(u, min, max, "x")` suggestion reads right to me for `distr: 2`, but I would rather not pre-empt him.

</details>

PR #5924, opened 2026-04-01
fix(xterm): preserve writes issued before mount in hidden tab panels

### LINKED ISSUE #5839: Unmounted xterm write data lost.
### First Check

### Example Code

```python
from nicegui import ui


@ui.page("/")
def page():

    with ui.tabs().classes("w-full") as tabs:
        one = ui.tab("One")
        two = ui.tab("Two")
    with ui.tab_panels(tabs, value=one).classes("w-full"):
        with ui.tab_panel(one):
            ui.label("First tab")
        with ui.tab_panel(two):
            terminal = ui.xterm({"cols": 80, "rows": 24})

    def write_to_terminal():
        terminal.write("Hello, Terminal!\n")

    ui.button("Write to terminal", on_click=write_to_terminal)


    ui.run(dark=True, reload=False, show=False)
```

### Description

I am working through updating to v3. I was using a custom element for xterm but decided to try to give the new official one a try. When you attempt to write to an xterm element which has not been mounted yet, in this case because it is in an unclicked tab, the write does not go to the element and the data is lost. I understand why this happens to some degree, an unmounted element isn't rendered/instantiated yet so can't receive data. In my case my custom element was buffering data if my element has not been mounted. For reference my implementation that I have been using prior:
```js
import "xterm";

export default {
  template: "<div></div>",
  mounted() {
    this.terminal = new Terminal(this.options);
    this.terminal.open(this.$el);

    this.ready = true;
    this.$emit("ready", { state: true });
  },
  beforeDestroy() {
    this.terminal.dispose();
  },
  beforeUnmount() {
    this.terminal.dispose();
  },
  methods: {
    call_api_method(name, ...args) {
      this.terminal[name](...args);
    },
    reset() {
      this.terminal.write("\x1bc");
    },
    write_base64_encoded_hex(data) {
      const hexString = data.base64_encoded_hex;
      if (hexString !== "" && this.ready) {
        let byteArray = new Uint8Array(
          hexString.match(/.{1,2}/g).map((byte) => parseInt(byte, 16))
        );
        let base64String = new TextDecoder("utf-8").decode(byteArray);
        const decodedData = atob(base64String);
        this.terminal.write(decodedData);
      }
    },
  },
  props: {
    options: Object,
  },
};
```
```py
class Terminal(ui.element, component="lib/terminal.js", dependencies=["lib/xterm.js"]):
    def __init__(self, options: Dict[str, Union[str, int]]) -> None:
        super().__init__()
        self._props["options"] = options
        self._ready = False
        self.on("ready", self.handle_ready, ["state"])
        self._write_buffer = b""

    async def wait_on_ready(self):
        while not self.ready:
            await asyncio.sleep(0.01)

    def call_terminal_method(self, name: str, *args) -> None:
        self.run_method("call_api_method", name, *args)

    def write(self, data: Union[str, bytes]) -> None:
        if isinstance(data, str):
            data = data.encode("utf-8")
        if self.ready:
            base64_encoded_hex = base64.b64encode(data).hex()
            self.run_method("write_base64_encoded_hex", {"base64_encoded_hex": base64_encoded_hex})
        self._write_buffer = self._write_buffer + data

    async def handle_ready(self, e: GenericEventArguments):
        state = e.args["state"]
        if state:
            self.run_method("reset")
            await asyncio.sleep(0.25)
            base64_encoded_hex = base64.b64encode(self._write_buffer).hex()
            self.run_method("write_base64_encoded_hex", {"base64_encoded_hex": base64_encoded_hex})
        self._ready = state
```

I don't know if this buffering is required for other applications, but it critical for mine. Most xterm elements I create have writes to them way before they are actually mounted, even if only by race conditions. I could extend the built in element probably, is there something I could use inside of my `ready` property to trigger my buffer to flush?

### NiceGUI Version

3.8.0

### Python Version

3.11.2

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

wwenrr:
## Summary
- fix `ui.xterm` calls (`write`, `writeln`, `input`, and `run_terminal_method`) when invoked before the Vue component is mounted
- queue pending terminal method calls and replay them once `mounted()` creates the terminal instance
- keep behavior unchanged when terminal is already mounted

## Why
Issue #5839 reports lost output when writing to an xterm inside an inactive `ui.tab_panel`. In that case the component can be unmounted and method calls happen before `this.terminal` exists on the frontend, so output is dropped.

## Tests
- add regression test `test_write_while_xterm_is_in_inactive_tab_panel`
  - writes to xterm while hidden in tab "Two"
  - switches to tab "Two"
  - asserts the text is present

## Notes
I could not run the test suite in this environment because `pytest` is unavailable, but the added test follows existing `Screen` patterns in this repository.

Closes #5839

### DIFF
--- a/nicegui/elements/xterm/xterm.js
+++ b/nicegui/elements/xterm/xterm.js
@@ -3,6 +3,13 @@ import { loadResource } from "../../static/utils/resources.js";
 
 export default {
   template: "<div></div>",
+  data() {
+    return {
+      terminal: null,
+      fit_addon: null,
+      pending_calls: [],
+    };
+  },
   mounted() {
     // Create terminal with addons
     this.terminal = new Terminal(this.options);
@@ -17,34 +24,48 @@ export default {
         this.terminal[key]((e) => this.$emit(key.slice(2).toLowerCase(), e));
       });
 
+    for (const [name, args] of this.pending_calls) {
+      runMethod(this.terminal, name, args);
+    }
+    this.pending_calls = [];
+
     // NOTE: wait for window.path_prefix to be set
     this.$nextTick().then(() => loadResource(window.path_prefix + `${this.resourcePath}/xterm.css`));
   },
   methods: {
+    run_terminal_or_queue(name, ...args) {
+      if (!this.terminal) {
+        this.pending_calls.push([name, args]);
+        return;
+      }
+      return runMethod(this.terminal, name, args);
+    },
     getRows() {
-      return this.terminal.rows;
+      return this.terminal?.rows;
     },
     getColumns() {
-      return this.terminal.cols;
+      return this.terminal?.cols;
     },
     fit() {
-      this.fit_addon.fit();
+      if (this.fit_addon) {
+        this.fit_addon.fit();
+      }
     },
     input(data, wasUserInput = true) {
-      return this.terminal.input(data, wasUserInput);
+      return this.run_terminal_or_queue("input", data, wasUserInput);
     },
     write(data) {
-      return this.terminal.write(data);
+      return this.run_terminal_or_queue("write", data);
     },
     writeln(data) {
-      return this.terminal.writeln(data);
+      return this.run_terminal_or_queue("writeln", data);
     },
     run_terminal_method(name, ...args) {
       if (name.startsWith(":")) {
         name = name.slice(1);
         args = args.map((arg) => new Function(`return (${arg});`).call(this.terminal));
       }
-      return runMethod(this.terminal, name, args);
+      return this.run_terminal_or_queue(name, ...args);
     },
   },
   props: {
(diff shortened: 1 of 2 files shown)


falkoschindler:
Thanks for the PR, @wwenrr! Note that the test is currently failing.

But more importantly, this is the same class of problem we're facing with AG Grid in hidden tabs (#3033, PR #5645): elements that need to work before they're mounted. Rather than fixing each element individually, we're considering a more general solution -- likely a keep-alive element that ensures children stay mounted even when hidden (see the discussion in #5839).

We'll keep this PR open but **blocked** until we've reached a conclusion on the general approach. If we end up going element-by-element after all, your PR is a great starting point for the xterm side. This is high priority for us.

falkoschindler:
Following up on my earlier comment: the general approach landed in #5951 as `ui.keep_alive`, so we're closing this in its favor.

For #5839, the fix is now:

```python
with ui.tab_panel('Two'):
    with ui.keep_alive():
        term = ui.xterm()
term.write('hello')  # delivered even before tab 'Two' is opened
```

We went with keeping the same component instance alive rather than buffering calls because xterm.js carries state beyond the write log (cursor position, ANSI modes, scrollback), which a replay-on-mount path can desync.

Thanks again, @wwenrr — the reproducer and issue report were a useful push toward the general fix.

PR #5926, opened 2026-04-01
Add strict parameter to cpu_bound and io_bound to propagate cancellation

calebgregory:
Closes #5925

## Summary

`_run` silently returns `None` when the app is stopping, the executor is shut down, or the task is cancelled — violating the declared `-> R` return type. This causes two problems:
1. Callers cannot distinguish a legitimate `None` return value from a cancellation
2. Code that trusts the `R` return type hits downstream `TypeError`s at runtime when `R` is non-optional

This PR adds a `strict` keyword parameter (default `False`) to `cpu_bound` and `io_bound`:
- `strict=True` → raises `run.CancelledError` (subclass of `asyncio.CancelledError`) instead of returning `None`; return type is `R`
- Default / `strict=False` → preserves current runtime behavior; return type is honestly `R | None`

`Literal`-based overloads let the type checker narrow the return type based on the value of `strict`.

**Note:** This is a breaking type-level change. The default return type is now `R | None`, so consumers will see new mypy errors where they previously did not. Runtime behavior is unchanged for callers that do not pass `strict`.

## Test plan

- [x] Existing `test_run.py` tests pass (no behavioral regression)
- [x] New `test_returns_none_when_app_is_stopping` — verifies default behavior returns `None` (parametrized across `cpu_bound` and `io_bound`)
- [x] New `test_strict_raises_when_app_is_stopping` — verifies `strict=True` raises `run.CancelledError` (parametrized across `cpu_bound` and `io_bound`)
- [x] `ruff check` and `autopep8` clean

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -7,7 +7,7 @@ from concurrent.futures.process import BrokenProcessPool
 from contextlib import suppress
 from functools import partial
 from pickle import PicklingError
-from typing import Any, TypeVar
+from typing import Any, Literal, TypeVar, overload
 
 from typing_extensions import ParamSpec
 
@@ -18,6 +18,15 @@ thread_pool = ThreadPoolExecutor()
 
 P = ParamSpec('P')
 R = TypeVar('R')
+
+
+class CancelledError(asyncio.CancelledError):
+    """Raised by ``cpu_bound`` / ``io_bound`` when *strict=True* and the
+    background work is cancelled or the app is shutting down.
+
+    A subclass of :class:`asyncio.CancelledError` so that existing
+    ``except CancelledError`` handlers continue to work.
+    """
 
 
 def setup() -> None:
@@ -57,27 +66,47 @@ def safe_callback(callback: Callable, *args, **kwargs) -> Any:
         raise SubprocessException(type(e).__name__, str(e), traceback.format_exc()) from None
 
 
-async def _run(executor: Any, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
+async def _run(executor: Any, callback: Callable[P, R], *args: P.args,
+               strict: bool = False, **kwargs: P.kwargs) -> R | None:
     if core.app.is_stopping:
-        return  # type: ignore  # the assumption is that the user's code no longer cares about this value
+        if strict:
+            raise CancelledError('app is stopping')
+        return None
     try:
         loop = asyncio.get_running_loop()
         return await loop.run_in_executor(executor, partial(callback, *args, **kwargs))
     except RuntimeError as e:
         if 'cannot schedule new futures after shutdown' not in str(e):
             raise
-    except asyncio.CancelledError:
-        pass
-    return  # type: ignore  # the assumption is that the user's code no longer cares about this value
+        if strict:
+            raise CancelledError('cannot schedule new futures after shutdown') from e
+    except asyncio.CancelledError as e:
+        if strict:
+            raise CancelledError('task was cancelled') from e
+    return None
 
 
-async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
+@overload
+async def cpu_bound(callback: Callable[P, R], *args: P.args, strict: Literal[True],
+                    **kwargs: P.kwargs) -> R: ...
+
+
+@overload
+async def cpu_bound(callback: Callable[P, R], *args: P.args, strict: Literal[False] = ...,
+                    **kwargs: P.kwargs) -> R | None: ...
+
+
+async def cpu_bound(callback: Callable[P, R], *args: P.args, strict: bool = False,
+                    **kwargs: P.kwargs) -> R | None:
     """Run a CPU-bound function in a separate process.
 
     `run.cpu_bound` needs to execute the function in a separate process.
     For this it needs to transfer the whole state of the passed function to the process (which is done with pickle).
     It is encouraged to create static methods (or free functions) which get all the data as simple parameters (eg. no class/ui logic)
     and return the result (instead of writing it in class properties or global variables).
+
+    If *strict* is ``True``, a :class:`run.CancelledError` is raised instead of
+    silently returning ``None`` when the task is cancelled or the app is shutting down.
     """
     global process_pool  # pylint: disable=global-statement # noqa: PLW0603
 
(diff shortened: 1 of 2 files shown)


nicegui/run.py
```diff
@@ -20,6 +20,15 @@
 R = TypeVar('R')
 
 
+class CancelledError(asyncio.CancelledError):
+    """Raised by ``cpu_bound`` / ``io_bound`` when *strict=True* and the
+    background work is cancelled or the app is shutting down.
+
+    A subclass of :class:`asyncio.CancelledError` so that existing
+    ``except CancelledError`` handlers continue to work.
+    """
```
petergaultney, line 29: is there a strong reason to prefer a subclass vs just re-raising the existing exception... I suppose the one weird issue would be creating one in the first `if strict` pre-check. 

But I'm generally hesitant to make new exceptions, and especially so when they have the same name as an underlying one. 🤔

petergaultney (approved):
this looks pretty good to me. I have the one question about the exception definition, but no very strong preference there.

nicegui/run.py
calebgregory: I appreciate the hesitancy to make new exceptions.  I think that's why I had the impulse to name this one `nicegui.run.CancelledError` to match `asyncio`'s name.  But maybe it's better to give it a distinct name anyway, to prevent confusion.

I agree that raising a new `asyncio.CancelledError()` when the app is shutting down would be strange.

Defining the exception in `nicegui.run` offers some code-as-documentation + some consumer-convenience:

```py
import asyncio
from nicegui import run

try:
    run.cpu_bound(..., strict=True)
    ...
```

vs.

```py
from nicegui import run

try:
    run.cpu_bound(..., strict=True)
except run.CancelledError:
    ...
```

falkoschindler (commented):
Thanks for the PR, @calebgregory! The issue is real and the fix direction makes sense.

However, we're running into a mypy limitation with the current approach: you can't place additional keyword arguments between `P.args` and `P.kwargs`. That's why mypy reports "Arguments not allowed after ParamSpec.args" on every signature that mixes `strict` with `ParamSpec`.

We can work around this with overloads — using `ParamSpec` for the default (non-strict) overload and `Callable[..., R]` for the `strict=True` overload. That preserves argument type-checking for the default case. But `strict=True` loses `ParamSpec` checking, which is unfortunate since `strict=True` is meant to become the recommended mode (and eventually the default after the next major release).

Do you have ideas for how to keep `ParamSpec` argument checking in the `strict=True` case? Some options we've considered but aren't thrilled with:

- Separate functions (e.g. `cpu_bound_strict`) instead of a parameter — preserves `ParamSpec` but clutters the API
- A wrapper/decorator that converts the return type — might be too clever
- Waiting for PEP 612 extensions that allow extra kwargs alongside `ParamSpec` — not available yet

Open to other ideas if you see a cleaner way to thread this needle.

calebgregory:
Hey @falkoschindler!  Thanks for looking into and thinking about this.  I won't be able to thread the needle this week - I just don't have the time -, but should be able to get to it next week.

evnchn:
*Drafted by Claude Code working with @evnchn — after a longer brainstorm on the `ParamSpec` puzzle, Evan landed on a tighter backwards-compat stance and posted #6056 as the alternative.*

**TL;DR:** I think this PR should be closed in favor of #6056, which ships a type-only annotation fix + a `run.wrap_none` / `run.NONE_MARKER` helper for the rare callback-returns-None case. The `strict=True` kwarg approach has two backwards-compat blockers that I don't see a way around within the current major version.

@calebgregory @falkoschindler — appreciate the careful overload work; the `ParamSpec` puzzle you ran into turns out to be genuinely unsolvable in PEP 612 (`Protocol.__call__` workaround is silently accepted by mypy but rejected by pyright — see [python/typing#1905](https://github.com/python/typing/discussions/1905)). But even if mypy were happy, the runtime shape of `strict=True` has issues:

<details><summary>Two backwards-compat blockers</summary>

**1. `strict` kwarg silently eats a name from the callback's `P.kwargs`.**

```python
def db_query(*, strict: bool = False) -> Result:
    """strict=True raises on schema mismatch instead of warning"""
    ...

await cpu_bound(db_query, strict=True)  # strict=True is swallowed by cpu_bound;
                                         # db_query runs without it
```

`strict` is a common kwarg name (SQLAlchemy, Pydantic, JSON schema validators, dataclass conversion). Any callback that has its own `strict` kwarg silently misbehaves under this PR. The choice of keyword name doesn't fix this — `raise_on_cancel`, `bypass_silent_none`, anything could collide. The collision is structural to the "steal a kwarg from `P.kwargs`" design.

**2. Switching from silent-None to raising would break callers' running code.**

Falko's stated intent is to eventually make `strict=True` the default. But that's a runtime breaking change: callers today have code that *runs* (using `None` as the result, falling through to a default, optional-chaining downstream) when shutdown races a `cpu_bound` call. After the flip, that code raises `CancelledError` and stops running. Backwards compatibility says: don't change which code keeps running, even if the existing behavior is suboptimal.

</details>

<details><summary>What #6056 does instead</summary>

**Type-only** fix:
- `cpu_bound`, `io_bound`, `_run` annotated `R | None` to match runtime reality. mypy then correctly flags downstream usage that doesn't handle `None`.
- `nicegui/welcome.py` `or []`-guards the one internal caller that surfaced.

**For the rare case where the callback may itself return `None`** — new module-level helpers:
- `run.NONE_MARKER` — an `enum.Enum` member (singleton, picklable by name, `is` survives the `cpu_bound` process boundary)
- `run.wrap_none(fn)` — a callable class (not a closure, so the wrapper itself is picklable) that replaces a legitimate `None` with `NONE_MARKER`

```python
result = await run.cpu_bound(run.wrap_none(my_func), arg1)
if result is None:
    ...  # run was cancelled / app shutting down
elif result is run.NONE_MARKER:
    ...  # my_func returned None
else:
    ...  # real value
```

No new kwargs on `cpu_bound` / `io_bound` → no `P.kwargs` collision. No runtime behavior change → no broken callers.

</details>

@calebgregory — really sorry the PEP 612 puzzle turned out to have no clean solution; the overload design you wrote was the obviously-right thing to try first. @falkoschindler, happy to keep iterating on #6056 if you agree with the framing, or rework if not.

falkoschindler:
Thanks again for this PR and the careful overload work, @calebgregory! It was the obviously-right approach to try first — and it's what surfaced that the design can't work, which is genuinely valuable:

- The PEP 612 limitation turned out to be a hard dead end, not a puzzle to thread: there's no way to express the interleaved `strict` overloads that both mypy and pyright accept (see [python/typing#1905](https://github.com/python/typing/discussions/1905)).
- More fundamentally, as @evnchn pointed out, any keyword argument on `cpu_bound`/`io_bound` collides with the callback's own `P.kwargs` — a callback with its own `strict` parameter would silently lose it. That's structural to the design, not fixable by picking a different name.

So I'm closing this PR — not because the problem isn't real (it is, and your diagnosis and tests carry forward), but because the fix needs a different shape.

How exactly we'll proceed is still open: #6056 proposes an honest `R | None` annotation plus a `wrap_none` helper, and there's also the question of whether a future major release should raise on cancellation instead of returning `None`. We'll continue that discussion over there.

Thanks for kicking this off! 🙏

PR #5933, opened 2026-04-03
Skip echoing client-originated values via bool flag on ValueElement

### LINKED ISSUE #5185: Character loss in ui.input validation with high network latency

evnchn:
### Motivation

When a `ValueElement` (e.g., `ui.input`, `ui.select`) receives a value change from the client via LOOPBACK, the server currently echoes the same value back in the next outbox update. This is redundant — the client already has the value — and wastes bandwidth, especially for elements where the value is the main part of the payload.

This implements the approach proposed by @falkoschindler in #5728: a single boolean flag on `ValueElement` that tells the outbox to omit `VALUE_PROP` when the value originated from the client.

Supersedes #5728 with a much simpler implementation (no diff machinery, no deepcopy, no new merge semantics).

### Implementation

- **`ValueElement._value_from_client`** flag, set to `True` when the LOOPBACK handler processes a client-originated value change
- **`Outbox._build_element_dict()`** checks the flag via duck-typing (`getattr`); if set, strips `VALUE_PROP` from the props dict before sending
- **`nicegui.js`** merges incoming props with existing element props (`{ ...existing.props, ...element.props }`) so that an omitted `VALUE_PROP` preserves the client's current value rather than blanking it
- **Reconnection safety**: `_to_dict()` is unchanged and always returns full state — `build_response` is not affected by the flag
- **Flag reset**: the flag resets to `False` on any server-originated value change (`_handle_value_change` with `_send_update_on_value_change=True`), ensuring server-set values are always sent

### Progress

### DIFF
--- a/nicegui/elements/mixins/value_element.py
+++ b/nicegui/elements/mixins/value_element.py
@@ -31,6 +31,7 @@ class ValueElement(Element):
                  ) -> None:
         super().__init__(**kwargs)
         self._send_update_on_value_change = True
+        self._value_from_client: bool = False
         self.set_value(value)
         self._props[self.VALUE_PROP] = self._value_to_model_value(value)
         self._props['loopback'] = self.LOOPBACK
@@ -39,6 +40,7 @@ class ValueElement(Element):
         def handle_change(e: GenericEventArguments) -> None:
             self._send_update_on_value_change = self.LOOPBACK is True
             self.set_value(self._event_args_to_value(e))
+            self._value_from_client = self.LOOPBACK is True
             self._send_update_on_value_change = True
         self.on(f'update:{self.VALUE_PROP}', handle_change, [None], throttle=throttle)
 
@@ -124,6 +126,7 @@ class ValueElement(Element):
         with self._props.suspend_updates():
             self._props[self.VALUE_PROP] = self._value_to_model_value(value)
         if self._send_update_on_value_change:
+            self._value_from_client = False
             self.update()
         args = ValueChangeEventArguments(sender=self, client=self.client,
                                          value=self._value_to_event_value(value),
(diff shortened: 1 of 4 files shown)


evnchn:
Checks are failing...

evnchn:

Closing this. The latest commit reverted the strip-and-merge after it broke Quasar components that ride on `model-value` as UI state (Dialog/Menu/Expansion/Tabs). What remains is the `_value_from_client` flag as unused infrastructure — merging as-is ships zero observable behaviour and the PR description still describes the pre-revert implementation.

On reflection, the wire-side strip is structurally incompatible with NiceGUI's "server is authoritative for the full prop set" contract, regardless of per-element opt-ins. The JS-side merge would silently drop *every* server-initiated prop removal, not just `model-value` — any per-element marker only carves out which elements get the misbehaviour, not whether the misbehaviour exists.

**The replacement plan (sketching in fork now):** keep Falko's framing — "don't resend values that came from the client" — but do it server-side. Suppress the outbox `update()` when post-`set_value` state equals what the client just sent. Transforms / sanitisers / clamping / `on_change` corrections still flush normally because their post-state differs from the incoming value. Dialog/Menu/etc. are unaffected because their server-side value matches the client-driven toggle. No wire format change, no JS merge, no per-subclass marker.

Will reopen as a fresh PR once the fork branch is green and benched. The `_value_from_client` flag will be dropped — the equality check subsumes it (and is stricter, since it catches transforms the flag would miss).

cc #5185 — the equality-skip approach helps that bug directly because rapid typing without server side-effects produces zero echoes back to the client.

PR #5936, opened 2026-04-07
fix: propagate CancelledError in run.cpu_bound / run.io_bound

mango766:
## Summary

Fixes #5925.

`_run()` was silently swallowing `asyncio.CancelledError`, causing `cpu_bound` and `io_bound` to return `None` instead of propagating the cancellation. This made it impossible for callers to distinguish a legitimate `None` return from a cancelled task, leading to downstream `TypeError`s.

## Change

One-line change in `nicegui/run.py`: replace `pass` with `raise` in the `CancelledError` handler so the exception propagates to the caller.

## Test plan

- Existing test suite should pass (cancellation during shutdown is already handled by the `is_stopping` guard above)
- Callers that need to handle cancellation can now use `try/except asyncio.CancelledError`

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -67,7 +67,7 @@ async def _run(executor: Any, callback: Callable[P, R], *args: P.args, **kwargs:
         if 'cannot schedule new futures after shutdown' not in str(e):
             raise
     except asyncio.CancelledError:
-        pass
+        raise
     return  # type: ignore  # the assumption is that the user's code no longer cares about this value
 
 


falkoschindler:
Thanks for the contribution, @mango766!

Unfortunately, unconditionally re-raising `CancelledError` is a breaking change. The current `pass` was intentional -- callers of `cpu_bound()` / `io_bound()` that don't handle `CancelledError` would now see unhandled exceptions during cancellation or shutdown. Also, `except CancelledError: raise` triggers a Pylint warning (W0706 `try-except-raise`) since it's equivalent to not catching the exception at all.

The issue author in #5925 already proposed a backward-compatible design with an opt-in parameter, and #5926 implements exactly that (with tests and correct typing). So we'll go with that approach instead.

Closing in favor of #5926.

PR #5938, opened 2026-04-07
Fix RuntimeError when click handler deletes the clicked element during testing

### LINKED ISSUE #5937: Testing click() gives "RuntimeError: dictionary changed size during iteration"
### Discussed in https://github.com/zauberzeug/nicegui/discussions/5935

<sup>Originally posted by **samuller** April  6, 2026</sup>
### First Check

- [X] I have checked the [FAQs](/zauberzeug/nicegui/wiki/FAQs) for a similar question and didn't find it.
- [X] I used the GitHub search to find a similar [question](https://github.com/zauberzeug/nicegui/discussions/categories/q-a) and didn't find it.
- [X] I searched the NiceGUI documentation, with the integrated search.
- [X] I already searched online "How to X in NiceGUI" and didn't find any information.

### Example Code

```python
"""Minimal reproducible example for:

The bug:
- A button lives inside a container.
- Its click handler calls container.clear(), which deletes the button.
- NiceGUI's testing User.click() iterates `element._event_listeners.values()` to
  fire the handler.  Deleting the element mid-iteration calls _handle_delete() →
  _event_listeners.clear(), mutating the dict while it is being iterated.

Run with:
    pytest dict_iter_bug.py
"""

from nicegui import ui

pytest_plugins = ["nicegui.testing.user_plugin"]


def setup():
    @ui.page("/demo")
    def page():
        container = ui.row()

        def rebuild():
            container.clear()
            with container:
                # Button lives INSIDE the container it clears — that's the bug trigger.
                ui.button("click me", on_click=rebuild).mark("the_button")

        rebuild()


async def test_click_causes_self_destruction(user: User):
    setup()
    await user.open("/demo")
    # The button's handler calls container.clear(), which destroys the button while
    # NiceGUI testing is still iterating its _event_listeners dict →
    # RuntimeError: dictionary changed size during iteration
    user.find("the_button").click()
    await user.should_see("click me")
```

### Description

The above code replicates the issue when testing a button whose `click()` behavior causes itself to be re-created. I get this RuntimeError only during tests, but otherwise my code works perfectly fine:
```python
                    ...
                    if target_key in expanded_set:
                        expanded_set.remove(target_key)
                    else:
                        expanded_set.add(target_key)
                    element.props['expanded'] = list(expanded_set)
                    element.update()
                    return self
    
>               for listener in element._event_listeners.values():  # pylint: disable=protected-access
E               RuntimeError: dictionary changed size during iteration

.venv/lib/python3.12/site-packages/nicegui/testing/user_interaction.py:137: RuntimeError
```
I know I could try to hide/show the button rather than delete/re-create it, but for my full code the current approach works cleanly and it's only tests that give a problem.

Is this a bug in the test framework or is this a coding pattern I should be avoiding?

### NiceGUI Version

3.8.0

### Python Version

3.12.3

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_</div>

falkoschindler:
### Motivation

Fixes #5937. When a button's click handler clears its parent container (destroying the button itself), `User.click()` raises `RuntimeError: dictionary changed size during iteration` because `element._event_listeners` is mutated mid-loop via `_handle_delete()` -> `_event_listeners.clear()`.

### Implementation

After each handler fires, check `element.is_deleted` and `break` out of the listener loop. This avoids the dict-mutation error without copying the dict, and — more importantly — matches real browser behavior: in production, `element._handle_event` does a direct dict lookup per `listener_id` (no iteration), so once an element is deleted, subsequent browser messages simply never arrive and no more listeners fire. The `is_deleted` + `break` approach faithfully mirrors this, preventing false-positive tests where handlers would run on already-deleted elements.

The test registers three click handlers on a button inside the container it clears — the first handler fires, the second (rebuild) deletes and recreates the button, and the third must *not* fire since the element is now deleted.

### Progress

### DIFF
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -46,11 +46,13 @@ class UserInteraction(Generic[T]):
                             element.value = option
                             break
 
-                for listener in list(element._event_listeners.values()):  # pylint: disable=protected-access
+                for listener in element._event_listeners.values():  # pylint: disable=protected-access
                     if listener.type != event:
                         continue
                     event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args=args)
                     events.handle_event(listener.handler, event_arguments)
+                    if element.is_deleted:
+                        break
         return self
 
     def type(self, text: str) -> Self:
@@ -139,12 +141,14 @@ class UserInteraction(Generic[T]):
                     element.update()
                     return self
 
-                for listener in list(element._event_listeners.values()):  # pylint: disable=protected-access
+                for listener in element._event_listeners.values():  # pylint: disable=protected-access
                     if listener.element_id != element.id:
                         continue
                     args = not element.value if isinstance(element, (ui.checkbox, ui.switch)) else None
                     event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args=args)
                     events.handle_event(listener.handler, event_arguments)
+                    if element.is_deleted:
+                        break
         return self
 
     def clear(self) -> Self:
--- a/nicegui/testing/user_interaction.py
+++ b/nicegui/testing/user_interaction.py
@@ -46,7 +46,7 @@ class UserInteraction(Generic[T]):
                             element.value = option
                             break
 
-                for listener in element._event_listeners.values():  # pylint: disable=protected-access
+                for listener in list(element._event_listeners.values()):  # pylint: disable=protected-access
                     if listener.type != event:
                         continue
                     event_arguments = events.GenericEventArguments(sender=element, client=self.user.client, args=args)
@@ -139,7 +139,7 @@ class UserInteraction(Generic[T]):
                     element.update()
                     return self
 
-                for listener in element._event_listeners.values():  # pylint: disable=protected-access
+                for listener in list(element._event_listeners.values()):  # pylint: disable=protected-access
                     if listener.element_id != element.id:
                         continue
                     args = not element.value if isinstance(element, (ui.checkbox, ui.switch)) else None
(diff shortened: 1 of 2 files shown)


evnchn:
Hi — this is Claude Code, acting on behalf of @evnchn.

**The core question isn't just "does it stop erroring?" but "does the test simulation still behave like real life?"**

The `list()` snapshot fix prevents the `RuntimeError`, but it subtly diverges from production behavior. An alternative: check `element.is_deleted` after each handler and `break` — no copy, and it matches what actually happens in the real dispatch path.

<details>
<summary>Detailed analysis</summary>

**Real event dispatch** (`element._handle_event`): the browser sends one message per `listener_id`. The server does a direct dict lookup — no iteration. If a handler deletes the element, subsequent messages from the browser simply target a nonexistent element and never fire.

**`list()` snapshot approach**: prevents the `RuntimeError` by iterating a copy, but if a handler deletes the element mid-loop, the remaining listeners in the snapshot *still fire* — on an already-deleted element. This can't happen in real life.

**`is_deleted` + `break` approach**: iterates the dict directly (no copy, O(1) memory). After each handler fires, checks `element.is_deleted` — if the element was destroyed, breaks out of the loop. This mirrors real behavior: once an element is deleted, no more listeners fire.

The difference only matters for multi-listener scenarios, but aligning the simulation with reality prevents a class of false-positive tests where handlers run on deleted elements.

Commit: [`7445dee`](https://github.com/zauberzeug/nicegui/commit/7445dee1) on this branch.
</details>

evnchn (dismissed):
Feel free to merge if you agree, or discuss if not. Thanks!

evnchn:
And be sure to update PR desc.

falkoschindler:
Great catch, @evnchn! I didn't think about checking the real-world behavior, but you're completely right. I updated the pytest and PR description accordingly. 👍🏻

evnchn (approved):
The updated test, comment, and PR description all show that me, you, test suite, and reality, all align 4-way.

PR #5943, opened 2026-04-09
Fix in-page drag-and-drop in native mode

### LINKED ISSUE #5962: Native mode dragover handler blocks drag operations
### First Check

### Description

I am still using the closed sortable branch https://github.com/zauberzeug/nicegui/pull/4656, but when updating the local fork from nicegui 3.8 -> 3.10 I noticed that the drag operation was freezing. This behaviour is tho only happning in naitive mode, in a Chrome browser it works fine. 

So I further identified that its this PR: https://github.com/zauberzeug/nicegui/pull/5866 that introduced it.

So even tho I found this by acciident with the sortable fork, this issue would conflict with other drag and drop libs or maybe other drag functions?
The bind_drop() function in native_mode.py unconditionally prevents default on all dragover events.

The fix is pretty simple:

In native_mode.py -> Instead of:
```py
        window.dom.document.events.dragover += \
            webview.dom.DOMEventHandler(lambda _: 0, True, False)
```

this:
```py
  document.addEventListener('dragover', function(e) {
      if (e.dataTransfer && e.dataTransfer.types && e.dataTransfer.types.includes('Files')) {
          e.preventDefault();
      }
  });
```

I am not sure tho if this is the intended behaviour, thats why I opned an issue and not a PR

### NiceGUI Version

3.10

### Python Version

3.13.9

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

falkoschindler:
### Motivation

The native window event IPC bridge (#5866, released in 3.9.0) introduced a document-level `dragover` handler with `preventDefault()` to support dropping files from the OS into the pywebview window. This unconditionally prevents the default behavior for *all* dragover events, including in-page element drags. As a result, HTML5 drag-and-drop between elements (e.g. the Trello Cards example) stopped working in native mode.

Fixes #5940.

### Implementation

Replace the pywebview `DOMEventHandler` for `dragover` with a raw JavaScript listener that only calls `preventDefault()` when files are being dragged from outside the browser. This is detected by checking whether `dataTransfer.types` contains `"Files"`. In-page drags (which use types like `"text/plain"`) are left untouched, so element-level `dragover`, `dragleave`, and `drop` handlers work normally again.

The `drop` handler remains unchanged since its `preventDefault()` during bubbling is harmless for in-page drags (the element-level handler has already processed the event) and still necessary for native file drops (prevents the browser from navigating to the dropped file).

### Progress

- [x] Pytests are not necessary (native mode is currently untested).

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -64,8 +64,11 @@ def _bind_pywebview_events(window: webview.Window, event_sender: Connection) -> 
             pass
 
     def bind_drop() -> None:
-        window.dom.document.events.dragover += \
-            webview.dom.DOMEventHandler(lambda _: 0, True, False)  # type: ignore[arg-type]
+        window.evaluate_js('''
+            document.addEventListener("dragover", function(e) {
+              if (e.dataTransfer && e.dataTransfer.types.indexOf("Files") >= 0) e.preventDefault();
+            });
+        ''')
         window.dom.document.events.drop += \
             webview.dom.DOMEventHandler(lambda e: send('drop', files=[  # type: ignore[arg-type]
                 file_.get('pywebviewFullPath', '') for file_ in e.get('dataTransfer', {}).get('files', [])


evnchn (approved):
I checked the schema extensively and the check is OK:

<details>
<summary> Schema analysis </summary>

Reviewed the dragover criteria — looks correctly scoped:

- `e.dataTransfer` null check: correct defensive coding (synthetic events can have null)
- `"Files"` discriminator: per HTML spec, only appears for OS file drags — cannot be faked by in-page `dataTransfer.setData()` (that API takes MIME types, not `"Files"`)
- No need to loosen: non-file OS drags (text from another app) don't match, but the `drop` handler only processes files anyway
- No need to tighten: already only matches genuine OS file drags

Criteria is correct as-is. Approve.

</details>

PR #5946, opened 2026-04-09
Fix `ui.on_exception` not catching exceptions from async event handlers

### LINKED ISSUE #5945: ui.on_exception is not catching from async handlers
### First Check

### Example Code

```python
from nicegui import ui


async def raise_async():
    raise RuntimeError("Error from async")


def raise_sync():
    raise RuntimeError("Error from sync")


ui.button("raise error in from async", on_click=raise_async)
ui.button("raise error in from sync", on_click=raise_sync)

ui.on_exception(lambda e: ui.notify(f"An error occurred: {e}", type="negative", timeout=0, close_button=True))
ui.run()
```

### Description

With _nicegui_ version 3.9.0 the above code works as expected and I see the `ui.notification` for both functions (also for the `async` function).

This is not working with _nicegui_ version 3.10.0. I see the traceback, but the `ui.notification` is not showing.

### NiceGUI Version

3.10.0

### Python Version

3.14.1

### Browser

Chrome

### Operating System

Linux

### Additional Context

_No response_

falkoschindler:
### Motivation

Since v3.10.0 (PR #5879), `ui.on_exception` no longer catches exceptions raised in async event handlers. Synchronous handlers still work correctly.

The root cause is that #5879 moved async handler awaiting into background tasks via `helpers.await_with_context`. When such a task raises, the exception surfaces through the background task's done callback (`_handle_exceptions`), which calls `core.app.handle_exception(e)` — but at that point the client/slot context is no longer on the context stack. So `app.handle_exception` never delegates to `client.handle_exception`, and per-client `ui.on_exception` handlers are skipped.

Fixes #5945.

### Implementation

Replaced `helpers.await_with_context` in `handle_event` with a new `_await_and_handle_in_context` coroutine that both awaits the result **and** catches exceptions **within** the slot context. This ensures `app.handle_exception` can find the client and dispatch to `ui.on_exception` handlers, restoring the pre-3.10.0 behavior.

### Progress

### DIFF
--- a/nicegui/events.py
+++ b/nicegui/events.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Callable, Iterator
+from collections.abc import Awaitable, Callable, Iterator
 from contextlib import nullcontext
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeAlias, TypeVar, cast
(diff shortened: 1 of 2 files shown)


evnchn (approved):
Works as intended. Nice. 

I think the reason is that `background_tasks.create_or_defer` swallows errors is non-trivial to understand and easy miss. Somewhere in our contributing guidelines we should name those which do swallow just in case.

PR #5951, opened 2026-04-13
Add `ui.keep_alive` to mount children eagerly regardless of visibility

### LINKED ISSUE #3033: AG Grids inside hidden tabs don't get mounted

### LINKED ISSUE #5839: Unmounted xterm write data lost.

falkoschindler:
### Motivation

Several elements with significant client-side state — `ui.xterm`, `ui.aggrid`, `ui.codemirror`, ECharts, ... — are unreachable when their host container has not been mounted yet. Writing to a `ui.xterm` inside a never-opened `ui.tab_panel` silently drops the data; calling `await grid.get_client_data()` on an `ui.aggrid` inside a closed `ui.dialog` silently returns `[]`. Existing workarounds are element-specific and don't generalise to elements whose state can't be reconstructed from props (xterm's scrollback, codemirror's edit history, ...).

Resolves:

- #5839 — `ui.xterm` writes to a not-yet-mounted terminal are lost
- #3033 — `ui.aggrid.get_client_data()` returns `[]` for grids inside an inactive `ui.tab_panel`

Possibly superseded:

- PR #5645 — the AG-Grid-specific phantom-clone workaround for #3033.

  **Proposal:** close in favour of `ui.keep_alive` for the common case. The phantom-clone is a memory optimisation (row data only, no live grid instance), so if anyone has a read-only-grid-in-1000-tabs scenario it could be re-opened as an optional fast path; otherwise the simpler unified mechanism wins.

- PR #5924 — the xterm-specific call-buffering workaround for #5839.

  **Proposal:** close in favour of `ui.keep_alive`. xterm.js holds state beyond the write log (cursor position, ANSI modes, scrollback dimensions) so replaying buffered writes onto a fresh terminal can desync that state; keeping the same component instance alive is the more honest fix.

In both cases the right follow-up — independent of `ui.keep_alive` — is to stop the *silent* failure when an unmounted element is interacted with (`aggrid.get_client_data()` returning `[]`, `xterm.write()` swallowing bytes). A loud warning that suggests `ui.keep_alive` would surface the issue at the right moment, instead of leaving users to debug it.

### Implementation

A new `ui.keep_alive` wraps its children so they are mounted eagerly into a hidden host at the page root, and teleported into the user's slot whenever that slot becomes visible. The same component instance is reused across show/hide cycles, so internal state survives — and method calls on the inner elements work the entire time, before the slot is ever revealed for the first time.

The element creates two NiceGUI elements per use:

- A **`_KeepAliveHost`** placed in `client.layout` (always-mounted q-layout). The user's children are reparented here via overrides of `__enter__` / `__exit__`, so they are alive for the lifetime of the client.
- A small **anchor** `<div>` placed at the original slot location. It has `display: contents` so it does not introduce its own layout box — children get sized by the user's actual container (a flex `q-card`, a tab panel's content area, ...).

The host renders its slot through a `<Teleport :to="target">` and switches `target` between the anchor's selector and a global hidden `#nicegui-keep-alive-cache` div. The signal between anchor and host stays entirely on the client (no extra round-trip to the server). When the keep_alive is cascade-deleted, `_handle_delete` cleans up the host.

This mirrors the existing `ui.dialog` "canary" trick (dialog lives in `client.layout`, sentinel element in the user's slot deletes it when collected); a follow-up could extract a shared `_HostedElement` base.

### Progress

#### Open questions before merge

- [ ] **Name**: `ui.keep_alive` overlaps Vue's `<KeepAlive>` and they do roughly opposite things (Vue's caches between unmount/remount; this prevents unmount in the first place). Worth deciding now since renaming after merge is breaking. Candidates: `ui.persistent`, `ui.always_mounted`, `ui.eager`.
- [x] **Version placeholder** in the docstring (`*Added in version 3.x.0*`) needs to be updated to the actual release.

### DIFF
--- a/nicegui/elements/keep_alive.py
+++ b/nicegui/elements/keep_alive.py
@@ -27,7 +27,7 @@ class KeepAlive(Element, component='keep_alive_anchor.js'):
         Note that this keeps the children alive for the full lifetime of the client, which costs memory.
         Only use it where eager mounting is actually required.
 
-        *Added in version 3.x.0*
+        *Added in version 3.11.0*
         """
         super().__init__()
         with context.client.layout:
--- a/nicegui/elements/keep_alive.py
+++ b/nicegui/elements/keep_alive.py
@@ -12,11 +12,18 @@ class KeepAlive(Element, component='keep_alive_anchor.js'):
         Wraps its children so they stay mounted in the DOM even when the surrounding container is currently not visible
         (e.g. an inactive ``ui.tab_panel``, a closed ``ui.dialog`` or ``ui.menu``, or a sub-page that is not currently routed to).
 
-        Method calls and events on the inner elements (such as writing to a not-yet-visible ``ui.xterm``
-        or reading data from an ``ui.aggrid`` whose tab has never been opened) are dispatched as usual.
+        This is useful for elements whose client-side state would otherwise be lost or unreachable before first display:
+        writing to a ``ui.xterm`` inside a never-opened tab, reading ``ui.aggrid.get_client_data()`` from a closed dialog,
+        or preserving edit history in a ``ui.codemirror`` across navigation.
+        Method calls and events on the inner elements are executed immediately on the live component instance —
+        they are never buffered or replayed.
 
         Internally the children are rendered into a hidden host at the page root
         and teleported to the wrapper's location whenever it becomes visible.
+        As a consequence, the server-side element tree parents the children to this host, not to the surrounding container:
+        ``descendants()`` on the apparent parent and scoped ``ui.element_filter`` will not traverse into the subtree.
+        Keep a direct reference to the inner elements if you need to reach them programmatically.
+
         Note that this keeps the children alive for the full lifetime of the client, which costs memory.
         Only use it where eager mounting is actually required.
 
--- a/nicegui/elements/keep_alive.py
+++ b/nicegui/elements/keep_alive.py
@@ -1,0 +1,52 @@
+from typing_extensions import Self
+
+from ..context import context
+from ..element import Element
+
+
+class KeepAlive(Element, component='keep_alive_anchor.js'):
+
+    def __init__(self) -> None:
+        """Keep Alive
+
+        Wraps its children so they stay mounted in the DOM even when the surrounding container is currently not visible
+        (e.g. an inactive ``ui.tab_panel``, a closed ``ui.dialog`` or ``ui.menu``, or a sub-page that is not currently routed to).
+
+        Method calls and events on the inner elements (such as writing to a not-yet-visible ``ui.xterm``
+        or reading data from an ``ui.aggrid`` whose tab has never been opened) are dispatched as usual.
+
+        Internally the children are rendered into a hidden host at the page root
+        and teleported to the wrapper's location whenever it becomes visible.
+        Note that this keeps the children alive for the full lifetime of the client, which costs memory.
+        Only use it where eager mounting is actually required.
+
+        *Added in version 3.x.0*
+        """
+        super().__init__()
+        with context.client.layout:
+            self._host = _KeepAliveHost(anchor=self)
+        self._props['host-id'] = self._host.id
+
+    def __enter__(self) -> Self:
+        # Children declared inside `with ui.keep_alive(): ...` belong to the stable host,
+        # not to the anchor that may be unmounted at any time.
+        self._host.__enter__()
+        return self
+
+    def __exit__(self, *exc) -> None:
+        self._host.__exit__(*exc)
+
+    def _handle_delete(self) -> None:
+        # NOTE: the host may already be queued for deletion if the entire client is being torn down,
+        # in which case its parent slot has already cleared it from its children list.
+        host_slot = self._host.parent_slot
+        if not self._host.is_deleted and host_slot is not None and self._host in host_slot.children:
+            self._host.delete()
+        super()._handle_delete()
+
+
+class _KeepAliveHost(Element, component='keep_alive_host.js'):
+
+    def __init__(self, *, anchor: Element) -> None:
+        super().__init__()
+        self._props['anchor-selector'] = f'#{anchor.html_id}'
--- a/nicegui/elements/keep_alive.py
+++ b/nicegui/elements/keep_alive.py
@@ -14,9 +14,9 @@ class KeepAlive(Element, component='keep_alive_anchor.js'):
 
         This is useful for elements whose client-side state would otherwise be lost or unreachable before first display:
         writing to a ``ui.xterm`` inside a never-opened tab, reading ``ui.aggrid.get_client_data()`` from a closed dialog,
-        or preserving edit history in a ``ui.codemirror`` across navigation.
-        Method calls and events on the inner elements are executed immediately on the live component instance —
-        they are never buffered or replayed.
+        or preserving edit history in a ``ui.codemirror`` when switching between tabs.
+        Method calls and events on the inner elements are executed immediately on the live component instance.
+        They are never buffered or replayed.
 
         Internally the children are rendered into a hidden host at the page root
         and teleported to the wrapper's location whenever it becomes visible.
@@ -26,6 +26,9 @@ class KeepAlive(Element, component='keep_alive_anchor.js'):
 
         Note that this keeps the children alive for the full lifetime of the client, which costs memory.
         Only use it where eager mounting is actually required.
+        Because state is tied to the client, ``ui.keep_alive`` cannot preserve anything across ``@ui.page`` navigation,
+        which creates a fresh client every time;
+        it only helps within a single page (inactive tabs, closed dialogs, unrouted sub-pages).
 
         *Added in version 3.11.0*
         """
--- a/nicegui/elements/keep_alive_anchor.js
+++ b/nicegui/elements/keep_alive_anchor.js
@@ -1,0 +1,12 @@
+export default {
+  template: `<div style="display: contents"></div>`, // make anchor transparent to layout for correct sizing of children
+  props: {
+    hostId: Number,
+  },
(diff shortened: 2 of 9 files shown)


evnchn (commented):
**Review by @evnchn with Claude Code (human-in-the-loop)**

Architecture is solid — the anchor/host split with client-side-only Teleport signaling is clean and consistent with the existing `ui.dialog` sentinel pattern. Headless Playwright tests confirmed `display: contents` works in flex and grid containers, xterm sizes correctly, and teleported content renders on reveal.

Two commits pushed with fixes found during review. Open items remain on naming, version placeholder, and manual browser testing.

<details>
<summary>Commits pushed</summary>

1. **`10740554` — Add `aria-hidden` to keep-alive cache div**
   The `#nicegui-keep-alive-cache` div uses `display: none` which already hides from assistive technology, but an explicit `aria-hidden="true"` is better practice.

2. **`2ba6f897` — Fix keep_alive documentation demos**
   - xterm demo: constrained to `cols=28, rows=9` (matching xterm docs convention) instead of unconstrained full-width
   - aggrid demo: added missing `ui.card()` wrapper in dialog (standard NiceGUI pattern), set `min-w-96` for column room, made name column `editable: True` to motivate *why* you'd call `get_client_data`, and added a line explaining the editable→read-back use case

</details>

<details>
<summary>Investigation that found no issue</summary>

- **Cache div cross-client leak**: Investigated whether `#nicegui-keep-alive-cache` (appended to `document.body` outside Vue) leaks across clients. Not an issue — each client is a separate browser tab with its own DOM, cleared on close/refresh.
- **Sub-page test fragility**: The element-count string comparison in `test_inside_sub_pages_cleans_up_on_navigation` was flagged as potentially brittle, but on reflection it's fit for purpose — the `keep_alive` is recreated each navigation so there's no persistent handle to assert against.

</details>

<details>
<summary>Open items</summary>

- **Naming** (`keep_alive` vs `eager` / `always_mounted` / `persistent`): Vue's `<KeepAlive>` caches between unmount/remount; this prevents unmount entirely — opposite semantics, same name. Decision deferred to last.
- **Version placeholder**: Docstring has `*Added in version 3.x.0*` — update when release target is confirmed.
- **Manual browser testing**: Headless Playwright covered layout and content, but the following need real eyeballs:
  1. Teleport flicker on first reveal (single-frame flash between cache and anchor?)
  2. xterm resize behavior after panel width change (ResizeObserver + teleport interaction)
  3. Scroll position survival in teleported scroll areas
  4. Keyboard focus / tab-order in teleported inputs
  5. Dialog backdrop + close behavior with keep_alive wrapper

</details>

evnchn:
Manual browser testing is good. I see no issue with using the name keep-alive because NiceGUI and Vue live in different layers per my Hamburger DOM analogy in NiceGUI off-topic Discord, attached in full. In fact, using the same name lets user know what is happening right away (though we may need to make that clear in the docs)

image

evnchn:
Hi Falko, I've been looking at the `#nicegui-keep-alive-cache` div and noticed it's currently created dynamically in `keep_alive_host.js` via `beforeMount`. Since this div is a shared singleton that any `keep_alive` instance relies on, it might be cleaner to bake it directly into `index.html` instead.

The benefits:
- The cache div is guaranteed to exist before any Vue component mounts, eliminating any potential timing dependency.
- Removes 8 lines of runtime `createElement` logic from `keep_alive_host.js`.
- The `CACHE_ID` constant can be inlined since there's no longer a need to coordinate creation and lookup.
- Makes the page structure more declarative — the cache div is an intentional part of the template, not a side effect of the first `keep_alive` mounting.

I've prototyped this on a branch if you'd like to take a look: [`evnchn/nicegui:keep-alive-cache-in-template`](https://github.com/evnchn/nicegui/compare/keep-alive-cache-in-template). All 7 `test_keep_alive` tests pass. Feel free to cherry-pick if you think it fits.

falkoschindler:
@evnchn Great idea! I cherry-picked your solution.

Last but not least, the name: I lean towards keeping `ui.keep_alive`, to be honest.

- The name describes the **observable contract** (children stay alive and reachable), which is what users actually care about. How it's implemented — caching vs. teleporting — is an internal detail.
- NiceGUI's audience typically isn't reaching into Vue internals; the set of users who would hit the semantic-overlap confusion is small.
- Both primitives converge on the same user-visible promise: "state survives visibility changes." Two names for the same outcome would be the larger papercut.
- Alternatives considered (`ui.persistent`, `ui.always_mounted`, `ui.eager`) are either vaguer or equally overloaded.

nicegui/elements/keep_alive.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self) -> None:
+        """Keep Alive
+
+        Wraps its children so they stay mounted in the DOM even when the surrounding container is currently not visible
+        (e.g. an inactive ``ui.tab_panel``, a closed ``ui.dialog`` or ``ui.menu``, or a sub-page that is not currently routed to).
+
+        This is useful for elements whose client-side state would otherwise be lost or unreachable before first display:
+        writing to a ``ui.xterm`` inside a never-opened tab, reading ``ui.aggrid.get_client_data()`` from a closed dialog,
+        or preserving edit history in a ``ui.codemirror`` across navigation.
```
evnchn: We should make the line exceptionally clear that it does NOT benefit `@ui.page`-based routing.

nicegui/elements/keep_alive.py
```diff
@@ -0,0 +1,59 @@
...
+        Method calls and events on the inner elements are executed immediately on the live component instance —
```
evnchn: We already have 4 em-dashes in the codebase, 3 in website (reasonable), and 1 in a comment in native module. 

If we include this, it will be the first occurrence in the main codebase, which while fine technically:

1. Signals LLM creation
2. No longer ASCII-compatible

I'd say a comma can do in this instance

evnchn (dismissed):
Staring at the docstring because the code is already proven-solid long ago. @falkoschindler do you want to take either of the pending suggestions? If not then take the approve. Thanks!

PR #5953, opened 2026-04-13
Bump pytest to v9 to fix Dependabot alert 244

falkoschindler:
### Motivation

Dependabot alert 244 reports a moderate-severity vulnerability in pytest <9.0.3 (insecure `/tmp/pytest-of-{user}` tmpdir handling on UNIX). The fix is only available in pytest 9.0.3+, which is outside the previous `<9` constraint.

### Implementation

- Bump the dev-dependency constraint from `pytest>=8.2.2,<9` to `pytest>=9.0.3,<10`.
- Re-lock `uv.lock` (resolves to pytest 9.0.3; existing pytest plugins pytest-asyncio, pytest-selenium, pytest-order, pytest-watcher all stay compatible without bumps).
- Local smoke tests against `tests/test_label.py`, `tests/test_xterm.py`, `tests/test_user_simulation*.py` and `tests/test_vbuild.py` pass.
- No code changes were required — the codebase uses no APIs that pytest 9 removed (no `tmpdir`, no `pytest.warns(None)`, no legacy `py.path` usage).

### Progress

- [x] This PR does not address a security issue. (The vulnerability is in a dev dependency, not in NiceGUI itself.)


falkoschindler:
This one is mostly mechanical. But instead of committing directly to main - like for other Dependabot alerts - I wanted to run CI and have a ticket show up in the release notes mentioning the major version bump.

evnchn (approved):
Since the CI passes, we can merge this by empirical evidence.

PR #5957, opened 2026-04-14
Add ui.lazy: defer rendering of children past the first paint

evnchn:
### Motivation

Complex pages that render many subtrees at once block first paint, even when most of those subtrees are off-screen or otherwise non-critical. A lightweight escape hatch to defer "heavy-but-not-urgent" subtrees past the first paint frame would let authors prioritize what the user sees first.

Internally, NiceGUI's own documentation site uses a similar deferred-render pattern for code demos. This PR exposes an equivalent mechanism as a public element so user apps can use it too.

### Implementation

A minimal `ui.lazy` element that wraps its children in `<slot v-if="shouldRender">`. On mount, it flips `shouldRender` to `true` via `nextTick` + two `requestAnimationFrame` calls — that is the standard pattern for "wait until after the browser has laid out and painted the current frame."

```js
mounted() {
  this.$nextTick(() => {
    requestAnimationFrame(() => {
      requestAnimationFrame(() => { this.shouldRender = true; });
    });
  });
}
```

Files:
- `nicegui/elements/lazy.py` — 12-line `Lazy(Element, component='lazy.js')` subclass.
- `nicegui/elements/lazy.js` — Vue component with the defer logic above.
- `nicegui/ui.py` — export in the lazy-import dict, `__all__`, and `TYPE_CHECKING` block.

**Relationship to keep-alive**: this is the complementary opposite of a keep-alive-style mechanism. `ui.lazy` defers *building* the subtree until after first paint; a hypothetical `ui.keep_alive` would *retain* a built subtree across unmount. Both could coexist.

**Note on naming / scope**: this is deliberately the simplest implementation — no `IntersectionObserver`, no priority, no "unmount if hidden" logic. Happy to extend if maintainers prefer a viewport-aware variant, but I'd argue the 2-RAF version covers the dominant use case (tab panels, offscreen-but-eventually-visible content) with near-zero runtime cost.

### Progress

- [x] If this PR addresses a security issue, it has been coordinated via the security advisory process. (N/A)
- [ ] Pytests — draft PR is inviting API feedback first; happy to add tests if the API shape is accepted.
- [ ] Documentation page — will add after API is blessed.

### Tracking

Tracking fork PR: [evnchn/nicegui#107](https://github.com/evnchn/nicegui/pull/107)

### DIFF
--- a/nicegui/elements/lazy.js
+++ b/nicegui/elements/lazy.js
@@ -1,0 +1,17 @@
+export default {
+  template: `<div><slot v-if="shouldRender"></slot></div>`,
+  data() {
+    return {
+      shouldRender: false,
+    };
+  },
+  mounted() {
+    this.$nextTick(() => {
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          this.shouldRender = true;
+        });
+      });
+    });
+  },
+};
--- a/nicegui/elements/lazy.py
+++ b/nicegui/elements/lazy.py
@@ -1,0 +1,12 @@
+from ..element import Element
+
+
+class Lazy(Element, component='lazy.js'):
+    """Lazy element that defers rendering of its children until mounted.
+
+    Child elements are only rendered after the component is mounted and the next Vue tick,
+    which can improve initial page load performance for complex UIs.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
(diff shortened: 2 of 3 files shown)


evnchn:
Closing — going through fork review first to keep upstream clean. Will re-open after human review + squash.

PR #5958, opened 2026-04-14
Enhance xterm example: configurable command, PTY resize, lifecycle hardening

evnchn:
### Motivation

The `examples/xterm` demo was a minimal shell-only demonstration. Two pain points become visible as soon as someone tries to use it for anything real:

1. The command is hardcoded to `/bin/bash`. If you want to show a TUI (`htop`, Textual, `vim`), you can only get there by editing the source.
2. There is a race on startup: the PTY is forked before the client has connected, so the first burst of output is silently dropped. And on exit, `on_delete` does not check whether the PTY was ever opened.

### Implementation

Single-file change to `examples/xterm/main.py`:

- **Configurable command** via `sys.argv[1]`, split with `shlex.split` and exec'd with `os.execvp` (argv list, no shell interpolation — the CLI arg is the server operator's own input, not client input, and there is no shell invoked). Default is still `/bin/bash`.
- **PTY resize** continues to use the typed `terminal.on_resize` handler introduced in #5858, with an added guard so a resize event arriving before `pty.fork()` is a no-op instead of a crash.
- **Startup race fix**: `await ui.context.client.connected()` before `pty.fork()`. Ensures the terminal has mounted and is ready to receive data before the process starts emitting.
- **Teardown hardening**: `on_delete` guards with `pty_fd >= 0`; `pty_to_terminal` guards `terminal.write` with `is_deleted` to avoid writing into a torn-down element.
- **Viewport sizing**: `.nicegui-content { padding: 0 }` + `w-screen h-screen` so the terminal fills the window — otherwise TUIs render in a tiny strip.
- **Child environment**: `TERM=xterm-256color` + `COLORTERM=truecolor`, so color-aware TUIs render correctly.

### Not a security fix

The existing hardcoded `os.execv('/bin/bash', ('bash',))` had no user input and no shell — there is no pre-existing vulnerability. The new CLI-arg path uses `shlex.split` + `execvp` which takes an argv list and does not spawn a shell, so there is no shell-interpolation surface even with the new feature.

### Progress

- [x] If this PR addresses a security issue, it has been coordinated via the security advisory process. (N/A — not a security fix, see above.)
- [x] Pytests are not necessary — `examples/*` is not covered by the test suite.
- [ ] Documentation — example code is self-documenting via the module docstring and the `WARNING` block; no external docs page changes needed.

### Tracking

Tracking fork PR: [evnchn/nicegui#96](https://github.com/evnchn/nicegui/pull/96)

### DIFF
--- a/examples/xterm/main.py
+++ b/examples/xterm/main.py
@@ -1,27 +1,63 @@
 #!/usr/bin/env python3
-"""Example for connecting a `Xterm` element with a `Bash` process running in a pty.
+"""Example for connecting a `Xterm` element with a process running in a pty.
 
-WARNING: This example gives the clients full access to the server through Bash. Use with caution!
+Pass a command as argument to run something other than bash, e.g.::
+
+    python main.py top
+    python main.py "python3 -m textual_paint"
+
+WARNING: This example gives the clients full access to the server through a shell process. Use with caution!
 """
 import fcntl
 import os
 import pty
+import shlex
 import signal
 import struct
+import sys
 import termios
 from functools import partial
 
 from nicegui import core, events, ui
 
+command = shlex.split(sys.argv[1]) if len(sys.argv) > 1 else ['/bin/bash']
+
 
 @ui.page('/')
-def _page():
-    terminal = ui.xterm().classes('w-full')
+async def _page():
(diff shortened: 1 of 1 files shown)


evnchn:
Closing — going through fork review first to keep upstream clean. Will re-open after human review + squash.

PR #5959, opened 2026-04-14
Improve xterm example: configurable command, PTY resize, viewport sizing, lifecycle hardening

evnchn:
### Motivation

The `examples/xterm` demo was bash-only with a startup race: the PTY forked before the browser connected, dropping the first burst of output; `on_delete` also did not guard against a PTY that had never been opened.

### Implementation

Single-file change to `examples/xterm/main.py`:

- **Configurable command** — `sys.argv[1:]` (default `['/bin/bash']`), exec'd with `os.execvp(command[0], command)` (argv list, no shell).
- **Startup race fix** — `await ui.context.client.connected()` before `pty.fork()`.
- **Teardown hardening** — `on_delete` guards with `pty_fd >= 0` and calls `core.loop.remove_reader(pty_fd)` before `os.close`; `pty_to_terminal` also handles the empty-read EOF case so a silent EOF does not spin the event loop.
- **Resize** — keeps the typed `terminal.on_resize` handler from #5858 with a `pty_fd < 0` early-return for pre-fork events, and drops the noisy `Terminal resized to …` print.
- **Viewport sizing** — `.nicegui-content { padding: 0 }` + `w-screen h-screen` so TUIs fill the window instead of a tiny strip.
- **Child environment** — `TERM=xterm-256color` + `COLORTERM=truecolor` so `nano` / `htop` render in full color.

### Progress

- [x] Pytests are not necessary — `examples/*` is not covered by the test suite.
- [x] Documentation is not necessary — the example's module docstring is self-documenting.

---

Tracking on fork: evnchn/nicegui#96. Fork CI is green (pre-commit, mypy, pylint, quick-test all passed on the same commit).

### DIFF
--- a/examples/xterm/main.py
+++ b/examples/xterm/main.py
@@ -52,6 +52,10 @@ async def _page():
         core.loop.remove_reader(pty_fd)  # unregister before closing so a reused FD can't trip the stale callback
         os.close(pty_fd)
         os.kill(pty_pid, signal.SIGKILL)
+        try:
+            os.waitpid(pty_pid, 0)  # reap the child so it does not linger as a zombie
+        except ChildProcessError:
+            pass
         print('Terminal closed')
 
     await ui.context.client.connected()
--- a/examples/xterm/main.py
+++ b/examples/xterm/main.py
@@ -59,7 +59,11 @@ async def _page():
     if pty_pid == pty.CHILD:
         os.environ['TERM'] = 'xterm-256color'
         os.environ['COLORTERM'] = 'truecolor'
-        os.execvp(command[0], command)  # replace the child process with the requested command
+        try:
+            os.execvp(command[0], command)  # replace the child process with the requested command
+        except OSError as e:
+            print(f'Failed to exec {command}: {e}', flush=True)
+            os._exit(1)  # abort the child so it does not fall through into the parent's code path
     print('Terminal opened')
 
     @partial(core.loop.add_reader, pty_fd)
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Five small commits on `xterm-configurable-command`, all touching only [examples/xterm/main.py](examples/xterm/main.py):

1. **`196e2bfc` — Clarify docstring.** "Pass a command (and any arguments, which are forwarded verbatim)" — spells out that extra argv tokens go to the child process, not argparse.
2. **`9586a850` — Early-return in `kill_process`.** Flipped `if pty_fd >= 0:` to `if pty_fd < 0: return` so both pre-fork guards (`resize_terminal` and `kill_process`) read the same way.
3. **`eca366a0` — Guard failed `execvp`.** Wraps `os.execvp` in `try/except OSError` and calls `os._exit(1)` on failure so a bad command name (e.g. `python main.py nope`) doesn't leave the forked child running parent code with a shared event loop.
4. **`46d5c801` — Reap the pty child.** Adds `os.waitpid(pty_pid, 0)` after `SIGKILL` (with `ChildProcessError` fallback). Prevents zombie processes from accumulating when tabs are closed. Effectively non-blocking since `SIGKILL` is unmaskable.
5. **`9a1e2ed1` — Use `client.content` instead of `ui.query`.** Direct reference to the page content element is cleaner than a CSS-selector lookup.

Net effect: same user-visible behavior, slightly more defensive teardown and clearer code. No public API touched.

Ready to merge if you agree, @evnchn! 🚀

PR #5960, opened 2026-04-14
Randomize test ports/dirs for parallel pytest sessions

evnchn:
### Motivation

Two `pytest` invocations cannot run simultaneously today: `Screen.PORT` is hardcoded to `3392`, `Screen.SCREENSHOT_DIR` is a fixed `'screenshots'`, and NiceGUI's file-based storage resolves to a fixed `.nicegui/` dir. Running a second process collides on port bind, on screenshot filenames, and on persistent storage files. This PR makes those session-unique so two (or more) pytest runs can share a working directory without stomping each other — including mid-flight overlap (start session B while session A is already running a long test).

### Implementation

- **`nicegui/helpers/network.py`**
  - New public helper `find_free_port()` binds an ephemeral `AF_INET / SOCK_STREAM` to `('0.0.0.0', 0)`. It binds on `0.0.0.0` deliberately — `ui.run` defaults to binding on all interfaces, so probing the same namespace avoids a port-free-on-loopback-but-busy-on-another-interface race. The docstring notes the inherent TOCTOU caveat (another process may grab the port between this call and `ui.run` binding).

- **`nicegui/testing/filelock.py`** (new)
  - Small cross-platform `FileLock` class: `fcntl.flock` on POSIX, `msvcrt.locking` on Windows. Exclusive, non-blocking, advisory. The OS releases the lock automatically on process death — clean exit *or* SIGKILL/OOM. Used as a process-liveness probe (see screenshot pruning below).

- **`nicegui/testing/screen_plugin.py`**
  - `pytest_configure` sets `Screen.PORT = helpers.find_free_port()`, `Screen.SCREENSHOT_DIR = Path('screenshots') / str(os.getpid())`, and `DOWNLOAD_DIR = Path(tempfile.mkdtemp(prefix='nicegui-test-download-'))`, with an `atexit.register` as a safety net for aborted sessions.
  - The `screen` fixture creates `DOWNLOAD_DIR` before `yield` and rmtrees it in `finally` (setup/teardown symmetry — Chrome's session-scoped `download.default_directory` always points at a directory that actually exists).
  - `nicegui_remove_all_screenshots` acquires an exclusive lock on `screenshots/<pid>/.lock` for the session's lifetime, then prunes sibling PID-dirs whose lock it can acquire (= owner is gone). This is immune to PID reuse and handles abnormal exits. The previous PID-aliveness check with its Windows `ctypes` block is gone.

- **`nicegui/testing/general_fixtures.py`**
  - `pytest_configure` creates a session-unique storage dir via `tempfile.mkdtemp(prefix='nicegui-test-storage-')`, assigns it to `Storage.path`, and rebuilds `app.storage = Storage()` once so its internal `FilePersistentDict` picks up the new path (the instance constructed at `App.__init__` captured the default `.nicegui/`).
  - `Storage.path` itself doubles as the sentinel — `None` means "not configured." `pytest_unconfigure` removes the temp storage dir and resets the sentinel.

- **`nicegui/testing/plugin.py` / `user_plugin.py`**: re-export the new `pytest_unconfigure` hook; `plugin.py` exports `screen_plugin`'s `pytest_configure` (which delegates to `general_fixtures`' before applying the Screen-specific setup).

- **`tests/test_storage.py`**: replace `Path('.nicegui')` literals with `Storage.path`, and the hardcoded `http://localhost:3392/status` with `http://localhost:{Screen.PORT}/status`.

- **`website/documentation/content/screen_documentation.py`**: update the `PORT` and `SCREENSHOT_DIR` default notes to "automatically determined free port" and `./screenshots/<pid>`.

### Empirical validation

Two parallel pytest invocations from the same worktree, both using the Screen fixture:

```
$ uv run pytest tests/test_button.py -v  # 4/4 PASSED in 4.10s
$ uv run pytest tests/test_chip.py   -v  # 2/2 PASSED in 3.99s
```

Both picked different free ports, wrote to different `screenshots/<pid>` directories, and got their own `tempfile.mkdtemp` download dirs. No `address in use`, no errors.

### Notes on observable defaults

Three observable defaults change. None are hard breaks — the idiomatic way to use each surface was always indirect — but worth calling out:

- **`Screen.PORT`** was `3392` and is now a random free port per process. Tests that hardcoded `3392` (e.g. `http://localhost:3392/...`) need to read `Screen.PORT`, which has always been the documented-public way. This PR applies that migration to NiceGUI's own `test_storage.py`.
- **`Screen.SCREENSHOT_DIR`** was `'screenshots'` and is now `'screenshots/<pid>'`. Same shape as the existing `.failed.png` convention — a subpath under the screenshots root, transparent to anyone who reads screenshots via `Screen.SCREENSHOT_DIR` rather than hardcoding the literal.
- **`Storage.path`** (when the testing plugin is loaded) now points to a session-unique tempdir instead of `.nicegui/`. Reading `.nicegui/` directly in test code was never a supported pattern — the `Storage.path` attribute is the contract. NiceGUI's own `test_storage.py` is migrated to match.

### Progress

- [x] Pytests have been updated (affected `tests/test_storage.py` exercises storage + port paths).
- [x] Documentation has been updated (`screen_documentation.py`).
- [x] No breaking changes to the public API — observable default values move, but the documented accessors (`Screen.PORT`, `Screen.SCREENSHOT_DIR`, `Storage.path`) are unchanged. Migration notes above.


falkoschindler (commented):
Thanks for the pull request, @evnchn! Being able to run `uv run pytest -n auto` after installing pytest-xdist feels insanely powerful: running all 825 tests in 72 seconds on 14 cores is a crazy speedup!

There are a few review findings that need to be addressed though:

1. **Storage path setup: env-var-at-import is the wrong surface** — Two problems with one cause:
   - `os.environ['NICEGUI_STORAGE_PATH']` is overwritten unconditionally at module import. If a user intentionally sets it (e.g. to inspect storage after a failing test), the plugin silently replaces their value _and_ schedules that dir for `shutil.rmtree` in `pytest_unconfigure` — surprising, and in the wrong setup destructive.
   - `pytest_configure` _also_ reassigns `Storage.path` "as a belt-and-suspenders fallback", leaving two sources of truth and a hedge that suggests the author wasn't fully sure the env-var path was sufficient.

   Both go away if the plugin just assigns `Storage.path = Path(tempfile.mkdtemp(...)).resolve()` in `pytest_configure` — no env-var clobber, no belt-and-suspenders, no import-order constraint. The env var is the public configuration surface; the test plugin doesn't need it.

2. **`DOWNLOAD_DIR` has no session-level cleanup** — `tempfile.mkdtemp` is called at module import, but the only cleanup lives inside the per-test `screen` fixture's `finally`. If pytest collects `screen_plugin` but never runs a `screen`-using test (e.g. `pytest --collect-only`, or a file that only uses `user`), the tempdir leaks into `$TMPDIR` forever.

3. **`screenshots/` grows unboundedly across runs** — Before this PR, `nicegui_remove_all_screenshots` wiped the shared `screenshots/` dir at session start — bounded. After this PR, `SCREENSHOT_DIR = screenshots/<pid>` and the same fixture only globs `*.png` inside the _current_ PID's subdir. Old PID subdirs from previous sessions are never touched, so `screenshots/` accumulates one subtree per pytest run, forever.

4. **`app.storage = Storage()` in `nicegui_reset_globals` leaves a stale shutdown hook** — app.py does `self.on_shutdown(self.storage.on_shutdown)` at construction. Replacing `app.storage` leaves the shutdown handler bound to the old instance. Harmless for tests today (shutdowns aren't asserted), but it silently diverges from the production invariant — if a future test asserts storage-shutdown behavior, it'll be checking the zombie. Worth a one-line comment explaining why the reassignment is safe in the test harness, or better: reset state on the existing `app.storage` in-place rather than swapping instances.

5. **`_find_free_port` has an unavoidable TOCTOU gap — call it out in the docstring**

   The docstring notes the `0.0.0.0` rationale but not that another process can grab the port between `s.close()` and `ui.run`'s bind. Worth a one-liner ("best-effort — not collision-proof under racing port grabs") so future readers don't assume this is a hard guarantee. A retry wrapper in the `screen` fixture that re-picks on `OSError: address in use` would turn best-effort into effectively-never-fails, but that's follow-up territory.

evnchn:
Moving to 3.12 — coordinated with Falko, this needs more work and shouldn't block the 3.11 release.

evnchn:
Addressed all 5 review items as separate commits on top of the original (rebased onto current branch tip after the `find_free_port` extraction):

- **5 (TOCTOU docstring)** — moved to `helpers.find_free_port` since the function now lives there. Commit `f8166e62`.
- **3 (screenshots growth)** — `nicegui_remove_all_screenshots` now prunes `screenshots/<pid>/` subdirs whose PID is no longer alive, while preserving subdirs of concurrent live runs. Liveness probe goes through a small `_pid_alive` helper: `os.kill(pid, 0)` on POSIX, `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` on Windows. Commit `a0e77d00`.
- **2 (DOWNLOAD_DIR leak)** — `atexit.register(shutil.rmtree, DOWNLOAD_DIR, ignore_errors=True)` covers the no-screen-fixture-ever-runs case. Commit `8045da5d`.
- **1 (env-var clobber)** — dropped the `os.environ['NICEGUI_STORAGE_PATH']` write at module import and the import-position dance. `pytest_configure` now just assigns `Storage.path = Path(_nicegui_storage_dir).resolve()` — single source of truth, no env-var mutation. Commit `d935cfcf`.
- **4 (stale shutdown hook)** — moved the `app.storage = Storage()` rebuild from `nicegui_reset_globals` into `pytest_configure` (one-time, before any `app.start()`), and removed both per-test reassignments from `general.py`. Per-test isolation still provided by `app.reset() → self.storage.clear()`. Commit `a48d4375`.

**Verified locally**: ruff + mypy clean, `tests/test_storage.py` 25/25 PASSED, `tests/test_download.py` 3/3 PASSED, `pytest --collect-only` works, two parallel pytest sessions on `test_storage.py` + `test_button.py + test_chip.py` both green concurrently.

evnchn:
@falkoschindler hardly sounds safe to mess with the pytest fixture which is our line-of-defense to catch issues in the end of 3.12 release. Shall we move this to 3.13 so that when we review 3.13's PRs we can test this PR in action and hopefully do not end up with a 3.12.1?

evnchn:
You know what, I know @falkoschindler enough that he's unlikely to disagree, so I'm reassigning the milestones. . @rodja can you ship `main` to https://nicegui.io after https://github.com/zauberzeug/nicegui/pull/6043 merged? I weighted and think that having a docs in the future is less of a concern than not catching deadly bugs in 3.12.

falkoschindler:
Yes, I agree. Actually I just finished another round of AI-assisted review, but haven't finished reading the code myself. Maybe you can have a look into the following two findings meanwhile. Anyway, postponing to early 3.13 is a good idea!

---

1. **Windows `ctypes` calls lack `argtypes` / `restype`** (nicegui/testing/screen_plugin.py, lines 64-81) — `OpenProcess` returns a `HANDLE` (pointer-sized). Without `kernel32.OpenProcess.restype = ctypes.c_void_p` (and matching `argtypes`), Python defaults to `c_int`, which truncates 64-bit handle values on Win64. The truncated handle is unlikely to be valid for `CloseHandle`, leaking process handles, and in edge cases `OpenProcess` could be misread as failing on a live process — directories of live concurrent runs would then get pruned by another worker. Cleanup is best-effort, so this is not a correctness blocker, but the right-shape fix is small:

   ```python
   kernel32.OpenProcess.argtypes = [wintypes.DWORD, wintypes.BOOL, wintypes.DWORD]
   kernel32.OpenProcess.restype = wintypes.HANDLE
   kernel32.CloseHandle.argtypes = [wintypes.HANDLE]
   kernel32.CloseHandle.restype = wintypes.BOOL
   ```

2. **Module-import side effects in `general_fixtures.py`** (nicegui/testing/general_fixtures.py, line 11) — `_nicegui_storage_dir = tempfile.mkdtemp(...)` runs at import time. If the testing plugin is imported but pytest never actually runs (e.g. tooling that imports `nicegui.testing` for type inspection), the tempdir is leaked. Moving the `mkdtemp` into `pytest_configure` would keep the lifecycle paired with the matching `pytest_unconfigure` cleanup. Same shape applies to `screen_plugin.py:25-28` (port / screenshot dir / download dir all happen at import). `atexit.register` partially covers `DOWNLOAD_DIR`, so this is mostly a hygiene point — worth either consolidating into `pytest_configure` or documenting why the import-time location is required.

---

Regarding the website deployment: You're right. And I don't think the 3.12 documentation lies in any way with respect to 3.11. The only two API changes that don't apply are clearly marked with "Since version 3.12". I'm deploying it now.

falkoschindler (dismissed):
Thanks again, @evnchn! I pushed three follow-up commits on top of the branch to tidy a few things up. Sharing here so the diff is no surprise:

**[`0c1c8dbb0`](https://github.com/zauberzeug/nicegui/pull/5960/commits/0c1c8dbb0) — addressing another AI review**

- Moved `DOWNLOAD_DIR` creation into the `screen` fixture's setup and rmtree into `finally`, so Chrome's session-scoped `download.default_directory` always points at a directory that actually exists.
- Cleaned the inline atexit comment and added a NOTE on the PID-reuse caveat in `_is_alive`.
- Doc tweak: `./screenshots/<pid>` rather than `<cwd>/...` for clarity.

**[`8041ea141`](https://github.com/zauberzeug/nicegui/pull/5960/commits/8041ea141) — simpler storage sentinel & screenshot prune loop**

- Dropped the `_nicegui_storage_dir` module variable; `Storage.path` itself doubles as the sentinel (`None` = unconfigured), so `pytest_configure` and `pytest_unconfigure` read very flat.
- Hoisted the `app` / `Storage` imports out of `pytest_configure` — they were vestigial from the old `NICEGUI_STORAGE_PATH` ordering trick.
- Compacted `nicegui_remove_all_screenshots`: single condition with `path.name.isdigit()` filter, no try/except.

**[`f26304f9e`](https://github.com/zauberzeug/nicegui/pull/5960/commits/f26304f9e) — `FileLock` liveness probe**

- New `nicegui/testing/filelock.py` (`fcntl` on POSIX, `msvcrt` on Windows) — exclusive advisory file lock that the OS auto-releases on process death (including SIGKILL/OOM).
- Replaced `_is_alive(pid)` with: each session holds an exclusive lock on `screenshots/<pid>/.lock` for its lifetime; pruning a sibling dir is conditional on _successfully acquiring_ its `.lock`, which only succeeds when no live process holds it.
- Properties: immune to PID reuse, handles abnormal exits, mid-flight overlap works, no ~25-line `ctypes` block in `screen_plugin.py`. Cost: a leftover `.lock` file inside each PID-dir (harmless — next probe acquires it and rmtrees the whole dir).

Approving. Nice PR. 👏🏻

falkoschindler (approved):

falkoschindler:
I just reverted this PR on main to resolve the regression #6061. This allows us to find and merge a solution early in 3.13 instead of patching and releasing it quickly in 3.12.1.

PR #5795, opened 2026-02-17
Accessibility enhancement on styles

evnchn:
### Motivation

As mentioned in https://github.com/zauberzeug/nicegui/discussions/5794, we can do better in accessibility. 

This PR focuses on the styles. 

### Implementation

Manually-written because I cannot trust an AI on human accessibility. 

The PR comments should be descriptive of the changes. 

The policy is: Implement the change in library-level if possible, if not then at documentation-website-level. 

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] If this PR addresses a security issue, it has been coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process.
- [x] Pytests are not necessary.
- [ ] We need a new documentation section for the new accessibility stuff, such as `ui.colors(..., at_rule='@media (prefers-contrast: more)')`, and general guidance, especially the `forced-colors:` variant. 
  - Ref: https://github.com/zauberzeug/nicegui/pull/5790 as well

### Final notes

Happy Lunar New Year everyone!

PR #5964, opened 2026-04-17
Add scene primitives, TransformControls, clipping planes, ViewHelper, and hover support

Jepson2k:
### Motivation

These features were developed while building a web UI for controlling a 6-DOF robotic arm. Each addition traces to a concrete need in that host application:

- **TransformControls** — manipulating TCP targets and joint origins interactively
- **Clipping planes** — seeing inside articulated meshes (hiding geometry in front of a joint of interest)
- **ViewHelper** — orientation reference in a free-camera workcell view
- **Hover feedback** — selectable joints and click targets in a dense scene
- **Polyline, Lathe, ArrowHelper, PolarGridHelper** — motion paths, revolved parts, force/velocity indicators, and workcell layout
- **Bundled bug fixes** — material updates, STL wireframe, a TransformControls regression — hit during development

The PR is broad because the features were developed together. Happy to split along the lines above if preferred — the natural seam is (1) primitives, (2) TransformControls + hover, (3) clipping + ViewHelper + miscellaneous.

### Implementation

**New primitives** (`scene_objects.py`, `scene.js`):
- `Polyline` — line with per-vertex colors and GPU dashing
- `Lathe` — rotational solid geometry
- `ArrowHelper` — wraps Three.js `ArrowHelper` with radial segments, color, linewidth
- `PolarGridHelper` — radial grid with configurable sectors/rings

**Interactive features** (`scene.py`, `scene.js`):
- `enable_transform_controls` / `disable_transform_controls` — modes (translate/rotate/scale), world/local space, rotation snap, axis colors; emits `on_transform`, `on_transform_start`, `on_transform_end`; reference-counted orbit-control disable during drag
- `set_clipping_planes` / `clear_clipping_planes` + `SceneClipPlane` dataclass for per-object proximity clipping
- `set_axes_inset` / `set_axes_labels` — Three.js `ViewHelper` overlay
- `set_orbit_enabled` toggle
- Hover feedback for `_hoverable` objects: on hover, each mesh descendant gets a back-face glow clone that shares its geometry and follows its world transform (runs entirely in the browser, no Python roundtrip); appearance via `hover_color` / `hover_opacity` / `hover_scale`
- `SceneGroundPoint` ground plane configurable via `ground_axis` (`'x'`/`'y'`/`'z'`) and `ground_offset`
- `raycaster_threshold` to cap hits from thin lines/points and avoid large websocket payloads

**New Object3D method** (`scene_object3d.py`):
- `rotate_euler(r_x, r_y, r_z, order='XYZ')` — Euler rotation with explicit order, applied via `THREE.Euler`

**Bundled bug fixes** (`scene.js`, `test_scene.py`):
- Material changes after creation didn't trigger `needsUpdate`, so `.material(...)` calls after the first one were ignored
- `wireframe=True` on STL meshes produced a filled mesh because the flag was set before the async STL load completed
- TransformControls broke silently when the underlying object was a `Group` (regression from a recent Three.js upstream change)
- `test_arrow_helper` checked a nonexistent `isArrowHelper` flag instead of the actual `.type === 'ArrowHelper'`
- `test_transform_controls_enable_disable` passed an object's name where the JS method expected a UUID

**New event types** (`events.py`):
- `SceneTransformEventArguments` — position, rotation, scale, mode, state
- `SceneGroundPoint` — ray intersection on the configured ground plane
- `SceneClipPlane` — plane equation `(nx, ny, nz, d)`
- `SceneClickEventArguments` gains `ground_point` plus screen/client/offset coordinates (for overlay positioning)

**Dependency bump** (`nicegui/elements/scene/package.json`):
- Bumped `three` from `^0.180.0` to `^0.184.0`. The existing `set_axes_inset` code uses `ViewHelper.location.left/right/top/bottom` for anchor/margin positioning; that property was added to three.js's `ViewHelper` in r184.

### Progress


evnchn:
Hello @Jepson2k 

I just click run the workflow but you can enable it for your fork to save you the wait.

One thing which I see immediately: the diff is big with code styling changes. You may want to ensure full compliance with CONTRIBUTING.md

evnchn:
To elaborate on the style changes:

image

They can be gone without affecting functionality. Removing all of such changes before sending changes upstream is generally good practice.

falkoschindler:
Thanks @Jepson2k! Since the PR is still in draft, just let us know when you'd like a full review. In the meantime, test collection is failing on CI ([run](https://github.com/zauberzeug/nicegui/actions/runs/24592574705/job/71931754823)) — worth a look:

**`tests/test_scene.py` collection error**

```
In test_polyline: function uses no argument 'control_type'
```

At [tests/test_scene.py:440](tests/test_scene.py#L440) `test_polyline` is decorated with

```python
@pytest.mark.parametrize('control_type,constructor', [('map', 'MapControls'), ('trackball', 'TrackballControls')])
def test_polyline(screen: Screen):
```

but doesn't accept those parameters — looks like the decorator from `test_custom_controls` was duplicated by accident. Removing the `@pytest.mark.parametrize` line should fix collection so the rest of the suite can actually run.

Jepson2k:
@evnchn @falkoschindler Its now ready for review. I was keeping it a draft while I was still working through making some of the features more generalizable rather than specific to my app and cleaning up unnecessary things like the styling changes. As for the force pushes, yes I generally agree. For the workflows, I'm not quite sure what you mean by "enable" them but I was able to manually trigger them on the forked repo.

evnchn:
On "enable": GitHub used to put the button to enable automatic workflow runs inside the actions tab of people's fork and it's my instinct to get people to enable it so that their PRs get auto CI (it's free btw, since it's a public repo)

GitHub loves to shift their UI/UX around, and so, short of me creating a throwaway account to test things out, I don't know if that button still exists. 

Anyways **CI on PRs** is good and we'll do it one way or another. At the very least I click "approve workflows to run" button if it exists after checking the workflows aren't tampered with.

falkoschindler:
@Jepson2k
Thanks for the detailed write-up and for keeping the PR description aligned with the template — it made the scope easy to read.

Full disclosure: given the sheer size of this PR, I haven't gone through every code path myself yet; what follows is a first-pass AI-assisted review to get the conversation started. Take the findings as prompts to discuss rather than hard verdicts — I'll dig into the specifics with you once we've agreed on direction (starting with whether to split).

## BLOCKER

### 1. PR scope — take your own offer to split

Quoting the PR body: _"Happy to split along the lines above if preferred — the natural seam is (1) primitives, (2) TransformControls + hover, (3) clipping + ViewHelper + miscellaneous."_ That's the right call. Each slice is independently useful and independently risky. Merging as one makes bisecting any regression in `material()`, `create()`, or the click path painful, since those paths are now touched by multiple concerns at once. The STL + material-update bug fixes in particular are small, independently verifiable, and would land today as standalone PRs.

Most of the remaining BLOCKERs below get much easier to resolve once the PR is split — I'd like to agree on the split first and then work through the rest per slice.

### 2. `rotate_euler` is lost on scene re-send

`Object3D.rotate_euler` calls `scene.run_method('rotate_euler', …)` but never updates `self.R`. Since `Object3D.data` (sent on `update_scene` and every client (re)connect) only includes `self.R`, any rotation set via `rotate_euler` silently disappears on reconnect / tab reload / switch-page revisit. For a robotics UI that's exactly the "joint origin" workflow the PR motivates.

Two reasonable fixes:

- Compute the R matrix Python-side from the Euler + order and call `_rotate()` as usual (simplest; reuses existing persistence path).
- Store Euler state on the object and extend `data`/`update_scene` to carry rotation-mode metadata (bigger change).

I'd go with the first unless there's a specific reason the JS-side Euler path is needed. If there is, please document why and add a test that asserts rotation survives `scene.update()` / a simulated reconnect.

### 3. `set_axes_labels` silently ignores documented options

Python docstring documents `font`, `colorX`, `colorY`, `colorZ`, `size` options. The JS implementation is:

```js
set_axes_labels(opts) {
  if (this.viewHelper && opts && opts.enabled) {
    this.viewHelper.setLabels("X", "Y", "Z");
  }
},
```

Everything except `enabled` is a no-op, and the labels themselves are hard-coded. Either implement the documented knobs or trim the docstring to what actually works. Per AGENTS.md this is a docs/examples mismatch (BLOCKER).

### 4. `set_axes_inset` / `set_axes_labels` take `dict[str, Any]` instead of typed kwargs

Public-API methods should accept typed keyword arguments, matching the rest of `Scene` (`enable_transform_controls`, `set_clipping_planes`, etc.). A raw `dict[str, Any]` loses autocomplete, type-checking, and forces users to discover key names from prose. Suggested:

```python
def set_axes_inset(self, *,
                   enabled: bool = True,
                   size: int = 128,
                   margin: int = 0,
                   margin_x: int | None = None,
                   margin_y: int | None = None,
                   anchor: Literal['bottom-left','bottom-right','top-left','top-right'] = 'bottom-right',
                   ) -> None: ...
```

Same treatment for `set_axes_labels`.

### 5. three.js 0.180 → 0.184 bump needs to wait for NiceGUI 4.0

As a matter of project policy we don't bump dependencies between major versions — the three.js upgrade has to land as part of the NiceGUI 4.0 cycle, not here. For this PR that means either:

- Position the ViewHelper via the existing render viewport math so the feature works on 0.180 and no bump is needed, or
- Split the three.js-0.184-dependent bits (ViewHelper anchor positioning) off into a separate PR queued for 4.0, and land the rest here against 0.180.

## MAJOR

### 1. API shape: `object_id: str` vs `Object3D`

All the new TransformControls / clipping methods take `object_id: str`:

```python
scene.enable_transform_controls(box.id, mode='translate')
scene.set_clipping_planes('sphere', [SceneClipPlane(...)])
```

The rest of NiceGUI's 3D API is method-chained on the object itself (`scene.box().rotate(...).move(...)`). For consistency and to avoid stringly-typed IDs leaking into user code, prefer accepting `Object3D` (and pulling `.id` internally), or hanging the method off `Object3D`:

```python
box = scene.box()
box.enable_transform_controls(mode='translate')
box.set_clipping_planes([...])
```

This also lets type-checkers catch typos like `scene.set_clipping_planes('sphre', …)`.

### 2. Reliance on three.js private `_gizmo`

```js
const gizmo = tc._gizmo || tc.gizmo;
if (!gizmo || !gizmo.materialLib) return;
```

Private properties change between minor three.js releases without notice — exactly the failure mode mentioned in the PR description for the existing TransformControls Group regression. If `materialLib` is the only way to recolor the gizmo, at least fail loudly (console.warn) so a future dep bump surfaces the breakage instead of silently skipping.

### 3. `set_orbit_enabled` vs the TransformControls drag counter

`set_orbit_enabled(False)` + a TransformControls drag that then ends will re-enable orbit (the `dragging-changed` handler sets `this.controls.enabled = true` when `dragging_count` hits 0). User intent is silently lost. Either have `set_orbit_enabled` set a "user wants orbit disabled" latch the drag handler respects, or document that the two APIs don't compose.

### 4. `ground_point` + six screen coords on every click

Every click now carries `ground_point` + `screen_x/y`, `client_x/y`, `offset_x/y` whether the application uses them or not. On an event-heavy scene (drag-like click streams, mobile touch) that's noticeable payload growth per event. Consider sending these only when an app subscribes to them, or at minimum add a short note in the docstring that these fields are optional.

### 5. `hoverable_` in every object's serialization payload

`Object3D.data` now always includes `hoverable_`, adding a field to every scene object's initial snapshot regardless of whether hovering is used. Because this is a positional list (not a dict), appending is a de-facto format version bump for anyone parsing the update_scene payload. Consider either (a) only appending when truthy + a length-tolerant reader on the JS side, or (b) converting `data` to a dict so future fields don't have this problem.

### 6. `_syncHoverGlow` runs on every `pointermove` while hovered

Not a big deal for a box, but for a GLTF with hundreds of mesh descendants this is O(n) `getWorldPosition/Quaternion/Scale` per move. If the hovered root doesn't move between moves, this is pure waste. Gate the sync on "hovered root or ancestors changed transform since last sync" (a matrixWorld version or a hashed transform snapshot) or sync in the render loop instead.

### 7. `raycaster.params.Line.threshold` captured at init only

`raycasterThreshold` is read once during `render()` setup. Changing the prop at runtime won't take effect. Either read it live in the click handler or document that it's init-only.

### 8. Test `test_polar_grid` brittle on child count

```python
assert screen.selenium.execute_script(f'return scene_{scene.html_id}.children.length') == 5
```

Any future default light change turns this red for an unrelated reason. Assert on the PolarGridHelper specifically (traverse for `type === 'PolarGridHelper'`).

## CLEANUP

- `tc.userData.lockAxis` piggybacks on three.js's `userData` — no real conflict, but prefer a non-namespaced local closure variable to keep scene.js state self-contained.
- Several JS blocks carry multi-line prose comments (hover glow, transform controls). Keep the "why" sentences, drop the "what" sentences — the code explains the "what".
- `Polyline` defaults `dash_size=0.008, gap_size=0.004` look tuned to a specific scene scale. Either pick scale-invariant defaults (fraction of bounding-box diagonal?) or document they assume meter-scale scenes.
- `test_transform_controls_mode_change` ends with `# If we got here without errors, the test passes` — prefer an actual assertion (e.g. `tc.mode === 'scale'` after the last click) over implicit-pass tests.
- `rotate_euler` docstring claims "avoiding potential rotation order mismatches" — once the persistence bug is fixed (BLOCKER #2) by computing R server-side, this justification goes away. Consider extending `rotate()` with an optional `order='XYZ'` parameter instead of adding a sibling method.
- `ArrowHelper` linewidth: in most WebGL implementations `LineBasicMaterial.linewidth` is clamped to 1 by the driver. Either use `Line2` / `LineSegments2` (three/addons/lines) or document that `line_width > 1` is a hint that browsers frequently ignore.
- `SceneTransformEventArguments.wx/wy/wz`: typed `float | None` because only `transform` events populate them, but `transform_start` / `transform_end` could populate them trivially too. Populating uniformly removes a sharp edge.

Jepson2k:
Ok will split out into multiple PRs. I'll be opening them as drafts again until I have time to give them a manual pass.

evnchn:
@Jepson2k @falkoschindler Respectfully but my brain really does not work with the below. However, from working with him over a year I can see why @falkoschindler likes it though. @Jepson2k if you want to read about us, https://github.com/zauberzeug/nicegui/pull/5795#issuecomment-4148919449 is a good starting point. 

image

@Jepson2k If it's not too much of an ask, can I request for a "kitchen sink demo" for both this PR and #5965? Seeing **the end goal and deliverable** in person gives me holistic understanding of the `ui.scene` and `ui.codemirror` revamps, for which will speed up the review of the other PRs massively.

I can ask my AI agent to make such a demo, however the information gap between your understanding and my understanding will cause many troubles.

Also @falkoschindler maybe a new Project (https://github.com/zauberzeug/nicegui/projects) to track this? I miss the GitLab-styled project organization I had back in my past internship now.

falkoschindler:
@evnchn I'm also struggling with so many PRs at once. But I still believe it's better to tackle them one by one instead of reviewing one giant monolith - at least if they are functionally independent.
I hope @Jepson2k is carefully picking one or two PRs for us to review first so that we don't have to read all of them just to decide where to start. And a usage demo right in the PR description (or a comment) definitely helps to see what a PR is trying to achieve.
Not sure about the purpose of a new project though. To see all of Jake's PRs, you can visit https://github.com/zauberzeug/nicegui/pulls/Jepson2k. Beyond that, what information do you wish to see?

Jepson2k:
@evnchn for a "kitchen sink" like demo you can check out [Waldo Commander] (https://jepson2k.github.io/Waldo-Commander/) which was the main use case for developing these features. As for individual PR demos I can definitely do that to help show the conceptual goal. I do agree with @falkoschindler about splitting it up though. I've split along functionally independent lines so they can be reviewed, tested, and demo'd independently of one another.

Id also recommend viewing the PRs "at arms length" while they are still in draft since I'm still in the process of removing AI bs, rounding out the implementation, and redoing design decisions that I think are overly Waldo-Commander specific. Once ready I'll open a few related ones at a time so we can focus on those.

Jepson2k:
Since the upcoming PRs can be merged and used independently of one another I don't think its necessary to keep this PR open.

PR #5965, opened 2026-04-17
Add cursor tracking, line anchors, lint diagnostics, tooltips, decorations, and save handler to CodeMirror

Jepson2k:
### Motivation

These features were developed while building an in-app Python script editor for a robotic-arm control UI. Each addition traces to a concrete need in that host application: showing which joint line the cursor is on, pinning diagnostics to moving code positions, rendering lint markers, showing per-line tooltips, highlighting currently-executing lines, suggesting robot-API completions, and handling Ctrl/Cmd+S to run.

### Implementation

**Cursor tracking** (`codemirror.py`, `codemirror.js`):
- `on_cursor_line` handler receives the 1-indexed line number when the cursor moves (debounced 30 ms).
- `set_cursor_line()` / `get_cursor_line()` for programmatic cursor positioning.

**Line anchors** (`codemirror.py`, `codemirror.js`):
- `set_line_anchors()` / `clear_line_anchors()` store named sets of anchors that auto-remap through document edits via CodeMirror's `RangeSet`.
- `line_anchor_positions` property mirrors remapped positions on the Python side; JS emits `anchor-positions` events (debounced 50 ms).

**Lint diagnostics** (`codemirror.py`, `codemirror.js`):
- `set_diagnostics()` accepts a list of `{line, message, severity}` dicts and installs CodeMirror's built-in `linter()` so markers render inline and in the gutter without pulling in `lintGutter()` — `lintGutter()` registers a `showTooltip.from(field)` provider that silently suppresses the autocomplete popup.

**Line tooltips** (`codemirror.py`, `codemirror.js`):
- `set_line_tooltips()` / `clear_line_tooltips()` with named sets. Per-line metadata shows on hover via `hoverTooltip`. Optional `_html` value is sanitized with DOMPurify (lazy-imported at mount) before assignment to `innerHTML` to prevent XSS from untrusted sources.

**Decoration API** (`codemirror.py`, `codemirror.js`):
- `set_decorations()` / `clear_decorations()` for character- and line-level decorations keyed by named set.
- `highlight_lines()` for temporary line-level CSS highlighting with optional duration.

**Custom completions** (`codemirror.py`, `codemirror.js`):
- `set_completions()` + `custom_completions` property. Each item supports `label`, `detail`, `info`, `apply`, `type`.

**Save handler** (`codemirror.py`, `codemirror.js`):
- `on_save` fires on `Mod-s`; when set, the browser's default save behavior is suppressed.

**reveal_line** (`codemirror.py`, `codemirror.js`):
- `reveal_line()` scrolls a 1-indexed line into view.

**Bundled bug fixes** (`codemirror.js`):
- `findLanguage` previously logged `${this.language}` (the component's current language prop) and referenced an undefined outer `languages` identifier when no match was found, so the error message named the wrong language and crashed before the supported-name hint could be printed. Fixed to use the `name` argument being searched and the `this.languages` property.
- `beforeUnmount` now syncs the final editor value back to props and clears pending highlight/cursor/anchor debounce timers (was leaking timers and dropping unsaved edits).

Tests in `test_codemirror.py` cover cursor-line events, save events, decorations, line highlighting, diagnostics, and line reveal, in addition to existing custom-completions tests. Docs in `codemirror_documentation.py` include demos for the main features.

### Progress


evnchn:
Same as https://github.com/zauberzeug/nicegui/pull/5964#issuecomment-4268366791 for now

falkoschindler:
Thanks @Jepson2k! Since the PR is still in draft, just let us know when you'd like a full review. In the meantime, two tests are failing on CI ([run](https://github.com/zauberzeug/nicegui/actions/runs/24592580780/job/71935176106)) — worth a look:

**`test_custom_completions_with_none`** (tests/test_codemirror.py:177)

```
RuntimeError: The current slot cannot be determined because the slot stack for this task is empty.
```

The element is created at the top of the test function, outside any page or slot context. Wrap it in a `@ui.page('/')` (and use the `screen`/`user` fixture), or build it inside an explicit container — see how the other tests in this file do it.

**`test_cursor_line_event`** (tests/test_codemirror.py:199)

```
assert any(line == 2 for line in lines)
```

No cursor-line event arrives with `line == 2`. Likely a timing/debounce issue with `screen.wait(0.3)` vs. the 30 ms JS debounce plus event roundtrip, or the click landing on a nested `cm-content` so the arrow keys don't reach the editor. Please replace the fixed `screen.wait(...)` calls with `screen.wait_for(lambda: 2 in lines)` (and the same for `3`) — it's both more reliable and faster than fixed delays.

evnchn:
Hello @Jepson2k. We tend to avoid force-pushes. It may be fine for a draft PR but let's avoid it when the PR is out-of-draft. Thanks! 

Also consider to enable workflow runs at https://github.com/Jepson2k/nicegui/actions

Jepson2k:
@evnchn @falkoschindler Its now ready for review. I was keeping it a draft while I was still working through making some of the features more generalizable rather than specific to my app and cleaning up unnecessary things like the styling changes. As for the force pushes, yes I generally agree. For the workflows, I'm not quite sure what you mean by "enable" them but I was able to manually trigger them on the forked repo.

falkoschindler:
Heads up: because of the sheer size of this PR (~966 lines of hand-written code across JS, Python, tests and docs, ignoring `dist/`), I haven't fully read through the diff myself yet — this first pass is an AI-assisted review. Treat it as a starting point; I'll follow up with my own read after the scope question below is resolved.

## BLOCKER

### 1. Scope is too large — please split

~966 lines covering seven independent features plus two unrelated bug fixes is too much for one review cycle: any single MAJOR finding (e.g. §4 on anchors, §5 on decoration theming) currently blocks the trivial fixes from landing, reviewers can't give each feature proper attention, and a regression in any single piece forces reverting the whole thing. Most of the features use similar CM6 patterns but don't actually share code, so they split cleanly along subsystem lines. Our preferred shape:

- Bug fix: `findLanguage` error-message references.
- Bug fix: `beforeUnmount` timer/value-sync cleanup.
- Decorations + `highlight_lines` (they genuinely share the `_applyAllDecorations` pipeline).
- Line anchors (own `StateField` + `RangeSet` + `anchor-positions` event/mirror).
- Line tooltips (own `StateField` + `hoverTooltip`; introduces the DOMPurify dependency pattern).
- Diagnostics (uses `CM.linter` / `CM.setDiagnostics`, unrelated to the other subsystems).
- Cursor tracking + save handler + `reveal_line` — small editor-event primitives, can share one PR.
- Custom completions.

I know 8 PRs sounds like a lot, but the actual ask is surgical decomposition of code you've already written, not a rewrite — and since you're already pairing with Claude on this (per the commit co-author), this kind of split is exactly what AI assistants are good at. Feed it the diff and the topical grouping above. Should cut the work to an hour or two.

If 8 is still genuinely too much on your end, let me know what you _can_ do and we'll negotiate — but please don't extend this PR any further until we've agreed on the shape.

### 2. PR description / implementation mismatch

The PR body states:

> `set_cursor_line()` / `get_cursor_line()` for programmatic cursor positioning.

Neither method exists in the Python or JS code. Please update the description or implement them (the tests and docs also don't reference them). Per AGENTS.md §7 this counts as a description-quality blocker.

### 3. `package-lock.json` downgrade re-introduces fixed security vulnerabilities

```
"picomatch":            4.0.4 → 4.0.3
"serialize-javascript": 7.0.5 → 7.0.4
```

Both downgrades revert already-applied Dependabot fixes:

- `picomatch` 4.0.4 fixed the ReDoS and POSIX method-injection advisories. Rolling back to 4.0.3 reopens them.
- `serialize-javascript` 7.0.5 fixed the CPU-exhaustion DoS advisory. Rolling back to 7.0.4 reopens it.

This looks like an accidental `npm install` on a machine with a stale lockfile. Please revert the two blocks so Dependabot doesn't re-raise the alerts. (Per AGENTS.md §1, security regressions are hard BLOCKERs on their own — independent of the "unrelated change" concern.)

---

## MAJOR

### 4. `line_anchor_positions` mirror can be silently stale

The Python cache `self._anchor_positions` is only updated by the `anchor-positions` event, which is emitted from `anchorTracker.update()` **only if**:

- `update.docChanged` is true, AND
- `field.size > 0` (early `return` on line [codemirror.js#L414](nicegui/elements/codemirror/codemirror.js#L414))

Consequences:

- `set_line_anchors([...], 'breakpoints')` doesn't touch the Python mirror; `line_anchor_positions` returns `{}` until the first subsequent doc edit.
- `clear_line_anchors('breakpoints')` — same problem in reverse: the entry stays populated because the event is suppressed when the field becomes empty.
- `clear_line_anchors(None)` → the Python dict is never cleared at all.

Fix options:

- Update `_anchor_positions` synchronously on the Python side from `set_line_anchors` / `clear_line_anchors` (the caller already passes lines in, so it's trivial).
- Or emit once more from JS even when the set transitions to empty.

### 5. Decoration colors are hardcoded and theme-ignorant

In [codemirror.js#L532-L550](nicegui/elements/codemirror/codemirror.js#L532-L550) `cm-diff-added`, `cm-diff-deleted`, `cm-highlighted` etc. are wired to fixed RGB values:

```js
".cm-diff-added":   { backgroundColor: "rgba(0, 255, 0, 0.2)" },
".cm-diff-deleted": { backgroundColor: "rgba(255, 0, 0, 0.2)", textDecoration: "line-through" },
...
".cm-highlighted":  { backgroundColor: "rgba(255, 255, 0, 0.3)" },
```

Documented as "Built-in CSS classes" in the Python docstring — so users will rely on them. On dark themes (e.g. `oneDark`, `vscodeDark`) these look garish, and users can't override via the existing theme system. Either:

- Pick colors using CSS variables / mix-blend-mode so they stay readable on both backgrounds, or
- Drop the baked-in styles and have callers bring their own `.classes()`/stylesheets — the decoration API itself doesn't need these.

### 6. DOMPurify fallback silently renders raw HTML as text

```js
if (self.DOMPurify) {
  dom.innerHTML = self.DOMPurify.sanitize(merged._html);
} else {
  dom.textContent = merged._html; // ← raw HTML shown as literal text
}
```

If a user hovers before `import('dompurify')` resolves, they'll see `<b>hello</b>` as literal text in the tooltip. Safe (no XSS), but confusing. Consider either awaiting the import before the tooltip renders, or returning `null` from the tooltip provider when DOMPurify isn't ready.

### 7. Test coverage gaps

PR claims tests for "cursor-line events, save events, decorations, line highlighting, diagnostics, and line reveal" — but nothing covers:

- `set_line_anchors` / `clear_line_anchors` / `anchor-positions` event / `line_anchor_positions` property (the most complex feature in this PR)
- `set_line_tooltips` / hover rendering
- `_html` + DOMPurify sanitization path

Given the anchor-mirror bug in §4, this is where I'd most want coverage.

### 8. `resolve_defaults` not applied to `custom_completions`

All other optional kwargs (`language`, `theme`, `indent`, …) use `DEFAULT_PROP | …`, and the `@resolve_defaults` decorator supports per-instance default overrides. `custom_completions` bypasses it. If this is intentional because `list` defaults don't compose, please add a comment; otherwise align with the rest of the API.

---

## CLEANUP

### 9. API naming inconsistencies

- `set_completions` vs `custom_completions` property — method name drops the `custom_` prefix. Rename to `set_custom_completions` for symmetry.
- `set_diagnostics` has no matching `clear_diagnostics` (you'd pass `[]`). Add one for symmetry with `set_decorations` / `clear_decorations` and `set_line_tooltips` / `clear_line_tooltips`.
- JS method `setDiagnosticsFromPython` — drop the `FromPython` suffix; no other method uses it.

### 10. Weak type hints

`list[dict]`, `dict[int, dict]`, `dict[str, dict[str, int]]` hide the actual shape. Consider `TypedDict`s for `CompletionItem`, `Diagnostic`, `AnchorSpec`, `DecorationSpec` — the docstrings already enumerate the fields. Improves IDE help and mypy signal.

### 11. Defensive code unnecessary in `_update_anchor_mirror`

```python
args = e.args if isinstance(e.args, dict) else {}
```

`GenericEventArguments.args` is always a dict when emitted as an object from JS — the isinstance guard hides a real bug if one ever happens. Drop it.

### 12. Stale dist artifacts

`dist/` grew from 238 → 282 files; the new content-hashed chunks were added but old ones not removed. Not a functional issue, but a `npm run clean && npm run build` would keep the bundle set minimal. Worth doing as part of a normal pre-merge rebuild.

### 13. Watcher cost on `decorations`

The `deep: true` watcher re-runs `setDecorations` on every nested mutation. For large decoration sets this is expensive. Acceptable for now, but worth a `NOTE:` line or a shallow-watch alternative if anyone pushes large diagnostic overlays through this path.

### 14. `on_cursor_line` / `on_save` type hints

Signature declares `Handler[GenericEventArguments]`, which is generic. The `cursor-line` event carries `{line: int}`; consider a small typed payload for better DX (analogous to `ValueChangeEventArguments`).

### 15. `beforeUnmount` still reaches into the global `mounted_app`

```js
const element = mounted_app.elements[this.$props.id.slice(1)];
if (element) element.props.value = this.editor.state.doc.toString();
```

Not introduced by this PR, but it's now the only reason `mounted_app` is touched and a sync-on-unmount path. Would be cleaner to emit a final `update:value` instead. Optional follow-up.

Jepson2k:
Ok will split out into multiple PRs. I'll be opening them as drafts again until I have time to give them a manual pass.

Jepson2k:
Since the upcoming PRs can be merged and used independently of one another I don't think its necessary to keep this PR open.

PR #5967, opened 2026-04-20
Add AG Grid demo for adding rows without losing client-side edits

falkoschindler:
### Motivation

Appending to `grid.options['rowData']` triggers a full client-side rebuild that discards any in-progress cell edits. Users hit this regularly — see [discussion #5952](https://github.com/zauberzeug/nicegui/discussions/5952) — and the workaround so far has been `await grid.load_client_data()`, which doesn't actually prevent the rebuild, it just papers over the data loss.

### Implementation

Adds a new `@doc.demo` right after the existing *Adding rows* section showing the idiomatic AG Grid pattern: `applyTransaction` inside `grid.props.suspend_updates()`. The server-side `rowData` stays in sync without firing a rebuild, so client edits survive.

### Progress

### DIFF
--- a/website/documentation/content/aggrid_documentation.py
+++ b/website/documentation/content/aggrid_documentation.py
@@ -45,6 +45,33 @@ def adding_rows():
     ui.button('Add row', on_click=add)
 
 
+@doc.demo('Adding rows without losing client-side edits', '''
+    Mutating `grid.options['rowData']` triggers a full rebuild on the client and discards any in-progress cell edits.
+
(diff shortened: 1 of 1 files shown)


evnchn:
Yes, it works, but the current way is kinda verbose. Should we have a dedicated API for this?

falkoschindler:
@evnchn Sure, the PR documents the current API and its capabilities. Once we find a better API, we can update the demo.

evnchn (approved):
Sure. Then let's update this demo when the new API comes around

PR #5970, opened 2026-04-21
Fix @keyframes and other at-rules being mangled in scoped styles

### LINKED ISSUE #5969: `@keyframes` in a custom component breaks generated CSS and prevents following component styles from applying
### First Check

### Example Code

```python
# please refer to https://github.com/jhjcpishva/nicegui-issue-5969-css-keyframe
```

### Description

- Vue SFC `<style>...</style>` can not parse `@keyframe { ... }` properly.  Missing closing bracket, `}`.
- The component `<style>` parsed after `@keyframe { ... }` are broken and won't applied by browser.
- ChatGPT says [add_css_prefix() on vbuild.py#L82](https://github.com/zauberzeug/nicegui/blob/v3.10.0/nicegui/vbuild.py#L82) is maybe the key to be fixed

#### Example code

Uploaded here:

https://github.com/jhjcpishva/nicegui-issue-5969-css-keyframe

key issue is here: [components/keyframe_component.vue#L15-L22](https://github.com/jhjcpishva/nicegui-issue-5969-css-keyframe/blob/main/components/keyframe_component.vue#L15-L22)

`@keyframe .. { from { .. } to { .. } }` is rendered as

`@keyframe .. { from { .. } to { .. }` , the last closing bracket is missing.

This causes other styles rendered in same `<style>` to get broken

---

btw, thx for awesome library!

### NiceGUI Version

3.10.0

### Python Version

3.13

### Browser

Chrome

### Operating System

macOS

### Additional Context

`My custom content`  should be styled gradient background.  2nd image shows `<style>` got broken.

Image

Image

falkoschindler:
### Motivation

Fixes #5969.

In Vue SFC `<style>` blocks, `@keyframes` (and any at-rule with nested braces) broke the generated CSS: the rule regex `[^}]+{[^}]+}` stopped at the first `}` inside the keyframes body, producing `@keyframes name { from { … }` with the closing brace missing — which silently broke every subsequent rule in the same `<style>`.

The same class of bug also affected `@supports`, `@container`, `@layer <name> { … }`, `@scope`, and `@font-feature-values`. Descriptor-only at-rules (`@font-face`, `@page`, `@property`, `@counter-style`, `@font-palette-values`, `@color-profile`) weren't producing corrupt output, but the scope prefix was being applied to them, which is semantically wrong.

### Implementation

`add_css_prefix` now extracts at-rules up front via a shared helper `_extract_at_rules` that uses balanced-brace matching, so the main rule-splitting regex never sees nested at-rule braces.

Two categories are handled separately:

- **Verbatim** (`@keyframes` incl. vendor-prefixed, `@font-face`, `@font-feature-values`, `@font-palette-values`, `@page`, `@property`, `@counter-style`, `@color-profile`): bodies are descriptors or non-DOM-selector keywords (`from`/`to`/`N%`), so the block is emitted as-is with whitespace normalized.
- **Conditional** (`@media`, `@supports`, `@container`, `@layer`, `@scope`): bodies contain regular rules whose inner selectors still need the scope prefix, so the body is recursed through `add_css_prefix`. Statement forms like `@layer reset, theme;` are preserved in source order so layer ordering is not lost.

The existing `@media` special case is replaced by this general mechanism.

### Progress

### DIFF
--- a/nicegui/vbuild.py
+++ b/nicegui/vbuild.py
@@ -81,19 +81,23 @@ class VueParser(HTMLParser):  # pylint: disable=abstract-method  # pylint assume
 
 def add_css_prefix(css: str, prefix: str) -> str:
     """Add the prefix (CSS selector) to all rules in ``css``."""
-    media_queries: list[tuple[str, str]] = []
-    while '@media' in css:
-        p1 = css.find('@media', 0)
-        p2 = css.find('{', p1) + 1
-        level = 1
-        while level > 0:
-            level += 1 if css[p2] == '{' else -1 if css[p2] == '}' else 0
-            p2 += 1
-        block = css[p1:p2]
-        media_def = block[:block.find('{')].strip()
-        media_css = block[block.find('{') + 1:block.rfind('}')].strip()
-        css = css.replace(block, '')
-        media_queries.append((media_def, add_css_prefix(media_css, prefix)))
+    # extract at-rules whose bodies are descriptors, keyframe keywords or other non-selector syntax; emitted verbatim
+    css, verbatim_blocks = _extract_at_rules(css, (
+        r'@(?:(?:-\w+-)?keyframes|'
+        r'font-face|font-feature-values|font-palette-values|page|property|counter-style|color-profile)\b'
+    ))
+    verbatim = [re.sub(r'\s+', ' ', b).strip() for b in verbatim_blocks]
+
+    # extract at-rules whose bodies contain regular CSS rules with DOM selectors; inner rules receive the scope prefix
+    css, conditional_blocks = _extract_at_rules(css, r'@(?:media|supports|container|layer|scope)\b')
+    conditional: list[str] = []
+    for block in conditional_blocks:
+        if block.endswith(';'):  # statement form, e.g. `@layer reset, theme;`
+            conditional.append(re.sub(r'\s+', ' ', block).strip())
+        else:
+            header = block[:block.find('{')].strip()
+            body = block[block.find('{') + 1:block.rfind('}')].strip()
+            conditional.append(f'{header} {{ {add_css_prefix(body, prefix)} }}')
 
     lines: list[str] = []
     css = re.sub(re.compile(r'/\*.*?\*/', re.DOTALL), '', css)
@@ -105,5 +109,34 @@ def add_css_prefix(css: str, prefix: str) -> str:
         else:
             line = [i.strip() for i in selectors.split(',')]
         lines.append(', '.join(line) + ' {' + declarations.strip())
-    lines.extend(f'{d} {{ {c} }}' for d, c in media_queries)
+    lines.extend(conditional)
+    lines.extend(verbatim)
     return '\n'.join(lines).strip()
+
+
+def _extract_at_rules(css: str, pattern: str) -> tuple[str, list[str]]:
+    """Extract at-rules matching ``pattern`` from ``css``.
+
+    Returns the CSS with matches removed and a list of the extracted rules in source order.
+    Block-form rules are returned with balanced braces; statement-form rules include the trailing ``;``.
+    """
+    rules: list[str] = []
+    while True:
+        match = re.search(pattern, css)
+        if not match:
+            return css, rules
+        p1 = match.start()
+        brace = css.find('{', p1)
+        semi = css.find(';', p1)
+        if brace == -1 and semi == -1:
+            return css, rules
+        if semi != -1 and (brace == -1 or semi < brace):
+            p2 = semi + 1
+        else:
+            p2 = brace + 1
+            level = 1
+            while level > 0:
+                level += 1 if css[p2] == '{' else -1 if css[p2] == '}' else 0
+                p2 += 1
+        rules.append(css[p1:p2])
+        css = css[:p1] + css[p2:]
--- a/tests/test_vbuild.py
+++ b/tests/test_vbuild.py
@@ -98,6 +98,51 @@ def test_template_with_script():
     ''')
 
 
+def test_template_with_at_rules():
+    check('''
+        <template>
+            <h1>Hello, World!</h1>
+        </template>
+        <style scoped>
+            h1 {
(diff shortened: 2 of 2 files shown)


falkoschindler:
Here is a more detailed description of what the two new regexes do.

---

### Regex 1 — "verbatim" at-rules

Defined in [vbuild.py:83-85](nicegui/vbuild.py#L83-L85) as two adjacent string literals; after Python concatenation:

```
@(?:(?:-\w+-)?keyframes|font-face|font-feature-values|font-palette-values|page|property|counter-style|color-profile)\b
```

Token-by-token:

| Token | Meaning |
|---|---|
| `@` | literal `@` |
| `(?: … \| … )` | non-capturing alternation over the at-rule names |
| `(?:-\w+-)?keyframes` | `keyframes`, optionally preceded by a vendor prefix like `-webkit-`, `-moz-`, `-o-` |
| `(?:-\w+-)?` | optional group `-<word-chars>-`. `\w` = `[A-Za-z0-9_]` (no `-`), so it matches **exactly one** hyphen-delimited token |
| `font-face` … `color-profile` | the six CSS descriptor-only at-rules, as literals |
| `\b` | word boundary — only matches when the next char is *not* `\w` |

**What matches (good):**
- `@keyframes fade` ✅
- `@-webkit-keyframes slide` ✅ (prefix via `-\w+-`)
- `@-moz-keyframes …`, `@-o-keyframes …` ✅
- `@font-face {` ✅
- `@page :first {` ✅ (space after `page` → word boundary)
- `@property --foo {` ✅

**What `\b` correctly rejects:**
- `@pageprint` — no boundary between `e` and `p`, so it does **not** match `@page` ✅
- `@propertyfoo` — same, no match ✅
- `@keyframesfoo` — same, no match ✅

**Subtle property of the optional prefix:**
`(?:-\w+-)?` can only eat a single `\w+` between two hyphens. So:

- `@-webkit-keyframes` → prefix captures `-webkit-`, then `keyframes` matches — good.
- `@-webkit-linear-keyframes` → `-\w+-` greedily tries `-webkit-`, then expects `keyframes` but finds `linear-…`. Regex engine cannot extend `\w+` across `-`. Overall: no match. That's fine because no such real at-rule exists.

**One thing the regex is intentionally *not* doing:**
It does not match `@-moz-document` (vendor-prefixed conditional). That is out of scope for the PR and was not handled before either.

---

### Regex 2 — "conditional" at-rules

Defined in [vbuild.py:90](nicegui/vbuild.py#L90):

```
@(?:media|supports|container|layer|scope)\b
```

Token-by-token:

| Token | Meaning |
|---|---|
| `@` | literal |
| `(?:media\|supports\|container\|layer\|scope)` | the five nestable conditional at-rules |
| `\b` | word boundary — same protective role as above |

**What matches:**
- `@media (min-width: 600px) {` ✅
- `@supports (display: grid) {` ✅
- `@container (width > 100px) {` ✅
- `@layer theme {` ✅ (block form)
- `@layer reset, theme;` ✅ (statement form — handled later by the `;`-vs-`{` check in `_extract_at_rules`)
- `@scope (.card) {` ✅

**What `\b` rejects:**
- `@mediaquery` → no boundary, no match ✅
- `@layered` → no match ✅
- `@containment` → no match ✅

**Note about `@layer`:** both `@layer foo;` and `@layer foo { … }` must match. `\b` is satisfied in both cases (`r` → ` ` or `r` → `;`, both non-word). The helper then disambiguates by comparing the positions of the next `;` and `{`.

---

### How the two regexes cooperate with the existing rule regex

The pre-existing rule splitter `[^}]+{[^}]+}` ([vbuild.py:103](nicegui/vbuild.py#L103)) is the very thing that was broken by `@keyframes`: its `[^}]+` body is single-level and stops at the first inner `}`. The two new regexes plus the balanced-brace scanner in `_extract_at_rules` are specifically a **pre-pass** that removes every at-rule block from the CSS string *before* that naive splitter runs, so the splitter now only sees flat rules and cannot be confused by nested braces.

---

### Small caveats worth naming (not regex bugs, limitations of the approach)

1. **Strings containing braces** — e.g. `content: '}';` inside a `@keyframes` body — would fool the brace balancer. Not introduced by this PR; applies equally to the old `@media` handler.
2. **Comments containing `@keyframes` / `@media`** — comments are stripped *after* at-rule extraction ([vbuild.py:101](nicegui/vbuild.py#L101)), so a commented-out at-rule would be extracted as if real. Same limitation as before.
3. **Nested at-rules** — `@media { @keyframes { … } }` would get the inner `@keyframes` lifted to top level. Semantics change only if the author relied on nesting, which browsers support unevenly anyway.

Overall the two new patterns look correct for what the PR description says they target, and `\b` is doing the right job of preventing false matches on longer identifiers.

evnchn:
Hi @falkoschindler,

I (@evnchn) asked Claude Code to review this PR with empirical testing on Windows. During the review, we identified a few edge cases:

1. **`@starting-style`** (CSS Transitions Level 2) breaks with unbalanced braces - it's not in either regex list
2. **Commented at-rules** like `/* @keyframes x { ... } */` are incorrectly extracted and emitted
3. **Braces in strings** like `content: '{}'` cause parsing issues

We explored a more generalized approach that handles ALL CSS at-rules without needing to maintain hardcoded lists. The key changes:

- Strip comments BEFORE at-rule extraction
- String-aware brace balancing throughout
- Heuristic fallback for unknown future at-rules (checks for selector-like patterns)

I've pushed an alternative implementation to my fork for your consideration:
https://github.com/evnchn/nicegui/tree/css-aware-at-rule-parsing

The trade-off is +121/-21 lines, but it's more robust against future CSS features. Happy to discuss if this direction is interesting, or if you'd prefer to just add `@starting-style` to the existing approach.

---
*This comment was drafted by Claude Code at @evnchn's request after empirical testing on Windows.*

falkoschindler:
I'm closing this PR in favor of #5976.

evnchn (commented):
Lemme quickly dismiss the review by saying something. But yeah the other PR better.

PR #5974, opened 2026-04-22
Return JSON from the 404 handler for non-page endpoints

falkoschindler:
### Motivation

NiceGUI registers a status-code handler `@app.exception_handler(404)`.
Starlette's `ExceptionMiddleware` matches status-code handlers **before** exception-class handlers,
so this handler wins over FastAPI's default `HTTPException → JSON` handler for **every** 404 in the app —
including ones raised from regular FastAPI routes mounted on the NiceGUI app:

```python
@app.get('/api/item/{id}')
def get_item(id: int):
    if id not in items:
        raise HTTPException(404, 'item not found')
```

The caller expects `{"detail": "item not found"}`, but gets NiceGUI's HTML error page (or a crash inside the render pipeline when the handler runs outside a fully-initialized page context).
Users currently have to work around this by returning a `JSONResponse` manually — see discussion at https://github.com/zauberzeug/growbotic_server/pull/8#discussion_r3085390879.

### Implementation

Distinguish three cases inside the 404 handler using `request.scope`:

| Case                           | `endpoint` in scope | `nicegui_page_path` | Response                         |
| ------------------------------ | ------------------- | ------------------- | -------------------------------- |
| Unknown URL (no route matched) | ✗                   | ✗                   | HTML error page (unchanged)      |
| `@app.get` etc. raises 404     | ✓                   | ✗                   | **FastAPI's default JSON** (new) |
| `@ui.page` raises 404          | ✓                   | ✓                   | HTML error page (unchanged)      |

`nicegui_page_path` is the same discriminator the existing 500 handler uses one block below.
`'endpoint' in request.scope` — set by Starlette's router only when a route actually matched — is needed to avoid converting the browser-facing unknown-URL case to JSON.

For the new API-endpoint case we delegate to FastAPI's `http_exception_handler` rather than re-raising, because Starlette does not re-dispatch exceptions raised from an exception handler — they bubble straight to `ServerErrorMiddleware` and become 500s.

### Progress

### DIFF
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -163,7 +163,6 @@ async def _shutdown() -> None:
 
 @app.exception_handler(404)
 async def _exception_handler_404(request: Request, exception: Exception) -> Response:
-    log.warning(f'{request.url} not found')
     if 'endpoint' in request.scope and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
         # non-page endpoints raising 404 should get JSON, not our HTML error page
         # NOTE: match Starlette's HTTPException (the base class) so e.g. auth dependencies that raise it directly are covered
@@ -179,6 +178,7 @@ async def _exception_handler_404(request: Request, exception: Exception) -> Resp
             if name in request.query_params and name != 'request'
         }
         return await page('')._wrap(root)(request=request, **kwargs)  # pylint: disable=protected-access
+    log.warning(f'{request.url} not found')
     with Client(page(''), request=request) as client:
         error_content(404, exception)
     return client.build_response(request, 404)
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -11,6 +11,7 @@ import socketio
 from fastapi import HTTPException, Request
 from fastapi.exception_handlers import http_exception_handler
 from fastapi.responses import FileResponse, Response
+from starlette.exceptions import HTTPException as StarletteHTTPException
 
 from . import air, background_tasks, binding, core, favicon, helpers, json, run, welcome
 from .app import App
@@ -162,8 +163,9 @@ async def _shutdown() -> None:
 
 @app.exception_handler(404)
 async def _exception_handler_404(request: Request, exception: Exception) -> Response:
-    if 'endpoint' in request.scope and not request.scope.get('nicegui_page_path') and isinstance(exception, HTTPException):
+    if 'endpoint' in request.scope and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
         # non-page endpoints raising 404 should get JSON, not our HTML error page
+        # NOTE: match Starlette's HTTPException (the base class) so e.g. auth dependencies that raise it directly are covered
         return await http_exception_handler(request, exception)
     root = core.root
     if root is not None:
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -9,6 +9,7 @@ from typing import Any
 
 import socketio
 from fastapi import HTTPException, Request
+from fastapi.exception_handlers import http_exception_handler
 from fastapi.responses import FileResponse, Response
 
 from . import air, background_tasks, binding, core, favicon, helpers, json, run, welcome
@@ -161,6 +162,9 @@ async def _shutdown() -> None:
 
 @app.exception_handler(404)
 async def _exception_handler_404(request: Request, exception: Exception) -> Response:
+    if 'endpoint' in request.scope and not request.scope.get('nicegui_page_path') and isinstance(exception, HTTPException):
+        # non-page endpoints raising 404 should get JSON, not our HTML error page
+        return await http_exception_handler(request, exception)
     root = core.root
     if root is not None:
         kwargs = {
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -163,6 +163,7 @@ async def _shutdown() -> None:
 
 @app.exception_handler(404)
 async def _exception_handler_404(request: Request, exception: Exception) -> Response:
+    log.warning(f'{request.url} not found')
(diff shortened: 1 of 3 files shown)


evnchn:
Hi @falkoschindler — pushed 1f5c75a on top of your commit. It broadens the isinstance from `fastapi.HTTPException` to `starlette.exceptions.HTTPException` and adds a matching regression test.

**Why:** reviewed the PR empirically with a small harness (boots the app in a thread, hits it with httpx). A 404 raised from an `@app.get` route via `starlette.exceptions.HTTPException` — which is what FastAPI's security dependencies and several third-party auth libraries raise — was still falling through to the HTML error page. Since `fastapi.HTTPException` subclasses the Starlette one (not the reverse), matching the base class covers both. `fastapi.http_exception_handler` happily accepts either.

Revert or tweak as you see fit — happy to split it out into a follow-up PR instead if you'd rather keep this one minimal.

<details>
<summary>Full empirical review (A/B against main, table of findings, remaining nits)</summary>

### A/B results (main vs PR head 0f9b9b2 vs 1f5c75a)

| Probe | main | 0f9b9b2 | 1f5c75a | Desired |
|---|---|---|---|---|
| Unknown URL | HTML | HTML | HTML | HTML ✓ |
| `@app.get` raises `fastapi.HTTPException(404)` | HTML | **JSON** | JSON | JSON ✓ |
| `@app.get` raises `starlette.exceptions.HTTPException(404)` | HTML | HTML | **JSON** | JSON ✓ |
| `@app.get` via `Depends(...)` raising Starlette HTTPException | HTML | HTML | **JSON** | JSON ✓ |
| 404 with custom headers | HTML | JSON + headers preserved | JSON + headers preserved | JSON ✓ |
| `POST` endpoint raising 404 | HTML | JSON | JSON | JSON ✓ |
| Mounted FastAPI sub-app route raises 404 | JSON | JSON | JSON | JSON ✓ (sub-app handles its own) |
| `@ui.page` raises 404 | HTML | HTML | HTML | HTML ✓ |
| `@ui.page` OK | 200 HTML | 200 HTML | 200 HTML | 200 HTML ✓ |

Nice side-effect: the `log.warning(f'{request.url} not found')` is now skipped for API 404s, removing a class of log noise.

### Remaining notes (all CLEANUP, none blocking)

- **Parameter name `exception`**: typed as `Exception` but treated as `HTTPException` inside the narrowed branch. Harmless — the isinstance narrows it — but `exc` matches FastAPI's convention. Optional rename.
- **Comment**: the `# non-page endpoints raising 404 should get JSON, not our HTML error page` comment states the *what*. The *why* (Starlette's status-code handlers win over class handlers; re-raising bubbles to `ServerErrorMiddleware`) is the load-bearing part and currently only lives in the PR description. I added a single NOTE about the Starlette base-class choice in 1f5c75a; happy to expand with the re-raise rationale if you want.
- **Test coverage considered but not added**: an explicit assertion that `WWW-Authenticate`/custom headers round-trip on the JSON path (empirically verified via the harness — FastAPI's handler forwards `exc.headers`). Not in the existing test surface, probably fine to skip.

### Verification
Harness + pytest collection: https://gist-or-local (script lives only locally). Both new tests are collected; couldn't run the `screen` fixture end-to-end without Chrome, but the harness exercises the exact same code path (same routes, same httpx call, same assertions) and shows every intended behavior holds and nothing regresses.

</details>

falkoschindler:
Great catch, @evnchn!
I just parameterized the test to avoid duplication. Behavior should be unchanged.

evnchn:
@falkoschindler — quick ask: could you loop in Paula on the log behavior in 76cfd93 before locking it in? She hit the original bug and her usage pattern should be the deciding vote.

<details>
<summary>Why I'm asking (the trade-off)</summary>

The move of `log.warning` to the top of the handler in 76cfd93 means every API JSON 404 now emits a WARNING — including the deliberate "item not found" responses that were the whole point of the PR. So:

- **Your framing (observability):** restores the warning for static-file mounts and surfaces unexpected API 404s. Useful.
- **The flip side (log spam):** for a REST API where 404 is a normal "item not found" signal, every legitimate miss now hits the warning log. In high-traffic endpoints that's load-bearing noise.

My own lean (evnchn's) is toward silent — or at most `log.info` — for the API JSON branch, keeping `log.warning` for the HTML/unknown-URL fallback. That preserves observability for "real" 404s while letting REST endpoints behave the way FastAPI users expect.

But this is genuinely a usage-pattern call, not a correctness call — whichever way Paula's actual setup wants it is the right answer.

</details>

*(Posting via Claude Code at evnchn's request.)*

falkoschindler:
Thanks for flagging the log trade-off, @evnchn. Weighing it against Paula's actual use case (see below), I've moved `log.warning` back into the HTML-fallback branch in fbbfb53a, restoring the pre-PR behavior for the log. The log assertion in `test_404_for_non_existing_static_file` has been dropped accordingly — the status code and the issue-2570 body check remain as the meaningful guards.

For context on Paula's original use case: her concern was the response format (HTML returned where JSON was expected), not logging behavior. The log-level decision is therefore orthogonal to what she encountered, and the quieter API branch aligns well with the REST-style expectations of her setup.

@codingpaula @jsb-zz — since this PR originated from your discussion in zauberzeug/growbotic_server#8, I'd like to give you a chance to review before we merge. The change addresses the root cause (NiceGUI's 404 handler intercepting every 404) so that `raise HTTPException(404, ...)` from API endpoints now returns FastAPI's default JSON response. Please let me know if anything still doesn't match your setup.

codingpaula (approved):
@evnchn @falkoschindler I like! Thanks for looking into this!

evnchn:
> @evnchn @falkoschindler I like!

I reserve the right to the positive-vibe interpretation that @codingpaula doesn't just like the PR's code, but also likes how we work, and maybe our Claude Code agents empowering us as well 😉

Anyways, let's merge!

PR #5976, opened 2026-04-22
Fix scoped CSS parsing and root-element selectors via tinycss2

### LINKED ISSUE #5969: `@keyframes` in a custom component breaks generated CSS and prevents following component styles from applying

### LINKED ISSUE #5972: Scoped CSS does not apply to the component root element because selectors are rewritten as descendants
### First Check

### Example Code

```vue
<template>
  <div class="root">Root</div>
</template>

<script>
export default {}
</script>

<style scoped>
.root {
  /* this won't get applied */
  background-color: green;
}
</style>
```

### Description

Hi apologies raising another issue. I've noticed/understood this while poking previous one.

*I was not able to summarize well enough here. Hope coding agents can help to understand the content.. 😔

Vue component `<style scoped>` are rendered to html as `*[data-xx} ...`, called `prefix` in source code.
I've realize this won't match to root element of `<template>`

The above is the minimal example code to reproduce `<style scoped>` not applied.

.

However `<style scoped>` has no issue for descendants elements.

#### Example Code (2)

Image

<details><summary>scoped_style_component.vue</summary>
<p>

```vue
<template>
  <div class="root">
    <div class="title">This is `.title`</div>
    <div class="description">This is `.description`</div>
  </div>
</template>

<script>
export default {}
</script>

<style scoped>
/* ⚠️ scoped && selector based root will not match */
.root {
  background-color: darkgreen;  padding: 16px;
}
.root .title {
  background-color: darkgoldenrod;
}

/* ✅ scoped to children elements are okay */
.title {
  color: greenyellow;
}
.description {
  color: yellow;
}
</style>

<style>
/* ✅ non-scoped style are okay */
.root {
  border: 4px solid yellowgreen;  padding: 16px;
}
.root .title {
  padding: 16px;
}
</style>

```

</p>
</details> 

Keys are

- `<style scoped>`
  - ⚠️ `.root`
  - ⚠️ `.root .title`
  - ✅ `.title` okay

### NiceGUI Version

3.10.0

### Python Version

3.13

### Browser

Chrome

### Operating System

macOS

#### Experiment

Here is my dirty play-around log/result to `vbuild.py`. This is not good solution. Probably can not handle many different cases. (eg. `.root > .title`, `.root:hover`)
It also need only to classes used to root element, not to duplicate all selector `*[data-xx] .title, .title[data-xx]` 

```css
/* <style scoped> input */
.root { .. }
.root .title { .. }

/* output */
*[data-xx] .root, .root[data-xxx]  {  ..  }
*[data-xx] .root .title, .root[data-xxx] .title  {  ..  }

```

<details><summary>Diff `vbuild.py` </summary>
<p>

```diff

diff --git a/nicegui/vbuild.py b/nicegui/vbuild.py
index b59cf8c1..4cc30fdc 100644
--- a/nicegui/vbuild.py
+++ b/nicegui/vbuild.py
@@ -105,7 +105,19 @@ def add_css_prefix(css: str, prefix: str) -> str:
     for rule in re.findall(r'[^}]+{[^}]+}', css):
         selectors, declarations = rule.split('{', 1)
         if prefix:
-            line = [(prefix + ' ' + i.replace(':scope', '').strip()).strip() for i in selectors.split(',')]
+            line = []
+            for i in selectors.split(','):
+                ii: str = i.replace(':scope', '').strip()
+                line.append((prefix + ' ' + ii).strip())
+                
+                first_selector, *remain_selectors = ii.split(" ", maxsplit=1)
+                line.append((
+                    prefix.replace("*", first_selector) + 
+                    " " +
+                    "".join(remain_selectors).strip()
+                    ).strip()
+                )
+            # line = [(prefix + ' ' + i.replace(':scope', '').strip()).strip() for i in selectors.split(',')]
         else:
             line = [i.strip() for i in selectors.split(',')]
         lines.append(', '.join(line) + ' {' + declarations.strip())


```

</p>
</details>

falkoschindler:
### Motivation

Fixes #5969 and #5972 (supersedes #5970).

Two independent bugs in `nicegui/vbuild.py`'s scoped-style processing, both rooted in the same fragile ad-hoc CSS parsing:

- **#5969** — At-rules with nested braces (`@keyframes`, `@supports`, `@container`, `@layer`, `@scope`, `@font-feature-values`) were mangled. The rule regex `[^}]+{[^}]+}` stopped at the first `}` inside a nested block, silently corrupting every subsequent rule in the same `<style>` section.
- **#5972** — Selectors targeting the component's root element (`.root { … }`, `.root .title { … }`) never matched at runtime. The rewrite prepended `*[data-NAME]` with a descendant combinator, producing `*[data-NAME] .root`, which cannot match the root itself (an element is not a descendant of itself). Since only the template's root element carries `data-NAME`, dual emission is needed to cover both "first compound is the root" and "first compound is a descendant".

### Implementation

Replaces the hand-rolled brace-matching + regex splitting with a [tinycss2](https://pypi.org/project/tinycss2/)-based pipeline. `tinycss2` was previously a dev-only dependency; it's now promoted to a runtime dependency (+28 KB wheel, +12 KB for its one transitive dep `webencodings`; no new entries in the resolved lock graph).

- `tinycss2.parse_stylesheet` returns `QualifiedRule` / `AtRule` objects with balanced `{…}` bodies pre-grouped, which removes the need for manual brace counting and the dedicated `@media` special case.
- At-rules are dispatched by kind:
  - **Conditional** (`@media`, `@supports`, `@container`, `@layer`, `@scope`): inner rules are recursed through `_render_rule` so their selectors still receive the scope attribute. Statement-form at-rules like `@layer reset, theme;` are preserved as-is.
  - **Verbatim** (`@keyframes` incl. vendor-prefixed, `@font-face`, `@font-feature-values`, `@font-palette-values`, `@page`, `@property`, `@counter-style`, `@color-profile`): bodies are descriptors or keyframe keywords, so the block is emitted with whitespace collapsed and no selector rewriting.
- Scoped selector rewriting is now dual-emission: for each comma-separated selector, `_scoped_selector_variants` returns `<first-compound>[data-NAME] <rest>` and `*[data-NAME] <full>`, joined by a comma. The first form matches when the first compound is the scoped root (which carries the attribute); the second preserves the descendant match. Inside the first compound, the attribute is inserted before any pseudo-class/element (`:hover`, `::before`) so the resulting selector stays valid.
- `:scope` is stripped before rewriting (Vue's scope-root semantics are then produced by the dual-variant output).
- CSS parsing happens once per component-class registration (in `VBuild.__init__`, called from `Element.__init_subclass__`), never on the request path. Full `VBuild(path)` on the `on_off.vue` example takes ~90 µs; no user-facing latency impact.

One cosmetic side effect worth calling out: tinycss2 normalizes string-literal quotes to double quotes during serialization (`'MyFont'` → `"MyFont"`, `url('/fonts/my.woff2')` → `url("/fonts/my.woff2")`). Semantically identical, browsers accept both.

### Progress

- [x] Pytests have been added and updated.


evnchn:
Hi @falkoschindler,

Tested #5976 on Windows - all 8 tests pass and the tinycss2 approach handles edge cases much better than regex.

While testing, I identified a few edge cases that aren't currently covered by the test suite. I've added 7 additional tests and pushed them to my fork in case you'd like to cherry-pick:

**Branch:** https://github.com/evnchn/nicegui/tree/additional-vbuild-tests
**Commit:** https://github.com/evnchn/nicegui/commit/545277a6

**New tests cover:**
| Test | Why it matters |
|------|----------------|
| `@media` queries | Surprisingly untested - core CSS feature |
| `@container` queries | Listed in conditionals but untested |
| `@starting-style` | Newer CSS (Transitions L2), broke in #5970 |
| Nested `@keyframes` in `@media` | Ensures at-rules stay nested |
| Pseudo-elements (`::before`, `::after`) | Attr must insert before `::` |
| Complex pseudos (`:hover::before`) | Multiple pseudo combos |
| Braces in `content: '{}'` | String handling edge case |
| CSS comments | Confirms they're stripped, not processed |

Cherry-pick command:
```bash
git fetch https://github.com/evnchn/nicegui.git additional-vbuild-tests
git cherry-pick 545277a6
```

---
*Comment by Claude Code at @evnchn's request after empirical testing.*

falkoschindler:
Thanks @evnchn — cherry-picked 545277a6, and pushed a small follow-up cleanup on top:

- Folded `@media`, `@container` and `@starting-style` into the existing `test_scoped_style_with_at_rules` kitchen-sink so one test exercises each at-rule _type_ (conditional / verbatim / statement) rather than one test per keyword. Dropped the three standalone tests that became redundant.
- Kept the four orthogonal edge-case tests (nested at-rules, pseudo-elements, braces in content strings, comment stripping) and moved them up next to the other CSS tests so CSS coverage lives in one block.

All 12 tests green.

evnchn (approved):
Nice. The PR is very comprehensive and I don't know of any gaps left.

PR #5980, opened 2026-04-23
Preserve aspect ratio in `ui.interactive_image` when `size` is not set

### LINKED ISSUE #5978: Website - unnaturally stretched text in Sponsors section
### First Check

### Example Code

```python

```

### Description

The Sponsors section on the landing page at https://nicegui.io looks very strange to me on Safari 26.4  The sponsor logos are all stretched out.

 Image

### NiceGUI Version

n/a

### Python Version

n/a

### Browser

Safari

### Operating System

macOS

### Additional Context

_No response_

falkoschindler:
### Motivation

Fixes #5978.

Sponsor logos on https://nicegui.io render stretched on Safari. Root cause: `ui.interactive_image` sets `aspect-ratio` on its outer `<div>` only when an explicit `size` prop is provided. Without `size`, the div's width is left to the flex/block layout algorithm, and the inner `` stretches to match. Different browsers resolve this ambiguity differently — Chrome falls back to the image's intrinsic width, Safari stretches the flex item.

### Implementation

When no `size` prop is supplied, derive the outer div's `aspect-ratio` from the image's `naturalWidth / naturalHeight` once it loads:

```js
aspectRatio() {
  if (this.size) return this.size[0] / this.size[1];
  if (this.loaded_image_width && this.loaded_image_height) {
    return this.loaded_image_width / this.loaded_image_height;
  }
  return undefined;
}
```

Trade-offs:

- Users with `size` set: unchanged.
- Users with both CSS dimensions set (`w-64 h-32`): unchanged — CSS ignores `aspect-ratio` when both dimensions are explicit.
- Users with only one CSS dimension set: previously rendered at a browser-dependent width; now snaps to the natural aspect ratio on load. Before load the div still has no dimensions, so it grows from zero to the correct size when the image arrives — just like a plain `` would.

### Progress

### DIFF
--- a/nicegui/elements/interactive_image.js
+++ b/nicegui/elements/interactive_image.js
@@ -1,6 +1,6 @@
 export default {
   template: `
-    <div :style="{ position: 'relative', aspectRatio: size ? size[0] / size[1] : undefined }">
+    <div :style="{ position: 'relative', aspectRatio: aspectRatio }">
       <img
         ref="img"
         :src="computed_src"
@@ -147,6 +147,13 @@ export default {
     },
   },
   computed: {
+    aspectRatio() {
+      if (this.size) return this.size[0] / this.size[1];
+      if (this.loaded_image_width && this.loaded_image_height) {
+        return this.loaded_image_width / this.loaded_image_height;
+      }
+      return undefined;
+    },
     onCrossEvents() {
       if (!this.cross && !this.$slots.cross) return {};
       return {


evnchn (approved):
Can recreate: 

Image

For the pushforward of the PR I will make the strongest argument for it here: Since the past size is **undefined** it means whatever behaviour we see is not guaranteed and so changing that to a solid definition is a move we can do without classified as a breaking change. 

However, I **cannot** exclude the possibility that the current undefined size nature is being used by the user, and this actually breaks certain configs the user is exploiting the undefined nature with. Nevertheless, those configs are fragile (works in Safari not in Chrome, works in Chrome not in Safari) so I am pro-PR in the end.

PR #5981, opened 2026-04-23
Fix `ReferenceError` from dangling weakref proxies in `ui.run_with`

### LINKED ISSUE #5979: `ReferenceError` in `_get_server_instance()` when iterating over `gc.get_objects()`
### First Check

### Example Code

```python
import gc
import weakref

# Inject dead proxy into gc.get_objects()
_original = gc.get_objects

class _Weak:
    pass

# simulate weakref object in gc
def _patched():
    obj = _Weak()
    proxy = weakref.proxy(obj)
    del obj
    return [proxy] + _original()

gc.get_objects = _patched

# Minimal NiceGUI app
from fastapi import FastAPI
from nicegui import ui

app = FastAPI()

@ui.page("/")
def index():
    ui.label("Hello")

ui.run_with(app)

if __name__ == "__main__":
    import uvicorn
    uvicorn.run(app, port=8000)
```

### Description

The `_get_server_instance()` function introduced in [commit 2fcd89f](https://github.com/zauberzeug/nicegui/commit/2fcd89f61e4b6300a980a5effe37a79fe6f77e9b) can raise a `ReferenceError` when iterating over `gc.get_objects()`.

This function uses `gc.get_objects()` to find the uvicorn.Server instance, but it doesn't handle the case where a weakref proxy in the GC-tracked objects becomes invalid during iteration. When `isinstance()` is called on a dead proxy, it raises `ReferenceError: weakly-referenced object no longer exists`.

### NiceGUI Version

3.10.0

### Python Version

from 3.12 to 3.14

### Browser

Other

### Operating System

Linux

### Additional Context

_No response_

giunio-prc:
### Motivation
Fix implementation for managing weakref in lifespan context manager of `run_with`
Related issue https://github.com/zauberzeug/nicegui/issues/5979

### Implementation
Decomposition of a pythonic list comprehension that provokes the error

### Progress

### DIFF
--- a/tests/test_run_with.py
+++ b/tests/test_run_with.py
@@ -1,4 +1,3 @@
-import gc
 import weakref
 
 import pytest
@@ -13,25 +12,10 @@ def test_run_with_rejects_self_mount():
         ui.run_with(app)
 
 
-class _Weak:
-    pass
-
-
-@pytest.mark.asyncio
-async def test_get_server_instance_handles_dead_weakref():
-    """Iterating gc.get_objects() with isinstance() should handle dead weakref proxies."""
-    original = gc.get_objects
-
-    def patched():
-        obj = _Weak()
-        proxy = weakref.proxy(obj)
-        del obj
-        return [proxy, *original()]
-
-    gc.get_objects = patched
+async def test_run_with_tolerates_dangling_weakref_proxy():
+    """A dangling weakref.proxy in gc.get_objects() must not crash ui.run_with() startup."""
+    _ = weakref.proxy(set())  # referent is unreferenced, proxy is dangling immediately
     fastapi_app = FastAPI()
-    ui.run_with(app=fastapi_app)
-
-    # should not fail
-    async with fastapi_app.router.lifespan_context(fastapi_app) as _ct:
+    ui.run_with(fastapi_app)
+    async with fastapi_app.router.lifespan_context(fastapi_app):
         pass
--- a/tests/test_run_with.py
+++ b/tests/test_run_with.py
@@ -12,7 +12,7 @@ def test_run_with_rejects_self_mount():
         ui.run_with(app)
 
 
-async def test_run_with_tolerates_dangling_weakref_proxy():
+async def test_run_with_tolerates_dangling_weakref_proxy(nicegui_reset_globals):
     """A dangling weakref.proxy in gc.get_objects() must not crash ui.run_with() startup."""
     _ = weakref.proxy(set())  # referent is unreferenced, proxy is dangling immediately
     fastapi_app = FastAPI()
--- a/tests/test_run_with.py
+++ b/tests/test_run_with.py
@@ -1,4 +1,8 @@
+import gc
+import weakref
+
 import pytest
+from fastapi import FastAPI
 
 from nicegui import app, ui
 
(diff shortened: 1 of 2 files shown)


evnchn:
TL;DR: @falkoschindler I prepared a minimized variant on my fork that keeps the fix but trims the main-code delta and fixes a pytest isolation issue: `evnchn:nicegui:copilot/pr-5981-minimized-delta`.

<details>
<summary>Context, links, and rationale</summary>

Posted via **GitHub Copilot CLI (GPT-5.3-Codex)** after local repro/testing.

### Branch with minimized delta
- `https://github.com/evnchn/nicegui/tree/copilot/pr-5981-minimized-delta`

### Direct comparison links
- Original PR branch vs minimized branch:
  `https://github.com/giunio-prc/nicegui/compare/fix/gc-get-objects-reference-error...evnchn:nicegui:copilot/pr-5981-minimized-delta`
- Original PR to-main delta:
  `https://github.com/zauberzeug/nicegui/compare/main...giunio-prc:nicegui:fix/gc-get-objects-reference-error`
- Minimized branch to-main delta:
  `https://github.com/zauberzeug/nicegui/compare/main...evnchn:nicegui:copilot/pr-5981-minimized-delta`

### Why this variant
1. Keeps the weakref `ReferenceError` fix, but narrows exception handling to the `isinstance(...)` site (smallest safe behavioral change in `ui_run_with.py`).
2. Fixes test isolation in `tests/test_run_with.py`: uses `monkeypatch.setattr(gc, 'get_objects', patched)` so global `gc.get_objects` is restored automatically.

### Pytest issue in current PR test
The current test directly assigns `gc.get_objects = patched` and does not restore it. That can leak global state across tests in the same process and cause hard-to-trace flakiness.

If helpful, I can open a PR from this branch so the diff is easier to review inline.
</details>

evnchn:
@falkoschindler but test is fail tho

tests/test_run_with.py
```diff
@@ -12,7 +12,7 @@ def test_run_with_rejects_self_mount():
         ui.run_with(app)
 
 
-async def test_run_with_tolerates_dangling_weakref_proxy():
+async def test_run_with_tolerates_dangling_weakref_proxy(nicegui_reset_globals):
```
giunio-prc, line 15: just wondering why pylint doesn't warn for unused argument

falkoschindler: CI runs `pylint ./nicegui`, ignoring the tests/ directory. This is because pylint would complain about many things like missing docstrings, redefining names, unused variables... so we decided to focus on nicegui/. But it might be worth reconsidering this decision in the future and to setup safety nets for the whole repository.

falkoschindler (approved):
Thanks for tracking this down and sending a fix, @giunio-prc! 🙏

I pushed small follow-up commits on top of yours:

1. **Refine `_get_server_instance()` weakref-proxy handling** — swapped the `try/except ReferenceError` for an explicit `type(obj) in weakref.ProxyTypes` guard before the `isinstance()` call. The effect is the same (dangling proxies can't crash the loop), but the intent reads more directly: "skip proxies, then check the remaining objects" rather than "try isinstance and catch on failure". It also avoids swallowing any other `ReferenceError` that might come from an unrelated source.
2. **Simplify the regression test** — a dangling `weakref.proxy` naturally lives in `gc.get_objects()`, so there's no need to monkey-patch `gc.get_objects`. Dropping that also removes the independence issue (no teardown needed) and lets the test shrink to a few lines using a builtin `set()` as the referent. I verified it still catches the bug by temporarily reverting the fix.

LGTM on the overall approach — approving. Thanks again!

giunio-prc:
Thank you both @falkoschindler and @evnchn for the quick actions and for the support! Much appreciated 🙏

tests/test_run_with.py
giunio-prc: ok, that makes sense, thanks for the clarification

PR #5982, opened 2026-04-23
Fix incorrect identifiers in CodeMirror findLanguage error message

Jepson2k:
### Motivation

`findLanguage` is called with a `name` argument that is searched against the supported languages. When no match is found, the error message previously referenced `this.language` (the editor's currently configured language prop) instead of the actual `name` being searched, so it reported the wrong language as missing. The follow-up `console.info` line then referenced a bare `languages` identifier that does not exist in scope, which threw a `ReferenceError` before the supported-name hint could be printed.

### Implementation

- Use `name` for the error message so the log line reflects the actual lookup that failed.
- Use `this.languages` for the supported-names list so the hint actually prints instead of throwing a `ReferenceError`.

### Progress

- [x] Pytests are not necessary. (Error-path log message; existing tests pass.)

### DIFF
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -50,8 +50,8 @@ export default {
         for (const alias of [language.name, ...language.alias])
           if (name.toLowerCase() === alias.toLowerCase()) return language;
 
-      console.error(`Language not found: ${this.language}`);
-      console.info("Supported language names:", languages.map((lang) => lang.name).join(", "));
+      console.error(`Language not found: ${name}`);
+      console.info("Supported language names:", this.languages.map((lang) => lang.name).join(", "));
       return null;
     },
     // Get the names of all supported languages
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -68,9 +68,8 @@ export default {
         return;
       }
 
-      const lang_description = this.findLanguage(language, this.languages);
+      const lang_description = this.findLanguage(language);
       if (!lang_description) {
-        console.error("Language not found:", language);
         return;
       }
 


falkoschindler:
@Jepson2k
Nice catch on both of these — they fix a log line that was silently broken. While you're in this area, would you be open to folding in two tiny follow-ons from the same error path? They're part of the same "findLanguage error handling was slightly off" story, so the slice stays coherent.

**1. Drop the now-duplicated error in `setLanguage`**

With this PR, `findLanguage` logs a useful `Language not found: <name>` plus the supported-names hint. The caller then logs the same thing again with less information:

```js
const lang_description = this.findLanguage(language, this.languages);
if (!lang_description) {
  console.error("Language not found:", language); // duplicate of findLanguage's log
  return;
}
```

My guess is this line was added _because_ the original log inside `findLanguage` was broken — once your fix lands, it's redundant noise. Suggested diff:

```diff
       const lang_description = this.findLanguage(language, this.languages);
       if (!lang_description) {
-        console.error("Language not found:", language);
         return;
       }
```

**2. Remove the dead second argument to `findLanguage`**

`findLanguage(name)` only declares one parameter and reads `this.languages` internally, so the second argument is silently dropped — probably a leftover from the same confusion about scope:

```diff
-      const lang_description = this.findLanguage(language, this.languages);
+      const lang_description = this.findLanguage(language);
```

Thanks again for splitting this out!

evnchn:
Hi @Jepson2k — Claude Code here, on behalf of @evnchn

When you have a moment, could you rebase your open PRs (#6000, #5993, #5992, #5991, #5990, #5989, #5988, #5986, #5985, #5984, and this one) onto current `main`? Non-blocking — CI can be retried until green — but a rebase is the cleanest path to clear the current intermittent failures.

<details>
<summary>Why</summary>

These PRs branched before commit 985fecb ("Stop hitting third-party origins in absolute-URL navigation test", merged 2026-04-29), which replaced an `https://www.google.com/` navigation in `test_navigate_to_absolute_url` with a local `@ui.page('/external')`. The old version intermittently fails with `JavaScript console error: ... Refused to get unsafe header "X-Sg-Cs"` because Chrome surfaces SEVERE log entries from Google's page, and the Screen fixture treats those as failures.

After a rebase, the test no longer hits a third-party origin and the flake is gone.

</details>

Thanks!

falkoschindler (approved):
Thanks @Jepson2k!
Ready to merge 🚀

PR #5984, opened 2026-04-23
Add CodeMirror editor signals (selection, focus, viewport, geometry) and reveal_line

Jepson2k:
### Motivation

Hosts that embed `ui.codemirror` as an in-app editor often need feedback about the editor's state — cursor line / column, focus, the visible line range, and editor geometry. None of this is exposed through the Python API today, and there's no programmatic way to scroll a specific line into view.

pr-5984

### Implementation

- `reveal_line(line_number)` — scrolls a 1-indexed line into view via `EditorView.scrollIntoView`.
- Four typed signal handlers: `on_selection_change(line, column, from_line, to_line, empty)`, `on_focus_change(focused)`, `on_viewport_change(from_line, to_line)`, `on_geometry_change(width, height, content_height)`.

A single JS `ViewPlugin` reads the four CM6 `ViewUpdate` flags and emits per-signal events. Each signal has its own enable prop (auto-flipped when a handler is registered); when nothing is subscribed the dispatcher bails before computing a payload, so unsubscribed signals cost nothing on the wire. Payloads are deduped against the last one sent, so an update that leaves a signal unchanged never reaches the socket either.

Rate limiting is `Element.on(..., throttle=...)`, passed by the four registrars with per-signal defaults of 30 ms / none / 100 ms / 100 ms — the same approach `slider`, `range`, `knob` and `splitter` take with `throttle=0.05`. An earlier revision carried a bespoke `ui.codemirror.handler(cb, debounce_ms=...)` factory for this; it was removed in favour of the existing mechanism, which is per-listener rather than per-element and delivers the first event of a burst immediately.

No test covers the throttling itself: `tests/test_events.py` already exercises `throttle` / `leading_events` / `trailing_events`, and a per-element re-test would only be a slower, timing-dependent copy of it.

`geometry-change` reads the editor's size through `requestMeasure` rather than in `ViewPlugin.update`, which CodeMirror rules out: `geometryChanged` is set by every document change, so a registered geometry handler forced a layout on every keystroke. The `isConnected` guard in the write phase is load-bearing — `beforeUnmount` does not destroy the view, so a queued measure can outlive the editor's DOM and would otherwise report a 0×0 geometry. One consequence: the event now lands in a later measure cycle rather than in synchronous plugin-update order. No test covers the unmount path — reaching "view alive, DOM detached" needs scaffolding that mirrors the implementation, so `test_geometry_change_event` remains the coverage for the events themselves.

### Progress


falkoschindler:
Hi @Jepson2k, this branch has drifted out of sync with `main`: the conflicts are in `codemirror.js`, `codemirror.py` and `codemirror_documentation.py`, and they come mostly from your own line-anchor work — #5988 landed in 3.16, the follow-up #6282 moved the `on_anchor_change` dispatch to Python, and the "Added in" annotations were standardized in between — so the CodeMirror element has seen quite some motion since this branch was cut. Could you merge current `main` into the branch and resolve the conflicts (no rebase/force-push needed)?

Since #5984, #5985 and #5986 all touch the same spots, I'd suggest doing them one at a time rather than keeping all three in sync: pick the one you want to land first, bring it up to date, mark it "ready for review", and once it is merged, do the same for the next. That also keeps you well within the three-ready-PRs limit Evan mentioned. Note that #5991 (decoration API) is planned for 3.17 and lands in the same files, so the first one may need one more small update after that — sorry about the churn.

Until then I'm assigning this to you and marking it as draft; flip it back to "ready for review" whenever it is green and I'll take it from there. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Conflict%20analysis%20from%20the%20model%20%28merge-tree%20against%20main%29%2C%20wording%20from%20the%20model%2C%20substance%20and%20the%20sequencing%2Fdraft%20decision%20from%20the%20maintainer.&d=2026-08-19&a=claude-code "claude-fable-5: Conflict analysis from the model (merge-tree against main), wording from the model, substance and the sequencing/draft decision from the maintainer.")

evnchn:

I see you added `ui.codemirror.handler(cb, debounce_ms=...)` and `CodeMirrorHandlerSpec` to rate-limit the new events, but `Element.on(..., throttle=, leading_events=, trailing_events=)` already does this for every event in seconds, and in my eyes it covers everything yours does (checked: with the JS debounce at 0, `throttle=0.2, leading_events=False` coalesced 10 cursor moves at 30 ms into 2 events carrying the latest payload). If you agree, please remove it; if not, let's discuss.

Jepson2k:
*Posted by Claude Code on @Jepson2k's behalf.*

Agreed, and removed — net −106 lines.

Two reasons beyond the ones you gave: `throttle` keys on `listener_id`, so it's genuinely per-registration, whereas my `<signal>-debounce-ms` were element props that a second handler with a different value silently clobbered. And `leading_events=True` delivers the first cursor move immediately, where the trailing-only debounce added 30 ms of latency to every one.

The four `on_*_change` methods now pass `throttle=` with the old values as defaults (30 ms selection, 100 ms viewport and geometry, none for focus), the same shape as `slider.py`. I kept the JS payload dedupe and the `<signal>-tracking-enabled` props — both gate traffic at the source rather than rate-limiting it.

One call for you: since `on_*_change` flips the tracking props, a raw `.on('selection-change', …, throttle=…)` would never fire, so those defaults aren't overridable. `slider` has the same limitation so I left it — say the word if you'd rather have a `throttle` kwarg like `Joystick(throttle=…)`.

`test_debounce_override` is deleted with no replacement; `tests/test_events.py` already covers `throttle` / `leading_events` / `trailing_events`.

@falkoschindler — also merged current `main`. Your conflict report was accurate when you wrote it; I merged the day after, so `merge-tree` against today's `main` now reports no conflicts. Still outstanding was the `*Added in version 3.17.0*` annotations, which the new API was missing — those are in now.

nicegui/elements/codemirror/codemirror.js
```diff
@@ -239,6 +243,15 @@ export default {
       this._lastAnchors = positions;
       this.$emit("anchor-positions", { anchors: positions });
     },
+    revealLine(lineNumber) {
+      if (!this.editor) return;
+      const doc = this.editor.state.doc;
+      const lineNum = Math.max(1, Math.min(lineNumber, doc.lines));
```
evnchn, line 249: `reveal_line(2.5)` on a 3-line doc currently scrolls somewhere and logs nothing, and `line_tooltips` / `line_anchors` both warn for the same mistake. Measured after: 2.5, 0, -1, 500 and a Python `float('nan')` (arrives as `null`) all warn, none throw.

```suggestion
      if (!Number.isInteger(lineNumber) || lineNumber < 1 || lineNumber > doc.lines) {
        logAndEmit("warning", `reveal_line: line ${lineNumber} is not in [1, ${doc.lines}]`);
      }
      const lineNum = Math.min(Math.max(Math.trunc(lineNumber) || 1, 1), doc.lines);
```

nicegui/elements/codemirror/codemirror.js
```diff
@@ -319,6 +332,58 @@ export default {
...
+        class {
+          constructor() {
+            this._last = {};
+          }
+          update(u) {
+            // A focus transition makes selection state meaningful again: hosts that
+            // ignore unfocused selection events (programmatic echoes) must still hear
+            // about the first post-focus selection even if it matches the last payload.
+            if (u.focusChanged) delete this._last["selection-change"];
+            if (self.selectionTrackingEnabled && (u.selectionSet || u.docChanged)) {
```
evnchn, line 349: A server-side `editor.set_value()` with no user input emits a `selection-change`, and the payload carries nothing that lets a host tell it from a real cursor move. `changeSender` above already gates on `self.emitting` for exactly this. Measured: 1 event before, 0 after; user typing still emits.

```suggestion
            if (self.selectionTrackingEnabled && (u.selectionSet || (u.docChanged && self.emitting))) {
```

nicegui/elements/codemirror/codemirror.js
```diff
@@ -319,6 +332,58 @@ export default {
...
+              const sel = u.state.selection.main;
+              const line = u.state.doc.lineAt(sel.head);
+              this._maybeEmit("selection-change", {
+                line: line.number,
+                column: sel.head - line.from + 1,
```
evnchn, line 354: `column` is a UTF-16 code-unit offset, but the Python side indexes `str` by code point, which is why `_encode_codepoints` exists in `codemirror.py`. On `"\U0001F60Eabc"` a host slicing `line[:e.column - 1]` gets one character too many. Measured: cursor at end reported 6 before, 5 after; walking right from the line start now gives 1, 2, 3, 4 instead of 1, 3, 4, 5.

```suggestion
                column: Array.from(u.state.doc.sliceString(line.from, sel.head)).length + 1,
```

evnchn (commented):

I have spotted some issues. See if these suggestion blocks make sense to you. If not, I can explain further.

All three reproduced in a browser against 2f7acc5, fixed, and re-checked; `tests/test_codemirror.py` stays at 26 passed with all three applied.

evnchn:

Oh, and by the way — I noticed +0.27 ms/keystroke.

**TL;DR:**

Not a bug, and not part of my suggestion blocks. Take it or leave it.

- `geometry-change` reads `u.view.dom.clientWidth` / `clientHeight` inside `ViewPlugin.update`, which CodeMirror's docs specifically say not to do — "should *not* read the DOM layout".
- `geometryChanged` is true on every `docChanged`, so with a geometry handler registered that is a forced layout **per keystroke**. Measured at ~0.27 ms each.
- `throttle=0.1` on the Python side cannot help: the read happens client-side, before anything is emitted.

Small enough that ignoring it is a perfectly good answer — I am flagging it because it is invisible from the Python side, not because I think it blocks anything.

<details>
<summary><b>How it was measured</b></summary>

Two pages with identical DOM, the geometry handler being the only difference. `clientWidth`/`clientHeight` getters wrapped on `Element.prototype` to count accesses and time them, then 10 keystrokes typed at 60 ms intervals, 3 repetitions. Head 2f7acc5, Chromium via Playwright.

| | layout reads | time in those reads |
|---|---|---|
| no geometry handler | 606 | 0.2 / 0.7 / 0.6 ms |
| geometry handler | 626 | 2.6 / 3.7 / 2.9 ms |

The read *count* delta is exactly 20 = 2 per keystroke, which is the geometry payload. The cost delta lands entirely in those 20 reads: ~0.16 ms each against ~0.001 ms for CodeMirror's own 606, which is the signature of a forced layout rather than a cached one.

</details>

<details>
<summary><b>The patch, if you want it — and the trap in it</b></summary>

Move the measurement into CodeMirror's own DOM-read phase:

```js
            if (self.geometryTrackingEnabled && u.geometryChanged) {
              u.view.requestMeasure({
                key: "nicegui-codemirror-geometry",
                read: (view) => ({
                  width: view.dom.clientWidth,
                  height: view.dom.clientHeight,
                  content_height: Math.round(view.contentHeight),
                }),
                write: (payload) => {
                  if (u.view.dom.isConnected) this._maybeEmit("geometry-change", payload);
                },
              });
            }
```

After: 0.4 / 0.5 / 0.6 ms against a control of 0.4 / 0.5 / 0.1 ms — gone. Events still fire on content growth and on viewport resize, and the 26 tests in `tests/test_codemirror.py` stay green.

**The `isConnected` guard is load-bearing, and this is the part worth your attention.** A queued measure survives the editor's DOM being detached, because `beforeUnmount` never destroys the view. Control run, same probe, guard removed:

| | live edits | edit while the editor is detached |
|---|---|---|
| with guard | 2 events | nothing |
| without guard | 2 events | `0x0 content=59` |

So the naive version of this optimisation hands the host a 0×0 geometry payload on unmount. Current `main` has no such problem — it is introduced by the fix, which is a fair reason to decide the 0.27 ms is not worth it.

One more consequence either way: `geometry-change` moves from synchronous plugin-update order into a later measure cycle, so a host watching several signals at once can observe the ordering change.

The unmount case was caught by a second-lineage review (Codex), not by me; I verified it in the browser afterwards.

</details>

Jepson2k:
*Posted by Claude Code on @Jepson2k's behalf.*

All three taken, plus the geometry one.

- **`reveal_line`** — one word changed: `is not an integer in [1, N]`, since 2.5 genuinely *is* in [1, 3].
- **`selection-change` on `set_value`** — as written.
- **Code-point `column`** — as written. The real bug of the three.
- **The 0.27 ms** — taken too, `isConnected` guard included. The forced layout only exists because this PR adds `geometry-change`, so it's ours to fix rather than inherit. The measure-cycle ordering change is in the commit message.

No test for the unmount path: it needs the view alive with its DOM detached, which the public API doesn't expose without scaffolding that mirrors the implementation. Reason is in **Implementation**.

Each of the three regression tests was verified to fail with its fix reverted. Four commits, one per item.

evnchn:
**TL;DR:** where the host thinks the cursor is and where it actually is must never desync.
The `emitting` gate I asked for desyncs them, and nothing corrects it until the user happens to
move the cursor themselves. Three ways out below, none mine to pick.

Cursor on line 3, server inserts a line above it:

| | events emitted | host believes | cursor actually on | |
|---|---|---|---|---|
| `2f7acc51` parent | 1 | line 4 | line 4 | in sync |
| `1792efe3` head | **0** | **line 3** | line 4 | **desynced** |

A later real cursor move does resync it. Until then nothing can: `on_selection_change` is the
only channel reporting the cursor, and there is no Python-side getter to read instead.

What decides it: is a host meant to track the cursor across a `set_value`? If no, A is enough.
If yes, only B or C close the desync.

Diffs folded so the three read side by side, and not suggestion blocks on purpose: three
Commit buttons on a decision point is a way to decide by accident.

**A. Keep the gate, reword the docstring.** `codemirror.py` L138. Cheapest, and leaves the
desync in place.

<details><summary>diff</summary>

```diff
-        Fires on selection moves and on document edits that shift the cursor line or column.
+        Fires on selection moves and on user edits that shift the cursor line or column.
+        A server-driven ``set_value`` does not emit, even when it moves the cursor.
```

</details>

**B. Keep the gate, add `programmatic: bool` and emit again.** Echo-ignorers filter on it,
cursor-trackers stay in sync. Untested, and it joins `_maybeEmit`'s whole-payload comparison.

<details><summary>diff</summary>

```diff
--- a/nicegui/elements/codemirror/codemirror.js
-            if (self.selectionTrackingEnabled && (u.selectionSet || (u.docChanged && self.emitting))) {
+            if (self.selectionTrackingEnabled && (u.selectionSet || u.docChanged)) {
               const sel = u.state.selection.main;
               const line = u.state.doc.lineAt(sel.head);
               this._maybeEmit("selection-change", {
                 line: line.number,
+                programmatic: u.docChanged && !self.emitting,

--- a/nicegui/events.py
 class CodeMirrorSelectionChangeEventArguments(UiEventArguments):
     line: int
     column: int
+    programmatic: bool

--- a/nicegui/elements/codemirror/codemirror.py
             empty=e.args['empty'],
+            programmatic=e.args['programmatic'],
         )), throttle=0.03)
```

</details>

**C. Emit on a document edit only when the selection payload changed.** The only one I built,
and my first attempt was wrong: comparing the head offset misses a column change when `ab`
becomes `😎`, and a `to_line` change when a backward selection loses its middle line. Comparing
the whole payload against `u.startState` handles both. 27 passed, table back to in-sync, and
both counterexamples now emit correctly.

<details><summary>diff</summary>

```diff
-            if (self.selectionTrackingEnabled && (u.selectionSet || (u.docChanged && self.emitting))) {
-              const sel = u.state.selection.main;
-              const line = u.state.doc.lineAt(sel.head);
-              this._maybeEmit("selection-change", {
-                line: line.number,
-                column: Array.from(u.state.doc.sliceString(line.from, sel.head)).length + 1,
-                from_line: u.state.doc.lineAt(sel.from).number,
-                to_line: u.state.doc.lineAt(sel.to).number,
-                empty: sel.empty,
-              });
+            if (self.selectionTrackingEnabled && (u.selectionSet || u.docChanged)) {
+              const payload = (state) => {
+                const sel = state.selection.main;
+                const line = state.doc.lineAt(sel.head);
+                return {
+                  line: line.number,
+                  column: Array.from(state.doc.sliceString(line.from, sel.head)).length + 1,
+                  from_line: state.doc.lineAt(sel.from).number,
+                  to_line: state.doc.lineAt(sel.to).number,
+                  empty: sel.empty,
+                };
+              };
+              const now = payload(u.state);
+              if (u.selectionSet || JSON.stringify(now) !== JSON.stringify(payload(u.startState))) {
+                this._maybeEmit("selection-change", now);
+              }
             }
```

<details><summary>the two cases that killed my first attempt</summary>

Head-offset comparison. Both emit nothing, both leave the host wrong:

| | host believed | truth after the edit | emitted |
|---|---|---|---|
| `ab` before the cursor becomes `😎` | column 3 | column 2 | 0 |
| backward selection over 3 lines, middle one deleted | `to_line` 3 | `to_line` 2 | 0 |

The head offset is UTF-16 so it does not change in the first, and the head does not move at all
in the second. With the payload comparison both emit, `(1, 2, 1, 1, True)` and
`(1, 1, 1, 2, False)`.

</details>

</details>

I asked for the gate, so which one lands is your call and @falkoschindler's.

<details>
<summary><b>One dead end, in case you try the same thing</b></summary>

Dropping `&& self.emitting` and leaning on the existing `_maybeEmit` dedupe looks like C for
free. It fails your test with `assert 2 == (0 + 1)`: `_last["selection-change"]` is still empty
there, so the first echo has nothing to dedupe against. The comparison has to be against
`u.startState`.

</details>

<details>
<summary><b>The probe, run on both trees</b></summary>

```python
def test_cursor_after_set_value(screen: Screen):
    events, editor = [], None

    @ui.page('/')
    def page():
        nonlocal editor
        editor = ui.codemirror('L1\nL2\nL3',
                               on_selection_change=lambda e: events.append(e.line))

    screen.open('/')
    screen.should_contain('L3')
    screen.selenium.execute_script(
        f'const el = getElement({editor.id});'
        'el.editor.dispatch({selection: {anchor: el.editor.state.doc.line(3).from}});')
    screen.wait_for(lambda: 3 in events)
    before = len(events)

    editor.set_value('NEW\nL1\nL2\nL3')
    screen.should_contain('NEW')
    time.sleep(1.0)                      # give any echo time to arrive

    actual = screen.selenium.execute_script(
        f'const el = getElement({editor.id});'
        'const s = el.editor.state; return s.doc.lineAt(s.selection.main.head).number;')
    print(f'emitted={len(events) - before} host={events[-1]} actual={actual}')
```

`parent: emitted=1 host=4 actual=4` · `head: emitted=0 host=3 actual=4`

</details>

CI has not run since `2f7acc51`, so today's four commits have not been through it. I merged
`main` locally to check the above, and both conflicts are keep-both, in case it saves you a look:

- the docstring paragraph that landed beside the Decorations one
- #6336 moving the Line Anchors demo to the end of the file


---

<sub>Colophon: drafted by Claude Code working with @evnchn. Figures are medians of repeated runs
through the repo's `screen` fixture, head `1792efe3` against parent `2f7acc51`. I have moved the
attribution to the foot so the first line can be about the work. Adopt it or not, no rule here.</sub>

nicegui/elements/codemirror/codemirror.js
```diff
@@ -319,6 +335,69 @@ export default {
...
+            // A focus transition makes selection state meaningful again: hosts that
+            // ignore unfocused selection events (programmatic echoes) must still hear
+            // about the first post-focus selection even if it matches the last payload.
+            if (u.focusChanged) delete this._last["selection-change"];
+            if (self.selectionTrackingEnabled && (u.selectionSet || (u.docChanged && self.emitting))) {
+              const sel = u.state.selection.main;
+              const line = u.state.doc.lineAt(sel.head);
+              this._maybeEmit("selection-change", {
+                line: line.number,
+                column: Array.from(u.state.doc.sliceString(line.from, sel.head)).length + 1,
```
evnchn: OPTIONAL, ignoring it is a perfectly good answer. `Array.from()` made `column` O(line length):
0.565 ms per selection change on a 200 000-char line against 0.106 before, 0.112 with this.
Long lines only, so minified, JSON and logs.

```suggestion
              const line = u.state.doc.lineAt(sel.head);
              const prefix = u.state.doc.sliceString(line.from, sel.head);
              this._maybeEmit("selection-change", {
                line: line.number,
                // Array.from() allocates one entry per code point; only pay that on a surrogate.
                column: (/[\uD800-\uDBFF]/.test(prefix) ? Array.from(prefix).length : prefix.length) + 1,
```

9/9 cases unchanged, 27 passed. Goes stale if C lands first, where it is the `column:` line plus
its `prefix` declaration.

<details>
<summary><b>UTF-16 against UTF-8, and why the surrogate check is exact</b></summary>

UTF-16 against UTF-8: a JS string is a UTF-16 sequence, while Python counts the code points it
holds and encodes to UTF-8 on the way out. Three numbers for one character:

| | UTF-8 bytes | UTF-16 units, JS `.length` | code points, Python `len()` |
|---|---|---|---|
| `a` | 1 | 1 | 1 |
| `中` | 3 | 1 | 1 |
| `😎` | **4** | **2** | **1** |

`column` has to be the right-hand number, because that is what indexes a Python `str`. The
original `sel.head - line.from` gave the middle one. They diverge only above U+FFFF.

Above U+FFFF a character is stored as a high surrogate (U+D800-U+DBFF) plus a low one, and that
is the only way one reaches a JS string. So no high surrogate in the prefix means code units
equal code points, which is all the fast path needs. That implication only runs one way, which
is fine: a lone high surrogate takes the slow branch and `Array.from` returns the same number.

Code points are still not what a reader counts as a character: written decomposed as `e` +
U+0301, `é` is 2 in both languages. `column` matches Python, which is what it is for. None of
this is something the PR was missing.

</details>

---

<sub>Colophon: drafted by Claude Code working with @evnchn. Timings are medians of repeated runs
on one 200 000-character line, same fixture and same two trees as the other comment.</sub>

Jepson2k:
*Posted by Claude Code on @Jepson2k's behalf.*

C, pushed as `32b44c22`. You were right both times — the echo was real, and so was the desync the gate traded it for. The payload-vs-`startState` comparison is the thing that answers the actual question; `emitting` only ever meant "the server already knows the text".

`test_set_value_emits_only_when_the_cursor_actually_moves` pins it from both sides: it fails against the gate on the remap, and against the ungated original on the echo. Your two counterexamples are why it compares payloads rather than the head offset.

Also merged `main` — both conflicts keep-both, as you said. CI is running on these five commits now.

nicegui/elements/codemirror/codemirror.js
```diff
@@ -514,6 +530,78 @@ export default {
...
+            // CodeMirror forbids DOM layout reads in update(), and geometryChanged is set by every
+            // document change, so reading clientWidth/clientHeight here forces a layout per keystroke.
+            // requestMeasure defers the read into CM's own measure cycle, where it is already batched.
+            if (self.geometryTrackingEnabled && u.geometryChanged) {
+              u.view.requestMeasure({
+                key: "nicegui-codemirror-geometry",
+                read: (view) => ({
+                  width: view.dom.clientWidth,
+                  height: view.dom.clientHeight,
+                  content_height: Math.round(view.contentHeight),
```
evnchn: `content_height` comes out in scaled px while its two siblings are layout px: at `scale(0.5)` it reports 43 where 86 is correct, and `width`/`height` stay at 724/128.

image

```suggestion
                  content_height: Math.round(view.contentHeight / view.scaleY),
```

nicegui/elements/codemirror/codemirror.js
```diff
@@ -434,6 +438,18 @@ export default {
       this._lastAnchors = positions;
       this.$emit("anchor-positions", { anchors: positions });
     },
+    revealLine(lineNumber) {
+      if (!this.editor) return;
+      const doc = this.editor.state.doc;
+      if (!Number.isInteger(lineNumber) || lineNumber < 1 || lineNumber > doc.lines) {
+        logAndEmit("warning", `reveal_line: line ${lineNumber} is not an integer in [1, ${doc.lines}]`);
+      }
```
evnchn, line 446: Nit: an out-of-range integer reports `is not an integer`, which reads as a type error; `setLineTooltips` already says `out of range`. The non-integer string stays, since `test_reveal_line` asserts it.

```suggestion
      if (!Number.isInteger(lineNumber)) {
        logAndEmit("warning", `reveal_line: line ${lineNumber} is not an integer in [1, ${doc.lines}]`);
      } else if (lineNumber < 1 || lineNumber > doc.lines) {
        logAndEmit("warning", `reveal_line: line ${lineNumber} out of range [1, ${doc.lines}]`);
      }
```

nicegui/elements/codemirror/codemirror.py
```diff
@@ -76,10 +85,16 @@ def __init__(
         *Since version 3.17.0:*
         Decorations style, hide or annotate parts of the document without changing it.
         Assign a list of specs to ``decorations`` or mutate ``decorations`` in place.
+        Editor signals report the cursor selection, focus, visible line range and geometry,
+        and ``reveal_line`` scrolls a given line into view.
```
evnchn, line 89: Nit: this sentence reads as part of the Decorations paragraph; a blank line puts it on its own.

```suggestion
        Assign a list of specs to ``decorations`` or mutate ``decorations`` in place.

        Editor signals report the cursor selection, focus, visible line range and geometry,
        and ``reveal_line`` scrolls a given line into view.
```

Jepson2k:
*Posted by Claude Code on @Jepson2k's behalf.*

Both in.

**`content_height`** — this is a real bug in the API this PR adds, not a nitpick. Fixed as suggested, and `test_geometry_change_event` now pins it: a CSS transform changes neither the document nor the space it is laid out in, so all three numbers have to sit still. It fails `assert 14 == 28` without the divide.

**`Array.from`** — taken. Worth noting your measurement understates it now: C builds the payload twice per update to compare against `startState`, so the cost had doubled since you timed it. The surrogate check is exact where it needs to be, and a lone high surrogate taking the slow branch costs nothing.

Two commits, green on the codemirror and events suites.

evnchn (approved):
Approving. I checked that the new test fails without the divide (`assert 14 == 28`) and that the value holds at `scale(0.5)`, `scale(2)`, `scale(1.7)`, `scaleY(0.25)` and a scaled parent. The two suggestions are optional, resolve them if you would rather not take them.

[※](https://zauberzeug.github.io/colophon/#m=deepseek-v4.1-flash&t=Checks%20and%20wording%20from%20opencode%2C%20re-checked%20in%20Claude%20Code%20%28claude-opus-5%29%3B%20evnchn%20chose%20the%20checks%20to%20run.&d=2026-09-22&a=opencode "deepseek-v4.1-flash: Checks and wording from opencode, re-checked in Claude Code (claude-opus-5); evnchn chose the checks to run.")

PR #5988, opened 2026-04-24
Add line-anchor tracking to CodeMirror

Jepson2k:
### Motivation

Line anchors give you a more stable reference to specific lines than line numbers. CodeMirror knows where each anchor moved to as the user types — paste 30 lines above an anchor and it follows; delete the line it sits on and it slides to the next one (an anchor only disappears when a deletion runs across it). Useful when the host application needs a handle to a specific line that survives edits without recomputing positions in Python on every keystroke.

The new demo on the docs page shows the idea: two lines are tagged with anchors "A" and "B", and a pair of buttons rewrites whichever line the anchor currently points at. You can edit the code freely — insert blank lines, reformat, reorder — and clicking "Update" still hits the right line.

https://github.com/user-attachments/assets/62b74fdc-1306-473e-857e-778b1f898192

### Implementation

- `editor.line_anchors` is a `{id: line}` dict; assigning it (or passing it to the constructor) syncs to the client. An optional `on_anchor_change=` handler fires whenever tracked positions change.
- `editor.line_anchor_positions` reflects the current `{id: line}` mapping. The browser is the source of truth — anchors are remapped through edits by CM6's `RangeSet.map()` and the snapshot is pushed back after every state change.

### Progress


evnchn (dismissed):

Thanks so much for this, @Jepson2k — line anchors are a genuinely useful addition, and the test suite is a real highlight (behavior-level Screen tests, no mocks, and the debounce-resurrection test shows you thought carefully about the tricky cases). I ran the PR locally and dug into the anchor mapping with a few minimal repros. The core is solid; there are a few things to sort before it lands, and I've tried to keep each one concrete and small. One is a blocker, but it's a contained fix — happy to pair on any of these.

**🔴 The one blocker — a single unrelated prop update wipes the tracked positions.** Any server-side prop change re-broadcasts *all* props, which re-fires the deep `lineAnchors` watcher and re-runs `applyLineAnchors([A@1, B@2])` from the original creation-time `line` field (`codemirror.js` ~L84-89, ~L218-220), discarding the live remapped positions. Concretely, after an edit has remapped anchors to A:2/B:3: setting `editor.theme = 'oneDark'` or `editor.set_language('Python')` snaps them back to `{A:1, B:2}`, and `editor.value = ...` corrupts the set differently (I saw `{A:1}`, with B dropped). The docs demo writes the value after the first action, so the anchors break on the second use. The fix is contained: instead of re-applying from props on every broadcast, deep-compare the incoming list against a stored `_lastAppliedAnchors` snapshot and no-op when it's unchanged (a deliberate reassignment still applies normally).

**🟠 Deleting the *last* anchor doesn't notify Python.** `anchorTracker` early-returns when the field size is 0 (`codemirror.js` ~L279), which swallows the 1→0 transition, so `line_anchor_positions` stays stale afterwards (I confirmed it survives the deletion plus two further edits with no emit). With two or more anchors a partial drop emits correctly — it's only the last-one case. One-line fix: also check `update.startState.field(anchorField).size === 0` before returning.

**🟠 A small docs mismatch on deletion.** The description says "delete the line and it's gone," but anchors are zero-width ranges whose `mapMode` defaults to `MapMode.TrackDel`, and per the CM6 docs TrackDel only drops a position when a deletion happens *across* it ([`RangeValue.mapMode`](https://codemirror.net/docs/ref/#state.RangeValue.mapMode): "Defaults to MapMode.TrackDel"; [`MapMode.TrackDel`](https://codemirror.net/docs/ref/#state.MapMode): "Return null if deletion happens across the position"). A natural full-line delete starts *at* the anchor's leading edge, not across it, so the anchor slides to the next line rather than vanishing. Just worth matching the docs to that (or choosing a different `mapMode` if "gone on delete" is the intent).

**A few minor things,** none blocking:
- the three `*Added in version X.Y.Z*` placeholders need a real version;
- `id` is typed `str` but `{'id': 1}` round-trips as `'1'` (JSON keys + `str()` coercion);
- `_update_anchor_mirror` guards that the payload is a dict but not the value types, so a malformed `{'a': 'x'}` would make `int(v)` raise;
- with a shared element (script-mode page / multiple tabs) the single Python mirror is last-writer-wins — worth a docstring sentence.

**And one optional API thought** (entirely up to you): `line_anchors=[{'id': 'A', 'line': 1}, ...]` could become `line_anchors={'A': 1}` — it'd match the sibling `line_tooltips: dict`, mirror the `line_anchor_positions: dict` output, and remove the duplicate-id handling entirely. Is there an ordering need I'm missing? Similarly, the demo teaches the raw event string (`editor.on('anchor-positions', ...)`); an `on_anchor_change=` handler param (like `on_change`) would keep that out of user code.

Good news on the build side — no dist rebuild needed, the bundle already re-exports the CM APIs you use. I have small runnable repros for each of these if they'd help — happy to drop them in a gist. The two JS fixes are small, and I'm glad to take another quick pass once they're in. Really nice first cut. 🙂

Jepson2k:
ok I believe I've address the latest feedback in my latest commits including the suggest dict type change. The failing test appears to be a flaky one related to the tree component. It doesn't appear that I have the ability to re-run the tests, but I'd appreciate a re-run whenever possible. Thank you

evnchn:

**TL;DR:**

All my 06-14 review points are addressed at `fa656cda` — the 🔴 blocker, both 🟠 fixes, the nits, and both 💡 optional suggestions (dict API + `on_anchor_change`). Checked against the diff, not the commit message.

A holistic re-review with end-to-end verification follows in a separate comment.

<details>
<summary><b>Per-point verification (diff at <code>fa656cda</code>)</b></summary>

| Review point | Status | Evidence |
|---|---|---|
| 🔴 Unrelated prop re-broadcast re-fired the deep `lineAnchors` watcher and re-applied from declared lines, wiping remapped positions | ✅ | `applyLineAnchors()` computes a `JSON.stringify(Object.entries(anchors).sort())` snapshot, stores `_lastAppliedAnchors`, and no-ops when unchanged; a genuine reassignment still applies. Test `test_anchor_positions_survive_unrelated_prop_update`. |
| 🟠 Last-anchor deletion (1→0) didn't notify | ✅ | `anchorTracker` now skips only when **both** `startState` and `state` field sizes are 0. Test `test_last_anchor_deletion_notifies`. |
| 🟠 Docs vs TrackDel slide/vanish semantics | ✅ | `line_anchors` docstring: dropped only when a deletion spans across the position; a full-line delete slides to the next line. Matches CM6 default `MapMode.TrackDel`. |
| ⚪ `X.Y.Z` version placeholders | ✅ | All `3.14.0` (commit `fa656cda` fixed a 3.13.0 slip). |
| ⚪ `id` str/int coercion | ✅ | `{str(k): int(v) …}` in init + setter. |
| ⚪ malformed value crashes `int()` | ✅ | `_update_anchor_mirror` try/except per anchor, warns + skips. |
| ⚪ shared-element last-writer-wins | ✅ | Documented in `line_anchor_positions`. |
| 💡 dict API `{'A': 1}` | ✅ adopted | `line_anchors: dict[str, int]`; the list-of-dicts proxy is gone. |
| 💡 `on_anchor_change=` handler | ✅ adopted | New param + method + `CodeMirrorAnchorChangeEventArguments`. |

Nicely done, @Jepson2k.
</details>

evnchn:

**TL;DR:**

Not approving solo — I'd like @falkoschindler's eyes before this merges, given the CodeMirror position-mapping history (#4676 / #4678 / #4702).

That said, my end-to-end testing says the anchor machinery is solid. I exercised the core behaviors (remap, TrackDel slide, vanish-on-delete-across, the 1→0 notify, the blocker regression) **and** the historically-risky edit families that motivated my caution — UTF-16 / astral text, multi-cursor, and undo. All hold up (tables below).

So this is a confident "looks right to me, second pair of eyes for the merge call" — not a list of open worries.

<details>
<summary><b>Core behaviors — verified end-to-end (branch at <code>fa656cda</code>, real browser)</b></summary>

Demo with `line_anchors={'a': 3}` + an `on_anchor_change` handler mirroring `line_anchor_positions` into a label; drove real CM6 document edits via `view.dispatch(...)` and read the server mirror after each (WS round-trip + 50 ms debounce):

| Edit | Mirror result | Confirms |
|---|---|---|
| insert a line above the anchor | `{'a': 4}` | remap up |
| delete that line | `{'a': 3}` | remap down |
| **full-line delete of the anchor's own line** | `{'a': 3}` (now points at old `line4`) | **TrackDel slide**, not vanish |
| insert 2 lines on top → `a`=5, then **toggle an unrelated prop** (`line_wrapping`) | `{'a': 5}` (not reset to declared 3) | **🔴 blocker fix** — re-broadcast doesn't wipe remapped positions |
| deletion spanning **across** the anchor | `{}` | vanish-on-delete-across **and** the 🟠 last-anchor 1→0 notify (mirror emptied) |

</details>

<details>
<summary><b>Unicode / UTF-16 (the #4678 corner) — robust</b></summary>

Re-ran the harness with emoji (astral / surrogate-pair), CJK, and combining-mark content:

| Edit | Mirror | Confirms |
|---|---|---|
| insert astral-only line above the anchors | `{'a': 4, 'b': 6}` | line counting unaffected by surrogate pairs |
| **prepend 🚀🚀 (4 UTF-16 code units) with no newline** | `{'a': 3, 'b': 5}` unchanged | **line-based anchors are immune to UTF-16 offset shifts** |
| full-line delete of a CJK anchor line | `{'a': 3, 'b': 4}` (a slid to `café delta`) | TrackDel slide with non-ASCII |
| insert 🔥🔥 mid anchor-line (`ca🔥🔥fé delta`) | `{'a': 3, 'b': 4}` unchanged | a mid-line astral edit doesn't move the line number |
| delete across an anchor through astral content | `b` dropped, `a` kept | vanish works with surrogate pairs in range |

Root reason it's safe: anchors are reported as **line numbers** (`doc.lineAt().number`), never char offsets — so the JS-UTF-16-vs-Python-codepoint mismatch that bit #4678 has no surface here.
</details>

<details>
<summary><b>Multi-cursor / multi-change (#4702) + undo — robust, one semantic to note</b></summary>

Multi-change (single transaction):

| Edit | Mirror | Confirms |
|---|---|---|
| two `\n` inserts on different lines above the anchor | `a` += 2 | multi-change `ChangeSet` remaps correctly |
| two astral inserts, no newlines | unchanged | multi-change + UTF-16 together |

Undo: `Cmd-Z` after a delete-across remaps `a` correctly through the inverse change (no crash). **But** an anchor dropped by a delete-across is **not resurrected** by undo — the `RangeSet` no longer contains it. Defensible (a deleted anchor stays deleted), but a user might expect undo to restore it; possibly worth a docstring line if you find it surprising.
</details>

<details>
<summary><b>How I drove it (reproducible — replay against any checkout of this branch)</b></summary>

```python
# anchor_demo.py
from nicegui import ui
mirror = ui.label('mirror: {}').props('id=mirror')
editor = ui.codemirror('line1\nline2\nline3\nline4\nline5',
                       line_anchors={'a': 3},
                       on_anchor_change=lambda e: mirror.set_text(f'mirror: {e.anchors}')).classes('h-40')
ui.button('toggle wrap', on_click=lambda: setattr(editor, 'line_wrapping', not editor._props['line-wrapping'])).props('id=wrap')
ui.run(port=8233, show=False, reload=False)
```
```bash
PYTHONPATH=<worktree-at-fa656cda> <venv>/bin/python anchor_demo.py
```
In the browser, `getElement('<editor-id>').editor` is the `EditorView`; edits via `view.dispatch({changes: {from, to, insert}})` (an array of changes for the multi-cursor case), positions read from the `#mirror` label after ~400 ms. Anchor pos sourced from `doc.line(n).from`; "delete across" = a change range that strictly contains the anchor position. Unicode runs used 😀/🚀/🔥 (astral), 中文 (CJK), and `café` (combining mark).
</details>

<details>
<summary><b>Code-quality pass (no blockers)</b></summary>

- `AnchorValue extends RangeValue` with id-based `eq`; zero-width point ranges in a `StateField`, remapped via `set.map(tr.changes)` — idiomatic CM6, relying on the default `MapMode.TrackDel` for the slide/vanish semantics (which the runs above confirm).
- The deep-compare snapshot is the right shape for the blocker; `_anchorTimer` is cleared on `destroyed()`.
- 50 ms debounce coalesces bursts; the fire-time callback reads live state, so a stale timer still emits correct positions.
- `_update_anchor_mirror` defensively warns+skips malformed payloads; `line_anchor_positions` returns a copy.
- 9 new tests, including both regression encodings.
</details>

On CI: the only red was an unrelated flaky `test_tree.py` (not in this PR's file set, untouched since Feb); I re-triggered it and it's re-running at the time of this edit.

Sibling #6000 (now approved) touches the same `codemirror.py`, so whichever merges first will want a quick rebase. Thanks @Jepson2k — the implementation itself looks careful.

---
*Changelog*
- 2026-06-18: Added empirical Unicode/UTF-16 (#4678), multi-cursor (#4702), and undo results — all hold up; updated the TL;DR from "didn't exercise those corners" to "exercised, robust". Merge call still deferred to @falkoschindler.

falkoschindler:
Really nice feature — I went through parts of it and pushed a few small, behavior-preserving cleanups to the branch:

- **Docstrings** now follow CONTRIBUTING's one-sentence-per-line rule.
- **Shared the RangeSet scaffold.** `AnchorValue`/`anchorField` and `TooltipValue`/`tooltipField` had two near-identical `StateEffect` + `StateField` blocks; extracted a `defineRemappableRangeSet()` factory so both reuse one definition (and the two `update()` bodies no longer drift apart).
- **Trust declared types.** Dropped the defensive `str()`/`int()` coercion from the `line_anchors` constructor/setter and the `isinstance`/try-except validation from the inbound `_update_anchor_mirror` — the typed API and our own `emitAnchorPositions` already guarantee the shapes, matching how the sibling `line_tooltips` stores its dict verbatim (also removed the now-unused `log` import).
- **Event subscription for handlers.** `on_anchor_change` now registers via `self.on('anchor-positions', ...)` like `Button.on_click`, instead of maintaining a private handler list and dispatching to it manually; `_update_anchor_mirror` is reduced to just keeping the mirror current.

These are all small and self-contained — please sanity-check them, and the Selenium `tests/test_codemirror.py` is the right gate since a couple touch the tooltip path too.

---

One design thought on the anchor sync — not a blocker, more a "there might be a cleaner foundation now" note. Feel free to take it on, or leave it for a follow-up.

**The current `_lastAppliedAnchors` guard is a smart fix for a real problem, but it leaves one ambiguity unresolved.** Every server-side prop change re-broadcasts _all_ props, which re-fires the deep `lineAnchors` watcher with the originally declared lines. Without the guard, an unrelated change like `editor.theme = 'oneDark'` would re-run `applyLineAnchors` and snap every anchor back to its declared line, destroying the live remapped positions. The snapshot guard correctly suppresses that.

The catch is that a _deliberate_ re-assignment and an _unrelated_ re-broadcast deliver a byte-identical input to `applyLineAnchors` — both arrive as `{'a': 5}`. The guard can only compare values, so it can't tell them apart and no-ops both. The consequence: re-asserting an unchanged declaration does nothing, so there's no way to snap an anchor back to its declared line, or to restore an anchor that was dropped (deletion across it) or clamped (line > doc length, then the doc grew). The workaround is to perturb the value (`= {}` then `= {'a': 5}`), which is non-obvious.

**Now that #6104 (preserve client-originated values in element updates) is on `main`, there's a cleaner foundation for this.** Its `preserved_props` mechanism lets the server tell the client "keep your current value for this prop on an update" — the client skips patching it, so the component's internal state survives. The JS half is prop-name-agnostic, so it applies to `line-anchors` directly: marking `line-anchors` as preserved means the deep watcher never fires on a spurious re-broadcast, and the live positions in the CM `anchorField` are kept — which is exactly what `_lastAppliedAnchors` does today, but via the framework instead of a bespoke guard. That guard could then go away.

**One adaptation is needed.** #6104 decides preserve-or-not by _value comparison_ (`outgoing == client's last emit`), which works for `ValueElement` because the live client state _is_ the prop. For anchors the live state lives in the `anchorField` and comes back via `anchor-positions` events — it's not the `line-anchors` prop, which always holds the declared lines — so there's nothing to compare. Instead the decision should be driven by _intent_: preserve `line-anchors` on every update **except** when the `line_anchors` setter/constructor was actually invoked since the last send. That intent signal (which value-comparison can't see) is exactly what resolves the ambiguity above — an unrelated prop change preserves and keeps live positions, while a deliberate `editor.line_anchors = {'a': 5}` is _not_ preserved and re-applies, even when the value is unchanged.

Rough shape:

```python
# setter / constructor
self._props['line-anchors'] = anchors or {}
self._apply_anchors = True  # mark deliberate (re)assignment

# serialization
def _to_dict(self) -> dict[str, Any]:
    dict_ = super()._to_dict()  # ValueElement may already add VALUE_PROP
    if not self._apply_anchors:
        dict_.setdefault('preserved_props', []).append('line-anchors')
    self._apply_anchors = False
    return dict_
```

Worth sanity-checking that mutating a flag inside `_to_dict` is safe (i.e. `_to_dict` is only called for outgoing sends), and that extending — not replacing — `ValueElement`'s `preserved_props` list composes correctly. `test_anchor_positions_survive_unrelated_prop_update` already covers the preserve path; a deliberate-re-assign-snaps-back test would cover the other direction.

falkoschindler:
One more, on the API surface: what's the intended purpose of having both `line_anchors` and `line_anchor_positions`?

As far as I can tell there are really two operations a user needs — **declare** anchors (the setter) and **observe** where they currently are (`line_anchor_positions`). The piece I'm unsure about is the `line_anchors` *getter*: it returns the original declared seed, which goes stale the moment the document is edited, so it rarely matches what you'd actually want to read. And because it looks like it should return the current state, it's easy to reach for by mistake.

Would it make sense to merge the two into a single `line_anchors` — assign it to declare, read it back for the live positions — and drop `line_anchor_positions`? The only thing that buys the split is a synchronous read-back of exactly what you just declared, which I'm struggling to find a use for (the live read is the interesting one, and it's async by nature anyway).

Is there a use case for reading the declared values back that I'm missing?

Jepson2k:
@falkoschindler I like both suggestions and incorporated the new preserved_props mechanism and merged the setter and getter into line_anchors.

evnchn:

**TL;DR:** Re-reviewed the delta since my 06-18 pass — @Jepson2k's `preserved_props` migration + getter merge (`192a632e`) and @falkoschindler's follow-up cleanups incl. the mixin extraction (91b7a6d9). Clean improvement: the API simplification is a real win, the mixin is behavior-preserving, and I couldn't break the flag mechanism (incl. the `_to_dict` mutation you flagged, @falkoschindler — safe, analysis folded below). **LGTM** — one small test worth adding, otherwise good to land from my side (merge call still yours).

Local gates green: pre-commit ✓, mypy ✓, pylint 10.00 ✓, `test_codemirror_line_anchors.py` 8✅, sibling `test_codemirror.py` 19✅.

### 🟡 Missing test for the snap-back direction (the one actionable item)

@falkoschindler you noted *"a deliberate-re-assign-snaps-back test would cover the other direction."* Still the one real gap: the preserve direction is covered (`test_anchor_positions_survive_unrelated_prop_update`), but the deliberate-reassign-**same-declared-value** → snap-back direction (the whole reason the intent flag exists — value-comparison can't see it) isn't. I verified it works empirically (remap `3→4`, reassign `{mid: 3}`, snaps back to 3), so it's coverage, not a bug — but it's the direction most likely to regress if the flag lifecycle is ever touched.

<details>
<summary>Ready-to-drop test (passes on the current branch, ~1.9s)</summary>

```python
def test_reassign_same_declared_value_snaps_back(screen: Screen):
    editor = None
    @ui.page('/')
    def page():
        nonlocal editor
        editor = ui.codemirror('a\nb\nc\nd\ne')
    screen.open('/')
    _wait_for_editor(screen)
    editor.line_anchors = {'mid': 3}
    screen.wait_for(lambda: editor.line_anchors.get('mid') == 3)
    screen.selenium.execute_script(
        f'const el = getElement({editor.id});'
        'el.editor.dispatch({changes: {from: 0, insert: "X\\n"}});')
    screen.wait_for(lambda: editor.line_anchors.get('mid') == 4)  # remapped
    editor.line_anchors = {'mid': 3}                              # deliberate reassign, SAME value
    screen.wait_for(lambda: editor.line_anchors.get('mid') == 3)  # snaps back
```
</details>

<details>
<summary><b>✅ Flag-in-<code>_to_dict</code> safety — the sanity-check you asked for (@falkoschindler)</b></summary>

> *"worth sanity-checking that mutating a flag inside `_to_dict` is safe (i.e. `_to_dict` is only called for outgoing sends)."*

I chased this hard (it's the one spot where the migration adds server-side state). **It holds up.** `_apply_anchors` is set by the setter and consumed by `_to_dict()`, which has two callers — the per-update outbox flush (`outbox.py`) and the full page-build (`client.py`) — but on a *live, socket-connected* client the reassignment is consumed by exactly **one** `_to_dict()` (the single outbox flush, deduped by element id). The full page-build can't race it: each page GET (incl. root/auto-index) builds a **fresh `Client` with fresh element instances** (`page.py` `with Client(self, request=request)`), so there's no shared live element another viewer could serialize; and `build_response()` clears `outbox.updates` *before* it serializes anyway. So no reachable double-consumption → the deliberate reassignment isn't dropped.

**Dead end worth recording:** I first suspected a race — a second viewer's page-build `_to_dict()` eating the flag before the pending outbox update to an open tab flushes (→ that update serializes with `preserved_props` and the tab silently ignores the reassignment). Built an MRE, but the "production" framing was **unfaithful**, confirmed via a second-lineage pass over `client.py`/`page.py`/`outbox.py`: fresh `Client` per GET (no shared live element across viewers; root routes through the same per-request machinery via `page('')._wrap(root)`), and `build_response()` does `self.outbox.updates.clear()` *before* the full-build. So the only live-element caller of `_to_dict()` with a pending reassignment is the outbox flush, once. Safe. (Recording so nobody re-walks it.)

**One latent note, not a blocker:** `_to_dict()` is a serializer with a side effect (resetting `_apply_anchors`). Safe under today's per-request-client model, but it would silently break if `_to_dict()` were ever called speculatively on a live element (debug/repr, a future shared-client or resync path). A one-line comment on the setter/`_to_dict` — "correctness relies on exactly one outgoing serialization per deliberate reassignment" — would save the next person the walk. Take-or-leave.
</details>

<details>
<summary>✅ Also confirmed correct (so you don't re-verify)</summary>

- **`preserved_props` composes with `ValueElement`.** MRO `CodeMirror → KeyBindingElement → LineAnchorElement → ValueElement → …`; `ValueElement._to_dict` hard-assigns `['value']`, `LineAnchorElement` (outer) appends via `setdefault(...).append('line-anchors')` → `['value','line-anchors']`. (Only fragile if the bases were ever reordered so `ValueElement` ran outermost — its hard-assign would clobber. Fine as-is.)
- **Fresh client applies declared anchors** even if its build carries `preserved_props` — the `nicegui.js` `if (this.elements[id] && …)` guard skips preservation when the element doesn't exist yet.
- **Getter merge is clean** — no dangling `line_anchor_positions` references in library/tests/docs.
- **Mixin extraction behavior-preserving** — the shared `defineRemappableRangeSet` scaffold keeps the tooltip path green.
</details>

Nice work both of you — solid feature.

falkoschindler:
Sorry, we didn't manage to land this in 3.15 either, so it has to move once more — but we'll firmly schedule it for 3.16. Thanks for bearing with us.

falkoschindler (approved):
Thanks for sticking with this, @Jepson2k — line anchors are a genuinely useful addition, and the foundation you picked (a remappable `RangeSet` whose positions the browser owns, with the server holding a mirror) turned out to be exactly right. Everything below builds on that.

Your fixes from @evnchn's round all check out: the last-anchor deletion notifies Python (the `startState` size check, with a test), the deletion wording matches `MapMode.TrackDel`, and both API suggestions landed — `line_anchors` is a `{id: line}` dict matching `line_tooltips`, and `on_anchor_change=` is a constructor parameter, so the demo no longer teaches the raw event string. The shared-element caveat is moot on 3.x, where every request gets a fresh client.

Rather than send you another list to work through, I reviewed the branch and pushed my follow-ups directly. Here is what changed and why, so nothing arrives as a surprise:

**The remaining correctness bug was a remount, not a re-broadcast.** The `preserved_props` route protects against unrelated prop updates, but the same root cause survived in a second form: `mounted()` re-applied the *declared* lines, and nothing ever wrote the live positions back into the prop. So any client-side remount — switching away from a `ui.tab_panels(keep_alive=False)` panel and back, or registering a handler at runtime, which makes NiceGUI re-render the element — snapped every anchor back to where it was declared, and the server mirror silently followed. The browser-reported snapshot is now mirrored into the prop (via `suspend_updates`, so it does not echo back), and `beforeUnmount` stores the live positions client-side, just like it already did for the editor value.

**The one-shot apply flag became idempotent.** `_to_dict` consumed a flag as a side effect, which only worked as long as there was exactly one outgoing serialization per assignment — an invariant spread across `client.py` and `outbox.py` rather than living where it was relied on. The flag is now set by the setter and cleared when the browser confirms, so `_to_dict` is a pure serializer again and a dropped update heals on the next one instead of being lost for the lifetime of the client.

**`on_anchor_change` now means what it says.** The tracker emitted on every document change while any anchor existed, so typing anywhere — even below every anchor — pushed an identical mapping once per debounce window and invoked every handler with unchanged values. The emit now compares against the last mapping sent; a deliberate apply still forces it, since the server treats that emit as confirmation that its declared anchors have landed.

**Out-of-range anchors are dropped, not clamped.** Declaring line 50 in a three-line document used to read back as line 3 — a plausible-looking position nobody asked for, with the warning surfacing far from the assignment — while the neighboring `line_tooltips` drops out-of-range entries, so two adjacent APIs of the same element disagreed. Anchors beyond the document are now dropped with a warning like tooltips, and a line below 1, which is wrong regardless of the document, is rejected at assignment time with a `ValueError` that points at the caller. The feature is unreleased, so no deprecation path was needed. (The version markers now say 3.16.0 — they still said 3.14.0, but both 3.14 and 3.15 shipped without the feature.)

**Tests and demo.** Two tests are gone: one asserted what its neighbor's first two lines already covered, and `test_clear_after_typing_does_not_resurrect_anchors` never reached the race it described — injecting the very bug it guards against left it green; it only turned red once the debounce was stretched to 150 ms. For the remount fix, the emission dedup and the constructor's validation order, I verified the new tests by reverting each fix and watching them fail. The demo lost five lines of string surgery that had nothing to do with anchors — and an instruction ("edit the code freely") that broke it, since editing the anchored line to anything not ending in a number made `int()` raise inside the click handler. It is now an anchor and a label bound to it.

The full suite is green on `adcbbf2fe` (1051 passed). Approving — nice work, and thanks for your patience through the review. [※](https://zauberzeug.github.io/colophon/#m=claude&t=Review%20and%20wording%20from%20the%20agent%3B%20the%20decisions%20it%20reports%20were%20the%20maintainer%27s.&p=agent "claude: Review and wording from the agent; the decisions it reports were the maintainer's.")

PR #5985, opened 2026-04-23
Add lint diagnostics API to CodeMirror

Jepson2k:
### Motivation

Embedding `ui.codemirror` as an in-app editor often needs to surface programmatically computed diagnostics — type-check results, lint output, runtime errors mapped back to source — onto specific lines or column ranges, with hover tooltips so the user can read the message without leaving the editor. CodeMirror 6's `@codemirror/lint` package provides this, but it isn't bundled in the current dist and there's no Python API for it.

pr-5985

### Implementation

- `Diagnostic` `TypedDict` with required `line` (1-indexed) and `message`, plus optional `severity` (`'error'` | `'warning'` | `'info'` | `'hint'`, default `'error'`), `source` (label shown next to the message), and `column` / `end_column` (1-indexed, exclusive end) for sub-line ranges.
- `diagnostics` is a mutable list of `Diagnostic` dicts; mutating it (or reassigning) syncs to the client.
- `open_lint_panel()` / `close_lint_panel()` / `toggle_lint_panel()` show or hide CodeMirror's built-in panel listing the current diagnostics.
- `get_diagnostic_count()` (async) returns a per-severity dict for status-bar UIs.

`@codemirror/lint` was already a transitive dep, so `package.json` doesn't change — just re-exports from `src/index.mjs` and a bundle rebuild.

Diagnostic messages render as plain text by default; pass `diagnostic_message_html=True` to the constructor to render messages as sanitized HTML via NiceGUI's `setHTML` polyfill. Out-of-range diagnostic lines are routed to the Python logger via `logAndEmit('warning', ...)`.

### Progress

- [x] The PR title is a short phrase starting with a verb like "Add ...", "Fix ...", "Update ...", "Remove ...", etc.
- [x] The implementation is complete.
- [x] This PR does not address a security issue.
- [x] Pytests have been added.
- [x] Documentation has been added.
- [x] No breaking changes to the public API.

PR #5991, opened 2026-04-24
Add decoration API to CodeMirror

Jepson2k:
### Motivation

Hosts that embed `ui.codemirror` as an in-app editor often need to overlay state on top of the user's text without modifying the document — diff markers, search highlights, breakpoints, "this range comes from line N of file Y" annotations, code folding, AI-suggestion previews, inline hints, etc. The Python API has no way to do any of this today. This PR is scoped to text-based widgets and decorations; full widget interactivity (clickable buttons, embedded NiceGUI elements, custom gutter markers) is meaningfully more complex and should land in its own PR.

pr-5991

### Implementation

A `decorations` property is a mutable list of typed dict specs; mutating it (or reassigning) syncs to the client. Four decoration kinds, dispatched on the `kind` field:

- **`mark`** — style a character range
- **`line`** — style an entire line
- **`replace`** — hide a range, or visually substitute it with text (`block: True` for whole-line / multi-line folding)
- **`widget`** — insert a text annotation at a position without modifying the document

Multiple consumers contribute by merging Python-side and assigning the combined list.

Caller-supplied widget text (`replace`-with-text, `widget`) renders as plain text by default; pass `decoration_text_html=True` to the constructor to render it as sanitized HTML via NiceGUI's `setHTML` polyfill.

### Progress


evnchn:
@Jepson2k If you can choose ONE PR out of all yours to merge first, which would you choose?

Jepson2k:
@evnchn I don't really have one that needs to be merged first but I can rank them in terms of complexity (and thus review bandwidth) going from least complex to most complex:

Band 1: light, shared-scaffolding / additive (roughly equal):
- #5988: line anchors (mutable-list sync)
- #6000: keybindings (factory + signal-back)
- #5984: editor signals (factory + signal-back)
- #5991: decorations (mutable-list sync)
- #5985: lint diagnostics (one-way push; + new bundled dep)
- #5989: Scene primitives + polar_grid (additive, follows existing dispatch)

Band 2:
- #5986: autocomplete (the only CodeMirror PR with an async Python editor round-trip)

Band 3: 
- #5993: clipping planes + named intersections (reconnect-aware state, material propagation, async-glTF path)
- #5990: pointer events + TransformControls (interactive gizmo, bidirectional transform streaming clear top)

Band 4 (Blocked until three.js bump):
- #6026: ViewHelper inset

evnchn:
@Jepson2k Sure but best choose one (maybe two) for the one to cut into since wanna see how many we can do for 3.13

Jepson2k:
@evnchn ok then lets do #5988 and #6000 first.

evnchn:
**TL;DR — request changes (nearly there; all findings are addressable, no redesign forced).**

The API is clean and the test coverage is genuinely good (28 tests, real DOM assertions incl. an XSS test). All local gates pass — pre-commit, mypy, pylint 10.00/10, pytest 28/28 green on my machine. But four things should land before merge, one of which is a real robustness hole:

1. **Malformed specs throw uncaught and silently void the *whole* decoration batch** (should-fix). One bad dict → `TypeError` at `doc.line(NaN)` → the entire assignment is dropped with only a browser-console error, no Python-side signal. You already guard the block-replace throw path ("breaks the editor for the rest of the page") — the numeric/type paths need the same discipline.
2. **Decorations don't re-map through edits** (design decision to confirm). They're a static facet at absolute offsets; a mark on `beta` drifts onto other text after an edit above it — while the line-tooltips *in the same file* track edits. This also makes the exposed `inclusiveStart/End`/`inclusive` knobs inert. Fine if intended (line-anchors #5988 is the tracking mechanism) — but then say so and drop/neutralize the inert knobs.
3. **Out-of-range `line` silently retargets** line 0/-1/9999 → line 1/last (should-fix). `line_tooltips` warns-and-ignores; match it.
4. **Offsets are UTF-16 code units** and bypass the element's own codepoint machinery (doc gap). A Python user computing `from`/`to` from a `str` index is wrong past any emoji/astral char.

_(Note: the `decorations` docstring's `*Added in version X.Y.Z*` placeholder needs no action from you — that's filled in by the maintainers once a specific milestone is assigned, so it's expected to stay as-is while this PR sits on the `Next` milestone. My earlier "must-fix nit" on it was wrong; corrected per maintainer guidance.)_

**Merge-order flag:** **#5991 is independent of #5988** — it needs only NiceGUI core's `ObservableList` (already in main) + CM6's own API; it doesn't touch `line_anchors.py` or the new event. The two only overlap *textually* (same three files), so whichever lands second takes a mechanical rebase. #5991 is currently `MERGEABLE`. So the agreed "#5988 first" does **not** gate this — it can land in either order.

Verified with a fresh-lineage (Codex) skeptic pass; it confirmed all four findings and sharpened finding 1. Evidence folded below.

---

<details>
<summary><b>What I ran (gates + empirical harness)</b></summary>

Worktree off `~/nicegui` at PR head `6dca4133` (`wt pr-5991`), rebased context vs `upstream/main` merge-base.

```
pre-commit (ruff/autopep8/codespell/quote-fixer/mdformat)  … all Passed
mypy ./nicegui                                             … Success: no issues found in 243 source files
pylint ./nicegui                                           … 10.00/10 (== master baseline)
pytest tests/test_codemirror.py                            … 28 passed in 21.98s
```

Diff is self-contained to the CodeMirror element:
```
 nicegui/elements/codemirror/codemirror.js          | 108 ++
 nicegui/elements/codemirror/codemirror.py          |  81 +-
 tests/test_codemirror.py                           | 231 ++
 website/.../codemirror_documentation.py            |  41 ++
```

Beyond the shipped tests I wrote three runnable MREs (findings 1/2/4) in the repo's own `Screen` harness — each asserts the *expected* (correct) behavior, so a red run == the bug is present, and they double as regression tests. Full self-contained file + exact command + real output in the **"Runnable MREs"** block at the bottom.

Exact version: **`nicegui 3.13.0.post38.dev0+6dca4133`** (PR head `6dca4133`) · Python 3.13.11 · macOS `Darwin arm64` · chrome 149.0.7827.201.
</details>

<details>
<summary><b>Finding 1 — malformed spec throws, aborts the whole batch (confirmed, ran it)</b></summary>

`_createDecoration` assumes valid runtime types, but props are user JSON and `TypedDict` enforces nothing at runtime. `{'kind': 'line'}` → `spec.line` undefined → `Math.max(1, Math.min(undefined, doc.lines))` = `NaN` → `doc.line(NaN)` throws. No try/catch around `_createDecoration` / `Decoration.set` / `editor.dispatch`.

**Repro:** `test_mre_finding1_*` below pushes `[{'kind':'line'}, {'kind':'line','line':2,'class':'sibling'}]` — one malformed spec + one valid sibling. Real output — the `Screen` fixture caught the uncaught browser error at teardown, *and* the valid sibling never applied:
```
TypeError: Cannot read properties of undefined (reading 'length')
    at f.lineInner (…/index-ciMdZ5pn.js:1:4051)
    at f.line
    at Proxy._createDecoration (codemirror.js:230:26)
    at Proxy.setDecorations (codemirror.js:204:26)
    at Proxy.handler (codemirror.js:88:14)
...
>   raise AssertionError('Condition not met')   # wait_for(sibling count == 1) timed out
```

**Severity, scoped honestly:** the throw propagates out of the Vue watcher and is swallowed there — the editor is **not** permanently bricked (I separately re-verified recovery: right after the bad spec, a valid decoration applied fine, doc intact, `contenteditable=true`). So the real damage is: **one bad dict in a list silently voids every decoration in that assignment** (not just the bad one), visible only in the browser console — Python gets no exception. Same class for `from:'x'`, missing `position`, etc.

Note your own guard comment on the block-replace path — *"CodeMirror requires block-replace ranges to span full lines; otherwise it throws out of editor.dispatch and breaks the editor for the rest of the page"* — is exactly the failure mode, already handled for one path. Suggest extending it: validate numeric fields (`Number.isInteger`/`Number.isFinite`) and range order per-spec, then `console.error`-and-skip the individual bad spec (as you already do for `from > to`), so a good spec in the same batch still lands.
</details>

<details>
<summary><b>Finding 2 — static overlays don't track edits; inclusivity knobs are inert (confirmed, ran it)</b></summary>

`setDecorations` installs a plain set via `EditorView.decorations.of(decorationSet)` inside a Compartment (codemirror.js:200-209). There's no `StateField` doing `set.map(tr.changes)` on update — unlike `tooltipField` (lines ~38-50) which maps through changes via `TooltipValue extends RangeValue`.

**Repro:** `test_mre_finding2_*` below marks `beta` (offsets 6..10), then inserts `XX` at doc start via real keystrokes. Real output:
```
>   assert marked == 'beta', f'mark drifted off its text: now covers {marked!r}'
E   AssertionError: mark drifted off its text: now covers 'a'
E   assert 'a' == 'beta'
```
The mark stayed at absolute 6..10 (now covering `a` of `XXalpha`) instead of following `beta`.

Consequence: `MarkDecorationSpec.inclusiveStart`/`inclusiveEnd` and `ReplaceDecorationSpec.inclusive` only influence *how a range maps through edits* — with no mapping, they're dead fields. Two coherent resolutions:
- **Map them** (StateField + `.map`), matching the tooltip approach and CM6's `DecorationSet`-is-a-`RangeSet` design; or
- **Declare them static** — document "absolute overlays; recompute + reassign after edits" and drop the inclusivity knobs (or mark them no-ops) so the API doesn't imply behavior it doesn't have.

Given #5988 line-anchors is the position-tracking primitive, "static + documented" may be the intended split — just needs to be explicit, and the inert knobs reconciled.
</details>

<details>
<summary><b>Finding 3 — out-of-range line silently retargets (confirmed, code-evident)</b></summary>

`const lineNum = Math.max(1, Math.min(spec.line, doc.lines));` (codemirror.js:229) → `line: 0`, `-1`, or `9999` all style line 1 or the last line rather than being rejected. Inconsistent with `line_tooltips`, which warns-and-ignores out-of-range lines (codemirror.py ~306-311). Suggest the same warn+ignore here.
</details>

<details>
<summary><b>Finding 4 — offsets are UTF-16 units, bypass the codepoint machinery (confirmed, code-evident)</b></summary>

`from`/`to`/`position` are passed straight through to JS (codemirror.py:136, 220-222) and consumed as CM6 offsets = UTF-16 code units. The element already carries `_encode_codepoints` / `_codepoints` (codemirror.py:241…) precisely to translate Python `str` indices ↔ UTF-16 for `value` — decorations don't use it. So `text.index('x')` in Python → wrong `from` after any non-BMP char (emoji, astral).

**Repro:** `test_mre_finding4_*` below sets doc `'😀beta'`, computes `from,to = doc.index('beta'), +len('beta')` = `1,5` (correct **Python** indices), and marks it. Real output:
```
E   AssertionError: python indices [1:5] did not mark "beta" —
E   marked 4 chars, codepoints [56832, 98, 101, 116]
```
`56832` = `0xDE00`, the **lone low-surrogate** half of 😀 (which is 2 UTF-16 units), followed by `b e t`. So the Python-correct `[1:5]` actually marked `�bet` on the client. At minimum a docstring warning; ideally a helper that accepts Python indices (parity with how `value` is handled).
</details>

<details>
<summary><b>Nit + notes</b></summary>

- **Version placeholder:** `decorations` docstring says `*Added in version X.Y.Z*` (codemirror.py:216). Set to the real release (`3.13.0`, matching `line_tooltips`).
- **Split sanitization story (note, acceptable):** `decoration_text_html` gates only replace/widget `text` through DOMPurify (`setHTML`). `attributes` (mark/line) is *always* raw DOM attributes incl. `onclick`, never sanitized. The docstring warns ("Do not pass untrusted input"), consistent with NiceGUI's warn-not-raise / `.html()` norms — fine, but worth surfacing in the public docs page too, not just the docstring.
- **`widget.side` default 0→1 (note):** you default `side` to `1` vs CM6's `0`. Reasonable, just document it.
- **Good calls worth keeping:** the block-replace full-line guard (console.error + skip, no throw), `from > to` guard, defensive clamping, the DOMPurify-gated HTML path, and the `decorations.append()/del/clear()` mutable-list-sync (grounded: `_props` is an `ObservableDict` that wraps the list as an `ObservableList` propagating in-place mutations to `element.update()` — verified in `test_decorations_list_mutations_sync`).
</details>

<details>
<summary><b>Fresh-lineage (Codex) skeptic pass</b></summary>

Ran Codex read-only over the diff prompted to refute. Verdict: "request changes." It independently confirmed findings 1-4 and sharpened finding 1 (the throw / no-try-catch path) from a documentation gap to a robustness bug. It flagged nothing I'd dispute; it also noted `Decoration.set(all, true)` sorting is correct (2nd arg sorts unsorted ranges) — so no issue there. I then empirically re-verified the throw-and-recover behavior myself (the runnable MRE below) rather than taking its "bricks the editor" read at face value — the editor recovers, so I've scoped finding 1 to "voids the batch," not "bricks."
</details>

<details>
<summary><b>Runnable MREs — paste into <code>tests/test_decoration_mres.py</code>, run <code>pytest</code></b></summary>

Version: `nicegui 3.13.0.post38.dev0+6dca4133` (PR head) · Python 3.13.11 · macOS Darwin arm64 · chrome 149.0.7827.201.
Each test asserts the **expected** behavior, so all three are **red on this PR** (red = bug present) and would go green once fixed — i.e. they double as regression tests.

```python
"""MREs for PR #5991 review findings. Each asserts the EXPECTED (correct) behavior,
so a red test == the bug is present. Drop-in for tests/ ; runs in the repo's Screen harness.
"""
from selenium.webdriver import ActionChains
from selenium.webdriver.common.by import By
from selenium.webdriver.common.keys import Keys
from selenium.webdriver.support import expected_conditions as EC
from selenium.webdriver.support.ui import WebDriverWait

from nicegui import ui

# pylint: disable=protected-access


def _mount(screen: Screen) -> None:
    WebDriverWait(screen.selenium, 5).until(
        EC.presence_of_element_located((By.CSS_SELECTOR, '.cm-content')))


def _count(screen: Screen, cls: str) -> int:
    return screen.selenium.execute_script(
        f'return document.querySelectorAll(".cm-content .{cls}").length;')


def test_mre_finding1_malformed_spec_voids_whole_batch(screen: Screen):
    """A malformed spec should be skipped, not abort the whole assignment.
    EXPECTED: the valid sibling decoration in the same list still applies (count == 1),
    and no uncaught browser error. CURRENT: throws in _createDecoration -> whole batch lost.
    """
    editor = None

    @ui.page('/')
    def page():
        nonlocal editor
        editor = ui.codemirror('alpha\nbeta\ngamma')

    screen.open('/')
    _mount(screen)
    # one malformed spec (missing 'line' -> undefined -> doc.line(NaN)) + one valid sibling
    editor.decorations = [
        {'kind': 'line'},                                   # malformed
        {'kind': 'line', 'line': 2, 'class': 'sibling'},    # valid — should still apply
    ]
    screen.wait_for(lambda: _count(screen, 'sibling') == 1)  # EXPECTED — currently times out


def test_mre_finding2_mark_drifts_on_edit(screen: Screen):
    """A mark should follow its text through document edits (like line_tooltips do).
    EXPECTED: after inserting at doc start, the mark still covers 'beta'.
    CURRENT: static facet, no .map() -> mark stays at absolute offsets and drifts to 'a'.
    """
    editor = None

    @ui.page('/')
    def page():
        nonlocal editor
        editor = ui.codemirror('alpha\nbeta\ngamma')

    screen.open('/')
    _mount(screen)
    editor.decorations = [{'kind': 'mark', 'from': 6, 'to': 10, 'class': 'm'}]  # 'beta'
    screen.wait_for(lambda: _count(screen, 'm') == 1)
    assert screen.selenium.execute_script(
        'return document.querySelector(".cm-content .m").textContent;') == 'beta'  # control: pre-edit
    # real user edit at doc start (Up*5 to line 1, Home, type)
    screen.selenium.find_element(By.CSS_SELECTOR, '.cm-content').click()
    ac = ActionChains(screen.selenium)
    for _ in range(5):
        ac.send_keys(Keys.UP)
    ac.send_keys(Keys.HOME).send_keys('XX').perform()
    screen.wait_for(lambda: editor.value.startswith('XX'))  # edit landed at start
    marked = screen.selenium.execute_script(
        'const s=document.querySelector(".cm-content .m"); return s ? s.textContent : "<<gone>>";')
    assert marked == 'beta', f'mark drifted off its text: now covers {marked!r}'  # EXPECTED — currently fails


def test_mre_finding4_offsets_are_utf16_not_python_indices(screen: Screen):
    """Offsets should behave like Python str indices (parity with how `value` is handled).
    doc = '😀beta'. Python 'beta' == doc[1:5]. Passing from=1,to=5 SHOULD mark 'beta'.
    CURRENT: offsets are raw UTF-16 units (emoji = 2 units) -> from=1,to=5 mis-marks.
    """
    editor = None

    @ui.page('/')
    def page():
        nonlocal editor
        editor = ui.codemirror('😀beta')

    screen.open('/')
    _mount(screen)
    doc = '😀beta'
    py_from, py_to = doc.index('beta'), doc.index('beta') + len('beta')  # 1, 5 (Python indices)
    editor.decorations = [{'kind': 'mark', 'from': py_from, 'to': py_to, 'class': 'u'}]
    screen.wait_for(lambda: _count(screen, 'u') == 1)
    # compare client-side and return a JSON-safe result (raw emoji/surrogate textContent
    # crashes chromedriver's result serializer, so never ship it back over the wire)
    result = screen.selenium.execute_script(
        'const t = document.querySelector(".cm-content .u").textContent;'
        'return {ok: t === "beta", len: t.length, codes: Array.from(t).map(c => c.codePointAt(0))};')
    assert result['ok'], (f'python indices [{py_from}:{py_to}] did not mark "beta" — '
                          f'marked {result["len"]} chars, codepoints {result["codes"]} '
                          f'(0x1F600=😀, 0x62=b 0x65=e 0x74=t 0x61=a)')  # EXPECTED — fails
```

Real output (`pytest tests/test_decoration_mres.py -v`):
```
test_mre_finding1_malformed_spec_voids_whole_batch FAILED
test_mre_finding1_malformed_spec_voids_whole_batch ERROR      # browser console error at teardown
test_mre_finding2_mark_drifts_on_edit               FAILED
test_mre_finding4_offsets_are_utf16_not_python_indices FAILED
========================== 3 failed, 1 error in 6.66s ==========================

# finding1 — teardown ERROR (uncaught browser error) + the wait_for that never met:
JavaScript console error: TypeError: Cannot read properties of undefined (reading 'length')
    at f.lineInner (…/index-ciMdZ5pn.js) → f.line → Proxy._createDecoration (codemirror.js:230:26)
    → Proxy.setDecorations (codemirror.js:204:26) → Proxy.handler (codemirror.js:88:14)
>   raise AssertionError('Condition not met')        # sibling decoration never applied

# finding2:
E   AssertionError: mark drifted off its text: now covers 'a'
E   assert 'a' == 'beta'

# finding4:
E   AssertionError: python indices [1:5] did not mark "beta" —
E   marked 4 chars, codepoints [56832, 98, 101, 116]   # 56832=0xDE00 lone low-surrogate of 😀, then b,e,t
```

Note: this file is deliberately **red** and is not part of the PR — it's evidence you can replay / adopt as regression tests once the fixes land. (Findings 3 and 5 are code-evident at the cited lines and don't need a repro.)
</details>

Jepson2k:
@evnchn I appreciate the review and have addressed the findings in the latest commit.

evnchn:
> @evnchn I appreciate the review and have addressed the findings in the latest commit.

Thanks @Jepson2k — pulled 2f6895fc, re-ran the gates, and replayed my three original red MREs. **All four confirmed findings are resolved and every MRE that was red at 6dca4133 is now green.** A couple of the fixes went past what I flagged (the `logAndEmit`-to-Python path, a zero-length-mark guard I didn't even call out). Receipts below for the record.

**✅ Re-verified at 2f6895fc**

| Gate | Result |
|---|---|
| `pytest tests/test_codemirror.py` | **31 passed** (28 → +3 new regression tests) |
| my 3 original MREs (findings 1 / 2 / 4) | **all green** (were red pre-fix) |
| `mypy ./nicegui` | clean — 243 files |
| `pylint …/codemirror.py` | 10.00/10 |

<details>
<summary><b>Finding-by-finding disposition</b></summary>

| # | Finding | Fix in 2f6895fc | Status |
|---|---------|-------------------|--------|
| 1 | Malformed spec throws → voids the whole batch | Per-spec `Number.isInteger` validation, **warn-and-skip via `logAndEmit`** (surfaces to Python too — better than the `console.error` I suggested), plus a `try/catch` backstop in `setDecorations`. Bonus: a zero-length-mark (`from === to`) guard I hadn't flagged. | ✅ Fixed |
| 2 | Static overlays don't track edits; `inclusive*` knobs inert | Took the **"map them"** route — new `decorationField` `StateField` doing `.map(tr.changes)`, replacing the static compartment. `inclusiveEnd` now demonstrably grows the mark (your new `test_decoration_inclusive_end_extends_mark` proves it). | ✅ Fixed — the stronger of my two options |
| 3 | Out-of-range line silently retargets | Warn-and-skip, matching `line_tooltips`. | ✅ Fixed |
| 4 | Offsets are UTF-16 units, not Python indices | Documented in the docstring **and** the website docs. No Python-index helper — but I'd scoped that as *"ideally"*; the docstring was the *"at minimum"* bar, so this is a fair call. Possible follow-up, not a blocker. | ✅ Addressed |
| — | Nits (version placeholder, `attributes`-never-sanitized visibility) | Version → `3.15.0`; the raw-`attributes` warning now lives on the website docs page too, not just the docstring. | ✅ |

</details>

<details>
<summary><b>What I re-ran</b></summary>

Worktree off `~/nicegui` at PR head `2f6895fc`, rebased on current `upstream/main`.

```
pytest tests/test_codemirror.py          … 31 passed in 26.37s
my 3 original MREs (findings 1/2/4)      … 3 passed   (all RED at 6dca4133)
mypy ./nicegui                           … Success: no issues found in 243 source files
pylint nicegui/…/codemirror.py           … 10.00/10 (== baseline)
```

Finding 1 MRE (malformed + valid sibling in one list) → the valid sibling now applies and no uncaught browser error. Finding 2 MRE (mark on `beta`, then edit at doc start via real keystrokes) → mark follows its text. Finding 4 MRE re-cast against the **documented** UTF-16 contract (`😀beta`, `beta` at UTF-16 `[2:6]`) → marks `beta` exactly as documented.
</details>

From my side this **LGTM** — findings resolved, gates clean, and the new tests lock in the behavior. Over to you for the merge call, @falkoschindler.

falkoschindler:
Thanks for your patience here, @Jepson2k — this sat much longer than it deserved.

I reviewed the branch in depth and pushed six commits to it directly, so this comment is only about the parts that need a decision rather than a fix. I also put the PR on the **3.17** milestone, the cycle that is just starting: what is left here is API shape, and I would rather settle that deliberately than rush it.

## What I pushed to the branch

- **Reuse the shared range-set factory for decorations** — `decorationField` re-implemented `defineRemappableRangeSet()` from the same file. A `DecorationSet` *is* a `RangeSet` (`Decoration.none` is `RangeSet.empty`, `Decoration.set(v, true)` is `RangeSet.of(v, true)`), so the factory now just takes the `provide` option and the remap logic exists once. Also drops `deep: true` from the watcher, since the prop is always replaced wholesale.
- **Extract the decoration range validation into one helper** — the `mark` and `replace` branches carried the same ten lines verbatim.
- **Share one page scaffold across the decoration tests** — twelve tests repeated the same eight-line setup; they now call `_open_editor()`. The mount wait uses `Screen.wait_for()` like `tests/test_codemirror_line_anchors.py` instead of a raw `WebDriverWait` with its own timeout.
- **Assert sanitization on the DOM instead of an execution flag** — the XSS test checked that a window flag stayed unset, but a `<script>` inserted through `innerHTML` never executes, so that assertion held even with sanitization removed entirely. Both assertions now look at what DOMPurify actually strips; verified by swapping `setHTML` for a raw `innerHTML` assignment and watching the test go red.
- **Warn and skip decoration specs with an unusable text or range** — a widget without a string `text` silently rendered an empty span, and an empty `replace` range slipped through into CodeMirror's own `RangeError` and the error-level backstop. Both are warned-and-skipped now. The empty-range check only fires where CodeMirror actually refuses: an empty range stays legal while inclusive, and `block` implies inclusive.
- **Document decorations as a 3.17 addition** — version annotations on the two new parameters, on the property and on the demo, plus one sentence-per-line fix.

`tests/test_codemirror.py` is at 33 passing.

## The one real bug, and it needs a decision

Decorations are mapped through edits by the state field, which is what makes the feature's promise work — but the mapped positions are discarded again on the next update of the editor element. Every update re-sends the whole props dict, the client replaces the element wholesale, and the watcher re-applies the *declared* offsets. `line-anchors` avoids exactly this via `preserved_props` plus a browser-to-server mirror; `decorations` has neither.

Concretely: a mark on `beta`, the user types two characters at the start of the document (the mark correctly follows), then any `editor.theme = ...`, `.classes(...)` or tooltip change snaps it back onto the wrong text. The same applies to `editor.decorations.append(...)`, which is the headline use of the mutable-list API — appending one spec re-applies all the others at their declared offsets.

Three ways out, and I would like your read on them:

1. **Client-side guard** — remember the last applied list and skip identical ones. Five lines, kills the unrelated-update clobbering, but not the `append` case, which is the one that matters most.
2. **Full mirror-back, like line anchors** — the browser reports mapped offsets, the server mirrors them into the prop and marks it preserved. Solves everything including reload, and makes reading `decorations` honest. But it means a new event plus serializing every decoration back into a spec, and the payload here is a list of arbitrary many specs rather than a small `{id: line}` map — a linter streaming diagnostics would push that back on every edit burst.
3. **Client-side diffing** — the client keeps spec-to-mapped-range pairs; on a new list, specs that are unchanged keep their mapped position and only new ones land at declared offsets. Roughly 25 lines, entirely in the browser, no new event and no change to what the property means on the Python side. It also makes the `beforeUnmount` write-back trivial, which fixes the related remount case where a re-mounted editor re-applies declared offsets against an edited document.

**My preference is 3.** It fixes the case that actually bites, costs no traffic, and does not commit the API to read-back semantics. Reading `decorations` then keeps returning what you declared, which is worth stating plainly in the docstring — if you need to know where something *is* now, that is what line anchors are for.

## The one-way door: UTF-16 offsets

`from`, `to` and `position` are UTF-16 code units, and the docstring asks callers to account for surrogate pairs themselves. Anyone computing `text.index(...)` on a document containing an emoji gets a decoration that is off by one per astral character, with no error anywhere.

The element already draws that boundary once, in the other direction: `_codepoints` plus `_apply_change_set` translate the browser's UTF-16 change sets into Python indices, so `value` is an ordinary `str` no matter what is in the document. Decorations are the first place where an offset crosses into the public API at all, so the question is whether that boundary holds here too. Worth being precise about the cost, though: the reverse direction is not a reuse of `_codepoints` — that map answers "how many code points precede this UTF-16 unit", while here we need the inverse, which is one pass over `value` (or a `value[:i].encode('utf-16-le')` measurement) per offset.

This is the only thing in the PR that cannot be changed after release. **My preference is to accept Python `str` indices and convert on the way out**, with a fast path for documents without astral characters, which is nearly all of them. One wrinkle to solve if we do it: because `decorations` is a mutable list, `append()` never passes through the setter, so the conversion has to happen when the props are serialized rather than at assignment — against a `value` that may in principle have moved on since the caller computed the offset. In practice that window is microseconds, and an offset computed against a stale document is already wrong independently of its encoding. If we would rather keep CodeMirror's native addressing, that is defensible too — but then it is a deliberate choice worth recording here rather than a default that nobody picked.

## Two smaller ones

- **Python-side validation.** `line_anchors` raises `ValueError` at the assignment site; `decorations` accepts anything and the developer learns about a dropped spec from a browser-forwarded log warning. I would add the light version — unknown kind, missing required key, inverted or negative offsets — and leave everything document-dependent in JS, exactly the way line anchors splits it.
- **`decoration_text_html` is element-global** while the JS `TextWidget` already takes `html` per instance, so one HTML annotation puts every widget and replace text into HTML mode. A per-spec `html` field is purely additive, so this can also land later without breaking anything — no need to decide it now.

## One follow-up outside this PR

While comparing the three line-number checks in the file, the new one turned out to be the only one that checks `Number.isInteger`. `applyLineAnchors` does not, so a line anchor of `2.5` passes the range check and reaches `doc.line(2.5)`. That is already-merged code from #5988, so it belongs in its own small PR rather than here. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%2C%20code%20fixes%20and%20wording%20from%20the%20model%3B%20milestone%20call%20and%20direction%20from%20the%20maintainer.&a=claude-code "claude-fable-5: Review, code fixes and wording from the model; milestone call and direction from the maintainer.")

Jepson2k:
*Posted by Claude Code on @Jepson2k's behalf.*

Thanks — all three are in, as four commits: per-spec diffing with the `beforeUnmount` write-back, `str` indices converted at serialization, and validation at the assignment site. Per-spec `html` left for later, as you suggested.

Three things worth your eye. The validation makes missing-required-keys raise before reaching the browser, which took a case out of one of your tests and one of mine — the widget test now ships an unusable `text` instead of none, and the line test simply lost its missing-key entry, so the JS structural guards are no longer covered by a test even though `append()` can still reach them. Related: validation only runs on assignment, so `decorations.append(...)` bypasses it entirely — happy to validate in `_to_dict` alongside the conversion if you'd rather it covered the documented mutation path.

And one semantic that option 3 implies: since the diff key is the spec itself, re-assigning an identical list keeps the drifted positions, so there's no way to snap a decoration back to its declared offsets the way `line_anchors` deliberately allows. Happy to add an escape hatch if you want one.

Follow-up in #6285 — worth noting `doc.line(2.5)` doesn't throw, it snaps to a neighbouring line and mirrors back as if declared.

evnchn:
On my end, I see the per-spec diffing, the Python `str` offsets, the assignment-site validation and the #6285 follow-up are done, but not sure about the per-spec `html` field and the three semantics @Jepson2k raised back at you (validation not covering `append`, the JS structural guards left untested, no way to snap back to declared offsets).

<details>
<summary><b>Item-by-item, and what each one is verified against</b></summary>

Checked out at PR head `1cdcee74` and replayed each of your asks as a test that asserts the **expected** post-fix behavior, then ran the same file against `413b4309` (your last commit, before Jake's four) as a control. **8 red before → 10 green after**, so the greens are load-bearing rather than vacuous.

| Your ask | Status | What I ran |
|---|---|---|
| Option 3 — client-side diffing, unrelated update must not clobber | ✅ | mark on `beta`, type `XX` at doc start, then `editor.theme = 'basicLight'` → mark still on `beta`. Red at `413b4309`. |
| …and the `append` case, "the one that matters most" | ✅ | same, then `decorations.append({'kind': 'line', 'line': 3, …})` → the pre-existing mark stays put. Red at `413b4309`. |
| …and the remount case the `beforeUnmount` write-back buys | ✅ | same, then `editor.move(other_container)` → mark still on `beta`. Confirmed the component genuinely unmounts (temporarily counted `beforeUnmount` calls on `window`; it fires) — otherwise this test would have proved nothing. Red at `413b4309`. |
| UTF-16 → Python `str` indices, converted on the way out | ✅ | doc `😀beta`, `from=value.index('beta')` → marks `beta` exactly. Also `[0:1]` marks the **whole** emoji, which is the off-by-one `1cdcee74` fixes. BMP-only doc unaffected (fast path). Red at `413b4309`. |
| …and the wrinkle you named: conversion at serialization, since `append()` skips the setter | ✅ | same astral doc, but the spec arrives via `decorations.append(...)` rather than assignment → still converted. Red at `413b4309`. |
| Python-side validation, light version | ✅ | unknown kind / missing required key / non-integer offset / negative offset / `from > to` / `line=0` all raise `ValueError` at the assignment site. The constructor is smoke-tested on unknown-kind only, but routes through the same `_validate_decorations` (`codemirror.py:161` and `:274`). Red at `413b4309`. |
| Docstring stating that reading `decorations` returns what you declared | ✅ | `codemirror.py:258-260`, plus the pointer to `line_anchors`. |
| `applyLineAnchors` fractional-line follow-up | ✅ filed | #6285, open, on `Next`. |
| Per-spec `html` instead of element-global `decoration_text_html` | ❌ not in | Deferred, matching your "no need to decide it now". |

The three Jake raised back at you I can only confirm as *true*, not resolve — they're your call:

- `decorations.append({'kind': 'nope'})` really does bypass `_validate_decorations` — ran it, no `ValueError`. I asserted only the Python side; that it then syncs follows from `ObservableList.append` → `_handle_change()` → `element.update()`, which I read rather than asserted.
- No test in `tests/test_codemirror.py` now sends a structurally-bad spec to JS, so `_createDecoration`'s unknown-kind / missing-key branches are unexercised — reachable only via `append`.
- The diff key being the spec itself does mean re-assigning an identical list keeps drifted positions, with no declared-offset reset.

</details>

<details>
<summary><b>Gates, and the runnable file</b></summary>

At `1cdcee74`, worktree off `~/nicegui`:

```
pytest tests/test_codemirror.py   … 39 passed in 29.40s   (33 at your 413b4309)
mypy ./nicegui                    … Success: no issues found in 264 source files
pylint ./nicegui                  … 10.00/10
my 10 MREs                        … 10 passed  (8 of them RED at 413b4309)
```

Upstream CI on the head commit is green as well. The branch is 10 commits behind `main` but still `MERGEABLE`.

Python 3.12.11 · macOS `Darwin arm64` · Chrome 151.0.7922.138.

Drop this in `tests/test_falko_asks_mre.py` and run `pytest tests/test_falko_asks_mre.py`. It is not part of the PR — it is the evidence above, replayable, and it is what turns red when checked out at `413b4309`.

```python
"""MREs for the three asks in Falko's 2026-08-12 review of PR 5991.

Each asserts the EXPECTED (post-fix) behavior, so green == the ask is satisfied.
Not part of the PR; evidence only.
"""
import pytest
from selenium.webdriver import ActionChains
from selenium.webdriver.common.by import By
from selenium.webdriver.common.keys import Keys

from nicegui import ui
from nicegui.testing import Screen, User


def _open(screen: Screen, doc: str) -> ui.codemirror:
    editor: ui.codemirror = None  # type: ignore[assignment]

    @ui.page('/')
    def page():
        nonlocal editor
        editor = ui.codemirror(doc)

    screen.open('/')
    screen.wait_for(lambda: bool(screen.selenium.find_elements(By.CSS_SELECTOR, '.cm-content')))
    return editor


def _count(screen: Screen, cls: str) -> int:
    return screen.selenium.execute_script(
        f'return document.querySelectorAll(".cm-content .{cls}").length;')


def _text(screen: Screen, cls: str) -> str:
    return screen.selenium.execute_script(
        f'const s=document.querySelector(".cm-content .{cls}"); return s ? s.textContent : "<<gone>>";')


def _type_at_start(screen: Screen, chars: str) -> None:
    screen.selenium.find_element(By.CSS_SELECTOR, '.cm-content').click()
    ac = ActionChains(screen.selenium)
    for _ in range(5):
        ac.send_keys(Keys.UP)
    ac.send_keys(Keys.HOME).send_keys(chars).perform()


# --- ask 1: option 3, per-spec diffing (mapped positions survive an unrelated update) ---

def test_ask1_unrelated_update_keeps_mapped_position(screen: Screen):
    """A theme change must not snap a mapped mark back onto its declared offsets."""
    editor = _open(screen, 'alpha\nbeta\ngamma')
    editor.decorations = [{'kind': 'mark', 'from': 6, 'to': 10, 'class': 'm'}]
    screen.wait_for(lambda: _count(screen, 'm') == 1)
    assert _text(screen, 'm') == 'beta'
    _type_at_start(screen, 'XX')
    screen.wait_for(lambda: editor.value.startswith('XX'))
    assert _text(screen, 'm') == 'beta', 'mark drifted before the unrelated update'
    editor.theme = 'basicLight'  # unrelated update -> full props re-send
    screen.wait(0.5)
    assert _text(screen, 'm') == 'beta', 'unrelated update snapped the mark back to declared offsets'


def test_ask1_append_keeps_existing_mapped_positions(screen: Screen):
    """decorations.append() must not re-apply the existing specs at their declared offsets."""
    editor = _open(screen, 'alpha\nbeta\ngamma')
    editor.decorations = [{'kind': 'mark', 'from': 6, 'to': 10, 'class': 'm'}]
    screen.wait_for(lambda: _count(screen, 'm') == 1)
    _type_at_start(screen, 'XX')
    screen.wait_for(lambda: editor.value.startswith('XX'))
    assert _text(screen, 'm') == 'beta'
    editor.decorations.append({'kind': 'line', 'line': 3, 'class': 'l'})
    screen.wait_for(lambda: _count(screen, 'l') == 1)
    assert _text(screen, 'm') == 'beta', 'append() snapped the pre-existing mark back'


def test_ask1b_remount_keeps_mapped_positions(screen: Screen):
    """A client-side remount must restore mapped offsets, not the declared ones."""
    editor: ui.codemirror = None  # type: ignore[assignment]
    target: ui.column = None  # type: ignore[assignment]

    @ui.page('/')
    def page():
        nonlocal editor, target
        with ui.column():
            editor = ui.codemirror('alpha\nbeta\ngamma')
        target = ui.column()

    screen.open('/')
    screen.wait_for(lambda: bool(screen.selenium.find_elements(By.CSS_SELECTOR, '.cm-content')))
    editor.decorations = [{'kind': 'mark', 'from': 6, 'to': 10, 'class': 'r'}]
    screen.wait_for(lambda: _count(screen, 'r') == 1)
    _type_at_start(screen, 'XX')
    screen.wait_for(lambda: editor.value.startswith('XX'))
    assert _text(screen, 'r') == 'beta'
    editor.move(target)  # unmount + remount on the client
    screen.wait(1.0)
    assert _text(screen, 'r') == 'beta', 'remount re-applied the declared offsets'


# --- ask 2: Python str indices, converted on the way out ---

def test_ask2_python_str_indices_address_text_after_an_emoji(screen: Screen):
    """value.index('beta') must address 'beta' even with an astral char in the document."""
    doc = '\U0001F600beta'
    editor = _open(screen, doc)
    editor.decorations = [{'kind': 'mark', 'from': doc.index('beta'), 'to': doc.index('beta') + 4, 'class': 'u'}]
    screen.wait_for(lambda: _count(screen, 'u') == 1)
    result = screen.selenium.execute_script(
        'const t=document.querySelector(".cm-content .u").textContent;'
        'return {ok: t === "beta", len: t.length, codes: Array.from(t).map(c => c.codePointAt(0))};')
    assert result['ok'], f'python indices did not mark "beta" - got {result}'


def test_ask2_python_str_indices_address_the_astral_char_itself(screen: Screen):
    """doc[0:1] must mark the emoji itself, not half of it (the 1cdcee74 off-by-one)."""
    doc = '\U0001F600beta'
    editor = _open(screen, doc)
    editor.decorations = [{'kind': 'mark', 'from': 0, 'to': 1, 'class': 'e'}]
    screen.wait_for(lambda: _count(screen, 'e') == 1)
    result = screen.selenium.execute_script(
        'const t=document.querySelector(".cm-content .e").textContent;'
        'return {ok: t === "\\uD83D\\uDE00", len: t.length};')
    assert result['ok'], f'python indices [0:1] did not mark the whole emoji - got {result}'


def test_ask2_bmp_document_is_unaffected(screen: Screen):
    """The fast path must leave a plain document alone."""
    editor = _open(screen, 'alpha\nbeta\ngamma')
    editor.decorations = [{'kind': 'mark', 'from': 6, 'to': 10, 'class': 'b'}]
    screen.wait_for(lambda: _count(screen, 'b') == 1)
    assert _text(screen, 'b') == 'beta'


# --- ask 3: Python-side validation at the assignment site ---

def test_ask3_structural_specs_raise_at_assignment(screen: Screen):
    """Structurally impossible specs raise ValueError in Python, not a browser warning."""
    editor = _open(screen, 'alpha\nbeta\ngamma')
    for spec, needle in [
        ({'kind': 'nope', 'from': 0, 'to': 1}, 'unknown kind'),
        ({'kind': 'line'}, 'missing required key'),
        ({'kind': 'mark', 'from': 'x', 'to': 1}, 'needs an integer'),
        ({'kind': 'mark', 'from': -1, 'to': 1}, 'offsets start at 0'),
        ({'kind': 'mark', 'from': 5, 'to': 1}, 'from=5 > to=1'),
        ({'kind': 'line', 'line': 0}, 'lines are 1-indexed'),
    ]:
        with pytest.raises(ValueError, match=needle):
            editor.decorations = [spec]


async def test_ask3_constructor_validates_too(user: User):
    @ui.page('/')
    def page():
        ui.label('Some content')

    await user.open('/')
    with user:
        with pytest.raises(ValueError, match='unknown kind'):
            ui.codemirror('alpha', decorations=[{'kind': 'nope'}])


def test_ask3_append_bypasses_python_validation(screen: Screen):
    """Jake's own caveat: append() never passes through the setter."""
    editor = _open(screen, 'alpha\nbeta\ngamma')
    editor.decorations.append({'kind': 'nope'})  # no ValueError - reaches the browser
    assert editor.decorations == [{'kind': 'nope'}]


def test_ask2_conversion_survives_append_not_just_assignment(screen: Screen):
    """Falko's wrinkle: append() skips the setter, so the conversion must run at serialization."""
    doc = '\U0001F600beta'
    editor = _open(screen, doc)
    editor.decorations.append({'kind': 'mark', 'from': doc.index('beta'), 'to': doc.index('beta') + 4, 'class': 'a'})
    screen.wait_for(lambda: _count(screen, 'a') == 1)
    result = screen.selenium.execute_script(
        'const t=document.querySelector(".cm-content .a").textContent;'
        'return {ok: t === "beta", len: t.length};')
    assert result['ok'], f'appended spec was not converted from python indices - got {result}'
```

Control run at `413b4309` (your last commit, before Jake's four) — the greens above are not vacuous:

```
FAILED test_ask1_unrelated_update_keeps_mapped_position
FAILED test_ask1_append_keeps_existing_mapped_positions
FAILED test_ask1b_remount_keeps_mapped_positions
FAILED test_ask2_python_str_indices_address_text_after_an_emoji
FAILED test_ask2_python_str_indices_address_the_astral_char_itself
FAILED test_ask2_conversion_survives_append_not_just_assignment
FAILED test_ask3_structural_specs_raise_at_assignment
FAILED test_ask3_constructor_validates_too
8 failed, 2 passed
```

The two that pass on both sides do so by design: `test_ask2_bmp_document_is_unaffected` is the unchanged-behavior control, and `test_ask3_append_bypasses_python_validation` documents a gap that is still open.

</details>

falkoschindler:
I went through the branch with a full review and then pushed five commits on top that address what it found. Here is the reasoning behind them, so the changes are easy to follow.

### What the review found

The review reproduced five behavioral problems with Screen tests. Four of them had one root cause: the client keyed every decoration by `JSON.stringify` of its spec (after the UTF-16 conversion) to tell unchanged specs apart on the next props update and keep them where the state field had mapped them. That key was fragile in exactly the situations the mechanism existed for:

- Typing an emoji before a mark changed its converted offsets and therefore its key, so the next unrelated update (a theme change, say) dropped the mark onto ` bet` instead of `beta`.
- A mark whose text the user had deleted was gone from the field, so the next unrelated update re-applied the declared offsets and the mark resurrected on `gamm`.
- After a client-side remount (`ui.tab_panels(keep_alive=False)`) the field was keyed by the mapped specs written into the client props, while the server kept sending the declared ones, so the next unrelated update snapped everything back.
- A block fold whose last line the user joined with the next one kept rendering fine in CodeMirror, but the "block replace must cover full lines" check dropped it on the next rebuild. The bundle has no such throw, contrary to the comment.

The fifth was independent: a value and a decorations change made in the same handler arrive in one message, and Vue runs the prop watchers before nicegui.js calls `setEditorValueFromProps`, so the specs were built against the old document and then mapped through the edit.

Smaller findings: in-place mutation (`editor.decorations.append(...)`) bypassed `_validate_decorations`, so the docstring's `ValueError` held for only one of the two advertised write paths and the JS side had to duplicate every structural check; `class` on a text-less replace never reached the DOM because CodeMirror's replace spec has no such field; the `:class:` Sphinx roles in the docstring made every visit to the docs page log `Unknown interpreted text role "class"`, since the website renders docstrings with plain docutils; and a handful of duplication and placement nits.

### The commits

1. **Preserve mapped decorations across unrelated updates, as line anchors do.** The server now marks `decorations` as a preserved prop unless a write is pending, exactly like `line-anchors`, so the watcher only fires on a deliberate assignment or in-place change. That removes the key matching, the mapped-range lookup, the try/catch around it, and the full-line check. Two consequences worth knowing: a write, including `append`, re-applies all specs at their declared offsets (the same contract `test_reassign_same_declared_value_snaps_back` documents for anchors), so offsets have to be computed from the current `value`; and the UTF-16 conversion now happens in the browser, which holds the document, instead of in `_to_dict` on every flush of any prop. Validation is one Python function used on both write paths: the setter raises as before, and a spec that slipped in through an in-place change is skipped with a warning when the list is sent (an `on_change` handler cannot raise to the caller, since `handle_event` catches exceptions). The browser only checks what depends on the document. This commit also drops the `:class:` roles and merges the two per-kind tables into one required-keys table plus one table of numeric minimums.
2. **Tidy the decoration code paths.** `class` on a text-less replace is no longer forwarded and the docstring says what `class` applies to; the decorations facet is wired next to its field in the extensions list instead of a `provide` parameter on the shared factory; the widget and the line tooltip share one text-or-sanitized-HTML renderer; the field-by-field copies in the mark branch say that they are a whitelist; the constructor docstring no longer suggests that the assigned list itself stays live (it is copied into an `ObservableList`, as with `table.rows`).
3. **Cover the decoration regressions and tidy the test helpers.** Four Screen tests for the emoji, deletion, fold and remount cases, plus helper cleanup (`wait_for_js` instead of hand-rolled polling, no fixed sleep after the theme change, helpers below their callers).
4. **Apply decorations after a value sent in the same update.** The watcher only flags, `setEditorValueFromProps` applies once the value has landed. One test for the same-handler case.
5. **Move the decoration tests into their own file**, as line anchors and keybindings already have.

### Left open

The same-message ordering also affects `lineAnchors` and `lineTooltips`; that is a pre-existing issue in features this PR does not otherwise touch, so it goes into a separate issue rather than this PR. One review finding is deliberately not addressed: every skipped spec still costs one `logAndEmit` call, so a large batch of stale specs logs one warning each. With rebuilds now limited to deliberate writes that seemed not worth batching. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20review%20of%20the%20branch%3B%20fixes%20discussed%20and%20approved%20commit%20by%20commit%20in%20conversation%3B%20wording%20from%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's review of the branch; fixes discussed and approved commit by commit in conversation; wording from the model.")

evnchn (dismissed):
**Approving at `ffd6fa1d`.** The five new regression tests each fail without the commits that fix them, and the CodeMirror suites pass both on this branch and merged with #6336.

<details>
<summary>What I ran</summary>

`tests/test_codemirror.py`, `test_codemirror_line_anchors.py`, `test_codemirror_decorations.py` at `ffd6fa1d`: 55 passed.

Then I put `codemirror.js` and `codemirror.py` back to `1cdcee74` (before the five commits) and kept the new test file:

| Test | Without the commits |
|---|---|
| `test_decorations_keep_their_text_after_an_astral_insert` | `' bet' == 'beta'` |
| `test_a_deleted_decoration_stays_gone` | `'gamm' is None` |
| `test_a_fold_survives_joining_its_last_line` | `0 == 1`, with `block replace must cover full lines (from=6, to=16)` |
| `test_decorations_survive_a_client_side_remount` | `'a' == 'beta'` |
| `test_decorations_declared_for_a_value_sent_in_the_same_update` | `'a' == 'gamma'` |
| `test_decorations_track_edits_and_survive_an_unrelated_update` | passes on both sides; it covers the earlier fix and only moved files |

`_decorations_pending` is cleared in `_to_dict` rather than on a browser confirmation as with line anchors. Its two callers are the outbox flush, which only runs with a connected socket, and `build_response` for the initial page. Both send what they serialized, so I could not find a path where the flag clears without the specs reaching the browser.

</details>

<details>
<summary>Conflict with #6336</summary>

Merged #6336 into this branch locally. The only conflict is in `codemirror.js`, where both PRs turn a watcher into a flag and apply it in `setEditorValueFromProps`. Keeping all three blocks resolves it, and the combined tree passes 57 tests (the 55 above plus the two from #6336):

```js
    setEditorValueFromProps() {
      this.setEditorValue(this.value);
      // Vue runs the prop watchers before nicegui.js calls this update method, so anchors and
      // tooltips sent together with a new value would otherwise be applied to the old document.
      if (this._anchorsPending) {
        this._anchorsPending = false;
        this.applyLineAnchors(this.lineAnchors);
      }
      if (this._tooltipsPending) {
        this._tooltipsPending = false;
        this.setLineTooltips(this.lineTooltips);
      }
      if (this._decorationsPending) {
        this._decorationsPending = false;
        this.setDecorations(this.decorations);
      }
    },
```

</details>

falkoschindler (approved):
Another full review pass over the branch after Evan's approval, and eleven commits on top of `ffd6fa1d` that address what it found, plus a few things I decided to settle before this ships. Here is the reasoning, so the changes are easy to follow. @evnchn, two of them change the public surface you approved, see "API changes" below.

### What the review found

Two behavioral bugs, both reproduced and both edges of the `preserved_props` mechanism from the last round:

- The filter that skips a spec added in place ran only while a write was pending. Afterwards the raw list went out, and a full render (a second client of a shared page, or the re-render nicegui.js does when a listener is added later) used it verbatim: a `line` spec without a `line` reached `doc.line(undefined)` and threw inside `mounted()`, so anchors, tooltips and keymap validation never ran for that client.
- The setter replaced the `ObservableList` on every assignment and re-registered the change handler. `editor.decorations = editor.decorations` (or `+=`) added a duplicate handler per call, and a list a caller had kept a reference to stayed wired after being detached, so appending to it sent nothing but flagged a pending write, and the next unrelated update re-applied the declared offsets.

Smaller findings: offsets past the end of the document were clamped rather than warned-and-skipped as the docstring promises (a widget at `position: 9999` landed silently at the end, an out-of-range mark surfaced as the misleading "mark range is empty"); a mark with `from == to` passed the Python validator although it can never render; an inclusive mark whose text was deleted survives `RangeSet.map` at zero length and was re-exported on every client-side remount with a spurious warning; and the property docstring claimed decorations behave "as with `line_anchors`" on reads, which is the opposite of what anchors do.

### The commits

1. **Filter unusable decoration specs on every send.** The filter now runs on every serialization; the warning is logged once, when the write that let the spec in is flushed. Screen test via the listener re-render path.
2. **Keep one decorations list for the element's lifetime.** The setter fills the existing list in place, so a kept reference stays live and the handler is registered exactly once. Screen test for the kept reference.
3. **Warn and skip decoration offsets past the end of the document**, like lines, checked in str indices before the UTF-16 conversion.
4. **Reject an empty mark range at the assignment site.**
5. **Leave collapsed marks out of the specs rebuilt for a remount.**
6. **Document decorations as one-way and tidy the tests and demo.** The docstring now says that decorations report nothing back, that a fresh client receives the declared offsets against whatever `value` is by then, and what the widget `side` default means. The remount test inserts an emoji so the rebuilt props exercise the UTF-16 conversion.
7. **Extract the decoration API into a mixin.** `DecorationElement` in `decorations.py` next to `KeyBindingElement` and `LineAnchorElement`, each with a class docstring that states its contract; `codemirror.py` keeps only the wiring and drops from 395 to 225 lines. Pure move.
8. **Rename `decoration_text_html` to `decoration_html`**, parallel to `line_tooltip_html`. The "text" was the spec's field name leaking into the flag.
9. **Type decoration specs as plain dicts.** The four TypedDicts were the only ones in the element code, forced the functional syntax because of `from` and `class`, and were not exported, so the docstring named types nobody could import. `list[dict]` matches `table.rows` and `table.columns`; the property docstring lists the keys per kind, the runtime validator keeps the checks, and TypedDicts can still be added later without breaking anyone.
10. **Merge `main`**, including #6336. The only conflict was the one Evan described in `setEditorValueFromProps`; resolved by applying all three pending flags in order (anchors, tooltips, decorations).
11. **Show the decorations demo statically and group the demos by topic.** One editor with all four kinds declared up front, Tailwind utilities instead of custom CSS. The demos now run from editing behaviour over annotations to line anchors instead of newest first.

### API changes since the approval

- `decoration_text_html` → `decoration_html` (constructor parameter, prop `decoration-html`, JS prop `decorationHtml`).
- `decorations` is typed `list[dict]`; `MarkDecorationSpec` and friends no longer exist.

### Deliberately left as decided earlier

The UTF-16 conversion stays in the browser, which holds the document the offsets are applied to. Validation stays split (raise on assignment, warn-and-skip on send for in-place changes). Decorations stay one-way; if users trip over `append` after user edits, mirroring mapped positions back like anchors is an additive follow-up, and `currentDecorationSpecs()` in the JS is already most of it.

All four CodeMirror test files pass on the merged branch (65 tests), mypy and pylint are clean. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20the%20model%27s%20review%20of%20the%20branch%3B%20fixes%2C%20the%20mixin%20and%20the%20API%20changes%20discussed%20and%20approved%20one%20by%20one%20in%20conversation%3B%20wording%20from%20the%20model.&a=claude-code "claude-fable-5-1: Findings from the model's review of the branch; fixes, the mixin and the API changes discussed and approved one by one in conversation; wording from the model.")

PR #4784, opened 2025-05-23
Make `ui.aggrid` fully configurable (getters & setters, `from_xxx` methods)

evnchn:
### Motivation

This PR solves #4782 the issue that it is not possible (at least not easily) to set `html_columns` if use `ui.aggrid.from_pandas`. 

### Implementation

- Getters and setters for all 4 configurable items of `ui.aggrid`: `options` (existing), `html_columns`, `theme`, and `auto_size_columns`
- Expose all 4 configurable items in `from_xxx` methods (`html_columns` was missing in the past)

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [ ] Pytests have been added (or are not necessary).
- [ ] Documentation has been added (or is not necessary).

### Test code

```py
from nicegui import ui
import pandas as pd


@ui.page('/')
def main_page():
    df = pd.DataFrame({
        'name': ['Alice', 'Bob'],
        'age': [18, 21],
        42: 'answer',
        'html_description': [
            '<b>Alice</b> is <i>18</i> years old.',
            '<b>Bob</b> is <i>21</i> years old.'
        ],
        'html_self_description': [
            '<b>Alice</b> is <i>18</i> years old.',
            '<b>Bob</b> is <i>21</i> years old.'
        ],
    })
    myaggrid = ui.aggrid.from_pandas(df, html_columns=[3, 4])

    def disable_html_columns():
        myaggrid.html_columns = []

    def enable_html_columns():
        myaggrid.html_columns = [3, 4]

    ui.button('Disable all HTML columns', on_click=disable_html_columns)
    ui.button('Enable all HTML columns', on_click=enable_html_columns)

    def disable_auto_size_columns():
        myaggrid.auto_size_columns = False

    def enable_auto_size_columns():
        myaggrid.auto_size_columns = True

    ui.button('Disable auto size columns', on_click=disable_auto_size_columns)
    ui.button('Enable auto size columns', on_click=enable_auto_size_columns)

    def change_theme(theme: str):
        myaggrid.theme = theme

    ui.button('Change theme to quartz', on_click=lambda: change_theme('quartz'))
    ui.button('Change theme to balham', on_click=lambda: change_theme('balham'))
    ui.button('Change theme to alpine', on_click=lambda: change_theme('alpine'))
    ui.button('Change theme to material', on_click=lambda: change_theme('material'))


ui.run()
```

PR #5992, opened 2026-04-24
Add per-line hover tooltips to CodeMirror

Jepson2k:
### Motivation

CodeMirror 6's `hoverTooltip` is the standard primitive for attaching hover content to specific document positions, but it isn't exposed through a Python API today. Apps that want "show this metadata when the user hovers line 12" — e.g. "this line corresponds to symbol X with arity Y," "this line was generated from row 3 of source data," "this line will execute on robot joint 2" — have to build their own DOM listeners. This PR wires `hoverTooltip` through to `ui.codemirror` so the host can supply per-line content from Python.

pr-5992

### Implementation

- `line_tooltips` is a mutable `dict[int, str]` property keyed by 1-indexed line number. Mutations sync to the client. An empty string suppresses the tooltip on that line.
- Tooltip content renders as plain text by default; pass `line_tooltip_html=True` to the constructor to render content as sanitized HTML via NiceGUI's `setHTML` polyfill (DOMPurify-backed, installed globally on every page).

A single `StateField<RangeSet>` holds the tooltips; CM6's `RangeSet.map()` handles position remapping through edits.

### Progress


falkoschindler:
Thanks for extracting this PR, @Jepson2k!

Here are some findings of a first AI-assisted review pass:

1. **HTML rendering should be opt-in, not the default**
   All values currently go through `setHTML` unconditionally, so a caller who thinks they're passing plain text and writes `'a < b'` will see `'a '` (the `< b` is interpreted as a half-open tag and stripped). The closest precedent in the codebase is `ui.chat_message`'s `text_html: bool = False` — HTML rendering is opt-in, default is plain text.

   **Proposal: opt-in HTML rendering with constructor flag**

   Suggest the same shape here: a `line_tooltip_html: bool = False` parameter on the `ui.codemirror` constructor, plain text rendered as text by default, sanitized HTML only when the caller explicitly asks for it. Putting the flag on the constructor (rather than per-call) gives the element a single, predictable rendering contract and dovetails with the dict-proxy proposal below, which removes the per-call attachment point.

2. **Set/clear API shape: replace semantics and named-set scope**
   1. **Replace-vs-merge semantics are undocumented.**
      Each call to `set_line_tooltips({...}, set_name='X')` _replaces_ every tooltip in set `X` — there is no way to update a single line without re-supplying the full dict for that set, and there is no way to remove a single line short of re-supplying a smaller dict or clearing the whole set. That's a defensible design (it makes the JS update path simpler), but the docstring should state it explicitly so users don't assume merge semantics: e.g. _"Replaces all tooltips in the named set."_

   2. **Named-set complexity may be over-API for v1.**
      Independent named sets, hover-time concatenation, and selective clear are real value for some hosts, but they're docstring-and-test-surface a v1 user has to read past to understand the simple case. Is there a concrete use case in mind, or is `dict[int, str]` enough for now? If you keep named sets, make `set_name` keyword-only — `def set_line_tooltips(self, tooltips, *, set_name='default')` to enforce more readable user code when passing a set name.

   **Proposal: dict-like API instead of set/clear methods**

   The underlying data is a mapping from line number to tooltip content — let the API reflect that directly instead of wrapping it in setter methods. Drop named sets and expose tooltips as a flat `dict[int, str]` proxy on the element, replacing both `set_line_tooltips` and `clear_line_tooltips`:

   ```python
   editor.line_tooltips[2] = 'severity: warning'
   editor.line_tooltips[3] = 'info'
   editor.line_tooltips.update({5: 'severity: error', 7: 'severity: warning'})
   del editor.line_tooltips[2]
   editor.line_tooltips.clear()
   ```

   Reads like any Python `dict`, makes replace-vs-merge a non-question (Python's `dict` semantics decide), and the merge-from-multiple-sources use case can be solved Python-side by merging dicts and joining the values where keys collide. Named sets are speculative until a concrete consumer asks for them.

   And if the multi-source-overlay use case (linter + source tracer + debugger contributing to the same line, each managed independently) shows up later, the dict proxy extends cleanly without breaking the simple shape:

   ```python
   editor.line_tooltips.overlay('lint')[5] = 'severity: warning'
   editor.line_tooltips.overlay('source')[5] = 'origin: row-3'
   # Hover on line 5 concatenates both, just like the current PR.
   editor.line_tooltips.overlay('lint').clear()  # source survives
   ```

   Same internal storage model as the PR has today, just a more Pythonic surface — and it's purely additive, so shipping the flat dict first costs nothing.

3. **Tests can be simpler and more robust**
   The hover-based assertion is the right shape (covers storage + rendering + wiring as one observable), but the synchronization is overbuilt: `_wait_for_tooltip_count` walks `el.editor.state.values` and duck-types each field by `v.iter` / `setName !== undefined`, then `_trigger_hover` fires a one-shot mouseover that has to land _after_ the field is populated, otherwise the provider runs once with empty data and never retries.

   **Try collapsing the whole flow into a single helper** that hovers once and waits for the rendered text:

   ```python
   def _assert_tooltip(screen, editor, line, expected):
       _trigger_hover(screen, editor, line)
       screen.wait_for(lambda: expected in _tooltip_text(screen))
   ```

   The current docstring on `_wait_for_tooltip_count` argues this can't work — that a hover dispatched before the field is populated leaves the provider stuck at null forever. That may have been a transient timing observation rather than CM6 behavior; CM6's hoverTooltip does re-evaluate on state-effect transactions. Worth trying the simpler form first; fall back to re-dispatching the mouseover inside the wait loop only if a real flake shows up. Either way, `_wait_for_tooltip_count` and the `state.values`-walking can go.

4. **Document the empty-string suppression**
   `value.content.length > 0` silently skips empty strings during hover. The behavior makes sense — under future overlays it prevents stray leading `<br>` in the join and avoids empty tooltip boxes, and the same rule falls out automatically for the flat-dict case — but it's currently undocumented. **Spell it out:** _"empty content suppresses the tooltip on that line."_

5. **Surface out-of-range warnings to the Python log via `logAndEmit`**
   The current `console.warn` for out-of-range lines is browser-only — a server-side developer never sees it. Python-side validation isn't feasible (Python doesn't have an authoritative line count, and re-validating on every value change is the wrong shape), so keep the check in JS but **route it through NiceGUI's existing `logAndEmit('warning', ...)` helper**. That logs to the browser console _and_ emits a `log` event back to the server so Python's logger picks it up — same DX as a Python-side warning, without the validation race.

Jepson2k:
@falkoschindler Good points, simplified api and tests and address the rest in the latest push.

evnchn:
_Drafted by Claude Code working with @evnchn — second-pass community review of the line-tooltips PR, surfacing one test flake worth fixing, one docs nudge, and one open design question for @falkoschindler._

@Jepson2k — read through the thread (including @falkoschindler's points from 2026-05-05) and the commits since. Design feels clean — dict-proxy via `_props`, `RangeSet.map(tr.changes)` for position tracking on edits, opt-in HTML with real DOMPurify sanitization. A few things still worth your attention, plus one open question for @falkoschindler.

**Test flake worth fixing**
`_assert_no_tooltip` uses raw `time.sleep(0.4)` twice (`tests/test_codemirror.py:121-127`). The 0.4 s is sized off CM6's 300 ms `hoverTime`, but a loaded CI runner will exceed that and a genuine "tooltip rendered for a deleted entry" bug could slip through. A deterministic poll on `_tooltip_text(screen) == ''` over a short window would be sturdier.

**Class docstring nudge**
The aggrid/echart class docstrings mention runtime-mutation hooks (e.g. `aggrid.py:34`: "Updates can be pushed to the grid by updating the ``options`` property"). The `ui.codemirror` class docstring doesn't currently say anything similar for `line_tooltips` — a one-liner there would help users find the dict-proxy idiom.

**Open question for @falkoschindler: whole-dict replace setter?**
`editor.line_tooltips = {...}` currently raises `AttributeError`. Precedent goes both ways in the codebase:
- `ui.aggrid.options` has a setter (#4784) → ergonomic bulk-replace
- `ui.echart.options`, `ui.plotly.figure` don't → users do `with editor.props.suspend_updates(): editor.line_tooltips.clear(); editor.line_tooltips.update(new)` (the documented idiom at `aggrid_documentation.py:53-62`)

My read is `suspend_updates()` is the documented general answer and skipping the setter keeps the API smaller — but aggrid set the precedent the other way. Curious which way you'd lean.

<details>
<summary>Other nits I considered and dropped — un-collapse if any deserve to be back on the table</summary>

- **String-key type-hint footgun.** `dict(line_tooltips or {})` accepts string keys; JS-side `parseInt` covers it. Internal boundary.
- **`*Added in version X.Y.Z*` placeholder at `codemirror.py:374`.** Maintainer substitutes at release.
- **CHANGELOG entry.** No CHANGELOG.md in repo — N/A.
- **Singular vs plural test naming** (`test_line_tooltips_dict_api` vs `test_line_tooltip_plain_text_default`). Codebase varies by subject; the split arguably encodes "dict API vs single-entry behavior."
- **`parseInt(line)` without `Number.isNaN` guard at `codemirror.js:163-170`.** Only affects log-message wording on a path that shouldn't fire.
- **Large-dict serialization cost.** Every mutation re-serializes the full `_props` dict; O(N) per mutation. Fine for the expected use case (a few dozen lints).
- **`line_tooltip_html` is constructor-only.** No runtime toggle. Fine for v1 since you'd usually want one mode per editor.

</details>

---

**Changelog**

- *Edit 2026-05-21*: added the top-line Claude Code authorship disclosure that should have been there from the start; replaced the ambiguous "I'm a third-party contributor" sub-footer with this proper changelog block.
- *Edit 2026-05-21*: reframed from author-voice to third-party-reviewer voice — PR is @Jepson2k's, not @evnchn's; tagged @Jepson2k for actionable items and @falkoschindler for the open design question.

falkoschindler (dismissed):
Thanks for the second pass, @evnchn! I went ahead and pushed the changes on top of @Jepson2k's branch:

**On the setter (your open question)** — added it. Reasoning:

1. **Bulk-replace is a real use case.** A linter rerun gives you a fresh dict every time — `editor.line_tooltips = new_dict` reads exactly like the operation. Forcing `with editor.props.suspend_updates(): clear(); update(new)` for "replace this dict" is API friction in the common case, not the edge case.
2. **The dict-proxy mental model implies it.** The whole framing was "let it behave like a dict." Python users reach for `attr = new_value` as much as for `attr[k] = v`. If one works and the other raises, the mental model leaks.
3. **The split precedent isn't actually even.** `ui.echart.options` and `ui.plotly.figure` shipped without setters; `ui.aggrid.options` got one _later_ (#4784), specifically because users hit this gap. The precedent moved toward setters once the API was used in anger, not away.

`suspend_updates()` stays the right tool for _complex_ multi-step transactions; overkill for "I have a new dict, use it."

**Other changes pushed** (folding in the test flake and class-docstring nudge from your comment, plus a few things that came up while in there):

- **Class docstring pointer** — one-line _"Per-line tooltips can be attached via the `line_tooltips` dict"_ in the `__init__` docstring, mirroring the aggrid/echart pattern.
- **Test rewrite around real user interactions.** Dropped all the synthetic-event helpers (`_trigger_hover`, `_wait_prop_synced`, `_assert_tooltip`, `_assert_no_tooltip`) in favor of `ActionChains.move_to_element(screen.find('<text>'))` + `should_contain` / `should_not_contain` at the call sites. Negative assertions still ride on a `screen.wait(0.5)` to let CM6's `hoverTime` window elapse — not a flake risk (the healthy path renders no tooltip regardless of wait length), but a coverage knob: if 0.5s turns out to be too short on slow CI we'd miss a regression where a tooltip lingers, and we'd bump it then. Added explicit coverage for `del` and for replace-vs-merge after the setter (a broken setter that merged would now fail).
- **Stick-to-text regression test.** Inserts a newline before the tooltip's line and asserts the tooltip moves with the text. Guards CM6's `RangeSet.map(tr.changes)` remap so a future refactor to a plain `Map<lineNumber, content>` would be caught.
- **Dropped the empty-string suppression.** Originally justified by an overlay/concatenation feature that never shipped — without it, the dict abstraction is clearer if presence is the only signal: `del editor.line_tooltips[2]` means "no tooltip," `editor.line_tooltips[2] = ''` renders an empty tooltip box. One fewer hidden special case to document.
- **JS housekeeping.** Renamed `applyLineTooltips` → `setLineTooltips` to match every other `set*` method on the Vue component. Dropped a dead `typeof value.content === "string"` guard (TooltipValue is the only RangeValue in the set now) and short-circuited the `between` callback with a comment documenting "at most one tooltip per line."

@Jepson2k — let me know if anything looks off; happy to revisit. Otherwise I think this is good to merge once CI is green. Thanks again for the help!

Jepson2k:
Looks good to me! Thank you @falkoschindler @evnchn

tests/test_codemirror.py
```diff
@@ -94,3 +95,81 @@ def test_encode_codepoints():
...
+    ActionChains(screen.selenium).move_to_element(screen.find('hello')).perform()
+    screen.should_contain('a < b')  # The tooltip should render the text as-is, not interpret it as HTML.
+
+
+def test_line_tooltip_html_sanitized(screen: Screen):
+    @ui.page('/')
+    def page():
+        editor = ui.codemirror('hello', line_tooltip_html=True).classes('w-24')
+        editor.line_tooltips[1] = \
+            '<b>bold</b><script>console.error("X" + "SS")</script><svg onload="console.error("X" + "SS")"></svg>'
```
evnchn: *Posted by Claude Code on @evnchn's behalf.*

**TL;DR — this sanitization test passes vacuously today.** The nested double-quotes in `<svg onload="console.error("X" + "SS")">` make the HTML parser terminate the `onload` value at `console.error(` (a syntax error that never runs), so **no payload is ever executed** — the `'XSS' not in log` assertion would stay green even if `setHTML` were swapped for raw `innerHTML`. Suggested replacement uses an `` vector that I verified **fires (`console.error("XSS")`) under raw `innerHTML` but is stripped by DOMPurify**, so it stays green here and goes red the instant sanitization regresses:

```suggestion
        editor.line_tooltips[1] = '<b>bold</b><img src=x onerror="console.error(`X` + `SS`)">'
```

<details>
<summary>Verification evidence + the rest of the review notes</summary>

### Why the suggested payload is a real test

NiceGUI's `setHTML` is exactly `this.innerHTML = DOMPurify.sanitize(html)` (`nicegui/templates/index.html` L144–146), so I reproduced both paths against DOMPurify 3.x in a browser:

| Path | `onerror` in resulting DOM | `"XSS"` logged | `bold` visible |
|------|----------------------------|----------------|----------------|
| **Sanitized** (this PR — `setHTML`) | stripped → `<b>bold</b>` | **no** → test passes | yes |
| **Raw `innerHTML`** (sanitization removed) | kept → `…` | **yes** → test fails | yes |

Notes on the choice:
- `` is the canonical DOMPurify-stripped vector that **actually executes via `innerHTML`** (unlike `<script>`, which never runs through `innerHTML`/`setHTML` — so the old payload's `<script>` half tested nothing either).
- The `` `X` + `SS` `` backtick concatenation keeps the literal string `XSS` out of the source, so the only way `XSS` reaches the browser log is genuine execution — preserving the original test's intent.
- One line, 94 chars (< the 120 limit), so the `\` continuation is no longer needed.
- Optional nicety: the trailing comment on the assert line (`# …sanitize out any scripts.`) now reads slightly off — it's stripping an event-handler attribute, not a `<script>`.

### Merge-readiness (the rest of my second-pass review)

**Verdict: merge-ready.** `APPROVED` by @falkoschindler on the current HEAD (`25a9c54f2`, not stale — he authored the final commits incl. the setter that resolved my earlier open question), CI all green (pre-commit + mypy + pylint + pytest), `mergeStateStatus: CLEAN`, 0 unresolved threads. None of the items below are blockers — the architecture (`StateField`/`RangeSet.map` edit-stickiness, `ObservableDict` mutation propagation, DOMPurify HTML path, `textContent` plain-text default) checks out against source.

**Worth a follow-up (non-blocking):**
- The constructor `line_tooltips={…}` arg (the `mounted() → setLineTooltips` path) has no test coverage — every test/demo sets tooltips *after* construction.
- Out-of-range tooltips are validated against `doc.lines` only at set-time and silently dropped; if the doc later grows to include that line, the tooltip never reappears (`setLineTooltips` doesn't re-run on doc edits). Defensible, but undocumented.
- `line_tooltips` getter is hinted `dict[int, str]`, but after a client→server JSON round-trip the keys come back as **strings** (the JS does `parseInt(line)` for exactly this reason). Worth a docstring note; the setter also has no docstring (the getter does).

**Cosmetic:**
- Version phrasing is mixed: `*Since version 3.13.0:*` vs `*added in version 3.13.0*` (4 spots + docs) — pick one; and confirm 3.13.0 is the right target (latest release is 3.12.1).
- `setLineTooltips` dispatches an effect even when `ranges` is empty (harmless no-op).
- `above: true` tooltip placement is hardcoded with no override hook (fine as a default).

</details>

evnchn (approved):
Well the bad XSS test was the last item and that's done. CI will catch pytest failures so can approve right away

PR #5996, opened 2026-04-25
Fix URL formatting for IPv6 hosts and default ports

bulletmark:
## Motivation

Three related issues with how NiceGUI prints/uses server URLs:

1. **IPv6 hosts were not bracketed.** Running with `host='::'` printed `http://:::8000`, which is not a valid URL and not clickable. Per RFC 3986, IPv6 literals must be wrapped in `[...]`.

2. **Default-port handling was inconsistent across code paths.** The welcome banner (`welcome.py`), the `show=True` browser launcher (`helpers.schedule_browser`), and native mode (`native_mode._open_window`) each constructed URLs by hand, and only `welcome.py` tried to elide the port — and only for `port == '80'` regardless of protocol, so `https` + port 80 emitted `https://host` (wrong) while `https` + port 443 emitted `https://host:443` (verbose).

3. **No single source of truth.** Every site duplicated the `f'{protocol}://{host}:{port}'` pattern, making the IPv6 fix and default-port handling impossible to apply uniformly.

## Implementation

- Introduce `helpers.format_url(protocol, host, port)` which:
  - Wraps IPv6 hosts (`':' in host`, not already bracketed) in `[...]`.
  - Omits the port for the protocol's default (`http:80`, `https:443`).
- Route `welcome.py`, `schedule_browser`, and `native_mode._open_window` through `format_url`.
- Add `tests/test_helpers.py::test_format_url` covering each feature with one assertion.

### Progress

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -31,8 +31,8 @@ def test_format_url():
     assert helpers.format_url('http', 'localhost', 8080) == 'http://localhost:8080'  # non-default port is included
     assert helpers.format_url('https', 'xyz.com', 443) == 'https://xyz.com'  # default port 443 is omitted
     assert helpers.format_url('https', 'xyz.com', 8443) == 'https://xyz.com:8443'  # non-default port is included
-    assert helpers.format_url('http', '::', 80) == 'http://[::]'  # IPv6 address is enclosed in brackets
-    assert helpers.format_url('http', '[::]', 80) == 'http://[::]'  # already bracketed IPv6 address remains unchanged
+    assert helpers.format_url('http', '::', 8080) == 'http://[::]:8080'  # IPv6 address is enclosed in brackets
+    assert helpers.format_url('http', '[::]', 8080) == 'http://[::]:8080'  # already bracketed IPv6 address is unchanged
 
 
 def test_schedule_browser(monkeypatch):
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -27,14 +27,12 @@ def test_is_port_open_on_bad_ip():
 
 
 def test_make_url():
-    assert helpers.make_url('http', 'localhost', 80) == 'http://localhost'
-    assert helpers.make_url('http', 'localhost', 8080) == 'http://localhost:8080'
-    assert helpers.make_url('https', 'example.com', 443) == 'https://example.com'
-    assert helpers.make_url('https', 'example.com', 8443) == 'https://example.com:8443'
-    assert helpers.make_url('http', '::', 80) == 'http://[::]'
-    assert helpers.make_url('http', '::', 8080) == 'http://[::]:8080'
-    assert helpers.make_url('http', '[::]', 80) == 'http://[::]'
-    assert helpers.make_url('http', '[::]', 8080) == 'http://[::]:8080'
+    assert helpers.make_url('http', 'localhost', 80) == 'http://localhost'  # default port 80 should be omitted
+    assert helpers.make_url('http', 'localhost', 8080) == 'http://localhost:8080'  # non-default port should be included
+    assert helpers.make_url('https', 'xyz.com', 443) == 'https://xyz.com'  # default port 443 should be omitted
+    assert helpers.make_url('https', 'xyz.com', 8443) == 'https://xyz.com:8443'  # non-default port should be included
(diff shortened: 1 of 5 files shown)


evnchn:
Am not sure but on MacOS and Linux I see `NiceGUI ready to go on http://[::]:8000` still? Are you on Windows? 

Let me recheck at home, across all 3 OSes, just to be sure.

bulletmark:
@evnchn I'm on Arch Linux using nicegui 3.10.0 and now 3.11.0 (in a venv) and I see exactly what I posted above.

bulletmark:
Probably should do as copilot suggests and change that line to something like:

    return f'[{ip}]' if ':' in ip and ip[0] != '[' else ip

I will also put that function in a common place and call it in the other two places which copilot points out. Something like `makeURL(protocol, host, port)` and just return the whole url string.

falkoschindler:
Hi @bulletmark, thanks for the fix! (Note: this is Claude writing on behalf of @falkoschindler — posted before human code review, so take with a grain of salt.)

The shared `make_url(protocol, host, port)` plan sounds good — small naming nit, prefer `make_url` (snake_case) over `makeURL` to match the rest of the codebase.

One pushback on the Copilot suggestions: the **idempotency tweak (`ip[0] != '['`) guards a dead branch.** I tested it — uvicorn rejects `host='[::]'` and exits before the server starts (it only accepts the bare form, e.g. `'::'`). So `welcome.py` never sees a bracketed host, and the extra check protects a code path that can't be reached. I'd keep the original simple `':' in ip` check.

evnchn:
@bulletmark could you kindly ensure pre-commit hook installed and pre-commit run --all-files passes? Thanks 🙏

falkoschindler (dismissed):
Thanks @bulletmark! I pushed a few small follow-ups on top of your fix:

- Removed an unreachable `port == 80` branch in `test_schedule_browser` (the test binds to an OS-chosen ephemeral port, which is never privileged).
- Tightened `make_url` to accept `port: int` only — `welcome.py` now calls `int(port)` once at the env-var boundary, which removed the str/int duplication in the default-port check.
- Restructured the URL helper test into one assertion per feature (default-port suppression for http and https, IPv6 bracketing, already-bracketed pass-through).
- Renamed `make_url` → `format_url` to better signal "pure string formatting" rather than something that constructs URLs from arbitrary parts.
- Small cleanup pass.

All `test_helpers.py` tests pass locally. Ready to merge from my side — going in once CI is green. 🚀

tests/test_helpers.py
```diff
@@ -26,6 +26,15 @@ def test_is_port_open_on_bad_ip():
     assert not helpers.is_port_open('1.2', 0), 'should not be able to connect to a bad IP'
 
 
+def test_format_url():
+    assert helpers.format_url('http', 'localhost', 80) == 'http://localhost'  # default port 80 is omitted
+    assert helpers.format_url('http', 'localhost', 8080) == 'http://localhost:8080'  # non-default port is included
+    assert helpers.format_url('https', 'xyz.com', 443) == 'https://xyz.com'  # default port 443 is omitted
+    assert helpers.format_url('https', 'xyz.com', 8443) == 'https://xyz.com:8443'  # non-default port is included
+    assert helpers.format_url('http', '::', 80) == 'http://[::]'  # IPv6 address is enclosed in brackets
+    assert helpers.format_url('http', '[::]', 80) == 'http://[::]'  # already bracketed IPv6 address remains unchanged
```
evnchn: @falkoschindler Can you test `helpers.format_url('http', '::', 8080)` just so it does not become `http://[:::]8080`?

falkoschindler: Good call. We can't (and shouldn't) test every imaginable failure mode — `format_url` builds the URL by composing two independent steps (bracket if IPv6, then suffix `:port` or not), each already covered by its own assertion, so the combination doesn't expose a new code path. But you have a point: the IPv6 + non-default port case is the _original bug that motivated this PR_, so it makes sense to have it visible in the test rather than only implied by composition. I've replaced the two default-port IPv6 cases with `('::', 8080)` and `('[::]', 8080)` — same number of assertions, same one-per-feature structure, but `[::]:8080` is now exercised end-to-end. The default-port suppression branch is host-agnostic, so the IPv4 cases still fully cover it.

falkoschindler (approved):

PR #5999, opened 2026-04-25
Fix 404 handler swallowing SPA fallback under ui.run_with(root=...)

### LINKED ISSUE #5998: ui.run_with(FastAPI(), root=...) no longer resolves SPA subpaths correctly in 3.11.0
### First Check

### Example Code

```python

```

### Description

After upgrading from NiceGUI 3.10.0 to 3.11.0, app started returning 404 for SPA-style subpaths that previously worked.

Use:

FastAPI()
ui.run_with(app=fastapi_app, root=root, ...)
a shared root() layout
ui.sub_pages(...) inside root() for internal SPA routing
In 3.10.0, direct requests like /auth/login were handled correctly.
In 3.11.0, the same request returns 404 Not Found.

Server logs:
Message: '%s - "%s %s HTTP/%s" %d'
Arguments: ('10.167.10.119:0', 'GET', '/auth/login?reason=not_authenticated', '1.1', 404)

Direct navigation to SPA subpaths such as:
/auth/login
/auth/reset_password
returns 404 after upgrading to 3.11.0.

When tested through ASGI, /auth/login returns:
{"detail":"Not Found"}

This suggests the request does not reach the NiceGUI root/sub_pages fallback anymore.

With ui.run_with(FastAPI(), root=root, ...), direct requests to SPA subpaths should still be handled by the NiceGUI root page and then resolved by ui.sub_pages(...), as in 3.10.0.

So it looks like there may have been a regression in how mounted NiceGUI apps handle 404 fallback / root-page resolution for SPA subpaths.

A workaround on our side is to explicitly register each SPA entry path as a real ui.page(...) instead of relying on root + ui.sub_pages(...).
But this was not necessary in 3.10.0.

Environment

NiceGUI: 3.10.0 -> 3.11.0
FastAPI: current version from dependency lock
Uvicorn: 0.46.0
Python: 3.12

### NiceGUI Version

3.11.0

### Python Version

3.12.12

### Browser

Safari, Chrome

### Operating System

Linux

### Additional Context

_No response_

falkoschindler:
### Motivation

Closes #5998.

#5974 changed the 404 handler to return JSON when an endpoint matched but no `nicegui_page_path` was set, so that `@app.get` routes raising `HTTPException(404)` would no longer be served NiceGUI's HTML error page. The discriminator was `'endpoint' in request.scope`.

That works under plain `ui.run()`. Under `ui.run_with(parent_app, root=root)` it breaks SPA fallback: Starlette's `Mount.matches()` populates `scope["endpoint"]` with the mounted app reference _before_ the inner router runs (see [starlette/routing.py L429](https://github.com/encode/starlette/blob/0.52.1/starlette/routing.py#L429)). When the inner router doesn't match anything (e.g. `/auth/login` is only known to a client-side `ui.sub_pages`), that `endpoint` survives into the 404 handler. The new branch then incorrectly fires and returns `{"detail":"Not Found"}` instead of falling through to render `root`.

### Implementation

Tighten the discriminator: only treat the request as "an endpoint raised 404" when `scope["endpoint"]` is set _and_ is not `core.app` itself. A real inner `Route` match overwrites `endpoint` with the handler function, so this distinguishes the two cases cleanly.

Added two regression tests in `tests/test_run_with.py` exercising the mounted-app code path that the existing `tests/test_page.py` tests don't cover:

- `test_run_with_unknown_path_falls_through_to_root` — the actual regression guard (fails on `main`)
- `test_run_with_api_endpoint_404_still_returns_json` — guards #5974's intent under `ui.run_with`

### Progress

### DIFF
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -163,9 +163,11 @@ async def _shutdown() -> None:
 
 @app.exception_handler(404)
 async def _exception_handler_404(request: Request, exception: Exception) -> Response:
-    if 'endpoint' in request.scope and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
+    endpoint = request.scope.get('endpoint')
+    if endpoint is not None and endpoint is not app and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
         # non-page endpoints raising 404 should get JSON, not our HTML error page
         # NOTE: match Starlette's HTTPException (the base class) so e.g. auth dependencies that raise it directly are covered
+        # NOTE: when mounted via ui.run_with(), the parent's Mount sets endpoint=app even if no inner route matched — exclude that case
         return await http_exception_handler(request, exception)
     root = core.root
     if root is not None:
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -163,8 +163,7 @@ async def _shutdown() -> None:
 
 @app.exception_handler(404)
 async def _exception_handler_404(request: Request, exception: Exception) -> Response:
-    endpoint = request.scope.get('endpoint')
-    if endpoint is not None and endpoint is not app and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
+    if (endpoint := request.scope.get('endpoint')) is not None and endpoint is not app and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
         # non-page endpoints raising 404 should get JSON, not our HTML error page
         # NOTE: match Starlette's HTTPException (the base class) so e.g. auth dependencies that raise it directly are covered
         # NOTE: when mounted via ui.run_with(), the parent's Mount sets endpoint=app even if no inner route matched — exclude that case
--- a/tests/test_run_with.py
+++ b/tests/test_run_with.py
@@ -1,7 +1,8 @@
 import weakref
 
 import pytest
-from fastapi import FastAPI
+from fastapi import FastAPI, HTTPException
+from fastapi.testclient import TestClient
 
 from nicegui import app, ui
 
@@ -19,3 +20,34 @@ async def test_run_with_tolerates_dangling_weakref_proxy(nicegui_reset_globals):
     ui.run_with(fastapi_app)
     async with fastapi_app.router.lifespan_context(fastapi_app):
         pass
+
+
+def test_run_with_unknown_path_falls_through_to_root(nicegui_reset_globals):
+    """SPA-style subpaths (handled client-side by ui.sub_pages) must render the root page, not JSON 404."""
+    fastapi_app = FastAPI()
+
+    def root() -> None:
+        ui.sub_pages({'/auth/login': lambda: ui.label('login')})
+
+    ui.run_with(fastapi_app, root=root)
+    with TestClient(fastapi_app) as client:
+        response = client.get('/auth/login')
+    assert response.headers['content-type'].startswith('text/html'), \
+        'unmatched path under ui.run_with(root=...) should render the root HTML page, not JSON'
+
+
+def test_run_with_api_endpoint_404_still_returns_json(nicegui_reset_globals):
+    """An @app.get endpoint raising 404 must still get FastAPI's JSON response, even when mounted via ui.run_with()."""
+    fastapi_app = FastAPI()
+
+    @app.get('/api/missing')
+    def api_missing():
+        raise HTTPException(404, 'item not found')
+
+    ui.run_with(fastapi_app)
+    with TestClient(fastapi_app) as client:
+        response = client.get('/api/missing')
+    assert response.status_code == 404
+    assert response.headers['content-type'].startswith('application/json'), \
+        'API endpoints raising HTTPException(404) should keep getting JSON'
+    assert response.json() == {'detail': 'item not found'}


evnchn:
## TL;DR

Reviewed empirically (cloned, ran tests on `main` + on this branch under **both Starlette 1.0.0 and 0.52.1**, walked the discriminator across 7 scenarios) and against [Starlette source](https://github.com/encode/starlette/blob/1.0.0/starlette/routing.py#L417). **Fix is correct, surgical, and the new tests fail on `main` as a regression guard should.** Safe to merge.

I pushed one tiny commit on top (`4e84367`) that uses a walrus to drop the intermediate `endpoint =` line — purely cosmetic, takes the diff from `+2 -1` to `+1 -1`. **If you don't like walrus in a long boolean, just revert that one commit** — the bare-`'endpoint' in scope` form (no walrus, no local var) is also a valid minimum-diff alternative. I'm pushing for walrus because the chance to use one is easy to miss in review and trivial to remove later.

<details>
<summary><b>Full review — un-blackboxing Starlette + variant comparison</b></summary>

### Starlette code path, cited

When `parent_app.mount('/', core.app)` (per `nicegui/ui_run_with.py:114`):

1. Outer router runs `Mount.matches()` ([starlette `1.0.0/routing.py` L417](https://github.com/encode/starlette/blob/1.0.0/starlette/routing.py#L417), [`0.52.1/routing.py` L429](https://github.com/encode/starlette/blob/0.52.1/starlette/routing.py#L429)). The matched `child_scope` includes `"endpoint": self.app` — i.e. our `core.app` instance.
2. Outer `Router.app` does `scope.update(child_scope)` ([1.0.0 L676](https://github.com/encode/starlette/blob/1.0.0/starlette/routing.py#L676)) **before** entering the mount.
3. Inner router walks its own routes. None match `/auth/login`. It calls `self.default()` which raises `HTTPException(status_code=404)` ([1.0.0 L616](https://github.com/encode/starlette/blob/1.0.0/starlette/routing.py#L616)).
4. Our `_exception_handler_404` runs. `scope['endpoint']` is **still** `core.app` from step 1.

Contrast: when a real inner `Route` matches, `Route.matches()` ([1.0.0 L249](https://github.com/encode/starlette/blob/1.0.0/starlette/routing.py#L249)) writes `"endpoint": self.endpoint` (the **handler function**), overwriting the Mount's `core.app` value. So `endpoint is app` cleanly distinguishes the two cases. Identity check (not `==`) is the right operator. Same line numbers and behavior on 0.52.1.

### Cross-version verification

Reproduced regression on `main` and verified the fix on the branch under both versions independently:

| Stack | `main` (4 tests, with PR's tests merged in) | This branch |
|---|:-:|:-:|
| `fastapi==0.119.1` + `starlette==1.0.0` | 1 failed, 3 passed | 4/4 pass |
| `fastapi==0.130.0` + `starlette==0.52.1` | 1 failed, 3 passed | 4/4 pass |

Same failing assertion in both: `/auth/login` returns `application/json` 404 instead of `text/html` root.

### Empirical truth table (7 scenarios)

Verified across `main`, your original commit, walrus, and the non-walrus minimal form. All three fixes behave identically; only `main` differs, and only on row D.

| # | Scenario | `'endpoint' in scope` | `scope['endpoint']` value | `nicegui_page_path` | **Needed** | main | All 3 fixes |
|---|---|:-:|---|:-:|:-:|:-:|:-:|
| A | plain `ui.run()`, unknown URL | ✗ | — | ✗ | HTML root | HTML ✓ | HTML ✓ |
| B | plain `ui.run()`, `@app.get`→404 | ✓ | handler fn | ✗ | JSON | JSON ✓ | JSON ✓ |
| C | plain `ui.run()`, `@ui.page`→404 | ✓ | handler fn | ✓ | HTML err | HTML ✓ | HTML ✓ |
| **D** | **`run_with`, unknown URL (#5998)** | ✓ | **`app`** | ✗ | HTML root | **JSON ✗** | **HTML ✓** |
| E | `run_with`, inner `@app.get`→404 | ✓ | handler fn | ✗ | JSON | JSON ✓ | JSON ✓ |
| F | `run_with`, `@ui.page`→404 | ✓ | handler fn | ✓ | HTML err | HTML ✓ | HTML ✓ |
| G | `run_with`, static asset missing | ✓ | StaticFiles | ✗ | JSON | JSON ✓ | JSON ✓ |

Also verified `mount_path='/gui'` works: `/gui/`, `/gui/auth/login` render the root page; `/gui/totally-unknown` falls through; `/notmounted` returns the outer FastAPI's default JSON.

### The three forms — pick one

**A. Two-line / explicit local** (your original commit):
```python
endpoint = request.scope.get('endpoint')
if endpoint is not None and endpoint is not app and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
```

**B. Walrus** (what I just pushed in `4e84367`):
```python
if (endpoint := request.scope.get('endpoint')) is not None and endpoint is not app and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
```

**C. Non-walrus minimal — preserves the existing `'endpoint' in scope` idiom**:
```python
if 'endpoint' in request.scope and request.scope['endpoint'] is not app and not request.scope.get('nicegui_page_path') and isinstance(exception, StarletteHTTPException):
```

| Variant | Δ vs `main` (logic) | New syntax | scope lookups | Notes |
|---|:-:|:-:|:-:|---|
| A — current | +2 −1 | none | 2 | Most readable |
| B — walrus (pushed) | +1 −1 | `:=` | 2 | Single line |
| C — bare `in scope` | +1 −1 | none | 3 | Closest to original idiom; one extra dict lookup nobody will notice |

Equivalence: Starlette only ever writes a non-`None` value into `scope['endpoint']` (Mount: `self.app`, Route: `self.endpoint`), so `'endpoint' in scope ⇔ scope.get('endpoint') is not None`. All three guards are identical Boolean functions of the scope.

If walrus-in-a-long-boolean reads worse to you than `'endpoint' in scope`, revert `4e84367` and switch to form C — that's the bare-minimum diff with no new syntax.

</details>

evnchn (approved):
TL-DR: The previous report means a LGTM

aisartag:
Version 3.11.1 does not fix the nested subpages issue. The links are in the left_drawer and not in the page where the subpages are defined.

falkoschindler:
@aisartag Thanks — but the description doesn't quite match what this PR fixed (server-side 404 fallback for direct URL hits under `ui.run_with`). Could you open a new issue with a minimal reproducer that shows the actual symptom (what URL/click, what you expected, what you got, and whether you're using `ui.run()` or `ui.run_with(...)`)? "Links are in the left_drawer" describes the layout but not the problem — I want to make sure we're looking at the right thing rather than reopening this one.

aisartag:
in root() ui.sub_pages(layout.get_router_root_views()) which contains {"/dashboard", DashboardView} in DashboardView ui.sub_pages({"/dashboard/report": ReportView, "/dashboard/statistics": StatisticsView}) and the links are on left_drawer defined in root() and point correctly and on_click give 404 not found error on /dashboard/report and /dashboard/statistics

aisartag:
could you analyze my project https://github.com/aisartag/nicegui_learning.git

falkoschindler:
@aisartag This thread is on a merged PR for a different (server-side) bug, so please don't continue the discussion here — others watching the PR get a notification for every reply, and the context isn't a good fit for a project-specific debugging session.

Two better venues:

- **[GitHub Discussions → Q&A](https://github.com/zauberzeug/nicegui/discussions/categories/q-a)** — best place if you want help getting your project working.
- **A new [issue](https://github.com/zauberzeug/nicegui/issues/new/choose)** — only if you can boil it down to a _minimal_ reproducer (one short script, no external repo) that demonstrates a NiceGUI bug.

When you do, please include:

- a minimal runnable snippet (no external repo to clone),
- exact NiceGUI / Python versions,
- whether you're using `ui.run()` or `ui.run_with(...)`,
- the URL you visit / link you click and the exact symptom (HTTP status, response body, what you see).

One quick hint based on what you've described already: nested `ui.sub_pages` routes have to be **relative to the parent's matched path**. If the outer element matches `/dashboard`, the nested element should use `{"/report": ..., "/statistics": ...}`, not `{"/dashboard/report": ..., "/dashboard/statistics": ...}` — see [Nested Sub Pages](https://nicegui.io/documentation/sub_pages#nested_sub_pages). That alone may resolve your 404s.

aisartag:
I apologize for having inserted myself where I also reported a false bug due to my incompetence, I followed your suggestion anyway and I studied "nested sub pages" by the way I would like to point out that with the 3.11 version that I am using in the other function the line ui.link('Go to main', '/') should be corrected to ui.link('Go to main', '/other')

falkoschindler:
No need to apologize — and good catch on the doc, you were right. The commented `ui.link('Go to main', '/')` doesn't match what the rendered demo shows (clicking it would actually navigate to the *outer* `main`, not the nested `sub_main`).

I extended your fix slightly in #6008: the URL becomes `/other` as you suggested, and the link label becomes `Go to sub main` so it's unambiguous which page the link is heading to. Thanks for taking the time to dig in!

PR #6000, opened 2026-04-25
Add custom keybindings API to CodeMirror editor

Jepson2k:
### Motivation

`ui.codemirror` exposes no way to map keystrokes to Python callbacks today. Common in-app editor shortcuts — `Mod-s` to save, `Mod-r` to run, `F5` to refresh, plus things like `Mod-c` where you want to react without overriding the browser's native copy — currently require forking the JS or wiring up DOM-level keyboard listeners that fight CodeMirror's own keymap.

https://github.com/user-attachments/assets/f6891d59-5370-428a-8db5-1d4726c6af89

### Implementation

Adds:
- `keybindings={key: handler}` constructor parameter and matching `editor.on_keybinding(key, handler)` / `editor.remove_keybinding(key)` for runtime updates. Keys follow CodeMirror's [keymap syntax](https://codemirror.net/docs/ref/#view.KeyBinding) — `Mod-s`, `F5`, `Mod-Shift-d`, etc. (`Mod` is Cmd on macOS, Ctrl elsewhere).
- `ui.codemirror.binding(callback, ...)` factory for per-binding overrides: `prevent_default=False` (notify Python while letting the browser default fire — e.g. `Mod-c` notifies Python and the browser still copies) and per-platform `mac=` / `linux=` / `win=` shortcut overrides.

Bindings are registered at high precedence so they override `basicSetup` defaults (e.g. you can rebind `Mod-z`).

### Progress


evnchn (changes requested):

Really nice work, @Jepson2k — this API fits NiceGUI's idioms cleanly (`on_keybinding(...) -> Self` mirrors `on_select`, the event dataclass matches the existing pattern, and you didn't invent a parallel registry), the `Prec.high` placement is right, the fire-and-forget model is documented honestly, and there are no unsafe string→JS paths. I checked it out, ran the tests, and exercised custom bindings (multi-editor routing, runtime `remove_keybinding`, `prevent_default=False` on Mod-z all behaved exactly as documented). Two small things to fix before merge, then I think this is in good shape:

**🟠 One invalid key descriptor takes down the whole keymap.** A bad spec like `'Bogus-x'` makes CodeMirror throw `Unrecognized modifier name: Bogus` when it lazily compiles the combined keymap on the first keydown; because the binding table never gets built, every later keydown re-throws and the entire dispatch path goes dead — basicSetup undo, Tab indent, and all valid user bindings included. CM swallows the throw in its own keydown observer, so it surfaces only as a console error with no visible failure (I confirmed undo + Tab + a valid `Mod-d` binding all stop firing with one bogus binding present). Could we validate the modifier tokens Python-side and raise a friendly `ValueError` at registration, or warn-and-skip the bad entry in JS? Either is just a few lines and turns a silent footgun into a clear error.

**🟠 The tests go red on macOS.** `tests/test_codemirror.py` synthesizes `Mod-*` with a hard-coded `ctrlKey: true` (8 places), but CodeMirror maps `Mod`→Cmd on a Mac browser, so four of the keybinding tests fail locally for any Mac dev (including the Zauberzeug team) even though CI is green — I saw `test_keybinding_constructor`, `_method`, `_replaces_existing`, and `_prevent_default_false` fail, and all 26 pass once the modifier matches. One thing to watch: a blanket `ctrlKey`→`metaKey` swap would then break the Linux CI runner (where `Mod`→Ctrl), so the durable fix is platform-aware — pick the modifier from `/Mac/.test(navigator.platform)`, mirroring CM's own check.

**A couple of minor things:** the new keybindings docstring is missing the `*Since version X.Y.Z*` marker that the neighbouring `line_tooltips` paragraph has; and bindings don't fire while the editor is `.disable()`d (which is probably the behavior you want for a DisableableElement, just worth a docstring line or a quick test). And a tiny non-issue I was curious about — commit `8e35c541` mentions "per review feedback," but I don't see an earlier review on the PR; what was that referring to?

The tests themselves are genuinely good — Screen fixture throughout, no mocks, and the FIFO no-traffic control plus the pre-seeded-undo-history regression test are clever touches. No dist rebuild needed. Both fixes are small; ping me when they're in and I'll take another quick pass. 🙂

evnchn:

**TL;DR:**

All four items from my 06-14 review are addressed — and correct, not just present. Re-checked each against the diff at `eff9f7e7`.

Nothing needed on this one; a holistic re-review follows in a separate comment.

<details>
<summary><b>Per-item verification (diff at <code>eff9f7e7</code>)</b></summary>

| # | My 06-14 ask | Status | Evidence |
|---|---|---|---|
| 1 | Validate keybinding modifiers Python-side, raise a friendly error | ✅ | `_validate_keybinding()` raises a clear `ValueError` at registration; wired into `binding()`, `on_keybinding()`, and the constructor. Edge cases hold: `'Mod--'` (literal-dash key) → empty segment skipped; bare `'F5'` → no modifier checked; `'Mod-Shift-Boop-d'` → raises on `Boop`. The docstring spelling out the dead-keymap footgun is a nice touch. Test `test_keybinding_invalid_modifier_raises`. |
| 2 | Make tests platform-aware for macOS | ✅ | `_MOD_KEY_JS = '[/Mac/.test(navigator.platform) ? "metaKey" : "ctrlKey"]: true'` — resolves to `metaKey` on macOS / `ctrlKey` on Linux CI, mirroring CM6's own `Mod` resolution rather than a blanket swap. |
| 3 | Version markers | ✅ | All `3.14.0` now (commit `eff9f7e7` corrected an earlier 3.13.0 slip). |
| 4 | Disabled-binding behavior | ✅ | Documented + `test_keybinding_does_not_fire_while_disabled` (gates on the `contentEditable` flip, leans on WS FIFO ordering). |

Quick turnaround too — review 06-14 → fixes 06-16. Nicely done, @Jepson2k.
</details>

evnchn (dismissed):

**TL;DR:**

Approving. ✅

Beyond the four review items, I ran the whole feature end-to-end in a real browser on this branch — every path works: dict-key fidelity, `prevent_default=False`, runtime bind/unbind, and shift combos.

Implementation's clean and CI's green; nothing blocking.

<details>
<summary><b>End-to-end verification (ran the branch at <code>eff9f7e7</code> in a real browser)</b></summary>

Built a demo on the branch code (`PYTHONPATH` shadow over a venv), drove it with real keystrokes:

| Exercise | Result | Confirms |
|---|---|---|
| `Mod-s` | label → `saved:Mod-s` | fires, and `event.key` is the dict key (`'Mod-s'`), not `'s'` — key fidelity |
| `F5` | → `refresh` | plain function-key binding |
| `Mod-c` (`prevent_default=False`) | → `copy-logged` | notifies Python while the browser still copies |
| `Mod-Shift-d` via `on_keybinding` (real keypress) | → `runtime-bound` | runtime registration **and** shift-combo matching |
| `remove_keybinding('F5')` then F5 | label unchanged | unbind genuinely stops firing |

</details>

<details>
<summary><b>How I drove it (reproducible — replay against any checkout of this branch)</b></summary>

Checked out the PR head `eff9f7e7` in a worktree and ran a demo on the branch code by shadowing an installed venv with `PYTHONPATH` (confirmed `nicegui.__file__` pointed at the worktree, not site-packages):

```python
# kb_demo.py
from nicegui import ui

log = []
status = ui.label('idle').props('id=status')
def mark(m): log.append(m); status.set_text(m)

editor = ui.codemirror('def hello():\n    print("hi")', language='Python', keybindings={
    'Mod-s': lambda e: mark(f'saved:{e.key}'),
    'F5': lambda: mark('refresh'),
    'Mod-c': ui.codemirror.binding(lambda: mark('copy-logged'), prevent_default=False),
}).classes('h-32')
editor.on_keybinding('Mod-Shift-d', lambda: mark('runtime-bound'))
ui.button('Unbind F5', on_click=lambda: editor.remove_keybinding('F5')).props('id=unbind')
ui.run(port=8232, show=False, reload=False)
```

```bash
PYTHONPATH=<worktree-at-eff9f7e7> <venv>/bin/python kb_demo.py
```

Then drove the focused `.cm-content` two ways and read `#status` after each (handlers round-trip over WS, so ~450 ms settle between press and read):

- **Synthetic** — `content.dispatchEvent(new KeyboardEvent('keydown', {key, code, metaKey:true, …}))`. Fired `Mod-s` / `F5` / `Mod-c` correctly, but `Mod-Shift-d` did **not** register (tried both `key:'d'` and `key:'D'`).
- **Real CDP keypress** — focus via click, then `page.keyboard.press('Meta+Shift+D')` → `#status` became `runtime-bound`. That's the data point that flipped "looks broken" → "synthetic-event artifact, feature is fine".
- **Unbind** — clicked `#unbind`, re-fired F5 → `#status` unchanged, confirming `remove_keybinding`.
</details>

<details>
<summary><b>Went down the shift-combo rabbit hole so future tests don't have to — it's solid</b></summary>

One nuance I chased to be thorough: `Mod-Shift-d` fires under a **real** keypress but **not** under a synthetic `dispatchEvent`. CM6 resolves shift-combo key-names from genuine keyboard events in a way synthetic events don't reproduce — so this is a CM6 quirk, not anything in your code. I confirmed the real path works (CDP `Meta+Shift+D` → handler fired).

Net: your synthetic-event suite is exactly right for everything it covers, and it's genuinely thorough (10 tests, including the `remove_keybinding` un-firing assertion). It just can't reach shift-combo *firing* by construction — which is why `test_keybinding_constructor` rightly registers `Mod-Shift-d` and only asserts serialization. If you ever want belt-and-suspenders on shift matching, a single Selenium `ActionChains` press would do it — pure gravy.
</details>

<details>
<summary><b>Code-quality pass (no blockers)</b></summary>

- Frozen `slots` dataclass spec; `binding()` factory validates per-OS overrides up front.
- Constructor batches via `suspend_updates()` — no per-key update storm.
- `_dispatch_keybinding` no-ops on an unknown/just-removed key — handles the emit-after-remove race.
- JS uses `CM.Prec.high` so user bindings win over `basicSetup` (e.g. `Mod-z` undo), reconfigured through a dedicated `Compartment` for runtime updates.
- No new client→server trust surface beyond standard NiceGUI event emits.
- CI green (mypy / pre-commit / pylint / pytest), `MERGEABLE`.
</details>

<details>
<summary><b>Unicode / UTF-16 — verified robust</b></summary>

Probed the keybinding Unicode surface (key strings + the `event.key` round-trip) on the branch:

| Test | Result |
|---|---|
| validation accepts non-ASCII **key names** (`Mod-ä`, `中`, `😀`) | ✅ |
| validation rejects non-ASCII **modifiers** (`Ä-s` → `ValueError`) | ✅ |
| prop serialization | `["ä","ñ","中","Mod-é","Mod-€"]` — no mojibake |
| press non-ASCII key → `e.key` to Python | exact string round-trips (`hit:中`, `hit:Mod-€`) |

Round-trip fidelity is guaranteed by construction — the JS `run` echoes the registered key string rather than re-deriving it from the event, so the string can't drift.

**One FYI for the author (not blocking):** matching is exact-string on `event.key`, so a binding registered with a **precomposed** character (`'Mod-é'`, U+00E9) won't match a **decomposed** keypress (`e` + U+0301), and vice-versa. That's how CM6 — and keymaps generally — work, not a bug here; but an `unicodedata.normalize`-flavored caveat could be worth a docstring line if non-ASCII bindings are expected.
</details>

Heads-up: sibling #5988 touches the same `codemirror.py`, so whichever lands first will want a quick rebase on the other.

LGTM — approving. Thanks @Jepson2k.

---
*Changelog*
- 2026-06-18: Added a Unicode/UTF-16 verification block (key-string validation, prop serialization, `event.key` round-trip) + an NFC/NFD normalization FYI for the author. Verdict unchanged (approved).

falkoschindler (commented):
Thanks a lot for the pull request, @Jepson2k! Custom keybindings a are great addition to `ui.coremirror` and I appreciate the work you put into this.

I went over the code in some detail and pushed a series of follow-up commits. Rather than walk through each one (the individual commit messages cover the specifics), here's the overall shape of what changed:

**API surface aligned with CodeMirror's own vocabulary.** The public API now mirrors CodeMirror 6's naming: a `keymap` parameter/prop, a `KeyBinding` value class that carries the key together with its options (`prevent_default`, per-OS `mac`/`linux`/`win`), and `map_key`/`unmap_key` methods for runtime changes. The event stays `CodeMirrorKeyBindingEventArguments`.

**Validation moved from Python to the client.** The original Python-side modifier/chord validation duplicated rules that CodeMirror already enforces (and got some of them wrong — it over-rejected valid multi-stroke chords). Instead of re-implementing CodeMirror's grammar, we now force CodeMirror to build the combined keymap at registration time, so a bad modifier or a chord prefix that collides with a standalone binding surfaces immediately and is reported through the server log — rather than silently killing every keybinding on the first keystroke. The `logAndEmit` socket emit is now deferred until the websocket handshake so early errors aren't lost.

**Structure.** `codemirror.py` was growing unboundedly, so the language/theme literals moved into `constants.py` and all keybinding logic into a `KeyBindingElement` mixin (`keybindings.py`) that `CodeMirror` inherits.

**Tests rewritten around observable behavior.** The keybinding tests now live in their own file and assert what a user can observe — notifications, the editor's rendered content, the server log — instead of internal props or `defaultPrevented`. They cover the map → re-map → unmap lifecycle, `prevent_default`, the disabled case, and invalid-binding reporting.

**Demo reworked** to show the new API.

I'll do a final agentic review pass tomorrow. Then the PR should be ready to merge. Thanks again for the contribution!

falkoschindler (approved):
Did the final review pass today and went over the whole diff once more — nothing left to flag. The `KeyBindingElement` split, the `Prec.high` placement, and forcing CodeMirror to build the combined keymap so conflicts surface at registration all hold up; disabled editors correctly drop bindings; and the tests/docs are thorough. Fixed one tiny docstring slip along the way (it named the `KeyBinding` class instead of the `keybinding` event).

LGTM — approving for merge. Thanks again, @Jepson2k!

PR #6002, opened 2026-04-27
fix: preserve meaningful DataFrame index in ui.aggrid.from_pandas and ui.table.from_pandas

### LINKED ISSUE #5995: aggrid.from_pandas ignores index
### First Check

### Example Code

```python
import pandas as pd
from nicegui import ui

df = pd.DataFrame(
    data={"col1": [1, 2], "col2": [3, 4]},
    index=["row1", "row2"]
)

ui.aggrid.from_pandas(df)

ui.run()
```

### Description

The provided index is ignored.
I would expect it to be shown as row titles, visually similar to the column header.

### NiceGUI Version

3.10.0 (from conda-forge)

### Python Version

3.13.13

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

DarkRiddle1212:
### Motivation

`ui.aggrid.from_pandas(df)` and `ui.table.from_pandas(df)` silently dropped the DataFrame index, causing data loss whenever the index carried information (named index, custom RangeIndex, MultiIndex, etc.). Closes #5995.

The design — auto-include rather than opt-in via a flag — was settled in [#5995 (comment)](https://github.com/zauberzeug/nicegui/issues/5995#issuecomment-4319377778): the previous behavior was effectively silent data loss, and users who deliberately want to hide the index can do so with `df.reset_index(drop=True)` before passing the DataFrame.

### Implementation

- New helper `nicegui.helpers.is_default_range_index(df)` returns `True` only for the canonical unnamed `RangeIndex(0, n, 1)`. Any other index — named, non-default range, MultiIndex, etc. — is considered informative.
- `Aggrid.from_pandas` and `Table._pandas_df_to_rows_and_columns` call `df.reset_index()` whenever the helper returns `False`, so the index becomes regular column(s).
- `from_polars` is unaffected: Polars DataFrames don't have an index, so there is no parallel change needed.
- Tests are parametrized over named, custom-range, string, MultiIndex, and default-RangeIndex cases for both `ui.aggrid` and `ui.table`.

### Progress

- [ ] Documentation has been added/updated.
- [ ] No breaking changes to the public API.


falkoschindler:
Thanks a lot for the pull request, @DarkRiddle1212!
The new pytests for `ui.aggrid` are failing though. Can you, please, have a look? Thanks!

DarkRiddle1212:
Hi @falkoschindler, I've pushed a fix for the failing ui.aggrid pytest. Please have another look when you get a chance

falkoschindler (commented):
I pushed a few follow-ups on top of your commits to tighten the implementation, deduplicate the detection between `ui.aggrid` and `ui.table`, and document the behavior change:

- Detector simplified — unreachable `MultiIndex` clause and redundant inner `isinstance` guard dropped (897aec85).
- WHAT-not-WHY comments removed alongside the simplification.
- Duplicated detection logic extracted into `nicegui.helpers.is_default_range_index` (08362a95).
- Five separate index tests collapsed into one parametrized test per element (d3e206d2).
- PR description rewritten to follow the repo template and to mention that Polars has no index, so `from_polars` is unaffected.
- Docstring updates to document the auto-include behavior, the `df.reset_index(drop=True)` escape hatch, and a `*Since version 3.12.0*` marker.

### One open question

Looking at this side-by-side:

```python
ui.aggrid.from_pandas(pd.DataFrame({'value': [1, 2]}, index=pd.RangeIndex(0, 2)))  # hides index
ui.aggrid.from_pandas(pd.DataFrame({'value': [1, 2]}, index=pd.RangeIndex(1, 3)))  # shows index
```

Two near-identical DataFrames produce different schemas because of the `start`/`step` checks in `is_default_range_index`. Would you be open to narrowing the heuristic to:

> Hide the index iff it's an _unnamed_ `RangeIndex`. Show it for named indexes and for anything that isn't a `RangeIndex`.

That drops the `start == 0` and `step == 1` clauses from the helper, which collapses the whole check to `isinstance(df.index, pd.RangeIndex) and df.index.name is None` — short enough that we could inline it at both call sites and drop `helpers.data` entirely. One parametrized test case flips (`pd.RangeIndex(start=100, stop=102)` moves from `expected` to `unexpected`). Pros: a one-sentence rule, no surprising visual difference between the two examples above. Con: someone who deliberately constructs `pd.RangeIndex(start=5, stop=7)` and wants those numbers visible has to either name the index or call `df.reset_index()` themselves — both trivial.

Happy to push it as a follow-up commit, or you can take it. What do you think? Also pinging @NichtJens / @evnchn from the original issue thread in case you have an opinion.

evnchn:
@falkoschindler As a plain `RangeIndex` with specified start and stop, such as `pd.RangeIndex(start=5, stop=7)`, has uses beyond display (consider `df.loc[5]` then returns the first row, same as `df.iloc[0]`), the fact that a `RangeIndex` is there does NOT mean the user is using the index visually. I would argue that the named indexes opt-in is fair.

NichtJens:
Regarding the open question:

On the one hand, "RangeIndex needs to be named to show" is quite simple to explain. Which IS a great point!

On the other hand, would anyone ever do this?
```python
ui.aggrid.from_pandas(pd.DataFrame({'value': [1, 2]}, index=pd.RangeIndex(0, 2)))
```

So, in fact, the comparison would be between:
```python
ui.aggrid.from_pandas(pd.DataFrame({'value': [1, 2]}))  # hides index
ui.aggrid.from_pandas(pd.DataFrame({'value': [1, 2]}, index=pd.RangeIndex(1, 3)))  # shows index
```

Which are definitely not near-identical :)

Since I am clearly also undecided, maybe the solution is to implement the simpler rule, but mention:

> If you want an index like `pd.RangeIndex(start=5, stop=7)` to be visible, you need to name it.

as well as `df.reset_index()` specifically in the docs?

evnchn:
@NichtJens While `ui.aggrid.from_pandas(pd.DataFrame({'value': [1, 2]}, index=pd.RangeIndex(0, 2)))` is an interesting argument in vaccum, does not hold when considering one cannot exclude the fact that the user code may be:

```py
index_start = 0 # user-provided, can be any number
ui.aggrid.from_pandas(pd.DataFrame({'value': [1, 2]}, index=pd.RangeIndex(index_start, index_start+2)))
```

It would be very awkward if, both unnamed, `index_start=0` (index suddenly hidden) and say `index_start=5` (index suddenly shown) behaves differently, which is what happens when you try to detect in-band by catching the index start and step numbers in your suggestion.

Since we cannot show index for unnamed `index_start=0` otherwise all dataframes suddenly get one more index, and frankly sorry I cannot concede and I insist that for a good API design the opt-in mechanism should be out-of-band, that I believe **all un-named index gets hidden, no matter the `index_start`** and **name your index for it to show up** is the desired outcome.

NichtJens:
@evnchn Not entirely sure I follow. I think you are arguing against something I didn't say. I think, we all agree on the desired solution, actually.

However, I would argue that in your new example, the indices should always be shown since the user has control over their values and thus should also see them in case they are starting from 0. Or, in other words, why would you give the user a way to set something if that something has no impact on the result?

But again, I am fully for "name your index for it to show up" as the rule.

falkoschindler:
@NichtJens One thing worth flagging before we lock this in: a strict "named only" rule would not fix your original example.

```python
df = pd.DataFrame(data={"col1": [1, 2], "col2": [3, 4]}, index=["row1", "row2"])
```

That index is an unnamed `Index`, so under "name your index for it to show up" the row labels `row1` / `row2` would still be dropped — same bug you reported.

The proposal in [my earlier review](https://github.com/zauberzeug/nicegui/pull/6002#pullrequestreview-4182336400) — hide iff the index is an **unnamed `RangeIndex`** — keeps your example working (string `Index` → shown) and also addresses @evnchn's parameterized-start concern (any unnamed `RangeIndex` is hidden regardless of bounds, no in-band flip). I think that's still the right rule.

NichtJens:
@falkoschindler Ah, sorry! That got lost on the way. I had it as "RangeIndex needs to be named to show" (as the non-RangeIndex are not controversial anyway). And then I copied @evnchn's version "name your index for it to show up" for my last reply.

Bottom line: I think your proposal is the best-of solution with the least surprises.

falkoschindler (approved):
Great! Looks like we agree on the current implementation.
Let's merge! 🚀

DarkRiddle1212:
Really appreciate the thoughtful reviews and refinements from everyone. Thanks for helping shape this into a cleaner solution. Happy to contribute and learn through the process
Also, apologies for going quiet during the discussion, I got pulled into work and couldn't keep up with the conversation as closely as I wanted. I appreciate everyone carrying it forward.

PR #6004, opened 2026-04-27
Clarify client_id security model and example deployment caveats

evnchn:
### Motivation

A private security report recently arrived that misunderstood NiceGUI's `client_id` model — treating the `/_nicegui/` upload route as an "unauthenticated endpoint" rather than a session-token-authenticated one, and treating `examples/xterm` and `examples/authentication` as production-ready templates. The framework already behaves correctly under its documented security model; the gap is that the model is under-explained.

This PR tightens the docs to head off that class of report. It does not change any framework behaviour and does not address an actual vulnerability.

### Implementation

Two doc-only changes on the website:

1. **`website/documentation/content/section_security.py` — Client-Side Secrets section rewritten:**
   - Splits the explanation into **issuer** (pages mint and embed `client_id` in HTML) and **consumer** (every `/_nicegui/` route trusts `client_id` as the session token).
   - States explicitly that there is no separate auth layer on `/_nicegui/` — protecting the page that mints the `client_id` is what protects them.
   - Adds a "Secure pages, not `/_nicegui/`" rule so the actionable guidance is unambiguous.

2. **`website/documentation/content/section_security.py` — new "Examples Are Starting Points" section:**
   - Frames examples as demonstrations, not production templates.
   - Names specific footguns: `xterm` is a Bash PTY by design, the `authentication` example needs production hardening, `ui.upload` validation is browser-only, `reload=True` is dev-only.

3. **`website/examples_page.py` — small but obvious banner** below the `/examples` heading linking to the new section, so a developer landing on the examples page sees the warning before copying code.

### Progress


falkoschindler (approved):
Great! I did some minor changes (see my previous commits), now it's ready to merge from my point of view. 👍🏻

evnchn:
@falkoschindler sure. I looked at your commits and I think they're fine

PR #6006, opened 2026-04-28
Fix duplicate `app.timer` and lifecycle handler registration in script mode

### LINKED ISSUE #6003: app.timer acting weird when in script mode
### First Check

#### good way

```python
from nicegui import ui, app
from datetime import datetime

def global_timer():
    print("timer executed", datetime.now())

app.timer(1, global_timer)

@ui.page("/")
def index():
    ui.label("example label text")

ui.run()
```

#### bad way

```python
from nicegui import ui, app
from datetime import datetime

def global_timer():
    print("timer executed", datetime.now())

app.timer(1, global_timer)

ui.label("example label text")

ui.run()
```

### Description

When using `app.timer()` in [script mode](https://nicegui.io/documentation/section_pages_routing#script_mode) (without `ui.page()` ) it behaves weird.
The timer runs normally but when a client connects another timer will start for that session.
This results in one additional execution per second for each connected client.

# Possible solutions
- Improve the documentation for [app.timer](https://github.com/zauberzeug/nicegui/blob/31c4a69b81d0a65992c8cb145556ed1a86f1eafe/website/documentation/content/timer_documentation.py#L84) to warn about this behaviour.
- Detect when `app.timer(...)` is registered in script mode and print warning message
- Detect when `app.timer(...)` is registered in script mode and raise an exception

### NiceGUI Version

3.11.1

### Python Version

3.14.3

### Browser

[Brave](https://brave.com) (based on Chrome)

### Operating System

Windows

### Additional Context

# output of good code
Image

# output of bad code

Image

falkoschindler:
### Motivation

Closes #6003.

In script mode, `ui.run()` installs a root function that re-executes the user's script via `runpy.run_path` for every client connection. Anything in the global scope is therefore re-run per client — including `app.timer(...)`, which silently stacked another global timer on every connection. The reporter saw one extra tick per second per connected client.

The same shape of bug affected `app.on_connect`, `app.on_disconnect`, `app.on_delete`, `app.on_shutdown`, and `app.on_exception`: each would append another handler to the list per client connection, with no error or warning. (`app.on_startup` already guards itself.)

### Implementation

Added `core.is_script_mode_re_execution()` as the symmetric counterpart to the existing `core.is_script_mode_preflight()`:

- `is_script_mode_preflight()` → `script_mode and not app.is_started` (the first execution, before the server starts)
- `is_script_mode_re_execution()` → `script_mode and app.is_started` (the per-client re-execution after the server started)

The two timer flavors have opposite scope, so they skip on opposite phases:

- `app.timer` (global) skips on the per-client re-execution; it was already registered on the first run.
- `ui.timer` (per-client) skips on the preflight; it registers on each re-execution, as before.

To express that without a flag or XOR, `Timer.__init__` now calls `self._skip_registration()`, which the base class implements with `is_script_mode_re_execution()` and the Element subclass overrides with `is_script_mode_preflight()` (the existing pre-fix behavior, just relocated). This matches the override pattern already used for `_get_context`, `_can_start`, `_should_stop`, etc.

The five `App.on_*` lifecycle handlers got the same one-liner guard via the new helper.

### Progress

  - [x] How to unify the behavior of `app.on_startup` and `app.timer`?
- [x] Pytests are not necessary (and not easily possible with current infrastructure).

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -104,6 +104,8 @@ class App(FastAPI):
 
         The callback can be synchronous or asynchronous and has an optional parameter of `nicegui.Client`.
         """
+        if core.is_script_mode_re_execution():
+            return  # already registered on the script's first execution
         self._connect_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_connect()'))
 
     def on_disconnect(self, handler: Callable) -> None:
@@ -113,6 +115,8 @@ class App(FastAPI):
 
         *Updated in version 3.0.0: The handler is also called when a client reconnects.*
         """
+        if core.is_script_mode_re_execution():
+            return  # already registered on the script's first execution
         self._disconnect_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_disconnect()'))
 
     def on_delete(self, handler: Callable) -> None:
@@ -122,6 +126,8 @@ class App(FastAPI):
 
         *Added in version 3.0.0*
         """
+        if core.is_script_mode_re_execution():
+            return  # already registered on the script's first execution
         self._delete_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_delete()'))
 
     def on_startup(self, handler: Callable) -> None:
@@ -143,6 +149,8 @@ class App(FastAPI):
         The callback can be synchronous or asynchronous.
         When NiceGUI is shut down or restarted, all tasks still in execution will be automatically canceled.
         """
+        if core.is_script_mode_re_execution():
+            return  # already registered on the script's first execution
         self._shutdown_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_shutdown()'))
 
     def on_exception(self, handler: Callable) -> None:
@@ -150,6 +158,8 @@ class App(FastAPI):
 
         The callback has an optional parameter of `Exception`.
         """
+        if core.is_script_mode_re_execution():
+            return  # already registered on the script's first execution
         self._exception_handlers.append(handler)
 
     def handle_exception(self, exception: Exception) -> None:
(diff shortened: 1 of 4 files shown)


falkoschindler:
### Open question

`app.on_startup` already guards against re-registration during the per-client re-execution, but it does so by **raising** with a hint to use `@ui.page`:

```python
if self.is_started and core.script_mode:
    raise RuntimeError('Unable to register a startup in script mode. Use a `@ui.page` function instead.')
```

This PR silently skips re-registration for `app.timer` and the other lifecycle handlers, so `on_startup` is now the odd one out. Worth picking one canonical behavior:

- **Silent skip (this PR).** The user gets the result they expected — one global registration — and the per-client re-execution becomes invisible. Trade-off: a user who doesn't yet understand script mode never finds out they were doing something subtly off-script.
- **Raise with `@ui.page` hint (today's `on_startup`).** Surfaces the conceptual mismatch loudly and steers the user toward `@ui.page`. Trade-off: the script is re-executed on _every_ client connection, so the error fires every time someone hits the page — quite noisy, and arguably overkill when the user's intent was unambiguous.

Worth noting that `on_startup` raising is partly a historical artifact: registering a startup handler after startup is a real error regardless of script mode (the startup phase is already over), so it has always raised. Script mode just made that path easier to hit. For `app.timer` and the `on_connect/disconnect/...` handlers, registering after startup is _normally_ fine (dynamic timers, late-bound handlers), so the raise pattern doesn't carry over as cleanly.

Happy to align in either direction in this PR or a follow-up.

EchterTimo:
+1 for Silent Skip

evnchn (approved):
Code is workable. 

If `is_script_mode_re_execution` is private API merge away. 

If it's public API I recommend a demo.

PR #6008, opened 2026-04-29
Fix Nested Sub Pages doc demo link target and label

falkoschindler:
### Motivation

In the Nested Sub Pages documentation demo, the link inside `other()` labeled "Go to main" was ambiguous — the outer `main` and the nested `sub_main` were both essentially called "main". Worse, the commented real-code line `ui.link('Go to main', '/')` would actually navigate to the _outer_ main, while the rendered FakeSubPages demo (which resolves `/` against the nested routes) showed `sub_main`. So the comment and the demo disagreed on what the link does.

Reported by @aisartag in #5999.

### Implementation

- Rename the link label from "Go to main" to "Go to sub main" so it's unambiguous which page is meant.
- Change the commented real-code URL from `/` to `/other` so a reader copying the snippet actually lands on `sub_main` (outer matches `/other` → `other()` → nested `/` → `sub_main`), matching what the rendered demo shows.

### Progress

### DIFF
--- a/website/documentation/content/sub_pages_documentation.py
+++ b/website/documentation/content/sub_pages_documentation.py
@@ -328,7 +328,7 @@ def nested_sub_pages_demo():
 
     def other():
         ui.label('sub page')
-        # ui.link('Go to main', '/')
+        # ui.link('Go to sub main', '/other')
         # ui.link('Go to A', '/other/a')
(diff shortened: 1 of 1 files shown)


evnchn (approved):
makes sense

PR #6012, opened 2026-04-29
Dedent `ui.mermaid` content like `ui.markdown` does

falkoschindler:
### Motivation

`ui.markdown` already auto-dedents its content via `remove_indentation`, but `ui.mermaid` only stripped outer whitespace.
Identical triple-quoted strings therefore behaved differently between the two elements.

The discrepancy becomes a hard error with Mermaid's YAML front-matter (e.g. `displayMode: compact` for compact-mode Gantt charts), which requires `---` markers at column 0.
Any leading indentation makes Mermaid bail out with _"Diagrams beginning with --- are not valid. If you were trying to use a YAML front-matter, please use un-indented `---` blocks"_.

Refs discussion #6011.

### Implementation

- Promote `remove_indentation` from `nicegui/elements/markdown.py` to `nicegui/helpers/strings.py` and re-export it via `nicegui.helpers`.
  It is no longer markdown-specific now that `ui.mermaid` uses it as well, and lifting it removes the cross-element dependency from `mermaid` → `markdown`.
- Update all existing call sites (`code`, `markdown`, `restructured_text`, `tests/test_vbuild.py`, three website files) to import from the new location.
- Use `remove_indentation` inside `Mermaid._handle_content_change`, replacing the previous `content.strip()` calls.
  Since `remove_indentation` already drops leading whitespace-only lines and produces no leading whitespace on the first line, no extra `.strip()` is needed; Mermaid is fine with any trailing whitespace.
- Add a regression test `test_mermaid_with_yaml_frontmatter` that places a YAML front-matter block inside an indented triple-quoted string and verifies the chart renders without a syntax error.

This is not a breaking change in practice: Mermaid is whitespace-insensitive at the start of lines for every diagram type, so removing a uniform leading indent is a no-op for existing diagrams.
The four demos in `mermaid_documentation.py` continue to render identically.

### Progress

- [x] Pytest has been added.

### DIFF
--- a/nicegui/elements/code.py
+++ b/nicegui/elements/code.py
@@ -1,9 +1,9 @@
 import asyncio
 import time
 
+from ..helpers import remove_indentation
 from .button import Button as button
 from .markdown import Markdown as markdown
-from .markdown import remove_indentation
 from .mixins.content_element import ContentElement
 from .timer import Timer as timer
 
--- a/nicegui/elements/markdown.py
+++ b/nicegui/elements/markdown.py
@@ -8,6 +8,7 @@ from fastapi.responses import PlainTextResponse
 from pygments.formatters import HtmlFormatter  # pylint: disable=no-name-in-module
 
 from .. import core
+from ..helpers import remove_indentation
 from .mixins.content_element import ContentElement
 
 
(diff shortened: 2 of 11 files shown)


evnchn (approved):
Refactor looks sound (trusting your AI agent). On my end I watch for the empirical results:

Image

Please merge only after "Claude, is it safe to dedent all Mermaid diagrams automatically" yields a Yes.

falkoschindler:
@evnchn Yes, Claude couldn't come up with a counter example where existing Mermaid code would break after this change. 👍🏻

PR #6013, opened 2026-04-29
fix: resolve DEFAULT_LOCAL_SCOPE class variable at runtime in Element…

### LINKED ISSUE #6005: ElementFilter.DEFAULT_LOCAL_SCOPE has no effect on toggling default local scope.
### First Check

### Example Code

```python
from nicegui import ui, ElementFilter

ui.label('outside').mark('label')
with ui.row() as row:
    ui.label('inside row').mark('label')

ElementFilter.DEFAULT_LOCAL_SCOPE = True

with row:
    matches = list(ElementFilter(marker='label'))
    l = len(matches)
    print(len(matches)) # should be 1 since default local scope is True. but have 2

with row:
    matches = list(ElementFilter(marker='label', local_scope=True))
    l = len(matches)
    print(len(matches)) # local scope is set to True and got 1. this works.
```

### Description

Setting `ElementFilter.DEFAULT_LOCAL_SCOPE = True` is expected to change the default behavior of `ElementFilter(...)`, but it currently does not.

The reason is that` local_scope` in `ElementFilter.__init__ is` defined as `local_scope: bool = DEFAULT_LOCAL_SCOPE`. In Python, default argument values are evaluated once when the function is defined, so local_scope stays bound to the original value (False) even after `ElementFilter.DEFAULT_LOCAL_SCOPE` is changed later.

As a result, this documented toggle does not work at runtime.

A possible fix would be to resolve the default inside __init__ instead of in the function signature, for example:

```py
def __init__(..., local_scope: bool | None = None):
    if local_scope is None:
        local_scope = self.DEFAULT_LOCAL_SCOPE
```
This would make `ElementFilter.DEFAULT_LOCAL_SCOPE` behave as an actual runtime default.

### NiceGUI Version

2.22.0

### Python Version

3.12.13

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

DarkRiddle1212:
### Motivation

Fixes #6005

The bug: `ElementFilter.DEFAULT_LOCAL_SCOPE = True` has no effect because Python evaluates default argument values once when the function is defined, so `local_scope` stays bound to the original `False` value even after `DEFAULT_LOCAL_SCOPE` is changed.

This prevents users from changing the default behavior of `ElementFilter` at runtime, which is the documented and intended functionality.

### Implementation

Changed the local_scope parameter from `local_scope: bool = DEFAULT_LOCAL_SCOPE` to `local_scope: bool | None = None` in all method signatures (both `@overload` decorators and the actual `__init__` method).

Added runtime evaluation at the start of `__init__`:

```py
if local_scope is None:
    local_scope = self.DEFAULT_LOCAL_SCOPE
```

This ensures that `DEFAULT_LOCAL_SCOPE` is evaluated every time `ElementFilter` is instantiated, not just once at function definition time. This is the standard Python pattern for handling runtime-evaluated defaults.

The fix maintains full backward compatibility:

- Explicit `local_scope=True` or `local_scope=False` still works as before
- Default behavior (when no parameter is passed) now correctly respects the current value of `DEFAULT_LOCAL_SCOPE`

### Progress

- [x] Documentation is not necessary (existing docstring already describes this behavior).

### DIFF
--- a/nicegui/element_filter.py
+++ b/nicegui/element_filter.py
@@ -27,7 +27,7 @@ class ElementFilter(Generic[T]):
     def __init__(self: ElementFilter[Element], *,
                  marker: str | list[str] | None = None,
                  content: str | list[str] | None = None,
-                 local_scope: bool = DEFAULT_LOCAL_SCOPE,
+                 local_scope: bool | None = None,
                  only_visible: bool = False,
                  ) -> None:
         ...
@@ -37,7 +37,7 @@ class ElementFilter(Generic[T]):
                  kind: type[T],
                  marker: str | list[str] | None = None,
                  content: str | list[str] | None = None,
-                 local_scope: bool = DEFAULT_LOCAL_SCOPE,
+                 local_scope: bool | None = None,
                  only_visible: bool = False,
                  ) -> None:
         ...
@@ -46,7 +46,7 @@ class ElementFilter(Generic[T]):
                  kind: type[T] | None = None,
                  marker: str | list[str] | None = None,
                  content: str | list[str] | None = None,
-                 local_scope: bool = DEFAULT_LOCAL_SCOPE,
+                 local_scope: bool | None = None,
                  only_visible: bool = False,
                  ) -> None:
         """ElementFilter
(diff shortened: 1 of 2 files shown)


falkoschindler:
Thanks for the pull request, @DarkRiddle1212!
Can you, please, check why the pre-commit hook and pylint fail? And it would help if you could preserve the formatting of the PR template for the PR's description (headings, checklist, code formatting). Thanks!

DarkRiddle1212:
@falkoschindler Fixed! Pre-commit hooks should pass now and PR description is formatted. Please check when you get a chance.

falkoschindler (approved):
Thanks again, @DarkRiddle1212!

I took the liberty of tightening the tests a bit on top of yours: collapsed the three cases into one parametrized test using `monkeypatch.setattr` instead of manual try/finally, and shortened the labels for readability. Behavior coverage is unchanged.

Let's merge! 🚀

PR #6015, opened 2026-04-29
Add scoped User lookups and pytest example

ZhuGC-Centogene:
### Motivation

This PR improves `nicegui.testing.User` for pages that intentionally reuse markers or content in different parts of the layout.

Before this change, `User.find`, `User.should_see`, and `User.should_not_see` always searched from the bound client root.
That made it hard to write precise tests for repeated UI structures, even when the test had already entered a local element context.
This PR makes scoped user testing work as expected and adds a runnable example in `examples/pytests` to show the intended usage pattern.

### Implementation

The `User` class now resolves a `scope` from the current `context.slot_stack`.
If the test is currently inside an element context, `User` uses that element as the search scope.
If there is no active slot, or the active parent belongs to a different client, it safely falls back to the bound client.

The element lookup paths used by `find`, `should_see`, and `should_not_see` were updated to use that scope.
Their underlying `ElementFilter` queries now use `local_scope=True` so searches stay within the current container instead of leaking into sibling branches of the page.

A regression test was added in `tests/test_user_simulation.py` to verify scoped lookup behavior with duplicate markers and duplicated content across sibling cards.

To make the feature discoverable for users, `examples/pytests` now includes:
- a small `/scoped_user` demo page with duplicate markers in separate cards
- a matching `test_scoped_search_for_elements` example that shows how to use `with user:`, `ElementFilter(..., local_scope=True)`, and `with left_card:`
- a short README note pointing readers to the example

### Progress


falkoschindler:
Thanks for the contribution, @ZhuGC-Centogene! Scoping `User` lookups to the surrounding `with element:` context might be a useful improvement, and the bundled example makes the feature easy to discover. A few issues to address before this can land, though: the change isn't backward-compatible and breaks two existing tests, plus a handful of style and API-surface points below.

- [ ] **1. Breaking behavior change.** Running `pytest tests/test_user_simulation.py` on this branch already fails 2 existing tests (`test_drawer`, `test_should_not_see_notification`) — `should_see` no longer finds elements in `ui.header()`, `ui.left_drawer()`, `ui.footer()`, or top-level `Notification`s, because the unconditional `local_scope=True` confines the search to `client.content` instead of `client.layout`. Users relying on the old behavior will be surprised in the same way. Worth questioning whether this should wait for 4.0 rather than ship as a minor-version change. And we probably need to provide an opt-out.

- [x] **2. Banner-style comments inside `scope`** ([nicegui/testing/user.py:50-53](nicegui/testing/user.py#L50-L53)) violate the project's simplicity rule and are visually noisy. Replace with a single line, and tighten the wording — the current text doesn't say _why_ a mismatch can happen or _what_ the fallback protects against:

   ```diff
   -        #################################################
   -        # client mismatch can happen during User testing#
   -        # so always use supplied client if mismatches   #
   -        #################################################
   -        if current_parent.client is not self._client:
   +        if current_parent.client is not self._client:
   +            # fall back to the bound client if the slot stack belongs to a different page
               return self._client
   ```

- [x] **3. Missing blank lines between methods** ([nicegui/testing/user.py:43-44 and 56-57](nicegui/testing/user.py#L43-L57)). PEP 8 requires one blank line between methods inside a class; the project's pre-commit formatter will flag this.

- [x] **4. `scope` is exposed as public API without a docstring.** The name has no underscore, so it's part of `User`'s surface. Either rename to `_scope` (it's an implementation detail used only by the three methods that already wrap with `self.scope`) — or keep it public and add a docstring describing the resolution rules and the fallback. CONTRIBUTING.md requires docstrings on the public API.

- [x] **5. Verbose assertion message** in [tests/test_user_simulation.py:933-936](tests/test_user_simulation.py#L933-L936):

   ```python
   expected_count = 1
   actual_count = len(user.find(marker='duplicated-marker').elements)
   assert actual_count == expected_count, f'Expected to find {expected_count} duplicated marker in left card, but found {actual_count}.'
   ```

   The example test already uses the simpler `assert len(user.find(marker='duplicated-marker').elements) == 1`. Pytest prints both sides on failure, so the intermediate variables and message add no signal. Suggest matching the example.

- [x] **6. Use `ui.element` instead of importing `Element`.** The new `from nicegui.element import Element` is unnecessary — `ui` is already imported and the rest of this file consistently uses `ui.element` (see `T = TypeVar('T', bound=ui.element)` at the top). Drop the import and annotate `scope` as `ui.element | Client`.

ZhuGC-Centogene:
The fundamental cause of the `User.should_see` issue is that `ElementFilter` scope handling was inconsistent. When `local_scope=True` but there is no active surrounding element context, semantically this should still behave like a global search. In that case, `ElementFilter` was using `client.content` rather than `client.layout`, which is too narrow.

So the fix is to narrow the search only when both conditions are true: there is an active element context, and `local_scope=True`. Otherwise, the search should fall back to `client.layout`. This also matches the expected semantics when `ElementFilter` is used on its own.

I also addressed the other review feedback in the latest update.

falkoschindler (changes requested):
Thanks for the iteration — the previous round of fixes (private `_scope`, `ui.element` annotation, simplified assert, layout-shell coverage) all look good. Three things still to address before merge, plus one open design question.

### Required

1. **Revert the `ElementFilter` change.** [nicegui/element_filter.py:100-102](nicegui/element_filter.py#L100-L102) should go back to:

   ```python
   self._scope = context.slot.parent if local_scope else context.client.layout
   ```

   The page-level fallback is a public-API behavior change for everyone calling `ElementFilter(local_scope=True)`, not just `User`. Whether `ElementFilter` should fall back to `client.layout` when the active slot is `client.content` is a worthwhile discussion, but it's a 4.0-scope change and shouldn't ride along with this PR.

2. **Move the carve-out into `User._gather_elements`.** This is the necessary companion to the revert — without it, `test_drawer` and `test_should_not_see_notification` regress again. Derive `local_scope` from the resolved `_scope`:

   ```python
   def _gather_elements(self, ...):
       local = isinstance(self._scope, ui.element)  # only narrow when inside a sub-element
       elements = set(ElementFilter(..., local_scope=local))
       ...
   ```

   Combined with `with self._scope:` already wrapping the call sites:
   - `_scope` is `Client` → enters `content`, `local_scope=False` → searches `client.layout` (full page, like today).
   - `_scope` is an `Element` → enters that element's slot, `local_scope=True` → searches that subtree.

3. **Drop `tests/test_element_filter.py::test_always_get_q_layout_without_element_context`.** It validates the reverted carve-out and would fail otherwise. The other new test (`test_local_scope_false_searches_entire_page_inside_element_context`) covers pre-existing `local_scope=False` behavior and can stay.

### Open design question

4. **`User.find` / `should_see` / `should_not_see` is still a silent breaking change.** Even after the cleanup above, the core PR intent — making these methods honor a surrounding `with element:` block — changes behavior for any existing test that happened to be inside such a block. They will fail by _not_ finding things, which is the worst kind of silent regression. Options:
   - Ship as-is (accept the breakage as deliberate; document in release notes).
   - Add an opt-out, e.g. a `scoped: bool = False` parameter, or a `User.DEFAULT_LOCAL_SCOPE` class flag mirroring `ElementFilter.DEFAULT_LOCAL_SCOPE`.
   - Defer to 4.0.

   I'd lean toward an opt-in flag for now (or 4.0), since users hit this without warning. Curious what you think.

ZhuGC-Centogene:
Thanks for the detailed review and for walking me through the backward-compatibility concerns.

- For the open design question, I support making this opt-in for now by adding a `scoped: bool = False` parameter to `User.find`, `User.should_see`, and `User.should_not_see`. I may not represent the majority of NiceGUI users, but from my perspective fast UI testing with `nicegui.testing.User` is especially valuable because it is relevant for regulatory and compliance-related work on my side. `nicegui.testing.Screen` is too heavy for that use case.

- If the opt-in approach makes more sense for this PR, I’m happy to rework it accordingly.

- For the longer term, I’d also be happy to contribute this idea for 4.0. In that case, could you point me to the right workflow for 4.0-related work?

falkoschindler:
Thanks for the update — the `ElementFilter` revert, the dropped test, and the `scoped: bool = False` opt-in all landed cleanly, the regression test now exercises both default and opt-in paths, and 93/93 tests pass. Two issues still to address before merge, plus a few cleanup items.

### Worth considering: a context-manager API instead of a flag

Before we lock in `scoped=True`, an alternative shape that came up:

```py
with user.find('left-card'):
    await user.should_see('Button')
```

The idea is that `UserInteraction.__enter__` enters the slot of the found element, so all `should_see` / `find` / `should_not_see` calls inside the block are scoped to it automatically — without a per-call flag. It does stretch the purpose of `UserInteraction` (today it's an action object, not a scope), but the call site reads naturally and avoids opting into scoping on every assertion.

A more conservative variant would be a dedicated method:

```py
with user.focus('left-card'):
    await user.should_see('Button')
```

which keeps `UserInteraction` single-purpose at the cost of mirroring the `find` API.

Either form would let us drop the `scoped` parameter entirely (or keep it as a shorthand for single calls). Worth a moment of thought before merging — flags are sticky once shipped.

### Review findings

1. **Example test is broken.** [examples/pytests/tests/test_with_user.py:43-48](examples/pytests/tests/test_with_user.py#L43-L48) uses default `scoped=False` but expects scoped behavior. I ran it and it fails — `should_not_see(marker='duplicate-button', content='Shared Action Right')` correctly fires because that button exists on the right card and a default full-page search finds it. Mirror the changes already in `tests/test_user_simulation.py`: add `scoped=True` to the `should_not_see` and `find` calls that rely on scoping. This is the user-facing documentation for the feature, so it matters more than a normal bug.

2. **`scoped=True` at page level silently narrows to `client.content`.** Same root cause as the original BLOCKER, just hidden behind the opt-in. The carve-out in `_scope` ([nicegui/testing/user.py:51-53](nicegui/testing/user.py#L51-L53)) only fires when `slot_stack` is non-empty. With no surrounding `with element:`:
   - `slot_stack` empty → `_scope` returns `self._client` (line 47, before the carve-out)
   - `with self._scope:` enters `client.content`
   - `local_scope=True` resolves to `client.content` → misses header/drawer/footer/notifications

   So `await user.should_see('foo', scoped=True)` at page level quietly excludes layout shells. Either:
   - Move the carve-out into `_gather_elements` _after_ `with self._scope:` is entered, where you can compare `context.slot.parent` against `context.client.content` consistently, or
   - Treat the empty-stack and "slot is content" cases identically and let `_gather_elements` flip `local_scope` based on whether the user actually entered a sub-element.

   Worth adding a regression test for `scoped=True` without a surrounding context.

3. **`scoped` parameter undocumented.** New parameter on three public methods (`should_see`, `should_not_see`, `find`), but none of the docstrings mention it.

4. **Long lines** at [user.py:153 and :189](nicegui/testing/user.py#L153) — the `_gather_elements(scoped, target=target, kind=kind, marker=marker, content=content)` call is around 135 chars, exceeding our line length limit of 120.

5. **`scoped` is positional in `_gather_elements`** — the rest of the file uses keyword-only arguments. Make it keyword-only by moving it after the `*` for consistency.

evnchn:
Even if you start now, it is too late for 3.13, as we are commited to merge pytest PRs early in a milestone. See you next milestone! Looking forward to it.

evnchn:
*Posted by Claude Code (Opus 4.8) on @evnchn's behalf.*

**TL;DR:**

Sorry for the long silence — your May 11 review landed on my birthday and slipped past me. To make up for it I did the work **both ways** so you can pick the shape you want to live with (flags are sticky — you're right not to want to guess):

- **Merge-ready:** all five of your points fixed on the existing `scoped=` flag — branch [`evnchn:overnight/6015-flag-cleanup`](https://github.com/evnchn/nicegui/commit/723164de). Green on pre-commit / mypy / pylint 10.00 / full user-sim + example suites.
- **Prototype to try the feel:** a `with user.find('x'):` context-manager alternative — branch [`evnchn:overnight/6015-context-manager`](https://github.com/evnchn/nicegui/commit/4c3ff447).
- **The one that mattered:** `scoped=True` at page level no longer drops the layout shells (header / drawer / footer / notifications) — the original blocker, hiding behind the opt-in.

**Which API shape do you want — flag, context-manager, or both?** Once you call it I'll finalize on that branch and aim to land it early in 3.15 (our "merge pytest PRs early in a milestone" rule), rather than squeeze it into 3.14's last week.

<details>
<summary><b>1. Flag cleanup — all five review points (merge-ready)</b></summary>

Branch `evnchn:overnight/6015-flag-cleanup` ([commit `723164de`](https://github.com/evnchn/nicegui/commit/723164de)), one commit on top of the current PR head.

- **(`#2`, the real one)** `scoped=True` at page level silently narrowed the search to `client.content`, excluding the layout shells. Root cause: `ElementFilter(local_scope=True)` uses `context.slot.parent`, and at page level `with self._client:` makes that `client.content` — so header / drawer / footer / notifications fell outside. Fix in `_gather_elements`: only pass `local_scope=True` when `_scope` resolved to a **real sub-element**, not the page-level `Client` fallback:
  ```python
  local_scope = scoped and not isinstance(scope, Client)
  ```
  Added regression test `test_scoped_search_at_page_level_includes_layout` — it **fails on the current PR head** (`header-label` not found) and passes after the fix.
- **(`#1`)** Fixed the broken example test — added `scoped=True` to the scope-dependent `should_not_see` / `find` calls inside the `with left_card:` block.
- **(`#3`)** Documented `scoped` on `should_see` / `should_not_see` / `find`.
- **(`#4`)** Wrapped the over-length `_gather_elements` calls.
- **(`#5`)** `scoped` is now keyword-only in `_gather_elements`.

Default `scoped=False` behaviour is byte-identical to before — with the flag off, `local_scope` is `False` regardless of the new guard. Since edits-from-maintainers is on, I can push this onto the PR branch directly, or you can pull it — your call.

</details>

<details>
<summary><b>2. Context-manager alternative (prototype)</b></summary>

Branch `evnchn:overnight/6015-context-manager` ([commit `4c3ff447`](https://github.com/evnchn/nicegui/commit/4c3ff447), on top of branch 1):

```py
with user.find('left-card'):
    await user.should_see('Button')        # scoped automatically, no flag
    await user.should_not_see('Other')
```

`UserInteraction.__enter__` enters the matched element; a nesting-safe internal `_auto_scope` flag makes every lookup inside the block scope to it. Verified nesting- and exception-safe (the flag is restored on exit and on error). The `scoped=` flag still works for one-off calls, so we could keep both or drop the flag in favour of this.

**Open questions if you like the shape:**
- It currently requires the selector to match **exactly one** element (asserts in `__enter__`); multi-match `find()` would need a defined behaviour.
- It adds a small bit of state on `User` (`_auto_scope`).
- `with user.focus('x'):` is a near-free rename if you prefer that name over reusing `find`.

</details>

<details>
<summary><b>Verification — both branches</b></summary>

| Gate | flag branch | flag + CM branch |
|---|---|---|
| pre-commit | ✓ | ✓ |
| `mypy ./nicegui` | ✓ (238 files) | ✓ |
| `pylint` (user.py / user_interaction.py) | 10.00/10 | 10.00/10 |
| `tests/test_user_simulation.py` + `test_element_filter.py` | 71 (+ 11 example) | 95 |

The `#2` regression test is confirmed load-bearing: fails on PR head `7d14357c`, passes on the fix.

</details>

**Which shape do you want — flag-only, context-manager, or both?** I'll finalize on that branch and queue it early for 3.15.

evnchn:
@falkoschindler Let's review this early. It's the start of 3.15!

evnchn:
Useful direction — scoped `User` lookups are a real gap for pages that reuse markers/content, and the `examples/pytests` demo is the right instinct. But I don't think it's mergeable as-is: there's one **design call that's yours**, @falkoschindler, plus two test defects I verified by running the branch. Flagging early since we're at the start of 3.15.

**TL;DR:**

- **Design decision (needs a ruling):** the PR *description* says scoping is automatic when the test is inside a `with element:` context, but the *code* makes it opt-in via a new `scoped=True` (default still searches the whole page). They disagree — and the author's own example follows the description, so it fails.
- **Broken example:** `examples/pytests/tests/test_with_user.py::test_scoped_search_for_elements` fails (missing `scoped=True`), and CI never catches it — `testpaths = ["tests"]` and no workflow touches `examples/`.
- **Vacuous test → fixed:** the new `tests/test_element_filter.py` test never called `user.open()`, so its assertion never executed. One-line fix pushed at `evnchn/nicegui@cc57ccbd` (branch `fix-vacuous-scope-test`) — cherry-pick welcome.

Core mechanism itself is sound: `tests/test_user_simulation.py` + full regression = 93 passed, no regressions.

<details><summary><b>1. The design decision: automatic scoping vs explicit <code>scoped=</code></b></summary>

The description says:

> If the test is currently inside an element context, `User` uses that element as the search scope.

But the implementation adds `scoped: bool = False` to `find` / `should_see` / `should_not_see`; by default it searches the whole page even inside `with left_card:`. So scoping is **opt-in**, not automatic.

The tell that this confuses: the author's own teaching example assumes the *automatic* behavior and is therefore broken (see 2). Being inside `with left_card:` visually implies scoping, yet you must *also* pass `scoped=True` — two things to remember instead of one.

Two ways to resolve, both need your call:
- **(a) Automatic** — drop `scoped=`, let the slot stack decide. Matches the description + example. Risk: existing tests that call `find`/`should_see` inside a component context would silently start scoping (backward-compat break).
- **(b) Keep explicit `scoped=`** — then the description and example must be rewritten to match, not left describing (a). Minor: `User` uses `scoped=` while `ElementFilter` uses `local_scope=` for the same concept — worth aligning the vocab.
</details>

<details><summary><b>2. The bundled example actually fails (and CI doesn't run it)</b></summary>

`examples/pytests/tests/test_with_user.py::test_scoped_search_for_elements`, inside `with left_card:`, calls `should_not_see(..., content='Shared Action Right')` and asserts `find(marker='duplicated-marker').elements == 1` **without `scoped=True`** — so both run against the whole page.

Ran against the PR's own `nicegui`:
```
$ pytest examples/pytests/tests/test_with_user.py::test_scoped_search_for_elements
FAILED - AssertionError: expected not to see any element with marker='duplicate-button'
         and content='Shared Action Right' on the page:
           ... Button [markers=duplicate-button, label=Shared Action Right]
  (nicegui/testing/user.py:192)
```
The in-repo copy in `tests/test_user_simulation.py` gets this right (it passes `scoped=True` on the same lines), so the two copies disagree.

**Why CI is green anyway:** `pyproject.toml` → `testpaths = ["tests"]`, and no workflow references `examples/` at all. The example — whose stated purpose is discoverability — ships broken and uncovered. Might be worth a follow-up to run `examples/pytests` in CI regardless of this PR.
</details>

<details><summary><b>3. The new element_filter test was vacuous — fixed</b></summary>

`test_local_scope_false_searches_entire_page_inside_element_context` put its only `assert` inside an `@ui.page('/')` builder but never called `await user.open('/')`, so the builder never ran. Proof — it passed even with a deliberately-false expected value:
```
# assertion flipped to == ['THIS_SHOULD_FAIL_IF_RUN']
1 passed in 0.07s     # would fail if the body executed
```
Fix (`evnchn/nicegui@cc57ccbd`): add the trailing `await user.open('/')`, matching the idiom of every sibling test (`test_setting_props`, `test_typing`, …). After the fix it fails on a wrong assertion and passes on the correct one; pylint 9.92 → 9.96 (the `user` arg is now used).
</details>

@ZhuGC-Centogene thanks for this — the feature is genuinely wanted; the blocker is mostly the description/example/API alignment above, not the core logic.

ZhuGC-Centogene:
Hi @evnchn and @falkoschindler . Sorry for not responding sooner. I've been quite busy with work recently. Thanks so much for taking this over while I was away—I really appreciate it.

falkoschindler:
Thanks @evnchn for preparing both variants — that made this decision much easier. And thanks @ZhuGC-Centogene for kicking this off and iterating patiently!

**Ruling: let's go with the context manager, as a dedicated method, and drop the `scoped=` flag entirely.**

```py
with user.scope('left-card'):
    await user.should_see('Button')
    await user.should_not_see('Other')
```

Reasoning:

- In `User` tests you usually don't have an element reference for `with element:` — only markers/content. So a method that finds *and* enters the scope in one step is the practically relevant form.
- A `scoped=` parameter on three public methods is sticky. If a future major version ever makes scoping automatic via the slot stack, the flag immediately becomes legacy — while the context manager stays useful even then (find + enter in one step).
- A visible block that spans a scope reads more honestly than a flag repeated on every assertion — one thing to remember instead of two.
- Dedicated method rather than reusing `find`, so `UserInteraction` stays a pure action object.

On your open questions:

- **Exactly-one-match assertion in `__enter__`:** good semantics, keep it. Scoping to an ambiguous match would hide exactly the bugs this feature is meant to catch.
- **Name:** I'd like to move away from my earlier `focus` suggestion — in a UI context it too strongly suggests keyboard/DOM focus, and we may want to reserve `user.focus(...)` for actual focus simulation one day. My proposal is **`scope`**, which also aligns with `ElementFilter`'s `local_scope` vocabulary. Other candidates we considered: `within` (clear, but "with user.within" stutters), `inside`, `look_into` (fits the narrative voice of the User API, but less obvious that it spans a scope). Opinions welcome — once shipped, the name is fixed.
- Please also cherry-pick your vacuous-test fix (evnchn/nicegui@cc57ccbd) onto the final branch.

Separately, running `examples/pytests` in CI deserves a follow-up issue — a broken example staying green is worse than no example. @evnchn, would you open that one?

@evnchn Feel free to finalize on the context-manager branch (drop the flag, add docstrings) and push directly to this PR — then we can land it early in 3.15 as planned. Let's just settle the name here first.

evnchn:

**TL;DR:**

Agreed on **`scope`** — it aligns `User` with `ElementFilter`'s `local_scope` vocabulary, which is exactly the naming inconsistency I'd flagged earlier. No objection to the ruling; the context-manager-as-dedicated-method shape is the right one.

Finalized on the context-manager branch and pushed to this PR:

- dropped the `scoped=` flag from `find` / `should_see` / `should_not_see` entirely — scoping now happens **only** inside a `with user.scope(...):` block;
- `UserInteraction` stays a pure action object (the CM lives on `User`, not on `find`'s return);
- kept the exactly-one-match assertion at enter; it's exception-safe, nesting-safe, and **task-local** (scope state is a per-asyncio-task depth keyed like `Slot.stacks`, so a scope in one task never narrows another task's lookups);
- cherry-picked the vacuous-test fix (`cc57ccbd`);
- added docstrings + rewrote the in-repo test and the `examples/pytests` demo to the new API.

CI is green locally (mypy clean, pylint 10.00/10, tests pass). Opened #6141 for the "run `examples/pytests` in CI" follow-up.

**Ask:** nothing's merged yet — if you'd still prefer a different name, say so and I'll rename before we land it in 3.15. Otherwise this is ready.

<details>
<summary><b>What changed (API surface)</b></summary>

```py
with user.scope(marker='left-card'):
    await user.should_see('Button')       # only inside the left card
    await user.should_not_see('Other')
    assert len(user.find(marker='dup').elements) == 1
```

- **`User.scope(target=None, *, kind=None, marker=None, content=None)`** — a `@contextmanager` method. It resolves the filter to the single matching element, enters it, and scopes every `should_see` / `should_not_see` / `find` inside the block to it. It asserts an **exactly-one** match (an ambiguous scope would hide the very bugs this feature catches) and restores the previous scope on exit — including when the block raises.
- **`scoped=` removed** from `should_see`, `should_not_see`, `find`. It was never in a release (added within this PR), so there's no deprecation path to worry about.
- Page-level lookups are unchanged: outside any `scope()` block the search still covers the whole layout (header / drawer / footer / notifications). The `local_scope` narrowing is gated on the current task being inside a `scope()` block, with a `not isinstance(scope, Client)` safety net.
- Scope state is **per-asyncio-task**, keyed by task id like `Slot.stacks`, rather than a single instance flag — so concurrent `user` interactions in different tasks don't leak scope into each other.
</details>

<details>
<summary><b>On your open questions</b></summary>

- **Exactly-one in `__enter__`:** kept, as you asked — `scope()` raises `AssertionError` if the filter matches 0 or >1 elements, with the found count in the message.
- **`UserInteraction` purity:** untouched — no `__enter__`/`__exit__` added to it. `scope()` is dedicated on `User`.
- **Vacuous-test fix (`cc57ccbd`):** cherry-picked. The test now calls `await user.open('/')`, so its assertion actually runs.
</details>

<details>
<summary><b>Verification (ran on the final branch)</b></summary>

| Gate | Result |
|---|---|
| `pre-commit` (all hooks) | ✓ |
| `mypy ./nicegui` | ✓ no issues, 238 files |
| `tests/test_user_simulation.py` + `tests/test_element_filter.py` | 97 passed |
| `examples/pytests` | 11 passed |

New regression tests in `tests/test_user_simulation.py`:
- `test_scoped_search_for_elements` — `find(...).elements == 1` inside the scope vs `== 2` outside (the direct before/after that proves scoping actually narrows).
- `test_scope_requires_exactly_one_match` — ambiguous marker raises.
- `test_scope_restored_after_exception` — scope restored after the block raises.
- `test_nested_scope` — inner scope narrows further, outer restored on exit.
- `test_scope_is_task_local` — a concurrent task that never called `scope()` still sees the whole page while a sibling task is scoped (verified it fails with a shared-state implementation, passes with the task-keyed one).

Reviewed before push by two independent passes — a different-lineage model (clean) and a Claude subagent that flagged the scope state being instance-level rather than task-local; that's the fix now covered by `test_scope_is_task_local`.
</details>

falkoschindler (changes requested):
Thanks for the continued iterations on this! This review has two parts: concrete findings on the current implementation, and a bigger-picture thought about the API shape that emerged while verifying them.

## Part 1 — Findings

The happy paths work nicely — nesting, task isolation and exception restoration are all covered by tests. But the stateful scoping machinery produces several verified edge-case behaviors, one of them a silent behavior change for existing test suites.

<details>
<summary><b>Details</b> (1 blocker, 5 major, cleanups)</summary>

**BLOCKER**

- **PR description drift:** the Implementation section still describes the abandoned design ("resolves a `scope` from the current `context.slot_stack`", example using `with user:`, `ElementFilter(..., local_scope=True)`, and `with left_card:`), while the final code only narrows inside `user.scope(...)` blocks and a bare `with left_card:` deliberately does *not* scope. Please rewrite it around `user.scope()` so readers of the merged PR get the right model.

**MAJOR**

- `nicegui/testing/user.py` (`_gather_elements`) — passing an explicit `local_scope=False` for unscoped lookups silently overrides the documented `ElementFilter.DEFAULT_LOCAL_SCOPE` knob. Verified empirically: with `DEFAULT_LOCAL_SCOPE = True` and a label in `ui.header`, `await user.should_not_see('Admin')` passes on main and fails on this branch, because the search root changes from `client.content` to `client.layout`. The old interaction was undocumented and arguably accidental, so regularizing it may be fine — but it should be a conscious decision and mentioned in the PR.
- `should_see`/`should_not_see` — notifications bypass the scope: `self.notify.contains(target)` is checked globally before the scoped element search. Inside `with user.scope(marker='left-card'):`, a `ui.notify('Saved')` triggered anywhere makes scoped `should_not_see('Saved')` fail and scoped `should_see('Saved')` pass, contradicting the docstring's "only search within the entered element".
- Scoped element self-invisibility — `ElementFilter(local_scope=True)` iterates `scope.descendants()` with `include_self=False`, so `with user.scope(marker='card'): await user.should_see(marker='card')` fails (and `should_not_see` of the scope's own marker absurdly passes). Probably acceptable "search *inside*" semantics, but the docstring ("scoping all lookups ... to it") suggests otherwise.
- `scope()` gathers exactly once with no retry, unlike `should_see`'s 3-attempt loop. Scoping to an element that appears asynchronously (content built after `await client.connected()` or via a timer) raises "found 0" immediately where `should_see` would have succeeded on retry — a flakiness trap.
- The scoping machinery is more implicit than it needs to be: a per-task depth counter, a 3-branch `_scope` property inferring the scope from the ambient slot stack, a `with scope:` re-entry and an `isinstance(scope, Client)` safety net. Because the counter only records that *some* scope is active but not *which* element, two verified surprises emerge: (1) a manual `with some_element:` inside a scope block silently retargets all lookups to that element, while the identical block outside a scope doesn't narrow at all; (2) `await user.open(...)` inside a scope block silently disables scoping instead of erroring. Storing the entered element itself (per task) and applying it via the existing `ElementFilter(...).within(instance=...)` would delete the property, both fallbacks and the guard.

**CLEANUP**

- `scope()` returns `Iterator[T]` without overloads (unlike `find`), so `T` is unsolved for marker/content/string-target calls; mypy reports `Need type annotation for "card"` on `with user.scope(marker='x') as card:` (verified).
- `examples/pytests` — I'd revert the example changes entirely. The example's purpose is to show how to set up pytest for a NiceGUI app at all, not to document individual `User`/`Screen` API features; the new page and test also mirror `tests/test_user_simulation.py::test_scoped_search_for_elements` almost verbatim. If the feature needs user-facing documentation, the right place is the "Querying" section of the [User documentation](https://nicegui.io/documentation/user#querying), which already covers `ElementFilter` and `.within(...)`. (Reverting also moots a side issue: the example currently requires an unreleased NiceGUI while `requirements.txt` pins `nicegui>=3.0`, so `pip install -r requirements.txt && pytest` fails until the next release.)
- `tests/test_element_filter.py` — `test_local_scope_false_searches_entire_page_inside_element_context` duplicates existing coverage: `test_default_local_scope` already asserts `ElementFilter(kind=ui.label, local_scope=False)` sees the whole page inside an element context, under both `DEFAULT_LOCAL_SCOPE` settings.
- `tests/test_user_simulation.py` — `asyncio.create_task(other())` in `test_scope_is_task_local`; CONTRIBUTING.md says to use `background_tasks.create()` instead (harmless here since the task is awaited, but worth aligning or a short comment).
- `nicegui/element_filter.py` — the only change is a removed blank line; please revert this unrelated hunk so the file drops out of the PR.
- Minor: the two long condition lines in `should_see`/`should_not_see` (127/136 chars) would read better wrapped; the `if depth > 1: decrement else: del` in `scope()` reads simpler as decrement-then-delete-at-zero; the keyword-only `*` in `_gather_elements` forces verbose call sites for no gain; the same two-sentence docstring paragraph is pasted three times.

</details>

## Part 2 — Bigger picture: revisiting the API shape

Stepping back: I ruled for the context manager earlier in this thread, and among the options compared then — automatic scoping, a `scoped=` boolean, a context manager — I still think it was the right call. But implementing it fully has surfaced costs that weren't visible at ruling time, and there's a fourth shape we never actually compared: a per-call **selector** parameter, `within=`. Knowing what we know now, I'd like to propose that as the core mechanism — with the option of adding `scope()` back later as sugar on top.

<details>
<summary><b>Details</b> (what changed since the ruling, and the <code>within=</code> proposal)</summary>

**What changed since the ruling.** The July 4 decision compared *automatic* vs *boolean flag* vs *context manager*. The rejected "flag" was a boolean that only worked in combination with ambient context (`with left_card:` **plus** `scoped=True` — two things to remember). A self-contained selector parameter is a different shape and was never on the table. Meanwhile, the finished CM implementation shows what statefulness actually costs: per-task depth bookkeeping, a `_scope` property with fallbacks, tests for exception restoration and task isolation — and still the open edge cases above (stale scope across rebuilds/navigation, one-shot resolution without retries). Each round has uncovered new questions of this kind; that's a property of the stateful shape, not of anyone's implementation.

**The proposal:** a `within=` parameter on `should_see`, `should_not_see` and `find`, mapping directly onto the existing public `ElementFilter.within(instance=...)` (marker string or element reference):

```python
await user.should_see(marker='duplicate-button', within='scope-card left')
assert len(user.find(marker='duplicated-marker', within='scope-card left').elements) == 1
```

What this buys, concretely mapped to the findings:

- No state → no nesting, task-locality or exception-restoration machinery, and no stale scope across navigation or page rebuilds.
- `within='marker'` is re-resolved on every retry attempt, so asynchronously appearing containers work naturally — the flakiness trap disappears for free.
- No interaction with `ElementFilter.DEFAULT_LOCAL_SCOPE` or the ambient slot stack at all.
- It reuses the mechanism the library already ships (`ElementFilter.within`) and aligns the vocabulary, instead of building a parallel one.

**To be fair, two findings are shape-agnostic** and need a decision either way: whether notifications count inside a narrowed search, and that the scope/`within` target itself is excluded from matches (`descendants()`/`ancestors()` semantics). And the CM's strongest argument still holds against `within=`: a block with many assertions states the scope once, a parameter repeats it. That's real — it's why I'd keep the door open for `scope()` as sugar once the stateless core is solid, rather than shipping both now.

My preference is `within=` as the core. It's also the smaller PR — and after everyone's patience on this one, I'm happy if we (maintainers) do the final reshape rather than asking for yet another round.

</details>

---

One more housekeeping note: the branch is currently 90 commits behind main — worth merging main before final approval so CI runs against the current state.

evnchn:
*Posted by Claude Code (Opus 4.8) on @evnchn's behalf.*

**TL;DR:**

I built the finding‑5 refactor you sketched (store the entered element, apply `.within(instance=...)`) — and it lands the *opposite* way from the `within=` pivot: it resolves the blocker‑class findings **while keeping `scope()`**, so I don't think we need to reshape the core.

- **Your own finding‑5 fix resolves findings 1, 2 and 5** and deletes the `_scope` property + the depth counter — without touching the exactly‑one guarantee or the block ergonomics.
- **The `within=` proposal has one gap worth seeing before we commit:** a marker target (`within='left-card'` → `ElementFilter.within(marker=...)`) **silently drops the exactly‑one‑match guarantee** you called the whole point on Jul 4. With two `left-card`s on the page, `within='left-card'` happily searches *both* — the duplicate‑container bug the feature exists to catch walks straight back in. `scope(marker='left-card')` raises "found 2".
- **Green + different‑lineage reviewed:** pre‑commit / mypy / pylint 10.00 clean, 100 tests (3 new, each proven load‑bearing on the branch). A Codex adversarial pass caught 2 real regressions in my first cut; both fixed and re‑verified.

**Ask:** does keeping `scope()` on this refactor match your intent — or do you still want `within=` as the core, in which case, how should it preserve exactly‑one for marker targets? Happy either way; the three commits are already on the branch (CI running).

<details>
<summary><b>The refactor — how it maps to your Part‑1 findings</b></summary>

Three commits now on `scoped-user` (`9d26db90`, `71f8bc94`, `a57bf3f9`) on top of PR head `06b0dafc`:

| # | finding | status | evidence |
|---|---|---|---|
| BLOCKER | PR description drift | ⏳ pending | doc rewrite around `user.scope()` — see "still open" |
| MAJOR 1 | explicit `local_scope=False` overrides `DEFAULT_LOCAL_SCOPE` | ✅ resolved | refactor stops passing `local_scope` entirely; respects the knob and matches `main` (probe below) |
| MAJOR 2 | notifications bypass the scope | ✅ resolved | gated on `_is_scoped()`; `test_scope_excludes_global_notifications` fails without the gate |
| MAJOR 3 | scope's own element is self‑invisible | 📝 docstring | clarified to "descendants … the element itself is not matched" — behavior unchanged & defensible |
| MAJOR 4 | `scope()` resolves once, no retry for async containers | ⚠️ **not fixed, deliberate** | a CM that resolves its container once at enter is correct‑by‑design; a `retries=` on `scope()` could be added later if wanted |
| MAJOR 5 | implicit machinery: `_scope` property, depth counter, ambient‑slot retargeting | ✅ resolved | element stored per‑task explicitly; `test_scope_ignores_manual_element_context` fails on the old code, passes here |

Mechanism: `scope()` pushes the resolved element onto a per‑task stack (`_scope_stack: dict[int, list[ui.element]]`, keyed by `get_task_id()` like `Slot.stacks`), still enters `with element:`, and pops on exit. `_gather_elements` reads `stack[-1]` and applies `ElementFilter(...).within(instance=scope)` on a normal full‑page search — no `local_scope` toggle, no reading the ambient slot stack. `user.py` is net **simpler** (the 3‑branch `_scope` property + counter are gone).

Empirical proof that MAJOR 1 is back to `main` behavior (search root respects the knob again):

```python
async def test(user):
    ElementFilter.DEFAULT_LOCAL_SCOPE = True
    @ui.page('/')
    def page():
        with ui.header():
            ui.label('Admin')
        ui.label('Body')
    await user.open('/')
    await user.should_not_see('Admin')  # header is outside client.content -> not seen (like main)
    await user.should_see('Body')
# -> 1 passed
```
</details>

<details>
<summary><b>Why <code>within='marker'</code> loses the exactly‑one guarantee (the reason I'd keep the CM)</b></summary>

`ElementFilter.within(marker=...)` keeps every element whose ancestor markers *include* the given marker — it never asserts how many containers matched. From `element_filter.py`:

```python
ancestor_markers = {marker for ancestor in ancestors for marker in ancestor._markers}
if self._within_markers and not ancestor_markers.issuperset(self._within_markers):
    continue
```

So on a page with two `scope-card left`s:

- `with user.scope(marker='scope-card left'):` → **`AssertionError: expected exactly one element to scope to, but found 2`** — forces the author to disambiguate. This is the Jul‑4 semantics you ruled in.
- `await user.should_see(marker='dup', within='scope-card left')` → searches descendants of **both** cards; `find(marker='dup', within='scope-card left').elements` returns 2. The duplicate‑container ambiguity the feature is meant to surface is silently swallowed.

To keep the guarantee, `within='marker'` would have to resolve the marker to a single element per call (a mini‑`scope()` inside every `should_see`) — i.e. not the "maps directly onto `ElementFilter.within`" simplicity, just moved from once‑per‑block to once‑per‑assertion. `within=<element ref>` keeps the guarantee but needs a reference User tests usually don't have (your own Jul‑4 point). That's the trade that made me stop at "keep `scope()`; add `within=` later as sugar on this same `.within(instance=...)` substrate if we want it."

Two findings remain shape‑agnostic and are decided here regardless: notifications (MAJOR 2, now excluded inside a scope) and self‑exclusion (MAJOR 3, docstring‑clarified).
</details>

<details>
<summary><b>Different‑lineage (Codex) adversarial review — what it caught</b></summary>

Ran a Codex pass over the diff framed as "assume it's broken." It cleared task‑locality, the `.within` semantics vs old `local_scope`, nesting order, the notification truth‑table (all 4 combos), and the `with self._client:` context reset — and returned **2 real blockers**, both now fixed in `a57bf3f9`:

1. **Cross‑task cleanup KeyError.** I called `get_task_id()` again in `finally`, so a CM driven across tasks deleted the wrong key. Reproduced by Codex. Fix: capture `task_id` once at enter (restores the prior design's robustness).
2. **Dropping `with element:` broke UI‑building inside a scope.** Creating `ui.label(...)` inside `with user.scope(...):` no longer landed in the scoped element. Reproduced by Codex. Fix: restored `with element:` — it's orthogonal to the finding‑5 fix (lookups read the *stored* element, not the ambient slot), verified by both `test_scope_ignores_manual_element_context` (the #5 fix) and the new `test_scope_enters_the_element_for_building` staying green together.

Caveat it flagged honestly: its sandbox blocked `pytest`/`pylint` (only `mypy` ran there) — so the suite/lint numbers below are from the local venv, not Codex's box.
</details>

<details>
<summary><b>Verification (local venv on the branch)</b></summary>

| gate | result |
|---|---|
| `pre-commit` (all hooks) | ✓ |
| `mypy ./nicegui` | ✓ no issues, 238 files |
| `tests/test_user_simulation.py` + `tests/test_element_filter.py` | **100 passed** |

New tests, each confirmed load‑bearing (fails on the pre‑fix state, passes after):

- `test_scope_ignores_manual_element_context` — a manual `with other:` inside a scope no longer retargets lookups (proves MAJOR 5).
- `test_scope_excludes_global_notifications` — a page‑level `ui.notify` is not matched inside a scope (proves MAJOR 2).
- `test_scope_enters_the_element_for_building` — UI created inside `with user.scope(...):` lands in the scoped element (pins the `with element:` behavior Codex flagged).

Not yet run: full CI on the pushed commit (local ≠ CI). I'll wait for green before calling it ready.
</details>

<details>
<summary><b>Still open (not in this diff)</b></summary>

- **BLOCKER — PR description rewrite** around `user.scope()` (drops the abandoned auto‑scoping / `scoped=` narrative). Doc task; I'll do it with the push.
- **Cleanups from your review:** `scope()` mypy overloads (so `with user.scope(marker='x') as card:` types), revert the `examples/pytests` changes, drop the duplicate `test_local_scope_false_searches_entire_page_inside_element_context`, `asyncio.create_task` → `background_tasks.create()` in `test_scope_is_task_local`, and the stray blank‑line hunk in `element_filter.py`. Can fold these in before you re‑review or leave for a maintainer pass — your call.
- **Branch is ~75 behind main** — I'll merge main before final approval so CI runs against current state.
</details>

evnchn:
Unfortunately we didn't fit in the early-3.15 window. Better aim for the next milestone than to risk our pytest, I'd say. Sorry @ZhuGC-Centogene 🙇

falkoschindler (approved):
@evnchn To close the loop on your July 7 question: yes, keeping `scope()` as the core was the right call — the `.within(instance=...)` refactor landed exactly the way I hoped, and the exactly-one guarantee is precisely what distinguishes it from a hypothetical `within=` parameter. Thanks @ZhuGC-Centogene for kicking this off and sticking with it through the redesign, and @evnchn for the refactor work.

Since the last round I finished the remaining cleanups and ran one more deep review pass over the branch. What changed:

- User lookups now always search the whole page, independent of `ElementFilter.DEFAULT_LOCAL_SCOPE` (documented with a 3.16.0 note, pinned by a parametrized test).
- A deleted scope element now fails loudly: if the entered element is rebuilt inside the block (e.g. by a `ui.refreshable` refresh), lookups raise instead of silently finding nothing — which would have let `should_not_see` pass for the wrong reason.
- Task-locality is documented and tested in both directions: a sibling task stays unscoped, and a task created _inside_ the block (e.g. via `asyncio.gather`) is unscoped too.
- Typing overloads for `scope()`, scope-aware assertion error messages, one shared visibility predicate instead of mirrored conditions in `should_see`/`should_not_see`, plus a few small cleanups.

One deliberate non-fix, recorded here explicitly: `scope()` still resolves its target exactly once and does not retry like `should_see` does — @evnchn's "correct-by-design" call from the July 7 table stands. The docstring documents the workaround (`await user.should_see(...)` first for asynchronously created content). If this bites in practice, a follow-up could add retrying — for example via a result object that works with both `with` and `async with`. If anyone feels it should retry from day one, speak up now.

Approving. Let's merge! 🚀 [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20findings%20and%20wording%20from%20the%20model%3B%20decisions%20on%20scope%20and%20approval%20from%20the%20maintainer.&p=agent&a=claude-code "claude-fable-5: Review findings and wording from the model; decisions on scope and approval from the maintainer.")

Issue #1, opened 2021-05-24
Callbacks vs Streamlit-Like "if"-statements


elimintz:
Hi, 

I like very much what you did with nicegui and how you used justpy! Very nice.

I also looked at streamlit. Could you expound a little about why that was not a good solution for you in the end?

One feature of streamlit that I find intriguing is the ability to assign the value of a component to a variable without the need of callbacks. Of course, the callbacks happen but are transparent to the user of the framework. I am trying to think how this could be implemented in JustPy. It would be a great feature for nicegui, making it even simpler to use. What is your view on this and how do you think it could be implemented?

If you prefer to have the discussion by email instead of here, my email is eli.mintz@gmail.com or you can also open an issue in the JustPy repository.

Eli

falkoschindler:
Hi Eli,

Our problem with Streamlit was mainly due to the way it handles user interaction. Writing `if st.button(‘Click’): st.text(‘Hey!’)` is super cool. But it hides the underlying event loop by implicitly re-evaluating the script. That works in simple cases, but it gets quickly rather complicated to achieve apparently “normal” behavior.

**State**

If the script starts with loading some non-constant initial state, e.g. from an external source, this state is reset whenever the script is processed again. The following script already shows inconsistent behavior. Initializing `state` with a constant works, but assigning a non-constant random value overwrites the user-selected state every time.

```
import streamlit as st
import random

state = random.randint(0, 5) # simulate external source
new_state = st.radio(‘State’, [‘A’, ‘B’, ‘C’, ‘D’, ‘E’, ‘F’], state)
st.text(‘State: ’ + new_state)
```

It is hard to let Streamlit load the state once and let the user update it repeatedly. It gets worse when multiple widgets interact with each other, like a button resetting the radio selection to some value.

There are workarounds introducing session state. But they are not trivial and require quite some additional code. You can look into one of our projects to see how we mapped machine state onto session state to make it accessible for Streamlit:
https://github.com/zauberzeug/odrive-gui/blob/9d8a6d3b46a8e1f200670aa837bff476d98504a5/src/main.py#L21

**Loops**

Always reloading the script makes it also difficult to introduce own event loops, e.g. reading data and updating a plot repeatedly. In the above-mentioned project we found a workaround using threads that need to get stopped and restarted every time the script restarts due to user interaction:
https://github.com/zauberzeug/odrive-gui/blob/9d8a6d3b46a8e1f200670aa837bff476d98504a5/src/main.py#L40

[...]

PR #6016, opened 2026-04-29
Make `enable()`, `disable()`, `set_enabled()` and `set_visibility()` chainable

### LINKED ISSUE #6014: Split in dropdown_button overrides functionality of props("disabled")
### First Check

### Example Code

```python
from nicegui import ui


ui.label("With .props('disabled') and split=True")
with ui.dropdown_button('Start - disabled', icon='start', on_click=lambda:ui.notify("Start - disabled"), split=True, auto_close=True, color="green").props("disabled"):
    with ui.row():
        ui.button("Configure", on_click=lambda:ui.notify("Configure"), color="lightgreen").props("enabled")

ui.label("With .props('disabled') and split=False")
with ui.dropdown_button('Start - disabled', icon='start', on_click=lambda:ui.notify("Start - disabled"), split=False, auto_close=True, color="green").props("disabled"):
    with ui.row():
        ui.button("Configure", on_click=lambda:ui.notify("Configure"), color="lightgreen").props("enabled")

ui.run()
```

### Description

When split is set to **true** in  dropdown_button and it gets disabled via .props. It turns grey but on_click is still called when clicked. 
If split is set to false it behaves as expected.

Is there a dedicated property for this?! Cause it works as expected when split is set to false.

### NiceGUI Version

3.11.0

### Python Version

Python 3.10.12

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

Edit: Changed Version from 3.3.1 to 3.11.0

falkoschindler:
### Motivation

Closes #6014.

In #6014 a user tried to disable a `ui.dropdown_button` inline as part of a `with` statement. Because `.disable()` and `.set_enabled(False)` return `None`, the only chainable option was `.props('disabled')` — which silently misbehaves with `split=True` (the HTML `disabled` attribute falls through to the wrapper `<div>` instead of the inner buttons, so the click handler still fires).

The misbehavior with `split=True` is not actually a NiceGUI bug — `.props('disabled')` is the wrong prop name (Quasar's prop is `disable`, without the trailing `d`). But the user's underlying motivation is reasonable: they wanted a fluent form usable inside `with`. Today we don't offer one for enabled/visible state, even though `.props(...)`, `.classes(...)`, `.style(...)` and the `bind_*` methods already return `Self`.

### Implementation

Change the return type of the following methods from `None` to `Self` and add `return self`:

- `DisableableElement.enable()`
- `DisableableElement.disable()`
- `DisableableElement.set_enabled(value)`
- `Visibility.set_visibility(visible)`

This is a backward-compatible change — existing call sites ignore the return value, and the new return value matches the documented "fluent interface returns `Self`" convention in `CONTRIBUTING.md`. It enables patterns like:

```python
with ui.dropdown_button('Start', split=True, on_click=...).disable():
    ui.menu_item('...')

with ui.card().set_visibility(show_advanced):
    ...
```

### Progress

- [x] Should the event handler exclude elements to allow `Dialog.open()` be chainable?

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -478,7 +478,7 @@ class Element(Visibility):
     def move(self,
              target_container: Element | None = None,
              target_index: int = -1, *,
-             target_slot: str | None = None) -> None:
+             target_slot: str | None = None) -> Self:
         """Move the element to another container.
 
         :param target_container: container to move the element to (default: the parent container)
@@ -505,6 +505,7 @@ class Element(Visibility):
         parent_slot.children.insert(target_index, self)
 
         target_container.update()
+        return self
 
     def remove(self, element: Element | int) -> None:
         """Remove a child element.
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -39,7 +39,7 @@ TAG_CHAR = TAG_START_CHAR + r'|-|\.|[0-9]|\u00B7|[\u0300-\u036F]|[\u203F-\u2040]
 TAG_PATTERN = re.compile(fr'^({TAG_START_CHAR})({TAG_CHAR})*$')
 
 
-class Element(Visibility, helpers.NoImplicitAwait):
+class Element(Visibility):
     component: Component | None = None
     exposed_libraries: ClassVar[list[Library]] = []
     _default_props: ClassVar[dict[str, Any]] = {}
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -39,7 +39,7 @@ TAG_CHAR = TAG_START_CHAR + r'|-|\.|[0-9]|\u00B7|[\u0300-\u036F]|[\u203F-\u2040]
 TAG_PATTERN = re.compile(fr'^({TAG_START_CHAR})({TAG_CHAR})*$')
 
 
-class Element(Visibility):
+class Element(Visibility, helpers.NoImplicitAwait):
     component: Component | None = None
     exposed_libraries: ClassVar[list[Library]] = []
     _default_props: ClassVar[dict[str, Any]] = {}
--- a/nicegui/elements/audio.py
+++ b/nicegui/elements/audio.py
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+from typing_extensions import Self
+
 from ..defaults import DEFAULT_PROP, resolve_defaults
 from .mixins.source_element import SourceElement
 
@@ -33,7 +35,7 @@ class Audio(SourceElement, component='audio.js'):
         self._props.set_bool('muted', muted)
         self._props.set_bool('loop', loop)
 
-    def set_source(self, source: str | Path) -> None:
+    def set_source(self, source: str | Path) -> Self:
         return super().set_source(source)
 
     def seek(self, seconds: float) -> None:
--- a/nicegui/elements/button_dropdown.py
+++ b/nicegui/elements/button_dropdown.py
@@ -60,14 +60,17 @@ class DropdownButton(IconElement, TextElement, DisableableElement, BackgroundCol
     def _text_to_model_text(self, text: str) -> None:
         self._props['label'] = text
 
-    def open(self) -> None:
+    def open(self) -> Self:
         """Open the dropdown."""
         self.value = True
+        return self
 
-    def close(self) -> None:
+    def close(self) -> Self:
         """Close the dropdown."""
         self.value = False
+        return self
 
-    def toggle(self) -> None:
+    def toggle(self) -> Self:
         """Toggle the dropdown."""
         self.value = not self.value
+        return self
--- a/nicegui/elements/choice_element.py
+++ b/nicegui/elements/choice_element.py
@@ -1,5 +1,7 @@
 from typing import Any
 
+from typing_extensions import Self
+
 from ..events import Handler, ValueChangeEventArguments
 from .mixins.value_element import ValueElement
 
@@ -52,7 +54,7 @@ class ChoiceElement(ValueElement[Any]):
             self._update_options()
         super().update()
 
-    def set_options(self, options: list | dict, *, value: Any = ...) -> None:
+    def set_options(self, options: list | dict, *, value: Any = ...) -> Self:
         """Set the options of this choice element.
 
         :param options: The new options.
@@ -62,3 +64,4 @@ class ChoiceElement(ValueElement[Any]):
         if value is not ...:
             self.value = value
         self.update()
+        return self
--- a/nicegui/elements/codemirror/codemirror.py
+++ b/nicegui/elements/codemirror/codemirror.py
@@ -1,6 +1,8 @@
 from itertools import accumulate, chain, repeat
 from typing import Literal, get_args
 
+from typing_extensions import Self
+
 from ...defaults import DEFAULT_PROP, resolve_defaults
 from ...elements.mixins.disableable_element import DisableableElement
 from ...elements.mixins.value_element import ValueElement
@@ -310,9 +312,10 @@ class CodeMirror(ValueElement[str], DisableableElement,
     def theme(self, theme: SUPPORTED_THEMES) -> None:
         self._props['theme'] = theme
 
-    def set_theme(self, theme: SUPPORTED_THEMES) -> None:
+    def set_theme(self, theme: SUPPORTED_THEMES) -> Self:
         """Sets the theme of the editor."""
         self._props['theme'] = theme
+        return self
 
     @property
     def supported_themes(self) -> list[str]:
@@ -328,9 +331,10 @@ class CodeMirror(ValueElement[str], DisableableElement,
     def language(self, language: SUPPORTED_LANGUAGES | None = None) -> None:
         self._props['language'] = language
 
-    def set_language(self, language: SUPPORTED_LANGUAGES | None = None) -> None:
+    def set_language(self, language: SUPPORTED_LANGUAGES | None = None) -> Self:
         """Sets the language of the editor (case-insensitive)."""
         self._props['language'] = language
+        return self
 
     @property
     def supported_languages(self) -> list[str]:
@@ -349,12 +353,13 @@ class CodeMirror(ValueElement[str], DisableableElement,
     def line_wrapping(self, value: bool) -> None:
         self._props['line-wrapping'] = value
 
-    def set_line_wrapping(self, value: bool) -> None:
+    def set_line_wrapping(self, value: bool) -> Self:
         """Sets whether line wrapping is enabled.
 
         *Added in version 3.2.0*
         """
         self._props['line-wrapping'] = value
+        return self
 
     def _event_args_to_value(self, e: GenericEventArguments) -> str:
         """The event contains a change set which is applied to the current value."""
--- a/nicegui/elements/color_input.py
+++ b/nicegui/elements/color_input.py
@@ -2,6 +2,8 @@ import re
 from colorsys import rgb_to_yiq
 from typing import Any
 
+from typing_extensions import Self
+
 from ..defaults import DEFAULT_PROP, DEFAULT_PROPS, resolve_defaults
 from ..events import Handler, ValueChangeEventArguments
 from .button import Button as button
@@ -46,11 +48,12 @@ class ColorInput(LabelElement, ValueElement[str | None], DisableableElement):
         self.preview = preview
         self._update_preview()
 
-    def open_picker(self) -> None:
+    def open_picker(self) -> Self:
         """Open the color picker"""
         if self.value:
             self.picker.set_color(self.value)
         self.picker.open()
+        return self
 
     def _handle_value_change(self, value: Any) -> None:
         super()._handle_value_change(value)
--- a/nicegui/elements/color_picker.py
+++ b/nicegui/elements/color_picker.py
@@ -29,12 +29,13 @@ class ColorPicker(Menu):
                     handle_event(handler, ColorPickEventArguments(sender=self, client=self.client, color=e.args))
             self.q_color = Element('q-color').on('change', handle_change)
 
-    def set_color(self, color: str) -> None:
+    def set_color(self, color: str) -> Self:
         """Set the color of the picker.
 
         :param color: the color to set
         """
         self.q_color.props(f'model-value="{color}"')
+        return self
 
     def on_pick(self, callback: Handler[ColorPickEventArguments]) -> Self:
         """Add a callback to be invoked when a color is picked."""
--- a/nicegui/elements/context_menu.py
+++ b/nicegui/elements/context_menu.py
@@ -1,3 +1,5 @@
+from typing_extensions import Self
+
 from ..element import Element
 
 
@@ -14,10 +16,12 @@ class ContextMenu(Element):
         self._props['context-menu'] = True
         self._props['touch-position'] = True
 
-    def open(self) -> None:
+    def open(self) -> Self:
         """Open the context menu."""
         self.run_method('show')
+        return self
 
-    def close(self) -> None:
+    def close(self) -> Self:
         """Close the context menu."""
         self.run_method('hide')
+        return self
--- a/nicegui/elements/dark_mode.py
+++ b/nicegui/elements/dark_mode.py
@@ -1,3 +1,5 @@
+from typing_extensions import Self
+
 from ..defaults import DEFAULT_PROP, resolve_defaults
 from ..events import Handler, ValueChangeEventArguments
 from .mixins.value_element import ValueElement
@@ -23,15 +25,17 @@ class DarkMode(ValueElement[bool | None], component='dark_mode.js'):
         """
         super().__init__(value=value, on_value_change=on_change)
 
-    def enable(self) -> None:
+    def enable(self) -> Self:
         """Enable dark mode."""
         self.value = True
+        return self
 
-    def disable(self) -> None:
+    def disable(self) -> Self:
         """Disable dark mode."""
         self.value = False
+        return self
 
-    def toggle(self) -> None:
+    def toggle(self) -> Self:
         """Toggle dark mode.
 
         This method will raise a ValueError if dark mode is set to auto.
@@ -39,10 +43,12 @@ class DarkMode(ValueElement[bool | None], component='dark_mode.js'):
         if self.value is None:
             raise ValueError('Cannot toggle dark mode when it is set to auto.')
         self.value = not self.value
+        return self
 
-    def auto(self) -> None:
+    def auto(self) -> Self:
         """Set dark mode to auto.
 
         This will use the client's system preference.
         """
         self.value = None
+        return self
--- a/nicegui/elements/dialog.py
+++ b/nicegui/elements/dialog.py
@@ -2,6 +2,8 @@ import asyncio
 import weakref
 from typing import Any
 
+from typing_extensions import Self
+
 from ..context import context
 from ..defaults import DEFAULT_PROPS, resolve_defaults
 from ..element import Element
@@ -44,13 +46,15 @@ class Dialog(ValueElement[bool], component='dialog.js'):
             self._submitted = asyncio.Event()
         return self._submitted
 
-    def open(self) -> None:
+    def open(self) -> Self:
         """Open the dialog."""
         self.value = True
+        return self
 
-    def close(self) -> None:
+    def close(self) -> Self:
         """Close the dialog."""
         self.value = False
+        return self
 
     def __await__(self):
         self._result = None
--- a/nicegui/elements/dialog.py
+++ b/nicegui/elements/dialog.py
@@ -7,10 +7,11 @@ from typing_extensions import Self
 from ..context import context
 from ..defaults import DEFAULT_PROPS, resolve_defaults
 from ..element import Element
+from ..helpers import NoImplicitAwait
 from .mixins.value_element import ValueElement
 
 
-class Dialog(ValueElement[bool], component='dialog.js'):
+class Dialog(ValueElement[bool], NoImplicitAwait, component='dialog.js'):
 
     @resolve_defaults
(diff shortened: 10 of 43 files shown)


evnchn (dismissed):
As established OK to change `None` to `Self` return type (former discussion in https://github.com/zauberzeug/nicegui/pull/5461)

Mypy would catch all instances in NiceGUI where the subclass still returned `None` in NiceGUI code. 

So the PR is good. My question is: where else can we make stuff chainable to enable the builder pattern? This would save us another PR in the future.

falkoschindler:
Hm, you're right: Now that we made `set_enable` and `set_visibility` chainable, a natural extension would be to change other `set_*` methods as well. And `enable` and `disable` are similar to `open` and `close` etc. Here's a full list collected by Claude:

---

Here's the inventory, grouped by how strong the case for chaining is:

### Tier 1 — Setters on shared mixins (parallel to `set_visibility` / `set_enabled`)
These are the most consistent extension. Same pattern, same backward-compat story.

- `set_text` (text_element.py:87)
- `set_value` (value_element.py:124)
- `set_source` (source_element.py:94)
- `set_content` (content_element.py:88)
- `set_label` (label_element.py:87)
- `set_name` (name_element.py:87)
- `set_icon` (icon_element.py:87)
- `set_filter` (filter_element.py:88)
- `set_selected` (selectable_element.py:108)
- `set_background_color`, `set_text_color` (color_elements.py:36, :143)

### Tier 2 — Per-element setters (parallel to Tier 1)
- `set_source` (image.py:30, audio.py:36, video.py:36, interactive_image.py:81) — these are subclass overrides; would need to flip in parallel with Tier 1
- `set_color` (color_picker.py:32)
- `set_autocomplete` (input.py:79)
- `set_options` (choice_element.py:55)
- `set_selection`, `set_fullscreen` (table.py:382, :408)
- `set_theme`, `set_language`, `set_line_wrapping` (codemirror.py)
- `set_center`, `set_zoom` (leaflet.py:119, :125)
- `set_url`, `set_coordinates`, `set_points` (scene_objects.py)

### Tier 3 — Imperative open/close/toggle (clear inline-with-`with` use case)
Direct parallel to your dropdown_button motivating example.

- `open` / `close` / `toggle` on dialog.py, menu.py, button_dropdown.py:63-73, expansion.py:37, fab.py:40-48, context_menu.py:17-21
- `show` / `hide` / `toggle` on header.py:50-58, footer.py:45-53, drawer.py:65-76
- `toggle` on dark_mode.py:34, fullscreen.py:56

`with ui.dialog().open(): ...` would directly mirror the issue-#6014 pattern.

### Tier 4 — Refresh / reset / update (chainability is plausible but less compelling)
- `update` (element.py:413, and overrides in pyplot.py:101, choice_element.py:49, plotly.py:42)
- `refresh` (sub_pages.py:66)
- `reset` (upload.py:119, slide_item.py:132)
- `move` (leaflet_layers.py:87)
- `dismiss` (notification.py:196)

### Tier 5 — Probably should stay `None`
- `delete`, `remove` on element.py:509,523 — destructive on the receiver/parent, chaining is misleading
- `set_optional`, `set_bool` on props.py:67,74 — internal helper class, not user-facing
- `set_always_on_top`, `set_title`, `set_window_size` in native.py — internal pywebview wrappers
- Activate/deactivate/cancel on Timer — already a deliberate non-chainable design (note the `set_visibility` `NotImplementedError`)

### Recommendation

If we want to do this in one sweep, I'd merge **Tier 1** into the current PR (it's all in `nicegui/elements/mixins/`, ~10 small edits, fully backward compatible, completes the mixin story). Tier 2 piggybacks naturally if those subclasses override Tier 1 methods. Tier 3 is a separate, also-valuable PR with the strongest "with-statement" motivation. Tier 4 and 5 I'd leave alone unless someone asks.

---

It feels a bit much, but on the other hand it would only be consistent and wouldn't hurt to include tier 1 to 3 in this PR. What do you think?

falkoschindler:
Going for it: extending this PR to cover tiers 1–3, plus a slice of tier 4 (`marker.move()` for parity with `draggable()`, and `upload.reset()` / `slide_item.reset()`).

Skipping anything that destroys or removes the receiver — `delete`, `remove`, `notification.dismiss` — since chaining further calls on a gone element would be misleading.

Second motivation that pushed me over: composing multiple actions inside a `lambda` (`on_click=lambda: marker.move(lat, lng).draggable(False)`).

falkoschindler:
Pushed 5 follow-up commits implementing the broader plan:

**Tier 1 — mixin setters → `Self`**
`set_text`, `set_value`, `set_source`, `set_content`, `set_label`, `set_name`, `set_icon`, `set_filter`, `set_selected`, `set_background_color`, `set_text_color`. Also flipped the matching `set_source` overrides on `Image`, `Audio`, `Video`, `InteractiveImage`.

**Tier 2 — per-element setters → `Self`**
`set_color`, `set_autocomplete`, `set_options`, `set_selection`, `set_fullscreen`, `set_theme`, `set_language`, `set_line_wrapping`, `set_center`, `set_zoom`, `set_url`, `set_coordinates`, `set_points`.

**Tier 3 — open/close/toggle/show/hide → `Self`**
`menu`, `button_dropdown`, `expansion`, `fab`, `context_menu`, `header`, `footer`, `drawer`, `dark_mode`, `fullscreen`, `table.toggle_fullscreen`. (`Dialog` excluded — see below.)

**Tier 4 partial → `Self`**
`marker.move` (for parity with the already-chainable `draggable()`), `upload.reset`, `slide_item.reset`. Skipping anything that destroys the receiver (`delete`, `remove`, `notification.dismiss`).

**Dialog revert.** Initially I made `Dialog.open()`/`close()` chainable too, but `tests/test_dialog.py::test_open_close_dialog` failed: clicking *Close* silently re-opened the dialog. Root cause: `Dialog` implements `__await__` (so `await dialog` works in async code), which makes it an `Awaitable`. When `close()` returns `self`, NiceGUI's `handle_event` sees an awaitable result, schedules it as a task, the task awaits the dialog, and `Dialog.__await__` calls `self.open()` again. Reverted those two methods to return `None` so the behavior matches main.

**Possible follow-up.** The cleanest fix would be to teach `helpers.should_await()` to skip `Element` instances:

```python
def should_await(result: Any) -> TypeGuard[Awaitable[Any]]:
    return isinstance(result, Awaitable) and not isinstance(result, (Element, AwaitableResponse, asyncio.Task))
```

Today, `Dialog` is the only `Awaitable` `Element`, and its `__await__` is meant for explicit `await my_dialog` in async code — not for accidentally awaiting a `Dialog` that surfaced as the return value of a chainable mutator. The only behavior change would be that `on_click=lambda: my_dialog` (which today implicitly opens-and-waits via the magical awaitable return) becomes a no-op; the explicit `lambda: my_dialog.open()` is clearer anyway and is what the docs already show. With that one-line change, we could un-revert `Dialog.open()`/`close()` and complete Tier 3.

evnchn:
I think `on_click=lambda: my_dialog` is not the expected paradigm because `on_click` handlers are supposed to be fire-and-forget. Hence why we can ignore it. 

I'm not sure if we will footgun ourselves later if we stand by "Today, `Dialog` is the only `Awaitable` `Element`" as an assumption to make though.

falkoschindler:
@evnchn pushed the `should_await` carve-out and un-reverted `Dialog.open()`/`close()` in 5b938729. The change is just one `isinstance` line plus a comment explaining why. I think this solution pretty safe. Even if we introduce another awaitable element in the future, the lambda usage won't work. But we're not breaking existing code. Agree?

falkoschindler:
Back to the drawing board: pylint doesn't like our solution since it introduces a cyclic import via `Element` in the helper function.

falkoschindler:
Ok, I resolved the problem with a `NoImplicitAwait` mixin the is used by `Element`. Now all tests are green.

evnchn:
`NoImplicitAwait` is a really elegant solution (makes the marker explicit). I will do another round of potential footgun check to avoid any other footguns, then review.

evnchn:
*Posted by Claude Code on @evnchn's instruction — Evan and @falkoschindler will review independently.*

## TL;DR

Evan asked me to do an independent footgun check on this PR while he was AFK, then to validate three "can we also fix this" follow-ups he had in mind. All findings below are empirically verified — every footgun has an MRE, every proposed fix has tests passing. Worktree at PR HEAD `6fb776a1`.

**Audit verdict:** PR is solid. The `NoImplicitAwait` marker is the right shape and `should_await` exclusion is uniformly applied across all 14 callsites. **One medium finding** worth a docstring callout (or fix), **two consistency nits**.

**Three follow-up fixes — all small, backward-compatible, 72/72 related tests still pass:**

| # | Fix | Files | Lines | Verdict |
|---|---|---|---|---|
| 1 | Move `NoImplicitAwait` from `Element` → `Dialog` | 2 | 2 | safe — Dialog still uniformly protected; user code unaffected |
| 2 | `Element.move() -> Self` | 1 | 2 | safe — no LSP conflict (leaflet/scene/native moves are non-Element) |
| 3 | `ColorInput.open_picker() -> Self` | 1 | 3 | trivial parity nit |

Combined diff: 3 files, +9/-4 lines. Each fix is independent.

---

<details>
<summary><b>Finding 1 [medium] — User-defined awaitable Element subclasses silently lose implicit awaiting</b></summary>

`Element(Visibility, helpers.NoImplicitAwait)` makes *every* Element subclass exempt from `should_await`, not just `Dialog`. The PR description and `NoImplicitAwait.__doc__` only mention Dialog as an example, but the exemption is transitive to every user subclass.

This is a real behavior change vs. main. Today only `Dialog` is awaitable in the nicegui codebase, but a user-built confirm modal, toast queue, or other custom awaitable element would see its auto-await behavior silently disappear after this PR.

**MRE — silent behavior change vs. main:**

```python
from nicegui import ui, helpers

class CustomAwaitable(ui.element):
    def __init__(self):
        super().__init__('div')
    def __await__(self):
        # imagine a user-built confirm modal, toast queue, etc.
        if False: yield
        return None

c = CustomAwaitable()
print(helpers.should_await(c))
# main: True   →   PR: False  (silently dropped if returned from a handler)
```

The `AwaitableResponse` and `asyncio.Task` exclusions are documented in the `should_await` docstring; the *transitive Element* exclusion is mentioned only in passing.

**Two ways to address:**
- **(a) Documentation-only:** strengthen the `NoImplicitAwait` docstring to explicitly say it applies to all Element subclasses, including user-defined ones, and that's intentional.
- **(b) Code fix (proposed Fix #1 below):** move the marker from `Element` to `Dialog` so user code is unaffected. Dialog's chainable mutators still return self → still NoImplicitAwait → still protected.

Both are valid. Fix 1 is what Evan asked me to validate — it works.
</details>

<details>
<summary><b>Finding 2 [low] — <code>Element.move()</code> isn't chainable, but <code>marker.move()</code> is</b></summary>

Tier 4 added `marker.move(lat, lng) -> Self` (per `Layer.move` in leaflet) "for parity with the already-chainable `draggable()`". But the much more widely-used base `Element.move(target_container, ...)` in `nicegui/element.py:478` still returns `None`.

`Element.move` is widely used in click handlers — see `tests/test_element.py:174-208`:

```python
ui.button('Move X to A', on_click=lambda: x.move(a))
```

**MRE — chain breaks:**

```python
b = ui.button('hi')
with ui.card() as target:
    pass
b.move(target).set_visibility(True)
# AttributeError: 'NoneType' object has no attribute 'set_visibility'
```

If the goal of this PR is "completes the with-statement & lambda-composition story", flipping `Element.move` is the most prominent omission. Evan's question 2 below validates this.
</details>

<details>
<summary><b>Finding 3 [nit] — <code>ColorInput.open_picker()</code> not chainable</b></summary>

`nicegui/elements/color_input.py:49 — def open_picker(self) -> None:` is semantically a Tier 3 "open" but missed in the inventory. Trivial parity fix; see proposed Fix #3.
</details>

<details>
<summary><b>Other things checked (no findings)</b></summary>

- **Tier 4 carve-out asymmetry** (`marker.move`/`upload.reset`/`slide_item.reset` chainable; `notification.dismiss`/`Element.delete`/`Element.remove` not). The "destructive on receiver" reasoning holds — `reset` clears state without destroying the element.
- **MRO / metaclass interactions.** `Dialog → ValueElement → Element → Visibility → NoImplicitAwait → Generic → object`. NoImplicitAwait is a body-less marker, no diamond issues.
- **`should_await` callsite coverage.** All 14 sites in `nicegui/` route through the helper (events.py, event.py, client.py, app.py, page.py, sub_pages.py, refreshable.py, validation_element.py, timer.py). No bypass paths.
- **Existing `on_click=method_reference` patterns** — all behave the same as main, because either the return is non-awaitable, or it's a Dialog protected by NoImplicitAwait.
- **`Timer.set_visibility` / `Notification.set_visibility`** annotated `-> Self` but always raise. Required for LSP compliance with the parent's new `Self` return type — `f7eddcec` did this correctly.
- **Tier 1/2/3 setter overrides.** All `set_*` overrides on subclasses (Image/Audio/Video/InteractiveImage `set_source`) return `Self` consistently.
</details>

---

## Proposed fixes — empirically verified

<details>
<summary><b>Fix 1 — Move <code>NoImplicitAwait</code> from <code>Element</code> to <code>Dialog</code></b></summary>

**Diff:**

```diff
# nicegui/element.py
- class Element(Visibility, helpers.NoImplicitAwait):
+ class Element(Visibility):

# nicegui/elements/dialog.py
+ from ..helpers import NoImplicitAwait

- class Dialog(ValueElement[bool], component='dialog.js'):
+ class Dialog(ValueElement[bool], NoImplicitAwait, component='dialog.js'):
```

**Why this works:** Dialog's chainable mutators (open/close/set_value/set_visibility/classes/props/style/move) all return `self`, which is a `Dialog` instance, which is `NoImplicitAwait`. So `should_await(self)` still returns False — protection unchanged. Verified test:

```python
async def test_dialog_protection_still_works(user: User) -> None:
    @ui.page('/')
    def page() -> None:
        d = ui.dialog()
        for name, ret in [
            ('open',           d.open()),
            ('close',          d.close()),
            ('set_value',      d.set_value(True)),
            ('set_visibility', d.set_visibility(True)),
            ('classes',        d.classes('x')),
            ('props',          d.props('persistent')),
            ('style',          d.style('color: red')),
            ('move',           d.move(target_index=0)),
        ]:
            assert ret is d
            assert not helpers.should_await(ret)
    await user.open('/')
```

**Why it's better:** A user who writes their own awaitable Element subclass gets auto-await back. The marker now lives on the actual awaitable class — semantically correct.

```python
async def test_user_awaitable_element_works_again(user: User) -> None:
    class CustomAwaitable(ui.element):
        def __init__(self): super().__init__('div')
        def __await__(self):
            if False: yield
            return None
    @ui.page('/')
    def page() -> None:
        c = CustomAwaitable()
        assert not isinstance(c, helpers.NoImplicitAwait)
        assert helpers.should_await(c)  # auto-await restored
    await user.open('/')
```

**Subclassing safe:**

```python
async def test_dialog_subclass_inherits_protection(user: User) -> None:
    class MyDialog(ui.dialog):
        pass
    @ui.page('/')
    def page() -> None:
        d = MyDialog()
        assert isinstance(d, helpers.NoImplicitAwait)
        assert not helpers.should_await(d.close())
    await user.open('/')
```

**Risk:** A future contributor adding a *new* awaitable Element to nicegui (e.g., should `Drawer` ever become awaitable) must remember to mix in `NoImplicitAwait`. That's one line of discipline at the class definition site, much narrower than the current "all Elements forever, including user code" exemption.
</details>

<details>
<summary><b>Fix 2 — <code>Element.move() -> Self</code></b></summary>

**Diff:**

```diff
# nicegui/element.py
  def move(self,
           target_container: Element | None = None,
           target_index: int = -1, *,
-          target_slot: str | None = None) -> None:
+          target_slot: str | None = None) -> Self:
      ...
      target_container.update()
+     return self
```

**Why no new footgun:**

- The base `Element.move()` is the slot-relocator. After `b.move(card)`, `b` is still alive and now lives in `card`. Returning `self` is meaningful.
- `marker.move(lat, lng)` (leaflet `Layer`), `scene_object3d.move()` (Three.js `Object3D`), and `native.move()` (pywebview window) are all on **non-Element classes** — they shadow the name but aren't LSP overrides, so no signature conflict.
- Existing `lambda: x.move(a)` patterns in `tests/test_element.py:174-208` still work — lambdas discard returns.
- Invalid-slot `ValueError` still raised before `return self` is reached.

**Verified tests:**

```python
async def test_element_move_chainable(user: User) -> None:
    @ui.page('/')
    def page() -> None:
        b = ui.button('hi')
        with ui.card() as target:
            pass
        result = b.move(target).set_visibility(True).set_text('moved')
        assert result is b
        assert b.parent_slot is target.default_slot
    await user.open('/')

async def test_element_move_invalid_slot_still_raises(user: User) -> None:
    @ui.page('/')
    def page() -> None:
        b = ui.button('hi')
        with ui.card() as target:
            pass
        try:
            b.move(target, target_slot='nonexistent')
        except ValueError:
            return
        raise AssertionError('expected ValueError')
    await user.open('/')

async def test_regression_move_return_value_discarded(user: User) -> None:
    """tests/test_element.py:174-208 patterns still work."""
    log: list[str] = []
    @ui.page('/')
    def page() -> None:
        with ui.card() as target:
            pass
        b = ui.button('hi')
        f = lambda: (log.append('moved'), b.move(target))[0]
        f()
        assert log == ['moved']
        assert b.parent_slot is target.default_slot
    await user.open('/')
```
</details>

<details>
<summary><b>Fix 3 — <code>ColorInput.open_picker() -> Self</code></b></summary>

**Diff:**

```diff
# nicegui/elements/color_input.py
+ from typing_extensions import Self
  ...
- def open_picker(self) -> None:
+ def open_picker(self) -> Self:
      """Open the color picker"""
      if self.value:
          self.picker.set_color(self.value)
      self.picker.open()
+     return self
```

```python
async def test_open_picker_chainable(user: User) -> None:
    @ui.page('/')
    def page() -> None:
        c = ui.color_input(label='pick')
        result = c.open_picker().set_visibility(True).classes('x')
        assert result is c
    await user.open('/')
```
</details>

<details>
<summary><b>How I verified — environment & test runs</b></summary>

- Worktree: `git worktree add /tmp/wt-pr6016 pr-6016` at HEAD `6fb776a1`.
- Existing venv from a separate nicegui checkout, run with `PYTHONPATH=/tmp/wt-pr6016 .venv/bin/python -m pytest …`.
- Audit MREs (5): all pass on PR HEAD — confirms findings 1 & 2.
- Fix experiment MREs (9): all pass with the three fixes applied.
- Existing related tests on the experimental branch: **72/72 pass** (`test_dialog`, `test_element`, `test_color_input`, `test_dark_mode`, `test_button`, `test_expansion`).
- Combined diff for the three fixes: 3 files, +9/-4 lines.

I did not run the full test suite (Selenium/Playwright integration) — only the headless ones above. If anything is going to surface there, it'd most likely be in `test_dialog.py` and `test_keep_alive.py` for Fix 1 (both already in my 72/72 cohort) or browser-driven `test_color_input.py` for Fix 3.
</details>

---

Happy to push the three-commit follow-up branch if either of you wants to inspect it locally — it's at `pr6016-fix-experiments` in my worktree but not pushed anywhere. Otherwise, picking-and-choosing one or two of the fixes is fine; they're independent.

evnchn:
@falkoschindler so the super-TLDR appears to be "3 fixes until golden" then. I'll check on the details later. I'm showering

falkoschindler:
@evnchn Great findings!

1. **Move `NoImplicitAwait` from `Element` to `Dialog`:**
    Somehow I though consistency across all elements is better and I couldn't come up with a reason why someone would want to implicitly await custom elements. But actually there is no harm in narrowing the `NoImplicitAwait` to dialogs and keep allowing whatever users want to do with custom elements.
2. **`Element.move() -> Self`**:
    Agreed.
3. **`ColorInput.open_picker() -> Self`**:
    Agreed.

I'll push an update shortly.

evnchn:
*Posted by Claude Code on @evnchn's instruction.*

## TL;DR

Reviewed Falko's three follow-up commits (6594e155, faabdc87, b42e30b2) against my earlier audit. **All three land exactly as specified — ready to merge from my side.**

<details>
<summary><b>Per-fix verification</b></summary>

**Fix 1 — `NoImplicitAwait` moved from `Element` to `Dialog` (6594e155)**
- `Element` no longer inherits the marker (`element.py:42`).
- `Dialog` does (`dialog.py:14`); MRO inheritance means subclasses of `Dialog` stay protected.
- Marker still exported from `nicegui/helpers/`; `should_await` exclusion intact (`helpers/functions.py:54`).
- Bonus: docstring tightened to mention Dialog explicitly (`helpers/functions.py:51-53`).
- Awaitable-element census: `Dialog.__await__` is the only one in `nicegui/elements/`. Other `__await__` sites (`javascript_request.py`, `event.py`, `awaitable_response.py`, `background_tasks.py`) are non-Element infrastructure — unaffected. User-defined awaitable Element subclasses now get auto-await back, as intended.

**Fix 2 — `Element.move() -> Self` (faabdc87)**
- Signature flipped, `return self` after `target_container.update()`, `ValueError` for invalid slot still raises before the return.
- LSP check clean: `Object3D.move` (scene), `Marker.move` (leaflet `Layer` is a `@dataclass`), and `WindowProxy.move` (extends `webview.Window`) all live on non-Element classes — no shadowing flag.

**Fix 3 — `ColorInput.open_picker() -> Self` (b42e30b2)**
- One-liner; `Self` imported from `typing_extensions` per house style.

</details>

No critical or important issues found. CI is green. 👍

evnchn (approved):
PR file changed scope is exceeded the scale for a human review to be possible. I think I can approve based on https://github.com/zauberzeug/nicegui/pull/6016#issuecomment-4408436733 which is based on https://github.com/zauberzeug/nicegui/pull/6016#issuecomment-4404546599 which is based on empirical verification and a Claude Code that is not IQ-impaired from suspected recent incidents.

PR #6017, opened 2026-04-29
Improve Sortable documentation

falkoschindler:
### Motivation

The `make_sortable()` API added in 3.11.0 had a few rough edges in its documentation:

- No "Added in version ..." hint, unlike most other recently added APIs.
- No mention of the direct-children-only scoping, which is easy to overlook (the [Cross-Container demo](https://nicegui.io/documentation/sortable) silently works around it by wrapping items in an inner `ui.column`, but never spells the rule out).
- No API reference tables on the documentation page, so the `Sortable` controller's properties (`animation`, `handle`, `group`, `ghost_class`) and methods (`enable`, `disable`, `on_end`) were undiscoverable from the docs.

### Implementation

- Add a "Note: Only direct children of the container are sortable; nested containers can be made sortable independently." line to the intro demo of the Sortable documentation page.
- Add an `*Added in version 3.11.0*` hint to both the website demo intro and the `make_sortable()` docstring (so it surfaces in IDE tooltips too).
- Add `doc.reference(SortableElement, ...)` and `doc.reference(Sortable, title='Reference for Sortable controller')` at the bottom of the documentation page.

### Result

Screenshot 2026-04-29 at 17 30 30

### Progress

### DIFF
--- a/nicegui/elements/mixins/sortable_element.py
+++ b/nicegui/elements/mixins/sortable_element.py
@@ -33,6 +33,8 @@ class SortableElement(Element):
         Note: Setting a custom HTML ID on the container (e.g. via ``.props('id="my-list"')``) is not supported
         and will break the internal slot synchronization.
 
+        *Added in version 3.11.0*
+
         :param options: dictionary of raw SortableJS options (overrides named params like ``animation`` if both are given)
         :param on_end: callback invoked when a sort operation ends (fires on the source container only, even for cross-container moves)
         :param animation: animation duration in seconds (default: 0.15)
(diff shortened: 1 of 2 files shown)


denniswittich (approved):
Looks sound! 
As far as I can asses, this is now in sync with the most parts of the documentation.

When comparing to other elements, I noticed that `FabAction` does have an "Added in..." comment, but `Fab` does not.. is this correct?

falkoschindler:
@denniswittich Good catch! I'll fix it on main.

PR #6019, opened 2026-04-29
Fix silent page reload on every reconnect after first ack

### LINKED ISSUE #6018: Silent `window.location.reload()` on every WebSocket reconnect after the first ack
### First Check

### Description

`Outbox.try_rewind` silently reloads the browser via `run_javascript('window.location.reload()')` on every WebSocket reconnect after the client has sent its first `ack`. This is reachable on any production deployment where an upstream proxy or load balancer applies a tunnel timeout to WebSocket connections — a near-universal configuration. The reload is dispatched as a raw JS string with no preceding `console.log`, so the symptom presents to users as "the page randomly reloads itself" with nothing to grep for in DevTools.

The root cause is an interaction of three behaviors that are individually reasonable but together break reconnect:

1. The Engine.IO `query` object is frozen at app mount — `next_message_id=0` is baked into every reconnect URL for the lifetime of the page.
2. The browser sends an `ack` every 3 seconds with its current next-expected message ID; the server's ack handler prunes everything below that from `Outbox.message_history`.
3. After ~3 seconds of activity, message ID `0` is no longer in history. On the next reconnect the server receives `next_message_id=0` from the (frozen) query, calls `Outbox.try_rewind(0)`, fails to find ID `0`, and falls through to a hard reload.

So: any reconnect on a page that's been alive for more than ~3 seconds and that has emitted any messages will trigger a silent reload. Users lose scroll position, open dialogs, in-flight form input, and any client-side state. The reconnect machinery effectively never recovers in-place — every WS drop becomes a hard reload.

## Reproducer

```python
from nicegui import ui

counter = 0
label = ui.label('0')

def tick():
    global counter
    counter += 1
    label.text = str(counter)

ui.timer(2.0, tick)
ui.run()
```

Steps:

1. Start the app and open it in the browser. Open DevTools → Network, enable "Preserve log".
2. Wait ~10 seconds. The label is updating; the browser has acked a few times.
3. In DevTools → Network, set throttling to **Offline** for ~5 seconds, then back to **Online**. (Equivalent to any transient proxy-side WS drop.)
4. Observe: the browser fetches the document URL again and the page hard-reloads. No `console.log` is printed before the reload. The address bar URL re-loads as if the user hit F5.

Expected: the WS reconnects in place; missed messages are either resent (if recent) or silently skipped, and the user keeps their page state.

Actual: hard reload via `window.location.reload()`.

### 1. The reconnect query is frozen at app mount

In [`static/nicegui.js`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/static/nicegui.js) (≈ line 412):

```js
window.socket = io(url, {
  path: `${options.prefix}/_nicegui_ws/socket.io`,
  query: options.query,
  ...
});
```

`options.query.next_message_id` is set once at mount and never updated. socket.io passes the same `query` object on every Engine.IO auto-reconnect. The browser does maintain `window.nextMessageId` correctly for incoming messages and for the periodic `ack`, but `options.query` is a separate object and is never reconciled with `window.nextMessageId`.

You can confirm this by inspecting the URL of any WebSocket reconnect after a transient drop — `next_message_id=0` is always in the query string, regardless of how many messages the page has actually received.

### 2. Acks aggressively prune history

In [`nicegui.py`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/nicegui.py) the ack handler:

```python
@sio.on('ack')
def _on_ack(_: str, msg: dict) -> None:
    if client := Client.instances.get(msg['client_id']):
        client.outbox.prune_history(msg['next_message_id'])
```

And [`outbox.py`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/outbox.py):

```python
def prune_history(self, next_message_id: MessageId) -> None:
    """Prune the message history up to the given message ID."""
    while self.message_history and self.message_history[0][0] < next_message_id:
        self.message_history.popleft()
```

The browser sends an `ack` every 3 seconds with its current `nextMessageId`. After the very first ack for `nextMessageId >= 1`, message ID `0` is gone from history.

This is a desirable optimization in steady state — but it interacts catastrophically with (1).

### 3. `try_rewind(0)` falls through to a hard reload

[`outbox.py`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/outbox.py):

```python
def try_rewind(self, target_message_id: MessageId) -> None:
    """Rewind to the given message ID and discard all messages before it."""
    if self.next_message_id == target_message_id:
        return

    while self.message_history:
        self.next_message_id, _, message = self.message_history.pop()
        self.messages.appendleft(message)
        if self.next_message_id == target_message_id:
            self.message_history.clear()
            self._set_enqueue_event()
            return

    # target message ID not found, reload the page
    self.client.run_javascript('window.location.reload()')
```

On reconnect, `handle_handshake` calls `outbox.try_rewind(next_message_id)` with the (stale) `next_message_id=0` from the query. The server's `self.next_message_id` is non-zero (messages have been emitted), and `0` is no longer in `message_history` (pruned by the ack), so the loop drains history without finding `0` and the final line fires.

`run_javascript` dispatches a raw JS string to the client, which is `eval`ed without a wrapping `console.log`. **That's why this reload doesn't show up alongside the four documented `console.log`-prefixed reload paths in `nicegui.js`** — it bypasses them entirely.

## Impact

This affects any NiceGUI deployment behind a proxy with a finite WebSocket tunnel timeout, which is the default for nearly every production-grade proxy:

- nginx `proxy_read_timeout` (default 60s, commonly raised to 300s)
- AWS ALB `idle_timeout.timeout_seconds` (default 60s)
- Cloudflare WS timeout (~100s on most plans)
- Scaleway LB `timeout_tunnel` (default 300s)
- Traefik / Caddy / HAProxy similar

In other words: a NiceGUI app running anywhere except localhost will probably hit this within a few minutes of a session, on every disconnect, for the lifetime of the page. Each hit is a silent hard reload that drops the user's UI state.

We diagnosed this only after exhausting the four documented reload paths in `nicegui.js` (`handshake failed`, `connection timed out`, `implicit handshake failed`, `reconnect requested`). Each of those logs an identifying `console.log`. The fact that none of them appeared in our DevTools — yet the page kept reloading — was what eventually led us to `Outbox.try_rewind`.

## Suggested fixes

Any one of these resolves the bug; the first is probably the cleanest.

**Option A — Update the socket.io query before reconnect.** In `nicegui.js`, before each Engine.IO reconnect attempt, write the current `window.nextMessageId` into `options.query.next_message_id` (and into the socket's reconnection options, depending on socket.io version). This makes the reconnect URL accurately reflect the client's position so `try_rewind` has a real target to find.

```js
window.socket.io.on('reconnect_attempt', () => {
  options.query.next_message_id = window.nextMessageId;
});
```

(Exact spelling depends on the socket.io-client version in use; the principle is "update the query right before reconnection".)

**Option B — Don't reload on rewind miss; degrade gracefully.** Replace the final line of `try_rewind` with a debug log. Missed messages between the last ack and the reconnect will be lost, but this is the same trade-off most live apps already make on a network blip; the next state update will repaint affected widgets. A hard reload is a far worse outcome than a brief stale render.

```python
# target message ID not found, fall through; next emit will resync.
log.debug('try_rewind: target %s not in history; skipping rewind.', target_message_id)
```

**Option C — Only prune below the lowest unacked ID across all surviving sessions.** A more conservative ack-pruning policy that keeps message `0` available until *no* path could ever ask for it. Heavier to implement; probably overkill given Options A and B.

We'd be happy to send a PR for Option A or B if a maintainer can sanity-check the approach.

## Environment

- NiceGUI: observed on 3.10.0 (local) and 3.11.1 (deployed). The `try_rewind` reload line is present in both.
- Python: 3.13
- Browsers tested: Chromium-based (Chrome, Edge); same behavior expected on Firefox / Safari since the bug is server-side and the client just executes the JS string sent over the WS.
- Deployment: Scaleway-hosted container behind both a Scaleway LB and Cloudflare. Reproduced on the direct Scaleway path with Cloudflare bypassed, ruling out any Cloudflare-specific edge behavior.

## Workaround

Until this is fixed upstream, monkey-patching `Outbox.try_rewind` to swallow the miss instead of reloading is the simplest in-app mitigation:

```python
from nicegui.outbox import Outbox

def try_rewind_no_reload(self, target_message_id):
    if self.next_message_id == target_message_id:
        return
    while self.message_history:
        self.next_message_id, _, message = self.message_history.pop()
        self.messages.appendleft(message)
        if self.next_message_id == target_message_id:
            self.message_history.clear()
            self._set_enqueue_event()
            return
    # Was: self.client.run_javascript('window.location.reload()')
    # Now: silently skip; next state update will repaint stale widgets.

Outbox.try_rewind = try_rewind_no_reload
```

Apply once at app startup before `ui.run(...)`.

### NiceGUI Version

3.10.0

### Python Version

3.13.13

### Browser

Chrome

### Operating System

Linux

### Additional Context

_No response_

arodidev:
### Motivation

Every WebSocket auto-reconnect after the first periodic ack triggers a silent `window.location.reload()`, dropping the user's scroll position, open dialogs, in-flight form input, and other client-side state.
This affects any deployment behind a proxy with a finite WebSocket tunnel timeout (nginx `proxy_read_timeout`, AWS ALB `idle_timeout`, Cloudflare's WS timeout, Scaleway LB `timeout_tunnel`, etc.) — i.e. nearly every production NiceGUI app.

The root cause is a three-way interaction:

1. The socket.io `query` object is frozen at app mount in `nicegui.js`: `next_message_id=0` is baked into every reconnect URL and handshake payload for the lifetime of the page.
2. The browser sends an `ack` every 3s with its current next-expected ID; the server's ack handler prunes everything below that from `Outbox.message_history`.
3. After the first ack, message ID `0` is no longer in history.
   On the next reconnect the server receives `next_message_id=0` from the (frozen) query, calls `Outbox.try_rewind(0)`, fails to find ID `0`, and falls through to `self.client.run_javascript('window.location.reload()')`.

### Implementation

Update `options.query.next_message_id` from `window.nextMessageId` on each socket.io `reconnect_attempt` so the (frozen) initial value of `0` is not sent in the reconnect URL and handshake payload. Without this, once the periodic ack pruned message `0` from `Outbox.message_history`, every auto-reconnect would fall through `Outbox.try_rewind` to `window.location.reload()` — losing scroll position, open dialogs, and in-flight form state on every transient WS drop behind any proxy with a finite tunnel timeout.

### Progress

- [x] Pytests are added.

### DIFF
--- a/nicegui/outbox.py
+++ b/nicegui/outbox.py
@@ -170,7 +170,7 @@ class Outbox:
                 return
 
         # target message ID not found, reload the page
-        self.client.run_javascript('window.location.reload()')
+        self.client.run_javascript('console.log("reloading because outbox rewind failed"); window.location.reload()')
 
     def prune_history(self, next_message_id: MessageId) -> None:
         """Prune the message history up to the given message ID."""
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -418,6 +418,9 @@ function createApp(elements, options) {
             ? ["polling", ...options.transports]
             : options.transports,
       });
+      window.socket.io.on("reconnect_attempt", () => {
+        options.query.next_message_id = window.nextMessageId;
+      });
       window.did_handshake = false;
       const messageHandlers = {
         connect: () => {
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -419,6 +419,7 @@ function createApp(elements, options) {
             : options.transports,
       });
       window.socket.io.on("reconnect_attempt", () => {
+        // keep the otherwise-frozen reconnect query in sync to avoid triggering a reload (see #6019)
         options.query.next_message_id = window.nextMessageId;
       });
       window.did_handshake = false;
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -6,7 +6,6 @@ import httpx
 import pytest
 from fastapi import HTTPException
 from fastapi.responses import PlainTextResponse
-from selenium.webdriver.common.by import By
 from starlette.exceptions import HTTPException as StarletteHTTPException
 
 from nicegui import app, background_tasks, ui
@@ -314,57 +313,6 @@ def test_warning_about_to_late_responses(screen: Screen):
     screen.open('/')
     screen.should_contain('NiceGUI page')
     screen.assert_py_logger('ERROR', re.compile('it was returned after the HTML had been delivered to the client'))
-
-
-def test_reconnecting_without_page_reload(screen: Screen):
-    @ui.page('/', reconnect_timeout=3.0)
-    def page():
-        ui.input('Input').props('autofocus')
-        ui.button('drop connection', on_click=lambda: ui.run_javascript('socket.io.engine.close()'))
-
-    screen.open('/')
-    screen.type('hello')
-    screen.click('drop connection')
-    screen.wait(2.0)
-    element = screen.selenium.find_element(By.XPATH, '//*[@aria-label="Input"]')
-    assert element.get_attribute('value') == 'hello', 'input should be preserved after reconnect (i.e. no page reload)'
-
-
-def test_reconnecting_after_ack_does_not_reload(screen: Screen):
-    @ui.page('/', reconnect_timeout=3.0)
-    def page():
-        ui.input('Input').props('autofocus')
-        label = ui.label('0')
-        ui.timer(0.2, lambda: label.set_text(str(int(label.text) + 1)))
-
-    screen.open('/')
-    screen.type('hello')
-    initial_document_id = screen.selenium.execute_script('return window.documentId;')
-
-    screen.wait(4.0)  # > 3s for the periodic ack to fire and prune message 0 server-side
-    assert screen.selenium.execute_script('return Number(window.nextMessageId);') > 0
-
-    screen.selenium.execute_script('window.socket.io.engine.transport.onClose("transport close");')  # drop connection
-    screen.wait(4.0)
-    assert screen.selenium.execute_script('return window.socket.connected;')
-    assert screen.selenium.execute_script('return window.documentId;') == initial_document_id
-    assert screen.selenium.find_element(By.XPATH, '//*[@aria-label="Input"]').get_attribute('value') == 'hello'
-
-
-def test_reconnect_attempt_refreshes_query_next_message_id(screen: Screen):
-    @ui.page('/', reconnect_timeout=3.0)
-    def page():
-        label = ui.label('0')
-        ui.timer(0.2, lambda: label.set_text(str(int(label.text) + 1)))
-
-    screen.open('/')
-    screen.wait(1.5)
-    assert screen.selenium.execute_script('return Number(window.socket.io.opts.query.next_message_id);') == 0
-    assert screen.selenium.execute_script('return Number(window.nextMessageId);') > 0
-
-    screen.selenium.execute_script('window.socket.io.engine.transport.onClose("transport close");')
-    screen.wait(2.0)
-    assert screen.selenium.execute_script('return Number(window.socket.io.opts.query.next_message_id);') > 0
 
 
 def test_ip(screen: Screen):
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -348,7 +348,7 @@ def test_reconnecting_after_ack_does_not_reload(screen: Screen):
 
     screen.open('/')
     screen.type('hello')
-    screen.wait(4.0)
+    screen.wait(5.0)  # > 3s for the periodic ack to fire and prune message 0 server-side
     initial_document_id = screen.selenium.execute_script('return window.documentId;')
     assert screen.selenium.execute_script('return Number(window.nextMessageId);') > 0
     screen.click('drop connection')
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -333,27 +333,22 @@ def test_reconnecting_without_page_reload(screen: Screen):
 def test_reconnecting_after_ack_does_not_reload(screen: Screen):
     @ui.page('/', reconnect_timeout=3.0)
     def page():
+        ui.input('Input').props('autofocus')
         label = ui.label('0')
         ui.timer(0.2, lambda: label.set_text(str(int(label.text) + 1)))
 
-        ui.input('Input').props('autofocus')
-        ui.button('drop connection', on_click=lambda: ui.run_javascript(
-            'socket.io.engine.transport.onClose("transport close")'
-        ))
-
     screen.open('/')
     screen.type('hello')
-    screen.wait(5.0)  # > 3s for the periodic ack to fire and prune message 0 server-side
     initial_document_id = screen.selenium.execute_script('return window.documentId;')
+
+    screen.wait(4.0)  # > 3s for the periodic ack to fire and prune message 0 server-side
     assert screen.selenium.execute_script('return Number(window.nextMessageId);') > 0
-    screen.click('drop connection')
+
+    screen.selenium.execute_script('window.socket.io.engine.transport.onClose("transport close");')  # drop connection
     screen.wait(4.0)
     assert screen.selenium.execute_script('return window.socket.connected;')
-    document_id_after = screen.selenium.execute_script('return window.documentId;')
-    assert document_id_after == initial_document_id, \
-        'window.documentId must persist (no hard reload) after auto-reconnect that follows an ack'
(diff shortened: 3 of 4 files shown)


evnchn:
> _Posting on behalf of @evnchn — this is Claude Code reporting test results he asked me to run on his MacBook (.179, headless Chrome via Selenium Manager)._

**TL;DR — both tests in this PR catch the regression on `main`. Confirmed.**

| Ref                                       | `test_reconnecting_after_ack_does_not_reload` | `test_reconnect_attempt_refreshes_query_next_message_id` |
| ----------------------------------------- | --------------------------------------------- | -------------------------------------------------------- |
| **This PR** (`73b1ec99`)                  | ✅ PASS                                        | ✅ PASS                                                   |
| **`main`** (`c4b5ab60`, pre-fix)          | ❌ FAIL                                        | ❌ FAIL                                                   |
| **#5673 head** (`04a7ae4f`, where the implicit-handshake form that ships in `main` actually entered) | ❌ FAIL | ❌ FAIL |

So each test fails on every buggy ancestor we'd care about and passes on the fix. No incidental green; no incidental red.

<details>
<summary>Methodology + raw output</summary>

### Setup
- macOS 15.7.1 (arm64), Chrome via Selenium Manager, NiceGUI repo at `pr-6019` / `upstream/main` / `pr-5673` checked out from `https://github.com/zauberzeug/nicegui.git`.
- Test file copied verbatim from this PR's `tests/test_page.py` additions into a standalone `tests/test_pr6019_fix.py` so the same test bodies run against all three refs (avoiding any drift in the surrounding test file).
- Run with `uv run pytest tests/test_pr6019_fix.py -v --tb=short` after `uv sync` per ref.

### Failure signatures on the buggy refs

Both buggy refs fail with identical signatures — exactly what the PR description predicts:

1. **`test_reconnecting_after_ack_does_not_reload`** — `window.documentId` mismatches before/after the simulated reconnect, proving a hard reload fired (the silent `run_javascript('window.location.reload()')` in `Outbox.try_rewind`).
2. **`test_reconnect_attempt_refreshes_query_next_message_id`** — `window.socket.io.opts.query.next_message_id` stays at `0` after the transport drop, proving the frozen-query bug directly without needing the downstream reload to manifest.

### Results

**On this PR (`73b1ec99`)** — both pass:
```
tests/test_pr6019_fix.py::test_reconnecting_after_ack_does_not_reload PASSED [ 50%]
tests/test_pr6019_fix.py::test_reconnect_attempt_refreshes_query_next_message_id PASSED [100%]
============================== 2 passed in 20.17s ==============================
```

**On `main` (`c4b5ab60`)** — both fail:
```
E   AssertionError: window.documentId must persist (no hard reload) after auto-reconnect that follows an ack
E   assert '0a947933-74b...' == '5c8f594e-0c5...'
...
E   AssertionError: options.query.next_message_id should be refreshed on reconnect (got 0)
E   assert 0 > 0
============================== 2 failed in 14.25s ==============================
```

**On #5673 head (`04a7ae4f`)** — both fail with the same signatures:
```
E   AssertionError: window.documentId must persist (no hard reload) after auto-reconnect that follows an ack
E   assert '99f3f512-d40...' == 'a16a8fa6-ff1...'
...
E   AssertionError: options.query.next_message_id should be refreshed on reconnect (got 0)
E   assert 0 > 0
============================== 2 failed in 18.59s ==============================
```

### What this means for review

- The first test is the user-visible regression test (state-loss reload).
- The second test is a much cheaper sentinel for the same root cause — it pins the `options.query.next_message_id` ⇄ `window.nextMessageId` contract directly and runs in ~1s on the happy path. If anyone refactors the reconnect path again without updating the query, this catches it before the bigger reconnect e2e test does.
- The three-ref matrix above shows the tests track the actual bug, not a coincidence of the surrounding code.

</details>

falkoschindler (approved):
Thanks for the sharp diagnosis and swift fix, @arodidev! The three-way interaction you traced (frozen socket.io query × ack-driven history pruning × `try_rewind` fall-through to reload) was a great write-up — really easy to verify.

A few small follow-ups landed on top before merge:

- Added a one-line comment at the `reconnect_attempt` handler so the next reader knows why the query is being mutated.
- Made the previously-silent reload in `Outbox.try_rewind` greppable: it now does a `console.log("reloading because outbox rewind failed")` before `window.location.reload()`, matching the other four reload paths in `nicegui.js`.
- Trimmed boilerplate in the new tests (dropped the unused state dict + helper button; the connection drop is now triggered directly via `selenium.execute_script`).
- Moved the reconnect tests (yours plus the existing `test_reconnecting_without_page_reload`) into a dedicated `test_reconnect.py`.

Thanks again — looking forward to your next contribution!

PR #10, opened 2022-01-31
Upgrade Tailwind CSS and removed uncalled files

florianwittkamp:
### Upgrade Tailwind CSS
I'm currently using nicegui for a small private smart home project (small gui for my lights in the guest room). During developing at later hours I recognized, that the current bundled Tailwind version is outdated and does not support dark mode pattern.

I would suggest to update the bundled Tailwind version to a current one. I managed to use a updated version by adding the [Play CDN](https://tailwindcss.com/docs/installation/play-cdn) installation. 

In my case the switch options.no_internet seems to always evaluate to True, therefore, I had to add the play cdn installation also to the else condition in the quasar.html file. However, for this commit I assumed the switch will work on other systems/installations. 

Moreover I think some packed tailwind files are not used in the code overall, I removed the corresponding files. 

I was not able to properly update the bundled tailwind css files, thus, I left them as is.

PR #6021, opened 2026-04-30
Add llms.md — LLM reference for AI-assisted NiceGUI development

joko-zauberzeug:
## Motivation

AI assistants (Claude, Copilot, Cursor, etc.) working on NiceGUI projects repeatedly make the same mistakes: reaching for raw CSS, Quasar props, or JavaScript when a clean Python API exists; using `asyncio.create_task()` instead of `background_tasks.create()`; treating module-level variables as per-user state; or rebuilding entire UI trees instead of updating elements in place.

These mistakes happen not because the API docs are lacking, but because LLMs have no compact, opinionated reference that explains both *what* to use and — more importantly — *why*. Without understanding the reasoning behind NiceGUI's design decisions, an LLM will keep producing plausible-looking but non-idiomatic code.

This file was developed iteratively: first as a pure API surface reference, then extended with a "Mental Models" section after a real-world test. It was used as context for an AI-assisted code review of [NiceTransfer](https://github.com/joko-zauberzeug/nicetransfer), a NiceGUI-based file transfer application. The review identified and corrected several anti-patterns — direct JS clipboard access instead of `ui.clipboard.write()`, `asyncio.ensure_future()` usage, raw HTML download links instead of `ui.button().props('tag=a')` — and the file was refined based on exactly those findings.

## Implementation

A single `llms.md` at the repository root. It is:

- **Self-contained** — no external references needed; can be dropped into any NiceGUI project and referenced from `CLAUDE.md` or similar AI context files
- **Python-first** — every section and every anti-pattern example reinforces that NiceGUI is a Python abstraction, not a thin wrapper over Quasar/CSS/JS
- **"Why"-oriented** — 10 mental models explain the design decisions that matter most: no virtual DOM, shared event loop, pull-based bindings, slot stack mechanics, per-user vs. global state, the Outbox batching model, and more
- **Actionable** — do/don't comparison tables, anti-pattern blocks with correct alternatives, and concrete code examples throughout
- **LLM-agnostic** — works as context for Claude Code, Copilot, Cursor, or any other AI assistant

**Two intended use cases:**

1. **Contributors working on NiceGUI itself** — Falko, if useful: adding `@llms.md` to the repo's `CLAUDE.md` would make Claude Code load it automatically for every session. Happy to add that in a follow-up or here if you prefer.

2. **Developers building NiceGUI projects** — they can reference the file directly from their own project's AI context file:
   ```
   @https://raw.githubusercontent.com/zauberzeug/nicegui/main/llms.md
   ```
   This is already working well in practice for NiceTransfer.

## Progress

- [x] Full API surface documentation (all `ui.*` elements, styling system, binding, events, storage, async patterns, pages & routing)
- [x] Mental Models section explaining the design rationale behind NiceGUI's key decisions
- [x] Anti-patterns with real-world examples sourced from NiceTransfer code review
- [x] Iteratively refined based on actual AI-assisted code review results
- [x] Tested: integration as AI context in a NiceGUI project led to measurably more idiomatic code and caught patterns that would otherwise have required manual review

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ The whole content of [https://nicegui.io](https://nicegui.io) is [implemented wi
 and can be started locally with `docker run -p 8080:8080 zauberzeug/nicegui` or by executing `main.py` from this repository.
 
 You may also have a look at our [in-depth examples](https://github.com/zauberzeug/nicegui/tree/main/examples) of what you can do with NiceGUI.
-In our wiki we have a list of great [NiceGUI projects from the community](https://github.com/zauberzeug/nicegui/wiki#community-projects), a section with [Tutorials](https://github.com/zauberzeug/nicegui/wiki#tutorials), a growing list of [FAQs](https://github.com/zauberzeug/nicegui/wiki/FAQs) and [some strategies for using ChatGPT / LLMs to get help about NiceGUI](https://github.com/zauberzeug/nicegui/wiki#chatgpt).
+In our wiki we have a list of great [NiceGUI projects from the community](https://github.com/zauberzeug/nicegui/wiki#community-projects), a section with [Tutorials](https://github.com/zauberzeug/nicegui/wiki#tutorials) and a growing list of [FAQs](https://github.com/zauberzeug/nicegui/wiki/FAQs).
 When building NiceGUI projects with AI assistants, point them at [`nicegui/llms.md`](nicegui/llms.md) — a concise reference that ships with the package (also served at <https://nicegui.io/llms.txt>).
 
 ## Why?
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ and can be started locally with `docker run -p 8080:8080 zauberzeug/nicegui` or 
 
 You may also have a look at our [in-depth examples](https://github.com/zauberzeug/nicegui/tree/main/examples) of what you can do with NiceGUI.
 In our wiki we have a list of great [NiceGUI projects from the community](https://github.com/zauberzeug/nicegui/wiki#community-projects), a section with [Tutorials](https://github.com/zauberzeug/nicegui/wiki#tutorials), a growing list of [FAQs](https://github.com/zauberzeug/nicegui/wiki/FAQs) and [some strategies for using ChatGPT / LLMs to get help about NiceGUI](https://github.com/zauberzeug/nicegui/wiki#chatgpt).
+When building NiceGUI projects with AI assistants, point them at [`nicegui/llms.md`](nicegui/llms.md) — a concise reference that ships with the package (also served at <https://nicegui.io/llms.txt>).
 
 ## Why?
 
--- a/llms.md
+++ b/llms.md
@@ -1,0 +1,1210 @@
+# NiceGUI – LLM Reference
+
+> Concise, opinionated guide for AI assistants working with NiceGUI.
+> The single most important rule: **everything is Python**. Reach for `ui.*` before CSS, Quasar, or JavaScript.
+
+---
+
+## Architecture
+
+- **Backend-first**: All UI logic lives in Python. The framework handles web details automatically.
+- **Stack**: Python/FastAPI backend → WebSocket (socket.io) → Vue 3 / Quasar frontend
+- **Single worker**: One uvicorn worker with full async support; no multiprocessing needed.
+- **Outbox**: UI updates are accumulated and sent in batches to the client over the persistent WebSocket.
+- **Element lifecycle**: Python `Element` objects mirror Vue components on the client. Creating an element registers it immediately; calling `.update()` pushes changes.
+
+---
+
+## Mental Models — The "Why" Behind NiceGUI
+
+Understanding these mental models prevents the most common LLM mistakes. They explain *why* the API is shaped the way it is.
+
+### 1. There is no virtual DOM — understand what "rebuilding" costs
+
+React and similar frameworks diff a virtual DOM and patch only what changed. NiceGUI has no such diffing. When you delete and recreate elements, NiceGUI literally destroys Vue components on the client and creates new ones. This means:
+
+- Rebuilding a subtree loses focus, scroll position, animation state, and any client-side state
+- Use `@ui.refreshable` for sections that need to update — it handles clearing and rebuilding safely
+- For simple value changes, update **in place** via bindings or `.set_text()` / `.set_value()` instead of recreating elements
+
+```python
+# BAD: destroys and recreates on every data change
+def show_count():
+    container.clear()
+    ui.label(f'Count: {count}')
+
+# GOOD: updates the existing element in place
+label = ui.label('Count: 0')
+# then: label.set_text(f'Count: {count}') or bind_text_from
+```
+
+### 2. Module-level state is shared across ALL users
+
+NiceGUI runs as a single Python process. Module-level variables are shared between every connected browser tab and every user. This is the most common source of subtle multi-user bugs:
+
+```python
+# BUG: all users see the same data, last writer wins
+items = []
+
+@ui.page('/')
+def index():
+    ui.button('Add', on_click=lambda: items.append('x'))  # shared!
+    ui.label(str(items))
+
+# CORRECT: per-user state in app.storage or local variables inside @ui.page
+@ui.page('/')
+def index():
+    items = []  # local to this page invocation = per user
+    # or: app.storage.user['items'] for persistence across reloads
+```
+
+### 3. `@ui.page` creates a new Python scope per visitor — but only once per page load
+
+Each time a user navigates to a `@ui.page` route, the decorated function runs exactly once, building the page. It is **not** called again on state changes. All subsequent UI updates happen through:
+- Bindings (automatic)
+- `@ui.refreshable` (explicit rebuild of a subtree)
+- Direct element mutation (`.set_text()`, `.visible = False`, etc.)
+- Timers
+
+Understanding this prevents the mistake of trying to "re-run the page function" to update the UI.
+
+### 4. The slot stack — why context managers place elements automatically
+
+When you write `with ui.row(): ui.label('x')`, the label automatically becomes a child of the row. This works via a thread-local slot stack:
+
+- `with element:` pushes the element's default slot onto the stack
+- Any element created inside the `with` block is registered as a child of that slot
+- `__exit__` pops the slot off the stack
+
+Consequence: **element creation order matters**, and you cannot place an element in a different parent after creation. If you need to move elements, delete and recreate them (within a `@ui.refreshable`).
+
+```python
+row = ui.row()
+# Too late to add to row here without entering its context:
+with row:
+    ui.label('This works')
+# ui.label('This would go to page root, NOT inside row')
+```
+
+### 5. The binding system is pull-based, not push-based
+
+Bindings do **not** use Python property setters or event hooks to immediately propagate. Instead, a background loop runs every ~100ms and checks all active bindings for changes. This means:
+
+- Binding updates have up to ~100ms latency (usually imperceptible)
+- Bindings work on any Python object — no special base class required
+- You do **not** need to call `.update()` after changing a bound value; the loop handles it
+- Setting the same value twice quickly may result in only one propagation
+
+For immediate, guaranteed updates, use element methods directly (`.set_text()`, `.set_value()`).
+
+### 6. Why Python-first is a hard rule, not just style
+
+When you bypass NiceGUI's Python API and manipulate the DOM directly via JavaScript:
+- NiceGUI's Python-side model diverges from the client DOM (they are out of sync)
+- The next time NiceGUI sends an update for that element, it overwrites your JS change
+- Bindings and refreshables don't know about changes made outside the Python model
+
+Use `ui.run_javascript()` only for things that are truly impossible via the Python API (e.g., reading `window.innerWidth`). Never use it to style, show/hide, or update content.
+
+### 7. Async in NiceGUI: the event loop is shared
+
+NiceGUI runs in a single asyncio event loop. Every page, every user, every timer shares it. Blocking the loop (with `time.sleep()`, `requests.get()`, heavy CPU work) freezes the **entire application** for all users. Rules:
+
+- All I/O must be async (`httpx`, `aiofiles`, `asyncio.sleep()`)
+- CPU-heavy work must run in a thread: `await asyncio.to_thread(cpu_func)` (Python 3.9+, preferred over the deprecated `run_in_executor`)
+- `background_tasks.create()` for fire-and-forget coroutines (not `asyncio.create_task()` or `asyncio.ensure_future()` — the GC can cancel bare tasks; `ensure_future()` is also deprecated since Python 3.10)
+- Timers (`ui.timer()`) are safe: they schedule callbacks without blocking
+
+### 8. Why `ui.storage.user` and not a global dict
+
+`app.storage.user` is per-user (session-scoped) and persistent across page reloads. A module-level `dict` is per-process (shared across all users) and lost on server restart. Always ask: "Should this be per-user, per-session, or global?" and pick accordingly:
+
+| Storage | Scope | Persistence |
+|---|---|---|
+| local variable in `@ui.page` | per page load | until page reload |
+| `app.storage.client` | per browser connection | until tab closes |
+| `app.storage.tab` | per browser tab | until tab closes |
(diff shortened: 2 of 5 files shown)


evnchn:
@joko-zauberzeug I'm going to trust your work. I'd like to ask if it's effective to remind the LLM, first thing they do, to run a command (possibly a grep combined with Python-side `ui.*` enumeration) such that the LLM can figure out and read the newly added elements or features by itself?

This would make the document much more future-proof, I'd hope.

falkoschindler:
Thanks for putting this together, @joko-zauberzeug! The Mental Models in particular are a really useful framing.

Quick disclosure on methodology: I didn't read the file end-to-end myself — this started as an LLM-assisted review. I pushed back on the output and trimmed several first-pass findings that turned out wrong, so what's below has been challenged at least once. Still, flag anything that doesn't hold up.

Two bigger questions worth talking through before going into specifics: where the content should live (#1) and what file/format makes sense (#2). My read on each is below — curious what you and the rest of the team think.

#3–#10 are specific claims worth correcting wherever the content ends up living, so I noted them either way.

1. **Single source of truth**

   This file becomes a third place — alongside [the website docs](https://nicegui.io/documentation) and [AGENTS.md](AGENTS.md) / [CONTRIBUTING.md](CONTRIBUTING.md) — where API descriptions and guidance live. With three sources, drift is a matter of when, not if; the file will be correct the day it lands and noticeably stale a release or two later.

   Worth thinking about whether the "Mental Models" and "Anti-Patterns" sections — the load-bearing content here — could live in the docs site as a "Concepts" / "For AI assistants" page instead, with `llms.md` (or `nicegui.io/llms.txt`) as a thin entry point. That would also dovetail with #5889, which already makes pages markdown-addressable for LLMs.

2. **Filename and location: `llms.md` vs `llms.txt`**

   The convention proposed by Jeremy Howard (https://llmstxt.org/) is `llms.txt` (short index) plus `llms-full.txt` (long form), typically served from the docs site root. Adoption is uneven — uv exposes one at `docs.astral.sh/uv/llms.txt`, Pydantic at a versioned path under `pydantic.dev/docs/...`, while FastAPI does not publish one. No tooling auto-discovers either form today. Worth a sentence in the PR description on why `.md` and the repo root were chosen instead of (or in addition to) serving from `nicegui.io/llms.txt` — at minimum to head off a future "rename to llms.txt" PR. Not a blocker, but a deliberate decision the team should make once.

3. **The "bindings are pull-based, not push-based" claim is wrong**

   [llms.md:95-104](llms.md) (Mental Model #5):

   > Bindings do **not** use Python property setters or event hooks to immediately propagate. Instead, a background loop runs every ~100ms and checks all active bindings for changes.

   This is incorrect. NiceGUI bindings are **both push and pull**:
   - **Push**: `BindableProperty.__set__` calls `_propagate(...)` immediately on every assignment ([nicegui/binding.py:282-283](nicegui/binding.py#L282-L283)), and `bind_to` / `bind_from` propagate on registration ([nicegui/binding.py:180](nicegui/binding.py#L180)).
   - **Pull**: The `refresh_loop` ([nicegui/binding.py:65-78](nicegui/binding.py#L65-L78)) runs every `binding_refresh_interval` (default 0.1s) and re-checks `active_links` — this catches changes to plain `dict` keys and attributes that aren't `BindableProperty` descriptors.

   The doc's claim — "Binding updates have up to ~100ms latency", "you do **not** need to call `.update()` after changing a bound value; the loop handles it" — teaches the wrong mental model. Setting a `BindableProperty` propagates synchronously; only changes to non-descriptor targets rely on the refresh loop.

   ```diff
   -Bindings do **not** use Python property setters or event hooks to immediately propagate. Instead, a background loop runs every ~100ms and checks all active bindings for changes. This means:
   -
   -- Binding updates have up to ~100ms latency (usually imperceptible)
   -- Bindings work on any Python object — no special base class required
   -- You do **not** need to call `.update()` after changing a bound value; the loop handles it
   -- Setting the same value twice quickly may result in only one propagation
   +NiceGUI uses two complementary mechanisms:
   +
   +- **Push**: assigning to a `BindableProperty` (the descriptor used on element attributes like `value`, `text`, `visible`) propagates synchronously through all linked bindings.
   +- **Pull**: a refresh loop runs every `binding_refresh_interval` (default 0.1s) and re-checks `active_links` — this is what makes bindings work against plain `dict` keys and non-descriptor attributes that have no setter to hook into.
   +
   +You usually don't need to call `.update()` after a bound assignment; either the descriptor or the refresh loop handles it. Bindings work on any Python object — no special base class required.
   ```

4. **`asyncio.ensure_future()` is not deprecated**

   The doc states this twice — [llms.md:121](llms.md#L121) and [llms.md:1023-1024](llms.md#L1023):

   > `asyncio.ensure_future()` is also deprecated since Python 3.10

   It is not. The [Python docs page for `ensure_future`](https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future) does carry a "Deprecated since version 3.10" notice, but it applies **only to the `loop` parameter** ("Deprecation warning is emitted if _obj_ is not a Future-like object and _loop_ is not specified and there is no running event loop") — not to `ensure_future()` itself. The function is still documented and supported in 3.13. The docs recommend `create_task()` / `TaskGroup` over `ensure_future()` for new tasks, but stylistic preference is not deprecation. Verified locally on 3.11 with `warnings.simplefilter('error')`: no `DeprecationWarning` is emitted.

   The recommendation to prefer `background_tasks.create()` is correct and well-justified — but justify it on its actual merits (GC-cancel protection, exception handling, naming) and drop the false deprecation claim. An LLM that ingests this will repeat the falsehood as fact.

   ```diff
   -# BAD: asyncio.create_task() or asyncio.ensure_future() — GC may cancel bare tasks;
   -#      ensure_future() is also deprecated since Python 3.10
   +# BAD: asyncio.create_task() or asyncio.ensure_future() — the GC may cancel bare
   +#      task references, and exceptions are silently swallowed unless awaited.
   ```

   And in the Mental Model #7 bullet:

   ```diff
   -- `background_tasks.create()` for fire-and-forget coroutines (not `asyncio.create_task()` or `asyncio.ensure_future()` — the GC can cancel bare tasks; `ensure_future()` is also deprecated since Python 3.10)
   +- `background_tasks.create()` for fire-and-forget coroutines — wraps `loop.create_task` but keeps a reference (so the GC won't cancel it) and routes exceptions through NiceGUI's exception handler
   ```

5. **`loop.run_in_executor` is not deprecated**

   [llms.md:120](llms.md#L120) and the styling table at [llms.md:178](llms.md#L178):

   > `await asyncio.to_thread(cpu_func)` (Python 3.9+, preferred over the deprecated `run_in_executor`)

   `run_in_executor` is not deprecated. NiceGUI itself uses it internally — see [nicegui/run.py:65](nicegui/run.py#L65). `to_thread` is a thin wrapper and is preferable for ergonomics, but `run_in_executor` is the only way to dispatch work to a custom executor (`ProcessPoolExecutor` for CPU work, etc.). Recommend ergonomic preference, not false deprecation.

   ```diff
   -| CPU work in async | `asyncio.get_event_loop().run_in_executor(None, fn)` | `await asyncio.to_thread(fn)` |
   +| Run blocking fn in thread | `asyncio.get_event_loop().run_in_executor(None, fn)` | `await asyncio.to_thread(fn)` (3.9+; for custom executors, `run_in_executor` is still correct) |
   ```

6. **"Shared page" describes pre-3.0 behavior; script mode is the current model**

   [llms.md:585-586](llms.md#L585):

   > Shared page (all users see the same elements — rare, usually not desired)
   > Just put `ui.*` calls at module level (outside `@ui.page`)

   This is the old auto-index page model, which 3.0 replaced with **script mode** (see [section_pages_routing.py:25-30](website/documentation/content/section_pages_routing.py#L25-L30) and [ui_run.py:127-160](nicegui/ui_run.py#L127-L160)). Today, when `ui.*` calls live at module level with no `@ui.page` or `root` function, NiceGUI re-executes the script once per client connection (via `runpy.run_path`) — so each visitor gets a fresh execution and their own UI instance. The "all users see the same elements" warning is no longer true.

   Suggested rewording:

   ```diff
   -# Shared page (all users see the same elements — rare, usually not desired)
   -# Just put ui.* calls at module level (outside @ui.page)
   +# Script mode: when ui.* calls are at the top level of the entry script (no @ui.page,
   +# no root function), NiceGUI re-executes the script once per client connection.
   +# Each visitor gets a fresh execution and their own UI — convenient for prototypes
   +# and demos. State defined in *other* imported modules is still process-wide.
   ```

7. **`ui.color_picker` description is wrong**

   [llms.md:800](llms.md#L800):

   > `ui.color_picker()` — Standalone color picker

   `ColorPicker` subclasses `Menu` ([nicegui/elements/color_picker.py:9](nicegui/elements/color_picker.py#L9)) — it is a popover, not a standalone widget. The "standalone" claim will lead LLMs to use it as a top-level element when it should be attached to a button/icon. Correct phrasing: "popover color picker (anchors to a parent element)".

8. **Self-contradicting recommendation**

   [llms.md:1109](llms.md#L1109):

   > Keep files under 200–300 lines; refactor when exceeding this

   The file itself is 1,210 lines, and the NiceGUI library does not enforce any such limit — see e.g. `nicegui/element.py` (~1k lines), `nicegui/binding.py` (~400 lines). This rule isn't in `CONTRIBUTING.md`. Either drop it or move it under a clearly-labeled "Recommended for downstream apps" header so an LLM doesn't apply it to PRs against `nicegui/` itself.

9. **`ui.button().props('href="/download" tag=a ...')` quoting**

   [llms.md:1039](llms.md#L1039): The props string mixes quoted (`href="/download"`) and unquoted (`tag=a`) values. NiceGUI's props parser does handle this, but it would be clearer to either consistently quote or rely on the simpler form (`href=/download tag=a`). Easy place for an LLM to mis-copy.

10. **Stale-on-arrival claims to double-check**
    - "ECharts (recommended for most charts)" — there is no such recommendation in the website docs; this is a personal preference. Either drop "recommended" or add a sentence on why.
    - The Storage table at [llms.md:128-135](llms.md#L128) gives `app.storage.client` and `app.storage.tab` identical "until tab closes" persistence, which hides the actual difference: `client` is per-connection and dropped on reload; `tab` survives reload within the same tab (within `storage_purge_lapse`). That reconnect distinction is what users actually trip over.

joko-zauberzeug:
**@evnchn @falkoschindler** — thanks for the thorough review, both of you. A quick word of context before I respond to the specifics:

I'm not a software developer at your level — I work mostly on electronics and hardware at Zauberzeug. What opened up a completely new world for me is "vibe coding" with Claude: describing what I want and iterating together. But I kept hitting a wall: Claude knows NiceGUI, but without the right context it would drift toward raw CSS, JavaScript, and Quasar props instead of using the Python API. Feeding it the full docs is technically possible but very token-heavy in practice.

`llms.md` mostly solved that problem for me. So Evan — please don't trust this too much! It came from someone who knows just enough Python to be dangerous, not from a NiceGUI expert. The whole point was to build something useful precisely *because* I'm not an expert.

**On Evan's suggestion (#auto-discovery):** I love the idea in principle. Teaching an LLM to discover new elements dynamically via a grep or enumeration command would make the file future-proof. My concern is token cost — by the time an LLM runs the discovery and reads the results, you've spent a lot of context budget. Could be elegant as a server-side generated artifact for new releases though (a `llms.txt` generated from the actual codebase). Worth exploring separately.

**On the factual errors (#3, #4, #5):** Corrected and pushed. The binding model was genuinely wrong — thank you for catching it with code references. The deprecation claims for `ensure_future` and `run_in_executor` were sloppy; I leaned on what Claude told me and didn't verify. Fixed.

I'd like to go through the remaining points (#1, #2, #6–#10) one by one — happy to do that here or in whatever format works best for you.

joko-zauberzeug:
**On #1 (single source of truth) and #2 (filename/location):**

The drift concern is completely valid. Long-term, the right architecture is the inverse: LLM-relevant hints live *in the source* — docstrings or tagged comments per element/function — and `llms.txt` gets generated automatically from those. Similar to how OpenAPI docs are generated from code annotations. Whoever touches a feature also updates its LLM guidance; drift becomes structurally impossible. `llms.md` is the proof-of-concept to figure out what content actually helps before distributing it across the codebase.

On `.md` vs `.txt`: the llmstxt.org convention says `.txt`, which makes sense for auto-discovery tooling. For humans, `.md` is easier to read and GitHub renders it nicely. For LLMs, the extension doesn't matter. If the file gets generated from code tags eventually, the maintainers can choose the format at generation time — that decision belongs with the programmers who build the pipeline, not with me.

On location: for now, the repo root is fine for manual referencing (`@https://raw.githubusercontent.com/zauberzeug/nicegui/main/llms.md`). But the ideal long-term home might be **inside the package** — at `nicegui/llms.md`. That way it ships with `pip install nicegui` and is available locally without any network fetch. Serving it additionally from `nicegui.io/llms.txt` would cover auto-discovery for tools following the llmstxt.org standard. Again, the programmers know best where this fits the package structure.

Liebe Grüße, Joko

falkoschindler:
@joko-zauberzeug @evnchn — pulling this thread together.

My initial review leaned toward "migrate the load-bearing content into the docs site as a single source of truth." After the discussion with both of you and some self-correction, I'm walking that back.

Updates to my thinking:

- **On the structural sync ideas raised in the thread** (Evan's auto-discovery, Joko's tagged-docstrings, and my own docs-site migration suggestion). All three try to dissolve the duplication problem — by having the LLM grep at runtime, by generating `llms.txt` from source annotations, or by spreading content across the docs site. But none of them help with what's actually in `llms.md`: mental models and anti-patterns are hand-written prose, and prose carries drift regardless of where it lives or how it's distributed. The auto-derivation ideas remain interesting for future API-doc work (signatures and types _are_ mechanically derivable); the migration is reasonable as a future discoverability move. Neither resolves what this PR contributes.

- **On bundling vs website.** Joko had the better instinct here. For someone on `nicegui==3.7`, an "always current" website would be _worse_ — an LLM gets pointed at APIs that don't exist in their installed version. A file bundled with the package is versioned correctly, faster (no fetch), and offline-capable. Both forms are useful for different audiences.

**Resulting proposal:**

1. **Move the file to `nicegui/llms.md`** so it ships with each install and is versioned automatically.
2. **Add `/llms.txt` and `/llms.md` website routes** (same body, `.txt` canonical) for auto-discovery and current-`main` browsing. Link from the docs footer for discoverability. Include a one-liner that `Accept: text/markdown` (#5889) makes every docs page Markdown-addressable, so agents don't need to parse HTML.
3. **Add a pre-release checklist step** to have an LLM check `llms.md` for correctness and completeness against the codebase. Catches drift at release boundaries — which is when it matters, since the bundled file is what users get pinned to their installed version.

If this direction works for you, I'll handle the move and the route setup myself — no need to push more onto your branch. Shout if anything doesn't sit right.

evnchn:
I think given our release schedule on Tuesday means that we are often done circa Friday and bunhunting in the meantime, we can afford to spend some time to sync `nicegui/llms.md` to each version after we are done with the PRs. I agree.

falkoschindler (approved):
LGTM, merging once CI is green.

Final state of the PR after the back-and-forth:

- **Content** — the factual corrections from the earlier review pass are all in (binding model is push+pull, `ensure_future`/`run_in_executor` no longer claimed deprecated, script mode replaces "shared page", color picker is a popover, storage table distinguishes client vs. tab correctly). Mental Model #5 heading now matches its body.
- **Distribution** — bundled inside the package at `nicegui/llms.md` (verified the wheel and sdist both contain it via `uv build`), plus `/llms.md` and `/llms.txt` served by the website. Both routes return `text/markdown; charset=utf-8` — the llmstxt.org spec defines the content as Markdown but specifies no MIME type, so this is the honest answer for both.
- **Discoverability** — README points downstream users at the bundled file, footer links to `/llms.txt`, top of `llms.md` explains the two ways to get it (pinned-to-installed-version vs. current-`main`) and the `Accept: text/markdown` content negotiation that complements it.

Thanks @joko-zauberzeug for the initial draft and the patient iteration on the factual fixes; thanks @evnchn for the auto-discovery angle, which is the right pointer for the eventual generated-from-source successor.

KrilleGH:
Before #6067/ #6070 I wanted to say that also for humans this is a nice document, but didn't get around to write it down. I think it still applies, maybe to somewhat lesser degree. It could be better to have two documents, targeted to the respective audience. Although this probably incurs some duplication, some things just don't change.

evnchn:
Hot take on my side: llms.md is the modern version of the "distraction-free reader". 

It IS both for humans and LLMs because _humans are creatures born and thriving in language._

PR #6026, opened 2026-05-03
Add orientation axes inset overlay to Scene

Jepson2k:
### Motivation

Three.js ships with a `ViewHelper` — a small XYZ orientation gizmo that indicates the camera's view direction and snap-animates the camera to look down an axis when an axis sprite is clicked. This PR exposes it through `Scene` so users can drop a small inset into a corner of the canvas without writing JS.

This was originally part of #5993 (scene-clipping-axes), but supporting it on the current `^0.180.0` three.js pin required a manual scissor + viewport block in the render loop and a fake-clientX/clientY synthesis around `ViewHelper.handleClick` to compensate for r180's hardcoded bottom-right/dim=128 assumptions — both fragile workarounds. r184 added a `viewHelper.location` field that handles positioning natively, so I split this out as a follow-up that targets nicegui v4 (where the three.js pin can move to r184).

### Implementation

Two new methods on `Scene`:

- `set_axes_inset(*, enabled, margin, margin_x, margin_y, anchor)` — toggles the inset on a configurable corner. Position is wired through `viewHelper.location` (`{top|bottom: my, left|right: mx}`). The inset is fixed at 128 px (a hardcoded constant inside three.js' `ViewHelper`).
- `set_axes_labels(*, enabled, labels, font, color, radius)` — replaces the default `X`/`Y`/`Z` labels and customizes their typography. Forwards to `viewHelper.setLabels` and `setLabelStyle`.

Click-to-snap is wired via a single capture-phase `pointerdown` listener that calls `viewHelper.handleClick(event)` and stops propagation when an axis sprite is hit, so OrbitControls doesn't start a drag-rotate on what was meant to be an inset click.

Both methods cache their opts on the Python side and replay them in `_handle_init`, so a fresh client connect (e.g. after a reload) re-applies the inset without the host having to re-call them.

The scene-bundle's `three` pin is bumped from `^0.180.0` to `^0.184.0` because `viewHelper.location` was added in r184. As a result this change can only land in nicegui v4 — earlier versions can't merge it without backporting the location math by hand.

### Progress

- [ ] The implementation is complete. (Otherwise, open a [draft PR](https://github.blog/news-insights/product-news/introducing-draft-pull-requests/).) — drafted, blocked on three.js pin bump for v4

---

Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


Jepson2k:
@evnchn @falkoschindler Not sure whats the process for PR that require package upgrades (and therefore next nicegui v4 release). Let me know if this should be switched back to a draft or if there is a v4 branch this should target instead.

falkoschindler:
Hi @Jepson2k, there's no separate v4 branch — we keep development on `main` and use the 4.0 milestone for what's deferred to the next major. I'll attach this PR to that milestone. Once we cut 4.0 and bump the three.js pin on `main`, you can rebase and we'll take a full review pass.

evnchn:

Reinforcing the v4 gate so this doesn't get accidentally swept onto a 3.x bump in a triage pass — `^0.180.0` → `^0.184.0` crosses four releases with hard removals and renames, not just soft deprecations. User-facing footguns that would silently break downstream JS:

- **r181** renames: `PI2` → `TWO_PI`, `PassNode.setResolution()` → `…Scale()`, `WaterMesh.resolution` → `resolutionScale`, `AfterImageNode.damp` no longer accepts plain numbers. PBR indirect-specular brightness also shifts visibly.
- **r182** `MeshPostProcessingMaterial` deleted; `WebGLCubeRenderTarget` no longer accepted by `WebGPURenderer`; `colorBufferType` → `outputBufferType`.
- **r183** `PostProcessing` → `RenderPipeline`; `FileLoader.load()` / `ImageBitmapLoader.load()` no longer return a value (callers using the return silently break); `Sky` legacy gamma path gone; `RoomEnvironment` lighting shifts.
- **r184** `FBXLoader` auto-rotates +Z-up → +Y-up (existing imports may flip 90°); `SSAAPassNode.clearColor` / `clearAlpha` removed.

`Scene`'s own Python surface is fine, but user JS touching `scene.three` / custom loaders / post-processing absorbs these silently. Defer-to-4.0 is the right call.

PR #6027, opened 2026-05-03
Fix worker traceback on Ctrl-C in run.cpu_bound

### LINKED ISSUE #6025: Using run.cpu_bound() makes ctrl-c throw exception
### First Check

### Example Code

```python
from nicegui import run, ui

def p():
    return "foo"

@ui.page("/")
def main():
    async def f():
        print(await run.cpu_bound(p))

    ui.button("compute", on_click=f)

ui.run()
```

### Description

When I try to exit the program using ctrl+c after using run.cpu_bound(), I get this:

```
^CProcess SpawnProcess-1:1:
  File "/home/justus/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
    ~~~~~~~~^^
  File "/home/justus/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/multiprocessing/process.py", line 108, in run
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/justus/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/concurrent/futures/process.py", line 242, in _process_worker
    call_item = call_queue.get(block=True)
  File "/home/justus/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/multiprocessing/queues.py", line 101, in get
    res = self._recv_bytes()
  File "/home/justus/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/multiprocessing/connection.py", line 216, in recv_bytes
    buf = self._recv_bytes(maxlength)
  File "/home/justus/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/multiprocessing/connection.py", line 430, in _recv_bytes
    buf = self._recv(4)
  File "/home/justus/.local/share/uv/python/cpython-3.13.12-linux-x86_64-gnu/lib/python3.13/multiprocessing/connection.py", line 395, in _recv
    chunk = read(handle, remaining)
KeyboardInterrupt
```

This also happens using the [progress example](https://github.com/zauberzeug/nicegui/blob/main/examples/progress/main.py), except that you have to press ctrl+c twice. This is caused by Manager(), but I don't know why. Maybe a separate issue.

If I use app.shutdown() instead, everything works fine (except for the Manager() thing, so I still have to press ctrl+c).

### NiceGUI Version

3.11.1

### Python Version

Python 3.13.12

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

falkoschindler:
### Motivation

Closes #6025.

After calling `run.cpu_bound`, pressing Ctrl-C prints an alarming traceback even though shutdown actually succeeds:

```
Process SpawnProcess-1:1:
  ...
  File ".../concurrent/futures/process.py", line 242, in _process_worker
    call_item = call_queue.get(block=True)
  ...
KeyboardInterrupt
```

Root cause: SIGINT is delivered to the whole foreground process group. The parent (uvicorn) handles it gracefully and `_shutdown()` calls `run.tear_down()` → `_kill_processes()` to terminate workers. But the workers themselves had no custom SIGINT handler, so the default raised `KeyboardInterrupt` mid-`call_queue.get()` and dumped a traceback before the parent could kill them.

### Implementation

Pass `initializer=_ignore_sigint` to `ProcessPoolExecutor` (both in `setup()` and in the `BrokenProcessPool` recovery path). The initializer installs `signal.SIG_IGN` for `SIGINT` so workers stay quiet — the parent's existing `tear_down()` is what actually stops them.

The `multiprocessing.Manager()` traceback the reporter also mentions in the progress example is a stdlib quirk outside NiceGUI's surface; it's not addressed here.

### Progress

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import signal
 import traceback
 from collections.abc import Callable
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
@@ -24,7 +25,7 @@ def setup() -> None:
     """Setup the process pool. (For internal use only.)"""
     global process_pool  # pylint: disable=global-statement # noqa: PLW0603
     try:
-        process_pool = ProcessPoolExecutor()
+        process_pool = ProcessPoolExecutor(initializer=_ignore_sigint)
     except NotImplementedError:
         logging.warning('Failed to initialize ProcessPoolExecutor')
 
@@ -94,7 +95,7 @@ async def cpu_bound(callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs)
         try:
             await _run(process_pool, safe_callback, lambda: None)
         except BrokenProcessPool:
-            process_pool = ProcessPoolExecutor()
+            process_pool = ProcessPoolExecutor(initializer=_ignore_sigint)
         finally:
             raise e
 
@@ -128,6 +129,11 @@ def tear_down() -> None:
         _kill_processes()
         process_pool.shutdown(wait=True, cancel_futures=True)
     thread_pool.shutdown(wait=False, cancel_futures=True)
+
+
+def _ignore_sigint() -> None:
+    """Ignore SIGINT in worker processes so only the parent handles Ctrl-C (see #6025)."""
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
 
 
 def _kill_processes() -> None:


evnchn:
I'm trying to question myself whether the fix is really that simple, with empirical testing to back things up. 

Is it me or "multiprocessing = tough battle" is engrained in my mind.

evnchn:
**TL;DR:** Claude Code (here as "I") empirically validated on Linux + macOS — bug reproduces 3/3 without the fix, stays silenced 3/3 with it. No regressions in the edge cases I poked. Looks good.

<details>
<summary>Test details</summary>

Driver: `subprocess.Popen(start_new_session=True)` then `os.killpg(pgid, SIGINT)` to simulate a terminal Ctrl-C reaching the whole process group.

| Platform | Python | mp start method | main (no fix) | PR 6027 |
|---|---|---|---|---|
| Linux | 3.10 + uvloop | spawn (forced) | worker traceback 3/3 | clean 3/3 |
| Linux | 3.10 + uvloop | fork (default) | clean 3/3 | clean 3/3 |
| macOS | 3.14.3 + uvloop | spawn (default) | worker traceback 3/3 | clean 3/3 |

Worth a footnote: on Linux with the default fork context, the bug doesn't manifest at all on my box — parent's `_kill_processes` SIGKILL appears to win the race against the worker's KeyboardInterrupt print. The bug surfaces cleanly with spawn (macOS default), which matches the `Process SpawnProcess-1:1` name in the original report. Fix is correct regardless of which context the user has selected.

Edge cases on the PR branch — all clean:

- **SIGINT mid-`cpu_bound`** (long-running job): `cpu_bound` returns `None` via `_run`'s existing `CancelledError` swallow. Pre-existing behavior, no new traceback.
- **`cpu_bound` never called**: process pool created but no worker spawned — no regression, clean shutdown.
- **Parent SIGKILLed externally** (no `tear_down` path): worker survival behavior identical on `main` and PR branch. Pre-existing concern, not introduced by this PR.

`multiprocessing.Manager()` noise in `examples/progress/main.py` confirmed untouched, matching the acknowledgment in the PR body.
</details>

evnchn (approved):
See above. I tested things empirically so it's an OK for me. If you are unsure then you can wait for Windows but that's another can of worms...

PR #6042, opened 2026-05-06
Fix KeyError when validating after error prop is removed

### LINKED ISSUE #5977: ValidationElement.validate() raises KeyError: 'error' after removing error via the public props() API
### First Check

### Example Code

```python
from nicegui import ui

input_element = ui.input(value="valid", validation=lambda value: None if value else "required")

input_element.props(remove="error error-message")
input_element.validate()
```

### Description

I am using a validated `ui.input` and clearing Quasar validation props through NiceGUI's public `props()` API.

I expect either:
- the public API to reject removal of reserved validation props, or
- `validate()` to recover gracefully.

What currently happens is that `validate()` raises:

```py
KeyError: 'error'
```

This seems to happen because `ValidationElement.error` later reads `self._props['error']`, but that key can still be removed through `props(remove="error ...")`.

### NiceGUI Version

3.9.0

### Python Version

3.12.3

### Browser

Chrome

### Operating System

Linux

### Additional Context

This looks related to #3809, which also hit a `ValidationElement KeyError` around the reserved `error` prop. The trigger here seems different: in this case the key can be removed through the public `props(remove="error ...")` API and the later `validate()` call then crashes.

falkoschindler:
### Motivation

Closes #5977.

`ValidationElement.error` setter compared `self._props['error']` directly. If a user had previously removed the `error` prop via `element.props(remove='error error-message')`, the next call to `validate()` would raise `KeyError: 'error'`.

### Implementation

Replace the bare subscript with `.get()` so the comparison is robust to the prop having been removed externally. Behavior is otherwise unchanged: a missing prop compares not-equal to the new value and the setter proceeds to re-seed both `error` and `error-message`.

Adds a regression test that reproduces the original repro: `ui.input(validation=...)`, `props(remove='error error-message')`, then `validate()` — must not raise.

### Progress

- [x] Pytest has been added.

### DIFF
--- a/tests/test_input.py
+++ b/tests/test_input.py
@@ -134,6 +134,16 @@ def test_setting_error_without_validation(screen: Screen):
     screen.should_contain('Something is wrong')
 
 
+def test_validate_after_removing_error_prop(screen: Screen):
+    @ui.page('/')
+    def page():
+        input_ = ui.input('Name', value='valid', validation=lambda v: None if v else 'required')
+        ui.button('Remove error', on_click=lambda: input_.props(remove='error error-message').validate())
+
+    screen.open('/')
+    screen.click('Remove error')
+
+
 def test_autocompletion(screen: Screen):
     input_ = None
 
--- a/tests/test_input.py
+++ b/tests/test_input.py
@@ -138,10 +138,11 @@ def test_validate_after_removing_error_prop(screen: Screen):
     @ui.page('/')
     def page():
         input_ = ui.input('Name', value='valid', validation=lambda v: None if v else 'required')
-        ui.button('Remove error', on_click=lambda: input_.props(remove='error error-message').validate())
+        ui.button('Reset', on_click=lambda: (input_.props(remove='error error-message').validate(), ui.label('done')))
 
     screen.open('/')
-    screen.click('Remove error')
+    screen.click('Reset')
+    screen.should_contain('done')
 
 
 def test_autocompletion(screen: Screen):
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -46,7 +46,7 @@ class ValidationElement(ValueElement[ValueT]):
         :param error: The optional error message
         """
         new_error_prop = None if error is None and self.validation is None else (error is not None)
-        if self._error == error and self._props['error'] == new_error_prop:
+        if self._error == error and self._props.get('error') == new_error_prop:
             return
         self._error = error
         self._props['error'] = new_error_prop


evnchn:
_Claude Code drive-by, acting independently — drafted by an agent on Evan's behalf so he can review both the PR and this comment as separate signals. No human pre-read._

## TL;DR

Mostly LGTM, with one structural nit on the test. Empirically verified on Linux / Python 3.10 / NiceGUI HEAD:

- **Bug reproduces on plain `origin/main`**: bare REPL of issue #5977's repro raises `KeyError: 'error'` at `validation_element.py:49`, exactly as described.
- **Patch fixes it**: `.get('error')` returns `None` for the missing key, so `None != False`, the short-circuit falls through, and the setter re-seeds both props — happy-path behavior preserved.
- **No regressions**: full `tests/test_input.py` green (14/14) on the patched branch.
- **Test catches the bug in CI** — `pytest` exits `1` on plain main + new test, `0` on patched. **But the failure is surfaced by `conftest`'s "unexpected ERROR logs" teardown hook, not by any assertion in the test body.** Worth a small follow-up so the regression check doesn't depend on the implicit log-level gate.

<details>
<summary>Empirical run results (Linux .68, Python 3.10.20, pytest 9.0.3)</summary>

### Plain `origin/main` (`0e842eb1`) + new test only, no code fix

This is the load-bearing question: does the new test actually fail CI without the fix?

```
$ git worktree add /tmp/nicegui-main-with-test origin/main
# inserted the new test verbatim from the PR at tests/test_input.py:137

$ CHROME_BINARY_LOCATION=/usr/bin/chromium-browser \
    uv run pytest tests/test_input.py::test_validate_after_removing_error_prop -x --tb=short
collected 1 item
tests/test_input.py .E
=========== ERROR at teardown of test_validate_after_removing_error_prop =======
Failed: There were unexpected ERROR logs.
ERROR    nicegui:app.py:170 'error'
  File ".../nicegui/events.py", line 469, in handle_event
    result = cast(Callable[[], Any], handler)()
  File ".../tests/test_input.py", line 141, in <lambda>
    ui.button('Remove error', on_click=lambda: input_.props(remove='error error-message').validate())
  File ".../nicegui/elements/mixins/validation_element.py", line 76, in validate
    self.error = cast(str | None, result)
  File ".../nicegui/elements/mixins/validation_element.py", line 49, in error
    if self._error == error and self._props['error'] == new_error_prop:
KeyError: 'error'
========================== 1 passed, 1 error in 2.56s ==========================
EXIT=1
```

So at item level the test is **PASSED** (no asserts in the body, the click handler raises server-side, the screen fixture returns). The teardown then notices unexpected ERROR-level log output and converts that into a non-zero exit. CI does go red — so the test isn't useless — but it's surfacing the regression via a `conftest` side-channel rather than via the test itself making a claim.

### Patched branch (`pr-6042` HEAD `782edf2b`) + same test

```
$ uv run pytest tests/test_input.py::test_validate_after_removing_error_prop -x
collected 1 item
tests/test_input.py .                                                    [100%]
============================== 1 passed in 3.30s ===============================
EXIT=0
```

### Python REPL on broken (issue #5977's exact repro, no pytest, no server)

```
$ uv run python -c "from nicegui import ui; \
  i = ui.input(value='valid', validation=lambda v: None if v else 'required'); \
  i.props(remove='error error-message'); \
  i.validate(); print('OK')"
  File "<string>", line 1, in <module>
  File ".../nicegui/elements/mixins/validation_element.py", line 76, in validate
    self.error = cast(str | None, result)
  File ".../nicegui/elements/mixins/validation_element.py", line 49, in error
    if self._error == error and self._props['error'] == new_error_prop:
KeyError: 'error'
```

### Python REPL on patched (same one-liner)

```
$ uv run python -c "from nicegui import ui; \
  i = ui.input(value='valid', validation=lambda v: None if v else 'required'); \
  i.props(remove='error error-message'); \
  i.validate(); print('OK')"
OK
```

### Full `tests/test_input.py` on patched

```
$ uv run pytest tests/test_input.py
collected 14 items
tests/test_input.py ..............                                       [100%]
============================= 14 passed in 27.46s ==============================
EXIT=0
```

</details>

### Verdict

The fix does exactly what the PR says: swapping `self._props['error']` for `self._props.get('error')` in the equality short-circuit removes the `KeyError` when the prop was stripped via `props(remove=...)`. Bug reliably reproduces pre-patch (pytest **and** bare REPL), gone post-patch, no collateral damage to the rest of `test_input.py`.

**Suggested tweak to the test (non-blocking).** Today the test has no `assert` — it relies on `conftest`'s "unexpected ERROR logs" teardown hook to flip CI red. If that hook is ever relaxed (level filtered, log capture scoped differently, etc.), this regression silently re-mutes. I left a concrete `suggestion` block as an [inline review comment on the test](https://github.com/zauberzeug/nicegui/pull/6042#discussion_r3194674655) so it's one-click apply-or-dismiss; functionally the test catches the bug today, structurally it would be a touch more robust.

The thornier design questions raised in the issue thread (slot-aware `error` setter / `bool` validation returns) are explicitly out of scope here, and that seems right — ship the cheap defensive hardening first, debate the API question separately.

tests/test_input.py
```diff
@@ -134,6 +134,16 @@ def page():
...
 
 
+def test_validate_after_removing_error_prop(screen: Screen):
+    @ui.page('/')
+    def page():
+        input_ = ui.input('Name', value='valid', validation=lambda v: None if v else 'required')
+        ui.button('Remove error', on_click=lambda: input_.props(remove='error error-message').validate())
+
+    screen.open('/')
+    screen.click('Remove error')
```
evnchn: _Claude Code, suggestion via inline review comment._

Non-blocking, structural nit (rationale in [the parent comment](https://github.com/zauberzeug/nicegui/pull/6042#issuecomment-4386984758)): the test currently has no body-level assertion — it relies on `conftest`'s "unexpected ERROR logs" teardown gate to fail CI when the bug fires server-side. Two-line variant that makes the regression check explicit using `screen.should_contain` (matches the pattern at line 134 of this file):

```suggestion
        ui.button('Remove error', on_click=lambda: (input_.props(remove='error error-message').validate(), ui.label('validated')))

    screen.open('/')
    screen.click('Remove error')
    screen.should_contain('validated')
```

Mechanism: lambda's comma-tuple is evaluated left-to-right — if `validate()` raises (pre-patch), `ui.label` never runs and `screen.should_contain('validated')` fails at the assertion. Post-patch: clean evaluation, label renders, check passes.

**Verified empirically** on `origin/main` (without the fix) + this variant → pytest `1 failed, 1 error`, exit 1, `AssertionError: Could not find "validated"`. On `pr-6042` HEAD + this variant → `1 passed`, exit 0. So the variant catches the regression at body-assertion level (with the conftest teardown gate also firing as defense-in-depth).

_(Edited: previous post had a 4-space indent on the `ui.button` line; should be 8 spaces — inside `def page()`. Re-posted with correct indent.)_

evnchn (dismissed):
LGTM. Extra comment with suggestion is to make the test more robust, but it works as it is. The fix is literally a few characters what could go wrong.

tests/test_input.py
falkoschindler: You're right. An assertion is more robust and communicates the intention more clearly. Fix it in 2964671a4afe2e9513f9031fb39999edb1a68f63.

evnchn (approved):
Then let's go, my previous suggestion was adopted.

PR #6043, opened 2026-05-06
Fix click handler dispatch in user simulation

Jepson2k:
### Motivation

`UserInteraction.click()` short-circuits with `return self` after handling several element types (`ui.tab`, `ui.link`, `ui.select`), so click handlers attached via `.on('click', ...)` don't fire under user simulation even though they would in a real browser. On top of that, the dispatch loop fires every listener on the element regardless of event type — and the implicit toggle of `ui.checkbox` / `ui.switch` was riding on that very lack of filtering, dispatching to the `update:modelValue` listener with `args=not element.value`.

### Implementation

- Drop the `return self` short-circuits for `ui.tab`, `ui.link`, and `ui.select`. For `ui.select`, the click handler fires only on the wrapper-click paths (opening, and closing without selection) — option clicks stay silent, matching Quasar.
- Extract the dispatch into a `_dispatch_click` helper that filters listeners by `type == 'click'`.
- Replace the implicit checkbox/switch toggle with an explicit `element.value = not element.value`.
- Fire the `ui.link` click handler before `navigate.to`, matching browser semantics.
- Make every dispatch branch self-contained (no implicit fall-through to a trailing handler).
- Make `click()` always-first single-target: pick the lowest-id enabled element and dispatch once. A real user clicks one element at a time, so multi-match `find().click()` no longer broadcasts.
- Add coverage for tab, link, and select click handlers, plus a click-handler assertion on the checkbox/switch test.

### Notes for users

- Click handlers on `ui.checkbox` / `ui.switch` now receive `e.args = None` instead of `not element.value`. Read `e.sender.value` (or use `.on_value_change(...)`) for the post-toggle value.
- `find(...).click()` no longer broadcasts to every matched element — it picks the lowest-id enabled match and dispatches once. Tests that relied on simultaneous multi-match clicks need to issue separate calls.

### Progress

- [x] Breaking changes to the public API: see _Notes for users_ above.


evnchn:
> _Posted on Evan's behalf by Claude Code (Opus 4.7) — empirical work against main 5c8fd15a._

TL;DR: confirmed the regression and the fix locally. Also checked whether `return self` should be dropped from the `ui.radio` / `ui.select` branches too. **It shouldn't** — Selenium ground truth says those don't fire `.on('click', ...)` on the real browser either, so the existing short-circuits happen to be faithful. #6043 is correctly scoped to `ui.tab`.

<details>
<summary>Empirical detail (side-by-side <code>screen</code> vs <code>user</code> on main 5c8fd15a)</summary>

| element | `screen` (Chromium) | `user` (post-fix) |
|---|---|---|
| `tab.on('click', h)` | fires | fires |
| `radio.on('click', h)` | does NOT fire | does NOT fire |
| `select.on('click', h)` | does NOT fire | does NOT fire |

Test pattern: persistent label set via `.on('click', ...)` (not `ui.notify`, since toasts fade before `screen.should_contain` polls). Quasar's `q-option-group` / `q-select` don't bubble option-clicks to the root in a way that fires the root's `@click` — so `ui.radio(...).on('click', ...)` is effectively a no-op in real browsers too.

Naively dropping `return self` from radio/select also crashes under `user`: `TypeError: list indices must be integers or slices, not NoneType` at `radio._event_args_to_value(e)`. The generic loop reaches the internal `update:VALUE_PROP` handler registered at `value_element.py:49` with `args=None`. So broadening would both crash *and* introduce a new divergence from the browser.

Latent observation (not in scope here): the generic listener loop in `click()` is event-type-agnostic — it fires every listener in `_event_listeners` regardless of registered event name. Doesn't bite this PR but worth a separate look.

</details>

evnchn (dismissed):
Empirical test OK for me. @falkoschindler please look before merge thanks since you're more pro in user fixture.

falkoschindler:
Thanks for the fix, @Jepson2k! The diagnosis is spot-on and the regression test is exactly the right shape.

While reviewing, I got curious whether the same `return self` pattern in other branches of `UserInteraction.click()` was also dropping handlers that _do_ fire in a real browser. I ran the snippet below and clicked each element manually:

```python
ui.link('Link', '#').on('click', lambda: ui.label('You clicked the link!'))
ui.radio(['A', 'B', 'C'], value='A').on('click', lambda: ui.label('You clicked the radio!'))
ui.select(['A', 'B', 'C'], value='A').on('click', lambda: ui.label('You clicked the select!'))
with ui.tabs():
    ui.tab('Tab A').on('click', lambda: ui.label('You clicked Tab A!'))
ui.tree([{'id': 'A'}], label_key='id').on('click', lambda: ui.label('You clicked the tree!'))
```

| Element     | Real `on('click')` fires?                       | User-sim fires it? | Verdict                                      |
| ----------- | ----------------------------------------------- | ------------------ | -------------------------------------------- |
| `ui.link`   | yes                                             | no                 | **same regression as `ui.tab`** — needs fix  |
| `ui.radio`  | no                                              | no                 | fine — `return self` matches reality         |
| `ui.select` | yes (only on input field; not on popup options) | no                 | **partial regression** — needs investigation |
| `ui.tab`    | yes                                             | no                 | fixed by this PR                             |
| `ui.tree`   | no                                              | no                 | fine — `return self` matches reality         |

So the same root-cause regression also affects `ui.link` (clearly) and `ui.select` (with a wrinkle).

The `ui.select` case is more nuanced: Quasar's `q-select` dispatches `click` on the input/wrapper itself, but clicks on options inside the popup do _not_ bubble to the user's `on('click')` handler. That means user-sim fidelity needs to distinguish:

- popup-closed → opened (user clicks input): real fires handler → user-sim should fire too
- popup-open, option clicked: real does _not_ fire handler → user-sim correctly drops it
- popup-open, wrapper re-clicked to close: real fires handler → user-sim should fire too

I'd like to dig into the `ui.select` branch a bit more before recommending a precise fix there. For `ui.link` the fix is the same shape as the `ui.tab` one in this PR.

falkoschindler (dismissed):
Approving — and pushed a follow-up commit on top, coming back from the "dig into `ui.select`" detour. It addresses the three regressions in one shape:

1. **`ui.link` click handler** — same fix as your `ui.tab` one: drop `return self` and let the dispatch loop run. Updated `test_click_link` to assert the handler fires.

2. **`ui.select` click handler** — fires only on the wrapper-click paths (popup open, popup close without selection) and stays silent when an option is picked, matching Quasar's behavior. Added `test_select_click_handler` covering all three cases.

3. **Type-agnostic listener loop** — picking up the latent observation @evnchn flagged. Listeners are now filtered by `type == 'click'`. Surfaced an implicit dependency: the previous unfiltered loop was the only thing toggling `ui.checkbox` / `ui.switch` — it was reaching their `update:modelValue` listener with `args=not element.value`. Replaced that hack with an explicit toggle branch in the shape of the `ui.tab` one.

The dispatch is now extracted into a small `_handle_click_event` helper so all the special-cased branches can share it. 71 tests in `test_user_simulation.py` pass.

falkoschindler:
Layered a few minor readability follow-ups (handler-fires-before-navigate for `ui.link`, self-contained dispatch per branch with `_handle_click_event` → `_dispatch_click`, and a click-handler assertion on the checkbox/switch test) plus one scope change that I want to flag explicitly since it lands after approval — happy to revert any of them.

**Scope change** (02e7df19): `click()` now picks a single matched element and dispatches once, instead of fanning out to every match.

This came up while flattening the dispatch. The if/elif chain had two unrelated control-flow patterns: `ui.link` / `ui.select` / `ui.radio` / `ui.tree` stopped at the first match, while `ui.button` / `ui.tab` / `ui.checkbox` / `ui.switch` broadcast to all matches. Once I tried to articulate _why_ it was split that way, I couldn't — it tracked which branches happened to have `return self` historically, not anything semantic.

Two arguments for unifying as always-first (rather than always-broadcast):

- Simulating a _user_: a real user clicks one thing at a time, not N buttons in zero time.
- Determinism: with broadcast, a multi-match `find('go to').click()` on links would queue N navigations and the last one would win — and `set` iteration order isn't insertion-ordered, so "last" is non-deterministic.

The observable change: any test where the selector matches multiple elements (probably a test-authoring mistake regardless) would now click only the lowest-id match instead of all of them. I've added a _Notes for users_ sub-section to flag this.

falkoschindler:
@evnchn Sorry for the noise and the growing diff. But I couldn't resist to clean up this for-loop. For reviewing it's best to view in VS Code where the changed indentation is more obvious than on GitHub.

evnchn:
No need to apologize to me. I'm always fine to review whatever diff comes my way. Let me wrap my head around the new diff. The person who's in shock is probably @Jepson2k. The TL-DR is that "we found more implementation gaps while reviewing your PR and preferred to fix it right away otherwise your diff is not even adopted at all."

Overall I need to double check whether "`click()` now picks a single matched element and dispatches once" is a good idea because, as I reiterate, then end goal of user simulation is to **simulate the `screen` fixture** which is **what happens were a browser was involved.** 

As such, **if `click()` dispatches to multiple elements in `screen` fixture and now it doesn't for `user` fixture** then I **will** reject the PR on the grounds of `user` VS `screen` mismatch. 

TL-DR: whether to dispatch to multiple elements or not is not a taste preference which can go one way or another, but factual alignment where there is only one correct answer. (MRE will be available to prove it one way or another)

evnchn:
> _Posted on Evan's behalf by Claude Code (Opus 4.7) — empirical work against PR head `02e7df19`, baseline against upstream `main` `e8a720ad`._

**TL;DR:** the `02e7df19` "always-first" change matches `screen` and **fixes** a user/screen divergence that already ships on `main`. It doesn't introduce one. Recommendation: keep the scope change.

| call | dispatches | source of truth |
|---|---|---|
| `screen.click('GO')` | **1 element** (first DOM match) | [`screen.py:182`](https://github.com/zauberzeug/nicegui/blob/02e7df19/nicegui/testing/screen.py#L182) → `self.find()` → Selenium `find_element` (singular) |
| `user.find('GO').click()` on `main` (`e8a720ad`) | **3 elements** (broadcast) | `for element in self.elements: ...` |
| `user.find('GO').click()` just before scope change (`02e7df19~1`) | **3 elements** (broadcast) | same loop, unchanged from main |
| `user.find('GO').click()` on PR head (`02e7df19`) | **1 element** (lowest-id match) | `min(enabled_elements, key=lambda e: e.id)` |

So the `user`/`screen` divergence runs the other way today: shipping `user` broadcasts where `screen` clicks once. The PR closes that gap.

<details>
<summary>MRE + run output</summary>

Three buttons sharing the prefix `'GO'`, each appending its own click marker.

```python
# test_multimatch_click.py
from nicegui import ui
from nicegui.testing import Screen, User


def page_with_three_buttons() -> None:
    @ui.page('/')
    def index() -> None:
        results = ui.column()
        for i in range(3):
            btn = ui.button(f'GO {i}')
            btn.on('click', lambda _e, idx=i: ui.label(f'clicked-{idx}').move(results))


def test_screen_clicks_one(screen: Screen) -> None:
    page_with_three_buttons()
    screen.open('/')
    screen.click('GO')
    screen.wait(0.5)
    body = screen.selenium.find_element('xpath', '//body').text
    assert 'clicked-0' in body
    assert 'clicked-1' not in body
    assert 'clicked-2' not in body


async def test_user_clicks_one(user: User) -> None:
    page_with_three_buttons()
    await user.open('/')
    user.find('GO').click()
    await user.should_see('clicked-0')
    await user.should_not_see('clicked-1')
    await user.should_not_see('clicked-2')
```

**Result on `02e7df19` (PR head):** both tests pass. Only `clicked-0` rendered.

**Result on `e8a720ad` (upstream/main) and `02e7df19~1` (PR before the scope change):** `test_user_clicks_one` fails on the first `should_not_see('clicked-1')` — page contains all three labels (`clicked-0`, `clicked-1`, `clicked-2`). `user.find('GO').click()` broadcast.

`screen.click('GO')` body dump on PR head:

```
clicked-0
GO 0
GO 1
GO 2
```

Single dispatch. Buttons 1 and 2 received nothing.

**One caveat for the future**, not a blocker here: post-PR `user` picks `min(elements, key=lambda e: e.id)`, i.e. lowest creation-order id. `screen` picks first DOM match via XPath. In normal pages where elements are created top-to-bottom, these coincide (and they do here). They could in principle diverge if elements are reparented after creation (e.g. `.move()`d into different order from creation). Worth noting if anyone hits a weird "wrong element clicked" mismatch later — but no observable divergence in this MRE.

</details>

---

**Edit log:** swapped the user-fixture test from a clunky `try/except should_see` pattern to plain `should_not_see('clicked-1')` / `should_not_see('clicked-2')`, mirroring the `screen` test's `assert 'clicked-1' not in body` shape (per [@evnchn](https://github.com/evnchn)). Also added the upstream `main` (`e8a720ad`) row to the comparison table — confirms the broadcast is the shipping behavior today, not just an artifact of an intermediate commit on this branch.

evnchn:
> _Posted on Evan's behalf by Claude Code (Opus 4.7) — empirical work against PR head `02e7df19` and `main` `e8a720ad`._

**TL;DR (correctness):** the lowest-id heuristic in `02e7df19` aligns with `screen.click()` for *every* page where elements stay in creation order — but it diverges from `screen` if any matched element has been `.move()`-reparented after creation. Whether that gap is worth closing depends on whether you consider post-construction reparenting in scope.

## The gap, made concrete

Three buttons created in order (ids 0, 1, 2), then `buttons[0].move(target_index=-1)` puts button 0 at the end of the column. DOM is now `[GO 1, GO 2, GO 0]`; ids unchanged.

| call | dispatches to | source of "first match" |
|---|---|---|
| `screen.click('GO')` | button 1 | first XPath match in document order ([screen.py:182](https://github.com/zauberzeug/nicegui/blob/02e7df19/nicegui/testing/screen.py#L182) → Selenium `find_element`) |
| `user.find('GO').click()` on `02e7df19` (lowest-id) | **button 0** | `min(elements, key=e.id)` — picks creation order, not DOM |

So `user` clicks the moved-away button while `screen` clicks the one that's now visually first. Same shape of `user`/`screen` mismatch the PR was opened to fix, just gated on `.move()`.

<details>
<summary>MRE</summary>

```python
# tests/test_move_divergence.py
def page_with_moved_button():
    @ui.page('/')
    def index():
        col = ui.column()
        with col:
            buttons = []
            for i in range(3):
                btn = ui.button(f'GO {i}')
                btn.on('click', lambda _, idx=i: ui.label(f'clicked-{idx}'))
                buttons.append(btn)
        buttons[0].move(target_container=col, target_index=-1)


def test_screen_clicks_first_in_dom(screen: Screen) -> None:
    page_with_moved_button()
    screen.open('/')
    screen.click('GO')
    body = screen.selenium.find_element('xpath', '//body').text
    assert 'clicked-1' in body and 'clicked-0' not in body  # first DOM = button 1


async def test_user_status_quo_diverges(user: User) -> None:
    page_with_moved_button()
    await user.open('/')
    user.find('GO').click()
    await user.should_see('clicked-0')          # ← lowest-id, not first-in-DOM
    await user.should_not_see('clicked-1')
```

Both tests pass on `02e7df19`. The user-test passing is itself the divergence.

</details>

## How niche is "niche"?

To weight the perf discussion below, I measured 16 representative pages from `website/` (running each in the user fixture and counting `client.layout.descendants`):

| metric | median | max | example max |
|---|---:|---:|---|
| **N** (total elements descendant of layout) | 388 | 756 | `/documentation/table` |
| **d** (max depth) | 10 | 14 | — |
| **s** (largest sibling list seen) | 64 | 129 | `/documentation/table` |
| **k** (largest single-type bucket) | 76 | 235 | `/documentation/page_layout` |

In **none** of these pages does `.move()` participate in normal construction — `.move()` is reserved for apps that explicitly reparent (drag-and-drop, dynamic layout, sortable lists). So the divergence above triggers only for tests of reparenting-using apps.

## Drop-in fix if you want to close the gap

Walk the slot tree depth-first preorder, returning the first match. This is the same traversal NiceGUI's existing `Element.descendants(include_self=True)` does (and matches XPath document order by construction), just hand-recursive to dodge generator-frame overhead:

```python
def _first_in_dom_order(matched, root):
    matched_set = set(matched)
    def walk(node):
        if node in matched_set:
            return node
        for slot in node.slots.values():
            for child in slot.children:
                hit = walk(child)
                if hit is not None:
                    return hit
        return None
    return walk(root)


# in click(): replace `min(elements, key=lambda e: e.id)` with:
element = _first_in_dom_order(enabled_elements, self.user.client.layout)
```

**Cost vs. status quo** (median µs/call, 2000 iters × 5 repeats, .68):

| variant | N=388, k=5 | N=388, k=50 | N=750, k=200 |
|---|---:|---:|---:|
| `min(.., key=e.id)` (status quo) | 0.66 | ~3 | ~10 |
| `_first_in_dom_order` (this) | 2.7 | 3.8 | 7.7 |

For k=5 the drop-in costs ~2 µs more per click; for k≥50 the descendants walk is actually cheaper than the lowest-id `min` because `min` is linear in k while the walk early-exits at the first DOM-match. Either way, both are dwarfed by the ~hundreds-of-µs Quasar event round-trip in the user simulation.

**Correctness vs. status quo:** identical answer on every test in `tests/test_user_simulation.py` (71/71 pass with this drop-in), and on every page where elements are in creation order (so all current behavior preserved). Diverges from status quo only on `.move()`-reparented pages, where it matches `screen`.

Cross-validated against the descendants-generator reference on **130+ random page shapes** with random nesting and 0-15 random `.move()` perturbations per page. Zero mismatches.

<details>
<summary>Approach I tried and discarded — path-tuple <code>min</code></summary>

`min(matched, key=path_tuple)` where `path_tuple(e)` is the `(slot_idx, child_idx)` sequence from root to `e`, lex-ordered by document order. Asymptotically `O(k·d)` vs. `O(N)` for the descendants walk, so I expected a win when `k << N`. I didn't get one.

Best optimized variant (walk-up via `_parent_slot` weakref, scan parent's slots only):

| variant | N=388, k=5 | N=388, k=50 | N=750, k=200 |
|---|---:|---:|---:|
| `_first_in_dom_order` (chosen) | 2.7 | 3.8 | 7.7 |
| `min(matched, key=walkup_key)` | 18.2 | 182 | **751** |

Crossover with the descendants walk is at `k ≈ 2`. Above that, the path-key approach pays a full ancestor-walk for *every* match, while the descendants walk early-exits on the first DOM hit. By `k=200` it's ~100× slower despite the better Big-O. So no realistic regime where it wins on this codebase.

</details>

## Recommendation

If `.move()` use cases are out of scope for `user` simulation: do nothing. The lowest-id heuristic in `02e7df19` is an objective improvement over the broadcasting on `main`, and the remaining gap doesn't trigger for normal pages.

If you want `user.find().click()` to behave like `screen` for *all* pages including reparented ones: drop in the snippet above. 4 lines plus a 12-line helper, costs ~2 µs at k=5 (cheaper than status quo for k≥50), no behavior change for any currently-tested case, closes the divergence by construction.

Happy to open this as a follow-up PR if useful — wanted to surface the gap and the data first since it's tangential to what `02e7df19` set out to fix.

Sources:
- [W3C XPath 1.0 §5 — document order](https://www.w3.org/TR/1999/REC-xpath-19991116/)
- [Selenium WebDriver — Finding web elements (`find_element` returns first DOM match)](https://www.selenium.dev/documentation/webdriver/elements/finders/)

evnchn:
> _Posted on Evan's behalf by Claude Code (Opus 4.7) — independent code-reviewer subagent dispatched against PR head `02e7df19`, with the `.move()` corner case from my earlier follow-up comment held out of scope per Evan's call._

**TL;DR:** Ready to merge. No Critical, no Important. One process recommendation around release notes; otherwise the PR matches its plan precisely and tests cover the breaking change.

<details>
<summary>Strengths</summary>

- **`_dispatch_click` extraction** (`user_interaction.py:155-163`): clean separation; the `listener.type != 'click'` filter resolves the latent type-agnostic listener bug from the original thread, while `listener.element_id != element.id` preserves the existing self-listener guard. Reusing the same helper across all branches makes the dispatch surface uniform.
- **Self-contained branches**: each `elif` handles its own dispatch (or deliberately omits it for `ui.radio`/`ui.tree`). Eliminates the implicit fall-through the previous checkbox/switch toggle relied on.
- **`ui.select` wrapper-vs-option fidelity** (`user_interaction.py:90-115`): matches Falko's three-state table — popup-open + valid option = silent; popup-open + close-without-selection = handler; popup-closed + open = handler.
- **`ui.link` ordering** (`user_interaction.py:84-86`): handler fires before `navigate.to`, matching browser semantics.
- **`enabled_elements` guard hoisted to top** (`user_interaction.py:79-81`): single-pass filter, avoids per-branch `isinstance` checks.
- **Tests scoped right**: `test_select_click_handler` covers all three popup states (`tests/test_user_simulation.py:578-595`); `test_tab_click_handler` (`:914-922`); `test_click_link` updated with handler assertion (`:357,365`); checkbox/switch test updated with `e.sender.value` read (`:195,205,209`).
- **Multi-match → single-match scope change** correctly identified as *closing* a `user`/`screen` divergence (Selenium `find_element` is also single-target), not introducing one.

</details>

<details>
<summary>Recommendation</summary>

Carry the PR body's "Notes for users" verbatim into the GitHub release notes when this ships, since the `e.args` change for `ui.checkbox`/`ui.switch` click handlers is a behavior break (existing handlers receiving `not element.value` will start receiving `None`). The PR description handles it cleanly; just want to make sure it lands somewhere users see at upgrade time, since the project ships release notes rather than a CHANGELOG.

</details>

**Verdict:** Yes, merge. The `.move()` corner case in my earlier comment is a separate follow-up if you want it; not a blocker for this.

evnchn (approved):
Super-TLDR: Since last time we meet, we revamped the PR to cover the browser-to-user-fixture alignment more in depth, with the only gap being `.move()` which needs work if we decide to narrow that gap, if not can merge right away.

falkoschindler:
Thanks for the thorough analysis, @evnchn!
Let's merge this one. If there's demand, we can improve the ordering in a follow-up PR.

evnchn:
@falkoschindler please be sure to make clear the behavioural change in the release notes, as suggested in the "Recommend" block in https://github.com/zauberzeug/nicegui/pull/6043#issuecomment-4408663801, since one can argue it's a breaking change but we argue it's user-to-screen alignment.

PR #6045, opened 2026-05-08
Fix SemLock fork/spawn RuntimeError in `ui.run(native=True, reload=True)`

### LINKED ISSUE #1841: Runtime Error when trying to run native=True
### Description

I'm getting a Fork vs Spawn error when attempting to run a native app on Ubuntu.

**Sample Application (from documentation)**
```py
from nicegui import app, ui

app.native.window_args['resizable'] = False
app.native.start_args['debug'] = True

ui.label('app running in native mode')
ui.button('enlarge', on_click=lambda: app.native.main_window.resize(1000, 700))

ui.run(native=True, window_size=(400, 300), fullscreen=False)

```

Invoked from command line from within a virtual environment:  python main.py

**Trackback**
```
  File "/mnt/ext3tbdrive/sandbox/try-nicegui/main.py", line 15, in <module>
    main()
  File "/mnt/ext3tbdrive/sandbox/try-nicegui/main.py", line 11, in main
    ui.run(native=True, window_size=(400, 300), fullscreen=False)
  File "/mnt/ext3tbdrive/sandbox/try-nicegui/env/lib/python3.11/site-packages/nicegui/run.py", line 180, in run
    ChangeReload(config, target=globals.server.run, sockets=[sock]).run()
  File "/mnt/ext3tbdrive/sandbox/try-nicegui/env/lib/python3.11/site-packages/uvicorn/supervisors/basereload.py", line 49, in run
    self.startup()
  File "/mnt/ext3tbdrive/sandbox/try-nicegui/env/lib/python3.11/site-packages/uvicorn/supervisors/basereload.py", line 85, in startup
    self.process.start()
  File "/home/tim/.pyenv/versions/3.11.6/lib/python3.11/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
                  ^^^^^^^^^^^^^^^^^
  File "/home/tim/.pyenv/versions/3.11.6/lib/python3.11/multiprocessing/context.py", line 288, in _Popen
    return Popen(process_obj)
           ^^^^^^^^^^^^^^^^^^
  File "/home/tim/.pyenv/versions/3.11.6/lib/python3.11/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/home/tim/.pyenv/versions/3.11.6/lib/python3.11/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/home/tim/.pyenv/versions/3.11.6/lib/python3.11/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/home/tim/.pyenv/versions/3.11.6/lib/python3.11/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
  File "/home/tim/.pyenv/versions/3.11.6/lib/python3.11/multiprocessing/synchronize.py", line 107, in __getstate__
    raise RuntimeError('A SemLock created in a fork context is being '
RuntimeError: A SemLock created in a fork context is being shared with a process in a spawn context. This is not supported. Please use the same context to create multiprocessing objects and Process.
```

OS: Ubuntu 20.04
Python: 3.11.6
Pyenv: 2.3.29-5-g62e2e9c7
nicegui: 1.3.17
pywebview: 4.3.3

evnchn:
### Motivation

Fixes #1841. `ui.run(native=True, reload=True)` raises `RuntimeError: A SemLock created in a fork context is being shared with a process in a spawn context.` on CPython ≥ 3.11.5. NiceGUI's native-mode `Queue`/`Pipe`/`Event` use the default (fork) context, but uvicorn's `ChangeReload` worker is spawn-context, so when ChangeReload pickles parent state for the worker the fork-context SemLocks fail the cross-context check that landed in [cpython#77377](https://github.com/python/cpython/issues/77377). Older CPython silently corrupted state instead of raising, which is the long tail of "leaked semaphore" / "semaphore released too many times" reports in the same issue.

Empirical evidence — distro × backend × Python sweep, exact 3.11.4↔3.11.5 boundary, upstream commit, all in [#1841 (comment)](https://github.com/zauberzeug/nicegui/issues/1841#issuecomment-4408850083).

### Implementation

Switch the three affected primitives to an explicit `multiprocessing.get_context('spawn')`. This is the pattern Python's docs recommend for libraries ([Contexts and start methods](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods): *"Libraries using multiprocessing should be designed to allow their users to provide their own multiprocessing context"*) and what CPython's own SemLock error message tells you to do (*"Please use the same context to create multiprocessing objects and Process"*). Pywebview already uses the same pattern in [`examples/pystray_icon.py`](https://github.com/r0x0r/pywebview/blob/master/examples/pystray_icon.py).

The alternative considered — `multiprocessing.set_start_method('spawn', force=True)` — was rejected because it mutates the user's global default and would break user code that relies on fork (CoW model loading, post-fork file descriptors, ML pipelines). The patch here keeps the user's default untouched and only spawn-types NiceGUI's three IPC primitives.

Verified on `python:3.12-slim` Docker against the patched repo via `PYTHONPATH` (no monkeypatch): unpatched fires `SEMLOCK_FIRED`, patched runs to event-loop block. After-import `multiprocessing.get_start_method(allow_none=True)` is unchanged; `mp.get_context('fork').Queue()` in user code still works.

### Progress

---

*Drafted by Claude Code working with @evnchn — empirical investigation across distros / CPython patch versions, then surgical fix matching pywebview's existing pattern.*

### DIFF
--- a/nicegui/native/native.py
+++ b/nicegui/native/native.py
@@ -12,8 +12,7 @@ from typing import Any
 from .. import run
 from ..logging import log
 
-# spawn context so primitives can be pickled into uvicorn's spawn-context ChangeReload worker (#1841)
-_SPAWN_CTX = multiprocessing.get_context('spawn')
+SPAWN_CONTEXT = multiprocessing.get_context('spawn')  # match uvicorn's ChangeReload worker (#1841)
 
 method_queue: Queue | None = None
 response_queue: Queue | None = None
@@ -24,9 +23,9 @@ event_sender: Connection | None = None
 def create_queues() -> None:
     """Create the message queues and event pipe. (For internal use only.)"""
     global method_queue, response_queue, event_receiver, event_sender  # pylint: disable=global-statement # noqa: PLW0603
-    method_queue = _SPAWN_CTX.Queue()
-    response_queue = _SPAWN_CTX.Queue()
-    event_receiver, event_sender = _SPAWN_CTX.Pipe(duplex=False)
+    method_queue = SPAWN_CONTEXT.Queue()
+    response_queue = SPAWN_CONTEXT.Queue()
+    event_receiver, event_sender = SPAWN_CONTEXT.Pipe(duplex=False)
 
 
 def remove_queues() -> None:
--- a/nicegui/native/native.py
+++ b/nicegui/native/native.py
@@ -2,14 +2,18 @@
 from __future__ import annotations
 
 import inspect
+import multiprocessing
 import warnings
 from collections.abc import Callable
-from multiprocessing import Pipe, Queue
+from multiprocessing import Queue
 from multiprocessing.connection import Connection
 from typing import Any
 
 from .. import run
 from ..logging import log
+
+# spawn context so primitives can be pickled into uvicorn's spawn-context ChangeReload worker (#1841)
+_SPAWN_CTX = multiprocessing.get_context('spawn')
 
 method_queue: Queue | None = None
 response_queue: Queue | None = None
@@ -20,9 +24,9 @@ event_sender: Connection | None = None
 def create_queues() -> None:
     """Create the message queues and event pipe. (For internal use only.)"""
     global method_queue, response_queue, event_receiver, event_sender  # pylint: disable=global-statement # noqa: PLW0603
-    method_queue = Queue()
-    response_queue = Queue()
-    event_receiver, event_sender = Pipe(duplex=False)
+    method_queue = _SPAWN_CTX.Queue()
+    response_queue = _SPAWN_CTX.Queue()
+    event_receiver, event_sender = _SPAWN_CTX.Pipe(duplex=False)
 
 
 def remove_queues() -> None:
--- a/nicegui/ui_run.py
+++ b/nicegui/ui_run.py
@@ -244,7 +244,7 @@ def run(root: Callable | None = None, *,
         width, height = window_size or (800, 600)
         native_host = '127.0.0.1' if host == '0.0.0.0' else host
         if reload:
-            shutdown_event = multiprocessing.Event()
+            shutdown_event = multiprocessing.get_context('spawn').Event()  # match uvicorn ChangeReload worker (#1841)
         native_favicon = str(Path(favicon).resolve()) if favicon and helpers.is_file(favicon) else None
         native_module.activate(protocol, native_host, port, title, width, height, fullscreen, frameless,
                                shutdown_event, native_favicon)
--- a/nicegui/ui_run.py
+++ b/nicegui/ui_run.py
@@ -244,7 +244,7 @@ def run(root: Callable | None = None, *,
         width, height = window_size or (800, 600)
         native_host = '127.0.0.1' if host == '0.0.0.0' else host
         if reload:
-            shutdown_event = multiprocessing.get_context('spawn').Event()  # match uvicorn ChangeReload worker (#1841)
+            shutdown_event = native_module.native.SPAWN_CONTEXT.Event()
         native_favicon = str(Path(favicon).resolve()) if favicon and helpers.is_file(favicon) else None
         native_module.activate(protocol, native_host, port, title, width, height, fullscreen, frameless,
                                shutdown_event, native_favicon)


evnchn:
_This comment was written by [Claude Code](https://claude.com/claude-code) running autonomously on @evnchn's homelab while he's away; the content below was **not screened by him** — his only involvement was kicking the run off. — Claude_

## TL;DR

@falko-schindler — confident for the **3.12** merge. Extended the validation matrix beyond what's already in the [#1841 evidence comment](https://github.com/zauberzeug/nicegui/issues/1841#issuecomment-4408850083) (which was about the bug, not the fix): added both pywebview backends, `reload={True,False}`, Python 3.10–3.14, macOS, fork-preservation, full pytest. Everything that should be green is green; the one red is pre-existing on `main`.

## Test matrix (patched code, not monkey-patch)

Per cell: fresh `python:X-slim` Docker, apt-installs backend, `pip install nicegui pywebview[…]`, then `PYTHONPATH=/nicegui-src` to swap in the patched repo. Probe runs `ui.run(native=True, reload=R, show=False)` under `timeout --signal=KILL 12s`; SIGKILL @ 12s means we got past SemLock and were blocking on the event loop = WORKS.

| Python | Backend | `reload` | Variant   | `_SPAWN_CTX` | Verdict |
|--------|---------|----------|-----------|--------------|---------|
| 3.10   | GTK     | True     | unpatched | —            | RAN _(no SemLock check pre-3.11.5)_ |
| 3.10   | GTK     | True     | patched   | ✓            | RAN |
| 3.11   | GTK     | True     | **unpatched** | —        | **SEMLOCK_FIRED** |
| 3.11   | GTK     | True     | patched   | ✓            | RAN |
| 3.12   | GTK     | True     | **unpatched** | —        | **SEMLOCK_FIRED** |
| 3.12   | GTK     | True     | patched   | ✓            | RAN |
| 3.12   | GTK     | False    | unpatched | —            | RAN _(reload=False bypasses ChangeReload, sanity)_ |
| 3.12   | GTK     | False    | patched   | ✓            | RAN _(no regression)_ |
| 3.13   | GTK     | True     | **unpatched** | —        | **SEMLOCK_FIRED** |
| 3.13   | GTK     | True     | patched   | ✓            | RAN |
| 3.12   | **Qt**  | True     | **unpatched** | —        | **SEMLOCK_FIRED** _(bug is backend-independent)_ |
| 3.12   | Qt      | True     | patched   | ✓            | RAN |
| 3.13   | Qt      | True     | **unpatched** | —        | **SEMLOCK_FIRED** |
| 3.13   | Qt      | True     | patched   | ✓            | RAN |

Both backends, both reload modes, full 3.10–3.13 sweep: every unpatched ≥3.11.5 fires SEMLOCK, every patched is clean.

## Python 3.14 — orthogonal quirk, worth a heads-up

Python 3.14 changed the POSIX default start method from `fork` → `forkserver`. Empirically (both patched and **unpatched**):

- **Without** `if __name__ == '__main__'` guard at the call site → both error with `RuntimeError: An attempt has been made to start a new process before the current process has finished its bootstrapping phase ... freeze_support()`. Affects all multiprocessing-using code on 3.14, not specific to NiceGUI.
- **With** main-guard → both run cleanly (SIGKILL @ 12s). 3.14's SemLock semantics apparently tolerate `forkserver` → `spawn` cross-context, so the patch is a no-op there but isn't harmful.

Not a regression introduced by this PR. Likely worth a separate `docs/native.md` note or its own ticket when 3.14 enters the support matrix.

## No global side effect (user fork preservation)

Ran on macOS (3.12) and inside the Docker containers above. Patched code:

```
before nicegui import: start_method = None
after  nicegui import: start_method = None     ← user's global default UNCHANGED
_SPAWN_CTX.get_start_method() = spawn          ← internal context is spawn
fork-context Queue still works: got 'hello'    ← user's fork-using code still works
user set_start_method(fork, force=True): ok    ← user can still pick fork explicitly
```

The patch uses `multiprocessing.get_context('spawn')` only for NiceGUI's own three IPC primitives — it does **not** call `set_start_method('spawn', force=True)`. Users with fork-dependent workflows (CoW model loading, post-fork file descriptors, ML pipelines that fork from large parents) are unaffected.

## macOS (Darwin/arm64)

Darwin defaults to spawn since CPython 3.8, so the bug doesn't reproduce here, but the patched code imports cleanly, `_SPAWN_CTX.get_start_method() == 'spawn'`, fork-context Queue still works. No-op on macOS, but verified non-regression.

## Lint, type-check, pytest

- `pylint nicegui/native/native.py nicegui/ui_run.py` → **10.00/10**
- `mypy nicegui/native/native.py nicegui/ui_run.py` → **no issues**
- `pytest tests/` on patched: **365 passed, 1 pre-existing failure**

NiceGUI's pytest harness short-circuits native mode via `is_pytest()`, so no test currently exercises the changed lines. The Docker matrix above is the load-bearing verification. The lone red — `tests/test_lazy_imports.py::test_module_access_does_not_import_others` — reproduces identically on un-patched `main` HEAD (`2c04631d`): the test invokes system `python3` instead of the venv interpreter, and the system interpreter lacks `socketio`. Pure env-setup issue on my macOS sandbox, not introduced by this PR.

## Recommendation

Ship as-is in **3.12**. The patch is minimal, surgical, matches pywebview's existing precedent, and doesn't touch user-facing API. If you want a regression test, the simplest shape is an import-time assertion `assert nicegui.native.native._SPAWN_CTX.get_start_method() == 'spawn'` — happy to add it if you want.

evnchn:
@falkoschindler Super-TLDR: can ship in 3.12 :rocket:

falkoschindler (approved):

PR #6046, opened 2026-05-11
Improved authentication example

NichtJens:
### Motivation

The current authentication example has a few small problems:

- favicons are showing inconsistently
  - png favicons are shown only when logged in, emoji favicons always work
- the username/password page has unexpected (/no) focusing logic
- the widgets are not aligned

### Implementation

- png favicons are accessed via `/favicon.ico` (emojis are inlined)
  - => added `"/favicon.ico"` to `unrestricted_page_routes`
- enter on both username and password tried to log in
  - => implemented the more usual input focus logic:
    - auto-focus username
    - enter on username --> focus password
    - enter on password --> try to log in
- added `items-stretch` to `ui.card()`
- also, simplified `AuthMiddleware.dispatch` by inverting the logic as it was a bit hard to understand

In particular the favicon problem is not easy to understand when it happens. So, having it in the example is probably a good idea. I imagine, the same problem might trigger elsewhere as well, e.g., for `robots.txt`. This might warrant a comment somewhere? (where?)

If the `items-stretch` is desired, I can add a new screenshot...

### Progress

- [x] No breaking changes.

### DIFF
--- a/examples/authentication/main.py
+++ b/examples/authentication/main.py
@@ -14,6 +14,7 @@ from nicegui import app, ui
 # in reality users passwords would obviously need to be hashed
 passwords = {'user1': 'pass1', 'user2': 'pass2'}
 
+# top-level static routes like /favicon.ico must be unrestricted, otherwise the middleware redirects them to /login
 unrestricted_page_routes = {'/favicon.ico', '/login'}
 
 
--- a/examples/authentication/main.py
+++ b/examples/authentication/main.py
@@ -56,7 +56,7 @@ def login(redirect_to: str = '/') -> RedirectResponse | None:
     def try_login() -> None:
         if passwords.get(username.value) == password.value:
             app.storage.user.update(username=username.value, authenticated=True)
-            ui.navigate.to(redirect_to)
+            ui.navigate.to(redirect_to)  # go back to where the user wanted to go
         else:
             ui.notify('Wrong username or password', color='negative')
 
--- a/examples/authentication/main.py
+++ b/examples/authentication/main.py
@@ -61,7 +61,7 @@ def login(redirect_to: str = '/') -> RedirectResponse | None:
             ui.notify('Wrong username or password', color='negative')
 
     with ui.card().classes('absolute-center items-stretch'):
-        username = ui.input('Username').props('autofocus').on('keydown.enter', try_login)
+        username = ui.input('Username').props('autofocus').on('keydown.enter', lambda: password.run_method('focus'))
         password = ui.input('Password', password=True, password_toggle_button=True).on('keydown.enter', try_login)
         ui.button('Log in', on_click=try_login)
 
--- a/examples/authentication/main.py
+++ b/examples/authentication/main.py
@@ -14,7 +14,7 @@ from nicegui import app, ui
 # in reality users passwords would obviously need to be hashed
 passwords = {'user1': 'pass1', 'user2': 'pass2'}
 
-unrestricted_page_routes = {'/login'}
+unrestricted_page_routes = {'/favicon.ico', '/login'}
 
 
 @app.add_middleware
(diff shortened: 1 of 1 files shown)


examples/authentication/main.py
```diff
@@ -49,21 +55,37 @@ def test_page() -> None:
...
+    with ui.card().classes('absolute-center items-stretch'):
+        username = ui.input('Username').props('autofocus').on('keydown.enter', try_login)
         password = ui.input('Password', password=True, password_toggle_button=True).on('keydown.enter', try_login)
         ui.button('Log in', on_click=try_login)
+
     return None
 
 
+def focus(element):
+    ui.run_javascript(f'getElement({element.id}).$refs.qRef.focus()')
```
evnchn: Not sure if it works but heard `.run_method('focus')` works for an element. 

If it does, then we don't need this function at all.

evnchn (commented):
Will review in detail tomorrow but there's one thing which I think _could_ work.

examples/authentication/main.py
NichtJens: Yup, that works. Changed it. I had only found the JS version in a comment by @falkoschindler ...

falkoschindler (commented):
Thanks for the cleanup! The favicon fix and the inverted `dispatch` are clear wins. I do want to push back on the form-focus changes and suggest a tighter `dispatch`.

### 1. Drop the form-UX policy from `try_login`

CONTRIBUTING.md (line 235) asks for examples to be "focused on one concept and as minimal as possible." This example's concept is _authentication_ — middleware, storage, redirect — not form UX. The new `try_login` adds three branches of form-input policy (notify on empty username, silent refocus on empty password, notify on bad credentials), which distracts from what the example is teaching.

Two further reasons to keep it out:

- The PR description calls the original behavior "unexpected (/no) focusing logic," but doesn't pin down a concrete bug. Browsers already submit on Enter, Quasar advances on Tab — there's no malfunction described, just a preference.
- The replacement is itself asymmetric: empty username → notify + focus, empty password → focus only, both empty → notify + focus username. That asymmetry invites exactly the bikeshedding the original two-branch form avoided.

Suggestion: keep `autofocus` on the username input (one-line ergonomic win, no extra policy) and revert `try_login` to its original two-branch shape:

```python
def try_login() -> None:
    if passwords.get(username.value) == password.value:
        app.storage.user.update(username=username.value, authenticated=True)
        ui.navigate.to(redirect_to)
    else:
        ui.notify('Wrong username or password', color='negative')
```

### 2. `dispatch` can collapse further

The inverted form is much easier to read — thanks. It can go further: the condition is short enough to fit one line, which makes the whole policy visible at a glance:

```python
async def dispatch(self, request: Request, call_next):
    path = request.url.path
    if app.storage.user.get('authenticated') or path in unrestricted_page_routes or path.startswith('/_nicegui'):
    return RedirectResponse(f'/login?redirect_to={path}')
```

Dropping `, False` from `.get('authenticated')` keeps it inside the line-length cap; `None` is falsy and a bare `.get(key)` is the standard idiom (same in `login()`).

---

Otherwise looks good — `items-stretch` on the card, moving the early-return to the top of `login`, and `update(username=..., authenticated=True)` are all small but welcome.

NichtJens:
Sounds good! Should be all adjusted accordingly.

Also, thanks for adding the comment. I wasn't sure where/how you'd prefer that.

I absolutely see your point that it's better not to mix several things in a single example. I will add, however, that submitting a login form when pressing enter on the username will be unexpected for most. So, unless I was particularly blind when searching for it, it feels like `.run_method('focus')` should be made easier to find. Then everyone can indeed bikeshed their own focusing logic as much as they want ;)

falkoschindler (dismissed):
Thanks for the quick turnaround on the feedback — everything looks good.

To your point about Enter-on-username being unexpected: fair, and easy to fix without bringing the policy back. I pushed a small follow-up — Enter on the username field now advances focus to the password field via `password.run_method('focus')`. Captures the muscle-memory bit you were after; the bad-credentials notification handles an empty-username submit just fine.

Your separate point about `.run_method('focus')` discoverability is a good one and worth its own thread — feel free to open an issue if you'd like to chase it.

Let's merge! 🚀

falkoschindler (approved):

NichtJens:
Awesome :) That is a nice and minimal compromise.

I'll look into suggestions for the `.run_method('focus')` discoverability and open a ticket.

PR #6050, opened 2026-05-13
Improve diagnostic when storing a non-JSON-serializable value

### LINKED ISSUE #6044: Unhelpful error when storing sets in app.storage.user
### First Check

### Example Code

```python
from nicegui import app, ui

@ui.page("/")
def main():
    app.storage.user["a set"] = set()

ui.run(storage_secret="TEST")
```

### Description

The above example gives the following traceback:

```
  File "xxx/lib/python3.13/site-packages/nicegui/json/orjson_wrapper.py", line 62, in _orjson_converter
    raise TypeError(f'Object of type {obj.__class__.__name__} is not JSON serializable')
TypeError: Object of type ObservableSet is not JSON serializable

The above exception was the direct cause of the following exception:

  File "xxx/lib/python3.13/site-packages/nicegui/background_tasks.py", line 152, in _handle_exceptions
    task.result()
    ~~~~~~~~~~~^^
  File "xxx/lib/python3.13/site-packages/nicegui/background_tasks.py", line 131, in wrapper
    return await awaitable
           ^^^^^^^^^^^^^^^
  File "xxx/lib/python3.13/site-packages/nicegui/persistence/file_persistent_dict.py", line 52, in async_backup
    await f.write(json.dumps(self, indent=self.indent))
                  ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx/lib/python3.13/site-packages/nicegui/json/orjson_wrapper.py", line 43, in dumps
    return orjson.dumps(obj, option=opts, default=_orjson_converter).decode('utf-8')
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Type is not JSON serializable: ObservableSet
```

In the simple example, it is of course obvious where the problem is.

However, in a more complex code it is not, since I am not using `ObservableSet` directly at all and the traceback does not reference anything that helps me find the problem.

Questions/Ideas:
1. Better traceback
    - I have no good idea how to show the line of *inserting* the set as part of the traceback. (A bad idea, though -- there could be a custom check at insert time.)
    - However, since `app.storage.user` is always a dict (right?), it should be possible to show the key in the traceback. This does not look trivial either, though, as it would need custom logic in the serialization ... Thus:
2. Better serialization
    - It would be relatively simple to extend [this](https://github.com/zauberzeug/nicegui/blob/5c8fd15ad273a710c6643ac1114f5f1a77ceb438/nicegui/json/builtin_wrapper.py#L49) and [this](https://github.com/zauberzeug/nicegui/blob/5c8fd15ad273a710c6643ac1114f5f1a77ceb438/nicegui/json/orjson_wrapper.py#L54) to automatically convert `Observable*` (and `Persistent*` ?) types:
      ```python
      if isinstance(obj, ObservableSet):
          return list(obj)
      # ... etc ...
      ```
    - Bonus question: Wouldn't it be better if the two referenced pieces of code would do the exact same conversions to serializable types? I.e., be the same common function?

---

Since 2. looks quite straight forward (with or without the bonus point), I am happy to provide a PR for this.

### NiceGUI Version

3.11.1

### Python Version

3.13.13

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

falkoschindler:
### Motivation

Closes #6044.

When `app.storage.user` held a `set` (or any other value the JSON encoder cannot handle), the backup task surfaced `Type is not JSON serializable: ObservableSet` — referencing an internal class the user never typed, and pointing at the deferred backup task rather than the offending data.

### Implementation

- **`nicegui/persistence/serialization.py`** (new module) — exports a `dumps(data, identifier, *, indent=False)` helper that wraps `json.dumps` in `try/except TypeError`. On failure, a small recursive walker descends into `dict`/`list`/`tuple` containers to locate the offending key path and the leaf value, and a new `TypeError` is raised that mentions the storage identifier, the path, and the leaf's underlying type (`set` rather than the internal `ObservableSet` wrapper).

- **`nicegui/persistence/file_persistent_dict.py`** and **`nicegui/persistence/redis_persistent_dict.py`** — call the new `dumps(self, ...)` instead of `json.dumps(self)` at their respective backup sites, passing the storage filepath / Redis key as the identifier.

	Resulting error:
	
	```
	Could not serialize /…/storage-user-….json at ['a set']: value of type 'set' is not JSON-serializable
	```
	
	The walker only runs once the underlying `dumps` already failed, so steady-state cost is zero.

- **`tests/test_storage.py`** — adds `test_storage_serialization_error_points_at_offending_key`. Uses the `screen` fixture to drive a button click that writes a nested `set` into `app.storage.general`, then asserts via `assert_py_logger` that the error log mentions both the storage file and the offending nested key path with the type name.

### Progress


falkoschindler:
Note on why this PR doesn't include the orjson `default`-fallback tweak I proposed [earlier in the issue](https://github.com/zauberzeug/nicegui/issues/6044#issuecomment-4405714359):

Once the persistence wrapper builds its own message (with storage identifier + key path + underlying type), orjson's outer "ObservableSet" string never reaches the user-facing error anyway — we replace it. And on the FastAPI response path, a returned value is a plain `set`, not an `ObservableSet`, so the proposed `isinstance(obj, set)` branch would say exactly what the default branch already says. The only place the change would actually be observable is if someone called `nicegui.json.dumps(ObservableSet(...))` directly outside the persistence layer, and even then it would only change the inner chained exception. Narrow enough that the extra two lines weren't worth carrying.

evnchn:
@falkoschindler I quite like the idea! 2 things:

**1. Naming and parameter adjustment:**

IMO the core idea of the PR is an "identified dumps" as opposed to `json.dumps` which is annonymous. 

To make this more obvious to humans and AI alike (also defends against `from json import *`), I recommend renaming to `identified_dumps` with `identifier` being the first parameter. 

**2. Async diagnostic triggering:**

We need to do benchmark on the "small recursive walker"'s performance (task for Claude Code), especially against deeply nested datasets. If it is slow, then it could be harmful, because downstream code expects fast response, success of fail. We can't grind for a second or more just for hunting the offending key. 

If it is indeed slow, or could be slow in a circumstance, I recommend to start an async task to find the offending key instead, so it'd something like:

`Could not serialize BLAH.json. Please check logs for more details.`

Then either:

`On BLAH.json: Offending key path is ['a set'].`

Or, if we implement a 1-second timeout to avoid some _really_ nested datasets from getting in our way:

`On BLAH.json: Unable to find offending key path in 1 second.`

---

As we are acting on the `dumps`, this is more elegant than https://github.com/evnchn/nicegui/pull/146. I think I will close that if my Claude Code has no disagreements.

evnchn:
*Comment drafted by Claude Code (Opus 4.7) on Evan's behalf — bench numbers measured locally on Apple Silicon M4.*

@falkoschindler — followed up on the "benchmark the walker" item from my earlier comment. **The walker is synchronous-safe for realistic storage sizes; only adversarial multi-MB poisoned blobs stall it noticeably.** No async-task indirection needed for the common path, but a small **time budget** would cheaply bound the worst case.

<details>
<summary><b>Bench results (median of 5, Apple Silicon M4)</b></summary>

Walker source: copied verbatim from this PR's `nicegui/persistence/serialization.py`. Bench replicates the failure path (poisoned input → walker runs).

| workload | stdlib `json` | `orjson` |
|---|---:|---:|
| shallow bad (4 keys) | 0.008 ms | 0.002 ms |
| deep bad (depth 100) | 1.7 ms | 0.17 ms |
| wide bad (width 10000) | 14.6 ms | 3.3 ms |
| **realistic ~720KB user blob, poison in `users[150]`** | **16 ms** | **1.7 ms** |
| adversarial, poison-LAST among 1000 heavy siblings (~30MB JSON) | 277 ms | 41 ms |
| adversarial, poison-LAST among 5000 heavy siblings (~150MB JSON) | 1384 ms | 205 ms |
| any of the above with poison-FIRST | <0.01 ms | <0.01 ms (early exit) |

Key observations:
- The walker re-serializes every sibling at every level until it finds the failing one. Cost is dominated by **good-sibling serialization**, not bad-leaf detection.
- For ≤1MB storage (the realistic regime) even stdlib stays under 20ms — comfortably synchronous.
- Adversarial ordering matters more than depth: poison-first is free; poison-last on huge wide dicts is what stalls.

</details>

<details>
<summary><b>Proposal: add a time budget to the walker</b></summary>

Simpler than the async-task idea I floated earlier, and addresses the same worry (don't let diagnostics block the failure path indefinitely). Sketch:

```python
import time

class _DiagTimeout(Exception):
    pass

def dumps(data: object, identifier: str, *, indent: bool = False, diag_budget_s: float = 0.5) -> str:
    try:
        return json.dumps(data, indent=indent)
    except TypeError as e:
        try:
            leaf = _find_offending_leaf(data, time.perf_counter() + diag_budget_s)
        except _DiagTimeout:
            raise TypeError(
                f'Could not serialize {identifier} (offending key path search exceeded {diag_budget_s}s): {e}'
            ) from e
        if leaf is None:
            raise TypeError(f'Could not serialize {identifier}: {e}') from e
        path, value = leaf
        type_name = 'set' if isinstance(value, set) else type(value).__name__
        raise TypeError(
            f'Could not serialize {identifier} at {path}: value of type {type_name!r} is not JSON-serializable'
        ) from e


def _find_offending_leaf(data: object, deadline: float) -> tuple[str, object] | None:
    items = data.items() if isinstance(data, dict) else enumerate(data) if isinstance(data, (list, tuple)) else ()
    for key, value in items:
        if time.perf_counter() > deadline:
            raise _DiagTimeout
        try:
            json.dumps(value)
        except TypeError:
            deeper = _find_offending_leaf(value, deadline)
            if deeper is None:
                return f'[{key!r}]', value
            return f'[{key!r}]' + deeper[0], deeper[1]
    return None
```

Why **time-based** beats node-count:
- The user cares about wall-clock latency, not nodes visited. A 500ms cap means "the failure path never blocks for more than ~500ms" regardless of which JSON backend is installed (stdlib vs orjson have ~6× perf delta).
- Node-count budgets are hardware-independent but their meaningful values are opaque (100k? 1M?) and need re-tuning as serializer perf shifts.
- The check is one `time.perf_counter()` call per iteration — negligible cost.

Why **0.5s** as default:
- 30× headroom over the worst realistic case in the bench (16ms stdlib on 720KB).
- 2× under the 1s threshold I worried about being noticeable to downstream code.
- Easy to override per-call if a deployer knows their storage is huge and wants to allow longer searches.

If you'd prefer to skip the budget entirely and accept the rare multi-second stall on pathological storage, that's also defensible — the bench shows it really is rare. But the budget is ~10 lines, doesn't change the happy path, and gives a clean upper bound. Happy to send a PR on top of this one if you'd like.

</details>

Separately: closing [evnchn/nicegui#146](https://github.com/evnchn/nicegui/pull/146) in favor of this PR — your `dumps`-side approach is more surgical (zero steady-state cost, single incision point) than the insert-time validation I was exploring.

evnchn:
@falkoschindler Super TL-DR: Under all sane cases, the runtime is short, so "async" is a bad idea. However, unbounded offending key exploration would likely land us in a GHSA so we just need to ensure bounded, where 0.5s is a reasonable bound as recommended from empirical testing.

falkoschindler:
Thanks @evnchn — and the bench numbers are genuinely useful, appreciate the work. Two pushbacks though:

**On renaming to `identified_dumps`:** the function already lives in its own module (`nicegui/persistence/serialization.py`) and every call site imports it explicitly as `from .serialization import dumps`, so there's no realistic collision unless someone wildcard-imports from a private module — which isn't a case I want to design around. The `data`-first order also mirrors `json.dumps(data, …)`, which makes the wrap-relationship obvious at a glance. I'd rather keep the current name.

**On the time budget:** this path only fires when the user's own code has stuffed a non-serializable value into their own storage. That's a dev-time mistake, not a production hot path — once you see the error, you fix it and it doesn't fire again. And if an attacker can write 150MB into a user's `app.storage`, the diagnostic walker's wall-clock is not the problem we have. So I don't think the extra ~10 lines + new exception type + tunable knob pulls its weight here. Happy to revisit if anyone actually hits the slow path in the wild.

evnchn (approved):
Fair. I think the tunable error can be added when there is a feature request. After all, we don't have use cases to back it up yet, only very synthetic data

PR #6052, opened 2026-05-13
Add aria-label and child-content labels to button markdown

evnchn:
### Motivation

Phase 2 polish for #5889 (markdown content negotiation). `Button._render_markdown` only inspected the `label` and `icon` props, so a button that carried its meaning through an `aria-label` or a child element — e.g. `ui.button().props('aria-label="Save"')`, or a button wrapping a `ui.label` — collapsed to a bare `[Button]`, dropping information a reading agent could have used.

This PR surfaces that information, while **keeping `[Button]` as the representation for genuinely empty / decorative buttons** (such as the copy buttons whose only child is a phosphor icon). Dropping the `[Button]` token itself for decorative buttons was considered (per [#6007 finding 4](https://github.com/zauberzeug/nicegui/discussions/6007)) but not done here: the placeholder is a useful structural signal in the stream, and #6054 already stops the decorative child markup from leaking.

### Implementation

`nicegui/elements/button.py:_render_markdown()` now resolves, in order:

1. `label` prop → `[Button: <label>]` (existing)
2. `aria-label` prop → `[Button: <aria-label>]` (new)
3. `icon` prop → `[Button: icon:<icon>]` (existing)
4. A single plain line of child markdown (no newlines, no `[`/`]`) → `[Button: <child>]` (new)
5. Otherwise → `[Button]` (unchanged)

The child fallback is deliberately conservative: multi-child, bracketed (e.g. nested links), or empty/decorative children fall through to `[Button]` rather than garbling the `[Button: ...]` wrapper. Children whose markdown is suppressed via `override_markdown(..., '')` (#6054) render empty and are skipped, so a copy button stays `[Button]` without exposing raw `<i class="ph-...">` HTML.

Tests in `tests/test_markdown_response.py` cover the `aria-label` priority (below `label`, above `icon`), single-line child content, and children that fall back to `[Button]` (a nested-link child, a multi-child button, and a whitespace-only child).

### Progress

- [x] Documentation is not necessary (agent-facing rendering change).
- [x] No breaking changes to the public API — empty `ui.button()` still renders `[Button]`; only buttons carrying an `aria-label` or a single-line child gain a label. No Python API signature changes.


evnchn:
Hmm, is `[Button: <i class="...">] exposure: addressed by the companion decorative-HTML PR` even true, in this context?

falkoschindler (changes requested):

I've got some points that we should address before merging. I'm especially curious what you think about 1 and 2.

### 1. Empty buttons should keep the `[Button]` placeholder — _major_

The PR drops `button.py:55` from `'[Button]'` to `''`. Combined with the `if ... markdown` filter in `_children_to_markdown`, a button with no label, no aria-label, no icon, and no rendered children now vanishes entirely — no placeholder, no separator. That removes a structural signal: a reading agent can no longer tell that a button is present at all.

The motivating case — copy buttons whose only content is a decorative phosphor `<i>` — is the right thing to fix in #6054, where `phosphor_icon()` is wrapped with `override_markdown(..., '')`. With that in place, `_children_to_markdown()` returns `''` for such buttons (the empty child is filtered out by `element.py:218`), the children-fallback's truthy check fails, and we fall through to the final return — which should stay as `[Button]`, not `''`. Copy buttons then show up as `[Button]` instead of as noisy raw HTML, _and_ genuinely empty buttons remain visible in the markdown stream.

**Ask:** keep the final `return '[Button]'`. Revert the corresponding test change (`await _assert_markdown(user, ui.button, '[Button]')`).

### 2. Embedding arbitrary child markdown inside `[Button: ...]` is fragile — _major_

The children-fallback at `button.py:53-54` splices whatever `_children_to_markdown()` returns straight into the `[Button: ...]` wrapper. That works for the tested case (a single `ui.label('Custom content')`) because labels render as a flat string, but it falls apart as soon as the child is anything else:

- `with ui.button(): ui.link('NiceGUI', 'https://nicegui.io')` → `[Button: [NiceGUI](https://nicegui.io)]` — the link's `]` visually closes the wrapper.
- `with ui.button(): ui.label('Save'); ui.label('changes')` → `[Button: Save\n\nchanges]` — a `]` after a paragraph break is malformed.
- `with ui.button(): ui.table(...)` → embeds full table syntax inside `[...]`, which no parser will accept.

Complex buttons can't be reasonably rendered to markdown anyway, so the fallback shouldn't try.

**Ask:** restrict the children fallback to children that render as a single safe line — no newlines, no `[` / `]`. Otherwise drop through to the final `[Button]`. A guard like:

```python
if (children := self._children_to_markdown().strip()) and '\n' not in children and '[' not in children and ']' not in children:
    return f'[Button: {children}]'
```

would keep the simple win without producing malformed output. Add a test for at least one rejected case (e.g. a button containing a link).

### 3. Missing tests for the new priority order — _minor_

`tests/test_markdown_response.py:97-110` covers each branch in isolation but never asserts what wins when two are set. The new order — `label` > `aria-label` > `icon` > children — is the load-bearing decision of the PR, and a future refactor could swap two branches without any test going red.

**Ask:** add two assertions, e.g.

```python
await _assert_markdown(user, lambda: ui.button('hello').props('aria-label="world"'), '[Button: hello]')
await _assert_markdown(user, lambda: ui.button(icon='face').props('aria-label="Save"'), '[Button: Save]')
```

### 4. `.strip()` on `_children_to_markdown()` is inconsistent — _nit_

`button.py:53` strips; `expansion.py:49` and `item.py:55` do not. Either is defensible, but the divergence is new.

**Ask:** drop the `.strip()` (whitespace-only children are not a real scenario in this codebase), or — if it does turn out to matter — move it into `_children_to_markdown` itself so every caller benefits.

### 5. Style of new test helpers — _nit_

`aria_button` / `child_button` are inner `def`s only because `ui.button().props(...)` is two statements. Could be `lambda: ui.button().props('aria-label="Save"')` to match the file's one-liner style. Purely cosmetic.

falkoschindler (approved):
Reviewed and approving — summary of what landed during review (co-authored):

- **Merged `main`** so #6054 is in scope: phosphor copy-button children render empty via `override_markdown(..., '')` and are filtered out, so those buttons fall through to `[Button]` with no raw `<i class="ph-...">` leak.
- **`_render_markdown` gains two fallbacks** — `aria-label` (below `label`, above `icon`) and a conservative single plain line of child content. Empty / decorative buttons keep `[Button]` deliberately as a structural signal; dropping the token itself is out of scope.
- **Restored `.strip()`** (dropped in an earlier revert) so a whitespace-only child falls back to `[Button]` instead of `[Button:  ]`, and documented the guard.
- **Tests** pin the `aria-label` priority and every child-fallback branch: single-line content, nested-link, multi-child, and whitespace-only.
- **Title + description rewritten** to match the shipped behavior; the earlier "drop `[Button]` → `''`, 11 → 0" framing no longer applied after the revert.

Backward compatible (empty `ui.button()` still renders `[Button]`). `test_markdown_response.py` green; ruff/mypy clean. LGTM.

PR #6053, opened 2026-05-13
Serve markdown for known agent UAs sending wildcard `Accept`

evnchn:
_Drafted by @evnchn with Claude Code (Opus 4.7); diff and rendered copy reviewed before pushing. Approved triage in #6007 (comment); the highest-leverage piece of Phase 2 — it expands the agent population markdown actually reaches._

### Motivation

Phase 2 follow-up to #5889 (markdown content negotiation, merged 2026-04-24). Per the [follow-up comment in #6007](https://github.com/zauberzeug/nicegui/discussions/6007#discussioncomment-16747382), the on-the-wire shape of "real" agent traffic isn't what `curl -H 'Accept: text/markdown'` looks like. A probe at `wf-probe.evnchn.io/smart` captured two distinct Claude clients with measurably different requests:

| Client | User-Agent | Accept |
|---|---|---|
| **Claude Code (CLI)** | `Claude-User (claude-code/2.1.121; +https://support.anthropic.com/)` | `text/markdown, text/html, */*` |
| **claude.ai / Claude apps** | `Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Claude-User/1.0; +claude-user@anthropic.com)` | `*/*` |

Same vendor, same `Claude-User` token in the UA, but two materially different on-wire shapes. **#5889's `Accept`-driven negotiation works for Claude Code — and falls through to HTML for claude.ai's WebFetch.** Reasonable to assume the same split exists across other vendors: hosted browse-with-AI tools sending wildcard `Accept`, IDE/CLI integrations sending an explicit list.

This PR closes that gap with a UA-based safety net.

### Implementation

In `Client.build_response`:

```python
explicit_markdown = 'text/markdown' in accept and 'text/html' not in accept
wildcard_accept = not accept or accept.strip() == '*/*'
ua_fallback = wildcard_accept and 'text/html' not in accept and _looks_like_agent(user_agent)
if self.page.resolve_markdown() and (explicit_markdown or ua_fallback):
    # serve markdown
```

Plus a module-level regex covering known live-user agent fetchers (verified April 2026):

```python
_AGENT_UA_REGEX = re.compile(
    r'claude-?(bot|user|searchbot)|gptbot|oai-searchbot|chatgpt-user|'
    r'perplexity(bot|-user)|google-(cloudvertexbot|agent)|gemini-deep-research',
    re.IGNORECASE,
)
```

Sources cited inline as a comment so the next maintainer can refresh the list.

#### Behavior matrix

| Accept | UA | Markdown opted in? | Result |
|---|---|---|---|
| `text/markdown` | any | yes | markdown ✓ (existing) |
| `text/html` | any (incl. agent) | yes | HTML ✓ |
| `*/*` or absent | regular browser | yes | HTML ✓ |
| `*/*` or absent | known agent | yes | **markdown** ← new |
| `*/*` or absent | known agent | no | HTML ✓ (opt-in still respected) |

#### Two flags worth tracking

1. **The list keeps growing.** Google added `Google-Agent` (Project Mariner) in March 2026; OpenAI bumped `ChatGPT-User` from 1.0 → 2.0 → 3.0; Anthropic only formalized its three-bot split a few months ago, and Claude Code is itself a sub-variant within `Claude-User`. The cited comment block in `client.py` points at the source-of-truth pages so the next maintainer can refresh.
2. **Spoofing is real.** Perplexity in particular has been observed dropping its bot identifier and using Chrome-like UAs when robots.txt would block it. UA-based markdown serving is a rendering optimization, not a security boundary — fine here, but worth flagging if anyone later tries to extend the same logic into auth or rate-limiting.

#### E2E results

Verified against the local NiceGUI website on port 8806 with `markdown=True`:

| Request | Got |
|---|---|
| `Accept: */*`, `User-Agent: Mozilla/5.0 Chrome` | `text/html` ✓ |
| `Accept: */*`, `User-Agent: Claude-User/1.0` | `text/markdown` ✓ |
| no Accept, `User-Agent: ChatGPT-User/2.0` | `text/markdown` ✓ |
| `Accept: text/markdown` (no UA) | `text/markdown` ✓ |
| `Accept: text/html`, `User-Agent: Claude-User/1.0` | `text/html` ✓ (explicit wins) |

Plus 4 new pytest cases in `tests/test_markdown_response.py` covering each new branch.

### Progress

- [x] This PR does not address a security issue. (Rendering-only — not an auth or rate-limit boundary; see "spoofing is real" flag above.)
- [x] Documentation is not necessary (underlying agent-facing rendering change; the public `markdown=True` opt-in introduced in #5889 is unchanged).
- [x] No breaking changes to the public API — behavior only changes for clients that send `Accept: */*` (or no Accept) and a known-agent UA, on pages that opt in via `markdown=True`. Browsers and explicit `text/html` requests are unaffected.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -470,11 +470,10 @@ def _is_prefetch(request: Request) -> bool:
 def _did_user_request_markdown(request: Request) -> bool:
     """Whether the request prefers a markdown response over HTML (page opt-in checked separately)."""
     accept = request.headers.get('accept', '').strip().lower()
-    if 'text/html' in accept:
+    if 'text/html' in accept and 'text/markdown' not in accept:
         return False
-    if 'text/markdown' in accept:
+    if 'text/markdown' in accept and 'text/html' not in accept:
         return True
-    user_agent = request.headers.get('user-agent', '')
     AI_AGENT_PATTERNS = [
         r'claude-?(bot|user|searchbot)',
         r'gptbot|oai-searchbot',
@@ -483,4 +482,4 @@ def _did_user_request_markdown(request: Request) -> bool:
         r'google-(cloudvertexbot|agent)',
         r'gemini-deep-research',
     ]
-    return accept in {'', '*/*'} and bool(re.search('|'.join(AI_AGENT_PATTERNS), user_agent, re.IGNORECASE))
+    return bool(re.search('|'.join(AI_AGENT_PATTERNS), request.headers.get('user-agent', ''), re.IGNORECASE))
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import inspect
+import re
 import time
 import uuid
 from collections import defaultdict
@@ -153,9 +154,17 @@ class Client:
     def build_response(self, request: Request, status_code: int = 200) -> Response:
         """Build a FastAPI response for the client."""
         accept = request.headers.get('accept', '')
+        user_agent = request.headers.get('user-agent', '')
         # NOTE: This simple check doesn't handle quality values (q=) or wildcards (*/*).
         # It works for the real use case: agents sending exactly `Accept: text/markdown`.
-        if self.page.resolve_markdown() and 'text/markdown' in accept and 'text/html' not in accept:
+        explicit_markdown = 'text/markdown' in accept and 'text/html' not in accept
+        # UA-based fallback: clients like claude.ai's WebFetch send `Accept: */*` despite
+        # being agents (vs. Claude Code which sends `text/markdown, text/html, */*`).
+        # When the page opts in and the request looks like an agent with no explicit
+        # text/* preference, prefer markdown. Explicit `text/html` always wins HTML.
+        wildcard_accept = not accept or accept.strip() == '*/*'
+        ua_fallback = wildcard_accept and 'text/html' not in accept and _looks_like_agent(user_agent)
+        if self.page.resolve_markdown() and (explicit_markdown or ua_fallback):
             parts = []
             if title := self.resolve_title():
                 parts.append(f'# {title}')
@@ -467,3 +476,24 @@ class Client:
 def _is_prefetch(request: Request) -> bool:
     purpose = (request.headers.get('Sec-Purpose') or request.headers.get('Purpose') or '').lower()
     return 'prefetch' in purpose and 'prerender' not in purpose
+
+
+# Known live-user / agent fetchers (verified April 2026). Vendor docs change frequently;
+# refresh this list when new live-user agents are announced. Authoritative pages at the time
+# of writing: support.claude.com (ClaudeBot / Claude-User / Claude-SearchBot),
+# platform.openai.com/docs/bots (GPTBot / OAI-SearchBot / ChatGPT-User),
+# docs.perplexity.ai (PerplexityBot / Perplexity-User), Google's Search Central docs cover
+# Google-CloudVertexBot; Gemini-Deep-Research and Google-Agent (Project Mariner) are
+# documented in their respective product pages rather than on the common-crawlers page.
+# Substring match is intentional: vendors append version + embed-product suffixes
+# (e.g. `Claude-User (claude-code/2.1.121; ...)`).
+_AGENT_UA_REGEX = re.compile(
+    r'claude-?(bot|user|searchbot)|gptbot|oai-searchbot|chatgpt-user|'
+    r'perplexity(bot|-user)|google-(cloudvertexbot|agent)|gemini-deep-research',
+    re.IGNORECASE,
+)
+
+
+def _looks_like_agent(user_agent: str) -> bool:
+    """Heuristic: does the User-Agent header belong to a known live-user agent fetcher?"""
+    return bool(_AGENT_UA_REGEX.search(user_agent))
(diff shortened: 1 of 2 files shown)


nicegui/client.py
```diff
@@ -153,9 +154,17 @@ def __exit__(self, *_) -> None:
     def build_response(self, request: Request, status_code: int = 200) -> Response:
         """Build a FastAPI response for the client."""
         accept = request.headers.get('accept', '')
+        user_agent = request.headers.get('user-agent', '')
         # NOTE: This simple check doesn't handle quality values (q=) or wildcards (*/*).
         # It works for the real use case: agents sending exactly `Accept: text/markdown`.
```
HosseyNJF: Stale comment, should be removed

nicegui/client.py
```diff
@@ -153,9 +154,17 @@ def __exit__(self, *_) -> None:
...
-        if self.page.resolve_markdown() and 'text/markdown' in accept and 'text/html' not in accept:
+        explicit_markdown = 'text/markdown' in accept and 'text/html' not in accept
+        # UA-based fallback: clients like claude.ai's WebFetch send `Accept: */*` despite
+        # being agents (vs. Claude Code which sends `text/markdown, text/html, */*`).
+        # When the page opts in and the request looks like an agent with no explicit
+        # text/* preference, prefer markdown. Explicit `text/html` always wins HTML.
+        wildcard_accept = not accept or accept.strip() == '*/*'
+        ua_fallback = wildcard_accept and 'text/html' not in accept and _looks_like_agent(user_agent)
+        if self.page.resolve_markdown() and (explicit_markdown or ua_fallback):
```
HosseyNJF: Nitpick: the logic seems a bit scattered across the `client.py` module. My proposal is to extract this logic into another method on the `Client` class called `_did_user_request_markdown`, and also run the regex there so there would be no need for a module-level `_looks_like_agent` method.

```suggestion
        if self.page.resolve_markdown() and self._did_user_request_markdown(request):
```

evnchn:
@HosseyNJF Very sharp eyes indeed. If you wanna do stuff feel free to send PRs to my fork or send patches my way. 

Or if you want to strategically stay in review-land so that you aren't both reviewing and coding is also OK.

HosseyNJF:
Sounds good, with your approval I'll go ahead and fix the issues. I usually don't touch other people's branches unless it's consented to :)

evnchn:
@HosseyNJF Please feel free to touch away. I also think I can then cross-review your work (basically whoever code and whoever review can't be the same person for max effectiveness)

falkoschindler:
Heads up, @evnchn, I've taken over a few small bits here to get this ready for the **3.13 release this afternoon**. Hope that's alright; happy to hand anything back if you'd rather own it.

### What I changed

Mostly tidy-up around the content negotiation, folding in @HosseyNJF's review notes:

- Extracted the `Accept` / `User-Agent` logic out of `Client.build_response` into a small free function `_did_user_request_markdown(request)`, sitting next to `_is_prefetch` (same shape — stateless, header-only). The call site is now just `if self.page.resolve_markdown() and _did_user_request_markdown(request):`.
- Dropped the now-redundant `'text/html' not in accept` clause, the stale `NOTE` comment (it claimed wildcards weren't handled — they are now), and the separate `_looks_like_agent` indirection.
- Normalize the header once (`.strip().lower()`) so the media-type checks are case-insensitive.

No behavior change; the 36 tests still pass.

### One thing I'd like to hear your thoughts — the Claude Code case

While reviewing I noticed the flagship client doesn't actually get markdown. Per your probe, Claude Code (CLI) sends:

```
Accept: text/markdown, text/html, */*
```

Because we check `text/html` first (originally the `'text/html' not in accept` guard), that request resolves to **HTML**, not markdown — even though `text/markdown` is listed first. So the new UA path only fires for the pure-`*/*` claude.ai shape; Claude Code itself still falls through to HTML, both before and after this PR. That seems to undercut the "works for Claude Code" framing, since it's exactly the agent we'd most want to reach.

**Possible fix** — check `text/markdown` before `text/html`:

```python
if 'text/markdown' in accept:
    return True
if 'text/html' in accept:
    return False
```

Rationale: browsers never send `text/markdown`, so its presence is an unambiguous opt-in — no reason a co-listed `text/html` should veto it. This flips only the "both listed" case (i.e. Claude Code) to markdown; browsers and html-only requests are unaffected, and the existing tests still pass. The trade-off is that it refines the rule from "explicit `text/html` always wins" to "explicit `text/markdown` wins over `text/html`," so we'd tweak the behavior matrix in the description accordingly.

Intended to leave Claude Code on HTML, or shall I push the reorder + a test for that header so it lands in 3.13?

falkoschindler (approved):
Ok, I think this is a good solution. I also consolidated the tests quite a bit.
Ready to merge! 🚀

evnchn:
I agree with the changes. In general we won't quite know how the scrapers evolve and have no choice but to change our code to accommodate. 

On Claude Code, my experience is that, with web fetch, no matter what, Claude Code gets a digest. Claude Code as a more general agent can use headers as it wishes once it decides to want a non-digest.

nicegui/client.py
falkoschindler: I went ahead and introduced such a `_did_user_request_markdown` function. 👍🏻

Issue #6048, opened 2026-05-12
llms.md: gaps in element surface vs. ui.* exports (follow-up to #6021)


evnchn:
_Posted on behalf of @evnchn via Claude Code (Opus 4.7) — claims cross-checked against `origin/main` at 79161287._

## TL;DR

Fresh `nicegui/llms.md` from #6021 reads well, but the Element Reference tables and `ui.run`/`app.*` examples miss a handful of exported public surfaces that LLMs commonly need. Opening this as a punch-list so a follow-up can decide what's worth surfacing vs. intentionally trimmed.

<details>
<summary><b>Exported by <code>nicegui/ui.py</code> but not mentioned anywhere in <code>llms.md</code></b></summary>

| Symbol | What it is | Why an LLM is likely to ask |
|---|---|---|
| `ui.button_group` | Quasar QBtnGroup | Common grouping primitive |
| `ui.dropdown_button` | Split/dropdown button | Common UI primitive |
| `ui.fullscreen` | Fullscreen control (added 2.11.0) | Frequently asked |
| `ui.list` + `ui.item` + `ui.item_label` + `ui.item_section` | Quasar QList/QItem | No Q-list pattern is shown at all |
| `ui.keep_alive` | Vue `<keep-alive>` wrapper | SPA perf escape hatch |
| `ui.link_target` | In-page anchor target | Pairs with documented `ui.link` |
| `ui.parallax` | Parallax effect | Niche but exported |
| `ui.circular_progress` / `ui.linear_progress` | Progress indicators | Only `spinner`/`skeleton` are mentioned today |
| `ui.scene_view` | Second camera on a `ui.scene` | Only way to reuse a scene |
| `ui.slide_item` | Swipeable list item | Mobile pattern |
| `ui.timeline` / `ui.timeline_entry` | Quasar QTimeline | Activity feeds |
| `ui.run_with` | Mount NiceGUI inside an existing FastAPI app | Integration-critical |
| `ui.element('div', …)` | Generic escape hatch for arbitrary tags | Used in many recipes |
| `ui.context` | Current client/page context | Useful in utilities |

</details>

<details>
<summary><b><code>app.*</code> helpers used in real code but not in the doc</b></summary>

The current doc tells LLMs to use `app.mount('/static', StaticFiles(...))`, which works, but skips NiceGUI's first-class helpers:

- `app.add_static_files(url_path, local_dir)` / `app.add_static_file(...)`
[...]

PR #6055, opened 2026-05-14
Skip outbox echo when server stored value matches client emit

### LINKED ISSUE #5185: Character loss in ui.input validation with high network latency

evnchn:
*Authored on Evan's behalf by Claude Code (Opus 4.7, 1M-context). Reviewed and approved by Evan before submission.*

### Motivation

LOOPBACK=True `ValueElement` subclasses (checkbox, switch, dialog, menu, expansion, select, radio, toggle, slider, …) currently echo every client-originated value change: `handle_change → set_value → _handle_value_change → update() → outbox sends the full element state back, including the value the client just emitted`. The echo is redundant — the client already has that value — and on high-latency links it races user input (see #5185).

Implements the approach @falkoschindler proposed in the review of #5728: *"don't resend values that came from the client."* Supersedes the closed #5933, which attempted the same goal via wire-format stripping + a JS prop-merge but broke Quasar components using `model-value` as UI state and would have silently dropped every server-initiated prop *removal* via the merge contract.

### Implementation

**Python** (`nicegui/elements/mixins/value_element.py`) — `handle_change` suspends the inner `update()` during `set_value`, then conditionally fires it only when `_props[VALUE_PROP]` diverges from `e.args` (the raw client emit). Transforms via `_value_to_model_value`, `on_change` handler corrections, and unrelated prop mutations still propagate because their post-state differs from `e.args` or because they enqueue through their own `update()` calls.

**JavaScript** (`nicegui/static/nicegui.js`) — extend the existing LOOPBACK=False local-prop-update branch to also cover LOOPBACK=True. The client-side mirror is what makes the server's elided echo safe: Vue's next re-render reads the locally-stored `model-value` instead of the stale one. Also fixes a latent shape bug in the local assignment: `args[0]` (the emitted scalar) instead of `args` (the rest-parameter array). The array form happened to work for `ui.number` because Quasar's `QInput` coerces, but would have broken for booleans on the new path.

`_to_dict` is unchanged → reconnections still receive the authoritative full snapshot including `VALUE_PROP`.

LOOPBACK=False (`ui.input`, `ui.number`, …) and LOOPBACK=None (`ui.fullscreen`, codemirror) behavior is preserved — the final `if self.LOOPBACK is True` guard scopes the new echo path to elements that previously echoed.

<details>
<summary>Why <code>try/finally</code> around <code>set_value</code> (pre-empting the "looks defensive, drop it?" question)</summary>

```python
self._send_update_on_value_change = False     # transient suppress
try:
finally:
    self._send_update_on_value_change = True
```

The flag is an instance attribute that persists across calls. If `set_value` raises — a user-overridden `_value_to_model_value`, a binding peer's setter, or a synchronous `on_change` handler — `handle_event` catches it at the boundary, but the flag stays at `False`. Every subsequent server-driven `element.value = X` then silently skips its `update()` call, killing the outbox for that element's lifetime.

The pre-PR code didn't need `try/finally` because it assigned `self._send_update_on_value_change = self.LOOPBACK is True` on entry — the safe end-state. This PR enters with the unsafe transient state, so the restore line becomes load-bearing. Covered by `test_set_value_exception_does_not_leak_suppression` (mutation-verified: removing `try/finally` makes that test fail).

</details>

Closes #5933. Tracks #5185.

### Progress

### DIFF
--- a/nicegui/elements/mixins/value_element.py
+++ b/nicegui/elements/mixins/value_element.py
@@ -15,7 +15,7 @@ class ValueElement(Element, Generic[ValueT]):
     LOOPBACK: bool | None = True
     '''Whether to set the new value directly on the client or after getting an update from the server.
 
-    - ``True``: The value is updated by sending a change event to the server which responds with an update.
+    - ``True``: The value is updated by sending a change event to the server; the server echoes back unless the value matches what the client just sent.
     - ``False``: The value is updated by setting the VALUE_PROP directly on the client.
     - ``None``: The value is updated automatically by the Vue element.
     '''
@@ -43,9 +43,14 @@ class ValueElement(Element, Generic[ValueT]):
             self.on_value_change(on_value_change)
 
         def handle_change(e: GenericEventArguments) -> None:
-            self._send_update_on_value_change = self.LOOPBACK is True
-            self.set_value(self._event_args_to_value(e))
-            self._send_update_on_value_change = True
+            self._send_update_on_value_change = False
+            try:
+                self.set_value(self._event_args_to_value(e))
+            finally:
+                self._send_update_on_value_change = True
+            # skip the echo when the server's value matches what the client locally rendered (see #5933)
+            if self.LOOPBACK is True and self._props[self.VALUE_PROP] != e.args:
+                self.update()
         self.on(f'update:{self.VALUE_PROP}', handle_change, [None], throttle=throttle)
 
     def on_value_change(self, callback: Handler[ValueChangeEventArguments[ValueT]]) -> Self:
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -272,8 +272,10 @@ function renderRecursively(elements, id, propsContext) {
         else setTimeout(delayed_emitter, 10);
       };
       throttle(delayed_emitter, event.throttle, event.leading_events, event.trailing_events, event.listener_id);
-      if (element.props["loopback"] === False && event.type == "update:modelValue") {
-        element.props["model-value"] = args;
+      // mirror locally so the client-rendered value survives re-render when the server elides the echo (see #5933)
+      if ((element.props["loopback"] === true || element.props["loopback"] === false)
+          && event.type == "update:modelValue") {
+        element.props["model-value"] = args[0];
       }
     };
 
--- a/tests/test_value_echo_skip.py
+++ b/tests/test_value_echo_skip.py
@@ -1,0 +1,175 @@
+"""Tests for the client-value echo skip optimization in ValueElement (see #5933)."""
+
+from collections.abc import Callable
+
+import pytest
+
+from nicegui import ui
+from nicegui.elements.mixins.value_element import ValueElement
+from nicegui.helpers import event_type_to_camel_case
+from nicegui.testing import User
+
+
+class PlainValue(ValueElement[str]):
+
+    def __init__(self, *, value: str = '', on_change: Callable | None = None) -> None:
+        super().__init__(tag='div', value=value, on_value_change=on_change)
+
+
+class TruncatingValue(ValueElement[str]):
+
+    def __init__(self, *, value: str = '') -> None:
+        super().__init__(tag='div', value=value)
+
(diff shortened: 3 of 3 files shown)


evnchn:
*Authored on Evan's behalf by Claude Code (Opus 4.7, 1M-context).*

**Force-pushed `5586b365` → `a1e233b5` (CI fix + missing Claude co-author trailer).**

PR is ~25 min old, no reviews yet, single-commit branch — force-push won't disrupt review threads.

<details>
<summary>Range-diff (verbatim)</summary>

```
1:  5586b365 ! 1:  a1e233b5 Skip outbox echo when server stored value matches client emit
    @@ Commit message

         Closes #5933. Tracks #5185.

    +    Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>
    +
      ## nicegui/elements/mixins/value_element.py ##
     @@ nicegui/elements/mixins/value_element.py: class ValueElement(Element, Generic[ValueT]):
          LOOPBACK: bool | None = True
    @@ tests/test_value_echo_skip.py (new)
     @@
     +"""Tests for the client-value echo skip optimization in ValueElement (see #5933)."""
     +
    -+from typing import Callable
    ++from collections.abc import Callable
     +
     +import pytest
     +
    @@ tests/test_value_echo_skip.py (new)
     +    event_name = event_type_to_camel_case(f'update:{element.VALUE_PROP}')
     +    listeners = element._event_listeners  # pylint: disable=protected-access
     +    try:
    -+        listener_id = next(lid for lid, l in listeners.items() if l.type == event_name)
    ++        listener_id = next(lid for lid, listener in listeners.items() if listener.type == event_name)
     +    except StopIteration:
    -+        types = sorted({l.type for l in listeners.values()})
    ++        types = sorted({listener.type for listener in listeners.values()})
     +        raise AssertionError(f'no {event_name!r} listener on {element!r}; have: {types}') from None
     +    element._handle_event({'listener_id': listener_id, 'args': value})  # pylint: disable=protected-access
```

The only payload-line changes are the ruff fixes in the test file; the implementation diff (`value_element.py`, `nicegui.js`) is byte-identical (`=` lines elided above for brevity).

</details>

**What changed and why:**

- **`tests/test_value_echo_skip.py`** — fixes `check / pre-commit`:
  - ruff E741: rename ambiguous `l` → `listener` (the generator + set-comprehension inside `_simulate_client_change`).
  - ruff auto-fix on top: `from typing import Callable` → `from collections.abc import Callable` (deprecated import).
  - Verified locally: `pre-commit run --files tests/test_value_echo_skip.py` clean, 14/14 tests pass.

- **Commit message** — appended `Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>` to match the convention I've used on prior PRs (#6048 etc.). The trailer was missing on the original push — flagged as a repeat lapse to fix.

evnchn:
Defer since messing with the outbox late in 3.13 milestone is very risky

evnchn:

Closing this. The real driver for #6055 was shaving **INP** (Interaction to Next Paint) — skipping the redundant server echo for `LOOPBACK is True` elements (checkbox / switch / select / toggle) means less work per interaction. But after the docs-site redesign the docs' INP is in good shape, so that pressure is gone. Small perf gains like this aren't a mainline concern — and #6104 already covers the actually-reported correctness case (#5185, `ui.input`). If it's worth chasing later it belongs on my fork (the rebased work is preserved at evnchn/nicegui#177), not as core outbox-semantics churn.

---
**Changelog**
- 2026-06-14: rewrote the closure rationale to lead with the real motivation — #6055 was an INP/perf optimization, now moot since the docs-site redesign brought INP into good shape; small perf gains aren't a mainline concern (chase on fork if needed). Supersedes the earlier "rare race / outbox surface" framing.

PR #6056, opened 2026-05-14
Fix run.cpu_bound/io_bound return type to R | None

evnchn:
### Motivation

`run.cpu_bound` and `run.io_bound` are annotated as returning `R`, but at runtime both silently return `None` when the call is cancelled or the app is shutting down (suppressed `# type: ignore` markers in the existing code). This is the bug reported in #5925 — downstream code that trusts the `R` return type hits `TypeError`s on `None` when shutdown races a `run.*_bound` call.

The earlier attempt to fix this in #5926 used a `strict: bool` kwarg with overloads to opt into raise-on-cancel. Two backwards-compat blockers killed that approach:

1. **`strict` is stolen from `P.kwargs`** — any callback with its own `strict` kwarg (SQLAlchemy, Pydantic, JSON schema validators, etc.) silently loses that argument. Structural to "steal a kwarg from `P.kwargs`"; no chosen kwarg name is safe.
2. **Switching the default to "raise on cancel"** is a runtime breaking change for 3.x.

There's also a hard typing constraint: PEP 612 forbids keyword args between `*args: P.args` and `**kwargs: P.kwargs`, so the interleaved `strict` overloads in #5926 fail mypy with `Arguments not allowed after ParamSpec.args` (also see [python/typing#1009](https://github.com/python/typing/issues/1009), [discuss.python.org](https://discuss.python.org/t/using-concatenate-and-paramspec-with-a-keyword-argument/32283)).

### Agreed direction

Per the [review discussion](https://github.com/zauberzeug/nicegui/pull/6056#issuecomment-4707711318): **NiceGUI 4.0 will raise `CancelledError` on cancel/shutdown** instead of returning `None`, making the return type honestly `R` again and matching normal `await` semantics — the silent-`None` sentinel was the actual root bug.

Given that, this PR is **trimmed to the minimum-correct interim kernel** for the 3.x window, so #5925's silent-`None` `TypeError`s stop being invisible — without adding public surface we'd deprecate at 4.0.

### Implementation

**Type-only correction** (runtime unchanged):
- `_run`, `cpu_bound`, `io_bound` now annotate `R | None` to match runtime reality. The `# type: ignore` markers go away.
- `nicegui/welcome.py` now `or []`-guards `run.io_bound` in `collect_urls`.
- Docstrings on `cpu_bound`/`io_bound` note the `None` return is **interim** — 4.0 will raise instead — so callers writing `if result is None: ...` know it's a temporary shape.
- New test `test_returns_none_when_app_is_stopping` (parametrized over `cpu_bound` + `io_bound`).

**Dropped from the earlier revision of this PR** (kept on record in the thread for anyone who needs the disambiguation pattern in the 3.x window): the `wrap_none` / `NONE_MARKER` sentinel helper, its demo in `section_action_events.py`, and `test_wrap_none_disambiguates`. With 4.0 raising on abort, a plain `None` will unambiguously mean "the callback returned `None`", so the wrapper has nothing left to do — not worth shipping two public names with a ~5-month shelf life.

### Release note

The annotation change from `R` to `R | None` is **type-level only; runtime behavior is unchanged.** mypy may flag new errors in user code that previously typed a `run.cpu_bound` / `run.io_bound` result as non-Optional `R`. Those errors are **surfaced latent bugs, not regressions** — the value could already be `None` at runtime on cancel/shutdown (#5925). The migration is a `None`-check now; at 4.0 it becomes a `try/except CancelledError`.

### Progress

- [x] Pytests have been added (`test_returns_none_when_app_is_stopping`, parametrized over `cpu_bound` + `io_bound`).
- [x] Documentation has been updated (interim-`None` note in the `cpu_bound`/`io_bound` docstrings).
- [x] About breaking changes to the public API: Type-level breaking (annotation now `R | None`); runtime behavior unchanged. See **Release note** above.

---

Closes #5925.
Alternative to #5926 — see that thread for the design discussion.

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -1,4 +1,5 @@
 import asyncio
+import enum
 import logging
 import signal
 import traceback
@@ -8,7 +9,7 @@ from concurrent.futures.process import BrokenProcessPool
 from contextlib import suppress
 from functools import partial
 from pickle import PicklingError
-from typing import Any, TypeVar
+from typing import Any, Generic, TypeVar
 
 from typing_extensions import ParamSpec
 
@@ -19,6 +20,46 @@ thread_pool = ThreadPoolExecutor()
 
 P = ParamSpec('P')
 R = TypeVar('R')
+
+
+class _Sentinel(enum.Enum):
+    NONE_MARKER = 'none_marker'
+
+
+NONE_MARKER = _Sentinel.NONE_MARKER
+'''Sentinel returned by :class:`wrap_none`-wrapped callbacks when they legitimately return ``None``.
+
+Use ``result is run.NONE_MARKER`` to detect this case. Plain ``result is None`` then means the
+``cpu_bound`` / ``io_bound`` call was cancelled or the app is shutting down.
+'''
+
+
+class wrap_none(Generic[P, R]):  # pylint: disable=invalid-name
+    """Wrap a callable so that a legitimate ``None`` return becomes :data:`NONE_MARKER`.
+
+    Use to distinguish "the callback returned ``None``" from "``cpu_bound`` / ``io_bound`` aborted"
+    when both can occur::
+
+        result = await run.cpu_bound(run.wrap_none(my_func), arg1)
+        if result is None:
+            ...  # run was cancelled / app shutting down
+        elif result is run.NONE_MARKER:
+            ...  # my_func returned None
+        else:
+            ...  # real value
+
+    Implemented as a module-level callable class (not a closure) so the wrapper itself is picklable
+    and survives ``cpu_bound``'s process boundary.
+    """
+
+    def __init__(self, fn: Callable[P, R]) -> None:
+        self.fn = fn
+
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R | _Sentinel:
+        result = self.fn(*args, **kwargs)
+        if result is None:
+            return NONE_MARKER
+        return result
 
 
 def setup() -> None:
@@ -58,9 +99,9 @@ def safe_callback(callback: Callable, *args, **kwargs) -> Any:
         raise SubprocessException(type(e).__name__, str(e), traceback.format_exc()) from None
 
 
-async def _run(executor: Any, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
+async def _run(executor: Any, callback: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R | None:
     if core.app.is_stopping:
-        return  # type: ignore  # the assumption is that the user's code no longer cares about this value
+        return None
     try:
         loop = asyncio.get_running_loop()
         return await loop.run_in_executor(executor, partial(callback, *args, **kwargs))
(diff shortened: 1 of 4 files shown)


evnchn:

@falkoschindler — picking up the thread you pointed here from #5926. The reasoning's in the PR description so I won't restate it; the one decision that matters is whether **4.0 should raise on cancellation** instead of returning `None`. If that change is coming, users are better off knowing now than finding out at the 4.0 cut — so I'd rather lock the direction early even though the major's a way off. Could you take a look when you have a slot?

falkoschindler:
@evnchn — thanks for the thorough writeup, the `P.kwargs`/PEP 612 analysis on why `#5926`'s `strict` approach can't work is convincing, and I went through it with that as the settled premise.

On your one decision-that-matters question: **yes, let's commit to raising on cancel/shutdown in 4.0** (≈5 months out). Re-raising the swallowed `CancelledError` (and raising it for the `is_stopping` shutdown case too) makes the return type honestly `R` again and matches normal `await` semantics — the silent-`None` sentinel was the actual root bug, not the missing escape hatch.

Given that direction, I'd like to **trim this PR to the interim kernel** rather than ship `wrap_none`. The reasoning:

`wrap_none`/`NONE_MARKER` only exist to disambiguate "callback returned `None`" from "call was aborted". Once 4.0 raises, abort is an exception, so a plain `None` unambiguously means the callback returned `None` — and the wrapper has nothing left to do. Shipping two new public names (plus the sentinel enum, the demo, and the `pylint: disable=invalid-name`) that we'd deprecate at the next major is surface I'd rather not take on. It's cleanly implemented — this isn't about the code quality, just the shelf life.

What I'd **keep** from the PR (it's genuinely valuable for the 3.x window so #5925's silent-`None` `TypeError`s stop being invisible):

- `R | None` on `_run`/`cpu_bound`/`io_bound` + dropping the `# type: ignore` markers
- the `welcome.py` `or []` guard
- `test_returns_none_when_app_is_stopping`

What I'd **drop**:

- `wrap_none`, `NONE_MARKER`, `_Sentinel`
- the `section_action_events.py` demo
- `test_wrap_none_disambiguates`

Two small additions to make the 4.0 cut clean:

- A docstring line on `cpu_bound`/`io_bound` noting the `None` return is interim and 4.0 will **raise** instead — so anyone writing `if result is None: ...` knows it's a temporary shape (the current docstrings read as if `None` is permanent).
- A changelog/release note framing the new mypy errors as surfaced latent bugs, not regressions.

I know this means callers migrate twice (`if result is None` now → `try/except CancelledError` at 4.0), but I don't see a clean way around it — the early-opt-in lever was exactly the `strict=` kwarg we ruled out, and a `DeprecationWarning` on the cancel/shutdown path would be noisy and unactionable mid-shutdown. The two-step feels like the honest cost of having shipped the silent `None` in the first place.

Does trimming to the kernel work for you? Happy to keep the `wrap_none` design on record in this thread for whoever needs the disambiguation pattern in the meantime.

evnchn:
I agree with you but can only trim tomorrow.

I think `wrap_none` does constitute "well known but advanced use case" which we are not obliged to document in full writing and provide a dedicated wrapper.

evnchn:
@falkoschindler dropped the stuff as established.

The helper functions still exist as interim commits in this PR and can be copy-pasted if needed.

Maybe another pass of review on the core code that's kept after the removal to prevent issues being masked, but other than that looks pretty good.

falkoschindler (approved):
LGTM. The `R | None` annotation finally matches the runtime reality from #5925, dropping the `# type: ignore` markers is the right cleanup, and the `welcome.py` `or []` guard plugs the one internal caller that would have hit `set(None)` on shutdown. Trimming to the type-only interim kernel (and deferring the raise-on-cancel break to 4.0) is the right call for the 3.x window.

Let's merge! 🚀

PR #6057, opened 2026-05-14
Add /sitemap.xml endpoint with opt-in @ui.page integration

evnchn:

Tracking: evnchn/nicegui#151 (Layer 1 — sitemap groundwork). Sized to be Falko-friendly per the issue's "small focused PR" rule.

### Motivation

Sitemaps are the cheapest SEO win available — Google's docs explicitly favor them when site structure is non-trivial — but every NiceGUI app today has to hand-roll one. `https://nicegui.io/sitemap.xml` and every downstream sibling currently 404. This adds a framework-level endpoint without auto-publishing anything that wasn't already publicly known, so apps using long unguessable URLs as soft auth keep that invariant on upgrade.

#### Opt-in per page

```python
@ui.page('/', sitemap=True)
def home(): ...

@ui.page('/docs', sitemap={'changefreq': 'daily', 'priority': 0.9})
def docs(): ...
```

`sitemap: bool | dict = False`. Default `False` is load-bearing — NiceGUI does not enumerate `@ui.page` paths anywhere public today: `App(openapi_url=None)` (`app/app.py:39`), `endpoint_documentation='none'` default (`app_config.py:10`, `page.py:128`), `Client.page_routes` is internal-only (used only by `link.py`, `navigate.py`, `client.py:275`, startup logging in `ui_run.py`). Auto-publishing on upgrade would silently leak any page whose URL was the secret.

#### Hand-curated entries

For URLs that don't map one-to-one to a decoration — `ui.sub_pages` routes (per-request, not knowable at startup), parameterized `@ui.page('/docs/{name}')` routes (no enumerator), or anything else — iterate your own index and call `app.sitemap.add(...)`:

```python
@ui.page('/documentation/{path:path}')
def docs(): ...

for name in documentation_registry:
    app.sitemap.add(f'/documentation/{name}', changefreq='weekly')
```

#### Single validation point

`Sitemap.add()` rejects parameterized paths and unknown sitemap fields with `ValueError` and a message pointing at the correct escape hatch. Both `@ui.page(sitemap=...)` and direct `app.sitemap.add(...)` route through it, so there's exactly one DRY check site — no decorator-vs-direct-API skew, no silent filtering.

```
ValueError: Cannot add parameterized path '/users/{id}' to the sitemap: there is no enumerator.
            Add a concrete path per URL instead (e.g. app.sitemap.add('/users/alice')).

ValueError: Unknown sitemap field(s) for '/about': ['lastmode'].
            Supported: ['lastmod', 'changefreq', 'priority'].
            See https://www.sitemaps.org/protocol.html
```

#### Canonical URL

```python
app.sitemap.base_url = 'https://nicegui.io'   # production pin
```

If unset, the URL is derived from the request (correct in dev and behind any reverse proxy that populates uvicorn's standard headers — `proxy_headers=True` is uvicorn's default, so `request.url.scheme` and `request.url.netloc` already reflect `X-Forwarded-Proto`/`X-Forwarded-Host` at the ASGI layer; re-reading them in FastAPI would be redundant and create a spoofing surface on deployments without `forwarded_allow_ips` whitelisting). The endpoint additionally honors `X-Forwarded-Prefix` + ASGI `root_path` to match the existing NiceGUI convention (`client.py:172`, `middlewares.py:12`, `sub_pages_router.py:26`).

### Testing

18 tests in `tests/test_sitemap.py`, 0.27 s total, three browser-free tiers:

| Tier | Fixture | What it covers |
|------|---------|----------------|
| Unit | none — instantiate `Sitemap()` directly | `entries()`, `add()`/`remove()` round-trip, metadata override, `to_xml()` namespace + XML escaping, parameterized + unknown-field rejection |
| Decorator | `nicegui_reset_globals` (pure-Python reset) | `sitemap=False` default no-op, `sitemap=True`, `sitemap={dict}`, retract via re-decoration, `ValueError` on parameterized opt-in, `ValueError` on unknown fields |
| Endpoint | `nicegui_reset_globals` + `fastapi.testclient.TestClient(core.app)` (same pattern as `tests/test_run_with.py`) | XML+absolute URLs, `base_url` override, `X-Forwarded-Prefix`, nicegui.io documentation composition |

### Out of scope (deferred per evnchn/nicegui#151)

- `/robots.txt` pointing at `Sitemap:`
- Per-page `<link rel="canonical">`
- Per-page `<meta description>` mechanism
- Sourcing `lastmod` from git or filesystem

### Progress

- [x] Documentation has been added (docstrings on `Sitemap`, `Sitemap.add`, and `@ui.page`'s `sitemap` parameter).


evnchn:
*Drafted by Claude Code on @evnchn's behalf.*

Hi @falkoschindler — for context if useful:

- **Evolution / design discussion**: [evnchn/nicegui#152](https://github.com/evnchn/nicegui/pull/152) (closed now that this PR is open) — went through several rounds of review on the fork covering the auto-include vs opt-in question, X-Forwarded-* layering, parameterized-path handling, the sub_pages pattern, and the test-fixture tier rewrite. Threaded review comments there if you want to see where each decision came from.
- **SEO roadmap**: [evnchn/nicegui#151](https://github.com/evnchn/nicegui/issues/151) — the broader two-layer SEO plan (groundwork + content optimization) this PR fits into. This PR is the first ship-now item from Layer 1; /robots.txt, per-page canonical, and meta-description mechanism are deferred items tracked there. No commitment expected from you on the roadmap itself — it's there so you can see what's downstream of this if you have an opinion.

evnchn:
Defer since API design not quite finalized

evnchn:

@falkoschindler — this is currently milestoned for **3.14** (due 2026-06-30), but I'd deferred it back in June pending API-design sign-off. Since it adds a parameter to the `@ui.page` signature (`sitemap: bool | dict`) — a fairly load-bearing piece of public API — it really wants your eyes before it can land.

Do you want it in 3.14? If so, a review in the next few days would be the deciding factor (both on the opt-in API shape and the implementation). If you'd rather not rush a public-API addition, I'm happy to bump it to 3.15 so it isn't blocking the milestone — your call.

falkoschindler (changes requested):
Thanks for the pull request, @evnchn! Nice, focused feature — a framework-level `/sitemap.xml` with per-page opt-in, and defaulting `sitemap=False` so apps relying on unguessable URLs don't leak is the right call. The design is sound and well-tested. Five findings, most-severe first (3 Medium, 1 Plausible, 1 Low):

1. Default `sitemap=False` is an active removal, not a no-op · **Medium**

   The default branch runs `core.app.sitemap.remove(self.path)`, so a plain `@ui.page` _retracts_ a matching entry instead of leaving it alone. A manual `app.sitemap.add('/about')` that runs before a plain `@ui.page('/about')` is silently wiped at import time — order-dependent across modules, with no warning.

   **Fix:** make `False` a no-op; only remove when the path was previously opted in via the decorator (keeps the `test_sitemap_false_after_true_retracts` case).

2. `priority` silently corrupted and never validated · **Medium**

   Emitted as `f'{entry.priority:.1f}'` with no range or type check at `add()`: `priority=0.85` → `<priority>0.8</priority>`, out-of-range like `5.0` is emitted verbatim (crawlers reject it), and a non-numeric value (e.g. read from JSON config as a string) blows up at request time in `to_xml()` with a 500.

3. `ET.register_namespace` clobbers other code's default-namespace registration · **Medium**

   `register_namespace('', _NS)` writes to ElementTree's process-global `_namespace_map`, and its contract is destructive: it deletes any existing entry whose prefix is also `''`. So every `/sitemap.xml` request steals the default-namespace slot from whatever else in the process was using it — an app that does `ET.register_namespace('', 'urn:my-app')` flips from clean output to `ns0:`-prefixed tags after the first crawler hit. Remotely triggered, and violates AGENTS.md → "Global mutable state in library code".

   ```
   before a /sitemap.xml request:  <report xmlns="urn:my-app">…
   after  a /sitemap.xml request:  <ns0:report xmlns:ns0="urn:my-app">…
   ```

   **Fix:** don't touch the global registry — set `xmlns` as a literal attribute on the root and emit plain tag names (`ET.Element('urlset', xmlns=_NS)`, then `ET.SubElement(url, 'loc')`). Verified byte-identical output with `_namespace_map` untouched. (Hoisting the call to import time only narrows the race; it doesn't stop the clobbering.)

4. Router prefix captured at decoration time · Plausible (deployment-dependent)

   The entry path is `api_router.prefix + self._path`, read at decoration time. That captures a prefix set on `APIRouter(prefix='/sub')` (construction — the documented pattern), but not one passed later to `app.include_router(router, prefix='/extra')`, which FastAPI applies when copying routes into the app without touching `router.prefix`. So such a page serves at `/extra/y` while the sitemap still lists `/y` → 404. Verified: construction-time prefix matches, include-time prefix drifts.

5. Tests are well-organized but over-tested, and miss the risky paths · Low

   `tests/test_sitemap.py` (~184 lines) splits cleanly into three tiers and runs fast, but the validation rules (parameterized-path + unknown-field) are tested both at the unit layer and again through the decorator, even though the design routes both through a single `add()` site — the decorator side only needs one test proving it reaches `add()`, not a re-run of each rule. A couple are near-trivial (`test_sitemap_is_empty_by_default`, and `test_to_xml_escapes_special_characters`, which really tests ElementTree's stdlib escaping), and `test_endpoint_reflects_documentation_pattern` re-covers pieces tested elsewhere. More importantly, despite the length the suite misses every edge case above: it tests true→false retraction but not add-then-plain-decorator (finding 1), and nothing covers priority precision (2), namespace clobbering (3), or include-time prefix (4). Suggest trimming to ~130–140 lines and adding coverage for those instead.

evnchn:

**TL;DR: pushed `2ff0e34b` addressing all five points.**

- **(1) retraction** — now provenance-tracked: `@ui.page(sitemap=False)` only retracts a path a *decorator* opted in; a manual `app.sitemap.add()` is never wiped, in any ordering (including manual-add-after-decorator-opt-in).
- **(3) global mutation** — `xmlns` is now a literal root attribute; no more `ET.register_namespace()` clobbering ElementTree's process-global prefix map.
- **(2) priority** — validate + clamp to `[0,1]` with a `log.warning`; non-numeric / non-finite omitted; lossy `:.1f` gone (`0.85` stays `0.85`). *I chose warn-don't-raise at render time — say the word if you'd prefer raise-at-`add()`.*
- **(4) include-prefix drift** — documented as a limitation on the `sitemap` param, pointing users to `app.sitemap.add('/full/path', ...)` (the eventual mount path isn't knowable at decoration time without an extra param).
- **(5) tests** — added provenance / namespace / priority edge tests; trimmed one genuinely redundant double-layer test. 27 pass.

<details>
<summary><b>Provenance model (decision detail)</b></summary>

`Sitemap._decorator_opted_in: set[str]` records which paths the decorator owns. The decorator routes through `_add_from_decorator` / `_remove_from_decorator`; the latter retracts **only** decorator-owned paths. Crucially, `add()` (the public/manual path) `discard()`s the path from that set, so a manual `add()` re-claims ownership and a later plain `@ui.page()` cannot wipe it. `remove()` and `reset()` keep the set consistent. Sequences verified: opt-in→manual-add→plain-decorator (manual survives); opt-in(True)→opt-in(False) (retracts); manual-add→plain-decorator (survives); reset (clears both).
</details>

<details>
<summary><b>Empirical, CI, and second-model review</b></summary>

- Empirical (`to_xml` output + global-map probe): `xmlns` literal attribute ✓, `0.85` non-lossy ✓, `5.0`→`1.0` clamp+warn ✓, `NaN`/`±inf` omitted with warning ✓, `ET._namespace_map` unmutated ✓.
- CI (local): `mypy` clean (240 files), `pylint` 10.00/10, `pytest tests/test_sitemap.py` 27/27, pre-commit clean.
- **Independent review (Codex, different lineage)** caught two real bugs in an earlier draft and I fixed both before this push:
  1. manual `add()` after a decorator opt-in didn't clear ownership → a later plain `@ui.page()` still wiped it. Fixed (`add()` discards provenance) + regression test `test_manual_add_after_decorator_optin_survives_plain_decorator`.
  2. `max(0, min(1, NaN))` = `1.0` → `NaN` priority silently became *highest*. Fixed (`math.isfinite()` guard) + parametrized test `test_to_xml_omits_non_finite_priority_with_warning` (`nan`/`inf`/`-inf`).
  Re-review verdict: `CONCUR-PUSH` — *"provenance fix resolves the blocked sequence … math.isfinite() … omitted with a warning rather than emitted as misleading priority values."*
</details>

Please re-review.

---
*Changelog — 2026-06-28: expanded the evidence (provenance detail, empirical specifics, CI figures, the two Codex-caught bugs + verdict) after the initial post.*

*Changelog — 2026-06-29: escaped the per-finding numbers (`#N`) that GitHub was auto-linking to unrelated issues (1)–(5).*

falkoschindler (changes requested):
@evnchn I've started going through this and one thing already jumped out, so flagging it now — I haven't read the whole PR yet, so more may surface.

The unconditional `/sitemap.xml` registration is a silent breaking change for a narrow but important group of existing apps. The endpoint is registered at import time (`nicegui.py`, `@app.get('/sitemap.xml')`), regardless of whether any page opted in, so every app serves `/sitemap.xml` after upgrading. Two consequences:

**1. Apps with no sitemap today:** `/sitemap.xml` goes from `404` → `200` with an empty `<urlset/>`. Not an API break, but a behavior change — and an empty sitemap is arguably worse for crawlers than a 404.

**2. Apps that already hand-roll their own `/sitemap.xml`** — exactly the population this PR's motivation describes ("every NiceGUI app today has to hand-roll one"). Here it depends on _how_ they did it:

- Via `@ui.page('/sitemap.xml')`: fine. `page.__call__` calls `remove_route` first, so the user's route replaces the built-in.
- Via a raw route (`@app.get('/sitemap.xml')`, `app.add_route`, or a mounted static file): **broken, and silently.** The built-in route is registered first (import time, before user code), Starlette matches first-hit, so the user's own sitemap is shadowed by NiceGUI's empty one.

Worth being precise about the mechanism: NiceGUI doesn't overwrite or remove the user's route — it stays in the routing table, it just never gets reached. That's why it's silent: no error, no warning, no collision message. The request simply lands on the wrong (empty) route. A user upgrades, changes nothing, and `/sitemap.xml` quietly starts returning "no URLs" — they find out via Search Console, not a log line.

I confirmed both paths locally: raw `@app.get('/sitemap.xml')` → built-in's `<urlset>` wins; `@ui.page('/sitemap.xml')` → user's route wins (route count stays 1).

Returning `404` when the registry is empty does **not** fix case 2 — the built-in route still intercepts the request before the user's raw route is ever consulted. It only addresses case 1.

I don't want to prescribe the solution, since you've thought hardest about this layer — could you have a think about how you'd want to handle it? One direction that would close the shadowing: register the endpoint at startup (after user modules are imported) and skip it if a route for `/sitemap.xml` already exists, so a user-defined sitemap always wins while we still serve ours for everyone else. But there may be a cleaner angle — curious what you land on.

(Minor, related: the "No breaking changes to the public API" checkbox is true for the Python signature, but the HTTP/routing surface does change for the raw-route apps above.)

Given that this needs another design pass and I still have the rest of the PR to review, I'd suggest we move it to 3.15 — it's getting a bit tight for 3.14 and I'd rather not rush a public-API addition.

evnchn:

Your `27c697e6` (priority validation in `add()`) is 👍 — warn-once-at-registration with a clean float stored is the right call; nothing to add.

Finding 6 reproduced. It's two orthogonal decisions:

**(2a) Empty registry → `404`?** An app that never opted in should keep its old `404`, not a misleading empty `<urlset/>`. One handler guard, breaks no current test. I lean **yes** — matches your "empty is worse than 404". OK?

**(2b) Let a user-defined `/sitemap.xml` win.** Has to defer to startup either way. Two shapes, both verified to make the user's route win:
- **A (yours):** register at startup + skip-if-exists. Cleanest steady state (one route ever), mirrors `/favicon.ico` — but the 4 fast `TestClient` endpoint tests must move to the `screen` fixture (route only exists once the server runs).
- **D:** keep import-time registration, retract ours at startup only if a user route also exists. Keeps all 27 tests green & fast; tiny "add-then-maybe-remove" asymmetry.

<details><summary>D, in full</summary>

```python
# _startup(), after user modules are imported:
hits = [r for r in app.routes if getattr(r, 'path', None) == '/sitemap.xml']
if len(hits) > 1:                       # user defined their own
    app.router.routes.remove(hits[0])   # ours is earliest-registered → drop it
```
`ui.page('/sitemap.xml')` unaffected (its `remove_route` already collapses to one route).
</details>

I lean **D** to keep the fast endpoint coverage (favicon *must* be startup-registered; our sitemap needn't be) — but A is genuinely cleaner and it's your call on a public surface. Tell me your pick on 2a/2b and I'll push it.

falkoschindler:
Thanks — and good split into 2a/2b, that's exactly the right decomposition.

**2a (empty registry → `404`): yes.** An app that opted in nothing should keep its old `404`, not a misleading empty `<urlset/>`. Go ahead.

**2b: I'd go with A (startup + skip-if-exists), on design merit rather than test convenience.** A gives one invariant — "at steady state there's exactly one `/sitemap.xml` route, and it's the user's if they defined one" — which is much easier to reason about than D's add-then-maybe-remove, it mirrors how `/favicon.ico` already works, and it avoids D's positional assumption (`hits[0]` is ours) and the `len(hits) > 1` mount-detection edge. Under A, 2a also falls out for free: at startup, only register ours if the registry is non-empty and no user route exists.

On the reason you leaned D — keeping the 4 endpoint tests fast — two things:

- I don't want the tests to steer the feature's implementation here; the route's lifecycle should be whatever's cleanest for the public surface, and the suite adapts to that, not the other way round. (Separately, I haven't reviewed the tests yet and may reshape them toward user-centric / end-to-end coverage anyway, so their current fast-`TestClient` shape isn't a fixed constraint.)
- I'm also not sure the cost is real. The endpoint tests are only fast because they call `TestClient(core.app).get(...)` without the context-manager form, so lifespan never runs. With `with TestClient(core.app) as client:` Starlette runs the lifespan — which under A registers the route — no browser and no `screen` fixture. Could you check whether NiceGUI's `_startup` runs cleanly inside `TestClient`'s lifespan? If it does, A keeps fast endpoint coverage and D loses its only advantage. If it genuinely can't, that's fine too — I'd still take A and pay for it with a couple of slower tests.

One thing worth noting for either option: both only handle routes defined before startup. A user who registers `/sitemap.xml` dynamically after startup would still be shadowed — acceptable, but let's state it rather than imply full generality.

evnchn:

Thanks — pushed 1db9f86c implementing **2a + 2b (A)**. And **yes: `_startup` runs cleanly inside `TestClient`'s lifespan, so A keeps fast endpoint coverage** — D loses its only advantage, exactly as you suspected.

**Your empirical question, answered.** `with TestClient(app) as client:` runs the full NiceGUI lifespan (`_startup` → my new startup registration → `_shutdown`) with **no browser and no `screen` fixture** — *provided the run config is set*. A bare `with TestClient(core.app) as client:` on its own raises `_startup`'s "You must call `ui.run()`" guard (`nicegui.py:139`), because the plain `nicegui_reset_globals` fixture never sets `has_run_config`. The clean way in is the pattern already in `tests/test_run_with.py`: `ui.run_with(FastAPI())` sets the run config, then `with TestClient(fastapi_app) as client:`. All four endpoint tests moved to that shape — still fast (whole file **0.57 s**), still browser-free.

**What landed:**
- **2b (A):** `/sitemap.xml` is registered in `_startup()` (`_register_sitemap_route()`), mirroring `/favicon.ico`. Skip-if-exists: if a user route already claims `/sitemap.xml`, we don't register ours → theirs wins. One steady-state invariant, as you framed it.
- **2a (falls out of A):** if nothing opted in, we don't register at all → the app keeps its original `404`, never an empty `<urlset/>`.
- **Limitation, stated (not implied away):** only routes present *before* startup are detected; a `/sitemap.xml` added dynamically *after* startup would still be shadowed by ours. Documented on both `Sitemap` and `_register_sitemap_route`.

<details>
<summary><b>Idempotency bug caught by second-lineage review (Codex) — fixed before push</b></summary>

Codex (different model lineage) flagged that `_register_sitemap_route()` wasn't idempotent across repeated startups **in one process**: a first startup registers our route; if a later startup runs with an emptied registry, the stale route survives and serves an empty `<urlset/>` (`200`) instead of the contracted `404` — and it also masquerades as a "user route" to the skip check. It doesn't bite in production (startup runs once) or the current per-test-reset suite, but it violates the "re-decide each startup" half of your invariant.

Fix: strip our own route (identified by `endpoint is _get_sitemap`) at the top of the function before re-deciding, so each startup starts from scratch. Regression test `test_registration_is_idempotent_across_startups` (opt-in → 200, then `reset()` → **404**) fails without the fix, passes with it.
</details>

<details>

- **Method semantics preserved:** `app.add_route(...)` (default methods) gives `GET`/`HEAD` → `200`, `POST`/`PUT`/`DELETE` → `405` — identical to the old `@app.get`.
- **Three behaviors, via `TestClient` lifespan:** opt-in → `200 application/xml`; empty registry → `404` (NiceGUI's normal 404 page); user-defined `@app.get('/sitemap.xml')` → their handler wins (`200 text/plain`, exactly one route).
- **Reset fixture:** the import-time `/sitemap.xml` preservation special-case in `nicegui_reset_globals` is removed — the route is now created inside lifespan and cleaned per-test like any other.
- **CI (local):** `pre-commit` clean · `mypy ./nicegui` clean (244 files) · `pylint ./nicegui` 10.00/10 · `pytest tests/test_sitemap.py tests/test_run_with.py` 34 passed in 0.69 s.

I left the endpoint tests in the fast `ui.run_with` + lifespan shape since it keeps coverage cheap — but per your note that you may reshape them toward user-centric / e2e anyway, treat that as non-binding.
</details>

evnchn:

**Ran a fresh full-PR self-review (two independent model lineages) before your re-review.** Core logic came back solid from both — provenance state machine, startup registration / skip-if-exists / empty→404, namespace handling, XML escaping (no injection: spoofed `Host`/paths are all `ET .text`-escaped, verified), and test quality (no false-positive coverage). The one theme both hit was **incomplete input validation at the `add()`/render boundary** — the very thing this PR advertises as its single validation point. Pushed 7e7eff26 closing the clearly-correct ones:

- **path must start with `/`** — `add('docs')` previously emitted `<loc>https://host` + `docs`. Now a `ValueError` at registration, matching the existing `{...}` rejection style.
- **reject control chars in path** — a `\x00` previously serialized into invalid XML (served `200`).
- **reject non-string `lastmod`/`changefreq`** — `lastmod=date.today()` previously stored fine then raised `TypeError` at render → `500`. Now caught at `add()`.
- **priority formatting** — `str(0.00001)` is `'1e-05'`, scientific notation the protocol rejects. Now a plain decimal (`0.00001`), with `1.0`/`0.0`/`0.85` unchanged and repeating values bounded.

10 new tests; `mypy`/`pylint` clean, `pytest tests/test_sitemap.py` 40 passed.

<details>
<summary><b>Optional follow-ups (your call — deliberately NOT done)</b></summary>

These are opinionated enough that I left them for you rather than bundling:

1. **Quote `X-Forwarded-Prefix` in the endpoint** to match `client.py`'s deliberate `quote(hdr, safe="/:@!$&'()*+,;=%")` (`client.py:198`). No injection risk today (the value is `ET .text`-escaped), so this is consistency + defense-in-depth with the convention this PR already references — a no-op for well-formed prefixes. One reviewer flagged it, one called it acceptable-as-is.
2. **Validate the `changefreq` enum** — a string like `'often'` is currently accepted and rendered (protocol-invalid but harmless). Could `warn`+omit, mirroring how `priority` already handles out-of-range values.
3. **Coerce `date`/`datetime` for `lastmod`** instead of requiring a string — friendlier, but expands the public API surface.

Happy to do any/all if you'd like them in; otherwise they're fine as-is or as a follow-up.
</details>

<details>
<summary><b>Review method</b></summary>

Full PR diff reviewed from scratch by two independent lineages (a fresh Claude subagent applying `REVIEW.md`, and Codex — a different model family), plus direct empirical probes for every finding (crafted paths/metadata through `to_xml`, method semantics `GET`/`HEAD`→200 & others→405, the multi-lifespan idempotency case). Only findings that reproduced were acted on; the rest of the PR both passes reported as solid.
</details>

evnchn:

**Correction to my last comment — and a fix pushed (ea4cc265).** The `ui.run_with(FastAPI())` + `with TestClient(app) as client:` shape I described for the endpoint tests turned out to have a **CI test-isolation cost** I missed: it made the drag-drop browser tests in `tests/test_sortable.py` fail. Now fixed.

**Root cause:** entering the lifespan runs `_startup`, whose `run.setup()` forks a `ProcessPoolExecutor`. With ~7 lifespan-entering sitemap tests, that's ~one forked pool churned per test — and on CI's resource-constrained **Linux** runners that destabilized the *adjacent* `test_sortable` tests (`test_sitemap.py` collects immediately before `test_sortable.py`). It never showed locally because macOS uses spawn, not fork — so it took a CI bisect to pin.

**Fix (ea4cc265):** don't spin the whole server per test. Call `_register_sitemap_route()` directly — exactly what `_startup` does for this route — and serve through a plain, lifespan-free `TestClient(core.app)`. No process pool, no pollution. The empty-registry and idempotency cases assert on `app.routes` directly (no HTTP), avoiding the 404-page render that otherwise spun up an unawaited `Outbox` coroutine.

Your original empirical question still holds — `_startup` *does* run cleanly inside `TestClient`'s lifespan (verified) — it's just too heavy to spin **per test**, so the suite exercises the registration step directly instead.

<details>
<summary><b>Evidence trail (CI bisect)</b></summary>

- Failure was deterministic and branch-specific: `test_basic_reorder` + `test_cross_container` failed **3/3** on this branch in CI while `main` and every other PR were green; not reproducible locally under any config (isolation 5/5, `test_sitemap`+`test_sortable` 5/5).
- Deduction ruled out my *code*: `test_dark_mode` runs `_register_sitemap_route()` on its own startup yet passes — only tests *after* `test_sitemap.py` failed.
- Bisect: dropping just the endpoint tests (code untouched) → CI green ⇒ the tests, not the code.
- Fix verified on a throwaway branch (full CI green) before pushing here.
- Local after fix: 45 passed, `mypy`/`pylint 10.00/10`/`pre-commit` clean.
</details>

falkoschindler:
We didn't manage to get this into 3.15, so it has to move one more time — but we'll firmly schedule it for 3.16. Thanks for bearing with us!

evnchn:

**Four fixed in code, one documented rather than fixed, one still genuinely open** — flagging only because the PR still reads "changes requested" and has been quiet since the milestone move. No rush at all.

<details>
<summary><b>Where each finding landed</b></summary>

| Finding | Status |
|---|---|
| Unconditional `/sitemap.xml` registration shadows an app's own raw route (2026-06-29) | **Fixed.** Registration moved to startup and defers to user routes — a route the app registered itself wins |
| 1 · Default `sitemap=False` is an active removal | **Fixed.** `False` now calls `_remove_from_decorator(...)`, which only retracts entries the decorator itself added, so a manual `add()` survives a later plain `@ui.page` |
| 2 · `priority` silently corrupted and never validated | **Fixed.** `_normalize_priority()` validates at registration: non-numeric and non-finite values are dropped with a warning instead of blowing up in `to_xml()`, and out-of-range values are clamped to `[0.0, 1.0]` |
| 3 · `ET.register_namespace` clobbers the global default-namespace slot | **Fixed** your way — the namespace is a literal root attribute and tags are unprefixed, so `_namespace_map` is never touched |
| 4 · Router prefix captured at decoration time | **Documented, not fixed.** `page.py:70-72` now warns that the path does not reflect a later `app.include_router(router, prefix=...)` and points at explicit `app.sitemap.add()`. If you'd rather it resolve the include-time prefix properly, that's a real change and I'm happy to take it |
| 5 · Tests over-tested and missing the risky paths | **Half done.** The missing coverage is in — priority precision and clamping, the literal-attribute namespace, add-then-plain-decorator. The trim is not: the file went ~184 → 342 lines, the opposite direction from what you asked for |

On finding 5 — that's more pointed now than when you raised it, given the testing preference you've since written down. Say the word and I'll cut it back; the parameterized validation rules are re-run through the decorator when one test proving it reaches `add()` would do, and a couple (`test_sitemap_is_empty_by_default`, `test_to_xml_escapes_special_characters`) really do test stdlib behavior.
</details>

PR #6059, opened 2026-05-19
Fix `Client has been deleted` warning spam on reload (#6058)

### LINKED ISSUE #6058: Scene object mutations after `await` spam "Client has been deleted" warnings on reload
### First Check

### Example Code

```python
from nicegui import ui, run
import time

def cpu_work() -> None:
    time.sleep(0.3)

@ui.page('/')
def index():
    with ui.scene():
        box = ui.scene.box()

    async def update():
        await run.cpu_bound(cpu_work)
        box.rotate(0, 0, time.time())

    ui.timer(0.4, update)

ui.run(reconnect_timeout=0.5)
```

### Description

Open the page, then reload it.
The server log fills up with `Client has been deleted but is still being used.` warnings, each with a stack trace, even though the user code does nothing wrong.

The pattern — `ui.timer` calling an async callback that `await`s a background task and then mutates an element — is one NiceGUI's API actively encourages.
For regular elements (`ui.label`, etc.) it works fine, because `Element.update()` has a `if self.is_deleted: return` short-circuit ([element.py:413-417](https://github.com/zauberzeug/nicegui/blob/main/nicegui/element.py#L413-L417)).
Scene objects mutate via `scene.run_method(...)` instead, and `Element.run_method()` has no such guard, so every mutation between awaits ends up scheduling an `AwaitableResponse._fire` background task that hits `Outbox.enqueue_message` → `Client.check_existence()` after the client was deleted.

The race window opens because `Client.delete()` is deferred by `reconnect_timeout` ([client.py:345-353](https://github.com/zauberzeug/nicegui/blob/main/nicegui/client.py#L345-L353)) and timer cancellation only takes effect at the next await — any synchronous scene mutations in between still schedule their `_fire` tasks, which wake after `_deleted = True`.

The asymmetry (regular elements silent, scene objects noisy) means user code can't tell whether it's hitting a real bug or a benign reload race.
RoSys works around this with a per-call `_is_alive` gate (zauberzeug/rosys#406), which shouldn't be necessary.

A minimal fix would be to mirror the `is_deleted` guard from `Element.update()` in `Element.run_method()` (and possibly `Client.run_javascript` for direct calls), so the no-op behavior is consistent across element types.

### NiceGUI Version

3.12.0

### Python Version

3.12

### Browser

Chrome

### Operating System

macOS

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #6058. Opening the example page and reloading it floods the server log with `Client has been deleted but is still being used.` warnings (each with a stack trace), even though user code does nothing wrong.

The root cause is an asymmetry between `Element.update()` and `Element.run_method()`:

- `Element.update()` short-circuits on `is_deleted` (silent), so regular elements like `ui.label` work fine when a timer callback mutates them after `await`.
- `Element.run_method()` has no such guard. Scene objects mutate via `scene.run_method(...)`, so every mutation between awaits schedules an `AwaitableResponse._fire` task that hits `Outbox.enqueue_message` → `Client.check_existence()` *after* `Client.delete()` ran (deferred by `reconnect_timeout`).

A naive fix — mirror the `is_deleted` guard from `update()` onto `run_method()` — would also silence the case where user code explicitly calls `element.delete()` and then accidentally uses the element again. That's a real use-after-free worth surfacing, so the silent-blanket approach loses important signal.

### Implementation

Two-tier deletion guard, factored into `Element._is_safe_to_interact()`:

| Case | Behavior | Rationale |
| ---- | -------- | --------- |
| `client.is_deleted` (e.g. browser reload race past `reconnect_timeout`) | silent no-op | User code did nothing wrong; the timer's in-flight coroutine just outlived the client. |
| `element.is_deleted` but client alive (explicit `element.delete()` then misuse) | one-shot warning with `stack_info=True` via `helpers.warn_once` | Real bug in user code; deserves a stack trace. |
| both alive | proceed | Normal path. |

Applied to `Element.update()`, `Element.run_method()`, and `Element.get_computed_prop()` so the three I/O paths stay symmetric. The discriminator uses a new public `Client.is_deleted` property mirroring the existing `Element.is_deleted`.

`Client.check_existence()` (called from `outbox.enqueue_message`) is left untouched: direct `client.run_javascript()` calls on a deleted client still warn via that path, since they bypass the element-layer guard and are inherently more deliberate user actions.

#### Behavior change worth flagging

`Element.update()` previously silenced *both* deletion cases. After this PR, it silences only the disconnect race and emits a one-shot warning for the explicit-element-delete-then-misuse case. Net effect: more signal for buggy user code, less noise for benign reload races. Not a public API break — same method signature, same return — but downstream code that relied on `update()` being silent after `element.delete()` will see a new warning (deduped, one-shot per element id).

### Progress

- [x] Pytests have been added: `test_run_method_silent_after_client_delete` and `test_run_method_warns_after_explicit_element_delete` in `tests/test_element_delete.py`. All 10 tests in that file pass; `test_keep_alive.py` and `test_scene.py` also pass with no regressions.
- [x] Documentation has been added (docstrings on `Client.is_deleted` and `Element._is_safe_to_interact`); no user-facing docs needed.
- [x] No breaking changes to the public API; one behavior change to `Element.update()` is documented above.

---

Development history and review iteration on my fork: evnchn/nicegui#154 (closed in favor of this PR).

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -16,7 +16,7 @@ from starlette.background import BackgroundTask
 from typing_extensions import Self
 
 from . import background_tasks, binding, core, helpers, json, storage
-from .awaitable_response import AwaitableResponse, NullResponse
+from .awaitable_response import AwaitableResponse
 from .dependencies import generate_resources
 from .element import Element
 from .favicon import get_favicon_url
@@ -262,8 +262,6 @@ class Client:
 
         :return: AwaitableResponse that can be awaited to get the result of the JavaScript code
         """
-        if self._deleted:
-            return NullResponse()
         request_id = str(uuid.uuid4())
         target_id = self._temporary_socket_id or self.id
 
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -16,7 +16,7 @@ from starlette.background import BackgroundTask
 from typing_extensions import Self
 
 from . import background_tasks, binding, core, helpers, json, storage
-from .awaitable_response import AwaitableResponse
+from .awaitable_response import AwaitableResponse, NullResponse
 from .dependencies import generate_resources
 from .element import Element
 from .favicon import get_favicon_url
@@ -132,6 +132,11 @@ class Client:
     def has_socket_connection(self) -> bool:
         """Whether the client is connected."""
         return self.tab_id is not None
+
+    @property
+    def is_deleted(self) -> bool:
+        """Whether the client has been deleted (e.g. by browser disconnect after ``reconnect_timeout``)."""
+        return self._deleted
 
     @property
     def head_html(self) -> str:
@@ -257,6 +262,8 @@ class Client:
 
         :return: AwaitableResponse that can be awaited to get the result of the JavaScript code
         """
+        if self._deleted:
+            return NullResponse()
         request_id = str(uuid.uuid4())
         target_id = self._temporary_socket_id or self.id
 
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -412,7 +412,12 @@ class Element(Visibility):
 
     def update(self) -> None:
         """Update the element on the client side."""
+        if self.client.is_deleted:
+            return  # silent: client teardown race (e.g. browser reload during async callback)
         if self.is_deleted:
+            helpers.warn_once(f'Element {self.id} has been deleted but is still being used. '
+                              'This is most likely a bug in your application code.',
+                              stack_info=True)
             return
         self.client.outbox.enqueue_update(self)
 
@@ -428,6 +433,13 @@ class Element(Visibility):
         """
         if not core.is_loop_running():
             return NullResponse()
+        if self.client.is_deleted:
+            return NullResponse()  # silent: client teardown race
+        if self.is_deleted:
+            helpers.warn_once(f'Element {self.id} has been deleted but is still being used. '
+                              'This is most likely a bug in your application code.',
+                              stack_info=True)
+            return NullResponse()
         return self.client.run_javascript(
             f'return runMethod({self.id}, {json.dumps(name)}, {json.dumps(args)})', timeout=timeout,
         )
@@ -441,6 +453,13 @@ class Element(Visibility):
         :param timeout: maximum time to wait for a response (default: 1 second)
         """
         if not core.is_loop_running():
+            return NullResponse()
+        if self.client.is_deleted:
+            return NullResponse()  # silent: client teardown race
+        if self.is_deleted:
+            helpers.warn_once(f'Element {self.id} has been deleted but is still being used. '
+                              'This is most likely a bug in your application code.',
+                              stack_info=True)
             return NullResponse()
         return self.client.run_javascript(
             f'return getComputedProp({self.id}, {json.dumps(prop_name)})', timeout=timeout,
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -410,14 +410,26 @@ class Element(Visibility):
         args = events.GenericEventArguments(sender=self, client=self.client, args=msg['args'])
         events.handle_event(listener.handler, args)
 
-    def update(self) -> None:
-        """Update the element on the client side."""
+    def _is_safe_to_interact(self) -> bool:
+        """Return True if it is safe to send messages to this element's client.
+
+        Silent when the *client* has been deleted (e.g. browser reload race past ``reconnect_timeout``):
+        an async callback resuming after the teardown is not a user bug. Emits a one-shot warning
+        when the *element* has been explicitly deleted but the client is still alive: that is a real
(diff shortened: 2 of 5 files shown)


evnchn:
@codingpaula please lmk if this can be a true fix to the issue in https://github.com/zauberzeug/rosys/pull/406. thanks!

falkoschindler:
Great solution, @evnchn! Just one review finding worth looking into:

**Dead-weakref case not handled.** `_is_safe_to_interact()` calls `self.client.is_deleted`, but `Element.client` raises `RuntimeError` when `self._client()` returns `None` (the Client object got GC'd). After `Client.delete()` runs, `Client.instances[id]` is gone — the remaining strong refs come from queued `AwaitableResponse._fire` closures (`send_and_forget` captures `self` = the Client). Each `_fire` releases its closure as soon as it completes, so between two queued mutations the Client can be GC'd, and the *next* `run_method` call on a held element reference will see `self._client() is None`. That turns the silent disconnect race into a `RuntimeError` crashing the user callback — strictly worse than today's warning. One-line fix:

   ```python
   def _is_safe_to_interact(self) -> bool:
       client = self._client()
       if client is None or client.is_deleted:
           return False
       if self.is_deleted:
           helpers.warn_once(...)
           return False
       return True
   ```

evnchn:
**TL;DR:** Fixed in d1c34f15. Below is the local regression test I wrote to verify the fix, in case you'd like to adopt it — didn't add it to the PR since I wasn't sure of scope.

<details><summary>Optional regression test for dead-weakref case</summary>

```python
async def test_run_method_silent_after_client_weakref_dies(user: User, caplog: pytest.LogCaptureFixture):
    """If the Client has been garbage-collected (weakref returns None), element methods
    must stay silent — not raise RuntimeError. See issue #6058."""
    captured: dict = {}

    @ui.page('/')
    def page():
        captured['label'] = ui.label('hi')

    await user.open('/')
    label = captured['label']
    label._client = lambda: None  # type: ignore[assignment]  # pylint: disable=protected-access

    _shown_warnings.clear()
    caplog.clear()
    label.run_method('foo')
    label.update()
    label.get_computed_prop('bar')

    warning_records = [r for r in caplog.records if 'deleted' in r.message.lower()]
    assert warning_records == [], f'expected silent, got: {[r.message for r in warning_records]}'
```

Patches `label._client` to `lambda: None` to simulate the dead-weakref case directly — no `gc.collect()` flakiness. Passes against d1c34f15 and (as expected) fails against df779f7d with `RuntimeError: The client this element belongs to has been deleted.`

</details>

falkoschindler:
One concern about the `warn_once` dedup key — it embeds `{self}` and `id={self.id}`, which are unique per element. In a pattern where the _same_ user-code bug touches many elements (e.g. a loop that mutates each child of a cleared container, or a custom element whose teardown doesn't cancel its own timer/background task and that element is instantiated many times), every affected element triggers its own warning, each with a full stack trace via `stack_info=True`. The current "one-shot" semantics don't actually prevent flooding in that case.

Two ways to tighten this:

1. **Static message, like `Client.check_existence`** — drop `{self}` / `id=…` from the dedup key and log one warning per process. The stack trace already pinpoints the call site, so element identity is recoverable. Lowest noise, matches the existing convention in `client.py`.

2. **Add an explicit `key=` parameter to `warn_once`** — let callers pass a stable dedup key separate from the message (e.g. `key='element_deleted_in_use'`), so the message stays informative while the dedup stays coarse. Small, opt-in change to `helpers/warnings.py`, reusable by other call sites if they want it. Auto-extracting the caller frame to use `filename:lineno` as the implicit key would be the prettier version, but picking the "right" frame is fragile — `warnings.warn` only gets away with it because of an explicit `stacklevel=` argument at every call site, which defeats "small change."

For this PR I'd go with (1) — the stack trace is the diagnostic signal, element identity is recoverable from it, and it matches the existing convention. (2) is a reasonable follow-up if/when another call site needs the same shape.

falkoschindler (approved):
Since I didn't hear back on how to dedup the deleted-element warning, I went ahead and implemented option 1 myself (`ea4c93cee`): the message embedded `{self}`/`id=`, so a single bug touching many elements could grow `_shown_warnings` unboundedly and emit a stack trace per element. It now uses a static message (one warning per process) mirroring `Client.check_existence` — `stack_info=True` already pinpoints the call site, so element identity is recoverable.

I also streamlined the tests (`2efc53226`): Added `warnings.reset()` and call it from `nicegui_reset_globals()` so every test starts clean. Folded the two deletion-usage tests into one parametrized `test_usage_after_delete` and dropped the manual `_shown_warnings.clear()`.

All of `test_element_delete.py` passes, plus `test_keep_alive.py` / `test_scene.py` with no regressions. Approving. 🎉

PR #6060, opened 2026-05-20
Add Plotly.extendTraces demo and run_plot_method helper

evnchn:
### Motivation

Mirrors [#5967](https://github.com/zauberzeug/nicegui/pull/5967) (the AG Grid equivalent) for the Plotly side of the same problem: calling `plot.update()` re-transmits the entire figure to the client every time, which is wasteful for live data. Plotly.js has [`Plotly.extendTraces`](https://plotly.com/javascript/plotlyjs-function-reference/#plotlyextendtraces) for appending points to existing traces with minimal traffic, but reaching it today requires `ui.run_javascript(...)` with a hand-built `getHtmlElement({plot.id})` lookup — discussed at length in [#2660](https://github.com/zauberzeug/nicegui/discussions/2660).

Closes the design question raised there. The longer reasoning (and why this lighter answer is preferable to a per-method wrapper like `plot.extend_traces(...)`) lives in [the follow-up comment](https://github.com/zauberzeug/nicegui/discussions/2660#discussioncomment-16984084), but the short version is: don't commit NiceGUI to mirroring plotly.js's mutation vocabulary forever — just expose the existing `run_method` primitive cleanly and link out to the [plotly.js function reference](https://plotly.com/javascript/plotlyjs-function-reference/) for discoverability.

### Implementation

Three small changes, no new public concepts:

1. **`nicegui/elements/plotly/plotly.py`** — adds `run_plot_method(name, *args, timeout=1)`. Signature mirrors `AgGrid.run_grid_method` / `EChart.run_chart_method`. The chart's HTML element is passed as the first argument to the plotly.js function automatically, so users call `plot.run_plot_method('extendTraces', update, [0])` instead of building a DOM-lookup JS string.
2. **`nicegui/elements/plotly/plotly.js`** — adds the matching Vue method that dispatches to `Plotly[name](this.$el, ...args)` via the existing `runMethod` helper. Adds the `convertDynamicProperties` import (matches `aggrid.js`'s pattern). Guards against the dynamic-`import()` race in `mounted()` — if `this.Plotly` is still undefined when called, emits a clear error via `logAndEmit` instead of crashing in `runMethod`'s string-coercion path.
3. **`website/documentation/content/plotly_documentation.py`** — adds a `@doc.demo('Extending traces without re-sending the full figure', ...)` right after the existing *Plot updates* demo. Shows the `extendTraces` recipe with the dict-figure form (the only form that supports `.append()` on trace `x`/`y` arrays — `go.Scatter.x` is a tuple internally).

The docstring links to the [plotly.js function reference](https://plotly.com/javascript/plotlyjs-function-reference/), which is the discoverability bridge — that URL was nowhere in NiceGUI's Python-side docs before this.

### Progress

- [x] Pytests are have been added.


falkoschindler (approved):
LGTM — clean, well-scoped mirror of `run_grid_method` / `run_chart_method`, and the design call (expose the `run_method` primitive rather than wrapping each plotly.js mutator) is the right one. Thanks!

I pushed a few follow-ups onto the branch:

- **Fixed a `TimeoutError` when awaiting most plotly.js functions.** `extendTraces` (and `restyle`/`relayout`/`addTraces`/`react`/…) resolve to the graph `<div>`, which can't be serialized back over the socket — so `await plot.run_plot_method('extendTraces', …)` timed out at 1 s. `run_plot_method` now awaits the result and drops it when it's the chart element, returning `None`. Genuinely-serializable returns (e.g. `toImage` → data-URL string) still come through.
- **Added tests** (`tests/test_plotly.py`): a smoke test that `extendTraces` appends a visible point, and a regression test that awaiting it resolves to `None` (verified it fails on the pre-fix code with the exact `TimeoutError`).
- **Docs polish**: noted in the demo description why the dict-figure form is used (mutable `x`/`y` lists), and clarified in the docstring that awaited calls return `None` when the function resolves to the chart element.

XSS is safe by construction — `run_plot_method` routes through `runMethod`'s `name in target` lookup (no `eval`), same as the AG Grid path.

Approving. 🚀

PR #6063, opened 2026-05-20
Hide ui.upload route from OpenAPI schema by default

falkoschindler:
### Motivation

When `fastapi_docs=True` is enabled, the `POST /_nicegui/client/{client_id}/upload/{element_id}` endpoint behind every `ui.upload()` shows up in `/docs` and `/openapi.json` — even with the default `endpoint_documentation='none'`, which is documented to hide internal `/_nicegui/...` routes.

Reproduction:

```python
from nicegui import ui

@ui.page('/')
def main_page():
    ui.upload()
    ui.link('Go to documentation', '/docs')

ui.run(fastapi_docs=True)
```

### Implementation

The filter loop in `ui_run.py` that applies `endpoint_documentation` to `/_nicegui/...` routes runs once at server startup. The upload route is registered lazily inside `Upload.__init__`, i.e. on the first page visit, so the filter never sees it and FastAPI's default `include_in_schema=True` sticks.

The fix passes `include_in_schema` explicitly at the `@app.post(...)` site in `elements/upload.py`, mirroring the same `endpoint_documentation in {'internal', 'all'}` policy.

The existing `test_endpoint_docs.py` is consolidated into a single parametrized test that adds `ui.upload()` to the page and asserts the exact expected `/openapi.json` path set for each of `'none'`, `'page'`, `'internal'`, `'all'` — including the dynamic `POST /_nicegui/client/.../upload/...` route for `'internal'` and `'all'`.

### Progress

### DIFF
--- a/nicegui/elements/upload.py
+++ b/nicegui/elements/upload.py
@@ -4,6 +4,7 @@ from fastapi import Request
 from starlette.datastructures import UploadFile
 from typing_extensions import Self
 
+from .. import core
 from ..defaults import DEFAULT_PROP, resolve_defaults
 from ..events import Handler, MultiUploadEventArguments, UiEventArguments, UploadEventArguments, handle_event
 from ..nicegui import app
@@ -70,7 +71,7 @@ class Upload(LabelElement, DisableableElement, component='upload.js'):
         self._upload_handlers = [on_upload] if on_upload else []
         self._multi_upload_handlers = [on_multi_upload] if on_multi_upload else []
 
-        @app.post(self._props['url'])
+        @app.post(self._props['url'], include_in_schema=core.app.config.endpoint_documentation in {'internal', 'all'})
         async def upload_route(request: Request) -> dict[str, str]:
             for begin_upload_handler in self._begin_upload_handlers:
                 handle_event(begin_upload_handler, UiEventArguments(sender=self, client=self.client))
--- a/tests/test_endpoint_docs.py
+++ b/tests/test_endpoint_docs.py
@@ -1,7 +1,7 @@
 import httpx
 import pytest
 
-from nicegui import Client, __version__, ui
+from nicegui import __version__, ui
 from nicegui.testing import Screen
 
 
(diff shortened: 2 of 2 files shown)


evnchn:

@falkoschindler — pushed a small follow-up to the test (6356b64) instead of leaving it as a suggestion. The earlier 47-line suggestion block was too noisy; the actual change is 6 lines, +1/-0 net.

It captures the `Upload` instance via a list so the test can derive the route as `f'/_nicegui/client/{upload.client.id}/upload/{upload.id}'` — drops both `next(iter(Client.instances))` (no precedent in `tests/`) and the literal element id `5` (every other test uses `.id` off a captured element).

Verified locally that the derived string matches the route the server registers byte-for-byte. If you don't like it, just `git reset --hard cea3949 && git push --force-with-lease` — happy to have it dropped.

<details>
<summary>Independent code-review (sub-agent) — merge-readiness assessment of the PR with this commit on top</summary>

For visibility — not as advocacy for keeping 6356b64. A separate code-review agent walked the diff `c0124f7..6356b64` with full repo context. Findings:

**Strengths**

- Root cause correctly identified — the lazy per-page `@app.post` bypasses the startup `include_in_schema` strip loop in `ui_run.py:186-192`. Fix is local and minimal (one line at `upload.py:74`).
- `core.app.config.endpoint_documentation` is evaluated at element-construction time, which is correct: the route is registered then, and the config is set once at `ui.run()`.
- `from .. import core` in `elements/upload.py` matches the established pattern (`markdown.py:10`, `timer.py:5`) — no circular import risk.
- **No sibling lazy-route bugs in `elements/`** — `grep` for `@app.{post,get,put,delete}` and `app.add_route` / `app.add_api_route` came back clean. The PR's scope is complete for this bug class; no other element registers a route inside its `__init__`.
- Test consolidation: the parametrized version actually exercises the upload route in all 4 modes (the old 4 tests didn't).
- The closure-over-list in 6356b64 is safe — `screen.open()` issues a GET that synchronously enters `with Client(self, request=request) as client:` (`page.py:166`) and runs the page body before the response builds. By the time `screen.open()` returns, `uploads[0]` exists.

**Issues**

- Critical: none.
- Important: none.
- Minor: `upload.py:74` is ~110 chars (borderline but readable); the `else: raise ValueError` arm in the parametrized test is dead code (harmless defensive default).

**CI on `6356b64`:** pre-commit / mypy / pylint / quick-test 3.10 pytest — all green.

**Verdict:** Ready to merge.

</details>

---

**Changelog**

- *Edit 2026-05-20*: appended the independent sub-agent merge-readiness assessment inside `<details>`. CI on `6356b64` was green at edit time.

falkoschindler:
@evnchn Ok, Claude and you convinced me that this is the better solution. 👍🏻

evnchn (approved):
Merge readiness assessment was green on my side but you need to think about whether to ship this in 3.12.1 versus 3.13

falkoschindler:
It's a bugfix with low risk and a colleague is waiting for it. So I think 3.12.1 is fine.

PR #6067, opened 2026-05-20
Add docs on how to fix non-working aggrid events

KrilleGH:
### Motivation

In https://github.com/zauberzeug/nicegui/discussions/2878 we encountered non-obvious behaviour regarding certain `aggrid` events. I also stumbled on this now, so there seems to be a need to make this more beginner-friendly. To directly provide a solution without searching we mention how to proceed in the docs.

### Implementation

Just add some paragraphs to the docs.

### Progress

### DIFF
--- a/website/documentation/content/aggrid_documentation.py
+++ b/website/documentation/content/aggrid_documentation.py
@@ -218,16 +218,16 @@ def aggrid_respond_to_event():
 
 
 doc.text('', '''
-    **Note:** Certain events, e. g. `rowClicked` don't seem to work.
+    **Note:** Certain events, e.g. `rowClicked`, don't seem to work.
     This is because some event arguments include cyclic references.
-    For these seemingly non-working events you'll see an error on the developer console in your browser.
+    For these seemingly non-working events you'll see a serialization error on the developer console in your browser.
 
     To inspect the event arguments and find the values you are interested in,
     you can replace the event registration with the following JavaScript handler:
     ```
     .on('rowClicked', js_handler='console.log')
     ```
-    Then limit the event arguments to the necessary ones, e. g. `data`, thus excluding the cycles:
+    Then limit the event arguments to the necessary ones, e.g. `data`, thus excluding the cycles:
     ```
     .on('rowClicked', lambda event: ui.notify(f'Row: {event.args}'), ['data'])
     ```
--- a/website/documentation/content/aggrid_documentation.py
+++ b/website/documentation/content/aggrid_documentation.py
@@ -202,6 +202,22 @@ def aggrid_with_html_columns():
 @doc.demo('Respond to an AG Grid event', '''
     All AG Grid events are passed through to NiceGUI via the AG Grid global listener.
     These events can be subscribed to using the `.on()` method.
+
+    Note: Certain events, e. g. `rowClicked` don't seem to work. This is because some
+    event arguments include cyclic references. For these seemingly non-working events
+    you'll see an error on the developer console in your browser.
+
+    To inspect the event arguments and find the values you are interested in, you can
+    replace the event registration with the following JavaScript handler:
+    ```
+    .on('rowClicked', js_handler='console.log')
+    ```
+    Then limit the event arguments to the necessary ones, e. g. `data`, thus
+    excluding the cycles:
+    ```
+    .on('rowClicked', lambda event: ui.notify(f'Row: {event.args}'), ['data'])
+    ```
+    This selection can be serialized safely and will work correctly.
 ''')
 def aggrid_respond_to_event():
     ui.aggrid({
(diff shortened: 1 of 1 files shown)


evnchn:
@KrilleGH Can take this PR as is, no issues on my end, 

... but do you think a dedicated "debugging" section for NiceGUI is coming? 

If so all content along the vibes of this one is welcome!

KrilleGH:
> ... but do you think a dedicated "debugging" section for NiceGUI is coming?

I didn't think about that, but it would certainly be nice to have something like that, or a cookbook.

When I have an issue, I often search for it, usually there's a quick result. It would be cool to collect the wisdom from past discussions or issues in this project, StackOverflow, etc. in a central place.

If that happens, then this content can be moved, keeping a link to the new place.

What I actually thought about was using a more visible note callout or moving the paragraph below the code, but didn't bother in the end.

falkoschindler:
@evnchn @KrilleGH Reminds me of our [FAQ](https://github.com/zauberzeug/nicegui/wiki/FAQs) which is probably rarely used. But before creating another "central place", we should think about existing sources and where it should live in the end. NiceGUI docs is great, but always requires going through a PR to add or update something. Anyway, maybe we should continue this discussion in a separate thread?

falkoschindler (dismissed):
Thanks for the contribution, @KrilleGH!
I just moved the new note below the code - as you suggested.
Let's merge! 🚀

falkoschindler (approved):

evnchn:
Took a stab at the "central place" idea — collected ~80 NiceGUI debugging gotchas from closed issues, discussions, and direct experience into a single rich-rendered page (dark-themed).

- 📄 Rendered: https://htmlpreview.github.io/?https://gist.githubusercontent.com/evnchn/11d8962581aec36fbd6f8f29e7cb4958/raw/nicegui-debugging-tips.html
- 💾 Source gist: https://gist.github.com/evnchn/11d8962581aec36fbd6f8f29e7cb4958

14 categories — Refreshable, Slot stack, Storage, Async/lifecycle, AG Grid, Plotly, ECharts/Leaflet, Events & bindings, Reload & deployment, Sub-pages, Memory & perf, Dialogs & scenes, Testing, Native & misc. Every card cites the originating issue/discussion (#NNNN / D#NNNN). The AG Grid event gotcha from this PR is §5.1.

@KrilleGH — would something like this be useful as a starting point if a dedicated debugging section happens later? Happy to iterate or to seed it into a PR against `website/documentation/content/`.

evnchn:
And if yes maybe open new discussion. Otherwise this gets sunk.

KrilleGH:
> Took a stab at the "central place" idea — collected ~80 NiceGUI debugging gotchas from closed issues, discussions, and direct experience into a single rich-rendered page (dark-themed).

Wow, that looks cool. :open_mouth: I suppose you used AI to extract it?

I'd rather keep it in the same design as the main docs. From a quick glance it seems it needs some more prompting or editing.

3.1 may no longer be appropriate as there has been a code change to fix it. It may be better to omit these, to reduce the amount of information and make it easier to spot something that may fix a current issue.

5.1 misses the point, the card mentions something completely different.

So I'm not sure how useful it is, but at least a good start.

evnchn:
Did use AI. 

I think I also want it inside NiceGUI. @falkoschindler maybe you can just use AI to speed-merge those PRs (or find me) once see those are in "safe folders", with "debugging section" being one? 

Because https://gist.github.com/evnchn/11d8962581aec36fbd6f8f29e7cb4958 digs https://github.com/zauberzeug/nicegui/discussions and don't say you know the tips and tricks without the gist. At least even I didn't have all of them, and I think my memory of NiceGUI is already very sharp.

evnchn:
Moving the side-thread about a dedicated debugging / cookbook section (and the ~80-gotcha gist) to a proper discussion so it doesn't get lost now that this PR is merged:

👉 https://github.com/zauberzeug/nicegui/discussions/6070

@KrilleGH @falkoschindler — please continue the "central place" conversation over there. I summarised the points so far (including @KrilleGH's feedback on §3.1 / §5.1 and the styling) and listed open questions on home, format, merge path and handling stale entries.

KrilleGH:
> Did use AI.

I'm curious, which was the prompt?

evnchn:
@KrilleGH

> Read the thread of https://github.com/zauberzeug/nicegui/pull/6067#issuecomment-4501627756. I think your value would be to look across nicegui for "pro tips for debugging nicegui applications". If you agree then please have at it. Must rich-render for this one.

Where "Rich render" is an established term in my agent's memory that being to "make a pretty HTML website"

evnchn:
Oh also:

> yes but the real goldmine is github and other community forums

> Also you can look for debugging tips and tricks pulled by local agents or other agents in the homelab

And for the posted comment: 

> can you post the rich render to public gist and link below the original PR, after ensure all private information is stripped?

Issue #5939, opened 2026-04-07
Runtime Error when calling gui app with an entry point.


carsten-engelke:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python

```

### Description

I cant get nicegui to run if I call my app from an Entry Point. I created a litte demonstration:

Steps to perform:

    mkdir niceguitest
    cd niceguitest
    uv init --package

Create files in src/niceguitest:

logic.py

    class myTestAdd:
    def __init__(self, a, b):
        self.a = a
        self.b = b

    def add(self):
        return self.a + self.b
cli.py

    from niceguitest import logic
    
    def main() -> None:
        a = int(input("Enter first number: "))
        b = int(input("Enter second number: "))
        print(f"Result: {logic.myTestAdd(a, b).add()}")

    if __name__ == "__main__":
        main()

gui.py

    # This file is the main entry point for the GUI version of the niceguitest package. It defines a simple NiceGUI application that allows users to input two numbers and see their sum. The logic for the addition is defined in the logic.py module, and this file focuses on setting up the GUI and handling user interactions.
    from niceguitest.logic import myTestAdd

    def main() -> None:
        # This entry point is used when the user runs the command "niceguitest-gui" after installing the package with the "gui" extra.
        try:
            from nicegui import ui
            run_gui_app(ui)
        except ImportError:
            print("NiceGUI is not installed. Please install the 'gui' extra to use the GUI features with 'pip install niceguitest[gui]' .")
            from niceguitest.cli import main as cli_main
            cli_main()

    def run_gui_app(ui) -> None:
        from niceguitest import logic
        ui.label("Please enter two numbers to add:")
        a_input = ui.input(label="First number")
        b_input = ui.input(label="Second number")
        result_label = ui.label("Result will be shown here.")
        ui.button("Calculate", on_click=lambda: calculate_sum(a_input, b_input, result_label))
        ui.run()

    def calculate_sum(a_input, b_input, result_label) -> None:
        try:
            a = int(a_input.value)
            b = int(b_input.value)
            result = myTestAdd(a, b).add()
[...]

Issue #5280, opened 2025-10-14
NiceGUI renders blank page when behind a main guard on v3


ClementPinard:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

if __name__ in {"__main__", "__mp_main__"}:
    ui.label("Hello world")
    ui.run(reload=True)
```

### Description

This code works with nicegui<3, but with version 3.0.4 I get a blank page.

However, this code works:

```python
if __name__ == "__main__" or current_process().name != "MainProcess":
    ui.label("Hello world")
    ui.run()
```

I first came upon this problem when trying to run it as an entrypoint script.

Pyproject.toml would look like

```toml
[project.scripts]
main = "test_project.main:run"
```

and the code

```python
def run(reload=False):
    """This is the main function that gets run"""
    ui.label("Hello world")
    ui.run(reload=reload)


if __name__ in {"__main__", "__mp_main__"}:
    run(reload=True)
```

which is a pattern I saw in this discussion : https://github.com/zauberzeug/nicegui/discussions/3769

And this does not work as well: `uv run main` gives me blank page.

Not sure these two problems are related though, but they both disappear when I use nicegui==2.24.2 (last version <3)

### NiceGUI Version

3.0.4

### Python Version

3.13.7

### Browser

Chrome

### Operating System

macOS

evnchn:
Hello @ClementPinard thanks for the report. 

Can recreate. Using `print('__name__:', __name__)` we see `__name__` is `<run_path>`. 

That is from defaults of https://github.com/zauberzeug/nicegui/blob/b6ebab64da58e2ff2b776e4a5f6498b8862a9cef/nicegui/ui_run.py#L129

Short term fix: You can use main guard `if __name__ in {'__main__', '__mp_main__', '<run_path>'}:`

Long-term fix: We will have to consider `runpy.run_path(sys.argv[0], run_name='__main__')`

evnchn:
@falkoschindler So you are in progress of a PR on this? Because I am quite hesitant. 

I fear that we may run into cases of infinite-loop, if we do `runpy.run_path(sys.argv[0], run_name='__main__')` in our code, as it would basically make our invocation "indistinguishable". 

[...]

Issue #181, opened 2022-11-25
CLI pattern `if __name__ == "__main__": main()` does not work


nazarewk:
One of the patterns for running CLI programs is:

```python
from nicegui import ui

def main():
    ui.label('Hello NiceGUI!')
    ui.button('BUTTON', on_click=lambda: ui.notify('button was pressed'))
    ui.run(port=8082)

if __name__ == "__main__"
    main()
```

This does not work with `reload=True`, because the reloadable `__name__` is `__mp_main__` instead. Most likely related to #143 

So I made it work using second entry:

```python
from nicegui import ui

def main():
    ui.label('Hello NiceGUI!')
    ui.button('BUTTON', on_click=lambda: ui.notify('button was pressed'))
    ui.run(port=8082)

if __name__ in ("__main__", "__mp_main__"):
    main()
```

Any idea what causes this and/or whether it can be fixed in `nicegui` itself?

falkoschindler:
When Uvicorn is reloading the app (which happens for the first time right after starting the script), it re-evaluates main.py while searching for the nicegui:app object. In contrast to the original process, this "reloader" process has `__name__` "__mp_main__" with "mp" indicating "multi-processing". The main guard (successfully...) prevents the new process from evaluating `main()`, running `ui.run` and initializing `globals.config`. This is not what you want in this case.

Technically I don't see an easy solution for supporting the default main guard `if __name__ == '__main__':` for auto-reloading NiceGUI apps. Sure, you can extend the guard with `__mp_main__`. The question is, what is you motivation for the guard in the first place - besides it being a common pattern?

If you want to avoid the code in `main()` being evaluated a second time when starting with `reload=True`, you can wrap it in a page function:

```python
from nicegui import ui

@ui.page('/')
def main_page():
    ui.label('Hello NiceGUI!')
    ui.button('BUTTON', on_click=lambda: ui.notify('button was pressed'))

ui.run(port=8082)
```

Personally I'm not sure if you really need a main guard. Ideally your main script is short and concise, so that it does not need to be imported from anywhere.

nazarewk:
I want my main script to serve as a CLI primarily and be able to spin up a web gui as an option.

falkoschindler:
[...]

PR #6068, opened 2026-05-20
Replace misleading reload error for `python -m <pkg>` (#6066)

### LINKED ISSUE #6066: App as module fails when reloading is enabled (with wrong error message)
### First Check

### Example Code

```python

```

### Description

- Create folder called `xxx`.
- Add empty file `__init__.py` into `xxx`.
- Add file `__main__.py` into `xxx` with below contents.
- From level above `xxx` run `python -m xxx`.

`__main__.py`:
```python
from nicegui import ui
ui.label("xxx")
ui.run(reload=True)
```

This crashes with:
```python
   if __name__ == "__main__":
to allow for multiprocessing.

ERROR:    Application startup failed. Exiting.
```

After changing to `ui.run(reload=False)`, it runs as expected. From this, I would conclude that the automatic reloading has problems reloading apps written as modules. Is this expected?

The error message is wrong since the code definitely has `ui.run()` and does not hide it behind a wrong guard.

### NiceGUI Version

3.12.0

### Python Version

3.13.13

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

evnchn:
### Motivation

Closes #6066.

When launching as `python -m <pkg>` (a package directory with `__init__.py` + `__main__.py`), `ui.run(reload=True)` fails with a misleading error that tells the user to fix a main guard. The user has no main guard to fix, and CPython will refuse to re-run `__main__.py` regardless.

<details>
<summary>The current (misleading) error</summary>

```
   if __name__ == "__main__":
to allow for multiprocessing.
```

</details>

This same trap has been reported repeatedly: #181, #5280, #3769, #5939, and now #6066.

### Root cause

`multiprocessing.spawn._fixup_main_from_name` (`Lib/multiprocessing/spawn.py`) returns early when the parent was launched as `-m <pkg>`, per CPython policy:

> `__main__.py` files for packages, directories, zip archives, etc, run their "main only" code unconditionally, so we don't even try to populate anything in `__main__`, nor do we make any changes to `__main__` attributes

NiceGUI's reload model assumes the user's top-level script re-executes in the spawn child (that's how `app.config.has_run_config` gets set there). For `-m package`, spawn declines to re-execute, so `has_run_config` stays `False` → the misleading error fires.

### Implementation

In `_startup()`, before the legacy error, detect the specific failure mode and raise a targeted message naming the package and pointing at the two working alternatives. The legacy error still fires for the original main-guard case.

<details>
<summary>Detection signals (and why <code>sys.argv[0]</code> is the only reliable one in the spawn child)</summary>

A probe in the spawn child showed:

```
proc='SpawnProcess-1'
__main__.__spec__ = None
__main__.__file__ = None
sys.argv[0]       = '/private/tmp/ng-test/xxx/__main__.py'
```

`__spec__` and `__file__` are `None` because spawn refused to fix them up. `sys.argv[0]` is inherited from the parent unchanged. Combined with the `__init__.py` sibling check, this uniquely identifies a `-m <pkg>` invocation without false positives on regular scripts.

</details>

### Empirical verification

| Invocation | Outcome |
|---|---|
| `python -m xxx` (broken case) | **New** targeted error fires |
| `python script.py` | `NiceGUI ready` ✓ (new branch correctly skipped — `has_run_config` is True) |
| `python xxx/__main__.py` (direct) | `NiceGUI ready` ✓ (spawn's `_fixup_main_from_path` runs the file) |
| `python noguard.py` with a legitimate `if __name__ == "__main__":` guard | **Legacy** "remove the guard" error fires unchanged |

The new error reads:

```
Auto-reload is not supported with `python -m xxx`.
Pass `reload=False` to ui.run() or run `python xxx/__main__.py` directly.
```

### Alternatives considered

A **salvager** that re-runs the user's package via `runpy.run_module(pkg, run_name='__mp_main__')` at module load time was prototyped and confirmed to work end-to-end (including across multiple WatchFiles reload cycles — verified 3 consecutive edits, browser saw `MARKER_V1` → `V2` → `V3`).

It was rejected for being too aggressive:
- bypasses CPython's deliberate spawn policy
- runs user code as a side effect of `import nicegui`
- adds a 30-line helper to handle one obscure invocation form

An accurate error is a smaller, more honest change.

<details>
<summary>The full salvager prototype (~30 lines) — for contrast with the 7-line fix in this PR</summary>

```python
def _salvage_module_reload_child() -> None:
    """Workaround for `python -m <pkg>` + reload=True (issue #6066).

    multiprocessing.spawn._fixup_main_from_name refuses to re-execute __main__.py
    files in spawn children, so the reload child never runs the user's code and
    `app.config.has_run_config` stays False. We detect that situation here (at
    import time, before lifespan starts) and re-run the user's package.
    """
    import multiprocessing
    import os
    import runpy
    import sys
    if multiprocessing.current_process().name == 'MainProcess':
        return
    if app.config.has_run_config:
        return
    argv0 = sys.argv[0] if sys.argv else ''
    if not argv0.endswith('__main__.py'):
        return
    pkg_dir = os.path.dirname(argv0)
    if not os.path.exists(os.path.join(pkg_dir, '__init__.py')):
        return
    pkg_name = os.path.basename(pkg_dir)
    parent = os.path.dirname(pkg_dir)
    while os.path.exists(os.path.join(parent, '__init__.py')):
        pkg_name = os.path.basename(parent) + '.' + pkg_name
        parent = os.path.dirname(parent)
    runpy.run_module(pkg_name, run_name='__mp_main__', alter_sys=True)


_salvage_module_reload_child()
```

Placed at module level of `nicegui/nicegui.py`, after `app` is constructed.

</details>

### Progress

- [x] Pytests not added — the failure surfaces in a uvicorn-reload subprocess driven by the user's invocation form, which the existing test harness doesn't model.
- [x] Documentation not necessary — behavior is unchanged for all working invocations; only the error message for one specific broken invocation is improved.
- [x] No breaking changes.


falkoschindler (approved):
Thanks @evnchn — nice diagnosis, and I fully agree with choosing the accurate error over the `runpy` salvager.

I pushed a couple of small tweaks to the message while merging `main` in:

- **Dropped the `python <pkg>/__main__.py` suggestion.** Running a package's `__main__.py` as a plain script breaks any package that uses relative imports (`from . import ...` → `ImportError: attempted relative import with no known parent package`) — which is often exactly _why_ someone runs it via `-m`. So it's a footgun for real packages, and `reload=False` is the one universally-correct fix.
- **Stopped printing the package name.** It's not meant to be copy-pasted — the user just typed the command themselves — so `python -m` is explanatory enough. This also sidesteps a nested-package edge case (`python -m a.b.c` would otherwise have reported just `c`) without needing any dotted-name reconstruction.

The message now reads:

```
Auto-reload is not supported when running a package with `python -m`.
Pass `reload=False` to ui.run() to start the server.
```

Approving. 👍

PR #6071, opened 2026-05-20
Don't inject sad-face error_content when ui.sub_pages handles 404 (fixes #6069)

### LINKED ISSUE #6069: `ui.sub_pages` 404 sad-face persists after client-side navigation when page is first loaded at an unknown URL
### First Check

### Example Code

```python
from nicegui import ui

def root():
    ui.link('Go to home', '/')

ui.run(root)
```

### Description

Steps:

1. Start the app and open `/some_other_page` directly in the browser.
2. The full NiceGUI 404 error page (sad face, "404", "This page doesn't exist.") is shown — together with the `Go to home` link from `root()`.
3. Click `Go to home`.

Expected: the `Home` label is shown, the sad-face error block disappears.

Actual: the sub_pages content updates to `Home`, but the sad-face / "404" / "This page doesn't exist." block remains on the page.

### Why this happens

- When `/some_other_page` is requested, no FastAPI route matches, so the 404 exception handler in `nicegui/nicegui.py` wraps `core.root` as a page and serves it.
- The page handler in `nicegui/page.py` builds `root()` (link + sub_pages). The sub_pages cannot match the path and renders its inner `"404: sub page ... not found"` label.
- `page.py` then calls `_can_resolve_full_path`, sees the unresolved sub_pages, and does `with client: error_content(404, …)` — adding the sad-face block as a **sibling** of the sub_pages, directly in the client layout.
- Clicking the link triggers client-side navigation via `sub_pages_navigate`. The router (`sub_pages_router.py`) only re-renders the `SubPages` elements; the `error_content` column added on the initial response is never cleaned up, so it stays visible.

The existing test `test_navigate_from_404_to_root_path` doesn't cover this because it opens `/` first (no `error_content` injected), and only then navigates client-side to a bad path.

### Possible fix directions (open for discussion)

1. Don't inject the global `error_content` block when a `SubPages` is already rendering its own 404. The HTTP status stays 404, but only the sub_pages 404 line shows — SPA navigation stays clean.
2. Track the injected `error_content` elements and remove them whenever `SubPages._handle_open` resolves a new path.

### NiceGUI Version

3.12.0.post4.dev0+ec74aa6d4

### Python Version

3.11.7

### Browser

Chrome

### Operating System

macOS

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #6069.

When a page handler returns a 404 because `ui.sub_pages` cannot resolve the path, `nicegui/page.py` was injecting `error_content` (the sad-face block) as a sibling of the user's `root()` chrome. Client-side navigation via `sub_pages_router._handle_open` only re-renders `SubPages` elements, so the sibling sad-face leaked past the recovered navigation — that's the bug in #6069.

### Implementation

This PR implements Falko's "fix 1" proposal from the issue: delete the `error_content` injection in `page.py` and let `SubPages` continue to render its own inline `404: sub page X not found` label inside its own slot. Because that label lives inside the slot, `_show()` naturally cleans it up on SPA navigation. HTTP 404 status is unchanged.

Changes:
- `nicegui/page.py`: drop the `with client: error_content(...)` block and the now-unused `HTTPException` / `error_content` imports.
- `tests/test_sub_pages.py`:
  - 4 existing assertions updated from `'HTTPException: 404: /bad_path not found'` (sad-face text) to `'404: sub page /bad_path not found'` (inline label).
  - New `test_navigate_from_initial_404_does_not_leak_sad_face` regression test for #6069.

Test plan:
- CI on my fork is green (mypy, pre-commit, pylint, quick-test/3.10/pytest 20m44s): https://github.com/evnchn/nicegui/pull/159
- Manually confirm the original repro from #6069: direct-URL `/bad_path` shows only the inline label inside chrome, and clicking the `Go to home` link leaves only home content (no stale 404 text).

Note: this is a user-visible behavior change — apps that relied on the styled sad-face overlay appearing on `ui.sub_pages` 404s will now see only the inline label inside their own chrome. HTTP status and the `show_404=False` escape hatch are unchanged. See the trade-off discussion below.

### Progress

---

## Note to @falkoschindler

A couple of things I'd like your call on before merging:

**Milestone — 3.12.1 patch or 3.13?**

I'd lean 3.13. The fix changes user-visible behavior (apps relying on the sad-face overlay on sub_pages 404 will stop seeing it), so it doesn't feel right for a patch release even though it closes a bug. A minor release with an explicit changelog note feels safer.

**Does fix 1 really sit well with you?**

You proposed two directions in the issue. I went with fix 1 because, framing-wise, when `root()` contains `ui.sub_pages` the surrounding UI is user chrome and the framework arguably shouldn't be drawing a styled 404 page on top of it. But there's a real trade-off and I want to flag it rather than just advocate.

### Neutral review of fix 1

**Pros**
- Smallest possible patch (2 imports + 3 lines deleted in `page.py`).
- Respects user chrome: once you opt into `ui.sub_pages`, you've taken responsibility for the page surface.
- Closes #6069 by construction — there's no sibling element to leak past SPA nav, so no router cleanup needed now or later.
- Preserves HTTP 404 status; bookmarks / SEO / `curl` outputs unchanged.
- Consistent with the existing `show_404=False` escape hatch, which already treats the sub_pages 404 as user-controlled.

**Cons / things to consider**
- The inline `SubPages` 404 is currently a plain `Label('404: sub page X not found')` — debug-quality, leaks framework-speak ('sub page'). Apps that were quietly relying on the sad-face overlay as their 404 UI lose that polish and just see a small line of text inside their chrome.
- There's no customization hook yet. Users who want a pretty 404 today have to either `show_404=False` plus a conditional render in `root()`, or wait for a follow-up (e.g. `not_found_builder=` callback on `ui.sub_pages`, or a `'*'` wildcard route). Worth tracking as a separate issue regardless of which fix you pick.
- 4 existing tests in `test_sub_pages.py` asserted on the sad-face text; that's evidence that downstream code/snippets/demos may also assume the overlay is present. The changelog entry should call this out so users aren't surprised.
- Scope check: this PR only touches the sub_pages-aware 404 path. Direct `HTTPException(404)` raises and FastAPI-level 404s (no matching route, no root) still render `error_content` via `nicegui/nicegui.py`. That's intentional — the framework still owns those surfaces — but worth noting.

**Trade-off framing**
- Fix 1 says: framework should not own UX inside user chrome.
- Fix 2 (track sibling + cleanup on resolve) says: framework provides a polished default 404, but is SPA-aware.
- Both close #6069. The decision is framework philosophy more than engineering. I'm comfortable with either; just flagging that fix 2 wouldn't be wrong, only different.

If you'd rather go with fix 2, happy to rework — say the word.

### DIFF
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -202,8 +202,6 @@ class page:
                     task.add_done_callback(check_for_late_return_value)
 
             if not await client.sub_pages_router._can_resolve_full_path(client):  # pylint: disable=protected-access
-                # The SubPages element renders its own 404 label inside the page; don't inject
-                # a sibling sad-face block on top of user-supplied chrome. Just return 404 status.
                 log.warning(f'{request.url} not found')
                 return client.build_response(request, 404)
 
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -7,11 +7,10 @@ from functools import wraps
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
-from fastapi import HTTPException, Request, Response
+from fastapi import Request, Response
 
 from . import background_tasks, binding, core, helpers
 from .client import Client, ClientConnectionTimeout
-from .error import error_content
 from .favicon import create_favicon_route
 from .language import Language
 from .logging import log
@@ -203,10 +202,9 @@ class page:
                     task.add_done_callback(check_for_late_return_value)
 
             if not await client.sub_pages_router._can_resolve_full_path(client):  # pylint: disable=protected-access
-                # Handle 404 gracefully without re-raising exception (similar to 404 handler when no root function)
+                # The SubPages element renders its own 404 label inside the page; don't inject
+                # a sibling sad-face block on top of user-supplied chrome. Just return 404 status.
                 log.warning(f'{request.url} not found')
-                with client:
-                    error_content(404, HTTPException(404, f'{client.sub_pages_router.current_path} not found'))
                 return client.build_response(request, 404)
 
             if isinstance(result, Response):  # NOTE if setup returns a response, we don't need to render the page
--- a/tests/test_sub_pages.py
+++ b/tests/test_sub_pages.py
@@ -976,7 +976,7 @@ def test_on_path_changed_event(screen: Screen):
     assert calls == {'index': 2, 'main': 1, 'other': 2}
 
     screen.open('/bad_path')
-    screen.should_contain('HTTPException: 404: /bad_path not found')
+    screen.should_contain('404: sub page /bad_path not found')
     assert paths == ['/other']
     assert calls == {'index': 3, 'main': 2, 'other': 2}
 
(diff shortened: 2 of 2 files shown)


falkoschindler (approved):
Awesome! Claude and I have nothing to object. Fix 1 is the right decision and we're on 3.13 now anyway. Let's merge! 🚀

PR #6074, opened 2026-05-21
Fix auto-prune timers being skipped in script mode

### LINKED ISSUE #6072: Auto-prune timers never run in script mode (regression from #6006)
### Reproduction

Script (`leak.py`):

```python
from nicegui import app, core, ui

ui.label('any module-level UI puts NiceGUI in script mode')

@app.get('/_count')
def count():
    return {
        'script_mode': core.script_mode,
        'clients': len(Client.instances),
        'routes': len(app.routes),
    }

ui.run(reload=False, port=8080)
```

Driver:

```bash
python leak.py &
sleep 4
for i in $(seq 1 1000); do curl -s -o /dev/null http://localhost:8080/; done
curl -s http://localhost:8080/_count        # → clients=1001 routes=1010
sleep 70                                    # past default client_age_threshold=60s
curl -s http://localhost:8080/_count        # → clients=1001 routes=1010 (unchanged)
```

### Expected

After `sleep 70`, `Client.prune_instances` (registered via `app.timer(10, Client.prune_instances)` in `nicegui/nicegui.py:147`) should have run several times and pruned the 1000 stale clients whose `created` is older than 60 s.

### Actual

`prune_instances` is **never called**. Clients and their routes accumulate without bound.

### Bisect

Regression introduced by **#6006** (commit `8a527282`, 2026-05-01) — *Fix duplicate `app.timer` and lifecycle handler registration in script mode*.

| Commit | Clients after 130 s grind |
|---|---|
| `8a527282^` (pre-#6006) | 1 |
| `8a527282` (current `main`) | 1001 |

<details>
<summary>Root cause</summary>

`Timer.__init__` in `nicegui/timer.py:42-45`:

```python
coroutine = self._run_once if once else self._run_in_loop
if self._skip_registration():
    return
background_tasks.create_or_defer(coroutine(), name=str(callback))
```

`Timer._skip_registration()` returns `core.is_script_mode_re_execution()`, which is `script_mode and app.is_started`.

In `nicegui/nicegui.py:_startup`:

```python
app.start()                                 # sets _state = STARTED → is_started = True
background_tasks.create(binding.refresh_loop(), …)
app.timer(10, Client.prune_instances)       # ← skipped: is_started=True AND script_mode=True
app.timer(10, Slot.prune_stacks)            # ← skipped
app.timer(10, prune_tab_storage)            # ← skipped
if app.storage.secret is not None:
    app.timer(10, prune_user_storage)       # ← skipped
```

\`app.start()\` flips \`is_started=True\` before the four internal \`app.timer(...)\` registrations. In script mode this puts the \`_skip_registration\` predicate permanently in the True branch, so the four NiceGUI-internal auto-prune timers are misclassified as per-client re-execution registrations and skipped — even though \`_startup\` runs **once** per process.

\`Slot.prune_stacks\`, \`prune_tab_storage\`, and \`prune_user_storage\` are affected the same way; this is not just a route/client leak — tab storage and slot stacks are also no longer pruned in script mode.

</details>

<details>
<summary>Why the fix in #6006 over-applies</summary>

PR #6006 correctly guards against duplicate registration when **user code at module level** does \`app.timer(...)\`: in script mode the user's script is re-executed via \`runpy.run_path\` per client connection, so module-level \`app.timer\` would otherwise stack a fresh timer each time.

But the same predicate is checked unconditionally inside \`Timer.__init__\`, which is also reached from NiceGUI's **internal** \`nicegui.py:_startup\`. Those four calls happen exactly once per process, never via the per-client re-execution path. The predicate has no way to distinguish "internal \`_startup\`-registered timer" from "user-script-registered timer".

</details>

<details>
<summary>Fix sketches (leaving the choice to maintainers)</summary>

Three plausible directions, in increasing order of intrusion:

1. **Reorder \`_startup\`** so the four internal \`app.timer(...)\` calls run **before** \`app.start()\` (while \`is_started=False\`). One-line move; \`_skip_registration\` would then return False for them. Need to verify the timers don't depend on anything \`app.start()\` produces.
2. **Internal-vs-user distinction**: add \`_internal: bool = False\` to \`Timer.__init__\` (or a class-level marker on a NiceGUI-only \`_AppTimer\` subclass) and bypass \`_skip_registration\` when set. The four internal calls in \`_startup\` pass \`_internal=True\`.
3. **Disable script-mode handling for the duration of \`_startup\`**: wrap the internal section in a \`with core.script_mode_suspended():\` context manager.

Option 1 is the smallest diff; option 2 is the most explicit.

</details>

### Environment

- NiceGUI \`main\` @ \`ba4b7c4\`
- Python 3.12.11, macOS 15.7.1 arm64
- Empirical confirmation: same code at \`8a527282^\` cleans up; at \`8a527282\` it doesn't.

falkoschindler:
### Motivation

Fixes #6072.

PR #6006 added a `_skip_registration` guard to `Timer.__init__` so user-script `app.timer(...)` calls at module level wouldn't be re-registered on every per-client re-execution in script mode. The guard returns `True` whenever `script_mode and app.is_started`.

That predicate also fires for NiceGUI's own four internal `app.timer(...)` calls in `_startup`, because they ran **after** `app.start()` had set `is_started=True`. Result in script mode: `Client.prune_instances`, `Slot.prune_stacks`, `prune_tab_storage`, and (when storage secret is set) `prune_user_storage` were never scheduled. Clients, their routes, slot stacks, and tab storage accumulated without bound — see #6072 for the repro.

### Implementation

Move the internal startup wiring into `App.start()` itself, before the `_state = State.STARTED` flip:

- The four `app.timer(10, …)` registrations for the prune helpers.
- The `background_tasks.create(binding.refresh_loop(), …)` task (not affected by the bug, but moved for symmetry — it's part of the same "start NiceGUI internals" wiring).

The two helpers `prune_tab_storage` / `prune_user_storage` are moved alongside, from `nicegui/nicegui.py` into `nicegui/app/app.py`, since they manipulate `app.storage` internals and now live next to their only registration site.

Because the timers register while `_state == STARTING` (so `is_started == False`), `_skip_registration()` no longer matches them. The ordering invariant — *internal timers must register before `STARTED` is set* — is now physically inside the method that owns the state machine, which makes accidental re-regression much less likely than when the registrations lived in `_startup` separated from `app.start()` by other code.

Test update: `tests/test_storage.py` was importing the prune helpers via `nicegui.prune_tab_storage` / `nicegui.prune_user_storage` (the `nicegui.nicegui` module); switched to importing them from `nicegui.app.app` to follow the move. The helpers aren't referenced anywhere user-facing (website, examples, README), so no compatibility shim is provided.

### Progress

- [x] Pytests are not necessary (script mode is not easily testable at the moment).

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -1,7 +1,9 @@
+import asyncio
 import inspect
 import os
 import platform
 import signal
+import time
 import urllib
 from collections.abc import Callable, Iterator
 from enum import Enum
@@ -10,7 +12,7 @@ from pathlib import Path
 from fastapi import FastAPI, HTTPException, Request, Response
 from fastapi.responses import FileResponse
 
-from .. import background_tasks, core, helpers
+from .. import background_tasks, binding, core, helpers
 from ..client import Client
 from ..context import context
 from ..elements.mixins.color_elements import QUASAR_COLORS
@@ -18,8 +20,9 @@ from ..logging import log
 from ..native import NativeConfig
 from ..observables import ObservableSet
 from ..server import Server
+from ..slot import Slot
 from ..staticfiles import CacheControlledStaticFiles
-from ..storage import Storage
+from ..storage import PersistentDict, Storage
 from .app_config import AppConfig
 from .range_response import get_range_response
 
@@ -79,6 +82,12 @@ class App(FastAPI):
             self.safe_invoke(t)
         self.on_shutdown(self.storage.on_shutdown)
         self.on_shutdown(background_tasks.teardown)
+        background_tasks.create(binding.refresh_loop(), name='refresh bindings')
+        self.timer(10, Client.prune_instances)
+        self.timer(10, Slot.prune_stacks)
+        self.timer(10, prune_tab_storage)
+        if self.storage.secret is not None:
+            self.timer(10, prune_user_storage)
         self._state = State.STARTED
 
     async def stop(self) -> None:
@@ -393,3 +402,31 @@ class App(FastAPI):
         for client in Client.instances.values():
             if path is None or client.page.path == path:
                 yield client
+
+
+async def prune_tab_storage(*, force: bool = False) -> None:
+    """Prune tab storage that is older than the configured ``max_tab_storage_age``."""
+    tab_storages = core.app.storage._tabs  # pylint: disable=protected-access
+    for tab_id, tab in list(tab_storages.items()):
+        if force or time.time() > tab.last_modified + core.app.storage.max_tab_storage_age:
+            tab.clear()
+            if isinstance(tab, PersistentDict):
+                await tab.close()
+            del tab_storages[tab_id]
+
+
+async def prune_user_storage(*, force: bool = False) -> None:
+    """Remove user storage objects without a client session."""
+    client_session_ids = {client.request.session['id'] for client in Client.instances.values()}
+    storages_to_close: list[PersistentDict] = []
+    now = time.time()
+    user_storages = core.app.storage._users  # pylint: disable=protected-access
+    for session_id in list(user_storages):
+        if session_id not in client_session_ids:
+            age = now - user_storages[session_id].last_modified
+            if force or age > 10.0:  # NOTE: do not remove storages created by middleware and still wait for client
+                storages_to_close.append(user_storages.pop(session_id))
+    results = await asyncio.gather(*[storage.close() for storage in storages_to_close], return_exceptions=True)
+    for result in results:
+        if isinstance(result, Exception):
+            log.exception(result)
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -1,7 +1,6 @@
 import asyncio
 import inspect
 import mimetypes
-import time
 import urllib.parse
 from contextlib import asynccontextmanager
 from pathlib import Path
@@ -13,7 +12,7 @@ from fastapi.exception_handlers import http_exception_handler
 from fastapi.responses import FileResponse, Response
 from starlette.exceptions import HTTPException as StarletteHTTPException
 
-from . import air, background_tasks, binding, core, favicon, helpers, json, run, welcome
+from . import air, core, favicon, helpers, json, run, welcome
 from .app import App
 from .client import Client
 from .dependencies import dynamic_resources, esm_modules, js_components, libraries, resources, vue_components
@@ -22,8 +21,6 @@ from .json import NiceGUIJSONResponse
 from .logging import log
 from .page import page
 from .page_arguments import PageArguments
-from .persistence import PersistentDict
-from .slot import Slot
 from .staticfiles import CacheControlledStaticFiles
 from .version import __version__
 
@@ -143,12 +140,6 @@ async def _startup() -> None:
     core.loop = asyncio.get_running_loop()
     run.setup()
     app.start()
-    background_tasks.create(binding.refresh_loop(), name='refresh bindings')
-    app.timer(10, Client.prune_instances)
-    app.timer(10, Slot.prune_stacks)
-    app.timer(10, prune_tab_storage)
-    if app.storage.secret is not None:
-        app.timer(10, prune_user_storage)
     air.connect()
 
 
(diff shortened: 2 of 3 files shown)


evnchn:
> _Posted independently by a Claude Code session reviewing this PR via the `requesting-code-review` skill — not affiliated with the PR author's session. Same reviewer that looked at #6073. Verdict and reasoning below; push back on anything off._

**Verdict: ready to merge.** Prefer this over #6073.

Fix is mechanically correct: in `nicegui/app/app.py:80-93` the four `self.timer(...)` registrations and the `binding.refresh_loop` task are placed BEFORE `self._state = State.STARTED`, so `core.is_script_mode_re_execution()` (`script_mode and app.is_started`) returns `False` for these internal calls. Bug genuinely closed.

Co-locating the registrations inside `App.start()` immediately above the state flip makes the ordering invariant local and obvious — a reader sees "register internal pollers" and "mark started" in the same six-line block, rather than scattered across `_startup` separated by other code. That's tighter than the `_internal=True` marker in #6073, and there's no new private API parameter to remember at future callsites.

Helper relocation (`prune_tab_storage` / `prune_user_storage` from `nicegui/nicegui.py` → `nicegui/app/app.py`) is sound — these only exist to be scheduled by `App.start()`, and now live next to their sole call site. GitHub code search confirms no other importer in the repo or any user-facing surface (examples, website, README).

<details>
<summary>Verified during review</summary>

- All four call sites preserved: `Client.prune_instances`, `Slot.prune_stacks`, `prune_tab_storage`, `prune_user_storage` (latter still gated by `self.storage.secret is not None`).
- Imports clean — `PersistentDict` is re-exported from `nicegui/storage.py:18`, so `from ..storage import PersistentDict, Storage` resolves without circular-import risk.
- `tests/test_storage.py:9-10` correctly imports the helpers from their new home; two invocations at `tests/test_storage.py:217,373` updated to match. Stale `nicegui` import dropped.
- `background_tasks.create(binding.refresh_loop(), ...)` moved alongside but its relative ordering vs. `air.connect()` is unchanged (still: `refresh_loop` before `air.connect`). No functional concern.

</details>

<details>
<summary>Minor non-blocking observations</summary>

1. **Add a single-line ordering-invariant comment** above `self.timer(10, Client.prune_instances)` in `App.start()` — something like `# NOTE: register internal timers BEFORE flipping _state to STARTED; see #6072.` This is the single most fragile invariant in the fix. Without it, the next contributor refactoring `App.start()` for some unrelated reason could silently re-introduce the bug. Adding the comment turns "fragile to reordering" (the one thing #6073 had going for it) into a non-issue.

2. **Optional 5-line monkeypatch regression test.** Full script-mode end-to-end testing is awkward, but a unit-level assertion that `App.start()` registers four timers when called with `script_mode=True`, and that `core.app._state == STARTING` at the moment `self.timer(...)` runs, would guard the ordering invariant automatically.

3. **Scope creep, very mild.** The helper relocation is logically a separate refactor from the bug fix. Both belong, but a reviewer reading `git blame` later loses the ability to see "the bug fix" as a small atomic change. If splitting commits is desired: minimal fix is ~6 lines, helper relocation is the other ~30. Not a blocker.

4. **Add a one-line module comment** on `prune_tab_storage` / `prune_user_storage` like `# Internal helpers — scheduled by App.start(). Not public API.` so the import path change (`nicegui.prune_tab_storage` → `nicegui.app.app.prune_tab_storage`) sets expectations for future contributors.

5. **Re-export shim?** Probably not — these helpers are private and external code shouldn't reference them. PR description's claim of zero external references holds.

</details>

<details>
<summary>Comparative note: #6073 vs #6074</summary>

Both fix the same bug; the difference is *where* the "this is internal" knowledge lives.

- **#6073 (option 2, `_internal=True` kwarg):** explicit at the callsite. `app.timer(10, prune_tab_storage, _internal=True)` is self-documenting and survives arbitrary reordering. Cost: introduces a private API parameter on `Timer.__init__`, and any future internal `app.timer(...)` callsite must remember to set the flag — easy to forget, no static guarantee.

- **#6074 (option 1, reorder + co-locate):** the knowledge is structural. No flag to forget because the registration happens inside `App.start()` BEFORE `is_started` flips. Cost: relies on textual ordering of statements inside one method body. Mitigated significantly by co-locating with the state flip — a future contributor would have to deliberately separate them.

As a maintainer I'd prefer #6074:
1. No new private API parameter — `Timer` stays simpler.
2. The ordering constraint is physically visible (six adjacent lines) rather than a flag that has to be remembered.
3. `_skip_registration` keeps a single, simple predicate — no carve-out branch.
4. Bonus: relocating the helpers next to their only registration site fixes a long-standing "why is this in `nicegui.py`?" weirdness.

The one thing #6073 has over #6074 is robustness to reordering. Patch that here with the single comment line suggested above, and #6074 is strictly better.

</details>

Worth closing #6073 explicitly in this PR's merge message, and mentioning in release notes that auto-prune was broken in script mode between PR #6006 and the fix — long-lived script-mode deployments may want to check for leaked tab storage.

evnchn (approved):
No comments from copilot, so even better than #6073. Claude agrees. I can't review due to pain issues but this is enough to approve.

PR #6078, opened 2026-05-26
Fix Event.emitted() dropping the emitted arguments

### LINKED ISSUE #6077: `Event.emitted()` returns `None` for events emitted with arguments
### First Check

### Example Code

```python
import asyncio
from nicegui import Event

async def main():
    e: Event[int] = Event()

    async def waiter():
        return await e.emitted(timeout=1)

    task = asyncio.create_task(waiter())
    await asyncio.sleep(0)  # let waiter subscribe
    e.emit(42)
    result = await task
    print('expected 42, got', result)  # -> got None

asyncio.run(main())
```

### Description

`Event.emitted()` is documented as *"Wait for an event to be fired and return its arguments."* In practice it returns `None` for any event whose `emit()` is called with arguments. Callers awaiting a value (e.g. a frame, payload, message) silently receive `None` instead of the emitted argument.

The bug is in the interaction between `Event.emitted()`'s internal variadic receiver and `helpers.expects_arguments`:

1. [`Event.emitted()`](nicegui/event.py#L111-L128) registers an internal callback declared as `def callback(*args, **kwargs)`.
2. When the event fires, `Callback.run()` ([nicegui/event.py:30-33](nicegui/event.py#L30-L33)) decides whether to forward arguments based on `helpers.expects_arguments(self.func)`.
3. [`expects_arguments()`](nicegui/helpers/functions.py#L31-L36) explicitly excludes `VAR_POSITIONAL` and `VAR_KEYWORD` parameters, so a purely variadic callable is reported as "expects no arguments".
4. Consequently `Callback.run()` invokes the receiver with no arguments, the receiver hits the `else None` branch, and the future resolves to `None`.

### Impact

Every call to `Event.emitted()` on an event whose `emit()` passes arguments yields `None`. The bug doesn't raise — it just returns the wrong value — so it's easily masked by surrounding `except Exception` handlers and fallback paths.

### Suggested fix

The most localized fix is in `helpers.expects_arguments()`: return `True` when the signature has only `VAR_POSITIONAL`/`VAR_KEYWORD` parameters. A purely variadic callable *can* always accept arguments — labelling it as "expects nothing" is incorrect.

Alternatively, `Callback.run()` could special-case fully variadic signatures and always forward `*args, **kwargs`.

### NiceGUI Version

3.12.1

### Python Version

3.11

### Browser

Chrome

### Operating System

macOS

### Additional Context

Reported by @pascalzauberzeug.

falkoschindler:
### Motivation

`Event.emitted()` is documented as _"Wait for an event to be fired and return its arguments"_, but it currently returns `None` for any event whose `emit()` is called with arguments.

Reproducer:

```python
import asyncio
from nicegui import Event

async def main():
    e: Event[int] = Event()

    async def waiter():
        return await e.emitted(timeout=1)

    task = asyncio.create_task(waiter())
    await asyncio.sleep(0)
    e.emit(42)
    print('expected 42, got', await task)  # -> got None

asyncio.run(main())
```

Internally, `Event.emitted()` registers a `def callback(*args, **kwargs)` receiver and waits on a future that the receiver is supposed to fill. The receiver is dispatched through `Callback.run`, which only forwards arguments when `expect_args` is true. `expect_args` is auto-detected via `helpers.expects_arguments`, which returns `False` for purely variadic signatures — by design, since for a UI handler "no required parameters" correctly means "don't forward anything". So the receiver runs with `args=()` and stores `None` on the future.

The helper is right: a `*args, **kwargs` wrapper doesn't _expect_ arguments, it _accepts_ them. The missing piece is a way for the caller to opt into forwarding when it knows its wrapper is intentionally permissive.

### Implementation

- Add an `expect_args: bool | None = None` keyword argument to `Event.subscribe()`. When `None` (the default), behavior is identical to before — same auto-detection, including the existing special case for bound `Event.emit` / `Event.call` chaining. When set explicitly, the caller's choice wins.
- Use `self.subscribe(callback, expect_args=True)` inside `Event.emitted()` so its variadic internal receiver actually receives the emitted arguments and writes them to the future.
- Add a `test_await_emitted` regression test.

No public-API changes for existing callers; the new parameter is opt-in.

### Progress

- [x] Documentation is not necessary (internal behavior fix; existing `Event.emitted()` docstring already describes the intended contract).

### DIFF
--- a/nicegui/event.py
+++ b/nicegui/event.py
@@ -56,6 +56,7 @@ class Event(Generic[P]):
         self.instances.add(self)
 
     def subscribe(self, callback: Callable[P, Any] | Callable[[], Any], *,
+                  expect_args: bool | None = None,
                   unsubscribe_on_delete: bool | None = None) -> None:
         """Subscribe to the event.
 
@@ -65,6 +66,8 @@ class Event(Generic[P]):
         to prevent memory leaks.
 
         :param callback: the callback which will be called when the event is fired
+        :param expect_args: whether to forward the event's arguments to the callback
+            (default: ``None`` meaning auto-detected from the callback's signature, *added in version 3.13.0*)
         :param unsubscribe_on_delete: whether to unsubscribe the callback when the current client is deleted
             (default: ``None`` meaning the callback is automatically unsubscribed if subscribed from within a UI context)
         """
@@ -77,7 +80,7 @@ class Event(Generic[P]):
             expect_args=helpers.expects_arguments(callback) or (
                 isinstance(getattr(callback, '__self__', None), Event) and
                 getattr(callback, '__name__', None) in {'emit', 'call'}
-            ),
+            ) if expect_args is None else expect_args,
             filepath=frame.f_code.co_filename,
             line=frame.f_lineno,
         )
@@ -119,7 +122,7 @@ class Event(Generic[P]):
             if not future.done():
                 future.set_result(args[0] if len(args) == 1 else args if args else None)
 
-        self.subscribe(callback)
+        self.subscribe(callback, expect_args=True)
         try:
             return await asyncio.wait_for(future, timeout)
         except TimeoutError as error:
--- a/tests/test_event.py
+++ b/tests/test_event.py
@@ -86,6 +86,19 @@ async def test_exception_during_emit(user: User, caplog: pytest.LogCaptureFixtur
     caplog.records.pop(0)
 
 
+async def test_await_emitted(user: User):
+    event = Event()
+    app.timer(0.1, lambda: event.emit(42))
+
+    @ui.page('/')
+    async def page():
+        number = await event.emitted()
+        ui.label(f'Emitted number: {number}')
+
+    await user.open('/')
+    await user.should_see('Emitted number: 42')
+
+
 async def test_exception_during_call(user: User):
     event = Event()
     event.subscribe(lambda: print(1 / 0))


nicegui/event.py
```diff
@@ -65,6 +66,8 @@ def subscribe(self, callback: Callable[P, Any] | Callable[[], Any], *,
         to prevent memory leaks.
 
         :param callback: the callback which will be called when the event is fired
+        :param expect_args: whether to forward the event's arguments to the callback
+            (default: ``None`` meaning auto-detected from the callback's signature, *added in version 3.13.0*)
```
evnchn, line 70: *Posted via Claude Code on Evan's behalf.*

Tiny docstring polish (non-blocking): name the actual rule so the param stops looking arbitrary, and keep the `*added in version*` tag inline + lowercase to match the house convention (`timer.py`, `element.py`) rather than a standalone line.

```suggestion
        :param expect_args: whether to forward the event's arguments to the callback;
            by default (``None``) this is inferred from the signature: forwarded only to a callback
            that declares a parameter for them, so pass ``True`` for a variadic ``*args`` callback
            that should receive them (*added in version 3.13.0*)
```

evnchn:
*Posted via Claude Code on Evan's behalf.*

Happy with this PR — correct, minimal, backward-compatible, and I verified it end-to-end. One tiny docstring suggestion above; non-blocking.

<details>
<summary>Verification &amp; reasoning</summary>

**Empirically verified (Python 3.12.11):**
- The repro from the PR body flips `None` → `42` (base `2c8402d4` vs head `62b32407`).
- Edge cases: multi-arg → `(1, 2)`, zero-arg → `None`, explicit `expect_args=False` suppresses, `expect_args=True` forwards.
- Full `tests/test_event.py` → 11 passed.
- Precedence of `(... or ...) if expect_args is None else expect_args` confirmed via AST to group as `((A or B) if C else D)`.

**Why the scoping is right** (vs. the tempting “simpler” alternative of making `expects_arguments` forward to variadic globally): that helper is the framework-wide “forward args only to a callback that requires one” gate — used in `events.py`, `client.py`, `page.py` and `event.py` — so widening it would feed unexpected args to every `*args` / `partial` / wrapper handler across all apps. Setting the already-existing `Callback.expect_args` field directly via `subscribe(expect_args=...)` is the well-contained move: it's the same bit `Callback.run` already keys on, default `None` changes nothing, and only callers that *know* their variadic receiver wants the args opt in.

Backward-compat: new param defaults to `None` (identical behavior); `Callback.expect_args` already existed.
</details>

evnchn (approved):
Am fine with the PR. 

The jewel of the suggestion block is 

````
``True`` for a variadic ``*args`` callback
````

which is not apparent in the old docstring version, and could lead one to be confused with "why is it not working for me, `expect_args=None`"

nicegui/event.py
falkoschindler: Thanks for the look. Going to keep the docstring as-is and merge — two reasons against the longer prose:

1. "Forwarded only to a callback that declares a parameter for them" isn't quite the inference rule. `def f(x=1)` declares a parameter but auto-detect returns `False` because of the default — the actual predicate is "has a required, non-variadic parameter". The simplified version reads inviting until someone hits the `x=1` case and feels misled.

2. Documenting the inference algorithm is the inverse of what readers reaching for `expect_args=True` need. They're already past "how does auto-detect work"; spelling the rule out risks inviting the exact question the bug starts with ("I have `*args`, why doesn't that count?").

(Side note on the style point: the `*added in version*` tag is already inline + lowercase in the current diff — it sits inside the same parenthetical, not on a standalone line.)

Issue #5087, opened 2025-08-28
Faced a trouble with using Ag-Grid's Infinite Row Model


denis-x:
Hello,

I have recently noted, that in my application Ag-Grid's Infinite Row Model discontinued to operate.
I have traced this issue, that function assigned in grid's option ":onGridReady" is not executed, because is redefined in 
https://github.com/zauberzeug/nicegui/blob/20b55d2fd46949b5d7c8faa0d6a272aa78d1d44e/nicegui/elements/aggrid.js#L45.
When I have commented line 45 of aggrid.js it starts to work as expected.

Because Ag-Grid's "datasource" cannot be assigned with nicegui on python side, it requires to execute JS function on gridReady.
Such function has construction like "params.api.setGridOption('datasource', dataSource);" to assign datasource. 
However, approach with grid.on('gridReady', js_handler=... does not work, because params.api is not accessible (by AI not serialized).

Please, reassess or make conditional aggrid.js#L45 to allow Ag-Grid's Infinite Row Model.
BR Denis

falkoschindler:
Thanks for reporting this issue, @denis-x!
Can you, please, share a minimum reproduction? Then we can tackle the problem more efficiently.

denis-x:
```py
from nicegui import ui

custom_js = """
<script>
onGridReady = params => {

  fetch("https://www.ag-grid.com/example-assets/olympic-winners.json")
    .then((response) => response.json())
    .then(function (data) {
      const dataSource = {
        rowCount: undefined, // behave as infinite scroll

        getRows: (params) => {
          console.log("asking for " + params.startRow + " to " + params.endRow);

          // At this point in your code, you would call the server.
          // To make the demo look real, wait for 500ms before returning
          setTimeout(() => {
            // take a slice of the total rows
            const rowsThisPage = data.slice(params.startRow, params.endRow);
            // if on or after the last page, work out the last row.
            let lastRow = -1;
            if (data.length <= params.endRow) {
              lastRow = data.length;
            }
            // call the success callback
            params.successCallback(rowsThisPage, lastRow);
          }, 500);
        },
      };

    params.api.setGridOption('datasource', dataSource);
    });
};
</script>
"""

ui.add_head_html(custom_js)

grid = ui.aggrid(
    {
        'defaultColDef': {'flex': 1},
        'columnDefs': [
            {'headerName': 'ID', 'maxWidth': 100, 'valueGetter': 'node.id', 'field': 'name'},
[...]

PR #6080, opened 2026-05-27
Fix blank ui.aggrid when flex columns use the infinite row model

evnchn:
### Motivation

`ui.aggrid` renders a **blank** grid when columns use `flex` together with the **infinite row model**. The data loads (rows and headers are in the DOM), but every cell ends up `visibility: hidden`, so the user sees nothing. Reported in the comments of #5087; minimal repro:

```python
from nicegui import ui

ui.aggrid({
    'defaultColDef': {'flex': 1},
    'columnDefs': [{'field': 'athlete'}, {'field': 'age'}, {'field': 'country'}],
    'rowData': None,
    'rowModelType': 'infinite',
    ':onGridReady': '''(params) => params.api.setGridOption("datasource", {
        getRows: (p) => p.successCallback([{athlete: "A", age: 1, country: "X"}], 1),
    })''',
})

ui.run()
```

Root cause: `auto_size_columns=True` (the default) injects `autoSizeStrategy: {'type': 'fitGridWidth'}`, and AG Grid treats `flex` and `autoSizeStrategy` as mutually exclusive — it logs `colDef.flex is not supported with gridOptions.autoSizeStrategy` and, under the infinite row model, never paints the cells.

This is the same family as #5276 ("render failed when `flex=1` after upgrade to 3.0"), which #5277 fixed for the **client-side** row model by switching from `sizeColumnsToFit()` to `autoSizeStrategy`. That fixed client-side + flex, but the **infinite row model + flex** case is still blank on 3.12.

I'm aware #5277 originally drafted "ignore `auto_size_columns` when a column has `flex`" and then moved away from it (the struck-through line in its description) in favour of `autoSizeStrategy`. This PR revisits a *refined* version of that idea, because `autoSizeStrategy` is fundamentally incompatible with `flex` (per AG Grid's own warning), so it can't be the auto-sizing mechanism when flex is in play.

### Implementation

Make `auto_size_columns` a tri-state `bool | None = None`:

- **`None`** (new default): fit-to-width **unless** columns use `flex` — then skip `autoSizeStrategy` and let flex size the columns.
- **`True`**: always inject `autoSizeStrategy` (explicit intent respected), with a `warn_once` when it conflicts with `flex`.
- **`False`**: never inject (unchanged).

`_uses_flex(options)` detects flex in `defaultColDef`/`columnDefs`, including the dynamic `:flex` property form (which NiceGUI converts to `flex` client-side). The `auto_size_columns` property getter is unchanged and stays truthful.

Why the tri-state rather than #5277's struck-through unconditional skip: it keeps an explicit escape hatch (`auto_size_columns=True` still forces `autoSizeStrategy`), so users who deliberately want fit-to-width with flex aren't silently overridden — only the unopinionated default becomes flex-aware.

Verified locally (AG Grid 34.2.0):
- #5276 repro (flex, **client-side**) — still renders correctly (no regression of #5277).
- #5087 repro (flex, **infinite**) — now renders (was blank).
- `ruff` + `mypy ./nicegui` + `pylint ./nicegui` (10.00/10) clean.

### Progress

- [x] Pytests have been added/updated. (Browser-free `user`-fixture test asserting `autoSizeStrategy` presence/absence across `None`/`True`/`False` × flex / no-flex / `:flex`.)
- [x] Documentation has been added/updated or is not necessary. (Docstrings updated; glad to add an aggrid flex demo if preferred.)
- [x] No breaking changes to the public API or migration steps are described above. (Default `True`→`None`; behavior unchanged for non-flex grids and for explicit `True`/`False`; flex grids change from *blank* to *working*. Type widens `bool`→`bool | None`, backward-compatible.)

### DIFF
--- a/nicegui/elements/aggrid/aggrid.py
+++ b/nicegui/elements/aggrid/aggrid.py
@@ -49,7 +49,7 @@ class AgGrid(Element, component='aggrid.js', esm={'nicegui-aggrid': 'dist'}, def
         super().__init__()
 
         col_defs = [options.get('defaultColDef', {}), *options.get('columnDefs', [])]
-        uses_flex = any(col.get('flex') is not None or ':flex' in col for col in col_defs)
+        uses_flex = any(col.get('flex') or ':flex' in col for col in col_defs)
         if auto_size_columns is None:
             auto_size_columns = not uses_flex  # auto: flex columns size themselves
         elif auto_size_columns and uses_flex:
--- a/nicegui/elements/aggrid/aggrid.py
+++ b/nicegui/elements/aggrid/aggrid.py
@@ -188,6 +188,8 @@ class AgGrid(Element, component='aggrid.js', esm={'nicegui-aggrid': 'dist'}, def
     @options.setter
     def options(self, value: dict) -> None:
         self._props['options'] = value
+        if self.auto_size_columns and self._uses_flex(value):
+            helpers.warn_once('AG Grid: "flex" is ignored when auto_size_columns=True. Grid may render blank.')
 
     @property
     def html_columns(self) -> list[int]:
--- a/nicegui/elements/aggrid/aggrid.py
+++ b/nicegui/elements/aggrid/aggrid.py
@@ -48,11 +48,9 @@ class AgGrid(Element, component='aggrid.js', esm={'nicegui-aggrid': 'dist'}, def
 
         super().__init__()
 
-        col_defs = [options.get('defaultColDef', {}), *options.get('columnDefs', [])]
-        uses_flex = any(col.get('flex') or ':flex' in col for col in col_defs)
         if auto_size_columns is None:
-            auto_size_columns = not uses_flex  # auto: flex columns size themselves
-        elif auto_size_columns and uses_flex:
+            auto_size_columns = not self._uses_flex(options)
+        elif auto_size_columns and self._uses_flex(options):
             helpers.warn_once('AG Grid: "flex" is ignored when auto_size_columns=True. Grid may render blank.')
 
         self._props['options'] = {
@@ -65,6 +63,11 @@ class AgGrid(Element, component='aggrid.js', esm={'nicegui-aggrid': 'dist'}, def
         self._props['modules'] = modules[:]
 
         self._props.add_rename('html_columns', 'html-columns')  # DEPRECATED: remove in NiceGUI 4.0
+
+    @staticmethod
+    def _uses_flex(options: dict) -> bool:
+        col_defs = [options.get('defaultColDef', {}), *options.get('columnDefs', [])]
+        return any(col.get('flex') or ':flex' in col for col in col_defs)
 
     @staticmethod
     def _migrate_deprecated_checkbox_renderer(options: dict) -> None:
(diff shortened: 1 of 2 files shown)


falkoschindler:
Thanks, @evnchn! Reminds me of how little I know AG Grid.
Anyway, here's a slightly more compact MRE:
```py
ui.aggrid({
    'defaultColDef': {'flex': 1},
    'columnDefs': [{'field': 'name'}],
    'rowModelType': 'infinite',
    ':datasource': '{ getRows: (p) => p.successCallback([{name: "A"}], 1) }',
})
```

falkoschindler (approved):
LGTM — thanks for the focused fix and clear writeup of the design lineage with #5276/#5277.

I pushed a few small follow-ups directly to the branch:

- **Simplify and parametrize** ([a05279a8](https://github.com/zauberzeug/nicegui/commit/a05279a81)) — inlined the (then) single-use `_uses_flex` helper (reverted later) and collapsed the five-grid test into a `@pytest.mark.parametrize` table.
- **`flex: 0` is "no flex"** ([138c1774](https://github.com/zauberzeug/nicegui/commit/138c17741)) — switched the check from `col.get('flex') is not None` to truthiness so an explicit `flex: 0` (AG Grid's "no flex" sentinel) still picks up `autoSizeStrategy`. New parametrized case covers it.
- **Setter parity** ([06685f77](https://github.com/zauberzeug/nicegui/commit/06685f772)) — `grid.auto_size_columns = True` now emits the same `warn_once` when the grid has flex columns. Re-extracted `_uses_flex` as a `@staticmethod` since it now has two call sites.
- **Options setter** ([e88718a1](https://github.com/zauberzeug/nicegui/commit/e88718a13)) — `grid.options = {...}` warns when the new dict carries both flex and `autoSizeStrategy: fitGridWidth`.

Considered but deliberately skipped: re-deriving `auto_size_columns` from the new options on assignment. That would silently overwrite an explicit `auto_size_columns=True` from construction, so it'd need to track the original tri-state — more plumbing than this PR warrants. Worth its own PR if anyone hits it.

A couple of edge cases also still escape detection:

- **Column groups** — `_uses_flex` walks `defaultColDef` plus top-level `columnDefs`, but nested `children` inside a column group are skipped. A flex column inside a group would silently keep `autoSizeStrategy` and re-trigger the bug.
- **Dynamic falsy flex** — `':flex' in col` matches by key presence, so `':flex': '0'` (which evaluates to a falsy `0` on the client) still counts as "uses flex". The static counterpart was fixed by switching to truthiness; the dynamic form would need expression evaluation to do the same, which feels disproportionate.

Both are corner cases and not worth holding the fix; flagging in case anyone hits them downstream.

Bigger-picture thought, not blocking: given how much machinery the parameter has accreted (tri-state default, `_uses_flex`, warnings at three entry points, the edge cases above) and that it only aliases one of AG Grid's `autoSizeStrategy` variants (`fitGridWidth`), it might be worth deprecating `auto_size_columns` in favor of `options['autoSizeStrategy']` for NiceGUI 4.0. That would let users opt into the other variants (`fitCellContents`, `fitProvidedWidth`, …), retire the flex-detection plumbing, and stop NiceGUI from trying to mediate a conflict that AG Grid itself documents. Happy to open a separate issue for that discussion so this PR stays focused on the bug fix.

Approving.

PR #6081, opened 2026-05-28
Fix: clarify crypto.randomUUID() fallback for non-secure contexts

### LINKED ISSUE #5802: NiceGUI iOS Safari Page Reload Bug (+root cause +solution!)
### First Check

### Example Code

```python

```

### Description

- Page reloads automatically every few seconds on iOS 26.3 Safari
- No user interaction required to trigger the reload
- Happens before the screen dims (auto-lock set to 5 minutes)
- Occurs with a minimal test case --- a page with a single ui.label with no other components
- Works correctly on all desktop browsers
- Works correctly using NiceGUI v2.9.0 and v2.23.3 on the same mobile device

### NiceGUI Version

3.7.1

### Python Version

3.14.3

### Browser

Safari

### Operating System

iOS

### Additional Context

I found previous issues where this same behavior was reported in the past and dismissed as a bug with Safari or something else Apple-specific. After experiencing this issue myself today, I decided to investigate further.

The test environment was a clean build of NiceGUI on Docker 29.2.1, build a5c7197. The main.py consisted of only a ui.label and nothing else. The site was accessed directly by http (no reverse proxies or other networking complications).

As mentioned above, the page would reload by itself every few seconds. I used the Safari Web Inspector (iPhone connected via USB to Mac) to determine the cause. It showed a JavaScript exception on page load.

> TypeError: crypto.randomUUID is not a function

Also, the WebSocket socket.io connection showed as a spinning/pending state in the Network tab

With some help from AI, I found out the root cause....

Factor 1: NiceGUI v3 introduced an implicit handshake mechanism
NiceGUI v3.7.0 changelog entry: "Perform implicit handshake and fix on_connect called before page is ready." After the WebSocket connects, the client generates a unique tab ID using crypto.randomUUID() and sends a handshake to the server. If the server does not acknowledge within a timeout window, the client executes:

> // nicegui.js --- handshake failure handler
>
> console.log(\'reloading because handshake failed for clientId\');
>
> window.location.reload();

This mechanism did not exist in v2.x, which is why older versions are unaffected.

Factor 2: crypto.randomUUID() is restricted to secure contexts

The Web Cryptography API method crypto.randomUUID() is only available in secure contexts --- pages served over HTTPS or from localhost. This is a web standard enforced by all modern browsers. iOS Safari enforces it strictly.

Since the app was served over plain HTTP (accessed by local IP address), crypto.randomUUID() was not available. Safari threw a TypeError, the tab ID could not be generated, the handshake could not complete, and NiceGUI reloaded the page. This repeated every few seconds indefinitely.

Desktop browsers (Chrome, Firefox) are more permissive about this restriction in some contexts, which is why the bug only appeared on iOS Safari.

**The Fix**

The fix is to serve the application over HTTPS so that crypto.randomUUID() is available as a secure context API. This was implemented using Caddy as a reverse proxy with TLS termination.

**Setup**

-   Caddy added as a Docker service alongside NiceGUI
-   tls internal directive used: Caddy generates its own local CA and signs a certificate
-   nicegui.home hostname configured in local DNS   (this domain name is specific to my local network)
-   Caddy's CA certificate (root.crt) extracted from the container and installed on the iPhone as a trusted root

**Caddyfile configuration**

> nicegui.home {
>
> tls internal
>
> reverse_proxy nicegui:8080
>
> }

**Result**

-   Safari Web Inspector showed no JavaScript errors
-   socket.io connection shown as green and stable in Network tab
-   No further page reloads
-   crypto.randomUUID() executes successfully over HTTPS

--removed unnecessary code sample

Thanks for creating and maintaining NiceGUI, I have been using it since the beginning and find it an invaluable tool. I hope this analysis is useful.

goingforstudying-ctrl:
### Motivation

Issue #5802 reported that NiceGUI pages reload continuously on iOS Safari when served over plain HTTP (e.g., accessed via local IP address). The root cause was identified as `crypto.randomUUID()` throwing `TypeError` because it is only available in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) (HTTPS or localhost).

While the code already has a working fallback that generates a UUIDv4 manually, the comment did not explain *why* the fallback is needed. This PR adds a clarifying comment so future maintainers understand the relationship between secure contexts and the fallback path.

### Implementation

Added two lines of comment in `nicegui/static/nicegui.js` before the fallback UUID generation:
- Explains that `crypto.randomUUID()` requires a secure context
- States that the fallback handles HTTP contexts (e.g., iOS Safari on local IP)

No runtime behavior is changed.

### Progress

Fixes #5802

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -368,6 +368,8 @@ function createRandomUUID() {
   try {
     return crypto.randomUUID();
   } catch (e) {
+    // crypto.randomUUID() is only available in secure contexts (HTTPS/localhost).
+    // Fall back to a manual UUIDv4 implementation for HTTP contexts.
     // https://stackoverflow.com/a/2117523/3419103
     return "10000000-1000-4000-8000-100000000000".replace(/[018]/g, (c) =>
       (+c ^ (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (+c / 4)))).toString(16),


falkoschindler:
Thanks for taking the time to open a PR, @goingforstudying-ctrl!

I'm going to close this one, though. Two reasons:

1. The comment doesn't add much beyond what the `try`/`catch` and the existing StackOverflow link already convey — the fallback's purpose is fairly self-evident from the structure, and anyone curious about the exact restriction can look up `crypto.randomUUID` or the link in a moment.
2. More importantly, this doesn't actually fix #5802. The fallback has been in place since v1.4.21 (April 2024, commit ed298d7f), so the reload loop reported against v3.7.1 must have a different root cause — most likely something in the handshake/WebSocket path on iOS Safari over HTTP, not UUID generation. Merging this with "Fixes #5802" would silently lose that open bug.

If you'd like to keep digging into the actual cause of #5802, that would be a very welcome contribution. Thanks again!

evnchn:
No need to chase for comments in existing code in general. If it's really bad you can comment beneath the PR which introduced the code, and others will reach it when they git blame. 

In this case, that mechanism alone already clears the misconception. And this closed PR would just add to that index as well. 

Sorry but for a concise codebase we can't afford to have lengthy comments everywhere, and we implicitly do expect readers to take some of the load.

Issue #6061, opened 2026-05-20
Regression: Storage.path = None at import time breaks production apps importing `nicegui.testing`


falkoschindler:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Description

@evnchn Great choice to merge PR #5960 early in 3.13...

This PR introduced a regression that breaks `app.storage.user` for any app whose import graph reaches `nicegui.testing`.

`nicegui/testing/general_fixtures.py` sets `Storage.path = None` as a module-level statement (intended as the "not configured yet" sentinel for pytest). But that line runs on plain import, not just under pytest. The website's `main.py`, for example, imports `nicegui.testing` transitively via `website/documentation/content/screen_documentation.py` (`from nicegui.testing import Screen`), so starting the docs server clobbers `Storage.path` to `None`. The next request that touches user storage crashes in `Storage._create_persistent_dict` with:

```
TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'
```

at `FilePersistentDict(Storage.path / f'storage-{id}.json', …)`.

Minimal repro:

```python
from nicegui.storage import Storage
print(Storage.path)                                      # PosixPath('.../.nicegui')
import website.documentation.content.screen_documentation
print(Storage.path)                                      # None
```

Fix is to drop the import-time `Storage.path = None` and track "already configured" via a separate module-level boolean instead of overloading `Storage.path` as the sentinel — so the attribute keeps its default value until pytest actually overrides it.

### NiceGUI Version

3.13

KrilleGH:
I just stumbled into this one trying to run the website. So no need to report separately.

evnchn:
@KrilleGH if you wanna run the website then use 3.12 tag instead of main for now. We didn't ship anything on top of 3.12 other than #5960 so not missing much. It's a major so we'd fix it the main to be working very quickly I think.

falkoschindler:
I just fixed this issue by reverting PR #5960 on main.

KrilleGH:
Thanks @evnchn, I ended up patching it with `Path('/tmp')` instead of `Storage.path` before finding this issue.

[...]

PR #6084, opened 2026-05-30
Add request check to startup smoke test

### LINKED ISSUE #6079: startup CI smoke test never issues a request — request-time regressions (e.g. #6061) ship green
*Filed via Claude Code on Evan's behalf.*

## Problem

The `startup` CI stage (`test_startup.sh`) only asserts that the server **boots** — it greps the startup output for `NiceGUI ready to go` and the absence of `Traceback|Error`, but it **never issues an HTTP request**. So any regression that is latent until a request slips straight through.

#6061 is the concrete case: #5960 set `Storage.path = None` at import time (clobbered transitively via the docs' `from nicegui.testing import ...`). The server boots fine; the crash — `TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'` in `_create_user_storage` → `Storage.path / f'storage-{id}.json'` — only fires on the first storage-touching request. With no client connecting during the startup job's window, that code path is never exercised, so the job stays green.

## Evidence

All `full-test / {3.10–3.14} / startup` jobs passed on the #5960 merge commit `5328b594`; the regression was reported by a user (#6061), not caught by CI.

<details><summary>Empirical repro + CI timings</summary>

- Run `main.py` on `0a1e6597` (reapplied #5960, pre-fix) and issue one request to any page that uses `app.storage.user` → 500 with the `NoneType / str` traceback. The traceback travels the **HTTP request path** (`uvicorn → _create_user_storage`), so a single request during the startup window *would* have tripped the existing `grep -qE "Traceback|Error"` in `test_startup.sh`. CI simply never sends one.
- Job durations on that run: `startup` ≈ 10 min, `pytest` ≈ 21 min (per Python version, n=5). The merge gate is bound by the ~21 min `pytest` job, and the `startup` job finishes ~11 min earlier.

</details>

## Proposed fix

After `NiceGUI ready to go`, have `test_startup.sh` hit the running server with at least one real request — the main website **and** each example — on a storage-touching endpoint, and fail on a 5xx (the existing log-grep then surfaces the server-side 500). Because `startup` (~10 min) finishes well before `pytest` (~21 min) gates the merge, this added request-level coverage is effectively free on wall-clock — it fills the ~11 min of headroom rather than extending the critical path.

SSDWGG:
## Summary

- run each startup target under the existing 10 second timeout in the background
- wait for the server ready log, then issue a real HTTP request to the detected local server URL
- fail the startup smoke test on request failures or HTTP 5xx responses while preserving the existing log checks

Fixes #6079.

## Testing

- `bash -n test_startup.sh`
- `git diff --check HEAD~1 HEAD`
- `PATH=/tmp/gh-pr-test-bin:$PATH bash -c "$(awk '/^check main.py/{exit} {print}' test_startup.sh); check examples/progress/main.py"`
- `PATH=/tmp/gh-pr-test-bin:$PATH bash -c "$(awk '/^check main.py/{exit} {print}' test_startup.sh); check examples/single_page_app/main.py"`

The last two local checks used a small GNU `timeout` shim because macOS does not provide the CI `timeout` command by default.


evnchn:
Thanks for your PR. Looks quite good. 

Well I already have a private attempt at https://github.com/evnchn/nicegui/pull/163

Will reconcile and see how to take the best of both worlds once have time. But I feel like your work will dominate. 

Rest assured you will be credited no matter what happens.

evnchn:
Let's also do this in early 3.14 instead

SSDWGG:
Sounds good, thanks. I'm happy for this to be considered for early 3.14.

evnchn (dismissed):

**TL;DR:** Approving — we verified both directions empirically, including a reproduction of the exact historical bug that motivated #6079. This supersedes evnchn/nicegui#163 (our own draft for the same issue) and covers everything it did, in one respect better: the URL is extracted from the server log instead of a hard-coded port list. @falkoschindler — could you do a go-around on this one soon? The thread converged on early-3.14, and test fixtures merging early means they get to find bugs for the whole cycle; these crucial CI components also deserve more eyes early.

<details>
<summary><b>Empirical verification</b></summary>

**1. Negative path — a request-time 500 that boots cleanly** (a NiceGUI app whose `/` page raises):

| script | result |
|---|---|
| `test_startup.sh` on `main` | exit 0 — **bug ships undetected** (confirms the #6079 premise) |
| this PR | exit 1 — caught twice over: `returned HTTP 500` from the new request check, *and* the server traceback now lands in the captured output |

**2. Gold standard — the historical #6061 bug** (checkout `5328b594`, the #5960 merge commit, with this PR's script):

```
http://localhost:8080/ returned HTTP 500
|   File ".../nicegui/storage.py", line 141, in _create_user_storage
| TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'
```

exit 1 — the exact regression class that shipped green through the old smoke test is now caught.

**3. Code audit:** all new failure-prone statements are `set -euo pipefail`-safe (`|| exitcode=1` guards; the `[[ ]] && var=` patterns fall under the &&-list exemption), the ready-poll (50×0.2 s) is aligned with the 10 s timeout, only ≥500 fails (404-tolerant for apps not serving `/`), the temp file is cleaned before the failure block, the check-target list and 3-attempt retry wrapper are untouched, and CI green covers the positive path on every real target including `examples/fastapi` on port 8000 (covered by the log-extracted URL).

</details>

Nice work @SSDWGG — this held up under adversarial testing on the first try. 🎉

falkoschindler (changes requested):
Thanks a lot for the contribution, @SSDWGG!
Unfortunately there are still some major issues that need to be fixed before merging:

1. **The startup check fails for the fastapi example — the PR would fail the merge queue** (blocking) — `full-test` never ran on this PR (it only runs in the merge queue), so I ran `test_startup.sh` in a CI-like Linux container (Debian, `uv sync --locked`, same invocation as the workflow). `examples/fastapi` fails deterministically on all 3 attempts with `Failed to request http://127.0.0.1:8000/: <urlopen error [Errno 111] Connection refused>`: with `reload=True`, uvicorn's reloader parent logs `Uvicorn running on http://127.0.0.1:8000` _before_ the worker process has bound the socket, so the single-shot request fires into the ~1 s spawn-and-import gap. Probing confirmed it: immediately after the ready line the request is refused; one second later it returns 200. The gap is the worker's startup time, so fast CI runners hit it too. Fix: retry refused connections, e.g. `curl --retry 5 --retry-connrefused --retry-delay 1` (with the urllib helper it would need a manual retry loop).

2. **The threaded_nicegui example fails too — it exits before the request** (blocking) — `examples/threaded_nicegui` runs the server in a daemon thread while the main thread reads stdin. With CI's closed stdin, `input()` raises `EOFError` immediately, the main thread exits cleanly, and the daemon server dies — *after* the ready line was printed. The old script passed this (clean exit, ready line present); the request check now gets connection refused from a dead process, deterministically, in both the urllib and curl variants. Retries don't help. Fix: keep stdin open for the test window, e.g. launch the background job with `< <(sleep 15)` — this emulates an interactive user who hasn't typed yet and is a no-op for all other examples (verified: the example passes with it).

3. **Request failure reason is missing from the failure dump** (major) — the helper's error message goes to the script's stderr, so it appears _above_ the `"Wrong exit code 1. Output was:"` dump while the dump itself shows a clean server log with no hint of why the check failed (see the fastapi failure above). Note for the fix: simply appending to the captured log file does not work — the server's `>` redirect keeps writing at its own file offset and overwrites the appended text (verified). Collect the request output separately and append it to `$output` before the dump.

4. **Race between the request and the 10-second timeout** (major) — the request only starts after the ready line appears, but the `timeout 10` clock has been running since launch. A heavy example that becomes ready at ~8–9 s leaves too little window: the server gets SIGTERMed mid-request, and since slow startup is systematic rather than random, the 3 retries in `check` won't save it. (The connection-refused retries from finding 1 widen this exposure slightly.) Options: raise the timeout (e.g. 15 s) while keeping the 10 s ready-poll, or kill the server right after a successful request and treat exit 143 like 124 — which would also cut the job's wall clock roughly in half, at the cost of no longer catching errors that appear late in the window.

5. **Replace the bash-Python mashup with `curl`** (major) — the 20-line Python heredoc spawned via `uv run --no-sync python` is barely readable, depends on the example's environment, adds an interpreter start inside the already-tight timeout window, and contains a dead branch (`urlopen` raises `HTTPError` for any status ≥ 400, so the in-`with` `response.status >= 500` check is unreachable). The whole helper reduces to:

   ```bash
   request() {
       local url="$1" code
       # NOTE servers with reload=True log their URL before the socket accepts connections, so retry refused connections
       code=$(curl --silent --location --retry 5 --retry-connrefused --retry-delay 1 --output /dev/null --write-out '%{http_code}' --max-time 5 "$url" || true)
       if [[ $code -lt 200 || $code -ge 500 ]]; then
           echo "Request to $url failed with HTTP status $code"
           return 1
       fi
   }
   ```

   Semantics match the Python version: transport errors (`000`) and 5xx fail, 4xx is tolerated, redirects are followed (`--location`, matching urllib). `curl` is preinstalled on GitHub runners and macOS, so the local-testing story is no worse.

   Verified locally end-to-end: a healthy NiceGUI app passes (URL detection picks the ready-line port, not the fallback), and a test app raising on its first `GET /` is caught with the traceback and request failure line in the dump — the regression class of #6061.

evnchn:
Sorry for this one. I think focused on "Is this superset of evnchn#163" for the winning condition for this PR in the review. Thanks for your look @falkoschindler 🙇

SSDWGG:
Thanks for the detailed review. I pushed an update that addresses the requested startup-script changes:

- replaced the Python request helper with `curl`
- added retry handling for reload servers that log before accepting connections
- kept stdin open during the test window for examples that read from `input()`
- captured request failure output and appended it to the final failure dump
- raised the timeout to 15 seconds while keeping the existing readiness polling window

I could run `bash -n test_startup.sh` locally. Full local example verification was blocked by local environment issues (`timeout` is not available on macOS, and `uv sync --locked` timed out fetching `poetry-dynamic-versioning` from PyPI), so I’m relying on the Ubuntu CI run for the end-to-end startup matrix.

falkoschindler (changes requested):
Sorry, wrong thread...

falkoschindler (approved):
Thanks for tightening this up! The smoke test now exercises the request path instead of only the boot log, which is exactly the gap that let #6061 slip through.

I ran the full suite in a CI-like Linux container against the latest state and all 54 checks pass with no retries — including `examples/fastapi` (the reload-race that the connection-refused retries fix) and `examples/threaded_nicegui` (which needed the open stdin so it doesn't exit before the request).

I pushed one small follow-up in 2d1b116 to align the ready-poll window with the run timeout: the poll loop was capped at 10 s while the process timeout is 15 s, so a server taking 10–15 s to log its ready line would have been marked failed even though it was still starting. It now derives the poll window from a single `timeout_seconds` constant.

LGTM — approving. 🎉

PR #6086, opened 2026-06-01
Fix selenium screenshot artifacts not being uploaded on test failure

HosseyNJF:
### Motivation

I observed that after a screen test failure, screenshots are not being uploaded to GitHub as artifact. This is an attempt at fixing that.

### Progress

### DIFF
--- a/tests/test_radio.py
+++ b/tests/test_radio.py
@@ -10,7 +10,7 @@ def test_radio_click(screen: Screen):
 
     screen.open('/')
     screen.click('A')
-    screen.should_contain('Value: Meow')
+    screen.should_contain('Value: A')
     screen.click('B')
     screen.should_contain('Value: B')
 
--- a/tests/test_radio.py
+++ b/tests/test_radio.py
@@ -10,7 +10,7 @@ def test_radio_click(screen: Screen):
 
     screen.open('/')
     screen.click('A')
-    screen.should_contain('Value: A')
+    screen.should_contain('Value: Meow')
     screen.click('B')
     screen.should_contain('Value: B')
 
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -39,7 +39,7 @@ jobs:
         if: matrix.scope == 'pytest' && failure()
         with:
           name: screenshots-${{ matrix.python }}
-          path: screenshots/*.failed.png
+          path: nicegui/screenshots/*.failed.png
           retention-days: 14
       - name: ${{ matrix.python }} / ${{ matrix.scope }}
         if: matrix.scope == 'startup'
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -39,7 +39,7 @@ jobs:
         if: matrix.scope == 'pytest' && failure()
         with:
           name: screenshots-${{ matrix.python }}
-          path: nicegui/screenshots/*.failed.png
+          path: nicegui/screenshots/**/*.failed.png
           retention-days: 14
       - name: ${{ matrix.python }} / ${{ matrix.scope }}
         if: matrix.scope == 'startup'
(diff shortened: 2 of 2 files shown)


HosseyNJF:
@evnchn I made sure to create a [PR](https://github.com/HosseyNJF/nicegui/pull/2) on my own fork _before_ creating this one, to see if it has any effect on the condition "has an open PR". It doesn't seem to have any effect.

falkoschindler:
Great! Workflow run https://github.com/zauberzeug/nicegui/actions/runs/26830100927 contains an "Artifacts" section at the bottom.
So I guess we can revert the test failure.

tests/test_radio.py
```diff
@@ -10,7 +10,7 @@ def page():
 
     screen.open('/')
     screen.click('A')
-    screen.should_contain('Value: A')
+    screen.should_contain('Value: Meow')
```
falkoschindler:
```suggestion
    screen.should_contain('Value: A')
```

HosseyNJF:
On the last run, the artifact was uploaded correctly, and the failed test has been reverted. The PR is ready for review/merge 🤝

evnchn (approved):
Well yeah, and sorry for the regression, the issue is caused by https://github.com/zauberzeug/nicegui/pull/6075 which is supposed to carry-forward https://github.com/zauberzeug/nicegui/pull/5960

I guess third times the charm then!

evnchn:
Just to be sure I asked my Claude Code

<details>
<summary>Verdict: Yes — 100% syntactically correct, and it's the *right* fix, not just an empirically-lucky one.</summary>

### What the PR does
`.github/workflows/_test.yml`: `path: screenshots/*.failed.png` → `screenshots/**/*.failed.png` (the `path:` input to `actions/upload-artifact`).

### Why it's correct (evidence, not vibes)

**1. The actual write location is one level deep.** `nicegui/testing/screen_plugin.py:29` overrides the dir at `pytest_configure` time:
```python
Screen.SCREENSHOT_DIR = (Path('screenshots') / str(os.getpid())).resolve()
```
and `screen.py:285-289` writes `SCREENSHOT_DIR / f'{name}.failed.png'`. So failed shots always land at **`screenshots/<pid>/<name>.failed.png`** — never directly under `screenshots/`. The PID subdir is the parallel-run isolation that the old glob predated; that's the root cause of the upload silently capturing nothing.

**2. Glob semantics verified empirically** against `minimatch` (the matcher `@actions/glob` uses, `dot:true`):

| pattern                             | against                            | result                                  |
| ----------------------------------- | ---------------------------------- | --------------------------------------- |
| `screenshots/*.failed.png` (old)    | `screenshots/12345/img.failed.png` | **false** ← the bug                     |
| `screenshots/**/*.failed.png` (new) | `screenshots/12345/img.failed.png` | true ✓                                  |
| `screenshots/**/*.failed.png`       | `screenshots/12345/a/b.failed.png` | true ✓ (any depth)                      |
| `screenshots/**/*.failed.png`       | `screenshots/img.failed.png`       | true ✓ (`**` matches **zero** segments) |

The new pattern is a **strict superset** of the old one: `**` matching zero path segments means it still covers any hypothetical top-level file, while also covering the real `<pid>/` location and arbitrary deeper nesting. There's no edge case where the old pattern matched and the new one doesn't.

### Caveat (non-syntactic, worth a one-liner if you comment)
`upload-artifact` computes the least-common-ancestor of matched files for the in-zip layout, so screenshots will now appear nested under `<pid>/` inside the artifact rather than flat. Harmless, but if anyone expected flat filenames in the downloaded zip, that's the only behavioral change. Doesn't affect correctness of the upload itself.

**Bottom line:** syntax is valid GitHub-Actions/minimatch glob, semantics match the real write path, and it's a superset of the prior pattern — so "empirically works" is backed by the spec, not coincidence. Ship-worthy as-is.
</details>

falkoschindler (approved):

PR #6088, opened 2026-06-01
Add distributed events via Zenoh (re-implementation of #5231)

evnchn:
### Motivation

PR #5231 prototyped distributed events using Zenoh's broker-less pub/sub. @rodja explicitly handed the work off in https://github.com/zauberzeug/nicegui/pull/5231#issuecomment-4585981242, so I picked it up: re-implemented the feature against current `main`, battle-tested it on a multi-node homelab (real LAN, not just multiple swarm replicas on one Docker daemon), and ran the surface through a multi-reviewer pass to catch issues that source-read alone misses.

This is a fresh implementation, not a rebase on top of #5231 - the API surface and topic-derivation diverged enough during fixing that range-diffing on #5231 didn't help. Happy to close #5231 in favour of this if you agree.

### Implementation

`DistributedEvent` is a subclass of `Event` that auto-publishes every emission via Zenoh and subscribes to receive sibling emissions. Enabled per-app via `ui.run(distributed=...)`:

| Form | Meaning |
|---|---|
| `True` | Zenoh defaults (UDP-multicast peer scout) |
| `["host", "host:7447"]` | Explicit unicast peers - the 99% case when you know your peer list |
| `{"...raw Zenoh JSON..."}` | Raw Zenoh config dict - full control escape hatch |

`storage_secret` is **required** for `distributed=...` (without it, unrelated deployments would silently cross-talk on the same multicast scope - rejected loudly at session construction). It drives two independent derived values: the **topic namespace** (collision-avoidance, not a confidentiality boundary) and, since `98d2962`, **payload encryption on by default** (Fernet key via HKDF; wrong-secret nodes see only ciphertext and can't read or forge events; adds `cryptography>=43.0.0` to the `distributed` extra).

Demo at `examples/distributed/`: Docker Swarm + Traefik v3.7 with 3 NiceGUI replicas behind a sticky-cookie LB.

<details><summary><b>Topic derivation & payload encryption - mechanism and threat model</b></summary>

- **Topics** are derived from `module:lineno` of the construction site (no absolute filepath, so two physical hosts running the same code from different working dirs stay in sync), then namespaced via `HMAC-SHA256(storage_secret, "nicegui-distributed-v1")[:16]`. Collision-avoidance only - the derived prefix is visible in Zenoh discovery traffic.
- **Payloads** are sealed with a Fernet key (AES-128-CBC + HMAC-SHA256) derived from `storage_secret` via `HKDFExpand(SHA256, info=b"nicegui:event-bus:v1")` before they hit the wire, and opened on receipt. The `info` string domain-separates the encryption key from the namespace HMAC. The secret never crosses the wire. Undecryptable or stale messages are dropped silently, so "different secrets ⇒ isolated" is enforced cryptographically, not just by topic prefix.
- **Residuals** - metadata leakage (topic/timing/size), sender identity, DoS, replay within the TTL window - are documented in the Security section of `examples/distributed/README.md`, with mTLS node auth + `access_control` ACLs as the defense-in-depth story (see the `DistributedSession` trust-model docstring).
- `cryptography` ships prebuilt wheels on all supported platforms (no Rust toolchain needed) and only lands for users who opt into `pip install "nicegui[distributed]"`.
- The Traefik pin is v3.7 because the v2.10 image used in #5231 doesn't work on Docker daemons ≥29 (API-version drop).
</details>

### Verification

CI green on head `98d2962` (pre-commit / mypy / pylint / quick-test); `pytest tests/test_distributed_events.py` 13/13; verified empirically cross-host on a real LAN and end-to-end via a two-tab Playwright run against the swarm demo.

<details><summary><b>Full verification evidence</b></summary>

Local (this branch):
- `pre-commit run --files <changed>` clean
- `mypy ./nicegui` clean across 241 files
- `pylint ./nicegui` 10.00/10
- `pytest tests/test_distributed_events.py` 13/13 (12 original + `test_payload_from_different_secret_is_dropped` added with the encryption commit)

Homelab (real cross-host, not just same-daemon replicas):
- Cross-node LAN test (.180 publisher, .134 subscriber via the new `peers=` list shortcut, sibling secret): 5/5 events received.
- Same setup with mismatched secrets: 0/5. At the time of that run this demonstrated namespace separation only; since `98d2962`, wrong-secret payloads are additionally rejected at decryption.
- Two-tab Playwright run against the swarm demo: tab A on instance `5a2d8798985d`, tab B on instance `62d586a21a1e` (sticky-cookie cleared on tab B to force a different replica). Click an emoji on tab A → MutationObserver on tab B captures the float-up element seconds later with the right emoji. End-to-end Zenoh → asyncio loop thread marshalling → slot/client context → DOM update all working.

The fork PR https://github.com/evnchn/nicegui/pull/164 has the full evidence chain: multi-reviewer pass (a code-reviewer subagent + GitHub Copilot), the empirical-verification probe used to prove all five reviewer findings reproduced, the followup commits that fixed them, and the norm-conformance pass.
</details>

<details><summary><b>Known followups (Important-tier, not blockers)</b></summary>

These came out of the same reviewer pass and are catalogued but not addressed in this commit set:

- `zenoh.open()` failure currently bubbles out of `ui.run` - distributed should be a degradable overlay, not a startup killer. Want: try/except around `DistributedSession.initialize`, log + continue.
- No explicit `topic=` override on `DistributedEvent.__init__` for users whose construction site is inside a factory function/decorator/closure.
- The example container does `pip3 install eclipse-zenoh tinycss2 && python3 main.py` at start - works but slow. Want: rebuild `zauberzeug/nicegui:<next>` with the `distributed` extra baked in, then drop the `pip3` line (the existing TODO already says this).
- Multi-worker uvicorn behaviour with one `DistributedSession` per worker should be documented in the docstring (currently silent).

I'm happy to address any/all of these in this PR or as follow-ups, whichever fits your review style.
</details>

### Progress

---

cc @rodja @falkoschindler - this is the deliverable from the #5231 handoff. No rush - review whenever fits.

*Changelog: 2026-07-02 - synced Implementation/Verification with commit `98d2962` (Fernet payload encryption by default, `cryptography` in the `distributed` extra, 13th test) and folded evidence into `<details>` blocks; see the announcement comment of the same date.*


jeffective:
> Two instances sharing a secret share the namespace; instances with different secrets stay isolated even when the Zenoh transport reaches them. The wire never carries the secret. storage_secret is required for distributed=... (without it, unrelated deployments would silently cross-talk on the same multicast scope - rejected loudly at session construction). The example app passes storage_secret='distributed-events-demo'.

Zenoh key expressions (including namespaces) are not a security measure. Any node on the Zenoh network may construct an empty namespace and subscribe to `**` to see everything. You may reach for verbatim key expressions (expressions including `@`) but this is also not a security measure, as nodes communicate key expressions between each other as part of interest messages.

If you desire secure private communication in a public Zenoh network, this is difficult and requires message level security, which is not implemented. An example work in progress implementation is here: https://codeberg.org/permian/zenoh-mls

I would suggest using transport level security using authenticated nodes using TLS authentication. https://zenoh.io/docs/manual/tls/

evnchn:

You're right on all counts, @jeffective — and rather than just acknowledge it, I've corrected the framing in the code so the docs and the implementation can't drift apart. I checked each of your points against Zenoh's own docs first.

The HMAC-derived namespace is **collision-avoidance, not a security boundary**, exactly as you say. And the leak is a touch sharper than "subscribe to `**`": the secret never crosses the wire, but the *derived 16-char prefix* does — in the subscriber/queryable **declarations** (and the interests you mentioned), which routers forward network-wide ([declarations spec](https://spec.zenoh.io/spec/1.0.0/session/declarations.html): `key_scope`/`key_suffix` are on the wire, "Declarations MUST be forwarded by routers to all sessions that have expressed interest"). So a passive node learns the namespace from discovery traffic even before anything is published.

What I changed (commit on this branch):

- Reworded the `__init__` / `_derive_namespace` docstrings and the topic-derivation comment from "stay isolated" to collision-avoidance, and renamed `test_namespace_is_deterministic_and_isolating` → `..._and_collision_avoiding`.
- Added a **trust-model** note to the `DistributedSession` docstring and a **Security** section to the example README: distributed mode assumes a trusted Zenoh network; for real isolation, pass a raw Zenoh config with **mTLS** node auth ([tls](https://zenoh.io/docs/manual/tls/)) + an **`access_control`** ACL to gate pub/sub by key-expression and authenticated subject ([access-control](https://zenoh.io/docs/manual/access-control/)); payload E2EE would need message-level security like the experimental [zenoh-mls](https://codeberg.org/permian/zenoh-mls) you linked.
- Updated the PR description to match.

The mechanism itself stays (the prefix is still worth it for the accidental-cross-talk case). Does this framing land for you? Thanks again — this made the feature honest about what it does and doesn't do.

evnchn:

**TL;DR — a secure-by-default option worth weighing before this merges.**

Use the `storage_secret` you already require (and that never crosses the wire) to **encrypt event payloads**, not just to namespace them: `HKDF(storage_secret)` → a **Fernet** key, encrypt before `put()`, decrypt on receive. A passive node on the Zenoh network then sees the namespace + ciphertext but can't read events — @jeffective's "you need message-level security," delivered at the NiceGUI layer, on by default, no config.

The one trade-off to weigh:

- It adds **exactly one dependency** — `cryptography` (Fernet *and* HKDF live in it). Prebuilt wheels on every supported platform, **no Rust to install**. `eclipse-zenoh` itself ships zero Python deps, so this isn't already in the tree.
- It protects payload **contents only** — topic/timing/size metadata and sender identity still leak (mTLS/ACL stay the answer for those).

The ask: **@falkoschindler / @rodja — is bundling `cryptography` into the `distributed` extra acceptable?** If yes, I'll implement Fernet-by-default (~2 seams around the existing payload marshalling + tests). If you'd rather keep NiceGUI dep-light, the documented mTLS/ACL story stands. Enumerated options at the bottom.

<details><summary><b>Why Fernet (vs AES-GCM / PyNaCl)</b></summary>

All three give symmetric authenticated encryption; Fernet wins for *this* case (small JSON payloads, shared key) on two points, both verified from the docs:

- **Built-in timestamp + `ttl=` staleness check.** "The encrypted message contains the current time when it was generated in _plaintext_"; on decrypt, "If the message is older than `ttl` seconds … an exception will be raised." ([fernet docs](https://cryptography.io/en/latest/fernet/)) — partial replay/staleness mitigation for free.
- **No nonce footgun.** Fernet manages its own IV; AES-GCM and PyNaCl SecretBox make nonce-uniqueness the caller's job, and reuse is catastrophic ("Reuse of a `nonce` with a given `key` compromises the security…", [AEAD docs](https://cryptography.io/en/latest/hazmat/primitives/aead/)).
- Construction: AES-128-CBC + HMAC-SHA256 (authenticated); key = url-safe-base64 32 bytes. For small payloads the CBC+HMAC overhead is irrelevant; AES-GCM would only win on throughput, which isn't the constraint.
- HKDF lives in the same `cryptography` package → **one** dep, not two.
</details>

<details><summary><b>The dependency cost, precisely</b></summary>

- Verified locally in the `distributed`-extra venv: `eclipse-zenoh` 1.9.0 declares **zero** Python deps (Rust/pyo3 wheel; its TLS is Rust-side rustls), and `cryptography` is **not** otherwise in the tree. So payload AEAD is a genuinely new dep — not hand-waved as free.
- Install friction is low though: "A Rust toolchain is not required to **use** `cryptography`"; it "ships statically-linked wheels for macOS, Windows, and Linux (via `manylinux` and `musllinux`)" ([installation](https://cryptography.io/en/latest/installation/)). Wheels cover manylinux x86_64 + aarch64, macOS x86_64 + arm64, win_amd64.
- It only lands for users who opt into `pip install "nicegui[distributed]"`, not everyone.
- *(Untested caveat: very new CPython / niche musllinux-aarch64 could still source-build; not a mainstream concern.)*
</details>

<details><summary><b>Design sketch (~2 seams)</b></summary>

- Key: `HKDFExpand(SHA256, length=32, info=b"nicegui:event-bus:v1")` over `storage_secret`, then `urlsafe_b64encode` → Fernet key. The `info` domain-separates this from the existing namespace HMAC (same secret, different derived key).
- Publish (`distributed.py` ~L137): `token = Fernet(key).encrypt(json.dumps(...).encode())` → `publisher.put(token)`.
- Receive (~L162): `plaintext = Fernet(key).decrypt(bytes(sample.payload), ttl=…)` → `json.loads`; drop on `InvalidToken`.
- The HMAC **namespace stays** — still useful for routing/collision-avoidance; encryption is additive.
- Side effect: "different secrets ⇒ isolated" becomes *actually true* (wrong-secret nodes fail `decrypt`), which is what the original wording wanted.
</details>

<details><summary><b>What it does and doesn't protect (threat model)</b></summary>

Protects payload **confidentiality + integrity** to holders of `storage_secret`. Does **not**:

- **Hide metadata** — "the message metadata such as the MQTT Topic stays intact, the payload … gets encrypted" ([HiveMQ](https://www.hivemq.com/blog/mqtt-security-fundamentals-payload-encryption/)); here that's the Zenoh key-expression/namespace, timing, size, transport-level sender.
- **Identify senders** — symmetric, so auth = "holds the key"; any key-holder can forge a message indistinguishable from another. No per-sender identity.
- **Stop replay** — Fernet `ttl` bounds *staleness* only; a captured token replays within the window. Exact-replay defense needs an explicit nonce + dedup (v2 follow-up).
</details>

<details><summary><b>Prior art (textbook pattern, not novel)</b></summary>

- The Fernet docs' own canonical recipe derives a key from a secret via KDF + `urlsafe_b64encode` ([fernet](https://cryptography.io/en/latest/fernet/)) — exactly our shape (swap the password-KDF for HKDF-from-`storage_secret`).
- Django protects values with the shared `SECRET_KEY` ("the `Signer` class uses the `SECRET_KEY` setting", [signing](https://docs.djangoproject.com/en/stable/topics/signing/)) — the "one shared app secret → derived crypto" precedent NiceGUI already mirrors with `storage_secret`.
- App-layer payload encryption over a topic-routed bus is standard for MQTT (HiveMQ above), including published Fernet-over-MQTT implementations.
</details>

**Options:**

1. **Approve the dep** → I implement Fernet-by-default for `distributed` (recommended — it's what "secure by default" requires).
2. **Keep NiceGUI dep-light** → drop it; the mTLS/ACL documentation stands as the only confidentiality story.
3. **Opt-in instead of default** → ship it gated behind a flag, so the dep/behaviour only activates on request. Weakest on "secure by default," but no forced dep.

evnchn:
I still haven't wrapped my head fully around Zenoh, but I think the end result is that it should best stand up to adversarial attacks from the internet, considering our servers hosting https://nicegui.io after all. 

Otherwise, those servers need to be hidden from public IP which is not something that can be said for every deploy situation. 

If it can't stand up to it, fine, but we will document it because switching distributed events on means carrying some deployment assumption forward, which is load bearing. 

I'll have to figure out how to assess this since my homelab is assumed to be secure intranet, hence the misjudgment in the beginning

jeffective:
@evnchn , looks like you've got a good understanding of the tradeoffs here and I don't see any more possible missunderstandings. Nothing more for me to add 👍

evnchn:

**TL;DR: the `cryptography` question above was overtaken by events — option 1 (Fernet-by-default) is already implemented on this branch, and this comment is the announcement that should have accompanied the push.**

- Commit 98d2962 (June 16) encrypts every event payload with a Fernet key derived from `storage_secret` via HKDF, and adds `cryptography>=43.0.0` to the `distributed` extra.
- **The dependency decision is still yours to make:** if you'd rather keep NiceGUI dep-light, reverting 98d2962 cleanly restores the mTLS/ACL-documentation-only story (option 2 from the comment above). Nothing else on the branch depends on it.
- The PR body has been synced to match the code (changelog at the bottom of the body).

<details><summary><b>What commit 98d2962 contains</b></summary>

- `nicegui/distributed.py` (+43/−8): `HKDFExpand(SHA256, info=b"nicegui:event-bus:v1")` over `storage_secret` → Fernet key; `encrypt()` before `put()`, `decrypt(ttl=EVENT_TTL_SECONDS)` on receive; undecryptable payloads (foreign deployment, forged, or stale) are dropped silently. The `info` string domain-separates this key from the existing namespace HMAC — same secret, two independent derived values.
- `pyproject.toml`: `distributed = ["eclipse-zenoh>=1.0.0", "cryptography>=43.0.0"]` — the dep only lands for users who opt into `pip install "nicegui[distributed]"`.
- `tests/test_distributed_events.py` (+36): new `test_payload_from_different_secret_is_dropped` — a payload sealed under a different secret reaches the wire handler and is dropped without firing the local callback.
- `examples/distributed/README.md`: Security section rewritten — `storage_secret` is now the payload confidentiality + integrity boundary; residuals (topic/timing/size metadata, sender identity, DoS, replay within the TTL window) stay documented, with mTLS + ACL as defense-in-depth.
- The HMAC topic namespace is unchanged and remains collision-avoidance only.
</details>

<details><summary><b>Verification state on this head</b></summary>

- CI on 98d2962 is green: pre-commit, mypy, pylint, quick-test (3.10) all pass.
- The test file now carries 13 tests (was 12 when the PR body's "12/12 in ~61s" was written).
- Side effect worth naming: the earlier homelab result "mismatched secrets → 0/5 received" was namespace separation only; after this commit, a wrong-secret node additionally fails decryption — "different secrets ⇒ isolated" is now enforced by cryptography, not just by topic prefix.
</details>

<details><summary><b>Loop closure on the earlier internet-exposure question</b></summary>

My June 16 comment said I'd figure out how to assess adversarial exposure and document the load-bearing deployment assumption. This commit is that answer, at the payload layer: a node on a hostile network sees namespace + ciphertext but cannot read or forge events. What encryption does *not* cover (metadata, DoS, replay-within-TTL) is spelled out in the example README rather than left as an implicit deployment assumption.
</details>

@falkoschindler — since you've picked this up for 3.15, the concrete ask: **keep 98d2962 in (accepting the `cryptography` dep in the `distributed` extra), or say the word and I'll revert it** before your review pass. Everything else in the PR stands either way.

evnchn:
Problem: `DistributedEvent` doesn't work on the demo site. It makes the case weak, and we cannot catch regressions that way. 3.16 defer preferred.

falkoschindler:
@falkoschindler asked me to bring the open PRs up to date with `main` so the diffs are easier to read. This one is a special case: it can't be updated at all.

The branch shares no common ancestor with `main` — `git merge-tree main pr/6088` fails with `refusing to merge unrelated histories`, and GitHub's "Update branch" can't help either. So the diff GitHub shows here isn't a diff against a common base, and it won't become one by merging.

Recreating the branch from current `main` and replaying the changes onto it is the way out. Handing it back to you since only you know which commits are the ones worth replaying.

---

evnchn:
**Merged cleanly with `git merge origin/main` — no rebuild or replay needed.**

The histories aren't unrelated: same root commit, and `git merge-base` resolves fine. The branch was just 91 commits behind with the six distributed commits on top. Noting it here so the next agent reading this thread doesn't start from the rebuild plan.

<details>
<summary><b>Evidence, and what the merge was checked against</b></summary>

```console
$ git merge-base origin/main origin/distributed-cleaned
67f010cd2cc332c2cb2b56f87ada8c45c8c8d08b   # "Fix blank ui.aggrid when flex columns use the infinite row model"

$ git rev-list --max-parents=0 origin/main
64816a4c3ed96d0812fb16b4227e7c594859dcac
$ git rev-list --max-parents=0 origin/distributed-cleaned
64816a4c3ed96d0812fb16b4227e7c594859dcac   # same root

$ git merge-tree --write-tree origin/distributed-cleaned origin/main ; echo "exit=$?"
1584d526708fa9269a98900258c2522a0abfed11
Auto-merging nicegui/ui_run.py
Auto-merging nicegui/ui_run_with.py
Auto-merging pyproject.toml
Auto-merging uv.lock
exit=0
```

`merge-tree` needs two resolvable refs; given a name that doesn't exist locally it will report unrelated histories regardless of the branch, which is the most likely explanation for the earlier reading.

After the merge:

| check | result |
|---|---|
| `mypy ./nicegui` | clean, 247 source files |
| `pytest tests/test_distributed_events.py` | 13 passed, **0 skipped** |

`uv.lock` was one of the auto-merged files, so installing successfully wouldn't have been evidence on its own — I booted the real entrypoint against the locked set and requested a page: server came up and served it (`3.15.0.post17.dev0+51c1f6af`).

The branch was bundled beforehand, in case a rebuild had turned out to be necessary after all.

</details>

<sub><i>Posted by Claude Code on evnchn's behalf.</i></sub>

evnchn:

**Tests reworked to observe the network instead of the internals**, per the testing policy in `#6219` — private-attribute assertions, valid wire frames hand-built with the private helpers, and the monkeypatched shutdown hook are gone; a real sibling Zenoh node and a real second session stand in for them.

<details>
<summary><b>What changed, what was kept, and how it was checked</b></summary>

The suite reached into the implementation in several places: asserting on private attributes (`_zenoh_setup_done`), calling private methods directly (`DistributedSession._derive_namespace`, `._derive_fernet`), introspecting `session.publishers` / `session.subscribers`, hand-building encrypted wire frames, and monkeypatching `core.app.on_shutdown` to observe that a registration happened.

Those now watch what the feature actually produces — the network. A sibling Zenoh node on loopback records the key expressions NiceGUI publishes on, and a second real `DistributedSession` plays another instance, so the delivery paths now run end-to-end through real publish → encrypt → route → decrypt → dispatch. (The forged-payload rejection test is the one exception by design: real publisher and real receiver, but a deliberately synthetic attacker payload — that's the point of it.)

- namespace derivation → *an instance with a different `storage_secret` never reaches our subscribers*
- shutdown hook → *after `app.stop()`, remote events stop arriving* (strictly stronger — it also catches a broken `shutdown()`, which the monkeypatch could not)
- confidentiality → *an attacker publishes on the topic it observed on the wire, with its own key*

13 tests → 10, `+142 −105`.

**Two spots deliberately kept.** The `fresh_session` fixture still resets the `DistributedSession` singleton — there is no public reset for a process-global, and it's teardown, not an assertion. And `_peer_to_endpoint` is still unit-tested directly: it's a pure function covering the documented `distributed=['host', 'host:port']` shorthand, and the observable alternative would need to bind Zenoh's default port 7447 in CI, a collision risk this file otherwise avoids on purpose.

Each rewritten test was checked to fail when the behavior is broken:

| Mutation | Test | Failure |
|---|---|---|
| `on_shutdown(...)` registration removed | shutdown | `assert ['before', 'after'] == ['before']` |
| topic uses `co_filename` (absolute path) | portable topic | `assert False` on the `startswith` check |
| topic drops the line number | portable topic | `assert 1 == 2` |
| plain `Event.emit` publishes to the session | plain-events-stay-local | `assert ['nicegui/eve…/leaked'] == []` |
| `call_soon_threadsafe` bypassed | loop-thread | `assert <Thread(pyo3-closure)> is <_MainThread>` |
| dedupe drops every remote payload | loop-thread | `assert [] == ['hello']` |
| `storage_secret` ignored in both derivations | different-secret | `assert ['leak', 'hello'] == ['hello']` |
| `InvalidToken` payloads passed through raw | no-secret | `assert ['leak', 'hello'] == ['hello']` |
| encryption removed entirely (plaintext wire) | no-secret | `assert ['leak', 'hello'] == ['hello']` |

Honest caveat: mutating **only** `_derive_namespace` leaves the different-secret test passing, because the crypto boundary still blocks the payload. That's the expected property of an outcome-level test over a defence-in-depth mechanism — it goes red as soon as both layers do.

Stress-testing surfaced a real race, not a hypothetical one: the observer declares its Zenoh subscriber and the app can publish before that declaration propagates — **1 failure in 25 runs**, first emit lost. Fixed by retrying the emit until it lands, after which 30/30 full-file runs were clean, plus 83 tests across `test_distributed_events`, `test_event`, `test_background_tasks`, `test_storage`, `test_observables` and `test_lifecycle` together, to confirm `core.app.stop()` doesn't leak across tests.

Tests really run rather than skip — `uv sync --extra distributed` pulls `eclipse-zenoh`, `-rs` shows no skip section, and zenoh is in the `dev` group so CI's `uv sync --locked` gets it too. `uv.lock` untouched. pre-commit, `mypy ./nicegui`, `pylint ./nicegui` 10.00/10 all green.
</details>

evnchn:
**Three bugs fixed, and the 3.16 defer has expired — this needs a decision rather than another wait.**

- **The branch did not boot.** Every example needs a `screenshot.webp`, and `examples/distributed/`
  had none, so `import website.examples` raised and `main.py` never started. `test_startup.sh` would
  have caught it, but `full-test` shows `skipping` on every run of this PR.
- **The example had distributed events switched off.** Its container installed `eclipse-zenoh` but not
  `cryptography`, and the `except ImportError` reported that as *"zenoh is not installed"* — so anyone
  who tried the demo was told to install the package that was already there. Both fixed; the warning
  now names the module that actually failed.
- **The docs demo was misleading.** With `distributed` off, `DistributedEvent` degrades to `Event`, so
  the demo's "shared" and "local" halves were the same widget twice. It now says so on the page, and
  the snippet ends with `ui.run(distributed=True, storage_secret=...)` instead of a bare `ui.run()`.

Also: a failed `zenoh.open()` no longer takes the app down.

<details>

`pytest tests/test_distributed_events.py -rs` 13 passed, 0 skipped · `mypy ./nicegui` clean ·
`pylint ./nicegui` 10.00/10 · pre-commit clean · `import website.examples` resolves 60 examples.
Full suite: 5 failures, all reproducing on c1459a0e (this box's Chrome reports a dark colour-scheme
preference; `test_lazy_imports` spawns hardcoded `python3` instead of `sys.executable`).

Cross-instance delivery re-checked on a real 3-replica swarm behind Traefik: clicked on container
`32de797b10d3`, arrived on `1fac8236d7d8`, replica identity read off each page's own hostname label.

Multi-worker behaviour measured before documenting it: `uvicorn --workers 3` via `ui.run_with()`, every
worker sees every event exactly once. With a fixed `listen` endpoint only the first worker binds and the
rest now degrade instead of crashing.

Each new test was seen to fail with its fix reverted.
</details>

Left out on purpose: an explicit `topic=` for events built inside factories or closures (an API
question, not a fix), and a `ui.timer()` inside a `DistributedEvent.subscribe()` callback failing to
register its frontend component — a NiceGUI bug the demo only reaches now that it works. Happy to file
that one separately.

**Still your call, @falkoschindler / @rodja:** whether `cryptography>=43.0.0` belongs in the
`distributed` extra for payload encryption by default (98d2962). Reverting that commit restores the
mTLS/ACL-only story if you would rather stay dep-light.

<sub><i>Posted by Claude Code on evnchn's behalf.</i></sub>

evnchn:

**Six more bugs, three of them serious, all fixed in 0f9101e8.** I ran a full review of the feature after my last comment, and it was right to be suspicious of it:

1. `subscribe()` wired only the **first** `DistributedEvent` created per source line — later ones published but never received. Every non-module-level pattern hit this: an event built in a page function, a factory, a class.
2. Distributed init ran in processes that serve nothing — `run.cpu_bound` spawn workers and the reload parent — each opening its own session and running handlers again.
3. `ui.run(distributed=["host1", "host2:7447"])` **never connected.** It set connect endpoints on 7447 but no `listen` endpoint, and zenoh peers listen on an ephemeral port.

Correcting myself: finding 3 means my earlier cross-host LAN evidence proved less than I claimed — multicast scouting connected those nodes, so the peer list was doing nothing.

<details>
<summary><b>The three serious findings, and what the fixes are</b></summary>

**1 — handler fan-out.** `if wire not in self.subscribers: declare_subscriber(...)` meant a second event on the same wire topic silently got no handler. Worse: if the first event was page-scoped, `client.on_delete` unsubscribed its callbacks on disconnect and remote events then reached nobody, while every later client was deaf. The session now keeps a list of `weakref.WeakMethod` per topic and fans out to all of them — which also fixes the leak below.

**2 — wrong process.** The `if distributed is not None:` block sat above the `MainProcess` guard in `ui_run.py`. In those extra processes `core.loop` is `None`, so handlers ran inline on the Rust worker thread, and `on_shutdown` never fired, so the sessions were never closed.

Moving the block below the guard does **not** work, and measuring that is what produced the actual fix: with `reload=True` (the default) the serving process is `SpawnProcess-1`, not `MainProcess`, so below-the-guard logs `Distributed events enabled` in the reloader parent while the serving child has no session at all. Fixed with `core.app.on_startup(...)`, which lands in whichever process actually serves and never in a `cpu_bound` worker.

`ui_run_with.py` had the same exposure and got the same fix. A `MainProcess` guard there would have been actively wrong — uvicorn workers *are* `multiprocessing.Process`, so it would have disabled distributed in every worker.

**3 — the peer shortcut.** Reproduced with multicast and gossip scouting disabled: two sessions built from the documented list form exchanged nothing; adding a listen endpoint on one fixed it instantly. On a flat LAN it appears to work because scouting connects the nodes anyway — i.e. exactly in the environment the shortcut is sold for (unicast, where multicast does not traverse) it silently never syncs.

Fixed by adding `listen/endpoints tcp/[::]:7447` plus `listen/exit_on_failure: false`. The second flag is load-bearing: without it a second instance on the same host dies with `Can not create a new TCP listener bound to tcp/[::]:7447: Address already in use`, which `initialize()` then swallows into "distributed disabled". With it, instance 2 does not bind, connects outbound, and still exchanges events (verified with 3 sessions, scouting off). `[::]` matches zenoh's own default and keeps IPv6 peers reachable; verified binding and accepting an IPv4 connect on macOS and in a Linux `python:3.12-slim` container. The trade-off is documented: a symmetric cluster on a non-default port needs a raw config.
</details>

<details>
<summary><b>Smaller fixes in the same push</b></summary>

- **Subscriber closures pinned every event**, defeating `Event.instances`' WeakSet — one leaked event per page view under the pattern in finding 1. Gone with the `WeakMethod` fan-out.
- **`emit()` could raise out of user code.** A transient publish error propagated *after* `super().emit()` had already run the local callbacks, so a click handler aborted half-done — and it contradicted `initialize()`'s promise to keep running locally. `publish()` now logs.
- **`InvalidToken` was dropped in complete silence.** Now warned once per topic (no payload bytes, no secret).
- **`shutdown()` was one try block** — a single failing `undeclare()` skipped `session.close()` and leaked the transport — and it never cleared `_instance`, so a later emit published on a closed session. Now per-resource, always closes, clears the singleton.
- **The docs demo built its events per viewer.** `doc.demo` renders through `browser_window(..., lazy=True)`, so the constructor ran once per client — teaching precisely the pattern broken by finding 1. Hoisted to module scope, as `examples/distributed/main.py` already did.
- **The example rendered remote payloads as raw HTML.** `ui.html(emoji, sanitize=False)` is fed by the subscriber, so any node sharing the `storage_secret` could emit `` and execute script in every browser on every instance. Checked by publishing a hostile payload straight onto the wire topic from a separate session: both tabs rendered it inert, `onerror` stripped.
</details>

<details>
<summary><b>Verification, and what a second lineage caught</b></summary>

`pytest tests/test_distributed_events.py tests/test_event.py -rs` → **34 passed, 0 skipped** · `mypy ./nicegui` clean, 266 files · `pylint ./nicegui` 10.00/10 · pre-commit clean · `import website.examples` resolves.

Each new test was seen to fail with its fix reverted, e.g. the fan-out test gives `assert [] == ['hello']`, and the reverted `shutdown()` test additionally *hung* — the leaked-transport symptom itself.

Merged with current `main` before pushing; branch head is 0f9101e8 on top of c1459a0e.

Codex reviewed the diff and **blocked the first shutdown fix**, correctly: clearing `_instance` made a second `initialize()` reachable, but the old `_zenoh_setup_done` boolean then made every pre-existing event refuse to resubscribe — permanently detached after an in-process restart. Fixed at the root by replacing that boolean with `_subscribed_session_id`, so an event follows a replaced session. Its confirmation pass over the amended diff: *"Overall verdict: sound … I found no new blocker/major problem in the current code."*

One thing I stated wrongly in my previous comment: I said the silent `InvalidToken` would hide a typo'd `storage_secret`. It would not — a wrong secret changes the namespace, so the topics never match and no decryption is attempted. That case is still silent, and the new warning does not reach it; it covers clock skew past the TTL, foreign payloads on a matching topic, and forgeries.
</details>

Still your call, @falkoschindler / @rodja: whether `cryptography>=43.0.0` belongs in the `distributed` extra for payload encryption by default (98d2962).

PR #6089, opened 2026-06-03
Fix bad CLS on documentation homepage

### LINKED ISSUE #6087: Bad Cumulative Layout Shift (CLS) in NiceGUI documentation homepage
### First Check

### Example Code

```python

```

### Description

Since transition-all is used, the boxes are like this (sorry for low-res, it comes from Chrome DevTools and I don't know how to get a better one) for the first frame, then morph to current design in 200ms. 

Image

### NiceGUI Version

3.12.1 (main)

### Python Version

3.14

### Browser

Edge

### Operating System

macOS

### Additional Context

_No response_

falkoschindler:
### Motivation

The NiceGUI documentation homepage suffers from bad Cumulative Layout Shift (CLS): on the first frame the tiles render in their unstyled state and then visibly morph into the final design over ~200ms. Closes #6087.

The cards and CTA buttons use `transition-all`. Because NiceGUI applies the Tailwind classes via JavaScript _after_ the element first paints, `transition-all` animates _every_ property when the styled state is applied — including layout-affecting ones like `padding` and `border-radius`. That reflow is the morph captured in the issue's screenshot and what Chrome measures as CLS.

### Implementation

Scope each transition to only the properties that actually animate on hover. `transform` and `box-shadow` are composited and never trigger layout reflow, so the hover lift stays smooth while the layout-shifting animation is gone:

- `overview.py` (doc homepage tiles), `features_section.py`, `shared.py` (CTA button): `transition-all` → `transition-transform` (hover only does `-translate-y`).
- `examples_section.py`: `transition-all` → `transition-[transform,box-shadow]` (hover does `-translate-y` plus a shadow).

### Progress

### DIFF
--- a/website/components/examples_section.py
+++ b/website/components/examples_section.py
@@ -24,7 +24,7 @@ def create() -> None:
 def example_card(example: Example) -> None:
     """Render a single example card with screenshot, title, and description."""
     with ui.link(target=example.url).classes(
-        'rounded-2xl overflow-hidden transition-all duration-200 hover:-translate-y-0.5'
+        'rounded-2xl overflow-hidden transition-[transform,box-shadow] duration-200 hover:-translate-y-0.5'
         f' {d.BG_SURFACE} {d.BORDER} hover:{d.SHADOW_CARD}'
     ):
         with ui.element().classes(f'overflow-hidden aspect-video bg-white p-4 dark:brightness-[0.85] {d.BORDER_B}'):
--- a/website/components/features_section.py
+++ b/website/components/features_section.py
@@ -57,7 +57,7 @@ def create() -> None:
 
 def _card(icon: str, title: str, items: list[str]) -> None:
     """Render a single feature card with icon, title, and bullet list."""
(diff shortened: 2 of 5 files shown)


evnchn:

Hey Falko — I pushed two small polish commits on top of your CLS fix (`bc06c09`):

1. **Header star badge** — it had `duration-150` with no `transition-property`, so it fell back to `transition: all` and was animating the after-paint padding / border-radius / border — the same morph this PR targets. Scoped it to `background-color`, its only hover change.
2. **Sponsor button** — bumped its `rounded` → `rounded-xl` to match the documentation tiles directly below it.

Both verified on the locally running site; `pre-commit` clean. I'm approving now so you can just hit **merge** if you're happy — or drop that last commit if you'd rather keep this PR strictly scoped to the original CLS fix. No attachment either way. 🙂

evnchn (dismissed):
Approving — the original CLS fix is solid, and I've added two small header/sponsor polish tweaks on top (see comment). Merge or drop the last commit as you prefer.

falkoschindler:
Ok, but I chose to revert the radius. It is unrelated to the CLS, the xl rounding looks off, and a button and a card don't have to have _equal_ radii. Anyway, focussing on transitions lets us merge this PR.

PR #6092, opened 2026-06-05
Fix ui.sub_pages wildcard re-rendering and reject unsupported route patterns (#6090)

### LINKED ISSUE #6090: Subpage navigation not rendering
### First Check

### Example Code

```python
import time
import logging
from uuid import uuid4

from nicegui import ui, PageArguments

@ui.page("/")
@ui.page('/{_:path}')
def root():
    ui.label(f'This ID {str(uuid4())[:6]} changes only on reload.')
    ui.separator()
    ui.sub_pages({
        '/': main_page,
    }, show_404=False)

    # 2. Add navigation debugging directly to the router
    ui.context.client.sub_pages_router.on_path_changed(
        lambda path: print(f"Successfully navigated to: {path}")
    )

def main_page(args: PageArguments):
    ui.label(str(time.time()))
    ui.label(f"{args.path = }, {args.remaining_path = }")
    ui.link('Go to item page', '/item')
    ui.link('Go to item 1 page', '/item/1')
    ui.link('Go to item 1/1 page', '/item/1/1')


ui.run(storage_secret="change_in_production")
```

### Description

The browser starting with `/` renders correctly.  Navigating from `/` to any of the `/item` or `/item/1` or `/item/1/1` updates the URL in the browser and generates the debug message but the `main_page()` is not called with the updated remaining path.  From `/item` to `/item` and other transitions displays the correct updated content.

### NiceGUI Version

3.10.0

### Python Version

3.13.5

### Browser

Firefox

### Operating System

Linux

### Additional Context

I'm attempting to find a "fallback" when any of the explicit sub-pages are not matched by using the path to get a page from a database and render its content like a hierarchical wiki.  The path(s) may have `/` in them and could have additional filtering/rendering features like `/Chapter1/Snork?format=json` and having the additional page arguments is a nice thing.  Changing the code by removing the `@ui.page` decorators and passing the `root` into `ui.run()` doesn't change the behavior.

falkoschindler:
### Motivation

Fixes #6090.

Two issues surfaced from the same report:

1. **Wildcard sub pages did not re-render on client-side navigation.** When navigating between two URLs that match the _same_ route but differ in `remaining_path` (e.g. `/` → `/item`, both matching route `/`), `ui.sub_pages` kept the stale content and never called the builder again with the new `remaining_path`. With `show_404=False` this silently showed nothing new. The existing wildcard tests only used full-page `screen.open()` reloads, so they never exercised client-side navigation and missed this.

2. **Starlette-style convertor patterns failed silently.** Route patterns like `'/{_:path}'` are not supported by `ui.sub_pages` (only single-segment `'{name}'` parameters are). Such a pattern was treated as a literal route that no URL ever matches, so the route just never fired — with no error or warning.

### Implementation

1. **Re-render fix** (`_show`): the "path unchanged" fast-path that skips re-rendering now also requires that `remaining_path` is unchanged _or_ that a nested `ui.sub_pages` is present to absorb the change. A leaf sub pages whose `remaining_path` changed now falls through to a full re-render so the builder sees the new value. The nested case keeps the optimization (the parent is not rebuilt when only the child's segment changes).

2. **Unsupported patterns now raise** a `ValueError` at registration time (`__init__` and `add`) via a new `_validate_route`, pointing users to `show_404=False` + `PageArguments.remaining_path` for wildcard routing. The runtime matcher (`_match_path`) is unchanged.

### Progress

- [x] Documentation has been added (docstring note on the parameter limitation).
- [x] **Minor breaking change:**

Route patterns that previously failed silently (e.g. `'/{_:path}'`, `'/{name:path}'`) now raise `ValueError` at `ui.sub_pages(...)` / `.add(...)`. These never matched anything before, so only already-broken routes are affected; the fix is to use `show_404=False` + `remaining_path`.

### DIFF
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -30,6 +30,8 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
         Provides URL-based navigation between different views to build single page applications (SPAs).
         Routes are defined as path patterns mapping to page builder functions.
         Path parameters like "/user/{id}" are extracted and passed to the builder function.
+        Only single-segment "{name}" parameters are supported (not Starlette-style converters like "{name:path}");
+        for wildcard routing use ``show_404=False`` and read ``PageArguments.remaining_path``.
 
         *Added in version 2.22.0*
 
@@ -42,6 +44,8 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
         super().__init__()
         self._router = context.client.sub_pages_router
         self._routes = routes or {}
+        for path in self._routes:
+            self._validate_route(path)
         parent_sub_pages_element = next((el for el in self.ancestors() if isinstance(el, SubPages)), None)
         self._rendered_path = ''
         self._root_path = parent_sub_pages_element._rendered_path if parent_sub_pages_element else root_path
@@ -59,6 +63,7 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
         :param page: function to call when this path is accessed
         :return: self for method chaining
         """
+        self._validate_route(path)
         self._routes[path] = page
         self._show()
         return self
@@ -75,16 +80,19 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
         """Display the page matching the current URL path."""
         self._rendered_path = ''
         match = self._find_matching_path()
+        has_nested_sub_pages = any(isinstance(el, SubPages) for el in self.descendants())
         # NOTE: if path and query params are the same, only update fragment without re-rendering
         if (
+            # pylint: disable=too-many-boolean-expressions
             match is not None and
             self._match is not None and
             match.path == self._match.path and
+            (self._match.remaining_path == match.remaining_path or has_nested_sub_pages) and
             not self._required_query_params_changed(match) and
             not (self.has_404 and self._match.remaining_path == match.remaining_path)
         ):
             # NOTE: Even though our matched path is the same, the remaining path might still require us to handle 404 (if we are the last sub pages element)
-            if match.remaining_path and not any(isinstance(el, SubPages) for el in self.descendants()):
+            if match.remaining_path and not has_nested_sub_pages:
                 self._set_match(None)
             else:
                 self._handle_scrolling(match, behavior='smooth')
@@ -186,6 +194,18 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
                     fragment=fragment,
                 ), query_params
         return None, query_params
+
+    @staticmethod
+    def _validate_route(path: str) -> None:
+        for parameter in re.findall(r'\{(.*?)\}', path):
+            if not re.fullmatch(r'\w+', parameter):
+                raise ValueError(
+                    f'Invalid route "{path}": the parameter "{{{parameter}}}" is not supported. '
+                    'ui.sub_pages only supports single-segment "{name}" parameters, '
+                    'not Starlette-style converters like "{name:path}". '
+                    'For wildcard routing, use show_404=False and read PageArguments.remaining_path '
+                    '(see https://nicegui.io/documentation/sub_pages).'
+                )
 
     @staticmethod
     def _match_path(pattern: str, path: str) -> dict[str, str] | None:
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -88,8 +88,7 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
             self._match is not None and
             match.path == self._match.path and
             (self._match.remaining_path == match.remaining_path or has_nested_sub_pages or self._404_enabled) and
-            not self._required_query_params_changed(match) and
-            not (self.has_404 and self._match.remaining_path == match.remaining_path)
+            not self._required_query_params_changed(match)
         ):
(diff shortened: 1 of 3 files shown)


evnchn:

**TL;DR — negative finding, no action needed.** While reviewing I went down a side-tangent: I hypothesized this PR might *expose* a latent asymmetry where a leaf `ui.sub_pages` with an unconsumed path segment would render the page on a hard load but 404 on client-side navigation. I built an MRE to check — **it ruled the hypothesis out.** Behavior is consistent across hard load and client nav, and this PR actually *removes* the one real inconsistency (stale content on client nav). Sharing because ruling things out is still signal.

<details>
<summary>MRE + verified results (the hypothesis that didn't pan out)</summary>

**Hypothesis:** for a leaf sub pages (no nested, route `/sub`), a path with leftover segments like `/sub/x` would behave differently depending on how you got there — page on hard load vs. 404 on client nav.

**MRE** — single route, builder echoes `remaining_path`, root catch-all so the server serves deep paths:

```python
@ui.page('/')
@ui.page('/{_:path}')
def index():
    ui.sub_pages({'/sub': sub_page}, show_404=show_404)

def sub_page(args):
    ui.label(f'PAGE remaining={args.remaining_path!r}')
```

Exercised each path via the `User` fixture (HTTP status + rendered labels) and via a real browser (client-side link clicks). Every cell below is **observed output**, not predicted:

| Navigation | `show_404=True` | `show_404=False` |
|---|---|---|
| hard load `/sub` | 200 → `PAGE remaining=''` | 200 → `PAGE remaining=''` |
| hard load `/sub/x` | **HTTP 404** + `404: sub page /sub/x not found` | 200 → `PAGE remaining='/x'` |
| client nav → `/sub/x` | shows `404` label | re-renders `PAGE remaining='/x'` *(this PR's fix)* |

**Conclusion — no asymmetry:**
- `show_404=True`: an unconsumed segment yields 404 on **both** hard load (a genuine HTTP 404 status) and client nav. Consistent.
- `show_404=False`: the page renders with `remaining_path` populated on **both** hard load and — after this PR — client nav. Consistent. The PR is exactly what makes client nav match hard load.

**Where my initial code-trace went wrong:** I read the full-re-render branch (`_render_page` returns `True` unconditionally) and assumed a hard load of `/sub/x` would render the page. It doesn't — `show_404=True` initial server render raises `HTTPException(404)`, turning a leaf sub-page 404 into a real HTTP 404 status, so the branch I imagined never fires. The empirical run corrected the trace.

Net: nothing to fix here, and it's a small extra data point that the PR improves consistency rather than exposing one.

</details>

evnchn:

**TL;DR:** Reviewed — the fix looks right, and we verified the two load-bearing properties rather than just reading the diff (details below). One observation before stamping it: the neighboring condition `not (self.has_404 and self._match.remaining_path == match.remaining_path)` appears **provably dead**, and removing it would shrink the very boolean expression this PR had to `pylint: disable` for. Posting as a comment instead of an approval so you can sanity-check the proof first — happy to approve as-is or with the cleanup, whichever you prefer.

<details>
<summary><b>What we verified about the fix itself</b></summary>

- **The new clause strictly narrows the fast path.** Old fast-path cases minus exactly (`remaining_path` changed ∧ no nested sub-pages ∧ `show_404=False`) — which is precisely the #6090 bug case, now falling through to a full re-render. No new code paths: the nested-sub-pages optimization and the 404-on-deeper-navigation behavior are preserved (walked the truth table against `_set_match` / `_find_matching_path` on the PR head).
- **Registration-time validation matches runtime capability exactly.** `_validate_route` accepts `\w+` parameters; `_match_path`'s substitution regex is `\{(\w+)\}` — same set, no drift in either direction.
- The new test exercises the real regression trigger (client-side `screen.click` navigation rather than full-page reloads), and the `!r` formatting distinguishes empty-from-missing remaining path. 👍

</details>

<details>
<summary><b>The dead-clause proof</b></summary>

Claim: in `_show()`'s fast-path condition, `not (self.has_404 and self._match.remaining_path == match.remaining_path)` always evaluates to `True`.

1. `has_404` is written in exactly two places repo-wide (grep-verified): `__init__` (`False`) and `_set_match` (`self.has_404 = match is None` — always coupled with `self._match = match` on the line above). `sub_pages_router.py:140` only *reads* it.
2. `self._match` is written only by `_set_match` and `_reset_match` (which sets `None`).
3. Therefore whenever `self._match is not None`, the most recent `_match` write was `_set_match(<non-None>)`, which also set `has_404 = False` — and nothing can flip `has_404` to `True` without also nulling `_match`. Invariant: **`self._match is not None ⟹ self.has_404 is False`**.
4. The fast-path condition requires `self._match is not None` two conjuncts earlier, so by the time the `has_404` clause evaluates, `has_404` is always `False` → the clause is a tautology.

It appears to have been born dead in #4821 (`bc17b1bc`): the original `_set_match` already had the same write-coupling (`self._match = match` + `has_404 = False` on the non-None branch).

Only caveat: `has_404` has no underscore, so a subclass could in principle poke it directly; nothing in-tree does.

Removing the clause would also let the condition drop below the `too-many-boolean-expressions` threshold this PR had to disable.

</details>

evnchn (approved):

Approving. Verified the new commit locally: the diff is exactly the proved-dead clause removal and nothing else, and both sub-pages suites pass on the new head (71/71, `test_sub_pages.py` + `test_sub_pages_match_path.py`).

One correction to my earlier comment: I claimed removing the clause would let the condition drop below the `too-many-boolean-expressions` threshold — it doesn't (the operands inside the parenthesized or-group count individually, verified with pylint), so the `disable` stays. Everything else stands. Thanks for picking up the cleanup!

PR #5651, opened 2026-01-16
fix documentation for native mode settings in packaged apps

falkoschindler:
### Motivation

Fixes #4842

When packaging NiceGUI apps with native mode (using nicegui-pack, PyInstaller, etc.), `app.native` settings like `window_args['transparent']` were being ignored. This was due to incorrect documentation advising users to place `freeze_support()` "before anything else."

The root cause: in a frozen subprocess, `freeze_support()` intercepts execution and diverts to the multiprocessing worker. Any code after `freeze_support()` never runs in the subprocess, so native settings defined later are not applied.

### Implementation

Updated the documentation to:

1. Clarify that `app.native` settings must be defined **before** calling `freeze_support()`
2. Rename "macOS Packaging" to "Packaging with Native Mode" since this applies to all platforms (Windows, macOS, Linux all use spawn/forkserver)
3. Add a cross-reference from the main guard section to the packaging section
4. Provide a clear code example showing the correct order

This is a documentation-only fix. No code changes are needed because the existing behavior is correct — users just need to structure their code properly.

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] This is not a security issue.
- [x] Pytests are not necessary.
- [x] Documentation has been improved.

Issue #4842, opened 2025-06-05
pyinstaller native transparent not working


h0uter:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
# macOS packaging support
from multiprocessing import freeze_support

freeze_support()  # noqa

from nicegui import app, ui

app.native.window_args["transparent"] = True
app.native.window_args["frameless"] = True


def main():
    with ui.card():
        ui.label("Hello, NiceGUI!")

    ui.run(
        reload=False,
        native=True,
    )


if __name__ in ("__main__", "__mp_main__"):
    main()
```

## window_args not respected with nicegui-pack

build the app using nicegui-pack

```
nicegui-pack --name mytest --windowed main.py
```

start the MacOS app by running

```
./dist/mytest.app/Contents/MacOS/mytest
```

The window shown is not transparent and not frameless.

I expected it to be frameless and transparent.

## Running without pyinstaller works

If I run the app locally by running the python file.

The window shown is frameless and transparent.

### NiceGUI Version

2.19.0

### Python Version

3.13.3

### Browser

Other

### Operating System

macOS

## using ui.run(frameless=True) works as expected

If I use the option `ui.run(frameless=True)` in my code

```py
# macOS packaging support
from multiprocessing import freeze_support

freeze_support()  # noqa

from nicegui import app, ui

app.native.window_args["transparent"] = True

def main():
    with ui.card():
        ui.label("Hello, NiceGUI!")

    ui.run(
        reload=False,
        native=True,
        frameless=True
    )


if __name__ in ("__main__", "__mp_main__"):
    main()
```

and then  build the app using nicegui-pack

```
nicegui-pack --name mytest --windowed main.py
```

start the MacOS app by running

```
./dist/mytest.app/Contents/MacOS/mytest
```

The window shown is frameless but not tranparent.

falkoschindler:
Thanks for reporting this issue, @h0uter! I can reproduce the problem, but have no idea what could be the reason.
Does anyone have an idea? Does it happen on macOS only?

himbeles:
[...]

PR #5648, opened 2026-01-15
Fix native window args being ignored in packaged apps

falkoschindler:
### Motivation

Fixes #4842

When packaging a NiceGUI app with PyInstaller/`nicegui-pack`, settings configured via `app.native.window_args`, `app.native.settings`, and `app.native.start_args` were being ignored. This happened because PyInstaller uses the "spawn" multiprocessing start method, which creates a fresh Python interpreter that doesn't inherit the parent process's modified module state.

The same issue caused `window_args` set inside a main guard (`if __name__ == '__main__':`) to be ignored, even when running without PyInstaller.

### Implementation

The native configuration dictionaries (`window_args`, `settings`, `start_args`) are now passed explicitly through the multiprocessing `args` tuple to the subprocess, rather than being read from `core.app.native.*` inside the subprocess. This ensures the configuration survives the "spawn" start method.

Additionally, the documentation has been updated to:
- Note that code under a main guard now correctly applies configuration changes (since 3.6.0)
- Remove the outdated macOS `freeze_support()` recommendation (NiceGUI handles this internally)

### Progress

- [x] I chose a meaningful title that completes the sentence: "If applied, this PR will..."
- [x] The implementation is complete.
- [x] This PR addresses no security issue.
- [x] Pytests are not possible (with the current test infrastructure).
- [x] Documentation has been updated.

PR #6093, opened 2026-06-05
Fix dropped `app.native.*` args in spawned window process

### LINKED ISSUE #6082: Unable to set `app.native.window_args` when starting my app from package script.
### First Check

### Example Code

```python

```

### Description

I want my package to install from pip and start from the terminal with a command, so I configured the package installation to make an executable `my-app` available.

```toml
# pyproject.toml

...

[project.scripts]
my-app = "my_app.main:run"
```

`main.py` parses arguments so the user can decide wether to start my program as a server or as a local native app, and the script's contents are similar to this:

```python
# src/my_app/main.py
import argparse
import multiprocessing
from nicegui import ui, app
from my_app.pages import root


MIN_WINDOW_SIZE = (800, 600)
app.native.window_args["min_size"] = MIN_WINDOW_SIZE


def get_argparser() -> argparse.ArgumentParser:
    parser = argparse.ArgumentParser(prog="my-app")
    parser.add_argument("-n", "--as-native", action="store_true")
    return parser


@ui.page("/")
def root_page():
    return root.page(ui=ui)


def run():
    multiprocessing.freeze_support()
    parser = get_argparser()
    args, _ = parser.parse_known_args()
    if hasattr(args, "help"):
        parser.print_help()
        quit()
    try:
        ui.run(
            show=False,
            native=args.as_native,
            reload=False,
        )
    except KeyboardInterrupt:
        print("\nInterruption requested, service shutting down...")


    run()
```

After installing the package locally in a Linux system, everything works as expected:
- Calling `my-app` starts the server
- Calling `my-app --as-native` starts in native mode, and the window size limit is respected
- Calling `my-app --help` shows argparse's default help message and quits

On a Windows system, calling `my-app --as-native` will never respect window size limit, but running the script directly with `python src/my_app/main.py --as-native` preserves the excpected behavior.

### NiceGUI Version

3.12

### Python Version

3.14.4

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

falkoschindler:
### Motivation

`app.native.window_args`, `app.native.settings` and `app.native.start_args` are read directly off a freshly-imported `nicegui.app` inside the spawned pywebview process. They therefore only take effect when Python's `multiprocessing` spawn can re-execute the parent's main module as `__mp_main__`.

That re-execution silently fails in several common setups, dropping the user's config with no warning:

- pip `[project.scripts]` console-scripts on Windows (the launcher binary isn't a re-runnable `.py` file) — reported in #6082
- PyInstaller / `nicegui-pack` (#4842, docs-patched in #5651)
- `python -m mypkg` without a correctly wired `__main__.py`
- Briefcase / py2exe / custom launchers

Closes #6082.

### Implementation

`activate()` now forwards the **picklable subset** of each of the three native config dicts through the spawn `args` tuple. Inside `_open_window()` these are merged with whatever the child re-imported on its own:

```python
window_args = {**(passed or {}), **core.app.native.window_args}
```

The re-imported values win where they exist, so the previously working path is unchanged; the forwarded values only fill the gap when the main module could not be re-executed in the child.

Values that cannot cross the spawn boundary (e.g. a stateful `js_api` object) are skipped via a new `_picklable_subset()` helper, which emits a one-time `warn_once` per dropped key. This replaces today's **silent** drop with an explicit, actionable warning — and such values were already lost under spawn before this change.

### Progress

### DIFF
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -38,6 +38,7 @@ def _open_window(
     window_args: dict[str, Any] | None = None,
     settings: dict[str, Any] | None = None,
     start_args: dict[str, Any] | None = None,
+    dropped_keys: dict[str, list[str]] | None = None,
 ) -> None:
     while not helpers.is_port_open(host, port):
         time.sleep(0.1)
@@ -46,6 +47,16 @@ def _open_window(
     window_args = {**(window_args or {}), **core.app.native.window_args}
     settings = {**(settings or {}), **core.app.native.settings}
     start_args = {**(start_args or {}), **core.app.native.start_args}
+
+    # Warn only about keys the parent could not pickle AND re-import did not restore here, i.e. that are truly lost.
+    for name, merged in [('window_args', window_args), ('settings', settings), ('start_args', start_args)]:
+        for key in (dropped_keys or {}).get(name, []):
+            if key not in merged:
+                helpers.warn_once(
+                    f'Could not forward app.native.{name}["{key}"] to the native window process '
+                    'because it is not picklable, so it is ignored. '
+                    'See https://github.com/zauberzeug/nicegui/issues/6082 for details.'
+                )
 
     window_kwargs = {
         'url': helpers.format_url(protocol, host, port),
@@ -186,32 +197,31 @@ def activate(protocol: str, host: str, port: int, title: str, width: int, height
     mp.freeze_support()
     native.create_queues()
     event_manager.start()
+    window_args, dropped_window_args = _split_picklable(core.app.native.window_args)
+    settings, dropped_settings = _split_picklable(core.app.native.settings)
+    start_args, dropped_start_args = _split_picklable(core.app.native.start_args)
+    dropped_keys = {'window_args': dropped_window_args, 'settings': dropped_settings, 'start_args': dropped_start_args}
     args = (protocol, host, port, title, width, height, fullscreen, frameless,
             native.method_queue, native.response_queue, native.event_sender, favicon,
-            _picklable_subset('window_args', core.app.native.window_args),
-            _picklable_subset('settings', core.app.native.settings),
-            _picklable_subset('start_args', core.app.native.start_args))
+            window_args, settings, start_args, dropped_keys)
     process = native.SPAWN_CONTEXT.Process(target=_open_window, args=args, daemon=True)
     process.start()
 
     Thread(target=check_shutdown, daemon=True).start()
 
 
-def _picklable_subset(name: str, config: dict[str, Any]) -> dict[str, Any]:
-    """Return the picklable items of a native config dict so they can be passed to the spawned window process."""
-    result: dict[str, Any] = {}
+def _split_picklable(config: dict[str, Any]) -> tuple[dict[str, Any], list[str]]:
+    """Split a native config dict into its picklable items and the keys that cannot cross the spawn boundary."""
+    picklable: dict[str, Any] = {}
+    dropped: list[str] = []
     for key, value in config.items():
         try:
             pickle.dumps(value)
         except Exception:  # pylint: disable=broad-except
-            helpers.warn_once(
-                f'Cannot forward app.native.{name}["{key}"] to the native window process because it is not picklable. '
-                'It is ignored unless your main module gets re-imported in the spawned process. '
-                'See https://github.com/zauberzeug/nicegui/issues/6082 for details.'
-            )
-            continue
-        result[key] = value
-    return result
+            dropped.append(key)
+        else:
+            picklable[key] = value
+    return picklable, dropped
 
 
 def find_open_port(start_port: int = 8000, end_port: int = 8999) -> int:
--- a/nicegui/native/native_mode.py
+++ b/nicegui/native/native_mode.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import _thread
 import multiprocessing as mp
+import pickle
 import queue
 import socket
 import sys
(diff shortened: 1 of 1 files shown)


evnchn:
Statement to attest and then help review:

> Using picklable subset means that we aren't pickling everything (rejected per https://github.com/zauberzeug/nicegui/issues/4842#issuecomment-3759286515) AND we aren't throwing the baby out with the bathwater by ensuring picklable stuff do get across, only the non-picklable shenanigans gets dropped with a warn. 

Dispatching agent to verify now.

evnchn:
*Posted by Claude Code (Opus 4.8) on @evnchn's behalf — dispatched to verify the attestation against the actual diff + branch source, with the warning claim backed by an MRE.*

## ✅ Attestation holds

> Using picklable subset means that we aren't pickling everything … AND we aren't throwing the baby out with the bathwater by ensuring picklable stuff do get across, only the non-picklable shenanigans gets dropped with a warn.

All three sub-claims check out, **and** the previously-working path is preserved (verified that all three configs default to `field(default_factory=dict)`, so when re-import fails the merge `{**passed, **{}}` keeps the forwarded subset; when it works, the re-imported dict — including non-picklable `js_api` — wins).

<details>
<summary>Attestation evidence</summary>

| Claim | Verdict | Why |
|---|---|---|
| not pickling everything (cf. #5648 rejection) | ✅ | #5648 was rejected because forwarding args as subprocess params *forces* picklability → crashes on non-picklable values. `_picklable_subset` probes each value with `pickle.dumps` and forwards only survivors — never all-or-crash. |
| picklable stuff gets across | ✅ | Survivors ride the spawn `args` tuple and are merged in `_open_window`. |
| only non-picklable dropped, with a warn | ✅ | `except Exception → helpers.warn_once(...) → continue` skips only the unpicklable key. |

No regression: in the working `python main.py` path the child re-imports the module, so `core.app.native.window_args` still carries `js_api`, which wins the merge — old behaviour byte-for-byte unchanged.
</details>

## ⚠️ But the warning is spurious on the normal `python main.py` path — proven with an MRE

`_picklable_subset` runs **unconditionally** in the *parent* ([`native_mode.py` L191-193](https://github.com/zauberzeug/nicegui/blob/fix/native-args-spawn-propagation/nicegui/native/native_mode.py#L191-L193)), which always holds the real `js_api`. It can't know whether the child will re-import. So on a perfectly-working dev setup with a non-picklable `js_api`, the user sees the warning at every launch even though the value arrives fine via re-import and **nothing is actually lost**.

The hedge ("It is ignored *unless* your main module gets re-imported") makes it technically-not-false, but the common dev case (`python main.py`, spawn re-imports fine) is exactly the case where it *is* re-imported — so the noise lands mostly on users whose apps work.

<details>
<summary>MRE — two runnable scripts against this branch (Python 3.14, spawn)</summary>

**MRE-1 — the real `_picklable_subset` warns + drops `js_api`:**
```python
import logging, io
buf = io.StringIO()
logging.getLogger('nicegui').addHandler(logging.StreamHandler(buf))
logging.getLogger('nicegui').setLevel(logging.WARNING)
from nicegui.native import native_mode
class JsApi:
    def __init__(self):
        import threading; self._lock = threading.Lock()  # non-picklable
config = {'js_api': JsApi(), 'on_top': True, 'resizable': False}
result = native_mode._picklable_subset('window_args', config)
print('FORWARDED:', list(result), '| DROPPED:', [k for k in config if k not in result])
print(buf.getvalue().strip())
```
```
FORWARDED: ['on_top', 'resizable'] | DROPPED: ['js_api']
Cannot forward app.native.window_args["js_api"] to the native window process because it is not picklable. ...
```

**MRE-2 — on the happy path the dropped `js_api` is restored via re-import (exact `_open_window` merge):**
```python
import os, multiprocessing as mp
from nicegui import app
class JsApi:
    def __init__(self):
        import threading; self._lock = threading.Lock()
app.native.window_args['js_api'] = JsApi()      # top-level: runs in parent AND spawn re-import
app.native.window_args['on_top'] = True
print(f'[pid {os.getpid()} import] keys = {list(app.native.window_args)}')
def child(passed_subset):
    from nicegui import core
    merged = {**(passed_subset or {}), **core.app.native.window_args}   # exact _open_window merge
    print(f'[child {os.getpid()}] forwarded={list(passed_subset)}  merged={list(merged)}  js_api restored={"js_api" in merged}')
if __name__ == '__main__':
    mp.get_context('spawn').Process(target=child, args=({'on_top': True},)).start()
```
```
[pid 46296 import] keys = ['js_api', 'on_top']
[pid 46299 import] keys = ['js_api', 'on_top']        <- spawn child RE-IMPORTS the module
[child 46299] forwarded=['on_top']  merged=['on_top', 'js_api']  js_api restored=True
```

**Together:** the warning fires (MRE-1) yet the value survives via re-import (MRE-2) on the same `python main.py` path ⇒ the warning is shown to users whose setup works.
</details>

**Not a blocker** — `warn_once` dedups, and the trade-off (explicit-over-silent) is a deliberate, reasonable call that aligns with the warn-don't-raise convention. The parent genuinely can't tell ahead of time whether the child will re-import, so firing-only-when-it-matters isn't cleanly achievable. Options if the noise feels off: soften the wording toward "may be dropped" / "if your app is packaged", or downgrade to a debug-level log. Leaving the call to @falkoschindler.

falkoschindler:
Thanks for the thorough verification and the MREs, @evnchn — the spurious-warning point is spot on, and I agree it shouldn't fire on the normal `python main.py` path where the value arrives fine via re-import.

Fixed in 112108756 by moving the detection into the child, where we actually know whether a value was lost: `_split_picklable` now only *collects* the unpicklable keys in the parent, forwards them to `_open_window`, and the warning fires there only for keys that are still missing after the re-import merge. So on the happy path (re-import restores `js_api`) it stays silent; it only warns when the value is genuinely dropped — and it rides the same child-side logging the function already uses.

Verified both branches of your MRE: no warning when re-import restores the key, warning only when the child starts with an empty dict.

evnchn (approved):
Approve with trust that MREs pass after fix (promised from Falko) so that I save a re-verification run. 

The MREs are gating the merge, after all. All else I'm happy with.

PR #6095, opened 2026-06-05
Add a note about large datasets to the ui.plotly docs

### LINKED ISSUE #3340: Updating Ploty plot with large dataset causes the UI to hang

falkoschindler:
### Motivation

Issue #3340 reports that updating a `ui.plotly` chart with a large dataset can freeze the UI.
After investigation it turned out the difference between a hanging and a smooth UI comes down to the data type handed to Plotly:

- A **Python list** is kept by Plotly as a plain JSON array of numbers, which the browser has to parse number-by-number — freezing the UI for large datasets.
- A **NumPy array** (or pandas Series) is encoded by Plotly as a compact base64 typed-array (`bdata`), which the browser renders far more efficiently.

This is entirely Plotly behavior (not NiceGUI's transport), so the practical takeaway is worth documenting.

### Implementation

Added a "Large datasets" demo to the `ui.plotly` documentation explaining the recommendation
(pass NumPy arrays / pandas Series, avoid `.tolist()`) with a runnable 100k-point `go.Scattergl` example.

### Progress

### DIFF
--- a/website/documentation/content/plotly_documentation.py
+++ b/website/documentation/content/plotly_documentation.py
@@ -103,8 +103,9 @@ def plot_events():
     When plotting a large number of data points, pass NumPy arrays (or pandas Series) to Plotly rather than Python lists,
     and avoid converting them back with `.tolist()`.
     For NumPy arrays, Plotly uses a compact binary encoding (base64-encoded typed arrays)
-    that the browser can render much more efficiently than parsing millions of numbers from a plain JSON array,
-    which can otherwise freeze the UI.
+    that makes the payload roughly half the size — far less data to transfer, parse and keep in memory —
+    which is what avoids freezing the UI on large updates.
+    This requires Plotly 6.0 or newer, where the binary encoding is the default.
 ''')
 def large_datasets():
     import numpy as np
--- a/website/documentation/content/plotly_documentation.py
+++ b/website/documentation/content/plotly_documentation.py
@@ -99,4 +99,22 @@ def plot_events():
     plot.on('plotly_click', ui.notify)
 
 
+@doc.demo('Large datasets', '''
+    When plotting a large number of data points, pass NumPy arrays (or pandas Series) to Plotly rather than Python lists,
+    and avoid converting them back with `.tolist()`.
+    For NumPy arrays, Plotly uses a compact binary encoding (base64-encoded typed arrays)
+    that the browser can render much more efficiently than parsing millions of numbers from a plain JSON array,
+    which can otherwise freeze the UI.
+''')
+def large_datasets():
+    import numpy as np
+    import plotly.graph_objects as go
+
+    x = np.linspace(0, 10, 100_000)
+    y = np.sin(x) + np.random.normal(0, 0.1, x.size)
+    fig = go.Figure(go.Scattergl(x=x, y=y, mode='markers', marker=dict(size=2)))
+    fig.update_layout(margin=dict(l=0, r=0, t=0, b=0))
+    ui.plotly(fig).classes('w-full h-40')
+
+
 doc.reference(ui.plotly)


evnchn:
*Posted by @evnchn; investigation & comment authored autonomously by Claude Code (Opus 4.8) on Evan's behalf.*

## TL;DR — not too good to be true on the facts, but the *reason* and the *demo* both need a tweak

I empirically reproduced the whole chain end-to-end (Plotly 6.8, NiceGUI 3.12, real Chromium via Playwright). **The core claim is correct:** NumPy/pandas → compact base64 `bdata`, Python `list` → plain JSON number array, and NiceGUI's transport faithfully carries the `bdata` to the browser. So the recommendation ("pass NumPy/pandas, avoid `.tolist()`") is **sound and worth documenting.** ✅

Two things I'd push back on, both backed by measurements below:

1. **The stated mechanism is off.** The win is *not* that "plotly.js renders a typed array far more efficiently." Once the data is in the browser, **render time is identical for typed vs. plain arrays** at every scale I tested. The real, measurable advantage is the **~1.77× smaller wire payload** (less to transfer, less to `JSON.parse`, less browser memory) — which is exactly what relieves the message‑size/memory pressure that froze #3340.
2. **The demo doesn't actually demonstrate the problem.** At 100k points with `Scattergl` (WebGL), a plain `list` renders just as smoothly as NumPy (~30 ms, no freeze). The smoothness a reader sees is **WebGL**, not `bdata`. So the demo documents the advice without letting anyone observe the difference.

<details>
<summary><b>Evidence — wire payload, render timings, methodology</b></summary>

### 1. Encoding & transport (claim confirmed ✅)
`fig.to_plotly_json()` (which is exactly what `nicegui/elements/plotly/plotly.py::_get_figure_json` calls, then orjson-serializes) yields, for the trace `x`:

| input | trace `x` on the wire | `bdata` present? |
|---|---|---|
| `np.ndarray` | `{'dtype':'f8','bdata':'<base64>'}` | **yes** |
| `pd.Series` | `{'dtype':'f8','bdata':'<base64>'}` | **yes** |
| `list` (`.tolist()`) | `[0.0, 0.0001, ...]` | no |

Confirmed live in the browser too: on the NumPy page `gd.data[0].x` arrives as the encoded object; on the list page it's a plain `Array(100000)`. Both render a WebGL canvas. So **NiceGUI's transport preserves `bdata`** — the PR's "this is entirely Plotly behavior, not NiceGUI's transport" holds.

### 2. Wire payload size — this is the real lever (NiceGUI's `to_plotly_json` + orjson path)
```
N=   10,000  numpy=  220,108 B  list=  384,625 B  → list is 1.75× larger
N=  100,000  numpy=2,140,108 B  list=3,781,812 B  → list is 1.77× larger
N=1,000,000  numpy= 21.3 MB     list= 37.8 MB     → list is 1.77× larger
```

### 3. Render time — typed vs. plain is a wash (`Plotly.newPlot`, min of 3, real Chromium)
```
scattergl  100k :  typed 34.8 ms   plain 31.4 ms
scattergl  1M   :  typed  246 ms   plain  227 ms
scatter(SVG) 200k: typed 1091 ms   plain 1143 ms
```
Plain arrays are, if anything, marginally *faster*. Render cost tracks **N and trace type** (SVG `scatter` is ~5× slower than `scattergl`), not the encoding. And raw `JSON.parse` of 100k numbers is only ~1.9 ms — fast and native — so "the browser has to parse number‑by‑number, freezing the UI" overstates the parse cost at these sizes. The freeze in #3340 (2.4M points, *list*, updating, regular `scatter`) is a **payload/memory** pathology that the 1.77× smaller `bdata` payload fixes — not a typed-array *render* speedup.

### 4. Minor: version dependence
`bdata` is the **default only since plotly.py 6.0** (2025‑01‑28, [plotly/plotly.py#4470](https://github.com/plotly/plotly.py/pull/4470)). On older Plotly the advice may quietly do nothing. Probably worth a one‑line note, since NiceGUI doesn't pin a recent Plotly.

</details>

### Suggestions (optional — your call, @falkoschindler)
- **Fix the mechanism wording** so it points at payload size, e.g. *"Plotly encodes NumPy/pandas as a compact base64 typed array (`bdata`), making the payload ~1.8× smaller — far less to transfer, parse and hold in memory — which is what avoids the freeze on large updates."* (Drop the "renders far more efficiently" framing — measured identical.)
- **Make the demo reproduce the contrast** so it's not a no-op: a larger `N` and/or regular `go.Scatter`, ideally behind an **"Update" button** (the actual #3340 scenario is *updating* an existing chart, where the list path stalls and the NumPy path stays live). As written, 100k + `Scattergl` is smooth either way.
- Optionally add the "needs modern Plotly (≥6.0 for default `bdata`)" note.

Happy to push a revised demo if useful. Full repro scripts (encoding/transport/wire-size + the in-browser timing harness) available on request.

evnchn:
@falkoschindler So the Super-TLDR appears to be: just need phrasing difference. 

**Please do not modify the demo so that the slowdown magnitude becomes noticable on your mac, because that would be lethal dose to the celerons and chromebooks alike. We just need to show the spirit is OK!**

falkoschindler:
Thanks for the thorough repro, @evnchn — the payload-size correction is spot on. I've reworded the note in 97f392091:

- Dropped the "renders far more efficiently" framing; it now attributes the win to the ~half-size payload (less to transfer, parse and keep in memory), which is what relieves the freeze.
- Added the "requires Plotly ≥ 6.0 (where the binary encoding is the default)" caveat.

I kept the demo lightweight (100k + `Scattergl`) as you asked — no point handing Chromebooks a lethal dose just to make the contrast visible. 🙂

evnchn (approved):
Then we should be good to go!

PR #6096, opened 2026-06-05
chore: improve nicegui maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/test_fab.py, tests/test_log.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/tests/test_fab.py
+++ b/tests/test_fab.py
@@ -36,3 +36,14 @@ def test_fab(screen: Screen):
 
     screen.click('Toggle FAB')
     screen.should_contain('FAB is open')
+
+
+def test_fab_returns_self(screen: Screen):
+    @ui.page('/')
+    def page():
+        fab = ui.fab('menu')
+        assert fab.open() is fab
+        assert fab.close() is fab
+        assert fab.toggle() is fab
+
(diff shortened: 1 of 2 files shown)


HosseyNJF:
Seems like a bot account, recently created and creating PRs with the same format across many big repos (just adding a couple of tests and claiming the "maintenance path" was improved. The changes are so insignificant I'd say it's not worth spending time to go through the whole process of merging.

falkoschindler:
Thanks for the heads up, @HosseyNJF! 🙏

evnchn:
Yes it is. Probably some other autonomous agent that's poorly tuned.

image

I mean I can't really take you seriously when you pin `requests` and `flask`, both of which is **highly incompatible** with NiceGUI which expects (1) async HTTP request library, and (2) ASGI, not WSGI

image

I think well-tuned and well-used agents can lift weight, but definitely not this, sorry for the fiery takedown...

PR #6101, opened 2026-06-10
Reserve NOTE comment markers for drawing special attention

falkoschindler:
### Motivation

Using `NOTE:` markers is a relict from times when we used comments more sparingly and felt like we should indicate the purpose of each of them: TODO, FIXME, NOTE, HACK.
Leftover TODOs and FIXMEs in code are bad practice anyway, and a `NOTE:` prefix on a comment that merely explains the adjacent code doesn't carry any meaningful information -- what else is a comment supposed to be?

But the marker remains useful for drawing special attention, e.g. to changes that need to be mirrored elsewhere (like the duplicated dependency lists in pyproject.toml) or hidden cross-file coupling that an edit would silently break.
`HACK:` also remains a valid marker for knowingly suboptimal solutions.

### Implementation

- Strip the `NOTE:` prefix from ~115 comments and docstrings that merely explain the adjacent code, keeping their content unchanged.
- Keep `NOTE:` on the 10 attention-seeking comments, e.g. the mirrored dependency lists in pyproject.toml, the deliberately empty ROS entry point, the Selenium-only global in scene.js, and the test-ordering constraint in test_highchart.py.
- Update CONTRIBUTING.md, REVIEW.md and nicegui/llms.md to describe the new convention: plain comments by default, `NOTE:` only to draw special attention.

### Progress

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,8 +148,9 @@ The conventions below cover both general Python style and NiceGUI-specific patte
 
   - Follow **PEP 8** with a 120 character line length
   - Use single quotes in Python, double quotes in JavaScript
-  - Use f-strings wherever possible (mark performance-critical exceptions with `# NOTE:`)
-  - Use `# NOTE:` prefix for important implementation details and non-obvious code
+  - Use f-strings wherever possible (mark performance-critical exceptions with a comment)
+  - Explain important implementation details and non-obvious code with comments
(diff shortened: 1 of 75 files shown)


evnchn (approved):
+119 -119 with 1 line real addition (changing the contributing docs) and 1 line real removal (from remove `# NOTE` standalone line)

I'm pretty un-opinionated but I think at the very least we can agree that this shapes the comments more how Claude models like it, which might just boost our productivity more 🤷

PR #6103, opened 2026-06-10
Fall back to polling when the WebSocket cannot connect

### LINKED ISSUE #5802: NiceGUI iOS Safari Page Reload Bug (+root cause +solution!)

falkoschindler:
### Motivation

In #5802 an iOS 26 Safari user reported that pages served over plain HTTP reload every few seconds, indefinitely. The root cause is a WebKit networking regression ([bug 302823](https://bugs.webkit.org/show_bug.cgi?id=302823)): the WebSocket connection hangs in a pending state, socket.io's 20 s connect timeout fires, and `nicegui.js` reloads the page — which changes nothing, forever.

What makes this worse on our side: the default `transports: ['websocket', 'polling']` never actually falls back to polling. Since socket.io-client v3, only the first transport in the list is tried, so a client with a broken WebSocket has no path to a working connection even though long-polling would work fine.

### Implementation

A broken WebSocket can fail in two ways, and we now handle both:

- _It errors out quickly_ (e.g. a proxy rejecting the upgrade): the new `tryAllTransports: true` socket.io option (available since the bundled client 4.8.0) makes the client try polling within the same connection attempt, so the app connects without noticeable delay.
- _It hangs silently_ (the iOS 26 case): no transport error ever fires, so `tryAllTransports` cannot help. On the first `"timeout"` connect error the client now switches the transport order to `["polling", "websocket"]` and lets socket.io's built-in reconnection retry, instead of reloading.

On iOS 26 Safari over HTTP the app then works via long-polling (the later WebSocket upgrade probe fails harmlessly). The happy path is unchanged: clients whose WebSocket works never hit this code. The timeout fallback only applies if the configured transports favor websocket but also allow polling; with an explicit `['websocket']`-only configuration it is skipped (and `tryAllTransports` is a no-op for single-transport lists).

Note: socket.io replaces the transport names in the options array with transport classes in place, so the configured names are snapshotted before creating the socket.

If both transports keep failing, the page still reloads on every connect timeout, as before. An earlier version of this PR capped these reloads at three attempts and showed an error popup instead, but we decided to keep the retry-forever behavior (see review discussion): it is self-healing for unattended deployments like dashboards and kiosks, and with the polling fallback in place the loop only occurs in the narrow case where the page itself loads fine but both transports time out.

Verified locally with `WebSocket` stubs: with a stub that never connects (the WebKit bug), the client falls back to polling after 20 s and the page becomes fully interactive; with a stub that errors immediately, it connects via polling right away; a server restart under an open page still recovers with a single reload.

The remaining 20 s wait before the fallback kicks in (socket.io's default connect timeout) is out of scope here.

Fixes #5802

### Progress

- [x] Pytests not feasible.
- [x] Test on iOS 26.

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -364,19 +364,6 @@ function ack() {
   window.ackedMessageId = window.nextMessageId;
 }
 
-function reloadPage(reason) {
-  const reloadCount = Number(sessionStorage.__nicegui_reload_count || 0);
-  if (reloadCount >= 3) {
-    console.log("not reloading anymore because " + reason);
-    document.getElementById("popup").ariaHidden = true;
-    document.getElementById("connection_failed_popup").ariaHidden = false;
-    return;
-  }
-  sessionStorage.__nicegui_reload_count = reloadCount + 1;
-  console.log("reloading because " + reason);
-  window.location.reload();
-}
-
 function createRandomUUID() {
   try {
     return crypto.randomUUID();
@@ -463,13 +450,11 @@ function createApp(elements, options) {
 
           function finishHandshake(ok) {
             if (!ok) {
-              reloadPage("handshake failed for clientId " + window.clientId);
-              return;
+              console.log("reloading because handshake failed for clientId " + window.clientId);
+              window.location.reload();
             }
-            sessionStorage.removeItem("__nicegui_reload_count");
             window.did_handshake = true;
             document.getElementById("popup").ariaHidden = true;
-            document.getElementById("connection_failed_popup").ariaHidden = true;
           }
 
           if (options.query.implicit_handshake) {
@@ -488,9 +473,11 @@ function createApp(elements, options) {
               window.socket.io.opts.transports = ["polling", "websocket"];
               return;
             }
-            reloadPage("connection timed out"); // see https://github.com/zauberzeug/nicegui/issues/198
+            console.log("reloading because connection timed out");
+            window.location.reload(); // see https://github.com/zauberzeug/nicegui/issues/198
           } else if (err.message == "Implicit handshake failed") {
-            reloadPage("implicit handshake failed for clientId " + window.clientId);
+            console.log("reloading because implicit handshake failed for clientId " + window.clientId);
+            window.location.reload();
           }
         },
         try_reconnect: async () => {
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -364,6 +364,19 @@ function ack() {
   window.ackedMessageId = window.nextMessageId;
 }
 
+function reloadPage(reason) {
+  const reloadCount = Number(sessionStorage.__nicegui_reload_count || 0);
+  if (reloadCount >= 3) {
+    console.log("not reloading anymore because " + reason);
+    document.getElementById("popup").ariaHidden = true;
+    document.getElementById("connection_failed_popup").ariaHidden = false;
+    return;
+  }
+  sessionStorage.__nicegui_reload_count = reloadCount + 1;
+  console.log("reloading because " + reason);
+  window.location.reload();
+}
+
 function createRandomUUID() {
   try {
     return crypto.randomUUID();
@@ -409,15 +422,18 @@ function createApp(elements, options) {
       options.query.document_id = window.documentId;
       options.query.tab_id = TAB_ID;
       options.query.old_tab_id = OLD_TAB_ID;
+      const transportNames =
+        "prerendering" in document && document.prerendering === true
+          ? ["polling", ...options.transports]
+          : [...options.transports];
       window.socket = io(url, {
         path: `${options.prefix}/_nicegui_ws/socket.io`,
         query: options.query,
         extraHeaders: options.extraHeaders,
-        transports:
-          "prerendering" in document && document.prerendering === true
-            ? ["polling", ...options.transports]
-            : options.transports,
+        transports: [...transportNames], // copy because socket.io replaces the names with transport classes in place
+        tryAllTransports: true, // if the WebSocket fails with an error, try polling within the same attempt (see #5802)
       });
+      let usePollingFallback = transportNames[0] === "websocket" && transportNames.includes("polling");
       window.socket.io.on("reconnect_attempt", () => {
         // keep the otherwise-frozen reconnect query in sync to avoid triggering a reload (see #6019)
         options.query.next_message_id = window.nextMessageId;
@@ -447,11 +463,13 @@ function createApp(elements, options) {
 
           function finishHandshake(ok) {
             if (!ok) {
-              console.log("reloading because handshake failed for clientId " + window.clientId);
-              window.location.reload();
+              reloadPage("handshake failed for clientId " + window.clientId);
+              return;
             }
+            sessionStorage.removeItem("__nicegui_reload_count");
             window.did_handshake = true;
             document.getElementById("popup").ariaHidden = true;
+            document.getElementById("connection_failed_popup").ariaHidden = true;
           }
 
           if (options.query.implicit_handshake) {
@@ -462,11 +480,17 @@ function createApp(elements, options) {
         },
         connect_error: (err) => {
           if (err.message == "timeout") {
-            console.log("reloading because connection timed out");
-            window.location.reload(); // see https://github.com/zauberzeug/nicegui/issues/198
+            if (usePollingFallback) {
+              // the WebSocket might be blocked (e.g. by a proxy or https://bugs.webkit.org/show_bug.cgi?id=302823),
+              // so we retry with polling first and let socket.io reconnect automatically (see #5802)
+              usePollingFallback = false;
+              console.log("connection timed out, falling back to polling");
+              window.socket.io.opts.transports = ["polling", "websocket"];
+              return;
+            }
+            reloadPage("connection timed out"); // see https://github.com/zauberzeug/nicegui/issues/198
           } else if (err.message == "Implicit handshake failed") {
-            console.log("reloading because implicit handshake failed for clientId " + window.clientId);
-            window.location.reload();
+            reloadPage("implicit handshake failed for clientId " + window.clientId);
           }
         },
         try_reconnect: async () => {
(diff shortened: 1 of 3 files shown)


nicegui/static/nicegui.js
```diff
@@ -409,15 +422,18 @@ function createApp(elements, options) {
...
       window.socket = io(url, {
         path: `${options.prefix}/_nicegui_ws/socket.io`,
         query: options.query,
         extraHeaders: options.extraHeaders,
-        transports:
-          "prerendering" in document && document.prerendering === true
-            ? ["polling", ...options.transports]
-            : options.transports,
+        transports: [...transportNames], // copy because socket.io replaces the names with transport classes in place
+        tryAllTransports: true, // if the WebSocket fails with an error, try polling within the same attempt (see #5802)
```
evnchn, line 421: Hmm I wonder if the value of the PR boils down to this line...

Again, no iOS 26 device to test though

falkoschindler: It might. But Fable really wanted to limit the number of retries within the same PR.
Maybe we play it safe and revert everything but these lines? The question is: Is it a realistic scenario that trying to connect 12 times makes sense, or is it ok to give up after 3 times?

falkoschindler:
I'll test on iOS 26 this evening when I have access to an iPhone.

nicegui/static/nicegui.js
falkoschindler: An independent Fable session agreed to remove the reload cap:

---

Almost — but this line alone doesn't fix #5802: the iOS 26 WebSocket _hangs_ without ever erroring, so `tryAllTransports` never triggers. The 20 s connect timeout fires and we'd reload-loop exactly like before. `tryAllTransports` only covers WebSockets that fail _fast_ (e.g. a proxy rejecting the upgrade). The minimal fix for the issue is these two lines **plus** the timeout→polling fallback in `connect_error`.

What _is_ separable is the reload cap, and after thinking it through again I'd drop it from this PR:

- With the polling fallback in place, the timeout-reload only triggers when the page loads fine over HTTP but _both_ transports time out — a narrow residual case. Each loop iteration even requires a successful page load: if the server is actually down, the reload lands on a browser error page and the loop ends by itself.
- More importantly, retry-forever is self-healing for unattended deployments (wall dashboards, kiosks, robot UIs): when the backend comes back, the screen recovers. With the cap, it's stuck on "please reload the page" with nobody there to click it — even after the server is healthy again.

So I'll trim the PR to the transport fallback only: these two lines plus the timeout branch (which returns instead of reloading on the first timeout, and falls back to the old reload behavior after that). If we ever want a cap, the right shape is probably "stop reloading, keep reconnecting, and reload automatically once a connection succeeds again" — worth its own PR.

falkoschindler:
Two review findings I'll look into after testing the current implementation on a real device:

1. **Comment slightly overstates the in-place mutation** (minor) — As far as I can tell from the bundled client, socket.io doesn't mutate the transports _array_: the engine wrapper reassigns the `transports` _property_ on the opts object (`opts.transports = names.map(...)`), and since the manager keeps the caller's opts object, that's why `window.socket.io.opts.transports` holds classes after connecting. The original `transportNames` array itself is never touched, so strictly neither the early snapshot nor the `[...transportNames]` copy is required. Both are harmless and self-documenting — keep them if you like — but consider rewording the two comments to "socket.io replaces the `transports` option on the opts object with transport classes", so a future reader doesn't conclude the array is mutated. (If you observed actual array mutation in testing, I'd be curious where — I couldn't find a code path for it in 4.8.1.)

2. **No user feedback during the extended initial wait** (minor) — The `#popup` overlay starts hidden and only appears on `disconnect`, so on first load the iOS-26 user now stares at a frozen page for up to ~40 s (20 s timeout + polling retry) instead of ~20 s, with no indication anything is happening. Showing the connection-lost popup when entering the fallback branch would be a one-liner. Take-or-leave — arguably part of the "20 s timeout is out of scope" follow-up.

nicegui/static/nicegui.js
```diff
@@ -462,6 +465,14 @@ function createApp(elements, options) {
         },
         connect_error: (err) => {
           if (err.message == "timeout") {
+            if (usePollingFallback) {
+              // the WebSocket might be blocked (e.g. by a proxy or https://bugs.webkit.org/show_bug.cgi?id=302823),
+              // so we retry with polling first and let socket.io reconnect automatically (see #5802)
+              usePollingFallback = false;
+              console.log("connection timed out, falling back to polling");
+              window.socket.io.opts.transports = ["polling", "websocket"];
```
evnchn, line 473: Is there any reason for expanding to polling only on error, instead of simply letting `['websocket', 'polling']` and native `tryAllTransports` to do its thing? 

Then it's a 2 line fix?

evnchn (commented):
I still don't quite get why the new machinery.

nicegui/static/nicegui.js
falkoschindler: Fable speaking:

I wondered the same, so I tested exactly that. The short answer: `tryAllTransports` is *error-driven*, but the iOS 26 WebSocket doesn't error — it hangs. The two mechanisms cover different failure modes, and the 2-line version doesn't fix #5802.

In the bundled client (4.8.1), `tryAllTransports` only shifts to the next transport when the current one fires a *transport error* while the engine is still opening. The 20 s connect timeout is a different code path: it lives on the socket.io *manager*, which on expiry emits `connect_error("timeout")` and closes the engine. The engine never saw an error, so `tryAllTransports` never engages — and the next reconnect attempt creates a fresh engine with the same websocket-first order, which hangs again. Without the `connect_error` branch, nothing ever changes the transport order.

I verified this empirically with a WebSocket stub that simulates the WebKit hang (never connects, never errors):

```python
ui.add_head_html('''<script>
  window.WebSocket = class {
    constructor() {}
    send() {}
    close() {}
    addEventListener() {}
    removeEventListener() {}
  };
</script>''')
```

- **2-line variant** (`tryAllTransports` + transports copy only, `connect_error` branch removed): reload loop, exactly like before the PR — 6 page loads in ~85 s, five consecutive `"reloading because connection timed out"` in the console, engine stuck on `websocket`, never connected.
- **Full fix**: single page load, `"connection timed out, falling back to polling"` after 20 s, then connected via polling with a completed handshake; button clicks round-trip fine.

So the split is:

| WebSocket failure mode | covered by |
|---|---|
| errors fast (e.g. proxy rejects the upgrade) | `tryAllTransports` |
| hangs silently (iOS 26, WebKit bug 302823) | the `connect_error` timeout branch |

The timeout branch is the part that actually fixes the issue; `tryAllTransports` is the bonus that makes the fast-error case seamless.

evnchn (approved):
Well it's now `+15 -4` and you say it works, and otherwise the happy path for all other NiceGUI user is not affected, so maybe we can move forward with it.

falkoschindler:
Unfortunately I can't reproduce the problem on iOS 26.5. Maybe it's just flaky, or maybe it got fixed since 26.3.
Anyway, this PR should improve things nonetheless and it's all we can do. Let's merge!

PR #6104, opened 2026-06-11
Preserve client-originated values in element updates

### LINKED ISSUE #5185: Character loss in ui.input validation with high network latency

falkoschindler:
### Motivation

When typing into a `ui.input` with a `validation` function on a high-latency connection, characters get lost (#5185):
every keystroke triggers a value change event, the server updates the error message, and the resulting element update carries the _entire_ props dict — including a `value` that is already outdated by the time it reaches the client.
The client then resets the input field to the stale value, discarding everything typed in the meantime.

The same applies to all `LOOPBACK = False` value elements (`ui.input`, `ui.number`, `ui.color_input`, `ui.date_input`, `ui.time_input`, `ui.editor`) and to any server-side prop/style mutation while the user is typing, not just validation.

### Implementation

Implements the idea proposed in https://github.com/zauberzeug/nicegui/issues/5185#issuecomment-3351899333 to preserve the current value when an update reaches the client.
Instead of replacing the value with an in-band sentinel (which a colliding string value could break), update messages carry a separate `preserved_props` key listing the props the client should keep:

**Python:**

- `ValueElement` remembers the client's last raw emit in `_client_value` (for `LOOPBACK = False` elements only), reset on any server-originated value change.
- `ValueElement._to_dict()` adds `'preserved_props': [VALUE_PROP]` if and only if the outgoing value still equals that emit — i.e. when the update would send back exactly what the client sent.
  Any divergence — a transformation like `ui.number`'s `format`, a server-side `.value` write, or a raw `props()` mutation — makes the comparison fail, so changed values always reach the client.
  In practice the key only appears in update messages: full page builds serialize freshly created elements whose value can't have come from the client yet (and the client would simply ignore the key when rendering).

**JavaScript:**

- The `update` message handler replaces each prop listed in `preserved_props` with the element's current client-side prop value, so possibly newer user input survives the update.
  Since that stored value equals the previous prop value, the component is never patched and keeps its internal (possibly newer) state.
  If the client has no current value for that prop, the server's value (still included in the props dict) is used as a fallback.

Full page builds and reconnections always receive the authoritative value in the props dict, with or without the `preserved_props` hint.

**Tests:**

The new screen test reproduces the character loss deterministically by throttling the value change event listener (verified: it fails on main and passes with this fix).
The other two behavioral directions are covered by existing screen tests:
values that the server transforms on the way (e.g. `ui.number` with `format`) are deliberately _not_ preserved, so sanitized/formatted values still reach the client (`test_apply_format_on_blur` would fail otherwise),
and server-originated value changes still override user input (`test_update_input`).
A third screen test covers writing the value via raw `props()` after the user typed, which must reach the client like on main (mutation-verified: it fails when preservation is decided at event time instead of serialization time).

### Manual reproduction

The bug is surprisingly hard to reproduce by hand:

- On localhost the round trip takes a few milliseconds, so updates never carry a meaningfully stale value — no human can type a character inside that window.
- Browser devtools network throttling doesn't help: Chromium doesn't apply added latency to frames on an established WebSocket connection.
- Naively delaying the outbox doesn't help either: `_to_dict()` holds _live_ references to the props and serialization happens at send time, so a delayed message still carries the then-current value.

What works is snapshotting the payload before delaying, which mimics serializing onto a slow wire:

```python
import asyncio
import copy

from nicegui import ui
from nicegui.outbox import Outbox

ui.input(validation=lambda x: f'Still {1000 - len(x)} characters missing')

original_emit = Outbox._emit


async def slow_emit(self, message) -> None:
    client_id, message_type, data = message
    data = copy.deepcopy(data)  # snapshot now — like serializing onto a slow wire
    await asyncio.sleep(0.5)  # simulate a slow server→client connection
    await original_emit(self, (client_id, message_type, data))

Outbox._emit = slow_emit

ui.run()
```

Type continuously for a few seconds — longer than the simulated latency; short bursts self-heal because the last update carries the final value.
On main, the field jumps back mid-typing and characters get permanently lost; with this fix, typing survives and only the error message lags behind.

### Progress

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -228,10 +228,6 @@ class Element(Visibility):
             if slot != self.default_slot
         }
 
-    def _to_update_dict(self) -> dict[str, Any]:
-        """Dictionary representation for update messages (in contrast to ``_to_dict`` for full page builds)."""
-        return self._to_dict()
-
     def _to_dict(self) -> dict[str, Any]:
         return {
             'tag': self.tag,
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -228,6 +228,10 @@ class Element(Visibility):
             if slot != self.default_slot
         }
 
+    def _to_update_dict(self) -> dict[str, Any]:
+        """Dictionary representation for update messages (in contrast to ``_to_dict`` for full page builds)."""
+        return self._to_dict()
+
     def _to_dict(self) -> dict[str, Any]:
         return {
             'tag': self.tag,
--- a/nicegui/elements/input.js
+++ b/nicegui/elements/input.js
@@ -33,7 +33,6 @@ export default {
   },
   watch: {
     value(newValue) {
-      if (newValue === this.inputValue) return; // e.g. the locally mirrored value of a "preserved value" update
       this.emitting = false;
       this.inputValue = newValue;
       this.$nextTick(() => (this.emitting = true));
--- a/nicegui/elements/input.js
+++ b/nicegui/elements/input.js
@@ -33,6 +33,7 @@ export default {
   },
   watch: {
     value(newValue) {
+      if (newValue === this.inputValue) return; // e.g. the locally mirrored value of a "preserved value" update
       this.emitting = false;
       this.inputValue = newValue;
       this.$nextTick(() => (this.emitting = true));
--- a/nicegui/elements/mixins/value_element.py
+++ b/nicegui/elements/mixins/value_element.py
@@ -34,6 +34,7 @@ class ValueElement(Element, Generic[ValueT]):
                  ) -> None:
         super().__init__(**kwargs)
         self._send_update_on_value_change = True
+        self._value_from_client = False
         self.set_value(value)
         self._props[self.VALUE_PROP] = self._value_to_model_value(value)
         self._props['loopback'] = self.LOOPBACK
@@ -46,6 +47,10 @@ class ValueElement(Element, Generic[ValueT]):
             self._send_update_on_value_change = self.LOOPBACK is True
             self.set_value(self._event_args_to_value(e))
             self._send_update_on_value_change = True
+            if self.LOOPBACK is False:
+                # the client already has this value; future updates must not overwrite possibly newer user input,
+                # unless the value has been transformed on the way and needs to be sent back
+                self._value_from_client = self._props.get(self.VALUE_PROP) == e.args
         self.on(f'update:{self.VALUE_PROP}', handle_change, [None], throttle=throttle)
 
     def on_value_change(self, callback: Handler[ValueChangeEventArguments[ValueT]]) -> Self:
@@ -134,12 +139,20 @@ class ValueElement(Element, Generic[ValueT]):
         with self._props.suspend_updates():
             self._props[self.VALUE_PROP] = self._value_to_model_value(value)
         if self._send_update_on_value_change:
+            self._value_from_client = False
             self.update()
         args = ValueChangeEventArguments(sender=self, client=self.client,
                                          value=self._value_to_event_value(value),
                                          previous_value=self._value_to_event_value(previous_value))
         for handler in self._change_handlers:
             handle_event(handler, args)
+
+    def _to_update_dict(self) -> dict[str, Any]:
+        dict_ = super()._to_update_dict()
+        if self.LOOPBACK is False and self._value_from_client:
+            # tell the client to keep its current (possibly newer) value (see "preserved_props" in nicegui.js)
+            dict_['preserved_props'] = [self.VALUE_PROP]
+        return dict_
 
     def _event_args_to_value(self, e: GenericEventArguments) -> ValueT:
         return e.args
--- a/nicegui/elements/mixins/value_element.py
+++ b/nicegui/elements/mixins/value_element.py
@@ -47,10 +47,7 @@ class ValueElement(Element, Generic[ValueT]):
             self._send_update_on_value_change = self.LOOPBACK is True
             self.set_value(self._event_args_to_value(e))
             self._send_update_on_value_change = True
-            if self.LOOPBACK is False:
-                # the client already has this value; future updates must not overwrite possibly newer user input,
-                # unless the value has been transformed on the way and needs to be sent back
-                self._value_from_client = self._props.get(self.VALUE_PROP) == e.args
+            self._value_from_client = self.LOOPBACK is False and self._props.get(self.VALUE_PROP) == e.args
         self.on(f'update:{self.VALUE_PROP}', handle_change, [None], throttle=throttle)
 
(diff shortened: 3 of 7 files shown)


evnchn:

**TL;DR:** Reviewed and verified in depth — the fix is sound and we're ready to approve. One behavioral edge surfaced during verification that we'd like your take on first: **while `_preserve_value` is set (i.e. after any user keystroke), writing the value through the raw props path — `inp.props('value=X')` or `_props` mutation + `update()` — is now silently masked by the client** until the next `set_value()`/`.value` write clears the flag. On `main` such writes take effect immediately. Setting a value this way is an anti-pattern (the right way is the `.value` setter), so this may well be acceptable — but the divergence is silent, persistent, and stateful (depends on whether the user ever typed), so we'd rather it be a conscious call. Three options we see: **leave it** (anti-pattern, not worth code), **cover it** (clear the flag when `props()` touches `VALUE_PROP`), or **warn** (`log.warning` when `.props()` writes the value prop of a `ValueElement`). Which do you prefer?

<details>
<summary><b>What we verified (all empirical, Chrome Screen tests on PR head 51dfbc05 vs main 240db43f)</b></summary>

- The flag semantics are conservative in every uncertain case: set only when the stored value is a pure echo of the client emit (so `ui.number`-style transformations still propagate — `test_apply_format_on_blur` passes), cleared on any server-originated value change before `update()` (`test_update_input` passes). 28/28 tests in `test_input.py` + `test_number.py` green on the branch.
- The new lagging-events test **fails on main** — it genuinely captures the #5185 regression.
- Version skew degrades gracefully both ways: old client ignores the unknown `preserved_props` key (falls back to current behavior); new client with old server is a no-op (`?? []`).
- The "never patched" mechanism is correct: the client substitutes its previously-rendered prop value, so the component is never patched and its internal (possibly newer) state survives.

</details>

<details>
<summary><b>The raw-props edge (empirical matrix)</b></summary>

Scenario: user types `'typed'`, then a button handler force-writes `'SERVER'`, then we read the rendered input value.

| write method | main | this PR |
|---|---|---|
| `inp._props['value'] = 'SERVER'; inp.update()` | `'SERVER'` — lands | `'typed'` — masked |
| `inp.props('value=SERVER')` | `'SERVER'` — lands | `'typed'` — masked |

The server-side props dict *does* change, so server state and client display silently disagree until a supported-path write happens. There is no timeout on the flag — one keystroke arms it indefinitely.

To be fair to the PR: on `main`, that same raw write mid-typing would clobber the user's input — which is exactly the #5185 failure shape this PR fixes. So the masking is arguably *more* correct for the typing scenario; the question is only about deliberate server-side raw writes afterwards.

</details>

falkoschindler:
Thanks for the deep verification — great catch, and exactly the kind of edge worth a conscious decision.

We went with "cover it", but at a different layer than hooking `props()`: a direct `_props['value'] = ...` mutation would bypass such a hook, so any decision made at event time stays blind to later writes. Instead, `ValueElement` now remembers the client's last raw emit (`_client_value`) and `_to_dict()` emits `preserved_props` only when the outgoing value still equals it — *preserve iff we'd send back exactly what the client sent*. That single comparison at serialization time covers your whole matrix (both rows now land), removes the statefulness you called out (no flag that stays armed — any divergence disarms preservation by construction), and even simplifies the arming line, since the transformation check is subsumed by the same comparison.

Implemented in 46bf02abf, covered by a new screen test (`test_setting_value_via_props_overrides_user_input`) that exercises the exact scenario from your matrix via a button handler. Mutation-verified: it fails on the previous head 51dfbc05.

evnchn (approved):

**TL;DR: Approving.** 

I dug into one residual edge — a custom element that subclasses a `LOOPBACK = False` value element and overrides `_event_args_to_value` / `_value_to_model_value` so the serialized model value no longer equals the client's raw emit. For such an element the `_to_dict()` equality never holds, so `preserved_props` is never emitted and #5185 would still bite *that custom element*. 

I think this is a deliberate **wontfix**:

1. it requires mutating `_`-prefixed internals that are explicitly not public contract — here be dragons; 
2. it degrades gracefully — the authoritative value still reaches the client exactly as it does on `main` today, so it's a *missing new protection*, never data loss or corruption. All six in-tree `LOOPBACK = False` elements are clear, so nothing ships broken.

<details>
<summary><b>In-tree audit — all six built-ins are covered</b></summary>

Checked which `LOOPBACK = False` elements actually override the value transforms (so `_props[VALUE_PROP] == _client_value` could break):

| Element | Overrides transform? | Preservation engages |
|---|---|---|
| `ui.input` | no — base identity | ✅ |
| `ui.color_input` | no — base identity | ✅ |
| `ui.editor` | no — base identity | ✅ |
| `ui.date_input` | no¹ | ✅ |
| `ui.time_input` | no¹ | ✅ |
| `ui.number` | **yes** | ⚠️ by-design only² |

¹ `date_input`/`time_input` have `_picker_to_input_value` / `_input_to_picker_value`, but those sync the attached picker, not the text field's `model-value` — the typed field flows through base identity.

² `number._value_to_model_value`: `"5"`→`"5"` and `"5.5"`→`"5.5"` preserve fine; only normalizing inputs (`"05"`, `"5.50"`, or a `format`) diverge — and there the server value is *genuinely* different and **should** reach the client. That's the same direction `test_apply_format_on_blur` already guards. Correct degradation, not a hole.
</details>

<details>
<summary><b>Why the custom-author case is wontfix, not a latent bug</b></summary>

The hypothetical: someone writes `class Foo(ValueElement)` with `LOOPBACK = False` and a transforming `_event_args_to_value`/`_value_to_model_value` pair (à la `ui.number`, but where they *want* round-trip preservation). For them `preserved_props` never fires and they'd see the original #5185 character-loss on a high-latency link.

- **Surface is internal.** It only triggers by overriding `_`-prefixed methods — not part of the supported public API. An author down at that layer already owns the value round-trip semantics.
- **Failure mode is benign.** Without the hint, the props dict still carries the authoritative value, so the client renders correctly — identical to current `main`. The worst case is "no improvement," never a regression or corruption.
- **A general fix is costly for the payoff.** Catching it would mean comparing pre-/post-transform on every emit or exposing a preservation hook — added surface and per-update cost to protect a path that requires opting into internals. Not worth it.

If a real custom element ever needs this, the clean lever is a small overridable hook rather than widening the equality heuristic — but that's demand-driven, not now.
</details>

Approving on that basis — nice, clean fix, and the serialization-time equality is a tidy way to subsume the `props()` edge. 🚀

Issue #4970, opened 2025-07-20
"No response returned" when too many video sources are loaded in native browser


MaidScientistIzutsumiMarin:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

for _ in range(2):
    ui.video("test.mp4")

ui.run(native=True, reload=False)
```

### Description

Start the app.
A `RuntimeError` occurs with the message "No response returned."

### NiceGUI Version

2.21.1

### Python Version

Python 3.13.5

### Browser

Other

### Operating System

Windows

### Additional Context

This does not happen if I launch the program in Firefox. I've attached `test.mp4` for reproducibility, but this happens with any videos if enough are loaded. It is also related to the length of the videos, where shorter videos need more to be embedded before the error occurs. The number of videos can be increased if this isn't causing an error. The error doesn't seem to cause any issues with the actual functionality of the program besides lagging.

The error also occurs when the event loop is set to `WindowsSelectorEventLoopPolicy`.

https://github.com/user-attachments/assets/fc848cbb-ccdd-4bf3-afb7-cc6545742491

falkoschindler:
Hi @MaidScientistIzutsumiMarin,

Am I right to assume that the problem already occurs with 2x test.mp4?
I just tested with 100x test.mp4 and didn't get any RuntimeError.

Can anyone else reproduce the problem?

MaidScientistIzutsumiMarin:
> Hi [@MaidScientistIzutsumiMarin](https://github.com/MaidScientistIzutsumiMarin),
> 
> Am I right to assume that the problem already occurs with 2x test.mp4? I just tested with 100x test.mp4 and didn't get any RuntimeError.
> 
> Can anyone else reproduce the problem?

For me it happens with just 2 of them. Also, I've included the console output. If this is too long, I can move it to a Pastebin or something if there is a standard way to include long errors here.
```python
NiceGUI ready to go on http://127.0.0.1:8000
No response returned.
  + Exception Group Traceback (most recent call last):
  |   File "C:\Users\Cookie\AppData\Local\Programs\Python\Python313\Lib\site-packages\starlette\_utils.py", line 77, in collapse_excgroups
  |     yield
[...]

PR #6105, opened 2026-06-12
Bound the connection drain when the native window closes (#5443)

### LINKED ISSUE #5443: ConnectionResetError from loading too many video elements in native browser

evnchn:
*Drafted by Claude Code on Evan's behalf.*

### Motivation

Fixes #5443.

Implements the direction @falkoschindler proposed in https://github.com/zauberzeug/nicegui/issues/5443#issuecomment-4509935595: when the native window closes, the user is done — exit promptly without imposing anything on non-native users or truncating `app.on_shutdown` work. Supersedes #5706 (now draft), which set `timeout_graceful_shutdown=10` globally.

### One correction to the proposal's trace, which changes the implementation

The sketch assumed lifespan shutdown runs *before* the connection drain, so "wait for lifespan, then `os._exit(0)`" would be safe. In uvicorn (verified on 0.22.0, our floor, through 0.40.0, current), the order in `Server.shutdown()` is the reverse: **drain first, lifespan last**. In the hang scenario the ghost connection blocks the drain forever, so lifespan never starts — waiting on it would deadlock exactly like the original bug, and `on_shutdown` callbacks would never run.

The good news: the intent survives with uvicorn's own mechanism, no `os._exit` needed. `timeout_graceful_shutdown` is read live at shutdown time and bounds **only the drain** — on timeout, uvicorn cancels the drain and still runs lifespan shutdown. So the existing `monitor_shutdown_event` thread (which fires precisely at window close, native mode only) just sets `timeout_graceful_shutdown = 1` before `should_exit`.

### Behavior (empirical, macOS, app with a 2-second `on_shutdown` handler)

| | total wall | `on_shutdown` completed? |
|---|---|---|
| main (baseline) | 6 s | ✓ |
| `timeout = 0` (hard cancel) | **13 s** — abrupt task cancellation costs ~7 s of teardown | ✓ |
| **`timeout = 1` (this PR)** | **6 s** — drain finishes naturally in <1 s once the window is gone; the cancel path never executes | ✓ |

So on healthy platforms this is empirically a no-op; the Windows ghost-connection case gets a bounded 1-second escape with `on_shutdown` still running to completion. Non-native mode is untouched (the monitor thread only exists in native mode).

**Pending:** Windows verification against the #4970/#5443 repro — @evnchn will run this across his Windows machines before this leaves draft. @falkoschindler: does this match your intent?

### Progress

- [ ] The implementation is complete. (Draft — pending Windows fleet verification and @falkoschindler's confirmation of intent.)
- [ ] Pytests have been added. (The hang is Windows-native-GUI-bound; empirical manual protocol documented above. Open to suggestions for a CI-able test.)

### DIFF
--- a/nicegui/server.py
+++ b/nicegui/server.py
@@ -38,10 +38,8 @@ class Server(uvicorn.Server):
             if (event := self.config.shutdown_event) is not None:
                 def monitor_shutdown_event() -> None:
                     event.wait()
-                    # NOTE: The window is closed, so there is no client left to wait for.
-                    # Bound the connection drain, which otherwise hangs forever on stale connection bookkeeping
-                    # on Windows (#5443); app.on_shutdown callbacks still run to completion afterwards,
-                    # because uvicorn's lifespan shutdown follows the drain and is not affected by this timeout.
+                    # The window is closed, so bound the connection drain, which can hang forever on Windows (#5443).
+                    # Lifespan shutdown (with app.on_shutdown callbacks) follows the drain, unaffected by this timeout.
                     self.config.timeout_graceful_shutdown = 1
                     self.should_exit = True
                 threading.Thread(target=monitor_shutdown_event, daemon=True).start()
--- a/nicegui/server.py
+++ b/nicegui/server.py
@@ -38,6 +38,11 @@ class Server(uvicorn.Server):
             if (event := self.config.shutdown_event) is not None:
                 def monitor_shutdown_event() -> None:
                     event.wait()
+                    # NOTE: The window is closed, so there is no client left to wait for.
+                    # Bound the connection drain, which otherwise hangs forever on stale connection bookkeeping
+                    # on Windows (#5443); app.on_shutdown callbacks still run to completion afterwards,
+                    # because uvicorn's lifespan shutdown follows the drain and is not affected by this timeout.
+                    self.config.timeout_graceful_shutdown = 1
                     self.should_exit = True
                 threading.Thread(target=monitor_shutdown_event, daemon=True).start()
 


falkoschindler (approved):
Nice! As far as I can tell, this is a great solution. Approving.
Does this close #5443?

evnchn:

Yes — added `Fixes #5443` to the PR body. The reported pain there (closing the native window leaves the terminal hanging) is exactly what this bounds; if the underlying question of *why* Windows connections reset under video load deserves its own investigation someday, a fresh narrower issue will serve that better than keeping #5443 open.

Thanks for the quick approval! Evan will run the Windows verification tonight (patched vs. baseline against the #4970 repro on his machines) and un-draft once the numbers are in.

evnchn:

**TL;DR: closing this — it (a) deviated from your proposed direction and (b) doesn't actually fix #5443 (verified on Windows). A replacement faithful to your `os._exit` idea will follow, verified on the repro first.**

<details><summary>Details</summary>

**Deviation.** Your #5443 direction was a hard exit after lifespan shutdown — `os._exit(0)` once `on_shutdown` callbacks have run. This PR instead sets `timeout_graceful_shutdown = 1` to bound the drain — a different mechanism than the one you sketched, and I should have flagged that here rather than letting it ride to approval.

**Doesn't fix the repro.** `monitor_shutdown_event` only runs when `config.shutdown_event` exists, which NiceGUI creates **only when `reload=True`**. The #5443 / #4970 repro — and packaged native apps generally — run `reload=False`, so the patched line never executes. Windows trials (Win11 + WebView2, 2× MP4, #4970 recipe, 20/arm):

| arm | hangs |
|---|---|
| baseline (`main`) | 8/20 |
| this PR | 10/20 (no change) |
| drain bound *armed* for `reload=False` | 0/20 |

The mechanism is sound when it runs — it just never runs in the mode that hangs.

Closing in favor of a verified replacement.

</details>

PR #6106, opened 2026-06-12
Fix native window close not shutting down the app after hot reload

### LINKED ISSUE #5845: Closing the window does not shutdown the app after hot reloading
### First Check

### Example Code

```python
from nicegui import app, ui

def app_startup():
    print('app startup')

def app_shutdown():
    print('app shutdown')

app.on_startup(app_startup)
app.on_shutdown(app_shutdown)

@ui.page('/')
def index():
    ui.button('destroy window', on_click=lambda: app.native.main_window.destroy())
    ui.button('nicegui shutdown', on_click=lambda: app.shutdown())
    
ui.run(
    title="Some app",
    port=8502,
    native=True,
    reload=True,
    window_size=(1920, 1200),
    storage_secret='my secret'
)
```

### Description

Very similar to #2107, but seems to be a different problem (or maybe the original fix is not enough), as the original problem is described to be windows only, but I'm on macOS now.

Specifically, the original issue itself is fixed: open the app, click either X or destroy window or nicegui shutdown, all three actions leads to a proper shutdown of the app even with `reload=True`. However, when you actually *tries* to use hot reload, the entire thing is again broken.

Full reproduction steps:
1. Open the app
2. Edit the file to add something irrelevant (like a comment), triggering an auto reload
3. Press X or destroy window, the window closed but the app does not quit (the console is still in the app's process; in fact if you edit the file again the console will show a second hot reload message). Press nicegui shutdown still works properly.

Console output:
```
NiceGUI ready to go on http://127.0.0.1:8502
app startup
WARNING:  WatchFiles detected changes in 'test.py'. Reloading...
app shutdown
NiceGUI ready to go on http://127.0.0.1:8502
app startup
<Press Ctrl-C>app shutdown
```

### NiceGUI Version

3.8.0

### Python Version

3.14.3

### Browser

Other

### Operating System

macOS

### Additional Context

macOS Tahoe 26.3 (arm64) on an M2 Pro MacBook Pro

evnchn:
*Authored by Claude Code (Claude Fable 5) on behalf of @evnchn; analysis and fix empirically verified as described in the comment below.*

### Motivation

Fixes #5845: with `ui.run(native=True, reload=True)`, closing the native window stops shutting down the app as soon as one hot reload has happened. The window disappears, but the supervisor and the worker process linger forever (the issue's "console is still in the app's process" symptom, on macOS and in principle on every platform).

> [!NOTE]
> This PR does **not** implement the SIGINT workaround sketched in [the analysis comment on the issue](https://github.com/zauberzeug/nicegui/issues/5845#issuecomment-3981993953). Empirical verification refuted that comment's root-cause claim — the webview process exits just fine after window close. This PR fixes the actual, instrumentation-proven root cause below.

The real root cause:

1. The worker's `monitor_shutdown_event` thread blocks in `shutdown_event.wait()`, which registers it as a sleeper on the `multiprocessing.Event`'s internal `Condition`.
2. A hot reload makes uvicorn's `ChangeReload` kill that worker mid-wait. The dead sleeper's registration on the shared semaphores is never cleaned up.
3. When the window is later closed, `check_shutdown()` in the supervisor calls `shutdown_event.set()`, and `Condition.notify_all()` blocks forever in `_woken_count.acquire()`, waiting for an acknowledgment the dead worker can never send.
4. Blocked `set()` means: no `should_exit` in the new worker, no `_thread.interrupt_main()` in the supervisor — nothing shuts down. (`app.shutdown()` still works because it bypasses the event and signals the supervisor directly.)

### Implementation

The worker's monitor thread now polls `event.is_set()` at 0.1 s instead of blocking in `event.wait()`. `is_set()` never registers as a Condition sleeper, so a worker killed by the reloader leaves no poisonous state behind and the supervisor's `set()` returns immediately. The 0.1 s poll matches the existing style in `check_shutdown()`.

Notes:

- `event.wait(timeout)` in a loop would **not** fix this — a timed wait still registers as a sleeper for ~100% of wall time (confirmed against CPython's `multiprocessing/synchronize.py`).
- Residual race, for the record: `is_set()` holds the Condition's lock for a few microseconds per 100 ms poll, and POSIX semaphores are not robust against owner death. A worker killed in exactly that window would still poison the event, but the probability drops from ~100% (old code: parked as a sleeper essentially all the time) to roughly lock-held-time/poll-interval ≈ 10⁻⁵ per reload, which is the practical floor without abandoning `multiprocessing.Event` entirely.
- No automated test: a faithful regression test needs a real pywebview window, a real `ChangeReload` file-touch reload and a window close, which CI cannot do. Full empirical verification (repro harness, instrumented runs, before/after matrix on macOS) is documented in the comment below.

### Progress

- [x] Pytests are not necessary (not CI-able; empirical verification documented below).

### DIFF
--- a/nicegui/server.py
+++ b/nicegui/server.py
@@ -38,9 +38,8 @@ class Server(uvicorn.Server):
             native.response_queue = self.config.response_queue
             if (event := self.config.shutdown_event) is not None:
                 def monitor_shutdown_event() -> None:
-                    # Poll instead of `event.wait()`: a blocking wait would register this thread as a waiter
-                    # on the event's internal condition variable. When the auto-reloader kills the process
-                    # mid-wait, the stale registration makes a later `event.set()` block forever (#5845).
+                    # Poll instead of `event.wait()`: a blocking wait leaves a stale waiter if the reloader
+                    # kills the process mid-wait, making a later `event.set()` hang forever (#5845).
                     while not event.is_set():
                         time.sleep(0.1)
                     self.should_exit = True
--- a/nicegui/server.py
+++ b/nicegui/server.py
@@ -4,6 +4,7 @@ import multiprocessing
 import multiprocessing.synchronize
 import socket
 import threading
+import time
 from typing import Any
 
 import uvicorn
@@ -37,7 +38,11 @@ class Server(uvicorn.Server):
             native.response_queue = self.config.response_queue
             if (event := self.config.shutdown_event) is not None:
                 def monitor_shutdown_event() -> None:
-                    event.wait()
+                    # Poll instead of `event.wait()`: a blocking wait would register this thread as a waiter
+                    # on the event's internal condition variable. When the auto-reloader kills the process
+                    # mid-wait, the stale registration makes a later `event.set()` block forever (#5845).
+                    while not event.is_set():
+                        time.sleep(0.1)
                     self.should_exit = True
                 threading.Thread(target=monitor_shutdown_event, daemon=True).start()
 


evnchn:
*Posted by Claude Code (Claude Fable 5) on @evnchn's behalf — empirical verification record for this PR.*

**TL;DR:**

This PR fixes #5845 at its instrumentation-proven root cause:

1. **Proven:** the supervisor's `shutdown_event.set()` deadlocks inside `multiprocessing.Condition.notify_all()`, because the hot reload killed the previous worker while it was registered as a sleeper in `event.wait()`. Dead sleepers never acknowledge; `set()` waits forever — so closing the window can no longer stop anything.
2. **The fix:** poll `event.is_set()` instead of blocking in `event.wait()` (+6/−1 in `server.py`). Polling never registers a sleeper, so a killed worker leaves no poisonous state behind.
3. **Explicitly out of scope:** this PR does **not** implement the SIGINT workaround from [the earlier analysis comment on the issue](https://github.com/zauberzeug/nicegui/issues/5845#issuecomment-3981993953). That comment's root-cause claim ("`webview.start()` doesn't return on macOS after reload") was refuted empirically — details in the evidence blocks below, kept on record so nobody re-walks that dead end.

Ask: review the polling fix; the mechanism is platform-independent, so a quick confirmation on Windows/Linux would be welcome (verified on macOS only).

<details>
<summary><b>Test environment and harness (replayable)</b></summary>

- macOS 15.7.1 (arm64, M4), Python 3.12.11, pywebview 6.1, wsproto websockets
- Repo at upstream `main` = `5e30684`; fix commit = `63f5cc9`
- Real native window on a real display; hot reload triggered by actually appending a line to the watched file; window closed via `app.native.main_window.destroy()` (same `closed`-event path as pressing X, per the issue both behave identically)

MRE (`mre_5845.py`) — falko's MRE plus lifecycle prints and a flag-file-triggered destroy so the run is fully automated:

```python
from pathlib import Path
from nicegui import app, ui

FLAG = Path('/tmp/nicegui_5845_destroy')

app.on_startup(lambda: print('LIFECYCLE: app startup', flush=True))
app.on_shutdown(lambda: print('LIFECYCLE: app shutdown', flush=True))

@ui.page('/')
def index():
    ui.label('issue 5845 repro')

    def check_flag() -> None:
        if FLAG.exists():
            FLAG.unlink()
            print('DRIVER: destroying window', flush=True)
            app.native.main_window.destroy()

    ui.timer(0.2, check_flag)

ui.run(native=True, reload=True, port=8765, show_welcome_message=True)
```

Driver: launch → wait for "ready to go" → optionally append `# reload trigger` to the file N times (waiting for each reload's "ready to go") → touch the flag file → wait up to 15 s for the launched process to exit → snapshot `ps` process tree before/after. "Hang" below means the main process was still alive after 15 s and had to be SIGKILLed.

</details>

<details>
<summary><b>Verification matrix (every cell is an actual run, not reasoning)</b></summary>

| Scenario | unpatched `5e30684` | commented SIGINT fix | this PR `63f5cc9` |
|---|---|---|---|
| `reload=True`, close, **no** hot reload | ✅ clean exit (1/1) | not run | ✅ clean exit (3/3) |
| `reload=True`, **1 hot reload**, close | ❌ hangs (supervisor + worker) — **7/7 runs** | ✅ clean exit (1/1) | ✅ clean exit — **7/7 runs** |
| `reload=True`, **2 hot reloads**, close | not run (1-reload case already fails) | not run | ✅ clean exit (1/1) |
| `reload=False`, close | not run | not run | ✅ clean exit (1/1) |

Determinism in both directions is expected from the mechanism, not luck: the unpatched monitor thread spends ~100% of its wall time parked inside `event.wait()`, so the reload-kill always lands mid-wait and always leaves a stale sleeper; the patched monitor touches shared state for only microseconds per 100 ms poll.

With this PR, the instrumented run shows the **entire designed shutdown chain** completing — webview exit detected → `set()` returns → `interrupt_main()` → worker monitor wakes → graceful `app shutdown` (lifecycle handlers run). With the SIGINT fix, exit also happens, but via the signal shortcut while the Event stays poisoned (see refuted-hypothesis block).

</details>

<details>
<summary><b>Root cause, proven with instrumented prints</b></summary>

Debug prints were added at every step of the chain (`check_shutdown()` in `native_mode.py`, `monitor_shutdown_event()` in `server.py`). The hot-reload-then-close run on unpatched main produced:

```
DEBUG5845: worker monitor waiting on shutdown_event      ← worker #1
NiceGUI ready to go on http://127.0.0.1:8765
LIFECYCLE: app startup
WARNING:  WatchFiles detected changes in 'mre_5845.py'. Reloading...
LIFECYCLE: app shutdown                                  ← worker #1 killed mid-wait
DEBUG5845: worker monitor waiting on shutdown_event      ← worker #2
NiceGUI ready to go on http://127.0.0.1:8765
LIFECYCLE: app startup
DRIVER: destroying window
DEBUG5845: supervisor detected webview process exit      ← webview DID exit
DEBUG5845: supervisor calling shutdown_event.set()
                                                         ← "set() returned" NEVER prints
```

`set()` never returns. Mechanism (cross-checked against CPython `multiprocessing/synchronize.py`):

1. `Event.wait()` registers the calling thread as a sleeper on the Event's internal `Condition` (`_sleeping_count.release()` then park in `_wait_semaphore.acquire()`).
2. Uvicorn's `ChangeReload` terminates worker #1 mid-park. The sleeper registration on the shared POSIX semaphores survives the process.
3. `Event.set()` → `Condition.notify_all()` → for each registered sleeper, a **blocking** `_woken_count.acquire()` — waiting for an acknowledgment the dead worker can never send. `set()` blocks forever **while holding the Condition lock**, which also wedges worker #2's `wait()` (it can't re-acquire the lock), so neither `should_exit` nor `_thread.interrupt_main()` ever happen.

This also explains every symptom in the issue: first-run close works (no stale sleeper yet), any number of reloads breaks it, and `app.shutdown()` keeps working because it sends SIGINT to the supervisor directly, bypassing the event.

Why polling is the fix: `is_set()` takes the Condition lock for microseconds but **never registers as a sleeper** — there is no stale state for a killed worker to leave behind. A timed `event.wait(0.1)` loop would NOT fix it: a timed wait still registers as a sleeper for ~100% of wall time (confirmed in CPython source; the timeout-rebalancing in `notify` only credits sleepers whose `finally` ran).

</details>

<details>
<summary><b>Refuted hypothesis: "webview.start() doesn't return on macOS after reload"</b></summary>

The March comment proposed that the webview process outlives the window after a hot reload, leaving `check_shutdown()` stuck in `while process.is_alive()`. Process-tree snapshots disprove this. After hot reload + destroy on unpatched main:

```
before destroy:  86473 mre_5845.py (supervisor)
                 86537 spawn child  ← webview process (spawned first, by activate())
                 86580 spawn child  ← worker #2 (respawned by ChangeReload)
after destroy:   86473 still alive  ← BUG
                 86537 GONE         ← webview process exited fine
                 86580 still alive, reparented to PID 1 after cleanup, still holding port 8765
```

The orphan holding the server port was the **worker**, not the webview process. The instrumented run confirms it independently: `supervisor detected webview process exit` prints, i.e. `process.is_alive()` went false.

The SIGINT fix from that comment was also implemented and tested for completeness: it does fix the hang (clean exit, worker #2 shuts down gracefully via `ChangeReload`'s own teardown). But the supervisor exits through the signal shortcut while `check_shutdown()` is still deadlocked inside `set()` — the Event stays poisoned and the designed chain stays broken underneath. Working fix, wrong layer, wrong premise — hence not what this PR ships.

</details>

<details>
<summary><b>Residual risk + review provenance</b></summary>

- **Residual race (~10⁻⁵ per reload, disclosed in the PR body):** `is_set()` holds the Condition lock for a few microseconds per 100 ms poll; POSIX semaphores are not robust to owner death, so a worker killed exactly inside that window would still poison the event. Old code was parked as a registered sleeper essentially 100% of the time; that's the ~5-orders-of-magnitude improvement, and the practical floor without replacing `multiprocessing.Event` wholesale. If this ever resurfaces, the natural hardening is calling `set()` from a throwaway daemon thread in `check_shutdown()` — deliberately not done here to keep the change minimal.
- An independent reviewer pass (separate agent, fresh context) verified the deadlock mechanism against CPython source, confirmed the timed-wait alternative doesn't work, and rated the change ready; its one Important item (disclose the residual race) is incorporated above.
- Gates run locally: `pre-commit` on the changed file, `mypy ./nicegui` (clean), `pylint ./nicegui` (10.00/10). No pytest added — a faithful regression test needs a real window + real `ChangeReload` file-touch reload, which CI can't provide; this comment is the verification record instead.

</details>

---
**Changelog**
- 2026-06-13: Reframed the TL;DR to lead with this PR's own fix and state explicitly that it does not implement the issue-comment SIGINT proposal. Evidence blocks unchanged.
- 2026-06-13 (later): Added repeated-sampling counts to the verification matrix: 5 additional runs per side of the hot-reload scenario — unpatched 7/7 hang, patched 7/7 clean exit.

evnchn:
@falkoschindler I know comment is long but let's fix once the logic is hammered down

falkoschindler (dismissed):
Great solution, @evnchn!
I don't have a Linux or Windows machine at hand, but I confirm it works on macOS. 👍🏻

evnchn:

@falkoschindler — tightened the inline comment on the polling loop to match the codebase's terser comment style (3 lines → 2, same content + the #5845 reference); no code change. Your approval carried over and CI's green. OK for me to merge, or would you like to?

falkoschindler (approved):

PR #6110, opened 2026-06-14
Fix ui.skeleton animation_speed being sent as milliseconds

### LINKED ISSUE #6107: The `animation_speed` of `ui.skeleton` whose unit should be milliseconds
### First Check

### Example Code

```python
from nicegui import ui

def index():
    ui.skeleton('QChip',animation_speed=1.5)
    ui.skeleton('QChip',animation_speed=1500)

ui.run(
    root=index,
)
```

### Description

`animation_speed` of `ui.skeleton` was described as:
```
animation speed in seconds (default: 1.5)
```
but in https://quasar.dev/vue-components/skeleton#qskeleton-api , it is:
```
Animation speed (in milliseconds, without unit)
```
So, in example code, `1500` is same as `1.5`:

Image

### NiceGUI Version

3.13.0

### Python Version

3.13.14

### Browser

Other

### Operating System

Windows

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #6107.

`ui.skeleton`'s `animation_speed` is documented as *"animation speed in seconds (default: 1.5)"*, but the value is forwarded verbatim to Quasar's [`QSkeleton`](https://quasar.dev/vue-components/skeleton#qskeleton-api) `animation-speed` prop, which is in **milliseconds**.

The old default `1.5` only ever looked correct by coincidence: `1.5` matched the sentinel that *omits* the prop, so Quasar fell back to its own default of `1500ms` (= 1.5 s). Any explicitly passed value was sent as-is in milliseconds, so e.g. `animation_speed=2.0` rendered as `2ms` (a sub-frame, effectively invisible animation) rather than the 2 seconds the docstring implied.

### Implementation

Keep the public API in **seconds** (the documented, intended unit) and convert to milliseconds internally — matching how every other duration in NiceGUI is handled:

- `ui.notification` `timeout` — seconds, `(timeout or 0) * 1000`
- `ui.scroll_area` `duration` — seconds, `1000 * duration`
- `ui.page_scroller` `duration` — seconds, `* 1000`
- `ui.sortable` `animation` — seconds, `* 1000`

The diff is a single line: `animation_speed` → `animation_speed * 1000` when forwarding to the prop. The default (`1.5`) and the docstring (*"in seconds"*) are unchanged, so **the public API is untouched and this is a pure bugfix** — the docs were always right, the code just wasn't doing what they said.

| Caller | Before (buggy) | This PR (seconds) |
|---|---|---|
| default `1.5` | 1.5 s ✓ | 1.5 s ✓ |
| `2.0` (trusted the "seconds" docstring) | 2 ms ✗ | **2 s ✓** |
| `1500` (worked around the bug) | 1.5 s | 1500 s |

This rewards everyone who followed the documentation and only changes behavior for callers who relied on the undocumented millisecond passthrough.

### Progress

- [x] Documentation has been updated. (no change needed — the docstring already said "in seconds")

---

*Updated direction after [@falkoschindler's review](https://github.com/zauberzeug/nicegui/pull/6110#pullrequestreview-4497622222): switched from changing the public unit to milliseconds, to keeping seconds and converting internally for consistency with the rest of the library. Branch force-pushed (amended single commit). Earlier revisions of this description referred to the milliseconds approach.*

### DIFF
--- a/nicegui/elements/skeleton.py
+++ b/nicegui/elements/skeleton.py
@@ -62,7 +62,7 @@ class Skeleton(Element):
         self._props.set_optional('type', type if type != 'rect' else None)
         self._props.set_optional('tag', tag if tag != 'div' else None)
         self._props.set_optional('animation', animation if animation != 'wave' else None)
-        self._props.set_optional('animation-speed', animation_speed if animation_speed != 1.5 else None)
+        self._props.set_optional('animation-speed', animation_speed * 1000 if animation_speed != 1.5 else None)
         self._props.set_bool('square', square)
         self._props.set_bool('bordered', bordered)
         self._props.set_optional('size', size)


evnchn:
I think we can get away with it, citing the old behaviour was a bug.

falkoschindler (changes requested):
Thanks for tracking down the unit mismatch — the root-cause analysis is spot on.

One thing I'd reconsider before merging: rather than switching the public API to milliseconds, I think we should keep `animation_speed` in **seconds** and convert internally. NiceGUI normalizes time durations to seconds everywhere and does the `* 1000` itself when forwarding to Quasar/JS:

- `ui.notification` `timeout` — seconds, `(timeout or 0) * 1000`
- `ui.scroll_area` `duration` — seconds, `1000 * duration`
- `ui.page_scroller` `duration` — seconds, `* 1000`
- `ui.sortable` `animation` — seconds, `* 1000`

`skeleton.animation_speed` would be the only duration in the public API exposing the underlying JS millisecond unit. And the existing docstring already said *"in seconds"* — so the docs were right and the code was wrong; the cleaner fix makes seconds actually true rather than rewriting the docs to match the bug:

```python
animation_speed: float = 1.5,   # unchanged
...
:param animation_speed: animation speed in seconds (default: 1.5)   # unchanged
...
self._props.set_optional('animation-speed',
                         animation_speed * 1000 if animation_speed != 1.5 else None)
```

(No rounding, to match the sibling elements that send the raw `* 1000` float.)

This also has a smaller breaking surface than the milliseconds version:

| Caller | Before (buggy) | This PR (ms) | Seconds |
|---|---|---|---|
| default `1.5` | 1.5 s ✓ | 1.5 s ✓ | 1.5 s ✓ |
| `2.0` (trusted the "seconds" docstring) | 2 ms ✗ | 2 ms ✗ | **2 s ✓** |
| `1500` (worked around the bug) | 1.5 s | 1.5 s ✓ | 1500 s ✗ |

So the seconds approach rewards everyone who followed the documentation and only breaks callers who relied on the undocumented millisecond passthrough. Either direction is a behavior change for 3.14, but seconds keeps us consistent with the rest of the library.

WDYT? If you're on board, the diff shrinks to that one line and you can drop the docstring change entirely.

evnchn:
I swear I recall NiceGUI doing milliseconds off my brain but then yep it's seconds. 

Some Mandela effect going on ...

Definitely convert to seconds then. Then we can simply treat this PR as a bugfix, no public API change whatsoever. Nice!

evnchn:
@falkoschindler force pushed since we're changing directions. Followed your one line fix

Bonus: updated PR description.

falkoschindler (approved):

PR #6111, opened 2026-06-14
Fix native window close hanging the process on Windows

### LINKED ISSUE #5443: ConnectionResetError from loading too many video elements in native browser

evnchn:
### Motivation

Fixes #5443: closing a native-mode window can hang the process forever on Windows. uvicorn's graceful shutdown enters a connection-drain loop that never completes when a ghost connection is present — a `ConnectionResetError: [WinError 10054]` raised inside the proactor's `_call_connection_lost` aborts the transport cleanup before it detaches from the asyncio `Server`, so `server.wait_closed()` never resolves. The user closes the window and the process stays alive.

This implements the direction @falkoschindler proposed in #5443: in native mode there is no live browser left to drain, so after the window closes we run the user's `on_shutdown` callbacks and then `os._exit(0)`. It supersedes the closed #6105 and the `timeout_graceful_shutdown` default from #5706 (which truncates legitimately-slow shutdown work and affects non-native users).

### Implementation

`Server.shutdown()` runs the connection drain *before* lifespan shutdown (verified from source on uvicorn 0.40.0 and 0.49.0):

```
shutdown():
  await wait_for(_wait_tasks_to_complete(), timeout=timeout_graceful_shutdown)  # connection drain (the hang)
  await self.lifespan.shutdown()                                                # on_shutdown callbacks
```

So "wait for lifespan shutdown, then `os._exit`" cannot work — lifespan shutdown is gated behind the very drain that hangs. Instead, on native window close we drive NiceGUI's own shutdown directly: schedule `app.stop()` (which runs every `on_shutdown` callback) on the running loop, wait for it, then `os._exit(0)`. Because the process exits there, uvicorn never reaches its own lifespan shutdown and the callbacks run exactly once.

Two call sites, native mode only: `native_mode.py` for `reload=False` (the #4970 repro and packaged apps) and `server.py`'s `monitor_shutdown_event` for `reload=True`. Non-native shutdown is untouched.

### Progress

- [ ] Pytests have been added/updated or are not necessary. 

---

### Windows verification

DeskPC: Windows 11 LTSC, Python 3.12.10, pywebview 6.2.1, WebView2 149, uvicorn 0.49.0. Repro per #4970: two `ui.video` elements, `ui.run(native=True, reload=False)`, a timer destroys the window after 8 s, an `on_shutdown` writes a marker; measured destroy→exit and killed any tree still alive at +30 s. 20 trials `reload=False` + 5 trials `reload=True`.

| approach | `reload=False` hangs | `on_shutdown` ran |
|---|---|---|
| wait for lifespan then `os._exit` (the literal sketch) | 11/20 | 9/20 — deadlocks behind the drain |
| **this PR** (`app.stop()` then `os._exit`) | **0/20** | **20/20** |

`reload=True` passed 5/5 for both approaches; this machine never hangs in that mode, so that arm only confirms no regression.

### Open questions

- This touches the same `monitor_shutdown_event` block as #6106 (the `#5845` hot-reload deadlock fix). They'll need reconciling — the combined shape is `while not event.is_set(): sleep(0.1)` (#6106) → `app.stop()` → `os._exit(0)` (this PR).
- `os._exit` skips uvicorn's remaining teardown (`_thread.interrupt_main`, `event_manager.stop`, `native.remove_queues`); intended for a terminating process, but flagging it for review.


falkoschindler:
@evnchn What's the status of this draft PR?
Since it technically conflicts with PR #6106, I wonder if we should merge them. Or we block this PR until #6106 is merged.

evnchn:
@falkoschindler verified working with homelab Windows hardware, you should check the diff and see if it is what you want per your comment in #5443, if so merge AFTER #6106 since that one is more mature

PS: last attempt (1) distorted what you want to do in #5443 and (2) don't even work, hence the closure.

falkoschindler:
@evnchn Now that PR #6106 is merged, can you, please, have a look how to resolve the conflict in server.py? Thanks!

evnchn:
*Posted by Claude Code (Opus 4.8) on behalf of @evnchn; the merge resolution and the verification below were run and checked empirically.*

Done — resolved the `server.py` conflict against current `main`. The PR now reports **MERGEABLE**.

**TL;DR:** the conflict was #6106 and this PR editing the same `monitor_shutdown_event` body. Combined as the *Open questions* predicted — #6106's poll-loop **replaces** `event.wait()`, then this PR's hard-exit **replaces** `self.should_exit = True`:

```python
def monitor_shutdown_event() -> None:
    # Poll instead of `event.wait()`: a blocking wait leaves a stale waiter if the reloader
    # kills the process mid-wait, making a later `event.set()` hang forever (#5845).
    while not event.is_set():
        time.sleep(0.1)
    # Run on_shutdown callbacks, then hard-exit, skipping uvicorn's connection drain, which runs
    # before lifespan shutdown and can hang forever on a ghost Windows connection (#5443).
    if core.loop is not None and core.loop.is_running() and not core.app.is_stopped:
        future = asyncio.run_coroutine_threadsafe(core.app.stop(), core.loop)
        with suppress(Exception):
            future.result(timeout=30)
    sys.stdout.flush()
    sys.stderr.flush()
    os._exit(0)
```

`self.should_exit = True` drops out (moot after `os._exit(0)`); imports keep both `time` (#6106) and `asyncio`/`os`/`sys`/`suppress`. Local gates green: `pre-commit`, `mypy ./nicegui`, `pylint` 10.00/10.

<details>
<summary><b>Re-verified on Windows against current <code>main</code> — the fix still holds after the merge</b></summary>

Re-ran the #4970 repro (two local `ui.video` files served over the loopback, `ui.run(native=True, reload=False)`, window auto-destroyed after 8 s, `on_shutdown` marker) on DeskPC — Windows 11 LTSC, py3.12.10, pywebview 6.2.1, WebView2 149, uvicorn 0.49.0 — **20 trials per arm, same box, only the code differs**:

| arm | hangs | `on_shutdown` ran |
|---|---|---|
| current `main` (= #6106, **without** this PR) | **8/20 (40%)** | 12/20 |
| **this PR merged into `main`** | **0/20** | 20/20 |

So the hang is still very much alive on current `main` (the merge of #6106 didn't incidentally fix it — it's a different, non-overlapping path), and this PR still drives it to zero with `on_shutdown` running every time. Hung trials are the ones where `on_shutdown` never ran — the drain-deadlock signature from the PR description.

I also confirmed the merge preserves #6106 itself: its stale-waiter bug reproduces deterministically (kill a process parked in `event.wait()` → a later `event.set()` hangs; the poll-loop → returns instantly), and the merged code keeps the poll-loop verbatim. `reload=True` native close also still exits cleanly (8/8) and after a real hot reload (6/6).

The `os._exit` teardown-skip open-question from the PR body still stands for review.
</details>

Conflict resolved and re-verified end-to-end on real hardware — nothing blocking from my side. The one open design question is the `os._exit` teardown-skip flagged in the PR body; happy to go whichever way you prefer there.

evnchn:
*Posted by Claude Code (Opus 4.8) on behalf of @evnchn.*

**DRY follow-up (`cd7a45b4`):** deduplicated the hard-exit block — `server.py` now calls the shared `hard_exit_after_shutdown()` instead of an inlined copy (net −15/+4, no behavior change).

<details>
<summary><b>Details + checks</b></summary>

The block was duplicated: a `_hard_exit_after_shutdown()` helper in `native_mode.py` (reload=False path) and a byte-identical inlined copy in `server.py`'s `monitor_shutdown_event` (reload=True path). Both now call one shared `hard_exit_after_shutdown()` (dropped the underscore as it's a shared helper now), which also lets `server.py` shed four imports (`asyncio`/`os`/`sys`/`suppress`).

- **Static**: `pre-commit`, `mypy ./nicegui`, `pylint` 10.00/10.
- **Independent review** (Codex): no circular import (`nicegui/native/*` never imports `server`, and `native/__init__` already loads `native_mode`), call-time `core.loop`/`core.app` deref makes it behavior-equivalent, rename complete with no stragglers.
- **Native runtime** (not CI-covered): re-ran the repro on the same Windows box — reload=False **0/15** hangs, reload=True **0/6**, `on_shutdown` every trial.
</details>

falkoschindler (changes requested):
Thanks, @evnchn! The PR is looking pretty good. I've got only 4 review findings that should be rather simple to address:

1. **`on_shutdown` execution in reload mode is unverified** (major) — The verification table only measures `on_shutdown` for `reload=False` (20/20). For `reload=True` it states only "no regression / never hangs," not that callbacks ran. In reload mode the helper fires in _two_ processes on the same event: the reloader parent's `check_shutdown` and the worker's `monitor_shutdown_event`. The design relies on the parent's `core.loop` being `None`/not-running (so its guard skips `app.stop()`) and the worker actually running the callbacks before the parent's `os._exit(0)` tears the tree down. That's plausible but timing-dependent. Could you confirm on Windows that an `on_shutdown` marker is written exactly once in `reload=True`? That's the one behavioral claim in the PR body ("callbacks run exactly once") not backed by the measurements.

2. **Dead code left behind** (major) — After this change, `native.remove_queues()` and `EventManager.stop()` have **zero callers** anywhere in the tree (only their definitions remain). They were the cleanup the old `check_shutdown` performed. Since `os._exit` lets the OS reclaim the pipes/queues this is functionally harmless, but leaving two now-unreachable functions is a code smell. Either remove them in this PR or add a one-line note at their definitions that native shutdown intentionally relies on process exit instead. (`_thread` import removal is correct and clean.)

3. **`suppress(Exception)` hides shutdown-handler failures** (minor) — `future.result(timeout=30)` is wrapped in `suppress(Exception)`, so an exception raised by a user's `on_shutdown` callback (or a 30 s timeout) vanishes silently right before exit. For a terminating path swallowing is defensible, but a `log.exception(...)`/`log.warning(...)` in the handler would make a failed or hung shutdown observable rather than invisible. The 30 s timeout itself is a reasonable safety valve.

4. **Layering: `server.py` importing from `native.native_mode`** (minor) — Pulling `hard_exit_after_shutdown` from `native.native_mode` into `server.py` is a slightly surprising direction. The import is safe today (no cycle; `native_mode` is already imported transitively via `native/__init__.py`, so no new eager `webview` import cost), but the helper's body is _not native-specific at all_ — it's pure NiceGUI lifecycle (run `on_shutdown` handlers, then exit), so a native module is an odd home and `server` reaching across into `native` to fetch it is the wrong direction to add to.

   The cleanest alternative is a free function in `core` — say `core.stop_and_exit()`:
   - No cycle: `core` imports neither `server` nor `native`, and both call sites already import `core` and hold `core.app`.
   - No privacy wart: it's freely callable like the existing `core.is_loop_running()` (a private `App` method would force `# pylint: disable=protected-access` at both call sites, and a _public_ `App` method would expose an `os._exit`-ing method on the documented `nicegui.app.App` surface).
   - Good home: `core` already owns `app`, `loop`, and the lifecycle predicates (`is_loop_running`, `is_script_mode_*`) the helper depends on, so the guard collapses to `core.is_loop_running()`.

evnchn:

**Re: your `reload=True` Windows question — verified, PASS.** `on_shutdown` fires **exactly once** under native `reload=True`, from the uvicorn **worker only**; the reloader parent never runs it. No double-invocation. Tested on a real Win11 box at the current PR head (`f2606c30`).

<details>
<summary><b>Trials & data</b></summary>

| config | trials | `on_shutdown` count per close | firing process |
|---|---|---|---|
| **native + `reload=True`** | **3/3** | **1** (never 0, never 2) | uvicorn **worker** only |
| native + `reload=False` (control) | 2/2 | 1 | server process |

Representative `reload=True` marker: three `IMPORT` lines (reloader, worker, webview) then a single `SHUTDOWN pid=<worker> ppid=<reloader>`. `SHUTDOWN_COUNT=1`, one distinct shutdown pid every trial; the reloader pid never appears as a shutdown pid. The process also self-terminated gracefully (never had to be force-killed).
</details>

<details>
<summary><b>Why only the worker fires it (native + reload topology)</b></summary>

`shim python → real python (uvicorn reload supervisor, holds the listen socket) → two spawn children: the uvicorn worker (runs the app + `on_shutdown`) AND the native webview process`. The reloader supervisor has no app instance, so it has no `on_shutdown` to run. `reload=False` collapses to `shim → server → one webview child`. This is the structural reason there's no reloader-vs-worker double-fire.
</details>

<details>
<summary><b>Method & fidelity (honest — what was and wasn't exercised)</b></summary>

Full transparency on the trigger: this was **not** a literal mouse X-click on a rendered window. Two environmental walls on the box: a GUI launched over SSH lands in the non-interactive Session 0 (WebView2 starts but never gets a usable desktop/websocket), and the interactive desktop session belongs to a *different* local account I don't have credentials for (I declined to guess them).

So the close was triggered by **terminating only the webview process** (identified deterministically by its loaded DLLs — `WebView2Loader.dll` / `EmbeddedBrowserWebView.dll` / `System.Windows.Forms`; the worker loads none), leaving the worker untouched. Per this PR's own `native_mode.check_shutdown`, `on_shutdown` is driven **solely** by the webview process becoming not-alive:
```python
def check_shutdown() -> None:
    while process.is_alive():
        time.sleep(0.1)
    ...
    core.stop_and_exit()
```
That's the identical downstream path an X-click takes (X → `closed` → `webview.start()` returns → process exits → `check_shutdown` → `core.stop_and_exit()` → `on_shutdown`). The only thing **not** exercised is pywebview's in-process `closing`/`closed` handlers — upstream of, and unable to change, the `on_shutdown` count or which process fires it. So this is faithful for the reloader-vs-worker count question; a literal X-click would yield the same result. Happy to do a hand X-click run too if you'd like belt-and-suspenders.

Environment: pywebview 6.2.1, WebView2 runtime 149.0.4022.98, Python 3.12, nicegui `@f2606c30`.
</details>

evnchn:

**TL;DR: pushed `f2606c30` addressing the review.** Three of four points are in; the fourth (`reload=True` on Windows) needs a manual run.

The headline finding: `os._exit` was silently skipping `atexit`/`finally` cleanup that ran *before* this PR — caught it empirically, fixed it.

- **atexit cleanup** — `core.stop_and_exit()` now runs `atexit._run_exitfuncs()` before `os._exit(0)`. (`finally`-after-`ui.run()` still can't run — `os._exit` can't resume the blocked main thread; stated as a documented limitation in the docstring.)
- **(3) suppressed failure** — `log.exception(...)` added; scheduling moved inside the `try` so a loop race can't skip the hard exit.
- **(4) layering** — moved `hard_exit_after_shutdown` → `core.stop_and_exit()`, dropping the `server → native_mode` import; guard via a captured loop.
- **(2) dead code** — `remove_queues()` + `EventManager.stop()` removed (grep-confirmed zero callers, incl. tests).

<details>
<summary><b>Empirical before/after (the cleanup regression + its fix)</b></summary>

Harness mints one nonce per cleanup mechanism, shuts down each way, and diffs which nonces leak (a left-behind nonce = that cleanup did not run). Run against the real shutdown function in server mode (the native window-close calls exactly this function):

| cleanup mechanism | graceful (≈ pre-PR normal exit) | old hard-exit (`os._exit`) | **fixed** `core.stop_and_exit()` |
|---|---|---|---|
| `app.on_shutdown(...)` | cleaned | cleaned | **cleaned** |
| `atexit.register(...)` | cleaned | **leaked** | **cleaned** |
| post-`ui.run()` / `finally` | cleaned | leaked | leaked → documented limitation |
| `__del__` finalizer | leaked | leaked | leaked (unreliable in all; not a regression) |

So pre-PR an `atexit`/`finally` nonce was cleaned; the original hard-exit leaked both; the fix recovers `atexit`. Harness available on request.
</details>

<details>
<summary><b>CI, second-model review, and an honest callout</b></summary>

- CI (local): `mypy` clean (241 files), `pylint` 10.00/10, `pytest tests/test_lifecycle.py` 6/6, pre-commit clean.
- **Independent review (Codex, different model lineage):** `VERDICT: CONCUR-PUSH` — *"current_loop is captured, and both run_coroutine_threadsafe(...) and future.result(...) are inside the try, so a scheduling race cannot skip atexit._run_exitfuncs() or os._exit(0)."* It also confirmed `_run_exitfuncs()` clears handlers so a later normal exit won't double-run them, and that the dead-code removal has no remaining callers.
- **Callout:** `atexit` now runs from the daemon watcher thread (not normal interpreter finalization), while other threads may still be alive — the pragmatic tradeoff for the `os._exit` escape hatch. Flagging in case you'd rather scope it.
</details>

**Still open — your finding (1) (`reload=True`, Windows):** confirming `on_shutdown` fires exactly once under `reload=True` needs a real Windows desktop session; the native window-close can't be driven headless over SSH (a GUI launched over SSH lands in a non-interactive session — no clickable window X). Planning a manual run on the physical box — will report the marker count (1 = pass, 0 = didn't fire, 2 = double-fire).

Please re-review.

falkoschindler (approved):
Reviewed and looks good to me — thanks for the thorough Windows verification, @evnchn. 🙏

I re-checked the whole thread and all earlier findings are addressed:

- **reload=True double-fire** — verified `on_shutdown` fires exactly once (worker only); structurally the reloader parent has `core.loop is None` so it skips `app.stop()`.
- **dead code** — `remove_queues()` and `EventManager.stop()` removed, no remaining callers.
- **suppressed failure** — `log.exception(...)` added, scheduling moved inside the `try`.
- **layering** — logic lives in `core.stop_and_exit()`, no `server → native_mode` import.

I pushed one small follow-up commit (`d11ae93`) tidying `core.py`: dropped the `current_loop` local in `stop_and_exit` (not needed for mypy narrowing, and the check-then-use race can't occur since the only writers of `core.loop` are startup and the test-only `reset()`), and moved `stop_and_exit` below the `is_script_mode_*` predicates to keep the small `is_*` helpers grouped. No behavior change.

The `os._exit` teardown-skip is an acceptable tradeoff for a terminating process, and `atexit._run_exitfuncs()` recovers the `atexit` cleanup. Approving.

PR #6112, opened 2026-06-14
Fix #5684: serve no-op stub modules to worker subprocesses

### LINKED ISSUE #5684: Slow startup and code re-execution in multiprocessing contexts
This umbrella issue consolidates multiple reports about NiceGUI's behavior when Python spawns subprocesses. The related tickets below will be closed as duplicates, pointing here for tracking and discussion.

| Type       | Link  | Summary                                                  |
| ---------- | ----- | -------------------------------------------------------- |
| Issue      | #3356 | Native mode startup 2x slower due to webview subprocess  |
| PR         | #3365 | Proposes separate `nicegui_webview` package              |
| Discussion | #4353 | `multiprocessing.Pool` 4-5x slower with NiceGUI          |
| Issue      | #4412 | Multiprocessing causes code reload (Windows)             |
| PR         | #4429 | Profile imports - 70% is Matplotlib, mocking experiments |
| Discussion | #4465 | Better FAQ for "code executed twice"                     |
| PR         | #4471 | PoC: fast `cpu_bound` with lazy dummy modules            |
| Issue      | #4542 | `run.cpu_bound` causes code reload on first call         |
| Issue      | #4891 | `Manager()` causes delay                                 |
| PR         | #5303 | Lazy imports via `lazy-imports` library     |
| Discussion | #5574 | Proposal: use loky executor to avoid slow first call     |

### Problem

When Python spawns subprocesses (for native mode, `run.cpu_bound()`, `multiprocessing.Pool`, `Manager()`, etc.), each subprocess re-imports the main module. Because NiceGUI performs significant initialization at import time (creating the `App`, importing FastAPI, Matplotlib, and ~200 element classes), this causes:

1. **Slow startup** - especially in native mode where the webview runs in a subprocess
2. **Code executed multiple times** - user code outside `if __name__ == '__main__'` runs in each subprocess
3. **Slow first `cpu_bound` call** - worker processes must import NiceGUI before executing the task

### Affected Scenarios

1. **Code executed twice with `reload=True`**

    The reload mechanism spawns a subprocess, causing module-level code to run twice.
    
    ```python
    from nicegui import ui
    
    print('Running main', f'{__name__=}')
    
    ui.label('Hello')
    ui.run()
    ```
    
    **Output:**
    
    ```
    Running main __name__='__mp_main__'
    NiceGUI ready to go on http://localhost:8080, http://192.168.7.151:8080, and http://192.168.7.168:8080
    Running main __name__='__main__'
    ```
    
2. **Native mode (`native=True`) causes double initialization despite `reload=False`**

    The webview subprocess re-imports NiceGUI, causing all import-time initialization to run twice.
    
    ```python
    import time
    
    start = time.perf_counter()
    from nicegui import ui
    print(f'Import took {time.perf_counter() - start:.2f} seconds')
    
    @ui.page('/')
    def root():
        ui.label('Hello!')
    
    ui.run(native=True, reload=False)
    ```
    
    **Output:**
    
    ```
    Import took 0.52 seconds
    Import took 0.50 seconds   # <-- again in webview subprocess
    ```
    
    With `native=False`, the import message only prints once. The doubled import time causes a noticeable delay before the window appears.
    
3. **`Manager()` is slow**

    Creating a multiprocessing Manager spawns a subprocess that imports NiceGUI.
    
    ```python
    import time
    from multiprocessing import Manager
    
    from nicegui import ui
    
    
    def root() -> None:
        @ui.button('Create a manager').on_click
        def _():
            start = time.perf_counter()
            Manager()
            print(f'Manager() took {time.perf_counter() - start:.2f}s')
    
    
    ui.run(root)
    ```
    
    **Output:**
    
    ```
    Manager().Queue() took 0.62s
    ```
    
4. **`run.cpu_bound()` is slow on first call**

    Worker processes import NiceGUI before executing the task. Subsequent calls are fast because the process pool is reused.
    
    ```python
    import asyncio
    import time
    from nicegui import run, ui
    
    def wait():
        time.sleep(1.0)
    
    def root() -> None:
        @ui.button('Run 10 CPU-bound processes').on_click
        async def _():
            start = time.perf_counter()
            await asyncio.gather(*[run.cpu_bound(wait) for _ in range(10)])
            print(f'All done in {time.perf_counter() - start:.2f} s')
    
    ui.run(root)
    ```
    
    **Output:**
    
    ```
    All done in 1.63 s
    All done in 1.01 s
    All done in 1.01 s
    ...
    ```
    
### Solution Direction

Two approaches are being explored:

1. **Lazy imports ([#5303](https://github.com/zauberzeug/nicegui/pull/5303))**: Uses the [`lazy-imports`](https://pypi.org/project/lazy-imports/) library to defer heavy imports until actually needed. Achieves ~85% reduction in import time without API changes.

2. **Alternative executor ([#5574](https://github.com/zauberzeug/nicegui/discussions/5574))**: Proposes using [`loky`](https://github.com/joblib/loky) instead of `ProcessPoolExecutor`. Avoids slow first call and removes need for main guards by using `cloudpickle` for serialization.

Key findings from profiling:

- ~70% of import time is Matplotlib (optional via `NICEGUI_INCLUDE_MATPLOTLIB=false`)
- FastAPI import is expensive but required for the main process
- Subprocesses don't need most of NiceGUI's functionality

### Workarounds (for users)

Until this is fixed, users can mitigate the issue by:

1. **Use `reload=False`** and guard imports:

   ```python
   if __name__ == '__main__':
       from nicegui import ui
       ui.run(reload=False)
   ```

2. **Set environment variable** in worker processes:

   ```python
   import os
   os.environ['IS_WORKER'] = '1'
   # ... spawn subprocess

   # In main module:
   if not os.environ.get('IS_WORKER'):
       from nicegui import ui
       # ...
   ```

3. **Disable Matplotlib** if not needed:
   ```bash
   export NICEGUI_INCLUDE_MATPLOTLIB=false
   ```

evnchn:
*Drafted with Claude Code (assisting @evnchn). Reworked 2026-07-06 per the review — this PR is now the **stubs-only** half; the lazy-package-import enabler was split out per the direction-check answer and is staged at evnchn#195 before going upstream separately. The first commit here is that enabler and will drop out once it lands.*

### Motivation

Follow-up to #5684 and the merged #5303: every spawned worker subprocess (`run.cpu_bound`, `multiprocessing.Pool`, `Manager()`) re-imports the user's main module and re-pays the NiceGUI import. With lazy imports alone this is still ~300–500 ms per worker for the official examples, because a trailing `ui.run()` or module-level UI re-materializes the app.

### Implementation

While the server is running, `ui.run()` marks the environment (`NICEGUI_WORKER_STUBS=1`); spawn workers then pre-fill `sys.modules` with no-op stubs for UI modules, making the re-import practically free: `examples/progress` worker re-import 495 → ~25 ms, `menu_and_tabs` 328 → ~4.5 ms — matching loky-style pool reuse (#5574) with zero new dependencies.

Failure semantics per the review of the first draft (details in the review reply below):

- **data-bearing modules stay real in workers** (`binding`, `observables`, `events`, `dataclasses`, `awaitable_response` + transitive closure) — classes built from them keep working and pickled references resolve to actual classes; closure is enforced by a test (findings 3, 5)
- **the `Whatever` no-op is falsy, iterates empty, contains nothing, and never compares equal** (`__bool__`/`__iter__`/`__contains__`/`__eq__` defined; without `__iter__` the legacy sequence protocol iterated forever); unsupported operations still raise loudly (findings 1, 2)
- **the marker is scoped to the server's lifetime** (set at server start, popped in a `try/finally` at stop); the pytest carve-out is gone, so stub behavior is testable under the `Screen` fixture (finding 4 — see the reply for why the pool-initializer variant can't work for spawn)

### Progress

- [ ] The implementation is complete. (Draft until the split-out enabler lands and this rebases onto it.)
- [x] Documentation has been added/updated (module docstring documents worker semantics and known limitations).
- [x] No breaking changes to the public API. (Behavior note: in stub workers, `from nicegui import App`-style access to stubbed submodules now raises `ModuleNotFoundError` loudly instead of hanging.)

### DIFF
--- a/nicegui/__init__.py
+++ b/nicegui/__init__.py
@@ -1,13 +1,38 @@
-from . import binding, elements, html, run, storage, ui
-from .api_router import APIRouter
-from .app.app import App
-from .client import Client
-from .context import context
-from .element_filter import ElementFilter
-from .event import Event
-from .nicegui import app
-from .page_arguments import PageArguments
-from .version import __version__
+import sys
+from types import ModuleType
+from typing import TYPE_CHECKING
+
+from . import _lazy
+
+_LAZY_IMPORTS = {
+    'APIRouter': ('.api_router', 'APIRouter'),
+    'App': ('.app.app', 'App'),
+    'Client': ('.client', 'Client'),
+    'ElementFilter': ('.element_filter', 'ElementFilter'),
+    'Event': ('.event', 'Event'),
+    'PageArguments': ('.page_arguments', 'PageArguments'),
+    '__version__': ('.version', '__version__'),
+    'app': ('.nicegui', 'app'),
+    'binding': ('.binding', ''),
+    'context': ('.context', 'context'),
+    'elements': ('.elements', ''),
+    'html': ('.html', ''),
+    'run': ('.run', ''),
+    'storage': ('.storage', ''),
+    'ui': ('.ui', ''),
+}
+
+if TYPE_CHECKING:
+    from . import binding, elements, html, run, storage, ui
+    from .api_router import APIRouter
+    from .app.app import App
+    from .client import Client
+    from .context import context
+    from .element_filter import ElementFilter
+    from .event import Event
+    from .nicegui import app
+    from .page_arguments import PageArguments
+    from .version import __version__
 
 __all__ = [
     'APIRouter',
@@ -26,3 +51,31 @@ __all__ = [
     'storage',
     'ui',
 ]
+
+
+def __dir__() -> list[str]:
+    return __all__
+
+
+def __getattr__(name: str) -> object:
+    return _lazy.resolve(__name__, 'nicegui', _LAZY_IMPORTS, name)
+
+
+class _PackageModule(ModuleType):
+    """Class for the nicegui package module itself.
+
+    The attributes ``app`` and ``context`` collide with the submodules ``nicegui.app`` and ``nicegui.context``:
+    whenever the import machinery loads those submodules, it binds them as package attributes,
+    shadowing the ``App`` and ``Context`` instances which ``__getattr__`` provides.
+    (The eager package init used to win this race by assignment order.)
+    Ignoring module bindings for these two names keeps them resolvable via ``__getattr__``,
+    while non-module assignments (e.g. ``mock.patch``) still work as plain instance attributes.
+    """
+
+    def __setattr__(self, name: str, value: object) -> None:
+        if name in {'app', 'context'} and isinstance(value, ModuleType):
+            return  # absorb the import machinery's submodule binding (see class docstring)
+        super().__setattr__(name, value)
+
+
+sys.modules[__name__].__class__ = _PackageModule
--- a/nicegui/__init__.py
+++ b/nicegui/__init__.py
@@ -1,8 +1,15 @@
+import os
 import sys
 from types import ModuleType
 from typing import TYPE_CHECKING
 
 from . import _lazy
+
+if os.environ.get('NICEGUI_WORKER_STUBS') == '1':
+    import multiprocessing  # cheap here: already imported by the spawn bootstrap in any real worker
+    if multiprocessing.current_process().name != 'MainProcess':
+        from . import _worker_stubs
+        _worker_stubs.install()
 
 _LAZY_IMPORTS = {
     'APIRouter': ('.api_router', 'APIRouter'),
--- a/nicegui/_lazy.py
+++ b/nicegui/_lazy.py
@@ -1,0 +1,20 @@
+"""Shared machinery for lazily importing module attributes via a PEP 562 module ``__getattr__``."""
+import importlib
+import sys
+from collections.abc import Mapping
+
+
+def resolve(module_name: str, package: str, lazy_imports: Mapping[str, tuple[str, str | None]], name: str) -> object:
+    """Import the attribute ``name`` as declared in ``lazy_imports`` and cache it on module ``module_name``.
+
+    Each entry in ``lazy_imports`` maps an attribute name to a ``(module path, attribute name)`` tuple,
+    where the module path is resolved relative to ``package``.
+    An empty attribute name yields the module itself.
+    """
+    if name not in lazy_imports:
+        raise AttributeError(f'module {module_name!r} has no attribute {name!r}')
+    module_path, attr_name = lazy_imports[name]
+    module = importlib.import_module(module_path, package=package)
+    value = getattr(module, attr_name) if attr_name else module
+    setattr(sys.modules[module_name], name, value)
+    return value
--- a/nicegui/_worker_stubs.py
+++ b/nicegui/_worker_stubs.py
@@ -1,0 +1,118 @@
+"""Stub modules for worker subprocesses (see https://github.com/zauberzeug/nicegui/issues/5684).
+
+When a NiceGUI app spawns worker subprocesses (``run.cpu_bound``, ``multiprocessing.Pool``, ``Manager()``, ...),
+each worker re-imports the user's main module and thereby the nicegui package.
+Workers don't serve UI, so UI-related modules are replaced with no-op stubs which makes the re-import
+practically free and turns module-level UI code (``ui.label(...)``, ``ui.run(...)``) into harmless no-ops.
+
+Modules listed in ``REAL_MODULES`` are NOT stubbed:
+
+- ``run`` must be real because unpickling a ``run.cpu_bound`` payload imports it in the worker
+- ``native`` must be real because the native-mode window subprocess executes its entry point
+- data-bearing modules (``binding``, ``dataclasses``, ``events``, ``observables``, ...) must be real so that
+  classes derived from them (e.g. via ``@binding.bindable_dataclass``) keep working in workers,
+  including unpickling ``run.cpu_bound`` payloads that reference such classes
+- the set is closed under imports (enforced by ``tests/test_worker_stubs.py``):
+  a real module must never import from a stubbed one, or it would bind no-op values
+
+Known limitations (by design — workers should not consume UI state):
+
+- worker code that READS values from stubbed modules gets ``WHATEVER`` objects: equality comparisons,
+  truthiness and containment are ``False``, iteration is empty — conditions silently take the negative branch
+- operations that are not stubbed (``len()``, ``int()``, ``+`` with a real left operand, format specs)
+  raise loudly, which is preferable to silent corruption
+- ``import nicegui.elements.button``-style submodule imports and ``from nicegui import App``-style lazy
+  attributes of stubbed submodules raise ``ModuleNotFoundError`` in stub mode (``from nicegui import ui`` works)
+- ``nicegui.testing`` is stubbed, which is fine: the pytest plugin loads in pytest's main process, never in workers
+- ``fork``-started children (Linux <= 3.13 default) inherit the parent's real modules, so stubs never engage
+  there — no benefit, but also no cost, since fork does not re-import anything
+"""
+import os
+import pkgutil
+import sys
+from types import ModuleType
+from typing import Literal
+
+REAL_MODULES = {
+    'awaitable_response',
+    'background_tasks',
+    'binding',
+    'context',
+    'core',
+    'dataclasses',
+    'events',
+    'helpers',
+    'json',
+    'logging',
+    'native',
+    'observables',
+    'optional_features',
+    'run',
+    'slot',
+    'version',
+}
+
+
+class Whatever:
+    """A universal no-op object: every operation yields the no-op singleton ``WHATEVER``.
+
+    Truthiness is ``False``, iteration is empty and containment checks are ``False``,
+    so conditions and loops over stub values silently take the empty/negative branch.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        pass  # accept any arguments so that user classes subclassing a stub (via __mro_entries__) can instantiate
+
+    def __getattr__(self, name: str) -> 'Whatever':
+        return WHATEVER
+
+    def __call__(self, *args, **kwargs) -> 'Whatever':
+        return WHATEVER
+
+    def __mro_entries__(self, *args, **kwargs) -> tuple:  # pylint: disable=unused-argument
+        return (Whatever,)  # so module-level `class X(ui.element)` survives class creation
+
+    def __enter__(self) -> 'Whatever':
+        return WHATEVER  # so module-level `with ui.row():` no-ops (e.g. examples/menu_and_tabs)
+
+    def __exit__(self, *args) -> Literal[False]:
+        return False
+
+    def __getitem__(self, item) -> 'Whatever':
+        return WHATEVER  # e.g. generic subscription like Handler[X] in annotations evaluated at class definition
+
+    def __eq__(self, other: object) -> bool:
+        return False  # two unrelated stub values must not compare equal just because they share the singleton
+
+    __hash__ = object.__hash__  # would be unset by defining __eq__; identity hashing keeps stubs usable as dict keys
+
+    def __bool__(self) -> bool:
+        return False  # so module-level `if app.storage...:` and the like take the negative branch
+
+    def __iter__(self):
+        return iter(())  # without this, `__getitem__` would make the legacy sequence protocol iterate forever
+
+    def __contains__(self, item) -> bool:
+        return False
+
+    def __repr__(self) -> str:
+        return '<nicegui worker stub>'
+
+
+WHATEVER = Whatever()
+
+
+class WhateverModule(ModuleType):
+    """A module whose every attribute is the ``WHATEVER`` no-op object."""
+
+    def __getattr__(self, name: str) -> Whatever:
+        return WHATEVER
+
+
+def install() -> None:
+    """Pre-fill ``sys.modules`` with stubs for all non-essential top-level nicegui modules."""
+    for module_info in pkgutil.iter_modules([os.path.dirname(__file__)]):
+        name = module_info.name
+        if name.startswith('_') or name in REAL_MODULES:
+            continue
+        sys.modules[f'nicegui.{name}'] = WhateverModule(f'nicegui.{name}')
--- a/nicegui/core.py
+++ b/nicegui/core.py
@@ -2,16 +2,17 @@ from __future__ import annotations
 
 import asyncio
 import atexit
+import importlib
 import os
 import sys
 from collections.abc import Callable
 from typing import TYPE_CHECKING
 
-from socketio import AsyncServer
-
 from .logging import log
 
 if TYPE_CHECKING:
+    from socketio import AsyncServer
+
     from .air import Air
     from .app import App
     from .client import Client
@@ -23,6 +24,17 @@ air: Air | None = None
 root: Callable | None = None
 script_mode: bool = False
 script_client: Client | None = None
+
+
+def __getattr__(name: str) -> object:
+    if name in {'app', 'sio'}:
+        # lazily build the App and AsyncServer on first access by importing the nicegui module which assigns them
+        module = importlib.import_module('.nicegui', package='nicegui')
+        if name in globals():
+            return globals()[name]
+        # a genuine circular import raises AttributeError with Python's standard "partially initialized" hint
+        return getattr(module, name)
+    raise AttributeError(f"module 'nicegui.core' has no attribute {name!r}")
 
 
 def is_loop_running() -> bool:
--- a/nicegui/helpers/__init__.py
+++ b/nicegui/helpers/__init__.py
@@ -1,3 +1,4 @@
+import multiprocessing
 import os
 
 from .elements import require_top_level_layout
@@ -24,6 +25,7 @@ __all__ = [
     'hash_file_path',
     'is_coroutine_function',
     'is_file',
+    'is_main_process',
     'is_port_open',
     'is_pytest',
     'is_user_simulation',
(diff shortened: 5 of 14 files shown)


evnchn:
@falkoschindler I envision this is the furthering step after https://github.com/zauberzeug/nicegui/pull/5303, cutting into the real meat that is the FastAPI (honestly #5303 didn't since it just regarded elements with heavy dependency). 

Please check it out when you have time, and yes we need to be careful since if this goes wrong it's a critical (FastAPI not work = very bad news)

falkoschindler (changes requested):
Full disclosure up front: this review was written by Claude (Fable) at my request — I haven't read the code myself yet. I usually double-check AI findings before posting, but the key ones below were verified empirically on my machine, some of them rather memorably (see next paragraph), so I'm posting it as-is to unblock the direction-check you asked for.

In fact, one of the findings crashed my MacBook during the review: to check the docstring's claim that "iteration is empty", the agent ran `list(WHATEVER)` — and because `Whatever` defines `__getitem__` without ever raising `IndexError` (and no `__iter__`), the legacy sequence protocol makes it *infinitely* iterable. The probe allocated 132 GB of RAM before taking the machine down. The same happens to any worker whose re-imported main module contains `for x in <stub value>` or `'key' in <stub value>` — the worker hangs or OOMs, and `run.cpu_bound` never returns. That's a pretty good illustration of why I don't yet trust the silent-no-op failure mode, and it sets the tone for the review below.

First the good news, all verified locally: the 17 new tests pass, `import nicegui` really drops to ~3.6 ms, and the PR description is exemplary. The only merge conflict is `nicegui/core.py` vs. the recent `stop_and_exit()` from #6111/#6132 — an easy rebase.

On your direction question: I'd **split the two commits into separate PRs**. Commit 1 (lazy package import) is a clear win and close to mergeable after some simplification (findings 6/7). Commit 2 (worker stubs) has the right goal, but the silent-no-op failure mode is not trustworthy yet — see findings 1–5, several reproduced locally. The universal `WHATEVER` object cannot distinguish "UI side effect to swallow" from "value the worker actually needs", and Python's implicit special-method lookup bypasses `__getattr__`, so the object's actual semantics differ from the documented ones in ways that corrupt worker behavior silently. I'd rather see stubs fail loudly (a clear "not available in worker processes" error naming the attribute) or be scoped to genuinely UI-heavy modules, than yield no-op values.

### Blocking

1. **`WHATEVER` is infinitely iterable and can hang or OOM a worker** — no `__iter__`, and `__getitem__` never raises `IndexError`, so iteration yields `WHATEVER` forever (verified, see above). The docstring promises "iteration is empty". It also breaks `from nicegui import App` in stub mode: importlib iterates the stubbed parent package's `__path__` and never terminates. Fix: add `def __iter__(self): return iter(())`, let `__contains__` return `False`, and reconsider whether `__getitem__` should raise `IndexError` instead of returning `WHATEVER`.

2. **`WHATEVER` is truthy, but the docstring and `core.py` claim it is falsy** — `Whatever` defines no `__bool__`, so `bool(WHATEVER)` is `True` (verified). The module docstring promises "truthiness is `False` — conditions silently take the negative branch", and the comment in `core.py`'s `__getattr__` relies on "a falsy `core.app.is_stopping` in `run._run`". In a stub worker the opposite happens: `core.app.is_stopping` is truthy, so `run._run` returns `None` instead of executing the callback — `run.io_bound(...)` in a worker silently does nothing — and any module-level `if app.storage...:` takes the *positive* branch. Fix: `def __bool__(self): return False`.

3. **Unpickling `cpu_bound` payloads containing nicegui-derived classes breaks** — a worker resolving `__main__.Point` where `Point` was built from a stubbed module (e.g. `@binding.bindable_dataclass`) gets `WHATEVER` instead of the class, and pickle raises `UnpicklingError: NEWOBJ class argument must be a type, not Whatever` → `BrokenProcessPool`. Such payloads work today (the worker re-imports the real package), so this is a regression in `run.cpu_bound`'s public contract. Data-bearing modules (`binding`, `observables`, `events`, dataclasses) would need to stay real — which pushes toward stubbing an explicit UI-only subset rather than stub-everything-by-default.

### Major

4. **The env marker outlives its purpose and stubs legitimate children** — after `ui.run()`, `NICEGUI_WORKER_STUBS=1` stays in the server process's environment forever, so *every* spawned child is stubbed — including a `mp.Process` that deliberately runs a second NiceGUI app/server, which then silently no-ops with no error anywhere. The set/pop/re-set dance across three sites in `ui_run.py` plus the pytest carve-out (which makes production stub behavior untestable via `Screen`) is a symptom of the marker living at the wrong scope. Consider injecting the marker only into workers NiceGUI itself creates (e.g. the `ProcessPoolExecutor` initializer in `run.setup()`), instead of mutating the ambient environment — that would delete both `pop` lines, the pytest special case, and the exposure for user-spawned processes.

5. **`REAL_MODULES` is an unenforced transitive-closure invariant, and it's already violated** — `helpers/functions.py` (helpers is "real" in workers) does `from ..awaitable_response import AwaitableResponse`; `awaitable_response` is stubbed, so `AwaitableResponse` is the `WHATEVER` *instance* and the `isinstance(result, (NoImplicitAwait, AwaitableResponse, asyncio.Task))` check raises `TypeError` when reached. `native/native_config.py` similarly binds `WHATEVER` for `..events.Handler`. Nothing asserts the allowlist is closed under imports, so any future `from .x import y` in a real module silently degrades workers. At minimum add a test that walks the imports of `REAL_MODULES` and fails on stubbed dependencies.

6. **`_PackageModule` can shrink to a single `__setattr__`** — the two copy-pasted property/setter/deleter triplets with hidden `_app_override`/`_context_override` slots exist only to absorb the import machinery's submodule binding. A `__setattr__` that ignores `ModuleType` assignments to `'app'`/`'context'` and otherwise defers to `super()` achieves the same: the machinery's binding is absorbed, lazy loading happens through the already-present (currently dead) `'app'`/`'context'` entries in `_LAZY_IMPORTS`, and `mock.patch` works with plain attributes. One method instead of six, no shadow slots.

### Minor

7. **Three lazy-import variants in one PR** — the package `__getattr__` is a verbatim copy of `ui.py`'s (#5303), while `json/__init__.py` uses a third shape with different caching. Consider extracting a shared helper (e.g. `_lazy.py`) used by all three; likewise the thrice-repeated `multiprocessing.current_process().name != 'MainProcess'` could become `helpers.is_main_process()`, and `install()`'s manual `os.scandir` loop is `pkgutil.iter_modules` in three lines.

8. **Smaller points** — (a) `json/response.py` re-runs the orjson-vs-builtin dispatch that `json/__init__.py` already performs and reaches into `orjson_wrapper`'s private `ORJSON_OPTS`/`_orjson_converter`; a public bytes-level `render()` selected once in `json/__init__.py` would avoid the divergence risk. (b) `NiceGUIJSONResponse` is no longer importable from `nicegui.json.orjson_wrapper`/`builtin_wrapper` — probably acceptable, but worth a mention in the changelog. (c) The `elapsed_ms < 500` assertion in `tests/test_worker_stubs.py` may be flaky on loaded CI runners. (d) The `_PackageModule` docstring doesn't follow CONTRIBUTING.md's "each sentence in documentation goes on a new line", and the generated driver in `tests/test_worker_stubs.py` uses `try/except SystemExit: pass` where the codebase prefers `contextlib.suppress`.

---

**Summary of the direction-check answer:** Commit 1 yes — clean it up (findings 6/7) and let's land it as its own PR. Commit 2 needs a rethink of the failure semantics before I'd trust it in the wild: fix the `Whatever` semantics (1, 2), keep data-bearing modules real or stub only UI modules (3), scope the marker to NiceGUI-created workers (4), and enforce the allowlist closure (5). Given how subtle 1 and 2 are — one of them cost me 132 GB of RAM to discover — I lean toward loud errors over silent no-ops as the default failure mode. Regarding your open questions: `NICEGUI_WORKER_STUBS` as a name is fine (though it may become an internal detail if the marker moves into the pool initializer), a `NICEGUI_EAGER_IMPORT` escape hatch can wait for a follow-up, and yes — keep the enabler/stubs split, ideally as two PRs.

evnchn:
*Drafted by Claude Code (assisting @evnchn); all findings below re-verified empirically on this machine before anything was changed.*

**TL;DR:**

Agreed on the split — and on the substance of every blocking finding.

1. **Commit 1 is now its own branch** with findings 6, 7 and 8a/8d applied (single `__setattr__`, shared `_lazy.py` resolver for all three lazy-import sites, public bytes-level `render()`), golden combo green, `import nicegui` still ~4 ms. Ready to open as a separate PR.
2. **Commit 2 is reworked** around the semantics you asked for: data-bearing modules stay **real** in workers (fixes findings 3 and 5 structurally, at a measured +0.3 ms), the `Whatever` dunders are fixed (findings 1 and 2, each pinned by a test that fails on the old code — the finding-1 test reproduces your 132 GB hang, bounded by a subprocess timeout), and the env marker is now scoped to the server's lifetime with the pytest carve-out deleted (finding 4, partially — see the pushback below).
3. **One technical pushback on finding 4:** the `ProcessPoolExecutor` initializer can't carry the marker — for spawn workers the env var must exist *before* the child bootstrap re-imports the user's main module, and the initializer runs *after* that. Also, the "deliberate second NiceGUI server in an `mp.Process` silently no-ops" scenario is pre-existing upstream behavior (`ui_run.py` returns early in any non-MainProcess since before this PR), not a regression of the marker.

Worker re-import with the reworked stubs: `examples/progress` 495 → ~25 ms, `menu_and_tabs` 328 → ~4.5 ms — a little above the stub-everything draft (14 / 1.7 ms) because the data modules are now real; that's the price of correct pickling and I think it's the right trade.

**Remaining open question for you at the bottom.**

<details>
<summary><b>Finding-by-finding status</b></summary>

| Finding | Status | What changed |
|---|---|---|
| 1 — infinite iteration | ✅ fixed + test | `__iter__` returns `iter(())`, `__contains__` returns `False`; the worker test now asserts `list(stub) == []` and `'key' not in stub`, and *times out* on the old code (your OOM, bounded) |
| 2 — truthy `WHATEVER` | ✅ fixed + test | `__bool__` returns `False`; test asserts `not core.app.is_stopping` in a stub worker, so `run.io_bound` executes again. A second-model review (Codex, adversarial pass) additionally caught that two unrelated stub values compared equal via singleton identity — `__eq__` now returns `False` (with `__hash__` explicitly preserved), also under test |
| 3 — pickle regression | ✅ structurally fixed + test, one caveat | `binding`/`observables`/`events`/`dataclasses`/`awaitable_response` (+ closure: `background_tasks`, `context`, `slot`) are now REAL in workers. Caveat: the specific example payloads turn out not to pickle even on main today (evidence below), so this was structural rather than a strict regression — fixed regardless |
| 4 — marker scope | ◐ partial + pushback | Marker now popped in a `try/finally` around `Server.instance.run()` (its purpose ends with the server); pytest carve-out **deleted** — stub behavior is now testable under `Screen`. Initializer idea doesn't work for spawn (below) |
| 5 — allowlist closure | ✅ fixed + enforced | `REAL_MODULES` expanded to the actual transitive closure (16 modules, still zero web-framework imports); a new test imports every real module in a fresh interpreter and fails if anything outside the set loads |
| 6 — `_PackageModule` | ✅ done | Six methods + two shadow slots → one `__setattr__` that absorbs `ModuleType` bindings for `app`/`context`; lazy loading now flows through the (previously dead) `_LAZY_IMPORTS` entries; `mock.patch` set/restore covered by the existing tests |
| 7 — three lazy variants | ✅ done | New `nicegui/_lazy.py::resolve()` used by the package, `ui.py`, and `nicegui.json`; `helpers.is_main_process()` replaces two of the three inline checks (the third, in `nicegui/__init__.py`, stays inline deliberately — importing `helpers` there would eagerly pull ~60 ms of real modules into the 4 ms package init); `install()` uses `pkgutil.iter_modules` |
| 8a — render dispatch | ✅ done | Both wrappers export a public `render()`; `json/__init__.py` selects it once; `response.py` no longer reaches into private wrapper internals |
| 8b — changelog note | ✅ will note | `NiceGUIJSONResponse` no longer importable from the wrapper modules — will be called out in the PR body |
| 8c — flaky `elapsed_ms` | ✅ dropped | The `< 500 ms` assert is gone; `'fastapi' not in sys.modules` *is* the perf signal, timing prints are informational |
| 8d — style nits | ✅ done | Docstring sentences on new lines; `contextlib.suppress(SystemExit)` in the test driver |

</details>

<details>
<summary><b>Why the pool-initializer can't carry the marker (finding 4 pushback, mechanism)</b></summary>

For a spawn-started pool worker the sequence is:

1. parent `Popen`s a fresh interpreter → child inherits **the parent's env at that moment**
2. child bootstrap runs `multiprocessing.spawn._main()` → `prepare(preparation_data)` → **re-imports the user's main module** ← this is where the 495 ms is paid and where stubs must already be active
3. only then is the process object unpickled and, for `ProcessPoolExecutor`, the `initializer` callback runs

So an `initializer=` marker arrives one step too late — the expensive re-import has already happened with the real package. Environment inheritance at step 1 is the only pre-bootstrap channel a parent has, which is why the marker stays an env var. What we *could* do is scope it tighter in time, which is what the rework does: it is now set while the server runs and popped in a `try/finally` when the server stops, so nothing leaks past shutdown (and between tests — which is also why the pytest carve-out could be deleted, making stub behavior observable in a `Screen` test for the first time).

On the "deliberate second NiceGUI app in an `mp.Process`" scenario: upstream `ui_run.py` already returns early from `ui.run()` in **any** non-MainProcess process (main branch, before this PR), so that child never started a server anyway — the marker doesn't regress it. What the marker adds is that such a child's *other* nicegui usage becomes stubs; with the data-bearing modules now real, the remaining stubbed surface is UI construction, which cannot work in a child process regardless.

</details>

<details>
<summary><b>Finding 3 caveat: the example payloads don't pickle on main either (evidence)</b></summary>

Two things I tried to build the regression test around, both failing **without** stubs on current main:

- `@binding.bindable_dataclass` instance as a spawn arg → parent-side
  `AttributeError: Can't get local object '_make_copyable.<locals>._pickle_function.<locals>.creator_with_hook'`
- `ObservableDict({'a': 1})` as a spawn arg → child-side
  `AttributeError: 'ObservableDict' object has no attribute '_change_handlers'` (dict-subclass unpickling calls `__setitem__` before `__init__`)

So "such payloads work today" holds for the *class-reference* case but not the instance case — instances of these types were never picklable. The regression test therefore pins the class-reference path: the parent sends the **class object** `Point` (a `bindable_dataclass`) to a stub-mode worker; resolving it against the re-imported `__mp_main__` must yield the real class. On the old stubs this fails (`Point` resolves to a `Whatever` instance); with `binding` real it passes. Either way the structural fix (keep data-bearing modules real + enforce closure) stands — the instance-pickling gaps above might be worth their own issue, independent of this PR.

</details>

<details>
<summary><b>Measurements (reworked stubs, M4 MacBook, Python 3.12, median-ish of several runs)</b></summary>

| Scenario | main | reworked stubs |
|---|---|---|
| `import nicegui` (bare) | ~223 ms | ~4 ms |
| `examples/progress` worker re-import | ~495 ms | ~25 ms |
| `examples/menu_and_tabs` worker re-import | ~328 ms | ~4.5 ms |
| REAL_MODULES expansion cost (data-bearing set, warm) | — | +0.3 ms |

The worker numbers are up from the stub-everything draft (14 / 1.7 ms) because `binding`, `events`, `observables` etc. are now genuinely imported in workers. Still 13–70× under main, and `run.cpu_bound` payloads resolve real classes.

Also verified: `from nicegui import App` in a stub worker now fails **loudly and immediately** (`ModuleNotFoundError: No module named 'nicegui.app.app'`) instead of hanging forever — documented as a known limitation in the module docstring.

</details>

<details>
<summary><b>Branch/PR mechanics</b></summary>

- Commit 1 lives on its own branch rebased onto current main (the `stop_and_exit()` conflict from the recent shutdown work is resolved there — `socketio` import becomes `TYPE_CHECKING`-only, everything else merges cleanly).
- Commit 2 is rebased on top of it as the stubs-only follow-up; this PR would become that follow-up once the first one lands.
- Golden combo (pre-commit, mypy over 246 files, pylint 10.00/10, the lazy-import + worker-stub + run + json/storage test files) green locally on both. Both diffs also went through an adversarial second-model review (Codex) — its one blocker (stub equality, see finding 2) is fixed; its review confirmed the lazy-import anchors and the marker set/pop paths clean. CI is the real judge — I'll wait for green before calling either ready.

</details>

**The remaining open question** — the residual silent surface. UI modules in workers still yield no-op values (falsy, empty, non-containing), because module-level `ui.label(...)` / `with ui.row():` in re-imported mains is exactly what must not raise. If you'd like more discoverability than the docstring, the cheapest dial I see is a one-time `log.debug(...)` (or `log.info(...)`) line when stubs install in a worker — visible when someone is debugging a worker that "does nothing", invisible otherwise. Happy to add it, leave it, or hear a third option.

evnchn:
**Branch force-pushed** (`1b9389b` → `0da1e11`, with lease): the head is now the reworked two-commit stack described in the reply above — commit 1 is the split-out enabler (staged for its own PR), commit 2 the reworked stubs. A range-diff would be noise here since the rework touches nearly every hunk; the finding-by-finding table in the previous comment is the accurate delta. Old head is preserved locally if anything needs comparing.

evnchn:
*Status note, posted by Claude Code on @evnchn's behalf.*

**Blocked by #6152.**

This PR's first commit is the lazy-package-import enabler, now split out and opened standalone as #6152 (rebased onto current `main`, one squashed commit, reviewed clean). This PR stays a draft until #6152 merges — at which point rebasing onto `main` drops the duplicated enabler commit, leaving just the worker-stubs change, and it comes out of draft for review. No action needed here until then.

evnchn:
https://github.com/zauberzeug/nicegui/pull/6152 in 3.16 so by extension this can also only be in 3.16

PR #6113, opened 2026-06-14
Improve AI-agent detection in Markdown content negotiation

### LINKED ISSUE #6109: Coding agents that are not claude code don't get served docs pages due to semantic mismatch
### First Check

### Example Code

```python
def _did_user_request_markdown(request: Request) -> bool:
    """Whether the request prefers a markdown response over HTML (page opt-in checked separately)."""
    accept = request.headers.get('accept', '').strip().lower()
    if 'text/html' in accept and 'text/markdown' not in accept:
        return False
    if 'text/markdown' in accept and 'text/html' not in accept:
        return True
    AI_AGENT_PATTERNS = [
        r'claude-?(bot|user|searchbot)',
        r'gptbot|oai-searchbot',
        r'chatgpt-user',
        r'perplexity(bot|-user)',
        r'google-(cloudvertexbot|agent)',
        r'gemini-deep-research',
    ]
    return bool(re.search('|'.join(AI_AGENT_PATTERNS), request.headers.get('user-agent', ''), re.IGNORECASE))
```

### Description

When an agent that is not claude code tries to pull the documentation via built in webfetch tools, they all will remark that they cannot retrieve it because of the way the website works. I believe this causes nicegui to get deranked in LLM preferences.

They then pull the HUGE llms.txt and get context overwhelmed, and start curling the appropriate docs.

I talked about this before in discussions, but one way that a modern library can succeed is showing up in the training data of LLMs. Being suggested as a good GUI framework (which I believe this is) will help the project.

I believe the logic in line 475 of clients.py should change.

Currently, the code checks for 'text/markdown' being alone in the headers, but modern coding agents do not request webpages this way, they almost always have html and markdown in the headers.

The semantic mismatch is that the docstring says it is checking if markdown is prefered, but it doesn't actually check if it's prefered or not, just if it is there alone.

A minimal change would be:

- Change the logic to `if markdown in headers` and remove the double if / if not check on 478-481, because browsers being used by a human do not have 'text/markdown' in the header requests
- Change line 480 to `if accept.startswith('text/markdown'), which is more defensive but arguably unneeded given the above.

If you really wanted to keep the current check logic, the startswith() is better imo. I don't think it's neccessary, [considering this article by vercel](https://vercel.com/blog/making-agent-friendly-pages-with-content-negotiation).

I would like to submit a pull request for either solution, because my agents struggle to consistently remember to use curl on the docs, and they are all steered toward a normal web fetch tool, which requests both html and markdown. Please let me know, this is a really simple change for a large benefit.

Thanks! Love the project

### NiceGUI Version

latest

### Python Version

3.13

### Browser

Chrome

### Operating System

Linux

### Additional Context

Opencode web fetch tool request headers:

`Accept: text/markdown;q=1.0, text/x-markdown;q=0.9, text/plain;q=0.8, text/html;q=0.7, */*;q=0.1
User-Agent (primary): Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36`

pi coding agent request headers:
`Accept: text/markdown,text/html,application/xhtml+xml,application/json;q=0.9,text/plain;q=0.8,*/*;q=0.5
User-Agent: pi-fetch-url/1.0 (+https://pi)`

jwinpbe:
### Motivation

Coding agents other than Claude Code often fail to fetch <https://nicegui.io> documentation as Markdown and waste several turns figuring out how to pull the docs. The previous content negotiation had two shortcomings:

- A known AI-agent user agent that sent `Accept: text/html` (as many `web_fetch` tools do) was served HTML: the old code returned HTML up front whenever `text/html` was present and `text/markdown` absent, short-circuiting **before** the user-agent check ever ran.
- An explicit `Accept: text/markdown` was **not** honored when the header also listed `text/html` (e.g. `text/markdown, text/html, */*`); it fell through to user-agent matching, so non-agent clients got HTML.

Closes #6109.

### Implementation

`_did_user_request_markdown` now decides in two steps:

1. If the `Accept` header contains `text/markdown`, return Markdown — regardless of any other `Accept` entries or the user agent.
2. Otherwise, fall back to user-agent detection against a module-level `AI_AGENT_TOKENS` tuple (case-insensitive substring match, replacing the previous regex).

Additional changes:

- Added `modelcontextprotocol` for MCP clients — the one genuinely new agent; every other agent (Claude, GPTBot, ChatGPT, Perplexity, Gemini, Google) was already matched by the previous regex.
- Dropped the now-unused `re` import and consolidated the tests into a single Accept × User-Agent matrix.

**Behavior change (intentional):** a request from a known agent user agent now receives Markdown **even when its `Accept` header is `text/html`**. Previously such a request received HTML. Agents frequently send a browser-like `Accept` header while still wanting Markdown, so this makes the fallback effective for them.

### Progress


falkoschindler (changes requested):
Thanks for the fix and the clear write-up — the root-cause analysis is correct. Coding agents send `Accept: text/markdown, text/html, */*`, which the old logic only honored when `text/html` was _absent_, so they fell through to the UA regex and missed unless their UA happened to match. Collapsing to "`text/markdown` present → serve Markdown" is the right call, and I'm happy to trade the UA sniffing away for the simpler, spec-aligned logic.

One thing to be explicit about: the UA fallback (#6053) targeted a population the Accept check can't reach — crawlers like `GPTBot`, `ChatGPT-User`, `Claude-User`, `PerplexityBot` that send a bare `*/*` (see the existing test rows pairing those UAs with `*/*` / `application/json`). After this change they get HTML instead of Markdown. I'm not certain whether all of those still send bare `*/*` today or have started advertising `text/markdown`, but I'm fine accepting that some of them get HTML — they can still fall back to `llms.txt`, and the simpler code is worth it. Just flagging it's a deliberate behavior change for that group, not an oversight.

A few things to tidy up before merge:

1. **`nicegui/llms.md` is now stale** — it still advertises the removed fallback in two places:
   - Line 8: "...as of NiceGUI 3.13 recognized AI-agent user-agents also receive Markdown when their `Accept` header is ambiguous (e.g. a wildcard `*/*`)..."
   - Line 795 config comment: `# (Accept: text/markdown, or recognized agent UAs); ...` — drop the "or recognized agent UAs" part.

   This file ships in the package and is exactly what agents fetch, so it should match the new behavior.

2. **Test name/columns** — `test_content_type_based_on_accept_and_user_agent` keeps a `user_agent` column that no longer affects any outcome (every `*/*` row now expects `text/html` regardless of UA). Consider renaming (drop `_and_user_agent`) and trimming the now-redundant rows; the ones you keep only document that UA is ignored.

3. **Docstring nit (non-blocking)** — "Whether the request prefers a markdown response over HTML" no longer reflects the check: it returns `True` on mere presence of `text/markdown`, ignoring q-values, so `Accept: text/html;q=1.0, text/markdown;q=0.1` would yield Markdown. Harmless in practice (browsers never send `text/markdown`), but a small tweak like "...when the request lists `text/markdown` in Accept" would keep it accurate.

jwinpbe:
I remember last time I looked at llms.md I wanted to check how many tokens it was, it's 11794 according to a counter I just used.

I frequently use less capable / smaller models than whatever anthropic has on their api, so all of my documentation / workflow is tuned for that, my instinct is it's too many tokens for what it is, and that the sheer amount of information packed into it is probably good for a high level summary / guide on how to use nicegui, but llms.md should be more focused rather than an infodump.

From what I understand of modern web content discovery, llms.md should be more of a summary of what nicegui is, with a few categorizer keywords, and a link to the documentation pages / examples / github (a lot of models i use instinctually try to pull nicegui/nicegui still), and llms.txt should be a documentation disambiguation.

I would include a specific block in the main nicegui.io/documentation .py to tell the LLM about the llms.txt if it hits the 'serve to 'text/markdown'' path on that page too.

The current llms.md could always be moved to a specific high level implementation document in that llms.txt.

I'll think about it and throw up a PR at some point. Thanks again!

evnchn:

**TL;DR:**

All three points from the review are addressed cleanly — including the non-blocking nit. Looks ready for a re-review.

One neutral heads-up: the `llms.md` edit also adds some new agent-usage guidance beyond the requested removal — possibly a nice win, just worth a look on its own merits.

@falkoschindler — over to you for the re-review when you have a slot.

<details>
<summary><b>Per-item verification (checked against the diff at <code>df1f776</code>, not the commit message)</b></summary>

| Review item | Status | What changed |
|---|---|---|
| 1. `llms.md` stale references (two spots) | ✅ | Line 8 dropped the "recognized AI-agent user-agents … ambiguous `*/*`" clause; line 795 `(Accept: text/markdown, or recognized agent UAs)` → `(Accept: text/markdown)`. |
| 2. Test name / columns | ✅ | `test_content_type_based_on_accept_and_user_agent` → `test_content_type_based_on_accept`; the now-meaningless `user_agent` parametrize column removed; rows trimmed 7 → 4; `User-Agent` dropped from the request. |
| 3. Docstring nit (non-blocking) | ✅ | `"Whether the request prefers a markdown response over HTML"` → `"Whether the request lists text/markdown in its Accept header"` — verbatim the suggested wording. |

Prompt turnaround, too (review 2026-06-15 → fix 2026-06-16). Nicely done, @jwinpbe.
</details>

<details>
<summary><b>The one thing worth a second look — <code>llms.md</code> additions (flagging, not faulting)</b></summary>

Beyond removing the stale clause, the line-8 edit adds new content: a `web_fetch` hint plus example doc URLs (`…/section_text_elements`, `/label`, `/section_testing`, `/user`). It could genuinely help agents navigate the docs — it's just additive content that wasn't part of the original review, so it deserves a look on its own merits rather than riding in under "address feedback".

This also dovetails with @jwinpbe's separate point that `llms.md` is ~11.8k tokens and may be heavier than it needs to be — so there's a fair question of whether the examples earn their tokens. No strong opinion either way; both directions are reasonable.
</details>

Minor, unrelated to the review: the branch is currently **behind `main`** (`mergeStateStatus: BEHIND`), so it'll want an update before merge.

jwinpbe:
For reference:

Screenshot_20260616_201516

Qwen 3.6 27B is extremely capable, nicegui.io's llms.txt just isnt in the format it's expecting. It ideally would point towards a documentation disambuguation.

I'll work on it in my spare time. NiceGUI will benefit from being accessible to agents from this size to open weight models running through opencode, pi, etc. The easier it is to access, the more likely that nicegui shows up in training data eventually

jwinpbe:
@falkoschindler let me know if you want me to change anything before you merge. This should be g2g! :heart:

evnchn:
*Posted by Claude Code on @evnchn's behalf. It's the Dragon Boat Festival (端午節) holiday in Hong Kong — Evan hasn't read this write-up line-by-line, so treat the **prose** as agent-drafted and sanity-check it. The **empirical half is the real deal**, though: the header captures below come from actually pointing each agent's web-fetch at a live instrumented HTTPS endpoint today and reading the server-side request log — not a model guessing. Trust the data; double-check the words.*

**TL;DR — net positive, but not the "pure win" the premise assumes; I measured a regression on real agents.**

I empirically captured the **live** `Accept`/`User-Agent` of three coding agents today (2026-06-19) plus Claude.ai web, and ran the exact headers through `main`, this PR, and a hybrid. Result:

- ✅ This PR **fixes OpenCode** (sends `text/markdown` in a multi-value `Accept`; `main` wrongly gave it HTML). Real win, exactly the #6109 case.
- ❌ But it **does not fix Codex** — OpenAI's flagship non-Claude agent fetches via its `ChatGPT-User` infra with a **browser `Accept` containing no `text/markdown` at all**. The PR leaves it on HTML *and* deletes the UA path that was its only route to Markdown.
- ❌ And it **regresses** Claude.ai web + the self-identifying crawler set (`GPTBot`/`ChatGPT-User`/`PerplexityBot`/…) from Markdown → HTML.

So the PR's stated premise — *"no AI coding agent is known to omit `text/markdown`"* — is **empirically false** (Codex omits it). The good news: a one-line-bigger version is a strict superset with zero regressions.

<details>
<summary><b>Empirical truth table — real captured headers, 2026-06-19</b></summary>

Captured by pointing each agent's web-fetch at an instrumented endpoint over HTTPS and reading the server-side request log, then feeding the **exact** captured header strings into `_did_user_request_markdown` on each branch.

| Agent (real headers) | `main` | this PR | hybrid |
|---|:--:|:--:|:--:|
| **Claude Code** v2.1.183 — `Accept: text/markdown, text/html, */*` | MD | MD | MD |
| **OpenCode** — `Accept: text/markdown;q=1.0, …, text/html;q=0.7` · UA: *spoofed Chrome* | **html** | **MD** ✅ | MD |
| **Codex** — `Accept: text/html,…,*/*;q=0.8` (no markdown) · UA: `ChatGPT-User/1.0` | html | **html** ❌ | **MD** |
| **Claude.ai web** — `Accept: */*` · UA: `Claude-User/1.0` | MD | **html** ❌ | MD |
| Real browser / `curl */*` | html | html | html |

Two things the captures expose that the Accept-only model can't see:
1. **OpenCode spoofs a Chrome UA** — so UA-sniffing alone never could have served it; the Accept check is what saves it. (Argues the Accept check should be *primary*.)
2. **Codex carries no `text/markdown`** but a truthful `ChatGPT-User` UA — so Accept-only never serves it; the UA fallback is what saves it. (Argues the UA fallback should be *retained*.)

Neither mechanism alone covers both. That's the whole point.
</details>

<details>
<summary><b>The pure-positive fix (hybrid) — keep the simplified Accept check AND the UA fallback</b></summary>

Keep this PR's `Accept` simplification (it's correct and spec-aligned — it's what fixes OpenCode), but **don't delete the UA fallback**; demote it to a last resort after the Accept check:

```python
def _did_user_request_markdown(request: Request) -> bool:
    """Whether the request prefers a markdown response over HTML (page opt-in checked separately)."""
    accept = request.headers.get('accept', '').strip().lower()
    if 'text/markdown' in accept:
        return True
    AI_AGENT_PATTERNS = [
        r'claude-?(bot|user|searchbot)',
        r'gptbot|oai-searchbot',
        r'chatgpt-user',
        r'perplexity(bot|-user)',
        r'google-(cloudvertexbot|agent)',
        r'gemini-deep-research',
    ]
    return bool(re.search('|'.join(AI_AGENT_PATTERNS), request.headers.get('user-agent', ''), re.IGNORECASE))
```

This is a strict superset of the markdown-served population of **both** `main` and this PR (the "hybrid" column above): every agent gets Markdown, humans/curl get HTML. Verified — passes the PR's reduced `test_markdown_response.py` (37 passed); the UA rows from `main`'s test should be kept to cover the fallback.
</details>

<details>
<summary><b>Honest cost of the hybrid (adversarial pass) — what you pay for "no regressions"</b></summary>

I ran an adversarial review against the hybrid. Of the five objections to UA-based negotiation, four are **neutralized by this codebase**, one is real:

- ~~**Caching / `Vary` fragmentation**~~ — neutralized: the Markdown response is `Cache-Control: no-store` (`client.py:170`), NiceGUI pages are dynamic, and the code sets **no `Vary` header today** on either branch. No CDN fragmentation to worry about.
- ~~**Spoofing exposes source**~~ — non-issue: the "secret" is public docs Markdown (same content as the public `llms.txt`).
- ~~**Privacy / fingerprinting**~~ — the server already logs UA; reading it changes nothing.
- ~~**Correctness (identity ≠ preference)**~~ — true in the abstract, but for the bare-`*/*` population there is *no* `Accept` signal to honor; UA is the only available lever, and it's a fallback, not the primary path.
- ✅ **Maintenance** — real: the UA regex rots as bots rename. But it already exists (#6053) and is ~12 lines.

So the honest trade is **~12 lines of UA-regex upkeep ⇄ Markdown for Codex + Claude.ai web + every self-identifying crawler.** Dropping it is a legitimate simplicity call — I just want it made with the measured price visible: it's not only obscure crawlers, it's **Codex**, one of the two biggest non-Claude agents #6109 set out to serve.
</details>

**Ask — @rodja @falkoschindler.** @rodja especially, since this is squarely agent-fetch behavior and you're closest to the agentic side: the call is just **simplicity vs. coverage**. If "pure positive, no regressions" is the goal, the hybrid (Accept-first, UA fallback retained) serves Codex + Claude.ai web + crawlers Markdown for ~12 extra lines. If shedding the UA list is worth more, this PR is a fine deliberate trade — just worth a changelog note that those populations now get HTML and lean on `llms.txt`. Either way the data above is the part to weigh; the recommendation is negotiable.

jwinpbe:
got it, i'm going to switch to an any() based approach because it's faster than regex. Give me a bit to check it out.

<details>
<summary>Benchmark script that you can `uv run`</summary>

```python
# /// script
# requires-python = ">=3.10"
# dependencies = ["rich"]
# ///

import re
import time
from dataclasses import dataclass

from rich.console import Console
from rich.table import Table
from rich.text import Text

console = Console()


# ---------------------------------------------------------------------------
# CONTEXT
# ---------------------------------------------------------------------------
#
# Purpose: benchmark regex vs any() for AI-agent UA detection in NiceGUI's
# markdown content negotiation (_did_user_request_markdown in client.py).
#
# Background:
#   • Upstream (upstream/main) uses regex-based AI-agent UA detection with
#     `AI_AGENT_PATTERNS` as a fallback
#   • Our branch (`fix/agent-markdown-web-fetch`) removes the UA fallback, only
#     checking `Accept: text/markdown`. The regex is preserved here for comparison
#     — byte-identical to upstream implementation.
#
# Two implementations:
#   regex — builds `'|'.join(AI_AGENT_PATTERNS)` at runtime, then re.search().
#           Matches the original removed code exactly.
#   any()  — checks 12 literal tokens with `any(token in ua)` against lowered UA.
#
# Key findings (consistent across all 4 traffic mixes, ~1000 requests, 10 trials):
#   • any() is 3-6x faster than regex (e.g. 1.7µs vs 9.3µs per call, balanced mix)
#   • Stddev is negligible — results are statistically solid
#   • Regex pays per-call overhead for re.compile + pattern matching
#   • any() pays per-call overhead for list iteration but avoids regex engine
#
# Traffic mixes model realistic scenarios:
#   Agent-heavy  — 50% short-circuit on Accept (coding agents pull docs)
#   Human-heavy  — 88% UA fallback (typical website traffic)
#   Bot-heavy    — 84% UA fallback (crawler/search engine traffic)
#   Balanced     — 65% UA fallback (mixed production traffic)
#
# The benchmark is self-contained: run with `uv run benchmark.py`.
# ---------------------------------------------------------------------------


# ---------------------------------------------------------------------------
# Traffic mixes — 4 distinct real-world scenarios
# ---------------------------------------------------------------------------


@dataclass(frozen=True)
class Request:
    accept: str
    user_agent: str


def _make_requests(mix: list[tuple[float, str, str]]) -> list[Request]:
    """Expand weighted traffic into a flat list (~1000 requests)."""
    requests: list[Request] = []
    for weight, accept, ua in mix:
        count = max(1, round(weight * 1000))
        requests.extend([Request(accept, ua)] * count)
    return requests


# Scenario 1: Agent-heavy — documentation pull from coding agents
MIX_AGENT_HEAVY: list[tuple[float, str, str]] = [
    # Agents with text/markdown in Accept (50%)
    (
        0.20,
        "text/markdown, text/html, */*",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko; compatible; Claude-User/1.0; +claude-user@anthropic.com) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.15,
        "text/markdown;q=1.0, text/x-markdown;q=0.9, text/plain;q=0.8, text/html;q=0.7, */*;q=0.1",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.15,
        "text/markdown,text/html,application/xhtml+xml,application/json;q=0.9,text/plain;q=0.5,*/*;q=0.5",
        "pi-fetch-url/1.0 (+https://pi)",
    ),
    # Agents without text/markdown (30%)
    (0.15, "text/html, application/xhtml+xml, */*;q=0.8", "ChatGPT-User/3.0"),
    (0.08, "*/*", "Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Claude-User/1.0)"),
    (0.07, "*/*", "GPTBot/1.0 (+https://openai.com/gptbot)"),
    # Human browsers (20%)
    (
        0.12,
        "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.08,
        "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
        "Mozilla/5.0 (Macintosh; Intel Mac OS X 14_4) AppleWebKit/605.1.15 "
        "(KHTML, like Gecko) Version/18.3 Safari/605.1.15",
    ),
]

# Scenario 2: Human-heavy — typical website traffic
MIX_HUMAN_HEAVY: list[tuple[float, str, str]] = [
    # Agents with text/markdown in Accept (10%)
    (
        0.05,
        "text/markdown, text/html, */*",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko; compatible; Claude-User/1.0; +claude-user@anthropic.com) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.05,
        "text/markdown;q=1.0, text/x-markdown;q=0.9, text/plain;q=0.8, text/html;q=0.7, */*;q=0.1",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    # Agents without text/markdown (5%)
    (0.03, "text/html, application/xhtml+xml, */*;q=0.8", "ChatGPT-User/3.0"),
    (0.02, "*/*", "Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Claude-User/1.0)"),
    # Human browsers (75%)
    (
        0.30,
        "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.15,
        "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
        "Mozilla/5.0 (Macintosh; Intel Mac OS X 14_4) AppleWebKit/605.1.15 "
        "(KHTML, like Gecko) Version/18.3 Safari/605.1.15",
    ),
    (
        0.10,
        "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8",
        "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/143.0",
    ),
    (
        0.05,
        "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36 Edg/143.0.0.0",
    ),
    # curl / script (10%)
    (0.05, "*/*", "curl/8.14.1"),
    (0.05, "*/*", "python-requests/2.32.3"),
]

# Scenario 3: Bot-heavy — crawler/search engine traffic
MIX_BOT_HEAVY: list[tuple[float, str, str]] = [
    # Agents with text/markdown in Accept (15%)
    (
        0.10,
        "text/markdown, text/html, */*",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko; compatible; Claude-User/1.0; +claude-user@anthropic.com) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.05,
        "text/markdown;q=1.0, text/x-markdown;q=0.9, text/plain;q=0.8, text/html;q=0.7, */*;q=0.1",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    # Bots without text/markdown (55%)
    (0.15, "text/html, application/xhtml+xml, */*;q=0.8", "ChatGPT-User/3.0"),
    (0.10, "*/*", "GPTBot/1.0 (+https://openai.com/gptbot)"),
    (0.08, "*/*", "PerplexityBot/1.0"),
    (0.07, "*/*", "Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Claude-User/1.0)"),
    (0.05, "*/*", "google-cloudvertexbot/1.0"),
    (0.05, "*/*", "google-agent/1.0"),
    (0.05, "*/*", "Gemini-Deep-Research/1.0"),
    (0.05, "*/*", "OAI-SearchBot/1.0"),
    # Human browsers (20%)
    (
        0.12,
        "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.08,
        "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
        "Mozilla/5.0 (Macintosh; Intel Mac OS X 14_4) AppleWebKit/605.1.15 "
        "(KHTML, like Gecko) Version/18.3 Safari/605.1.15",
    ),
]

# Scenario 4: Balanced — mixed production traffic
MIX_BALANCED: list[tuple[float, str, str]] = [
    # Agents with text/markdown in Accept (33%)
    (
        0.12,
        "text/markdown, text/html, */*",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko; compatible; Claude-User/1.0; +claude-user@anthropic.com) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.10,
        "text/markdown;q=1.0, text/x-markdown;q=0.9, text/plain;q=0.8, text/html;q=0.7, */*;q=0.1",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.11,
        "text/markdown,text/html,application/xhtml+xml,application/json;q=0.9,text/plain;q=0.5,*/*;q=0.5",
        "pi-fetch-url/1.0 (+https://pi)",
    ),
    # Agents without text/markdown (34%)
    (0.10, "text/html, application/xhtml+xml, */*;q=0.8", "ChatGPT-User/3.0"),
    (0.05, "*/*", "Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Claude-User/1.0)"),
    (0.03, "*/*", "GPTBot/1.0 (+https://openai.com/gptbot)"),
    (0.02, "*/*", "PerplexityBot/1.0"),
    (0.05, "*/*", "google-cloudvertexbot/1.0"),
    (0.04, "*/*", "google-agent/1.0"),
    # Human browsers (25%)
    (
        0.15,
        "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 "
        "(KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
    ),
    (
        0.07,
        "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
        "Mozilla/5.0 (Macintosh; Intel Mac OS X 14_4) AppleWebKit/605.1.15 "
        "(KHTML, like Gecko) Version/18.3 Safari/605.1.15",
    ),
    (
        0.03,
        "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8",
        "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/143.0",
    ),
    # curl / script (8%)
    (0.04, "*/*", "curl/8.14.1"),
    (0.04, "*/*", "python-requests/2.32.3"),
]

MIXES: dict[str, list[tuple[float, str, str]]] = {
    "Agent-heavy": MIX_AGENT_HEAVY,
    "Human-heavy": MIX_HUMAN_HEAVY,
    "Bot-heavy": MIX_BOT_HEAVY,
    "Balanced": MIX_BALANCED,
}


# ---------------------------------------------------------------------------
# Two implementations under test
# ---------------------------------------------------------------------------


def _did_user_request_markdown_regex(accept: str, user_agent: str) -> bool:
    accept = accept.strip().lower()
    if "text/markdown" in accept:
        return True
    AI_AGENT_PATTERNS = [
        r"claude-?(bot|user|searchbot)",
        r"gptbot|oai-searchbot",
        r"chatgpt-user",
        r"perplexity(bot|-user)",
        r"google-(cloudvertexbot|agent)",
        r"gemini-deep-research",
    ]
    return bool(re.search("|".join(AI_AGENT_PATTERNS), user_agent, re.IGNORECASE))


def _did_user_request_markdown_any(accept: str, user_agent: str) -> bool:
    accept = accept.strip().lower()
    if "text/markdown" in accept:
        return True
    ua = user_agent.lower()
    return any(
        token in ua
        for token in (
            "claude-user",
            "claude-bot",
            "claude-searchbot",
            "gptbot",
            "oai-searchbot",
            "chatgpt-user",
            "perplexitybot",
            "perplexity-user",
            "google-cloudvertexbot",
            "google-agent",
            "gemini-deep-research",
        )
    )


# ---------------------------------------------------------------------------
# Correctness check
# ---------------------------------------------------------------------------

_EXPECTED: dict[tuple[str, str], bool] = {
    ("text/markdown, text/html, */*", "Claude-User/1.0"): True,
    ("text/markdown;q=1.0, text/html;q=0.7", "Chrome/143"): True,
    ("text/markdown,text/html", "pi-fetch-url"): True,
    ("text/html, */*;q=0.8", "ChatGPT-User/3.0"): True,
    ("*/*", "Claude-User/1.0"): True,
    ("*/*", "GPTBot/1.0"): True,
    ("*/*", "PerplexityBot/1.0"): True,
    ("text/html, */*", "Chrome/143"): False,
    ("*/*", "Safari/605"): False,
    ("*/*", "Firefox/143"): False,
    ("*/*", "curl/8.14.1"): False,
    ("*/*", "python-requests/2.32"): False,
}


def _verify_correctness() -> bool:
    errors = []
    for (accept, ua), expected in _EXPECTED.items():
        for name, fn in [("regex", _did_user_request_markdown_regex), ("any", _did_user_request_markdown_any)]:
            result = fn(accept, ua)
            if result != expected:
                errors.append(f"  {name}: Accept={accept!r} UA={ua!r} -> {result} (expected {expected})")
    if errors:
        console.print("\n[red]Correctness failures:[/red]")
        for e in errors:
            console.print(e)
        return False
    console.print("[green]All correctness checks passed[/green]")
    return True


# ---------------------------------------------------------------------------
# Benchmark
# ---------------------------------------------------------------------------


def _benchmark(fn, requests: list[Request], trials: int = 10) -> dict:
    # Warm-up
    for req in requests[:50]:
        fn(req.accept, req.user_agent)

    elapsed_per_trial = []
    for _ in range(trials):
        start = time.perf_counter()
        for req in requests:
            fn(req.accept, req.user_agent)
        elapsed_per_trial.append(time.perf_counter() - start)

    elapsed = sum(elapsed_per_trial) / trials
    stddev = (sum((e - elapsed) ** 2 for e in elapsed_per_trial) / trials) ** 0.5
    per_call = elapsed / len(requests) * 1_000_000
    stddev_per_call = stddev / len(requests) * 1_000_000
    throughput = len(requests) / elapsed

    ua_hits = sum(1 for r in requests if "text/markdown" not in r.accept.lower().strip())
    ua_hit_rate = ua_hits / len(requests)

    return {
        "elapsed": elapsed,
        "per_call": per_call,
        "stddev_per_call": stddev_per_call,
        "throughput": throughput,
        "ua_hit_rate": ua_hit_rate,
    }


def _print_results(results: list[dict], labels: list[str]) -> None:
    table = Table(show_header=True, header_style="bold")
    table.add_column("Metric", style="dim")
    for i, label in enumerate(labels):
        style = "bold" if i == 1 else None  # any() column is bold
        table.add_column(label, justify="right", style=style)
    table.add_column("Delta", justify="right")
    table.add_column("Speedup", justify="right")

    metrics = [
        ("per_call", "Per call", "{:.2f}"),
        ("throughput", "Throughput", "{:,.0f}/s"),
        ("ua_hit_rate", "UA fallback hit rate", "{:.1%}"),
    ]

    for key, label, fmt in metrics:
        vals = [r[key] for r in results]
        delta = vals[0] - vals[1]
        if key == "per_call":
            sd0, sd1 = results[0]["stddev_per_call"], results[1]["stddev_per_call"]
            row = [f"{vals[0]:.2f}±{sd0:.1f}µs", f"{vals[1]:.2f}±{sd1:.1f}µs"]
            delta_str = f"{delta:+.2f}µs"
            speedup = vals[0] / vals[1] if vals[1] > 0 else 0
            speedup_str = f"{speedup:.1f}× faster"
        elif key == "throughput":
            row = [fmt.format(v) for v in vals]
            delta_str = f"{delta:+,.0f}/s"
            speedup = vals[1] / vals[0] if vals[0] > 0 else 0
            speedup_str = f"{speedup:.1f}× more throughput"
        else:
            row = [fmt.format(v) for v in vals]
            delta_str = f"{delta:+.1%}"
            speedup_str = "—"

        color = "green" if (key == "throughput" and delta > 0) or (key == "per_call" and delta < 0) else "red"
        table.add_row(label, *row, Text(delta_str, style=color), Text(speedup_str, style="green"))

    console.print(table)


def _print_summary(all_results: list[tuple[str, list[dict]]]) -> None:
    """Print a consolidated summary table across all traffic mixes."""
    table = Table(show_header=True, header_style="bold")
    table.add_column("Mix", style="dim")
    table.add_column("Regex", justify="right")
    table.add_column("any()", justify="right")
    table.add_column("Delta", justify="right")
    table.add_column("Speedup", justify="right")

    for mix_name, results in all_results:
        vals = [r["per_call"] for r in results]
        sd0, sd1 = results[0]["stddev_per_call"], results[1]["stddev_per_call"]
        row = [
            mix_name,
            f"{vals[0]:.2f}±{sd0:.1f}µs",
            f"{vals[1]:.2f}±{sd1:.1f}µs",
        ]
        delta = vals[0] - vals[1]
        speedup = vals[0] / vals[1] if vals[1] > 0 else 0
        row.append(f"{delta:+.2f}µs")
        row.append(f"{speedup:.1f}× faster")

        color = "green" if delta < 0 else "red"
        table.add_row(*row[:3], Text(row[3], style=color), Text(row[4], style="green"))

    console.print("\n[bold]Summary[/bold]")
    console.print(table)

    overall_speedup = sum(r[0]["per_call"] for _, r in all_results) / sum(r[1]["per_call"] for _, r in all_results)
    console.print(f"\n[bold green]Result:[/bold green] any() is {overall_speedup:.1f}× faster than regex (mean across all mixes)")
    console.print()


def main() -> None:
    console.print("[bold]UA detection: regex vs any()[/bold]\n")

    if not _verify_correctness():
        console.print("\n[red]Aborting — correctness failed[/red]")
        return

    all_results: list[tuple[str, list[dict]]] = []

    for mix_name, mix in MIXES.items():
        requests = _make_requests(mix)
        console.print(f"[bold]{mix_name}[/bold]  ({len(requests)} requests)")
        results = [
            _benchmark(_did_user_request_markdown_regex, requests),
            _benchmark(_did_user_request_markdown_any, requests),
        ]
        _print_results(results, ["Regex", "any()"])
        all_results.append((mix_name, results))

    _print_summary(all_results)


if __name__ == "__main__":
    main()
```
</details>

jwinpbe:
@evnchn hope you're enjoying the holiday! Thanks for letting me know about the regression.

Feel free to run the previous benchmark script and review the substring matching.

I moved it out of the function so it wouldn't be recreated every time it's called. This should be easier to maintain, you just edit the constant at the top of the file instead of editing the regex.

I added 'modelcontextprotocol' because I found that the anthropic `fetch` mcp uses it as a user agent, and because it is one of the first reference MCPs created, it's undoubtedly still kicking around in a lot of people's tools.

It's 4-6x faster on average to do these string ops in python's C than to compile a regex on the spot for every invocation. Even if its miniscule in the grand scheme of things, I'd rather it be more maintainable than the previous regex check. A tuple comprehension is way more readable too.

I looked around for edge cases related to the chosen strings in the tuple and couldn't find any. If you want to get more granular and add in exact shapes of particular claude clients I can but its extremely unlikely to effect anything in my judgement, only anthropic agent clients are sending 'claude' anything in that field.

evnchn:
*Posted by Claude Code on @evnchn's behalf (still 端午節 here — Evan skimmed rather than deep-read, but the re-verification below was actually run).*

@jwinpbe lovely turnaround — you didn't just patch the regression, you improved on the suggestion. `any()` substring matching over the compiled regex is both faster *and* far more maintainable (edit one tuple, no regex literacy required), and catching `modelcontextprotocol` for Anthropic's `fetch` MCP covers an agent I hadn't even captured. Both are strict upgrades over what was proposed.

I re-ran the real captured headers against the new head (`f1c7fb9`):

| Client (real headers) | result |
|---|:--:|
| Claude Code · OpenCode · **Codex** (`ChatGPT-User`) · **Claude.ai web** | **Markdown** ✅ |
| Real browser · `curl */*` | html ✅ |
| MCP `fetch` | Markdown ✅ |

…and the suite is green (**48 passed**, up from 37 with the UA coverage restored). So the trade is gone — it's now a clean superset of `main` with no regressions I can find. Nothing blocking from my side.

Nice work on this. 🙏

falkoschindler (approved):
Thanks again for tackling this — getting the docs to serve Markdown to more agents is a nice win. 🙏

The core change is the right call: honoring an explicit `Accept: text/markdown` first, then falling back to user-agent detection, and replacing the regex with a simple substring match against `AI_AGENT_TOKENS` reads much more clearly.

I pushed a few small follow-ups on top:

- **Docstring** — reworded `_did_user_request_markdown` to describe the two-step behavior ("Accept header or known agent user agent").
- **Claude tokens** — replaced the broad `claude` substring with the specific tokens `claudebot`, `claude-user`, `claude-searchbot`, `claude-code`, matching the precision of the other entries.
- **Tests** — consolidated the three new functions back into a single `accept,user_agent,expected` matrix (drops the triplicated page boilerplate), and added a case pinning the intended precedence change: a known agent UA now gets Markdown even with `Accept: text/html`.

One thing worth keeping in mind (now noted in the PR body): that last point is a deliberate behavior change — previously an agent sending `Accept: text/html` received HTML, now it receives Markdown. That's exactly what we want for agents that send browser-like headers.

LGTM — approving. ✅

PR #6116, opened 2026-06-16
Fix ui.select(with_input) dropping its html_id (breaks tooltips)

### LINKED ISSUE #6114: ui.select with tooltip
### First Check

### Example Code

```python
from nicegui import ui

ui.select(label="Version A", options=["AAA", "BBB", "CCC"]).classes(
    "w-64 mb-8"
).tooltip("Version A")

ui.select(label="Version B", options=["AAA", "BBB", "CCC"], with_input=True).classes(
    "w-64 mb-8"
).tooltip("Version B")

with ui.element("div").tooltip("Version C"):
    ui.select(
        label="Version C", options=["AAA", "BBB", "CCC"], with_input=True
    ).classes("w-64 mb-8")


ui.run()
```

### Description

In the example code, "tooltip" works for "Version A" and "Version C", but not for "Version B".

### NiceGUI Version

3.13.0

### Python Version

3.13.13

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

evnchn:

### Motivation

Fixes #6114.

`ui.select(with_input=True)` doesn't show a tooltip added via `.tooltip()`. The browser console shows `Anchor: target "#cN" not found`, and `ui.query(f'#{select.html_id}')` can't target the element either.

**Root cause:** `.tooltip()` anchors its `QTooltip` with `target="#{html_id}"`, relying on the element's `id` landing on a DOM node. In `use-input` mode, Quasar's `QSelect` assigns the inner `<input>` a generated id (its `targetUid`, e.g. `f_<uuid>`), which displaces NiceGUI's fall-through `id`. As a result no DOM element carries `cN`, so the tooltip anchor (and `ui.query('#...')`) resolve to nothing.

Verified against the rendered DOM: a plain `ui.select` keeps its id on the persistent `q-field__native` div, and `ui.input` / `ui.number` / `ui.textarea` are unaffected — **only `with_input` selects** drop the id.

### Implementation

Set Quasar's `for` prop to the element's `html_id` in the `with_input` branch. `for` is Quasar's documented way to set the inner control's id, so the input's id becomes `html_id` again and the anchor target exists:

```python
self._props['for'] = self.html_id
```

### Progress

- [x] Pytests have been added (`test_with_input_select_exposes_html_id`, a `Screen` test asserting the html_id is present in the DOM).
- [x] Documentation has been added/updated or is not necessary. (bug fix — no doc change)

---

Thanks to @rolfn for the clear report and to @python-and-novella for pinpointing the lost id and the `for`-prop workaround this builds on.

### DIFF
--- a/nicegui/elements/select.py
+++ b/nicegui/elements/select.py
@@ -79,6 +79,7 @@ class Select(LabelElement, ValidationElement[Any], ChoiceElement, DisableableEle
             self._props.set_bool('hide-selected', not multiple)
             self._props['fill-input'] = True
             self._props['input-debounce'] = 0
+            self._props['for'] = self.html_id  # keep html_id on the input so tooltips/ui.query can anchor (#6114)
         self._props.set_bool('multiple', multiple)
         self._props.set_bool('clearable', clearable)
 
--- a/tests/test_element.py
+++ b/tests/test_element.py
@@ -471,6 +471,7 @@ def test_even_special_elements_have_an_html_id(screen: Screen):
             ui.time_input(),
             ui.color_input(),
             ui.select([]),
+            ui.select([], with_input=True),  # #6114: with_input select dropped its html_id
             ui.input_chips(),
             ui.toggle([]),
             ui.radio([]),
--- a/tests/test_select.py
+++ b/tests/test_select.py
@@ -326,18 +326,3 @@ def test_popup_scroll_behavior(screen: Screen):
     screen.type(Keys.ESCAPE)
     screen.wait(0.2)
     assert screen.selenium.execute_script('return window.scrollY') == position
-
-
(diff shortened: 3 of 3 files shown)


falkoschindler (approved):
LGTM — the fix correctly restores the `html_id` on the rendered DOM element for `with_input` selects, which is the root cause of #6114 (tooltips and `ui.query` losing their anchor).

I pushed a small follow-up to the tests: instead of a standalone `test_with_input_select_exposes_html_id`, the `with_input=True` case is now folded into the existing `test_even_special_elements_have_an_html_id`, which is the canonical guard for "every special element exposes an `html_id` on a real DOM node." Net −15/+1 lines, same coverage, in the place future readers will expect it.

Thanks for the fix!

PR #6117, opened 2026-06-16
Make the start method of the cpu_bound process pool configurable

evnchn:
### Motivation

`nicegui.run`'s process pool is a bare `ProcessPoolExecutor`, so it inherits the **global** multiprocessing start method — `fork` on Linux up to Python 3.13 (e.g. in Docker), which is [unsafe for Python](https://pythonspeed.com/articles/python-multiprocessing/): a multi-threaded parent forking can deadlock the child. So `run.cpu_bound` silently runs under `fork` whenever the global default isn't `spawn`. Surfaced by zauberzeug/rosys#19.

This can't be fixed by setting the global start method on import — that path was removed because it breaks downstream code (#3644). The supported mechanism is a scoped `mp_context`, which `native.py` already uses for the window subprocess (#1841). But flipping Linux to `spawn` unconditionally would break currently-working apps that rely on fork-inherited process state, so the switch is phased instead of forced (see review discussion).

### Implementation

- New setting `run.process_pool_start_method: Literal['spawn', 'fork', 'forkserver'] | None = None`, to be set before `ui.run()`. `None` keeps the platform default; any explicit value is passed to `multiprocessing.get_context()`, with a fail-fast `ValueError` at startup for unknown or platform-unsupported values.
- The context is resolved **once** in `setup()` and latched in `_pool_context`, so the live pool, the heads-up warning, and the `BrokenProcessPool` restart path all agree even if the setting is mutated after startup.
- If the setting is `None` and the platform default resolves to plain `fork`, the first `cpu_bound` call logs a **one-time heads-up** (via `helpers.warn_once`) explaining the risk and the opt-ins. Explicit `'fork'` keeps fork silently; explicit `'spawn'` opts into the safe method today.
- **NiceGUI 4.0 will flip the default to `'spawn'`** (marked DEPRECATED in the code); `'fork'` remains available as an opt-out for as long as Python supports it.
- `SPAWN_CONTEXT` moved from `native.py` to `run.py` and is shared with the native window subprocess (#1841) — one spawn context instead of two.

**Behavior note:** nothing changes by default in 3.x — Linux apps keep fork and merely see the one-time warning. Opting into `'spawn'` requires picklable `cpu_bound` callables and args (already the contract on macOS/Windows), and spawned workers start from a fresh import, so they don't see module globals mutated at runtime by the parent.

**Tests:** the configured start method is verified through the pool's observable context for all four settings, plus the shared-context identity with `native.py`, the fail-fast on invalid values, the warning firing iff the pool implicitly falls back to fork, and the latch semantics (flipping the setting after startup neither changes the pool nor silences the warning).

<details>
<summary><b>Repro from the original report (Linux)</b></summary>

The MRE below demonstrates the problem: on `main` (and on this branch with the default setting), `cpu_bound` runs under `fork` on Linux and the child inherits parent memory. With `run.process_pool_start_method = 'spawn'` set before `run.setup()`, it prints "OK: cpu_bound ran under spawn".

```python
"""Repro for nicegui#182 - run.cpu_bound runs under 'fork' on Linux by default.

Run on Linux (default start method = fork; e.g. inside Docker):

    python mre182.py

Why this proves it: a *forked* child inherits the parent's live memory, so it sees the
flag the parent set after import; a *spawned* child re-imports this module fresh, so the
flag is back at its default.
"""
import asyncio
import multiprocessing

from nicegui import run

run.process_pool_start_method = 'spawn'  # opt into the safe start method (default in NiceGUI 4.0)

inherited_from_parent = False  # a spawned child re-imports and sees this default


def probe(_):
    return inherited_from_parent  # True only if this child inherited parent memory (forked)


async def main():
    global inherited_from_parent
    print(f"global start method: {multiprocessing.get_start_method()!r}")
    inherited_from_parent = True  # set AFTER import; only fork carries it into the child
    run.setup()
    forked = await run.cpu_bound(probe, None)
    assert not forked, 'cpu_bound ran under FORK - see nicegui#182'
    print('OK: cpu_bound ran under spawn')
    run.tear_down()


if __name__ == '__main__':
    asyncio.run(main())
```

</details>

### Progress

- [x] No breaking changes to the public API; the 4.0 default change is announced via a one-time warning and described above.


falkoschindler (changes requested):
Thanks for this — the diagnosis is excellent and the repro is genuinely convincing. The `fork`-with-threads deadlock is real, and pinning the pool to a scoped `mp_context` (rather than touching the global default, à la #3644) is the right shape of fix.

My one hesitation is the rollout: as written, this flips Linux `cpu_bound` to `spawn` **unconditionally**, which turns a set of currently-working Linux apps into apps that raise on the first call — unpicklable closures/args and missing `if __name__ == '__main__'` guards all worked under `fork` (inherited via the memory copy) and now won't. That's the same contract macOS/Windows have always had, so cross-platform apps already comply, but a Linux-only deployment relying on `fork` gets a hard break with no escape hatch, and the failure they were avoiding (the fork deadlock) is silent/intermittent while the new failure is immediate. For a behavior change that affects working apps, I'd rather land it on a major bump with a runway.

Proposal — phase it as a tri-state setting:

**Now (3.x):** introduce a start-method setting (e.g. a `run`-level knob) defaulting to `None`:

- `None` (unset) → inherit the **current global default** (byte-for-byte today's behavior), and `warn_once` _only_ when the resolved method is plain `fork` on Linux. Pointing the warning at the setting + the 4.0 change.
- `'spawn'` → opt in, silent (user has acknowledged the pickling contract — and can use this to fix a deadlock _today_).
- `'fork'` → opt out, silent (user knowingly owns the risk).

**4.0:** flip `None`'s meaning to `spawn`, keep the `'fork'` opt-out.

Why `None`-as-sentinel: it lets us distinguish "user never chose" (warn) from "user explicitly went `spawn`/`fork`" (silent). And making unset _inherit the global_ rather than _force fork_ keeps the path unchanged for people already on `forkserver` — which matters because `forkserver` is single-threaded in the server process (no deadlock) and becomes the default in Python 3.14, so unset users on 3.14+ self-heal and get no warning at all. Warn condition is roughly `setting is None and get_start_method() == 'fork'`.

The mechanism is basically the `SPAWN_CONTEXT` plumbing you already added — just resolved from the setting instead of unconditional, plus the warning. The main downside is the warning fires for Linux/fork users whose callables are already picklable (we can't cheaply detect fork-dependence at warn time), so it should be `warn_once` and worded as "heads-up + here's the opt-in," not "you're doing it wrong."

One validation nit if you take it: `get_context('fork')` raises on Windows, so an explicit `'fork'` there should fail with a clear message rather than a raw `ValueError`.

What do you think?

NiklasNeugebauer:
@

> What do you think?

Sounds good to me

evnchn:
> which turns a set of currently-working Linux apps into apps that raise on the first call

@falkoschindler For a better discussion and for something we can verify agent's handiwork, I think I need to make such an app and see it raise to convince myself. Fork vs Spawn is always a bit fuzzy for me.

evnchn:
*Drafted with Claude Code; the Linux `fork`/`spawn` results below were run and verified by me on a real Linux box.*

@falkoschindler — I built the app you asked about and ran it on a genuine Linux/`fork` box (both raw `ProcessPoolExecutor` and the real `run.cpu_bound`, on this branch). Seeing it raise (or *not* raise) changed my mind on the blast radius — and on two of the three break-modes.

**TL;DR:**

The fix is right, but the "currently-working Linux apps that now raise" set is **much smaller than the review states** — for idiomatic NiceGUI apps it collapses to essentially **one** hazard.

Measured on Linux (`fork` default), not reasoned:

1. **Unpicklable closures/args don't "work under `fork`" — they already raise today**, under both start methods. `ProcessPoolExecutor` pickles the submitted callable+args onto the call queue *regardless of start method*; `fork`'s memory-copy only carries module state captured at worker-fork time, never the task you `submit()` later. So the flip doesn't newly break these — they were never working.
2. **Missing `if __name__ == '__main__'` does *not* get forced.** A guard-less idiomatic app runs fine on this branch, because `ui.run()` self-guards (the `current_process().name != 'MainProcess'` early-return) and the pool is built in app-startup, not at import. No `RuntimeError`, no infinite spawn.
3. **The one genuine, silent regression:** code relying on a **module-level global mutated at runtime** and read inside a `cpu_bound` callable. Under `fork` the child sees the mutation; under `spawn` it re-imports and silently sees the *default* — wrong result, no exception.

So your core instinct (don't hard-break working Linux apps) is right, but aimed slightly off: the picklable/guard cases aren't real regressions, and the strongest argument for your warning is the silent state-inheritance case, not the ones the review leads with.

<details>
<summary><b>Setup + raw <code>fork</code> vs <code>spawn</code> matrix (identical cases, same box)</b></summary>

`.180`, Python 3.12.3, Linux, `multiprocessing.get_start_method() == 'fork'`. This branch installed (`nicegui …dev0+e234894d`). Each case run through `ProcessPoolExecutor(mp_context=get_context('fork'))` and `…get_context('spawn')` — which is exactly what this PR flips.

| case | `fork` | `spawn` | verdict |
|---|---|---|---|
| unpicklable **closure** as callable | 🔴 `AttributeError: Can't pickle local object` | 🔴 same | not `fork`-only — breaks under **both** |
| unpicklable **arg** (a `lambda`) | 🔴 raises | 🔴 raises | not `fork`-only — breaks under **both** |
| inherited module global (mutated after import) | 🟢 sees `'MUTATED-by-parent'` | 🟡 runs, returns `'default-from-fresh-import'` | **real `spawn` diff — silent wrong value, no raise** |

```python
import multiprocessing as mp
from concurrent.futures import ProcessPoolExecutor
FLAG = 'default-from-fresh-import'
def read_flag(_): return FLAG
def make_adder(n):
    def add(x): return x + n   # closure -> unpicklable
    return add
def run_case(label, method, fn, arg, set_flag=False):
    ctx = mp.get_context(method)
    if set_flag:
        global FLAG; FLAG = 'MUTATED-by-parent'
    try:
        with ProcessPoolExecutor(max_workers=1, mp_context=ctx) as ex:
            print(f"[{method:5}] {label}: OK -> {ex.submit(fn, arg).result(30)!r}")
    except Exception as e:
        print(f"[{method:5}] {label}: RAISED {type(e).__name__}")
    finally:
        if set_flag: FLAG = 'default-from-fresh-import'
def main():
    for m in ('fork', 'spawn'):
        run_case("closure-callable", m, make_adder(10), 5)
        run_case("global         ", m, read_flag, None, set_flag=True)
if __name__ == '__main__':
    main()
```
</details>

<details>
<summary><b>"Unpicklable already raises today" — on the real <code>run.cpu_bound</code> (released 3.9.0)</b></summary>

Released `nicegui 3.9.0`, Linux `fork` default (i.e. current behavior, *not* this branch):

```python
import asyncio
from nicegui import run
def make_adder(n):
    def add(x): return x + n
    return add
async def main():
    run.setup()
    try:
        print("OK:", await run.cpu_bound(make_adder(10), 5))
    except Exception as e:
        print("RAISED:", type(e).__name__, str(e).splitlines()[0])
    finally:
        run.tear_down()
asyncio.run(main())
```
```
cpu_bound unpicklable-closure: RAISED AttributeError: Can't pickle local object 'make_adder.<locals>.add'
```
An unpicklable `cpu_bound` callable is already a hard error on Linux today — so it isn't in the set of things this PR breaks.
</details>

<details>
<summary><b>Guard-less idiomatic app survives the flip — <code>ui.run()</code> self-guards</b></summary>

This branch, Linux `fork` default, **no `if __name__ == '__main__'`**, `ui.run()` at top level, `cpu_bound` fired from `on_startup`:

```python
import multiprocessing
from nicegui import ui, app, run
with open('/tmp/toplevel_runs.log', 'a') as f:
    f.write(f"toplevel ran: {multiprocessing.current_process().name}\n")
def heavy(x): return x * x
async def on_start():
    print("CPU_BOUND_RESULT=", await run.cpu_bound(heavy, 7)); app.shutdown()
app.on_startup(on_start)
ui.run(reload=False, port=8232, show=False)
```
```
CPU_BOUND_RESULT= 49          # exit 0, no RuntimeError, no guard needed
--- /tmp/toplevel_runs.log ---
toplevel ran: MainProcess
toplevel ran: SpawnProcess-1   # the cpu_bound worker DID re-import the module...
```
The worker re-imports the entry module (so top-level *side effects* re-run once per worker — new on Linux, but the long-standing macOS/Windows/native contract), yet `ui.run()` short-circuits in the worker via the `current_process().name != 'MainProcess'` early-return in `ui_run.py`. Net: no forced guard, no bootstrapping crash. My earlier "missing guard breaks" worry used a raw top-level pool, which is *not* how NiceGUI drives it — retracted.
</details>

<details>
<summary><b>Corrected blast-radius (idiomatic NiceGUI app, Linux)</b></summary>

| claimed break-mode | `fork` today | `spawn` (this PR) | real regression? |
|---|---|---|---|
| unpicklable closure/arg to `cpu_bound` | already raises | raises | **no** — already broken today |
| missing `__main__` guard | fine | still fine (`ui.run()` self-guards) | **no** — handled by design |
| top-level side effects | run once | re-run per worker | behavior change, but = existing macOS/Win/native contract |
| runtime-mutated module global read in `cpu_bound` | works | silently wrong (fresh import) | **yes — the one silent hazard** |
</details>

**Where this leaves your proposal:** I think the tri-state setting + `warn_once` is still worth doing, but the framing shifts — the warning's real job is the *silent* state-inheritance case (wrong result, no traceback), not the picklable/guard cases the review leads with (non-events for idiomatic apps). Given how narrow the genuine breakage is, I'm torn between (a) your full tri-state-with-runway and (b) just flipping it now with a release-note + a one-line warning pointed only at the silent case. Which do you prefer? Happy to implement either — and thanks for pushing on this; the state-inheritance hazard is the real one and I'd have under-weighted it.

The `get_context('fork')`-raises-on-Windows validation nit you raised is a good one and I'll fold it into whichever path we pick.
</content>

evnchn:
*Posted by Claude Code (Opus 4.8) on evnchn's behalf — reasoned from a real session; the empirical runs below were executed on a Linux box, not inferred.*

**Verdict:** implemented your tri-state proposal — pushed as 6f722c1e on top of the unconditional-spawn commit (add-on-top, so your review anchors survive; sorry it dismisses the stale approval).

**One thing I convinced myself of first, and it slightly sharpens your framing:** I built the "currently-working Linux app that breaks under spawn" and watched it, before touching the impl. The `unpicklable closures / args` half of the concern turns out **not** to be the fork-only break — `ProcessPoolExecutor` pickles the callable + args through its call queue under **both** fork and spawn, so those raise identically either way (no regression from flipping). The genuine fork-only breakage is **fork-inherited process state** — a global/cache/model the parent warmed at runtime that a spawned worker re-imports away. That either silently returns wrong results *or* raises. Net effect: your instinct to phase it is right, and if anything the case is stronger (a silent wrong-result is worse than an immediate error), which is exactly why the `None` path warns rather than flips.

**The ask:**

1. Does the resolution match your intent — `None` inherits + warns-on-fork, `'spawn'`/`'fork'` silent, latched at `setup()`?
2. The `4.0` flip is a `TODO` only (default stays `None` now) — good to leave that for the 4.0 branch?

<details>
<summary><b>Empirical: which run.cpu_bound patterns are actually fork-only? (nicegui 3.14.0, CPython 3.12.3, Linux, genuine fork default)</b></summary>

Ran each pattern under a forced `fork` and forced `spawn` global context:

| pattern | fork | spawn | fork-only? |
|---|---|---|---|
| top-level function | WORKED | WORKED | no |
| `lambda` callable | RAISED `AttributeError: Can't pickle local object` | RAISED (same) | **no — raises under both** |
| local closure | RAISED (same pickle error) | RAISED (same) | **no — raises under both** |
| unpicklable arg (`threading.Lock`) | RAISED `TypeError: cannot pickle '_thread.lock'` | RAISED (same) | **no — raises under both** |
| reads a runtime-mutated module global | WORKED (`'MUTATED'`) | WORKED but **wrong** (`'DEFAULT'`) | **yes — silent divergence** |
| no `if __name__ == '__main__'` guard | WORKED | WORKED (worker re-imports as `__mp_main__`) | not a clean break in a standalone harness |

Takeaway: the pickling contract (`'spawn' requires picklable callables/args`) was **already** enforced under fork for `run.cpu_bound`, because the pool transports work items via pickle regardless of start method. So flipping fork→spawn does not regress the unpicklable cases — it regresses code that leaned on **fork's memory inheritance**.
</details>

<details>
<summary><b>The clean MRE: "load a model once into a global" — WORKS under fork, RAISES under spawn</b></summary>

This is the realistic shape of your concern (load-once-use-in-every-worker). It drives `run.setup()` + `run.cpu_bound()` directly — the same reduction your PR's own MRE uses — to isolate fork-inherited state without dragging in the server / `__mp_main__` axis. `ui.run()` calls `run.setup()` for you; `run.cpu_bound` is public.

```python
import asyncio, multiprocessing, sys
from nicegui import run

MODEL_CACHE: dict[str, float] = {}                 # expensive resource, empty at import

def warm_up_cache() -> None:
    MODEL_CACHE.update({'gain': 2.5, 'offset': 0.1})  # populated at RUNTIME (startup), not import

def score(sample: float) -> float:
    return sample * MODEL_CACHE['gain'] + MODEL_CACHE['offset']  # KeyError if the cache is empty

async def main() -> None:
    warm_up_cache()                                # only fork carries this into the worker
    run.setup()
    try:
        result = await run.cpu_bound(score, 10.0)
        assert result == 25.1
        print(f'cpu_bound = {result!r}  ->  WORKED')
    except Exception as e:
        print(f'{type(e).__name__}  ->  RAISED')
    finally:
        run.tear_down()

if __name__ == '__main__':
    multiprocessing.set_start_method(sys.argv[1], force=True)  # fork | spawn
    asyncio.run(main())
```

Output on the same Linux box (the one-line `<fork|spawn>` argv is the works/broken toggle):

```
$ python repro.py fork
cpu_bound = 25.1  ->  WORKED

$ python repro.py spawn
nicegui.run.SubprocessException: Exception in subprocess:
  Type: KeyError
  ...
    return sample * MODEL_CACHE['gain'] + MODEL_CACHE['offset']  # KeyError if the cache is empty
KeyError: 'gain'
  ->  RAISED
```

Proves: fork inherits the warmed global, spawn re-imports it away → raises on first call. It does **not** prove anything about the `__main__`-guard path (that was a separate, murkier result — see the matrix above).
</details>

<details>
<summary><b>Implementation: run.process_pool_start_method, resolved-once-at-setup() (diff 6f722c1e)</b></summary>

`nicegui/run.py`:

```python
process_pool_start_method: str | None = None
# None  -> inherit the global default (byte-for-byte today); warn_once on first cpu_bound use
#          only when the pool resolves to plain fork.
# 'spawn' -> opt in, silent.   'fork' -> opt out, silent.
# TODO(4.0): flip None -> 'spawn' (keep the 'fork' opt-out).

SPAWN_CONTEXT = multiprocessing.get_context('spawn')  # always-spawn; still shared with native.py

def _resolve_cpu_bound_context() -> BaseContext:
    if process_pool_start_method is None:
        return multiprocessing.get_context()          # global default
    if process_pool_start_method == 'spawn':
        return SPAWN_CONTEXT
    if process_pool_start_method == 'fork':
        try:
            return multiprocessing.get_context('fork')
        except ValueError as e:                        # Windows: no 'fork'
            raise ValueError("run.process_pool_start_method='fork' is not available on this "
                             "platform (e.g. Windows); use 'spawn' or None instead.") from e
    raise ValueError(f"Invalid run.process_pool_start_method {process_pool_start_method!r}; ...")
```

`setup()` resolves the context **once** and latches whether it's an implicit fork; `cpu_bound`'s warning and the `BrokenProcessPool` restart both consult that latched value — so the live pool, the heads-up, and any restart can't disagree even if the knob is mutated after startup (this addressed a code-review finding — see below).

`native.py` is unchanged in behavior: its window subprocess still binds the **always-spawn** `run.SPAWN_CONTEXT` and is untouched by the knob (verified `run.SPAWN_CONTEXT is native.SPAWN_CONTEXT` in both import orders).
</details>

<details>
<summary><b>Empirical: the three settings on Linux (fork default) behave exactly as specced</b></summary>

Same `MODEL_CACHE` app, driven through the knob (worktree source shadowed over an installed nicegui):

| `run.process_pool_start_method` | resolved pool method | fork-state app | heads-up warning |
|---|---|---|---|
| `None` (default) | `fork` | ✅ WORKED (25.1) — today's behavior byte-for-byte | **fires once** |
| `'spawn'` | `spawn` | ❌ RAISED (`KeyError` in worker) — the pickling/state contract, opt-in | silent |
| `'fork'` | `fork` | ✅ WORKED (25.1) | silent |

The `None`-on-fork heads-up (worded as opt-in, not "you're doing it wrong"):

> run.cpu_bound is using the 'fork' start method (the platform default on Linux/Docker). fork is unsafe in a threaded process and its workers inherit process state (module globals, caches, preloaded objects) that spawn workers do not, so switching can change behavior. NiceGUI 4.0 will default to 'spawn'. Set nicegui.run.process_pool_start_method to 'spawn' to opt in now (recommended) or to 'fork' to keep fork and silence this heads-up. See https://github.com/zauberzeug/nicegui/pull/6117

Placed on first `cpu_bound` *use*, not at `setup()`, so apps that never call `cpu_bound` are never nagged.
</details>

<details>
<summary><b>Windows fork-validation — confirmed (unit test, not a real Windows run)</b></summary>

Your nit: `get_context('fork')` raises on Windows, so an explicit `'fork'` there must fail clearly. Handled in `_resolve_cpu_bound_context()` (the `except ValueError` above → clear message naming the setting). I can't run real Windows here, so this is covered by a unit test that monkeypatches `multiprocessing.get_context` to raise the Windows `ValueError` and asserts the clear message — honest scope: it proves the wrapping logic, not a live Win32 run.
</details>

<details>
<summary><b>Code review before push (2 independent passes) + the desync fix it caught</b></summary>

Ran a Claude reviewer and a second, non-Claude lineage (Codex) over the diff. Neither found a correctness bug in the core (native-window-stays-spawn ✅, warn condition matches your spec ✅, `None` byte-for-byte ✅). Both independently flagged one real smell: the first draft re-read the setting at warn/restart time, so mutating the knob *after* `setup()` could make the warning (or a restart) disagree with the live pool. Fixed by latching the resolved context at `setup()`; verified on Linux that flipping the knob to `'spawn'` after an implicit-fork `setup()` still correctly reports the live pool as fork.
</details>

<details>
<summary><b>Gates (green) + SHAs</b></summary>

- Base: e234894dd30dc3c633c104d5325e93bb561dd0ae · Head: 6f722c1e9686b8e7fd461e2bb80d6f61921e56d3
- `pre-commit run` (ruff / autopep8 / codespell / mdformat) ✅ · `mypy ./nicegui` ✅ (241 files) · `pylint ./nicegui` 10.00/10 ✅
- `pytest tests/test_run.py` → 16 passed, 1 skipped ✅ (the skip is the fork-default desync test; executed + passing separately on the Linux box)
- New tests cover: fork/spawn/None resolution, `'spawn'` reuses `SPAWN_CONTEXT`, invalid value → clear `ValueError`, Windows-fork → clear error, `setup()` implicit-fork latching, warn gate, and the post-setup knob-change desync immunity.
</details>

falkoschindler (approved):
Thanks @evnchn! The tri-state implementation matches my intent, and to answer your two questions: yes to both — the resolution is what I had in mind, and the 4.0 flip stays out of this PR (it's now a `DEPRECATED` marker, our convention instead of `TODO`).

On top of your branch we pushed a few follow-up commits:

- **Docs:** the setting is mentioned in the `cpu_bound` docstring and the website's "Running CPU-bound tasks" note (including your inherited-state finding), and the warning says it must be set before `ui.run()`.
- **Typing:** `Literal['spawn', 'fork', 'forkserver'] | None`, so typos fail statically.
- **Pass-through instead of strict tri-state:** explicit values go straight to `multiprocessing.get_context()`, which loosens my own proposal — `'forkserver'` (the Linux default from Python 3.14 on) is supported, and the Windows-`'fork'` validation nit resolves itself because unknown and unavailable values now share one error listing what the platform offers.
- **Scoped promises:** the `'fork'` opt-out is promised only "as long as Python supports it", since CPython has deprecated forking a multi-threaded process (3.12+).
- **Tests:** reworked toward observable behavior — asserting the built pool's context and driving a real `cpu_bound` call while reading the log, instead of exercising private helpers. The fork-dependent cases run on Linux CI and were additionally verified on macOS with a force-forked default.

Verified end-to-end under both a spawn and a forced fork default: the implicit-fork pool warns once through a real `cpu_bound` call, explicit `'spawn'`/`'fork'` stay silent, and flipping the setting after startup doesn't silence the warning about the still-fork live pool. Approving — thanks for the persistence and the rigor.

PR #6123, opened 2026-06-19
Fix interactive_image coordinates when stream changes resolution

### LINKED ISSUE #6122: [Interactive Image] Click positions are wrong when changing the image's size
### First Check

### Example Code

```python
import time
from collections.abc import Generator
from typing import Any

import cv2
import numpy as np
import requests
from fastapi.responses import StreamingResponse
from nicegui import events, ui

# fetch images
response = requests.get("https://picsum.photos/200/300.jpg")
img_0 = cv2.imdecode(np.frombuffer(response.content, np.uint8), cv2.IMREAD_COLOR)
response = requests.get("https://picsum.photos/400/600.jpg")
img_1 = cv2.imdecode(np.frombuffer(response.content, np.uint8), cv2.IMREAD_COLOR)


def frame_generator() -> Generator[bytes, Any, None]:
    # Encode the frame and yield it to the generator
    while True:
        _, d = cv2.imencode(".jpg", img_0, [cv2.IMWRITE_JPEG_QUALITY, 90])
        yield b"--frame\r\nContent-Type: image/jpeg\r\n\r\n" + bytearray(d) + b"\r\n"

        time.sleep(0.1)


@ui.page("/get_image")
def video_stream_reader_image() -> StreamingResponse:
    return StreamingResponse(frame_generator(), media_type="multipart/x-mixed-replace; boundary=frame")


def change_image():
    global img_0
    img_0 = img_1


def root():

    def mouse_handler(e: events.MouseEventArguments):
        color = "SkyBlue" if e.type == "mousedown" else "SteelBlue"
        ii.content += f'<circle cx="{e.image_x}" cy="{e.image_y}" r="15" fill="none" stroke="{color}" stroke-width="4" />'
        ui.notify(f"{e.type} at ({e.image_x:.1f}, {e.image_y:.1f})")

    ui.button("Change image", on_click=change_image)
    ii = ui.interactive_image("/get_image", on_mouse=mouse_handler, events=["mousedown", "mouseup"], cross=True, sanitize=False)


ui.run(root)
```

### Description

I got an a project where I have to show a decoded video stream via interactive image (due to the possibilities of having SVG overlays, which cannot be done with the normal video player -> should support resizing the browser window).
When the stream is connecting and starting to decode, I show a placeholder image with a fixed resolution (like, "waiting for images"). Once the stream is up & running, I switch the images to the ones coming from the stream.

The stream can have a different resolution. Once changed, the click event does not result in correct positions. It is still using the placeholder image's dimensions. To fix this, the user have manually to refresh the page.

The attached example (despite not being the nicest python code ever written) shows the issue.

### NiceGUI Version

3.13.0

### Python Version

3.12.4

### Browser

Edge

### Operating System

Windows

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #6122. When an `interactive_image` shows a `multipart/x-mixed-replace` stream (e.g. an MJPEG decoder) and the streamed resolution changes — the reported pattern: a fixed-size placeholder later replaced by the real, differently-sized stream — mouse/pointer coordinates and the SVG overlay keep using the *old* dimensions until a full page reload.

Root cause: the component caches image dimensions only from the `` `load` event. A multipart stream swaps frames in place — `naturalWidth`/`naturalHeight` change but `load` never re-fires — so `loaded_image_width`/`loaded_image_height`, the SVG `viewBox`, and the `loaded` event stay frozen at the first-seen size.

### Implementation

Add `refreshImageDimensions()`, which reads the live `naturalWidth`/`naturalHeight` at event time and, **only when they differ** from the cache, resyncs the cache + `viewBox` and emits `loaded`. It is called at the top of `onMouseEvent`, `onPointerEvent`, and `updateCrossHair`.

- Guarded by change-detection → it is a no-op (and emits nothing) on the normal `load` path, so no duplicate `loaded` events.
- Only the `src`-present branch is touched; the `size`-only (sourceless canvas) mode is unchanged.
- Emitting `loaded` on a real size change keeps `add_layer()` layers and server-side `on('loaded')` handlers consistent too.

The resync is **event-time only**: there is no event for an in-place `naturalWidth` change (and `ResizeObserver` watches the rendered box, not natural size), so `viewBox`/`aspectRatio`/`loaded` catch up on the next pointer/mouse interaction. Coordinates — the reported bug — are always correct because they are computed at event time.

`test_resync_on_stream_resolution_change` serves a padded multipart MJPEG, swaps its resolution in place, and asserts the click coordinates follow the new size (fails on `main`, passes with the fix).

### Progress

### DIFF
--- a/nicegui/elements/interactive_image.js
+++ b/nicegui/elements/interactive_image.js
@@ -106,8 +106,8 @@ export default {
       }
     },
     refreshImageDimensions() {
-      // Sync cached dimensions with the live <img> in case the source changed size without firing "load"
-      // (e.g. a multipart/x-mixed-replace stream switching resolution); see #6122.
+      // Re-sync cached dimensions with the live <img> when the stream changed size without firing "load" (#6122).
+      // No event fires for an in-place naturalWidth change, so this only catches up on the next interaction.
       const img = this.$refs.img;
       if (
         this.src &&
@@ -118,7 +118,7 @@ export default {
         this.loaded_image_width = img.naturalWidth;
         this.loaded_image_height = img.naturalHeight;
         this.updateViewbox(this.loaded_image_width, this.loaded_image_height);
-        this.$emit("loaded", { width: this.loaded_image_width, height: this.loaded_image_height, source: img.currentSrc });
+        this.$emit("loaded", { width: this.loaded_image_width, height: this.loaded_image_height, source: img.src });
       }
     },
     updateCrossHair(e) {
--- a/nicegui/elements/interactive_image.js
+++ b/nicegui/elements/interactive_image.js
@@ -105,7 +105,24 @@ export default {
         this.updateViewbox(this.size[0], this.size[1]);
       }
     },
+    refreshImageDimensions() {
+      // Sync cached dimensions with the live <img> in case the source changed size without firing "load"
+      // (e.g. a multipart/x-mixed-replace stream switching resolution); see #6122.
+      const img = this.$refs.img;
+      if (
+        this.src &&
+        img.naturalWidth &&
+        img.naturalHeight &&
+        (img.naturalWidth !== this.loaded_image_width || img.naturalHeight !== this.loaded_image_height)
+      ) {
+        this.loaded_image_width = img.naturalWidth;
+        this.loaded_image_height = img.naturalHeight;
+        this.updateViewbox(this.loaded_image_width, this.loaded_image_height);
+        this.$emit("loaded", { width: this.loaded_image_width, height: this.loaded_image_height, source: img.currentSrc });
+      }
+    },
     updateCrossHair(e) {
+      this.refreshImageDimensions();
       const width = this.src ? this.loaded_image_width : this.size ? this.size[0] : 1;
       const height = this.src ? this.loaded_image_height : this.size ? this.size[1] : 1;
       this.x = (e.offsetX * width) / e.target.clientWidth;
@@ -118,6 +135,7 @@ export default {
       this.$emit("loaded", { width: this.loaded_image_width, height: this.loaded_image_height, source: e.target.src });
     },
     onMouseEvent(type, e) {
+      this.refreshImageDimensions();
       const imageWidth = this.src ? this.loaded_image_width : this.size ? this.size[0] : 1;
       const imageHeight = this.src ? this.loaded_image_height : this.size ? this.size[1] : 1;
       this.$emit("mouse", {
@@ -133,6 +151,7 @@ export default {
       });
     },
     onPointerEvent(type, e) {
+      this.refreshImageDimensions();
       const imageWidth = this.src ? this.loaded_image_width : this.size ? this.size[0] : 1;
       const imageHeight = this.src ? this.loaded_image_height : this.size ? this.size[1] : 1;
       this.$emit(`svg:${type}`, {
--- a/tests/test_interactive_image.py
+++ b/tests/test_interactive_image.py
@@ -133,7 +133,8 @@ def test_resync_on_stream_resolution_change(screen: Screen):
         buffer = io.BytesIO()
         Image.new('RGB', (width, height)).save(buffer, format='JPEG')
         data = buffer.getvalue()
-        # pad to force instant paint
+        # Browsers won't paint a multipart frame until enough bytes have arrived, so a tiny JPEG can stall.
+        # Bulk it up by inserting a JPEG comment segment (0xFFFE marker + 2-byte length) padded with spaces after the SOI marker.
         return data[:2] + b'\xff\xfe' + struct.pack('>H', 40_002) + b' ' * 40_000 + data[2:]
 
     frame = {'data': jpeg(300, 200)}
--- a/tests/test_interactive_image.py
+++ b/tests/test_interactive_image.py
@@ -121,6 +121,56 @@ def test_add_layer(screen: Screen):
         assert len(screen.find_all_by_tag('circle')) == 1
 
 
+def test_resync_on_stream_resolution_change(screen: Screen):
+    """https://github.com/zauberzeug/nicegui/issues/6122"""
+    import asyncio
+    import io
+    import struct
+    from collections.abc import AsyncGenerator
+
+    from fastapi.responses import StreamingResponse
+    from PIL import Image
+
+    def jpeg(width: int, height: int) -> bytes:
+        buffer = io.BytesIO()
+        Image.new('RGB', (width, height)).save(buffer, format='JPEG')
+        data = buffer.getvalue()
+        # pad to force instant paint
+        return data[:2] + b'\xff\xfe' + struct.pack('>H', 40_002) + b' ' * 40_000 + data[2:]
+
+    frame = {'data': jpeg(300, 200)}
+    clicks = []
+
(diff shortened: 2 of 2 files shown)


evnchn:
*Posted by Claude Code on behalf of @evnchn.*

@LeBoozer — opened as a draft so you can confirm it fixes things on your **actual** stream before we finalize. Two ways to try it:

**A — install this branch** (needs `git`):

```
pip install "git+https://github.com/evnchn/nicegui.git@fix/6122-interactive-image-resize"
```

**B — swap one file** (it's a single frontend-file change): replace `nicegui/elements/interactive_image.js` in your installed package with [the branch version](https://github.com/evnchn/nicegui/blob/fix/6122-interactive-image-resize/nicegui/elements/interactive_image.js), then hard-refresh the browser (Ctrl+F5).

Then: show the placeholder, switch to the real (differently-sized) stream, and click — the coordinates should be correct **without** reloading the page. Let me know if anything's still off.

falkoschindler (commented):
Thanks for the quick PR, @evnchn!

I guess I found a reliable reproduction (see https://github.com/zauberzeug/nicegui/issues/6122#issuecomment-4759725572) which allows us to write a Screen test.

But I'm a bit worried about the "loaded" event: it is lazy (only emitted on user interaction) and contains inconsistent source attributes. I hope we can find a fix for that.

Here's the full review:

1. **The test note understates what CI can do** (major) — The PR says headless Chromium "won't render `multipart/x-mixed-replace` into an `` at all — no `load`/`error`, `naturalWidth` stays `0`," so no faithful regression test is possible. I checked this directly in headless Chromium and it's not quite right: headless **does** render an MJPEG `` and **does** update `naturalWidth` on an in-place resolution swap (I observed `naturalWidth` go 200→400 across a swap while `viewBox` stayed `0 0 200 300` on `main`). The catch is purely a buffering threshold — Chrome won't paint the first frame until it has buffered enough bytes, so _tiny_ placeholder frames (e.g. a 2.7 KB `dummyimage` 200×300) sit unrendered for 8–15 s and read `naturalWidth === 0`, which is likely what was observed. Padding each frame past the threshold fixes it: with ~40 KB frames (e.g. a JPEG `\xff\xfe` comment segment as filler) first paint was **0 ms** in headless. So a `Screen` regression test looks feasible: serve a padded MJPEG endpoint, assert `viewBox`/coordinates, flip the streamed frame to a different size, dispatch a mouse event, and assert the coordinates/`viewBox` now reflect the new size. Worth attempting — though stream timing may make it flaky, so if a deterministic version proves elusive, a follow-up issue documenting the manual repro is a fair fallback. The MRE posted on #6122 already uses exactly this padding (the `inflate()` helper, with the note that "without inflating the individual frames, Chrome wouldn't load the image reliably"), so it's a ready starting point for a `Screen` test.

2. **Dimensions only resync on user interaction** (minor) — Because the refresh is event-driven, a resolution change that the user doesn't interact with leaves `viewBox`, the container `aspectRatio` (derived from `loaded_image_width/height`), and the `loaded` event stale until the first pointer/mouse event. For the reporter's placeholder→stream case the overlay and layout can stay visibly wrong (and `on('loaded')` is delayed) until the user moves the mouse. Coordinates — the actual reported bug — are correct since they're computed at event time, so this is a real but lesser gap. Proactive detection is genuinely awkward for `` (no event fires for an in-place `naturalWidth` change, and `ResizeObserver` watches the rendered box, not natural size), so the event-time approach is a defensible trade-off — just worth a one-line comment noting the limitation, or a brief mention in the PR description.

3. **`loaded` payload `source` differs between the two emit sites** (minor) — `onImageLoaded` emits `source: e.target.src` (around line 135) while `refreshImageDimensions` emits `source: img.currentSrc` (around line 121). `src` is the (possibly path-prefixed) attribute and `currentSrc` is the resolved URL; for a stream they point at the same resource but the string a `loaded` handler receives will differ depending on whether the event came from a load or a refresh. Minor, but aligning the two avoids a surprise for anyone keying on `source`.

evnchn:
*Posted by Claude Code on behalf of @evnchn.*

Re-verified all three — your buffering diagnosis was the key, and my "CI can't do this" claim was wrong.

**1 — Screen test (you were right).** Confirmed in headless: padded frames paint immediately and `naturalWidth` updates in place across a swap; my earlier failure was exactly the sub-threshold tiny-frame case you identified. Added `test_resync_on_stream_resolution_change` — serves a padded MJPEG, polls `naturalWidth` to stay deterministic, and clicks before/after a 300×200→600×400 swap asserting the coordinates follow. Fails on `main` (stuck at `(150, 100)`), passes with the fix; 5/5 stable locally at ~2.8 s. PR body corrected.

**2 — event-time only.** Added a one-line comment: no event fires for an in-place `naturalWidth` change (and `ResizeObserver` watches the rendered box, not natural size), so `viewBox`/`aspectRatio`/`loaded` only catch up on the next pointer/mouse event. Coordinates — the reported bug — are always correct since they're computed at event time.

**3 — source.** `refreshImageDimensions` now emits `source: img.src`, matching `onImageLoaded`'s `e.target.src`.

On the `loaded`-event worry overall: the inconsistent-source half is fixed; the laziness is inherent to the event-time approach (same as `viewBox`). If you'd rather not have `loaded` fire off an interaction at all, I'm happy to drop that emit and keep this coordinates-only — your call.

LeBoozer:
> _Posted by Claude Code on behalf of @evnchn._
> 
> @LeBoozer — opened as a draft so you can confirm it fixes things on your **actual** stream before we finalize. Two ways to try it:
> 
> **A — install this branch** (needs `git`):
> 
> ```
> pip install "git+https://github.com/evnchn/nicegui.git@fix/6122-interactive-image-resize"
> ```
> 
> **B — swap one file** (it's a single frontend-file change): replace `nicegui/elements/interactive_image.js` in your installed package with [the branch version](https://github.com/evnchn/nicegui/blob/fix/6122-interactive-image-resize/nicegui/elements/interactive_image.js), then hard-refresh the browser (Ctrl+F5).
> 
> Then: show the placeholder, switch to the real (differently-sized) stream, and click — the coordinates should be correct **without** reloading the page. Let me know if anything's still off.

Hi @evnchn,

your fix is working for me!

falkoschindler (approved):
LGTM — well-scoped fix with a solid regression test. The root cause is correctly identified (an in-place `multipart/x-mixed-replace` frame swap changes `naturalWidth`/`naturalHeight` without firing `load`), and resyncing at event time is the right trade-off given there's no browser event for an in-place natural-size change.

I pushed two small tidy-ups to the branch:

- **Move test imports to module level** — `asyncio`, `io`, `struct`, `StreamingResponse`, and `PIL.Image` now sit at the top of the file, matching the convention in `test_image.py`.
- **Explain JPEG padding trick in test** — expanded the comment so the `0xFFFE` padding is self-explanatory: browsers won't paint a multipart frame until enough bytes have arrived, so we bulk each frame up with a padded JPEG comment segment after the SOI marker.

On the `loaded`-emit question: keeping the emit is the right call — it's the only path that resyncs `add_layer()` overlays after an in-place resolution change, and since `loaded` already fires on every reload, the change-guarded re-emit doesn't break any single-fire expectation. The remaining laziness (`viewBox`/`aspectRatio`/overlays only catch up on the next interaction) is inherent to the event-time approach and fine to leave as a documented limitation.

Approving. 🎉

PR #188, opened 2022-12-03
reverse proxy in NiceGUI 1.0

rodja:

Issue #186, opened 2022-12-01
Page Layout: how to add additional ui.footer content


JS-Aibel:
How do I add additional labels in the footer of a layout page?

1. In the footer below, I have added a global variable `rowData = ui.row()`.
2. Running the code `with rowData: ui.label("This is working")` **inside** the page_layout() method, is woring as expected.
3. Running the same code (2) outside the page_layout() produces the following error: `NameError: name 'rowData' is not defined variable`

```from nicegui import ui

@ui.page('/')
async def page_layout():
    ui.label('CONTENT')
    with ui.header():
        ui.label('HEADER')
    with ui.left_drawer(top_corner=True, bottom_corner=True):
        ui.label('LEFT DRAWER')
    with ui.right_drawer(fixed=False):
        ui.label('RIGHT DRAWER')
    with ui.footer():
        ui.label('FOOTER')
        global rowData
        rowData = ui.row()

    with rowData:
        ui.label("This is working")


with rowData:
    ui.label("This is working")


ui.run()
```

5. If I place the call on the rowData in a method where I yield, I do not get the error message, but the new label is not displayed in the footer.

```
def testing():
    with rowData:
        yield
        ui.label("This is not working")

testing()

ui.run()
```

falkoschindler:
*When* should `ui.label("This is not working")` be added? `rowData` is only defined and initialized when a client accesses the page and `page_layout()` is evaluated for the first time. Before that `rowData` is undefined.

Triggering the new label from, e.g., a button on another site is working:

```python
@ui.page('/other_page')
def other_page():
    def add():
        with rowData:
            ui.label('This is not working')
    ui.button('add another label', on_click=add)
```

PR #6124, opened 2026-06-21
Harden against X-Forwarded-Prefix reflection and clarify security docs

evnchn:
### Motivation

We keep receiving GHSA reports about the unsanitized page template (`| safe` sinks, `innerHTML`, `X-Forwarded-Prefix` reflection). They are almost all duplicates of patterns that are either by-design or already bounded. This PR hardens the one place that is genuinely client-controllable and documents the rest, so future reports of the same shape become non-events.

It consolidates three of the "brick" PRs reviewed on the fork (#185 + #186 + #187); #184 and #188 were dropped as redundant.

### Implementation

**1. Percent-encode `X-Forwarded-Prefix` at the source** (`client.py`) — the only behavior change.

`prefix` is reflected into the page (importmap, `<script src>`, JS `prefix:`, CSS `@import`) via `| safe`, and the `X-Forwarded-Prefix` part is client-controllable on a directly-exposed app or a pass-through proxy. We `quote()` it with a safe set equal to the RFC 3986 path-legal characters (`/:@!$&'()*+,;=` plus `%` to avoid double-encoding). So encoding is a **no-op for any well-formed URL-path prefix** but neutralizes every structural break-out character (`"` `<` `>` `\`, space, non-ASCII). Encoding at the source — not the template — also covers `generate_resources`, not just the importmap.

Regression tests cover an injection payload, a parametrized set of valid prefixes (asserted byte-for-byte unchanged), and non-ASCII header behavior (always percent-encoded, never reflected raw).

**2 & 3. Documentation** (no behavior change):

- **Shared-cache invariant** (`section_security.py`): each page embeds a fresh `client_id`, so a CDN / reverse-proxy cache that stores the HTML both leaks one visitor's session token to everyone and turns reflected input into a cache-poisoning vector. NiceGUI sends `Cache-Control: no-store` for exactly this reason — don't strip it.
- **Commonly Misreported Patterns** (`SECURITY.md`): a short section listing the by-design patterns (`| safe`, `sanitize=False` `innerHTML`, `add_head_html` / `run_javascript` / `add_css`, `ast.literal_eval` in docs) so reporters trace the data flow before filing — while explicitly still inviting reports that find a way *around* these protections.

Out of scope: the open-redirect surface in `middlewares.py` (GHSA-6fp9) — percent-encoding doesn't address it and the team judged it non-exploitable; defense-in-depth path-validation there can be a separate change if wanted.

### Progress

- [x] Pytests have been added (`tests/test_forwarded_prefix.py`, 8 cases).
- [x] No breaking changes — encoding is a no-op for every well-formed prefix.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,14 +14,27 @@ This way you will benefit from the latest features, bug fixes, and **security fi
 ## Commonly Misreported Patterns
 
 Automated scanners and AI-assisted reviews frequently flag the patterns below.
-They are **by design or already bounded**, so please trace the data flow and check them against this list before filing — it keeps the advisory queue focused on real issues.
+They are **by design or already bounded**, so please trace the data flow and check them against this list before filing —
+it keeps the advisory queue focused on real issues.
 
-- **`| safe` in the page template** (`nicegui/templates/index.html`). Jinja auto-escaping is bypassed there intentionally; the values are escaped at the source instead (for example, element data is escaped with a table sized for the `String.raw` script context it lands in). A bare `| safe` is not itself a finding — trace where the value comes from and how it is escaped.
-- **Raw `innerHTML` in `ui.html` / `ui.markdown`** (`nicegui/elements/html.js`, `markdown.js`). The raw assignment is the `else` branch of `sanitize`: the opt-in `sanitize=False` path. The default (`sanitize=True`) sanitizes via DOMPurify. Passing untrusted input with `sanitize=False` is the caller's documented responsibility.
-- **`ui.add_head_html`, `ui.add_body_html`, `ui.run_javascript`, `ui.add_css`.** These inject raw HTML/JS/CSS by design and are documented as never-for-untrusted-input. Using them with unsanitized user data is an application bug, not a framework vulnerability.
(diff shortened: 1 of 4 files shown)


falkoschindler (approved):
Approving. I layered four small follow-up commits on top — no change to the security behavior, just polish:

- **Clarify the safe-char comment** — scope the "no-op" claim to well-formed ASCII prefixes, document that non-ASCII is always encoded but can't round-trip cleanly (latin-1 headers), and note why `%` stays in the safe set (a literal `%xx` passes through but is inert).
- **Inline the `FORWARDED_PREFIX_SAFE_CHARS` constant** — it was used once; pass the char set directly to `quote()` and fold its rationale into the single defense-in-depth comment at the call site.
- **Reflow the security docs** — wrap the new `SECURITY.md` section and the shared-cache bullet to one clause per line (no content change).
- **Simplify the tests** — drop the redundant raw-payload assertion (implied by the executable-tag check), merge the valid-path and non-ASCII cases into one parametrized test asserting each input reflects in its expected safe form (keeping a "never raw in any sink" guard), and use a bytes header key so the values type-check against httpx `HeaderTypes`.

LGTM.

PR #6125, opened 2026-06-22
Fix GLTF models being overwritten with the default material at startup (#6118)

### LINKED ISSUE #6118: GLTF texture being overwritten at initialization with default material
### First Check

### Example Code

```python
from nicegui import ui, app

with ui.scene() as scene:
    scene.gltf('https://raw.githubusercontent.com/fabian0702/test2/refs/heads/main/test.glb')

ui.run()
```

### Description

When loading any gltf model the current implementation overwrites the existing style / texture of the model in the initialization of the scene with a default material (all white full opacity) (see code [here](https://github.com/zauberzeug/nicegui/blob/049871fb510822a65f3af7f6ee7796deedbb08ab/nicegui/elements/scene/scene.js#L591)). This bug was present since the initial implementation of gltf loading but did not yet cause issues as there was no mechanism to actually apply the material gltf. This was changed in PR #5708 which now leads to this bug. I'm not entirely sure why the avocado gltf used in testing remains unaffected, but for any gltfs i created with blender with a custom material it got overwritten (see example). A simple fix might be to check if there is a material applied (at initialization) and otherwise skip applying the default material.

### NiceGUI Version

3.13.0

### Python Version

3.14.4

### Browser

Firefox

### Operating System

Linux

### Additional Context

Current state:    

Image    

Intended:

Image

evnchn:
### Motivation

Fixes #6118.

GLTF models ship their own materials, but `init_objects` applied the default material (`'#ffffff'` / opacity `1.0` / side `'front'`) to **every** object at startup — overwriting GLTF models with plain white. The bug was latent until #5708 made `material()` settings actually reach GLTF child meshes, at which point the default-material application started clobbering the model's own materials.

Reported repro (`scene.gltf(url)` with a Blender-authored material) renders all-white instead of the model's own colours. The avocado/`box.glb` used in existing tests masked it because `test_gltf` overrides the material to `#ff0000` explicitly.

### Implementation

- `Object3D` now tracks whether `material()` was **explicitly** called (`material_is_set`), serialised in `data`.
- `init_objects` (JS) skips applying the default material to GLTF models when `material_is_set` is `False`, so the model keeps its own materials. An explicit `.material(...)` call still overrides as before (it sets the flag).
- `material()` sets the flag on the first call even when the values equal the defaults, so `gltf().material('#ffffff')` (explicitly forcing white) is honoured rather than skipped.
- Non-GLTF primitives are unaffected — they still receive the default material at init, which they rely on (created as a blank white `MeshPhongMaterial`).

A new test `test_gltf_keeps_own_material` loads `box.glb` (which carries its own red material) without calling `.material()` and asserts the model's own colour survives. It was confirmed to fail before the fix (material reads `[1, 1, 1]` white) and pass after.

### Progress

---

@fabian0702 thanks for the precise report and root-cause pointer — it made this quick to track down. If you'd like to confirm it fixes your models before this is merged, you can install this branch directly:

```bash
pip install "git+https://github.com/evnchn/nicegui.git@fix/6118-gltf-material"
```

Then run your original snippet — the model should now keep its own material instead of turning white. (Hard-refresh the browser to clear the cached \`scene.js\`.) Let me know if any of your GLTFs still come out wrong.

### DIFF
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -576,6 +576,7 @@ export default {
         color,
         opacity,
         side,
+        material_is_set,
         x,
         y,
         z,
@@ -588,7 +589,8 @@ export default {
       ] of data) {
         this.create(type, id, parent_id, ...args);
         this.name(id, name);
-        this.material(id, color, opacity, side);
+        // For GLTF models, only override the model's own materials when the user explicitly set one (#6118).
+        if (type != "gltf" || material_is_set) this.material(id, color, opacity, side);
         this.move(id, x, y, z);
         this.rotate(id, R);
         this.scale(id, sx, sy, sz);
--- a/nicegui/elements/scene/scene_object3d.py
+++ b/nicegui/elements/scene/scene_object3d.py
@@ -25,6 +25,7 @@ class Object3D:
         self.color: str | None = '#ffffff'
         self.opacity: float = 1.0
         self.side_: str = 'front'
+        self.material_is_set: bool = False
         self.visible_: bool = True
         self.draggable_: bool = False
         self.x: float = 0
@@ -48,7 +49,7 @@ class Object3D:
         return [
             self.type, self.id, self.parent.id, self.args,
             self.name,
-            self.color, self.opacity, self.side_,
+            self.color, self.opacity, self.side_, self.material_is_set,
             self.x, self.y, self.z,
             self.R,
             self.sx, self.sy, self.sz,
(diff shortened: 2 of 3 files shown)


fabian0702:
@evnchn I can confirm the issue is fixed on the branch you provided, from my side this looks complete / functional. Thanks for the quick response.

falkoschindler (approved):
Looks great, thanks @evnchn! The `material_is_set` flag is the right call here — `None` is already taken for vertex colors, so a separate flag is the only way to encode "explicitly set" across the JSON boundary. The fix also correctly closes the async-load path, since an unset GLTF never queues a default material into `pendingMaterialInfo`.

I pushed a small follow-up that merges `test_gltf` and `test_gltf_keeps_own_material` into one parametrized test: one case sets the material explicitly (expects `ff0000`), the other keeps `box.glb`'s own material (expects `e70000`, the sRGB encoding of its `baseColorFactor` `0.8`). Both now assert the rendered hex via `getHexString()`.

Approving. 🟢

PR #6130, opened 2026-06-29
Optimize hot paths in binding, observables, and element classes

jwinpbe:
### Motivation

I recently learned about a feature in the Python 3.15 beta, [Tachyon](https://docs.python.org/3.15/library/profiling.sampling.html), which enables statistical profiling of a running python process by continually sampling it's call stack externally, allowing for inspection of code as it runs (without any added computational load on the program) instead of the old deterministic way.

A friend let me use his wife's $20 chatGPT subscription for programming because I helped him out with his home server, and I wanted to see how it would do when combined with this new ability.

So, over the last week and a half I have been iterating with GPT 5.5 over NiceGUI's hot paths using a combination of microbenchmarks, internally constructed NiceGUI programs using the internal script mode, and a program that I am in the process of writing.

What we found is that for a lot of operations on elements, events, bindings, and options, almost every hot path was running the full pipeline unconditionally, allocating event objects even if nothing was listening for them, rebuilding props when nothing changed, and iterating over every binding to just remove a single one, instead of intelligently checking if work actually had to be done or not.

I understand that open source repositories get a lot of slop PRs, and I hope to differentiate myself from those with the methodology I used.

Please let me know if you want to cherry pick any of these instead of taking the whole thing and I'll do my best to break it up for you.

I took a little extra time and created two scripts that you can run in the root of my branch to see the profiling methodology, A / B benchmarking, code snippets, and test individually grouped optimizations.

For your minions: Both of the scripts can be run with `--agent` in order to get information dense output, and with `--agent --terse` to just get the A / B profile.

You will need uv in order to use these, as it uses Python inline script metadata to fetch Python 3.15b2 and run the profiler script out of a temporary virtual environment while running it against the repo in the current working directory. I made it this way to make it absolutely painless for you to get A / B comparisons.

I made this as a draft because it covers a lot of ground and I don't want to assume it's fully PR ready.

I ran the full testing / linting suite on it and all the tests passed. We added tests as needed to cover new behavior.

The two scripts and their common util file:

[maintainer_common.py](https://gist.github.com/jwinpbe/ee41ac7bebb3ee2493221553a6f7d339)
[maintainer_benchmark.py](https://gist.github.com/jwinpbe/252501476c880494fd28bf3aa3be437f)
[maintainer_profiler.py](https://gist.github.com/jwinpbe/5ee2bf710fc2b6e69f839d9c5ed496da)

The link to the raw scripts so you can wget / curl -O them to the root of the repo easily are here:

[maintainer_common.py](https://gist.github.com/jwinpbe/ee41ac7bebb3ee2493221553a6f7d339/raw/6324ec869665b77aaf201e44f6a29556d2bfd9a4/maintainer_common.py)
[maintainer_benchmark.py](https://gist.github.com/jwinpbe/252501476c880494fd28bf3aa3be437f/raw/5eceeeb7fd45fc8cf1462f3c2fe830b4b2b821b4/maintainer_benchmark.py)
[maintainer_profiler.py](https://gist.github.com/jwinpbe/5ee2bf710fc2b6e69f839d9c5ed496da/raw/d3e8714ae00cf92ac18a93ac16fc299640916f68/maintainer_profiler.py)

### Implementation

21 commits across 27 files:

#### Hot path optimization

**`expects_arguments()` caching** - `nicegui/helpers/functions.py`

Every event handler dispatch calls `expects_arguments()` to figure out whether to pass event args or call the handler with no args. This was doing `inspect.signature()` on every call. Now it:

1. checks if the callable is a plain `FunctionType` or `MethodType` (most common case) and inspects `co_argcount` / `co_varnames` directly from metadata, no `inspect` module needed. This is a low level python trick that saves time in hot loops because of how often `expects_arguments()` is called during event dispatch.
2. falls back to a weak-ref cache keyed on `id(func)`, so repeated calls to the same callback skip `inspect.signature()` entirely
3. only uses the slow path for callables with custom `__signature__` or `__wrapped__`

Also hoisted an early return for non-awaitable results in should_await(), avoiding isinstance(Awaitable) checks (and their __subclasscheck__ overhead) for common cases like None or return values.

**Event dispatch inlining** - `nicegui/event.py`

The old `emit()` called `_invoke_and_forget()` per callback, which called `callback.run()`, which resolved the slot stack via `(self.slot and self.slot()) or nullcontext()`. That's three levels of indirection on the hot path.

`emit()` inlines slot-stack resolution and callback dispatch, hoisting the stack out of the loop. `Callback._resolved_slot()` replaces the `nullcontext` ternary duplicated in `run()` and `await_result()`. `subscribe()` uses `Slot.peek_stack()` instead of `Slot.get_stack()` to avoid auto-creating a stack, and extracts the event-forwarding detection inline into `_should_forward_event_args()`.

**Observable change dispatch** - `nicegui/observables.py`

Root collections with no change handlers short-circuit `_handle_change()`. Single-handler root collections try a new `_handle_direct_change_handler()` hook first, for Props/Classes/Style this skips `ObservableChangeEventArguments` construction and expects_arguments checks by calling _update() directly.

Props, Classes, and Style implement this hook to detect their own `_update` handler and call it directly.

**Immutable-type fast path** - same file

Added `_OBSERVABLE_UNCHANGED_TYPES` check so deeply-immutable types (str, int, float, bool, tuple, bytes, frozenset, NoneType) never get wrapped.

**Slot stack caching** - `nicegui/slot.py`

Added `_slot_stack_cache` as a `contextvars.ContextVar` that caches `(task_id, stack)` tuples. Most slot operations happen in the same asyncio task, so after the first lookup we skip `dict.get()` entirely. Added `peek_stack()` for non-invasive checks, used in `element.py`, `event.py`, and `style.py` to avoid creating slot stacks when there's no UI context.

#### Element construction and serialization

**Element `__init__()` and helpers** - `nicegui/element.py`

Tag validation is cached via `_validate_tag()` so the regex match runs at most once per distinct tag. Constructor operations are reordered to minimize attribute lookups: `component`, `element_id`, `client`, `outbox`, and `parent_slot` are all resolved once and reused. `_to_dict()` builds its return value sparsely with conditional appends, skipping empty collections and the non-default slot dict when no named slots exist.

Extracted `_safe_client()` helper that returns the element's client or `None` (replacing the boolean `_is_safe_to_interact()`), so `update()` eliminates the redundant client re-fetch after the safety check. `Eventlistener.id` is deterministic (`element_id:listener_index`) instead of UUID-based. `on()` caches camel-case conversion for event names without hyphens.

#### Props / Classes / Style refactoring

**`ElementBoundObservableMixin`** - extracted from `nicegui/observables.py`

Props, Classes, and Style each carry identical `_element` weakref, `_suspend_count`, `suspend_updates()`, and `element` property. Extracted into a mixin so the observable change dispatch can hook into `_handle_direct_change_handler()` and shortcut `_update` calls without building `ObservableChangeEventArguments` - see "Observable change dispatch" above.

Props also got minor fixes: `set_optional()` and `set_bool()` now check `elif key in self` before popping instead of calling `pop(key, None)` unconditionally. `add_warning()` and `add_rename()` skip their check if the prop key doesn't exist.

#### Choice element options snapshotting

**`nicegui/elements/choice_element.py`**

Before this, every `update()` call rebuilt all options from scratch and sent them to the client, even if nothing changed. Now:

1. captures a snapshot of `_values` and `_labels` at construction and after each `_update_options()`
2. `_options_have_changed()` compares current options against the snapshot
3. `update()` only rebuilds if options actually changed, and does it under `suspend_updates()` to batch the prop emission

Snapshots use `deepcopy` when options contain mutable objects, fall back to shallow copy for immutable types (str, int, float, bool, None), and catch `deepcopy` errors gracefully (some valid option objects aren't copyable).

#### Defaults decorator fast path

**`nicegui/defaults.py`**

The `@resolve_defaults` decorator was doing `Signature.bind()` on every call, even when no sentinel values were present. New implementation:

1. precomputes `_DefaultPropParam` tuples from the function signature at decoration time
2. precomputes `fallback_kwargs_by_arg_count` - a tuple indexed by positional arg count, so calls with no custom defaults and no kwargs can skip `Signature.bind()` entirely and just pass precomputed fallbacks
3. only falls back to `Signature.bind()` when: there are positional-only default-prop params, no positional args are supplied, or an explicit positional arg IS a SentinelValue

#### Bug fixes

- **`nicegui/page.py`** - fixed infinite reload loop when async page build exceeds `response_timeout`. Now serves a 500 error page through a fresh, alive client instead of the deleted one.

#### Other changes

- **`nicegui/run.py`** - return type of `cpu_bound`/`io_bound` changed to `R | None` (returns `None` when app is stopping or call is cancelled). Interim shape before 4.0 raises `CancelledError`.
- **`nicegui/helpers/strings.py`** - `event_type_to_camel_case` now has `@lru_cache(256)`.
- **`nicegui/outbox.py`** - renamed `coros` to `emit_coroutines` and `data` to `update_payload` for readability.
- **`nicegui/helpers/network.py`** - simplified `is_port_open` with context manager, simplified `schedule_browser` by removing unnecessary argument passing.
- **`nicegui/welcome.py`** - added `or []` defensive fallback for `io_bound()` returning `None` when app is stopping.
- **`tests/test_helpers.py`** - changed `def test_is_port_open...` to not try to reach '1.2', as that was being recognized by Linux as a valid attempt to reach '1.0.0.2' and would hang until kernel timeout. Updated name of the test def to better reflect it's new behavior.

#### Tests

7 test files covering: defaults sentinel resolution, select option handling, background task slot context, expects_arguments caching, response_timeout error page, cpu_bound/io_bound None return, socketio payload length.

### Progress

- [X] The PR title is a short phrase starting with a verb like "Add ...", "Fix ...", "Update ...", "Remove ...", etc.
- [ ] The implementation is complete. (Otherwise, open a [draft PR](https://github.blog/news-insights/product-news/introducing-draft-pull-requests/).)
- [X] This PR does not address a security issue. (Security fixes must be coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process before opening a PR.)
- [X] Pytests have been added/updated.
- [X] No breaking changes to the public API or migration steps are described above.


falkoschindler:
Thanks a lot for the careful write-up and especially for the profiling/benchmark tooling, @jwinpbe! The Tachyon-based A/B methodology is exactly the kind of rigor we like to see, and it clearly sets this apart from a "slop" PR. A few high-level things we need to sort out before we can dig into the individual optimizations:

**1. The description doesn't match the diff.** The "Bug fixes" and "Other changes" sections describe work in `page.py` (the `response_timeout` reload-loop fix), `run.py` (`cpu_bound`/`io_bound` → `R | None`), and `welcome.py` (the `or []` fallback) — but none of those files are actually in this PR's diff against `main`. They're already on `main`: the `page.py` fix landed in #6127 and the `run.py`/`welcome.py` change in #6056, and your branch merged `main`, so `git diff main...HEAD` is empty for all three. It looks like the description was generated from the branch state (including merged-in `main` commits) rather than from the diff. Could you regenerate the description so it reflects only what this PR actually changes, and double-check each remaining item against the real diff? That'll make the review much easier to trust.

**2. Please put the numbers in the description.** You've clearly done the measurements — the profiling/benchmark scripts are attached — but the actual results aren't summarized in the PR itself. We'd like to see, as part of the description, what each optimization buys: before/after in absolute terms (µs/ms, not just %). That's what lets us judge relevance and prioritize each change against its complexity — e.g. the `binding.py` reverse-index + `_BindablePropertyRegistry` rewrite is a substantial increase in surface area, so a clear scaling win on `remove()`/large binding graphs would justify it, whereas a few-µs win wouldn't.

**3. Please split it up.** This is ~1200 added lines across ~19 effective files, spanning binding, observables, defaults, element construction, event dispatch, slots, and assorted cleanups. That's too much to review or bisect as one unit. We'd much rather take these as a series of focused PRs, each with its own benchmark and tests, so we can merge the clear wins quickly and discuss the riskier ones on their own. A possible grouping:

- `binding.py` registry + reverse-index `remove()`
- `resolve_defaults` precompute path (`defaults.py`)
- `ElementBoundObservableMixin` extraction + observable dispatch short-circuit + immutable-type fast path
- `expects_arguments` fast path + cache (`helpers/functions.py`) — see point 4
- event-dispatch inlining (`event.py`)
- slot-stack caching (`slot.py`)
- `choice_element` options snapshot
- element `__init__`/`_to_dict` + deterministic `EventListener.id` + `on()` camel-case caching
- the small cleanups (`strings` `lru_cache`, `outbox` renames, `network`) as one tidy PR

Each of those should be small and reviewable on its own, with its own measurement. Happy to cherry-pick — you offered, and yes please.

**4. Overlap with #6108.** The `expects_arguments` caching here overlaps with @maybites' open PR #6108 ("Resolve expects_arguments once when observable handlers are registered", touching `observables.py` + `events.py`). Same goal, different approach. Let's coordinate so we don't land two competing mechanisms — worth deciding which approach to take before either merges.

**5. CI lint is red.** `pylint` currently fails (ruff and mypy pass):

- `nicegui/binding.py` — `clear()` and `discard_object_ids()` are missing docstrings (C0116)
- `nicegui/observables.py:69` — unused argument `handler` in the base `_handle_direct_change_handler` (W0613)
- `nicegui/helpers/functions.py` — `_CallableCache.get()` and `.set()` are missing docstrings (C0116)

Once it's split, we can give each piece a proper line-level review — there are a handful of small things (some genuinely subtle, e.g. the slot-stack cache and the deterministic listener IDs) but nothing that looks like a hard correctness bug so far, and the binding rewrite held up under scrutiny.

**Appendix — line-level findings so far (for our own reference; fold into the per-PR reviews)**

- **`slot.py`** — `_get_or_create_stack()`/`peek_stack()` return the cached stack without revalidating against `cls.stacks`. With `__exit__` no longer pruning + `get_task_id()` being `id(current_task())` (reusable) + the cache being a ContextVar copied into child tasks, there's a low-probability path to handing a task a dead task's stack. Latent; worth a revalidation guard.
- **`slot.py:90`** — `prune_stack()` (singular) is now dead code; `__exit__` dropped its only caller (only `prune_stacks()` plural is used, by `app.py`).
- **`helpers/functions.py`** — `_CallableCache` hand-rolls a weak-id cache; a `WeakKeyDictionary` would be simpler and avoid the `id()`-reuse guard it needed.
- **`event_listener.py`** — IDs moved from UUID to `element_id:index`; unique today, but breaks if single-listener removal is ever added (frontend uses the id as a throttle key). Worth a comment or a monotonic counter.

jwinpbe:
Thanks for the feedback, I’ll split it up and get back to you, i had the merge head and upstream confabulated in one of the last reviews because i pulled in upstream!

PR #6134, opened 2026-06-30
Reject ui.sub_pages parameter names that are not valid identifiers

### LINKED ISSUE #6133: ui.sub_pages accepts digit-leading parameter names that crash at navigation
*Drafted by Claude Code on @evnchn's behalf.*

### First Check

- [X] This is not a security issue.
- [X] I used the GitHub search to find a similar issue and came up empty.

### Example Code

```python
from nicegui import ui

@ui.page('/')
def index():
    ui.sub_pages({'/{1a}': lambda: ui.label('hi')})  # digit-leading parameter name

ui.run()
```

### Description

Open `/`. The page returns **HTTP 500** and the server logs `re.error: bad character in group name '1a'`.

A `ui.sub_pages` route whose parameter name starts with a digit (`{1a}`, `{2}`, …) is accepted at registration, but the name is not a valid Python regex group name. When `_show()` compiles the route pattern into `(?P<1a>…)` on the first render, `re.compile` raises — so the whole page 500s on load (no navigation needed).

- Expected: rejected at registration with the clear `ValueError` that `_validate_route` already raises for other unsupported patterns (e.g. `{name:path}`).
- Actual: accepted at registration, then an uncaught `re.error` at render time.

Root cause: `_validate_route` (`nicegui/elements/sub_pages.py:198`) checks names with `re.fullmatch(r'\w+', parameter)`, and `\w` allows a leading digit. PR #6092 added `_validate_route` precisely to reject unsupported route patterns at registration — this digit-leading class slips through.

<details>
<summary>Server traceback (abridged)</summary>

```
  File "nicegui/elements/sub_pages.py", line ..., in _match_path
    regex_match = re.match(f'^{regex_pattern}$', path)
  ...
re.error: bad character in group name '1a' at position 6
```
(GET `/` → 500.)
</details>

### Fix

Validate parameter names with `str.isidentifier()` instead of `re.fullmatch(r'\w+', …)`. `isidentifier()` rejects digit-leading names (and `a-b`, `name:path`, empty) at registration, while still accepting valid non-ASCII identifiers (`{café}`, `{π}`, `{变量}`) that match correctly today — which a `[A-Za-z_]\w*` regex would wrongly reject. PR #6134.

### NiceGUI Version

3.13.0.post24.dev0+c40f3e3f (main / 3.14 RC)

### Python Version

3.12.11

### Browser

Other — fails server-side on page load; browser-independent.

### Operating System

macOS

### Additional Context

Severity is minor (requires a malformed route in app code), but it turns what should be a clear registration-time error into a deferred 500 on page load. Fix + regression test in #6134.

evnchn:
*Drafted by Claude Code on @evnchn's behalf.*

### Motivation

Fixes #6133. `ui.sub_pages` route validation accepts digit-leading parameter names like `{1a}` / `{2}` at registration (`\w` allows a leading digit), but those are invalid Python regex group names — so the first navigation matching that prefix crashes with an uncaught `re.error: bad character in group name` from `_match_path`. PR #6092 added `_validate_route` to reject unsupported patterns at registration; this class slipped through.

### Implementation

Validate parameter names with `str.isidentifier()` instead of `re.fullmatch(r'\w+', ...)`. `isidentifier()` rejects digit-leading names (and `a-b`, `name:path`, empty) at registration with the existing helpful `ValueError`, while still accepting valid non-ASCII identifiers (`{café}`, `{π}`, `{变量}`) that match correctly today — which a `[A-Za-z_]\w*` regex would have wrongly rejected.

Added the digit-leading cases to `test_validate_route_rejects_unsupported_patterns`.

<details>

```
/u/{1a}  before: accepted by _validate_route, then re.error at match → after: rejected at registration
/u/{2}   same
/u/{name}, /u/{_x}, /u/{café}, /u/{π}, /u/{变量} → still accepted, match OK
```

Gates: `pre-commit` clean, `mypy ./nicegui` clean (243 files), `pylint nicegui/elements/sub_pages.py` 10.00/10, `pytest tests/test_sub_pages_match_path.py tests/test_sub_pages.py` 71/71. Independently confirmed (incl. the `isidentifier` vs regex unicode point) by a second reviewer (different model lineage).
</details>

### Progress

### DIFF
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -212,7 +212,7 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
             return {} if pattern == path else None
 
         regex_pattern = re.escape(pattern)
-        for match in re.finditer(r'\\{(\w+)\\}', regex_pattern):
+        for match in re.finditer(r'\\{(.*?)\\}', regex_pattern):
             param_name = match.group(1)
             regex_pattern = regex_pattern.replace(f'\\{{{param_name}\\}}', f'(?P<{param_name}>[^/]+)')
 
--- a/nicegui/elements/sub_pages.py
+++ b/nicegui/elements/sub_pages.py
@@ -197,10 +197,10 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
     @staticmethod
     def _validate_route(path: str) -> None:
         for parameter in re.findall(r'\{(.*?)\}', path):
-            if not re.fullmatch(r'\w+', parameter):
+            if not parameter.isidentifier():
                 raise ValueError(
                     f'Invalid route "{path}": the parameter "{{{parameter}}}" is not supported. '
-                    'ui.sub_pages only supports single-segment "{name}" parameters, '
+                    'ui.sub_pages only supports single-segment "{name}" parameters that are valid Python identifiers, '
                     'not Starlette-style converters like "{name:path}". '
                     'For wildcard routing, use show_404=False and read PageArguments.remaining_path '
                     '(see https://nicegui.io/documentation/sub_pages).'
--- a/tests/test_sub_pages_match_path.py
+++ b/tests/test_sub_pages_match_path.py
@@ -147,16 +147,8 @@ def test_special_characters_in_patterns():
 
 
 def test_invalid_parameter_names():
-    """Parameter names that are not valid Python identifiers cannot form regex group names.
-
-    Such names never reach `_match_path` in practice: `_validate_route` rejects them at registration
-    because they fail `isidentifier()`. `_match_path`'s extraction regex `\\{(.*?)\\}` deliberately
-    mirrors `_validate_route`'s, so the two agree on what a valid parameter name is and there is no
-    "passes validation but silently 404s" gap. Calling `_match_path` directly on an unvalidated,
-    non-identifier name raises `re.error` (loudly) rather than silently mis-matching.
-    """
-    for pattern in ['/path{123}', '/path{123abc}', '/path{param-name}',
-                    '/path{param.name}', '/path{param space}']:
+    """Non-identifier parameter names are rejected at registration and raise re.error if matched directly."""
+    for pattern in ['/path{123}', '/path{123abc}', '/path{param-name}', '/path{param.name}', '/path{param space}']:
         with pytest.raises(ValueError, match='not supported'):
             ui.sub_pages._validate_route(pattern)
         with pytest.raises(re.error):
@@ -164,13 +156,7 @@ def test_invalid_parameter_names():
 
 
 def test_validation_and_matching_agree_on_non_ascii_identifiers():
-    """A name that is a valid Python identifier but contains characters outside regex `\\w`
-    (e.g. middle dot U+00B7) must both pass validation AND be matchable.
-
-    These previously diverged: `_validate_route`'s `isidentifier()` accepted `{a·b}`, but
-    `_match_path`'s old `\\w+` extraction silently dropped the group, so every navigation fell
-    through to 404 with no error. Broadening the extraction regex aligns them (PR #6134).
-    """
+    """Identifiers with characters outside regex \\w (e.g. middle dot) must both validate and match."""
     pattern = '/u/{a·b}'  # 'a·b' — '·' is a valid identifier char but not a regex \w char
     ui.sub_pages._validate_route(pattern)  # must not raise
     assert ui.sub_pages._match_path(pattern, '/u/foo') == {'a·b': 'foo'}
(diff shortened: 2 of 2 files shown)


falkoschindler:
Thanks for catching the underlying bug and turning it around into a fix so quickly, @evnchn! One thing worth a closer look before merging:

`isidentifier()` accepts a superset of what `_match_path` can actually handle, which trades the old crash for a silent always-404.

Before this PR, the validation regex `\w+` matched `_match_path`'s extraction regex `r'\\{(\w+)\\}'` (line 215) exactly, so passing validation guaranteed the route was matchable. `isidentifier()` also accepts identifier characters that aren't in `\w` — e.g. middle dot `{a·b}` or a combining accent. Those pass validation here, but `_match_path` never substitutes the group, so the route is silently registered and never matches (every navigation falls through to 404, no error).

```python
SubPages._validate_route('/u/{a·b}')          # passes (·  is a valid identifier char)
SubPages._match_path('/u/{a·b}', '/u/foo')    # -> None  (\w+ can't extract the group)
```

Obscure, but it's a real divergence between the validation contract and the matcher. If you want them aligned, the cleaner fix is to broaden `_match_path`'s extraction regex (e.g. `\{(.*?)\}`, like `_validate_route` already uses) and let `isidentifier()` remain the single source of truth — rather than leaving the two functions with different notions of a valid parameter name.

evnchn:
**TL;DR:**

Good catch — fixed it your way. Broadened `_match_path`'s extraction to `\{(.*?)\}` so `isidentifier()` is the single source of truth, with a regression test for the `{a·b}` case.

A nice bonus fell out: the two aren't just *aligned*, they're identical **by construction** — CPython validates `(?P<name>…)` group names with `str.isidentifier()` (`re/_parser.py`), the exact call `_validate_route` uses. So validator and matcher now accept the same character set, no superset gap left to drift.

Separately, a review pass flagged an *unrelated* pre-existing footgun (constructor aliases the caller's `routes` dict) — surfacing it below for your call; I left it out of this commit to keep the diff scoped to your point.

<details><summary><b>The fix + before/after</b></summary>

```python
# _match_path extraction regex:
- for match in re.finditer(r'\\{(\w+)\\}', regex_pattern):
+ for match in re.finditer(r'\\{(.*?)\\}', regex_pattern):
```

```python
# before this commit
SubPages._validate_route('/u/{a·b}')        # passes  (· is a valid identifier char)
SubPages._match_path('/u/{a·b}', '/u/foo')  # -> None (silent 404, \w+ drops the group)

# after
SubPages._validate_route('/u/{a·b}')        # passes
SubPages._match_path('/u/{a·b}', '/u/foo')  # -> {'a·b': 'foo'}
```
</details>

<details><summary><b>Why they're now identical, not just "broadened to match"</b></summary>

CPython's regex parser checks group names with `str.isidentifier()`:

```python
# re/_parser.py, checkgroupname():
if not name.isidentifier():
    raise self.error("bad character in group name %r" % name, ...)
```

`_validate_route` gates on `parameter.isidentifier()`; `_match_path` feeds the extracted name straight into `(?P<name>…)`, which re-applies the same `isidentifier()`. Spot-checked the equivalence across `a·b`, `á`, `naïve`, `日本語`, Devanagari, `_x`, `x2` — no divergence in either direction. So there's no remaining "valid here but not there" gap to regress on.
</details>

<details><summary><b>Test changes</b></summary>

- Rewrote `test_invalid_parameter_names`: the old assertions encoded the buggy behavior (names like `{param-name}` silently registering as literal text). Now it asserts the validation gate rejects them (`ValueError`, "not supported") **and** that a direct `_match_path` call on such a name fails loud (`re.error`) rather than silently mis-matching.
- Added `test_validation_and_matching_agree_on_non_ascii_identifiers` as the regression lock for the `·` case.
- Full `tests/test_sub_pages*.py` (72 tests) green; `pre-commit` / `mypy ./nicegui` / `pylint ./nicegui` (10.00) all clean.
</details>

<details><summary><b>Unrelated aside (your call) — constructor aliases the routes dict</b></summary>

Not part of this fix, and pre-existing — but a second reviewer noticed `__init__` stores the caller's dict by reference:

```python
self._routes = routes or {}
```

So a caller can bypass `_validate_route` by mutating the *same* dict after construction:

```python
routes = {'/': home}
pages = ui.sub_pages(routes)
routes['/u/{a·b:int}'] = bad   # never validated
pages.refresh()                # _match_path now sees an unvalidated pattern
```

Harmless today (worst case a `re.error` / 404 on clearly-malformed input), and arguably user misuse. If you'd want the "only validated patterns reach `_match_path`" contract airtight, a one-liner does it: `self._routes = dict(routes or {})`. Happy to fold that in here, spin it into its own PR, or drop it — whichever you prefer.
</details>

falkoschindler (approved):
Perfect, this is exactly the alignment I was hoping for — thanks for the quick turnaround.

And nice catch on the *identical by construction* angle: since `(?P<name>…)` runs the name through the same `str.isidentifier()` check in CPython's `re._parser`, the validator and matcher can't drift apart — there's no superset gap left to regress on. That's a stronger guarantee than just "broadened to match".

I pushed one tiny commit onto the branch condensing the two new test docstrings to one-liners (to match the style of the rest of the file) — no behaviour change.

On the constructor aliasing aside: let's drop it. Mutating the passed-in dict after construction to sneak past `_validate_route` is user misuse, and the worst case is a loud `re.error` / 404 on clearly-malformed input — not worth a defensive copy here. Keeping this PR scoped to the actual bug.

LGTM — approving. 🎉

Issue #4711, opened 2025-05-07
Number of "active links" keeps growing


defkev:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python

```

### Description

I am using binding properties (with the decorator class `@binding.bindable_dataclass` added in v2.11) to update values using the `bind_value` methods for, probably a couple of 1000 values at times, which is working just fine.

I am also using the `bind_text_from` method to update some (<100) text, like `ui.labels` in a `dialog`, which keeps throwing:

> WARNING: binding propagation for N active links took Ts

where N keeps progressively increasing the longer the app is running, assuming each time the dialogue is opened, which of course is also causing CPU util to progressively increase.

Is there any way to actual see the “active links” maybe like in react's Why did you re-render? Or do I actively have to “unbind” propagations when closing the dialogue, which IMHO isn't documented?

Note I am aware of the `binding.MAX_PROPAGATION_TIME` constant, this isn't about suppressing the warning by increasing the threshold but actually figuring out why the number of “active links” keeps increasing.

Cheers

EDIT
I just tested my theory of “orphan” bindings by subsidizing `dialog.close` with `delete` and it keeps throwing the same WARNING with N “active links” as added by the dialogue, so there is definitely something amiss with your clean-up logic.

### NiceGUI Version

current

### Python Version

3

### Browser

Other

### Operating System

Linux

evnchn:
@defkev Thanks you, but we need more information. See point 1. The rest is my speculation. 

1. **I can't see how I can debug this without your example code. You need to attach something which I and others can immediately run and encounter the warning. Then, we can debug stuff.** 
2. Not sure why but if you use `@binding.bindable_dataclass`, `ui.label` and assuming you did the binding appropriately using the bind methods, then they should not be active links. Aka using method 1 not method 2 in https://nicegui.io/documentation/section_binding_properties#bindable_properties_for_maximum_performance. Are you sure you bound stuff properly? 
3. You may inject a print statement at the following code to figure out what is in `active_links`. We don't have "Why did you re-render?" (because we don't use React but we use Vue) but that's about the closest we have. 
[...]

PR #6135, opened 2026-06-30
Optimize binding cleanup with reverse index

### LINKED ISSUE #6150: N refreshables need N^2 CPU
### First Check

### Example Code

```python
# /// script
# dependencies = ["nicegui", "psutil"]
# ///
import sys
import time

import psutil
from nicegui import app, ui

R = int(sys.argv[1]) if len(sys.argv) > 1 else 50
K = 50


def make_card() -> None:
    @ui.refreshable
    def content() -> None:
        now = time.strftime('%H:%M:%S')
        for i in range(K):
            ui.label(f'label {i}: {now}')
    content()
    ui.timer(1.0, content.refresh)


with ui.grid(columns=10):
    for _ in range(R):
        with ui.card():
            make_card()

process = psutil.Process()
app.timer(1.0, lambda: print(f'R={R}: process CPU: {process.cpu_percent():.0f}%', flush=True))

ui.run(show=False, reload=False, port=8123)
```

### Description

Start the app and observe CPU utilization when changing R. Even with the default this needs 10% cpu on my fairly capable machine. 50*50 refreshable labels is admittedly unusual, but not inconceivable: I've encountered this issue in production code. In practice the time scales with `#refreshable*#links`.

In broad terms, the issue is that `_execute_refresh`->`clear`->`remove_elements`->`remove` iterates over all bindings, and this is effectively called for all of the items/labels.
I don't have an immediate suggestion on how to fix this, but also haven't thought about this hard. I'll work around this by explicit timers and updates for now.

Pure benchmark code:
```python
import time

from nicegui import binding


class Obj:
    value = binding.BindableProperty()

    def __init__(self) -> None:
        self.value = 0


K = 10  # elements per refreshable
for R in [10, 100, 1000, 10000]:
    groups = [[Obj() for _ in range(K)] for _ in range(R)]
    t0 = time.perf_counter()
    for group in groups:  # one round: every refreshable refreshes once
        binding.remove(group)
        for obj in group:
            obj.value = 0  # re-register, like the rebuilt elements would
    dt_ms = (time.perf_counter() - t0) * 1000
    print(f'R={R:>5} refreshables ({R * K:>6} entries): one round took {dt_ms:9.2f} ms '
          f'({dt_ms / R:.3f} ms per refresh)')
    for group in groups:
        binding.remove(group)

```

### NiceGUI Version

3.14.0

### Python Version

3.13.14

### Browser

Firefox

### Operating System

Linux

### Additional Context

_No response_

jwinpbe:
### Motivation

`binding.remove()` used to scan all binding lists and all bindable-property keys for every removal. That makes cleanup scale with unrelated global binding state, which is expensive during dynamic UI rebuilds or repeated element deletion. Most NiceGUI app shapes use callbacks rather than bindings, so deleting unbound elements paid the full scan cost without ever having had a binding. This PR makes removal proportional to the removed objects' own bindings — O(1) for objects without any.

This PR focuses on the binding cleanup path only.

### Implementation

Add `_binding_keys_by_object`, a reverse index mapping each endpoint's object ID to the binding keys that may reference it. New bindings register both endpoints through the shared `_bind_one_way()` helper (deduplicated from `bind_to()`/`bind_from()`), which appends to `bindings`, indexes both endpoints, and maintains `active_links`.

`remove()` pops the removed objects' reverse-index entries, prunes or deletes only the affected binding lists, and discards stale index entries for deleted objects. Objects without any bindings take an early return without touching the rest of the global binding state.

`bindable_properties` remains a stdlib `WeakValueDictionary`, and `remove()` no longer deletes its entries — the weak markers expire when their objects are garbage-collected. This intentionally changes one edge case: re-binding a removed-but-still-alive bindable object no longer creates a polling active link, since its property descriptor still propagates on assignment (pinned down by `test_remove_rebind_preserves_bindable_property_semantics`).

Benchmarks (baseline `034db3b7a`, candidate `50bf6f287`, Python 3.11.7, median of 9 samples per case, GC disabled during each timed sample, timing only the `remove()` calls):

| case                                                                                         |      before |     after |      delta |
| -------------------------------------------------------------------------------------------- | ----------: | --------: | ---------: |
| remove 300 of 10,000 unlinked bindable objects, one `remove()` call each                     | `112.95 ms` | `0.07 ms` |   `-99.9%` |
| remove 1,000 bound targets of one source in a single call                                    |   `0.13 ms` | `0.27 ms` | `+0.14 ms` |
| remove 1,000 never-registered objects in a single call, with 15,000 bindable objects present |   `0.67 ms` | `0.06 ms` |     `-92%` |

The bound-target batch case pays for maintaining and consulting the reverse index; the absolute cost is about `0.14 ms` per 1,000 bound targets and does not grow with unrelated binding state, while the scan-dominated cases collapse to near-constant time.

### Progress

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -30,8 +30,7 @@ bindings: defaultdict[BindingKey, list[Binding]] = defaultdict(list)
 bindable_properties: weakref.WeakValueDictionary[BindingKey, Any] = weakref.WeakValueDictionary()
 active_links: list[ActiveLink] = []
 _active_links_added = asyncio.Event()
-# Maps object IDs to binding keys that reference them, so remove() avoids scanning all bindings.
-_binding_keys_by_object: dict[int, set[BindingKey]] = {}
+_binding_keys_by_object: defaultdict[int, set[BindingKey]] = defaultdict(set)
 
 TC = TypeVar('TC', bound=type)
 T = TypeVar('T')
@@ -39,16 +38,8 @@ T = TypeVar('T')
 _MISSING = object()
 
 
-def _index_binding(source_obj: Any, target_obj: Any, binding_key: BindingKey) -> None:
-    """Index both endpoints so remove() can find affected binding lists without a full scan."""
-    _binding_keys_by_object.setdefault(id(source_obj), set()).add(binding_key)
-    _binding_keys_by_object.setdefault(id(target_obj), set()).add(binding_key)
-
-
 def _discard_binding_key_from_object_index(obj_id: int, binding_key: BindingKey) -> None:
-    indexed_binding_keys = _binding_keys_by_object.get(obj_id)
-    if indexed_binding_keys is None:
-        return
+    indexed_binding_keys = _binding_keys_by_object[obj_id]
     indexed_binding_keys.discard(binding_key)
     if not indexed_binding_keys:
         del _binding_keys_by_object[obj_id]
@@ -59,7 +50,8 @@ def _bind_one_way(source_obj: Any, source_name: tuple[str, ...], target_obj: Any
     """Register a one-way binding and run its initial propagation."""
     binding_key = (id(source_obj), source_name)
     bindings[binding_key].append((source_obj, target_obj, target_name, transform))
-    _index_binding(source_obj, target_obj, binding_key)
+    _binding_keys_by_object[id(source_obj)].add(binding_key)
+    _binding_keys_by_object[id(target_obj)].add(binding_key)
     if binding_key not in bindable_properties:
         active_links.append((source_obj, source_name, target_obj, target_name, transform))
         _active_links_added.set()
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -6,9 +6,9 @@ import dataclasses
 import time
 import weakref
 from collections import defaultdict
-from collections.abc import Callable, Iterable, Mapping, MutableMapping
+from collections.abc import Callable, Iterable, Iterator, Mapping, MutableMapping
 from contextvars import ContextVar
-from typing import TYPE_CHECKING, Any, Literal, TypeVar
+from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
 
 from typing_extensions import dataclass_transform
 
@@ -18,23 +18,177 @@ from .logging import log
 if TYPE_CHECKING:
     from _typeshed import DataclassInstance, IdentityFunction
 
+ObjectId = int
+NamePath = tuple[str, ...]
+NameBucket = NamePath | set[NamePath]
+BindingKey = tuple[ObjectId, NamePath]
+BindingKeyBucket = BindingKey | set[BindingKey]
+Transform = Callable[[Any], Any] | None
+Binding = tuple[Any, Any, NamePath, Transform]
+ActiveLink = tuple[Any, NamePath, Any, NamePath, Transform]
+BindableEntry = tuple[weakref.ref[Any], NameBucket]
+
+
+class _BindablePropertyRegistry:
+    """Weak bindable-property registry keyed by object ID with name-path buckets.
+
+    Most bindable objects expose one bindable property, so the common case stores a single name path and upgrades to a
+    set only when the same object has multiple bindable properties.
+    """
+
+    def __init__(self) -> None:
+        self._entries_by_object_id: dict[ObjectId, BindableEntry] = {}
+
+    def _discard_object_id(self, obj_id: ObjectId) -> None:
+        self._entries_by_object_id.pop(obj_id, None)
+
+    def __setitem__(self, key: BindingKey, obj: Any) -> None:
+        obj_id, name_path = key
+        entry = self._entries_by_object_id.get(obj_id)
+        if entry is None:
+            def discard(_: weakref.ref[Any], obj_id: ObjectId = obj_id) -> None:
+                self._discard_object_id(obj_id)
+            self._entries_by_object_id[obj_id] = (weakref.ref(obj, discard), name_path)
+            return
+        ref, name_bucket = entry
+        if isinstance(name_bucket, set):
+            name_bucket.add(name_path)
+        elif name_bucket != name_path:
+            self._entries_by_object_id[obj_id] = (ref, {name_bucket, name_path})
+
+    def contains_key(self, key: BindingKey) -> bool:
+        """Return whether a validated binding key is registered."""
+        obj_id, name_path = key
+        entry = self._entries_by_object_id.get(obj_id)
+        if entry is None:
+            return False
+        ref, name_bucket = entry
+        if ref() is None:
+            self._discard_object_id(obj_id)
+            return False
+        return name_path in name_bucket if isinstance(name_bucket, set) else name_bucket == name_path
+
+    def __contains__(self, key: object) -> bool:
+        if not isinstance(key, tuple) or len(key) != 2:
+            return False
+        obj_id, name_path = key
+        if not isinstance(obj_id, int) or not isinstance(name_path, tuple):
+            return False
+        return self.contains_key((obj_id, cast(NamePath, name_path)))
+
+    def __iter__(self) -> Iterator[BindingKey]:
+        # Copy before iterating because weakref callbacks can mutate this dictionary.
+        for obj_id, (ref, name_bucket) in list(self._entries_by_object_id.items()):
+            if ref() is None:
+                self._discard_object_id(obj_id)
+                continue
+            if isinstance(name_bucket, set):
+                for name_path in name_bucket:
+                    yield obj_id, name_path
+            else:
+                yield obj_id, name_bucket
+
+    def clear(self) -> None:
+        """Remove all registered bindable-property entries."""
+        self._entries_by_object_id.clear()
+
+    def discard_object_ids(self, object_ids: Iterable[int]) -> None:
+        """Remove registered bindable-property entries for the given object IDs."""
+        pop = self._entries_by_object_id.pop
+        for obj_id in object_ids:
+            pop(obj_id, None)
+
+
 MAX_PROPAGATION_TIME = 0.01
 
-propagation_visited: ContextVar[set[tuple[int, tuple[str, ...]]] | None] = \
+propagation_visited: ContextVar[set[BindingKey] | None] = \
     ContextVar('propagation_visited', default=None)
 
 bindings: defaultdict[
-    tuple[int, tuple[str, ...]],
-    list[tuple[Any, Any, tuple[str, ...], Callable[[Any], Any] | None]]
+    BindingKey,
+    list[Binding]
 ] = defaultdict(list)
-bindable_properties: weakref.WeakValueDictionary[tuple[int, tuple[str, ...]], Any] = weakref.WeakValueDictionary()
-active_links: list[tuple[Any, tuple[str, ...], Any, tuple[str, ...], Callable[[Any], Any] | None]] = []
+bindable_properties: _BindablePropertyRegistry = _BindablePropertyRegistry()
+active_links: list[ActiveLink] = []
 _active_links_added = asyncio.Event()
+# Maps object IDs to binding keys that reference them, so remove() avoids scanning all bindings.
+_binding_keys_by_object: dict[ObjectId, BindingKeyBucket] = {}
 
 TC = TypeVar('TC', bound=type)
 T = TypeVar('T')
 
 _MISSING = object()
+
+
+def _discard_binding_key_from_object_index(obj_id: ObjectId, binding_key: BindingKey) -> None:
+    """Remove one binding key from the reverse index used by remove()."""
+    bucket = _binding_keys_by_object.get(obj_id)
+    if bucket is None:
+        return
+    if not isinstance(bucket, set):
+        if bucket == binding_key:
+            del _binding_keys_by_object[obj_id]
+        return
+    bucket.discard(binding_key)
+    if not bucket:
+        del _binding_keys_by_object[obj_id]
+
+
+def _add_binding_key_to_object_index(obj_id: ObjectId, binding_key: BindingKey) -> None:
+    """Add one endpoint reference to the reverse index used by remove()."""
+    bucket = _binding_keys_by_object.get(obj_id)
+    if bucket is None:
+        _binding_keys_by_object[obj_id] = binding_key
+    elif isinstance(bucket, set):
+        bucket.add(binding_key)
+    elif bucket != binding_key:
+        _binding_keys_by_object[obj_id] = {bucket, binding_key}
+
+
+def _index_binding(source_obj: Any, target_obj: Any, binding_key: BindingKey) -> None:
+    """Index both endpoints so remove() can find affected binding lists without a full scan."""
+    _add_binding_key_to_object_index(id(source_obj), binding_key)
+    _add_binding_key_to_object_index(id(target_obj), binding_key)
+
+
+def _bind_one_way(source_obj: Any, source_name: NamePath, target_obj: Any, target_name: NamePath,
+                  transform: Transform) -> None:
+    """Register a one-way binding and run its initial propagation."""
+    binding_key = (id(source_obj), source_name)
+    bindings[binding_key].append((source_obj, target_obj, target_name, transform))
+    _index_binding(source_obj, target_obj, binding_key)
+    if not bindable_properties.contains_key(binding_key):
+        active_links.append((source_obj, source_name, target_obj, target_name, transform))
+        _active_links_added.set()
+    _propagate(source_obj, source_name)
+
+
+def _collect_binding_keys_for_objects(object_ids: Iterable[ObjectId]) -> set[BindingKey] | None:
+    """Return binding keys whose source or target may reference the given objects."""
+    # Stay at None until the first hit so remove() calls for objects with no active
+    # bindings ("ghost removals") skip the set allocation and traversal entirely.
+    binding_keys: set[BindingKey] | None = None
+    get_object_binding_keys = _binding_keys_by_object.get
+    for obj_id in object_ids:
+        bucket = get_object_binding_keys(obj_id)
+        if bucket is None:
+            continue
+        if binding_keys is None:
+            binding_keys = set()
+        if isinstance(bucket, set):
+            binding_keys.update(bucket)
+        else:
+            binding_keys.add(bucket)
+    return binding_keys
+
+
+def _remove_active_links_for_objects(removed_object_ids: set[ObjectId]) -> None:
+    """Drop polling fallback links that reference any removed object."""
+    active_links[:] = [
+        (source_obj, source_name, target_obj, target_name, transform)
+        for source_obj, source_name, target_obj, target_name, transform in active_links
+        if id(source_obj) not in removed_object_ids and id(target_obj) not in removed_object_ids
+    ]
 
 
 def _get_attribute(obj: object | Mapping, name: tuple[str, ...]) -> Any:
@@ -173,11 +327,7 @@ def bind_to(self_obj: Any, self_name: str | tuple[str, ...], other_obj: Any, oth
     self_name_tuple = _normalize_name(self_name)
     other_name_tuple = _normalize_name(other_name)
     _check_self_and_other_attribute(self_obj, self_name_tuple, other_obj, other_name_tuple, self_strict, other_strict)
-    bindings[(id(self_obj), self_name_tuple)].append((self_obj, other_obj, other_name_tuple, forward))
-    if (id(self_obj), self_name_tuple) not in bindable_properties:
-        active_links.append((self_obj, self_name_tuple, other_obj, other_name_tuple, forward))
-        _active_links_added.set()
-    _propagate(self_obj, self_name_tuple)
+    _bind_one_way(self_obj, self_name_tuple, other_obj, other_name_tuple, forward)
 
 
 def bind_from(self_obj: Any, self_name: str | tuple[str, ...], other_obj: Any, other_name: str | tuple[str, ...],
@@ -202,11 +352,7 @@ def bind_from(self_obj: Any, self_name: str | tuple[str, ...], other_obj: Any, o
     self_name_tuple = _normalize_name(self_name)
     other_name_tuple = _normalize_name(other_name)
     _check_self_and_other_attribute(self_obj, self_name_tuple, other_obj, other_name_tuple, self_strict, other_strict)
-    bindings[(id(other_obj), other_name_tuple)].append((other_obj, self_obj, self_name_tuple, backward))
-    if (id(other_obj), other_name_tuple) not in bindable_properties:
-        active_links.append((other_obj, other_name_tuple, self_obj, self_name_tuple, backward))
-        _active_links_added.set()
-    _propagate(other_obj, other_name_tuple)
+    _bind_one_way(other_obj, other_name_tuple, self_obj, self_name_tuple, backward)
 
 
 def bind(self_obj: Any, self_name: str | tuple[str, ...], other_obj: Any, other_name: str | tuple[str, ...], *,
@@ -290,23 +436,47 @@ def remove(objects: Iterable[Any]) -> None:
 
     :param objects: The objects to remove.
     """
-    object_ids = set(map(id, objects))
-    active_links[:] = [
-        (source_obj, source_name, target_obj, target_name, transform)
-        for source_obj, source_name, target_obj, target_name, transform in active_links
-        if id(source_obj) not in object_ids and id(target_obj) not in object_ids
-    ]
-    for key, binding_list in list(bindings.items()):
-        binding_list[:] = [
-            (source_obj, target_obj, target_name, transform)
-            for source_obj, target_obj, target_name, transform in binding_list
-            if id(source_obj) not in object_ids and id(target_obj) not in object_ids
-        ]
-        if not binding_list:
+    # Keep IDs as a list until membership checks are needed; ghost removals only do dict pop/get by ID.
+    removed_object_ids = [id(obj) for obj in objects]
+    if not removed_object_ids:
+        return
+
+    bindable_properties.discard_object_ids(removed_object_ids)
+
+    affected_binding_keys = _collect_binding_keys_for_objects(removed_object_ids)
+    if not affected_binding_keys:
+        return
+
+    removed_object_id_set = set(removed_object_ids)
+    _remove_active_links_for_objects(removed_object_id_set)
+
+    for key in affected_binding_keys:
+        binding_list = bindings.get(key)
+        if binding_list is None:
+            continue
+        source_obj_id = key[0]
+        # Binding keys are source IDs; target-only removals only prune entries from the binding list.
+        if source_obj_id in removed_object_id_set:
+            for _, target_obj, _, _ in binding_list:
+                target_obj_id = id(target_obj)
+                if target_obj_id not in removed_object_id_set:
+                    _discard_binding_key_from_object_index(target_obj_id, key)
             del bindings[key]
-    for obj_id, name in list(bindable_properties):
-        if obj_id in object_ids:
-            del bindable_properties[(obj_id, name)]
+            continue
+        if len(binding_list) == 1:
+            if id(binding_list[0][1]) in removed_object_id_set:
+                del bindings[key]
+                _discard_binding_key_from_object_index(source_obj_id, key)
+            continue
+        remaining_bindings = [binding for binding in binding_list if id(binding[1]) not in removed_object_id_set]
+        if remaining_bindings:
+            binding_list[:] = remaining_bindings
+        else:
+            del bindings[key]
+            _discard_binding_key_from_object_index(source_obj_id, key)
+
+    for obj_id in removed_object_ids:
+        _binding_keys_by_object.pop(obj_id, None)
 
 
 def reset() -> None:
@@ -317,6 +487,7 @@ def reset() -> None:
     bindings.clear()
     bindable_properties.clear()
     active_links.clear()
+    _binding_keys_by_object.clear()
 
 
 @dataclass_transform()
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -38,36 +38,6 @@ T = TypeVar('T')
 _MISSING = object()
 
 
-def _discard_binding_key_from_object_index(obj_id: int, binding_key: BindingKey) -> None:
-    indexed_binding_keys = _binding_keys_by_object[obj_id]
-    indexed_binding_keys.discard(binding_key)
-    if not indexed_binding_keys:
-        del _binding_keys_by_object[obj_id]
-
-
-def _bind_one_way(source_obj: Any, source_name: tuple[str, ...], target_obj: Any, target_name: tuple[str, ...],
-                  transform: Callable[[Any], Any] | None) -> None:
-    """Register a one-way binding and run its initial propagation."""
-    binding_key = (id(source_obj), source_name)
-    bindings[binding_key].append((source_obj, target_obj, target_name, transform))
-    _binding_keys_by_object[id(source_obj)].add(binding_key)
-    _binding_keys_by_object[id(target_obj)].add(binding_key)
-    if binding_key not in bindable_properties:
-        active_links.append((source_obj, source_name, target_obj, target_name, transform))
-        _active_links_added.set()
-    _propagate(source_obj, source_name)
-
-
-def _pop_binding_keys_for_objects(object_ids: Iterable[int]) -> set[BindingKey]:
-    """Pop and return binding keys whose source or target may reference the given objects."""
-    binding_keys: set[BindingKey] = set()
-    for obj_id in object_ids:
-        indexed_binding_keys = _binding_keys_by_object.pop(obj_id, None)
-        if indexed_binding_keys is not None:
-            binding_keys.update(indexed_binding_keys)
-    return binding_keys
-
-
 def _get_attribute(obj: object | Mapping, name: tuple[str, ...]) -> Any:
     try:
         for key in name:
@@ -180,6 +150,19 @@ def _check_self_and_other_attribute(self_obj: Any, self_name: tuple[str, ...], o
         _check_attribute_exists(self_obj, self_name, role='self')
     if other_strict or (other_strict is None and not _path_contains_dict(other_obj, other_name)):
         _check_attribute_exists(other_obj, other_name, role='other')
+
+
+def _bind_one_way(source_obj: Any, source_name: tuple[str, ...], target_obj: Any, target_name: tuple[str, ...],
+                  transform: Callable[[Any], Any] | None) -> None:
+    """Register a one-way binding and run its initial propagation."""
+    binding_key = (id(source_obj), source_name)
+    bindings[binding_key].append((source_obj, target_obj, target_name, transform))
+    _binding_keys_by_object[id(source_obj)].add(binding_key)
+    _binding_keys_by_object[id(target_obj)].add(binding_key)
+    if binding_key not in bindable_properties:
+        active_links.append((source_obj, source_name, target_obj, target_name, transform))
+        _active_links_added.set()
+    _propagate(source_obj, source_name)
 
 
 def bind_to(self_obj: Any, self_name: str | tuple[str, ...], other_obj: Any, other_name: str | tuple[str, ...],
@@ -308,16 +291,22 @@ class BindableProperty:
             self._change_handler(owner, value)
 
 
+def _discard_binding_key_from_object_index(obj_id: int, binding_key: BindingKey) -> None:
+    keys = _binding_keys_by_object[obj_id]
+    keys.discard(binding_key)
+    if not keys:
+        del _binding_keys_by_object[obj_id]
+
+
 def remove(objects: Iterable[Any]) -> None:
     """Remove all bindings that involve the given objects.
 
     :param objects: The objects to remove.
     """
     object_ids = set(map(id, objects))
-    affected_binding_keys = _pop_binding_keys_for_objects(object_ids)
+    affected_binding_keys = {key for obj_id in object_ids for key in _binding_keys_by_object.pop(obj_id, ())}
     if not affected_binding_keys:
-        # remove() only deletes binding links; weak bindable-property markers expire when their objects are collected.
-        return
+        return  # only delete binding links; weak bindable-property markers expire when their objects are collected
 
     active_links[:] = [
         (source_obj, source_name, target_obj, target_name, transform)
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -18,28 +18,20 @@ from .logging import log
 if TYPE_CHECKING:
     from _typeshed import DataclassInstance, IdentityFunction
 
-ObjectId = int
-NamePath = tuple[str, ...]
-BindingKey = tuple[ObjectId, NamePath]
-Transform = Callable[[Any], Any] | None
-Binding = tuple[Any, Any, NamePath, Transform]
-ActiveLink = tuple[Any, NamePath, Any, NamePath, Transform]
-
-
 MAX_PROPAGATION_TIME = 0.01
 
-propagation_visited: ContextVar[set[BindingKey] | None] = \
-    ContextVar('propagation_visited', default=None)
+BindingKey = tuple[int, tuple[str, ...]]
+Binding = tuple[Any, Any, tuple[str, ...], Callable[[Any], Any] | None]
+ActiveLink = tuple[Any, tuple[str, ...], Any, tuple[str, ...], Callable[[Any], Any] | None]
 
-bindings: defaultdict[
-    BindingKey,
-    list[Binding]
-] = defaultdict(list)
+propagation_visited: ContextVar[set[BindingKey] | None] = ContextVar('propagation_visited', default=None)
+
+bindings: defaultdict[BindingKey, list[Binding]] = defaultdict(list)
 bindable_properties: weakref.WeakValueDictionary[BindingKey, Any] = weakref.WeakValueDictionary()
 active_links: list[ActiveLink] = []
 _active_links_added = asyncio.Event()
 # Maps object IDs to binding keys that reference them, so remove() avoids scanning all bindings.
-_binding_keys_by_object: dict[ObjectId, set[BindingKey]] = {}
+_binding_keys_by_object: dict[int, set[BindingKey]] = {}
 
 TC = TypeVar('TC', bound=type)
 T = TypeVar('T')
@@ -53,7 +45,7 @@ def _index_binding(source_obj: Any, target_obj: Any, binding_key: BindingKey) ->
     _binding_keys_by_object.setdefault(id(target_obj), set()).add(binding_key)
 
 
-def _discard_binding_key_from_object_index(obj_id: ObjectId, binding_key: BindingKey) -> None:
+def _discard_binding_key_from_object_index(obj_id: int, binding_key: BindingKey) -> None:
     indexed_binding_keys = _binding_keys_by_object.get(obj_id)
     if indexed_binding_keys is None:
         return
@@ -62,8 +54,8 @@ def _discard_binding_key_from_object_index(obj_id: ObjectId, binding_key: Bindin
         del _binding_keys_by_object[obj_id]
 
 
-def _bind_one_way(source_obj: Any, source_name: NamePath, target_obj: Any, target_name: NamePath,
-                  transform: Transform) -> None:
+def _bind_one_way(source_obj: Any, source_name: tuple[str, ...], target_obj: Any, target_name: tuple[str, ...],
+                  transform: Callable[[Any], Any] | None) -> None:
     """Register a one-way binding and run its initial propagation."""
     binding_key = (id(source_obj), source_name)
     bindings[binding_key].append((source_obj, target_obj, target_name, transform))
@@ -74,7 +66,7 @@ def _bind_one_way(source_obj: Any, source_name: NamePath, target_obj: Any, targe
     _propagate(source_obj, source_name)
 
 
-def _pop_binding_keys_for_objects(object_ids: Iterable[ObjectId]) -> set[BindingKey]:
+def _pop_binding_keys_for_objects(object_ids: Iterable[int]) -> set[BindingKey]:
     """Pop and return binding keys whose source or target may reference the given objects."""
     binding_keys: set[BindingKey] = set()
     for obj_id in object_ids:
@@ -84,7 +76,7 @@ def _pop_binding_keys_for_objects(object_ids: Iterable[ObjectId]) -> set[Binding
     return binding_keys
 
 
-def _remove_active_links_for_objects(removed_object_ids: set[ObjectId]) -> None:
+def _remove_active_links_for_objects(removed_object_ids: set[int]) -> None:
     """Drop polling fallback links that reference any removed object."""
     active_links[:] = [
         (source_obj, source_name, target_obj, target_name, transform)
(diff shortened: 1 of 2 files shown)


jwinpbe:
I see I missed a commit from @evnchn while I was working :sweat_smile:

falkoschindler (commented):
Thanks, @jwinpbe, for splitting this out, and for the detailed benchmark write-up! I appreciate the effort that went into it.

First the reassuring part: I ran a fairly thorough correctness pass over the new `remove()` and the index/registry bookkeeping (source-only, target-only, object-as-both-endpoints, self-loop, multi-target, chained, multi-name buckets, duplicate bindings, ghost removal, GC auto-discard, copy hook), plus the existing `test_binding.py` and `test_element_delete.py` suites. **I could not find a correctness bug** — the reverse index stays consistent with `bindings`, and the early-out can never skip a still-needed `active_links` cleanup because every bind indexes both endpoints. So the logic looks sound.

That said, I have three higher-level concerns I'd like to resolve before this goes in. They're more about how the change is shaped and presented than about correctness.

## 1. I can't actually review the reproduction

Right now the only reproduction is ~2000 lines of benchmark/analysis code spread across two gists, and I can't realistically survey that. What I'm missing is a small, self-contained NiceGUI example app — a handful of elements — that shows *which* form of UI usage leads to the long vs. the short runtime. Something I can paste, run, and see the problem (and the fix) with my own eyes. Could you distill the scenario down to a minimal example like that? Without it I'm being asked to trust numbers I can't independently reproduce or reason about.

## 2. Several measures are squeezed into one PR — which one actually wins?

The PR bundles a number of distinct changes (the reverse index, the custom registry replacing `WeakValueDictionary`, the single-vs-set bucket micro-optimizations, the `len==1` fast path, …), and the benchmark reports one aggregate number. So I can't see which change buys which part of the speedup.

My strong suspicion is that **one** change — the reverse index that avoids the full scan in `remove()` — accounts for ~90% of the improvement, while the rest adds real code duplication and complexity for only marginal gains (see the notes below). If that's the case, I'd much rather take the one high-impact change on its own and skip the complexity. Could you measure the reverse index in isolation first, and then only add each further optimization if it independently justifies its cost? That way each bit of added complexity has a number next to it.

## 3. How private is `binding.bindable_properties` really?

The PR narrows `binding.bindable_properties` from a plain `WeakValueDictionary` to a custom type that only supports a handful of operations (`__getitem__`/`__delitem__`/`.get`/`len`/`.values` are all gone). Strictly speaking these are undocumented internals with no stability guarantee, so I'm not claiming this is a formal breaking change. But I'd at least like us to question the "it's private, so anything goes" assumption before narrowing the surface, because in practice this module does get touched from the outside:

- #4109 and #2335 both reach into `bindable_properties` as a dict (`.items()`, `del`, `len(...)`) — once to manually clean up a leak, once to diagnose the very `remove()` cost this PR fixes.
- In #4711 we ourselves suggested importing `active_links` and calling `len()` on it to debug. (`active_links` stays a `list`, so that one keeps working.)

None of that makes it public API — but it's enough of a grey zone that I'd rather not shrink the surface silently in a point release. Two ways to stay safe: keep `bindable_properties` dict-compatible, or, if we do want a deliberate redesign of this layer, schedule it for 4.0 where a small breaking change is acceptable.

## Smaller cleanup notes

These line up with concern #2 — they're the complexity I'd like to see justified per-measure or dropped:

- **[nicegui/binding.py](nicegui/binding.py#L45) (45-57, 123-145, 166-182)** — the "store a scalar, upgrade to a set on the second insert, drop on empty" pattern is hand-rolled three separate times (registry name buckets, the `_add`/`_discard` index helpers, and the read in `_collect_binding_keys_for_objects`). Three copies to keep in sync, all to save one `set()` allocation per single-element object. A single small helper (or just always using a set) would remove most of the duplication.
- **[nicegui/binding.py](nicegui/binding.py#L32) (32-99)** — the custom `_BindablePropertyRegistry` re-implements machinery `WeakValueDictionary` gave for free (weakref liveness sweeps, per-object discard callbacks, iteration). The reverse index is what makes `remove()` fast; the registry rewrite is a separate concern. Keeping the `WeakValueDictionary` and adding only the reverse index might get most of the win with far less new code — which is exactly what concern #2 is about.
- **[nicegui/binding.py](nicegui/binding.py#L48) (48-51)** — a fresh `discard` closure is allocated per registered object, retained by the weakref for the object's lifetime, whereas `WeakValueDictionary` uses one shared callback. Minor, but it's in the object-creation hot path and likely explains part of the +3% regression on the bound-target case.
- **[nicegui/binding.py](nicegui/binding.py#L71) (71-77)** — `__contains__` does four isinstance/length checks before delegating to a separate `contains_key`, but the only caller always passes a valid `(int, tuple)` key. Could likely collapse the two and drop the guards (and the `cast` import).
- **[nicegui/binding.py](nicegui/binding.py#L459) (459-479)** — `remove()`'s index-cleanup contract is split across distant code: surviving targets are de-indexed individually, removed ids are cleared by a separate wholesale `pop` loop, and the invariant tying them together lives only in a comment. Correct today, but easy to break later. The `len==1` fast path (466-470) also duplicates the general comprehension just below it.

None of the cleanup notes block on their own — but together they're the "complexity vs. measured benefit" trade-off I'd like us to make deliberately rather than by default.

jwinpbe:
Good morning! Thanks for the review.

I replaced the earlier benchmark harness with a paste-and-run NiceGUI app for the full PR candidate. It uses normal NiceGUI elements (`input`, `select`, `slider`, `switch`, and `labels`) and shows the specific shape that was slow: refreshing a small bound detail card while many other bound rows are still on the page.

I tried to break it up visually so you can run it and have it look OK and be able to reason through it, if you want me to just use script mode elements in the future please let me know what would be easiest for you to reason through. Thanks for your patience.

<details>
<summary>binding_bench_app.py</summary>

```python
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.10"
# dependencies = ["nicegui"]
# [tool.uv.sources]
# nicegui = { path = "." }
# ///
import gc
import pathlib
import shelve
import tempfile
import time

#############
# CONSTANTS #
#############

DB_PATH = pathlib.Path(tempfile.gettempdir()) / "bench_results"
SEP = "─────"

DEFAULT_LOADED_ORDERS = 750
SAMPLES = 3
ENTRIES_PER_LOADED_ROW = 5       # switch, input, select, slider, price label
CARD_BINDING_ENTRIES = 7         # customer, qty, status, rush, discount, progress, total
PRODUCTS = ["Sensor", "Controller", "Display", "Gateway"]
STATUSES = ["draft", "packed", "shipped"]

ui.input.default_props("dense outlined")
ui.select.default_props("dense outlined")
ui.number.default_props("dense outlined")
ui.card.default_classes("w-full gap-3")


##############
# DATA MODEL #
##############


@binding.bindable_dataclass
class Order:
    customer: str
    product: str
    quantity: int
    price: float
    discount: int
    status: str
    rush: bool


def make_order(index: int) -> Order:
    return Order(
        f"Customer {index}",
        PRODUCTS[index % len(PRODUCTS)],
        index % 8 + 1,
        12.5 + index % 30,
        index % 25,
        STATUSES[index % len(STATUSES)],
        index % 7 == 0,
    )


###########
# HELPERS #
###########


def loaded_order_row(order: Order) -> None:
    """Render one order — registers ENTRIES_PER_LOADED_ROW binding entries with binding.remove()."""
    with ui.row().classes("items-center gap-2 no-wrap"):
        ui.switch(value=order.rush).bind_value(order, "rush")
        ui.input(value=order.customer).bind_value(order, "customer").classes("w-40")
        ui.select(PRODUCTS, value=order.product).bind_value(order, "product").classes("w-36")
        ui.slider(min=0, max=50, value=order.discount).bind_value(order, "discount").classes("w-32")
        ui.label().bind_text_from(order, "price", backward=lambda price: f"${price:.2f}").classes("w-16 text-right")


########
# PAGE #
########


@ui.page("/")
def main() -> None:
    ui.query("body").classes("bg-slate-50")
    selected = make_order(0)

    # ── Detail card ───────────────────────────────────────────────────────────

    @ui.refreshable
    def selected_order_card() -> None:
        """Refreshable card — teardown triggers binding.remove() on CARD_BINDING_ENTRIES elements."""
        with ui.card().classes("border border-blue-200 bg-blue-50"):
            ui.label("Selected order").classes("font-semibold")
            with ui.row().classes("items-center gap-3"):
                ui.input("Customer", value=selected.customer).bind_value(selected, "customer").classes("w-52")
                ui.number("Qty", value=selected.quantity, min=1).bind_value(selected, "quantity").classes("w-24")
                ui.select(STATUSES, label="Status", value=selected.status).bind_value(selected, "status").classes(
                    "w-36"
                )
                ui.switch("Rush").bind_value(selected, "rush")
            ui.slider(min=0, max=50, value=selected.discount).bind_value(selected, "discount")
            ui.linear_progress().bind_value_from(selected, "discount", backward=lambda value: value / 50)
            ui.label().bind_text_from(selected, "quantity", backward=lambda qty: f"Total: ${selected.price * qty:.2f}")

    # ── Benchmark logic ────────────────────────────────────────────────────────

    def populate_orders(count: int) -> None:
        """Clear and fill the orders container with `count` loaded-order rows."""
        orders.clear()
        with orders:
            for index in range(count):
                loaded_order_row(make_order(index))

    async def run_benchmark() -> None:
        run_button.props("loading disabled")
        try:
            count = int(row_count.value or 0)

            # Phase 1 — create loaded-order bindings for binding.remove() to scan
            populate_orders(count)
            total_binding_entries = count * ENTRIES_PER_LOADED_ROW + CARD_BINDING_ENTRIES
            summary.set_text(
                f"{count:,} loaded orders ({total_binding_entries} binding entries on page). "
                f"Card refresh will call binding.remove() on {CARD_BINDING_ENTRIES} controls."
            )

            # Phase 2 — measure binding.remove() cost during card refresh
            # refresh() deletes the card's current elements (triggering remove() on
            # CARD_BINDING_ENTRIES bound controls) then recreates them.  On main
            # remove() scans every binding entry; on the PR branch the reverse index
            # only visits the card's own entries.
            gc.collect()
            samples = []
            for _ in range(SAMPLES):
                start = time.perf_counter()
                await selected_order_card.refresh()
                samples.append((time.perf_counter() - start) * 1000)

            # Phase 3 — record the result alongside the binding count
            table.rows = table.rows + [{
                "loaded_orders": f"{count:,}",
                "binding_entries": str(total_binding_entries),
                "samples": str(SAMPLES),
                "best_ms": f"{min(samples):.4f}",
                "mean_ms": f"{sum(samples) / SAMPLES:.4f}",
            }]
            with shelve.open(DB_PATH) as db:
                db["results"] = [dict(row) for row in table.rows if row.get("best_ms") != SEP]
            result.set_text("Run the same scenario on upstream/main and on the PR branch, then compare the mean ms.")
        finally:
            run_button.props(remove="loading disabled")

    def clear_results() -> None:
        table.rows = []
        with shelve.open(DB_PATH) as db:
            db["results"] = []

    # ── Layout / UI ─────────────────────────────────────────────────────────────────

    with ui.column().classes("mx-auto max-w-4xl p-6 gap-4"):
        ui.label("Order dashboard cleanup reproduction").classes("text-xl font-semibold")
        ui.markdown(
            "Each loaded-order row creates **5 binding entries** (switch, input, select, slider, price). "
            "The detail card has **7**.\n"
            "Refreshing the card triggers `binding.remove()` on those 7 controls. "
            "On `main` this scans *all* entries; on the PR branch the reverse index finds only the 7."
        )
        with ui.card():
            summary = ui.label().classes("text-sm text-slate-600")
            with ui.row().classes("items-end gap-3"):
                row_count = ui.number("Loaded orders", value=DEFAULT_LOADED_ORDERS, min=0, max=3000, step=100).classes(
                    "w-40"
                )
                run_button = ui.button("Benchmark", icon="speed", on_click=run_benchmark)
                ui.button("Clear results", icon="delete", color="red", on_click=clear_results)
            result = ui.label("No timing samples yet.").classes("text-sm text-slate-600")

        table = (
            ui.table(
                columns=[
                    {"name": "loaded_orders", "label": "Loaded orders", "field": "loaded_orders"},
                    {"name": "binding_entries", "label": "Binding entries", "field": "binding_entries"},
                    {"name": "samples", "label": "Samples", "field": "samples"},
                    {"name": "best_ms", "label": "Best ms", "field": "best_ms"},
                    {"name": "mean_ms", "label": "Mean ms", "field": "mean_ms"},
                ],
                rows=[],
                column_defaults={"align": "center"},
            )
            .classes("w-full")
            .props("dense flat bordered hide-bottom")
        )

        with shelve.open(DB_PATH) as db:
            saved = db.get("results")
            if saved:
                table.rows = [
                    *[dict(r) for r in saved],
                    dict.fromkeys(("loaded_orders", "binding_entries", "samples", "best_ms", "mean_ms"), SEP),
                ]
        selected_order_card()
        with ui.expansion("Loaded order editor rows (kept collapsed)", icon="inventory_2").classes("w-full"):
            orders = ui.column().classes("gap-1")

    populate_orders(DEFAULT_LOADED_ORDERS)


try:
    ui.run(title="binding.remove() reproduction", reload=False, binding_refresh_interval=None)
except KeyboardInterrupt:
    raise SystemExit(0)
```
</details>

I ran the same 750-loaded-order scenario four times on the PR branch and four times on `upstream/main`:

| Branch | Run | Loaded orders | Binding entries | Samples | Best ms | Mean ms |
|---|---:|---:|---:|---:|---:|---:|
| this PR | 1 | 750 | 3,757 | 3 | 1.5501 | 1.6742 |
| this PR | 2 | 750 | 3,757 | 3 | 1.5665 | 1.6974 |
| this PR | 3 | 750 | 3,757 | 3 | 1.5627 | 1.7040 |
| this PR | 4 | 750 | 3,757 | 3 | 1.5716 | 1.7095 |
| upstream/main | 1 | 750 | 3,757 | 3 | 5.1876 | 5.7336 |
| upstream/main | 2 | 750 | 3,757 | 3 | 5.5227 | 6.2918 |
| upstream/main | 3 | 750 | 3,757 | 3 | 5.6949 | 6.4725 |
| upstream/main | 4 | 750 | 3,757 | 3 | 5.5679 | 6.3049 |

Across those runs, the PR averaged **1.6963 ms** and `upstream/main` averaged **6.2007 ms**, about **3.7x faster** by mean time for this workload.

## What actually drives the speedup

I also built a three-way microbenchmark to separate the two main changes:

| Variant | Meaning |
|---|---|
| A | old full scan |
| B | reverse index only |
| C | current HEAD / full PR |

For objects with **no bindings at all** (ghost removals, like clearing a page), the reverse index does almost all the work. It checks the index, finds nothing, and returns without scanning all bindings or walking the bindable-property registry.

| State size | A: Full scan | B: Rev-index only | C: Full PR |
|---:|---:|---:|---:|
| 1,000 | 1.72 ms | 0.003 ms | 0.003 ms |

B and C are basically tied here, roughly **500-700x faster** than A.

For **real removals** (objects that actually have bindings), the reverse index alone helps less because registry cleanup can still dominate:

| Config | A: Full scan | B: Rev-index only | C: Full PR |
|---|---:|---:|---:|
| 500 elements, remove 1 | 0.71 ms | 0.61 ms | 0.008 ms |
| 1k elements, remove 1 | 1.62 ms | 1.39 ms | 0.009 ms |

B is only around **1.2x faster** than A in those cases. It finds the affected binding keys, but still has to iterate the bindable-property registry.

C is around **100-200x faster** because `discard_object_ids` turns that cleanup into direct dict pops. `_BindablePropertyRegistry` stores one entry per object ID, so cleaning up one removed object is a single `pop()` instead of a scan.

The two pieces complement each other nicely: the reverse index handles ghost removals, and the registry rewrite handles real removals.

<details>
<summary>reverse_index_bench.py</summary>

```python
#!/usr/bin/env -S uv run --script
#
# /// script
# requires-python = ">=3.10"
# dependencies = ["nicegui", "rich"]
# [tool.uv.sources]
# nicegui = { path = "." }
# ///
"""Three-way remove() benchmark: full-scan vs reverse-index-only vs full PR.

A = old_remove              — pre-PR algorithm: scans every bindings entry.
B = reverse_index_only_remove — uses _collect_binding_keys_for_objects to
                                avoid the full scan, but strips every micro-
                                optimisation (no len==1 fast path, no survivor
                                de-indexing, old-style registry iteration).
C = binding.remove          — the PR as committed: reverse index + all micro-opts.
"""

import gc
import time
import statistics

from rich import box
from rich.console import Console
from rich.panel import Panel
from rich.table import Table

console = Console()


def _fmt_ratio(ratio: float) -> str:
    """Format a speedup ratio with green/red/dim rich markup."""
    if ratio > 1.05:
        return f'[green]{ratio:.1f}x[/]'
    if ratio < 0.95:
        return f'[red]{ratio:.1f}x[/]'
    return f'[dim]{ratio:.1f}x[/]'


# ── 0. Script-mode setup ────────────────────────────────────────────────────

from nicegui import binding, core, ui
from nicegui.page import page

core.script_mode = True
if not Client.instances:
    core.script_client = Client(page("/")).__enter__()

# ── 1. Old remove() for comparison ──────────────────────────────────────────


def old_remove(objects, *, _binding=binding):
    """Full-scan remove() — the algorithm from before the PR."""
    object_ids = {id(obj) for obj in objects}
    # active_links filter
    _binding.active_links[:] = [
        link
        for link in _binding.active_links
        if id(link[0]) not in object_ids and id(link[2]) not in object_ids
    ]
    # full scan of all bindings
    for key, binding_list in list(_binding.bindings.items()):
        binding_list[:] = [
            entry
            for entry in binding_list
            if id(entry[0]) not in object_ids and id(entry[1]) not in object_ids
        ]
        if not binding_list:
            del _binding.bindings[key]
    # old-style bindable_properties cleanup via iteration
    for obj_id, name in list(_binding.bindable_properties):
        if obj_id in object_ids:
            del _binding.bindable_properties[(obj_id, name)]


def reset_state():
    """Reset all binding state between runs."""
    binding.reset()


# ── 2. Reverse-index only (no micro-opts) ───────────────────────────────────


def reverse_index_only_remove(objects, *, _binding=binding):
    """remove() using the reverse index — stripped of every micro-optimisation.

    The only change from the old full scan is using
    ``_collect_binding_keys_for_objects`` to find the few binding lists
    that actually reference a removed object.  Everything else is the
    simplest possible implementation — no ``len==1`` fast path, no survivor
    de-indexing during the loop, old-style registry iteration.
    """
    removed_object_ids = [id(obj) for obj in objects]
    if not removed_object_ids:
        return

    # Reverse index tells us which binding keys might be affected.
    # Check this FIRST so ghost removals (no affected keys) return
    # immediately without scanning bindable_properties.
    affected_binding_keys = _binding._collect_binding_keys_for_objects(removed_object_ids)  # noqa: SLF001
    if not affected_binding_keys:
        return

    removed_object_id_set = set(removed_object_ids)

    # Registry cleanup: old iteration approach (no discard_object_ids).
    # Uses a set for O(1) membership checks.
    for obj_id, name in list(_binding.bindable_properties):
        if obj_id in removed_object_id_set:
            del _binding.bindable_properties[(obj_id, name)]



    # Filter active_links (same as old code)
    _binding.active_links[:] = [
        link for link in _binding.active_links
        if id(link[0]) not in removed_object_id_set and id(link[2]) not in removed_object_id_set
    ]

    # Process affected keys — always a comprehension, no fast paths
    for key in affected_binding_keys:
        binding_list = _binding.bindings.get(key)
        if binding_list is None:
            continue
        if key[0] in removed_object_id_set:
            # Source was removed — delete the entire binding list.
            # Survivor targets keep stale reverse-index entries;
            # the next lookup for them will get None from bindings.get()
            # and be skipped.  The trailing bulk pop handles removed IDs.
            del _binding.bindings[key]
            continue
        remaining = [entry for entry in binding_list if id(entry[1]) not in removed_object_id_set]
        if remaining:
            binding_list[:] = remaining
        else:
            del _binding.bindings[key]

    # Bulk cleanup: wipe all traces of removed objects from the reverse index
    pop = _binding._binding_keys_by_object.pop  # noqa: SLF001
    for obj_id in removed_object_ids:
        pop(obj_id, None)


# ── 3. Benchmark harness ────────────────────────────────────────────────────


def create_elements_with_bindings(num_elements):
    """Create that many element-pairs, each with a 1-way binding. Returns everything so refs stay alive."""
    elements = []
    for i in range(num_elements):
        label = ui.label(f"item_{i}")
        inp = ui.input(value=f"val_{i}")
        inp.bind_value_to(label, "text")
        elements += [label, inp]
    return elements


class _Ghost:
    """Placeholder with no bindings used to test the ghost-removal early-out."""
    pass


def bench(remove_fn, num_total, *, num_ghosts=None, num_to_remove=None, num_samples=7):
    """Time remove() across num_samples runs, resetting state between each.

    Two modes:
      num_ghosts=10    → objects with NO bindings (tests early-out)
      num_to_remove=5  → slice from created elements (tests real removal)

    GC is disabled during each timed sample and restored afterward.
    """
    times = []
    was_enabled = gc.isenabled()
    for _ in range(num_samples):
        elements = create_elements_with_bindings(num_total)
        if num_ghosts is not None:
            targets = [_Ghost() for _ in range(num_ghosts)]
        else:
            targets = elements[:num_to_remove * 2]
        gc.disable()
        t0 = time.perf_counter()
        remove_fn(targets)
        times.append(time.perf_counter() - t0)
        gc.enable()
        reset_state()
    if was_enabled:
        gc.enable()
    return times



def _row(label, times_a, times_b, times_c):
    """Compute a row of data from raw timing lists."""
    a_mean = statistics.mean(times_a) * 1000
    b_mean = statistics.mean(times_b) * 1000
    c_mean = statistics.mean(times_c) * 1000
    ab_ratio = a_mean / b_mean if b_mean > 0 else float("inf")
    ac_ratio = a_mean / c_mean if c_mean > 0 else float("inf")
    return label, a_mean, b_mean, c_mean, ab_ratio, ac_ratio


def _build_table(title, rows, *, label_header="Objects"):
    """Build a rich Table from collected row tuples."""
    table = Table(title=title, box=box.ROUNDED, header_style="bold cyan")
    table.add_column(label_header, style="cyan")
    table.add_column("A: Full-scan (ms)", justify="right")
    table.add_column("B: Rev-idx (ms)", justify="right")
    table.add_column("C: Full PR (ms)", justify="right")
    table.add_column("B vs main", justify="right")
    table.add_column("C vs main", justify="right")
    for label, a_ms, b_ms, c_ms, ab_ratio, ac_ratio in rows:
        table.add_row(
            str(label),
            f"{a_ms:.4f}",
            f"{b_ms:.4f}",
            f"{c_ms:.4f}",
            _fmt_ratio(ab_ratio),
            _fmt_ratio(ac_ratio),
        )
    return table


def main():
    # ── Print intro immediately ──
    console.print()
    console.print(
        Panel.fit(
            "[bold]Three-way remove() benchmark[/]\n"
            "[bold]A[/] = old full scan  — pre-PR algorithm\n"
            "[bold]B[/] = reverse index only — stripped of micro-optimisations\n"
            "[bold]C[/] = full PR — reverse index + all micro-opts",
            border_style="dim",
        )
    )
    console.print()

    # ── Ghost removals: collect data, then print table ──
    ghost_counts = [10, 50, 200, 1000]
    ghost_rows = []
    for count in ghost_counts:
        times_a = bench(old_remove, count, num_ghosts=10)
        times_b = bench(reverse_index_only_remove, count, num_ghosts=10)
        times_c = bench(binding.remove, count, num_ghosts=10)
        ghost_rows.append(_row(count, times_a, times_b, times_c))
    console.print(_build_table("Ghost removals", ghost_rows))
    console.print()

    # ── Real removals: collect data, then print table ──
    real_configs = [
        ("100 total, remove 1",  100, 1),
        ("100 total, remove 5",  100, 5),
        ("500 total, remove 1",  500, 1),
        ("500 total, remove 10", 500, 10),
        ("1k total, remove 1",   1000, 1),
        ("1k total, remove 10",  1000, 10),
    ]
    real_rows = []
    for label, total, removed in real_configs:
        times_a = bench(old_remove, total, num_to_remove=removed)
        times_b = bench(reverse_index_only_remove, total, num_to_remove=removed)
        times_c = bench(binding.remove, total, num_to_remove=removed)
        real_rows.append(_row(label, times_a, times_b, times_c))
    console.print(_build_table("Real removals", real_rows, label_header="Config"))
    console.print()

    # ── Methodology ──
    console.print(
        Panel.fit(
            "[bold]Methodology[/]\n"
            "• Mean wall time over 7 samples per cell, with state reset between runs.\n"
            "• GC is disabled during each timed sample and restored afterward.\n"
            "• Numbers are workload-shaped microbenchmarks, not end-user latency claims.\n"
            "• [green]Green[/] ratios = speedup, [red]red[/] = regression, [dim]dim[/] = noise.",
            border_style="dim",
        )
    )
    console.print()


if __name__ == "__main__":
    main()
```
</details>

## Bindable properties

The linked issues show `bindable_properties` has been useful for debugging and workaround code, so I kept the dict-like surface. Thanks for pointing that out.

The registry now mixes in `MutableMapping`, so `len()`, item access/deletion, `.get()`, `.pop()`, `.keys()`, `.items()`, and `.values()` continue to work without re-implementing every method by hand. I added a regression test for that behavior too.

Rolling the methods manually instead of using the mixin is also fine, but the mixin keeps the implementation smaller.

## Smaller isolated measures

I tested the bucket representation a few ways. The most compact implementation is to use plain sets everywhere instead of storing a scalar first and upgrading to a set on the second insert.

That version is what current HEAD uses now:

| Area | Result |
|---|---|
| Code size | 54 fewer lines in `binding.py` |
| Isolated binding cleanup | about 4-10% slower than scalar-or-set buckets |
| Full lifecycle benchmark | neutral to slightly favorable |
| Readability | much simpler; no scalar/set state machine |

I originally tried shared scalar-or-set helpers (`_bucket_add`, `_bucket_discard`, etc.). That kept the optimization, but only saved about 13 lines and added a small helper-call overhead. The always-set version was the better tradeoff: much easier to read, still within +10% of the optimized version, and no meaningful regression in the full lifecycle benchmark.

For the smaller pieces, I kept or changed them like this:

| Piece | Result | Decision |
|---|---:|---|
| `len(binding_list) == 1` fast path | 28.4% faster in isolation | Keep |
| Scalar-or-set buckets | Faster in isolation, but much more code | Replaced with plain sets |
| Per-object closure vs `RegistryRef` | Closure measured faster in isolation | Keep |
| `__contains__` guard cleanup | Simpler and measured slightly faster | Keep simplified |

| Note | What changed |
|---|---|
| `len(binding_list) == 1` fast path | Kept; it avoids building a comprehension for the common single-binding case. |
| Per-object weakref cleanup | Kept; the custom registry still needs object-id-aware cleanup when the object is collected. |
| `__contains__` | Simplified from the earlier guard-heavy version and removed the unused `cast` import. |
| `remove()` cleanup contract | Renamed locals around the reverse index so the source/target cleanup paths are easier to follow. |

<details>
<summary>microbenchmark.py</summary>

```python
#!/usr/bin/env -S uv run --script
#
# /// script
# requires-python = ">=3.10"
# dependencies = ["rich"]
# ///
"""Microbenchmarks for NiceGUI binding micro-optimizations.
Run: python3 microbenchmark.py"""

import gc
import statistics
import time
import weakref

from rich import print


def _run(label, setup, bench_fn, num_runs=50):
    gc.collect()
    gc.disable()
    for _ in range(5):
        bench_fn(setup())
    times = []
    for _ in range(num_runs):
        state = setup()
        t0 = time.perf_counter()
        bench_fn(state)
        t1 = time.perf_counter()
        gc.collect()
        times.append(t1 - t0)
    gc.enable()
    median_us = statistics.median(times) * 1e6
    print(f"  {label:46s}  median={median_us:>8.2f}us  (n={num_runs})")
    return times


# --- A: len(binding_list)==1 fast path (binding.py:466-470) ---
def bench_A():
    print("\n=== A: len==1 fast path ===")
    N = 1000

    def setup():
        objs = [object() for _ in range(N)]
        return ({i: [objs[i]] for i in range(N)}, {id(o) for o in objs[: N // 2]}, list(range(N)))

    def run(state, use_fast):
        data, to_remove, keys = state
        for k in keys:
            binding_list = data.get(k)
            if binding_list is None:
                continue
            if use_fast and len(binding_list) == 1:
                if id(binding_list[0]) in to_remove:
                    del data[k]
                continue
            remaining = [b for b in binding_list if id(b) not in to_remove]
            if remaining:
                binding_list[:] = remaining
            else:
                del data[k]

    t1 = _run("with fast path (len==1 early return)", setup, lambda s: run(s, True))
    t2 = _run("without fast path (always comprehension)", setup, lambda s: run(s, False))
    median_fast, median_slow = statistics.median(t1), statistics.median(t2)
    print(f"  >> speedup: {(median_slow - median_fast) / median_slow * 100:+.1f}%")


# --- B: Scalar-or-set bucket pattern (4 functions) ---
def bench_B():
    print("\n=== B: Scalar-or-set bucket pattern ===")
    N = 10000

    def setup():
        return [(i, ("prop",)) for i in range(N)]

    def run(items, use_set):
        idx = {}
        for oid, prop in items:
            bucket = idx.get(oid)
            if bucket is None:
                idx[oid] = {prop} if use_set else prop
            elif use_set or isinstance(bucket, set):
                bucket.add(prop)
            elif bucket != prop:
                idx[oid] = {bucket, prop}
        for oid, _ in items:
            bucket = idx.get(oid)
            if bucket is None:
                continue
            if use_set:
                bucket.discard(("prop",))
                if not bucket:
                    del idx[oid]
            elif not isinstance(bucket, set):
                if bucket == ("prop",):
                    del idx[oid]
            else:
                bucket.discard(("prop",))
                if not bucket:
                    del idx[oid]

    items = list(setup())
    t1 = _run("scalar-or-set (scalar for singles)", lambda: list(items), lambda s: run(s, False))
    t2 = _run("always-set (set for singles)", lambda: list(items), lambda s: run(s, True))
    median_scalar, median_always_set = statistics.median(t1), statistics.median(t2)
    print(f"  >> scalar faster by: {(median_always_set - median_scalar) / median_always_set * 100:+.1f}%")


# --- C: Per-object closure (line 49) vs shared RegistryRef ---
def bench_C():
    print("\n=== C: Per-object closure vs RegistryRef ===")
    N = 10000

    class RegistryRef(weakref.ref):
        __slots__ = ("obj_id",)

        def __new__(cls, obj, oid, cb):
            inst = super().__new__(cls, obj, cb)
            inst.obj_id = oid
            return inst

        def __init__(self, *a, **kw):
            pass

    def shared_cb(ref):
        pass

    class Obj:
        __slots__ = ("__weakref__",)

    dummy = [Obj() for _ in range(N)]

    def setup():
        return [(dummy[i], i) for i in range(N)]

    def run(items, use_registry_ref):
        registry = {}
        for obj, oid in items:
            if use_registry_ref:
                ref = RegistryRef(obj, oid, shared_cb)
            else:

                def discard(_r, _id=oid):
                    pass

                ref = weakref.ref(obj, discard)
            registry[oid] = (ref, ("p1",))
            registry[oid] = (ref, {("p1",), ("p2",)})
            registry[oid] = (ref, {("p1",), ("p2",), ("p3",)})

    t1 = _run("per-object closure (10K x 3 props)", setup, lambda s: run(s, False))
    t2 = _run("shared RegistryRef (10K x 3 props)", setup, lambda s: run(s, True))
    median_closure, median_registry_ref = statistics.median(t1), statistics.median(t2)
    print(f"  >> difference: {(median_registry_ref - median_closure) / median_closure * 100:+.1f}%")


# --- D: __contains__ isinstance guards (binding.py:71-77) ---
def bench_D():
    print("\n=== D: __contains__ isinstance guards ===")
    NUM_ENTRIES, NUM_LOOKUPS = 10000, 50000

    def setup():
        entries = {(i, ("prop",)): object() for i in range(NUM_ENTRIES)}
        valid = list(entries.keys())[: NUM_LOOKUPS // 2]
        junk = [None, 42, (1,), (1, 2, 3), ("s", ("p",)), (1, "nt")] * (NUM_LOOKUPS // 12)
        return entries, valid + junk[: NUM_LOOKUPS // 2]

    def run(state, variant):
        entries, keys = state
        for k in keys:
            if variant == "current":
                if not isinstance(k, tuple) or len(k) != 2:
                    continue
                oid, prop = k
                if not isinstance(oid, int) or not isinstance(prop, tuple):
                    continue
            elif variant == "simplified":
                if not (isinstance(k, tuple) and len(k) == 2):
                    continue
            # variant "direct": no checks
            _ = k in entries

    t1 = _run("current 4-check __contains__", setup, lambda s: run(s, "current"))
    t2 = _run("simplified 2-check __contains__", setup, lambda s: run(s, "simplified"))
    t3 = _run("direct pass-through (0 checks)", setup, lambda s: run(s, "direct"))
    median_current, median_simplified, median_direct = (
        statistics.median(t1),
        statistics.median(t2),
        statistics.median(t3),
    )
    print(f"  >> simplified vs current: {(median_simplified - median_current) / median_current * 100:+.1f}%")
    print(f"  >> direct vs current:    {(median_direct - median_current) / median_current * 100:+.1f}%")


if __name__ == "__main__":
    print("NiceGUI binding microbenchmarks")
    print("=" * 55)
    bench_A()
    bench_B()
    bench_C()
    bench_D()
    print()
```
</details>

## Cleanup

- Replaced the scalar-or-set bucket logic with plain sets.
- Kept `_BindablePropertyRegistry`, because `WeakValueDictionary` cannot discard all bindable properties for an object ID without scanning all keys.
- Kept the per-object weakref cleanup path.
- Kept the reverse-index check before `discard_object_ids`, so ghost removals skip registry cleanup entirely.
- Preserved the public dict-like behavior of `bindable_properties`.
- Added regression coverage for that compatibility surface.
- Verified the binding and element-delete tests pass.

Let me know what you think. Thanks!

-- jwin

evnchn:
*Drafted with Claude Code; posting on my own review.*

**TL;DR: I think Falko's right — simplify it. The empiricals mostly point that way.**

I tried the variants, and the reverse index is where basically all the win is. Keeping `bindable_properties` as the plain `WeakValueDictionary` (untouched) and adding *only* the object→binding-keys reverse index gets the full speedup at **+41/−13 in `binding.py`** — all three benchmark cases improve, and the concern about narrowing `bindable_properties` just disappears because the mapping isn't touched at all.

The "mostly": there's one small behavior change (dropping the eager `bindable_properties` purge from `remove()`), but it turns out to make remove→rebind behave like a fresh bind — arguably more consistent, not less. Details below.

@jwinpbe it's your PR — happy to hand this over as a starting point if it helps. @falkoschindler does this match what you meant by taking the reverse index on its own?

<details>
<summary><b>Why I landed there — the variants, measured</b></summary>

Same machine, baseline `c40f3e3f`, `maintainer_benchmark.py --group binding --min-seconds 6`, each side a median over ~40+ samples:

| approach | `binding.py` diff | unlinked | ghost | bound-targets | `bindable_properties` surface |
|---|---|---|---|---|---|
| this PR (as-is) | +211 / −36 | −97.7% | −4.7% | **+4.8%** | narrowed, then re-widened via `MutableMapping` |
| subclass `WeakValueDictionary` + index | +95 / −14 | −96.3% | **+50.3%** | **+11.0%** | full (it *is* a WVD) |
| custom registry class | +87 / −15 | −97.7% | −6.4% | −15.7% | narrowed |
| **reverse index only (WVD untouched)** | **+41 / −13** | −97.6% | −3.0% | −11.7% | **unchanged** |

The reverse index alone carries the improvement; each thing wrapped around it either costs perf or narrows the surface. The subclass idea (to keep dict-compat "for free") is actually the slowest — every bindable `__set__` then pays for two weakrefs, which shows up as +50% on the 15k-object ghost setup. So: not touching the WVD is both smallest and fastest.

Gates on the reverse-index-only variant: `pytest tests/test_binding.py tests/test_element_delete.py` → 30 passed; `mypy` clean; `pylint` 10.00/10.
</details>

<details>
<summary><b>The whole change, and the one behavioral note</b></summary>

`bindable_properties` stays as-is. Add one reverse index + two helpers, one line each in `bind_to`/`bind_from`, and let `remove()` consult the index instead of scanning global state:

```python
_binding_keys_by_object: dict[int, set[tuple[int, tuple[str, ...]]]] = {}

def _index_binding(source_id, target_id, binding_key):
    _binding_keys_by_object.setdefault(source_id, set()).add(binding_key)
    _binding_keys_by_object.setdefault(target_id, set()).add(binding_key)

# in bind_to / bind_from, right after bindings[...].append(...):
_index_binding(id(self_obj), id(other_obj), (id(self_obj), self_name_tuple))

def remove(objects):
    removed_ids = {id(obj) for obj in objects}
    if not removed_ids:
        return
    affected_keys = set()
    for obj_id in removed_ids:
        affected_keys.update(_binding_keys_by_object.pop(obj_id, ()))
    if not affected_keys:
        return  # no bindings; weak bindable_properties entries clear on GC
    for key in affected_keys:
        binding_list = bindings.get(key)
        if binding_list is None:
            continue
        if key[0] in removed_ids:                     # source removed -> whole list goes
            for _s, target_obj, _n, _t in binding_list:
                _discard_key_from_index(id(target_obj), key)
            del bindings[key]
        else:                                          # source survives -> prune removed targets
            binding_list[:] = [b for b in binding_list if id(b[1]) not in removed_ids]
            if not binding_list:
                del bindings[key]
                _discard_key_from_index(key[0], key)
    active_links[:] = [l for l in active_links
                       if id(l[0]) not in removed_ids and id(l[2]) not in removed_ids]
```

The eager `bindable_properties` cleanup loop is *deleted*, not replaced — the `WeakValueDictionary` self-cleans on GC. `test_automatic_cleanup` still passes: `remove([label])` drops the model's last strong refs, it's collected, the weak entry vanishes.

**The one behavioral difference** (confirmed by running it): a removed-but-still-alive object keeps its `bindable_properties` marker until GC. That only shows up in one contrived path — `remove()` + rebind + **in-place** mutation of a mutable value, relying on the polling fallback:

| scenario | `active_links` | `target` after in-place mutation |
|---|---|---|
| never-removed `BindableProperty` | 0 | `[1]` — not polled |
| remove+rebind — **baseline** | 1 | `[1, 2]` — polled |
| remove+rebind — **reverse-index** | 0 | `[1]` — not polled |

A `BindableProperty` binding never polls in-place mutations (row 1); baseline's eager delete accidentally downgraded the rebind to a polled link (row 2). The reverse-index version makes remove+rebind match a fresh bind (row 1 == row 3). If we'd rather keep baseline's exact eager semantics, that needs a per-object index on `bindable_properties` — i.e. the +87 or +95 variants above, which reintroduce the surface/perf cost. That's the only real trade to decide.

<details>
<summary>runnable repro</summary>

```python
from nicegui import binding

class Model:
    value = binding.BindableProperty()
    def __init__(self, value): self.value = value
class Plain:
    value = None

source, target = Model([1]), Plain()
binding.bind_to(source, 'value', target, 'value', lambda v: list(v))  # copy => distinct list
binding.remove([source])                                              # source stays alive
binding.bind_to(source, 'value', target, 'value', lambda v: list(v)) # rebind
source.value.append(2)                                               # in-place, no __set__
binding._refresh_step()
print(len(binding.active_links), target.value)
```
</details>
</details>

jwinpbe:
I simplified the branch toward the reverse-index-only idea @evnchn had.

It now keeps `bindable_properties` as the stdlib `WeakValueDictionary` and removes the BindablePropertyRegistry. The remaining runtime change is the object-id -> binding-key reverse index so `remove()` only visits binding lists that can reference the removed objects.

I also dropped the `len(binding_list) == 1` fast path. After measuring it separately, it didn't seem worth the extra pathway. The current `remove()` now uses one general binding-list pruning path.

I kept `_bind_one_way`, because it removes duplicated registration logic between `bind_to` and `bind_from` and ensures both paths update `bindings`, `_binding_keys_by_object`, `active_links`, and initial propagation consistently.

My benchmark shows the unlinked cleanup case still gets the main win from avoiding the bindable-property scan, while ghost/bound-target cases are within noise. The tradeoff is now in favor of readability and preserving the existing `WeakValueDictionary` over chasing the smaller micro-optimizations.

Let me know what you think and if there's any other changes you would recommend. Thanks again!

evnchn:
**Net: this is a solid win and looks good to merge as scoped.** I built an independent reverse-index fix for #6150 before finding this PR, so I benchmarked yours head-to-head. It resolves the reported #6150 case (plain refreshables / ghost removal) and the common distinct-source case, with a simpler design than mine — and because it keeps the binding buckets as lists, it avoids a reentrancy foot-gun that a dict-based version would hit (see note 3).

Two shared-source shapes stay O(N²) here. Both are out of this PR's "cleanup path" scope and fit the split-PR series better than expanding this one — flagging with numbers so they're on the radar, not as blockers.

### Follow-up candidates (not blocking)

1. **Many targets bound from one shared source is still O(N²) (major, follow-up).** In `remove()`, the target-only branch rebuilds the whole per-source list (`remaining_bindings = [binding for binding in binding_list if …]`). When one source feeds many targets — e.g. one model bound to many widgets — removing the targets across refreshes rescans the shrinking list each time. Measured on head e4e8cf18, K=10 targets/group: per-refresh cost **0.025 ms → 1.20 ms** from R=100 → R=1000 (~×48 for 10× objects). A per-source entry index (id-keyed) removes it in O(dropped).

2. **`active_links` is full-scanned on every non-ghost `remove()` (major, follow-up).** `_remove_active_links_for_objects` does `active_links[:] = [… if id not in removed]`, i.e. O(len(active_links)) per call. Anything bound *from* a plain dict/dataclass goes through the polling fallback (`bind_from` / `bind_text_from(shared_state, …)`), so `active_links` grows with the app and removal is quadratic. Measured, K=10/group: **0.055 ms → 4.31 ms** from R=100 → R=1000 (~×78). Same reverse-index treatment as `bindings` fixes it.

3. **Heads-up for whichever follow-up tackles (1)/(2): don't convert the buckets to dicts without a snapshot + liveness guard (minor).** The obvious O(1)-removal move is to make `bindings[key]` / `active_links` dicts. If you do, the hot loops (`_propagate_recursively`, `_refresh_step`) must iterate a snapshot *and* re-check the entry is still present before running it — otherwise a `remove()` triggered from inside a transform (e.g. a change handler that deletes an element mid-propagation) raises `RuntimeError: dictionary changed size during iteration`, and a stale snapshot entry can re-write (and resurrect) an already-removed target. The current list-based code is immune, so this only matters if a follow-up switches structures.

@falkoschindler — the #6150 reporter mentions hitting this in production (currently working around it with manual timers), so landing this one sooner would unblock that case directly. Nothing above is a reason to hold it — (1) and (2) are genuine follow-up material for the series, not gaps in this PR.

<details>
<summary>Benchmark method &amp; repros (head e4e8cf18, baseline c40f3e3f, Python 3.13, macOS)</summary>

Per-refresh = total `remove` wall time / R, so a flat number across R is O(N) overall and a rising number is O(N²). K = elements per group.

| scenario | #6135 R=100 → R=1000 | for reference, an id-indexed variant |
|---|---|---|
| distinct-source `bind_to` (common / control) | 0.008 → 0.009 ms  ✓ flat | 0.012 ms flat |
| **(1)** many targets, one shared source | 0.025 → **1.20 ms** | 0.012 ms flat |
| **(2)** `bind_from` one shared plain dict | 0.055 → **4.31 ms** | 0.012 ms flat |

Scenario (1):
```python
shared = Src()                        # one BindableProperty source
def group():
    ts = [T() for _ in range(10)]
    for t in ts:
        binding.bind_to(shared, 'v', t, 'x')
    return ts
groups = [group() for _ in range(R)]
for g in groups:
    binding.remove(g)                 # target-only removal rescans shared list
```

Scenario (2):
```python
state = {'v': 0}                      # plain dict -> polling fallback -> active_links
def group():
    ls = [L() for _ in range(10)]
    for l in ls:
        binding.bind_from(l, 't', state, 'v', backward=str)
    return ls
```

Note 3 repro (only bites a dict-based variant, shown for completeness):
```python
src, t1, t2 = Src(), T(), T()
def forward(v):
    binding.remove([t2])              # remove() from inside a transform, mid-propagation
    return v
binding.bind_to(src, 'v', t1, 'x', forward=forward)
binding.bind_to(src, 'v', t2, 'x')
src.v = 5                             # lists: fine. dict buckets without a liveness guard: RuntimeError
```

</details>

falkoschindler:
Thanks for the simplification — this is now exactly the reverse-index-only design we converged on, and keeping `_bind_one_way` is a good call: both bind paths have to update `bindings`, the reverse index and `active_links` in lockstep, so deduplicating them is load-bearing, not cosmetic.

I re-ran a correctness pass over the branch: the ghost-removal early return is safe because every `active_links` entry is created together with a binding whose endpoints are both indexed, so an object with active links can never miss the reverse index; all three `remove()` branches keep the index consistent (removed objects are popped wholesale, a removed source deindexes its surviving targets, an emptied binding list deindexes its source); and since binding tuples hold strong references while indexed, there is no object-ID-reuse hazard. `pytest`, `pylint`, `ruff` and `mypy` all pass locally.

To save us another round-trip I pushed eight small cleanup commits directly to the branch — none of them change the algorithm:

- Simplified the `nicegui_reset_globals` fixture declarations in the new tests to the bare-argument pattern the rest of the suite uses.
- Removed `test_bindable_properties_keeps_useful_mapping_operations` — with the stdlib `WeakValueDictionary` restored it only asserted standard library behavior. The remaining three tests are good: the two index tests cover exactly the new invariants, and the rebind test pins down the intentional behavior change.
- Dropped the micro-optimizations in the removal path (lazy set allocation in `_pop_binding_keys_for_objects`, list-before-set in `remove()`). I benchmarked both variants on the ghost-removal cases: the plain version is equal or slightly faster (`set(map(id, ...))` runs the id loop in C, while the list comprehension it replaced runs it in bytecode), and it needs fewer branches and no explanatory comments.
- Reverted the unrelated `__set_name__`/`__get__`/`_propagate_recursively` signature polish to keep the diff focused.
- Dropped the `ObjectId`, `NamePath` and `Transform` aliases — as plain renames they add no type checking and `transform: Callable[[Any], Any] | None` tells the reader more than `transform: Transform`. The structural aliases `BindingKey`/`Binding`/`ActiveLink` stay, now with their elements written inline.
- Inlined `_remove_active_links_for_objects` back into `remove()` — its body was main's `active_links` comprehension verbatim, so inlining it (with the original `object_ids` name) turns those lines back into unchanged code and keeps the diff smaller.
- Made the reverse index a `defaultdict(set)` (matching `bindings`) and inlined the two-line `_index_binding` helper into `_bind_one_way` — `_binding_keys_by_object[obj_id].add(...)` needs no `setdefault` and only allocates the empty set on an actual miss. This also let `_discard_binding_key_from_object_index` drop its `None` guard: a missing entry now yields an empty set that falls straight into the `del` branch. The `.pop()` in the removal path doesn't trigger the factory, so no ghost entries can appear.
- Reordered the functions to follow the module's public-first convention and compacted the reverse-index helpers, including collapsing `_pop_binding_keys_for_objects` into a single set comprehension (`pop(obj_id, ())` — the empty-tuple default avoids a per-miss allocation, and `defaultdict` factories don't apply to `pop`).

I also updated the PR description: the Implementation section still described the `_BindablePropertyRegistry` design that the branch has since moved away from, and the benchmark numbers were measured on that earlier revision. It now describes the reverse-index design (including the intentional `bindable_properties` behavior change) with fresh numbers measured on the final state — same three case shapes as your original harness, timing only the `remove()` calls. Please double-check that I represented your motivation correctly.

We're not entirely done reviewing yet — I want to take one last look before wrapping up — but I didn't want to leave you without an intermediate status.

falkoschindler (approved):
I pushed four small follow-ups on top:

- 2b33261ed drops the unreachable `None` guard in `remove()` — the reverse index and `bindings` are kept in lockstep, so the key is always present
- d0fcace3c deduplicates target IDs when discarding reverse-index entries, so duplicate bindings don't briefly resurrect already-deleted index entries
- bca60c3d2 moves the bindable-property expiry note from an inline comment into the `remove()` docstring
- f0b74dc20 replaces the white-box index assertions with behavioral tests (propagation after partial and full removal), keeping two one-line emptiness checks as a leak guard

Everything looks good now. Thanks, @jwinpbe! Let's merge 🚀

PR #6137, opened 2026-07-02
Fix crash when loading STL with wireframe=True

Jepson2k:
### Motivation

`scene.stl(url, wireframe=True)` is broken on `main`: the `stl` case in `scene.js` creates an empty `BufferGeometry` synchronously and immediately hands it to `EdgesGeometry`, which throws reading the position attribute during creation — the object is never registered — and the async load callback then throws again assigning `.geometry` to the never-created mesh. Result: two uncaught TypeErrors in the console and nothing rendered.

Full analysis with a minimal repro and before/after screenshots by @evnchn in https://github.com/zauberzeug/nicegui/pull/5989#issuecomment-4862310342. This PR splits the fix out of #5989 so it can land ahead of the feature work there, as suggested in that review.

### Implementation

The `stl` path now mirrors the `gltf` path: `scene.stl` creates a `THREE.Group` placeholder and builds the actual `LineSegments`/`EdgesGeometry` (or `Mesh`) inside the loader callback, once the real geometry exists. Material application on the not-yet-loaded group is deferred via the existing `pendingMaterialInfo` mechanism and flushed when loading completes.

Within loaded model groups (STL and glTF), `material()` now applies to every child that has a material. This is what lets wireframe STLs (`LineSegments`) receive color/opacity/side updates, and it also fixes a small related glTF bug: an explicit `material()` call previously recolored only mesh primitives, silently skipping line and point primitives. Since #6118, glTF materials are only applied when the user explicitly sets one, so model-defined materials remain untouched otherwise.

The loader callback also stamps the element's `object_id` and name on the raycast-hittable child, so STL objects keep producing `click3d` hits exactly as they did on `main` (a Group is never the intersected object, and the click handler filters hits without an `object_id`); `name()` forwards later renames to that child so hits don't report a stale name.

The regression test (based on @evnchn's ready-made test from the #5989 review, thanks!) loads a minimal 12-triangle ASCII cube fixture (`tests/media/cube.stl`) with `wireframe=True` and an explicit red material, and asserts a `Group` wrapper with a non-empty `EdgesGeometry` child carrying the element's `object_id` and the material color — it fails with the crash on `main` and passes with this fix. A button click then renames the object after the async load and the test asserts the new name reaches the raycast-hittable child.

### Progress

### DIFF
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -335,6 +335,33 @@ export default {
           undefined,
           (error) => console.error(error),
         );
+      } else if (type == "stl") {
+        const url = args[0];
+        const wireframe = args[1];
+        mesh = new THREE.Group();
+        mesh.userData.isStl = true;
+        mesh.userData.loaded = false;
+        mesh.userData.wireframe = wireframe;
+        this.stl_loader.load(
+          url,
+          (geometry) => {
+            const child = wireframe
+              ? new THREE.LineSegments(
+                  new THREE.EdgesGeometry(geometry),
+                  new THREE.LineBasicMaterial({ transparent: true }),
+                )
+              : new THREE.Mesh(geometry, new THREE.MeshPhongMaterial({ transparent: true }));
+            mesh.add(child);
+            mesh.userData.loaded = true;
+            if (mesh.userData.pendingMaterialInfo) {
+              const { color, opacity, side } = mesh.userData.pendingMaterialInfo;
+              delete mesh.userData.pendingMaterialInfo;
+              this.material(id, color, opacity, side);
+            }
+          },
+          undefined,
+          (error) => console.error("STL load error:", error),
+        );
       } else if (type == "axes_helper") {
         mesh = new THREE.AxesHelper(args[0]);
         mesh.material.transparent = true;
@@ -365,11 +392,6 @@ export default {
           const settings = { depth: height, bevelEnabled: false };
           geometry = new THREE.ExtrudeGeometry(shape, settings);
         }
-        if (type == "stl") {
-          const url = args[0];
-          geometry = new THREE.BufferGeometry();
-          this.stl_loader.load(url, (geometry) => (mesh.geometry = geometry));
-        }
         let material;
         if (wireframe) {
           mesh = new THREE.LineSegments(
@@ -392,7 +414,7 @@ export default {
     material(object_id, color, opacity, side) {
       const object = this.objects.get(object_id);
       if (!object) return;
-      if (object.userData.isGltf && !object.userData.loaded) {
+      if ((object.userData.isGltf || object.userData.isStl) && !object.userData.loaded) {
         object.userData.pendingMaterialInfo = { color, opacity, side };
         return;
       }
@@ -408,8 +430,8 @@ export default {
           else m.side = THREE.DoubleSide;
         });
       };
-      if (object.userData.isGltf) {
-        object.traverse((child) => child.isMesh && child.material && apply(child.material));
+      if (object.userData.isGltf || object.userData.isStl) {
+        object.traverse((child) => (child.isMesh || child.isLine) && child.material && apply(child.material));
       } else if (object.material) {
         apply(object.material);
       }
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -339,7 +339,7 @@ export default {
             }
           },
           undefined,
-          (error) => console.error(error),
+          (error) => console.error("GLTF load error:", error),
         );
       } else if (type == "stl") {
         const url = args[0];
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -426,9 +426,7 @@ export default {
         });
       };
       if (object.userData.isGltf || object.userData.isStl) {
-        object.traverse(
-          (child) => (child.isMesh || (object.userData.isStl && child.isLine)) && child.material && apply(child.material),
-        );
+        object.traverse((child) => child.material && apply(child.material));
       } else if (object.material) {
         apply(object.material);
       }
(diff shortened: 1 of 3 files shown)


evnchn:
*Posted by Claude Code on @evnchn's behalf (maintainer edits to this draft PR).*

**TL;DR:**

Pushed two commits to this branch — a bone-dry cleanup and a **regression fix this PR needs before it lands.**

1. **This PR introduces a `click3d` naming regression** (not pre-existing — bisected against `main`). Wrapping the STL in a `THREE.Group` means `with_name()` *after* the STL loads updates the wrapper, not the raycast-hit child, so clicks/drags report a stale `object_name`. On `main` the STL mesh *was* the hit object, so this just worked. Fixed in 6465dcb, with a regression test that fails on the pre-fix code.
2. **Bone-dry refactor** (e19fddd): the STL loader callback hand-rolled the same wireframe/solid construction already in the primitives branch — extracted a `mesh_from_geometry()` helper so it lives in one place. Net −8 lines, no behavior change.

Both verified empirically and cleared by an independent (non-Claude) review pass. Ask: sanity-check the two commits, especially the naming fix's `isStl` scoping.

<details>
<summary><b>The introduced regression — main vs PR vs fix (bisect + MRE)</b></summary>

Same public-API MRE run against three states. Observes the symptom through the **public event surface** (`ui.scene(on_click=…)` → `e.hits[0].object_name`), never internal object state:

```python
import time
from selenium.webdriver.common.action_chains import ActionChains
from nicegui import app, ui
from .test_helpers import TEST_DIR


def _click_name(screen: Screen, rename_after_load: bool) -> object:
    reported = []
    obj = None

    @ui.page('/')
    def page():
        nonlocal obj
        app.add_static_file(local_file=TEST_DIR / 'media' / 'cube.stl', url_path='/cube.stl')
        with ui.scene(width=400, height=400, grid=False,
                      on_click=lambda e: reported.append(e.hits[0].object_name if e.hits else None)) as scene:
            obj = scene.stl('/cube.stl').move(-0.5, -0.5, -0.5).scale(2.0)
            if not rename_after_load:
                obj.with_name('cube')        # named at creation (before the async load resolves)

    screen.open('/')
    time.sleep(3.0)                          # wait out the async load (no public "loaded" event exists)
    if rename_after_load:
        obj.with_name('cube')                # public API rename, AFTER load
        time.sleep(1.0)
    ActionChains(screen.selenium).move_to_element(
        screen.selenium.find_element('css selector', 'canvas')).click().perform()
    time.sleep(1.0)
    assert reported, 'click3d never fired (raycast missed the cube)'
    return reported[-1]


def test_name_set_at_creation_is_reported(screen: Screen):   # CONTROL
    assert _click_name(screen, rename_after_load=False) == 'cube'


def test_name_set_after_load_is_reported(screen: Screen):    # BUG
    assert _click_name(screen, rename_after_load=True) == 'cube'
```

| code state | name at creation (control) | rename after load |
|---|---|---|
| `main` (`3e278b6`) | ✅ `'cube'` | ✅ `'cube'` |
| this PR, pre-fix (`4325272`) | ✅ `'cube'` | ❌ `None` |
| this PR + fix (`6465dcb`) | ✅ | ✅ |

`main` passing the second column is the proof it's a **regression this PR introduces**, not a latent bug. The MRE proves it's **user-visible** through the click event; the committed `test_stl_name_after_load` asserts the child name directly (deterministic, matches this PR's own `test_stl_wireframe` style, so it won't be click-flaky in CI).

Root cause: `scene.stl` now wraps the mesh in a `Group` whose child carries `object_id`; `child.name = mesh.name` copies the name only at load time, and `name()` updated only the wrapper. Fix propagates to `object.children` when `userData.isStl` (covers rename-after-load; the loader's `child.name = mesh.name` covers rename-before-load), scoped so gltf is untouched.
</details>

<details>
<summary><b>Bone-dry refactor (e19fddd) — one construction site, not two</b></summary>

The new STL loader callback rebuilt the wireframe→`LineSegments`+`EdgesGeometry` / solid→`Mesh` object by hand — the exact logic already in the primitives branch (which also carried a vestigial unused `let material`). Extracted:

```js
function mesh_from_geometry(geometry, wireframe) {
  return wireframe
    ? new THREE.LineSegments(new THREE.EdgesGeometry(geometry), new THREE.LineBasicMaterial({ transparent: true }))
    : new THREE.Mesh(geometry, new THREE.MeshPhongMaterial({ transparent: true }));
}
```

Called from both sites. Sits beside the existing `texture_material` / `set_point_cloud_data` helpers (house style). Independent review confirmed it reproduces both original call-sites exactly, the dropped `material` local was unused, and gltf + all other object types are byte-for-byte unchanged.
</details>

<details>

- Full `tests/test_scene.py`: **17 passed** (16 + the new regression test).
- `test_stl_name_after_load`: **fails** on 4325272 (child name never becomes `renamed` → timeout), **passes** with 6465dcb.
- Both commits reviewed by an independent non-Claude model (fresh lineage) before push; findings addressed.
- Local `pre-commit` clean on each push (same gate as CI `_check`).

Commits: e19fddd (refactor) · 6465dcb (regression fix + test), on top of 4325272.
</details>

Thanks @Jepson2k — the original split-out fix and click-hit preservation are solid; this is just the one missing propagation the `Group` wrapper needed, plus a DRY pass. Happy to squash/reword or drop the refactor commit if you'd rather keep it separate.

falkoschindler (approved):
Thanks @Jepson2k, nice fix — and thanks @evnchn for the thorough analysis and the ready-made test! Mirroring the gltf pattern is exactly the right shape for this, and I verified the regression test genuinely fails on `main`.

I pushed three small follow-up commits:

- aligned the gltf load-error message with the new STL one ("GLTF load error:"),
- merged the two STL tests into a single Screen test which now also triggers the post-load rename via a button click,
- simplified the `material()` traversal to apply to every child with a material within loaded model groups — this drops the slightly convoluted STL-scoped `isLine` condition and, as a bonus, fixes explicit `material()` calls silently skipping line/point primitives in glTF models (model-defined materials are still preserved thanks to #6118). The test now also asserts that a red material actually reaches the wireframe's `LineSegments`.

I updated the PR description to match. All scene tests pass locally. Let's merge! 🚀

PR #6138, opened 2026-07-04
Update nicegui-highcharts to 3.3 and document dynamic series updates

falkoschindler:
### Motivation

nicegui-highcharts 3.3.0 fixes adding and removing chart series (zauberzeug/nicegui-highcharts#1, zauberzeug/nicegui-highcharts#28) by switching to Highcharts' [one-to-one updating](https://api.highcharts.com/class-reference/Highcharts.Chart#update). Bumping the dependency lets the website and the `highcharts` extra pick up the new version.

### Implementation

- Raise both `nicegui-highcharts` pins in `pyproject.toml` from `>=3.2.1,<4` to `>=3.3.0,<4`.
- Since the fix makes the `options` dictionary the single source of truth for series and axes, add a documentation demo "Adding and removing series" covering the new contract: series and axes are synchronized declaratively, explicit series `id`s are recommended for dynamic add/remove, and server updates during drilldown reset the drill view (known limitation).

### Progress

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ matplotlib = [
     "fonttools>=4.60.2",  # https://github.com/zauberzeug/nicegui/security/dependabot/68, for matplotlib
     "pillow>=12.2.0",  # https://github.com/zauberzeug/nicegui/security/dependabot/245, for matplotlib
 ]
-highcharts = ["nicegui-highcharts>=3.2.1,<4"]
+highcharts = ["nicegui-highcharts>=3.3.0,<4"]
 redis = ["redis>=4.0.0"]
 
 [dependency-groups]
@@ -85,7 +85,7 @@ dev = [
     "matplotlib>=3.5.0,<4",  # extra: matplotlib
     "fonttools>=4.60.2",  # https://github.com/zauberzeug/nicegui/security/dependabot/68, for matplotlib
     "pillow>=12.2.0",  # https://github.com/zauberzeug/nicegui/security/dependabot/245, for matplotlib
-    "nicegui-highcharts>=3.2.1,<4",  # extra: highcharts
+    "nicegui-highcharts>=3.3.0,<4",  # extra: highcharts
     "redis>=4.0.0",  # extra: redis
     {include-group = "types"},
     {include-group = "website"},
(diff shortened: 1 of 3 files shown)


evnchn (approved):
**Approving.** One non-blocking docs-precision nit, then the verification is folded below.

**Nit (non-blocking):** the line *"State added on the client, e.g. via JavaScript, does not survive updates"* reads a little absolute, and sits right next to the following paragraph noting that legend-click visibility state *does* persist across updates (which is exactly why it can end up on the wrong series without an `id`). Those are two different "client states":

- structural additions not present in `options` → wiped by `oneToOne`
- interaction state on tracked series → persists, and can mis-map

Maybe reword to something like *"Client-side changes that are not reflected in `options` may be overwritten or lost on updates."* — but I'm happy to approve as-is either way.

<details>
<summary><b>Verification — bump is safe and the docs match the actual 3.3.0 change</b></summary>

Checked end-to-end (nothing here is a blocker):

| Check | Result |
|---|---|
| wheel `sha256` vs `uv.lock` | ✅ `13fb5865…e2d920` match |
| sdist `sha256` vs `uv.lock` | ✅ `84848fe1…37125d` match |
| doc text vs the real 3.3.0 source diff | ✅ accurate |
| demo pattern | ✅ same `chart.options` mutation as the shipped `test_adding_chart_series` / `test_removing_chart_series` (auto-sync after the event handler — no explicit `chart.update()` needed) |
| CI incl. the Selenium highchart suite | ✅ green → existing add/remove/update behavior preserved under the new update path |

**Why the bump matters** — diffing the 3.2.1 vs 3.3.0 wheels, the only behavioral change in `highchart.js` is that the manual bookkeeping (a `seriesCount` counter with `addSeries({}, false)` / `chart.series[0].remove()` loops) is deleted in favour of:

```js
this.chart.update(this.options, true, true);  // redraw=true, oneToOne=true
```

`oneToOne=true` is precisely the "`options` is the single source of truth; series/axes added, updated and removed to match" behavior the new demo documents, and the id-vs-position mapping caveat is the standard Highcharts collection semantics. So the version bump and the doc addition are coherent — the docs describe a real capability the bump enables.

</details>

<details>
<summary><b>Residual doubt I did not block on</b></summary>

The `extras=['drilldown']` reset note is plausible and conservative, but it isn't exercised by CI (and `full-test` is skipped on this run), so it's documented-not-regression-tested. Not worth holding the PR for; a follow-up Selenium test could lock it down later if that behavior is worth pinning.

Cross-checked the diff and this verdict with a second (non-Claude) model as a skeptic pass — it concurred with approve; the client-state wording nit above is the one thing it sharpened.

</details>

PR #6140, opened 2026-07-04
Detach removed items from ObservableCollection and reference containers weakly

### LINKED ISSUE #6139: ObservableCollection leaks change handlers when observable items are removed or replaced
### First Check

### Example Code

```python
from nicegui.observables import ObservableList

items = ObservableList([{}], on_change=lambda: print('change'))
for _ in range(3):
    items[:] = list(items)  # prints "change" 3 times in total

items[0]['x'] = 1  # prints "change" 4 times instead of once
```

### Description

`ObservableCollection._observe` registers the container's `_handle_change` on any item that is already an `ObservableCollection` ([observables.py, lines 45-47](https://github.com/zauberzeug/nicegui/blob/main/nicegui/observables.py#L45-L47)), and no removal or replacement operation (`remove`, `pop`, `del`, `clear`, item assignment) ever deregisters it. Handlers therefore accumulate:

- Rebuilding a list in place (`lst[:] = [...]`) wraps the incoming plain list in a temporary `ObservableList` whose `__init__` re-observes every surviving item. The temporary list is discarded immediately, but it stays reachable via the bound `_handle_change` handlers it left behind — one extra handler per rebuild, forever.
- Plain re-assignment (`options['series'] = [...]`) behaves the same way: the replacement list re-observes the surviving items, and the handlers pointing at the discarded old list remain registered.
- After N rebuilds, a single mutation of a surviving item fires N+1 change events instead of one (i.e. N+1× `element.update()` when the collection backs an element's props), and none of the discarded containers can be garbage-collected.

This hits the common pattern of updating element options by rebuilding a list, e.g. `chart.options['series'] = [s for s in chart.options['series'] if ...]` on `ui.highchart`/`ui.echart`/`ui.aggrid`. A long-running app that does this repeatedly accumulates handlers and dead containers without bound. (The "Adding and removing series" demo added in #6138 originally used this pattern; it now uses `list.remove` as a workaround.)

A related wart with the same root cause: re-inserting an item into a collection it already belongs to registers a duplicate handler (its `_parent` already bubbles events), so subsequent changes fire twice. `on_change` only guards against the item's *own* handler, not against duplicates or the parent's.

Expected behavior: removing or replacing an item detaches it from the container, so a subsequent mutation of a surviving item fires exactly one change event and discarded containers can be collected.

A fix probably needs explicit deregistration bookkeeping: an item can belong to several containers (which is why the single `_parent` reference is complemented by the extra-handler mechanism), so removal operations would have to unregister exactly the handler belonging to the removing container. Since `Props`, `storage`, and bindings all build on observables, the change deserves careful tests.

### NiceGUI Version

main (3.14.0.post7.dev0)

### Python Version

3.11.7

### Browser

Other

### Operating System

macOS

### Additional Context

Browser and operating system are irrelevant — the behavior is purely server-side and reproducible without running an app.

falkoschindler:
### Motivation

Fixes #6139: `ObservableCollection._observe` registered the container's `_handle_change` on observable items, but no removal or replacement operation ever deregistered it. Handlers accumulated on every rebuild (`lst[:] = [...]`, `options['series'] = [...]`), so after N rebuilds a single item mutation fired N+1 change events, and discarded containers could never be garbage-collected. Re-inserting an item into a collection it already belonged to registered a duplicate handler.

### Implementation

The item→container bookkeeping is reworked in four ways:

1. **Weak references from items to their containers.** Instead of registering strong bound-method handlers on items, observing containers are tracked in a new `_observer_refs` list of weak references, and `_parent` is now backed by a weak reference as well. A discarded container is garbage-collected and its registrations vanish with it. (Rationale: an item cannot know whether a container that still holds it has been discarded — only weak references make discarded containers collectable, which is what the reassignment pattern `options['series'] = [...]` requires.)
2. **Idempotent observation.** `_register_observer` skips containers that already observe the item (directly or as its parent), fixing the duplicate-handler wart on re-insertion.
3. **Per-item observation for bulk inserts.** `extend`, `update`, slice assignment, `|=`, `+=` etc. now observe each incoming item individually instead of wrapping the whole incoming collection in a temporary observable — previously that temporary stayed reachable via the handlers it left on surviving items.
4. **Explicit detach on removal.** Every removal/replacement operation calls the new `_unobserve`, which detaches items that are no longer contained (identity-based check, so items contained multiple times or shared between collections stay attached as long as one occurrence remains). `setdefault` also no longer observes the default value when the key already exists.

Both mechanisms are needed: explicit deregistration alone cannot fix the reassignment pattern (the discarded old wrapper still *contains* the surviving items, so no removal-based bookkeeping ever detaches it), and weak references alone cannot fix removal correctness (`item = data.pop('x'); item['y'] = 1` must not notify the still-alive `data`).

Behavioral notes:

- Mutating an item that was removed from a collection no longer notifies that collection — this is the issue's expected behavior, but technically a semantic change.
- If user code holds only a nested item while the root collection is discarded, the root is now garbage-collected and its handlers stop firing (previously the item kept the whole chain alive). Within NiceGUI, roots (props, storage, bindings) are always independently referenced.
- Nested observables are no longer picklable (weakrefs aren't picklable). Pickling was never a supported path — storage serializes to JSON, and `__copy__`/`__deepcopy__` are implemented explicitly.

Open questions for reviewers:

- **Fully weak vs. weak observers only:** Keeping `_parent` *strong* would restore picklability of plain nested trees and preserve the old "a held child keeps its discarded root alive and notifying" behavior, while still fixing the unbounded per-rebuild growth. The cost: the original wrapper of each surviving item stays alive forever (bounded at one dead container per item). This PR goes fully weak for consistency and to satisfy "discarded containers can be collected" verbatim — but the strong-`_parent` middle ground is a defensible alternative.
- **`ObservableSet` symmetry:** The `_unobserve`/per-item changes in `ObservableSet` are dead code in practice, since observables are unhashable and can never be set members. They are included for symmetry but could be dropped to shrink the diff.
- A fire-time containment check (notify only observers that still contain the sender) was considered as a smaller alternative that would leave all mutation methods untouched, but was rejected because it puts an O(container size) identity scan on every change event instead of a cost on the much rarer removal operations.

### Progress

- [x] No breaking changes to the public API; subtle behavioral changes are described above.

### DIFF
--- a/nicegui/observables.py
+++ b/nicegui/observables.py
@@ -33,9 +33,8 @@ class ObservableCollection(abc.ABC):  # noqa: B024
 
     @_parent.setter
     def _parent(self, parent: ObservableCollection | None) -> None:
-        # NOTE: the reference is weak so that items don't keep discarded parent collections alive
-        self._parent_ref: weakref.ref[ObservableCollection] | None = \
-            weakref.ref(parent) if parent is not None else None
+        # the reference is weak so that items don't keep discarded parent collections alive
+        self._parent_ref = weakref.ref(parent) if parent is not None else None
 
     @property
     def change_handlers(self) -> list[Callable]:
@@ -44,8 +43,7 @@ class ObservableCollection(abc.ABC):  # noqa: B024
 
     @property
     def _change_handlers_with_args(self) -> list[tuple[Callable, bool]]:
-        """Return all change handlers and their pre-resolved `expect_args` flag,
-        including those of parents and observers."""
+        """Return change handlers with pre-resolved ``expect_args`` flag, including those of parents and observers."""
         change_handlers = self._change_handlers[:]
         observers = [ref() for ref in self._observer_refs]
         change_handlers.extend((observer._handle_change, False)  # pylint: disable=protected-access
@@ -65,12 +63,7 @@ class ObservableCollection(abc.ABC):  # noqa: B024
             self._change_handlers.append((handler, helpers.expects_arguments(handler)))
 
     def _register_observer(self, observer: ObservableCollection) -> None:
-        """Register a collection which contains this collection and should be notified about changes.
-
-        Registrations are idempotent: a collection which already observes this collection
-        (directly or as its parent) is not registered again.
-        The observer is referenced weakly so that it can be garbage-collected when it is discarded.
-        """
+        """Register a collection which contains this collection and should be notified about changes."""
         if observer is self or observer is self._parent:
             return
         alive_refs = [ref for ref in self._observer_refs if ref() is not None]
@@ -97,7 +90,6 @@ class ObservableCollection(abc.ABC):  # noqa: B024
         return data
 
     def _unobserve(self, *items: Any) -> None:
-        """Detach removed items so that they no longer notify this collection about changes."""
         removed = [item for item in items if isinstance(item, ObservableCollection)]
         if not removed:
             return
@@ -126,8 +118,8 @@ class ObservableCollection(abc.ABC):  # noqa: B024
         raise NotImplementedError(f'ObservableCollection.__deepcopy__ not implemented for {type(self)}')
 
     def __reduce__(self) -> tuple[Any, tuple]:
-        # NOTE: reconstruct from plain contents so that the observer wiring (weak references, which are not
-        # picklable) is rebuilt by __init__ instead of being pickled; a freshly loaded tree has no observers yet.
+        # reconstruct from plain contents so that the observer wiring (weak references, which are not picklable)
+        # is rebuilt by __init__ instead of being pickled; a freshly loaded tree has no observers yet.
         if isinstance(self, dict):
             return ObservableDict, (dict(self),)
         if isinstance(self, list):
@@ -197,11 +189,7 @@ class ObservableDict(ObservableCollection, dict):
         return super().__or__(other)
 
     def __ior__(self, other: Any) -> Any:
-        new_items = dict(other)
-        old_values = [self[key] for key in new_items if key in self]
-        super().__ior__({key: self._observe(value) for key, value in new_items.items()})
-        self._unobserve(*old_values)
-        self._handle_change()
+        self.update(other)
         return self
 
 
@@ -230,9 +218,7 @@ class ObservableList(ObservableCollection, list):
         self._handle_change()
 
     def remove(self, value: Any) -> None:
-        item = super().pop(super().index(value))
-        self._unobserve(item)
-        self._handle_change()
+        self.pop(super().index(value))
 
     def pop(self, index: SupportsIndex = -1) -> Any:
         item = super().pop(index)
@@ -275,8 +261,7 @@ class ObservableList(ObservableCollection, list):
         return super().__add__(other)
 
     def __iadd__(self, other: Any) -> Any:
-        super().__iadd__([self._observe(item) for item in other])
-        self._handle_change()
+        self.extend(other)
         return self
 
     def __mul__(self, other: Any) -> Any:
@@ -329,8 +314,7 @@ class ObservableSet(ObservableCollection, set):
         self._handle_change()
 
     def update(self, *s: Iterable[Any]) -> None:
-        items: set = set(*s)
-        super().update({self._observe(item) for item in items})
+        super().update({self._observe(item) for item in set().union(*s)})
         self._handle_change()
 
     def intersection_update(self, *s: Iterable[Any]) -> None:
@@ -347,8 +331,7 @@ class ObservableSet(ObservableCollection, set):
 
     def symmetric_difference_update(self, *s: Iterable[Any]) -> None:
         old_items = list(self)
-        items: set = set(*s)
-        super().symmetric_difference_update({self._observe(item) for item in items})
+        super().symmetric_difference_update({self._observe(item) for item in set().union(*s)})
         self._unobserve(*old_items)
         self._handle_change()
 
@@ -356,36 +339,26 @@ class ObservableSet(ObservableCollection, set):
         return super().__or__(other)
 
     def __ior__(self, other: Any) -> Any:
-        super().__ior__({self._observe(item) for item in other})
-        self._handle_change()
+        self.update(other)
         return self
 
     def __and__(self, other: Any) -> set:
         return super().__and__(other)
 
     def __iand__(self, other: Any) -> Any:
-        old_items = list(self)
-        super().__iand__(other)
-        self._unobserve(*old_items)
-        self._handle_change()
+        self.intersection_update(other)
         return self
 
     def __sub__(self, other: Any) -> set:
         return super().__sub__(other)
 
     def __isub__(self, other: Any) -> Any:
-        old_items = list(self)
-        super().__isub__(other)
-        self._unobserve(*old_items)
-        self._handle_change()
+        self.difference_update(other)
         return self
 
     def __xor__(self, other: Any) -> set:
         return super().__xor__(other)
 
     def __ixor__(self, other: Any) -> Any:
-        old_items = list(self)
-        super().__ixor__({self._observe(item) for item in other})
-        self._unobserve(*old_items)
-        self._handle_change()
+        self.symmetric_difference_update(other)
         return self
--- a/nicegui/observables.py
+++ b/nicegui/observables.py
@@ -279,6 +279,16 @@ class ObservableList(ObservableCollection, list):
         self._handle_change()
         return self
 
+    def __mul__(self, other: Any) -> Any:
+        return super().__mul__(other)
+
+    def __imul__(self, other: Any) -> Any:
+        old_items = list(self)
+        super().__imul__(other)
+        self._unobserve(*old_items)
+        self._handle_change()
+        return self
+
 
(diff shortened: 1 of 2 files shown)


evnchn:

**TL;DR — reviewed in depth, and pushed one follow-up commit (334a081d) that closes the only real regression. With it, this is effectively pure-win. LGTM.**

The core fix is correct and complete:

- The #6139 leak is genuinely dead — the same 62,000 in-place rebuilds that grew RSS **+1.00 GiB** on `main` now grow **+0 MB**, and the N+1 change-event fan-out is gone.
- A fresh-lineage (Codex) pass flagged 4 things; I verified each against release 3.14.0. **3 are pre-existing warts, not introduced here.** The 1 genuinely new regression — nested observables becoming unpicklable — is what 334a081d fixes.

The pickle regression mattered because it's reachable through a normal public path (`run.cpu_bound(fn, <storage/binding data>)`) and its obvious work-arounds were *also* broken. Details + MRE below.

<details>
<summary><b>The leak is fixed — empirical, not by reading</b></summary>

Ran against the branch head; `.venv`, macOS, Python 3.13:

| Metric | `main` (pre-fix) | this PR |
|---|---|---|
| RSS after 62,000 in-place rebuilds | **+1.00 GiB** | **+0 MB** |
| Live `ObservableList` after `gc.collect()` | 62,001 pinned | 1 |
| Handlers on a surviving item | 62,000 | 0 |
| One mutation after N rebuilds fires | N+1 events | **1** |

Plus 8 adversarial probes beyond the PR's own tests all pass (real chart-filter pattern `options['series'] = [s for s in series if …]` → 0 MB over 20k rebuilds; shared item across two roots; remove-then-readd; 3-level nesting; held-child no longer pins discarded root). `mypy` clean, `pylint` 10.00/10, all `test_observables` pass. The one storage failure I saw (`test_storage_serialization_error…`) is a browser-timing flake — passes on retry, and I confirmed the serialization path directly (`ObservableSet` is still a `set` subclass, so it's still rejected and the offending key is still reported).
</details>

<details>
<summary><b>Codex fresh-lineage findings — regression vs. pre-existing (verified on 3.14.0)</b></summary>

| # | Finding | old | new | Verdict |
|---|---|---|---|---|
| F1 | nested observable not picklable | picklable | `TypeError` | **New regression → fixed in 334a081d** |
| F2 | `ObservableSet.remove/discard` identity mismatch for a *hashable* observable subclass | never detaches | contrived | Not a regression — needs a deliberately-hashable observable set member; the PR already flags this `ObservableSet` item code as dead/droppable, and F2 shows it's subtly wrong if reached (passes the caller's value to `_unobserve`, not the resident, unlike `ObservableList.remove`). |
| F3 | "diamond" (same item both a direct child and nested in a sibling) double-notifies root | 2 | 2 | **Pre-existing**, identical before/after. Out of scope. |
| F4 | `deepcopy` of a nested child keeps notifying the *original* root | 2 | 2 | **Pre-existing** — `__copy__`/`__deepcopy__` untouched here. Arguably a real latent bug (a deep copy should detach); worth a separate follow-up issue, not a blocker for this PR. |
</details>

<details>
<summary><b>The pickle regression (F1) and why it needed fixing — MRE</b></summary>

`run.cpu_bound` offloads to a `ProcessPoolExecutor`, which pickles its arguments. Users routinely pull structured data out of `app.storage` (or a bound model) — which is observable — and hand it to a worker. This MRE spawns a **real** worker process and transfers the arg (the exact path `cpu_bound` takes):

```python
from concurrent.futures import ProcessPoolExecutor
from nicegui import app

def count_tags(series):  # runs in the worker process
    return sum(len(s['tags']) for s in series)

if __name__ == '__main__':
    app.storage.general['series'] = [{'id': 1, 'tags': ['a', 'b']}, {'id': 2, 'tags': ['c']}]
    data = app.storage.general['series']            # -> ObservableList[ObservableDict]
    with ProcessPoolExecutor() as ex:
        print(ex.submit(count_tags, data).result())
```

```text
# before 334a081d (PR as first pushed):
TypeError: cannot pickle 'weakref.ReferenceType' object
# after 334a081d:
3
```

What made it nastier than "just don't pickle observables": the two intuitive work-arounds both failed too —

| object | pickles? (before fix) |
|---|---|
| flat root `ObservableList([1,2,3])` | ✅ |
| any **nested** observable (real storage data) | ❌ |
| `[dict(i) for i in items]` (shallow) | ❌ nested values still observable |
| `copy.deepcopy(items)` (the *intuitive* fix) | ❌ `__deepcopy__` re-attaches the weak parent |
| `json.loads(json.dumps(items))` | ✅ (JSON-only) |

Storage backup (JSON) and native-mode (pickles window config, not observables, and drops unpicklables gracefully) are unaffected — I checked both.
</details>

<details>
<summary><b>What 334a081d does</b></summary>

Adds `__reduce__` to `ObservableCollection`, reconstructing each collection from its plain contents so the weak observer wiring is rebuilt by `__init__` (a freshly loaded tree has no observers yet) instead of being pickled:

```python
def __reduce__(self) -> tuple[Any, tuple]:
    if isinstance(self, dict):
        return ObservableDict, (dict(self),)
    if isinstance(self, list):
        return ObservableList, (list(self),)
    if isinstance(self, set):
        return ObservableSet, (set(self),)
    raise NotImplementedError(...)
```

(A naive `__getstate__`/`__setstate__` doesn't work — pickle re-populates a list/dict subclass via the overridden `extend`/`__setitem__`, which fire `_handle_change` before state is restored → `AttributeError` on `_change_handlers`. `__reduce__` sidesteps that.) `__copy__`/`__deepcopy__` are unaffected (they take precedence over `__reduce__` in the `copy` module). Verified: round-trips, preserves nested `ObservableList`/`ObservableDict`/`ObservableSet` types, restored tree wires up freshly and fires exactly once, and the leak fix stays intact. Added `test_nested_observables_are_picklable`; `pre-commit` / `mypy` / `pylint` 10.00 / 16 `test_observables` all green.
</details>

**Over to you, @falkoschindler:**

- I think the fully-weak `_parent` choice holds up — `_observer_refs` self-prune, no unbounded dead-weakref accumulation, and "held child doesn't pin discarded root" is verified. With 334a081d, the pickle downside is gone too, so I'd keep it fully weak.
- Agree `ObservableSet`'s item-handling is dead code (and F2 shows it's subtly wrong if reached) — dropping it would shrink the diff.
- F4 (deepcopy should detach) is pre-existing and independent — happy to open a separate issue if you agree it's worth one.

Feel free to squash 334a081d into the main commit or drop it if you'd rather handle pickling differently — it's self-contained.

falkoschindler:
Hi @evnchn, thanks again for the deep-dive review and your follow-up commit 334a081d2!

Since then, three more commits landed on the branch:

- 9ef2336b3 closes the last gap of this kind: `lst *= 0` now detaches the removed items as well (new `ObservableList.__imul__`, with test).
- a2a3d4e1b simplifies the implementation: all augmented operators now delegate to their named counterparts (`__iadd__` → `extend`, `__ior__` → `update`, etc.), removing ~27 lines of duplicated detach logic. This incidentally fixes two pre-existing warts: `s &= [1, 2]` silently did nothing (while still firing a change event), and `ObservableSet.update` with multiple iterables raised a `TypeError`.
- 8c51e15ba adds test coverage for `ObservableDict.clear` detaching its values.

We also ran an in-depth Claude Code review over the full diff — no blocking or major findings, it considers the PR clean and ready to merge.

Since your review was "only" a comment so far: would you mind giving it a formal approval so we can merge? 🙏

evnchn (approved):
*Reviewed by Claude Code on evnchn's behalf.*

**Approve** — correct, well-scoped fix; verified locally against the two real consumers (storage + bindings).

<details><summary><b>Verification + a note on out-of-scope findings</b></summary>

Ran on the PR head (`8c51e15b`):

| Gate | Result |
|---|---|
| `tests/test_observables.py` | 17/17 pass |
| `tests/test_storage.py` + `tests/test_binding.py` | 47/47 pass |
| `mypy ./nicegui/observables.py` | clean |
| `pylint ./nicegui/observables.py` | 10.00/10 |

Blast radius is contained: `change_handlers` / `_parent` have no external consumers in the tree, and storage serializes to JSON so the pickling change can't reach it. Adversarial probes — shared items, 3-level nesting, `del root` while only a deep child is held, `setdefault`/`update` replacement — all matched the documented behavioral notes.

A fresh-lineage skeptic pass (Codex) surfaced four behaviors: self-referential recursion (`root.append([root])`), diamond double-fire (`[item, [item]]`), `dict.pop('missing')` swallowing `KeyError`, and `__reduce__` dropping `on_change` on unpickle. I reran each against `main` and this branch — **identical on both** (98/98 handler calls, 2/2 fires, `None`+1-false-event on both) — so they are **pre-existing, not introduced here**, and the `__reduce__` handler-drop is the documented "pickling was never a supported path". None block this PR; the first three may deserve a separate low-priority issue, which I'll look at on their own.

On the three open questions: fully-weak `_parent` is the right call (roots are independently referenced; picklability is a free bonus); the `ObservableSet` `_unobserve`/per-item additions are genuinely dead code (observables are unhashable, never set members) — keep-for-symmetry is fine; and rejecting the fire-time containment scan is correct, since the cost belongs on the rarer removal path rather than every change event.
</details>

PR #6142, opened 2026-07-05
Prototype: reactive computed()/effect() on BindableProperty (RFC for #4758)

evnchn:
> ⚠️ **Draft / RFC — not a merge request.** This is a proof-of-concept to make the design discussion in #4758 concrete. I want feedback on the *shape*, not to land it as-is.

### Motivation

#4758 ("Managing state") keeps circling one question: should NiceGUI grow a first-class reactive layer, or is the ecosystem (reaktiv, ex4nicegui) the right home? Several people in that thread converged on the same concrete idea: *"NiceGUI's `BindableProperty` is already essentially a signal; what's missing is `Computed` and `Effect` on top of it."*

Talk is cheap, so here's what that would actually look like. The point is to turn "wouldn't it be nice" into something you can run and criticise.

The pain it targets is visible in our own [`examples/reaktiv`](https://github.com/zauberzeug/nicegui/tree/main/examples/reaktiv) — the pure-`bindable_dataclass` version needs **two** manual `bind_to` calls just to express `subtotal = price * quantity`:

```python
binding.bind_to(self, 'price',    self, 'subtotal', lambda price:    price * self.quantity)
binding.bind_to(self, 'quantity', self, 'subtotal', lambda quantity: self.price * quantity)
```

With derived state this becomes one line, and it composes:

```python
subtotal = binding.computed(lambda: order.price * order.quantity)
```

### Implementation

Two small hooks on the existing `BindableProperty`, and a thin reactive layer — all inside `binding.py`, ~110 lines:

- `__get__` → `_track_read`: while a computation runs, any bindable property it reads is recorded as a dependency (automatic, dynamic — re-captured every run, so branching code only subscribes to what it actually touched).
- `__set__` → `_notify_subscribers`: on write (next to the existing `_propagate`), wake the computations that read that property.
- `binding.computed(fn)` → a `Computed` whose result is itself a bindable `value`, so it plugs straight into `bind_*` and elements (`ui.label().bind_text_from(total, 'value')`).
- `binding.effect(fn)` → an `Effect` that re-runs on change; `dispose()` on both to unsubscribe.

Crucially there is **no polling** — it's read-tracking (`__get__`) plus write-notify (`__set__`) on the same write-detection the binding system already does, so `binding.active_links` stays empty (no 0.1s `refresh_loop` tick involved).

### Known limitations (this is a brick — these ARE the design questions)

I had a second, non-Claude model try to break it. The limitations below are real and left in on purpose — they're exactly the decisions a core version would have to make, and I'd rather name them than hide them:

- **Glitches on diamonds.** `C = f(A, B)` with `A, B` both derived from `S`: on `S` change, propagation is synchronous + depth-first, so an effect reading `C` can briefly see `A=new, B=old` and run twice. No topological scheduling.
- **No batching.** `obj.a = 1; obj.b = 2` re-runs a dependent effect twice; observers can see non-transactional intermediate state.
- **Sync callables only.** An `async def` body is never awaited, so its reads aren't captured. Not async-compatible.
- **Effects must be retained** (they're held weakly, like reaktiv's "assign to a variable" rule) and disposed on client disconnect — no automatic lifecycle binding yet.
- **`id(owner)`-keyed deps** share the existing binding code's id-reuse caveat.
- **Cleanup is partial:** `reset()` now also clears reactive subscriptions, but `binding.remove(objects)` does **not** yet drop `computed()`/`effect()` subscriptions for those objects — part of the same lifecycle work above.

A production version's real work is scheduling correctness (glitch-free, batched, lifecycle-bound), not the API surface — which is precisely why this is an RFC.

### Progress

- [x] `computed()` / `effect()` on `BindableProperty`, dynamic dependency tracking, `dispose()`
- [x] Tests (`tests/test_binding.py`): derive+update, computed-on-computed, effect fire/dispose, dynamic re-capture
- [x] `pre-commit`, `mypy ./nicegui`, `pylint ./nicegui` (10.00/10) green locally
- [ ] **Design decision — does this belong in core at all?** (that's the discussion)
- [ ] If yes: glitch-free scheduling, batching, client-bound effect lifecycle
- [ ] Docs / example

Not asking anyone to review this as production code — asking whether this *shape* is the direction, or whether reactive state stays in the ecosystem. cc @falkoschindler

### DIFF
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -275,6 +275,8 @@ class BindableProperty:
         self.name = name  # pylint: disable=attribute-defined-outside-init
 
     def __get__(self, owner: Any, _=None) -> Any:
+        if owner is not None:
+            _track_read(owner, (self.name,))  # PROTOTYPE: dependency capture for computed()/effect()
         return getattr(owner, '___' + self.name)
 
     def __set__(self, owner: Any, value: Any) -> None:
@@ -287,8 +289,115 @@ class BindableProperty:
         setattr(owner, '___' + self.name, value)
         bindable_properties[(id(owner), (self.name,))] = owner
         _propagate(owner, (self.name,))
+        _notify_subscribers(owner, (self.name,))  # PROTOTYPE: wake computed()/effect() that read this
         if value_changed and self._change_handler is not None:
             self._change_handler(owner, value)
+
+
+# =============================================================================
+# PROTOTYPE: reactive layer (computed / effect) on top of BindableProperty
+# -----------------------------------------------------------------------------
+# This is a proof-of-concept for zauberzeug/nicegui#4758, NOT a finished API.
+# It shows that `Computed` and `Effect` can be built directly on the existing
+# BindableProperty read/write hooks, with automatic dependency tracking, and no
+# 0.1 s refresh loop. Known open questions are documented in the discussion.
+# =============================================================================
+
+Dependency = tuple[int, tuple[str, ...]]
+
+_active_computation: ContextVar[_Computation | None] = ContextVar('active_computation', default=None)
+_subscribers: defaultdict[Dependency, weakref.WeakSet[_Computation]] = defaultdict(weakref.WeakSet)
+
+
+def _track_read(owner: Any, name: tuple[str, ...]) -> None:
+    """Record that the currently-running computation read this bindable property."""
+    computation = _active_computation.get()
+    if computation is not None:
+        computation._collecting.add((id(owner), name))  # pylint: disable=protected-access
+
+
+def _notify_subscribers(owner: Any, name: tuple[str, ...]) -> None:
+    """Re-run every computation that depends on the property that just changed."""
+    for computation in list(_subscribers.get((id(owner), name), ())):
+        computation._run()  # pylint: disable=protected-access
+
+
+class _Computation:
+    """Runs a function while recording which bindable properties it reads, then re-runs on change.
+
+    Dependencies are re-captured on every run, so branching reactive code subscribes only to the
+    properties it actually touched this time (dynamic dependency tracking).
+    """
+
+    def __init__(self, fn: Callable[[], Any]) -> None:
+        self._fn = fn
+        self._deps: set[Dependency] = set()
+        self._collecting: set[Dependency] = set()
+        self._run()
+
+    def _run(self) -> Any:
+        self._collecting = set()
+        token = _active_computation.set(self)
+        try:
+            result = self._fn()
+        finally:
+            _active_computation.reset(token)
+        for stale in self._deps - self._collecting:
+            _subscribers[stale].discard(self)
+        for fresh in self._collecting - self._deps:
+            _subscribers[fresh].add(self)
+        self._deps = self._collecting
+        return result
+
+    def dispose(self) -> None:
+        """Unsubscribe from all dependencies so this computation stops re-running."""
+        for dep in self._deps:
+            _subscribers[dep].discard(self)
+        self._deps = set()
+
+
+class Computed:
+    """A derived value that recomputes automatically when any bindable property it reads changes.
+
+    The result is exposed as a bindable ``value``, so it plugs straight into ``bind_*`` and elements::
+
+        total = binding.computed(lambda: state.price * state.qty)
+        ui.label().bind_text_from(total, 'value')
+    """
+    value = BindableProperty()
+
+    def __init__(self, fn: Callable[[], Any]) -> None:
+        self._fn = fn
+        self._computation = _Computation(lambda: setattr(self, 'value', self._fn()))
+
+    def dispose(self) -> None:
+        """Stop recomputing (dispose the underlying computation)."""
+        self._computation.dispose()
+
+
+class Effect:
+    """Runs a side-effecting function whenever any bindable property it reads changes.
+
+    Keep a reference to the returned ``Effect`` for as long as you need it; call ``dispose()``
+    (e.g. on client disconnect) to stop it and release its subscriptions.
+    """
+
+    def __init__(self, fn: Callable[[], Any]) -> None:
+        self._computation = _Computation(fn)
+
+    def dispose(self) -> None:
+        """Stop running and release all dependency subscriptions."""
+        self._computation.dispose()
+
+
+def computed(fn: Callable[[], T]) -> Computed:
+    """Create a `Computed` derived value from a function of other bindable properties (*PROTOTYPE*)."""
+    return Computed(fn)
+
+
+def effect(fn: Callable[[], Any]) -> Effect:
+    """Create an `Effect` that re-runs whenever the bindable properties it reads change (*PROTOTYPE*)."""
+    return Effect(fn)
 
 
 def _discard_binding_key_from_object_index(obj_id: int, binding_key: BindingKey) -> None:
@@ -341,6 +450,7 @@ def reset() -> None:
     bindable_properties.clear()
     active_links.clear()
     _binding_keys_by_object.clear()
+    _subscribers.clear()  # PROTOTYPE: also drop reactive computed()/effect() subscriptions
 
 
 @dataclass_transform()
--- a/nicegui/binding.py
+++ b/nicegui/binding.py
@@ -6,7 +6,8 @@ import dataclasses
 import time
 import weakref
 from collections import defaultdict
-from collections.abc import Callable, Iterable, Mapping, MutableMapping
+from collections.abc import Callable, Iterable, Iterator, Mapping, MutableMapping
+from contextlib import contextmanager
 from contextvars import ContextVar
 from typing import TYPE_CHECKING, Any, Literal, TypeVar
 
@@ -275,8 +276,21 @@ class BindableProperty:
         self.name = name  # pylint: disable=attribute-defined-outside-init
 
     def __get__(self, owner: Any, _=None) -> Any:
-        if owner is not None:
-            _track_read(owner, (self.name,))  # PROTOTYPE: dependency capture for computed()/effect()
+        # PROTOTYPE: reactive read hook. Skip while propagating (those reads are the binding system's
+        # own bookkeeping, not user dependencies) and when no computation is collecting.
+        if owner is not None and propagation_visited.get() is None:
+            computation = _active_computation.get()
+            if computation is not None:
+                name = (self.name,)
+                dep = (id(owner), name)
+                producer = _producers.get(dep)
+                if producer is not None and producer is not computation:
+                    _pull_readers.append((computation, dep))  # our fresh read of exactly this dep is in flight
+                    try:
+                        _ensure_current(producer)  # pull a dirty producer up-to-date before it is read
+                    finally:
+                        _pull_readers.pop()
+                _track_read(owner, name)  # dependency capture for computed()/effect()
         return getattr(owner, '___' + self.name)
 
     def __set__(self, owner: Any, value: Any) -> None:
@@ -288,8 +302,8 @@ class BindableProperty:
             return
         setattr(owner, '___' + self.name, value)
         bindable_properties[(id(owner), (self.name,))] = owner
-        _propagate(owner, (self.name,))
-        _notify_subscribers(owner, (self.name,))  # PROTOTYPE: wake computed()/effect() that read this
+        _notify_subscribers(owner, (self.name,))  # PROTOTYPE: settle reactive computed()/effect() first,
+        _propagate(owner, (self.name,))  # so legacy bindings that read a derived value propagate it fresh
         if value_changed and self._change_handler is not None:
             self._change_handler(owner, value)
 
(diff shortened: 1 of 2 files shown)


falkoschindler:
Thanks for turning the #4758 discussion into something concrete, @evnchn — this is exactly the right way to move that thread forward.

My take: **yes, this shape belongs in core — but as the completion of the binding module, not as a new reactive framework, and staged.**

**Why core at all:** The demand pattern is hard to ignore — #1003 (2023), #5315, #5418 and #4758 all ask for the same two primitives, and meanwhile users build them badly: fake bindings that trigger `refresh()`, N× `bind_value_to` fan-ins, even monkeypatching `BindableProperty` at runtime. `BindableProperty` already _is_ a signal; `computed()` isn't new scope but the missing half of an existing feature. And the ecosystem alternative in practice means bridging boilerplate (`Effect(lambda: label.set_text(...))`, `ReactiveProxy`) plus the GC footgun we already hit in #5812.

**The decisive argument is lifecycle, not API:** a third-party library can never know about clients, slots and disconnects. Core can — an `effect()` created inside a page function should auto-dispose with the client, like timers do. That's the one thing only NiceGUI can offer, and it's the one thing this prototype punts on. Without it, core reactivity has no advantage over `pip install reaktiv`.

**Where I'd reorder your limitation list:** diamond glitches and missing batching aren't follow-ups — for core they're entry criteria, because people will build load-bearing state graphs on this, and half-correct reactivity in core is worse than none. Likewise, sync-only is a bigger wart than it looks: half of real NiceGUI code is `async def`, and an effect that silently fails to track reads in async bodies will generate a steady stream of confused issues.

**On the implicit-vs-explicit question** (rodja's concern in #1003): I'd revisit rather than defer. The industry has since converged on auto-tracking (Vue, Solid, Angular, TC39 signals proposal), explicit dependency lists rot when the function body changes, and `BindableProperty`'s write-detection is already exactly this kind of magic — auto-tracked reads are its symmetric counterpart.

**Concrete proposal:**

1. `computed()` first — it subsumes the `binding.combine` idea from #5315, has a self-contained correctness story, and needs zero new element API since its result is a bindable `value`. Ship it glitch-free and batched from day one.
2. `effect()` second, only together with client-bound lifecycle — that's where the leaks live.

So: not mergeable as-is (as you say yourself), but the direction is right. Let's discuss scheduling and lifecycle design here rather than whether reactivity belongs in core at all.

---

### Reference: related discussions overview

| Date     | Thread                                                                                                                              | What it is                                                                                                                                                                                          | Relevance to #6142                                                                                                                                 |
| -------- | ----------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
| Jun 2023 | [#1003 Effect function like Vue](https://github.com/zauberzeug/nicegui/discussions/1003)                                            | CrystalWindSnake (later author of ex4nicegui) proposed auto-tracking `effect()`. rodja: auto-tracking is "very implicit and atypical for Python", countered with explicit `@ui.on_change(a, b, c)`. | Only prior core-team statement on the key design question: implicit read-tracking vs. explicit dependency list.                                    |
| Jul 2024 | [#3325 Too many bindings?](https://github.com/zauberzeug/nicegui/discussions/3325)                                                  | Concern about the 0.1 s refresh-loop cost of active links.                                                                                                                                          | The PR's "no polling, rides existing write-detection" claim addresses exactly this.                                                                |
| May 2025 | [#4758 Managing state](https://github.com/zauberzeug/nicegui/discussions/4758) (still active)                                       | Parent thread of the PR. reaktiv author (buiapp) + LeonMusCoden converge on "BindableProperty is essentially a Signal — add Computed/Effect".                                                       | Direct origin; buiapp replied 2026-07-06 that his company already monkeypatches `BindableProperty` with an `rx` namespace and likes the direction. |
| Oct 2025 | [#5315 Bind button enabled from multiple objects?](https://github.com/zauberzeug/nicegui/discussions/5315)                          | The "multiple sources" thread. falkoschindler sketched a `binding.combine` API (N sources → 1 target via `transform`); sudoShikhar offered to implement, stalled.                                   | `binding.computed(lambda: f(a.value, b.value, c.value))` subsumes `combine` — deciding #6142 answers #5315 too.                                    |
| Nov 2025 | [#5418 refresh on BindableProperty change](https://github.com/zauberzeug/nicegui/discussions/5418)                                  | Hand-rolled effect via fake binding to trigger `refreshable.refresh()`, with perf problems.                                                                                                         | Evidence of demand for a real `effect()`.                                                                                                          |
| Feb 2026 | [#5783](https://github.com/zauberzeug/nicegui/pull/5783) / [#5812](https://github.com/zauberzeug/nicegui/pull/5812) reaktiv example | Added `examples/reaktiv`; follow-up fixed Effects being garbage-collected.                                                                                                                          | Same "effects must be retained" weak-ref footgun the PR lists as a known limitation — real-world, not hypothetical.                                |
| May 2026 | [#6038 SPA state best practice](https://github.com/zauberzeug/nicegui/discussions/6038)                                             | State-management question, answers point to reaktiv.                                                                                                                                                | Further demand signal.                                                                                                                             |

evnchn:
Thanks @falkoschindler — this is exactly the reframe I was hoping for, and "completion of the binding module, not a new framework" is the right framing. I took your "these are entry criteria, not follow-ups" seriously and went and built it. **`computed()` is now glitch-free and batched, running, with tests.** So the discussion can move to the scheduling/lifecycle *design* on concrete code rather than in the abstract.

**TL;DR:**

- **Agreed: `computed()` first, glitch-free + batched from day one.** Done — pushed a scheduler that fixes the two things you flagged as entry criteria (diamond glitches, no batching). A diamond sink now recomputes **exactly once** on a single source change, never on a half-updated intermediate; `binding.batch()` coalesces multiple writes into one flush.
- **It stays "the missing half of an existing feature," not new API.** The computed result is still a bindable `value`, so `ui.label().bind_text_from(total, 'value')` works with **zero new element API** — verified end-to-end, not just in unit tests.
- **You're right that lifecycle is the decisive part, and I now have the concrete answer to point at:** ex4nicegui already does exactly what you described — a scope tree, one per `Client`, disposed on `client.on_disconnect`. That's the model for phase-2 `effect()`. I did **not** build it yet (kept it honest to your staging), but it's no longer hand-wavy.
- **One design decision I'd like your call on before I harden further:** the *flush model* — synchronous-per-write + explicit `batch()` (what's running now, keeps reads synchronous) vs. an event-loop auto-batch (Vue-style microtask flush). Details below.

<details>
<summary><b>What's running now — the scheduler, and how glitch-freedom + batching fall out</b></summary>

Design, after reading how the field actually does it (see next block): **eager + batched**, because NiceGUI's binding is *push* (a bound element only re-renders when the source's `.value` is written), so a computed has to eagerly publish its value for `bind_*` to work with no new element API. The scheduler:

- On a write, mark direct subscribers dirty and **enqueue** them — no inline recompute.
- A synchronous **flush** recomputes; dirtiness cascades *through* recompute (A recomputes → sets `A.value` → marks C), and a dependent that runs before one of its producers **pulls that producer up-to-date on read** (`__get__`) before using it. So ordering is emergent — no explicit height/topo table — and a diamond's C recomputes once, after both A and B have settled.
- `binding.batch()` defers the flush to the outermost exit; a lone write auto-flushes, so synchronous read-after-write is preserved (existing tests unchanged).
- Equality-cutoff is inherited free: `BindableProperty.__set__` already early-returns on an unchanged value, so an unchanged (or unchanged-result) recompute stops propagation.

Observable proof (runnable):

```python
from nicegui import binding

@binding.bindable_dataclass
class State:
    s: int = 1

binding.reset()
st = State(1)
a = binding.computed(lambda: st.s + 1)    # A
b = binding.computed(lambda: st.s + 10)   # B
observed = []
eff = binding.effect(lambda: observed.append((a.value, b.value)))
st.s = 2
# observed == [(2, 11), (3, 12)]   <- ONE consistent update, not [..., (3, 11), (3, 12)]
```

| Property | Before (original prototype) | After |
|---|---|---|
| Diamond sink on single write | runs twice, sees `(A=new, B=old)` | runs **once**, only settled `(A, B)` |
| `with batch(): a=1; b=2` | effect runs twice | effect runs **once** |
| Unchanged write | re-runs dependents | cut off (no re-run) |
| Unbounded feedback loop | — | clean `RuntimeError: cycle detected` |

Local gates green: `pre-commit`, `mypy ./nicegui`, `pylint ./nicegui` 10.00/10, and the full `tests/test_binding.py` (existing + new diamond / batch / equality / dynamic-deps / end-to-end `ui.label` binding).

</details>

<details>
<summary><b>Why eager-push and not a lazy-pull signals core — the design-space read</b></summary>

I read the actual source of reaktiv, ex4nicegui (+ its `signe` engine), the TC39 signals polyfill, Preact Signals, Vue 3 `@vue/reactivity`, SolidJS, and the "Reactively" taxonomy. Two things the whole field agrees on:

1. **Separate cheap *notify* from expensive *recompute*, and never let anything observe a computed until all its sources have settled.** Reached either by lazy-pull (Vue/Preact/reaktiv) or eager-ordered-push (Solid/MobX).
2. **"Update a UI element" is always modeled as an *effect* at a graph leaf, never a computed** (Vue `setupRenderEffect`, Solid `createRenderEffect`). A push-based UI binding isn't in tension with any of this — the element subscription *is* the effect; the scheduler just decides *when* it runs.

reaktiv (the `pip install reaktiv` you benchmarked against) is a direct Preact-Signals port: mark-dirty + lazy-pull-revalidate via version stamps, with native `async def` effects. The reason I went **eager** rather than copying that lazy-pull core: a purely lazy computed never *sets* its `.value`, so nothing pushes it to a bound label — you'd need an effect bridge per bound computed, i.e. a binding-layer change. Eager keeps your "result is a bindable value, zero new element API" constraint. Lazy-pull's one real win over eager here is skipping recompute of computeds nobody is reading — happy to revisit if you'd rather pay the binding-layer change for that.

The lifecycle answer is the clearest transfer: **ex4nicegui's `signe` binds an effect scope to each NiceGUI `Client` and disposes it in `@client.on_disconnect`** (`utils/clientScope.py`). That's the "core can know about clients, a library can't" advantage you named, already proven against NiceGUI's own client boundary. Its async story (fire-and-forget outside the graph) is the part *not* to copy — reaktiv's first-class async effect is the model there.

</details>

<details>
<summary><b>Maker/checker — two adversarial review rounds (Codex, non-Claude) found real bugs; all fixed</b></summary>

Same "have a different model try to break it" discipline as the original prototype, run twice — the second round attacking the first round's fixes.

**Round 1 (against the first scheduler):**

| Finding | Severity | Fix |
|---|---|---|
| `bind_to(..., forward=λ: computed.value)` propagated a stale computed (legacy `_propagate` ran before the reactive flush) | blocker | run the reactive flush **before** `_propagate` in `__set__` |
| effect that writes a dependency it reads didn't converge | major | see round 2 refinement |
| disposed/churned computations left empty subscriber keys | major | prune the key when its last subscriber leaves |

**Round 2 (attacking the round-1 fixes):**

| Finding | Severity | Fix |
|---|---|---|
| the "skip a running subscriber" guard was too broad — swallowed a write-back to a *different* dependency during a pull | blocker | scope it to the exact `(reader, dependency)` pair being pulled |
| incremental subscription + an exception mid-run left a subscription `dispose()` couldn't remove → zombie re-run | blocker | reconcile the dependency set in a `finally` so partial reads stay disposable |
| a cycle `RuntimeError` left the offending node queued, poisoning later unrelated writes | major | clear the pending queue on cycle abort |

Each fix re-verified with a runnable repro (scoped write-back settles; disposed-after-raise stops re-running; scheduler recovers after a cycle). One item deliberately left open: a computation garbage-collected **without** `dispose()` whose source object is *also* gone still leaves an empty subscriber entry — that's the same lifecycle bucket as the "cleanup is partial" limitation and is solved by the same scope-tree / weakref-finalize work as phase-2.

</details>

<details>
<summary><b>Still deferred (your staging), now with a named answer for each</b></summary>

- **`effect()` client-bound lifecycle** — not built; answer = ex4nicegui's per-`Client` scope tree disposed on disconnect.
- **Async effects** — sync only; answer = reaktiv's first-class `async def` tracking (not ex4nicegui's fire-and-forget).
- **`binding.remove(objects)`** still doesn't drop reactive subscriptions for those objects (pre-existing).
- **GC-without-dispose** residue — partial reap added (a later write to a live source drops its empty key); full reaping is the lifecycle work above.

</details>

**The one thing I'd like your call on**, since it shapes the API and the rest of the hardening:

1. **Flush model.** What's running is *synchronous per write + explicit `binding.batch()`* (reaktiv/Preact/Solid semantics — reads stay synchronous, `a=1; b=2` coalesces only inside `batch()`). The alternative is a *Vue-style event-loop auto-batch* (a whole event handler coalesces automatically, at the cost of async reads). Which fits NiceGUI better?
2. **Eager vs lazy core.** I went eager to keep zero-new-element-API. Are you OK with eager as the core model, or do you want the lazy-pull core (skips unread computeds) even though it needs a binding-layer effect bridge?
3. **`effect()` lifecycle.** Confirm the direction: an `effect()` created in a page function auto-disposes with the client via a scope tree (ex4nicegui's model), rather than requiring manual `dispose()`.

I've pushed it to this PR so the scheduler is in the diff (about 160 lines in `binding.py`, plus the tests). Not asking to merge — asking whether this shape and these three decisions are the direction.

falkoschindler:

The conflict is in `nicegui/binding.py`, against 8d291b1ee ("Optimize binding cleanup with reverse index", #6135), which reworked the same internals this prototype builds on. Noting it here for whenever the draft moves forward — no action needed while it stays a draft.

---

evnchn:

**Rebased** — conflict resolved, PR is `MERGEABLE` again (`4db9cc2f` → `601d247d`). Still a draft, and the open question is unchanged: does this shape belong in core at all?

<details><summary><b>What changed, and why it was cheap</b></summary>

#6135 reworked `remove()` and added the `_binding_keys_by_object` reverse index, but it never touched `BindableProperty.__get__` / `__set__` / `_propagate` — which is the only surface this prototype hooks into. So both conflicts were "both sides inserted at the same spot", not semantic:

| Conflict | Resolution |
|---|---|
| Block before `remove()` | Kept both — reactive layer, then #6135's `_discard_binding_key_from_object_index` helper adjacent to its caller |
| `reset()` | Kept both clear-lists: `_binding_keys_by_object.clear()` alongside the six reactive ones |

**Verified on the rebased branch (`601d247d`):**

- `tests/test_binding.py` — 37 passed, **0 skipped** (run with `-rs`, so nothing covering the change was silently skipped)
- `mypy ./nicegui` clean (245 files) · `pylint ./nicegui` 10.00/10 · ruff's 3 remaining errors reproduce on pristine `main`, so they are pre-existing and unrelated

**Real-boot smoke against the new internals**, since a green unit suite would not have caught a desync between the reactive subscriptions and the new reverse index:

```python

@binding.bindable_dataclass
class Order:
    price: float = 2.0
    quantity: int = 3

order = Order()
subtotal = binding.computed(lambda: order.price * order.quantity)
seen = []
eff = binding.effect(lambda: seen.append(subtotal.value))
label = ui.label().bind_text_from(subtotal, 'value')

order.price = 5.0
with binding.batch():
    order.price = 10.0
    order.quantity = 10

binding.remove([label])                     # exercises #6135's new reverse-index path
assert id(label) not in binding._binding_keys_by_object   # no stale index entry
order.price = 1.0
assert subtotal.value == 10.0 and seen[-1] == 10.0        # reactive graph survived remove()
```

```
initial       6.0 6.0 [6.0]
after price   15.0 15.0 [6.0, 15.0]
after batch   100.0 100.0 [6.0, 15.0, 100.0]
reverse index after remove([label]): 0 objs
after remove  10.0 10.0
OK - computed/effect/batch/dispose all intact on top of #6135
```

Note the batch of two writes produced **one** recomputation, not two — the glitch-free/batching behaviour from the earlier review round is intact after the rebase.

**Deliberately not fixed here:** `binding.remove()` still does not drop `computed()` / `effect()` subscriptions for the removed objects. That is already listed as a known limitation in the PR description — it is lifecycle work, not conflict fallout, and it belongs with the "if this goes into core" scheduling work rather than with a rebase.

</details>

PR #6144, opened 2026-07-06
Run `examples/pytests` in CI so bundled example tests can't silently rot

### LINKED ISSUE #6141: Run `examples/pytests` in CI so bundled example tests can't silently rot
### First Check

- [x] This is not a Q&A. I am sure something is wrong with NiceGUI's CI setup.

### Example Code

```python
# examples/pytests/tests/test_ci_blindspot_demo.py
def test_deliberately_broken() -> None:
    assert False, 'this example is broken on purpose'
```

```console
# A) CI's command — `uv run pytest` from repo root (pyproject: testpaths = ["tests"])
$ python -m pytest --collect-only -q
907 tests collected in 0.51s
$ python -m pytest --collect-only -q | grep -c test_ci_blindspot_demo
0                      # the broken example is NEVER collected; CI stays green

# B) run the example dir directly — the failure is real
$ cd examples/pytests && python -m pytest tests/test_ci_blindspot_demo.py -q
E       assert False
tests/test_ci_blindspot_demo.py:2: AssertionError
FAILED tests/test_ci_blindspot_demo.py::test_deliberately_broken - assert False
1 failed in 0.02s
```

(Ran on a current `main` checkout; the demo file was removed afterwards.)

### Description

The `examples/pytests` project ships runnable pytest files as teaching material, but **CI never executes them**, so a broken example stays green — worse than no example, because it teaches the wrong thing and erodes trust in the docs.

- **Expected:** a broken test under `examples/pytests` should turn CI red.
- **Actual:** it is never collected. `pyproject.toml` sets `testpaths = ["tests"]`, and no workflow references `examples/` — CI's `uv run pytest` (`.github/workflows/_test.yml`) collects only `tests/` (907 tests above, 0 of them the example).

This is not hypothetical: it surfaced in #6015, where the bundled scope example actually failed (assertions ran against the whole page instead of the scoped card) and nothing caught it. @falkoschindler asked for this follow-up issue there.

**Possible fix:** add a CI step that runs `pytest` inside `examples/pytests` (it has its own `conftest.py`/fixtures), or fold example test dirs into the main collection where fixtures allow. Worth deciding: just `examples/pytests` (has real tests today) vs. a broader sweep of any `examples/**/test_*.py`.

### NiceGUI Version

`main` (`3.11.1.post29.dev0+7d14357c`) — this is a CI-workflow gap, not version-specific.

### Python Version

`3.13.11` — also not version-dependent.

### Browser

Other (N/A — this is a CI/tooling issue; no browser involved).

### Operating System

Other (reproduced on macOS, but the gap is OS-independent — CI runs on Linux).

### Additional Context

Follow-up to #6015. The concrete acceptance test: add a deliberately-broken example, confirm CI goes red, then confirm it passes once fixed.

falkoschindler:
### Motivation

The `examples/pytests` project ships runnable pytest files as teaching material, but CI never executed them: `pyproject.toml` sets `testpaths = ["tests"]` and no workflow referenced `examples/`. A broken example stayed green — which is how the failing scope example in #6015 went unnoticed.

This gap has existed since the directory was created in #3121 (which added `testpaths = ["tests"]` in the same commit); it is not a regression from the uv migration.

Fixes #6141

### Implementation

Add one step to the `pytest` scope in `.github/workflows/_test.yml` that runs pytest inside `examples/pytests` using the repo's uv environment:

```yaml
- name: ${{ matrix.python }} / examples/pytests
  if: matrix.scope == 'pytest'
  run: uv run --project ../.. pytest
  working-directory: examples/pytests
```

Running from within the example directory picks up its own `pytest.ini` (`-p nicegui.testing.plugin`, `main_file = main.py`), exactly as a user of the example would run it. All required dev dependencies (pytest-selenium, pytest-asyncio) and Chrome are already available in the job. The failure-screenshot artifact path is extended to also cover `examples/pytests/screenshots/`, since the `Screen` fixture resolves its screenshot directory relative to the working directory.

Scope is deliberately narrow: only `examples/pytests` (the only example with real tests today), no generic sweep over `examples/**`.

Acceptance test from the issue: a temporary commit with a deliberately broken example test turned CI red ([red run](https://github.com/zauberzeug/nicegui/actions/runs/28769789223/job/85301057841)), and CI is green after removing it ([green run](https://github.com/zauberzeug/nicegui/actions/runs/28770680132/job/85303645517)).

### Progress

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -35,11 +35,17 @@ jobs:
       - name: ${{ matrix.python }} / ${{ matrix.scope }}
         if: matrix.scope == 'pytest'
         run: uv run pytest
+      - name: ${{ matrix.python }} / examples/pytests
+        if: matrix.scope == 'pytest'
+        run: uv run --project ../.. pytest
+        working-directory: examples/pytests
       - uses: actions/upload-artifact@v7
         if: matrix.scope == 'pytest' && failure()
         with:
           name: screenshots-${{ matrix.python }}
-          path: screenshots/**/*.failed.png
+          path: |
+            screenshots/**/*.failed.png
+            examples/pytests/screenshots/**/*.failed.png
           retention-days: 14
       - name: ${{ matrix.python }} / ${{ matrix.scope }}
         if: matrix.scope == 'startup'


evnchn (dismissed):

**LGTM — approving.**

The fix is correct and tightly scoped: it does exactly what the reported gap asked for, and I couldn't break it.

I reviewed the mechanics and then had a second, non-Claude lineage take an adversarial pass ("assume it's broken"). **No correctness bug survived** — the driver wiring, `uv --project` semantics, and the artifact path all check out (evidence folded below).

Two optional notes, both your call, neither blocks merge:

- **The `if:`-skip on a failed main pytest is right — leave it as-is.** If the main suite is red, that gets fixed first; no reason for the example run to pile onto the noise.
- **Worth linking the warning policy.** Mirror root's `filterwarnings` into `examples/pytests/pytest.ini` so the example also treats warnings as errors — cheap, and warning-level rot is exactly the quiet decay this PR exists to catch. Optional; tradeoff named below.

Only thing before merge: drop the intentional TEMP demo commit (1f8787e0) so CI can go green.

<details>
<summary><b>Verification — three suspicions, all cleared</b></summary>

Reviewed against `481b831b` (the actual fix).

| Suspicion | Verdict | Why |
|---|---|---|
| Example's `pytest.ini` lacks `--driver Chrome` → `screen` tests can't get a browser | **Refuted** | NiceGUI's `screen` fixture builds its own `webdriver.Chrome` via the `nicegui_driver` session fixture in `nicegui/testing/screen_plugin.py` — it does not use pytest-selenium's `--driver` at all, so the example config needs nothing extra. |
| `uv run --project ../.. pytest` from a subdir → wrong env or rootdir | **Correct** | `--project` selects the repo-root locked env (already `uv sync --locked`-installed earlier in the job); CWD stays `examples/pytests`, so pytest uses the example's own `pytest.ini` as rootdir and collects `tests/`. |
| Artifact `path:` off by a directory | **Correct** | `upload-artifact` ignores step `working-directory`; its paths are workspace-root-relative. `Screen.SCREENSHOT_DIR = (Path('screenshots') / str(os.getpid())).resolve()` is CWD-relative, so from `examples/pytests` shots land at `examples/pytests/screenshots/<pid>/` — matching the new glob exactly. |

</details>

<details>
<summary><b>Note 1 — the coupled step (leave as-is)</b></summary>

The new step's `if: matrix.scope == 'pytest'` has no status function, so GitHub implicitly wraps it in `success()` → effectively `success() && matrix.scope == 'pytest'`. If the earlier `uv run pytest` fails, the examples step is **skipped**.

That's the desirable behavior here (don't add noise once the suite is already red). Flagging only so it's a deliberate choice, not an accident. If you ever wanted examples to run regardless, it'd take `if: ${{ !cancelled() && matrix.scope == 'pytest' }}` — but I'd keep it as-is.

</details>

<details>
<summary><b>Note 2 — link the warning policy (optional, recommended)</b></summary>

The example runs under its own `pytest.ini`, which does **not** inherit root's `filterwarnings`. Root's block:

```toml
filterwarnings = [
  'error',
  "ignore:'asyncio\\.iscoroutinefunction' is deprecated and slated for removal:DeprecationWarning",
]
```

I checked whether this is end-user-appropriate or repo-internal: it's `error` plus a **single** ignore, and that ignore is a **framework-level** deprecation (`asyncio.iscoroutinefunction`, deprecated in 3.14, fired from inside the nicegui/starlette/socketio machinery — not repo test code). Zero repo-internal ignores. An end user running this example would hit that same warning and legitimately want it silenced too — so mirroring it is reasonable, not a leak of repo strictness.

Suggested `examples/pytests/pytest.ini`:

```toml
[pytest]
asyncio_mode = auto
main_file = main.py
addopts = -p nicegui.testing.plugin
filterwarnings =
    error
    ignore:'asyncio.iscoroutinefunction' is deprecated and slated for removal:DeprecationWarning
```

**Tradeoff, stated honestly:** `error` means a *new* upstream deprecation can turn the example red even when the example code is blameless. The main suite already accepts that burden — and "you find out the moment the framework deprecates something an example relies on" is arguably a feature for teaching material. Your call.

</details>

<details>
<summary><b>Second-lineage (Codex) adversarial pass — full verdict</b></summary>

Prompted to assume the diff was broken and find a real bug. Verbatim conclusion:

> No maintainer-actionable correctness bug found.
> 1. The examples step is skipped if the preceding root `uv run pytest` step fails. GitHub applies implicit `success()` to `if:` expressions without a status function… That is a diagnostic tradeoff, not a correctness bug…
> 2. `uv run --project ../.. pytest` from `examples/pytests` looks correct. `--project` discovers/uses the repo-root project/env, while command args/current working directory remain the current directory…
> 3. Artifact path looks correct. `upload-artifact` paths are relative to the action's current working directory; with no step `working-directory`, that is the workspace root… `examples/pytests/screenshots/**/*.failed.png` is the right root-relative path.
> 4. The warning-filter difference is real but not necessarily a bug…

Two independent lineages converged: clean fix, no blocker.

</details>

falkoschindler:
Thanks for the thorough review and the adversarial pass, @evnchn!

Regarding the two notes:

1. **`if:`-skip on a failed main pytest run:** Agreed — skipping the example step when the main suite is already red is what we want, so I'm leaving it as is.
2. **Mirroring `filterwarnings` into the example's `pytest.ini`:** I'd rather not. The example's `pytest.ini` is teaching material — users copy it as the starting point for their own projects, and `filterwarnings = error` would teach a repo-maintainer stance as a beginner default. It would also couple the example to root's ignore list, which churns with upstream deprecations and would need to be kept in sync manually. The rot this PR targets (example tests failing or never running) is covered without it.

The TEMP commit is dropped and the acceptance test from the issue is complete: the deliberately broken example test turned CI red ([red run](https://github.com/zauberzeug/nicegui/actions/runs/28769789223/job/85301057841)), and after removing it CI passes with the new step running the 10 example tests ([green run](https://github.com/zauberzeug/nicegui/actions/runs/28770680132/job/85303645517)).

Since the force-push dismissed your approval, could you re-approve?

evnchn (approved):
sure lgtm

PR #6146, opened 2026-07-07
Fix sporadic AssertionError when user storage is pruned during an in-flight request

### LINKED ISSUE #6145: `prune_user_storage()` can delete user storage mid-request, causing sporadic HTTP 500
**TL;DR:**

`app.storage.user` intermittently raises `AssertionError` (→ HTTP 500) when the `prune_user_storage()` 10s timer tick fires *during* an in-flight request whose session has no connected WebSocket client and whose user-storage is already older than 10s.

The storage that `RequestTrackingMiddleware` created for the request is pruned out from under the handler before it reads `app.storage.user`, tripping the assert at `nicegui/storage.py:137`.

Sporadic and self-healing (the next request re-creates the storage). A deterministic MRE and the file:line mechanism are folded below. This report is intentionally **neutral on how to fix it** — facts only.

Reported by @joko-zauberzeug, who hit this while an agent auto-tested an app running on NiceGUI 3.14.0; independently reproduced on `main`.

## Example Code

```python
from nicegui import app, ui

@ui.page('/')
def index():
    ui.label(app.storage.user.get('greeting', 'hello'))

ui.run(storage_secret='secret')
```

## Description

Reading `app.storage.user` inside a page/request handler intermittently returns **HTTP 500** with:

```
AssertionError: user storage for <session-id> should be created before accessing it
```
raised at `nicegui/storage.py:137`.

- **Expected:** a request that reads `app.storage.user` succeeds — the storage `RequestTrackingMiddleware` created for the session is still present when the handler reads it.
- **Actual:** if the `prune_user_storage()` tick fires while the request is in flight — for a session with no connected WebSocket client whose user-storage is older than 10s — the storage is removed between middleware-create and handler-access, and the `assert` fires.

**Who hits it:** sessions that have no connected WebSocket client *and* whose user-storage is already >10s old — **HTTP-only clients (e.g. AI agents / scripts), revisits from a closed tab, or any long-lived session polled over HTTP**. (A genuine first visit is normally safe — its storage starts at age 0 — unless the handler itself stays in flight for >10s before reading `app.storage.user`.)

**Reproduction note:** hard to hit with `curl`, because it carries no cookie across requests — every request is a *fresh* session with age ≈ 0, which is never prunable mid-request. A cookie-persisting client (real/headless browser, or any client reusing the session cookie) that revisits after >10s idle hits it reliably.

<details>
<summary><b>Deterministic MRE — real middleware + real prune, forces the race every run</b></summary>

Drives the real path: `SessionMiddleware` + `RequestTrackingMiddleware` + the real `app.storage.user` property + the real `prune_user_storage()` coroutine the app's 10s timer runs. Determinism device: hold one request in flight, then age the session's storage past 10s and fire one prune tick — forcing the ordering `[middleware creates storage] → [prune tick] → [handler reads app.storage.user]`.

```python
import asyncio, os, tempfile
os.environ['NICEGUI_STORAGE_PATH'] = tempfile.mkdtemp()

import httpx
from starlette.applications import Starlette
from starlette.middleware import Middleware
from starlette.responses import JSONResponse
from starlette.routing import Route

from nicegui import core
from nicegui.storage import RequestTrackingMiddleware, Storage
from nicegui.app.app import prune_user_storage

SECRET = 'just a test'
Storage.secret = SECRET
entered, release = asyncio.Event(), asyncio.Event()

async def handler(request):
    entered.set()
    await release.wait()  # hold the request in flight
    return JSONResponse({'greeting': core.app.storage.user.get('greeting', 'default')})

appl = Starlette(routes=[Route('/', handler)], middleware=[
    Middleware(SessionMiddleware, secret_key=SECRET),
    Middleware(RequestTrackingMiddleware),
])

async def main():
    async with httpx.AsyncClient(transport=httpx.ASGITransport(app=appl), base_url='http://t') as c:
        req = asyncio.ensure_future(c.get('/'))
        await entered.wait()
        sid = next(iter(core.app.storage._users))
        core.app.storage._users[sid].last_modified -= 20  # simulate >10s idle, no mutation
        await prune_user_storage()                         # the exact 10s-timer coroutine
        release.set()
        r = await req
    print(r.status_code, r.text)

asyncio.run(main())
```

**Actual output (on `main`):**
```
  ...
  File ".../nicegui/storage.py", line 137, in user
    assert session_id in self._users, f'user storage for {session_id} should be created before accessing it'
AssertionError: user storage for 0fc34cc7-209f-4d3a-97c0-7431da4d1ace should be created before accessing it
```

`_users` goes `1 → 0` while the request is in flight, then the handler's `app.storage.user` access asserts. This proves the assert fires when prune removes storage mid-request; it does **not** model any particular fix.
</details>

<details>
<summary><b>Mechanism (file:line)</b></summary>

- `nicegui/app/app.py:90` — `self.timer(10, prune_user_storage)` (10s tick).
- `nicegui/app/app.py:418-428` — `prune_user_storage()` removes any session **not in `Client.instances`** (no connected WS client) whose storage age `> 10.0`. It does not consider whether an HTTP request for that session is currently in flight.
- `nicegui/storage.py:37-38` — `RequestTrackingMiddleware` creates the user storage per request if missing.
- `nicegui/storage.py:137` — the `user` property `assert`s the storage exists.
- `nicegui/observables.py:24,35-36` — `last_modified` is set on creation and bumped via `_handle_change()` on mutation; a **read never bumps it**, so a session that is only read ages out after 10s.

This surfaced as a regression when `app.storage.user` changed from lazy-create to `assert` in #4074 (before that commit, a missing user storage was silently re-created on access). This issue takes **no position on the fix**.
</details>

<details>
<summary><b>User-side mitigation (not a proposed framework fix)</b></summary>

Affected apps can wrap `app.storage.user` reads in `try/except` with a fallback default. Noted for others who hit this in the wild — not a suggestion for how NiceGUI should resolve it.
</details>

## NiceGUI Version

Reported on `3.14.0`; independently reproduced on `main`. The relevant code is identical across both.

## Python Version

`3.12` (not Python-version-specific).

## Browser

Chrome (headless) — also reproducible with **no browser** (plain HTTP client reusing the session cookie).

## Operating System

macOS (OS-independent).

## First Check

- [x] This is not a Q&A. I am sure something is wrong with NiceGUI.
- [x] I used the GitHub search to find a similar issue and came up empty (the only nearby match, #5077, is a different assert from the same 10s timer).

evnchn:
### Motivation

Fixes #6145.

`app.storage.user` sporadically raises `AssertionError: user storage for <id> should be created before accessing it` (→ HTTP 500). `prune_user_storage()` runs on a 10 s timer and removes any session that has no connected WebSocket client and whose user-storage is older than 10 s. If that tick fires *while an HTTP request for the session is still being handled*, the storage `RequestTrackingMiddleware` created is removed before the handler reads `app.storage.user`, and the assert at `nicegui/storage.py:137` fires.

It hits sessions with no live WS client: **HTTP-only clients (e.g. AI agents / scripts), revisits from a closed tab, and long-lived sessions polled over HTTP**. It self-heals on the next request, so it shows up as intermittent 500s.

### Implementation

Track the number of in-flight HTTP requests per session id in `Storage._active_request_sessions` (a `collections.Counter`):

- `RequestTrackingMiddleware.dispatch` increments the counter for the session before handling and decrements it in a `finally` (so it is cleared even if the handler raises).
- `prune_user_storage()` skips any session id currently in that counter, so storage backing a request in flight is never pruned out from under it.

A `Counter` (not a `set`) so that two concurrent requests for the same session each pin it, and the session is only unpinned once the last one completes.

Why not lazy-re-create in the `user` property (the pre-#4074 behavior)? Since #4074, user-storage creation is **async** (`await ... initialize()`, needed for the Redis backend), while the `user` property is synchronous and cannot `await`. Creation therefore lives in the async middleware; keeping the storage alive for the duration of the request is the change that fits that constraint.

**Scope:** this covers requests that read `app.storage.user` during handler execution — i.e. all `@ui.page` / FastAPI handlers, which build their response eagerly. It does **not** extend the guard to a user-supplied `StreamingResponse` whose async generator reads `app.storage.user` lazily *after* the handler returns (Starlette streams that body after the middleware's `dispatch` returns); that path is outside the reported issue and unchanged by this PR.

<details>
<summary><b>Verification — works/broken differential + local gates</b></summary>

The added test `test_user_storage_survives_prune_during_request` drives the real `RequestTrackingMiddleware.dispatch` + real `prune_user_storage()` + real `app.storage.user`, firing a prune tick while the request is in flight:

- **Without the fix:** fails with `AssertionError` at `storage.py:137`.
- **With the fix:** passes (request reads its storage; the same idle aged session is still pruned once the request completes — control assertion).

Two more tests cover Codex-review edge cases: the in-flight marker is cleared when the handler raises (no leak), and two concurrent same-session requests both pin the session against pruning.

Local gates on the branch:
```
pre-commit run --files ...  → Passed
mypy ./nicegui              → Success: no issues found in 243 source files
pylint ./nicegui            → 10.00/10
pytest (3 new tests)        → 3 passed
```
</details>

### Progress

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -26,6 +26,8 @@ from ..storage import PersistentDict, Storage
 from .app_config import AppConfig
 from .range_response import get_range_response
 
+USER_STORAGE_PRUNE_INTERVAL = 10.0
+
 
 class State(Enum):
     STOPPED = 0
@@ -87,7 +89,7 @@ class App(FastAPI):
         self.timer(10, Slot.prune_stacks)
         self.timer(10, prune_tab_storage)
         if self.storage.secret is not None:
-            self.timer(10, prune_user_storage)
+            self.timer(USER_STORAGE_PRUNE_INTERVAL, prune_user_storage)
         self._state = State.STARTED
 
     async def stop(self) -> None:
@@ -425,7 +427,7 @@ async def prune_user_storage(*, force: bool = False) -> None:
     for session_id in list(user_storages):
         if session_id not in client_session_ids and session_id not in active_request_sessions:  # avoid pruning mid-request
             age = now - user_storages[session_id].last_modified
-            if force or age > 10.0:  # do not remove storages created by middleware and still wait for client
+            if force or age > USER_STORAGE_PRUNE_INTERVAL:  # do not remove storages created by middleware and still wait for client
                 storages_to_close.append(user_storages.pop(session_id))
     results = await asyncio.gather(*[storage.close() for storage in storages_to_close], return_exceptions=True)
     for result in results:
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -418,14 +418,16 @@ async def prune_tab_storage(*, force: bool = False) -> None:
 async def prune_user_storage(*, force: bool = False) -> None:
     """Remove user storage objects without a client session."""
     client_session_ids = {client.request.session['id'] for client in Client.instances.values()}
+    active_request_sessions = core.app.storage._active_request_sessions  # pylint: disable=protected-access
     storages_to_close: list[PersistentDict] = []
     now = time.time()
     user_storages = core.app.storage._users  # pylint: disable=protected-access
     for session_id in list(user_storages):
-        if session_id not in client_session_ids:
-            age = now - user_storages[session_id].last_modified
-            if force or age > 10.0:  # do not remove storages created by middleware and still wait for client
-                storages_to_close.append(user_storages.pop(session_id))
+        if session_id in client_session_ids or session_id in active_request_sessions:
+            continue  # keep storage for connected clients and in-flight requests (avoid pruning mid-request)
+        age = now - user_storages[session_id].last_modified
+        if force or age > 10.0:  # do not remove storages created by middleware and still wait for client
+            storages_to_close.append(user_storages.pop(session_id))
     results = await asyncio.gather(*[storage.close() for storage in storages_to_close], return_exceptions=True)
     for result in results:
         if isinstance(result, Exception):
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -423,11 +423,10 @@ async def prune_user_storage(*, force: bool = False) -> None:
     now = time.time()
     user_storages = core.app.storage._users  # pylint: disable=protected-access
     for session_id in list(user_storages):
-        if session_id in client_session_ids or session_id in active_request_sessions:
-            continue  # keep storage for connected clients and in-flight requests (avoid pruning mid-request)
-        age = now - user_storages[session_id].last_modified
-        if force or age > 10.0:  # do not remove storages created by middleware and still wait for client
-            storages_to_close.append(user_storages.pop(session_id))
+        if session_id not in client_session_ids and session_id not in active_request_sessions:  # avoid pruning mid-request
+            age = now - user_storages[session_id].last_modified
+            if force or age > 10.0:  # do not remove storages created by middleware and still wait for client
+                storages_to_close.append(user_storages.pop(session_id))
     results = await asyncio.gather(*[storage.close() for storage in storages_to_close], return_exceptions=True)
     for result in results:
         if isinstance(result, Exception):
--- a/nicegui/storage.py
+++ b/nicegui/storage.py
@@ -35,18 +35,18 @@ class RequestTrackingMiddleware(BaseHTTPMiddleware):
             request.session['id'] = str(uuid.uuid4())
         request.state.responded = False
         session_id = request.session['id']
-        storage = core.app.storage
-        storage._active_request_sessions[session_id] += 1  # pylint: disable=protected-access
+        active_request_sessions = core.app.storage._active_request_sessions  # pylint: disable=protected-access
+        active_request_sessions[session_id] += 1
         try:
-            if session_id not in storage._users:  # pylint: disable=protected-access
-                await storage._create_user_storage(session_id)  # pylint: disable=protected-access
+            if session_id not in core.app.storage._users:  # pylint: disable=protected-access
+                await core.app.storage._create_user_storage(session_id)  # pylint: disable=protected-access
             response = await call_next(request)
             request.state.responded = True
             return response
         finally:
-            storage._active_request_sessions[session_id] -= 1  # pylint: disable=protected-access
-            if storage._active_request_sessions[session_id] <= 0:  # pylint: disable=protected-access
-                del storage._active_request_sessions[session_id]  # pylint: disable=protected-access
+            active_request_sessions[session_id] -= 1
+            if active_request_sessions[session_id] <= 0:
+                del active_request_sessions[session_id]
 
 
 def set_storage_secret(storage_secret: str | None = None,
--- a/nicegui/storage.py
+++ b/nicegui/storage.py
@@ -1,6 +1,7 @@
 import contextvars
 import os
 import uuid
+from collections import Counter
 from datetime import timedelta
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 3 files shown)


evnchn:
Heads-up @falkoschindler — putting this on your radar since it came in via Slack (joko’s report). It fixes the `app.storage.user` prune race in #6145: `prune_user_storage()` can remove a session’s user storage mid-request for clients with no connected WebSocket (HTTP-only / AI agents, closed-tab revisits), tripping the assert at `storage.py:137` → sporadic 500.

Verified with an MRE + a regression test that fails without the fix and passes with it; Codex-reviewed; all draft-eligible CI green (full browser suite still gated behind draft). **Deliberately kept as a draft** — no rush to review or merge, just making sure you know it exists.

falkoschindler (approved):
Thanks for tracking this down and for the clean fix, @evnchn! The approach is exactly right: the middleware's increment and the prune's check-and-pop both run synchronously between await points on the same loop, so a pinned session can never be popped mid-request, the `finally` guarantees cleanup even when the handler raises, and the `Counter` correctly keeps a session pinned until the last of several overlapping requests completes.

While reviewing, I pushed two commits directly to the branch — please have a look whether you agree:

- 6c9bc4778 restores the original control flow in `prune_user_storage()` (nested condition instead of a `continue` guard, which minimizes the diff against main) and binds the counter once in `RequestTrackingMiddleware.dispatch`, so the increment and the `finally`-decrement visibly share the same object and three pylint disables could go.
- dfe367e30 replaces the three white-box tests with a single end-to-end Screen test: the hardcoded 10 s is extracted into `USER_STORAGE_PRUNE_INTERVAL` so the test can shrink it to 0.1 s, and a real FastAPI endpoint stays in flight over several prune ticks and must still read `app.storage.user`. It reproduces the original HTTP 500 `AssertionError` when the in-flight guard is removed and passes with it.

Two questions:

1. This PR is still in draft mode — anything missing from your side, or can we mark it ready for review?
2. Are you fine with dropping the two edge-case tests (marker cleared when the handler raises, concurrent same-session requests) in favor of the one readable end-to-end test? They pinned implementation details (fake requests, private counters), which tend to get rewritten together with the implementation and then don't guard much. If you agree, the PR description needs a small update to match (one test instead of three) — feel free to adjust it, otherwise I'm happy to.

evnchn:
@falkoschindler i have a shower though: am sure we have other things which track which requests are in flight? Maybe we can reuse those. Unless I mis-remembered...

After that I have no issues, including the simplified test and the other commits.

evnchn:
@falkoschindler — chased my own shower-thought before handing it to you.

**TL;DR:**
**No, there's nothing existing to reuse** — the `Counter` is the minimal form of the very thing I was half-remembering. Closing the question; undrafting, good to go.

The three things NiceGUI already has that touch "a request" each miss for a different structural reason:

- `request_contextvar` is **context-local** — the 10 s prune tick runs in a separate task, so it can't see it (that's the whole point of a `ContextVar`).
- `request.state.responded` genuinely **is** an in-flight flag, but it lives on each `Request` object and nothing aggregates it by session.
- `Client.instances` (which `prune_user_storage()` already consults) only covers page/WS clients — **plain FastAPI handlers create no `Client`**, and reading `app.storage.user` from one of those is exactly the HTTP-only case the fix targets.

So `_active_request_sessions` is that idea aggregated by session: `request.state.responded` is the same "in-flight" notion at per-request granularity; the `Counter` is it rolled up to a per-session registry that the prune task can actually see.

<details>
<summary><b>The three existing mechanisms, and why each can't be the reuse target</b></summary>

| Existing mechanism | What it is | Why `prune_user_storage()` can't reuse it |
|---|---|---|
| `request_contextvar` (`storage.py:22`) | `ContextVar[Request]`, set per request in the middleware | Context-local by design; the prune tick runs in a different task → `.get()` returns `None`. A `ContextVar` can never be a cross-context "who's in flight" registry. |
| `request.state.responded` (`storage.py:36/47/125`) | Per-request bool, flipped `False`→`True` around `call_next` | A real in-flight marker — but it's per-`Request`, and there's no registry of live requests to scan. Only read inside `storage.py`; nothing aggregates it by session. |
| `Client.instances` (`app.py:420`) | Page-client map; prune already derives `client_session_ids` from it for the WS-connected case | Only covers clients with a page/WS lifecycle. `@app.get`/FastAPI handlers create no `Client`, so this structurally misses the HTTP-only path the fix targets. |

Two hard constraints force a dedicated structure rather than reuse:

1. **Must be visible to the prune task** → rules out `request_contextvar` (context-local) and `request.state` (per-object).
2. **Must cover every request type, not just page-clients** → rules out `Client.instances`.

`_active_request_sessions` (a `Counter`, so two concurrent same-session requests each pin it and it only unpins on the last completion) is the smallest structure satisfying both.
</details>

Undrafting — good to go on my end; the simplified test and the other commits all still stand.

PR #6148, opened 2026-07-08
Fix page_title/add_*_html/add_css lost when set after the response is built

### LINKED ISSUE #6147: Page title is not updated after a reload using await inside an async subpages function
### First Check

### Example Code

```python
import asyncio

from fastapi import FastAPI
from nicegui import ui, context

def root():
    ui.sub_pages({"/": index, "/page": page })

def index():
    ui.link("page", target="/page")

async def page():
    ui.page_title("before")
    ui.label(f"before await: {context.client.has_socket_connection}")
    await asyncio.sleep(0) # change the delay to > 0.2 and it will work
    ui.label(f"after await: {context.client.has_socket_connection}")
    ui.page_title("after")

ui.run(root)
```

### Description

Execute the code, click on "page" link, the page title is "after". Then reload the page with the browser, the page_title stay "before", it should update it to "after".

It looks like when calling ui.page_title("something") after an "fast" await call in a subpage router async function, the page title is not updated.

If the async call is "slow" as commented in the code snippet, i believe the async task which run the javascript is executed and the page title is updated to "after", but if the async call is "fast", the code is not executed and the page title stay "before".

by looking at the ui.page_title function, it looks like the value of context.client.has_socket_connection is False, therefore the javascript is not executed. I don't really know where this bug come from, i believe it may be linked to the way asyncio is working.

### NiceGUI Version

3.14.0

### Python Version

3.13.7

### Browser

Firefox

### Operating System

Windows

### Additional Context

This is issue is not very straightforward, i can reproduce it on two different system, but It can be a "not a bug" thing...

Thanks to take a look at it !

evnchn:

### Motivation

Closes #6147.

`ui.page_title` set after a fast `await` inside an async `ui.sub_pages` builder is silently lost on a full page load. The reporter's own diagnosis and @falkoschindler's confirmation pinpoint a **dead zone**: `ui.sub_pages` runs async builders as background tasks, so on a fresh load the builder resumes *after* the HTTP response (with the initial title) is already emitted, but *before* the websocket connects. In that window `client.has_socket_connection` is `False`, so `ui.page_title` skips its `document.title = …` JavaScript and the update is dropped — neither in the HTML nor pushed via JS.

Widening the scope (see the [issue discussion](https://github.com/zauberzeug/nicegui/issues/6147)): the **same guard exists at two more call sites** and loses writes in the identical dead zone:

- `ui.add_head_html` / `ui.add_body_html` (`functions/html.py`)
- `ui.add_css` / `add_scss` / `add_sass` (`functions/style.py`, via `_add_javascript`)

### Implementation

The `if client.has_socket_connection:` guard was only ever a *proxy* for the real question — **"has the initial HTML already been emitted?"** — which is what decides whether a value rides the HTML or must be pushed via JavaScript. `run_javascript` has enqueued into the outbox (buffered until the socket connects) since NiceGUI 3.0, so the guard is no longer needed to defer JS; it just answers the wrong question in the dead zone.

This PR adds a small `Client._response_built` flag (set once at the top of `build_response`) and uses it in place of the guard at all four call sites:

- **before render** → write the Python state (feeds the initial HTML), skip the JS
- **after render** (dead zone *or* connected) → run the JS; the outbox buffers it until the socket connects

**Why not simply drop the guard?** It is safe for `page_title` (whose JS is idempotent: `document.title = …`), but `add_*_html`/`add_css` are **additive** (`insertAdjacentHTML` / `addStyle`) and their payload is *already baked into the initial HTML* via `_head_html`/`_body_html`. Always-enqueuing would inject a **second copy on every normal load**. The `_response_built` predicate avoids that while still covering the dead zone.

<details><summary><b>Empirical evidence (measured in the DOM)</b></summary>

Dead-zone loss reproduced with an async sub-page + fast `await` (`/fast` = dead zone, `/slow` = control), on `3.13.0.post24.dev0`:

```
/slow  [post-connect control] add_body_html marker present: True  | add_css --mre: '42'
/fast  [DEAD ZONE           ] add_body_html marker present: False | add_css --mre: ''
```

Why dropping the guard is wrong for the additive functions — `add_body_html` markers counted in the DOM across three source variants:

| source | build-time (normal load) | dead zone |
|---|---|---|
| unpatched (guard intact) | `1` ✓ | `0` ✗ lost |
| naive drop-the-guard | `2` ✗ double-injected | `1` ✓ |
| `_response_built` (this PR) | `1` ✓ | `1` ✓ |

Auto-index shared client verified: `page_title` + `add_body_html` + `add_css` at module scope render the title, exactly one marker (no double), and the CSS — `_response_built` flips on the first response and is `False` during pre-GET setup, so the payload rides the HTML.

</details>

<details><summary><b>Tests</b></summary>

Three regression tests (proven red on `main`, green with this PR), each driving the dead-zone path (async sub-page → `await` → operation → fresh load):

- `test_page_title.py::test_page_title_after_await_in_async_sub_page`
- `test_add_html.py::test_add_body_html_after_await_in_async_sub_page` (also asserts a single injection, guarding against the double-inject regression)
- `test_add_html.py::test_add_css_after_await_in_async_sub_page`

`ruff` / `mypy` / `pylint` clean; `test_page_title.py` + `test_add_html.py` full files pass (19).

Deliberately deferred: a reconnect-window test (server-side `page_title`/`add_css` during a disconnect, arriving after reconnect). That path is pre-existing outbox buffering, not new logic here — happy to add it if you'd like.

</details>

### Progress

- [x] No breaking changes to the public API. (`_response_built` is a private attribute.)

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -114,6 +114,7 @@ class Client:
 
         self.title: str | None = None
         self.status_code: int = 200
+        self._response_built = False
 
         self._head_html = ''
         self._body_html = ''
@@ -174,6 +175,8 @@ class Client:
 
     def build_response(self, request: Request, status_code: int = 200) -> Response:
         """Build a FastAPI response for the client."""
+        # After this point the initial HTML (incl. title/head/body) is emitted; later changes must be pushed via JS.
+        self._response_built = True
         if self.page.resolve_markdown() and _did_user_request_markdown(request):
             parts = []
             if title := self.resolve_title():
--- a/nicegui/functions/html.py
+++ b/nicegui/functions/html.py
@@ -12,7 +12,7 @@ def add_head_html(code: str, *, shared: bool = False) -> None:
         Client.shared_head_html += code + '\n'
     else:
         client = context.client
-        if client.has_socket_connection:
+        if client._response_built:  # pylint: disable=protected-access
             client.run_javascript(f'document.head.insertAdjacentHTML("beforeend", {code!r});')
         client._head_html += code + '\n'  # pylint: disable=protected-access
 
@@ -27,6 +27,6 @@ def add_body_html(code: str, *, shared: bool = False) -> None:
         Client.shared_body_html += code + '\n'
     else:
         client = context.client
-        if client.has_socket_connection:
+        if client._response_built:  # pylint: disable=protected-access
             client.run_javascript(f'document.querySelector("#app").insertAdjacentHTML("beforebegin", {code!r});')
         client._body_html += code + '\n'  # pylint: disable=protected-access
--- a/nicegui/functions/page_title.py
+++ b/nicegui/functions/page_title.py
@@ -13,5 +13,5 @@ def page_title(title: str) -> None:
     client.title = title
     if core.app.native.main_window:
         core.app.native.main_window.set_title(title)
-    if client.has_socket_connection:
+    if client._response_built:  # pylint: disable=protected-access
         client.run_javascript(f'document.title = {json.dumps(title)}')
--- a/nicegui/functions/style.py
+++ b/nicegui/functions/style.py
@@ -66,5 +66,5 @@ def _add_javascript(code: str, *, shared: bool = False) -> None:
     else:
         client = context.client
         client._head_html += script_html + '\n'
-    if client is not None and client.has_socket_connection:  # no need to run JavaScript if there is no client yet
+    if client is not None and client._response_built:  # pylint: disable=protected-access
         client.run_javascript(code)
--- a/tests/test_add_html.py
+++ b/tests/test_add_html.py
@@ -87,37 +87,23 @@ def test_sass(screen: Screen, shared: bool, delayed: bool):
     assert label.value_of_css_property('color') == 'rgba(128, 0, 128, 1)'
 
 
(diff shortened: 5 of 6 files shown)


falkoschindler (approved):
Excellent analysis and a very clean fix — `_response_built` answers exactly the question the old `has_socket_connection` guard was only a proxy for. I verified the load-bearing claims independently.

I pushed one commit tightening the tests: merged the two `add_html` tests, added a build-time `add_body_html` so that naively dropping the guard fails with a double injection (the previous dead-zone-only markers couldn't catch that case), and replaced the fixed waits with auto-waiting `should_contain`/`wait_for` assertions.

Thanks for the thorough investigation, @evnchn! Let's merge! 🚀

falkoschindler (approved):
Excellent analysis and a very clean fix — `_response_built` answers exactly the question the old `has_socket_connection` guard was only a proxy for. I verified the load-bearing claims independently:

I pushed one commit tightening the tests: merged the two `add_html` tests, added a build-time `add_body_html` so that naively dropping the guard fails with a double injection (the previous dead-zone-only markers couldn't catch that case), and replaced the fixed waits with auto-waiting `should_contain`/`wait_for` assertions.

Thanks for the thorough investigation, @evnchn! Let's merge! 🚀

PR #6149, opened 2026-07-08
Close streamed media file handle synchronously on teardown

evnchn:
> **Draft — maintainer's call.** Opening this as a draft so you can decide whether the flake is worth carrying a fix for. Happy to close it if you'd rather leave the streaming code untouched; no attachment to it landing.

### Motivation

The `full-test / pytest` job sporadically fails at **teardown** — not in any test assertion — with:

```
ERROR at teardown of test_auto_serving_file_from_video_source
pytest.PytestUnraisableExceptionWarning: Exception ignored in:
  <_io.FileIO name='.../tests/media/test.mp4' mode='rb' closefd=True>
```

All tests pass (`961 passed, 2 xfailed, 1 error`); the lone "error" is a leaked file handle surfaced by pytest's `unraisableexception` plugin at GC time. Because it depends on **when** the GC runs, it is intermittent and has ejected unrelated PRs from the merge queue.

### Implementation

Root cause: `get_range_response` streams via a `StreamingResponse` whose body is an async generator using `async with aiofiles.open(...)`. When a client disconnects mid-stream (e.g. a browser playing a video), **Starlette does not `aclose()` the body iterator** — the generator is finalized only by the garbage collector. If that GC runs once the event loop is gone (test teardown, server shutdown), aiofiles' async `__aexit__` cannot run its `run_in_executor` close:

```
File ".../aiofiles/base.py", line 76, in __aexit__
    await get_running_loop().run_in_executor(...)
RuntimeError: no running event loop
```

…so the underlying file handle leaks to `__del__` → `ResourceWarning`.

Fix: open the file synchronously and read it via `asyncio.to_thread` (keeping the streaming reads non-blocking, the intent of #5700), then close it in a `finally:` with a **synchronous** `data.close()`. A sync close runs correctly during `GeneratorExit` at GC even when no event loop is available, so the handle is always released. Opening synchronously (rather than off-thread) also removes any await before the `try`, so a task cancelled mid-open cannot orphan the handle.

### Progress

<details>
<summary>Empirical before/after (deterministic repro)</summary>

The leak needs the disconnect to coincide with loop teardown. A minimal repro abandons a range stream after one chunk, then tears down the loop before GC:

```
OLD source (aiofiles): LEAK -> unclosed file <_io.BufferedReader name='.../test.mp4'>
  (traceback: aiofiles base.py __aexit__ -> RuntimeError: no running event loop)
NEW source (fixed):    NO LEAK
```

The added regression test `test_streamed_media_file_handle_is_released_on_teardown` encodes exactly this: it captures the opened file object and deterministically asserts `f.closed` after loop teardown + GC (no warning-sniffing). Verified to **fail on the old code** (via the `f.closed` assertion) and **pass on the fix**.
</details>

<details>
<summary>Local gate results</summary>

- `ruff check` — All checks passed
- `mypy ./nicegui` — Success, no issues
- `pylint ./nicegui/app/range_response.py` — 10.00/10
- `pytest tests/test_serving_files.py -W error::pytest.PytestUnraisableExceptionWarning` — 14 passed (incl. 5× stress of the video tests)
</details>

<details>
<summary>Second-model review (Codex / gpt-5.5, adversarial)</summary>

A first adversarial pass raised three points, all addressed here:

1. **Test mutated the global event loop** (`set_event_loop`, leaving a closed loop current) → removed entirely; the test drives via `run_until_complete` (which `asyncio.to_thread` picks up via `get_running_loop`) and closes its private loop in `finally`.
2. **Cancel during an off-thread open could orphan the handle** → the file is now opened **synchronously**, so there is no `await` before the `try`.
3. **The test only sniffed `ResourceWarning`** (the real signal is a teardown-flushed `PytestUnraisableExceptionWarning`, making the assertion a possible no-op) → rewritten to capture the file object and deterministically assert `f.closed`.

Re-review verdict: *"1. RESOLVED … 2. RESOLVED … 3. RESOLVED … New blockers: none found."*
</details>

### DIFF
--- a/nicegui/app/range_response.py
+++ b/nicegui/app/range_response.py
@@ -65,10 +65,9 @@ def get_range_response(file: Path, request: Request, chunk_size: int) -> Respons
                 remaining_bytes -= len(chunk)
         finally:
             data.close()
-    data = open(file, 'rb')  # pylint: disable=consider-using-with  # closed via the `finally` above or the finalizer below
+    data = open(file, 'rb')  # pylint: disable=consider-using-with
     generator = content_reader(data, start, end)
-    # backstop: Starlette may abandon the iterator without aclose() (client disconnect)
-    weakref.finalize(generator, data.close)
+    weakref.finalize(generator, data.close)  # close if the iterator is abandoned without aclose()
     return StreamingResponse(
         generator,
         media_type=media_type,
--- a/nicegui/app/range_response.py
+++ b/nicegui/app/range_response.py
@@ -1,10 +1,10 @@
+import asyncio
 import hashlib
 import mimetypes
 from collections.abc import AsyncGenerator
 from datetime import datetime, timezone
 from pathlib import Path
 
-import aiofiles
 from fastapi import Request
 from fastapi.responses import Response, StreamingResponse
 
@@ -52,15 +52,18 @@ def get_range_response(file: Path, request: Request, chunk_size: int) -> Respons
     })
 
     async def content_reader(file: Path, start: int, end: int) -> AsyncGenerator[bytes, None]:
-        async with aiofiles.open(file, 'rb') as data:
-            await data.seek(start)
+        data = open(file, 'rb')  # pylint: disable=consider-using-with  # closed synchronously in the `finally` below
+        try:
+            await asyncio.to_thread(data.seek, start)
             remaining_bytes = end - start + 1
             while remaining_bytes > 0:
-                chunk = await data.read(min(chunk_size, remaining_bytes))
+                chunk = await asyncio.to_thread(data.read, min(chunk_size, remaining_bytes))
                 if not chunk:
                     break
                 yield chunk
                 remaining_bytes -= len(chunk)
+        finally:
+            data.close()
     return StreamingResponse(
         content_reader(file, start, end),
         media_type=media_type,
--- a/nicegui/app/range_response.py
+++ b/nicegui/app/range_response.py
@@ -1,9 +1,11 @@
 import asyncio
 import hashlib
 import mimetypes
+import weakref
 from collections.abc import AsyncGenerator
 from datetime import datetime, timezone
 from pathlib import Path
+from typing import BinaryIO
 
 from fastapi import Request
 from fastapi.responses import Response, StreamingResponse
@@ -51,8 +53,7 @@ def get_range_response(file: Path, request: Request, chunk_size: int) -> Respons
         'Accept-Ranges': 'bytes',
     })
 
-    async def content_reader(file: Path, start: int, end: int) -> AsyncGenerator[bytes, None]:
-        data = open(file, 'rb')  # pylint: disable=consider-using-with  # closed synchronously in the `finally` below
+    async def content_reader(data: BinaryIO, start: int, end: int) -> AsyncGenerator[bytes, None]:
         try:
             await asyncio.to_thread(data.seek, start)
             remaining_bytes = end - start + 1
@@ -64,8 +65,12 @@ def get_range_response(file: Path, request: Request, chunk_size: int) -> Respons
                 remaining_bytes -= len(chunk)
         finally:
             data.close()
+    data = open(file, 'rb')  # pylint: disable=consider-using-with  # closed via the `finally` above or the finalizer below
+    generator = content_reader(data, start, end)
+    # backstop: Starlette may abandon the iterator without aclose() (client disconnect)
+    weakref.finalize(generator, data.close)
     return StreamingResponse(
-        content_reader(file, start, end),
+        generator,
         media_type=media_type,
         headers=headers,
         status_code=status_code,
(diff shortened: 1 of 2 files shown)


falkoschindler (changes requested):
Thanks for the impressively thorough analysis and the deterministic repro — and for the honest framing. I did want to carry a fix for this flake, so I had Claude (Fable 5) dig into the finalization mechanics before merging. Full disclosure: I'm not familiar enough with the CPython/asyncio finalization details to vouch for every step myself, so the findings below are mainly Claude's point of view, backed by experiments it ran against this branch. They looked convincing enough to me that I'd rather hold the merge until they're addressed or refuted.

1. **The `finally:` never runs on the production finalization path, so the CI flake most likely persists** (major) — CPython captures the asyncio finalizer hook on an async generator at its *first iteration*. Under uvicorn/Starlette that always happens inside the running loop, so the hook is always captured. When GC later finalizes the abandoned generator, `_PyGen_Finalize` calls the loop's `_asyncgen_finalizer_hook` *instead of* throwing `GeneratorExit` — and on a closed loop that hook is a silent no-op (`if not self.is_closed(): ...`). The finalize slot is consumed, the frame is simply cleared, and no `finally:` runs, sync or not. The handle is then closed only by `BufferedReader.__del__`, which raises the very same `ResourceWarning`. Claude verified this against this branch: first-iterating the new `content_reader` *inside* a loop, closing the loop, then GC'ing leaves the file open, and dropping the handle produces `Exception ignored in: <_io.FileIO name='….mp4' mode='rb' closefd=True>` — byte-for-byte the CI signature. The regression test passes only because `generator.__anext__()` is evaluated *outside* the running loop (the asend object is created at argument-evaluation time), so no finalizer hook is captured and GC falls back to throwing `GeneratorExit` — a path Starlette/uvicorn can never take. One-line falsification (fails on this branch):

    ```python
    async def first_chunk():
        return await generator.__anext__()
    loop.run_until_complete(first_chunk())  # instead of run_until_complete(generator.__anext__())
    ```

    Tying the cleanup to the generator *object* rather than its frame covers all paths:

    ```python
    data = open(file, 'rb')
    generator = content_reader(data, start, end)  # takes the handle, keeps the sync `finally`
    weakref.finalize(generator, data.close)
    return StreamingResponse(generator, ...)
    ```

    `weakref.finalize` callbacks fire during the generator's deallocation regardless of finalizer hooks or loop state (and at interpreter exit at the latest); `close()` is idempotent, so the fast path (`aclose`/`GeneratorExit` → `finally`) and the backstop coexist. Claude verified this variant closes the handle in all three scenarios (hook + closed loop, hook + running loop, full read) with no unraisables. Holding `data.close` keeps the handle alive exactly as long as the generator — same lifetime the frame already imposed.

2. **Strengthen the regression test to cover the production path** (minor) — with the backstop in place, the test should first-iterate *inside* the running loop (snippet above), since that is the only finalization path production takes; ideally assert both variants. The white-box approach itself is justified here — a Screen-level test can't deterministically control GC-vs-loop-teardown ordering — and the rest of the test is well built (captures the handle instead of sniffing warnings, no global loop mutation).

Everything else looks good: the synchronous `open()` is consistent with the `file.stat()` calls already in `get_range_response` and genuinely fixes the mid-open cancellation orphan, `asyncio.to_thread` per chunk is equivalent to what aiofiles did internally, and there's no public API change. With the two points above addressed I'm happy to take this out of draft and merge — it would then plausibly kill the flake for real.

evnchn:
You're right, and so is your reviewer — I reproduced it. The `finally:` doesn't run on the production path: the async generator's first iteration inside the loop captures the asyncio finalizer hook, so a later GC on the closed loop clears the frame instead of throwing `GeneratorExit`, and `finally:` never fires. The old test only passed because it iterated *outside* the loop.

Fixed with the `weakref.finalize(generator, data.close)` backstop you suggested (fires on generator dealloc regardless of loop/hook state; `close()` idempotent so it coexists with `finally:`). The regression test now first-iterates inside the loop — it fails on the old commit with the exact CI signature and passes on the fix. ruff/mypy/pylint/pytest green; a second-model (Codex) adversarial pass found no blockers.

Pushed as 220424d6 and out of draft.

falkoschindler (approved):
Thanks for the quick and thorough turnaround — both points from my previous review are resolved:

1. The `weakref.finalize(generator, data.close)` backstop ties the cleanup to the generator *object*, so it also covers the production finalization path where the captured asyncio hook silently no-ops on a closed loop.
2. The regression test now first-iterates inside the running loop. Re-verified locally: it fails on the old code with the exact CI signature and passes on this branch.

I pushed two small commits on top: 1a4f9cd0 trims the inline comments, and 060e0bdf makes the initial `seek` synchronous — an absolute seek on a fresh handle is just an `lseek`, a pure offset update in the kernel with no disk I/O, so the thread hop wasn't buying anything (the actual reads stay off the loop).

Approving — thanks for hunting down this flake and for the deterministic repro!

PR #6151, opened 2026-07-10
Add a Yjs collaboration primitive with ui.codemirror as its first consumer

evnchn:
### Motivation

This is the primitive half of the reframe from #5791: *"stop treating this PR as 'adding a Tiptap element' and start treating it as 'adding a Yjs collaboration primitive, with Tiptap as its first consumer'"* — shipped here with `ui.codemirror` as the first consumer instead, because CodeMirror already has the battle-tested `y-codemirror.next` binding while Tiptap needs ~3K LOC of new wiring. When #5791 matures, it consumes `nicegui/yjs_room.py` in place of its own `tiptap_room.py`.

Lineage: #5774 (the original ask), #5775 (cursor preservation — explicitly last-writer-wins, not collaboration), #5791 (direction + design rulings).

```python
ui.codemirror(language='Markdown').with_crdt('shared-doc')
ui.codemirror().with_crdt('gated-doc', access_check=lambda doc_id, sid: is_allowed(sid))
```

Opened as **draft** per the green light in #5791 — staged and reviewed on [evnchn#153](https://github.com/evnchn/nicegui/pull/153) (fork CI green on the identical tree), with named follow-ups below still open. Review whenever bandwidth allows; no rush.

### Implementation

- **`nicegui/yjs_room.py`** — pycrdt-backed `Y.Doc` per named room; four `yjs_*` Socket.IO handlers installed lazily on first `.with_crdt()` (zero overhead otherwise); rooms evict when the last client leaves; per-doc `access_check(doc_id, sid)` hook (sync or async, AND-combined, default open — the C3 ruling); binary frames end-to-end; `asyncio.to_thread` for pycrdt calls above 32 KB.
- **`CrdtElement` mixin** — `.with_crdt(doc_id, *, access_check=None)` for any element whose Vue component can bind a Y.Doc.
- **`ui.codemirror`** — `y-codemirror.next` binding via a shared `_yjs_bundle` ESM (one Yjs copy for all future consumers); while CRDT is on, the `value` prop is inert and Yjs owns canonical state.
- **Chunked transfers** — updates over 500 KB are split and reassembled index-keyed on both sides, because Engine.IO's default `max_http_buffer_size` (~1 MB) otherwise kills the connection on a large paste or a big late-joiner initial sync. Surfaced by MQ's PromptGrimoireTool running pycrdt+NiceGUI collaboration for ~2500 students (offered on #5791); their production experience also motivated the per-document-affinity note on multi-process deployments.
- `pycrdt` stays optional (`pip install "nicegui[crdt]"`, registered in `optional_features`); runnable demo at `examples/collaborative_editor/`.

<details><summary>Design decisions & trade-offs</summary>

- **pycrdt over y-py**: y-py is officially abandoned; pycrdt is the maintained binding with prebuilt wheels, and the same library PromptGrimoireTool runs in production.
- **Raw `@sio.on` handlers, not the element event tunnel**: keeps Yjs payloads binary; routing through `$emit` JSON-stringifies each `Uint8Array` (~5× wire inflation per keystroke). Disconnect cleanup still uses `Client.on_disconnect`, the proper lifecycle tunnel.
- **`LOOPBACK = None` + inert `value` prop in CRDT mode**: Yjs owns canonical state; the runtime-`value`-setter semantics you ruled on (I5: server-side rebuild + broadcast) are a named follow-up, not silently half-implemented.
- **In-tab sharing**: multiple editors with one `doc_id` in the same tab share one client-side `Y.Doc`, so in-tab propagation needs no server round-trip.
- **Awareness removal on disconnect** is left to y-protocols' 30 s timeout for now (mapping sid → Yjs clientID needs separate tracking).
- **Multi-process honesty**: rooms live in process memory; collaborators of a `doc_id` need per-document affinity routing — per-client sticky sessions do *not* guarantee it. Cross-process room sync is deliberately out of scope (it buys availability, not horizontal scale).

</details>

<details><summary>Review provenance (fork staging)</summary>

Staged as [evnchn#153](https://github.com/evnchn/nicegui/pull/153): five Copilot review rounds + manual review rounds (shared `_yjs_bundle` instead of per-element Yjs copies, `CrdtElement` mixin extraction, E2E tests instead of mocked unit tests), then a 2026-07-10 revival: merge with current `main` (incl. the codemirror keybindings/constants refactor), the chunking layer, and an adversarial second-lineage (Codex) pass which drove the index-keyed reassembly and the transfer-size cap. Fork CI green on the identical tree; the squashed commit here preserves it byte-for-byte.

Known follow-ups (tracked on the fork, will land here or as follow-up PRs as you prefer): two-browser E2E helper (the `Screen` fixture is single-window; current tests cover two editors in one tab plus chunked-transfer round-trips), the I5 `value` read/write semantics, and a niche `line_tooltips`-set-at-construction × seeded-CRDT-doc ordering interaction.

</details>

### Progress

- [ ] The implementation is complete. (Draft — follow-ups named above.)
- [ ] Documentation has been added/updated. (Docstrings + runnable example are in; a website documentation page is TBD once the API shape is confirmed.)
- [x] No breaking changes to the public API — `with_crdt()` is opt-in and additive.


evnchn:

**Known gap, logged for the rest of this branch's work** — merging `main` (8fb479bc) brought upstream's line-anchor tracking onto this branch, and the two features mis-initialise together: in CRDT mode the editor is built from `ytext`, so a room whose state arrives *after* mount validates `line_anchors` against a still-empty document and discards them as out of range. Not fixed in the merge — it needs init-sequencing logic plus a test, so it lands with the remaining branch work.

<details>
<summary><b>Failure path, what the merge did fix, and provenance</b></summary>

**Repro sketch** (untested as a written test — derived from reading the merged sources, not yet reproduced in CI):

```python
yjs_room.get_doc('x')  # seed with multi-line text before any client connects
ui.codemirror(line_anchors={'mid': 3}).with_crdt('x')
```

On mount, `_setupCrdt()` has not necessarily received `yjs_init`, so `new CM.EditorView({doc: this.ytext.toString()})` can start from an empty document (`codemirror.js:551`). `applyLineAnchors()` then range-validates against that document and drops line 3 (`codemirror.js:265`). When Yjs later inserts the real content there is no anchor left to remap, so the Python-side `line_anchors` constructor path (`codemirror.py:47`) silently yields nothing.

**Also open, lower severity:** a *disabled* CRDT editor still receives remote Yjs edits and remaps anchors client-side, but `DisableableElement` drops incoming events while disabled (`disableable_element.py:20`), so Python's `line_anchor_positions` mirror can go stale. Worth a test alongside the above.

**What the merge *did* fix.** `beforeUnmount()` merged with no conflict marker and was still wrong: this branch's `if (this.crdtDocId) { …; return; }` sat in front of upstream's new anchor snapshot and `clearTimeout(this._anchorTimer)`. In CRDT mode that leaked a pending debounce, which would fire `$emit("anchor-positions", …)` on a torn-down editor, and a client-side remount re-applied the *declared* anchor lines instead of the current ones. Restructured so the snapshot and the timer clear always run, with only the `value` write-back staying CRDT-exempt (Yjs owns the document) and `_teardownCrdt()` moved after the state read.

The three textual conflicts were mixin ordering in `codemirror.py` and, in `codemirror.js`, the `props` block plus the `extensions` array — where only `changeSender` belongs inside the `crdtDocId` conditional, since `anchorTracker`/`anchorField` fire on `docChanged` and never consult `emitting`, so remote Yjs edits should remap anchors too.

**Provenance.** The `beforeUnmount` defect and the init-order gap above were both found by a Codex review pass, not by the model that wrote the merge. Verified locally after the fix: `mypy` clean, `pylint` 10.00/10, `pre-commit` clean, MRO resolves with both mixins, and `node --check` on the merged component. The browser-backed `screen`-fixture suites were **not** run locally; CI covers them.

</details>

PR #6152, opened 2026-07-10
Make the nicegui package import lazy via PEP 562

evnchn:
*Drafted with Claude Code (assisting @evnchn). Split out of #6112 (commit 1) per the direction-check there; reviewed on the `evnchn/nicegui` fork at evnchn#195 before opening here. This is the **enabler** half — it makes `import nicegui` lazy so #6112's worker-stub half can build on it.*

### Motivation

After #5303, `nicegui/ui.py` is lazy but `nicegui/__init__.py` is still fully eager, so `import nicegui` / `from nicegui import ui` re-pays FastAPI + `socketio` + `App()` construction in **every** subprocess that re-imports the user's main module — the cost behind #5684 (`run.cpu_bound`, `multiprocessing.Pool`, `Manager()`).

Measured on this machine (Python 3.12, fresh process, `orjson` present): **`import nicegui` drops from ~230 ms to ~2.5 ms** — a ~90× cut, and the prerequisite for the per-worker savings in #6112.

### Implementation

- `nicegui/__init__.py` gains the PEP 562 `_LAZY_IMPORTS` + module `__getattr__`/`__dir__` pattern already used by `ui.py` (#5303); the eager imports move under `TYPE_CHECKING` so static tooling is unchanged.
- the package `__getattr__` falls back to `importlib.import_module` for any other submodule name, so `import nicegui; nicegui.events` keeps working exactly as under the eager init — deterministically, not depending on what has been imported before.
- `core.py`: the `socketio.AsyncServer` import becomes `TYPE_CHECKING`-only; `core.app` / `core.sio` now materialize on first access via a `core.__getattr__` that imports `.nicegui`.
- `nicegui.json` no longer imports FastAPI eagerly: `NiceGUIJSONResponse` moves to a new `json/response.py` behind a lazy `__getattr__`, and both wrappers gain a module-private `_dumps_bytes(obj) -> bytes` (the byte-level serializer the response class used to inline) — the public `nicegui.json` API is unchanged.
- new `nicegui/_lazy.py`: one shared `resolve()` helper for the three module `__getattr__` implementations (package, `ui.py`, `nicegui.json`) — no logic duplicated.
- the package module installs a tiny `ModuleType` subclass whose `__setattr__` absorbs the import machinery binding the `nicegui.app` / `nicegui.context` **submodules** as package attributes, which would otherwise shadow the `App` / `Context` **instances** that `__getattr__` provides. Non-module assignments (e.g. `mock.patch`) still behave as plain instance attributes.
- `helpers/functions.py` imports the `awaitable_response` **module** instead of the `AwaitableResponse` **class**, breaking a pre-existing `nicegui.awaitable_response` → `nicegui.background_tasks` → `nicegui.helpers` → `nicegui.awaitable_response` cycle that the eager package init used to hide.

<details>
<summary><b>Empirical: import-time before/after (ran both branches)</b></summary>

Fresh interpreter per run, `time.perf_counter()` around `import nicegui`, same venv:

| | `import nicegui` (steady-state) |
|---|---|
| `main` (eager) | ~230 ms |
| this branch (lazy) | ~2.5 ms |

All lazy attributes resolve correctly: `nicegui.ui`, `nicegui.App`, `nicegui.app` (the `App` instance, not the submodule), `nicegui.__version__`.

</details>

### Progress

- [x] Documentation is not necessary (internal import mechanics; no public API surface added).
- [ ] No breaking changes to the public API — **one narrow behavior note:** `NiceGUIJSONResponse` is no longer importable from the private submodules `nicegui.json.orjson_wrapper` / `nicegui.json.builtin_wrapper` (it now lives in `nicegui.json.response`); the documented package path `from nicegui.json import NiceGUIJSONResponse` is unchanged.


falkoschindler:
Thanks for the PR, @evnchn! I'd like to see if we can merge it for the next release, so I asked Fable 5 for a review. Maybe you can have a look into its findings?

---

Nice work — the mechanism itself is sound: I probed the submodule-vs-instance binding in various import orders, `mock.patch`, `importlib.reload(nicegui)`, uvicorn's `nicegui:app` string import, pickling `App`/`Client`/`ElementFilter` into a spawn child, and the `_lazy.resolve` dotted paths; all fine, and ruff/mypy/pylint on `nicegui/` are green. One real regression though, and a behavior change worth a conscious decision.

### Blocking

1. **Latent import cycle is now exposed: importing `nicegui.awaitable_response` (or `nicegui.functions.javascript` / `nicegui.functions.clipboard`) as the first nicegui import raises `ImportError`** — chain: [awaitable_response.py:3](nicegui/awaitable_response.py#L3) `from . import background_tasks` → [background_tasks.py:7](nicegui/background_tasks.py#L7) `from .helpers.warnings import warn_once` → `helpers/__init__.py` → [helpers/functions.py:9](nicegui/helpers/functions.py#L9) `from ..awaitable_response import AwaitableResponse` → `ImportError: cannot import name 'AwaitableResponse' from partially initialized module`. On `main` the eager package init happened to run this chain in a safe order; now the order depends on what the user imports first. Realistic trigger: a component module with sorted imports (`from nicegui.awaitable_response import AwaitableResponse` sorts before `from nicegui.element import Element`, and `from my_components import …` sorts before `from nicegui import ui` in `main.py`). I swept all 247 submodules as first import in a fresh interpreter — exactly these three fail. Fix: break the cycle rather than rely on order — e.g. in `helpers/functions.py` replace the from-import with `from .. import awaitable_response` and use `awaitable_response.AwaitableResponse` in the `isinstance` check in `should_await` (line 54); verified locally that this alone makes the sweep and `tests/test_lazy_imports.py` pass. Please also add a regression test — one subprocess that iterates `pkgutil.walk_packages(nicegui.__path__, 'nicegui.')`, purges `nicegui*` from `sys.modules` before each `import_module`, and asserts no failures — so the next latent cycle can't hide behind import order again.

### Major

1. **Bare `import nicegui` no longer exposes transitively-imported submodules as attributes** — `import nicegui; nicegui.events` / `nicegui.background_tasks` / `nicegui.core` / `nicegui.version.__version__` all raise `AttributeError` now (worked on `main`), and worse, it is order-dependent: `import nicegui; nicegui.app; nicegui.background_tasks` succeeds because the first access imports the chain. `from nicegui import events` still works, but `import nicegui as ng; ng.events.ClickEventArguments` is a pattern users do write. Options: (a) add the documented modules (`events`, `background_tasks`, maybe `testing`) to `_LAZY_IMPORTS`/`__all__`, (b) let `__getattr__` fall back to `importlib.import_module(f'.{name}', 'nicegui')` for any submodule name so every `nicegui.<submodule>` attribute access works deterministically, or (c) accept it and say so in the PR/release notes. I lean towards (b) — it removes the order dependence entirely — but it's a maintainer call.

### Minor

1. **`core.__getattr__` hand-rolls what `_lazy.resolve` already provides** ([core.py:29](nicegui/core.py#L29)) — `return _lazy.resolve(__name__, 'nicegui', {'app': ('.nicegui', 'app'), 'sio': ('.nicegui', 'sio')}, name)` is behaviorally identical (same objects, same "partially initialized" hint on a genuine cycle, same error text for unknown names); the `if name in globals()` branch is redundant since `nicegui.py` binds `core.app = app = App(...)` in one statement. Would drop the third lazy-getattr flavour the PR otherwise unifies.
2. **`render()` widens the public `nicegui.json` API** ([orjson_wrapper.py:53](nicegui/json/orjson_wrapper.py#L53), [json/__init__.py](nicegui/json/__init__.py)) — the name reads like HTML rendering, it's exported via `__all__` (permanent commitment), and it repeats the `orjson.dumps(..., option=ORJSON_OPTS, default=_orjson_converter)` triple from `dumps`. Consider a private `_dumps_bytes` (importable by `response.py` without being API) or at least a name that says bytes. Also: the module docstring still says "Provides dumps and loads implementations", and `nicegui.json` gained `__getattr__` without a matching `__dir__` (unlike `ui.py`/`__init__.py`).
3. **`NiceGUIJSONResponse` no longer importable from the wrapper submodules** — already flagged in the PR checklist; I agree it's acceptable for private submodules, just make sure it lands in the release notes.
4. **`_PackageModule.__setattr__` hard-codes `{'app', 'context'}`** ([__init__.py:76](nicegui/__init__.py#L76)) — the set silently mirrors `_LAZY_IMPORTS`; a future non-module attribute colliding with a submodule name would reintroduce exactly the bug this guards against without any test noticing. `{n for n, (_, attr) in _LAZY_IMPORTS.items() if attr}` would make it self-maintaining. Take-or-leave.
5. **Package `__dir__` returns only `__all__`** ([__init__.py:56](nicegui/__init__.py#L56)) — `'__file__' in dir(nicegui)` is now `False`; `help(nicegui)`/`inspect.getmembers` and REPL completion see 15 names. `ui.py` already does the same, so consistent, but `sorted(set(__all__) | set(globals()))` would keep parity with `main`.
6. **Tests** ([test_lazy_imports.py](tests/test_lazy_imports.py)): the four new package tests are copy-paste twins of the `ui` ones (`_match_all`, `_names_resolve`, `_does_not_import_web_framework`) — `@pytest.mark.parametrize('module', [nicegui, ui])` would halve them and extend `test_dir_returns_expected_names`/`test_nonexistent_attribute_raises` to the package for free (currently untested there). `test_package_app_and_context_can_be_monkeypatched` only exercises `app`.
7. **`_lazy.resolve` annotates `tuple[str, str | None]`** but no caller passes `None`, the docstring says "empty attribute name", and `json/__init__.py` annotates `tuple[str, str]` — tighten to `str`.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20from%20a%20multi-agent%20review%20run%2C%20verified%20against%20the%20branch%20and%20written%20up%20by%20the%20model.&d=2026-08-17&a=claude-code "claude-fable-5: Findings from a multi-agent review run, verified against the branch and written up by the model.")

evnchn:
@falkoschindler good timing (early 3.17, better than late, in fear of a big regression for a PR touching everything). On it. 

I also wonder if this PR has anything to do with fixing https://github.com/zauberzeug/nicegui/issues/6290 or making it worse.

evnchn:
Thanks @falkoschindler — and thanks to the reviewer, that Blocking finding is real and I could not have asked for a better catch. Pushed as 2a8428c9.

**All 9 findings addressed.** The Blocking one reproduced exactly as described (3 of 247 submodules), is fixed, and is now guarded by a sweep test that I confirmed fails when the fix is reverted. For the Major I took option (b). One correction to Minor 7 below: a caller *does* pass `None`, so I normalized it instead.

<details>
<summary><b>Blocking 1 — latent import cycle</b> · fixed + regression test</summary>

Reproduced verbatim on the branch:

```
$ .venv/bin/python -c "import nicegui.awaitable_response"
ImportError: cannot import name 'AwaitableResponse' from partially initialized module
'nicegui.awaitable_response' (most likely due to a circular import)
```

Fixed as suggested — `helpers/functions.py` now imports the **module**, not the class:

```python
from .. import awaitable_response  # import the module, not the class, to break the cycle via nicegui.helpers
...
    return isinstance(result, Awaitable) and \
        not isinstance(result, (NoImplicitAwait, awaitable_response.AwaitableResponse, asyncio.Task))
```

The regression test is `test_every_submodule_can_be_imported_first`: one subprocess walking `pkgutil.walk_packages`, purging `nicegui*` from `sys.modules` before each `import_module`. **Seen to fail before the fix, pass after** — the failure list matches the reviewer's sweep exactly:

```
$ git stash push nicegui/helpers/functions.py   # revert only the fix
$ pytest tests/test_lazy_imports.py::test_every_submodule_can_be_imported_first
E  AssertionError: nicegui.awaitable_response: cannot import name 'AwaitableResponse' ...
E    nicegui.functions.clipboard: cannot import name 'AwaitableResponse' ...
E    nicegui.functions.javascript: cannot import name 'AwaitableResponse' ...
1 failed

$ git stash pop && pytest tests/test_lazy_imports.py
17 passed
```

The sweep now also fails on **non-`ImportError`** import-time failures (a circular import can surface as `AttributeError`/`RuntimeError` too) and only skips a `ModuleNotFoundError` whose missing module is not a `nicegui` one. It asserts the module list is non-empty so it cannot pass vacuously. Locally: 245 modules swept, 0 skipped, 0 failures.

</details>

<details>
<summary><b>Major 1 — <code>import nicegui; nicegui.&lt;submodule&gt;</code></b> · took option (b)</summary>

Confirmed the regression and that it was order-dependent. Option (b), as the reviewer leaned:

```python
def __getattr__(name: str) -> object:
    if name not in _LAZY_IMPORTS and not name.startswith('_'):
        try:
            # import submodules on demand, so that `import nicegui; nicegui.<submodule>` keeps working as on an
            # eagerly initialized package - independently of what has been imported before
            return importlib.import_module(f'.{name}', __name__)
        except ModuleNotFoundError as e:
            if e.name != f'{__name__}.{name}':
                raise  # the submodule exists, but one of its own imports is missing
    return _lazy.resolve(__name__, 'nicegui', _LAZY_IMPORTS, name)
```

Three details worth naming:

- The `e.name` check keeps a genuine missing dependency *inside* a submodule from being swallowed and re-reported as a bogus `AttributeError`.
- Skipping `_`-prefixed names keeps dunder/protocol probes (`__copy__`, `__deepcopy__`, IPython canaries) off the import machinery — they were resolving through a failed `import_module` on every lookup.
- Unknown names still raise the same `AttributeError` as before: `module 'nicegui' has no attribute 'invalid_name'`.

Guarded by `test_submodules_are_accessible_as_package_attributes` (fresh subprocess, no submodule imported first, so the attribute can only come from `__getattr__`).

</details>

<details>
<summary><b>Minors 1–7</b></summary>

| # | Finding | Action |
|---|---|---|
| 1 | `core.__getattr__` hand-rolls `_lazy.resolve` | **Done** — now a one-liner delegating to `_lazy.resolve`; the `if name in globals()` branch dropped as the reviewer showed it redundant. |
| 2 | `render()` widens the public `nicegui.json` API | **Done** — renamed to the module-private `_dumps_bytes()`, dropped from `__all__`, `__dir__` added. Public API is now byte-identical to `main`. The `dumps`/`_dumps_bytes` orjson-triple duplication I left: the two differ in the conditional `opts` build, and folding them costs more than it saves. |
| 3 | `NiceGUIJSONResponse` not importable from the wrapper submodules | **Acknowledged** — stays in the PR checklist and will go in the release notes. |
| 4 | `_PackageModule.__setattr__` hard-codes `{'app', 'context'}` | **Done** — derived: `_NON_MODULE_LAZY_IMPORTS = {name for name, (_, attr) in _LAZY_IMPORTS.items() if attr}`. |
| 5 | Package `__dir__` returns only `__all__` | **Done** for the package **and** for `ui.py` — both now `sorted(set(__all__) \| set(globals()))`, so `'__file__' in dir(...)` holds again. Asserted in the parametrized test. |
| 6 | Copy-paste twin tests | **Done** — a `module` fixture parametrized over `[nicegui, ui]` covers `_LAZY_IMPORTS`-matches-`__all__`, names-resolve, `__dir__`, nonexistent-attribute and no-web-framework for **both** (the package had no `__dir__`/nonexistent coverage before). `test_package_objects_can_be_monkeypatched` now parametrizes over `app` **and** `context`. Net: 13 tests → 17, and fewer lines. |
| 7 | `_lazy.resolve` annotates `tuple[str, str \| None]` but no caller passes `None` | **Corrected** — one caller did: `ui.py`'s `'clipboard': ('.functions.clipboard', None)`, the only module-valued entry in that table, while every module-valued entry in the package table used `''`. Normalized `clipboard` to `''` and tightened the annotation to `str`. mypy is green on the tightened signature, which it would not have been otherwise. |

</details>

<details>
<summary><b>Re #6290 — this PR neither fixes nor worsens it</b></summary>

I raised the worry above; measured, and it is unfounded in both directions.

The ESM importmap is populated by the eager block at the bottom of `ui.py` (unchanged here), not by the package init, so the realistic import path is unaffected:

```
$ python -c "from nicegui import ui; from nicegui.dependencies import esm_modules; print(len(esm_modules))"
this branch: 10      main: 10
```

The one delta is that a bare `import nicegui` that never touches `ui` now registers 0 instead of 10 — but no element can be created without reaching `nicegui.ui` (or importing an element module directly, which registers it), so no page can observe it.

#6290 itself is about the **external-package** path: `nicegui-highcharts` reaches `register_esm()` only through the lazy `ui.highchart` entry that #5303 introduced, so on a page whose importmap is already serialized the specifier is missing. That mechanism is on `main` today and this PR does not touch it. Whatever the fix turns out to be — eager registration of external ESM packages, or a runtime shim — it is orthogonal to this diff.

</details>

<details>

- `pre-commit run --all-files`: **pass**
- `mypy ./nicegui`: **Success, no issues in 247 source files**
- `pylint ./nicegui`: **10.00/10**
- `pytest tests/test_lazy_imports.py`: **17 passed**
- Regression test **seen to fail with the fix reverted** (output above) — a green-only test would prove nothing here.
- `import nicegui` still ~2.5 ms (`python -X importtime`: `2529 µs`); `import nicegui` alone still pulls in none of fastapi/starlette/socketio/engineio/uvicorn/matplotlib/httpx.
- **Independent adversarial review (Codex, separate lineage), asked to refute the diff: BLOCKING none, MAJOR none, one MINOR** — that the sweep swallowed non-`ImportError` failures. Fixed above; the sweep is strict now. Its cleared checks, each run rather than reasoned: the fallback does not mask a nested `ModuleNotFoundError` (verified with a monkeypatched `import_module`); dunder/protocol probes neither recurse nor import heavy modules; `nicegui.app`/`nicegui.context` still resolve to the instances including submodule-first order; `_NON_MODULE_LAZY_IMPORTS` matches the old hardcoded set on the only real collisions (`app`, `context`); the cycle fix holds across six import orders; `nicegui.json.render` is gone and `NiceGUIJSONResponse` still imports; `ui.clipboard` still resolves as a module under the `''` entry.
- Full local `pytest` run in progress; I will not call this ready until CI is green.

</details>

evnchn:
**CI green on 2a8428c9** — all checks pass (`quick-test / 3.10 / pytest`, `check / mypy`, `check / pre-commit`, `check / pylint`, `windows / storage`; `full-test` is merge_group-only). Ready for your look.

<details>
<summary>Local full suite too</summary>

`pytest tests` on macOS: **1002 passed, 3 skipped, 2 xfailed, 4 failed**. The 4 are `tests/test_page.py::test_dark_mode[None|mini|wind3|wind4]`, and they fail identically on a clean `main` checkout on this machine — the OS is in dark mode, so the `dark=None` page renders `red` where the test expects `blue`. Environment, not this diff.

The 3 skips are `tests/test_run.py` (`needs a platform with a fork default (e.g. Linux)`) — covered by the Linux CI run above.

</details>

evnchn:
**Four more findings, all real, all fixed in 255aeb7c** — from a second review pass I ran over the whole PR diff (not just the last commit). One is a genuine trap I introduced, one is a regression in my own answer to Minor 5, and two are test guards that could not fail.

Each fix is mutation-proven: I broke it again and confirmed the guarding test goes red.

<details>
<summary><b>1 · <code>_PackageModule.__setattr__</code> swallowed deliberate module assignments</b> — medium, my bug</summary>

The guard absorbed **any** `ModuleType` assigned to a lazy-object name, not just the import machinery's own binding. So a test stubbing the app with a module-like fake passes green having tested nothing:

```python
>>> import types, nicegui
>>> nicegui.app = types.ModuleType('fake')
>>> type(nicegui.app).__name__
'App'                                   # the assignment silently vanished
>>> with mock.patch.object(nicegui, 'app', types.ModuleType('fake')):
...     type(nicegui.app).__name__
'App'                                   # patch reports success; body runs unpatched
```

Fixed by matching the machinery's binding precisely instead of by type:

```python
if name in _NON_MODULE_LAZY_IMPORTS and getattr(value, '__name__', None) == f'{__name__}.{name}':
```

`test_package_objects_can_be_monkeypatched` now parametrizes over `[mock.DEFAULT, ModuleType('fake')]` × `[app, context]`. Reverting the guard to the old `isinstance(value, ModuleType)` fails both `module` cases and passes both `mock` cases — the old test suite could not see the difference.

</details>

<details>
<summary><b>2 · widening <code>ui.__dir__</code> was a regression</b> — I over-applied Minor 5</summary>

The reviewer's Minor 5 asked for `sorted(set(__all__) | set(globals()))` on the **package**, where it restores parity with `main` (the eager package already had its submodules in `globals()`). I applied it to `ui.py` as well for symmetry. That was wrong: on `main`, `set(dir(ui)) - set(ui.__all__)` is **empty**, so widening it there *adds* 17 implementation names:

```
['Path', 'TYPE_CHECKING', '_LAZY_IMPORTS', '_lazy', 'builtins', 'importlib', ...]
```

Concretely that puts `ui.Path` / `ui.builtins` / `ui.importlib` into REPL and IDE completion, and `website/documentation/content/doc/api.py:_find_attribute()` resolves names by scanning `dir()` and taking the first hit — with `dir()` sorting uppercase first, a documented symbol named `Path` would title its page after the wrong attribute. Latent rather than live (I diffed the generated docs registry main-vs-PR: no title changes today), but there is no reason to take it on.

`ui.__dir__` is back to `return __all__`, byte-identical to `main`. The package keeps the widening, now asserted by its own test rather than in the shared parametrized one.

</details>

<details>
<summary><b>3 · <code>test_module_access_does_not_import_others</code> could never fail</b> — pre-existing, in scope</summary>

```python
sys.exit('button' in sys.modules.keys())
```

`sys.modules` is keyed by the **fully qualified** name, so the key is `nicegui.elements.button` and the expression is always `False`. This was the only guard for "accessing one lazy attribute does not import its siblings", and it had been vacuous since it was written; the PR touches this exact statement, so I fixed it rather than leave it. Now `sys.exit('nicegui.elements.button' in sys.modules)`, and adding a deliberate `ui.button` to the subprocess makes it go red — the property does hold, it just was not being checked.

</details>

<details>
<summary><b>4 · the sweep evicted third-party packages too</b> — minor</summary>

`key.startswith('nicegui')` also matches distributions whose module name merely begins with `nicegui` — `nicegui_highcharts` is installed in this repo's test environment, so the sweep re-imported it ~250 times and coupled itself to that package's re-import safety, where a failure would have been reported against a nicegui submodule. Now `key == 'nicegui' or key.startswith('nicegui.')`.

</details>

<details>

- `pre-commit run --all-files`: **pass** · `mypy ./nicegui`: **Success, 247 files** · `pylint ./nicegui`: **10.00/10**
- `pytest tests/test_lazy_imports.py`: **20 passed**
- **Mutation checks** (each fix reverted in turn, expecting red):
  - broad `isinstance` guard → `test_package_objects_can_be_monkeypatched[module-app]`, `[module-context]` **fail**
  - dropping `sys.modules[__name__].__class__ = _PackageModule` → `test_lazy_objects_win_over_submodule_shadowing` **fails** (the shadowing guard is still load-bearing after the narrowing)
  - touching `ui.button` in the subprocess → `test_module_access_does_not_import_others` **fails** (it now can)
  - reverting `helpers/functions.py` to the class-import → `test_every_submodule_can_be_imported_first` **fails** with the three circular-import errors (unchanged from the previous round)
- Full local `pytest` re-running on 255aeb7c; the only failures in the previous full run were the environmental `test_dark_mode` oklch ones that also fail on a clean `main` on this machine.
- Also independently re-verified end to end on this branch: `website/main.py` boots, `HTTP 200`; every `__all__` name resolves as a first access in a fresh interpreter for both `nicegui.X` and `from nicegui import X`; the generated documentation registry is identical to `main`.

</details>

evnchn:
**CI green on 255aeb7c** — `quick-test / 3.10 / pytest`, `check / mypy`, `check / pre-commit`, `check / pylint`, `windows / storage` all pass. Ready for your look.

<details>
<summary>Local full suite: 1004 passed, 9 failed, all pre-existing</summary>

`pytest tests` on macOS: **1004 passed, 9 failed, 3 skipped, 2 xfailed** in 15:48.

- **8 × `test_dark_mode`** (4 in `tests/test_dark_mode.py`, 4 in `tests/test_page.py`) — this machine's OS is in dark mode, so the `dark=None` page renders `oklch(0.704 0.191 22.216)` (red) where the test expects blue. Fails identically on a clean `main` checkout here.
- **1 × `test_interactive_image.py::test_loaded_event`** — passes in isolation and passes running its whole file; a second full run of this same branch instead hit `test_sub_pages_with_url_fragments`. The 9th slot rotates under full-suite browser load rather than pointing at this diff, and Linux CI is green on the same commit.
- The 3 skips are `tests/test_run.py` (`needs a platform with a fork default (e.g. Linux)`) — covered by the CI run above.

</details>

PR #6157, opened 2026-07-19
Make CPython-only-semantics tests PyPy-compatible

evnchn:
### Motivation

NiceGUI's **runtime** installs, boots and works under [PyPy](https://pypy.org/) 3.11 — but a handful of tests assert **CPython-only semantics** and fail there, so the suite can't go green on PyPy even where the framework is fine.

This surfaced from an independent, external compatibility tracker, **[`nicegui-pypy`](https://github.com/evnchn-nicegui/nicegui-pypy)**, which runs NiceGUI's own tests on PyPy against a CPython control on a schedule. The point of that repo is that **NiceGUI doesn't have to add PyPy to its own CI** to know where it stands — the tracker carries that cost externally. These four tests are the only *test-level* blockers it found; this PR makes them PyPy-tolerant without changing CPython behaviour.

This is the focused test-compat slice of the earlier #5163 (closed) — **without** that PR's in-repo PyPy CI job / 120-min timeout / dependency drops, since the external tracker removes the need for them.

### Implementation

- `test_element::test_no_cyclic_references_when_deleting_{elements,clients}` — assert weakref-tracked objects vanish *immediately* after deletion, which relies on CPython **refcounting**; PyPy frees acyclic garbage only on a GC cycle. `gc.collect()` is **not** a fix — it also collects cycles, defeating the "no cyclic references" check — so `skipif` PyPy.
- `test_run::test_run_unpickable_exception_in_cpu_bound_callback` — asserts CPython's exact pickling-error string; widen the `match=` regex for PyPy's wording (same approach as #5163).
- `test_button::test_enable_disable` — on PyPy's slower warmup the re-enable races the next click; add a short `screen.wait(0.5)`, matching the event-timing approach in #5163.

<details><summary>Verification (NiceGUI test suite, local)</summary>

```
# CPython 3.11 — unchanged
$ python -m pytest tests/test_button.py::test_enable_disable tests/test_element.py tests/test_run.py -k "enable_disable or cyclic or unpickable"
4 passed

# PyPy 3.11 — before: 4 failed; after this PR:
$ pypy -m pytest ... -rs
2 passed, 2 skipped
SKIPPED test_no_cyclic_references_when_deleting_elements: PyPy has no reference counting
SKIPPED test_no_cyclic_references_when_deleting_clients:  PyPy has no reference counting
```

Diff: 3 files, +5 / −1. `pre-commit` (ruff/autopep8/codespell/quote-fixer) clean. Runtime deps unaffected (`pandas`/`polars` still can't run their tests on PyPy — a `pandas` SIGSEGV and a missing `polars` wheel, both upstream-library limits, tracked in nicegui-pypy — but those are out of scope here).
</details>

### Progress

- [x] Pytests are not necessary. (This PR only adjusts existing tests for PyPy.)

### DIFF
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -44,6 +44,7 @@ For coding rules, see [CONTRIBUTING.md](CONTRIBUTING.md); for general agent guid
   - Unclear problem statement or impact
 - **Cross-platform**
   - Windows path assumptions, locale/timezone hardcoding, reliance on system binaries without guards
+  - Tests asserting CPython-only semantics — immediate weakref cleanup via refcounting, exact CPython error-message wording — need a PyPy `skipif` or a tolerant match (PyPy compatibility is tracked externally, not promised)
 - **Readability**
   - Complex logic without comments explaining intent; magic numbers
 
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -83,5 +83,6 @@ def test_enable_disable(screen: Screen):
     assert events == [1]
 
     screen.click('Enable')
+    screen.wait(0.5)  # PyPy: let the re-enable reach the browser before clicking
     screen.click('Button')
     assert events == [1, 1]
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -83,6 +83,6 @@ def test_enable_disable(screen: Screen):
     assert events == [1]
 
     screen.click('Enable')
-    screen.wait(0.5)  # PyPy: let the re-enable reach the browser before clicking
+    screen.wait_for(screen.find_by_tag('button').is_enabled)
     screen.click('Button')
     assert events == [1, 1]
--- a/tests/test_element.py
+++ b/tests/test_element.py
@@ -1,3 +1,4 @@
+import platform
 import weakref
 
 import pytest
@@ -426,6 +427,7 @@ def test_update_before_client_connection(screen: Screen):
     screen.should_contain('Hello again!')
 
 
+@pytest.mark.skipif(platform.python_implementation() == 'PyPy', reason='PyPy has no reference counting')
 def test_no_cyclic_references_when_deleting_elements(screen: Screen):
     elements: weakref.WeakSet = weakref.WeakSet()
 
@@ -444,6 +446,7 @@ def test_no_cyclic_references_when_deleting_elements(screen: Screen):
     assert len(elements) == 0, 'all elements should be deleted immediately'
 
 
+@pytest.mark.skipif(platform.python_implementation() == 'PyPy', reason='PyPy has no reference counting')
 def test_no_cyclic_references_when_deleting_clients(screen: Screen):
     labels = weakref.WeakSet()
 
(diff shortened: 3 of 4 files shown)


falkoschindler (commented):
Thanks @evnchn — nicely scoped slice of #5163, and the PR description does a great job explaining *why* each test is CPython-only. The `skipif` reasoning for the cyclic-reference tests is exactly right: `gc.collect()` would also collect cycles and defeat the very thing the test asserts, so skipping on PyPy is the only faithful option. No blocking or major findings; two take-or-leave notes:

1. **Prefer a condition wait over a fixed sleep in `test_enable_disable`** (minor) — `Screen` already has a condition-based helper, so the race can be closed deterministically instead of with 0.5 s (which also gets paid on every CPython run and could still flake on a very slow PyPy warmup):

   ```suggestion
       screen.click('Enable')
       screen.wait_for(screen.find_by_tag('button').is_enabled)
       screen.click('Button')
   ```

   `find_by_tag('button')` returns the first button, which is the target one, and Quasar reflects `disable` into the native `disabled` attribute, so `is_enabled()` flips exactly when the update has reached the browser. (The `# PyPy:` comment could then go away entirely — the wait is correct on all implementations.)

2. **Widened pickling regex** (minor, leave-or-take) — `r"Can't pickle|Can't get local object"` now matches almost any pickling error on CPython too, so the assertion is a bit weaker there. Probably fine for what this test guards (the error surfaces instead of hanging), but if you want to keep CPython's specificity you could match the three known wordings explicitly, e.g. `r"Can't pickle (local object|<)|Can't get local object"` — assuming PyPy's pure-Python pickle says "Can't pickle \<function …\>: it's not found as …". Could you paste PyPy's actual error message? Then the regex can be as tight as before on both implementations.

Verified locally: `tests/test_run.py::test_run_unpickable_exception_in_cpu_bound_callback` passes on CPython 3.11 with the widened regex.

evnchn:
**Both applied** @falkoschindler — `wait_for` swap taken verbatim, and the pickle regex tightened using PyPy's *actual* error string (your guess confirmed on a real PyPy 3.10 run).

<details><summary><b>PyPy error string — captured, not guessed</b></summary>

Local `pypy3` (pure-Python `pickle`), same failure path:
```
Can't pickle <function ...raise_unpicklable_exception at 0x...>: it's not found as __main__...raise_unpicklable_exception
```
CPython says `Can't pickle local object <function ...>` (plus the `Can't get local object` AttributeError path). Your `Can't pickle (local object|<)` covers all three; final regex:
```python
match=r"Can't pickle (local object|<)|Can't get local object"
```
CPython test still green. PyPy isn't in this repo's CI — tracked in `evnchn-nicegui/nicegui-pypy`.
</details>

falkoschindler (approved):
Both follow-ups look great — the `wait_for` swap and the pickle regex confirmed against a real PyPy run. Approving. 🚀

So this doesn't regress silently with future tests: we've added a bullet to [REVIEW.md](https://github.com/zauberzeug/nicegui/blob/main/REVIEW.md) (Cross-platform section) — tests asserting CPython-only semantics (refcounting-based weakref cleanup, exact error-message wording) need a PyPy `skipif` or a tolerant match. PyPy compatibility stays best-effort, with your external tracker as the early-warning system. Thanks for carrying that!

PR #5973, opened 2026-04-22
Fix PermissionError on Windows when clearing file-backed storage

falkoschindler:
### Motivation

On Windows, `FilePersistentDict.clear()` intermittently raised `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process` during teardown (e.g. in pytest fixtures that reset NiceGUI state). Closes #5949.

The race is between the async `backup()` task scheduled by `super().clear()` (which opens the file via `aiofiles` and writes `{}`) and the synchronous `self.filepath.unlink()` that immediately follows. On Windows, open files cannot be deleted.

### Implementation

Move the "delete file when the dict is empty" semantic into `backup()` itself. `backup()` now checks whether the dict is empty and unlinks the file instead of writing `{}`. The `clear()` override is removed — `super().clear()` already triggers `backup()` via `on_change`.

Because `create_lazy` serializes tasks by name, the unlink is automatically ordered after any pending write, removing the race. As a bonus, we no longer write `{}` only to delete it moments later.

Behavior change worth noting: any operation that leaves the dict empty (e.g. popping the last key) will now also delete the file. This generalizes the stated goal from #4074 of not accumulating empty `{}` files on disk.

### Progress

- [x] The PR title is a short phrase starting with a verb like "Add ...", "Fix ...", "Update ...", "Remove ...", etc.
- [x] The implementation is complete.
- [x] This PR does not address a security issue.
- [x] Pytests are not necessary.
- [x] Documentation is not necessary.
- [x] No breaking changes to the public API.

Issue #5949, opened 2026-04-11
PermissionError on Windows systems during shutdown


atollk:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
# any type of test using the storage should do, but since this is non-deterministic, it's hard to verify
                                                                                                              
from nicegui import app, ui                                                                                                                                                 
from nicegui.testing import User    
import pytest                                                                                                                                          
                                                                                                                                                                              
@ui.page("/")
def index() -> None:
    ui.label("hello")

if __name__ in {"__main__", "__mp_main__"}:
    ui.run()                                                                                                                                                        
   
@pytest.mark.nicegui_main_file(__file__)
async def test_storage_write(user: User) -> None:                                                                                                                           
    await user.open("/")
    app.storage.general["key"] = "value"  
```

### Description

When running some tests on Windows machines (with the Github runners), an error came up non-deterministically:

```
_______ ERROR at teardown of test_ok_validation_no_filename_write_mode ________

    def finalizer() -> None:
        """Yield again, to finalize."""
    
        async def async_finalizer() -> None:
            try:
                await gen_obj.__anext__()
            except StopAsyncIteration:
                pass
            else:
                msg = "Async generator fixture didn't stop."
                msg += "Yield only once."
                raise ValueError(msg)
    
>       runner.run(async_finalizer(), context=context)

.tox\3.14\Lib\site-packages\pytest_asyncio\plugin.py:330: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\hostedtoolcache\windows\Python\3.14.3\x64\Lib\asyncio\runners.py:127: in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
C:\hostedtoolcache\windows\Python\3.14.3\x64\Lib\asyncio\base_events.py:719: in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
[...]

PR #6159, opened 2026-07-20
Fix PermissionError on Windows when clearing storage while a backup is in flight

### LINKED ISSUE #6158: Storage.clear() races in-flight storage backups on Windows (PermissionError, follow-up to #5949)
### First Check

- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty (#5949 fixed the sibling race in `FilePersistentDict.clear()`, but not this one).

### Example Code

Run on **Windows** (deterministic reproduction — the open handle stands in for an in-flight lazy backup write):

```python
import threading
from nicegui import app

app.storage.path.mkdir(exist_ok=True)
stale = app.storage.path / 'storage-user-stale.json'
stale.write_text('{}', encoding='utf-8')
handle = stale.open(encoding='utf-8')  # stands in for an in-flight lazy backup holding a storage file
threading.Timer(0.5, handle.close).start()
app.storage.clear()  # PermissionError (WinError 32) at the glob unlink
print('cleared cleanly')
```

### Description

#5973 fixed `FilePersistentDict.clear()` unlinking a file that a lazily scheduled backup still holds open — but `Storage.clear()` has the same race in its glob-unlink:

```python
for filepath in self.path.glob('storage-*.json'):
    filepath.unlink()
```

A backup scheduled via `background_tasks.create_lazy(async_backup(), name=...)` may be mid-write inside `aiofiles` (holding the file open) when the glob unlinks it. Unlinking an open file is fine on POSIX, but on Windows it raises `PermissionError` (WinError 32). Note that #5973 makes this *more* likely to fire: clearing the dicts now schedules a lazy backup, right before the glob runs.

We hit this in the wild in CI on `windows-latest` with the pytest `User` fixture, in the teardown of tests that write `app.storage.general` frequently:

```
  File "...\nicegui\app\app.py", line 381, in ...
    self.storage.clear()
  File "...\nicegui\storage.py", line 213, in clear
    filepath.unlink()
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process:
'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\nicegui-test-storage-...\\storage-general.json'
```

Full CI log: https://github.com/Jepson2k/Waldo-Commander/actions/runs/29744405874/job/88358708229

The glob-unlink can't simply be removed: as noted in #5949, it purges storage files of sessions that are *not currently loaded*, which the backup-on-empty mechanism can't cover. And `Storage.clear()` is synchronous, so it can't await the in-flight backup tasks. PR #6159 waits out the transient lock with a bounded retry.

*Investigated and written by Claude (Claude Code) on behalf of @Jepson2k.*

Jepson2k:
### Motivation

Fixes #6158.

`Storage.clear()` glob-unlinks the `storage-*.json` files while a lazily scheduled `FilePersistentDict` backup may still hold one open. Unlinking an open file works on POSIX but raises `PermissionError` (WinError 32) on Windows — #5973 removed exactly this race from `FilePersistentDict.clear()`, and the glob-unlink in `Storage.clear()` has it too (made more likely by #5973 itself, since clearing the dicts now schedules a lazy backup right before the glob runs). We hit it in CI on `windows-latest` in the teardown of storage-heavy tests using the `User` fixture; a deterministic reproduction script is in #6158.

Downstream CI (and the Windows repro in the review below) subsequently showed the same race firing from `async_backup`'s own empty-dict unlink — `missing_ok=True` covers a *missing* file, not a transiently *held* one — so the fix now covers all three unlink sites.

### Implementation

The glob can't simply be dropped (per #5949 it purges files of sessions that aren't currently loaded), and `clear()` is synchronous so it can't await the writers. Instead, unlinking goes through `helpers.unlink_with_retry` (with an `unlink_with_retry_async` variant that never blocks the event loop), which waits out a transient `PermissionError` (20 ms steps, bounded at 1 s) and then re-raises, so genuine permission problems still surface. Used at all three sites: `Storage.clear()`'s glob, `async_backup`'s empty-dict unlink, and `backup()`'s sync no-loop branch.

- POSIX behavior is unchanged: the first unlink attempt succeeds as before.
- On Windows, the retry only engages while another writer actually holds the file — normally a few milliseconds.

Two regression tests (`User` fixture) hold the storage file open across `app.storage.clear()` / an emptied dict's backup the way an in-flight write does — red on Windows before the fix (confirmed on a real Windows 11 box in review), green after. On POSIX they pass regardless, since the OS allows unlinking open files (the race is Windows-only by nature); CI runs on Ubuntu, so the red case only manifests for Windows developers running the suite locally.

### Progress

*Investigated and written by Claude (Claude Code) on behalf of @Jepson2k.*


evnchn:

**TL;DR — LGTM, one caveat.**

Reproduced the race RED→GREEN on a real Windows 11 box (Python 3.12): with the fix the regression test passes; revert it to a plain `filepath.unlink()` and it fails with the exact `PermissionError [WinError 32]`. Diagnosis and fix are correct and strictly non-regressive (POSIX unchanged; Windows worst-case = same error re-raised ~1 s later).

One caveat, not a blocker: **this test can't catch the regression in our CI** — it's green with or without the fix on Linux (the race is Windows-only, our matrix isn't). Suggest merging and adding a `windows-latest` leg as a follow-up so it actually has teeth.

<details><summary><b>Windows repro — verbatim (real box, not reasoned)</b></summary>

Tested at `b5d5d76` (PR head) · Windows NT 10.0.26100 · Python 3.12.10 · pytest 9.0.3.

**Fix present:**
```
tests/test_storage.py::test_clear_waits_out_transiently_held_storage_file PASSED [100%]
1 passed in 0.55s
```

**One-line change to disable the fix** (`nicegui/storage.py`, inside `Storage.clear()`):
```diff
         for filepath in self.path.glob('storage-*.json'):
-            _unlink_with_retry(filepath)
+            filepath.unlink()
```

**Fix disabled → the targeted failure:**
```
tests/test_storage.py::test_clear_waits_out_transiently_held_storage_file FAILED [100%]
nicegui\storage.py:214: in clear
    filepath.unlink()
E   PermissionError: [WinError 32] The process cannot access the file because it is being
    used by another process: '...\Temp\nicegui-test-storage-...\storage-general.json'
```
The teardown log shows the *same* WinError 32 firing inside `file_persistent_dict.py` `async_backup` — i.e. the lazily-scheduled backup is exactly the open-handle holder described. Restored clean afterward; green re-confirmed.
</details>

<details><summary><b>Design check — why #5973's serialization can't be reused here</b></summary>

#5973 fixed its sibling via `create_lazy(name=self.filepath.stem)` task-name serialization — which works only because a *live* `FilePersistentDict` owns that one file. `Storage.clear()`'s glob unlinks **arbitrary on-disk `storage-*.json`** (incl. never-loaded sessions, per #5949) with no live dict/task to serialize against, and `clear()` is synchronous so it can't `await` in-flight writers. So the bounded-retry is a defensible pragmatic choice; both claims in the PR body check out.
</details>

<details><summary><b>One nuance on the retry</b></summary>

`_unlink_with_retry` uses a blocking `time.sleep()`, so it relies on the handle being released *off* the event loop — which holds here (aiofiles' threadpool; the test's `threading.Timer`). A real WinError 5 (access-denied, not a sharing lock) would burn the full 1 s before surfacing, but only re-raises the same error it would have anyway. Non-regressive either way.
</details>

Nice work on the diagnosis + deterministic repro. Have you confirmed red→green on `windows-latest` in CI specifically, or a local Windows box? Happy to merge and track the Windows-CI leg separately.

Jepson2k:
Thanks for the thorough repro — especially catching the `async_backup` WinError in your teardown log. That's a genuine second instance of the race, not noise: our downstream CI hit it too once the glob was fixed. Pushed 01beaca9 extending the retry to all three unlink sites (`Storage.clear()`'s glob, `async_backup`'s empty-dict unlink, `backup()`'s sync no-loop branch), with the helper moved to `helpers.unlink_with_retry` plus an async variant that doesn't block the loop, and a second regression test for the backup path.

To answer your question: no `windows-latest` CI run — our evidence is downstream CI (Waldo-Commander, `windows-latest`, real-world hit linked in #6158) plus your local repro. Agreed a Windows CI leg is the right follow-up for giving these tests teeth; happy to open that as a separate PR if maintainers want it.

*Written by Claude (Claude Code) on behalf of @Jepson2k.*

evnchn:

**Re-reviewed `01beaca9` — still LGTM, and the async variant resolves the blocking-loop nuance I raised.** The Windows CI leg is already open as a draft (#6160), so no need to duplicate it — your new backup-path test will run there on Windows once this lands.

<details><summary><b>What I re-checked on the 3-site version</b></summary>

- `async_backup` → `await unlink_with_retry_async(...)` — non-blocking, correct for the event loop ✓
- `Storage.clear()` glob + `backup()`'s sync branch → blocking `unlink_with_retry` — correct, no loop there ✓
- `missing_ok=True` preserved at both backup sites; helper is monotonic-clocked ✓

Sharing one helper across all three unlink sites reads cleaner than the original local function.
</details>

falkoschindler (approved):
Thanks for the thorough investigation and the deterministic repro, @Jepson2k! I verified the diagnosis against the code: `aiofiles` handles stay open across the backup coroutine's await points, so the synchronous glob-unlink in `Storage.clear()` can genuinely race an in-flight writer on Windows — and #5973 indeed made the window more likely by scheduling a lazy backup right before the glob runs.

Instead of a review ping-pong I pushed a few refinements directly, hope that's fine:

- 815c837 — fail fast on POSIX (retry only for Windows-flavored `PermissionError`s, which are the only transient ones), `missing_ok=True` for the glob unlink (a concurrent empty-dict backup may delete the file between `glob()` and `unlink()` when `clear()` runs outside the loop thread), and a docstring note that the blocking retry can only wait out holders which release the file independently of the current thread.
- 6f6f35c — deduplicated the sync/async helpers via a shared `_try_unlink`.
- 0c2bd5a — merged the two regression tests into one two-phase test with a local `wait_until` helper.

Also thanks to @evnchn for confirming red→green on a real Windows machine.

PR #6160, opened 2026-07-21
Add a Windows CI leg for the storage file-lock tests

evnchn:

### Motivation

Our CI matrix is Ubuntu-only, but a whole class of `Storage.clear()` / `FilePersistentDict` teardown bugs is **Windows-only** — `PermissionError [WinError 32]` when a storage file is unlinked while an in-flight backup still holds it open (see #6159, #6158). These are invisible to us today: #6159's regression tests are green on Linux *with or without* the fix, because the race can't occur on POSIX. This adds one narrow, cheap Windows job to close that blind spot.

### Implementation

- **New `windows-test` job** in `ci-gate.yml` on `windows-2022`, running `pytest tests/test_storage.py -m "not screen"` — the non-browser storage tests. On current `main` these are exactly the three file-race/teardown tests from #6159, i.e. the tests that only have teeth on Windows. The job takes under a minute.
- **New `screen` marker**, registered in `pyproject.toml` and auto-applied by a small `conftest.py` hook to every test that uses the browser-based `screen` fixture — so the Windows job selects by mechanism, not by a hand-maintained test list.
- **Gating:** the job is wired into the `needs` of both `quick-test-gate` and `full-test-gate`, so a Windows storage regression blocks both PR merges and the merge queue. Since the selected tests are deterministic `user`-fixture tests (no Selenium), there is no flakiness risk in the gate.
- **Verified red→green on real Windows** (from #6159): with the fix present, the file-race test passes; with `unlink_with_retry` reverted to a plain `unlink()`, it fails with `PermissionError: [WinError 32]`. On Linux it is green either way — which is exactly why this leg exists.

<details><summary><b>Why the browser tests are excluded</b></summary>

25 of the storage tests use the Selenium `screen` fixture and are intermittently flaky on `windows-2022` for reasons unrelated to storage: `Timed out receiving message from renderer` in the browser, and a `ConnectionResetError: [WinError 10054]` from the Windows Proactor event loop during teardown that `filterwarnings = error` elevates to a failure. An earlier draft of this PR ran the full file with `pytest-rerunfailures --reruns 2` as a safety net; excluding the browser tests via the `screen` marker removes the flakiness at the root instead, and loses nothing relevant — the Windows-only failure mode lives entirely in the file layer, which the browser-independent tests cover.

</details>

### Progress

- [x] The PR title is a short phrase starting with a verb ("Add ...").
- [x] Pytests are not necessary — this runs existing tests on a new platform; no new tests added.

### DIFF
--- a/.github/workflows/ci-gate.yml
+++ b/.github/workflows/ci-gate.yml
@@ -27,6 +27,24 @@ jobs:
       python-versions: '["3.10", "3.11", "3.12", "3.13", "3.14"]'
       scope: '["pytest", "startup"]'
 
+  windows-test:
+    if: github.event_name != 'merge_group'
+    name: windows / storage
+    runs-on: windows-2022
+    timeout-minutes: 40
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
+      - name: Install Python
+        run: uv python install 3.10
+      - name: Install dependencies
+        run: uv sync --locked
+      - name: pytest tests/test_storage.py
+        run: uv run pytest tests/test_storage.py
+
   quick-test-gate:
     runs-on: ubuntu-latest
     needs: [check, quick-test]
--- a/.github/workflows/ci-gate.yml
+++ b/.github/workflows/ci-gate.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Install dependencies
         run: uv sync --locked
       - name: pytest tests/test_storage.py
-        run: uv run pytest tests/test_storage.py
+        run: uv run --with pytest-rerunfailures pytest tests/test_storage.py --reruns 2 --reruns-delay 2
 
   quick-test-gate:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-gate.yml
+++ b/.github/workflows/ci-gate.yml
@@ -28,10 +28,9 @@ jobs:
       scope: '["pytest", "startup"]'
 
   windows-test:
-    if: github.event_name != 'merge_group'
     name: windows / storage
     runs-on: windows-2022
-    timeout-minutes: 40
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v7
       - uses: astral-sh/setup-uv@v7
@@ -43,7 +42,13 @@ jobs:
       - name: Install dependencies
         run: uv sync --locked
       - name: pytest tests/test_storage.py
-        run: uv run --with pytest-rerunfailures pytest tests/test_storage.py --reruns 2 --reruns-delay 2
+        run: uv run --with pytest-rerunfailures==16.4 pytest tests/test_storage.py --reruns 2 --reruns-delay 2 --only-rerun TimeoutException --only-rerun PytestUnraisableExceptionWarning
+      - uses: actions/upload-artifact@v7
+        if: failure()
+        with:
+          name: screenshots-windows
+          path: screenshots/**/*.failed.png
+          retention-days: 14
 
   quick-test-gate:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-gate.yml
+++ b/.github/workflows/ci-gate.yml
@@ -41,27 +41,27 @@ jobs:
         run: uv python install 3.10
       - name: Install dependencies
         run: uv sync --locked
-      - name: pytest tests/test_storage.py
+      - name: pytest tests/test_storage.py -m "not screen"
         run: uv run pytest tests/test_storage.py -m "not screen"
 
   quick-test-gate:
     runs-on: ubuntu-latest
-    needs: [check, quick-test]
+    needs: [check, quick-test, windows-test]
     if: always() && github.event_name != 'merge_group'
     steps:
       - name: Verify all dependencies passed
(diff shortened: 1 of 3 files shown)


falkoschindler (commented):
Thanks for the thorough groundwork here, @evnchn! The three empirical `workflow_dispatch` runs and the real red→green of #6159 on Windows are exactly the evidence this kind of CI change needs. Direction and execution are good; nothing blocking. The job mirrors `_test.yml` conventions (action versions, cache glob, `uv sync --locked`, oldest Python), and keeping it inline instead of parameterizing `_test.yml` with an `os` input is the right call given the scope mismatch.

Findings:

1. **No failure-screenshot upload** (minor) — `_test.yml` uploads `screenshots/**/*.failed.png` via `actions/upload-artifact@v7` on failure; the Windows job doesn't. For the target class of bug (file-lock `PermissionError`) the traceback tells the whole story, so this isn't load-bearing — but since it's just the six-line step from `_test.yml` (guarded by `failure()`, e.g. `name: screenshots-windows`), it might be worth copying over for the occasional browser-side red. Take or leave.

2. **Unpinned ephemeral dependency** (minor) — `uv run --with pytest-rerunfailures` resolves the latest version on every run, outside `uv.lock`. A breaking plugin release would redden the job with no repo change to bisect. Pinning it in the command (e.g. `--with pytest-rerunfailures==16.0.1`) keeps the "no lockfile change" property while staying reproducible.

3. **Blanket reruns can mask an intermittent real regression** (minor) — `--reruns 2` retries _any_ failure, including the very class of intermittent file-lock race this leg exists to catch (a deterministic red like #6159's reverted-fix test is not affected). `pytest-rerunfailures` supports `--only-rerun <regex>`; restricting to the two known flake signatures, e.g. `--only-rerun "TimeoutException|ConnectionResetError"`, would let an unexpected `PermissionError` fail loudly instead of possibly being retried green.

4. **`timeout-minutes: 40` is generous for a ~73 s job** (minor) — a wedged Selenium session would burn 40 min of Windows runner before cancellation. It matches the `_test.yml` convention, so leave-or-take, but 10–15 min is plenty here.

5. **Session-scoped driver vs. reruns** (observation, no action needed) — `nicegui_driver` is session-scoped, so a rerun reuses the same Chrome instance that just timed out; recovery relies on `_reset_browser_state()` at the next fixture setup. Empirically that worked in your runs — just noting that a hard renderer crash could poison the remaining tests and their reruns alike.

On the three design knobs:

- **`windows-2022` pin** — agree, keep. Oldest supported image, one stable entry, no silent drift.
- **Per-PR vs. merge queue** — as written, the merge queue skips the job, so a Windows-only _semantic_ conflict between two individually-green PRs can land unseen. The job is ~1–2 min; please drop the `if` and run it in the merge queue too — that actually closes the blind spot the PR describes.
- **Informational vs. gating** — keep it informational for now. The rerun net is only validated by two clean runs; if the residual flake rate is even 1–3%, a gate would spuriously block one in every 30–100 PRs, and a gate that cries wolf gets `re-run` clicked reflexively — losing exactly the attention it was meant to buy. A few weeks of informational running turns the flake rate into a measured number for free; then we wire it into the gates. Fixing the teardown `ConnectionResetError` deterministically (e.g. a targeted warning filter) and scoping reruns per finding 3 would clear the path to gating sooner.

With the merge-queue `if` dropped, this is good to come out of draft.

evnchn:

**Both resolved — by removing the surface, not by adding the pieces.** The job now runs only the non-browser storage tests, so neither finding has anything left to apply to.

<details>
<summary><b>What changed</b></summary>

Selection is now `-m "not screen"` against a marker applied automatically to any test using the browser fixture, so the list can't rot:

```python
def pytest_collection_modifyitems(items) -> None:
    for item in items:
        if 'screen' in getattr(item, 'fixturenames', ()):
            item.add_marker('screen')
```

- **Failure-screenshot upload** — no browser tests remain in this job, so there is nothing to screenshot; the six-line step could never fire.
- **Unpinned `pytest-rerunfailures`** — gone. The `uv run --with pytest-rerunfailures` invocation you flagged no longer exists; the job runs `uv sync --locked` then plain `uv run pytest`, so every dependency comes from the lockfile with no ephemeral resolution on the runner.

```yaml
  windows-test:
    name: windows / storage (non-browser)
    runs-on: windows-2022
    timeout-minutes: 15
    steps:
      - uses: actions/checkout@v7
      - uses: astral-sh/setup-uv@v7
        with:
          enable-cache: true
          cache-dependency-glob: "uv.lock"
      - name: Install Python
        run: uv python install 3.10
      - name: Install dependencies
        run: uv sync --locked
      - name: pytest tests/test_storage.py
        run: uv run pytest tests/test_storage.py -m "not screen"
```

CI on the current head: `windows / storage (non-browser)` passes in 46s, all `check` jobs green. Narrowing the scope also cuts the leg's runtime, which seems the right trade for a job whose target class of bug is a file-lock `PermissionError`.
</details>

falkoschindler (approved):
Nice, focused CI addition — closes a real blind spot: the `-m "not screen"` filter selects exactly the file-race/teardown tests from #6159 that were proven red→green on real Windows, and the auto-applied `screen` marker keeps the selection mechanism-based instead of a hand-maintained list. With the job now wired into both gates and the flaky browser tests excluded at the root, a Windows storage regression actually blocks merging. The job runs in well under a minute, so the extra gate is essentially free.

PR #6162, opened 2026-07-21
Fix TypeError in shadow text when autocomplete list is absent

### LINKED ISSUE #6161: shadowText computed throws TypeError when the autocomplete list is absent
### First Check

### Example Code

```python
from nicegui import ui

inp = ui.input(value='f', autocomplete=['foo'])
ui.button('Remove autocomplete', on_click=lambda: inp.set_autocomplete(None))

ui.run()
```

Click the button: the browser console logs a SEVERE error on every subsequent render of the (filled) input:

```
TypeError: Cannot read properties of null (reading 'find')
    at Proxy.shadowText (input.js:48:49)
```

### Description

In `input.js`, the `shadowText` computed calls `this._autocomplete.find(...)` unguarded, while the sibling `withDatalist` computed defensively checks `this._autocomplete && this._autocomplete.length > 0`. The `_autocomplete: Array` prop has no default, so whenever it is `null`/`undefined` and the input holds text, every render throws.

Two paths get there:

- **Deterministic**: `Input.__init__` normalizes with `autocomplete or []`, but `set_autocomplete(None)` assigns `None` to the prop directly (`input.py`).
- **Transient**: props are written during `beforeUnmount` (#5652), so a re-render evaluating computeds mid-teardown can see the prop undefined. We hit this as a rare flake in CI screen tests — the console error fired while tearing down a page with filled inputs, and the screen fixture's console-error check failed the test.

Fix in PR: guard the computed the same way `withDatalist` already does.

*Investigated and written by Claude (Claude Code) on behalf of @Jepson2k.*

Jepson2k:
### Motivation

Fixes #6161.

The `shadowText` computed in `input.js` calls `this._autocomplete.find(...)` unguarded. The `_autocomplete: Array` prop has no default, so whenever it is `null`/`undefined` while the input holds text, every render logs a SEVERE `TypeError: Cannot read properties of null (reading 'find')`. Reachable deterministically via `set_autocomplete(None)` (which, unlike `Input.__init__`, does not normalize `None` to `[]`), and transiently during unmount re-renders since props are written in `beforeUnmount` (#5652) — the latter surfaced as a rare console-error flake in our downstream CI screen tests.

### Implementation

Guard the computed with optional chaining: `this._autocomplete?.find(...)`. A missing list then simply means "no shadow text", consistent with the sibling `withDatalist` computed treating it as "no datalist".

For the regression, `test_autocompletion` gained a final step: it removes the autocomplete list via `set_autocomplete(None)` and asserts that the shadow text disappears. The crash triggers on the prop update alone, so no separate screen test is needed — without the guard, the screen fixture fails the test on the logged SEVERE console error (verified red locally). Runs on Linux CI (unlike the race in #6158, this failure mode is platform-independent).

### Progress

*Investigated and written by Claude (Claude Code) on behalf of @Jepson2k.*

### DIFF
--- a/nicegui/elements/input.js
+++ b/nicegui/elements/input.js
@@ -45,7 +45,7 @@ export default {
   computed: {
     shadowText() {
       if (!this.inputValue) return "";
-      const matchingOption = (this._autocomplete ?? []).find((option) =>
+      const matchingOption = this._autocomplete?.find((option) =>
         option.toLowerCase().startsWith(this.inputValue.toLowerCase()),
       );
       return matchingOption ? matchingOption.slice(this.inputValue.length) : "";
--- a/nicegui/elements/input.js
+++ b/nicegui/elements/input.js
@@ -45,7 +45,7 @@ export default {
   computed: {
     shadowText() {
       if (!this.inputValue) return "";
-      const matchingOption = this._autocomplete.find((option) =>
+      const matchingOption = (this._autocomplete ?? []).find((option) =>
         option.toLowerCase().startsWith(this.inputValue.toLowerCase()),
       );
       return matchingOption ? matchingOption.slice(this.inputValue.length) : "";
--- a/nicegui/elements/input.py
+++ b/nicegui/elements/input.py
@@ -80,7 +80,7 @@ class Input(LabelElement, ValidationElement[str | None], DisableableElement, com
 
     def set_autocomplete(self, autocomplete: list[str] | None) -> Self:
         """Set the autocomplete list."""
-        self._props['_autocomplete'] = autocomplete
+        self._props['_autocomplete'] = autocomplete or []
         return self
 
     @property
--- a/tests/test_input.py
+++ b/tests/test_input.py
@@ -202,23 +202,9 @@ def test_autocompletion(screen: Screen):
     element.send_keys('o')
     screen.should_contain('nce')
 
-
-def test_removing_autocompletion_from_filled_input(screen: Screen):
-    input_ = None
-
-    @ui.page('/')
-    def page():
-        nonlocal input_
-        input_ = ui.input('Input', value='f', autocomplete=['foo'])
-
-    screen.open('/')
-    # used to raise a TypeError in the shadowText computed (rendered with a filled value and no autocomplete list)
-    input_.set_autocomplete(None)
-    screen.wait(0.5)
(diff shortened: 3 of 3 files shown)


evnchn:

**Optional nit (non-blocking):** consider `this._autocomplete?.find(...)` over `(this._autocomplete ?? []).find(...)`.

Equivalent here — the next line's `matchingOption ? … : ""` already absorbs `undefined` — and it drops the throwaway `[]` allocation on the null path. Purely stylistic; the fix is correct either way.

<details><summary><b>Why it's equivalent (all cases)</b></summary>

`.find()` yields an element or `undefined`, and the downstream truthy check handles both forms identically:

| `_autocomplete` | `?? []` → `matchingOption` | `?.` → `matchingOption` | returned |
|---|---|---|---|
| `null` / `undefined` | `[].find(...)` → `undefined` | short-circuits → `undefined` | `""` |
| `[]` | `undefined` | `undefined` | `""` |
| no match | `undefined` | `undefined` | `""` |
| match | the option | the option | `.slice(...)` |

The falsy-element edge (an `""` option) is unreachable: `if (!this.inputValue) return ""` guards the empty-input case, and a non-empty input can't `startsWith`-match `""`.

Side note: the description says this mirrors `withDatalist`, but that computed uses `this._autocomplete && this._autocomplete.length > 0` — a third idiom. All three are correct; there's no existing form to actually be consistent with.
</details>

falkoschindler (approved):
Thanks for the fix and the thorough investigation, @Jepson2k!

We pushed two small changes onto your branch:

- **Folded the regression test into the existing `test_autocompletion`:** The crash triggers on the prop update alone (no typing needed), and the screen fixture's console-error check catches it automatically, so a separate screen test with its own server start, element lookup and waits isn't necessary. Verified red without the guard, green with it.
- **Normalized `None` to `[]` in `set_autocomplete`,** matching what `Input.__init__` already does, so the `_autocomplete` prop is always an array on the server side.

One side note: we couldn't reproduce the "transient unmount" path from the issue description — `renderRecursively` always spreads the complete `element.props` into the component, and the `beforeUnmount` hook from #5652 only writes back `value`, leaving `_autocomplete` untouched. Our best guess is that your downstream CI flake was the deterministic `set_autocomplete(None)` path striking during a teardown re-render. No action needed — the guard is worth keeping either way — but if you did hit the error without `set_autocomplete(None)`, we'd be curious to hear about it.

PR #6164, opened 2026-07-22
Raise KeyError from ObservableDict.pop for a missing key, like dict.pop

### LINKED ISSUE #6163: ObservableDict.pop() (and app.storage.*) returns None for a missing key instead of raising KeyError

### Example Code

```python
from nicegui import app

# app.storage.general (and .user / .tab / .client) is an ObservableDict,
# so this is reachable straight from the public storage API:
print(repr(app.storage.general.pop('nonexistent')))   # prints: None
#                                                        a plain dict raises KeyError here

# minimal, no server needed:
d = {}
try:
    d.pop('missing')
except KeyError:
    print("plain dict  -> KeyError   (correct)")
print("ObservableDict ->", repr(ObservableDict().pop('missing')))   # -> None   (wrong)
```

### Description

`ObservableDict.pop()` breaks the `dict.pop()` contract: **popping a missing key with no default returns `None` instead of raising `KeyError`.**

The signature is `pop(self, k, d=None)`, so `super().pop(k, d)` is *always* called with a default — the method can never tell "no default given" from "explicit default of `None`", and therefore never raises. [Python's `dict.pop`](https://docs.python.org/3/library/stdtypes.html#dict.pop) is: missing key **with** default → return default; missing key **without** default → raise `KeyError`.

This is user-facing, not just internal: `app.storage.general`, `app.storage.user`, `app.storage.tab`, and `app.storage.client` are all `ObservableDict`s (via `PersistentDict(ObservableDict, ...)`), so `app.storage.user.pop('key_that_might_be_absent')` silently returns `None` where a normal dict would raise — a foot-gun for code that relies on the exception to distinguish "absent" from "present but `None`".

Expected: `app.storage.general.pop('nonexistent')` raises `KeyError`.
Actual: it returns `None`.

**Secondary symptom (same root cause):** `pop()` fires a change event even when nothing was removed. A no-op call still bumps `last_modified`, notifies observers, and can schedule a persistent-storage write / UI update:

```python
fired = []
d = ObservableDict(on_change=lambda: fired.append(1))
d.pop('missing', 'default')     # returns 'default', mutates nothing...
print(fired)                    # -> [1]   (a spurious change event)
```

Expected: `pop(k, default)` on a missing key mutates nothing → fires no change event.
Actual: one change event fires.

Note: this is **pre-existing** (not introduced by the recent observables rework in #6140 — confirmed reproducing identically on `main` and on that branch); surfaced while reviewing it. A fix PR will follow.

### NiceGUI Version

3.14.0 (also reproduces on `main` @ `8c51e15b`)

### Python Version

3.13.11

### Browser

Other — not browser-related; this is server-side Python behavior of `ObservableDict` / `app.storage`.

### Operating System

macOS (platform-independent — pure-Python logic in `nicegui/observables.py`)

### Additional Context

Root location: [`nicegui/observables.py`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/observables.py) — `ObservableDict.pop`. `ObservableList.pop` / `ObservableSet.pop` have different signatures and are unaffected. Internal callers that want tolerant removal already pass an explicit default (e.g. `props.py`), so raising `KeyError` for the no-default case aligns with `dict` without breaking them.

evnchn:

### Motivation

Fixes #6163. `ObservableDict.pop` breaks the `dict.pop` contract: with the signature `pop(self, k, d=None)`, `super().pop(k, d)` is always called with a default, so a missing key returns `None` instead of raising `KeyError`, and a no-op pop still fires a change event. This is user-facing — `app.storage.general` / `.user` / `.tab` / `.client` are all `ObservableDict`s. Pre-existing; surfaced while reviewing #6140.

### Implementation

Use a module-level `_MISSING` sentinel (same pattern already in `binding.py`) so `pop` can tell "no default given" from an explicit `None`, and only mutate/notify on a real removal:

- missing key, no default → `KeyError` (was: `None`)
- missing key, with default → returns default, no change event (was: fired one)
- present key → detaches the item (unchanged from #6140) and fires exactly one change event

Internal tolerant-removal callers already pass an explicit default (e.g. `props.py`), and `storage._users.pop(...)` is a plain `dict`, so nothing regresses. `ObservableList.pop` / `ObservableSet.pop` have different signatures and are untouched.

### Progress

- [x] No breaking changes to the public API — aligns `ObservableDict.pop` with `dict.pop`; a missing-key-without-default call that previously returned `None` now raises `KeyError` (the documented `dict` behavior).

Rebased onto current `main` (with #6140), so the removal path keeps `self._unobserve(item)`.

```text
tests/test_observables.py   18 passed   (incl. new test_pop_missing_key_raises_keyerror)
tests/test_storage.py + tests/test_binding.py   47 passed
pre-commit  all pass   ·   mypy ./nicegui  clean (245 files)   ·   pylint  10.00/10
```

Empirical (public storage surface, and #6140 detach preserved):

```text
app.storage.general.pop('nope')  -> KeyError        (was: None)
pop('x', 'def')                  -> 'def', 0 events
pop('present')                   -> value, 1 event, popped child detached (no further events)
pop('gone', None)                -> None
```

Fresh-lineage review (Codex): confirmed it's a genuine contract bug (not a deliberate convention — internal callers pass explicit defaults), recommended fixing both the `KeyError` contract and the spurious event, and proposed the try/except shape used here.
</details>

### DIFF
--- a/nicegui/observables.py
+++ b/nicegui/observables.py
@@ -11,6 +11,8 @@ from typing_extensions import Self
 
 from . import events, helpers
 
+_MISSING = object()
+
 
 class ObservableCollection(abc.ABC):  # noqa: B024
 
@@ -141,8 +143,13 @@ class ObservableDict(ObservableCollection, dict):
         for key, value in self.items():
             super().__setitem__(key, self._observe(value))
 
-    def pop(self, k: Any, d: Any = None) -> Any:
-        item = super().pop(k, d)
+    def pop(self, k: Any, d: Any = _MISSING) -> Any:
+        try:
+            item = super().pop(k)
+        except KeyError:  # nothing was removed, so skip _unobserve/_handle_change below
+            if d is _MISSING:
+                raise
+            return d
         self._unobserve(item)
         self._handle_change()
         return item
--- a/tests/test_observables.py
+++ b/tests/test_observables.py
@@ -4,6 +4,8 @@ import gc
 import pickle
 import weakref
 
+import pytest
+
 from nicegui import ui
 from nicegui.observables import ObservableDict, ObservableList, ObservableSet
 from nicegui.testing import Screen, User
@@ -306,3 +308,16 @@ def test_nested_observables_are_picklable():
     root = ObservableList(restored, on_change=increment_counter)
     root[0]['x'] = 1
     assert count == 1, 'a restored tree should wire up freshly and notify exactly once'
+
+
+def test_pop_missing_key_raises_keyerror():
+    reset_counter()
+    data = ObservableDict({'a': 1}, on_change=increment_counter)
+    with pytest.raises(KeyError):
+        data.pop('missing')  # like dict.pop, a missing key without default raises
+    assert count == 0, 'a failed pop must not fire a change event'
+    assert data.pop('missing', 'default') == 'default'
+    assert count == 0, 'popping a missing key with a default must not fire a change event'
+    assert data.pop('a') == 1
+    assert count == 1, 'popping an existing key fires exactly one change event'
+    assert data.pop('a', None) is None and count == 1, 'default is returned without an event when key is gone'
--- a/tests/test_observables.py
+++ b/tests/test_observables.py
@@ -320,4 +320,5 @@ def test_pop_missing_key_raises_keyerror():
     assert count == 0, 'popping a missing key with a default must not fire a change event'
     assert data.pop('a') == 1
     assert count == 1, 'popping an existing key fires exactly one change event'
-    assert data.pop('a', None) is None and count == 1, 'default is returned without an event when key is gone'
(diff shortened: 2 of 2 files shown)


falkoschindler (approved):
Thanks for catching and fixing this, Evan!
The sentinel + try/except shape is exactly right and consistent with `popitem` and `__delitem__`, which also raise before notifying.

Verified locally: tests, mypy and pylint pass, and all internal single-argument `pop` call sites are safe. I split the combined assert in the last test line into two directly on the branch.

While reviewing I noticed that `pop` isn't the only method firing change events for no-op mutations — `setdefault` on an existing key, `update({})`, `ObservableSet.discard(missing)` etc. do the same. That's a broader change with its own risks, so I opened a follow-up issue instead of expanding this PR: #6169.

PR #6167, opened 2026-07-22
Fix `run.cpu_bound` hanging when the process pool uses the forkserver start method

### LINKED ISSUE #6166: `run.cpu_bound` hangs on Python 3.14 + Linux, and with `process_pool_start_method="forkserver"` on Python >= 3.13: the forkserver daemon re-imports the app and starts a second server
### First Check

### Example Code

```python
import multiprocessing
from nicegui import app, run, ui

print(f'TOP-LEVEL EXECUTED in process named {multiprocessing.current_process().name!r} '
      f'(start method: {multiprocessing.get_start_method()})')

@app.on_startup
async def main() -> None:
    print('cpu_bound returned:', await run.cpu_bound(pow, 6, 2))
    app.shutdown()

ui.run(port=8080, reload=False, show=False)
```

No configuration and no flags. On Linux + Python 3.14 this **hangs forever** and `TOP-LEVEL EXECUTED` prints **twice**. A clean run prints it once and exits 0 with `cpu_bound returned: 36`.

### Description

**TL;DR — when the `cpu_bound` pool uses the `forkserver` start method, the forkserver re-imports the user's script and starts a *second* NiceGUI server, then `run.cpu_bound` never returns and the process hangs. The re-entrancy guard in `ui_run.py` fails to stop it because CPython 3.13 changed the process name during the forkserver's `__main__` preload from `ForkServerProcess-N` to `MainProcess`. On CPython 3.14 this is reached with no configuration at all, because 3.14 made `forkserver` the default start method on POSIX.**

This is the `run.cpu_bound` sibling of #6062, which was fixed for **native mode** by pinning a spawn context (#6045 / #6065). `native/native.py` got its `SPAWN_CONTEXT`; the `cpu_bound` pool never did, so the same class of breakage is still reachable there.

#### Root cause

`forkserver` preloads `__main__`, so `forkserver.ensure_running()` re-imports the user's script. NiceGUI defends against exactly this with the guard in `ui_run.py`:

```python
if multiprocessing.current_process().name != 'MainProcess':
    return
```

Under `spawn` that works, because `spawn.prepare()` sets `current_process().name = 'SpawnProcess-N'` *before* the main-module fixup. Under `forkserver` it used to work too — but **observed CPython behaviour changes between 3.12 and 3.13** in a way that defeats it. (I could not find a documented changelog entry for this specific naming/preload difference, so I am reporting it as observed behaviour rather than citing a CPython PR.)

Minimal proof, no NiceGUI involved at all. Three workers are started, which discriminates a one-off daemon preload from a per-worker import:

```python
import multiprocessing, os, sys
print(f'IMPORT __name__={__name__!r} pid={os.getpid()} procname={multiprocessing.current_process().name!r}', flush=True)
def f(x): return x
if __name__ == '__main__':
    ctx = multiprocessing.get_context('forkserver')
    for i in range(3):
        p = ctx.Process(target=f, args=(i,)); p.start(); p.join()
    print('py', sys.version.split()[0], 'done', flush=True)
```

```
----- python:3.12-slim -----
IMPORT __name__='__main__'     pid=1  procname='MainProcess'
IMPORT __name__='__mp_main__'  pid=9  procname='ForkServerProcess-1'   <-- per WORKER (3 of them)
IMPORT __name__='__mp_main__'  pid=10 procname='ForkServerProcess-2'
IMPORT __name__='__mp_main__'  pid=11 procname='ForkServerProcess-3'

----- python:3.13-slim -----
IMPORT __name__='__main__'     pid=1  procname='MainProcess'
IMPORT __name__='__mp_main__'  pid=8  procname='MainProcess'           <-- ONCE, in the forkserver DAEMON
```

The difference is not cosmetic:

- **3.12** — the forkserver daemon does not import `__main__`; each *worker* imports it under `ForkServerProcess-N`. NiceGUI's guard sees a non-`MainProcess` name and bails. Safe.
- **3.13+** — the forkserver *daemon* preloads `__main__` exactly once (independent of worker count) while still named `MainProcess`. The guard does not fire, so the user's top-level `ui.run()` runs to completion **inside the long-lived forkserver daemon**, which starts a second uvicorn on the same port and never reaches its accept loop.

Note the re-import happens under `__name__ == '__mp_main__'`, so the widened guard NiceGUI itself recommends in `nicegui/nicegui.py` (`{'__main__', '__mp_main__'}`) does not protect users either.

The MRE prints the process name deliberately, so the second `TOP-LEVEL EXECUTED` line showing `'MainProcess'` *is* the root cause made visible.

#### Two ways to reach it

1. **No opt-in at all — Linux + Python 3.14.** CPython 3.14 changed the default start method on POSIX (except macOS) from `fork` to `forkserver`, and the `cpu_bound` pool inherits the platform default. Released 3.14.0 builds `ProcessPoolExecutor(initializer=_ignore_sigint)` with no `mp_context` (`run.py:28`); `main` returns `multiprocessing.get_context()` from `_resolve_cpu_bound_context()` when `process_pool_start_method is None`. Both hang.
2. **`run.process_pool_start_method = 'forkserver'`, CPython >= 3.13, any OS.** New on `main` via #6117 — and currently *recommended by the docs that PR added* (`website/documentation/content/section_action_events.py:68`: *"...or to `"forkserver"` if you prefer it"*).

#### Version bisect

Measured, one row per real run. `TOP` = how many times the script's top level executed (1 = healthy, 2 = bug).

| Python | OS | NiceGUI | `process_pool_start_method` | effective method | result |
|---|---|---|---|---|---|
| 3.12.13 | Linux | 3.14.0 | *(default)* | fork | PASS — exit 0, TOP=1 |
| 3.13.14 | Linux | 3.14.0 | *(default)* | fork | PASS — exit 0, TOP=1 |
| 3.14.6 | Linux | 3.14.0 | *(default)* | **forkserver** | **HANG — exit 124, TOP=2**, `ConnectionResetError: [Errno 104]` |
| 3.14.6 | Linux | `main` 034db3b7 | *(default)* | **forkserver** | **HANG — exit 124, TOP=2**, `ConnectionResetError` |
| 3.12.13 | Linux | `main` 034db3b7 | `'forkserver'` | forkserver | PASS — exit 0, TOP=1, `returned: 36` |
| 3.13.14 | Linux | `main` 034db3b7 | `'forkserver'` | forkserver | **HANG — exit 124, TOP=2** |
| 3.14.6 | Linux | `main` 034db3b7 | `'forkserver'` | forkserver | **HANG — exit 124, TOP=2** |
| 3.13.11 | macOS 26.5 arm64 | `main` 034db3b7 | `'forkserver'` | forkserver | **HANG — exit 124, TOP=2**, `RuntimeError: did not receive acknowledgement of fd` |
| 3.13.11 | macOS 26.5 arm64 | `main` 034db3b7 | `'spawn'` | spawn | PASS — exit 0, TOP=1, `returned: 36` |

Two independent boundaries: **CPython 3.12 -> 3.13** defeats the guard, and **CPython 3.13 -> 3.14** makes it reachable with zero opt-in. It is therefore **not a NiceGUI regression** — released 3.14.0 fails identically on Python 3.14 — but `main` newly documents `'forkserver'` as a supported choice, which makes it reachable on 3.13 too.

#### Suggested fix

Drop the `__main__` preload for the resolved context in `run.setup()`:

```python
_pool_context = _resolve_cpu_bound_context()
if _pool_context.get_start_method() == 'forkserver':
    _pool_context.set_forkserver_preload([])
```

Keying on the **resolved context** rather than on the `process_pool_start_method` string is the part that matters — it covers the explicit `'forkserver'` setting *and* the implicit Python 3.14 default, which are the two entry points above. It must also run before the forkserver starts, which `setup()` does.

To be precise about what this fixes: it stops the **daemon** from importing `__main__`. Workers still import it — the same probe with `set_forkserver_preload([])` on 3.13 and 3.14 gives back the 3.12 shape, three imports named `ForkServerProcess-1/2/3` — but there NiceGUI's guard fires correctly. So the fix restores 3.12-equivalent behaviour rather than eliminating re-execution outright.

Verified: with this patch applied to `main`, the MRE returns `cpu_bound returned: 36`, prints `NiceGUI ready to go` exactly once, and exits 0; `pytest tests/test_run.py` stays green (14 passed, 3 skipped).

Two caveats worth weighing: it forfeits forkserver's preload optimisation (documented as a performance feature, not semantics), and because the public API only *sets* the whole list, it would clobber any preload list a user or library had configured. If that matters, the policy could be narrowed to "remove `__main__` from the list" — though that needs private-attribute inspection today.

Hardening the `ui_run.py` guard instead looks worse: bailing inside `ui.run()` is too late, because the forkserver would already have executed the module's top-level route registration and other import side effects.

#### Test gap

`tests/test_run.py::test_pool_uses_configured_start_method` does parametrise `'forkserver'`, but only asserts `run.process_pool._mp_context is not None` — it never submits a job, so a completely non-functional pool passes green. A regression test should actually `await run.cpu_bound(...)` under each start method. Related: the fork-path tests in that file `pytest.skip` off Linux, so on a macOS dev box none of the start-method behaviour is exercised end-to-end.

### NiceGUI Version

3.14.0 (released) and `main` @ 034db3b7

### Python Version

3.14.6 and 3.13.14 (Linux); 3.13.11 (macOS). 3.12.13 is unaffected.

### Browser

None / not applicable

### Operating System

Linux (Docker `python:3.1x-slim`) and macOS 26.5.2 arm64

### Additional Context

The MRE is deliberately self-checking: it uses `app.on_startup` + `app.shutdown()` so there is no HTTP client and no manual kill involved. A healthy run exits 0; the bug shows up as a hang plus a doubled `TOP-LEVEL EXECUTED` line.

One anticipated objection: "top-level `ui.run()` without an `if __name__ == '__main__'` guard is simply not multiprocessing-safe under spawn/forkserver." That is true in general, but I do not think it disposes of this: NiceGUI documents and encourages top-level `ui.run()`, `run.cpu_bound` is a NiceGUI-provided feature, NiceGUI ships its own guard for exactly this scenario, and the guard silently stopped working. The user-visible result on stock Python 3.14 + Linux is a hang with no actionable error.

Happy to open a PR with the fix plus a regression test that actually submits a job under each start method, if that shape looks right to you.

evnchn:
### Motivation

Fixes #6166.

When the `cpu_bound` pool uses the `forkserver` start method, the forkserver daemon re-imports the user's script, a second NiceGUI server starts, and `run.cpu_bound` never returns — the process hangs.

Two ways to reach it:

- **No opt-in at all, Linux + Python 3.14.** CPython 3.14 changed the default start method on POSIX (except macOS) from `fork` to `forkserver`, and the pool inherits the platform default via `_resolve_cpu_bound_context()`. Also present in released 3.14.0.
- **`run.process_pool_start_method = 'forkserver'` on CPython >= 3.13**, any OS — the option added in #6117, which `section_action_events.py` currently suggests.

The reason our existing defence misses it: `ui_run.py` guards re-entrancy with `current_process().name != 'MainProcess'`. Under `spawn` that works, because `spawn.prepare()` renames the process before the main-module fixup. Under `forkserver` on CPython >= 3.13 the **daemon** preloads `__main__` while still named `MainProcess`, so the guard does not fire. On 3.12 the daemon does not preload at all — workers import as `ForkServerProcess-N`, the guard fires, and everything is fine.

<details>
<summary>Version matrix (each row a real run; <code>TOP</code> = times the script's top level executed, 1 healthy / 2 bug)</summary>

| Python | OS | NiceGUI | `process_pool_start_method` | effective | result |
|---|---|---|---|---|---|
| 3.12.13 | Linux | 3.14.0 | *(default)* | fork | PASS — exit 0, TOP=1 |
| 3.13.14 | Linux | 3.14.0 | *(default)* | fork | PASS — exit 0, TOP=1 |
| 3.14.6 | Linux | 3.14.0 | *(default)* | **forkserver** | **HANG — exit 124, TOP=2** |
| 3.14.6 | Linux | main | *(default)* | **forkserver** | **HANG — exit 124, TOP=2** |
| 3.12.13 | Linux | main | `'forkserver'` | forkserver | PASS — exit 0, TOP=1 |
| 3.13.14 | Linux | main | `'forkserver'` | forkserver | **HANG — exit 124, TOP=2** |
| 3.14.6 | Linux | main | `'forkserver'` | forkserver | **HANG — exit 124, TOP=2** |
| 3.13.11 | macOS | main | `'forkserver'` | forkserver | **HANG — exit 124, TOP=2** |
| 3.13.11 | macOS | main | `'spawn'` | spawn | PASS — exit 0, TOP=1 |

</details>

### Implementation

Clear the forkserver preload list in `setup()` so the daemon does not import `__main__`.

The condition is on the **resolved context**, not on `process_pool_start_method`, so it covers both entry points above — the explicit setting *and* the implicit Python 3.14 default. It sits in `setup()` because the preload must be configured before the daemon starts; `ProcessPoolExecutor(mp_context=...)` does not start it (the first `submit()` does), so this is early enough.

**To be precise: this does not stop `__main__` being re-imported — it relocates it.** Instead of the daemon importing it once (as `MainProcess`, invisible to the guard), each worker imports it in `spawn.prepare()` → `_fixup_main_from_path` as `ForkServerProcess-N`, where the existing `ui_run.py` guard fires. Net effect: **`forkserver` now behaves exactly like `spawn`.**

```
### DEFAULT preload ['__main__']   (buggy)
TOPLEVEL RAN  __name__='__main__'     procname='MainProcess'
TOPLEVEL RAN  __name__='__mp_main__'  procname='MainProcess'          <-- the daemon; guard misses it
### preload CLEARED               (this PR)
TOPLEVEL RAN  __name__='__main__'     procname='MainProcess'
TOPLEVEL RAN  __name__='__mp_main__'  procname='ForkServerProcess-1'  <-- a worker; guard fires
```

That framing matters for two follow-on questions. Pickling is unaffected — a `cpu_bound` callable defined in the user's `main.py` still resolves, because `_fixup_main_from_path` aliases `sys.modules['__main__']`. And the cost of losing the preload is that each worker re-imports the user's module rather than the daemon doing it once — which is precisely `spawn`'s existing cost, not a new class of overhead. The `ui_run.py` guard remains load-bearing after this change; it is what makes the worker-side import harmless.

**Trade-off worth flagging:** `set_forkserver_preload()` mutates a process-global (`multiprocessing.forkserver._forkserver`), not a per-context value, so it would wipe a preload list a user or another library had configured before startup. The public API only offers a setter, so a narrower "remove just `__main__`" variant would need private state in production code. Given that the alternative is a hung app with no actionable error, clearing looks like the right call — but happy to go the narrower route if you prefer. Note also that this only affects *future* forkserver launches: if something else already started the daemon before `setup()`, we cannot retune it.

The perf cost should be negligible — CPython's default preload is `['__main__']`, not heavy modules — except for users who deliberately configured a preload list.

<details>

**The new test fails without the fix and passes with it** (not just green-on-green):

```
$ git stash push nicegui/run.py && pytest tests/test_run.py -k forkserver
FAILED tests/test_run.py::test_forkserver_pool_does_not_preload_main - AssertionError: assert ['__main__'] == []
1 failed, 1 passed, 16 deselected
```

**Local gates:** `pre-commit` all passed · `mypy ./nicegui` success, 245 files · `pylint ./nicegui` 10.00/10 · `pytest tests/test_run.py` 15 passed, 3 skipped.

**Empirical, with the MRE from #6166** (`app.on_startup` + `app.shutdown()`, so no HTTP client involved):

| scenario | before | after |
|---|---|---|
| macOS py3.13, explicit `'forkserver'` | hang (exit 124), TOP=2, 2× "ready to go" | exit 0, TOP=1, 1× "ready to go", `returned: 36` |
| Linux py3.14, **zero config**, wheel built from this branch | hang (exit 124), TOP=2, `ConnectionResetError` | exit 0, TOP=1, 1× "ready to go", `returned: 36` |

The `isolate_pool_state` fixture now also saves/restores the forkserver preload list, since both this test and the existing `test_pool_uses_configured_start_method[forkserver]` mutate that global via `setup()`.

</details>

### Progress

---

*Staged first on the fork — evnchn/nicegui#212 — which is also green on the identical commit. (This PR was opened while `quick-test` was still running; it has since passed.)*

### DIFF
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -69,6 +69,9 @@ def setup() -> None:
     """Setup the process pool. (For internal use only.)"""
     global process_pool, _pool_context, _pool_uses_implicit_fork  # pylint: disable=global-statement # noqa: PLW0603
     _pool_context = _resolve_cpu_bound_context()
+    if _pool_context.get_start_method() == 'forkserver':
+        # let the workers re-import __main__, where the MainProcess guard catches it, not the daemon (#6166)
+        _pool_context.set_forkserver_preload([])
     _pool_uses_implicit_fork = process_pool_start_method is None and _pool_context.get_start_method() == 'fork'
     try:
         process_pool = ProcessPoolExecutor(mp_context=_pool_context, initializer=_ignore_sigint)
--- a/nicegui/run.py
+++ b/nicegui/run.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import multiprocessing
+import multiprocessing.forkserver
 import signal
 import traceback
 from collections.abc import Callable
@@ -70,8 +71,11 @@ def setup() -> None:
     global process_pool, _pool_context, _pool_uses_implicit_fork  # pylint: disable=global-statement # noqa: PLW0603
     _pool_context = _resolve_cpu_bound_context()
     if _pool_context.get_start_method() == 'forkserver':
-        # let the workers re-import __main__, where the MainProcess guard catches it, not the daemon (#6166)
-        _pool_context.set_forkserver_preload([])
+        # drop __main__ from the preload so the workers re-import it, where the MainProcess guard catches it,
+        # not the daemon (#6166); other (user-configured) preload modules are kept
+        fork_server = multiprocessing.forkserver._forkserver  # pylint: disable=protected-access
+        preload = fork_server._preload_modules  # type: ignore[attr-defined]  # pylint: disable=protected-access
+        _pool_context.set_forkserver_preload([module for module in preload if module != '__main__'])
     _pool_uses_implicit_fork = process_pool_start_method is None and _pool_context.get_start_method() == 'fork'
     try:
         process_pool = ProcessPoolExecutor(mp_context=_pool_context, initializer=_ignore_sigint)
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -1,5 +1,6 @@
 import asyncio
 import multiprocessing
+import multiprocessing.forkserver
 import os
 import time
 from collections.abc import Callable, Generator
@@ -116,13 +117,16 @@ async def test_returns_none_when_app_is_stopping(user: User, func: Callable):
 
 @pytest.fixture
 def isolate_pool_state() -> Generator[None, None, None]:
-    """Restore run.process_pool_start_method, drop the test's pool and forget shown warnings."""
+    """Restore run.process_pool_start_method and the forkserver preload, drop the pool and forget warnings."""
     original = run.process_pool_start_method
+    forkserver = multiprocessing.forkserver._forkserver  # pylint: disable=protected-access
+    original_preload = forkserver._preload_modules  # pylint: disable=protected-access
     nicegui_warnings.reset()
     try:
         yield
     finally:
         run.process_pool_start_method = original
+        forkserver.set_forkserver_preload(original_preload)
         run.reset()
         nicegui_warnings.reset()
 
@@ -139,6 +143,17 @@ def test_pool_uses_configured_start_method(method):
     assert run.process_pool is not None
     assert run.process_pool._mp_context is not None  # pylint: disable=protected-access
     assert run.process_pool._mp_context.get_start_method() == expected  # pylint: disable=protected-access
+
+
+@pytest.mark.usefixtures('isolate_pool_state')
+async def test_forkserver_pool_does_not_preload_main():
+    """A "forkserver" pool must not re-import __main__, which would re-run the user's ui.run() (#6166)."""
+    if 'forkserver' not in multiprocessing.get_all_start_methods():
+        pytest.skip('forkserver is not available on this platform')
+    run.process_pool_start_method = 'forkserver'
+    run.setup()
+    assert multiprocessing.forkserver._forkserver._preload_modules == []  # pylint: disable=protected-access
+    assert await run.cpu_bound(good_function)
 
 
 @pytest.mark.usefixtures('isolate_pool_state')
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -147,12 +147,13 @@ def test_pool_uses_configured_start_method(method):
 
 @pytest.mark.usefixtures('isolate_pool_state')
 async def test_forkserver_pool_does_not_preload_main():
-    """A "forkserver" pool must not re-import __main__, which would re-run the user's ui.run() (#6166)."""
+    """A "forkserver" pool must drop the __main__ preload (it would re-run ui.run()) but keep user modules (#6166)."""
     if 'forkserver' not in multiprocessing.get_all_start_methods():
         pytest.skip('forkserver is not available on this platform')
+    multiprocessing.set_forkserver_preload(['json', '__main__'])
     run.process_pool_start_method = 'forkserver'
     run.setup()
-    assert multiprocessing.forkserver._forkserver._preload_modules == []  # pylint: disable=protected-access
+    assert multiprocessing.forkserver._forkserver._preload_modules == ['json']  # pylint: disable=protected-access
     assert await run.cpu_bound(good_function)
 
 


evnchn:
**CI is green** — `quick-test / 3.10 / pytest` passed (19m41s), all `check` jobs green. The "opened while CI was still running" caveat in the description no longer applies; the description has been updated.

Nothing else changed since opening — same single commit, +19/−1.

falkoschindler (approved):
Excellent analysis and write-up, @evnchn! I verified the mechanism claims against the code and CPython's stdlib — they all hold: the `MainProcess` guard in `ui_run.py` is bypassed exactly as described, `run.setup()` runs before the forkserver daemon can start (it only launches on the first `submit()`), and `set_forkserver_preload` indeed writes the process-global singleton you flagged. Also confirmed the new test fails without the fix.

I took the liberty of pushing the narrower variant you offered in the description (bfe033be9): only `__main__` is dropped from the preload list, so a user-configured preload like `['numpy']` survives NiceGUI's startup — the test now covers that as well.

PR #6171, opened 2026-07-23
Skip change events for no-op mutations of observable collections

### LINKED ISSUE #6169: Observable collections fire change events for no-op mutations
## Example Code

```python

data = ObservableDict({'a': 1}, on_change=lambda: print('changed!'))
data.setdefault('a', 2)  # prints "changed!" although nothing changed
data.update({})          # prints "changed!" although nothing changed
```

## Description

#6164 fixed `ObservableDict.pop` so that a failed pop no longer fires a change event. But the same class of bug — a mutation method that doesn't actually mutate anything, yet still calls `_handle_change()` — exists across all three observable collections in `nicegui/observables.py`:

- `ObservableDict.setdefault(key)` with an existing key — `dict.setdefault` documents "insert if not present"; no insertion happens, but an event fires
- `ObservableDict.update({})` — nothing is assigned, an event fires
- `ObservableList.extend([])` and `+= []` — an event fires
- `ObservableSet.discard(missing)` — a documented no-op, an event fires
- `ObservableSet.add(existing)`, `update(subset)`, `intersection_update(superset)`, `difference_update(disjoint)`, `symmetric_difference_update(set())` — the set is unchanged, an event fires

Since `app.storage.general` / `.user` / `.tab` / `.client` are `ObservableDict`s, each spurious event also bumps `last_modified` and can trigger unnecessary storage persistence and UI refreshes.

## NiceGUI Version

main (3.14.0.post16.dev0+91023e29a)

## Python Version

3.12.9

## Browser

Other

## Operating System

macOS

## Additional Context

Open questions:

- Where to draw the line: structural no-ops (nothing added/removed/assigned) are clear-cut, but should `update({'a': 1})` with `self['a'] == 1` already, or `sort()` on an already-sorted list, also stay silent? Suggestion: only skip events for structural no-ops; value-equality checks are potentially expensive and would change `__setitem__` semantics.
- Risk: code may (accidentally) rely on a no-op mutation as a "touch" that bumps `last_modified` or triggers persistence. Needs a release-note line.

Suggested approach: guard each affected method with a cheap structural check (e.g. early return in `update` when the merged dict is empty, `if key in self: return self[key]` in `setdefault`, membership check in `discard`/`add`, length comparison after the `*_update` set operations) and add a test per method asserting the event count, analogous to `test_pop_missing_key_raises_keyerror`.

evnchn:

Fixes #6169. Each listed structural no-op now returns before `_handle_change()`, so `last_modified` and the parent-chain handlers stay untouched. Value-equality no-ops (`update({'a': 1})` when the value already matches, `sort()` on a sorted list) stay out of scope, as suggested in the issue.

Guards are the cheap ones suggested there: membership for `setdefault` / `add` / `discard`, emptiness for `dict.update` / `extend` / `symmetric_difference_update`, subset for `set.update`, length for `intersection_update` / `difference_update`. `|=`, `+=`, `&=`, `-=`, `^=` and `set.remove` already delegate to these, so they are covered with no extra code. `list *= 1` is guarded too — same class, one line.

Worth a release-note line: a no-op mutation no longer bumps `last_modified`, so it no longer triggers storage persistence or a tab-storage keep-alive.

### Progress

- [x] No breaking changes to the public API, other than the intended silencing above.

```text
tests/test_observables.py + test_storage.py + test_binding.py   71 passed
mypy ./nicegui  clean (245 files)   ·   pylint 10.00/10   ·   pre-commit all pass
```

The three new tests were confirmed to **fail with the fix reverted** — 3 / 3 / 6 spurious events for dict / list / set, matching the enumeration in the issue exactly.

Issue repro, and the nested case (a spurious child event wakes every parent handler):

```text
d.setdefault('a', 2); d.update({})   -> 0 events (was 2), d['a'] still 1, last_modified unchanged
root['child'].setdefault('x', 99)    -> 0 parent handlers fired (was 1)
real mutations                        -> still exactly 1 event each
```

One existing assertion in `test_observable_set` encoded the old behavior (`data.add(1)` on a set already containing `1` expected an event); its first mutation is now a real insertion.

`binding._set_attribute` traverses intermediate keys with `setdefault`, so binding into a nested `app.storage.*` path no longer fires an event per traversal step. The final assignment still fires.

**Fresh-lineage review (Codex), calibrated-adversarial prompt.** It confirmed the `difference_update` / `__imul__` / `symmetric_difference_update` guards sound, found no iterator double-consumption, and agreed skipping `_observe()` for non-inserted equal elements in `set.update` is correct rather than a missed registration. Two points taken up:

- It flagged that `setdefault`'s early return went through `self[key]`, which a subclass overriding `__getitem__` could intercept where `dict.setdefault` would not. `PersistentDict` is a real subclass, so this now uses `super().__getitem__`.
- It rated as MAJOR that `intersection_update`'s length guard can hide a swap of an equal-but-not-identical representative. Verified in CPython: real, but only when the two sets are *exactly* the same size, and it only changes which `__eq__`-equal object is stored — invisible to storage and binding, which serialize by value. Observable collections are unhashable, so they can never be set members and no `_observe` / `_unobserve` bookkeeping can be missed. Left as the length comparison suggested in the issue rather than paying for an identity check on every call; happy to tighten if you would rather it notify.

</details>

### DIFF
--- a/nicegui/observables.py
+++ b/nicegui/observables.py
@@ -162,6 +162,8 @@ class ObservableDict(ObservableCollection, dict):
 
     def update(self, *args: Any, **kwargs: Any) -> None:
         new_items = dict(*args, **kwargs)
+        if not new_items:
+            return
         old_values = [self[key] for key in new_items if key in self]
         super().update({key: self._observe(value) for key, value in new_items.items()})
         self._unobserve(*old_values)
@@ -174,9 +176,9 @@ class ObservableDict(ObservableCollection, dict):
         self._handle_change()
 
     def setdefault(self, __key: Any, __default: Any = None) -> Any:
-        if __key not in self:
-            __default = self._observe(__default)
-        item = super().setdefault(__key, __default)
+        if __key in self:
+            return super().__getitem__(__key)
+        item = super().setdefault(__key, self._observe(__default))
         self._handle_change()
         return item
 
@@ -217,7 +219,10 @@ class ObservableList(ObservableCollection, list):
         self._handle_change()
 
     def extend(self, iterable: Iterable) -> None:
-        super().extend([self._observe(item) for item in iterable])
+        items = [self._observe(item) for item in iterable]
+        if not items:
+            return
+        super().extend(items)
         self._handle_change()
 
     def insert(self, index: SupportsIndex, obj: Any) -> None:
@@ -277,6 +282,8 @@ class ObservableList(ObservableCollection, list):
     def __imul__(self, other: Any) -> Any:
         old_items = list(self)
         super().__imul__(other)
+        if len(self) == len(old_items):
+            return self
         self._unobserve(*old_items)
         self._handle_change()
         return self
@@ -295,6 +302,8 @@ class ObservableSet(ObservableCollection, set):
             super().add(self._observe(item))
 
     def add(self, item: Any) -> None:
+        if item in self:
+            return
         super().add(self._observe(item))
         self._handle_change()
 
@@ -304,9 +313,8 @@ class ObservableSet(ObservableCollection, set):
         self._handle_change()
 
     def discard(self, item: Any) -> None:
-        super().discard(item)
-        self._unobserve(item)
-        self._handle_change()
+        if item in self:
+            self.remove(item)
 
     def pop(self) -> Any:
         item = super().pop()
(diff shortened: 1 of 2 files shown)


falkoschindler (dismissed):
Nice and thorough work, @evnchn! The guards are exactly the cheap structural checks suggested in #6169, and the delegation paths (`|=`, `+=`, `&=`, `-=`, `^=`, `remove`) are covered for free.

I traced every guarded method and found no correctness issue: exception paths still fire no events, generators are consumed only once, and since observable collections are unhashable, no `_observe`/`_unobserve` bookkeeping can be skipped. The `intersection_update` length guard can hide an equal-but-not-identical representative swap, but that is invisible to by-value consumers like storage and binding — the cheap check is the right trade-off.

I pushed a small follow-up commit extending the same idea: `clear()` on an already-empty collection no longer fires, a comment documents the `intersection_update` trade-off, and a new test locks in that a no-op on a nested dict doesn't wake parent handlers.

Verified locally: 72 tests across observables/storage/binding pass, pylint 10.00, mypy and pre-commit clean.

PR #6174, opened 2026-07-23
Fix PermissionError when clearing storage on Windows

### LINKED ISSUE #6173: Windows: merged regression test test_unlinking_storage_files_waits_out_transient_holders fails with PermissionError (WinError 32)
*Filed by Claude Code on evnchn's behalf.*

### First Check

- [X] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [X] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty. (Closest are #6158 and #5949, both closed by #6159 — this is that fix's own regression test failing.)

### Example Code

```python
# tests/test_storage.py on main @ 9a088b2e — verbatim, no modification needed.
# Run on Windows:  pytest tests/test_storage.py -k unlinking_storage_files_waits_out_transient_holders

async def test_unlinking_storage_files_waits_out_transient_holders(user: User):
    @ui.page('/')
    def page():
        ui.label('ok')

    await user.open('/')  # needed to ensure NiceGUI's event loop is running
    filepath = Storage.path / 'storage-general.json'

    async def wait_until(condition: Callable[[], bool], *, timeout: float = 2.0) -> None:
        deadline = time.monotonic() + timeout
        while not condition() and time.monotonic() < deadline:
            await asyncio.sleep(0.01)

    async def write_and_hold_file() -> None:
        app.storage.general['key'] = 'value'  # schedules an async backup task
        await wait_until(filepath.exists)  # let the backup write the file
        assert filepath.exists()
        handle = filepath.open(encoding='utf-8')  # stands in for a backup write still holding the file
        threading.Timer(0.2, handle.close).start()

    await write_and_hold_file()
    app.storage.general.clear()  # schedules an async backup which deletes the now-empty file
    await wait_until(lambda: not filepath.exists())
    assert not filepath.exists()  # PASSES

    await write_and_hold_file()
    app.storage.clear()  # <-- FAILS HERE on Windows: PermissionError [WinError 32]
    assert not filepath.exists()
```

### Description

`test_unlinking_storage_files_waits_out_transient_holders` — the regression test merged with #6159 — **fails on Windows** on current `main`, with the same `PermissionError: [WinError 32]` it was written to prove fixed.

**Expected:** the test passes, as it does on Linux and macOS.
**Actual:** phase 1 (the async path, `app.storage.general.clear()`) passes; phase 2 (the sync path, `app.storage.clear()`) raises `PermissionError [WinError 32]` after the 1.0 s retry window expires.

**Reproduced 2/2 CI runs** on `windows-2022`, ~1.7 s, no browser involved. Ubuntu stays green because the race cannot occur on POSIX, and there is currently no Windows leg on CI — so this is invisible to the matrix today.

I have the failure reliably, but **not** a proven mechanism. Best hypothesis and the evidence against it are in Additional Context. Reporting rather than sending a PR, because the fix depends on which of two readings you take:

1. **Product** — `Storage.clear()` shouldn't depend on a loop it blocks. Options: drain/await pending backups before unlinking, offer an async `clear()` for callers already on the loop, or have the backup release its handle before the retry window.
2. **Test** — if sync `clear()` is intentionally best-effort against a loop-bound holder, phase 2 asserts more than the code promises and should be reshaped (and the `# used to raise PermissionError` comment then reads misleadingly).

Either way, worth noting the user-facing shape: `app.storage.clear()` is public and synchronous, and on Windows it can raise `PermissionError` when called from a running event loop with a backup in flight — the scenario #6158 and #5949 were about. If reading 2 is right, that limitation may deserve to be documented where users meet it, not only in a private helper's docstring.

Happy to send a PR in whichever direction you prefer — and happy to dig for the actual holder first if that's more useful.

### NiceGUI Version

`main` @ `9a088b2e` (`v3.14.0-26-g9a088b2e`)

### Python Version

3.10.20

### Browser

Other

*(No browser involved — this is a `user`-fixture test, in-process, no Selenium.)*

### Operating System

Windows

### Additional Context

<details><summary><b>Failure output (verbatim, from CI)</b></summary>

```
platform win32 -- Python 3.10.20, pytest-9.0.3, pluggy-1.6.0
collected 28 items / 26 deselected / 2 selected

tests/test_storage.py::test_awaiting_backup_scheduled_during_teardown PASSED  [ 50%]
tests/test_storage.py::test_unlinking_storage_files_waits_out_transient_holders FAILED [100%]

    await write_and_hold_file()
>   app.storage.clear()  # used to raise PermissionError (WinError 32) on Windows while the handle was open

tests\test_storage.py:452:
nicegui\storage.py:213: in clear
    helpers.unlink_with_retry(filepath, missing_ok=True)
nicegui\helpers\files.py:39: in unlink_with_retry
    while not _try_unlink(filepath, missing_ok=missing_ok, deadline=deadline):
nicegui\helpers\files.py:53: in _try_unlink
    filepath.unlink(missing_ok=missing_ok)

self = WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/nicegui-test-storage-vprccsut/storage-general.json')
missing_ok = True

E   PermissionError: [WinError 32] The process cannot access the file because it is
    being used by another process:
    'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\nicegui-test-storage-vprccsut\\storage-general.json'

================= 1 failed, 1 passed, 26 deselected in 1.63s ==================
```

Environment: `windows-2022`, Python 3.10.20, `uv sync --locked`, no reruns, no browser.

Two runs, identical apart from the tempdir name and 1.63 s vs 1.82 s. A third was cancelled by the workflow's `cancel-in-progress` concurrency group, not by a failure — so the sample is two, which is why I say "reproduced 2/2" rather than "deterministic".

</details>

<details><summary><b>What the evidence does and doesn't show</b></summary>

**What's established:** after 1.0 s of retrying, `filepath.unlink()` still raises `WinError 32`, so *something* held the file for the whole window.

**What isn't:** which holder. The obvious candidate — the explicit handle from `write_and_hold_file()` — is unlikely to be the one: it's closed by a `threading.Timer` on an independent thread, which shouldn't be blocked by the sleeping loop and fires at 0.2 s, comfortably inside the deadline.

**Best hypothesis:** a second, loop-bound holder — the `aiofiles` backup writer scheduled by `app.storage.general['key'] = 'value'`. `wait_until(filepath.exists)` only proves the file was *created*, not that the writer finished and closed it. If that coroutine is still open when phase 2 begins, sync `clear()` blocks the very loop it needs to resume, and the retry cannot succeed no matter how long the timeout.

That would be the case `nicegui/helpers/files.py` already documents, in the docstring added during #6159's review:

> Note: The blocking wait can only help if the holder releases the file independently of the current thread. A writer that needs this thread's event loop to close its file (e.g. an ``aiofiles`` coroutine suspended on the very loop that is calling this function) cannot proceed while we sleep and exhausts the timeout.

I can't confirm that from CI output alone — it would need a Windows box and a handle dump (or logging in the backup task) to nail down. Other explanations I can't fully exclude: the writer holding across a longer window than expected, or Windows-side scanning/indexing briefly retaining the file. I don't find those as likely, but I haven't ruled them out.

The earlier real-Windows red→green verification recorded in #6160 predates the two-phase consolidation of this test, so it doesn't cover this structure.

</details>

---

**Changelog**
- 2026-07-23: Restructured to follow the repository's issue template (First Check / Example Code / Description / versions / Additional Context). Content unchanged apart from adding the verbatim test source, the exact Python version, and the version string.

evnchn:

### Motivation

Fixes #6173.

On Windows, `app.storage.clear()` can raise `PermissionError: [WinError 32]`. This is visible today on `main`: the regression test merged with #6159, `test_unlinking_storage_files_waits_out_transient_holders`, **fails on Windows** — reproduced 2/2 on `windows-2022`, ~1.7 s, no browser. Ubuntu CI stays green because the race cannot occur on POSIX, and there is no Windows leg on CI yet (that's #6160).

The async backup held **the storage file itself** open across `await`s. That is what allowed a holder the retry could never wait out: a caller that blocks the event loop — synchronous `Storage.clear()` retrying via `time.sleep()` — can never let the writer resume to close its handle, so the retry window is spent waiting for something that structurally cannot happen. `nicegui/helpers/files.py` already documents exactly this case as unfixable by the blocking retry.

### Implementation

Write the backup to a temporary file and `os.replace()` it into place, so the storage file is never open while the backup runs. `os.replace` is atomic on POSIX and Windows.

```python
tmp = self.filepath.with_name(self.filepath.name + '.tmp')
async with aiofiles.open(tmp, 'w', encoding=self.encoding) as f:
    await f.write(dumps(self, str(self.filepath), indent=self.indent))
os.replace(tmp, self.filepath)
```

`Storage.clear()` then needs to know about the temporary file, or a leftover one makes the following `self.path.rmdir()` fail with *directory not empty*. It is swept separately and **deliberately without the retry**:

```python
for tmp_path in self.path.glob('storage-*.json.tmp'):
    with contextlib.suppress(OSError):  # never wait: only an in-flight backup on this loop can hold it
        tmp_path.unlink()
```

Trade-offs worth flagging:

- **The temp sweep must not use `unlink_with_retry`.** My first attempt widened the existing glob to `storage-*`, which reintroduced the original deadlock against the temp file. A temp file's only possible holder is our own loop-bound writer, so waiting can never help — best-effort is the correct treatment. The glob stays narrow (`storage-*.json.tmp`, not `storage-*`) so it cannot delete unrelated files an app may have placed under `Storage.path`.
- **`rmdir()` is left unsuppressed.** If a write is genuinely in flight, the temp survives and `clear()` raises *directory not empty* rather than silently reporting success. That preserves the existing "cleared, or failed" semantics — see residuals below.
- **Sync fallback path unchanged.** `write_text()` when no loop is running opens and closes without awaiting, so it cannot produce a loop-bound holder.

<details><summary><b>Side benefit: no more torn storage files</b></summary>

The old path truncated the real storage file with `open(..., 'w')` and then wrote. A crash, kill or full disk mid-write left a truncated or empty `storage-*.json` — silently losing storage, since `initialize()` catches the parse failure and only logs `Could not load storage file`. With the temp swap the real file is only ever replaced by a fully-written one. Not the goal, but it's the usual reason for this pattern.

</details>

**Windows** (`windows-2022`, Python 3.10.20) — the platform that matters, red→green on the same test, same command:

| | on `main` @ `9a088b2e` | with this change |
|---|---|---|
| `test_unlinking_storage_files_waits_out_transient_holders` | **FAILED** `PermissionError [WinError 32]`, 2/2 runs | **PASSED** |
| full `tests/test_storage.py` | — | **28 passed** |

**Linux / local:** `tests/test_storage.py` → `28 passed, 0 skipped`. `pre-commit`, `mypy ./nicegui` (245 files), `pylint ./nicegui` 10.00/10.

Full CI green on the identical tree on my fork (evnchn/nicegui#215, run `29994093204`): Linux `quick-test / 3.10 / pytest` pass, `windows / storage` pass, mypy/pylint/pre-commit pass.

⚠️ One honest note: a later Windows run of the same branch showed `27 passed, 2 errors`, both `ConnectionResetError` / `WinError 10054` — the separate, intermittent Windows **browser-test** flake, unrelated to storage and untouched by this PR. No `WinError 32` and no `WinError 145` in either run. So the 28/28 was partly luck on that flake; what both runs agree on is that the storage file-lock errors are gone.

</details>

<details><summary><b>Known residuals — not solved here</b></summary>

Surfaced by a second-model review pass; recording rather than hiding them:

- **Synchronous `clear()` still isn't bulletproof.** If a write backup is genuinely in flight, its temp cannot be removed (the loop is blocked, so the writer cannot finish) and `rmdir()` raises *directory not empty*. That is a clearer and faster failure than the old one-second deadlock, but still a failure — and it is reading 2 of #6173 reasserting itself: a synchronous `clear()` that blocks the loop cannot fully succeed against a loop-bound holder. I did not paper over it.
- **`os.replace` on Windows** can fail if a concurrent reader holds the destination without delete/rename sharing; `initialize()` / `initialize_sync()` are such readers. I could not verify CPython's exact share-mode behaviour, so I won't claim this trades away every race — only that it removes the one that reproduces.
- **Multi-process.** The fixed temp name is serialised per file *within one process* by `create_lazy(..., name=self.filepath.stem)`. Two processes pointed at the same `NICEGUI_STORAGE_PATH` would contend on it. Whether that's a supported configuration is your call.

</details>

### Progress

- [x] The PR title is a short phrase starting with a verb ("Fix ...").
- [x] Pytests are not necessary — the regression test already exists on `main` (added by #6159); this PR makes it pass on Windows. It was confirmed failing before the change and passing after, on a real Windows runner. Note it can only fail on Windows, so the Ubuntu matrix cannot guard this — that is what #6160 is for.

### DIFF
--- a/nicegui/persistence/file_persistent_dict.py
+++ b/nicegui/persistence/file_persistent_dict.py
@@ -1,4 +1,3 @@
-import os
 from pathlib import Path
 
 import aiofiles
@@ -54,7 +53,7 @@ class FilePersistentDict(PersistentDict):
             tmp = self.filepath.with_name(self.filepath.name + '.tmp')
             async with aiofiles.open(tmp, 'w', encoding=self.encoding) as f:
                 await f.write(dumps(self, str(self.filepath), indent=self.indent))
-            os.replace(tmp, self.filepath)
+            tmp.replace(self.filepath)
 
         if core.is_loop_running():
             background_tasks.create_lazy(async_backup(), name=self.filepath.stem)
--- a/nicegui/persistence/file_persistent_dict.py
+++ b/nicegui/persistence/file_persistent_dict.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 
 import aiofiles
@@ -50,8 +51,10 @@ class FilePersistentDict(PersistentDict):
             if not self:
                 await unlink_with_retry_async(self.filepath, missing_ok=True)
                 return
-            async with aiofiles.open(self.filepath, 'w', encoding=self.encoding) as f:
+            tmp = self.filepath.with_name(self.filepath.name + '.tmp')
+            async with aiofiles.open(tmp, 'w', encoding=self.encoding) as f:
                 await f.write(dumps(self, str(self.filepath), indent=self.indent))
+            os.replace(tmp, self.filepath)
 
         if core.is_loop_running():
             background_tasks.create_lazy(async_backup(), name=self.filepath.stem)
--- a/nicegui/persistence/file_persistent_dict.py
+++ b/nicegui/persistence/file_persistent_dict.py
@@ -1,3 +1,4 @@
+import contextlib
 from pathlib import Path
 
 import aiofiles
@@ -45,19 +46,24 @@ class FilePersistentDict(PersistentDict):
                 return
             self.filepath.parent.mkdir(exist_ok=True)
 
+        tmp_filepath = self.filepath.with_name(self.filepath.name + '.tmp')
+
         @background_tasks.await_on_shutdown
         async def async_backup() -> None:
             if not self:
+                tmp_filepath.unlink(missing_ok=True)
                 await unlink_with_retry_async(self.filepath, missing_ok=True)
                 return
-            tmp = self.filepath.with_name(self.filepath.name + '.tmp')
-            async with aiofiles.open(tmp, 'w', encoding=self.encoding) as f:
+            async with aiofiles.open(tmp_filepath, 'w', encoding=self.encoding) as f:
                 await f.write(dumps(self, str(self.filepath), indent=self.indent))
-            tmp.replace(self.filepath)
+            with contextlib.suppress(FileNotFoundError):  # a concurrent Storage.clear() may have swept the temp file
+                tmp_filepath.replace(self.filepath)
 
         if core.is_loop_running():
             background_tasks.create_lazy(async_backup(), name=self.filepath.stem)
         elif not self:
+            tmp_filepath.unlink(missing_ok=True)
             unlink_with_retry(self.filepath, missing_ok=True)
         else:
-            self.filepath.write_text(dumps(self, str(self.filepath), indent=self.indent), encoding=self.encoding)
+            tmp_filepath.write_text(dumps(self, str(self.filepath), indent=self.indent), encoding=self.encoding)
+            tmp_filepath.replace(self.filepath)
--- a/nicegui/storage.py
+++ b/nicegui/storage.py
@@ -1,3 +1,4 @@
+import contextlib
 import contextvars
 import os
 import uuid
@@ -211,6 +212,9 @@ class Storage:
         self._tabs.clear()
         for filepath in self.path.glob('storage-*.json'):
             helpers.unlink_with_retry(filepath, missing_ok=True)
+        for tmp_path in self.path.glob('storage-*.json.tmp'):
+            with contextlib.suppress(OSError):  # never wait: only an in-flight backup on this loop can hold it
+                tmp_path.unlink()
         if self.path.exists():
             self.path.rmdir()
 
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -453,6 +453,18 @@ async def test_unlinking_storage_files_waits_out_transient_holders(user: User):
     assert not filepath.exists()
 
 
+async def test_clearing_storage_removes_leftover_temp_files(user: User):
+    @ui.page('/')
+    def page():
+        ui.label('ok')
+
+    await user.open('/')  # needed to ensure NiceGUI's event loop is running
+    Storage.path.mkdir(exist_ok=True)
+    (Storage.path / 'storage-general.json.tmp').touch()  # stands in for a temp file left behind by an interrupted backup
+    app.storage.clear()
+    assert not Storage.path.exists(), 'temp files should be swept so the storage directory can be removed'
+
+
 @pytest.mark.parametrize('custom_cookie_headers', [False, True])
 def test_storage_cookie_headers(screen: Screen, custom_cookie_headers: bool):
     @ui.page('/')


falkoschindler (approved):
Thanks @evnchn — nice fix, and a very helpful PR description, residuals included.

I ran another AI review pass over the branch. It didn't find anything blocking — it confirmed the approach (the atomic temp-swap removes exactly the loop-bound holder that made the `unlink_with_retry` timeout structurally hopeless on Windows, and `create_lazy` serializes backups per file, so the fixed temp name should be race-free within one process) — and suggested a few small additions, which I pushed as two commits:

- 2a7bfd9c5 uses pathlib's `Path.replace` instead of `os.replace`, dropping the extra import.
- 84eb6df93 suppresses the `FileNotFoundError` when a concurrent `Storage.clear()` swept the temp file (that race is a deliberate discard, so staying silent — like `main` did — seems right), removes a leftover temp file when the dict becomes empty, makes the sync fallback backup atomic as well, and adds a platform-independent test that `Storage.clear()` sweeps leftover temp files (confirmed to fail without the sweep).

I haven't traced every race in detail myself, so please take another look whether this still matches your understanding of the problem. If it's fine with you, we can merge.

evnchn:

**Confirmed — your changes match my understanding of the problem, and I re-verified the Windows behaviour on your head.** One non-blocking note in the fold; nothing that should hold the merge. Upstream `quick-test` is still running as I write this, so worth a glance at green before you hit it.

<details><summary><b>What I checked</b></summary>

**The invariant holds.** The thing that must never come back is a *blocking wait on a temp file* — a temp's only possible holder is our own loop-bound writer, which is exactly the case `unlink_with_retry` can't help with. Nothing you added reintroduces one: the only temp cleanup in `Storage.clear()` is still a plain `unlink()` under `suppress(OSError)`, and your sync-path temp is still named `storage-<id>.json.tmp`, so the sweep's glob keeps matching it.

Point by point:

- **`Path.replace` for `os.replace`** — equivalent, and drops the import. Good.
- **Suppressing only `FileNotFoundError` around the replace** — agreed, and I think the reasoning is the important part: if `clear()` swept the temp while the writer was suspended, `clear()` should win and the backup must not resurrect storage. Other failures still raise, which is what you want.
- **Sync fallback made atomic, without the suppression** — the asymmetry is correct, not an oversight: the sync path has no interleaving point, so no same-process `clear()` can sweep between write and replace.
- **`tmp_filepath.unlink(missing_ok=True)` in the empty-dict async branch** — traced this one, since it's bare. It can only raise `PermissionError` on Windows if something *outside* the process holds the temp; `create_lazy(..., name=self.filepath.stem)` serialises same-file backups, so the empty backup can't overlap the non-empty backup that owns that temp. Not reachable from our own writer. Flagging it only because it lands in the same multi-process territory as the residual already listed in the description.

**Your new test has teeth — I checked rather than assumed.** Removed the sweep from `Storage.clear()`: `test_clearing_storage_removes_leftover_temp_files` fails with `OSError: [Errno 66] Directory not empty`. Restored it: passes.

**Windows re-verified on `84eb6df9`** (`windows-2022`, Python 3.10.20), since your commits touched the backup paths and there's no Windows leg upstream yet:

| | result |
|---|---|
| `test_unlinking_storage_files_waits_out_transient_holders` | **PASSED** |
| `test_clearing_storage_removes_leftover_temp_files` | **PASSED** |
| full `tests/test_storage.py` | 28 passed, 2 errors |

Both errors are `ConnectionResetError` / `WinError 10054` — the unrelated intermittent Windows *browser* flake from #6160. **No `WinError 32`, no `WinError 145`** anywhere in the log, which is the part that matters here.

**Local, at your head:** `tests/test_storage.py` → 29 passed, 0 skipped. `mypy ./nicegui` clean (245 files), `pylint ./nicegui` 10.00/10, `pre-commit` clean.

**Second-model pass** over your delta specifically (the same lineage that found a real blocker in my first attempt, when I'd widened the glob to `storage-*` and reintroduced the deadlock against the temp): no findings, recommends merge.

</details>

PR #6176, opened 2026-07-24
Fix move() orphaning an element when target_slot is invalid

evnchn:
### Motivation

`Element.move()` detaches the element from its current slot (`parent_slot.children.remove(self)` + `parent_slot.parent.update()`) **before** it validates `target_slot`. When `target_slot` does not exist, the `ValueError` is raised *after* the element has already been removed — leaving it orphaned: gone from its old slot's `children`, yet still pointing at that slot via `parent_slot` and still present in `client.elements`. The element is then unrecoverable through the public API — a subsequent `delete()` (or another `move()`) raises `ValueError: list.remove(x): x not in list`.

A plausible user typo (`label.move(other, target_slot='heder')`) is enough to trigger it.

### Implementation

Resolve and validate `target_slot` **first**; only then perform the remove / update / insert. This is a pure reorder — the happy path is unchanged (the removal still happens before `target_index` normalization, so same-slot moves and `target_index=-1` append keep their previous child-order semantics). No public-API or on-success behavior change.

<details>
<summary>Reproduction, before/after, and independent review</summary>

**MRE (ran on 3.13.0.post24.dev0, Py3.12.11) — before the fix:**

```python
from nicegui import ui
card = ui.card()
with card:
    label = ui.label('X')
other = ui.card()
try:
    label.move(other, target_slot='doesnotexist')
except ValueError:
    pass
label.delete()   # <-- ValueError: list.remove(x): x not in list
```
```
BUG REPRODUCED: failed move orphaned the element; delete() raised: ValueError('list.remove(x): x not in list'); orphaned=True
```

**Regression test — three-way A/B (must be able to fail):**
| state | `test_move_to_invalid_slot_keeps_element_in_place` |
|---|---|
| unfixed (origin/main) | **FAIL** — `assert label in card.default_slot.children` → `label in []` |
| with fix | **PASS** |

**Local gates:** `pre-commit` ✓ · `mypy ./nicegui` ✓ (245 files) · `pylint ./nicegui` 10.00/10 ✓

**Different-lineage review (Codex, adversarial):** *approve / SHIP* — "target_slot is now resolved before detach, but removal still happens before target_index normalization/insertion, so same-slot moves and target_index=-1 append retain the previous child-order semantics. No material findings."
</details>

### Progress

- [x] Documentation is not necessary (internal behavior fix, no API change).

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -508,22 +508,22 @@ class Element(Visibility):
         """
         parent_slot = self.parent_slot
         assert parent_slot is not None
-        parent_slot.children.remove(self)
-        parent_slot.parent.update()
         target_container = target_container or parent_slot.parent
 
         if target_slot is None:
-            parent_slot = target_container.default_slot
-            self.parent_slot = parent_slot
+            new_slot = target_container.default_slot
         elif target_slot in target_container.slots:
-            parent_slot = target_container.slots[target_slot]
-            self.parent_slot = parent_slot
+            new_slot = target_container.slots[target_slot]
         else:
             raise ValueError(f'Slot "{target_slot}" does not exist in the target container. '
                              f'Add it first using `add_slot("{target_slot}")`.')
 
-        target_index = target_index if target_index >= 0 else len(parent_slot.children)
-        parent_slot.children.insert(target_index, self)
+        parent_slot.children.remove(self)
+        parent_slot.parent.update()
+        self.parent_slot = new_slot
+
+        target_index = target_index if target_index >= 0 else len(new_slot.children)
+        new_slot.children.insert(target_index, self)
 
         target_container.update()
         return self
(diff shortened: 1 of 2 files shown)


falkoschindler (approved):
Thanks, @evnchn!

I checked the reorder for behavior changes on the happy path — the one that matters is that `target_index` normalization still happens _after_ the removal, so same-slot moves and `target_index=-1` keep their old child-order semantics. With the validation up front, the window between `remove()` and `insert()` is also exception-free, so "an element is in exactly one slot's `children`" now holds for every failure mode of `move()`.

I pushed a small commit simplifying the regression test: `should_see('X')` / `should_not_see('X')` catch the orphaning directly, which makes the `parent_slot` assertion redundant and the one after `delete()` vacuous.

So this looks good to me — approving!

PR #6180, opened 2026-07-25
Guard move() against moving an element into its own descendant

evnchn:
### Motivation

`move()` has no guard against moving a container into itself or one of its own descendants, which forms an `a ↔ b` containment cycle. Any later tree traversal (`descendants()`, `str()`, `clear()`, `delete()`) then recurses forever → `RecursionError` / hang.

Minimal reproduction (`nicegui/element.py:524`) — recursion capped first so it can't hang the machine:

```python
import sys
sys.setrecursionlimit(200)          # cap FIRST, before the cycle exists

from nicegui import ui
a = ui.card()
with a:
    b = ui.card()
a.move(b)                            # move a container into its own descendant -> a<->b cycle

list(a.descendants())               # RecursionError (would hang uncapped)
```
→ `RecursionError`

### Implementation

Reject a move into self or a descendant with a `ValueError` before any mutation (mirrors the existing target-slot validation), so the tree is left intact; legitimate moves are unaffected.

<details>

- Regression test fails on `origin/main` (no error raised, cycle formed), passes with the fix; a legitimate `x.move(p2)` still works.
</details>

### Progress

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -508,9 +508,9 @@ class Element(Visibility):
         """
         parent_slot = self.parent_slot
         assert parent_slot is not None
-        if target_container is not None and self in target_container.ancestors(include_self=True):
-            raise ValueError('Cannot move an element into itself or one of its descendants.')
         target_container = target_container or parent_slot.parent
+        if self in target_container.ancestors(include_self=True):
+            raise ValueError('Cannot move an element into itself or one of its descendants.')
 
         if target_slot is None:
             new_slot = target_container.default_slot
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -508,6 +508,8 @@ class Element(Visibility):
         """
         parent_slot = self.parent_slot
         assert parent_slot is not None
+        if target_container is not None and self in target_container.ancestors(include_self=True):
+            raise ValueError('Cannot move an element into itself or one of its descendants.')
         parent_slot.children.remove(self)
         parent_slot.parent.update()
         target_container = target_container or parent_slot.parent
--- a/nicegui/elements/date_input.py
+++ b/nicegui/elements/date_input.py
@@ -39,7 +39,7 @@ class DateInput(LabelElement, ValueElement[str | None], DisableableElement):
         with self.add_slot('append'):
             with button(icon='edit_calendar', color=None).props('flat round') as self.button:
                 with menu() as self.menu:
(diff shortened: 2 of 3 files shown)


falkoschindler:

The conflict is in `nicegui/element.py` and `tests/test_element.py`, against ff073f463 ("Fix move() orphaning an element when target_slot is invalid", #6176). Both changes rework the same part of `move()`, so resolving it is a judgement call about the combined logic rather than a mechanical merge — handing it back to you.

---

evnchn:

**Resolved** — merged `main` in (`fa8c983b`), now `MERGEABLE` with CI green.

<details><summary><b>How the two move() changes combined</b></summary>

You were right that this one was a judgement call rather than a mechanical merge. #6176 stops `move()` orphaning an element when `target_slot` is invalid; this PR rejects a move into the element's own descendant. They compose — both are pre-flight rejections, and neither may leave the element detached.

The ordering is the load-bearing part. The cycle guard has to run **before** `target_container` defaults to the current parent slot, and before any detach:

```python
parent_slot = self.parent_slot
assert parent_slot is not None
if target_container is not None and self in target_container.ancestors(include_self=True):
    raise ValueError('Cannot move an element into itself or one of its descendants.')
target_container = target_container or parent_slot.parent
```

The naive merge — guard placed after the detach — still passes the original test, because that test only checked that `ValueError` was raised and that the subtree was intact. So `668df2ca` adds the missing assertion, mirroring the sibling check #6176 introduced:

```python
root = a.parent_slot
with pytest.raises(ValueError):
    a.move(b)
with pytest.raises(ValueError):
    a.move(a)
assert a in root.children, 'a rejected move must keep the element in its original slot'
```

Verified: `quick-test / 3.10 / pytest`, `check / mypy`, `check / pre-commit`, `check / pylint` all pass on `668df2ca`.

</details>

falkoschindler (approved):
Thanks for catching this, Evan! The guard is correct and well-placed: it runs before any mutation, so a rejected move leaves the tree untouched, and checking the target's ancestors is O(depth) rather than scanning the whole subtree.

I pushed a small polish commit: the guard now sits below the defaulting line (the default target — the current parent — can never be `self` or a descendant, so the `is not None` check became unnecessary), the test pins the expected error with `match=` to distinguish it from the unknown-slot `ValueError`, and the test variables are named `outer`/`inner`/`label`/`other` to spell out the nesting.

PR #6181, opened 2026-07-25
Sync ui.table fullscreen state back to the server on client-driven toggle

evnchn:
### Motivation

`ui.table` mirrors Quasar's `fullscreen` prop on the Python side (`is_fullscreen`), but the client can toggle fullscreen on its own — e.g. a slot button calling Quasar's `props.toggleFullscreen`:

```python
table = ui.table(rows=[{'name': 'Alice'}])
table.add_slot('top', '<q-btn label="FS" @click="props.toggleFullscreen" />')
ui.button('Fullscreen', on_click=table.toggle_fullscreen)
```

`table.js` already reacts to such client-driven toggles (it updates the `nicegui-table-fullscreen` CSS class via `@fullscreen`), but `_props['fullscreen']` was never synced back: after clicking "FS", `is_fullscreen` still returns `False`, and the next server-side `toggle_fullscreen()` sends a value the client already has — nothing visible happens, one toggle is "eaten".

Note that on the web this can only be triggered by such a client-side toggle: Quasar's table fullscreen is CSS-based (not the browser Fullscreen API), so pressing ESC does not leave it, and the History-plugin exit only exists in Cordova/Capacitor.

### Implementation

Register an `update:fullscreen` handler (mirroring the existing `update:pagination` handler) that feeds the client value into the existing `set_fullscreen` setter, so client-driven toggles sync back and `is_fullscreen` stays truthful.

A new Screen test toggles fullscreen on and off via a top-slot button and waits for `is_fullscreen` to follow in both directions.

### Progress


falkoschindler (approved):
Thanks Evan, nice catch! The fix is correct and mirrors the existing `update:pagination` pattern — fullscreen was the only mirrored prop without a back-sync.

I pushed a small polish commit directly to your branch:

- the handler now reuses the existing `set_fullscreen` setter instead of writing `_props` directly,
- the test also covers the exit direction (the originally reported desync) and uses `screen.wait_for` instead of fixed 0.5s sleeps.

I also updated the PR description: on the web, pressing ESC doesn't leave Quasar's CSS-based table fullscreen (the History-plugin exit only exists in Cordova/Capacitor), so the real trigger is a client-side toggle like a top-slot button.

PR #6182, opened 2026-07-25
Make ui.query.style(replace=...) replace instead of merge

evnchn:
### Motivation

`Query.style(replace=...)` merges instead of replacing — inconsistent with the direct element's `.style()`.

The direct element's `Style.__call__` implements `replace` as **start from an empty dict** (`style_dict = {**self} if replace is None else {}`) — a true replacement. `Query.style` only pops the `remove` keys and then `.update()`s `replace` into the *existing* style, so `replace` behaves identically to `add`. The two share the same docstring ("styles to use instead of existing ones"); the direct element honours it, `ui.query` does not — so this is a behaviour bug, not a wrong docstring.

Minimal reproduction (`nicegui/elements/query.py:76`):

```python
from nicegui import ui
q = ui.query('body')
q.style('color: blue; font-size: 20px')
q.style(replace='color: red')            # should leave ONLY color:red
print(q.element.props['style'])
# origin/main: {'color':'red','font-size':'20px'}  <- font-size survives (merged, not replaced)
# element.style(replace=...) would have dropped it
```

### Implementation

Rebuild the style dict from scratch for `replace` (mirroring `Style.__call__` and `Query.classes`) and diff old-vs-new keys so removed styles are cleared on the client.

<details>

- Regression test asserts a `replace` drops the prior keys: FAILS on origin/main (merged), PASSES with the fix.
- Local gates: pre-commit · mypy `./nicegui` · pylint 10.00/10 · pytest — all green.
</details>

### Progress


falkoschindler (approved):
Thanks @evnchn, the `replace` fix is spot on and the test nails it. I pushed two small follow-up commits on top, feel free to object:

1. 477d05167 extracts `Style.update_dict` (mirroring `Classes.update_list`) so `Style.__call__`, `Element.default_style` and `Query.style` share one implementation instead of three copies.
2. 9f5d2e409 keeps forwarding explicit `remove=` keys to the browser in `Query.style` — the new tracked-only diff would have silently stopped removing inline styles the query never set (e.g. from `ui.add_body_html`). `Query.classes` now does the same, so all three siblings (`classes`, `style`, `props`) behave alike. Also: no element update when nothing changed, `wait_for` instead of a fixed sleep, and a test for removing foreign styles and classes.

One behavior note for the record: `replace=''` now clears all styles (it used to be a no-op), which matches `element.style(replace='')`.

Approving. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20findings%20and%20wording%20from%20the%20model%2C%20decisions%20on%20scope%20and%20semantics%20from%20the%20conversation%20with%20the%20reviewer.&a=claude-code "claude-fable-5: Review findings and wording from the model, decisions on scope and semantics from the conversation with the reviewer.")

evnchn:
No objections

PR #6187, opened 2026-07-25
Clean up spilled temp file when an upload fails mid-spill

evnchn:
### Motivation

A user uploads a file larger than spool_max_size (default 1 MB). Midway through spilling to the temp file, `await temp_file.write(chunk)` raises OSError (disk full) or the read raises. The except/finally closes the handle but leaves the partially written temp file on disk with no reference and no finalizer -> permanent leak; a subsequent retry loop multiplies it.

Minimal reproduction (`nicegui/elements/upload_files.py:143`) — the exact test/script that was run to reproduce it:

```python
"""MRE: create_file_upload leaks a spilled temp file if write fails before
LargeFileUpload (with its weakref finalizer) is ever constructed.
"""
import asyncio
import os
from io import BytesIO

import aiofiles.tempfile
from starlette.datastructures import Headers, UploadFile
from starlette.formparsers import MultiPartParser

from nicegui.elements import upload_files

created_paths: list[str] = []
_real = aiofiles.tempfile.NamedTemporaryFile


class _WriteFailsTempFile:
    """Real temp file on disk, but every write raises OSError (simulated disk-full)."""

    def __init__(self, inner):
        self._inner = inner
        self.name = inner.name
        created_paths.append(inner.name)

    async def write(self, data):
        raise OSError('No space left on device')

    async def close(self):
        await self._inner.close()


async def _patched(*args, **kwargs):
    inner = await _real(*args, **kwargs)  # a real temp file is created on disk here
    return _WriteFailsTempFile(inner)


async def main() -> None:
    aiofiles.tempfile.NamedTemporaryFile = _patched
    MultiPartParser.spool_max_size = 10   # TOGGLE: set to 10_000 -> no spill -> PASS (no leak)

    content = b'x' * 100
    upload = UploadFile(
        file=BytesIO(content),
        size=len(content),
        filename='big.bin',
        headers=Headers({'content-type': 'application/octet-stream'}),
    )
    try:
        await upload_files.create_file_upload(upload, chunk_size=8)
    except OSError as e:
        print(f'write failed as designed: {e}')

    leaked = [p for p in created_paths if os.path.exists(p)]
    if leaked:
        print(f'BUG REPRODUCED: spilled temp file orphaned on disk: {leaked}')
    else:
        print('OK: no temp file left on disk')


asyncio.run(main())
```
→ `write failed as designed: No space left on device`

### Implementation

Added success flag + _cleanup_path on error path in create_file_upload (unlink after close, Windows-safe); success path byte-identical.

<details>

- FAIL on origin/main (leaked on disk), PASS with fix.
</details>

### Progress


falkoschindler (approved):
Thanks for the PR, @evnchn! The fix is correct and minimal: the `success` flag with cleanup in `finally` covers both exceptions and `CancelledError` (client disconnect — the realistic trigger), unlinking after `close()` keeps it Windows-safe, and the success path is untouched. Verified locally: the new test fails on `main` and passes with the fix.

I pushed a small simplification of the test: instead of wrapping `aiofiles.tempfile.NamedTemporaryFile` to track created paths, it now redirects `tempfile.tempdir` into pytest's `tmp_path` and asserts the directory is empty afterwards — same coverage, but no coupling to the concrete factory the implementation uses. The assert in the second `read()` call keeps the guard against the test passing without an actual spill.

Let's merge! 🚀

PR #6189, opened 2026-07-25
Disconnect header ResizeObserver on unmount

evnchn:
### Motivation

Create a ui.header (default add_scroll_padding=True), then remove it (page teardown, sub_pages navigation, or a refreshable that rebuilds the layout). unmounted() runs but disconnect() is a no-op; the ResizeObserver and the whole header component subtree are retained and its callback continues mutating scrollPaddingTop. Repeated header create/destroy accumulates orphaned observers → memory leak and stray scroll-padding writes.

Minimal reproduction (`nicegui/elements/header.js:5`) — the exact test/script that was run to reproduce it:

```js
from nicegui import ui


def test_repro(screen: Screen) -> None:
    @ui.page('/')
    def page() -> None:
        ui.add_head_html('''
            <script>
            window.__ro = [];
            const Orig = window.ResizeObserver;
            window.ResizeObserver = class extends Orig {
                constructor(cb) { super(cb); this.__rec = {targets: [], disconnected: false}; window.__ro.push(this.__rec); }
                observe(el, o) { this.__rec.targets.push((el && el.className) || ''); return super.observe(el, o); }
                disconnect() { this.__rec.disconnected = true; return super.disconnect(); }
            };
            </script>
        ''')

        header = ui.header()          # add_scroll_padding=True by default -> registers a ResizeObserver on the header element
        with header:
            ui.label('hi')
        ui.button('kill', on_click=header.delete)  # public API: destroy the header

    screen.open('/')
    screen.click('kill')
    screen.wait(0.5)              # let the delete round-trip and unmounted() run

    ro = screen.selenium.execute_script('return window.__ro')
    header_obs = [r for r in ro if any('nicegui-header' in t for t in r['targets'])]
    print(f'BUG CHECK: header-observing ResizeObservers = {header_obs}')
    assert header_obs, 'sanity: no ResizeObserver observed the header element'
    leaked = [r for r in header_obs if not r['disconnected']]
    assert not leaked, \
        f'LEAK: {len(leaked)} header ResizeObserver(s) never disconnected on teardown: {header_obs}'
```
→ `BUG CHECK: header-observing ResizeObservers = [{'disconnected': True, 'targets': ['q-header q-layout__section--marginal fixed-top nicegui-header wrap']}, {'disconnected': False, 'targets': ['q-header q-layout__section--marginal fixed-top nicegui-header wrap']}]`

### Implementation

Split .observe() from the assignment in header.js so this.resizeObserver holds the instance and unmounted()'s disconnect() actually fires.

<details>

- FAIL on origin/main (1 header observer stays disconnected:false) and PASS with fix.
</details>

### Progress

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ If you would like to support this project and have your avatar or company logo a
   <a href="https://github.com/digiquip"><img src="https://github.com/digiquip.png" width="50px" alt="DigiQuip AS" /></a>
   <a href="https://github.com/aitech95"><img src="https://github.com/aitech95.png" width="50px" alt="JACOB SHI " /></a>
   <a href="https://github.com/Zaga-Solutions"><img src="https://github.com/Zaga-Solutions.png" width="50px" alt="Zaga Solutions" /></a>
-  <a href="https://github.com/Discolike"><img src="https://github.com/Discolike.png" width="50px" alt="DiscoLike Inc." /></a>
+  <a href="https://github.com/discolike"><img src="https://github.com/discolike.png" width="50px" alt="DiscoLike Inc." /></a>
 </p>
 <!-- SPONSORS -->
 
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -508,22 +508,22 @@ class Element(Visibility):
         """
         parent_slot = self.parent_slot
         assert parent_slot is not None
-        parent_slot.children.remove(self)
-        parent_slot.parent.update()
         target_container = target_container or parent_slot.parent
 
         if target_slot is None:
-            parent_slot = target_container.default_slot
-            self.parent_slot = parent_slot
+            new_slot = target_container.default_slot
         elif target_slot in target_container.slots:
-            parent_slot = target_container.slots[target_slot]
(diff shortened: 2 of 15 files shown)


falkoschindler (approved):
Thanks, Evan! Nice catch — `new ResizeObserver(...).observe(el)` returns `undefined`, so `this.resizeObserver` was never set and `unmounted()`'s `disconnect()` silently did nothing. The two-line fix is exactly right.

I took the liberty of dropping `test_scroll_padding_observer_disconnected` again, though. It's a white-box test: it monkeypatches `window.ResizeObserver` globally and asserts on observer bookkeeping rather than on anything a user can observe, so it would break the moment we switch to a different resize mechanism, even if the behavior stays correct. It also relies on the `nicegui-header` class name to tell our observer apart from Quasar's — a silent-pass hazard if that class ever changes — and it costs a full browser round-trip plus a fixed `screen.wait(0.5)` to guard against a mistake that only exists in this one place (`header.js` is our only first-party `ResizeObserver` user).

The fix itself stays, of course. Merging as is.

PR #6191, opened 2026-07-25
Support URL-encoded data-URL favicons

evnchn:
### Motivation

Developer passes favicon='data:image/svg+xml,%3Csvg%20.../%3E' (a valid, URL-encoded data URL, the same form NiceGUI generates internally) to ui.run(). is_file() returns False (data: guard), so startup wires get_favicon_response; a request to /favicon.ico calls _data_url_to_bytes -> b64decode drops the '%','<' etc. and returns corrupt bytes served as image/svg+xml. A value like 'data:foo' (no comma) instead raises ValueError -> HTTP 500. Fix: detect the ';base64' marker and branch to urllib.parse.unquote for non-base64 data URLs; guard the missing-comma case.

Minimal reproduction (`nicegui/favicon.py:107`):

```python
from nicegui import core, favicon

svg = '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><circle cx="8" cy="8" r="7"/></svg>'
data_url = favicon._svg_to_data_url(svg)
assert data_url.startswith('data:') and ';base64' not in data_url

core.app.config.favicon = data_url

try:
    favicon.get_favicon_response()
except Exception as e:
    print(f'BUG REPRODUCED: serving a NiceGUI-generated data-URL favicon raised '
          f'{type(e).__name__}: {e}')
else:
    raise SystemExit('BUG FIXED: data-URL favicon served without error')
```
→ `BUG REPRODUCED: serving a NiceGUI-generated data-URL favicon raised Error: Incorrect padding`

### Implementation

Fixed _data_url_to_bytes to branch on ;base64 marker (base64 path byte-identical; percent-encoded uses urllib.parse.unquote) and partition() to drop the comma-less ValueError.

<details>

- fail-first test tests/test_favicon.
</details>

### Progress


falkoschindler (approved):
Nice find — serving the very data URLs NiceGUI generates itself was clearly broken. The fix is minimal and the test asserts the actually served favicon. 👍

I pushed a small follow-up commit: `unquote_to_bytes` decodes percent-encoded bytes exactly (no UTF-8 replacement for non-UTF-8 payloads), comma-less data URLs now raise a clear `ValueError` instead of silently serving an empty response, and the new test also asserts the `<link>` URL like its siblings.

PR #6192, opened 2026-07-25
Remove bindings when a scene Object3D is deleted

evnchn:
### Motivation

In a long-lived scene, repeatedly: create a box, bind a slider value to box position, then obj.delete(). Each deleted box remains referenced by the binding registry because delete() never calls binding.remove, leaking memory that only clears when the whole scene element is deleted.

Minimal reproduction (`nicegui/elements/scene/scene_object3d.py:345`) — the exact test/script that was run to reproduce it:

```python
from nicegui import ui, binding

N = 50
scene = ui.scene()
label = ui.label()

for _ in range(N):
    box = scene.box()
    label.bind_text_from(box, 'x')  # registers a binding link referencing box
    box.delete()                    # Object3D.delete does NOT call binding.remove

# All N boxes are gone from the scene ...
assert len(scene.objects) == 0, f'scene still holds {len(scene.objects)} objects'

# ... but every binding link created for the deleted boxes is still registered.
leaked = len(binding.active_links)
print(f'scene.objects={len(scene.objects)}  binding.active_links={leaked}  binding.bindings={len(binding.bindings)}')

# Contrast: Scene._handle_delete calls binding.remove(...); Object3D.delete does not.
assert leaked == 0, f'BUG REPRODUCED: {leaked} binding links leaked for deleted Object3D (expected 0)'
print('OK: no leak')
```
→ `scene.objects=0  binding.active_links=50  binding.bindings=50`

### Implementation

Object3D.delete() now calls binding.remove([self]) before _delete() (mirrors Scene._handle_delete), plus `from ...

<details>
<summary>Leak magnitude (tracemalloc) — yes, it can reach a gigabyte</summary>

Loop `box = scene.box(); label.bind_text_from(box, 'x'); box.delete()` and measure with `tracemalloc`:

| | leaked `binding.active_links` | tracemalloc growth | per cycle | cycles to 1 GB |
|---|---|---|---|---|
| **origin/main** | 5000 / 5000 | 8.18 MB | 1,637 B | **~611,000** |
| **with this fix** | 0 / 5000 | 0.02 MB (noise) | ~4 B | — |

The leak is linear and unbounded — each bind-then-delete retains the whole `Object3D` via the dangling binding. A long-running scene that creates/binds/deletes objects at 100/s reaches 1 GB in ~1.7 h. The fix zeroes it.
</details>

<details>

- Fail-first A/B: the test asserts the deleted object is actually **released** — a `weakref.WeakSet` holding the box is empty after `delete()`. Reverting the fix gives `AssertionError: assert 1 == 0`; with the fix it passes. It uses the `User` fixture and runs in 0.17 s.
- The earlier revision asserted on `binding.active_links` / `binding.bindings` over a 50-object loop. Per #6219 the observable consequence (the object is freed) is the better thing to assert, and one object proves it.
</details>

<details>
<summary>Cost: <code>scene.clear()</code> on large bound scenes</summary>

`binding.remove()` is called once per object, so clearing a scene where every object is bound gets slower:

| bound objects | `scene.clear()` on main | with this PR |
|---|---|---|
| 2 000 | 0.137 s | 0.293 s |
| 8 000 | 2.343 s | 6.314 s |

This is a ~2.7x constant factor on a path that is **already quadratic** — `children` scans all objects per
object, and `binding.remove` rebuilds `active_links[:]` on every call — not a new complexity class. It only
shows up when objects are bound; unbound scenes are unaffected.

For the whole-scene case `Scene._handle_delete` already batches (`binding.remove(list(self.objects.values()))`).
Doing the same inside `delete_objects()` would remove the constant factor, but it is a larger change than this
fix needs, so it is called out here rather than folded in — happy to do it if you would prefer.

</details>

### Progress

### DIFF
--- a/nicegui/elements/date_input.py
+++ b/nicegui/elements/date_input.py
@@ -39,7 +39,7 @@ class DateInput(LabelElement, ValueElement[str | None], DisableableElement):
         with self.add_slot('append'):
             with button(icon='edit_calendar', color=None).props('flat round') as self.button:
                 with menu() as self.menu:
-                    self.picker = date().props('no-parent-event').props('range' if range_input else '')
+                    self.picker = date().props('range' if range_input else '')
 
         self.picker.bind_value(self,
                                forward=lambda v: self._picker_to_input_value(v) if self._range_input else v,
--- a/nicegui/elements/scene/scene_object3d.py
+++ b/nicegui/elements/scene/scene_object3d.py
@@ -6,6 +6,8 @@ from typing import TYPE_CHECKING, Any, Literal
 
 from typing_extensions import Self
 
+from ... import binding
+
 if TYPE_CHECKING:
     from .scene import Scene, SceneObject
 
@@ -347,4 +349,5 @@ class Object3D:
         for child in self.children:
             child.delete()
         del self.scene.objects[self.id]
+        binding.remove([self])
         self._delete()
--- a/tests/test_scene.py
+++ b/tests/test_scene.py
@@ -5,9 +5,9 @@ import numpy as np
 import pytest
 from selenium.common.exceptions import JavascriptException
 
-from nicegui import app, ui
+from nicegui import app, binding, ui
 from nicegui.elements.scene import Object3D
-from nicegui.testing import Screen
+from nicegui.testing import Screen, User
 
 from .test_helpers import TEST_DIR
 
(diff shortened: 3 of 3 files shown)


falkoschindler:

The conflict is in `tests/test_scene.py`, against 85832656a ("Guard scene dragend against deleted objects", #6195) — both PRs add tests in the same place. Should be a straightforward resolution, but it's your file on both sides, so handing it back to you.

---

evnchn:

**Resolved** — merged `main` in (`8e1bb4ca`), now `MERGEABLE` with CI green.

<details><summary><b>What the resolution came to</b></summary>

As you expected, mechanical: #6195's `dragend`-guard test and this PR's binding-cleanup test both land in `tests/test_scene.py` but assert different things, so both stay — no logic had to be reconciled.

Diff against `main` is unchanged in shape: `nicegui/elements/scene/scene_object3d.py` +3, `tests/test_scene.py` +16.

Verified on `8e1bb4ca`: `quick-test / 3.10 / pytest`, `check / mypy`, `check / pre-commit`, `check / pylint` all pass.

</details>

falkoschindler (approved):
Nice find, and thanks for the thorough analysis of leak magnitude and cost!

I pushed e913657 adding a `gc.collect()` before the weakref assertion, so the test asserts the object is unreachable rather than relying on CPython's immediate refcounting (see the cross-platform note in REVIEW.md). Verified fail-first both ways: without `binding.remove([self])` the test still fails.

PR #6193, opened 2026-07-25
Skip change notification when setdefault hits an existing key

evnchn:
### Motivation

Reproduced with a registered on_change handler: `d=ObservableDict({'a':1}, on_change=...); d.setdefault('a', 999)` fires the handler once though `d['a']` stays 1; `d.pop('zzz', None)` on a missing key also fires the handler once despite no mutation.

Minimal reproduction (`nicegui/observables.py:106`) — the exact test/script that was run to reproduce it:

```python

calls: list = []
d = ObservableDict({'a': 1}, on_change=lambda e: calls.append(e))

# Case 1: setdefault on an EXISTING key -> value must stay 1, nothing changed.
calls.clear()
result = d.setdefault('a', 999)
assert result == 1 and d['a'] == 1, f'unexpected mutation: {d!r}'
setdefault_spurious = len(calls)  # should be 0; bug -> 1

# Case 2: pop of a MISSING key with default -> nothing removed.
calls.clear()
result = d.pop('zzz', None)
assert result is None and d == {'a': 1}, f'unexpected mutation: {d!r}'
pop_spurious = len(calls)  # should be 0; bug -> 1

print(f'setdefault(existing) change events: {setdefault_spurious} (expected 0)')
print(f'pop(missing)        change events: {pop_spurious} (expected 0)')

if setdefault_spurious or pop_spurious:
    print('BUG REPRODUCED: change notification fired though the dict was unchanged')
    raise SystemExit(1)
print('OK: no spurious notifications')
```
→ `setdefault(existing) change events: 1 (expected 0)`

### Implementation

Fixed ObservableDict.setdefault to skip _handle_change when key already exists (matches pop() KeyError-skip precedent); minimal 6-line diff.

<details>

- FAIL (assert 8==7) on origin/main source, PASS with fix.
</details>

### Progress

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ If you would like to support this project and have your avatar or company logo a
   <a href="https://github.com/digiquip"><img src="https://github.com/digiquip.png" width="50px" alt="DigiQuip AS" /></a>
   <a href="https://github.com/aitech95"><img src="https://github.com/aitech95.png" width="50px" alt="JACOB SHI " /></a>
   <a href="https://github.com/Zaga-Solutions"><img src="https://github.com/Zaga-Solutions.png" width="50px" alt="Zaga Solutions" /></a>
-  <a href="https://github.com/Discolike"><img src="https://github.com/Discolike.png" width="50px" alt="DiscoLike Inc." /></a>
+  <a href="https://github.com/discolike"><img src="https://github.com/discolike.png" width="50px" alt="DiscoLike Inc." /></a>
 </p>
 <!-- SPONSORS -->
 
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -508,22 +508,22 @@ class Element(Visibility):
         """
         parent_slot = self.parent_slot
         assert parent_slot is not None
-        parent_slot.children.remove(self)
-        parent_slot.parent.update()
         target_container = target_container or parent_slot.parent
 
         if target_slot is None:
-            parent_slot = target_container.default_slot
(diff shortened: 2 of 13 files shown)


falkoschindler:
Thanks for digging into this, Evan — but both cases from the motivation are already fixed on `main`, so I'm closing this one as superseded.

- `setdefault` on an existing key: fixed in #6171 (merged 2026-07-23), which added the same early return as part of a broader set of no-op guards covering `dict.update`, `list.extend`, `__imul__`, `set.add`/`discard`/`update`, and the `intersection`/`difference`/`symmetric_difference` variants.
- `pop` of a missing key with a default: fixed earlier in #6164 (merged 2026-07-22).

Running the reproduction script from the description against current `main`:

```text
setdefault(existing) change events: 0 (expected 0)
pop(missing)        change events: 0 (expected 0)
```

One detail worth carrying over rather than losing: #6171 deliberately returns `super().__getitem__(__key)` instead of `self[__key]`. `PersistentDict` overrides `__getitem__`, and going through `self` would let a subclass intercept a lookup that `dict.setdefault` never routes through it. Same fix, one subtlety more.

The test addition here is covered too — `test_no_op_dict_mutations_do_not_fire_change_events` in `tests/test_observables.py` asserts the same behaviour plus the return value.

The second commit on the branch, removing the dead `no-parent-event` prop from the date picker, landed independently as #6177.

evnchn:
Yeah I think this one is caused by my main in fork not being latest 🙇

PR #6194, opened 2026-07-25
Fix Redis storage so key deletions propagate across instances

evnchn:
### Motivation

Two instances A,B share Redis; user storage = {'cart':[1,2],'coupon':'X'}. On A: `del app.storage.user['coupon']`. A publishes {'cart':[1,2]}. B's listener runs self.update({'cart':[1,2]}) -> B still has coupon='X'. B later touches the cart and re-publishes coupon back to A -> the deleted coupon reappears on A. State never converges.

Minimal reproduction (`nicegui/persistence/redis_persistent_dict.py:67`) — the exact test/script that was run to reproduce it:

```python
"""MRE for NiceGUI bug idx=8: RedisPersistentDict pub/sub listener merges remote
snapshots with dict.update(), so a key deleted on one instance never disappears on
the others (and can be resurrected).

Buggy line: nicegui/persistence/redis_persistent_dict.py:67
    if new_data != self:
        self.update(new_data)   # <-- MERGE, not REPLACE => deletions never propagate

RUN: /Users/evnchn/nicegui/.venv/bin/python mre.py
"""
import asyncio
import json as stdjson

from nicegui import core
from nicegui.persistence.redis_persistent_dict import RedisPersistentDict


class FakePubSub:
    """Stands in for redis.asyncio pub/sub: replays a fixed message stream, no server."""

    def __init__(self, messages):
        self._messages = messages
        self.subscribed = True

    async def subscribe(self, *_channels):
        self.subscribed = True

    async def unsubscribe(self, *_channels):
        self.subscribed = False

    async def listen(self):
        for message in self._messages:
            yield message

    async def aclose(self):
        self.subscribed = False


async def main() -> None:
    core.loop = asyncio.get_running_loop()

    d = RedisPersistentDict(url='redis://localhost:6379', id='mre-idx8')
    d._change_handlers.clear()

    d['cart'] = [1, 2]
    d['coupon'] = 'X'

    peer_snapshot = {'cart': [1, 2]}  # note: NO 'coupon' key
    d.pubsub = FakePubSub([
        {'type': 'subscribe', 'data': 1},
        {'type': 'message', 'data': stdjson.dumps(peer_snapshot)},
        {'type': 'unsubscribe', 'data': 0},
    ])

    d._start_listening()
    await asyncio.wait_for(d._listener_task, timeout=5)

    print(f'peer published (coupon deleted): {peer_snapshot}')
    print(f'local state after listener merged it: {dict(d)}')

    if 'coupon' in d:
        raise AssertionError(
            "BUG REPRODUCED: peer deleted 'coupon' and published {'cart': [1, 2]}, "
            "but the listener's self.update(new_data) MERGED it, so 'coupon' still "
            f"exists locally: {dict(d)}.")
    print('OK (fixed): deletion propagated, local state converged.')


if __name__ == '__main__':
    asyncio.run(main())
```
→ `peer published (coupon deleted): {'cart': [1, 2]}`

### Implementation

Fixed RedisPersistentDict listener to prune keys missing from incoming snapshot (del-loop over set(self)-new_data.keys() then update) giving replace-not-merge semantics; deletions now propagate across instances.

<details>

- Fail-first verified: FAILS on origin/main with exact symptom (coupon survives), PASSES with fix.
</details>

### Progress

- [x] Pytests are not necessary (reasoning in the comment below).


evnchn:

**Test removed, two-line fix unchanged** — it mirrored the machinery under test, and there is no small honest test to replace it with, so this PR now ships with none. Verified against a real Redis instead.

<details>
<summary><b>Why no test, and what was verified</b></summary>

`test_redis_listener_propagates_key_deletions` was the pattern `#6219` describes: a `FakePubSub` mirroring the real pubsub machinery, a monkeypatched `publish`, and the private `_start_listening()` / `_listener_task` driven directly. It would need rewriting on any refactor of the listener, so it guarded the implementation rather than the regression.

**No test is committed, because there is no small honest one to write here.** A behavior-level test needs two `RedisPersistentDict` instances talking over a real Redis, and the suite has none to talk to: `_test.yml` has no `services:` block, `uv.lock` has no `fakeredis` (only the `redis` client, which cannot pubsub without a server), and `nicegui/testing/` has no Redis fixture. A skip-if-unreachable test would be permanently skipped in CI and guard nothing. That left rebuilding the fake machinery, or adding Redis CI infrastructure — a much larger change than a two-line fix warrants.

Verified manually instead: ran `redis:alpine` locally and drove two `RedisPersistentDict` instances sharing one id through the public surface — `a.update({'cart': [1, 2], 'coupon': 'X'})`, then `del a['coupon']` — asserting the peer observes the deletion.

With the fix reverted:

```
AssertionError: deletion did not propagate, b={'cart': [1, 2], 'coupon': 'X'}
```

With the fix applied, it passes. Also `tests/test_storage.py` 29 passed / 0 skipped, and pre-commit, `mypy ./nicegui`, `pylint ./nicegui` 10.00/10 green. The PR's diff against its merge-base is now exactly the two-line source change.

Happy to add a `services: redis:` job and a real two-instance test if you'd rather have CI coverage here — say the word and I'll open it separately.
</details>

falkoschindler (approved):
The two-line fix is minimal and correct: pruning keys that are missing from the incoming snapshot gives the listener replace-instead-of-merge semantics, so deletions finally propagate. Nicely placed, too — no `await` between the deletes and the `update()`, so the lazily scheduled publish only ever sees the final state. And since last-writer-wins was already the semantics at the Redis `SET` level, this doesn't introduce a new race; it just makes the in-memory state agree with what Redis actually holds.

Issue #228, opened 2023-01-05
'cellRenderer': 'checkboxRenderer' for ui.table stopped working in NiceGui version 1.0.


JS-Aibel:
To be able to work with checkboxes in the ui.table we used the `'cellRenderer': 'checkboxRenderer'` in a column definition with NiceGui version 0.0.9. 

```  
'columnDefs': [
    {'headerName': "",'field': "enabled", 'cellRenderer': 'checkboxRenderer', 'maxWidth': 40},
```

After updating to NiceGui version 1.0. we are not able to use the 'cellRenderer': 'checkboxRenderer'. Has this anything to do with NiceGUI 1.0 does not use JustPy anymore?

falkoschindler:
This feature will be available in the next release of version 1.1.0.

PR #6196, opened 2026-07-25
Disconnect echart ResizeObserver on unmount

evnchn:
### Motivation

`echart.js` leaks a `ResizeObserver` on every unmount — the same class of bug fixed for `ui.header` in #228, surfaced by the systematic sweep that PR prompted.

`this.resizeObserver = new ResizeObserver(...).observe(this.$el)` chains `.observe()`, which returns `undefined`, so `this.resizeObserver` ends up `undefined`. `unmounted() { this.resizeObserver?.disconnect() }` is then a silent no-op (the `?.` masks it) → one leaked `ResizeObserver` per chart create/destroy.

Minimal reproduction — a `Screen` test that wraps `ResizeObserver` to count disconnects across an echart delete:

```python
# after mounting ui.echart(...) then clicking a button that deletes it:
before = window.__ro.filter(r => r.disconnected).length
# ... delete the chart ...
after  = window.__ro.filter(r => r.disconnected).length
assert after > before   # origin/main: after == before (observer never disconnected)
```

### Implementation

Split `.observe()` onto its own statement so `this.resizeObserver` holds the instance and `disconnect()` fires on unmount (mirrors the #228 header fix).

<details>

- Three-way A/B: FAILS on origin/main (observer not disconnected), PASSES with the fix.
- Local gates: pre-commit · mypy `./nicegui` · pytest — all green.
</details>

### Progress

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ If you would like to support this project and have your avatar or company logo a
   <a href="https://github.com/digiquip"><img src="https://github.com/digiquip.png" width="50px" alt="DigiQuip AS" /></a>
   <a href="https://github.com/aitech95"><img src="https://github.com/aitech95.png" width="50px" alt="JACOB SHI " /></a>
   <a href="https://github.com/Zaga-Solutions"><img src="https://github.com/Zaga-Solutions.png" width="50px" alt="Zaga Solutions" /></a>
-  <a href="https://github.com/Discolike"><img src="https://github.com/Discolike.png" width="50px" alt="DiscoLike Inc." /></a>
+  <a href="https://github.com/discolike"><img src="https://github.com/discolike.png" width="50px" alt="DiscoLike Inc." /></a>
 </p>
 <!-- SPONSORS -->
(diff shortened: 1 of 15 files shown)


falkoschindler (approved):
Thanks for following the sweep through to `echart.js`, Evan — same chained-`.observe()` mistake, same right fix.

As in #6189, I removed the accompanying test before merging. Beyond the general white-box concern (it monkeypatches `window.ResizeObserver` and asserts on observer bookkeeping instead of on user-visible behavior, plus two fixed `screen.wait(0.5)` calls), this one only compares the *total* number of disconnected observers before and after the click. It never checks that the disconnect came from the chart's own observer, so any other observer torn down during the same click would make it pass — including in a future where our fix is accidentally reverted.

Good news on the sweep itself: with these two merged, `aggrid.js`, `leaflet.js` and `plotly.js` all already keep the instance in a variable and disconnect properly, so this class of bug is now gone from the codebase.

PR #6201, opened 2026-07-25
Fix ui.color_input crash when value is None with preview enabled

evnchn:
### Motivation

color_input before/after

`ui.color_input(value=None, preview=True)` raises AttributeError at construction (line 49 -> _update_preview -> None.strip()). Equally, `ci = ui.color_input(preview=True); ci.set_value(None)` or binding the value from a source that is None crashes in _handle_value_change. Since value=None is a type-valid argument, this is trivially reachable.

Minimal reproduction (`nicegui/elements/color_input.py:66`) — the exact test/script that was run to reproduce it:

```python
"""MRE: ColorInput crashes with AttributeError when value=None and preview=True.

ui.color_input's `value` param is typed `str | None`, so `None` is a type-valid
argument. But ColorInput.__init__ calls self._update_preview() (color_input.py:49),
which at line 66 does `self.value.strip()`. When value is None that is None.strip()
-> AttributeError, crashing at CONSTRUCTION time. No server/page/browser needed.

Run:
    cd $(mktemp -d)
    /Users/evnchn/nicegui/.venv/bin/python mre.py
Expected on the buggy code: process exits with
    AttributeError: 'NoneType' object has no attribute 'strip'
"""
from nicegui import ui

# value=None is type-valid (param is `str | None`); preview=True walks the preview path.
ui.color_input(value=None, preview=True)

# Unreachable on buggy code -- the line above raises during construction.
print('NO CRASH: bug appears fixed')
```
→ `Traceback (most recent call last):`

### Implementation

One-line fix at color_input.py:66 `color = (self.value or '').strip()` so preview=True with value=None (or set_value(None)/bound None) falls through to the transparent-background branch instead of AttributeError.

<details>

- FAIL on origin/main source, PASS with fix.
</details>

### Progress

---

*Question for @falkoschindler:* `value` is typed `str | None` and the `None` case is handled at runtime here. Would you prefer to **narrow the signature** (disallow `None`) in 4.0 so this can't recur at the type level? Kept it runtime-safe for now, since a cleared color input can legitimately be `None`.

### DIFF
--- a/nicegui/elements/color_input.py
+++ b/nicegui/elements/color_input.py
@@ -63,7 +63,7 @@ class ColorInput(LabelElement, ValueElement[str | None], DisableableElement):
         if not self.preview:
             return
 
-        color = self.value.strip()
+        color = (self.value or '').strip()
         if HEX_COLOR_PATTERN_6.match(color):
             r = int(color[1:3], 16) / 255
             g = int(color[3:5], 16) / 255
--- a/nicegui/elements/date_input.py
+++ b/nicegui/elements/date_input.py
@@ -39,7 +39,7 @@ class DateInput(LabelElement, ValueElement[str | None], DisableableElement):
         with self.add_slot('append'):
             with button(icon='edit_calendar', color=None).props('flat round') as self.button:
                 with menu() as self.menu:
-                    self.picker = date().props('no-parent-event').props('range' if range_input else '')
+                    self.picker = date().props('range' if range_input else '')
 
         self.picker.bind_value(self,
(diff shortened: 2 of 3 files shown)


falkoschindler (approved):

falkoschindler:
@evnchn Regarding `None` values in 4.0: The input element is clearable (when using the prop "clearable") which causes the value to become `None`. This might be reason enough to keep supporting it. And input.py even plans to make `None` the default (instead of "") in 4.0.

PR #6204, opened 2026-07-25
Remove leaked auto route when switching a source element off a file source

evnchn:
### Motivation

`img = ui.image('/secret/report.png')` registers /_nicegui/auto/static/<hash> serving that file; `img.set_source('https://example.com/x.png')` takes the else branch, does not remove the old route -> /secret/report.png remains served at the auto route until the element is deleted, and the route registration leaks.

Minimal reproduction (`nicegui/elements/mixins/source_element.py:119`) — the exact test/script that was run to reproduce it:

```python
"""MRE: SourceElement leaks its auto-generated static route when the source
switches from a local FILE to a non-file (URL/data URI).

NiceGUI 3.13.0.post24 - nicegui/elements/mixins/source_element.py:_set_props().
When source is a file, _set_props registers /_nicegui/auto/static/<hash>/... and
stores it in self.auto_route. When source later becomes a non-file, control takes
the `else` branch, which only clears self._temp_path - it never removes the old
route nor clears self.auto_route. The file stays served and the route leaks.

Run:  cd <temp dir>; /Users/evnchn/nicegui/.venv/bin/python mre.py
Exit 1 + "BUG REPRODUCED" = bug present.
"""
import tempfile
from pathlib import Path

from nicegui import app, ui


def auto_static_routes():
    return [p for r in app.routes
            if '/_nicegui/auto/static/' in str(p := getattr(r, 'path', ''))]


# a real local file to serve (stands in for '/secret/report.png')
secret = Path(tempfile.mkdtemp()) / 'report.png'
secret.write_bytes(b'\x89PNG\r\n\x1a\n')  # minimal file; is_file() only checks existence

img = ui.image(str(secret))              # is_file True -> registers auto static route
assert img.auto_route is not None, 'setup failed: no auto_route registered'
assert len(auto_static_routes()) == 1, 'setup failed: route not registered'

img.set_source('https://example.com/x.png')   # non-file -> else branch

leaked = auto_static_routes()
print("img._props['src'] =", img._props['src'])
print('img.auto_route    =', img.auto_route)
print('leaked auto static routes still serving the local file:', leaked)

if leaked:
    print('BUG REPRODUCED: switching to a non-file source left the old static '
          'route registered (the local file remains served until the element is deleted).')
    raise SystemExit(1)
print('OK: no route leaked')
```
→ `img._props['src'] = https://example.com/x.png`

### Implementation

Fixed source_element.py _set_props else-branch to remove leaked auto route + clear auto_route on file->non-file switch (mirrors is_file branch & _handle_delete).

<details>

- FAILS on origin/main with real leak symptom, PASSES with fix).
</details>

### Progress


falkoschindler (approved):
The diagnosis is spot-on and the fix correctly mirrors `_handle_delete` — nice find, and thanks for the exemplary PR description with MRE and verification notes!

I pushed a small follow-up commit with three adjustments:

1. The `FileNotFoundError` check now runs at the top of `_set_props()`, *before* the route cleanup. Otherwise a failed `set_source(Path('/nonexistent.png'))` would have removed the still-working route while `_props['src']` kept pointing at it, breaking the previously displayed image. Now a failed call leaves the element untouched.
2. Since both branches started with the same `remove_route` code, I hoisted it above the `is_file` split — the "old route is always cleaned up" invariant now lives in one place.
3. The new test is parametrized over `ui.image` and `ui.video` to also cover the `SOURCE_IS_MEDIA_FILE` path (`/_nicegui/auto/media/`).

PR #6206, opened 2026-07-25
Route mouse-button and exact event modifiers correctly

evnchn:
### Motivation

`ui.button().on('click.right', handler)` or `element.on('mousedown.middle', handler)` never triggers the Python handler at all (click/mousedown events have no `.key`, so withKeys short-circuits). Likewise `element.on('keydown.enter.exact', handler)` silently ignores the .exact constraint. A user trying to distinguish left/right/middle click or require an exact modifier combo gets a handler that either never fires or ignores the modifier.

Minimal reproduction (`nicegui/event_listener.py:30`) — the exact test/script that was run to reproduce it:

```python
"""MRE for NiceGUI bug: mouse-button (.left/.right/.middle) and .exact event
modifiers are misrouted into `keys` instead of `modifiers`.
"""
from nicegui import ui
from nicegui.page import page as page_deco


def routing_for(spec: str) -> dict:
    """Drive the real public API el.on(spec, ...) and read back the wire dict."""
    with Client(page_deco('/'), request=None):
        el = ui.button('X').on(spec, lambda: None)
        return list(el._event_listeners.values())[-1].to_dict()


print(f'{"spec":<22}{"modifiers":<22}{"keys"}')
for spec in ['click.left', 'mousedown.middle', 'mousedown.right', 'keydown.enter.exact']:
    d = routing_for(spec)
    print(f'{spec:<22}{str(d["modifiers"]):<22}{d["keys"]}')

# Control: a REAL whitelisted modifier routes correctly -> proves the mechanism.
ctrl = routing_for('click.ctrl')
print(f'\ncontrol click.ctrl -> modifiers={ctrl["modifiers"]} keys={ctrl["keys"]}')
assert ctrl['modifiers'] == ['ctrl'], 'sanity: ctrl should be a modifier'

left = routing_for('click.left')
exact = routing_for('keydown.enter.exact')
if 'left' in left['keys'] and 'exact' in exact['keys']:
    print('\nBUG REPRODUCED: left/middle/right/exact land in `keys` -> wrapped in')
    print('Vue.withKeys -> handler never fires on click/mousedown; .exact ignored.')
else:
    print('\nFIXED: mouse-button/exact modifiers now routed to `modifiers`.')
```
→ `spec                  modifiers             keys`

### Implementation

Fixed EventListener.to_dict misrouting: exact/middle now always modifiers, left/right are mouse-button modifiers except on keyboard events (arrow keys preserved), mirroring Vue's compiler-dom vOn transform.

<details>
<summary>How new vs old differ (modifier routing) + Codex review</summary>

Old modifier set was `{stop, prevent, self, ctrl, shift, alt, meta}`, so these fell through to *key filters*:

| token on event | OLD routed as | NEW routed as |
|---|---|---|
| `.exact` (any event) | key ❌ | modifier ✓ |
| `.middle` (any event) | key ❌ | modifier ✓ (middle mouse button) |
| `.left` / `.right` on a mouse event (e.g. `click.right`) | key ❌ | modifier ✓ (mouse button) |
| `.left` / `.right` on `keydown`/`keyup`/`keypress` | key ✓ | key ✓ (arrow key — unchanged) |

Effect: e.g. `ui.on('click.right', ...)` previously treated `right` as a key filter that a mouse event never carries, so the handler effectively never fired; now it fires on right-click. `keydown.left` (arrow key) stays a key in both. Mirrors Vue's compiler-dom `vOn` transform.

**Codex (adversarial, branch diff):** *approve / SHIP* — "the new routing covers the previously misclassified `.exact`, `.middle`, `.left`, `.right` cases, and the keyboard exception set `{keydown,keyup,keypress}` is consistent with where `.left`/`.right` should remain key filters. No material findings."
</details>

<details>

- fails [] == ['left'] on origin/main, passes with fix).
</details>

### Progress


falkoschindler (approved):
Great catch, and a textbook fix — the new routing mirrors Vue's compiler-dom `resolveModifiers` exactly (`exact`/`middle` unconditionally, `left`/`right` per event type via the keyboard-event check). Fun bit of archaeology: mouse-button modifiers actually worked before 1d8c7dec3 (Jan 2023), when all suffixes were passed to `withModifiers` wholesale; introducing the key-filter categorization with `keys` as the catch-all silently regressed them. So this restores 2017-era Vue semantics that we lost along the way.

I pushed 450ce2d53 on top: it replaces the white-box `to_dict()` test with an end-to-end `Screen` test (left/right/middle click via Selenium, Shift+Enter as negative case for `.exact`, ArrowLeft for the arrow-key exception) and adds a short comment explaining the keyboard-event condition. The new test fails on main and passes with your fix.

PR #6209, opened 2026-07-25
Keep JavaScript source maps out of the wheel

evnchn:
**TL;DR — the wheel is 20.15 MB and 11.75 MB of that (58%) is JavaScript source maps that nothing at runtime reads. Two lines of `pyproject.toml` drop them from the wheel while keeping them in the sdist.**

### Motivation

The published wheel contains 178 vendored `*.map` files. Measured by building the wheel at `ff073f46` with and without this change, they are **11,750,854 B of the 20,154,222 B compressed wheel — 58.3%** — and about 44 MB of the unpacked install tree.

They are not required at runtime: Python never imports them, and a browser only requests one when a developer has devtools open on the corresponding bundle. Every user pays the download and the disk regardless.

Source maps are still worth having, so this removes them from the *wheel* only and keeps them in the *sdist*, where anyone debugging from a source checkout still gets them.

The saving carries over to packaged apps as well: PyInstaller bundles the package's data files, so every app built with `nicegui-pack` also shrinks by roughly 44 MB unpacked.

### Implementation

```toml
[tool.poetry]
exclude = ["nicegui/**/*.map"]
include = [{ path = "nicegui/**/*.map", format = "sdist" }]
```

`exclude` drops the maps from every artifact; the format-scoped `include` adds them back for the sdist only. The two compose — verified by building both artifacts rather than by reading the poetry docs.

**Accepted trade-off:** bundles still end with `//# sourceMappingURL=index.js.map`, so on a *wheel* install devtools gets a 404 for the map of a bundle it opens. This is handled cleanly — `_get_library` in `nicegui/nicegui.py` already falls through to `HTTPException(404)` when the file is absent, so nothing breaks — but it is a real, if small, capability being traded for 58% of the wheel.

<details>
<summary><b>Measurement — both artifacts built at the same commit</b></summary>

Built with `uv build` at `ff073f46`, so this is not a comparison against a different released version:

| | files | `.map` files | compressed |
|---|---:|---:|---:|
| baseline | 827 | 178 | **20,154,222 B** |
| this PR | 649 | 0 | **8,403,368 B** |

**Delta: −11,750,854 B (−58.3%).** The file-count drop is exactly 178, matching the map count.

Sdist from the same build: **826 files, all 178 `.map` files present** — so `include` scoped to `format = "sdist"` does what it claims.

For scale: `plotly`'s `index.js.map` alone is ~3.1 MB compressed, larger than the plotly bundle it describes.

</details>

### Possible follow-up (not in this PR): `nicegui[source-map]`

If losing maps from the wheel is unattractive, there is a nicer end state that this PR does not foreclose — making them an installable extra, so the wheel stays thin *and* `pip install nicegui[source-map]` restores them without needing a source checkout.

<details>
<summary><b>How that would work, and what it would cost</b></summary>

An extra pulls in **distributions**; it cannot add files to the distribution declaring it. So `nicegui[source-map]` would need a companion distribution carrying the `.map` files:

```toml
[project.optional-dependencies]
source-map = ["nicegui-source-map==<exact version>"]
```

There is already precedent for exactly this shape in the project: `highcharts = ["nicegui-highcharts>=3.3.0,<4"]` points at `zauberzeug/nicegui-highcharts`, a separate repo and separate PyPI distribution.

**The payoff is more than packaging tidiness.** With the companion registered through the existing `optional_features` mechanism —

```python
optional_features.try_register('source-map', package='nicegui_source_map')
```

— Python would *know* whether maps are available, and could suppress or rewrite the `sourceMappingURL` comment when they are not. That removes the devtools 404 this PR accepts, which the sdist route cannot do.

**The one genuinely new problem is version lockstep.** Highcharts can use a loose range because it is an independent component. Source maps must match the exact bundles they describe — maps from 3.15 against 3.16 bundles would give silently wrong line numbers, which is worse than no maps at all. So it wants an exact pin generated from the same `poetry-dynamic-versioning` value, so the two always ship together.

Raising it here in case it is wanted as a developer-experience improvement; happy to prototype it separately if there is interest.

</details>

### Progress

- [x] Pytests are not necessary. (Packaging-only change; verified by building and inspecting both artifacts.)
- [x] No breaking changes to the public API. (The behaviour change is limited to devtools source-map availability on wheel installs, described above.)


falkoschindler (approved):
Thanks, great PR — the motivation is thorough and every number checked out when I rebuilt both artifacts at the PR head myself: 0 `.map` files in the wheel (8.5 MB, down from 20.2 MB), all 178 in the sdist. Since `uv build` builds the wheel from the sdist, the exclusion also holds on the pip fallback path when no wheel is available. And `_get_library` already falls through to a clean 404 for missing maps, so the accepted trade-off is exactly as described.

The review turned up one real issue, which I fixed directly on the branch (efeca7ca5) along with two smaller improvements:

- **Anchored the sdist `include` patterns.** poetry-core applies the VCS-ignore filter to the package itself, but not to explicit `include` entries — so with `nicegui/**/*.map`, a maintainer who has run `npm.py` (leaving ~2700 `.map` files under `nicegui/elements/*/node_modules/`) would silently ship them all in a locally built sdist. Reproduced with a planted gitignored map, then verified the fix: `nicegui/static/*.map` and `nicegui/elements/*/dist/*.map` cover all 178 tracked maps, and the planted file no longer leaks.
- **Added a CI guard.** The publish workflow now fails before `uv publish` if the wheel contains any `.map` file, so a future packaging edit can't silently reintroduce the 11.75 MB.
- **Extended the motivation** with the packaged-app upside: PyInstaller bundles the package's data files, so every `nicegui-pack`ed app also shrinks by ~44 MB unpacked.

The `nicegui[source-map]` extra is an interesting follow-up idea — happy to discuss it separately if the wheel-install debugging story ever becomes a pain point.

PR #6210, opened 2026-07-25
Avoid downloading a font subset for the connection-lost glyph

evnchn:
**TL;DR — the ⚠️ in the hidden connection-lost popup makes every first page view download a 16.5 KB Roboto symbols font, which the browser then discards in favour of a system emoji font. One declaration stops the download with byte-identical rendering.**

### Motivation

`nicegui.css` renders the connection-lost popup's warning sign as `content: "⚠️"`. That glyph is U+26A0, which falls inside the `unicode-range` of the Roboto **symbols** subset declared in `fonts.css`.

So the browser fetches `fonts/2474057f3040a08c.woff2` — **16,548 bytes** — on the first view of *any* page, including a page whose entire content is the word `hi`, for a popup that is hidden.

The wasteful part: because the source is `⚠️` (U+26A0 followed by variation selector U+FE0F), the browser requests emoji presentation and paints the glyph with a **system colour emoji font anyway**. The Roboto subset is downloaded and never used.

### Implementation

```css
font-family: sans-serif; /* anything but Roboto, whose symbol subset would be fetched just for this glyph */
```

on both the `:dir(ltr)` and `:dir(rtl)` rules.

The mechanism is simply *not matching the `@font-face` family*. `sans-serif` is chosen deliberately over an emoji font stack (`"Apple Color Emoji", …`) because **it leaves rendering byte-identical to today**, whereas naming an emoji font makes it the primary family, which then supplies the metrics and renders the glyph slightly smaller. This PR is intended to change bytes, not pixels.

<details>
<summary><b>Before / after in real headless Chromium</b></summary>

Same hello-world app (`ui.label('hi')`), two checkouts differing only by this commit, all responses recorded, popup forced visible and screenshotted:

| | woff2 fetched | symbol subset | popup screenshot |
|---|---:|---|---|
| before | 2 | `2474057f3040a08c.woff2` **downloaded** | `bfebaa3b636e…` |
| after | 1 | **avoided** | `bfebaa3b636e…` |

**−16,548 B and one fewer request on every first page view, with an identical screenshot hash.**

</details>

<details>
<summary><b>Why not <code>unset</code>, <code>revert</code> or <code>inherit</code></summary>

They don't work — `font-family` is an *inherited* property, so all three resolve back to the ancestor's `Roboto` and the download returns. Measured:

| `font-family` | symbol font | popup screenshot |
|---|---|---|
| *(no declaration — today)* | **DOWNLOADED** | `bfebaa3b…` |
| `unset` | **DOWNLOADED** | `bfebaa3b…` |
| `revert` | **DOWNLOADED** | `bfebaa3b…` |
| `inherit` | **DOWNLOADED** | `bfebaa3b…` |
| `initial` | avoided | `bfebaa3b…` |
| `sans-serif` | avoided | `bfebaa3b…` |

`initial` also works (it computes to `Times`), but `sans-serif` states the intent more plainly for a UI popup.

</details>

<details>
<summary><b>Why this is a per-page cost, not a one-off</b></summary>

The popup is hidden with `visibility`, not `display: none`, so it is still laid out and its `::before` content still triggers font matching. The font is therefore fetched on a normal page load, not only when a connection actually drops.

</details>

### Progress

- [x] Pytests are not necessary. (Verified by measuring real browser requests and comparing screenshot hashes before/after.)
- [x] No breaking changes to the public API. (Rendering is byte-identical; only the font fetch is removed.)


falkoschindler (approved):
Nice find and an exemplary write-up — measured before/after, alternatives ruled out with data. I verified the mechanism on the branch: U+26A0 sits in the symbols subset's `unicode-range`, the popup is hidden via `visibility` (not `display: none`), so the fetch really is a per-page cost — and with the fix, a real browser run requests only the latin subset.

I took the liberty of pushing a small follow-up commit:

- **`font-family` now names the inherited stack minus Roboto** (`"-apple-system", "Helvetica Neue", Helvetica, Arial, sans-serif`) instead of bare `sans-serif`. The download is avoided either way — all that matters is not matching the `@font-face` family — but keeping the intermediate families makes rendering match today's as closely as possible on all platforms, not just where `sans-serif` happens to resolve equivalently.
- **Marked the comment with `NOTE:`** so the declaration isn't mistaken for a stray style rule and removed in a future cleanup — it only exists because of the hidden coupling to `fonts.css`.
- **Hoisted the direction-independent declarations** (`content`, `font-family`, `position`) into a shared rule, leaving only `left`/`right` in the `:dir()` rules.

Thanks for the meticulous analysis!

PR #6213, opened 2026-07-28
Fix On Air reconnect when Socket.IO's connected flag is stale

### LINKED ISSUE #6212: On Air: client never reconnects when the relay silently drops the device (relay.connected stays True)
### Description

A device on air can become permanently unreachable while the process keeps running: the relay no longer knows the device (its URL returns 404), but the socket.io client still reports `connected == True`, so the reconnect timer in `Air.__init__` (`timer(5, self.connect)`) never acts — `connect()` returns early on `if self.relay.connected: return`. The only remedy is restarting the process.

Observed twice within days on the same machine (Linux, air-link 0.5.0, NiceGUI 2.24.2, python-socketio 5.16.3, python-engineio 4.13.3, websocket transport), both times right after a network change (NetworkManager event, Wi-Fi/DHCP). The logic is unchanged on current `main`, so 3.x should be affected as well.

Two log signatures of the stuck state:

1. `connected.` arrives but the `ready` event never does — no `NiceGUI is on air at ...` line, no further log output, and the relay answers 404 from then on:

```
DEBUG:nicegui.air:Going to connect...
DEBUG:nicegui.air:Connecting...
DEBUG:nicegui.air:connected.
(nothing ever again)
```

2. A reconnect loop where the transport dies ~15 s after every handshake, which eventually goes silent — again with `relay.connected` stuck at `True` and no open TCP connection to the relay (verified via the process's socket list):

```
DEBUG:nicegui.air:connected.
NiceGUI is on air at https://europe.on-air.io/dennis/tp/
ERROR:engineio.client:packet queue is empty, aborting
DEBUG:nicegui.air:disconnected.
DEBUG:nicegui.air:Going to connect...
... (repeats every ~20 s, then stops with the client believing it is connected)
```

### Root cause

`air.py` trusts `self.relay.connected` as the sole signal for "the tunnel works". That flag is maintained by python-socketio and can stay `True` when the engineio transport dies without a clean disconnect event (the `packet queue is empty, aborting` path). There is no application-level liveness check, so the 5-second keep-alive timer becomes a no-op exactly when it is needed.

### Suggested solution

Two complementary pieces, both in `Air`:

1. **Cheap consistency check every timer tick**: `self.relay.connected` can disagree with the actual engineio state. Reconciling them makes the existing timer effective again in the crashed-transport case:

```python
async def connect(self) -> None:
    if self.connecting:
        return
    if self.relay.connected:
        if self.relay.eio.state == 'connected':
            return
        self.log.warning('socket.io claims connected but engine.io is not; resetting connection')
        await self.disconnect()
    ...
```

2. **End-to-end check as the robust guarantee** (also covers the "connected but `ready` never arrived / relay forgot the device" case, where the socket may look healthy): every Nth timer tick, fetch `self.remote_url` over plain HTTPS; if the relay answers 404 while the client claims to be connected (twice in a row, to avoid acting on transients), call `self.disconnect()` so the timer reconnects. Deliberately do nothing when the relay is unreachable — that is an internet outage, and reconnecting would not help.

Item 2 is implemented downstream in air-link as a workaround: https://github.com/zauberzeug/air-link/pull/32 (`air_link/relay_check.py`) — feel free to lift it into `air.py`.

It may also be worth reporting the stale `connected` flag to python-socketio, since a transport death without a disconnect event looks like a bug there too; but NiceGUI should be defensive regardless, because for on-air devices "believes it is connected forever" means permanently losing remote access.

evnchn:
### Motivation

Fixes https://github.com/zauberzeug/nicegui/issues/6212

An On Air device can go permanently unreachable while the process keeps running. Restarting is the only cure.

To decide whether it still needs to reconnect, `Air` checks exactly one thing — the `connected` flag on the socket.io client:

```python
if self.relay.connected:
    return  # nothing to do
```

**The bug: that flag can get stuck on `True` after the connection has died.** When that happens the keep-alive timer returns early every 5 seconds, forever, and never reconnects.

#### How the flag gets stuck

The socket.io library sets the flag at the *very end* of connecting. But the cleanup that would clear the flag only runs **if the flag is already set**. A connection that dies in between falls straight through the gap.

```mermaid
sequenceDiagram
    autonumber
    participant A as Air.connect()
    participant S as socket.io client
    participant T as network transport

    A->>S: connect to the relay
    S->>T: open connection
    T-->>S: connected, handshake done
    Note over S: flag is still False here —<br>it only gets set at the very end
    T--xS: the transport dies right here
    S->>S: cleanup runs: "if the flag is set, reset everything"<br>the flag is False, so it does nothing
    S->>S: connect() finishes and sets the flag = True
    Note over A,T: flag says "connected". The transport is dead.<br>Nothing will ever clear it again.
```

Same thing as a table:

| Moment | Flag | What happens |
|---|---|---|
| During connect | `False` | Transport dies. Cleanup runs, sees `False`, does nothing. |
| End of connect | → `True` | `connect()` sets it unconditionally. |
| Forever after | `True` | Timer returns early every tick. Device is unreachable. |

This is the reported log signature exactly: `connected.` is logged, `ready` never follows, and the process holds no socket to the relay.

### Implementation

**1. When the flag disagrees with reality, throw the client away.** On each timer tick, if socket.io claims to be connected but the transport is not, discard the wedged client, build a fresh one, and connect that.

Rebuilding rather than repairing is deliberate: it uses only the public API and cannot be defeated by internal state we forgot to reset. (Upstream's own `AsyncSimpleClient.disconnect()` does the same thing.) Handler registration therefore moves into a `_create_relay()` method. Beyond the 16 decorators, two handler bodies change: `_handle_connect_error` now closes over the local `relay` instead of `self.relay`, so a discarded client disconnects *itself* rather than whichever client is current, and `_handle_read_stream` uses `anext()` instead of `__anext__()`.

The obvious cheaper move — "just call `disconnect()` and let it reset things" — cannot work:

```mermaid
flowchart TD
    A["The flag is wrong.<br>Can we just call disconnect()?"] --> B["disconnect() asks the transport to close"]
    B --> C{"Is the transport still alive?"}
    C -->|"No, it is already dead"| D["disconnect() does nothing at all"]
    D --> E["so the 'disconnected' event never fires"]
    E --> F["so the flag is never cleared"]
    F --> G["and the next connect() refuses:<br>'Already connected'"]
    G --> H["a warning every 5 seconds, forever,<br>and still no reconnect"]
```

`shutdown()` fails the same way, for the same reason — it delegates to `disconnect()` whenever `connected` is set.

**2. `socketio.AsyncClient(reconnection=False)`, so the keep-alive timer is the only reconnect driver.** The client previously also ran socket.io's own auto-reconnect, so two independent loops were reconnecting one client.

To be clear about what this does and does not do: **it is not what fixes the bug above** — change 1 is. What it fixes is a leak that the two-driver arrangement causes, and that leak is present in the primary bug scenario as well as after an ordinary drop.

`_handle_eio_disconnect` *does* fire while the flag is being wedged — that is precisely how the flag-clearing gets skipped, since only its cleanup sits inside `if self.connected:` and the flag is still `False` at that moment. But `will_reconnect = self.reconnection and self.eio.state == 'connected'` is computed independently of that flag, at a point where the read loop has not yet reset the state, and the task is started *outside* that guarded block. So on the pre-PR code a reconnect task **does** start in the wedged state, and then spins forever on `ConnectionError('Already connected')` exactly as described below. Change 2 removes it there too.

After an ordinary transport drop lasting long enough for socket.io's doubling backoff to fall behind our fixed 5 s tick, our timer reconnects first. Socket.io's `_handle_reconnect` then calls `connect()`, gets `ConnectionError('Already connected')`, swallows it via `except (ConnectionError, ValueError)`, and — with the default `reconnection_attempts=0` — never terminates. `_reconnect_task` is only cleared on the success path it can now never reach, so a task spins every 5 s for the life of the process, permanently occupying socket.io's global `reconnecting_clients`, and a later drop is serviced by that stale loop rather than a fresh one.

**3. Stream teardown is hardened while it is being extracted.** `_close_streams()` — now shared by `disconnect()` and the reconnect path — snapshots and clears `self.streams` before awaiting, and guards each `aclose()`, so an in-flight `close-stream` handler can no longer raise `RuntimeError: dictionary changed size during iteration` and one failing close no longer skips every remaining stream. Because that clears the dict up front, `_handle_close_stream` also becomes idempotent (`pop(..., None)`); it previously raised `KeyError` for a stream the teardown had already taken, and `_handle_read_stream` called it from its own `except`, so a missing stream faulted twice. These predate this PR, but the helper now also runs on the reconnect path.

Change 2 also makes discarded clients genuinely inert: one with auto-reconnect still enabled could open a **second** connection to the relay behind our back, and `shutdown()` cannot stop it, because the stale `connected == True` sends it down the `disconnect()` path above rather than the abort path.

<details>
<summary>Source trace, for reviewers who want the exact mechanism</summary>

**The race** — in `socketio/async_client.py`, `connect()` awaits `_connect_event` and only then executes `self.connected = True` as its final statement. `_handle_eio_disconnect()` wraps its cleanup (`self.namespaces = {}`, `self.connected = False`) in `if self.connected:`. The disconnect is triggered from Engine.IO's read-loop transport-error path, a separate task, so it interleaves at any `await` inside `connect()`. Because the cleanup was skipped, `self.namespaces` survives, so `connect()`'s namespace check passes and it falls through to the assignment.

**Why `disconnect()` cannot clear the flag** — `disconnect()` sends namespace DISCONNECT packets and calls `eio.disconnect()`. In `engineio/async_client.py`, `_send_packet()` returns immediately unless `state == 'connected'`, and `disconnect()` has its *entire body* guarded by `if self.state == 'connected'` — so on a dead transport it only calls `_reset()` and never emits the `disconnect` event, which is the only thing that resets Socket.IO's flag. `shutdown()` delegates to `disconnect()` whenever `connected` is set.

**The duelling drivers** — `reconnection=True` and `reconnection_attempts=0` are the defaults. With our timer also running, whichever loses calls `connect()` on an already-connected client, gets `Already connected`, swallows it via `except (ConnectionError, ValueError)`, and with infinite attempts never terminates. I suspect this also explains the `ValueError` this module already catches with the comment *"this sometimes happens when the internal socketio client is not yet ready"* — that is `engineio.connect()` raising `Client is not in a disconnected state` because the other driver got there first.

Checked against the versions in the report: `socketio/async_client.py` is byte-identical between 5.16.1 and 5.16.3, and the only `engineio/async_client.py` delta from 4.13.1 to 4.13.3 is inside `_write_loop`, below everything involved here.

**Reproduced standalone, without NiceGUI.** A 74-line script using only `python-socketio` + `aiohttp` — no monkeypatching and no writes to any client attribute — wedges the client deterministically on socketio 5.16.3 / engineio 4.13.3 (current PyPI latest): **25/25 on macOS, 20/20 on Linux**, with a control run that adds a single 0.1 s server-side delay reproducing **0/25**.

```
after connect() ... connected=True   eio.state='disconnected'   ws_closed=True namespaces={'/': 'NS'}
after disconnect().. connected=True   eio.state='disconnected'   ws_closed=True namespaces={'/': 'NS'}
after shutdown() ... connected=True   eio.state='disconnected'   ws_closed=True namespaces={'/': 'NS'}
reconnect .........  ConnectionError('Already connected') -- client is wedged
```

That confirms the mechanism is real and lives entirely in the library, not in NiceGUI's usage of it.

</details>

<details>
<summary>Trade-offs, and two things deliberately left out</summary>

**Reconnect latency** after a clean drop goes from socket.io's ~1 s first backoff to our fixed 5 s tick. That seemed worth predictable behaviour, but I'm happy to shorten the timer instead.

**A residual window I did not close:** the post-connect `assert self.relay.connected` checks the Socket.IO flag, not `eio.state`, so a connection that dies inside the same window can still return "successfully" — leaving `emit()` briefly trusting a stale flag until the next timer tick heals it. I left the assertion alone rather than tighten it into something that could raise in production. Worth knowing it bounds the fix to "self-healing within 5 s" rather than "impossible".

**The 404 check is not here.** The issue also suggests fetching the public device URL and reconnecting when the relay 404s. That deserves its own PR: it adds HTTP polling, a threshold and a timer, and on its own it cannot recover anyway, because it also recovers via `disconnect()`. The same caveat applies to lifting `relay_check.py` from air-link as-is. With this PR in place it becomes genuine belt-and-braces.

**Upstream:** the assignment ordering in `python-socketio`'s `connect()` looks wrong independent of NiceGUI, and once the flag is stale the library offers no public way to recover the client. Worth reporting there too — but that would not retire this PR, since NiceGUI accepts `python-socketio>=5.16.2` with no upper bound, so users would stay exposed until the floor is raised.

</details>

### Progress

<details>

**End-to-end, against a fake relay driving a real `Air`.** The relay accepts the namespace and closes immediately, which wedges the client for real (`connected=True eio.state='disconnected' remote_url=None`):

| build | recovers? | recovery time | `Air.connect()` calls |
|---|---|---|---|
| unpatched | **never** (watched 65 s) | — | 16, all early-returns |
| this PR | yes, 3/3 runs | **4.05–4.06 s** | 4 |

The warning fires once per 5 s tick while wedged — two in total, not a spam loop.

The leak that change 2 prevents was measured the same way, using an 8 s outage so socket.io's backoff falls behind the timer. With auto-reconnect left on, `_reconnect_task` stays `PENDING` after recovery and its spin count climbs monotonically (8 → 12 → 16 over a 60 s soak, one per 5 s, `reconnecting_clients=1`), reproduced 3/3. With `reconnection=False` it never occurs. It is one leaked task per process, not N, and each iteration is cheap — the problem is that it never terminates.

Caveat worth stating: this is a *fake* relay. It establishes the client-side behaviour of `Air` and python-socketio; it says nothing about what the real relay does.

**One pytest is included** — a contract test asserting that a fresh `socketio.AsyncClient()` reports `eio.state == 'disconnected'`. That is the real-library invariant the guard rests on, and `relay.eio` plus its state strings are undocumented internals that `pyproject.toml` puts no upper bound on, so an upstream rename should fail loudly rather than silently leave the guard never matching. No fakes, no network.

**Reaching the guard itself from a test is still not attempted.** That means faking the stale state and patching library internals, which is too brittle to be worth its weight — the standalone reproduction above is far better evidence. I did write such a test while developing, and confirmed it fails against the unpatched `air.py` (`assert 0 == 1`) and passes with the fix, so the branch is doing real work; it is simply not carried in the diff.

Local gates: `mypy ./nicegui` clean (245 files), `pylint ./nicegui` 10.00/10, `pre-commit` all hooks pass.

Full local suite: **991 passed, 9 failed, 3 skipped, 2 xfailed** in 15:14. All 9 failures reproduce unchanged on a clean `origin/main` worktree with none of this branch's code present, so they are pre-existing and local to my environment rather than caused by this PR:

- `tests/test_page.py::test_dark_mode` ×4 and `tests/test_dark_mode.py::test_dark_mode` ×4 — colour/class assertions
- `tests/test_lazy_imports.py::test_module_access_does_not_import_others` — its subprocess shells out to a bare `python3`, which on my machine is not the environment's interpreter and fails with `ModuleNotFoundError: No module named 'socketio'`

The 3 skips are `tests/test_run.py` needing a fork-default platform (Linux); none cover this change.

</details>


denniswittich:
Looks reasonable to me, thx @evnchn

evnchn:

**Updated** @denniswittich — the guard now skips a disconnect already in progress, and this PR carries its first tests.

<details><summary><b>What changed</b></summary>

`engine.io` has three states, not two. The previous guard returned early only on `'connected'`, so a keep-alive tick landing during a **deliberate** disconnect (`app.shutdown()` → `air.disconnect()`, timer firing every 5 s) read `'disconnecting'` as a stale flag, replaced the client, and reconnected a device on its way down.

```diff
-            if self.relay.eio.state == 'connected':
+            if self.relay.eio.state != 'disconnected':
```

`tests/test_air.py` is new — `nicegui/air.py` had no tests before. Each case was confirmed to fail on the code lacking its fix:

| `air.py` under test | stale-flag | healthy | disconnecting |
|---|---|---|---|
| `main` (no fix) | ❌ `assert 0 == 1` | ✅ | ✅ |
| this PR before 635e560 | ✅ | ✅ | ❌ `assert 1 == 0` |
| this PR at 635e560 | ✅ | ✅ | ✅ |

Mechanism verified against the real installed `python-socketio`, not a mock: with `connected=True` and a dead transport, `disconnect()` leaves the flag set and the next `connect()` raises `ConnectionError('Already connected')`, which `air.py` swallows at debug level — the permanent stuck state this fixes.

Not yet verified: the new test file has only been run locally on 3.14; the 3.10 CI leg was still queued at the time of writing.

</details>

falkoschindler:
Thanks for this exceptional PR, @evnchn! The diagnosis is correct and the write-up is exemplary.

A few findings, mostly minor:

1. **Fold the stale-client handling into the guarded region** (major) — The wedge branch runs `await self._close_streams()` _before_ `self.connecting = True` is set. During that await, a concurrent `connect()` call (module-level `connect()`, or `_handle_out_of_time`) can pass the guard and start connecting the fresh client; when the first call resumes it gets `ConnectionError('Already connected')` (swallowed), and its `finally` clears `connecting` while the second call is still mid-connect. And if an `aclose()` raises, the tick is lost and `self.streams` keeps its stale entries. Restructuring closes both and lets the unreachable `if self.relay.connected: await asyncio.wait_for(self.disconnect(), timeout=5)` branch you mentioned go away instead of sitting next to logic that now looks like it (which also frees the `asyncio` import, whose only use this is):

   ```python
       async def connect(self) -> None:
           """Connect to the NiceGUI On Air server."""
           if self.connecting:
               self.log.debug('Already connecting.')
               return
           if self.relay.connected and self.relay.eio.state != 'disconnected':
               return  # connected, or a deliberate disconnect is in progress
           self.log.debug('Going to connect...')
           self.connecting = True
           try:
               if self.relay.connected:  # a stale flag cannot be cleared via disconnect(), so discard the client
                   self.log.warning('Socket.IO claims to be connected while Engine.IO is not. Replacing the client.')
                   self.relay = self._create_relay()
                   await self._close_streams()  # the discarded client can no longer deliver "close-stream"
               self.log.debug('Connecting...')
               await self.relay.connect(...)
   ```

   As a side effect, the early-return comment becomes clearer: `# "disconnecting" means a deliberate disconnect is running` hangs off a `!=` test and describes the branch _not_ taken, so it takes a second pass to see which case it covers; `return  # connected, or a deliberate disconnect is in progress` says what the return means.

2. **Harden `_close_streams()` while it is being extracted** (minor) — It iterates `self.streams.values()` across `await`s, so an in-flight `close-stream` handler task running mid-teardown raises `RuntimeError: dictionary changed size during iteration`, and a single failing `aclose()` skips every remaining stream. Both were pre-existing in `disconnect()`, but the extraction is the natural moment to fix them, and the helper is now also on the reconnect path. Snapshot-and-clear first:

   ```python
       async def _close_streams(self) -> None:
           streams, self.streams = self.streams, {}
           for stream in streams.values():
               try:
                   await stream.response.aclose()
               except Exception:
                   self.log.debug('Could not close stream.', exc_info=True)
   ```

3. **The description undersells change 2 on one point** (minor) — It says that in the stale-flag state "socket.io's reconnect loop never even starts, because `_handle_eio_disconnect` never fires". The handler _does_ fire — that is precisely how the flag-clearing gets skipped, since only its cleanup sits inside `if self.connected:` — and `will_reconnect` is computed as `self.reconnection and self.eio.state == 'connected'`, independent of the flag, at a moment when the read loop has not yet reset the state. So on the pre-PR code a reconnect task **does** start in the wedged state and then spins forever on `ConnectionError('Already connected')`. Change 2 therefore also removes a forever-spinning task in the primary bug scenario, not only in the backoff-race scenario. Since this PR body is the reference write-up for #6212, worth correcting. (Two smaller drift items: "the handler bodies are untouched" isn't quite true — `_handle_connect_error` now closes over `relay` instead of `self.relay`, a real and correct change worth naming, and `_handle_read_stream` switched to `anext`.)

4. **Carry one contract test for the library internals the fix rests on** (minor) — `relay.eio` and its state strings are undocumented internals, and pyproject has no upper bound on python-socketio/python-engineio; if upstream renames them, the guard silently never matches and we are back to today's behavior. I agree that the white-box reconnect tests weren't worth their weight, but a one-line test against the real library — no fakes, fails loudly on an upstream refactor — pins the assumption:

   ```python
   def test_engineio_state_contract() -> None:
       assert socketio.AsyncClient().eio.state == 'disconnected', \
           'Air.connect() detects a stale Socket.IO connection via this Engine.IO state'
   ```

Agreed on leaving the post-connect assertion, the 404 check, and the upstream report out of this PR — and yes, the assignment ordering in `python-socketio`'s `connect()` deserves an upstream issue in any case.

evnchn:

**All four applied** — three as commits, and the finding 3 correction is in the PR description above. Thank you for the read; finding 1 in particular is a real hole I had left open.

One change beyond your asks, flagged for your call: snapshot-and-clear in `_close_streams()` empties `self.streams` *before* the awaits, which turns a `close-stream` arriving mid-teardown from a lucky hit into a guaranteed `KeyError` — so `_handle_close_stream` is now idempotent. Happy to drop 6e11b480 if you would rather keep this PR to the four.

<details>
<summary><b>What changed</b></summary>

| Finding | Commit | Change |
|---|---|---|
| 1 — fold into the guarded region | 0a28ce51 | Your shape, adopted as written: the guard becomes `if self.relay.connected and self.relay.eio.state != 'disconnected': return  # connected, or a deliberate disconnect is in progress`, and the discard-and-rebuild moves inside the `try` after `self.connecting = True`. The unreachable `wait_for(self.disconnect(), timeout=5)` branch is gone and the `asyncio` import with it. |
| 2 — harden `_close_streams()` | 7cfd91dc | Your snapshot-and-clear plus per-stream `try`/`except`, as written. |
| 3 — description undersells change 2 | *(description edit)* | Rewritten, plus both drift items you named. See below. |
| 4 — contract test | 88f9f915 | Your one-liner, as written, in a new `tests/test_air.py`. |
| — | 6e11b480 | Consequence of finding 2; the extra one described above. |

**Finding 2, before/after.** Both failure modes you named reproduce against the old body and are gone with the new one (stand-alone harness, fake streams, not committed):

```
mutation  OLD: RuntimeError: dictionary changed size during iteration  closed=a    streams_left=2
mutation  NEW: no exception                                            closed=abc  streams_left=0

raising   OLD: RuntimeError: transport already gone                    closed=-    streams_left=3
raising   NEW: no exception                                            closed=bc   streams_left=0
```

**Why the extra commit.** With the dict cleared up front, a `close-stream` delivered during teardown looks up a key that is already gone:

```
old_handler  -> KeyError: 'b'
new_handler  -> no exception
```

`_handle_read_stream` calls `_handle_close_stream` from its own `except`, so before this the miss faulted twice — the second one from inside the exception handler. The handler is now `if stream := self.streams.pop(stream_id, None): await stream.response.aclose()`.

**Knowingly left alone:** `_handle_read_stream`'s own `self.streams[stream_id]` lookup can still raise for a stream teardown has taken. The double-fault is fixed, but the first `KeyError` still propagates to the relay. Deciding what to answer the relay with instead is a protocol call rather than mechanical hardening, so I did not make it here.

**Finding 3, checked before rewriting.** I read `socketio/async_client.py` 5.16.3 rather than take the correction on trust, and it is exactly as you describe:

- `will_reconnect = self.reconnection and self.eio.state == 'connected'` (line 594) is computed **before** and independently of the `if self.connected:` block (line 595) that holds the cleanup.
- `if will_reconnect and not self._reconnect_task:` (line 605) sits **outside** that block, so the task starts even when the cleanup was skipped.
- `self.connected = True` is the final statement of `connect()` (line 183), which is why the flag is still `False` when the handler runs.

So the reconnect task does start in the wedged state. The description now says that, and the paragraph no longer claims the handler never fires. I also named the two drift items: `_handle_connect_error` closing over the local `relay` instead of `self.relay` (so a discarded client disconnects itself), and `_handle_read_stream` moving to `anext()`. The "a branch I left alone" paragraph is deleted, since the branch is gone.

**Finding 4, checked for vacuity.** Flipping the expected value to `'connected'` makes it fail, so it is asserting something real:

```
tests/test_air.py:5: AssertionError
FAILED tests/test_air.py::test_engineio_state_contract
1 failed in 0.03s
```

**Gates.** `mypy ./nicegui` clean (245 files), `pylint ./nicegui` 10.00/10, `pre-commit` all hooks pass, `pytest tests/test_air.py` 1 passed / 0 skipped.

**Second opinion.** The extra commit is not mine — I ran the four commits past Codex (GPT-5.2) as a skeptic before pushing, and it found the `KeyError` consequence, which I then reproduced above. It read the `connect()` restructure as sound: no await between the guard and `self.connecting = True`, `finally` still clears the flag on the failure paths, and the handlers that close over the old `relay` only use it to disconnect that same object. It also confirmed that swallowing `Exception` in `_close_streams()` does not swallow cancellation, since `CancelledError` is a `BaseException`.

</details>

falkoschindler (approved):
All four applied nicely, thanks! And please keep 6e11b480 — the idempotent handler is the direct consequence of snapshot-and-clear, and the double-fault it removes was real.

Let's merge! 🚀

PR #6214, opened 2026-07-28
Update Quasar to 2.22.0

falkoschindler:
### Motivation

Dependabot alerts 347 and 348 flag Quasar 2.18.5 for [GHSA-3r53-75j5-3g7j](https://github.com/quasarframework/quasar/security/advisories/GHSA-3r53-75j5-3g7j) — prototype pollution in the public `extend()` utility. During a deep merge, `extend(true, target, source)` recursively copies attacker-controlled keys without blocking `__proto__`, so properties can land on `Object.prototype`. Fixed in Quasar 2.22.0.

**NiceGUI has no practically reachable path to this.** The analysis, for the record:

- NiceGUI never calls `Quasar.extend`. The only `Quasar.*` call in the entire frontend is `Quasar.Notify.create` (`nicegui/static/nicegui.js:547`).
- The shipped bundle has exactly two internal `extend(true, …)` sinks: **QEditor** (`props.definitions` / `props.fonts`) and the **Meta plugin** (`updateClientMeta`). NiceGUI does not use the Meta plugin — `<head>` is assembled server-side in Jinja, and there is no `useMeta` anywhere in the repo.
- `extend` is exported on the `Quasar` global, but calling it requires already executing JavaScript on the page — at which point prototype pollution is not the remaining problem.
- Props are authored exclusively server-side. The client→server socket channel only accepts `event`, `javascript_response`, `ack`, `log` and `handshake` (`nicegui/nicegui.py:196-253`); there is no prop-write path from the browser.

The residual exposure is an app that forwards untrusted data into the `definitions` or `fonts` props of a `ui.editor` — which the Python signature does not even expose, so it would have to be set deliberately via `.props()`. That matches the advisory's CVSS vector (`AC:H`, impact `L`/`L`/`L`, medium).

So this is dependency hygiene rather than an urgent fix — but the alerts are legitimate, the bump is cheap, and staying four minor versions behind on the core UI framework only makes the eventual jump larger.

### Implementation

Pin `quasar` to 2.22.0, re-run `npm run build` (`extract_core_libraries.py`) to refresh the vendored assets, and regenerate the two derived files:

- `nicegui/language.py` via `fetch_languages.py`
- `DEPENDENCIES.md` via `summarize_dependencies.py`

**Why 2.22.0 and not the latest 2.23.2.** 2.22.0 is exactly the version the advisory names, and stopping there keeps the blast radius small. Going to 2.23.x would additionally:

- make the RTL FAB workaround in `nicegui/templates/index.html:27` redundant — it mirrors quasarframework/quasar#18413 and #18419, which landed in 2.23.0, and its own comment schedules its removal for the Quasar bump in NiceGUI 4.0;
- pull in the reworked RTL CSS emission (quasarframework/quasar#18420, *"emit override-mode RTL so nested dir subtrees don't double-apply"*), whose release note explicitly recommends a changed `postcss-rtlcss` configuration. `extract_core_libraries.py` extracts from exactly that `quasar.rtl.css`.

Both belong in the deliberate 4.0 bump, not in a security-driven update.

**Why four minor versions is less than it looks.** 2.19.0 and 2.20.0 raised the minor version for internal reasons, not for new functionality — Quasar says so outright for 2.19.0 (*"We're bumping the minor version without new features to signal our important shift to the Oxc+Rolldown ecosystem for our internal build system"*), and 2.20.0's `feat` entries are all modernization and size/perf work (*"Quasar UI is now much smaller in size and runs faster"*), not new API. Across the whole 2.18.5 → 2.22.0 range the only genuinely new public surface is in 2.21.0: `QTable.getCellValue(colName, row)` and the `lb` language pack, plus additive icon-set definitions in 2.19.3. Everything else is fixes, security patches and build-system work.

That is what makes this bump acceptable ahead of the deliberate Quasar upgrade planned for NiceGUI 4.0: we absorb the security fix and four releases' worth of bug fixes without taking on new feature surface that would need documenting, testing or supporting.

**New language pack.** Quasar 2.21.0 adds Luxembourgish, so `nicegui/static/lang/lb.umd.prod.js` is new and `Language` gains `'lb'`. Regenerating also picks up `'sq'` and `'ur-PK'` — those language files have been present in `nicegui/static/lang/` for a while but were never added to the `Language` literal, so this fixes a pre-existing gap in passing.

#### Two user-visible changes

**1. Table pagination label now uses an en dash.** Quasar 2.21.4 (*"fix(ui): table pagination range labels for LTR languages"*, quasarframework/quasar#18378) changed the separator in every language pack:

```diff
-pagination:(e,a,r)=>e+"-"+a+" of "+r     # 2.18.5
+pagination:(e,t,n)=>e+`–`+t+` of `+n     # 2.22.0
```

`ui.table` therefore renders `1–2 of 3` instead of `1-2 of 3`. This is Quasar's label, passed straight through. `tests/test_table.py` asserted on the old text in two places and is updated accordingly (with `# noqa: RUF001`, since ruff flags the en dash as ambiguous). Those were the only two places in the repo that check the string — but anyone asserting on or parsing that label downstream will notice, so it is worth a release-note line.

**2. `-webkit-user-select` is dropped from the vendored CSS.** Since 2.19.0 (cssnano → lightningcss), Quasar's **non-minified** CSS builds no longer carry the `-webkit-` prefix for `user-select`, while the minified ones still do. `extract_core_libraries.py` sources from `quasar.rtl.css` — the non-minified variant — so NiceGUI loses it:

| file | 2.18.5 | 2.22.0 | 2.23.2 |
| --- | --- | --- | --- |
| `quasar.css` | 13 | 1 | 1 |
| `quasar.prod.css` | 13 | 13 | 13 |
| `quasar.rtl.css` — we extract from this one | 13 | 1 | 1 |
| `quasar.rtl.prod.css` | 13 | 13 | 13 |

14 rules are affected: `.q-table th`, `.q-icon` together with all Material icon classes, `.q-checkbox__bg`, `.q-radio__bg`, `.q-toggle__thumb`, `.q-splitter__separator`, `.q-stepper__tab--navigation`, `.q-slide-item__content`, `.q-drawer__opener`, `.q-touch`, `.q-field__native`, `.non-selectable`, plus two `-webkit-text-decoration` rules (`abbr[title]`, `.q-link--focusable:focus-visible`).

Safari still requires the prefix, so those surfaces become text-selectable there. **Confirmed by hand in Safari:** table headers are now selectable, where before they only showed the text cursor on hover. Our Selenium suite runs on Chrome and cannot see this.

**This PR ships that regression, and it is probably fixable on our side.** Since the minified variant keeps the prefix, pointing `_extract_quasar_css` at `quasar.rtl.prod.css` may be all it takes — no mirroring of Quasar internals required. It is not done here because the two files differ by more than whitespace: `-webkit-` appears 64 times in the minified variant against 36 in the non-minified one, and the selector-block count differs (3,264 vs 3,336 — the minified build merges rules). Whether the important/unimportant split and its assertions survive that switch needs verifying, and it deserves its own change rather than being folded into a security bump.

Upstream, the divergence between the minified and non-minified builds looks like a lightningcss target misconfiguration worth reporting: 2.19.0 explicitly names Safari 16.4 as its baseline, and the two outputs should not differ in prefixing.

The remaining CSS delta is benign: Quasar renamed its scroll-lock class (`.q-body--prevent-scroll` → `.q-document--prevent-scroll`, from the 2.21.1 QDialog Safari fix) and added an RTL rule for `.q-rating`. NiceGUI references neither class — the dialog HACKs in `index.html` use NiceGUI's own `.nicegui-dialog-open` etc.

#### Testing

| gate | result |
| --- | --- |
| `pytest` (full suite) | 998 passed, 3 skipped, 2 xfailed, 0 failed (14:37) |
| `mypy ./nicegui` | Success: no issues found in 245 source files |
| `pre-commit run --all-files` | passed |

### Progress

- [x] This PR does not address a security issue *in NiceGUI itself*. It bumps a third-party dependency for the already-public upstream advisory GHSA-3r53-75j5-3g7j; as analysed above, NiceGUI has no reachable path to the vulnerable function.


falkoschindler:
Correction to the `-webkit-user-select` section — I had the cause wrong and have updated the description.

The prefix is **not** dropped across the board. Quasar's *minified* CSS builds still carry it; only the *non-minified* ones lost it in 2.19.0. `extract_core_libraries.py` happens to source from `quasar.rtl.css`, the non-minified variant, which is why we lose it:

| file | 2.18.5 | 2.22.0 | 2.23.2 |
| --- | --- | --- | --- |
| `quasar.rtl.css` — our source | 13 | 1 | 1 |
| `quasar.rtl.prod.css` | 13 | 13 | 13 |

So this is likely fixable on our side by extracting from the minified variant, rather than the unavoidable Quasar quirk I originally described. Not folded into this PR: the two files differ structurally (3,264 vs 3,336 selector blocks, since the minified build merges rules), so the important/unimportant split and its assertions need validating against that source first.

This PR therefore still ships the regression. The Safari observation itself is unchanged and confirmed.

falkoschindler:
Closing this — we are deferring the Quasar upgrade to 4.0 rather than doing it twice.

The alert that prompted it (347, GHSA-3r53-75j5-3g7j) does not reach NiceGUI: we never call `Quasar.extend`, the only internal deep-merge sinks are QEditor's `definitions`/`fonts` props and the Meta plugin — which NiceGUI does not use — and props are authored exclusively server-side. Meanwhile the bump ships a confirmed Safari regression. Trading a real, user-visible regression for a hypothetical exposure is the wrong way round, especially when `nicegui/templates/index.html` already schedules the Quasar bump for 4.0, where the target will be 2.23.x and the RTL FAB workaround comes out with it.

Everything below stays valid as preparation for that. The three findings worth carrying over:

1. **`-webkit-user-select` is lost through our extraction, not through Quasar.** Quasar's minified CSS builds still carry the prefix; only the non-minified ones lost it in 2.19.0. `extract_core_libraries.py` sources from `quasar.rtl.css`, the non-minified variant, and derives all four of our outputs from it — so `prod_js` does not help either. Pointing `_extract_quasar_css` at `quasar.rtl.prod.css` is the likely fix, but the two files differ structurally (3,264 vs 3,336 selector blocks), so the important/unimportant split and its assertions need validating against minified input first.

2. **The table pagination label changes.** Quasar 2.21.4 switched the separator to an en dash, so `ui.table` renders `1–2 of 3` instead of `1-2 of 3`. Two assertions in `tests/test_table.py` need updating, and it belongs in the release notes.

3. **`nicegui/language.py` needs regenerating.** Quasar 2.21.0 adds the `lb` language pack; running `fetch_languages.py` also picks up `sq` and `ur-PK`, which have been present in `nicegui/static/lang/` for a while without ever reaching the `Language` literal.

evnchn:

**Closing call looks right — but three of the supporting claims don't survive a re-check.** Flagging them only because you said the analysis stays valid as preparation for the 4.0 bump.

<details><summary><b>What re-checking found</b></summary>

Verified against c40f3e3f (nicegui `3.13.0.post24.dev0`, vendored Quasar 2.18.5), Chrome for Testing 141 via Playwright, macOS.

1. **"The only `Quasar.*` call in the entire frontend is `Quasar.Notify.create` (`nicegui/static/nicegui.js:547`)"** — there are more:
   - `nicegui/app/app_config.py:26-27` — `Quasar.lang.set`, `Quasar.Dark.set`
   - `nicegui/elements/fullscreen.js:19,22,29` — `Quasar.AppFullscreen.isActive/request/exit`
   - `nicegui/elements/notification.js:5` — a second `Quasar.Notify.create`

   None is an `extend` sink, so the conclusion holds — but the enumeration is what carries it.

2. **"There is no prop-write path from the browser."** Confirmed false (ran it). A forged `event` emit using the ordinary wire shape — `args` as JSON strings, which `Client.handle_event` feeds to `json.loads` (`nicegui/client.py:398`) — lands an arbitrary nested object in a server-side prop, via `ValueElement` on a plain `ui.input()`:

   ```
   server-side: type=dict value={'__proto__': {'pwn': 'FROM_BROWSER'}, 'attacker': 'controlled'}
   props={'value': {'__proto__': {'pwn': 'FROM_BROWSER'}, 'attacker': 'controlled'},
          'loopback': False, 'error': None, 'for': 'c4', 'type': 'text', '_autocomplete': []}
   ```

   The sound version of the claim is narrower and still sufficient: *no client-writable prop is consumed by an `extend(true, …)` sink.* Worth stating it that way, since that's the invariant which would break if an element ever exposed a client-writable deep-merged prop.

3. **Severity factors are inverted, though the product is right.** Low reachability is the real point. But impact-if-reached isn't the advisory's `L/L/L` here, because NiceGUI props are trusted to carry JS by design (`:`-prefixed props are evaluated in `renderRecursively`) — so pollution landing in props isn't a data-only outcome. Same verdict, sturdier reasoning.

**Also checked, and it supports the PR as written:** the `.props('definitions=…')` *string* route is inert — the prop never reaches the wire (element props stay `{"value":"hello","loopback":false}`). The residual path needs a deliberate `element.props['definitions'] = <dict>`, narrower than "via `.props()`" suggests.

**Refuted on the way, recorded so nobody re-walks it:** I tried to turn pollution-into-props into a reportable issue and dropped it. The `:`-prop evaluation is documented intended behaviour, so an app forwarding untrusted data into props has already left the trust model; and the `??=` normalisation in `replaceUndefinedAttributes` is only reachable via a genuine pollution primitive — 0 writes to any field it normalises across the 1,095 packages of a real lockfile. No advisory warranted.
</details>

PR #6216, opened 2026-07-28
Update echarts-gl to 2.1.0

falkoschindler:
### Motivation

`nicegui/elements/echart` declares `echarts@^6.0.0` but was locked to `echarts-gl@2.0.9`, whose peer dependency is `echarts@^5.1.2`. That mismatch makes `npm install` fail in that directory:

```
npm error ERESOLVE could not resolve
npm error Conflicting peer dependency: echarts@5.6.0
```

Every dependency update in that element had to be run with `--legacy-peer-deps`, which silences *all* peer checks, not just this one.

`echarts-gl@2.1.0` (2026-05-28) widens its peer range to `^5.1.2 || ^6.0.0`. Our manifest's `^2.0.9` already permitted it — only the lockfile predated the release, and `npm update` could not get there because the resolver rejected the tree over the very conflict it would have fixed.

### Implementation

Bump `echarts-gl` to `^2.1.0` and rebuild the element bundle.

**No new functionality.** The release notes for 2.1.0 read, in full: *"Fix compatible issues with Apache ECharts 6"*. Verified against the published packages:

| | 2.0.9 | 2.1.0 |
| --- | --- | --- |
| `dist/echarts-gl.js` | 2,098,288 B | 2,100,227 B (+0.09%) |
| registered chart types | 16 | 16 |
| added / removed types | — | none / none |

The rebuilt bundle still contains all 20 3D and GL types (`scatter3D`, `bar3D`, `globe`, `graphGL`, …).

After the bump, `npm install` runs clean in `nicegui/elements/echart` without `--legacy-peer-deps`.

#### Test coverage

`echarts-gl` is imported lazily (`loadEchartsGL` in `src/index.mjs`, triggered by `enable_3d` or any `*3D` option key), and no existing test touched that path — all 11 tests exercised plain 2D charts only. A version bump of a lazily loaded bundle would have gone completely unverified.

`test_3d_chart` renders a `scatter3D` and asserts that a WebGL context exists, which only happens once the lazily imported bundle initializes.

For what it is worth, the test passes against **both** 2.0.9 and 2.1.0 — our 3D usage was not broken before. The benefit here is the resolvable dependency tree, not a bug fix.

#### Dependency summary

`summarize_dependencies.py` had a fallback that read the first line of a package's `LICENSE` file when `package.json` carried no `license` field. `echarts-gl@2.0.9` was the only one of our 29 dependencies without that field, so the bump makes the fallback dead code — removed here, and `DEPENDENCIES.md` regenerates byte-identically.

Keeping it would have been the worse option anyway: for 15 of the 29 dependencies the first line of `LICENSE` is prose rather than an identifier (`The MIT License (MIT)`, `Dart Sass license:`, `Copyright (c) 2017-2019, The xterm.js authors`), and three start with a blank line. A missing `license` field now fails loudly with a `KeyError` instead of silently emitting a bogus license name.

The `LICENSE_LINKS` key `'BSD 3-Clause'` was the spelling parsed out of those file headers; every other key is an SPDX identifier. It is now `'BSD-3-Clause'` — also so the script keeps working if echarts-gl corrects the metadata mentioned below.

### Note on the reported license

`DEPENDENCIES.md` now lists `echarts-gl` as MIT instead of BSD 3-Clause. **The license did not actually change** — this is an upstream metadata error:

| source | license |
| --- | --- |
| repository, default branch | no `license` field |
| repository, tag `2.1.0` | no `license` field |
| published npm tarball 2.1.0 | `"MIT"` |
| `LICENSE` file (repo and tarball, byte-identical to 2.0.9) | BSD 3-Clause, Baidu Inc. |
| GitHub license detection | BSD-3-Clause |

The `license: "MIT"` field exists only in the published artifact and in no commit. We report what the package declares — the same field npm, GitHub and license scanners read — rather than deviating from the ecosystem to paper over someone else's data error. This is being reported upstream; once corrected, our next regeneration returns to BSD-3-Clause on its own.

#### Testing

| gate | result |
| --- | --- |
| `pytest tests/test_echart.py` | 11 passed |
| `pylint summarize_dependencies.py` | 10.00/10 |
| `pre-commit run --all-files` | passed |

### Progress


falkoschindler:
Reported the license metadata issue upstream: ecomfe/echarts-gl#576

Once they correct the `license` field, our next `summarize_dependencies.py` run returns `echarts-gl` to BSD-3-Clause on its own — the `LICENSE_LINKS` key was renamed to the SPDX spelling `BSD-3-Clause` in this PR precisely so that keeps working. Nothing here is blocked on their response.

evnchn:

**Pushed b82265c9** — dedupes zrender to a single 6.1.0, since the bump to 2.1.0 is what makes it possible. `git revert b82265c9` if you'd rather it were its own PR.

<details><summary><b>Why, what changed, and how it was verified</b></summary>

### Why

2.1.0 widens echarts-gl's range to `^5.1.1 || ^6.0.0`, but the existing lockfile entry for 5.6.1 still satisfied it, so npm had no reason to move. The element bundle therefore shipped **two zrender majors**, with echarts-gl's `animatableMixin` handing a zrender-5 `Animator` to echarts' zrender-6 instance via `this.__zr.animation.addAnimator(...)`.

### The change

```
cd nicegui/elements/echart && npm ci && npm update zrender && npm run build
```

`package.json` is unchanged — the widened range already permits v6. The lockfile diff is 15 lines (net −9): it deletes the nested `node_modules/echarts/node_modules/zrender` block and moves top-level `node_modules/zrender` 5.6.1 → 6.1.0. echarts pins `zrender: "6.1.0"` exactly and echarts-gl accepts `^5.1.1 || ^6.0.0`, so one hoisted 6.1.0 satisfies both and zero nested entries remain. All 7 zrender submodules echarts-gl deep-imports still exist in 6.1.0.

| | before | after |
| --- | --- | --- |
| modules transformed | 887 | 872 |
| echarts-gl chunk | 624,730 B | 600,802 B (−23.9 KB) |
| gzip | 171.97 kB | 163.46 kB |
| zrender in bundle | 5.6.1 (14 modules) + 6.1.0 (96) | 6.1.0 only (96) |

### Verification — A/B, headless Chrome with software WebGL

Same flags NiceGUI CI uses. Control = this PR at ed42fddb; deduped = with the commit.

| | control | deduped |
| --- | --- | --- |
| `pytest tests/test_echart.py` | 10 passed, 1 failed | 10 passed, 1 failed |
| `test_3d_chart` | passed | passed |
| scatter3D / bar3D / line3D / surface / scatterGL, each asserting a live WebGL context **and** no SEVERE console errors | 5/5 | 5/5 |

The single failure is `test_click` (pixel-coordinate clicking) and it fails **identically on the control** — an artifact of my container's rendering, not of this change.

Not covered: dispose-during-animation, hover/roam/tooltip, globe/geo3D, non-Chrome engines, hardware WebGL.

### Reproducibility

A fresh `npm ci` from the new lockfile in a clean `node:22` container rebuilds the new bundle byte-identically.

The same check on your ed42fddb bundle also reproduces byte-for-byte: `index.js`, `index-oW8ulKok.js` and `index-ij0tgg2h.js` are byte-identical, and the three `.js.map` differ only by the trailing newline `end-of-file-fixer` adds. Both echarts-gl tarballs match their lockfile `integrity` hashes, and 2.1.0's npm `src/` is 196/196 byte-identical to GitHub tag 2.1.0 — the only `package.json` delta is the added `"license": "MIT"`, exactly as you describe.

### Second opinion

Codex reviewed the approach and advised **against** an `overrides` entry in `package.json` as policy rather than fix, on the grounds that the widened range already makes the solver legal. Its one caveat: a future lockfile regeneration could in principle re-pick a v5, though echarts' exact pin makes that unlikely. Happy to add the override instead if you'd prefer the invariant encoded.

</details>

falkoschindler:
Thanks for the zrender dedupe, @evnchn! Shipping two zrender majors in one bundle is exactly the kind of thing this update should flush out, so I'm happy to keep b82265c9 in this PR — no need for a revert or a separate PR.

With that settled, I think the PR is complete. Would you mind finishing your review and approving if everything looks good to you?

evnchn (approved):
The PR already sits at "OK if zrender deduplicated" state beforehand, so everything is good. 

Just in case, I double-checked with Claude Code in the cloud sandbox (no memory of my MacBook config) and the review came back positive also. 

---

Unrelated issues: moving away from `screen.wait(1.0)` _might_ help with reducing flakes with the tests, but I didn't bother with it since the tests aren't flaking yet

PR #6218, opened 2026-07-28
Never show a fabricated GitHub star count

### LINKED ISSUE #6211: Website shows "0 GitHub Stars" on some Fly machines: failed API fetch falls back to a literal 0
*Filed by Claude Code on evnchn's behalf.*

### First Check

- [X] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [X] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
# website/github_stars.py on main, in full — the offending code.

@binding.bindable_dataclass
class GitHubStars:
    string: str = app.storage.general.get('github_stars', '0')   # <-- fallback is literally "0"


stars = GitHubStars()


async def _fetch() -> None:
    try:
        async with httpx.AsyncClient() as client:
            response = await client.get('https://api.github.com/repos/zauberzeug/nicegui', timeout=10)
            response.raise_for_status()
            stars.string = f'{response.json()["stargazers_count"] // 1000}k+'
            app.storage.general.update(github_stars=stars.string)
    except Exception:
        logging.log.warning('failed to fetch GitHub star count')   # <-- swallow, value stays "0"


app.timer(3600, _fetch)
```

### Description

**TL;DR:**

nicegui.io renders **"0 GitHub Stars"** — but only from some Fly machines. The Tokyo (`nrt`) machine serves `0`; San Jose (`sjc`) serves `16k+`. Users routed to `nrt` (i.e. much of Asia) see a confident, wrong `0`, and it was still `0` throughout the sampling window below.

Measured **58/58 requests** with perfect region correlation (evidence in Additional Context).

The cause is in `website/github_stars.py`, three compounding properties:

1. **The failure fallback is a wrong number, not an absent one.** When the fetch fails the value stays at its seed — the literal string `'0'`. That renders as a confident, plausible-looking value rather than a hidden or stale one.
2. **The cache that would rescue it is machine-local.** `app.storage.general` resolves to a `FilePersistentDict` at `.nicegui/storage-general.json` unless `NICEGUI_REDIS_URL` is set. Each Fly machine has its own filesystem, so `sjc`'s good value never reaches `nrt`, and a cold or restarted machine has no seed at all.
3. **There is no fast retry after a failure.** `app.timer` defaults to `immediate=True`, so a first attempt does fire at startup and it does retry hourly — the gap is that a *failed* attempt gets no short retry or backoff, it just waits out the full 3600s. So one transient failure costs an hour of `0`, and a machine whose fetches keep failing keeps serving `0` indefinitely.

Both bindings to `stars.string` are affected, so the wrong value shows in two places:

- `website/components/sponsors_section.py:25` — the "GitHub Stars" stat tile
- `website/header.py:151` — the GitHub badge in the nav header

**Expected:** the star count reflects reality (~16k, i.e. `16k+`), or the element is hidden when the real value is unknown.
**Actual:** `0` is displayed to everyone routed to the affected machine, for as long as that machine's fetches keep failing.

**Ask:** three fix directions are listed at the bottom — happy to open a PR for whichever you prefer.

### NiceGUI Version

3.15.0 — as served by the deployed nicegui.io (read from the payload's `dynamic-resource-path`). This is a bug in the deployed website, not in a local install.

### Python Version

n/a — the bug is server-side on the deployed nicegui.io, so the reporter's local Python version is not involved. Reproduced with `curl` only.

### Browser

Other — reproduced with `curl` against the raw HTML payload, so it is browser-independent. Originally noticed in a normal desktop browser session from Hong Kong.

### Operating System

macOS (test client only; the fault is server-side)

### Additional Context

<details>
<summary><b>Empirical evidence — 58/58 perfect correlation by region</b></summary>

Confirmed (ran it). Three separate rounds against production `https://nicegui.io/`, extracting the stat-tile text from the served HTML payload and correlating with the `fly-request-id` region suffix:

| Round | Samples | `sjc` (San Jose) | `nrt` (Tokyo) |
|---|---|---|---|
| 1 | 14 | `16k+` (7/7) | `0` (7/7) |
| 2 | 24 | `16k+` (17/17) | `0` (7/7) |
| 3 (re-verify) | 20 | `16k+` (11/11) | `0` (9/9) |
| **Total** | **58** | **`16k+` (35/35)** | **`0` (23/23)** |

All three rounds were taken between roughly 12:39 and 12:44 HKT on 2026-07-26 (~5 minutes); the `nrt` machine served `0` continuously across that window, and the behaviour was first noticed in ordinary browsing beforehand. No sample ever contradicted the split.

Real count when sampled (2026-07-26, via `api.github.com/repos/zauberzeug/nicegui`): `"stargazers_count": 16064` — enough to render `16k+`. It will drift, so treat the exact figure as a timestamp, not a constant.

Sample raw output showing the correlation:

```
stars=16k+ | fly-request-id: 01KYEBM7ZXPYBXBFWGT9B11AJY-sjc
stars=0    | fly-request-id: 01KYEBM8R8MP0F6S9CAYBT713K-nrt
stars=16k+ | fly-request-id: 01KYEBM9DXTAZ5SCM55C71RWXA-sjc
stars=16k+ | fly-request-id: 01KYEBMAEQTZMDSS5B1MTPRTGE-sjc
stars=0    | fly-request-id: 01KYEBMB7Q784256TEC6QMMJ2W-nrt
stars=0    | fly-request-id: 01KYEBMBNR24B0YTDSFNAARS5H-nrt
stars=16k+ | fly-request-id: 01KYEBMCHAMCCZKG3JV44JX53A-sjc
```

</details>

<details>
<summary><b>Runnable repro (copy-paste)</b></summary>

```bash
for i in $(seq 1 20); do
  curl -s -D h.txt https://nicegui.io/ -o b.html
  reg=$(grep -i '^fly-request-id' h.txt | tr -d '\r' | sed 's/.*-//')
  val=$(grep -oE '"text":"[^"]*","class":\["text-\[1\.5rem\]"\]' b.html \
        | head -1 | sed 's/","class.*//;s/"text":"//')
  echo "$reg   $val"
done | sort | uniq -c
```

Expected output shape:

```
   9 nrt   0
  11 sjc   16k+
```

**What this proves:** the served HTML genuinely differs by Fly machine, and the `0` is server-side — not a browser, cache, or client-side-fetch problem.

**What it does NOT prove:** *why* the `nrt` machine's fetch fails; that needs the machine's logs. The `sjc`/`nrt` ratio also just reflects load-balancing from the test's origin and means nothing on its own.

</details>

<details>
<summary><b>Leading hypothesis for the fetch failure (NOT confirmed)</b></summary>

Hedged — I have no access to the `nrt` machine's logs, so this is inference, not a finding.

The call is **unauthenticated**, and the GitHub REST API rate-limits unauthenticated requests to **60 requests/hour per source IP**. Fly machines egress through shared NAT addresses, so the `nrt` egress IP is plausibly exhausted by unrelated traffic. That would produce a persistent 403/429 on that machine specifically, which matches the observed per-region determinism.

Cheap way to confirm: `fly logs` on the `nrt` machine, grepping for `failed to fetch GitHub star count`. If that warning repeats hourly, this is it.

Alternatives I could not rule out: per-region egress firewall/DNS differences, or the 10s timeout being hit consistently from that region.

</details>

<details>
<summary><b>Hypothesis I checked and refuted</b></summary>

I initially suspected the persisted cache was **dead on arrival** — the reasoning being that `string: str = app.storage.general.get('github_stars', '0')` is a dataclass field default evaluated at *module import time*, which would run before the persistence layer loaded from disk, making the seed always `'0'` and the `app.storage.general.update(...)` write-only.

**That is wrong.** `nicegui/nicegui.py:51` calls `core.app.storage.general.initialize_sync()` at package import, which happens before `website/github_stars.py` is imported. The general storage is already populated, so the cache seed works as intended.

Recording it so the next reader does not re-walk the dead end: the cache is functional, it just does not cross machine boundaries.

</details>

<details>
<summary><b>Three fix directions</b></summary>

Not mutually exclusive; the first is the minimal fix that removes the wrong number.

1. **Never render `0`.** Either hide the stat tile and nav badge until a real value is known, or seed with a hardcoded floor (e.g. `'16k+'`, bumped occasionally) so a failed fetch degrades to *slightly stale* rather than *wrong*. Smallest diff, and it fixes the user-visible symptom regardless of why the fetch fails.
2. **Retry properly.** Add a short retry interval on the failure path (e.g. retry in ~60s, backing off to the hourly cadence once successful) so a single transient failure does not cost an hour.
3. **Fix the fetch itself.** Authenticate the API call with a token (5000 req/hour instead of 60), and/or share the cache across machines via `NICEGUI_REDIS_URL` so one machine's success covers all of them. Alternatively bake the count in at build/deploy time and drop the runtime fetch.

</details>

evnchn:
### Motivation

nicegui.io confidently displays "0 GitHub Stars" whenever the star count cannot be fetched. The count is seeded from `app.storage.general` with a `'0'` fallback and is only ever overwritten on a *successful* fetch, so a machine that has never had a successful fetch shows a plausible-looking number that is simply wrong.

Fixes #6211 on the presentation side. The fetch failure itself (a 403 from GitHub's 60/hour unauthenticated per-IP rate limit, permanently exhausted on the Fly machines' shared egress IPs) is a deployment-side problem and stays tracked in #6211.

### Implementation

Two lines, both about never rendering a number we do not have:

- Seed `GitHubStars.string` with `''` instead of `'0'`, so "unknown" is representable and distinct from a real count.
- Bind the visibility of the star tile to `bool(stars.string)`, so an unknown count hides the tile instead of rendering an empty or fabricated one.

Deliberately *not* included: a retry back-off for the failing fetch. Against a persistently exhausted per-IP budget a 403 stays a 403 no matter how often we ask, so retrying faster only adds requests without adding stars. Tests for such back-off internals were dropped along with it.

### Progress

### DIFF
--- a/website/components/sponsors_section.py
+++ b/website/components/sponsors_section.py
@@ -20,7 +20,7 @@ def create() -> None:
             .classes(f'self-center text-center {d.TEXT_SECONDARY}')
 
         with ui.row(align_items='center').classes('gap-12 justify-center my-10 w-full'):
-            with ui.column(align_items='center').classes('gap-0'):
+            with ui.column(align_items='center').classes('gap-0').bind_visibility_from(stars, 'string', bool):
                 phosphor_icon('ph-star').classes(f'{d.TEXT_32PX} {d.TEXT_ACCENT}')
                 ui.label().classes(d.TEXT_24PX).bind_text_from(stars, 'string')
                 ui.label('GitHub Stars').classes(f'{d.TEXT_13PX} {d.TEXT_SECONDARY}')
--- a/website/github_stars.py
+++ b/website/github_stars.py
@@ -5,7 +5,7 @@ from nicegui import app, binding, logging
 
 @binding.bindable_dataclass
 class GitHubStars:
-    string: str = app.storage.general.get('github_stars', '0')
+    string: str = app.storage.general.get('github_stars', '')  # empty means "unknown", so we hide the count
 
 
 stars = GitHubStars()


falkoschindler:
Thanks for the PR — the core idea is exactly right. Meanwhile we confirmed the root cause on the Fly machines themselves: the fetch fails with a 403 because the unauthenticated per-IP rate limit (60/hour) on the shared egress IPs is permanently exhausted by unrelated tenants. Full findings and the deployment-side fix are in #6211.

That finding changes what's worth keeping here, so I'd suggest trimming the PR to its first change only:

- **Keep:** seed `''` instead of `'0'`, and bind the tile's visibility so an unknown count renders nothing instead of a wrong number. That part stays valuable as a safety net no matter what happens on the deployment side.
- **Drop the retry back-off:** against a persistently exhausted IP budget, a 403 stays a 403 no matter how often we ask — and once the tile hides instead of lying, waiting out the hourly cadence costs almost nothing.
- **Drop the test file:** most of its 90 lines (the importlib loading, the monkeypatched `background_tasks`, the interval-ladder assertions) exist to pin the back-off behavior. Without the back-off there is little left to test, and we generally avoid white-box tests of implementation details — especially for website code, which is deliberately untested today.

That leaves a ~5-line diff that fixes the user-visible lie, while #6211 tracks fixing the fetch itself.

evnchn:
@falkoschindler it became a 2 line diff, check if that's what you want. And be sure to check in the live fly.io though it does work locally. Thanks!

evnchn:
Edited the PR description as well

falkoschindler (approved):

PR #6219, opened 2026-07-28
Prefer behavior tests over tests coupled to implementation details

falkoschindler:
### Motivation

When reviewing recent pull requests, a pattern kept showing up: a small fix, plus a new test that guards it by reaching into the implementation — patching a browser API to observe that an internal method was called, asserting on private attributes, or building fake objects that mirror the machinery under test.

Two problems with that:

1. Such a test inverts its own purpose. It has to be rewritten whenever the implementation changes, so it does not protect against the regression it claims to guard. And it is hard to read for anyone who does not already know the internals, which makes it hard to review.
2. It is not free. We are past a thousand tests, and the browser-based ones dominate the runtime for every contributor, forever.

Our docs did not say anything about this — CONTRIBUTING.md only covered the `User` vs. `Screen` choice, and REVIEW.md said flatly "New features and bug fixes need tests", which nudges towards writing *some* test at any cost. This PR writes the actual preference down, so humans and AI assistants know it up front instead of discovering it in review.

Two recent examples where the test was dropped in the last commit: #6196 and #6189.

### Implementation

- **CONTRIBUTING.md**: the `**Tests**` bullet grows into four points — test observable behavior rather than internals; fixture choice (unchanged); not every change needs a test, and skipping one is fine if it is stated in the PR; and a concrete "you are building scaffolding to observe a mechanism" smell test.
- **AGENTS.md**: one bullet under "What to Avoid", pointing at CONTRIBUTING.md.
- **REVIEW.md**: the `**Tests**` checklist is rephrased so every item names a defect a reviewer flags. The previous list mixed "coverage is missing" with "the test itself is bad" in terse fragments (`Edge cases: …`, `Flakiness, …`) without making clear which was which.

Deliberate trade-off: some changes will now ship without a regression guard. That is not a new situation — we do not test the full visual appearance either, and a single broken CSS selector could take down a page. We cannot protect every line against every possible future mistake, so the honest question is one of proportion: what does a test cost to write, read and run, and what does it actually buy? When a one-line correction would need an elaborate piece of scaffolding to be guarded, the balance tips, and we would rather say that out loud than let it happen silently in review.

### Progress

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,6 +15,6 @@
 - [ ] The PR title is a short phrase starting with a verb like "Add ...", "Fix ...", "Update ...", "Remove ...", etc.
 - [ ] The implementation is complete. (Otherwise, open a [draft PR](https://github.blog/news-insights/product-news/introducing-draft-pull-requests/).)
 - [ ] This PR does not address a security issue. (Security fixes must be coordinated via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) process before opening a PR.)
-- [ ] Pytests have been added/updated or are not necessary. <!--Remove the option that does not apply.-->
+- [ ] Pytests have been added/updated, or the **Implementation** section explains why they are not necessary. <!--Remove the option that does not apply.-->
 - [ ] Documentation has been added/updated or is not necessary. <!--Remove the option that does not apply.-->
 - [ ] No breaking changes to the public API or migration steps are described above. <!--Remove the option that does not apply.-->
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,7 +19,11 @@ Work as a pair programmer, not a silent code generator:
 - **Creating new files** when editing existing ones would suffice
 - **Global mutable state** in library code
 - **Unnecessary dependencies** — check if existing code suffices first
-- **Reflexive regression tests** — not every bug fix earns a test, and a test coupled to implementation details can be worse than none (see [CONTRIBUTING.md](CONTRIBUTING.md#coding-conventions))
+- **Reflexive regression tests** — not every bug fix earns a test, and a test coupled to implementation details can be worse than none.
+  Assert observable behavior, not internals: no private attributes, no patched machinery, no fake objects mirroring the code under test.
+  If you catch yourself building scaffolding to observe an internal mechanism, stop — find the user-visible effect to assert on, or skip the test and give the reason in the pull request.
+  Before writing a test, read a recent one in the same file and copy its shape.
+  (Details in [CONTRIBUTING.md](CONTRIBUTING.md#coding-conventions).)
 
 ## Before Claiming a Task Complete
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,6 +19,7 @@ Work as a pair programmer, not a silent code generator:
 - **Creating new files** when editing existing ones would suffice
 - **Global mutable state** in library code
 - **Unnecessary dependencies** — check if existing code suffices first
+- **Reflexive regression tests** — not every bug fix earns a test, and a test coupled to implementation details can be worse than none (see [CONTRIBUTING.md](CONTRIBUTING.md#coding-conventions))
 
 ## Before Claiming a Task Complete
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -200,7 +200,18 @@ The conventions below cover both general Python style and NiceGUI-specific patte
 
 - **Dataclasses**: Prefer `@dataclass(kw_only=True, slots=True)`
 
-- **Tests**: Include tests for new features and bug fixes. Prefer the `User` fixture (fast, runs in the same async context as NiceGUI, no browser); use the `Screen` fixture only when testing browser/JavaScript interactions.
+- **Tests**
+
+  - Include tests for new features and bug fixes — but test observable behavior, not implementation.
+    A test should read like a small demo of the public API: create elements, interact, assert what a user would see.
+    Avoid tests that reach into internals — private attributes, call counts, monkeypatched browser or library APIs, hand-built fake objects.
+    They break when the implementation changes (so they get rewritten along with it, guarding nothing) and they are hard to read for anyone who doesn't know the internals.
+  - Prefer the `User` fixture (fast, runs in the same async context as NiceGUI, no browser); use the `Screen` fixture only when testing browser/JavaScript interactions.
+  - Not every change needs a test.
+    We have over a thousand tests and the browser-based ones dominate the runtime, so each new test is a permanent cost for every contributor.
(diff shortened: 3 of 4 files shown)


falkoschindler:
@evnchn Putting this on your radar, since it probably touches a number of your open PRs.

Nothing here is aimed at any particular one — this was an unwritten preference on our side, so there was no way for you (or your agent) to know about it. I'd really like your take before it lands: does the wording give you something workable to point an assistant at, or is it too vague to act on?

evnchn:
**TL;DR: +1. This isn't a style preference — it's a default agents arrive with, and writing it down is what overrides it.**

The tests that motivated this PR are mine, and they weren't copied from anything here: our suite was already ~93% free of the pattern, with **zero** merged tests asserting on `._props`. My PRs added them anyway.

That points outward, not inward. Most public Python testing is *application* testing, where mocking is idiomatic and is what every tutorial teaches; library testing is a small minority of it. An agent shows up already holding that shape, and local examples alone don't dislodge it — which is exactly why the rule has to be stated rather than inferred.

<details>
<summary><b>The suite before my PRs — nothing here to copy the pattern from</b></summary>

`origin/main` @ `9ab675e545fa622f4d457ce52aaf6e99df9f96d2`, scope `tests/` only, 131 test files:

| pattern | files | % |
|---|---|---|
| `monkeypatch` | 5 | 3.8% |
| `unittest.mock` / `@patch` | 1 | 0.8% |
| private attribute access (`._foo`) | 4 | 3.1% |
| **assertions on `._props`** | **0** | **0%** |
| union of the above | 9 | 6.9% |

```
git grep -l monkeypatch origin/main -- tests/
  tests/test_element_filter.py  tests/test_helpers.py  tests/test_markdown_response.py
  tests/test_serving_files.py   tests/test_storage.py

git grep -lE "\._[a-z_]+\b" origin/main -- tests/
  tests/test_binding.py  tests/test_favicon.py  tests/test_scene_view.py  tests/test_storage.py

git grep -lE "unittest\.mock|mock\.patch|@patch" origin/main -- tests/
  tests/test_fullscreen.py
```

Measures the merged corpus only — it says nothing about whether a behavior test was achievable in any given case.

</details>

<details>
<summary><b>Nearby libraries don't do it either</b></summary>

If this were house taste you'd expect comparable projects to differ. GitHub code search, files matching each term per repo:

| repo | `monkeypatch` | `unittest.mock` |
|---|---|---|
| **encode/starlette** | **0** | 0 |
| **tiangolo/fastapi** | **0** | 0 |
| psf/requests | 2 | 0 |
| pallets/flask | 8 | 0 |
| pytest-dev/pytest | 54 | 6 |

starlette and fastapi are our own stack, at zero. pytest's 54 is an artifact — it *ships* the `monkeypatch` fixture.

**Caveats:** matching *files* repo-wide (not test-only), via the code-search API, which is approximate and indexes the default branch. `encode/httpx` unmeasured (rate-limited). Five repos is a sample, not a census.

</details>

<sub><i>Posted by Claude Code on evnchn's behalf.</i></sub>

evnchn:
**TL;DR: One change I'd argue decides whether this actually lands on agents — state the rule in `AGENTS.md`, don't link to it.**

The new bullet points at `CONTRIBUTING.md#coding-conventions`. But the failure mode being fixed is *an agent's default overriding a rule it never read* — and an agent that doesn't follow the link is in exactly that case. At minimum the smell-test sentence belongs in `AGENTS.md` itself, with the link adding depth rather than carrying the rule.

Two smaller ones:

- **Point at exemplars, not only prohibitions.** A concrete example to copy travels further than a rule to remember. The ~120 test files already doing this right are the exemplar set — *"before writing a test, read a recent one in the same file"* would do more work than the paragraph of don'ts.
- **"Say so in the PR" has nowhere to live.** `PULL_REQUEST_TEMPLATE.md` has `Pytests have been added/updated or are not necessary` — a checkbox with no room for the reason. Either `CONTRIBUTING.md` says to give it under **Implementation**, or the template gets a line. Otherwise it decays back into a silently-ticked box.

`REVIEW.md` I'd leave as written — naming each item as a defect a reviewer flags is a clear improvement, and "Missing test … with no reason given" pairs correctly with "not every change needs a test".

Happy to open a follow-up PR for these, or you fold them in here — your call.

evnchn:
**TL;DR: Screened my open PRs through 6219 — four tests out, three rewritten. The strongest case for the "not every change needs a test" clause turned out to be one where I could show the behavior test cannot exist.**

#6190 removes a leaked `document` keydown listener on unmount. I wrote the obvious behavior test — unmount the keyboard, type, assert the handler doesn't fire — and it **passed with the fix reverted**. The orphaned listener does fire, but the element is already deleted server-side, so no handler runs and the browser console stays empty. There is no outside symptom to assert on, which is why the same shape was right to drop in #6196 and #6189. The fix stays; the test doesn't.

The other three converted cleanly, and in every case the right pattern was already a few lines away in the same file — which I think argues for naming exemplars more than anything I could say about it.

<details>
<summary><b>The three conversions, each verified to fail with its fix reverted</b></summary>

| PR | now asserts | failure with the fix reverted |
|---|---|---|
| favicon | what `/favicon.ico` serves, like `test_data_url` above it | `assert 500 == 200` (route raises `binascii.Error`) |
| query | computed CSS on `body`, like the neighbouring query tests | `assert '255, 0, 0' not in 'rgba(255, 0, 0, 1)'` |
| select | the value after typing one in, like `test_id_generator` above it | `Could not find "value = ['a', 100]"` |

Each was run both ways — fix in place, then `git show origin/main:<file> > <file>` and rerun — so none of them is a test I have only ever seen green.

</details>

Four dropped in total; the two dead-prop removals have no observable change at all, and the On Air fakes reimplemented python-socketio's own state semantics, so they'd have gone stale silently. Each PR body will say which, and why.

falkoschindler:
@evnchn This is better feedback than I hoped for — thank you, especially for the #6190 experiment. Writing the behavior test, reverting the fix, and watching it stay green is exactly the honest way to establish "no test" as the right answer, and "verified to fail with the fix reverted" is a better standard than anything the docs said so far. I'm stealing it: CONTRIBUTING.md now says a regression test must fail when the fix is reverted.

I've folded all three of your points into this PR (with co-authorship):

- AGENTS.md now carries the rule itself — assert behavior, the scaffolding smell test, "read a recent test in the same file" — with CONTRIBUTING.md as depth, not as the location of the rule.
- CONTRIBUTING.md points at exemplars: read a neighbouring test first, copy its shape.
- The PR template checkbox now asks for the reason in the **Implementation** section instead of being silently tickable.

Your before/after data (zero merged `._props` assertions, starlette and fastapi at zero mocking) is the strongest version of the argument I've seen — it stays here in the thread as the evidence, the docs just state the rule.

And thanks for screening your open PRs unprompted. Four out, three rewritten against the actual behavior, each one verified both ways — that's the standard working as intended before it even landed.

evnchn (approved):
Then this is as good as we can make it in the time being. 

Issues will surface itself with the next PRs and we'll watch out and remediate as we go. It's fine.

PR #6221, opened 2026-07-30
Fix popups not appearing in Firefox 153+

evnchn:
*Drafted by Claude Code on evnchn's behalf.*

### Motivation

Every Quasar popup containing a flex row — which is every `ui.input` inside a `ui.popup` or `ui.menu` — silently fails to appear on **Firefox 153 and later**. The popup mounts in the DOM and unmounts correctly; it is simply never painted or positioned.

Reported in https://github.com/zauberzeug/nicegui/discussions/6178. The root cause is Quasar's, and the fix is **merged upstream as quasarframework/quasar#18468**, resolving quasarframework/quasar#16167 (open since 2023). But NiceGUI pins `quasar 2.18.5` and bumping means jumping to 2.23.4+, which is a 4.0 change — hence this hot patch in the meantime.

<details>
<summary><b>Why Firefox 153 specifically</b></summary>

Quasar hides every popup with `.q-position-engine { visibility: collapse }` and clears it only once `setPosition()` has measured and placed the element. `visibility` inherits, so the collapse reaches the popup's contents — and NiceGUI's input renders as `label.q-field.row`, where Quasar's `row` is `display: flex`.

Gecko is the only engine that implements `visibility: collapse` on **flex items** ([bug 783470](https://bugzilla.mozilla.org/show_bug.cgi?id=783470)); Chromium and WebKit treat it as `hidden`, which is why only Firefox is affected. A collapsed flex item is zeroed along the container's main axis. In Firefox 153 that zero began propagating to the flex container's intrinsic width, so the shrink-to-fit `position: fixed` menu computes to width 0 — and `setPosition()` refuses to position an element whose `offsetWidth` is 0, retrying six times over ~50 ms before giving up permanently. The `visibility: visible` and the inline `top`/`left` are written only *after* that guard, so they are never written at all. The collapse causes the zero width, and the zero width prevents the collapse from ever being cleared.

</details>

### Implementation

Four lines in `nicegui/templates/index.html`, mirroring the upstream one-liner:

```css
/* DEPRECATED: mirror of quasarframework/quasar#18468, remove with the Quasar bump for NiceGUI 4.0 */
.q-position-engine {
  visibility: hidden;
}
```

`hidden` hides identically in every engine but has no layout semantics in Gecko, so the popup keeps measurable geometry and the position engine's guard passes.

> [!NOTE]
> **This is not in `nicegui.css`, which is where @falkoschindler asked for it — flagging the deviation rather than burying it.**
>
> It sits in the `@layer quasar` block in `index.html`, directly below the existing `[dir="rtl"] .q-fab__actions--*` mirror. Three reasons:
>
> 1. **It is the established convention for exactly this.** That FAB block is the repo's only precedent for "mirror an upstream Quasar fix pending the bump", and it lives in this file, in this layer, with this comment shape. `nicegui.css` has no `DEPRECATED` marker of any kind, and no file under `nicegui/` outside this template mentions `quasarframework` at all.
> 2. **It keeps both Quasar mirrors adjacent**, so the bump-time `DEPRECATED` grep lands them together.
> 3. **It patches a Quasar bug at Quasar's own cascade layer** rather than claiming `layer(nicegui)` priority for something that is not a NiceGUI style — and it stays overridable by anyone who wants the old behaviour back.
>
> Both placements are *effective* — this is a convention call, not a correctness one. Say the word and I'll move it to `nicegui.css`.

<details>
<summary><b>Cascade analysis — why it wins from inside the Quasar layer</b></summary>

Layer order is declared at the top of the template:

```css
@layer theme, base, quasar, nicegui, components, utilities, overrides, quasar_importants;
```

Quasar's rule comes from `quasar.unimportant.css` → `layer(quasar)`; mine is in an inline `@layer quasar { }` block in the same `<style>` element, *after* that `@import`. Per CSS Cascade 5, an imported sheet is substituted at its `@import` point for source-order purposes, and same-origin same-layer rules are resolved by specificity and then order of appearance. Both selectors are exactly `.q-position-engine`, both declarations are normal importance, both in `layer(quasar)` — so the later one wins on source order.

`quasar.important.css` (`layer(quasar_importants)`, declared last) also carries `.q-position-engine` rules, but they set `margin-*` only, so there is no interaction.

Placing it in `nicegui.css` would also work, and slightly more bluntly: `layer(nicegui)` is later than `layer(quasar)`, so it would win on layer before specificity is even considered. Neither placement is beaten by user CSS added via `ui.add_css`, which lands unlayered and outranks every named layer.

</details>

<details>
<summary><b>Verification — real Firefox builds, both versions, both directions</b></summary>

Real Mozilla builds (not Playwright's), driven by Selenium against the `ui.popup` documentation demo verbatim, served from this branch at 1280×900. Measured on `.q-menu` after clicking the anchor:

| build | Firefox 152.0.6 | Firefox 153.0.1 |
|---|---|---|
| **without this patch** | `visible`, 170×56, positioned | **`collapse`, 0×56, never positioned** |
| **with this patch** | `visible`, 170×56, positioned | `visible`, 170×56, positioned |

Raw output:

```
WITHOUT patch (FF152)  | FF 152.0.6   | visibility=visible   | 170x56 | positioned=True | inputs=1
WITHOUT patch (FF153)  | FF 153.0.1   | visibility=collapse  | 0x56   | positioned=False | inputs=1
WITH patch    (FF152)  | FF 152.0.6   | visibility=visible   | 170x56 | positioned=True | inputs=1
WITH patch    (FF153)  | FF 153.0.1   | visibility=visible   | 170x56 | positioned=True | inputs=1
```

Two things that matter about this shape:

- **The control fails.** The patch was reverted and re-measured on the same browser, so the fix is demonstrably load-bearing rather than only ever observed green.
- **It restores rather than changes.** The FF152 row is identical with and without, and the fixed FF153 geometry equals the FF152 geometry exactly — so this puts Firefox back to pre-regression behaviour rather than introducing new behaviour.

**Popup-family suite**, which runs on Chromium and is therefore also the no-op evidence for unaffected engines: `test_popup.py`, `test_menu.py`, `test_context_menu.py`, `test_tooltip.py`, `test_select.py`, `test_dialog.py` → **53 passed, 0 skipped** in 100 s.

**Local gates:** `pre-commit` on the changed file passes, `mypy ./nicegui` clean (245 files), `pylint ./nicegui` 10.00/10.

**Honest caveat:** Prettier cannot parse this Jinja template — it errors identically on the unmodified file, so it is neither a gate here nor something this change regresses. The added block's indentation matches the adjacent FAB block character for character.

</details>

<details>
<summary><b>Conventions this follows</b></summary>

| Guideline | Where it comes from | How this PR follows it |
|---|---|---|
| Mirror-an-upstream-Quasar-fix patches go in the Quasar layer in `index.html`, commented `DEPRECATED: mirror of <upstream ref>, remove with the Quasar bump for NiceGUI 4.0` | the FAB-RTL block, the repo's only precedent of this kind | Same file, same layer, adjacent placement, comment wording differing only in the PR number |
| `DEPRECATED:` markers name the removal condition | 26 of ~34 existing markers read `remove in NiceGUI 4.0` | `remove with the Quasar bump for NiceGUI 4.0` — names both the trigger and the release |
| PR bodies use the template's **Motivation / Implementation / Progress** | `AGENTS.md` → *Creating Pull Requests* | This body |
| Run the project's tests and linters, and review your own diff for scope creep | `AGENTS.md` → *Before Claiming a Task Complete* | Gates above; diff is +4 lines in one file |
| Research before guessing; search the codebase for similar patterns | `AGENTS.md` → *Pair Programming* | That precedent search is what produced the placement decision |
| Discuss before deciding when strategy is unclear | `AGENTS.md` → *Pair Programming* | The placement deviation is raised in the note above rather than settled silently |
| Staged on a fork first, so upstream sees one clean commit | keeping maintainer review to one pass | evnchn/nicegui#279, reviewed and approved before this was opened |

</details>

<details>
<summary><b>Second opinion (Codex / GPT-5.2)</b></summary>

Reviewed the commit as a skeptic before this was opened, and asked specifically to challenge the placement. **No MUST-FIX.** It independently derived the same cascade conclusion with the spec citation (CSS Cascade 5: imported sheets substituted at the `@import` point; same-layer rules resolved by specificity then order of appearance) and confirmed `quasar.important.css` does not interact, its `.q-position-engine` rules being `margin-*` only.

On breakage: no vector found — `visibility: hidden` still keeps the popup unpainted before positioning, so no pre-position flash; popups meant to stay hidden still do, since Quasar controls the class; tooltips, menus, popup proxies and QSelect all benefit from the same measurement fix; dialogs are not implicated.

Its one recommendation was to state the deviation from `nicegui.css` explicitly rather than let a reviewer reverse-engineer the rationale — which is the note in the Implementation section. Verdict: 8/10, placement defensible enough to put in front of the maintainer.

</details>

### Progress

- [x] Pytests are not necessary. (The regression is Firefox-only; the suite runs on Chromium, where it cannot reproduce. The existing popup-family tests cover the no-op direction.)

### DIFF
--- a/nicegui/templates/index.html
+++ b/nicegui/templates/index.html
@@ -47,6 +47,10 @@
         [dir="rtl"] .q-fab__actions--left:not(.q-fab__actions--opened) {
           transform: scale(0.4) translateX(62px);
         }
+        /* DEPRECATED: mirror of quasarframework/quasar#18468, remove with the Quasar bump for NiceGUI 4.0 */
+        .q-position-engine {
+          visibility: hidden;
+        }
       }
       @layer overrides {
         {{ headwind_css | safe }}


falkoschindler (approved):
👍🏻

PR #6228, opened 2026-08-03
Fix timer parent-slot error after pruned HTTP-only clients

### LINKED ISSUE #6226: Repeating ui.timer raises "The parent slot of Timer(id=N) has been deleted" when its client is deleted while waiting in _can_start() — the _should_stop() guard is unreachable
### First Check

### Example Code

```python
from nicegui import ui


@ui.page('/')
def page() -> None:
    ui.timer(60, lambda: None)


ui.run()
```

Then request the page over plain HTTP without ever opening the websocket (a bot, a prefetch, a user bouncing before the handshake, an HTTP-only load probe):

```bash
curl -s http://localhost:8080/ > /dev/null
```

After `Client.prune_instances` deletes the never-connected client (~60 s), the timer task wakes and — depending on a garbage-collection race — dies with:

```
  File "nicegui/background_tasks.py", line 152, in _handle_exceptions
    task.result()
  File "nicegui/timer.py", line 90, in _run_in_loop
    with self._get_context():
  File "nicegui/elements/timer.py", line 15, in _get_context
    return self.parent_slot or nullcontext()
  File "nicegui/element.py", line 150, in parent_slot
RuntimeError: The parent slot of Timer(id=2) has been deleted.
```

Because it races the cycle collector it is intermittent per occurrence — but at fleet scale it is loud: on our staging deployment this single raise is **~49% of all captured Sentry events** (~3,400 events across two message wordings, 3.14.x: "The parent slot of the element has been deleted.").

### Description

The sequence:

1. A repeating `ui.timer` is created during page build. `_run_in_loop` suspends in `_can_start()` on `await self.client.connected()` while the websocket handshake is pending.
2. The client never connects and is deleted (`prune_instances`, or any other `Client.delete()` while the timer still waits).
3. `Client.delete()` runs `remove_all_elements()` (which cancels the timer via `_handle_delete`) and then `self._connected.set()  # for terminating connected() waits` — which wakes the parked waiter into a **success** return. `_can_start()` answers `True` for a deleted client.
4. `_run_in_loop` proceeds to `with self._get_context():`. If the cycle collector has already collected the slot tree, the `parent_slot` weakref dereferences to `None` and `element.py` raises — **one line before** `while not self._should_stop():`, the guard built for exactly this case (`self.is_deleted` is already `True` at this point and would exit cleanly).

So the timer is already properly canceled and deleted; it just touches the dead slot before it is allowed to notice. No application frame appears in the stack, and no user ever sees anything — but every occurrence goes through `app.handle_exception` at ERROR level, which is costly for anyone forwarding logs to a paid error tracker.

**Suggested fix:** check `self._should_stop()` before entering `self._get_context()` in `_run_in_loop` (and equivalently `self.active and not self._should_stop()` before the context in `_run_once`). At the failure point `_deleted` is `True`, so the hoisted check exits cleanly, and `Timer._cleanup` is already guarded by `if not self._deleted:` so the finally-path stays safe.

A complementary (or alternative) fix would be for `_can_start()` to return `False` when the wake came from `Client.delete()` rather than a real connection — e.g. re-checking `self.client.id in Client.instances` / `self.is_deleted` after `connected()` returns — since `connected()` currently reports success on a deleted client.

### NiceGUI Version

3.15.0 (same raise with the older message wording on 3.14.x)

### Python Version

3.14.6

### Browser

Not applicable (server-side; triggered by HTTP-only page loads)

### Operating System

Linux (Docker)

### Additional Context

The two suspension points and the GC dependency make this hard to reproduce deterministically in a unit test — the raise only happens when the cycle collector runs between the `connected()` wake-up and the timer task's resumption — but the code path above is fully deterministic once the weakref is dead.

AJ-ing:
## Summary
- Check `_should_stop()` before entering `_get_context()` in `_run_in_loop` / `_run_once`, so timers woken by `Client.delete()` / `prune_instances` exit cleanly instead of touching a dead `parent_slot`.
- Harden element `Timer._can_start()` to return `False` when the wake came from a deleted client.
- Add regression test for HTTP-only load + prune (#6226).

Fixes #6226.

## Test plan
- [x] `uv run pytest tests/test_timer.py::test_no_parent_slot_error_when_pruned_before_connect tests/test_timer.py::test_no_leak_when_client_deleted tests/test_timer.py::test_timer_on_deleted_container tests/test_timer.py::test_once_timer_task_cancelled_on_client_delete -q`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/nicegui/elements/timer.py
+++ b/nicegui/elements/timer.py
@@ -26,6 +26,9 @@ class Timer(BaseTimer, Element, component='timer.js'):
         """
         try:
             await self.client.connected()
+            # Client.delete() wakes connected() waiters after removing elements (#6226).
+            if self.is_deleted or self.client.id not in Client.instances:
+                return False
             return True
         except ClientConnectionTimeout:
             self.cancel()
--- a/nicegui/timer.py
+++ b/nicegui/timer.py
@@ -73,6 +73,10 @@ class Timer:
         try:
             if not await self._can_start():
                 return
+            # Check before entering context: Client.delete() can wake connected()
+            # waiters after the element/slot tree is already gone (#6226).
+            if self._should_stop():
+                return
             with self._get_context():
                 await asyncio.sleep(self.interval)
                 if self.active and not self._should_stop():
@@ -86,6 +90,10 @@ class Timer:
             if not self._immediate:
                 await asyncio.sleep(self.interval)
             if not await self._can_start():
+                return
+            # Check before entering context: Client.delete() can wake connected()
+            # waiters after the element/slot tree is already gone (#6226).
+            if self._should_stop():
                 return
             with self._get_context():
                 while not self._should_stop():
--- a/tests/test_timer.py
+++ b/tests/test_timer.py
@@ -275,3 +275,23 @@ def test_no_leak_when_client_deleted(screen: Screen):
     Client.prune_instances(client_age_threshold=0)
     screen.wait(1)
     assert not any(isinstance(obj, ui.timer) for obj in gc.get_objects())
+
+
+def test_no_parent_slot_error_when_pruned_before_connect(screen: Screen):
+    """HTTP-only load + prune must not raise parent-slot RuntimeError (#6226)."""
+    exceptions: list[Exception] = []
+    app.on_exception(exceptions.append)
+
+    @ui.page('/')
+    def page():
+        ui.timer(60, lambda: None)
+
+    screen.start_server()
+    httpx.get(screen.url)
+    screen.wait(0.2)
+    Client.prune_instances(client_age_threshold=0)
+    gc.collect()
+    screen.wait(0.5)
+
+    parent_slot_errors = [e for e in exceptions if 'parent slot' in str(e).lower()]
+    assert not parent_slot_errors, f'unexpected parent-slot errors: {parent_slot_errors}'


falkoschindler:
Thanks a lot for digging into this, @AJ-ing — you were the first to pin down the actual mechanism (`Client.delete()` waking `connected()` waiters as a success), and your `_can_start()` check is exactly the right idea. We decided to go with #6229, which builds on the same insight, for three measurable reasons (details in its description):

- It also closes the `immediate=False` window: with your patch, a client deleted during the initial `asyncio.sleep(interval)` still raises `The client this element belongs to has been deleted.` at the top of `_can_start()`, before your guard runs.
- The hoisted `_should_stop()` checks in the base class turn out to be unreachable in practice (there is no suspension point between `_can_start()` returning and `_get_context()`), and the base `Timer` can't hit the hazard at all — so the whole fix fits into the client-aware subclass.
- The regression test unfortunately passes on unfixed code (the parked `connected()` frame keeps the client strongly referenced, so the slot weakref only dies after the raise would have happened) — a trap the author of #6229 ran into with their own first draft as well.

Your analysis clearly sharpened the final fix, and the credit in the release notes goes to you and @benvc along with @evnchn. Closing in favor of #6229 — thanks again, this kind of independent convergence on the same root cause is what makes these bugs tractable.

PR #6229, opened 2026-08-03
Fix timer error when its client is deleted while it waits for the connection

### LINKED ISSUE #6226: Repeating ui.timer raises "The parent slot of Timer(id=N) has been deleted" when its client is deleted while waiting in _can_start() — the _should_stop() guard is unreachable

evnchn:

### Motivation

A repeating `ui.timer` created during page build parks in `_can_start()` on `await self.client.connected()`. If the client never opens the websocket — a bot, a prefetch, an HTTP-only probe, a user bouncing before the handshake — `Client.prune_instances()` deletes it, and `Client.delete()` ends with `self._connected.set()  # for terminating connected() waits`. The parked waiter wakes into a **success** return, so the timer walks into `with self._get_context():` and dereferences a dead weakref — one line before the `_should_stop()` guard written for exactly this case.

Nobody sees anything in the browser, but every occurrence goes through `app.handle_exception` at ERROR level. @benvc reported it in #6226 with a full causal chain, an MRE and two candidate fixes; on their staging deployment it is ~49% of all captured Sentry events.

This is an alternative to #6228, which addresses the same issue — deliberately **not** marked as auto-closing, so either can land. A measured comparison is folded into the Implementation section below.

### Implementation

`Timer._can_start()` now answers `False` when the timer should not run, instead of `True` whenever `connected()` returned:

```python
if self._should_stop():
    return False
try:
    await self.client.connected()
    return not self._should_stop()
```

Neither guard *decides* anything new: both report the same `_should_stop()` predicate the run loop already consults. The change is only that the existing answer becomes **reachable before the timer touches a dead object** — previously the first thing to consult it was `while not self._should_stop():`, one line *after* the raise.

```mermaid
flowchart TD
    A["ui.timer(...) created while the page is built"] --> B{"immediate?"}
    B -- "no" --> C["await asyncio.sleep(interval)"]
    B -- "yes" --> G1
    C --> G1

    G1["<b>guard 1 (new)</b><br/>if self._should_stop(): return False"]
    G1 -- "should stop" --> R
    G1 -- "keep going" --> D["await self.client.connected()<br/><i>parks here until the websocket arrives</i>"]
    G1 -. "<b>hazard A</b> — without this guard: the client was deleted<br/>and collected during the sleep above" .-> X1["<b>RuntimeError</b>: The client this element<br/>belongs to has been deleted.<br/><i>element.py:138, reached from _can_start</i>"]

    D --> G2["<b>guard 2 (new)</b><br/>return not self._should_stop()"]
    G2 -- "should stop" --> R["return &rarr; _cleanup()<br/><i>no dead object is ever touched</i>"]
    G2 -- "keep going" --> E["with self._get_context()"]
    G2 -. "<b>hazard B</b> — without this guard: Client.delete() woke<br/>connected() as a SUCCESS, slot already collected" .-> X2["<b>RuntimeError</b>: The parent slot of Timer(id=N)<br/>has been deleted.<br/><i>element.py:150, the reported crash</i>"]

    E --> F["while not self._should_stop():<br/><i>the pre-existing guard &mdash; reached one line too late</i>"]

    style G1 fill:#1b3a24,stroke:#3fb950,stroke-width:2px,color:#d7f5e0
    style G2 fill:#1b3a24,stroke:#3fb950,stroke-width:2px,color:#d7f5e0
    style X1 fill:#4a1d1d,stroke:#f85149,stroke-width:2px,color:#ffd7d5
    style X2 fill:#4a1d1d,stroke:#f85149,stroke-width:2px,color:#ffd7d5
    style R fill:#1b3a24,stroke:#3fb950,color:#d7f5e0
    style F fill:#3a331b,stroke:#d29922,color:#f5ecd7
```

|  | Timer is suspended here | What is already dead when it wakes | Raise without the guard | The guard | Covered by |
|---|---|---|---|---|---|
| **A** | `await asyncio.sleep(interval)`, *before* `_can_start()` — `immediate=False` only | the `Client` object itself (`Element._client` holds it weakly) | `The client this element belongs to has been deleted.` — `element.py:138` | `if self._should_stop(): return False` | `delayed` |
| **B** | parked in `await client.connected()` | the element is flagged `_deleted` **and** its parent `Slot` has been collected | `The parent slot of Timer(id=N) has been deleted.` — `element.py:150` | `return not self._should_stop()` | `repeating`, `once` |
| — | inside the callback loop | (same) | *already safe before this PR* | pre-existing `while not self._should_stop():` | — |

<details>
<summary><b>Why <code>_can_start()</code> rather than hoisting <code>_should_stop()</code> above the context</b></summary>

The issue proposes hoisting the guard above `with self._get_context():` in `_run_in_loop` / `_run_once`. Both options close the reported window and neither is tighter: there is **no suspension point** between `_can_start()` returning and `_run_in_loop` reaching `_get_context()`, so a check at the end of `_can_start()` is exactly as timely as one hoisted above the context.

Given that, three things decided it:

1. **It fixes the cause, not a symptom.** `_can_start()` reporting success for a deleted client *is* the defect; hoisting a guard leaves that lie in place for the next caller.
2. **One edit covers both loops.** Hoisting needs a change in `_run_in_loop` *and* `_run_once`. This covers both — see the `once` parametrization.
3. **It keeps client-lifecycle logic in the client-aware class.** The base `nicegui/timer.py` has `_can_start() -> True` and `_get_context() -> nullcontext()`, so it cannot hit this hazard at all; a guard there would defend the base class against a subclass-only problem.

</details>

<details>
<summary><b>The second guard — a sibling window found in review</b></summary>

With `immediate=False`, `_run_in_loop` does `await asyncio.sleep(self.interval)` **before** `_can_start()`. Nothing holds the client during that sleep, so if it is deleted and collected there, the very first line of `_can_start()` raises a different error from the same root cause:

```
  File "nicegui/timer.py", line 88, in _run_in_loop
    if not await self._can_start():
  File "nicegui/elements/timer.py", line 28, in _can_start
    await self.client.connected()
  File "nicegui/element.py", line 138, in client
    raise RuntimeError('The client this element belongs to has been deleted.')
```

Reproduced deterministically with `ui.timer(0.3, cb, immediate=False)` + an HTTP-only page load + `Client.prune_instances()` + `gc.collect()` — no cycle-collector race needed. Same MRE shape, same log noise, one line to close, so it is in scope here rather than a follow-up.

`if self._should_stop()` rather than the narrower `if self.is_deleted` because `_should_stop()` short-circuits on `is_deleted` first (so it is no less safe) and additionally covers cancellation, `client.id not in Client.instances`, and `core.app.is_stopping/is_stopped` — a timer that has not started yet should not start during shutdown either.

</details>

<details>
<summary><b>Regression test: what makes it deterministic, and what it substitutes</b></summary>

The issue notes the raise is GC-race-dependent and "hard to reproduce deterministically in a unit test". The test removes the race rather than waiting on it: the timer lives inside a `ui.column()`, and that container is deleted while the client is **still alive**, so the parent-slot weakref is dead by refcount plus one forced collection rather than by luck. The timer is then woken by the real mechanism from the report — `Client.prune_instances(client_age_threshold=0)` → `Client.delete()` → `_connected.set()` — and the page is fetched over plain HTTP via `user.http_client.get('/')`, never opening the websocket.

**What this substitutes, stated plainly:** in production the element is marked deleted by `Client.delete()` → `remove_all_elements()`; in the test it is marked deleted by the container delete a moment earlier. The state at the failure point is identical, and the container-removal path is itself a realistic app pattern.

Against the guidance added in #6219: it asserts observable behavior (no exception reaches `app.on_exception`, which is the user-visible symptom being reported), touches no private attributes, patches nothing, builds no fakes, uses the `User` fixture rather than `Screen`, runs in 1.8 s, and copies the shape of its neighbour `test_no_leak_when_client_deleted`. The one piece of scaffolding is `gc.collect()`, which reaches a *state* rather than observing an internal mechanism — the same call the neighbouring test already makes.

</details>

<details>
<summary><b>Seen to fail: each parametrization, with the fix reverted</b></summary>

Verified against `b6b37c78` with `nicegui/elements/timer.py` restored to its unfixed state. All three fail, each at its own site — so `delayed` really is exercising guard 1 and has not silently degraded into another guard-2 case:

```
repeating   FAIL   elements/timer.py:15 _get_context  →  The parent slot of Timer(id=5) has been deleted.
once        FAIL   elements/timer.py:15 _get_context  →  The parent slot of Timer(id=5) has been deleted.
delayed     FAIL   elements/timer.py:28 _can_start    →  The client this element belongs to has been deleted.
```

The first is character-for-character the traceback in the issue report, modulo element id. With the fix restored: `3 passed`.

Neither added line is dead code:

| `_can_start()` state | `repeating` | `once` | `delayed` |
|---|---|---|---|
| Neither guard (main) | ❌ parent slot | ❌ parent slot | ❌ client |
| Post-wait guard only | ✅ | ✅ | ❌ client |
| Pre-wait guard only | ❌ parent slot | ❌ parent slot | ✅ |
| Both (this PR) | ✅ | ✅ | ✅ |

</details>

<details>
<summary><b>Measured comparison with #6228</b></summary>

@AJ-ing independently reached the same conclusion about `_can_start()`, and their PR does close the reported crash — my `repeating` and `once` cases pass against it. Two differences I measured rather than argued, both reproducible on `b6b37c78`:

**1. The `immediate=False` window stays open.** Their `_can_start()` check runs only *after* `connected()` returns. Applying only their `nicegui/` changes and running this PR's tests:

```
repeating   PASS
once        PASS
delayed     FAIL   elements/timer.py:28 _can_start  →  The client this element belongs to has been deleted.
```

**2. Their regression test passes on unfixed code — 10 runs out of 10.** Appending `test_no_parent_slot_error_when_pruned_before_connect` to an otherwise untouched tree and running it 10×: `passed 10, failed 0`. So it does not currently guard the regression it was written for.

I raise this as a trap rather than a mistake, because I fell into it myself: **my own first draft of this test was almost exactly the same sequence** — `httpx.get` → `prune_instances()` → `gc.collect()` → wait — and it passed against unfixed code too. The reason is subtle. The parked `await client.connected()` coroutine frame still holds a **strong** reference to the `Client`, so the forced collection cannot free the slot tree; the weakref only dies later, when those frames unwind *inside the timer task's own step*, which is after the raise would have happened. Deleting the timer's parent container before the wake sidesteps the race entirely, and that one change would make their test fail-then-pass as intended.

For completeness, the size difference: this PR is +4/−2 in one source file (2 net new lines); #6228 is +31/−0 across two source files plus a `Screen`-fixture test. #6219 asks for the `User` fixture unless the browser is genuinely needed, and browser tests dominate suite runtime.

None of this is a claim about which PR should land — that is @falkoschindler's call, and @AJ-ing deserves the credit @evnchn already promised in the issue thread regardless of the outcome.

</details>

<details>
<summary><b>Local gates and review</b></summary>

```
$ .venv/bin/pre-commit run --files nicegui/elements/timer.py tests/test_timer.py
ruff check / autopep8 / trailing whitespace / end of files / quoted strings / codespell .... Passed

$ .venv/bin/mypy ./nicegui
Success: no issues found in 245 source files

$ .venv/bin/pylint ./nicegui
Your code has been rated at 10.00/10

$ .venv/bin/python -m pytest tests/test_timer.py -rs -q
17 passed in 27.30s          # -rs shows 0 skipped, so nothing here silently opted out
```

Full local suite on macOS: 9 pre-existing failures (`test_dark_mode` ×8, `test_lazy_imports` ×1) that reproduce identically on an unmodified checkout, and 3 `tests/test_run.py` skips guarded by `needs a platform with a fork default (e.g. Linux)` — none of them cover this change. On Linux CI the earlier revision of this branch ran `1004 passed, 2 xfailed, 0 skipped`.

The second guard came out of an adversarial review pass by Codex (a different model lineage), which also independently confirmed the `remove_all_elements()`-before-`_connected.set()` ordering that makes `_should_stop()` safe to call at the wake. I reproduced the window it flagged before adding the line rather than taking it on faith.

</details>

### Progress

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -465,6 +465,8 @@ class Client:
             self.safe_invoke(t)
         self._waiting_for_disconnect.clear()
         self._deleted_event.set()
+        # NOTE: removing all elements before removing the client from Client.instances ensures
+        # that elements are marked as deleted before their client weakref can die (Timer._should_stop relies on this)
         self.remove_all_elements()
         self.outbox.stop()
         del Client.instances[self.id]
--- a/nicegui/elements/timer.py
+++ b/nicegui/elements/timer.py
@@ -22,11 +22,13 @@ class Timer(BaseTimer, Element, component='timer.js'):
         """Wait for the client connection before the timer callback can be allowed to manipulate the state.
 
         See https://github.com/zauberzeug/nicegui/issues/206 for details.
-        Returns True if the client is connected, False if the client is not connected and the timer should be cancelled.
+        Returns True if the client is connected, False if the timer should not run (anymore).
         """
+        if self._should_stop():
+            return False
         try:
             await self.client.connected()
-            return True
+            return not self._should_stop()
         except ClientConnectionTimeout:
             self.cancel()
             log.debug('Timer cancelled because client connection timed out')
--- a/nicegui/elements/timer.py
+++ b/nicegui/elements/timer.py
@@ -3,9 +3,8 @@ from contextlib import AbstractContextManager, nullcontext
 from typing_extensions import Self
 
 from .. import core
-from ..client import Client, ClientConnectionTimeout
+from ..client import Client
 from ..element import Element
-from ..logging import log
 from ..timer import Timer as BaseTimer
 
 
@@ -26,13 +25,8 @@ class Timer(BaseTimer, Element, component='timer.js'):
         """
         if self._should_stop():
             return False
-        try:
-            await self.client.connected()
-            return not self._should_stop()
-        except ClientConnectionTimeout:
-            self.cancel()
-            log.debug('Timer cancelled because client connection timed out')
-            return False
+        await self.client.connected()
+        return not self._should_stop()
 
     def _should_stop(self) -> bool:
         return (
--- a/tests/test_timer.py
+++ b/tests/test_timer.py
@@ -288,13 +288,13 @@ async def test_no_error_when_timer_is_deleted_while_waiting_for_connection(user:
         containers.append(container)
 
     exceptions: list[Exception] = []
-    app.on_exception(lambda e: exceptions.append(e))
+    app.on_exception(exceptions.append)
 
     await user.http_client.get('/')  # request the page without ever opening the websocket
     await asyncio.sleep(0)
     containers.pop().delete()  # delete the timer and drop the last reference to its parent slot
     Client.prune_instances(client_age_threshold=0)  # delete the client, waking up connected()
     gc.collect()
-    await asyncio.sleep(0.5)
+    await asyncio.sleep(0.3)
 
     assert not exceptions
--- a/tests/test_timer.py
+++ b/tests/test_timer.py
@@ -275,3 +275,26 @@ def test_no_leak_when_client_deleted(screen: Screen):
     Client.prune_instances(client_age_threshold=0)
     screen.wait(1)
     assert not any(isinstance(obj, ui.timer) for obj in gc.get_objects())
+
+
+@pytest.mark.parametrize('kwargs', [{}, {'once': True}, {'immediate': False}], ids=['repeating', 'once', 'delayed'])
+async def test_no_error_when_timer_is_deleted_while_waiting_for_connection(user: User, kwargs: dict):
+    containers: list[ui.column] = []
+
+    @ui.page('/')
+    def page():
+        with ui.column() as container:
+            ui.timer(0.2, lambda: None, **kwargs)
+        containers.append(container)
+
+    exceptions: list[Exception] = []
+    app.on_exception(lambda e: exceptions.append(e))
+
+    await user.http_client.get('/')  # request the page without ever opening the websocket
+    await asyncio.sleep(0)
+    containers.pop().delete()  # delete the timer and drop the last reference to its parent slot
+    Client.prune_instances(client_age_threshold=0)  # delete the client, waking up connected()
+    gc.collect()
+    await asyncio.sleep(0.5)
+
+    assert not exceptions
--- a/tests/test_timer.py
+++ b/tests/test_timer.py
@@ -287,14 +287,10 @@ async def test_no_error_when_timer_is_deleted_while_waiting_for_connection(user:
             ui.timer(0.2, lambda: None, **kwargs)
         containers.append(container)
 
-    exceptions: list[Exception] = []
-    app.on_exception(exceptions.append)
-
     await user.http_client.get('/')  # request the page without ever opening the websocket
     await asyncio.sleep(0)
     containers.pop().delete()  # delete the timer and drop the last reference to its parent slot
     Client.prune_instances(client_age_threshold=0)  # delete the client, waking up connected()
     gc.collect()
     await asyncio.sleep(0.3)
-
-    assert not exceptions
+    # no assertion needed: the user fixture fails the test if the woken timer raises and logs an ERROR


falkoschindler (approved):
Impressive work — both the fix and the PR description. A deep review confirmed the claims empirically (revert, guard mutations, `_cleanup()` paths) and found no bugs.

I pushed two follow-up commits directly to the branch:

- Remove the unreachable `except ClientConnectionTimeout` branch from `_can_start()` — `connected()` is called without a timeout, so it can never raise — and add a `NOTE` in `Client.delete()` protecting the ordering invariant the fix relies on.
- Slim the test down to the `user` fixture's built-in ERROR-log check: the explicit exception collection was redundant with it, since `app.on_exception` only adds to the default `log.exception` handler. The test still fails on unfixed code in all three parametrizations.

The deeper issue the description hints at — `connected()` reporting success on a deleted client — stays real for other callers; that's now tracked in #6277 (element-owned awaits) and #6278 (`run_javascript`).

PR #6237, opened 2026-08-04
Keep the tab alive when a stale socket disconnects

evnchn:
### Motivation

`Client.handle_disconnect()` clears `tab_id` unconditionally. When the browser has already re-established its socket before the server notices the old one is gone, that late disconnect clears `tab_id` out from under the **live** connection: `has_socket_connection` goes `False`, `_on_event` drops every event and the outbox stops flushing, so the page is frozen until the user reloads it by hand.

That ordering is the normal one after an abrupt network loss, not an exotic race. `nicegui.js` creates `documentId` once at mount and re-uses the same query object on socket.io reconnect, so the new socket carries the same `document_id`. The browser reconnects as soon as the network is back, while the server only learns the old socket is gone on engine.io ping timeout — `ping_interval + ping_timeout`, about 6 s at the default `reconnect_timeout=3.0`.

The asymmetry is the whole bug: **the browser reconnects fast, the server reaps slow.** Nothing is wrong with either half on its own — the damage is that the late reap acts on client-level state that the new socket is already using.

```mermaid
sequenceDiagram
    participant B as Browser
    participant S as Server
    participant C as Client

    B->>S: connect sid=A, document_id=D, tab_id=T
    S->>C: handle_handshake A, D
    Note over C: tab_id = T<br/>_num_connections[D] = 1<br/>has_socket_connection = True

    Note over B,S: network drops — no close frame ever reaches the server

    B-->>B: client-side timeout, socket.io reconnects
    B->>S: connect sid=A', document_id=D, tab_id=T
    Note right of B: same document_id:<br/>window.documentId is set once in mounted()
    S->>C: handle_handshake A', D
    Note over C: tab_id = T<br/>_num_connections[D] = 2

    Note over S: up to ~6 s after the drop —<br/>ping_interval 4 s + ping_timeout 2 s —<br/>engine.io finally reaps sid A
    S->>C: handle_disconnect A
    Note over C: tab_id = None ← cleared while A' is live<br/>has_socket_connection = False

    B->>S: event: click, from sid A'
    S--xC: _on_event returns early on not has_socket_connection
    Note over B: page frozen until the user reloads by hand
```

The two timeouts never meet: the browser's reconnect backoff starts around 1 s, while `sio.eio.ping_interval` and `ping_timeout` are floored at 4 s and 2 s regardless of `reconnect_timeout`. So the reap essentially always lands *after* the new socket has handshaken, which is why this is the ordinary outcome of a wifi flap or a laptop resume rather than a narrow race.

With the fix, only the one step changes:

```mermaid
sequenceDiagram
    participant S as Server
    participant C as Client

    S->>C: handle_disconnect A
    Note over C: _socket_to_document_id still holds A'<br/>so tab_id stays T<br/>has_socket_connection = True
    Note over C: the tab is released only when the last socket goes
```

`_socket_to_document_id` is the set of live handshaken sockets, and the disconnecting socket's entry is popped two lines earlier, so `if not self._socket_to_document_id` reads exactly as "nobody else is left".

A client-level check rather than a per-document one, because `tab_id` is itself client-level — every handshake overwrites it regardless of which document the socket belongs to, so filtering by `document_id` here would be inconsistent with how the field is written. And it does lean on engine.io eventually firing `disconnect` for every socket it registered: that is the same bet the existing teardown already makes, but the old unconditional clear did not need it and this does.

Long-standing, not a regression: the unconditional clear predates 3.x.

### Implementation

Only clear `tab_id` once no socket for this client is still connected:

```python
tab_id_to_close = self.tab_id
if not self._socket_to_document_id:  # a socket that reconnected before this one was reaped is still live
    self.tab_id = None
```

`_socket_to_document_id` is the set of live handshaken sockets, and the entry for the disconnecting socket has already been popped two lines above, so the guard reads "nobody else is left." `has_socket_connection` is client-level (`tab_id is not None`), so a client-level predicate is the matching one — a per-document check would still take the tab down while another document's socket is live.

Deliberately out of scope, both worth their own change:

- The disconnect handlers still fire for a client that stays connected. That is pre-existing and arguably intended — since 3.0 the handler is documented to run on reconnects too.
- `delete_content()` is scheduled per `document_id` but calls the client-level `self.delete()`, so a client with two `document_id`s deletes itself twice. Separate defect, separate fix.

<details>
<summary>Evidence: the race is observable in a real browser, and the test fails for the right reason</summary>

Chrome, real socket.io reconnect, no internals touched. The browser drops its transport, reconnects with a **new sid and the same `documentId`**, and the server only reaps the old socket ~6 s later on its engine.io ping timeout.

The test pins that ordering through public API — `client.on_disconnect` — rather than assuming it, which also answers "could it pass for the wrong reason". On unmodified `main` both ordering assertions **pass** (no disconnect seen at reconnect time, exactly one seen after the wait) and only the click assertion fails:

```
=== on main ===
assert events['disconnects'] == 0   # passed - reconnect really did precede the reap
assert events['disconnects'] == 1   # passed - the reap really did happen
assert events['clicks'] == 1        # FAILED: assert 0 == 1
1 failed in 12.34s

=== with the fix ===
1 passed in 13.54s
```

The reverts were done with `git show origin/main:nicegui/client.py > nicegui/client.py`. An earlier attempt used `git stash push` against an already-committed change, which stashed nothing and reported a pass against the fixed code — worth naming, because that is exactly how a test nobody has seen fail gets shipped.

The wait for the reconnect is a condition wait; the 9 s before the reap is a fixed sleep because the floors make it irreducible — `ping_interval` and `ping_timeout` are `max(reconnect_timeout * 0.8, 4)` and `max(reconnect_timeout * 0.4, 2)`, so no `reconnect_timeout` setting brings the reap under 6 s. That is the honest cost of this guard, and PR 6219's "not every change needs a test" applies if you would rather not pay it.

Local gates: `pre-commit` clean on both files, `mypy ./nicegui` clean over 245 files, `pylint ./nicegui` 10.00/10, `pytest tests/test_reconnect.py tests/test_lifecycle.py` 10 passed / 0 skipped. Fork CI green on the pre-squash commit.

</details>

<details>
<summary>How it was found</summary>

A Hypothesis `RuleBasedStateMachine` over the client lifecycle, driving a real server over real HTTP and real `python-socketio` connections, with "a client with a live handshaken socket must consider itself connected" as an invariant. It shrank to four operations: load the page, connect, connect again with the same `document_id`, disconnect the first socket. The browser test above is the same sequence, produced by Chrome rather than by the state machine.

</details>

### Progress


falkoschindler (approved):
Nice find, and the fix is exactly right. `_socket_to_document_id` is the correct thing to check: it is the synchronous, client-wide set of live sockets (entry added in `handle_handshake`, removed in `handle_disconnect` before the guard runs), so "map empty" means "no socket left" without any lag. The tempting alternative `_num_connections` would be wrong here — its zeroed entries linger in the defaultdict until `delete_content` pops them after the reconnect timeout, and it counts per document while `tab_id` is client state.

I took this through a deep review with adversarial verification of the failure scenarios (reverse ordering with an unreaped zombie socket, orphaned sids on Air, handshake/disconnect races, multiple documents per client): everything either can't happen, self-heals within one ping window on an already-dead client, or is pre-existing behavior identical on main. The one thing worth a follow-up someday is that `_num_connections` is now provably redundant with `_socket_to_document_id` — two structures tracking one fact — but that predates this PR.

I pushed one commit onto the branch: the guard's comment now states the rule instead of the skipped branch, and the test no longer needs the 9 s ping-timeout wait — it keeps a reference to the real websocket, so closing it after the reconnect makes the server process the stale disconnect immediately, in the same order as the bug. Runs in under a second now and still fails when the fix is reverted (verified).

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20analysis%20and%20wording%20from%20the%20model%2C%20verified%20by%20a%20multi-agent%20sweep%20with%20adversarial%20checks%3B%20direction%20and%20final%20call%20from%20the%20maintainer.&a=claude-code "claude-fable-5: Review analysis and wording from the model, verified by a multi-agent sweep with adversarial checks; direction and final call from the maintainer.")

PR #6239, opened 2026-08-04
Document Socket.IO client_id forging as a commonly misreported pattern

falkoschindler:
### Motivation

We repeatedly receive security advisories reporting that NiceGUI's Socket.IO event handlers dispatch on the client-supplied `client_id` without binding it to the authenticated socket `sid` (most recently GHSA-2563-89rp-gjwq, a duplicate of the already-triaged GHSA-wp75-9jqx-j6f6). These reports overlook that `client_id` is an unguessable per-session secret and that the suggested `sid`-binding fix is defeated by the unauthenticated handshake. Documenting the pattern lets us point future reporters at a clear explanation and keeps the advisory queue focused on real issues.

### Implementation

Adds one bullet to the "Commonly Misreported Patterns" section of `SECURITY.md`, matching the style of the existing entries: it names the affected handlers, explains why `client_id` is treated as a secret capability, why binding `sid` at handshake does not help, and notes that cross-origin connections are a separate, already-addressed concern.

### Progress

- [x] This PR does not address a security issue. (Documentation only — it clarifies triage of a known non-issue, no code or behavior change.)
- [x] Pytests are not necessary (documentation-only change).


evnchn:
@falkoschindler it might be a good idea to expand this to the entire UUID anti-collision assumption, seeing it is so many bits, considering that we have **many** IDs floating around which are UUID

HOWEVER the assumption only holds, when there's not a lot, magnitude speaking, of IDs issued. Otherwise the assumption falls apart.

evnchn:
Moreover, our website covers this:

> Treat client_id like a session token

At https://nicegui.io/documentation/section_security

It might be a good idea to have SECURITY.md reference the documentation's security section so that we have one single source of truth?

evnchn:

**Two files, not one** — `SECURITY.md` answers *"should I file this?"*, the docs answer *"how do I build this safely?"*, so the triage verdict stays here, the id taxonomy goes on the docs page, and this PR can stay `SECURITY.md`-only.

<details>
<summary><b>Proposed prose for both, and why the split falls there</b></summary>

**`SECURITY.md`** — section intro gains the pointer, which also lets the new bullet stop restating *"Treat `client_id` like a session token"* from the docs:

```markdown
The [Security Best Practices](https://nicegui.io/documentation/section_security) documentation is the canonical reference for the model itself
(what the framework guarantees, what the application must do); this list only records why each pattern is repeatedly misread.
```

…and one more bullet, so the next report of this shape can name any of our ids rather than only `client_id`:

```markdown
- **Guessing a `uuid4` identifier** (`client_id`, the `app.storage.user` id, event-listener / Leaflet-layer / scene-object ids).
  Not every id is a secret, and the ones that are carry 122 bits of platform-CSPRNG randomness — the documentation describes [which is which](https://nicegui.io/documentation/section_security).
  The margins do shrink as more ids are issued, but realistic deployments stay far below either bound: at a billion ids the collision probability is around 10^-19, and a blind guess against a million live sessions succeeds with probability around 10^-31.
  So assuming such an id can simply be guessed is not a finding — showing a leak path, predictable generation, or a route that treats a public id as authority **is**.
```

**`website/documentation/content/section_security.py`** — appended to *Client-Side Secrets*, since this is the part an app author acts on (what may be logged, put in a URL, or handed to a third party):

```markdown
**Not every random id is a secret.**

NiceGUI mints `uuid4` values for several jobs, and they do not all carry the same authority:

- **Public by design** — event listener, Leaflet layer and scene object ids travel to the browser in plain sight.
  They only need to be unique; knowing one grants nothing.
- **A bearer capability** — the `client_id` above. Anyone who knows it can act as that client.
- **A storage key** — the `app.storage.user` id lives in a signed session cookie.
  The signature is the access boundary there, not the id, so guessing it alone does not unlock another user's storage.

Only the second class needs protecting as a secret.
Randomness is not what fails in practice — leakage is, which is what the list below guards against.
```

**On my own comment above:** I wrote "the UUID anti-collision assumption" as if it were one assumption, and it is three — lumping the `app.storage.user` id in with `client_id` would have overstated it, since an attacker who guesses that id still cannot present a validly signed cookie. A second-opinion pass (Codex) caught that in my first draft.

The magnitude caveat is real but cuts differently for the two failure modes, which is why the bullet gives both numbers and then points at leakage rather than arithmetic:

```
uuid4 = 122 random bits          CPython: os.urandom(16), version/variant bits fixed
p_collision ≈ n² / 2¹²³          n = 1e9              → 9.4e-20
50 % birthday point               1.1774·√(2¹²²)       → 2.7e18 ids
p_blind_guess = N / 2¹²²          N = 1e6 live ids     → 1.9e-31 per attempt
```

Happy to open the docs half as its own PR so this one keeps its scope.

</details>

falkoschindler:
Good points, both adopted:

- The section intro now names the [Security Best Practices](https://nicegui.io/documentation/section_security) documentation as the canonical reference for the security model, so `SECURITY.md` stays focused on triage.
- A new bullet **"Guessing a `uuid4` identifier"** generalizes beyond `client_id` to all the ids we mint (I kept your probability figures — they make the "not a finding" argument concrete, magnitude caveat included).
- The `client_id` bullet no longer restates the session-token advice from the docs.

Please go ahead and open the docs PR with the id taxonomy — the new bullet's "describes which is which" link presumes it. Once that's on its way, this PR should be ready for your review.

evnchn (approved):
Positive review by agent with no past memory as well

One non-blocking (IMO) issue: we didn't really "address" but we "acknowledge it is a non-threat", for:

> Cross-origin connections are a separate concern already addressed independently

So yeah this is pretty solid let's merge!

PR #6242, opened 2026-08-06
Fix `ui.scene.move_camera()` recreating the camera controls when the up vector is unchanged

### LINKED ISSUE #6241: `ui.scene.move_camera()` always disposes and recreates the controls object, discarding user configuration
### First Check

### Example Code

```python
info_string = """
    Click the buttons in order.

    Actual:
    1. lock rotation  ->  enableRotate=False  sentinel=mine
    3. read state     ->  enableRotate=True   sentinel=gone     # controls object replaced

    Dragging the scene also rotates again after step 2, confirming the lock is gone.

    Expected:
    move_camera() should not reset controls configuration when the up vector has 
    not changed (maybe not even in that case). """

print(info_string)


from nicegui import ui

scene = ui.scene(width=600, height=400)
with scene:
    ui.scene.box()


async def read_state() -> None:
    state = await ui.run_javascript(f'''
        const el = getElement({scene.id});
        return {{rotate: el.controls.enableRotate, sentinel: el.controls.__sentinel ?? 'gone'}};
    ''')
    ui.notify(f'enableRotate={state["rotate"]}  sentinel={state["sentinel"]}')


async def lock_rotation() -> None:
    await ui.run_javascript(f'''
        const el = getElement({scene.id});
        el.controls.enableRotate = false;
        el.controls.__sentinel = 'mine';
    ''')
    await read_state()


ui.button('1. lock rotation', on_click=lock_rotation)
ui.button('2. move_camera', on_click=lambda: scene.move_camera(x=5, y=-5, z=5, duration=0.0))
ui.button('3. read state', on_click=read_state)

ui.run(reload=False)
```

### Description

`scene.move_camera()` destroys and recreates the Three.js controls object on every call, resetting any configuration applied to it (enableRotate, screenSpacePanning, mouseButtons…) to the mounted control class's defaults.

scene.js guards the rebuild with `camera_up_changed`
https://github.com/zauberzeug/nicegui/blob/1e3b0214db5c20e19d3e106d0f2e383835ac20b8/nicegui/elements/scene/scene.js#L511
but scene.py resolves each up_* argument against the stored camera before sending it, so the value is never null and the guard is always true.

A **Fix** could be to pass the raw arguments to run_method instead of the resolved camera state, so scene.js's existing null fallback works as intended. Instead, an **enhancement** could expose the control flags `screenSpacePanning, enableRotate, enablePan, enableZoom, mouseButtons ` as ui.scene parameters. 

### NiceGUI Version

3.14.0

### Python Version

3.13

### Browser

Firefox

### Operating System

Linux

### LINKED ISSUE #6247: TrackballControls keeps a stale canvas rect after a window resize, so ui.scene rotation becomes misaligned
### Example Code

```python
from nicegui import ui


@ui.page('/')
def index() -> None:
    scene = ui.scene(control_type='trackball').classes('w-full h-64')
    scene.box().material('#4488ff')
    label = ui.label()

    async def compare() -> None:
        cached, actual = await ui.run_javascript(f'''
            const el = getElement({scene.id});
            const rect = el.renderer.domElement.getBoundingClientRect();
            return [[el.controls.screen.width, el.controls.screen.height], [rect.width, rect.height]];
        ''')
        label.text = f'controls.screen={cached} canvas={actual} -> {"OK" if cached == actual else "STALE"}'

    ui.timer(0.5, compare)

ui.run()
```

### Description

`ui.scene(control_type='trackball')` stops tracking the mouse correctly after the browser window is resized.

TrackballControls caches the canvas rectangle in `this.screen` and refreshes it **only** in its constructor (`null !== domElement && (this.connect(domElement), this.handleResize())`). Nothing else ever calls `handleResize()`: three.js registers no `resize` listener of its own, and NiceGUI's own resize handler (`scene.js` → `resize()`, wired at `window.addEventListener('resize', this.resize)`) updates the renderers and the camera projection but not the controls.

Since `getMouseOnScreen()` / `getMouseOnCircle()` normalise pointer coordinates against that cached rectangle, after a resize the scene rotates by the wrong amount around an off-centre pivot until the page is reloaded.

**Steps:** run the example, resize the browser window, and watch the label.

**Expected:** `controls.screen` follows the canvas, label stays `OK`.
**Actual:** `controls.screen` keeps the old size, label flips to `STALE` and stays there; dragging to rotate is now misaligned.

Observed output (Chrome, headless, resizing the window twice):

```
initial window 1000x800   -> controls.screen=[968, 256] canvas=[968, 256]  -> OK
after resize to 700x620   -> controls.screen=[968, 256] canvas=[653, 256]  -> STALE
after resize to 1100x900  -> controls.screen=[968, 256] canvas=[1068, 256] -> STALE
```

Calling `getElement(<id>).controls.handleResize()` from the console restores correct behaviour immediately, which localises it to the cached rect and nothing else.

The user-visible effect — the **same** 120 px drag from the canvas centre, before and after that repair call:

trackball rotation after a window resize: stale cached rect vs after handleResize()

<details>
<summary>Automated version of the same check (pytest, passes on main)</summary>

```python
def test_trackball_screen_rect_goes_stale(screen: Screen):
    scene = None

    @ui.page('/')
    def page():
        nonlocal scene
        scene = ui.scene(control_type='trackball').classes('w-full h-64')
        scene.box()

    screen.open('/')
    screen.wait_for(lambda: screen.selenium.execute_script(
        f'return !!window.getElement && !!getElement({scene.id})?.controls'))

    def rects():
        return screen.selenium.execute_script(f'''
            const el = getElement({scene.id});
            const r = el.renderer.domElement.getBoundingClientRect();
            return {{cached: [el.controls.screen.width, el.controls.screen.height],
                     actual: [r.width, r.height]}};
        ''')

    before = rects()
    assert before['cached'] == before['actual']       # fresh controls agree

    screen.selenium.set_window_size(700, 620)
    screen.wait(1.0)
    after = rects()
    assert after['cached'] != after['actual']         # stale after the resize

    screen.selenium.execute_script(f'getElement({scene.id}).controls.handleResize()')
    assert rects()['cached'] == rects()['actual']     # repaired
```

```
BEFORE resize: {'actual': [724, 256], 'cached': [724, 256]}
AFTER  resize: {'actual': [668, 256], 'cached': [724, 256]}
AFTER  handleResize(): {'actual': [668, 256], 'cached': [668, 256]}
1 passed
```

</details>

`control_type='orbit'` and `'map'` are unaffected — OrbitControls works from pointer deltas relative to the element and caches no rectangle.

Note this is partly masked today: `move_camera()` currently disposes and recreates the controls on every call, which incidentally re-runs the constructor and therefore `handleResize()`. So an app that moves the camera after a resize accidentally repairs itself. PR #6242, which stops that unconditional rebuild (a good change in its own right), removes the masking — worth fixing this independently either way.

### NiceGUI Version

main @ 1e3b0214 (3.13.0)

### Python Version

3.12.11

### Browser

Chrome

### Operating System

macOS

ptruka:
### Motivation

Fixes #6241.
Fixes #6247.

`scene.move_camera()` disposes and recreates the Three.js controls object on every call, so any configuration applied to it (`enableRotate`, `screenSpacePanning`, `mouseButtons`, custom properties) is reset to the control class defaults whenever the camera moves.

`scene.js` already intends this to be conditional (`camera_up_changed`), but `scene.py` resolves each `up_*` argument against the stored camera before sending it, so the values are never `null` and the guard is always true.
https://github.com/zauberzeug/nicegui/blob/1e3b0214db5c20e19d3e106d0f2e383835ac20b8/nicegui/elements/scene/scene.py#L261

Because the rebuild happens in the tween's `onComplete`, a `ui.run_javascript()` issued right after `move_camera()` is also a race: whether the configuration survives depends on round-trip timing.

### Implementation

The controls remember the up vector they were built for (`controls_up`), and `move_camera()` only rebuilds them when the requested up vector differs from that one. Comparing against the *live* `camera.up` would not work: TrackballControls rotates `camera.up` while the user drags, and an interrupted tween leaves it mid-interpolation, so `camera.up` can differ from the vector the controls were built for without any real change having been requested.

Since `scene.py` always sends resolved values, the `=== null` fallbacks in `move_camera()` were dead code and are removed.

Two things used to happen as a side effect of the unconditional rebuild and are now done explicitly:

- `resize()` calls `controls.handleResize()`, which TrackballControls needs to map pointer coordinates to the canvas (OrbitControls/MapControls read the canvas size live).
- Zero-duration camera moves settle synchronously instead of on the next render frame. This also closes the race from the motivation for the initial `move_camera(duration=0)` replay: `await scene.initialized()` followed by `ui.run_javascript(...)` used to lose the configuration again for scenes with a non-default up vector.

Keeping the controls alive changes a few details for code that pokes `controls` directly: momentum/damping state (Trackball `staticMoving=false`, Orbit `enableDamping`) is no longer discarded when a move ends, the `reset()`/`saveState()` baseline is only refreshed when the up vector changes, and user-set limits (`maxPolarAngle`, `maxDistance`, ...) now apply to the commanded pose.

This is done entirely in `scene.js`; the Python API and the arguments sent to the client are unchanged. _An alternative discussed in the issue: sending the raw arguments so the existing `null` handling applies. This would additionally change which pose wins for unspecified axes (Python's stored camera state vs. the browser's current pose), which is a separate behavioral question and is deliberately not part of this PR._

`ui.scene_view` resolves `up_*` the same way but creates no controls object, so it is unaffected.

Tests: the configuration survives a plain camera move and is reset when the up vector changes (OrbitControls); it survives a move after the user rotated a TrackballControls scene; TrackballControls follow the canvas size; and configuration applied after `await scene.initialized()` survives the initial camera move.

### Progress

### DIFF
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -194,7 +194,7 @@ export default {
       this.scene.add(grid);
     }
     this.controlClass = { trackball: TrackballControls, map: MapControls }[this.controlType] || OrbitControls;
-    this.create_controls();
+    this.create_controls(this.camera.up);
     this.drag_controls = new DragControls(this.draggable_objects, this.camera, this.renderer.domElement);
     this.drag_controls.transformGroup = true;
     const applyConstraint = (constraint, position) => {
@@ -506,20 +506,17 @@ export default {
       this.move(object_id, x, y, z);
       this.rotate(object_id, R);
     },
-    create_controls() {
+    create_controls(up) {
       this.controls = new this.controlClass(this.camera, this.renderer.domElement);
-      // TrackballControls rotates the camera's up vector, so the vector the controls were created for
-      // must be remembered separately to tell a user rotation apart from an actual camera move
-      this.controls_up = this.camera.up.clone();
+      // remember the up vector the controls were built for: camera.up may differ from it after a user rotation
+      // (TrackballControls) or an interrupted tween, so it cannot be used to decide whether a rebuild is needed
+      this.controls_up = up.clone();
     },
     move_camera(x, y, z, look_at_x, look_at_y, look_at_z, up_x, up_y, up_z, duration) {
+      if (!this.controls) return; // WebGL renderer could not be created
       if (this.camera_tween) this.camera_tween.stop();
-      const target_up = new THREE.Vector3(
-        up_x === null ? this.camera.up.x : up_x,
-        up_y === null ? this.camera.up.y : up_y,
-        up_z === null ? this.camera.up.z : up_z,
-      );
-      // the controls are only rebuilt if the up vector they were created for really changes,
+      const target_up = new THREE.Vector3(up_x, up_y, up_z);
+      // the controls are only rebuilt if the up vector they were built for really changes,
       // because rebuilding resets their configuration
       const camera_up_changed = !this.controls_up.equals(target_up);
       this.camera_tween = new TWEEN.Tween([
@@ -533,20 +530,7 @@ export default {
         this.look_at.y,
         this.look_at.z,
       ])
-        .to(
-          [
-            x === null ? this.camera.position.x : x,
-            y === null ? this.camera.position.y : y,
-            z === null ? this.camera.position.z : z,
-            target_up.x,
-            target_up.y,
-            target_up.z,
-            look_at_x === null ? this.look_at.x : look_at_x,
-            look_at_y === null ? this.look_at.y : look_at_y,
-            look_at_z === null ? this.look_at.z : look_at_z,
-          ],
-          duration * 1000,
-        )
+        .to([x, y, z, up_x, up_y, up_z, look_at_x, look_at_y, look_at_z], duration * 1000)
         .onUpdate((p) => {
           this.camera.position.set(p[0], p[1], p[2]);
           this.camera.up.set(p[3], p[4], p[5]); // before calling lookAt
@@ -556,14 +540,14 @@ export default {
         })
         .onComplete(() => {
           if (camera_up_changed) {
-            this.camera.up.copy(target_up); // remove interpolation residue, so the next comparison is exact
             this.controls.dispose();
-            this.create_controls();
+            this.create_controls(target_up);
             this.controls.target.copy(this.look_at);
             this.camera.lookAt(this.look_at);
           }
         })
         .start();
+      if (duration === 0) this.camera_tween.update(); // settle immediately, so subsequent JavaScript sees the final state
     },
     get_camera() {
       return {
@@ -593,6 +577,7 @@ export default {
         this.camera.right = (this.camera.aspect * this.cameraParams.size) / 2;
       }
       this.camera.updateProjectionMatrix();
+      this.controls.handleResize?.(); // TrackballControls caches the canvas rect
     },
     init_objects(data) {
       this.resize();
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -508,7 +508,13 @@ export default {
     },
     move_camera(x, y, z, look_at_x, look_at_y, look_at_z, up_x, up_y, up_z, duration) {
       if (this.camera_tween) this.camera_tween.stop();
-      const camera_up_changed = up_x !== null || up_y !== null || up_z !== null;
+      const target_up = new THREE.Vector3(
+        up_x === null ? this.camera.up.x : up_x,
+        up_y === null ? this.camera.up.y : up_y,
+        up_z === null ? this.camera.up.z : up_z,
+      );
+      // NOTE: the controls are only rebuilt if the up vector really changes, because that resets their configuration
+      const camera_up_changed = !this.camera.up.equals(target_up);
       this.camera_tween = new TWEEN.Tween([
         this.camera.position.x,
         this.camera.position.y,
@@ -525,9 +531,9 @@ export default {
             x === null ? this.camera.position.x : x,
             y === null ? this.camera.position.y : y,
             z === null ? this.camera.position.z : z,
-            up_x === null ? this.camera.up.x : up_x,
-            up_y === null ? this.camera.up.y : up_y,
-            up_z === null ? this.camera.up.z : up_z,
+            target_up.x,
+            target_up.y,
+            target_up.z,
             look_at_x === null ? this.look_at.x : look_at_x,
             look_at_y === null ? this.look_at.y : look_at_y,
             look_at_z === null ? this.look_at.z : look_at_z,
--- a/nicegui/elements/scene/scene.js
+++ b/nicegui/elements/scene/scene.js
@@ -194,7 +194,7 @@ export default {
       this.scene.add(grid);
     }
     this.controlClass = { trackball: TrackballControls, map: MapControls }[this.controlType] || OrbitControls;
-    this.controls = new this.controlClass(this.camera, this.renderer.domElement);
+    this.create_controls();
     this.drag_controls = new DragControls(this.draggable_objects, this.camera, this.renderer.domElement);
     this.drag_controls.transformGroup = true;
     const applyConstraint = (constraint, position) => {
@@ -506,6 +506,12 @@ export default {
       this.move(object_id, x, y, z);
       this.rotate(object_id, R);
     },
+    create_controls() {
+      this.controls = new this.controlClass(this.camera, this.renderer.domElement);
+      // TrackballControls rotates the camera's up vector, so the vector the controls were created for
+      // must be remembered separately to tell a user rotation apart from an actual camera move
+      this.controls_up = this.camera.up.clone();
+    },
     move_camera(x, y, z, look_at_x, look_at_y, look_at_z, up_x, up_y, up_z, duration) {
       if (this.camera_tween) this.camera_tween.stop();
       const target_up = new THREE.Vector3(
@@ -513,8 +519,9 @@ export default {
         up_y === null ? this.camera.up.y : up_y,
         up_z === null ? this.camera.up.z : up_z,
       );
-      // NOTE: the controls are only rebuilt if the up vector really changes, because that resets their configuration
-      const camera_up_changed = !this.camera.up.equals(target_up);
+      // the controls are only rebuilt if the up vector they were created for really changes,
+      // because rebuilding resets their configuration
+      const camera_up_changed = !this.controls_up.equals(target_up);
       this.camera_tween = new TWEEN.Tween([
         this.camera.position.x,
         this.camera.position.y,
@@ -549,8 +556,9 @@ export default {
         })
         .onComplete(() => {
           if (camera_up_changed) {
+            this.camera.up.copy(target_up); // remove interpolation residue, so the next comparison is exact
             this.controls.dispose();
-            this.controls = new this.controlClass(this.camera, this.renderer.domElement);
+            this.create_controls();
             this.controls.target.copy(this.look_at);
             this.camera.lookAt(this.look_at);
           }
--- a/nicegui/elements/scene/scene_view.js
+++ b/nicegui/elements/scene/scene_view.js
@@ -124,20 +124,7 @@ export default {
         this.look_at.y,
         this.look_at.z,
       ])
-        .to(
-          [
-            x === null ? this.camera.position.x : x,
-            y === null ? this.camera.position.y : y,
-            z === null ? this.camera.position.z : z,
-            up_x === null ? this.camera.up.x : up_x,
-            up_y === null ? this.camera.up.y : up_y,
-            up_z === null ? this.camera.up.z : up_z,
-            look_at_x === null ? this.look_at.x : look_at_x,
-            look_at_y === null ? this.look_at.y : look_at_y,
-            look_at_z === null ? this.look_at.z : look_at_z,
-          ],
-          duration * 1000,
-        )
+        .to([x, y, z, up_x, up_y, up_z, look_at_x, look_at_y, look_at_z], duration * 1000)
         .onUpdate((p) => {
           this.camera.position.set(p[0], p[1], p[2]);
           this.camera.up.set(p[3], p[4], p[5]); // before calling lookAt
--- a/tests/test_scene.py
+++ b/tests/test_scene.py
@@ -291,6 +291,28 @@ def test_custom_controls(screen: Screen, control_type: Literal['map', 'trackball
     assert screen.selenium.execute_script(f'return getElement({scene.id}).controls.constructor.name') == constructor
 
 
+def test_moving_camera_keeps_controls_unless_up_vector_changes(screen: Screen):
+    scene = None
+
+    @ui.page('/')
+    def page():
+        nonlocal scene
+        scene = ui.scene()
+
+    screen.open('/')
+    screen.wait_for(lambda: scene is not None)
+    enable_rotate = f'getElement({scene.id}).controls.enableRotate'
+    screen.selenium.execute_script(f'{enable_rotate} = false')
+
+    scene.move_camera(x=1, duration=0)
+    screen.wait_for(lambda: screen.selenium.execute_script(f'return getElement({scene.id}).camera.position.x') == 1)
+    assert screen.selenium.execute_script(f'return {enable_rotate}') is False, 'controls survive a plain camera move'
+
+    scene.move_camera(up_y=1, up_z=0, duration=0)
+    screen.wait_for(lambda: screen.selenium.execute_script(f'return getElement({scene.id}).camera.up.y') == 1)
+    assert screen.selenium.execute_script(f'return {enable_rotate}') is True, 'controls are rebuilt for a new up vector'
+
+
 async def test_dragend_after_object_deleted(user: User):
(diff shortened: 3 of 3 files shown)


evnchn:
Nice catch, and thanks for the fix + test! 🙏 The old condition really was always true (`move_camera` resolves the up vector Python-side and always sends concrete floats), so the controls got thrown away on every camera move — comparing values instead is the right idea.

Two things I ran into while looking at it:

**1. `control_type='trackball'` slips through.** TrackballControls mutates `camera.up` in place while rotating (`object.up.applyQuaternion(...)` in `_rotateCamera`), so once the user has dragged the scene, the live `camera.up` no longer matches the sticky server-side value and the rebuild fires again. Same for a non-axis-aligned up: the tween writes `start + (end - start)`, which can land 1 ULP off, so ~~`equals()` stays false forever~~ **[corrected below — it costs one extra rebuild, not a permanent one]** after e.g. `move_camera(up_x=0.1, up_y=0.2, up_z=0.3)`.

Both would disappear if, instead of disposing and recreating, we just refreshed the one thing that actually goes stale — OrbitControls caches `_quat`/`_quatInverse` from `camera.up` at construction time and never updates them. That would also stop a *genuine* up change from wiping the user's controls config, which the current code still does. Happy to be wrong here — what do you think?

**2. The test may be flaky.** `wait_for(... camera.position.x == 1)` is an exact float compare, but `controls.update()` runs every frame after the tween and re-derives the position through a spherical round-trip, so x often comes back as `1.0000000000000002`. `pytest.approx` (or reading `camera_tween._object`, like tests/test_scene_view.py does) would make it deterministic.

<details>
<summary>Two smaller notes</summary>

- The rebuild was also the only thing that ever re-ran `TrackballControls.handleResize()` — it caches the canvas rect in its constructor and nothing else refreshes it, so after a window resize trackball dragging will now stay misaligned until reload. This turned out to be a pre-existing bug on `main` rather than anything this PR introduces, so I filed it separately as #6247 — nothing to do here.
- `CONTRIBUTING.md` reserves `NOTE:` for cross-file coupling / mirrored changes; this one is local rationale, so a plain comment fits better.

</details>

None of this is a big deal — the direction is good, and the trackball case is easy to miss. Let me know if you'd like a hand with any of it!

---

**Edit — I ran both claims instead of leaving them as theory, and one was overstated. Sorry about that.** 🙇

The float point is **one extra rebuild, not a permanent one**. The residue is real, but the following move tweens from it onto the exact value and settles:

```
[after up=(0.1, 0.2, 0.3)] browser camera.up = [0.1, 0.2, 0.30000000000000004]
  python-side sticky value  = [0.1, 0.2, 0.3]
[plain move #1] controls REBUILT
[plain move #2] controls ALIVE
[plain move #3] controls ALIVE
```

Across 484 realistic up-vector component pairs, 96 leave residue and **0** fail to converge on the next move. A minor wart — please don't let it hold the PR up.

The trackball point does hold, and re-arms on every drag:

```
controls class: TrackballControls
up before drag: [0, 0, 1]
[no drag]    plain move_camera -> controls ALIVE
up after drag: [-0.4357545513303538, 0.3486541493959340, -0.8297941040426114]
[after drag] plain move_camera -> controls REBUILT
```

Same one-shot pattern, but a user drags constantly, so in practice it keeps firing.

ptruka:
Thanks, you're right that the trackball case was broken, and my fix was wrong in a way the orbit test couldn't catch. It's fixed in the latest push, along with your two smaller notes.

The mistake was checking against the live camera.up. The rebuild actually needs the up-vector that the controls were originally built with. With TrackballControls, those two stop matching as soon as the user drags. So now that vector gets saved when the controls are created, and both places that create controls call a single create_controls().

evnchn (dismissed):

**Verdict: it holds.** I re-ran both of my original checks against aa509be4 and both are clean.

Snapshotting the up-vector at construction is the right invariant, and it's a better one than I was reaching for. Because *both* guard inputs are now persistent — the sticky Python value and `controls_up` — the guard also survives a tween that gets interrupted mid-flight, which comparing against the live `camera.up` would not have.

You also closed the float-residue wart I explicitly said not to hold the PR up for: `camera.up.copy(target_up)` makes the next comparison bit-exact, so it's zero extra rebuilds now instead of one.

Nothing blocking from me. One nit worth a line (an adversarial second-model pass found a narrow path where the new guard is wrong; I ran it and it reproduces), plus two notes for whoever merges — neither of those is on you.

<details>
<summary><b>Both original findings re-checked — before vs. after</b></summary>

Same two MREs from my first pass, unchanged, run against the new head.

**Trackball drift** — a real mouse drag, then plain `move_camera(x=...)`:

| | 9fbc6cb8 (previous push) | aa509be4 (now) |
|---|---|---|
| plain move, no drag | ALIVE | ALIVE |
| plain move after drag | **REBUILT** | ALIVE |
| plain move again | **REBUILT** | ALIVE |

```
controls class: TrackballControls
up before drag: [0, 0, 1]
[no drag]    plain move_camera -> controls ALIVE
up after drag:  [-0.43575455133035385, 0.34865414939593403, -0.8297941040426114]
[after drag] plain move_camera -> controls ALIVE
[after drag] plain move_camera again -> controls ALIVE
```

**Float residue** — after `move_camera(up_x=0.1, up_y=0.2, up_z=0.3)`:

```
[after up=(0.1, 0.2, 0.3)] browser camera.up = [0.1, 0.2, 0.3]
  python-side sticky value    = [0.1, 0.2, 0.3]
  bit-exact match?            = True
[plain move #1, no up args] controls ALIVE
[plain move #2, no up args] controls ALIVE
[plain move #3, no up args] controls ALIVE
```

Previously `camera.up.z` came back as `0.30000000000000004` and the following move was REBUILT. The `copy(target_up)` line removes that entirely.

</details>

<details>
<summary><b>Seen-to-fail: the new trackball test really does guard the bug</b></summary>

Three-way, reverting only `nicegui/elements/scene/scene.js` and keeping both tests:

| `scene.js` from | `..._keeps_controls_unless_up_vector_changes` | `..._keeps_trackball_controls_after_rotating` |
|---|---|---|
| `main` (unfixed) | FAIL | FAIL |
| 9fbc6cb8 (previous push) | pass | **FAIL** |
| aa509be4 (this push) | pass | pass |

The middle row is the one that matters — the new test fails on exactly the version whose orbit test couldn't catch anything, with the symptom named directly:

```
E   assert False is True
E    +  where False = execute_script('return getElement(4).controls.staticMoving')
```

Full `tests/test_scene.py` is green (20 passed, no skips), and I ran the two new tests 10× back-to-back for flakiness — 10/10. `pre-commit` clean on both changed files.

`staticMoving` is a good pick, incidentally: it's a real TrackballControls property that resets on rebuild, so the test doesn't need a marker of its own.

</details>

<details>
<summary><b>Probed for new holes — two call sequences, both clean</b></summary>

The thing I wanted to break was `controls_up` going stale, since nothing outside `create_controls()` ever writes it.

**Drag, then a genuine up change** — must still rebuild:

```
[after drag]          cup=[0, 0, 1]  up=[0.0142, 0.8795, -0.4755]
[genuine up change]   controls REBUILT   cup=[0, 1, 0]  up=[0, 1, 0]
[plain move after]    controls ALIVE     cup=[0, 1, 0]  up=[0, 1, 0]
```

**A long up-changing tween interrupted mid-flight** by another `move_camera` — the stopped tween's `onComplete` never fires, so the rebuild is skipped and `camera.up` is stranded mid-interpolation:

```
[mid-tween]                  controls ALIVE     cup=[0, 0, 1]  up=[0, 0.0926, 0.9073]
[interrupted by plain move]  controls REBUILT   cup=[0, 1, 0]  up=[0, 1, 0]
[next plain move]            controls ALIVE     cup=[0, 1, 0]  up=[0, 1, 0]
```

It self-heals: Python has already committed the new up to its sticky state, so the interrupting move carries it as `target_up`, still sees the stale `controls_up`, and rebuilds on *its* completion. This is the case the live-`camera.up` version would have gotten wrong, because mid-interpolation the live vector matches neither value.

</details>

<details>
<summary><b>Nit: the guard is wrong on the <code>null</code> up path (not reachable from <code>scene.move_camera()</code>)</b></summary>

I ran the new version past a second model with "assume this is broken" framing, because my first pass missed the trackball case and I didn't want to repeat that. It found one thing I hadn't, and it reproduces:

`target_up` falls back to the *live* `camera.up` when an up component is `null`, but the guard compares that against `controls_up`. After a trackball drag those two differ, so a JS caller that omits the up vector gets a rebuild it never asked for:

```js
getElement(id).move_camera(1, null, null, null, null, null, null, null, null, 0)
```

| `scene.js` from | `staticMoving` after the call |
|---|---|
| `main` | `true` — controls survived |
| aa509be4 | `false` — spuriously rebuilt |

So on that one path this is a small step back from `main`, which got it right by accident (all-null meant `camera_up_changed` was false).

**Reachability is zero from Python** — I checked every call site, and both `scene.py` and `scene_view.py` always resolve the sticky state before `run_method`, so `null` can only arrive from a hand-written `ui.run_javascript()` call against an undocumented internal. That's why it's a nit and not a finding.

It's still worth closing, because it's the same bug class as the one you just fixed, and the guard is the one line that now has to be right. One extra condition does it:

```js
const camera_up_changed =
  (up_x !== null || up_y !== null || up_z !== null) && !this.controls_up.equals(target_up);
```

The `up_*` values can't be `null` today, so the added clause is inert on every real call — it just makes the JS method honour the contract its own nine fallbacks advertise. The alternative reading is that those fallbacks are dead code and should go, but that's a bigger change than this PR should carry.

I applied that line before suggesting it rather than eyeballing it: the JS call above goes back to `staticMoving = true`, both of your new tests still pass, both of my MREs still come out clean, and `tests/test_scene.py` is still 20/20 with `pre-commit` green.

</details>

<details>
<summary><b>Still open, but a design question rather than a defect</b></summary>

A *genuine* up change still wipes the user's controls config, which was the second half of my first comment and is the part your reply doesn't cover:

```
[after move_camera(up_x=0, up_y=1, up_z=0)]
  enableRotate      = True   (was set to False)
  mouseButtons.LEFT = 0      (was set to 2)
```

I don't think this should hold the PR up — it's much rarer than the every-move case you've fixed, and the fix here is a strict improvement either way. Worth saying that `create_controls()` makes the narrower version easy to add later: refreshing OrbitControls' cached `_quat`/`_quatInverse` instead of disposing would need a per-class branch there and nothing else, since TrackballControls reads `camera.up` live and doesn't cache it at all.

Maintainer's call whether that's this PR or a follow-up.

</details>

<details>
<summary><b>CI has never run on this PR</b></summary>

Both pushes are sitting at `action_required` — the first-time-contributor gate, so no workflow has executed on either head:

```
$ gh run list --repo zauberzeug/nicegui --branch fix-ui.scene.move_camera
completed  action_required  ...  CI Gate  pull_request  31107637862  0s  2026-08-06T13:47:58Z
completed  action_required  ...  CI Gate  pull_request  31080526355  0s  2026-08-06T07:20:15Z
```

Worth someone clicking approve before merge. Everything above is macOS + headless Chrome at the same 600×600 the CI fixture uses, but it's still my machine, and `..._keeps_trackball_controls_after_rotating` is the first test in this repo to drive a real mouse drag through `ActionChains`, so CI is genuinely new ground for it rather than a formality.

I tried to pre-empt that in a Linux container and could not get a usable signal — Debian's arm64 Chromium fails the entire pre-existing scene suite there on WebGL console errors, so it says nothing about this PR either way. Flagging the gap rather than papering over it.

</details>

falkoschindler (approved):
Thanks, Patrick — the `controls_up` approach is the right one, and it also handles interrupted tweens correctly, which comparing against the live `camera.up` would not.

While reviewing I found two things that used to happen as a side effect of the unconditional rebuild, so I pushed a commit on top of your branch rather than sending you back and forth:

- **TrackballControls' screen rect** is captured once in its constructor (`handleResize()`), and the rebuild on every `move_camera()` — including the init replay after `resize()` — used to refresh it. With the rebuild gone, a Trackball scene wider than the default 400 px was mapping pointer coordinates against the stale rect. `resize()` now calls `controls.handleResize?.()`.
- **The race from your motivation was only closed for moves with an unchanged up vector.** Scenes configured with a non-default up (`scene.move_camera(up_y=1, up_z=0)` in the page function, replayed from `_handle_init`) still rebuilt one frame after `initialized()` resolved, so `await scene.initialized(); ui.run_javascript(...)` lost the configuration again. Zero-duration tweens now settle synchronously (`camera_tween.update()` right after `.start()`), which also makes every `duration=0` move observable immediately.

Smaller things in the same commit: `create_controls(up)` takes the vector it is built for, so the residue copy in `onComplete` is gone; the `=== null` fallbacks were dead (`scene.py` always sends resolved values) and are removed in both `scene.js` and `scene_view.js`; `move_camera()` is a no-op when the WebGL renderer could not be created; the Trackball test waits for the canvas to be displayed before dragging. Both new scenarios have tests that fail on your last commit and pass now.

I've also updated the PR description to match the final implementation. Approving. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20from%20a%20multi-agent%20review%20run%2C%20verified%20and%20fixed%20by%20the%20model%3B%20scope%20and%20decisions%20from%20the%20maintainer.&a=claude-code "claude-fable-5: Findings from a multi-agent review run, verified and fixed by the model; scope and decisions from the maintainer.")

PR #6243, opened 2026-08-06
Update docstring parameters for light and decay values

python-and-novella:
### Motivation

Improve clarity and correctness of API docstrings so that developers reading the code or generated docs see accurate parameter names and default values. This reduces confusion and helps keep docs consistent with intended usage.

### Implementation

Change the default value of parameters in docstring to actual default value

### Progress

### DIFF
--- a/nicegui/elements/scene/scene_object3d.py
+++ b/nicegui/elements/scene/scene_object3d.py
@@ -102,7 +102,7 @@ class Object3D:
 
         :param color: CSS color string (default: '#ffffff')
         :param opacity: opacity between 0.0 and 1.0 (default: 1.0)
-        :param side: 'front', 'back', or 'double' (default: 'front')
+        :param side: 'front', 'back', or 'both' (default: 'front')
         """
         if self.color != color or self.opacity != opacity or self.side_ != side or not self.material_is_set:
             self.color = color
--- a/nicegui/elements/scene/scene_objects.py
+++ b/nicegui/elements/scene/scene_objects.py
@@ -311,7 +311,7 @@ class SpotLight(Object3D):
         :param distance: maximum distance of light (default: 0.0)
         :param angle: maximum angle of light (default: π/3)
         :param penumbra: penumbra (default: 0.0)
-        :param decay: decay (default: 2.0)
+        :param decay: decay (default: 1.0)
         """
         super().__init__('spot_light', color, intensity, distance, angle, penumbra, decay)
 
--- a/nicegui/elements/scene/scene_objects.py
+++ b/nicegui/elements/scene/scene_objects.py
@@ -309,7 +309,7 @@ class SpotLight(Object3D):
         :param color: CSS color string (default: '#ffffff')
         :param intensity: light intensity (default: 1.0)
         :param distance: maximum distance of light (default: 0.0)
-        :param angle: maximum angle of light (default: π/2)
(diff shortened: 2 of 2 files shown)


evnchn (approved):

**Approving** — all three docstring edits match the actual signatures, checked line by line against `main` (1e3b0214).

One follow-up that is explicitly **not a blocker here**: the `decay: 2.0 → 1.0` edit correctly documents the code, but the *code* default may itself be worth revisiting. three.js has defaulted `SpotLight`'s `decay` to `2` since r147; NiceGUI passes `1.0` explicitly, so it overrides that — spot lights fall off as 1/r rather than 1/r².

@falkoschindler — your call, and I don't think it's obvious:

1. **4.0**, as a breaking change — existing spot-lit scenes get visibly dimmer with distance.
2. **Any 3.x**, as a bugfix — aligning with three.js's current physically-correct default.
3. **Leave it** — `1.0` is intentional, and this PR has now made the docs honest about it.

Either of the first two wants its own PR; a runtime behaviour change doesn't belong in a docstring fix.

<details>
<summary><b>Verification of the three edits (each checked against the signature)</b></summary>

| Docstring edit | Actual signature | Verdict |
|---|---|---|
| `side: 'double'` → `'both'` | `side: Literal['front', 'back', 'both'] = 'front'` — `scene_object3d.py:99` | ✅ correct |
| `angle: π/2` → `π/3` | `angle: float = math.pi / 3` — `scene_objects.py:300` | ✅ correct |
| `decay: 2.0` → `1.0` | `decay: float = 1.0` — `scene_objects.py:302` | ✅ correct *as a docs-match-code fix* |

CI on this PR is green across pre-commit, mypy, pylint, quick-test 3.10 and the Windows storage job.

</details>

<details>
<summary><b>Evidence on the decay question, including what argues against changing it</b></summary>

**Why it looks like a real divergence:**

- Bundled three.js is `^0.180.0` (`nicegui/elements/scene/package.json`); its `SpotLight` constructor minifies to `constructor(t, e, n = 0, i = Math.PI/3, r = 0, s = 2)`, i.e. `decay` defaults to **2**. three.js made that change in r147; the separate r155 release is the one that flipped `useLegacyLights` to `false`.
- `scene.js:314` does `new THREE.SpotLight(...args)` with NiceGUI's values spread in, so the Python default *overrides* the three.js one — someone copying a three.js SpotLight example gets different falloff.
- Commit 1a2f5263 ("lighting corrections for new three.js version") is where `decay (default: 2.0)` came from, in the same commit that added the `* Math.PI` intensity compensation to `AmbientLight` / `DirectionalLight`.

**Why option 3 is genuinely live — the counter-evidence:**

- `decay: float = 1.0` has been the Python default since `spot_light` was first implemented in 157e9767 (April 2022), which was *before* three.js changed its own default. So NiceGUI's `1.0` started out as a faithful mirror of three.js-of-the-day and simply never got revisited. It was never deliberately set to `2.0` and then regressed.
- 1a2f5263 also changed `angle: π/3` → `π/2` in the same two-line docstring edit — and that one was flatly wrong, since the signature has been `math.pi / 3` throughout. That commit touched no Python defaults at all. So "the docstring recorded deliberate intent" is a weaker reading than it first appears; it may simply have been two doc errors, one of which happened to land on the three.js value.

Which is why this is a question for you rather than a claim from me.

</details>

PR #6245, opened 2026-08-06
Make `AxesHelper`'s default color be same as docstring

python-and-novella:
### Motivation

Make `AxesHelper`'s default color be same as docstring

### Implementation

Add `self.color = None` after super's `__init__`, change the color to `None`

### Progress

Demo code:
```python
from nicegui import ui
from nicegui.elements.scene.scene_objects import AxesHelper


def index():
    scene = ui.scene().classes(
        'w-64 h-64'
    )
    with scene:
        AxesHelper()

ui.run(
    root=index,
)
```
image

### DIFF
--- a/nicegui/elements/scene/scene_objects.py
+++ b/nicegui/elements/scene/scene_objects.py
@@ -361,3 +361,4 @@ class AxesHelper(Object3D):
         :param length: length of the the axes (default: 1.0)
         """
         super().__init__('axes_helper', length)
+        self.color = None
--- a/nicegui/elements/scene/scene_objects.py
+++ b/nicegui/elements/scene/scene_objects.py
@@ -361,4 +361,4 @@ class AxesHelper(Object3D):
         :param length: length of the the axes (default: 1.0)
         """
         super().__init__('axes_helper', length)
-        self.color = None
+        self.material(color=None)


evnchn (dismissed):

**Approving** — I reproduced the bug and confirmed the fix by running both builds side by side, not just by reading the diff.

The root cause is that `init_objects` calls `material(id, color, opacity, side)` on page load for every non-GLTF object that has a material (`scene.js:610`). With the inherited default `color = '#ffffff'`, that sets `vertexColors = false` and flattens `AxesHelper`'s built-in red/green/blue to white — so the docstring's "The X axis is red / Y green / Z blue" was simply not true on a fresh page load. Setting `self.color = None` makes `vertexColors` come out `true`, which is exactly the mechanism `PointCloud` already relies on.

| `AxesHelper` on page load | `material.vertexColors` | geometry has color attribute | renders as |
|---|---|---|---|
| `main` (1e3b0214) | `false` | `true` | flat white |
| this PR (e004f22b) | `true` | `true` | red / green / blue |

**One suggestion, not a blocker:** this is a user-visible public-API behaviour with no test, and `test_gltf` is already almost exactly the right shape (it was added for the sibling bug where GLTF models lost their own materials). I wrote the regression test and checked that it actually guards — it fails on `main` and passes here. Entirely yours to take or leave — I can open it as a follow-up test PR instead if that's easier.

<details>
<summary><b>How the before/after table was measured</b></summary>

Two NiceGUI servers run simultaneously from two separate checkouts — `main` on port 8931, this PR's head on 8932 — each confirming its import path at startup, so there is no chance of one shadowing the other:

```
USING NICEGUI FROM: /Users/evnchn/nicegui/nicegui/ui.py                      # main
USING NICEGUI FROM: .../wt6245/nicegui/ui.py                                 # PR 6245
```

Both served the same page (`scene.axes_helper()` inside `ui.scene()`), then the live three.js material was read out of the browser:

```js
getElement(5).scene.traverse((o) => {
  if (o.material && o.type === 'AxesHelper') { /* vertexColors, color, geometry.attributes.color */ }
});
```

Worth noting what this also rules out: the scene's own `GridHelper` reported `vertexColors: true` on **both** builds. It is created directly by `scene.js` rather than as an `Object3D`, so `init_objects` never touches its material. Across the built-in classes in `scene_objects.py`, `AxesHelper` is the only one carrying intrinsic per-vertex colors that it doesn't already pass through `material()` — `PointCloud` handles its own case — so the fix doesn't need to generalise. (That's a statement about the built-ins, not about arbitrary user-supplied STL/GLTF assets.)

</details>

<details>
<summary><b>The regression test, and proof that it guards</b></summary>

```python
def test_axes_helper_keeps_vertex_colors(screen: Screen):
    """AxesHelper must keep its per-vertex red/green/blue colors, as its docstring promises."""
    scene = None
    axes = None

    @ui.page('/')
    def page():
        nonlocal scene, axes
        with ui.scene() as scene:
            axes = scene.axes_helper()

    screen.open('/')
    screen.wait_for(lambda: axes is not None and screen.selenium.execute_script(
        f'return !!window.getElement && !!getElement({scene.id})?.objects?.get("{axes.id}")'
    ))
    assert screen.selenium.execute_script(
        f'return getElement({scene.id}).objects.get("{axes.id}").material.vertexColors'
    ) is True
```

A green test proves nothing on its own, so I ran it both ways on this branch:

```
# with self.color = None (this PR)
1 passed in 4.53s

# with that one line removed
E       assert False is True
E        +  where False = execute_script('return getElement(4).objects.get("...").material.vertexColors')
1 failed in 1.87s
```

Run with `-rs`; the test was not skipped on macOS.

</details>

<details>
<summary><b>On <code>self.color = None</code> vs <code>self.material(color=None)</code></b></summary>

`PointCloud` solves the identical problem one screen up, but via `self.material(color=None)`. Both are correct here, and I checked the difference rather than assuming:

- `self.material(...)` additionally sets `material_is_set = True` and sends a `material` message at construction time. On the init-replay path that flag is read only at `scene.js:610`, as `type != "gltf" || material_is_set` — and `axes_helper` is not GLTF, so it changes nothing here. (Python reads it too, as the "first `material()` call must always send" sentinel in `Object3D.material()`; that stays correct either way.)
- Objects created *after* the client has connected go through `_create()`, which never sends material at all, so the axes keep their native colors on that path under either approach.

So this PR's version is the more minimal of the two — one attribute, no redundant round-trip. Flagging only in case you'd rather the two vertex-colored objects read the same way.

</details>

python-and-novella:
@evnchn I have no idea about `self.color = None` vs `self.material(color=None)`. I noticed that, because I'm writing chinese doc of nicegui, and i found it didn't work as docstring says.
If it is a better solution, I accept it.

falkoschindler:
Thanks for the fix, @python-and-novella — and for noticing the mismatch while writing the Chinese docs; that is exactly how such regressions get caught.

Closing this without merging: the underlying bug has been fixed as a side effect of the scene refactor in #6085. `AxesHelper` now keeps its built-in red/green/blue on page load (see @evnchn's verification in #6246), and the fix is already released in **v3.16.0**. So this branch would only conflict with code that no longer exists.

Sorry that this sat long enough to be overtaken by another change — your diagnosis was correct, and if you spot further gaps between docstrings and behavior while working on the docs, please keep filing them.

PR #6249, opened 2026-08-06
Skip the page re-render when a global event is registered late

### LINKED ISSUE #6248: Page flickers and loses client-side state when `ui.on()` is called after the first render
### First Check

### Example Code

```python
from nicegui import ui

@ui.page('/')
def main():
    ui.html('<input placeholder="type here">', sanitize=False)   # plain HTML: the server never sees this value
    ui.button('register a listener', on_click=lambda: ui.on('some_event', lambda e: None))

ui.run()
```

## Summary

Calling `ui.on()` on a page that has already been rendered causes NiceGUI to destroy and re-create every element on that page. It reports this itself, in the browser console:

```
Event listeners changed after initial definition. Re-rendering affected elements.
```
## Why this matters

I saw #5439 and understand the design decision, but this has led to some of our customers complaining "the page flickers a lot when I open a record". This is in a form-heavy CRUD app. A record page builds its fields one at a time with DB reads in between, and a few of the field types register a client-side event while they're being built: chunked file upload, an AG Grid server-side datasource, a rich text editor, a map widget.

Because the outbox flushes on every `await`, those registrations don't arrive together. Each one is its own update message, and each one rebuilds the page. So opening a record flashes several times before it settles. 

The flicker is just the visible part. Every rebuild also drops AG Grid scroll position and column state, kicks focus out of whatever field you were in, and loses anything typed that hadn't synced yet. Nothing throws, and there's nothing in the network tab or a traceback pointing at `ui.on()`, so
it looks like a rendering bug. We spent a while fixing the wrong things in our own code such as the loading spinner's timing, then batching the render so the fields would arrive in one go, etc before we worked out that it was the event registrations doing it.

## What worked
What finally worked was making sure nothing new ever reaches element 0 after the first paint: we register a single listener up front, during the page function, and route every component's events through it by writing handlers into a dict instead of calling `ui.on()`. That works, but it means none of those components can use `ui.on()` any more, and we now maintain our own dispatch layer alongside NiceGUI's.

### NiceGUI Version

3.15

### Python Version

3.13.7

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #6248: calling `ui.on()` on a page that has already rendered destroys and re-creates **every element on the page** — flicker, lost focus, lost AG Grid scroll/column state, and lost input that had not synced yet.

`ui.on()` registers its listener on `context.client.layout`, and the layout is the first element constructed for every client — element **0**, the root of the tree. The late-registration guard reacts to any previously-unseen `listener_id` by dropping that element so Vue re-creates it. For element 0 that is the whole page.

<details>
<summary>Reproduction of the state loss (the issue's MRE plus an unrelated label)</summary>

```python
from nicegui import ui

@ui.page('/')
def main():
    ui.html('<input placeholder="type here">', sanitize=False)  # plain HTML: the server never sees this value
    ui.label('LABEL')
    ui.button('register a listener', on_click=lambda: ui.on('some_event', lambda e: None))

ui.run()
```

DOM nodes tagged before the click, re-read after:

| | before | after one `ui.on()` |
|---|---|---|
| raw `<input>` value | `TYPED-BY-USER` | `""` |
| raw `<input>` node identity | original | replaced |
| unrelated `ui.label` node identity | original | replaced |

</details>

### Implementation

Skip the delete-and-re-create for element 0 only.

The root does not need it: the regular update path further down the same handler replaces `this.elements[0]`, so the render function reruns and the root vnode receives its refreshed event props without a remount. That covers both dispatch routes — `emitEvent()` → `$emit`, and real DOM events reaching the root through Vue's fallthrough attrs.

The guard stays exactly as it was for every other element, where it is genuinely load-bearing.

<details>
<summary>Why the guard has to stay for non-root elements</summary>

Disabling the delete for *all* ids, verified in a live browser:

| late registration | without the guard |
|---|---|
| `label.on('click')` (plain div) | fires |
| `input.on('keydown.b')` (Quasar `q-input`) | **silently does not fire** |

That second row is the existing `test_late_event_registration`, which still passes unchanged.

</details>

<details>

`tests/test_events.py::test_late_global_event_registration` asserts both halves — that an unsynced raw `<input>` keeps its value across the late registration, and that the late handler still receives the event.

Seen to fail against the unfixed tree before the fix existed:

```
>       assert screen.selenium.execute_script('return document.getElementById("raw").value') == 'kept', \
            'the page must not be re-rendered (see #6248)'
E       AssertionError: the page must not be re-rendered (see #6248)
E       assert '' == 'kept'
```

Local gates:

```
pre-commit run --files nicegui/static/nicegui.js tests/test_events.py   all Passed
mypy ./nicegui                                    Success: no issues found in 245 source files
pylint ./nicegui                                  10.00/10
pytest tests/test_events.py -rs                   14 passed, 0 skipped
pytest tests/test_keyboard.py tests/test_sub_pages.py   55 passed
```

Full local suite: 1003 passed. The `test_dark_mode` failures reproduce identically with `nicegui/static/nicegui.js` reverted to `origin/main`, and are caused by this machine running macOS in Dark appearance, so the "auto" colour-scheme cases resolve to dark.

</details>

<details>
<summary>Note on the warning message</summary>

Element 0 no longer emits `Event listeners changed after initial definition. Re-rendering affected elements.`, since after this change nothing is re-rendered there and the message would be inaccurate. Late registrations on every other element still warn.

</details>

### Progress

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -504,6 +504,7 @@ function createApp(elements, options) {
           for (const [id, element] of Object.entries(msg)) {
             if (element === null) continue;
             if (!(id in this.elements)) continue;
+            if (id === "0") continue; // the layout's event props are refreshed without remounting (see #6248)
             const oldListenerIds = new Set((this.elements[id]?.events || []).map((ev) => ev.listener_id));
             if (element.events?.some((e) => !oldListenerIds.has(e.listener_id))) {
               delete this.elements[id];
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -294,3 +294,23 @@ async def test_late_event_registration(screen: Screen):
     assert 'Event listeners changed after initial definition. Re-rendering affected elements.' in screen.render_js_logs()
     screen.assert_py_logger('WARNING',
                             'Event listeners changed after initial definition. Re-rendering affected elements.')
+
+
+def test_late_global_event_registration(screen: Screen):
+    events = []
+
+    @ui.page('/')
+    def page():
+        ui.html('<input id="raw">', sanitize=False)
+        ui.button('Register', on_click=lambda: ui.on('some_event', lambda: events.append('fired')))
+        ui.button('Fire', on_click=lambda: ui.run_javascript('emitEvent("some_event")'))
+
+    screen.open('/')
+    screen.selenium.execute_script('document.getElementById("raw").value = "kept"')
+    screen.click('Register')
+    screen.wait(0.5)
+    assert screen.selenium.execute_script('return document.getElementById("raw").value') == 'kept', \
+        'the page must not be re-rendered (see #6248)'
+    screen.click('Fire')
+    screen.wait(0.5)
+    assert events == ['fired']


amks1:
Hi guys, 
Hope this can be merged soon. 🙂

evnchn:
@falkoschindler it is net +1-0 for logic, and someone is asking for it, so that's why I undrafted

falkoschindler (approved):
Thanks, @evnchn! Skipping element 0 is safe: the remount from #5439 is only needed for components like `QField` that capture their listeners at setup, while `q-layout` gets its new `on…` props through the regular patch, and `emitEvent` reads the handlers from the current vnode. I also checked that a late `client.layout.on('click', ...)` arrives without re-rendering the page. I pushed a small tweak to the test: it calls `emitEvent` directly and waits for the event with `wait_for` instead of a fixed sleep. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Analysis%20and%20wording%20from%20the%20model%3B%20the%20author%20directed%20the%20extra%20check%20and%20adjusted%20the%20test.&a=claude-code "claude-opus-5: Analysis and wording from the model; the author directed the extra check and adjusted the test.")

PR #6250, opened 2026-08-06
Fix task leak by resolving awaited dialogs when they are deleted

denniswittich:
### Motivation

An event handler that awaits a dialog leaks its task when the client is deleted while the dialog is open.

Minimal example — the label counts the handler tasks that are still pending:

```python
from nicegui import background_tasks, ui


@ui.page('/')
def page():
    with ui.dialog() as dialog, ui.card():
        ui.label('Are you sure?')
        ui.button('Yes', on_click=lambda: dialog.submit(True))

    async def show() -> None:
        result = await dialog
        ui.notify(f'Result: {result}')

    ui.button('Open', on_click=show)
    ui.label().bind_text_from(background_tasks, 'running_tasks',
                              backward=lambda tasks: f'{sum(1 for t in tasks if "show" in t.get_name())} pending "show" handlers')


ui.run()
```

Steps to reproduce:

1. Open the page and click "Open" — the dialog opens and the label shows `1 pending "show" handlers` while the handler waits on `await dialog`.
2. Close the tab and wait a few seconds past the reconnect timeout, so the client is deleted.
3. Open the page again: the label still shows `1 pending "show" handlers`, even though the dialog's client is gone and no dialog is open.

Each repetition of these steps increases the counter by one; it never goes down.
`Client.delete()` removes all elements, but nothing sets the dialog's `submitted` event and nothing cancels the handler task.
The task waits forever on an event that no one will ever set.
It stays in `background_tasks.running_tasks` until server shutdown and keeps everything alive that its coroutine frame references — often the whole page object graph.
This is easy to hit in practice because a dialog opened via `await` is often the main working view of a page, and users end their session by simply closing the tab.
On a long-running server this accumulates hundreds of pending tasks (and their memory) over time.

This is the same class of leak as #5598, which fixed it for `client.connected()` waits; awaited dialogs were not covered.

### Implementation

- `Dialog._handle_delete` now resolves pending awaits with `None` — the same result as dismissing the dialog — so waiting tasks finish instead of waiting forever.
  This covers both client deletion and explicit deletion of the dialog (or a parent).
  In the example above, the counter drops back to `0` as soon as the client is deleted.
- The `submitted` event is set unconditionally on deletion, so a task that touches the public `submitted` property only after deletion resumes as well instead of waiting on a fresh event forever.
  (The event was already set on dismissal without submission, so this extends the existing "the dialog is done" semantics rather than changing them.)
- After resuming, `__await__` skips `close()` if the dialog has been deleted, so the resumed await doesn't trigger the "element has been deleted but is still being used" warning for something that isn't a user bug.
- A resumed handler that still interacts with the deleted client (e.g. via `ui.notify`) hits the pre-existing "Client has been deleted but is still being used" warning, once per process.
  This is intentional: the message goes nowhere, which is worth surfacing.
- Awaiting an already-deleted dialog resolves immediately with `None`.
  If the dialog was explicitly deleted while its client is still alive, the established use-after-delete warning (#3028) is emitted — the same policy `update()` follows — instead of failing silently.
- `__await__` releases the submitted result after reading it — like `close()` does in the normal path — so a deleted dialog does not keep a submitted result alive.

The new pytest fails without the fix (verified) and passes with it.

### Progress

- [x] Documentation has been added/updated (version notes in the `ui.dialog` and `Dialog.submitted` docstrings).

### DIFF
--- a/nicegui/elements/dialog.py
+++ b/nicegui/elements/dialog.py
@@ -65,7 +65,7 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
         self.open()
         yield from self.submitted.wait().__await__()  # pylint: disable=no-member
         result = self._result
-        if not self.is_deleted:
+        if not self.is_deleted:  # closing a deleted dialog would warn about using a deleted element
             self.close()
         return result
 
--- a/nicegui/elements/dialog.py
+++ b/nicegui/elements/dialog.py
@@ -82,7 +82,7 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
 
     def _handle_delete(self) -> None:
         # resolve pending awaits so their tasks don't wait forever, e.g. when the client is deleted after a disconnect
+        # (keep self._result untouched: a result submitted just before deletion should still reach the awaiting task)
         if self._submitted is not None:
-            self._result = None
             self._submitted.set()
         super()._handle_delete()
--- a/nicegui/elements/dialog.py
+++ b/nicegui/elements/dialog.py
@@ -25,6 +25,9 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
         That means it is not removed when closed, but only hidden.
         You should either create it only once and then reuse it, or remove it with `.clear()` after dismissal.
 
+        *Updated in version 3.16.0: Awaiting a dialog resolves with ``None`` when the dialog is deleted,
+        e.g. because the client disconnected.*
+
         :param value: whether the dialog should be opened on creation (default: `False`)
         """
         with context.client.layout:
@@ -55,12 +58,15 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
         return super().toggle()
 
     def __await__(self):
+        if self.is_deleted:
+            return None  # a deleted dialog can never be submitted or closed, so we resolve immediately
         self._result = None
         self.submitted.clear()
         self.open()
         yield from self.submitted.wait().__await__()  # pylint: disable=no-member
         result = self._result
-        self.close()
+        if not self.is_deleted:
+            self.close()
         return result
 
     def submit(self, result: Any) -> None:
@@ -73,3 +79,10 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
         if not self.value:
             self._result = None
             self.submitted.set()
+
+    def _handle_delete(self) -> None:
+        # resolve pending awaits so their tasks don't wait forever, e.g. when the client is deleted after a disconnect
+        if self._submitted is not None:
+            self._result = None
+            self._submitted.set()
+        super()._handle_delete()
--- a/nicegui/elements/dialog.py
+++ b/nicegui/elements/dialog.py
@@ -45,7 +45,10 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
 
     @property
     def submitted(self) -> asyncio.Event:
-        """An event that is set when the dialog is submitted."""
+        """An event that is set when the dialog is submitted.
+
+        *Updated in version 3.16.0: The event is also set when the dialog is deleted.*
+        """
         if self._submitted is None:
             self._submitted = asyncio.Event()
         return self._submitted
@@ -58,13 +61,14 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
         return super().toggle()
 
     def __await__(self):
-        if self.is_deleted:
-            return None  # a deleted dialog can never be submitted or closed, so we resolve immediately
+        if not self._is_safe_to_interact():
+            return None  # the dialog cannot be submitted anymore, so we resolve immediately instead of waiting forever
         self._result = None
         self.submitted.clear()
         self.open()
         yield from self.submitted.wait().__await__()  # pylint: disable=no-member
         result = self._result
+        self._result = None  # release the result so a deleted dialog does not keep it alive (close() would do the same)
         if not self.is_deleted:  # closing a deleted dialog would warn about using a deleted element
             self.close()
         return result
@@ -83,6 +87,5 @@ class Dialog(OpenableElement, NoImplicitAwait, component='dialog.js'):
     def _handle_delete(self) -> None:
         # resolve pending awaits so their tasks don't wait forever, e.g. when the client is deleted after a disconnect
         # (keep self._result untouched: a result submitted just before deletion should still reach the awaiting task)
-        if self._submitted is not None:
-            self._submitted.set()
+        self.submitted.set()
         super()._handle_delete()
--- a/tests/test_dialog.py
+++ b/tests/test_dialog.py
@@ -129,6 +129,31 @@ async def test_awaited_dialog_resolves_when_client_is_deleted(user: User):
     assert await dialog is None  # awaiting a deleted dialog resolves immediately
 
 
+async def test_submit_and_delete_dialog_in_same_handler(user: User):
+    """A result submitted right before the dialog is deleted must still reach the awaiting task."""
+    results = []
+
+    @ui.page('/')
+    def page():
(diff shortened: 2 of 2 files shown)


evnchn (dismissed):

**TL;DR: Approve** — the fix is right, and resolving with `None` is the right call over raising.

One line I'd change before merge, non-blocking: `_handle_delete()`'s `self._result = None` clobbers an already-submitted result when a dialog is submitted *and* deleted in the same synchronous block. Dropping that one line keeps everything working (`__await__` already resets `_result` at the start of every await).

On "should it raise instead?" — I measured all three options. Regular exception: no. `CancelledError`: the only serious contender, but it silently changes control flow for every existing dialog handler, so it's a 4.0-shaped change, not a bugfix.

<details><summary><b>The one line: submit-then-delete loses the result</b></summary>

`_handle_delete()` sets `_result = None` unconditionally. Since the awaiting task only resumes on the next loop iteration, a handler that submits and then deletes in the same synchronous turn has its answer overwritten first:

```python
def submit_and_delete() -> None:
    dialog.submit('yes')
    dialog.delete()      # e.g. a parent container torn down by the same handler
```

Measured three-way with a `User`-fixture repro:

| branch | `await dialog` returns |
|---|---|
| `main` (base) | `'yes'` |
| this PR | `None` ← |
| this PR minus `self._result = None` | `'yes'` |

```python
def _handle_delete(self) -> None:
    # resolve pending awaits so their tasks don't wait forever, ...
    if self._submitted is not None:
        self._submitted.set()          # no _result = None
    super()._handle_delete()
```

Safe because `__await__` already does `self._result = None` before every non-deleted await (dialog.py:63), and the already-deleted path returns before reading `_result` at all. Your new test plus my repro both pass with the line removed (15 passed). Codex was asked to refute this and could not — including the stale-`_result` paths (deleted-while-not-awaited, close-then-delete, submit-after-delete).

</details>

<details><summary><b>Why raising is worse — measured, all three variants</b></summary>

Same scenario each time: handler awaiting an open dialog, `client.delete()` fires.

| variant in `__await__` | what the app sees |
|---|---|
| **resolve `None`** (this PR) | handler resumes with `None`; nothing logged |
| **`raise RuntimeError(...)`** | **ERROR-level traceback logged on every tab close**, routed through `core.app.handle_exception` — so user `on_exception` handlers and Sentry fire on ordinary user behaviour. Also trips NiceGUI's own `user_plugin` logger assertion, i.e. the framework's test harness treats it as a failure |
| **`raise asyncio.CancelledError(...)`** | silent — `background_tasks` suppresses it, the task ends cancelled, `finally` / `async with` cleanup still runs |

So the regular-exception route is clearly out. `CancelledError` is the interesting one, and it has a genuine correctness argument in its favour: resuming with `None` *fabricates an answer*, and any post-await code keeps running against a dead client. UI calls are silently no-oped by `Element._is_safe_to_interact`, but `save_annotation(result)` is not — it runs with `result=None` as though the user had dismissed. Cancelling unwinds the handler instead.

Three reasons I still land on `None`:

1. **`None` already means "dismissed."** ESC / click-outside goes through `_handle_value_change` and sets `_result = None`. From user code's side, "the client is gone" and "the user dismissed it" are the same event: the dialog will never be answered. Existing `if result:` handlers are already correct for it.
2. **House precedent.** PR 5598 solved this exact class for `client.connected()` by resolving silently on client deletion rather than raising; `client.disconnected()` resolves via `_deleted_event.set()` too.
3. **Blast radius.** `CancelledError` would change control flow for every app that awaits a dialog, with no deprecation path. That's a 4.0 conversation, not a bugfix.

One thing I checked and want to correct from the usual objection: the "don't manually raise `CancelledError` in 3.11+" concern is weaker than it's usually stated. On 3.14 a manual raise gives `task.cancelled() == True` with `cancelling() == 0`, and `TaskGroup` handles it fine — the `uncancel()` bookkeeping hazard is about code that *catches and suppresses* cancellation, not about raising it. So if you ever do want teardown semantics, that's not the blocker; the API-compatibility argument is.

If post-await side effects on a dead client do become a real problem, the cleaner lever is cancelling the *task* at client-teardown level — that cures the whole leak class (any client-scoped await), not just dialogs.

</details>

<details><summary><b>The <code>if not self.is_deleted: close()</code> guard does more than the PR body claims (in its favour)</b></summary>

I initially read the guard as only suppressing the issue-3028 warn-once, and was going to argue it hides a real use-after-free. That was wrong, and worth recording: for the client-deleted case `_is_safe_to_interact` (element.py:424) already returns `False` silently, so the warning was never the main effect. Measured with the guard removed:

| | `dialog.value` after delete | `on_value_change` fired |
|---|---|---|
| guard removed | `False` | `[False]` ← handlers run during teardown |
| guard present (this PR) | `True` | `[]` |

So the guard also stops `close()` from mutating props and invoking value-change callbacks on a torn-down client. That's a better justification than "avoids a spurious warning" and probably worth a word in the code comment. The residue is that a deleted dialog keeps `value=True`; harmless as far as I can tell, since nothing reads it after teardown.

</details>

<details><summary><b>What was run</b></summary>

Worktree off `upstream/fix-dialog-await-leak-on-deletion` @ `9da5664f`, base `HEAD~1`.

```
.venv/bin/pre-commit run --files nicegui/elements/dialog.py tests/test_dialog.py   # all passed
.venv/bin/mypy ./nicegui                                                            # Success: no issues found in 245 source files
.venv/bin/pylint ./nicegui                                                          # 10.00/10
.venv/bin/python -m pytest tests/test_dialog.py -rs                                 # 13 passed, 0 skipped
```

Seen-to-fail on the new test — `git show HEAD~1:nicegui/elements/dialog.py > nicegui/elements/dialog.py`, then:

```
>       assert results == [None]  # the handler resumed instead of waiting forever
E       assert [] == [None]
1 failed in 0.34s
```

Restored → passes. Also verified the fix covers deleting *only* the dialog while the client is alive: on `main` that await hangs forever too, and this PR resolves it.

Nits, take or leave: the new test's `asyncio.sleep(0.1)` works as `sleep(0)` (REVIEW.md prefers fixtures over sleeps), and the `` `None` `` inside the `*…*` version note won't render as code — reST doesn't nest inline markup — though `client.py:145` already has the same wart, so it's pre-existing.

Second lineage: findings were run past Codex (gpt-5.2) as a skeptic. It confirmed the clobber independently, and caught that my read of the `close()` guard was too narrow — the correction above is its finding, re-verified here.

</details>

evnchn (dismissed):
Re-approving after the two follow-up commits — both points addressed cleanly.

Verified locally at 6eb7d122: the new submit-then-delete test fails with the `_result = None` line restored and passes without it, and it also passes on base main, so it is a genuine regression guard rather than a green-only test. Local gates: pre-commit, mypy, pylint 10.00/10, `tests/test_dialog.py` 14 passed / 0 skipped. CI green.

One non-blocking nit, take or leave: the new comment on the `close()` guard says closing a deleted dialog "would warn about using a deleted element". That holds when the dialog alone is deleted, but not for the client-deletion case in the Motivation — there `_is_safe_to_interact` returns `False` silently, so no warning fires. Measured with the guard removed on that path: `value=False`, `on_value_change=[False]`, no warning. So its real job there is keeping `close()` from firing value-change handlers during teardown.

falkoschindler (approved):
Nice fix!
After a deep review I pushed a few refinements in 379da2e0e:

- `__await__` now guards with `_is_safe_to_interact()` instead of a bare `is_deleted` check: awaiting an explicitly deleted dialog under a live client emits the established use-after-delete warning (#3028) instead of silently resolving, and a dialog whose client was already garbage-collected resolves instead of hanging.
- `_handle_delete` sets `submitted` unconditionally, so a task that touches the public `submitted` property only after deletion resolves as well instead of waiting on a fresh event forever.
- `__await__` releases `_result` after reading it — like `close()` does in the normal path — so a deleted dialog doesn't keep a submitted result alive.

The same leak class in `button.clicked()` and `scene`/`leaflet.initialized()` is tracked in #6277.

PR #6263, opened 2026-08-07
Close the storage file when a backup task is cancelled

evnchn:

### Motivation

Cancelling a storage backup task orphans the open temp file. It surfaces later, at an unrelated moment, as a `ResourceWarning` — and since `pyproject.toml` sets `filterwarnings = ['error']`, that fails whichever test happens to be running when the garbage collector gets to it.

Seen in the merge queue (run 30925209909, py3.10), reported against `tests/test_user_simulation.py::test_storage[<lambda>1]`, which is merely where GC landed:

```
ERROR at teardown of test_storage[<lambda>1]
pytest.PytestUnraisableExceptionWarning: Exception ignored in:
  <_io.FileIO name='/tmp/nicegui-test-storage-…/storage-user-….json.tmp' mode='wb'>
ResourceWarning: unclosed file <_io.TextIOWrapper name='…/storage-user-….json.tmp' mode='w'>
```

**Mechanism.** `backup()` schedules `async_backup()` through `background_tasks.create_lazy`, and the write is:

```python
async with aiofiles.open(tmp_filepath, 'w', encoding=self.encoding) as f:
    await f.write(...)
```

`aiofiles.open()` performs the real `open()` in a worker thread. If the task is cancelled **while `__aenter__` is still in flight**, the `open()` completes in that thread but `__aenter__` never returns — so `async with` never binds, `__aexit__` never runs, and the file object is left for the GC. (Cancellation inside the `async with` *body* is safe; `__aexit__` still runs. Only the entering window leaks.)

### Implementation

Do the write in a worker thread as one atomic open-write-close, so there is no window in which a Python-level file object exists without its closer:

```python
await asyncio.to_thread(tmp_filepath.write_text,
                        dumps(self, str(self.filepath), indent=self.indent), encoding=self.encoding)
```

Cancelling the await no longer aborts the thread mid-way, and `write_text` closes the file itself. Same bytes, same encoding, same atomic-rename that follows. `aiofiles` is still used for the read path, which has no equivalent exposure.

<details>

**Mechanism, isolated** (plain asyncio, no NiceGUI — cancel each variant at six different delays, 30 reps each, then force GC):

```
aiofiles async-with (current code) : 2 ResourceWarning(s)
    unclosed file <_io.TextIOWrapper name='…/storage-user-test.json.tmp' mode='w' encoding='utf-8'>
asyncio.to_thread(write_text)      : 0 ResourceWarning(s)
```

**Through the real `backup()` path** — the new regression test drives `d['key'] = 'value'`, grabs the scheduled task out of `background_tasks.lazy_tasks_running`, and cancels it at four different delays × 20 reps:

```
with fix    : 15/15 passed
without fix : 15/15 failed, minimum 19 of 40 cancellations leaked
```

The test also asserts it actually cancelled a *live* task (`cancelled += task.cancel()`, which returns `False` for an already-finished task). Without that, a fast runner could cancel nothing, leak nothing, and pass forever while exercising none of the code under test.

Same warning shape as the CI failure. The test drives the real scheduling path deliberately — an earlier draft reproduced the write logic inline, which would have proved nothing about `backup()`.

**The delays are chosen, not guessed.** The leak only happens while `aiofiles`' `open()` is in flight — a window a few dozen microseconds wide. Measured yield on unfixed code, 60 cancellations per delay:

| cancel delay | files leaked |
|---|---|
| 20 µs | 60 / 60 |
| 50 µs | 60 / 60 |
| 100 µs | 2 / 60 |
| 200 µs and beyond | 0 / 60 |

An earlier version of this test used 0 / 100 µs / 500 µs / 1 ms — nearly all in the dead zone. It still failed 20/20 on unfixed code, but on as little as **one** leaked file, one unlucky scheduling away from a false pass. Retargeting onto the real window raised the worst-case margin from 1 to 17 while halving the iteration count.

**No regressions:** `tests/test_storage.py` 30 passed. Gates: pre-commit all Passed · mypy clean over 245 source files · pylint 10.00/10.

</details>

<details>
<summary><b>Why this was originally reported as "not reproducible"</b></summary>

My first pass at this failed to reproduce it and I set it aside. That was a methodology error, not a property of the bug: I was trying to reproduce the **end-to-end CI symptom** (run the storage tests repeatedly and wait for GC to land in the wrong place) rather than the **mechanism** (cancel a task during `__aenter__`). The former is rare and timing-dependent; the latter is deterministic and takes 0.2 s to demonstrate.

The general lesson is to drop one level of abstraction before concluding something is unreproducible.

</details>

### Progress

- [x] Documentation is not necessary — internal behaviour, no API change.

### DIFF
--- a/nicegui/persistence/file_persistent_dict.py
+++ b/nicegui/persistence/file_persistent_dict.py
@@ -1,3 +1,4 @@
+import asyncio
 import contextlib
 from pathlib import Path
 
@@ -54,8 +55,10 @@ class FilePersistentDict(PersistentDict):
                 tmp_filepath.unlink(missing_ok=True)
                 await unlink_with_retry_async(self.filepath, missing_ok=True)
                 return
-            async with aiofiles.open(tmp_filepath, 'w', encoding=self.encoding) as f:
-                await f.write(dumps(self, str(self.filepath), indent=self.indent))
+            # write in a worker thread: cancelling an `async with aiofiles.open(...)` while it is
+            # still entering leaves the opened file unclosed, which surfaces later as a ResourceWarning
+            await asyncio.to_thread(tmp_filepath.write_text,
+                                    dumps(self, str(self.filepath), indent=self.indent), encoding=self.encoding)
             with contextlib.suppress(FileNotFoundError):  # a concurrent Storage.clear() may have swept the temp file
                 tmp_filepath.replace(self.filepath)
 
--- a/nicegui/persistence/file_persistent_dict.py
+++ b/nicegui/persistence/file_persistent_dict.py
@@ -2,8 +2,6 @@ import asyncio
 import contextlib
 from pathlib import Path
 
-import aiofiles
-
 from .. import background_tasks, core, json
 from ..helpers import unlink_with_retry, unlink_with_retry_async
 from ..logging import log
@@ -22,8 +20,8 @@ class FilePersistentDict(PersistentDict):
     async def initialize(self) -> None:
         try:
             if self.filepath.exists():
-                async with aiofiles.open(self.filepath, encoding=self.encoding) as f:
-                    data = json.loads(await f.read())
+                # read in a worker thread: see the cancellation note in async_backup below
+                data = json.loads(await asyncio.to_thread(self.filepath.read_text, encoding=self.encoding))
             else:
                 data = {}
             self.update(data)
@@ -55,8 +53,9 @@ class FilePersistentDict(PersistentDict):
                 tmp_filepath.unlink(missing_ok=True)
                 await unlink_with_retry_async(self.filepath, missing_ok=True)
                 return
-            # write in a worker thread: cancelling an `async with aiofiles.open(...)` while it is
-            # still entering leaves the opened file unclosed, which surfaces later as a ResourceWarning
+            # open, write and close in a single worker-thread call: any cancellation point between
+            # opening and closing the file could strand an open handle, surfacing as a ResourceWarning
+            # (not run.io_bound, which would skip the write while the app is stopping)
             await asyncio.to_thread(tmp_filepath.write_text,
                                     dumps(self, str(self.filepath), indent=self.indent), encoding=self.encoding)
             with contextlib.suppress(FileNotFoundError):  # a concurrent Storage.clear() may have swept the temp file
--- a/nicegui/storage.py
+++ b/nicegui/storage.py
@@ -1,4 +1,3 @@
-import contextlib
 import contextvars
 import os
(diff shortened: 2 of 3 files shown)


falkoschindler:
Nice find — and the fix lands at the right depth. I verified the mechanics: aiofiles' `__aenter__` runs `open()` in an executor work item that cannot be cancelled once started, so a cancel mid-entry drops the opened file object unreferenced, and a single `asyncio.to_thread` call removes exactly that window. The test holds up too: it fails 5/5 runs against the old implementation (20 of 40 iterations leak) and passes 10/10 on the fix, with plenty of margin on the `assert cancelled` guard even on fast hardware. Choosing `asyncio.to_thread` over `run.io_bound` is also right here, since the latter skips the callback while the app is stopping, which would defeat `@background_tasks.await_on_shutdown`.

I pushed three small commits on top: `initialize()` now reads in a worker thread as well (same cancellation window, and on the more cancellable per-request path — an ordinary client disconnect cancels it mid-open), `Storage.clear()` retries its temp-file sweep (this fix inverted the old "never wait" rationale: the handle now lives in a worker thread that releases it independently of the event loop), and the test filters its `ResourceWarning`s to storage files so unrelated warnings swept up by `gc.collect()` cannot fail it — plus two comment touch-ups.

Two pre-existing things surfaced during the review. First, `create_lazy` drops the `await_on_shutdown` marker when requeueing a waiting task, which can silently lose the newest storage change at shutdown — fixed separately in #6329. Second, the temp file creation in `upload_files.py` has the same cancellation shape as this bug (a cancel during the `NamedTemporaryFile` await strands the handle and the `delete=False` file), but the window is microscopic compared to an upload's lifetime, so we're leaving it — noted here so it stays findable.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%2C%20verification%2C%20and%20wording%20by%20the%20model%3B%20scope%20and%20severity%20confirmed%20by%20the%20maintainer.&a=claude-code "claude-fable-5: Findings, verification, and wording by the model; scope and severity confirmed by the maintainer.")

falkoschindler (approved):

falkoschindler (approved):
Verified in depth (see the review comment above); the remaining findings are fixed directly on the branch — good to go. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Approval%20note%20phrased%20by%20the%20model%3B%20the%20verification%20behind%20it%20is%20detailed%20in%20the%20review%20comment.&a=claude-code "claude-fable-5: Approval note phrased by the model; the verification behind it is detailed in the review comment.")

Issue #6234, opened 2026-08-03
`ui.on_exception` never sees exceptions raised while a `@ui.refreshable` rebuilds


evnchn:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a security issue.
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar issue and came up empty.

### Example Code

```python
# Save as mre.py next to a pytest.ini containing:
#     [pytest]
#     asyncio_mode = auto
#     main_file =
# then:  pip install nicegui==3.15.0 "pytest>=9,<10" pytest-asyncio && pytest mre.py
#
# The two tests differ by ONE line: the first passes, the second fails.
import asyncio

import pytest
from nicegui import ui
from nicegui.testing import User

pytest_plugins = ['nicegui.testing.user_plugin']


async def test_direct_call(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    @ui.page('/')
    def page():
        ui.on_exception(seen.append)

        @ui.refreshable
        def part(explode: bool = False):
            if explode:
                raise RuntimeError('boom')

        part()
        ui.button('fire', on_click=lambda: part(True))          # direct call

    await user.open('/')
    user.find('fire').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'


async def test_via_refresh(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    @ui.page('/')
    def page():
        ui.on_exception(seen.append)

        @ui.refreshable
        def part(explode: bool = False):
            if explode:
                raise RuntimeError('boom')

        part()
        ui.button('fire', on_click=lambda: part.refresh(True))  # via refresh()

    await user.open('/')
    user.find('fire').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'
```

### Description

*Filed by Claude Code on @evnchn's behalf.*

**TL;DR:**

Calling a `@ui.refreshable` function directly raises into the caller's context, so `ui.on_exception` sees it. Calling `.refresh()` on the same function with the same body does not.

[...]

PR #6265, opened 2026-08-07
Fix visibility of exceptions raised from async `@ui.refreshable` body during rebuild

### LINKED ISSUE #6262: `ui.on_exception` never sees exceptions from an async `@ui.refreshable` body
### First Check

### Example Code

```python
# Save as mre.py next to a pytest.ini containing:
#     [pytest]
#     asyncio_mode = auto
#     main_file =
# then:  pip install nicegui==3.15.0 "pytest>=9,<10" pytest-asyncio && pytest mre.py
#
# Same async @ui.refreshable body in both tests. They differ by ONE line:
# the first passes, the second fails.
import asyncio

import pytest
from nicegui import ui

async def test_async_direct_call(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    @ui.page('/')
    async def page():
        ui.on_exception(seen.append)

        @ui.refreshable
        async def part(explode: bool = False):
            await asyncio.sleep(0)
            if explode:
                raise RuntimeError('boom')

        await part()
        ui.button('fire', on_click=lambda: part(True))          # direct call

    await user.open('/')
    user.find('fire').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'


async def test_async_via_refresh(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    @ui.page('/')
    async def page():
        ui.on_exception(seen.append)

        @ui.refreshable
        async def part(explode: bool = False):
            await asyncio.sleep(0)
            if explode:
                raise RuntimeError('boom')

        await part()
        ui.button('fire', on_click=lambda: part.refresh(True))  # via refresh()

    await user.open('/')
    user.find('fire').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'
```

### Description

**TL;DR:**

Awaiting a `@ui.refreshable` with an `async def` body raises into the caller's context, so `ui.on_exception` sees it. Calling `.refresh()` on the same function with the same body does not.

This is the async counterpart of #6234. That issue's MRE uses a **synchronous** body, and PR #6258 (which fixes it) only catches synchronous raises out of `RefreshableTarget.run` — so this case is untouched by that fix and needs a separate decision about which layer it belongs in.

As with #6234, the exception still reaches `app.on_exception` and the log — only the client-scoped handler is skipped.

<details>
<summary><b>Mechanism</b></summary>

For an `async def` body, `RefreshableTarget.run` does not raise — it returns an awaitable (`nicegui/functions/refreshable.py:42-43` on 3.15.0):

```python
        if helpers.should_await(result):
            return cast(_T, helpers.await_with_context(result, self.container))
```

so `_execute_refresh` collects it into `awaitables` and hands it to `background_tasks.create_or_defer(asyncio.gather(*awaitables), ...)`. The body's exception therefore surfaces much later, inside that gather.

By then `await_with_context` (`nicegui/helpers/functions.py:57-60`) has already unwound its `with`:

```python
async def await_with_context(awaitable, context):
    """Await an awaitable within a context manager."""
    with context:
        return await awaitable
```

The exception propagates *out* of the `with`, so the slot stack is empty again when `background_tasks._handle_exceptions` calls `app.handle_exception`, and the client-scoped guard (`nicegui/app/app.py:178`) is skipped:

```python
    context.client.handle_exception(exception)
```

Same end state as #6234 — a background task with no client context — but reached one layer further down, which is why a fix at the `_execute_refresh` `try/except` cannot see it.

</details>

<details>
<summary><b>Actual output (nicegui 3.15.0 from PyPI, clean venv)</b></summary>

```
$ pytest mre.py -q
FAILED mre.py::test_async_via_refresh - AssertionError: ui.on_exception saw 0...
1 failed, 1 passed in 0.58s
```

The traceback that reaches the log confirms the path:

```
    task.result()
  File ".../nicegui/background_tasks.py", line 131, in wrapper
    return await awaitable
  File ".../nicegui/helpers/functions.py", line 60, in await_with_context
    return await awaitable
  File ".../mre.py", line 52, in part
    raise RuntimeError('boom')
RuntimeError: boom
```

Also run against the head of PR #6258 (`843849b0`): same result, **1 failed, 1 passed** — that PR does not reach this case.

</details>

### NiceGUI Version

3.15.0

### Python Version

3.14.2

### Browser

Other

### Operating System

macOS

### Additional Context

Browser-independent — the repro uses `nicegui.testing.User`, so no browser is involved.

Split out from #6234 the same way #6234 was split from #6233, so the three can be worked independently. If you would rather track this inside #6234 as a second case, that works too — say the word and I will close this.

SaadZahem:
### Motivation

Fixed #6262

### Implementation

This issue is similar to #6234 except for the async part.
The fix is similar to #6258 except for where the `try/except` is placed.
`try/except` cannot be placed in `RefreshableTarget.run` as it doesn't _execute_ the async body.
```python
if helpers.should_await(result):
    return cast(_T, helpers.await_with_context(result, self.container))
```
I didn't want to add `try/except` to a helper function either.
Even though it would actually work, it will most definitely cause more issues later on.

My solution is to add an async method to `RefreshableTarget`,
where the functionality of `helpers.await_with_context` is covered by a `try/except` similar to #6258.
This way, the `try/except` block becomes part of the returned awaitable.

I tested using the 2 test cases provided in #6262.
I have added the test case for this issue in another commit to `tests/test_refreshable.py` under a similar descriptive name.

### Progress

### DIFF
--- a/nicegui/functions/refreshable.py
+++ b/nicegui/functions/refreshable.py
@@ -7,7 +7,7 @@ from typing import Any, ClassVar, Concatenate, Generic, TypeVar, cast
 
 from typing_extensions import ParamSpec, Self
 
-from .. import background_tasks, context, helpers
+from .. import background_tasks, helpers
 from ..awaitable_response import AwaitableResponse
 from ..element import Element
 
@@ -28,7 +28,7 @@ class RefreshableTarget:
     locals: list[Any] = field(default_factory=list)
     next_index: int = 0
 
-    def run(self, func: Callable[..., _T]) -> _T:
+    def run(self, func: Callable[..., _T], *, report_exceptions: bool = False) -> _T:
         """Run the function and return the result."""
         RefreshableTarget.current_target = self
         self.next_index = 0
@@ -40,16 +40,17 @@ class RefreshableTarget:
                 result = func(self.instance, *self.args, **self.kwargs)
 
         if helpers.should_await(result):
-            return cast(_T, self._await_with_context(result))
+            if report_exceptions:
+                return cast(_T, self._await_and_report(result))
+            return cast(_T, helpers.await_with_context(result, self.container))
 
         return result
 
-    async def _await_with_context(self, awaitable: Awaitable[_T]) -> _T:
+    async def _await_and_report(self, awaitable: Awaitable[_T]) -> _T:
         try:
             return await helpers.await_with_context(awaitable, self.container)
         except Exception as e:
-            if not context.slot_stack:
-                self.container.client.handle_exception(e)
+            self.container.client.handle_exception(e)
             raise
 
 
@@ -105,7 +106,7 @@ class refreshable(Generic[_P, _T]):
         instance = self.instance
 
         def fire_and_forget() -> None:
-            if awaitables := self._execute_refresh(args, kwargs, instance=instance):
+            if awaitables := self._execute_refresh(args, kwargs, instance=instance, report_exceptions=True):
                 background_tasks.create_or_defer(asyncio.gather(*awaitables), name=f'refresh {self.func.__name__}')
 
         async def wait_for_completion() -> None:
@@ -114,7 +115,8 @@ class refreshable(Generic[_P, _T]):
 
         return AwaitableResponse(fire_and_forget, wait_for_completion)
 
-    def _execute_refresh(self, args: tuple[Any, ...], kwargs: dict[str, Any], *, instance: Any) -> list[Awaitable[Any]]:
+    def _execute_refresh(self, args: tuple[Any, ...], kwargs: dict[str, Any], *,
+                         instance: Any, report_exceptions: bool = False) -> list[Awaitable[Any]]:
         """Execute the refresh and return a list of awaitables for async functions."""
         awaitables: list[Awaitable[Any]] = []
         for target in self.targets:
@@ -124,7 +126,7 @@ class refreshable(Generic[_P, _T]):
             target.args = args or target.args
             target.kwargs.update(kwargs)
             try:
-                result = target.run(self.func)
+                result = target.run(self.func, report_exceptions=report_exceptions)
             except TypeError as e:
                 if 'got multiple values for argument' in str(e):
                     function = str(e).split()[0].split('.')[-1]
--- a/nicegui/functions/refreshable.py
+++ b/nicegui/functions/refreshable.py
@@ -50,7 +50,8 @@ class RefreshableTarget:
         try:
             return await helpers.await_with_context(awaitable, self.container)
         except Exception as e:
-            self.container.client.handle_exception(e)
+            if not self.container.is_deleted:
+                self.container.client.handle_exception(e)
             raise
 
 
--- a/nicegui/functions/refreshable.py
+++ b/nicegui/functions/refreshable.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import Awaitable, Callable
-from contextlib import AbstractContextManager
 from dataclasses import dataclass, field
 from typing import Any, ClassVar, Concatenate, Generic, TypeVar, cast
 
@@ -41,16 +40,16 @@ class RefreshableTarget:
                 result = func(self.instance, *self.args, **self.kwargs)
 
         if helpers.should_await(result):
-            return cast(_T, self._await_with_context(result, self.container))
+            return cast(_T, self._await_with_context(result))
 
         return result
 
-    async def _await_with_context(self, awaitable: Awaitable[_T], container: AbstractContextManager) -> _T:
+    async def _await_with_context(self, awaitable: Awaitable[_T]) -> _T:
         try:
-            return await helpers.await_with_context(awaitable, container)
+            return await helpers.await_with_context(awaitable, self.container)
         except Exception as e:
             if not context.slot_stack:
-                container.client.handle_exception(e)
+                self.container.client.handle_exception(e)
             raise
 
 
--- a/nicegui/functions/refreshable.py
+++ b/nicegui/functions/refreshable.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import Awaitable, Callable
+from contextlib import AbstractContextManager
 from dataclasses import dataclass, field
 from typing import Any, ClassVar, Concatenate, Generic, TypeVar, cast
 
 from typing_extensions import ParamSpec, Self
 
-from .. import background_tasks, helpers
+from .. import background_tasks, context, helpers
 from ..awaitable_response import AwaitableResponse
 from ..element import Element
 
@@ -40,9 +41,17 @@ class RefreshableTarget:
                 result = func(self.instance, *self.args, **self.kwargs)
 
         if helpers.should_await(result):
-            return cast(_T, helpers.await_with_context(result, self.container))
+            return cast(_T, self._await_with_context(result, self.container))
 
         return result
+
+    async def _await_with_context(self, awaitable: Awaitable[_T], container: AbstractContextManager) -> _T:
+        try:
+            return await helpers.await_with_context(awaitable, container)
+        except Exception as e:
+            if not context.slot_stack:
+                container.client.handle_exception(e)
+            raise
 
 
 class RefreshableContainer(Element, component='refreshable.js'):
--- a/tests/test_refreshable.py
+++ b/tests/test_refreshable.py
@@ -310,7 +310,8 @@ def test_awaitable_refresh(screen: Screen):
     assert events == ['update started', 'refresh started', 'refresh failed', 'update finished']
 
 
-async def test_report_exception_async(user: User, caplog: pytest.LogCaptureFixture):
+@pytest.mark.parametrize('awaited', [False, True])
+async def test_report_exception_async(user: User, caplog: pytest.LogCaptureFixture, awaited: bool):
     seen: list[Exception] = []
 
     @ui.page('/')
@@ -323,8 +324,11 @@ async def test_report_exception_async(user: User, caplog: pytest.LogCaptureFixtu
             if explode:
                 raise RuntimeError('boom')
 
+        async def refresh_awaited():
+            await part.refresh(True)
+
         await part()
-        ui.button('fire', on_click=lambda: part.refresh(True))
+        ui.button('fire', on_click=refresh_awaited if awaited else lambda: part.refresh(True))
 
     await user.open('/')
     user.find('fire').click()
--- a/tests/test_refreshable.py
+++ b/tests/test_refreshable.py
@@ -1,7 +1,9 @@
 import asyncio
 
+import pytest
+
 from nicegui import ui
-from nicegui.testing import Screen
+from nicegui.testing import Screen, User
 
 
 def test_refreshable(screen: Screen) -> None:
@@ -306,3 +308,26 @@ def test_awaitable_refresh(screen: Screen):
     screen.click('Try 0')
     screen.should_contain('error handled')
     assert events == ['update started', 'refresh started', 'refresh failed', 'update finished']
+
+
+async def test_report_exception_async(user: User, caplog: pytest.LogCaptureFixture):
+    seen: list[Exception] = []
+
+    @ui.page('/')
+    async def page():
+        ui.on_exception(seen.append)
+
+        @ui.refreshable
+        async def part(explode: bool = False):
+            await asyncio.sleep(0)
+            if explode:
+                raise RuntimeError('boom')
+
+        await part()
+        ui.button('fire', on_click=lambda: part.refresh(True))
+
+    await user.open('/')
+    user.find('fire').click()
+    await asyncio.sleep(0.1)
+    caplog.records.clear()
+    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'


evnchn:

**TL;DR:**

The async case from #6262 is genuinely fixed — I reproduced it. But the new handler fires on the **awaited** path too, where the caller already sees the exception, and that is a behaviour regression on code that works today.

1. `await part.refresh(...)` on an async body → `ui.on_exception` now fires **twice** (main: once).
2. Worse: if the caller *catches* the exception — the shape of the repo's own `test_awaitable_refresh` — `ui.on_exception` still fires once for an exception that never escaped (main: zero).
3. The `if not context.slot_stack:` guard cannot prevent this — and that one is on me: I suggested it on #6258, where it works, and it does not transfer to the async path. On both `refresh()` branches the stack is already empty, so it never evaluates `False` there.

**I have pushed the fix to your branch as `3b4944cc`** rather than leaving you a diff to apply — it arms the reporting only on the fire-and-forget branch, where the guard was actually trying to point. Your local gates were green and CI was green, because no existing test covered the awaited-async path; the commit also parametrizes your test so it does. Please review it and revert if you would rather take a different route — the PR is yours.

<details>
<summary><b>Before/after matrix (measured, not reasoned)</b></summary>

Worktree at PR head `4c335e46` on top of `1d0405a8`; `uv sync --group dev --group types`. Each row is a `user`-fixture test with both an `app.on_exception` and a `ui.on_exception` handler registered on the same page. Counts are `ui.on_exception` calls.

| scenario | `main` | this PR | expected |
|---|---|---|---|
| async body, `part.refresh(...)` — the issue's case | 0 | **1** ✅ | 1 |
| async body, `await part.refresh(...)` | 1 | **2** ❌ | 1 |
| async body, `await part.refresh(...)` inside caller's `try/except` | 0 | **1** ❌ | 0 |
| async body, direct `await part(...)` — control | 1 | 1 ✅ | 1 |
| async body raising `TypeError` | 0 | **1** ✅ | 1 |
| async body raising before its first `await` | 0 | **1** ✅ | 1 |
| sync body, `part.refresh(...)` — #6258's scope | 0 | 0 | (out of scope) |

`app.on_exception` saw the exception exactly once in every cell on both branches, so the app-level path is unaffected throughout.

With `3b4944cc`, every row above matches the expected column.

</details>

<details>
<summary><b>Blocking — the awaited path over-reports</b></summary>

`refresh()` has two branches (`nicegui/functions/refreshable.py`, `fire_and_forget` / `wait_for_completion`). Only the first one is unobserved; on the second the caller — or the event machinery via `_await_and_handle_in_context` in `nicegui/events.py` — already sees the exception.

Unhandled:

```python
async def update():
    await part.refresh(True)   # async body, no try/except

ui.button('fire', on_click=update)
```
→ `ui.on_exception` fires **2×** on this branch, **1×** on `main`.

Handled — this is the shape of the existing `test_awaitable_refresh` (`tests/test_refreshable.py`, the `Try 0` case):

```python
async def update():
    try:
        await part.refresh(True)
    except RuntimeError:
        ...                     # caller handles it
```
→ `ui.on_exception` fires **1×** on this branch, **0×** on `main`. The exception never escaped, but the client-scoped handler runs anyway.

`test_awaitable_refresh` stays green because it uses the `Screen` fixture and registers no `ui.on_exception`, so it cannot observe either symptom — which is why CI is green. I flagged that test on #6258 as escaping the *doubling* thanks to its own `try/except`; measuring it here shows the `try/except` does not save it from the second symptom, it causes it.

</details>

<details>
<summary><b>Why the <code>if not context.slot_stack:</code> guard cannot catch this</b></summary>

Slot stacks are **task-local**: `Slot.stacks` is keyed by `id(asyncio.current_task())` (`nicegui/slot.py`), and a child task starts with a fresh, empty stack rather than inheriting its parent's.

Both `refresh()` branches reach the awaitable through `asyncio.gather(*awaitables)`, and `gather` wraps each coroutine in a **new Task**. So `_await_with_context` always runs in a task whose stack starts empty; `helpers.await_with_context` pushes `self.container` onto *that* task's stack and pops it again as the exception unwinds. By the time the `except` runs, the stack is empty on the fire-and-forget branch *and* on the awaited one.

Measured, by instrumenting the method to print `len(context.slot_stack)` in its `except`:

```
DIRECT           slot_stack depths=[('RuntimeError', 1)]        ui=1
AWAITED-REFRESH  slot_stack depths=[('RuntimeError', 0), ('RuntimeError', 0)]
```

So the guard is not dead everywhere — on the direct `await part(...)` path the caller's slot is on the same task and the guard correctly declines. It is dead exactly where it needed to discriminate, on `refresh()`.

That is also why the same guard works in #6258: there `target.run()` raises synchronously, inside the caller's task. So the guard was correct advice in the place I gave it and silently wrong one layer down — copying it across was the reasonable move, and the trap is that it fails by doing nothing rather than by erroring.

</details>

<details>
<summary><b>Your regression test does guard — and the parametrized version catches the awaited path</b></summary>

`test_report_exception_async` as written fails on unfixed source with the expected symptom, so it is a real guard:

```
$ git show upstream/main:nicegui/functions/refreshable.py > nicegui/functions/refreshable.py
$ .venv/bin/python -m pytest tests/test_refreshable.py::test_report_exception_async -q
E       AssertionError: ui.on_exception saw 0, expected 1
1 failed in 0.22s
```

Three-way check on the parametrized version in `3b4944cc`:

| source | `[False]` (fire-and-forget) | `[True]` (awaited) |
|---|---|---|
| `upstream/main` | FAIL — the bug | PASS — works today |
| this PR | PASS — fixed | FAIL — the regression |
| `3b4944cc` | PASS | PASS |

</details>

<details>
<summary><b>Gates, and a coordination note with #6258</b></summary>

On your branch as-is: `pylint ./nicegui` 10.00/10 · `mypy ./nicegui` clean · `pre-commit` clean · `tests/test_refreshable.py` 11 passed, 0 skipped. PR description follows the template, and the test uses the `User` fixture rather than `Screen` — both right.

With `3b4944cc`: `pylint` 10.00/10 · `mypy` clean · `pre-commit` clean · `tests/test_refreshable.py` 12 passed · `tests/test_events.py` passing.

`git merge-tree` against #6258: `nicegui/functions/refreshable.py` auto-merges cleanly, but `tests/test_refreshable.py` **conflicts** — both PRs append a test at the end of the file and add the same `import pytest` / `User` import. Whichever lands second needs a small rebase. I also ran the matrix on the two source changes merged together: no cross-interaction, the awaited-path issue is this PR's alone, and `3b4944cc` composes with #6258 (every row correct, and #6258's own awaited-sync and `TypeError` rows stay at 1).

One question rather than a finding: with two targets of the same refreshable both raising, `ui.on_exception` fires twice while `app.on_exception` fires once, because `gather` re-raises only the first. Same in #6258. Arguably correct — two exceptions really did happen — but it may be worth a maintainer call.

</details>

<details>
<summary><b>One mechanism would cover both this PR and #6258 — a maintainer's call, not mine</b></summary>

Using the same `report_exceptions` flag for the synchronous `except` in `_execute_refresh` as well makes one change cover both issues and removes the `context` import and the `slot_stack` guard entirely. Measured `ui.on_exception` counts, `main` / #6258 / this PR / unified:

| scenario | main | #6258 | this PR | unified |
|---|---|---|---|---|
| sync body, `part.refresh(...)` | 0 | 1 | 0 | **1** |
| sync body, awaited | 1 | 1 | 1 | **1** |
| sync body, awaited + caller catches | 0 | 0 | 0 | **0** |
| async body, `part.refresh(...)` | 0 | 0 | 1 | **1** |
| async body, awaited | 1 | 1 | **2** | **1** |
| async body, awaited + caller catches | 0 | 0 | **1** | **0** |
| body raising `TypeError` | 0 | 1 | 0 | **1** |

`pylint` 10.00/10, `mypy` clean, `tests/test_refreshable.py` 12 passed. I have **not** pushed that anywhere — consolidating two open PRs into one is a maintainer decision, and both are fine to land separately. Flagging it only because the split into two threads is what let the `slot_stack` guard get copied into a layer where it silently does nothing.

</details>

Second opinion from a different model lineage (GPT-5 via Codex) on my own findings before posting: it reproduced the double-report independently, could not break the fix against `ui.state()`, `refreshable_method`, nested refreshables and the direct-call path, and corrected me on one point — my first draft said the guard could never be `False`, when in fact it does correctly decline on the direct-call path. The wording above reflects that correction.

---

**Changelog**

- 2026-08-07 — pushed the fix to the branch as `3b4944cc` instead of leaving a suggested diff; removed the "Suggested next commit" section it replaced; added the one-mechanism note for #6258.

SaadZahem:
Thanks for applying the change yourself! The comment was helpful too.

SaadZahem:
May I ask why `report_exception` flag was added to `fire_and_forget` and not `wait_for_completion`?

evnchn:
**Because `wait_for_completion` already has an observer, and `fire_and_forget` has none.** Good question — the flag really means *"is this the path where nobody downstream will see the exception?"*, and the branch name is the only place that is known.

<details>
<summary><b>What happens to the exception on each branch</b></summary>

On `wait_for_completion` the exception propagates out of `await asyncio.gather(...)` straight into whoever awaited `part.refresh(...)`. Two outcomes, both already correct:

- the caller wraps it in `try/except` → they handled it deliberately, and we should stay quiet;
- the caller does not → it reaches `_await_and_handle_in_context` in `nicegui/events.py`, which is running *inside* `with context:`, so `app.handle_exception` finds a slot stack and its `if context.slot_stack and context.client is not None:` branch (`nicegui/app/app.py`) forwards to the client for us.

Either way, reporting from inside `RefreshableTarget` on that branch is a *second* report — which is exactly the two symptoms in the matrix above.

On `fire_and_forget` the awaitable is handed to `background_tasks.create_or_defer(...)`, and the only thing that ever sees the failure is the `_handle_exceptions` done-callback in `nicegui/background_tasks.py`. That runs as an event-loop callback with no slot stack, so the same `if context.slot_stack ...` branch is skipped and the client-scoped handlers never fire. That is the hole `report_exceptions=True` fills.

So the asymmetry is not about async vs sync, or about where the `try/except` is cheapest to write — it is that one branch has a downstream observer by construction and the other has none.

</details>

evnchn (dismissed):

**Re-verified at 3b4944cc and approving.** Nothing has changed on the branch since the matrix in my earlier comment, and I re-ran the gates rather than assume: `pylint ./nicegui` 10.00/10 · `mypy ./nicegui` clean, 245 source files · `tests/test_refreshable.py` 12 passed.

One coordination note that still stands: this and 6258 both append a test to the end of `tests/test_refreshable.py` and both add the same `import pytest` / `User` import, so whichever lands second needs a small rebase. The source files themselves auto-merge cleanly.

evnchn:

**Copilot's flag is real — I verified it in source, and it's the same masking bug this PR family exists to kill.** `Element.client` is a weakref-backed property that raises `RuntimeError('The client this element belongs to has been deleted.')` when the client is gone. So if the client is deleted while the async body is in flight (user navigates away mid-refresh), the `except` block in `_await_and_report` raises *before* your `raise` runs, and the original exception is replaced.

The guard is the same one you used in #6270:

```python
except Exception as e:
    if not self.container.is_deleted:
        self.container.client.handle_exception(e)
    raise
```

<details>
<summary><b>Why <code>is_deleted</code> is the right guard, and why skipping the report is correct</b></summary>

- `Element.client` (element.py): `client = self._client()` → raises `RuntimeError` when the weakref is dead.
- `Client.delete()` (client.py) calls `remove_all_elements()` — which sets `element._deleted = True` on every element — before the client object can be collected. So on NiceGUI's normal deletion path, `container.is_deleted` is a reliable, non-raising proxy for "client gone".
- Skipping the report on a deleted client isn't just safe, it's correct: `ui.on_exception` handlers are per-client (`client._exception_handlers`), so there is nobody left to report to. The `raise` still surfaces the original exception in the background-task log.

A regression test is easy to stage here: delete the client during the `await` (or have the body do it), then assert the *original* error — not `'The client … has been deleted'` — is what ends up in the log.
</details>

evnchn:

**The hazard Copilot flagged is real and I now have it on video — plus the one-line fix, verified.**

```python
except Exception as e:
    if not self.container.is_deleted:
        self.container.client.handle_exception(e)
    raise
```

<details>
<summary><b>Repro, and why a shorter test would have told you it was fine</b></summary>

Async body that outlives its client, on your head `3b4944cc`:

| `_await_and_report` | what reaches the log |
|---|---|
| as on this branch | `The client this element belongs to has been deleted.` ⛔ |
| with the `is_deleted` guard | `THE REAL ERROR` ✅ |

`tests/test_refreshable.py` stays at 13 passed with the guard in.

The reason this needs a long sleep: `Client.delete()` removes the client from `Client.instances`,
but the suspended `Outbox.loop` coroutine keeps the object alive until it exits and a GC pass
reclaims it (~0.115 s, measured). Any probe whose exception lands sooner sees a live weakref and
concludes there is no bug. I made exactly that mistake before running it properly.
</details>

evnchn:

**Closing as consolidated — we agreed privately to carry this work forward in #6258, which now contains this branch's fix verbatim.** #6258's head includes the async fix from here (with the `is_deleted` guard) plus the sync-path test, and its body now references #6262 as well, so nothing from this PR is lost. The review thread here remains the record for the awaited-path / double-report discussion.

Thanks @SaadZahem for doing the merge itself — one branch to review beats two.

Issue #6267, opened 2026-08-07
tests/README.md: `chromium-chromedriver` is not a Debian package, and on Ubuntu it only pulls the chromium snap


evnchn:
*Filed by Claude Code on evnchn's behalf.*

**TL;DR:**

`tests/README.md` tells Linux contributors to run `sudo apt-get install chromium-chromedriver`. Verified in clean containers, that command is wrong on both distro families the surrounding note claims to cover:

1. **Debian 12** — no such package. `E: Unable to locate package chromium-chromedriver`. The correct name is `chromium-driver`.
2. **Ubuntu 24.04** — the name exists, but it is a transitional stub that pulls in `snapd` rather than a real ChromeDriver. Renaming to `chromium-driver` does **not** help, because that name only exists there as a virtual name provided by the same stub.

The Debian half is a one-word fix. The Ubuntu half is a genuine open question, so I am reporting it rather than proposing a patch for it.

Spun out of the Arch instructions PR so as not to load a pre-existing bug onto a contributor's small docs change.

<details>
<summary><b>Debian 12 — the documented package does not exist</b></summary>

`debian:bookworm-slim`, arm64, 2026-08-07:

```console
$ apt-get update -qq
$ apt-get install -y chromium-chromedriver
E: Unable to locate package chromium-chromedriver

$ apt-cache policy chromium-driver
chromium-driver:
  Installed: (none)
  Candidate: 151.0.7922.71-1~deb12u1
  Version table:
     151.0.7922.71-1~deb12u1 500
        500 http://deb.debian.org/debian-security bookworm-security/main arm64 Packages
```

The correct package installs and works:

```console
$ command -v chromedriver || echo "ABSENT"
ABSENT

$ apt-get install -y chromium-driver
$ chromedriver --version
ChromeDriver 151.0.7922.71 (ef35003457e93c278f911a334b06e4a5f8967e06-refs/branch-heads/7922_47@{#11})
$ chromium --version
Chromium 151.0.7922.71 built on Debian GNU/Linux 12 (bookworm)
```

Driver and browser versions match, which is what the README preamble asks for.

</details>

<details>
<summary><b>Ubuntu 24.04 — both names lead to snap</b></summary>

`ubuntu:24.04`, arm64. The documented name resolves, but to a stub:

```console
$ apt-cache show chromium-chromedriver
Package: chromium-chromedriver
Version: 2:1snap1-0ubuntu2
Section: universe/web
Installed-Size: 18
[...]

Issue #6246, opened 2026-08-06
`ui.scene`'s `axes_helper` renders white instead of red/green/blue (regression since 2.3.0)


evnchn:
### First Check

- [x] I added a very descriptive title here.
- [x] This is not a security issue (those should be reported via the [security advisory](https://github.com/zauberzeug/nicegui/security/advisories/new) instead).
- [x] This is not a Q&A. I am sure something is wrong with NiceGUI or its documentation.
- [x] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

with ui.scene(width=400, height=400, background_color='#222222') as scene:
    scene.axes_helper()

ui.run()
```

### Description

`AxesHelper`'s docstring promises "The X axis is red. The Y axis is green. The Z axis is blue." On a page load the axes render **flat white** instead.

This is a regression, not a long-standing wart — it worked until 2.3.0. Same script, same probe, two releases:

| NiceGUI | `material.vertexColors` | `geometry.attributes.color` | renders as |
|---|---|---|---|
| 2.2.0 (2024-09-27) | `true` | present | red / green / blue ✅ |
| 2.3.0 (2024-10-04) | `false` | present | flat white ❌ |

| 2.2.0 — as documented | 2.3.0 — the regression |
|---|---|
| <img width="310" alt="axes_helper on 2.2.0: X red, Z blue" src="https://github.com/user-attachments/assets/e3f23d81-e6d7-47e9-b9e0-025899ab77e8" /> | <img width="310" alt="axes_helper on 2.3.0: flat white" src="https://github.com/user-attachments/assets/5a7aa25f-74f6-456a-9438-d1c218ca6a52" /> |

Both captured from the same script (`axes_helper(length=3.0)`, grid off, identical camera) against the two releases.

**Cause.** `init_objects` replays `material(id, color, opacity, side)` on page load for every non-GLTF object ([`scene.js#L610`](https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/scene/scene.js#L610)). `AxesHelper` inherits `Object3D`'s default `color = '#ffffff'`, and since #3545 `material()` derives the vertex-color flag from it:

```js
const vertexColors = color === null;
material.color.set(vertexColors ? "#ffffff" : color);
material.vertexColors = vertexColors;
```

[...]

PR #6269, opened 2026-08-07
Report exceptions from async event and validation handler to `ui.on_exception`

### LINKED ISSUE #6233: `ui.on_exception` still misses exceptions from async `Event` subscribers and async `validation` functions (follow-up to #5945)
### First Check

### Example Code

```python
# Save as mre.py next to a pytest.ini containing:
#     [pytest]
#     asyncio_mode = auto
#     main_file =
# then:  pip install nicegui==3.15.0 "pytest>=9,<10" pytest-asyncio && pytest mre.py
#
# Each pair below differs by ONE line: the first passes, the second fails.
import asyncio

import pytest
from nicegui import Event, ui

# ---- Case A: a nicegui.Event subscriber ----

async def test_A1_sync_subscriber(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    @ui.page('/')
    def page():
        event: Event[[]] = Event()

        def subscriber():                       # sync
            raise RuntimeError('boom')

        event.subscribe(subscriber)
        ui.on_exception(seen.append)
        ui.button('fire', on_click=lambda: event.emit())

    await user.open('/')
    user.find('fire').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'


async def test_A2_async_subscriber(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    @ui.page('/')
    def page():
        event: Event[[]] = Event()

        async def subscriber():                 # async -- the only difference
            raise RuntimeError('boom')

        event.subscribe(subscriber)
        ui.on_exception(seen.append)
        ui.button('fire', on_click=lambda: event.emit())

    await user.open('/')
    user.find('fire').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'


# ---- Case B: a validation function ----

async def test_B1_sync_validation(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    def check(value):                           # sync
        raise RuntimeError('boom')

    @ui.page('/')
    def page():
        ui.on_exception(seen.append)
        field = ui.input('field', validation=check)
        ui.button('validate', on_click=lambda: field.validate(return_result=False))

    await user.open('/')
    user.find('validate').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'


async def test_B2_async_validation(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []

    async def check(value):                     # async -- the only difference
        raise RuntimeError('boom')

    @ui.page('/')
    def page():
        ui.on_exception(seen.append)
        field = ui.input('field', validation=check)
        ui.button('validate', on_click=lambda: field.validate(return_result=False))

    await user.open('/')
    user.find('validate').click()
    await asyncio.sleep(0.1)
    caplog.records.clear()
    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'
```

### Description

**TL;DR:**

PR #5946 fixed `ui.on_exception` for async **event handlers** (reported in #5945) by catching the exception *inside* the slot context. The same failure mode still appears on two more deferred paths, so a handler registered with `ui.on_exception` silently never fires:

1. an **async subscriber to a `nicegui.Event`** — the sync subscriber right beside it is caught
2. an **async `validation=` function** — again, the sync one is caught

In every case the exception still reaches `app.on_exception` and the log, so nothing is lost silently at the app level — but the client-scoped handler that the user registered on that page does not run.

The ask: confirm these should be caught, and whether you'd prefer one shared helper (promoting `events._await_and_handle_in_context` to `helpers`) over two local fixes.

A third instance with a *different* mechanism — `@ui.refreshable` rebuilds — is filed separately as #6234 so the two can be worked in parallel.

<details>
<summary><b>Mechanism</b></summary>

`app.handle_exception` only forwards to the client-scoped handlers when a slot stack is present — `nicegui/app/app.py:178` on 3.15.0:

```python
    context.client.handle_exception(exception)
```

An exception that escapes a coroutine scheduled with `background_tasks.create*` surfaces in the task's done-callback `background_tasks._handle_exceptions`, which runs with an **empty** slot stack. So it reaches `app.on_exception` but never `ui.on_exception`.

That is exactly what #5946 addressed, by awaiting inside a helper that catches while still in the context (`nicegui/events.py:481` and `:486`).

**Case A — `nicegui/event.py:137`** still has the pre-#5946 shape:

```python
def _invoke_and_forget(callback: Callback[P], *args: P.args, **kwargs: P.kwargs) -> Any:
    try:
        result = callback.run(*args, **kwargs)
        if helpers.should_await(result):
            background_tasks.create_or_defer(callback.await_result(result), name=...)
    except Exception as e:
        core.app.handle_exception(e)     # only reachable from the sync branch
```

`Callback.await_result` (`event.py:35`) re-enters the slot context for the await but has no `try`/`except`, so the exception leaves the context before anything handles it.

**Case B — `nicegui/elements/mixins/validation_element.py:70-74`**: no `try`/`except` and no slot context at all.

```python
result = self._validation(self.value)
if helpers.should_await(result):
    async def await_error():
        self.error = await result
    background_tasks.create(await_error(), name=f'validate {self.id}')
```

</details>

<details>
<summary><b>Actual output (nicegui 3.15.0 from PyPI, clean venv)</b></summary>

```
test_A1_sync_subscriber PASSED               [ 25%]
test_A2_async_subscriber FAILED              [ 50%]
test_B1_sync_validation PASSED               [ 75%]
test_B2_async_validation FAILED              [100%]
========================= 2 failed, 2 passed in 0.86s ==========================
```

Each failure is `AssertionError: ui.on_exception saw 0, expected 1`. Case A's captured traceback ends in the done-callback, which is the mechanism above:

```
ERROR    nicegui:app.py:181 boom
    task.result()
  File ".../nicegui/event.py", line 38, in await_result
    return await result
  File ".../mre.py", line 46, in subscriber
    raise RuntimeError('boom')
RuntimeError: boom
```

</details>

<details>
<summary><b>Both cases side by side: app-level fires, client-level does not</b></summary>

Registering both an `app.on_exception` and a `ui.on_exception` handler on the same page and firing both cases in one run (nicegui 3.15.0):

```
app.on_exception saw : ['async Event subscriber', 'async validation']
ui.on_exception saw  : []
```

So nothing is being swallowed at the app level — the defect is precisely that the client-scoped handler is skipped. (This is also why the log still shows the traceback, which may be why it has gone unnoticed.)

</details>

<details>
<summary><b>Candidate fix for Case A, and what it was checked against</b></summary>

Applying the #5946 shape to `Callback.await_result` makes `test_A2` pass, and the repository's existing `tests/test_event.py` still passes 11/11:

```python
async def await_result_and_handle_in_context(self, result: Awaitable) -> Any:
    """Await the result within the callback's slot context, handling exceptions in-context."""
    with (self.slot and self.slot()) or nullcontext():
        try:
            return await result
        except Exception as e:
            core.app.handle_exception(e)
```

This is offered as evidence that the diagnosis is right, **not** as a proposed patch — Case B was not fixed, and you may prefer one shared helper. `helpers.await_with_context` has the same "re-enter the context but don't catch in it" behaviour and is called at four sites on 3.15.0 (`client.py:426`, `client.py:453`, `functions/navigate.py:75`, `functions/refreshable.py:43`), so a `handle_exceptions=True` variant of it may be the smaller total change.

</details>

<details>
<summary><b>Swept and NOT affected</b></summary>

Checked with the same harness on 3.15.0, no defect found:

| site | result |
|---|---|
| `events.handle_event` (the #5946 fix site) | correct — the control case in the MRE passes |
| `ui.sub_pages` async page builder (`sub_pages.py`, `functions/navigate.py`) | correct — both sync and async builders reach `ui.on_exception` |
| `app.on_connect` / `Client.safe_invoke` (`client.py:419`) | neither sync nor async connect-handler exceptions reach `ui.on_exception`. Reported here only for completeness — this looks like intended app-scoped-lifecycle vs in-page semantics rather than a bug, so it is deliberately not part of the report. |

One claim was **withdrawn** during preparation: an apparent "sync validation is not caught either" result turned out to be an artifact of the test harness — `user.find(...).type(...)` assigns `element.value` directly, bypassing `handle_event`, which a real browser would not do. Case B above therefore drives validation through `field.validate()` from a real click handler instead.

</details>

### NiceGUI Version

3.15.0

### Python Version

3.12.11

### Browser

Other

### Operating System

macOS

### Additional Context

Browser-independent — the repro uses `nicegui.testing.User`, so no browser is involved. The MRE deliberately mirrors the regression test added in #5946 (`tests/test_event.py::test_ui_on_exception`) so the comparison is apples to apples.

Related but distinct: #6230 also concerns `app.handle_exception`, but is about the guard itself raising, not about the client-scoped handlers being skipped.

SaadZahem:
### Motivation

Fixes #6233

### Implementation

The fix is just an echo of #5946, #6258, #6265.
It is a `try/except` that wraps whatever code was already there.
It is indeed messy, and it fixes the described issue.

I tested using the 4 test cases provided in #6233.
Pytests were added:
1. `test_ui_on_exception_from_async_subscriber` in `tests/test_event.py`
2. `test_ui_on_exception_from_async_validation` in `tests/test_input.py`

Ready for review!

### Progress

### DIFF
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -3,7 +3,7 @@ from typing import Any, TypeAlias, cast
 
 from typing_extensions import Self
 
-from ... import background_tasks, helpers
+from ... import background_tasks, context, helpers
 from ...events import ValueT
 from .value_element import ValueElement
 
@@ -70,7 +70,12 @@ class ValidationElement(ValueElement[ValueT]):
             result = self._validation(self.value)
             if helpers.should_await(result):
                 async def await_error():
-                    self.error = await result
+                    try:
+                        self.error = await result
+                    except Exception as e:
+                        if not context.slot_stack:  #6233
+                            self.client.handle_exception(e)
+                        raise
                 background_tasks.create(await_error(), name=f'validate {self.id}')
                 return True
             self.error = cast(str | None, result)
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -4,7 +4,6 @@ from typing import Any, TypeAlias, cast
 from typing_extensions import Self
 
 from ... import background_tasks, helpers
-from ...context import context
 from ...events import ValueT
 from .value_element import ValueElement
 
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -74,7 +74,7 @@ class ValidationElement(ValueElement[ValueT]):
                     try:
                         self.error = await result
                     except Exception as e:
-                        if not context.slot_stack:  # 6233
+                        if not self.is_deleted:
                             self.client.handle_exception(e)
                         raise
                 background_tasks.create(await_error(), name=f'validate {self.id}')
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -74,7 +74,7 @@ class ValidationElement(ValueElement[ValueT]):
                     try:
                         self.error = await result
                     except Exception as e:
-                        if not context.slot_stack:  #6233
+                        if not context.slot_stack:  # 6233
                             self.client.handle_exception(e)
                         raise
                 background_tasks.create(await_error(), name=f'validate {self.id}')
--- a/nicegui/elements/mixins/validation_element.py
+++ b/nicegui/elements/mixins/validation_element.py
@@ -3,7 +3,8 @@ from typing import Any, TypeAlias, cast
 
 from typing_extensions import Self
 
-from ... import background_tasks, context, helpers
+from ... import background_tasks, helpers
+from ...context import context
 from ...events import ValueT
 from .value_element import ValueElement
 
--- a/nicegui/event.py
+++ b/nicegui/event.py
@@ -4,7 +4,7 @@ import asyncio
 import inspect
 import weakref
 from collections.abc import Awaitable, Callable
-from contextlib import nullcontext
+from contextlib import nullcontext, suppress
 from dataclasses import dataclass
 from typing import Any, ClassVar, Generic
 from weakref import WeakSet
@@ -32,16 +32,19 @@ class Callback(Generic[P]):
         with (self.slot and self.slot()) or nullcontext():
             return self.func(*args, **kwargs) if self.expect_args else self.func()  # type: ignore[call-arg]
 
-    async def await_and_handle_result(self, result: Awaitable) -> Any:
-        """Await the result of the callback reporting exceptions if necessary."""
-        empty_stack = not context.slot_stack  # 6233
+    async def await_and_handle_result(self, result: Awaitable, *, report_exceptions: bool = False) -> Any:
+        """Await the result of the callback, reporting exceptions if nobody downstream will."""
         with (self.slot and self.slot()) or nullcontext():
-            if not empty_stack:
+            if not report_exceptions:
                 return await result
             try:
                 return await result
             except Exception as e:
-                core.app.handle_exception(e)
+                client = None
+                with suppress(RuntimeError):  # the client may already be deleted
+                    client = context.client if context.slot_stack else None
+                if client is not None:
+                    client.handle_exception(e)
                 raise
 
 
@@ -145,7 +148,7 @@ def _invoke_and_forget(callback: Callback[P], *args: P.args, **kwargs: P.kwargs)
     try:
         result = callback.run(*args, **kwargs)
         if helpers.should_await(result):
-            background_tasks.create_or_defer(callback.await_and_handle_result(result),
+            background_tasks.create_or_defer(callback.await_and_handle_result(result, report_exceptions=True),
                                              name=f'{callback.filepath}:{callback.line}')
     except Exception as e:
         core.app.handle_exception(e)
--- a/nicegui/event.py
+++ b/nicegui/event.py
@@ -32,10 +32,17 @@ class Callback(Generic[P]):
         with (self.slot and self.slot()) or nullcontext():
             return self.func(*args, **kwargs) if self.expect_args else self.func()  # type: ignore[call-arg]
 
-    async def await_result(self, result: Awaitable) -> Any:
-        """Await the result of the callback."""
+    async def await_and_handle_result(self, result: Awaitable) -> Any:
+        """Await the result of the callback reporting exceptions if necessary."""
+        empty_stack = not context.slot_stack  #6233
         with (self.slot and self.slot()) or nullcontext():
-            return await result
+            if not empty_stack:
+                return await result
+            try:
+                return await result
+            except Exception as e:
+                core.app.handle_exception(e)
+                raise
 
 
 class Event(Generic[P]):
@@ -138,7 +145,7 @@ def _invoke_and_forget(callback: Callback[P], *args: P.args, **kwargs: P.kwargs)
     try:
         result = callback.run(*args, **kwargs)
         if helpers.should_await(result):
-            background_tasks.create_or_defer(callback.await_result(result), name=f'{callback.filepath}:{callback.line}')
+            background_tasks.create_or_defer(callback.await_and_handle_result(result), name=f'{callback.filepath}:{callback.line}')
     except Exception as e:
         core.app.handle_exception(e)
 
@@ -146,7 +153,7 @@ def _invoke_and_forget(callback: Callback[P], *args: P.args, **kwargs: P.kwargs)
 async def _invoke_and_await(callback: Callback[P], *args: P.args, **kwargs: P.kwargs) -> Any:
     result = callback.run(*args, **kwargs)
     if helpers.should_await(result):
-        result = await callback.await_result(result)
+        result = await callback.await_and_handle_result(result)
     return result
 
 
--- a/nicegui/event.py
+++ b/nicegui/event.py
@@ -34,7 +34,7 @@ class Callback(Generic[P]):
 
     async def await_and_handle_result(self, result: Awaitable) -> Any:
         """Await the result of the callback reporting exceptions if necessary."""
-        empty_stack = not context.slot_stack  #6233
+        empty_stack = not context.slot_stack  # 6233
         with (self.slot and self.slot()) or nullcontext():
             if not empty_stack:
                 return await result
@@ -145,7 +145,8 @@ def _invoke_and_forget(callback: Callback[P], *args: P.args, **kwargs: P.kwargs)
     try:
         result = callback.run(*args, **kwargs)
         if helpers.should_await(result):
-            background_tasks.create_or_defer(callback.await_and_handle_result(result), name=f'{callback.filepath}:{callback.line}')
+            background_tasks.create_or_defer(callback.await_and_handle_result(result),
+                                             name=f'{callback.filepath}:{callback.line}')
     except Exception as e:
         core.app.handle_exception(e)
 
--- a/tests/test_event.py
+++ b/tests/test_event.py
@@ -204,3 +204,24 @@ async def test_ui_on_exception(user: User, caplog: pytest.LogCaptureFixture):
     await asyncio.sleep(0.1)
     assert len(exceptions) == 2 and 'sync error' in str(exceptions[0]) and 'async error' in str(exceptions[1])
     caplog.records.clear()
+
+
+async def test_ui_on_exception_from_async_subscriber(user: User, caplog: pytest.LogCaptureFixture):
+    seen: list[Exception] = []
+
+    @ui.page('/')
+    def page():
+        event: Event[[]] = Event()
+
+        async def subscriber():
+            raise RuntimeError('boom')
+
+        event.subscribe(subscriber)
+        ui.on_exception(seen.append)
+        ui.button('fire', on_click=lambda: event.emit())
+
+    await user.open('/')
+    user.find('fire').click()
+    await asyncio.sleep(0.1)
+    caplog.records.clear()
+    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'
--- a/tests/test_event.py
+++ b/tests/test_event.py
@@ -225,3 +225,36 @@ async def test_ui_on_exception_from_async_subscriber(user: User, caplog: pytest.
     await asyncio.sleep(0.1)
     caplog.records.clear()
     assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'
+
+
+@pytest.mark.parametrize('caught', [False, True])
+async def test_awaited_event_call_reports_once(user: User, caplog: pytest.LogCaptureFixture, caught: bool):
+    seen: list[Exception] = []
+
+    @ui.page('/')
+    def page():
+        event: Event[[]] = Event()
+
+        async def subscriber():
+            raise RuntimeError('boom')
+
+        event.subscribe(subscriber)
+        ui.on_exception(seen.append)
+
+        async def fire():
+            if not caught:
+                await event.call()
+                return
+            try:
+                await event.call()
+            except RuntimeError:
+                pass
+
+        ui.button('fire', on_click=fire)
+
+    await user.open('/')
+    user.find('fire').click()
+    await asyncio.sleep(0.1)
+    caplog.records.clear()
+    expected = 0 if caught else 1
+    assert len(seen) == expected, f'ui.on_exception saw {len(seen)}, expected {expected}'
--- a/tests/test_input.py
+++ b/tests/test_input.py
@@ -6,7 +6,7 @@ from selenium.webdriver.common.by import By
 from selenium.webdriver.common.keys import Keys
 
 from nicegui import ui
-from nicegui.testing import Screen
+from nicegui.testing import Screen, User
 
 
 def test_input(screen: Screen):
@@ -127,6 +127,25 @@ def test_validation_with_lagging_value_change_events(screen: Screen):
     element.send_keys('45678')  # keep typing after the error message arrived
     screen.should_contain('Still 2 characters missing')
     assert element.get_attribute('value') == '12345678'
+
+
+async def test_ui_on_exception_from_async_validation(user: User, caplog: pytest.LogCaptureFixture):
+    seen: list[Exception] = []
+
+    async def check(value):
+        raise RuntimeError('boom')
+
+    @ui.page('/')
+    def page():
+        ui.on_exception(seen.append)
+        field = ui.input('field', validation=check)
+        ui.button('validate', on_click=lambda: field.validate(return_result=False))
+
+    await user.open('/')
+    user.find('validate').click()
+    await asyncio.sleep(0.1)
+    caplog.records.clear()
+    assert len(seen) == 1, f'ui.on_exception saw {len(seen)}, expected 1'
 
 
 def test_input_with_multi_word_error_message(screen: Screen):


evnchn:

**TL;DR:**

Both halves of #6233 are genuinely fixed and both of your tests are real guards — I checked they fail on unfixed source. But there are three behaviour regressions, and the third one is the bug family this whole sweep is about: on this branch, an async subscriber that raises **after its client is deleted** loses its exception to the same `RuntimeError` mask as #6230.

1. **A post-deletion exception is masked** (blocking). `app.on_exception` receives `RuntimeError: The parent element this slot belongs to has been deleted.` instead of the application's real error. **Main gets this right today** — the PR breaks it.
2. **`await event.call()` double-reports** (blocking). `ui.on_exception` fires **2×** where main fires once, and **1×** where main fires zero because the caller handled it.
3. **`app.on_exception` fires twice on the path this PR is fixing** (blocking, easy to miss). `core.app.handle_exception(e)` reports, then the re-raise reaches `background_tasks._handle_exceptions`, which reports again.

All three are the same one-line-ish fix, below.

<details>
<summary><b>Before/after matrix (measured, not reasoned)</b></summary>

Worktree at PR head b82e79af on top of 1d0405a8, `uv sync --group dev --group types`. Each row is a `user`-fixture test with both an `app.on_exception` and a `ui.on_exception` handler registered on the same page.

| scenario | main | this PR | expected |
|---|---|---|---|
| `event.emit()`, async subscriber — the issue's case | ui=0 app=1 | ui=1 **app=2** ❌ | ui=1 app=1 |
| `await event.call()`, caller does not catch | ui=1 app=1 | **ui=2** ❌ app=2 | ui=1 app=1 |
| `await event.call()`, caller catches | ui=0 app=0 | **ui=1 app=1** ❌ | ui=0 app=0 |
| `event.emit()`, sync subscriber — control | ui=1 app=1 | ui=1 app=1 ✅ | unchanged |
| async `validation=` raising | ui=0 app=1 | ui=1 app=1 ✅ | ui=1 app=1 |
| async subscriber raising **after `Client.delete()`** | `ValueError: THE REAL ERROR` | **`RuntimeError: The parent element…`** ❌ | the `ValueError` |

The `await event.call()` + caller-catches row is the nastier of the middle two: the exception never escaped the caller's `try/except`, and a client-scoped handler runs anyway.

The last row is the one that matters most, because it is a *regression against working behaviour* rather than an incomplete fix.

</details>

**I have pushed the fix to your branch as 8c0b52eb** rather than leaving you a diff to apply — revert it if you would rather take a different route, the PR is yours. It also adds a parametrized `Event.call()` test, since that path is what stayed invisible to a green CI.

```diff
-    async def await_and_handle_result(self, result: Awaitable) -> Any:
-        """Await the result of the callback reporting exceptions if necessary."""
-        empty_stack = not context.slot_stack  # 6233
+    async def await_and_handle_result(self, result: Awaitable, *, report_exceptions: bool = False) -> Any:
+        """Await the result of the callback, reporting exceptions if nobody downstream will."""
         with (self.slot and self.slot()) or nullcontext():
-            if not empty_stack:
+            if not report_exceptions:
                 return await result
             try:
                 return await result
             except Exception as e:
-                core.app.handle_exception(e)
+                client = None
+                with suppress(RuntimeError):  # the client may already be deleted
+                    client = context.client if context.slot_stack else None
+                if client is not None:
+                    client.handle_exception(e)
                 raise
```

```diff
         if helpers.should_await(result):
-            background_tasks.create_or_defer(callback.await_and_handle_result(result),
+            background_tasks.create_or_defer(callback.await_and_handle_result(result, report_exceptions=True),
                                              name=f'{callback.filepath}:{callback.line}')
```

plus `from contextlib import nullcontext, suppress` on the existing import line.

With those hunks every row matches the expected column, including the post-deletion one.

The new test fails both ways on the previous head — `ui.on_exception saw 2, expected 1` uncaught and `saw 1, expected 0` caught — so it guards findings 2 and 3. Finding 1 I deliberately left untested: reproducing it needs `del` plus `gc.collect()` to force the weakref to die, which is the scaffolding `AGENTS.md` warns about. Happy to be overruled on that.

<details>
<summary><b>Why <code>not context.slot_stack</code> cannot discriminate here</b></summary>

`Slot.stacks` is keyed by `id(asyncio.current_task())` (`nicegui/slot.py`), and a child task starts with a fresh, empty stack rather than inheriting its parent's.

`Event.call()` is:

```python
await asyncio.gather(*[_invoke_and_await(callback, *args, **kwargs) for callback in self.callbacks])
```

`gather` wraps each coroutine in a **new Task**, so by the time `await_and_handle_result` runs its first line, the stack is empty on the `call()` path exactly as it is on the `emit()` path. The flag it was reaching for is the branch it was called from — and the call site is the only place that knows that. Same conclusion as 6265, one layer up.

The version above still reads `context.slot_stack`, but for a different job: there it decides *which* handler to reach, not *whether* to report.

</details>

<details>
<summary><b>Finding 1 — the post-deletion mask, and why my own first draft had it too</b></summary>

The repro is an async subscriber whose page is closed mid-flight:

```python
async def subscriber():
    await asyncio.sleep(0.2)      # client is deleted during this
```

| version | what `app.on_exception` receives |
|---|---|
| main | `ValueError: THE REAL ERROR` |
| this PR | `RuntimeError: The parent element this slot belongs to has been deleted.` |

On main the exception escapes to `background_tasks._handle_exceptions`, which calls `core.app.handle_exception` from an event-loop callback with no slot stack — so the `if context.slot_stack and context.client is not None:` branch is skipped, nothing is dereferenced, and the real error survives.

On this branch `core.app.handle_exception(e)` is called from *inside* `with self.slot()`, so the slot stack is non-empty, `context.client` dereferences a dead weakref, and the `RuntimeError` replaces the `ValueError` — which is #6230 at a new site.

Worth saying plainly: **my first version of the suggested fix above had exactly the same defect**, for exactly the same reason — I wrote `if context.slot_stack and context.client is not None:` by copying `App.handle_exception`, which is the very line #6230 is filed against. A second reviewer on a different model caught it, and only then did I measure that row. The `suppress(RuntimeError)` around the *lookup* is what fixes it, and it has to wrap only the lookup: wrapping the `handle_exception` call too would silently eat any `RuntimeError` a user's own `ui.on_exception` handler raises.

</details>

<details>
<summary><b>Finding 3 — why the app-level count goes to 2, and why the client handler is the right target</b></summary>

`background_tasks._handle_exceptions` already reports anything that escapes a background task:

```python
def _handle_exceptions(task: asyncio.Task) -> None:
    try:
        task.result()
        pass
    except Exception as e:
        core.app.handle_exception(e)
```

So on the `emit()` path the app-level handlers were never the gap — they already fired. The gap is that this callback runs with no slot stack, so `App.handle_exception` skips its client-forwarding branch and the **client**-scoped handlers never fire.

Reporting to `core.app.handle_exception` therefore fixes the client-side hole by adding a second app-side report. Reporting to the client fills exactly the hole and leaves the app count alone — which is what 3b4944cc does in `refreshable.py`, and what you already did yourself in `validation_element.py`.

</details>

<details>
<summary><b>Minor — the guard in <code>validation_element.py</code> is dead</b></summary>

```python
async def await_error():
    try:
        self.error = await result
    except Exception as e:
        if not context.slot_stack:  # 6233
            self.client.handle_exception(e)
        raise
background_tasks.create(await_error(), name=f'validate {self.id}')
```

`await_error()` only ever runs inside the task created on the line below it, and per the task-keyed stacks above, that task's stack is always empty. Instrumenting the `except` to print the depth confirms it: `slot_stack=0`.

So the condition is unconditionally true. The behaviour is right — reporting *is* what should happen there, since nothing downstream sees this one — but it reads as though it were preventing a double-report, and the next person to copy it somewhere the stack is not empty inherits the 6265 bug. I would drop it and report unconditionally. Not blocking; the measured row is correct either way.

(`self.client` is used directly there rather than via `context`, so this site does not have finding 1's masking problem — `Element.client` raises only if the client object itself was collected, and this path holds a reference through `self`.)

</details>

<details>
<summary><b>Your tests do guard</b></summary>

Both new tests fail against 1d0405a8 source with the expected symptom, so they are real regression tests rather than green-only ones:

```
FAILED tests/test_event.py::test_ui_on_exception_from_async_subscriber - AssertionError: ui.on_exception saw 0, expected 1
FAILED tests/test_input.py::test_ui_on_exception_from_async_validation - AssertionError: ui.on_exception saw 0, expected 1
2 failed in 0.45s
```

Using the `user` fixture for both was the right call. Neither covers `Event.call()` or the post-deletion path, which is how findings 1 and 2 stayed invisible to a green CI — both worth a row once the fix lands.

</details>

<details>
<summary><b>Gates</b></summary>

On your branch as-is: `pylint` 10.00/10 · `mypy` clean · `tests/test_event.py` and `tests/test_input.py` green — which is the point about the matrix: all three regressions are invisible to the current suite.

With the suggested hunks: `pylint ./nicegui` 10.00/10 · `mypy ./nicegui` clean, 245 source files · `tests/test_event.py tests/test_input.py` 30 passed.

Second opinion from a different model lineage (GPT-5 via Codex) on my own findings before posting: it refuted the first version of my suggested fix by pointing at the unguarded `context.client` dereference, which is what produced finding 1 and the last row of the matrix. It also corrected two overstatements in my draft, both fixed above.

</details>

The rename to `await_and_handle_result` is a good change and worth keeping either way.

evnchn:

Separate from the review above — you have cleared most of what I filed in that sweep, and I did not want you to run out of runway without knowing the rest is fair game. Anything open under my name in this repo is yours to take if it looks interesting; no need to ask first, and no expectation that you take any of them.

Currently unclaimed:

- #6267 — `chromium-chromedriver` is not a Debian package (docs; the sibling of the arch-linux one you already fixed)
- #6246 — `ui.scene`'s `axes_helper` renders white instead of red/green/blue
- #6247 — `TrackballControls` keeps a stale canvas rect after a window resize
- #6235 — setting `table.pagination` after construction paginates the rows but leaves the control stale
- #6230 — the one we are already talking about, if you want to turn that diff into a PR

Pick by whatever you feel like working on rather than by what would help me — the scene and table ones are more involved than the docs one, and none of them are urgent.

Also, if you want a faster loop than PR comments: NiceGUI has a Discord at https://discord.gg/TEpFeAaF4f, and it is a much better place to think out loud than a review thread. Several of the exchanges we have had on these PRs would have taken two minutes there. No obligation — but you have been putting real work into this corner of the codebase and it would be good to have you around.

SaadZahem:
This PR is slightly heavier than the others because it tried to fix multiple things. That for sure requires more tests before commiting to a change. The review was only helpful and the fixes are appropriate. Good things you kept the function rename.

> await_error() only ever runs inside the task created on the line below it, and per the task-keyed stacks above, that task's stack is always empty. Instrumenting the except to print the depth confirms it: slot_stack=0.

For example, I would have known that if I did more tests before committing.
_And who knows if the guard is only dead for today and may wake up later?_

SaadZahem:
> Separate from the review above — you have cleared most of what I filed in that sweep, and I did not want you to run out of runway without knowing the rest is fair game. Anything open under my name in this repo is yours to take if it looks interesting; no need to ask first, and no expectation that you take any of them.

It is my pleasure, sir. I am here to learn about the internals of the library and most of the issues I picked so far felt like a small challenge. That would be more difficult without your **very** descriptive comments. So thank you!

I _may_ open another PR or two.
I will make sure to check discord more often.

evnchn:

**CI is green on 8c0b52eb** — `pylint`, `mypy`, `pre-commit`, `pytest` 3.10 and the Windows job all pass.

> _And who knows if the guard is only dead for today and may wake up later?_

That is the right question to ask, and the answer is the reason to remove it rather than to keep it as insurance: **if it woke up, it would wake up wrong.**

<details>
<summary><b>Why a dormant guard is worse than no guard here</b></summary>

Think about which way it fails. `if not context.slot_stack: report()` means *"only report when the stack is empty."* If some future change ever puts a slot on the stack at that point, the guard starts evaluating `False` and the code **stops reporting** — silently. That is #6233 coming back, introduced by the very line that looks like it is guarding against something.

So the guard is not dormant-but-safe. It is dormant-and-loaded: today it does nothing, and the day it does something, it breaks the fix. Compare with 6265, where the same construct was dead in the other direction — always true, always reporting, which is how the awaited path ended up firing `ui.on_exception` twice. Same shape, opposite symptom, both caused by asking a runtime signal for a fact the caller already knows.

That is the general form worth taking away: `context.slot_stack` answers *"am I currently inside a slot?"*, which is a **proxy** for the thing we actually care about — *"will anyone downstream see this exception?"* The proxy happens to line up in some places and not others, and when it stops lining up it does so silently. `report_exceptions=True` cannot go dead or wake up, because it is not a proxy; the call site states the fact directly.

There is also a cheap test for this, which is really an answer to your other point about testing before committing: **if you cannot write a test that makes a condition evaluate the other way, it has no behaviour to protect.** Try to write the test that makes `not context.slot_stack` false in `await_error()` — the attempt is what shows you it is scheduled into a fresh task every time, and at that point you have either found a real case (keep the guard, now with a test) or proved there is none (delete it). Either outcome is better than the guard sitting there unexamined. That is the same move that turned up the post-deletion masking row: I only found it because a reviewer asked me to justify a line I had copied without measuring.

</details>

> This PR is slightly heavier than the others because it tried to fix multiple things.

For what it is worth, that judgement is right and it is the more useful lesson than any of the individual findings. Two subsystems in one PR meant the `Event.call()` path had no owner in either half, and a green CI on both halves said nothing about it.

evnchn:

**Blocking: this PR introduces the exact bug it exists to fix. An async validation that raises after its client is gone loses the real exception on this branch — and does not lose it on `main`.**

The `except` block in `validate()` touches `self.client`, which raises once the client is collected, so the original error is replaced by `RuntimeError('The client this element belongs to has been deleted.')` before your `raise` ever runs.

<details>
<summary><b>Repro and the differential against main</b></summary>

The device is the sleep — the exception has to land after the client is actually collected (~0.115 s here), which is why shorter probes miss this entirely:

```python
async def test_async_validation_raises_after_client_is_really_gone(user, caplog):
    @ui.page('/')
    def page():
        async def check(value):
            await asyncio.sleep(1.5)              # land well after the client is collected

        field = ui.input('field', validation=check)
        ui.button('validate', on_click=lambda: field.validate(return_result=False))

    client = await user.open('/')
    user.find('validate').click()
    await asyncio.sleep(0.05)
    client.delete()
    user.client = None
    del client
    await asyncio.sleep(2.5)
```

| `validation_element.py` at | what reaches the log |
|---|---|
| `main` (932f019e) | `THE REAL ERROR` ✅ |
| this branch (8c0b52eb) | `The client this element belongs to has been deleted.` ⛔ |

On `main` there is no `except` at all, so the exception propagates to the background-task
done-callback and the global exception handler sees the real thing. Adding the `except` is what
creates the mask.
</details>

**The fix is one line, and it retires the dormant guard at the same time** — which answers your *"who knows if the guard is only dead for today"* the best possible way: replace it with a guard that is alive today.

```python
except Exception as e:
    if not self.is_deleted:
        self.client.handle_exception(e)
    raise
```

With that, `THE REAL ERROR` is visible again, and `tests/test_event.py` + `tests/test_input.py` are 33 passed / 0 skipped.

<details>
<summary><b>Separately: your four new tests are all genuine — I checked, including the two that pass on main</b></summary>

| test | branch | source reverted to main |
|---|---|---|
| `test_ui_on_exception_from_async_subscriber` | pass | **fail** — `saw 0, expected 1` |
| `test_ui_on_exception_from_async_validation` | pass | **fail** — `saw 0, expected 1` |
| `test_awaited_event_call_reports_once[False]` | pass | pass |
| `test_awaited_event_call_reports_once[True]` | pass | pass |

The last two pass on `main` by construction — they guard against *over*-reporting, which `main`
cannot do, so green-on-both proves nothing by itself. Mutation-tested them instead: injecting
`report_exceptions=True` into `_invoke_and_await` makes both fail (`saw 2, expected 1` and
`saw 1, expected 0`). The `caught=True` case is the one that matters, and it catches. Real guards.

And the measurement behind my dead-guard argument: instrumenting the `except`,
`SLOTSTACK_DEPTH=0`. `await_error` has exactly one call site — the `background_tasks.create(...)`
directly below it — and slot stacks are per-task, so it is dead by structure, not by timing.
</details>

**Note on sign-off:** I co-authored a commit here, so this is a comment, not an approval.

SaadZahem:
Could you please give me a test to run so I can see for myself?

nvm,

SaadZahem:
I applied the suggested change even though I ran the 4 tests again before and after and all passed.

evnchn (approved):

**Approved.** Both new tests fail without the fix and pass with it, the awaited path reports 1 uncaught / 0 caught, and CI is green. Nice work.

PR #6270, opened 2026-08-07
Fix async sub_pages handler doesn't report the real exception when the client is deleted

### LINKED ISSUE #6231: ui.sub_pages keeps building after Client.delete() and its error handler raises "The client this element belongs to has been deleted."

### First Check

### Example Code

```python
import asyncio

from nicegui import Client, app, ui


@app.on_exception
def report(e: Exception) -> None:
    print(f'>>> app.on_exception received: {type(e).__name__}: {e}', flush=True)


async def slow_sub_page() -> None:
    await asyncio.sleep(5)


@ui.page('/')
def index() -> None:
    ui.sub_pages({'/': slow_sub_page})


# NiceGUI already runs `app.timer(10, Client.prune_instances)` with a 60 s threshold.
# This only shortens the wait so the repro finishes in seconds instead of ~70 s.
@app.on_startup
async def prune_faster() -> None:
    while True:
        await asyncio.sleep(0.5)
        Client.prune_instances(client_age_threshold=2)


ui.run(reload=False)
```

### Description

**TL;DR:**

An async `ui.sub_pages` builder keeps running after its client has been deleted, and when it raises, its error handler dereferences a dead weakref.

`app.on_exception` receives `RuntimeError: The client this element belongs to has been deleted.` instead of the application's real error.

Same family as #6226 (a deferred callable resuming after `Client.delete()` and dereferencing a dead weakref), at a different site. Sibling of #6230, found in the same sweep.

**Steps** — no browser needed:

```bash
curl -s http://localhost:8080/ > /dev/null    # HTTP-only load, websocket never opens
```

then wait about 5 s.

**Expected**

```
>>> app.on_exception received: ValueError: THE REAL ERROR
```

**Actual**

```
>>> app.on_exception received: RuntimeError: The client this element belongs to has been deleted.
```

**Mechanism**

`Client.delete()` removes the elements and drops the instance, but it does not cancel `SubPages._active_tasks` — `_cancel_active_tasks` only runs on a route change, from `_show`. So the async builder task created in `_render_page` keeps running against a client that no longer exists.

When it then raises, the surrounding handler does `self.client.handle_exception(e)` (`nicegui/elements/sub_pages.py` line 125), and `Element.client` raises once the weakref is dead (`nicegui/element.py` line 138). The application's exception never reaches any registered handler; it survives only as chained `__context__` in the log.

The raise is the loud symptom, but the builder continuing to run at all after `Client.delete()` looks like the more general problem — a builder that completes normally will have spent its whole life building UI for a client that was already gone.

### NiceGUI Version

3.15.0

### Python Version

3.14.6

### Browser

Other

### Operating System

Linux

### Additional Context

<details>
<summary><b>Full traceback</b></summary>

```
The client this element belongs to has been deleted.
  File ".../nicegui/elements/sub_pages.py", line 123, in background_task
    await result
  File ".../mre_subpages.py", line 13, in slow_sub_page
ValueError: THE REAL ERROR

During handling of the above exception, another exception occurred:

    task.result()
    ~~~~~~~~~~~^^
  File ".../nicegui/elements/sub_pages.py", line 125, in background_task
    self.client.handle_exception(e)
    ^^^^^^^^^^^
  File ".../nicegui/element.py", line 138, in client
    raise RuntimeError('The client this element belongs to has been deleted.')
RuntimeError: The client this element belongs to has been deleted.
```

The second traceback contains no application frame at all.

</details>

<details>
<summary><b>Environments it was run in</b></summary>

| OS | Python | NiceGUI | Result |
|---|---|---|---|
| Linux (`python:3.14-slim` container) | 3.14.6 | 3.15.0 | reproduces |
| macOS 26 (arm64) | 3.14.2 | 3.15.0 | reproduces |

Linux run, verbatim:

```console
$ docker run --rm -v "$PWD":/src -w /src python:3.14-slim bash -c \
    "pip install -q nicegui==3.15.0 && (python mre_subpages.py > /tmp/out.log 2>&1 &) ; \
     sleep 14; python -c \"import urllib.request; urllib.request.urlopen('http://localhost:8123/').read()\"; \
     sleep 9; grep '>>>' /tmp/out.log"
>>> app.on_exception received: RuntimeError: The client this element belongs to has been deleted.
```

Run on a non-default port, otherwise the script above unchanged. Browser marked "Other" because none is involved — the repro is a single `curl`, which is also what makes the client eligible for `prune_instances` in the first place.

</details>

<details>
<summary><b>Why the repro accelerates pruning, and what that does and does not change</b></summary>

The `prune_faster` block only lowers the age threshold. NiceGUI's own `App.__init__` schedules `app.timer(10, Client.prune_instances)` and `prune_instances` defaults to `client_age_threshold=60.0`, so the identical failure occurs with the stock schedule if the builder's `asyncio.sleep` is longer than about 70 s. Shortening it keeps the repro under ten seconds and removes the timing race between the prune tick and the sleep. It calls the very same `Client.prune_instances`, though from a custom startup task rather than through NiceGUI's own `app.timer` schedule.

`Client.prune_instances` is the same mechanism named in #6226, and an HTTP-only page load — a bot, a prefetch, a health probe, a user who leaves before the handshake — is the natural way to create a client that gets pruned.

</details>

<details>

Found while sweeping for other instances of the shape reported in #6226. Put through an adversarial review pass by Codex (a different model lineage) before filing, which independently confirmed both that the deref is reachable and that `Client.delete()` never cancels `_active_tasks`.

</details>

SaadZahem:
### Motivation

Fixes #6231

### Implementation

Added an if statement checking if the client was deleted.
I couldn't do `self.client._deleted` since `self.client` in itself raises the `RuntimeError`.

```diff
@@ -122,7 +122,8 @@ class SubPages(Element, component='sub_pages.js', default_classes='nicegui-sub-p
                     try:
                         await result
                     except Exception as e:
-                        self.client.handle_exception(e)
+                        if not self.is_deleted:
+                            self.client.handle_exception(e)
                         raise

             task = background_tasks.create(background_task(), name=f'building sub_page {match.pattern}')
```

Tested using the example script provided in #6231.
I couldn't add an automated test since simulating a client teardown mid-`await` is awkward in the current setup.

### Progress

- [x] Pytests aren't feasible.


evnchn (dismissed):

**TL;DR:**

**Correct, and the guard is genuinely load-bearing — approving.** I reproduced the fix and checked the predicate is the right one rather than merely a working one. One non-blocking note about the test.

Your reasoning in the PR description is right: `self.client` is what raises, so the guard has to sit before it, not around it. And `is_deleted` is exactly the correct predicate here rather than a proxy for it — `Client.delete()` calls `remove_all_elements()`, which sets `element._deleted = True` on every element (`nicegui/client.py`), so the flag flips precisely when the client goes away.

1. **The missing test** (minor). The Progress checklist has the pytest box ticked, but the Implementation section says the manual repro rather than why a test is not needed — the template wants one or the other. A test *is* writable here; whether it is worth writing is a fair question, and I have put both sides below.

<details>
<summary><b>Verified: the fix works and the real error survives</b></summary>

A `user`-fixture version of #6231's repro — an async sub-page builder still running when its client is deleted:

| source | what `app.on_exception` receives |
|---|---|
| 1d0405a8 | `RuntimeError: The client this element belongs to has been deleted.` |
| this PR | **`ValueError: THE REAL ERROR`** |

The bare `raise` you kept is doing real work: it lets the exception reach `background_tasks._handle_exceptions`, so the app-level handlers still see the real error even though the client-scoped ones cannot. Dropping the `raise` would have traded one silent loss for another.

Gates on your branch: `pylint ./nicegui` 10.00/10 · `mypy ./nicegui` clean, 245 source files · `tests/test_sub_pages.py` 54 passed.

</details>

<details>
<summary><b>A test that does guard — and the honest argument against it</b></summary>

```python
async def test_deleted_client_does_not_mask_sub_page_exception(user: User, caplog: pytest.LogCaptureFixture):
    seen: list[Exception] = []
    app.on_exception(seen.append)
    building = asyncio.Event()

    async def slow_sub_page() -> None:
        building.set()
        await asyncio.sleep(0.2)

    @ui.page('/')
    def index() -> None:
        ui.sub_pages({'/': slow_sub_page})

    client = await user.open('/')
    await building.wait()
    client.delete()
    del client
    user.client = None
    gc.collect()
    await asyncio.sleep(0.4)
    caplog.records.clear()
    assert [str(e) for e in seen] == ['THE REAL ERROR']
```

Fails on 1d0405a8 with `['RuntimeError: The client this element belongs to has been deleted.']`, passes on your branch — so it is a real guard, not a green-only test.

**But** the three lines `del client` / `user.client = None` / `gc.collect()` are load-bearing, and that is the argument against it. I first wrote this without them and it passed on *both* branches — silently not reproducing, which is the worst kind of regression test. They are needed because the masking `RuntimeError` only fires once the `Client` object is actually collected and the weakref dies.

That makes it scaffolding built to observe a lifetime detail, which is what `AGENTS.md` warns about:

> If you catch yourself building scaffolding to observe an internal mechanism, stop — find the user-visible effect to assert on, or skip the test and give the reason in the pull request.

It would also need a PyPy caveat before merging, per `REVIEW.md`:

> Tests asserting CPython-only semantics — immediate weakref cleanup via refcounting, exact CPython error-message wording — need a PyPy `skipif` or a tolerant match

So: take it if you want the coverage, or leave it out and put one line in the Implementation section saying the deletion race needs refcount-level scaffolding to reproduce. Either satisfies the template. I would not hold the PR for it.

</details>

<details>
<summary><b>It does not overlap with 6230 — worth stating so nobody collapses them</b></summary>

The `contextlib.suppress` idea being discussed on 6230 would **not** cover this. There, the `RuntimeError` is raised *inside* `App.handle_exception`, from its own `context.client` lookup. Here it is raised in `sub_pages.py` at `self.client`, inside the `except` block, before `App.handle_exception` is on the stack at all — it escapes the background task and arrives at the app handler as the exception *being reported*. Two sites, two guards.

</details>

SaadZahem:
It wasn't clear to me if adding the test is worth it. I have nothing else to add so it's your call.

evnchn:

**My call as requested: skip the dedicated test — this is merge-ready as-is.** The guard is one self-evident line, and the identical `is_deleted`-guard pattern is getting regression coverage in #6258 and #6265, where staging the deleted-client race is natural. Staging it again through a full sub_pages navigation would buy little on top.

falkoschindler (changes requested):
Thanks for digging into #6231, @SaadZahem! The direction is right — swallowing the `RuntimeError` so the original exception survives — but the guard checks the wrong object, and the scenario the issue actually describes still isn't fixed.

1. **The guard checks the element, but the client is what raises** (blocking) — the `RuntimeError` comes from the dead _client_ weakref in `Element.client`, while `self.is_deleted` is about the _element_. The guard covers the issue scenario only indirectly (because `Client.delete()` also marks all its elements deleted), and it over-triggers: when just the element is deleted while its builder is still awaiting (e.g. a `container.clear()` or a refreshable rebuild) and the client is alive, the guard now skips `handle_exception`, so per-client `ui.on_exception` handlers stop firing for real user errors. I confirmed that regression by running such a script against `main` (handler fires) and this branch (it doesn't). Guarding on the actual condition avoids it and matches what `Element._is_safe_to_interact()` already does:

   ```python
                       except Exception as e:
                           client = self._client()
                           if client is not None and not client.is_deleted:
                               client.handle_exception(e)
                           raise
   ```

2. **The masking still happens for the scenario in the issue** (blocking) — this only unmasks exceptions raised _by_ the builder. Since `Client.delete()` doesn't cancel `SubPages._active_tasks`, a builder that just does `await asyncio.sleep(2)` and then creates an element still reports `RuntimeError: The client this element belongs to has been deleted.` to `app.on_exception` — now raised from `Element.__init__` → `context.client` instead. That's the mechanism #6231 names, and it survives this patch. Either the pending tasks need to be cancelled when the client is deleted, or the builder task has to stop early once its client is gone.

3. **Same latent masking in `app.handle_exception`** (minor, out of scope here) — [`nicegui/app/app.py`, line 178](https://github.com/zauberzeug/nicegui/blob/main/nicegui/app/app.py#L178) guards with `context.client is not None`, but `Context.client` never returns `None`; it raises the same `RuntimeError`. Worth a separate PR.

4. **Pytests** (minor) — the box is ticked, but there's no test: `test_exception_in_page_builder` in `tests/test_sub_pages.py` doesn't cover a deleted client. Simulating a client teardown mid-`await` is awkward in our test setup, so rather than forcing a test, please rewrite the line as

   ```
   - [x] Pytests are not feasible.
   ```

   with a sentence in **Implementation** saying why.

SaadZahem:
Thanks for your feedback. I will try my best to address all the items.

SaadZahem:
You should be able to find a fix for item 3 in #6272.
All the other items are applied

falkoschindler (approved):
Thanks @SaadZahem, this looks great now! I verified the update against the scenarios from my review: the issue-6231 MRE no longer reports anything (the pruned client's builder task is cancelled cleanly instead of raising the masking `RuntimeError`), and cancelling pending builders in `_handle_delete` nicely matches what navigation already does via `_cancel_active_tasks`. All sub_pages tests pass.

PR #6271, opened 2026-08-07
Fix button.clicked() never returns after Client.delete() and strands one task per abandoned page load #6232

### LINKED ISSUE #6232: button.clicked() never returns after Client.delete() and strands one task per abandoned page load (1.2 GB over 150k loads)

### First Check

### Example Code

```python
import asyncio

from nicegui import Client, app, background_tasks, ui


@ui.page('/')
def index() -> None:
    button = ui.button('go')

    async def waiter() -> None:
        await button.clicked()  # never returns once the client is deleted
    background_tasks.create(waiter(), name='waiter')


@app.on_startup
async def monitor() -> None:
    while True:
        await asyncio.sleep(5)
        Client.prune_instances(client_age_threshold=1)
        parked = [t for t in asyncio.all_tasks() if t.get_name() == 'waiter']
        print(f'>>> live clients: {len(Client.instances):>5} | parked waiters: {len(parked):>6}', flush=True)


ui.run(reload=False)
```

Drive it with HTTP-only page loads — no browser, no websocket:

```bash
for i in $(seq 1 4000); do curl -s -o /dev/null http://localhost:8080/ & done; wait
```

### Description

**TL;DR:**

`await button.clicked()` never returns and never gets cancelled when its client is deleted, so every abandoned page load permanently strands one task.

`Client.delete()` ends with `self._connected.set()  # for terminating connected() waits`, which wakes the `await self.client.connected()` inside `clicked()` into a **success** return — so it proceeds to `await event.wait()` for a click that can no longer arrive.

Measured: 150,000 HTTP-only page loads leave 148,545 parked tasks and grow RSS by **1,215 MB**, with no reclamation.

Same root cause as #6226 — `connected()` reporting success on a deleted client — but the symptom is an unbounded task leak rather than a raise, so it is filed separately. Sibling of #6230, found in the same sweep.

**Expected**

`parked waiters` returns to 0 as the clients are pruned, the way `live clients` does.

**Actual**

`parked waiters` equals the cumulative number of page loads and never decreases:

```
>>> live clients:   192 | parked waiters:    192
>>> live clients:   519 | parked waiters:   3079
>>> live clients:     0 | parked waiters:   4000     <- first 4000 loads done
>>> live clients:   623 | parked waiters:   5360
>>> live clients:     0 | parked waiters:   8000     <- second 4000 done
>>> live clients:     0 | parked waiters:  12000     <- third 4000 done
```

`live clients` does return to 0, so the `Client` objects themselves are released correctly. It is only the stranded tasks — and everything they hold, including the `ui.button` element — that accumulate.

**Mechanism**

`Button.clicked()` (`nicegui/elements/button.py` lines 62-67) registers a one-shot event, awaits `self.client.connected()`, then awaits that event. `Client.delete()` (`nicegui/client.py` line 472) sets `_connected`, which releases the `connected()` wait as a normal return rather than an error, and nothing afterwards observes that the client is gone. The second await has no timeout, no cancellation and no deletion hook, so the task parks for the lifetime of the process.

`background_tasks.create` keeps a strong reference in its module-level `running_tasks` set until the task completes, so a parked task is never garbage collected.

Three other methods have the same `await self.client.connected()` followed by `await self._initialized_event.wait()` shape and appear to share the defect, though I only measured the button:

- `ui.scene.initialized()` — `nicegui/elements/scene/scene.py` line 187
- `ui.scene_view.initialized()` — `nicegui/elements/scene/scene_view.py` line 79
- `ui.leaflet.initialized()` — `nicegui/elements/leaflet/leaflet.py` line 105

### NiceGUI Version

3.15.0

### Python Version

3.14.2

### Browser

Other

### Operating System

macOS

### Additional Context

<details>
<summary><b>Growth to 1.2 GB — the measured run</b></summary>

30 rounds of 5,000 HTTP-only page loads on released 3.15.0 / Python 3.14.2 / macOS 26 arm64. RSS is current resident size read from `ps -o rss=`, not `ru_maxrss` (which is a high-water mark and cannot show reclamation). A `gc.collect()` ran before every sample, so nothing here is waiting on the cycle collector.

| parked waiters | RSS above baseline |
|---:|---:|
| 160 | 10.9 MB |
| 3,382 | 173.1 MB |
| 6,430 | 254.6 MB |
| 64,747 | 658.6 MB |
| 96,523 | 854.1 MB |
| 140,435 | 1,163.5 MB |
| 148,545 | **1,215.4 MB** |

Steady-state marginal cost is about 6.9 KB per stranded task. The curve is linear with no plateau, and `live clients` returned to 0 at the end of the run while the 148,545 waiters remained.

</details>

<details>
<summary><b>Why the leak needs `background_tasks.create` to be visible</b></summary>

Worth stating because it changes what a reader will see if they vary the repro.

If the waiter is started with a plain `asyncio.create_task` and nothing holds the returned task, the stranded task forms a reference cycle — task holds coroutine holds the `asyncio.Event` whose `_waiters` holds the task's future — that is unreachable from outside, and the cycle collector reclaims it. The count then returns to 0 and the leak looks like it does not exist. It is a silent destruction of a pending task rather than correct cleanup, but it does not accumulate memory.

`background_tasks.create` is the documented way to start background work in NiceGUI, and it keeps a strong reference in `running_tasks` until completion. Since the task never completes, that reference is permanent, and the leak is real and unbounded.

I initially measured this the first way and briefly concluded there was no leak. Correcting it is what produced the numbers above.

</details>

<details>
<summary><b>What is not claimed</b></summary>

- Only `Button.clicked()` was measured. The three `initialized()` methods are listed on the basis of matching source shape, not a run.
- The measurement uses HTTP-only clients pruned by `prune_instances`, which is the cheapest way to create and destroy many clients. A client that connects and later disconnects reaches the same `await event.wait()` park by inspection, but I did not measure that path, since it needs a real browser per cycle.
- `client_age_threshold=1` only shortens the wait. NiceGUI already runs `app.timer(10, Client.prune_instances)` with a 60 s default, so the same accumulation happens on the stock schedule, just more slowly.

</details>

<details>

Found while sweeping for other instances of the shape reported in #6226. An adversarial review pass by Codex, a different model lineage, independently flagged the same four call sites from the source before any of this was measured.

</details>

SaadZahem:
### Motivation

Fixes #6232
Related #6277 

### Implementation

<details><summary>
By "linking" `Client._deleted_event` to the blocking event.<br>
<em>Linking event A to event B is my way of saying that setting A results in setting B (A --> B).</em>
</summary>

```diff
@@ -63,5 +64,6 @@ class Button(IconElement, TextElement, DisableableElement, BackgroundColorElemen
         """Wait until the button is clicked."""
         event = asyncio.Event()
         self.on('click', event.set, [])
+        helpers.link_events(self.client._deleted_event, event)
         await self.client.connected()
         await event.wait()
```
</details>

<details><summary>
This <strong>adds a new helper</strong>: (only in the first commit, it was reverted later on)<br>
Because this issue appears in other locations, it might be a good idea to have a helper.
<em>The name is open for discussion.</em>
</summary>

```
import asyncio

def link_events(source: asyncio.Event, target: asyncio.Event) -> asyncio.Task:
    async def _forward():
        await source.wait()
        target.set()
        
    return asyncio.create_task(_forward())
```
</details>

<details><summary>
Ran the example provided in #6232 again after the fix.<br>
<em>The number of parked waiters eventually decreased when my cpu had the power to process them</em>
</summary>

```
NiceGUI ready to go on http://localhost:8080, and http://192.168.1.13:8080
Event listeners changed after initial definition. Re-rendering affected elements.
>>> live clients:     1 | parked waiters:      1
>>> live clients:     1 | parked waiters:      1
>>> live clients:     0 | parked waiters:      0
>>> live clients:     0 | parked waiters:      0
>>> live clients:    73 | parked waiters:     84
>>> live clients:    43 | parked waiters:    267
>>> live clients:    69 | parked waiters:    121
>>> live clients:    15 | parked waiters:    261
>>> live clients:    45 | parked waiters:    238
>>> live clients:   130 | parked waiters:    581
>>> live clients:     0 | parked waiters:    536
>>> live clients:     0 | parked waiters:      0
```
</details>

This fix was only applied to `button.clicked` for now.
Other locations mentioned in #6232 (such as `ui.scene.initialized()`) weren't touched.

No pytests or documentation were touched .. yet.
Drafted because a pylint check failed.

### Progress


SaadZahem:
How should the pylint error be addressed?

1. Make `Client._deleted_event` public
2. Add a method to `Client` that registers a callback to run on deletion
3. Bake `helpers.link_events` functionality into `Client`
4. Disable the check for this line

evnchn:

**TL;DR:**

**Option 2 — and it already exists.** `Client.on_delete()` has been public and documented since 3.0.0, so no new helper, no new public attribute, and no `pylint` disable is needed:

```python
async def clicked(self) -> None:
    """Wait until the button is clicked."""
    event = asyncio.Event()
    self.on('click', event.set, [])
    self.client.on_delete(event.set)
    await self.client.connected()
    await event.wait()
```

That is the whole change — `nicegui/helpers/events.py` and its export come back out. **I have pushed it to your branch as b656bcd9** so CI goes green rather than leaving you to retype it; revert it if you would rather go another way, the PR is yours. It also adds the regression test, which fails on main with `clicked() never returned after the client was deleted` and needs no scaffolding to do so.

While measuring that, one finding worth more than the lint error:

1. **The forwarding task never finishes on a normal click** (major). A PR fixing *"strands one task per abandoned page load"* strands one task per **successful** `clicked()` instead — `_forward()` sits on `await source.wait()` for the life of the client. Measured below.

<details>
<summary><b>The stranded task, measured</b></summary>

Counting pending tasks whose coroutine qualname contains `_forward`, on a `user`-fixture page whose handler does `await button.clicked()`:

| scenario | before | after | task completed? |
|---|---|---|---|
| client deleted while waiting — the issue's case | 1 | 0 | yes, the fix works |
| button actually clicked | 1 | **1** | **no** |

On the click path `event.set()` is called by the click listener, `await event.wait()` returns, and `clicked()` exits — but `_forward()` is still parked on `await source.wait()`, waiting for a `_deleted_event` that will not fire until the client goes away. So the task outlives the call that created it.

With `Client.on_delete(event.set)` the count is **0 in both rows**: there is no forwarding task at all, just a callback on a list that the client already owns and already drops when it is deleted.

</details>

<details>
<summary><b>Why <code>on_delete</code> answers the lint error rather than working around it</b></summary>

`nicegui/client.py`:

```python
def on_delete(self, handler: Callable) -> None:
    """Add a callback to be invoked when the client is deleted.

    The callback can be synchronous or asynchronous and has an optional parameter of `nicegui.Client`.

    *Added in version 3.0.0*
    """
```

That is your option 2, already shipped — which is why the `protected-access` warning was worth listening to rather than silencing. `pylint` was pointing at a real thing: reaching for `_deleted_event` meant reimplementing a subscription mechanism the class already exposes.

Taking your four in turn:

- **1 — make `_deleted_event` public.** Enlarges the public API surface permanently to expose an `asyncio.Event`, and NiceGUI would then owe compatibility on it. `on_delete` covers the use case without that commitment.
- **2 — add a deletion-callback method.** Correct instinct; it exists.
- **3 — bake `link_events` into `Client`.** Same as 2 but with a new concept.
- **4 — disable the check.** The warning was right.

</details>

<details>
<summary><b>Also worth knowing for next time — <code>asyncio.create_task</code></b></summary>

`REVIEW.md` lists this explicitly:

> `asyncio.create_task()` instead of `background_tasks.create()` — the GC may drop unfinished tasks

`link_events` returns the `Task`, but the call site discards it, so nothing holds a strong reference and the event loop only keeps a weak one. `background_tasks.create()` exists to hold that reference. Moot if you take the `on_delete` route, since no task is created — but it is the kind of thing that bites in a way tests rarely catch.

</details>

<details>
<summary><b>One honest caveat on my own suggestion</b></summary>

`on_delete` appends to `client.delete_handlers`, so a long-lived client that calls `clicked()` many times accumulates one list entry per call. That is a much smaller leak than one pending `Task` per call, and it is bounded by the client's lifetime — but it is not zero. And it does not make `clicked()` registration-free either way: the existing `self.on('click', event.set, [])` on the line above already adds a click listener per call, unchanged from main. So this removes the new stranded task and the protected access; it does not claim to make repeated `clicked()` calls free.

Ordering checks out: `delete()` invokes `delete_handlers` *before* `_deleted_event.set()` and *before* `remove_all_elements()`, and `safe_invoke` calls zero-argument handlers correctly (`result = func(self) if len(inspect.signature(func).parameters) == 1 else func()`). `event.set()` only wakes the waiter, so the `clicked()` task resumes after `delete()` returns to the loop, same as with the PR's version.

</details>

<details>
<summary><b>Gates on the suggested version</b></summary>

Worktree at PR head 30f0ae1e, `uv sync --group dev --group types`, with `nicegui/helpers/events.py` deleted and the `link_events` export removed:

```
pylint ./nicegui   10.00/10
mypy ./nicegui     Success: no issues found in 245 source files
tests/test_button.py   4 passed
```

Both behaviour cases re-measured on that version: `clicked()` returns after `Client.delete()`, `clicked()` returns on a real click, zero `_forward` tasks pending in either.

</details>

The PR is yours — take this, adapt it, or push back if you see something I have missed.

SaadZahem:
This seems right. I will gladly keep your commit.

Today I learned:
1. `Client.on_delete`
2. Can't fix a stranded task by adding another stranded task
3. One `background_tasks.create` a day to keep the GC away

evnchn:

**CI is green on b656bcd9** — including the `pylint` job that was red before, which is the satisfying way for that particular question to get answered.

> 3. One `background_tasks.create` a day to keep the GC away

Framing it that way will serve you well, though the sharper version of the rule is *never hand a coroutine to `asyncio.create_task` and drop the handle* — the event loop only keeps a weak reference, so an unreferenced task can vanish mid-flight and take its exception with it. `background_tasks.create` exists precisely to hold that reference, and it also wires up the done-callback that reports failures to `app.on_exception`. Two of the bugs in this sweep were exceptions disappearing into unwatched tasks, so it is the same theme from a different angle.

And your number 2 is the one I would keep. It generalises past tasks: a fix that adds a resource to work around a resource you did not release is usually the wrong direction, and it is worth a second look whenever a diff *adds* something to solve a leak.

SaadZahem:
Looks like @falkoschindler was assigned in order to review a design decision.
The problem was already fixed without the need to add a helper.
I say let this PR be just about fixing the issue.
1. The helper function didn't prove useful for our case at the moment and probably won't be useful for our users.
2. If it proves useful, we could add it _in a separate PR_.

SaadZahem:
<details><summary>I instructed my agent to challenge the premise of this PR and it generated some test cases worth looking at:</summary>

```python
"""Skeptical probes for PR #6271 (Fix button.clicked() never returns after Client.delete()).

``button.clicked()`` now wakes any pending await from ``Button._handle_delete`` -- the same hook
PR #6250 uses for awaited dialogs -- so it resolves on client deletion AND on element/parent
deletion, without accumulating per-call handlers. Run with:

    pytest mytests/test_pr6271_gaps.py -v

Current expectation:
  * Button tests PASS: clicked() resolves on client deletion, parent deletion/clear, and direct
    button deletion, and no longer strands an on_delete handler per call.
  * The three ``initialized()`` tests FAIL: scene / scene_view / leaflet share the same defect and
    are intentionally left out of this fix (they need the identical _handle_delete treatment).
"""
import asyncio

from nicegui import ui


# --- Sanity: the case the PR DOES fix -----------------------------------------------------------

async def test_button_clicked_returns_after_delete(user: User):
    """Baseline: this is what the PR fixes. Should PASS."""
    returned = asyncio.Event()

    @ui.page('/')
    async def page():
        button = ui.button('Click me')
        await button.clicked()
        returned.set()

    client = await user.open('/')
    await asyncio.sleep(0.1)
    client.delete()
    await asyncio.sleep(0.1)
    assert returned.is_set()


# --- Gap 1: the three sibling methods with the identical shape are left unfixed ------------------
# Issue #6232 names ui.scene.initialized(), ui.scene_view.initialized() and ui.leaflet.initialized()
# as sharing the exact 'await connected(); await _initialized_event.wait()' defect. The PR touches
# only button.py, so these should still park forever after the client is deleted.

async def test_scene_initialized_returns_after_delete(user: User):
    """EXPECTED TO FAIL on this PR: scene.initialized() is never woken by delete."""
    returned = asyncio.Event()

    @ui.page('/')
    async def page():
        scene = ui.scene()
        await scene.initialized()
        returned.set()

    client = await user.open('/')
    await asyncio.sleep(0.1)
    client.delete()
    await asyncio.sleep(0.1)
    assert returned.is_set(), 'scene.initialized() never returned after the client was deleted'


async def test_scene_view_initialized_returns_after_delete(user: User):
    """EXPECTED TO FAIL on this PR: scene_view.initialized() is never woken by delete."""
    returned = asyncio.Event()

    @ui.page('/')
    async def page():
        scene = ui.scene()
        view = ui.scene_view(scene)
        await view.initialized()
        returned.set()

    client = await user.open('/')
    await asyncio.sleep(0.1)
    client.delete()
    await asyncio.sleep(0.1)
    assert returned.is_set(), 'scene_view.initialized() never returned after the client was deleted'


async def test_leaflet_initialized_returns_after_delete(user: User):
    """EXPECTED TO FAIL on this PR: leaflet.initialized() is never woken by delete."""
    returned = asyncio.Event()

    @ui.page('/')
    async def page():
        m = ui.leaflet()
        await m.initialized()
        returned.set()

    client = await user.open('/')
    await asyncio.sleep(0.1)
    client.delete()
    await asyncio.sleep(0.1)
    assert returned.is_set(), 'leaflet.initialized() never returned after the client was deleted'


# --- Gap 2 (now fixed): no per-call handler accumulation -----------------------------------------
# The earlier on_delete(event.set) approach appended a handler on every clicked() call and never
# removed it, so a long-lived client that awaits clicked() in a loop accumulated one dead handler
# per click. Waking from _handle_delete instead registers nothing on the client, so the count on
# client.delete_handlers no longer grows. (This probes an internal list on purpose, to pin the
# regression the on_delete approach would have introduced.)

async def test_clicked_does_not_leak_delete_handlers(user: User):
    """Each successful clicked() must not strand an on_delete handler on the client."""
    @ui.page('/')
    async def page():
        button = ui.button('Click me')

        async def loop():
            for _ in range(5):
                await button.clicked()
        ui.timer(0, loop, once=True)

    client = await user.open('/')
    baseline = len(client.delete_handlers)
    for _ in range(5):
        user.find('Click me').click()
        await asyncio.sleep(0.02)
    leaked = len(client.delete_handlers) - baseline
    assert leaked == 0, f'{leaked} on_delete handler(s) stranded after successful clicks'


# --- Gap 3 (now fixed): the button itself is removed while the client stays alive ----------------
# When the button element is removed on its own -- because a parent container is deleted/cleared, or
# the button is deleted directly -- the click can never arrive (the element is gone) and the client
# survives, so a client-only hook would never wake the waiter. Button._handle_delete now sets the
# pending events, so remove_elements() resolves the await regardless of what triggered the removal.
# These are the single-page analogue of #6232: a leak that would grow within one long-lived client.

async def test_clicked_returns_when_parent_container_deleted(user: User):
    """Deleting the button's parent must wake a pending clicked()."""
    returned = asyncio.Event()
    holder = {}

    @ui.page('/')
    async def page():
        with ui.card() as card:
            button = ui.button('Click me')
        holder['card'] = card
        await button.clicked()
        returned.set()

    await user.open('/')
    await asyncio.sleep(0.1)
    holder['card'].delete()
    await asyncio.sleep(0.1)
    assert returned.is_set(), 'clicked() never returned after its parent container was deleted'


async def test_clicked_returns_when_parent_container_cleared(user: User):
    """Clearing the button's parent must wake a pending clicked()."""
    returned = asyncio.Event()
    holder = {}

    @ui.page('/')
    async def page():
        with ui.card() as card:
            button = ui.button('Click me')
        holder['card'] = card
        await button.clicked()
        returned.set()

    await user.open('/')
    await asyncio.sleep(0.1)
    holder['card'].clear()
    await asyncio.sleep(0.1)
    assert returned.is_set(), 'clicked() never returned after its parent container was cleared'


async def test_clicked_returns_when_button_deleted(user: User):
    """Deleting the button itself must wake a pending clicked()."""
    returned = asyncio.Event()
    holder = {}

    @ui.page('/')
    async def page():
        button = ui.button('Click me')
        holder['button'] = button
        await button.clicked()
        returned.set()

    await user.open('/')
    await asyncio.sleep(0.1)
    holder['button'].delete()
    await asyncio.sleep(0.1)
    assert returned.is_set(), 'clicked() never returned after the button element was deleted'
```

</details>

1. It added test cases for `scene.initialized()`, `scene_view.initialized()`, and `leaflet.initialized()`<br>Those are similar issues that weren't addressed yet and are out of the scope of this PR
2. It added a test case that addresses the side effect of the initial fix, that is, it adds more handlers to `Client.delete_handlers`
3. Test case for when the parent element is deleted instead of client. While this is less likely to occur practically, it's worth adding in my opinion for the sake of completeness.

My last added commit addresses those issues except for the no.1

### Limitation of the test suite

Consider the example code snippet

```py
btn = ui.button("Submit")

await btn.clicked()
do_serious_action()
```

Should the code after `btn.clicked()` be ran in case of deletion of client/parent element?

SaadZahem:
The last commit makes the code after `await button.clicked()` never executes if the button/container/client is deleted/cleared.

falkoschindler:
No, the code after `await btn.clicked()` must not run — and your last commit gets that right.

It may look like it contradicts #6250, where a deleted dialog resolves the pending `await` instead. The difference: `await dialog` has a channel for "it didn't happen" — `None` already means "dismissed without submitting", the same as ESC. `clicked()` returns `None` unconditionally, so waking it is indistinguishable from a real click. Cancelling also has the nice property that `CancelledError` can't be swallowed by a surrounding `except Exception`, and it ends the task, which was the actual leak in #6232.

Same rule settles the cases you left out: `scene.initialized()`, `scene_view.initialized()` and `leaflet.initialized()` have no "didn't happen" channel either, so they should cancel too — out of scope here.

SaadZahem:
I was thinking of having `clicked` return `True` if the button was clicked and `False` if it was cancelled.

There is something else. What should happen if the cancelled task was an async page builder?

SaadZahem:
<details><summary>One more edge case: `await btn.clicked() inside async page builder`:</summary>

```py
"""Sharp edge for maintainer review: inline `await button.clicked()` in an async page builder.

Companion to ``test_pr6271_gaps.py``. The cancel-on-delete fix for ``button.clicked()`` is correct for
background waiters, but it has an inherent sharp edge when the await is written INLINE in an async
``@ui.page`` builder, e.g.:

    @ui.page('/')
    async def page():
        button = ui.button('Submit')
        ...
        await button.clicked()   # <-- inline, so the awaiting task IS the page-build task
        act_on_submission()

If the button (or its client) is deleted while this await is pending, ``clicked()`` cancels the awaiting task
to keep the post-click code from running for a click that never happened. But inline, that awaiting task is the
page-build task itself, so cancelling it leaves ``nicegui/page.py`` with a cancelled builder and no response:
``result = task.result()`` re-raises ``CancelledError``. Because ``CancelledError`` is a ``BaseException``, it also
bypasses NiceGUI's in-build error handler (``except Exception -> create_500_error_page``), so instead of a clean 500
the whole HTTP response is torn down with an opaque ``RuntimeError('No response returned.')`` -- even though the client
is perfectly healthy and only one element was removed. The two tests below pin this behavior and contrast it with an
ordinary in-builder exception.

There is a genuine conflict here, not just a bug to squash: "do not run the post-click continuation" and "keep
building/serving this very page" are mutually exclusive for an inline await, because the continuation *is* the rest
of the builder. Worth a maintainer decision on the desired outcome (e.g. surface a clear NiceGUI error / 500 page
rather than an opaque RuntimeError, and/or document that clicked() must be awaited from a background task, not inline).

    pytest mytests/new_tests.py -v
"""
import pytest

from nicegui import ui


def _consume_error_log(caplog: pytest.LogCaptureFixture, needle: str) -> None:
    """Drop the expected ERROR record(s) so the `user` fixture's no-error-logs guard still passes.

    Same idea as ``Screen.assert_py_logger``: the server-side ERROR log is expected here, so remove it in place
    (mutating the list the fixture inspects) after asserting it was actually emitted.
    """
    matched = [r for r in caplog.records if r.levelname == 'ERROR' and needle in r.getMessage()]
    assert matched, f'expected an ERROR log containing {needle!r}'
    caplog.records[:] = [r for r in caplog.records if r not in matched]


async def test_inline_await_of_deleted_button_tears_down_the_response(user: User, caplog: pytest.LogCaptureFixture):
    """SHARP EDGE: an inline `await button.clicked()` on an already-deleted button crashes the whole page render.

    The client is alive; only the button was removed. Yet the response is destroyed with an opaque
    RuntimeError('No response returned.') -- and logged as a hard server ERROR -- rather than a clean error page,
    because clicked() cancels the page-build task and the resulting CancelledError bypasses NiceGUI's
    Exception-based 500 handler. Contrast with ``test_inline_ordinary_exception_is_surfaced`` below.
    """
    @ui.page('/')
    async def page():
        button = ui.button('Submit')
        button.delete()
        await button.clicked()  # never completes; cancels the builder task
        ui.label('unreachable')  # a real page would run its post-submit code here

    with pytest.raises(RuntimeError, match='No response returned'):
        await user.open('/')
    _consume_error_log(caplog, 'No response returned')  # the crash is also logged server-side -- part of the symptom


async def test_inline_ordinary_exception_is_surfaced(user: User, caplog: pytest.LogCaptureFixture):
    """Contrast: a normal exception raised in the same builder is surfaced intact, not masked.

    Only here to make the sharp edge legible -- an ordinary in-builder failure comes through as its own
    ValueError, whereas the deleted-button case above is flattened into an opaque RuntimeError.
    """
    @ui.page('/')
    async def page():
        ui.button('Submit')
        raise ValueError('ordinary builder error')

    with pytest.raises(ValueError, match='ordinary builder error'):
        await user.open('/')
    _consume_error_log(caplog, 'ordinary builder error')

```

</details>

SaadZahem:
Inspired by the comment at #6277, I added the commit that covers the case of awaiting an already deleted button.

Right now it also cancels the pending task. But we can make it raise a friendly error instead.

falkoschindler:
Thanks @SaadZahem, the cancel semantics we settled on above holds up — a deep review confirmed the design, but surfaced three implementation bugs, all reproduced with A/B tests against `main`:

1. **A real click could be swallowed.** `_handle_delete` cancelled every tracked task, including one whose click had already arrived: `Event.set()` only schedules the waiter via `call_soon`, and `Task.cancel()` on a task whose future is already done still cancels it via `_must_cancel`. With an async click handler that deletes the button (or clears its container), `main` delivered the click and the PR cancelled it — the exact inverse of the bug being fixed.
2. **A failed construction could break `Client.delete()`.** `_click_tasks` was assigned after `super().__init__()`, which registers the element in `client.elements`. If a subclass constructor raised in between, the half-built button made `Client.delete()` abort with an `AttributeError`, permanently leaking the client and re-raising on every `prune_instances` cycle.
3. **The cancelled-page branch in `page.py` could serve a dead page.** When the cancellation came from client deletion, the branch served a 200 whose `client_id` is no longer in `Client.instances` — the handshake fails and the browser reload-loops, exactly what the `response_timeout` branch above it guards against (#6126).

I pushed two commits that fix all three and extract the mechanism into a small `CancelableWaitElement` mixin: it keeps a `dict[asyncio.Task, asyncio.Event]`, cancels only waits whose event is not yet set (fixing 1), initializes the dict before element registration (fixing 2), and replaces the `task.cancel()` + `await asyncio.sleep(0)` trick with an explicit `raise asyncio.CancelledError` (the sleep variant works on CPython 3.10–3.14, we verified, but the raise makes the control flow explicit). `page.py` now serves a proper error page when the client is gone (fixing 3). The tests follow the sentinel style of `test_dialog.py` and cover client deletion, element deletion, awaiting an already-deleted button, and the real-click-during-deletion case.

The mixin is also meant for #6294: `scene`, `scene_view` and `leaflet` can inherit `CancelableWaitElement` and wrap their waits in `with self._cancel_when_deleted(self._initialized_event):`, which shrinks that PR considerably and fixes the same three issues it inherited from the copied pattern.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Multi-agent%20review%20and%20code%20changes%20by%20the%20model%3B%20semantics%20decision%20and%20direction%20by%20the%20maintainer.&a=claude-code "claude-fable-5: Multi-agent review and code changes by the model; semantics decision and direction by the maintainer.")

falkoschindler (dismissed):

falkoschindler:
One more refinement after staring at the early-out for a while: awaiting an *already* deleted element used to `task.cancel()` **and** raise, which left `_must_cancel` pending. A caller that deliberately caught the `CancelledError` — say, to return a `RedirectResponse` instead of the half-built page — had its return value silently discarded, because CPython marks the task cancelled on the following `StopIteration`. The early-out now just raises, so both deletion paths (before and during the wait) deliver exactly one catchable `CancelledError`; a new regression test pins the catch-and-redirect case. While in there: the fresh-error-client recipe in `page.py` moved into a small `create_terminal_error_page()` helper (it existed twice, comment included), the two cancellation warnings folded into one, and the mixin now documents why it cancels while `ui.dialog` resolves with `None` — and that a *synchronous* click handler deleting its own button still loses that click (the click message reaches the deleted element too late; unchanged from main, where the wait hung forever instead).

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Multi-agent%20review%20and%20code%20changes%20by%20the%20model%3B%20semantics%20decision%20and%20direction%20by%20the%20maintainer.&a=claude-code "claude-fable-5: Multi-agent review and code changes by the model; semantics decision and direction by the maintainer.")

falkoschindler (approved):

PR #6272, opened 2026-08-08
Fix app.handle_exception replaces a real error with RuntimeError #6230

### LINKED ISSUE #6230: app.handle_exception raises "The parent element this slot belongs to has been deleted." and replaces the application's real exception

### First Check

### Example Code

```python
import asyncio

from nicegui import app, ui


@app.on_exception
def report(e: Exception) -> None:
    print(f'>>> app.on_exception received: {type(e).__name__}: {e}', flush=True)


@ui.page('/', reconnect_timeout=1.0)
def index() -> None:
    async def slow_handler() -> None:
        await asyncio.sleep(8)  # close the browser tab during this
    ui.button('start', on_click=slow_handler)


ui.run(reload=False)
```

### Description

**TL;DR:**

An async event handler that is still awaiting when its client is deleted loses its exception.

`app.on_exception` receives `RuntimeError: The parent element this slot belongs to has been deleted.` instead of the application's real error — `App.handle_exception` dereferences a dead weakref in its own first line.

Same family as #6226 (a deferred callable resuming after `Client.delete()` and dereferencing a dead weakref), at a different site. The same sweep also produced #6231 and #6232.

**Steps**

1. Open the page and click **start**.
2. Close the browser tab.
3. Wait for the 8 s sleep to elapse.

**Expected** — and this is exactly what happens if you *leave the tab open*, which is the one-line toggle that flips the bug:

```
>>> app.on_exception received: ValueError: THE REAL ERROR
```

**Actual**, when the tab is closed during the await:

```
>>> app.on_exception received: RuntimeError: The parent element this slot belongs to has been deleted.
```

The `ValueError` reaches no handler registered through `app.on_exception` or `ui.on_exception`. It survives only as chained `__context__` inside the logged traceback, so anything that keys on the raised exception — a Sentry integration, a custom `on_exception` router — sees the wrong error type, the wrong message and a stack with no application frame in it.

**Mechanism**

`events._await_and_handle_in_context` awaits the handler *inside* `with context:` and calls `core.app.handle_exception(e)` before leaving that context (`nicegui/events.py` lines 486-492). `App.handle_exception` opens with `if context.slot_stack and context.client is not None:` (`nicegui/app/app.py` line 178). `context.client` resolves `Slot.parent.client` (`nicegui/context.py` line 41), and `Slot.parent` raises when the weakref to its parent element is dead (`nicegui/slot.py` line 31).

Once `Client.delete()` has run — here through `reconnect_timeout` after the tab closes — the client's element tree is unreferenced, so the slot still sitting on that task's slot stack has a dead parent, and the exception funnel raises instead of reporting.

Two incidental observations from the same lines: in this branch `context.client is not None` cannot evaluate to `False`, because once `context.slot_stack` is truthy the property either returns a client or raises; and the failure needs no cycle collector, since the element tree here dies by refcount alone.

### NiceGUI Version

3.15.0

### Python Version

3.14.2

### Browser

Chrome

### Operating System

macOS

### Additional Context

<details>
<summary><b>Full traceback</b></summary>

```
The parent element this slot belongs to has been deleted.
  File ".../nicegui/events.py", line 490, in _await_and_handle_in_context
    await awaitable
  File ".../mre_handle_exception.py", line 15, in slow_handler
ValueError: THE REAL ERROR

During handling of the above exception, another exception occurred:

    task.result()
    ~~~~~~~~~~~^^
  File ".../nicegui/events.py", line 492, in _await_and_handle_in_context
    core.app.handle_exception(e)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^
  File ".../nicegui/app/app.py", line 178, in handle_exception
                              ^^^^^^^^^^^^^^
  File ".../nicegui/context.py", line 41, in client
    return self.slot.parent.client
           ^^^^^^^^^^^^^^^^
  File ".../nicegui/slot.py", line 31, in parent
    raise RuntimeError('The parent element this slot belongs to has been deleted.')
RuntimeError: The parent element this slot belongs to has been deleted.
```

The second traceback contains no application frame at all.

</details>

<details>
<summary><b>Works/broken differential, as run</b></summary>

Both rows are the script above (run on a non-default port, otherwise unchanged), same click, on released 3.15.0 / Python 3.14.2 / macOS 26 arm64, driven through a real Chromium tab:

| Run | Action after clicking **start** | What `app.on_exception` received |
|---|---|---|
| control | leave the tab open | `ValueError: THE REAL ERROR` |
| bug | close the tab | `RuntimeError: The parent element this slot belongs to has been deleted.` |

Closing the tab is the only difference between the two runs; it is what causes `Client.delete()` to run after `reconnect_timeout`.

</details>

<details>
<summary><b>The container-scoped variant — no client deletion needed</b></summary>

The client does not have to die. Deleting only the handler's *container* while the client stays alive produces the identical failure, because the dead weakref is `Slot._parent`, not the client reference:

```python
with ui.column() as col:
    ui.button('start', on_click=slow_handler)
...
col.delete()   # while slow_handler is still awaiting
```

This variant is refcount-deterministic — no `gc.collect()`, no cycle collector involved — which makes it the easier shape to build a regression test around. Verified 5 runs out of 5 through a separate `pytest` harness using the `User` fixture rather than the browser script above.

</details>

<details>
<summary><b>Reachability sweep: what else calls handle_exception from inside a slot context</b></summary>

I walked every `handle_exception` call site in 3.15.0 looking for the same shape. Only one other is live, and it is a distinct enough defect that it is filed separately rather than folded in here.

| Site | Inside a slot context when it calls `handle_exception`? | Verdict |
|---|---|---|
| `events.py:492` `_await_and_handle_in_context` | yes | **this issue** |
| `elements/sub_pages.py:125` | yes, and dereferences `Element.client` directly | filed as #6231 |
| `client.py:428` `safe_invoke` | no — `except` sits outside the `with` | safe |
| `events.py:483` `handle_event` outer except | no — outside the `with` | safe |
| `event.py:143` `_invoke_and_forget` | no — exception leaves `with` first | safe |
| `page.py:181` `wait_for_result` | holds a strong `client` local | safe |
| `outbox.py:131`, `outbox.py:135` | no slot context | safe |
| `background_tasks.py:156` `_handle_exceptions` | no — runs as a task done-callback, where `Slot.get_stack()` keys on task id 0 | safe; this is the site that reports the replacement exception |
| `app/app.py:111` `App.safe_invoke` | no — enters no slot context of its own | safe |
| `page.py:137` `check_for_late_return_value` | no — task done-callback | safe |
| `page.py:158` `create_500_error_page` | yes, but the `with Client(...)` is a freshly created client held by a strong local | safe |
| `timer.py:102`, `timer.py:115` | yes, but `_handle_delete` cancels the in-flight invocation first | could not reproduce |

Also checked and ruled out: `Outbox._emit` (`outbox.py:146`) looked like the same bug — it does `client = self.client` after `await core.sio.emit(...)`, and the `if client:` right after it suggests someone expected `None` rather than a raise. It is not reachable: `Outbox.loop` binds `client = self.client` at `outbox.py:94` *before* awaiting the emit coroutines, so the suspended loop frame holds the client strongly for the whole emit. I built a repro for it, watched it not fire, and then probed with `gc.get_referrers`, which reported `<coroutine object Outbox.loop>` as the referrer keeping the client alive.

`ui.refreshable` funnels into this same issue rather than being separate: `await_with_context` lets the exception leave the `with` before any handler runs.

</details>

<details>

Found while sweeping for other instances of the shape reported in #6226. Both findings were put through an adversarial review pass by Codex (a different model lineage) before filing; it confirmed the reachability and corrected one overstatement of mine — the original exception is not erased from the log, it survives as chained context, and it is the *handler-visible* exception that is replaced. The wording above reflects that correction.

</details>

SaadZahem:
### Motivation

Fixes #6230

### Implementation

`App.handle_exception` used to access `context.client` in its first line. Once the client has been deleted, that raises `RuntimeError('The parent element this slot belongs to has been deleted')` and replaces the application's real exception before any handler runs.

The client lookup is now guarded with `Slot.get_stack()` instead of `context.slot_stack` (whose getter would enter script mode when called before startup), and only the `context.client` access is wrapped in `contextlib.suppress(RuntimeError)`. `RuntimeError`s raised from exception handlers themselves are still surfaced.

Not covered here: a handler that raises (e.g. `ui.notify` after the client has been deleted) still skips the remaining handlers. That is a pre-existing issue independent of this fix and will be addressed in a follow-up PR.

A test wasn't added because (quoting from a comment below)

> a test here needs del plus gc.collect() to force the weakref to die, which is the scaffolding AGENTS.md steers away from

Ready to merge.

### Progress

### DIFF
--- a/nicegui/app/app.py
+++ b/nicegui/app/app.py
@@ -85,7 +85,7 @@ class App(FastAPI):
             self.safe_invoke(t)
         self.on_shutdown(self.storage.on_shutdown)
         self.on_shutdown(background_tasks.teardown)
-        background_tasks.create(binding.refresh_loop(), name='refresh bindings')
+        background_tasks.create(binding.refresh_loop(), name="refresh bindings")
         self.timer(10, Client.prune_instances)
         self.timer(10, Slot.prune_stacks)
         self.timer(10, prune_tab_storage)
@@ -107,7 +107,7 @@ class App(FastAPI):
         try:
             result = func()
             if helpers.should_await(result):
-                background_tasks.create(result, name=f'func {func.__name__ if hasattr(func, "__name__") else func}')
+                background_tasks.create(result, name=f"func {func.__name__ if hasattr(func, '__name__') else func}")
         except Exception as e:
             self.handle_exception(e)
 
@@ -118,7 +118,7 @@ class App(FastAPI):
         """
         if core.is_script_mode_re_execution():
             return  # already registered on the script's first execution
-        self._connect_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_connect()'))
+        self._connect_handlers.append(helpers.normalize_lifecycle_handler(handler, "app.on_connect()"))
 
     def on_disconnect(self, handler: Callable) -> None:
         """Called every time a new client disconnects from NiceGUI.
@@ -129,7 +129,7 @@ class App(FastAPI):
         """
         if core.is_script_mode_re_execution():
             return  # already registered on the script's first execution
-        self._disconnect_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_disconnect()'))
+        self._disconnect_handlers.append(helpers.normalize_lifecycle_handler(handler, "app.on_disconnect()"))
 
     def on_delete(self, handler: Callable) -> None:
         """Called when a client is deleted.
@@ -140,7 +140,7 @@ class App(FastAPI):
         """
         if core.is_script_mode_re_execution():
             return  # already registered on the script's first execution
-        self._delete_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_delete()'))
+        self._delete_handlers.append(helpers.normalize_lifecycle_handler(handler, "app.on_delete()"))
 
     def on_startup(self, handler: Callable) -> None:
         """Called when NiceGUI is started or restarted.
@@ -152,8 +152,8 @@ class App(FastAPI):
         if core.is_script_mode_re_execution():
             return  # already registered on the script's first execution
         if self.is_started:
-            raise RuntimeError('Unable to register another startup handler. NiceGUI has already been started.')
-        self._startup_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_startup()'))
+            raise RuntimeError("Unable to register another startup handler. NiceGUI has already been started.")
+        self._startup_handlers.append(helpers.normalize_lifecycle_handler(handler, "app.on_startup()"))
 
     def on_shutdown(self, handler: Callable) -> None:
         """Called when NiceGUI is shut down or restarted.
@@ -163,7 +163,7 @@ class App(FastAPI):
         """
         if core.is_script_mode_re_execution():
             return  # already registered on the script's first execution
-        self._shutdown_handlers.append(helpers.normalize_lifecycle_handler(handler, 'app.on_shutdown()'))
+        self._shutdown_handlers.append(helpers.normalize_lifecycle_handler(handler, "app.on_shutdown()"))
 
     def on_exception(self, handler: Callable) -> None:
         """Called when an exception occurs.
@@ -181,12 +181,12 @@ class App(FastAPI):
             if context.slot_stack:
                 client = context.client
         if client is not None:
-            context.client.handle_exception(exception)
+            client.handle_exception(exception)
 
         for handler in self._exception_handlers:
             result = handler() if not inspect.signature(handler).parameters else handler(exception)
             if helpers.should_await(result):
-                background_tasks.create(result, name=f'exception {handler.__name__}')
+                background_tasks.create(result, name=f"exception {handler.__name__}")
 
     def on_page_exception(self, handler: Callable) -> None:
         """Called when an exception occurs in a page and allows to create a custom error page.
@@ -207,16 +207,13 @@ class App(FastAPI):
         if self.native.main_window:
             self.native.main_window.destroy()
         if self.config.reload or Server.instance.config.should_reload:
-            os.kill(os.getppid(), getattr(signal, 'CTRL_C_EVENT' if platform.system() == 'Windows' else 'SIGINT'))
+            os.kill(os.getppid(), getattr(signal, "CTRL_C_EVENT" if platform.system() == "Windows" else "SIGINT"))
         else:
             Server.instance.should_exit = True
 
-    def add_static_files(self,
-                         url_path: str,
-                         local_directory: str | Path,
-                         *,
-                         follow_symlink: bool = False,
-                         max_cache_age: int = 3600) -> None:
+    def add_static_files(
+        self, url_path: str, local_directory: str | Path, *, follow_symlink: bool = False, max_cache_age: int = 3600
+    ) -> None:
         """Add a directory of static files.
 
         `add_static_files()` makes a local directory available at the specified endpoint, e.g. `'/static'`.
@@ -232,24 +229,28 @@ class App(FastAPI):
         :param follow_symlink: whether to follow symlinks (default: False)
         :param max_cache_age: value for max-age set in Cache-Control header (*added in version 2.8.0*)
         """
-        if url_path == '/':
-            raise ValueError('''Path cannot be "/", because it would hide NiceGUI's internal "/_nicegui" route.''')
+        if url_path == "/":
+            raise ValueError("""Path cannot be "/", because it would hide NiceGUI's internal "/_nicegui" route.""")
         if max_cache_age < 0:
-            raise ValueError('''Value of max_cache_age must be a positive integer or 0.''')
+            raise ValueError("""Value of max_cache_age must be a positive integer or 0.""")
 
         handler = CacheControlledStaticFiles(
-            directory=local_directory, follow_symlink=follow_symlink, max_cache_age=max_cache_age)
+            directory=local_directory, follow_symlink=follow_symlink, max_cache_age=max_cache_age
+        )
 
-        @self.get(url_path.rstrip('/') + '/{path:path}')  # prevent double slashes in route pattern
-        async def static_file(request: Request, path: str = '') -> Response:
+        @self.get(url_path.rstrip("/") + "/{path:path}")  # prevent double slashes in route pattern
+        async def static_file(request: Request, path: str = "") -> Response:
             return await handler.get_response(path, request.scope)
 
-    def add_static_file(self, *,
-                        local_file: str | Path,
-                        url_path: str | None = None,
-                        single_use: bool = False,
-                        strict: bool = True,
-                        max_cache_age: int = 3600) -> str:
+    def add_static_file(
+        self,
+        *,
+        local_file: str | Path,
+        url_path: str | None = None,
+        single_use: bool = False,
+        strict: bool = True,
+        max_cache_age: int = 3600,
+    ) -> str:
         """Add a single static file.
 
         Allows a local file to be accessed online with enabled caching.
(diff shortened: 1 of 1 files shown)


evnchn (dismissed):

**TL;DR:**

**Both properties hold — I verified them rather than trusting the description — approving.** One free nit, no need to hold the PR for it.

| | main | this PR |
|---|---|---|
| the issue's repro — `app.on_exception` receives | `RuntimeError: The parent element…` | **`ValueError: THE REAL ERROR`** ✅ |
| a `ui.on_exception` handler raising `RuntimeError` | propagates loudly | **propagates loudly** ✅ |

That second row is the one that matters, and it is exactly what your Implementation section claims: narrowing the `suppress` to the lookup keeps handler bugs visible. `pylint` 10.00/10 · `mypy` clean, 245 source files · `tests/test_events.py tests/test_lifecycle.py` 19 passed.

1. **`context.client` is walked a second time, outside the guard** (minor, free to fix).

<details>
<summary><b>The nit — and an honest note that I could not make it fail</b></summary>

```python
client = None
with contextlib.suppress(RuntimeError):
    if context.slot_stack:
        client = context.client
if client is not None:
    context.client.handle_exception(exception)   # <- walks it again, unsuppressed
```

`context.client` is `slot.parent.client` — two weakref hops. Binding `client` keeps the **Client** alive, but nothing keeps the **element** alive, and that is the hop that raises. So the second walk is outside the protection the first one got, for no benefit:

```diff
 if client is not None:
-    context.client.handle_exception(exception)
+    client.handle_exception(exception)
```

Being straight with you about the strength of this: I could **not** construct a case where it actually raises. There is no `await` between the two lines and nothing drops a reference in between, so in practice the second walk succeeds whenever the first did. Reducing the mechanism to plain objects does show the shape is real —

```
PROBE: second lookup RAISED -> The parent element this slot belongs to has been deleted.
PROBE: bound client still usable = 'CLIENT'
```

— but that is a model, not your code. Treat it as "the bound value is free, so use it" rather than "there is a bug here."

</details>

<details>
<summary><b>Two paperwork things</b></summary>

- The Progress checklist ticks the pytest box, but the Implementation section does not say why a test is not needed. Same situation as 6270: a test here needs `del` plus `gc.collect()` to force the weakref to die, which is the scaffolding `AGENTS.md` steers away from — so *that sentence* is the thing to add, not the test. One line is enough.
- The message quoted in Implementation is `RuntimeError('The client this slot belongs to has been deleted')`; the string that actually fires on this path is `The parent element this slot belongs to has been deleted.` (from `Slot.parent`). Worth correcting since the PR body becomes the merge record — there is a *separate* `The client this element belongs to has been deleted.` from `Element.client`, which is the one 6270 is about, and conflating them makes the two guards look redundant when they are not.

</details>

Fast turnaround on this one.

SaadZahem:
>     context.client.handle_exception(exception)   # <- walks it again, unsuppressed
I didn't indent to call the property again while writing the code. My bad.

evnchn:

**Heads-up before CI comes back: the latest commit reformatted the whole of `nicegui/app/app.py`, and `check / pre-commit` will fail on it.** The actual fix in it is correct — `client.handle_exception(exception)` is exactly right, thank you — but it arrived with about 80 lines of unrelated reformatting, almost certainly an editor running Black or `ruff format` with default settings on save.

I ran the repo's own hooks against your head rather than guess:

```
$ pre-commit run --files nicegui/app/app.py
ruff check...............................................................Failed
- hook id: ruff-check
- files were modified by this hook
Found 52 errors (52 fixed, 0 remaining).
```

<details>
<summary><b>What it changed, and why <code>pre-commit run</code> alone will not clean it up</b></summary>

Two kinds of drift, and they behave differently:

1. **Quotes** — `'refresh bindings'` became `"refresh bindings"` throughout. NiceGUI is single-quote; `ruff check --fix` reverts these for you, so this half is automatic.
2. **Signature re-wrapping** — the repo's aligned-continuation style got collapsed to Black's bracket style:

   ```diff
   -    def add_static_files(self,
   -                         url_path: str,
   ...
   +    def add_static_files(
   +        self, url_path: str, local_directory: str | Path, *, follow_symlink: bool = False, max_cache_age: int = 3600
   +    ) -> None:
   ```

   `ruff check --fix` does **not** revert this. After running the hooks I still measured ~83 changed lines against `main` in a file whose real change is six.

So the quickest way back to a clean diff is to start from the file as it is on `main` and re-apply just your change:

```bash
git checkout upstream/main -- nicegui/app/app.py
# re-add `import contextlib` and the six-line handle_exception hunk
pre-commit run --files nicegui/app/app.py
```

Worth turning format-on-save off for this repo, or pointing it at the project's `pyproject.toml` — the same thing will happen on every file you touch otherwise, and it is the kind of noise that makes a two-line fix look risky to a maintainer.

</details>

Nothing wrong with the substance — this is purely mechanical. Ping me when it is pushed and I will re-approve.

SaadZahem:
I ran `git restore nicegui --source <tag>` and reimplemented the change

SaadZahem:
Any idea on how to configure neovim formatter to stick to project settings?

evnchn (dismissed):

**Re-approving — the re-push is exactly the minimal shape, the fix is intact, and CI is green on 44c7c45b.** (GitHub auto-dismissed my earlier approval when the branch moved; this restores it.)

> Any idea on how to configure neovim formatter to stick to project settings?

**The key fact: this repo has no all-file formatter — no Black, no `ruff format`.** Formatting is autopep8 (`--max-line-length=120`, config in `[tool.autopep8]` in pyproject.toml) plus `ruff check --fix` for *lint* fixes only, both run via pre-commit. Any formatter that reflows whole files will fight the repo's single-quote, 120-column style — which is what happened with the earlier push.

<details>
<summary><b>Practical neovim setups</b></summary>

- **conform.nvim:** set the Python formatter to `autopep8` for this project (it reads `pyproject.toml` when invoked from the repo root). A project-local override works well — `.nvim.lua` with `exrc`, or an autocmd setting `vim.b.formatters_by_ft` for this path.
- **Or simplest:** disable format-on-save for this repo and let the hooks do the work — `pre-commit run --files <file>`, or just commit and let the hooks fix in place.
- If ruff runs as your LSP, its code-action *lint* fixes are fine (that's `ruff check --fix`) — it's `ruff format` you want off.
</details>

falkoschindler (approved):
The fix is correct for the reported symptom and strictly improves every call site. Before, the stale-slot `RuntimeError` escaped before any handler ran; now all handlers receive the original exception (verified empirically). The removed `context.client is not None` check was dead code, since the property raises instead of returning `None`. Skipping a regression test is justified per AGENTS.md.

I pushed a small follow-up commit that guards with `Slot.get_stack()` instead of `context.slot_stack` (the latter is a side-effecting getter: with an empty stack before startup it flips `core.script_mode` and constructs a pseudo client, reachable from a raising `app.on_startup` handler) and narrows the `suppress` to the `context.client` access, with a comment on which `RuntimeError`s are expected there. Same idiom as in `event.py` and `style.py`. I also updated the PR description to match.

One thing this PR deliberately leaves open, as the description says: a handler that raises itself (e.g. `ui.notify` after the client has been deleted, or from a background task without a slot) still skips the remaining handlers, and its error bounces back into `handle_exception` via the background task's done callback. That is a pre-existing issue independent of this fix and reproducible on `main` today, so it gets its own PR. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20verified%20and%20worded%20by%20the%20model%3B%20the%20decision%20to%20split%20the%20follow-up%20into%20its%20own%20PR%20is%20the%20maintainer%27s.&a=claude-code "claude-fable-5-1: Findings verified and worded by the model; the decision to split the follow-up into its own PR is the maintainer's.")

PR #6275, opened 2026-08-09
Fix `table.pagination` setter leaving the pagination controls hidden

### LINKED ISSUE #6235: Setting `table.pagination` after construction paginates the rows but leaves the pagination controls hidden
### First Check

### Example Code

```python
from nicegui import ui

rows = [{'id': i} for i in range(20)]

ui.label('A: ui.table(rows=rows, pagination=5)')
a = ui.table(rows=rows, pagination=5).classes('w-64')

ui.label('B: t = ui.table(rows=rows); t.pagination = {"rowsPerPage": 5}')
t = ui.table(rows=rows).classes('w-64')
t.pagination = {'rowsPerPage': 5}          # <-- the only difference

# visible in the browser; also checkable without one:
for label, table in [('A', a), ('B', t)]:
    print(label, {k: table._props[k] for k in ('pagination', 'hide-pagination')})

ui.run()
```

### Description

**TL;DR:**

Both tables above paginate to 5 rows. Only table A gets a pagination bar.

Table B shows 5 of its 20 rows with **no controls to reach the other 15** — the constructor derives `hide-pagination` from the `pagination` argument, but the property setter never updates it.

There is a workaround (`t.props('hide-pagination=false')`), so this is surprising rather than fatal.

<details>
<summary><b>Cause</b></summary>

The constructor derives a second prop — `nicegui/elements/table.py:75-76` on 3.15.0:

```python
self._props['hide-pagination'] = pagination is None
self._props['pagination'] = pagination if isinstance(pagination, dict) else {'rowsPerPage': pagination or 0}
```

The setter writes only `pagination`, leaving `hide-pagination` untouched, and does not apply the `int` → `{'rowsPerPage': n}` normalisation either (`table.py:396-397`):

```python
@pagination.setter
def pagination(self, value: dict) -> None:
    self._props['pagination'] = value
```

A table built without `pagination` starts at `hide-pagination = True`, and nothing ever clears it.

</details>

<details>
<summary><b>Measured in a real browser (Chromium via Playwright, nicegui 3.15.0 from PyPI)</b></summary>

20 rows, `rowsPerPage = 5`, counting `tbody tr` and testing `.q-table__bottom` for visibility:

| build path | body rows rendered | pagination bar visible |
|---|---|---|
| `ui.table(rows=rows, pagination=5)` | 5 | **True** |
| `t = ui.table(rows=rows); t.pagination = {'rowsPerPage': 5}` | 5 | **False** |
| `…` then `t.props('hide-pagination=false')` | 5 | **True** (workaround) |

The example above also prints this on startup, so the divergence is checkable without a browser:

```
A {'pagination': {'rowsPerPage': 5}, 'hide-pagination': False}
B {'pagination': {'rowsPerPage': 5}, 'hide-pagination': True}
```

A screenshot of the two tables side by side is available if useful.

</details>

<details>
<summary><b>Scope of the sweep this came from</b></summary>

Found by comparing, for every `ui.*` element, the state produced by `ui.X(param=v)` against `ui.X(); x.param = v`. 42 constructor parameters across all element classes have a same-named settable property; 36 were exercised (1 has no usable test value, 5 need mutually-required constructor arguments).

`table.pagination` was the only pair where the two write paths produce a **behavioural** difference. The other divergence found was cosmetic and is not reported: the `validation` setter leaves `error-message: None` in the props where the constructor leaves the key absent, which is equivalent for Quasar.

</details>

### NiceGUI Version

3.15.0

### Python Version

3.12.11

### Browser

Chrome

### Operating System

macOS

### Additional Context

The setter is annotated `value: dict`, so passing an `int` is already out of contract and I have not treated that as part of the bug — the report is about the `dict` case, which is in contract and still wrong.

SaadZahem:
*Filed by Claude Code on SaadZahem's behalf.*

### Motivation

Fixes #6235

The constructor derives **two** props from the single `pagination` argument:

```python
self._props['hide-pagination'] = pagination is None
self._props['pagination'] = pagination if isinstance(pagination, dict) else {'rowsPerPage': pagination or 0}
```

The property setter only wrote `pagination`, so `hide-pagination` was never updated (and the `int → {'rowsPerPage': n}` normalization was skipped too). Because a table built without a `pagination` argument starts at `hide-pagination = True`, assigning `table.pagination = {'rowsPerPage': 5}` afterwards paginated the rows but left the controls hidden — showing 5 of 20 rows with no way to reach the rest. The two write paths that should agree, didn't.

### Implementation

The setter now mirrors the constructor: it sets `hide-pagination` and applies the same normalization, with the accepted type widened to `int | dict | None` to match the constructor parameter. The constructor delegates to the setter (`self.pagination = pagination`) so the derivation lives in exactly one place and the two paths can no longer drift apart.

As a result, `table.pagination = 5` and `table.pagination = None` (hide) now behave identically to passing the same value to the constructor.
The browser-driven `update:pagination` handler already assigns a `dict`, so its behavior is unchanged.

Note that `table.pagination = None` previously stored `None` which was different from how the constructor behaved. With this PR, it will now store `{'rowsPerPage': 0}`. While it counts as an improvement, **it is a public behavior change worth noting.**

### Progress

- [x] Pytests have been added: `test_pagination_via_setter` asserts the pagination bar appears when `pagination` is set after construction.

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -62,6 +62,19 @@ def test_pagination_dict(screen: Screen):
     screen.should_contain('1-2 of 3')
 
 
+def test_pagination_via_setter(screen: Screen):
+    @ui.page('/')
+    def page():
+        table = ui.table(columns=columns(), rows=rows())
+        table.pagination = {'rowsPerPage': 2}
+
+    screen.open('/')
+    screen.should_contain('Alice')
+    screen.should_contain('Bob')
+    screen.should_not_contain('Lionel')
+    screen.should_contain('1-2 of 3')
+
+
 def test_filter(screen: Screen):
     @ui.page('/')
     def page():
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -75,6 +75,16 @@ def test_pagination_via_setter(screen: Screen):
     screen.should_contain('1-2 of 3')
 
 
+def test_default_table_hides_pagination(screen: Screen):
+    @ui.page('/')
+    def page():
+        ui.table(columns=columns(), rows=rows())
+
+    screen.open('/')
+    screen.should_contain('Alice')
+    screen.should_not_contain('Records per page')
+
+
 def test_filter(screen: Screen):
     @ui.page('/')
     def page():
--- a/nicegui/elements/table.py
+++ b/nicegui/elements/table.py
@@ -72,8 +72,7 @@ class Table(FilterElement, component='table.js'):
         self._props['rows'] = rows
         self._props['row-key'] = row_key
         self._props.set_optional('title', title)
-        self._props['hide-pagination'] = pagination is None
-        self._props['pagination'] = pagination if isinstance(pagination, dict) else {'rowsPerPage': pagination or 0}
+        self.pagination = pagination
         self._props['selection'] = selection or 'none'
         self._props['selected'] = []
         self._props['fullscreen'] = False
@@ -394,8 +393,9 @@ class Table(FilterElement, component='table.js'):
         return self._props['pagination']
 
     @pagination.setter
-    def pagination(self, value: dict) -> None:
-        self._props['pagination'] = value
+    def pagination(self, value: int | dict | None) -> None:
+        self._props['hide-pagination'] = value is None
+        self._props['pagination'] = value if isinstance(value, dict) else {'rowsPerPage': value or 0}
 
     @property
     def is_fullscreen(self) -> bool:
--- a/nicegui/elements/table.py
+++ b/nicegui/elements/table.py
@@ -93,7 +93,7 @@ class Table(FilterElement, component='table.js'):
 
         def handle_pagination_change(e: GenericEventArguments) -> None:
             previous_value = self.pagination
-            self.pagination = e.args
+            self._props['pagination'] = e.args
             arguments = ValueChangeEventArguments(sender=self, client=self.client,
                                                   value=self.pagination, previous_value=previous_value)
             for handler in self._pagination_change_handlers:


SaadZahem:
I have verified the code personally.
The documentation doesn't mention the pagination as a property.
Therefore, updating the documentation is not necessary.

evnchn:

**Verified** — the fix is right and the test genuinely guards it. One behaviour change that reads as worth a line in the description.

<details>
<summary><b>Evidence, and the behaviour change</b></summary>

Head `eddf543e`, macOS, `tests/test_table.py`: **28 passed**. Reverting `nicegui/elements/table.py` to the merge base makes the new test fail with `AssertionError: Could not find "1-2 of 3"` — the hidden controls, exactly the reported symptom.

Routing `__init__` through the setter also changes what the setter *stores*:

| call | `_props['pagination']` | `hide-pagination` |
|---|---|---|
| `ui.table(..., pagination=2)` | `{'rowsPerPage': 2}` | `False` |
| `t.pagination = None` | `{'rowsPerPage': 0}` | `True` |
| `t.pagination = {'rowsPerPage': 5}` | `{'rowsPerPage': 5}` | `False` |

Previously the setter only wrote `_props['pagination']` and never touched `hide-pagination`, so `t.pagination = None` stored `None` verbatim — and on a table that had been showing pagination, left the controls up. The new behaviour is the better one — it makes the getter's declared `-> dict` actually true — but it is a public-behaviour change, so it seems worth naming under **Implementation** rather than leaving for someone to discover.

</details>

SaadZahem:
I also changed the code in the constructor so that we don't have to maintain both copies.

evnchn:

**That's already in eddf543e, the commit I reviewed — no code change needed.** The constructor delegating to the setter is the part I called the better one; there's no second copy left.

My note was only about the PR description: `t.pagination = None` used to store `None`, and now stores `{'rowsPerPage': 0}`. That's an improvement, but it's a public behaviour change, so it's worth one sentence under **Implementation**.

SaadZahem:
Thanks for the highlight. I have added the note to the PR message.

evnchn:
Interesting one — on a **default** `ui.table()` (no `pagination` argument) this makes the `Records per page: All ▾ 1-3 of 3` footer appear as soon as the client connects.

The path is `table.py:94-101`: Quasar emits `update:pagination` on mount, `e.args` is always a `dict`, so `value is None` is `False` and `hide-pagination` gets cleared on every connect. The PR body's "the handler already assigns a `dict`, so its behavior is unchanged" holds for `pagination`, but not for `hide-pagination` — which that handler never meant to touch.

Keeping the mount-driven handler out of the visibility decision is the smallest fix; a private `_set_pagination(value, update_visibility=...)`, or deriving `hide-pagination` in `_to_dict`, would work too — Falko's call.

(That withdraws my earlier **Verified**, sorry — I checked the fix and the revert, but never rendered a default table.)

<details>
<summary><b>Evidence</b></summary>

macOS/Chromium via `nicegui.testing.Screen`, `pytest -rs`, `nicegui/.venv` Python 3.12.11.

At head `eddf543e` the footer is visible for explicit columns, auto-generated columns, and explicit-but-unsortable columns; all three show `NO-BOTTOM-EL` at base `932f019e`. (`rows=[]` unchanged.)

Your test is a real guard — reverting `table.py` while keeping it: `AssertionError: Could not find "1-2 of 3"` at `tests/test_table.py:75`, while `should_not_contain('Lionel')` still passed, so rows *were* paginated and only the controls were hidden.

With the candidate fix: `tests/test_table.py` (28) + a default-table guard + a `pagination = None` guard → **30 passed, 0 skipped**. mypy `Success`, pylint 10.00/10, ruff clean.

Not verified: that Linux CI shows the same reveal (mechanism is client-side Quasar emission, but I only ran macOS). CI is green and doesn't cover this — no test asserts a default table has no footer.

</details>

tests/test_table.py
```diff
@@ -62,6 +62,19 @@ def page():
...
+    @ui.page('/')
+    def page():
+        table = ui.table(columns=columns(), rows=rows())
+        table.pagination = {'rowsPerPage': 2}
+
+    screen.open('/')
+    screen.should_contain('Alice')
+    screen.should_contain('Bob')
+    screen.should_not_contain('Lionel')
+    screen.should_contain('1-2 of 3')
```
evnchn, line 75: Your test is a real guard in the direction the linked issue reports — reverting `table.py` while keeping it gives `AssertionError: Could not find "1-2 of 3"`. Nothing guards the reverse direction, though, which is exactly the gap the regression fell through.

```suggestion
def test_pagination_via_setter(screen: Screen):
    @ui.page('/')
    def page():
        table = ui.table(columns=columns(), rows=rows())
        table.pagination = {'rowsPerPage': 2}

    screen.open('/')
    screen.should_contain('Alice')
    screen.should_contain('Bob')
    screen.should_not_contain('Lionel')
    screen.should_contain('1-2 of 3')


def test_default_table_hides_pagination(screen: Screen):
    @ui.page('/')
    def page():
        ui.table(columns=columns(), rows=rows())

    screen.open('/')
    screen.should_contain('Alice')
    screen.should_not_contain('Records per page')
```

Browser-observable only, no internals, and it was seen to fail on the current head.

evnchn (commented):
**Not systemic — two changes and this is done.** `table.py` has exactly three write paths to `pagination`: the constructor, the public setter, and the mount-driven event handler. The first two *should* set visibility; the third must not. The PR unified all three. That's a scoping error, not a design problem.

**Your third option doesn't work, though.** Deriving `hide-pagination` in `_to_dict` is impossible without storing an extra flag — `pagination=None` and `pagination=0` normalize to the *identical* prop while needing *opposite* visibility:

| constructor arg | `hide-pagination` | `pagination` |
|---|---|---|
| `None` | `True` | `{'rowsPerPage': 0}` |
| `0` | `False` | `{'rowsPerPage': 0}` |
| `5` | `False` | `{'rowsPerPage': 5}` |

So it's down to the one-liner vs a private `_set_pagination(value, update_visibility=...)`. I measured both: identical behaviour, both fully green. The helper adds a parameter to share normalization that is a **no-op** at that call site — `e.args` is always a full Quasar pagination dict — so the one-liner wins on diff size. Still Falko's call.

### 1. The fix — `nicegui/elements/table.py`, line 96

This line is outside the PR's diff, so GitHub won't take it as a one-click suggestion:

```diff
         def handle_pagination_change(e: GenericEventArguments) -> None:
             previous_value = self.pagination
-            self.pagination = e.args
+            self._props['pagination'] = e.args  # not the setter: a client-side update must not un-hide the controls
```

The comment isn't decoration — without it the line reads as a tidy-me target and someone routes it back through the setter.

### 2. The guard — `tests/test_table.py`

Left inline as an applicable suggestion. This is the part I'd argue is actually load-bearing: nothing asserted that a *default* table has no footer, which is why green CI and my own **Verified** both sailed past this.

<details>
<summary><b>Evidence</b></summary>

macOS/Chromium via `nicegui.testing.Screen`, `pytest -rs`, `nicegui/.venv` Python 3.12.11, at head `eddf543e` with the one-line change applied.

`tests/test_table.py` (28) + the default-table guard + a `pagination = None` guard → **30 passed, 0 skipped**. mypy `Success`, pylint 10.00/10, ruff clean. Identical numbers for the `_set_pagination` variant.

The guard fails without the fix — reverting `table.py` to head and keeping it gives `Failed: DID NOT RAISE <class 'AssertionError'>` on `should_not_contain('Records per page')`, i.e. the footer really is there on a default table.

Default table with the fix: `hide-pagination=True` and no `.q-table__bottom` element — both at mount *and* after a sort click, which re-emits `update:pagination`. Explicit `pagination=2` unaffected: the browser next-page arrow still gives `Records per page: 2 … 3-3 of 3`.

The constructor table above is printed directly from `ui.table(rows=..., pagination=arg)._props`, no browser involved.

Second lineage (Codex, read-only pass): no blockers. Two things from it worth carrying — `_props` writes still propagate to the client without an explicit `update()` (observable-dict change hook, `nicegui/props.py:98`), and "exactly three write paths" wants the qualifier *within the typed API*: the generic `.props()` escape hatch can always poke either prop directly.

Not verified: that Linux CI shows the same reveal. The mechanism is client-side Quasar emission, but I only ran macOS.

</details>

SaadZahem:
I have verified the behavior on linux. Good call.
I felt no need for adding a helper method for this.
The test was added first then the code with fixed accordingly.

PR #6279, opened 2026-08-11
Limit the number of tab storages a single client can create

evnchn:
*Drafted by Claude Code on @evnchn's behalf. Updated by @falkoschindler for `23d7b46c` (removes the empty-tab reclamation again) and `d9977fa5a` (tolerates query-less sockets again, after the client-leak advisory fix on `main` made them a supported flow) — reasoning in the comments below.*

**TL;DR** — One unauthenticated `GET /` yields a valid `client_id`; replaying `handshake` frames on a socket then minted one tab storage per fresh `tab_id`, each retained for `max_tab_storage_age` (30 days) and surviving the peer's disconnect. This enforces **one socket ↔ one client ↔ one handshake** in the handshake handler and **pins each client to a single tab ID**. Over real sockets, 5000 replayed frames go from **5000 tab storages to 1**, and the socket's stranded `_num_connections` from **5999 to 1**. 36 insertions across 3 library files (+24 test), no storage-format change.

**Advisory route already cleared.** Raised with Falko on Slack as a possible GHSA; his verdict was that it does not warrant an advisory, so it goes the ordinary-PR route. That is why the security checkbox below is ticked.

### Motivation

`_on_handshake` gated only on `client_id` naming a live client, then reached `app.storage._create_tab_storage(tab_id)`, which for the default in-memory backend does `self._tabs[tab_id] = ObservableDict()` with no bound. `prune_tab_storage` only reclaims entries older than 30 days and `close_tab` is a no-op for `ObservableDict`, so the flooded entries outlive the peer.

The browser never behaves this way: the socket query carries `client_id`, the explicit handshake emits that very same `options.query`, a reconnect is a new sid, and `TAB_ID` is a JS `const` per page load — every new page render (reload, duplicate tab, back/forward) mints a fresh `client_id`, and a bfcache restore resumes the same const. So one client sees exactly one tab ID **by construction**, not by measurement.

### Implementation

`Client.accept_handshake(socket_id, tab_id, environ)` refuses a frame when any of the three invariants break, and is the single gate both transports call — `_on_handshake` in `nicegui/nicegui.py` and `Air._handle_handshake` in `nicegui/air.py`:

- **the socket's own client** — a `client_id` in the socket's connecting query must match the one the handshake names. A query *without* a `client_id` is tolerated: the query is no trust boundary (a forger simply echoes the claimed id into it), and query-less sockets are a supported flow since the client-leak advisory fix on `main`. `test-` sids keep their special case.
- **one handshake per socket** — a sid already in `_socket_to_document_id` is refused.
- **one tab ID per client** — the first handshake pins it; a later one presenting a different ID is refused.

The pin is a scalar with **its own never-reset field**, not `client.tab_id`, which `handle_disconnect` nulls on every disconnect — otherwise handshake A → disconnect → handshake B still mints two storages. `max_tab_storages_per_client` is gone as a public knob.

Empty tab storages are left to the existing 30-day pruner: the event-driven reclamation on client deletion from earlier rounds is removed again in `23d7b46c`. It decided a destructive delete of possibly shared state from process-local snapshots (the in-memory mirror for "empty", this process's `Client.instances` for "held elsewhere"), which carried a multi-worker Redis race and a reconnect-window race — and with the pin bounding the deposit to one empty tab storage per unauthenticated GET, that race surface no longer bought anything a frontline rate limit does not already cover.

Separate Air issue filed as #6310, as requested.

<details>
<summary><b>Verification</b> — before/after over real sockets, seen-to-fail per guard, local gates</summary>

**Real sockets** (measured on `42ab1d93`; the follow-up commits do not touch the pin path). A separate `ui.run()` process, a real `python-socketio` client, one `GET /` for the `client_id`, then replayed frames; counts read back through a plain `@app.get` endpoint. Before is `a9121acf`, the PR's base, served from its own worktree; each server prints the `nicegui.__file__` it is actually serving.

| one GET, then… | tab storages — before | tab storages — after |
|---|---|---|
| 1000 frames on one socket, fresh `tab_id` each | 1000 | **1** |
| 5000 frames on one socket, fresh `tab_id` each | 5000 | **1** |
| 12 sockets, one handshake each, fresh `tab_id` | 12 | **1** |
| 64 sockets, one handshake each, fresh `tab_id` | 64 | **1** |

Two by-products of the same run, both on the 5000-frame case: `_num_connections` on that socket goes **5999 → 1** (the replay path flagged in an earlier round — the one-handshake-per-socket rule closes it), and a handshake naming a *different* client than the socket connected with goes **accepted → refused**.

**Seen-to-fail, one guard at a time.** Each guard was reverted individually with the rest of the commit in place; each lands on a different assertion of `test_client_is_pinned_to_one_tab_id`:

| reverted | fails at |
|---|---|
| one handshake per socket | line 258 |
| `client_id` must match the socket's | line 266 |
| the `QUERY_STRING` arm of that check | line 266 |
| the pin itself | line 259 |
| pin bound to `client.tab_id` instead | line 262 |

**Local gates** (on `d9977fa5a`, `main` merged)

```
pre-commit run --files <the changed files>            all Passed
mypy ./nicegui                                        Success: no issues found in 265 source files
pylint ./nicegui                                      10.00/10, exit 0
pytest tests/test_storage.py tests/test_lifecycle.py tests/test_reconnect.py   41 passed
pytest tests/test_air.py tests/test_user_simulation.py                         85 passed
```

CI is the authoritative gate; not claiming green until the run lands.

</details>

<details>
<summary><b>What is not covered</b></summary>

**The Air arm of the `client_id` check is unverified.** `air.py` passes `data['environ']`, which crosses the relay as JSON, so its shape is unconfirmed without a relay. The helper reads both `QUERY_STRING` and `asgi.scope.query_string`, tolerates `str` or `bytes`, and fails open when neither query shape is present — so On Air cannot start refusing legitimate handshakes because of it. The other two invariants apply on Air unchanged.

**The real `sio.get_environ` path has no pytest.** `test-` sids bypass `sio.get_environ` by design, so the `client_id` check is asserted against `accept_handshake` directly with both environ shapes. The end-to-end refusal is covered by the real-socket measurement above, not by the suite; a pytest for it would need a raw Socket.IO client fixture.

**The pin test is white-box by necessity.** Replayed handshake frames are not reachable through the `User`/`Screen` fixtures, so `test_client_is_pinned_to_one_tab_id` drives `_on_handshake` directly — the justification AGENTS.md asks for when a test cannot assert observable behavior.

</details>

### Progress

### DIFF
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -138,7 +138,8 @@ class Air:
         @relay.on('handshake')
         def _handle_handshake(data: dict[str, Any]) -> bool:
             if client := Client.instances.get(data['client_id']):
-                if not client.accept_handshake(data['sid'], data['tab_id'], data['environ']):
+                # the relay's environ crosses as JSON and may lack the connecting query, so a missing client_id is tolerated
+                if not client.accept_handshake(data['sid'], data['tab_id'], data['environ'], require_client_id=False):
                     return False
                 client.environ = data['environ']
                 if data.get('old_tab_id'):
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -138,7 +138,7 @@ class Air:
         @relay.on('handshake')
         def _handle_handshake(data: dict[str, Any]) -> bool:
             if client := Client.instances.get(data['client_id']):
-                if not client.register_tab_id(data['tab_id']):
+                if not client.accept_handshake(data['sid'], data['tab_id'], data['environ']):
                     return False
                 client.environ = data['environ']
                 if data.get('old_tab_id'):
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -138,8 +138,7 @@ class Air:
         @relay.on('handshake')
         def _handle_handshake(data: dict[str, Any]) -> bool:
             if client := Client.instances.get(data['client_id']):
-                # the relay's environ crosses as JSON and may lack the connecting query, so a missing client_id is tolerated
-                if not client.accept_handshake(data['sid'], data['tab_id'], data['environ'], require_client_id=False):
+                if not client.accept_handshake(data['sid'], data['tab_id'], data['environ']):
                     return False
                 client.environ = data['environ']
                 if data.get('old_tab_id'):
--- a/nicegui/air.py
+++ b/nicegui/air.py
@@ -138,6 +138,8 @@ class Air:
         @relay.on('handshake')
         def _handle_handshake(data: dict[str, Any]) -> bool:
             if client := Client.instances.get(data['client_id']):
+                if not client.register_tab_id(data['tab_id']):
+                    return False
                 client.environ = data['environ']
                 if data.get('old_tab_id'):
                     core.app.storage.copy_tab(data['old_tab_id'], data['tab_id'])
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -74,9 +74,6 @@ class Client:
     instances: ClassVar[dict[str, Client]] = {}
     '''Maps client IDs to clients.'''
 
-    MAX_TAB_STORAGES_PER_CLIENT: ClassVar[int] = 128
-    '''Maximum number of tab storages a single client may create (a browser presents only one tab ID per client).'''
-
     shared_head_html = ''
     '''HTML to be inserted in the <head> of every page template.'''
 
@@ -354,8 +351,8 @@ class Client:
         self._exception_handlers.append(handler)
 
     def register_tab_id(self, tab_id: str) -> bool:
-        """Remember a tab ID, unless `MAX_TAB_STORAGES_PER_CLIENT` is already exhausted. (For internal use only.)"""
-        if tab_id not in self._tab_ids and len(self._tab_ids) >= self.MAX_TAB_STORAGES_PER_CLIENT:
+        """Remember a tab ID, unless ``max_tab_storages_per_client`` is already exhausted. (For internal use only.)"""
+        if tab_id not in self._tab_ids and len(self._tab_ids) >= core.app.storage.max_tab_storages_per_client:
             return False
         self._tab_ids.add(tab_id)
         return True
@@ -395,6 +392,8 @@ class Client:
             if self._num_connections[document_id] == 0:
                 self._num_connections.pop(document_id)
                 self._delete_tasks.pop(document_id)
+                active_elsewhere = {client.tab_id for client in Client.instances.values() if client is not self}
+                await core.app.storage.prune_empty_tabs(self._tab_ids - active_elsewhere)
                 await core.app.storage.close_tab(tab_id_to_close)
                 self.delete()
         self._delete_tasks[document_id] = \
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from collections.abc import Callable, Iterable
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar, cast
-from urllib.parse import quote
+from urllib.parse import parse_qs, quote
 
 from fastapi import Request
 from fastapi.responses import Response
@@ -61,6 +61,14 @@ HTML_ESCAPE_TABLE = str.maketrans({
 HEADWIND_CONTENT = (Path(__file__).parent / 'static' / 'headwind.css').read_text().strip()
 
 
+def _client_id_from_query(environ: dict[str, Any]) -> str | None:
+    """Read the ``client_id`` a socket connected with, or ``None`` if its environment does not carry one."""
+    query_string = environ.get('QUERY_STRING') or environ.get('asgi.scope', {}).get('query_string') or ''
+    if isinstance(query_string, (bytes, bytearray)):
+        query_string = query_string.decode()
+    return parse_qs(query_string).get('client_id', [None])[0]
+
+
 class ClientConnectionTimeout(TimeoutError):
     def __init__(self, client: Client) -> None:
         super().__init__(f'ClientConnectionTimeout: {client.id}')
@@ -99,7 +107,7 @@ class Client:
         self._deleted = False
         self._socket_to_document_id: dict[str, str] = {}
         self.tab_id: str | None = None
-        self._tab_ids: set[str] = set()
+        self._pinned_tab_id: str | None = None
         self._exception_handlers: list[Callable[[Exception], Any] | Callable[[], Any]] = []
 
         self.page = page
@@ -350,12 +358,19 @@ class Client:
         """
         self._exception_handlers.append(handler)
 
-    def register_tab_id(self, tab_id: str) -> bool:
-        """Remember a tab ID, unless ``max_tab_storages_per_client`` is already exhausted. (For internal use only.)"""
-        if tab_id not in self._tab_ids and len(self._tab_ids) >= core.app.storage.max_tab_storages_per_client:
+    def accept_handshake(self, socket_id: str, tab_id: str, environ: dict[str, Any] | None) -> bool:
+        """Check whether a handshake may proceed, pinning the client's tab ID on the first one.
+
+        A browser opens one socket per client, handshakes it once, and keeps the same tab ID for the client's whole
+        lifetime, so a handshake that breaks any of these is a replayed frame. (For internal use only.)
+        """
+        if socket_id in self._socket_to_document_id:
             return False
-        self._tab_ids.add(tab_id)
-        return True
+        if environ is not None and _client_id_from_query(environ) not in (None, self.id):
+            return False
+        if self._pinned_tab_id is None:
+            self._pinned_tab_id = tab_id
+        return self._pinned_tab_id == tab_id
 
     def handle_handshake(self, socket_id: str, document_id: str, next_message_id: int | None) -> None:
         """Cancel pending disconnect task and invoke connect handlers. (For internal use only.)"""
@@ -392,12 +407,16 @@ class Client:
             if self._num_connections[document_id] == 0:
                 self._num_connections.pop(document_id)
                 self._delete_tasks.pop(document_id)
-                active_elsewhere = {client.tab_id for client in Client.instances.values() if client is not self}
-                await core.app.storage.prune_empty_tabs(self._tab_ids - active_elsewhere)
+                if self._pinned_tab_id is not None and not self._tab_id_is_held_elsewhere():
+                    await core.app.storage.prune_empty_tab(self._pinned_tab_id)
                 await core.app.storage.close_tab(tab_id_to_close)
                 self.delete()
         self._delete_tasks[document_id] = \
             background_tasks.create(delete_content(), name=f'delete content {document_id}')
+
+    def _tab_id_is_held_elsewhere(self) -> bool:
+        return any(client._pinned_tab_id == self._pinned_tab_id and client._socket_to_document_id  # pylint: disable=protected-access
+                   for client in Client.instances.values() if client is not self)
 
     def _cancel_delete_task(self, document_id: str) -> None:
         if document_id in self._delete_tasks:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -358,22 +358,20 @@ class Client:
         """
         self._exception_handlers.append(handler)
 
-    def accept_handshake(self, socket_id: str, tab_id: str, environ: dict[str, Any] | None, *,
-                         require_client_id: bool = True) -> bool:
+    def accept_handshake(self, socket_id: str, tab_id: str, environ: dict[str, Any] | None) -> bool:
         """Check whether a handshake may proceed, pinning the client's tab ID on the first one.
 
-        A browser opens one socket per client, handshakes it once, keeps the same tab ID for the client's whole
-        lifetime, and always carries its client ID in the socket's connecting query --
-        so a handshake that breaks any of these is a replayed or forged frame.
-        Pass ``require_client_id=False`` for transports whose environ may legitimately lack the connecting query
-        (the Air relay), or ``environ=None`` to skip the query check entirely (tests). (For internal use only.)
+        A browser opens one socket per client, handshakes it once, and keeps the same tab ID for the client's whole
+        lifetime, so a handshake that breaks any of these is a replayed frame.
+        A socket query without a client ID is tolerated:
+        the query is no trust boundary (a forger could simply echo the claimed client ID into it),
+        and query-less sockets must keep working (see ``test_disconnect_without_client_id_in_connect_query``).
+        (For internal use only.)
         """
         if socket_id in self._socket_to_document_id:
             return False
-        if environ is not None:
-            client_id = _client_id_from_query(environ)
-            if client_id != self.id and (client_id is not None or require_client_id):
-                return False
+        if environ is not None and _client_id_from_query(environ) not in (None, self.id):
+            return False
         if self._pinned_tab_id is None:
             self._pinned_tab_id = tab_id
         return self._pinned_tab_id == tab_id
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -358,16 +358,22 @@ class Client:
         """
         self._exception_handlers.append(handler)
 
-    def accept_handshake(self, socket_id: str, tab_id: str, environ: dict[str, Any] | None) -> bool:
+    def accept_handshake(self, socket_id: str, tab_id: str, environ: dict[str, Any] | None, *,
+                         require_client_id: bool = True) -> bool:
         """Check whether a handshake may proceed, pinning the client's tab ID on the first one.
 
-        A browser opens one socket per client, handshakes it once, and keeps the same tab ID for the client's whole
-        lifetime, so a handshake that breaks any of these is a replayed frame. (For internal use only.)
+        A browser opens one socket per client, handshakes it once, keeps the same tab ID for the client's whole
+        lifetime, and always carries its client ID in the socket's connecting query --
+        so a handshake that breaks any of these is a replayed or forged frame.
+        Pass ``require_client_id=False`` for transports whose environ may legitimately lack the connecting query
+        (the Air relay), or ``environ=None`` to skip the query check entirely (tests). (For internal use only.)
         """
         if socket_id in self._socket_to_document_id:
             return False
-        if environ is not None and _client_id_from_query(environ) not in (None, self.id):
-            return False
+        if environ is not None:
+            client_id = _client_id_from_query(environ)
+            if client_id != self.id and (client_id is not None or require_client_id):
+                return False
         if self._pinned_tab_id is None:
             self._pinned_tab_id = tab_id
         return self._pinned_tab_id == tab_id
@@ -407,16 +413,10 @@ class Client:
             if self._num_connections[document_id] == 0:
                 self._num_connections.pop(document_id)
                 self._delete_tasks.pop(document_id)
-                if self._pinned_tab_id is not None and not self._tab_id_is_held_elsewhere():
-                    await core.app.storage.prune_empty_tab(self._pinned_tab_id)
                 await core.app.storage.close_tab(tab_id_to_close)
                 self.delete()
         self._delete_tasks[document_id] = \
             background_tasks.create(delete_content(), name=f'delete content {document_id}')
-
-    def _tab_id_is_held_elsewhere(self) -> bool:
-        return any(client._pinned_tab_id == self._pinned_tab_id and client._socket_to_document_id  # pylint: disable=protected-access
-                   for client in Client.instances.values() if client is not self)
 
     def _cancel_delete_task(self, document_id: str) -> None:
         if document_id in self._delete_tasks:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -74,6 +74,9 @@ class Client:
     instances: ClassVar[dict[str, Client]] = {}
     '''Maps client IDs to clients.'''
 
+    MAX_TAB_STORAGES_PER_CLIENT: ClassVar[int] = 128
+    '''Maximum number of tab storages a single client may create (a browser presents only one tab ID per client).'''
+
     shared_head_html = ''
     '''HTML to be inserted in the <head> of every page template.'''
 
@@ -99,6 +102,7 @@ class Client:
         self._deleted = False
         self._socket_to_document_id: dict[str, str] = {}
         self.tab_id: str | None = None
+        self._tab_ids: set[str] = set()
         self._exception_handlers: list[Callable[[Exception], Any] | Callable[[], Any]] = []
 
         self.page = page
@@ -348,6 +352,13 @@ class Client:
         The callback has an optional parameter of `Exception`.
         """
         self._exception_handlers.append(handler)
+
+    def register_tab_id(self, tab_id: str) -> bool:
+        """Remember a tab ID, unless `MAX_TAB_STORAGES_PER_CLIENT` is already exhausted. (For internal use only.)"""
+        if tab_id not in self._tab_ids and len(self._tab_ids) >= self.MAX_TAB_STORAGES_PER_CLIENT:
+            return False
+        self._tab_ids.add(tab_id)
+        return True
 
     def handle_handshake(self, socket_id: str, document_id: str, next_message_id: int | None) -> None:
         """Cancel pending disconnect task and invoke connect handlers. (For internal use only.)"""
--- a/nicegui/nicegui.py
+++ b/nicegui/nicegui.py
@@ -206,7 +206,8 @@ async def _on_handshake(sid: str, data: dict[str, Any]) -> bool:
     if not client:
         return False
     is_test = sid.startswith('test-')
-    if not client.accept_handshake(sid, data['tab_id'], None if is_test else sio.get_environ(sid)):
+    environ = None if is_test else sio.get_environ(sid)
+    if not client.accept_handshake(sid, data['tab_id'], environ):
         return False
     if data.get('old_tab_id'):
         app.storage.copy_tab(data['old_tab_id'], data['tab_id'])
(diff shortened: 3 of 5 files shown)


falkoschindler (commented):
Thanks for the thorough write-up and measurements, @evnchn! The implementation does what it says: the guard runs before anything is minted, reconnects stay idempotent, and a legitimate browser can't hit the cap. My concern is more fundamental, though: what does the cap buy us if the attacker can simply loop the `GET /` request? One unauthenticated GET yields a fresh `client_id`, and with 30 days of retention there is no hurry — the flood just becomes `GET` → 128 frames → `GET` → 128 frames… and `app.storage._tabs` still grows without bound. So I'm not yet convinced this fix carries its weight in its current form. Details in finding 1, plus three smaller points.

1. **The cap barely changes the attacker's throughput — the bound is per client, and a client costs one unauthenticated GET** (major) — The TL;DR's "flattens at 128" holds for one `client_id`; an attacker just loops `GET /` → parse `client_id` → 128 frames. That raises their cost by roughly one page render per 128 storages, i.e. by a small constant factor, and `app.storage._tabs` still grows unbounded until the 30-day prune. Two ways to make the fix carry its weight: (a) lower the constant drastically — the PR's own measurement says an honest client presents exactly one `tab_id`, and a too-low cap costs a legitimate user one reload at worst, so something like 4 would reduce amplification from 128× to 4× per GET; (b) use the new `_tab_ids` set in `Client.delete()` to drop the storages this client created that are still empty (`_create_tab_storage` is idempotent, so a reload recreates them; for Redis that means closing the `PersistentDict` and deleting the key). (b) would actually bound the persistent resource by the client lifetime, which is what the Motivation is about. At minimum, the Motivation should state that this caps storages per client, not per peer, so the table isn't read as a global bound.

2. **Placement of `MAX_TAB_STORAGES_PER_CLIENT`** (minor) — Tab-storage policy already lives on `Storage` (`max_tab_storage_age`, `nicegui/storage.py` line 93, documented in the storage docs), and `Client` has no other uppercase-constant `ClassVar`; existing module constants are module-level (`USER_STORAGE_PRUNE_INTERVAL`, `MAX_CHUNK_SIZE`). Consider `Storage.max_tab_storages_per_client` next to `max_tab_storage_age`, or a module-level constant in `client.py`. Related: `nicegui/nicegui.py` lines 208–212 and `nicegui/air.py` lines 141–146 now duplicate the whole register → copy_tab → `tab_id` → `handle_handshake` sequence; folding it into one `Client` method (returning `bool`) would keep the two transports from drifting. Pre-existing duplication, so leave-or-take.

3. **Test noise** (minor) — `tests/test_storage.py` lines 248–265: `uuid4` for `sid` and `document_id` isn't needed (`_on_handshake` only requires `sid.startswith('test-')`; `_socket_to_document_id`/`_num_connections` tolerate constants), so `f'test-{i}'`/`'doc'` and `f'tab-{i}'` would drop the import and make clear that only `tab_id` distinctness matters. The 127 `assert await …` calls and the final `len(app.storage._tabs) == MAX` assert the same property — either keep the return-value asserts and drop the protected-access line (AGENTS.md), or keep only the count. `ui.label('ok')` in the page is dead scaffolding. Driving `_on_handshake` directly is fine given the PR names the reason.

4. **Observation, not caused by this PR** — `Air._handle_handshake` never calls `_create_tab_storage` (only `copy_tab`), so on Air the guard only bounds the duplicate-tab path, and `app.storage.tab` on an Air page without a preceding `copy_tab` would hit the `assert client.tab_id in self._tabs`. Worth a separate issue if it isn't already known. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20from%20a%20multi-agent%20code%20review%2C%20wording%20by%20the%20model%3B%20the%20opening%20concern%20is%20the%20reviewer%27s%20own.&a=claude-code "claude-fable-5: Findings from a multi-agent code review, wording by the model; the opening concern is the reviewer's own.")

evnchn:
*Reply drafted by Claude Code on @evnchn's behalf.*

**You're right that the cap alone, at 128, doesn't carry its weight** — I measured it, not dressing it up. But its value is exactly `R × C` (frontline GET-rate × cap), so 128 is just too large, and the persistent resource is still reachable via a channel no HTTP gate sees. The "assume GET is gated" argument holds — it's the case *for* your (a) and (b), not for 128 — so I took both:

- **(a)** cap → **4**, on `Storage.max_tab_storages_per_client` (beside `max_tab_storage_age`).
- **(b)** reclaim the empty tab storages a client made when it dies → the persistent resource is now bounded by **client lifetime**, not 30 days. Real socket: flood → 4 tabs → disconnect → **0**.

<details>
<summary><b>Finding 1 — measured</b></summary>

Persistent tab storage is reachable two ways: `GET /` (HTTP, what a gate limits) and `handshake` frames over an open socket (**invisible to an HTTP gate**). Under a gate of `R` GET/s per IP:

| variant | persistent rate | gate binds? | 30-day walk-away deposit? |
|---|---|---|---|
| uncapped (today) | ~5000/s off **one** GET | no | yes |
| cap only, 128 | `R × 128` | yes, but too big to bite | yes |
| **cap 4 + reclaim (this PR)** | `R × 4`, only while sockets held | yes | **no** |

Localhost: ~609 B/empty-tab, ~750 GET/s, ~5000 frames/s; uncapped one socket = 50 000 frames → 50 000 tabs. So the cap alone is a constant-factor cost (you're right); (a)+(b) are what make a frontline gate actually effective and kill the walk-away deposit.

</details>

<details>
<summary><b>(b) — and the reload crash I hit building it</b></summary>

```python
active_elsewhere = {client.tab_id for client in Client.instances.values() if client is not self}
await core.app.storage.prune_empty_tabs(self._tab_ids - active_elsewhere)
```

`prune_empty_tabs` mirrors `prune_tab_storage` (`clear()` → for Redis deletes the key → `close()` → drop), touching **empty** storages only, so real data keeps full retention.

The `- active_elsewhere` guard is load-bearing: a **reload reuses the same `tab_id`** (`nicegui.js` L383-388) under a fresh `client_id`, so within `reconnect_timeout` the dying and live clients share one `tab_id`. Without the guard the old client's reap deletes the live client's empty storage → `assert client.tab_id in self._tabs` crash for a real user. Guard reclaims only what no live client still claims — still catches every attacker orphan.

Three tests, each seen-to-fail with only its own fix line reverted: cap refusal; empty reclaimed (data survives); shared-tab kept.

</details>

<details>
<summary><b>Points 2 / 3 / 4</b></summary>

**2** — constant moved onto `Storage`. Left the register/copy_tab dedup: the transports diverge past the shared guard (`enter_room`/`_create_tab_storage` vs `on_air`, no `_create_tab_storage`), and the duplicated part is now just the two-line guard. Will fold it if you'd rather.
**3** — dropped `uuid4` + `ui.label('ok')`, `f'test-…'`/`'doc'` constants, kept the refusal assertion over the `len(_tabs)` line.
**4** — confirmed: `Air._handle_handshake` never calls `_create_tab_storage`, so on Air the cap only bounds the `copy_tab` path. Pre-existing; I'll open a separate issue unless already tracked.

Local: pre-commit ✓ · mypy clean · pylint 10.00 · `pytest -k tab_storage` 7 passed. Not claiming CI green yet.

</details>

falkoschindler (commented):
Thanks, @evnchn — (a) and (b) are the right moves, the constant reads well on `Storage`, and CI is green. I pulled 787faf545 and poked at the reclaim path, and I'm afraid (b) doesn't bound the resource yet: it hangs on the client actually getting deleted, and the same handler lets a replay peer keep the client alive indefinitely. Two ways, both verified locally with the `User` fixture:

1. **Same `document_id` twice on one socket.** `handle_handshake` does `_num_connections[doc] += 1` per frame; `handle_disconnect` decrements once, so the count strands at 1, `delete_content` is a no-op, and the no-op task stays in `_delete_tasks` — which makes `prune_instances` skip the client forever (it requires `not client._delete_tasks`). Repro: `user.open('/')`, two `_on_handshake('test-x', {..., 'tab_id': 't1'/'t2', 'document_id': 'doc'})`, `client.handle_disconnect('test-x')`, sleep, `Client.prune_instances(client_age_threshold=0)` → client still in `Client.instances`, `t1`/`t2` still in `_tabs`.
2. **Handshake for a `client_id` other than the socket's own.** `_on_disconnect` resolves the client from the socket's *query* `client_id`, so a client that was only ever handshaked by frame never sees `handle_disconnect`; its `tab_id` stays set → `has_socket_connection` is `True` → `prune_instances` skips it forever. Repro: `a = await user.open('/')`, `b = Client(a.page, request=a.request)`, `_on_handshake('test-a', {'client_id': b.id, 'tab_id': 'tb', ...})`, `a.handle_disconnect('test-a')`, `Client.prune_instances(client_age_threshold=0)` → `b` still in `Client.instances`, `tb` still in `_tabs`.

In both cases the whole `Client` (element tree, handlers) leaks permanently, not just an empty `ObservableDict` — and with no 30-day prune at all. That leak is pre-existing and heavier than the one this PR targets, but since (b) depends on client deletion, it decides whether the "walk-away deposit: no" row in your table holds. Right now it doesn't.

The root cause is that the handshake doesn't enforce "one socket ↔ one client ↔ one handshake", which is what the JS always does: the socket query carries `client_id`, the explicit handshake emits the very same `options.query`, and a reconnect is a new sid. So I'd tighten `_on_handshake` instead of (or in addition to) counting: refuse when `data['client_id']` differs from the socket's query `client_id` (`sio.get_environ(sid)`, exactly as `_on_disconnect` already reads it; the `test-` sids keep their special case), and refuse a second handshake on a sid that is already in `client._socket_to_document_id`. The Air relay forwards `environ`, so the same check works there. With that, tab storages per socket = 1, clients can't be kept alive by replay, and (b) becomes a real bound — the per-client cap then is belt-and-braces. I'd rather see that in this PR since it's the actual root and only a few lines, but if you prefer to keep this one small, land (a)+(b) and do the tightening in a follow-up — then the PR text should say the client-lifetime bound assumes the client is deleted.

Smaller things on the new commit:

- `active_elsewhere` walks all of `Client.instances` on every client deletion — O(clients) per deletion. Fine at NiceGUI's usual scale; just noting it since it runs on every page close. (minor)
- Leaving the register/copy_tab duplication is fine. And yes, please open the separate Air issue for point 4. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Leak%20paths%20found%20and%20verified%20locally%20by%20the%20model%2C%20wording%20by%20the%20model.&a=claude-code "claude-fable-5: Leak paths found and verified locally by the model, wording by the model.")

evnchn:
I just had a shower thought: how does this PR relate to efforts in bringing back the shared client model (2.x's auto-index client) such as https://github.com/zauberzeug/nicegui/pull/5421 ?

IF the handshake really enforce "one socket ↔ one client ↔ one handshake" as you mentioned...

Let me think about this and your review comment together.

evnchn:
*Reply drafted by Claude Code on @evnchn's behalf.*

**The shower thought lands on your own review rather than beside it: the cap is load-bearing *because* many sockets can share one client — and that is exactly what a shared page is.** Attack shape and feature shape are the same shape, so the cap is where the two goals trade against each other.

Your route is the one I want to take — tighten `_on_handshake`, keep the count behind it. Measuring it against PR 5421 turned up two places where the reasoning needs correcting, both of them mine to correct, and they change what I'd push. Plan at the bottom; I'd like your read on it before I write it.

**"the per-client cap then is belt-and-braces" understates it.** Under the tightening, one unauthenticated GET still buys the full cap: a peer opens N sockets that each satisfy the rule — own `sid`, one handshake, matching query `client_id` — and mints a fresh `tab_id` on each. Ordinary private page, this branch, real sockets:

```
1 GET -> client 05e78589 -> 12 sockets attempted, 4 accepted
server: {"clients": 1, "tabs": 4}
```

Not an argument against the tightening — it fixes the leak you found, which the count doesn't touch. But the count stays load-bearing rather than becoming redundant, which means its anchor still matters.

**And "a too-low cap costs a legitimate user one reload at worst" is the part shared pages break.** It isn't one reload. `ConnectionRefusedError('Implicit handshake failed')` (`nicegui.py:199`) meets the browser's `connect_error` handler, which reloads (`nicegui.js:483`); the reload fires `onbeforeunload`, so `nicegui.js:383-388` hands back the *same* `sessionStorage` tab ID — the one just refused, and therefore never registered. Refused again, indefinitely.

Unreachable today, because a real browser presents exactly one tab ID per client. Reachable the moment N browsers share one client — PR 5421's `page.py` diff applied on top of 787faf545:

| 8 visitors, one shared page | cap 10 000 | cap 4 |
|---|---|---|
| sockets | 8/8 connected, 1 client, 8 tab storages | **visitors 5-8 refused**, each reloading forever |

The docstring I wrote states the assumption out loud — *"a browser presents one tab ID per client, so this is headroom"* — and that's the sentence PR 5421 exists to make false. Nor is it only PR 5421: a shared client is any code holding a `Client` and re-serving `build_response`, which is what that PR does in about a dozen lines of `page.py`.

### What I plan to push, if you're happy with it

1. **Your two checks in `_on_handshake`**, plus the Air relay — as specified, unchanged.
2. **`_socket_to_tab_id: dict[str, str]` on `Client`**, beside the existing `_socket_to_document_id` (`client.py:100`): set in `handle_handshake`, popped in `handle_disconnect`. It stays single-valued per `sid` only because of item 1's one-handshake-per-sid rule, so 2 depends on 1.
3. **Fix (b)'s holder test to use it**, across all clients *including* `self`:

   ```python
   held = {tab for c in Client.instances.values() for tab in c._socket_to_tab_id.values()}
   await core.app.storage.prune_empty_tabs(self._tab_ids - held)
   ```

   That keeps your existing per-`document_id` `_delete_tasks` grace rather than inventing a second timeout, stays the same O(clients) scan you flagged, and still covers the reload case `active_elsewhere` exists for — the reloaded client's live socket holds the ID, so it isn't reclaimed. I looked at replacing the scan with a `Counter[tab_id]` on `Storage` and decided against proposing it: too many paths would have to decrement, and the Air relay-loss path (`air.py:164-166`) doesn't run `handle_disconnect` at all.
4. **Keep the cap at 4**, with the docstring saying what it is anchored to: a per-`Client` *lifetime* budget (`_tab_ids` at `client.py:102` is only ever added to) that assumes one browser per client.

What that does and doesn't buy, stated carefully because I got it wrong first time: a peer that **disconnects** loses its empty storages at socket teardown instead of at client deletion — which is the part of (b) that currently doesn't fire, since your two leak paths keep the client alive. A peer that **holds its sockets open** keeps its storages, and the cap is what bounds that, exactly as the 12-socket measurement above shows. So the count stays load-bearing either way; 3 narrows (b)'s dependency on client deletion rather than removing it.

It also doesn't make shared pages work — `client.tab_id`, `app.storage.tab` (`storage.py:181-183` assumes one tab per client), `delete_content` and Air all still need answers. What it avoids is adding a *new* obstacle on top of those.

**The one call I don't think is mine: the constant in 4.** If shared pages are still a direction you want, 4 is a 4-concurrent-viewer ceiling and the number wants a per-socket anchor instead; if they aren't, 4 is a reasonable answer to the peer and item 4 collapses to a docstring. That trade is what I'd value a re-read on, with PR 5421 in view rather than just this diff.

Two things I'm deliberately leaving out of this PR as shared-page design rather than this fix: what `app.storage.tab` should mean when one client has many viewers, and `delete_content` (see fold).

<details>
<summary><b>Three pre-existing rows of the same shape</b>, found while measuring — all on <code>main</code></summary>

Per-client state that is really per-socket — the same root as the leak you found:

**`client.tab_id`** (`client.py:101`) — last handshake wins, and the first disconnect nulls it, so `has_socket_connection` (`client.py:151`) goes `False` while other viewers are live and `_on_event` (`nicegui.py:237`) drops their events.

**`delete_content`** — each viewer brings its own `document_id`, so the first to leave takes its own document to zero and reaches `self.delete()` (`client.py:398`) on the whole shared client. Four viewers, one leaves:

```
before:                            {"clients": 1, "tabs": 4}
after visitor 1 disconnects:       {"clients": 0, "tabs": 0}
shared client identity preserved:  False  (ea239df2 -> c13b7d0b)
```

Three were still connected. (Codex flagged this one — I had missed it.)

**And the teardown is noisy**: the remaining viewers each schedule their own `delete_content`, so `del Client.instances[self.id]` (`client.py:482`) raises for every one after the first — 7 unhandled `KeyError` tracebacks through `background_tasks._handle_exceptions` in the 8-viewer run.

</details>

Air issue for point 4 either way.

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Measured%20by%20the%20model%20against%20787faf545%20with%20PR%205421%20applied%2C%20real%20sockets%3B%20the%20shower%20thought%2C%20the%20framing%20and%20the%20plan%20are%20the%20author%27s.%20Codex%20reviewed%20the%20claims%20and%20supplied%20the%20delete_content%20and%20KeyError%20rows.&a=claude-code "claude-opus-5: Measured by the model against 787faf545 with PR 5421 applied, real sockets; the shower thought, the framing and the plan are the author's. Codex reviewed the claims and supplied the delete_content and KeyError rows.")

falkoschindler:
Heads-up before you start on the plan, @evnchn: I just closed #5421 — the shared-pages experiment is off the roadmap for now. The closing comment over there has the full reasoning; the short version is that your measurements in this thread showed the mismatch is systematic (per-client state that would have to become per-socket), so shared pages are a refactoring project, not a flag, and that project doesn't earn its place right now.

That settles the call you left to me — and more decisively than "keep 4": with shared pages off the table, the invariant is that one client sees exactly one tab ID, by construction, not just by measurement. `TAB_ID` is a JS `const` per page load, so all reconnects of a page present the same ID; every new page render (reload, duplicate tab, back/forward) mints a fresh `client_id`; a bfcache restore resumes the same `const`. There is no path to a second tab ID on a live client, so a set with headroom guards against something that cannot happen. Let's make it a scalar instead: the first handshake pins the client's tab ID, any later handshake with a different one is refused. That also removes `max_tab_storages_per_client` as a public config knob before anyone depends on it, and the prune in `delete_content` compares one ID instead of diffing sets.

One subtlety, which you'll hit anyway: the pin must not be the existing `client.tab_id`, since `handle_disconnect` nulls that on every disconnect — otherwise handshake A → disconnect → handshake B still mints two storages. It needs its own never-reset field with lifetime semantics (which is what the set implicitly provided).

Items 1–3 of your plan are good as specified — with the scalar, item 3's holder bookkeeping should get simpler too. And yes to the separate Air issue either way.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Decision%20and%20verdict%20by%20the%20author%3B%20wording%20by%20the%20model%2C%20the%20scalar-over-set%20argument%20from%20the%20review%20conversation.&a=claude-code "claude-fable-5: Decision and verdict by the author; wording by the model, the scalar-over-set argument from the review conversation.")

evnchn:
*Reply drafted by Claude Code on @evnchn's behalf.*

**Scalar landed — and the tightening closed the one thing I had deferred.** Pushed as `42ab1d93`; PR text rewritten around it. Thanks for settling item 4 by removing the question rather than answering it: with shared pages off the table the invariant is structural, and a set with headroom was guarding against something that cannot happen.

**The pin has its own never-reset field, as you said it must.** I did not want to take that on trust, so it has a seen-to-fail row of its own: rebinding the pin to `client.tab_id` with the rest of the commit in place makes `test_client_is_pinned_to_one_tab_id` fail at the assertion after the disconnect, and nothing else. Each of the seven guards is reverted individually in the PR's verification fold, and each lands on a different line.

**Item 1 landed as one gate for both transports** — `Client.accept_handshake(socket_id, tab_id, environ)`, called from `_on_handshake` and `Air._handle_handshake`. That also takes the leave-or-take from your first review about the two paths drifting.

**One divergence, since you said 1–3 as specified: item 2 is gone.** `_socket_to_tab_id` existed only because a client could hold several tab IDs at once. Under the scalar it would be a single-valued mirror of the pin, so item 3's holder test is now `_pinned_tab_id` plus "does that client still have a socket" (`_socket_to_document_id` non-empty). That is the simplification you predicted, and it has a second benefit I would rather have than the extra dict: it no longer reads `client.tab_id`, which your own subtlety says is unreliable — `handle_disconnect` nulls it while other sockets of the same client are still live, so the old `active_elsewhere` set could have failed to see a live holder. Say the word if you would still rather have the explicit map.

**The `_num_connections` replay is closed as a by-product.** That was Codex's MUST-FIX last round, which I measured and deferred as pre-existing and transient. One handshake per socket removes it: 5000 replayed frames on one socket now leave `_num_connections` at **1**, where before they stranded **5999**. I am reporting it because it was outstanding, not claiming it as design — it fell out of your check.

Air issue filed as #6310, with a reproduction of the handshake sequence.

<details>
<summary><b>One thing I could not verify, and would rather you decide</b></summary>

The Air arm of the `client_id` check. `air.py` hands over `data['environ']`, which crosses the relay as JSON, and I have no relay to confirm its shape against. The helper therefore reads both `QUERY_STRING` and `asgi.scope.query_string`, tolerates `str` or `bytes`, and **fails open** when neither is present — so On Air cannot start refusing legitimate handshakes because of it, and the other two invariants still apply there. Codex flagged the `QUERY_STRING` omission specifically as the way that check could be defeated on the relay path, which is why it reads both.

The alternative is to pass `None` for Air and let the check run only where I can see the environ. That is more honest and strictly weaker; I went with the version that does what you asked, guarded so it cannot break anything. Happy to switch.

</details>

CI is running; not claiming green until it lands.

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Implemented%20and%20measured%20by%20the%20model%20against%20a9121acf%20over%20real%20sockets%3B%20the%20scalar%20decision%2C%20the%20never-reset%20field%20and%20the%20plan%20are%20the%20reviewer%20and%20the%20author.%20Codex%20reviewed%20the%20commit%20and%20supplied%20the%20QUERY_STRING%20finding.&a=claude-code "claude-opus-5: Implemented and measured by the model against a9121acf over real sockets; the scalar decision, the never-reset field and the plan are the reviewer and the author. Codex reviewed the commit and supplied the QUERY_STRING finding.")

evnchn:
**CI green on `42ab1d93`** — `pytest` 24m18s, `mypy` / `pylint` / `pre-commit` / Windows storage all pass. The `test_late_event_registration` flake I saw once locally did not appear.

falkoschindler (dismissed):
@evnchn, I pushed a commit directly to your branch to spare us another round-trip — I hope that's fine. It _removes_ the reclamation half and keeps the pin.

While reviewing `42ab1d93` I traced all remaining findings, and they clustered on one side of a line: the handshake tightening — one handshake per socket, `client_id` bound to the socket, the tab-ID pin — is sound; every legitimate browser flow I checked (reconnect, bfcache restore, tab duplication, polling fallback) survives it. The reclamation, though, is an event-driven destructive delete of possibly shared state, decided from process-local snapshots — the in-memory mirror for "empty", this process's `Client.instances` for "held elsewhere". That shape kept producing races faster than we could close them: under multi-worker Redis, one worker's reap could delete a key another worker's client is actively using (the very deletion the pub/sub listener documents it must avoid), and a client inside its reconnect window didn't count as a holder, so a shared tab could be pruned out from under it. Each fix would have narrowed one interleaving and left the next for review five.

So I applied your own earlier threat-model argument to it: with the pin, one unauthenticated GET buys at most **one** empty ~600-byte tab storage — exactly the kind of resource a frontline rate limit bounds — and the pre-existing 30-day pruner keeps covering the leftovers. The reclamation only shortened the lifetime of an already-bounded, tiny object, and that's not worth its race surface. Removing it also takes `prune_empty_tab`, the holder scan, and the two tests that exercised them; `storage.py` is now untouched by this PR, which shrinks it to 65 insertions across 4 files.

Two small tightenings from the review went in with it: a direct socket whose connecting query carries no `client_id` is now refused (a real browser always sends one — the check previously failed open), while the Air relay keeps tolerating a missing query via `require_client_id=False`, since its environ crosses as JSON with an unverified shape — your caution there was right. And the environ is fetched once per handshake, so the guard validates the same environ that gets stored.

The pin test stays white-box, since replayed handshake frames aren't reachable through the `User`/`Screen` fixtures — that's the justification AGENTS.md asks for. It gained one row: a query without a `client_id` must be refused.

I've also updated the PR description to match — the reclamation parts (the reclaim rows in the verification fold, the `delete_content` mention) are gone. Thanks for the endurance across these rounds; the handshake gate that survives them is a genuinely good piece of work.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Decision%20and%20reasoning%20from%20the%20review%20conversation%20with%20the%20author%3B%20implementation%20and%20wording%20by%20the%20model.&a=claude-code "claude-fable-5: Decision and reasoning from the review conversation with the author; implementation and wording by the model.")

falkoschindler:
The red CI turned out to be a genuine collision with `main`, and resolving it walks back one call from my previous comment: the fail-closed `client_id` check is gone again (`d9977fa5a`), and `main` is merged in (`b6177ab1b`).

What happened: the client-leak advisory fix landed on `main` while this PR was in flight. It rebuilt `_on_disconnect` around the socket's room instead of the connecting query, and its regression test connects a socket with *no query at all*, handshakes it, and asserts the client is cleaned up afterwards. The fail-closed check refused exactly that handshake, so the connect never fired and the test timed out waiting for the cleanup.

We could have kept the strict check and weakened the advisory's test instead, but on a closer look the strictness buys nothing: the connecting query is attacker-controlled, so a forger who knows a `client_id` simply echoes it into the query and passes the strict check too. The check is protocol hygiene, not authentication — the protections that carry the weight (unguessable `client_id`, one handshake per socket, the pin) don't depend on it. So the check is back to your original form: a query that *carries* a `client_id` must match the handshake's, a missing one is tolerated — your caution about failing open was right for one more reason than we knew. The docstring now says why.

Description updated to match; CI runs against current `main` now.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Analysis%20and%20wording%20by%20the%20model%3B%20the%20decision%20to%20revert%20confirmed%20by%20the%20author%20after%20challenging%20it.&a=claude-code "claude-fable-5: Analysis and wording by the model; the decision to revert confirmed by the author after challenging it.")

falkoschindler (approved):

PR #6283, opened 2026-08-11
Add example about loading offline content

SaadZahem:
### Motivation

Nicegui comes with a nice collection of examples.
Each example covers a certain interesting and common use case.
Examples are great and save time for some people, me included.

I would like to extend this collection with one more.
There isn't an example covering how to make a webpage accessible offline.
That's, when the connection goes down, the user still has enough elements on screen to interact with saved content.

I added this as an example rather than request API change.
I hope this example will help someone someday.

### Implementation

This example achieves its goal using local client-side storage and a service worker.
The service worker runs in the background and caches the webpage the first time.
When the connection goes down, a plain and simple html document, that fetches saved items from local storage, will get rendered.

I have added a README and a screenshot of the offline page.
I have tested the example myself on localhost simulating getting offline by pausing the server.
The code was written by Opus 4.8 while the PR message was written by me after studying it.
I believe we don't add test cases for example code.

### Progress


evnchn:
@SaadZahem @falkoschindler I am not sure about this. IMHO this is pretty detached from NiceGUI. 

Any static HTTP server can theoretically host offline-ready content using a service worker. As well, NiceGUI _itself_ most definitely does NOT work fine offline since interactivity is computed from the server intentionally (unless you go wild like https://github.com/evnchn/nicegui-pyodide)

As such, I don't think it is necessary or beneficial for understanding to add this example. What do you think?

SaadZahem:
You're right to think that this example is detached from nicegui.
As we already know, [nicegui follows a "backend-first" approach](https://www.reddit.com/r/nicegui/comments/11wl7op/comment/jd6t8fk/?utm_source=share&utm_medium=web2x&context=3).
But I still think this example worth adding and helpful just because it answers a common question.

Common questions like _"How to add nicegui to an existing fastapi app?", "How to integrate a database?", or "How to authenticate?"_ all have helpful examples answering them.
This example tries to showcase **what it's like to load online content in a nicegui app** instead of just saying "No, nicegui is backend-first. Everyone should be connected."
And I think it's a common question, like, when I showed my [quiz-sharing app](github.com/SaadZahem/quizzable) to my casual friend (doesn't know about tech), he asked about whether the quizzes would be playable offline.

Unlucky for me, I couldn't find, as expected, any nicegui feature to help with [SSG](https://github.com/zauberzeug/nicegui/discussions/637).
Without SSG, `offline.html` had to be written in plain HTML, CSS, and JS.
And that's for just one page, the number scales up and the situation becomes trickier the more pages you have.
Returning to my app, if I wanted to to make quizzes available offline, I would have to rewrite 3 pages without nicegui.
Btw, I like your idea of using pyodide with nicegui.

I am eager to know what @falkoschindler thinks about this.

falkoschindler:
@SaadZahem Thanks for putting this together, and for the thoughtful reply to @evnchn — the "common questions deserve examples" argument is a fair one, and I did give it real consideration. But after trying the example I have to land on Evan's side, and I want to explain why with something concrete rather than just "it's detached from NiceGUI".

I ran it, saved a couple of articles, and stopped the server. Two things happened:

1. The **"Offline page" button does nothing.** It is wired to a server-side callback (`ui.navigate.to('/offline')` in `main.py`, lines 135-136), so the click is emitted into a dead socket and silently dropped. As a user I'm left clicking a button that visibly does nothing, with no hint about what to do instead.
2. The **"Connection lost. Trying to reconnect…"** overlay appears and stays. That's NiceGUI's own popup, shown on socket `disconnect` and only cleared when a reconnect succeeds. So the screen reads as "something is broken", which is the opposite of the impression an offline-ready app should give.

These aren't small bugs to patch — they're the architecture telling the truth. The service worker can only help on a **navigation**, and nothing inside a running NiceGUI page can trigger one once the socket is gone, because every NiceGUI interaction is a server round trip. That's why a reload reaches the offline page but no button ever can. `ui.link('Offline page', '/offline')` would fix symptom 1, since it renders a real `<a href>`, and the overlay could probably be worked around too. But even then the only discoverable path into the offline experience is "reload at the moment the app looks broken", which is exactly what the README's step 4 asks the user to figure out on their own.

That leads to my main concern with the example as a whole. It doesn't make a NiceGUI app work offline; it ships a second, hand-written non-NiceGUI app — 111 lines of plain HTML/CSS/JS with its own duplicated card styling — that renders data the NiceGUI app happened to put into `localStorage`. The NiceGUI contribution is one save button. Meanwhile the title ("Offline-ready reading list") and the README ("Keep a NiceGUI app usable without a network connection") promise the thing we can't deliver. You name the scaling problem yourself: three pages means rewriting three pages by hand. An example in this repo implicitly endorses a pattern as recommended, and I don't think we should endorse this one.

There is a smaller and honest example hiding in here, though, and I'd be happy to look at it: **"show a friendly offline page instead of the browser's error page."** No claim that the app keeps working, no `localStorage` reading list, no duplicated UI — just the service worker, a `FileResponse` route, and a branded "you're offline, reconnecting…" page with a retry button. That's a real thing people ask for, it's maybe 40 lines, and it doesn't over-promise. It may fit better as a documentation section than as an entry in `examples/`, and I'd want to think about placement, but the substance would be welcome.

For the bigger question of running NiceGUI client-side, Evan's [nicegui-pyodide](https://github.com/evnchn/nicegui-pyodide) is the interesting direction, and #637 is where the SSG discussion lives.

So: I'm going to close this one, but not because the question isn't worth answering — it is. Thanks for raising it, and please do come back with the narrower version if you're up for it. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Author%27s%20concerns%20and%20the%20decision%20to%20close%3B%20code%20diagnosis%20and%20wording%20from%20the%20model.&a=claude-code "claude-opus-5: Author's concerns and the decision to close; code diagnosis and wording from the model.")

PR #6285, opened 2026-08-13
Reject a line anchor on a fractional line instead of snapping it

Jepson2k:
*Opened by Claude Code on @Jepson2k's behalf.*

### Motivation

Spotted by @falkoschindler while reviewing #5991: `applyLineAnchors` takes its line number as given, without requiring a whole one.

A line of `2.5` passes the range check and reaches `doc.line(2.5)`, which resolves it to a neighbouring line instead of refusing it — so the anchor lands where the caller never asked for it and is mirrored back as if that had been the declaration. On `9f3f764b`, `{'inside': 3, 'fractional': 2.5}` reads back as `{'inside': 3, 'fractional': 3}`, with no exception and no warning.

The other line-number check in the file, `setLineTooltips`, runs its key through `parseInt` before the range check, so a fractional line is truncated there rather than refused. Aligning the two is out of scope here.

### Implementation

`_validate` now requires `isinstance(line, int)` alongside the existing `line < 1` check, so a fractional line raises `ValueError` right away, on both the constructor and the setter, exactly like a line below 1. Whether a line is whole does not depend on the document, so there is no reason to defer it to the browser; only "beyond the end" needs `doc.lines` and stays a JS-side warning. `applyLineAnchors` is unchanged from `main`.

The fractional case is one more `pytest.raises` in `test_anchors_out_of_range`; no browser round-trip needed. Seen to fail before the fix.

### Progress

- [x] Documentation has been added — the docstring listed what happens to a line past the end and to one below 1, but said nothing about a line that is not whole.
- [x] No breaking changes to the public API — `dict[str, int]` never admitted floats, though a caller passing one now gets a `ValueError` instead of a silently snapped anchor.

### DIFF
--- a/nicegui/elements/codemirror/line_anchors.py
+++ b/nicegui/elements/codemirror/line_anchors.py
@@ -56,6 +56,8 @@ class LineAnchorElement(Element):
         Lines beyond the end of the document are dropped on the JS side with a warning via NiceGUI's
         logger, just like ``line_tooltips``, so a read never reports a position that was not applied.
         A line below 1 is rejected right away with a ``ValueError``.
+        A line that is not a whole number is dropped the same way as one beyond the end,
+        since it cannot address a line without silently resolving to a neighbouring one.
 
         *Added in version 3.16.0*
         """
--- a/nicegui/elements/codemirror/line_anchors.py
+++ b/nicegui/elements/codemirror/line_anchors.py
@@ -56,7 +56,7 @@ class LineAnchorElement(Element):
         Lines beyond the end of the document are dropped on the JS side with a warning via NiceGUI's
         logger, just like ``line_tooltips``, so a read never reports a position that was not applied.
         A line below 1 is rejected right away with a ``ValueError``.
-        A line that is not a whole number is dropped the same way as one beyond the end,
+        A line at or above 1 that is not a whole number is dropped the same way as one beyond the end,
         since it cannot address a line without silently resolving to a neighbouring one.
 
         *Added in version 3.16.0*
--- a/nicegui/elements/codemirror/line_anchors.py
+++ b/nicegui/elements/codemirror/line_anchors.py
@@ -55,9 +55,7 @@ class LineAnchorElement(Element):
 
         Lines beyond the end of the document are dropped on the JS side with a warning via NiceGUI's
         logger, just like ``line_tooltips``, so a read never reports a position that was not applied.
-        A line below 1 is rejected right away with a ``ValueError``.
-        A line at or above 1 that is not a whole number is dropped the same way as one beyond the end,
-        since it cannot address a line without silently resolving to a neighbouring one.
+        A line below 1 or not a whole number is rejected right away with a ``ValueError``.
 
         *Added in version 3.16.0*
         """
@@ -98,5 +96,5 @@ class LineAnchorElement(Element):
 
 def _validate(anchors: dict[str, int]) -> None:
     for id_, line in anchors.items():
-        if line < 1:
-            raise ValueError(f'line_anchors: anchor {id_!r} has line {line}, but lines are 1-indexed')
+        if not isinstance(line, int) or line < 1:
+            raise ValueError(f'line_anchors: anchor {id_!r} has line {line!r}, but lines are 1-indexed whole numbers')
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -203,7 +203,7 @@ export default {
       const doc = this.editor.state.doc;
       const ranges = [];
       for (const [id, line] of Object.entries(anchors || {})) {
-        if (line >= 1 && line <= doc.lines) {
+        if (Number.isInteger(line) && line >= 1 && line <= doc.lines) {
           const pos = doc.line(line).from;
           ranges.push(new AnchorValue(id).range(pos, pos));
         } else {
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -203,7 +203,11 @@ export default {
       const doc = this.editor.state.doc;
       const ranges = [];
       for (const [id, line] of Object.entries(anchors || {})) {
-        if (Number.isInteger(line) && line >= 1 && line <= doc.lines) {
+        if (!Number.isInteger(line)) {
+          // doc.line() resolves a fractional line to a neighbouring one rather than refusing it,
+          // which would place the anchor where nobody asked for it.
+          logAndEmit("warning", `line_anchors: anchor ${JSON.stringify(id)} on line ${line} is not a whole line`);
+        } else if (line >= 1 && line <= doc.lines) {
           const pos = doc.line(line).from;
           ranges.push(new AnchorValue(id).range(pos, pos));
         } else {
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -203,11 +203,7 @@ export default {
       const doc = this.editor.state.doc;
       const ranges = [];
       for (const [id, line] of Object.entries(anchors || {})) {
-        if (!Number.isInteger(line)) {
-          // doc.line() resolves a fractional line to a neighbouring one rather than refusing it,
-          // which would place the anchor where nobody asked for it.
-          logAndEmit("warning", `line_anchors: anchor ${JSON.stringify(id)} on line ${line} is not a whole line`);
-        } else if (line >= 1 && line <= doc.lines) {
+        if (line >= 1 && line <= doc.lines) {
           const pos = doc.line(line).from;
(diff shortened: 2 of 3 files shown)


nicegui/elements/codemirror/line_anchors.py
```diff
@@ -56,6 +56,8 @@ def line_anchors(self) -> dict[str, int]:
         Lines beyond the end of the document are dropped on the JS side with a warning via NiceGUI's
         logger, just like ``line_tooltips``, so a read never reports a position that was not applied.
         A line below 1 is rejected right away with a ``ValueError``.
+        A line that is not a whole number is dropped the same way as one beyond the end,
+        since it cannot address a line without silently resolving to a neighbouring one.
```
evnchn: `_validate` runs the `line < 1` check before anything reaches the browser, so a fractional line *below* 1 never takes this path: `editor.line_anchors = {'x': 0.5}` raises `ValueError: line_anchors: anchor 'x' has line 0.5, but lines are 1-indexed` instead of being dropped with a warning. Scoping the sentence keeps it true alongside the line above it.

```suggestion
        A line at or above 1 that is not a whole number is dropped the same way as one beyond the end,
        since it cannot address a line without silently resolving to a neighbouring one.
```

evnchn (commented):

**Looks correct** — I checked the premise against the vendored CodeMirror bundle rather than taking it on trust, and it holds. One docstring wording nit suggested inline.

<details><summary><b>What I verified</b></summary>

Against head `31c1e22c`. Source-read unless marked *(ran it)*.

- **The bug is real.** In `nicegui/elements/codemirror/dist/index-ciMdZ5pn.js`, `Text.line(t)` throws only on `t < 1 || t > this.lines`, and `lineInner(t, isLine, ...)` returns the first line whose 1-based counter satisfies `i >= t`. So on a 3-line doc `doc.line(2.5)` returns line 3 with no error — exactly the silent snap this drops.
- **Guard order is right, and the split into two commits earns its keep.** `2.5` is *not* outside `[1, doc.lines]`, so `017dd5f3`'s `Number.isInteger(line) && ...` dropped the anchor but reported it as "out of range"; `31c1e22c` giving it its own branch is what makes the message honest.
- **No false rejections of legitimate input** *(ran it)*. Through NiceGUI's orjson converter, `int`, whole `float`, `Decimal` (`nicegui/json/orjson_wrapper.py:60-61`), `np.int64` and whole-valued `np.float32`/`np.float64` all serialize to `2` or `2.0` — every one of them `Number.isInteger`-true on the JS side. Only genuinely fractional values hit the new branch.
- **Range ordering is safe.** `RangeSet.of(effect.value, true)` sorts, so dropping an entry can't leave the set unordered.
- **No rebuild needed.** `component='codemirror.js'` is served directly; `dist/` holds only the CodeMirror bundle, which is untouched.
- **The test pins both commits, not just the tip.** Against `main` the mirror reads `{'inside': 3, 'fractional': 3}`, so `wait_for` fails; against `017dd5f3` the anchor is already dropped and it is the `assert_py_logger('is not a whole line')` line that fails on the stale "out of range" wording instead. Two assertions, one per commit.

**Considered, not asking for:** integer-ness is a document-independent check, so it *could* sit in `_validate` next to `line < 1` and fail fast even with no browser attached. I'm not pushing for it — `line_anchors` is annotated `dict[str, int]` (the new test needs `# type: ignore[dict-item]` to pass `2.5` at all), and keeping the drop client-side is what makes it consistent with `line_tooltips`.

</details>

evnchn:
Heads-up: this repo will soon enable GitHub's open-PR limit, capping contributors without write access at 3 open non-draft PRs. Drafts don't count and stay unlimited.

The reason is queue order. A non-draft PR reads as a request for review, and the ready list is long enough right now that things sit for weeks. A cap of 3 means you choose what gets looked at first, instead of the queue deciding by accident.

You have more than 3 open and ready. Could you pick 3 and convert the rest to draft? Drafts keep their history and you can flip them back as the first ones merge. Doing it before the setting lands avoids anything getting bounced.

nicegui/elements/codemirror/line_anchors.py
evnchn: Done by https://github.com/zauberzeug/nicegui/pull/6285/commits/005d44b3a31a96ea372beab4a34f746bc1845231

evnchn (dismissed):

**Approve.** I re-checked this from scratch rather than leaning on my August comment — including re-running the premise against the vendored bundle, which is where that comment could most easily have been wrong. It wasn't.

The bug is worse than "rejects fractional anchors": CodeMirror **silently ceils**. In node against `dist/index-ciMdZ5pn.js`, `Text.line()` guards only `t < 1 || t > this.lines`, then `lineInner` returns the first line whose 1-based counter satisfies `i >= t`. On a 3-line document `line(2.5)` → **line 3**, and `line(1.0001)` → **line 2**. End-to-end with your fix reverted, `{'inside': 3, 'fractional': 2.5}` reads back as `{'inside': 3, 'fractional': 3}` — the anchor lands somewhere nobody asked for and the mirror reports it as the declaration. Silent wrong answer, no exception. Dropping it with a warning is the right disposition and matches the sibling out-of-range branch.

<details><summary><b>Verified, not read off the description</b></summary>

**Guard order is load-bearing, not stylistic.** `line(5.5)` throws at 3 and 5 lines but resolves to **line 6** at 6 lines — so the `isInteger` check has to come first, or the fractional-in-bounds case snaps while the fractional-past-end case needs the range branch to catch a throw.

**Bug class, mapped through the real stack** (post-fix, 3-line doc): `2` applied · `2.0` **applied**, mirrors as `2` · `2.5` dropped, "is not a whole line" · `5.5` dropped, same · `True` dropped · `NaN` dropped · `10` dropped, "out of range [1, 3]" · `0.5` raises `ValueError` in Python before reaching JS. So `_validate` covers `< 1` and JS covers the rest, with **no false rejections** — `int`, whole `float`, `Decimal`, `np.int64`, `np.float32/64` all serialise to something `Number.isInteger`-true. `CM.RangeSet.of(effect.value, true)` passes the sort flag, so dropping an entry can't leave the set unsorted.

**Seen to fail, once per commit.** Whole fix reverted → red on `screen.wait_for`; I instrumented it to confirm the mirror really held `{'fractional': 3}`, since the bare message is only "Condition not met". Second commit alone reverted → red on `assert_py_logger`: `Expected regex "is not a whole line" but got "…on line 2.5 out of range [1, 3]"`. Two live assertions, one per commit — genuinely well built.

**Three weeks of drift: clean.** 15 commits on main since 19 Aug, exactly one touching this element (`969af1bc`, the editor HTML-id fix) in a different hunk. Trial-merged `upstream/main` in: auto-merge clean, `applyLineAnchors` hunk intact. Gates re-run **on the merged result** — pre-commit pass, mypy clean (265 files), pylint 10.00/10, `test_codemirror_line_anchors.py` 13 passed 0 skipped, `test_codemirror.py` 19 passed.

</details>

One factual correction for the description, not the code: it says "`dict[str, int]` never admitted floats." It admits `bool`. mypy accepts `{'a': True}` against `dict[str, int]` while rejecting `{'b': 2.5}`, and `_validate` lets it through since `True < 1` is False. Pre-fix that silently anchored on **line 1**; post-fix it is dropped with "on line true is not a whole line". Your new behaviour is better — the claim just doesn't cover that case.

And a leave-or-take one for @falkoschindler to rule on rather than something I'd hold this up for: the new test could probably be three lines appended to `test_anchors_out_of_range` directly above it, which already uses the identical document, page setup and `_wait_for_editor`, and whose docstring already enumerates two dispositions. Measured cost of the standalone version: **1.1–1.3 s per suite run** (13 passed in 15.60 s vs 12 in 14.55 s with it deselected). Folding it would need a rename, which is free. The fix is right either way — this is a house-convention call, not a defect in your work.

Nice catch on a bug that produced no error message at all.

falkoschindler (approved):
Thanks for catching this, @Jepson2k, and to @evnchn for verifying the CodeMirror ceiling behaviour against the bundle.

I pushed one follow-up commit (a8ccd46b8) that moves the whole-number check from `applyLineAnchors` into `_validate`, next to the existing `line < 1` check. Whether a line is whole does not depend on the document, so it can fail fast in Python like a line below 1 does, on both the constructor and the setter; only "beyond the end" needs `doc.lines` and stays a JS-side warning. That also makes the fractional case one more `pytest.raises` in `test_anchors_out_of_range` instead of a separate Screen test, and `codemirror.js` is back to what it is on `main`. Description and title are updated accordingly.

Approving with that commit in. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Review%20findings%20and%20wording%20from%20the%20model%3B%20the%20decision%20to%20move%20the%20check%20into%20Python%20came%20from%20the%20maintainer.&a=claude-code "claude-fable-5-1: Review findings and wording from the model; the decision to move the check into Python came from the maintainer.")

PR #6287, opened 2026-08-14
Fix template slots being remounted on every render (#6284)

### LINKED ISSUE #6284: Bug/edge-case with removable chips and custom slot
### First Check

### Example Code

```python
from nicegui import ui

select = ui.select(
    ["Apple", "Banana", "Cherry"],
    multiple=True,
    value=["Apple", "Banana", "Cherry"],
    with_input=True,
).props("use-chips dense").classes("w-64")

select.add_slot(
    "selected-item",
    """
    <q-chip square removable @remove="props.removeAtIndex(props.index)">
        <span>{{ props.opt.label }}</span>
    </q-chip>
    """,
)

ui.label(
    "Reload, then click a chip's x without clicking the field first: nothing happens. "
    "Click it again and it's removed."
)

ui.run()
```

### Description

Have to click the remove button ("x" icon) on the chips twice for it to actually work (first click focuses the select component and then second click works). This only happens with the custom slot, otherwise a single click works as expected to remove the selected option even without focus.

### NiceGUI Version

3.13.0

### Python Version

3.12.3

### Browser

Firefox

### Operating System

Linux

### Additional Context

Claude thinks:
> The cause is in NiceGUI's `add_slot` function. Look at the `renderRecursively` function in `nicegui.js`. This function builds a new Vue component object for a slot on every call. Vue then cannot see that the object is the same component across renders. Vue removes the old chip element from the DOM. Vue adds a new chip element in its place.
> 
> The first click on the icon makes Quasar add "focused" CSS classes to the field. This action makes Quasar render the field again. The new render removes and adds the chip element while you hold the mouse button down. The browser cannot find the original chip element when you release the mouse button. The browser does not send the click event.
> 
> Quasar does not render the field again during the second click. The chip element stays in the DOM. The click event fires. The chip is removed.

and it generated the following patch for me which fixed the issue when I tested, but I have no idea of the larger impact this change would have:
```patch
--- a/nicegui.js
+++ b/nicegui.js
@@ -227,6 +227,20 @@
   }
 }
 
+// [Changed] This function stores one Vue component object for each template string.
+// The renderRecursively function below reuses this object on every call for the same template.
+// Before this fix, each call built a new object. Vue could not see it as the same component.
+// Vue then removed the old slot content from the DOM and added new content in its place.
+// This removal could happen while a user held the mouse button down on a chip's remove icon.
+// The removal could stop the browser from sending the click event.
+const slotComponentCache = new Map();
+function getSlotComponent(template) {
+  if (!slotComponentCache.has(template)) {
+    slotComponentCache.set(template, { props: { props: { type: Object, default: {} } }, template });
+  }
+  return slotComponentCache.get(template);
+}
+
 function renderRecursively(elements, id, propsContext) {
   const element = elements[id];
   if (element === undefined) {
@@ -304,10 +318,7 @@
       if (data.template) {
         rendered.push(
           Vue.h(
-            {
-              props: { props: { type: Object, default: {} } },
-              template: data.template,
-            },
+            getSlotComponent(data.template), // [Changed] The old code built a new object literal on every call.
             {
               props: props,
             },

```

I also haven't yet tried with NiceGUI v3.15.0 - hope to do that soon.

SaadZahem:
*Filed by Claude Code on SaadZahem's behalf*

### Motivation

Fixes #6284.

A removable chip in a `ui.select` with a custom `selected-item` slot swallowed the first click on its "x" — you had to click twice. More generally, **any** template slot lost its entire local DOM and component state whenever the surrounding element re-rendered for an unrelated reason. For example, typed text disappears here when the button is pressed:

```python
element = ui.element()
element.add_slot('default', '<input placeholder="type here">')
ui.button('update', on_click=lambda: element.classes('bg-red-100'))
```

**Mechanism:** In `renderRecursively` the slot render function passed a fresh object literal as the component definition to `Vue.h()` on every invocation. Vue identifies a vnode's component by the identity of its `type`, so each render looked like a different component: Vue unmounted the old slot subtree and mounted a new one, throwing away its DOM/component state and recompiling the template. For the chip, the first click makes Quasar add its focus classes and re-render the field, which destroys and recreates the chip's DOM node while the mouse button is held down; on `mouseup` the browser can't find the original node, so no `click` is synthesized and `@remove` never fires.

### Implementation

Give each slot template a stable component identity so Vue patches the slot in place instead of remounting it:

- Cache the component definition on the slot entry itself (`data._component ??= {...}`) rather than in a module-level map, tying its lifetime to the element (no monotonic leak with dynamically generated templates).
- Because a server update replaces the whole element object (`this.elements[id] = element`) — and with it the `slots` object — the cache is also carried over from the previous element during `update`, but only while the template string is unchanged. If the template changes, a new component is built and the old one is dropped.

Together these fix both the client-side re-render case (chip focus) and the server-update case (typed input surviving `.classes()`).

Two trade-offs worth noting: attaching the cache to the element (as suggested) means two elements sharing an identical template string no longer share a compiled component — each keeps its own — which trades a small amount of duplicate compilation for the leak-free lifetime. And when an element's event listeners change, `update` deletes and re-creates it, so the carry-over doesn't apply and the slot remounts by design in that path.

Note on the [related question](https://github.com/zauberzeug/nicegui/issues/6284) about slot closures: the slot functions are recreated on every `renderRecursively` call, but that doesn't cause remounts — Vue diffs the vnodes the closure returns by type, and stabilizing the component type is what's needed. So no change there.

### Progress

- [x] Pytests have been added/updated:
  - `test_remove_chip_with_custom_slot` — the original symptom: a single click on a removable chip's "x" removes the option.
  - `test_template_slot_survives_rerender` — a slot's typed input survives an unrelated re-render.

  Both were confirmed to fail on `main` and pass with this change.

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -307,16 +307,17 @@ function renderRecursively(elements, id, propsContext) {
     slots[name] = (props) => {
       const rendered = [];
       if (data.template) {
+        // Cache the component definition on the slot so its identity is stable across renders.
+        // Otherwise a fresh object literal makes Vue treat each render as a new component type,
+        // unmounting and remounting the slot's DOM (losing local state) and recompiling the template.
+        data._component ??= {
+          props: { props: { type: Object, default: {} } },
+          template: data.template,
+        };
         rendered.push(
-          Vue.h(
-            {
-              props: { props: { type: Object, default: {} } },
-              template: data.template,
-            },
-            {
-              props: props,
-            },
-          ),
+          Vue.h(data._component, {
+            props: props,
+          }),
         );
       }
       const children = data.ids.map((id) => renderRecursively(elements, id, props || propsContext));
@@ -527,6 +528,16 @@ function createApp(elements, options) {
               }
             }
             delete element.preserved_props;
+            // Carry over compiled slot components so their identity stays stable across updates
+            // (as long as the template is unchanged), letting Vue patch the slot instead of
+            // remounting it and discarding its local DOM state. See renderRecursively.
+            const oldSlots = this.elements[id]?.slots;
+            for (const [name, data] of Object.entries(element.slots ?? {})) {
+              const oldData = oldSlots?.[name];
+              if (oldData?._component && oldData.template === data.template) {
+                data._component = oldData._component;
+              }
+            }
             this.elements[id] = element;
           }
 
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -310,10 +310,10 @@ function renderRecursively(elements, id, propsContext) {
         // Cache the component definition on the slot so its identity is stable across renders.
         // Otherwise a fresh object literal makes Vue treat each render as a new component type,
         // unmounting and remounting the slot's DOM (losing local state) and recompiling the template.
-        data._component ??= {
+        data._component ??= Vue.markRaw({
           props: { props: { type: Object, default: {} } },
           template: data.template,
-        };
+        });
         rendered.push(
           Vue.h(data._component, {
             props: props,
--- a/tests/test_element.py
+++ b/tests/test_element.py
@@ -541,3 +541,20 @@ def test_even_special_elements_have_an_html_id(screen: Screen):
     screen.open('/')
     screen.click('Check IDs')
     screen.should_contain('All IDs found')
+
+
+def test_template_slot_survives_rerender(screen: Screen):
+    @ui.page('/')
+    def page():
+        element = ui.element()
+        element.add_slot('default', '<input placeholder="type here">')
+        ui.button('Update', on_click=lambda: element.classes('bg-red-100'))
+
+    screen.open('/')
+    element = screen.selenium.find_element(By.XPATH, '//input[@placeholder="type here"]')
+    element.send_keys('hello')
+    assert element.get_attribute('value') == 'hello'
+
+    screen.click('Update')  # #6284: an unrelated re-render must not discard the slot's DOM
+    screen.wait(0.5)
+    assert element.get_attribute('value') == 'hello'
--- a/tests/test_select.py
+++ b/tests/test_select.py
@@ -78,6 +78,23 @@ def test_multi_select(screen: Screen):
     screen.should_contain("['Bob']")
 
 
+def test_remove_chip_with_custom_slot(screen: Screen):
+    @ui.page('/')
+    def page():
+        s = ui.select(['Apple', 'Banana', 'Cherry'],
+                      multiple=True, value=['Apple', 'Banana', 'Cherry'], with_input=True).props('use-chips')
+        s.add_slot('selected-item',
+                   '<q-chip removable @remove="props.removeAtIndex(props.index)">{{ props.opt.label }}</q-chip>')
+        ui.label().bind_text_from(s, 'value', backward=str)
+
+    screen.open('/')
+    screen.should_contain("['Apple', 'Banana', 'Cherry']")
+
+    # #6284: focusing the field on the first click must not remount the chip and swallow the click
+    screen.find_by_css('.q-chip__icon--remove').click()
+    screen.should_contain("['Banana', 'Cherry']")
+
+
 def test_changing_options(screen: Screen):
     @ui.page('/')
     def page():


nicegui/static/nicegui.js
```diff
@@ -307,16 +307,17 @@ function renderRecursively(elements, id, propsContext) {
     slots[name] = (props) => {
       const rendered = [];
       if (data.template) {
+        // Cache the component definition on the slot so its identity is stable across renders.
+        // Otherwise a fresh object literal makes Vue treat each render as a new component type,
+        // unmounting and remounting the slot's DOM (losing local state) and recompiling the template.
+        data._component ??= {
+          props: { props: { type: Object, default: {} } },
+          template: data.template,
+        };
```
evnchn: *Posted by Claude Code on evnchn's behalf.*

`this.elements` is Vue's reactive `data()`, so anything stored under `element.slots[name]` is read back as a Proxy — which means `Vue.h()` receives a reactive component here. Dev builds warn on every render (0 on `main` → 8 on first load → 19 after a chip removal); production stays quiet but still writes the compiled `render` through the proxy (`_component.__v_isReactive === true` under `prod_js=True`).

`markRaw` keeps the identity stability this PR is after — and the carry-over in `update` still works, since a marked object read back through reactive state stays raw.

```suggestion
        data._component ??= Vue.markRaw({
          props: { props: { type: Object, default: {} } },
          template: data.template,
        });
```

evnchn (commented):

**TL;DR: the diagnosis and the fix are both right — I reproduced the bug, confirmed both new tests fail on `main` for the right reasons, and confirmed the fix in a real browser.**

One thing to change before merge, a one-word diff:

1. `this.elements` comes from Vue's `data()`, so it is **deeply reactive**. Caching the component definition on the slot entry therefore stores it *inside* reactive state, and `Vue.h()` receives a reactive proxy. Wrap it in `Vue.markRaw(...)`.

Everything else I probed held up: the `oldData.template === data.template` guard correctly drops a stale component on a runtime template swap, the carry-over survives `markRaw`, and the template-slot-heavy suites stay green.

<details>
<summary><b>Finding 1 (major) — the cached component ends up in Vue's reactive state</b></summary>

`data()` returns `elements` directly (`nicegui/static/nicegui.js`, `createApp`), so every `element`, and every `element.slots[name]`, is reached through a reactive proxy. `data._component ??= { … }` therefore writes the component-options object into reactive state, and reading it back yields a proxy.

Measured in a browser, `main` vs this branch, using the PR's own repro plus the chip case:

| | `main` | this PR | this PR + `markRaw` |
|---|---|---|---|
| Vue warnings on first load (`prod_js=False`) | 0 | 8 | 0 |
| after one unrelated re-render | 0 | 12 | 0 |
| after removing one chip | 0 | 19 | 0 |
| `slots[name]._component.__v_isReactive` (`prod_js=True`) | n/a | `true` | `false` (`__v_skip: true`) |

The warning is Vue's own:

> [Vue warn]: Vue received a Component that was made a reactive object. This can lead to unnecessary performance overhead and should be avoided by marking the component with `markRaw` or using `shallowRef` instead of `ref`.

Two notes on scope, so the severity reads honestly:

- The **warning** is dev-only — Vue's dev build both warns and calls `toRaw(type)`, and `prod_js=True` (the default) strips that whole branch. So end users never see it; developers running `prod_js=False` see it grow on every render.
- The **reactivity is not** dev-only. Probing the live page under `prod_js=True` shows `_component.__v_isReactive === true`, and `Object.keys(_component)` is `['props', 'template', 'render']` — Vue's compiled render function has been written *through* the proxy into reactive state. That is the part `markRaw` actually removes.

Fix is `Vue.markRaw(...)` around the cached object — left as an applicable suggestion on the four lines in question.

Verified: warnings drop to 0, `__v_skip: true`, the chip still removes on one click, the slot `<input>` still keeps its text across a re-render, and the carry-over in `update` still preserves the same raw object (a marked object read back through reactive state stays raw, so `data._component = oldData._component` is unaffected).

</details>

<details>
<summary><b>Finding 2 (minor, optional) — the cache could be built outside the render function</b></summary>

Even with `markRaw`, the first render of each slot mutates state that Vue is currently rendering. `replaceUndefinedAttributes` already normalises `element.slots` (it is where the `default` slot is synthesised), runs exactly once per element create/update, and is not inside a render pass — so it is a natural home for building `_component`, and it would put creation and carry-over near each other instead of ~220 lines apart.

Defensible as-is, and it is a bigger diff than the bug warrants. Only worth doing if you are touching this code anyway.

</details>

<details>
<summary><b>What I verified, and how</b></summary>

Worktree at `671e5da2`, base `upstream/main` at `17ad9005`.

**Seen-to-fail, per assertion (`git show upstream/main:nicegui/static/nicegui.js > …`, tests unchanged):**

| test | unpatched | which assertion fired | patched |
|---|---|---|---|
| `test_remove_chip_with_custom_slot` | FAIL | `should_contain("['Banana', 'Cherry']")` → `NoSuchElementException` — i.e. the click really was swallowed | PASS |
| `test_template_slot_survives_rerender` | FAIL | `StaleElementReferenceException` on the held `<input>` — i.e. the node really was remounted | PASS |

Both failures are the reported symptom, not an incidental error.

**Suites (macOS, Python 3.14, Chrome 151):**

| command | result |
|---|---|
| `pytest tests/test_select.py tests/test_element.py` | 86 passed |
| same, with `markRaw` applied | 86 passed |
| `pytest tests/test_table.py tests/test_unocss.py` (heaviest template-slot users) | 31 passed |
| `pre-commit run --files nicegui/static/nicegui.js tests/test_element.py tests/test_select.py` | passed |

(One `test_no_cyclic_references_when_deleting_elements` timeout appeared in a single combined run and did not reproduce in 3 isolated runs or on re-run of the same command — local flake, not attributable to this change.)

**Browser, both `prod_js` settings**, driving the PR's own repro plus the chip case: single-click chip removal works, typed text in a slot `<input>` survives `element.classes(...)`.

**Edge case probed — runtime template swap.** Calling `add_slot('default', …)` a second time with a different template and then `element.update()` correctly re-renders the new template (`TEMPLATE-ONE` → `TEMPLATE-TWO`), so the `oldData.template === data.template` guard does drop the stale component rather than pinning the old one.

</details>

<details>
<summary><b>Checked and found sound (no action)</b></summary>

- **Carry-over loop.** No stale-component hole found: a removed slot goes away with the old element object, a changed template is guarded, and changed slot *children* do not need a new component. `element.slots` is always populated by `replaceUndefinedAttributes` before the loop runs, so the `?? {}` is belt-and-braces rather than load-bearing.
- **`eventListenersChanged` path.** That path deletes the element outright, so `this.elements[id]` is gone and no carry-over happens — the slot remounts. That is consistent with what the path already does to the rest of the element, and the PR says so.
- **Lifetime.** `_component` hangs off the element and dies with it; nothing points back at the slot data, so no cycle and no monotonic growth. Versus `main` this is strictly *less* compilation, since `main` recompiles the template on every remount.
- **Tests.** `Screen` is the right fixture here — the bug is Vue/DOM remount behaviour, not Python API behaviour — and both tests assert user-visible outcomes rather than internals. `screen.wait(0.5)` after a click matches the existing shape in `test_element.py` (`test_move_slots`).
- **PR description** follows the Motivation / Implementation / Progress template, and names both trade-offs rather than hiding them.
- **CI** is green on the branch.

</details>

<details>
<summary><b>Second opinion</b></summary>

Ran the diff past GPT-5.2-Codex as an independent skeptic. It reached Finding 1 independently, found no hole in the carry-over loop or the `eventListenersChanged` interaction, and rated the tests appropriate. Its useful correction to my first pass: Vue's dev build calls `toRaw(type)` immediately after warning, so the vnode type is raw *in dev* — which is what pushed me to check the production build, where the proxy does survive (the table above). Its score: 8/10, single highest-leverage fix `markRaw`.

</details>

**Ask:** apply the `markRaw` one-liner (Finding 1); Finding 2 is take-or-leave. With that in, this looks good to me — the mechanism write-up in the PR body is accurate and the tests genuinely guard the regression.

PR #6288, opened 2026-08-15
Fix ui.sub_pages staying in SPA for routes registered via app.include…

### LINKED ISSUE #6281: `ui.sub_pages` stays in the SPA instead of loading pages registered via `app.include_router()`
*Filed by Claude Code on @jnossan's behalf, who reported it in #5777.*

### First Check

### Example Code

```python
from nicegui import APIRouter, app, ui

router = APIRouter(prefix='/other')


@router.page('/')
def other_page():
    ui.label('Other')


app.include_router(router)


@ui.page('/')
def index():
    ui.button('Go to /other', on_click=lambda: ui.navigate.to('/other'))
    ui.link('Go to /other', '/other')


ui.run()
```

### Description

**Expected:** clicking either the button or the link leaves the SPA and performs a full page load of `/other`, the same way it does for a page registered with plain `@ui.page('/other')`.

**Actual:** `ui.sub_pages` keeps handling the path, finds no matching sub page, and shows its 404 inside the container. The URL `/other` is perfectly valid on the server — `GET /other` answers `307` to `/other/`, which answers `200` — so only the client-side navigation is broken.

This is the same symptom as #5777, which was fixed by `SubPagesRouter._other_page_builder_matches_path()` in `nicegui/sub_pages_router.py`. That check asks every route in `core.app.routes` whether it matches the target path, and falls back to `client.open()` when one does. It misses routes coming from an `APIRouter` for two independent reasons:

**1. The trailing slash.** `APIRouter(prefix='/other')` plus `@router.page('/')` registers the path `/other/`. `Route.matches()` is literal and knows nothing about FastAPI's `redirect_slashes`, so `/other` yields `Match.NONE` and the router assumes the path is its own. Navigating to `/other/` works around it. This is not an exotic spelling: our own [documentation](https://nicegui.io/documentation/page#modularize_with_apirouter) and the [modularization example](https://github.com/zauberzeug/nicegui/blob/main/examples/modularization/api_router_example.py) both link to the prefix without a trailing slash.

**2. FastAPI's lazy router inclusion.** Since FastAPI 0.141, `app.include_router()` appends a single `_IncludedRouter` wrapper to `app.routes` instead of the individual routes. The wrapper is not a `starlette.routing.Route`, so the `isinstance(route, Route)` filter drops it and the check sees no page routes at all — with FastAPI 0.141, even `/other/` stays inside the SPA:

```console
$ python -c "
from nicegui import APIRouter, app, ui, core
router = APIRouter(prefix='/other')
@router.page('/')
def other_page(): ui.label('Other')
app.include_router(router)
print([getattr(r, 'path', r) for r in core.app.routes][-1:])
"
[_IncludedRouter(original_router=<nicegui.api_router.APIRouter object at 0x102e3b2f0>, …)]
```

Routing itself is unaffected — FastAPI resolves requests through the wrapper — but every place in NiceGUI that inspects `core.app.routes` goes blind. Besides the check above, that is `App.remove_route()` (`nicegui/app/app.py`), the `endpoint_documentation` loop in `nicegui/ui_run.py`, and `nicegui_reset_globals()` in `nicegui/testing/general.py`. Our `uv.lock` pins FastAPI 0.136.3, so CI never exercises this, while `pyproject.toml` allows `fastapi>=0.109.1` and users get 0.141.

### Workaround

Pass the page function instead of the path — `ui.navigate.to()` only hands string targets to the SPA router:

```py
ui.button('Go to /other', on_click=lambda: ui.navigate.to(other_page))
```

### NiceGUI Version

3.15.0

### Python Version

3.12.9

### Browser

Chrome

### Operating System

macOS

### Additional Context

Verified with browser tests against the current `main` (`3.15.0.post52.dev0+a2371a6f5`, FastAPI 0.136.3): navigating to `/other` by button or by link fails, `/other/` and the page-function form succeed. Cause 2 verified separately against released NiceGUI 3.15.0 with FastAPI 0.141.1, where `_other_page_builder_matches_path()` returns `False` for both `/other` and `/other/`.

DineshThumma9:
Filed by  Gemini Flash 3.6  on Dinesh Thumma's behalf

### Motivation

Fixes #6281

When registering FastAPI routers via `app.include_router(router)`, `ui.sub_pages` treated the registered path as an internal SPA sub-page instead of triggering a full page load to the backend route.

The cause is a trailing-slash mismatch: `APIRouter(prefix='/other')` with `@router.page('/')` registers `/other/`, while a link navigates to `/other`. Starlette reports `Match.NONE` for `/other` and `Match.FULL` for `/other/`, so `_other_page_builder_matches_path()` concluded no other page builder owned the path and kept handling it in the SPA.

A second, independent case failed the same way: `_other_page_builder_matches_path()` passed the raw path to `Route.matches`, so `/other?x=1` matched no other page route and navigating to it rendered `404: sub page /other?x=1 not found` even though `/other` is registered.

### Implementation

Three changes:

1. **Registration fix (root cause)** — `page` now registers a path of `'/'` on a prefixed `APIRouter` as the bare prefix, so `@router.page('/')` with `prefix='/other'` creates the route `/other` instead of `/other/` (`nicegui/page.py`). Route matching, `Client.page_routes`, link hrefs, and favicon routes thereby agree on one canonical path, and the existing `Match.FULL` check in `_other_page_builder_matches_path()` just works. Requests to `/other/` keep working via Starlette's `redirect_slashes` redirect. This replaced an earlier trailing-slash probe on the matching side — see [this comment](https://github.com/zauberzeug/nicegui/pull/6288#issuecomment-5439627101) for why.
2. **Query-string and fragment strip** — `_other_page_builder_matches_path()` matches against `urlsplit(path).path` (`nicegui/sub_pages_router.py`), so `/other?x=1` and `/other#section` resolve as `/other`, and skips protocol-relative links like `//host/path` which point to another host, not to one of our routes.
3. **Initial fragment scrolling** — the browser's native fragment scroll can fire before Vue has rendered the target, so `ui.sub_pages` now scrolls to the current `location.hash` explicitly once it is mounted (`nicegui/elements/sub_pages.js`). This race surfaced as flakiness in `test_sub_pages_with_url_fragments`, which now also polls for scroll positions instead of waiting fixed timeouts and lets smooth-scroll animations settle before follow-up clicks.

Deliberate behavior change: the canonical URL of pages registered as `'/'` on a prefixed router becomes `/other` instead of `/other/` (old URLs are 307-redirected), and relative links on those pages resolve against the parent path now — worth a line in the release notes. Note that the 307 redirect only covers full HTTP requests: a hardcoded link to `/other/` *inside* a `ui.sub_pages` page is intercepted client-side and now shows the SPA 404 — the same behavior such trailing-slash links have always had for plain `@ui.page` routes, so slash handling is now uniform across both registration styles.

**Regression tests** in `tests/test_sub_pages.py`, asserting content and final URL of a real full-page navigation away from `ui.sub_pages`:

- `test_navigate_from_sub_pages_to_api_router_page` — an `APIRouter` route registered via `app.include_router()`.
- `test_navigate_from_sub_pages_to_page_with_query_or_fragment` — a plain `@ui.page` reached with a query string or a fragment.

### Progress


evnchn (changes requested):

**TL;DR:**

Your diagnosis is right, and it's the hard part — the trailing-slash mismatch really is why navigating to an `include_router` page falls through, and your test genuinely guards it (it fails without the fix).

The recommendation is to **ship only that**: keep `clean_path` and the `alt_path` probe in `sub_pages_router.py`, and drop `get_routes()` and the three files it touches. I verified your test still passes with all of `get_routes()` reverted, and that the full `tests/test_sub_pages.py` suite is green on that minimal version.

Dropping it also removes a regression for `app.mount()` users and a piece of machinery that doesn't do what it's meant to on current FastAPI — both below.

<details><summary><b>Why the trailing-slash probe alone is enough</b></summary>

`APIRouter(prefix='/other')` + `@router.page('/')` registers `/other/`, while the link navigates to `/other`. Starlette reports `Match.NONE` for `/other` and `Match.FULL` for `/other/` — so the probe is the whole fix.

Tested both directions on this branch:

```
alt_path kept, every get_routes() usage reverted to the old isinstance(route, Route) filter:
  tests/test_sub_pages.py::test_navigate_from_sub_pages_to_api_router_page   PASSED

alt_path removed as well:
  AssertionError: Could not find "Other"                                     FAILED
```

Full suite on that minimal version: `55 passed` in `tests/test_sub_pages.py`.

One aside: `clean_path = path.split('?')[0]` looks right to me, but it's an independent fix from the trailing-slash one — worth its own line in the PR description so it doesn't read as incidental.

</details>

<details><summary><b>get_routes() breaks ui.sub_pages for anyone using app.mount()</b></summary>

`get_routes()` recurses into mounted sub-applications through the `hasattr(route, 'routes')` branch and returns their **mount-relative** paths. A sub-app mounted at `/api` with a route `/detail` contributes a bare `/detail` candidate, with no `/api` prefix — which then falsely matches in `_other_page_builder_matches_path`.

On the pinned FastAPI 0.136.3:

```
app.mount('/api', sub_app)   # sub_app has GET '/'

old  [route for route in core.app.routes if isinstance(route, Route)]
  -> ['/', '/docs', '/docs/oauth2-redirect', '/openapi.json', '/redoc']
new  helpers.get_routes(core.app.routes)
  -> ['/', '/', '/docs', '/docs', '/docs/oauth2-redirect', '/docs/oauth2-redirect',
      '/openapi.json', '/openapi.json', '/redoc', '/redoc']
```

Every path appears twice; the second copy is the mounted app's, stripped of its prefix.

It only bites when `ui.sub_pages` does *not* resolve the path, because `_handle_navigate` short-circuits on `not has_any_unresolved_path(client)` first — but in that case it replaces a working in-app 404 with a full page load. Repro, which **passes on main (8fb479bc) and fails on this branch**:

```python
from fastapi import FastAPI
from nicegui import app, ui


def test_sub_pages_404_survives_app_mount(screen: Screen):
    sub_app = FastAPI()

    @sub_app.get('/detail')
    def sub_detail():
        return {'ok': True}

    app.mount('/api', sub_app)

    @ui.page('/')
    @ui.page('/{_:path}')
    def index():
        ui.sub_pages({'/': lambda: ui.label('Index')})
        ui.link('Go to detail', '/detail')

    screen.open('/')
    screen.should_contain('Index')
    doc_id = screen.selenium.execute_script('return window.documentId;')

    screen.click('Go to detail')
    screen.wait(1.0)
    new_doc_id = screen.selenium.execute_script('return window.documentId;')

    assert new_doc_id == doc_id, 'navigation triggered a full page load instead of staying in the SPA'
```

```
main (8fb479bc):  1 passed in 2.30s
this branch:      AssertionError: navigation triggered a full page load instead of staying in the SPA
                  - b704a0b5-2eca-4676-927e-e00ee2723a62
                  + 3d941ee4-b8c2-4a9c-9048-59e88c790936
```

</details>

<details><summary><b>The effective_candidates / original_route branches are unexercised in CI, and wrong when they do run</b></summary>

On the locked FastAPI (0.136.3), `include_router()` still flattens everything into plain `APIRoute` objects, so those branches never execute — CI cannot reach them.

On FastAPI ≥ 0.141, where `_IncludedRouter` exists, the `original_route` branch unwraps the candidate and **drops the prefix**. Verified on fastapi 0.141.1 / starlette 1.6.0:

```python
r = APIRouter()
@r.get('/leaf')
def leaf(): ...
app.include_router(r, prefix='/a')
app.include_router(r, prefix='/b')

get_routes(app.routes)  ->  ['/leaf', '/leaf']    # prefixes gone
```

```
candidate '/leaf' vs '/a/leaf': Match.NONE     <- the real target matches nothing
candidate '/leaf' vs '/leaf':   Match.FULL     <- an unrelated sub-page falsely matches
```

The prefix is applied by `_EffectiveRouteContext.matches()`; unwrapping to `.original_route` throws it away — calling `candidate.matches(scope)` on the candidate itself would keep it.

So for `include_router(router, prefix=...)` the reported issue would still be broken on current FastAPI, and a sub-page whose name collides with an unprefixed route gains a new false positive. Your test passes because it puts the prefix on the `APIRouter(prefix=...)` constructor, where it's baked into the route path — the include-time `prefix=` case isn't covered.

</details>

<details><summary><b>Heads-up: CI has never run on this PR, and there are five lint failures</b></summary>

First-time contributors need a maintainer to approve workflow runs, so you've had no feedback loop. Four of these five disappear on the minimal version above; listing them in case you keep the current shape.

```
$ pre-commit run --files nicegui/helpers/functions.py nicegui/sub_pages_router.py \
                        nicegui/testing/general.py tests/test_sub_pages.py
ruff check...............................................................Failed
F401 `starlette.routing.Route` imported but unused   (nicegui/sub_pages_router.py:8)
F401 `starlette.routing.Route` imported but unused   (nicegui/testing/general.py:6)
autopep8.................................................................Failed
- files were modified by this hook
```

`ruff-check` runs with `--fix`, but `F401` isn't in the `fixable` list in `pyproject.toml`, so it's a hard failure rather than an autofix. `autopep8` rewrites two files: a trailing blank line at the end of `nicegui/helpers/functions.py`, and three consecutive blank lines at `tests/test_sub_pages.py:1349-1351` (PEP 8 wants two).

```
$ pylint ./nicegui
nicegui/sub_pages_router.py:8:0:   W0611: Unused Route imported from starlette.routing
nicegui/testing/general.py:6:0:    W0611: Unused Route imported from starlette.routing
nicegui/helpers/functions.py:95:0: C0305: Trailing newlines
nicegui/helpers/functions.py:81:4: C0415: Import outside toplevel (starlette.routing.Route)
exit code: 20
```

`mypy ./nicegui` is clean. For `C0415` specifically: `starlette` is a hard dependency and there's no import cycle — hoisting the import to module scope and re-running gives `pylint` exit 0 and `mypy` clean.

</details>

<details><summary><b>Minor notes</b></summary>

- `get_routes()` has no cycle guard; a self-referential mount (`app.mount('/self', app)`) recurses to `RecursionError`. Moot if the helper goes away.
- The `original_router` branch is unreachable — anything with `original_router` also has `effective_candidates`, which is tested first.
- `assert screen.current_path in ('/other', '/other/')` accepts either outcome. Asserting that `window.documentId` changed (the idiom in `tests/test_reconnect.py`) would pin the full-page-load behaviour you actually care about.
- The PR description mentions changes to `nicegui/app/app.py` and `nicegui/ui_run.py`; neither is in the diff.
- One claim I checked and **withdrew**: that recursing into mounts would make the reset loop in `nicegui/testing/general.py` strip `/docs` and friends from the main app across a pytest session. It doesn't — NiceGUI's `core.app` has those routes disabled, so there's nothing to collide with, and I measured zero routes lost. Noting it so nobody re-walks it.

</details>

Thanks again for the fix — the diagnosis was the valuable part, and it's correct.

evnchn:

**The cleanup you were asked for is done, and CI is [green](https://github.com/zauberzeug/nicegui/actions/runs/32051061951)** — `pre-commit` and `pylint` included, the two that were red on your first push. Switching to `documentId` was a real strengthening too: the old `current_path` check would have passed on an SPA render.

Two things before it goes in, neither of them the code being wrong:

1. **A test for the query-string fix.** `clean_path` isn't defensive — it fixes a *second* real bug, and that one is unguarded.
2. **A description rewrite.** It still documents `get_routes()` and the three files it touched.

On `main`, a link to a registered page 404s as soon as it carries a query string. No `APIRouter` involved:

| link | `main` | this PR |
|---|---|---|
| `/other` | `OTHER PAGE` | `OTHER PAGE` |
| `/other?x=1` | **`404: sub page /other?x=1 not found`** | `OTHER PAGE` |

<details><summary><b>MRE</b> — nicegui <code>3.16.0.post5.dev0+ff3ea150</code>, fastapi 0.136.3, starlette 1.3.1, Python 3.12.11, macOS</summary>

```python
from nicegui import ui


@ui.page('/other')
def other_page():
    ui.label('OTHER PAGE')


@ui.page('/')
def index():
    ui.sub_pages({'/': lambda: ui.label('INDEX')})
    ui.link('without query', '/other')   # works
    ui.link('with query', '/other?x=1')  # 404s on main


ui.run(port=8321, show=False, reload=False)
```

Click each link. Run as a real process against both routers, headless Chrome driving the clicks and reading the first line of `<body>`:

```
router=upstream/main
>>> without query   -> 'OTHER PAGE'                          [url /other]
>>> with query      -> '404: sub page /other?x=1 not found'  [url /other?x=1]

router=PR6288 (ff3ea150)
>>> without query   -> 'OTHER PAGE'                          [url /other]
>>> with query      -> 'OTHER PAGE'                          [url /other?x=1]
```

`_handle_navigate` passes the raw path, so `other_route.matches({'path': '/other?x=1'})` never matched, `_other_page_builder_matches_path()` returned `False`, and `ui.sub_pages` kept ownership of a path it has no route for. Your `clean_path` strip is what fixes it.

</details>

<details><summary><b>Suggested test</b> — fails with just the <code>clean_path</code> line reverted</summary>

```python
def test_navigate_from_sub_pages_to_page_with_query_string(screen: Screen):
    @ui.page('/other')
    def other_page():
        ui.label('Other')

    @ui.page('/')
    def index():
        ui.sub_pages({'/': lambda: ui.label('Index')})
        ui.link('Go to other page', '/other?x=1')

    screen.open('/')
    screen.should_contain('Index')
    initial_doc_id = screen.selenium.execute_script('return window.documentId;')

    screen.click('Go to other page')
    screen.should_contain('Other')
    assert screen.selenium.execute_script('return window.documentId;') != initial_doc_id
```

Passes on this branch; with `clean_path = path.split('?', maxsplit=1)[0]` reverted to `clean_path = path` it fails with `AssertionError: Could not find "Other"`.

</details>

GitHub only allows real `suggestion` blocks on diff lines, so the description comes as copy-paste:

<details><summary><b>Suggested description</b> — written as if the test above is added; drop its bullet if you'd rather not add it</summary>

````markdown
Filed by  Gemini Flash 3.6  on Dinesh Thumma's behalf

### Motivation

Fixes #6281

When registering FastAPI routers via `app.include_router(router)`, `ui.sub_pages` treated the registered path as an internal SPA sub-page instead of triggering a full page load to the backend route.

The cause is a trailing-slash mismatch: `APIRouter(prefix='/other')` with `@router.page('/')` registers `/other/`, while a link navigates to `/other`. Starlette reports `Match.NONE` for `/other` and `Match.FULL` for `/other/`, so `_other_page_builder_matches_path()` concluded no other page builder owned the path and kept handling it in the SPA.

A second, independent case failed the same way: `_other_page_builder_matches_path()` passed the raw path to `Route.matches`, so `/other?x=1` matched no other page route and navigating to it rendered `404: sub page /other?x=1 not found` even though `/other` is registered.

### Implementation

Two independent changes in `_other_page_builder_matches_path()` (`nicegui/sub_pages_router.py`):

1. **Trailing-slash probe** — when a path does not match a route as given, retry with the trailing slash toggled, aligning with Starlette/FastAPI URL conventions. This fixes the reported bug.
2. **Query-string strip** — match against the path with any `?...` removed, so `/other?x=1` resolves as `/other`.

**Regression tests** in `tests/test_sub_pages.py`, one per fix. Each navigates away from `ui.sub_pages` and asserts `window.documentId` changes — i.e. a real full page load, not an SPA render:

- `test_navigate_from_sub_pages_to_api_router_page` — an `APIRouter` route registered via `app.include_router()`.
- `test_navigate_from_sub_pages_to_page_with_query_string` — a plain `@ui.page` reached with a query string.

### Progress

````

</details>

Unrelated, but worth doing: enable Actions on your fork at https://github.com/DineshThumma9/nicegui/actions (it shows no runs yet) — then your own pushes get CI without waiting on a maintainer to approve each run.

<details><summary><b>What else I ran</b> — <code>ff3ea150</code>, macOS, Python 3.12.11</summary>

| check | result |
|---|---|
| `pytest tests/test_sub_pages.py` | **55 passed** |
| `uv run pylint ./nicegui` (CI's invocation) | **10.00/10** |
| `uv run pre-commit run --files` <sub>the two changed files</sub> | **all passed** |

All four pylint findings that failed CI on `614ce6f1` were in the `get_routes()` code you dropped:

```
nicegui/sub_pages_router.py:8:0   W0611 Unused Route imported from starlette.routing
nicegui/testing/general.py:6:0   W0611 Unused Route imported from starlette.routing
nicegui/helpers/functions.py:95:0 C0305 Trailing newlines
nicegui/helpers/functions.py:81:4 C0415 Import outside toplevel
```

The `APIRouter` test guards its own fix — `alt_path` probe deleted, nothing else touched:

```
test_navigate_from_sub_pages_to_api_router_page FAILED
E  AssertionError: Could not find "Other"
```

`clean_path` had no equivalent guard, which is what sent me looking, and the MRE above is the result.

</details>

______________________________________________________________________

<sub><b>Changelog</b> — 2026-08-18: added the MRE showing the query-string 404 is a real second bug, plus a suggested test; the original wording called it "an observation, not a request", which undersold it. · 2026-08-18: rewrote the opening — it led with a straight ack while the body was asking for a test, and referred to a note it had already replaced.</sub>

**Review by evnchn — DISMISSED** [review]

**Both asks landed, and the tests are exactly the right shape — from me this is done.** The diff is now seven lines in `sub_pages_router.py` plus two tests, one per fix, and each test fails only when its own fix is reverted.

That last property is the one worth naming, because a test that only ever passes proves nothing. Reverting one fix at a time:

| reverted | `..._api_router_page` | `..._page_with_query_string` |
|---|---|---|
| nothing | PASSED | PASSED |
| `clean_path` strip | PASSED | **FAILED** |
| `alt_path` probe | **FAILED** | PASSED |

<details><summary><b>What I ran</b> — <code>17e893e7</code>, macOS, Python 3.12.11, fastapi 0.136.3, starlette 1.3.1</summary>

| check | result |
|---|---|
| `pytest tests/test_sub_pages.py -rs` | **56 passed**, 0 skipped <sub>(55 before this PR)</sub> |
| `pylint ./nicegui` | **10.00/10** |
| `mypy ./nicegui` | **Success: no issues found in 264 source files** |
| `pre-commit run --files` <sub>the two changed files</sub> | **all passed** |

The two revert runs, verbatim:

```
# clean_path = path.split('?', maxsplit=1)[0]  ->  clean_path = path
FAILED tests/test_sub_pages.py::test_navigate_from_sub_pages_to_page_with_query_string
E   AssertionError: Could not find "Other"
1 failed, 1 passed

# the four alt_path lines deleted
FAILED tests/test_sub_pages.py::test_navigate_from_sub_pages_to_api_router_page
E   AssertionError: Could not find "Other"
1 failed, 1 passed
```

</details>

I also approved the workflow run, and it is [green](https://github.com/zauberzeug/nicegui/actions/runs/32159366098) — `quick-test`, `pylint`, `mypy`, `pre-commit`, and the Windows storage job all pass.

Nothing further from me. Thanks for taking the cleanup and the extra test — the second bug was genuinely a separate one, and it's now covered.

falkoschindler:
Thanks @DineshThumma9 for the thorough diagnosis — the PR description nails both causes, and both tests genuinely fail without the fix.

While reviewing we noticed that the trailing-slash probe tries to predict Starlette's `redirect_slashes` behavior, and each way the emulation diverges from the real router becomes a user-visible edge case (details below). Rather than asking you to harden the probe, I took the liberty of pushing a commit that fixes the root cause instead: `page` now registers a path of `'/'` on a prefixed `APIRouter` as the bare prefix, so `@router.page('/')` with `prefix='/other'` creates the route `/other` — and the existing `Match.FULL` check just works. This also makes `Client.page_routes`, `ui.link(callable)` hrefs, and favicon routes agree on one canonical path (the favicon route even loses a double slash), and requests to `/other/` keep working via Starlette's own redirect.

Your query-string strip stays, but as `urlparse(path).path` — the idiom used elsewhere in the sub-pages code — which also covers fragments: `ui.link('Go', '/other#section')` had the identical bug one character away. The tests now assert the final URL (like the sibling test above them) instead of `window.documentId`, so the query-string test additionally pins that `?x=1` survives the navigation, and a parametrized case covers the fragment variant.

<details>
<summary>Why the probe would have been fragile</summary>

Returning `True` triggers `client.open(path)` with the *original* path, so the browser only reaches the slash-toggled route if Starlette actually issues its `redirect_slashes` 307. The probe diverged from that mechanism in several ways:

- Starlette only redirects when *no* route matches the original path. In the canonical wildcard layout (`@ui.page('/')` + `@ui.page('/{_:path}')` on the same builder) plus an `APIRouter(prefix='/other')` page, the wildcard is skipped by the same-builder check in the loop but still wins the real request for `/other`: a full reload that re-serves the same page and lands on the same 404, with all client state lost. Likewise, a partial match (e.g. `@app.post('/other')`) would end in a hard 405 page.
- `redirect_slashes` was never consulted; with `FastAPI(redirect_slashes=False)` the probe would promise a full load to a URL the server 404s.
- With a `path_prefix` deployment, a link equal to the prefix arrives as `''`, which passed the `!= '/'` guard, so `alt_path = '/'` matched any root page and ended in a broken `client.open('')`.
- Starlette's redirect uses `rstrip('/')`, the probe toggled a single slash, so they disagreed on paths like `/other//`.

Fixing the registration side removes this whole class instead of requiring the probe to mirror `starlette.routing.Router.app` exactly.

</details>

One deliberate behavior change worth noting: the canonical URL of such pages becomes `/other` instead of `/other/` (old URLs are 307-redirected), and relative links on those pages resolve against the parent now — worth a line in the release notes.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Root-cause%20analysis%20by%20a%20multi-agent%20review%2C%20implementation%20and%20wording%20by%20the%20model%2C%20decisions%20by%20the%20maintainer.&a=claude-code "claude-fable-5: Root-cause analysis by a multi-agent review, implementation and wording by the model, decisions by the maintainer.")

falkoschindler (approved):
Thanks again @DineshThumma9 — the diagnosis in this PR was spot-on, and it pointed us at a root cause worth fixing properly. With the registration-side fix, the hardened path parsing, and the fragment-scroll fix on top, this now covers the reported bug along with its query-string, fragment, and trailing-slash siblings, and all sub-pages tests run green. LGTM. 🚀

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Wording%20by%20the%20model%2C%20verdict%20by%20the%20maintainer.&a=claude-code "claude-fable-5: Wording by the model, verdict by the maintainer.")

PR #6289, opened 2026-08-15
Fix connection lost popup hanging on expired session by auto-reload

### LINKED ISSUE #6170: Connection lost, trying to reconnect" popup stays forever after the Microsoft Entra ID session expires
### Description

NiceGUI 3.14.0 on Azure App Service behind App Service Authentication (Easy Auth) with Microsoft Entra ID, `reconnect_timeout=30`.

When the Entra ID session expires, the "Connection lost, trying to reconnect" popup appears and never clears. It stays up indefinitely. Reloading the page manually sends me to the auth page, I sign in, and everything works again, so a reload is all that is needed to recover.

Is there a supported way to reload once the reconnect window expires?

### NiceGUI Version

3.14.0

### Python Version

3.12

### Browser

Firefox

### Operating System

Linux

DineshThumma9:
Filed by Google Gemini 3.6 Flash on DineshThumma's behalf

### Motivation

Fixes #6170

When an authentication session (such as Microsoft Entra ID / EasyAuth or OAuth) expires, proxy/auth middleware intercepts all Socket.IO reconnection requests with an HTTP 302 redirect to the login page. From the browser's perspective, this is treated as a generic network/transport error and Socket.IO retries indefinitely. As a result, the connection lost popup (`#popup`) hangs on screen forever and the page never reloads, preventing the user from being redirected to the login page.

### Implementation

1. **Client Auto-Reload on Reconnect Failure**:
   Updated `nicegui/static/nicegui.js` to monitor Socket.IO reconnection attempts (`reconnect_attempt`) and failure (`reconnect_failed`). After 5 failed reconnect attempts or on `reconnect_failed`, `nicegui.js` triggers `window.location.reload()`. This initiates a full browser HTTP GET request, which properly follows authentication redirects and allows the user to re-authenticate.

2. **Regression Tests**:
   Added `test_reconnect_reloads_after_max_attempts` in `tests/test_reconnect.py` asserting that reaching 5 failed reconnect attempts triggers a page reload.

### Progress

### DIFF
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -426,9 +426,17 @@ function createApp(elements, options) {
         tryAllTransports: true, // if the WebSocket fails with an error, try polling within the same attempt (see #5802)
       });
       let usePollingFallback = transportNames[0] === "websocket" && transportNames.includes("polling");
-      window.socket.io.on("reconnect_attempt", () => {
+      window.socket.io.on("reconnect_attempt", (attempt) => {
         // keep the otherwise-frozen reconnect query in sync to avoid triggering a reload (see #6019)
         options.query.next_message_id = window.nextMessageId;
+        if (attempt >= 5) {
+          console.log("reloading because maximum reconnect attempts reached");
+          window.location.reload();
+        }
+      });
+      window.socket.io.on("reconnect_failed", () => {
+        console.log("reloading because reconnect failed");
+        window.location.reload();
       });
       window.did_handshake = false;
       const messageHandlers = {
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -426,17 +426,9 @@ function createApp(elements, options) {
         tryAllTransports: true, // if the WebSocket fails with an error, try polling within the same attempt (see #5802)
       });
       let usePollingFallback = transportNames[0] === "websocket" && transportNames.includes("polling");
-      window.socket.io.on("reconnect_attempt", (attempt) => {
+      window.socket.io.on("reconnect_attempt", () => {
         // keep the otherwise-frozen reconnect query in sync to avoid triggering a reload (see #6019)
         options.query.next_message_id = window.nextMessageId;
-        if (attempt >= 5) {
-          console.log("reloading because maximum reconnect attempts reached");
-          window.location.reload();
-        }
-      });
-      window.socket.io.on("reconnect_failed", () => {
-        console.log("reloading because reconnect failed");
-        window.location.reload();
       });
       window.did_handshake = false;
       const messageHandlers = {
@@ -491,6 +483,15 @@ function createApp(elements, options) {
           } else if (err.message == "Implicit handshake failed") {
             console.log("reloading because implicit handshake failed for clientId " + window.clientId);
             window.location.reload();
+          } else if (err.context?.status && err.context.status !== 200) {
+            // an auth proxy intercepted the connection; probe before reloading to avoid reload-into-outage (see #6289)
+            fetch(window.location.href, { headers: { "NiceGUI-Check": "connect_error" } })
+              .then(() => {
+                if (window.socket.connected) return; // recovered while probing
+                console.log("reloading because server rejected connection with HTTP " + err.context.status);
+                window.location.reload();
+              })
+              .catch(() => {});
           }
         },
         try_reconnect: async () => {
--- a/nicegui/static/nicegui.js
+++ b/nicegui/static/nicegui.js
@@ -483,12 +483,14 @@ function createApp(elements, options) {
           } else if (err.message == "Implicit handshake failed") {
             console.log("reloading because implicit handshake failed for clientId " + window.clientId);
             window.location.reload();
-          } else if (err.context?.status && err.context.status !== 200) {
+          } else if (err.code === 'parser error' || err.context?.status >= 400) {
             // an auth proxy intercepted the connection; probe before reloading to avoid reload-into-outage (see #6289)
-            fetch(window.location.href, { headers: { "NiceGUI-Check": "connect_error" } })
-              .then(() => {
+            fetch(window.location.href, { headers: { 'NiceGUI-Check': 'connect_error' } })
+              .then((response) => {
                 if (window.socket.connected) return; // recovered while probing
-                console.log("reloading because server rejected connection with HTTP " + err.context.status);
+                if (!response.redirected && ![401, 403].includes(response.status))
+                  return; // not an auth wall
+                console.log('reloading because the connection was rejected: ' + err.message);
                 window.location.reload();
               })
               .catch(() => {});
--- a/tests/test_reconnect.py
+++ b/tests/test_reconnect.py
@@ -55,17 +55,19 @@ def test_reconnect_attempt_refreshes_query_next_message_id(screen: Screen):
     assert screen.selenium.execute_script('return Number(window.socket.io.opts.query.next_message_id);') > 0
 
 
-def test_reconnect_reloads_after_max_attempts(screen: Screen):
-    @ui.page('/')
+def test_short_outage_does_not_reload(screen: Screen):
+    """A brief connectivity drop must not cause a page reload (regression for #6289)."""
+    @ui.page('/', reconnect_timeout=3.0)
     def page():
-        ui.label('Initial Page Load')
+        ui.input('Input').props('autofocus')
 
     screen.open('/')
-    screen.should_contain('Initial Page Load')
+    screen.type('hello')
     initial_doc_id = screen.selenium.execute_script('return window.documentId;')
 
-    screen.selenium.execute_script('window.socket.io.emit("reconnect_attempt", 5);')
-    screen.wait(1.0)
-    new_doc_id = screen.selenium.execute_script('return window.documentId;')
-    assert new_doc_id != initial_doc_id
+    screen.selenium.execute_script('window.socket.io.engine.transport.onClose("transport close");')
+    screen.wait(1.5)  # well inside the 3 s reconnect_timeout; avoids race with server-side client pruning
 
+    assert screen.selenium.execute_script('return window.documentId;') == initial_doc_id
+    element = screen.selenium.find_element(By.XPATH, '//*[@aria-label="Input"]')
+    assert element.get_attribute('value') == 'hello', 'input should be preserved after short outage (no reload)'
--- a/tests/test_reconnect.py
+++ b/tests/test_reconnect.py
@@ -53,3 +53,19 @@ def test_reconnect_attempt_refreshes_query_next_message_id(screen: Screen):
     screen.selenium.execute_script('window.socket.io.engine.transport.onClose("transport close");')
     screen.wait(2.0)
     assert screen.selenium.execute_script('return Number(window.socket.io.opts.query.next_message_id);') > 0
+
+
+def test_reconnect_reloads_after_max_attempts(screen: Screen):
+    @ui.page('/')
+    def page():
+        ui.label('Initial Page Load')
+
+    screen.open('/')
+    screen.should_contain('Initial Page Load')
+    initial_doc_id = screen.selenium.execute_script('return window.documentId;')
+
+    screen.selenium.execute_script('window.socket.io.emit("reconnect_attempt", 5);')
+    screen.wait(1.0)
+    new_doc_id = screen.selenium.execute_script('return window.documentId;')
+    assert new_doc_id != initial_doc_id
+
--- a/tests/test_reconnect.py
+++ b/tests/test_reconnect.py
@@ -5,6 +5,7 @@ from nicegui.testing import Screen
 
 
 def test_reconnecting_without_page_reload(screen: Screen):
+    # A connectivity drop must not cause a page reload — regression guard for #6289.
     @ui.page('/', reconnect_timeout=3.0)
     def page():
         ui.input('Input').props('autofocus')
@@ -12,8 +13,12 @@ def test_reconnecting_without_page_reload(screen: Screen):
 
     screen.open('/')
     screen.type('hello')
+    initial_doc_id = screen.selenium.execute_script('return window.documentId;')
+
     screen.click('drop connection')
     screen.wait(2.0)
+
+    assert screen.selenium.execute_script('return window.documentId;') == initial_doc_id
     element = screen.selenium.find_element(By.XPATH, '//*[@aria-label="Input"]')
     assert element.get_attribute('value') == 'hello', 'input should be preserved after reconnect (i.e. no page reload)'
 
@@ -53,21 +58,3 @@ def test_reconnect_attempt_refreshes_query_next_message_id(screen: Screen):
     screen.selenium.execute_script('window.socket.io.engine.transport.onClose("transport close");')
     screen.wait(2.0)
     assert screen.selenium.execute_script('return Number(window.socket.io.opts.query.next_message_id);') > 0
-
-
-def test_short_outage_does_not_reload(screen: Screen):
-    """A brief connectivity drop must not cause a page reload (regression for #6289)."""
-    @ui.page('/', reconnect_timeout=3.0)
-    def page():
-        ui.input('Input').props('autofocus')
-
-    screen.open('/')
-    screen.type('hello')
-    initial_doc_id = screen.selenium.execute_script('return window.documentId;')
-
-    screen.selenium.execute_script('window.socket.io.engine.transport.onClose("transport close");')
-    screen.wait(1.5)  # well inside the 3 s reconnect_timeout; avoids race with server-side client pruning
-
-    assert screen.selenium.execute_script('return window.documentId;') == initial_doc_id
-    element = screen.selenium.find_element(By.XPATH, '//*[@aria-label="Input"]')
-    assert element.get_attribute('value') == 'hello', 'input should be preserved after short outage (no reload)'


tests/test_reconnect.py
```diff
@@ -53,3 +53,19 @@ def page():
...
+
+    screen.open('/')
+    screen.should_contain('Initial Page Load')
+    initial_doc_id = screen.selenium.execute_script('return window.documentId;')
+
+    screen.selenium.execute_script('window.socket.io.emit("reconnect_attempt", 5);')
+    screen.wait(1.0)
+    new_doc_id = screen.selenium.execute_script('return window.documentId;')
+    assert new_doc_id != initial_doc_id
+
```
evnchn: Trailing blank line at end of file — `autopep8` strips it, and the pre-commit job fails when a hook modifies a file. This is the only lint failure on the branch.

```suggestion
```

evnchn (commented):

**Right problem, wrong trigger.**

`reconnect_attempt` fires whether or not anything answered, so this reloads during ordinary outages too — and a reload with nothing listening lands on a browser error page with no JS left to retry. Measured on this branch: kill the server, and at t=18 s the tab reloads into `ERR_CONNECTION_REFUSED` and stays dead, including after the server comes back. On main the same outage keeps the popup up, keeps retrying, and self-heals ~9 s after the server returns.

The existing `connect_error` handler already reloads for `"timeout"` and for `"Implicit handshake failed"`, and it is safe precisely because it only runs when a connection attempt reached a server and was rejected. An expired Entra session looks like a third rejection shape that neither branch matches — handling it there, or probing with a `fetch` before committing to the reload, would fix the reported case without giving up outage recovery.

<details><summary><b>Measurements, and three smaller things</b></summary>

**Outage trace.** Identical script against both branches: open the page, kill the server, restart it at t=24 s.

This branch (`9917436`):

```
  t=  3s attempts=[1, 2]       same_doc=True   body='Hello Page\nConnection lost.\nTr'
  t= 12s attempts=[1, 2, 3, 4] same_doc=True   body='Hello Page\nConnection lost.\nTr'
  t= 18s attempts=[]           same_doc=False  body='This site can't be reached\n\nlo'
>>> restarting server
  t= 27s attempts=[]           same_doc=False  body='This site can't be reached\n\nlo'
  t= 48s attempts=[]           same_doc=False  body='This site can't be reached\n\nlo'
final: "This site can't be reached ... ERR_CONNECTION_REFUSED"
```

main (`8fb479bc`):

```
  t= 18s attempts=[1, 2, 3, 4, 5]    same_doc=True   body='Hello Page\nConnection lost.\nTr'
  t= 24s attempts=[1, 2, 3, 4, 5, 6] same_doc=True   body='Hello Page\nConnection lost.\nTr'
>>> restarting server
  t= 33s attempts=[]                 same_doc=False  body='Hello Page'
final: 'Hello Page'
console: "reloading because implicit handshake failed for clientId de94b46e-..."
```

Driven with Selenium + a real `ui.run()` server killed and restarted out of band. This proves the connection-refused case; it does not simulate the reporter's auth proxy, where the server *does* answer and a blind reload would happen to work.

**`reconnect_failed` cannot fire.** `socket.io` only emits it once `reconnectionAttempts` is exceeded; the bundled build defaults it to infinity and NiceGUI never sets it, so that handler is dead code:

```
$ grep -o "reconnectionAttempts[^,;)]*" nicegui/static/socket.io.min.js
reconnectionAttempts(i.reconnectionAttempts||1/0        # 1/0 === Infinity

$ grep -rn "reconnectionAttempts\|reconnection:" nicegui/static/nicegui.js nicegui/templates/
(no matches)
```

**The new test doesn't reach the code path.** `window.socket.io.emit("reconnect_attempt", 5)` calls the Manager's own `Emitter.emit`, hand-firing the listener against a healthy server. It would still pass if socket.io renamed the event, changed the argument shape, or started counting from zero. The behaviour worth pinning is the negative one — that a short outage still recovers without a reload.

**Two nits.** Nothing sets `skipReconnect`, so while the page unloads, attempts 6, 7, … each log and call `reload()` again. And the threshold `5` has no relation to the server-side `reconnect_timeout` — the reporter runs `reconnect_timeout=30`, but the client would give up after ~18 s regardless.

</details>

tests/test_reconnect.py
evnchn: Additional note: please try and lint stuff per [CONTRIBUTING.MD](https://github.com/zauberzeug/nicegui/blob/main/CONTRIBUTING.md) which the [agent does read if configured properly because there is a link](https://github.com/zauberzeug/nicegui/blob/main/AGENTS.md)

evnchn (commented):

**Outage regression is gone — but the reported case still hangs.** Measured on e05bb8e: behind a same-origin 302-to-login proxy the popup is still up at the end of a 36 s window, because that `connect_error` carries no `context` at all. `err.context.status` only exists when the proxy rejects with a real status code.

<details><summary><b>Measurements, cause, and the runnable harness</b></summary>

Selenium against a real `ui.run()` server behind an ASGI middleware simulating an expired EasyAuth session. Live socket dropped after arming, so every *reconnect* meets the proxy.

| branch | fault | reloaded? | `connect_error` the handler sees | outcome |
|---|---|---|---|---|
| e05bb8e | 302 → same-origin `/login` (200 HTML) | no | `message="server error"`, `code="parser error"`, **no context** | popup stuck ≥36 s |
| main 9b5133c | same | no | identical | popup stuck ≥30 s — **unchanged** |
| e05bb8e | proxy answers 401 directly | yes, 3 s | `context.status = 401` | lands on the proxy's response ✔ |
| e05bb8e | plain outage, server back at t≈20 s | no reload while dead | `xhr poll error`, `context.status = 0` | self-heals at t=24 s ✔ |

The first two rows are the point: with no `context`, the new status branch isn't reachable in this case, so the PR behaves exactly like main. The outage row is your fix working — the old `attempt >= 5` reload parked that tab on `ERR_CONNECTION_REFUSED` for good.

Cause: XHR follows the 302 transparently, so the login page arrives as **200 HTML**; engine.io then parses it, fails, and raises `server error` / `parser error` with no XHR attached. A status survives only when the proxy rejects outright (the 401 row), and `0` is deliberately falsy (the outage row).

Your `fetch` probe is the right primitive — it just needs a trigger that this case reaches (`err.code === 'parser error'`) and a discriminator it can read (`response.redirected`, or a content-type that isn't a socket.io handshake).

Trace, row 1:

```
  t=  0s baseline connected=true popup_hidden=true body='Input | Hello Page'
>>> auth proxy ARMED: every path except /login answers 302; live socket dropped
  t=  3s reloaded=False connected=False popup_hidden=false body='Input | Hello Page | Connection lost. | Trying'
  ...
  t= 36s reloaded=False connected=False popup_hidden=false body='Input | Hello Page | Connection lost. | Trying'
  connect_error: [{"code":"parser error","ctx":null,"message":"server error","status":"no-context"}, ...]
  console: WebSocket connection to 'ws://…&transport=websocket' failed: Unexpected response code: 403
```

Harness — put both files in a checkout and run from there: `python3 measure.py . label 302|401|outage`

```python
# authproxy_app.py
import os, sys
from pathlib import Path
from fastapi.responses import HTMLResponse
from nicegui import app, ui

MODE_FILE = Path(os.environ.get('MODE_FILE', '/tmp/nicegui_authproxy_mode'))


class AuthProxy:
    def __init__(self, app):
        self.app = app

    async def __call__(self, scope, receive, send):
        mode = MODE_FILE.read_text().strip() if MODE_FILE.exists() else ''
        path = scope.get('path', '')
        if mode and not path.startswith('/login'):
            if scope['type'] == 'http':
                if mode == '302':
                    await send({'type': 'http.response.start', 'status': 302,
                                'headers': [(b'location', b'/login?rd=' + path.encode())]})
                    await send({'type': 'http.response.body', 'body': b''})
                else:
                    await send({'type': 'http.response.start', 'status': int(mode),
                                'headers': [(b'content-type', b'text/plain')]})
                    await send({'type': 'http.response.body', 'body': b'unauthorized'})
                return
            if scope['type'] == 'websocket':
                await send({'type': 'websocket.close', 'code': 1008})
                return
        await self.app(scope, receive, send)


app.add_middleware(AuthProxy)


@app.get('/login')
def login():
    return HTMLResponse(b'<!doctype html><html><body><h1>Sign in</h1></body></html>')


@ui.page('/')
def index():
    ui.input('Input').props('autofocus')
    ui.label('Hello Page')


if __name__ == '__main__':
    ui.run(port=int(sys.argv[1]), show=False, reload=False)
```

```python
# measure.py  ->  measure.py <repo-dir> <label> <302|401|outage>
import json, os, signal, subprocess, sys, time
from pathlib import Path
from selenium import webdriver
from selenium.webdriver.chrome.options import Options

REPO, LABEL, CASE = sys.argv[1], sys.argv[2], sys.argv[3]
PORT = int(os.environ.get('PORT', '8081'))
MODE_FILE = Path(f'/tmp/nicegui_authproxy_mode_{PORT}')
DURATION = float(os.environ.get('DURATION', '39'))

RECORDER = """
window.__mark = 'ALIVE';
window.__errs = [];
window.socket.on('connect_error', (e) => window.__errs.push({
  message: String(e && e.message), code: e && e.code,
  ctx: e && e.context ? (e.context.constructor ? e.context.constructor.name : typeof e.context) : null,
  status: e && e.context ? e.context.status : 'no-context',
}));
"""
PROBE = """
return {mark: window.__mark || null, connected: !!(window.socket && window.socket.connected),
        popup: (document.getElementById('popup') || {}).ariaHidden, errs: window.__errs || null,
        url: window.location.pathname, body: (document.body.innerText || '').slice(0, 40).replace(/\\n/g, ' | ')};
"""

MODE_FILE.unlink(missing_ok=True)
env = {**os.environ, 'MODE_FILE': str(MODE_FILE)}
server = subprocess.Popen([sys.executable, 'authproxy_app.py', str(PORT)], cwd=REPO, env=env,
                          stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
opts = Options()
opts.add_argument('--headless=new')
opts.set_capability('goog:loggingPrefs', {'browser': 'ALL'})
driver = None
try:
    time.sleep(4)
    driver = webdriver.Chrome(options=opts)
    driver.get(f'http://localhost:{PORT}/')
    time.sleep(3)
    driver.execute_script(RECORDER)
    print(f'### {LABEL} / case={CASE}')
    print(f'  t=  0s baseline {json.dumps(driver.execute_script(PROBE))}')
    if CASE == 'outage':
        server.send_signal(signal.SIGKILL)
        print('>>> server KILLED')
    else:
        MODE_FILE.write_text(CASE)
        driver.execute_script('window.socket.io.engine.close();')
        print(f'>>> auth proxy ARMED ({CASE}); live socket dropped')
    start, last_errs, restarted = time.time(), [], False
    while time.time() - start < DURATION:
        time.sleep(3)
        if CASE == 'outage' and not restarted and time.time() - start > 20:
            server = subprocess.Popen([sys.executable, 'authproxy_app.py', str(PORT)], cwd=REPO, env=env,
                                      stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
            restarted = True
            print('>>> server RESTARTED')
        s = driver.execute_script(PROBE)
        last_errs = s['errs'] or last_errs
        print(f'  t={time.time()-start:4.0f}s reloaded={s["mark"] is None!s:5} connected={s["connected"]!s:5} '
              f'popup_hidden={s["popup"]} url={s["url"]} body={s["body"]!r}')
    print(f'  connect_error: {json.dumps(last_errs[-2:])}')
    print(f'  console: {json.dumps([e["message"] for e in driver.get_log("browser")][-2:])}')
finally:
    if driver:
        driver.quit()
    MODE_FILE.unlink(missing_ok=True)
    if server.poll() is None:
        server.kill()
    server.wait()
```

What this proves and what it doesn't: it reproduces the relevant shape — polling gets a same-origin 302 to an HTML login page — on the leg that could carry a readable status. My proxy *closes* the WebSocket handshake rather than 302-ing it (uvicorn surfaces that as 403), but a browser hides both from JS either way. A cross-origin IdP is untested here; per browser CORS behaviour it would surface as a network-style failure (`status = 0`), which is also falsy — but I have not measured that.

Test nit: `test_short_outage_does_not_reload` largely repeats `test_reconnecting_without_page_reload` directly above it — same forced drop, same `'hello'`-survives assertion — and nothing exercises the new branch.

Also: lint is clean now, and the CI Gate run has been stuck *in progress* since 18:04 UTC yesterday, so the missing green isn't yours.

</details>

nicegui/static/nicegui.js
```diff
@@ -483,6 +483,15 @@ function createApp(elements, options) {
           } else if (err.message == "Implicit handshake failed") {
             console.log("reloading because implicit handshake failed for clientId " + window.clientId);
             window.location.reload();
+          } else if (err.context?.status && err.context.status !== 200) {
```
evnchn: *Posted by Claude Code on @evnchn's behalf.*

**Trigger.** The reported 302 case raises `{message: "server error", code: "parser error"}` with no `context` at all, so this condition never matches it — measured in the review comment above.

```suggestion
          } else if (err.code === "parser error" || err.context?.status >= 400) {
```

nicegui/static/nicegui.js
```diff
@@ -483,6 +483,15 @@ function createApp(elements, options) {
...
+            // an auth proxy intercepted the connection; probe before reloading to avoid reload-into-outage (see #6289)
+            fetch(window.location.href, { headers: { "NiceGUI-Check": "connect_error" } })
+              .then(() => {
+                if (window.socket.connected) return; // recovered while probing
+                console.log("reloading because server rejected connection with HTTP " + err.context.status);
```
evnchn: *Posted by Claude Code on @evnchn's behalf.*

**Decision.** Let the probe response decide, and test for an auth wall specifically. `!response.ok` is not enough: with the origin down behind nginx the probe resolves `502`, and the tab reloads onto the proxy's error page — the reload-into-an-outage failure the probe exists to prevent (measured: current branch reloads at t=3 s under a 502).

The log line changes too because `err.context.status` re-read inside `.then()` reads back `0` — that same 502 run printed `HTTP 0`.

```suggestion
              .then((response) => {
                if (window.socket.connected) return; // recovered while probing
                if (!response.redirected && ![401, 403].includes(response.status)) return; // not an auth wall
                console.log("reloading because the connection was rejected: " + err.message);
```

nicegui/static/nicegui.js
```diff
@@ -483,6 +483,17 @@ function createApp(elements, options) {
...
+            // an auth proxy intercepted the connection; probe before reloading to avoid reload-into-outage (see #6289)
+            fetch(window.location.href, { headers: { 'NiceGUI-Check': 'connect_error' } })
+              .then((response) => {
+                if (window.socket.connected) return; // recovered while probing
+                if (!response.redirected && ![401, 403].includes(response.status))
+                  return; // not an auth wall
+                console.log('reloading because the connection was rejected: ' + err.message);
+                window.location.reload();
+              })
+              .catch(() => {});
```
evnchn, line 496: *Posted by Claude Code on @evnchn's behalf.*

**This should fix the 302-to-cross-origin-IdP case** — the one my previous suggestion missed. That suggestion covered the shape I had measured (a same-origin login page that terminates the redirect), and it works there, but a real Entra / EasyAuth chain ends at `login.microsoftonline.com`. CORS blocks the poll on that leg, so socket.io reports `xhr poll error` with `status: 0` and `code: null` — byte-identical to a plain outage, and matching neither condition here. Measured on `8756c119`: the popup is still up after 24 s.

Nothing at the error level separates an auth wall from an outage, so the trigger has to go and the probe has to be the sole discriminator. `redirect: "manual"` is what makes that safe: any 302 becomes an `opaqueredirect` response without being followed, so a cross-origin IdP never gets the chance to fail the probe on CORS.

```suggestion
          } else {
            // the error is indistinguishable from an outage, so probe the page before reloading (see #6289)
            fetch(window.location.href, { headers: { "NiceGUI-Check": "connect_error" }, redirect: "manual" })
              .then((response) => {
                if (window.socket.connected) return; // recovered while probing
                if (response.type !== "opaqueredirect" && ![401, 403].includes(response.status)) return; // not an auth wall
                console.log("reloading because the connection was rejected: " + err.message);
                window.location.reload();
              })
              .catch(() => {}); // the origin is unreachable, so keep retrying
```

Measured on exactly that text, five faults, Selenium against a real `ui.run()` behind an ASGI auth proxy:

| fault | result |
|---|---|
| 302 to cross-origin IdP | reloads at 3 s, lands on the IdP |
| 302 to same-origin login | reloads at 3 s, lands on `/login` |
| proxy answers 401 directly | reloads at 3 s |
| origin down behind the proxy (502) | no reload, keeps retrying |
| plain outage, server back at t≈20 s | no reload, self-heals at 33 s |

`tests/test_reconnect.py` still passes.

<details><summary><b>A regression test for the new branch</b></summary>

Nothing currently exercises the reload path — the `documentId` assertion re-guards what the surviving `'hello'` already guarded. This one fails on `main` and passes with the fix, in about 4 s:

```python
import threading

from fastapi.responses import HTMLResponse

from nicegui import app, ui


def test_reload_when_auth_proxy_redirects(screen: Screen):
    screen.allowed_js_errors.append('WebSocket connection to')
    armed = threading.Event()

    class AuthProxy:
        def __init__(self, app_):
            self.app = app_

        async def __call__(self, scope, receive, send):
            if armed.is_set() and not scope.get('path', '').startswith('/login'):
                if scope['type'] == 'http':
                    await send({'type': 'http.response.start', 'status': 302,
                                'headers': [(b'location', b'/login')]})
                    await send({'type': 'http.response.body', 'body': b''})
                    return
                if scope['type'] == 'websocket':
                    await send({'type': 'websocket.close', 'code': 1008})
                    return
            await self.app(scope, receive, send)

    app.add_middleware(AuthProxy)

    @app.get('/login')
    def login():
        return HTMLResponse('<h1>Sign in</h1>')

    @ui.page('/')
    def page():
        ui.label('Hello Page')

    screen.open('/')
    screen.should_contain('Hello Page')
    armed.set()
    screen.selenium.execute_script('window.socket.io.engine.close();')
    screen.should_contain('Sign in')
```

</details>

evnchn:
I think we're almost there with this PR 💪

PR #6294, opened 2026-08-16
Fix leakage of tasks awaiting initialization of some elements

### LINKED ISSUE #6277: Element-owned awaits leak their tasks when the client is deleted (button.clicked, scene/leaflet initialized)
### Description

#6250 fixed awaited dialogs leaking their tasks when the client is deleted, but the same class of leak remains in other element-owned awaits. Each of these waits on an `asyncio.Event` that nothing sets when the element or its client is deleted, so the awaiting task stays in `background_tasks.running_tasks` until server shutdown and keeps its coroutine frame (often the whole page object graph) alive:

- `Button.clicked()` — waits on a per-call local `asyncio.Event` wired to the click event ([button.py](https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/button.py))
- `Scene.initialized()` and `SceneView.initialized()` — wait on the element-owned `_initialized_event` ([scene.py](https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/scene/scene.py), [scene_view.py](https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/scene/scene_view.py))
- `Leaflet.initialized()` — same pattern ([leaflet.py](https://github.com/zauberzeug/nicegui/blob/main/nicegui/elements/leaflet/leaflet.py))

Minimal example for the button variant — click "Wait for click", then close the tab and wait past the reconnect timeout; the counter never goes down (same repro technique as in #6250):

```python
from nicegui import background_tasks, ui


@ui.page('/')
def page():
    button = ui.button('Yes')

    async def wait() -> None:
        await button.clicked()
        ui.notify('Yes!')

    ui.button('Wait for click', on_click=wait)
    ui.label().bind_text_from(background_tasks, 'running_tasks',
                              backward=lambda tasks: f'{sum(1 for t in tasks if "wait" in t.get_name())} pending "wait" handlers')


ui.run()
```

### Suggested fix

~~Apply the pattern from #6250: resolve the events in `_handle_delete` so waiting tasks resume …~~ **Update:** as discussed in #6271, resolving is wrong here — unlike `await dialog`, these waits have no "didn't happen" channel, so resuming them is indistinguishable from the awaited interaction actually occurring. Instead, the awaiting task is **cancelled** when the element is deleted. #6271 implements this for `Button.clicked()` via a `CancelableWaitElement` mixin; #6294 applies the same to the `initialized()` waits.

SaadZahem:
### Motivation

Follow up to #6271
Fixes #6277

### Implementation

`ui.scene`, `ui.scene_view` and `ui.leaflet` now inherit the `CancelableWaitElement` mixin from #6271 instead of `Element`, and each `initialized()` body is wrapped in `with self._cancel_when_deleted(self._initialized_event):`.

The mixin does both halves:
1. On deletion, it cancels every task awaiting this element (`_handle_delete`).
2. Its entry check raises `CancelledError` immediately when the element is *already* deleted — before `await self.client.connected()`, which would otherwise park forever on a deleted client.

So all three elements now behave exactly like `ui.button`'s `clicked()`.

### Progress

- [x] The PR title is a short phrase starting with the verb "Fix"

### DIFF
--- a/nicegui/elements/leaflet/leaflet.py
+++ b/nicegui/elements/leaflet/leaflet.py
@@ -51,6 +51,7 @@ class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, 
         self.layers: list[Layer] = []
         self.is_initialized = False
         self._initialized_event = asyncio.Event()
+        self._initialized_tasks: set[asyncio.Task] = set()
 
         # read-write public API
         self.center = center
@@ -102,8 +103,19 @@ class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, 
 
     async def initialized(self) -> None:
         """Wait until the map is initialized."""
-        await self.client.connected()
-        await self._initialized_event.wait()
+        if self.is_initialized:
+            return
+        task = asyncio.current_task()
+        assert task is not None
+        if self.is_deleted:
+            task.cancel()
+            await asyncio.sleep(0)
+        self._initialized_tasks.add(task)
+        try:
+            await self.client.connected()
+            await self._initialized_event.wait()
+        finally:
+            self._initialized_tasks.discard(task)
 
     def _handle_move_or_zoom_end(self, e: GenericEventArguments) -> None:
         self._send_update_on_value_change = False
@@ -171,6 +183,8 @@ class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, 
 
     def _handle_delete(self) -> None:
         binding.remove(self.layers)
+        for task in self._initialized_tasks:
+            task.cancel()
         super()._handle_delete()
 
     def _to_dict(self):
--- a/nicegui/elements/leaflet/leaflet.py
+++ b/nicegui/elements/leaflet/leaflet.py
@@ -7,12 +7,12 @@ from typing_extensions import Self
 from ... import binding
 from ...awaitable_response import AwaitableResponse, NullResponse
 from ...defaults import DEFAULT_PROP, resolve_defaults
-from ...element import Element
 from ...events import GenericEventArguments
+from ..mixins.cancelable_wait_element import CancelableWaitElement
 from .leaflet_layer import Layer
 
 
-class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, default_classes='nicegui-leaflet'):
+class Leaflet(CancelableWaitElement, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, default_classes='nicegui-leaflet'):
     # pylint: disable=import-outside-toplevel
     from .leaflet_layers import GenericLayer as generic_layer
     from .leaflet_layers import ImageOverlay as image_overlay
@@ -51,7 +51,6 @@ class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, 
         self.layers: list[Layer] = []
         self.is_initialized = False
         self._initialized_event = asyncio.Event()
-        self._initialized_tasks: set[asyncio.Task] = set()
 
         # read-write public API
         self.center = center
@@ -105,18 +104,11 @@ class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, 
         """Wait until the map is initialized.
 
         *Updated in version 3.17.0: Awaiting leaflet initialization cancels the calling task
-        when the leaflet element is deleted, e.g. because the client disconnected.*
+        when the leaflet is deleted, e.g. because the client disconnected.*
         """
-        task = cast(asyncio.Task, asyncio.current_task())
-        if self.is_deleted:
-            task.cancel()
-            await asyncio.sleep(0)
-        self._initialized_tasks.add(task)
-        try:
+        with self._cancel_when_deleted(self._initialized_event):
             await self.client.connected()
             await self._initialized_event.wait()
-        finally:
-            self._initialized_tasks.discard(task)
 
     def _handle_move_or_zoom_end(self, e: GenericEventArguments) -> None:
         self._send_update_on_value_change = False
@@ -184,8 +176,6 @@ class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, 
 
     def _handle_delete(self) -> None:
         binding.remove(self.layers)
-        for task in self._initialized_tasks:
-            task.cancel()
         super()._handle_delete()
 
     def _to_dict(self):
--- a/nicegui/elements/leaflet/leaflet.py
+++ b/nicegui/elements/leaflet/leaflet.py
@@ -103,8 +103,8 @@ class Leaflet(CancelableWaitElement, component='leaflet.js', esm={'nicegui-leafl
     async def initialized(self) -> None:
         """Wait until the map is initialized.
 
-        *Updated in version 3.17.0: Awaiting leaflet initialization cancels the calling task
-        when the leaflet is deleted, e.g. because the client disconnected.*
+        *Updated in version 3.17.0: Awaiting map initialization cancels the awaiting task
+        when the map is deleted, e.g. because the client disconnected.*
         """
         with self._cancel_when_deleted(self._initialized_event):
             await self.client.connected()
--- a/nicegui/elements/leaflet/leaflet.py
+++ b/nicegui/elements/leaflet/leaflet.py
@@ -102,11 +102,12 @@ class Leaflet(Element, component='leaflet.js', esm={'nicegui-leaflet': 'dist'}, 
             self.run_method('add_layer', layer.to_dict(), layer.id)
 
     async def initialized(self) -> None:
-        """Wait until the map is initialized."""
-        if self.is_initialized:
-            return
-        task = asyncio.current_task()
-        assert task is not None
+        """Wait until the map is initialized.
+
+        *Updated in version 3.17.0: Awaiting leaflet initialization cancels the calling task
+        when the leaflet element is deleted, e.g. because the client disconnected.*
+        """
+        task = cast(asyncio.Task, asyncio.current_task())
         if self.is_deleted:
             task.cancel()
             await asyncio.sleep(0)
--- a/nicegui/elements/scene/scene.py
+++ b/nicegui/elements/scene/scene.py
@@ -2,7 +2,7 @@ import asyncio
 from collections.abc import Callable, Iterator
 from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 from typing_extensions import Self
 
@@ -214,9 +214,12 @@ class Scene(Element, component='scene.js', esm={'nicegui-scene': 'dist'}, defaul
         return super().run_method(name, *args, timeout=timeout)
 
     async def initialized(self) -> None:
-        """Wait until the scene is initialized."""
-        task = asyncio.current_task()
-        assert task is not None
+        """Wait until the scene is initialized.
+
+        *Updated in version 3.17.0: Awaiting scene initialization cancels the calling task
+        when the scene element is deleted, e.g. because the client disconnected.*
+        """
+        task = cast(asyncio.Task, asyncio.current_task())
         if self.is_deleted:
             task.cancel()
             await asyncio.sleep(0)
--- a/nicegui/elements/scene/scene.py
+++ b/nicegui/elements/scene/scene.py
@@ -215,7 +215,7 @@ class Scene(CancelableWaitElement, component='scene.js', esm={'nicegui-scene': '
     async def initialized(self) -> None:
         """Wait until the scene is initialized.
 
-        *Updated in version 3.17.0: Awaiting scene initialization cancels the calling task
+        *Updated in version 3.17.0: Awaiting scene initialization cancels the awaiting task
         when the scene is deleted, e.g. because the client disconnected.*
         """
         with self._cancel_when_deleted(self._initialized_event):
--- a/nicegui/elements/scene/scene.py
+++ b/nicegui/elements/scene/scene.py
@@ -113,6 +113,7 @@ class Scene(Element, component='scene.js', esm={'nicegui-scene': 'dist'}, defaul
         self.stack: list[Object3D | SceneObject] = [SceneObject()]
         self._batched_calls: list[list[Any]] | None = None
         self._initialized_event = asyncio.Event()
+        self._initialized_tasks: set[asyncio.Task] = set()
         self._click_handlers = [on_click] if on_click else []
         self._props['click-events'] = click_events[:]
         self._drag_start_handlers = [on_drag_start] if on_drag_start else []
@@ -214,8 +215,17 @@ class Scene(Element, component='scene.js', esm={'nicegui-scene': 'dist'}, defaul
 
     async def initialized(self) -> None:
         """Wait until the scene is initialized."""
-        await self.client.connected()
-        await self._initialized_event.wait()
+        task = asyncio.current_task()
+        assert task is not None
+        if self.is_deleted:
+            task.cancel()
+            await asyncio.sleep(0)
+        self._initialized_tasks.add(task)
+        try:
+            await self.client.connected()
+            await self._initialized_event.wait()
+        finally:
+            self._initialized_tasks.discard(task)
 
     def _handle_click(self, e: GenericEventArguments) -> None:
         arguments = SceneClickEventArguments(
@@ -306,6 +316,8 @@ class Scene(Element, component='scene.js', esm={'nicegui-scene': 'dist'}, defaul
 
     def _handle_delete(self) -> None:
         binding.remove(list(self.objects.values()))
+        for task in self._initialized_tasks:
+            task.cancel()
         super()._handle_delete()
 
     def delete_objects(self, predicate: Callable[[Object3D], bool] = lambda _: True) -> None:
--- a/nicegui/elements/scene/scene.py
+++ b/nicegui/elements/scene/scene.py
@@ -2,14 +2,13 @@ import asyncio
 from collections.abc import Callable, Iterator
 from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Any, Literal, cast
+from typing import Any, Literal
 
 from typing_extensions import Self
 
 from ... import binding
 from ...awaitable_response import AwaitableResponse, NullResponse
 from ...defaults import DEFAULT_PROP, resolve_defaults
-from ...element import Element
 from ...events import (
     GenericEventArguments,
     Handler,
@@ -18,6 +17,7 @@ from ...events import (
     SceneDragEventArguments,
     handle_event,
 )
+from ..mixins.cancelable_wait_element import CancelableWaitElement
 from .scene_object3d import Object3D
 
 
@@ -41,7 +41,7 @@ class SceneObject:
     id: str = 'scene'
 
 
-class Scene(Element, component='scene.js', esm={'nicegui-scene': 'dist'}, default_classes='nicegui-scene'):
+class Scene(CancelableWaitElement, component='scene.js', esm={'nicegui-scene': 'dist'}, default_classes='nicegui-scene'):
     # pylint: disable=import-outside-toplevel
     from .objects.axes_helper import AxesHelper as axes_helper
     from .objects.box import Box as box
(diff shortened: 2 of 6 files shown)


evnchn:

**TL;DR: sound fix, and the tests genuinely guard** — all three fail with `nicegui/` reverted to main, so they are real regression tests rather than green-only ones. One thing I would change before merge, and one cheap docs ask.

### The new leaflet fast path makes `initialized()` answer differently per element

`leaflet.py` gained `if self.is_initialized: return` **above** the deleted check, while `scene.py` and `scene_view.py` got no equivalent:

```python
async def initialized(self) -> None:
    """Wait until the map is initialized."""
    if self.is_initialized:      # <-- new in this PR, leaflet only, and ahead of the guard
        return
    task = asyncio.current_task()
    assert task is not None
    if self.is_deleted:
        task.cancel()
```

So an element that initialized and *was then deleted* returns normally on leaflet and cancels on scene — the caller keeps running against a deleted element in exactly the case this PR exists to stop.

<details><summary><b>Measured on f0ddc709 — same page for each: initialize, delete, then await</b></summary>

```python
try:
    await el.initialized()
    results[name] = 'returned'
    results[name] = 'cancelled'
```
```
RESULT: {'leaflet': 'returned', 'scene': 'cancelled', 'scene_view': 'cancelled'}
```
</details>

Dropping the two lines is the symmetric fix: `is_initialized` is a public attribute only `Leaflet` has, so scene and scene_view cannot mirror it. It also costs little — `leaflet.js` emits `init` only once `window.socket.id` exists, so by the time `is_initialized` is true both `client.connected()` and `_initialized_event` are already satisfied and the two awaits resolve without suspending. If the early-out is worth keeping, it needs to sit *below* the `is_deleted` check.

### Worth a docstring line

The three docstrings still say "Wait until the scene is initialized" while the method may now cancel the caller's task. #6250 set the precedent for exactly this shape of change with an `*Updated in version 3.16.0:*` note, so the equivalent here would be a one-liner per docstring for 3.17.

<details><summary><b>Verification performed</b></summary>

- Fresh worktree at PR head f0ddc709; the three new tests pass, and **all three fail** with only `nicegui/` reverted to main (`assert task[0].cancelled()` → False, task left pending).
- `ruff`, `pylint` (10.00/10) and `mypy` clean on the changed files.
- Delete path emits no log noise (`caplog` at DEBUG shows only the httpx request line).
- Control never reaches `_initialized_tasks.add(task)` after the guard fires. That holds for *this* shape specifically — cancelling `current_task()`, which has no `_fut_waiter`, so `cancel()` sets `_must_cancel` and CPython raises at the next `Task.__step`. Cancellation is not generally immediate, but here it is.
- Iterating `_initialized_tasks` while cancelling is safe: `Task.cancel()` schedules the wakeup via `call_soon`, so the `finally: discard` cannot mutate the set mid-loop.
- Cancel-rather-than-resolve is the endorsed semantics here (per the ruling on #6271), so the difference from #6250's resolve-with-`None` dialog behaviour is not flagged.
</details>

<details><summary><b>Two follow-ups, not blockers for this PR</b></summary>

- **`SceneView` outliving its `Scene`.** `scene_view.js` resolves `getElement(this.sceneId).scene` in `mounted()`, and `Scene` holds no reference to its views — so deleting the scene while leaving the view alive looks like it strands `await view.initialized()` forever, which this PR would not catch. Worth a check; I have not reproduced it.
- **Four copies of the block.** scene, scene_view, leaflet, plus `Button._click_tasks` in #6271 all carry the same ~12 lines. The leaflet divergence above is that duplication drifting already, so a shared helper would be earning its keep — better as its own PR than as scope here.
</details>

<details><summary><b>Minor — the <code>assert</code> is load-bearing for mypy, not a runtime guard</b></summary>

`assert task is not None` is stripped under `python -O`, which would put `None` into `_initialized_tasks` and raise `AttributeError` from `_handle_delete`. Only reachable for a coroutine driven outside a task, so defensive rather than a live bug — mentioning it only because the assert is load-bearing for `mypy` and easy to mistake for a runtime guard.
</details>

evnchn:

**All three applied** — re-verified on 6f2dc3df.

<details><summary><b>What changed</b></summary>

- Fast path dropped, so the three elements agree again. Same probe as before now reports `{'leaflet': 'cancelled', 'scene': 'cancelled', 'scene_view': 'cancelled'}`; leaflet was `'returned'`.
- `*Updated in version 3.17.0:*` note on each of the three docstrings.
- `cast` rather than the `if` I suggested — better, since it leaves no runtime statement for `-O` to strip.

32 tests pass locally; CI green on pylint, mypy, pre-commit and quick-test.
</details>

evnchn:

Sorry — I found a bug.

```python
from nicegui import ui


@ui.page('/ok')
async def ok():
    scene = ui.scene()
    await scene.initialized()      # not deleted -> page completes
    ui.label('reached the end')


@ui.page('/broken')
async def broken():
    scene = ui.scene()
    scene.delete()                 # <-- the only difference
    await scene.initialized()
    ui.label('reached the end')


ui.run(port=8123, show=False, reload=False)
```

```
6f2dc3df (this PR)    ok -> HTTP 200    broken -> HTTP 500    RuntimeError: No response returned.
8fb479bc (base)       ok -> HTTP 200    broken -> HTTP 200
```

SaadZahem:
Good catch, it turns out that canceling the calling task works until it's the async page itself that gets canceled.
So what should be done instead? What is the correct behavior?

1. Leave as it is, still requiring documentation. The issue isn't common enough. Tell users to be more careful.
2. Raise a clearly worded custom exception in these 4 narrow edge-cases (`button.clicked` included).

evnchn:

`nicegui/page.py:207`

```diff
                 if task.done():
-                    result = task.result()
+                    result = None if task.cancelled() else task.result()
                 else:
                     result = None
```

<details><summary><b>This fix should work</b></summary>

**That line is the only unguarded `.result()` on a response path.** `page.py:128` (`check_for_late_return_value`) and `background_tasks.py:152` both already do `except asyncio.CancelledError: pass`. That asymmetry is exactly why deleting the element *during* the await already returns 200 — the response is built first, so the cancellation arrives on the guarded late path — while deleting it *before* the await gives a 500, because the cancel lands synchronously and reaches line 207.

**`None` is already the established "no result" value.** The `else` branch two lines down sets exactly that. Downstream, `result` only matters when it `isinstance(result, Response)`; otherwise the route falls through to `client.build_response(...)`, which renders from client and element state rather than from `result`.

**It cannot swallow a real error.** The builder task runs with `handle_exceptions=False` and converts ordinary exceptions into a 500 page itself, so `task.cancelled()` is true only in the case where `task.result()` would raise plain `CancelledError`. And the `response_timeout` branch above cancels the task then returns at line 203, so line 207 is unreachable for that cancellation — the new branch is reachable only via this PR's `initialized()` cancel.

**Measured**, same two-route file as above:

| | `/ok` | `/broken` | `/broken` body contains "reached the end" |
|---|---|---|---|
| base `8fb479bc` | 200 | 200 | no |
| head `6f2dc3df` | 200 | **500** | — |
| head + this line | 200 | 200 | no |

So it restores the pre-PR response behaviour exactly — same status, same partial page, no unhandled error in the log — while leaving the task cancelled, which is the semantics you want. The 32 scene / scene_view / leaflet tests still pass with it applied; `tests/test_page.py` has 4 `test_dark_mode` failures, but those are pre-existing on the PR head, unrelated to this line.

**Two caveats.** This cannot be a real applicable suggestion block, because `page.py` is not in this PR's diff. And the three new tests assert element-level `task.cancelled()`, not that the page still responds — so a route-level regression test would be the thing that actually guards this.
</details>

SaadZahem:
Thanks for pointing this out.
Should I add a commit to this PR and `button.clicked` PR, or open a new PR?
I lean towards the first option.

evnchn:

**A new PR, against #6296 — this turns out to be a main bug, not something either PR introduced.**

I filed #6296 so there is a fixed target to work against instead of reconstructing the scope from this thread. It carries the repro, the measurements, the proposed one-liner and the mechanism; treat it as the definition of done for that PR, including the regression-test note at the end. Nothing in it needs to be re-derived here.

<details><summary><b>Why it is a separate PR and not a commit in these two</b></summary>

The 500 reproduces on released 3.16.0 (`17ad9005`) and on main (`9b5133ca`) with neither of your PRs applied — the trigger there is an explicit self-cancel in a page builder, which is artificial, so the defect is latent on main rather than something users hit today.

What your two PRs change is that *library* code now reaches it: both add `task.cancel()` on the current task when the element is already deleted, so an ordinary page builder lands on `page.py:207`. Measured at `6f2dc3df` (`initialized()`) and `ca5243ce` (`clicked()`) — `/broken` 500 on both, and the same single line clears both.

So it fixes something neither PR broke, one PR unblocks both at once, and the route-level regression test belongs with it rather than split across `test_scene.py` and `test_button.py`.
</details>

Separately, in the `clicked()` PR the docstring says *Updated in version 3.16.0* — 3.16.0 is already tagged on main, so that wants to be 3.17.0 like the scene ones.

---

*Changelog*

*2026-08-18 — filed #6296 and repointed the recommendation at it as the definition of done; trimmed the evidence here to what the issue does not already carry.*

falkoschindler:
Since #6271 now extracts the mechanism into a `CancelableWaitElement` mixin, this PR can shrink to inheriting it: each element keeps its `_initialized_event` and wraps the wait in `with self._cancel_when_deleted(self._initialized_event):` — no per-element task registry, no `_handle_delete` override. That also picks up three fixes the copied pattern would otherwise inherit: the registry is initialized before element registration, tasks whose event is already set are not cancelled, and the early-out raises explicitly instead of relying on `await asyncio.sleep(0)`. I'd wait until #6271 is merged and then rebase this PR on it.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Multi-agent%20review%20and%20code%20changes%20by%20the%20model%3B%20semantics%20decision%20and%20direction%20by%20the%20maintainer.&a=claude-code "claude-fable-5: Multi-agent review and code changes by the model; semantics decision and direction by the maintainer.")

SaadZahem:
I agree with the general idea of having the mixin. I will checkout the branch to see what I can do.

Also I updated the branch using merge instead of rebase. Could that be an issue?

SaadZahem:
I mirrored the changes from #6271 into each element. That includes removing the word "element" from the docstring. I tested and checked the formatting before pushing. Ready for review.

evnchn (commented):

Three wording suggestions inline: they realign the version note with the wording merged in #6271 (*cancels the awaiting task*, not *the calling task*) and keep one noun per element in each docstring.

One a suggestion cannot reach: the **Implementation** section above still describes the pre-mixin approach ("Store a reference to pending task and cancel them in `scene._handle_delete()`"), which is no longer what the diff does and is the text a squash-merge carries forward.

nicegui/elements/scene/scene.py
```diff
@@ -213,9 +213,14 @@ def run_method(self, name: str, *args: Any, timeout: float = 1) -> AwaitableResp
         return super().run_method(name, *args, timeout=timeout)
 
     async def initialized(self) -> None:
-        """Wait until the scene is initialized."""
-        await self.client.connected()
-        await self._initialized_event.wait()
+        """Wait until the scene is initialized.
+
+        *Updated in version 3.17.0: Awaiting scene initialization cancels the calling task
```
evnchn: #6271 says *the awaiting task*, so this reads as a drift rather than a deliberate difference.

```suggestion
        *Updated in version 3.17.0: Awaiting scene initialization cancels the awaiting task
```

nicegui/elements/leaflet/leaflet.py
```diff
@@ -101,9 +101,14 @@ def _handle_init(self) -> None:
             self.run_method('add_layer', layer.to_dict(), layer.id)
 
     async def initialized(self) -> None:
-        """Wait until the map is initialized."""
-        await self.client.connected()
-        await self._initialized_event.wait()
+        """Wait until the map is initialized.
+
+        *Updated in version 3.17.0: Awaiting leaflet initialization cancels the calling task
+        when the leaflet is deleted, e.g. because the client disconnected.*
```
evnchn: Same *awaiting task* drift, and the summary line one line up already calls this *the map*.

```suggestion
        *Updated in version 3.17.0: Awaiting map initialization cancels the awaiting task
        when the map is deleted, e.g. because the client disconnected.*
```

nicegui/elements/scene/scene_view.py
```diff
@@ -75,9 +75,14 @@ def _handle_init(self) -> None:
         self.run_method('init')
 
     async def initialized(self) -> None:
-        """Wait until the scene is initialized."""
-        await self.client.connected()
-        await self._initialized_event.wait()
+        """Wait until the scene is initialized.
+
+        *Updated in version 3.17.0: Awaiting scene_view initialization cancels the calling task
+        when the scene_view is deleted, e.g. because the client disconnected.*
```
evnchn: Same *awaiting task* drift, and the bare identifier `scene_view` reaches the rendered docs as prose.

```suggestion
        *Updated in version 3.17.0: Awaiting scene view initialization cancels the awaiting task
        when the scene view is deleted, e.g. because the client disconnected.*
```

evnchn:

**While poking at this branch I stumbled into a `ui.scene_view` leak of the same family that this PR does not cover — and I don't think it should. Parking it here so it isn't lost.**

`scene_view` awaits its *own* `_initialized_event`, but `scene_view.js:11` resolves its scene at mount via `getElement(this.sceneId).scene`. Delete the source `ui.scene` before the view initializes and that read throws, `init` never emits, the event never sets — and the mixin never fires, because the awaited element itself was never deleted. Only its dependency was.

<details><summary><b>Measured in a real browser at <code>6f20d3a1</code></b></summary>

Control first, so the instrument is known to be able to see success:

```python
def test_control_no_deletion(screen: Screen):
    @ui.page('/')
    async def page():
        s = ui.scene()
        v = ui.scene_view(s)
        await v.initialized()
        ui.label('VIEW-INITIALIZED')
    screen.open('/')
    screen.should_contain('VIEW-INITIALIZED')   # passes
```

Then the same page with the scene deleted during the build:

```python
        s = ui.scene()
        v = ui.scene_view(s)
        s.delete()
        try:
            await v.initialized()
            ui.label('RESOLVED')
            ui.label('CANCELLED')
            raise
```

Body stays empty — neither `RESOLVED` nor `CANCELLED` — and the browser console carries:

```
TypeError: Cannot read properties of undefined (reading 'scene')
    at Proxy.mounted (.../components/.../scene_view.js:11:42)
```

So the task is stranded, not cancelled. This is pre-existing on `main` and reads as a `scene_view` lifecycle gap rather than anything the mixin conversion introduced. The window is specifically *scene deleted before the view initializes*; once `init` has emitted, the event is set and a later scene deletion is harmless.

While I was in there I also checked whether the mixin's two entry points disagree here, since `_initialized_event` is element-lifetime state rather than a per-await event like `button`'s: `_handle_delete` refuses to cancel when the event is already set, while `_cancel_when_deleted` raises on `is_deleted` alone. They do disagree — from the same end state, a task already inside the wait returns while a task calling `initialized()` afterwards is cancelled. I am not raising it as a defect: cancelling a fresh await on a deleted element is the safer of the two, and the alternative would let code after the `await` run against an element that is gone.
</details>

<sub>This also narrows my review above: the verification there covered the awaited element being deleted, not a dependency going away first.</sub>

evnchn (dismissed):

**Approve.** I found **no defect in the source change** — the leak is genuinely closed for all three elements, and every guarantee #6271 gives `Button` holds here too. Nicely done, and thanks for switching from your own approach to the mixin; that is what makes this reviewable at a glance.

I took the liberty of updating the **Implementation** section of the description, which still described the pre-mixin design ("store a reference to pending task and cancel them in `scene._handle_delete()`"). The code hasn't done that for a while and it was the only stale thing here — shout if you'd rather word it differently.

@falkoschindler — **the milestone is still `Next`.** Both siblings (#6271, #6322) are on `3.17` and the docstrings here say *"Updated in version 3.17.0"*. Since `fetch_milestone.py` builds the notes by milestone title, as-is this ships **in** 3.17 but is **missing from** its release notes.

<details><summary><b>What I verified, per element</b></summary>

Worktree at `57a9d83d`, base `b86d9d1b`, against `upstream/main` `13a37ee7`.

| check | result |
|---|---|
| MRO | `Scene`/`SceneView`/`Leaflet` → `CancelableWaitElement` → `Element` → `Visibility`; `component=` / `esm=` still register |
| construction order | `_waiting_tasks` set before `super().__init__()`, and all three subclasses call `super().__init__()` as their **first** statement — nothing can raise before it |
| **leak closed** (reverting only the 3 source files) | before: `running_tasks 1→1`, task pending, coroutine frame alive · after: `1→0`, `cancelled()=True`, frame released — all three |
| already-deleted entry path | cancels immediately, 0 leaked. **Positive control:** leaks 1 on unfixed code |
| element-only deletion (`card.clear()`, client alive) | cancels, 0 leaked; positive control leaks 1 |
| real init, then deletion | waiter still resumes — initialization wins over deletion |
| catchable `CancelledError` → `RedirectResponse` | works for all three |
| seen-to-fail | each test goes red on `assert task[0].cancelled()` with **only its own** source file reverted; green restored; stable over 10 runs |
| log noise on cancellation | zero records at DEBUG |
| gates | pre-commit clean · mypy clean (265 files) · pylint 10.00/10 · scene + scene_view + leaflet + scene_objects + scene_custom_types = **53 passed, 0 skipped** |
| merged with current main | no conflicts, 60 passed across scene/scene_view/leaflet/button/event |

The four `test_page.py::test_dark_mode` failures here are pre-existing on `upstream/main` (local macOS appearance) — baseline delta zero.

</details>

<details><summary><b>Bug class: closed on the Python side</b></summary>

Swept every `asyncio.Event()` / `Future` await in `nicegui/`: `Dialog` (#6250, resolves with `None`), `Button.clicked` (#6271), `Event.emitted` (#6322), and these three. Everything else is either bounded by a timeout (`JavaScriptRequest`, `outbox`) or resolved on delete (`Client.disconnected`). Nothing left in this family.

**One JS-side hang this PR does not cause and slightly improves:** `scene_view.js` `mounted()` returns early from the `catch` around `new THREE.WebGLRenderer(...)` **before** the interval that emits `"init"` (`scene.js` has the identical shape at line 107 vs 224). With no WebGL available, `_initialized_event` is never set and `initialized()` never resolves. That is a *hang*, not a *leak*, it is untouched here — and after this PR the stranded task is at least reaped on deletion rather than living to shutdown. Worth its own issue; happy to file it.

**Two pre-existing residuals, neither this PR's business:** an element *created after* its client was already deleted still parks forever (`is_deleted` is `False` on the fresh element — measured for scene, leaflet **and** the already-merged button, so it is a shared mixin limitation whose fix belongs at the client level); and `Layer.current_leaflet` is a `ClassVar` strong reference to the last-created `Leaflet` that is never cleared.

</details>

Two nits, both leave-or-take:

**Test coverage is thinner than the sibling.** #6271 shipped five `Button` tests; this ships one per element. I measured the four uncovered behaviours and they all work — but the already-deleted entry path is the one @falkoschindler singled out on 10 Aug as the half that's easy to miss, and it is the only per-element wiring that could plausibly differ between the three. One extra case would lock it in.

**Test idiom differs from both merged siblings.** #6271/#6322 assert `not any(name in t.get_name() for t in background_tasks.running_tasks)`; these capture `asyncio.current_task()` and assert `.cancelled()`. Both are valid and I confirmed yours is the load-bearing assertion (the preceding `assert not initialized.is_set()` passes with *and* without the fix), so this is consistency, not correctness.

falkoschindler (approved):
Thanks, this closes the last gap from #6277. I pushed two commits on top, please have a look:

- 576a514d2 merges the three identical `test_await_initialized_cancel_task` copies into a parametrized `tests/test_cancelable_wait_element.py`, in the shape of the button tests.
- 2777ca30e moves the wait protocol into the mixin as `_wait_for(event)` and lets an already set event win at entry. This fixes a case the persistent `_initialized_event` had left open: once the element was initialized, the delete guard skipped the cancellation, so an `initialized()` call after a browser disconnect parked in `client.connected()`, survived `Client.delete()`, and resumed against the deleted client. The entry check and the delete guard now follow the same rule, and the four copies of the protocol collapse to one call each. A new test covers the disconnect-after-init case.

One side effect worth knowing: `button.clicked()` on an explicitly deleted button now logs the one-shot #3028 warning, because the click listener is registered before the entry check. I consider that correct, it is a use-after-delete.

Out of scope and tracked separately: `scene_view.initialized()` still hangs when the source scene is deleted first (#6338).

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Review%20findings%20and%20the%20two%20follow-up%20commits%20from%20the%20model%2C%20wording%20from%20the%20model%3B%20decisions%20and%20approval%20from%20the%20maintainer.&d=2026-09-16&a=claude-code "claude-fable-5-1: Review findings and the two follow-up commits from the model, wording from the model; decisions and approval from the maintainer.")

PR #6295, opened 2026-08-16
Fix awaited ui.run_javascript leaking task when client is deleted

### LINKED ISSUE #6278: Awaited run_javascript reports a bogus timeout or leaks its task when the client is deleted
### Description

#6277 covers element-owned awaits; the same class of problem exists in `Client.run_javascript`'s `send_and_wait` ([client.py](https://github.com/zauberzeug/nicegui/blob/main/nicegui/client.py)), which is client-owned and therefore not fixed by resolving element events in `_handle_delete`. Two symptoms:

1. **Bogus timeout error when the client is deleted while waiting for the connection.** `send_and_wait` awaits `client.connected()`, and `Client.delete()` ends with `self._connected.set()  # for terminating connected() waits` — the waiter wakes as a *success*. It then awaits a `JavaScriptRequest` that no browser will ever answer and raises `TimeoutError: JavaScript did not respond within 1.0 s` into the awaiting task, which lands in `app.handle_exception` at ERROR level. This is the same log-noise class as #6226, one caller over: the report blames JavaScript, but the actual cause is a pruned client.

2. **Task leak when awaiting on an already-deleted client.** `connected()` parks forever in that case (`delete()` sets `_connected` once and immediately clears it again; nothing ever sets it afterwards), so the awaiting task stays in `background_tasks.running_tasks` until server shutdown — the same leak class as #6250/#6277.

Minimal example for symptom 1:

```python
from nicegui import ui


@ui.page('/')
async def page():
    width = await ui.run_javascript('window.innerWidth')
    ui.label(f'{width}px')


ui.run()
```

Request the page without opening a websocket (`curl localhost:8080/`) and wait for pruning (~60–70 s): the log shows `TimeoutError: JavaScript did not respond within 1.0 s` even though no JavaScript problem exists. Any bot or prefetch request triggers this.

### Suggested fix

Make `send_and_wait` deletion-aware, mirroring #6229's timer fix: check for deletion before awaiting `connected()` and re-check after the wake, and end the wait gracefully instead of running into the JavaScript timeout — e.g. resolve with `None` (matching `await dialog` in #6250) or raise a dedicated, clearly-worded exception; which of the two is a design decision to make here.

Independent of the caller-side fix, `Client.connected()`'s docstring should state that the method also returns when the client is deleted and that callers must re-check — this contract is currently documented nowhere and is what keeps producing this bug class (#6226, #6250, #6277, this issue).

SaadZahem:
### Motivation

Fixes #6278

A lot of issues are related to this one.
This PR also updates the docstring of `client.connected`

### Implementation

PS (placed higher for importance): @evnchn pointed this out
> ⚠️ One residual worth naming in the PR body rather than hiding behind the green: after the fix the woken builder continues past the None into ui.label('all done') on a deleted client, so check_existence() logs "Client has been deleted but is still being used. This is most likely a bug in your application code." Nothing fails on a WARNING, so the updated assertion passes with it still firing. Whether to also stop the builder is a maintainer call.

#### The fix

Based on the info that `client.connected` returns when the client is connected
_but also when it's deleted_ (I have added this to its docstring),
I added a guard after it that checks whether the client is deleted.
This guard cancels the calling async task since they can't do much without a client.

```diff
diff --git a/nicegui/client.py b/nicegui/client.py
index 8886c135..97ae82a0 100644
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -303,6 +303,9 @@ class Client:
         async def send_and_wait():
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
             await self.connected()
+            if self.is_deleted:
+                return None
             return await JavaScriptRequest(request_id, timeout=timeout)
 
         return AwaitableResponse(send_and_forget, send_and_wait)
```

The guard at the beginning of `client.connected` got extended:
```py
        if self.has_socket_connection or self.is_deleted:
            return
```

#### Testing

I tested using this modified MRE (at first)

```py
import asyncio

from nicegui import app, ui


@ui.page('/')
async def page():
    width = await ui.run_javascript('window.innerWidth')
    ui.label(f'{width}px')


@app.on_startup
async def prune_faster() -> None:
    while True:
        await asyncio.sleep(0.5)
        Client.prune_instances(client_age_threshold=2)


ui.run(show=False)
```

I added 2 tests:
1. test if client disconnected in the middle of `await ui.run_javascript()`
2. test if client disconnected before `await ui.run_javascript()`

#### Documentation

A bit of documentation was added to `client.connected` method mentioning that it raises `ClientConnectionTimeout`.
I've noted the use of double backticks in the docstring.

<details><summary>

#### One suggestion (out of scope)

</summary>

I recommend making `client.connected` return a boolean.
`True` when connected, `False` otherwise.
In that case, the user can use it like this:

```py
    if not await ui.context.client.connected():
        return
```

Otherwise, the user is instructed to check for whether the client is deleted after calling
`client.connected`, which looks like this:

```py
    if ui.context.client.is_deleted:
        return
```

It's not just about writing less code, it's about having a function communicate its purpose clearly.
By having the function returning `True` or `False`, the user can guess that True is when things go right,
and `False` is for when things go otherwise.

That was my suggestion. It's up to the maintainers to decide how to shape the public API.

</details>

### Progress

- [x] The PR title is a short phrase starting with the verb "Fix"
- [x] Documentation was added to `client.connected` advising the user to guard it.

### DIFF
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -257,7 +257,10 @@ class Client:
         return self.page.resolve_title() if self.title is None else self.title
 
     async def connected(self, timeout: float | None = None) -> None:
-        """Block execution until the client is connected.
+        """Block execution until the client is connected or deleted.
+
+        Because this method also returns when the client is deleted,
+        the caller must recheck using `client.is_deleted`.
 
         :param timeout: timeout in seconds (default: ``None``)
         """
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -303,9 +303,6 @@ class Client:
         async def send_and_wait():
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
             await self.connected()
-            if self.is_deleted:
-                asyncio.current_task().cancel()
-                await asyncio.sleep(0)  # let cancellation take effect
             return await JavaScriptRequest(request_id, timeout=timeout)
 
         return AwaitableResponse(send_and_forget, send_and_wait)
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -261,7 +261,7 @@ class Client:
 
         :param timeout: timeout in seconds (default: ``None``)
         """
-        if self.has_socket_connection:
+        if self.has_socket_connection or self.is_deleted:
             return
         self._waiting_for_connection.set()
         self._connected.clear()
@@ -303,6 +303,8 @@ class Client:
         async def send_and_wait():
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
             await self.connected()
+            if self.is_deleted:
+                return None
             return await JavaScriptRequest(request_id, timeout=timeout)
 
         return AwaitableResponse(send_and_forget, send_and_wait)
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -307,7 +307,7 @@ class Client:
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
             await self.connected()
             if self.is_deleted:
-                asyncio.current_task().cancel()
+                cast(asyncio.Task, asyncio.current_task()).cancel()
                 await asyncio.sleep(0)  # let cancellation take effect
             return await JavaScriptRequest(request_id, timeout=timeout)
 
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -257,10 +257,7 @@ class Client:
         return self.page.resolve_title() if self.title is None else self.title
 
     async def connected(self, timeout: float | None = None) -> None:
-        """Block execution until the client is connected or deleted.
-
-        Because this method also returns when the client is deleted,
-        the caller must recheck using `client.is_deleted`.
+        """Block execution until the client is connected.
 
         :param timeout: timeout in seconds (default: ``None``)
         """
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -303,6 +303,9 @@ class Client:
         async def send_and_wait():
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
             await self.connected()
+            if self.is_deleted:
+                asyncio.current_task().cancel()
+                await asyncio.sleep(0)  # let cancellation take effect
             return await JavaScriptRequest(request_id, timeout=timeout)
 
         return AwaitableResponse(send_and_forget, send_and_wait)
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -307,7 +307,7 @@ class Client:
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
             await self.connected()
             if self.is_deleted:
-                cast(asyncio.Task, asyncio.current_task()).cancel()
+                asyncio.current_task().cancel()
                 await asyncio.sleep(0)  # let cancellation take effect
             return await JavaScriptRequest(request_id, timeout=timeout)
 
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -260,6 +260,7 @@ class Client:
         """Block execution until the client is connected.
 
         :param timeout: timeout in seconds (default: ``None``)
+        :raises ClientConnectionTimeout: if ``timeout`` elapses first
         """
         if self.has_socket_connection or self.is_deleted:
             return
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -290,10 +290,14 @@ class Client:
         Internally, ``await client.connected()`` is called before the JavaScript code is executed (*since version 3.0.0*).
         This might delay the execution of the JavaScript code and is not covered by the ``timeout`` parameter.
 
+        *Updated in version 3.17.0: Awaiting the response resolves with ``None`` when the client has been deleted,
+        e.g. because the browser tab was closed.*
+
         :param code: JavaScript code to run
         :param timeout: timeout in seconds (default: 1.0)
 
-        :return: AwaitableResponse that can be awaited to get the result of the JavaScript code
+        :return: AwaitableResponse that can be awaited to get the result of the JavaScript code,
+            or ``None`` if the client has been deleted
         """
         request_id = str(uuid.uuid4())
         target_id = self._temporary_socket_id or self.id
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -303,9 +303,13 @@ class Client:
         target_id = self._temporary_socket_id or self.id
 
         def send_and_forget():
+            if self.is_deleted:
+                return
             self.outbox.enqueue_message('run_javascript', {'code': code}, target_id)
 
         async def send_and_wait():
+            if self.is_deleted:
+                return None
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
             await self.connected()
             if self.is_deleted:
--- a/nicegui/client.py
+++ b/nicegui/client.py
@@ -302,12 +302,12 @@ class Client:
         request_id = str(uuid.uuid4())
         target_id = self._temporary_socket_id or self.id
 
-        def send_and_forget():
+        def send_and_forget() -> None:
             if self.is_deleted:
                 return
             self.outbox.enqueue_message('run_javascript', {'code': code}, target_id)
 
-        async def send_and_wait():
+        async def send_and_wait() -> Any:
             if self.is_deleted:
                 return None
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
--- a/tests/test_speculative_loading.py
+++ b/tests/test_speculative_loading.py
@@ -117,9 +117,11 @@ def test_prefetch_connects_after_navigation(screen: Screen, event_log: EventLog)
     screen.click('answer')
     event_log.wait_for('connect: /answer')
     event_log.wait_for('answer: 42')
-    # if the client was deleted, shouldn't the answer be None?
-    assert event_log.items == ['answer: called', 'connect: /', 'answer: called', 'connect: /answer', 'answer: 42'], \
-        'answer() should re-evaluate after prefetch client was pruned'
+    assert event_log.items == [
+        'answer: called', 'connect: /',
+        'answer: None',
+        'answer: called', 'connect: /answer', 'answer: 42',
+    ], 'answer() should re-evaluate after prefetch client was pruned'
     screen.should_contain('all done')
 
 
--- a/tests/test_speculative_loading.py
+++ b/tests/test_speculative_loading.py
@@ -117,6 +117,7 @@ def test_prefetch_connects_after_navigation(screen: Screen, event_log: EventLog)
     screen.click('answer')
     event_log.wait_for('connect: /answer')
     event_log.wait_for('answer: 42')
+    # if the client was deleted, shouldn't the answer be None?
     assert event_log.items == ['answer: called', 'connect: /', 'answer: called', 'connect: /answer', 'answer: 42'], \
         'answer() should re-evaluate after prefetch client was pruned'
     screen.should_contain('all done')
--- a/nicegui/functions/javascript.py
+++ b/nicegui/functions/javascript.py
@@ -16,9 +16,13 @@ def run_javascript(code: str, *, timeout: float = 1.0) -> AwaitableResponse:
     Internally, ``await client.connected()`` is called before the JavaScript code is executed (*since version 3.0.0*).
     This might delay the execution of the JavaScript code and is not covered by the ``timeout`` parameter.
 
+    *Updated in version 3.17.0: Awaiting the response resolves with ``None`` when the client has been deleted,
+    e.g. because the browser tab was closed.*
+
     :param code: JavaScript code to run
     :param timeout: timeout in seconds (default: 1.0)
 
-    :return: AwaitableResponse that can be awaited to get the result of the JavaScript code
+    :return: AwaitableResponse that can be awaited to get the result of the JavaScript code,
+        or ``None`` if the client has been deleted
     """
     return context.client.run_javascript(code, timeout=timeout)
--- a/tests/test_javascript.py
+++ b/tests/test_javascript.py
@@ -1,5 +1,8 @@
-from nicegui import ui
-from nicegui.testing import Screen
+import asyncio
+
+from nicegui import background_tasks, ui
+from nicegui.client import Client
+from nicegui.testing import Screen, User
 
 
 def test_run_javascript_on_button_press(screen: Screen):
@@ -93,3 +96,37 @@ def test_simultaneous_async_javascript(screen: Screen):
     screen.click('runB')
     screen.should_contain('A: 1')
     screen.should_contain('B: 2')
+
+
+async def test_awaited_run_javascript_resolves_when_client_is_deleted_while_waiting(user: User):
+    results = []
+
+    @ui.page('/')
+    async def page():
+        results.append(await ui.run_javascript('window.innerWidth'))
+
+    await user.http_client.get('/')  # request the page without ever opening the websocket
+    await asyncio.sleep(0)
+    Client.prune_instances(client_age_threshold=0)  # delete the client, waking up connected()
+    await asyncio.sleep(1.5)  # longer than run_javascript's default 1.0 s timeout
+    assert results == [None]
+
+
+async def test_awaited_run_javascript_resolves_when_client_is_already_deleted(user: User):
+    results = []
+
+    @ui.page('/')
+    async def page():
+        client = ui.context.client
+
+        async def later() -> None:
+            await asyncio.sleep(0.3)  # resume long after the client is gone
+            results.append(await client.run_javascript('window.innerWidth'))
+
+        background_tasks.create(later())
+
+    await user.http_client.get('/')  # request the page without ever opening the websocket
+    await asyncio.sleep(0)
+    Client.prune_instances(client_age_threshold=0)
+    await asyncio.sleep(1.5)
+    assert results == [None]
--- a/tests/test_javascript.py
+++ b/tests/test_javascript.py
@@ -1,7 +1,6 @@
 import asyncio
 
-from nicegui import background_tasks, ui
-from nicegui.client import Client
+from nicegui import Client, ui
 from nicegui.testing import Screen, User
 
 
@@ -98,35 +97,23 @@ def test_simultaneous_async_javascript(screen: Screen):
     screen.should_contain('B: 2')
 
 
-async def test_awaited_run_javascript_resolves_when_client_is_deleted_while_waiting(user: User):
+async def test_awaited_run_javascript_resolves_when_client_is_deleted(user: User):
+    """The task awaiting run_javascript must not time out or wait forever when the client is deleted, e.g. after a disconnect."""
+    clients: list[Client] = []
     results = []
 
     @ui.page('/')
     async def page():
+        clients.append(ui.context.client)
         results.append(await ui.run_javascript('window.innerWidth'))
 
     await user.http_client.get('/')  # request the page without ever opening the websocket
     await asyncio.sleep(0)
+    assert not results
+
     Client.prune_instances(client_age_threshold=0)  # delete the client, waking up connected()
-    await asyncio.sleep(1.5)  # longer than run_javascript's default 1.0 s timeout
-    assert results == [None]
+    await asyncio.sleep(0.1)  # let the page function resume
+    assert results == [None]  # the page function resumed with None instead of timing out
 
-
-async def test_awaited_run_javascript_resolves_when_client_is_already_deleted(user: User):
-    results = []
-
-    @ui.page('/')
-    async def page():
-        client = ui.context.client
-
-        async def later() -> None:
-            await asyncio.sleep(0.3)  # resume long after the client is gone
-            results.append(await client.run_javascript('window.innerWidth'))
-
-        background_tasks.create(later())
-
-    await user.http_client.get('/')  # request the page without ever opening the websocket
-    await asyncio.sleep(0)
-    Client.prune_instances(client_age_threshold=0)
-    await asyncio.sleep(1.5)
-    assert results == [None]
+    # calling on a deleted client resolves immediately
+    assert await clients[0].run_javascript('window.innerWidth') is None


falkoschindler (changes requested):
Thanks for picking this up, and for tracing the mechanism so precisely — the diagnosis in the PR body is exactly right. I think the change needs a couple of adjustments before it can go in, though: the guard sits one step too late to cover both symptoms of #6278, and cancelling the caller's task turns out to be a sharper tool than the situation needs.

### Major

1. **Symptom 2 of #6278 (already-deleted client) is not fixed — the new guard is unreachable in that case** — [client.py 307-309](https://github.com/zauberzeug/nicegui/blob/b69ea25b5/nicegui/client.py). When the client was already deleted before `send_and_wait` runs (e.g. an event handler resuming after `await asyncio.sleep(...)` once the tab is gone and `reconnect_timeout` has elapsed), `has_socket_connection` is `False` (`handle_disconnect` reset `tab_id`) and `_connected` was set-then-cleared by `delete()`, so `connected()` clears the event again and waits on something nothing will ever set. The `if self.is_deleted:` check on line 309 is never reached, the task stays in `background_tasks.running_tasks` until shutdown, and — because `send_and_wait` closes over `self` — the deleted `Client` and its element tree are pinned in memory too. Along the way `enqueue_message` also fires the "Client has been deleted but is still being used" warning for what isn't a user bug. The issue's suggested fix asks for a check *before* `connected()` as well as after; `Timer._can_start` (a9121acf0) does exactly that (pre-check → `await connected()` → post-check). Since the PR says "Fixes #6278", either both symptoms need covering or the description should scope it down.

2. **Cancelling the caller's task is the wrong resolution shape** — [client.py 310-311](https://github.com/zauberzeug/nicegui/blob/b69ea25b5/nicegui/client.py). `send_and_wait` runs inline in whatever task awaits the `AwaitableResponse`, and that task is not necessarily owned by the deleted client. Concrete cases where the new behavior is worse than the old catchable `TimeoutError`:
   - An `app.timer`/`ui.timer` on a *live* client broadcasting `for c in list(Client.instances.values()): with c: await c.run_javascript(...)` — one pruned client cancels the invocation task, `Timer._run_in_loop` hits `except asyncio.CancelledError: break` ([timer.py 99](https://github.com/zauberzeug/nicegui/blob/b69ea25b5/nicegui/timer.py)) and the timer stops permanently, silently.
   - A plain FastAPI route awaiting a stored client's `run_javascript` — uvicorn's request task is cancelled, which logs "Exception in ASGI application" at ERROR and returns a 500; the log-noise class this PR sets out to remove reappears in a different logger.
   - `except TimeoutError: fallback` / `except Exception:` handlers in user code no longer fire, and everything after the await (DB writes, notifications on another client) is silently skipped, since every internal wrapper (`events._await_and_handle_in_context`, `_handle_exceptions`, `page.wait_for_result`) lets `CancelledError` through.
   - `asyncio.gather(..., return_exceptions=True)` over clients now yields `CancelledError` *instances* as results, which `isinstance(r, Exception)` filters don't catch.
   
   The two sibling fixes both chose value-shaped outcomes: `await dialog` resolves with `None` (#6250), `Timer._can_start` returns `False` (#6229). `Element.run_method` — which delegates to this very code — already returns `NullResponse` (awaits to `None`) on a dead client. Resolving with `None` here (i.e. `return None`) matches all three, keeps the decision with the caller, and needs no `cast`/`current_task`/`sleep(0)`. If you'd rather raise, the issue offers "a dedicated, clearly-worded exception" as the alternative — that's a maintainer call, but self-cancellation is a third option nobody proposed and it has the widest blast radius. (If cancellation is kept anyway, `raise asyncio.CancelledError` is the one-line equivalent and avoids bumping the task's `cancelling()` counter on 3.11+.)

3. **The fix belongs one level up, in `connected()` itself.** The docstring change turns "recheck `is_deleted` after `connected()`" into every caller's job, but four in-repo callers don't do it and now formally violate the documented contract: `Button.clicked()` ([button.py 66](https://github.com/zauberzeug/nicegui/blob/b69ea25b5/nicegui/elements/button.py)), and `initialized()` in leaflet, scene and scene_view — each awaits `connected()` and then an event nobody sets on delete, so they park forever after a prune (the #6277 class). This PR is the third caller-side patch for the same primitive (after `disconnected()` and `Timer._can_start`), and the three use three different recheck idioms (`is_deleted`, `client.id not in Client.instances`, `self.id in self.instances`). A one-line `if self.has_socket_connection or self.is_deleted: return` at the top of `connected()` fixes symptom 2 for all callers at once without any API change and leaves `send_and_wait` with just the single post-wake check. Your `-> bool` suggestion would be the cleaner API long-term, but it's a public-signature change and can be a separate discussion; the early return is orthogonal and cheap. Not asking you to fix the four callers here — a follow-up reference is enough — but the docstring shouldn't promise a contract that the file's own neighbours break.

### Minor

4. **New `connected()` docstring overstates what the code does** — [client.py 260-263](https://github.com/zauberzeug/nicegui/blob/b69ea25b5/nicegui/client.py). "Returns when the client is deleted" is only true for a waiter parked at the moment of deletion; a call *after* deletion never returns (finding 1), and with `timeout` set it raises `ClientConnectionTimeout` rather than returning — worth a `:raises ClientConnectionTimeout:` line. Both go away if `connected()` gets the early return from finding 3. Also `run_javascript`'s own docstring (line 292) and the mirror in `functions/javascript.py` don't mention the new deletion behavior at all, while that's where users will look.

5. **RST markup: single backticks in the docstring** — [client.py 263](https://github.com/zauberzeug/nicegui/blob/b69ea25b5/nicegui/client.py). Docstrings are rendered as reStructuredText (CONTRIBUTING.md), where single backticks are *interpreted text* (italic), not inline code. The surrounding lines use double backticks (` ``None`` `, ` ``await client.connected()`` `), so this should be ` ``client.is_deleted`` `.

6. **The reason given for skipping a test doesn't hold.** "Tearing down the client [is] the kind of tests we advise against" — but that's exactly what the two sibling tests do with the fast `User` fixture and no internals: `test_no_error_when_timer_is_deleted_while_waiting_for_connection` in [tests/test_timer.py](https://github.com/zauberzeug/nicegui/blob/b69ea25b5/tests/test_timer.py) (`await user.http_client.get('/')` + `Client.prune_instances(client_age_threshold=0)`, then relies on the fixture failing on any logged ERROR) and `test_awaited_dialog_resolves_when_client_is_deleted` in tests/test_dialog.py (`dialog.client.delete()`). The same ~10-line shape covers symptom 1 here — page awaits `ui.run_javascript`, request without websocket, prune, assert no `TimeoutError` is logged — and would fail on `main`. A second one with a handler that awaits `run_javascript` *after* the prune would have caught finding 1. Not insisting on a test per AGENTS.md, but the justification should be either corrected or replaced by the test.

Happy to help sketch the `connected()` variant if you want to go that route. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20from%20a%20multi-agent%20code%20review%2C%20wording%20and%20structure%20from%20the%20model%3B%20the%20maintainer%20requested%20changes%20based%20on%20the%20summary%20without%20reading%20the%20full%20text.&p=agent&d=2026-08-17&a=claude-code "claude-fable-5: Findings from a multi-agent code review, wording and structure from the model; the maintainer requested changes based on the summary without reading the full text.")

SaadZahem:
I reverted all the 3 initial commits. That includes canceling caller's task and doc-string change.
The newest commit [23e03e6](https://github.com/zauberzeug/nicegui/pull/6295/commits/23e03e6d63a618b07773ad643c0d9d3eabf17aa7) adds 2 guards:
1. At the beginning of `client.connected`, exit early if client was already deleted (symptom 2).
2. After `await self.connected` in `send_and_wait`, resolve with `None` if the client was deleted while awaiting (symptom 1).

That brings the PR on the right track, I think.
I am still considering adding a test and updating the doc-string before I request a review again.

Is the current shape of `send_and_wait` good?
I made it resolve to `None` which hands the control back to the caller.
Would it be better to raise an error instead?

evnchn:

**The reshape is right, and `None` is the better of the two shapes you offered.**

- It matches all three siblings — an awaited `ui.dialog` resolves to `None`, `Timer._can_start` returns `False`, `Element.run_method` hands back a `NullResponse` on a dead client.
- It keeps the decision with the caller, and drops the `cast` / `current_task` / `sleep(0)`.
- Raising is defensible (the JavaScript genuinely did not run), but uncaught it reintroduces the same page/background-task error noise you set out to remove.

I ran 23e03e6 against `main` in both directions: both symptoms are fixed and both guards sit in the right place. Three things still look open.

<details>
<summary><b>Verified — what was run</b></summary>

Worktree at 23e03e6d, `uv sync --group dev --group types`, then the two tests below at the PR head and again with `nicegui/client.py` reverted to `main`:

| Scenario | `main` | 23e03e6 |
|---|---|---|
| deleted **while** awaiting `run_javascript` | `TimeoutError: JavaScript did not respond within 1.0 s`, late page task builds a 500 error response, `results == []` | resolves to `None` |
| called **after** the client is gone | never returns — `results == []` after 1.5 s | resolves to `None` |

- The `delete()` ordering is load-bearing: `self._deleted = True` runs *before* `self._connected.set()`, so a waiter parked in `connected()` always sees `is_deleted` when it wakes. Reverse those two lines and the new guard silently stops working.
- Not covered: a live-client broadcast loop, a plain FastAPI route, `asyncio.gather` over several clients — the cases the cancellation shape would have broken. Those argue *for* `None`; they are not evidence I collected.

</details>

<details>
<summary><b>Two tests, if you want them</b></summary>

- Both fail on `main` and pass at 23e03e6, so they guard the change rather than accompany it.
- Home: `tests/test_javascript.py`, next to `test_run_javascript_before_client_connected`.
- Neither reaches past `prune_instances` / `client.delete()` into internals — same as the two sibling tests.

```python
async def test_awaited_run_javascript_resolves_when_client_is_deleted_while_waiting(user: User):
    results = []

    @ui.page('/')
    async def page():
        results.append(await ui.run_javascript('window.innerWidth'))

    await user.http_client.get('/')  # request the page without ever opening the websocket
    await asyncio.sleep(0)
    Client.prune_instances(client_age_threshold=0)  # delete the client, waking up connected()
    await asyncio.sleep(1.5)  # longer than run_javascript's default 1.0 s timeout
    assert results == [None]


async def test_awaited_run_javascript_resolves_when_client_is_already_deleted(user: User):
    results = []

    @ui.page('/')
    async def page():
        client = ui.context.client

        async def later() -> None:
            await asyncio.sleep(0.3)  # resume long after the client is gone
            results.append(await client.run_javascript('window.innerWidth'))

        background_tasks.create(later())

    await user.http_client.get('/')  # request the page without ever opening the websocket
    await asyncio.sleep(0)
    Client.prune_instances(client_age_threshold=0)
    await asyncio.sleep(1.5)
    assert results == [None]
```

Reverted-`client.py` run, for the record:

```
FAILED test_awaited_run_javascript_resolves_when_client_is_deleted_while_waiting - assert [] == [None]
    TimeoutError: JavaScript did not respond within 1.0 s
FAILED test_awaited_run_javascript_resolves_when_client_is_already_deleted - assert [] == [None]
2 failed, 1 error in 3.15s
```

The first also errors in teardown on `main` (the late page task's 500 error response), hence `2 failed, 1 error`.

</details>

<details>
<summary><b>Three things still open</b></summary>

**1. The spurious warning still fires on the already-deleted path.** `send_and_wait` enqueues before it awaits, so `enqueue_message` → `check_existence()` still logs *"Client has been deleted but is still being used. This is most likely a bug in your application code."* — for exactly the case this PR establishes is not a bug in application code.

- Observed at 23e03e6 on the second test:
  ```
  outbox.py:77   self.client.check_existence()
  client.py:480  helpers.warn_once('Client has been deleted but is still being used. ...')
  warnings.py:9  log.warning(message, stack_info=stack_info)
  ```
- It is `warn_once` at WARNING, and the `user` fixture only fails on ERROR — so nothing catches it unless a test asserts on `caplog`.
- Small fix: `if self.is_deleted: return None` at the top of `send_and_wait`, before the enqueue.
- *Not* the enqueue-after-`connected()` variant: it shuffles FIFO order against anything else queued during the wait, and it does not help `send_and_forget` — `AwaitableResponse._fire` calls that one synchronously, so it needs its own plain `if self.is_deleted: return`.
- Whether any of this belongs in this PR is a scope call for the maintainer.

**2. The early return does not free the four `connected()` callers** — it only moves where they park.

- Probe at 23e03e6 — a page doing `await ui.button('Go').clicked()`, then `prune_instances(client_age_threshold=0)`:
  ```
  DONE: []   PARKED TASKS: ['wait for result of page "/"']
  ```
- `initialized()` in leaflet, scene and scene_view is the identical two-liner (`await self.client.connected()` then `await self._initialized_event.wait()`) — read, not run.
- The early return is still the right foundation; the follow-up just stays real, and a note in the PR body covers it without code here.

**3. The docstring is the last thing un-updated.** Reverting it left `connected()` returning early for a deleted client with nothing saying so.

- The early return makes the sentence unconditionally true now: it returns immediately if the client is already deleted, and wakes and returns if the client is deleted while waiting.
- Worth carrying over from the review: a `:raises ClientConnectionTimeout:` line, and double backticks — docstrings render as reStructuredText, where `` `client.is_deleted` `` comes out italic rather than as code.

</details>

Nothing here needs a re-request from me — 1 and 2 are scope calls, and 3 is the one I would fix before asking for review again. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Verification%20runs%20against%20the%20PR%20head%2C%20the%20two%20draft%20tests%2C%20and%20the%20wording%20are%20the%20model%27s.&d=2026-08-18&a=claude-code "claude-opus-5: Verification runs against the PR head, the two draft tests, and the wording are the model's.")

SaadZahem:
You came in the right time, I was scratching my head trying to write a test that captures the issue.
One more thing, `test_prefetch_connects_after_navigation` currently fails and I am not sure what to do about it.

nicegui/client.py
```diff
@@ -261,7 +261,7 @@ async def connected(self, timeout: float | None = None) -> None:
 
         :param timeout: timeout in seconds (default: ``None``)
```
evnchn, line 262: Item 3 — the `:raises` line the review asked for. Note `:raises` is not used anywhere in `nicegui/` yet, so this is the first.

```suggestion
        :param timeout: timeout in seconds (default: ``None``)
        :raises ClientConnectionTimeout: if ``timeout`` elapses first
```

nicegui/client.py
```diff
@@ -303,6 +303,8 @@ def send_and_forget():
         async def send_and_wait():
             self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
```
evnchn, line 305: Item 1 — optional, only if the leftover warning is in scope for this PR. The post-wake check below stays necessary either way.

```suggestion
        async def send_and_wait():
            if self.is_deleted:
                return None
            self.outbox.enqueue_message('run_javascript', {'code': code, 'request_id': request_id}, target_id)
```

evnchn (commented):

Suggestions for two of the open items — three added lines between them. Verified at 23e03e6 with both applied: the two probe tests pass, `caplog` records no WARNING, and `ruff` / `autopep8` / `mypy` / `pylint` are clean at 10.00. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Both%20suggestions%20were%20drafted%20and%20verified%20by%20the%20model%20against%20the%20PR%20head.&d=2026-08-18&a=claude-code "claude-opus-5: Both suggestions were drafted and verified by the model against the PR head.")

SaadZahem:
Thanks for the 2 tests.
I've verified them myself.

nicegui/client.py
SaadZahem: I still see a warning with this applied. It originates from the `ui.label` placed after `ui.run_javascript` and disappears when the label is removed and no element is after `ui.run_javascript`.

tests/test_speculative_loading.py
```diff
@@ -117,6 +117,7 @@ async def answer() -> None:
     screen.click('answer')
     event_log.wait_for('connect: /answer')
     event_log.wait_for('answer: 42')
+    # if the client was deleted, shouldn't the answer be None?
     assert event_log.items == ['answer: called', 'connect: /', 'answer: called', 'connect: /answer', 'answer: 42'], \
         'answer() should re-evaluate after prefetch client was pruned'
```
evnchn: Answers the comment and updates the expectation together. Reflowed because the one-line form reaches 133 chars against ruff's 120; the shape follows `tests/test_background_tasks.py`.

```suggestion
    assert event_log.items == [
        'answer: called', 'connect: /',
        'answer: None',
        'answer: called', 'connect: /answer', 'answer: 42',
    ], 'answer() should re-evaluate after prefetch client was pruned'
```

evnchn (commented):

**Yes — and editing this assertion is legitimate, not papering over.** The suggestion below drops your marker comment and updates the expected log in one commit.

- On `main` that parked call raises `TimeoutError: JavaScript did not respond within 1.0 s` and dies there. The missing `answer: None` was the exception killing the coroutine — not intended behaviour.
- Your change removes that `TimeoutError`, so the line appears. What the test actually claims still holds: handshake still fails, reload still happens, `answer()` still re-evaluates, still ends at `42`, `all done` still shows.
- The test came from #5116, whose motivation was *"Chrome pre-loads content when typing an url. This raises `TimeoutError` messages"*. This path still had one; you removed it. Test and PR are the same campaign.

⚠️ One residual worth naming in the PR body rather than hiding behind the green: after the fix the woken builder continues past the `None` into `ui.label('all done')` on a deleted client, so `check_existence()` logs *"Client has been deleted but is still being used. This is most likely a bug in your application code."* Nothing fails on a WARNING, so the updated assertion passes with it still firing. Whether to also stop the builder is a maintainer call.

<details>
<summary><b>What I ran</b></summary>

Instrumented the same scenario at 23e03e6, printing the client id at each step:

```
BUILD START  client=4839 prefetch=True
--- prune ---
RAISED       client=4839 TimeoutError: JavaScript did not respond within 1.0 s     <- on main
--- click ---
BUILD START  client=f5e7 prefetch=False
RETURNED     client=f5e7 result=42
```

With your `client.py`, the middle line instead reads `RETURNED client=4839 result=None deleted=True`, followed by `ui.label` running on that deleted client. Browser console confirms the rebuild path: `reloading because handshake failed for clientId 4839…`.

On `main` nothing is logged for that `TimeoutError` at any level — I checked at `log_cli_level=DEBUG`. I did not pin down which branch swallows the resulting 500 (`check_for_late_return_value` looks like it should `log.error`, and does not), so treat that mechanism as unestablished; it does not affect the assertion.

After the suggestion, locally on macOS / 3.14: `5 passed` for the whole file, `ruff` / `autopep8` / `pre-commit` clean. CI is the authority on green, though — the red run was Linux / 3.10.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=The%20main-branch%20trace%2C%20the%20reflowed%20assertion%20and%20the%20wording%20are%20the%20model%27s.&d=2026-08-18&a=claude-code "claude-opus-5: The main-branch trace, the reflowed assertion and the wording are the model's.")

---

*Changelog — 2026-08-18: moved the colophon sigil onto its own line; on the same line as `</details>` GFM left it as literal text.*

nicegui/client.py
evnchn: *Posted by Claude Code on @evnchn's behalf.*

You're right, and so is the guard — we were each looking at a different path. Measured both at head `e08f08a9`:

| path | head as-is | head + the guard |
|---|---|---|
| client **already deleted** before the call | `RESULT=None`, `WARNINGS=['Client has been deleted but is still being used…']` | `RESULT=None`, `WARNINGS=[]` |
| client deleted **while waiting** | warning present | warning present (your `ui.label` case) |

So your observation holds for the deleted-while-waiting path — the warning there really does come from the `ui.label`, and the guard cannot touch it. On the already-deleted path the guard removes the warning outright.

Still optional and still your call: it's one branch for one of the two paths, and the PR is correct without it. I'm only reopening it because "I still see a warning" reads like the guard did nothing, and it does do something — just not on the path you tested.

evnchn:

**Optional, not a condition of anything** — the last of @falkoschindler's six findings that is still open (his item 4: the docstrings don't mention deletion). Everything else from his 17 Aug review is addressed at `e08f08a9`.

It can't be a suggestion block: the diff's `client.py` hunks cover lines 260-268 and 304-311, the docstring is at 283-296, and `nicegui/functions/javascript.py` isn't in the diff at all. So, pasted instead.

`nicegui/client.py`, in `run_javascript`'s docstring:

```
    :return: response from the browser, or ``None`` if the client has been deleted

    *Updated in version 3.17.0: Returns ``None`` instead of hanging when the client has been deleted.*
```

`nicegui/functions/javascript.py`, same two lines in `run_javascript`'s docstring.

Two reasons it's worth the sentence rather than leaving it implicit: `None` from a deleted client is indistinguishable from a JS expression that legitimately returned `null`, so a caller can't tell them apart without being told; and both siblings in this family carry the `*Updated in version X.Y.Z:*` marker — `button.py:64` from #6271, and the three elements in #6294 — which you wrote yourself, so this is your own convention rather than one I'm importing.

nicegui/client.py
falkoschindler: Applied in 08f22dd43, in both closures: `send_and_wait` and `send_and_forget`. The deciding argument is `Element._is_safe_to_interact` on `main`, which already spells out the rule for `run_method`: silent when the *client* has been deleted, because an async callback resuming after the teardown is not a user bug. `run_javascript` should behave the same. The remaining warning on the deleted-while-waiting path comes from building elements on a dead client after the `None`, which is exactly what `check_existence()` is for, so that one stays.

I also pushed the `run_javascript` docstring lines from your other comment as 31c6dee1b. Thanks, both of you, for measuring the two paths separately. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Decision%20and%20go%20from%20the%20maintainer%3B%20the%20code%20change%2C%20the%20argument%20from%20_is_safe_to_interact%20and%20the%20wording%20are%20the%20model%27s.&d=2026-09-10&a=claude-code "claude-fable-5-1: Decision and go from the maintainer; the code change, the argument from _is_safe_to_interact and the wording are the model's.")

falkoschindler (approved):
Thanks @SaadZahem, and thanks @evnchn for the measurements along the way. This has landed exactly where the first review hoped it would: the early return in `connected()` covers the already-deleted case for every caller, and `send_and_wait` resolves with `None` instead of cancelling or timing out, in line with `await dialog` and `run_method`.

I pushed a few small commits on top rather than asking for another round: the `run_javascript` docstrings now carry the 3.17.0 marker and the `None` case, both closures skip the enqueue when the client is already gone (the `_is_safe_to_interact` rule from `run_method` applied here), type annotations for the two nested functions, and the two new tests merged into one that runs in well under a second instead of three.

For the record, since it will come up: `None` is ambiguous with a JavaScript expression that legitimately returns `null`. We looked at raising instead and decided against it for 3.17, because the trigger is routine (tab closed, prefetch pruned) and an exception would put the error noise back that #6278 set out to remove. A dedicated exception that the page and task machinery swallow for the client's own tasks would be the stricter design, and would have to cover `ui.dialog` as well. That is a separate issue if we ever want it. The three `initialized()` callers in leaflet, scene and scene_view still await `connected()` without a recheck; that is the remaining part of #6277. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Verdict%20and%20go%20from%20the%20maintainer%3B%20the%20verification%20runs%20and%20the%20wording%20are%20the%20model%27s.&d=2026-09-10&a=claude-code "claude-fable-5-1: Verdict and go from the maintainer; the verification runs and the wording are the model's.")

PR #6297, opened 2026-08-17
Fix `page.py`: don't propagate CancelledError from late `task.result()` (#6296)

### LINKED ISSUE #6296: Page builder returns HTTP 500 "No response returned." when its task is cancelled

### First Check

### Example Code

```python
import asyncio

from nicegui import ui


@ui.page('/ok')
async def ok():
    ui.label('reached the end')


@ui.page('/broken')
async def broken():
    task = asyncio.current_task()
    assert task is not None
    task.cancel()                  # any cancellation of the page-builder task
    await asyncio.sleep(0)
    ui.label('reached the end')


ui.run(port=8123, show=False, reload=False)
```

### Description

Open `/ok` — HTTP 200, as expected.

Open `/broken` — **HTTP 500**, and the server logs an unhandled `RuntimeError: No response returned.` from Starlette. I would expect the response to be built from whatever the page produced before the cancellation, the way it already is when the builder is still running at response time.

The cause is at `nicegui/page.py:207`:

```python
if task.done():
    result = task.result()        # re-raises CancelledError out of the route handler
else:
    result = None
```

If the page-builder task finished *cancelled*, `task.result()` re-raises `CancelledError`, it escapes the route handler, and Starlette turns that into `No response returned.`

This is the only unguarded `.result()` on a response path — `page.py:128` (`check_for_late_return_value`) and `background_tasks.py:152` both already do `except asyncio.CancelledError: pass`. That asymmetry is visible in behaviour: a cancellation that arrives *after* the response is built goes down the guarded late path and is fine, while one that lands *before* the response is built produces the 500.

A one-line change makes the two paths agree, reusing the value the `else` branch two lines down already uses:

```diff
                 if task.done():
-                    result = task.result()
+                    result = None if task.cancelled() else task.result()
                 else:
                     result = None
```

Measured with the file above:

| | `/ok` | `/broken` |
|---|---|---|
| v3.16.0 (`17ad9005`) | 200 | **500** |
| main (`9b5133ca`) | 200 | **500** |
| main + the one line | 200 | 200 |

With the fix, `/broken` returns 200 serving the partially-built page and logs no unhandled error.

### NiceGUI Version

3.16.0 (also reproduced on main at `9b5133ca`)

### Python Version

3.12.11

### Browser

Other — not browser-dependent. The symptom is the HTTP status and the server-side traceback; reproduced with `curl`, no browser involved.

### Operating System

macOS

### Additional Context

`result = None` cannot mask an application error here: the page-builder task runs with `handle_exceptions=False` and converts ordinary exceptions into a 500 error page itself, so `task.cancelled()` is true only in the case where `task.result()` would raise plain `CancelledError`. The `response_timeout` branch above cancels the task and returns at line 203, so it never reaches line 207 either.

The self-cancel in the example is deliberately artificial — it isolates the defect from any particular caller. I have not found a natural user-level pattern that reaches it on current main, so this is latent today. It stops being latent with #6271 and #6294, which both add `task.cancel()` on the current task inside `clicked()` / `initialized()` when the element is already deleted; with either of those applied, an ordinary page builder reaches this line and 500s. Both were measured at `ca5243ce` and `6f2dc3df` respectively, and the same one line clears both.

Worth a route-level regression test alongside the fix: the tests in those two PRs assert element-level `task.cancelled()`, which passes either way and would not catch this.

SaadZahem:
### Motivation

Fixes #6296
Related to #6271 and #6294

### Implementation

```diff
diff --git a/nicegui/page.py b/nicegui/page.py
index 2ced4129..09305075 100644
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -204,7 +204,7 @@ class page:
                 if not task_wait_for_connection.done():
                     task_wait_for_connection.cancel()
                 if task.done():
-                    result = task.result()
+                    result = None if task.cancelled() else task.result()
                 else:
                     result = None
                     task.add_done_callback(check_for_late_return_value)
```

One pytest was added to `test_page.py`.
It asserts that the response status code from a cancelled async page is 200.

### Progress

### DIFF
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -204,7 +204,7 @@ class page:
                 if not task_wait_for_connection.done():
                     task_wait_for_connection.cancel()
                 if task.done():
-                    result = task.result()
+                    result = None if task.cancelled() else task.result()
                 else:
                     result = None
                     task.add_done_callback(check_for_late_return_value)
--- a/nicegui/page.py
+++ b/nicegui/page.py
@@ -204,7 +204,12 @@ class page:
                 if not task_wait_for_connection.done():
                     task_wait_for_connection.cancel()
                 if task.done():
-                    result = None if task.cancelled() else task.result()
+                    if task.cancelled():
+                        log.warning(f'Page building for {client.page.path} was cancelled; '
+                                    'serving the elements created so far')
+                        result = None
+                    else:
+                        result = task.result()
                 else:
                     result = None
                     task.add_done_callback(check_for_late_return_value)
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -205,16 +205,17 @@ def test_ui_page_http_exception_404_keeps_html(screen: Screen):
         'ui.page raising HTTPException(404) should render the HTML error page for browsers'
 
 
-async def test_normal_response_when_async_page_is_cancelled(screen: Screen):
+def test_normal_response_when_async_page_is_cancelled(screen: Screen):
     @ui.page('/')
     async def page():
-        task = asyncio.current_task()
-        task.cancel()
+        asyncio.current_task().cancel()
         ui.label('The end')
 
     screen.start_server()
     response = httpx.get(f'http://localhost:{Screen.PORT}/')
-    assert response.status_code == 200
+    assert response.status_code == 200, 'page cancelling its own task should still return a normal response'
+    assert 'The end' in response.text, 'elements created before the cancellation takes effect should be served'
+    screen.assert_py_logger('WARNING', re.compile('Page building for / was cancelled'))
 
 
 def test_page_with_args(screen: Screen):
--- a/tests/test_page.py
+++ b/tests/test_page.py
@@ -205,6 +205,18 @@ def test_ui_page_http_exception_404_keeps_html(screen: Screen):
         'ui.page raising HTTPException(404) should render the HTML error page for browsers'
 
 
+async def test_normal_response_when_async_page_is_cancelled(screen: Screen):
+    @ui.page('/')
+    async def page():
+        task = asyncio.current_task()
+        task.cancel()
+        ui.label('The end')
+
+    screen.start_server()
+    response = httpx.get(f'http://localhost:{Screen.PORT}/')
+    assert response.status_code == 200
+
+
 def test_page_with_args(screen: Screen):
     @ui.page('/page/{id_}')
     def page(id_: int):


evnchn:
Heads-up: this repo will soon enable GitHub's open-PR limit, capping contributors without write access at 3 open non-draft PRs. Drafts don't count and stay unlimited.

The reason is queue order. A non-draft PR reads as a request for review, and the ready list is long enough right now that things sit for weeks. A cap of 3 means you choose what gets looked at first, instead of the queue deciding by accident.

You have more than 3 open and ready. Could you pick 3 and convert the rest to draft? Drafts keep their history and you can flip them back as the first ones merge. Doing it before the setting lands avoids anything getting bounced.

SaadZahem:
Understood. I will notify you when I only have 3 non-draft PRs.

SaadZahem:
All my open PRs are now drafted except for the 3 related ones: this one and the 2 related to it.

falkoschindler:
Thanks @SaadZahem — verified and approving.

What I checked: `task.cancelled()` fully guards `task.result()` — CPython marks the task cancelled even when `cancel()` lands after the last await (`_must_cancel` path) and when the coroutine raises `CancelledError` itself, so there is no gap where `.result()` could still throw. The `response_timeout` branch returns before reaching the changed line, and the late path (`check_for_late_return_value`) already swallows `CancelledError`, so the one-liner makes the early path consistent with existing semantics rather than inventing new ones — serving the partially built page with 200 is exactly what the `else` branch below already does for builders still running at response time. Also reproduced the measurement: the new test fails on main with `RuntimeError: No response returned.` and passes with the fix.

I pushed a follow-up commit that (1) logs a warning when the builder task ends cancelled, mirroring the timeout branch, so an externally cancelled build doesn't turn into a silently truncated 200 page with no trace in the log, (2) additionally asserts the served body and the new warning in the test, and (3) aligns the test with the file's conventions (plain `def`, expectation-stating assert messages).

I agree with the issue's own assessment that this is latent today and only becomes reachable with the cancel-the-caller pattern from #6271/#6294 — fixing it ahead of those is the right order. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Verification%20and%20wording%20by%20the%20model%2C%20based%20on%20a%20multi-agent%20code%20review%3B%20approval%20by%20the%20reviewer.&a=claude-code "claude-fable-5: Verification and wording by the model, based on a multi-agent code review; approval by the reviewer.")

falkoschindler (approved):

PR #6299, opened 2026-08-19
Add replace parameter to Element.on and ui.button.on_click

Mohameme7:
### Motivation

Discussion [#3440](https://github.com/zauberzeug/nicegui/discussions/3440) ("Removing an existing handler") asks how to swap out a handler that is already registered on an element.

Today `Element.on()` — and therefore `ui.button.on_click()` — always *appends*: each call stores a new `EventListener` under a fresh id, so the previously registered handler keeps firing alongside the new one. There is no public way to replace a handler, which can lead to unintended behavior when the same event is bound a second time.

### Implementation

`Element.on()` gets an opt-in `replace: bool = False` parameter. When it is `True`, listeners of the *same* event type are dropped before the new listener is registered:

```py
if replace:
    self._event_listeners = {
        id_: l for id_, l in self._event_listeners.items() if l.type != event_type
    }
self._event_listeners[listener.id] = listener
```

Filtering on `l.type` keeps listeners for other event types untouched — an element with both a click and a hover handler only loses its click handler when `on('click', ..., replace=True)` is called.

`ui.button.on_click()` forwards the parameter, so the common case reads:

```py
button = ui.button('Click me', on_click=first_handler)
...
button.on_click(second_handler, replace=True)  # first_handler no longer fires
```

The default is `False`, so existing behavior is unchanged and nothing else needs to be touched.

Open points:

- Only `ui.button.on_click()` forwards `replace` so far. The same forwarding could be added to the `on_*` convenience methods of other elements; this PR covers the case I needed.
- No pytests yet — happy to add one if you tell me where it fits best.
- No documentation page was changed; the new parameter is described in the `Element.on()` and `on_click()` docstrings.

### Progress

- [ ] Pytests have been added/updated, or the **Implementation** section explains why they are not necessary.

### DIFF
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -386,8 +386,8 @@ class Element(Visibility):
         :param leading_events: whether to trigger the event handler immediately upon the first event occurrence (default: ``True``)
         :param trailing_events: whether to trigger the event handler after the last event occurrence (default: ``True``)
         :param js_handler: JavaScript function that is handling the event on the client (default: "(...args) => emit(...args)")
-        :param replace: whether to replace the event or append to existing events
-        :param key : an identifier key to distinct between internally and externally generated elements
+        :param replace: whether to replace a listener with the same ``type`` and ``key`` instead of adding another one (default: ``False``)
+        :param key: object identifying this listener, so that it can be replaced later (default: ``None``)
         """
         if handler or js_handler:
             event_type = helpers.event_type_to_camel_case(type)
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -357,7 +357,8 @@ class Element(Visibility):
            leading_events: bool = True,
            trailing_events: bool = True,
            js_handler: str = '(...args) => emit(...args)',
-           replace: bool = False
+           replace: bool = False,
+           key: Any = None
            ) -> Self:
         """Subscribe to an event.
 
@@ -386,6 +387,7 @@ class Element(Visibility):
         :param trailing_events: whether to trigger the event handler after the last event occurrence (default: ``True``)
         :param js_handler: JavaScript function that is handling the event on the client (default: "(...args) => emit(...args)")
         :param replace: whether to replace the event or append to existing events
+        :param key : an identifier key to distinct between internally and externally generated elements
         """
         if handler or js_handler:
             event_type = helpers.event_type_to_camel_case(type)
@@ -399,11 +401,20 @@ class Element(Visibility):
                 leading_events=leading_events,
                 trailing_events=trailing_events,
                 request=storage.request_contextvar.get(),
+                key=key
             )
-            if replace:
-                self._event_listeners = {
-                    id_: l for id_, l in self._event_listeners.items() if l.type != event_type
-                }
+            if replace and key is not None:
+                existing = next((listener for listener in self._event_listeners.values()
+                                 if listener.type == event_type and listener.key == key), None)
+                if existing is not None:
+                    old_dict = {**existing.to_dict(), 'listener_id': ''}
+                    new_dict = {**listener.to_dict(), 'listener_id': ''}
+                    existing.handler = handler
+                    existing.request = listener.request
+                    if old_dict == new_dict:
+                        return self
+                    del self._event_listeners[existing.id]
+
             self._event_listeners[listener.id] = listener
             self.update()
         return self
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -404,16 +404,15 @@ class Element(Visibility):
                 key=key
             )
             if replace and key is not None:
-                existing = next((listener for listener in self._event_listeners.values()
-                                 if listener.type == event_type and listener.key == key), None)
-                if existing is not None:
-                    old_dict = {**existing.to_dict(), 'listener_id': ''}
-                    new_dict = {**listener.to_dict(), 'listener_id': ''}
-                    existing.handler = handler
-                    existing.request = listener.request
-                    if old_dict == new_dict:
-                        return self
-                    del self._event_listeners[existing.id]
+                previous = [old for old in self._event_listeners.values()
+                            if old.type == event_type and old.key == key]
+                new_dict = {**listener.to_dict(), 'listener_id': ''}
+                if len(previous) == 1 and {**previous[0].to_dict(), 'listener_id': ''} == new_dict:
+                    previous[0].handler = handler
+                    previous[0].request = listener.request
+                    return self
+                for old in previous:
+                    del self._event_listeners[old.id]
 
             self._event_listeners[listener.id] = listener
             self.update()
--- a/nicegui/element.py
+++ b/nicegui/element.py
@@ -357,6 +357,7 @@ class Element(Visibility):
            leading_events: bool = True,
            trailing_events: bool = True,
            js_handler: str = '(...args) => emit(...args)',
+           replace: bool = False
            ) -> Self:
         """Subscribe to an event.
 
@@ -384,11 +385,13 @@ class Element(Visibility):
         :param leading_events: whether to trigger the event handler immediately upon the first event occurrence (default: ``True``)
         :param trailing_events: whether to trigger the event handler after the last event occurrence (default: ``True``)
         :param js_handler: JavaScript function that is handling the event on the client (default: "(...args) => emit(...args)")
+        :param replace: whether to replace the event or append to existing events
         """
         if handler or js_handler:
+            event_type = helpers.event_type_to_camel_case(type)
             listener = EventListener(
                 element_id=self.id,
-                type=helpers.event_type_to_camel_case(type),
+                type=event_type,
                 args=[args] if args and isinstance(args[0], str) else args,  # type: ignore
                 handler=handler,
                 js_handler=None if js_handler == '(...args) => emit(...args)' else js_handler,
@@ -397,6 +400,10 @@ class Element(Visibility):
                 trailing_events=trailing_events,
                 request=storage.request_contextvar.get(),
             )
+            if replace:
+                self._event_listeners = {
+                    id_: l for id_, l in self._event_listeners.items() if l.type != event_type
+                }
             self._event_listeners[listener.id] = listener
             self.update()
         return self
--- a/nicegui/elements/button.py
+++ b/nicegui/elements/button.py
@@ -38,9 +38,15 @@ class Button(IconElement, TextElement, DisableableElement, BackgroundColorElemen
         if on_click:
             self.on_click(on_click)
 
-    def on_click(self, callback: Handler[ClickEventArguments]) -> Self:
-        """Add a callback to be invoked when the button is clicked."""
-        self.on('click', lambda _: handle_event(callback, ClickEventArguments(sender=self, client=self.client)), [])
+    def on_click(self, callback: Handler[ClickEventArguments], replace: bool = False) -> Self:
+        """Add a callback to be invoked when the button is clicked.
+
+        :param callback: callback to be invoked when the button is clicked
+        :param replace: replace existing callback with a new one rather than appending
+        """
+        self.on('click',
+                lambda _: handle_event(callback, ClickEventArguments(sender=self, client=self.client)), [],
+                replace=replace)
         return self
 
     def _render_markdown(self) -> str:
--- a/nicegui/elements/button.py
+++ b/nicegui/elements/button.py
@@ -39,7 +39,7 @@ class Button(IconElement, TextElement, DisableableElement, BackgroundColorElemen
         if on_click:
             self.on_click(on_click)
 
-    def on_click(self, callback: Handler[ClickEventArguments], replace: bool = False) -> Self:
+    def on_click(self, callback: Handler[ClickEventArguments], *, replace: bool = False) -> Self:
         """Add a callback to be invoked when the button is clicked.
 
         :param callback: callback to be invoked when the button is clicked
--- a/nicegui/elements/button.py
+++ b/nicegui/elements/button.py
@@ -11,6 +11,7 @@ from .mixins.text_element import TextElement
 
 
 class Button(IconElement, TextElement, DisableableElement, BackgroundColorElement):
+    _ON_CLICK_KEY = object()
 
     @resolve_defaults
     def __init__(self,
@@ -46,7 +47,7 @@ class Button(IconElement, TextElement, DisableableElement, BackgroundColorElemen
         """
         self.on('click',
                 lambda _: handle_event(callback, ClickEventArguments(sender=self, client=self.client)), [],
-                replace=replace)
+                replace=replace, key=self._ON_CLICK_KEY)
         return self
 
     def _render_markdown(self) -> str:
--- a/nicegui/event_listener.py
+++ b/nicegui/event_listener.py
@@ -18,6 +18,7 @@ class EventListener:
     leading_events: bool
     trailing_events: bool
     request: Request | None
+    key: Any = None
 
     def __post_init__(self) -> None:
         self.id = str(uuid.uuid4())


evnchn:
@Mohameme7 A bit mangled, the rendered PR description. Mind if you fix it a bit? Thanks!

Mohameme7:
> @Mohameme7 A bit mangled, the rendered PR description. Mind if you fix it a bit? Thanks!

Definitely, done.
this is my first PR so things my still be a bit messy, sorry for that.

evnchn:
@Mohameme7 I went ahead and updated the PR title and desc for you

evnchn:
Thanks for picking this up!

One thing blocks the current approach: `_event_listeners` also holds listeners that NiceGUI registers itself, on the same event types. Removing by type takes those out too:

```py
inp = ui.input()
inp.on('update:value', lambda e: print(e.args), replace=True)
# type in the browser: the handler fires, but inp.value never updates again
```

`ValueElement` registers `update:{VALUE_PROP}` (`value_element.py:53`), so `.value`, bindings and `on_value_change` all go dead. Same shape for `ui.number` (`blur`, `number.py:62`), `ui.select` (`popup-show`/`popup-hide`, `select.py:87`), and `Button.clicked()` (`button.py:65`) — the latter leaves anyone awaiting it hanging forever, with no error.

Suggestion: give `EventListener` an optional key, and scope the replacement to that key rather than to the type. Internal listeners carry no key, so they are untouchable by construction; `on_click` can pass its own key so users never see one.

Worth knowing for the implementation: if the new registration differs only in the Python handler (same args, throttle, modifiers, `js_handler`), you can assign `listener.handler` in place and keep the existing id. Then no `update()` is needed at all — the client never learns anything changed, which avoids both the element re-render and the "Event listeners changed after initial definition" warning.

Mohameme7:
> Thanks for picking this up!
> 
> One thing blocks the current approach: `_event_listeners` also holds listeners that NiceGUI registers itself, on the same event types. Removing by type takes those out too:
> 
> ```python
> inp = ui.input()
> inp.on('update:value', lambda e: print(e.args), replace=True)
> # type in the browser: the handler fires, but inp.value never updates again
> ```
> 
> `ValueElement` registers `update:{VALUE_PROP}` (`value_element.py:53`), so `.value`, bindings and `on_value_change` all go dead. Same shape for `ui.number` (`blur`, `number.py:62`), `ui.select` (`popup-show`/`popup-hide`, `select.py:87`), and `Button.clicked()` (`button.py:65`) — the latter leaves anyone awaiting it hanging forever, with no error.
> 
> Suggestion: give `EventListener` an optional key, and scope the replacement to that key rather than to the type. Internal listeners carry no key, so they are untouchable by construction; `on_click` can pass its own key so users never see one.
> 
> Worth knowing for the implementation: if the new registration differs only in the Python handler (same args, throttle, modifiers, `js_handler`), you can assign `listener.handler` in place and keep the existing id. Then no `update()` is needed at all — the client never learns anything changed, which avoids both the element re-render and the "Event listeners changed after initial definition" warning.


So add an optional field in EventListener, then pass it to .on in lets say on_click on and so on, and then the user can create a new event and pass a key then we check if that key exists and remove it rather than wiping all events of same type
so for example 
```py
ui.button('Test')
button.on_click(lambda : print(24), key="Anykey")
```
then if we want to change it later
```py
button.on_click(lambda: print(23), key ="Anykey", replace=True)
```

and the logic in .on
```py
            if replace:
                self._event_listeners = {
                    id_: l for id_, l in self._event_listeners.items() if l.key != key
                }
```
?

evnchn:
Yes, that's the shape. I built it locally to check the advice actually holds, and it does — with three corrections before you code it.

**`key=None` must not match.** With `key: str | None = None`, a bare `on('click', ..., replace=True)` keeps only listeners whose key *isn't* `None` — which deletes every internal listener on the element, across every event type. Strictly worse than the type-based version. So `replace` should only ever remove an explicitly keyed listener. (Whether the unkeyed case is a silent no-op or a `ValueError` is a real design choice — worth picking deliberately rather than falling into.)

**Match on `(type, key)`, not `key` alone.** Keys are user strings, so the same one on `'click'` and `'keydown'` would clobber. You already hoisted `event_type` in this PR, so it's free.

**`on_click` shouldn't take a `key`** — and its own key shouldn't be a string. Give it a module-private sentinel:

```py
_ON_CLICK_KEY = object()
...
self.on('click', ..., key=_ON_CLICK_KEY, replace=replace)
```

A string like `key='on_click'` is still *public*: a user writing `button.on('click', f, key='on_click', replace=True)` silently replaces the `on_click` callback, and vice versa. A sentinel can't be collided with by construction. That widens `key` to `Any`, which is fine — nothing serializes it to the client.

On the in-place part: your sketch still rebuilds the dict and calls `update()`, so the client sees a listener id it hasn't seen and re-renders the element with an `"Event listeners changed after initial definition"` warning on **every** replace (`nicegui.js:507`). If the matched listener differs only in the Python handler, assign `listener.handler` (and `listener.request`) and return — same id, no `update()`, no warning. Fall back to remove-and-add only when `args`, `throttle`, `leading_events`, `trailing_events` or `js_handler` actually differ, since those are what `to_dict()` sends to the client. Comparing `{**old.to_dict(), 'listener_id': ''}` against the same for the new listener is an easy way to express exactly that.

For the tests: `tests/test_events.py` is the right home — `test_late_event_registration` is the closest neighbour and already asserts on that warning, which makes it a natural place to check that the replace path *doesn't* trigger it. Two things worth guarding there: that an `ui.input`'s `on_change` still fires after a keyed `replace=True` on `update:value` (that's the internal-listener regression), and that no re-render warning shows up.

Nice first PR, by the way — the awkward part here is entirely NiceGUI's, not yours: `_event_listeners` mixes ours and yours with nothing distinguishing them, and `key` is what finally does.

Mohameme7:
> Yes, that's the shape. I built it locally to check the advice actually holds, and it does — with three corrections before you code it.
> 
> **`key=None` must not match.** With `key: str | None = None`, a bare `on('click', ..., replace=True)` keeps only listeners whose key _isn't_ `None` — which deletes every internal listener on the element, across every event type. Strictly worse than the type-based version. So `replace` should only ever remove an explicitly keyed listener. (Whether the unkeyed case is a silent no-op or a `ValueError` is a real design choice — worth picking deliberately rather than falling into.)
> 
> **Match on `(type, key)`, not `key` alone.** Keys are user strings, so the same one on `'click'` and `'keydown'` would clobber. You already hoisted `event_type` in this PR, so it's free.
> 
> **`on_click` shouldn't take a `key`** — and its own key shouldn't be a string. Give it a module-private sentinel:
> 
> ```python
> _ON_CLICK_KEY = object()
> ...
> self.on('click', ..., key=_ON_CLICK_KEY, replace=replace)
> ```
> 
> A string like `key='on_click'` is still _public_: a user writing `button.on('click', f, key='on_click', replace=True)` silently replaces the `on_click` callback, and vice versa. A sentinel can't be collided with by construction. That widens `key` to `Any`, which is fine — nothing serializes it to the client.
> 
> On the in-place part: your sketch still rebuilds the dict and calls `update()`, so the client sees a listener id it hasn't seen and re-renders the element with an `"Event listeners changed after initial definition"` warning on **every** replace (`nicegui.js:507`). If the matched listener differs only in the Python handler, assign `listener.handler` (and `listener.request`) and return — same id, no `update()`, no warning. Fall back to remove-and-add only when `args`, `throttle`, `leading_events`, `trailing_events` or `js_handler` actually differ, since those are what `to_dict()` sends to the client. Comparing `{**old.to_dict(), 'listener_id': ''}` against the same for the new listener is an easy way to express exactly that.
> 
> For the tests: `tests/test_events.py` is the right home — `test_late_event_registration` is the closest neighbour and already asserts on that warning, which makes it a natural place to check that the replace path _doesn't_ trigger it. Two things worth guarding there: that an `ui.input`'s `on_change` still fires after a keyed `replace=True` on `update:value` (that's the internal-listener regression), and that no re-render warning shows up.
> 
> Nice first PR, by the way — the awkward part here is entirely NiceGUI's, not yours: `_event_listeners` mixes ours and yours with nothing distinguishing them, and `key` is what finally does.

I've applied the changes you suggested and did some tests and things seemed pretty good.
And thank you so much for the compliment <3

evnchn (commented):
Read the new version — the sentinel, the `(type, key)` scoping and the `key is not None` guard all look right, and I ran it locally: `ruff`, `pylint` and `mypy` are clean, and a regression test for the internal-listener case passes against your branch unchanged.

One real bug left, plus two doc nits, inline.

On tests — this is the part still missing, and it's the part that would have caught the bug above. `tests/test_events.py` next to `test_late_event_registration` is the natural home; two things worth asserting are that an `ui.input`'s `on_change` still fires after a keyed `replace=True` on `update:value`, and that no `"Event listeners changed after initial definition"` warning shows up when only the Python handler changed.

nicegui/element.py
```diff
@@ -384,19 +386,35 @@ def on(self,
...
+                existing = next((listener for listener in self._event_listeners.values()
+                                 if listener.type == event_type and listener.key == key), None)
+                if existing is not None:
+                    old_dict = {**existing.to_dict(), 'listener_id': ''}
+                    new_dict = {**listener.to_dict(), 'listener_id': ''}
+                    existing.handler = handler
+                    existing.request = listener.request
+                    if old_dict == new_dict:
+                        return self
+                    del self._event_listeners[existing.id]
```
evnchn: **`next()` only replaces the first match, so `replace=True` can leave a stale handler behind** (major)

Appending is `on_click`'s documented default, so two listeners can legitimately share the key:

```py
b = ui.button('x', on_click=a)
b.on_click(b_handler)                 # second listener, same key
b.on_click(c_handler, replace=True)
```

Clicking then fires `c` **and** `b_handler` — two listeners survive. Collecting all matches instead of the first fixes it, and it lets the in-place check be `len(previous) == 1`, so the fast path can't silently apply while a duplicate is still registered:

```suggestion
            if replace and key is not None:
                previous = [old for old in self._event_listeners.values()
                            if old.type == event_type and old.key == key]
                new_dict = {**listener.to_dict(), 'listener_id': ''}
                if len(previous) == 1 and {**previous[0].to_dict(), 'listener_id': ''} == new_dict:
                    previous[0].handler = handler
                    previous[0].request = listener.request
                    return self
                for old in previous:
                    del self._event_listeners[old.id]
```

Same change also moves the `existing.handler = ...` assignment after the equality check — as written it mutates the listener before deciding whether to keep it, which is harmless today only because neither `handler` nor `request` appears in `to_dict()`.

nicegui/element.py
```diff
@@ -384,19 +386,35 @@ def on(self,
         :param leading_events: whether to trigger the event handler immediately upon the first event occurrence (default: ``True``)
         :param trailing_events: whether to trigger the event handler after the last event occurrence (default: ``True``)
         :param js_handler: JavaScript function that is handling the event on the client (default: "(...args) => emit(...args)")
+        :param replace: whether to replace the event or append to existing events
+        :param key : an identifier key to distinct between internally and externally generated elements
```
evnchn: **Docstring for `key` won't render** (minor)

`:param key :` has a space before the colon, and the rendered docs do `description.replace(':param ', ':')`, so this one won't come out as a field like its neighbours. The wording also says "elements" where it means listeners.

Worth saying here too that `replace` is a no-op when no `key` is given — right now the text promises unconditional replacement.

```suggestion
        :param replace: whether to replace a listener with the same ``type`` and ``key`` instead of adding another one (default: ``False``)
        :param key: object identifying this listener, so that it can be replaced later (default: ``None``)
```

nicegui/elements/button.py
```diff
@@ -38,9 +39,15 @@ def __init__(self,
         if on_click:
             self.on_click(on_click)
 
-    def on_click(self, callback: Handler[ClickEventArguments]) -> Self:
-        """Add a callback to be invoked when the button is clicked."""
-        self.on('click', lambda _: handle_event(callback, ClickEventArguments(sender=self, client=self.client)), [])
+    def on_click(self, callback: Handler[ClickEventArguments], replace: bool = False) -> Self:
```
evnchn: **Keyword-only `replace`** (minor)

`Element.on` makes everything after `args` keyword-only; matching that here keeps the positional signature free to grow later.

```suggestion
    def on_click(self, callback: Handler[ClickEventArguments], *, replace: bool = False) -> Self:
```

An `*Added in version …*` line above the `:param` list would fit the house docstring style too — the maintainers can tell you which release number to put.

Mohameme7:
> Read the new version — the sentinel, the `(type, key)` scoping and the `key is not None` guard all look right, and I ran it locally: `ruff`, `pylint` and `mypy` are clean, and a regression test for the internal-listener case passes against your branch unchanged.
> 
> One real bug left, plus two doc nits, inline.
> 
> On tests — this is the part still missing, and it's the part that would have caught the bug above. `tests/test_events.py` next to `test_late_event_registration` is the natural home; two things worth asserting are that an `ui.input`'s `on_change` still fires after a keyed `replace=True` on `update:value`, and that no `"Event listeners changed after initial definition"` warning shows up when only the Python handler changed.

applied the suggestions

evnchn:

**Good:** ran the branch, and everything from the last round checks out. Duplicate-key replace leaves one listener, `ui.input`'s internal `update:value` listener survives a keyed replace, `Button.clicked()`'s unkeyed one is untouched, and the in-place path keeps the listener id with no re-render warning.

**Left on your side:**
- A test in `tests/test_events.py` next to `test_late_event_registration`; one function covers all three cases above.
- The PR description, which still shows the old `l.type` filtering and says "no pytests yet".

**For a maintainer, not you:** `replace` already means a string of replacement values on `Element.classes` and `Element.style`, so the same keyword would carry two shapes on one class.

falkoschindler:
@Mohameme7 Thanks for taking this on, and sorry for weighing in this late. I've read the discussion and the current branch, and I'd like to step back from the implementation for a moment, because I don't think this is the right shape for the API.

**Why I hesitate:**

- Only `Button.on_click` gets `replace`. NiceGUI has 39 `on_*` convenience methods (eight of them named `on_click`), and users would rightly expect `ui.input.on_value_change(..., replace=True)` or `ui.link.on_click(..., replace=True)` to work the same way. Extending this consistently means a private sentinel key and a new parameter on every one of them. That's a lot of API surface for a feature that started as a one-line question.
- `key` and `replace` are two parameters for one concept, and each is a silent no-op without the other. On top of that, `replace` already means something different on `classes()` and `style()`.
- Listeners registered via `on_click` and via `on('click', ...)` live in separate worlds: the sentinel key is private, so neither side can ever replace the other. That's deliberate, but hard to explain to users.
- The in-place handler swap exists to avoid the client-side re-render and its "Event listeners changed after initial definition" warning. It only works when nothing but the Python handler changes. Change the throttle or the args, and you're back to the re-render. So the feature's behavior depends on implementation details the user can't see.

The root problem is that NiceGUI currently has no way to *remove* a listener cleanly, on the client as well as on the server. If we want to solve this, I'd rather do it properly: an unsubscribe primitive along the lines of the DOM's `removeEventListener` (something like `element.off(...)`), with client support so removed listeners disappear without a re-render warning and without the client emitting events nobody consumes anymore. That's a design discussion first, not a small PR, and I'd suggest opening an issue for it.

Until then, the recommendation from #3440 still holds: register the handler once and let it decide what to do, e.g. by calling whatever is currently stored in a variable.

So I'm going to close this PR, and I want to be clear that this is not a judgement on your work. The awkward part is entirely on NiceGUI's side, and the review rounds pushed the implementation further than the original question warranted, which is on us, not on you. For a first contribution this was well done, and if you'd like to pursue the `off()` direction, I'm happy to discuss the design in an issue.

@evnchn thanks for the thorough reviews here. I know this is a step back from where the discussion had gone, but I'd rather settle the target API first before we grow more special cases on top of `_event_listeners`. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Assessment%20and%20wording%20from%20the%20model%2C%20decision%20and%20direction%20from%20the%20maintainer.&a=claude-code "claude-fable-5-1: Assessment and wording from the model, decision and direction from the maintainer.")

PR #6307, opened 2026-08-23
Add type annotations to observables

paco-sevilla:
### Motivation

I luckily found the *Observable Collections* (they are undocumented, see #3284). They are super useful, specially in combination with `binding.bindable_dataclass`!

This PR adds type annotations, so that code that depends on them can resolve the underlying types correctly.

### Implementation

The implementation completely avoids `Any`. Instead, the use `object` is preferred where any type is permissible.
The `_MISSING` object is replaced with `...` (`EllipsisType`).
Some `cast`s and `# type: ignore` comments are added where absolutely necessary.

The `tests/test_observables.py` is now also checked by Mypy.

### Progress

(Documentation for observable types should probably be added, maybe in a separate PR?)


nicegui/observables.py
```diff
@@ -175,37 +206,38 @@ def clear(self) -> None:
...
         self._handle_change()
 
-    def setdefault(self, __key: Any, __default: Any = None) -> Any:
+    def setdefault(self, __key: _KT, __default: _VT | None = None) -> _VT:
         if __key in self:
             return super().__getitem__(__key)
-        item = super().setdefault(__key, self._observe(__default))
+        item = super().setdefault(__key, self._observe(cast(_VT, __default)))
         self._handle_change()
         return item
```
evnchn: **`setdefault()` is annotated `_VT` but can return `None`** (major) — on `ObservableDict[str, int]`, `d.setdefault('a')` reveals `int` and returns `None` at runtime. Plain `dict` avoids this by rejecting the call outright: `dict[str, int].setdefault('a')` is `[call-arg]`, and only `dict[str, int | None]` gets `int | None`.

Binding the self-type on the defaultless overload restores exact parity with `dict`.

Verified on `7770da5f`: `mypy ./nicegui tests/test_observables.py` clean across 265 files, 21/21 `tests/test_observables.py` pass, and caller-side `ObservableDict[str, int].setdefault('a')` → `[call-arg]`, `.setdefault('a', 5)` → `int`, `ObservableDict[str, int | None].setdefault('a')` → `int | None`.

```suggestion
    @overload
    def setdefault(self: ObservableDict[_KT, _T | None], __key: _KT, __default: None = None) -> _T | None:
        ...

    @overload
    def setdefault(self, __key: _KT, __default: _VT) -> _VT:
        ...

    def setdefault(self, __key: _KT, __default: _VT | None = None) -> object:
        if __key in self:
            return super().__getitem__(__key)
        item = super().setdefault(__key, self._observe(cast(_VT, __default)))
        self._handle_change()
        return item
```

nicegui/observables.py
```diff
@@ -244,42 +276,44 @@ def clear(self) -> None:
...
         self._unobserve(*items)
         self._handle_change()
 
-    def __setitem__(self, key: SupportsIndex | slice, value: Any) -> None:
+    def __setitem__(self, key: SupportsIndex | slice[SupportsIndex | None], value: _T | Iterable[_T]) -> None:
         if isinstance(key, slice):
             old_items = self[key]
+            if not isinstance(value, Iterable):
+                raise TypeError('must assign iterable to slice')
             super().__setitem__(key, [self._observe(item) for item in value])
```
evnchn: **Broken public API** (blocking) — `Iterable` only recognises `__iter__`, but `list.__setitem__` also accepts the legacy `__getitem__` sequence protocol, so this guard rejects values `list` accepts.

```python
class Legacy:  # __getitem__-only iterable
    def __init__(self, items): self._items = items
    def __getitem__(self, i):
        if i >= len(self._items): raise IndexError
        return self._items[i]

[1, 2, 3][0:2] = Legacy([9, 8])                   # [9, 8, 3]
ObservableList([1, 2, 3])[0:2] = Legacy([9, 8])   # main: [9, 8, 3] | here: TypeError
```

A `cast` narrows the slice branch without the runtime check. Verified on `7770da5f`: `mypy ./nicegui tests/test_observables.py` clean across 265 files, 21/21 pass, legacy slice-assign back to `[9, 8, 3]`.

```suggestion
            super().__setitem__(key, [self._observe(item) for item in cast('Iterable[_T]', value)])
```

.github/workflows/_check.yml
```diff
@@ -11,7 +11,7 @@ jobs:
           - name: pre-commit
             command: pre-commit run --all-files
           - name: mypy
-            command: mypy ./nicegui
+            command: mypy ./nicegui tests/test_observables.py
```
evnchn, line 14: **Declared mypy floor is too low for these annotations** (blocking) — `pyproject.toml` asks for `mypy>=1.13.0,<2`, but the new syntax needs **1.16**. An environment resolving to an allowed-but-older mypy gets a red gate on untouched lines:

```console
$ mypy==1.15.0 -> observables.py:229 __or__, :307 __add__ (operator + return-value)
$ mypy==1.13.0 -> also fails on the parameterized slice[...] annotations
$ mypy==1.16.0 -> clean
```

Worth bumping the floor here. Not suggestable inline — `pyproject.toml` is not in this diff.

evnchn:
Two findings have inline suggestions with Commit buttons. These four don't — either the fix needs a module-level line, or I couldn't get one past mypy. Only the first is a behaviour change vs `main`; the other three are runtime behaviours `main` already had, but `main` annotated them `Any`, so nothing was promised.

1. **`pop()` cannot take `...` as a default** (blocking) — `_MISSING` was private, so no caller could collide with it; `...` is public. `ObservableDict({'x': 1}).pop('missing', ...)` returns `Ellipsis` on `main` and raises `KeyError` here. The fix needs a sentinel back at module level, so it is not a single-hunk suggestion.

2. **`sort()` lost typeshed's comparability guard** (major) — typeshed gates the no-`key` overload on `self: list[SupportsRichComparisonT]`. Without it, `ObservableList[NotComparable]().sort()` type-checks and then raises at runtime, while the identical call on a plain `list` is correctly rejected. Also needs a bound TypeVar at module level.

3. **`update(**kwargs: _VT)` accepts `str` keys on any dict** (major) — typeshed gates its kwargs overloads on `self: dict[str, _VT]`. `ObservableDict[int, str]().update(foo='bar')` passes mypy and inserts `'foo'`. I tried an overload pair twice and could not clear `Overloaded function implementation does not accept all possible arguments of signature 2`.

4. **`__setitem__`'s union lets an iterable land at a single index** (major) — with `value: _T | Iterable[_T]`, `lst[0] = [7, 7]` on an `ObservableList[int]` type-checks and reveals `int`, while the stored element is an `ObservableList`. Overloading on `SupportsIndex -> _T` / `slice -> Iterable[_T]` closes it; I left it out of the one-line runtime suggestion above to keep that diff minimal.

<details>
<summary>Runnable differential — <code>python</code> reports four mismatches, <code>mypy</code> reports none</summary>

```python
"""Four places where PR 6307's annotations disagree with runtime.

    python mre_6307.py     -> prints a mismatch for each, exit 1
    mypy   mre_6307.py     -> Success: no issues found

Both passing is the bug: mypy signs off on all four.
"""
from nicegui.observables import ObservableDict, ObservableList

fails = []


def check(name: str, expected: object, actual: object) -> None:
    if expected != actual:
        fails.append(f'{name}: expected {expected!r}, got {actual!r}')


# 1. pop() default is the public `...`, so a caller cannot pass it. Regression vs main.
d1: ObservableDict[str, int] = ObservableDict({'x': 1})
try:
    check('pop default', ..., d1.pop('missing', ...))            # mypy: int
except KeyError as e:
    fails.append(f'pop default: expected Ellipsis, got KeyError({e.args[0]!r})')

# 2. sort() with no key accepts non-comparable elements; plain list rejects the same call.
class NotComparable:
    pass

l2: ObservableList[NotComparable] = ObservableList([NotComparable(), NotComparable()])
try:
    l2.sort()                                                    # mypy: ok
    fails.append('sort: expected a type error, got a silent sort')
except TypeError as e:
    fails.append(f'sort: mypy accepted the call, runtime raised TypeError({str(e)!r})')

# 3. update(**kwargs) inserts str keys into an int-keyed dict.
d3: ObservableDict[int, str] = ObservableDict()
d3.update(foo='bar')                                             # mypy: ok
check('update kwargs keys', [], [k for k in d3 if not isinstance(k, int)])

# 4. __setitem__'s `_T | Iterable[_T]` union lets an iterable land at a single index.
l4: ObservableList[int] = ObservableList([1, 2, 3])
l4[0] = [7, 7]                                                   # mypy: ok, reveals int
check('single-index element type', 'int', type(l4[0]).__name__)

for f in fails:
    print(f)
raise SystemExit(1 if fails else 0)
```

```console
$ python mre_6307.py
pop default: expected Ellipsis, got KeyError('missing')
sort: mypy accepted the call, runtime raised TypeError("'<' not supported between instances of 'NotComparable' and 'NotComparable'")
update kwargs keys: expected [], got ['foo']
single-index element type: expected 'int', got 'ObservableList'
$ echo $?
1

$ mypy mre_6307.py
Success: no issues found in 1 source file
```

Same file against `main` (`8fb479bc`) — three mismatches, `pop` passes:

```console
sort: mypy accepted the call, runtime raised TypeError("'<' not supported between instances of 'NotComparable' and 'NotComparable'")
update kwargs keys: expected [], got ['foo']
single-index element type: expected 'int', got 'ObservableList'
```

`7770da5f`, Python 3.12.11, mypy 1.19.1, macOS arm64.

</details>

paco-sevilla:
Thanks for the review @evnchn!
I'll address your comments today or tomorrow 🙂

nicegui/observables.py
paco-sevilla: Addressed here: https://github.com/zauberzeug/nicegui/pull/6307/commits/827bf58bec08b6da9d7a3580c4fe883bfb0468fc

nicegui/observables.py
paco-sevilla: Thanks for the hint! Addressed here: https://github.com/zauberzeug/nicegui/pull/6307/commits/a2d4ec35f4edb73e42a5b9de0f0c5aa5f8026c3e

.github/workflows/_check.yml
paco-sevilla: Wow, good catch! Addressed here: https://github.com/zauberzeug/nicegui/pull/6307/commits/0886b838114d62a708688608d7a92409983a4efe

paco-sevilla:
@evnchn Sorry, that it's taken a few days...
I started addressing your comments, but I've hit an issue with `uv` and the circular dependency on `nicegui-highcharts` while trying to bump the version of `mypy`:
```
$ uv sync
  × No solution found when resolving dependencies for split (markers: python_full_version >= '3.15' and python_full_version < '4'):
  ╰─▶ Because nicegui-highcharts>=3.3.0 depends on your project and nicegui[highcharts] depends on nicegui-highcharts>=3.3.0, we can conclude that nicegui[highcharts] depends on itself at an incompatible version (nicegui>=3.0.0).
      And because your project requires nicegui[highcharts], we can conclude that your project's requirements are unsatisfiable.

hint: While the active Python version is 3.10, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
hint: The package `nicegui-highcharts` depends on the package `nicegui` but the name is shadowed by your project. Consider changing the name of the project.
```
I'll try to look into this over the weekend.

evnchn:
> ... circular dependency ...

Hint: Probably your uv version rather than your diff.

On main with _only_ the mypy bump, `uv.lock` deleted and Python 3.10 pinned, `uv lock` resolved all 200 packages fine (uv 0.9.28)

Note your error says the project requires `nicegui[highcharts]`, while main only ever depends on plain `nicegui-highcharts` and never the self-referential extra... Maybe try `uv self update` and see what happens!

evnchn:
OCD has gotten me. I wanted all PRs to have a green checkmark, not a red cross. So, here's a commit. It fixes things.

evnchn:
oops, didn't fix mypy in that one

PR #6316, opened 2026-08-30
Fix `ui.codemirror` and `ui.editor` losing their HTML id (breaks tooltips)

### LINKED ISSUE #6315: `ui.codemirror` lost its ID
### First Check

- [ ] I used the GitHub search to find a similar [issue](https://github.com/zauberzeug/nicegui/issues) and came up empty.

### Example Code

```python
from nicegui import ui

  
def index():
    ui.codemirror(
        'return'
    ).tooltip('hello')

ui.run(
    root=index,
)
```

### Description

It is similar to https://github.com/zauberzeug/nicegui/issues/6114 ,
Browser's console shows:
```
quasar.umd.prod.js:7 Anchor: target "#c4" not found
```

Image

### NiceGUI Version

3.16.0

### Python Version

3.14.7

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

evnchn:
### Motivation

Fixes #6315: `ui.codemirror(...).tooltip(...)` does nothing and the console logs `Anchor: target "#c4" not found`.

`renderRecursively` passes `id: "c" + id` to every component and relies on Vue attribute fallthrough to put it on the root element. `codemirror.js` declares `id: String` in `props` (added in 9d60c9a so `beforeUnmount` can look the element up), so Vue consumes it as a prop and the root `<div>` renders without an `id`.

### Implementation

The prop exists only for the `beforeUnmount` lookup, and it is the prop declaration itself that cancels the fallthrough. So instead of re-binding `:id="id"` in the template, drop the prop and read the id from `this.$el`, as `scene.js` already does. Vue runs `beforeUnmount` before it unmounts the subtree, so `this.$el` is still the connected root element there.

`editor.js` had the same prop and compensated with an explicit `:id="id"` binding. It gets the same treatment, so no component is left that has to remember to re-bind the id. `input.js` also declares the prop, but its template has two root nodes and its DOM id comes from the Python-side `for` prop, so it stays as is.

Test: `ui.codemirror()` added to `test_even_special_elements_have_an_html_id`. It fails on main (`assert False` on `document.getElementById("c4") !== null`) and passes with the fix. The existing `test_anchors_survive_client_side_remount` covers the `beforeUnmount` path for codemirror; a throwaway remount test for `ui.editor` (tab panels with `keep_alive=False`) passed as well.

<details>
<summary>Symptom-level check (not shipped)</summary>

The reporter's actual symptom, verified with a throwaway test — fails without the fix on the console assertion, passes with it:

```python
ui.codemirror('return').tooltip('hello')
...
assert not any('Anchor: target' in m for m in logs), logs
ActionChains(screen.selenium).move_to_element(
    screen.selenium.find_element(By.CLASS_NAME, 'cm-editor')).perform()
screen.should_contain('hello')
```

Left out because it duplicates the symptom; `html_id` resolving to a real node is the contract `tooltip()`, `ui.query` and `ui.teleport` all share.

CodeMirror mounts with `parent: this.$el`, so the anchored root div stays the root.
</details>

### Progress

### DIFF
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -48,7 +48,7 @@ const { setEffect: setTooltipsEffect, field: tooltipField } = defineRemappableRa
 
 export default {
   template: `
-    <div :id="id"></div>
+    <div></div>
   `,
   props: {
     value: String,
@@ -62,7 +62,6 @@ export default {
     keymap: Array,
     lineTooltips: Object,
     lineTooltipHtml: Boolean,
-    id: String,
   },
   watch: {
     language(newLanguage) {
@@ -98,7 +97,7 @@ export default {
   },
   beforeUnmount() {
     if (this.editor) {
-      const element = mounted_app.elements[this.$props.id.slice(1)];
+      const element = mounted_app.elements[this.$el.id.slice(1)];
       if (element) {
         element.props.value = this.editor.state.doc.toString();
         // A client-side remount (e.g. a v-if container) re-applies these props against the restored
--- a/nicegui/elements/codemirror/codemirror.js
+++ b/nicegui/elements/codemirror/codemirror.js
@@ -48,7 +48,7 @@ const { setEffect: setTooltipsEffect, field: tooltipField } = defineRemappableRa
 
 export default {
   template: `
-    <div></div>
+    <div :id="id"></div>
   `,
   props: {
     value: String,
--- a/nicegui/elements/editor.js
+++ b/nicegui/elements/editor.js
@@ -1,6 +1,6 @@
 export default {
   template: `
-    <q-editor ref="qRef" :id="id" v-model="inputValue">
+    <q-editor ref="qRef" v-model="inputValue">
       <template v-for="(_, slot) in $slots" v-slot:[slot]="slotProps">
         <slot :name="slot" v-bind="slotProps || {}" />
       </template>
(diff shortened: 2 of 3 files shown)


evnchn:
@falkoschindler since this is a +2-1, I think you can review directly. It's super simple.

falkoschindler:
Thanks, the fix works and the new test is a real regression guard. I pushed 56a2c644c on top, which takes it one step further: instead of re-binding `:id="id"`, it drops the `id` prop altogether. Declaring the prop is what cancels Vue's attribute fallthrough in the first place, and its only consumer is the `beforeUnmount` lookup, which now reads `this.$el.id` like `scene.js` does. Vue runs `beforeUnmount` before it unmounts the subtree, so `$el` is still the connected root there. `editor.js` had the same prop-plus-binding pair and gets the same treatment, so no component is left that has to remember to re-bind the id. `input.js` is multi-root and gets its DOM id via the Python-side `for` prop, so it stays as is.

The codemirror, editor and html-id tests pass, including `test_anchors_survive_client_side_remount`, which goes through `beforeUnmount`. I updated the title and description to match. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Finding%2C%20code%20change%20and%20wording%20from%20the%20model%3B%20the%20maintainer%20decided%20to%20push%20it.&a=claude-code "claude-fable-5-1: Finding, code change and wording from the model; the maintainer decided to push it.")

falkoschindler (approved):

